Issue #1387, opened 2021-12-09
Refactor talentsearch and admin projects to not involve Laravel/Lumen


tristan-orourke:
# Description
Our react apps don't need to be served via Laravel/Lumen apps. They could be compiled into static files. This would require changes to js compilation, some new html files to replace the blade templates, and changes to htaccess, among other things.

Questions: 
1. Should we combine talentsearch and admin into a single project, with a single set of npm dependencies? In that case, what do we do with the Common folder? Should we use a common public folder for images, js and css, etc, or keep it split up?
2. What bundler would we use? Webpack? We've looked a bit into Vite as well, which looks very nice for local dev

Answers:
1. We can keep components divided up among talentsearch, admin and common directories, but they should all be in the same project, with only a single package.json file. Images, css, etc, may be kept in a single public folder. 
2. We will use webpack. We already use some webpack configuration, though it may need updating.

mnigh:
We will meet as a team to discuss the questions in this issue before estimating.

petertgiles:
A good writeup of how to use a top-level package.json and node_modules directory to support a common project and and two different app projects with npm 7.
https://dev.to/limal/simplify-your-monorepo-with-npm-7-workspaces-5gmj

patcon:
I'm chipping away at getting npm spaces working. This will need a resolution in hydrogen before we can merge all our deps into one workspace package.json: https://github.com/hydrogen-design-system/hydrogen.css/issues/72

patcon:
Bonus: Merging node_modules will bring our file count waaaaaaay down, and so the mounted volumes will be much less slow. I think osx mounts are particularly slow, as setup.sh takes like 10-15 min to run each time. Is that the same for everyone?

tristan-orourke:
Honestly, I'm not sure we even need workspaces. Is there any real downside to making common, admin and talentsearch a single project, with a single package.json file?

petertgiles:
> Is there any real downside to making common, admin and talentsearch a single project

That's an interesting question.  The answer is probably "technically yes but practically no".  Right now the talentsearch bundle can be a bit smaller because it doesn't have a few things like urql auth and react-table.  And we would lose the ability to have different hydrogen configs though right now they are identical.  It could be argued that preventing the projects from diverging is actually a positive thing though.  I would caution that maybe we need some clarity about how the indigenous portal and "other CMO department" will fit into the ecosystem before deciding that.

patcon:
[...]

PR #2262, opened 2022-03-16
Inject some constant into admin html at runtime

### LINKED ISSUE #2261: Allow client-side to retrieve some env variables at run-time instead of build time
# Description
Any env variables used by frontend code are currently pulled in by webpack during build time and baked into adminConstants and talentSearchConstants. However, it would be beneficial to be able to reference some values from the server at run-time, especially those which are very likely to differ in different environments, and potentially feature flags.

This can be achieved by serving a simple json file on the server which includes some env vars. This might be doable with purely apache (ie htaccess file). If not, it can be done with php.

tristan-orourke:
Closes #2261

Based on [this trick](https://create-react-app.dev/docs/title-and-meta-tags/#injecting-data-from-the-server-into-the-page), this allows some environment variables to be injected directly into the html when served. ~I tried to do with this with pure PHP (instead of passing variables from the Laravel controller) so as to conflict as little as possible with #1387.~ Environment variables are injected using [Apache SSI](https://httpd.apache.org/docs/current/howto/ssi.html). Any environment variables that we want available at run-time like this must be defined in three places: the htaccess file, the config.sjs file, and the adminConstants file (or other _____Constants file).

To test, simply run `npm run dev` in frontend/admin, and try to login, then logout. The logout will fail if this isn't working.

Contrary to the warning [referenced](https://medium.com/node-security/the-most-common-xss-vulnerability-in-react-js-applications-2bdffbcc1fa0) in the [create-react-app docs](https://create-react-app.dev/docs/title-and-meta-tags/#injecting-data-from-the-server-into-the-page), this solution doesn't seem to cause an XSS vulnerability. Try setting `OAUTH_POST_LOGOUT_REDIRECT='"};alert("You have an XSS vulnerability!");const x={"test":"''` in your admin/.env file to test it.

:warning: This PR updates the Docker infrastructure ⚠️ 
This PR updates the php dockerfile, and updates the docker-compose file to pass env variables to the apache environment. To ensure your environment is up to date, rebuild your docker image with the following commands:
```
docker-compose build
docker-compose up -d
```

### DIFF
--- a/frontend/admin/public/.htaccess
+++ b/frontend/admin/public/.htaccess
@@ -18,4 +18,16 @@
     RewriteCond %{REQUEST_FILENAME} !-d
     RewriteCond %{REQUEST_FILENAME} !-f
     RewriteRule ^ index.php [L]
+
+
+
 </IfModule>
+
+# Set SSI to run on .shtml and .sjs files in this directory
+Options +Includes
+AddType text/html .shtml
+AddType text/javascript .sjs
+AddOutputFilter INCLUDES .shtml .sjs
+
+# Make certain environment variables available for SSI
+PassEnv OAUTH_POST_LOGOUT_REDIRECT OAUTH_LOGOUT_URI
--- a/frontend/admin/public/config.sjs
+++ b/frontend/admin/public/config.sjs
@@ -1,0 +1,14 @@
+/**
+ * This javascript file uses Apache Server Side Includes (SSI) to inject server-side environment
+ * variables into the frontend.
+ * https://httpd.apache.org/docs/current/howto/ssi.html
+ *
+ * The sjs filetype lets apache know that this file should be parsed for SSI directives.
+ */
+
+const data = {
+    "OAUTH_LOGOUT_URI": "<!--#echo var="OAUTH_LOGOUT_URI" -->",
+    "OAUTH_POST_LOGOUT_REDIRECT": "<!--#echo var="OAUTH_POST_LOGOUT_REDIRECT" -->",
+}
+
+window.__SERVER_CONFIG__ = data;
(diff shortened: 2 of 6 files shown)


petertgiles:
OK, we're getting somewhere with this already!  I ran this myself and it to does seem to work just fine, including the XSS test.  I'm not sure about the PHP part though.  We're in the middle of removing PHP as a dependency, right?   Do we want to add it back in, even if it isn't Laravel?  When we were talking yesterday it sounded more like we wanted to use Apache itself to inject the variables in, using something like [mod_substitute](https://httpd.apache.org/docs/2.4/mod/mod_substitute.html).

tristan-orourke:
@petertgiles I wasn't able to get mod_substitute working with environment variables, but I figured out how to inject the variables with [Apache SSI](https://httpd.apache.org/docs/current/howto/ssi.html)!

I'm not if this is strictly better, though? We want to remove the dependency on Laravel, but it will still be served from a PHP server. And even if we want to change that to Node one day, we also want to change Apache to Nginx eventually.

tristan-orourke:
With the new version (the config.sjs file) I tested XSS with `OAUTH_POST_LOGOUT_REDIRECT='"};alert("You have an XSS vulnerability!");const x={"test":"'`

tristan-orourke:
Because the new stuff is happening outside the Laravel environment, I had to load update the docker-compose file to load the relevant env variables.
I also had to update the php container Dockerfile to load the [mod_include](https://httpd.apache.org/docs/current/mod/mod_include.html) apache module. I hope that's available on our server, or we can set it to be...

petertgiles (approved):
Yeah, it works for me!  It's a bit different that now you have to rerun docker-compose to make changes to the .env file. :grin:

> I'm not if this is strictly better, though?

I think it really is!  Two thoughts:
1. As we switch to the "best practices docker" with one container per service this will allow a much simpler container to be used for the frontend project.
2. We have our frontend and backend on the same server right now but it might not always be like that.  A fully static site serve for just the frontend could even be done from a much simpler server (or many servers).  It could help us to avoid "to use PHP 8 you have to switch to ngninx" kinds of issues.

PR #2268, opened 2022-03-18
API_APP_URL to APP_URL

petertgiles:
Changes the environment variable used in the _API_ project from `API_APP_URL` to `APP_URL`.  The _API_ project uses this variable for post login redirect with the _from_ parameter.  So if _from_ is /admin/pools they will be redirected to API_APP_URL/admin/pools.  The value of API_APP_URL on the server is tbs-dev.whatever/api so you end up with tbs-dev.whatever/api/admin/pools which fails.  I've switched it to using the APP_URL variable which is just tbs-dev.whatever and can properly accept a relative url.  There's no routing for /api anyway so that variable may not even need to exist.

### DIFF
--- a/.github/workflows/check-lighthouse-schema.yml
+++ b/.github/workflows/check-lighthouse-schema.yml
@@ -17,10 +17,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2.4.0
 
-      - name: Create admin .env file, which is referenced in docker-compose
-        working-directory: frontend/admin
-        run: copy .env.example .env
-
       - name: "Build docker image: api"
         working-directory: infrastructure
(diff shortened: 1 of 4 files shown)


petertgiles:
I don't understand why the check schema action is failing.  I'll check on Monday.

tristan-orourke (commented):
Nice! So this solves the redirect issue. Could the failing test have something to do with the fact that I added a reference to the admin env file to our docker-compose?
```
env_file:
      - ../frontend/admin/.env
```

edit: Yes, I'm pretty sure this is the issue.

tristan-orourke (approved):
Makes sense!

PR #2315, opened 2022-03-23
Bashrc for script runs

patcon:
Re-ticketed from [Slack message](https://talent-cloud.slack.com/archives/C0259G6KXJ6/p1648051752658949?thread_ts=1647896080.351719&cid=C0259G6KXJ6).

We're running bash scripts in docker with `bash myscript.sh`. This works fine, but we're running scripts that expect nvm to be loaded. But nvm is loaded as part of bashrc. We've added some work-arounds to scripts to manually load NVM, but there may be other edge-cases where the "script" environment is unlike the "user" environment.

To bring this more inline, we can start running our scripts as if they were login shells, which I understand to be very common (and I've used this a lot in the past).

Basically:

```sh
bash myscript.sh
# becomes...
bash --login myscript.sh
```

Both `--login` and `-i` (for "interactive" shell) can be used to ensure bashrc is loaded, but I suspect "interactive" might tell underlying tools that we're in an interactive environment, and perhaps surface "user prompts" that we don't want.

See: [full `bash` man page](https://www.man7.org/linux/man-pages/man1/bash.1.html#:~:text=one%20started%20with%20the%20%2D%2Dlogin%20option)

### To Do
- [x] update maintenance README
- [x] update .vscode/tasks.json
- [x] ~~decide if stretch goal is here, or spawn another ticket :) -- @petertgiles @patcon~~ `wont-do`
- [x] ~~stretch goal? add makefile to abstract this stuff away in future~~ `wont-do`
  - ~~so that we have a central place to run these that can be updated, rather than storing it in people's brains~~
  - ~~ie. we now have to pester ppl to slowly stop using `docker-compose run [...] bash myscript.sh`, instead of just swapping out what `docker-compose run [...] make myscript` (or `make myscript`) does under the covers, with no comms necessary 🙃~~
- [x] revert 😐 
- [x] pull snippet into own file
- [x] document some non-obvious things (hashbang, using `sh`, explain nvm caveats)
- [x] test all scripts as working _independantly_: `setup.sh` and `refresh_*.sh`
- ~~stretch goals?~~
  - [ ] ~~add `make` to container (also workable outside?)~~
    - symlink to get it working in both places? https://stackoverflow.com/a/31885214
  - [ ] ~~shift `task.json` to running `make` inside container~~
  - [ ] ~~add `checkout.sh` script to have "idempotent" command to run on checkouts.~~
    - ~~to run "reload" steps when switching branches. here, just rebuilding docker maintenance container to get the `make` install~~

### DIFF
--- a/maintenance/scripts/lib/load_nvm.sh
+++ b/maintenance/scripts/lib/load_nvm.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env sh
+
+# In the hashbang line above, we use `sh` instead of `bash` as the script
+# shell, since `sh` is more portable across systems. Be aware of differences
+# with built-in functions and flow control syntax: e.g., `.` vs `source`
+# notation, `[ ]` vs `[[ ]]` conditionals, etc.
+# See: https://stackoverflow.com/a/5725402
+
+# Setup NVM.
+#
+# We must reload it all scripts, as NVM expects to be run via a login or
+# interactive user shell, and loads itself accordingly in ~/.bashrc . But
+# scripts don't load files like this.
+export NVM_DIR="$HOME/.nvm"
+[ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh"
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -8,7 +8,7 @@
       "type": "shell",
 
       "command": "docker-compose",
-      "args": ["run", "--rm", "maintenance", "bash", "refresh_auth.sh"],
+      "args": ["run", "--rm", "maintenance", "bash", "--login", "refresh_auth.sh"],
 
       "options": { "cwd": "infrastructure" },
       "presentation": {
@@ -26,7 +26,7 @@
       "type": "shell",
 
       "command": "docker-compose",
-      "args": ["run", "--rm", "maintenance", "bash", "refresh_api.sh"],
+      "args": ["run", "--rm", "maintenance", "bash", "--login", "refresh_api.sh"],
 
       "options": { "cwd": "infrastructure" },
       "presentation": {
@@ -44,7 +44,7 @@
       "type": "shell",
 
       "command": "docker-compose",
-      "args": ["run", "--rm", "maintenance", "bash", "refresh_common.sh"],
+      "args": ["run", "--rm", "maintenance", "bash", "--login", "refresh_common.sh"],
 
       "options": { "cwd": "infrastructure" },
       "presentation": {
@@ -62,7 +62,7 @@
       "type": "shell",
 
       "command": "docker-compose",
-      "args": ["run", "--rm", "maintenance", "bash", "refresh_talentsearch.sh"],
+      "args": ["run", "--rm", "maintenance", "bash", "--login", "refresh_talentsearch.sh"],
 
       "options": { "cwd": "infrastructure" },
       "presentation": {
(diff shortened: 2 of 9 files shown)


petertgiles:
This change seems to work just fine for me, including the VS code tasks.  Wouldn't we want to remove these blocks now?  This is the workaround that this change aims to eliminate, right?
```
export NVM_DIR="$HOME/.nvm"
[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
```

I think makefiles would require another dependency to install for our Windows friends so I'm not seeing the value for that change yet.

patcon:
> Wouldn't we want to remove these blocks now?

Done! cc @tristan-orourke as I removed from deploy.sh as well, under assumption that bash scripts should always be run with `bash --login foo.sh` now. Should you document usage of that script somewhere, as no mention of how/where/when its run is made anywhere.

> I think makefiles would require another dependency to install for our Windows friends so I'm not seeing the value for that change yet.

Agree that these can be out-of-scope here, but as discusses at dockercon, make is installed/run inside container, so no issue for windows users :)

patcon:
Ready for review/merge

petertgiles:
This didn't work for me, unfortunately.  I got `nvm: command not found` when I ran it.

To try to help the shell find the config I trie adding the nvm stuff to `/etc/profile` in the maintenance container during nvm setup:
```
RUN wget -qO- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | PROFILE=/etc/profile bash
```

This worked for individual scripts, like `refresh_admin` or `refresh_talentsearch`.  I still couldn't get it to work in `refresh_all` where the script is calling a second script.  Any ideas?

patcon:
I am a bit ashamed to learn through troubleshooting this that I have been doing this the wrong way for years... :/

I'm going to switch the approach in a tiny way, just to hint at the next dolt like me as to how to properly do this.

petertgiles:
Cool!  I'll put in draft status while you work on the new approach.

patcon:
Bah, I'm realizing my pyschology here. This is small PR. It's ready for review. It's pretty inconquential. I'm ashamed how small it ended up being. I added the stretch goals because of that shame. I've now removed them, as they're scope-creep.

@petertgiles This PR can either be (1) reviewed and merged quickly, or (2) closed without merge. Because screw the sunk cost fallacy 🙃

maintenance/scripts/lib/load_nvm.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env sh
+
+# In the hashbang line above, we use `sh` instead of `bash` as the script
+# shell, since `sh` is more portable across systems. Be aware of differences
+# with built-in functions and flow control syntax: e.g., `.` vs `source`
+# notation, `[ ]` vs `[[ ]]` conditionals, etc.
+# See: https://stackoverflow.com/a/5725402
```
patcon, line 7: I'd like to later move this up to top-level, and convert all scripts to `sh`, but out-of-scope here. Will do it in a more suitable PR.

petertgiles: I like it.  Throw it up on the board and let's do it!

petertgiles: I'd like to move all the update_* scripts to the lib folder, too.

petertgiles (approved):
Woo!  Pushed it across the finish line.  :1st_place_medal:

PR #2317, opened 2022-03-23
User shouldn't be able to mutate own sub field

tristan-orourke:
Removes sub field from UpdateUserAsUser mutation.


petertgiles (approved):
Good catch!  Builds for me.

Issue #2252, opened 2022-03-14
Run smoketest on Prod with SiC client credentials


tristan-orourke:
- [x] add OAUTH_LOGOUT_URI and OAUTH_POST_LOGOUT_REDIRECT to env variables
- [x] add production SiC client credentials to kv
- [x] Add OAUTH_LOGOUT_URI and OAUTH_POST_LOGOUT_REDIRECT env variables to build pipeline (they are required in frontend code)
- [ ] Logout env variables are different for each environment, which means we need three build artifacts/pipelines again. Or logout requires a refactor...

PR #2318, opened 2022-03-23
Update acr_values to enable gckey-passthrough

tristan-orourke:
During final testing of our connection to the production SiC service, we realized that we no longer need to include acr_values=mfa in our request. MFA has been configured as mandatory for our SiC client account. Instead, we can set acr_values=gckey to enable gckey-passthrough, or sending users directly to the gckey login page, ignoring the option of signing in with a bank.

Related to #2252


petertgiles (approved):
Big improvement!  Nice!

PR #2330, opened 2022-03-25
Restore field pick for user form

### LINKED ISSUE #2321: Cannot update users through admin panel
# Description
After recent changes to updateUser mutation, updating users through the Admin panel doesn't work.

# Steps to reproduce
1. Go to User table on admin portal
2. Click edit on existing user
3. Ensure all fields contain values
4. Click submit
5. See error

petertgiles:
This branch restores the field pick logic for the user edit form.

Closes #2321

### DIFF
--- a/frontend/admin/resources/js/components/user/UpdateUser.tsx
+++ b/frontend/admin/resources/js/components/user/UpdateUser.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import { useIntl } from "react-intl";
 import { FormProvider, SubmitHandler, useForm } from "react-hook-form";
+import pick from "lodash/pick";
 import { toast } from "react-toastify";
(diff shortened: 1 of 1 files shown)


vd1992:
Checked out Main, refreshed environment, attempted to replicate error, error reproduced
Checked out this branch, refreshed environment, attempted to replicate error, error did not appear
Issue appears resolved

Nice work Peter, I think things are solved? Not sure if anything else was broken.
Solution makes sense, cannot update all those other fields newly added to User, need to exclude them for the time being. Will need to adjust the forms later for the new fields though?

petertgiles:
> Will need to adjust the forms later for the new fields though?

Yeah, I imagine the form will probably be enriched with more fields later.

tristan-orourke (approved):
Code looks solid, I love how simple a change it is!

PR #2357, opened 2022-03-29
Add mutations for experiences

### LINKED ISSUE #1789: API mutations to Create, Edit and Delete Experiences
# Description
Add mutations to the API schema to create, edit, and delete experiences. Note that experiences comprise 5 different types of models. Unfortunately, each type needs its own mutations, ie `createWorkExperience` + `updateWorkExperience` + `deleteWorkExperience`, `createPersonalExperience` + `updatePersonalExperience` + `deletePersonalExperience`, etc. Sorry, this issue might get a bit tedious. 

These will need to be protected by policies, which means using the `@guard`, `@can(ability: "create")` and `@can(ability: "update", find: "id")` directives. At least you should only need to define a single ExperiencePolicy which can apply to all the Experience models.

[Larvel documentation on Policies](https://laravel.com/docs/9.x/authorization#writing-policies)

Ideally, we'd want to update the Experience-Skill connection with the `sync with pivot data` [nested mutation](https://lighthouse-php.com/master/eloquent/nested-mutations.html#storing-pivot-data) pattern. This would let us specify the skill id and description text we want to associate with an experience, without having to worry about whether that ExperienceSkill already exists. This might require a bit of refactoring of the Experience and Skills models, so that Laravel considers them "relationship + intermediate pivot data", instead of treating ExperienceSkill as its own model. Maybe we can have both? Honestly, still unclear - figuring this out will be the most challenging part of this story. 
  -- @tristan-orourke is ready to help with this part even if not tackling the whole story

Finally, in addition to defining `createWorkExperience` etc mutations, it would also be convenient to allow these operations as nested mutations on the createUser and updateUser models. You should be able to reuse the intput types you already created for the direct mutations. eg something like this:
```graphql
mutation {
  createUser(user: CreateUserInput!)
  createWorkExperience(userId: ID!, experience: CreateWorkExperienceInput!)
}
CreateWorkExperienceInput {
  ...details
}
CreateUserInput {
  ...details
  workExperience: WorkExperienceHasMany
}
# WorkExperienceHasMany is the Nested Mutation type
WorkExperienceHasMany {
  create: [CreateWorkExperienceInput!] # this lets you create a whole array at once, using the same Input type as the createWorkExperience mutation
  # do the same thing for updates and deletes
  # update: [UpdateWorkExperienceInput!] 
  # delete: [ID!]
}
```

# Acceptance Criteria
- [x] Its possible to create, edit and delete experiences through the API
    - [x] through their own mutations
    - [x] or through nested mutations on the User
- [x] Its possible to update the ExperienceSkills attached to experience at the same time
- [x] Its possible to create and delete ExperienceSkills, and edit their text, but not to edit their skill_id or experience_id
- [x] Experiences (and ExperienceSkills) mutations should only be accessible by the users which own them, or admins

JamesHuf:
Resolves #1789 

Adds graphql mutations to create, update and delete experiences.

Notes:
- Changes the way experienceSkills are handled to make them a more proper pivot table.
- I added nested mutations on create and update users to allow creating experiences but not update or delete them because there doesn't seem to be a way to enforce the updated experience is owned by the user.
- ~~I haven't added authentication policies yet.~~ Experience policy now working thanks to @tristan-orourke 
- ~~I'm currently working on fixing all the errors.~~


JamesHuf:
Ok, this _should_ be ready for review now.

api/graphql/schema.graphql
```diff
@@ -396,74 +404,76 @@ enum EducationStatus {
...
+    skills: [Skill!] @morphToMany
+    skillPivot: ExperienceSkillPivot @rename(attribute: "pivot")
 }
 
-type ExperienceSkill {
-  id: ID!
-  details: String
-  skill: Skill! @belongsTo
-  experience: Experience! @morphTo
+type ExperienceSkillPivot {
```
petertgiles, line 475: Are we able to get rid of this type?  Is it necessary? If we're trying to align with how the other ones are done, the pivot tables are invisible to the schema, I think.  The experience interface has its own `details` property now, right?

JamesHuf:
> Do you have any queries or mutations written that you can share to jumpstart testing?

Query to get experiences (this needs to be queried through users since we don't have any queries for them directly right now)
```
query GetUsers {
  users {
    id
    firstName
    lastName
    experiences {
      id
      details
      skills {
        id
        name {
          en
        }
        experiencePivot {
          details
        }
      }
    }
  }
}
```
Mutation to create experience through nested operation on a user
```
// Graphql operation
mutation UpdateUser($id: ID!, $user: UpdateUserAsUserInput!) {
    // Query data
  }
}

// and passed in data:
"id": "81d0dd7f-43d4-4963-98ab-c9d06487dd5f",
"user": {
    "firstName": "James",
    "awardExperiences": {
      "create": [{
        // New experiences data
        "details": "details about experience",
        "skills": { "connect": [
          { "id": "53f8e98d-1755-4542-bac4-76c048317ce6", "details": "Experience skill details" }
        ]}
      }]
    }
  }
```
We have to pass in the userId if we create it directly
```
// Graphql operation
mutation CreateAwardExperience($userId: ID!, $awardExperience: AwardExperienceMutationInput!) {
  createAwardExperience(userId: $userId, awardExperience: $awardExperience) {
    // Query data
  }
}

// and passed in data:
"userId": "81d0dd7f-43d4-4963-98ab-c9d06487dd5f",
"awardExperience": {
    // other experience data
    "skills": { "sync": [
      { "id": "53f8e98d-1755-4542-bac4-76c048317ce6", "details": "Experience skill details" }
    ]}
}
```
Finally an update mutation (this can't be done through a user mutation)
```
// Graphql operation
mutation UpdateWorkExperience($id: ID!, $workExperience: WorkExperienceMutationInput!) {
  updateWorkExperience(id: $id, workExperience: $workExperience) {
    // Query data
  }
}

// and passed in data:
"id": "81d0dd7f-43d4-4963-98ab-c9d06487dd5f",
"workExperience": {
    // other experience data
    "skills": { "sync": [
      { "id": "53f8e98d-1755-4542-bac4-76c048317ce6", "details": "Experience skill details" },
      { "id": "53f8e98d-1755-4542-bac4-76c048317ce6", "details": "Other experience skill details" }
    ]}
  }
```

api/graphql/schema.graphql
JamesHuf: I'm pretty sure we still need this since we're just querying through this to get the details property of the relationship. The details property on experiences is for details on the experience itself.

petertgiles: Ah, OK, is it this?
image
I dislike the name - it's leaking implementation details.  I'm having trouble coming up with a better one, though. :laughing: ExperienceSkillUsage?  ExperienceSkillRelationsip?  ExperienceSkillAssignment?

JamesHuf: Yeah, it's that. I can see your complaint about the name but I'm also unsure about the best replacement.

petertgiles (changes requested):
Nice, I ran through the sample queries you sent and it looks like it all ran great!  :100: 

Just the pivot rename and I'm good to approve.

api/graphql/schema.graphql
petertgiles: Jerbo has recommended "[experience-skill-record](https://talent-cloud.slack.com/archives/CFCE2FSR4/p1649101910273059?thread_ts=1649094859.312789&cid=CFCE2FSR4)" and Tristan is on board.  So I suggest ExperienceSkillRecord for the schema type.  Hopefully that will be clearer when communicating with designers and users.  I noticed that the table name should be experience_skill (without the s) to comply with convention but I think that's out of scope.  I opened #2403 for that.

petertgiles (changes requested):
Don't forget the property names. :smile: 
image
They should probably match the type name, but in camelCase.  For example, experienceSkillRecord.

petertgiles (approved):
I love it.  :muscle: I'll leave it unmerged and hopefully Tristan will get a chance to review before sprint end.

api/storage/app/lighthouse-schema.graphql
```diff
@@ -385,6 +385,7 @@ interface Experience {
   applicant: Applicant!
   details: String
   skills: [Skill!]
+  experienceSkillRecord: ExperienceSkillRecord
```
petertgiles, line 388: lol, good catch!

tristan-orourke (changes requested):
Here's the review I half-finished last week. I haven't had a chance yet to look at the schema file, but there's a few small suggestions about other files - quick fixes

api/graphql/schema.graphql
```diff
@@ -360,7 +360,15 @@ type Skill {
...
     families: [SkillFamily!] @belongsToMany
-    experienceSkills: [ExperienceSkill] @hasMany
+
+    # Experiences
+    experiences: [Experience!]
+    awardExperiences: [AwardExperience!]
+    communityExperiences: [CommunityExperience!]
+    educationExperiences: [EducationExperience!]
+    personalExperiences: [PersonalExperience!]
+    workExperiences: [WorkExperience!]
```
tristan-orourke, line 370: These need `@hasMany` directives, like in User type

api/graphql/schema.graphql
```diff
@@ -873,6 +904,79 @@ input UpdateSkillInput {
     keywords: [String!]
 }
 
+input WorkExperienceMutationInput {
```
tristan-orourke, line 907: The word Mutation is redundant in these names, as graphql `inputs` are almost all used in mutations.

api/graphql/schema.graphql
```diff
@@ -360,7 +360,15 @@ type Skill {
...
-    experienceSkills: [ExperienceSkill] @hasMany
+
+    # Experiences
+    experiences: [Experience!]
+    awardExperiences: [AwardExperience!]
+    communityExperiences: [CommunityExperience!]
+    educationExperiences: [EducationExperience!]
+    personalExperiences: [PersonalExperience!]
+    workExperiences: [WorkExperience!]
+    experienceSkillRecord: ExperienceSkillRecord @rename(attribute: "pivot")
```
tristan-orourke, line 371: I don't think experienceSkillRecord applies here, though I could be wrong?

api/graphql/schema.graphql
```diff
@@ -398,74 +406,77 @@ enum EducationStatus {
...
 interface Experience {
-  id: ID!
-  applicant: Applicant! @belongsTo(relation: "user")
-  details: String
-  experienceSkills: [ExperienceSkill] @morphMany
+    id: ID!
+    applicant: Applicant! @belongsTo(relation: "user")
+    details: String
+    skills: [Skill!] @morphToMany
+    experienceSkillRecord: ExperienceSkillRecord @rename(attribute: "pivot")
```
tristan-orourke, line 413: I haven't tested these queries yet in the playground, but if I understand correctly the pivot record is only available in certain contexts. ie when Experience is nested inside Skill. 
eg `{skill { experiences { experienceSkillRecord { detail }}}}` would be valid but `{user {experiences {experienceSkillRecord {detail }}}}` would not.
I'm not sure what the best way around this is. If nothing else, can we leave a comment in the schema explaining this?
This applies equally to other instances of experienceSkillRecord fields.

tristan-orourke (changes requested):
Added a few more comments about the schema, after finally looking through it. Excellent work overall!

api/app/Models/EducationExperience.php
```diff
@@ -47,8 +47,10 @@ public function user(): BelongsTo
         return $this->belongsTo(User::class);
     }
 
-    public function experienceSkills(): MorphMany
+    public function skills(): MorphToMany
     {
-        return $this->morphMany(ExperienceSkill::class, 'experience');
+        return $this->morphToMany(Skill::class, 'experience', 'experience_skills')
+            ->withTimestamps()
+            ->withPivot('created_at', 'updated_at', 'details');
```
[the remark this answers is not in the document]
JamesHuf: Your right about the timestamps but `->as('experience_skill')` seems to just rename the attribute on the model to "experience_skill" (from "pivot"). The only useI can see for this here is renaming it to "experienceSkillRecord" so we don't need to use the `@rename(attribute: "pivot")` directive in the graphql.

api/app/Policies/ExperiencePolicy.php
```diff
@@ -0,0 +1,70 @@
...
+{
+    use HandlesAuthorization;
+
+    /**
+     * Determine whether the user can create models.
+     *
+     * @param  \App\Models\User  $user
+     * @return \Illuminate\Auth\Access\Response|bool
+     */
+    public function create(User $user, $second)
```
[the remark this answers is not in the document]
JamesHuf: This is already changed to $request after @petertgiles found it.

api/graphql/schema.graphql
JamesHuf: I think it does since it allows us to be able to see the relationship details when querying for skills through experiences. Like this: 
image

api/graphql/schema.graphql
JamesHuf: Yeah, that's how it works (if you try to query for it directly detail just returns null) but I'm also unsure of any good way around this. I'll add a comment explaining this now though.

tristan-orourke (approved):
Looks great! Thanks for adding the comments, I just suggested some minor changes in wording.

Also, I want to suggest a change to the `->as()` name for the pivot attribute. ExperienceSkillRecord is a great name for the type in the API, but I think referencing the fact its a pivot table is helpful when working with the Laravel models. (Helpful in terms of reading the code and easily understanding how it all fits together.) What do you think about changing `->as('experienceSkillRecord')` to `->as('experience_skill_pivot')`. (I also like using python_casing here, to match other Eloquent attributes.) It will require re-adding the `@rename` directive in the schema, but I think it might be worth it. 

If you disagree and think tying the attribute and name to the schema type name is more legible, I'm open to leaving as it is.

Either way, these changes are super minor. I'm going to approve so that I'm not blocking the merge after these last few suggestions are considered!

Nice job!

PR #2359, opened 2022-03-29
Feature/1044 pagination admin table

### LINKED ISSUE #1044: Admin - tables - horizontal scroll bar hidden
The horizontal scroll bar for tables like users and pool candidates is only visible at the bottom of the table.  If there are more than about 10 rows then scroll bar is not visible.  For example, on the pool candidates table page it isn't clear that there is an edit column in the table.  If a user wants to edit the first row then they need to scroll to the bottom of the page, scroll to the right, then scroll back to the top of the page, then click edit.  Is there a better way to handle scrolling and visibility on large tables?

Screenshot 2021-11-04 112458

yonikid15:
Resolves #1044 

# Notes
- Adds pagination to admin table using React Table [usePagination hook](https://github.com/TanStack/react-table/tree/v7/examples/pagination). Now that the number of rows can be limited, I believe it should indirectly fix the horizontal scroll issue since the user doesn't need to scroll that much.


vd1992:
- Checked out the branch, built, no errors or anything seeming to have broken
- Pagination is present everywhere
- Visually looks fine, I suppose you could add a little space between page jumper and "Show X" selector

I think I found a bug however: 
- Navigate to Departments
- Click up arrow next to the page jump thing, the one to skip through pages
- Number stuck at 7, cannot go above it to 8

Good work Yoni! Double check if that bug exists on your end, or if it was just something being funny on my end.

vd1992 (changes requested):
-  Checked out the branch, built, no errors or anything seeming to have broken
-  Pagination is present everywhere
-  Visually looks fine, I suppose you could add a little space between page jumper and "Show X" selector

I think I found a bug however:

-  Navigate to Departments
-  Click up arrow next to the page jump thing, the one to skip through pages
-  Number stuck at 7, cannot go above it to 8

Good work Yoni! Double check if that bug exists on your end, or if it was just something being funny on my end.

frontend/common/src/components/Pagination/Pagination.tsx
```diff
@@ -37,7 +54,7 @@ const Pagination: React.FunctionComponent<PaginationProps> = ({
   });
 
   // If there are less than 2 times in pagination range we shall not render the component
-  if (currentPage === 0 || paginationRange.length < 2) {
+  if (currentPage === 0 || totalCount < 2) {
     return null;
   }
```
vd1992, line 59: Does this work at the moment? Pagination thing appears with less than two items, or have I misunderstood what this is supposed to do.

image

auth/api_secret.txt
```diff
@@ -0,0 +1,5 @@
+New client created successfully.
```
petertgiles, line 1: Oops, this shouldn't be checked in.

frontend/admin/resources/js/components/Table.tsx
```diff
@@ -1,8 +1,21 @@
...
+  useTable,
+  useGlobalFilter,
+  useSortBy,
+  Column,
+  usePagination,
+} from "react-table";
 import { Button } from "@common/components";
+import Pagination from "@common/components/Pagination";
+import {
+  ChevronDoubleLeftIcon,
```
petertgiles, line 14: Unused?

auth/api_secret.txt
petertgiles, line 1: Oops, this shouldn't be checked in

petertgiles (changes requested):
Nice, swapping out for the homegrown one!  I confirm that I see the two bugs @vd1992 noticed.  Also, could you do a linting pass on Table.tsx?

frontend/common/src/components/Pagination/Pagination.tsx
yonikid15: Hmm, so I was trying to fix the problem of the pagination component suddenly disappearing when a user changes the number of how many items to show (the select box at the end). I'll ask what the designers want to do about that..

yonikid15:
Alright, should be good to go.. I talked to the design team, and they said to disable buttons except the select box which changes the total number of items on the page. Please do your best to break it though when testing 😆

vd1992:
> Alright, should be good to go.. I talked to the design team, and they said to disable buttons except the select box which changes the total number of items on the page. Please do your best to break it though when testing 😆

💥

vd1992 (approved):
My requests all seem to be fixed, noticed no other issues or glitches when playing around with it. 
All that is left is Peter's linting request and I think it is all good?

petertgiles (approved):

Issue #2329, opened 2022-03-25
Missing field from Type User - gov employee type


vd1992:
Per the wireframe, there is the field there corresponding to gov employee type, student vs contract etc. 
It isn't present in Type User
https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/api/graphql/schema.graphql#L55

![image](https://user-images.githubusercontent.com/40485260/160154425-9f3ee110-52e8-45d0-8e33-f817f380b470.png)

# Acceptance Criteria
- [x] Add missing types to database (with migrations)
- [x] Add missing types to graphql schema
- [x] Add missing types to GovInfoFormOperations.graphql
- [x] Fix TODOs in GovernmentInfoForm.tsx
- [x] Fix TODOs in /profile/ProfilePage/ProfilePage.tsx

PR #2369, opened 2022-03-31
Fill in the profile form with values

### LINKED ISSUE #1996: Populate sections of main profile page
# Description
Fill in all the sections which display user data. Each one needs to check if it is missing required data, and if so show a text box and a link to the edit form.

This applies to Pools, About me, language, Gov info, Work Location and Work Preferences

[Wireframe](https://www.figma.com/proto/PrvXEDac0HVp7dF7SLWVFm/Candidate-profile-interfaces---V5.0.0?node-id=2340%3A4185&starting-point-node-id=2340%3A4185)

# Acceptance Criteria
- [ ] Includes Graphql operation to get user data.
- [ ] Fill in all the sections which display user data.
- [ ] Each one needs to check if it is missing required data, and if so show a text box and a link to the edit form.

vd1992:
resolves #1996 

Some issues, per wireframe, there isn't anything in the schema yet for "job duration" or "disliked operational requirements"
Intl-ing and localizing ignored until the form is appropriately filled out
GraphQL stuff works fine I think locally


vd1992:
From local environment: 

image

vd1992:
All needed fields pulled in now I think from Schema.
Functionality all present.

vd1992:
You can use the Controls tool for Chromatic Storybook to pass in specific values if you have something in mind to test

petertgiles (changes requested):
Whoah, this is looking cool!  :tada: A few suggestions:

There are some issues with the page when no information is there (for example, when not logged in):
- Many boxes are missing the "There are required fields missing" message.
- Field labels with no entered data should not be displayed.
It would be good to have a storybook story in this state to be able to review.

The naming of the "inner component" and "outer component" is a little out of convention, I think.  I suggest _ProfilePage_ for the outer component and _ProfileForm_ for the inner component to be more like _SearchPage_ and _SearchForm_.  I think keeping them in the same file like this with this name is fine though.

Do we need the giant _ProfilePageProps_ interface?  This is the _User_ type, isn't it?

Regarding _skills and experience_, James' PR should be in soon.  I think it's worth holding this one back until that's in so that you can complete that section.

petertgiles (changes requested):
Please check all the enums, as well.  There are a few that are displaying as the raw enum instead of the localized description.

vd1992:
> Whoah, this is looking cool! 🎉 A few suggestions:
> 
> There are some issues with the page when no information is there (for example, when not logged in):
> 
>     * Many boxes are missing the "There are required fields missing" message.
> 
>     * Field labels with no entered data should not be displayed.
>       It would be good to have a storybook story in this state to be able to review.
> 
> 
> The naming of the "inner component" and "outer component" is a little out of convention, I think. I suggest _ProfilePage_ for the outer component and _ProfileForm_ for the inner component to be more like _SearchPage_ and _SearchForm_. I think keeping them in the same file like this with this name is fine though.
> 
> Do we need the giant _ProfilePageProps_ interface? This is the _User_ type, isn't it?
> 
> Regarding _skills and experience_, James' PR should be in soon. I think it's worth holding this one back until that's in so that you can complete that section.

- Skill and Experiences is out of scope of the issue though
- Required boxes won't apply to Inclusion, and Government Info is missing that one field so that one will be pending. The other ones seem to have it appear. Not sure if they fail to appear in some cases though. 
- I don't know if I need the interface, so far I just try to avoid the red underlines through whatever means. 

Renamed components, empty fields hidden now, and added in Enum stuff. Standard Intl I'm leaving until I've made sure the page is complete with all the stuff I need to add.

vd1992:
types tamed, type monstrosity eliminated

frontend/talentsearch/src/js/components/profile/ProfilePage/ProfilePage.tsx
```diff
@@ -160,7 +269,25 @@ export const ProfilePage: React.FC<ProfilePageProps> = ({
                   description: "Title of the My hiring pools section",
                 })}
               </h2>
-              <p>Pool details</p>
+              <div
+                data-h2-bg-color="b(gray)"
+                data-h2-padding="b(all, m)"
+                data-h2-radius="b(s)"
+              >
+                {(candidateArray === null || candidateArray.length === 0) && (
```
petertgiles, line 277: You're doing a lot of `=== null` and `=== undefined` and `=== 0` in this file. These are all considered falsy in javascript.  You could using the more concise form of direct boolean evalution like `!candidateArray || !candidateArray.length` to shorten the code.  I see you doing this in other places on this file so I suggest that form to be consisent.

frontend/talentsearch/src/js/components/profile/ProfilePage/ProfilePage.tsx
```diff
@@ -249,7 +648,64 @@ export const ProfilePage: React.FC<ProfilePageProps> = ({
...
               </div>
-              <p>Government status details</p>
+              <div
+                data-h2-bg-color="b(gray)"
+                data-h2-padding="b(all, m)"
+                data-h2-radius="b(s)"
+              >
+                {/* !!!!!!!!!!!!
+                TODO
+                ADD GOVERNMENT EMPLOYEE TYPE TO THIS FORM FUNCTION, THEN ADD EMPLOYEE TYPE LOGIC HERE
```
petertgiles, line 658: This is the missing field you mentioned?  Could you add the issue # here?

frontend/talentsearch/src/js/components/profile/ProfilePage/ProfilePage.tsx
```diff
@@ -340,7 +920,74 @@ export const ProfilePage: React.FC<ProfilePageProps> = ({
...
+                        defaultMessage: "I identify as:",
+                        description:
+                          "Label preceding what groups the user identifies as part of, followed by a colon",
+                      })}{" "}
+                    </p>{" "}
+                    <ul data-h2-padding="b(left, l)">
+                      {isWoman && (
+                        <li data-h2-font-weight="b(700)">
+                          {intl.formatMessage({
+                            defaultMessage: "Woman",
```
petertgiles, line 957: I feel like we use these EE groups everywhere even though they're not a graphql enum.  I suggest adding them to the localized constants file.

petertgiles (changes requested):
Looking great!  I noticed a few more things: 

Checking null state:
- For _Diversity, equity and inclusion_ there should be a link for _employment equity groups_.
- Many of the boxes should have a _You haven't added any information here yet_ message.

For non null state:
- None of the stories have data for pool candidates.  Bug in component or query or seeder?
- For _Government information_ the items should be in an unordered list (dots)
- For the _About me_ section could you make the two columns collapse for small screens?
image

frontend/talentsearch/src/js/components/profile/ProfilePage/ProfilePage.tsx
vd1992: going to mention this component in #2329

vd1992:
will edit comment this later

petertgiles (approved):
Nice work, Vachan.  It looks great!

vd1992:
> Annoyingly pedantic comment: Should the component be called ProfileForm? It doesn't contain a form! ;)

Peter suggested it above, so I rolled with it!
"The naming of the "inner component" and "outer component" is a little out of convention, I think. I suggest ProfilePage for the outer component and ProfileForm for the inner component to be more like SearchPage and SearchForm. I think keeping them in the same file like this with this name is fine though."

PR #2373, opened 2022-04-01
Fix failing tests on admin

yonikid15:
Resolves #2373


vd1992 (approved):
Jest Test doesn't fail
Skill Accordions weren't broken in Storybook

PR #2414, opened 2022-04-06
Add package integrity check

### LINKED ISSUE #2153: Ensure integrity checksums exist in lockfile for all npm packages
npm packages have a content-specific "integrity" hash for all their contained files. This comes from the npmjs repository, and so if any files are different from what was in the package, then npm can notice. These integrity hashes are stored in package-lock.json.
https://docs.npmjs.com/cli/v6/configuring-npm/package-lock-json#integrity

Sometimes these are missing, and we should know when that happens.

This may involve:
- a GitHub Action or npm package that checks for missing integrity hashes in `package-lock.json` during CI, or
- some flag in npm that enforces strictness (??), or
- all of the above, for added confidence

Related issue: https://github.com/npm/cli/issues/4460

Related: No pressure to pollute our labels, but do you think it's worth having a `security` label to make security tasks more visible, @tristan-orourke?

## To Do
- [x] ~~research npm packages that validate integrity~~
- [x] ~~research npm flags that force this~~
- [x] optional: add `security` label
- [x] maybe: create/modify github action to do this
- [ ] confirm integrity hashes are present https://github.com/GCTC-NTGC/gc-digital-talent/pull/2414
- [ ] confirm integrity hashes are well-formed https://github.com/GCTC-NTGC/gc-digital-talent/pull/2414
- [ ] validate hashes match local content. https://github.com/GCTC-NTGC/gc-digital-talent/issues/2557
    - confirm integrity hashes match cached tarball. see: https://gist.github.com/patcon/88e7e8332e49abd06b79feb1d0a134bd
- [ ] validate hashes match npmjs hashes. https://github.com/GCTC-NTGC/gc-digital-talent/issues/2558

petertgiles:
Add a script and workflow to verify that all node packages in the lock file have an integrity check value.
Closes #2153


tristan-orourke:
Cool! Makes sense to me. I like that we can trigger the check locally with an npm command too!

I'll leave this for Pat to make the final review.

patcon:
This is great! Ok, so I ran a quick check, and it seems like there's something we can still cover here, which is a failure of npm, I think.

**Issue:** `npm ci` which is supposed to be [mostly?] deterministic and secure (e.g., suppose to fail if sub-resource integrity values don't match, and not supposed to update lockfile with `npm ci`), but it's trooping along through install, even when SRI (sub-resource integrity) hashes don't match, and even if they're totally corrupt.

This is kinda messed up, but maybe a bug we can support a fix on. For now, maybe we work-around for ourselves, but try to push a more universal fix upstream?

### Steps to Reproduce

1. `cd frontend && nvm use` (we're now using npm v8.3.0)
2. `npm install` (ensure lockfile is proper)
3. edit lockfile to remove a random char from an integrity hash
4. `npm run --workspace common check-integrity` properly throws error about corrupt integrity hash
5. run `npm ci`

#### Expected Result

Expect `npm ci` to not change the `package-lock.json`, and to fail.

#### Observed Result

`npm ci` doesn't touch the `package-lock.json`, but it _successfully_ installs packages.

### Alt: Steps to Reproduce

Same as above, but for (3), copy another integrity check to another package entry.

BUT (4) **won't throw an error**, because technically the checksum is proper form, it just doesn't match.

#### Expected Result

Don't touch package-lock.json, and fail out because sub-resource integrity hash doesn't match.

#### Observer Result

Doesn't touch package-lock.json, but succeeds (ignore incorrect checksum)

patcon:
Ok, so for the alt reproduction, we need something else to protect us until an npm fix arrives.

@petertgiles your rad fix resolves situation when something is missing, or hash is present but malformed, but we're missing the case where a well-formed hash is present, but **doesn't match the files on disk**.

We should be able to use a library that checks SRI (sub-resource integrity) which I believe is the way npm generates those integrity hashes. So we can prob use a library like this (which might replace our current code too, since it should we checking for well-formed hashes as well):
- ⭐ `_32` https://github.com/npm/ssri
- ⭐ `__8` https://github.com/JimiC/nsri
- ⭐ `_24` https://github.com/mnismt/checksum
- ⭐ `_11` https://github.com/zkat/srisum
- ⭐ `_29` https://github.com/neftaly/npm-sri-toolbox
- ⭐ `_11` https://github.com/kopach/lockfix
- ⭐ `329` https://github.com/waysact/webpack-subresource-integrity
- ⭐ `__6` https://github.com/mikechau/sri-stats-webpack-plugin

Some of the above are SRI for included JS and CSS files, which is outside the scope of this issue. But might be nice to choose a library that maybe helps us with that later. So including them above in the research

patcon:
K, I'm piecing together what's needed here, for this extended part (which maybe we don't want to do here in this PR, but whatever, we can sort that later):

- why `npm ci`? [NPM CI: A Command You Should Know](https://lusbuab.medium.com/npm-ci-a-command-you-should-know-d8d67847884c)
- StackOverflow: [Validating NPM Package Integrity](https://superuser.com/questions/1564144/validating-npm-package-integrity)
  - W3C docs: [Subresource Integrity: The integrity attribute](https://w3c.github.io/webappsec-subresource-integrity/#the-integrity-attribute)
  - NPM Docs: [Verifying the PGP signature of a package from the npm public registry](https://docs.npmjs.com/verifying-the-pgp-signature-for-a-package-from-the-npm-public-registry)
  - StackOverflow: [How to manually confirm integrity field in package-lock.json file?](https://stackoverflow.com/a/70349349)
    - From here maybe: Mozilla MDN: [Tools for generating SRI hashes](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity#tools_for_generating_sri_hashes)

patcon:
OK, so it seems that npm's own ssri package has a bug bounty program and is maintained by GitHub themselves (and prob used in npm itself, ideally), so I wrote a quick script to cover a little bit more surface area:

https://gist.github.com/patcon/88e7e8332e49abd06b79feb1d0a134bd

This still assumes the cache has proper values, and just checks that cached checksums match the caches tarball.

In theory, there are still two shakey places:
- the cached tarball and cached checksum could match, but it might not be the contents from npm. we don't confirm that yet
- the cached tarball and cached checksum could match, but maybe our uncompressed files running in node_modules don't match that. we don't confirm that yet

I would have thought this was npm's job to watch, but am a little uncertain after interactions in the other issue. So I do feel like this is a worthwhile exercise to carry out for a government-run project that is open source and therefore very easy emulated by future gov projects.

Should these little sub-goals be split out, or keep them here? Also, would be great to push the essence of these fixes upstream into npm itself (if we realize it's truly missing during `npm ci` or `npm install`, and without good reason -- maybe there are edges which explain why this isn't present)

patcon:
I think we could also add little tests to this:
- remove an integrity hash => expect our CI to fail
- corrupt an integrity hash => expect failure
- change a file in node_modules => expect failure
- change file in node_modules BUT update "fake" hash to match => expect failure (won't match npmjs hash)

imho this should all be handled by npmjs somehow, but we can cover it ourselves to set a good example 🤷

patcon (changes requested):
As specified in other comments

tristan-orourke:
Thanks for the very thorough deep dive @patcon! 

I like your idea of writing some tests for this - would also help document what we're doing here. 

I think the remaining blind spots (validate cached integrity hash matches hash on npmjs; validate cached integrity hash matches hash on npmjs) could be pulled into a new ticket.

petertgiles:
I agree with Tristan that this is quite a different scope of project and should probably be in a new ticket (with a bigger estimate!).  I'll definitely add some tests to prove the current functionality though.

I'm not sure we should be writing code recreate what npm _should_ be doing though.  It makes sense for us to enforce packages having integrity hashes since we know npm does not want to enforce that.  However, if npm should be noticing corrupted hashes and isn't then I feel it makes more sense to work with them to do that rather than recreating the functionality ourselves.

frontend/common/src/tooling/checkIntegrityLib.js
```diff
@@ -0,0 +1,72 @@
...
+          `Package ${packageName} has the malformed integrity value ${packageProperties.integrity}`,
+        );
+        invalidIntegrityCount += 1;
+      }
+      return previousCount + invalidIntegrityCount;
+    },
+    0,
+  );
+
+module.exports = {
```
petertgiles, line 68: I couldn't figure out how to export the functions for testing without having the script run.  I factored them out to a separate lib file.  Correct approach?

petertgiles:
Honestly, I'm still not sure why the resolve and integrity are dropping out for @esizer.  I've updated the script to only check the `packages` section of the file since it seems like the `dependencies` section is only around for compatibility and is not used.

For my last commit I rebuilt the package lock file by running
```
rm -rf node_modules
rm package-lock.json 
npm cache clean --force
npm i
```
After running that there were a handful of packages that still were missing the integrity so I manually ran `npm i package-name --workspace=package-workspace` to fill them in.

patcon:
> After running that there were a handful of packages that still were missing the integrity so I manually ran npm i package-name --workspace=package-workspace to fill them in.

I'm assuming this is ok to keep doing in future, so long as we know it's happening?

petertgiles:
Nice, thanks for creating those new issues!

petertgiles:
> I'm assuming this is ok to keep doing in future, so long as we know it's happening?

Honestly, I'm still not even sure _why_ it's happening let alone whether this is best solution.  At least we'll know when it does, now. :shrug:

PR #2415, opened 2022-04-06
Add link to "holding tank" for proposed/deferred/discarded DoD/DoR items

patcon:
Small doc change to have a single shared place (current location just temporary; i'm sure there's a better place) to store candidates outside the "authoritative" files that have a bit of a barrier (as they should have).

cc: @tristan-orourke @gggrant 

Not urgent, just wanted to capture it in complete PR without talking about it, since it only took 5 min to do. Happy to see it merged, changed, or discarded after retro.

Can anyone suggest the clearest way to queue this up for consideration during backlog refinement?

tags: definition of done / ready


gobyrne:
Let's let this come up through retro if it's needed.

PR #2140, opened 2022-02-24
Skills in detail

yonikid15:
Resolves #1795 

# Notes
- I added the simple word counter we made on Talent Cloud and tweaked it a bit to fit this app.
- Also, I took the liberty of moving some files around in `/skills` to fit the new naming convention we decided to use.

PR #2424, opened 2022-04-08
Talent Search - Pluralize Candidate Result Count #2370

### LINKED ISSUE #2370: Talent Search - Pluralize Candidate Result Count
The candidate search should properly pluralize the word _candidates_ for a result count of 1 in the search results:

RVany:
The candidate search should properly pluralize the word _candidates_ for a result count of 1 in the search results:

Resolves #2370

### DIFF
--- a/frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx
+++ b/frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx
@@ -27,12 +27,6 @@ const EstimatedCandidates: React.FunctionComponent<
     }
   }, [updatePending, candidateCount]);
 
-  // React.useEffect(() => {
-  //   if (candidateCount === 1) {
-  //     setPluralizedWord("candidate");
-  //   }
-  // }, [candidateCount]);
-
   return (
     <div
       data-h2-bg-color="b(white)"
--- a/frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx
+++ b/frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx
@@ -47,7 +47,7 @@ const EstimatedCandidates: React.FunctionComponent<
           {intl.formatMessage(
             {
               defaultMessage:
-                "There are approximately <weight>{candidateCount}</weight> {candidateCount,plural, =0 {candidate} =1 {candidate} other {candidates}} right now who meet your criteria.",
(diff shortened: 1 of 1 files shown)


petertgiles:
There's actually a simpler way to do this that doesn't require useState or useEffect hooks.  Check out this commit from Yoni that does it using only the intl library: https://github.com/GCTC-NTGC/gc-digital-talent/pull/2140/commits/45a6e549e7019ab728ec849220128feac69f6a8a

petertgiles:
Here's to documentation on it.  The section on pluralizing is near the top: https://formatjs.io/docs/intl-messageformat/

RVany:
> 

Ok! I will do that!

frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx
```diff
@@ -47,7 +47,7 @@ const EstimatedCandidates: React.FunctionComponent<
           {intl.formatMessage(
             {
               defaultMessage:
-                "There are approximately <weight>{candidateCount}</weight> candidates right now who meet your criteria.",
+                "There are approximately <weight>{candidateCount}</weight> {candidateCount,plural, =0 {candidate} =1 {candidate} other {candidates}} right now who meet your criteria.",
```
petertgiles, line 50:
```suggestion
                "There are approximately <weight>{candidateCount}</weight> {candidateCount,plural, =0 {candidate} =1 {candidate} other {candidates}} right now who meet your criteria.",
```
```suggestion
                "There are approximately <weight>{candidateCount}</weight> {candidateCount,plural, =0 {candidates} =1 {candidate} other {candidates}} right now who meet your criteria.",
```

petertgiles (changes requested):
This is the right approach, great job!  The word for 0 is not right though.

Now, I just noticed something else that I did not see before so it was not in the issue.  But for one candidate there are two other words that should change as well. "There _is_ approximately 1 candidate right now who _meets_ your criteria.  Could you explore if it is possible to solve that in the same way?  On [this page](https://unicode-org.github.io/icu/userguide/format_parse/messages/#complex-argument-types) they suggest writing the full sentence as the submessage which would allow you to substitute several words in the sentence at the same time.
image

RVany:
Ha Ha! I didn't think about There are/There is part. Yes this is way better.

petertgiles (approved):
Nice, we got it!  :handshake:

Issue #2425, opened 2022-04-08
<legend> must be a direct descendant of <fieldset>


esizer:
## Summary

Currently we have the `<legend>` element nested within a `<div>` within our `<fieldset>`. Unfortunately, it is required that the `<legend>` be a direct descendant of the `<fieldset>`. Of course, this introduces an issue since we need to have the required and optional text be displayed next to the `<legend>`. Having the legend be a direct descendant ensures that the `<legend>` is properly associated with the `<fieldset>` and announced by assistive technology.

Ref: [H71: Providing a description for groups of form controls using fieldset and legend elements](https://www.w3.org/TR/WCAG20-TECHS/H71.html)
Ref: [Legend must be a direct child of fieldset](https://www.accessibility-developer-guide.com/examples/forms/grouping-with-fieldset-legend/#:~:text=Legend%20must%20be%20a%20direct%20child%20of%20fieldset,-%23&text=HTML%20requires%20the%20element,first%20child%20of%20.&text=As%20a%20result%2C%20some%20screen,is%20not%20its%20first%20child.)

## Proposed Solution

This is not ideal but would allow us to have the required layout while maintaining the `<legend>` as a direct descendant.

```jsx
<fieldset>
  <legend class="sr-only">Fieldset Legend</legend>
  <div>
    <span aria-hidden="true" role="presentation">Display Legend</span>
    <span>Optional/Required</span>
  </div>
</fieldset>
```

This would use an `sr-only` to only represent the text for screen readers, allowing the legend to be a direct descendant and hiding the display legend using `aria-hidden` and `role="presentation"` so we can have the layout required.

PR #2429, opened 2022-04-08
Move legend to be direct descendant of fieldset

esizer:
## Summary

This is a simple change to move the `<legend>` of our `<fieldset>` to be a direct child so they are properly associated for assistive technologies. There _should_ be **no visual changes**, just moving the markup around.

Resolves: #2425

### DIFF
--- a/frontend/common/src/components/inputPartials/Fieldset/Fieldset.tsx
+++ b/frontend/common/src/components/inputPartials/Fieldset/Fieldset.tsx
@@ -44,9 +44,7 @@ export const Fieldset: React.FC<FieldsetProps> = ({
       }}
       data-h2-margin="b(bottom, xxs)"
     >
-      <legend data-h2-visibility="b(invisible)" data-h2-font-size="b(caption)">
-        {legend}
-      </legend>
+      <legend data-h2-visibility="b(invisible)">{legend}</legend>
       <div
         data-h2-display="b(flex)"
         data-h2-flex-wrap="b(wrap)"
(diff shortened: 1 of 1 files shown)


petertgiles:
> There _should_ be **no visual changes**, just moving the markup around.

@gobyrne ... if only we had some software package that would highlight all the visual differences that occurred as a result of a change....

petertgiles:
> some software package that would highlight all the visual differences

@esizer we have full permission to nag Gray until he buys our Chromatic license. :grin: This would be a very handy time to have that working since it could tell us if there were any visual differences caused by this change.

esizer:
> > some software package that would highlight all the visual differences
> 
> @esizer we have full permission to nag Gray until he buys our Chromatic license. 😁 This would be a very handy time to have that working since it could tell us if there were any visual differences caused by this change.

@petertgiles that would be excellent for issues exactly like this where we change markup for a11y and want design integrity to stay intact.

frontend/common/src/components/inputPartials/Fieldset/Fieldset.tsx
```diff
@@ -44,13 +44,22 @@ export const Fieldset: React.FC<FieldsetProps> = ({
       }}
       data-h2-margin="b(bottom, xxs)"
     >
+      <legend data-h2-visibility="b(invisible)" data-h2-font-size="b(caption)">
```
petertgiles, line 47: I'm definitely not the expert here but is the font size attribute here doing anything?  It seems confusing the make something invisible and then set a font size but maybe it has some significance that I'm not thinking of?

esizer: No, it doesn't do anything useful outside of inflating the DOM. I was more worried about hiding it and forgot to remove the, now unnecessary, font size rule.

Resolved with: [remove unnecessary font-size rule](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2429/commits/0e41f58655512200705c065bdcf42a30e5493525)

petertgiles (approved):
Nice one!  Great accessibility catch!

Issue #1641, opened 2022-01-04
Node - permission denied


petertgiles:
It seems that node does not like to be run as root.  I think this is one of the issues @patcon is running into in PR #1449 as well.  It seems to not affect docker on a regular windows file system but does appear on a WSL or Apple file system.  I think our maintenance container should be updated to not run regular commands as root.

![image.png](https://images.zenhubusercontent.com/6172cf6e898e11fd20507433/955063f5-67dd-463e-af96-3b6776751d6e)

patcon:
to be clear, it seems to work locally on "docker for mac", but just gets complicated with edge-cases when running docker in CI environments. it's weird. something to do with docker shared volumes and mount types (but only with npm <7.x).

Any one of these would address the issue:
- pinning npm < 7.x resolves things without abandoning nvm
- using a docker image with specific node version (instead of nvm) resolves things
- somehow winding back use of shared volumes (seemingly a non-starter with our current dev process, though possible with some of the vscode flows mentioned here: https://code.visualstudio.com/docs/remote/containers)
- never running local scripts in CI resolves (this would seem increasingly limiting)

petertgiles:
@patcon I think I like option #2.  Since it's a controlled environment I think it makes sense to bake in the version of node and npm on container creation.  Trying to run nvm without an interactive shell is real hassle.  It also really speeds up script runs since it doesn't have to test node and npm on every run.

I also really like the idea of VS code dev containers and have been playing around with them for a bit.  I'll work on a PR for that after this issue is closed.

patcon:
Mind if I take a run at this? I think I've mostly gone through the steps while troubleshooting the CI environment :)

petertgiles:
Haha, I was working on this on Friday and got a big chunk done, too.  You can finish it off if you want though.  I just pushed my branch called feature/optimize-maint-container so you can either use that and add to it or just review it for ideas.  I was going to remove devcontainer.json file since I'm not ready to included that in a PR yet.  And I wanted to add an environment variable to the container for workspace root and use that instead of all the relative paths in the scripts.

patcon:
heh don't lemme step on your toes. i'll stick with review :) thanks peter!

petertgiles:
Alright, no problem, I'll finish it off.  Not gonna lie, I'll probably crib a few things from your draft PR before submitting it. 😉  I like how you've made the failures quicker and have the switching between dev and prod.

petertgiles:
[...]

PR #2430, opened 2022-04-08
Fix admin docker env

### LINKED ISSUE #2327: Docker Compose env file for php container overwrites local .env files
The docker compose file adds frontent/admin/.env to the php container on build.  Unfortunately, these values override the values from the local .env file when running Laravel apps.  Depending on what was in the .env file on container build this can cause a variety of issues.  Some that I've seen so far are both related to login: `No app key` and `Invalidate grant (bad/missing) redirect_uri)`.  I wouldn't be surprised if #2309 is related to this too.

### LINKED ISSUE #2328: Clean Setup - composer fails to build
When doing a completely clean setup the composer up command will fail due to /frontend/admin/.env not existing yet.

petertgiles:
This branch factors out the environment variables used by Apache to a separate env file.  

This resolves #2327 by ensuring that an incomplete .env file is not used for php container construction resulting in some environment variables being unset in the container environment.

This resolves #2328 by ensuring that the env file needed for the php container creation is present in a fresh install environment instead of relying on a previous setup script run.

### DIFF
--- a/frontend/admin/.apache_env
+++ b/frontend/admin/.apache_env
@@ -1,0 +1,5 @@
+# TODO: some sort of logout landing page
+OAUTH_POST_LOGOUT_REDIRECT="http://localhost:8000/admin"
+
+# Don't use with local Laravel auth
+# OAUTH_LOGOUT_URI="https://te-auth.id.tbs-sct.gc.ca/oxauth/restv1/end_session"
--- a/frontend/admin/.env.example
+++ b/frontend/admin/.env.example
@@ -25,12 +25,6 @@ QUEUE_CONNECTION=sync
 
 API_URI="/graphql"
 
-# TODO: some sort of logout landing page
-OAUTH_POST_LOGOUT_REDIRECT="http://localhost:8000/admin"
-
-# Don't use with local Laravel auth
-# OAUTH_LOGOUT_URI="https://te-auth.id.tbs-sct.gc.ca/oxauth/restv1/end_session"
-
 # continuous integration only
 MERGE_STORYBOOKS=false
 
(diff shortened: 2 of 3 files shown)


esizer:
@petertgiles I was going to pull this down onto a fresh machine (Windows 10 WSL2 Ubuntu 20.04 LTS). Though, I'm running into a permissions error when the `setup.sh` script gets to the `npm run dev` command. (works on my macbook with no issue)

I looked into the docker container and saw that all the commands are running as `root`, though the files are owned by `1000:1000`. 

If I run `chown -R root:root .` it resolves the issue but I can see other issues arising from this.

Could be out of scope for this PR. I also know you and @patcon have been working on refactoring docker as well and not sure if this would be impacted or not. This may also not be the place for this type of discussion and if so, we can move elsewhere.

petertgiles:
@esizer oh man, you're finding all the skeletons in the closet!  Yes, that's on the board as https://github.com/GCTC-NTGC/gc-digital-talent/issues/1641 but I agree it's out of scope for this one.  Most Windows folks on the team are not using WSL2 but I think it's great.  If you're able to figure that one out, though, I'll be forever indebted!

esizer:
> @esizer oh man, you're finding all the skeletons in the closet! Yes, that's on the board as #1641 but I agree it's out of scope for this one. Most Windows folks on the team are not using WSL2 but I think it's great. If you're able to figure that one out, though, I'll be forever indebted!

Okay thanks. At least I can take comfort in the fact that it is not just me. I did do some troubleshooting between my first day and trying to get the project running and today but so far nothing. I could circle back when #1641 gets promoted from the product backlog.

esizer (approved):
Finally got around to doing a review on a machine not running WSL and it works great!

From my initial setup last Monday which was slightly painful, this is now running perfectly.  I was able to get it up an running simply by following the `maintenanc/README` file with no issue :)

PR #2431, opened 2022-04-08
2423 buttons wrapped by links

### LINKED ISSUE #2423: Instances of Button component wrapped with Link
## Issue

There are some instances where we are wrapping the `<Button>` component with `<Link>`. I presume this is to get the styling of a button and the functionality of a link. Unfortunately this results in the following markup being generated:

```jsx
<a>
  <button>
     Link
  </button>
</a>
```

This is [not considered valid HTML](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) and can cause [issues with accessibility](https://accessibilityinsights.io/info-examples/web/nested-interactive/) and cause assistive technology to act in unexpected ways.

## Example

Currently the only example I have come across is the [CancelButton.tsx](https://github.com/GCTC-NTGC/gc-digital-talent/blob/0514af0b14268a51c1a80ba87a44fdd2dcd25c75/frontend/talentsearch/src/js/components/applicantProfile/CancelButton.tsx#L11-L33) component. 

### Easy Fix

An easy fix would be to add a "link" option to the type prop which would render a Link rather than a button and also provide an optional href prop that is used only when the type is set to link.

### Comprehensive

I'm not sure how necessary this would be, but we could instead us the "as" prop pattern to give the developer control over the rendered component while keeping the button styling. Though, this may add unnecessary complexity in annotating prop types.

#### Example

```jsx
import React from 'react';

const Button = ({ as = "button" }) => {
 const Component = as;
 return <Component>{children}</Component>
}

Button.defaultProps = {
 as: 'h1',
};

export default Button;
```

esizer:
## Summary

Introduced a new API for the `<Link>` component to apply the same styles found with the `<Button>` component to avoid having to nest interactive elements and improve accessibility.

Resolves #2423 

## New API

I did my best to make this as identical to the API for the `<Button>` component but we could make some changes to simplify.

```typescript
export interface LinkProps extends React.HTMLProps<HTMLAnchorElement> {
  /** The style colour of the link */
  color?: Color;
  /** The style mode of the element. */
  mode?: "solid" | "outline" | "inline";
  block?: boolean;
  type?: "button" | "link";
}
```

### Example

```tsx
import { Link } from '@common/components/Link';
...
<Link type="button" mode="solid" color="primary" href="/">Home</Link>
... 
```

### Simplify API

My only thought is that for the **type** prop, instead of accepting `"link" | "button"`, replace it with a `boolean` "button" prop that is false by default. The reason it is constructed that way is to mimc the `<Button>` which accepts `"button" | "submit" | "reset" | undefined`.  Though, with the `<Button>` component, it is mapping directly to an HTML property so it is not exactly apples to apples and could be changed to simplify.

#### Example

```tsx
export interface LinkProps extends React.HTMLProps<HTMLAnchorElement> {
...
button?: boolean;
...
}

const Link: React.FC<LinkProps> = ({ button = false, ... rest }) => {...}

<Link button mode="solid" color="primary" href="/">Home</Link>
```

### DIFF
--- a/frontend/admin/src/js/components/classification/ClassificationPage.tsx
+++ b/frontend/admin/src/js/components/classification/ClassificationPage.tsx
@@ -33,15 +33,19 @@ export const ClassificationPage: React.FC = () => {
             data-h2-flex-item="b(1of1) m(2of5)"
             data-h2-text-align="m(right)"
           >
-            <Button color="white" mode="outline">
-              <Link href={paths.classificationCreate()} title="">
-                {intl.formatMessage({
-                  defaultMessage: "Create Classification",
-                  description:
-                    "Heading displayed above the Create Classification form.",
-                })}
-              </Link>
-            </Button>
+            <Link
+              href={paths.classificationCreate()}
+              title=""
+              color="white"
+              mode="outline"
+              type="button"
+            >
+              {intl.formatMessage({
+                defaultMessage: "Create Classification",
+                description:
+                  "Heading displayed above the Create Classification form.",
+              })}
+            </Link>
           </div>
         </div>
       </header>
--- a/frontend/admin/src/js/components/classification/ClassificationPage.tsx
+++ b/frontend/admin/src/js/components/classification/ClassificationPage.tsx
@@ -1,6 +1,5 @@
 import React from "react";
 import { useIntl } from "react-intl";
-import Button from "@common/components/Button";
 import Link from "@common/components/Link";
 import { useAdminRoutes } from "../../adminRoutes";
 import { ClassificationTableApi } from "./ClassificationTable";
--- a/frontend/admin/src/js/components/cmoAsset/CmoAssetPage.tsx
+++ b/frontend/admin/src/js/components/cmoAsset/CmoAssetPage.tsx
@@ -1,6 +1,5 @@
 import React from "react";
 import { useIntl } from "react-intl";
-import Button from "@common/components/Button";
 import Link from "@common/components/Link";
 import { useAdminRoutes } from "../../adminRoutes";
 import { CmoAssetTableApi } from "./CmoAssetTable";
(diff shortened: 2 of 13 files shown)


petertgiles:
I think this is a PR that would benefit from a review by the designers.  I recommend posting a note on #desfam on slack asking them if they have time to take a look.  Maybe post a link to the issue, the PR, and the chromatic story page to make it quick and easy for them.

vd1992:
Going through, I think you've found all instances of `<Link>` nested in `<Button>` and vice-versa, didn't find anything else.
Good work!

petertgiles (changes requested):
I'm not sure there was any consensus in the discussion but at least they had a chance to review. :man_shrugging: Don't forget to remove the unused imports to the Button file.

esizer:
> I'm not sure there was any consensus in the discussion but at least they had a chance to review. 🤷‍♂️ Don't forget to remove the unused imports to the Button file.

I think there was enough to move forward. While they did not necessarily agree with two separate components for Button and Link, they approved the look and functionality.

petertgiles (changes requested):
I managed to find a few more unused Button imports :grin: 

DepartmentPage.tsx
PoolPage.tsx
PoolCandidatePage.tsx
SkillPage.tsx
SkillFamilyPage.tsx

esizer:
> I managed to find a few more unused Button imports 😁
> 
> DepartmentPage.tsx PoolPage.tsx PoolCandidatePage.tsx SkillPage.tsx SkillFamilyPage.tsx

I swear I ran the linter  😅

Resolved with [remove more unused Button imports](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2431/commits/ebaa1cb9309d5d50abb52779e7cb997f96202281)

petertgiles (approved):
Nice one!

PR #2446, opened 2022-04-11
Migrate faker.js to @faker-js/faker

### LINKED ISSUE #1700: Faker - project abandoned
The original dev behind faker.js has abandoned the project:
https://www.theverge.com/2022/1/9/22874949/developer-corrupts-open-source-libraries-projects-affected

We should consider moving off this dependency at some point.  It seems like there's an active fork of the project with open source support that has started at https://github.com/faker-js/faker.

Discussion around the fork: https://github.com/withshepherd/faker.js/issues/1

esizer:
## Summary

This fixes the security issue with the abandoned `faker.js` project by migrating to the maintained fork `@faker-js/faker`.

Ref: https://fakerjs.dev/migration-guide-v5/

Resolves #1700 

## Completed Tasks

 - [x] Removed old dev dependancy `faker` and associated types `@types/faker`
  - v6.0.0 now ships TypeScript types out of the box 🎉 
 - [x] Install forked `@faker-js/faker`
 - [x] Update imports of `faker` to new dep.
  - Update from default to named import to future proof for tree-shaking
 - [x] Add type annotations to `faker.random.arrayElements<T>`
 - [x] Fix bug where passing empty array to `faker.random.arrayElements<T>` caused issue


vd1992:
Do we want to do this step? As far as I can tell, we always set locale to 'En', don't see any case of using faked French data, in which case might as well import English only so the setLocale step is skipped. Performance impact is probably nigh non-existent 

image

petertgiles:
> As far as I can tell, we always set locale to 'En',

I would love to use some faked French for our localized strings but when I originally set it up there just wasn't much French in the library to use.  Has that changed with this fork of the project?

petertgiles:
> Performance impact is probably nigh non-existent

I'm not too worried about this.  Faker isn't used in our deployment bundle, only in Storybook.  At least that's how it should be!  :grin:

esizer:
> > Performance impact is probably nigh non-existent
> 
> I'm not too worried about this. Faker isn't used in our deployment bundle, only in Storybook. At least that's how it should be! 😁

Just took a look and it still seems to be mostly English. So maybe it wouldn't be a terrible idea to just import the EN locale as @vd1992 suggested.

esizer:
*Note:* Changed base temporarily to clear out PR incorrectly showing diffs from the main branch after merging to fix conflicts.

petertgiles (changes requested):
Nice, this is looking great!  I love that we can use generics with the generators now.  It looks like the package-lock file has been massively changed with all the integrity hashes stripped out.  Could you take a look?  Is your environment set up to comply with the engine consraints?
```
  "engines": {
    "node": "14.18.1",
    "npm": ">=8.0.0"
  },
```

esizer:
That is odd, @petertgiles. I'm not running those exact versions but it no major difference.

Screen Shot 2022-04-12 at 10 08 07 AM

esizer:
@petertgiles and @vd1992 Can you take a took at the lock file after [regen package-lock.json](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2446/commits/30028e38edcb93a6be3f371dc3cb64fea1668c88)? I reviewed it and it seems the integrity hashes are back. Not sure what happened but I just deleted it and re-ran `npm install`

petertgiles:
Sorry, still happening. :disappointed: 
image

PR #2447, opened 2022-04-11
Bug: 2358 - Fix Missing Breadcrumb Key

### LINKED ISSUE #2358: Breadcrumbs need key
The breadcrumbs component is emitting warnings:
```
Warning: Each child in a list should have a unique "key" prop.
```
Using the title should be unique enough for a breadcrumb key, I think.

esizer:
## Summary

Fixes the issue with missing keys on the `<Breadcrumb>` component.  While this is not the ideal solution, it should work since the breadcrumbs never need to re-render and are entirely replaced on page navigation.

Any suggestions on a better work around are very welcome.

Resolves #2358 

## Solution

To resolve the issue of the separators not containing a key coming from the `insertBetween<T>(separator: T, arr: T[]): T[]` function, we are looping through, cloning and applying the array index as a key for each element that does not have one. This should only be the separators.

```tsx
menuBar.map((el, index) =>
  React.cloneElement(el, { key: el.key || `sep-${index}` }),
)}
```


vd1992:
What is going on with the package-lock file?

tristan-orourke:
Good solution, and I think it would be fine. The only alternative I can think of would be to include the separator in the first links.map expression, so it can share the same key. I'm not sure its a better solution, but it might be a bit more conventional.

```tsx
// taking in the array of objects as defined above, create an array of Link components or a <span> if there is no href
  const arrayLinks = links.map((link, index) => {
    return (
      <Fragment key={link.title}>
        {/* Insert a Chevron before all but the first link, to separate them. */}
        {index > 0 && (
          <span
            data-h2-padding="b(right-left, xs)"
            data-h2-display="b(flex)"
            data-h2-align-items="b(center)"
          >
            <ChevronRightIcon style={{ width: "1.4rem" }} />
          </span>
        )}
        {/* The segment itself may be a link or just plain text. */}
        {link.href ? (
          <Link
            data-h2-display="b(flex)"
            data-h2-align-items="b(center)"
            href={link.href}
            title={link.title}
          >
            {link.icon || ""} {link.title}
          </Link>
        ) : (
          <span
            data-h2-display="b(flex)"
            data-h2-align-items="b(center)"
            data-h2-font-weight="b(700)"
          >
            {link.icon || ""} {link.title}
          </span>
        )}
      </Fragment>
    );
  });
```

esizer:
> Good solution, and I think it would be fine. The only alternative I can think of would be to include the separator in the first links.map expression, so it can share the same key. I'm not sure its a better solution, but it might be a bit more conventional.

I like the proposed solution. I'll implement it and confirm.

esizer:
@tristan-orourke Implemented your solution in [new breadcrumb key implementation](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2447/commits/77ccef3711b1e53a212d444c73ee88a512f0e22a) to avoid using array index as key. Thanks! 😄

Issue #1925, opened 2022-01-28
Experience Accordion component


tristan-orourke:
# Description
An accordion component representing an Experience. When opened, displays the list of associated Skills, additional information and a Link to the experience's edit form.

Note that the title and subtitle of the accordion are formatted slightly differently for different types of experience:
- work:
  - title:  <Role> at <Organization>
  - subtitle: "Since: <Start Month> <Start Year>" (is this only if still active? is it otherwise <Start Date> - <End Date>?)

[Wireframe](https://www.figma.com/proto/PrvXEDac0HVp7dF7SLWVFm/Candidate-profile-interfaces---V5.0.0?node-id=2340%3A4769&starting-point-node-id=1%3A2)

# Acceptance Criteria

- [ ] Storybook
- [ ] Jest tests:
  - [ ] Titles, subtitles and content appear correctly for different experiences

tristan-orourke:
Recent at top. Use end date, if no end date put at top

gobyrne:
Sort by most recent End Date at top.
All items with no end date should be at the top and sorted by most recent start date.

Jerryescandon:
Alright, hold on to your socks. 
I've added the Title and Subtitle for all the types of experiences in the next images,
but also,
we can't fit all the info we collect on the title alone, so I've added a mini section inside every expanded accordion to put in the complete info.

Here you go:

**Personal Experience**
![image](https://user-images.githubusercontent.com/16977254/156054401-c1930e87-c9c9-46f4-8b8b-72424880cd4d.png)

**Work Experience**
![image](https://user-images.githubusercontent.com/16977254/156054608-8c8fb16f-91a1-47f7-98a9-64547097e17d.png)

**Community Experience**
![image](https://user-images.githubusercontent.com/16977254/156054773-3e6733dd-bca0-48d8-90d1-e64acfd4f99b.png)

**Education**
![image](https://user-images.githubusercontent.com/16977254/156054847-e35c8bb1-7a4b-41f4-977c-f88e10258078.png)

**Awards**
![image](https://user-images.githubusercontent.com/16977254/156054882-23807178-2aa3-481a-9b49-cd55de947c11.png)

vd1992:
I was trying to figure out if I could tackle this and was playing around with things. 
Realized this appears to be blocked by #1788

If skills aren't linked to experiences, you cannot make the wireframe accordions I believe

tristan-orourke:
[...]

PR #2454, opened 2022-04-12
Feature/2325 few touchups for experience accordions

### LINKED ISSUE #2325: Few touchups For Experience Accordions
h/t @brindasasi noticed a few touchups needed for experience accordions:
1. Date format should be MMM YYYY
2. Skill titles should be in the primary colour

RVany:
Resolves #2325 
h/t @brindasasi noticed a few touchups needed for experience accordions:
1. Date format should be MMM YYYY
2. Skill titles should be in the primary colour

### DIFF
--- a/frontend/common/src/fakeData/fakeExperiences.ts
+++ b/frontend/common/src/fakeData/fakeExperiences.ts
@@ -58,6 +58,11 @@ const sampleSkill2: Skill = {
   name: theSkillString2,
   experienceSkillRecord: theExperienceSkillRecord,
 };
+function formattedDeate(date: string) {
+  const startMonth = date.substring(4, 7).toUpperCase();
+  const startYear = date.substring(11, 15);
+  return `${startMonth}  ${startYear}`;
+}
 
 // 5 generators to generate experiences of a certain type
 // actual generators start here
@@ -86,7 +91,7 @@ const generateAward = (): AwardExperience => {
       AwardedScope.Provincial,
       AwardedScope.SubOrganizational,
     ]),
-    awardedDate: faker.date.past().toString().slice(0, 15),
+    awardedDate: formattedDeate(faker.date.past().toString().slice(0, 15)),
     issuedBy: faker.company.companyName(),
     experienceSkillRecord: {
       details: faker.random.words(),
@@ -105,8 +110,8 @@ const generateCommunity = (): CommunityExperience => {
     title: faker.lorem.word(),
     organization: faker.company.companyName(),
     project: faker.lorem.word(),
-    startDate: faker.date.recent().toString().slice(0, 15),
-    endDate: faker.date.future().toString().slice(0, 15),
+    startDate: formattedDeate(faker.date.recent().toString().slice(0, 15)),
+    endDate: formattedDeate(faker.date.future().toString().slice(0, 15)),
     experienceSkillRecord: {
       details: faker.random.words(),
     },
@@ -140,8 +145,8 @@ const generateEducation = (): EducationExperience => {
       EducationStatus.SuccessCredential,
       EducationStatus.SuccessNoCredential,
     ]),
-    startDate: faker.date.recent().toString().slice(0, 15),
-    endDate: faker.date.future().toString().slice(0, 15),
+    startDate: formattedDeate(faker.date.recent().toString().slice(0, 15)),
+    endDate: formattedDeate(faker.date.future().toString().slice(0, 15)),
     thesisTitle: faker.random.words(),
     experienceSkillRecord: {
       details: faker.random.words(),
@@ -158,8 +163,8 @@ const generatePersonal = (): PersonalExperience => {
     skills: [sampleSkill1],
     details: faker.lorem.sentence(),
     title: faker.name.jobTitle(),
-    startDate: faker.date.recent().toString().slice(0, 15),
-    endDate: faker.date.future().toString().slice(0, 15),
+    startDate: formattedDeate(faker.date.recent().toString().slice(0, 15)),
+    endDate: formattedDeate(faker.date.future().toString().slice(0, 15)),
     description: faker.lorem.paragraph(),
     experienceSkillRecord: {
       details: faker.random.words(),
@@ -178,8 +183,8 @@ const generateWork = (): WorkExperience => {
     organization: faker.company.companyName(),
     role: faker.name.jobTitle(),
     division: faker.animal.bird(),
-    startDate: faker.date.past().toString().slice(0, 15),
-    endDate: faker.date.soon().toString().slice(0, 15),
+    startDate: formattedDeate(faker.date.past().toString().slice(0, 15)),
+    endDate: formattedDeate(faker.date.soon().toString().slice(0, 15)),
     experienceSkillRecord: {
       details: faker.random.words(),
     },
--- a/frontend/common/src/fakeData/fakeExperiences.ts
+++ b/frontend/common/src/fakeData/fakeExperiences.ts
@@ -91,7 +91,7 @@ const generateAward = (): AwardExperience => {
       AwardedScope.Provincial,
       AwardedScope.SubOrganizational,
     ]),
-    awardedDate: formattedDate(faker.date.past().toString().slice(0, 15)),
+    awardedDate: faker.date.past().toString().slice(0, 15),
     issuedBy: faker.company.companyName(),
     experienceSkillRecord: {
       details: faker.random.words(),
--- a/frontend/common/src/fakeData/fakeExperiences.ts
+++ b/frontend/common/src/fakeData/fakeExperiences.ts
@@ -58,7 +58,7 @@ const sampleSkill2: Skill = {
   name: theSkillString2,
   experienceSkillRecord: theExperienceSkillRecord,
 };
-function formattedDeate(date: string) {
+function formattedDate(date: string) {
   const startMonth = date.substring(4, 7).toUpperCase();
   const startYear = date.substring(11, 15);
   return `${startMonth}  ${startYear}`;
@@ -91,7 +91,7 @@ const generateAward = (): AwardExperience => {
       AwardedScope.Provincial,
       AwardedScope.SubOrganizational,
     ]),
-    awardedDate: formattedDeate(faker.date.past().toString().slice(0, 15)),
+    awardedDate: formattedDate(faker.date.past().toString().slice(0, 15)),
     issuedBy: faker.company.companyName(),
     experienceSkillRecord: {
       details: faker.random.words(),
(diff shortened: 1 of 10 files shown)


frontend/common/src/fakeData/fakeExperiences.ts
```diff
@@ -58,6 +58,11 @@ const sampleSkill2: Skill = {
   name: theSkillString2,
   experienceSkillRecord: theExperienceSkillRecord,
 };
+function formattedDeate(date: string) {
```
vd1992, line 61: Just going to be a bit picky, I assume you meant _formattedDate_

vd1992:
I think you forgot to include the
"Skill titles should be in the primary colour" 
bit in the issue. Skills do not appear highlighted

frontend/common/src/fakeData/fakeExperiences.ts
RVany: you are right!

RVany:
> I think you forgot to include the "Skill titles should be in the primary colour" bit in the issue. Skills do not appear in the purple text

I think I added the black color. 
image
@vd1992 @petertgiles It should be purple?

RVany:
@petertgiles Can you please share the wireframe? It looks like the bullet is misplaced and the Date Format is 
image
or 
image

vd1992:
Nice! Got those other things sorted out. 
Unfortunately, SkillAccordion stories appear to have broken as a result of changes to fakeExperiences 

image

RVany:
> Nice! Got those other things sorted out. Unfortunately, SkillAccordion stories appear to have broken as a result of changes to fakeExperiences
> 
> ![image](https://user-images.githubusercontent.com/40485260/163067157-cbc244df-3c6c-4541-bc66-f8d7a97bf623.

looks like woring in local. I will check again and didn't understand the reason you mentioned. Good to learn!

RVany:
> Nice! Got those other things sorted out. Unfortunately, SkillAccordion stories appear to have broken as a result of changes to fakeExperiences
> 
> image

Its working for me!
image

petertgiles (changes requested):
I see the broken skill accordions as well.  Since you've only touched the experience accordions it was probably not your fault. :smile:  Could you try to fix it while you're working on these, though?  If it turns out to be a big problem please open an issue on the board for it.

Also, could you use [the date formatting function that Brinda worked out](https://github.com/GCTC-NTGC/gc-digital-talent/blob/ca974164d5e50cb78d768aa8b38f964253700c3e/frontend/talentsearch/src/js/components/skills/SkillAccordion/SkillAccordion.tsx#L35)?  It's not as accurate to the wireframe but I think it's superior since it will localize months in French.  And they two types of accordions should probably match. :laughing:   You could copy/paste it but the better solution would probably be to move it to one of the common helper function files and update both accordion files to point to the shared location.

petertgiles:
> @petertgiles Can you please share the wireframe?

https://www.figma.com/proto/PrvXEDac0HVp7dF7SLWVFm/Candidate-profile-interfaces---V5.0.0?node-id=2340%3A4769&starting-point-node-id=1%3A2

RVany:
> 

Ok I try to do that. Thank you!

vd1992:
> I see the broken skill accordions as well. Since you've only touched the experience accordions it was probably not your fault. 😄 Could you try to fix it while you're working on these, though? If it turns out to be a big problem please open an issue on the board for it.
> 
> Also, could you use [the date formatting function that Brinda worked out](https://github.com/GCTC-NTGC/gc-digital-talent/blob/ca974164d5e50cb78d768aa8b38f964253700c3e/frontend/talentsearch/src/js/components/skills/SkillAccordion/SkillAccordion.tsx#L35)? It's not as accurate to the wireframe but I think it's superior since it will localize months in French. And they two types of accordions should probably match. 😆 You could copy/paste it but the better solution would probably be to move it to one of the common helper function files and update both accordion files to point to the shared location.

FakeExperiences was updated, Skill Accordion draws from that, so my guess was a tweak to that faker inadvertently broke Skill Accordion as that component needed things in a very specific way. 
All I see is more reason we need that Chromatic comparison thing to avoid something being broken in who-knows-what place and no one notices for weeks

RVany:
> 
> > I see the broken skill accordions as well. Since you've only touched the experience accordions it was probably not your fault. 😄 Could you try to fix it while you're working on these, though? If it turns out to be a big problem please open an issue on the board for it.
> > Also, could you use [the date formatting function that Brinda worked out](https://github.com/GCTC-NTGC/gc-digital-talent/blob/ca974164d5e50cb78d768aa8b38f964253700c3e/frontend/talentsearch/src/js/components/skills/SkillAccordion/SkillAccordion.tsx#L35)? It's not as accurate to the wireframe but I think it's superior since it will localize months in French. And they two types of accordions should probably match. 😆 You could copy/paste it but the better solution would probably be to move it to one of the common helper function files and update both accordion files to point to the shared location.
> 
> FakeExperiences was updated, Skill Accordion draws from that, so my guess was a tweak to that faker inadvertently broke Skill Accordion as that component needed things in a very specific way. All I see is more reason we need that Chromatic comparison thing to avoid something being broken in who-knows-what place and no one notices for weeks

How I couln't see any broken accordions. Anyway, I undo the changes in fakeExperiences

vd1992:
> > > I see the broken skill accordions as well. Since you've only touched the experience accordions it was probably not your fault. 😄 Could you try to fix it while you're working on these, though? If it turns out to be a big problem please open an issue on the board for it.
> > > Also, could you use [the date formatting function that Brinda worked out](https://github.com/GCTC-NTGC/gc-digital-talent/blob/ca974164d5e50cb78d768aa8b38f964253700c3e/frontend/talentsearch/src/js/components/skills/SkillAccordion/SkillAccordion.tsx#L35)? It's not as accurate to the wireframe but I think it's superior since it will localize months in French. And they two types of accordions should probably match. 😆 You could copy/paste it but the better solution would probably be to move it to one of the common helper function files and update both accordion files to point to the shared location.
> > 
> > 
> > FakeExperiences was updated, Skill Accordion draws from that, so my guess was a tweak to that faker inadvertently broke Skill Accordion as that component needed things in a very specific way. All I see is more reason we need that Chromatic comparison thing to avoid something being broken in who-knows-what place and no one notices for weeks
> 
> How I couldn't see any broken accordions. Anyway, I undo the changes in fakeExperiences

_Skill_ Accordions, you posted a picture of _Experience_ Accordions above. Different component

RVany:
> 



> > > > I see the broken skill accordions as well. Since you've only touched the experience accordions it was probably not your fault. 😄 Could you try to fix it while you're working on these, though? If it turns out to be a big problem please open an issue on the board for it.
> > > > Also, could you use [the date formatting function that Brinda worked out](https://github.com/GCTC-NTGC/gc-digital-talent/blob/ca974164d5e50cb78d768aa8b38f964253700c3e/frontend/talentsearch/src/js/components/skills/SkillAccordion/SkillAccordion.tsx#L35)? It's not as accurate to the wireframe but I think it's superior since it will localize months in French. And they two types of accordions should probably match. 😆 You could copy/paste it but the better solution would probably be to move it to one of the common helper function files and update both accordion files to point to the shared location.
> > > 
> > > 
> > > FakeExperiences was updated, Skill Accordion draws from that, so my guess was a tweak to that faker inadvertently broke Skill Accordion as that component needed things in a very specific way. All I see is more reason we need that Chromatic comparison thing to avoid something being broken in who-knows-what place and no one notices for weeks
> > 
> > 
> > How I couldn't see any broken accordions. Anyway, I undo the changes in fakeExperiences
> 
> _Skill_ Accordions, you posted a picture of _Experience_ Accordions above. Different component

Because Experience Accordions is the issue in this PR. But Skill Accordions also working for me both locally and remote. I even shared the screen to Peter. Check again please. I am trying to find out how I can format the date in 
Experince Acordinons without touching the fakeExperiences. Also that is not a proper way!

RVany:
> * experience accordion looks good
> * skill accordion back to normal
> 
> I think all the tasks are taken care of, good work. However, please take a look at the Jest test that is failing. Looks like the skillAccordion test, please update that so it works with your new date style

ok I will take a look. I never touch the test though!

vd1992:
the test probably broke because of formatting changes elsewhere, I think you just need to change the tests around a bit so they don't fail

RVany:
> the test probably broke because of formatting changes elsewhere, I think you just need to change the tests around a bit so they don't fail

ok. Thank you!

frontend/talentsearch/src/js/components/ExperienceAccordion/individualExperienceAccordions/CommunityAccordion.tsx
```diff
@@ -38,22 +36,12 @@ const CommunityAccordion: React.FunctionComponent<CommunityExperience> = ({
     <Accordion
       title={intl.formatMessage(
         {
-          defaultMessage: "{title} at {organization}",
+          defaultMessage: "{title} of {organization}",
```
petertgiles, line 39: Was this change intentional?  I'm having trouble finding Jerbo's post that had all the individual accordions.

petertgiles: Found it!  Community should be _at_ not _of_.  https://github.com/GCTC-NTGC/gc-digital-talent/issues/1925#issuecomment-1054640559

RVany: I don't think I change it.

petertgiles: You changed it in commit https://github.com/GCTC-NTGC/gc-digital-talent/pull/2454/commits/abb62a0633834bc3788d761179cb52d2adb549cf .  Hehe, git knows everything!

RVany: really? why I did that?

RVany: I changed it back.

RVany:
Now, the jest test is passed but date format in `frontend\common\src\fakeData\fakeExperiences.ts` is not matching the component's date format though.

petertgiles:
> s not matching the component's date format though.

The javascript Date parser is good enough to understand the date strings from both faker and the API so it's working OK. :smile:

RVany:
> > s not matching the component's date format though.
> 
> The javascript Date parser is good enough to understand the date strings from both faker and the API so it's working OK. 😄

Ok! Thank you!😀

PR #2458, opened 2022-04-12
2456 - fix config for jsx-ally

brindasasi:
I can see the linting working as per the axe-con. This will error out in the VSCODE editor  itself.  However CLI version of linting currently broken due to existing errors. We could only these accessiblity errors when current errors are cleared out. 
npm run lint exits with exit code 0 when it begins linting on talentsearch folder. 
So I'm blocked on there. 

Screen Shot 2022-04-12 at 1 57 01 PM
Screen Shot 2022-04-12 at 2 03 40 PM

We could also do a github action for  accessibility linting. I will look into that next. 

If above both are working , then I would think axe tool would be duplicated effort.


petertgiles (changes requested):
Looks cool!  I see the messages in VS code:
image
I do see the linting messages coming out in the github actions as well:
image

Please add the same changes to the indigenousapprenticeship folder.  Then I'll approve this spike!

I agree that this is a quick way to get some linting going and we can decide to move to the axe tool later if we see the value there.  Also, I added pile of new issues to the board to try to get our linting clean.  Once this is in I can do the same for any new a11y issues.

petertgiles (approved):
Nice one!  :+1:

PR #2478, opened 2022-04-13
Remove Required tag from some User & PoolCandidate fields

### LINKED ISSUE #2455: Admin - be less strict with required fields
# Description
Almost all fields are set to required on the admin dashboard. Some fields should NOT be required.

In PoolCandidates form, none of the multiselects should be required. The User Telephone should not be required.

In User form, telephone should not be required.

tristan-orourke:
Resolves #2455 

Removes required rule from multi-select inputs in PoolCandidate forms on Admin dashboard. An empty selection is reasonable for these fields.

Removes required rule from User telephone fields. 

Adds a default value for status field on new PoolCandidates, for convenience.

### DIFF
--- a/frontend/admin/src/js/components/poolCandidate/CreatePoolCandidate.tsx
+++ b/frontend/admin/src/js/components/poolCandidate/CreatePoolCandidate.tsx
@@ -16,7 +16,7 @@ import {
   Checkbox,
   RadioGroup,
 } from "@common/components/form";
-import { notEmpty } from "@common/helpers/util";
+import { empty, notEmpty } from "@common/helpers/util";
 import { currentDate, enumToOptions } from "@common/helpers/formUtils";
 import { navigate } from "@common/helpers/router";
 import { getLocale } from "@common/helpers/localize";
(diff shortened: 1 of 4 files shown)


petertgiles (changes requested):
When I try to create or update a user without a phone number I'm getting an error.  It looks like the validator is still running against the empty string.
image
Pool candidate changes seem OK though.

vd1992:
Yeah, the validate is tripping things up. I tried Googling, maybe this will help?

https://github.com/react-hook-form/react-hook-form/issues/1781#issuecomment-775632786

tristan-orourke:
@petertgiles @vd1992 Thanks. Instead of changing the API validation for phone numbers to include empty strings (which technically isn't a valid phone number 😛 ) I set the forms to submit null instead, which *is* allowed.

vd1992:
Is the linting/prettier being a bit weird for you too? I'm noticing some files get their spacing drastically changed, create and update poolCandidate in the latest commit

tristan-orourke:
> Is the linting/prettier being a bit weird for you too? I'm noticing some files get their spacing drastically changed, create and update poolCandidate in the latest commit

Definitely a bit awkward that one line of Prettier fixing changes the indentation of the entire file :/
But I do want to allow Prettier to do its thing. @vd1992 does it not recommend that change on your machine?

vd1992:
> > Is the linting/prettier being a bit weird for you too? I'm noticing some files get their spacing drastically changed, create and update poolCandidate in the latest commit
> 
> Definitely a bit awkward that one line of Prettier fixing changes the indentation of the entire file :/ But I do want to allow Prettier to do its thing. @vd1992 does it not recommend that change on your machine?

I had something weird where it wanted everything re-spaced, so I let it do it. Then later on, it was unhappy again wanting everything re-spaced. 
https://github.com/GCTC-NTGC/gc-digital-talent/pull/2323/commits/f907f621f7d7bb220f09956e5fe693a78168fa8b
https://github.com/GCTC-NTGC/gc-digital-talent/pull/2323/commits/3812aa1c315f17421b51b225d57eaaf33096b90b
InfoForm component spacing changed and then it wanted to change it back

vd1992:
Not happy at all! @tristan-orourke 
image

petertgiles (approved):
Works for me!  :+1:

PR #2481, opened 2022-04-14
Fix language search filtering out bilingual

### LINKED ISSUE #2457: Talent Search - searching for one language filters out bilingual people
# Description
On the search page, selecting English only or French only filters out a Bilingual candidate. They should still show up.

JamesHuf:
Resolves #2457 

Fixed language search filter to properly include bilingual people when searching for a specific language.


JamesHuf:
And I noticed the php tests are failing. I'll take a look at them tomorrow.

petertgiles:
I'm getting weird results locally.  Not sure if it's my environment, this change, or pre-existing yet.
image

petertgiles:
I dunno, I tried refreshing my environment and I'm still getting weird results.  I'm getting yet again different numbers when I submit the request and review it in admin.  Is this happening for you?

JamesHuf:
> I'm getting weird results locally. Not sure if it's my environment, this change, or pre-existing yet.

This is interesting. I (foolishly) hadn't actually run this query in the database but while I'm seeing different results than what's in the database I'm getting the same number I was getting before my changes so I'd assumed they'd worked. Can you try in main and see if you get different results?

petertgiles:
@Jerryescandon It's not really relevant to James' change here but I'm uncomfortable with "Any Language" here.  Like, will you accept someone who only speaks Korean?  Maybe something like "No Preference" would be better?
image

frontend/talentsearch/src/js/components/search/SearchForm.tsx
```diff
@@ -330,9 +330,10 @@ export const SearchForm: React.FunctionComponent<SearchFormProps> = ({
             idPrefix="languageAbility"
             legend="Language"
             name="languageAbility"
+            defaultSelected="any_language"
```
petertgiles, line 333: I understand why you're making this change but I find it a little confusing to introduce a magic string here that doesn't exist as an enum in the schema.  Can we do better?  Maybe declare a constant in this file with a comment about what it represents?  Then use it here and also on line 124.  The logic on line 124 is ugly enough that I might suggest factoring that out into its own little function, too.  Jerbo has suggested changing the name, too.

Jerryescandon:
> @Jerryescandon It's not really relevant to James' change here but I'm uncomfortable with "Any Language" here. Like, will you accept someone who only speaks Korean? Maybe something like "No Preference" would be better? image

This is actually on of the things that was flagged for the next update. This is how we have it on the upcoming prototype. But if y'all working on this might as well update it now.
image

petertgiles (dismissed):
It works!  I've suggesting a few tidy up steps but I'd say it's almost ready to merge.  Or if you don't like my tidy suggestions, tell me why and I'll approve as-is.  :grin:

frontend/talentsearch/src/js/components/search/SearchForm.tsx
petertgiles: If you do refactor the logic on line 124, please use the real constants, like `LanguageAbility.English` instead of `"ENGLISH"`.

tristan-orourke (approved):
Nice work. Code looks good, and the search results count now makes sense when testing locally.

PR #2500, opened 2022-04-15
logout bug in local environment

### LINKED ISSUE #2305: Empty Logout URI Not Handled Correctly
When the app is set to authentication with Laravel Auth it doesn't handle logout correctly.  It seems that even though the environment variable is empty it is still trying:

I think this is related to  #2262

vd1992:
resolves #2305 
just played around with things until the problem appeared to go away

I think there is some weirdness involved with the redirect, so I removed it, namely as it was throwing page not found errors while it was present. Not sure if there is some other problem present or if this "solution" fuddled things.


tristan-orourke (changes requested):
Commenting/uncommenting code in different environments is not a viable solution. This is exactly the problem environment variables are supposed to fix! Config files (in this case, our .env files) are the one thing expected to change in different environments. Config.sjs sounds like a config file, but its really a bit of code that's meant to make our environment variables accessible in code.

vd1992:
> Commenting/uncommenting code in different environments is not a viable solution. This is exactly the problem environment variables are supposed to fix! Config files (in this case, our .env files) are the one thing expected to change in different environments. Config.sjs sounds like a config file, but its really a bit of code that's meant to make our environment variables accessible in code.

Yep, Peter told me just that a little while ago 😅

vd1992:
@petertgiles 
Incorporated your ideas, SSIundefined changed it to what the default is due to a single space string not being falsy.

petertgiles (approved):
Nice.  I'd like to see some short comments in both files but I think the solution is solid.  I tested both with Laravel auth and SiC on my workstation and it worked well.

tristan-orourke (changes requested):
Nice, I'm pretty happy with this solution. I just have two suggestions:
1. I think it would make sense to do that check for all the variables in config.sjs
2. I think it should fallback to undefined instead of empty string. That way, other code (like adminConstants.ts) can tell if it was set or not, and use their own fallbacks if not.

vd1992:
followed @petertgiles suggestion with the tweak to adminConstants to get logout to work normally
(so that it operates as desired in AuthContainer)

tristan-orourke (approved):

PR #2524, opened 2022-04-20
Remove urql Errors from Toasts

### LINKED ISSUE #2354: Toast notification not styling correctly
Certain toast notifications appear to have the text overflow the bounds of the box as in the image below.

esizer:
## Summary

We were piping all urql errors to the toasts, displaying them to users which caused two issues.

1. Unfriendly error messages were being displayed to users
2. Came in the form of a long JSON string with no spaces which caused overflow issues

## Solution

Stopped displaying the urql errors from the toasts since it is not friendly to users.

Resolves #2354 

### Logging of urql Errors

If we are not currently, we should be logging these errors.

### Toast Character Count Limit

While it is still a good idea we come up with a character limit for toasts, I believe it should be a fairly soft one since a test was done with > 800 characters and, while not ideal, the toast handled the large amount of text decently. I believe that the main issue here was the lack of spaces which causes the odd overflow issues. As long as we ensure that friendly errors messages are being displayed, a character limit becomes a lower priority.

toast-large

### DIFF
--- a/frontend/admin/src/js/components/ClientProvider.tsx
+++ b/frontend/admin/src/js/components/ClientProvider.tsx
@@ -129,11 +129,14 @@ export const ClientProvider: React.FC<{ client?: Client }> = ({
       createClient({
         url: apiUri,
         exchanges: [
-          errorExchange({
-            onError: (error: CombinedError) => {
-              toast.error(error.message);
-            },
-          }),
+          /**
+           * Commented out to stop urlq errors being displayed in toasts
+           */
+          //   errorExchange({
+          //     onError: (error: CombinedError) => {
+          //       toast.error(error.message);
+          //     },
+          //   }),
           dedupExchange,
           cacheExchange,
           authExchange({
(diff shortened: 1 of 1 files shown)


petertgiles:
I agree with your premise that the root issue is urql messages being displayed as toasts, not the styling of the toasts themselves.  However, I'm uncomfortable with the error messages just disappearing.  Maybe we can send them to console.error for now, instead?

esizer:
> I agree with your premise that the root issue is urql messages being displayed as toasts, not the styling of the toasts themselves. However, I'm uncomfortable with the error messages just disappearing. Maybe we can send them to console.error for now, instead?

Agreed, updated in [log errors in console](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2524/commits/ef7c6ce7bd93ddd5833dcc4aa957bc3ff37323d7). We can leave this until we have a better logging solution.

tristan-orourke:
I think this is a good solution for this, for now. I'll create a separate issue to ensure we're logging errors correctly on the server.

tristan-orourke (approved):

patcon:
Maybe related to this general "where to log, but not leak too much info" problem, for later (not now 🙂 )

https://github.com/getsentry/sentry 

(Sentry = open core SaaS platform for logging/crash management used by many large orgs)

Issue #2217, opened 2022-03-10
Refactor Docker


gggrant:
Some loose thoughts as I think on this:

- inspiration of style will be https://github.com/compdemocracy/polis/blob/dev/docker-compose.yml
- we want to do e2e testing. this would be simpler with a more "conventional" docker setup, that we could run in GitHub Actions environment
- docker conventions of note
  - one process per container (php, apache, etc)
  - one dockerfile per section? (admin/talent/common)
    - potentially multistage build for static builds, ie single dockerfile (for react build)
- stretch goals (prob not in this PR)
  - deploy container image artifacts to github container repo
    - this would allow the containers to be built once, and then pulled into other jobs (no rebuild) for parallelized testing or other stuff
    - demo: https://github.com/g0v-network/polis/blob/dev/.github/workflows/release-docker-images.yml
      - https://github.com/orgs/g0v-network/packages?repo_name=polis
- great feature on smaller docker images: https://symflower.com/en/company/blog/2022/complete-guide-on-shrinking-container-images/

## Prior Art
(Potentially useful when this is revived)

- https://github.com/GCTC-NTGC/gc-digital-talent/pull/2451
- https://github.com/GCTC-NTGC/gc-digital-talent/pull/2945

### To Do
- [x] https://github.com/GCTC-NTGC/gc-digital-talent/issues/2324
  - [x] read peter's [docker manifesto](https://talent-cloud.slack.com/files/U02J583KLGP/F038HCD796E/docker_manifesto.md)
    - [x] @petertgiles  🙃 
    - [x] @patcon 
    - [x] @brindasasi 
    - [x] @JamesHuf 
    - [x] @gggrant 
    - [x] @RVany
    - [ ] others? [add yourself]
  - [x] schedule time -- @petertgiles [FRI MAR 25 @ 2-3pm]
  - [x] invite all interesteds [from thread](https://talent-cloud.slack.com/archives/CB424V61J/p1648129419092389) -- @petertgiles 
  - [x] do the call! -- all
- [x] optional: summarize goals as decided?
- [ ] optional: split out tasks/tickets/accept-criteria (nearterm, longterm, etc)
- [ ] https://github.com/GCTC-NTGC/gc-digital-talent/issues/2327

petertgiles:
Result of planning session 2022-03-25:

# Docker Manifesto

## Goals
1. An easy way for new hires/contributors to get a working state
2. A reliable way for existing developers to reset their working state
[...]

PR #2525, opened 2022-04-20
Fix node in maintenance container

### LINKED ISSUE #1641: Node - permission denied

petertgiles:
This branch changes the maintenance container build to solve the "permission denied" errors observed when running the container on linux (and linux in wsl).  It also removes the profiles usage that fails on older versions of docker compose.

Closes #1641


.github/workflows/check-lighthouse-schema.yml
```diff
@@ -23,7 +23,7 @@ jobs:
 
       - name: "Build docker image: api"
         working-directory: infrastructure
-        run: docker compose up api --detach
+        run: docker compose -f docker-compose.yml -f docker-compose.ci.yml up api --detach
```
petertgiles, line 26: Use an override file for CI services since profiles fails on slightly older versions of docker compose: https://docs.docker.com/compose/extends/#multiple-compose-files

esizer (approved):
Nice! Ran from fresh WSL2 Ubuntu 20.04 with no issue. Both `setup.sh` and `refresh_all.sh`! 🎉

petertgiles:
> Nice! Ran from fresh WSL2 Ubuntu 20.04 with no issue. Both `setup.sh` and `refresh_all.sh`! tada

Thanks Eric!  Any chance an Apple person could git it a whirl before I merge it?

patcon:
> Any chance an Apple person could give it a whirl before I merge it?

Doing now!

Will then try running our setup.sh script on GitHub Actions and see if this has unfucked that too 🎉

esizer:
Just tested this on my macbook and it worked fine on:

- MacOS 10.15.5
- Docker 4.6.1

patcon:
Yesssss ok, this fixed our ability to run current scripts on GitHub Actions, which isssss interesting...
https://github.com/patcon/gc-digital-talent/runs/6130579542?check_suite_focus=true

That means that we're no longer blocked on #2217 to get Cypress tests running 🤔

Issue #2427, opened 2022-04-08
Document should contain landmarks


esizer:
## Summary

With modern assistive technology, landmarks provide a simple way to navigate a webpage. The technologies allow users to navigate between HTML5 landmarks such as `<header>`, `<nav>`, `<main>` and `<footer>` landmarks.

Currently we have `<nav>`, `<header>` and `<footer>` but lack a `<main>` landmark.

This would work in conjunction with #2426 acting as a fallback for assistive technologies that do not support navigating by landmarks.

Related: #2426 

Ref: [Page must have one main landmark](https://dequeuniversity.com/rules/axe/4.4/landmark-one-main?application=AxeChrome)

## Proposed Solution

This would easily be solved by wrapping our router `{content}` in a `<main>`

```jsx
<main>
  {content}
</main>
```

Issue #2426, opened 2022-04-08
Missing "Skip to Main Content" / "Skip to nav" links


esizer:
## Summary

Both Admins and Talent Search applications are missing a link to allow assistive technology to skip the navigation and be directed to the main content. This is more of an issue for the Admin application as the main navigation is fairly significant.

Ref: https://webaim.org/techniques/skipnav/

## Proposed Solution

This would be as simple as adding a hidden link to a div wrapping the main content.

### Example

```jsx
<a class="sr-only" href="#main">Skip to main content</a>
<Header />
<main id="main">
  {content}
</main>
<Footer />
```

PR #2528, opened 2022-04-21
A11y - 2427 document landmarks

esizer:
## Summary

This is a simple one to add the `main` landmark which allows to [easier navigation](https://www.w3.org/TR/wai-aria-practices/examples/landmarks/at.html) for assistive technology users. As a fall back, we have also added a "skip to main content" link so users accessing the tools with assistive technology that do not support landmarks can still skip the main navigation.

Landmarks Ref: https://www.w3.org/TR/wai-aria-practices/examples/landmarks/main.html
Skip Links Ref: https://www.w3.org/WAI/WCAG21/Techniques/general/G1.html

Resolves: #2427 and #2426

### DIFF
--- a/frontend/admin/src/js/components/dashboard/Dashboard.tsx
+++ b/frontend/admin/src/js/components/dashboard/Dashboard.tsx
@@ -187,37 +187,47 @@ export const Dashboard: React.FC<{
   menuItems: ReactElement[];
   contentRoutes: Routes<RouterResult>;
 }> = ({ menuItems, contentRoutes }) => {
+  const intl = useIntl();
   const content = useRouter(contentRoutes, <AdminNotFound />);
   return (
-    <div className="container">
-      <section
-        className="dashboard"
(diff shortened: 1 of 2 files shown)


petertgiles (changes requested):
Very cool!  I'm not sure how to test this with a screen reader but I can confirm that the site still works properly without assistive technology and the links and main elements seem correct.  Could you please apply the change for the indigenousapprenticeship project?  It's a copy/paste of talentsearch right now.

esizer:
> Very cool! I'm not sure how to test this with a screen reader but I can confirm that the site still works properly without assistive technology and the links and main elements seem correct. Could you please apply the change for the indigenousapprenticeship project? It's a copy/paste of talentsearch right now.

I'm actually working on the homepage for that right now and it will be included in my PR next week sometime which is why it is not part of this.

petertgiles (approved):

PR #2535, opened 2022-04-22
Font weight 100 too light on Firefox Linux #2441

### LINKED ISSUE #2441: Font weight 100 too light on Firefox Linux
So, I know this is extremely niche but font weight 100 on Firefox in Linux (Pop OS) is too light to easily see.

When I bump it up to 200 it looks much better.

On Chromium in Linux and on Windows (Firefox, Chrome, and Edge) it looks heavier and there is no difference between 100 and 200.  This makes me think that the font in Firefox in Linux ([Fira Sans](https://fonts.google.com/specimen/Fira+Sans)) is the only one that actually has font weight light enough to show a realistic 100.  Could we bump all the uses of 100 up to 200?

- [ ] Test in Lighthouse for contrast ratio

RVany:
Resolves #2441 
So, I know this is extremely niche but font weight 100 on Firefox in Linux (Pop OS) is too light to easily see.
image.png
When I bump it up to 200 it looks much better.
image.png
On Chromium in Linux and on Windows (Firefox, Chrome, and Edge) it looks heavier and there is no difference between 100 and 200. This makes me think that the font in Firefox in Linux ([Fira Sans](https://fonts.google.com/specimen/Fira+Sans)) is the only one that actually has font weight light enough to show a realistic 100. Could we bump all the uses of 100 up to 200?

 Test in Lighthouse for contrast ratio

### DIFF
--- a/frontend/admin/src/js/components/dashboard/Dashboard.tsx
+++ b/frontend/admin/src/js/components/dashboard/Dashboard.tsx
@@ -68,7 +68,7 @@ export const MenuLink: React.FC<MenuLinkProps> = ({
       <span
         {...(isActive(href, location.pathname)
           ? { "data-h2-font-weight": "b(700)" }
-          : { "data-h2-font-weight": "b(100)" })}
+          : { "data-h2-font-weight": "b(200)" })}
       >
         {text}
       </span>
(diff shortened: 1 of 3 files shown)


petertgiles:
I used the regex search function in vs code to find a few more!
image
The files in _dist_ folders are autogenerated, though.

petertgiles:
Oh, and don't forget to add a change management label to your PRs, now, too.

RVany:
> Oh, and don't forget to add a change management label to your PRs, now, too.

I din't get it.

RVany:
> 

When I use regex 0 results came to me. I will try again😀

petertgiles:
> I din't get it.

image

petertgiles (approved):
Nice!  Looks great in Firefox and Chromium is still scoring 100 on lighthouse.

RVany:
Chromium? I never heard🤐

PR #2537, opened 2022-04-22
linting in frontend - prettier

### LINKED ISSUE #2496: Linting - prettier/prettier
Fix all instances of the linting error with rule prettier/prettier.

Tips:
- Run `npm run lint --workspaces` from the frontend folder to manually run linting on all subprojects
- The [eslint-formatter-summary](https://github.com/mhipszki/eslint-formatter-summary) package can be handy for generating a linting summary by rule

vd1992:
resolves #2496 

ran npm linting thing, fixed prettier errors that popped up
a bit of weirdness, a few things popped up for prettier in /indigenousapprenticeship, but upon opening the relevant file there were no prettier highlights

### DIFF
--- a/frontend/admin/src/js/components/menu/SideMenu.tsx
+++ b/frontend/admin/src/js/components/menu/SideMenu.tsx
@@ -4,10 +4,7 @@ export const SideMenu: React.FunctionComponent<{
   items: ReactElement[];
 }> = ({ items }) => {
   return (
-    <nav
-      data-h2-padding="m(top-bottom, m)"
-      style={{ top: "0" }}
-    >
+    <nav data-h2-padding="m(top-bottom, m)" style={{ top: "0" }}>
       <ul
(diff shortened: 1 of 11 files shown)


esizer (approved):
Nice! I wouldn't worry about anything in the IAP folder since most of that was just setup as boilerplate and is expected to change in the week or two.

Something we could also consider is adding prettier as part of our pipeline to catch stuff like this earlier? I wouldn't expect this to be a priority at all but something to think about.

vd1992:
> Nice! I wouldn't worry about anything in the IAP folder since most of that was just setup as boilerplate and is expected to change in the week or two.
> 
> Something we could also consider is adding prettier as part of our pipeline to catch stuff like this earlier? I wouldn't expect this to be a priority at all but something to think about.

Could be something to consider, to avoid missed cases.

Issue #2536, opened 2022-04-22
Run smoke tests on scripts as part of CI


patcon:
🎟️ **Re-ticketed from:** https://github.com/GCTC-NTGC/gc-digital-talent/pull/2525#issuecomment-1104248612
🎯 **Success Criteria:** The everyday scripts we use (in `maintenance/scripts`) should each be able to run successfully in CI smoke test, and a working app should be minimally validated after each script run.

Thanks to @petertgiles's clever find in #2525, we're now able to run our existing docker setup on GitHub Actions. This means:
1. we can start smoke testing all our scripts, to know whether they've broken or fail to spin up a working app, and
2. we can now implement a MVP Cypress end-to-end test workflow right away, since we can deploy the existing docker setup on GitHub Actions.

The simplest version of this is to just:
1. spin up the bare docker containers
2. ping the site and confirm failure
3. run `setup.sh`, then ping each app and confirm minimal success via 200 response
4. run each `refresh_*.sh`, and confirm appropriate app still works

Opportunities in future: (not expected to be in first PR)
- use GitHub container repo to build containers once, and other jobs just pull from there (current build only take 1m20s, so not much savings here)
- use [build matrixes](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-running-multiple-versions-of-nodejs) to run the scripts in parallel (but each would spin up its own docker containers, and need to run setup.sh first which is 6m build time, so maybe not worth the savings)

petertgiles:
Hi @patcon .  Since you're pulling in an issue that is not part of the current sprint backlog could you please:
- assign yourself as owner
- add the interruption label
- add the current sprint
- add an estimate

patcon:
Sure! And is there anything process-wise you'd recommend? Specifically if I prioritized this too unilaterally or did things in the wrong order?

(My thought process was that we kicked this to another issue in #2525, and that this gets us more directly to #1762, which I understood to be the reason we're prioritizing #2217 in current backlog. So this made sense in my mind, but I realize my process of checking in with others might need work 🙏 )

petertgiles:
> And is there anything process-wise you'd recommend?

Hehe, I'm not an agile expert but I think the recommendation is for devs to prioritize the issues in the sprint backlog.  :wink:  

But if you feel an issue is urgent enough to pull into the sprint (or we run out like last sprint!) you can use the interruption label so that it can be tracked and you'll need to add all the meta tags yourself.

[...]

PR #2538, opened 2022-04-22
Added smoke tests for setup scripts and docker environment

patcon:
Addresses https://github.com/GCTC-NTGC/gc-digital-talent/issues/2536

This does the following:
- creates a workflow that spins up the docker environment
- run it on pushes to "main" branch and any PRs, but skip if only markdown files have changed
- a series of tests are run that confirm:
  - there's no site served at first from /talent
  - the setup.sh completes successfully
  - a specific app successfully served for each of 4 paths after this
  - the app is broken intentionally and confirmed as such (deleting index.html)
  - the refresh_all.sh script completes successfully
  - a specific app successfully served for each of 4 paths after this
- this also adds docker layer caching between runs
- a little github action that cancels other running workflows, to keep two pushes from running at once (since this one's heavy)

### Open Questions
- is running setup.sh and refresh_all.sh good enough to give us confidence? Do we wish to add more or other scripts that might reasonably break on their own?

### DIFF
--- a/.github/workflows/docker-smoke-test.yml
+++ b/.github/workflows/docker-smoke-test.yml
@@ -13,6 +13,9 @@ jobs:
       COMPOSE_DOCKER_CLI_BUILD: 1
       # Use buildkit to speed up docker command
       DOCKER_BUILDKIT: 1
+    defaults:
+      run:
+        working-directory: infrastructure
     steps:
       # This ensures only the most recently push commit will keep running.
       - name: Cancel other active runs of this workflow
@@ -24,13 +27,27 @@ jobs:
         uses: actions/checkout@v2.3.4
 
       - name: Serve app via docker-compose
-        working-directory: infrastructure
         run: docker-compose up --detach
 
-      - name: Run setup.sh
-        working-directory: infrastructure
+      - name: "Confirm fail response: 404 Not Found"
+        run: curl http://localhost:8000/talent --output /dev/null --silent --write-out "%{http_code}"  | grep 404 --quiet
+
+      - name: "Run: setup.sh"
         run: docker-compose run --rm maintenance bash setup.sh
 
-      - name: Run refresh_all.sh
-        working-directory: infrastructure
+      - name: "Confirm success response: setup.sh"
+        run: |
+          curl --silent http://localhost:8000/talent | grep /talent/app.js
+          curl --silent http://localhost:8000/admin | grep /admin/app.js
+          curl --silent http://localhost:8000/indigenous-it-apprentice | grep /indigenous-it-apprentice/app.js
+          curl --silent http://localhost:8000/en/talent-cloud | grep "GC Talent | Talent Cloud"
+
+      - name: "Run: refresh_all.sh"
         run: docker-compose run --rm maintenance bash refresh_all.sh
+
+      - name: "Confirm success response: refresh_all.sh"
+        run: |
+          curl --silent http://localhost:8000/talent | grep /talent/app.js
+          curl --silent http://localhost:8000/admin | grep /admin/app.js
+          curl --silent http://localhost:8000/indigenous-it-apprentice | grep /indigenous-it-apprentice/app.js
+          curl --silent http://localhost:8000/en/talent-cloud | grep "GC Talent | Talent Cloud"
--- a/.github/workflows/docker-smoke-test.yml
+++ b/.github/workflows/docker-smoke-test.yml
@@ -13,9 +13,6 @@ jobs:
       COMPOSE_DOCKER_CLI_BUILD: 1
       # Use buildkit to speed up docker command
       DOCKER_BUILDKIT: 1
-    defaults:
-      run:
-        working-directory: infrastructure
     steps:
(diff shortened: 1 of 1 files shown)


patcon:
Ah, realized just now that the special action to cancel other running workflows (and avoid duplicate compute resources) has been merged into GitHub Action platform proper:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

I'll migrate to using that :)

patcon:
ok, done! ready for review!

petertgiles:
This is very cool!  I guess the layer caching isn't doing much right now.  Once loading vendor/node_modules is part of the image creation it will be doing a lot more lifting, right?

Would it be possible to add a test for the api project as well?  Maybe confirm a 200 from the /graphql endpoint or something?

patcon:
> Once loading vendor/node_modules is part of the image creation it will be doing a lot more lifting, right?

Exactly!

> Would it be possible to add a test for the api project as well? Maybe confirm a 200 from the /graphql endpoint or something?

Good idea. Will do!

Related: Is there any sort of "ping" or "healthcheck" or "meta" endpoint you'd suggest? And if not, is it worth adding something for that? The ideal imho would be a "ping" endpoint (simply checking existence, without database check) and then some other endpoint whose response depends in the lightest way on something in database.

petertgiles:
> Is there any sort of "ping" or "healthcheck" or "meta" endpoint you'd suggest?

If a database exists you should be able to hit /graphql or /graphql-playground.  The api project also handle login/logout so you should get responses from /login, /auth-callback, and /refresh.

I like the idea of healthcheck endpoint.  I've been seeing references to this in my docker research.  I'm not sure it's too useful for a smoketest though.  For a healthcheck endpoint you mostly want to know if your service is still running and not down because the server crashed or you've got a flood of traffic.  For a github actions smoketest I want to know if the program still performs basic functionality after a code test.  Pretty different tests, I think.

patcon:
I don't particularly like that I'm repeating myself in the workflows. I also don't like that running things from the workflows requires copy-pasting from YAML workflow files to execute commands.

In my projects, I would have Makefile tasks for this, but I understand that that's a large conversation.

For now, any suggestion where I could put these? The most sensible place would strike me as something like `infrastructure/bin`:

```sh
# infrastructure/bin/check_failure_talent.sh
#!/usr/bin/env sh
set -e
# Match any non-200 HTTP response code (301 or 404)
curl http://localhost:8000/talent --output /dev/null --silent --write-out "%{http_code}"  | grep --invert-match 200
```

```
# infrastructure/bin/check_success_all.sh
#!/usr/bin/env sh
set -e
curl --silent http://localhost:8000/talent | grep /talent/app.js
curl --silent http://localhost:8000/admin | grep /admin/app.js
curl --silent http://localhost:8000/indigenous-it-apprentice | grep /indigenous-it-apprentice/app.js
curl --silent http://localhost:8000/en/talent-cloud | grep "GC Talent | Talent Cloud"
# Test that GraphQL API returns some data, and that the HTTP code is 200.
# See: https://www.apollographql.com/docs/apollo-server/monitoring/health-checks/#graphql-level-health-checks
curl "http://localhost:8000/graphql?query=\{__typename\}" --write-out "%{http_code}" --silent | grep '"data"' | grep 200
```

Thoughts?

petertgiles:
> I don't particularly like that I'm repeating myself in the workflows. I also don't like that running things from the workflows requires copy-pasting from YAML workflow files to execute commands.

Yeah, I get the desire to be DRY for sure.  At the same time, this is a new flow and we've got the docker refactoring and the IAP stand up happening simultaneously.  Let's get it merged in a simple form so we can start living with it then start to polish.  I have feeling we're going to want to refine the actual tests being run, too.

patcon:
(I pasted a graphql docs link above in an EDIT)

> For a github actions smoketest I want to know if the program still performs basic functionality after a code test.

Agree, we can perform /login or whatever, but I generally prefer self-describing things. For e.g., I'm using a suggested minimal check from the docs (`curl https://your.server/graphql?query=%7B__typename%7D
`), but I need to comment "this is a health check" so someone reading along can know _why_ it's there. That's usually my logic with API design. If something has a purpose, the path should tell a story of the purpose. Maybe it's just an alias though :) It's small here, but I think it generally affects how I design APIs

patcon:
Thanks for review! I'll merge whenever it looks ok 🙂

petertgiles:
> I'll kick to later

Sure, stick an issue on the board!

patcon:
> Sure, stick an issue on the board!

To explain, I don't see it as a specific debt issue (to address in the abstract all at once), but just something we start doing and try out in-flow of existing work? Then, if it looks/feels good, we'd decide to "cover our tracks" in that larger issue and "fix the debt" (once we have confidence that it is in fact debt). So it'll basically just be a future issue like this one that I just try it out in, in my suggested mode of work :)

Sound ok as a style of operation?

patcon:
weeeeee

PR #2540, opened 2022-04-22
Add cache option to BasicForm

### LINKED ISSUE #2530: Reusable solution for Form data caching
# Description
We want to make it hard to accidentally lose data in a partially-filled form. Our current solution is to continuously save the form data in localStorage, and load it from there (if possible) when first opening. This functionality should be made very easy to add to any new or existing form, whether using our BasicForm component or the react-hook-form useForm hook.

QUESTION: I shared some questions/concerns with how we're doing this [here](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2480/files#r855489951). My main thoughts are to use session storage instead of local storage, and very unique cache keys. Needs feedback from other devs.

# Acceptance Criteria
- [ ] Easy to turn on the caching solution in BasicForm
- [ ] Easy to use when using `useForm` manually
- [ ] Does not cause excessive re-renders (I worry that using `useWatch` or `useLocalStorage` naively will force the entire form to re-render each time one form input changes, which I think can be avoided?)
- [ ] Does not result in forms populated with "stale" or irrelevant data (see [comment](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2480/files#r855489951))

tristan-orourke:
Resolves #2530.

Allows you to supply a cacheKey to BasicForm to have it cache form contents in session storage. Using session storage instead of local storage addresses some concerns about stale data, though it won't save your data if you close a tab. I want to check in with designers on what they would consider the optimal business logic.


esizer (approved):
This is great, tested with the About Me form and it worked perfectly. Love the abstraction of the storage utils to allow for session or local.

PR #2559, opened 2022-04-25
More Prettier Linting

petertgiles:
I found a few more files to pretty lint

Related to #2496


vd1992:
I don't think some of those should even have linting problems, they were reviewed and merged recently, I think something is up with our Prettiers.

vd1992:
for comparison purposes, I could execute Prettier on a directory and commit it
gives a point of reference

vd1992:
prettier vscode

image

petertgiles:
I'll sit down with Vachan some time and try to figure this out.  For now I'll get rid of this PR.

PR #2560, opened 2022-04-25
Fix null operational requriement handling for search

### LINKED ISSUE #2516: Search Page - Operational Requirements Filter
🎟️ **Re-ticketed from** [Slack message](https://talent-cloud.slack.com/archives/CB424V61J/p1650373227569169) 🔒 

A blank search page produces a graphql query like this:
The issue is that the empty operational requirement filter is being interpreted as "any **non-null** operational requirements" rather than "any operational requirements".  If it is removed, all the candidates are returned.
The SQL query being generated:
image
A simpler query for demonstration:
image

Acceptance Critera:

- [ ] Automated test to detect bug
- [ ] Fix bug

petertgiles:
This branch adds a test to demonstrate, and a fix for, a search bug in the talent site.  If a pool candidate has a null field for operational requirement they are not returned when searching for candidates without an operational requirement filter.

The solution is simply to test for an empty filter clause and not populate the filter in the query.

I don't think any of the other fields have the same bug.  Only Work Locations uses the same JsonContains function and it works differently by building the query with a ForEach loop.

Closes #2516

### DIFF
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -179,6 +179,11 @@ RAWSQL2;
     }
     public function filterByOperationalRequirements(Builder $query, ?array $operationalRequirements): Builder
     {
+        // if no filters provided then return query unchanged
+        if (empty($operationalRequirements)) {
+            return $query;
+        }
+
         // OperationalRequirements act as an AND filter. The query should only return candidates willing to accept ALL of the requirements.
             $query->whereJsonContains('accepted_operational_requirements', $operationalRequirements);
         return $query;
--- a/api/tests/Feature/PoolCandidateTest.php
+++ b/api/tests/Feature/PoolCandidateTest.php
@@ -144,7 +144,7 @@ class PoolCandidateTest extends TestCase
   {
     // Create initial data.
     PoolCandidate::factory()->count(5)->create([
-      'accepted_operational_requirements' => [],
+      'accepted_operational_requirements' => null,
     ]);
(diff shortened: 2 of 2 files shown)


petertgiles:
To see how many total candidates there are in a pool you can run a SQL query like this:
```
select *
from "pool_candidates"
where "pool_id" in ('878f59c4-f2b5-4140-95c9-3cdad74dd99f')
```
And to run with a manual graphql query you can try this in playground:
```
query countPoolCandidates($where: PoolCandidateFilterInput) { 
  countPoolCandidates(where: $where) 
} 
```
```
{
  "where":{
    "classifications":[],
    "cmoAssets":[],
    "operationalRequirements":[],
    "hasDiploma":false,
    "hasDisability":false,
    "isIndigenous":false,
    "isVisibleMinority":false,
    "isWoman":false,
    "workRegions":[],
    "pools":[{"id":"878f59c4-f2b5-4140-95c9-3cdad74dd99f"}]
  }
}
```

vd1992 (approved):
- built main, then ran sample graphQL query at playground
- needed to replace pool ID locally (IDs may vary locally?)
- replicated bug, removing empty operational array brought candidate count from 0 to 30

then
- checked out and built branch
- presence of `operationalRequirements: []` no longer impacts results, candidates count remains at 30
- problem appears resolved 

Good work!

vd1992:
Also, in the sample query, removing or keeping the other `[]` selectors seemed to have no impact. So, nothing else seems like it suffers from the same error.

PR #2561, opened 2022-04-26
Feature - Common Dialog Component

### LINKED ISSUE #2533: Common - Modal dialog component

esizer:
## Summary

This adds a modal dialog component to the `@common`project for use in the other frontend projects. This is intentionally simplistic to allow for expanding on it using child components and will be expanded on with a Standard and Confirmation dialog as found in the  [Design System](https://www.figma.com/file/5w15OeN9YiOR4UX0XKW7so/Digital-Talent---Design-documentation?node-id=5%3A9).

Resolves #2533 

## Dependencies

With this PR we are introducing a new dependencies to which is a low-level component that provides base functionality and accessibility for a dialog.

- [@reach/dialog](https://reach.tech/dialog)

### To Do

- [x] Add different types outlined in [Design System](https://www.figma.com/file/5w15OeN9YiOR4UX0XKW7so/Digital-Talent---Design-documentation?node-id=5%3A9)
- [x] Provide a way to add buttons to footer
- [x] Allow for some customization (colours?)
- [x] Write tests
- [x] Write stories
- [x] ~~Export the child components to expose higher control over the functionality/styling?~~

### Usage

```tsx
...
import Dialog from "@common/Dialog"

const Example: React.FC = () => {
  const [isOpen, setOpen] = React.useState<boolean>(false);

 const open = () => setOpen(true);
 const close => () => setOpen(false);

  return (
    <>
      <button onClick={open}>Open</button>
      <Dialog isOpen={isOpen}  onDismiss={close} title="Dialog Title">
        <p>Dialog content.<p>
      </Dialog>
    </>
  );
}
...
```

#### title `string` - _required_

Adds a title to the `dialog` element that is both visual as well as descriptive for the `aria-labelledby` attribute.

#### subtitle `string`

Add a smaller subtitle below the title of the dialog.

#### isOpen `boolean` - _required_

This prop controls if the dialog should be displayed or not

#### onDismiss `(event: React.MouseEvent | React.KeyboardEvent) => void` - _required_

This function is called when the dialog would like to be dismissed via a user pressing `esc` or clicking outside of the modal.

#### initialFocusRef `React.RefObject<HTMLElement>`

By default, the modal will focus the first focusable element. If you would like to control which element is focused, you can pass a reference to it with this prop.

#### confirmation `boolean` - default: false

Determines if the dialog should use the confirmation style from the [Design System](https://www.figma.com/file/5w15OeN9YiOR4UX0XKW7so/Digital-Talent---Design-documentation?node-id=5%3A9)

#### color `Color {primary | secondary | tertiary}` - default: primary

Changes the colour of the header if `confirmation` is `false`.  Options are:

- primary (default): purple
- secondary: blue
- tertiary: pink


petertgiles:
Yeah, this works really nicely and the code looks super clean.

About the todos:
Add different types outlined in [Design System](https://www.figma.com/file/5w15OeN9YiOR4UX0XKW7so/Digital-Talent---Design-documentation?node-id=5%3A9) - yep
Provide a way to add buttons to footer - for sure
Allow for some customization (colours?) - I think so, probably take cues from the button component
Write tests - nice
Write stories - yeah
Export the child components to expose higher control over the functionality/styling? - I think probably not.  Let's enforce conformity until we have a good reason not to

petertgiles:
Lol, your NPM lock files are producing `resolve` and `integrity` values, too!

esizer:
> Export the child components to expose higher control over the functionality/styling? - I think probably not. Let's enforce conformity until we have a good reason not to

Agreed, I will remove this from the to do list

frontend/common/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,123 @@
+import React from "react";
+import { XIcon } from "@heroicons/react/outline";
+
+import Overlay from "./Overlay";
+import Content from "./Content";
+
+import "@reach/dialog/styles.css";
+import "./dialog.css";
+
+export type Color = "primary" | "secondary" | "tertiary";
```
petertgiles, line 10: How will this work if we share the component between talentsearch and indigenousapprenticeship?  When I set up the hello world PR I added the *ia* prefix to the colours in the hydrogen config.  Maybe something like that?

frontend/common/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,146 @@
...
+      name: "confirmation",
+      type: { name: "boolean", required: false },
+    },
+    color: {
+      options: ["primary", "secondary", "tertiary"],
+    },
+  },
+} as Meta;
+
+const TemplateDialog: Story = (args) => {
```
petertgiles, line 45: The template starts off with the dialog hidden.  This makes it interactive but unfortunately it is hidden from the Chromatic snapshots.  Can we have it start with the dialog open?

petertgiles: For example: 
image

frontend/common/src/components/Dialog/Dialog.tsx
esizer: I can add the ia prefix to match that, since the header has a gradient background i simply used the three gradients available in the hydrogen config.

We could make it more explicit and do `type Color = "purple" | "ia-purple" | "ia-pink";`?

petertgiles: Yeah, I like that better.  I think the best would be something like "ts-primary" and "ia-primary" but I'm not sure the designers have designated colours like that.

esizer: Updated with [update colour names](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2561/commits/36b7ae19914a86f04a5c627dbb4faece63a6b0de). I used your suggestion of prefixing primary, secondary etc. to match the button component more closely.

frontend/common/src/components/Dialog/Dialog.stories.tsx
esizer: Resolved with [make dialog open by default in story](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2561/commits/f26b1b1b1d63f9554ff5dc98b650830ced33413f)

petertgiles (approved):
Looks super nice.  Another A+, Eric!

PR #2564, opened 2022-04-26
Add Applicant Profile Feature Flag

### LINKED ISSUE #2363: Figure out a "feature flag" solution
# Description
We need a way to add features to main branch without releasing them to public. (eg routes to new Applicant Profile pages)

This issue currently up for discussion.

petertgiles:
This branch shows a possible solution to the "feature flag" spike.  It intentionally follows the "build it the stupid way first" philosophy. :grin: 

It adds to the talent search project:
- A sample Applicant Profile feature flag 
- Logic in the webpack config to add flag properties to the process.env object
- A guard on the government-information route

This solution was adapted from [this blog post](https://medium.com/nick-prokesch/usefeatureflag-feature-flags-with-react-and-webpack-13f4b8ece88f).

Another possible approach would be a short-lived release flag that would be simply removed from the code when ready to release instead of leaving the flags in code permanently.

Another possible approach would be a "javascript only" solution instead of using environment variables.  I think the environment variables make sense in our "PHP and Azure" world.

I think guarding pages by route is a sensible approach but if we wanted to make it less stupid (and more complicated!) here are some possible enhancements:
- a hook and component decorator like in the blog post
- A new library from NPM
- A third-party flag management service

Let me know what you think!

Closes #2363


esizer:
This seems like a good place to start for feature flags. The only concerns I have are fairly minor and I think it is a future issue that should be easily handled.

Obviously you already touched on this but once we have a certain number of flags, we may run into issues with scaling it. However, the implementation of a hook to manage this like the one in the blog post linked would handle this fairly easily and if we really need to, we can turn to a 3rd party library.

I think at some point we would need to remove certain flags after release to reduce debt. A process for a feature could be:

1. Add feature behind flag set to false in prod
2. Release feature by setting flag to true in prod
3. After a certain period of time, remove flag entirely

This would allow us to release the feature and quickly turn it off again if something bad happens. We then have a period of time where we can easily turn a feature off until we are confident in the new feature. I'm not sure a set period of time would work for this and it would be all confidence based.

tristan-orourke (commented):
Cool. I love how simple it is. I like prefixing env vars with FEATURE_. I have a few thoughts: 

I'm interested to hear your thinking for putting the guard in the routes file. My first thought would be to put it in the Router file (`frontend/talentsearch/src/js/components/Router.tsx`).

I agree with @esizer that we'd want to remove these conditions eventually. Could we make a little `checkFeatureFlag` utility function somewhere? It would be really dumb - just check for a feature key and return a boolean. But if we use it consistently, it'll be easy later to find all the places we are using it. I like the idea that we could use the condition pretty flexibly - sometimes to show/hide a whole route, sometimes to conditionally render some components on a page, maybe somewhere to change a url or functionality. Having a common utility function to use will make it easier to keep track of all the places we do that.

For injecting the FEATURE_ variable itself, shouldn't we use apache_env+config.sjs approach? We'll likely want to turn this on in DEV or UAT before turning it on in PROD.

petertgiles:
> shouldn't we use apache_env+config.sjs approach

Sorry, I'm still a bit hazy on the deployment process.  So the idea is that we want to compile one artifact for all verticals and anything that happens in webpack will be baked in for all three.  But Apache env can be different in each server, right?

tristan-orourke:
> we want to compile one artifact for all verticals and anything that happens in webpack will be baked in for all three. But Apache env can be different in each server, right?

Yeah, exactly. That's the situation right now anyway. The way we're importing apache env vars feels very hacky to me, but its like you said, "the stupid way first" 😛

petertgiles:
Here's a solution that uses Apache and a hook for the flags.  If you remove a flag from the hook the compiler should error all the uses for you before you can recompile.

I also converted the "data" object in the config.sjs file to a map to make it easier to access in a typed way.  If people like this I can convert the variables in admin to this form as well.

The downside of the Apache method is that you have to add or remove variables in 4 places:
- The apache_env file (or Azure environment)
- The useFeatureFlags.ts hook file
- The .htacess file (for the PassEnv directive)
- The config.sjs file to add it to the Window object

That's a bit error prone and painful!  I'll keep thinking of options.

petertgiles:
I've switched to a generalized `useRuntimeVariable` hook that can be used by anything injected by Apache.  

I've also switched to using the feature flag by name so that brings down variables setup from 4 places to 3.  I'm not sure if it's possible to bring it down any further.  I found the `printenv` directive but I'm uncomfortable with just dumping out all the variables publicly.  Maybe it would be fine but it seems too easy to accidentally lose a secret.

Anyway, it guards the Router file instead of the Routes file now (thanks Tristan :smile:).  Finding all the flags should be as simple as searching for `useRuntimeVariable("FEATURE_`.

frontend/talentsearch/src/js/components/Router.tsx
```diff
@@ -114,7 +120,12 @@ export const Router: React.FC = () => {
     <ClientProvider>
       <PageContainer
         menuItems={menuItems}
-        contentRoutes={routes(talentPaths, profilePaths)}
+        contentRoutes={[
+          ...talentRoutes(talentPaths),
+          ...(isStringTrue(useRuntimeVariable("FEATURE_APPLICANTPROFILE"))
```
tristan-orourke, line 125: I would define a helper function which combines isStringTrue with useRuntimeVariable for you, since the use of isStringTrue is not totally obvious. eg
```
checkFeatureFlag = (name: string): boolean => {
   // We use isStringTrue here because an ENV variable which is left undefined may show up as a runtime variable as a string (eg "false" or "(none)") instead of false or undefined.
    return isStringTrue(useRuntimeVariable(name));
 }
```

frontend/common/src/hooks/useRuntimeVariable.ts
```diff
@@ -0,0 +1,14 @@
+interface HasServerConfig {
+  __SERVER_CONFIG__?: Map<string, string>;
+}
+
+/**
+ * A hook to retrieve an environment variable value from the window object
+ */
+export const useRuntimeVariable = (name: string): string | undefined => {
```
tristan-orourke, line 8: From what I can till, this function has no reason to be a hook. Hooks tend to rely on other hooks and therefore only work inside react components, but this isn't under that restriction. It would make more sense to me to live in the helpers folder.

tristan-orourke (changes requested):
I see what you mean by the pain of having to put the feature flag in many places. I'm not sure what's the right way to improve that, but I'm okay with putting it off, especially when we expect to switch to Nginx at some point.

frontend/talentsearch/src/js/components/Router.tsx
petertgiles: Good call, I've added in the checkFeatureFlag help function.

tristan-orourke (approved):

PR #2576, opened 2022-04-28
2329 add missing field to user

### LINKED ISSUE #2329: Missing field from Type User - gov employee type

JamesHuf:
Resolves #2329 

- Added govEmployeeType to User in backend
- Updated GovernmentInfoForm to retrieve and update the new field
- Updated ProfilePage to display text for users current position type

### DIFF
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -47,6 +47,7 @@ use Illuminate\Auth\Authenticatable as AuthenticatableTrait;
  * @property array $expected_salary
  * @property boolean $would_accept_temporary
  * @property array $accepted_operational_requirements
+ * @property string $gov_employee_type
  * @property Illuminate\Support\Carbon $created_at
  * @property Illuminate\Support\Carbon $updated_at
  */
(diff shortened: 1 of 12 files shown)


esizer (approved):
Great work! Tested this and the changing of government type works well. 

However, I noticed an issue where if you change from any of the employment types other than student back to the student role, the group and classification is not unset. I don't believe this is within the scope of this issue/PR so I'll make a new issue for that.

PR #2582, opened 2022-04-28
Fix exhaustive deps linting errors

### LINKED ISSUE #2497: Linting - react-hooks/exhaustive-deps
Fix all instances of the linting error with rule [react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920#issuecomment-471070149).  If necessary, add a rule disable comment for instances where appropriate and add a justification comment.  

:warning:This is probably one of the trickier fixes!:warning:

Tips:
- Run `npm run lint --workspaces` from the frontend folder to manually run linting on all subprojects
- The [eslint-formatter-summary](https://github.com/mhipszki/eslint-formatter-summary) package can be handy for generating a linting summary by rule

petertgiles:
This branch addresses the [exhaustive deps](https://github.com/facebook/react/issues/14920) linting warnings we have in our code base.  I found four instances.

c6872c2a8d6d2b7e8e590d466b5f3c42e7179a98 solves the warning in router.tsx.  The solution provided is to stabilize the "not found" component in the parent by storing it in a ref.  This way it can be safely added to the dependency list without triggering continuous rendering.

8330f66cb16a336719b374b15446c230313e9c31 solves the warning in SearchRequestTable.tsx.  The missing dependency can be added without any issues.

aa27bdd4382b2e0c8ae0bc4115153895ee747cce solves the warning in SearchForm.tsx.  The solution provided stabilizes the debounce function in a ref.  The debounce function needs to be a single instance for the life of the component since it has its own internal state.  Once the instance is in a ref it does not need to be tracked as a dependency.

4496b6d5124beec141e64070b5c100681bf6aff7 solves the warning in useDeepCompareEffect.  I ended up disabling the warning here though it makes me a bit uneasy.  This function is essentially a wrapper around useEffect so I figured it wasn't too important to track the callback itself as a dependency since it was unlikely to change.  I did add useDeepCompareEffect to the eslint rules, though, so that proper checking can happen at the calling location.

Closes #2497

### DIFF
--- a/frontend/admin/.eslintrc
+++ b/frontend/admin/.eslintrc
@@ -63,7 +63,13 @@
         "react/jsx-props-no-spreading": "off",
         "react/require-default-props": "off",
         "react-hooks/rules-of-hooks": "error",
-        "react-hooks/exhaustive-deps": "warn",
+        "react-hooks/exhaustive-deps": [
+            "warn",
+            {
+                "additionalHooks":
+                    "(useDeepCompareEffect)"
+            }
+        ],
         "jsx-a11y/label-has-associated-control": [
             2,
             {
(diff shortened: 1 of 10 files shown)


esizer (approved):
Looks good to me, clever use of `useRef`! 😎

PR #2620, opened 2022-05-02
2573 - Improve search spinner a11y

### LINKED ISSUE #2573: Make front-page loading spinner accessible
<sup>Originally suggested by Grant [on Slack](https://talent-cloud.slack.com/archives/CB424V61J/p1635521467031500?thread_ts=1635520995.031000&cid=CB424V61J) 🔒 </sup>

The loading spinner that refreshes on each change to the form is not accessible. What's needed (from first blush, likely `aria-live` and `aria-busy` attributes) is covered in this great tutorial :)

https://dockyard.com/blog/2020/03/02/accessible-loading-indicatorswith-no-extra-elements

This is likely a one-off for now, but perhaps whoever does the work could think on whether there are any patterns for this moving forward, because any loading content will need these affordances 🙏 (also, perhaps a little research, to not reinvent)

## Screenshot

Screen Shot 2022-04-27 at 12 34 05 PM

## Relevant Code

https://github.com/GCTC-NTGC/gc-digital-talent/blob/3a5ef34603b7ca8f3bd878c7a4273c0951a40299/frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx#L14-L20

---

Things like this make me want to lean on other frameworks and not creating our design system from the ground up, as per @esizer's comment on modals in Slack (esp when we're still learning). Other systems have already figured out some of this stuff (even though we would need to be constrained by their choices)

https://carbondesignsystem.com/patterns/loading-pattern/#accessibility

esizer:
## Summary

implemented the changes discussed in issue #2573 

### `aria-live`

Added the `aria-live="polite"` attribute to the results/spinner container so that assistive technology will announce changes to the contents.

### Swap Contents

Instead of injecting the spinner into the message, we now replace the contents completely to avoid unnecessary audio clutter for assistive technology users. This will simplify improve the feedback provided.

Resolves #2573


petertgiles (changes requested):
TIL how aria live works.  Cool!  The spinner at the bottom of the page could probably use the same love. :smile:

frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx
```diff
@@ -6,59 +6,71 @@ interface EstimatedCandidatesProps {
...
+                    description:
+                      "Message to display when a search is in progress.",
+                  })}
+                </span>
+              </span>
+            ) : (
+              <>
+                {intl.formatMessage(
+                  {
+                    defaultMessage: `{candidateCount, plural, =1 {There is approximately <weight>{candidateCount}</weight> candidate right now who meets your criteria.} other {There are approximately <weight>{candidateCount}</weight> candidates right now who meet your criteria.}}`,
```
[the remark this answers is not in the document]
esizer: Hmm, I think it makes sense and am open to it. I'm fairly new to i18n with `react-intl` though so maybe someone else could weigh in?

frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx
```diff
@@ -6,59 +6,71 @@ interface EstimatedCandidatesProps {
...
-  function weight(msg: string) {
-    return updatePending ? (
-      <span className="lds-dual-ring" />
-    ) : (
-      <span data-h2-font-weight="b(800)">{msg}</span>
-    );
-  }
+    function weight(msg: string) {
+      return <span data-h2-font-weight="b(800)">{msg}</span>;
+    }
```
patcon, line 15: ❤️

frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx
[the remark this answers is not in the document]
esizer: So, I have read into this and I agree so I'm just going to commit your suggestion. Seems simple enough and allows for better scalability if we plan on adding more languages in the future.

frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx
```diff
@@ -6,59 +6,74 @@ interface EstimatedCandidatesProps {
...
```
patcon, line 15: Not sure if we have team conventions here, but for one-liner "components" I tend to omit the return:

```suggestion
    function weight(msg: string) (
      <span data-h2-font-weight="b(800)">{msg}</span>;
    )
```

frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx
```diff
@@ -6,59 +6,74 @@ interface EstimatedCandidatesProps {
...
+            {updatePending ? (
+              <span className="lds-dual-ring">
+                <span data-h2-visibility="b(invisible)">
+                  {intl.formatMessage({
+                    defaultMessage: "Searching...",
+                    description:
+                      "Message to display when a search is in progress.",
+                  })}
+                </span>
+              </span>
```
patcon, line 54: Thoughts on pulling these out into `spinner` function to make the logic more legible? (I'm trying to figure out my own balance of using ["Thinking in React"](https://reactjs.org/docs/thinking-in-react.html) style of componentization, but without so much overhead of putting them all in files, so I personally try to lean on little pseudo-component functions to bundle up markup. It's ok if you feel differently!)

```js
function spinner() (
  <span className="lds-dual-ring">
    <span data-h2-visibility="b(invisible)">
      {intl.formatMessage({
        defaultMessage: "Searching...",
        description:
          "Message to display when a search is in progress.",
      })}
    </span>
  </span>
)

**Review by patcon — APPROVED** [review]
LGTM! Left comments, but pls merge at your discretion!

**Comment by patcon** [comment]
This looks like a step in a good direction! I def agree that replacing the whole text with a spinner is more accessible than putting the spinner in the place of the number during loading :)

Also, since it was lost in a resolved comment, outstanding Q (not urgent it be addressed in this PR): Is the brevity of the dynamic text around the candidateCount an a11y issue? (since it's read aloud on each refresh) I feel we could make it more terse, without any loss. Example

> Candidate who meet criteria: 10

Also, less page jitter on change, which feels nice

**Comment by patcon** [comment]
Ah, just noticed: Seeing in chromatic that the page is shifted for loading forms.
https://www.chromatic.com/pullrequest?activeElementId=comparison-test-key-61e0a1efba513b709b06678f-1200&appId=61e099a1bb1465003a73d3ce&number=2620&view=changes

Is this expected?

**Comment by patcon** [comment]
Lastly, was looking at the chromatic diffs, and realized that only the big form was capturing this change. Is it worth adding some component stories to capture the states of this sidebox component? Thinking maybe:
- loading
- one result
- many results

File: frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx (hunk 4 above)
Comment (esizer) [reply frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx]: Great suggestion, the spinner has been moved to its own component for readability and reusability. 

[move spinner, add story, change candidate weight](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2620/commits/e5e0e3bc4eaf184123557ed90bfb0299fc3278a4)

**Comment by esizer** [comment]
> Ah, just noticed: Seeing in chromatic that the page is shifted for loading forms. https://www.chromatic.com/pullrequest?activeElementId=comparison-test-key-61e0a1efba513b709b06678f-1200&appId=61e099a1bb1465003a73d3ce&number=2620&view=changes
> 
> Is this expected?
> 
> EDIT: Funny, seems to not be an issue in live storybook... https://61e099a1bb1465003a73d3ce-leqgkkjeni.chromatic.com/?path=/story/search-page--updating-search-page

I've been having some issues with storybook in general today. Peter also noticed something to do with content shift being an issue with the IAP homepage that was only an issue in storybook, not sure if it is related or not.

**Comment by patcon** [comment]
> only an issue in storybook

Cool, thanks! Do you mean "only as issue in chromatic"? If it's something about their renderer, that's out of out control, but if it's happening in storybook (and we haven't upgraded storybook), then it's likely something in our own styling/components/layouts that's changed, so it feel more important to note when it started :)

**Comment by esizer** [comment]
> chromatic

I'm not sure actually, I took a second look and i **think** i know the cause for this PR but if I'm correct, it raises more questions.

By adding the spinner at the bottom of the page it is causing some shift in layout but that doesn't explain why the rest of the form shifted by a few pixels 🤔 

Screen Shot 2022-05-03 at 1 40 41 PM

File: frontend/talentsearch/src/js/components/Spinner/Spinner.stories.tsx [hunk 5]
```diff
@@ -0,0 +1,8 @@
+import type { Meta, Story } from "@storybook/react";
+
+import Spinner from "./Spinner";
+
+export default {
+  component: Spinner,
+  title: "Spinner",
+};
```
patcon, line 8: Cool! I think this is just the "metadata" object being exported, but you still need to export a Story in order for storybook to render one.

Something like

```tsx
const Template: Story = () => (
  <Spinner />
)

export const Default = Template.bind({});
```

esizer: Wow 😳, thanks! Resolved with [add spinner story component export](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2620/commits/f61d5c5ccd39fcb942d2b370b764cb78f32632bd)

patcon:
And this little spinner story is great! I def am in the camp for "more components get stories" (not just high-level forms).

But also, apologies, as I didn't make my suggestion very clear: I was asking whether you thought it worthwhile to add those stories (loading, one, many) for `EstimatedCandidates.stories.tsx`

patcon:
> By adding the spinner at the bottom of the page it is causing some shift in layout but that doesn't explain why the rest of the form shifted by a few pixels 🤔

Maybe we could just spin this out to track details? It feels a bit odd to track a bug in our tooling, but I have a feeling it might crop up in something user-facing eventually, and might be good to have a place to collate things ppl notice (even if non-urgent) :)

frontend/talentsearch/src/js/components/Spinner/Spinner.stories.tsx
patcon: haha Thank YOU!

petertgiles (approved):
Awesome :sunglasses:

PR #2623, opened 2022-05-02
Rebuild lock files

### LINKED ISSUE #2618: Failing Jest tests
After some dependabot updates were pushed in, Jest tests appear to have begun failing. A file might have shifted location 
The error logs range in length and extend over thousands of line. Here is a snippet attached

image

petertgiles:
It seems like something in the dependabot updates for today scrambled the lock files.  I've rebuilt them.  h/t @esizer for finding the dropped dependency.

Closes #2618


esizer (approved):
We may need to let people know to merge this in to get their tests (hopefully) passing.

PR #2626, opened 2022-05-02
Edit experience button prefills form

### LINKED ISSUE #1928: When clicking Edit Experience, you navigate to a pre-filled Experience form
# Description
Ensure the edit button links to a version of the Experience form which loads and populates with existing values.

This requires ensuring the Edit Experience links are correct and that routing is working as expected.

JamesHuf:
Resolves #1928 

Notes:
- I filled out the experience section of the profile page
- I changed the background color of the profile sections from grey to light grey to better match designs
- Updated edit button on accordions to link to appropriate prefilled form
- Fixed some bugs encountered while editing an existing experience


vd1992:
Do we want any sort of cap on how many experiences appear on the profile page? Or do we want on the profile page every single experience to be visible? 
Here
image

vd1992 (changes requested):
### Issues
- unable to delete skills, click to remove a skill does nothing
- toast still says "successfully added experience", might be good to add a separate toast in the update experience code chain to reflect a successful update operation
- might as well update the routing for the "cancel button", currently it points to `http://localhost:8000/profile` instead of `http:localhost:8000/talent/profile`
- if you add a skill in edit, it hides the other skills you have, it doesn't delete them but it makes you unable to edit their skill details as they've been hidden

### What Works
- pre-fill is successful and form fills with data
- edit operation is successful and experience updates are visible on profile page
- navigating to edit page works and it smoothly sends user back to profile page on update

Great job so far James! Not sure if all the issues need to be addressed in this

vd1992:
to demonstrate Issues bullet #4
http://g.recordit.co/Wi8TEjh6Vw.gif

frontend/talentsearch/src/js/components/Router.tsx
```diff
@@ -74,14 +74,29 @@ const routes = (
...
+      return {
+        component: (
+          <ExperienceFormContainer
+            experienceType={experienceType}
+            experienceId={experienceId}
+          />
+        ),
+      };
+    },
+  },
```
vd1992, line 99: It might be a good idea to introduce a sort of edit=true flag here. It will enable conditional operations that depends on someone editing an experience rather than creating one.

If Edit=True, you could have a different toast display, updated successfully instead of created successfully. 
It could also come in handy for the Delete Experience (separate issue), because then the delete button can be conditionally rendered. So it seems like something that could be of use for multiple tasks. 

Of course, you could implement it another way, rather than the edit=true passed in the route

JamesHuf: That's a good idea. Thanks for the suggestion.

frontend/talentsearch/src/js/components/ExperienceAccordion/individualExperienceAccordions/AwardAccordion.tsx
```diff
@@ -105,7 +110,14 @@ const AwardAccordion: React.FunctionComponent<AwardExperience> = ({
...
       <div data-h2-padding="b(left, l)">
-        <Button color="primary" mode="outline">
+        <Button
+          color="primary"
+          mode="outline"
+          onClick={(event) => {
+            event.preventDefault();
+            navigate(editUrl);
+          }}
+        >
```
tristan-orourke, line 120: Since this is doing a navigate, we should use a button-styled Link component instead of an actual Button component. (Button-styled Links are a new tool in our common project! 😁)

frontend/talentsearch/src/js/components/applicantProfile/ExperienceSection.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+const ExperienceSection: React.FunctionComponent<ExperienceSectionProps> = ({
+  experiences,
+}) => {
+  const intl = useIntl();
+  const locale = getLocale(intl);
+
+  const awardExperiences =
+    experiences
+      ?.filter((experience) => isAwardExperience(experience))
```
tristan-orourke, line 101: While the ExperienceSection is being refactored anyway, I can recommend a tiny stylistic change. 😁 Passing functions directly into `filter()`.
```suggestion
      ?.filter(isAwardExperience)
```

tristan-orourke (changes requested):
I haven't ran and tested it (since @vd1992's review seemed very thorough) but the code looks great! Just need to switch those buttons to links for a11y purposes.

tristan-orourke:
> Do we want any sort of cap on how many experiences appear on the profile page?

For now, no. If it gets out of hand, we could do something about it in a later issue.

JamesHuf:
I've addressed Tristan's feedback and the last three points Vachan raised. After struggling for a while I've given up on fixing the skills not properly being removed for now though. Feel free if anyone wants to pick it up and try fixing it.

vd1992:
- checked out latest
- built
- bullets 2 thru 4 addressed
- clicking to remove skill removes it from the form but yeah, as stated above, doesn't actually update yet
- no other quirks noted

nice!
I'll put in an approval once a decision is made about the removing skills bit

esizer:
I was going to see if I could get to the bottom of this since I put the form together but the route is not working for me 🤔 

I'm navigating to: `/en/talent/skills-and-experiences/award/8e859a4b-b340-4afa-bef1-7feb456e8400/edit` and `/en/talent/skills-and-experiences/award/create` but getting a 404 on both.

Am I doing something obviously wrong?

vd1992:
> I was going to see if I could get to the bottom of this since I put the form together but the route is not working for me 🤔
> 
> I'm navigating to: `/en/talent/skills-and-experiences/award/8e859a4b-b340-4afa-bef1-7feb456e8400/edit` and `/en/talent/skills-and-experiences/award/create` but getting a 404 on both.
> 
> Am I doing something obviously wrong?

have you injected the feature flag PHP var?

esizer:
> > I was going to see if I could get to the bottom of this since I put the form together but the route is not working for me 🤔
> > I'm navigating to: `/en/talent/skills-and-experiences/award/8e859a4b-b340-4afa-bef1-7feb456e8400/edit` and `/en/talent/skills-and-experiences/award/create` but getting a 404 on both.
> > Am I doing something obviously wrong?
> 
> have you injected the feature flag PHP var?

I believe so but i will check again, thanks for the suggestion. On another note,

Reviewing the code I wrote my suspicion lies with the sync portion of the [submissionData.ts](https://github.com/GCTC-NTGC/gc-digital-talent/blob/feature/1928-edit-experience-button-prefills-form/frontend/talentsearch/src/js/components/experienceForm/submissionData.ts):

```tsx
 if (data.skills) {
    skillSync = Object.keys(data.skills).map((key: string) => {
      if (!data.skills) {
        return undefined;
      }
      const skill = data.skills[key];
      return {
        id: key,
        details: skill.details,
      };
    });
  }

  return {
    details: data.details,
    skills: data.skills ? { sync: skillSync } : undefined,
    ...dataMap[type],
  };
```

It is supposed to remove any ids that do not exist in the array while adding those that do not. I actually had trouble with this directive on another one (cant remember where but will update when i find it).

JamesHuf:
The routes should currently be `/en/talent/profile/skills-and-experiences/award/8e859a4b-b340-4afa-bef1-7feb456e8400/edit` and  `/en/talent/profile/skills-and-experiences/award/create` although looking at the sitemap again I can see they're out of date as well

vd1992:
wait, why are you two navigating via URL to the forms?
I've been click `Edit Experience` on an experience in 
/en/talent/profile/skills-and-experiences

I don't trust the generated URLs because they vary by machine

JamesHuf:
I've generally been navigating via the edit links but Eric asked about the URLs so that's what I gave him.

esizer:
> wait, why are you two navigating via URL to the forms? I've been click `Edit Experience` on an experience in /en/talent/profile/skills-and-experiences
> 
> I don't trust the generated URLs because they vary by machine
> 
> Edit: Oh, maybe the /create URL is having issues

I'm just weird and navigate to everything via URLs rather than clicking on buttons 😆

esizer:
Okay, so now that I can view the form and inspect the network requests It is actually passing the deleted skill to the mutation which shifts the blame away from sync.

In the [ExperienceSkills.tsx](https://github.com/GCTC-NTGC/gc-digital-talent/blob/2efca484d524227536ac622c4ea5a021cd1b9134/frontend/talentsearch/src/js/components/experienceForm/ExperienceSkills.tsx#L56-L68) file we have:

```tsx
const { setValue, watch } = useForm();
...
  const handleRemoveSkill = (id: string) => {
    const newSkills = Object.keys(watchedSkills).reduce(
      (object: FormValues["skills"], key) => {
        if (key !== id) {
          // eslint-disable-next-line no-param-reassign
          object[key] = watchedSkills[key];
        }
        return object;
      },
      {},
    );
    setValue("skills", newSkills);
  };
```

When deleting a skill the `newSkills`  reduce function properly removed the skill but `setValue` doesn't seem to update the form data properly. I'm fairly new the `react-hook-form` and have much more experience with `formik` so I'm not really sure how this function works.

It's my understanding that calling this within the existing form context _should_ update the form data that ultimately gets submitted.

To confirm, we can add skills but not remove them, right?

vd1992:
yeah, as far as I can tell adding skills works just fine, you just can't remove them

esizer:
> yeah, as far as I can tell adding skills works just fine, you just can't remove them

Hm, okay that is odd because it uses the same function and as far as I can tell it is being provided the proper input. I'll keep looking.

JamesHuf:
> > yeah, as far as I can tell adding skills works just fine, you just can't remove them
> 
> Hm, okay that is odd because it uses the same function and as far as I can tell it is being provided the proper input. I'll keep looking.

From my understanding when I was looking around was that: watchedSkills and addedSkills are updating ok (which is why the skills disappear from the page) but something (and this is what I was stuck on) is re-adding the initial skills to the list when we submit the form.

petertgiles:
Sorry, I haven't figured it out yet, either.  It smells a little funny the `skills` prop into ExperienceSkills is the full possible list but that's also the same name used in form for selected skills.  But I'm not there yet.  :frowning_face:

vd1992:
> > yeah, as far as I can tell adding skills works just fine, you just can't remove them
> 
> Hm, okay that is odd because it uses the same function and as far as I can tell it is being provided the proper input. I'll keep looking.

I've been poking and prodding as well for fun. 
Yeah, I don't get it. Why does adding a skill work just fine, but removing it do nothing? The add and remove functions are pretty symmetrical, so I am confused why only one does not work.

I tried editing a skill description, clicking remove, and then submitting. What happened was that the skill was not deleted, but its description updated. So for whatever reason, deleting won't delete it. It merely hides it, and it updates with whatever the details field had before you "removed" it. 
Or maybe it does remove it, but the <FormValues> whatever thing just refusing to acknowledge removal as an operation.

petertgiles:
I've created bug #2693 for "can't remove skills".  Let's take that bug out of scope for this issue.

vd1992 (approved):
All other issues addressed, the remaining bug spun out into a separate issue
Success!

frontend/talentsearch/src/js/components/applicantProfile/ExperienceSection.tsx
```diff
@@ -0,0 +1,233 @@
...
+          ({
+            ...award,
+            startDate: award.awardedDate,
+            endDate: award.awardedDate,
+          } as AwardExperience & { startDate: string; endDate: string }),
+      )
+      .sort(compareByDate) || [];
+  const communityExperiences =
+    experiences
+      ?.filter((experience) => isCommunityExperience(experience))
```
tristan-orourke, line 113: The anonymous function is exactly equivalent to the named function, so we should just use the named function directly.
```suggestion
      ?.filter(isCommunityExperience)
```

frontend/talentsearch/src/js/components/applicantProfile/ExperienceSection.tsx
```diff
@@ -0,0 +1,233 @@
...
+      .sort(compareByDate) || [];
+  const educationExperiences =
+    experiences
+      ?.filter((experience) => isEducationExperience(experience))
```
tristan-orourke, line 117:
```suggestion
      ?.filter(isEducationExperience)
```

frontend/talentsearch/src/js/components/applicantProfile/ExperienceSection.tsx
```diff
@@ -0,0 +1,233 @@
...
+      .sort(compareByDate) || [];
+  const personalExperiences =
+    experiences
+      ?.filter((experience) => isPersonalExperience(experience))
```
tristan-orourke, line 121:
```suggestion
      ?.filter(isPersonalExperience)
```

tristan-orourke (changes requested):
Great job everyone on trying to identify and scope the issue with deleting skills.

My main remaining note is not to repeat the text of a link in the title attribute. Here are a few articles I read when trying to confirm that. One of them advices not to use the title attribute at all! But they all agree not to put redundant text there.
https://www.deque.com/blog/text-links-practices-screen-readers/
https://www.searchenginejournal.com/link-title-attribute/335415/
https://www.a11yproject.com/posts/creating-valid-and-accessible-links/

tristan-orourke (approved):

PR #2649, opened 2022-05-03
Add a template for creating new issues

patcon:
Just wanted to create a quick one. We can have a few, and this doesn't need to cover everything (a blank is always still an option).

Even if we create through ZenHub most of the time (and bypass GitHub's templates), this can be a file to copy-paste from for an idealized format.

Screen Shot 2022-05-03 at 6 11 41 PM

### DIFF
--- a/.github/issue_template.yml
+++ b/.github/issue_template.yml
@@ -4,12 +4,13 @@ about: Use this template new features with links to prototypes.
 # These aren't used at the moment.
 # See: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/manually-creating-a-single-issue-template-for-your-repository
 title: null
-labels: null
+labels: feature
 assignees: null
 ---
 <sup>ℹ️ [Figma (root)][figma] | [Sitemap][sitemap]</sup>
 <sup>_This initial comment is collaborative and open to modification by all._</sup>
 
+## Description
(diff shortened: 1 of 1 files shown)


.github/issue_template.yml
```diff
@@ -0,0 +1,25 @@
...
+about: Use this template new features with links to prototypes.
+# These aren't used at the moment.
+# See: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/manually-creating-a-single-issue-template-for-your-repository
+title: null
+labels: null
+assignees: null
+---
+<sup>ℹ️ [Figma (root)][figma] | [Sitemap][sitemap]</sup>
+<sup>_This initial comment is collaborative and open to modification by all._</sup>
+
```
tristan-orourke, line 12: Maybe add a `## Description` section to delineate from the `<sup>` section?

tristan-orourke (commented):
Yeah, this will be handy! I just left a couple suggestions. And the template can be used the same way in ZenHub, by the way.

tristan-orourke (changes requested):
@patcon I left a couple comments before, but maybe you didn't see them because I forgot to make it a official review

.github/issue_template.yml
patcon:
```suggestion

## Description
```

PR #2667, opened 2022-05-05
More DB constraints

### LINKED ISSUE #1847: Add missing Database integrity constraints
# Description
We could add some constraints to some tables to help ensure data integrity. For example, each user should only have a single PoolCandidate item in a given pool. Its worth taking a quick look for others.

petertgiles:
This branch adds some more database constraints.  Most of them are just making pivots unique but I also added one to make the user sub unique.

Testing tips:
Get into the maintenance container: `docker-compose run --rm maintenance bash`
Get to the api directory: `cd /var/www/html/api/`
Check your migration status: `php artisan migrate:status`
Start from scratch: `php artisan migrate:fresh --seed`
Rollback one step: `php artisan migrate:rollback --step=1`
Roll forward to the end: `php artisan migrate`

Closes #1847

### DIFF
--- a/api/database/migrations/2022_05_04_173542_add_extra_constraints.php
+++ b/api/database/migrations/2022_05_04_173542_add_extra_constraints.php
@@ -1,0 +1,125 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AddExtraConstraints extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        /** the first batch is a bunch of pivots to make unique **/
+
+        // a pool may not have duplicate CMO assets assigned to it
+        Schema::table('asset_cmo_asset_pool', function (Blueprint $table) {
+            $table->unique(['cmo_asset_id', 'pool_id'], 'asset_cmo_asset_pool_unique');
+        });
+
+        // a pool may not have duplicate classifications assigned to it
+        Schema::table('classification_pool', function (Blueprint $table) {
+            $table->unique(['classification_id', 'pool_id'], 'classification_pool_unique');
+        });
+
+        // a pool candidate may not have duplicate classifications assigned to it
+        Schema::table('classification_pool_candidate', function (Blueprint $table) {
+            $table->unique(['pool_candidate_id', 'classification_id'], 'classification_pool_candidate_unique');
+        });
+
+        // a user may not have duplicate classifications assigned to it
+        Schema::table('classification_user', function (Blueprint $table) {
+            $table->unique(['user_id', 'classification_id'], 'classification_user_unique');
+        });
+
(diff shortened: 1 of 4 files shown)


petertgiles:
Do we want to tighten up the classifications table?
- min salary not null
- max salary not null
- min salary >= 0
- max salary >= min salary

more?
- group - two uppercase characters
- level >= 0

even moar?
- make json arrays non-nullable with a default value of an empty array

api/database/migrations/2022_05_04_173542_add_extra_constraints.php
```diff
@@ -0,0 +1,140 @@
...
+     *
+     * @return void
+     */
+    public function up()
+    {
+        /** the first batch is a bunch of pivots to make unique **/
+
+        // a pool may not have duplicate CMO assets assigned to it
+        Schema::table('asset_cmo_asset_pool', function (Blueprint $table) {
+            $table->unique(['cmo_asset_id', 'pool_id'], 'asset_cmo_asset_pool_unique');
```
patcon, line 21: Does the order of constraints in these pairs matter to us?

I see that sometimes the parent entity is first in the constraint, and sometimes it's second. Not sure if that's intentional. If it is, maybe you could leave comment as to why? And if not, maybe we could keep it consistent to keep future behaviour more consistent? (e.g., partial indexes: https://stackoverflow.com/a/52619215)

patcon:
Ah hey, was looking at column order in the firs place because was wondering whether a utility function would make it more legible that the same thing was happening each time, and then noticed these variations that made it so a utility function wouldn't work: (code prob not working -- php is rusty)

```php
function addConstraintParentNoDupChildren($parentType, $childType, $tableName = NULL) {
  $tableName =  is_null($tableName) ? $childType.'_'.$parentType : $tableName
  Schema::table($tableName, function (Blueprint $table) {
    $table->unique([$childType.'_id', $parentType.'_id'], $childType.'_'.$parentType.'_unique');
  });
}

patcon (approved):
Good as-is, esp if questions don't give you any other ideas

patcon:
And I'm with tristan on not being overly clever with the constraints before the rubber has hit the road with "real world" usage :)

api/database/migrations/2022_05_04_173542_add_extra_constraints.php
petertgiles: Sorry, I think I may have mislead you with the description comment.  I'm not sure there's really a "parent" and a "child" in these relationships.  These are all pivot tables to support many-to-many relationships between two peer models.   My comment says "a pool has CMO assets" because that's the way a user would see it.  From a data modelling perspective it would be just as valid to say that "a CMO asset has pools" though it's not a useful use case and our API doesn't have any queries written to support that.

The idea of partial index usage is interesting.  These indexes weren't added for performance but in all these relationships there is one side that we would typically use for searching and filtering.  I just paired them in the order that the columns appear in the tables.
image
I'll switch them to have the "filterable" field first.  Thanks for the idea!

PR #2673, opened 2022-05-05
Bug - 2668 - Fix Missing Request Button

### LINKED ISSUE #2668: Bug - request button missing on Talent Search page

esizer:
## Summary

When changing the spinner to be more accessible and its own component, we moved the candidateResults to its own component to avoid nesting ternary operations. When this was done, we passed a reference of the function rather than calling it to have it render. 

This is a simple change to call the `candidateResults` function to actually render this component.

Resolves #2668

Screen Shot 2022-05-05 at 1 28 29 PM


petertgiles (approved):
Thanks for the quick fix!

PR #2674, opened 2022-05-05
Upgrade to Jest 28

### LINKED ISSUE #2627: refactor to upgrade Jest to v28
there are multiple breaking changes that appear to exist throughout multiple tests, meaning, multiple tests may need to be refactored to some extent, the online jest release notes addresses this

error:
`TypeError: Cannot read property 'html' of undefined`
possible resource:
https://jestjs.io/docs/upgrading-to-jest28#jsdom
occurs: 
seemingly in every single test

error
`Failed to collect coverage from __________ `
```
Invalid return value:
  `process()` or/and `processAsync()` method of code transformer found at 
  "/home/runner/work/gc-digital-talent/gc-digital-talent/frontend/node_modules/ts-jest/dist/index.js" 
  should return an object or a Promise resolving to an object. The object 
  must have `code` property with a string of processed code.
```
possible resource:
https://jestjs.io/docs/upgrading-to-jest28#transformer
occurs:
boatload of places that begin with `/home/runner/work`

petertgiles:
This branch upgrades the project to Jest 28.  It turns out the blocker was just that it needed the `jest-environment-jsdom` package dependency made explicit.  https://jestjs.io/docs/upgrading-to-jest28#jsdom
Closes #2627


vd1992:
well that ended up not nearly a daunting task 🤣

esizer (approved):
Works great! 👌 

One thing I noticed, which is out of scope for this PR, is that the test coverage is taking exceedling long (10+ mins) for the common project. Should we look into our jest config to collect coverage for less folders/files? Or is this only for me?

```tsx
  collectCoverageFrom: [
    "src/**/*.{js,jsx,ts,tsx}",
    "!<rootDir>/node_modules/",
  ],
```

petertgiles:
Confirming I see the long code coverage issue on my workstation, too.

esizer:
> Confirming I see the long code coverage issue on my workstation, too.

My suspicion is the ignore patter we have. I don't believe it is properly ignoring `node_modules` since the number of files we have in the project should not result in 10+ mins.

I see in the [documentation for jest](https://jestjs.io/docs/configuration#coveragepathignorepatterns-arraystring) that `node_modules` should be ignore by default though, so I could be wrong about this.

petertgiles:
I'm not sure.  I tried pointing the collection directory to src/components and it was slow and then pointed it to src/constants and it was fast.  I'm not sure where to go from here since I don't want to trial-and-error each component.  :man_shrugging:

esizer:
> I'm not sure. I tried pointing the collection directory to src/components and it was slow and then pointed it to src/constants and it was fast. I'm not sure where to go from here since I don't want to trial-and-error each component. 🤷‍♂️

Yeah, guess we can file an issue and just come back to it. Our tests are running and that is the important part 😜

Issue #2681, opened 2022-05-06
GraphQL error when searching for talent with CMO Assets


tristan-orourke:
The full error appeared in a toast, which is also not great.
This was found on UAT and PROD.

```
{
  "errors": [
    {
      "message": "Variable \"$where\" got invalid value {\"classifications\":[],\"cmoAssets\":[[],[],[],[]],\"operationalRequirements\":[\"SHIFT_WORK\"],\"hasDiploma\":false,\"hasDisability\":false,\"isIndigenous\":false,\"isVisibleMinority\":false,\"isWoman\":false,\"languageAbility\":\"FRENCH\",\"workRegions\":[],\"pools\":[{\"id\":\"85156e08-101b-4616-961d-2cb08ec2801e\"}]}; Field value.cmoAssets[0].key of required type KeyString! was not provided.",
      "extensions": {
        "category": "graphql"
      },
      "locations": [
        {
          "line": 1,
          "column": 27
        }
      ]
    },
    {
      "message": "Variable \"$where\" got invalid value {\"classifications\":[],\"cmoAssets\":[[],[],[],[]],\"operationalRequirements\":[\"SHIFT_WORK\"],\"hasDiploma\":false,\"hasDisability\":false,\"isIndigenous\":false,\"isVisibleMinority\":false,\"isWoman\":false,\"languageAbility\":\"FRENCH\",\"workRegions\":[],\"pools\":[{\"id\":\"85156e08-101b-4616-961d-2cb08ec2801e\"}]}; Field value.cmoAssets[1].key of required type KeyString! was not provided.",
      "extensions": {
        "category": "graphql"
      },
      "locations": [
        {
          "line": 1,
          "column": 27
        }
      ]
    },
    {
      "message": "Variable \"$where\" got invalid value {\"classifications\":[],\"cmoAssets\":[[],[],[],[]],\"operationalRequirements\":[\"SHIFT_WORK\"],\"hasDiploma\":false,\"hasDisability\":false,\"isIndigenous\":false,\"isVisibleMinority\":false,\"isWoman\":false,\"languageAbility\":\"FRENCH\",\"workRegions\":[],\"pools\":[{\"id\":\"85156e08-101b-4616-961d-2cb08ec2801e\"}]}; Field value.cmoAssets[2].key of required type KeyString! was not provided.",
      "extensions": {
        "category": "graphql"
      },
      "locations": [
        {
          "line": 1,
          "column": 27
        }
      ]
[...]

Issue #2256, opened 2022-03-15
Migrate to using use-debounce pkg for hooks for debouncing forms


patcon:
Re-ticketed from: https://github.com/GCTC-NTGC/gc-digital-talent/pull/2082/files#diff-30ab133bdf859b4d8f08fdd914329e912c392b01e818ff69f3714914258058caR144

We're doing debouncing with a function that useCallback linting rules are complaining about. We pass in a named function using `debounce` pkg, when we should apparently be passing an inline (anon) function to useCallback.

There's a very popular pkg that should help us with this, esp since it will be more commonly used in later forms: https://github.com/xnimorz/use-debounce

cc: @tristan-orourke 

## Acceptance Criteria
- [ ] replace our debounce logic with useDebounce hook
- [ ] remove "ignore" line for eslint rule `react-hooks/exhaustive-deps` [[?]](https://typeofnan.dev/you-probably-shouldnt-ignore-react-hooks-exhaustive-deps-warnings/)
- [ ] ...

gobyrne:
We think this has been addressed through work on linting errors.

PR #2694, opened 2022-05-09
Bug - 2682 - Classification causing graphql Error

### LINKED ISSUE #2682: GraphQL error when searching talent via Classification
Steps to reproduce:
1. Go to https://talent.canada.ca/en/talent/search
2. Select a classification
3. See noisy toast error
```
{
  "errors": [
    {
      "message": "Variable \"$where\" got invalid value {\"classifications\":[[],[]],\"cmoAssets\":[],\"operationalRequirements\":[\"WORK_WEEKENDS\",\"OVERTIME_SCHEDULED\"],\"hasDiploma\":false,\"hasDisability\":false,\"isIndigenous\":false,\"isVisibleMinority\":false,\"isWoman\":false,\"workRegions\":[\"NATIONAL_CAPITAL\",\"BRITISH_COLUMBIA\"],\"pools\":[{\"id\":\"92db3350-727b-429c-9c38-e8571cade8b1\"}]}; Field value.classifications[0].group of required type String! was not provided.",
      "extensions": {
        "category": "graphql"
      },
      "locations": [
        {
          "line": 1,
          "column": 27
        }
      ]
    },
    {
      "message": "Variable \"$where\" got invalid value {\"classifications\":[[],[]],\"cmoAssets\":[],\"operationalRequirements\":[\"WORK_WEEKENDS\",\"OVERTIME_SCHEDULED\"],\"hasDiploma\":false,\"hasDisability\":false,\"isIndigenous\":false,\"isVisibleMinority\":false,\"isWoman\":false,\"workRegions\":[\"NATIONAL_CAPITAL\",\"BRITISH_COLUMBIA\"],\"pools\":[{\"id\":\"92db3350-727b-429c-9c38-e8571cade8b1\"}]}; Field value.classifications[0].level of required type Int! was not provided.",
      "extensions": {
        "category": "graphql"
      },
      "locations": [
        {
          "line": 1,
          "column": 27
        }
      ]
    },
    {
      "message": "Variable \"$where\" got invalid value {\"classifications\":[[],[]],\"cmoAssets\":[],\"operationalRequirements\":[\"WORK_WEEKENDS\",\"OVERTIME_SCHEDULED\"],\"hasDiploma\":false,\"hasDisability\":false,\"isIndigenous\":false,\"isVisibleMinority\":false,\"isWoman\":false,\"workRegions\":[\"NATIONAL_CAPITAL\",\"BRITISH_COLUMBIA\"],\"pools\":[{\"id\":\"92db3350-727b-429c-9c38-e8571cade8b1\"}]}; Field value.classifications[1].group of required type String! was not provided.",
      "extensions": {
        "category": "graphql"
      },
      "locations": [
        {
          "line": 1,
          "column": 27
        }
      ]
    },
    {
      "message": "Variable \"$where\" got invalid value {\"classifications\":[[],[]],\"cmoAssets\":[],\"operationalRequirements\":[\"WORK_WEEKENDS\",\"OVERTIME_SCHEDULED\"],\"hasDiploma\":false,\"hasDisability\":false,\"isIndigenous\":false,\"isVisibleMinority\":false,\"isWoman\":false,\"workRegions\":[\"NATIONAL_CAPITAL\",\"BRITISH_COLUMBIA\"],\"pools\":[{\"id\":\"92db3350-727b-429c-9c38-e8571cade8b1\"}]}; Field value.classifications[1].level of required type Int! was not provided.",
      "extensions": {
        "category": "graphql"
      },
      "locations": [
        {
          "line": 1,
          "column": 27
        }
      ]
    }
  ]
}
```

esizer:
## Summary

It seems the `useRef` call was successful in preventing unnecessary re-renders, however that is due to the fact that it was caching submit function and the dependencies were not updating. This lead to the CMO and Classification maps being undefined.

~~I have moved the the defined functions to the `useDeepCompareEffect` to avoid the need for the `useRef` call as well as reducing the number of dependencies being passed into the hook. This has resolved the error, however it also required commenting out some functionality~~

Update: Modified to make use of the `watch` callback rather than the `useDeepCompareEffect` helper.

Resolves #2682 and #2681


tristan-orourke (commented):
You know how sometimes you look back on code you wrote a few months ago and just cringe? That's how I'm feeling about this debounce code. I'm leaning towards just removing it entirely. Instead, we could use the callback form of the [watch method](https://react-hook-form.com/api/useform/watch) to submit every time the form changes. That feels like the quickest and most reliable fix for this particular bug.

If we keep the debounce, we can probably make use of the callback watch method, something I didn't know about when first writing this. We can also consider using the debounce package mentioned here https://github.com/GCTC-NTGC/gc-digital-talent/issues/2256, which might help avoid some of the issues.

tristan-orourke:
I'd also like to refactor that use of LocationState for initial values, though it doesn't have to be in this high-priority PR. The fact that you had to hunt down where it was set doesn't speak well for this code 😅 . 

Passing intialValues into the SearchForm as a prop seems like a good start, it would give some more options for jest tests and storybook. 

Defaulting it to undefined instead of `{}` might help avoid some re-rendering.

Instead of using `location.state` at all, we could simply reuse our SessionStorage hook. That would accomplish the same thing, I think, with a more standard pattern.

esizer:
> You know how sometimes you look back on code you wrote a few months ago and just cringe? That's how I'm feeling about this debounce code. I'm leaning towards just removing it entirely. Instead, we could use the callback form of the [watch method](https://react-hook-form.com/api/useform/watch) to submit every time the form changes. That feels like the quickest and most reliable fix for this particular bug.
> 
> If we keep the debounce, we can probably make use of the callback watch method, something I didn't know about when first writing this. We can also consider using the debounce package mentioned here #2256, which might help avoid some of the issues.

I can take a stab at that if you'd like. Otherwise it may be quicker to release something if you have a plan to complete it if you wanna push something up to this branch.

I have a good idea of what you are talking about but if speed is the priority you may have a better handle on it instead of me spending 3/4 reading code to understand how to spend the other 1/4 writing it. But if this can wait a day then I am confident I could have something up by tomorrow for another review.

esizer:
> I'd also like to refactor that use of LocationState for initial values, though it doesn't have to be in this high-priority PR. The fact that you had to hunt down where it was set doesn't speak well for this code 😅 .
> 
> Passing intialValues into the SearchForm as a prop seems like a good start, it would give some more options for jest tests and storybook.
> 
> Defaulting it to undefined instead of `{}` might help avoid some re-rendering.
> 
> Instead of using `location.state` at all, we could simply reuse our SessionStorage hook. That would accomplish the same thing, I think, with a more standard pattern.

Assuming this actually works, do you think it would be appropriate to push it out as a fix and prepare a ticket for a more elegant fix/refactor in the next sprint?

I know what we have here is somewhat hacked in but from my testing it works and getting this out will give us more time to do a proper refactoring of the form using `watch`, `debounce` and `initialValues`.

@tristan-orourke

esizer:
@tristan-orourke Integrity check was introduced with 
#2696 but fixed in #2698 and can be merged in once it has been accepted.

frontend/talentsearch/src/js/components/search/SearchForm.tsx
```diff
@@ -99,14 +97,22 @@ export const SearchForm: React.FunctionComponent<SearchFormProps> = ({
...
+  const { watch } = methods;
+
+  React.useEffect(() => {
+    const formValuesToData = (values: FormValues): PoolCandidateFilterInput => {
       return {
         classifications: values.classifications
-          ? values.classifications?.map((id) => classificationMap.get(id))
+          ? values.classifications?.map((id) =>
+              id ? classificationMap.get(id) : null,
+            )
```
tristan-orourke, line 112: If there were a case where this returned null, I think it might break the api query. Maybe filter out nulls after? (Same for cmoAssets.)
```
          ? values.classifications
              ?.map((id) => (id ? classificationMap.get(id) : null))
              .filter(notEmpty)
```

tristan-orourke (approved):
Nice work! I've tested it as well. I suggest filtering out nulls, but you can merge after that minor change.

I'll create a new issue to refactor how we pass in initial values. We already have an issue for refactoring the debounce.

Issue #2637, opened 2022-05-03
Dashboard


tristan-orourke:

tristan-orourke:
Question: @Jerryescandon should it just show Pending requests?

Jerryescandon:
> Question: @Jerryescandon should it just show Pending requests?

Probably not, let's just stick with chronological order, regardless of status.

PR #2698, opened 2022-05-09
2638 - Create Dashboard Placeholder

### LINKED ISSUE #2638: Dashboard page placeholder
Placeholder page which shows up at correct url. Side menu link works.

esizer:
## Summary

Created the placeholder for the dashboard page, added the route and menu link. Name on the dashboard is a hardcoded string but ready to accept the current user name when #2637 is being worked on.

### Note

The text differs from the original text in the mock up slightly to avoid directional language.

"Don’t use directional terms as the only clue to location. Left, right, up, down, above, and below aren’t very useful for people who use screen-reading software. If you must use a directional term, provide additional text about the location, such as in the Save As dialog box, on the Standard toolbar, or in the title bar." - [Writing for all abilities](https://docs.microsoft.com/en-us/style-guide/accessibility/writing-all-abilities)

Ref: https://medium.com/docusign-design/a-ux-copywriters-guide-to-accessibility-79dbea85798d

Resolves #2638 

Screen Shot 2022-05-09 at 2 14 44 PM


petertgiles:
In my experience, when you are missing an integrity value or two you can get them resolved by a manual `npm i` of those packages.

esizer:
> actually fix integrity check

Worked like a charm!

Resolved with the aptly named [actually fix integrity check](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2698/commits/8924734fe38362657536c5889be4a8a14141cd90)

frontend/admin/src/js/components/PoolDashboard.tsx
```diff
@@ -46,6 +47,12 @@ const routes = (
       redirect: loggedIn ? paths.poolTable() : undefined,
```
tristan-orourke, line 47: Hey, we can change this to redirect to the dashboard now, like it was supposed to all along!
```suggestion
      redirect: loggedIn ? paths.dashboard() : undefined,
```

tristan-orourke (approved):
Looks good! Let's just change the default redirect.

PR #2701, opened 2022-05-09
Move docker compose

patcon:
Just pulling this out of #2217, as it's something I've wished was available already.

It should change anything, except we can now run docker-compose commands from anywhere in the project, instead of having to back out into `infrastructure/` to run them :)

Ready for review! If docker workflows pass, it should be all good.

### DIFF
--- a/.github/workflows/check-lighthouse-schema.yml
+++ b/.github/workflows/check-lighthouse-schema.yml
@@ -22,14 +22,12 @@ jobs:
         run: cp .env.example .env
 
       - name: "Build docker image: api"
-        working-directory: infrastructure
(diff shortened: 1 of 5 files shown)


patcon:
Checks all pass! Good for review/merge. (very small change)

patcon:
K done! Thanks peter

And to clarify, while people might appreciate knowing they no longer need to be in a specific docker directory, no current workflows (nor any scripts) will encounter issues with this, since Docker is simply working in _more_ places now :)

petertgiles (approved):
Nice.  Works on linux. :+1:

PR #2715, opened 2022-05-11
Test - 2669 - Search Container Tests

### LINKED ISSUE #2669: Regression Test - Jest test for the Talent Search component
Estimate numbers appear as expected, and then request button appears. If no results, no button.

esizer:
Resolves #2669 

## Summary

Adds tests to ensure the number of candidates shows up properly along with the displaying the request button when there are candidates available.

## Note

This also makes the `candidateFilter` an optional field to avoid needing to pass `candidateFilter={undefined}` as a prop.

### DIFF
--- a/frontend/talentsearch/src/js/components/search/SearchContainer.test.tsx
+++ b/frontend/talentsearch/src/js/components/search/SearchContainer.test.tsx
@@ -1,0 +1,59 @@
+/**
+ * @jest-environment jsdom
+ */
+import "@testing-library/jest-dom";
+import { screen } from "@testing-library/react";
+import React from "react";
+import { fakeClassifications, fakeCmoAssets } from "@common/fakeData";
+
+import { render } from "../../tests/testUtils";
+import { SearchContainer } from "./SearchContainer";
+import type { SearchContainerProps } from "./SearchContainer";
+
+const mockClassifications = fakeClassifications();
+const mockCmoAssets = fakeCmoAssets();
+
+type MockSearchContainerProps = Pick<SearchContainerProps, "candidateCount">;
+
+const renderSearchContainer = ({
+  candidateCount,
+}: MockSearchContainerProps) => {
+  const mockUpdate = jest.fn();
+  const mockValues = jest.fn();
+  const mockSubmit = jest.fn();
+  return (
+    <>
+      {render(
+        <SearchContainer
+          classifications={mockClassifications}
+          cmoAssets={mockCmoAssets}
+          candidateCount={candidateCount}
+          updateCandidateFilter={mockUpdate}
+          updateInitialValues={mockValues}
+          handleSubmit={mockSubmit}
+        />,
+      )}
+    </>
+  );
+};
+
+describe("SearchContainer", () => {
+  it("should render request button with candidates", async () => {
+    renderSearchContainer({ candidateCount: 10 });
+    await expect(
+      screen.getByRole("button", { name: /request candidates/i }),
+    ).toBeInTheDocument();
+  });
+
+  it("should not render request button with no candidates", async () => {
+    renderSearchContainer({ candidateCount: 0 });
+    await expect(
+      screen.queryByRole("button", { name: /request candidates/i }),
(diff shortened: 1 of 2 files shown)


frontend/talentsearch/src/js/components/search/SearchContainer.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+      )}
+    </>
+  );
+};
+
+describe("SearchContainer", () => {
+  it("should render request button with candidates", async () => {
+    renderSearchContainer({ candidateCount: 10 });
+    await expect(
+      screen.getByRole("button", { name: /request candidates/i }),
```
mnigh, line 44: Are we only writing tests for the English version of the site?

frontend/talentsearch/src/js/components/search/SearchContainer.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+  it("should not render request button with no candidates", async () => {
+    renderSearchContainer({ candidateCount: 0 });
+    await expect(
+      screen.queryByRole("button", { name: /request candidates/i }),
+    ).not.toBeInTheDocument();
+  });
+
+  it("should render number of candidates", async () => {
+    renderSearchContainer({ candidateCount: 10 });
+    expect(await screen.getAllByText(/10/i).length).toBeGreaterThanOrEqual(3);
```
mnigh, line 57: Pardon my ignorance in advance. I'm not sure I understand this test: what does `.toBeGreaterThanOrEqual(3);` mean, get all instances of _10_ and then check if its length is greater or equal to 3?

tristan-orourke: This makes me think this test is too indirect. I would think that we should find the three elements we expect to show the candidate count, and then ensure they each contain "10".

frontend/talentsearch/src/js/components/search/SearchContainer.test.tsx
esizer: Correct, the text in the button doesn't matter so much. If it were french we would just modify the query to look for "Demande de candidats" and it should still pass.

The text is just a way to query the screen for the button. A French test would look like:

```tsx
....
describe("SearchContainer", () => {
  it("should render request button with candidates", async () => {
    renderSearchContainer({ candidateCount: 10 });
    await expect(
      screen.getByRole("button", { name: /demande de candidats/i }),
    ).toBeInTheDocument();
  });
...
```

frontend/talentsearch/src/js/components/search/SearchContainer.test.tsx
esizer: Unfortunately, since we are placing them in spans which have no reliable way for querying them outside of  `byTestId` which is not recommended. Unless I'm missing a different query that would be more suitable here.

Ref: https://testing-library.com/docs/queries/bytestid

"In the spirit of [the guiding principles](https://testing-library.com/docs/guiding-principles), it is recommended to use this only after the other queries don't work for your use case. Using data-testid attributes do not resemble how your software is used and should be avoided if possible."

tristan-orourke: Maybe this is a place where getByTestId is worthwhile as a "last resort"?

tristan-orourke: Also, I've seen some disagreement on when/how testid should be used, so I started a slack discussion about it https://talent-cloud.slack.com/archives/C0259G6KXJ6/p1652299866870359

esizer: Alright, updated to use `data-testid` rather than selecting by text. As well, an additional test has been created to test that the proper value exists in each element.

Resolved by [update to use test id query](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2715/commits/5184184004e34653688d7526d4f78dd58229f833)

mnigh (approved):
👍

PR #2718, opened 2022-05-11
Auto create users

### LINKED ISSUE #2672: Automatically register people
If some comes from SiC with a valid SUB, just add add them as a new user to database.

petertgiles:
This branch adds the feature that users who visit the site with a valid token but don't yet exist in the database will have a new user created for them on-the-fly.

Suggested test methodology:
1) Switch your environment to Sign In Canada
2) Remove the seeded user row for your user (or disable it by clearing the sub value)
3) Log into the admin site
4) Verify that a new user was created with the correct sub value for your SiC ID
Note: New users are not admin so you still won't be able to use the admin site :smile: 
5) Run the test suite

I factored out most of the logic of `AuthServiceProvider` to the `resolveUserOrAbort` method to make it easier to test.  I was never able to figure out how to bootstrap the app with a mock `BearerTokenService` to be able to create a more end-to-end test, unfortunately.

I had to create a DataSetInterface to create test doubles for the JWT DataSet which is a final class.  Unfortunately since the real DataSet does not actually implement my DataSetInterface I had to remove the type hinting where that class was used.  If any PHP wizards have a way around that let me know.  PHP 8 (cough, cough) has type hinting with unions which would be nice.

Closes #2672

### DIFF
--- a/api/app/Providers/AuthServiceProvider.php
+++ b/api/app/Providers/AuthServiceProvider.php
@@ -100,7 +100,6 @@ class AuthServiceProvider extends ServiceProvider
                 // No user found for given subscriber - lets auto-register them
                 $newUser = new User;
                 $newUser->first_name = $sub;  // displayed on the landing page so should help us find the user
-                $newUser->email = $sub.'@new.user'; // junk, but email is currently non-nullable
                 $newUser->sub = $sub;
                 $newUser->roles = null;
                 $newUser->save();
--- a/api/app/Providers/AuthServiceProvider.php
+++ b/api/app/Providers/AuthServiceProvider.php
@@ -65,41 +65,57 @@ class AuthServiceProvider extends ServiceProvider
         Gate::policy(AwardExperience::class, ExperiencePolicy::class);
 
         $this->app['auth']->viaRequest('api', function ($request) use ($tokenService) {
-
-            $bearerToken = $request->bearerToken(); // 1. extract JWT access token from request.
-            if($bearerToken)
-            {
-                try {
-                    $claims = $tokenService->validateAndGetClaims($bearerToken);  // 2. validate access token.
-                    $sub = $claims->get('sub');
-                } catch (RequiredConstraintsViolated $e) {
-                    $violations = [];
-                    foreach ($e->violations() as $violationError) {
-                        array_push($violations, $violationError->getMessage());
-                    }
-                    Log::notice('Authorization token not valid: ', $violations);
-                    return abort(401, 'Authorization token not valid.');
-                } catch (Exception $e) {
-                    Log::notice('Error while validating authorization token: ' . $e->getMessage());
-                    return abort(401, 'Error while validating authorization token.');
-                }
-
-                $userMatch = User::where('sub', $sub)->first(); // 3. match "sub" claim to user 'sub' field.
-                if($userMatch) {
-                    return $userMatch;
-                } else {
-                    Log::notice('No user found for given subscriber: '.$sub);
-                    return null;
-                }
-            }
-
-            // If a default user is set, and there is no bearer token, treat the request as if it were from the default user.
-            if (config('oauth.default_user')) {
-                $user = User::where('email', config('oauth.default_user'))->first();
-                if ($user) {
-                    return $user;
-                }
-            }
+            return $this->resolveUserOrAbort($request->bearerToken(), $tokenService);
         });
+    }
+
+    /**
+     * Get the associated user model for a request or abort if something goes wrong with the lookup
+     *
+     */
+    public function resolveUserOrAbort($bearerToken, $tokenService): ?User {
+        if($bearerToken)
+        {
+            try {
+                $claims = $tokenService->validateAndGetClaims($bearerToken);  // 2. validate access token.
+                $sub = $claims->get('sub');
+            } catch (RequiredConstraintsViolated $e) {
+                $violations = [];
+                foreach ($e->violations() as $violationError) {
+                    array_push($violations, $violationError->getMessage());
+                }
+                Log::notice('Authorization token not valid: ', $violations);
+                return abort(401, 'Authorization token not valid.');
+            } catch (Exception $e) {
+                Log::notice('Error while validating authorization token: ' . $e->getMessage());
+                return abort(401, 'Error while validating authorization token.');
+            }
+
+            // By this point we have verified that the token is legitimate
+
+            $userMatch = User::where('sub', $sub)->first(); // 3. match "sub" claim to user 'sub' field.
+            if($userMatch) {
+                return $userMatch;
+            } else {
+                // No user found for given subscriber - lets auto-register them
+                $newUser = new User;
+                $newUser->first_name = $sub;  // displayed on the landing page so should help us find the user
+                $newUser->email = $sub.'@new.user'; // junk, but email is currently non-nullable
+                $newUser->sub = $sub;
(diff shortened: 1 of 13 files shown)


api/app/Services/LocalAuthBearerTokenService.php
```diff
@@ -33,7 +32,10 @@ function __construct(string $issuer, string $publicKey, Clock $clock, DateInterv
         $this->allowableClockSkew = $allowableClockSkew;
     }
 
-    public function validateAndGetClaims(string $bearerToken) : DataSet
+    /*
+    * @returns DataSet
```
tristan-orourke, line 36:
```suggestion
    * @returns Lcobucci\JWT\Token\DataSet
```

api/app/Services/OpenIdBearerTokenService.php
```diff
@@ -141,7 +140,10 @@ public function verifyJwtWithIntrospection(string $accessToken)
         }
     }
 
-    public function validateAndGetClaims(string $bearerToken) : DataSet
+    /*
+    * @returns DataSet
```
tristan-orourke, line 144:
```suggestion
    * @returns Lcobucci\JWT\Token\DataSet
```

api/app/Providers/AuthServiceProvider.php
```diff
@@ -65,41 +65,57 @@ public function boot(BearerTokenServiceInterface $tokenService)
...
+            // By this point we have verified that the token is legitimate
+
+            $userMatch = User::where('sub', $sub)->first(); // 3. match "sub" claim to user 'sub' field.
+            if($userMatch) {
+                return $userMatch;
+            } else {
+                // No user found for given subscriber - lets auto-register them
+                $newUser = new User;
+                $newUser->first_name = $sub;  // displayed on the landing page so should help us find the user
+                $newUser->email = $sub.'@new.user'; // junk, but email is currently non-nullable
```
tristan-orourke, line 103: What do you think about changing our model so that email *is* nullable. It's not really a primary identifier anymore (that's sub, or id), its just contact info now. I like that if it is initialized as null, we can know to direct users to fill it in. 

We could add logic to migrations to disallow changing it *to* a null value, if we want.

api/app/Providers/AuthServiceProvider.php
```diff
@@ -65,41 +65,57 @@ public function boot(BearerTokenServiceInterface $tokenService)
...
-                }
+            // By this point we have verified that the token is legitimate
+
+            $userMatch = User::where('sub', $sub)->first(); // 3. match "sub" claim to user 'sub' field.
+            if($userMatch) {
+                return $userMatch;
+            } else {
+                // No user found for given subscriber - lets auto-register them
+                $newUser = new User;
+                $newUser->first_name = $sub;  // displayed on the landing page so should help us find the user
```
tristan-orourke, line 102: Smart

tristan-orourke (changes requested):
I'm going to treat this as a lesson in using mocks in phpunit tests!

I'm a bit worried about setting a dummy email. The [plan](https://www.figma.com/file/kNHblb2NTAovHAOvmEILpd/User-flow---Registration-and-login?node-id=0%3A1) is to direct users to fill in their name and email the first time they log in. That would be easier if we could just check for a null email. I think setting it to nullable makes sense.

api/app/Providers/AuthServiceProvider.php
petertgiles: Yeah, definitely.  I'm on team "let's get rid of Laravel auth".  It doesn't really behave like SiC in a lot of ways and this is one of them. It's a very complicated project for what is essentially a poor mock or test double.  

I've added a migration to make email and sub both nullable.  I did sub since it's reasonable that many of our (non-admin) users will not have SiC subs to add.  I've tried to make the changes to the schema and app as light as possible for now.  I'll add an issue to the board to finish that job.

petertgiles: #2722

frontend/admin/src/js/components/poolCandidate/CreatePoolCandidate.tsx
```diff
@@ -255,7 +255,7 @@ export const CreatePoolCandidateForm: React.FunctionComponent<CreatePoolCandidat
         case "new":
           userObject = {
             create: {
-              email: values.email,
+              email: values.email ?? "",
```
tristan-orourke, line 258: What's the purpose behind this change?

frontend/admin/src/js/components/user/UpdateUser.tsx
```diff
@@ -155,7 +155,7 @@ export const UpdateUserForm: React.FunctionComponent<UpdateUserFormProps> = ({
               })}
               type="email"
               name="email"
-              value={initialUser.email}
+              value={initialUser.email ?? ""}
```
tristan-orourke, line 158: What's the purpose behind this change as well?

frontend/admin/src/js/components/poolCandidate/CreatePoolCandidate.tsx
petertgiles: This is because email is now nullable in the inital data but is still non-nullable for CreateUserInput.

frontend/admin/src/js/components/user/UpdateUser.tsx
petertgiles: This is a little weird because this is just for display, it's not submitted in the form.  I'm surprised but Maybe<string> (since email is nullable) cannot be assigned as undefined and Input.value can't be null.  This could be `initialUser.email ?? undefined` but I'm not sure that's any clearer.

frontend/admin/src/js/components/poolCandidate/CreatePoolCandidate.tsx
tristan-orourke: Ah, I see, its to get around type shenanigans :joy:

tristan-orourke: May I recommend adjusting the FormValues type to better reflect the form, instead of correcting for it here?
```
type FormValues = Pick<
  PoolCandidate,
  | "cmoIdentifier"
  | "expiryDate"
  | "hasDiploma"
  | "hasDisability"
  | "isIndigenous"
  | "isVisibleMinority"
  | "isWoman"
  | "languageAbility"
  | "expectedSalary"
  | "locationPreferences"
  | "acceptedOperationalRequirements"
  | "status"
> &
  Pick<User, "firstName" | "lastName" | "preferredLang" | "telephone"> & {
    email: string; // email is Maybe<string> in User definition, but is required in the form.
    cmoAssets: string[] | undefined;
    expectedClassifications: string[] | undefined;
    pool: string;
    user: string;
    userMode: UserModeValue;
  };
  ```

tristan-orourke (changes requested):

frontend/admin/src/js/components/user/UpdateUser.tsx
tristan-orourke: Got it, thanks for the explanation!

frontend/admin/src/js/components/poolCandidate/CreatePoolCandidate.tsx
petertgiles: Yeah, this is better, thanks.

tristan-orourke (approved):

Issue #1480, opened 2021-12-16
API - Class TestUser fails all tests


mnigh:
When running the PHPUnit tests within the class TestUser, all three of the tests result in failures: "No authorization token found.".

### Steps to reproduce
1. `cd infrastructure`
2. `docker-compose exec -w /var/www/html/api php sh -c "./vendor/bin/phpunit tests/UserTest.php"`

mnigh:
We would first need to generate an access token. Then, we would need to use the [postGraphQL function](https://github.com/nuwave/lighthouse/blob/master/src/Testing/MakesGraphQLRequests.php#L65-L83) instead of the graphQL function to be able to pass the access token (Bearer token) via a request header.

petertgiles:
@tristan-orourke Should this have been closed?  The tests are still broken.

tristan-orourke:
If the tests are still broken, I'm not sure why it was closed :/
I'll reopen it

mnigh:
> If the tests are still broken, I'm not sure why it was closed :/ I'll reopen it

It looks like the tests were [marked to be skipped](https://github.com/GCTC-NTGC/gc-digital-talent/pull/1993/files#diff-bc714e73f0e228fef8e2aba3bf3ed22a0ac1cc80d1b2f480913cf775945ddfff) in #1993.

PR #2720, opened 2022-05-11
Paginated users query

### LINKED ISSUE #2645: Create paginated users query (which accepts search filter)

JamesHuf:
Resolves #2645 

Created the usersPaginated query which can accept UserFilterInput and returns a paginated list of matching users.
Filters work for the following:
- pools: Returns users in ANY of the given pools
- languageAbility
- expectedClassifications: Returns users with ANY matching classifications OR salary expectations
- operationalRequirements: Only returns users accepting ALL of the requirements
- locationPreferences: Returns users with ANY matching locations
- hasDiploma
- wouldAcceptTemporary
- jobLookingStatus
- isProfileComplete
- skills: Returns users with ALL of the given skills
- isGovEmployee

Big thanks to @petertgiles for helping with the skills filter.


api/app/Models/User.php
```diff
@@ -10,6 +11,9 @@
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 use Illuminate\Contracts\Auth\Authenticatable;
 use Illuminate\Auth\Authenticatable as AuthenticatableTrait;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Support\Facades\Log;
```
esizer, line 15: Looks like we have an unused import

esizer (changes requested):
Looking really good! I tested it locally and got back exactly what I expected. Only noticed a simple unused import.

I might also suggested someone more familiar with the backend take a look as well. Looks good to me but I don't trust myself quite yet on the backend of this app 😅 

Also, this is up for discussion but I find lighthouses terminology "first" a bit confusing. I get what it does but seems like they chose an odd word to represent the number of items you are querying for. Is it just me?

Ref: https://lighthouse-php.com/5/api-reference/directives.html#paginate

```graphql
type Query {
  posts(
    "Limits number of fetched items."
    first: Int!

    "The offset from which items are returned."
    page: Int
  ): PostPaginator
}
```

vd1992:
is the lighthouse scheme build step missing? because upon checking out -> refresh API, I get this thing it wants to commit

image

petertgiles:
> I get this thing it wants to commit

Yeah, me too.  I wonder why the check schema github action didn't flag this?

JamesHuf:
> is the lighthouse scheme build step missing? because upon checking out -> refresh API, I get this thing it wants to commit
> 

Yeah, I totally forgot that step since graphql playground doesn't need it.

vd1992:
> > I get this thing it wants to commit
> 
> Yeah, me too. I wonder why the check schema github action didn't flag this?

broken test? 😬

petertgiles (changes requested):
Wow, this is looking super nice.  It's going to take me some time to do some manual tests.  In the meantime, could you add a few PHP tests?  I'd love to see a few for pagination and search by skill especially.  Yoni wrote some nice ones for PoolCandidate that you could use as a model.

api/app/Models/User.php
```diff
@@ -159,28 +187,238 @@ protected function generateSubject()
...
+        $query->where('job_looking_status', $jobLookingStatus);
+        return $query;
+    }
+    public function filterBySkills(Builder $query, array $skills): Builder
+    {
+        if (empty($skills)) {
+            return $query;
+        }
+
+        // skills act as an AND filter. The query should only return candidates with ALL of the skills.
```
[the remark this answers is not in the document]
JamesHuf: Thanks. I have one line (257) I had to leave as a selectRaw because I don't think laravel has a function for it but I got everything else.

api/storage/app/lighthouse-schema.graphql
```diff
@@ -770,6 +770,15 @@ type Query {
   skillFamilies: [SkillFamily]!
   skill(id: ID!): Skill
   skills: [Skill]!
+  usersPaginated(
+    where: UserFilterInput
+
+    """Limits number of fetched items."""
+    first: Int!
```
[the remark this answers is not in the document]
JamesHuf: It is possible to set a default page size but I wasn't sure what we'd prefer so I'd left it.

I don't like the name `first` either but I can't seem to find a way of renaming it.

petertgiles:
Pagination without any sorting is a bit weird.  There's really no guarantee from the database that you won't get back a different random set of items on your page for any given query if there's no sorting.  We're not using a database cursor queue our rows into a page, right?  I hate to blow scope but how hard would it be to add *something* for sorting?  Maybe default to sorting by the primary key so our results are at least deterministic?

petertgiles:
I checked pagination, language ability, and skills and it all looks great.  Once there are a few tests and we've decided what to do about sorting I'm happy to approve.

vd1992:
> Pagination without any sorting is a bit weird. There's really no guarantee from the database that you won't get back a different random set of items on your page for any given query if there's no sorting. We're not using a database cursor queue our rows into a page, right? I hate to blow scope but how hard would it be to add _something_ for sorting? Maybe default to sorting by the primary key so our results are at least deterministic?

yeah, it would be really handy to know how the results are ordered
it might be an additional argument to pass in, not sure

rerunning the query doesn't seem to return a random order so some order is being followed

api/storage/app/lighthouse-schema.graphql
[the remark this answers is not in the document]
petertgiles: OK, if it's not simple to change the name then let's leave it.  If you can set a default pages size let's do it.  I vote for 10. :laughing:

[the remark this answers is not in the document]
JamesHuf: > > Pagination without any sorting is a bit weird. There's really no guarantee from the database that you won't get back a different random set of items on your page for any given query if there's no sorting. We're not using a database cursor queue our rows into a page, right? I hate to blow scope but how hard would it be to add _something_ for sorting? Maybe default to sorting by the primary key so our results are at least deterministic?
> 
> yeah, it would be really handy to know how the results are ordered it might be an additional argument to pass in, not sure
> 
> rerunning the query doesn't seem to return a random order so some order is being followed

Yeah, I think it currently returns them in the order they are in the database. I'll change it though.

JamesHuf:
So I've set the default page size to 10 and added ordering to the response. Tests won't work right now though because of #1480. Do we want to prioritize that first or just skip tests for now?

petertgiles:
> Do we want to prioritize that first or just skip tests for now?

I think that won't prevent you from writing tests.  It's true that a few are marked to skip right now but you can still write new ones.
image

petertgiles:
> So I've set the default page size to 10 and added ordering to the response.

Ruh roh, it seems broken to me now.  No matter the query it's just returning everyone now.
image
When I log the SQL it's just doing a plain `select * from "users" where "users"."deleted_at" is null order by "id" asc limit 10 offset 0  ` every time.  Is this happening for you?

JamesHuf:
> > So I've set the default page size to 10 and added ordering to the response.
> 
> Ruh roh, it seems broken to me now. No matter the query it's just returning everyone now. When I log the SQL it's just doing a plain `select * from "users" where "users"."deleted_at" is null order by "id" asc limit 10 offset 0 ` every time. Is this happening for you?

Weird I'm not seeing that.
image

api/app/Models/User.php
```diff
@@ -159,28 +186,244 @@ protected function generateSubject()
...
+        $query->whereExists(function ($query) use ($skills) {
```
tristan-orourke, line 254: I'm impressed! This looks like it took some time to figure out :P It would take me some to just read through it and understand how it works.
Out of curiousity - am I correct that this only works because our ids are UUIDs? If they were non-unique integers, matching skills to experiences (this line `$join->on('experience_skills.experience_id', '=', 'experiences.experience_id');`) would be more complicated?

tristan-orourke:
Okay, another question...
The orderBy field works as intended. But I don't understand how? 😆 I would have expected the orderBy field in UserFilterAndOrderInput to require a `@scope(name: "order")` directive. Can you explain why this works without it?

api/graphql/schema.graphql
```diff
@@ -527,10 +527,26 @@ input PoolCandidateFilterInput {
...
+    skills: [ID] @builder(method: "App\\Models\\User@filterBySkills")
+    isGovEmployee: Boolean @scope
+    orderBy: String   # This needs to be the name of the field in the database eg. first_name
+}
+
 type Query {
     me: User @auth
     user(id: ID! @eq): User @find @guard @can(ability: "view", query: true)
     users: [User]! @all @guard @can(ability: "viewAny")
+    usersPaginated(where: UserFilterAndOrderInput): [User]! @paginate(defaultCount: 10, scopes: ["order"]) @guard @can(ability: "viewAny")
```
tristan-orourke, line 549: Should we add a maxCount as well? It could be pretty high. 500? 1000?

JamesHuf:
> Okay, another question for my own benefit... The orderBy field works as intended. But I don't understand how? 😆 I would have expected the orderBy field in UserFilterAndOrderInput to require a `@scope(name: "order")` directive. Can you explain why this works without it?

That one's interesting. Ultimately we don't need it because it's already being called by @paginate's scope field (which allows us to use default ordering even when orderBy is unspecified). I still need orderBy in UserFilterAndOrderInput though so people can use different ordering if needed.

api/graphql/schema.graphql
JamesHuf: Sounds good, I'll go with 1000 for now.

vd1992:
@JamesHuf 
Have you tried merging main and rebuilding your environment? I think there might have been a change of note or two with the Docker changes and it could merit attention

JamesHuf:
> @JamesHuf Have you tried merging main and rebuilding your environment? I think there might have been a change of note or two with the Docker changes and it could merit attention

I haven't yet but I will now. Good call!

JamesHuf:
> Could you add a few tests to exercise paging and sorting?

Yep, they're on the list.

vd1992:
sorting appears to work just fine
that should really come in handy, when you want to create a default view you can then pass in a value to sort by it

agreeing with peter on the tests, lets try to avoid more oopsies bubbling out that we catch down the line

vd1992:
> So I've set the default page size to 10 and added ordering to the response. Tests won't work right now though because of #1480. Do we want to prioritize that first or just skip tests for now?

In light of the above, I think it is a good idea to navigate to the test output and double check every time. Ensure no tests are skipped that are crucial or any other funny business. 
Looking at it, nothing in issue scope seems to have been skipped or thrown a tantrum. 

image

JamesHuf:
I've added some more tests. I might still add one for pagination but I'm not sure exactly what it should involve since the fact that most of the other tests are reading paginatorInfo means it's working.

petertgiles (commented):
Well, stymied again. 😄 It runs on the dev server but there's only two users there, not the full seeder, so I can't really test any filters.  Oh well, the pagination seems to work!  The tests look great!  Just one question on the test below.

api/app/Models/User.php
```diff
@@ -159,4 +151,265 @@ public function getIsProfileCompleteAttribute(): bool
...
+                    $join->on('experience_skills.experience_id', '=', 'experiences.experience_id');
+                })
+                ->groupBy('experiences.user_id');
+            }, "aggregate_experiences")
+            ->whereJsonContains('aggregate_experiences.user_skills_grouped', $skills)
+            ->whereColumn('aggregate_experiences.user_id', 'users.id');
+        });
+        return $query;
+    }
+    public function filterByClassifications(Builder $query, array $classifications): Builder
```
petertgiles, line 274: There was a small (but important!) fix to the filterByClassifications in the CandidatePool yesterday in https://github.com/GCTC-NTGC/gc-digital-talent/pull/2805 .  Could you please backport that fix here?

api/tests/Feature/UserTest.php
```diff
@@ -1031,4 +1037,421 @@ public function testFilterByJobLookingStatus(): void
...
+            'data' => [
+                'usersPaginated' => [
+                    'paginatorInfo' => [
+                        'total' => 3
+                    ]
+                ]
+            ]
+        ]);
+
+        // Assert query with isProfileComplete filter set to false will return all users
```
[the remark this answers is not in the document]
JamesHuf: The other booleans also work this way since that's how they work in the candidate count query and I mostly just ported them over. Do you think I should change them as well?

api/app/Models/User.php
JamesHuf: Will do.

api/tests/Feature/UserTest.php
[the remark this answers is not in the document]
petertgiles: Oh, you're right!  I feel like I've had this discussion before. deja-vu...
No, don't change it.  Consistency is much more important.  See, the test is good for showing expected behaviour, too.

api/app/Models/User.php
JamesHuf: I'd actually already fixed this but I added the tests and changed my method to match the one in #2805 since it feels clearer anyway.

petertgiles (approved):
Haha, I feel like I'm turning into the nightmare reviewer on this.  "Just one more thing..."  :laughing:   I love this, great work!

esizer (approved):

PR #2721, opened 2022-05-12
1933 toast notification when completing last required form2

### LINKED ISSUE #1933: Toast notification when completing last required form
# Description
We want to listen on the current state of application profile, and when it goes from being incomplete to being complete we want to display a toast notification saying as much.

We don't know which page the user will be on when this happens, so we want the toast to work consistently whether they're on the main profile page or any of the forms. In fact, the toast itself should persist when navigating between them.

Remember, triggering a toast is easy. Ensure our <Toast> component is somewhere in your component hierarchy (probably near the root, shared between the entire SPA), and then call:
```javascript
import { toast } from "react-toastify";
toast.success("my message");
```

Where and how to listen for the application status change is left to the implementer.

# Acceptance Criteria
- [ ] No storybook necessary
- [ ] Jest tests
  - [ ]  ???

RVany:
Resolves #1933 

# Description
We want to listen on the current state of application profile, and when it goes from being incomplete to being complete we want to display a toast notification saying as much.

We don't know which page the user will be on when this happens, so we want the toast to work consistently whether they're on the main profile page or any of the forms. In fact, the toast itself should persist when navigating between them.

Remember, triggering a toast is easy. Ensure our <Toast> component is somewhere in your component hierarchy (probably near the root, shared between the entire SPA), and then call:
```javascript
import { toast } from "react-toastify";
toast.success("my message");
```

Where and how to listen for the application status change is left to the implementer.

# Acceptance Criteria
- [ ] No storybook necessary
- [ ] Jest tests
  - [ ]  ???


RVany:
image
I don't think I made these changes😮. Any other reason?

petertgiles:
> I don't think I made these changesopen_mouth

Oh, it's a bad compare.  If you check the line numbers you can see it's comparing very different sections of the file.  If you do the compare in VS Code it does a much better job.

RVany:
> > I don't think I made these changesopen_mouth
> 
> Oh, it's a bad compare. If you check the line numbers you can see it's comparing very different sections of the file. If you do the compare in VS Code it does a much better job.

Ok, Good. Thank you!

frontend/talentsearch/src/js/components/GovernmentInfoForm/GovernmentInfoForm.tsx
```diff
@@ -434,7 +431,31 @@ export const GovInfoFormContainer: React.FunctionComponent = () => {
       return;
     }
     await handleUpdateUser(meId, data)
-      .then(() => {
+      .then((result) => {
+        if (result.isProfileComplete) {
```
petertgiles, line 435: This code block is repeated exactly on each page.  Is there any way to factor this out to a single location to be more DRY?

frontend/talentsearch/src/js/components/GovernmentInfoForm/GovernmentInfoForm.tsx
```diff
@@ -434,7 +431,31 @@ export const GovInfoFormContainer: React.FunctionComponent = () => {
...
+          if (result.isProfileComplete) {
+            sessionStorage.setItem("currentProfileStatus", "Complete");
```
petertgiles, line 437: I suggest using the session storage functions from frontend/common/src/helpers/storageUtils.ts instead of sessionStorage directly.  Just in case we every decide to do something else with session storage.

frontend/talentsearch/src/js/components/GovernmentInfoForm/GovernmentInfoForm.tsx
```diff
@@ -434,7 +431,31 @@ export const GovInfoFormContainer: React.FunctionComponent = () => {
...
+          } else {
+            sessionStorage.setItem("currentProfileStatus", "InComplete");
+          }
+          const preProfileStatus = sessionStorage.getItem("preProfileStatus");
+          const currentProfileStatus = sessionStorage.getItem(
+            "currentProfileStatus",
+          );
+
+          if (preProfileStatus === "InComplete") {
```
petertgiles, line 446: Could you constrain these values with enums or constants or something?  Also with your storage keys.

frontend/talentsearch/src/js/components/profile/ProfilePage/ProfilePage.tsx
```diff
@@ -76,12 +76,19 @@ export const ProfileForm: React.FC<ProfilePageProps> = ({
...
     experiences,
+    isProfileComplete,
   } = profileDataInput;
 
   const intl = useIntl();
   const paths = useApplicantProfileRoutes();
   const locale = getLocale(intl);
 
+  if (isProfileComplete) {
+    sessionStorage.setItem("preProfileStatus", "Complete");
```
petertgiles, line 87: Some concerns as other pages - use helper for storage and constrain strings.

petertgiles (changes requested):
Wow, it works - nice job!  I've suggested a few polish things below but you got it working very quickly!

frontend/talentsearch/src/js/components/GovernmentInfoForm/GovernmentInfoForm.tsx
```diff
@@ -434,7 +431,31 @@ export const GovInfoFormContainer: React.FunctionComponent = () => {
...
+                intl.formatMessage({
+                  defaultMessage:
+                    "All required fields are complete. You can now change your status.",
+                  description:
+                    "Message displayed to user when user profile completed.",
+                }),
+              );
+            }
+          }
+        }
```
petertgiles, line 458: Could you add some jest tests, too?  Maybe:
- If the page is changed and the profile goes from incomplete to complete a toast pops
- If the page is changed the profile stays incomplete then no toast
- If the page is changed and the profile stays complete then no toast

frontend/talentsearch/src/js/components/GovernmentInfoForm/GovernmentInfoForm.tsx
RVany: It makes sense. I will try.

frontend/talentsearch/src/js/components/GovernmentInfoForm/GovernmentInfoForm.tsx
RVany: It is inside the handler function. Do you think its possible to call a function here?

frontend/talentsearch/src/js/components/profile/ProfilePage/ProfilePage.tsx
tristan-orourke: One of the things the helpers do is serialize and deserialize the stored values, which allows you to store any kind of value, not just strings. This would let you store the boolean value directly: 
`setInSessionStorage("isProfileComplete", isProfileComplete)`

frontend/talentsearch/src/js/components/workPreferencesForm/WorkPreferencesForm.tsx
```diff
@@ -201,6 +201,30 @@ export const WorkPreferencesApi: React.FunctionComponent = () => {
       id,
       user: data,
     }).then((result) => {
+      if (result.data?.updateUserAsUser) {
+        if (result.data?.updateUserAsUser?.isProfileComplete) {
+          sessionStorage.setItem("currentProfileStatus", "Complete");
+        } else {
+          sessionStorage.setItem("currentProfileStatus", "InComplete");
+        }
```
tristan-orourke, line 209: Is there any reason to store the new status in sessionStorage as a different value? You can compare the current value to the previous one without storing the current value.

frontend/talentsearch/src/js/components/workPreferencesForm/WorkPreferencesForm.tsx
```diff
@@ -201,6 +201,30 @@ export const WorkPreferencesApi: React.FunctionComponent = () => {
...
+        const preProfileStatus = sessionStorage.getItem("preProfileStatus");
+        const currentProfileStatus = sessionStorage.getItem(
+          "currentProfileStatus",
+        );
+
+        if (preProfileStatus === "InComplete") {
+          if (currentProfileStatus === "Complete") {
```
tristan-orourke, line 216: After comparing the prev and current values, you should update the prev value to be the new value.

frontend/talentsearch/src/js/components/profile/ProfilePage/ProfilePage.tsx
RVany: > One of the things the helpers do is serialize and deserialize the stored values, which allows you to store any kind of value, not just strings. This would let you store the boolean value directly: `setInSessionStorage("isProfileComplete", isProfileComplete)`

Ok! Thank you!

tristan-orourke (changes requested):
Nice work figuring out a solution!
However, I think I see room for some edge-case bugs. If someone navigated directly to one of the form pages with the url, without going through the Profile page, the preProfileStatus won't be set. We should check and save the preProfileStatus on each page, not just the profile.

tristan-orourke:
Actually, if you add isProfileComplete to the initial get request on each page, you can do this without using session storage at all! For example, this could replace the `handleWorkPreferences` function in WorkPreferencesForm:
```
const handleWorkPreferences = (id: string, data: UpdateUserAsUserInput) => {
    const prevProfileIsComplete = initialData?.me?.isProfileComplete;
    return executeMutation({
      id,
      user: data,
    }).then((result) => {
      if (result.data?.updateUserAsUser) {
        const postProfileIsComplete =
          result.data?.updateUserAsUser?.isProfileComplete;
        // Check if profile is now complete, and it wasn't when the page loaded.
        if (postProfileIsComplete && !prevProfileIsComplete) {
          toast.success(
            intl.formatMessage({
              defaultMessage:
                "All required fields are complete. You can now change your status.",
              description:
                "Message displayed to user when user profile completed.",
            }),
          );
        }
        navigate(paths.home());
        toast.success(
          intl.formatMessage({
            defaultMessage: "Work preferences updated successfully!",
            description:
              "Message displayed to user after user is updated successfully.",
          }),
        );
        return result.data.updateUserAsUser;
      }
      return Promise.reject(result.error);
    });
  };
  ```
  Disclaimer - I haven't tested this! 😅

File: frontend/talentsearch/src/js/components/workPreferencesForm/WorkPreferencesForm.tsx (hunk 6 above)
Comment (RVany) [reply frontend/talentsearch/src/js/components/workPreferencesForm/WorkPreferencesForm.tsx:209]: Yeah. Since we are checking in each form, Current value don't need to save as a sessionStorage. Initially , I tried to check it in profile page but realized that logic is wrong.( After getting un expected results)

**Comment by RVany** [comment]
> Actually, I think that if you add isProfileComplete to the initial get request on each page, you can do this without using session storage at all! For example, this could replace the `handleWorkPreferences` function in WorkPreferencesForm:
> 
> ```
> const handleWorkPreferences = (id: string, data: UpdateUserAsUserInput) => {
>     const prevProfileIsComplete = initialData?.me?.isProfileComplete;
>     return executeMutation({
>       id,
>       user: data,
>     }).then((result) => {
>       if (result.data?.updateUserAsUser) {
>         const postProfileIsComplete =
>           result.data?.updateUserAsUser?.isProfileComplete;
>         // Check if profile is now complete, and it wasn't when the page loaded.
>         if (postProfileIsComplete && !prevProfileIsComplete) {
>           toast.success(
>             intl.formatMessage({
>               defaultMessage:
>                 "All required fields are complete. You can now change your status.",
>               description:
>                 "Message displayed to user when user profile completed.",
>             }),
>           );
>         }
>         navigate(paths.home());
>         toast.success(
>           intl.formatMessage({
>             defaultMessage: "Work preferences updated successfully!",
>             description:
>               "Message displayed to user after user is updated successfully.",
>           }),
>         );
>         return result.data.updateUserAsUser;
>       }
>       return Promise.reject(result.error);
>     });
>   };
> ```
> 
> Disclaimer - I haven't tested this! 😅

Yes! It should work. Again , Since unable to do the checking in profile page, we don't need the session Storage at all😀

File: frontend/talentsearch/src/js/components/GovernmentInfoForm/GovernmentInfoForm.tsx (hunk 5 above)
Comment (RVany) [reply frontend/talentsearch/src/js/components/GovernmentInfoForm/GovernmentInfoForm.tsx:446]: Since we are displaying the toast in all the necessary forms, Test needs to be done in all the forms. Am I correct?

**Review by petertgiles — APPROVED** [review]
This turned out really nicely!  If it's not possible to jest test the toast I'm satisfied anyway.

**Review by tristan-orourke — CHANGES_REQUESTED** [review]

**Comment by tristan-orourke** [comment]
Seems like testing the toast will be harder than we thought. Primarily because the creation of the toast (in this case) is tied closely with the API requests and response - so either we have to mock the API response via the urql library, or save this for integration testing.

File: frontend/talentsearch/src/js/components/GovernmentInfoForm/GovInfoFormOperations.graphql [hunk 8]
```diff
@@ -34,6 +35,7 @@ mutation UpdateGovAsUser($id: ID!, $user: UpdateUserAsUserInput!) {
     isGovEmployee
     govEmployeeType
     interestedInLaterOrSecondment
+    isProfileComplete
```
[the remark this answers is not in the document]
tristan-orourke: Nevermind, you were correct here, sorry!

tristan-orourke (changes requested):
Good work @RVany, I appreciate you moving more of the intl messages into a common file.

After looking at the code like this, I think the `getProfileCompleteToast` is too small to be useful. I think the code would be clearer if we deleted that function, and just repeated the lines
```
if (!preProfileStatus && currentProfileStatus) {
  return toast.success(message);
}
``` 
where they are needed. Many short functions are useful of course, but this one is an [impure function](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/documentation/style-guide.md#main-principles), which means its hard to understand what it does just by reading its name and looking at its type signature. Also, the logic inside is very simple, so we don't gain anything from hiding it away.

The other last change I'd like to see is remove the isProfileComplete field from the graphql schema.

RVany:
> Good work @RVany, I appreciate you moving more of the intl messages into a common file.
> 
> After looking at the code like this, I think the `getProfileCompleteToast` is too small to be useful. I think the code would be clearer if we deleted that function, and just repeated the lines
> 
> ```
> if (!preProfileStatus && currentProfileStatus) {
>   return toast.success(message);
> }
> ```
> 
> where they are needed. Many short functions are useful of course, but this one is an [impure function](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/documentation/style-guide.md#main-principles), which means its hard to understand what it does just by reading its name and looking at its type signature. Also, the logic inside is very simple, so we don't gain anything from hiding it away.
> 
> The other last change I'd like to see is remove the isProfileComplete field from the graphql schema.

Thank you! I will delete the helper and modifing all the forms. I didn't get `isProfileComplete` field part though.

frontend/talentsearch/src/js/components/aboutMeForm/AboutMeForm.tsx
```diff
@@ -324,6 +325,18 @@ const AboutMeFormContainer: React.FunctionComponent = () => {
...
+          const currentProfileStatus =
+            res.data?.updateUserAsUser?.isProfileComplete;
+          if (!preProfileStatus && currentProfileStatus) {
+            toast.success(
+              intl.formatMessage({
+                defaultMessage:
+                  "All required fields are complete. You can now change your status.",
+                description:
+                  "Message displayed to user when user profile completed.",
+              }),
```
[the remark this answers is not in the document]
RVany: Ok. I will do that.

tristan-orourke:
> I didn't get isProfileComplete field part though. Actually we added the isProfileCoplete field in graphql schema for previous issue. Do you mean delete from graphql mutation operations?

There's a difference between the graphql schema (where we define what you _can_ ask the API) and the operations files (where we define the queries and mutations we _actually_ send to the API). 

The places where you've added isProfileComplete to the operations makes sense (my mistake for suggesting otherwise earlier) because you're saying "We want to know this information (isProfileComplete) after doing a mutation". For this to work, isProfileComplete has to be in the schema, as part of the User and Applicant types, so that we can ask for that field in those types. 

However, isProfileComplete is already part of the User and Applicant types in the schema, from a previous PR. The place where you've added isProfileComplete to the schema in this PR is an `input` type, which defines what you can change in a mutation. This doesn't work, because isProfileComplete can't be set in a mutation.

api/graphql/schema.graphql
```diff
@@ -730,6 +730,9 @@ input UpdateUserAsUserInput {
     communityExperiences: CommunityExperienceHasMany
     educationExperiences: EducationExperienceHasMany
     awardExperiences: AwardExperienceHasMany
+
+    # Profile Status
+    isProfileComplete: Boolean
```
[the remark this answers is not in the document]
RVany: > However, isProfileComplete is already part of the User and Applicant types in the schema, from a previous PR. The place where you've added isProfileComplete to the schema in this PR is an `input` type, which defines what you can change in a mutation. This doesn't work, because isProfileComplete can't be set in a mutation.

Actually isProfileComplete is not added to Applicant types yet. Do we need that? I was thinking same but without adding it is giving typscript error when I added to mutations. I will try again and let you know.

RVany:
@tristan-orourke I updated the schema as you suggested. I didn't see any error so far and I am going to test all the forms in UI. Let you know if any error appear regarding this change.

RVany:
> This turned out really nicely! If it's not possible to jest test the toast I'm satisfied anyway.

Thank you! We are adding common profile messages.

RVany:
@tristan-orourke Your suggestion is perferctly working. I tested all the forms. May be I mis understood the error and change the input type of UpdateUserAsUser in the schema. 

- I deleted the isProfileComplete field from schema and light-house schema. 
- npm run codegen (To make sure delete field went away from generated file)
- npm run watch (To make sure updated reflecting in the UI)
- update the specific user directly in the db and set each form as last form and made sure profile complete message appeared.

Hope forms are working fine with the changes. Please let me know if I could check in any other ways.

tristan-orourke (approved):
This looks all good now, please go ahead and merge it 😁

RVany:
> This looks all good now, please go ahead and merge it 😁

Thank you!

Issue #2650, opened 2022-05-04
Deeplink Modals/Tabs


esizer:
h/t @patcon

## Summary

We are essentially looking to have the ability for someone to share a link to a modal to avoid "hidden" content. This would allow users to copy a URL and share it to someone to a page where the modal is open by default.

### Note

There is a [discussion about changing routers](https://github.com/gctc-ntgc/gc-digital-talent/issues/2368) so it would be beneficial to wait until that spike is resolved before we start this but not necessary.

## Approach

This could be accomplished by using optional parameters in the router and if that parameter exists, default the modal state to open.

## Example 

```tsx

// Router.tsx
...
const router = new UniversalRouter({
  path: '/path/segment/:modal?',
  action: ({ params }) => <Page currentModal={params.modal} />
}
...

// Page.tsx
...
interface PageProps {
  currentModal?: string;
}

const Page: React.FC<PageProps> = ({ currentModal }) => {
  const [isOpen, setOpen] = React.useState<boolean>(currrentModal === "learn-more");
  return (
    <a role="button" href="/path/segment/learn-more">Open Modal</a>
    <Modal isOpen={isOpen} />
  );
}
...
```

---

## Notes from Discussion

> However, im not sure how we can make this link obvious? Maybe convert the buttons that control the modals to links with the button role so they can have href to be copied?

@esizer I like your thinking on this! As for the href, I was actually thinking a fully qualified path, like `/en/indigenous-it-apprentice/learn-more`. A path fragment could work, but would probably break any future hope of browser deep-linking, which already uses hashes.

But yeah, wondering if maybe this is something we should hold off on being too clever about until we're 100% locked into a router. I recall we're maybe considering a more accessible router solution than universal-router. I'm finding posts about giving routes to modals, but they all seem to be for react-router:
- https://www.codedaily.io/tutorials/Create-a-Modal-Route-with-React-Router 👈 better one
- https://blog.logrocket.com/building-a-modal-module-for-react-with-react-router/
[...]

Issue #2368, opened 2022-03-31
Spike - Are there a11y advantages to React-Router?


tristan-orourke:
# Description
We're doing frontend routing with a custom hook and [UniversalRouter](https://github.com/kriasoft/universal-router). Are there any major features, especially around accessibility, that [React Router](https://reactrouter.com/) offers that we are currently missing?

Timeboxed to 1-2 hours.

tristan-orourke:
Just from a cursory search, I wonder if it has to do with this - announcing page navigations for screen readers? https://estevanmaito.me/blog/accessible-navigation-using-react-router

patcon:
What I recall of reach-router and react-router differences and fork/merge:

- https://reach.tech/router/accessibility
- https://reach.tech/router/credits (react vs reach)
  - polyfills aren't supported (re: small size) re: #2361 
  - accessibility
- https://reactrouter.com/docs/en/v6/upgrading/reach
  - react-router v6 incorporates many APIs of reach-router
  - accessibility: "Stop doing not-good-enough focus management by default."

Both reach and react-router now handle focus management in much more accessible ways, for both client-side and server-size.

Open questions:
1. Does accessibility concerns for "old" router versions only apply to server-side (of which we don't currently rely)? Will we ever use server-side rendering?
2. How will out browser support goals intersect with router choice?

patcon:
Still need to read this, but mentions (circa 2019) that **server-side rendering is highly recommended** for accessibility in govuk react sites: https://github.com/govuk-react/govuk-react/blob/main/ACCESSIBILITY.md

patcon:
Via [React Router v6 in Three Minutes](https://betterprogramming.pub/react-router-v6-in-two-minutes-a7a2963e2340) (Feb 2020):

https://twitter.com/mjackson/status/1229156979714605056

> what was the biggest factor that led to this change [in bundle size down 70%]?

> A couple of things:
> - dropping support for anything older than IE11
> - dropping support for React < 16.8
> - using Google Closure Compiler
> - writing better code 😅

patcon:
Not specifically a11y, but here's another downside of not having react-router.

I'm finding it hard to imagine how component state is exercised for stories whose components contain logic based on `location.pathname`. As it stands, we'd need to make pathname or location optionally injectable into each component using that logic. This is due to how we're implementing universal-router without any context to override:

[...]

PR #2728, opened 2022-05-12
Feature - 2633 -  Collapsible Menu

### LINKED ISSUE #2633: Collapsible menu
Question: where do the existing menu links go that aren't in the prototypes? eg skills, skill families, etc.

# Acceptance Criteria

- [x] new aesthetic
- [x] determine icons (from library?)
- [x] collapsibility (overflow and width shift?)

esizer:
Resolves #2633

## To Do

- [x] Highlight active link
- [x] [Fix footer placement](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2728#issuecomment-1125506624)
- [x] Look into how to write tests
- [x]  Slightly wider (try 18rem)
- [x] Icons slightly smaller
- [x] Investigate tooltip
- [x] Ensure all required links appear
- [x] Get designer sign off
- [x] Test a11y
- [x] pull in commits from admin's SideMenu.stories.js
    - https://github.com/patcon/gc-digital-talent/compare/253cd033363db8e420d17abfa13fb19b03d6ddcc...feature/2633-collapsible-admin-menu-patcon-stories
- [x] rename SideMenu to AdminSideMenu
    - Context+Component from [React components naming convention](https://medium.com/@wittydeveloper/react-components-naming-convention-%EF%B8%8F-b50303551505)
- [x] consider extra width compared to mockup

## Summary

This adds the collapsible menu component to the common project as well as using it in the admin project. In the process, some logic has moved around. We have reduced the responsibility of the `<PoolDashboard>` and `<Dashboard>` components moving it to the menu.

### Highlights

- New  components in common
  - `<SideMenu>` - Collapsible Sidebar Menu
  - `<SideMenuItem>` - Menu Link/Button with Icon
  - `<SideMenuContentWrapper>` - Provides the offset for the menu on larger screens
- Updates to the `<PoolDashboard>` moving links to the `<SideMenu>` component in admin
-  Rewrite of the admin `<SideMenu>` component to use new common `<SideMenu>`

## Comments

Overall I'm not super happy with the results and some of the component names.

Suggestions welcome! 😁 

### Component Names

Something that is very confusing is that we now have two components names `<SideMenu>`. The one in Admin abstracts the common component and I didn't really realize this at first and left it since I'm struggling to think of more appropriate name / approach to make.

### Moving of Admin Logic

The menu links have been moved to `/frontend/admin/js/components/menu/SideMenu.tsx` so I have moved the logic to check if it is admin. The reason for this was, we already have a lot of imports happening in that `<PoolDashboard>` component and it was getting fairly difficult to read and I didn't want to add all the icon imports on top of that.

Essentially going for, more components with less code in each in hope to improve legibility.

#### Defining a Menu Link

You can now define the menu links in the [SideMenu.tsx](https://github.com/GCTC-NTGC/gc-digital-talent/blob/c254794bb9b12f373b177d4f69259be85f789be8/frontend/admin/src/js/components/menu/SideMenu.tsx#L61-L70) file in admin.

```tsx
  const menuItems = [
    {
      key: "dashboard",
      href: paths.dashboard(),
      icon: HomeIcon,
      roles: [Role.Admin],
      text: intl.formatMessage({
        defaultMessage: "Dashboard",
        description: "Label displayed on the dashboard menu item.",
      }),
    },
...
]
````

##### Properties

- key: Used directly as the key in the map
- href: Path to where the link navigates to
- icon: Icon directly from [@heroicons](https://heroicons.com/)
- roles: Array of the roles who can see this link. null = anyone can see
- text: The text that appears in the menu link

### Mockup

Screen Shot 2022-05-13 at 9 04 00 AM

### Result

Screen Shot 2022-05-13 at 9 04 49 AM

### Differences

- Test seems to render slightly heavier (might have to do with browser v figma)
- Different links, used same as existing over the mockup

### CODE REVIEW

**Comment by vd1992** [comment]
**Appearance**
Hm... regarding appearances. Looking at the chromatic diff (thank you gray's wallet)
https://www.chromatic.com/test?appId=61e099a1bb1465003a73d3ce&id=627d6e477dac15004a58c06e

the footer shot up for some reason and I personally liked how the menu options were centred before, now there is so much space to the right of the menu options

**Usability** 
the icons don't reveal what they link to without looking at the URL in the browser corner, and in the case of the logout button you can't tell it is the logout button without expanding the menu
hover text might be a nice addition 

**Naming**
Regarding the naming of menu components, my understanding is that in /common you've got the core menu component and the one in /admin adds onto it, correct? 
In which case, you could say the latter extends the former so...
`SideMenu`
and
`SideMenuAdmin` or `SideMenuExtendAdmin`

Something along that line to show the base component and that the other extends it.

**Comment by esizer** [comment]
> **Appearance** Hm... regarding appearances. Looking at the chromatic diff (thank you gray's wallet) https://www.chromatic.com/test?appId=61e099a1bb1465003a73d3ce&id=627d6e477dac15004a58c06e
> 
> the footer shot up for some reason and I personally liked how the menu options were centred before, now there is so much space to the right of the menu options
> 
> **Usability** the icons don't reveal what they link to without looking at the URL in the browser corner, and in the case of the logout button you can't tell it is the logout button without expanding the menu hover text might be a nice addition
> 
> **Naming** Regarding the naming of menu components, my understanding is that in /common you've got the core menu component and the one in /admin adds onto it, correct? In which case, you could say the latter extends the former so... `SideMenu` and `SideMenuAdmin` or `SideMenuExtendAdmin`
> 
> Something along that line to show the base component and that the other extends it.

Thanks, @vd1992 !

I have fixed the footer with [fix footer placement](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2728/commits/2acc7a5da9c746696a9b11fa3e8cfa7dd8c02c24)

**Comment by esizer** [comment]
@vd1992 Love your idea of a tooltip, Im wondering if we want this as part of this PRs scope or add it to a new issue?

Since we are using `overflow: hidden;` to create the close effect on desktop screens a tooltip becomes a little more complicated. We should likely use a portal to render the tooltip outside of the menu tree and position it next to the item being hovered.

I don't think it is something so complicated we need to rely on a third party library, maybe just complicated enough it does not fall under the scope of the issue/PR?

**Comment by esizer** [comment]
@Jerryescandon Can you take a look at the icons I selected for the additional links and either confirm they are okay or provide alternatives from [@heroicons](https://heroicons.com/)? Thanks!

Screen Shot 2022-05-13 at 12 12 55 PM

**Comment by Jerryescandon** [comment]
love em, ship it.

**Comment by esizer** [comment]
> > I don't think it is something so complicated we need to rely on a third party library, maybe just complicated enough it does not fall under the scope of the issue/PR?
> 
> maybe new PR, with designer input? seems like it could be a larger pattern for places we use icons button :)

I can file a new ticket and iron out the details if both of you want to take a look and provide additional input/context once it is up? @vd1992 @patcon

**Comment by patcon** [comment]
What happens when y'all try to run `docker-compose run --rm maintenance bash setup.sh` locally?

I'm getting an issue with composer now trying to install from source, and not finding git in the container:

https://gist.github.com/patcon/f28bada64683ba6a7f51861fc5d525de

Screen Shot 2022-05-16 at 12 09 32 PM

The only real difference is that I merged in `main`, because that this includes the docker-compose.yml file move (and so I don't need to rebuild and blow away my volumes)

**Comment by patcon** [comment]
Ok, before having it built locally, thoughts on having stories of more variants the SideMenu component? I'm just seeing one for logged out?
https://www.chromatic.com/test?appId=61e099a1bb1465003a73d3ce&id=627e7391430614004a1a31a3

potentials:
- logged in
  - opened
  - closed
- logged out
  - open
  - closed

**Comment by patcon** [comment]
Hey! Really appreciate this Eric! I suspect you have more react chops than me, and I always learn a ton reading your code :)

Re: naming. How about `<AdminSideMenu />`? (I'll use that term below for clarity)

On diving in today, realized how tough it is to write stories in the way I'd alluded to above. But I made some progress. I had to rush out to an appointment after work, but will show code when I'm not on my phone. (Pls don't pay much attention to the comments below until I can push that diff in a gist -- don't worry, I won't push commits to your branch without discussing!)

But even without code, here's me bookmarking my most salient question: what are your thoughts on prioritizing stories for renderable components used in the apps themselves ("app components"), rather than stories for the abstracted components of common ("common components")? I am curious if that might help us write more atomic "app components", with clearer separation of rendering and data fetching, and ability to override that data/state. This focus would perhaps also more fully use chromatic for direct benefit to product development, since creating stories to exercise important component states of the app won't require _rebuilding_ a demo component from "common components" in the stories, such that it can revisit those important states (which can understandably feel like "make-work").

So in this case, I'm wondering if stories for admin's `AdminSideMenu` might be more helpful than those for common's `SideMenu`. (Even if those "app components" aren't going to be pretty right now)

I understand that there is currently some messiness in doing this. I can get them working with some changes to the supporting components. Mainly these changes seem to be adding some sort of optional `initialState` prop to context containers, for injecting overriding state data into contexts, and then decorating the story templates with these injectable context containers). It's still a bit messy, and maybe it's a larger project to get our existing "app components" more cleanly into storybook. (In my experiment, lots of errors on console log right now, but the clean-reading "app component" stories for `<AdminSideMenu />` succeed in simulating various states: open vs close, ADMIN role vs anon, changing active menu items (via `location.pathname`)... LoginOrLogout button is still defiant haha)

Also, I'm curious whether others have a rule of thumb for when to abstract components between common and admin? Do we do this as a matter of process and housekeeping, or do we wait until we're on the verge of implementing the component in the other app? I am certainly fond of abstracting quite early, though unsure if that's virtue or vice..! but seeing someone else do it brings the question to mind for how others decide :)

**Comment by esizer** [comment]
> Re: naming. How about `<AdminSideMenu />`? (I'll use that term below for clarity)

Good suggestion, I can rename that assuming it won't cause any conflict with what you are currently working on.

> Mainly these changes seem to be **adding some sort of optional `initialState` prop to context containers**, for injecting overriding state data into contexts, and then decorating the story templates with these injectable context containers).

Is this in reference to trying to mock auth in stories and tests? If so, I agree that using that prop would be useful so we can just inject something into the context like `initialStated={{loggedIn: true}}`

> Also, I'm **curious whether others have a rule of thumb for when to abstract components between common and admin?** Do we do this as a matter of process and housekeeping, or do we wait until we're on the verge of implementing the component in the other app? I am certainly fond of abstracting quite early, though unsure if that's virtue or vice..! but seeing someone else do it brings the question to mind for how others decide :)

When I start I simply think "**_could_** this be re-used elsewhere in another project", rather than "will we". So, I guess agreeing with your point about "abstracting quite early". I personally do not see much negative in abstracting UI and building block components. Obviously this won't work with components that are highly coupled with a specific project i.e. the edit profile forms in `talentsearch`.

**Comment by patcon** [comment]
Ok, here's the cleanest bit (i didn't commit the changes to make active path mockable, because that's even messier):
https://github.com/patcon/gc-digital-talent/compare/253cd033363db8e420d17abfa13fb19b03d6ddcc...feature/2633-collapsible-admin-menu-patcon-stories

Add `.diff` to the above path to see the raw diff, or here's the coloured gist:
https://gist.github.com/patcon/7b9ef55bc7711c03f24975a690c39541

How to stories look:


It's not perfect, but allows chromatic to visually capture the essence of various states, and spot changes. Making the console errors go away, and the login/logout button work, and the links no be wonky -- that feels like future opportunities for improvement to how we're building components and separating render vs logic, but feels like it can be out-of-scope for this PR :)

**Comment by patcon** [comment]
> I personally do not see much negative in abstracting UI and building block components.

I don't have strong war stories about this myself, but I've heard it advised by past coworkers (when I was more in devops than frontend) to **"avoid abstracting until there's a second implementation" or that "abstraction needs to see duplication [first]"**. I believe the supposed wisdom is that you can't truly abstract from a single example, and that the separation perhaps increases codebase complexity without confirmation that the abstraction is even the "right" one. (whatever "right" means haha -- it prob depends on the whims of the stakeholders, not how we think the second implementation _should_ work :) ) I understand it as a precautionary philosophy in the spirit of Knuth's words on [Premature Optimization](https://wiki.c2.com/?PrematureOptimization).

Just to show it's not just me making something up, some random posts discussing this: (the second even advocates WET, "write everything twice")
- [The cost of premature abstraction](https://archive.ph/Pgqzb#:~:text=When%20in%20doubt%2C%20duplicate%20code.%20The%20cost%20of%20a%20bad%20abstraction%20is%20much%2C%20much%20higher%20than%20the%20cost%20of%20dealing%20with%20duplicate%20code.) "When in doubt, duplicate code. The cost of a bad abstraction is much, much higher than the cost of dealing with duplicate code."
- [When DRY Doesn't Work, Go WET](https://dev.to/nickbulljs/when-dry-doesn-t-work-go-wet-4536#:~:text=need%20to%20remember%3A-,Duplication%20is%20far%20cheaper%20than%20the%20wrong%20abstraction.,-For%20example%2C%20you)
- [The false abstraction antipattern](https://mortoray.com/2014/08/01/the-false-abstraction-antipattern/#:~:text=An%20abstract%20type%20that%20has%20only%20a%20single%20implementation%20may%20be%20a%20false%20abstraction.) "An abstract type that has only a single implementation may be a false abstraction."
    - [Wikipedia: YAGNI "You aren't gonna need it"](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

To clarify, I'm not advocating that we collapse this back into simply admin components, but it's something I hope the team can discuss in future 🙏  I hope in future that we just write for admin (and focus on stories for those) until after we have copied and got it working elsewhere -- then we see what the shared  bits are :)

File: frontend/common/src/components/SideMenu/SideMenuContentWrapper.tsx [hunk 1]
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+
+const SideMenuContentWrapper: React.FC = ({ children }) => (
```
patcon, line 3: Could we call it something more descriptive, so it's more legible? You mentioned it's for wide screens, and it'd be great to know that without diving down a few layers into css.

Also, any reason the SideMenu wouldn't want this? Why it's own component?

esizer: Note, the naming is lack of creativity on my end and any suggestions would be helpful to better describe the components purpose.

esizer:
> To clarify, I'm not advocating that we collapse this back into simply admin components, but it's something I hope the team can discuss in future 🙏 I hope in future that we just write for admin (and focus on stories for those) until after we've duplicated code and got it working elsewhere -- then we see what the shared bits are :)

I agree, only abstract once needed. This is an example where it was likely unnecessary since, at-least to my knowledge, we have no plans to re-use it. 

Maybe when we create tickets that include new components we can see if we know if it will be reused and if an API can be determined for each use. I think something like a decision tree could help here. To start, two questions we could ask ourselves are:

- Will this component be re-used?
- Do we know how this component will be used in each project?

## Example

Something like a button, that is something we know will exist in each application. However, if we were creating a button for the first time we would need to determine if we know what the `props` for said button should be. If we cannot definitely answer those two questions then make it project specific.

So, for the `<Dialog />` I had previous knowledge that it would be reused and the designers provided the design files to allow me to see ahead of time the different variations on said component so I had the information to make an educated guess as to the API for the component.

patcon:
Anyhow, thanks Eric! I've got the LoginOrLogout button working now:
https://github.com/patcon/gc-digital-talent/compare/253cd033363db8e420d17abfa13fb19b03d6ddcc...feature/2633-collapsible-admin-menu-patcon-stories

Still figuring out if router is easily mockable. Open to suggestions. Happy to treat it as out-of-scope, but will be helpful eventually, in any place we use `useLocation` hook.

esizer:
> Still figuring out if router is easily mockable. Open to suggestions. Happy to treat it as out-of-scope, but will be helpful eventually, in any place we use `useLocation` hook.

I think someone else needs to weigh in on this, but we have #2368 where we are discussing react-router. I actually have no experience with `universal-router` so maybe this adds a bit more priority to making a decision on that spike? This spike was also brought up in #2650

esizer:
> Love this! And really like the idea of a decision tree, esp one with questions. I like that questions don't hardcode answers in the process :)

Great! Maybe we can create a spike to discuss how we want this to work and gather input on what types of questions will help us land at the appropriate amount of abstraction? I'm not sure what the next steps are but this seems like a good start?

frontend/admin/src/js/components/dashboard/Dashboard.tsx
```diff
@@ -169,34 +62,54 @@ export const Dashboard: React.FC<{
...
+          <MenuIcon
+            style={{ width: "1rem", height: "1rem", marginRight: "0.5rem" }}
+          />
+          <span>
+            {intl.formatMessage({
+              defaultMessage: "Open Menu",
+              description: "Text label for header button that opens side menu.",
+            })}
+          </span>
+        </Button>
```
patcon, line 109: Should this button be hidden when the menu is already open? It sticks around and the edge poking out from behind looks a little awkward.

Screen Shot 2022-05-17 at 4 53 02 PM

patcon: Also, for legibility, I often tease apart smaller components, though I don't store them in their own files, nor even export them. So like a `isMenuOpen ? null : <MobileOpenMenuButton onClickHandler={handleMenuToggle} />` 

Thought admittedly, this is a little less cumbersome in regular javascript, where it just feels like a helper function and prop definitions don't make it seem so heavy.

Not sure if this is your vibe. I've taken that "Thinking in React" article to heart, and tend to create components for any visually distinct items (and hide the presentation styles inside them), but I don't bog the project down with a file for each, at least not until things are really settled.

frontend/admin/src/js/components/menu/LoginOrLogout.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { useIntl } from "react-intl";
+import { LogoutIcon, LoginIcon } from "@heroicons/react/outline";
+import { SideMenuItem } from "@common/components/SideMenu";
+import { useLocation } from "@common/helpers/router";
+import { getLocale } from "@common/helpers/localize";
+
+import { AuthContext } from "../AuthContainer";
+import { useApiRoutes } from "../../apiRoutes";
+
+const LoginOrLogout = () => {
```
patcon, line 11: ❤️

frontend/admin/src/js/components/dashboard/Dashboard.tsx
esizer: So, I have moved it to its own component to make the return statement more readable. However, hiding it becomes tricky for a11y since we are using a focus trap to assist users with focus management.

## Focus Trap

1. User opens menu
2. Focus shifts to the first focusable item in the menu and prevents the user from changing focus outside of the menu (traps them)
3. User closes the menu (either with `esc` or by clicking "Close Menu")
4. Focus trap is released and focus shifts back to the element that invoked the trap (the button)

### Problem

If we remove the button from the DOM or hide it, the focus trap no longer knows what initially invoked it so it can't put focus in the proper spot. So we are kind of stuck with the button staying on screen.

### Solution?

The exact placement, style etc. were all decisions I made (close to bottom of screen for short thumbs) since we were  unfortunately not provided this in the mockups. So we can open this for discussion on where to place it so that it is not in an awkward position no matter the menu state.

patcon:
When menu is not open, I'm seeing a flicker of "closing" animation when screen size gets small enough to hit mobile breakpoint. Hard to see in the screencap, because it's quick and screencap is low framerate: https://recordit.co/3YWylr4KHw

Something to do with the way visibility is done, and perhaps to smooth transitions of desktop view interpolating with width values for mobile dimensions?

patcon:
Seems maybe room for improvement with width of menu in mobile mode, or breakpoint at which that menu kicks in?

Screen Shot 2022-05-17 at 4 53 02 PM

Seems like this could be narrower before kicking in? (I admittedly do a lot of window paneling on my one screen, so I notice when websites "mobilize" aggressively at mid-range breakpoints)

frontend/admin/src/js/components/dashboard/Dashboard.tsx
patcon: > close to bottom of screen for short thumbs

strong agree. and thanks for adding it! I have a friend who uses an assistive input device (and consults related to it), and thinking more about moving buttons and reach lately

focus trap considerations are new to me, so I'll trust you on that. Would this be a reason to add an `isVisible` prop which just hides it but keep the element itself in the DOM?

esizer: > focus trap considerations are new to me, so I'll trust you on that. Would this be a reason to add an isVisible prop which just hides it but keep the element itself in the DOM?

Yeah, i'll need to play around with that we can do to hide it will keeping it available to be focused later on.

esizer: So, I've been trying different things and it seems opacity is the best way to do this. Applying different `data-h2-*` to try and use hyddrogen resulted in re-render for some reason that cause issues with the focus trap.

Updated with [hide toggle button when menu open](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2728/commits/7183c3df4f50b138d05741ce67389d66ebe85988)

esizer:
> Something to do with the way visibility is done, and perhaps to smooth transitions of desktop view interpolating with width values for mobile dimensions?

You are 100% correct, we have a transition of 250ms on the menu and when it "closes" going down to mobile that is what you are seeing. I tried reducing this but it honestly looks worse imo because it is faster. I reduced it slightly and it is less noticeable without being a jarring flash. Can you check and see if it is slightly better?

I'm not too sure how this can be avoided without removing the transition entirely. It's totally only to make it look pretty and has no functional purpose so that is more up for discussion, what do you think?

esizer:
> Seems like this could be narrower in width, or maybe the mobile view kick in later? (I admittedly do a lot of window paneling on my one screen, so I notice when websites "mobilize" aggressively at mid-range breakpoints)

So, we do not have a better breakpoint available in hydrogen right now unfortunately and i'm not sure of the impact to add an additional one.

In regards to the actual width, this was just bad css on my part.

```css
/* Old */
.side-menu {
  width: 95vw; /* Was always 95% of the current screen */
}

/* New */
.side-menu {
  width: 18rem; /* Uses desktop width */
  max-width: 95vw; /* up to a maximum of 95% of the current screen */
}
```

frontend/admin/src/js/components/dashboard/Dashboard.tsx
```diff
@@ -151,15 +33,57 @@ const AdminNotFound: React.FC = () => {
...
-export const Dashboard: React.FC<{
-  menuItems: ReactElement[];
+interface OpenMenuButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
+  show: boolean;
+}
+
+const OpenMenuButton: React.FC<OpenMenuButtonProps> = ({
+  show,
+  onClick,
+  children,
```
patcon, line 43: Why use children, if this a component library button? Feels cleaner and more legible at the high level when buttons that don't have another use just get this hidden inside. Thoughts?

I'm not sure there's value in using children so readily? Again, with abstraction, I would pull children out only if we ever come back to re-used this exact button, and not before.

```jsx
      <OpenMenuButton onClick={handleMenuToggle} show={!isMenuOpen}>
        {intl.formatMessage({
          defaultMessage: "Open Menu",
          description: "Text label for header button that opens side menu.",
        })}
      </OpenMenuButton>
```

vs 

```jsx
      <OpenMenuButton onClick={handleMenuToggle} show={!isMenuOpen} />
```

patcon: i welcome push-back on this btw :)

frontend/common/src/components/SideMenu/SideMenuContentWrapper.tsx
patcon: That's helpful! I didn't realize the relation from the name, but I can see now why it makes sense :)

patcon: This is resolved unless you have any ideas to make confusion like mine go away haha

frontend/admin/src/js/components/dashboard/Dashboard.tsx
esizer: That would require adding `const intl = useIntl();` to the component introducing more complexity when it is just easier using children when `intl` is available in the parent scope. Only reason I chose that direction.

patcon:
> Can you check and see if it is slightly better?

> I'm not too sure how this can be avoided without removing the transition entirely. It's totally only to make it look pretty and has no functional purpose so that is more up for discussion, what do you think?

Not seeing any change, unfortunately. But actually, is this animation just the slide in/out that is only visible on page resize (not an animation that's seen when open/close'ing the menu)? If so, I'm inclined to think it's worth removing, since people will only see the slick animation if they're also seeing the janky glitching -- so not sure it's adding the intended benefit.

patcon:
> You are supposed to see it when opening and closing. 

Odd. I wasn't seeing and transitions except on screen resize. But because this isn't rebased on mainline lately (with its moved docker-compose), it's quite inconvenient to rebuild this when coming back to it, and maybe i missed a step.

I added my understanding of the outstanding todos above, with one final one:

I noticed in the mockup that the menu in the mockup was narrower. Any reason for adding extra width in menu? Since the content frame is pushed aside for the larger menu, I wonder if we could conserve as much content width as possible

frontend/common/src/components/SideMenu/SideMenuContentWrapper.tsx
patcon: This is helpful! Going to resolve

patcon (approved):
LGTM! Thanks! (Sorry for the delay on the final approval)

PR #2730, opened 2022-05-12
enabled go back on about-me submit

### LINKED ISSUE #2704: Profiles - About Me Does Not Go Back
If I fill in the "About Me" page and click "Save and Go Back" it saves but does not go back.

vd1992:
resolves #2704 

Issue: navigation was back to the same form on submit
Solution: changed path


petertgiles (approved):
Works for me!

PR #2731, opened 2022-05-12
Ran lint:fix in all frontend projects

### LINKED ISSUE #2631: Eslint and Prettier cause different results
It seems like Eslint with the Prettier plugin and Prettier (standalone) cause different results when beautifying a file.  I'm not clear why that is but we may want to check our configs to see if we have multiple conflicting results.

An example of two commands that result in different versions of the file (from the frontend/common folder):
`npx eslint src/components/SearchRequestFilters/SearchRequestFilters.tsx --fix`
`npx prettier src/components/SearchRequestFilters/SearchRequestFilters.tsx -w`

tristan-orourke:
Resolves #2631 

Now that eslint is properly using the Prettier config again, its making some PR diffs more cluttered than necessary. Better to get all the format changes out of the way at once.

I reran ` npm run lint:fix --workspaces --if-present` in the frontend directory. It seems to have just update the max line-length, and subsequent tab indents.


vd1992 (approved):
No more eslint prettier warnings, that's good. Hopefully things don't flip back again 😝 
For the sake of it, I also tried running in /frontend this `npm run prettier --workspaces`

these were the remaining suggested changes
nothing crazy, but I think cleaning up those graphQL files might be a nice added bit? Your choice, I'll approve regardless
image

vd1992:
oh yeah, no issue linked intentionally?

tristan-orourke:
Thanks @vd1992! I didn't realize that Prettier will format more file types than Eslint (at least with our current config).

Issue #2716, opened 2022-05-11
Remove reliance on `x5c` property at /jwks


patcon:
The staging `/jwks` endpoint: https://te-auth.id.tbs-sct.gc.ca/oxauth/restv1/jwks

**Success criteria:** Finding another way to check token sigs without using `5xc` property.

This would allow us to take a step toward replacing our custom auth app (for local development and CI) with one of the more turn-key mock oauth2 server.

I recommend we remove our usage of the poorly supported (as per a few mock oauth server projects) and unadvised ([as per Google Engineering](https://github.com/google/jws#:~:text=we%20eliminate%20easy%2Dto%2Dmisuse%20options%20such%20as%20%22x5c%22%2C%20%22x5u%22) and also [in-practice](https://www.googleapis.com/oauth2/v3/certs) in their infra) openid feature (the `x5c` property).

## Benefits

This would allow us to more easily benefit from migrating to a mock oauth2 server (for local development and CI). Migrating to a mock server would allow us to abandon our own `auth` app, and the special-case auth code paths within `api` app.

## Relevant code

Where we process the `x5c` property into a config object we use to later check token sigs:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/13ae7e3bb79d564a80c809226ba61966c968c122/api/app/Services/OpenIdBearerTokenService.php#L74-L80

Email from Doug Harris of SiC:
> The “x5c” property contains the same public key represented by the “n” and “e” properties, just presented as an X.509 certificate for any software out there that might prefer the older format. It is completely redundant and ignored by most software I’ve worked with.
>
> As long as your software is still checking the signature on our ID tokens the normal way you can ignore the “x5c”.

Added this to icebox for now. cc @tristan-orourke

patcon:
This is noted missing support from the library we're using: https://github.com/lcobucci/jwt/issues/32#issuecomment-392808889

This is a library suggested for it: https://github.com/acodercat/php-jwk-to-pem

Or this is perhaps better, as a more minimal dep is required, which doesn't pull in phpseclib:
https://github.com/lcobucci/jwt/issues/32#issuecomment-907556410

yonikid15:
This is resolved in #2813

PR #2741, opened 2022-05-16
Remove requirement for x5c key in jwks

patcon:
Closes https://github.com/GCTC-NTGC/gc-digital-talent/issues/2716

## To Do
- [x] Added `web-token/jwt-core` and updated lockfile
- [x] Added note about php ext required
- [x] Remove requirement for using `x5c` prop in JWKS endpoint data, in favour of generating its public key from the props `n` (modulus) and `e` (exponent).
- [x] update test data so jwks resources have "e" and "n", but lack "x5c"

Result: Can now use an off-the-shelf mock oauth2 server (which don't support the niche x5c prop)

## Alternatives
1. using a pure-php crypto library like `phpseclib` directly (without new system reqs like `bcmath` or `gmp`)
2. using another JWT library that uses `phpseclib`: https://phpseclib.com/docs/rsa#raw-rsa-public-keys
3. maybe doing it without a library: https://www.identityblog.com/?p=389

## Open Question

1. Is `bcmath` PHP extension available to enable/install on the UAT server? If not, what about `gmp`? (both are required for the currently implemented JWT library that handles working with JWKS endpoint data.) Are we open to using one of them?
2. would we rather use one of the alternatives?
    1. are these alternatives slower? and does speed even matter?


petertgiles:
Don't forget to inform people that they'll need to rebuild the image for *php* and then instantiate a new container.

patcon:
> It pegs my CPU at 100% for several minutes - I haven't let it finish yet. In the logs for Actions it looks like it completes normally. Is this happening for you?

Are you running phpunit in docker, or locally? If locally, you'll need to install bcmath pho ext (which I added to the workflow)

There's a command to run phpunit inside the php container of docker via `docker-compose run` but I recall our docs only mention running it locally?

patcon:
I feel funny installing a whole other jwt management lib for just these two jkws functions, though its true that our current lib doesn't have any supports for managing jwks. In theory, we could choose a new lib that handles jwks better, with an eye to perhaps fully migrating to it at some point (if it serves us better than lcobucci/jwt).

Options:
1. keep this PR. Aka install a second jwt management package, and install bcmath/gmp on docker+UAT (running phpunit locally also will not work without those)
2. Use alternative (3) above, and copy in our own minimal pure PHP public key derivation code.
3. Keep this basic PR, but use the [firebase-jwt lib](https://github.com/firebase/php-jwt) instead. Perks: built on a pure-php polyfill, that optimistically uses an audited, non-php crypto library (libsodium) if available on the OS. Most widely-used PHP jwt lib. Risks: many open GitHub issues, so that's a slight smell.

I'd recommend the third option, of using firebase-jwt instead of the jwt-core for jwks processing (again, for ONLY the jwks code in this PR, not swapping out [our whole lcobucci/jwt lib](https://github.com/lcobucci/jwt))

petertgiles:
Oh, of course, yeah it runs fine in the container with the new dep.  :blush:   That's weird behaviour - it doesn't display an error message but just locks up.  So, do we have this dep in production or is it going to lock up there, too?

For the decision - I guess the best dep is the one we have access to in production. It sounds like we can't just install whatever we want. :grin: I don't have a strong opinion beyond that.  @tristan-orourke?

patcon:
Closing this in favour of a new PR

PR #2779, opened 2022-05-17
Bug - 2777 - Check refreshed AuthState before returning 

### LINKED ISSUE #2777: User still considered logged in when token is expired
When the user logs in to the admin portal, and then leaves the site for a while, and comes back later, the site will still appear as Logged In, but api calls will fail because the access token is expired.

If the access token is refused, the refresh token should always be used to get a new access token. If the refresh token is rejected as well, the frontend should trigger a logout.

esizer:
Resolves #2777 

## Summary

This adds an `await` to the `refreshTokenSet` function so we can check the state before returning it. Additionally, if that state is no valid, we call `logout`.

Since we reuse `logout` and `return null;` now, this can been moved to a function.

```tsx
/**
 * Logout the user and return null AuthState
 *
  * @returns null
*/
const logoutNullState = () => {
  logout();
  return null;
};
```

## Testing

We should write some automated tests here to confirm we can refresh tokens and properly logout when needed. However, I am not sure where to start and opted to get this PR through and do the following manual testing.

### Invalid Access Token + Valid Refresh Token

This tests that the refresh token is able to successfully refresh the access token.

1. Login to the application
2. Open the DevTools and delete or modify the access token to dirty the AuthState and force a refresh attempt
3. Refresh the page
4. Make sure the token is refreshed and you are still logged in

### Invalid Access Token + Invalid Refresh Token

Test to make sure we are logged out when both tokens are invalid causing an error on refresh.

1. Login to the application
2. Open the DevTools and delete or modify the access and refresh token to dirty the AuthState and force a refresh attempt
3. Refresh the page
4. Make sure you are logged out and redirected to the homepage

### DIFF
--- a/frontend/common/src/components/ClientProvider/ClientProvider.tsx
+++ b/frontend/common/src/components/ClientProvider/ClientProvider.tsx
@@ -109,15 +109,28 @@ const ClientProvider: React.FC<{ client?: Client }> = ({
         return null;
       }
 
+      /**
+       * Logout the user and return null AuthState
+       *
+       * @returns null
+       */
+      const logoutNullState = () => {
+        logout();
+        return null;
+      };
+
       // If authState is not null, and getAuth is called again, then it means authentication failed for some reason.
       // let's try to use a refresh token to get new tokens
       if (refreshToken) {
-        const refreshedAuthState = refreshTokenSet();
-        return refreshedAuthState;
+        const refreshedAuthState = await refreshTokenSet();
+        if (refreshedAuthState) {
+          return refreshedAuthState;
+        }
+
+        return logoutNullState();
       }
 
-      logout();
-      return null;
+      return logoutNullState();
     },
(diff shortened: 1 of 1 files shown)


vd1992:
Played around with it, I think it works as intended 

Mangled Access Token then refresh
- refresh seems to trigger
- page reloads, logged in still

Mangled Access and Refresh Token, then refresh
- logout triggered
- logged out

Not sure about what is to be done regarding testing and approval metric for auth stuff, so I won't approve yet

esizer:
> Not sure about what is to be done regarding testing and approval metric for auth stuff, so I won't approve yet

Yeah, I think we have some issues related to mocking auth for testing purposes. @patcon are you able to provide any assistance in writing automated tests for this?

tristan-orourke (approved):
I'm happy to accept the manual tests for now, and tag this as another e2e test candidate

PR #2805, opened 2022-05-18
Group two halves of classification filter

### LINKED ISSUE #2801: No filters yields 40 candidates, but selecting all the classifications possible yields 50 candidates?
- initial candidate count appears correct now, query passes in variables
- filter toggling now behaves as expected
- one instance of confusing behaviour found, see below 

No filters yields 40 candidates, but selecting all the classifications possible yields 50 candidates?

image

_Originally posted by @vd1992 in https://github.com/GCTC-NTGC/gc-digital-talent/issues/2778#issuecomment-1129153479_

petertgiles:
This branch solves the bug where candidates can get pulled in from other pools in the salary-to-classification query.  h/t to Eagle-Eye @vd1992 .  I started by adding some tests for validation.

The query was being built like this:
```
select count(*) from pool_candidates where exists (
  select from joined tables where
  (exact classification match)
  or
  (salary to classification match)
  and
  (candidate_id predicate for exists clause)
)
```
Since AND has precedence the OR operator was not working as intended. I updated the builder to group the OR like this:
```
select count(*) from pool_candidates where exists (
  select from joined tables where
  (
    (exact classification match)
    or
    (salary to classification match)
  )
  and
  (candidate_id predicate for exists clause)
)
```

Closes #2801

### DIFF
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -74,22 +74,26 @@ class PoolCandidate extends Model
     {
         // Classifications act as an OR filter. The query should return candidates with any of the classifications.
         // A single whereHas clause for the relationship, containing multiple orWhere clauses accomplishes this.
-        $query->whereHas('expectedClassifications', function ($query) use ($classifications) {
-            foreach ($classifications as $index => $classification) {
-                if ($index === 0) {
-                    // First iteration must use where instead of orWhere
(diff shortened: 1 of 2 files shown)


tristan-orourke (approved):
Thanks @petertgiles! (And @vd1992.) Love the tests.

Issue #2648, opened 2022-05-03
Spike - Investigate production of files for sharing resumes


tristan-orourke:

esizer:
I see two options and neither is extremely difficult if we keep the PDFs simple and informational. Most of the complexity in generating PDFs comes from performance and styling.

1. Client - [react-pdf](https://react-pdf.org/)
2. Server - [laravel-dompdf](https://github.com/barryvdh/laravel-dompdf)

## Client - react-pdf

This library uses react components to render a PDF in the client and has the option for [on the fly rendering](https://react-pdf.org/advanced#on-the-fly-rendering) of a PDF download link.

When a user loads a page, the library will render the PDF in the background and once completed the button is enabled and clicking it will download the PDF. However, I have implemented this method in the past and it quickly becomes unreasonable as you add more information.

```jsx
import { PDFDownloadLink, Document, Page } from '@react-pdf/renderer';

const MyDoc = () => (
  <Document>
    <Page>
      // My document data
    </Page>
  </Document>
);

const App = () => (
  <div>
    <PDFDownloadLink document={<MyDoc />} fileName="somename.pdf">
      {({ blob, url, loading, error }) =>
        loading ? 'Loading document...' : 'Download now!'
      }
    </PDFDownloadLink>
  </div>
);
```

## Server - laravel-dompdf

While I have never used this specific library, I have used the one if wraps (dompdf). This allows you to pass it a blade view which it then converts to a PDF for download.

```php
use Barryvdh\DomPDF\Facade\Pdf;

$pdf = PDF::loadView('pdf.invoice', $data);
return $pdf->download('invoice.pdf');
```

PR #2816, opened 2022-05-19
Add React Print Demo

petertgiles:
This branch demonstrates a possible solution to the "export to PDF" spike.

I didn't really like the "render on the backend" approach.  The library Eric suggested installed very easily and was nice to use but bypassing the graphql schema (and it's guards) made me uncomfortable.  Also, we don't have the ability to access the localized constants (or any localization) on the backend.

I didn't really like the "render to PDF" in react approach.  It worked nicely but blew out the bundle size (dev bundle size) from about 6 MB to 25 MB.  It might be possible to use bundle splittling to make the user download that extra only on print.  It also freezes the browser for about 8 seconds to render about 30 KB worth of JSON data.

I preferred the approach of using a React print library.  It only increased the dev bundle size from 6 MB to about 8 MB.  It was nice to use and performs much better.  And the user could actually use it to print if they want. :laughing: It's a little less user friendly since the concept of printing to get a PDF file is a little confusing.  On the plus side, they could use it to export to XPS or whatever other export printer driver they have installed.

To try it out, click the "Download Users" on the dashboard:
image

Related to #2648 but will probably not be merged.  Just use this branch as a reference.

### DIFF
--- a/frontend/admin/package.json
+++ b/frontend/admin/package.json
@@ -113,6 +113,7 @@
     "react-intl": "^5.25.1",
     "react-select": "^5.3.1",
     "react-table": "^7.7.0",
+    "react-to-print": "^2.14.7",
     "react-toastify": "^9.0.1",
     "regenerator-runtime": "^0.13.9",
     "universal-router": "^9.1.0",
--- a/frontend/admin/src/js/adminRoutes.ts
+++ b/frontend/admin/src/js/adminRoutes.ts
@@ -63,6 +63,9 @@ const adminRoutes = (lang: string) => {
     searchRequestTable: (): string => path.join(home(), "search-requests"),
     searchRequestUpdate: (id: string): string =>
       path.join(home(), "search-requests", id, "edit"),
+
+    userAggregateDocument: (): string =>
+      path.join(home(), "user-aggregate-document"),
   };
 };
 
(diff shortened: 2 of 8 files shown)


petertgiles:
> * Chrome/Safari - Working great but page breaks are not confused

Ah, you're right!  I see it in Chromium on Linux, too.  Thanks for the docs, I'll work on getting it sorted.

petertgiles:
Hmm, it seems page break behavior is a bit squirrely.  I was never able to get Chromium to force a page break. :shrug: I settled for making sure it breaks between users which I could get to work the same way between Firefox and Chromium.  I also moved the styling inline.  I could see us wanting to use both methods - maybe shared print styling in the `pageStyle` prop pulled in from a shared location and document specific styling inline.

esizer:
> Hmm, it seems page break behavior is a bit squirrely. I was never able to get Chromium to force a page break. 🤷 I settled for making sure it breaks between users which I could get to work the same way between Firefox and Chromium. I also moved the styling inline. I could see us wanting to use both methods - maybe shared print styling in the `pageStyle` prop pulled in from a shared location and document specific styling inline.

Yeah, I have found that it is not very reliable. I think this should be sufficient since it is my understanding this is a temporary solution.

esizer (approved):
I think this looks good but I'd suggest leaving it up and let at least one other person take a look as well.

PR #2822, opened 2022-05-19
Component font consistency #2809

### LINKED ISSUE #2809: Component font consistency
There seems to be some font inconsistency between components. While "empty" and "filled" states might change in colour, it's best if they all have font type and size consistency.

Unsure which should be the base one so I'm tagging @JoshBeveridge on this one.

image

RVany:
Resolves #2809 
There seems to be some font inconsistency between components. While "empty" and "filled" states might change in colour, it's best if they all have font type and size consistency.

image

### DIFF
--- a/frontend/common/src/components/form/Input/Input.tsx
+++ b/frontend/common/src/components/form/Input/Input.tsx
@@ -59,6 +59,7 @@ const Input: React.FunctionComponent<InputProps> = ({
           data-h2-radius="b(s)"
           data-h2-border="b(darkgray, all, solid, s)"
           style={{ width: "100%" }}
+          data-h2-font-size="b(normal)"
           id={id}
           {...register(name, rules)}
           type={type}
--- a/frontend/common/src/components/form/Select/Select.tsx
+++ b/frontend/common/src/components/form/Select/Select.tsx
@@ -64,7 +64,7 @@ const Select: React.FunctionComponent<SelectProps> = ({
           )}
           {options.map((option) => (
             <option
-              data-h2-font-size="b(caption)"
+              data-h2-font-size="b(normal)"
               key={option.value}
               value={option.value}
             >
--- a/frontend/common/src/components/form/TextArea/TextArea.tsx
+++ b/frontend/common/src/components/form/TextArea/TextArea.tsx
@@ -46,6 +46,7 @@ const TextArea: React.FunctionComponent<TextAreaProps> = ({
           data-h2-padding="b(all, xxs)"
           data-h2-radius="b(s)"
           data-h2-border="b(darkgray, all, solid, s)"
+          data-h2-font-size="b(normal)"
           style={{ width: "100%", resize: "vertical" }}
           id={id}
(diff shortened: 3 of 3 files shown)


petertgiles:
@Jerryescandon do you want to get rid of the monospace font used in text boxes and date pickers, too?  Could you use the Chromatic _UI Review_ tool to approve or request changes on this PR?
image
image

Jerryescandon:
> do you want to get rid of the monospace font used in text boxes and date pickers, too?

Yeah fo sho. let's get them all to match.

RVany:
@petertgiles font family not included in any of the form component though! is it in any css file?🙂

petertgiles:
Looks fine to me but I'll leave it up to Jerbo to review.  I noticed the little calendar icon disappeared from the data field for me but that's probably a janky Linux font thing.  In Chromium it still looks fine, like in Chromatic.

RVany:
> Looks fine to me but I'll leave it up to Jerbo to review. I noticed the little calendar icon disappeared from the data field for me but that's probably a janky Linux font thing. In Chromium it still looks fine, like in Chromatic.

Thank you! I will check in my end and request review from Jerbo.

Jerryescandon (changes requested):
Date and Text fields look great.

Can we also update the size of the radio and checkboxes elements to match?
image

RVany:
> Date and Text fields look great.
> 
> Can we also update the size of the radio and checkboxes elements to match? 
Ok thank you, I will do that. What about other labels those look small for me.  Is that ok?
Screenshot 2022-05-20 154245

Jerryescandon:
> What about other labels those look small for me. Is that ok?

Yeah, labels are the only thing that's supposed to be smaller. So those are all good.

RVany:
> > What about other labels those look small for me. Is that ok?
> 
> Yeah, labels are the only thing that's supposed to be smaller. So those are all good.

ok thank you!

RVany:
> > What about other labels those look small for me. Is that ok?
> 
> Yeah, labels are the only thing that's supposed to be smaller. So those are all good.

@Jerryescandon radio and checkboxes elements are considered as labels so all the lebels are changing to same font and size.

Jerryescandon:
> @Jerryescandon Do you want me to bring the lebels back to privious format?

Yeah, I think so. I was thinking what would be the best thing to do in the meantime. At one point we need to separate those labels properly, but for now I would prefer all labels to be smaller. Even if the items in Radio and Checkboxes don't match yet.

RVany:
> > @Jerryescandon Do you want me to bring the lebels back to privious format?
> 
> Yeah, I think so. I was thinking what would be the best thing to do in the meantime. At one point we need to separate those labels properly, but for now I would prefer all labels to be smaller. Even if the items in Radio and Checkboxes don't match yet.

Ok. Thank you! I will do that. May be we need to do some major changes in form components.

Jerryescandon (approved):
Ship it.

tristan-orourke (approved):

PR #2827, opened 2022-05-20
Remove classification group constraint

### LINKED ISSUE #2806: Remove classification group constraint for salary-to-classification filter
From slack today:

> Peter
> I'm troubleshooting a search issue that Vachan found right now in the salary to classification query.  I noticed a quirk in the logic that I'm not sure is right.  Right now the logic requires you to have selected a classification in the searched group to do a salary-to-classification match.  So if you're searching for IT-01 the candidate will only be matched on salary if they included at least one IT classification in their profile.  Is this constraint correct or do we want to match on salary regardless of any classifications in their profile?
> 
> Based on Tristan's reply to Gray's question I'm pretty sure the constraint should not be there, right?
> 
> Tristan
> Yep, that does not sound right :confused:
> 

## Acceptance Criteria
- [ ] Ensure this fix applies to PoolCandidates and Users search
- [ ] PHPUnit test

petertgiles:
This branch removes the constraint that the candidate must have at least one attached classification of the group being search before they can be classification-to-salary matched.  Tests have been updated to exercise this behavior (or explicitly avoid it when not being tested).

Some queries I used for ad-hoc testing:
```
-- check salaries for IS 4 and 5
select * from classifications c 
where (c.group = 'IS' and c.level = 4) or (c.group = 'IS' and c.level = 5)

-- check candidates with expected salary ranges
select count(*) from pool_candidates pc 
where pc.expected_salary ?| array['_90_99K', '_100K_PLUS']
and "pool_id" in ('c2ed44d3-3bb9-4cf0-b99f-c961e0272779') 

-- check candidates without expected salary ranges but have classification matches
select count(*) from pool_candidates pc
join classification_pool_candidate cpc on pc.id = cpc.pool_candidate_id 
join classifications c on cpc.classification_id  = c.id
where not (pc.expected_salary ?| array[ '_90_99K', '_100K_PLUS'])
and "pool_id" in ('c2ed44d3-3bb9-4cf0-b99f-c961e0272779') 
and ( (c.group = 'IS' and c.level = 4) or (c.group = 'IS' and c.level = 5) )
```

Closes #2806

### DIFF
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -72,6 +72,11 @@ class PoolCandidate extends Model
 
     public function filterByClassifications(Builder $query, array $classifications): Builder
     {
+        // if no filters provided then return query unchanged
+        if (empty($classifications)) {
+            return $query;
+        }
+
         // Classifications act as an OR filter. The query should return candidates with any of the classifications.
         // A single whereHas clause for the relationship, containing multiple orWhere clauses accomplishes this.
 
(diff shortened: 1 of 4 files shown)


esizer (approved):
Looking good to me! 

- 🧪 Tests pass
- 🔧 Fixes [the bug](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2778#issuecomment-1129153479) @vd1992 found!

vd1992:
I watched this happen and can vouch for rigorously tested

PR #2832, opened 2022-05-20
Documentation - Environment Variables

petertgiles:
Hot off the press - environment variables.  Let me know if you find typos, things unclear, room for improvement...

### DIFF
--- a/documentation/environment-variables.md
+++ b/documentation/environment-variables.md
@@ -19,7 +19,7 @@ The frontend subprojects (admin, talentsearch, etc) have `.env` files that are c
 For deployment in production there needs to be a way to change variables in the program without rebuilding. This allows us to use environement variables as "feature flags", which enable/disable specific features without re-deployment. This is done using Apache Server Side Includes and the `config.sjs` files.  In a development environment these variables are injected into the docker container from the `/frontend/.apache_env` file.  To update these variables in the container a new container must be built.  For example:
 1) `docker-compose stop php`
 2) `docker-compose rm php`
-3) `docker-compose up -d`
+3) `docker-compose up --detach`
 
 To check if a particular environment variable is set in a container, a command like this could be used:
 `docker-compose exec php printenv FEATURE_APPLICANTPROFILE`
--- a/documentation/environment-variables.md
+++ b/documentation/environment-variables.md
@@ -1,0 +1,26 @@
+# Environment Variables
+
+Environment variables are created and used in many places throughout the repo.  This document covers the use of environment variables in a development environment.  Environment management in production is out of scope.
+
+## Laravel
+
+The Laravel projects (api and auth) read environment variables from the `.env` files in the root of their project directories.  These files are not checked into version control.  When running the setup scripts a new one is copied from the `.env.example` template and then customized.  In a development environment these files are reread on every new PHP request.  If needed a configuration reset can be forced with `php artisan config:clear`.
+
+## Frontend
+
+The frontend project uses environment variables in two distinct ways: build-time variables and run-time variables.
+
+### Build-Time Variables
+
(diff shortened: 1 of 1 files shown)


vd1992:
Very well-done. Really handy thing to document for sure. 
Can't think of anything else to address with that, no other forgotten steps or tips come to mind.

documentation/environment-variables.md
```diff
@@ -0,0 +1,26 @@
...
+
+Environment variables are created and used in many places throughout the repo.  This document covers the use of environment variables in a development environment.  Environment management in production is out of scope.
+
+## Laravel
+
+The Laravel projects (api and auth) read environment variables from the `.env` files in the root of their project directories.  These files are not checked into version control.  When running the setup scripts a new one is copied from the `.env.example` template and then customized.  In a development environment these files are reread on every new PHP request.  If needed a configuration reset can be forced with `php artisan config:clear`.
+
+## Frontend
+
+The frontend project uses environment variables in two distinct ways: build-time variables and run-time variables.
```
patcon, line 11:
```suggestion
The frontend project uses environment variables in two distinct ways: **build-time** variables and **run-time** variables.
```

patcon:
No specific micro-edits here, but advice I've received from friends who write policy whitepapers that are often skimmed, I find documentation can benefit a ton from:
- **bullets**. paragraphs are easy to get lost in. also more easily allows visual "diffing" of lines that say nearly the same thing, which i common in tech :)
- **bolding.** i often thing of this a meta-layer of the docs. like could someone ignore all the "fluff", and just read the bolding, and know the general content (and also know where to dive in deeper or slow down)

documentation/environment-variables.md
```diff
@@ -0,0 +1,26 @@
...
+
+The frontend React projects uses environment variables in two distinct ways: **build-time** variables and **run-time** variables.
+
+### Build-Time Variables
+
+The frontend subprojects (admin, talentsearch, etc) have `.env` files that are copied from `.env.example` template files like the Laravel projects.  When the frontend projects are built using Webpack, variables from the `.env` file are defined in the resulting bundle using DefinePlugin.  These values are "baked into" the bundle any if there are any variable changes the subproject must be rebuilt to reflect them.
+
+### Run-Time Variables
+
+For deployment in production there needs to be a way to change variables in the program without rebuilding.  This is done using Apache Server Side Includes and the `config.sjs` files.  In a development environment these variables are injected into the docker container from the `/frontend/.apache_env` file.  To update these variables in the container a new container must be built.  For example:
```
patcon, line 19:
```suggestion
For deployment in production there needs to be a way to change variables in the program without rebuilding. This allows us to use environement variables as "feature flags", which enable/disable specific features without re-deployment. This is done using Apache Server Side Includes and the `config.sjs` files.  In a development environment these variables are injected into the docker container from the `/frontend/.apache_env` file.  To update these variables in the container a new container must be built.  For example:
```

documentation/environment-variables.md
```diff
@@ -0,0 +1,26 @@
...
+1) `docker-compose stop php`
+2) `docker-compose rm php`
+3) `docker-compose up -d`
```
patcon, line 22: I believe this works the same: `docker-compose up --detach --build php`

Also, for your consideration: I've seen it advocated to avoid using short arg flags in both documentation and scripts, and prefer long form. The rationale being that the short flags are made for saving keystrokes when people know what they're doing, but documentation and scripts should always assume their intended audience is learners, for whom the long form helps make actions legible. (of course, always exceptions, but for me, blanket rules just feel simpler... rather than asking "am i expecting experts to be reading this?" all the time)

patcon (approved):
LGTM

documentation/environment-variables.md
petertgiles: Hmm, I just tried the one-liner and it appears to rebuild the image but doesn't recreate a running container.

I agree 100% about using the long-form for documentation and scripts though.  Good catch, I'll update that one.

vd1992:
per the devspam issue, this command
`docker-compose build php --no-cache`
might be something to add to the command list?

PR #2833, opened 2022-05-20
Create DataSeeder intended (and able) to run on UAT

### LINKED ISSUE #1763: Allow dev server to run data seeders
Seeders already exist. Need to ensure dev has the dev dependencies required to run the seeders.

tristan-orourke:
Closes #1763 

- Adds a new UserSeederUat class for where we can add users with subs specific to the UAT server.
- Adds PoolSeederUat which creates a digital_careers pool which should be identical to the one created on Prod. (This required adding a few more CmoAssets, for soft skills.)
- Modified ClassificationSeeder so it creates only realistic Classifications - ie the ones we created manually on Prod.
- Since none of the seeders called by UatSeeder use Model Factories or Faker, it can be run even when composer dev dependencies are not installed.

UatSeeder can be run with `php artisan db:seed --class=UatSeeder`.

This has been tested on the UAT server, and works!

### DIFF
--- a/api/database/seeders/PoolSeederUat.php
+++ b/api/database/seeders/PoolSeederUat.php
@@ -1,0 +1,58 @@
+<?php
+
+namespace Database\Seeders;
+
+use App\Models\Classification;
+use App\Models\CmoAsset;
+use App\Models\Pool;
+use App\Models\User;
+use Illuminate\Database\Seeder;
+use Database\Helpers\ApiEnums;
+
+class PoolSeederUat extends Seeder
+{
+    /**
+     * Run the database seeds.
+     *
+     * @return void
+     */
+    public function run()
+    {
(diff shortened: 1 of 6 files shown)


api/database/seeders/PoolSeederUat.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+namespace Database\Seeders;
+
+use App\Models\Classification;
+use App\Models\CmoAsset;
+use App\Models\Pool;
+use App\Models\User;
+use Illuminate\Database\Seeder;
+use Database\Helpers\ApiEnums;
```
petertgiles, line 10: Unused reference?

petertgiles (approved):
Very nice!

api/database/seeders/PoolSeederUat.php
tristan-orourke: Thank!

PR #2854, opened 2022-05-23
Consolidate typescript config files

patcon:
This does a few small house-keeping things to ensure our config stays aligned between project paths

Lingo:
- [frontend] base dir: `frontend/`
- [frontend] app dirs: `frontend/admin`, `frontend/talentsearch`, `frontend/indigenousapprenticeship`
- [frontend] common dir: `frontend/common`

## To Do
- [x] aligned config amongst app dirs (only tiny inconsequential changes)
- [x] consolidated app dir config into single file in frontend `tsconfig.base.json`
- [x] added minimal `extends` config back into app folders (this allows to override with a smaller subset of `include`, so doesn't run slower in app dirs)
- [x] set common dir config to extend base config, removing redundant settings
  - [x] confirm that these are important overrides to preserve and not revert -- @yonikid15 (from `git blame`)
- [x] added npm run script commands for typescript checking `npm run tsc`
  - in base, runs over whole frontend (single run, faster than individual) `cd frontend && npm run tsc`
  - in app dirs, runs for each project: `cd frontend/common &&  npm run tsc`
- [x] tell vscode where our workspace typescript compiler is, so it can use proper version by default [docs](https://code.visualstudio.com/docs/typescript/typescript-compiling#_using-the-workspace-version-of-typescript)
- [x] confirm everything still works fine in vscode -- vscode pro user (patcon did a cursory check)


patcon:
@yonikid15 specifically wondering why all the unique settings in here are needed:

https://github.com/GCTC-NTGC/gc-digital-talent/pull/2854/files#diff-b48a9107346b5ac17ccc767d5960ff8b50be392187a799cc917520eb10cd0b01

(compared to the much simpler and shared settings/defaults in `admin` `talentsearch` etc)

patcon:
Yoni said the customizations weren't important, so merging

PR #2872, opened 2022-05-25
Pool filter includes expired talent

### LINKED ISSUE #2810: Talent passed their expiry date in the pool should not show up in search results
## AC
- [x] apply this change to User and PoolCandidate search
- [x] phpunit tests

JamesHuf:
Resolves #2810.

Adds the viewExpired scope to various Pool Candidate queries to filter for only not expired candidates:
- poolCandidates
- countPoolCandidates
- searchPoolCandidates (can optionally filter for only expired candidates instead)

Updated the Pool filter for the usersPaginated query to take in a list of Pool IDs to filter for and optionally viewExpiredCandidates (defaults to False).


api/app/Models/User.php
```diff
@@ -174,8 +174,15 @@ public function filterByPools(Builder $query, array $pools): Builder
...
                   ->from('pool_candidates')
                   ->whereColumn('pool_candidates.user_id', 'users.id')
-                  ->whereIn('pool_candidates.pool_id', $pools);
+                  ->whereIn('pool_candidates.pool_id', $pools['ids']);
+
+            if (isset($pools['viewExpiredCandidates']) && $pools['viewExpiredCandidates']) {
+                $query->whereDate('expiry_date', '<', date("Y-m-d"));
+            } else {
+                $query->whereDate('expiry_date', '>=', date("Y-m-d"));
+            }
```
petertgiles, line 183: Again, null handling

api/graphql/schema.graphql
```diff
@@ -513,6 +513,11 @@ input PoolFilterInput {
     id: ID!
 }
 
+input UserPoolFilterInput {
+    ids: [ID!]!
+    viewExpiredCandidates: Boolean
```
petertgiles, line 518: It seems strange to me that the parameter is added here.  Why isn't it on PoolCandidateFilterInput like `isWoman` and the other booleans?

api/tests/Feature/PoolCandidateTest.php
```diff
@@ -836,5 +837,79 @@ public function testFilterByClassificationToSalaryWithPools(): void
...
+
+    // Create some valid users
+    $futureCandidates = PoolCandidate::factory()->count(4)->create([
+      'expiry_date' => '3000-05-13',
+    ]);
+    $todayCandidate = PoolCandidate::factory()->create([
+      'expiry_date' => date("Y-m-d"),
+    ]);
+    $futureCandidates->concat($todayCandidate);
+
```
petertgiles, line 863: Could you add some users with null expiry date?

api/app/Models/PoolCandidate.php
```diff
@@ -269,6 +269,13 @@ public function scopeIsWoman(Builder $query, bool $isWoman): Builder
         }
         return $query;
     }
-
-
+    public function scopeViewExpired(Builder $query, ?array $args) {
+        $viewExpired = isset($args['viewExpired']) ? $args['viewExpired'] : false;
+        if ($viewExpired) {
+            $query->whereDate('expiry_date', '<', date("Y-m-d"));
```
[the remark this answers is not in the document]
JamesHuf: I forgot `expiry_date` is nullable, I'll do that.

api/graphql/schema.graphql
petertgiles: Oh I see, the concept of being expired only exists in the context of a specific pool.  I wonder if we could name this better in case someone else is confused like me.  

Naming is hard!  Maybe we can change `ids` to `poolIds` since this isn't an input just for `pool` anymore.

Maybe `isPoolCandidateExpired` for the other parameter is a little more consistent with the other fields?

It's a little weird that this filter works differently from the other boolean filters.  For example, "is_woman = false" is the same as "is_woman is null".  In both cases you get all candidates.  For this filter, "viewExpiredCandidates = false" gets you all the actives only, right?  Do we want to change this to match the behavior of the other filters?  Return everybody unless the filter exists and is set to true?  I guess I would call it `includeOnlyActive` then?

I think I'm going to ask Tristan for a quick design meetup today in scrum.

JamesHuf: So I've changed `ids` to `pools` for UserPoolFilter input but I still don't like it because it leaves us with queries like this:
```
'where' => [
    'pools' => [
        'pools' => [$myPool->id],
        'expiredStatus' => ApiEnums::CANDIDATE_EXPIRY_FILTER_ACTIVE
    ]
]
```
I'm considering renaming the outer `pools` to something a little more accurate but I'm not sure what - `poolsAndStatus` maybe. Any other thoughts on naming improvements?

Also I've added the `CANDIDATE_EXPIRY_FILTER` enum and updated the user query with it. I'm currently switching the Pool Candidate queries to use it as well.

JamesHuf: Ok, Pool Candidate queries are changed over to use the `CANDIDATE_EXPIRY_FILTER` enum as well.

petertgiles: Yeah, I agree, that query is ugly with the two pools.  I also agree that changing the outer input is the way to go.  `PoolAndStatus` would become inaccurate as soon as we add anything else.. Maybe `poolCandidates`?  These are all filters in the pool_candidate table context, right?

petertgiles: I noticed a weird quirk while manually testing.  It seems like it's returning the filtered list of users but the paginator is pulling in everyone.  Is the paginator bypassing the contexts?
image

JamesHuf: I played around a bit myself but haven't been able to replicate your bug. Refreshing my environment didn't change anything. Would you mind taking another look and seeing if it's still happening for you?

Also in case it helps:
- This is the paginated query which defaults to only 10 results per page so that might be what made you think the list of users was filtered.
- The Pool Candidate factory currently only generates valid users so unless you change things the `ACTIVE` and `ALL`  statuses will both return all users in the pool while `EXPIRED` will not return anyone unless you change things.

vd1992: My attempt

image

vd1992: Yup, there is a default to 10 

image

JamesHuf: > My attempt

That looks right. The seeder generates 71 users, 2 of which are pool managers and the rest are active in one of those two pools.

petertgiles: So, the trouble here is that our seeder only seeds active pool candidates. :cry: 
```
select expiry_status, count(*) from (
  select
    case
      when expiry_date > current_date then 'active'
      when expiry_date is null then 'null'
      when expiry_date <= current_date then 'expired'
    end as expiry_status
  from pool_candidates pc
  where pool_id = '1a81dc03-32da-45be-9b49-d4d7da996058'
) t
group by expiry_status
```
image
I created #2894 to improve that.
So I manually fixed that in my workspace.  Here's a gnarly update statement:
```
UPDATE pool_candidates
SET expiry_date=u.new_expiry_date
FROM (
  select
    id,
    case
      when row_num < 20 then '1970-01-01'
      when row_num < 40 then null
      else expiry_date
    end as new_expiry_date
  from (
    select
      id,
      expiry_date,
      row_number() over (order by id) row_num 
    from pool_candidates pc
  ) as t
) AS u
WHERE pool_candidates.id=u.id;
```
That looks better:
image
Unfortunately, I'm seeing everyone in the paginator, all statuses and pools.  Also, the paginator doesn't seem to be limiting page size or page, either?
image

```
query q($where: UserFilterAndOrderInput, $first: Int, $page: Int) {
  usersPaginated(where: $where, first: $first, page: $page) {
    data {
      firstName
    }
    paginatorInfo {
      count
      currentPage
      hasMorePages
      perPage
      total
    }
  }
}
```
```
{
  "where": {
    "poolCandidates": {
      "pools": ["1a81dc03-32da-45be-9b49-d4d7da996058"],
      "expiryStatus": "ACTIVE"
    },
  },
  "first": 3,
  "page": 2
}
```

petertgiles (approved):
Amazing work, as always!  Thanks for sticking through my bumbling review. :smile:

PR #2874, opened 2022-05-25
placeholder user page

### LINKED ISSUE #2658: Create Placeholder User page
Screen Shot 2022-05-04 at 11 37 09 AM

Build first two tabs (NOT REQUEST HISTORY), but no content in the tabs.

Place the Download button, but no function.

vd1992:
resolves #2658 

Used the placeholder pool details component as a template


frontend/admin/src/js/components/user/ViewUser.tsx
```diff
@@ -0,0 +1,157 @@
...
+            }}
+          >
+            <span>
+              <DownloadIcon style={{ width: "1rem" }} />{" "}
+              {intl.formatMessage({
+                defaultMessage: "Download Profile",
+                description: "Text for button to download a user",
+              })}
+            </span>
+          </Button>
```
yonikid15, line 89: From a accessibility standpoint I'm pretty sure it's better to use a <Link /> in this situation. It should also include the [download attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download), which can be set to something like ` download="<username>-profile.<filetype>"` (but for now leaving it blank is fine since this is just the placeholder issue). The href can also be set to `href="#"` for now.

frontend/admin/src/js/components/user/ViewUser.tsx
```diff
@@ -0,0 +1,157 @@
...
+        defaultMessage: "Users",
+        description: "Breadcrumb title for the all users table link.",
+      }),
+      href: adminPaths.userTable(),
+      icon: <UserCircleIcon style={{ width: "1rem", marginRight: "5px" }} />,
+    },
+    {
+      title: userName,
+    },
+  ] as BreadcrumbsProps["links"];
```
yonikid15, line 53: Are these links supposed to match the breadcrumbs in the wireframe? Or am I looking at an outdated picture?
Screen Shot 2022-05-04 at 11 37 09 AM

yonikid15 (changes requested):
Looking good! Just a couple things left. Ones possibly just a question.

frontend/admin/src/js/components/user/ViewUser.tsx
vd1992: yeah, we were bouncing around on this, so far sticking with what we already have as a component for the time being I believe

frontend/admin/src/js/components/user/ViewUser.tsx
vd1992: check

yonikid15 (approved):
LGTM 👍

Issue #2875, opened 2022-05-25
Upgrade Storybook from 6.4 to 6.5


petertgiles:
When we upgraded to storybook 6.5 it stopped building correctly.  It seems like there are webpack errors now.
```
ℹ Storybook build output:
/home/runner/work/gc-digital-talent/gc-digital-talent/frontend/admin/build-storybook.log

info @storybook/react v6.5.5
info 
info => Cleaning outputDir: /tmp/chromatic--2118-LPsax8hQPlt0
info => Loading presets
info => Copying static files: /home/runner/work/gc-digital-talent/gc-digital-talent/frontend/admin/src at /tmp/chromatic--2118-LPsax8hQPlt0/admin
info => Copying static files: /home/runner/work/gc-digital-talent/gc-digital-talent/frontend/talentsearch/src at /tmp/chromatic--2118-LPsax8hQPlt0/talent
info => Copying static files: /home/runner/work/gc-digital-talent/gc-digital-talent/frontend/indigenousapprenticeship/src at /tmp/chromatic--2118-LPsax8hQPlt0/indigenous-it-apprentice
info => Compiling manager..
info => Compiling preview..
info => Using implicit CSS loaders
info => Using default Webpack5 setup
ERR! => Failed to build the preview
ERR! NormalModuleFactory.beforeResolve (IgnorePlugin) is no longer a waterfall hook, but a bailing hook instead. Do not return the passed object, but modify it instead. Returning false will ignore the request and results in no module created.
WARN Unable to close preview build!
ERR! undefined
info => Loading presets
attention => Storybook now collects completely anonymous telemetry regarding usage.
This information is used to shape Storybook's roadmap and prioritize features.
You can learn more, including how to opt-out if you'd not like to participate in this anonymous program, by visiting the following URL:
https://storybook.js.org/telemetry
    → Command failed: npm run --silent build-storybook -- --output-dir /tmp/chromatic--2118-LPsax8hQPlt0
```

tristan-orourke:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/GC-Digital-Talent-6092dc05157fa30010edaef9/issues/GCTC-NTGC/gc-digital-talent/2875?planning-poker) with ZenHub @JamesHuf

vd1992:
https://github.com/GCTC-NTGC/gc-digital-talent/pull/2905#issuecomment-1167935754

vd1992:
[...]

PR #2876, opened 2022-05-25
Rollback storybook 6.5

petertgiles:
Storybook 6.5 seems to have some webpack issues in our environment.  I've opened #2875 to take a closer look but in the meantime I've opened this PR to roll back the change.


vd1992:
seems to be a thing for multiple people
https://github.com/storybookjs/storybook/issues?page=2&q=is%3Aissue+is%3Aopen

vd1992:
counted like 10 issues related to Storybook not working proper after updating
one of them might be similar to ours, webpack one maybe, in any case can rule out it being a problem in your environment or on our half of the turf

petertgiles:
The snapshotting is interesting, too.  It looks like it's picking up all of Vany's font changes and the error messages became more verbose.  :shrug:

vd1992 (approved):
all appears good, per slack convo

PR #2877, opened 2022-05-25
Bug - 2693 - Allow Users to Delete Skills

### LINKED ISSUE #2693: Can't remove skills from experience
Currently it is not possible to remove skills from an experience.  The UI will allow the user to remove them but when the change is saved the mutation is executed with all the removed skills still there.  The form is failing to send the mutation with the skill removed.

This bug is proving to be very tricky to solve.  Right now ExperienceForm has a child component ExperienceSkills that manages adding and removing skills.  ExperienceSkills takes in its props a list of all skills and a list of initial skills then maintains its own state and changes the form by itself.  Possibly the solution is to instead have ExperienceSkills expose `onAddSkill` and `onRemoveSkill` and allow its parent to manage the state.

esizer:
Resolves #2693 

## Summary

This updates the `<ExperienceForm>` component to use the `useFieldArray` hook form `react-hook-form` to handle the skills array. Previously, we were manually updating using `setValue` which was causing problems when attempting to remove items.

## Testing

1. Navigate to the frontend directory, `cd frontend`
2. Build the talent search workspace `npm run production --workspace=talentsearch`
3. Navigate to the [Skills and Experiences](http://localhost:8000/en/talent/profile/skills-and-experiences) page.
4. Edit an existing experience
5. Test adding and removing skills
  - Add with the **Add Skills** section
  - Remove skills from
    - Chips under **Skills** attached to this experience
    - List under **Add Skills** section
    - Using the **Remove from experience** button in **Skills in Detail** section

All of the listed ways to remove a skill should remove it from the interface and send the proper mutation variables to execute the sync.

## Note

While working on this PR, I noticed that the `useForm` hook should be avoided when already within a form context. If you are already within that context, we should be using `useFormContext` in-place of that to get the existing form context. `useForm` seems to create a new context  which can cause problems.


vd1992 (changes requested):
### General 
_Relevant_
Request for keyprops as shown, appears when loading /edit when atleast one skill has been selected

image

### Accessibility 
_Out of scope?_
I took screenshots of the Lighthouse report, included
Multiples of these originate from components generated elsewhere, but given this is a page that uses many components it makes it easier to collate issues here. A bit out of scope here, but maybe something to spin-off down the line. Or maybe to be ignored if another tool is added to the Github Workflow. 

Your call really, but I wouldn't worry about it in this issue but just keep what the alerts say in mind. 

image

image

image

image

vd1992:
tried to push things through its paces manually, dunno if you want to add any new tests for this, whether they are even needed
so far, looks great, amazing!

- can create experience
- can update experience
- can delete experience
- can delete skills, save, skills adjusted accordingly (issue resolved)
- refresh form holds data

not sure what else to throw at it

esizer:
> _Out of scope?_
> I took screenshots of the Lighthouse report, included

These were simple enough to fix outside of the checkbox so I threw them in. Just updated some roles and adjusted the heading levels.

frontend/talentsearch/src/js/components/skills/AddedSkills/AddedSkills.tsx
```diff
@@ -19,25 +19,29 @@ const AddedSkills: React.FunctionComponent<AddedSkillsProps> = ({
...
-        );
-      })}
+      </h3>
+      {skills.length ? (
+        <Chips>
+          {skills.map((skill) => {
+            const handleDismiss = () => onRemoveSkill(skill.id);
+            return (
+              <Chip
+                key={skill?.id}
```
petertgiles, line 35: I'm getting a
```
 Warning: Each child in a list should have a unique "key" prop.
Check the render method of `AddedSkills`.
```
on the skill edit page.  Is it because this is possibly null?

petertgiles (approved):
Oh yeah, this is way nicer.  Nice fix and refactor! 💪

frontend/talentsearch/src/js/components/skills/AddedSkills/AddedSkills.tsx
esizer: So, this is the annoying bit. I updated this,`skill.id -> skill?.id` because when testing, we generate mock skills and experiences separately. This results in the `id` for skills not matching the `id` for skills attached to experiences.

It was actually breaking because when we pass in the skills that are added, they don't match the available skills and we get a "cannot access property `id` of `undefined`. This breaks the tests, adding the `?.id` throws some weird errors but does not break the tests.

I think we should create a ticket to update the `fakeExperiences` to allow us to pass in valid skills to be used in the generation. Once that is updated, we can revert this back to remove the warning.

vd1992 (approved):
Remaining issue to be addressed elsewhere. 
Great work!

petertgiles (approved):
Looks nice!

PR #2884, opened 2022-05-26
Adding accessibility testing

mgifford:
Obviously there is more that needs to be done than adding to a checklist, but thought it should be added here.


patcon:
Thanks @mgifford! Unless someone corrects me, I believe we have these conversations about improving our "definition of done" during our biweekly retrospectives, so we'll prob discuss putting things in place for this at that time 🙌

petertgiles:
Adding a11y steps to definition of done agreed upon as action item for this sprint.  Merging this in as a starting point.

petertgiles (approved):

PR #2888, opened 2022-05-26
hack: get FR iap story working for preview

esizer:
## Summary

Quick hack so Gray can share the FR version of IAP.

**Note**: Need to properly get locale working in storybook in the future.


esizer:
Closed with new issue to fix #2895

Issue #2873, opened 2022-05-25
API - File name casing bug


petertgiles:
Composer is complaining about the casing of the DatasetInferface file.  I think it's not causing any problems, normally.  It does seem to be causing fatal errors during testing on Patrick's laptop but I'm not clear on why.

![image.png](https://images.zenhubusercontent.com/6172cf6e898e11fd20507433/1cde4bc9-ccb9-48cf-ac66-d03b1fa5f4bb)

patcon:
Ah, sneaky, I didn't spot that case difference :)

PR #2889, opened 2022-05-26
Fixed phpunit tests for DataSet

patcon:
Resolves https://github.com/GCTC-NTGC/gc-digital-talent/issues/2873

Not sure why this was a quiet warning for everyone else and fatal errror for me. On investigating, it seems that the official mockery docks say that final classes can't be mocked, and they provide an alternative:
https://docs.mockery.io/en/latest/reference/final_methods_classes.html#dealing-with-final-classes-methods

I've swapped to using proxy mocks, and my phpunit tests are running now :)

### DIFF
--- a/api/app/Services/Contracts/DatasetInterface.php
+++ b/api/app/Services/Contracts/DatasetInterface.php
@@ -1,9 +1,0 @@
-<?php
-
-namespace App\Services\Contracts;
-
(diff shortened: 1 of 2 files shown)


tristan-orourke (approved):
@patcon if this works for you, and the PHPUnit workflow still passes, I'm happy

Issue #2896, opened 2022-05-27
Add Auth to TalentSearch


petertgiles:
We have our auth containers available in common now.  It would be a good time to enable auth in talentsearch.  We need auth there for all the profile and direct intake stuff.  Eric made a nice context component for admin with our "container sandwich" inside that could be reused.  I guess we would need log in and log out links?

# Acceptance Critera
- [ ] Root component in talentsearch includes AuthContainer 
- [ ] Profile page and direct intake routes require logged-in and APPLICANT role
- [ ] Cypress test for user without this role denied access profile and direct intake pages?

petertgiles:
Do we want to do it for IAP at the same time?  It's not really need there yet.

petertgiles:
Should we comment out the default AUTH_DEFAULT_USER as admin in the .env.example file after this?  It will be confusing to revert to an admin user when testing applicant pages.

tristan-orourke:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/GC-Digital-Talent-6092dc05157fa30010edaef9/issues/GCTC-NTGC/gc-digital-talent/2896?planning-poker) with ZenHub @esizer

Issue #2868, opened 2022-05-24
Admin - Flash of "Not Authorized" on reload


petertgiles:
Whenever I refresh the page while logged in I get a brief flash of the "Not Authorized" error message and then the page loads.  It doesn't actually prevent me from doing anything but I imagine a regular user would find it a bit disconcerting.  I also imagine the message would stay on the screen longer for slower connections.
![Peek 2022-05-24 13-29.gif](https://images.zenhubusercontent.com/6172cf6e898e11fd20507433/5493cd6f-b072-4f31-89e7-50b5a70a4235)

patcon:
Yeah, I noticed this, and thought maybe it was just noticeable for me due to the slowness of my local queries (though maybe comparable to a slow internet connection?)

Related, I was looking into how `auth0-react` handles its login state via it's useAuth0 hooks, as I would assume their patterns are pretty safe and worth emulating. They've got quite a few auth flows accommodated in their libs, so it's a bit to untangle. Didn't finish grokking, but I really love their component and hook APIs :)

They seem to manage various pieces of state (secure and insecure) in a few places, not just one: incl sessions, local storage, cookies, and in-memory cache, and even using web workers (*see below for neat trivia on this last thing). It's hard to tease apart which tactics are used together at the same time. 

But one thing I noticed is that they store a simple insecure cookie for `auth0.xxxxxx.isLoggedIn=true` that I suspect might deal with this flicker (and is removed if tokens stop validating)

The have a client object stored in their AuthContext that seems to handle all the complicated work.

Anyhow, it was enlightening looking at their library a bit, and felt neat to get potential approaches from them

---

Also, wondering if the useSWR hook (that I mentioned in #devspam on Slack) could address this for us. It aggressively caches responses based on a unique cached key (string, array, obj), and they serves cache first, then updates when the new response arrives. If we piped our graphql fetcher functions through it, and let it cache responses based on the data POSTed to /graphql, it could reduced our having to load our UI in cascades of renders with the various graphql calls :thinking_face:

(*re: web workers. Apparently web workers run in an inaccessible background process that the browser JS can only interact with through a developer-defined postMessage interface. And so it's a good place to hide refresh_tokens -- the worker can keep refreshing and passing the access_token forward. Even if JS code or a browser ext is compromised, it can only get access_tokens, not the longer-lived refresh_token. But I digress...!)

petertgiles:
[...]

Issue #2135, opened 2022-02-22
Logout landing page


petertgiles:
When the user is logged out of our auth provider they are redirected back to a whitelisted URL.  Do we want a *You have been logged out* landing page for them to come back to?

petertgiles:
Comments from standup:
- Since the redirection is happening with browser navigation (not javascript) it has to be a real URL
- We can have SiC whitelist as many URLs as we want so we can have multiple landing pages if desired
- With my current PR I'm redirecting back to the regular /admin page which is fine for now

tristan-orourke:
@JoshBeveridge could you add your prototype here?

tristan-orourke:
Will close this in favour of other tickets coming soon!

PR #2898, opened 2022-05-27
Deep Linking

### LINKED ISSUE #2784: Deep Linking (Post Login Redirect)
If a user navigates to a protected page they should automatically be redirected to log in and then redirected to the desired page after successful log in.  They should not just get the "Unauthorized" error.

Previous discussion about this existed in #1988

## User Acceptance
- [ ] Clicking a link to a protected page while logged in behaves normally 
- [ ] Clicking a link to a protected page while not logged in redirects to login
- [ ] After successfully logging in the user is redirected to the protected page
- [ ] If login is unsuccessful, or they do not have correct role permissions, they should get the regular "Unauthorized" error
- [ ] New redirect behavior also works when entering the URL directly (not navigating within React)
- [ ] Set up new behavior for protected pages on Admin, Profiles, and Pools pages

petertgiles:
This branch enables deep linking - linking to resources that are protected.  It will automatically send the user to login if they attempt to navigate to a protected page.

What to check:

- [ ] Clicking a link to a protected page while logged in behaves normally
- [ ] Clicking a link to a protected page while not logged in redirects to login
- [ ] After successfully logging in the user is redirected to the protected page
- [ ] If ~login is unsuccessful, or~ they do not have correct role permissions, they should get the regular "Unauthorized" error
- [ ] New redirect behavior also works when entering the URL directly (not navigating within React)
- [ ] Set up new behavior for protected pages on Admin, ~Profiles, and Pools pages~
- [ ] Talentsearch and IAP behavior should be unchanged

Caveats:
- When navigating there is a flash of "not authorized" before the contents loads - #2868
- Only admin is is scope for this - #2896
- There is no unprotected logout landing page so after logging out the user will loop right back to login - #2135 (but closed now?)
- There is only one role so to test out "unauthorized" you can remove the ADMIN role from your user
- If login is unsuccessful GCKey returns the user to `/` which is pretty good, I think
- If you want to truly log out of GCKey you'll need to enable the run-time variable `OAUTH_LOGOUT_URI` in `frontend`

Closes #2784

### DIFF
--- a/frontend/indigenousapprenticeship/src/js/components/PageContainer.tsx
+++ b/frontend/indigenousapprenticeship/src/js/components/PageContainer.tsx
@@ -101,8 +101,6 @@ export const PageContainer: React.FC<{
     contentRoutes,
     notFoundComponent.current,
     notAuthorizedComponent.current,
-    false,
-    null,
   );
   return (
     <ScrollToTop>
--- a/frontend/talentsearch/src/js/components/PageContainer.tsx
+++ b/frontend/talentsearch/src/js/components/PageContainer.tsx
@@ -104,8 +104,6 @@ export const PageContainer: React.FC<{
     contentRoutes,
     notFoundComponent.current,
     notAuthorizedComponent.current,
-    false,
-    null,
   );
   return (
     <ScrollToTop>


frontend/admin/src/js/components/PoolDashboard.tsx
```diff
@@ -37,291 +35,224 @@ import { UpdateSkillFamily } from "./skillFamily/UpdateSkillFamily";
...
-      redirect: loggedIn ? paths.dashboard() : undefined,
+      redirect: paths.dashboard(),
     }),
   },
   {
     path: paths.dashboard(),
     action: () => ({
-      component: loggedIn ? <DashboardPage /> : <AdminNotAuthorized />,
+      component: <DashboardPage />,
+      authorizedRoles: [Role.Admin],
```
petertgiles, line 53: How do people feel about defining the roles authorized to use the component right here?

frontend/common/src/helpers/router.tsx
```diff
@@ -87,35 +98,84 @@ export const pushToStateThenNavigate = (url: string, state: State): void => {
...
   missingRouteComponent: ReactElement,
+  notAuthorizedComponent: ReactElement,
+  isLoggedIn: boolean,
+  loggedInUserRoles: PossibleUserRoles,
 ): React.ReactElement | null => {
   const location = useLocation();
   const router = useMemo(() => new UniversalRouter(routes), [routes]);
   const [component, setComponent] = useState<React.ReactElement | null>(null);
   const pathName = location.pathname;
+  const apiRoutesRef = useRef(useApiRoutes()); // I'm not not sure why this function is unstable on its own!
```
petertgiles, line 117: Would love it if anyone could help me understand why this function was unstable.  If it is added directly to the useEffect dependencies it causes render loops.

frontend/common/src/helpers/router.tsx
```diff
@@ -87,35 +98,84 @@ export const pushToStateThenNavigate = (url: string, state: State): void => {
...
+        // may or may not be a promise, so attempt to resolve it. A non-promise value will simply resolve to itself.
+        const route = await Promise.resolve(routeMaybePromise);
+
+        // handling a redirect
+        if (route?.redirect) {
+          redirect(route.redirect);
+        }
+
+        // is authorization required for this route?
+        const authorizationRequired = route?.authorizedRoles;
```
petertgiles, line 133: Right now, the logic says that if they use needs to be logged in simply use an empty array.  A null means login not needed.  It's a bit "magicky" but I didn't really want to add another prop for this.  It's a good argument for adding a role like `USER` that everyone gets to test for being a logged in user.

frontend/admin/src/js/components/PoolDashboard.tsx
esizer: I personally like this. Not sure if there is even a better alternative outside of defining it in the component itself. 🤷

frontend/common/src/helpers/router.tsx
esizer: There is an unnecessary nesting of functions. I'm not sure why this causes infinite re-renders but it certainly adds complexity. Since `apiRoutes` is static and takes no arguments, its safer to declare it as an object rather than a function that returns an object. Try this:

```tsx
import path from "path-browserify";

export interface ApiRoutes {
  login: (from?: string, locale?: string) => string;
  refreshAccessToken: () => string;
}

const apiRoot = (): string => "/";

const apiRoutes = {
  login: (from?: string, locale?: string): string => {
    const searchTerms: string[] = [];
    if (from) searchTerms.push(`from=${encodeURI(from)}`);
    if (locale) searchTerms.push(`locale=${encodeURI(locale)}`);
    const searchString = searchTerms.join("&");

    const url =
      path.join(apiRoot(), "login") + (searchString ? `?${searchString}` : "");

    return url;
  },
  refreshAccessToken: (): string => path.join(apiRoot(), "refresh"),
};

/**
 * A hook version of loginRoutes which gets the locale from the intl context.
 * @returns LoginRoutes
 */
export const useApiRoutes = (): ApiRoutes => {
  return apiRoutes;
};

```

frontend/admin/src/js/components/dashboard/Dashboard.tsx
```diff
@@ -66,15 +90,28 @@ const OpenMenuButton: React.FC<OpenMenuButtonProps> = ({
 
 interface DashboardProps {
   contentRoutes: Routes<RouterResult>;
+  loggedIn: boolean;
```
petertgiles, line 93: All these props are just for the router.  I wonder if the `useRouter` call with its stuff should get lifted to the parent and just pass in `content`.

frontend/common/src/helpers/router.tsx
petertgiles: Yep, that was totally it.  Yer a wizard, @esizer!  :mage:

frontend/common/src/helpers/router.tsx
```diff
@@ -1,9 +1,20 @@
 import { createBrowserHistory, Location, Path, State } from "history";
 import UniversalRouter, { Routes } from "universal-router";
-import React, { useState, useEffect, useMemo, ReactElement } from "react";
+import React, {
+  useState,
+  useEffect,
+  useMemo,
+  ReactElement,
+  useRef,
```
esizer, line 8: Looks like we can removed the `useRef` import now that it is not wrapping the `useApiRoutes`.

frontend/common/src/helpers/router.tsx
esizer: Yeah, I agree. Adding some basic role like `USER` would reduce confusion on `null` vs `[]`.

esizer (changes requested):
This is amazing! Loved reviewing this one and learning more about how our router works.

Everything looks good except for a simple unused import 👍

frontend/common/src/helpers/router.tsx
petertgiles: Alternatively, remove this logic and use it as a pain point.  Anytime you have a page that needs a logged in user, use that as the opportunity to declare a new role.  For example, we need a role for CANDIDATE/APPLICANT/JOBSEEKER for the direct intake and profiles pages.

frontend/common/src/helpers/router.tsx
petertgiles: Good catch!

frontend/admin/src/js/components/dashboard/Dashboard.tsx
esizer: Just saw this, I tend to agree. I'd like to see if someone with more familiarity could weigh in. To me this makes sense so we can avoid excessive prop drilling.

esizer (approved):
Responded to a few comments but they seem to be out of scope for this PR.

Manually tested all scenarios and working as expected. Awesome! 🎉

frontend/admin/src/js/components/PoolDashboard.tsx
patcon: i like it too. from doing a few react-router tutorials in the past, this seems to be the common place to specify :)

frontend/common/src/helpers/router.tsx
tristan-orourke: For a bit more context, while it may be unnecessary for api routes, the talentSearchRoutes and adminRoutes use hooks to get the current locale from react-intl context, and inject it into the routes.

frontend/common/src/helpers/router.tsx
tristan-orourke: I don't like the idea of a generic `USER` role, as it would be entirely redundant. But I'm happy with adding a `CANDIDATE/APPLICANT` role, and try to keep it so that any page requiring login requires a role.

frontend/admin/src/js/components/PoolDashboard.tsx
tristan-orourke: Makes sense to me!

frontend/admin/src/js/components/PoolDashboard.tsx
```diff
@@ -330,11 +270,13 @@ export const PoolDashboard: React.FC = () => {
   const { loggedIn } = React.useContext(AuthenticationContext);
   const paths = useAdminRoutes();
   const { loggedInUserRoles } = React.useContext(AuthorizationContext);
```
tristan-orourke, line 272: Huh, we're getting the same context twice, that can be a single line.

frontend/admin/src/js/components/dashboard/Dashboard.tsx
tristan-orourke: Actually, why not just let useRouter access the AuthorizationContext and AuthenticationContext itself?

tristan-orourke (commented):
My one suggestion is to let useRouter access the auth contexts itself, instead of passing in logged-in status from the parent component.

frontend/common/src/helpers/router.tsx
petertgiles: Sounds good.  #2911 created.

frontend/admin/src/js/components/dashboard/Dashboard.tsx
petertgiles: Smart!  Added in bd1fb90a461ed6e17f57cffb07d6a931b097dbd2 and I like it much better.

tristan-orourke (approved):

PR #2908, opened 2022-05-30
Debt - 2771 - Ignore console.error for integrity tests

### LINKED ISSUE #2771: Jest - integrity check is noisy
The jest tests for integrity check are pretty noisy due to console logging.  It looks like there are a few ways to silence them. https://stackoverflow.com/questions/44467657/jest-better-way-to-disable-console-inside-unit-tests

esizer:
Resolves #2771 

## Summary

This updates the jest config to mock the `console.error` function globally to prevent the `checkIntegrity.test.js` from being too noisy muddying up the tests. It also makes tests verbose so we have better knowledge of the tests we are running.

h/t @petertgiles 

Ref: https://stackoverflow.com/questions/44467657/jest-better-way-to-disable-console-inside-unit-tests 

_**Note**: Sprint backlog is blocked and this seemed simple enough to update so I picked it up._

### DIFF
--- a/frontend/common/tests/setup.js
+++ b/frontend/common/tests/setup.js
@@ -1,0 +1,9 @@
+global.console = {
+  ...console,
+  // uncomment to ignore a specific log level
+  //log: jest.fn(),
+  //debug: jest.fn(),
+  //info: jest.fn(),
+  // warn: jest.fn(),
+  error: jest.fn(),
+};
--- a/frontend/common/jest.config.js
+++ b/frontend/common/jest.config.js
@@ -36,4 +36,7 @@ module.exports = {
 
   // Module file extensions for importing
   moduleFileExtensions: ["ts", "tsx", "js", "jsx", "json", "node"],
+
+  verbose: true,
+  setupFilesAfterEnv: ["<rootDir>/tests/setup.js"],
 };


petertgiles (approved):
Very cool!  ❄️

frontend/common/tests/setup.js
```diff
@@ -0,0 +1,9 @@
+global.console = {
```
patcon, line 1:
```suggestion
// source: https://stackoverflow.com/a/49591765
global.console = {
```

patcon:
Thanks!

- Pls do source the link when a solution comes from SO, as that helps tie external code snippets to the larger context :)
- Are there downsides to suppressing all errors in all tests in an ongoing way? (i'm seeing the SO solution isn't about overriding `console.error`)
- what about next solution down in SO? `jest.spyOn(console, 'error').mockImplementation(() => {});` in the single failing test itself? (principle of least surprise, if someone might _want_ to know about errors later while debugging)

esizer:
> * what about next solution down in SO? `jest.spyOn(console, 'error').mockImplementation(() => {});` in the single failing test itself? (principle of least surprise, if someone might _want_ to know about errors later while debugging)

I like that solution more since it prevents us missing `console.errors` while testing. Updated with [switch console.error mock implementation to specific test](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2908/commits/c76125ad1d0d9510d87931d08a7f7ff37809db65)

patcon (approved):

PR #2451, opened 2022-04-12
Refactor docker: the PR

patcon:
Resolves #2217 

## CAPTAINS LOG 🚀 

 - [x] did stuff together (last week) 👬 
 - TUE APR 12 cowork
   - [x] ensured htaccess being read
   - [x] removed LogMessage debug stuff
   - [x] moved htaccess to `frontend/conf/` dir
   - [ ] fix issue with `npm ci` from lockfile version change
   - [ ] ...

## Open Questions
1. in-container watch (node) vs manual vs on-host watch (docker)?
2. webserver routing structure / container arch?
    - frontend containers vs php container

PR #2945, opened 2022-06-01
Refactor Docker II: Tristan's Boogaloo

tristan-orourke:
I branched off feature/2217-refactor-docker (attached to PR #2451), so I'm currently using that as a base. 

Attempts to replace the refresh scripts with Dockerfiles for /api and /frontend. Building the docker image should result in a container that's properly set up, with all dependencies already installed.

Note that after any change (to code or to dependencies), the image must be rebuilt. Hopefully, the Docker cache should make this relatively painless. (Using WSL may be necessary for Windows users, to keep this quick.)

An unexpected complication was how to ensure codegen is using an up-to-date version of the graphql schema, when the api and frontend exist in different containers. The solution seems to be a script which builds the api image, copies the schema file out of it, and *then* builds the frontend image. I've written a Makefile which handles this, though the script could exist in other forms.

# TODO

- [x] Dockerfile which builds api project
- [x] Dockerfile which builds frontend projects as a single service
- [ ] Ensure frontend links to API correctly
- [ ] Ensure login still works correctly
    - [ ] mock auth
    - [ ] SiC auth
- [x] Script which ensures codegen is using the newest graphql schema
- [ ] A solution for live-reload in development
- [ ] Include tc-report in the frontend container
- [ ] Ensure github actions still work
- [ ] (Optional?) Optimize dockerfiles for rebuilding on windows, outside of WSL


patcon:
https://github.com/GCTC-NTGC/gc-digital-talent/pull/2451#issuecomment-1325459907

PR #2954, opened 2022-06-02
Added restart policy on mock-auth

patcon:
Pls drop an approval if it works to mitigate the issue you're experiencing :)

## Steps to confirm (my best guess)

**Before** checking out this branch

1. `docker-compose down`
2. `docker-compose up --detach` (presumably needs to be up)
3. shut down docker desktop and restart
4. `docker-compose ps` should show mock-auth container is down with exit code 143

Checkout this branch and repeat. If mock-auth container is up, it should me this is working

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -76,6 +76,10 @@ services:
     image: ghcr.io/navikt/mock-oauth2-server:0.4.8
     volumes:
       - ./infrastructure/conf:/app/conf
+    # Container shuts down with exit code 143 instead of the standard "0", so
+    # we ensure it comes back up after a Docker Desktop console restart.
+    # See: https://docs.docker.com/config/containers/start-containers-automatically/#use-a-restart-policy
+    restart: unless-stopped
     environment:
       # Things to note in config:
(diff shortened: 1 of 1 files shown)


RVany:
@patcon love the instruction you gave to test. The thing is auth container is running now even in the other branches too I tested  two other branches.

petertgiles (approved):
I checked out this branch and restarted my computer and it still didn't come up on its own.  I downed and then upped and then restarted my computer and it was up.  I'm not 100% sure this branch fixed it and maybe it was just me needing to down and up my network again?  I guess it can't hurt to have this set.

RVany:
> I checked out this branch and restarted my computer and it still didn't come up on its own. I downed and then upped and then restarted my computer and it was up. I'm not 100% sure this branch fixed it and maybe it was just me needing to down and up my network again? I guess it can't hurt to have this set.

I will approve too then,😀

RVany (approved):

patcon:
So weird...! Since I'm not seeing this myself, pls do drop an update about how this is working after you have a handle on whether anything got better. I don't want to forget about this pain point you're experiencing :)

PR #2955, opened 2022-06-02
a11y - 2915 - Add aria-required

### LINKED ISSUE #2915: Required fields are not announced by screen reader
## Description
Any required fields are not announced by a screen reader when navigating forms.

### Related

- https://github.com/GCTC-NTGC/gc-accessibility-passport/issues/111

esizer:
Resolves #2915 

## Summary

This adds the `aria-required="true"` attribute to inputs when `rules.required` is defined so they are properly announced to screen readers.


petertgiles (approved):
Looks great!  I'm not an expert but I turned on my screenreader and tabbing through the fields did announce that they were required.

PR #2958, opened 2022-06-02
Fix bug breaking going from search to request

### LINKED ISSUE #2930: Bug - first click from search to Request page results in blank page
# Description
Steps to reproduce
1. Go to https://talent.canada.ca/en/talent/search
2. Click Request (doesn't seem to matter what the filters are)
3. See blank page
4. Click refresh. Correct page loads (and still with the correct filters set!)

vd1992:
resolves #2930 

To test, make sure /search isn't breaking or off in anyway 
Added a basic Cypress test, first try. It is something to build up over time I'd say so it can be barebones for now

### DIFF
--- a/frontend/admin/src/js/components/Table/TableEditButton.tsx
+++ b/frontend/admin/src/js/components/Table/TableEditButton.tsx
@@ -1,7 +1,6 @@
 import React, { ReactElement } from "react";
-import { Button, Link } from "@common/components";
+import { Link } from "@common/components";
 import { useIntl } from "react-intl";
-import { navigate } from "@common/helpers/router";
 
 export interface TableEditButtonProps {
   /** Id of the object in the table. */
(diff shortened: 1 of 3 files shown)


esizer (approved):
Very nice! 👍 

Working well for me, tests pass and cannot reproduce the original issue anymore!

vd1992:
Thanks!

Gonna page @patcon 
New Cypress test!

petertgiles (changes requested):
This is very cool!  I wonder if we can do better on the test though?
- It has a bunch of built `wait`s that really slow it down.  The tests that Patrick wrote don't have that and can run much faster.
- The last assertion is just that the url has changed.  I think this isn't a great test since the bug does have the url changing, just nothing shows.  Can you make an assertion that the page has some expected content?  (The test does still fail on the bug since there are some exceptions thrown, though.)
- We're trying to keep related files together in a directory.  Is it possible to have this test file live near the search page component?

petertgiles (approved):
Nice, this test looks much better.  I'll start a slack discussion about test file placement.

PR #2960, opened 2022-06-02
Feature  -2923 - Register Page

### LINKED ISSUE #2923: Register Page
There will be a page to give a user context before registering for GCKey.

[image](https://www.figma.com/file/UDWAkiGXMt9YGbrfsuSzmr/Interface---Registration-and-login?node-id=133%3A1144)

## User Acceptance Critera
- [ ] Can navigate to /[lang]/register to view page
- [ ] Post login redirect should be the welcome page
- [ ] Files created in talentsearch directory
- [ ] The three links on the page operate

esizer:
Resolves #2923 

## Summary

This adds a page where users can register to use the application.

## Issue

While it is fine to have multiple links with different text that go to the same location, the issue is that the links on this page each have a different function.

- Register for a GC Key
- Login to the application
- Register for application with GC Key

Ref: https://www.w3.org/TR/WCAG20-TECHS/H30.html

### DIFF
--- a/frontend/common/src/components/Link/Link.tsx
+++ b/frontend/common/src/components/Link/Link.tsx
@@ -10,6 +10,7 @@ export interface LinkProps extends React.HTMLProps<HTMLAnchorElement> {
   mode?: "solid" | "outline" | "inline";
   block?: boolean;
   type?: "button" | "link";
+  /** For use when linking to a domain outside of the application */
   external?: boolean;
 }
 
--- a/frontend/common/src/components/Link/Link.tsx
+++ b/frontend/common/src/components/Link/Link.tsx
@@ -10,6 +10,7 @@ export interface LinkProps extends React.HTMLProps<HTMLAnchorElement> {
   mode?: "solid" | "outline" | "inline";
   block?: boolean;
   type?: "button" | "link";
+  external?: boolean;
 }
 
 const Link: React.FC<LinkProps> = ({
@@ -18,6 +19,7 @@ const Link: React.FC<LinkProps> = ({
   color,
   mode = "solid",
   block = false,
+  external = false,
   type = "link",
   children,
   className,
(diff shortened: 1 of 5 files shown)


petertgiles:
Yeah, I'm seeing that weird behavior I was worried about.  The *Register for one* link correctly hits the `/login` endpoint of the API while the two buttons at the bottom try to navigate within React to `/en/login` which doesn't exist.  I think all three should go to the API endpoint, right?

esizer:
> Yeah, I'm seeing that weird behavior I was worried about. The _Register for one_ link correctly hits the `/login` endpoint of the API while the two buttons at the bottom try to navigate within React to `/en/login` which doesn't exist. I think all three should go to the API endpoint, right?

I forgot that `navigate` only works on internal links. Fixed with [allow override of navigate on Link](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2960/commits/5c5540e976ea3d147c2b4247953a22f19c124fea)

vd1992:
- small thing, but in interface `LinkProps`, could you add a comment referring to the purpose of `external`, for ease of understanding when someone is looking through Link at a later time
- wireframe appears to want bolding around the "Dont have a gc account?" line

esizer:
> * small thing, but in interface `LinkProps`, could you add a comment referring to the purpose of `external`, for ease of understanding when someone is looking through Link at a later time

Added comment with [add external link comment](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2960/commits/0130e587c98282471b35d83ce6e84b1db71732a3)

> * wireframe appears to want bolding around the "Dont have a gc account?" line

Thanks for catching that, I have bolded it with [bold dont have a key?](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2960/commits/f50cd8a39c2473bc4c26455678d29975ce7b11d0)

petertgiles (approved):
Love it!

PR #2966, opened 2022-06-03
change equity to OR not AND filtering

### LINKED ISSUE #2931: EE groups should work as OR filter, not AND filter, in talent search
# Description
Currently, if you check more EE boxes, the search becomes increasingly restrictive. ie if you select "Indigenous" and "Woman", the count will only include candidates who are both women and indigenous. (Incidentally, that's also how the skill filters work. If you select multiple skills, you only get candidates who have all the skills.)

We need it to work differently, so that checking more EE boxes increases the number of candidates you can match. ie if you select "Indigenous" and "Woman", the count will include all women candidates and all indigenous candidates.

# Acceptance Criteria

- [ ] phpunit tests
- [ ] Must affect both PoolCandidate and User search

vd1992:
resolves #2931 
learning, so this'll be rough

- at /search, equity filters are no longer AND functionally 
- tests adjusted and new tests added
- manually exercising and unit tests are passing

to test, try playing with filters and comparing to expected results via Adminer
I think this is complete?


api/app/Providers/AppServiceProvider.php
```diff
@@ -26,5 +28,12 @@ public function boot()
             'personalExperience' => \App\Models\PersonalExperience::class,
             'workExperience' => \App\Models\WorkExperience::class,
         ]);
+        DB::listen(function($query) {
```
petertgiles, line 31: Oops, don't leave debugging code live.  I use this a lot though... maybe leave it in but commented out?  With a comment to enable it for debugging?

api/graphql/schema.graphql
```diff
@@ -336,15 +336,26 @@ enum PoolCandidateSearchStatus {
     DONE
 }
 
+type EquityObject {
```
petertgiles, line 339: I'm not sure `object` is the best suffix.  Naming is hard...  @Jerryescandon what are these?  EquityFilters? But filters sound like each will remove results and we're changing them to add results.  EquityAdditions?  EquityPreferences?  Maybe just plain-ole EmploymentEquity.

api/app/Models/PoolCandidate.php
```diff
@@ -234,41 +236,59 @@ public function filterByPools(Builder $query, array $pools): Builder
...
-            $query->where('is_visible_minority', true);
+        if ($equity["is_indigenous"]) {
+            array_push($equityVars, "is_indigenous");
+        }
+        if ($equity["is_visible_minority"]) {
+            array_push($equityVars, "is_visible_minority");
+        }
+
+        // then return queries depending on above array count, special query syntax needed for case of >=2 to ensure proper SQL query formed
+        if (count($equityVars) == 0) {
```
petertgiles, line 262: Are these branches for count == 0 and count == 1 necessary?  I think it would work without them?

frontend/talentsearch/src/js/components/request/CreateRequest.tsx
```diff
@@ -45,11 +46,12 @@ type FormValues = {
     cmoAssets?: {
       sync?: Array<Maybe<CmoAsset["id"]>>;
     };
+    // equity?: EquityObject["hasDisability"];
```
petertgiles, line 49: leftover debugging comment?

api/tests/Feature/PoolCandidateTest.php
```diff
@@ -411,7 +411,12 @@ public function testFilterByEmploymentEquity(): void
       }
     ', [
       'where' => [
-        'isIndigenous' => true,
+        'equity' => [
```
petertgiles, line 414: Try adding a test or two where the fields are missing instead of just false.  The schema has them as nullable, after all.  You'll find that you need to add some null handling. :grin:
image

petertgiles (changes requested):
Nice, overall it works great!  I've added a few things to check but you're most of the way there already.  :+1:

frontend/talentsearch/src/js/components/request/CreateRequest.tsx
vd1992: 😊

api/app/Models/PoolCandidate.php
```diff
@@ -21,6 +21,8 @@
  * @property boolean $is_indigenous
  * @property boolean $is_visible_minority
  * @property boolean $has_diploma
+ * @property object $equityObject
```
JamesHuf, line 24: Neither of these are actually properties of the poolCandidate model and should probably be removed.

JamesHuf (approved):
This is looking good so far. 👍 Peter's already covered pretty much everything I saw, I just have one minor note below.

api/app/Models/PoolCandidate.php
vd1992: I'll leave in the handle 0 for a just in case, but yeah, == 1 doesnt seem needed

PR #2967, opened 2022-06-03
Print candidate profile (2nd attempt PR)

### LINKED ISSUE #2663: Download Profile functionality
button styling done in #2658 

https://www.figma.com/proto/XS4Ag6GWcgdq2dBlLzBkay/Admin-v3?node-id=429%3A2292&scaling=min-zoom&page-id=0%3A1&starting-point-node-id=410%3A1984&hide-ui=1

Screen Shot 2022-05-04 at 11 48 22 AM

petertgiles:
This branch adds the ability to print the candidate profile.  Using the browser built-in PDF printer will allow a PDF export.  It uses the same "react-to-print" library as the spike.

Suggested Tests:
- [ ] The accordion styling remains the same on screen
- [ ] Pressing the Download button on the user page triggers the browser print dialog
- [ ] The printed document resembles the Candidate Profile tab contents
- [ ] The printed accordions should be all open, on the "by type" tab

Closes #2663


frontend/admin/src/js/components/user/UserProfileDocument.tsx
```diff
@@ -0,0 +1,222 @@
...
+          flex-shrink: 0;
+          flex: initial;
+          max-width: 100%;
+          align-items: center;
+          flex-direction: row;
+        }
+      }
+    `}</style>
+      <div ref={ref}>
+        <div className="print-container">
```
petertgiles, line 106: All this is copied from the UserProfile component to ensure it resembles it. It can be customized later if designs are provided.

frontend/admin/src/js/components/user/UserProfilePrintButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+export interface UserProfilePrintButtonProps {
+  userId: Scalars["ID"];
+}
+
+export const UserProfilePrintButton: React.FunctionComponent<{
+  userId: Scalars["ID"];
+}> = ({ userId, children }) => {
+  const intl = useIntl();
+
+  // would be nice to only fire this if the button was clicked but you would need an imperative version of the query that you could await on
```
petertgiles, line 18: If anyone knows how to get synchronous queries or a promise then I could hold back the query. But since the query is the same one as in the tab, URQL caches the results and only queries once anyway.

frontend/admin/src/js/components/user/UserProfilePrintButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const [{ data: initialData, fetching, error }] = useGetUserProfileQuery({
+    variables: { id: userId },
+  });
+
+  const componentRef = useRef(null);
+  const handlePrint = useReactToPrint({
+    content: () => componentRef.current,
+    pageStyle: `
```
petertgiles, line 26: I'd like to centralize some standard print styling in a file but I'm not sure how to get CSS into a string. I tried CSS modules but that provides an object. I starting looking into some webpack plugins to inline a file but decided it was more trouble than was worth.

frontend/common/src/components/UserProfile/ExperienceByTypeListing.tsx
```diff
@@ -0,0 +1,132 @@
+import {
```
petertgiles, line 1: Factored out this tab to use it in the print component.

frontend/admin/src/js/components/user/UserProfileDocument.tsx
```diff
@@ -0,0 +1,222 @@
...
+  return (
+    <div
+      style={{
+        display: "none",
+      }}
+    >
+      {/* a bunch of styling to tweak components for printing */}
+      <style type="text/css" media="print">{`
+
+        html{font-size:75%}
```
esizer, line 77: We should add some padding here. In chrome on MacOS, the text is bumping right up against the edge of the page.

```suggestion
        html{font-size:75%; padding: 16px;}
```

frontend/admin/src/js/components/user/UserProfilePrintButton.tsx
esizer: You _could_ just export it as a string from a `.ts` file. 🤔 

`styles.ts`

```
export default `
@page {
  size: letter portrait;
}

@media print {
  .page-section {
    margin-bottom: 2rem;
    display: block;
    page-break-after: auto;
    page-break-inside: avoid;
    -webkit-region-break-inside: avoid;
  }
`
```

esizer (changes requested):
Sweet! 🍭 

Outside of some minor styling, this is looking great!

frontend/admin/src/js/components/user/UserProfileDocument.tsx
petertgiles: Nice, thanks!

frontend/admin/src/js/components/user/ViewUser.tsx
```diff
@@ -72,21 +72,15 @@ export const ViewUserPage: React.FC<ViewUserPageProps> = ({ user }) => {
...
-            color="primary"
-            type="button"
-            href="/"
-            // download={ TODO }
-          >
+          <UserProfilePrintButton userId={user.id}>
             <span>
               <DownloadIcon style={{ width: "1rem" }} />{" "}
               {intl.formatMessage({
                 defaultMessage: "Download Profile",
```
petertgiles, line 79: I wonder if this should be change to *Print Profile* with a printer icon?  It would make it a little clearer that a print dialog is about to appear, not a save file dialog.

frontend/admin/src/js/components/user/UserProfilePrintButton.tsx
petertgiles: Sometimes the simplest solutions are best.  Thanks Eric!  Added them in https://github.com/GCTC-NTGC/gc-digital-talent/pull/2967/commits/a9901301b0ca671dff173fc2096ccc43aca868d8.

frontend/admin/src/js/components/user/ViewUser.tsx
petertgiles: @Jerryescandon switching it to *Print Profile*?

Jerryescandon: Switch it baby

frontend/admin/src/js/components/user/UserProfilePrintButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const userData = initialData;
+
+  if (fetching) return <p>{intl.formatMessage(commonMessages.loadingTitle)}</p>;
+  if (error)
+    return (
+      <p>
+        {intl.formatMessage(commonMessages.loadingError)}
+        {error.message}
+      </p>
+    );
```
esizer, line 42: With the new `<Pending />` component, this can safely be removed.

esizer (approved):
I left a small comment, but I actually prefer just a small loading message instead of the full screen loader.

Maybe we can add a prop to the pending component to allow it to be rendered inline rather than full screen. Also, I'm not seeing the background colour for that component anymore 😕 

Anyway, everything looks good and it is working wonderfully 😄

frontend/admin/src/js/components/user/UserProfilePrintButton.tsx
esizer: > Maybe we can add a prop to the pending component to allow it to be rendered inline rather than full screen. Also, I'm not seeing the background colour for that component anymore

This is just me making notes to myself. Feel free to ignore if you don't care 😅

Issue #2924, opened 2022-05-31
Welcome page (url: /create-account)


tristan-orourke:
When a user is auto-created they will arrive at this page to be given the opportunity to fill in their user record before using the site.

[![image](https://user-images.githubusercontent.com/8978655/171274055-fd208932-b0c4-42be-914f-b890322b1935.png)](https://www.figma.com/file/UDWAkiGXMt9YGbrfsuSzmr/Interface---Registration-and-login?node-id=0%3A1 
)

Acceptance Criteria
- [ ] Page can be navigated to at /create-account
- [ ] Files created in the talentsearch directory
- [ ] As much initial data is filled into the form as possible (maybe just preferred language?)
- [ ] A mutation saves the updates to the api and a successful save navigates to the user profile flow
- [ ] Form validation prevents incomplete form submission
- [ ] Government form elements only appear (and become required) when the user selects "Yes, I work with the government"

petertgiles:
@JoshBeveridge If a user navigates here with a completed profile should we prefill the fields and allow them to use the page as an editor?  If not, how do we detect that?  Would we assign them the APPLICANT role when they fill this form and check for that?  Would we redirect them to the profiles flow then?

JoshBeveridge:
@petertgiles Theoretically they shouldn't ever reach this page once it's completed. If they somehow did, I would imagine it would simply redirect to their profile, yes.

tristan-orourke:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/GC-Digital-Talent-6092dc05157fa30010edaef9/issues/GCTC-NTGC/gc-digital-talent/2924?planning-poker) with ZenHub @yonikid15

PR #2992, opened 2022-06-07
WIP: Feature - 2925 - Logged out page

### LINKED ISSUE #2925: Logout landing page
Use url `/logged-out`.
Build this in /talentsearch

[Prototype](https://www.figma.com/file/UDWAkiGXMt9YGbrfsuSzmr/Interface---Registration-and-login?node-id=84%3A1065)
[Screenshot](
Screenshot

# Acceptance Criteria

- [ ] If you land on this page, while logged in, throw up a confirmation dialog which asks if they want to log out.

esizer:
Resolves #2925 

## WIP

This blocked by a few other issues so just waiting on those to finish it off.

- #2924 Waiting on the implementation of the alert here
- #2923 This has some routing needed to display the page
- #2896 Auth container needed to display log out confirmation dialog

## Summary

This created a page at `/logged-out` to display to a user **_after_** they have logged out. If they land on this page while logged in, a dialog will appear asking if they would like to logout.

## New Component - `<TileLink />`

This is a simplified version of the `<Link />`  component that accepts the following:

- **color**: Color - _required_
  - This controls the colour of the border to the left
- **href**: string - _required_
  - URL to point the link to
-  **external**: boolean
  - If the link goes to external location, avoid using `universal-router`

### DIFF
--- a/frontend/common/src/components/TileLink/TileLink.stories.tsx
+++ b/frontend/common/src/components/TileLink/TileLink.stories.tsx
@@ -1,0 +1,45 @@
+import React from "react";
+import { Story, Meta } from "@storybook/react";
+import TileLink from "./TileLink";
+import type { TileLinkProps } from "./TileLink";
+
+export default {
+  component: TileLink,
+  title: "Components/Tile Link",
+  args: {
+    label: "Tile Link Label",
+    color: "primary",
+  },
+  argTypes: {
+    label: {
+      name: "Label",
+      type: { name: "string", required: true },
+    },
+    color: {
+      name: "Colour",
+      options: [
+        "primary",
+        "secondary",
+        "cta",
+        "white",
+        "black",
+        "ia-primary",
+        "ia-secondary",
+      ],
+    },
+  },
+};
+
+const TemplateTileLink: Story = (args) => {
+  const { label, color } = args;
+
+  return (
+    <div style={{ width: 250 }}>
+      <TileLink color={color} href="#">
+        {label}
+      </TileLink>
+    </div>
+  );
+};
+
+export const BasicTileLink = TemplateTileLink.bind({});
--- a/frontend/common/src/components/TileLink/TileLink.tsx
+++ b/frontend/common/src/components/TileLink/TileLink.tsx
@@ -1,0 +1,83 @@
+import { ChevronDoubleRightIcon } from "@heroicons/react/solid";
+import React from "react";
+import { navigate } from "../../helpers/router";
+import type { Color } from "../Button";
+
+const colorMap: Record<Color, Record<string, string>> = {
+  primary: {
+    "data-h2-border": "b(lightpurple, left, solid, m)",
+  },
+  secondary: {
+    "data-h2-border": "b(lightnavy, left, solid, m)",
+  },
+  cta: {
+    "data-h2-border": "b(darkgold, left, solid, m)",
+  },
+  white: {
+    "data-h2-border": "b(white, left, solid, m)",
+  },
+  black: {
+    "data-h2-border": "b(black, left, solid, m)",
+  },
+  "ia-primary": {
+    "data-h2-border": "b(ia-pink, left, solid, m)",
+  },
+  "ia-secondary": {
+    "data-h2-border": "b(ia-purple, left, solid, m)",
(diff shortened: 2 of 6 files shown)


petertgiles (changes requested):
So strictly speaking this is out of scope for this PR but it did happen in your PR from less than 24 hours ago so...
Could you please bump up the font weights in the header?  In Firefox-Linux (and I'm guessing Safari?) it becomes almost invisible.
image

petertgiles:
Much better, thanks!
image

esizer:
> Much better, thanks!

Still a bit lighter than the other two links and they are using the same component 😖

petertgiles (commented):
Some thoughts:
- You would need to add OAUTH_LOGOUT_URI to the talentsearch SJS page to make the post-logout redirect work properly, right?
- If I manually navigate to the /logged-out page to pop the dialog and click cancel I'm just ending up on /talentsearch.  Can we actually go back to the previous page?
- I think the difference in styling between Request and Login is that one is a link and one is a button.
- When you click logout there's a super simple browser dialog but on the logged-out page there's a super fancy react dialog.  Is there any advantage to picking one or the other for consistency?

esizer:
> > * I think the difference in styling between Request and Login is that one is a link and one is a button.
> 
> Woah, why are our links using open sans vs buttons using fira? I'm going to be honest, most fonts look the same to me and i never noticed this. I'll look into this. Is this intentional or should open sans be used everywhere?

Mentioned in slack, but this is due to an incomplete css reset. I have added a workaround for now in [add workaround font reset](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2992/commits/2a7e25f3d69a082a38a60a07ce643a72ac3c0431)

petertgiles:
> I'm going to be honest, that sounds correct but I did not and it worked for me during testing shrug . Did you need to do that to get it to work?

Yeah, I did.  I'm pretty confident that this is necessary

> So, I may be misunderstanding this. If you navigate directly to it that means typing it into the address bar, right? If so, the previous page only exists in browser history which we do not have access to.  The only thing I can think of is: window.history.goBack(). If they meant to go to talent.canda.ca from a different website, they would actually be directed back to that previous website and not to our application.

That's fair.  If a simple "back" is not possible then let's maybe send them back to the home page for the entire app "/"?

> Woah, why are our links using open sans vs buttons using fira? I'm going to be honest, most fonts look the same to me and i never noticed this. I'll look into this. Is this intentional or should open sans be used everywhere?

Sounds like you've got this one figured out.  I'm definitely not the expert on this one. :grin: 

> I like having consistency and prefer the react dialog. If you agree, I don't think it will be much to update them.

Sounds good to me!

petertgiles:
> I'm going to be honest, that sounds correct but I did not and it worked for me during testing shrug .

I have a theory about that... I actually had the opposite problem during testing.  I had the "disable cache" box selected but the update SJS file wouldn't get pulled in until I actually loaded that file directly in my browser!  I wonder if you had it cached with the logout URL populated even though Apache wouldn't have it.

esizer:
> Yeah, I did. I'm pretty confident that this is necessary

Fixed with [add logout uri to sjs](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2992/commits/42dfbf8fd722fb729b7f58e208d59dd5790a703b)

> That's fair. If a simple "back" is not possible then let's maybe send them back to the home page for the entire app "/"?

Updated the cancel url in [change cancel link destination](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2992/commits/c63ea0ff0128027bac5e11439c4c17e230a20f8b)

> Sounds like you've got this one figured out. I'm definitely not the expert on this one. 😁

Add workaround in [add workaround font reset](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2992/commits/2a7e25f3d69a082a38a60a07ce643a72ac3c0431)

> Sounds good to me!

Updated admin and talentsearch logout dialogs in [update logout confirmation dialogs](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2992/commits/5b99a19d1f1c03a42285f29dd18014ad18b8832e)

petertgiles (approved):

PR #3006, opened 2022-06-07
improve seeded data for local testing

### LINKED ISSUE #2894: Improve dev seeded data
Some of the dev seeded data is not very realistic and makes it tricky to do manual testing and development.

- Our candidates should have expected salary or expected classifications but not both
- Our candidates should actually conform to the requirements of the pool they are in
- Some of our candidates should be expired

vd1992:
resolves #2894 

- candidates can be expired now
- restricted to IT 
- salary or classifications now, not both

testing, make sure it seeds successfully locally and play around in the database with your preferred method (app/adminer/etc)

### DIFF
--- a/api/database/factories/PoolFactory.php
+++ b/api/database/factories/PoolFactory.php
@@ -43,9 +43,10 @@ class PoolFactory extends Factory
 
     public function configure()
     {
+        // take(5) will grab the first 5 classifications, in this case, all the IT classifications, restricting search to IT classifications
         return $this->afterCreating(function (Pool $pool) {
             $assets = CmoAsset::inRandomOrder()->limit(4)->get();
-            $classifications = Classification::inRandomOrder()->limit(3)->get();
+            $classifications = Classification::take(5)->get();
             $pool->essentialCriteria()->saveMany($assets->slice(0,2));
             $pool->assetCriteria()->saveMany($assets->slice(2,2));
             $pool->classifications()->saveMany($classifications);
--- a/api/database/factories/PoolFactory.php
+++ b/api/database/factories/PoolFactory.php
@@ -45,7 +45,7 @@ class PoolFactory extends Factory
     {
         return $this->afterCreating(function (Pool $pool) {
             $assets = CmoAsset::inRandomOrder()->limit(4)->get();
-            $classifications = Classification::where('group', 'IT')->inRandomOrder()->limit(3)->get();
+            $classifications = Classification::inRandomOrder()->limit(5)->get();
             $pool->essentialCriteria()->saveMany($assets->slice(0,2));
             $pool->assetCriteria()->saveMany($assets->slice(2,2));
             $pool->classifications()->saveMany($classifications);
--- a/api/database/factories/PoolFactory.php
+++ b/api/database/factories/PoolFactory.php
@@ -43,10 +43,9 @@ class PoolFactory extends Factory
 
     public function configure()
     {
-        // take(5) will grab the first 5 classifications, in this case, all the IT classifications, restricting search to IT classifications
         return $this->afterCreating(function (Pool $pool) {
             $assets = CmoAsset::inRandomOrder()->limit(4)->get();
-            $classifications = Classification::take(5)->get();
+            $classifications = Classification::where('group', 'IT')->inRandomOrder()->limit(3)->get();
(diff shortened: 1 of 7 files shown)


api/database/factories/PoolFactory.php
```diff
@@ -43,9 +43,10 @@ public function definition()
 
     public function configure()
     {
+        // take(5) will grab the first 5 classifications, in this case, all the IT classifications, restricting search to IT classifications
         return $this->afterCreating(function (Pool $pool) {
             $assets = CmoAsset::inRandomOrder()->limit(4)->get();
-            $classifications = Classification::inRandomOrder()->limit(3)->get();
+            $classifications = Classification::take(5)->get();
```
petertgiles, line 49: This is a bit dangerous.  You're depending on the classifications to always be seeded a certain way and the database behaving in a certain way.  The assumptions could very easily break.

api/database/seeders/DatabaseSeeder.php
```diff
@@ -52,22 +54,86 @@ public function run()
...
-             'roles' => [ApiEnums::ROLE_APPLICANT]
-        ])
-            ->count(60)
+            'roles' => [ApiEnums::ROLE_APPLICANT]
+       ])
+            ->count(20)
             ->afterCreating(function (User $user) {
                 $assets = CmoAsset::inRandomOrder()->limit(4)->pluck('id')->toArray();
-                $classifications = Classification::inRandomOrder()->limit(3)->pluck('id')->toArray();
+                $classifications = Classification::take(5)->inRandomOrder()->limit(3)->get()->pluck('id')->toArray();
```
petertgiles, line 66: Again, based on assumptions that can break.

api/database/seeders/DatabaseSeeder.php
```diff
@@ -52,22 +54,86 @@ public function run()
...
                 $user->cmoAssets()->sync($assets);
                 $user->expectedClassifications()->sync($classifications);
+                PoolCandidate::factory()->count(1)->sequence(fn () => [
+                    'pool_id' => Pool::pluck('id', 'key')->toArray()['digital_careers'],
```
petertgiles, line 70: Not a big deal but this is a bit sloppy.  Rather than querying the db for every key and selecting on you can query for just the one you need.

api/tests/Feature/PoolCandidateTest.php
```diff
@@ -30,14 +30,15 @@ public function testFilterByClassification(): void
     Classification::factory()->count(3)->create();
     PoolCandidate::factory()->count(5)->create([
       'expected_salary' => [], // remove salaries to avoid accidental classification-to-salary matching
+      'expiry_date' => '2050-01-01', // ensure no candidates are expired for this test
```
petertgiles, line 33: This is a "magic number" which is an anti-pattern.  Try making a constant for it with a descriptive name like FAR_FUTURE_DATE and use that through the file instead.

petertgiles (commented):
Looking good so far!  You're on the right track for most of it.  

The requirement of placing candidates in pools that they conform to is the trickiest AC by far.  I don't love the approach you're using here though.  I suggest writing a little function in the seeder that accepts an empty user, randomly selects a pool, places in the candidate in it, and then seeds the fields of the user based on the pool properties.

vd1992:
> Looking good so far! You're on the right track for most of it.
> 
> The requirement of placing candidates in pools that they conform to is the trickiest AC by far. I don't love the approach you're using here though. I suggest writing a little function in the seeder that accepts an empty user, randomly selects a pool, places in the candidate in it, and then seeds the fields of the user based on the pool properties.

I'm not sure, rolling seeding all pools into one go feels like it can get confusing, especially when you want to seed multiple pools in different ways 😕

api/database/helpers/ApiEnums.php
```diff
@@ -79,4 +79,21 @@ public static function roles() : array
             self::ROLE_APPLICANT,
         ];
     }
+
+    /**
+     * A collection of enums for operation_requirement in factories and seeders
```
petertgiles, line 84:
```suggestion
     * A collection of enums for salary ranges in factories and seeders
```

yonikid15 (approved):
Love it 👍 looks like it's working on my end.

PR #3013, opened 2022-06-09
Feature - 2951 - Redirect new users

### LINKED ISSUE #2951: Forward new users to to /create-account page
For all logged in users, if they have not initialized their account, forward them to the /create-account page.

esizer:
Resolves #2951 

## Summary

This checks to see if a user is logged in and has an email associated with their account. If the user has no email, it will redirect to `/talent/profile/create-account`.

### Hardcoded Path

~~I'm not happy about it but I had to hardcode the path since the functionality exists in `common` but the path is exclusive to `talentsearch`. Any suggestions to avoid hardcoding this?~~

Answered by @tristan-orourke :D We are now passing this through the `useRouter` hook.

### DIFF
--- a/frontend/common/src/components/Auth/AuthorizationContainer.tsx
+++ b/frontend/common/src/components/Auth/AuthorizationContainer.tsx
@@ -1,8 +1,8 @@
 import React from "react";
-import { Role } from "../../api/generated";
+import { Maybe, Role } from "../../api/generated";
 
 export type PossibleUserRoles = (Role | null | undefined)[] | null | undefined;
-export type PossibleEmail = string | null;
+export type MaybeEmail = Maybe<string>;
 
 interface AuthorizationState {
   loggedInUserRoles: PossibleUserRoles;
--- a/frontend/common/src/components/Auth/AuthorizationContainer.tsx
+++ b/frontend/common/src/components/Auth/AuthorizationContainer.tsx
@@ -2,28 +2,34 @@ import React from "react";
 import { Role } from "../../api/generated";
 
 export type PossibleUserRoles = (Role | null | undefined)[] | null | undefined;
+export type PossibleEmail = string | null;
 
 interface AuthorizationState {
   loggedInUserRoles: PossibleUserRoles;
+  loggedInEmail?: PossibleEmail;
 }
 
 export const AuthorizationContext = React.createContext<AuthorizationState>({
   loggedInUserRoles: null,
+  loggedInEmail: null,
 });
 
 interface AuthorizationContainerProps {
   userRoles?: PossibleUserRoles;
+  email?: PossibleEmail;
 }
 
 export const AuthorizationContainer: React.FC<AuthorizationContainerProps> = ({
   userRoles,
+  email,
   children,
 }) => {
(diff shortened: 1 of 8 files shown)


frontend/common/src/helpers/router.tsx
```diff
@@ -102,6 +102,7 @@ export const useRouter = (
   routes: Routes<RouterResult>,
   missingRouteComponent: ReactElement,
   notAuthorizedComponent: ReactElement,
+  welcomeRoute?: string,
```
esizer, line 105: @tristan-orourke I think this was what you were referring to, passing the route to the router hook?

frontend/common/src/components/Auth/AuthorizationContainer.tsx
```diff
@@ -2,28 +2,34 @@ import React from "react";
 import { Role } from "../../api/generated";
 
 export type PossibleUserRoles = (Role | null | undefined)[] | null | undefined;
+export type PossibleEmail = string | null;
```
patcon, line 5:
```suggestion
import { Maybe, Role } from "../../api/generated";

export type PossibleUserRoles = (Role | null | undefined)[] | null | undefined;
export type MaybeEmail = Maybe<string>;
```

Thoughts on using `Maybe<T>`? It's [pretty well-established](https://en.wikipedia.org/wiki/Monad_(functional_programming)#An_example:_Maybe) beyond our project (though we don't go full-monad :) )

This keeps the maybe/possible types consistent across the project

esizer: Makes sense to me, I was just following the pattern established in the file with `PossibleRoles`. Committed suggestion 👍

frontend/common/src/helpers/router.tsx
tristan-orourke: Yep, that's pretty much what I was thinking.

frontend/common/src/helpers/router.tsx
```diff
@@ -119,6 +121,16 @@ export const useRouter = (
         // may or may not be a promise, so attempt to resolve it. A non-promise value will simply resolve to itself.
         const route = await Promise.resolve(routeMaybePromise);
 
+        // Redirect authenticated user if no email exists yet
+        if (
+          welcomeRoute &&
+          loggedIn &&
+          (typeof loggedInEmail === undefined || loggedInEmail === null)
```
tristan-orourke, line 128: Is there any reason you're using `typeof` here instead of just checking if `loggedInEmail === undefined`?
Also btw we have a helper function for this in common/src/helpers/util.ts

esizer: Oh cool, thanks for the tip! This is actually a bug and I didn't mean to write it this way but `typeof` is fully intended because of weirdness with javascript. 

```tsx
loggedInEmail === undefined // throws an error "loggedInEmail is not defined" since it is being used as a value here

typeof loggedInEmail === 'undefined' // used as type so will not throw an error
```

patcon (approved):
Looks great! Agree with @petertgiles: yay for isLoading!

PR #3015, opened 2022-06-09
Increased cypress timeout

patcon:
Re-ticketed [from Slack message](https://talent-cloud.slack.com/archives/C0259G6KXJ6/p1654691932999009?thread_ts=1654635948.795159&cid=C0259G6KXJ6) :lock:

We have some actions that take longer than the 4-second timeout of cypress commands.

The ways to handle this, from best to worst:
1. wait for a related http request to finish via `cy.wait('@someAlias')`
2. look for progressive changes on the way, helpful for progressively loading UIs like ours (each `cy.get()` or equivalent has its own timeout)
3. Bump timeout either on the command or globally in `cypress.json`:
    ```js
    cy.get('.my-selector', {timeout: 10000}).should('not.exist')
    ```

I bumped the timeout globally (3), but also added an example of the intended way (1).

### DIFF
--- a/frontend/cypress.json
+++ b/frontend/cypress.json
@@ -1,4 +1,8 @@
 {
+  "defaultCommandTimeout": 4000,
+  "pageLoadTimeout": 60000,
+  "requestTimeout": 5000,
+  "responseTimeout": 30000,
   "reporter": "cypress-multi-reporters",
   "reporterOptions": {
     "configFile": "cypress/reporters-config.json"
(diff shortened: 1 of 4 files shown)


vd1992:
sorry, I am not too sure on this
not sure what a good course of action would be, I guess (3) is quick and simple it not being elegant isn't a big deal

petertgiles (approved):
This looks good to me.  I'm not able to replicate the timeouts on my system but I can confirm the tests still ran quickly after the changes.  I appreciate the two-pronged approach of demonstrating the correct way to do it and also bumping the timeouts for plan B.  :star2:

PR #3019, opened 2022-06-09
add filter for status = available

### LINKED ISSUE #3007: Talent Search should only return candidates with AVAILABLE status
Talent searches should only return candidates with the AVAILABLE status.

# Acceptance Criteria
- [x] PoolCandidates with any status except AVAILABLE do not appear in talent search results
- [x] When searching Users/Applicants instead of PoolCandidates but filtering by Pool, only users with PoolCandidates with AVAILABLE status in those Pools will appear in results.

vd1992:
resolves #3007 

- schema changed
- /search now searches for 'available'
- tests added

when more admin stuff is hit with the 2.0 or Talentsearch updated and select by status becomes implemented, the filter can be upgraded 
to test, make sure database and /search are in agreement for candidate count


api/app/Models/User.php
```diff
@@ -447,6 +447,20 @@ public function filterByEquity(Builder $query, array $equity): Builder
...
     }
 
+    public function filterByPoolCandidateStatus(Builder $query, ?string $poolCandidateStatus): Builder
+    {
+        if(empty($poolCandidateStatus)){
+            return $query;
+        }
+
+        if($poolCandidateStatus == 'AVAILABLE'){
+            $query->where('pool_candidate_status', 'AVAILABLE');
```
tristan-orourke, line 457: I don't think the user has a pool_candidate_status. This filter will actually be a bit more complicated on the User model, since it only applies to the Pools we're filtering by.

api/app/Models/PoolCandidate.php
```diff
@@ -270,6 +270,20 @@ public function filterByEquity(Builder $query, array $equity): Builder
         return $query;
     }
 
+    public function filterByPoolCandidateStatus(Builder $query, ?string $poolCandidateStatus): Builder
```
tristan-orourke, line 273: Oh! This is the perfect place for a [Laravel query scope](https://laravel.com/docs/9.x/eloquent#local-scopes)!

api/app/Models/PoolCandidate.php
```diff
@@ -270,6 +270,20 @@ public function filterByEquity(Builder $query, array $equity): Builder
...
+        }
+
+        if($poolCandidateStatus == 'AVAILABLE'){
+            $query->where('pool_candidate_status', 'AVAILABLE');
+            return $query;
+        }
+
+        return $query;
+    }
+
```
patcon, line 286: Take this with a grain of salt, as maybe I'm missing some consideration, but: Since this only handles "available" status, why not just make a `scopeAvailable` with no args?

```suggestion
    public function scopeAvailable(Builder $query): Builder
    {
        $query->where('pool_candidate_status', 'AVAILABLE');
    }

```

tristan-orourke (changes requested):
Okay. So I think the best approach would be to handle this similarly to how we handled expiry dates. That is, we added a field to UserFilterAndOrderInput, which is used to allow Admins to search over Users, but we didn't add anything to PoolCandidateFilterInput. Instead, we just added a scope to the countPoolCandidates and searchPoolCandidates queries.

eg `countPoolCandidates(where: PoolCandidateFilterInput): Int! @count(model: "PoolCandidate" scopes: ["expiryFilter"])`

This made sense because Admins might want options on how they filter for candidates, but when a manager is using the search page, they always want to see a count of unexpired candidates.

The same will be useful here. Admins may want to filter on all the statuses for users. Ideally, we could allow them to use a multiselect field, and pass in the array of statuses they want to allow. But the talent search should only ever count Available candidates, so the field doesn't need to be part of PoolCandidateFilter.

tristan-orourke:
Sorry @vd1992! I was in the middle of editing the review when I accidentally submitted it! There are some inconsistencies right now.

api/app/Models/PoolCandidate.php
tristan-orourke, line 286: In addition to a flexible dynamic scope, it would be handy to have a scope that [doesn't include any args at all](https://laravel.com/docs/9.x/eloquent#local-scopes).

```
    public function scopeIsAvailable(Builder $query): Builder
    {
            return $query->where('pool_candidate_status', 'AVAILABLE');
    }

```

api/graphql/schema.graphql
```diff
@@ -558,6 +559,7 @@ input PoolCandidateFilterInput {
     operationalRequirements: [OperationalRequirement] @builder(method: "App\\Models\\PoolCandidate@filterByOperationalRequirements")
     workRegions: [WorkRegion] @builder(method: "App\\Models\\PoolCandidate@filterByWorkRegions")
     pools: [PoolFilterInput] @builder(method: "App\\Models\\PoolCandidate@filterByPools")
+    status: PoolCandidateStatus @scope(name: "PoolCandidateStatus")
```
tristan-orourke, line 562: This can be removed. Instead we'll add the `available` scope to the countPoolCandidates and searchPoolCandidates queries.

api/graphql/schema.graphql
```diff
@@ -574,6 +576,7 @@ input UserFilterAndOrderInput {
     skills: [ID] @builder(method: "App\\Models\\User@filterBySkills")
     isGovEmployee: Boolean @scope
     orderBy: String   # This needs to be the name of the field in the database eg. first_name
+    status: PoolCandidateStatus @builder(method: "App\\Models\\User@filterByPoolCandidateStatus")
```
tristan-orourke, line 579: Since this only makes sense as part of PoolCandidates, it should be part of UserPoolFilterInput instead. Also, I think we should make it use an array of PoolCandidateStatuses, allowing for more flexibility.

api/graphql/schema.graphql
```diff
@@ -925,6 +928,7 @@ input CreatePoolCandidateFilterInput {
     operationalRequirements: [OperationalRequirement] @rename(attribute: "operational_requirements")
     workRegions: [WorkRegion] @rename(attribute: "work_regions")
     pools: PoolBelongsToMany
+    status: PoolCandidateStatus @rename(attribute: "pool_candidate_status")
```
tristan-orourke, line 931: Can be removed

tristan-orourke (commented):
Okay, I think I fixed all my comments to make sense again.

api/app/Models/PoolCandidate.php
patcon: Realizing I don't understand something about graphql setup that makes this required, but also would suggest it deal with all types so as not to be misleading

api/graphql/schema.graphql
```diff
@@ -547,6 +547,7 @@ enum CANDIDATE_EXPIRY_FILTER {
 input UserPoolFilterInput {
     pools: [ID!]!
     expiryStatus: CANDIDATE_EXPIRY_FILTER = ACTIVE
+    status: [String] @scope(name: "PoolCandidateStatus")
```
tristan-orourke, line 550:
```suggestion
    statuses: [PoolCandidateStatus] @scope(name: "PoolCandidateStatuses")
```

api/app/Models/User.php
```diff
@@ -182,6 +182,7 @@ public function filterByPools(Builder $query, array $poolCandidates): Builder
                     } else if ($poolCandidates['expiryStatus'] == ApiEnums::CANDIDATE_EXPIRY_FILTER_EXPIRED) {
                         $query->whereDate('expiry_date', '<', date("Y-m-d"));
                     }
+                    PoolCandidate::poolCandidateStatus(['AVAILABLE']);
```
tristan-orourke, line 185: We need to continue using the same queryBuilder object, and also use the statuses that were passed in the query.
```suggestion
                    $query->poolCandidateStatus($poolCandidates['statuses']);
```

tristan-orourke (changes requested):
Close! Thanks for being thorough about the unit tests. Along with my suggested changes, it would be good to add a couple more tests with other combinations of STATUSES filters for user search (including empty array, and several statuses at once).

tristan-orourke (approved):

PR #3021, opened 2022-06-10
Enable snazzy adminer theme

patcon:
Came across the baked-in adminer themes when messing with docker this evening. Seemed small and might be a little more delightful when adminer shows up on demos, but that's a stretch 🙃  

### What's here
- sets envvar for material design-y `hydra` theme
- updates hydra theme related to scroll bug
- bonus: sets envvar to default the server field to `postgres`

https://github.com/Niyko/Hydra-Dark-Theme-for-Adminer

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,6 +14,12 @@ services:
   adminer:
     image: adminer
     restart: always
+    environment:
+      - ADMINER_DESIGN=hydra2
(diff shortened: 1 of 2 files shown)


patcon:
Ah sorry, good call :) a file gets mounted, so this should work:

```
docker-compose down
docker-compose up --detach --build
```

esizer (approved):
Looking good! 😃

Issue #2988, opened 2022-06-06
Cypress Smoke Test


petertgiles:
This is my manual smoke test.  It might be good to turn this into a Cypress test.

Talent

    Navigate to /talent
    Expect that a non-zero candidate estimate comes up
    Scroll to the bottom and click Request Candidates
    Fill in junk data to the request form and click Submit Request
    Expect the Request created successfully! toast.

Admin

    Navigate to /admin
    If I'm logged in then click Logout
    Click Login
    Follow through the login procedure
    Click the Users link and expect to see the list of users
    Click the Edit link beside the first one
    Edit the first name of the user and click Submit
    Expect to see the edited name showing in the list

It doesn't include any profiles stuff so that should probably be added as well.

PR #3032, opened 2022-06-10
Fix search request submission

### LINKED ISSUE #3031: request creation in Talentsearch broken (local)
Attempting to submit a request errors at /talentsearch/request

error:
image

laravel log snippet
image

petertgiles:
h/t "Eagle Eyed" Vachan for finding the search submission bug.  Hopefully this sort of thing won't slip through once #2988 is in.

I think the problem was that [the EE fields are now in a sub-input](#2966).  Adding the spread directive puts them back where Eloquent expects them.

Suggested Test Steps
- [ ] Use the talent search site to submit a search request
- [ ] Use the admin site to verify search request was successfully saved

Closes #3031

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -920,7 +920,7 @@ input CreatePoolCandidateFilterInput {
     classifications: ClassificationBelongsToMany
     cmoAssets: CmoAssetBelongsToMany
     hasDiploma: Boolean @rename(attribute: "has_diploma")
-    equity: EquitySelectionsInput
+    equity: EquitySelectionsInput @spread
     languageAbility:LanguageAbility @rename(attribute: "language_ability")
     operationalRequirements: [OperationalRequirement] @rename(attribute: "operational_requirements")
     workRegions: [WorkRegion] @rename(attribute: "work_regions")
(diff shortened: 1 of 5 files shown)


vd1992:
that enables submission, but still have an issue with the create request form, it does not send the appropriate data
doing some prodding, by adding this button

image

I see that formvalues does not appear to acknowledge anything below the Summary of Filters header

image

vd1992:
Ah! My mistake! The only thing erroneous is the Equity object and that is because it isn't linked to the right variables!

vd1992:
sorry, hope you don't mind me commiting it myself to make it clearer what I meant above
okay...
now it submits almost everything and saves almost everything, what appears to remain non-functional, is that it won't sync pools and won't actually save the equity stuff to database

vd1992:
alright, it successfully saves to the database, so the issue is not there
image

the issue is admin-side, it fails to load the equity stuff, possibly due to it looking for different types
(viewing requests in /admin)

image

JamesHuf:
> This will help: Laravel accessors your mutators: https://laravel.com/docs/9.x/eloquent-mutators#defining-an-accessor https://laravel.com/docs/9.x/eloquent-mutators#defining-a-mutator

We currently have laravel 8.83.15 so we need to use the 8.x version of these: 
https://laravel.com/docs/8.x/eloquent-mutators#defining-an-accessor
https://laravel.com/docs/8.x/eloquent-mutators#defining-a-mutator

PR #3046, opened 2022-06-13
Move static files for frontend apps into conventional "public" dir

patcon:
Re-ticketed from #2217 (was poking at docker setup over the weekend)

tl;dr - Let's **consolidate static assets and use the "public" dir for corralling them**, as is conventional and least surprising :)

```
admin/public
├── config.sjs
├── images
│   ├── about.txt
│   ├── android-chrome-192x192.png
│   ├── android-chrome-512x512.png
│   ├── apple-touch-icon.png
│   ├── bg-mountains.jpg
│   ├── favicon-16x16.png
│   ├── favicon-32x32.png
│   ├── favicon.ico
│   ├── logo_canada.png
│   └── logo_goc_colour.svg
├── index.html
├── robots.txt
└── site.webmanifest
```

---

Currently, we tell webpack to specifically copy a few static files out of the root app dir when creating the build. This piecemeal moving of files generally makes effective use of **docker layer caching** more tedious, as each file should ideally be cherrypicked in a dockerfile COPY as each build step that requires it gets run. (doing COPY on whole app dir contents right away, leads to caches getting invalidated more often than we'd prefer)

The **convention of the popular react frameworks** (e.g., [nextjs](https://nextjs.org/docs/basic-features/static-file-serving), [nuxtjs](https://v3.nuxtjs.org/guide/directory-structure/public/), [CRA](https://create-react-app.dev/docs/using-the-public-folder/), etc) is to keep **each app's static assets in a single dir called `public/`**, whose contents are copied into the root of the build directory. By convention, the index.html template is also stored in "public/" (we kept it in `src/views`, a layover from laravel)


esizer (approved):
love it ❤️ 

I merged some changes to talent search that will need to be resoled but this is much cleaner 👍

PR #3066, opened 2022-06-14
Speed up phpunit test runs

patcon:
While trying to sort out a potential bug, I noticed that there was a faster way to do setUp in our phpunit tests. Full test run came **down to 12 sec from 45 sec**.

### Screenshot
Screen Shot 2022-06-14 at 2 49 58 PM

### To Validate
Just run this (might be faster for you, as my docker is slow):
```
docker-compose exec --workdir /var/www/html/api php php vendor/bin/phpunit --verbose
```

---

For anyone curious, this is **how it works**: It runs migration only once at start of each test run (rather than before each test), and keeps every test in a DB transaction that's rolled back after each test completes.

Video explaining it before it was brought into laravel: https://adamwathan.me/2016/11/14/a-better-database-testing-workflow-in-laravel/

### DIFF
--- a/api/tests/Feature/AuthServiceProviderTest.php
+++ b/api/tests/Feature/AuthServiceProviderTest.php
@@ -8,11 +8,11 @@ use Lcobucci\JWT\Token\DataSet;
 use Tests\TestCase;
 use Mockery\MockInterface;
 use Symfony\Component\HttpKernel\Exception\HttpException;
-use Illuminate\Foundation\Testing\DatabaseMigrations;
+use Illuminate\Foundation\Testing\RefreshDatabase;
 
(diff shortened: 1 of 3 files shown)


esizer (approved):
Cool! 😎 

```sh
[gc-digital-talent] docker-compose exec --workdir /var/www/html/api php php vendor/bin/phpunit --verbose                   faster-phpunit-tests 
PHPUnit 9.5.20 #StandWithUkraine

Runtime:       PHP 7.4.30
Configuration: /var/www/html/api/phpunit.xml

..........................................                        42 / 42 (100%)

Time: 00:08.639, Memory: 52.50 MB

OK (42 tests, 144 assertions)
```

PR #3069, opened 2022-06-14
Feature - Add department to user

### LINKED ISSUE #2990: Add "department" to Government Info profile form

yonikid15:
Resolves #2990 

# Notes
- This adds a `department` column/field to the users table/user model in schema. On the frontend, in the `<GovernmentInfoForm/>` component there should be a select input for choosing a department. Here are a list of pages where this form exists: 
- [profile/create-account](http://localhost:8000/en/talent/profile/create-account), 
- [/profile/government-info](http://localhost:8000/en/talent/profile/government-information). 
- There's also an addition to the "Government Information" section of the [/profile](http://localhost:8000/en/talent/profile) 

## Acceptance Criteria
- [x] Add department to User and Applicant type in schema.
- [x] Create migration to add department column to users table.
- [x] Add department to mutations and queries for gov info form and create account page.
- [x] Add select input for departments on gov info form.


api/database/factories/UserFactory.php
```diff
@@ -87,6 +89,7 @@ public function definition()
             ]),
             'is_gov_employee' => $this->faker->boolean(),
             'interested_in_later_or_secondment' => $this->faker->boolean(),
+            'department' => $randomDepartment ? $randomDepartment->id : null,
```
vd1992, line 92: for testing purposes, it might be handy to have department contingent on is_gov_employee so you don't end up with people with a department but are not gov employees

vd1992 (changes requested):
need to ensure departments are wiped if toggled back to not an employee
image

vd1992:
otherwise
great work!

yonikid15:
> need to ensure departments are wiped if toggled back to not an employee image

Nice catch, this is fixed with [Ensure departments is wiped if toggled back to not an employee](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3069/commits/bed684bbe4ca2a26978c9acbaf7fddf34c117fa9)

PR #3074, opened 2022-06-15
Remove min value from PoolCandidate expiry date validation

### LINKED ISSUE #3073: Admins shouldn't be restricted from using past dates in Expiry Date
The expiry date of the admin PoolCandidateForm only allows future dates in validation. In fact, admins may want to set a past date retroactively.

tristan-orourke:
Resolves #3073


vd1992 (approved):
👍

PR #3082, opened 2022-06-16
Bug - 3080 - Fix Create Account Redirect

### LINKED ISSUE #3080: Bug: Redirect to Create Account not completing
## Summary

When a user registers they are redirected to `/talent/profile/create-account`. Currently, the redirect occurs but the wrong component gets mounted (Not Authorized or Profile).

esizer:
Resolves #3080 

## Summary

Following the login redirect, there was a `return null;` added after the navigation to the new path. This should only be used when navigating away from the site. Returning null caused the application to navigate but never mount the new component and instead showed the last component that was mounted.

Also, added more safety around the authorisation, waiting until the roles have loaded before checking them.

### DIFF
--- a/frontend/cypress/integration/talentsearch/direct-intake.spec.js
+++ b/frontend/cypress/integration/talentsearch/direct-intake.spec.js
@@ -56,6 +56,7 @@ describe("Talentsearch Direct Intake Page", () => {
       cy.visit('/en/browse/pools');
 
       cy.location('pathname').should('eq', '/en/talent/profile/create-account');
+      cy.contains("successfully logged in");
     });
   })
 
--- a/frontend/common/src/helpers/router.tsx
+++ b/frontend/common/src/helpers/router.tsx
@@ -138,8 +138,7 @@ export const useRouter = (
           authorizationLoaded &&
           empty(loggedInEmail)
         ) {
-          redirect(welcomeRoute);
-          return null;
+          navigate(welcomeRoute);
         }
 
         // handling a redirect
@@ -161,7 +160,7 @@ export const useRouter = (
         let isAuthorized: boolean;
 
         // if there is a list of authorized roles required then let's see if the user is authorized
-        if (authorizationRequired) {
+        if (authorizationLoaded && authorizationRequired) {
           // the user is considered authorized if there are no roles needed or they have at least one of the required roles
           isAuthorized =
             authorizedRoles.length === 0 ||


petertgiles:
When I put in a new sub on login I'm sent to /create-account.  But after filling in the account I'm looped back again to /create-account even though I verified the email was saved for the user.  Is that happening for you?  Is that a totally unrelated bug?

esizer:
> When I put in a new sub on login I'm sent to /create-account. But after filling in the account I'm looped back again to /create-account even though I verified the email was saved for the user. Is that happening for you? Is that a totally unrelated bug?
> 
> Edit: If I put `/talent/profile` in the nav bar the router will let me manually go there.

Hmm, I am not able to reproduce either of these issues. Can you check the graphql response to see if the email is being returned properly? I used the existing `empty` function in helpers to check but im not sure that is totally sufficent.

```tsx
export function empty<T>(
  value: T | null | undefined,
): value is null | undefined {
  return value === null || value === undefined;
}
```

Since, if we return an empty string, or false this will return false. Also, we may want to consider updating the undefined check to `type of value === "undefined"` which is a more consistent way of checking if something is undefined.

patcon:
Seems to be working for me. Here are my steps:

1. Rebuilt containers
2. `docker-compose run --rm maintenance bash refresh_all.sh`
3. ensure `AUTH_DEFAULT_USER` is unset in `api/.env`
4. clear cookies and local storage
5. visit http://localhost:8000/talent
6. Click login
7. Submit new email
8. Get redirected back to http://localhost:8000/en/talent/profile/create-account

frontend/cypress/integration/talentsearch/direct-intake.spec.js
```diff
@@ -56,6 +56,7 @@ describe("Talentsearch Direct Intake Page", () => {
       cy.visit('/en/browse/pools');
 
       cy.location('pathname').should('eq', '/en/talent/profile/create-account');
+      cy.contains("successfully logged in");
```
patcon, line 59: Good catch. If it doesn't feel out-of-scope, consider using this as opp to add an aria role (whichever one fits best) to the "notification" in which your text is found, and using that instead of a whole page match.

```suggestion
      cy.findByRole('status', /successfully logged in/i);
```

petertgiles:
Confirming that failure to redirect after filling in the welcome page is out-of-scope for this PR.  I opened #3113.

esizer:
> Confirming that failure to redirect after filling in the welcome page is out-of-scope for this PR. I opened #3113.

Thanks @petertgiles 🙏

Issue #2640, opened 2022-05-03
Use backend filtering on User table search bar


tristan-orourke:
Search bar currently runs a client-side filter on any text fields. Convert it add a filter to the query, now that the table is using backend pagination. The text should search the name, email and phone number fields.

For this issue, don't worry about specifying a single field to search on.

tristan-orourke:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/GC-Digital-Talent-6092dc05157fa30010edaef9/issues/GCTC-NTGC/gc-digital-talent/2640?planning-poker) with ZenHub @patcon

tristan-orourke:
Please try adding some acceptance criteria as well

patcon:
@petertgiles and eric working together to refactor. back to draft.

esizer:
@petertgiles Looks like this was resolved with #3220 are we safe to close?

PR #3084, opened 2022-06-16
Cypress Workflow Tests

### LINKED ISSUE #2988: Cypress Smoke Test

petertgiles:
This branch adds some basic tests to verify essential functionality of the major sections of the site in a realistic way.

- [x] Talent search smoke test
- [x] Admin smoke test
- [x] profiles smoke test

Note: `searchpage.spec.js` was renamed to `search-page.spec.js` and the primary test was extracted as a foundation for the search workflow test.

Testing instructions: 
- Rerun seeder `php artisan migrate:fresh --seed`
- `npm run e2e`

Bonus: Prettier was set up at the `/frontend` level so that it can be used for linting Cypress tests

Closes #2988


frontend/cypress/integration/admin/admin-smoke-tests.spec.js
```diff
@@ -0,0 +1,68 @@
...
+import { getInputByLabel } from "../../support/helpers";
+
+describe("Admin Smoke Tests", () => {
+  const loginViaUI = (role) => {
+    cy.fixture("users.json").then((users) => {
+      const user = users[role];
+      cy.get("input[name=username]").type(user.email);
+    });
+    cy.findByText("Sign-in").click();
+  };
```
patcon, line 15: Unless I'm misunderstanding how "smoke test" might be different, I'd propose we just remove this and use `cy.login('admin')` below

The idea with E2E tests is usually that the UI is only tested once (in this case, in `auth.test.js`), and then after that things are done with more efficient commands (that ideally use `cy.request` directly) and are no longer exercising the UI (since that's already covered).

frontend/cypress/integration/admin/admin-smoke-tests.spec.js
```diff
@@ -0,0 +1,68 @@
...
+
+  beforeEach(() => {
+    // common requests
+    setUpCommonGraphqlIntercepts();
+    // page specific requests
+    cy.intercept("POST", "/graphql", (req) => {
+      aliasQuery(req, "AllUsers");
+      aliasQuery(req, "User");
+      aliasMutation(req, "UpdateUserAsAdmin");
+    });
```
patcon, line 25: Curious your thoughts here -- my use of "common requests" is just a convenience I made up, thinking it would make it clearer to set up and see them all in one place. Intercepts seem pretty lightweight, so it didn't seem to impose a cost

My sense if that we should just keep it simple and either:
1. set them up for each test file and have no "common" intercepts (how the docs show it), or
2. store them all in one place (`setUpCommonGraphqlIntercepts()`) and avoiding thinking about "common" vs page-specific

I don't have a preference between those, but don't think we should do anything in-between :)

frontend/cypress/integration/admin/admin-smoke-tests.spec.js
```diff
@@ -0,0 +1,68 @@
...
+
+    cy.visit("/en/admin");
+  });
+
+  it("Searches for a user, reviews the profile, and edits the phone number", () => {
+    // hit the home page, login, hit the dashboard
+    cy.contains("h1", "Home");
+    cy.contains("a", "Login").click();
+    loginViaUI("admin");
```
patcon, line 34:
```suggestion
    cy.login("admin");
```

frontend/cypress/integration/admin/admin-smoke-tests.spec.js
```diff
@@ -0,0 +1,68 @@
...
+      aliasQuery(req, "User");
+      aliasMutation(req, "UpdateUserAsAdmin");
+    });
+
+    cy.visit("/en/admin");
+  });
+
+  it("Searches for a user, reviews the profile, and edits the phone number", () => {
+    // hit the home page, login, hit the dashboard
+    cy.contains("h1", "Home");
```
patcon, line 32: Wondering about whether we should ever use cypress' native `contains` command, or instead the testing-library query selectors 🤔 Will leave comment in thread

```suggestion
    cy.findByRole("heading", { name: /Home/i });
```

frontend/.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
+  "semi": true,
+  "trailingComma": "all",
+  "singleQuote": false,
+  "printWidth": 80,
+  "tabWidth": 2,
+  "endOfLine": "auto"
+}
```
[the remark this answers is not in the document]
petertgiles: Yeah, I think you're right.  I'll try this out.

frontend/cypress/integration/admin/admin-smoke-tests.spec.js
petertgiles: I think I'd prefer option 1.  Each page tends to have its own set of customized requests and with the alias* functions you created it's really not much code to register them.  I'll try removing the common intercepts function.

frontend/cypress/integration/admin/admin-smoke-tests.spec.js
petertgiles: You're probably right that findByRole is better.  I'll have to do some experimentation.  I might have to add some roles, too, which is probably a good side-effect. :laughing:

frontend/cypress/support/helpers.js
```diff
@@ -0,0 +1,9 @@
+// https://glebbahmutov.com/cypress-examples/6.5.0/recipes/form-input-by-label.html#reusable-function
+export const getInputByLabel = (label) => {
+  return cy
+    .contains('label', label)
+    .invoke('attr', 'for')
+    .then((id) => {
+      cy.get('#' + id)
+    })
+}
```
patcon, line 9: consider using `cy.findByRole('textbox', ...)`

for testing-library, but confirms this is the right role and possible caveat: https://stackoverflow.com/questions/64864453/unable-to-get-the-name-of-the-input-when-using-getbyrole-while-testing-with-test

patcon: I think the findByRole approach might be so short that it no longer warrants a helper or command?

And it's a little fuzzy, but I'd recommend using custom commands (in `frontend/cypress/support/commands.js`) for something like this that seems to be valuable to use elsewhere. Custom functions are introduced pretty early in cypress docs, and it seems the the expectation is they'll do a lot of lifting. (I'm a little unsure why the official docs recommended helper functions for the graphql stuff, but I'm guessing it's just due to `cy.intercept` being a bit unique, and maybe better to batch 🤔 )

frontend/cypress/integration/admin/admin-smoke-tests.spec.js
```diff
@@ -0,0 +1,68 @@
...
+    cy.contains("a", "Manage users").click();
+    cy.wait("@gqlAllUsersQuery");
+    getInputByLabel("Search").clear().type("Applicant");
+    cy.contains("a", "applicant@test.com").click();
+
+    // exercise profile page
+    cy.contains("h1", "Candidate Details");
+    cy.contains("button", "Print Profile");
+    cy.contains("General Information").click();
+    cy.contains("span", "About");
```
patcon, line 48: This might be a place where exercising aria attributes is the thing to do to know if the tab is selected/loaded?
https://testing-library.com/docs/queries/byrole/#selected

Or maybe it's that it should be finding something in the tab content related to "General Information", which is more indicative of the tab being loaded from the screen-reader perspective?

frontend/cypress/integration/admin/admin-smoke-tests.spec.js
```diff
@@ -0,0 +1,68 @@
...
+    cy.contains("Candidate Profile").click();
+    cy.contains("span", "About");
+
+    // find the applicant user to play with
+    cy.contains("a", "Users").click();
+    //cy.wait("@gqlAllUsersQuery");  // will be cached and not fired a second time
+    getInputByLabel("Search").clear().type("Applicant");
+    cy.contains("td", "applicant@test.com")
```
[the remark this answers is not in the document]
patcon: Maybe this?

```suggestion
    cy.findByRole("cell", { name: "applicant@test.com" })
```

petertgiles:
> Should we be removing things from these "smoke tests" if they're already tested in other specs? E.g., if we tested redirect to `/dashboard` in another spec, I don't think we'd test it again here, but just assume it happens, right?
> 
> I was thinking these smoke tests were just to get us started with some big over-sized tests, but I was expecting them to get whittled away as we write more atomic tests. Does that sound right for you?

This is a really interesting question.  I'm beginning to regret my choice to call them smoke tests.  I think really what I want is a set of tests to exercise the common use cases of the application.  I don't think I want to make them as small and atomic as possible, either.  The strength of e2e is that you're moving all the pieces together and in sequence, right?  Maybe I do want to test that redirect to dashboard works properly again, after running the previous steps in the use case.  Though it would better if that repeated set of steps was DRY in a reusable command.  Does that make sense or am I missing the point?  :smile: I think the place for small fast tests is unit testing?

frontend/.prettierrc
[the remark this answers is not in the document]
petertgiles: Applied in https://github.com/GCTC-NTGC/gc-digital-talent/pull/3084/commits/6aae33b9caaeb6b9bc00f244321a0776f1755e22

frontend/cypress/integration/admin/admin-smoke-tests.spec.js
petertgiles: Applied in https://github.com/GCTC-NTGC/gc-digital-talent/pull/3084/commits/383b3126d8c6c36e4ef791e4cb651144a270797b

frontend/cypress/integration/admin/admin-smoke-tests.spec.js
petertgiles: I've removed all the contains and switched to findBy selectors.  I agree that it's probably a better way to go.  Applied in https://github.com/GCTC-NTGC/gc-digital-talent/pull/3084/commits/383b3126d8c6c36e4ef791e4cb651144a270797b

frontend/cypress/support/helpers.js
petertgiles: Yeah, the findByRole with name does the same thing as the helper so I removed it in https://github.com/GCTC-NTGC/gc-digital-talent/pull/3084/commits/383b3126d8c6c36e4ef791e4cb651144a270797b

frontend/cypress/integration/admin/admin-smoke-tests.spec.js
```diff
@@ -0,0 +1,68 @@
...
+    getInputByLabel("Search").clear().type("Applicant");
+    cy.contains("a", "applicant@test.com").click();
+
+    // exercise profile page
+    cy.contains("h1", "Candidate Details");
+    cy.contains("button", "Print Profile");
+    cy.contains("General Information").click();
+    cy.contains("span", "About");
+    cy.contains("Candidate Profile").click();
+    cy.contains("span", "About");
```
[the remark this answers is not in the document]
petertgiles: Applied in https://github.com/GCTC-NTGC/gc-digital-talent/pull/3084/commits/383b3126d8c6c36e4ef791e4cb651144a270797b

frontend/cypress/integration/admin/admin-smoke-tests.spec.js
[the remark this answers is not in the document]
petertgiles: I've updated the selector to find the row and chain down as suggested.  I'm not sure why `findByRole("link", { name: "applicant@test.com" })` doesn't work. :shrug: Oh well, this table will be changing soon anyway.

https://github.com/GCTC-NTGC/gc-digital-talent/pull/3084/commits/cc78fa41ff8267b51583a4ef3645fe72caa3b58c

frontend/cypress/integration/talentsearch/profile-workflows.spec.js
```diff
@@ -0,0 +1,54 @@
...
+    cy.url().should("contain", "/profile");
+
+    // work location
+    cy.findByRole("link", { name: /Edit Work Location/i }).click();
+    cy.findByRole("textbox", { name: /Location exemptions/i })
+      .clear()
+      .type("Test Locations");
+    cy.findByRole("button", { name: /Save and go back/i }).click();
+    cy.wait("@gqlcreateWorkLocationPreferenceMutation");
+    cy.findByRole("alert")
```
brindasasi, line 48: this can be a command as well..probably verifyUserUpdatedSuccessfully()

frontend/cypress/integration/talentsearch/profile-workflows.spec.js
```diff
@@ -0,0 +1,54 @@
...
+    cy.findByRole("link", { name: /Edit About Me/i }).click();
+    cy.findByRole("textbox", { name: /Current city/i })
+      .clear()
+      .type("Test City");
+    cy.findByRole("button", { name: /Save and go back/i }).click();
+    cy.wait("@gqlUpdateUserAsUserMutation");
+    cy.findByRole("alert")
+      .findByText(/User updated successfully/i)
+      .should("exist")
+      .and("be.visible");
```
brindasasi, line 38: this can be a command as well..probably verifyUserUpdatedSuccessfully()

frontend/cypress/integration/admin/admin-smoke-tests.spec.js
patcon: Thanks for removing it from the other specs too!

frontend/cypress/integration/admin/admin-smoke-tests.spec.js
patcon: yeah, the little a11y tidy tasks from using the findByRole selector totally sold me on the testing-library ethos :)

frontend/cypress/integration/admin/admin-workflows.spec.js
```diff
@@ -0,0 +1,99 @@
...
+      .should("exist")
+      .and("be.visible");
+  });
+
+  it("Searches for a user and edits the phone number", () => {
+    // find the applicant user to edit
+    cy.findByRole("link", { name: /Manage users/i }).click();
+    cy.wait("@gqlAllUsersQuery");
+    cy.findByRole("textbox", { name: /Search/i })
+      .clear()
```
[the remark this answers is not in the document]
patcon: But on second thought, let's just create it, and we'll figure out our style later with concrete examples as we make them :)

frontend/cypress/integration/admin/admin-workflows.spec.js
```diff
@@ -0,0 +1,99 @@
...
+      .type("+10123456789");
+    cy.findByRole("button", { name: /Submit/i }).click();
+
+    cy.wait("@gqlUpdateUserAsAdminMutation");
+
+    cy.findByRole("alert")
+      .findByText(/User updated successfully/i)
+      .should("exist")
+      .and("be.visible");
+  });
```
patcon, line 98: This looks great! Should wait and check that the new phone number is actually reflected on the page?

patcon:
> I'm beginning to regret my choice to call them smoke tests

heh yeah, was having a similar thought about us maybe using the term differently than elsewhere :)

> The strength of e2e is that you're moving all the pieces together and in sequence, right? Maybe I do want to test that redirect to dashboard works properly again, after running the previous steps in the use case. Though it would better if that repeated set of steps was DRY in a reusable command. Does that make sense or am I missing the point? 😄 I think the place for small fast tests is unit testing?

Yeah, it's a funny line. I'm realizing the line between react component unit/integration testing and e2e is a bit fuzzy for me...! We didn't do unit tests when I used it before, so that dividing line wasn't something I needed a sensibility about

frontend/cypress/integration/admin/admin-workflows.spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+    // wait for table to rerender
+    cy.findByText("applicant@test.com")
+      .should("exist")
+      .and("be.visible");
+
+    cy.findByRole("table")
+      .findByRole("row", { name: /applicant/i })
+      .findByText("applicant@test.com") // findByRole link doesn't work here
+      .click();
```
patcon, line 41: Tried running these, and getting a failure. Seems that chaining requires using cypress' `.within()` command:
https://github.com/testing-library/cypress-testing-library/issues/132#issuecomment-853297907

Testing out if this fixes my failure. A bit unsure why it was working for you and not me 🤔

patcon: Ah wait, there's something wrong and I think I need to refresh_all. So don't dive in here :)

frontend/cypress/integration/talentsearch/profile-workflows.spec.js
petertgiles: This set of commands is so small that I wonder whether its worth it but it will be used throughout the app so I added it anyway.  https://github.com/GCTC-NTGC/gc-digital-talent/pull/3084/commits/5a8c25ca861dea017d60e06585c55a19324e102c

frontend/cypress/integration/admin/admin-workflows.spec.js
petertgiles: Good point!  Added in https://github.com/GCTC-NTGC/gc-digital-talent/pull/3084/commits/9c925d6cb5f25e1ff09b706bb996aeb8355b299f

frontend/cypress/integration/admin/admin-workflows.spec.js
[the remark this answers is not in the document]
petertgiles: Agree that a helper function is probably more appropriate since this won't be used widely through the app.  Added it here: https://github.com/GCTC-NTGC/gc-digital-talent/pull/3084/commits/9c925d6cb5f25e1ff09b706bb996aeb8355b299f

brindasasi (approved):
Awesome work! Looks great to me!

frontend/cypress/integration/admin/admin-workflows.spec.js
patcon: as mentioned in slack last week (i think?) findByRole is a bit wonky, so maybe worth abandoning it in places where it gives any trouble. still new to cypress-testing-library, but have only seen this wonkiness within tables tbh, so might be small issue

petertgiles:
Blocked on #2640

PR #3103, opened 2022-06-16
Add default value for pool status in migration

### LINKED ISSUE #3102: Migrations unable to run with existing pools in database
# Description
If pools already exist in database, UpdatePoolsTableWithKeyTasksAndStatus will fail because we add a non-nullable column without setting a default value.

tristan-orourke:
Resolves #3102


api/database/helpers/ApiEnums.php
```diff
@@ -105,4 +105,6 @@ public static function salaryRanges() : array
     const CANDIDATE_STATUS_EXPIRED = 'EXPIRED';
     const CANDIDATE_STATUS_PLACED_CASUAL = 'PLACED_CASUAL';
 
+    const POOL_STATUS_NOT_TAKING_APPLICATIONS = 'NOT_TAKING_APPLICATIONS';
+    const POOL_STATUS_TAKING_APPLICATIONS = 'TAKING_APPLICATIONS';
```
patcon, line 109: Should this be used in other places where `TAKING_APPLICATIONS` and `NOT_TAKING_APPLICATIONS` is used?

Just these, I think:
`api/database/factories/PoolFactory.php`
`api/database/seeders/PoolSeederUat`

patcon (approved):
otherwise LGTM!

PR #3107, opened 2022-06-16
filter users by admin/users table search

### LINKED ISSUE #2812: Allow Users query to be filtered by email, name, or phone number
## Description

Add an optional filter to the paginated Users query which allows for filtering by email, name or phone number.

This is to support the table filter shown [here](https://www.figma.com/proto/XS4Ag6GWcgdq2dBlLzBkay/Admin-v3-(Profile-release)?node-id=465%3A8698&scaling=min-zoom&page-id=0%3A1&starting-point-node-id=410%3A1984&show-proto-sidebar=1)

screenshot

## Acceptance Criteria
- [x] email, name, phone number, or a general search field are optional filters that can be applied to users and usersPaginated queries
- [x] The different fields should act as an AND filter (eg only return users that match the email AND the phone number)
- [x] the "general" search field should work as an orWhere query, applied to email, firstName, lastName, and phoneNumber
- [x] a phpunit test for this

## Todo
- [ ] How to deal with firstName/lastName db fields vs single name search field?

vd1992:
resolves #2812 
 
Adds additional filtering to the UserFilterOrder thing, for UsersPaginated. 
To test, play around in GraphQL and see if anything appears off. This is not connected to anything frontend yet. 

Rolled in that one linting alert thing on all PRs because it was bugging me and it wasn't something that merited its own issue+PR

### DIFF
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -472,6 +472,49 @@ RAWSQL2;
         return $query;
     }
 
+    public function filterByGeneralSearch(Builder $query, ?string $search): Builder
+    {
+        if ($search) {
+            $query->where('first_name', "like", "%{$search}%")
+                  ->orWhere('last_name', "like", "%{$search}%")
+                  ->orWhere('email', "like", "%{$search}%")
+                  ->orWhere('telephone', "like", "%{$search}%");
+        }
+        return $query;
+    }
+
+    public function scopeTelephone(Builder $query, ?string $telephone): Builder
+    {
+        if ($telephone) {
+            $query->where('telephone', 'like', "%{$telephone}%");
+        }
+        return $query;
+    }
+
+    public function scopeEmail(Builder $query, ?string $email): Builder
+    {
+        if ($email) {
+            $query->where('email', 'like', "%{$email}%");
+        }
+        return $query;
+    }
+
+    public function scopeFirstName(Builder $query, ?string $firstName): Builder
+    {
+        if ($firstName) {
+            $query->where('first_name', 'like', "%{$firstName}%");
+        }
+        return $query;
+    }
+
+    public function scopeLastName(Builder $query, ?string $lastName): Builder
+    {
+        if ($lastName) {
+            $query->where('last_name', 'like', "%{$lastName}%");
+        }
+        return $query;
+    }
+
     public function scopeIsGovEmployee(Builder $query, bool $isGovEmployee): Builder
     {
         if ($isGovEmployee) {
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -476,10 +476,10 @@ RAWSQL2;
     {
         if ($search) {
             $query->where(function($query) use ($search) {
-                $query->where('first_name', "like", "%{$search}%")
-                      ->orWhere('last_name', "like", "%{$search}%")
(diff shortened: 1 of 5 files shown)


vd1992:
okay, I played around a bit and changed how name filtering was working, do you like it?
@tristan-orourke

petertgiles:
I just took a quick peek - seems good so far.  Just a quick caution - check that your OR clauses are being built in a sub query.  I can't tell just by looking at the code but it's a red flag for me.  Ref: https://github.com/GCTC-NTGC/gc-digital-talent/blob/953eca2e388b665059ff2eb736215abd5fe6c39e/api/app/Models/PoolCandidate.php#L84

vd1992:
> I just took a quick peek - seems good so far. Just a quick caution - check that your OR clauses are being built in a sub query. I can't tell just by looking at the code but it's a red flag for me. Ref:
> 
> https://github.com/GCTC-NTGC/gc-digital-talent/blob/953eca2e388b665059ff2eb736215abd5fe6c39e/api/app/Models/PoolCandidate.php#L84

👍

api/app/Models/User.php
```diff
@@ -472,6 +472,49 @@ public function filterByEquity(Builder $query, array $equity): Builder
...
+    public function filterByName(Builder $query, ?string $name): Builder
+    {
+        if ($name) {
+            $splitName = explode(" ", $name);
+            $query->where(function($query) use ($splitName) {
+                foreach($splitName as $index => $value){
+                    $query->where('first_name', "like", "%{$value}%")
+                        ->orWhere('last_name', "like", "%{$value}%");
+                }
+            });
```
tristan-orourke, line 497: I think this is good, in terms of OR groupings? Looks to me like this will require that each segment of splitName match at least one firstName or lastName, which makes sense to me. Of course, you'll want some good tests for this.

tristan-orourke (commented):
Yeah, I think splitting the name up like that makes sense.

As @petertgiles said, be very careful with OR groupings. It looks right to me, but test it thoroughly.

Lastly, I recommend using "ilike" instead of "like" to make the match case-insensitive: postgresqltutorial.com/postgresql-tutorial/postgresql-like/

JamesHuf (approved):
LGTM! 👍

api/tests/Feature/UserTest.php
```diff
@@ -2154,4 +2154,223 @@ public function testFilterByClassificationToSalaryWithPools(): void
...
+                'usersPaginated' => [
+                    'paginatorInfo' => [
+                        'total' => 2
+                    ]
+                ]
+            ]
+        ]);
+
+        // Assert filtering for last name in general search returns correct count
+        $this->graphQL(/** @lang Graphql */ '
```
tristan-orourke, line 2330: I'm most concerned about how the name query works, so could we have a few more tests for that?
eg
test that a full name works (bob rob returns 1 result)
test that two name segments work (bo ro returns 1 result)
test that name segments work in incorrect order (ro bo returns 1 result)

tristan-orourke (changes requested):
I'm sure its good, but there are a few more tests I'd like to see.

api/tests/Feature/UserTest.php
vd1992: added

tristan-orourke (approved):

Issue #2646, opened 2022-05-03
Frontend pagination component which supports backend pagination


tristan-orourke:
Build a version of the pagination component which can connect to a backend paginated query. The number of items returned from the query, and the number shown in the table, shouldn't have to match. It may need to distinguish between the parameters (eg filters) of the query changing (which necessitates a full re-querying of results), and reaching the end of the current set of results.

## Screenshot: Prototype
[![image][image]][figma]
<sup>[Click screenshot to visit Figma prototype.][figma]</sup>

   [image]: https://user-images.githubusercontent.com/305339/166570455-3fa34807-6682-4f9b-a3d3-4b67fead9116.png
   [figma]: https://www.figma.com/proto/XS4Ag6GWcgdq2dBlLzBkay/Admin-v3?node-id=465%3A8698&scaling=min-zoom&page-id=0%3A1&starting-point-node-id=410%3A1984&show-proto-sidebar=1&hide-ui=1

tristan-orourke:
Wait for backend paginated query to be complete, so we know the shape of paginated results.

PR #3114, opened 2022-06-17
WIP: Feature - 2654 - New User Table Style

### LINKED ISSUE #2654: Styling updates on User table (placement of buttons, etc)
Update the colors and style of the table itself, and the look and placement of buttons (search, filters, columns, new pool, etc). Even if a button is not used or doesn't exist on the current version, add a non-functional version to match the prototype.

[Prototype](https://www.figma.com/proto/XS4Ag6GWcgdq2dBlLzBkay/Admin-v3?node-id=465%3A8698&scaling=min-zoom&page-id=0%3A1&starting-point-node-id=410%3A1984&show-proto-sidebar=1)

image

esizer:
Resolves #2654 

## Summary

This updates the `<Table />` component to match the new style provided in the [prototype](https://www.figma.com/proto/XS4Ag6GWcgdq2dBlLzBkay/Admin-v3?node-id=465%3A8698&scaling=min-zoom&page-id=0%3A1&starting-point-node-id=410%3A1984&show-proto-sidebar=1).

## Notes

This is notably missing the dropdown in the search form for the column to search on. This was intentional since I believe we should source a headless component to handle these types of components. Something like [headlessui](https://headlessui.dev/react/popover#accessibility-notes) or [reachui](https://reach.tech/menu-button) is something that provides:

- Relevant `aria-*` attributes
- Focus management
- Void click to close

We used `@reach/ui` for the `<Dialog />` component, and it could be useful here as well. However, if we expect content other than `<button>` or `<a>` to exist, we may need to look into others.

## Testing

- Navigate to `/admin`
- Login with admin@test.com
- Navigate to `/admin/users`

Right now, search and filters do not function. This will be added with a few other issues this sprint.

## Screenshot

Screen Shot 2022-06-17 at 3 08 06 PM

### DIFF
--- a/frontend/admin/src/js/components/Table/Table.tsx
+++ b/frontend/admin/src/js/components/Table/Table.tsx
@@ -12,7 +12,6 @@ import {
 import { Button, Link } from "@common/components";
 import Pagination from "@common/components/Pagination";
 import { FilterIcon, PlusIcon, TableIcon } from "@heroicons/react/outline";
-import GlobalFilter from "../GlobalFilter";
 import SortIcon from "./SortIcon";
 import SearchForm from "./SearchForm";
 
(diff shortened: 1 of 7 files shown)


petertgiles:
I think the Reach dialogs have been working well for us so I'd be fine with using that library again for this.  On the page you linked it seems like you can put whatever you want in the MenuItem?

esizer:
> On the page you linked it seems like you can put whatever you want in the MenuItem?

You technically can, the only weird thing is adding a checkbox like we are in the columns dropdown. Since each item in the menu should have a function which leads to the menu being closed upon selecting it, it may make this interaction odd.

However, we could replace the columns dropdown with a dialog 🤔

petertgiles:
> However, we could replace the columns dropdown with a dialog thinking

@Jerryescandon do we know what's in this button?  Is it a menu with checkboxes or a dialog?  It sounds like a dialog would be more a11y friendly.
image

Jerryescandon:
> 

Right now I just wanted to update the current one (Checkbox list of columns). But had no plans for updating the functionality of it during this spring.  But I agree making it a dialog seems the better path. If we have time and are comfortable doing it without a design I say go for it.

petertgiles:
> Since each item in the menu should have a function which leads to the menu being closed upon selecting it, it may make this interaction odd.

A bit of a cheesy in-between would be a list of items that toggle the column and immediately close when selected.  So if you wanted to toggle a bunch of columns it would be a hassle since you would have to open the menu several times.  Maybe it's an option until we get a design for a dialog?

Jerryescandon:
I can't see the one we have right now, but I think is something like the left side one. So I figured the easiest path would be to throw that into a dialog? maybe?

image

esizer:
> A bit of a cheesy in-between would be a list of items that toggle the column and immediately close when selected. So if you wanted to toggle a bunch of columns it would be a hassle since you would have to open the menu several times. Maybe it's an option until we get a design for a dialog?

Sounds good! Our dialogs are fairly easy to implement so maybe I'll give that a go and see what we think. Otherwise, I can revert back to using the same as the search dropdown.

Thanks for looking at this, I think I can proceed now 👍

esizer:
> I can't see the one we have right now, but I think is something like the left side one. So I figured the easiest path would be to throw that into a dialog? maybe?

Thanks @Jerryescandon ! I'll proceed using this 😄

frontend/admin/src/js/components/Table/Table.tsx
```diff
@@ -220,20 +310,59 @@ function Table<T extends Record<string, unknown>>({
...
+              defaultMessage: "Selected actions:",
+              description: "Label for action buttons in footer of admin table.",
+            })}
+          </p>
+          <Spacer>
+            <Button type="button" mode="solid" color="primary">
+              {intl.formatMessage({
+                defaultMessage: "Download XML",
+                description:
+                  "Text label for button to download a csv of items in a table.",
```
vd1992, line 337: message says xml, description says csv

esizer: Good catch! Fixed with [update i18n description for xml download](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3114/commits/334ac437532223bbb14e5e61f9b787526dbf8883)

vd1992:
resizing the table appears to break the table footer, pagination slides over the download button

https://recordit.co/Y2u2Do02oc

currently to navigate to a user profile, you click on the email rather than the name? is this intentional, it feels a bit odd since normally clicking on an email leads to email handler stuff
though, I see the text pop-up that explains what happens on hover 

image

esizer:
> resizing the table appears to break the table footer, pagination slides over the download button

This is intended to be replaced in #2646 

> currently to navigate to a user profile, you click on the email rather than the name? is this intentional, it feels a bit odd since normally clicking on an email leads to email handler stuff

I'm not sure, this is was not defined in the issue itself or any other acceptance criteria. The purpose of this PR is to update the styling of the table and not the functionality.

vd1992 (changes requested):
a bit of a scattered review I started but, see the comments strewn above
as well as, table sort of email and languages appears to not work

https://recordit.co/xUvfALEFpC

esizer:
> as well as, table sort of email and languages appears to not work

Noted, though I don't think that falls within the scope of this PR since it is specifically related to styles of the table. Can you create an issue for this?

vd1992:
> > as well as, table sort of email and languages appears to not work
> 
> Noted, though I don't think that falls within the scope of this PR since it is specifically related to styles of the table. Can you create an issue for this?

[Done](https://github.com/GCTC-NTGC/gc-digital-talent/issues/3146)

vd1992 (approved):
I believe all my feedback has been addressed, so I will rescind my request. 
Great work 🥳

PR #3115, opened 2022-06-17
improve phone number usability

### LINKED ISSUE #2320: Improve usability of phone number form fields
# Description
Right now the API requires a specific format for telephone numbers, but it isn't the most obvious or what people are used to. 

Options:
1. We could make loosen the restriction altogether, just store a string in the database. Simple, easy. Downside - will end up with inconsistent data, making it hard to use in the future if we want to do something real with phone numbers.
2. Change the format - simply require a more familiar 123-123-1234 format, or (123)-123-1234. That only allows Canadian (and US) numbers - but that's maybe not so bad? 
3. Keep the international format in the database, but do some transformations when submitting to the database. Most flexible option, involves slightly more work.

WE'RE GOING WITH OPTION 1

## Acceptance Criteria
- [x] graphql schema still uses a custom scalar for phone number, but that scalar accepts any string as valid
- [x] any telephone number form fields likewise accept any string in frontend validation

vd1992:
resolves #2320 

Testing
- ensure nothing broke for changed forms
- phones in form successfully submit strings and save
- eye out for any weirdness that could cascade out as a result of changes


api/app/GraphQL/Scalars/PhoneNumber.php
```diff
@@ -23,6 +23,6 @@ class PhoneNumber extends Regex
 
     public static function regex(): string
     {
-        return '/^\+[1-9]\d{1,14}$/';
+        return '/./';
```
petertgiles, line 26: Could you update the description of this scalar?

frontend/admin/src/js/components/poolCandidate/CreatePoolCandidate.tsx
```diff
@@ -173,13 +172,13 @@ const UserFormSection: React.FunctionComponent<{
             defaultMessage: "Telephone",
             description: "Label displayed on the user form telephone field.",
           })}
-          type="tel"
+          type="text"
```
petertgiles, line 175: Why did you switch from `tel` to `text`?  Seems like that would be useful even without the pattern?

frontend/admin/src/js/components/poolCandidate/CreatePoolCandidate.tsx
```diff
@@ -29,7 +29,6 @@ import {
   getOperationalRequirement,
 } from "@common/constants/localizedConstants";
 import { errorMessages } from "@common/messages";
-import { phoneNumberRegex } from "@common/constants/regularExpressions";
```
petertgiles, line 32: I wonder if we should remove this regex from the common file if it's not going to be used anywhere?

api/app/GraphQL/Scalars/PhoneNumber.php
petertgiles: I wonder if its even worth having this scalar anymore if it's not really doing anything?

petertgiles (commented):
A few questions that I'm not sure about...

api/app/GraphQL/Scalars/PhoneNumber.php
tristan-orourke: I like the idea of keeping the scalar - even if its not validating anything, it makes it clear how to interpret the field.

petertgiles:
Could you double-check the Chromatic snapshots?  It looks like the phone numbers are switching from optional to required which I think is not desired for this PR.

vd1992:
> Could you double-check the Chromatic snapshots? It looks like the phone numbers are switching from optional to required which I think is not desired for this PR.

aye aye, and done!

petertgiles (approved):
Nice one!

PR #3117, opened 2022-06-17
add whitespace trimming

### LINKED ISSUE #2727: Trim whitespace in inputs and textareas
Allow it to be turned off.

Re-ticketed from #2499 

## Acceptance Criteria
- [x] Our input and textarea components should trim whitespace on their values on onBlur
- [x] Add an optional prop to these components which turns off this functionality

vd1992:
resolves #2727 

Testing: Play around with forms (inputs + textareas fields) and ensure they behave and submit as expected

### DIFF
--- a/frontend/common/src/components/form/Input/Input.tsx
+++ b/frontend/common/src/components/form/Input/Input.tsx
@@ -47,11 +47,10 @@ const Input: React.FunctionComponent<InputProps> = ({
   // To grab errors in nested objects we need to use lodash's get helper.
   const error = get(errors, name)?.message;
 
-  const whitespaceTrimmer = () => {
+  const whitespaceTrimmer = (e: React.FocusEvent<HTMLInputElement>) => {
     if (whitespaceTrim) {
-      const ele = document.getElementById(`${id}`) as HTMLInputElement;
-      ele.value = ele.value.trim();
-      setValue(name, ele.value);
+      const value = e.target.value.trim();
+      setValue(name, value);
     }
   };
 
--- a/frontend/common/src/components/form/Input/Input.tsx
+++ b/frontend/common/src/components/form/Input/Input.tsx
@@ -23,6 +23,8 @@ export interface InputProps
   /** If input is not required, hide the 'Optional' label */
   hideOptional?: boolean;
   errorPosition?: "top" | "bottom";
+  // Whether to trim leading/ending whitespace upon blurring of an input, default on
+  whitespaceTrim?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


frontend/common/src/components/form/Input/Input.tsx
```diff
@@ -34,15 +36,25 @@ const Input: React.FunctionComponent<InputProps> = ({
...
 }) => {
   const {
     register,
     formState: { errors },
+    setValue,
   } = useFormContext();
   // To grab errors in nested objects we need to use lodash's get helper.
   const error = get(errors, name)?.message;
 
+  const whitespaceTrimmer = () => {
```
esizer, line 50: Reach should pass as `React.FocusEvent<HTMLInputElement>` to `onBlur` event handler. You should be able to access it like such so you do not need the `document.getElementById`:

```tsx
const handleBlur = (e : React.FocusEvent<HTMLInputElement>) => {
  const el = e.target;
  ... 
}
```

frontend/common/src/components/form/TextArea/TextArea.tsx
```diff
@@ -24,15 +26,25 @@ const TextArea: React.FunctionComponent<TextAreaProps> = ({
...
 }) => {
   const {
     register,
     formState: { errors },
+    setValue,
   } = useFormContext();
   // To grab errors in nested objects we need to use lodash's get helper.
   const error = get(errors, name)?.message;
 
+  const whitespaceTrimmer = () => {
```
esizer, line 40: Since we reuse this, it could be useful to extract this into a utility/helper function.

esizer (changes requested):

frontend/common/src/components/form/TextArea/TextArea.tsx
vd1992: in order to use `trim()` I have to constrain things to `type: string` , so I am not experienced enough to know how to write this out so genericly
trim is a string method for instance so I don't think I can write things out to return an arbitrary `type V`
😖

esizer: This was just a suggestion, if we use the `React.FocusEvent<HTMLInputElement>` mentioned in the other comment, it think this function could be simple enough to not warrant extracting it.

vd1992: got it, thanks for the advice! 💯

esizer (approved):
Awesome! 😁

Issue #3106, opened 2022-06-16
Add new fields to PoolCandidateFilter


tristan-orourke:
## Description

We want to add new fields to PoolCandidateFilter so that it can store requests made on the new Talent Search page
New fields:
- skills
- wouldAcceptTemporary

At the same time, we want to ensure the countPoolCandidates and countApplicants query accept the same object, so that we can migrate everything to using countApplicants without losing backwards compatibility for our requests. To that end, we should remove the ApplicantFilterInput type and use PoolCandidateFIlterInput in both queries. 

In addition to adding the skills and wouldAcceptTemporary fields to PoolCandidateFIlter (and PoolCandidateFilterInput), we need to port over some other fields added to ApplicantFilterInput:

There may be some frontend refactorings to accommodate some name changes from ApplicantFilterInput to PoolCandidateFilterInput:
- poolCandidates -> pools
- expectedClassifications -> classifications
- locationPreferences -> workRegions
- poolCandidates -> pools

There will be some work required to make the `pools` field accomodate the expiryStatus and Status fields. I think this can be done by:
- refactoring UserPoolFilterInput to use a single pool id instead of an array of ids 
- replacing fields of UserPoolFilterInput with and array of UserPoolFilterInput
- replacing instances of ApplicantPoolFilterInput with UserPoolFilterInput
- adding the expiryStatus and statuses fields to the `pool_pool_candidate_filter` pivot table
- We can accomodate @petertgiles desire for as-unfiltered-as-possible defaults, without hurting backwards compatibility, if we use our current schema defaults as database defaults when adding columns in a migration

We can also add a jobLookingStatus to the PoolCandidateFilter, and we can remove the schema defaults if we use good defaults in the migration.

## Acceptance Criteria
- [ ] Add fields to PoolCandidateFilter model. Skills will be a relationship. UserPoolFilterInput data comes from the pivot table.
- [ ] Add new fields in migration. Skills requires a pivot table.
- [ ] Ensure default values are set in migration in order to retain backwards compatibility. ie after migrating candidate data to users (#3148) passing old PoolCandidateFilterInputs to countPoolCandidates and countApplicants should return the same number!
- [ ] Add fields to schema. (types, queries, and migrations)
- [ ] Refactor UserPoolFilterInput to work with `pool_pool_candidate_filter` pivot table
- [ ] Replace uses ApplicantPoolFilterInput with PoolCandidateFilterInput
[...]

PR #3153, opened 2022-06-22
employment duration filter

### LINKED ISSUE #2768: Employment duration -  New filter in request page.
[Wireframe](https://www.figma.com/proto/XS4Ag6GWcgdq2dBlLzBkay/Admin-v3?page-id=0%3A1&node-id=949%3A5635&viewport=-1637%2C-2263%2C0.73&scaling=min-zoom&starting-point-node-id=949%3A5635&show-proto-sidebar=1)

Add new filter to the request talent page called "Employment duration". This should be made to the NEW SearchForm.

Any Duration is equivalent to leaving this filter empty.

image

## Acceptance Criteria

- [ ] These inputs are connected to the api query
- [ ] UI is done

vd1992:
resolves #2768 

to connect the filter to API, there needs to well, be something to connect it to so I'm playing a bit stretchy with the issue scope
I don't think it is that much extra work to add the backend tweaks for it to connect, needed so the frontend filter is actually test-able

I think things are in order? Not sure if anything else is missing
Test: play around with /search and ensure the database actions match expectations 

labels are a bit funny, two options are supposed to do nothing but giving them the same value throws alerts


vd1992:
reference for added filter radiogroup

image

vd1992:
playing around with SQL queries, and learning how to make them!
so far, what appears on /search with duration = short term yields 1
this query in adminer also yields 1 so... I think things look to be working?

```
SELECT COUNT(*)
FROM "users"
JOIN "pool_candidates" ON ("pool_candidates"."user_id" = "users"."id")
WHERE "pool_candidates"."user_id" = "users"."id"
AND "pool_candidates"."pool_id" = '1dcfd143-083c-4a9a-a8da-88f7876ebb99'
AND "pool_candidates"."pool_candidate_status" = 'AVAILABLE'
AND "pool_candidates"."expiry_date" >= '2022-06-22'
AND "users"."would_accept_temporary" = '1'
```

api/database/migrations/2022_06_22_170914_update_pool_candidatate_filter_employment_duration.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('pool_candidate_filters', function (Blueprint $table) {
+            $table->boolean('would_accept_temporary')->nullable();
```
JamesHuf, line 17: While the query itself works fine as a boolean we should probably save this as an enum/string so we can actually tell whether a user selected 'any duration' or 'permanent'.

JamesHuf (changes requested):
Another great PR! There's just one change I think we'll want.

api/database/migrations/2022_06_22_170914_update_pool_candidatate_filter_employment_duration.php
vd1992: I realized I didn't even need to add that into this PR, I didn't need any API/DB changes.
That part is stealing from #3106 without real reason so just going to undo the migration bit 😬  🤣

JamesHuf: Sounds good!

JamesHuf (approved):
Everything looks great now!

PR #3174, opened 2022-06-24
Feat - Allow inline pending

### LINKED ISSUE #2953: Accessible loading spinner component
Re-ticketed from #2862 

- consider leveraging a third-party component to help get best-practices for free

Examples:
- https://carbondesignsystem.com/components/loading/accessibility/
- https://carbondesignsystem.com/patterns/loading-pattern/
- can also add accessibility to a non-accessible third-party component

cc: @esizer (i think you shared another link in zoom chat?)

### LINKED ISSUE #3088: Loading spinner is drawn on top of elements it blends with

Minimally urgent, as affects an uncommon window size.

## Description

When the window height is low, but width is high, the spinner can be drawn on top of the header elements, and it blends:

Screen Shot 2022-06-16 at 1 32 04 PM

Screencap: https://recordit.co/mdzN1RcqwQ

## Options

1. constrain loader within section being loaded, so background will be more knowable, or
2. add a background element behind loader, either transparent or opaque so it will always be visible

I personally think number two is best, since eric went to effort to make it work full-page.

## Acceptance Criteria
- [ ] spinner is visible over whatever element might show on page

esizer:
Resolves #3088

May also resolve #2953 ?

## Summary

Fixs and issue with the loading spinner background not rendering properly as well as adding the following props:

```tsx
export interface LoadingProps {
  inline?: boolean;
  live?: "polite" | "assertive";
}
```

- Inline: Allows the spinner to be displayed inline and not take over the entire page
- Live: Gives the option to announce "Loading" to assistive technologies
  - My recommendation would be to avoid this unless you are absolutely certain it should be used
  - Always prefer polite over assertive unless necessary :)

## Screenshot
Screen Shot 2022-06-24 at 2 55 55 PM


esizer:
@patcon I connected this to #2953 and think it is sufficient to close that issue. Do you mind confirming, if it does not then we can disconnect it 😄

petertgiles:
The code looks nice... but I have no idea how to test this.  :laughing: Sorry.  I'm not able to reproduce the "spinner is hard to see condition" either.

esizer:
> The code looks nice... but I have no idea how to test this. 😆 Sorry. I'm not able to reproduce the "spinner is hard to see condition" either.

The way I was testing was:

1. Pick a page to test on
2. Go to the pages component
3. Set swap out `fetching={fetching}` to just `fetching` to force spinner to always appear
4. Navigate to page and ensure it stands out (even on purple backgrounds)

frontend/common/src/components/Pending/Loading.tsx
```diff
@@ -1,17 +1,39 @@
...
+   *
+   * polite: wait for idle state (recommended)
+   * assertive: interrupt the current user workflow
+   *
+   * REF: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions
+   */
+  live?: "polite" | "assertive";
+}
+
+const Loading: React.FC<LoadingProps> = ({ inline, live }) => {
```
patcon, line 21: Is this the way we wanted to start doing this? (Happens to be no need to children here as well :) )

```suggestion
const Loading = ({ inline, live }: LoadingProps): JSX.Element => {
```

esizer: Did we agree on adopting this pattern? I know we ran a poll but I wasn't aware of the outcome so I'm just following the established pattern. 

If we did, I can start updating all the components I touch going forward.

patcon (approved):
LGTM

patcon:
Take comments as you see fit!

patcon:
actually, just going to scope out stories and see if they capture these states. one sec

patcon:
K, just scoped out storybook, and looks good! Ideally it might have something going on to show how the spinner works on a page, or with content behind it, but nbd :)

Small musings, for your future consideration:
- consolidating with Spinner component, now that code used for second time?
- the component names having some connection or shared root word: Spinner, Loading, Pending, ErrorMessage all seem very diff and don't give hints at how they might relate

Issue #2641, opened 2022-05-03
Fancy filter in modal


tristan-orourke:
## Description

Build the modal component to submit a User filter object to a callback when it is closed. Don't worry about using the filter in a query yet.

## Screenshot: Prototype
[![image][image]][figma]
<sup>[Click screenshot to visit Figma prototype.][figma]</sup>

   [image]: https://user-images.githubusercontent.com/305339/177175957-9f0908bd-6f8f-432f-a987-2448e255b05f.png
   [figma]: https://www.figma.com/proto/XS4Ag6GWcgdq2dBlLzBkay/Admin-v3?node-id=465%3A8698&scaling=min-zoom&page-id=0%3A1&starting-point-node-id=410%3A1984&show-proto-sidebar=1&hide-ui=1

### Notes
- _Multiselect search field_ is a new component
- _Checkbox field_ is a new component (if styled like a boolean switch)

### Acceptance Criteria
- [ ] We add a prop to our multiselect component to allow for the search  
- [x] The header of the modal includes a title and subtitle
- [x] The body of the modal includes a Pools multiselect search field
- [x] The body of the modal includes a Language select field
- [x] The body of the modal includes a Classifications multiselect field
- [x] The body of the modal includes a Conditions of employment multiselect field
- [x] The body of the modal includes a Work location multiselect field
- [x] The body of the modal includes an Education multiselect field
  - we realized the data model is still in flux, so this is behind logic to disable. story shows both.
- [x] The body of the modal includes a Duration preference select field
- [x] The body of the modal includes an Availability select field
- [x] The body of the modal includes a Complete profile checkbox field
- [x] The body of the modal includes a Skill filter multiselect search field
- [x] The body of the modal includes a Government employee checkbox field
- [x] All input values are reset to original form state once the user clicks "Clear filters"
  - "original form state" currently presumed "empty", but could easily be "current" (though maybe "Reset filters" makes more sense for that)
- [x] The modal closes once the user clicks "Show Results"
- [x] The form options are applied to search filters of related form
  - i wasn't sure how to final mutation and submission logic might work, so I assumed this was a separate integration task after this is merged. the onSubmit prop passed to the component is in turn passed an arg with all FormValues
- [ ] Jest tests to confirm the filter object looks right.

tristan-orourke:
[...]

PR #3188, opened 2022-06-27
Some house-cleaning on the MultiSelect and story

patcon:
Extracted from #2641 as things to make that PR more straightforward

- component
  - a7a2a18a5 stopped using React.FC (no longer recommended)
  - 539813e64 removed HTMLSelectElement props from type (not used, nor needed for react-select component that it wraps)
  - 042e46c6e used `{...{ foo, bar }}` spread to send unmodified (and so less important) props into component
  - c79e16d1e use same logic for `required` and `aria-required` props, which both accept bool
  - de9aa01d6 removed use of controller in `<Controller control={control} />`, which is optional when using FormProvider (which we use)
- storybook
  - 71cd4c6a1 less verbose boilerplate
  - 3f569fe42 migrated to the new storybook types Meta/Story => ComponentMeta/ComponentStory (prop types auto-detected)
  - 16740b382 made story titles less verbose (stopped mentioning component is each)

### DIFF
--- a/frontend/common/src/components/form/MultiSelect/MultiSelect.stories.tsx
+++ b/frontend/common/src/components/form/MultiSelect/MultiSelect.stories.tsx
@@ -1,19 +1,17 @@
-import { Meta, Story } from "@storybook/react";
+import { ComponentMeta, ComponentStory } from "@storybook/react";
 import { action } from "@storybook/addon-actions";
 import { uniqueId } from "lodash";
 import React from "react";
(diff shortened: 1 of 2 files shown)


tristan-orourke (commented):
I know this is still marked as a draft, but it all looks good to me.

patcon:
k thanks @tristan-orourke! Lemme know if it's good to merge

tristan-orourke (approved):

PR #3176, opened 2022-06-24
Add translations manually requested by DCMO for talent search page

tristan-orourke:
Resolves #3171

Adds the translations that DCMO requested. Some of these are fields we weren't translating at all yet (mostly the legend fields of some inputs). Mostly they're adjustments to existing French translations.

To Test:
1. Run `npm run intl-compile --workspaces --if-present` from /frontend folder
2. Run `docker-compose run --rm maintenance bash refresh_common.sh` and `docker-compose run --rm maintenance bash refresh_talentsearch.sh`
3. Navigate to http://localhost:8000/fr/talent/search
4. Check that the text on the page matches the following doc: https://docs.google.com/document/d/1dr6fKxscBDpaXVG1WLOvg-aiQvqULwqnxAMe93VAIgg/edit?usp=sharing

PR #3193, opened 2022-06-27
3078 Complete a round of translations

### LINKED ISSUE #3078: Extract all strings from current app and send for translation

tristan-orourke:
Resolves #3078 

Extract strings from all frontend projects, sent them for translation, and updated fr.json files.

I also modified checkIntl.js script to allow data to be read from yaml files as well as json files. I did this mainly to allow comments to be included in the whitelist files, so we can tell which translation entries are being whitelisted as the same in English and French.


vd1992:
not sure how to judge this but
checked out, intl-compile, run dev all worked
some stuff I know won't be translated yet as it was very recently merged in

/profile looks good 
/search looks good 

shouldn't "Close Menu" be translated at this point?
image

vd1992 (approved):
see comment above, approved in case everything is a non-issue

tristan-orourke:
> except for this little oddity

This is fixed in the other lang PR,  #3176

PR #3199, opened 2022-06-28
Ensure field height is consistent both with/without context info

patcon:
The non-zero line-heigh of the icon was causing the field layout to jitter and be slightly different depending on whether context info was set. The jitter wouldn't likely be noticed in live site conditions, but the slight difference in field height when two fields are alongside might be perceptible.

## Before
[](https://recordit.co/uxjMICz0hK)

## After

[](https://recordit.co/vJGd3O92rb)

### DIFF
--- a/frontend/common/src/css/common.css
+++ b/frontend/common/src/css/common.css
@@ -16,6 +16,7 @@ textarea {
   border: none;
   padding: 0;
   vertical-align: middle;
+  line-height: 0;
   text-decoration: none;
   background: transparent;
   cursor: pointer;


patcon:
@JoshBeveridge It does change the hotspot height unfortunately:



https://61e099a1bb1465003a73d3ce-vknswqzaqg.chromatic.com/?path=/story/form-select--select-required-with-info&args=context:a&globals=outline:true

`line-height: normal` variant: 20.5px (original)
`line-height: 0` variant: 14.4px
SVG: 14.4px (0.8rem)
visible icon: <11.5px

With this new non-jitter form, the hotspot now technically has a height equal to label text, but visually it appears smaller due to icon within SVG having a margin. This is perhaps not ideal, as often icons go right to edges of SVG, and any visual margin comes from CSS.

Options:
1. Keep current change, and merge as-is. Hotspot will change from height 20.5px => 14.4px
2. Keep current change, but Increase icon size to bring to edge of SVG bounding box. Hotspot still smaller, but less illusion of it being smaller than label.
3. Abort current change, but hardcode label div to have height or line-height of 20.5px, so will never jitter. A bit messy since it's a diff component, but works.
4. Abort current change, and leave unchanged. Required and optional fields on same line will have slightly diff alignment.

JoshBeveridge:
@patcon I think the existing target is more of an issue than this minor reduction, so for now, forge ahead with the change including the line height, and we'll revisit that button as a part of the eventual component review that will happen with the updated look and feel.

petertgiles (approved):
:100:

patcon:
Missed this, but thanks @JoshBeveridge! Merging as-is :)

PR #3200, opened 2022-06-28
Bug - 2868 - Fix unauthorized component flashing

### LINKED ISSUE #2868: Admin - Flash of "Not Authorized" on reload

esizer:
Resolves #2868 

## Summary

This fixes an issue where the unauthorized message was flashing on the screen while the query was running. Now, the spinner is displayed while the necessary authorization query is ran.

## Testing

-  Login to `/admin`
- Refresh a page
- Confirm the spinner appears instead of "Not Authorized" while the `useGetMeQuery` is ran.

### DIFF
--- a/frontend/common/src/components/context/AuthorizationProvider.tsx
+++ b/frontend/common/src/components/context/AuthorizationProvider.tsx
@@ -2,18 +2,20 @@ import React from "react";
 
 import { AuthorizationContainer } from "../Auth";
 import { useGetMeQuery } from "../../api/generated";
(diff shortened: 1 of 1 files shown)


JamesHuf (approved):
Works for me 👍

PR #3215, opened 2022-06-29
Feature - 2659 - New Tab Styles

### LINKED ISSUE #2659: Re-style tab component
Screen Shot 2022-05-04 at 11 39 37 AM

esizer:
Resolves #2659 

## Summary

This replaces the existing tabs with new ones that have the updated styles and use [@reach/tabs](https://reach.tech/tabs) as the underlying component.

## Testing

- View Storybook
- Check:
  - `/admin/user/{id}`
  - `/talent/profile`
  - `/talent/profile/skills-and-experiences/personal/create`

## Screenshot

Screen Shot 2022-06-29 at 12 48 01 PM


vd1992 (approved):
- nothing of concern in the Chromatic Diff
- builds and runs fine
- navigated to the addresses mentioned, nothing appears off

I think things are good. Though, I shall defer to Jerbo for final say on the appearance
@Jerryescandon

petertgiles:
Oops, I was a little too slow.  :smile:   @Jerryescandon the new tabs are missing the closer.  Is this a problem?
image

Jerryescandon:
Their only purpose is to make the UI look more tidy so the form doesn't look so intimidating on first contact. But I thing we can scrap it.

Issue #3160, opened 2022-06-23
Fix tc-report Register link


patcon:
Steps to confirm:

1. Visit a tc-report page like: https://talent.canada.ca/en/privacy-notice
2. Click "Login", and should work
3. Click "Register, and will not work.

## Acceptance Criteria
- [ ] register link on tc-report takes us to gckey
- [ ] if possible: some equivalent local test should reproduce break/fix

PR #3252, opened 2022-07-04
Bug/3160 broken tc report register link

JamesHuf:
Addresses #3160.

This should fix the register link on the tc-report pages by adding an actual /register route for them to link to.

I'm not certain this is the best way of solving this and unfortunately we're using absolute links on the tc-report pages so this isn't actually testable - thoughts on me changing them to be relative?

- Adds a new /register route identical to /login
- Changes the route to the register page from /register to /register-info similar to /login-info


petertgiles:
Yeah, you're right, it's not actually possible to fully test this locally since the Jekyll-built site isn't dynamic enough to change links based on environment.  That's why we just put the prod links in source.  We can sort-of test this since we know how it will work locally.  We know `talent.canada.ca/register?locale=en` will correspond to `localhost:8000/register?locale=en` and `talent.canada.ca/login?locale=en` will correspond to `localhost:8000/login?locale=en`.

In this PR you've made a route `http://localhost:8000/register-info` but that doesn't actually work since it doesn't exist in the router.  The `/login` route that the tc-report links to right now is actually a route in the api project, not the frontend.  If you want to make the register link work like the login link you'll need to make the changes in the api project.  Alternatively you could point both the login and register links to the info pages in the frontend.  That might be nicer but we definitely want both links to work the same way, whichever way that is.  :smile:

JamesHuf:
@petertgiles Ok, I think I got it working properly now.

petertgiles (approved):
Works for me.  :+1:

PR #3287, opened 2022-07-07
countApplicants unit tests

### LINKED ISSUE #2803: Write PHPUnit tests for countApplicants query
The many filters of the query make it complicated, and we've repeatedly found bugs with it. We need to write a suite of tests to have high confidence in this.

Issues we've found
- selecting all classifications can return a higher number than is actually in the pool you're searching on
- salary range & classification not interacting correctly
- filtering by English or French not including bilingual candidates
- ensure setting isWoman (and other EE fields) to false doesn't exclude women.

vd1992:
https://github.com/GCTC-NTGC/gc-digital-talent/issues/2803#issuecomment-1175200532
> These might be already written, but worth double checking.

Closes #2803 
There were countApplicant units tests already, I added a bit more
To review: proofread the added tests, judge what else may be needed

I think it is worth noting that a bunch of the filters for countApplicants are tested individually so I don't think there is much of a gap in test coverage? Just need to watch out for edge cases so suggestions welcome 

### Scope Adjustment
Per the advice of @petertgiles , countApplicant testing to be extracted into its own thing to be tested as a standalone with many tests


vd1992:
> The new tests look good to me.

Thanks!
I'll leave it up for a bit still

vd1992:
per Peter's suggestion, I have split out countApplicants into its own thing and made LOADS of tests
I think I got things in a good place? I used your code as example for chunks of it @JamesHuf , like a manual from a videogame case (back in the day)

JamesHuf (approved):
These changes make sense to me, the tests look good, and everything is passing. Great job!

PR #3291, opened 2022-07-07
3256 Store Applicant Filters

### LINKED ISSUE #3256: Allow ApplicantFilter to be attached to CandidateSearchRequest
PoolCandidateSearchRequests currently include a PoolCandidateFilter, which defines which candidates match the request. As we prepare to start using ApplicantFilters instead of PoolCandidateFilters, add an applicantFilter field to the SearchRequest type, so that either type of filter can be used.

## Acceptance Criteria

- [x] Create an ApplicantFilter type, model and table, based on ApplicantFilterInput
- [x] Create factory and seeder for ApplicantFilter
- [x] Create applicantFilter and applicantFilters queries
- [x] Add a one-to-one relation between ApplicantFIlter and PoolCandidateSearchRequest
- [x] When creating a request, allow applicantFilter to be set
- [x] make poolCandidateFilter field in PoolCandidateSearchRequest nullable
- [ ] Change name of PoolCandidateSearchRequest to CandidateSearchRequest to be more general and less confusing
- [x] Make minimal changes necessary to frontend to comply with the above changes without errors

tristan-orourke:
Resolves #3256 

- Creates an ApplicantFilter type, with its own eloquent model, queries, tables, factories and tests.
- Added methods to some existing model factories, to create them with extra or specific attributes, for more convenient testing.
- Allows ApplicantFilters to be attached to PoolCandidateSearchRequests instead of PoolCandidateFilters. Consequently, a PoolCandidateFilter is no longer required when creating a PoolCandidateSearchRequest.
- Made some minor changes to ApplicantFilterInput, to make it easier to store in tables and help clarify some of the fields.
- Made a graphql directive, `@pluck`, which acts like the [Laravel collection method](https://laravel.com/docs/9.x/collections#method-pluck). 
- Pluck directive is used to allow the ApplicantFilterInput to be the same shape as ApplicantFilter when returned by queries, without changing the underlying implementation of scope/filterBy methods. This means that you can retrieve an ApplicantFilter object from a query, and then pass it into an applicantCount query without massaging that data object. This is tested with phpunit.
- Refactored UserFilterInput to include an ApplicantFilterInput field, instead of duplicating fields and functionality.
- Tried to thoroughly test all changes to schema with PHPUnit. 

TODO:
- [x] Update client-side code to match breaking changes to schema

Remaining questions:
- [x] Should ApplicantFilters even be able to specify JobLookingStatus, PoolCandidate expiry status, or PoolCandidate status? All search requests should implicitly exclude NOT_ACTIVE, expired, and unavailable candidates, no?
    - Removed those filters from ApplicantFilter, and instead use scopes to force those filters on the applicantCount query. This allows us to expose only the desired functionality to applicant search page.
- [ ] Can we validate PoolCandidateSearchRequests to require one or the filter?
    - Spun off into issue #3317


vd1992:
warning at /admin/dashboard
image

api/database/factories/ApplicantFilterFactory.php
```diff
@@ -0,0 +1,102 @@
...
+            // $pools->each(function (Pool $pool) use ($filter, $sparse) {
+            //     $recordFactory = ApplicantFilterPoolRecord::factory();
+            //     if ($sparse) {
+            //         $recordFactory = $recordFactory->sparse();
+            //     }
+            //     $filter->applicantFilterPoolRecords()->save($recordFactory->make([
+            //         'pool_id' => $pool->id,
+            //         'applicant_filter_id' => $filter->id,
+            //     ]));
+            // });
```
vd1992, line 99: leftover troubleshooting?

api/database/helpers/ApiEnums.php
```diff
@@ -64,6 +64,19 @@ public static function govEmployeeTypes() : array
     const CANDIDATE_EXPIRY_FILTER_ACTIVE = 'ACTIVE';
     const CANDIDATE_EXPIRY_FILTER_EXPIRED = 'EXPIRED';
     const CANDIDATE_EXPIRY_FILTER_ALL = 'ALL';
+    /**
+     * A collection of enums for CandiateExpiryStatus in factories and seeders
```
vd1992, line 68: CandidateExpiryStatus*

api/tests/Feature/DirectivesTest.php
```diff
@@ -0,0 +1,132 @@
...
+        parent::setUp();
+
+        $this->setUpTestSchema();
+    }
+
+    public function testPluckFromArray(): void
+    {
+        // Mock up a resolver that returns the args we're interested in
+        $this->mockResolver(function ($root, array $args) {
+            var_dump($args);
```
vd1992, line 27: think that is no longer needed

vd1992:
I don't quite follow what is going on in ApplicantFilterPoolRecord, there is:
One pool ID
One candidate expiry status (active, expired, all)
An Array of pool candidate statuses (available, placed, etc)

Is this just collecting for recording purposes a single pool, whether you've selected active/expired/all and a break down of the candidates contained in terms of current status?
As in: selected pool number 123, picked ACTIVE expiry_status, see how many candidates contained pop up as expired candidate_status?

tristan-orourke:
> I don't quite follow what is going on in ApplicantFilterPoolRecord

Yeah, that makes sense. ApplicantFilterPoolRecord is left over from an earlier version of this PR... I stopped using it anywhere, but forgot to remove the model. I've deleted it and all references to it now.

For a full explanation: 
The usersPaginated query allows you to filter users in many ways (with UserFilterInput), including which pools they're in, and even via specific statuses in those pools (with UserPoolFilterInput). I originally allowed you to include those same status filters as part of an ApplicantFilter when storing it with a SearchRequest, which required the ApplicantFilterPoolRecord (and accompanying table).
But the ApplicantFilter is only used as part of a SearchRequest, and the SearchRequests aren't really intended to allow for searching for expired or unavailable candidates. So after this slack conversation [Gray agreed](https://talent-cloud.slack.com/archives/CB424V61J/p1657293078612919?thread_ts=1657231211.425819&cid=CB424V61J) we should only expose the minimum functionality we intend to, I removed those filters and instead forced the intended filters onto the countApplicants query with the `availableForOpportunities` scope and `filterByAvailableInPools` builder method.

esizer:
Is this meant to start storing the filters or just laying the ground work? All tests are passing and working well but the filters are not being stored yet from what I can see.

tristan-orourke:
@esizer This PR is only meant to cover API functionality. It makes it possible to make search requests with Applicant Filters, but does not make any changes to the client-side to actually start doing that.

esizer (approved):
Looks good 👍

PR #3249, opened 2022-07-04
Fancy filter

patcon:
Resolves #2641

Prototype: https://www.figma.com/proto/XS4Ag6GWcgdq2dBlLzBkay/Admin-v3-(Profile-release)?node-id=465%3A8698&scaling=min-zoom&page-id=0%3A1&starting-point-node-id=410%3A1984&show-proto-sidebar=1&hide-ui=1

## To Do
- [x] wire up clear button
- [x] ensure all strings localized
- [x] figure out localized select options
- [x] ~~import alternative grouped multiselectfield (was building demo outside repo)~~
- [x] wire up to search filter
- [x] jest tests
  - waiting for review first?
- [x] fix test that's timing out
- vachan feedback
  - [x] a6d3d43 a496b66 fix CSS issue 
  - [x] embed in User.stories.jsx "User Table" stories
  - [x] 0ba68ca fixed storybook brokenness
- peter feedback
  - [x] 3ccc00d7b create constants for `eductionRequirements`
  - [x] 1d79ba29a use `notEmpty` instead of new `notNullOrUndefined`
  - [x] 4a4d34cd8 improve docs of latency simulation in `MockGraphqlDecorator`
  - [x] e21acf01b avoid using `<FormattedMessage />` component
  - [x] add jest test(s) for `SelectFieldV2` component
  - [x] add jest test(s) for `MultiSelectFieldV2` component
  - [x] add jest test(s) for `UserTableFilterDialog` component (test filter object)
  - [x] add jest test(s) for `useFilterOptions` hook
- [x] split SelectFieldV2 and MultiSelectFieldV2 components/tests/stories into own PR https://github.com/GCTC-NTGC/gc-digital-talent/pull/3510
- tristan feedback
  - [x] 8976c920a move useLocale to common dir https://github.com/GCTC-NTGC/gc-digital-talent/pull/3249/files#r924925000

## Manual QA

1. Visit "SearchFilter > Default" component in storybook
2. Click the **dialog's "x"** in the corner and re-open.
3. With empty filters...
    - **Submit** and note **empty arrays** for each field's data in Action output.
5. **Add** some filter options to Select and MultiSelect fields. Clear each with **field's "x"** button.
    - **Submit** and note **empty arrays** for each field's data in Action output. Re-open.
6. **Add** some filter options to Select and MultiSelect fields, then close vis dialog's dismiss "x". Re-open and note **empty filters** (**no persistence**).
    - **Submit** and note **empty arrays** for each field's data in Action output. Re-open.
[...]

PR #3293, opened 2022-07-08
add new enums for overtime

### LINKED ISSUE #3286: Create new enum constants for new overtime options
The new version of the search page moves from "overtime - short notice" and "overtime - scheduled" to "overtime - occasionally" and "overtime - regularly".

We are currently treating the transition as a simple wording change:

However, they are quite different and should probably get new enum constants values to reflect that instead of reusing the older ones.

vd1992:
closes #3286 

some concerns, so for this I have just added the new enums and reworded a few things, now there are four overtime Enums
removing the prior enums will need to be done at some point and associated work (database migration?)

given the operationalRequirements are grabbed via get function for most of them, there is no manual selection of operational requirements, so by adding the 2 enums to localizedConstants, multiple components using a get function are pulling in more automatically 
for example
https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/frontend/admin/src/js/components/pool/CreatePool.tsx#L303
all the OperationalRequirement enums are passed in there
as well as based off the type in schema, which as mentioned above has four overtime enums now


vd1992:
I'm not sure, the old enums are needed for the old talent search for sure but I do not know about /admin end of things

patcon:
hold on, i'll poke in and see if what i was thinking makes sense :)

patcon:
My thinking on this (and my work in #3249) led to this discussion proposal, if you feel like weighing in:
https://github.com/GCTC-NTGC/gc-digital-talent/issues/3314

If we feel it's a problem that all 4 overtime options are showing everywhere, we could try to implement a simple `useOptions` for this, that just handles "operationalRequirements" for now

so this would be the proposed usage (which would return a filtered list of options with the two new additions, but without the other overtime values:

```js
const Foo = () => {
  const { operationalRequirements } = useOptions("operationalRequirements", { filterVariant: "v2" });
  return (
    <Checklist items={operationalRequirements.options} />
  )
}
```

api/database/helpers/ApiEnums.php
```diff
@@ -20,11 +20,12 @@ public static function operationalRequirements() : array
             'DRIVERS_LICENSE',
             'WORK_WEEKENDS',
             'OVERTIME_SCHEDULED',
-            'OVERTIME_SHORT_NOTICE'
+            'OVERTIME_SHORT_NOTICE',
+            'OVERTIME_OCCASIONAL',
+            'OVERTIME_REGULAR',
```
patcon, line 25: Larger proposal aside, it feels worth asking whether it's alright if all the types of overtime show up in forms. If so, then we could just merge this as-is, and leave the more clever solution for later :)

vd1992: added more subselection stuff
given the enum group may remain in flux for a while or tied to a feature_flag, this might be a compromise for now

vd1992: test by navigating the app I suppose? specifically where Operational Requirements are in play

frontend/admin/src/js/components/pool/CreatePool.tsx
```diff
@@ -129,6 +129,16 @@ export const CreatePoolForm: React.FunctionComponent<CreatePoolFormProps> = ({
     }),
   );
 
+  const operationalRequirementsSubset = [
```
patcon, line 132: Should we just say `operationalRequirementsV2` to hint that this is the new hotness? I'd def advocate we air toward using V1, V2, V3 suffixes as we evolve things in the app, rather than arbitrary modifier words

patcon:
```suggestion
  const operationalRequirementsV2 = [
```

patcon (approved):
Comments aside, looks good to me!

frontend/talentsearch/src/js/components/search/deprecated/SearchForm.tsx
```diff
@@ -184,7 +184,7 @@ export const SearchForm: React.FunctionComponent<SearchFormProps> = ({
     [cmoAssets, locale, intl],
   );
 
-  const operationalRequirementsSubset = [
+  const operationalRequirementsSubsetV1 = [
     OperationalRequirement.ShiftWork,
     OperationalRequirement.WorkWeekends,
     OperationalRequirement.OvertimeScheduled,
```
patcon, line 190: Ah, I hadn't realized there was an even earlier subset -- sorry if that confused my suggestion :)

vd1992: no worries
all good to merge?

frontend/admin/src/js/components/pool/UpdatePool.tsx
```diff
@@ -152,6 +152,16 @@ export const UpdatePoolForm: React.FunctionComponent<UpdatePoolFormProps> = ({
...
 
+  const operationalRequirementsSubsetV2 = [
+    OperationalRequirement.OvertimeOccasional,
+    OperationalRequirement.OvertimeRegular,
+    OperationalRequirement.ShiftWork,
+    OperationalRequirement.OnCall,
+    OperationalRequirement.Travel,
+    OperationalRequirement.TransportEquipment,
+    OperationalRequirement.DriversLicense,
+  ];
```
patcon, line 163: Pretty minor, but still seeing OVERTIME_SCHEDULED show up in the stories, which is a bit of a pain, since the faker stuff doesn't take into account the variations :/

Screen Shot 2022-07-13 at 1 11 28 PM

https://61e099a1bb1465003a73d3ce-qlakyjhnjl.chromatic.com/?path=/story/pool-candidates--update-pool-candidate-form

vd1992: I wouldn't worry too much about faker not accounting for different versions, at least I think it isn't something to worry about

patcon: oh hey, it's hardcoded in stories, so easy to fix :)

vd1992: [got it](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3293/commits/bcd8023bdd04ea30a139d8b55c2386f37105878f)

PR #3339, opened 2022-07-13
Fix Jest for GitHub workflow

### LINKED ISSUE #3309: Jest tests are failing but Github Action check passes
When Jest tests fail, the Github check should fail!

As an example:
1. Go to PR #3288
2. Go to the Checks tab
3. Go to the Jest Tests check
4. Click Unit Tests
5. Open "Run Unit Tests: All Workspaces"
6. See that some tests failed, but the overall check passed!

Obviously we want to fix the failing jest tests eventually, but first we want to ensure that failing tests will make the Check fail!

mnigh:
Resolves #3309.

### Description
Jest's coverage reports were slowing down our GitHub workflow but also causing a timeout which led to the job not completed with exit code 1 even when there was a failing test.

### Changes
- Sets `collectCoverage: false` in jest config files
- Removes `passWithNoTests` flag from jest
- Adds basic test for admin **HomePage** component
- Adds CssStub.js for admin
- Fixes CssStub.js case name reference in jest config
- Fixes case of test name

### Note
The **Jest Tests / Unit tests** check on this PR _should_ fail as failing tests do exist on the main branch (hence this issue and PR).


esizer (approved):
Nice! What exactly caused it, the coverage or improper letter case, or both?

vd1992:
🕵️

mnigh:
> What exactly caused it, the coverage or improper letter case, or both?

Coverage, pretty sure. **CssStub.js** case was a good fix though to be consistent with the other projects.

PR #3342, opened 2022-07-14
Bug - 3336 - side menu tests failing

### LINKED ISSUE #3336: Testing - SideMenu jest tests fail
## Description
Both tests listed fail with same error.

✕ Should be closed if isOpen false (30 ms)
✕ Should be open if isOpen true (11 ms)

```
  ● SideMenu › Should be closed if isOpen false

    [React Intl] Could not find required `intl` object. <IntlProvider> needs to exist in the component ancestry.

      27 |   children,
      28 | }) => {
    > 29 |   const intl = useIntl();
         |                       ^
      30 |
      31 |   const isSmallScreen = useIsSmallScreen();
      32 |   const handleToggle = () => {
```

esizer:
Resolves #3336 

## Summary

This moves the test utilities to `@common` so it can be used in other projects. It also uses the `customRender` function to fix the `SideMenu` test.

### DIFF
--- a/frontend/common/src/components/SideMenu/SideMenu.test.tsx
+++ b/frontend/common/src/components/SideMenu/SideMenu.test.tsx
@@ -2,8 +2,9 @@
  * @jest-environment jsdom
  */
 import "@testing-library/jest-dom";
-import { render, screen } from "@testing-library/react";
 import React from "react";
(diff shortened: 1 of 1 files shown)


mnigh (approved):
Nice work!

Issue #3172, opened 2022-06-24
SkillAccordion jest test throws a console error 


mnigh:
### Description
The `SkillAccordion` jest test throws a console error stating that the `SkillAccordion` component has two children with the same `key`:

```
console.error
    Warning: Encountered two children with the same key, `89bd9d8d-69a6-474e-8f46-7cc8796ed151`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
        at div
        at div
        at div
        at div
        at Accordion (/home/runner/work/gc-digital-talent/gc-digital-talent/frontend/common/src/components/accordion/Accordion.tsx:15:8)
        at SkillAccordion (/home/runner/work/gc-digital-talent/gc-digital-talent/frontend/common/src/components/UserProfile/SkillAccordion/SkillAccordion.tsx:35:8)
        at IntlProvider (/home/runner/work/gc-digital-talent/gc-digital-talent/frontend/node_modules/react-intl/src/components/provider.js:84:47)
```

### Screenshot
![Screen Shot 2022-06-24 at 13.23.08.png](https://images.zenhubusercontent.com/5f7b33373ac7b7279de9a3c5/12d56891-d1fa-4616-8d49-b87de7024878)

PR #3349, opened 2022-07-14
fix skill accordion jest test

### LINKED ISSUE #3331: Testing - SkillAccordion jest test fails
## Description
```
SkillAccordion tests › It renders proper context and detail when an award experience is provided
TestingLibraryElementError: Unable to find an element with the title: award.
```

## Notes
[The title attribute and value was removed](https://github.com/GCTC-NTGC/gc-digital-talent/commit/48a076c1b17370f9bcdbe0ee557528622371afd3#diff-7f7b32e68a1151c8712bb7655fb7828247ce6ade79300756082dcba4dbd6e73eL113) as part of #3267 since having a `title` attribute on a `span` element is not valid HTML. However, the test now fails. SkillAccordion tests should be refactored to try to avoid using `data-testid` attribute if possible and at least the [award](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/frontend/common/src/components/UserProfile/SkillAccordion/SkillAccordion.test.tsx#L70) to not use the attribute since it is not used by the other SkillAccordion tests.

vd1992:
resolves #3331 

Given no other test involved that element, I just elected to remove it. 
As well, as bonus, since I was already in the file I resolved the warning that was happening. It was simple, the generator always returns the same key for the first element, so when using two elements, have the second thing be the second item in the generated array. 

Testing: If it isn't green below, then that's a problem

### DIFF
--- a/frontend/common/src/components/UserProfile/SkillAccordion/SkillAccordion.test.tsx
+++ b/frontend/common/src/components/UserProfile/SkillAccordion/SkillAccordion.test.tsx
@@ -150,7 +150,7 @@ describe("SkillAccordion tests", () => {
 
   test("It renders proper context and detail when more than one experiences provided", () => {
     const experience1 = experienceGenerator.workExperiences()[0];
-    const experience2 = experienceGenerator.educationExperiences()[0];
+    const experience2 = experienceGenerator.educationExperiences(2)[1];
 
     testSkill.experiences = [experience1, experience2];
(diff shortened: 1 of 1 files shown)


esizer (approved):
I'd connect this to #3172 If it fixes that as well before merging.

mnigh:
Nice work on the double whammy fix @vd1992!

PR #3352, opened 2022-07-14
Tooling - 2757 - Upgrade Node

### LINKED ISSUE #2757: Update nodejs to Active LTS version 16
Project currently uses a version of nodejs, 14, that is on Maintenance LTS and will be [end of life in a year from now](https://nodejs.org/en/about/releases/) (2024-04-30).

Ensure Azure uses the same version when building artifacts.

esizer:
Resolves #2757 

## Summary

This upgrades node and npm to LTS

- node: 16.16.0
- npm: 8.11.0

## Testing

1. Rebuild container
   - `docker-compose stop php`
   - `docker-compose rm php`
   - `docker-compose up --detach --build`
   - `docker-compose run --rm maintenance bash setup.sh`
2. Confirm application builds and runs properly


vd1992:
am running maintenance setup at the moment, seeing this

image

which is weird because I am not on 14.X anymore

image

nothing broke yet so still keeping an eye out

vd1992 (approved):
navigated through parts of the app at random, saw nothing breaking
so, nothing to raise an alert on my end

saw this in the console when updating profile in the /profile section, I think it might be a separate issue so I will make it if it doesn't seem relevant to this issue
image

vd1992:
that console warn didn't pop up again 🤷 
I will keep an eye out for it

petertgiles:
Whoah, hol' up!  We need to get the artifact build in Azure to v16, too!

esizer:
> Whoah, hol' up! We need to get the artifact build in Azure to v16, too!

Wait, it doesn't build from the `Dockerfile`? 😱

petertgiles:
I'm actually not sure.  @tristan-orourke could you confirm?

Issue #3162, opened 2022-06-23
Edit Pool page - Readonly


Jerryescandon:
<sup>ℹ️ [Figma (root)][figma] | [Sitemap][sitemap]</sup>
<sup>_This initial comment is collaborative and open to modification by all._</sup>

## Description

*This first issue is to create a readonly version of the page - data is populated from the database but all the action buttons do nothing.  Mutations implemented in #3169*

This page contains an editable form for the information in the pool model.
This page also contains buttons that open modals to change the "Status" of this pool. Which in turn changes the behavior of the Pool Advertisement page.

If the pool status is in DRAFT, you can SAVE, PUBLISH, or DELETE.
If the pool status is PUBLISHED (Active before "Closing Date") you can CLOSE or EXTEND THE DATE.
If the pool status is PUBLISHED (Expired after "Closing Date") you can EXTEND THE DATE or ARCHIVE.

NOTE 01: GENERIC CLASSIFICATION: Leave out this dropdown selector from this issue, it will be addressed at a later point.

NOTE 02: DATE PICKER (yyyy/mm/dd): Date will always be set in Pacific Time at 11:59:59 p.m. (probably handled in the backend mutation) 

NOTE 03: We are not gonna implement "ARCHIVE" yet, just leave the button in but it won't do anything yet.

NOTE 04: CLASSIFICATION PICKER (e.g. IT-01): Front-end will be a 1 selection only (dropdown), but back end will allow multiple selections (just in case we need to allow this down the line). 

## Screenshot: Prototype
![image](https://user-images.githubusercontent.com/16977254/175546064-43c2c3d9-3d57-42dd-88f3-f41fd0112329.png)

## Acceptance Criteria
- [ ] URL "https://talent.canada.ca/en/admin/pools/[ID]/Edit" (Discuss what happens to old one)
- [ ] Create whole form, query for data, no actions taken on button clicks
- [ ] Put behind feature flag "Direct Intake"
- [ ] All modals need to work 
- [ ] Page needs to behave in different way depending on status (Draft, Published (Active), Closed (Published and Expired)

[Link to Edit Pool Advertisement wireframe](https://www.figma.com/proto/b6lDM86QCJQ6iaAh7qrTYH/Admin-v4-(Direct-Intake)?page-id=0%3A1&node-id=1618%3A6064&viewport=-3442%2C1747%2C0.59&scaling=min-zoom&starting-point-node-id=1618%3A6064)

[...]

PR #3353, opened 2022-07-14
Prep for edit pool advertisement readonly page

petertgiles:
The branch lays the foundation (but does not close) #3162 

Summary of changes:
- https://github.com/GCTC-NTGC/gc-digital-talent/commit/1c6635482ec97c27e0ca0590c3ef6f227b1fa65b - Removes the deprecated `pool` fields that will not be used on `poolAdvertisement`.  Also adds the missing (currently unused) ARCHIVED status.
- https://github.com/GCTC-NTGC/gc-digital-talent/commit/7264041f27f4e2d36111cf0b930139f56745113c - Exposes the `Option` interface from within the select component for use in typing parent component objects.
- https://github.com/GCTC-NTGC/gc-digital-talent/commit/48db2ed158fc0733272328a4b2d4a3c32a873fed - Update the `RadioGroup` component to make the legend optional and use multiple columns.  For use here: 
image
- https://github.com/GCTC-NTGC/gc-digital-talent/commit/a85b3189f24095d4c9baac1b1e3e2a24367c7c8b - Moves `SearchBar` and `SkillResults` to common so they can be shared with admin for this:
image
- https://github.com/GCTC-NTGC/gc-digital-talent/commit/1603927066af93cd38d1998e5fac0a1c7ad26fb8 - Some helpers for testing
- https://github.com/GCTC-NTGC/gc-digital-talent/commit/e36e5ba83bcecd50d8b0d3efe4efcf8f2693ac5c - Fix the label for this:
image
- https://github.com/GCTC-NTGC/gc-digital-talent/commit/315726f49fc526d08ccbaf50cb8524d467a68ca1 - Add a format helper to common for this: 
image
- https://github.com/GCTC-NTGC/gc-digital-talent/commit/55025c2a52e74fed5ac72b8ba82e65782525ffea - Catch up with PR in the meantime that moved the test utils to common


frontend/common/src/components/inputPartials/Fieldset/Fieldset.tsx
```diff
@@ -7,7 +7,7 @@ import InputError from "../InputError/InputError";
 
 export interface FieldsetProps {
   /** The text for the legend element. */
-  legend: string;
+  legend?: string;
```
esizer, line 10: I think providing a way for the legend to be hidden would be more appropriate. A `<fieldset>` must have a `<legend>` as the first child.

"The first element inside the fieldset must be a legend element, which provides a label or description for the group."

REF: https://www.w3.org/WAI/WCAG21/Techniques/html/H71.html

petertgiles: Good point!  I've made it mandatory again and added a `hideLegend` tag in the same way that `hideOptional` exists.

https://github.com/GCTC-NTGC/gc-digital-talent/pull/3353/commits/c59416a43f884ef93731f1f6c63dc165e80f0201

PR #3356, opened 2022-07-14
resolve backend advertiseStatus bug

### LINKED ISSUE #3354: Bug: Advertisement status expiring too early
## Summary

It appears as though our advertisement status is expiring too early.

## Possible Problem

I believe this is related to timezones. @vd1992 noticed this (screenshot attached) and even when the item should be open for another few hours, the status is `POOL_ADVERTISEMENT_IS_EXPIRED`

## Screenshot

image

### Backend

https://github.com/GCTC-NTGC/gc-digital-talent/blob/e97afa7f1569c0e9bbcc7873f0ae89b8c4024002/api/app/Models/Pool.php#L87-L109

### Frontend 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/e97afa7f1569c0e9bbcc7873f0ae89b8c4024002/frontend/talentsearch/src/js/components/pool/PoolAdvertisementPage.tsx#L110-L112

vd1992:
resolves #3354 

having the timezone set as Pacific in the accessor attribute was messing with things when current time and expiry time were within a few hours of one another
given the database deals in UTC, I elected to go with UTC time being enforced for the backend, I think it'll be easier that way, plus it seems that sticking with Laravel/database conventions would be preferred in that other issue

as well, added an additional test to ensure the bug was (ideally) resolved
testing: see if you can break the accessor in Adminer + viewing pool poster

sidenote: I think this convention should be kept in mind for mutations creating posters too, to convert time to UTC before executeMutation

### DIFF
--- a/api/app/Models/Pool.php
+++ b/api/app/Models/Pool.php
@@ -87,7 +87,8 @@ class Pool extends Model
     /* accessor to obtain Advertisement Status, depends on two variables regarding published and expiry */
     public function getAdvertisementStatusAttribute()
     {
-        date_default_timezone_set('America/Vancouver');
+        // given database is functioning in UTC, all backend should consistently enforce the same timezone
+        date_default_timezone_set('UTC');
         $isPublished = $this->is_published;
         $expiryDate = $this->expiry_date;
         $currentTime = date("Y-m-d H:i:s");
--- a/api/app/Models/Pool.php
+++ b/api/app/Models/Pool.php
@@ -88,7 +88,6 @@ class Pool extends Model
     public function getAdvertisementStatusAttribute()
     {
         // given database is functioning in UTC, all backend should consistently enforce the same timezone
-        date_default_timezone_set('UTC');
         $isPublished = $this->is_published;
         $expiryDate = $this->expiry_date;
         $currentTime = date("Y-m-d H:i:s");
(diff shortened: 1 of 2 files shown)


api/app/Models/Pool.php
```diff
@@ -87,7 +87,8 @@ public function nonessentialSkills(): BelongsToMany
     /* accessor to obtain Advertisement Status, depends on two variables regarding published and expiry */
     public function getAdvertisementStatusAttribute()
     {
-        date_default_timezone_set('America/Vancouver');
+        // given database is functioning in UTC, all backend should consistently enforce the same timezone
+        date_default_timezone_set('UTC');
```
esizer, line 91: Do we need this line at all? I believe laravel automatically sets the timezone to UTC for us.

api/tests/Feature/PoolTest.php
```diff
@@ -104,4 +104,52 @@ public function testPoolAdvertisementAccessor(): void
     ]);
   }
 
+  public function testPoolAdvertisementAccessorTime(): void
+  {
+    // test that expiry on day of functions as expected, that soon to expire can be applied to and just expired is longer open for application
+    date_default_timezone_set('UTC');
```
esizer, line 110: See previous note.

esizer (changes requested):
Awesome! I think we can safely remove the lines setting timezones though.

api/app/Models/Pool.php
vd1992: aye aye

esizer (approved):
Nice! 👍

PR #3361, opened 2022-07-15
workflow troubleshooting

vd1992:
trying to troubleshooting the docker/e2e issues...

okay, each commit represents a solution to the recent issue of workflows for smoketest and cypress
both appear to resolve the issue, I am partial to the latter since I do not know if there is a purpose to the postrun save, disabling it solves the issue for sure and shaves some time off test runs

Screen Shot 2022-07-18 at 12 04 01 PM

related to docker layer caching action: https://github.com/satackey/action-docker-layer-caching/issues/139#issuecomment-1007316528

per conversation here
https://talent-cloud.slack.com/archives/C0259G6KXJ6/p1657839973910039

### DIFF
--- a/.github/workflows/docker-smoke-test.yml
+++ b/.github/workflows/docker-smoke-test.yml
@@ -33,6 +33,11 @@ jobs:
       # Use buildkit to speed up docker command
       DOCKER_BUILDKIT: 1
     steps:
+
+      # See: https://github.com/satackey/action-docker-layer-caching/issues/139#issuecomment-1007316528
+      - name: reclaim space on runner
+        run: rm -rf /usr/local/android /usr/share/dotnet /usr/local/share/boost /opt/ghc
+
       - name: Checkout
         uses: actions/checkout@v2.3.4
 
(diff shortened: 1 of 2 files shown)


vd1992:
looking at some previous runs, this one for example, post-run took 6 minutes
https://github.com/GCTC-NTGC/gc-digital-talent/runs/7346142902?check_suite_focus=true

petertgiles:
What are the "docker/e2e issues"?  I'm not sure what this branch is solving.

patcon:
Nice! If I'm understanding correctly, 8782930 looks like it also resolved it by clearing up space on the VM? This feels like the least change to get things working.

(I believe that the `skip-save: true` alternative totally turns off updating of the cache, which is the same as removing the cache step.)

Preference is for the first solution 8782930, since it preserves caching. But if you do go with the `skip-save` option (because I'm missing some consideration), I'd propose adding a comment and maybe a TODO to "re-enable once PR #### is resolved", as my experience is that temporary work-arounds get forgotten if not annotated in code :)

patcon (changes requested):
Preference for solution in prior commit, unless I'm missing some consideration :)

vd1992:
> Preference for solution in prior commit, unless I'm missing some consideration :)

aye aye

patcon (approved):
Thanks!

vd1992:
> Thanks!

workflows please cooperate now!

PR #3362, opened 2022-07-15
Improve first-time setup experience via doc update

patcon:
The bad experience of @JoshBeveridge on Windows WSL seemed to involve a memory issue that was resolved by spinning down mock-auth during npm package resolution. Added a hint.


patcon:
(Sorry, over-ambitiously hit merge when I was shutting down my tabs end-of-day, but ideally would have awaited feedback on below comment before doing that...!)

To clarify, there was no error message, it just stalled for like 20 minutes on `npm install`. I'll add "unexpectedly hangs without error" in a quick self-merged PR to accommodate that feedback :)

PR #3385, opened 2022-07-18
Bug - Fix @formatjs warning for default rich text elements

esizer:
## Summary

This changes our compile method from string to [AST](https://formatjs.io/docs/guides/advanced-usage).  I can't tell you why this is the solution but it should actually speed up the application as well as fix the `console.warn` we are getting during tests.

Ref: https://stackoverflow.com/a/71384834


patcon (approved):
jest tests pass, so LGTM! Thanks Eric!

PR #3357, opened 2022-07-15
Restore old labels for old search page

petertgiles:
This branch restores the old Conditions of Employment labels for the old version of the search page.

Testing
- [ ] Refresh your local environment including intl-compile for talentsearch
- [ ] Set the `FEATURE_APPLICANTSEARCH` flag to false and recreate your PHP container
- [x] Confirm the old labels for CoE appear on the search page
image
- [x] Confirm the French labels are loaded, too
image

PR #3388, opened 2022-07-18
Use v1 Conditions of Employment on MVP pool candidate forms

### LINKED ISSUE #3387: Show the v1 conditions of employment on Admin PoolCandidate forms
# Description
The Admin CreatePoolCandidate and UpdatePoolCandidate forms should use the same four v1 conditions of employment as we use on the deprecated/SearchForm. Just like the deprecated SearchForm, these PoolCandidate forms will no longer be used after we widely release Applicant Profiles, but they should use the same Conditions of Employment until then.

tristan-orourke:
Resolves #3387 

To test:
- [ ] Set the FEATURE_APPLICANTSEARCH flag to false and recreate your PHP container
- [ ] Run refresh scripts
- [ ] If necessary, run data seeders with `docker-compose exec -w /var/www/html/api php sh -c "php artisan db:seed"`
- [ ] In one tab, open the search page (http://localhost:8000/en/talent/search) and find the Conditions of Employment options
- [ ] In another tab, get to the Create Pool Candidate form:
    1. Go to http://localhost:8000/en/admin/pools
    2. Select "View Candidates" for a pool
    3. Click "Create Pool Candidate" in top right corner
- [ ] In a third tab, get to the Update Pool Candidate form:
    1. Go to http://localhost:8000/en/admin/pools
    2. Select "View Candidates" for a pool
    3. Select "Edit" for a pool candidate
- [ ] For both forms, the Operational Requirements should show the same options with the same text as on the Search page. The text should match in both English and French. They should be the options shown in https://github.com/GCTC-NTGC/gc-digital-talent/pull/3357.


vd1992:
and in french
hm... doesnt appear to match here

image

vd1992:
as an aside

this is off, on /search
image

I think it is a quick fix, don't think that `/span` is needed and should be primary instead
image

petertgiles:
I mention this over here too, https://github.com/GCTC-NTGC/gc-digital-talent/pull/3432#issuecomment-1190684079 , but I much prefer the approach of moving files into deprecated directories that we've doing with components rather than using "v" suffixes.

tristan-orourke:
Closing this in favour of alternative PR #3432

PR #3389, opened 2022-07-19
Bugfix: Added formatjs --ast flag to appropriate command.

patcon:
Fixes bug in #3385 

@tristan-orourke [in Slack](https://app.slack.com/client/T5ZFRSQ3V/threads/thread/C0259G6KXJ6-1658167365.124319):
> Were you both able to run intl-extract like this? Now npm run intl-extract throws an error when I run it locally: error: unknown option '--ast'.


esizer (approved):
Oops, thanks!

PR #3405, opened 2022-07-19
semi-refactor WorkPreferencesSection

### LINKED ISSUE #3403: Unselecting optional fields for work preferences cause summary to mess up
## To reproduce
1. Log in
2. Navigate to profile and complete the Work Preferences section including checking some of the optional checkboxes.
3. Save and confirm the summary is correct on the profile page.
4. Go back into the Work Preference sub-page and uncheck all the optional checkboxes
5. Save and confirm the summary section is borked.

This is what it looks like for me:
image

##Desired behavior
1. The null state for this summary should list all the unselected optional fields like this:
image
2. The checkbox fields are not required for work preferences, so they shouldn't affect the part of the summary page that says "There are required fields missing."
3. The duration selection should still show up properly once all optional fields are unchecked.
image

vd1992:
resolves #3403 

The look of the WorkPreferences section needs some changing. Now what is _and_ what isn't selected needs to be represented.
As well as what appears required and how things look nulled. 

To test, go to /profile and view the Work Preferences section and...
- test null for wouldAcceptTemporary
- zero selected requirements
- some selected requirements
- all selected requirements

### DIFF
--- a/frontend/common/src/components/UserProfile/ProfileSections/WorkPreferencesSection.tsx
+++ b/frontend/common/src/components/UserProfile/ProfileSections/WorkPreferencesSection.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import { useIntl } from "react-intl";
 import { isEmpty } from "lodash";
 import { getOperationalRequirement } from "../../../constants/localizedConstants";
-import { Applicant } from "../../../api/generated";
+import { Applicant, OperationalRequirement } from "../../../api/generated";
 
 const WorkPreferencesSection: React.FunctionComponent<{
   applicant: Pick<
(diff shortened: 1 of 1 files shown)


frontend/common/src/components/UserProfile/ProfileSections/WorkPreferencesSection.tsx
```diff
@@ -26,45 +26,107 @@ const WorkPreferencesSection: React.FunctionComponent<{
...
   return (
     <div
       data-h2-bg-color="b(lightgray)"
       data-h2-padding="b(all, m)"
       data-h2-radius="b(s)"
     >
-      {anyCriteriaSelected && (
+      {wouldAcceptTemporary === null && (
         <p>
+          <p>
```
esizer, line 65: Can we remove on of the `<p>` here so we don't nest them?

frontend/common/src/components/UserProfile/ProfileSections/WorkPreferencesSection.tsx
```diff
@@ -26,45 +26,107 @@ const WorkPreferencesSection: React.FunctionComponent<{
...
+  const unselectedOperationalArray = operationalRequirementsSubsetV2.filter(
+    (requirement) => !acceptedOperationalRequirements?.includes(requirement),
+  );
+
+  // generate list of unaccepted operational requirements
+  const unacceptedOperationalArray = unselectedOperationalArray
+    ? unselectedOperationalArray.map((opRequirement) => (
+        <li data-h2-font-weight="b(700)" key={opRequirement}>
+          {opRequirement
+            ? getOperationalRequirement(opRequirement).defaultMessage
```
esizer, line 51: These should be wrapped in `intl.formatMessage` so that `react-intl` can properly localise them.

```suggestion
            ? intl.formatMessage(getOperationalRequirement(opRequirement))
```

esizer (approved):
Looking great so far, just two minor things :)

esizer (changes requested):
I'm not sure what happened but i accidentally marked this as approved 😱

frontend/common/src/components/UserProfile/ProfileSections/WorkPreferencesSection.tsx
vd1992: tackled in [this](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3405/commits/268a3d7742ca12e773ce6faa3793156204ca656e)

esizer: You are going to hate me, but i think the wrong tag was removed leading to a string not being wrapped in the `<p>`

vd1992: hopefully all good now

esizer (approved):
Awesome!

PR #3421, opened 2022-07-19
Bug/3409 download buttons on wrong pages

### LINKED ISSUE #3409: Download buttons show up in suspicious places
I'm adding Skill Families on the admin portal and noticed that the pagination component has the Download buttons. I don't think they make sense here (and are definitely non-functional on the page).

The fix would just be to removed the download buttons from all pages that are not the "candidates" page.

JamesHuf:
Resolves #3409.

Removes the download buttons from all the tables but Users because none of them have row selection at this point.

Notes:
- Removes the download buttons from the old frontend-paginated table component since it can't currently handle row selection.
- Hides the download buttons in the backend paginated table unless specified.
- Set the "Download XML" button to always be disabled since we don't have functionality for it yet anyway.

### DIFF
--- a/frontend/admin/src/js/components/Table/Table.tsx
+++ b/frontend/admin/src/js/components/Table/Table.tsx
@@ -289,36 +289,6 @@ function Table<T extends Record<string, unknown>>({
         data-h2-radius="b(none, none, s, s)"
         data-h2-padding="b(all, s)"
       >
-        <div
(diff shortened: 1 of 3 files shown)


esizer (approved):
Nice!  🎉

petertgiles:
Chromatic diff approved.

PR #3430, opened 2022-07-20
Fixed markup typo in translation string.

patcon:
Re-ticketed from https://github.com/GCTC-NTGC/gc-digital-talent/pull/3388#issuecomment-1189544806


patcon:
Actually, this is so small that I think self-merge is alright :)

PR #3432, opened 2022-07-20
Allow v1 and v2 conditions of employment (alternative)

patcon:
Resolves https://github.com/GCTC-NTGC/gc-digital-talent/issues/3387 (alt approach from https://github.com/GCTC-NTGC/gc-digital-talent/pull/3388)

Extracted from https://github.com/GCTC-NTGC/gc-digital-talent/pull/3249

the V1/V2 list refers to the subset of _inputs_ (the enum subsets), while the label variant key is more about "how do we show this" and is not necessarily time-specific (though it can be), e.g., we could create `shortV1` variant if there are still differences.

cc @vd1992 

To test:
- [x] Set the FEATURE_APPLICANTSEARCH flag to false and recreate your PHP container
- [x] Run refresh scripts
- [x] If necessary, run data seeders with `docker-compose exec -w /var/www/html/api php sh -c "php artisan db:seed"`
- [x] In one tab, open the search page (http://localhost:8000/en/talent/search) and find the Conditions of Employment options
- [x] In another tab, get to the Create Pool Candidate form:
    1. Go to http://localhost:8000/en/admin/pools
    2. Select "View Candidates" for a pool
    3. Click "Create Pool Candidate" in top right corner
- [x] In a third tab, get to the Update Pool Candidate form:
    1. Go to http://localhost:8000/en/admin/pools
    2. Select "View Candidates" for a pool
    3. Select "Edit" for a pool candidate
- [x] For both forms, the Operational Requirements should show the same options with the same text as on the Search page. The text should match in both English and French. They should be the options shown in https://github.com/GCTC-NTGC/gc-digital-talent/pull/3357.


frontend/admin/src/js/stories/Pool.stories.tsx
```diff
@@ -7,11 +7,11 @@ import {
   fakeUsers,
   fakePools,
 } from "@common/fakeData";
+import { OperationalRequirementV2 } from "@common/constants/localizedConstants";
```
patcon, line 10:
```suggestion
import { OperationalRequirementV1 } from "@common/constants/localizedConstants";
```

ah this should be V1 to match the Pool components

frontend/admin/src/js/stories/Pool.stories.tsx
```diff
@@ -60,8 +60,8 @@ stories.add("Update Pool Form", () => {
     assetCriteria: [fakeCmoAssets()[0], fakeCmoAssets()[1]],
     essentialCriteria: [fakeCmoAssets()[0], fakeCmoAssets()[1]],
     operationalRequirements: [
-      OperationalRequirement.DriversLicense,
-      OperationalRequirement.OnCall,
+      OperationalRequirementV2[0],
+      OperationalRequirementV2[1],
```
patcon, line 64:
```suggestion
      OperationalRequirementV1[0],
      OperationalRequirementV1[1],
```

frontend/common/src/constants/localizedConstants.tsx
```diff
@@ -649,14 +657,75 @@ export const operationalRequirements = defineMessages({
   },
 });
 
+export const operationalRequirementLabelShort = defineMessages({
```
patcon, line 660: These are the short strings used in the fancy filter with limited space. Will they suffice as the v1 labels, comparable to this:
https://github.com/GCTC-NTGC/gc-digital-talent/pull/3388/files#diff-fb251ec51456fdc6fbe09424d4f94b19ecb7dceac37971e853209b885bac2147R661-R699

If not, we could either make another label variant, or you can suggest the important changes, and we'll just have "not-quite-the-shortest" values for the fancy filter :)

frontend/talentsearch/src/js/components/search/deprecated/SearchForm.tsx
```diff
@@ -288,9 +283,9 @@ export const SearchForm: React.FunctionComponent<SearchFormProps> = ({
                 "Legend for the Conditions of Employment filter checklist",
             })}
             name="operationalRequirements"
-            items={operationalRequirementsSubsetV1.map((value) => ({
+            items={OperationalRequirementV1.map((value) => ({
               value,
-              label: intl.formatMessage(getOperationalRequirement(value)),
+              label: intl.formatMessage(getOperationalRequirement(value, "short")),
```
patcon, line 288: This is where we grab the short form for the deprecated component, instead of default "full".

patcon:
Ah shoot, I just realized this needs `fr.json` transposed, if all else looks good.
https://github.com/GCTC-NTGC/gc-digital-talent/pull/3388/files#diff-45bc9afedaf7884ee2994bbbb15ba5f3aeabd2e5703ccbcf62070afbf7b4bef9

I'm realizing this might be one of those places where autogenerated ID's are less helpful -- these are strings representing variations where the description might change to better describe differences or to give better hints to translators (this string is "short-form", and so be less verbose). In this case, the way that autogen IDs completely untether strings on each change, that feels inconvenient.

Might it be worth disabling the `no-id` lint rule, so that we can add manual ones in specific places and avoid this tedium of shifting IDs as we do housekeeping? We could also just decide to do eslint ignore that manual id lines, and leave the rule in-place.

petertgiles:
This is just a personal preference but I don't really like the v1 and v2 suffixes.  It's not obvious what the current "v" is at given point in time and I think it's kind of ugly.  :smile: 

I much prefer the approach being taken with many of the search and pool components where the current components have the standard name and location and deprecated components are moved to a separate "deprecated" directory.  You can even alias the import to make it extra obvious like this:
https://github.com/GCTC-NTGC/gc-digital-talent/blob/5df77a8f4b019abde493ec7f695b9d56c237fad3/frontend/admin/src/js/components/PoolDashboard.tsx#L26
When it comes time to clean out the old code you can just delete the directory with the files and the compiler will tell you if you missed anything at build time.  All the current components keep the clean name and location.

Also, I'm not sure if you're aware but I already started this with conditions of employment with this PR: https://github.com/GCTC-NTGC/gc-digital-talent/pull/3357

frontend/admin/src/js/stories/Pool.stories.tsx
tristan-orourke: I think leaving v2 in the Pool forms is fine - we're not expecting any new pools to be created with v1.

tristan-orourke (changes requested):
I'm happy with this. The main blocker now is needing to update the french strings.

> Might it be worth disabling the no-id lint rule, so that we can add manual ones in specific places and avoid this tedium of shifting IDs as we do housekeeping?

That sounds like a great idea to me. I really like being able to default to autogenerated ids, but when we keep changing them like this its more of a pain, and manual ids would help.

> @petertgiles preferring putting things in a deprecated file/folder instead of version numbers

You make a good argument. I think in this case we should keep the short version of requirements around, but the v1 list could be in a file marked deprecated. It just needs to be in the common project, so that admin and talentsearch can both use it.

tristan-orourke:
@patcon I've moved some translations around, and used some manual ids this time.
@vd1992 could you give it another look-over?

vd1992:
I guess this might as well be tacked on from the other PR
https://github.com/GCTC-NTGC/gc-digital-talent/pull/3388#issuecomment-1189544806

tristan-orourke:
@vd1992 I think @patcon fixed the span in another PR

tristan-orourke:
@petertgiles I like the idea of figuring out a more consistent "deprecated" pattern, but this fix is more urgent for CMO than I originally realized. For expediency, I'd like to merge it as is for now.

vd1992:
please fix the merge issue @patcon

frontend/admin/src/js/stories/Pool.stories.tsx
patcon: K, will leave, and ensure update story matches the choice. thanks!

frontend/common/src/constants/localizedConstants.tsx
patcon: > maybe those messages could be defined in the component instead of the general localizedConstants?

storing label variants in components admittedly goes pretty counter to how I was hoping to eventually consolidate into an option builder hook 🤔 

I don't think I fully understand the rationale for why these variants can't be merged when they're so close, but I'll assume the text is very dialled in, and I'll just create a new label variant set in the fancy filter PR, or just use the slightly longer text as-is :)

patcon: Ah, just now realized you'd already done this...!

patcon:
# Search
Screen Shot 2022-07-21 at 3 14 18 PM


# Create Pool Candidate
Screen Shot 2022-07-21 at 3 17 22 PM


# Update Pool Candidate
Screen Shot 2022-07-21 at 3 21 36 PM

patcon:
K conflict resolved, merging as soon as passing

patcon:
Ok, going to just bump bundlwatch in mainline, but not sure what's up with that failing jest test 🤔 

Will check into it

patcon:
k worked on rerun. Will just bump bundlewatch direct on mainline and not wait for rebuild

PR #3433, opened 2022-07-20
Add suspicious download buttons to app

patcon:
h/t @gobyrne 

EDIT: To clarify, this is not a real PR 😛

### DIFF
--- a/frontend/common/src/components/Button/Button.tsx
+++ b/frontend/common/src/components/Button/Button.tsx
@@ -195,7 +195,7 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
         }}
         {...rest}
       >
-        {children}
+        Download
       </button>
     );
   },


frontend/common/src/components/Button/Button.tsx
```diff
@@ -195,7 +195,7 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
         }}
         {...rest}
       >
-        {children}
+        Download
```
mnigh, line 198: Is there a reason why you have made this a static value for a reusable component like Button? I am confused.

mnigh: Might this PR be related to https://github.com/GCTC-NTGC/gc-digital-talent/issues/3409? If so, there is a PR that has resolved that issue: https://github.com/GCTC-NTGC/gc-digital-talent/pull/3421.

PR #3439, opened 2022-07-20
Bug - 3311 - Fix Auth e2e Workflow Failing

### LINKED ISSUE #3311: Cypress e2e test failing sporadically (admin-workflows.spec.js)
We're getting sporadic failures of one particular e2e test:
https://github.com/GCTC-NTGC/gc-digital-talent/blob/59bdda46a6672701aa64b7b207c93ded954365af/frontend/cypress/integration/admin/admin-workflows.spec.js#L71-L103

It's happening in line 94:
https://github.com/GCTC-NTGC/gc-digital-talent/blob/59bdda46a6672701aa64b7b207c93ded954365af/frontend/cypress/integration/admin/admin-workflows.spec.js#L94

https://github.com/GCTC-NTGC/gc-digital-talent/blob/59bdda46a6672701aa64b7b207c93ded954365af/frontend/cypress/integration/admin/admin-workflows.spec.js#L15-L24

First failing run I could see with same error:
https://github.com/GCTC-NTGC/gc-digital-talent/runs/7239951241?check_suite_focus=true

List of all e2e tests runs:
https://github.com/GCTC-NTGC/gc-digital-talent/actions/workflows/e2e-tests.yml?page=3&query=

esizer:
May Resolve #3311 

## Summary

Adds a `cy.wait` for the AllUsersPaginated query to complete when searching for a user.

## Note

Not sure if this will fix, going to run it a few times to see since it seems to be intermittent.

### DIFF
--- a/frontend/cypress/integration/admin/admin-workflows.spec.js
+++ b/frontend/cypress/integration/admin/admin-workflows.spec.js
@@ -17,6 +17,8 @@ describe("Admin Workflow Tests", () => {
      .clear()
      .type(name);
 
+     cy.wait("@gqlAllUsersPaginatedQuery");
+
     // wait for table to rerender
     cy.contains(expectedEmail)
       .should("exist")


petertgiles (approved):
Hmm, I didn't think this would work since URQL would cache the query and it wouldn't be fired after the first time.  I ran it a few times locally and it seems to have worked!

esizer:
> Hmm, I didn't think this would work since URQL would cache the query and it wouldn't be fired after the first time. I ran it a few times locally and it seems to have worked!

I'm just as shocked as you are. I just threw this up yesterday to see if it would work. I was thinking "It can't be this simple" lmao

PR #3441, opened 2022-07-20
Updated experience accordions to use the correct icon for their type

### LINKED ISSUE #3426: Icons should match type of experience
## Description

The experience accordions should have matching icons depending on the type of experience. And this should be the case for all tabs.

image

image

JamesHuf:
Resolves #3426.

Updated the experience accordions to use the appropriate icon for their type.


esizer (approved):
Nice! Much better 😁

PR #3447, opened 2022-07-21
Tooling - 1463 - Improve Webpack Optmizations

### LINKED ISSUE #1463: Ensure our webpack configuration is doing some basic optimizations
# Description
After getting away from the Laravel Mix defaults, we may need to some more configuration ourselves to ensure Webpack is optimizing our build (such as tree shaking).

esizer:
Resolves #1463

## Summary

This introduces the `css-minimizer-webpack-plugin` in order to minimize our CSS files as well as `sideEffects` options in `package.json`.

## Bundle Watch!

`Everything is in check (+0B, -314.26KB, -42.1%)`

### CSS Minimizer

Our CSS files are now being minimized and have changed in file size.

- admin: `72kb -> 69kb`
- indigenousapprenticeship: `70kb -> 68kb`
- talentsearch: `68kb -> 66kb`

### Tree Shaking

We have added the `sideEffects` option in our `package.json` files to hint to webpack which files which are imported but not used do not contain dead. Essentially, this is only our `common` css files.

No noticeable change in bundle size.

### Code Splitting

Updated to use `React.lazy` on all route components to reduce initial bundle size and only load components on demand.

- admin: `1.26mb -> 558kb`
- talentsearch: `1.12mb -> 612kb`

#### Note

> A majority of this is updating the imports to de default to support `React.lazy`


frontend/admin/webpack.base.js
```diff
@@ -99,6 +100,12 @@ module.exports = {
       },
     ],
   },
+  optimization: {
```
petertgiles, line 103: Does this take a noticable amount of time to run?  This seems like something that should only be in the PROD webpack config.  It may get in the way of troubleshooting, too.

petertgiles (commented):
I did a quick scan of the code.  The results are exciting but I won't pretend to really understand how it works.  :-)  I'll try to do a more thorough exercising of the code when I get back on Monday if no one else gets to it first.

frontend/admin/webpack.base.js
petertgiles: Hmm, you're right it only runs in production even thought the config is in the base file.  Is there a way to make a default behavior explicit so that it's clearer?  Or at least a comment to flag non-obvious behavior?  Regarding speed, prod build is already quite a bit slower and I'm not noticing any further slowdown.

esizer: Sounds good, I added a comment with [add note that optim runs only in prod mode](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3447/commits/d7974d93ddb09d2a73e2ae0b04414d718ed6165b). Let me know if it needs some additional information.

frontend/talentsearch/src/js/components/Router.tsx
```diff
@@ -22,28 +21,65 @@ import {
...
-import PoolApplicationThanksPage from "./pool/PoolApplicationThanksPage";
-import RegisterPage from "./register/RegisterPage";
-import LoggedOutPage from "./loggedOut/LoggedOutPage";
-import LoginPage from "./login/LoginPage";
-import { CreateAccount } from "./createAccount/CreateAccountPage";
 import { Role } from "../api/generated";
-import PoolAdvertisementPage from "./pool/PoolAdvertisementPage";
+
+/** Search */
+const SearchPage = React.lazy(() => import("./search/SearchPage"));
```
petertgiles, line 28: I observed a cypress failure to load a chunk during a local run though I'm not sure why.  When I opened my browser interactively it didn't have a problem.  It was successful when I ran it a second time, too.  Is this going to increase the number of random cypress failures?  Can we mitigate this somehow?
image

petertgiles (approved):
Overall, this worked really well for my manual testing.  I couldn't observe any unusual behavior at all.

I did observe a cypress failure to load a chunk during a local run though I'm not sure why.  When I opened my browser interactively it didn't have a problem.  It was successful when I ran it a second time, too.  Is this going to increase the number of random cypress failures?  Can we mitigate this somehow?

image

I'd recommend trying to get a second review before merging this if possible.

petertgiles:
Hmm, I thought of something else to try.  I think I'll deploy this to the dev server tomorrow to make sure it goes through the build and deploy properly.

petertgiles:
I deployed to dev.  Seems to work well.  🥳

vd1992:
ran `npm run dev` , no issues and can browse the the app

yonikid15 (approved):
Wow amazing work Eric 🎉 Everything is building smoothly, and I'm not seeing any unusual behaviour when testing throughout the site manually.

PR #3452, opened 2022-07-21
Hide experience types if user has none

### LINKED ISSUE #3411: Hide types of Experience that have not been used on profile
## Description
In the "My experience and skills" sub-page of profile, when you sort by "Type", you are shown all the types of skills, even if you haven't entered any skills of that type yet. It looks a little clunky.

Please hide all types that have no experiences associated with them.

Using Chrome on UAT 2022-07-19.

## Screenshot: Prototype
image

JamesHuf:
Resolves #3411.

Hides types of experience the user has none of when ordering experiences by type.

### DIFF
--- a/frontend/common/src/components/UserProfile/ExperienceByTypeListing.tsx
+++ b/frontend/common/src/components/UserProfile/ExperienceByTypeListing.tsx
@@ -96,41 +96,51 @@ const ExperienceByTypeListing: React.FunctionComponent<
 
   return (
     <>
-      <ExperienceByType
-        title={intl.formatMessage({ defaultMessage: "Personal" })}
-        icon={<LightBulbIcon style={{ width: "1.5rem" }} />}
-        experiences={personalExperiences}
-        defaultOpen={defaultOpen}
-        experienceEditPaths={editPaths}
-      />
-      <ExperienceByType
-        title={intl.formatMessage({ defaultMessage: "Community" })}
-        icon={<UserGroupIcon style={{ width: "1.5rem" }} />}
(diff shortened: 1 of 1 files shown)


frontend/common/src/components/UserProfile/ExperienceByTypeListing.tsx
```diff
@@ -96,41 +96,51 @@ const ExperienceByTypeListing: React.FunctionComponent<
...
-      />
+      {personalExperiences.length > 0 && (
+        <ExperienceByType
+          title={intl.formatMessage({ defaultMessage: "Personal" })}
+          icon={<LightBulbIcon style={{ width: "1.5rem" }} />}
+          experiences={personalExperiences}
+          defaultOpen={defaultOpen}
+          experienceEditPaths={editPaths}
+        />
+      )}
```
esizer, line 107: When checking length, it's safer to use the following pattern as this may output `0` or `false` to the screen.

```tsx
{array.length ? (<Component />) : null}
```

esizer (changes requested):

frontend/common/src/components/UserProfile/ExperienceByTypeListing.tsx
JamesHuf: Changed in [fb6fcdd](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3452/commits/fb6fcdd3092b26acce1e1e50bab554c3e73ee333).

esizer (approved):
Awesome! :D

PR #3456, opened 2022-07-22
Debt - 3212 - Combine Colour Variant Stories

### LINKED ISSUE #3212: Combine Individual Variant Stories into Single Story
## Description

Currently, we have individual stories for different variants of components which is resulting in a large number of chromatic snapshots. The proposal is to combine the individual variants into a single story where appropriate to reduce the number of snapshots.

Example: https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%A7%AC-atoms-cc-button--disabled

## Acceptance Criteria
- [ ] Reduce number of stories

esizer:
Resolves #3212 

## Summary

Combines colour variants of stories to reduce the number of snapshots.

### DIFF
--- a/frontend/common/src/components/Button/Button.stories.tsx
+++ b/frontend/common/src/components/Button/Button.stories.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import { Story, Meta } from "@storybook/react";
 import { action } from "@storybook/addon-actions";
 import Button from "./Button";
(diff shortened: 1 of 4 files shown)


vd1992:
think this one really needs a snapshot

yonikid15 (approved):
Checked everything out locally and it looks good 👍 Nice work!

PR #3482, opened 2022-07-25
Use common source for Employment Equity statement messages

### LINKED ISSUE #3408: Equity statements are different on profile and equity pages
## Description

- Equity statements are different on the **profile page** and the **diversity, equity and inclusion page**.
- Equity statements are in a different order on the **profile page** and the **diversity, equity and inclusion page**.

## Screenshots
Screen Shot 2022-07-19 at 15 44 41
Screen Shot 2022-07-19 at 15 44 10

## Acceptance Criteria
- [ ] Equity statements should be from the same source in every instance
- [ ] Equity statements should be displayed in the same [order](https://github.com/GCTC-NTGC/gc-digital-talent/issues/3408#issuecomment-1189510198) in every instance

mnigh:
Resolves #3408.


frontend/common/src/constants/localizedConstants.tsx
```diff
@@ -25,25 +25,34 @@ import {
...
+    description: "Statement for when someone indicates they have a disability",
+  },
+});
 
-export const disabilityLocalized = {
-  defaultMessage: "I identify as a person with a disability",
-  description: "Message for person with a disability",
-};
+export const getEmploymentEquityStatement = (
+  equityStatement: string,
```
petertgiles, line 49: Could we further constrain this type beyond string to make sure that a missing key is a compile-time error instead of runtime?  Something like `keyof employmentEquityStatements` or make a union type or enum or a set of constants?

frontend/talentsearch/src/js/components/diversityEquityInclusion/EquityOptions.tsx
```diff
@@ -89,50 +90,40 @@ const EquityOptions: React.FC<EquityOptionsProps> = ({
...
-                    title={intl.formatMessage({
-                      defaultMessage:
-                        '"I Identify as a person with a disability"',
-                      description:
-                        "Title for when someone indicates they have a disability on their profile",
-                    })}
+                    onOpen={() => setWomanOpen(true)}
+                    title={intl.formatMessage(
+                      getEmploymentEquityStatement("woman"),
+                    )}
```
esizer, line 99: More a question than a request, but should we also make this change for the unresolved items? Or at least move them to `localizedConstants.tsx`?

frontend/common/src/constants/localizedConstants.tsx
```diff
@@ -25,25 +25,34 @@ import {
...
-export const indigenousLocalized = {
-  defaultMessage: "I am indigenous",
-  description: "Message for indigenous",
-};
-
-export const minorityLocalized = {
-  defaultMessage: "I identify as a member of a visible minority group",
-  description: " Message for minority group",
-};
+export const employmentEquityStatements = defineMessages({
```
petertgiles, line 28: Did you want to add a set of strings for the search page that use the same keys?  It looks like these don't quite match Jerbo's option strings either.  Could be out of scope, I guess.

image

petertgiles (changes requested):
Nice!  It works great.  A few small suggestions below.

frontend/talentsearch/src/js/components/diversityEquityInclusion/EquityOptions.tsx
mnigh: Good catch. I think this is out of scope though and have created a new issue:

- https://github.com/GCTC-NTGC/gc-digital-talent/issues/3484

frontend/common/src/constants/localizedConstants.tsx
mnigh: Good catch. I think this is out of scope though and have created a new issue:

- https://github.com/GCTC-NTGC/gc-digital-talent/issues/3484

petertgiles (approved):
Very nice!

esizer (approved):
Great! 🥇

Issue #3128, opened 2022-06-20
Migrate Select to SelectFieldV2


mnigh:
### Description

We have Select and MultiSelect, which are very different underneath. We wish to migrate from using these prior ones to using the newer implementations of SelectFieldV2 and MultiSelectFieldV2, which both use react-select underneath, and have much more shared code.

This completes work done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/3510

Proposed approach:
- write cypress/jest tests that validate that important Select form submissions work
- replace `<Select>` with `<SelectFieldV2>`
  - wrap v1 around v2, ensuring types play nice together
  - replace all v1 with v2, making any change required if types need massaging
  - rename SelectFieldV2 to SelectField
  - delete Select
- fix cypress/jest tests as needed

## Acceptance Criteria
- [ ] Select should be replaced with SelectField
- [ ] all the forms that previously used Select should submit properly (pay attention to default and cleared fields)

patcon:
As part of this, maybe we could start using `FooField` (e.g. `SelectField`) for components that include our label wrappers, and just `Foo` (e.g., `Select`) for our custom versions of input elements. What we're calling Select is perhaps clearer to call SelectField (with labels and all our chrome), and then there's room for a lightweight wrapper around React's Select that's just our customization of their component, like a `CustomSelect` (which is how they recommend modifying their component in their docs)

patcon:
un-assigning since this is not being worked on atm

petertgiles:
Due to the a11y concerns (Fable testing failures) with our SelectFieldV2 components we've decided to abandon this migration and continue using Select as-is.


https://github.com/gctc-ntgc/gc-digital-talent/issues/3386
https://github.com/gctc-ntgc/gc-digital-talent/issues/4995

Sorry James.  :cry:

PR #3510, opened 2022-07-26
Add SelectFieldV2 and MultiSelectV2 components

patcon:
Extracted from https://github.com/GCTC-NTGC/gc-digital-talent/issues/2641

What this does:
- Adds components for using react-select (single select and multi-select), with MultiSelectFieldV2 as slim extension of SelectFieldV2
- adopts "XField" nomenclature to better communicate that component encloses more logic than just form/input element (label, context messages, input chrome, etc)
- rather than use react-select's more complex `Option: {value: string; label: string}` format, uses simpler `value` of `string` or `string[]`
- added prop `forceArrayFormValue` to optionally force `SelectFieldV2` to store values as array of single value (e.g., `["BAZ"]`). This should simplify submission logic that can now be more agnostic about field type (Select vs MultiSelect)
- makes text content inside react-select component localizable (e.g, default placeholder "Select...", "Loading...", etc.)
- adds tests for MultiSelectFieldV2 and SelectFieldV2
- allow `name`, `id`, `label` to be derived from one another. Specifically, `name` from `id`, and `id` from `label`. Setting `label` and `id` will result in working component.

## To Do
- [x] spawn issue to eventually migrate Select => SelectFieldV2 and MultiSelect => MultiSelectFieldV2 (remove V2 when full deprecated) https://github.com/GCTC-NTGC/gc-digital-talent/issues/3128

## Open Questions
1. ~~Will we ever need `id` and `name` to be different, or is this just a footgun for react-hook-form and validation? If not, should we just offer one prop and pass it to `InputWrapper` as `inputId`?~~ [EDIT: yes, helpful to allow diff between `id` and `name` for choosing formValue keys (`name`) and also ensuring uniqueness-on-page (`id`). keep both. inheritance of `name` from `id` is best]
2. upstream type definitions of react-hook-form say `rules: { required: true }` is valid, but it confuses our validation message logic -- it silently refuses to submit, but doesn't display a message [(example)](https://main--61e099a1bb1465003a73d3ce.chromatic.com/?path=/story/form-select--select-required&args=rules.required:true). Should we add a default message like "This field is required." when one isn't offered?

### DIFF
--- a/frontend/common/src/components/form/MultiSelect/MultiSelectFieldV2.stories.tsx
+++ b/frontend/common/src/components/form/MultiSelect/MultiSelectFieldV2.stories.tsx
@@ -16,7 +16,8 @@ export default {
           onSubmit={action("Submit Form")}
           options={{ defaultValues: { departments: [] } }}
         >
-          <Story />
+          { /* See: https://github.com/storybookjs/storybook/issues/12596#issuecomment-723440097 */ }
+          {Story() /* Can't use <Story /> for inline decorator. */ }
           <Submit />
         </BasicForm>
       );
--- a/frontend/common/src/components/form/MultiSelect/MultiSelectFieldV2.stories.tsx
+++ b/frontend/common/src/components/form/MultiSelect/MultiSelectFieldV2.stories.tsx
@@ -16,8 +16,8 @@ export default {
           onSubmit={action("Submit Form")}
           options={{ defaultValues: { departments: [] } }}
         >
-          { /* See: https://github.com/storybookjs/storybook/issues/12596#issuecomment-723440097 */ }
-          {Story() /* Can't use <Story /> for inline decorator. */ }
+          {/* See: https://github.com/storybookjs/storybook/issues/12596#issuecomment-723440097 */}
+          {Story() /* Can't use <Story /> for inline decorator. */}
           <Submit />
         </BasicForm>
       );
--- a/frontend/common/src/components/form/MultiSelect/MultiSelectFieldV2.stories.tsx
+++ b/frontend/common/src/components/form/MultiSelect/MultiSelectFieldV2.stories.tsx
@@ -1,0 +1,67 @@
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+import { action } from "@storybook/addon-actions";
+import { uniqueId } from "lodash";
+import React from "react";
+import BasicForm from "../BasicForm";
+import Submit from "../Submit";
+import MultiSelectFieldV2 from "./MultiSelectFieldV2";
+
+export default {
+  component: MultiSelectFieldV2,
+  title: "Form/MultiSelectFieldV2",
+  decorators: [
+    (Story) => {
+      return (
+        <BasicForm
+          onSubmit={action("Submit Form")}
+          options={{ defaultValues: { departments: "" } }}
+        >
+          <Story />
+          <Submit />
+        </BasicForm>
+      );
+    },
+  ],
+} as ComponentMeta<typeof MultiSelectFieldV2>;
+
+const Template: ComponentStory<typeof MultiSelectFieldV2> = (args) => {
+  return <MultiSelectFieldV2 {...args} />;
+};
+
+export const Default = Template.bind({});
+Default.args = {
+  id: uniqueId(),
+  label: "Departments",
+  name: "departments",
+  options: [
+    { value: 1, label: "CRA" },
+    { value: 2, label: "TBS" },
+    { value: 3, label: "DND" },
+  ],
+};
+
+export const Required = Template.bind({});
+Required.args = {
+  ...Default.args,
+  rules: { required: "This must be accepted to continue." },
+};
+
+export const RequiredWithInfo = Template.bind({});
+RequiredWithInfo.args = {
+  ...Default.args,
+  context: "We collect the above data for account purposes.",
+  rules: { required: "This must be accepted to continue." },
+};
+
+export const RequiredWithError = Template.bind({});
+RequiredWithError.args = {
+  ...Default.args,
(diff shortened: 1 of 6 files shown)


frontend/common/src/components/form/MultiSelect/MultiSelectFieldV2.stories.tsx
```diff
@@ -0,0 +1,68 @@
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+import { action } from "@storybook/addon-actions";
+import { uniqueId } from "lodash";
```
petertgiles, line 3: I know this is just the dev dependency but prefer importing the function from the single file for smaller bundle.  https://stackoverflow.com/a/35251059

frontend/common/src/components/form/MultiSelect/MultiSelectFieldV2.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+  component: MultiSelectFieldV2,
+  title: "Form/MultiSelectFieldV2",
+  decorators: [
+    (Story) => {
+      return (
+        <BasicForm
+          onSubmit={action("Submit Form")}
+          options={{ defaultValues: { departments: [] } }}
+        >
+          { /* See: https://github.com/storybookjs/storybook/issues/12596#issuecomment-723440097 */ }
```
petertgiles, line 19: There are a few linting errors to check on.

frontend/common/src/components/form/Select/SelectFieldV2.tsx
```diff
@@ -0,0 +1,205 @@
+import React from "react";
+import { Controller, useFormContext } from "react-hook-form";
+import type { RegisterOptions } from "react-hook-form";
+import ReactSelect, { components } from "react-select";
+import type { NoticeProps, PropsValue, GroupBase, Options } from "react-select";
+import camelCase from "lodash/camelCase";
+import { useIntl } from "react-intl";
+import { InputWrapper } from "../../inputPartials";
+
+export type Option = { value: string | number; label: string };
```
petertgiles, line 10: > rather than use react-select's more complex Option: {value: string; label: string} format, uses simpler value of string or string[]

This seems to contradict?

petertgiles (commented):
I'm just working through this but I have a few comments and questions.

> spawn issue to eventually migrate Select => SelectFieldV2 and MultiSelect => MultiSelectFieldV2 (remove V2 when full deprecated)

I'm missing some context.  What do these components do that our existing Select and MultiSelect don't?

> Will we ever need id and name to be different, or is this just a footgun for react-hook-form and validation? If not, should we just offer one prop and pass it to InputWrapper as inputId?

Yeah, I saw this recently.  :laughing: If you have two different components on the same page with the same set of options they can easily affect each other if they don't have unique ids or names.  This can easily happen if they're in different dialogs so it's not obvious they are together on the page.  You might want to make a story to prove that this doesn't happen.

> it silently refuses to submit, but doesn't display a message

I don't understand this.  In the example you linked to a message is displayed?  Having a fallback message seems useful though.

frontend/common/src/components/form/MultiSelect/MultiSelectFieldV2.stories.tsx
patcon: Ah thanks! total agreement -- I've def seen the bundlesize diff it makes. Copied this and missed it :)

frontend/common/src/components/form/Select/SelectFieldV2.tsx
patcon: We still need to track it internally for actually building options, but that pull quote is more about it not saving into form value as Option or Option[] :)

frontend/common/src/components/form/MultiSelect/MultiSelectFieldV2.stories.tsx
patcon:
```suggestion
import uniqueId from "lodash/uniqueId";
```

patcon:
> What do these components do that our existing Select and MultiSelect don't

I think I did a poor job of answering this, but I tried to in "what this does". maybe we could jump in a huddle and I can come back here and summarize better for anyone else reviewing?

> If you have two different components on the same page with the same set of options they can easily affect each other if they don't have unique ids or names.

Ah ok, then I think it's good as-is: in this PR, just setting `id` will auto-populate `name` to be the same, but if someone wishes a specific key in formState (rather than the unique-to-page one), they can override. That sound like the right logic to you?

> > it silently refuses to submit, but doesn't display a message
> 
> I don't understand this. In the example you linked to a message is displayed? Having a fallback message seems useful though.

Ah interesting. The querystring in the link doesn't seem to apply as expected.

`rules.required` can be a boolean according to react-hook-form type hints, but our wrappers don't like that:

patcon:
Post-huddle summary:

> What do these components do that our existing Select and MultiSelect don't

Mainly that:
- Existing `<Select>` and `<MultiSelect>` have some big diffs underneath: `<Select>` wraps HTML element `<select>`, and `<MultiSelect>` wraps <ReactSelect isMulti>` in a very single-purpose way.
- These new components work more in the spirit of react-select, where multiselect features are just a variant of the base `<ReactSelect>`. Now, `<MultiSelectFieldV2>` is just sugar on top `<SelectFieldV2 isMulti>`. The cost of this shared logic is a bit more attention to types.
- This approach was informed by an experiment patcon did to get grouped options working (one of many react-select features), where it became more clear what changes were needed to keep this path open. (This does not add support for grouped options, but it should now be more straightforward to implement in both SelectFieldV2 and MultiSelectFieldV2)

petertgiles (approved):
Thanks for talking me through it.  Seems good to me!

patcon:
Thanks peter!

PR #3511, opened 2022-07-26
Debt - 3317 - Improve candidate request validation

### LINKED ISSUE #3317: Add validation that PoolCandidateSearchRequest must include a filter
Re-ticketed from #3256.

# Description
PoolCandidateFilter used to be a required field in when creating a PoolCandidateSearchRequest. Now its not, since the request can have an ApplicantFilter instead. Add validation to ensure at least one of them is defined when creating a new request.

esizer:
Resolves #3317 

## Summary

This adds `required_without` validation to the `createPoolCandidateSearchRequest` mutation for `poolCandidateFilter` and `applicantFilter`.

### DIFF
--- a/.vscode/project-words.txt
+++ b/.vscode/project-words.txt
@@ -2,6 +2,7 @@ alertdialog
 Analyser
 APPLICANTPROFILE
 APPLICANTSEARCH
+Assertable
 Authenticatable
 BEHAVIOURAL
 colour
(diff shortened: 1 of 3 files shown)


vd1992:
for anyone else on a tangent about the world of laravel's built-in validators here you go

https://laravel.com/docs/8.x/validation#available-validation-rules

vd1992:
if neither filter is passed in, it successfully fails

image

vd1992:
passing in `applicantFilter` 

image

passing in `poolCandidateFilter`

image

vd1992 (approved):
appears to work as intended
~also, when both filters passed in it still appears to work, though I wonder if that breaks something on the database end
example: submit `poolCandidateFilter.hasDiploma = false` and `applicantFilter.hasDiploma = true`~

~it is out of scope so this is just thinking~
~example like this~

realized it isn't actually stored in the database by itself, but in `pool_candidate_filters` and `applicant_filters` so nevermind lol

image

PR #3536, opened 2022-07-28
Add default "required" message for SelectField components

patcon:
Forgot to push this in https://github.com/GCTC-NTGC/gc-digital-talent/pull/3510 and didn't want to re-trigger more approval on that one, nor merge it without.

This allows processing of the rules to use a default when `{required: true}` instead of the current behaviour of silently refusing to submit with no UI indication why.


frontend/common/src/components/form/Select/SelectFieldV2.tsx
```diff
@@ -79,6 +79,25 @@ const LocalizedNoOptionsMessage = <
...
+export const useRulesWithDefaultMessages = (rules: RegisterOptions = {}) => {
+  const { formatMessage } = useIntl();
+  const rulesWithDefaults = { ...rules };
+
+  if (rules.required === true) {
+    rulesWithDefaults.required = formatMessage({
+      defaultMessage: "This field is required.",
+      description: "The default validation message when a field is required",
+    });
+  }
```
patcon, line 97: We've already got this message somewhere we can import from

```jsx
import { errorMessages } from "../../../messages";
```

```suggestion
  if (rules.required === true)
    rulesWithDefaults.required = formatMessage(errorMessages.required);
```

esizer:
So, I have a few questions sort of related to each other.

1. Should we be allowing generic default error messages, especially when they are treated like alerts?
2. If yes, can we associate these with the input field via an `id` and `aria-describedby`.
    - Currently, a screen reader would announce "This field is required" with no association which may be difficult for users to know exactly which field is required.
    - It would be nice to have a more contextual error message like "The {fieldName} field is required" or,
    - Using `aria-describedby` to associate the message with the input

Ref: https://www.w3.org/WAI/tutorials/forms/notifications/#listing-errors

patcon:
Ah, I'm seeing that you raised another ticket for (2), right? https://github.com/GCTC-NTGC/gc-digital-talent/issues/3545

Def feel that's important to resolve, but it does seem outside scope of this.

> 1. Should we be allowing generic default error messages, especially when they are treated like alerts?

This PR is just codifying something we're doing in maybe 150 places throughout the app (setting a default error msg), but setting the defaults within SelectField component (rather than each implementation of it):

```tsx
<SelectField rules={{ required: true }} />
```

Instead of 

```jsx
<SelectField rules={{ required: intl.formatMessage(errorMessages.required) }} />
```

We could also use the hook to set other defaults, so that `{ min: 5 }` etc works too. I just wasn't working with that in the UserTableFilterDialog.

It's also making the type hints more honest, as react-hook-form's `rules` type already allows `required: true` be used instead of a string. But if used it confused our components and causes submission to be denied with no error displayed.

esizer:
> Let me know if I'm not understanding what you raised about alerts. I don't want it to get in the way of something you're pushing for a11y!

So since errors have `role=alert` they inherit  `aria-live=assertive` which will have them announced to screen readers to interrupt the current flow. 

As an example, assume you have a form with 4 fields which are required and attempt the submit the form with them not filled out. The screen reader, on submission, announces to the user:

- This field is required
- This field is required
- This field is required
- This field is required

With no context, these error messages are not useful. It may sound like a bug where the screen reader accidentally announced the same field multiple times. As a user, I cannot be certain.

Even though we are already doing this, making it easier to continue this trend by allowing us to do it with `rules={{ required: true}}` makes it more likely to occur.

It is my own opinion that we should be forced to provide a message so that we need to think about context and provide a message like:

- Email is a required field
- First Name is a required field
- Last Name is a required field
- Phone is a required field

patcon:
What about just "This field is required<hidden>: {fieldName}</hidden>." and the hook takes field name as arg?

esizer:
> What about just `This field is required<hidden>: {fieldLabel}</hidden>.` and the hook takes field name as arg?

I like this approach since it actually enforces us to provide context. I also agree that option 2 is likely the best. Even though it sounds kind of awkward, I think you are right that adding the most pertinent information is a positive.

> Tangentially, I wonder if `<screenreaderOnly></screenreaderOnly>` is a tag that will be more legible to translators ("hidden" feels a little unclear about it's purpose, so might result in friction with translations)

While I agree that for devs screenReaderOnly provides better instruction as to the tags purpose, I'm not sure that would extend to translators. Is it a safe assumption that any potential translator is aware of screen readers? 

Would something like `visuallyHidden` make more sense?

patcon:
Happy to punt the element name, since we can't even use it here, due to the issue we discussed in Slack.

But in the spirit of having a11y convos:

> While I agree that for devs screenReaderOnly provides better instruction as to the tags purpose, I'm not sure that would extend to translators. Is it a safe assumption that any potential translator is aware of screen readers?

not sure `<visuallyHidden>` is any better 🙃 -- It already seems clear that it's visually hidden, but it's not clear _why_. And I suspect translators need to know why, because they need to understand this part of accessibility in order to decide what words in the translation are best shown vs hidden. They are translating one string for two audiences, and should prob know that :) 

imho it feels in the spirit of that "whole system" approach to accessibility, to up-skill translators about screenreaders if they don't already

frontend/common/src/components/form/Select/SelectFieldV2.tsx
```diff
@@ -79,6 +81,26 @@ const LocalizedNoOptionsMessage = <
     </components.NoOptionsMessage>
   );
 };
+/**
+ * One-off hook to add default messages to validation rule object in place of booleans.
+ *
+ * @param rules initial rule object
```
patcon, line 87:
```suggestion
 * @param fieldLabel label to pass to default messages for accessibility purposes
 * @param rules initial rule object
```

frontend/common/src/components/form/Select/SelectFieldV2.tsx
patcon: done

patcon:
Re-building storybook and will be reviewable here when done:
https://2641-react-select-fields--61e099a1bb1465003a73d3ce.chromatic.com/?path=/story/form-selectfieldv2--required

Steps:
1. visit story "SelectFieldV2 > Required"
2. submit empty and note default error message
3. visit story "SelectFieldV2 > Required With Custom"
4. submit empty and note custom message.

Bah, just realized that our custom message should model the accessible behaviour :/

There's a very soft and unofficial recommendation for aria-describedby to point to error when input elements are in error state. I think this is valid, but mind if I create a new issue for it?
https://www.w3.org/WAI/GL/wiki/Using_ARIA_role_of_alert_for_Error_Feedback_in_Forms#Example:

esizer:
> There's a very soft and unofficial recommendation for aria-describedby to point to error when input elements are in error state. I think this is valid, but mind if I create a new issue for it?

We actually have an issue for this :) #3545

patcon:
Thanks! Ok, storybook built and ready for final review as described above.

Lemme know if it looks alright

esizer (approved):
Nice! Sorry about all the extra work I requested.

Imo, I really like that it displays the field name in the error 👍

patcon:
Ah jeez, no I'm glad! Thank you thank you for the review

PR #3424, opened 2022-07-20
Feature - Pool advertisement backend mutations

yonikid15:
Resolves #3168 

# Notes
- This PR creates new mutations for the Pool Advertisement pages. 
  - Create, Update, Publish, ExtendClosingDate, Close, and Delete.
- There is a `poolAdvertisementOperations.graphql` file in`talentsearch/src/js/api` with all the mutations which you use to test in the graphql-playground.

PR #3537, opened 2022-07-28
adjust poolFactory to attach single classification

vd1992:
currently the factory attaches 5 random classifications to a pool, going forward like with the pool advertisement the norm is one classification per pool
having multiple classifications attached to a pool causes a failure in validation

candidate classification seeding needed to be adjusted too as there is only a single classification involved with a pool, so each candidate has one expected classification now

see #3424 for context

**testing**
1) check out and migrate+seed
2) visit "Select: classification_pool" and "Select: classification_pool_candidate" in Adminer and ensure they look like what the above says the goal is
3) check anything you might think be suspicious??? not sure

### DIFF
--- a/api/database/factories/PoolFactory.php
+++ b/api/database/factories/PoolFactory.php
@@ -49,7 +49,7 @@ class PoolFactory extends Factory
     {
         return $this->afterCreating(function (Pool $pool) {
             $assets = CmoAsset::inRandomOrder()->limit(4)->get();
-            $classifications = Classification::inRandomOrder()->limit(5)->get();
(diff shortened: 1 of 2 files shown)


esizer:
Just a question so I understand this better, is there a reason why we are using a many to many relationship when we only ever attach a single classification to a pool?

petertgiles:
> Just a question so I understand this better, is there a reason why we are using a many to many relationship when we only ever attach a single classification to a pool?

Yeah, this is future proofing.  No one believes that the client will never ask for more than one.  :laughing:

petertgiles (approved):
Nice!

PR #3547, opened 2022-07-29
Debt - Publish pool mutation error messages

### LINKED ISSUE #3546: Publish pool advertisement error messages
# Description
Temporarily add more user friendly error messages for the `publishPoolAdvertisement` mutation, until issue #3271 is completed.

yonikid15:
Resolves #3546


api/app/GraphQL/Mutations/PublishPoolAdvertisement.php
```diff
@@ -21,9 +19,9 @@ public function __invoke($_, array $args)
         $poolAdvertisement = Pool::find($args['id'])
             ->load(['classifications', 'essentialSkills', 'nonessentialSkills']);
         $poolValidation = new PublishPoolAdvertisementValidator;
-        $validator = Validator::make($poolAdvertisement->toArray(), $poolValidation->rules()); // First validate pool advertisement before updating.
+        $validator = Validator::make($poolAdvertisement->toArray(), $poolValidation->rules(), $poolValidation->messages()); // First validate pool advertisement before updating.
         if ($validator->fails()) {
-            throw new ValidationException($validator);
+            throw FormattedError::setInternalErrorMessage($validator->errors()->first());
```
petertgiles, line 24: I think this is causing the weird debug error:
image
I think this function returns void and you can't throw that.

petertgiles (changes requested):
Very cool!  This is a great reference for future PRs.  One small thing to check below.

api/app/GraphQL/Mutations/PublishPoolAdvertisement.php
yonikid15: Okay so I fixed this problem by using the ValidationException in [commit #4fdf634 ](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3547/commits/4fdf6e42ad9ac81c6c79a1d73f75b5cbbecf02c9). However, this is a temporary fix until the Custom error messages issue gets done. The ValidationException returns all the fields that failed validation, and it would be better for the user if they can see all the inputs that failed at once. 

image

petertgiles: I like it.  The naive implementation can just use `message` while a smarter implementation can examine the extensions.

petertgiles (approved):

PR #3549, opened 2022-07-29
create new pool page

### LINKED ISSUE #3163: New Pool page

## Description
New pool page is an intermediary page that creates new pools based on a couple of required selections. (Classification, Job Title/Stream/Generic????)

This will later on be used to host templates for pools that include pre selected skills and work tasks based on job titles **_(Not relevant to this sprint)._** 

## Screenshot: Prototype
image

## Acceptance Criteria
- [ ] URL "https://talent.canada.ca/en/admin/pools/create"
- [ ] Put behind feature flag "Direct Intake"
- [ ] User creating the pool becomes pool owner by default
- [ ] After new pool is created, user is navigated to the Edit Pool page
- [ ] Cypress test that user gets navigated to next page after submission

[Wireframe for this bad boy](https://www.figma.com/proto/b6lDM86QCJQ6iaAh7qrTYH/Admin-v4-(Direct-Intake)?page-id=0%3A1&node-id=1656%3A5731&viewport=-3442%2C1747%2C0.59&scaling=min-zoom&starting-point-node-id=1618%3A6064)

vd1992:
resolves #3163 

Adds the new create pool page
Key field will be removed in #3561 as it is currently mandatory 

Testing: Create pools and ensure things are reflected in dashboard and adminer

### QUESTION(S)
Prototype has generic job title merged with classification for "Starting group and level", this is currently meaningless as selecting between IT-03 Team Lead and IT-03 Advisor is pointless as it does the same thing. How to resolve this discrepancy? 

### PROTOTYPE SCREENSHOT FOR REFERENCE

image


patcon:
Stream => Generic example from edit pool page (screencap has been updated):
https://github.com/GCTC-NTGC/gc-digital-talent/issues/3162

frontend/admin/src/js/components/pool/CreatePool.tsx
```diff
@@ -2,102 +2,81 @@ import * as React from "react";
...
   key: string;
-  name: {
-    en: string;
-    fr: string;
-  };
-  assetCriteria: string[] | undefined;
-  classifications: string[] | undefined;
-  essentialCriteria: string[] | undefined;
-  owner: string;
+  classification: string[] | undefined;
```
esizer, line 32: Would it be better to just mark this as optional?

```suggestion
  classification?: string[];
```

frontend/admin/src/js/components/pool/CreatePool.tsx
```diff
@@ -2,102 +2,81 @@ import * as React from "react";
...
 interface CreatePoolFormProps {
-  classifications: Classification[];
-  cmoAssets: CmoAsset[];
-  users: User[];
+  userId: string;
+  genericJobTitles: {
+    key: GenericJobTitleKey;
+    id: string;
+    classificationId: string;
+  }[];
```
esizer, line 41: This is personal preference so feel free to ignore. But I find it easier to read when more complex types like this are extracted out. So like:

```tsx
interface GenericJobTitle {
    key: GenericJobTitleKey;
    id: string;
    classificationId: string;
}

interface CreatePoolFormProps {
...
genericJobTitles: GenericJobTitle[]
...
}
```

frontend/admin/src/js/components/pool/CreatePool.tsx
```diff
@@ -188,185 +171,82 @@ export const CreatePoolForm: React.FunctionComponent<CreatePoolFormProps> = ({
...
-              label={intl.formatMessage({
-                defaultMessage: "Key Tasks (French)",
+            <Submit
+              text={intl.formatMessage({
+                defaultMessage: "Create new pool",
                 description:
-                  "Label displayed on the pool form key tasks (French) field.",
+                  "Label displayed on submit button for new pool form.",
               })}
             />
```
esizer, line 192: Looks like the mockup uses the `cta` colour option. Could we add that here?

frontend/admin/src/js/components/pool/CreatePool.tsx
```diff
@@ -188,185 +171,82 @@ export const CreatePoolForm: React.FunctionComponent<CreatePoolFormProps> = ({
               name="key"
```
esizer, line 171: I'm unable to test this because the "key" input returns the following error:

> Expected type KeyString at value.key; The given value "TEST_KEY" did not match the regex /^[a-z]+[_a-z0-9]*$/ 

Should we add some instruction for users on the format since it is a required field?

esizer (changes requested):
Looking great so far! Just a few minor things :)

frontend/admin/src/js/components/pool/CreatePool.tsx
vd1992: given it is required in the form, I dropped the undefined portion of it, hopefully that isn't worse

frontend/admin/src/js/components/pool/CreatePool.tsx
vd1992: good idea, added

frontend/admin/src/js/components/pool/CreatePool.tsx
vd1992: ah, I somehow missed the button colour, added

frontend/admin/src/js/components/pool/CreatePool.tsx
vd1992: key field will be removed in #3561 so I don't think you should worry too much about that one
I've just put strings of random letters for the time being like "asdasdasfas"

vd1992: hang on, I should double check if it being removed or changed

esizer: Sounds good, everything else working well. Once you figure this out let me know and I should be able to approve 👍

vd1992: Key Field will be gone gone per https://github.com/GCTC-NTGC/gc-digital-talent/issues/3561#issuecomment-1204232963 

Another big thing, currently there is unsureness as to whether classifications, job titles or both will used in Create/Edit Pool, so for the time being I mirrored the Edit Pool with select Classification bit but left the job title stuff in for use later
Mirroring Edit Pool for the time was a decision made in Huddle

esizer (approved):
Nice, working well for me. Based on your comments regarding the key, I think it should be safe to ignore the validation for that field.

PR #3569, opened 2022-08-03
Viewport support for Storybook and Chromatic

### LINKED ISSUE #2078: Add viewport support for Chromatic
Re-ticketed from https://github.com/GCTC-NTGC/gc-digital-talent/pull/1992#issuecomment-1030336335

Blocked on https://github.com/GCTC-NTGC/gc-digital-talent/issues/2075

Chromatic can test specific components (or all) on multiple viewport sizes. This would help us :
1. test story variants with long inputs on small viewports,
2. discover regressions we might miss by not doing QA on various viewports

https://www.chromatic.com/docs/viewports

petertgiles:
This branch adds some extra support for Viewports in Storybook and Chromatic.

1) The INITIAL_VIEWPORTS collection was added as a default to each workspace and adds many standard devices to every canvas.  No configuration to individual stories needed.
image

2) The default viewport can be set for a story as demonstrated in frontend/common/src/components/form/RadioGroup/RadioGroup.stories.tsx.  The possible keys can be found at [their github page](https://github.com/storybookjs/storybook/blob/master/addons/viewport/src/defaults.ts).  Unfortunately this default viewport is not used automatically for Chromatic snapshots.
image

3) Snapshots for Chromatic can be set for a story as demonstrated in frontend/indigenousapprenticeship/src/js/components/Home/Home.stories.tsx.  I've added the `widthOf` helper to this branch to help link snapshots widths to a real viewport instead of using magic numbers.  Unfortunately there's no way I know of to send a label on to the Chromatic UI.
image

Closes #2078


frontend/common/src/components/form/RadioGroup/RadioGroup.stories.tsx
```diff
@@ -89,7 +81,7 @@ RadioGroupOfElements.args = {
 export const LargeRadioGroup = TemplateRadioGroup.bind({});
 LargeRadioGroup.parameters = {
   viewport: {
-    defaultViewport: "mobile1",
+    defaultViewport: "iphone6p",
```
petertgiles, line 84: I tried so hard but could not figure out a way to type annotate the keys of INITIAL_VIEWPORTS and improve type safety.  For example, `keyof typeof INITIAL_VIEWPORTS` or `[key in INITIAL_VIEWPORTS]: Viewport`.  :shrug:

frontend/common/src/helpers/storybookUtils.ts
```diff
@@ -0,0 +1,36 @@
+import {
+  Viewport,
+  ViewportStyles,
+  Styles,
+  // eslint-disable-next-line import/no-unresolved
```
petertgiles, line 5: This is so bad!  I haven't been able to figure out how to avoid this override.  Chromatic clearly doesn't care.

vd1992:
I am not entirely sure what to test or judge here
Can confirm if you go to the published storybooks then select the "Large Radio Group" story, it has a different viewport as default

esizer (approved):
Really unfortunate that an essential package does not integrate easily with a product made by core contributors of said technology 😢 

That being said, the fact you got this working is amazing! 😁

PR #3575, opened 2022-08-04
Reduce logout logic complexity

### LINKED ISSUE #3012: Have frontend always try to end the session on logout.
Right now our frontend conditionally tries to end the session by visiting the end_session endpoint.  The condition is there because Laravel auth did not have this endpoint but our new mock auth does.

UAC:
- [ ] Remove conditional logic so that the frontend always tries to end the session
- [ ] Setup example environment with the mock-auth end_session endpoint
- [ ] Cypress test to confirm that session is ended on logout and user is redirected to the right place?
- [ ] Use `.well-known/openid-configuration` endpoint to find end_session endpoint instead of explicitly listing it?  It's a bit weird since all the other auth stuff is being our API server.

petertgiles:
We can't make logout fully mandatory as in the AC of the issue.  If the browser is redirected to the Sign In Canada logout URI when there is no active session then a scary error page is displayed.  The logic to guess if a session is still active must remain.  This branch reduces the logic complexity around logout a bit.

## Highlights
- Removed some leftover Laravel Auth logic from api/app/Http/Controllers/AuthController.php.
- Removed the `homePath` prop from AuthenticationContainer.  Browswer will always be sent to `logoutRedirectUri`.
- Added a `getRuntimeVariableNotNull` function that will log an error if a mandatory variable is missing and has a not-null return type.
- Refactored the logic of `logoutAndRefreshPage` to (hopefully) make it a bit clearer

## Testing
- Test with both SiC and MockAuth
- Log in
- Check when your token expires (try jwt.io)
- If you logout before your token expires then you should see your browser navigating to the logout URI
- If you logout after your token expires then you should see your browser directly navigating to the /logged-out page
- If you have no OAUTH_LOGOUT_URI variable set in your PHP container you should see a console error logged when you try to logout

## Limitations
- Chose note to write Cypress test since behavior requires waiting several minutes and (as Patcon mentioned) mock-auth is stateless
- Chose to not use .well-known/openid-configuration endpoint to find end_session endpoint since that variable would be the only use on the frontend.  This probably will make more sense when we switch to frontend auth with PKCE

Closes #3012

### DIFF
--- a/frontend/common/src/helpers/runtimeVariable.ts
+++ b/frontend/common/src/helpers/runtimeVariable.ts
@@ -5,12 +5,26 @@ interface HasServerConfig {
 }
 
 /**
- * A function to retrieve a environment variable value from the window object
+ * Retrieve an environment variable value from the window object
  */
 export const getRuntimeVariable = (name: string): string | undefined => {
   const windowWithConfig = window as unknown as HasServerConfig;
   // eslint-disable-next-line no-underscore-dangle
   return windowWithConfig.__SERVER_CONFIG__?.get(name);
+};
+
+/**
+ * Retrieve an environment variable value from the window object with error if missing
+ */
+export const getRuntimeVariableNotNull = (name: string): string => {
+  const runtimeVariable = getRuntimeVariable(name);
+  if (runtimeVariable) {
+    return runtimeVariable;
+  }
+
+  // eslint-disable-next-line no-console
+  console.error(`Missing mandatory runtime variable: ${name}`);
+  return " ";
 };
 
 /**
--- a/api/app/Http/Controllers/AuthController.php
+++ b/api/app/Http/Controllers/AuthController.php
@@ -31,10 +31,7 @@ class AuthController extends Controller
         else
             $ui_locales = $requestedLocale;
 
-        $scope = 'openid';
-        // Laravel auth server will error out if you request offline_access scope
-        if(config('oauth.authorize_uri') != 'http://localhost:8000/oauth/authorize')
-            $scope = $scope . ' offline_access';
+        $scope = 'openid offline_access';
 
         $query = http_build_query([
             'client_id' => config('oauth.client_id'),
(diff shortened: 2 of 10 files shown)


yonikid15 (changes requested):
Excellent work Peter! The logout process overall seems to be working still, and you were able to clean up some stale code which is great. I found a couple possible small issues:

- If you logout after your token expires then you should see your browser directly navigating to the /logged-out page
    - This works for the SiC logout flow, but I'm still seeing the browser navigate to the logout URI with the MockAuth. 
- I also found the error log wasn't showing up which I marked below

frontend/common/src/helpers/runtimeVariable.ts
```diff
@@ -5,14 +5,28 @@ interface HasServerConfig {
...
+ * Retrieve an environment variable value from the window object with error if missing
+ */
+export const getRuntimeVariableNotNull = (name: string): string => {
+  const runtimeVariable = getRuntimeVariable(name);
+  if (runtimeVariable) {
+    return runtimeVariable;
+  }
+
+  // eslint-disable-next-line no-console
+  console.error(`Missing mandatory runtime variable: ${name}`);
```
yonikid15, line 26: Weird, I'm not seeing this error popup due to `const runtimeVariable = getRuntimeVariable(name);` constant returning "(none)" instead of an empty string, which is technically `true`. You would think it would return `undefined`, since it's a Map. Double checking again.

yonikid15 (changes requested):

frontend/common/src/helpers/runtimeVariable.ts
yonikid15: image

petertgiles: Good catch!  I must have tested with admin but the filterEmpty function was missing on ts and iap.  Added in https://github.com/GCTC-NTGC/gc-digital-talent/pull/3575/commits/ac69b7064982860ce7a07a326d36a8dfa6a779c3 .
image

petertgiles: To test the fix you'll need to rerun webpack build to copy the sjs file from `public` to `dist` (or manually copy) and probably rebuild the php container.

yonikid15 (approved):
Everything looks great! I'll approve and leave one comment, which might be out of the scope of this issue.. When logging out of the site, with the OAUTH_LOGOUT_URI equal to `null`, the user is redirected back to the login URI causing them to get logged back in (in the case of SiC). Just want to make sure this is what we are expecting to happen.

petertgiles:
> When logging out of the site, with the OAUTH_LOGOUT_URI equal to `null`, the user is redirected back to the login URI causing them to get logged back in

Yeah, now that the code requires that variable and will show an error if it's null, the behavior becomes a bit strange if you ignore the error and try to use the site anyway.  In this case, what's happening is that it deletes the cookies and tries to navigate to nothing which the browser interprets as the same current page which is probably protected and will try to log you back in.

PR #3578, opened 2022-08-04
Stabilize Missing Skills Stories

petertgiles:
I noticed that something is randomizing the order the fake skills in the MissingSkills component stories.  I added a sort to stabilize them for snapshotting.

Testing - Run storybook two or more times and check if the order of the fake skills in the MissingSkills stories is stable.


vd1992 (approved):
BEFORE
Chips would be in different order

AFTER
Chips in the same order consistently 

Nice!

image

vd1992:
stick on an estimate to the PR itself before merging

PR #3610, opened 2022-08-08
Bug - Revert storybook/react

esizer:
## Summary

A dependabot merge I made earlier today #3592 broke out chromatic builds. Seems to be because of a `react-dom` hook introduced to bridge compatibility between react 17 and 18. Should be removed once storybook fully moves to react 18 and drops support for 17.


mnigh:
> Seems to be because of a `react-dom` hook introduced to bridge compatibility between react 17 and 18. Should be removed once storybook fully moves to react 18 and drops support for 17.

Is there an existing issue already for this in [their repo's issues](https://github.com/storybookjs/storybook/issues)? I couldn't find one that mirrors what you are describing but might be helpful if there is one already to reference it here.

esizer:
> Is there an existing issue already for this in [their repo's issues](https://github.com/storybookjs/storybook/issues)? I couldn't find one that mirrors what you are describing but might be helpful if there is one already to reference it here.

To be honest, I am not 100% sure this is the exact issue. I looked into it and this seems to exist in the previous version as well. 

I looked at every commit and can't understand why this is breaking our build 🤷

vd1992 (approved):
given it is just undo-ing something, I reviewed by just glancing at the files changed

PR #3613, opened 2022-08-08
Limit when Azure pipeline builds a new release artifact

tristan-orourke:
Only trigger azure build on changes to main, not every PR. Exclude changes to only documentation or maintenance scripts.

Main motivation:
Since we automatically deploy to the Dev server every time an artifact is built, and we're currently building an artifact every time a PR is created or updated (and not necessarily ready for merging into main), the Dev server is likely to be broken at any given time.


petertgiles (approved):
Not sure how to test this without actually applying it as our build pipeline but the logic looks solid.

PR #3614, opened 2022-08-08
Post deployment script

### LINKED ISSUE #2680: Errors during build/deploy on Azure should notify devs
First, errors during build scripts should cause the task on Azure to fail. This can be accomplished by including `set -e` in the build script.

Ideally, a failure during the build process could notify our slack bot.

The command that runs after the artifact is deployed (including artisan migrate) should also cause some sort of notification if it fails.

petertgiles:
This branch adds a post-deployment script which reports the status of each step to a Slack webhook which posts in devrobotfam.

You can see the webhook in the Slack API section:
image

This is what a failed post-deployment looks like:
image

Deployment notes:
- Currently the post-deployment steps are executed in the app service *Startup Command*
- Change the start command from `cd api/ && php artisan migrate -n --force && cd ../admin && mkdir -p /tmp/bootstrap/cache && chown www-data:www-data /tmp/bootstrap/cache` to `./infrastructure/bin/post_deployment.sh . https://hooks.slack.com/services/.... appsvc-dev-talentcloud-lin001`
  - The first argument (.) is the project root and should be the same for each vertical
  - The second argument contains a slack secret and should not be posted publicly.  Look it up or ask me for it.  Should be the same for each vertical
  - The last argument is the name of the source that will be shown in the message to indicate the vertical or app service.

Testing instructions:
- Try setting it up on the dev app service using the deployment instructions above.
- To force a failure try deploying commit b07d65f4f21ac856a0d016fde3700a3a2cfd7c69

Observation:
- devrobotfam is very noisy.  I could create an issue to remove Zenhub and Github from posting there.

Closes #2680


tristan-orourke (changes requested):
I love it! I changed to Startup Command for the Dev server as described, ran the build pipeline off this branch, and when it automatically deployed to Dev the expected messages appeared in devrobotfam! Beautiful!

I have just one suggestion: if migrations fail, add $MIGRATION_STDOUT to the slack message. (Maybe as a code block, if that's possible.) On UAT and PROD, the startup command is `mkdir -p /tmp/bootstrap/cache && chown -R www-data:www-data /tmp/bootstrap && cd api && php artisan migrate -n --force 2>&1 | tee /tmp/artisan-migrate.log` which logs the output, allowing us to see why it failed, if it failed.

I'm actually wondering... why *do* migrations seem to fail when deploying this PR??

tristan-orourke:
Also, +1 to #3646!

tristan-orourke:
Oh! One other suggestion. If migrations do fail, maybe throw an `@channel` in the slack message? Even if we make devrobotfam less noisy overall, messages about failures should inherently be more noisy than messages about successes.

tristan-orourke (approved):
Looks great! I love how you've organized this file as well.

tristan-orourke:
So, as part of the next deployment, we have to make sure to change the Start Command of UAT and Prod, and the startup command in the IaC.

petertgiles:
> why _do_ migrations seem to fail when deploying this PR??

I think I copied an old version of the startup command that had the migration before the cache directory setup.  I moved it around in the latest commit and it looks much better. :smile: 
image

Issue #3588, opened 2022-08-06
Field Required Error Message Broken


vd1992:
## Description
going to /edit on a pool throws up a warning
```
Error: [@formatjs/intl Error FORMAT_ERROR] Error formatting default message for: "IfCvm5", rendering default message verbatim
MessageID: IfCvm5
Default Message: {fieldLabel}: This field is required.
Description: undefined

Locale: en

The intl string context variable "fieldLabel" was not provided to the string "{fieldLabel}: This field is required."
```

## Screenshot: Prototype

![image](https://user-images.githubusercontent.com/40485260/183225274-dc4f1ffe-2bc8-416e-8e87-4bb07e0cac02.png)

## Acceptance Criteria
- [ ] fix the cause of error

   [image]: https://user-images.githubusercontent.com/305339/166570935-6425a615-82c7-4b2f-9840-5c7d698fa2ab.png
   [image-link]: https://example.com
   [figma]: https://go.talent.c4nada.ca/figma
   [sitemap]: https://go.talent.c4nada.ca/sitemap

petertgiles:
Observed on the Search Request page:
![image.png](https://images.zenhubusercontent.com/6172cf6e898e11fd20507433/07b99a15-37e1-4368-a958-03fe1f7d2568)

petertgiles:
I'm guessing related to #3536

gobyrne:
The cause has been removed. We have a better fix for custom error messages coming in later.

PR #3630, opened 2022-08-08
Bug - Fix Dep Updates

esizer:
## Summary

Looks like there were a few breaking changes that were not properly flagged in release notes for some packages. This resolves all outstanding bugs/deprecations related to the dependabot updates.

## What was done

- Temporarily forced storybook dependencies to `6.5.9`
- Temporarily set `react-hook-form` to only patch releases
- Removed `{fieldLabel}` from defaulted field error message (causing issues with `react-intl` and failing tests)
- Fixed deprecated `faker.company.companyName()` method

## New Issues

- https://github.com/GCTC-NTGC/gc-digital-talent/issues/3633
- https://github.com/GCTC-NTGC/gc-digital-talent/issues/3634


petertgiles (approved):
You're my hero!  Seems to fix everything, even #3588!

petertgiles:
Chromatic can't seem to do a compare but that's probably due to main being broken right now.

PR #3636, opened 2022-08-09
Tooling - 3496 - Complete profile storybook

### LINKED ISSUE #3496: Setup for Accessibility testing (storybook for starters)
Gray and patcon felt that storybook was the most appropriate path forward in the short term.

Timeline would be for this work to happen either in the next sprint, or this sprint (PPPRE) if someone finds time.

There were a number of other proposals (opening UAT up to public with or without disabling auth, heroku or other external host) but Gray and patcon both thought they'd present challenges in short-term. Please do chime in if you think differently :)

### Proposed To Dos

- [x] inventory which pages of profile have stories and which don't
- [x] share links with Gray for Fable testing

esizer:
Resolves #3496 

## Summary

This makes the Diversity, Equity and Inclusion story interactive so it can be tested by Fable.


petertgiles (approved):
Nice.  Works for me.

PR #3650, opened 2022-08-10
Fix null values on FilterBlock component

yonikid15:


vd1992 (approved):
think the small problem is resolved?

no skills selected, classification selected

image

skills selected, no classification selected 

image

PR #3657, opened 2022-08-11
Bug - 3512 - Fix empty unique user validation

### LINKED ISSUE #3512: Empty optional fields should not violate uniqueness validation rules
## Description
Currently, sub is an optional field, with no input it passes in `null`. 
Validation also exists for sub, so attempting to make another user with no sub would be trying to make a second user with sub as `null` which fails validation and would be confusing to understand from a user-perspective. 

This section is probably still a bit in flux so this will be something to keep in mind. 
If sub validation exists, then it is misleading to label it as optional. 
Part of this would also involve more user friendly error messages (toast that _____ field must be unique) which may already be an issue. 

Email too is unique and optional and therefore would cause the same issue. 

## Screenshot: 
image

image

## Acceptance Criteria
- [ ] multiple users can be created with empty email and sub fields

esizer:
Resolves #3512 

## Summary

This adds the `sometimes` validation rule to `sub` and `email` when creating a user.

### Before Each

2. Navigate to `/admin/users/create`

### Empty Email/Subject (Success)

1. Attempt to create a user with no Email or Subject
2. Confirm it submits

### Duplicate Email/Subject (Failure)

1. Create a user with known Email and Subject
2. Attempt to create second user with same Email and Subject
3. Confirm it fails

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -736,13 +736,13 @@ input PoolsHasMany {
 When creating a User, name is required.
 """
 input CreateUserInput {
-    sub: String @rules(apply: ["unique:users,sub"])
+    sub: String @rules(apply: ["sometimes","unique:users,sub"])
     roles: [Role] = []
 
     # Personal info
     firstName: String! @rename(attribute: "first_name")
     lastName: String! @rename(attribute: "last_name")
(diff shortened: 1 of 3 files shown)


yonikid15 (approved):
Nice work! Everything is working on my end 👍

Issue #3660, opened 2022-08-11
Add new statuses to PoolCandidate (simple version)


tristan-orourke:
## Description
This is a simpler alternative to #3659.

Add new statuses to PoolCandidateStatus enum. We need one status for each state in the following table:
### Status table

| Status | Copy on public (My Applications) |  Programmatically | Has Snapshot | Can be Archived | Can be Deleted | Search results |
| ------- | -------------------- | -------------------- | ---------------- | ------------------- | ------------------ | ------------------ |
| Draft  | "Closes in (time) days" | Yes - default status on application creation| no | no | YES |
| Draft-expired  | "Closed on (date)" | Yes - default status on a draft application past closed date of advertisement | no | no| YES |
| New application  | "Submitted and waiting for review"|  Yes - default when submitting | YES | no | no |
| Application under review  | "Application under review" | manual | YES | no | no |
| Screened in (available)  | "Waiting for assessment" | manual| YES | no | no |
| Screened out - On application| "Application not accepted" | manual| YES | YES | no |
| Under assessment | "Waiting for assessment" | manual| YES | no | no |
| Screened out - On assessment  | "Assessment not passed" | manual| YES | YES | no |
| Qualified - Available  | "Qualified" | manual | no | no| no |
| Qualified - Unavailable  | "Qualified" | manual | no | no| no |
| Qualified - Withdrew  | "Qualified" | manual | no | no| no |
| Placed Casual  | "Temporary placement" | manual | no | no| no |
| Placed Term  | "Completed" | manual | no | no| no |
| Placed Indeterminate  | "Completed" | manual | no | no| no |
| Expired  | "Expired on (date)" | YES - Rule goes here | no | YES | no |

While the PoolCandidate table does have an underlying status column, the model should have a status accessor which adds a bit of logic to the return value:
1. If the PoolCandidate.submitted_at date is null, always return DRAFT or DRAFT_EXPIRED
2. Otherwise, if the expiry_date is passed, and the status is not one of the PLACED/Completed statuses, always return EXPIRED 

## Acceptance Criteria
- [x] PoolCandidateEnum contains an enum for each status in the table
- [x] Status can be manually set to any of the statuses in the table
- [x] If the PoolCandidate.submitted_at date is null, API always returns DRAFT or DRAFT_EXPIRED
- [x] If submitted_at is non-null, and status is NOT Placed Casual, Placed Term, or Placed Indeterminate, then API always returns EXPIRED

[...]

PR #3662, opened 2022-08-11
create application mutation

### LINKED ISSUE #3332: Backend mutations for Creating Applications
# Description

## Create Draft mutation
Create a mutation which creates a stub application, aka a PoolCandidate, for the currently logged-in user and the specified pool. This PoolCandidate object can be devoid of any data besides the pool, the user, and something that marks it as a Draft (a new status?). 

QUESTION: if the user already has a PoolCandidate for the pool, do we want to throw an error or just return the pool that already exists?

ANSWER: (Gray) I think so... it seems like returning the existing one will solve more problems then it causes. So yes, unless there's a good reason not to.

# Acceptance Criteria
- [ ] Create mutation creates a new PoolCandidate object with very little data, but in a draft status
- [ ] Draft status simply means that the 'submitted_at' field is null (potentially)
- [ ] If user already has a PoolCandidate for the pool, create mutation returns that existing object instead of creating a new one.

vd1992:
resolves #3332 

testing: 
use this mutation
you will need to fill in values from adminer, if you attempt to create a duplicate application the row count should remain unchanged, and should increment by one when you attempt to make a new one
```
mutation createApplication {
  createApplication(userId: "USER_ID", poolId: "POOL_ID") {
    user {
      id
    }
    pool {
      id
    }
  }
}
```

status work I did not complete as #3660 deals with that, adding DRAFT as well as doing the status logic
so what I did was just implement the submitted_at field for that issue

### DIFF
--- a/api/app/GraphQL/Mutations/CreateApplication.php
+++ b/api/app/GraphQL/Mutations/CreateApplication.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace App\GraphQL\Mutations;
+use App\Models\PoolCandidate;
+
+final class CreateApplication
+{
+    /**
+     * Create an application
+     * @param  null  $_
+     * @param  array{}  $args
+     */
+    public function __invoke($_, array $args)
+    {
+        // attempt to find existing application, if found return that otherwise create new application
+        $application = PoolCandidate::where('user_id', $args['userId'])->where('pool_id', $args['poolId'])->first();
+        if($application) {
+            return $application;
+        }
+        $newApplication = new PoolCandidate;
+        $newApplication->user_id = $args['userId'];
+        $newApplication->pool_id = $args['poolId'];
+        $newApplication->save();
+        return $newApplication;
+    }
+}
--- a/api/app/GraphQL/Mutations/CreateApplication.php
+++ b/api/app/GraphQL/Mutations/CreateApplication.php
@@ -13,14 +13,10 @@ final class CreateApplication
     public function __invoke($_, array $args)
     {
         // attempt to find existing application, if found return that otherwise create new application
-        $application = PoolCandidate::where('user_id', $args['userId'])->where('pool_id', $args['poolId'])->first();
-        if($application) {
-            return $application;
-        }
-        $newApplication = new PoolCandidate;
-        $newApplication->user_id = $args['userId'];
-        $newApplication->pool_id = $args['poolId'];
-        $newApplication->save();
-        return $newApplication;
+        $application = PoolCandidate::firstOrCreate([
+            'user_id' => $args['userId'],
+            'pool_id' => $args['poolId']
+          ]);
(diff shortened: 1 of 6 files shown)


api/app/GraphQL/Mutations/CreateApplication.php
```diff
@@ -0,0 +1,26 @@
...
+        // attempt to find existing application, if found return that otherwise create new application
+        $application = PoolCandidate::where('user_id', $args['userId'])->where('pool_id', $args['poolId'])->first();
+        if($application) {
+            return $application;
+        }
+        $newApplication = new PoolCandidate;
+        $newApplication->user_id = $args['userId'];
+        $newApplication->pool_id = $args['poolId'];
+        $newApplication->save();
+        return $newApplication;
```
esizer, line 24: If you set these as fillable, you could convert this to:

```php
$application = PoolCandidate::firstOrCreate([
  'user_id' => $args['userId'],
  'pool_id' => $args['poolId']
]);
```

api/database/migrations/2022_08_11_161706_submitted_at_field_added_application.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('pool_candidates', function (Blueprint $table) {
+            $table->timestamp('submitted_at')->nullable();
```
esizer, line 17: Should this have a default of `null` so we don't accidentally create a pool candidate with this set? Or is it implied it is `null`? 

This is more of a question than a suggestion because I'm not sure 🤷

esizer (approved):
Nice work! 👍 Just a small suggestion and a question from me so I can learn a bit more :)

api/database/migrations/2022_08_11_161706_submitted_at_field_added_application.php
vd1992: It shouldn't be accidentally set anywhere, just empty/null I think
hm, let us be cautious 
@petertgiles please enlighten us

petertgiles: If it's nullable and there's no default, then yes it will be null for new rows.  That said, I'm all for making implied things more explicit.

vd1992: and done
https://github.com/GCTC-NTGC/gc-digital-talent/pull/3662/commits/f86eef5e4d07ed28fc731a03b0731943aeea7278

vd1992:
@esizer 
I made some changes, so I think it would be best if you please quickly glanced over it before I merge

esizer (approved):
This has my stamp of approval 🎖️

PR #3669, opened 2022-08-12
Feature - 3326 - Add missing skills experience page

### LINKED ISSUE #3326: Edit Experience & Skills page with Required Skills
# Description

[Figma Prototype](https://www.figma.com/proto/N9NoPunkqaynjsi89Wrbh0/Direct-Intake---Poster-and-application?node-id=138%3A2829&starting-point-node-id=226%3A4567&scaling=min-zoom)

Create a version of the Experience and Skills page which also tells you which skills you are missing (considering the pool you are applying to). Transferable skills are not considered required and should not be listed.

Note: The Transferable/Occupational category comes from the Skill Family, which skills may have several of. If a skill has zero skill families, list it here. Only  Consider a skill to be Transferable (and therefore unlisted) if it has at least one SkillFamily and all are Transferable.

image

# Acceptance Criteria

- [x] this version of the Experience and Skills page is rendered if the url has an applicationId query parameter
- [ ] when creating or editing an Experience from this page, the url and breadcrumbs are relative this one, and saving returns you here
- [x] All required and Nice to Have non-transferrable skills are listed. 
- [x] Transferable skills associated with the pool are not listed.

esizer:
Resolves #3326 

## Summary

This adds the `<MissingSkills />` component to the `<ExperienceAndSkills />` component when `application` is set in the query params.

## Note

There is an unused variable that will need to be used to construct the paths to return to the application once that route has been created.

## Screenshot 

Screen Shot 2022-08-12 at 3 00 26 PM

## Testing

1. Get a `PoolCandidate` `id` from the database
2. Navigate to `/talent/profile/skills-and-experiences`
3. Confirm there is no `<MissingSkills />` at the top of the page
4. Append `?application={id}` where `id` is the one retrieved in step 1
5. Confirm there is a `<MissingSkills />`
    - Note: There is a small chance it may not appear since you may have added all of the skills


api/graphql/schema.graphql
```diff
@@ -306,6 +306,7 @@ enum PoolAdvertisementLanguage {
 type PoolCandidate {
     id: ID!
     pool: Pool @belongsTo
+    poolAdvertisement: PoolAdvertisement @belongsTo(relation: "pool")
```
esizer, line 309: I'm not sure if this is correct but not sure how else to get the `essentialSkills` and `nonessentialSkills`

vd1992:
I am a little bit confused
As is, doesn't it let a user edit/view any application's Edit Skills and Experiences page if they have the `PoolCandidate id` to plug into the `application={id}`? Shouldn't it check that the `id` of the logged in user is the same as the `user_id` connected to the `PoolCandidate` instance?

vd1992:
error upon navigating to /profile/skills-and-experiences

image

esizer:
> I may be misunderstanding but that seems like a bad thing to allow.

This is a very bad thing to allow, thought we do have a policy to prevent this. If you login with admin@test.com, you will see all of them.  Though, I have updated this to fix a potential problem where applicants could not view their own application in [allow users to view their application](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3669/commits/ab46324d8a5e5f62f0604d326c73e60c71f371a6)

https://github.com/GCTC-NTGC/gc-digital-talent/blob/28e66ef2b61d5ad93016d862565b939791bf49c1/api/app/Policies/PoolCandidatePolicy.php#L24-L34

> error upon navigating to /profile/skills-and-experiences

I created an API wrapper to prevent this error in [fix error when no application id is provided](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3669/commits/34e45b5aa72299541079132b18298cd57468d9d9)

api/graphql/schema.graphql
yonikid15: I think this is correct. Even though pool and poolAdvertisement are technically fetching from the same model, this makes it more explicit. Also, I believe the Pool type doesn't have `essentialSkills` and `nonEssentialSkills`.

esizer: Thanks, I was a little unsure about this and wanted to confirm I wasn't doing something incorrect.

yonikid15 (commented):
Great work Eric! Everything's looking great. I'm seeing the correct skills are getting displayed. However, I'm looking at the acceptance criteria, and it mentions: 

**"when creating or editing an Experience from this page, the url and breadcrumbs are relative this one, and saving returns you here"**

I'm not seeing the url or breadcrumbs staying consistent with the previous url the user came from. Was that still part of the scope of this issue?

esizer:
> I'm not seeing the url or breadcrumbs staying consistent with the previous url the user came from. Was that still part of the scope of this issue?

Oops! Yeah, I'll work on getting that in before re-requesting a review. Thanks for catching that!

yonikid15 (changes requested):
Really nice work with the breadcrumbs and url! Just one last minor thing left and should be good to go!

frontend/talentsearch/src/js/components/experienceForm/ExperienceForm.tsx
```diff
@@ -68,15 +73,21 @@ export const ExperienceForm: React.FunctionComponent<ExperienceFormProps> = ({
...
   const locale = getLocale(intl);
   const paths = applicantProfileRoutes(locale);
+  const directIntakePaths = useDirectIntakeRoutes();
   const defaultValues = experience
     ? queryResultToDefaultValues(experienceType, experience)
     : undefined;
+  const { application } = useQueryParams();
+  const returnPath = `${paths.skillsAndExperiences(userId)}${
+    application && `?application=${application}`
+  }`;
```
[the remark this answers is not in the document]
esizer: Good catch! Fixed with [fix undefined being appended to url](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3669/commits/79163e0e4826bb558ab7b46dea49d43bd043e215)

yonikid15 (approved):

frontend/talentsearch/src/js/components/applicantProfile/ProfileFormWrapper.tsx
```diff
@@ -20,20 +21,24 @@ const ProfileFormWrapper: React.FunctionComponent<ProfileFormWrapperProps> = ({
...
-      href: profilePath.myProfile(),
-      icon: <UserIcon style={{ width: "1rem", marginRight: "5px" }} />,
-    },
-    ...crumbs,
-  ];
+  let links = [...crumbs];
+  if (prefixBreadcrumbs) {
+    links = [
+      {
+        title: intl.formatMessage({
```
mnigh, line 32:
```suggestion
          title: intl.formatMessage({
```

frontend/talentsearch/src/js/components/applicantProfile/ProfileFormWrapper.tsx
```diff
@@ -20,20 +21,24 @@ const ProfileFormWrapper: React.FunctionComponent<ProfileFormWrapperProps> = ({
...
+          defaultMessage: "My Profile",
+          description: "Breadcrumb from applicant profile wrapper.",
+        }),
+        href: profilePath.myProfile(),
+        icon: <UserIcon style={{ width: "1rem", marginRight: "5px" }} />,
```
mnigh, line 37: The `icon` key is no longer used in the [Breadcrumbs component](https://github.com/GCTC-NTGC/gc-digital-talent/commit/fc457b19722bf41493d4303eb4405233064ed4e1), though strangely, still [part of the props](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/frontend/common/src/components/Breadcrumbs/Breadcrumbs.tsx#L6).

mnigh (changes requested):
Nice work! A few small things to address before approval.

frontend/common/src/fakeData/fakeExperiences.ts
```diff
@@ -16,6 +16,7 @@ import {
   AwardedScope,
   EducationType,
   EducationStatus,
+  Experience,
```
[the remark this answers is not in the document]
esizer: Fixed with [fix lint errors](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3669/commits/80e7d277ad2689e1b50bf8db5d796545b87abfcd)

frontend/talentsearch/src/js/components/applicantProfile/ProfileFormWrapper.tsx
esizer: Removed in [remove icons from breadcrumbs](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3669/commits/c62df89324658bcdbeb127504a564c6ce20e83db)

mnigh: The icons still remain in some components and stories for Breadcrumbs:

Screen Shot 2022-08-24 at 12 26 44

esizer: Removed all in [remove icons from breadcrumbs](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3669/commits/dba59efa6059435cf6640f3def0f5fb8082c524d)

mnigh (approved):
🎊

PR #3675, opened 2022-08-15
Add azure  Application Insights instrumentation to IAP pages

### LINKED ISSUE #3219: Set up Analytics for IAP
# Description

Upgrade the existing GA account from v3 to v4 and apply to the app as well as TC-report.
Share analytics about the IAP page with IAP team.

## Background.

The IAP team has requested the ability to track page views. 

First, we should check if Azure already provides traffic analytics from which we could create reports for them.

If that is not possible, we should [set up a Google Analytics account](https://support.google.com/analytics/answer/10759417). From a cursory look at the [account structure](https://support.google.com/analytics/answer/9679158), we may want to create a separate "property" for IAP.

We will want to configure this to collect minimal data, and then add the necessary script to the IAP landing page.

In the future, we will likely want to set this up on all our pages (probably under different "properties" if using Google Analytics), but we can start with a minimal implementation, just on the IAP homepage.

brindasasi:
Resolves #3219.

### How to Test Local
1. Check out this branch
2. Add `"InstrumentationKey=8f4f1fd8-8831-4a3b-84a0-20245c86e418;IngestionEndpoint=https://canadacentral-1.in.applicationinsights.azure.com/;LiveEndpoint=https://canadacentral.livediagnostics.monitor.azure.com/"` as the value for `APPLICATIONINSIGHTS_CONNECTION_STRING` in .env 
3. Open the network tab in your browser
4. Navigate to '/en/indigenous-it-apprentice'
5. Search for **track** in the network tab and verify the headers and payload

### To Do
- [x] Investigate better place to store connection string (secrets?) 
- [x] Deploy in the dev environment so we can see the metrics ( created a jira for team access to this AppInsights resource in Azure https://jira.tbs-sct.gc.ca/browse/AZ-13533 ) 
- [ ] Could we split the instrumentation as a sep wrapper from page container in IAP? So we can reuse for talencloud folder as well. ( for team/code reviewers to comment on. )  

### Azure Metrics screenshot 
image

**Deployment Notes**

### Dev & UAT 

Add the below environment variable
 APPLICATIONINSIGHTS_CONNECTION_STRING=""  

( empty string as we don't want to collect metrics from these environments ) 

### Prod 

Prod needs to be given the actual connection string from Azure AppInsights. You can get it directly from @brindasasi OR 
By this time deployment of this story all devs would have those access in Azure. 
Here is how you can collect the connection string 
migrate-from-instrumentation-keys-to-connection-strings (1)
https://docs.microsoft.com/en-us/azure/azure-monitor/app/sdk-connection-string?tabs=net

 APPLICATIONINSIGHTS_CONNECTION_STRING= "[original connection string from azure] "


mnigh:
>  Investigate better place to store connection string (secrets?)

I think it's fine for the connection string to be an environment variable, it's purposefully user-facing similar to how Google Analytics exposes the tracking id. Also, putting it in GitHub secrets wouldn't allow us to access the value on remote servers (Local, UAT, Dev, Production).

>  Deploy in the dev environment so we can see the metrics

We should be able to see the metrics without having to deploy to the dev environment.

>  Could we split the instrumentation as a sep wrapper from page container in IAP? So we can reuse for talencloud folder as well.

We could make a reusable component and put it in /common I think and then call the component in each "project" (admin, talentsearch, indigenousapprenticeship). Whether or not we want a different resource (with a different connection string) is another question.

petertgiles:
I added a deployment label to the issue.  Adding an env var will take an extra step during the next deployment.

petertgiles:
It seems like the "AC" from the client was detecting where the page views are coming from.  Could you demonstrate that?

brindasasi:
> I added a deployment label to the issue. Adding an env var will take an extra step during the next deployment. Please add a deployment note to this PR.

yeah ..thanks for the reminder.  Adding it

brindasasi:
> It seems like the "AC" from the client was detecting where the page views are coming from. Could you demonstrate that? Also, is it possible for others to view a dashboard or have one mailed to them automatically?

Yeah we can see the page views more granular like location , device , browser. Can do a group review where I can demo that.
We don't have a dashboard yet as I don't have permission to create one. ( asked for it ) 
But all devs should be able to access the AppInsights transaction logs & page views as above once the permission is given. Will update once that jira is resolved(https://jira.tbs-sct.gc.ca/browse/AZ-13533).

frontend/indigenousapprenticeship/package.json
```diff
@@ -109,8 +109,10 @@
     "webpack-merge": "^5.8.0"
   },
   "dependencies": {
-    "@heroicons/react": "^2.0.10",
-    "core-js": "^3.25.0",
+    "@heroicons/react": "^1.0.6",
```
petertgiles, line 112: Careful, we don't want to roll back these two dependencies.  Probably a merge-main artifact.

frontend/indigenousapprenticeship/.env.example
```diff
@@ -2,3 +2,4 @@ INDIGENOUSAPPRENTICESHIP_APP_URL="http://localhost:8000/indigenous-it-apprentice
 INDIGENOUSAPPRENTICESHIP_APP_DIR="/indigenous-it-apprentice"
 NODE_ENV=development
 API_URI="/graphql"
+APPLICATIONINSIGHTS_CONNECTION_STRING=""
```
petertgiles, line 5: This variable does not have the "INDIGENOUSAPPRENTICESHIP" prefix.  If the connection string will be the same for the other apps then they will share the variable and this is OK.  If the others need to be different then you may want to add the prefix.

petertgiles (changes requested):
Sorry, I can't build this locally either.  Something is messed up with the package graph, I think.

frontend/indigenousapprenticeship/.env.example
brindasasi: This is going to be the same connection string for all the apps. So I leave it like that.

Issue #3322, opened 2022-07-12
Add custom error message with translation for existing unique validations in the schema.graphql file 


brindasasi:
**Problem** 

As a user I want to see  user friendly error messages translated to my preferred language (EN / FR ) 
for all the unique values in the system ( to start with )  

**A/C** 

- set up validation errors
- Create validation rule with error code
- Suggest conventions for unique, understandable error codes
- Create a frontend class to collect localized error messages keyed to error codes
- Create a helper function to take a error code and turn it into a localized error message for toasting
- Unit test with English and French validation

    
**Implementation Details** 
- [ ] Add custom error message in graphQL createUserInput mutation for 
- sub and  email field in User table,
-  key in skill_families table

- [ ] Add error message in common messages for translation 
- [ ] Use the added message that has been added in  common messages inside the API handler to bubble up the error on the front end

PR #3676, opened 2022-08-15
submit application backend

### LINKED ISSUE #3658: Backend mutations for Submitting Applications
## Submit Application mutation
Create a mutation which changes the status of a PoolCandidate from Draft to Submitted, and also saves a Signature field. Signature field will need to be added to model, schema, DB table, and Factory (in the case of non-draft PoolCandidates).

Add backend validation - this mutation should fail if the user does not have a completed profile, or does not have the required skills. It should also fail if the application is already submitted.

QUESTION: do we need to save the time of submission? In that case, maybe this shouldn't be handled with simply a new status.

ANSWER: So this is a much bigger thing that just changing status. We also need to take a full snapshot of the profile. That snapshot is definitely timestamped, so I think that will cover us here. We should probably add a reference to the snapshot though. BUT, since most of that happens in a different table, we could just change the status here if that's helpful. Mutation should FAIL if there is no snapshot - we are adding this as a acceptance criteria on the Snapshotting issue though. That should avoid this issue being blocked by it.

QUESTION: if we do add more PoolCandidateStatuses, what should they be called? PENDING? SUBMITTED?

ANSWER: Let's add Draft. Once submitted, the status should be changed to: Application Submitted - New. We will be adding a bunch of statuses to help DCMO sort applicants as well. Full list to come once we get confirmation from DCMO.

QUESTION: Two things can "expire", a candidate (once placed into a pool) is given an "expiry date". Also, you can have a draft application and the pool closes without you submitting it. So how are we changing the status for these? Will there be a process running on the server to check for expired things and then change the status of them if it finds any? Alternatively, we could determine "expired" things programmatically (by checking dates). @tristan-orourke ?

## Acceptance Criteria
- [x] Submit Application operation fails validation if:
    - [x] profile is incomplete
    - [x] profile is missing any required technical skills
    - [x] if application is already submitted
- [x] If validation passes, populate the `submitted_at` field with the current timestamp, and signature field is populated with data from mutation
- [x] This means adding a nullable `submitted_at` field to PoolCandidate table, model, schema, factory, fake js
- [x] PHPUnit test for this operation, which ensures validation works
- [x] Snapshotting not included in this ticket
- [x] Signature field

vd1992:
resolves #3658 
this one was tough 😓 

lots going on with this one, plenty to comb through
- added signature field
- remaining work for submitted_at field
- add mutation
- custom resolver with validation
- testing

the mutation to test is
```
mutation submitTest {
	submitApplication(applicationID: "POOL_CANDIDATE_ID", signature: "STRING") {
		submittedAt
	}
}
```


vd1992:
also, the specially thrown Exceptions have their messages present, just nested a bit further down
I figured validation errors may be changing in the future so I didn't copy the entire thing into the assertions

image

api/graphql/schema.graphql
```diff
@@ -287,6 +287,8 @@ enum PoolCandidateStatus {
     UNAVAILABLE
     EXPIRED
     PLACED_CASUAL
+    DRAFT
```
petertgiles, line 290: I thought we were just going to add a DRAFT status and whether or not it is expired would be programmatically calculated base on dates?

petertgiles:
> also, the specially thrown Exceptions have their messages present

Nice, it looks like they conform to the shape I'm expecting in #3322.  As for the test, as long as it's checking for some error being thrown I'm OK with it.

api/graphql/schema.graphql
```diff
@@ -1293,4 +1296,5 @@ type Mutation {
 
     archiveApplication(id:ID!): PoolCandidate @guard @can(ability: "archiveApplication", find: "id")
     createApplication(userId:ID!, poolId:ID!): PoolCandidate
+    submitApplication(applicationID: ID!, signature: String!): PoolCandidate
```
petertgiles, line 1299: Should be `applicationId` with lower-case d

petertgiles (changes requested):
Looks great!  Code is super-clean and seems to work well in my testing.  A few small things to check below.

api/graphql/schema.graphql
vd1992: no I don't think so, #3660 seems to require the addition of DRAFT_EXPIRED as a status too
I've just included the draft statuses for the sake of submitted_at logic, the stuff to handle statuses will come in that issue

api/app/GraphQL/Validators/Mutation/SubmitApplicationValidator.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace App\GraphQL\Validators\Mutation;
+use Database\Helpers\ApiEnums;
```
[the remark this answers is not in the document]
vd1992: yep 
https://github.com/GCTC-NTGC/gc-digital-talent/pull/3676/commits/0138d20c835b301d2345a9ba45fb0a91d2c158e4

api/app/GraphQL/Mutations/SubmitApplication.php
```diff
@@ -0,0 +1,110 @@
...
+                throw ValidationException::withMessages(['a required pool skill is missing from application']);
+            }
+        }
+
+        // all validation has successfully completed above, execute the core function of this resolver
+        $dateNow = Carbon::now();
+        $signature = $args['signature'];
+        $application->update([
+            'submitted_at' => $dateNow,
+            'signature' => $signature,
```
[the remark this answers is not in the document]
vd1992: sure can
https://github.com/GCTC-NTGC/gc-digital-talent/pull/3676/commits/8d11be6e6a19c1a541c44d6c921be21c37fca021

api/graphql/schema.graphql
petertgiles: Eugh, you're right but I don't like it.  :smile: We can remove the status later, I guess, if we need to.

petertgiles (approved):

Issue #3668, opened 2022-08-12
Remove IAP copy of ClientProvider


petertgiles:
The IAP has its own copy of the Client Provider instead of using the one from Common (or the one from the common ContextContainer.  Lets repoint it to a common component.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/frontend/indigenousapprenticeship/src/js/components/ClientProvider.tsx

esizer:
Closed by #4597

Issue #3330, opened 2022-07-12
🛠️ Unify error codes and add them into our translation checker


brindasasi:
> [!IMPORTANT]
> This is an old ticket so it has been re-scoped based on discussions that took place in the comments. You can look there for more context.

## ♻️ Debt/Refactor

We have a mixed approach of how we deal with errors being sent via graphql. We would like to unify it and ensure we have translations for all error messages.

## 🕵️ Details

We recently added the ability to have enums be localized on the backend and it may make sense to register possible error codes as enum values that can be localized into a human readable error message. This would automatically have the in our server side intl check.

## 🙋‍♀️ Proposed Solution

Use a localized enum with a special handler to get a localized error message.

```php
enum ApiError {
  
  use HasLocalization;

  case EXPIRY_DATE_REQUIRED
  
  public function localizedErrorMessage()
  {
     $message = $this->localize();
     
     // Need to return a flat assoc for laravel messages
     return [ ...$message['value'], ...$mesage['label'] ];
  }
}
```

## 🌎 Localization

Most messages should already be localized but we may have missed one (hence the point of this ticket :sweat_smile: ) So you may need to have some translated.

## ✅ Acceptance Criteria

- [ ] Error code pattern is unified
- [ ] All codes are localized
- [ ] Missing srings are flagged in CI

---

## Original

**Description**
Given that we have implemented #3322/ #3328 for custom error messages .
But still we have to maintain those error codes in the front end manually in order to get extracted for translations.
It is easy to forget and not to add those api error codes in the front end. 
Hence we would like to extend intl-extract tool to smart enough to find the error code from graphql file as well 

Proposed solution:
Add logic to the translation workflow to check for enum in laravel and verify that the messages exist in the message file.

 Adding missing translations is OUT OF SCOPE

**A/C**
- [ ] intl-extract tool should find the messages from graphql as well and create the english json file 
- [ ] remove all the front end code that adds , uses the error message from the front end common messages file

esizer:
Has this been solved with the updates to the error exchange?

https://github.com/GCTC-NTGC/gc-digital-talent/blob/417bf6edacbf3f374f70a4456e2291d5be4edeee/packages/client/src/components/ClientProvider/ClientProvider.tsx#L159-L181

tristan-orourke:
I think this is still relevant - though I don't know how to implement it

petertgiles:
[...]

PR #3680, opened 2022-08-16
Api error messages

### LINKED ISSUE #3322: Add custom error message with translation for existing unique validations in the schema.graphql file

petertgiles:
This branch adds localized API error messages.  Validation rules were added to the schema with error messages.  An error message class with intl messages was added to the common project.  A helper function was added to find the message with the matching string.  A convention was suggested in the comments to use the source rule name as the key in the error message object.

## Notes
- A test was added for the function to extract the error messages.  I couldn't reliably test the toasts themselves in Jest though.  I'm not sure that would be super useful anyway.
- The duplicate `ClientProvider` in talentsearch was removed.  I couldn't easily remove the one in IAP so I posted #3668.

## Limitations
-  In this implementation the messages must be manually added to a common file.  Someday it would be nice to have our intl utilities scan and find these strings automatically.  #3330
-  The static messages from the validators was added as well but not the parametric ones.  Building the parameter on the fly would not be useful since the substituted field names would not be translated.  It would be possible to add strings for each field if that is found to be useful but most validation will be happening on the front end anyway.

## Testing
- [ ] Add these strings to your common fr.json file and rerun `npm run intl-compile` and `npm run dev`
```
"l8p1el": {
    "defaultMessage": "FR This user identifier (sub) is already in use FR",
    "description": "Error message that the given user identifier is already in use."
  },
"nDaEgr": {
    "defaultMessage": "FR This email address is already in use FR",
    "description": "Error message that the given email address is already in use."
  },
```
- [ ] Trigger validation errors by attempting to add a new user with existing email and sub values (like "applicant@test.com") in the admin site
- [ ] Observe the errors in English and French
image

Closes #3322


vd1992:
- added to the intl file in `\common\src\lang\fr.json`
- ran intl-compile and dev 

if I only add in a duplicate email, no duplicate sub this happens

image

if it is a duplicate email _and_ duplicate sub, it works

image

vd1992:
checking elsewhere, it works! but could probably be improved

this one outputs the same message twice, presumably occurs twice due to location_en AND location_fr

image

vd1992:
trying with no essential skills attached

image

no essential skills attached as well as missing locations 

image

vd1992 (changes requested):
great work Peter! got the ball rolling on a very handy feature
messages will take time to build up and sort out, but good start

seems like it kinda glitches up when there is only one message to pass in, that probably ought to be fixed lol

frontend/common/src/components/ClientProvider/ClientProvider.tsx
```diff
@@ -81,10 +86,75 @@ const willAuthError = ({ authState }: { authState: AuthState | null }) => {
...
+        {localizedMessages.map((message) => (
+          <li key={message}>{message}</li>
+        ))}
+      </ul>
+    );
+  }
+
+  // if just 1, toast by itself
+  if (localizedMessages.length === 1) {
+    return <span>localizedMessages[0]</span>;
```
vd1992, line 146: I think this is the cause
should be in `{}`

petertgiles: Doh, how did I miss that?  Good catch!

vd1992 (approved):
my environment is busy with hydrogen, so I will just have faith in you 👍

Issue #3661, opened 2022-08-11
Trigger snapshot functionality during Application submission


tristan-orourke:
## Acceptance Criteria
- [ ] When application is successfully "submitted", snapshot is created and saved to the PoolCandidate object.
- [ ] Getting the profile data API should return the data with created snapshot json object in it along with the submitted field timestamp

esizer:
I'm not sure if we have a specific method on how we plan on doing this but using an [Eloquent Observer](https://laravel.com/docs/9.x/eloquent#observers) could be useful here.

petertgiles:
I think removing the "test snapshot" functionality from #3707 would be worthwhile to include in the scope of this.  I recommend bumping this up to 3 to include that.

PR #3707, opened 2022-08-18
Early Snapshot Functionality

### LINKED ISSUE #3340: Create snapshot of profile at time of Application Submission
# Description

During submission mutation, we need to snapshot the profile. For now, create a specific mutation for this, don't bother tying it to submission (which comes with validation and stuff).

In the PHP method get all the data together that we want to store into an array:
- Everything in the applicant object
- Experiences & experience/skills
- Contact info from the user object if it's not in the applicant object

Store it all as JSON in a JSONB column. We want to store it in a format/shape that matches what we'd get from the query we use to populate the Profile UI.

In fact, if possible, run the graphql query FROM the php method, and simply store the results as JSON.

Write a PHP unit test which creates an application from a profile, submits it, and confirms the json snapshot looks the same as the result of the profile query.

# Acceptance Criteria
- [ ] Create a test mutation which saves a snapshot of the profile to a jsonb column in PoolCandidate object
- [ ] snapshot json matches the result of the query we use to populate the profile view on the admin side
- [ ] phpunit test to confirm the above

petertgiles:
## Intro
This branch demonstrates early snapshot functionality.  It adds a mutation to manually trigger snapshot creation.  Tying it in with application submission will be handled in #3661.

## Includes:
- A migration to add a JSONB column to pool_candidates to store the snapshot.
- A TestPoolCandidateSnapshot mutation to manually trigger a snapshot for a specific poolCandidate
- A phpunit test to verify that the stored snapshot matches the snapshotting query

## Testing
- [ ] Here's a query to run the mutation 
```
mutation snapshot($poolCandidateId: ID!) {
  testPoolCandidateSnapshot(id: $poolCandidateId) {
    profileSnapshot
  }
}
```
- [ ] Run the phpunit test
```
php artisan test tests/Feature/SnapshotTest.php 
```

Closes #3340

### DIFF
--- a/api/app/GraphQL/Mutations/TestPoolCandidateSnapshot.graphql
+++ b/api/app/GraphQL/Mutations/TestPoolCandidateSnapshot.graphql
@@ -1,8 +1,7 @@
 # Comes from frontend/talentsearch/src/js/api/userOperations.graphql
-# Should match that file to take accurate snapshots
-
-        query getMe {
-            me {
+# Should match the getMe query from that file to take accurate snapshots
+query getProfile($userId:ID!) {
+            user(id:$userId) {
               id
               sub
               roles
--- a/api/app/GraphQL/Mutations/TestPoolCandidateSnapshot.graphql
+++ b/api/app/GraphQL/Mutations/TestPoolCandidateSnapshot.graphql
@@ -51,7 +51,6 @@ query getProfile($userId:ID!) {
                   fr
                 }
               }
-
               isWoman
               hasDisability
               isIndigenous
@@ -148,21 +147,6 @@ query getProfile($userId:ID!) {
                   endDate
                 }
               }
-              # awardExperiences {
-              #   id
-              # }
-              # communityExperiences {
-              #   id
-              # }
-              # educationExperiences {
-              #   id
-              # }
-              # personalExperiences {
-              #   id
-              # }
-              # workExperiences {
-              #   id
-              # }
               poolCandidates {
                 status
                 expiryDate
--- a/api/app/GraphQL/Mutations/TestPoolCandidateSnapshot.graphql
+++ b/api/app/GraphQL/Mutations/TestPoolCandidateSnapshot.graphql
@@ -1,0 +1,181 @@
+# Comes from frontend/talentsearch/src/js/api/userOperations.graphql
+# Should match that file to take accurate snapshots
+
+        query getMe {
+            me {
+              id
+              sub
+              roles
+              firstName
+              lastName
+              email
+              telephone
+              preferredLang
+              currentProvince
+              currentCity
+              citizenship
+              isVeteran
+              languageAbility
+              lookingForEnglish
+              lookingForFrench
+              lookingForBilingual
+              bilingualEvaluation
+              comprehensionLevel
+              writtenLevel
+              verbalLevel
+              estimatedLanguageAbility
+              isGovEmployee
+              govEmployeeType
+              interestedInLaterOrSecondment
+              department {
+                id
+                departmentNumber
+                name {
+                  en
+                  fr
+                }
+              }
+              currentClassification {
+                id
+                group
+                level
+                name {
+                  en
+                  fr
+                }
+              }
+              expectedGenericJobTitles {
+                id
+                key
+                name {
+                  en
+                  fr
+                }
+              }
+
+              isWoman
+              hasDisability
+              isIndigenous
+              isVisibleMinority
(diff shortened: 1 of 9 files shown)


api/tests/Feature/SnapshotTest.php
```diff
@@ -0,0 +1,71 @@
...
+     * @return void
+     */
+    public function testCreateSnapshot()
+    {
+        $snapshotQuery = $query = file_get_contents(base_path('app/GraphQL/Mutations/TestPoolCandidateSnapshot.graphql'), true);
+
+        // set up a user with a pool candidate
+        $userId = $this->faker->Uuid();
+        $poolCandidateId = $this->faker->Uuid();
+        print_r($userId);
```
esizer, line 34: Not a big deal, but I assume this `print_r` was just debugging and no longer needed?

esizer (changes requested):
Awesome! 😁  Everything is working well, just two very minor things

api/tests/Feature/SnapshotTest.php
petertgiles: Good catch!  Fixed in [Remove stray print_r](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3707/commits/08750701ae9d71dec0159a2a9178c5bb4d5ddb1c)

esizer (approved):

api/app/GraphQL/Mutations/TestPoolCandidateSnapshot.graphql
```diff
@@ -0,0 +1,180 @@
...
+              }
+              expectedGenericJobTitles {
+                id
+                key
+                name {
+                  en
+                  fr
+                }
+              }
+
```
mnigh, line 54:
```suggestion
```

api/app/GraphQL/Mutations/TestPoolCandidateSnapshot.graphql
```diff
@@ -0,0 +1,180 @@
...
+                }
+                ... on WorkExperience {
+                  role
+                  organization
+                  division
+                  startDate
+                  endDate
+                }
+              }
+              # awardExperiences {
```
mnigh, line 151: Any reason why these are commented out or can they be removed?

petertgiles: I literally copied this from frontend/talentsearch/src/js/api/userOperations.graphql and I'd rather not make any cosmetic changes to the query that create noise when comparing the two files since it's super important that they stay in sync.  The only change I made was to change it from "me" to a specific user.

petertgiles: I wonder if there's a way to make sure the same query is used between the frontend and api without introducing a dependency.  :thinking:

mnigh: Ok, I see. Not sure if it's possible, but what about creating a fragment and importing it? Not sure if urql can do it but Apollo certainly [does](https://www.apollographql.com/docs/react/data/fragments/).

I found the [origin](https://github.com/GCTC-NTGC/gc-digital-talent/pull/2369/files#diff-200ddc7e7d0a5cc02eab37e0bf1f102c8757983d7c7036f576b6204d636db9ff) of the commented out code. I know this was ages ago, but any idea why the commented code was added @vd1992?

vd1992: I think that was back when experiences were not quite ready to be loaded, so it was commented out for later
instead a separate block was added in for experiences and the commented out stuff left

mnigh: Thanks! @vd1992 can the commented stuff be removed then? If not, ignore my comments please @petertgiles. If so, can you remove from both places to cut our debt and future questions regarding the unused code?

vd1992: don't see why not

petertgiles: Queries tidied in [1493a06](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3707/commits/1493a062107331594d6283d4a016c878402c7ecf)

petertgiles: Spun out query consolidation to #3713

PR #3710, opened 2022-08-18
resolve broken chromatic build

vd1992:
my last merge updated the enum Pool Candidate Statuses, I didn't realize that was something another piece was tightly coupled to and so errors resulted as the piece in LocalizedConstants wasn't updated but components that fetched the enum tried to run the enum through LocalizedConstants

testing
- check out
- run codegen and dev in /admin
- run storybook in /admin and confirms it doesn't error


esizer (approved):
Cool! Ran chromatic and everything built fine 👍

PR #3712, opened 2022-08-19
Stabilize Chromatic

petertgiles:
We have a few stories that are unstable and triggering false positives in the diffs.
image
This branch stabilizes them.

Testing:
1. Run Storybook
2. Screenshot the three stories listed
3. Stop and restart Storybook
4. Compare the three stories to the screenshots


esizer (approved):
Thanks! I was actually about to make an issue for this. I ran chromatic but I realized it will likely fail until we get #3710 merged.

PR #3754, opened 2022-08-24
Fix Hydrogen Conflicts

esizer:
## Summary

This resolves some conflicts caused by merging PRs which overwrote hydrogen changes.


JoshBeveridge (approved):
As far as I can tell via diff checker and my original commit, these look good to go!

PR #3757, opened 2022-08-24
Add leading zero to classification label

mnigh:
Resolves missing leading zero in classification label used in `useFilterOptions` hook.

### Previously
Screen Shot 2022-08-24 at 14 17 09


esizer (approved):
Nice work 🧑‍🍳

Issue #3756, opened 2022-08-24
Fancy filter changed to  maroon theme after merging with main 


brindasasi:
**Problem**
It used to be on standard blue theme. However after merged to main , it got switched to old maroon theme. 
![Screen Shot 2022-08-24 at 2.29.09 PM.png](https://images.zenhubusercontent.com/61a7bd49e20af80694dd845e/ee45e23a-131b-491c-9f3e-7e9aaa4c3bad)

gobyrne:
merged

PR #3758, opened 2022-08-24
Fancy filter styling

JoshBeveridge:
# Requirements
- [x] Change dialog heading color
- [x] Rework the dialog's layout so that it doesn't shift around when values are added to it
- [x] Review and fix the mobile layout

# Notes
Addresses #3756


mnigh (approved):
Nice work! 🎊

PR #3759, opened 2022-08-24
Updates the H2 package namespace

JoshBeveridge:
# Requirements
- [x] update the namespace for the package
- [x] test all workspaces to ensure the change works as intended

# Notes
- Also fixed a bug where Storybook was still referring to the H2 variable file.


JoshBeveridge:
Thanks @mnigh - the `maintenance` script was definitely missed on my part, and I've updated the links in the config (they wouldn't break anything as they're just for reference). `tc-report` hasn't been migrated to `2.0.0`, and probably won't be unless we see a huge need for it. For now, the old `hydrogen-design-system` org still exists in NPM, so there won't be any issues with the current version it's sitting on. If I decide to axe that org in the future as a part of a clean up, I'll make sure to come in and update it in the `package.json` (this will require an upgrade to at least `1.1.20` as that's the only `1.X.X` version published to the new org)

PR #3763, opened 2022-08-25
Revert Hydrogen Overwrites

esizer:
## Summary

Looks like `TableHeader` and `Button` were reverted again, this undoes that.


yonikid15 (approved):
Looking good 👍

Issue #3742, opened 2022-08-22
A11y - Improper use of radiogroup role (SkillFamilyPicker)


esizer:
## Issue Description

Ensures elements with an ARIA role that require child roles contain them.

## Element Location

https://github.com/GCTC-NTGC/gc-digital-talent/blob/6e6639c47991e86b9f76e27f702280af53f827f1/frontend/common/src/components/skills/SkillFamilyPicker/SkillFamilyPicker.tsx#L53-L58

## Proposed Solution

- Remove `role="radiogroup"`
- Remove `hideLegend` prop from the `RadioGroup`

## Notes

We already have our `RadioGroup` component wrapped in `fieldset` which inherently groups elements.

## Acceptance Criteria

- [x] No longer has `role="radiogroup"` or, it is moved to the `RadioGroup > fieldset`
- [x] Legend for the `fieldset` is visible

PR #3770, opened 2022-08-26
Debt - 3500 - Normalize skill picker

### LINKED ISSUE #3500: Separate tabs of skill families
## Description

We want to separate the top level of skill groupings (Technical/Transferable) into their own tab, this would make it easier to browse through.

In addition, we can then convert the list of subgroupings from "checkboxes" into "radio lists" to simplify the interaction for users. Making it so they browse the lists one at a time.

## CURRENT 👎 

image

## NEW AND IMPROVED

image

## Acceptance Criteria
- [x] Separate families into different tabs
- [x] Convert from checkboxes to radios
- [x] Add: "All + (Skill family) + skills" radio option
- [x] Align different versions of this in the code.
- [x] Jest test to confirm that result list shows/changes correctly
- [x] Copy on the word for Transferable should be as "Behavioural" in English

esizer:
Resolves #3500 #3742 

## Summary

This combines all skill pickers into a single component in `common`. At the same time, it updates it to use skill families as the tabs rather than columns within the tabs.

## Screenshot 

Screen Shot 2022-08-26 at 3 40 24 PM

### Admin

1. Navigate to `/admin/pools/create`
2. Create a new pool using the skill picker in Essential/Asset skills

### Talent Search

1. Navigate to `/search`
2. Use the new skill picker
3. Ensure results match expected

### Applicant Profile

1. Navigate to `users/{id}/profile/experiences/personal/create`
2. Use the skill picker to add/update skills
3. Ensure it is saved/updated properly


vd1992 (changes requested):
looking pretty good! 

I think the filter heading on /search is missing

image

esizer:
> I think the filter heading on /search is missing

Oops! Fixed with [fix removal of search skills copy](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3770/commits/66ef5f0642a56483a5846f1004b07141e10393cd)

Issue #3751, opened 2022-08-23
Bug: Edit Pool Test Periodically Fails


esizer:
## Summary

The edit pool jest test periodically fails when run as part of our `.github/workflows`.

## Error

REF: https://github.com/GCTC-NTGC/gc-digital-talent/runs/7523418060?check_suite_focus=true

```sh
> admin@1.3.2 test
> jest

FAIL src/js/components/pool/EditPool/EditPool.test.tsx (46.963 s)
  ● Edit Pool tests › should have save buttons that emit a save event when the status is draft

    thrown: "Exceeded timeout of 5000 ms for a test.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."

      27 |
      2[8](https://github.com/GCTC-NTGC/gc-digital-talent/runs/7523418060?check_suite_focus=true#step:7:9) | describe("Edit Pool tests", () => {
    > 2[9](https://github.com/GCTC-NTGC/gc-digital-talent/runs/7523418060?check_suite_focus=true#step:7:10) |   it("should have save buttons that emit a save event when the status is draft", async () => {
         |   ^
      30 |     const handleSave = jest.fn();
      31 |     const props = {
      32 |       ...EditPoolStory.args,

      at src/js/components/pool/EditPool/EditPool.test.tsx:29:3
      at Object.<anonymous> (src/js/components/pool/EditPool/EditPool.test.tsx:28:1)
```

## Possible Issue

We are wrapping `fireEvent` in `act`. By default `@testing-library` wraps `fireEvent` in `act`. It is only necessary to wrap in `act` when there is a state change.

### Recommended Use

```tsx
import {render, screen, fireEvent} from '@testing-library/react'

const Button = ({onClick, children}) => (
  <button onClick={onClick}>{children}</button>
)

test('calls onClick prop when clicked', () => {
  const handleClick = jest.fn()
  render(<Button onClick={handleClick}>Click Me</Button>)
  fireEvent.click(screen.getByText(/click me/i))
  expect(handleClick).toHaveBeenCalledTimes(1)
})
```

REF:

- https://testing-library.com/docs/dom-testing-library/api-events/#fireevent
- https://davidwcai.medium.com/react-testing-library-and-the-not-wrapped-in-act-errors-491a5629193b

vd1992:
I made this #3697 
But this one has a possible solution, so I will close mine in favour of yours

patcon:
[...]

PR #3795, opened 2022-08-29
Increased timeout on sporadically failing test

patcon:
Re-ticketed from https://github.com/GCTC-NTGC/gc-digital-talent/pull/3781#issuecomment-1230731898

The above got merged before I pushed the fix, so just cherry-picked it into a new PR.

This is the error that's surfacing every so often. I think we've just been re-running and it goes away:
https://github.com/GCTC-NTGC/gc-digital-talent/runs/8078250319?check_suite_focus=true#step:7:24

We should probably avoid writing tests that do so many user actions at once, but this is the easiest fix.

### DIFF
--- a/frontend/admin/src/js/components/pool/EditPool/EditPool.test.tsx
+++ b/frontend/admin/src/js/components/pool/EditPool/EditPool.test.tsx
@@ -26,6 +26,8 @@ const renderWithReactIntl = (
 };
 
 describe("Edit Pool tests", () => {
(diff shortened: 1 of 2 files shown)


esizer:
Looks like this is related to #3751 are you able to connect it?

esizer:
Just re-running the jest workflow a few times to confirm :)

esizer (approved):
Awesome work! I re-ran 5 times and never got the error 😁

Issue #3014, opened 2022-06-09
DashboardContentContainer component inconsistently used


mnigh:
## Description
The `CmoAssetTable` component [uses](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/frontend/admin/src/js/components/cmoAsset/CmoAssetTable.tsx#L73) the `DashboardContentContainer` yet other similar table components, such as the `ClassificationTable` component, [do not](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/frontend/admin/src/js/components/classification/ClassificationTable.tsx#L102). This results in an inconsistent layout with components that use `DashboardContentContainer` having padding and ones that do not use `DashboardContentContainer` do not.

## Screenshots
![Screen Shot 2022-06-09 at 12 44 51](https://user-images.githubusercontent.com/3046459/172904750-57a2ab7a-b9d7-4263-a2bc-634578a95e39.png)
--
![Screen Shot 2022-06-09 at 12 45 07](https://user-images.githubusercontent.com/3046459/172904755-2e131bcf-75f1-4bc1-8973-0a0234fdefaa.png)

mnigh:
Referenced in a previous PR: https://github.com/GCTC-NTGC/gc-digital-talent/pull/2957#issuecomment-1147744149

gobyrne:
Visually fixed in #3051 but not functionally

PR #3799, opened 2022-08-30
Design - 2656 - Admin Heading Update

### LINKED ISSUE #2656: Update headings on all admin pages
Go through all admin pages and match heading styles. The users page is currently correct, so use that as the template.

These are the top level pages that you should check. We can catch others as they come up.
- [ ] Users (already done)
- [ ] Pools
- [ ] Pool Candidates
- [ ] Requests
- [ ] Classifications
- [ ] CMO Assets
- [ ] Departments
- [ ] Skill families
- [ ] Skills

Screenshot

esizer:
Resolves #2656 and #3014 

## Summary

This replaces the headings of all admin index pages with the new pattern as well as makes the use of `<DashboardContentContainer>` consistent across all of the index pages. 

## Testing

- [ ] Check chromatic diff
- [ ] navigate to all `/admin` index pages in side menu
  - [ ] Confirm it is using the new heading style

## New Style Screenshot

 > Note: Simple text with icon to left of heading, "Add" buttons moved to right side above table

Screen Shot 2022-08-30 at 1 15 42 PM

### DIFF
--- a/frontend/admin/src/js/components/Table/Table.tsx
+++ b/frontend/admin/src/js/components/Table/Table.tsx
@@ -28,7 +28,7 @@ export interface TableProps<
   title?: string;
   addBtn?: {
     path: string;
-    label: string;
+    label: React.ReactNode;
   };
   filter?: boolean;
   pagination?: boolean;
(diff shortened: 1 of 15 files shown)


vd1992 (approved):
- went through menu header by header, loads fine and heading looks right
- no translations broken for menu or the headers
- Create buttons look right

Nice!

PR #3800, opened 2022-08-30
A11y - 3734 - Add page size label

### LINKED ISSUE #3734: A11y - Pagination select has no accessible name
## Issue Description

Ensures select element has an accessible name

## Element Location

https://github.com/GCTC-NTGC/gc-digital-talent/blob/c2b51159237ca50bb903e3017c6663f0a3a1dcdd/frontend/common/src/components/Pagination/Pagination.tsx#L205-L226

## To solve this issue, you need to...

Fix at least one (1) of these issues:

- Form element does not have an implicit (wrapped) <label>
- Form element does not have an explicit <label>
- aria-label attribute does not exist or is empty
- aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
- Element has no title attribute
- Element's default semantics were not overridden with role="none" or role="presentation"

esizer:
Resolves #3734 

## Summary

Adds a label to the page size on pagination as well as increases colour contrast of the solid, white button component.

## Testing

- [x] Build the project `npm run production --workspaces`
- [x] Compile translations `npm run intl-compile --workspaces`
- [x] Navigate to any page with pagination
- [x] Run a full page axe scan
- [x] Confirm errors relation to pagination no longer appear
- [x] Check for the `aria-label` on the pagination page size `select`


frontend/common/src/components/Pagination/Pagination.tsx
```diff
@@ -231,6 +231,11 @@ const Pagination: React.FunctionComponent<PaginationProps> = ({
                 <select
                   style={{ cursor: "pointer" }}
                   value={pageSize}
+                  aria-label={intl.formatMessage({
+                    defaultMessage: "Change page size",
```
patcon, line 235: Just because we don't use verbs on any other field, and shorter seems preferable for a11y 🤷 

```suggestion
                    defaultMessage: "Page size",
```

frontend/common/src/lang/fr.json
```diff
@@ -1353,5 +1353,9 @@
   "VOheZB": {
     "defaultMessage": "<primary>{role}</primary> à {organization}",
     "description": "Role at organization in work experience block of skill accordion."
+  },
+  "DGR0FU": {
+    "defaultMessage": "Modifier la taille de la page",
```
patcon, line 1358:
```suggestion
    "defaultMessage": "Taille de la page",
```

mnigh: @esizer you'll need to re-generate as `DGR0FU` will no longer be the id since the original English value has changed as well.

frontend/common/src/lang/fr.json
```diff
@@ -1353,5 +1353,9 @@
   "VOheZB": {
     "defaultMessage": "<primary>{role}</primary> à {organization}",
     "description": "Role at organization in work experience block of skill accordion."
+  },
+  "DGR0FU": {
```
patcon, line 1357:
```suggestion
  "fSBslq": {
```

patcon (approved):
LGTM!

(Minor note that localization `compile` step needs to come before `npm run production` :) )

PR #3806, opened 2022-08-30
update UAT seeders

vd1992:
regarding the problem with UAT seeders being very out of sync with local seeders
this is me copying over a bunch of stuff from local seeders, very much open to adjustment

feel free to commit to this directly rather than running things by me only

testing: checkout and run in php
`php artisan db:seed --class=UatSeeder`
observe /search

@brindasasi 
@gobyrne

### DIFF
--- a/api/composer.json
+++ b/api/composer.json
@@ -17,11 +17,11 @@
         "mll-lab/graphql-php-scalars": "^5.1",
         "mll-lab/laravel-graphql-playground": "^2.5",
         "nuwave/lighthouse": "^5.8",
-        "web-token/jwt-core": "^2.2"
+        "web-token/jwt-core": "^2.2",
+        "fakerphp/faker": "^1.9.1"
     },
     "require-dev": {
         "facade/ignition": "^2.5",
-        "fakerphp/faker": "^1.9.1",
         "laravel/sail": "^1.0.1",
         "mockery/mockery": "^1.4.4",
         "nunomaduro/collision": "^5.10",
--- a/api/composer.lock
+++ b/api/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "61fdeb185fe07043fa315411596b8397",
+    "content-hash": "4586890f17a6c639c0cc7275a14ac2e2",
     "packages": [
         {
             "name": "brick/math",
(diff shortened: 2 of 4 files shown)


brindasasi (commented):
image
I get this error when I deploy this branch in dev & tried to run
` php artisan db:seed --class=UatSeeder
`

petertgiles:
I'm not sure I'm clear on the issue and the solution here.
- Is it really a bug that running the UAT seeder erased some data?  That seeder is to reinitialize the environment, right?  I would expect that seeder to be run once when initializing the db and never again.
- Is it a bug that the two seeders are very different?  We have two seeders to setup two different environments.
- The changes in this branch add a whole lot of new "truncate table" activities.  The new behavior will erase a bunch more data than the behavior before. :bomb: 
- The new behavior adds a bunch of randomness.  Indeterminate behavior when setting up UAT doesn't seem useful to me.

api/composer.json
```diff
@@ -17,11 +17,11 @@
         "mll-lab/graphql-php-scalars": "^5.1",
         "mll-lab/laravel-graphql-playground": "^2.5",
         "nuwave/lighthouse": "^5.8",
-        "web-token/jwt-core": "^2.2"
+        "web-token/jwt-core": "^2.2",
+        "fakerphp/faker": "^1.9.1"
```
brindasasi, line 21: Open for suggestions on this one.
Some seeders like Skill seeder required this and since its dev dependency these seeders wouldn't run in UAT ,prod env. and will cause miss the skills for ex.
This change required to populate skill & candidates in UAT env.

brindasasi:
> I'm not sure I'm clear on the issue and the solution here.
> 
> * Is it really a bug that running the UAT seeder erased some data?  That seeder is to reinitialize the environment, right?  I would expect that seeder to be run once when initializing the db and never again.
> * Is it a bug that the two seeders are very different?  We have two seeders to setup two different environments.
> * The changes in this branch add a whole lot of new "truncate table" activities.  The new behavior will erase a bunch more data than the behavior before. 💣
> * The new behavior adds a bunch of randomness.  Indeterminate behavior when setting up UAT doesn't seem useful to me.

We need that data back in UAT as DCMO is testing on this.  This #3810  is intended for that. 
Can we achieve this in a more non-destructible way?

patcon:
> Can we achieve this in a more non-destructible way?

What if we removed truncation, and relied on `migrate:fresh` for that? Or some custom artisan command that just truncates

brindasasi:
closing this PR as we figured out the problem is not with the seeder run

PR #3809, opened 2022-08-31
Fix deprecated component styling

### LINKED ISSUE #3808: Request page Form input fields alignment is broken in UAT environment
### Problem 

In UAT environment , https://uat-talentcloud.tbs-sct.gc.ca/en/search/request this page alignment is broken 
Its not reproducible with Dev or local with any build

mnigh:
Resolves #3808.

- Adds a storybook story for SearchContainerDeprecated
- Adds a storybook story for RequestFormDeprecated

### Notes
The stories were added since these deprecated components were not represented at all in storybook even though they are still implemented in the codebase (although, behind the `FEATURE_APPLICANTSEARCH` feature flag and making it easy to miss in manual visual testing).

### Question
The other deprecated components are not represented in storybook, should they be?


petertgiles:
> My basic question on this would be do we need to maintain deprecated components?

I agree with this.  Unless functionality is broken or CMO complains I don't think it's worth touching them.

brindasasi:
> > My basic question on this would be do we need to maintain deprecated components?
> 
> I agree with this. Unless functionality is broken or CMO complains I don't think it's worth touching them.

Actually styles are broken in UAT @petertgiles and AnnMarie already pointed out that. 
Now we have only 2 option , either we turn on the feature flag and remove deprecated codes OR fix it with this PR and keep maintaining it until we turn on the flags. 

I just checked with @gobyrne , our intent is to turn both Direct Intake & Applicant Search flags once direct intake is ready 
which I hope just a couple sprint away.

petertgiles (approved):
Flipped `FEATURE_APPLICANTSEARCH` and it looks nice.  Thanks for the Storybook stories, they look great.  :+1:

PR #3811, opened 2022-08-31
add calculated priority value

### LINKED ISSUE #3749: Create calculated "Priority" field on user object
Broken off of #3449

For applicants with the same status, they should be sorted in order of "Priority". Priority is a calculated field based on three separate questions that are asked on the profile.

These are the three questions:
**From the "About Me" page:**

**From the "Government information" page:**

This is the logic for calculating the **Priority** column (don't necessarily code it this bad though)....
If "i have a priority entitlement" return Priority
else (
  If "I am a veteran of the CAF" return Veteran
  else
    return citizenship_status (citizen, permanent resident or Other)
)

Sort order should be

1. Priority
Veteran
Citizen or Permanent resident (no sorting between these 2... unless that too complicated, then citizen can be first)
Other (Jerbo listed this as "work visa" on the table, but you can just leave it as "Other" instead)

vd1992:
resolves #3749 
adds a calculated field to types Applicant and User via Laravel Accessors 

to test you can use queries for user or applicants or one that pull something with user or applicant attached to it like
```
query poolCandidate {
  poolCandidate(id: "POOLCANDIDATE_ID") {
    user {
      hasPriorityEntitlement
      armedForcesStatus
      citizenship
      priorityDerived
    }
  }
}
```

also, give consideration to any possible limitations and problems that may arise


vd1992:
as a trying things out task, I went ahead to see if I could sort by the computed value, and it seems to actually work
the resolver is barebones, it already has an issue for, I just wanted to see if I could sort it

use this query, scroll through the results, the order should be apparent
```
query sortedTest {
  sortedPoolCandidates {
    user {
      priorityDerived
    }
  }
}
```

petertgiles:
I'm a little nervous that this solution isn't very scalable.  This solution is to load every row into memory on the server, sort them, and then emit through the API, right?  For a small table like classifications this would be fine but I think we'll quickly get into trouble with tables like pool_candidates.  I really think this sorting and filtering needs to be done at the database level - it's really good at that.  I'd like to see @esizer 's suggestion of emitting a numeric value in the API implemented as well, and not expose the logic of how it is calculated.

vd1992:
> I'm a little nervous that this solution isn't very scalable. This solution is to load every row into memory on the server, sort them, and then emit through the API, right? For a small table like classifications this would be fine but I think we'll quickly get into trouble with tables like pool_candidates. I really think this sorting and filtering needs to be done at the database level - it's really good at that. I'd like to see @esizer 's suggestion of emitting a numeric value in the API implemented as well, and not expose the logic of how it is calculated.

sure, I can revert the sorting try commit if you want
it was mostly just me playing around and trying something

petertgiles:
> it was mostly just me playing around and trying something

Yeah, definitely.  Unfortunately your current solution is probably much cleaner and nicer than a solution in the db will be.  :cry:  Here's a list of some possible approaches for postgres: https://stackoverflow.com/a/8250729

esizer:
> I'd like to see @esizer 's suggestion of emitting a numeric value in the API implemented as well, and not expose the logic of how it is calculated.

Yeah, I think we could assign the value when saving that field (wherever it is set) and store it as something like a `priority_weight` column in the DB. Then, we could just use laravels query builder `orderBy` method like:

```php
User::orderBy('priority_weight', 'asc')->all();
```

> Not sure if the code is accurate, but something close 😅

vd1992:
reverted it

 
> Yeah, definitely. Unfortunately your current solution is probably much cleaner and nicer than a solution in the db will be. 😢 Here's a list of some possible approaches for postgres: https://stackoverflow.com/a/8250729

we're running postgreSQL 13.2 so stored generated columns is definitely something we can try out

petertgiles:
> we're running postgreSQL 13.2

I just checked, and our prod database is actually 12.11.  I think we're still OK to try that feature though.

petertgiles:
> we're running postgreSQL 13.2 so stored generated columns is definitely something we can try out

Stored generated columns seem like a solid solution.  I'd love to see you try it out!

vd1992:
> > we're running postgreSQL 13.2 so stored generated columns is definitely something we can try out
> 
> Stored generated columns seem like a solid solution. I'd love to see you try it out!

well, it is out of scope for this issue so not going to try it quite yet

vd1992:
actually, yes, yes I will try it out
@petertgiles

petertgiles:
> actually, yes, yes I will try it out

Hehe, busy busy on a holiday!  Looking very cool... Could you switch it to returning a numeric value that can be easily sorted on the client side?

vd1992:
> > actually, yes, yes I will try it out
> 
> Hehe, busy busy on a holiday! Looking very cool... Could you switch it to returning a numeric value that can be easily sorted on the client side?

okay done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/3811/commits/92b972070a77cb5968528911a3fbd18343a2a83a

api/database/migrations/2022_09_06_032007_generated_column_priority.php
```diff
@@ -14,13 +14,13 @@ class GeneratedColumnPriority extends Migration
...
     {
-        DB::statement("ALTER TABLE users ADD COLUMN priority_status VARCHAR(50)
+        DB::statement("ALTER TABLE users ADD COLUMN priority_status INT
         GENERATED ALWAYS AS
         (case
-        when has_priority_entitlement = TRUE then 'PRIORITY'
-        when armed_forces_status = 'VETERAN' then 'VETERAN'
-        when citizenship = 'CITIZEN' OR citizenship = 'PERMANENT_RESIDENT' then 'CITIZEN_OR_PR'
-        else 'OTHER'
+        when has_priority_entitlement = TRUE then 1
```
mnigh, line 20: Might it be worth it to have values staggered so that we can add types in the future in between?
```
has_priority_entitlement = 1
armed_forces_status = 10
citizenship = 'CITIZEN' OR citizenship = 'PERMANENT_RESIDENT' = 20
else = 30
```

vd1992: https://github.com/GCTC-NTGC/gc-digital-talent/pull/3811/commits/b826ec992e16f4b98d58027658ebccc6ae30bbcc

api/graphql/schema.graphql
```diff
@@ -108,7 +108,7 @@ type User {
     # Profile Status
     isProfileComplete: Boolean
     expectedGenericJobTitles: [GenericJobTitle] @belongsToMany
-
+    priorityDerived: PriorityDerivedStatus
```
petertgiles, line 111: I think that the "Dervied" suffix is an implementation detail and shouldn't be leaked through the API.  I'd prefer something like what Eric suggested - priorityWeight.

api/database/helpers/ApiEnums.php
```diff
@@ -296,4 +296,21 @@ public static function armedForcesStatuses() : array
             self::ARMED_FORCES_NON_CAF,
         ];
     }
+
+    /**
+     * Computed Priority Status
```
petertgiles, line 301: I think these will be unnecessary when you switch to the numeric weighting?

petertgiles (changes requested):
So cool!  I think you've got this one on-lock.  Just need to finish moving to the numeric values.

api/graphql/schema.graphql
```diff
@@ -108,7 +108,7 @@ type User {
     # Profile Status
     isProfileComplete: Boolean
     expectedGenericJobTitles: [GenericJobTitle] @belongsToMany
-
+    priorityWeight: Int @rename(attribute: "priority_status")
```
petertgiles, line 111: Hehe, let's use priority weight or priority status (I'm fine with either) but let's use just one. :-)

petertgiles (changes requested):
I love this!  Just one small nomenclaure issue to tidy up.

api/graphql/schema.graphql
vd1992: makes sense to me!
https://github.com/GCTC-NTGC/gc-digital-talent/pull/3811/commits/992c3f93043b82be2bfe6bc740812da2a3e0bc1b

petertgiles (approved):

PR #3825, opened 2022-09-02
Remove incomplete copy change from h2

petertgiles:
I found another incomplete copy change from the hydrogen upgrade.  Let's remove it for now and add it back in when it has been translated and the TODO logic has been finished.


petertgiles:
Issue to add it back in created: #3826

mnigh (approved):
👍

PR #3838, opened 2022-09-02
Align PostgreSQL versions

### LINKED ISSUE #3827: Docker container running different version of postgresql than prod server
## Description
We are running postgresql 13.2 in our [docker container](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/docker-compose.yml#L4), but on our prod server we are running postgresql 12.11.

## Acceptance Criteria
- [x] Docker container running same version of postgresql as prod server (12.11)

mnigh:
Resolves #3827.

- Adds `pgcrypto` extension that is required for the `gen_random_uuid()` function called in several migration files and is [not available before PostgreSQL 13](https://pgpedia.info/g/gen_random_uuid-function.html).

### Steps to test
1. `docker-compose down`
2. `docker-compose up --detach --build`
3. `docker-compose run --rm maintenance bash setup.sh`
4. `docker-compose run --rm maintenance bash refresh_all.sh`
5. `docker-compose exec -w /var/www/html/api php sh -c "php artisan migrate"`
6. `docker-compose exec -w /var/www/html/api php sh -c "php artisan db:seed"`
7. Verify there are no build errors


vd1992:
a question
why roll back the postgres version elsewhere instead of updating the prod version?

mnigh:
i think the issue with this PR might be related to [caching docker containers ](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/.github/workflows/e2e-tests.yml#L49)in the e2e workflow since locally the containers are not cached when running `docker-compose run --rm maintenance bash setup.sh` outside of the github workflow. @patcon, is there any way to force a refresh of the container in the workflow without changing the source code?

patcon:
I'm seeing that it's pulling the image of 12.11 fine in its own container: (this is where postgres server is installed)
https://github.com/GCTC-NTGC/gc-digital-talent/runs/8210035324?check_suite_focus=true#step:5:12

It looks like it's because later on, in the actual app php container that (where postgres client is used) it's still installing package for postgres 13 in our scripts:
https://github.com/GCTC-NTGC/gc-digital-talent/runs/8210035324?check_suite_focus=true#step:5:52

So it looks in this PR like we're using the postgres 13 client against the postgres 12 server

You could bash into the php container and see whether its package manager has packages (i think `pdo_pgsql` and `libpq-dev`) that depend on postgres 12 instead of postgres 13

I believe the docker container for `php:7.4-apache` is built from here:
https://github.com/docker-library/php/tree/master/7.4

So it looks like it's using either debian buster or bullseye. Likely the newer bullseye since we're not specifying in the `7.4-apache` tag

So looking at `libpq-dev`'s requirements for these platforms:
https://packages.debian.org/buster/libpq-dev => postgresql-doc-11 tells us it's built against postgres 11
https://packages.debian.org/bullseye/libpq-dev => postgresql-doc-13 tells us postgres 13 (what you're seeing)

So it looks like the options are:
1. leave it as-is  with postgres 13
2. go back to debian buster and postgres 11 (for bother server and client lib in php container)
3. find another package repo that offers postgres 12 for our php container os, and install that onto image as part of dockerfile (will make build time and container images bigger)
4. compile postgres 12 ourselves while building the container (will add time, and might be big if we don't clean up well -- building a minimal docker container needs tricks to keep size small and not package build tools into image, which the official builds do already)

mnigh:
> 1. leave it as-is  with postgres 13 (need confirmation, but I assume the upcoming IMTD VM image with nginx+php uses this)

thanks for your very thorough look at this, @patcon. 

i don't disagree with @vachan [here](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3838#issuecomment-1235854683) and your first option which are the same. my worry with this option is that it's not in our control to update the VM but it seems like the work to try and get back to postgres 12 is overly complex. going to close this PR.

PR #3858, opened 2022-09-06
Add new skill seeders

### LINKED ISSUE #3684: Add Skills to the database
Add (seed) the full list of skills including "Skill Category", "Skill families", "Skill Name", "Skill definition", "Similar Terms" in English AND French.

https://docs.google.com/spreadsheets/d/17Y58sxWaCgj6qoFEn333UoK0sPoqwTeIFlC4Y1ZgP-8/edit#gid=1864884487

**Acceptance Criteria**
- [ ] Add the migration for seeding 
- [ ] Delete existing skills (broken relationships should be acceptable) ( Back up existing just in case ) 
- [ ] Works on Dev, UAT, & Prod

petertgiles:
This branch adds a SkillFamilySeeder and SkillSeeder to implemented the skills in [the given worksheet](https://docs.google.com/spreadsheets/d/17Y58sxWaCgj6qoFEn333UoK0sPoqwTeIFlC4Y1ZgP-8/edit?usp=sharing).  It also adds a new artisan command to delete all the existing skills and skill families in the application.

## Testing
- Try out the deployment script

## Deployment Script
1. Ensure the database has been backed up
2. Run `php artisan delete:skills_and_skill_families`
3. Run `php artisan db:seed --class=SkillFamilySeeder`
4. Run `php artisan db:seed --class=SkillSeeder`

Closes #3684

### DIFF
--- a/api/database/seeders/SkillSeeder.php
+++ b/api/database/seeders/SkillSeeder.php
@@ -5,6 +5,7 @@ namespace Database\Seeders;
 use App\Models\Skill;
 use App\Models\SkillFamily;
 use Illuminate\Database\Seeder;
+use Illuminate\Support\Str;
 
 class SkillSeeder extends Seeder
 {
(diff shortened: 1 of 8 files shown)


esizer:
Appears as though there are some duplicate keys in the skill seeders. 

## Instances (skill name - lines found)

- Respect for Diversity - 1446, 1835
- Inclusive Mindset - 1406, 1726
- Strategy Development - 800, 1576
- Enterprise Information Technology Architecture - 557, 989
- Requirements Analysis - 503, 575, 854
- Data Analysis - 458, 629, 818
- Business Analysis - 773, 809
- Implementation of GC Digital Standards in Operations - 728, 1133

petertgiles:
> Appears as though there are some duplicate keys in the skill seeders.

Good catch!  Turns out assertions aren't working our php environment.  smh   Anyway, I've fixed my testing and sent this up to Gray to figure out how we should handle them.

brindasasi (changes requested):
Not sure if its the same thing you mentioned above but
I get AssertionError on Line 1899 on SkillSeeder.php 
 ```
database/seeders/SkillSeeder.php:1899
      assert("assert(count(array_unique($keys)) == count($reshapedData))")
```

**Comment by petertgiles** [comment]
Turns out that the skills are are actually many-to-many with skill families in the data.  I updated the logic to handle that.

**Comment by vd1992** [comment]
manually navigating through things, heading to pool edit I see this error

image

this appears due to how `KeyString` is defined
keys may not start with `_`

image

it fails for this one, where NAME_EN starts with a period

image

update that one key to something that is no longer illegal and edit pool loads

image

**Comment by vd1992** [comment]
things are a bit different for the separate skill pickers

on admin, edit pool
image

on talent search
image

this appears to be due to the difference in these queries

the one for edit pool, it calls description
https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/frontend/admin/src/js/components/pool/EditPool/editPoolOperations.graphql#L101

the one for talent search, it does not call description
https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/frontend/talentsearch/src/js/api/poolCandidateOperations.graphql#L49

File: api/database/seeders/SkillSeeder.php [hunk 1]
```diff
@@ -2,28 +2,1925 @@
...
+                // Take the provided data and reshape it to our data model
+                $model = [
+                    'key' => KeyStringHelpers::toKeyString($englishName), // no key provided so making our own slug
+                    'name' => [
+                        'en' => trim($firstMatchingRecord['skill_name_en']),
+                        'fr' => trim($firstMatchingRecord['skill_name_fr'])
+                    ],
+                    'description' => [ // no descriptions provided so reusing name
+                        'en' => trim($firstMatchingRecord['skill_definition_en']),
+                        'fr' => trim($firstMatchingRecord['skill_definition_fr'])
```
vd1992, line 1885: this isn't recycling name though?
in some cases there isn't a skill_definition provided, should there be a placeholder message placed somewhere? like on the frontend component `useGetQuery -> if (!data.description.en) {then placeholder}` just to avoid having blanks

petertgiles: Yep, it's definitely recycling the name.  The problem with a placeholder is that there's nothing to hold the place for.  :laughing: This is all the data we were provided.  Since the description field is non-nullable I figured that displaying the name a second time would be less confusing than displaying a single space or some other arbitrary constant.

esizer (approved):
Awesome! Working well for me 😀

Issue #3402, opened 2022-07-19
Translations for Work Preferences on Profile


gobyrne:
## Description
There are several copy and translation fails on Work Preferences section of the Profile. See screenshots and acceptance criteria.

Note that this is all for the V2 of the Work Preferences. Please don't do anything to change V1... we are still using those and have had a few mix ups already.

## List of things to change/fix
- [x] For the copy of the select group on the Work Preference sub-page, update the English to match:
```
I would consider accepting a job that lasts for:
...any duration. (short term, long term, or indeterminate)
...indeterminate duration only. (permanent only)
```
- [x] For the copy of the select group on the Work Preference sub-page, update the French to match:
```
J'envisagerais d'accepter un emploi :
… de toute durée. (à court terme, à long terme ou indéterminée)
… d'une durée indéterminée seulement. (permanent seulement)
```
![image](https://user-images.githubusercontent.com/16977254/179803529-46fa5152-120c-422a-b2ad-ad65274d3095.png)

- [x] From the label of the checkbox group on Work Preference sub-page, removes the word (including ellipsis) "requires..." and replace with just a colon ":"
- [x] On the French version of above remove "exige..." and replace with colon " :" (non-breaking space needed before colons in French)
- [x] Confirm that the checkbox group options on the French page are correctly displaying in French. Right now the first 2 are showing in English.

![image](https://user-images.githubusercontent.com/16977254/179803749-48682ef4-ab01-44e2-a638-278e52885163.png)

- [x] Work preferences summary on the Profile page for the "I would consider accepting a job that lasts for:" section should use the same copy as on the sub-page. For example, right now it says "Permanent duration" on the profile page, but "... only those of an indeterminate duration. (permanent)" on the sub-page select input.  BUT here the catch: We don't want the ellipsis on the summary. Please double check that the changes show up in French too.

- [x] Work preferences summary on the Profile page uses wrong copy for "I would consider accepting a job that". You'll notice that the grammar doesn't work: "I would consider accepting a job that: Availability, willingness..." Instead it should read: "I would consider accepting a job that: requires me to work occasional overtime. For context, those other strings are the ones that should be displayed to managers on the search page. Please also confirm the French options are displaying correctly for this. Currently the English ones are showing up on the French page.

![image](https://user-images.githubusercontent.com/16977254/179805080-2cdc4d6e-d2b8-44cb-92be-89caf38eedc8.png)

[...]

PR #3862, opened 2022-09-06
Add formatjs/enforce-id rule to make autogen ID easier to work with

patcon:
Pulled this out of https://github.com/GCTC-NTGC/gc-digital-talent/issues/3402 when realized how awkward updating translation strings was.

Keeping this small, and only talentsearch for now, as it has high possibility for conflicts unless merged quickly.

With this, updating a string is more in-flow of code, and simple as:

1. change the text in vscode. (note the immediately eslint underline showing that id is out-of-date)
3. save the file, and id auto-updates (see GIF)
4. `git diff` to see the old ID v new (see diff below)
5. git grep old ID from diff, to find it in `fr.json` and update/delete as appropriate

## Simple diff to see how ID changed

```diff
diff --git a/frontend/talentsearch/src/js/components/workPreferencesForm/WorkPreferencesForm.tsx b/frontend/talentsearch/src/js/components/workPreferencesForm/WorkPreferencesForm.tsx
index b9837e202..63d7b9423 100644
--- a/frontend/talentsearch/src/js/components/workPreferencesForm/WorkPreferencesForm.tsx
+++ b/frontend/talentsearch/src/js/components/workPreferencesForm/WorkPreferencesForm.tsx
@@ -179,8 +179,8 @@ export const WorkPreferencesForm: React.FC<WorkPreferencesFormProps> = ({
                       value: "true",
                       label: intl.formatMessage({
                         defaultMessage:
-                          "...any duration (short term, long term, or indeterminate duration)",
-                        id: "X2Ivfb",
+                          "...any duration. (short term, long term, or indeterminate duration)",
+                        id: "oKaV/T",
                         description:
                           "Label displayed on Work Preferences form for any duration option",
                       }),
```

## Benefits
- changes in auto-gen IDs are highly visible immediately in both vscode and version control (diffs, etc)
- easier to spot proper changes in PR (each inline autogen ID should have a clear "old" and "new" value to help coordinate update of the french translation, even if not done right in the PR)
- no need to run npm scripts to manage translations

## Notes
- this rule ignores anything with the `.` in it (e.g. `.foo` or `common.bar`), so that's how custom IDs can still be used

### DIFF
--- a/frontend/talentsearch/.eslintrc
+++ b/frontend/talentsearch/.eslintrc
@@ -50,6 +50,13 @@
     ],
     "rules": {
         "formatjs/no-id": "off",
+        "formatjs/enforce-id": [
+            "error",
+            {
+                "idInterpolationPattern": "[sha512:contenthash:base64:6]",
+                "idWhitelist": ["\\."]
+            }
+        ],
         "camelcase": "warn",
         "consistent-return": "warn",
         "import/no-extraneous-dependencies": "off",
--- a/frontend/talentsearch/src/js/components/ApplicationPageWrapper/ApplicationPageWrapper.tsx
+++ b/frontend/talentsearch/src/js/components/ApplicationPageWrapper/ApplicationPageWrapper.tsx
@@ -94,6 +94,7 @@ const ApplicationPageWrapper = ({
               <p data-h2-font-size="base(h6, 1)">
                 {intl.formatMessage({
                   defaultMessage: "Closing date:",
+                  id: "GIN69n",
                   description:
                     "Label for a pool advertisements closing date on the application",
                 })}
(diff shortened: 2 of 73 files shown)


patcon:
> I would like to get rid of the "idWhitelist": ["\\."] rule though. Let's not hand out that footgun until we really need it.

Heh, I see your point, but it's actually already needing in admin/common, just not here yet. Is it ok if I leave it in, since those will be added right after merge?

patcon:
Thanks so much @petertgiles! Reading this as approved, and so merging even without official button click form you. But pls do let me know if I've abused the process :)

petertgiles:
> Thanks so much @petertgiles! Reading this as approved, and so merging even without official button click form you. But pls do let me know if I've abused the process :)

Please don't. Remember we discussed in scrum that we want Matt and Yoni to review.

patcon:
Shoot! Sorry, I forgot 😞

patcon:
Bah, ok, well as a closing note in case anyone arrives back here later: If you come across any merge conflicts on existing PRs, prob easiest to discard the IDs rather than manually resolve conflicts -- you can let eslint just regenerate.

Issue #2917, opened 2022-05-31
Discuss how to combine frontend projects


tristan-orourke:
# Description
Right now, admin, talentsearch and IAP are organized as separate npm projects. However, we have more and more need to share components between them. And soon, there may also be some navigation between them. It makes more sense now to frame them as a single app.

It may still make sense to keep a separate `common` folder to hold a component library which is less specific to the GC Talent app. It may helpful to frame it like so: anything we might want to reuse in the Accessibility Passport should go in the `common` folder.

## Acceptance Criteria

- [ ] Agree on a directory structure for the final project

tristan-orourke:
Possibly we can just let projects import from each other, as a shortcut.

esizer:
Just mulling things over, looking at our existing architecture they are fairly similar so simply merging them while keeping them pretty much the same should work.

## Proposed Architecture

```sh
frontend/
├── common
└── app/
    ├── .storybook
    ├── public/
    │   ├── images
    │   ├── .htaccess
    │   ├── config.sjs
    │   ├── index.html
    │   └── site.webmanifest
    ├── src/
    │   ├── css/
    │   │   └── app.css
    │   └── js/
    │       ├── api
    │       ├── components
    │       ├── context
    │       ├── hooks/
    │       │   └── use*Routes.ts
    │       ├── lang
    │       ├── pages
    │       ├── routes/
    │       │   └── Router.tsx
    │       ├── tests/
    │       │   └── coverage
    │       ├── types
    │       └── App.tsx
    ├── .babelrc
    ├── .browserslistrc
    ├── .env.example
    ├── .eslintrc
    ├── .gitattributes
    ├── .gitignore
    ├── .nvmrc
    ├── .prettierrc
    ├── .styleci.yml
    ├── codegen.yml
    ├── jest.config.js
    ├── LICENSE
    ├── package.json
    ├── README.md
    ├── tsconfig.json
    ├── webpack.base.js
    ├── webpack.dev.js
    └── webpack.prod.js
```

## Common

This would remain the same.

## App

This holds all of the actual application files and essentially merges `admin`, `talentsearch` and `indigenousapprenticeship` applications into a single directory.

### API 

Keep this pretty much the same while trying to move the `graphql` files to colocate them with where they are used.

### Components

This would be reworked to only include functional components and no longer be the place where page components live.

### Context (optional)

[...]

Issue #2895, opened 2022-05-27
Bug - Storybook i18n not working


esizer:
## Summary

We have applied a hack in #2888 to get a storybook displaying both French and English versions of the homepage story. We need to undo this change and properly implement i18n.

## Details

We are currently using [storybook-addon-intl](https://storybook.js.org/addons/storybook-addon-intl). It seems as though the locale is changing (the logo updates) however, the messages are not properly being loaded in `react-intl` so any text localized by it is not updating.

esizer:
Believe this magically fixed itself 🦄

esizer:
> Believe this magically fixed itself 🦄

Nvm, it works while running storybook locally but not while hosted on chromatic.

mnigh:
Seems to work on some stories in Chromatic and not others. Perhaps something to do with how the projects are importing the frCompiled.json file?

patcon:
This should be resolved in the above PR. It was only an issue with the merged storybook we render via chromatic (a special case of the admin storybook), but unmerged local storybooks worked fine

patcon:
Closing as associated PR is merged :)

PR #3866, opened 2022-09-07
Import talentsearch french strings into merged Storybook (hosted on Chromatic)

patcon:
talentsearch french string weren't being merged into the omni-storybook during build, and so storybook wasn't properly reflecting live app translations.

## Before

Screen Shot 2022-09-07 at 10 40 15 AM

https://main--61e099a1bb1465003a73d3ce.chromatic.com/?path=/story/workpreferencesform--individual-work-preferences

## After

Screen Shot 2022-09-07 at 10 39 39 AM

### DIFF
--- a/frontend/admin/.storybook/preview.js
+++ b/frontend/admin/.storybook/preview.js
@@ -4,6 +4,7 @@ import "../src/css/app.css"
 import { setIntlConfig, withIntl } from 'storybook-addon-intl';
 import AdminFrench from "../src/js/lang/frCompiled.json";
 import CommonFrench from "../../common/src/lang";
+import TalentFrench from "../../talentsearch/src/js/lang/frCompiled.json";
 import defaultRichTextElements from "../../common/src/helpers/format";
 import MockGraphqlDecorator from "../../common/.storybook/decorators/MockGraphqlDecorator";
 import { INITIAL_VIEWPORTS } from "@storybook/addon-viewport";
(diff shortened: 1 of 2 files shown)


patcon:
> Oh! Is this what is causing issues in https://github.com/GCTC-NTGC/gc-digital-talent/issues/2895 ?

Looks like it! Submitted a fix, and ready for review after Chromatic builds.

esizer (approved):
Yay! Working as expected 😄  Also, thank-you for reverting the French page hack 🙇‍♀️ 

## Talent Search

Screen Shot 2022-09-07 at 1 27 57 PM

## IAP

Screen Shot 2022-09-07 at 1 27 08 PM

patcon:
thanks for review @esizer 🙌

patcon:
Ugh, note that there are some irritating edge-cases if the same string+description exists in both common and talentsearch/admin, as keys will override.

Current load order in admin storybook is admin > common > talent > indigenous (with later ones overriding). We could adjust the order, but still possible to get caught

I just spent a bunch of time changing it in common's fr.json, and banging my head trying to figure out why it wasn't reflected in storybook. Turns out it's that talentsearch's fr.json was overriding my change :/

I suppose this is a subtle consequence of maintaining separate apps, which would be resolved by https://github.com/GCTC-NTGC/gc-digital-talent/issues/2917

PR #3867, opened 2022-09-07
Add formatjs/enforce-id rule (continued: `common`)

patcon:
> Another section of https://github.com/GCTC-NTGC/gc-digital-talent/pull/3862, whose purpose is to make string changes more reviewable, as each string change will have the old and new autogen id in the diff.
>
> Sorry, I mistakenly merged before getting your input on the last PR, but this is the same tactic.
>
> (Separate PRs per app because nature of strings peppered throughout make this prone to conflicts.)


mnigh (commented):
The eslint rule is pretty neat, nice find 👍.

A bit of a nitpick, but the order of the object [according to the docs](https://formatjs.io/docs/react-intl/api#message-descriptor) is: id, defaultMessage, description. I don't think there is a way to change this since it's auto-generated, so I guess ignore.

We still have some non-generated IDs in several places already (for example: https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/frontend/common/src/constants/localizedConstants.tsx#L930). Should we be changing these explicit IDs as well to be auto-generated?

yonikid15 (commented):
Ya I like it too! It should make editing easier which is great 👍

patcon:
thanks for checking it out, los dos!

> A bit of a nitpick, but the order of the object [according to the docs](https://formatjs.io/docs/react-intl/api#message-descriptor) is: id, defaultMessage, description

@mnigh unfortunately eslint's fixups can't handle that easily :( it can be manually done, but it feels like maybe a precedent not worth setting, as we'll have to keep fighting its preference and correcting it in PRs. Is it ok to just leave it as-is?

(Maybe this is the sort of thing we could bank as a potential custom eslint rule in the future. If you'd like, we could create an issue to track these sorts of small "nits" that are easier to address when they're automatic: https://eslint.org/docs/latest/developer-guide/working-with-rules#applying-fixes)

> Should we be changing these explicit IDs as well to be auto-generated?

I recall those were changed because the updates we were doing on a few sets of label variants were confusing and causing regressions due to detachment mistakes. My light sensibility is kinda Chesterton's Fence, and not to revert unless it's causing trouble. Would it be ok to leave as-is?

patcon:
OK, done @mnigh! But also, I do think manual IDs have their place, so I hope we can stay open to using them sometimes :)

mnigh (approved):
👍

PR #3868, opened 2022-09-07
Add formatjs/enforce-id rule (continued: `indigenousapprenticeship`)

patcon:
> Another section of https://github.com/GCTC-NTGC/gc-digital-talent/pull/3862, whose purpose is to make string changes more reviewable, as each string change will have the old and new autogen id in the diff.
>
> Sorry, I mistakenly merged before getting your input on the last PR, but this is the same tactic.
>
> (Separate PRs per app because nature of strings peppered throughout make this prone to conflicts.)


mnigh (approved):
👍

PR #3893, opened 2022-09-08
make fancy filter work on user table

### LINKED ISSUE #3085: Use user filter object to filter Users table (Fancy Filter)
## Description

When the filter modal is closed, use the filter object to refresh the query.
Filter can be stored in state on the user table.

## Acceptance Criteria
- [ ] When the modal is opened, it is prepopulated with the current state.
- [ ] When the modal is closed, the filter state is saved and the table query refreshed.
- [ ] Cypress tests

vd1992:
closes #3085 
This makes the filter actually operate and changes the query. 
To make this work, I added a state in the parent component UserTable that is changed in the (grand) child component UserTableFilterDialog. That way, the filter change is received in UserTable enabling query re-run. 
I also combined the search bar filter with the fancy filter enabling both to work together. 
A single test was added the existing Jest file, not sure what useful testing could be added because not sure a Cypress test may help much as it is important to ensure the generated request matches the filtering selected by manual testing

Testing: 
go to /admin/users
pop open the network tab on the side
play around with the filtering, try filtering something then ensure it is mirrored accurately in a network request 
manual exercise filtering cases, the goal is to ensure the appropriate request is generated


frontend/admin/src/js/components/user/util.ts
```diff
@@ -0,0 +1,103 @@
...
+): LanguageAbility | undefined {
+  if (selection === LanguageAbility.Bilingual) {
+    return LanguageAbility.Bilingual;
+  }
+  if (selection === LanguageAbility.English) {
+    return LanguageAbility.English;
+  }
+  if (selection === LanguageAbility.French) {
+    return LanguageAbility.French;
+  }
```
patcon, line 20: This should be the same without explicitly mentioning each:

```suggestion
  if (Object.values(LanguageAbility).includes(selection as LanguageAbility)) {
    return selection as LanguageAbility;
  }
```

frontend/admin/src/js/components/user/util.ts
```diff
@@ -0,0 +1,103 @@
...
+  if (selection === WorkRegion.Prairie) {
+    return WorkRegion.Prairie;
+  }
+  if (selection === WorkRegion.Quebec) {
+    return WorkRegion.Quebec;
+  }
+  if (selection === WorkRegion.Telework) {
+    return WorkRegion.Telework;
+  }
```
patcon, line 50:
```suggestion
  if (Object.values(WorkRegion).includes(selection as WorkRegion)) {
    return selection as WorkRegion;
  }
```

frontend/admin/src/js/components/user/util.ts
```diff
@@ -0,0 +1,103 @@
...
+  if (selection === OperationalRequirement.TransportEquipment) {
+    return OperationalRequirement.TransportEquipment;
+  }
+  if (selection === OperationalRequirement.Travel) {
+    return OperationalRequirement.Travel;
+  }
+  if (selection === OperationalRequirement.WorkWeekends) {
+    return OperationalRequirement.WorkWeekends;
+  }
```
patcon, line 86:
```suggestion
  if (Object.values(OperationalRequirement).includes(selection as OperationalRequirement)) {
    return selection as OperationalRequirement;
  }
```

frontend/admin/src/js/components/user/util.ts
```diff
@@ -0,0 +1,103 @@
...
+): JobLookingStatus | undefined {
+  if (selection === JobLookingStatus.ActivelyLooking) {
+    return JobLookingStatus.ActivelyLooking;
+  }
+  if (selection === JobLookingStatus.Inactive) {
+    return JobLookingStatus.Inactive;
+  }
+  if (selection === JobLookingStatus.OpenToOpportunities) {
+    return JobLookingStatus.OpenToOpportunities;
+  }
```
patcon, line 101:
```suggestion
  if (Object.values(JobLookingStatus).includes(selection as JobLookingStatus)) {
    return selection as JobLookingStatus;
  }
```

frontend/admin/src/js/components/user/util.ts
patcon: There maybe a more general way to do this in typescript (so one function works for all of these), but I'm unsure what this is off the top of my head :)

frontend/admin/src/js/components/user/UserTableFilterDialog.tsx
```diff
@@ -250,6 +263,42 @@ const UserTableFilterButton = ({
   const handleSubmit: SubmitHandler<FormValues> = (data) => {
     onSubmit(data);
     setActiveFilters(data);
+    // this state lives in the UserTable component, this step also acts like a formValuesToSubmitData function
+    onFilterChange({
```
patcon, line 267: Do you think this component needs 2 submission handlers? As in could this be moved into the onSubmit function that was already present? Maybe it was just poorly named, but `onSubmit` was the place where I assumed this logic would be injected/added

maybe `onSubmitFilter`? which identifies itself as "for a form" but also "for the filter" like you capture?

patcon: EDIT: Wait. Nevermind, might have mixed myself up here. gimme a sec :)

vd1992:
handy, merged it in, neat stuff
thanks @patcon

frontend/admin/src/js/components/user/UserTableFilterDialog.tsx
```diff
@@ -250,6 +263,42 @@ const UserTableFilterButton = ({
...
+      applicantFilter: {
+        expectedClassifications: data.classifications.map((classification) => {
+          const splitString = classification.toString().split("-");
```
patcon, line 270: I think we could prob adjust this type to only be string, which would let you clean this up a bit. We have some generic type variables of `Object<V>` but it's always string
```
# shows only string
git grep 'Option<[a-z]\+>'
```

For no particular reason I dragged forward an Option type from earlier that allowed key to be either string or number:
https://github.com/GCTC-NTGC/gc-digital-talent/blob/71ded63044636b25c0b5b0ccaad7a106eec85e34/frontend/common/src/components/form/Select/SelectFieldV2.tsx#L11

But I don't think we ever use a number, except in the stories for SelectField:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/71ded63044636b25c0b5b0ccaad7a106eec85e34/frontend/common/src/components/form/Select/SelectFieldV2.stories.tsx#L38-L40

https://github.com/GCTC-NTGC/gc-digital-talent/blob/71ded63044636b25c0b5b0ccaad7a106eec85e34/frontend/common/src/components/form/MultiSelect/MultiSelectFieldV2.stories.tsx#L38-L40

frontend/admin/src/js/components/user/UserTableFilterDialog.tsx
```diff
@@ -250,6 +263,42 @@ const UserTableFilterButton = ({
...
+        }),
+        operationalRequirements: data.operationalRequirement.map(
+          (requirement) => {
+            return stringToEnumOperational(requirement.toString());
+          },
+        ),
+        skills: data.skills.map((skill) => {
+          const skillString = skill.toString();
+          return { id: skillString };
+        }),
```
patcon, line 287: E.g.,

```suggestion
        skills: data.skills.map((skill) => ({ id: skill }),
```

frontend/admin/src/js/components/user/UserTable.tsx
```diff
@@ -96,13 +83,43 @@ export const UserTable: React.FC = () => {
     col: string | undefined;
   }>();
 
+  // merge search bar input with fancy filter state
+  const filterCombiner = (
```
patcon, line 87: re: starting with verb, maybe something descriptive like

```suggestion
      where: buildUserFilterInput(fancyFilter, searchState?.term, searchState?.col),
```

frontend/admin/src/js/components/user/UserTable.tsx
```diff
@@ -96,13 +83,43 @@ export const UserTable: React.FC = () => {
...
+    fancyFilterState: UserFilterInput | undefined,
+    searchBarTerm: string | undefined,
+    searchBarCol: string | undefined,
```
patcon, line 90: "column" feels like a leaky abstraction from the table

```suggestion
    searchType: string | undefined,
```

frontend/admin/src/js/components/user/UserTableFilterDialog.tsx
vd1992: yeah, I didn't want to be too destructive when it came to adding this pull request in so I tried to just adjust things to fit
typing changes https://github.com/GCTC-NTGC/gc-digital-talent/pull/3893/commits/70f4cfb7afae25937fa4320f5618497948611318

frontend/admin/src/js/components/user/UserTable.tsx
vd1992: sounds reasonable 
https://github.com/GCTC-NTGC/gc-digital-talent/pull/3893/commits/a2cb77df9d066a66a0701d8289082a4d0f68742f

frontend/admin/src/js/components/user/UserTable.tsx
vd1992: https://github.com/GCTC-NTGC/gc-digital-talent/commit/a2cb77df9d066a66a0701d8289082a4d0f68742f

frontend/admin/src/js/components/user/UserTableFilterDialog.tsx
vd1992: https://github.com/GCTC-NTGC/gc-digital-talent/commit/70f4cfb7afae25937fa4320f5618497948611318

frontend/admin/src/js/components/user/UserTableFilterDialog.tsx
patcon: Bah, I'm fumbling to think how to merge these nicely, but do you see any way that a single function is possible instead of two? I think ideally the processing of formvalues into input type would happen in a single place, but I'm not seeing a clean way to do that. I recall onSubmit is a handler set up one level that doesn't do much, so feels odd to have two

vd1992: if it is too much of a mess to attempt it, then it might not be worth the hassle frankly

patcon: i'm onboard with that. extracting the processing out of the deep component is the hard part...

but what about just stripping out current `onSubmit` and renaming `onFilterChange` to `onSubmitFilter` (to hint at form submission and show up when grepping for "onSubmit")

patcon: Oh! Wait... you could also use it to pass your function, and keep the processing out of the component:

```tsx
  const handleFilterSubmit: SubmitHandler<FormValues> = (data) => {
    onFilterChange( //...
    // ...
  };
```

vd1992:
alright made some changes @patcon 
removed the function here https://github.com/GCTC-NTGC/gc-digital-talent/pull/3893/commits/af0b74fac45f4a1e5ece1323677786aec4824835 
and changed filter format to add education here https://github.com/GCTC-NTGC/gc-digital-talent/pull/3893/commits/b87e3543dae9a148d9b6264127d90e6810e357b5

patcon:
Thanks vachan! I really appreciate accommodating the feedback :) I'm realizing it feels like maybe a step backward to have the UserTableFilterDialog test checking shapes of the _transformed_ state, as it exists in UserFilterInput (whose state lives up two levels of components in UserTable).

I'll show what I meant with the other way, which keeps outside concerns out of the test (I'll push a branch with the reversion and change, and feel free to yoink those commits into here if it looks ok!)

brindasasi (changes requested):
I don't see the network requests  when I play around with the filters. Fancy filter doesn't filter the rows as it should be. 
Possible it is broken with the latest changes?

patcon:
Hm, will test. Thanks @brindasasi! Just rebuilding now

Thoughts on e2e test for this now that it's working in the app? I'm happy to write it or support writing it, but if we're itching for merge on this, I could spin it out

vd1992:
huh, that is odd, still seems to work on my end

image

image

brindasasi:
hah .. my bad .. clearing cache & everything .. it works now .
In safari there is a spacing issue in there. Chrome looks fine. Is it easy enough to fix in this pr? 

Screen Shot 2022-09-13 at 5 09 19 PM

vd1992:
> hah .. my bad .. clearing cache & everything .. it works now . In safari there is a spacing issue in there. Chrome looks fine. Is it easy enough to fix in this pr?
> 
> Screen Shot 2022-09-13 at 5 09 19 PM

it looks fine in Firefox and in Edge too
hm, I am hesitant about playing around with this because it could be a tricky problem and actually end up being a Hydrogen thing

patcon:
Safari thing was pretty small, so pushed a commit :) thanks B

vd1992:
> Safari thing was pretty small, so pushed a commit :) thanks B

nice!

patcon:
As mentioned, pushed an example of the reversions, and the alt approach:
https://github.com/GCTC-NTGC/gc-digital-talent/pull/3955

(It's not as drastic as it may look, but bringing back the prior tests make it look bigger than it is 😬  Basically just moving the FormValues=>UserFilterInput transformation from UserTableFilterDialog to TableHeader, which is closer to where the UserFilterInput state lives in UserTable component)

When bringing back the very focussed tests from before, I'm not sure how to salvage the tests you'd repurposed from modifying the prior ones. If we want that transformation code tested (and we don't need to), I wonder if it should be tested on its own, rather than trying to wrap it up within the tests of UserTableFilterDialog?

Like everything inside onFilterChange is essentially a transformation that could have it's own function outside the component and be tested separately, and needn't be part of the full component test :)

vd1992:
> As mentioned, pushed an example of the reversions, and the alt approach: #3955
> 
> (It's not as drastic as it may look, but bringing back the prior tests make it look bigger than it is 😬 Basically just moving the FormValues=>UserFilterInput transformation from UserTableFilterDialog to TableHeader, which is closer to where the UserFilterInput state lives in UserTable component)
> 
> When bringing back the very focussed tests from before, I'm not sure how to salvage the tests you'd repurposed from modifying the prior ones. If we want that transformation code tested (and we don't need to), I wonder if it should be tested on its own, rather than trying to wrap it up within the tests of UserTableFilterDialog?
> 
> Like everything inside onFilterChange is essentially a transformation that could have it's own function outside the component and be tested separately, and needn't be part of the full component test :)

I already merged it in lol

patcon:
haha ok, i think it's good to merge if you don't feel like extracting the transformation into a testable `transformFormValuesToUserFilterInput(data)`! I'll just approve and leave it up to you

patcon:
And any feelings on cypress tests here? I could just mark it as `e2e-candidate` if you'd rather get this merged asap?

patcon (approved):
Approving, as this is blocking and I'd rather see you getting the endorphins of merging if you need it, but I think a cypress test and tests of that transformation could be nice! If you agree, let's track them in other issues (the ones blocked on this) :)

vd1992:
> And any feelings on cypress tests here? I could just mark it as `e2e-candidate` if you'd rather get this merged asap?

I wasn't sure how to meaningfully do a Cypress test here since there are so many filters at play that it is hard to accurately ensure things are working as intended without a thorough manual exercise.

vd1992:
@patcon 
uh-oh, styling broke

image

vd1992:
oops, I'm silly, I shrunk the window that is why the rendering changed 💀

patcon:
> I wasn't sure how to meaningfully do a Cypress test here since there are so many filters at play that it is hard to accurately ensure things are working as intended without a thorough manual exercise.

Happy to defer, but also, this feels like classic place for "don't let perfect be enemy of the good". Even a single check of one field would catch a whole class of wiring-things-together regressions :)

brindasasi (commented):
Going into the filter pop up didn't trigger the network calls in my local. 
It works in Pat's local as he mentioned. I'm fine if you go ahead with the granted approval. I can still continue testing on this in Dev environment to make sure.

vd1992:
@patcon 
I added a basic check with https://github.com/GCTC-NTGC/gc-digital-talent/pull/3893/commits/6d47667251a95312a1a00b0c652051ec1c2a6f15
if that is fine by you I will merge this in

patcon:
LGTM! Wooooo

PR #3895, opened 2022-09-08
add status sorting weight

### LINKED ISSUE #3874: Add sorting weight based on status
In the same way that priority weight was implemented as a stored numeric weight, also add a database column generated weight based on status: 

- [x] DB Migration
- [x] Model and schema changes

### Sorting based on Status
_Filtering has been broken off into a separate issue #3813_

Here is the full list of statuses. The applicants should be displayed in order the order their status appears here (except for Draft applications which should not even be retrieved from the DB). (So "New application" should be at the top of the table and "placed casual" at the bottom unless fileters are removed).

0. **Draft** - Never visible. This data should not be retrieved from server at all for this table.
1. **New application** 
2. **Application under review** 
3. **Screened in (available)** 
4. **Screened out - On application** - Filtered out by default
5. **Under assessment**	
6. **Screened out - On assessment** - Filtered out by default
7. **Qualified - Available**
8. **Qualified - Unavailable**	
9. **Qualified - Withdrew** - Filtered out by default
10. **Placed Casual**
11. **Placed Term** - Filtered out by default
12. **Placed Indeterminate** - Filtered out by default

vd1992:
closes #3874 
Adds the generated column that is based off of candidate status. Updates the API to use it as well. 
Also, for a high degree of reliability I added testing to ensure every status returns the correct weight. 

Testing: 
Use this query, observe status and statusWeight. Change Status and then observe the query results again. 
```
query poolCandidate {
  poolCandidate(id: "ID") {
    status
    statusWeight
  }
}
```

### DIFF
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -28,6 +28,7 @@ use Illuminate\Database\Eloquent\Builder;
  * @property array $location_preferences
  * @property array $expected_salary
  * @property string $pool_candidate_status
+ * @property int $status_weight
  * @property int $pool_id
  * @property int $user_id
  * @property array $accepted_operational_requirements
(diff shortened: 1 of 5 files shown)


petertgiles (approved):
Love it!  :dart:

PR #3898, opened 2022-09-08
Add ability to toggle MFA for GCKey using env variable

### LINKED ISSUE #3883: Make it possible to turn off MFA for GCKey
SiC agrees that it is helpful to be able to turn off MFA in non-prod environments.  Let's add an environment variable to be able to turn it off so that we can use shared accounts for testing.

mnigh:
Resolves #3883.

### Steps to test locally (for MFA-disabled)
1. Set api/.env OAUTH_ACR_VALUES=gckey
2. Navigate to /en/login-info
3. Click on **Continue to GC Key and Login** button
4. Confirm MFA is not required to successfully sign in

### Steps to test locally (for MFA-enabled)
1. Set api/.env OAUTH_ACR_VALUES=gckeymfa
2. Navigate to /en/login-info
3. Click on **Continue to GC Key and Login** button
4. Confirm MFA is not required to successfully sign in


petertgiles (commented):
This looks right to me but I suggest setting it so that the 'gckeymfa' is the default behavior with no env variable by setting it as the default value in the config/oauth.php file.  This way we won't need to set the variable in production and we won't need a manual deployment step.

Ping me later today if you want to test it out on the dev server together.

petertgiles:
Nice, this works locally for me now.

petertgiles:
I've added the deployment label to the issue since this will require creating a new environment variable in the app services.

petertgiles (approved):
Tested on dev and it works!

PR #3900, opened 2022-09-08
Added viewport size back into IAP Home component story

patcon:
This was mistakenly excised in #3866


esizer (approved):
Good catch, thanks!

Issue #3682, opened 2022-08-16
Show failing status check for eslint when any warnings


patcon:
<sup>ℹ️ [Figma (root)][figma] | [Sitemap][sitemap]</sup>
<sup>_This initial comment is collaborative and open to modification by all._</sup>

Re-ticketed [from Slack message](https://talent-cloud.slack.com/archives/C0259G6KXJ6/p1660590261270909) :lock:

## Description

Right now, we have annotations show up in PRs for any eslint warnings. In practice, we don't merge until the warnings are fixed, but the check shows as green unless there are errors. This makes it easy to miss warnings that slow down review.

Proposal: We should fail the check if there are any warnings, so we get a more visible hint. We can do it by using this in the workflow: `npm run lint --workspaces -- --max-warnings=0` (the `--` means that any following args get passed into the npm run-script command, which in this case is eslint CLI)

## Acceptance Criteria
- [ ] a single warning will show a failure in PR checks

   [image]: https://user-images.githubusercontent.com/305339/166570935-6425a615-82c7-4b2f-9840-5c7d698fa2ab.png
   [image-link]: https://example.com
   [figma]: https://go.talent.c4nada.ca/figma
   [sitemap]: https://go.talent.c4nada.ca/sitemap

patcon:
Something that would have been caught by this:
https://github.com/GCTC-NTGC/gc-digital-talent/pull/3900

PR #3901, opened 2022-09-08
Force eslint workflow to fail build when even one warning

patcon:
Closes https://github.com/GCTC-NTGC/gc-digital-talent/issues/3682

Quick fix, and no more merges with eslint issues :)


esizer (approved):
Sweet! 🍭  Once we merge the other PR containing the fix for this warning, this will be very welcome!

PR #3908, opened 2022-09-09
Fix up work preferences copy (form and profile section)

patcon:
Resolves https://github.com/GCTC-NTGC/gc-digital-talent/issues/3402

## To Do
- [x] 4392a0c9f8f8b2 use `candidateDescription` variant for operationalReqs on profile workpref section
- [x] b1f75eaaba6da3bea2e806c5ad9fc921e922625c update colons and ellipsis as requested for french
- [x] 2a176dda306f794d4c05733d3a744b7a724f2a59 fix french ellipsis whitespace 
- [x] 9cafb4cc7b0000038a12789c7e29d70410db2571 de-dup some strings we wanted same, but description was diff
- [x] lowercase strings we want to use in a few places
- [x] 9cafb4cc7b0000038a12789c7e29d70410db2571 Extract "..." from strings into separate formatting component `<WithEllipsisPrefix>` which can wrap translated strings
- [x] c870e8b4375de178d7339e51d46fcc3070017d10 fix formatting of negation sub-section
- [x] 6375028 rename operationalReqs variant to be more descriptive, from `candidateDescription` to `firstPerson` 
- [x] 01cf18f 1180ae6 converted all single-char ellipsis to 3-char variant

## To Review
- visit each User Profile stories and scroll to Work Pref section. Flip between FR and EN and note all changing.
  - in negation sub-section, note that formatting renders properly
  - mainline story: https://main--61e099a1bb1465003a73d3ce.chromatic.com/?path=/story/admin-user-profile--user-profile-story-5
  - PR story:  https://61e099a1bb1465003a73d3ce-caujsixcak.chromatic.com/?path=/story/admin-user-profile--user-profile-story-5
- visit WorkPreferencesForm story and cycle between EN and FR (links: [mainline story](https://main--61e099a1bb1465003a73d3ce.chromatic.com/?path=/story/workpreferencesform--individual-work-preferences) / [PR story](https://61e099a1bb1465003a73d3ce-caujsixcak.chromatic.com/?path=/story/workpreferencesform--individual-work-preferences))
  - note all the mistakes and inconsistencies in French, fixed in English
  - In options, note that all option text is same as in WorkPref section of UserProfile, but with no ellipses
  - In labels, note that colons are used with proper space before

### DIFF
--- a/frontend/talentsearch/src/js/components/workPreferencesForm/WorkPreferencesForm.tsx
+++ b/frontend/talentsearch/src/js/components/workPreferencesForm/WorkPreferencesForm.tsx
@@ -240,10 +240,7 @@ export const WorkPreferencesForm: React.FC<WorkPreferencesFormProps> = ({
                     label: (
                       <WithEllipsisPrefix>
                         {intl.formatMessage(
-                          getOperationalRequirement(
-                            value,
-                            "candidateDescription",
-                          ),
+                          getOperationalRequirement(value, "firstPerson"),
                         )}
                       </WithEllipsisPrefix>
                     ),
--- a/frontend/talentsearch/src/js/components/workPreferencesForm/WorkPreferencesForm.tsx
+++ b/frontend/talentsearch/src/js/components/workPreferencesForm/WorkPreferencesForm.tsx
@@ -165,8 +165,8 @@ export const WorkPreferencesForm: React.FC<WorkPreferencesFormProps> = ({
                   idPrefix="required-work-preferences"
                   legend={intl.formatMessage({
                     defaultMessage:
-                      "I would consider accepting a job that lasts for...",
-                    id: "/DCykA",
+                      "I would consider accepting a job that lasts for:",
+                    id: "GNtu/7",
                     description:
                       "Legend Text for required work preferences options in work preferences form",
                   })}
@@ -179,8 +179,8 @@ export const WorkPreferencesForm: React.FC<WorkPreferencesFormProps> = ({
                       value: "true",
                       label: intl.formatMessage({
                         defaultMessage:
-                          "...any duration (short term, long term, or indeterminate duration)",
-                        id: "X2Ivfb",
+                          "...any duration. (short term, long term, or indeterminate duration)",
+                        id: "oKaV/T",
                         description:
                           "Label displayed on Work Preferences form for any duration option",
                       }),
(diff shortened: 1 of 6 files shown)


frontend/talentsearch/src/js/components/workPreferencesForm/WorkPreferencesForm.tsx
```diff
@@ -39,6 +39,24 @@ export interface WorkPreferencesFormProps {
   ) => Promise<UpdateWorkPreferencesMutation["updateUserAsUser"]>;
 }
 
+interface WithEllipsisPrefixProps {
+  children: React.ReactNode;
+}
+const WithEllipsisPrefix = ({ children }: WithEllipsisPrefixProps) => {
+  const { formatMessage } = useIntl();
+  const ellipsisPrefix = formatMessage({
+    defaultMessage: "...",
```
esizer, line 48: I'm not sure if it would work but could we use a `&hellip;` here or would it just render as text rather than `...`?

esizer (approved):
Works for me! Just one small question about ellipsis character.

frontend/talentsearch/src/js/components/workPreferencesForm/WorkPreferencesForm.tsx
patcon: I can scope that! I tried to get non-breaking space working via that syntax, and it printed out, but maybe that's a specialcase for whitespace chars 🤷

patcon: Hm, seems that only unicode works, not html entities.

```diff
diff --git a/frontend/talentsearch/src/js/lang/fr.json b/frontend/talentsearch/src/js/lang/fr.json
index fe7a5063f..34928125a 100644
--- a/frontend/talentsearch/src/js/lang/fr.json
+++ b/frontend/talentsearch/src/js/lang/fr.json
@@ -1,6 +1,6 @@
 {
   ".ellipsis": {
-    "defaultMessage": "… "
+    "defaultMessage": "\u2026 "
   },
   "+ZXZj+": {
     "defaultMessage": "S’il n’y a aucun candidat qui répond à vos critères <a>Communiquez avec nous! </a>",
```

Any pref? My loose one is to stick with "..." (3 chars). The only reason I used the single char for french was because it was used in a bunch of translations, and I thought maybe it was a french language preference. But I suspect it was just the word processor auto-correcting.

I suppose I could ask @gobyrne: is the single char "horizontal ellipsis" preferred for french translations?

If not, I'm inclined to just standardize on "3 dot chars". Better to be consistent, bc they look slightly diff. And 3 dots is just easier to type.

esizer (approved):
🆒 , looks great!

frontend/talentsearch/src/js/components/workPreferencesForm/WorkPreferencesForm.tsx
patcon: ✅  Reverted to 3-char version for French

patcon:
As per scrum conversation, will just use 3-char version of ellipsis instead of single-char for all french and english strings.

Reason: There is no special imperative to use it for French, and it's simpler to type and search for strings that include it (no need to look up char code for copy).

PR #3907, opened 2022-09-09
Bug/3820 windows node path error

JoshBeveridge:
# 📝 Notes
The `shelljs` package wasn't executing chained `cd..;` commands the same way on different OSes, so after experimenting with @vd1992 we found out that using `shell.cd` in tandem with `shell.exec` was a reliable way to get consistent output between Windows and WSL.

# ✅ Acceptance criteria
- [ ] `npm run dev` successfully compiles Hydrogen on Windows from any workspace
- [ ] `npm run storybook` successfully compiles Hydrogen on Windows from any workspace
- [ ] Saving changes results in Storybook successfully recompiling without any errors on Windows from any workspace
- [ ] `npm run dev` successfully compiles Hydrogen on WSL/Linux from any workspace
- [ ] `npm run storybook` successfully compiles Hydrogen on Windows from any workspace
- [ ] Saving changes results in Storybook successfully recompiling without any errors on Windows from any workspace
- [ ] `npm run dev` successfully compiles Hydrogen on MacOS from any workspace
- [ ] `npm run storybook` successfully compiles Hydrogen on MacOS from any workspace
- [ ] Saving changes results in Storybook successfully recompiling without any errors on MacOS from any workspace

# 🛠️ Related issues
- resolves #3820

PR #3937, opened 2022-09-12
Low Priority - Polish Maintenance Scripts

### LINKED ISSUE #2756: Polish Maintenance Scripts
There are a few small tweaks to the maintenance scripts needed:

- [ ] The setup script still uses the "older" NPM commands style.  This might be [causing issues for Windows folks](https://talent-cloud.slack.com/archives/C0259G6KXJ6/p1652481252392069).  Let's switch to the workspaces style of command.
- [ ] There is no VS Code build shortcut for the setup script.  Let's add one.
- [ ] The setup script calls `chown -R www-data` several times.  This can cause issues on docker on Linux (and Mac?) where the user can't interact with files.  Can we get away with just change permissions and not owner?
- [ ] All the scripts reference `/var/www/html` which is where the files are in the containers.  Can we remove those to allow the scripts to run locally?  Maybe use an environment variable and fallback to `/var/www/html` if it is not set?
- [ ] The *refresh all* script calls the refresh scripts for the four workspaces all in a row which means hydrogen refresh gets called four times even though it should only be once.  I think we should have a *refresh frontend* script which one runs hydrogen once and call that from *refresh all*.

petertgiles:
This branch polishes the maintenance scripts:
- ~Adds a maint script for clean (started, could probably use more directories) h/t @patcon~
- Adds a maint script for refresh frontend (only calls `npm install` once)
- Adds VS Code build shortcuts for setup, refresh frontend, docker down, docker up ~, and clean~
- Adds php-bcmath to the maintenance container so that you can run phpunit tests
- Switches the setup and refresh scripts to use the workspace style commands
- Added `--prefer-dist` to composer install h/t @patcon 
- Change `cp .env.example .env` to use `--preserve=all` flag so that root doesn't own the file
- Change `chmod -R 775` to `a+r,a+w` so that executable bit not flipped

## Testing
Run the scripts...
- [ ] Works on Windows
- [ ] Works on Linux
- [x] Works on Mac

## Notes
- I wasn't able to get rid of `chown -R www-data` without upsetting Apache
- I wasn't able to get the script running locally without Docker because of the `chown -R www-data` :cry: 

Closes #2756


mnigh (approved):
❤️ for more shortcuts, nice work!

petertgiles:
If anyone wants to get these changes on your own branch to try refreshing whatever you're working on, here's a command to pull down the ten changed files:
```
git checkout feature/2756-v2-polish-maint-scripts --  .vscode/tasks.json  infrastructure/maintenance-container/Dockerfile  maintenance/scripts/clean.sh   maintenance/scripts/refresh_admin.sh  maintenance/scripts/refresh_api.sh  maintenance/scripts/refresh_common.sh    maintenance/scripts/refresh_frontend.sh  maintenance/scripts/refresh_indigenousapprenticeship.sh   maintenance/scripts/refresh_talentsearch.sh    maintenance/scripts/setup.sh
```
Any git gurus know a better way to do that?

patcon (commented):
Scoped it out and this looks stellar! Thanks Peter!

The one clarifying question is about the clean.sh script. What is your thought on its use-case? I ask because it appears you're trying not to touch config, whereas I recall that being the main "gotcha" in mind when it came up before :)

https://stackoverflow.com/a/46273201/504018

Original intention was to offer something foolproof in a situation where someone was otherwise tempted to do a fresh clone. gitignore'd files are usually the culprit when a reclone is needed. it was a way to get a "fresh" environment as close as possible to a new git clone, but without forcing someone to lose all their stashes and feature branches and unpushed history :)

But if you think there's still value in clean as it exists, maybe two? `npm run clean` and `npm run clean:full`? It could even copy config into a `.bak/` dir

patcon:
Also, happy to create a new issue and kick this there, as you've already got lots of good stuff in here :)

petertgiles:
> The one clarifying question is about the clean.sh script. What is your thought on its use-case?

Yeah, that's a good question.  The clean script as it is probably isn't all that useful.  I usually just rerun setup and that's enough to get to "blank slate" 99.9% of the time.  I think that running git clean inside the maintenance container probably won't work since the entire repo isn't mounted in there but I'll give it a try.  If it doesn't work I'll spin off another issue.

patcon:
> I think that running git clean inside the maintenance container probably won't work since the entire repo isn't mounted in there but I'll give it a try.


Ah, you're totally right :) I def got mixed up since I was originally thinking of it as a one-liner in npm run-scripts, rather than a bash

This might be the quickest way to get it, if you feel it fits!

```diff
diff --git a/frontend/package.json b/frontend/package.json
index 321c9a1c0..b3dce3a05 100644
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -9,6 +9,8 @@
   ],
   "scripts": {
     "tsc": "tsc --project tsconfig.base.json --noEmit",
+    "clean:full:dryrun": "git clean -dnX ..",
+    "clean:full": "npx cli-confirm 'Are you sure you want to DELETE ALL untracked and ignored files, including config? y/N' && git clean -dX ..",
     "storybook": "MERGE_STORYBOOKS=true npm run storybook --workspace=admin",
     "e2e": "npm run e2e:open",
     "e2e:open": "cypress open",
```

petertgiles:
Yeah, this is what I got:
```
fatal: not a git repository (or any parent up to mount point /var/www/html)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```
I removed the clean script from this PR and spun it off to #3954.  Definitely add your ideas there.
I also spun off #3953 for the inability to run the scripts out of container.

patcon (approved):
LGTM! Thanks for this peter!

vd1992:
the docker shortcuts seem to work fine
the directory refreshes (ie Refresh Admin) still seem to do the same thing where they spawn a folder that cannot be interacted with and some weirdness ensures

mnigh:
> the directory refreshes (ie Refresh Admin) still seem to do the same thing where they spawn a folder that cannot be interacted with and some weirdness ensures

@vd1992 are you describing what was fixed with [this PR](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3907)? if so, that work hasn't been merged into this PR's branch.

vd1992:
> > the directory refreshes (ie Refresh Admin) still seem to do the same thing where they spawn a folder that cannot be interacted with and some weirdness ensures
> 
> @vd1992 are you describing what was fixed with [this PR](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3907)? if so, that work hasn't been merged into this PR's branch.

no, that is a different Windows delight

petertgiles:
> the directory refreshes (ie Refresh Admin) still seem to do the same thing

Thanks for taking a look!  What is this issue?  It's something that happens with the current scripts already?  What folder are they spawning?

vd1992:
> > the directory refreshes (ie Refresh Admin) still seem to do the same thing
> 
> Thanks for taking a look! What is this issue? It's something that happens with the current scripts already? What folder are they spawning?

yep, an occuring issue, I just don't use Refresh scripts aside from the API one
the one we troubleshooted a few months back where the phantom folders you couldn't open would appear with names like "admin-a89ad"

Issue #3713, opened 2022-08-19
Consolidate profile queries.


petertgiles:
We currently query the employee profile in the frontend to... show the profile.   We also query the employee profile in the backend the make application snapshots.

api/app/GraphQL/Mutations/TestPoolCandidateSnapshot.graphql
frontend/talentsearch/src/js/api/userOperations.graphql

How can we consolidate these so that they don't drift out of sync?

- [Fragments?](https://www.apollographql.com/docs/react/data/fragments/)
- [Persisted Queries?](https://lighthouse-php.com/master/performance/query-caching.html#automated-persisted-queries)
- [Whatever the first option is in this stack overflow answer?](https://stackoverflow.com/a/45196861)

Acceptance Critera:
- The frontend and the backend issue employee profile snaphshot queries from the same shared place.

petertgiles:
Suggested implementation:
1) Switch [snapshot test query](https://github.com/GCTC-NTGC/gc-digital-talent/blob/5191129260baeaf895681a7894df4be7af3e7bdd/api/tests/Feature/SnapshotTest.php#L45) to `$snapshotQuery = file_get_contents(base_path('../frontend/talentsearch/src/js/api/userOperations.graphql'), true);`
1) Rename frontend query file to just "userProfile.graphql` and change from a "me" query to a "user" query

esizer:
Something to note, I think this issue I created is sort of related. Looks like codegen is recommending a new way to handle `urql` which could lead to less duplication imo.

https://github.com/GCTC-NTGC/gc-digital-talent/issues/4662

mnigh:
@petertgiles is this issue still relevant? if so, is it still blocked by #4662 or was that a soft block?

petertgiles:
Hmm, yes and no.  I put in a PHPUnit test to error out if they go out of sync so it's not as big a deal anymore.  It would be nice if they used the same file so we didn't have to sync them manually though.  I'd leave the dependency on since that migration will probably wreck whatever solution we come up with.  I'd be fine to icebox this.

esizer:
This is definitely important but will be significantly more tricky since our documents are now writing in TypeScript. 😢

mnigh:
> We currently query the employee profile in the frontend to...show the profile. We also query the employee profile in the backend the make application snapshots.
> 
> * `api/app/GraphQL/Mutations/TestPoolCandidateSnapshot.graphql`
[...]

PR #3938, opened 2022-09-12
Feature - 3661 - Application Snapshot

### LINKED ISSUE #3661: Trigger snapshot functionality during Application submission

esizer:
Resolves #3661 

## Summary

This triggers the profile snapshot laid out in #3707 

## What This Does

Adds a new event `ApplicationSubmitted` that is dispatched in the `submitApplication` mutation. `StoreApplicationSnapshot` listens to the event and stores the profile snapshot when it is dispatched.

## Testing

1. Start by clearing app cache `php artisan event:clear`
2. Ensure the event exists and has the listener `php artisan event:list`
3. Make a `PoolCandidate` that is ready to be submitted
  - Complete profile
  - All required skills
  - `DRAFT` status
  - `null` for the following
    - `submitted_at`
    - `signature`
    - `profile_snapshot`
4. Run the `submitApplication` mutation
5. Ensure the snapshot was saved with the proper `JSON`


vd1992:
😮 
crazy

do you have a linter/formatter for php files? I'm noticing lots of changes that seem to be spacing, in `PoolApplicantTest.php` to be specific 
I don't think the php files have been linted, might be a good idea to ensure everyone is on the same page for that

esizer:
> do you have a linter/formatter for php files? I'm noticing lots of changes that seem to be spacing, in `PoolApplicantTest.php` to be specific

Oh yeah, I used [`php-cs-fixer`](https://github.com/FriendsOfPHP/PHP-CS-Fixer)

api/app/Models/Skill.php
```diff
@@ -108,13 +115,18 @@ public function getExperiencesAttribute()
...
-    public function poolsNonessentialSkills() : BelongsToMany
+    public function poolsNonessentialSkills(): BelongsToMany
     {
         return $this->belongsToMany(Pool::class, 'pools_essential_skills');
     }
+
+    public function getDetailsAttribute()
+    {
+        return $this->experience_skill_pivot->details;
+    }
```
vd1992, line 131: this does not appear to work and errors 

`
Object { file: "/var/www/html/api/app/Models/Skill.php", line: 130, call: "Illuminate\\Foundation\\Bootstrap\\HandleExceptions::handleError(8, 'Trying to get property \\'details\\' of non-object', '/var/www/html/api/app/Models/Skill.php', 130, array(0))" }`

esizer: Hmm, do you have steps to reproduce? I took a chance and made a change but I'm not certain if it fixed this.

[check for skill pivot before accessing details](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3938/commits/e7dffe74ab6217ac032292a2c931217756d4201d)

vd1992: attempting to publish a pool from `/admin/pools/:id/edit` 

experience_skill_pivot appears to be null

image

image

vd1992: the latest changes fix the error
one question, what is the role of the added accessor? I don't spot any instance where getDetailsAttribute is called

esizer: It is used in the `SkillResource`

https://github.com/GCTC-NTGC/gc-digital-talent/blob/e7dffe74ab6217ac032292a2c931217756d4201d/api/app/Http/Resources/SkillResource.php#L24-L26

vd1992: ah, I  thought it was going to be explicitly declared as ->getDetails() somewhere
thanks

vd1992:
upon submit the json does save yep
is there any handy online formatter that will convert that to something a bit more legible? like split it into separate lines and get rid of newline characters
would make it easier to tell if all fields that should be present are present

esizer:
> is there any handy online formatter that will convert that to something a bit more legible? like split it into separate lines and get rid of newline characters

I have just been copy + paste it into and IDE and find and replace the escape characters. A little annoying but haven't found a better way.

vd1992:
> > is there any handy online formatter that will convert that to something a bit more legible? like split it into separate lines and get rid of newline characters
> 
> I have just been copy + paste it into and IDE and find and replace the escape characters. A little annoying but haven't found a better way.

okay 
if you delete the `\` character and then remove the wrapping `" "` that go around the outer-most `{ }` then you can use
https://codebeautify.org/jsonviewer

vd1992:
can you also check the stored fields? I'm seeing some odd nulls

image

image

esizer:
> can you also check the stored fields? I'm seeing some odd nulls

Fixed in [fix user resource language items](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3938/commits/e74aed14aa3947d3a96a3ad8f1f46acfc31fc3e0)

api/app/Http/Resources/UserResource.php
```diff
@@ -0,0 +1,79 @@
...
+        $collection = $collection->merge($educationExperiences);
+        $collection = $collection->merge($personalExperiences);
+        $collection = $collection->merge($workExperiences);
+
+        return [
+            'id' => $this->id,
+            'sub' => $this->sub,
+            'roles' => $this->roles,
+            'firstName' => $this->firstName,
+            'lastName' => $this->lastName,
```
vd1992, line 40: this too please

esizer: Updated name as well: [update user resource name attribute case](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3938/commits/843f4b75deebc223252a17e2d3c28f44c3edb46d)

Question, is there a reason we are case changing from `snake_case` in DB columns to `camelCase` in the `JSON`?

vd1992: good question, let's use the phone a friend lifeline
@petertgiles

vd1992:
just to check, all the French is stored funny due to escaping characters right? And that is easy enough to undo?

image

esizer:
> just to check, all the French is stored funny due to escaping characters right? And that is easy enough to undo?

I am not able to read the text in that screenshot but I'm going to assume yes. Some characters (we do not use in English) are stored as unicode in the database. My assumption is that you are seeing these characters.

vd1992:
> > just to check, all the French is stored funny due to escaping characters right? And that is easy enough to undo?
> 
> I am not able to read the text in that screenshot but I'm going to assume yes. Some characters (we do not use in English) are stored as unicode in the database. My assumption is that you are seeing these characters.

I find it hard to read too 😬 
I need to play around with the display settings for Adminer

vd1992 (approved):
Awesome job! Lot to wrap your head around. 
I can't think of anything else to try. I think @yonikid15 is partway through reviewing too

api/app/Http/Resources/UserResource.php
petertgiles: I think it's just a convention thing.  Convention in php/laravel is snake_case and JSON is cameCase?

petertgiles:
> And that is easy enough to undo?

I've got the PR that consumes the snapshot.  I'm using JSON.parse on the client side so as long that can deserialize it properly then we're golden.

api/app/Events/ApplicationSubmitted.php
```diff
@@ -0,0 +1,46 @@
...
+
+use App\Models\PoolCandidate;
+use Illuminate\Broadcasting\Channel;
+use Illuminate\Broadcasting\InteractsWithSockets;
+use Illuminate\Broadcasting\PresenceChannel;
+use Illuminate\Broadcasting\PrivateChannel;
+use Illuminate\Contracts\Broadcasting\ShouldBroadcast;
+use Illuminate\Foundation\Events\Dispatchable;
+use Illuminate\Queue\SerializesModels;
+use Illuminate\Support\Facades\Log;
```
petertgiles, line 13: Some unused references

api/app/Events/ApplicationSubmitted.php
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * Get the channels the event should broadcast on.
+     *
+     * @return \Illuminate\Broadcasting\Channel|array
+     */
+    public function broadcastOn()
+    {
+        return new PrivateChannel('channel-name');
```
petertgiles, line 44: Needs a real name?

api/app/Listeners/StoreApplicationSnapshot.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+namespace App\Listeners;
+
+use App\Events\ApplicationSubmitted;
+use App\GraphQL\Util\GraphQLClient;
```
petertgiles, line 6: Unused reference

petertgiles (changes requested):
A few stray things to check below.

I'm curious why you decided to build the JSON object from scratch rather than use the graphql client?  I'm worried that's going to be brittle and easily go out of sync with the frontend query.  And make you write a bunch of *Resource.php files.  :scream:

petertgiles:
Thinking about these:
> Start by clearing app cache php artisan event:clear
Ensure the event exists and has the listener php artisan event:list

Do we need to add these steps to the maintenance scripts?  Do we need to do these in server deployment?

esizer:
> I'm curious why you decided to build the JSON object from scratch rather than use the graphql client? I'm worried that's going to be brittle and easily go out of sync with the frontend query. And make you write a bunch of *Resource.php files.

While I was running tests, I had some problems with the graphql query. The largest issue was that it took a fairly long time to run in comparison. I noticed that it was taking sometimes 4-5 seconds to get a round trip. 

The resources are fairly easy to maintain now that they are setup. Unless we can find a way to automate keeping the backend and frontend queries in sync I'm not sure it is any more or less difficult. 

However, what this does give us is a significant advantage in speed and circumvent the need to run graphql queries and store data directly in the PHP.

That being said, it is not difficult to go back to making a `graphql` query and storing that instead, most of the resource code was generated and only took me about 20 mins to create them so no loss there.

esizer:
> Do we need to add these steps to the maintenance scripts? Do we need to do these in server deployment?

I don't think so, these are just cautionary. Laravel does some event caching (like its views and config) and I just put this up to make sure you had a fresh cache while testing.

api/app/Events/ApplicationSubmitted.php
esizer: Removed in [remove channels and unused imports](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3938/commits/970f5f898be17ef2a7cfbc4bacad4d332ea154b4)

api/app/Events/ApplicationSubmitted.php
esizer: We actually are not using channels so I removed this in [remove channels and unused imports](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3938/commits/970f5f898be17ef2a7cfbc4bacad4d332ea154b4)

api/app/Listeners/StoreApplicationSnapshot.php
esizer: Removed in [remove unused import](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3938/commits/c0effb3944d8e9ced544edd3d1210163823bd23e)

yonikid15 (commented):
Looking good so far Eric, just had a couple comments (looks like Vachan covered most things 👍)

api/app/Http/Resources/UserResource.php
```diff
@@ -0,0 +1,79 @@
...
+            'bilingualEvaluation' => $this->bilingual_evaluation,
+            'comprehensionLevel' => $this->comprehension_level,
+            'writtenLevel' => $this->written_level,
+            'verbalLevel' => $this->verbal_level,
+            'estimatedLanguageAbility' => $this->estimated_language_ability,
+            'isGovEmployee' => $this->is_gov_employee,
+            'hasPriorityEntitlement' => $this->has_priority_entitlement,
+            'govEmployeeType' => $this->gov_employee_type,
+            'department' => $this->department ? (new DepartmentResource(Department::find($this->department))) : null,
+            'currentClassification' => (new ClassificationResource($this->currentClassification)),
```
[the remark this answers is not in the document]
esizer: I think it has something to do with the relationship attribute matching the database column.

- **Department**
  - Column: `department`
  - Relationship: `department`
- **Classification**
  - Column: `current_classification`
  - Relationship: `currentClassification`

When creating relationships I usually create it like this to avoid conflicts:

- Column: `department_id`
- Relationship: `department`

api/app/Events/ApplicationSubmitted.php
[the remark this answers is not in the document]
esizer: Yeah, we are not using channels so this has been removed in [remove channels and unused imports](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3938/commits/970f5f898be17ef2a7cfbc4bacad4d332ea154b4)

api/app/Http/Resources/UserResource.php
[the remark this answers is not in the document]
yonikid15: Ahh gotcha 👍

yonikid15 (approved):

petertgiles:
> However, what this does give us is a significant advantage in speed and circumvent the need to run graphql queries and store data directly in the PHP.
> 
> That being said, it is not difficult to go back to making a `graphql` query and storing that instead, most of the resource code was generated and only took me about 20 mins to create them so no loss there.

Hmm, that's pretty slow...  However it is much easier to confirm that two graphql files match than it is a graphql file and a php file with eloquent commands.  I see you've got a test for comparison though.  Tomorrow I'm going to try to rebase my branch to render the snaphots on this to double-check compatibility and speed.

api/app/GraphQL/Mutations/SubmitApplication.php
```diff
@@ -114,7 +116,9 @@ public function __invoke($_, array $args)
         $application->pool_candidate_status = ApiEnums::CANDIDATE_STATUS_NEW_APPLICATION;
         $application->expiry_date = $expiryDate;
         $application->signature = $signature;
-        $application->save();
+        $success = $application->save();
```
brindasasi, line 119: what will happen if there is an error returned from the `save()` ?

esizer: It depends on the error. If it is a mysql error, an exception is thrown and we should be catching it. If there is any other sort of error, it will simply return `false`. I added this so we only dispatch the event that stores the snapshot if the save is successful with `ApplicationSubmitted::dispatchIf($success, $application);`.

petertgiles:
OK, I sat down with this and tried it out.  A few observations:
1) I'm not seeing the slowness in running the Graphql snaphsotting.  The test, including the database refresh, creating the application, submitting, and getting the snapshot twice, runs in about 1.5 seconds for me.
2) You're relying on the copied graphql file to run the test?  Isn't this three things we need to keep in sync now?  :laughing: 
2) The returned snapshot is not parsable in the frontend.  It looks like maybe its been double-escaped?
image
In the database it looks like maybe it is a single string storing escaped JSON instead of all native JSONB.  
image
3) I'm pretty sure the test is not working if it missed the double-escaped data.  :smile:   I tried dumping out the compared variables and it looks like they're both empty?
image

esizer:
Yeah, that was why I had the question about the casing earlier. If we did not have that difference we could automatically convert a profile to matching JSON without keeping anything in sync 😢 

I think maybe it might be safer to revert to the graphql query instead of using laravels built in resources. I'm not a fan of how long it takes, do you know if we have a queue we can add the event to so we don't need to wait on the graphql request to return a response? That would at least allow us to run the snapshot in the background.

> 1. I'm not seeing the slowness in running the Graphql snaphsotting.  The test, including the database refresh, creating the application, submitting, and getting the snapshot twice, runs in about 1.5 seconds for me.

Hmm, I was getting 1.54s with resources and 3.68s with the graphql. Are you using network throttling to simulate an average network request? Locally, with no throttling, the request runs much faster.

> 2\. You're relying on the copied graphql file to run the test?  Isn't this three things we need to keep in sync now?  😆

Yeah, that is a good point.

> 3\. The returned snapshot is not parsable in the frontend.  It looks like maybe its been double-escaped?

Yeah, we are storing escaped JSON. I think we can force it to store unescaped, laravel auto escapes it for safety but we can work around this.

> 4\. I'm pretty sure the test is not working if it missed the double-escaped data.  😄   I tried dumping out the compared variables and it looks like they're both empty?

Hmm, odd. That might be why it is passing lol

petertgiles (approved):
Yes!  It works properly with my branch! :partying_face: 
I would still like to switch the snapshot test to use the frontend query directly but that's going to take a bit of fiddling since the frontend uses a "me" query instead of a "user" query.  I'm fine to leave that until #3713.

PR #3942, opened 2022-09-12
Update straggler LoginIcon after @heroicon update

patcon:
Was missing this update after https://github.com/GCTC-NTGC/gc-digital-talent/pull/3849, and `npm run tsc` was giving a warning

This sort of thing should probably be caught is our linting workflow?


esizer (approved):
Great catch! 😀

PR #3956, opened 2022-09-14
All candidates page

### LINKED ISSUE #3687: Create API end point  to get the data for the All Candidates table
## Description
Create an api end point to retrieve the data for `All Candidates`  table so that admin can review the applicants. 

## Acceptance Criteria
- End point should retrieve the data for below columns
    Status
    Priority 
   Availability
   Candidate Name
   Email
   Preferred Language
  Current Location 
  Date Received 
- If the user is Admin the end point should return the above columns successfully
- If the user is NOT admin , the end point should result in error

### LINKED ISSUE #3688: Create the "All Candidates" page with the table displayed with the necessary data
Description
All candidates Page should be as displayed as in the figma
https://www.figma.com/proto/b6lDM86QCJQ6iaAh7qrTYH/Admin-v4-(Direct-Intake)?page-id=0%3A1&node-id=1870%3A3055&viewport=3469%2C7991%2C0.82&scaling=min-zoom&starting-point-node-id=1618%3A6064

A/C

1. Below Columns should be displayed 

> Status
> Priority
> Availability
> View
> Candidate Name
> Email
> Preferred Language
> Current Location
> Date Received

2. View Application doesn't do anything in this story 
3. Check box to select all rows or selected rows should work 
4. Paging should work (backend pagination)
5. Clicking on the Columns should display all the columns and selection/deselection in that pop up should reflect in the table 
6. Bread crumbs should be added as per mock up 
7. Go Back button should be added as per mockup

Question 
How about Download CSV ? Should work in this story ? A sep one ? - Yes . Same story. 
We do not need to download status & priority columns

Out of scope 
Sorting logic

JamesHuf:
Resolves #3688 and should resolve #3687.

Updates the existing pool candidate table to use backend pagination and contain different fields. Also creates a simple csv for pool candidate data.

Remaining:
- ~Breadcrumbs to page~
- ~Subtitle for pool~
- Download pdf behavior for pool candidates
- ~Review what fields should be included in pdf and csv~


api/graphql/schema.graphql
```diff
@@ -738,7 +742,8 @@ type Query {
     poolByKey(key: String! @eq): Pool @find
     pools: [Pool]! @all
     poolCandidate(id: ID! @eq): PoolCandidate @find @guard @can(ability: "view", query: true)
-    poolCandidates: [PoolCandidate]! @all(scopes: ["expiryFilter"]) @guard @can(ability: "viewAny")
+    poolCandidates(includeIds: [ID] @in(key: "id")): [PoolCandidate]! @all(scopes: ["expiryFilter"]) @guard @can(ability: "viewAny")
```
vd1992, line 745: something that comes to mind, we will probably want to remove the expiry scope from this as there will be toggled filters that allow viewing of expired candidates say for printing purposes
if the scope remains you get the oddity of being able to select the expired candidate due to the poolCandidatesPaginated query but unable to print or do anything due to the poolCandidates query auto-removing expired

JamesHuf: I've now removed the scope from the query.

frontend/common/src/components/PageHeader/PageHeader.tsx
```diff
@@ -3,11 +3,42 @@ import Heading from "../Heading";
 
 export interface PageHeaderProps {
   icon?: React.FC<{ className: string }>;
+  subtitle?: string;
```
esizer, line 6: In order to allow for `react-intl` messages this should be the type `React.ReactNode`

```suggestion
  subtitle?: React.ReactNode;
```

frontend/admin/src/js/components/poolCandidate/PoolCandidatesTable.tsx
```diff
@@ -1,217 +1,478 @@
...
+      return (
+        <span>
+          {intl.formatMessage({
+            defaultMessage: "Work Visa",
+            id: "EimWiB",
+            description: "Priority text for users with work visas",
+          })}
+        </span>
+      );
+  }
```
esizer, line 131: Hmm, if we plan on using this elsewhere could we move this to `localizedConstants.tsx` with something like:

```tsx
export const poolCandidatePriorities = defineMessages({
  [10]: {
    defaultMessage: "Priority Entitlement",
    ...
  },
  ...
});

```

esizer (changes requested):
Wow, this is looking really great! I just left two requests. Also, are we moving the print to a new issue to reduce scope or is it something you are still working on?

vd1992:
it is a bit odd to use "Work Visa" when we are using "Other" elsewhere
I went with Other based off this #3749 
image

frontend/common/src/components/PageHeader/PageHeader.tsx
JamesHuf: Changed with [Changed subtitle in PageHeader to be ReactNode](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3956/commits/7c43873984aef83e6a92a796b6b62b574513c43c)

frontend/admin/src/js/components/poolCandidate/PoolCandidatesTable.tsx
JamesHuf: Added with [Moved status text to localized constants and changed 'Work Visa' to 'Other'](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3956/commits/4bddfcc5e268de11913d07daf3ec5aca523ed483)

JamesHuf:
> it is a bit odd to use "Work Visa" when we are using "Other" elsewhere I went with Other based off this #3749 image

Updated with [Moved status text to localized constants and changed 'Work Visa' to 'Other'](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3956/commits/4bddfcc5e268de11913d07daf3ec5aca523ed483)

JamesHuf:
Also I have added print functionality now

esizer (approved):
Awesome! Looking great 👍

Issue #3947, opened 2022-09-13
✨ Read Status Overrides and updates manually added candidates


petertgiles:
## Description
Some pool candidate status are overridden based on other dimensions.  This was causing issues with the pool candidates that were manually added to the database.  Since they were never submitted with a "submitted_at" time they were always forced to draft.  These overrides were commented out in https://github.com/GCTC-NTGC/gc-digital-talent/pull/3940/commits/90477ffb2bf07672020f68f6dbab06b2bcf34121

Pseudocode of logic to re-enable:
```
if `submittedAt` is null and `isExpired` is true then force status to DRAFT_EXPIRED
if `submittedAt` is null then force status to DRAFT
if `isExpired` is true then force status to EXPIRED unless status is in (PLACED_CASUAL, PLACED_TERM, PLACED_INDETERMINATE))
```

## Acceptance Criteria
- [ ] Status overrides re-enabled
- [ ] Manually added candidates "massaged" to fit expectations

> **Note**
> This cannot be done until candidates can no longer be manually added using page on admin site.

tristan-orourke:
If we're just overriding statuses in a Laravel accessor, we'll have to be careful about using statuses to filter. We currently allow you to filter the PoolCandidates table by status, but I think that is implemented as a direct comparison to the database column.

petertgiles:
I'll try to work with Gray to better define this one.

petertgiles:
I did some analysis in prod.

```
-- if `submittedAt` is null and `isExpired` is true then force status to DRAFT_EXPIRED

select p.name, pc.pool_candidate_status, count(*) 
from pool_candidates pc
left join pools p on pc.pool_id = p.id
where pc.submitted_at is null -- not submitted
and current_date > pc.expiry_date  -- is expired
and (pc.pool_candidate_status is null or pc.pool_candidate_status <> 'DRAFT_EXPIRED') -- status is other than DRAFT_EXPIRED
group by p.name, pc.pool_candidate_status;

                                         name                                          | pool_candidate_status | count
---------------------------------------------------------------------------------------+-----------------------+-------
 {"en": "IT Technician - Help Desk", "fr": "Technicien(ne) en TI - équipe de soutien"} | DRAFT                 |   204
 {"en": "Digital Careers", "fr": "Carrières Numériques"}                               | EXPIRED               |   196
(2 rows)
```

```
-- if `submittedAt` is null and `isExpired` is not true then force status to DRAFT

[...]

PR #3957, opened 2022-09-14
Feature - 3691 - Application status form

### LINKED ISSUE #3691: Create the component to display and edit the applicant status in the snapshot page
## Description
As an admin when I click on the view application , snapshot page should open with the applicant status component so that I can review/edit the applicant status. I also want to be able to leave notes about why I changed the status and add an expiry date if it's relevant.

https://www.figma.com/proto/b6lDM86QCJQ6iaAh7qrTYH/Admin-v4-(Direct-Intake)?page-id=0%3A1&node-id=2501%3A3744&viewport=3469%2C7991%2C0.82&scaling=min-zoom&starting-point-node-id=1618%3A6064

## Screenshot

Screen Shot 2022-09-08 at 11 25 47 AM

## Acceptance Criteria
- Applicant pool status values should be pre loaded for selection. Its a `required` column. By default the status would be in `New` (for newly submitted applications)

- Pool status should be the manual values from this table  https://github.com/GCTC-NTGC/gc-digital-talent/issues/3660
          
> New (default value -  covers draft status) 
>           Application under review
>           Screened In (available)
>           Screened Out - On application
>           Under assessment
>           Screened out - On assessment
>           Qualified - Available
>           Qualified - Unavailable
>           Qualified - Withdrew
>           Placed Casual 
>           Placed Term
>           Placed Indeterminate
>           Expired

-  The "Pool Expiry Date" is the same expiry date that we have used in the past for candidates in a pool. It is optional and can always be displayed. (we will revisit this functionality when we refactor statuses, but for now, always showing this field is fine)

- The notes section is a duplicate of the notes section that is provided on the admin view of the candidate (aka applicant) profile. On that profile, there are notes for each pool the person is in and they should be displayed in an editable way here.

- Save Changes should save the applicant status, expiry date (if provided) and the notes section into the db, 

## Dev Notes
 Refer the enums added in #3744 for this implementation

esizer:
Resolves #3691 

> **Note**: Failing `e2e` test is unrelated.

## Summary

This adds the new application status form component to be used on the application page.

## Screenshot

Screen Shot 2022-09-14 at 9 41 41 AM

## TO DO

- [x] Integrate with page
- [x] Write unit tests
- [x] Write e2e tests


api/database/seeders/DatabaseSeeder.php
```diff
@@ -189,6 +190,8 @@ private function seedPoolCandidate(User $user, Pool $pool)
             // match arrays from the user
             $candidate->expectedClassifications()->sync($user->expectedClassifications()->pluck('classifications.id')->toArray());
             $candidate->cmoAssets()->sync($user->cmoAssets()->pluck('cmo_assets.id')->toArray());
+
+            $candidate->createSnapshot();
```
vd1992, line 194: minor quibble, but could this run provided `submitted_at != null`? mostly just for closer approximation to expected production results

vd1992:
I feel that there are some odd bugs and weird cases to keep an eye out for. 
In example: 
It is possible to update status to DRAFT for an application; if you navigate away it will become impossible to change it back as DRAFTs aren't pulled into the candidates table and the `submitted_at == not null` means re-submitting it will be tricky too. Essentially, creating a non-retrievable application without database intervention. 

I think the above case would probably be best resolved in the override Accessor here [link](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/api/app/Models/PoolCandidate.php#L331)
like `(status == draft && submited_at != null ) => return status = new_application `
relevant issue #3947

frontend/admin/src/js/components/poolCandidate/ApplicationStatusForm/ApplicationStatusForm.tsx
```diff
@@ -0,0 +1,267 @@
...
+            <Input
+              id="expiryDate"
+              name="expiryDate"
+              label={intl.formatMessage({
+                defaultMessage: "Pool expiry date",
+                id: "W4O1Ip",
+                description:
+                  "Label displayed on the pool candidate application form expiry date field.",
+              })}
+              type="date"
```
vd1992, line 128: minor thing, might be nicer to have it be "candidate expiry date" or something otherwise it could be confusing as you might think this changes the associated pool's expiry date instead

vd1992:
as I understand it, the toggle view is to output unformatted data in case the snapshot and page render become too out of sync
therefore, should the status form component still not exist in those cases on the side? to allow status changing to not break

image

vd1992 (changes requested):
the trail of thoughts above 😄

frontend/admin/src/js/components/poolCandidate/ApplicationStatusForm/ApplicationStatusForm.tsx
esizer: The mockup has "Pool expiry date" @Jerryescandon what are your thoughts?

Jerryescandon: 🤔 I see your point. Yeah I don't see why not. Makes sense to me. But would need to stay consistent with the previous one too. E.g. "Candidate pool status" + "Candidate expiry date"

esizer:
> as I understand it, the toggle view is to output unformatted data in case the snapshot and page render become too out of sync
> therefore, should the status form component still not exist in those cases on the side? to allow status changing to not break

I'm actually not sure what our plan with the raw snapshot is. @petertgiles do you know what the plan is for this?

esizer:
> It is possible to update status to DRAFT for an application; if you navigate away it will become impossible to change it back as DRAFTs aren't pulled into the candidates table and the `submitted_at == not null` means re-submitting it will be tricky too. Essentially, creating a non-retrievable application without database intervention.

@tristan-orourke Can we get confirmation that admins should not have the ability to return an application to the `DRAFT` status?

api/database/seeders/DatabaseSeeder.php
esizer: This has been updated with [only seed candidate snapshot if submitted](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3957/commits/dd9cbcd0dffcf8c91d50ec3d6cc7108be0398625)

frontend/admin/src/js/components/poolCandidate/ApplicationStatusForm/ApplicationStatusForm.tsx
esizer: Thanks! Labels changed in [update field labels to candidate](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3957/commits/591b9a1c83b7f5fc100b1218f472aa8f74a7e4ec)

petertgiles:
> I'm actually not sure what our plan with the raw snapshot is. @petertgiles do you know what the plan is for this?

Yeah, the raw view is in case the profile snapshot can't be well rendered in the rich view.  I think that shouldn't prevent admins from leaving notes or changing the status of the application, though.

esizer:
> Yeah, the raw view is in case the profile snapshot can't be well rendered in the rich view. I think that shouldn't prevent admins from leaving notes or changing the status of the application, though.

Hmm, okay. I will need to make some changes. Since `UserProfile` includes a sidebar we may need to remove it when the `isNavigationVisible` is false 🤔

tristan-orourke:
> > It is possible to update status to DRAFT for an application; if you navigate away it will become impossible to change it back as DRAFTs aren't pulled into the candidates table and the `submitted_at == not null` means re-submitting it will be tricky too. Essentially, creating a non-retrievable application without database intervention.
> 
> @tristan-orourke Can we get confirmation that admins should not have the ability to return an application to the `DRAFT` status?

Yeah, seems to me like Admins/Managers should be able to set status to anything _except_ DRAFT, DRAFT_EXPIRED or EXPIRED.

esizer:
> Yeah, seems to me like Admins/Managers should be able to set status to anything _except_ DRAFT, DRAFT_EXPIRED or EXPIRED.

Thank-you, this has been updated with [remove calculated statuses from select](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3957/commits/6d21f09eb0e761574112211e8cb51b0c2f369b15)

petertgiles:
Where did the view application link go?
image

esizer:
> Where did the view application link go?

It was moved to the "View" column in the middle in #3956 I believe.

frontend/admin/src/js/components/poolCandidate/ViewPoolCandidate/ViewPoolCandidatePage.tsx
```diff
@@ -197,7 +201,12 @@ export const ViewPoolCandidate = ({
           </>
         )}
       </h2>
-      {mainContent}
+      <TableOfContents.Wrapper>
```
petertgiles, line 204: I think the sidebar needs to be pulled up a bit so that the title is even with "Application's profile snapshot"
image

petertgiles (changes requested):
This is looking great!  Two small things to check..

vd1992 (approved):
amazing work as usual 😮

api/database/seeders/DatabaseSeeder.php
```diff
@@ -20,6 +20,7 @@
 use App\Models\GenericJobTitle;
 use App\Models\PersonalExperience;
 use App\Models\WorkExperience;
+use App\Http\Resources\UserResource;
```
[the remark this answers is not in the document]
esizer: Removed in [remove unused import](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3957/commits/aabf05ceb1cd14a925d42ca7172a1903cc2e9c20)

frontend/admin/src/js/components/poolCandidate/ViewPoolCandidate/ViewPoolCandidatePage.tsx
esizer: Good catch, I have updated a couple of the styles in [adjust styles to match mockup](https://github.com/GCTC-NTGC/gc-digital-talent/pull/3957/commits/6e52c7ce6ff7650d82fd9646201503eadedffdb5)

petertgiles (approved):
Awesome!

PR #3978, opened 2022-09-15
sort by newest-oldest requests on admin

### LINKED ISSUE #3975: Newest requests first on Admin request page

vd1992:
closes #3975 
a simple change of just adding a directive

testing:
open up `/admin/talent-requests`
open up `/search`
most likely the only requests are the ones seeded at basically the same time, so submit a request from the search then reload the requests table


esizer (approved):
Nice! I seeded with the requests with dates from the past year and they look to be in order 👍 

Screen Shot 2022-09-16 at 8 52 05 AM

PR #3984, opened 2022-09-16
Feature - 3976 - Location Field Required

### LINKED ISSUE #3976: Make location mandatory on search page
As a pool owner, I want to be sure talent seekers have considered where their employee will work, so I can confidently process their talent request. 

Currently too many talent seekers are leaving the location field blank and DCMO is having to follow up with some to ensure they actually have no preference when it comes to work location.

Since "telework" is one of the location options (and it's essentially the same as selecting nothing) we think that simply making this field required will address the problem.

MUST be done for both version of the search page.

### Acceptance criteria
- [ ] Location field is required on the search page
- [ ] Location field is required on the search page behind feature flag

esizer:
Resolves #3976 

## Summary

This makes the location field required on both the current and deprecated search form.

## Testing

2. Navigate to `/search`
3. Add a Classification
4. Attempt to make a request and ensure it has a error on the location field
5. Set the `FEATURE_APPLICANTSEARCH` feature flag to false in `.apache_env`
6. Rebuild the php container
    - `docker-compose stop php`
    - `docker-compose rm php`
    - `docker-compose up --build --detach`
7. Repeat steps 1-4

### DIFF
--- a/frontend/talentsearch/src/js/components/search/SearchContainer.tsx
+++ b/frontend/talentsearch/src/js/components/search/SearchContainer.tsx
@@ -97,11 +97,13 @@ export const SearchContainer: React.FC<SearchContainerProps> = ({
     applicantFilter?.expectedClassifications?.length ?? 0;
   const operationalRequirementFilterCount =
     applicantFilter?.operationalRequirements?.length ?? 0;
+  const locationPreferencesCount =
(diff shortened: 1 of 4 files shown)


mnigh (approved):
LGTM 👍

PR #3985, opened 2022-09-16
Bug - Fix Search e2e

esizer:
## Summary

Fixes the e2e test for search which was broken in #3984


petertgiles (approved):
Looks good!  Go ahead and merge once the gh tests are successful.

vd1992 (changes requested):
small thing to note, I feel the selection should pick a few location fields
mainly as over multiple seed -> test cycles we might have some occasions where the test fails due to the classification + Telework selection returning zero candidates

just trying to avoid a nuisance down the road 😛

vd1992 (approved):
sweet thanks!
it came to mind because I think we had a similar issue in the past where a test sometimes failed because filters filtered everyone out

PR #3988, opened 2022-09-16
Design - Change menu link to text if it's the current page

### LINKED ISSUE #3641: Header login link on login page is confusing
Having a link on a page that points to itself is a bit confusing to me.  I sometimes click on the link a few times, confused that I'm not going anywhere, before I realize I'm already on the page.  Having it in bold just makes me want to click it more.  I think the header links should be become just text when they're on the page they point to.

The same confusion could occur on Search, Request, and Register as well.

yonikid15:
Resolves #3641 

## Notes
Updates the TalentSearch navigation menu to use text instead of anchor tag when the current page is active.
image


petertgiles (changes requested):
Nice!  Thanks for picking up my pet peeve. :grin: Love the hidden text for screen readers, too.

It seem to work well but as soon as I log out it seems to revert to the old behavior.  Any idea why?  Here's a video of that:

https://user-images.githubusercontent.com/8978655/191263980-c928ed54-7765-460e-9e90-8fd711915426.mp4

yonikid15:
> Nice! Thanks for picking up my pet peeve. 😁 Love the hidden text for screen readers, too.
> 
> It seem to work well but as soon as I log out it seems to revert to the old behavior. 


Hmm, I'm not able to produce the same behaviour as you.. maybe try disabling cache on network requests?

https://user-images.githubusercontent.com/22059495/191322861-c368c64d-4cf9-4b86-b5cd-8e2b3750f25a.mp4

petertgiles (approved):
You're right, when I disabled cache the issue disappeared.  I was probably seeing stale versions of those pages.  Thanks!

PR #4015, opened 2022-09-20
backend sort of pool candidates with pagination

### LINKED ISSUE #3951: Pool Candidates Table: Sorting based on Status
This was split off from #3449 for the backend work.

Sorting based on Status

Here is the full list of statuses. The applicants should be displayed in order the order their status appears here (except for Draft applications which should not even be retrieved from the DB). (So "New application" should be at the top of the table and "placed casual" at the bottom unless fileters are removed).

    Draft - Never visible. This data should not be retrieved from server at all for this table.
    New application
    Application under review
    Screened in (available)
    Screened out - On application - Filtered out by default
    Under assessment
    Screened out - On assessment - Filtered out by default
    Qualified - Available
    Qualified - Unavailable
    Qualified - Withdrew - Filtered out by default
    Placed Casual
    Placed Term - Filtered out by default
    Placed Indeterminate - Filtered out by default

vd1992:
closes #3951 
candidates are sorted in the pagination query, DRAFT not retrieved 

testing:
head to `/admin/pools/:id/pool-candidates`
view the order of table rows

### DIFF
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -326,6 +326,11 @@ RAWSQL2;
         return $query;
     }
 
+    public function scopeNotDraft(Builder $query): Builder
+    {
+        return $query->whereNotIn('pool_candidate_status', ['DRAFT', 'DRAFT_EXPIRED']);
+    }
+
    /* accessor to obtain pool candidate status, additional logic exists to override database field sometimes*/
    // pool_candidate_status database value passed into accessor as an argument
    public function getPoolCandidateStatusAttribute($candidateStatus)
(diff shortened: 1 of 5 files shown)


JamesHuf:
Looks pretty good to me but I think priority sorting is supposed to happen after status sorting.

vd1992:
> Looks pretty good to me but I think priority sorting is supposed to happen after status sorting.

true
not sure why I was convinced it had to be the other way 😅

JamesHuf (approved):
Works for me!

vd1992:
> Works for me!

thanks!

PR #4018, opened 2022-09-20
FIx any users.armed_forces_status values which may have been migrated incorrectly

### LINKED ISSUE #4017: Migration can result in invalid values in the users.armed_forces_status column
`2022_08_25_165115_change_is_veteran_field.php` renamed `users.is_veteran` to `users.armed_forces_status`, and then changed the column type from Boolean to String. The result for any existing users is the column ends up with either the string "true" or "false" as its value. Those throw an error when they don't map to the expected enum values.

tristan-orourke:
Resolves #4017 

Creates a new migration, which converts any "true" or "false" strings to "VETERAN" and "NON_CAF".


JamesHuf (approved):
Works for me.

PR #4022, opened 2022-09-21
fix Chromatic workflow

vd1992:
resolves the Chromatic workflow being broken by reverting the dependency bump of TS-Loader back to 9.3.1 from 9.4.0
issue to update the dependency of TS-Loader again here #4032


esizer:
This looks good to me. Should we also open a ticket to bump this dep up again manually?

esizer (approved):
LGTM! 👍

PR #4037, opened 2022-09-21
Delete dependabot label workflow

### LINKED ISSUE #3928: Dependencies-priority not working
I suspect the logic to tag some PRs as "dependencies-priority" is not working as it seems like every dependabot PR gets this tag.  I suggest we remove it as it's just adding noise right now.  The pain point of knowing which PRs to merge first will soon be less anyway, when we start auto-merging some dependabot PRs.

vd1992:
closes #3928 
deletes the workflow


petertgiles (approved):
I have no idea how to test this but I guess we'll find out on Monday.  :laughing:

PR #4047, opened 2022-09-22
allow sorting by toggle of pool candidates

### LINKED ISSUE #3449: Implement sorting in "All candidates" table (Status, Priority should be frozen)
As a pool owner I want to view the "All Candidates" table in the pre defined sorting order so that I can review the candidates based on the priority order. 

**For discussion:**
- Status and Priority columns should display lock icon beside its name and clicking on that should not sort the table. Clicking on rest of the columns should sort the table based on that.  ( can we get the icon name? ) 
- Our current setup for Lighthouse sorting requires actual database column names.  Can it use a calculated field?  Maybe server-side sorting will need to be implemented in a different way?

Various fields on the user object cause them to be sorted in different ways.  For example, sort priority should always be given to veterans.  Add a calculated field to the user object that will encapsulate the business logic to provide a sort order that can be used for server-side sorting.

How sorting for user objects should work for "All candidates" admin page (for any given pool). This table:

### Second level sorting by Priority
~~For applicants with the same status, they should be sorted in order of "Priority". Priority is a calculated field based on three separate questions that are asked on the profile.~~ This part is complete.

This has been broken into it's own issue #3749.

### Final Sort
For applicants with the same priority oldest applications (the first one to submit their application) should be sorted to the top

vd1992:
closes #3449 
allows clicking on the column header to sort that column
this is the tertiary sort, so it will only sort among rows with same status + priority 
and a small addition to cypress to make sure something happens on clicking a column header

test: head to `/admin/pools/8364f3b1-33f9-4c9b-aa33-77bfb8c5d676/pool-candidates` and start toggling


tristan-orourke (approved):
I left one code recommendation, but it's basically a code-style change. Everything seems to work as expected.

tristan-orourke:
Double approved 👍

tristan-orourke (approved):

Issue #4035, opened 2022-09-21
Fix Cypress


petertgiles:
Cypress is experiencing many intermittent failures these days.  Suspicions are around caching causing testing of old code and the has diploma checkbox.

## Sporadic Failures

- Talent Search Workflow Tests [`cypress/integration/talentsearch/search-workflows.spec.js`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/frontend/cypress/integration/talentsearch/search-workflows.spec.js#L21)
    - [ ] searches for candidates and submits a request. [example](https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/3144207648/jobs/5110835775#step:9:543) [example2](https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/3152257539/jobs/5127326019#step:8:537) `TestingLibraryElementError: Timed out retrying after 8000ms: Unable to find an element with the text` / `expected **findAllByText(`Required diploma from post-secondary institution`)** to exist in the DOM`
- Admin Workflow `cypress/integration/admin/admin-workflows.spec.js`
    - [x] ~~Searches for a user and edits the phone number [example](https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/3144895181/jobs/5111492216#step:9:92) [example2](https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/3144947539/jobs/5111696191#step:9:92) `[GraphQL] Validation failed for the field [updateUserAsAdmin].`~~ [fixed by other PRs]
- User Profile Workflow Tests `cypress/integration/talentsearch/profile-workflows.spec.js`
    - [x] ~~Reviews a user profile and makes some edits. [example](https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/3144895181/jobs/5111492216#step:9:547) [example2](https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/3144947539/jobs/5111696191#step:9:528) `[GraphQL] Validation failed for the field [updateUserAsUser].`~~ [fixed by other PRs]
    - [ ] Reviews a user profile and makes some edits. [example](https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/3175611816/jobs/5173874502#step:8:472)
[...]

PR #4058, opened 2022-09-23
ensure Digital Careers Pool can always be applied to

vd1992:
split this out of #4035 
ensuring a pool exists that can always be applied to was the less confusing problem going on with Cypress 

evaluation: run Cypress multiple times and ensure _Direct-Intake_ spec never fails, Search spec may fail


vd1992:
view attempts here
https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/3116026323

brindasasi (approved):
It works fine .. Search page also spec didn't fail for me. Works fine  👍

PR #4080, opened 2022-09-26
Change default request policy to avoid incorrectly-cached empty lists

### LINKED ISSUE #3705: Table not refreshed on successful creation of new item (on empty list)
## Description

After creating a new item in an empty table, the user gets redirected back to the table, but the table itself does not get refreshed (cached value does not update), meaning the new item does not appear. 

urql has some documentation on [cache updates](https://formidable.com/open-source/urql/docs/graphcache/cache-updates/#updating-lists-or-links) that work involving updating the cached lists (tables) after the create mutation is finished. However, writing out these updates for every mutation seems like alot of work for such a small edge case.

Video example of the issue:
https://recordit.co/c7aWXrPao9

tristan-orourke:
Resolves #3705 

Changes the default urql [request policy](https://formidable.com/open-source/urql/docs/basics/document-caching/#request-policies) to `cache-and-network`, to make really sure we always have up-to-date data, and as an easy way to avoid [this known issue](https://formidable.com/open-source/urql/docs/basics/document-caching/#document-cache-gotchas). 

Following a [slack conversation](https://talent-cloud.slack.com/archives/C0259G6KXJ6/p1664213144566129), where we realized we could optimize certain queries to use a different request policy, I've set requests for Skills and for Classifications as examples on how to use the `cache-first` request policy where it makes sense to do so.

### DIFF
--- a/frontend/admin/src/js/components/classification/ClassificationTable.tsx
+++ b/frontend/admin/src/js/components/classification/ClassificationTable.tsx
@@ -112,14 +112,12 @@ export const ClassificationTable: React.FC<
   );
 };
 
+const context: Partial<OperationContext> = {
+  additionalTypenames: ["Classification"], // This lets urql know when to invalidate cache if request returns empty list. https://formidable.com/open-source/urql/docs/basics/document-caching/#document-cache-gotchas
+  requestPolicy: "cache-first", // The list of classifications will rarely change, so we override default request policy to avoid unnecessary cache updates.
+};
+
 export const ClassificationTableApi: React.FunctionComponent = () => {
-  const context = useMemo<Partial<OperationContext>>(
-    () => ({
-      additionalTypenames: ["Classification"], // This lets urql know when to invalidate cache if request returns empty list. https://formidable.com/open-source/urql/docs/basics/document-caching/#document-cache-gotchas
-      requestPolicy: "cache-first", // The list of classifications will rarely change, so we override default request policy to avoid unnecessary cache updates.
-    }),
-    [],
-  );
   const [result] = useGetClassificationsQuery({
     context,
   });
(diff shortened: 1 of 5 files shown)


frontend/admin/src/js/components/classification/ClassificationTable.tsx
```diff
@@ -112,7 +113,16 @@ export const ClassificationTable: React.FC<
...
 
 export const ClassificationTableApi: React.FunctionComponent = () => {
-  const [result] = useGetClassificationsQuery();
+  const context = useMemo<Partial<OperationContext>>(
+    () => ({
+      additionalTypenames: ["Classification"], // This lets urql know when to invalidate cache if request returns empty list. https://formidable.com/open-source/urql/docs/basics/document-caching/#document-cache-gotchas
+      requestPolicy: "cache-first", // The list of classifications will rarely change, so we override default request policy to avoid unnecessary cache updates.
+    }),
+    [],
+  );
```
esizer, line 122: It is best to only use `useMemo` for complex computations since this can result in a lack of garbage collection. This could also result in a performance hit since it needs to execute a function and assign values taking longer than just assigning the values.

REF: https://kentcdodds.com/blog/usememo-and-usecallback

esizer (changes requested):
Nice! Thanks for drilling into this and figuring it out. I think we could drop the `useMemo` and move the declarations outside of the component since it does not rely any internal variables/props.

frontend/admin/src/js/components/classification/ClassificationTable.tsx
tristan-orourke: You're right. I copied the use of useMemo from the [urql docs](https://formidable.com/open-source/urql/docs/basics/document-caching/#adding-typenames), but I can't imagine moving the context object outside of the component will be a problem.

esizer (approved):
Nice! Everything is working great 😁

PR #4089, opened 2022-09-27
work preferences prefill bug

### LINKED ISSUE #4086: Profile - Work Preferences form does not prefill
To recreate:
- Navigate to /talent/profile
- Click "Edit Work Preferences"
- Select Indeterminate and save
- Click "Edit Work Preferences"
- Observe that the previous selection was lost

https://images.zenhubusercontent.com/364642496/b760b2dd-2b56-4352-8cbd-c5fb88342307/simplescreenrecorder_2022_09_27_13_02_36.mp4

vd1992:
closes #4086 
the issue was that the `data.wouldAcceptTemporary` condition meant `boolToString()` only ran when the variable was true, needs to run when it is either true or false so boolean 

testing: 
ensure form prefills for either option in the "I would consider accepting a job that lasts for:" selection


esizer (approved):
Awesome! Working well for me :D

Not within scope of this issue but maybe this is a good test case!

PR #4091, opened 2022-09-27
Fix bugs with welcome page

### LINKED ISSUE #4042: Error message missing - Getting Started page
## Description
An error message is missing when logging into the Digital Talent app for the first time on the "Getting Started" page.

## Steps

1. Signed up with GCKey.
2. Linked to Getting Started page (https://talent.canada.ca/en/create-account)
3. Entered all required fields.
4. Received an "Error:creating account failed" message at the top of the screen, while my focus stayed at the bottom of the page.5. 

**Additional issue:**  If you try and toggle to French after this issue occurs, you are stuck in a never-ending loading screen.

## Expectation
I expected to be brought back up to the top of the page and provided with information regarding the error and why the account failed.  I believe it likely failed because the contact email I used was the same as a previous account I created.  Since users may switch between email account (GOC email) but still use the same contact account, this is something to consider.

### Image of "creating account failed"
Screen Shot 2022-09-22 at 2 49 00 PM

### Image of the top of page after error
Screen Shot 2022-09-22 at 2 36 46 PM

petertgiles:
This branch fixes two bugs on the welcome page:
- No user error message for duplicate email address on user update
- Changes languages on the welcome page triggers a navigation loop

This branch does not fix noted issue that the error message does not place browser focus on the right form field.  Since the error is server-side validation, implementing this would not be straightforward.

## Testing
- Try to add an existing email address on the welcome page, observe a descriptive error message in the toasts
- Try to change languages on the welcome page, should be possible without looping

Closes #4042

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -853,7 +853,13 @@ input CreateUserInput {
 When updating a User, all fields are optional
 """
 input UpdateUserAsAdminInput {
-    email: Email
+    email: Email @rules(
+        apply: ["sometimes","unique:users,email"],
+        messages: [{
+            rule: "unique"
+            message: "This email address is already in use"
+        }]
+    )
     sub: String
     roles: [Role]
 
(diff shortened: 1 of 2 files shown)


esizer (approved):
Sweet 🍭 ! Was able to toggle FR and got an appropriate error message.

Screen Shot 2022-09-28 at 9 41 52 AM

PR #4103, opened 2022-09-28
Bug - Improve unique email validation

esizer:
## Summary

This resolves an issue with not being able to update an existing user due to the unique email validation by ignoring the user you are updating.

## Testing

1. Regenerate the queries `npm run codegen --workspaces`
2. Build all projects `npm run production --workspaces`
3. Navigate tot `/admin/users/`
4. Edit a user and update field other than email
5. Ensure you can save the user
6. Edit it again, using an existing email
7. Ensure you get the unique email error
8. Navigate to `/talent/profile`
9. Go to to the "About Me" form and repeat steps 4-7
10. Logout and register as a new user
11. Attempt to create account with an existing email
12. Ensure you get the unique email error

### DIFF
--- a/api/app/GraphQL/Validators/UpdateUserInputValidator.php
+++ b/api/app/GraphQL/Validators/UpdateUserInputValidator.php
@@ -17,6 +17,15 @@ final class UpdateUserInputValidator extends Validator
         return [
             'email' => [
                 'sometimes',
+                /**
+                 * Note: Ignore the email for the user passed
+                 * in when executing the mutation so it is not
+                 * included in the unique check. This is required
+                 * for allowing a user to be updated while the email
+                 * remains the same.
+                 *
+                 * REF: https://laravel.com/docs/9.x/validation#rule-unique
+                 */
                 Rule::unique('users', 'email')->ignore($this->arg('id'), 'id'),
             ]
         ];
--- a/api/app/GraphQL/Validators/UpdateUserInputValidator.php
+++ b/api/app/GraphQL/Validators/UpdateUserInputValidator.php
@@ -17,6 +17,7 @@ final class UpdateUserInputValidator extends Validator
         return [
             'email' => [
                 'sometimes',
+                'nullable',
                 /**
                  * Note: Ignore the email for the user passed
                  * in when executing the mutation so it is not
--- a/api/app/GraphQL/Validators/UpdateUserInputValidator.php
+++ b/api/app/GraphQL/Validators/UpdateUserInputValidator.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace App\GraphQL\Validators;
+
+use Illuminate\Support\Facades\Log;
+use Illuminate\Validation\Rule;
+use Nuwave\Lighthouse\Validation\Validator;
+
+final class UpdateUserInputValidator extends Validator
+{
+    /**
+     * Return the validation rules.
+     *
+     * @return array<string, array<mixed>>
+     */
+    public function rules(): array
+    {
+
+        Log::debug($this->args->toArray());
+
+        return [
+            'email' => [
+                'sometimes',
+                Rule::unique('users', 'email')->ignore($this->arg('id'), 'id'),
+            ]
+        ];
+    }
+
+    /**
+     * Return the validation messages
+     */
+    public function messages(): array
+    {
+        return [
+            'email.unique' => 'This email address is already in use',
+        ];
+    }
+}
--- a/api/app/GraphQL/Validators/UpdateUserInputValidator.php
+++ b/api/app/GraphQL/Validators/UpdateUserInputValidator.php
@@ -2,7 +2,6 @@
 
 namespace App\GraphQL\Validators;
 
-use Illuminate\Support\Facades\Log;
 use Illuminate\Validation\Rule;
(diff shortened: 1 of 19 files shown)


petertgiles:
What bug does this fix?  How do we test it?

vd1992:
> What bug does this fix? How do we test it?

https://talent-cloud.slack.com/archives/C0259G6KXJ6/p1664382823644029

esizer:
> What bug does this fix? How do we test it?

Sorry @petertgiles . I was in such a rush to get this up I forgot to write a proper description with testing instructions.

Essentially, when we added the unique email rule for `/create-account` it was also triggering for updates and would not allow a user to update their own account since their email already existed. This ignores the user email marching the `id` passed to the mutation.

petertgiles:
> when we added the unique email rule for `/create-account` it was also triggering for updates

Oh no!  That's what I get for ignoring Cypress I guess.  :facepalm: Thanks for fixing this so quickly.

esizer:
> Oh no! That's what I get for ignoring Cypress I guess. 🤦 Thanks for fixing this so quickly.

Felt responsible since I approved it. I don't like that cypress is being marked as "skipped" rather than "failed" right now.

api/graphql/schema.graphql
```diff
@@ -433,948 +442,1223 @@ enum GenericJobTitleKey {
...
-            message: "This email address is already in use"
-        }]
+  roles: [Role] = []
+
+  # Personal info
+  firstName: String! @rename(attribute: "first_name")
+  lastName: String! @rename(attribute: "last_name")
+  email: Email
+    @rules(
+      apply: ["unique:users,email"]
```
petertgiles, line 912: Did you mean to remove the sometimes?  It now prevents more than one user from having a blank email.  At least to the API, email is an optional field.
image

api/app/GraphQL/Validators/UpdateUserInputValidator.php
```diff
@@ -0,0 +1,34 @@
...
+     * Return the validation rules.
+     *
+     * @return array<string, array<mixed>>
+     */
+    public function rules(): array
+    {
+        return [
+            'email' => [
+                'sometimes',
+                Rule::unique('users', 'email')->ignore($this->arg('id'), 'id'),
```
petertgiles, line 20: Could you add a comment to explain the ignore?  By next week I will have forgotten and it won't be clear to me.  :laughing:

esizer:
@petertgiles Sorry, I did not realize the massive diff for `schema.gql`. The important lines are:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/b0bbd904c1ec1afc10719363a6e6be418a564ee8/api/graphql/schema.graphql#L975-L977

https://github.com/GCTC-NTGC/gc-digital-talent/blob/b0bbd904c1ec1afc10719363a6e6be418a564ee8/api/graphql/schema.graphql#L1042-L1044

petertgiles (changes requested):
Nice job!  This fixes the error for me.  Two small things to check on.

api/graphql/schema.graphql
esizer: Nope, I did not mean to remove that 😆 thanks for catching that. Re-added in [readd sometimes to unique email when creating user](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4103/commits/74711c3d02ce4fff92bb7afa7d2fadec1b5cd6ad)

api/app/GraphQL/Validators/UpdateUserInputValidator.php
esizer: Yup! Hopefully what i wrote makes sense 😃 

[document ignore for unique email rule](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4103/commits/f2cbb2c152bc31ce65a0589c3cdb4865c0fc86aa)

api/graphql/schema.graphql
petertgiles: Shoot, this is still happening to me.  Is it fixed for you?

petertgiles: Oh wait, I can create a user with no email now, but I can't update a user to have no email. :confused:

esizer: Oh, I think I know what is going on. We need to prevent the email from being sent when it is empty. I assume this happens on all user forms.

esizer: Okay, it should work now with [prevent email from being passed when empty](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4103/commits/81b8c5852ef095d0011239a6b1eb126f3ee3582e)

petertgiles: Still no dice.  :cry: 
image
image

esizer: Update it to be `null` instead of `undefined` and added `nullable` to email and it looks to be working 🤞 

[make email field null when empty](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4103/commits/5a3f66969eebd65a1e960b06d8bb3d5c4bc7a805)

patcon:
Thanks a million @esizer! I think this may have resolved everything after https://github.com/GCTC-NTGC/gc-digital-talent/pull/4108 🎉

PR #4108, opened 2022-09-29
Disable docker layer caching in github action

patcon:
Spun out from: https://github.com/GCTC-NTGC/gc-digital-talent/issues/4035#issuecomment-1261116699 h/t @mnigh 

This was causing e2e-tests to show as "skipped", and rarely showing the proper pass/fail status, which makes the e2e tests pretty useless.

Disabling docker layer cache for now, to avoid the github action "out of storage" error. We'll see in the github action run whether it makes the workflow run significantly longer (20-25 min seems to be regular).

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -47,6 +47,10 @@ jobs:
       # This uses GitHub Action cache for docker container layers
       # See: https://github.com/marketplace/actions/docker-layer-caching
       - uses: satackey/action-docker-layer-caching@v0.0.11
+        # Disable for now, to avoid storage space error.
+        # #TODO investigate how to run this without error.
+        # See: https://github.com/GCTC-NTGC/gc-digital-talent/issues/4035#issuecomment-1261116699
+        if: false
         # Ignore a failure of this step and avoid terminating the job.
         continue-on-error: true
         # By default, this cache is keyed to the workflow name. We ar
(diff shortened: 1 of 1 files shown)


esizer:
The test actually ran 8 mins faster 👀  Was it the caching that was making it take so long? 😝

.github/workflows/e2e-tests.yml
```diff
@@ -47,6 +47,10 @@ jobs:
       # This uses GitHub Action cache for docker container layers
       # See: https://github.com/marketplace/actions/docker-layer-caching
       - uses: satackey/action-docker-layer-caching@v0.0.11
+        # Disable for now, to avoid storage space error.
+        # #TODO investigate how to run this without error.
```
petertgiles, line 51: Is there an issue on the board for this todo?

patcon:
@esizer Whaaaa! Bwahaha omg ok, maybe we should just remove it completely. That's wiiiild :)

Related: https://github.com/satackey/action-docker-layer-caching/issues/305

.github/workflows/e2e-tests.yml
patcon: Ended up removing this section, as it seems clear enough that this isn't helping us :)

patcon:
I wonder if lack of disk space was causing tests to overrun timeouts... cursory impression is that this seems to be less failure-prone

petertgiles (approved):
I'm not too sure how to test this other than observe the run on this PR, but the code looks good!

patcon:
Thanks @petertgiles and @mnigh!

(And yeah Peter, the CI is kinda the evaluation for this stuff, I guess? In the future, I can explicitly say that in the "testing instructions" when that's the case..!)

Issue #4112, opened 2022-09-29
Clean up rounded edges on CallToAction


patcon:
Spin-off from https://github.com/GCTC-NTGC/gc-digital-talent/pull/4049#discussion_r983459975

cc @JoshBeveridge 

Left = current
Right = corrected

![](https://user-images.githubusercontent.com/305339/193027679-88400744-3a54-4b19-91ca-855fea21f50e.png)

radius must be double the border width in order to "penetrate" into inner bevel. corner radius upped from 5px to 6px to have a clean "half" value.

If not bumped from 5=>6px, then little jaggies show up:

<img width="117" alt="Screen Shot 2022-09-29 at 3 16 03 PM" src="https://user-images.githubusercontent.com/305339/193055998-498711cb-c20e-4647-b205-f4129b0b10e9.png">

```diff
diff --git a/frontend/hydrogen.config.json b/frontend/hydrogen.config.json
index 455b305f0..35afe6dd5 100644
--- a/frontend/hydrogen.config.json
+++ b/frontend/hydrogen.config.json
@@ -363,7 +363,7 @@
         },
         {
           "key": "rounded",
-          "radius": "5px"
+          "radius": "6px"
         },
         {
           "key": "circle",
diff --git a/frontend/talentsearch/src/js/components/Home/partials/CallToAction/CallToAction.tsx b/frontend/talentsearch/src/js/components/Home/partials/CallToAction/CallToAction.tsx
index 8224fad1e..54639ee57 100644
--- a/frontend/talentsearch/src/js/components/Home/partials/CallToAction/CallToAction.tsx
+++ b/frontend/talentsearch/src/js/components/Home/partials/CallToAction/CallToAction.tsx
@@ -70,6 +70,7 @@ const CallToAction: React.FC<CallToActionProps> = ({
             data-h2-display="base(flex)"
             data-h2-height="base(100%)"
             data-h2-padding="base(x.5) p-tablet(x.25, x.4)"
+            data-h2-radius="base(6px 0 0 6px)"
             data-h2-width="base:children[svg](var(--h2-font-size-h5))"
           >
             {icon}
```

esizer:
Sorry, I think I forgot to comment. This is actually more complicated than adding some `border-radius` since the border is part of the rounded object. Adding radius actually causes more of a visual bug.

I have some ideas about how we could but I'm not 100% sure how to pull it off using hydrogen 🤷‍♀️

PR #4114, opened 2022-09-29
Added corner bevel to hover state of CallToAction component

patcon:
Resolves: https://github.com/GCTC-NTGC/gc-digital-talent/issues/4112 (see here for explanation)

### DIFF
--- a/frontend/hydrogen.config.json
+++ b/frontend/hydrogen.config.json
@@ -363,7 +363,7 @@
         },
         {
           "key": "rounded",
-          "radius": "5px"
+          "radius": "6px"
         },
         {
           "key": "circle",
--- a/frontend/talentsearch/src/js/components/Home/partials/CallToAction/CallToAction.tsx
+++ b/frontend/talentsearch/src/js/components/Home/partials/CallToAction/CallToAction.tsx
@@ -70,6 +70,7 @@ const CallToAction: React.FC<CallToActionProps> = ({
             data-h2-display="base(flex)"
             data-h2-height="base(100%)"
             data-h2-padding="base(x.5) p-tablet(x.25, x.4)"
+            data-h2-radius="base(6px 0 0 6px)"
             data-h2-width="base:children[svg](var(--h2-font-size-h5))"
           >
             {icon}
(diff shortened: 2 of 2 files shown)


esizer (approved):
Nice! 😃  Oddly enough I made the exact same change and it did not work 😖 

Anyway, I'd still wait for Josh since we are changing `hydrogen.config.json` but it looks good to me!

frontend/hydrogen.config.json
```diff
@@ -363,7 +363,7 @@
         },
         {
           "key": "rounded",
-          "radius": "5px"
+          "radius": "6px"
```
JoshBeveridge, line 366: As mentioned, this changes the `rounded` key sitewide, so I'd recommend against it for now.

frontend/talentsearch/src/js/components/Home/partials/CallToAction/CallToAction.tsx
```diff
@@ -70,6 +70,7 @@ const CallToAction: React.FC<CallToActionProps> = ({
             data-h2-display="base(flex)"
             data-h2-height="base(100%)"
             data-h2-padding="base(x.5) p-tablet(x.25, x.4)"
+            data-h2-radius="base(6px 0 0 6px)"
```
JoshBeveridge, line 73: `base(6px 0 0 6px)` => `base(rounded, 0, 0, rounded)`

JoshBeveridge (changes requested):

frontend/hydrogen.config.json
esizer: So, I did some testing today between `5px` and `6px` and in this specific use case, I think `6px` looks a bit better since they match the outer radius slightly better. I think this is because it is doubling the `3px` of the border that is has.

Screen Shot 2022-10-05 at 2 31 03 PM

patcon: Here's the extent of the differences that show up in storybook when changing "rounded" to a divisible number. Doesn't seem to be too sweeping?
https://www.chromatic.com/pullrequest?appId=61e099a1bb1465003a73d3ce&number=4114&view=changes

JoshBeveridge: Okiedoke, looks good to me then!

JoshBeveridge (approved):
Pull the lever Kronk!

patcon:
Thanks so much, @JoshBeveridge :)

Issue #3533, opened 2022-07-28
Setting up React-Router


gobyrne:
Related to #3017

Use React Router for one page, without refactoring the whole app to use it.

# Acceptance Criteria

- [ ] Profile About-Me page uses React Router to control its rendering
- [ ] Links to the about me page still work
- [ ] Links on the About Me page to pages not using react-router work
- [ ] Rest of app is unaffected
- [ ] Prompt the user for confirmation if they try to the leave the page with unsaved changes in the form.

patcon:
Was really interested in this, so clocked so evening hours and got a spike working 🎉 
https://github.com/GCTC-NTGC/gc-digital-talent/compare/main...patcon:gc-digital-talent:react-router-spike-v2

Realized it wasn't as hard to do the full admin app as I thought. Mostly  looks like additions because we can't remove all our helpers until we also migrate the other apps.

Most of the routing now happens here: (no legacy routes for now)
https://github.com/GCTC-NTGC/gc-digital-talent/compare/main...patcon:gc-digital-talent:react-router-spike-v2#diff-6cd43a91e81c996c8362d077f7d99869b2c16eafa395fe62796e3fd484a797caR125

I'm seeing how we should be able to refactor components to make more use of `useParams()` if we want :)

Auth seems really clean using `<RequireAuth />` component. I feel that stories can get more parity with live components since we can use `<MemoryRouter />` story decorators to "trick" any component into thinking it's living on a specific url with specific params. As I understand, this is a really great way to write minimal stories that don't need much special maintenance.

esizer:
Lot's of great stuff coming in `6.4.0`

[Loaders](https://reactrouter.com/en/dev/route/loader) and [Actions](https://reactrouter.com/en/dev/route/action) seems like something that could be extremely useful for us. They seem to basically be porting over `@remix/run` functionality to `react-router`

Ref: https://reactrouter.com/en/dev/start/overview

patcon:
Could prob even put this behind a feature flag, if all the id's passing into components were made optional. useLocation could be swapped conditionally in @common/helpers/router, and navigate set conditionally in components.

This task would pair well with a bunch of dead-simple cypress tests that just tested all the ways of navigating from each page. So if all the tests exercising connectivity continued to work, then we could confidently know the migration worked.

tristan-orourke:
Decided to go straight for the whole thing with #3017

PR #4119, opened 2022-09-29
Attempt at fixing search-workflows.spec.js cypress test

patcon:
Closes: #4035 

I suspect the issue may be that we just need to wait for the requests to run after each form field change, rather than blitz through a bunch of changes and then try to wait. It seems that there are maybe a few requests queued when we go to quick, and we don't properly wait for the final one before submitting the request

## What's changed
- each test now retries 3 times in CI's runMode (hides flakey tests)
  - does not retry when using console, as we want to see flakiness there
- `frontend/cypress/integration/talentsearch/search-workflows.spec.js`
  - each form change that causes a request waits for the request to finish before pushing through
  - add checks for button visibility and not being disabled
  - **forces the click even if button element was detached from DOM**
  - remove extraneous `cy.clear()` from inputs without pre-filled text

## Run History

test failures are sporadic, so going to re-run a bunch and see if failures persist.

- bcd3615bef19be44e110e6d1fb2e8e908a4b8a82 [3/4 successful](https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/3153065433)
- b9f04497888809ce8eaee2ffe2807a01866522d6 no e2e CI runs
- 8085548148c27ad4563cd4b53acdddb78b19ae6a [0/0 successful](https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/3175119578)
- ...


vd1992:
too many changes each with their own triggered API call executed almost simultaneously makes sense as a source of error
here is the history for the workflow file in question
https://github.com/GCTC-NTGC/gc-digital-talent/commits/main/frontend/cypress/integration/talentsearch/search-workflows.spec.js

maybe upon adding location selection into workflow, things got wonky once we hit 3 simultaneous form updates

vd1992:
just gonna trigger re-runs a few times

patcon:
> just gonna trigger re-runs a few times

good call! We can tag-team it until maybe 10-15 are done :)

Also, if there are still issues, we could prob be even more cautious by adding waits for each region added, instead of doing all at once, then waiting

patcon:
New error!
https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/3153065433/jobs/5129291173#step:8:532

Screen Shot 2022-09-30 at 2 10 22 PM

vd1992:
saw this one earlier

image

patcon:
Thanks @vd1992, was poking around that one too..! Documented in https://github.com/GCTC-NTGC/gc-digital-talent/issues/4035

Seems it might be related to ScrollToTop, which @esizer has raised as perhaps buggy (failure to clean-up). That sporadic error might resolve itself if we migrate to react-router (which has a component for that, presumably without this bug): https://github.com/GCTC-NTGC/gc-digital-talent/issues/3533#issuecomment-1244988917

patcon:
Working through this: https://www.cypress.io/blog/2020/07/22/do-not-get-too-detached/

patcon:
omg i think i finally sorted out what was up with our buttons not working as expected, and complaining about detachment from DOM even though it clear appears to be there in video.

(The quick-fix is just `.click({ force: true })`, as this skips cypresses ["actionability checks"](https://docs.cypress.io/api/commands/click#Options).)

But I think the bigger fix is related to trySubmitHandler:
https://github.com/GCTC-NTGC/gc-digital-talent/blob/37775d5ae3a92021d1305f13bc902b6338c68f68/frontend/talentsearch/src/js/components/search/SearchContainer.tsx#L106-L113

What led to this: In the profiler, I was seeing a final `Anonymous` state change in only the button. This is where the button is being recreated, but not in a way that cypress expects, so it doesn't not how to wait on this:

Screen Shot 2022-10-07 at 2 54 15 PM

There's no need to "fix" this now, but wondering your thoughts, to the ppl who I'm seeing in git-blame: @tristan-orourke @vd1992 @esizer. Will this tryHandleSubmit become obsoleted by some future action or migration/mitigation (e.g., react-router)? Or should we just assume the hack of `.click({ force: true })` is good enough, and not indicative of any problem or anti-pattern we care to spend time on?

Current proposal: `.click({ force: true })` with a comment or link to explanation. Maybe a comment at `tryHandleSubmit` so ppl know the consequence before replicating that pattern.

Question: Should I create another issue to change the underlying pattern of tryHandleSubmit, or just leave it as-is?

esizer:
> Will this tryHandleSubmit become obsoleted by some future action or migration/mitigation (e.g., react-router)?

I believe this is because we never actually submit the form. We are using `react-hook-form` to handle state/validation and not much else here. What is happening is:

1. Form state updated
2. If missing required fields, trigger validation otherwise continue
3. Trigger the `onSubmit` function
4. `onSubmit` navigates user to the request form with the filter state

## Possible Solution

`react-hook-form` has a way to validate on different events. What we should be doing is setting that to either "onChange" , "onBlur" or "all". This will trigger a validation of the form while filling it out instead of us needing to manually trigger it.

REF: https://react-hook-form.com/api/useform

```tsx
useForm({
  mode: 'onChange',
  ...
})
```

This would mean we no longer need the `trySubmitHandler` and instead we could just go ahead and submit the form. Though, as mention, we never actually submit a form so it might make more sense to rename `onSubmit` to something else to reduce confusion. Possible something like `onRequestCandidates` or `onNavigateRequest`?

esizer:
> Really appreciate the thoughtful comment! Mind if I spin out another issue for the point you raise above, and this one can be more about "do the tests now pass more reliably than before, and look good enough to merge"?
> 
> The main criteria here is that the tests are no longer flakey, and I think we can back out of the need for the `cy.click({ force: true })` work-around in another issue, if needed :)

Yeah, for sure! I think we already have an issue for refactoring search but I can't seem to find it :/

petertgiles (approved):
Hard to say if "everything is fixed" but it sure looks better! :grin: Thanks for helping to get our testing back on track.  I appreciate the warning comment to not duplicate the pattern while we plan to refactor it.

:heavy_check_mark: Ran search workflow 20 times locally with no failures
:heavy_check_mark: Ran entire test suite once locally with no failures

Issue #3813, opened 2022-09-01
Add Status and Priority to fancy filter on All Candidates table


gobyrne:
This issue assumes that the fancy filter is already working on the all candidates page and that we just need to add some extra filters.

The status and priority columns are important for filtering the All Candidates table. They should be added to the fancy filter on this page and some of the statuses should be filtered OUT by default.

The 5 statuses that should be filtered out by default are listed in #3449.

## Acceptance Criteria
- [ ] Filters for status and priority are functional in the fancy filter
- [ ] Some filters for status are applied by default on page load

Issue #3949, opened 2022-09-13
Migrate Pool Candidates to Users Again


petertgiles:
PR #3567 to migrate the pool candidate fields to the users table and PR #3837 to remove the ability to add more pool candidates was merged to main.  We caught the issue before it was moved to prod though: The replacement pages for adding pool candidates and direct intake were not ready yet.

The commits to back them out were:
https://github.com/GCTC-NTGC/gc-digital-talent/pull/3940/commits/00f15835c02f8c3d92a774bd10d0c46c74c5ea3c
and
https://github.com/GCTC-NTGC/gc-digital-talent/pull/3940/commits/aa0ad4a3f000be69b81e12852dfc44ed4eafe53a
in PR https://github.com/GCTC-NTGC/gc-digital-talent/pull/3940

AC:
- [ ] ~Make sure CMO has a new way to add pool candidates~
- [ ] ~Reapply the changes from #3837 and #3567~
- [x] re-add the migration

tristan-orourke:
We're planning to launch Direct Intake in our next deployment! Migrating Pool Candidates to Users needs to be part of that.

vd1992:
@tristan-orourke 
for this do you want just the migration or the removal of create/update Pool Candidate as well?

vd1992:
deployment tag present
command to use
`php artisan migrate:pool_candidate_to_user`

from

https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/api/app/Console/Commands/MigratePoolCandidatesToUser.php

PR #4129, opened 2022-09-29
Feature - Pool candidates fancy filter

### LINKED ISSUE #3875: Make fancy filter work with All Candidates table on Pool page
just like #3016

yonikid15:
Resolves #3875 and #3813

## Notes
Adds new fancy filter to Pool Candidates table. Keep in mind that the: "table already has links to both the user profile and the candidate application so it's already not homogeneous.  The way forward is to populate and filter all columns from the profile fields on the User model and not worry too much it goes out of sync with the profile snapshot." 

## Testing
To test the filter and search form you must go to the Pool Candidates table. This can be found by:
- Go to [Pool table](http://localhost:8000/en/admin/pools)
- Choose a pool and click "View Candidates"
- Start testing fancy filter 💻 

## Question
- Should we remove some of the duplicate filterByX builders in PoolCandidate as they are not being used anymore (eg. filterByCmoAssets, filterByLocationPreferences)?


vd1992:
I think general search should be operational now with https://github.com/GCTC-NTGC/gc-digital-talent/pull/4129/commits/bc330ca44d6ff09b748924a231c997169d0f8baa

vd1992:
and I think equity should be working too now with https://github.com/GCTC-NTGC/gc-digital-talent/pull/4129/commits/9edd0040657872cf9e18df3677bc3bceb5aefde9
check it out!

petertgiles (commented):
This is as far as I got before my review got derailed.  :grin:

frontend/admin/src/js/components/poolCandidate/PoolCandidateTableFilterDialog.tsx
```diff
@@ -0,0 +1,252 @@
...
+          defaultValues: activeFilters,
+        }}
+      >
+        <div data-h2-flex-grid="base(flex-start, x1, x.5)">
+          <div data-h2-flex-item="base(1of1) p-tablet(1of2) laptop(2of5)">
+            <SelectFieldV2
+              forceArrayFormValue
+              id="languageAbility"
+              label={formatMessage({
+                defaultMessage: "Languages",
```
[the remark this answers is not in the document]
yonikid15: Yea there are few columns that don't match the wireframe that I need to remove.

petertgiles:
This PR and its issues are making my head spin.  :dizzy_face: It seems like the queries that are being issues are still resolving to SQL queries with predicates on the deprecated columns in pool_candidates table, right?  It's extra confusing because the seeders are generating pool_candidates rows with completely different values than the ones on the associated users table.  I'm going to at least open an issue to update the seeders.

api/app/Models/PoolCandidate.php
```diff
@@ -280,19 +303,44 @@ public function filterByEquity(Builder $query, array $equity): Builder
...
+    public function filterByGeneralSearch(Builder $query, ?string $search): Builder
+    {
+        // used App\\Models\\User@filterByPools as reference
+        if ($search) {
+            $query->whereExists(function ($query) use ($search) {
+                $query->select('id')
+                    ->from('users')
+                    ->whereColumn('users.id', 'pool_candidates.user_id')
+                    ->where(function ($query) use ($search) {
+                        $query->where('users.first_name', "ilike", $search)
```
vd1992, line 333: should be  `"%{$search}%"`

vd1992: done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/4129/commits/1776e91b9c01c0645193ba4c446f2632bdc26923

vd1992:
added something that hopefully works to filter expected classification in https://github.com/GCTC-NTGC/gc-digital-talent/pull/4129/commits/1776e91b9c01c0645193ba4c446f2632bdc26923

to try, open graphQL playground

using query 
```
query poolCandidatesPage($input1: PoolCandidateFilterInput) {
  poolCandidatesPaginated(first: 10, where: $input1) {
    paginatorInfo {
      total
      perPage
    }
    data {
      user {
	expectedClassifications {
          group
          level
        }
      }
    }
  }
}
```
and input

```
"input1": {
    "pools": [ {"id" : "9bf8916a-5553-40ab-aec4-a49850a86577"} ],
        "expectedClassifications": [{
        "group": "IT",
        "level": 5
      }]
}
```
update pool ID to match a pool locally
update classification to whatever classification is attached to the above pool, given seeding assigns all candidates in a pool the expectedClassification of the pool, each pool has everyone associated with one expectedClassification

api/app/Models/PoolCandidate.php
yonikid15: 🙏

frontend/admin/src/js/components/poolCandidate/PoolCandidatesTable.tsx
```diff
@@ -435,18 +523,27 @@ const PoolCandidatesTable: React.FC<{ poolId: string }> = ({ poolId }) => {
...
-        onFilterChange={() => {
-          /* TODO: Implement this later */
-        }}
-        onSearchChange={() => {
-          /* TODO: Implement this later */
+        filterButton={
+          <PoolCandidateTableFilterDialog.Button
+            onSubmit={handlePoolCandidateFilterSubmit}
+          />
+        }
```
patcon, line 537: To make it clearer?

```suggestion
        filterButtonComponent={
          <PoolCandidateTableFilterDialog.Button
            onSubmit={handlePoolCandidateFilterSubmit}
          />
        }
```

yonikid15: I like that better too 👍 Fixed by [Make filter button prop name more explicit](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4129/commits/e4382eb36ac45231f297feb11da0cd691ba36f8b).

api/app/Models/PoolCandidate.php
```diff
@@ -194,17 +195,6 @@ private function filterByClassificationToSalary(Builder $query, array $classific
         return $query->whereRaw('EXISTS (' . $sql . ')', $parameters);
     }
 
-    public function filterByCmoAssets(Builder $query, array $cmoAssets): Builder
```
petertgiles, line 197: Uh, oh, I think you broke "old search".  Feature flag migration strikes again!
image

petertgiles (changes requested):
Sorry Yoni, I got pulled into other things and didn't finish my review today.  :disappointed:   I'll post what I've got so far though.  I'm worried that you may have broken old search!

frontend/admin/src/js/components/apiManagedTable/filterOperations.graphql
```diff
@@ -0,0 +1,71 @@
...
+    }
+    description {
+      en
+      fr
+    }
+    classifications {
+      group
+      level
+    }
+    status
```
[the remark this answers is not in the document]
yonikid15: You are correct! Fixed by [Remove unused fields from getFilterData query](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4129/commits/77fb04ee341563f163024ea2c7ae746ff37bc3b6).

frontend/admin/src/js/components/user/util.ts
```diff
@@ -45,3 +46,16 @@ export function stringToEnumJobLooking(
   }
   return undefined;
 }
+
+export function stringToEnumStatus(
```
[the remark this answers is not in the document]
yonikid15: Yes I can 👍 Fixed by [Make pool canadidate enum util function name more explicit](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4129/commits/79279f52dfdab3a2e748b518005341df1fa2060c).

petertgiles (approved):
I think it's all good?  :laughing:  One small suggestion below:

I think this query automatically filters out DRAFT and DRAFT_EXPIRED so I'm not sure it makes sense to leave those options in the filter since they will never return results.
image

yonikid15:
> I think it's all good? 😆 One small suggestion below:
> 
> I think this query automatically filters out DRAFT and DRAFT_EXPIRED so I'm not sure it makes sense to leave those options in the filter since they will never return results.)

Wow, that's great 😄! Oh ya good point 👍 I'll take those out.
I'll let @patcon and @tristan-orourke finish reviewing before merging.

api/app/Models/PoolCandidate.php
```diff
@@ -101,25 +101,26 @@ public function scopeClassifications(Builder $query, array $classifications): Bu
...
-                    if ($index === 0) {
-                        // First iteration must use where instead of orWhere
-                        $query->where(function ($query) use ($classification) {
-                            $query->where('group', $classification['group'])->where('level', $classification['level']);
-                        });
-                    } else {
-                        $query->orWhere(function ($query) use ($classification) {
-                            $query->where('group', $classification['group'])->where('level', $classification['level']);
-                        });
+            $query->whereHas('user', function ($query) use ($classifications) {
```
tristan-orourke, line 106: FYI, this would break the search page which is currently live in Production, since those PoolCandidates don't have users with complete profiles.
I think its safe to commit this right now, but only because we're planning on turning the APPLICANTSEARCH feature flag on and [migrating candidate data into the user objects](https://github.com/GCTC-NTGC/gc-digital-talent/issues/3949) with the next deployment.

tristan-orourke (changes requested):
Here's everything I'm not certain about, but I think I'm open to moving all of these into other tech-debt tickets.

This changes most of the scopes/filters on PoolCandidate to just be duplicates of the User filters. If we're not careful about when we deploy this commit, it will mess up Production. Also, can we do this in a way that duplicates less code? Can the PoolCandidate scopes make use of User scopes?

  
Most of the apiManagedTable seems to be very generic, but useFIlterOptions and filterOperations are specific to User filters. Can they be renamed and/or moved somewhere else?

apiManagedTable could use a story file demonstrating how to put the pieces together.

It's definitely confusing that PoolCandidateFilterInput now looks significantly different from PoolCandidateFilter and CreatePoolCandidateFilterInput. If they're going to diverge this much, they should probably be differently named. 

We should probably remove the `users` and `searchPoolCandidates` queries asap, no that they can be replaced by Paginated versions. (And also replace `applicants` with a paginated version.)

tristan-orourke (approved):
Sorry for the previous wishy-washy review - I was mostly concerned about how changes to the PoolCandidate filters might affect old requests. But I've double-checked with Gray and we're okay if old requests (using PoolCandidateFilters instead of ApplicantFilters) no longer work correctly after the next update.

Which means this is okay to merge in! I'll make `Debt` new tickets for potential improvements I noticed, but they're out of scope for now.

PR #4130, opened 2022-09-29
Pool status column not updating correctly

### LINKED ISSUE #4098: After publishing the advertisement,  pool status column did not change into "TAKING_APPLICATIONS" in dev environment.
Result of https://github.com/GCTC-NTGC/gc-digital-talent/issues/3877
After publishing the advertisement,  pool status column did not change into "TAKING_APPLICATIONS" in dev environment. 

**Description** 

In dev initial issue was https://github.com/GCTC-NTGC/gc-digital-talent/issues/3877#issuecomment-1260059218 
Once I updated the db manually to add the default value for pool_status column to be 'NOT_TAKING_APPLICATIONS"
I was able to save the advertisement. 

But now after publishing the advertisement, the status column on the dashboard did not change into "TAKING_APPLICATIONS" as it should be.

JamesHuf:
Resolves #4098.

Updates the pool tables status column to use 'advertisementStatus' instead of 'status'. Also deprecates the 'status' field on the Pool model in favor of the 'advertisementStatus' field.


frontend/admin/src/js/components/pool/PoolTable.tsx
```diff
@@ -125,7 +126,8 @@ export const PoolTable: React.FC<GetPoolsQuery & { editUrlRoot: string }> = ({
           id: "ioqFVF",
           description: "Title displayed for the Pool table status column.",
         }),
-        accessor: "status",
+        accessor: ({ advertisementStatus }) =>
+          intl.formatMessage(getAdvertisementStatus(advertisementStatus || "")),
```
petertgiles, line 130: I think this null coalesce isn't really useful.  If it's null and coalesces to an empty string the getAdvertisementStatus will throw an error anyway since that key doesn't exist.  If you do a test first you could "else" fallback to https://github.com/GCTC-NTGC/gc-digital-talent/blob/c36788f0f61bcd1da87aaa355c906773128c635d/frontend/common/src/messages/commonMessages.ts#L24 maybe?

petertgiles (commented):
I was able to publish a pool and see the status changed to "Published" in the table.
image

Do we have an interface to get it to change to _Accepting Applications_?  :confused: 

image

JamesHuf:
> I was able to publish a pool and see the status changed to "Published" in the table. image
> 
> Do we have an interface to get it to change to _Accepting Applications_? 😕
> 
> image

It's not using the poolStatus field anymore since I don't think the poolStatus is actually used anywhere else. This is now using the generated advertisementStatus field instead.

petertgiles (approved):
> It's not using the poolStatus field anymore since I don't think the poolStatus is actually used anywhere else

Ah, OK, the AC is referring to the deprecated column.  Looks good then.  Could you put on issue on the board to eventually remove the deprecated column and tag as debt?

brindasasi:
> > It's not using the poolStatus field anymore since I don't think the poolStatus is actually used anywhere else
> 
> Ah, OK, the AC is referring to the deprecated column. Looks good then. Could you put on issue on the board to eventually remove the deprecated column and tag as debt?

I'm pretty sure create & update operations still deal with this column. 
Screen Shot 2022-10-03 at 4 07 49 PM
Screen Shot 2022-10-03 at 4 07 00 PM
Let's remove them if not used in a sep ticket.

JamesHuf:
Created https://github.com/gctc-ntgc/gc-digital-talent/issues/4183 to remove the status field.

brindasasi (approved):
LGTM

PR #4131, opened 2022-09-29
Feature - 3980 - Toggle Component

### LINKED ISSUE #3980: Create toggle component
We need a new toggle component!

Jerbo: "I would imagine it is mechanically just a radio button"

## Designs:

Notes:
Might be able to use some of the theme switcher that Josh built for the hydrogen statically-generated site:
https://gitlab.com/hydrogen-css/hydrogen/-/blob/release/2.0.0-beta.28/docs/src/_includes/components/theme-switcher.njk
https://gitlab.com/hydrogen-css/hydrogen/-/blob/release/2.0.0-beta.28/docs/src/static/scripts/app.js

## AC
- [x] Works with two or more options
- [x] Works with text or icons
- [ ] Works with React Hook Forms
- [x] Has a Storybook story
- [x] Can be localized

esizer:
Resolves #3980 

## Summary

This creates a new `ToggleGroup` component that allows users to toggle a single or multiple button(s) in a group.

## Description

This is based off of the [`@radix-ui/react-toggle-group`](https://www.radix-ui.com/docs/primitives/components/toggle-group) primitive and exposes all of the same [props](https://www.radix-ui.com/docs/primitives/components/toggle-group#api-reference) as those components.

In addition to the base props, you can also pass a `Color` to the `ToggleGroup.Root` element to style it.

## Usage

```tsx
import * as ToggleGroup from "@common/components/ToggleGroup";

const ExampleToggleGroup = () => (
  <ToggleGroup.Root onValueChange={value => console.log(value)}>
    <ToggleGroup.Item value="one">One</ToggleGroup.Item>
    <ToggleGroup.Item value="two">Two</ToggleGroup.Item>
    <ToggleGroup.Item value="three">Three</ToggleGroup.Item>
  </ToggleGroup.Root>
);
```

## Screenshot

Screen Shot 2022-09-29 at 5 53 12 PM


petertgiles (commented):
This looks really, really nice.  This is the start of the radix stuff, eh?

Just reviewing the AC:
Works with two or more options - Clearly it does but maybe a story with just two options would be good since that's a common use case?
Works with text or icons - Could you add a story with icons?
Works with React Hook Forms - Not sure it is possible to prove this with just storybook.  Is it possible to do this with Jest?  Have you tried this out informally?
Has a Storybook story - :+1: 
Can be localized - Could you localize your example strings to see that?

petertgiles (approved):
:star_struck: 
image

PR #4132, opened 2022-09-29
Fix Experience Date off by 1 error

### LINKED ISSUE #4100: Experiences - Off by one day issue
Probably a UTC to local time zone issue.

tristan-orourke:
Resolves #4100.

We're already saving dates to the API and database in a sensible way, as timezone-agnostic YYYY-mm-dd strings. We just need to be careful about how we present dates in the frontend.

It seems that `new Date(date_string)` will assume `date_string` is a UTC date, but then convert it into local time. That's reasonable when working with times strings, but less so when the original string is just a date. An easy way to get around the issue is to display/format the date as if we were also in UTC timezone.


brindasasi (approved):
works just fine.

PR #4135, opened 2022-09-30
Fix e2e admin workflow

esizer:
## Summary

Due to the `e2e` tests being skipped, silently failing test was introduced in between fixes. This fixes that bug to get tests working again.

### DIFF
--- a/frontend/cypress/integration/admin/admin-workflows.spec.js
+++ b/frontend/cypress/integration/admin/admin-workflows.spec.js
@@ -49,7 +49,7 @@ describe("Admin Workflow Tests", () => {
 
     cy.findByRole("table")
       .findByRole("row", { name: /applicant test/i })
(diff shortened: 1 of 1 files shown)


frontend/cypress/integration/admin/admin-workflows.spec.js
```diff
@@ -49,7 +49,7 @@ describe("Admin Workflow Tests", () => {
 
     cy.findByRole("table")
       .findByRole("row", { name: /applicant test/i })
-      .findByText("applicant@test.com") // findByRole link doesn't work here
+      .findByRole("link", {name: /view applicant test/i}) // findByRole link doesn't work here
```
petertgiles, line 52:
```suggestion
      .findByRole("link", {name: /view applicant test/i})
```

petertgiles (approved):
Ran well for me.  I got the "diploma crash" when I ran it but otherwise looks good.

PR #4194, opened 2022-10-03
Bug/4162 archive link on archived card issue

### LINKED ISSUE #4162: Remove the link for "Archive" from Archived Applications list
- Remove the link for "Archive" in already archived applications as it is not necessary and it actually leads to an error. 

image (4)

vd1992:
resolves #4162 
adds another condition that the `archivedAt` value for an application must be null in order to render the archive action
adds a Jest test for it too

testing:
create an application that can be archived (submit application in /talentsearch then switch it to screened out in /admin)
assert application can be archived, and after archiving can no longer be archived


esizer (approved):
Nice work! Confirmed it not longer appears on archived applications.

Screen Shot 2022-10-04 at 9 10 50 AM

Issue #4195, opened 2022-10-04
Clicking a URL fragment link takes two clicks on Firefox


petertgiles:
- Testing on Firefox 105 on Linux.
- Testing on the search page.
- Clicking the first time scrolls to the top of the screen.  A second click actually scrolls to the right element.
- Is it interaction with the scrolltotop component?
- Is it interaction with the router?
- Works fine on Chromium.
- Video:

https://user-images.githubusercontent.com/8978655/193827873-12a6a779-68a1-4de5-b4c5-6cde80d965b4.mp4

PR #4196, opened 2022-10-04
Restore view results link

### LINKED ISSUE #3826: Restore View Results Link
An incomplete change was removed in https://github.com/GCTC-NTGC/gc-digital-talent/pull/3825/commits/00fff23f55b442461df0f40040c9b7c1860eb716 .  Let's restore it back.

AC:
- [ ] Translation is added
- [ ] The "only show if" logic is completed.

petertgiles:
This branch restores some partial work from the Hydrogen upgrade.  It adds a "view results" link to the new search page.

image

Testing:
- [ ] "View results" link appears on search page
- [ ] Link disappears when 0 candidates are available
- [ ] Clicking the link scrolls down to the results section

Notes:
- It seems like there is a URL fragment issue on Firefox.  I opened #4195.
- This page is still behind a feature flag so no translation needed yet.

Closes #3826

### DIFF
--- a/frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx
+++ b/frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx
@@ -78,13 +78,6 @@ const EstimatedCandidates: React.FunctionComponent<
             {candidateCount > 0 && (
               <a
                 href="#results"
-                title={intl.formatMessage({
-                  defaultMessage:
-                    "View the pools that contain matching talent.",
-                  id: "7/Bn8u",
-                  description:
-                    "The title for a link to view the pools that contain matching talent.",
-                })}
                 data-h2-color="base(dt-black) base:hover(dt-primary)"
                 data-h2-transition="base:hover(color, .2s, ease, 0s)"
                 data-h2-display="base(inline-block)"
--- a/frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx
+++ b/frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx
@@ -78,7 +78,13 @@ const EstimatedCandidates: React.FunctionComponent<
             {candidateCount > 0 && (
               <a
                 href="#results"
-                title="View the pools that contain matching talent."
+                title={intl.formatMessage({
+                  defaultMessage:
+                    "View the pools that contain matching talent.",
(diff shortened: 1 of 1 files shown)


frontend/talentsearch/src/js/components/search/EstimatedCandidates.tsx
```diff
@@ -75,6 +75,23 @@ const EstimatedCandidates: React.FunctionComponent<
                 </>
               )}
             </p>
+            {candidateCount > 0 && (
+              <a
+                href="#results"
+                title="View the pools that contain matching talent."
```
mnigh, line 81: This needs to be wrapped in `intl.formatMessage`

esizer: I think it may be better to remove the `title` attribute entirely since it has some accessibility concerns.

REF: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#accessibility_concerns

petertgiles: Doh!  Localizing was most of this PR.  :facepalm: Good catch.  Added in [Localize link title](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4196/commits/fd183c4748661b4ca6bad9d2c83c2564020f625a).

petertgiles: Ah, right, the link text is descriptive enough anyway.  Removed in [Remove title](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4196/commits/37d7d372b9e2244ec6e627b7b3e54d80c421dd0b).

mnigh:
> Looks like there is no element with the `id="results"` so it is just refreshing the page for me.

Strange, you don't have this?

Screen Shot 2022-10-04 at 12 17 50

esizer:
> Strange, you don't have this?

Oh odd, so it disappears from the DOM while loading and I guess I have very fast fingers and clicked it before it was mounted? 🤔 😆

esizer (approved):

petertgiles:
> Oh odd, so it disappears from the DOM while loading and I guess I have very fast fingers and clicked it before it was mounted?

If you're testing on Firefox you might be seeing the issue I mentioned at the top of the PR, too.

PR #4208, opened 2022-10-04
bug/4182 pool advertisement links dead

### LINKED ISSUE #4182: view pool advertisement button/link dead

vd1992:
closes #4182 
updates the link to the poster view
also cleans up a Todo or two in the same file

testing:
navigate to `/en/admin/pools/:id`
observe links

### DIFF
--- a/frontend/admin/src/js/components/pool/ViewPool.tsx
+++ b/frontend/admin/src/js/components/pool/ViewPool.tsx
@@ -112,6 +112,11 @@ export const ViewPoolPage = ({ pool }: ViewPoolPageProps): JSX.Element => {
     getSecurityClearance(pool.securityClearance ?? ""),
   );
 
+  const relativeToAbsoluteURL = (path: string): string => {
+    const { host } = window.location;
+    return `${host}${path}`;
+  };
+
   const links = [
     {
       title: intl.formatMessage({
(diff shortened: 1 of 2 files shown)


frontend/admin/src/js/components/pool/ViewPool.tsx
```diff
@@ -112,6 +112,11 @@ export const ViewPoolPage = ({ pool }: ViewPoolPageProps): JSX.Element => {
     getSecurityClearance(pool.securityClearance ?? ""),
   );
 
+  const relativeToAbsoluteURL = (path: string): string => {
+    const { host } = window.location;
+    return `${host}${path}`;
```
esizer, line 117: Not sure about this, but maybe we could add the protocol? I don't think it is required since we seem to upgrade all `http` to `https`.

```suggestion
    const { host, protocol } = window.location;
    return `${protocol}//${host}${path}`;
```

esizer (approved):
Nice! 😄  One small suggestion about the `relativeToAbsoluteURL` function but not going to let that hold this up if you disagree

frontend/admin/src/js/components/pool/ViewPool.tsx
vd1992: I have no strong feelings either way, so, sure!

PR #4209, opened 2022-10-04
Give admins access to edit pool candidates if APPLICANTSEARCH is false

tristan-orourke:
The APPLICANTSEARCH feature flag is meant to indicate that the search page will make use of Applicant Profile data entered by applicants themselves, instead of the PoolCandidate data maintained by admins. Until that feature is active, admins need direct access to the PoolCandidate models.


esizer (approved):
Awesome! Confirmed that I can see the current "Update Pool Candidate" page :D

## Tip

We also have a hook to reduce duplicated code when checking feature flags:

```tsx
import useFeatureFlags from "@common/hooks/useFeatureFlags"

const FlagControlledComponent = () => {
  const { applicantSearch } = useFeatureFlags();
  ...
};
```

tristan-orourke:
@esizer I like how the useFeatureFlag hook makes the available flags explicit! I moved the functionality into a non-hook function so I could use it outside of a component.

esizer (approved):
Nice! I like the addition of a function for feature flags 🥇

tristan-orourke:
So, with the behavior of the page changing based on a feature flag, some of the tests no longer work, depending on the feature flag. In 7638763820e5e6d6de3827a840b52cea3e283255, I added a parallel feature flag to Cypress config, which controls which tests run. Unfortunately, this has to be maintained in parallel to the original feature flag env variable.

If we [migrate to Cypress v10](https://docs.cypress.io/guides/references/migration-guide#Configuration-File-Changes), we can replace cypress.json with cypress.config.js and set the Cypress flag based on the system variable: https://github.com/cypress-io/cypress/issues/22523#issuecomment-1170212674

frontend/admin/src/js/components/poolCandidate/PoolCandidatesTable.tsx
```diff
@@ -123,6 +124,22 @@ const viewAccessor = (
...
+          adminRoutes.poolCandidateUpdate(candidate.pool.id, candidate.id),
+          intl.formatMessage({
+            defaultMessage: "Edit",
+            id: "aGEisH",
+            description:
+              "Title displayed for the Pool Candidates table Edit column.",
+          }),
+        )}
+      </span>
+    );
```
patcon, line 140: No block, but was curious where getFeatureFlags was needed and so came to be looking at this and wondering why we're not just treating viewAccessor like a component? That way, we avoid creating the sorts of functions we then need to accommodate...? e.g.,

```ts
interface ViewAccessorProps {
  candidate: PoolCandidate;
}
const ViewAccessor = ({ candidate }: ViewAccessorProps) => {
  const { applicantSearch }  = useFeatureFlags();
  const intl = useIntl();
  const adminRoutes = useAdminRoutes();

  if (!applicantSearch) {
    // ...
```

My  rough understanding is that hooks exist to not need to pass props, so I'm unclear why we'd instantiate values from hooks and pass them in.

I'm finding myself wondering if the reason for this architectural decision (accessor function instead of component) was to "allow" the original author to just keep it in the same file without violating our "one file per component" convention. Otherwise, seems it can easily be treated as a functional component? :)

Happy to see this deferred in the interest of a fast merge, but thought I'd ask in-flow of PR

tristan-orourke: Ah. I see what we were working around with these not-component functions:
Refactoring to:
```
accessor: ({ user }) => (
          <PreferredLanguageAccessor language={user?.preferredLang} />
        ),
```
results in a linting error:
>Do not define components during render. React will see a new component type on every render and destroy the entire subtree’s DOM nodes and state (https://reactjs.org/docs/reconciliation.html#elements-of-different-types). Instead, move this component definition out of the parent component “PoolCandidatesTable” and pass data as props. If you want to allow component creation in props, set allowAsProps option to true.eslint[react/no-unstable-nested-components](https://github.com/jsx-eslint/eslint-plugin-react/tree/master/docs/rules/no-unstable-nested-components.md)

esizer: So, it can be a pure function w/o the need for a hook. The important part was returning an object with keys so you don't need to know all of the feature flags.

## Before

I need to know all of the feature flags and there is no auto suggest or validating of the feature flag so a typo causes it to always be false.

```tsx
const applicantSearch = checkFeatureFlag("FEATURE_APPLICANTSEARCH"); // True
const applicantSearch = checkFeatureFlag("APPLICANTSEARCH"); // False + no hinting as to why
```

## After

With the hook, we get some auto suggest for what feature flags we have available to check and it will error if you use one that does not exist.

```tsx
const flags = useFeatureFlags();
const { applicantSearch } = flags; // Okay 
const { applicantSearch } = flags; // Error 
```

patcon: Ah, if it's complex to resolve as the warning recommends, I'll retract :)

frontend/cypress/integration/admin/pool-candidate.spec.js
```diff
@@ -20,93 +20,118 @@ describe("Pool Candidates", () => {
...
+      aliasQuery(req, "getUpdatePoolCandidateData");
+      aliasMutation(req, "updatePoolCandidate");
     });
 
     loginAndGoToPoolsPage();
   });
 
-  it("should update pool candidate status", () => {
-    cy.wait("@gqlgetPoolsQuery");
+  if (Cypress.env("FEATURE_APPLICANTSEARCH")) {
```
patcon, line 32: Just a heads up to others that adding `?w=1` to the end of any diff-related url allows for review of diffs that ignore whitepace (just like how `-w` works on CLI).

https://github.com/GCTC-NTGC/gc-digital-talent/pull/4209/files?w=1

frontend/cypress/integration/admin/pool-candidate.spec.js
```diff
@@ -20,93 +20,118 @@ describe("Pool Candidates", () => {
...
+    it("should update pool candidate status", () => {
```
patcon, line 33: Maybe just make it more clear in the test output which ones ran/passed/failed?

```suggestion
    it("should update pool candidate status (FEATURE_APPLICANTSEARCH:on)", () => {
```

frontend/cypress/integration/admin/pool-candidate.spec.js
```diff
@@ -20,93 +20,118 @@ describe("Pool Candidates", () => {
...
 
-    cy.findByRole("button", { name: /save changes/i })
-      .click();
+      cy.wait("@gqlUpdatePoolCandidateStatusMutation");
 
-    cy.wait("@gqlUpdatePoolCandidateStatusMutation")
+      cy.expectToast(/pool candidate status updated successfully/i);
+    });
+  } else {
+    it("should edit and update pool candidate", () => {
```
patcon, line 107:
```suggestion
    it("should edit and update pool candidate (FEATURE_APPLICANTSEARCH:off)", () => {
```

patcon (approved):
This looks good to me. Small stretch if you feel like it... (next comment)

patcon:
One small thing if you feel like it:

I recognize that this was just the format of the rest of the test, but wanted to offer that this might be a little more clear to read and for the test to be legible (e.g. "why are there multiple matching of which first is taken?")

```js
      cy.findAllByRole("row").eq(0).within($row => {
        cy.wrap($row).findByRole("link", { name: /view candidates/i }).click();
      });
```

Once we start creating new pools for each test (when we have that command, `.eq(0)` can be replaced with another filter to match the new pool :)

No pressure to defer this to later

PR #4216, opened 2022-10-05
Bug - 4212 - Scroll to hash on load

### LINKED ISSUE #4212: Firefox doesn't navigate to URLs with fragments properly
I opened issue https://github.com/GCTC-NTGC/gc-digital-talent/issues/4195 because an `a` tag with an href to a URL fragment had to be clicked twice.  Eric kindly made a new component in https://github.com/GCTC-NTGC/gc-digital-talent/pull/4210 that bypassed the issue.  Unfortunately, pasting the URL with a fragment directly in the URL bar still exhibits the behavior.

https://images.zenhubusercontent.com/364642496/d5ac30d5-c274-42d8-a6c3-3b36a27e20b8/simplescreenrecorder_2022_10_04_16_29_08.mp4

esizer:
Resolves #4212 

## Summary

This fixes an issue with the page not scrolling to a hash on page load.

## Details

Our `<ScrollToTop />` component was forcing the page to the top on every page load. This adds some code to check for a hash then wait for the matching element to exist in the DOM before scrolling to it.

## Testing

> **Note**
> Please test in as many browsers you have available and note the browser and version you were using when you encounter any issues.

1. Build talentsearch `npm run production --workspace=talentsearch`
2. Navigate to `/search#results`
3. Ensure it scrolls to the results element

### DIFF
--- a/frontend/common/src/components/Link/ScrollToLink.tsx
+++ b/frontend/common/src/components/Link/ScrollToLink.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { HISTORY } from "../../helpers/router";
+import { navigate } from "../../helpers/router";
 
 export interface ScrollToLinkProps
   extends Omit<React.HTMLProps<HTMLAnchorElement>, "href" | "onClick"> {
@@ -19,7 +19,7 @@ const ScrollToLink = ({ to, children, ...rest }: ScrollToLinkProps) => {
   const handleClick = (e: React.MouseEvent<HTMLAnchorElement>) => {
     if (targetSection) {
       e.preventDefault();
-      HISTORY.push(`#${to}`);
+      navigate(`#${to}`);
       targetSection.scrollIntoView({
         behavior: "smooth",
         block: "start",
--- a/frontend/common/src/components/Link/ScrollToLink.tsx
+++ b/frontend/common/src/components/Link/ScrollToLink.tsx
@@ -1,4 +1,5 @@
 import React from "react";
+import { HISTORY } from "../../helpers/router";
 
 export interface ScrollToLinkProps
   extends Omit<React.HTMLProps<HTMLAnchorElement>, "href" | "onClick"> {
@@ -18,6 +19,7 @@ const ScrollToLink = ({ to, children, ...rest }: ScrollToLinkProps) => {
   const handleClick = (e: React.MouseEvent<HTMLAnchorElement>) => {
     if (targetSection) {
       e.preventDefault();
+      HISTORY.push(`#${to}`);
       targetSection.scrollIntoView({
         behavior: "smooth",
         block: "start",
--- a/frontend/common/src/helpers/router.tsx
+++ b/frontend/common/src/helpers/router.tsx
@@ -238,17 +238,61 @@ export function useQueryParams() {
   return parseUrlQueryParameters(location);
 }
 
-export const ScrollToTop: React.FC<{ children: React.ReactElement }> = ({
-  children,
-}): React.ReactElement => {
-  useEffect(() => {
-    window.scrollTo({
-      top: 0,
-      left: 0,
-      behavior: "auto",
+/**
+ * Wait for Element
+ *
+ * @param {string} selector The element sector
+ * @returns {Promise<Element | null>}  A promise containing the element if found, null if not
+ */
+const waitForElm = (selector: string): Promise<Element | null> => {
+  return new Promise((resolve) => {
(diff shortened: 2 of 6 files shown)


frontend/common/src/components/Link/ScrollToLink.tsx
```diff
@@ -18,6 +19,7 @@ const ScrollToLink = ({ to, children, ...rest }: ScrollToLinkProps) => {
   const handleClick = (e: React.MouseEvent<HTMLAnchorElement>) => {
     if (targetSection) {
       e.preventDefault();
+      HISTORY.push(`#${to}`);
       targetSection.scrollIntoView({
         behavior: "smooth",
         block: "start",
```
mnigh, line 25: Any reason you removed smooth scroll in the [router](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4216/commits/1fdaa7241021f72d0c372337db47abfde8185bfd) but not here?

mnigh (commented):
Tested on macOS 12.6:
- Google Chrome Version 106
- Microsoft Edge Version 106
- Firefox Version 105

Page scrolls to a hash on page load. 🥳 

Happy to approve once my question about the [scroll `behavior` difference](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4216/files#r988095524) is addressed.

vd1992:
what's up with the checks? they aren't running?

frontend/common/src/components/Link/ScrollToLink.tsx
esizer: Smooth scroll is nice when you are clicking a link but looked kinda janky when it was happening on load. Open to changing if others prefer it on load.

petertgiles:
Tested on Linux.  On both Firefox and Chromium if I just paste "http://localhost:8000/en/search#results" in my URL it scrolls to the right div and then almost immediately something hydrates, I think the skill picker, and then I'm on the wrong spot on the page again.  So technically it does scroll on load but effectively I can't send someone a link to this URL fragment and have them end up at the right spot.  :cry: 

https://user-images.githubusercontent.com/8978655/194115672-b7707ba0-a8b2-4346-afea-0f850b811d27.mp4

esizer:
> Tested on Linux. On both Firefox and Chromium if I just paste "http://localhost:8000/en/search#results" in my URL it scrolls to the right div and then almost immediately something hydrates, I think the skill picker, and then I'm on the wrong spot on the page again. So technically it does scroll on load but effectively I can't send someone a link to this URL fragment and have them end up at the right spot. 😢

Dang layout shift! Thanks, taking a look at this now. 👀

esizer:
> > Tested on Linux. On both Firefox and Chromium if I just paste "http://localhost:8000/en/search#results" in my URL it scrolls to the right div and then almost immediately something hydrates, I think the skill picker, and then I'm on the wrong spot on the page again. So technically it does scroll on load but effectively I can't send someone a link to this URL fragment and have them end up at the right spot. 😢
> 
> Dang layout shift! Thanks, taking a look at this now. 👀

Okay, so it does not *completely* resolve the issue but it does reduce layout shift and makes it less noticeable. Since we are making 4 different queries on page load and what renders changes based on results we can not avoid all layout shit just try to manage it.

It should be a bit better with [add pending to search container to reduce cls](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4216/commits/4fd20eb41dc8b18e64d8f3856e1875d2935ea5f8)

frontend/common/src/components/Link/ScrollToLink.tsx
```diff
@@ -18,6 +19,7 @@ const ScrollToLink = ({ to, children, ...rest }: ScrollToLinkProps) => {
   const handleClick = (e: React.MouseEvent<HTMLAnchorElement>) => {
     if (targetSection) {
       e.preventDefault();
+      HISTORY.push(`#${to}`);
```
petertgiles, line 22: Should maybe use `navigate` from router.tsx instead of HISTORY.push directly?

petertgiles (approved):
Yeah, that's a bit better

frontend/common/src/components/Link/ScrollToLink.tsx
esizer: I did not realize that is what `navigate` was doing 😳 

Changed in: [use navigate in favour of history](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4216/commits/b5ed0942be7e329ffdffb2da691b9df0126ead30)

PR #4217, opened 2022-10-05
Debt - 4204 - Update lodash imports

### LINKED ISSUE #4204: Update lodash imports
We try to always import lodash functions from the specific file needed to keep bundle sizes down.  However, a few top level imports have snuck in that need to be fixed up.

image

Ref:
https://stackoverflow.com/a/35251059
https://github.com/GCTC-NTGC/gc-digital-talent/pull/1673

esizer:
Resolves #4204 

## Summary

This updates all named imports of lodash to use specific paths to keep bundle size down.

## Example

```tsx
// Previous
import { flatMap, uniqBy } from "lodash";

// New
import flatMap from "lodash/flatMap";
import uniqBy from "lodash/uniqBy";
```

## Testing

Confirm there is no instance of `import ... from "lodash";`


petertgiles (approved):
Nice!  I wish Actions were running but it seems good to me.

PR #4218, opened 2022-10-05
Bug - 4203 - Remove data-h2 from react-helmet

### LINKED ISSUE #4203: Admin - Loader appears unstyled when in French
## Description
When navigating from page to page in Admin, the loader appears unstyled when the language is set to French.

## Screenshot
Screen Shot 2022-10-04 at 12 23 40

## Screen Recording
https://user-images.githubusercontent.com/3046459/193875019-bfaee736-b2b2-4292-aa41-84e4dd41bc06.mov

## Acceptance Criteria
- [ ] loader appears the same for English and French

esizer:
Resolves #4203 

## Summary

This removes the `data-h2` attribute from `react-helment` components so it is always on the `html` element.

## Details 

We had `data-h2` on the html and within `<Helmet />` . This allowed `react-helmet` to take responsibility for it causing it to be removed while it figured out the locale. Removing it from `<Helmet />` gave the `index.html` responsibility and it never gets removed now.

## Testing

1. Build admin `npm run production --admin`
2. Navigate to `/admin`
3. Navigate to different pages and ensure hydrogen styles are always applied to the `<Loading />` spinner

### DIFF
--- a/frontend/admin/src/js/components/PoolDashboard.tsx
+++ b/frontend/admin/src/js/components/PoolDashboard.tsx
@@ -8,7 +8,6 @@ import { AuthenticationContext } from "@common/components/Auth";
 import { getLocale } from "@common/helpers/localize";
 import { checkFeatureFlag } from "@common/helpers/runtimeVariable";
 
-import Pending from "@common/components/Pending";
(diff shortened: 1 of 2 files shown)


mnigh (approved):
🥳 👍

PR #4233, opened 2022-10-06
feature/4201 pool card component

### LINKED ISSUE #4201: Build Pool Card component
Build the "Card" view of Pools, as a component. The rest of the Browse IT Jobs page is still in flux, but we can build the Card view based on the following designs.

[Figma Prototype](https://www.figma.com/file/NhEMp9tiBNi5VwuhmRr4JW/Browse-jobs%2C-Support-and-FAQs?node-id=11%3A717)
Browse IT jobs

## Acceptance Criteria
- [ ] a Storybook story which, when passed a Pool object, display's the Card view.
- [ ] The list of links on the Browse Opportunities page should be replaced with a list of Cards
- [ ] The "Apply to this recruitment" button should be a link to `/pools/:poolId`

vd1992:
closes #4201 
adds the Pool Card component and updates `/browse/pools` to render cards

testing:
view storybook build
navigate to `/browse/pools` locally

did some real environmentally friendly coding (I recycled code) thank you Eric and Pat for handy code


vd1992:
rationale for test deletion in https://github.com/GCTC-NTGC/gc-digital-talent/pull/4233/commits/8505ebaa757ca098e935d0b21abfcc190d8a2c57
redundant test, done in `/submit-application-workflows.spec.js`

frontend/talentsearch/src/js/components/browse/PoolCard.tsx
```diff
@@ -0,0 +1,163 @@
...
+        data-h2-padding="base(x.75)"
+      >
+        <h6
+          data-h2-font-weight="base(700)"
+          data-h2-padding="base(0, 0, x0.75, 0)"
+        >
+          {poolClassificationString()
+            ? poolClassificationString()
+            : intl.formatMessage({
+                defaultMessage: "No Classification",
```
brindasasi, line 79: if the classification is null , the pool card should not be displayed.

frontend/talentsearch/src/js/components/browse/PoolCard.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+} as Meta;
+
+const TemplatePoolCard: Story<{ pool: PoolAdvertisement }> = () => (
+  <div>
+    <p data-h2-padding="base(x0.5, 0, x0.5, 0)">First</p>
+    <PoolCard pool={fakedPool} />
+    <p data-h2-padding="base(x0.50, 0, x0.5, 0)">Second</p>
+    <PoolCard pool={fakedPool2} />
+    <p data-h2-padding="base(x0.5, 0, x0.5, 0)">Null</p>
+    <PoolCard pool={nullPool} />
```
brindasasi, line 33: I don't think this is a valid scenario. We don't want to display the null pool card to the applicant.  We can remove this.

brindasasi (changes requested):
Everything looks perfect. Only concern related to null pool. Rest looks fine to me.

brindasasi (changes requested):
Checked with @Jerryescandon on this null pool 
Please implement like below and its french equivalents when the corresponding elements are null. 

Salary range: (Needs classification)
Key skills: (No skills required)
Apply by: (To be determined)

vd1992:
> Checked with @Jerryescandon on this null pool Please implement like below and its french equivalents when the corresponding elements are null.
> 
> Salary range: (Needs classification) Key skills: (No skills required) Apply by: (To be determined)

aye aye. done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/4233/commits/d2f4e797cca0ecabe4f517ca389d5ece1434b9bd

brindasasi (approved):
LGTM

frontend/talentsearch/src/js/components/browse/PoolCard.tsx
```diff
@@ -0,0 +1,179 @@
...
+      if (classification.minSalary && !classification.maxSalary) {
+        return `$${classification.minSalary} -`;
+      }
+      if (!classification.minSalary && classification.maxSalary) {
+        return `- $${classification.maxSalary}`;
+      }
+      return `$${classification.minSalary} - $${classification.maxSalary}`;
+    }
+    return null;
+  };
```
esizer, line 43: ## Tip

You could use the helper to get the formatted salary which also does some localization for us as well.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/855abba85d6de4d4197c4a0984f2adcecfe64901/frontend/common/src/helpers/localize.ts#L84-L106

esizer (changes requested):
Awesome! Looking great, just a small suggestion for the salary range.

frontend/talentsearch/src/js/components/browse/PoolCard.tsx
vd1992: yep https://github.com/GCTC-NTGC/gc-digital-talent/pull/4233/commits/74bd2f4fbb08dd759ac22a492198f2e6b19bb196
need to read through all the helpers/util stuff at some point

frontend/talentsearch/src/js/components/browse/PoolCard.tsx
```diff
@@ -0,0 +1,178 @@
...
+    >
+      <div
+        className="card__body"
+        data-h2-background-color="base(dt-white)"
+        data-h2-padding="base(x.75)"
+      >
+        <h6
+          data-h2-font-weight="base(700)"
+          data-h2-padding="base(0, 0, x0.75, 0)"
+        >
```
esizer, line 74: This may lead to some a11y issues related to the heading ranks. I'd suggest using the `Heading` component and drilling a level prop down.

```tsx
import Heading, { type HeadingLevel } from "@common/components/Heading";

...
interface PoolCardProps {
  ...
  headingLevel?: HeadingLevel;
}

// Note: You can set the default to w/e we will use most commonly (I think for the first use it is h3, though
const PoolCard = ({ pool, headingLevel = "h3" }: CardProps & React.HTMLProps<HTMLDivElement>) => {
  ...
    <Heading level={headingLevel} size="h6" data-h2-margin="base(0, 0, x.75, 0)">
      ...
    </Heading>
...
```

REF: https://www.w3.org/WAI/tutorials/page-structure/headings/#:~:text=Heading%20ranks,-Nest%20headings%20by&text=The%20most%20important%20heading%20has,of%20the%20higher%20ranked%20section.

frontend/talentsearch/src/js/components/browse/PoolCard.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const locale = getLocale(intl);
+  const paths = useDirectIntakeRoutes();
+
+  const poolClassificationString = (): string | null => {
+    if (pool.classifications && pool.classifications[0]) {
+      const classification = pool.classifications[0];
+      return `${classification.group}-0${classification.level}`;
+    }
+    return null;
+  };
```
esizer, line 30: Not a *huge* deal, but I think we could move this outside of the component and make it a pure function so we aren't executing it twice later on.

```tsx
const poolClassificationString = (
  classifications: Maybe<Maybe<Classification>[]>,
): string | null => {
  if (!classifications || !classifications[0]) return null;

  const classification = classifications[0];

  return `${classification.group}-0${classification.level}`;
};

...

const classification = poolClassificationString(pool.classifications);

...

{classification || intl.formatMessage(...)}
```

frontend/talentsearch/src/js/components/browse/PoolCard.tsx
```diff
@@ -0,0 +1,178 @@
...
+    if (pool.classifications && pool.classifications[0]) {
+      const classification = pool.classifications[0];
+      return localizeSalaryRange(
+        classification.minSalary,
+        classification.maxSalary,
+        locale,
+      );
+    }
+    return undefined;
+  };
```
esizer, line 42: See previous comment about the classification.

frontend/talentsearch/src/js/components/browse/PoolCard.tsx
```diff
@@ -0,0 +1,178 @@
...
+          <span>
+            <Chips>
+              {poolEssentialSkillKeys()
+                ? poolEssentialSkillKeys()
+                : intl.formatMessage({
+                    defaultMessage: "(No skills required)",
+                    description: "No skills required",
+                    id: "WbET08",
+                  })}
+            </Chips>
```
esizer, line 134: `<Chips>` should be moved within the check since it contains `role=list` which should not be empty.

```tsx
{essentialSkills ? <Chips>{essentialSkills}</Chips> : intl.formatMessage(...)}
```

frontend/talentsearch/src/js/components/browse/PoolCard.tsx
```diff
@@ -0,0 +1,178 @@
...
+            type="button"
+            mode="outline"
+            color="black"
+          >
+            {intl.formatMessage({
+              defaultMessage: "Apply to this recruitment",
+              description:
+                "Message on link that say to apply to a recruitment advertisement",
+              id: "zvn+Gu",
+            })}
```
esizer, line 170: Just so we have a bit more context for this link, can we add a screen reader only part to the end of this string?

```tsx
{intl.formatMessage({
  defaultMessage: "Apply to recruitment<hidden> {name}</hidden>",
  description:
    "Message on link that say to apply to a recruitment advertisement",
  id: "zvn+Gu",
}, {
  name: cardTitle
})}
```

esizer (changes requested):
I'm so sorry, these should requests should have been part of my initial review 😞

frontend/talentsearch/src/js/components/browse/PoolCard.tsx
vd1992: okay https://github.com/GCTC-NTGC/gc-digital-talent/pull/4233/commits/d0b44396a67bc8e7b787cb643374022bc7c68a15

frontend/talentsearch/src/js/components/browse/PoolCard.tsx
vd1992: https://github.com/GCTC-NTGC/gc-digital-talent/pull/4233/commits/1ce95b75f58cdecd76253fc5d31dd4c84f5a5beb

frontend/talentsearch/src/js/components/browse/PoolCard.tsx
vd1992: https://github.com/GCTC-NTGC/gc-digital-talent/pull/4233/commits/1ce95b75f58cdecd76253fc5d31dd4c84f5a5beb

frontend/talentsearch/src/js/components/browse/PoolCard.tsx
vd1992: https://github.com/GCTC-NTGC/gc-digital-talent/pull/4233/commits/eb329dd298f5247f94f396666c3a054c773f889e

esizer (approved):
Looking good 👍

PR #4246, opened 2022-10-07
Fixes a bug with the weight of button text on the new homepage

JoshBeveridge:
# Purpose
Fixes a minor visual bug @BrookeCameron noticed with the buttons on the homepage being the wrong weight after the components were merged. Solved by adding an optional weight prop set to "bold".


esizer (approved):
Awesome! 💯

esizer (approved):
👍

PR #4248, opened 2022-10-07
a11y - 4247 - Remove nested h2

### LINKED ISSUE #4247: Edit Pool Page - h2 cannot be a child of h2
image

esizer:
Resolves #4247 

## Summary

This removes the nested `<h2> elements on the `EditPool` page.

## Testing

2. Navigate to a pool `/admin/pools/{poolId}/edit`
3. Confirm there are no nested `<h2>` elements or errors in the console

### DIFF
--- a/frontend/admin/src/js/components/pool/EditPool/AssetSkillsSection.tsx
+++ b/frontend/admin/src/js/components/pool/EditPool/AssetSkillsSection.tsx
@@ -58,8 +58,8 @@ export const AssetSkillsSection = ({
 
   return (
     <TableOfContents.Section id={sectionMetadata.id}>
-      <TableOfContents.Heading>
-        <h2 data-h2-margin="base(x3, 0, x1, 0)">{sectionMetadata.title}</h2>
+      <TableOfContents.Heading data-h2-margin="base(x3, 0, x1, 0)">
+        {sectionMetadata.title}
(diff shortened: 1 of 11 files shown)


vd1992:
are the font changes desired per Storybook?
https://www.chromatic.com/build?appId=61e099a1bb1465003a73d3ce&number=3493

vd1992 (commented):
are the font changes desired per Storybook?
https://www.chromatic.com/build?appId=61e099a1bb1465003a73d3ce&number=3493

esizer:
> are the font changes desired per Storybook?

I think so but I'm not certain. It is now using the styles defined in the `Heading` component where before it was using browser defaults.

@JoshBeveridge should these styles be updated or are they the correct ones? I think what is happening is that even though they are supposed to structurally be `<h2>` we want to present them as `<h3>`, am I correct in that assumption?

https://github.com/GCTC-NTGC/gc-digital-talent/blob/8b55555169f629bc9a8001ec5384291d52f448f2/frontend/common/src/components/Heading/styles.ts#L8-L12

JoshBeveridge:
@esizer Yeah, this happens in a handful of places, because the default font size of the pertinent heading might be too visually overwhelming for the context, despite the semantics requiring that level.

tristan-orourke (approved):
LGTM

PR #4265, opened 2022-10-11
a11y - 3836 - Remove nested labels, checkbox

### LINKED ISSUE #3836: Form field must not have multiple label elements
## Summary

Multiple label elements is not widely supported in assistive technologies. Ensure the first label contains all necessary information.

## Locations

https://github.com/GCTC-NTGC/gc-digital-talent/blob/c8ff1a196d4275a58269f4273774e165366ee676/frontend/common/src/components/form/Checkbox/Checkbox.tsx#L69-L101

esizer:
Resolves #3836 

## Summary

This removes the wrapping `<label>` element which was providing a second label to the checkbox when `boundingBox={true}` .

## Testing

1. Build talentsearch `npm run production --workspace=talentsearch`
2. Navigate to the following experience forms and confirm is only one label for the bonding box checkboxes
    - Community
    - Education
    - Personal
    - Work

### DIFF
--- a/frontend/admin/src/js/components/pool/EditPool/EditPool.tsx
+++ b/frontend/admin/src/js/components/pool/EditPool/EditPool.tsx
@@ -206,10 +206,7 @@ export const EditPoolForm = ({
       <Breadcrumbs links={links} />
       <div data-h2-container="base(left, large, 0)">
         <TableOfContents.Wrapper>
-          <TableOfContents.Navigation
-            data-h2-background-color="base(dt-gray.light)"
-            data-h2-radius="base(s)"
-          >
+          <TableOfContents.Navigation>
             <TableOfContents.AnchorLink id={sectionMetadata.poolName.id}>
               {sectionMetadata.poolName.title}
             </TableOfContents.AnchorLink>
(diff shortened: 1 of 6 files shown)


vd1992 (changes requested):
a bit of an issue, it makes the text no longer interactive, you cannot click on the text to toggle the box

esizer:
> a bit of an issue, it makes the text no longer interactive, you cannot click on the text to toggle the box

Right, that is why we had it wrapping the input. This may require a bit more reworking 🤔

vd1992:
that's a lot of Chromatic buzz
from what I can tell, it appears to be a shift in the location of the "required" or "optional" label, but that was shifted inconsistently 

image

esizer:
> that's a lot of Chromatic buzz
> from what I can tell, it appears to be a shift in the location of the "required" or "optional" label, but that was shifted inconsistently

Ahh, yeah. Good catch! Had the logic backwards 😓  Updated in [fix appendLabel logic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4265/commits/b91a8e57d2b1ab3e76be13f42a38939ac69b8cf6)

vd1992 (approved):
- checks green
- Chromatic Diff shows mostly noise rather than changes of concern
- manual testing the checkbox works 

whew, that one ended being trickier than expected. amazing job!

PR #4271, opened 2022-10-12
Bug - Fix enums in other requirements section of edit pool page

### LINKED ISSUE #4243: Edit Pool - Dropdowns with enums try to update with an empty string when opened.
So after fiddling with things for a bit it looks like it only fails for me after I click the Select so I can view the options but click away without selecting one. It looks like this isn't the only place this happens though since I was able to replicate it with the security requirement field.

_Originally posted by @JamesHuf in https://github.com/GCTC-NTGC/gc-digital-talent/issues/4232#issuecomment-1271814725_
      
The forms on the edit pool advertisement page are a bit unusual in that they don't use form validation to ensure all the fields are filled in before submission.  Dropdowns for enums can behave a little strangely.  If the dropdown is opened and then closed without making a selection the backing value is set to an empty string which is an invalid submission for the enum type.  Guards need to be put in to convert the empty string to undefined.

yonikid15:
Resolves #4243

### DIFF
--- a/frontend/admin/src/js/components/pool/EditPool/OtherRequirementsSection.tsx
+++ b/frontend/admin/src/js/components/pool/EditPool/OtherRequirementsSection.tsx
@@ -97,7 +97,9 @@ export const OtherRequirementsSection = ({
 
   const handleSave = (formValues: FormValues) => {
     onSave({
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
Thanks!

PR #4292, opened 2022-10-13
a11y - 4206 - Increase chip touch area

### LINKED ISSUE #4206: Chip dismiss button has no label
h/t @MaddyDaigle 

## Summary

Right now the button to dismiss a `<Chip />` is `<svg role="button" />` and has no text content. This is causing issues for users of assistive technology not being able to properly interact with it.

## Location 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/7ba3ec98cd8500546314ad90797bc5ed1b1a5fbd/frontend/common/src/components/Chip/Chip.tsx#L60-L70

### Wrap Icon

We should be able to just wrap the `svg` in an `button` with stripped styles and provide some text.

#### Example

```tsx
...
<button 
  onClick={onDismiss} 
  aria-label={intl.formatMessage({
    defaultMessage: "Remove {label}",
    description: "Button text to remove a chip"
  })}>
  <svg />
</button>
...
```

### Wrap Chip

Another option is, we could wrap the entire chip in a button? I'm not sure what appropriate text for this would be but might make it easier increasing the target size for things like:

- Small screens
- Motor control disabilities (Cerebral palsy, parkinsons, etc.)

esizer:
Resolves #4206 

## Summary

This wraps a dismissible chip in a button, increasing the touch area. As well, it adds an `aria-label` to improve context for the button action.

## Testing

1. Run storybook `npm run storybook --workspace=common`
2. Open the "Chip Dismiss" story
3. Confirm the following
    - you can click on any area of the 
    - has focus indicator
    - has hover indicator

### DIFF
--- a/frontend/common/src/components/Chip/Chip.stories.tsx
+++ b/frontend/common/src/components/Chip/Chip.stories.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import { Story, Meta } from "@storybook/react";
 import Chip from "./Chip";
+import Chips from "./Chips";
 import type { ChipProps } from "./Chip";
 
 export default {
@@ -36,13 +37,13 @@ const TemplateNoDismissChip: Story<ChipProps> = (args) => {
 
 const TemplateMultiChip: Story<ChipProps> = (args) => {
   return (
-    <>
+    <Chips>
       <Chip {...args} />
       <Chip {...args} />
       <Chip {...args} />
       <Chip {...args} />
       <Chip {...args} />
-    </>
+    </Chips>
   );
 };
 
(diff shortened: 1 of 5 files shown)


petertgiles (approved):
Looks great.  I love the new hover shadow and that you've got the translation already done!

I'm hoping we can get a review from @MaddyDaigle as well?
https://61e099a1bb1465003a73d3ce-gosydhifjv.chromatic.com/?path=/story/components-chip--chip-dismiss

image

MaddyDaigle:
@esizer @petertgiles - Did a quick little accessibility check & it looks great to me!

Thanks! 👍

PR #4301, opened 2022-10-13
Bug/4279 view candidate page return link + Apache var

### LINKED ISSUE #4279: View candidate - go back link goes to pool advertisement
"Go back link" goes to pool advertisement instead of candidate listing table.

vd1992:
closes #4279 
updated the redirect 
merge #4209 added a flag check onto admin side, this was causing me issue until I realized I needed to update htaccess/apache for it
also hopefully fixes the current failing cypress test on main

test: 
make sure the above works lol
ensure you navigate correctly back when viewing a pool candidate profile


vd1992:
failure on the feature flag off section 

image

vd1992:
above corresponds to this

https://github.com/GCTC-NTGC/gc-digital-talent/blob/90797874ffb696f03b607c669994b07e4685e644/frontend/cypress/integration/admin/pool-candidate.spec.js#L108-L138

vd1992:
same error it appears

image

is the htaccess stuff overriding things in Cypress?

tristan-orourke:
Unfortunately, the cypress tests don't have access to the same env variables as the rest of the app. They use their own set of environment variables, as defined in `frontend/cypress.json`. For the test to work the value of the feature flag there has to match what is saved in apache_env, and - my bad! - it currently does not.

After we upgrade Cypress to v10, we can set this flag to have the same value as the system feature flag.

vd1992:
success!

tristan-orourke (approved):
Great job finding that I missed adding the feature flag to admin/public/.htaccess!

PR #4318, opened 2022-10-14
Fix broken links on application pages

### LINKED ISSUE #4317: Broken links in apply to pool pages
There are a couple problematic links on the apply to pool pages:
- On page 1 the "Step 2: Sign and submit" button uses the user id in place of the application id causing page 2 to fail to find the application.

- On page 2 both the "Step 1: Review my profile" and the "Back to previous step" buttons appear to still use a placeholder link and should be updated to return to the review my application page.

JamesHuf:
Resolves #4317.

Fixes some broken links on the pool application pages.

Test by creating an application and ensuring all the links both application pages (review application page and sign and submit page) work properly.


petertgiles (approved):
Great catches!

PR #4319, opened 2022-10-14
Bug/4300 pool advertisements already applied to

### LINKED ISSUE #4300: Pool Advertisement should look different for existing candidates.
Currently there is no change in the appearance of the Pool Advertisement page if the user has an existing application in the pool. In addition if the user attempts to (re)apply to the pool it resets their status to DRAFT but they can't resubmit their application.
We should probably get some new copy and disable the "Apply for this process" buttons if the user already has a candidate in the pool.

vd1992:
closes #4300 
adds validation to throw an error if the user has a non DRAFT application for a pool
adds a test too, as well as updates ApiMessages (definitely something to start doing more of)

testing:
ensure checks pass
- open `/browse/pools` as well as `/users/:id/applications`
- apply to a posting, should take you to the review application, assert this is reflected in My Applications
- go back to the poster and apply again, should change nothing from the last step
- complete and submit the application, ensure this is reflected in My Applications
- go back to the poster now and apply again 

### assert
- it goes nowhere
- toast appears with Intl message  "You have already submitted an application to this opportunity."
- application status unchanged in My Applications 

### **EDIT**
- now no toast appears, but rather the button should change and no longer be active


petertgiles:
@Jerryescandon You agree with this behavior change and copy?

vd1992:
sorry if you already saw this and were busy, but I'm just going to ping you once more @Jerryescandon

vd1992:
as a stopgap until #4357 is addressed, I have made some tweaks to avoid the frustration of false positives for Cypress

diploma weighting has been adjusted to be vast majority territory

a prior PR increased the number of pools by 10 to a total of 12, now the same candidate count is being seeded over 12 rather than 2 pools which also ties into the above issue
since some of the pools aren't even published, they shouldn't even have candidates, the most important one to ensure has a healthy seeded count is the Digital Talent one
the simple solution I went with was to bias seeding in favour of digital talent, rather than check statuses of all pools

frontend/talentsearch/src/js/components/pool/PoolAdvertisementPage.tsx
```diff
@@ -770,6 +789,23 @@ const PoolAdvertisementPage = ({ id }: PoolAdvertisementPageProps) => {
     visible = true;
   }
 
+  // grab pool candidates of Me, then check whether a pool candidate exists that matches the advertisement AND is submitted
+  const meDataResponse = useGetMePoolCandidatesQuery();
+  const meData = meDataResponse[0]?.data;
```
petertgiles, line 794: Access the data directly out of the array is not the best way to do this.  Check useGetPoolAdvertisementQuery on L781 above this for a better way.

petertgiles (changes requested):
It works!  A little weird to have a button that can never be used but I'm no designer.  :shrug:  :laughing: Just need to patch up your query a bit.

frontend/talentsearch/src/js/components/pool/PoolAdvertisementPage.tsx
vd1992: okay
here https://github.com/GCTC-NTGC/gc-digital-talent/pull/4319/commits/0e3ba54c346724b4e8488c729f8d15865567550b

petertgiles (approved):

PR #4358, opened 2022-10-17
Remove search field for Pools table

### LINKED ISSUE #4280: Pools Table - Search does nothing
https://images.zenhubusercontent.com/364642496/fb0aa20d-e21a-48a3-b707-e654d71b75c8/simplescreenrecorder_2022_10_13_10_08_57.mp4

JamesHuf:
Resolves #4280.

Modifies the Table component to allow removing the search field and removed it on the Pools table.

To test:
- Navigate to the Pools table and confirm it no longer has a search field
- View some other tables and ensure they still have the search field if they did before.


esizer (changes requested):
Looking great so far. We do have some cypress tests that use the search field that will need to be adjusted though.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/7e8cb457ab4729d211b589ea59141ddad7240f7e/frontend/cypress/integration/admin/pools.spec.js#L130-L134

https://github.com/GCTC-NTGC/gc-digital-talent/blob/7e8cb457ab4729d211b589ea59141ddad7240f7e/frontend/cypress/integration/admin/pools.spec.js#L162-L166

JamesHuf:
I removed the searches done in the tests and it's passing locally but not here. I'm not sure why that is.

esizer:
> I removed the searches done in the tests and it's passing locally but not here. I'm not sure why that is.

So, I was able to review this and it looks like the pool we are trying to edit is actually on the second page 😓 

You will like need to add something along the lines of:

```js
cy.findByRole("button", { name: /goto page 2/i }).click();
```

esizer (approved):
Awesome! 😀

Issue #4273, opened 2022-10-12
Publish pool dialog - expiry date off by 1


petertgiles:
Observe how I set the publish date to Oct 31 but the publish dialog shows Oct 30.

https://images.zenhubusercontent.com/364642496/12cd5065-0c76-4be3-b42d-d95544f4986e/simplescreenrecorder_2022_10_12_13_52_48.mp4

Probably covered by  #3315

## Acceptance Criteria
- [ ] When choosing an expiry date, user selects a date (not a time), but the UI informs the user that the job will close at 11:59 PM Pacific Time on their chosen date.
- [ ] A datetime, set to 11:59 PM Pacific Time on the chosen date, is submitted to the API. (The api should be expected to convert to UTC before saving it.)
- [ ] When displaying the expiry date anywhere on the Edit Pool page, and when initializing the date input with existing data, first convert the datetime from the api from UTC to Pacific Time, and show very explicitly that it is a time in a particular timezone.

vd1992:
the cause seems simple enough
updating it submits something at day of, midnight, so 00:00 UTC time, then when it loads it from the database, the time zone conversion bumps it to the day prior

![image](https://user-images.githubusercontent.com/40485260/195415434-0ca45cfe-431a-4a71-b34b-c93609438d0a.png)

the tricky part is a solution 
proposal: 
given pool expiry times are known (11:59 PM Pacific time), an intermediary function can be added so what gets sent to database is actually 06:59 AM UTC (7 hour difference)

vd1992:
I think the big sticking point might be, pools will not close on the same day for everyone
if closing time is 11:59 PM PST, it closes 02:59 AM for EST, therefore the expiry date is actually two possible dates

petertgiles:
@Jerryescandon For the AC of this issue we need to inform the user of the time zone nuance.  For the picker I had two ideas:

Option 1: A tooltip with the question button like Process Number
![image](https://user-images.githubusercontent.com/8978655/195932945-3c90de39-f55f-4941-8874-20ba09fed294.png)

Option 2: Sub text like Your impact
![image](https://user-images.githubusercontent.com/8978655/195933347-6062d3c2-a8d8-4d4d-9446-ecd6cac9d892.png)

Same design needs for Extend Date:
![image](https://user-images.githubusercontent.com/8978655/195934585-598b67b3-0b43-42f1-a471-43e834fa160d.png)


[...]

PR #4366, opened 2022-10-18
Convert pool expiry TZ field to UTC

petertgiles:
# Introduction
## What does it do?
This branch converts the pool advertisement expiry date / closing date from timestamptz to timestamp in the database and from a DateTimeTz to a DateTime in the API.  It also ensures that the front end is always sending UTC datetimes to the API and always interpreting the responses as UTC.  It adds `time-fns-tz` as a dependency for handling time zone conversions.  This seemed like a reasonable choice since we're already using `time-fns`.
## What does it not do?
The designers have not gotten back to me on when to show the date as Pacific and when to show it as local and how the copy will make that obvious.  For now I have the admin site always show Pacific and the talentsearch site always show local but I'll push the final update off to #4273.  I've skipped the poolcard tests until then, too.  The relativeExpiryDate function is known to not be working correctly now, too.  I'll wait to fix that until #4273 when we have a clear idea of how it's supposed to look.
# Testing
## Resources
- I like [worldtimebuddy](https://www.worldtimebuddy.com/?pl=1&lid=6094817,6173331,100&h=6094817&hf=1) for checking time zones.
- I link to some [dateful](https://dateful.com/time-zone-converter) conversions in the Jest tests, too.
## Suggested Steps
1. Before running the new migration copy out your existing expiry dates `select id, expiry_date from pools order by id`
2. Run the migration and ensure the expiry dates are the same "instant" as they are converted from local time zone to UTC.
3. Roll back the migration and observe how they return to the original values
4. Rerun the migration, run codegen for talentsearch and admin, rebuild the sites
5. Run the Jest tests and ensure they all pass
6. Log in as admin and navigate to /admin/pools/create
7. Set the closing date on your new pool. Pick a day before the Nov 6 time change.  Check the database to ensure it has been saved as the last second of the day in Pacific, but in UTC.
8. Refresh the page to ensure that the closing date is still showing the day in Pacific, not the following UTC or local day.
9. Publish the pool, ensure the publish dialog shows the Pacific day.
10. Navigate to the admin View Pool page, ensure the closing date shows the Pacific day.
11. Use the Extend dialog to set a new closing date.  Pick a day after the Nov 6 time change. Check the database to ensure it has been saved as the last second of the day in Pacific, but in UTC.
12. Navigate to /browse/pools, observe that the pool card shows the local day, not Pacific.
13. Click apply and observe that the header shows the local day, not Pacific.
14. Return to the admin editing page, close the pool, observe the closing date is in Pacific.

### DIFF
--- a/api/database/migrations/2022_10_17_141813_change_expiry_date_to_datetime.php
+++ b/api/database/migrations/2022_10_17_141813_change_expiry_date_to_datetime.php
@@ -1,0 +1,34 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\DB;
+
+class ChangeExpiryDateToDatetime extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        DB::statement("
+            ALTER TABLE pools
+            ALTER COLUMN expiry_date TYPE timestamp(0)
+            USING (expiry_date at time zone 'UTC')
+        ");
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        DB::statement("
+            ALTER TABLE pools
+            ALTER COLUMN expiry_date TYPE timestamptz(0)
+        ");
+    }
+}
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -5,10 +5,6 @@ scalar Date @scalar(class: "Nuwave\\Lighthouse\\Schema\\Types\\Scalars\\Date")
 scalar DateTime
   @scalar(class: "Nuwave\\Lighthouse\\Schema\\Types\\Scalars\\DateTime")
 
-"A datetime and timezone string in ISO 8601 format `Y-m-dTH:i:sP`, e.g. `2020-04-20T13:53:12+02:00`."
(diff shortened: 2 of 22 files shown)


esizer (commented):
Awesome! 🤩 

Everything is working well, just need to clarify some things to make sure I understand from the description before approval. On `talentsearch` I assume the `relativeExpiryDate` is known to not be functioning properly now based on the fact that you skipped those tests.

## Example

In my timezone (America/Toronto) it is showing the correct time but one day off. Such as:

- Today when I expect tomorrow
- Tomorrow when I expect the date
- November 8th when I expect November 7th
Screen Shot 2022-10-18 at 10 43 46 AM
Screen Shot 2022-10-18 at 10 42 50 AM
Screen Shot 2022-10-18 at 10 42 20 AM

petertgiles:
> I assume the `relativeExpiryDate` is known to not be functioning properly no

Yeah, definitely.  I tried to avoid diving into that function since I don't have clear guidance from the designers on exactly how it's going to look.  I'll add a note to the PR description to make that clear.

esizer (approved):
> Yeah, definitely. I tried to avoid diving into that function since I don't have clear guidance from the designers on exactly how it's going to look. I'll add a note to the PR description to make that clear.

Cool, that was my assumption. Just making sure 😁

Issue #3979, opened 2022-09-16
Memory leak on profile -> salary edit


petertgiles:
Steps to reproduce
1) Navigate to /talent/profile
1) Scroll to the Salary section
1) Click the edit link
1) Make a change and click "Save and go back"

Note: I think this only happens on the final edit when the "You can now activate your profile" toast is popped

![image.png](https://images.zenhubusercontent.com/6172cf6e898e11fd20507433/b760816e-7546-4bd5-a909-146949c93834)

vd1992:
there appears to be a grouping of them now
#3740 
same root cause?

petertgiles:
> same root cause?

Yeah, I bet we're all seeing the same thing.

gobyrne:
@esizer thinks it's the scroll to top. We have a listener, but we don't close it. There's a scroll to component in React-Router we could use that might solve this.

esizer:
Closed by #4597

Issue #3740, opened 2022-08-22
Memory Leak in LanguageInformationForm.tsx


brindasasi:
**Problem** 
Memory leak error on profile page after editing language info form in profile.![185641408-e0329387-bd52-446a-980f-e4f7f35a1099.png](https://images.zenhubusercontent.com/61a7bd49e20af80694dd845e/1a94ed90-6912-4704-9b1b-1f10693fad02)

related PR : https://github.com/GCTC-NTGC/gc-digital-talent/pull/3051

petertgiles:
I had no luck with this one.  :disappointed:   My best guess is that it has to do with the navigate function in router.tsx but I'm really not sure.

esizer:
Should be closed by #4597 but would like confirmation from someone. Our old `useLocation` hook was causing memory leaks and that was replaced so this **shouldn't** be an issue anymore and I can no longer reproduce on that PRs branch.

petertgiles:
> I can no longer reproduce on that PRs branch.

Agreed.  I also can't replicate it on that branch any longer.  :+1:

Issue #3633, opened 2022-08-09
Dep: Bump react-hook-form


esizer:
## Summary

The latest version of `react-hook-form@7.34.0` introduces some changes to the types. Specifically, they have updated the type for errors which causes issues with out `<InputWrapper />` and `<InputError />` components which accept `error?: string;`. This type is no longer compatible with the type provided by `react-hook-form`.

## Acceptance Criteria

- [ ] Bump dep to `react-hook-form@7.34.0`
- [ ] Resolve type errors with
  - [ ] InputWrapper
  - [ ] InputError
- [ ] Attempt to reintroduce field labels into default error message

vd1992:
appears to have been resolved by 7.35

![image](https://user-images.githubusercontent.com/40485260/197621927-5daf4928-560d-4dc9-844f-98407e12c5a9.png)

PR #4377, opened 2022-10-18
Bug - 4310 - Add cache for talent search forms

### LINKED ISSUE #4310: Form field data is lost when user switches between languages.
### Description

When a user enters form fields on any of the Talent Portal profile forms and then switches languages, all the data that was entered is lost.

**Environment:** UA, probably prod as well.

### Why is this an issue?

During user testing, a user expressed that they often switch between languages to make sure they understand what they are asked.  They had to start over every time.

### Expectation

When language is switched, all fields that were filled are saved and can be viewed in both languages.

esizer:
Resolves #4310 

## 📋  Summary

This refactors some forms and adds `cacheKey` to all talent search forms so no data is lost on page refresh.

## ❗ Important

We are now styling inputs based on the current form state.

### What does this mean?

We must now wrap **_most_** (assume all) form components in the `react-hook-form` `FormProvider`. This is so the form components can access the state and render the field accordingly. This will help with styling the following states:

- Unsaved (this PR)
- Invalid
- Valid
- Base

## ✅  To Do

- [x] Track dirty form values
- [x] Unsaved changes summary

## 🧪 Testing

2. Navigate to all profile forms and test the folllowing:
    1. Edit the form
    2. Refresh page
    3. Ensure the edited state appears
3. Run tests `npm run test --workspace=talentsearch`


vd1992:
you didn't change this form much, but do you get a console warning when going from profile -> About Me ? 
doesn't happen for language info or government info it seems 

image

vd1992:
doesn't seem to happen always either, lovely

vd1992:
is there anyway to clear the cache, it sticks around when you refresh the page, but it also stays around if you return to /profile + refresh + go to /myapplications + refresh and then return to the form
which is kind of confusing because then you think the value changed without hitting save

esizer:
> is there anyway to clear the cache, it sticks around when you refresh the page, but it also stays around if you return to /profile + refresh + go to /myapplications + refresh and then return to the form
> which is kind of confusing because then you think the value changed without hitting save

Thanks for this info. I actually didn't develop the caching so im not 100% how it works. I'll dig into it a bit more and see if I can solve this 👍

petertgiles:
> do you get a console warning when going from profile -> About Me ?

I bet this is related to #3979 and #3740

esizer:
> I bet this is related to #3979 and #3740

It is `urql` cache 😬 .

Screen Shot 2022-10-19 at 9 53 27 AM

It is trying to run a query while caching it at the same time, i think...

esizer:
> is there anyway to clear the cache

We can likely try to refactor a bit to have a little more control on exactly when the cache clears. But, right now I updated it to clear when a user navigates away from the page in [clear basic form cache when navigating away](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4377/commits/9a45dc9e2c61560bff36ad7e1947959280c4c34b)

frontend/admin/src/js/components/pool/EditPool/CloseDialog.tsx
```diff
@@ -74,33 +76,35 @@ const CloseDialog = ({
...
-        required={false}
-      >
-        <div
-          data-h2-display="base(flex)"
-          data-h2-width="base(100%)"
-          data-h2-gap="base(.5rem)"
-          data-h2-background-color="base(dt-gray.light)"
-          data-h2-padding="base(x.25)"
-          data-h2-radius="base(s)"
+      <FormProvider {...methods}>
```
esizer, line 79: > **Note**
> We need to wrap all inputs and `InputWrapper` in a `<FormProvider>` so we can display the "unsaved changes" label requested by designers

frontend/admin/src/js/components/pool/EditPool/EditPool.test.tsx
```diff
@@ -84,7 +84,10 @@ describe("Edit Pool tests", () => {
     // render story and click the button to open the modal
     await act(async () => {
       render(<DraftAdvertisement {...props} />);
-      fireEvent.click(screen.getByRole("button", { name: /publish/i }));
+    });
+
+    await act(async () => {
```
esizer, line 89: > **Note**
> I was having trouble getting this test to pass consistently here and moving the dialog open to it's own act worked. I believe it is because we are rendering a new form in the dialog which must be wrapped in `act()` due to how `react-hook-form` renders work.

frontend/common/src/components/inputPartials/Fieldset/Fieldset.stories.tsx
```diff
@@ -11,20 +13,22 @@ export default {
...
-      <label data-h2-display="base(block)" htmlFor="two">
-        Two
-        <input type="checkbox" name={name} value="Two" id="two" />
-      </label>
-      <label data-h2-display="base(block)" htmlFor="three">
-        Three
-        <input type="checkbox" name={name} value="Three" id="three" />
-      </label>
-    </Fieldset>
+    <Form onSubmit={action("Submit Form")}>
```
esizer, line 16: > **Note**
> `Fieldset` needs to be wrapped in a `FormProvider` to compute the styles based off of state properly

vd1992:
104 stories in the Chromatic changeset... anyone notice anything concerning? from what I can tell the vast majority is things being shifted around by a few pixels and like three actual story changes (addition/removal) 

also, please revise the estimate and issue scope, this is a bit of a chunky one now 🤣

vd1992:
a bit picky, but is it possible to disable formstate for things like searching? 

image

esizer:
> 104 stories in the Chromatic changeset... anyone notice anything concerning?

Yeah, obviously take a good look at it. The major changes were to the `Alert` with the addition of a prop to control the style for `"basic" | "large"` (terrible names, i know...taking suggestions 😆 ). Also, style changes came from designers (designed by @BrookeCameron 👏 ):

- New `LargeAlert` 
- Request to have darker outline with min of `2px` width to have it stand out from background

vd1992:
changed state clears for some inputs on submission like
http://recordit.co/Adf6QRAqP6

but not for others like
https://recordit.co/lqrmz8AL4h

esizer:
> but not for others like

Can you try forcing a reload after submission, I have a hunch?

---

So, what I think this is, is not necessarily a problem with the logic of the unsaved changes but rather our form logic. Usually, the form is reset because there is a page reload. However, we never actually reload our page, so our forms are never reset and always considered dirty is changed.

## Current Logic

1. Form values changed and `dirtyFields` updated
2. Field checks if it `isDirty` and renders blue is true
3. Form is submitted
4. Submit handler posts to API
5. Success Toast

## Proper Logic

1. Current 1-5
2. Reset form with values returned
3. No longer has any `dirtyFields`
4. All fields reset to default state

> **Question**
> Should I go through and update all forms to reset themselves on submit

vd1992:
yeah, reloading it fixes it
I just wanted to point out that it updated some inputs to represent them no longer being dirty but not all

esizer:
> a bit picky, but is it possible to disable formstate for things like searching?

Good catch, I added a `trackUnsaved` prop to the inputs, set it to true by default and set it to false where I thought appropriate.

[allow to not track unsaved changes](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4377/commits/fe6f2f09a4c384137bf10644fa025bff8b3fbd8c)

tristan-orourke (commented):
Wow, this is really cool! (Definitely got bigger than the original ticket, but this seems worth it 😄)

> Question: 
> Should I go through and update all forms to reset themselves on submit

I think you're right about the proper reset logic, but reseting the form with the values returned might be overkill in most cases. I would say that if submission was "successful", we can pretty confidently just reset all the dirty flags. 

How can we handle that in BasicForm, without requiring every form to call a callback... 🤔
Maybe we require the onSubmit function passed to BasicForm to be a Promise (right now that's optional) which rejects if the submission didn't succeed?

vd1992:
multi-select for the skill picker, filter by skill family, could also use the addition of `trackUnsaved` I believe

image

tristan-orourke:
Found a couple inconsistencies and edge cases:

Sometimes reverting a selection makes the "Unsaved Changes" notice go away: https://recordit.co/cVtFIy5lIb
Sometimes it does not: https://recordit.co/DBtCeAxnW8
Sometimes the field-level dirty warning goes away, but the overall field remains dirty: https://recordit.co/8IFCqnq1WW

Sometimes, awkwardly, its possible for a field to be hidden or not-rendered but still show up as dirty: https://recordit.co/ixBp0NsM73

esizer:
> Found a couple inconsistencies and edge cases:

Wow, this is a good catch. Very interesting. We may need to find a way to ignore fields that are not rendered due to conditional logic for the last one. I'm not entirely sure what is happening with check lists though, I'll need to look into that 🤔

vd1992:
> Sometimes the field-level dirty warning goes away, but the overall field remains dirty: https://recordit.co/8IFCqnq1WW
> 
> Sometimes, awkwardly, its possible for a field to be hidden or not-rendered but still show up as dirty: https://recordit.co/ixBp0NsM73

I think issues with language/government pages info may stem from the `watch()` hook needed to conditionally render inputs

vd1992:
I wonder if this may play a part, checking and unchecking causes the array to be ordered differently
the contents are the same but the order being different may cause it to be treated as still dirty

image

esizer:
> How can we handle that in BasicForm, without requiring every form to call a callback... 🤔

Okay, we now have the `BasicForm` doing a reset on submission along with clearing the cache (doesn't do much since any change will reintroduce it again but might as well cleanup after ourselves 🤷 ). Since the admin forms don't have a normalized `BasicForm`, I added a reset to the `EditPool` page which, i think, is the only one where we do not navigate away on.

[add form devtools + reset](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4377/commits/4557fe2dba1bc98052758306cffcc2f0e64f8da9)

> Sometimes the field-level dirty warning goes away, but the overall field remains dirty:

So, I'm not 100% sure why this was happening. But, I just added a check for the array length before rendering the alert to prevent it from showing up empty. If there was supposed to be something in there but it did not show up, let me know and I'll dive deeper into it.

[check dirty field length before rendering alert](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4377/commits/a1cb9e043264f5deaa9cf9b6494a6247ef669897)

> Sometimes, awkwardly, its possible for a field to be hidden or not-rendered but still show up as dirty:

We are now resetting fields when they disappear from the DOM. I only noticed this on `GovernmentInoForm` and `LanguageInformationForm`. If we have any other forms with conditional rendering, let me know and I can get to those as well.

[reset fields when they are removed from DOM](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4377/commits/9f0003ce6e47a37e5d853707ea297b1a2e364bbe)

## Checklist Array

So, we did bump `react-hook-form` in this PR as well. While I was initially able to reproduce this issue, reinstalling deps fixed it. There could be a version mismatch surfacing an old bug in `react-hook-form`, maybe? Anyway, can you try removing all `node_modules` and reinstalling fresh and see if that fixes it for you both as well?

@vd1992 @tristan-orourke

frontend/admin/src/js/components/pool/EditPool/EditPool.test.tsx
patcon: act should not be needed around any fireEvent, so if it's helping, it's because it's just slowing things down and avoiding a race condition that still exists (and might crop up sporadically)

If you feel skeptical of whether act should be avoided, i can dig up the issue from the test-library author where he said this unambiguously

vd1992:
Nice, you appear to have solved the conditional rendering problem
But I think you missed one

image

vd1992:
I'm not sure when this happened, whether it was part of this PR, but the DEI page isn't working right anymore
Removing no longer does anything 

image

esizer:
> Nice, you appear to have solved the conditional rendering problem
> But I think you missed one

Great catch! Fixed with [reset priority entitlement](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4377/commits/1473e5c6307676d9b9124c5f81d049fbafa3197a)

esizer:
> I'm not sure when this happened, whether it was part of this PR, but the DEI page isn't working right anymore
> Removing no longer does anything

So, I agree the copy should be updated. However, I am unable to reproduce the issue with those forms not working.

https://recordit.co/pl9rFIL3Om

vd1992:
deleting node modules and re-installing, did not seem to fix the checklist problem for location/work preferences
but interestingly, the checklist works for roles + salary
please note the state in the side, I feel that matches with my guess above

https://recordit.co/LWNtf3Se5O

vd1992:
> > I'm not sure when this happened, whether it was part of this PR, but the DEI page isn't working right anymore
> > Removing no longer does anything
> 
> So, I agree the copy should be updated. However, I am unable to reproduce the issue with those forms not working.
> 
> https://recordit.co/pl9rFIL3Om

Opposite for me
https://recordit.co/qRr9LVQv8Q 

😵

esizer:
> Opposite for me

It is positive removal, you must uncheck the box to remove it 😄

vd1992:
💀

vd1992:
judging by no apparent errors popping up anywhere
this is resolved too #3633

esizer:
> deleting node modules and re-installing, did not seem to fix the checklist problem for location/work preferences

Do you have some _exact_ steps on how to reproduce this, @vd1992 ?. I still have yet to be able to reproduce it 😬 I don't know how you are getting the array to be out of order, If you watch the following screencast no matter what order I click the checkboxes in they show up in the same order in the devtools.

https://recordit.co/0nxzskhXVg

vd1992:
> > deleting node modules and re-installing, did not seem to fix the checklist problem for location/work preferences
> 
> Do you have some _exact_ steps on how to reproduce this, @vd1992 ?. I still have yet to be able to reproduce it 😬 I don't know how you are getting the array to be out of order, If you watch the following screencast no matter what order I click the checkboxes in they show up in the same order in the devtools.
> 
> https://recordit.co/0nxzskhXVg

oh boy, that's weird
well, let's see if @tristan-orourke has the issue or my environment is quirky

vd1992:
I don't think I have anything else to add, I'll leave final approval for someone else. 
Great work, this really enhanced the user experience!

yonikid15 (changes requested):
Wow Eric first off this looks amazing👍However, I found a few some small issues:
- Clicking "Add Skill" when creating an experience triggers the unsaved changes feature to popup for all input items.

frontend/common/src/components/inputPartials/InputLabel/InputLabel.tsx
```diff
@@ -91,6 +96,16 @@ const InputLabel: React.FC<InputLabelProps> = ({
...
+          {fieldState === "dirty" && trackUnsaved && (
+            <span
+              data-h2-font-size="base(caption)"
+              data-h2-display="base(inline-block)"
+              data-h2-margin="base(0, 0, 0, x.125)"
+              data-h2-color="base(darkest.tm-blue)"
+            >
+              {intl.formatMessage(commonMessages.unSaved)}
+            </span>
+          )}
```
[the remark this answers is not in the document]
esizer: We need to keep it next to the label for proximity when using assistive technology such as screen magnification. When using [macos accessible zoom](https://support.apple.com/en-ca/guide/mac-help/mh40579/mac#:~:text=To%20quickly%20enable%20or%20disable,click%20Accessibility%2C%20then%20click%20Shortcut.) this is what I see:

Screenshot 2022-10-26 at 9 55 39 AM

frontend/common/src/components/Alert/BasicAlert.tsx
```diff
@@ -0,0 +1,41 @@
...
+    >
+      <p
+        data-h2-display="base(flex)"
+        data-h2-align-items="base(center)"
+        data-h2-font-weight="base(600)"
+      >
+        <Icon
+          data-h2-vertical-align="base(top)"
+          data-h2-margin="base(0, x.25, 0, 0)"
+          style={{ height: "1em", width: "1em" }}
```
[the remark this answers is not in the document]
esizer: So, looks like I was wrong. Thought i'd try again and it looks like it seems to be working now! 🎉 

[prefer hydrogen to inline styles](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4377/commits/ecc98e748ee75e5dbe529a530f8ee574da60416a)

tristan-orourke:
After re-installing node modules as suggested, my previous issues went away! 👍

yonikid15 (changes requested):
I'm still seeing some behaviour which I'm not sure is correct on the the Add and Edit my experience and skills form:
1) When editing an experience, changing any field will trigger the unsaved changes feature on all fields even if they are not `dirty`. 

https://user-images.githubusercontent.com/22059495/198649179-1dedd60a-fcf6-4de1-995e-a76de554ff20.mp4

2) When adding an experience.. when I try adding a skill, it immediately triggers the unsaved changes on all fields.

https://user-images.githubusercontent.com/22059495/198649241-2de460cb-0e4c-4273-ada9-3e93f462f43c.mp4

esizer:
> I'm still seeing some behaviour which I'm not sure is correct on the the Add and Edit my experience and skills form:

I have no idea why but `react-hook-form` was adding those fields to the `dirtyFields` object with a value of `false` 🤷‍♀️ . Added a check to make sure they are actually dirty before adding them to the list in [check dirty field values](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4377/commits/6c97cf0cfa54b5c94db69dafc04da487dc40c590)

yonikid15 (commented):
I found another bug with the in-page links 😢. Some of the links site-wide are not navigating to the section of the page where the changes have been made. Probably due to the naming not matching up. Here's an example on the "Create Account" page:

https://user-images.githubusercontent.com/22059495/198681342-ee25f9a7-ebf1-4a22-b66e-8956115a619e.mp4

Broken links:
- #preferredLang
- #govEmployeeYesNo
- #govEmployeeType
- #priorityEntitlementYesNo

I'm seeing it occasionally across a lot of other forms as well. I would suggest going through and checking. If you want some help with that please don't hesitate to ask.

esizer:
> I'm seeing it occasionally across a lot of other forms as well. I would suggest going through and checking. If you want some help with that please don't hesitate to ask.

Would you say this only happens with fields that implement `fieldset`? i.e

- `RadioGroup`
- `Checklist`

yonikid15 (approved):
Alright looks great now! Adding the id's to the fieldset seems to have fixed it 👍 
Approved ✅

PR #4407, opened 2022-10-20
Add time zones to date parse and display

### LINKED ISSUE #4273: Publish pool dialog - expiry date off by 1

petertgiles:
# :wave: Introduction
This branch:
- Expands the `relativeExpiryDate` in dateUtils to accept a time zone so that it can generate a string from that perspective 
- Adds a suite of tests for that function
- Updates the display of pool expiry date to sometimes show local time zone, sometimes show Pacific time zone, and sometimes both
# :test_tube: Testing
## :dart: Expectations
- In Closing Date section of Edit Pool, explanatory message with Pacific date
- In Publish Dialog of Edit Pool, both local and Pacific relative strings
- In Extend Dialog of Edit Pool, same explanatory message
- In Admin View Pool, both local and Pacific relative strings unless they are the same
- In Talent Search pages (Browse Opportunities, Application header, Pool Advertisement), only local relative string

## :spiral_notepad: Suggested Steps
1. Run the Jest tests
1. Make sure you're in Eastern time (sorry James, this only works if you're not in Pacific already :laughing:)
1. Start a new pool and set the closing date to the last day of a month
image
1. Publish the pool
image
1. Extend the pool
image
1. View the pool on the admin site
image
1. View the pool on the talentsearch site
image
image
1. Apply to the pool
image

# :robot: Robot Stuff
Closes #4273


Jerryescandon:
Can we expand labels on 4 and 6 to go from just (Pacific) to full (Pacific time zone)?

petertgiles:
> Can we expand labels on 4 and 6 to go from just (Pacific) to full (Pacific time zone)?

You got it.  
[Expand Pacific time zone label](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4407/commits/8af003e1987f26d112d9a6b1bd40d8428cd4e8eb)
image
image

frontend/common/src/helpers/dateUtils.ts
```diff
@@ -10,7 +10,14 @@ import { getLocale, Locales } from "./localize";
...
 export const DATETIME_FORMAT_STRING = "yyyy-MM-dd HH:mm:ss";
-export const LONG_DAY_FORMAT_STRING = "EEEE, d MMMM yyyy";
+
+// https://stackoverflow.com/a/563442/18246559
+const addDays = (startDate: Date, days: number) => {
+  // not really const since setDate changes the value in place
+  const date = new Date(startDate.valueOf());
+  date.setDate(date.getDate() + days);
+  return date;
+};
```
esizer, line 20: Just a tip if we want to reduce code we need to maintain, `date-fns` has an [`add`](https://date-fns.org/v2.29.3/docs/add) helper ;)

```tsx
const result = add(date, { days: 1 });
```

petertgiles: Good catch!  I should have guessed that was in there.  Fixed in [Use date-fns dateadd](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4407/commits/728c9240d3174f7b2af21774c9ae93cd921fb248).

esizer (commented):
Awesome! 🎉 

I'm ready to approve this but I have one thing I noticed and if you agree, we can either create a new ticket or tack it onto this.

When doing this I found it a little odd and may cause confusion for some admins. Even though we tell them that expiry date will be set to 11:59 Pacific, it still found it odd that the dates were different. I know why, but think this may confuse some people editing these.

Do you think it may help to show the time as well so you can clearly see it is not a full 24 hours difference, rather just 3 hour difference. Otherwise, I see some people assuming any timezone east of pacific have a full 24 hours more to apply.

petertgiles:
@Jerryescandon Eric is still [finding this difficult to interpret](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4407#pullrequestreview-1151071527) and I don't disagree.  Could you please take a second look and see if we need to tweak it further?  I'm happy talk through it with you if you like.

Jerryescandon:
I agree that adding hours would make it clear about the difference.

petertgiles:
Better with [Add time to relative date string](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4407/commits/9fed7c3f5aa8882ac5c8ec859f5e25ffff36c937)? I've removed day of week and day count to keep length shorter.
image
image
image

tristan-orourke:
Just my two cents...

1. Do we need to show the local time at all on the admin side? Seems to me like just showing the Pacific timezone is more clear, clean, and useful.  image
2. I like showing the time on the Applicant side as well, it certainly makes it more clear to me. However, I wonder if users (who famously don't read any instructions) will miss that a pool closes at 2am? Maybe its not a problem, since we explicitly give them the number of hours left when the time gets closer.

petertgiles:
> 1. Do we need to show the local time at all on the admin side?

Makes sense to me, though that's more of a @Jerryescandon question

> since we explicitly give them the number of hours left when the time gets closer.

Yeah, the description does still change to "Closes tomorrow at 2:59:59 AM" the day before closure.

esizer:
Should we display the time in `PoolCard` as well? If you are in the mountain timezone, the pool closes at 12:59AM and seems a little misleading and showing the exact time will reduce any confusion.

## Example

The card indicates "Apply by November 7" which, as a user I may think I have all day November 7th. But, in reality, I only have about an hour of that day to apply. The inverse could also be true, in pacific, I would see "Apply by November 6" and I might assume it closes at 12:01AM and therefore must apply on November 5th.

Screen Shot 2022-10-24 at 9 53 32 AM

Jerryescandon:
All public facing times should always be displayed in local time. Also, the text changes as soon as the time slips under 24 hrs I believe?

Jerryescandon:
I do see your point now, yeah maybe add the time too?

petertgiles:
Add time to PoolCard in [Generalize formatDate and add time to PoolCard](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4407/commits/7ccbeced3c6997b1e81338bf7b3241ef1b0059ea)
image

Jerryescandon:
Looks good to me

frontend/common/src/helpers/dateUtils.ts
```diff
@@ -58,27 +58,32 @@ export function getDateRange({
       );
 }
 
+// parameters for the relativeExpiryDate function
+export type formatDateOptions = {
+  date: Date;
+  formatString: string;
+  intl: IntlShape;
+  timeZone?: string;
+};
```
esizer, line 67: Pure nitpick, but can we use `PascalCase` for type names. Also, I think the comment needs updating:

`// parameters for the formatDate function`

petertgiles: Good stuff!  Gotta keep things tidy.  [Tidy naming and constants](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4407/commits/676bd3492b6353dd42529568b685aaa426028e92)

esizer (approved):
Amazing! 😲

PR #4408, opened 2022-10-20
feature/3949-redo-poolcandidate-migration to user

### LINKED ISSUE #3949: Migrate Pool Candidates to Users Again

vd1992:
closes #3949 
adds a new migration that is just a copy of the removed one https://github.com/GCTC-NTGC/gc-digital-talent/commit/aa0ad4a3f000be69b81e12852dfc44ed4eafe53a
as far as I can tell, we have not added any new fields to Pool Candidate that should be migrated to User

testing: 
@petertgiles  re-review?

### DIFF
--- a/api/database/migrations/2022_10_20_170113_migrating_pool_candidate_data.php
+++ b/api/database/migrations/2022_10_20_170113_migrating_pool_candidate_data.php
@@ -1,0 +1,63 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\DB;
+use App\Models\PoolCandidate;
+
+class MigratingPoolCandidateData extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+
+    public function up()
+    {
+        // update table users with data from table pool_candidates
+        DB::statement('UPDATE users
+                    SET is_woman = p.is_woman,
+                        has_disability = p.has_disability,
+                        is_indigenous = p.is_indigenous,
+                        is_visible_minority = p.is_visible_minority,
+                        has_diploma = p.has_diploma,
+                        language_ability = p.language_ability,
+                        location_preferences = p.location_preferences,
+                        accepted_operational_requirements = p.accepted_operational_requirements,
+                        expected_salary = p.expected_salary
+                    FROM pool_candidates p
+                    WHERE users.id = p.user_id');
+
+        // insert data into cmo_users and classifications_user using PoolCandidate and User models with query builder
+        $allPoolCandidates = PoolCandidate::all();
+        foreach($allPoolCandidates as $candidate){
+            $user = $candidate->user()->getChild();
+
+            // build arrays of Assets and Classifications existing on PoolCandidate model and attached User
+            $classificationsCandidate = $candidate->expectedClassifications()->pluck('classifications.id')->toArray();
+            $assetsCandidate = $candidate->cmoAssets()->pluck('cmo_assets.id')->toArray();
+            $classificationsUser = $user->expectedClassifications()->pluck('classifications.id')->toArray();
+            $assetsUser = $user->cmoAssets()->pluck('cmo_assets.id')->toArray();
+
+            // merge into a combined Assets array as well as Classifications array, then remove duplicates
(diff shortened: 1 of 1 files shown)


petertgiles (changes requested):
OK, I did a double-check by importing the UAT database for a bit more realistic data and fixing a bug in my check query.  I'm getting some mismatches of the booleans but I think that's just that some users have multiple candidates in UAT which won't happen in prod.  But I think there's a problem somewhere with the migrations for classifications and CMO assets that's causing them to not being migrated and leaving users with an empty list.  It's probably simplest for you to manually set up that condition for yourself but if you want I can help you import the UAT dump that I have.

petertgiles (approved):
Yeah, this looks better now.  Nice!

tristan-orourke (changes requested):
I'd like this to be refactored a bit before merging. Explanation incoming.

api/database/migrations/2022_10_20_170113_migrating_pool_candidate_data.php
```diff
@@ -0,0 +1,64 @@
...
+            $combinedClassifications = array_merge($classificationsCandidate, $classificationsUser);
+            $combinedClassificationsDedupe = array_unique($combinedClassifications);
+            $combinedAssets = array_merge($assetsCandidate, $assetsUser);
+            $combinedAssetsDedupe = array_unique($combinedAssets);
+
+            // sync the User with the de-duped arrays
+            User::where('id', $userId)->first()->expectedClassifications()->sync($combinedClassificationsDedupe);
+            User::where('id', $userId)->first()->cmoAssets()->sync($combinedAssetsDedupe);
+        }
+    }
```
tristan-orourke, line 53: Can we rewrite this to avoid using the PHP models? 

Basically, our PHP code might change. eg Maybe we remove the expectedClassifications relationship from PoolCandidate. But we won't want this migration to break, because it still needs to run every time we spin up a new dev environment. 

The database will always have the same shape every time this migration runs, since migrations run in deterministic order, so we can ensure it doesn't break by using [DB queries](https://laravel.com/docs/5.0/database#running-queries) instead of PHP Eloquent models.

For example, this line
```
$classificationsCandidate = $candidate->expectedClassifications()->pluck('classifications.id')->toArray();
```
could be replaced with 
```
$classificationsCandidate = DB::select("select candidate_id from classification_pool_candidate where pool_candidate_id = ?", $classificationId);
```

tristan-orourke (changes requested):

vd1992:
alrighty, switched from migration file to artisan
run 
`php artisan migrate:pool_candidate_to_user`

tristan-orourke (changes requested):
Okay, I like this! Only comment - can we think of a different name for the command? I just want to make it obvious that it's distinct from the [Laravel migration](https://laravel.com/docs/9.x/migrations) command.

vd1992:
> Okay, I like this! Only comment - can we think of a different name for the command? I just want to make it obvious that it's distinct from the [Laravel migration](https://laravel.com/docs/9.x/migrations) command.

I just copied the format Peter had going
Open to suggestions

vd1992:
okay
https://github.com/GCTC-NTGC/gc-digital-talent/pull/4408/commits/12e5c0d2d79dee379c0dad1566e660e258dfc415

tristan-orourke (approved):

PR #4409, opened 2022-10-20
Modify search page to search across all pools 

### LINKED ISSUE #4374: Modify Search Page to search all pools, not just digital_careers
Currently, the search page is hard coded to filter by the digital_careers pool. Remove this restriction so that it counts candidates in all pools.

Note: this will make the way the Request Candidates button is displayed (in a card describing the Digital Careers pool) inaccurate. Do not worry about this for now.

## Acceptance Criteria
- [x] The search page will count applicants in all pools, as long as they meet all other constraints.
- [x] The search page does not rely on any data that comes from a specific pool (ie to populate any text or any form inputs)
- [x] Classification Filter should show only IT-01 , IT-02, IT-03, IT-04
- [x] This does not affect the "deprecated" version of the search - in other words, this change is only visible if APPLICANTSEARCH feature flag is true

brindasasi:
resolves #4374 

- Search page uses hard coded constants for the classifications named as NonExecutiveITConstants which are from IT-01 to IT-04. 
-  Eliminates pool id from the search
- Search page still filters by expected classifications as usual. No change on that in this PR.

### DIFF
--- a/frontend/common/src/constants/NonExecutiveITClassifications.tsx
+++ b/frontend/common/src/constants/NonExecutiveITClassifications.tsx
@@ -1,0 +1,50 @@
+import { Classification } from "../api/generated";
+
+export default (): Classification[] => {
+  return [
+    {
+      id: "IT-01",
+      name: {
+        en: "Information Technology",
+        fr: "Technologie de l'information",
+      },
+      group: "IT",
+      level: 1,
+      minSalary: 50000,
+      maxSalary: 80000,
+    },
+    {
+      id: "IT-02",
+      name: {
+        en: "Information Technology",
+        fr: "Technologie de l'information",
+      },
+      group: "IT",
+      level: 2,
+      minSalary: 65000,
+      maxSalary: 94000,
+    },
+    {
+      id: "IT-03",
+      name: {
+        en: "Information Technology",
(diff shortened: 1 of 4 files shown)


tristan-orourke (changes requested):
Hmm. I don't like the idea of reproducing the entire Classification objects as client-side constants, if we can avoid it. Especially since the id will not match the api/database value. If we can pass only the group and level into the SearchContainer, lets do that. If the search container needs more information than that, let's retrieve it from the api (by getting all Classifications and then filtering them on group and level).

tristan-orourke (approved):
I was able to modify SearchForm in https://github.com/GCTC-NTGC/gc-digital-talent/pull/4409/commits/ce48c521de4ea6e4cd7183ebf0fb00e23f6a4f90 so that its happy with just the group and level of Classifications, without really changing anything! Everything else looks good, and it seems to work correctly!

brindasasi:
> I was able to modify SearchForm in [ce48c52](https://github.com/GCTC-NTGC/gc-digital-talent/commit/ce48c521de4ea6e4cd7183ebf0fb00e23f6a4f90) so that its happy with just the group and level of Classifications, without really changing anything! Everything else looks good, and it seems to work correctly!

Thanks Tristan!

PR #4418, opened 2022-10-21
Bug - 4313 - Render new lines text area

### LINKED ISSUE #4313: New line formatting when creating a new pool (work task & impact info) is removed
### Description
When an admin adds Work task & impact information to a pool, all new line formatting is removed - so everything just looks like a block of text.

### Expectation
The admin user should be able to add new lines to their text and descriptions.

#### What the admin enters
image (1)

#### What is shown on the pool
image

## Acceptance Criteria
- [x] Ensure that newlines are taken from the textinput form and sent to api
- [x] Ensure newlines in api submission are saved to database
- [ ] Ensure we display newlines correctly when displaying longform text from the pool
    - [ ] Browse Jobs card?
    - [ ] Individual Job Poster page
    - [ ] Job Poster page - admin view

esizer:
Resolves #4313 

## 📋 Summary

This wraps the content coming from a multi-line input (`textarea`) with a component that displays new lines with css.

## ❓ Question

While reviewing this, if you know of anywhere else we are using `textarea` and displaying the values, can you let me know so it doesn't get missed? 😄 

> **Note**
> I purposefully did not include this on the descriptions for pools in the `PoolCard`. Thought that may look strange 🤔 
> If you disagree, let me know and we can add it there as well.

## 🧪 Testing

1. Build all projects `npm run production --workspaces`
2. Edit a pool, adding new lines in "Your impact" and "Work tasks"
3. View the pool on admin and confirm line breaks appear
4. View the pool on "Browse Opportunities" confirm line breaks appear
5. Edit an experience on the profile
6. Added line breaks in skill descriptions and additional information
7. View the skills in all both Experience and Skill accordions confirm line breaks appear


esizer:
> Question: any reason we wouldn't want to make this the default, either by putting the style on our root components or in a CSS file?

That is a good question, I have never really done it before. @JoshBeveridge may know more about this than I would.

esizer:
Property | New lines | Spaces and tabs | Text wrapping | End-of-line spaces | End-of-line other space separators
-- | -- | -- | -- | -- | --
`normal` | Collapse | Collapse | Wrap | Remove | Hang
`nowrap` | Collapse | Collapse | No wrap | Remove | Hang
`pre` | Preserve | Preserve | No wrap | Preserve | No wrap
`pre-wrap` | Preserve | Preserve | Wrap | Hang | Hang
`pre-line` | Preserve | Collapse | Wrap | Remove | Hang
`break-spaces` | Preserve | Preserve | Wrap | Wrap | Wrap

REF: https://developer.mozilla.org/en-US/docs/Web/CSS/white-space

I think our best bet is actually to globalize either `pre-wrap` or `pre-line`. The big difference is whether we want to preserve of collapse spaces and tabs.

@tristan-orourke @JoshBeveridge

JoshBeveridge:
@esizer I might need someone to walk me through the context of this (and why it's needed) a little more clearly to be of any help. 

Side note 👍 for use of emojis in issue titles 😎

esizer:
> @esizer I might need someone to walk me through the context of this (and why it's needed) a little more clearly to be of any help.

Yeah, if you want to talk about it we can. But, essentially what this is about is the following:

When users type in a `textarea` and insert line breaks, they are stored in the database as `\n` which, by default is collapsed in css so when an admin reviews the content they filled out, all the line breaks disappear. We are fixing that by wrapping the output of `textarea` fields in a `<LineBreaks>` component that essentially just wraps them in `css white-space: pre-wrap;` to render the `\n` as line breaks.

Though, so we never have to think about it, @tristan-orourke suggested we just make that a global rule so we don't need to remeber to wrap the output of `textarea` at all. We just want to see if you had any thoughts about making a global css rule like:

```css
p {
  white-space: pre-wrap | pre-line;
}
```

JoshBeveridge:
Shucks, I wish I had the Hydrogen code out for CSS properties like this. Soon! But yes, that seems fine to me. I'm surprised though, that the `\n` characters are ignored in the first place 🤔

tristan-orourke:
I think the reason its not standard is so that you can write html like this
```
<p>
  Hello.
</p>
```
and not get surprising newlines. I remember it making sense when we did everything in handlebars template files. But now all our text is stored in json strings anyway.

esizer:
Alright, I removed the `<LineBreaks>` component and made the global rule. Chromatic diff isn't showing anything major and it all looked good during my testing. @tristan-orourke 

> **Note**
> I wonder if we can increase the threshold for chromatic diff? There are some negligible diffs in this PR that are hardly worth reviewing 😖 

REF: https://www.chromatic.com/docs/threshold

PR #4454, opened 2022-10-26
Search by qualified classification by the pool instead of expected classification

brindasasi:
# :wave: Introduction
This PR will provide the pool id to count the applicants who are all qualified on the selected classification. 
This will also restrict the classification selection  list (NonExecutiveITclassifications) to only that are already existing in the database. 

This PR does NOT change count applicants query or any other. Hence as long as the page passes the api request with correct parameters the accuracy of applicant count doesn't change.  
In this case , the network call should go with correct pool id for the IT classification you selected.  

# ⚡ Impacted Areas
- Classification dropdown
- Estimated Total Count on the right hand side
- Validation for form Submission 

# :test_tube: Testing
-  Create the Pools in local db for ( AS-01 , AS-02, IT-01, IT-02) 
-  Verify the classification filter lists only IT-01 and IT-02 only
-  When you select any classification filter make sure the countApplicants query on the network tab is passed with the correct pool id. 
- Verify on the initial load classification filter loads without any selection 
- Verify the form can be submitted and the request form page is opened


brindasasi:
this failing cypress test caught a bug 🥇  .. let me fix it quickly

frontend/talentsearch/src/js/components/search/SearchContainer.tsx
```diff
@@ -222,6 +220,22 @@ export const SearchContainer: React.FC<SearchContainerProps> = ({
 const SearchContainerApi: React.FC = () => {
   const [{ data, fetching, error }] = useGetSearchFormDataAcrossAllPoolsQuery();
   const skills = data?.skills;
+  const pools = data?.pools;
+
+  const availableClassifications = pools?.map(
+    (pool) => pool?.classifications && pool?.classifications[0],
+  );
```
tristan-orourke, line 227: Huh, I would have thought the `pool?.classifications && pool?.classifications[0]` expression would result in a boolean, not a Classification, but the typescript says otherwise.😕 Anyway, I don't think it's a good idea here to assume that each pool only has one classification. I'd recommend replacing this with
```suggestion
  const availableClassifications = pools
    ?.flatMap((pool) => pool?.classifications)
    .filter(notEmpty);
```

frontend/talentsearch/src/js/types/PoolUtils.ts
```diff
@@ -0,0 +1,8 @@
+import { Classification } from "@common/api/generated";
+
+export type SimpleClassification = Pick<Classification, "group" | "level">;
+
+export type SimplePool = {
+  id: string;
+  classification?: SimpleClassification;
```
tristan-orourke, line 7: The type Pool has `classifications` (plural) but SimplePool has a single `classification`. That could could some confusion, and potentially bugs, I think.

frontend/talentsearch/src/js/components/search/SearchForm.tsx
```diff
@@ -121,6 +122,19 @@ const SearchForm = React.forwardRef<SearchFormRef, SearchFormProps>(
     );
     const methods = useForm<FormValues>({ defaultValues: initialValues });
     const { watch, trigger } = methods;
+    const filterPoolsBySelectedClassification = (
+      allPools: SimplePool[],
+      classification: Maybe<SimpleClassification>,
+    ) =>
+      allPools
+        ?.filter((pool: Maybe<Pick<Pool, "id" | "classifications">>) => {
```
tristan-orourke, line 130: Since allPools has a definite type (`SimplePool[]) you don't need the `?` on line 130, and you don't need to define the type of `pool` in the filter callback. Actually, when I changed line 130 to `.filter((pool) => {` typescript found a bug: allPools has `classification` (singular) but inside the filter function we're looking at `classifications` (plural).

tristan-orourke: Also, maybe this filterPoolsBySelectedClassification woud be worth moving to PoolUtils?

frontend/talentsearch/src/js/types/PoolUtils.ts
```diff
@@ -0,0 +1,8 @@
+import { Classification } from "@common/api/generated";
```
tristan-orourke, line 1: I think our convention is that ts files that aren't components should have camelCase file names, ie `poolUtils.ts`.

tristan-orourke (changes requested):
The overall logic looks right, I think we just need to embrace the possibility that pools have multiple classifications, since that's the contract with the API.

tristan-orourke (approved):
Code looks great and it works perfectly. I did notice that we now should make some changes on the Request Summary component, but that's out of scope and can be discussed separately.

PR #4455, opened 2022-10-26
Candidate count, filtered by user, grouped per pool

### LINKED ISSUE #4398: CountApplicants query can return total number, plus number per pool
Update countApplicants query to return a map of pool -> number in addition to the total number of candidates.
Return data in shape of:
```
{
  total: Number
  byPool: [
    {poolId: Number }
  ]
}
```

## Acceptance Criteria
- [ ] Test with phpunit test
- [ ] Make use of SQL groupBy to do this efficiently
- [ ] Make sure doesn't break any existing uses of the query (instead of using result as a number can use result.total, or something)

petertgiles:
# :wave: Introduction
This branch adds a new query to the API for counting the number of pool candidates by filtering on their parent users and then groups them by pool.  I wasn't able to get a "natural implementation" based on the existing model and so it was implemented by a standalone query and manual query building.  H/t to @vd1992 for getting me started on the implementation.

I wasn't able to get it to also return the total number of users as originally defined in the issue.  I suggest issuing the existing user query at the same time in the same network request from the frontend.

# :test_tube: Testing
## :tipping_hand_man: Tips
- I like to uncomment the DB listener in api/app/Providers/AppServiceProvider.php and review the logged SQL.
- You'll need to use GraphQL Playground since there's no frontend usage for this yet.
```
query ($where: ApplicantFilterInput) {
  countPoolCandidatesByPool(where: $where) {
    pool { id }
    candidateCount
  }
}
------
{
  "where": {
    /* add some filters here */
  }  
}
```
## :notebook: Suggested Steps
- Run the PHPunit tests (don't forget to reseed afterwards)
- Run some manual tests in GraphQL Playground
- Search page still works well

# :robot: Robot Stuff
- Closes #4398

### DIFF
--- a/api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
+++ b/api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
@@ -66,12 +66,12 @@ final class CountPoolCandidatesByPool
 
             // wouldAcceptTemporary
             if (array_key_exists('wouldAcceptTemporary', $filters)) {
-                User::filterByWouldAcceptTemporary($userQuery, $filters['wouldAcceptTemporary']);
+                User::scopeWouldAcceptTemporary($userQuery, $filters['wouldAcceptTemporary']);
             }
 
             // expectedClassifications
             if (array_key_exists('expectedClassifications', $filters)) {
-                User::filterByExpectedClassifications($userQuery, $filters['expectedClassifications']);
+                User::scopeClassifications($userQuery, $filters['expectedClassifications']);
             }
 
             // skills
--- a/api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
+++ b/api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
@@ -1,0 +1,97 @@
+<?php
+
+namespace App\GraphQL\Queries;
+
+use App\Models\PoolCandidate;
+use App\Models\User;
+use Database\Helpers\ApiEnums;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Support\Facades\Log;
+
+final class CountPoolCandidatesByPool
+{
+    /**
+     * @param  null  $_
+     * @param  array{}  $args
+     */
+    public function __invoke($_, array $args)
+    {
+        $filters = $args["where"];
+
+        // query counts pool candidate rows, so start on that model
+        $queryBuilder = PoolCandidate::query();
+
+        // pool candidate filters go here
+
+        // candidate pool scope
+        if (array_key_exists('pools', $filters)) {
+            // pool candidate filter uses Pool while Applicant filter users IdInput
+            $pools = array_map(function ($id) {
+                return ['id' => $id];
+            }, $filters['pools']);
+            PoolCandidate::filterByPools($queryBuilder, $pools);
+        }
+
+        $queryBuilder->whereHas('user', function (Builder $userQuery) use ($filters) {
+            // user filters go here
+
+            // user status scope
+            User::scopeAvailableForOpportunities($userQuery);
+
+            // has diploma
+            if (array_key_exists('hasDiploma', $filters)) {
+                User::scopeHasDiploma($userQuery, $filters['hasDiploma']);
+            }
+
+            // equity
+            if (array_key_exists('equity', $filters)) {
+                User::filterByEquity($userQuery, $filters['equity']);
+            }
+
+            // languageAbility
+            if (array_key_exists('languageAbility', $filters)) {
+                User::filterByLanguageAbility($userQuery, $filters['languageAbility']);
+            }
+
+            // operationalRequirements
+            if (array_key_exists('operationalRequirements', $filters)) {
+                User::filterByOperationalRequirements($userQuery, $filters['operationalRequirements']);
+            }
+
+            // locationPreferences
+            if (array_key_exists('locationPreferences', $filters)) {
+                User::filterByLocationPreferences($userQuery, $filters['locationPreferences']);
+            }
+
+            // wouldAcceptTemporary
+            if (array_key_exists('wouldAcceptTemporary', $filters)) {
+                User::filterByWouldAcceptTemporary($userQuery, $filters['wouldAcceptTemporary']);
+            }
+
+            // expectedClassifications
+            if (array_key_exists('expectedClassifications', $filters)) {
+                User::filterByExpectedClassifications($userQuery, $filters['expectedClassifications']);
+            }
+
+            // skills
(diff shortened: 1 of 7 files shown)


api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
```diff
@@ -0,0 +1,103 @@
+<?php
+
+namespace App\GraphQL\Queries;
+
+use App\Models\PoolCandidate;
+use App\Models\User;
+use Database\Helpers\ApiEnums;
```
mnigh, line 7: Symbol 'App\GraphQL\Queries\ApiEnums' is declared but not used.

api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
```diff
@@ -0,0 +1,103 @@
...
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Support\Facades\Log;
```
mnigh, line 9: Symbol 'App\GraphQL\Queries\Log' is declared but not used.

api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
```diff
@@ -0,0 +1,103 @@
...
+            PoolCandidate::filterByPools($queryBuilder, $pools);
+        }
+
+        $queryBuilder->whereHas('user', function (Builder $userQuery) use ($filters) {
+            // user filters go here
+
+            // user status scope
+            User::scopeAvailableForOpportunities($userQuery);
+
+            // has diploma
```
mnigh, line 42:
```suggestion
            // hasDiploma
```

api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
```diff
@@ -0,0 +1,103 @@
...
+            }
+        });
+
+        $databaseRows = $queryBuilder
+            ->selectRaw('count(*) as candidate_count, pool_id')
+            ->groupBy('pool_id')
+            ->get();
+
+        // would be nice to do this in the same query as above but alas...
+        $allPools = Pool::whereIn('id', $databaseRows->pluck('pool_id'))->get();
```
esizer, line 89: If I understand what you are getting at, I believe you can do this with the same query by [eager loading](https://laravel.com/docs/9.x/eloquent-relationships#eager-loading) the relationship. I think something like this would work:

```php
return $queryBuilder
  ->with('pool')
  ->select('pool_id', DB::raw('count(*) as candidate_count'))
  ->groupBy('pool_id')
  ->get();
```

You can then tell laravel to change the `candidate_count` to `candidateCount` in the `PoolCandidate` model with:

```php
public function getCandidateCountAttribute()
{
    return $this->attributes['candidate_count'];
}
```

esizer: I'm not 100% sure this produces the correct result, but it might help get you on the right track to making it a single query hopefully :)

esizer (changes requested):
Looking good! One small suggestion and maybe some help with the query 🤷‍♂️

mnigh (changes requested):
Pretty cool stuff! A few small suggestions.

Maybe I don't understand our filtering correctly, but shouldn't these two screenshot counts for each total per pool add up to another total for the empty where?

api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
petertgiles: Thanks!  Fixed in [Tidy up based on reviews](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4455/commits/74587e85db9aeb7ae9cd36053f21c2d005a4081a).

api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
petertgiles: Thanks!  Fixed in [Tidy up based on reviews](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4455/commits/74587e85db9aeb7ae9cd36053f21c2d005a4081a).

api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
petertgiles: Thanks!  Fixed in [Tidy up based on reviews](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4455/commits/74587e85db9aeb7ae9cd36053f21c2d005a4081a).

petertgiles:
> Maybe I don't understand our filtering correctly, but shouldn't Screenshot 1 and Screenshot 2 counts for each total per pool add up to another total for the empty where (Screenshot 3)?

Oh man, that's a classic gotcha of our API.
isWoman: true -> filter on rows with isWoman = true
isWoman: false -> don't filter on rows with isWoman = true
isWoman: undefined -> don't filter on rows with isWoman = true

There's no way to request rows where isWoman = false.  It's just "filter on =true" or don't "filter on =true".
It's because it's tightly coupled to our search page which represents that option with a checkbox where there's no way to "undefine" it.

api/tests/Feature/CountPoolCandidatesByPoolTest.php
```diff
@@ -0,0 +1,739 @@
...
+    // test skills
+    // creates a three users various skills and filter for the skills on two of them
+    public function testSkills()
+    {
+        $pool = Pool::factory()->create();
+        $skills = Skill::factory(3)->create();
+        $users = User::factory(3)
+            ->afterCreating(function ($user) use ($pool) {
+                $exp = AwardExperience::factory()->create(['user_id' => $user->id]);
+                print_r($exp);
```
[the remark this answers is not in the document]
petertgiles: Thanks! Fixed in [Tidy up based on reviews](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4455/commits/74587e85db9aeb7ae9cd36053f21c2d005a4081a).

api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
petertgiles: Whoah, this works!  Yer a wizard!  :mage: 

It actually still queries the database a second time to pull in the pools but the code is so much cleaner now that the framework does it for me.  I'll take it as a win.  :smile:  Thanks!

[Use query builder to attach pools](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4455/commits/c1ae4ec91616caefe3ad6a84e5441275edcc257f)

mnigh (approved):
LGTM 👍

PR #4486, opened 2022-10-28
Restrict skills list to only "need to have" and "nice to have" skills during an application

### LINKED ISSUE #4365: Restrict skills list to only "Need to have" and "nice to have" skills during an application
When applying to a job, the skills listed on the Experience form page should be limited to only the Need to have and Nice to have skills associated with the job. This means we no longer need an elaborate skill picker, or any search functionality, since the list will be small.

Note: this is only when going through an application! When coming from your profile, there _are_ no asset/required skills, and we should continue to use the SkillPicker component.

restricted skill list

[Link to Figma prototype](https://www.figma.com/file/spDwyr56BgQKrSdRiHztPL/Skills-and-EX---Poster-and-application-V2?node-id=1150%3A7498)

## Acceptance Criteria
- [x] During an application, Step 2 of the Experience form lists Need to have and Nice to have skills, instead of the skill picker component.
- [x] Skills listed use the same skill accordion component as the skill picker, and the "add skill" button works the same.
- [ ] When editing an existing Experience, we want to hide skills unrelated to this job, but ENSURE THEY ARE NOT DELETED when updating the skills for this job. (Need to be careful about how we use `connect` and `sync` on `UpdateExperienceSkills`.)
- [x] Accessing the Experience form via the user profile continues to use the Skill Picker component allowing user to search all skills.

JamesHuf:
Resolves #4365.

Changes skill selection in the experience forms to only "need to have" and "nice to have" skills when applying to a pool. The skill selection should remain the same while viewing the form from the users profile.

Note: This PR contains 2 new strings I still need to get translated.

To test:
1. From your profile navigate to the edit page for an existing experience and the add skills section appears unchanged
2. Start a pool application
3. Navigate to an edit experience page from the new application and confirm the add skills section has changed to only include skills marked as need/nice-to-have for the pool advertisement
4. Add and remove a couple skills for the experience - confirm it works as expected.
5. Save the form and confirm everything is saved as expected


frontend/talentsearch/src/js/components/experienceForm/ExperienceSkills.tsx
```diff
@@ -51,6 +57,18 @@ const ExperienceSkills: React.FC<ExperienceSkillsProps> = ({ skills }) => {
...
 
+  const handleAddSkill = (id: string) => {
+    const skillToAdd = skills.find((skill) => skill.id === id);
+
+    if (skillToAdd) {
+      append({
+        skillId: skillToAdd.id,
+        name: skillToAdd.name,
+        details: "",
+      });
```
esizer, line 68: I'm going to leave the descision up to @Jerryescandon , but I think we may prefer to add `{shouldFocus: false}` to the `append`.

@Jerryescandon What do you think? By default, when you add a skill it jumps down the to details section to start writing, do we want to prevent that so users don't need to scroll back up to add more skills?

esizer (commented):
Looking really good! 👍 Just have one small question about the focus, once we figure that out, i'm ready to approve.

esizer:
> Looking really good! 👍 Just have one small question about the focus, once we figure that out, i'm ready to approve.

Also, I'm sorry if this wasn't ready for review yet. Just noticing now it is still in progress 😅

esizer (approved):
Nice! 😁 This makes applying so much easier 🤯

PR #4136, opened 2022-09-30
Debt - 3952 - Replace primitives

esizer:
Resolves #3952 

## 📋 Summary

This replaces our underlying custom/reach `common` primitives with `radix-ui`

## ❓Rationale

> **Warning**
> [Reach is not currently maintained.](https://github.com/reach/reach-ui/issues/972)

Currently, we are stuck on `react@17` and the only real blocker is `reach-ui`. After some research and looking at a number of primitive libraries, it seems as though radix is the best option because:

- It has full keyboard navigation
- Managed focus
- Has proper semantics for assistive technology
- WAI-ARIA compliant
- Screen reader tested
- Allows us to worry about styling rather than API
- Exposes all sub-components for more control
- Has a fairly large library to use

## ✅ TO DO

- [ ] Fix linting
- [ ] Translation run

## 🧪 Testing

> Good luck! 🍀 

1. Build all projects `npm run production --workspaces`
2. Bounce around the place where the components were replaced and confirm they all still work and look fine
3. Check storybook for components and docs! `npm run storybook`

### Places to Check

- Admin
  - Logout
  - Edit Pool
  - Table column dialog
  - User/Pool Candidate filter dialogs
  - API Table search dropdown
  - User tabs
- IAP
  - Dialogs on homepage
- Talent Search
  - Employment equity form dialogs
  - Experience form delete confirmation
  - Logout
  - Role Salary form dialogs

## ⚛️ Components

| Component     | Created | Stories | Tests | Replaced |
| ------------- | ------- | --------- | ----- | -------- |
| `Accordion`     |     ✅     |      ✅     |    ✅   |     ✅     |
| `AlertDialog`  |    ✅     |     ✅      |   ✅    |      ✅    |
| `Collapsible`        |    ✅     |      ✅     |    ✅   |      ✅      |
| `Dialog`        |    ✅     |      ✅     |    ✅   |      ✅      |
| `DropdownMenu` |     ✅    |      ✅     |    ✅     |    ✅      |
| `ScrollArea` |     ✅    |      ✅     |    ✅     |    ✅      |
| `Separator` |     ✅    |      ✅     |    ✅     |    ✅      |
| `Switch` |     ✅    |      ✅     |    ✅     |    ✅      |
| `Tabs`          |   ✅       |     ✅       |     ✅   |    ✅      |
| `ToggleGroup` |     ✅    |      ✅     |    ✅     |    ✅      |

### Accordion

```tsx
import Accordion from "@common/components/Accordion";

/**
[...]

PR #4520, opened 2022-10-31
Updates Hydrogen to beta.31 in preparation for theming

substrae:
# ✨ Purpose

Updates the Hydrogen dependency so that we can begin working with themes. This update also includes support for automatic dark mode once we define sibling values.

# 🧪 Changes
- [x] Update the Hydrogen dependency
- [x] Migrate the configuration file
- [ ] If possible, leverage the new promise response in the Webpack compiler hooks
  - Turns out this isn't possible right now, due to the same reasons we struggled with Webpack's hooks in the past 
- [x] Fix breaking opacity changes

# 📑 Related issues
- https://github.com/hydrogen-css/hydrogen/pull/199
- https://github.com/hydrogen-css/hydrogen/pull/202
- https://github.com/hydrogen-css/hydrogen/pull/203


petertgiles (commented):
I'm seeing some sus changes in Chromatic.  Are these intentional?
image

image

image

image

substrae:
Thanks @petertgiles! Taking a peek.

substrae:
Oops, noticed one last thing to fix that was breaking that I forgot about - color variable exports.

petertgiles:
Nice!  Chromatic is down to just a few diffs now.

The first one I probably wouldn't have noticed and I don't know if it's worth worrying about.  It seems like it noticed some green jaggies on the edges of the diagonal gradients:
image
I can't even replicate it in my browser on the actual page.

The second is that our Hero lost the placeholder image.
image
It doesn't seem to be a problem on the actual Browse Opportunities page.
image

Just a small personal preference, is there a way to make the Hydrogen output less noisy?  Unless we're trying to debug a problem with the tool this seems like a lot.  :laughing: Single-spaced start and stop messages would be enough.
image

substrae:
@petertgiles
- The green jaggies are likely differences in subpixel rendering, because nothing would have changed there
- I can't seem to replicate the hero issue... opening Storybook from the build above shows me the 500x500 image placeholder like your first screenshot
  - image 
- And yes! The system messages can't be silenced at the moment, but the timers can, so I'll pop a commit up for that

esizer (approved):
LGTM! 👍 

Can't wait to go in and attempt to replace all the custom css in https://github.com/GCTC-NTGC/gc-digital-talent/pull/4136 once it is finally merged 😀

PR #4522, opened 2022-10-31
Bug - Lock hydrogen version to beta.29

esizer:
## Summary

We were allowing hydrogen to auto-update which was causing problems while it is in beta. This locks it to specific versions until `2.0.0` is released.


substrae (approved):
Awesome, thanks Eric!

PR #4571, opened 2022-11-03
Bug - Fix Classification on summary of filters

### LINKED ISSUE #4500: "Group and Level" always empty in Summary of Filters
PR #4454 changes how the Classification dropdown on the search page works. Instead of applying a filter directly to users' expected-classifications or expected-salary, it will now be used to filter which pools the searcher sees and can make requests for.

However, that's not reflected in the Summary of Filters component, which now will always be blank for "Group and Level":

We need some design feedback on how to deal with this. My suggestion is to hide the "Group and Level" block if the filter doesn't have any (that way viewing old requests still works) and replace it with the Name or description of any Pools we're searching in.

## Acceptance Criteria
- [ ] Instead of showing classifications in the summary based on the "expectedClassifications" field of the filter (which has essentially been deprecated) lets show the classifications associated with any pools used in the filter.

yonikid15:
Resolves #4500 

## Notes
- I had to add the `selectedClassification` (from the search page filters) to the browser history, since the `PoolCandidateSearchRequest` isn't created till after they submit a request.
- On the other hand, on the admin side. When you view the Request the classification(s) are fetched from pool(s).

## Testing
- Go to Search page, fill out the form and request some candidates.
- Check to see if the Classification has value. If there is no value then request changes and let me know I messed up. If there is a value then fill out the Request form and hit "Submit Request".
- If you are back on the Search page and the mutation succeeded, go to the admin side and view your new request. Now, check  the "Request Information" section to see if the Classification has the same value. If there is no value, request changes and let me know I messed up.


yonikid15:
Alright, looks like its ready to go

vd1992:
I feel the logic of classification filters count will need to change, it is going to be removed from Search Advice but is used elsewhere

image

you can only pick one classification and you must pick one
it appears to be counting pools returned instead

image

vd1992:
I mean that it is counting pools returned and not classifications selected which makes for a bit of a confusingly named variable.
It rendered the filters (3) variable above, I could only select one classification so the 3 comes from the pools returned. 
This is more a variable naming matter than actual functionality. Hopefully I explained myself here lol

petertgiles (approved):
Seems good to me.  :+1: It works well when I tried it.

Using the browser history to move that across is a bit gnarly but that's the way the other filters worked all along so it makes sense to keep using it.  Also the candidate list when viewing the request in admin is bringing back way too many results but I think you're aware of that and it's out of scope right?

vd1992:
ah, if @petertgiles doesn't think it is really an issue, then nevermind lol

petertgiles:
> ah, if @petertgiles doesn't think it is really an issue, then nevermind lol

Ah ah.  Just because you left comments instead of requesting changes doesn't mean I'm going to double-check your "not-requests" for you.  :laughing:

PR #4586, opened 2022-11-03
Bug - Adds Link text to go to IAP homepage on Browse IT jobs page

### LINKED ISSUE #4558: Bug - Link text to go to IAP homepage on browse IT jobs page
## Description
The link text to go to IAP homepage on browse IT jobs page does not include an action and does not lend itself to be translated well with `<hidden>` tags (used for providing context for screen readers) into French.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/66a95cffb5c7a8ffb27a7ba82a775cf255a7054d/frontend/talentsearch/src/js/components/Browse/BrowsePoolsPage.tsx#L281

## Screenshot
Screen Shot 2022-11-02 at 12 23 13

## Acceptance Criteria
- [ ] button text should fit in with nearby **Visit the talent search page**
- [ ] button text should avoid using `<hidden>`

mnigh:
Resolves #4558.

## Steps to test

1. Compile talentsearch `npm run intl-compile --workspace=talentsearch`
2. Build talentsearch `npm run production --workspace=talentsearch`
3. Navigate to /en/browse/pools
4. Confirm there is a button with the text **Apply now**
5. Navigate to /en/browse/pools
6. Confirm there is a button with the text **Inscrivez-vous maintenant**


frontend/talentsearch/src/js/components/Browse/BrowsePoolsPage.tsx
```diff
@@ -278,10 +278,10 @@ export const BrowsePools: React.FC<BrowsePoolsProps> = ({
                 external: true,
                 label: intl.formatMessage({
                   defaultMessage:
-                    "<hidden>Learn more about </hidden>Indigenous IT apprenticeship<hidden> opportunities</hidden>",
-                  id: "xgocDD",
+                    "Apply<hidden> to the Indigenous Apprenticeship Program</hidden> now",
```
petertgiles, line 281: The leading space inside the hidden tags seems to get chomped in Firefox.
image
When I changed it to `"Apply <hidden>to the Indigenous Apprenticeship Program </hidden>now"`
image
I'm not sure if this is the best solution or if it works the same way in different browsers.

It looks like we may have this issue in a number of places.
image

petertgiles (changes requested):
Much better string!  Looks like it needs a little tweak though.

frontend/talentsearch/src/js/components/Browse/BrowsePoolsPage.tsx
mnigh: Good catch @petertgiles. This indeed is an issue in other places as well in Firefox (but not Chrome).

### Screenshots

Screen Shot 2022-11-07 at 09 58 42
Screen Shot 2022-11-07 at 09 56 12

Strangely, this similar string does not have these same problems in Chrome or Firefox:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/8d57e6e41f7a3afc98c40703000f30878642a93c/frontend/talentsearch/src/js/components/employmentEquityForm/EquityOption.tsx#L34

@esizer is it because the text within the hidden tag cannot begin with a space? Is this a known Firefox bug?

mnigh: I've created a new issue for the other strings since those are out of scope: https://github.com/GCTC-NTGC/gc-digital-talent/issues/4619

esizer: My best guess is that since it is a `span` and by default has `display: inline;` and Firefox will collapse whitespace inside an inline element.

> And finally, sequences of spaces at the beginning and end of an element are removed...

REF: https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Whitespace#:~:text=And%20finally%2C%20sequences%20of%20spaces%20at%20the%20beginning%20and%20end%20of%20an%20element%20are%20removed%2C%20so%20we%20finally%20get%20this%3A

mnigh: I tried this `Apply <hidden>to the Indigenous Apprenticeship Program </hidden>now` and it is formatted correctly in the a11y inspector but for some reason, VoiceOver reads it in the wrong order as: **to the indigenous Apprenticeship Program Apply now** 😕. Any ideas?

Screen Shot 2022-11-07 at 10 32 52

petertgiles: Oof, I'm guessing this is also part of a much larger issue (and scope).  I'd love to get this PR merged it to green up our CI.  Thoughts on spinning off this issue and merging the PR as is?

mnigh: Yes, it is definitely part of a larger issue. Let's go green and worry about it as part of: 

- https://github.com/GCTC-NTGC/gc-digital-talent/issues/4619

petertgiles (approved):

frontend/talentsearch/src/js/components/Browse/BrowsePoolsPage.tsx
esizer: Looks like this is a [known issue](https://github.com/h5bp/main.css/issues/12) with VoiceOver when `margin: -1;`....

Seems like [it is required](https://github.com/twbs/bootstrap/issues/25686) though, so it may be out of our control until VoiceOver has a fix for this.

PR #4599, opened 2022-11-05
Update job poster title

### LINKED ISSUE #4574: Job Poster title
Re-ticketed from #4482.

Pool poster title should include Specific title. I think we can use #4479 as guidance for the Pool title.
ie [Specific title] + " ( " + [Classification] + [Stream/Job title] + " ) "

image

JamesHuf:
Resolves #4574.

Updates the title on the job poster page to use the `formattedPoolPosterTitle` function.

To test:
Navigate to a job posting (`/browse/pools/{id}`) and confirm the title displays in the expected format. Ex: `Pool name ( EX-01 Pool Stream )`


frontend/talentsearch/src/js/components/pool/PoolAdvertisementPage.tsx
```diff
@@ -131,8 +134,19 @@ const PoolAdvertisement = ({
   const genericTitle = classification?.genericJobTitles?.length
     ? classification.genericJobTitles[0]
     : null;
-  const classificationSuffix = `${classification?.group}-0${classification?.level}`;
-  const fullTitle = getFullPoolAdvertisementTitle(intl, poolAdvertisement);
```
yonikid15, line 135: We should probably consolidate the two functions `formattedPoolPosterTitle` and `getFullPoolAdvertisementTitle`, since they are both meant to return the same string. Also, `getFullPoolAdvertisementTitle` is used across the "Apply to a pool" process so we need to replace those as well.

I like that `getFullPoolAdvertisementTitle` does the heavy lifting and only requires the `poolAdvertisement`, so I'm thinking bring the logic you wrote on this file to that function. Then let it live in the `poolUtils.ts` helper file in common. The function would be something like this?

```ts
import { getPoolStream } from "@common/constants/localizedConstants";
import { getLocalizedName } from "@common/helpers/localize";
import { formatClassificationString } from "@common/helpers/poolUtils";
import type { IntlShape } from "react-intl";
import { Maybe, PoolAdvertisement } from "../../api/generated";

const getFullPoolAdvertisementTitle = (
  intl: IntlShape,
  advertisement: Maybe<PoolAdvertisement>,
): string => {
  if (advertisement === null || advertisement === undefined)
    return intl.formatMessage({
      id: "D91nGW",
      defaultMessage: "Job title not found.",
      description:
        "Message shown to user when pool name or classification are not found.",
    });

  const classification = advertisement.classifications
    ? advertisement.classifications[0]
    : null;

  let classificationSuffix = ""; // type wrangling the complex type into a string
  if (classification) {
    classificationSuffix = formatClassificationString({
      group: classification?.group,
      level: classification?.level,
    });
  }
  const genericTitle = getLocalizedName(advertisement.name, intl);

  return `${genericTitle ? `${genericTitle} ` : ""}(${classificationSuffix}${
    advertisement.stream
      ? ` ${intl.formatMessage(getPoolStream(advertisement.stream))}`
      : ""
  })`;
};

export default getFullPoolAdvertisementTitle;
```

yonikid15 (changes requested):
Nice work! Everything's working on my end, but I noticed a duplicate function that's already being used across the Apply to a pool process. I left a comment below.

frontend/talentsearch/src/js/components/pool/PoolAdvertisementPage.tsx
JamesHuf: Good call. I wasn't sure if they were supposed to return the same thing or not.

JamesHuf: Both functions have now been consolidated into a new `getFullPoolAdvertisementTitle` function located in `poolUtils.ts`.

frontend/admin/src/js/components/pool/PoolTable.tsx
```diff
@@ -50,30 +46,14 @@ function poolCandidatesLinkAccessor(
...
-        classification: classificationString,
-        stream,
-        intl,
+    <Link href={`${editUrlRoot}/${pool.id}`} type="link">
+      {getFullPoolAdvertisementTitle(intl, {
+        id: pool.id,
+        name: pool.name,
+        classifications: pool.classifications,
+        stream: pool.stream,
       })}
```
tristan-orourke, line 57: We can pass the pool in directly instead of creating a new object, since its already an object with all the (properly typed!) fields:
```suggestion
      {getFullPoolAdvertisementTitle(intl, pool)}
```

frontend/common/src/helpers/poolUtils.ts
```diff
@@ -31,3 +14,35 @@ export const formatClassificationString = ({
...
+    : null;
+
+  let classificationSuffix = ""; // type wrangling the complex type into a string
+  if (classification) {
+    classificationSuffix = formatClassificationString({
+      group: classification?.group,
+      level: classification?.level,
+    });
+  }
+  const genericTitle = getLocalizedName(poolAdvertisement.name, intl);
```
tristan-orourke, line 41: I'm pretty sure `poolAdvertisement.name` is considered the "Specific title", and the Classification + Stream is the "Generic" part of the title.

tristan-orourke (commented):
Looks good. Just left a couple code suggestions.

yonikid15 (approved):
Awesome, looks good👍 Nice work!

yonikid15 (commented):
Oops, sorry just need to fix that translation error then were all set!

tristan-orourke (approved):
Looks good, go ahead and merge.

PR #4661, opened 2022-11-08
4660 language ability null

### LINKED ISSUE #4660: ApplicantFilterInput.languageAbility null vs undefined provide different results
For ApplicantFilter and PoolCandidateFilter inputs (and graphql inputs in general), a `null` field should be considered equal to an undefined field. However, when passing an ApplicantFilterInput object to the countApplicants query, setting `{languageAbility: null}` gives different results than leaving it undefined!

tristan-orourke:
Resolves #4660.

I wrote a regression test for this! Which actually found a similar bug in a directive I wrote.

Testing:
1. seed the database (`docker-compose exec -w /var/www/html/api php php artisan db:seed`)
2. Run the following two queries against our api. I used http://localhost:8000/graphql-playground.
```
query {
  countApplicants(where:{
    languageAbility: null
  })
}
``` 
```
query {
  countApplicants(where:{})
}
``` 
Or you can run them with one query using _aliases_ 😉 
```
query {
  nullLang: countApplicants(where:{
    languageAbility: null
  })
  undefinedLang: countApplicants(where:{})
}
```
3. Both queries should return the same number. (And it shouldn't be zero!)

### DIFF
--- a/api/app/GraphQL/Directives/PluckDirective.php
+++ b/api/app/GraphQL/Directives/PluckDirective.php
@@ -30,9 +30,6 @@ directive @pluck(
 GRAPHQL;
     }
 
-    /**
-     * Remove whitespace from the beginning and end of a given input.
-     */
     public function transform($argumentValue)
     {
         $output = Utils::mapEach(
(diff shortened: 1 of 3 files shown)


vd1992 (approved):
Nice!
Might be good to run a similar test on the other models like PoolCandidate, but that is outside of this issue

image

vd1992:
Hang on, should this be rectified too?

image

tristan-orourke:
> Hang on, should this be rectified too?
> 
> image

Good catch, I guess that can also be fixed here. And it should get its own regression test.

PR #4665, opened 2022-11-08
mirror the null + undefined testing done for User on PoolCandidate

vd1992:
spun out of #4661 
copies that for the other model with paginated results

make sure tests pass!


brindasasi (approved):
LGTM

PR #4669, opened 2022-11-08
Bug - Adds missing wouldAcceptTemporary to CreateRequest

### LINKED ISSUE #4659: wouldAcceptTemporary is not being saved in ApplicantSearch Requests
On the search page, if I set duration to "Term", the countApplicants query happens with `{wouldAcceptTemporary: true}`. (In other cases, wouldAcceptTemporary is null). However, when I go submit the request and check the request in the database, the associated ApplicantFilter has wouldAcceptTemporary equal to null - the correct value was not saved to the database.

mnigh:
Resolves #4659.

## Summary

- [Adds missing wouldAcceptTemporary to CreateRequest](https://github.com/GCTC-NTGC/gc-digital-talent/commit/3988933b24423fd89c6095e3498d7536c64b1d35)
- [Fixes would_accept_temporary to be true or null since false is never an option based on the CreateRequest form value](https://github.com/GCTC-NTGC/gc-digital-talent/commit/a6d9ea45e23cc3318d45319aed98417883355c4a)

### DIFF
--- a/api/database/factories/ApplicantFilterFactory.php
+++ b/api/database/factories/ApplicantFilterFactory.php
@@ -31,7 +31,7 @@ class ApplicantFilterFactory extends Factory
             'is_indigenous' => $this->faker->boolean(),
             'is_visible_minority' => $this->faker->boolean(),
             'is_woman' => $this->faker->boolean(),
-            'would_accept_temporary' => $this->faker->boolean(),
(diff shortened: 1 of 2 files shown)


esizer (approved):
Awesome! 🎉  Everything working well for me.

**Out of scope**: Seems odd to use "Term duration (permanent)" when the user selects "Indeterminate duration (permanent)" on the search form. Do you think we should create an issue to sync up terminology from `talentsearch` and `admin`?

PR #4707, opened 2022-11-10
Feature/4701 modify certain user emails artisan command

### LINKED ISSUE #4701: Create an artisan command which modifies emails for users without gc_key sub
Re-ticketed from #4443 based on discussions.

Create an artisan command which will modify the contact email of users who cannot log in. This will allow users to register again with those same emails in the future, if they want.

We can determine users without a login by looking for users where either the `sub` field is null, or the `sub` is the same as the user's email (which we were using as a default at one point). 

We should replace the email for these users with a version using "plus addressing". We pick a tag and add add it to an email like so: `address@domain.com` becomes `address+tag@domain.com`.

QUESTION: what tag should we use?
- dcmo?
- admin_entry
- ...

vd1992:
closes #4701 
adds an artisan command to add a tag to certain user emails

testing:
tests pass
run command locally, open up adminer and ensure only users with email == sub or sub == null were updated


vd1992:
`update:obsolete_user_email`

PR #4710, opened 2022-11-10
Adds an issue template for component epics

substrae:


petertgiles:
This is super nice!  I'm a little confused though since the example issue linked is not an epic.  If this template is just for designers to describe epics then it's great and I'll approve.  If it's actually for dev issues then it will need some dev acceptance criteria like a storybook story, jest tests, etc.

substrae:
@petertgiles Oh sorry - that's likely because none of us use Zenhub, we're just using the default Projects features in Github. If tacking the epic label on is helpful for y'all, I can absolutely do that!

In terms of who it's for, these are going to kind of be a weird handoff issue. I want them to be a centralized place to track design decisions for the design team, but I also want them to be the place the dev team can come to for the most up-to-date component designs AND for a contextual history of why things are the way they are. 

When a new component is built or we decide updates are needed, we're still going to break that work into smaller dev task issues for your Zenhub board (where I assume the acceptance criteria will live), and they should be linked back to these giant component parent issues for documentation purposes. That way, you guys can track your work at the scale you're used to, but we also have this macro level perception of the work that's been done over time.

petertgiles (approved):
Nice, let's do it!  I'd like to add the epic label, too, if it's no bother.

PR #4721, opened 2022-11-11
Design - Update toast styles

esizer:
## 👋 Introduction

We recently updated our alert styles, this makes similar design updates to the toasts to match those new alerts.

## ✅ TO DO

- [x] Replace existing toasts

## 🛠️ Usage

The new styles are used in the same way, the difference is where you import the `toast` function from.

```tsx
import { toast } from "@common/components/Toast" // Note: This is now being imported from common

toast.info("text");
toast.success("text");
toast.warning("text");
toast.error("text");
```

## 📷 Screenshot

Screenshot 2022-11-14 at 10 57 00 AM

## 🧪 Testing

1. Run storybook `npm run storybook --workspace=common`
2. Confirm the toast story functions as expected
3. Navigate through the app, triggering toasts to confirm they have been updated


petertgiles:
I just merged an update to react-toastify.  It might be worth merging main.

petertgiles (changes requested):
This works great! Almost!  So this one is pretty subjective but I think we're going to run into trouble with narrow viewports.  I don't think we've ever officially decided on the most narrow we'll support.  But if the page is wider than the viewport then it's possible for the toast to be mostly or fully hidden which is a pretty big problem for important notifications.  I guess it could be argued that the real problem is that the page is wider than the viewport and not responsive.  :laughing: 

I'm not sure of the solution.  Bottom-left instead of bottom-right?  This is maybe a discussion for designers instead.  :shrug: 

https://user-images.githubusercontent.com/8978655/202481989-06ac3105-515a-4a07-934e-41a7594541fe.mp4

esizer:
> This works great! Almost! So this one is pretty subjective but I think we're going to run into trouble with narrow viewports. I don't think we've ever officially decided on the most narrow we'll support. But if the page is wider than the viewport then it's possible for the toast to be mostly or fully hidden which is a pretty big problem for important notifications. I guess it could be argued that the real problem is that the page is wider than the viewport and not responsive. 😆
> 
> I'm not sure of the solution. Bottom-left instead of bottom-right? This is maybe a discussion for designers instead. 🤷

Good catch, looks like we just needed to add a media query to handle smaller screens better here. Fixed with [update container width for better responsiveness](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4721/commits/91eaaec06a518e871b270ef5b8e82f269c69cefe)

petertgiles (approved):
Nice!  At the narrowest widths it might not be possible to dismiss the toast but this looks very reasonable to me.

PR #4763, opened 2022-11-15
Bug - Edit Pool skill picker submit button text

### LINKED ISSUE #4576: Save Buttons dont change state
## Description
In the Pool edit page, after you select essential or asset skill and hit the save button the save button will not change to the submitted state, unlike the other save buttons on this page.

## Screenshot: Prototype
image

## Acceptance Criteria
- [ ] ...

yonikid15:
Resolves #4576 

## Notes
- Fixes the submit buttons on the Essential and Asset skill sections found in the edit pool page.
- Fixed by wrapping the entire SkillPicker contents in the `<FormProvider />` (even if not part of the form). I'm currently not sure if this is an accessibility risk, would like insight if so.

## Testing
- Go to [Pools page](http://localhost:8000/en/admin/pools) on admin after logging in.
- Find a pool that is still a draft and click edit OR create a new pool.
- Scroll down to Essential and Asset skills section and try Adding/Removing skills. Ensure form works correctly.


yonikid15:
Damn, I thought the `<SkillPicker />` component was only used on the admin side. Figuring out a new solution to be handled on TalentSearch right now.

yonikid15:
Okay, I found a compromise btw the two apps (admin and talentsearch) using the `<SkillPicker /> component. It's definitely not the best solution, but let me know what you think. 

I added a `submitButton` prop to the SkillPicker that is wrapped by a `<form />` component (within the `<FormProvider />`). However, the submit button is the only element within the form, which seems like an accessibility nono. I'll searching for a better solution.

frontend/common/src/components/SkillPicker/SkillPicker.tsx
```diff
@@ -241,7 +244,10 @@ const SkillPicker = ({
           </Chips>
         </>
       ) : null}
-    </>
+      {submitButton && handleSave && (
+        <form onSubmit={handleSubmit(handleSave)}>{submitButton}</form>
```
esizer, line 248: Is it required to use a `<form>` element here? Could not just remove the form and have the button passed in call `onSave` instea?

yonikid15: So ya it is required for the Submit button text to work, since it uses the [`useForm()`](https://react-hook-form.com/api/useform/handlesubmit/) hook. I've tried some other ways including the way you suggested, but it didn't work. I'm still reading through the docs for a better solution.

esizer (approved):
Nice! Looking good 🎉 

One small thing I noticed is that we may want to update `Submit.tsxL46` to `if(isSubmitting || overrideSubmitting)` so the text says "Submitting" while it works. Though, I think that is out of scope for this issue so i won't let that block this.

You can add it here if you want. I'll approve anyway, if you don't add it I can just create a ticket 👍 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/92cb664302ccfad3944a1721abcc17dad55d681c/frontend/common/src/components/form/Submit/Submit.tsx#L46

frontend/common/src/components/SkillPicker/SkillPicker.tsx
yonikid15: Turns out I just needed to use onClick instead of onSubmit 😆

yonikid15:
> Nice! Looking good 🎉
> 
> One small thing I noticed is that we may want to update `Submit.tsxL46` to `if(isSubmitting || overrideSubmitting)` so the text says "Submitting" while it works. Though, I think that is out of scope for this issue so i won't let that block this.
> 

Okay, I pushed that last change up! Thanks for the review :)

PR #4815, opened 2022-11-18
Bug -  4786 - Improve code splitting

### LINKED ISSUE #4786: Code splitting causing `ChunkLoadError`
## 👋 Introduction

After updating the app the utilize code splitting, we some times encounter a `ChunkLoadError` which is caused when the app attempts to load a chunk that no longer exists. This **_should_** be happening infrequently in prod since it happens when a chunk no longer exists or the app has trouble loading.

## ❓ Possible Solution

REF: https://www.codemzy.com/blog/fix-chunkloaderror-react

We can implement a retry function as well as assigning names to our chunks to help stabilize the loading of chunks. The naming will also help in understanding which chunk is having trouble loading.

## ✅ Acceptance Criteria

- [ ] Name our chunks using `webpackChunkName` comments
- [ ] Implement `retryLazy` function (See: https://www.codemzy.com/blog/fix-chunkloaderror-react) 

---

        > I occasionally get these Chunk errors, but not always and refreshing fixes it and stops it from re-occurring

So, I think we noticed this on main as well and it less to do with this specific PR and more to do with code splitting in general. I will create a ticket that will hopefully fix this.

REF: https://www.codemzy.com/blog/fix-chunkloaderror-react

_Originally posted by @esizer in https://github.com/GCTC-NTGC/gc-digital-talent/issues/4597#issuecomment-1318637652_

esizer:
## 👋 Introduction

This implements a retry function for dynamic imports to help prevent `ChunkLoadError`

## 🕵️ Details

When we serve our chunks, they are cached in the browser. The `ChunkLoadError` occurs when the browser looks for the file and it does not exists for any number of reasons (original cached file no longer exists, renamed, etc.).

This PR does two things

1. Names our chunks (previously randomly generated)
2. Implements a retry function for loading chunks

### Naming Chunks

By naming chunks, we make debugging easier since file names are human readable and mirror the component name. Also, when the browser caches the file, the name remains the same for all builds so it should be able to find it more consistently.

### Retry Lazy

In the event that the file could not be found we force a browser refresh to hopefully clear the cache and look for the file again. We story this in locale storage to prevent an infinite loop in the case that the file actually does not exist. If it really does not exist, we throw an error and display the `errorElement` with an unknown error.

## 🧪 Testing

1. Build all projects `npm run production --workspaces`
2. Confirm Admin and Talent Search function normally
3. Open the `dist` folder and delete a `*.js` file of a page
4. Visit that page and make sure a nice error message appears
5. Watch the locale storage and confirm `component-refreshed` updates as we look for the component

## 🤖 Robot Stuff

Resolves #4786

### DIFF
--- a/frontend/admin/src/js/components/Router.tsx
+++ b/frontend/admin/src/js/components/Router.tsx
@@ -181,7 +181,7 @@ const DeprecatedCreatePool = React.lazy(() =>
   lazyRetry(
     () =>
       import(
-        /* webpackChunkName: "adminCreatePool" */ "./pool/deprecated/CreatePool"
+        /* webpackChunkName: "adminDeprecatedCreatePool" */ "./pool/deprecated/CreatePool"
       ),
   ),
 );
--- a/frontend/admin/src/js/components/Router.tsx
+++ b/frontend/admin/src/js/components/Router.tsx
@@ -173,7 +173,7 @@ const DeprecatedUpdatePool = React.lazy(() =>
   lazyRetry(
     () =>
       import(
-        /* webpackChunkName: "adminUpdatePool" */ "./pool/deprecated/UpdatePool"
+        /* webpackChunkName: "adminDeprecatedUpdatePool" */ "./pool/deprecated/UpdatePool"
       ),
   ),
 );
--- a/frontend/admin/src/js/components/Router.tsx
+++ b/frontend/admin/src/js/components/Router.tsx
@@ -9,92 +9,197 @@ import Layout from "./Layout";
 
 import { Role } from "../api/generated";
 
-const HomePage = React.lazy(() => import("./home/HomePage"));
-const ErrorPage = React.lazy(() => import("./Error/ErrorPage"));
-const DashboardPage = React.lazy(() => import("./dashboard/DashboardPage"));
+const HomePage = React.lazy(
+  () => import(/* webpackChunkName: "adminHomePage" */ "./home/HomePage"),
+);
+const ErrorPage = React.lazy(
+  () => import(/* webpackChunkName: "adminErrorPage" */ "./Error/ErrorPage"),
+);
+const DashboardPage = React.lazy(
+  () =>
+    import(
+      /* webpackChunkName: "adminDashboardPage" */ "./dashboard/DashboardPage"
+    ),
+);
 
 /** Users */
-const UserPage = React.lazy(() => import("./user/UserPage"));
-const CreateUser = React.lazy(() => import("./user/CreateUser"));
-const UpdateUser = React.lazy(() => import("./user/UpdateUser"));
-const ViewUser = React.lazy(() => import("./user/ViewUser"));
+const UserPage = React.lazy(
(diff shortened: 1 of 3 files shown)


frontend/admin/src/js/components/Router.tsx
```diff
@@ -9,92 +10,269 @@ import Layout from "./Layout";
...
+      ),
+  ),
 );
-const DeprecatedUpdatePool = React.lazy(
-  () => import("./pool/deprecated/UpdatePool"),
+const DeprecatedUpdatePool = React.lazy(() =>
+  lazyRetry(
+    () =>
+      import(
+        /* webpackChunkName: "adminUpdatePool" */ "./pool/deprecated/UpdatePool"
```
mnigh, line 176:
```suggestion
        /* webpackChunkName: "adminDeprecatedUpdatePool" */ "./pool/deprecated/UpdatePool"
```

frontend/admin/src/js/components/Router.tsx
```diff
@@ -9,92 +10,269 @@ import Layout from "./Layout";
...
+      ),
+  ),
 );
-const DeprecatedCreatePool = React.lazy(
-  () => import("./pool/deprecated/CreatePool"),
+const DeprecatedCreatePool = React.lazy(() =>
+  lazyRetry(
+    () =>
+      import(
+        /* webpackChunkName: "adminCreatePool" */ "./pool/deprecated/CreatePool"
```
mnigh, line 184:
```suggestion
        /* webpackChunkName: "adminDeprecatedCreatePool" */ "./pool/deprecated/CreatePool"
```

mnigh (changes requested):
Cool! I noticed a linting warning and some naming convention suggestions. Will give it another test after that.

frontend/admin/src/js/components/Router.tsx
esizer: Thanks! 👯

frontend/admin/src/js/components/Router.tsx
esizer: Thanks! 👯

mnigh (commented):
I followed your test steps and they work as expected. The one thing I have noticed was that on requests that I would expect to throw a 404, they instead show the **Sorry, we encountered an error.** message. This may be unrelated and if so, I can open another issue.

esizer:
> I followed your test steps and they work as expected. The one thing I have noticed was that on requests that I would expect to throw a 404, they instead show the **Sorry, we encountered an error.** message. This may be unrelated and if so, I can open another issue.

Looks like a separate issue. I was using their [FAQ documentation for 404](https://reactrouter.com/en/main/start/faq#how-do-i-add-a-no-match-404-route-in-react-router-v6) but looks like that is not necessary with our `errorElement`! 

We can remove the following and it should work:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/b8227c9a847868c569e84e765eb580ff9182a43e/frontend/talentsearch/src/js/components/Router.tsx#L341-L344

mnigh (approved):
👍

Issue #3313, opened 2022-07-12
Screen readers read out abbreviations as words


mnigh:
### Description
Screen readers read out abbreviations as words abbreviations (AS-01, for example on Update Pool form is read out as the word _as_ instead of the letters _A_ _S_). 

### Challenges

- Difficult to control for any values from the database that can be added or modified outside of the codebase (a skill like IT Business Analyst / IT Project Management)
- Cannot add hidden text on something like a form select option

### Questions
Other than having the full version of the acronym (in the case of AS, it would be Administrative Services), do other ways exist to make abbreviations more accessible for screen readers?

Answer: not worried about reading full words out for classifications.

## proposed solution
Split up classifications - add a space between classification letters in an aria-label.

## Acceptance Criteria
- [ ] Classification is read out as "A" "S" instead of "as" by screen reader.

patcon:
This is an interesting problem, as these will appear in a ton of option labels, and it feels odd to put html tags into database entries.

Couple thoughts:
- if we're saying "AS-01" (non-expanded) for sighted people is appropriate, then the issue for non-sighted ppl is better framed as "it should as ay-ess-one", rather than "it should say administrative services 1". If we think non-sighted people need the acronym expanded, then so do sighted people :)
- i understand that `<span aria-label="A.S. 1">AS-01</span>` is a way to force screen readers to pronounce properly. [ref](https://www.queensu.ca/accessibility/across-campus/web-standards-and-accessibility-development-guide/writing-web/acronyms-and)
- if we don't like the idea of storing html tags in database labels (which in theory are editable by admins who don't want to use html), then we can pattern match with regex to create html tags: https://regex101.com/r/eoPnrj/1
  - AS-01 => `<span aria-label="A.S. 1">AS-01</span>`
- we can process labels for this, and any other place we think these might show up.

patcon:
> Cannot add hidden text on something like a form select option

I think this might be mitigated if/when we fully move to react-select, as their options and labels are full HTML

MaddyDaigle:
So, I did some research, and all of these should be tagged as abbreviations` <abbr> ` with a title attribute defining the abbreviated text.  This allows the screen reader user to understand what the abbreviation means, even if it's not read out entirely correctly.

Here is a way you could code it so that it both has the abbreviation title and reads it out correctly.  I just tested this and it seems to work!

[...]

PR #4819, opened 2022-11-18
Fix accordion dates

### LINKED ISSUE #4625: Bug - Date formatting displayed in French incorrectly
spacing looking a little weird on the dates 🤐 

image

_Originally posted by @vd1992 in https://github.com/GCTC-NTGC/gc-digital-talent/issues/4556#issuecomment-1306082943_

petertgiles:
Sorry to pull something out of the ice box.  :laughing: I needed something quick and it was making me crazy to have a bug:prod in there.

And apparently French month abbreviations are not capitalized and can have four characters?

image

image

Closes #4625

### DIFF
--- a/frontend/common/src/components/UserProfile/ExperienceAccordion/individualExperienceAccordions/AwardAccordion.tsx
+++ b/frontend/common/src/components/UserProfile/ExperienceAccordion/individualExperienceAccordions/AwardAccordion.tsx
@@ -8,8 +8,7 @@ import {
   getAwardedTo,
   getAwardedScope,
 } from "../../../../constants/localizedConstants";
-import { getLocale } from "../../../../helpers/localize";
-import { getDateRange } from "../../../../helpers/dateUtils";
+import { getDateRange } from "../../accordionUtils";
 import { AwardExperience } from "../../../../api/generated";
 
 export const AwardContent = ({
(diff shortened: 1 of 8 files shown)


mnigh (commented):
This is so much better than before!

However, I did find a Government of Canada [document](https://www.noslangues-ourlanguages.gc.ca/fr/cles-de-la-redaction/mois.html) (strangely, only available in French though a similar [document](https://www.noslangues-ourlanguages.gc.ca/en/writing-tips-plus/abbreviations-days-and-months.html) is available in English) with a table for abbreviating months in both French and English.

## Screenshot
Screen Shot 2022-11-21 at 12 14 19

All of the French month abbreviations work except for July, which should have an extra `l` before the period, **juill.**, and currently shows **juil.** instead. Sadly, the library we are using actually uses the incorrect abbreviation for fr-CA, but I have opened an [issue](https://github.com/date-fns/date-fns/issues/3281) for it.

Also, the Government of Canada document in English uses periods after some of the months and the [`<abbr>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr) element. I know this is changing the scope, but I'm wondering if we should be using abbreviations at all to avoid having to include markup when we could just use the full months names in any language. Thoughts?

petertgiles:
@Jerryescandon Thoughts on using abbreviations on this component vs using full names?  If we go with abbreviations we'll need to eventually add abbr tags for a11y but, of course, it will be longer.

Jerryescandon:
Abbreviations do clean up this UI a lot, and are mostly secondary info. I would love it if we can work abbreviations properly. But its by no means an important ticket, if you have the time, go for it.

petertgiles:
Thanks @mnigh for the l18n deep-dive on months.  I think the problems are out of scope for this PR but I've added a new issue to the board:
#4860 - For aligning abbreviations

And added a note about month abbreviations to #3313.

mnigh (approved):
👍

Issue #4573, opened 2022-11-03
Add indigenous languages to IAP homepage


tristan-orourke:
Add a new language selector to the IAP homepage for indigenous languages. This will change the language for this page only, not including the header and footer.

This can be implemented by having the new language switcher pass a second locale and set of compiled messages into another IntlProvider component which wraps the content of the page, but not the header and footer. Each new language will get its own file in `frontend/indigenousapprenticeship/src/js/lang`.

[Figma Prototype](https://www.figma.com/proto/tUuPyWzaZWSEwCflpfOLq6/IAP-Homepage---V3.5.2?page-id=0%3A1&node-id=803%3A625&viewport=-5833%2C2071%2C0.5&scaling=min-zoom&starting-point-node-id=117%3A1358)

![IAP Language Picker.png](https://images.zenhubusercontent.com/5d9ded8fd25901000101f3f0/8c10c582-acfc-4aef-9c2e-13ad06e4a40e)

For now, translations into 4 different languages have been provided, more will be provided in the future.

## Acceptance Criteria
- [ ] Language switcher appears at top of page
- [ ] All language between header and footer changes with language switcher
- [ ] Selecting a language in the language picker links you to the page in the correct language code.
  - [ ] Maybe we use a url query parameter for IL code for now, to not interfere with our existing language system.
- [ ] ensure that links on the IAP page use the `<a href="/support" hreflang="en">` format, if the page isn't already in en or fr

## Out of Scope
- Any change to the existing english-french language provider
- Any change to the existing english-french translation workflow (in terms of the scripts we run)
- Any localization of header, footer, or non-iap page
- Language code in URI path (don't interfere with existing language routing)
- Saved sate, local storage
- Generating JSON files is happening in  #4587
- Proofreading translations
- Fixing layout issues due to language differences
- Graphics with text in them changes

tristan-orourke:
Translations:
[English text being translated.](https://github.com/GCTC-NTGC/gc-digital-talent/files/9930320/014-221021-022_001_TG_TR20760.-.IAP.Homepage.Text.English.to_Michif.docx)
[014-221021-022_001_TG_TR20760 - IAP Homepage Text English to_Michif.docx](https://github.com/GCTC-NTGC/gc-digital-talent/files/9930331/014-221021-022_001_TG_TR20760.-.IAP.Homepage.Text.English.to_Michif.1.docx)
[...]

PR #4821, opened 2022-11-18
Feature - 4795  - Create nested language provider

### LINKED ISSUE #4795: Create SecondaryLanguageContainer

esizer:
## 👋 Introduction

This creates the `NestedLanguageProvider` to support secondary languages based on a search param.

## 🕵️ Details

This uses `react-router` to get the search params and `LanguageProvider` so it must be used as a child of both. I believe the best place to use this is wrapping all items in the `<Layout>` component which is how it was done during testing.

## 🔧 Usage

Since we are use user input in search params, this makes use of maps for a more secure approach.

```tsx
import NestedLanguageProvider from "@common/components/context/NestedLanguageProvider";
import * as ojwMessages from "../lang/ojwCompiled.json";
import * as micMessages from "../lang/micCompiled.json";

const messages: Map<string, Messages> = new Map([
  ["ojw", ojwMessages],
  ["mic", micMessages],
]);

const Layout = () => {
  ...
  return (
     <NestedLanguageProvider messages={messages}>
        {...}
     </NestedLanguageProvider>
  );
}
```

## ❓ Question

I had some trouble with the message types so decided to cast the type. I'd prefer a better approach if anyone could provide some suggestions.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/221af6de0dc20418118461a34d84c0d7c2c63dad/frontend/common/src/components/context/NestedLanguageProvider.tsx#L46-L50

## 🧪 Testing

Testing is difficult without additional languages yet. How I tested locally was the following:

1. Duplicate the `frCompiled.json` file in a project
2. Pick out strings where you know they exist for checking
3. Replace those strings with some phrase ("testing")
4. Add the `<NestedLanguageProvider>` to a project (I did IAP since that is where it will be used first)
5. Pass the new compiled file to the messages prop as it is done in the `<ContextContainer>` in main
6. Navigate to the page and confirm the normal functionality remains (EN, FR)
7. Add the locale as a search param to the URL (http://...?locale=xxx)
8. Confirm the strings you replaced show up and the rest fallback to the expected language (EN, FR)

### Sample Compiled File for Western Ojibway (ojw)

 This will replace the "About the Program" title on the IAP.

```json
{
  "CqLV19": [
    {
      "type": 0,
      "value": "Test OJW"
    }
  ]
}
```

## 🤖 Robot Stuff

Resolves #4795

### DIFF
--- a/frontend/common/src/components/context/NestedLanguageProvider.tsx
+++ b/frontend/common/src/components/context/NestedLanguageProvider.tsx
@@ -1,0 +1,60 @@
+import React from "react";
+import { IntlProvider, useIntl } from "react-intl";
+import { useSearchParams } from "react-router-dom";
+
+import defaultRichTextElements from "../../helpers/format";
+import { LanguageProviderProps } from "./LanguageProvider";
+
+/**
+ * Indigenous Language codes
+ *
+ * crg - Michif
+ * crk - Plains Cree
+ * ojw - Western Ojibway
+ * mic - Mikmaq
+ */
+type SecondaryLocale = "crg" | "crk" | "ojw" | "mic" | undefined;
+
+/**
+ * Nested Language Provider
+ *
+ * Providers a secondary language within the
+ * existing Language Provider.
+ *
+ * Note: This must be used within a LanguageProvider
+ *
+ * @param param0
+ */
+const NestedLanguageProvider = ({
+  messages,
+  children,
+}: LanguageProviderProps) => {
+  const intl = useIntl();
+  const [searchParams] = useSearchParams();
+  const locale = searchParams.get("locale") as SecondaryLocale;
+
+  // Return just children if there is no
+  // locale in the search params
+  if (!locale) {
+    return children as JSX.Element;
+  }
+
+  // Merge the messages
+  const nestedMessages = {
+    ...intl.messages,
+    ...messages,
+  };
+
+  return (
+    <IntlProvider
+      locale={locale}
+      key={locale}
+      messages={nestedMessages}
+      defaultRichTextElements={defaultRichTextElements}
+    >
+      {children}
+    </IntlProvider>
+  );
+};
+
+export default NestedLanguageProvider;
--- a/frontend/common/src/components/context/NestedLanguageProvider.tsx
+++ b/frontend/common/src/components/context/NestedLanguageProvider.tsx
@@ -3,7 +3,7 @@ import { IntlProvider, useIntl } from "react-intl";
 import { useSearchParams } from "react-router-dom";
 
 import defaultRichTextElements from "../../helpers/format";
-import { LanguageProviderProps } from "./LanguageProvider";
+import { Messages } from "./LanguageProvider";
 
 /**
  * Indigenous Language codes
@@ -13,7 +13,12 @@ import { LanguageProviderProps } from "./LanguageProvider";
  * ojw - Western Ojibway
  * mic - Mikmaq
  */
-type SecondaryLocale = "crg" | "crk" | "ojw" | "mic" | undefined;
+export type SecondaryLocale = "crg" | "crk" | "ojw" | "mic";
+
+interface NestedLanguageProvider {
+  messages: Map<string, Messages>;
+  children: React.ReactNode;
+}
 
 /**
  * Nested Language Provider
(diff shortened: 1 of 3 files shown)


petertgiles:
Not in scope for this issue but we need to watch the image since it uses the locale string:
image

frontend/common/src/components/context/NestedLanguageProvider.tsx
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * Note: This must be used within a LanguageProvider
+ */
+const NestedLanguageProvider = ({
+  messages,
+  children,
+}: NestedLanguageProvider) => {
+  const { messages: fallbackMessages } = useIntl();
+  const [searchParams] = useSearchParams();
+  const locale = searchParams.get("locale") as SecondaryLocale;
```
petertgiles, line 37: I don't love this cast.  It doesn't actually check if the provided string matches a locale, right?  It might be better to check and then force it to null if it doesn't fit.

petertgiles (changes requested):
Wow, this is really nice and clean and simple.  I'm not sure how to improve the message cast though.  To be honest it doesn't bother me and I would be fine leaving it in.  I don't love the locale cast though.

frontend/common/src/components/context/NestedLanguageProvider.tsx
petertgiles: Oh, is this what you mean about the maps being safer?  In that case maybe get rid of the SecondaryLocale type since it's not really providing any value?  Or do you have plans for it down the road.  :smile:

mnigh:
> Not in scope for this issue but we need to watch the image since it uses the locale string: image

See: https://github.com/GCTC-NTGC/gc-digital-talent/issues/4573#issuecomment-1302417388

frontend/common/src/components/context/NestedLanguageProvider.tsx
esizer: Yeah, so we can check to make sure it is a locale but with the use of maps, `messages` would return undefined and exit early if the key does not exists.

The plan is the use the type for the map to constrain that. I should have included that in the test instructions but I was more focused on getting a MVP up for testing 😅 

Screenshot 2022-11-21 at 4 55 15 PM

esizer: As an example of the early exit:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/221af6de0dc20418118461a34d84c0d7c2c63dad/frontend/common/src/components/context/NestedLanguageProvider.tsx#L37-L44

If we pass something like "not-a-locale" in as a search param, `locale` would be "not-a-locale" and `localeMessages` would be `undefined` so line 42 should resolve to true and just render the children without the provider so nothing changes.

Though, we could write something like `isSecondaryLocale` and add that to the checks as well.

```tsx
export type SecondaryLocale = "crg" | "crk" | "ojw" | "mic";
const secondaryLocales = ["crg", "crk", "ojw", "mic"];

export function isSecondaryLocale(
  locale?: string | null,
): locale is SecondaryLocale {
  return locale ? secondaryLocales.includes(locale) : false;
}
```

petertgiles: Yeah, I like the type guard idea or just leave locale as a string.  It makes me uncomfortable to do an explicit cast when we know it's not actually constrained to that type.

esizer: Alright, updated with [add type guard instead of casting](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4821/commits/04e1b47b07c77fa2a7c9f7b359e8578dabb6e3d9).

I'm not a fan of maintaining a type and an array but I honestly couldn't figure out a better way to type guard a type alias like this 🤷‍♀️

petertgiles (approved):

frontend/common/src/components/context/NestedLanguageProvider.tsx
esizer: Okay, removed all the extra stuff in [remove secondary type and type guard](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4821/commits/637a28d9ec352df4b9cb9e7ae18ae80623cfc7da).

If this looks good to you, I think we can merge.

petertgiles: This looks great!  Thanks for being patient with me.  :grin:

PR #4837, opened 2022-11-18
bug/4619-hidden-tag-whitespace firefox

### LINKED ISSUE #4619: Bug - Firefox strips leading spaces from strings
## Description
Text within the `hidden` tag cannot begin with a space and text adjacent to the closing `hidden` tag cannot begin with a space as it is stripped away in Firefox and alters the copy for screen readers.

## Screenshot
Screen Shot 2022-11-07 at 09 58 42

## Acceptance Criteria
- [ ] update strings so that hidden strings are read properly by screen readers
- [ ] ensure French translations do not get disconnected and also are update appropriately

vd1992:
closes #4619 
adds something to preserve whitespace in elements that contain something with `<hidden>`

testing will be a bit tedious
- check chromatic
- checks the link text and accessiblity text across multiple places
- repeat for another browser (recommend Firefox because that is the source of the bug, and quickly another browser to ensure no other bugs creeped in)

some places with `<hidden>` text to check, check files tab if you want to see all the places
`/en/admin/pools` see "View Candidates" and "Edit"
`/en/admin/dashboard` see "View Request"
`/en/browse/pools` see "Apply Now"
`/en/users/:id/profile/employment-equity/edit` see buttons

### DIFF
--- a/frontend/admin/src/js/components/Table/TableEditButton.tsx
+++ b/frontend/admin/src/js/components/Table/TableEditButton.tsx
@@ -26,6 +26,7 @@ function TableEditButton({
       mode="inline"
       color="black"
       data-h2-padding="base(0)"
+      style={{ whiteSpace: "pre" }}
     >
       {intl.formatMessage(
         {
--- a/frontend/admin/src/js/components/Table/TableViewItemButton.tsx
+++ b/frontend/admin/src/js/components/Table/TableViewItemButton.tsx
@@ -19,7 +19,13 @@ function TableViewItemButton({
 }: TableViewItemButtonProps): React.ReactElement {
   const intl = useIntl();
   return (
-    <Link href={viewUrl} type="button" mode="inline" data-h2-padding="base(0)">
+    <Link
+      href={viewUrl}
+      type="button"
+      mode="inline"
+      data-h2-padding="base(0)"
+      style={{ whiteSpace: "pre" }}
+    >
       {intl.formatMessage(
         {
           defaultMessage: "View {name} <hidden>{hiddenLabel}</hidden>",
--- a/frontend/admin/src/js/components/pool/PoolTable.tsx
+++ b/frontend/admin/src/js/components/pool/PoolTable.tsx
@@ -33,6 +33,7 @@ function poolCandidatesLinkAccessor(
       mode="inline"
       color="black"
       data-h2-padding="base(0)"
+      style={{ whiteSpace: "pre" }}
     >
       {intl.formatMessage(
         {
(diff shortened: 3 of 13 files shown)


vd1992:
Cutting down on visual messiness in progress

petertgiles:
Thanks for picking this up!  So the approach is sound and definitely works.  :partying_face: 
image
Two questions about the specifics ( and I don't know the right answers ) :
1) Do we want to have this styling done at the level of the individual components or is it worth hiking up to app-level styling?
2) Can this styling be done in hydrogen or does it have to be in CSS?

mnigh:
> 2. Can this styling be done in hydrogen or does it have to be in CSS?

No, [white-space](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space) is not yet a part of hydrogen. @substrae, will it be anytime soon?

https://github.com/hydrogen-css/hydrogen/blob/6acea2adc54dfc312aa151b2d046d7e994688603/tests/test-control/unfinished.js#L319

substrae:
`white-space` should work as of `beta-30` 🤔

esizer (commented):
Would it work to just add this to the `<hidden>` tag rather than individually on each item where we use it?

https://github.com/GCTC-NTGC/gc-digital-talent/blob/5a64c609c99505e6530add72402d42af48e2c729/frontend/common/src/helpers/format.tsx#L18-L24

vd1992:
> Would it work to just add this to the `<hidden>` tag rather than individually on each item where we use it?
> 

oh yeah, thanks
for some reason I thought `<hidden>` was built in HTML or an import and so immutable 
https://github.com/GCTC-NTGC/gc-digital-talent/pull/4837/commits/b397f07f9ba86e20459256b4ff4c2775f990eb4b

petertgiles (approved):
That's a much lighter diff!  :smile: Looks great.

frontend/common/src/helpers/format.tsx
```diff
@@ -20,7 +20,9 @@ export const strong = (text: React.ReactNode): React.ReactNode => (
  * @param text text to wrap.
  */
 export const hidden = (text: React.ReactNode): React.ReactNode => (
-  <span data-h2-visibility="base(invisible)">{text}</span>
+  <span data-h2-visibility="base(invisible)" style={{ whiteSpace: "pre" }}>
```
mnigh, line 23: Replace style attribute with hydrogen attribute
```suggestion
  <span data-h2-visibility="base(invisible)" data-h2-white-space="base(pre)">
```

Accordion to @substrae, this should work. Does it?

vd1992: no it doesn't
image

mnigh:
I don't think this PR solved https://github.com/GCTC-NTGC/gc-digital-talent/issues/4619 😢

## Screenshot
Screen Shot 2022-11-21 at 17 01 26

petertgiles:
> I don't think this PR solved #4619 cry

Oh no!  So the text fragments are in order (with proper spacing) in the accessibility pane but the MacOS screen reader changes to the order?

PR #4858, opened 2022-11-21
Feature - 4828 - Hide skills details profile

### LINKED ISSUE #4828: HIDE skills in detail on profile
## Description

- Users are being overwhelmed by the amount of skill justifications they need to fill out when they claim it through their Profile.
- Because of how impactful this is, this is being considered as a very high emergency fix
- To reduce this we are considering several solutions to alleviate this problem.
  1. Remove steps 3 and 4 (skills in detail, additional information) from the experience form (DURING PROFILE ONLY)
  2. Add a banner to instruct optimal number of skills
  3. Convert justifications to be optional in profile (needs design)
  4. Hybrid make the first 3 skills require a justification, and all subsequent skills optional (needs design)

### CONCERNS

Concerns for option 1: 
- Skills selected on profile would still show up as required during applications, we would need a fix for that too.
- (Additional concern) Non-poster skills on applications, should we hide them all together?
Option 2:
- Behavioural concerns (is this enough to deter the users?)
Option 3:
- Current design doesn't not convey OPTIONAL interactions really well, even if we just change the label it still might feel required, we would need to implement a new design to better convey that additional justifications are secondary to the interaction.
Option 4:
- LOTS OF DESIGN CONSIDERATIONS

## Screenshot: Prototype
image

## Acceptance Criteria (going with option 1)
- [ ] When editing your profile, section 3 and 4 are removed from the Experience page entirely
- [ ] A user can still edit the set of skills associated with an experience, and those changes are saved.
- [ ] If a user already has skills with associated text saved to this experience, that text will not be lost, though it will not be visible. In other words, it can still be retrieved from the API, and would still appear during an application.
- [ ] If an experience already has additional text context, that will not be lost, though it not be visible.

esizer:
## 👋 Introduction

This hides steps 3 and 4 on the experience form when editing a user profile.

### Profile

1. Navigate to profile and add an experience
2. Confirm that Skills in Detail and Additional Information sections do not appear and you can still save the form

### Application

1. Start an application then add/edit an experience from the application review page
2. Confirm that Skills in Detail and Additional Information sections appear and you can still save the form

## 🤖 Robot stuff

Resolves #4828

### DIFF
--- a/frontend/talentsearch/src/js/components/experienceForm/ExperienceForm.test.tsx
+++ b/frontend/talentsearch/src/js/components/experienceForm/ExperienceForm.test.tsx
@@ -242,10 +242,6 @@ describe("ExperienceForm", () => {
     expect(
       screen.getByRole("heading", { name: /skills displayed/i }),
     ).toBeInTheDocument();
-
-    expect(
-      screen.getByRole("heading", { name: /skills in detail/i }),
-    ).toBeInTheDocument();
   });
 
   it("should render additional information", async () => {
@@ -260,8 +256,6 @@ describe("ExperienceForm", () => {
         skills: mockSkills,
       });
     });
-
-    expect(screen.getByRole("textbox", { name: /additional information/i }));
   });
 
   it("should not submit award with empty fields", async () => {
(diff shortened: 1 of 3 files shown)


JamesHuf (approved):
LGTM! 🚀

PR #4864, opened 2022-11-21
4824 Add stream dropdown to search page

### LINKED ISSUE #4824: Add stream dropdown to Search page
## Description

We want to reduce the amount of pool options that show up at the bottom of the Search page.

### To do

Add a dropdown to select Stream right after selecting group and level
This reduces the pools that show up, at the bottom, but we do not need to record this in the summary afterwards.
It should be the same list as the one in the Job Poster Edit page

### Search page mockup

image

### Job Poster list

image

## Acceptance Criteria
- [ ] Add a dropdown for streams next to the dropdown for Classifications
- [ ] Selecting a stream should restrict the pools used for the filters, in exactly the same way selecting Classifications already does.

### LINKED ISSUE #4863: Initial count on the Search page includes ALL users, not just those in pools
When the search page initially loads (and as long as a classification is not selected), the countCandidates query does not specify any pools. This means that the count may include users who are not in any pools at all. I believe the intended functionality is to only count all users who are in at least one pool.

tristan-orourke:
Resolves #4824.
Resolves #4863

The primary change is to add a Stream dropdown to the search page which, like the classification dropdown, affects which pools are included in the actual filter.

I noticed a few other things I fixed while working on this:
- The required form fields will now validate themselves onChange after they're dirty, meaning if you touch a required field and then unset it, it will show an error right away. Previously it was showing the error only when you clicked Request Candidates.
- The first countCandidates query waits until pools are loaded, and explicitly queries using all known pools instead of an empty array. This resolves #4863.
- I was able to remove a few `as` type casts, making our typing more reliable.


brindasasi:
🧪 Testcases 

1.  Stream dropdown is added as required field under classification and populated with correct values
2.  Reset the values on the required fields and verify on spot errors are thrown 
3.  When Classification & stream is selected the page should return the count of candidates who are all "qualified available"  in the specified pool.
4. The search page count should match candidates on Request Summary page 
5. The search page count should match candidates on Pool Candidates page 

For ( #4863) 
6. On Initial load when classification & stream are null, the total estimated candidates count  should be the total no of all pool candidates

brindasasi (changes requested):
Works nicely!!! Numbers are adding up. French translation for "Information and Data Functions" is missing as below in the streams drop down. 
Screen Shot 2022-11-22 at 2 14 46 PM

tristan-orourke:
@brindasasi Odd, the translation is there for me! 
image

I've just fixed a merge conflict. Maybe rebuild and check again?

brindasasi:
For ( https://github.com/GCTC-NTGC/gc-digital-talent/issues/4863) I get the discrepancy in total estimated candidates as below : 
Screen Shot 2022-11-22 at 2 38 18 PM
Screen Shot 2022-11-22 at 2 35 05 PM
what am I missing here ? 🤔 Are we counting candidates with more than 1 pool as well ?

brindasasi:
> @brindasasi Odd, the translation is there for me! image
> 
> I've just fixed a merge conflict. Maybe rebuild and check again?

yeah .. rebuild fixed it. Thanks!

tristan-orourke:
Ah, I found the number discrepancy. The countApplicants query also filters for users who are ACTIVELY LOOKING or OPEN TO OPPORTUNITIES, and when filtering by pool it also ensure the Pool Candidate is not expired. Try this sql command - it matched the search page, for me:

[select](https://www.postgresql.org/docs/12/static/sql-select.html) [count](https://www.postgresql.org/docs/12/static/functions-aggregate.html)(*) from [users](http://localhost:8080/?pgsql=postgres&username=postgres&db=gctalent&ns=public&table=users) u, [pool_candidates](http://localhost:8080/?pgsql=postgres&username=postgres&db=gctalent&ns=public&table=pool_candidates) pc
WHERE pc.user_id = u.id
[AND](https://www.postgresql.org/docs/12/static/functions-logical.html) u.job_looking_status [IN](https://www.postgresql.org/docs/12/static/functions-subquery.html) ('ACTIVELY_LOOKING','OPEN_TO_OPPORTUNITIES')
[AND](https://www.postgresql.org/docs/12/static/functions-logical.html) pc.pool_candidate_status [IN](https://www.postgresql.org/docs/12/static/functions-subquery.html) ('QUALIFIED_AVAILABLE','PLACED_CASUAL')
[AND](https://www.postgresql.org/docs/12/static/functions-logical.html) pc.expiry_date >= [now](https://www.postgresql.org/docs/12/static/functions-datetime.html)()

brindasasi (approved):
PERFECT!

PR #4899, opened 2022-11-23
Bug - Fix extend pool closing date

### LINKED ISSUE #4775: Cannot extend a Pool Closing Date
Steps to reproduce:
1. Log in to admin dashboard
2. Go to Pools table, and find an Expired pool. (If none exist, close an existing pool early, or create a new pool to do so.)
3. On the Pool Edit page, scroll to the bottom and click "Extend the date".
4. When the modal opens, select a future date and click "Extend closing date"
5. The modal closes but the pool doesn't update. If you open the browser inspector, you'll see no network request is made, and a console warning is shown saying "Form submission canceled because the form is not connected".

yonikid15:
Resolves #4775 

## Notes
- Fixes the broken form submission on extend pool closing date on Edit Pool page. 
- The reason for it not working was the Dialog closing (removed from the DOM) before the form could submit. I followed the suggestions from Dialog radix docs to fix [asynchronous form submissions](https://www.radix-ui.com/docs/primitives/components/dialog#close-after-asynchronous-form-submission).
- Also, adds a fix for `endDate` defaultValue not showing up on the `date` input. Removing the time from the date was the solution.

## Testing
- Go to admin page  and go to Pools table.
- Edit a pool with status Expired/Published. If none exist close a pool early, or create a new one.
- Scroll to bottom and try extending the pool date.
- Ensure default value shows up on date input. 
- Ensure form is submitted before the dialog closes.
- Make sure correct date shows up after submission.

### DIFF
--- a/frontend/admin/src/js/components/pool/EditPool/EditPool.test.tsx
+++ b/frontend/admin/src/js/components/pool/EditPool/EditPool.test.tsx
@@ -223,7 +223,7 @@ describe("Edit Pool tests", () => {
   });
 
   it("should have an extend button that pops a modal and emits an event when the status is published", async () => {
-    const handleEvent = jest.fn();
+    const handleEvent = jest.fn(() => Promise.resolve());
     const props = {
       ...EditPoolStory.args,
       ...PublishedAdvertisement.args,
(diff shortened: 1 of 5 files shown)


esizer (approved):
Nice! 😁 

Out of scope for the issue but we may want to look at changing the button state to indicate to the user something is happening (disabled + "Saving...")

yonikid15:
> Nice! 😁
> 
> Out of scope for the issue but we may want to look at changing the button state to indicate to the user something is happening (disabled + "Saving...")

Good Idea. I can make an issue if you haven't already @esizer.

esizer:
> Good Idea. I can make an issue if you haven't already @esizer.

Added 😄 

https://github.com/GCTC-NTGC/gc-digital-talent/issues/4905

PR #4903, opened 2022-11-23
Bug/4901 no search results

### LINKED ISSUE #4901: Null state not showing on Search page
The search used to show some text if there were 0 candidates who matched your search, including "We can still help! If there are no matching candidates Get in Touch!". Since changing from showing only a single Digital Careers result to showing a different result card for every pool, we've lost that null state.

Cause of the problem:
The CandidateResults component shows the null state text if candidateCount == 0. But now, we render a list of CandidateResults, and only for results which are non zero!

Line 239 of SearchContainer.tsx:
```tsx
 {poolCandidateResults?.map(({ pool, candidateCount }) => (
                <CandidateResults
                  key={pool.id}
                  candidateCount={candidateCount}
                  pool={pool}
                  handleSubmit={tryHandleSubmit}
                />
              ))}
```

vd1992:
closes #4901 
updates to have a state for no search results

- recycled from the deprecated search page
- it definitely is a bit verbose, I was struggling to get a nested ternary rendering going but all I saw were red squigglies so I went simple + wordy

testing:
head to `localhost:8000/en/search` and play with the search and make sure it all looks right


vd1992:
experimenting with commit https://github.com/GCTC-NTGC/gc-digital-talent/pull/4903/commits/5926850a918196601c031dbc69077b23672fa025
see https://github.com/GCTC-NTGC/gc-digital-talent/pull/4907#issuecomment-1326859919

if it doesn't look right, just say it and I will reverse it

vd1992:
successful 3/3 times now

frontend/talentsearch/src/js/components/search/SearchContainer.tsx
```diff
@@ -234,20 +246,52 @@ export const SearchContainer: React.FC<SearchContainerProps> = ({
...
-                  pool={pool}
-                  handleSubmit={tryHandleSubmit}
-                />
-              ))}
-            </div>
-          ) : (
-            <Spinner />
-          )}
+          {!updatePending &&
+            (poolCandidateResults && poolCandidateResults.length > 0 ? (
```
esizer, line 250: This is just my personal opinion and I'm more than willing to relent on this because it's pretty nit-picky but, I'm not a huge fan of this condition and find it kinda awkward to read. 

Could we maybe extract this out to simplify it? @petertgiles @vd1992 

## Definition

```tsx
interface ResultsDisplay {
  pending: boolean;
  results?: Array<PoolCandidate>;
}

const ResultsDisplay = ({pending, results}: ResultsDisplayProps) => {
  if(pending) {
    return <Spinner />;
  }

  return results && results.length ? (
    // CandidateResults
  ) : (
    // We can still help message
  )
}
```

## Usage

```tsx
<ResultsDisplay pending={updatePending} results={poolCandidateResults} />
```

esizer (commented):
Everything works great here, just leaving a small comment about implementation. 

Feel free to ignore/push back if you disagree and I can approve 😄

frontend/talentsearch/src/js/components/search/SearchContainer.tsx
petertgiles: Ah, yeah, I like that.  I tried to improve it with a ternary but this is better imo.  :mage:

vd1992: thanks for the implementation details
done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/4903/commits/79728bc98c4d8c4af94cf72c2e5371cb3eee8b4b

esizer (approved):
🆒 !

PR #4900, opened 2022-11-23
change pool seeding + cypress test tweak

vd1992:
two things I noticed

first: search workflow would fail sometimes due to chance, for this I constrained the seeded digital careers pool to be more predictable and adjusted the search workflow to look for it

second: stream requirement was merged in with #4864 but this was not reflected in the workflow, and it threw no errors probably due to this line below, I added stream selection into the workflow

image

closes #4902

PR #4907, opened 2022-11-24
Bug - 3824 - Table print warn no selection

### LINKED ISSUE #3824: Print link on users table shows "hand" icon even though link is disabled
CSV link behaves correctly though.

In UAT:

Acceptance Criteria:
- [ ] Remove "disabled" attribute from link (in all scenarios)
- [ ] If no rows are selected and the link is click, put up a toast notification (see text below) and do nothing else.
- [ ] Update the link to say "Download Profiles" "Télécharger les profils" instead of "Print"

English:
"No users selected"
French:
"Aucun utilisateur sélectionné"

esizer:
## 👋 Introduction

This updates the print button on the api table to show a toast when a user attempts to print with no rows selected. Also, it updates the font weight to match the CSV link.

## 📋 TO DO

- [x] Send off new text for translation
- [x] Add translations

## 🧪 Testing

2. Navigate to the users table
3. Click "print" in the footer without selecting rows
4. Confirm toast appears with error message
5. Select from rows and attempt to print again
6. Confirm print window appears

## 🤖 Robot Stuff

Resolves #3824

### DIFF
--- a/frontend/admin/src/js/components/apiManagedTable/TableFooter.tsx
+++ b/frontend/admin/src/js/components/apiManagedTable/TableFooter.tsx
@@ -45,10 +45,10 @@ function TableFooter({
 
   const handlePrint = () => {
     if (disableActions) {
-      toast.warning(
+      toast.error(
         intl.formatMessage({
-          defaultMessage: "No rows selected to print.",
-          id: "H6oeBv",
+          defaultMessage: "Print failed: No rows selected",
+          id: "x/sZcv",
           description:
             "Alert message displayed when a user attempts to print without selecting items first",
         }),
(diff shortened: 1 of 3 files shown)


JamesHuf (changes requested):
It looks like the copy on the link is supposed to be updated to "Download Profiles"/"Télécharger les profils".
Also the copy on the toast is somewhat different from what is provided in the issue.

esizer:
> It looks like the copy on the link is supposed to be updated to "Download Profiles"/"Télécharger les profils".
> Also the copy on the toast is somewhat different from what is provided in the issue.

Good catch! Updated in [update copy](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4907/commits/080e149e6670b0e1a6366bcfe99476b89cf07b39)

In regards to the toast copy, I spoke with @Jerryescandon this morning about the change so we could add a bit more context to the actual error.

JamesHuf (approved):
Works for me! 🚀

vd1992:
Saw the Cypress failed for the same reason as one of my runs earlier
Seems like my merge #4900 earlier introduced an edge case bug 😭 

where if there is only one pool present, the stream selection changes nothing and doesn't send a graphQL request 
so it is the marked bit below, should that line just get deleted or what?

image

PR #4912, opened 2022-11-24
Bug - 4820 - Throw 404 on pool advertisement page error

### LINKED ISSUE #4820: GraphQL error shown instead of `ThrowNotFound` component
## Description
When navigating to `/browse/pools/{poolId}` where `poolId` is not an existing value in the database, a `Oh no...[GraphQL] Internal server error` is displayed rather than a nice looking error.

## Screenshot
Screen Shot 2022-11-18 at 12 49 42

## Acceptance Criteria
- [ ] render `ThrowNotFound` component if `poolId` not found

esizer:
## 👋 Introduction

This returns 404 if there is an error loading the pool on the `/browse/pools/{poolId}` page.

> **Note**
> This may be too generic since it happens on any error, thoughts?

> **Note**
> This may be an issue on other pages also. We should maybe sweep through the pages and make sure we throw errors when necessary.

## 🧪 Testing

2. Navigtate to `/browse/pools/anything-but-a-vaild-id`
3. Confirm you see the 404 page

## 🤖 Robot Stuff

Resolves #4820

### DIFF
--- a/frontend/common/src/helpers/util.ts
+++ b/frontend/common/src/helpers/util.ts
@@ -119,20 +119,3 @@ export const emptyToUndefined = (s: InputMaybe<string>): string | undefined =>
 export function uniqueItems<T>(arr: T[]): T[] {
   return Array.from(new Set(arr));
 }
-
-/**
- * Determines if a given string is a UUID
- *
- * @param id  string  ID to test
- * @returns   boolean If the string is a UUID
- */
-export function isUUID(id?: string): boolean {
-  // IDs from router can be undefined
-  if (!id) return false;
-
-  // Regular expression to check if string is a valid UUID
-  const regexExp =
-    /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/gi;
-
-  return regexExp.test(id);
-}
--- a/frontend/common/src/helpers/util.ts
+++ b/frontend/common/src/helpers/util.ts
@@ -119,3 +119,20 @@ export const emptyToUndefined = (s: InputMaybe<string>): string | undefined =>
 export function uniqueItems<T>(arr: T[]): T[] {
   return Array.from(new Set(arr));
 }
+
+/**
+ * Determines if a given string is a UUID
+ *
+ * @param id  string  ID to test
+ * @returns   boolean If the string is a UUID
+ */
+export function isUUID(id?: string): boolean {
+  // IDs from router can be undefined
+  if (!id) return false;
+
+  // Regular expression to check if string is a valid UUID
+  const regexExp =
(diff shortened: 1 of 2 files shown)


petertgiles:
> This may be too generic since it happens on any error, thoughts?

Yeah, I agree with this.  The original issue was just about pools and the estimate of 1 makes sense for that but if you're up for it, I think this is a problem across the whole app and it would be worth coming up with a pattern that can be reapplied everywhere.  If you're not up for it right now, this PR does solve the problem just for pools but I think a new issue should go on the board to come up with a long-term solution.

Off the top of my head, I wonder about plugging in at the error that gets passed to `Pending`.
image
Maybe instead of sending the whole `CombinedError` into it, use the context of the page and maybe a helper function to send the best error message in there?

esizer:
> > This may be too generic since it happens on any error, thoughts?
> 
> Yeah, I agree with this.  The original issue was just about pools and the estimate of 1 makes sense for that but if you're up for it, I think this is a problem across the whole app and it would be worth coming up with a pattern that can be reapplied everywhere.  If you're not up for it right now, this PR does solve the problem just for pools but I think a new issue should go on the board to come up with a long-term solution.
> 
> Off the top of my head, I wonder about plugging in at the error that gets passed to `Pending`.
> image
> Maybe instead of sending the whole `CombinedError` into it, use the context of the page and maybe a helper function to send the best error message in there?

Agreed, I'll look into this and see if I can come up with something. If I run into trouble I'll create a new issue 🤞

esizer:
@petertgiles So, the error was coming from the fact that we were not passing a valid UUID to the API, causing an internal server error.

I added a helper to check the UUID is valid then prevented the query from running and throw a 404 (think this is an appropriate error code?) if it fails the UUID check.

Does this make sense to you? If so, should this be implemented every where we use UUIDs or is that a separate ticket?

[check ID before running query, 404 if invalid](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4912/commits/58a52140537f82aa19f471a15cc41e8788f19001)

petertgiles:
> Does this make sense to you? If so, should this be implemented every where we use UUIDs or is that a separate ticket?

Yeah, I noticed that, too.  I'm not sure this is worth checking on the frontend though - this seems like a backend concern.  The frontend shouldn't care if the backend is auto-numbering by UUID or integers, I think.  For a user, if a pool doesn't exist it probably doesn't matter to them if the ID could be used in theory.

esizer:
> Yeah, I noticed that, too. I'm not sure this is worth checking on the frontend though - this seems like a backend concern. The frontend shouldn't care if the backend is auto-numbering by UUID or integers, I think. For a user, if a pool doesn't exist it probably doesn't matter to them if the ID could be used in theory.

💯 Agree. I reverted the last commit in [revert uuid check on frontend](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4912/commits/49eaf645faed5bf4c746d14e47b38285267933ad) and created an issue for this.

Can you take a look at https://github.com/GCTC-NTGC/gc-digital-talent/issues/4921 and provide any additional info that could help?

petertgiles (approved):
This works for now.  :+1: It's not very specific but it is clean and friendly.

PR #4924, opened 2022-11-25
Bug - 4905 - Indicate saving state, extend dialog

### LINKED ISSUE #4905: Extend pool dialog does not indicate "saving" state to user
## 👋 Introduction

On the admin edit pool page, the dialog to extend the pool has no indication that there is work happening in the background.

## 🛠️ Proposed Solution

We can use the `useEditPoolContext` hook to determine when the form `isSubmitting` to disable the button and update the text to indicate to users when the form is submitting.

## ✅ Acceptance Criteria

- [ ] Extend dialog form has some indication to users while it works on the submission

esizer:
## 👋 Introduction

This uses `formState.isSubmitting` to disable the save button and update its text to indicate to users the form is in the process of saving. It also removes the `useMemo` from the footer which cause preventing the button from reliably updating.

## 🧪 Testing

2. Navigate to a published pool edit page
3. Open the Extend date dialog and submit the form
4. Ensure the button is disabled and reads "Saving..."
5. Ensure it returns to normal state and dialog closes

## 🤖 Robot Stuff

Resolves #4905

### DIFF
--- a/frontend/admin/src/js/components/pool/EditPool/ExtendDialog.tsx
+++ b/frontend/admin/src/js/components/pool/EditPool/ExtendDialog.tsx
@@ -5,7 +5,7 @@ import { PoolAdvertisement } from "@common/api/generated";
 import { Button } from "@common/components";
 import { FormProvider, useForm } from "react-hook-form";
 import { Input } from "@common/components/form";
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
Love it!

Issue #4671, opened 2022-11-09
Term duration (permanent) vs. Indeterminate duration (permanent)


mnigh:
Seems odd to use "Term duration (permanent)" when the user selects "Indeterminate duration (permanent)" on the search form. Do you think we should create an issue to sync up terminology from `talentsearch` and `admin`?

_Originally posted by @esizer in https://github.com/GCTC-NTGC/gc-digital-talent/pull/4669#pullrequestreview-1174082561_

mnigh:
@esizer is this issue still relevant?

esizer:
> @esizer is this issue still relevant?

You know what, I'm not even sure. I went back and looked at my original comment and I can't even remember where the terms are being used. I should have linked to the line of code or added a screenshot... 😢

petertgiles:
I can't find any discrepancies.
![image](https://user-images.githubusercontent.com/8978655/234674583-882222f3-bb2f-4771-8b75-d9d557dba8df.png)
![image](https://user-images.githubusercontent.com/8978655/234675627-a77dd6f8-9d53-4e5f-aa76-089946304053.png)
![image](https://user-images.githubusercontent.com/8978655/234675669-a17e804c-06a4-4dac-942d-8f0f0e49eff9.png)

mnigh:
@esizer is it [closing time](https://www.youtube.com/watch?v=xGytDsqkQY8) for this issue then?

esizer:
> @esizer is it [closing time](https://www.youtube.com/watch?v=xGytDsqkQY8) for this issue then?

Go for it! 👍

mnigh:
Closing as can't reproduce.

PR #4956, opened 2022-11-28
Debt/4667 would accept temporary successor

### LINKED ISSUE #4667: `wouldAcceptTemporary` in ApplicantFilter is a boolean but acts like an enum
## Description

wouldAcceptTemporary, and the way it works, is currently very confusing. It should be updated to be more intuitively understandable to the devs.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/9edc7906da0dd79abca399b5e6f89789b3f866ed/api/app/Models/ApplicantFilter.php#L18

https://github.com/GCTC-NTGC/gc-digital-talent/blob/9edc7906da0dd79abca399b5e6f89789b3f866ed/api/graphql/schema.graphql#L517

https://github.com/GCTC-NTGC/gc-digital-talent/blob/9edc7906da0dd79abca399b5e6f89789b3f866ed/frontend/talentsearch/src/js/components/search/SearchForm.tsx#L172-L173

Relevant slack discussion: https://talent-cloud.slack.com/archives/C0259G6KXJ6/p1667335340182079

### Proposal A:
Users specify a list of duration types they are interested in. Managers specify the list of duration types they can offer. Filtering works by simple intersection.
```graphql
enum PositionDuration {
  TEMPORARY
  PERMANENT
}
type User {
  ...
  positionDurations: PositionDuration[] # what durations am I interested in
}
type ApplicantFilter {
  ..
  positionDurations: PositionDuration[] # what durations the manager can offer
}
```
Pros: intuitive, flexible, expandable
Cons: Does not build in our domain expectation that every user should be interested in Permanent positions. This could be done at the UI level, if we want.

### Proposal B:
Enums should build in domain expectations about users always being open to Permanent positions (which is what wouldAcceptTemporary does, but very confusingly). To be properly explicit, we need different enums for Users and Filters, and some logic on the User model which maps them for filtering.
```graphql
enum PositionDurationInterest {
  TEMPORARY_OR_PERMANENT
  PERMANENT_ONLY
}
type User {
  ...
  positionDuration: PositionDurationInterest  # what durations am I interested in
}
enum PositionDuration {
  TEMPORARY
  PERMANENT
}
type ApplicantFilter {
  ..
  positionDuration: PositionDuration # what duration the manager can offer.
}
# Filter logic:
if (offeredDuration == TEMPORARY) {
  query->where(positionDuration, 'TEMPORARY_OR_PERMANENT')
}
# In all other cases (offering a permanent position, or not using this filter at all, we do not filter by positionDuration
```
Pros: builds our domain expecations into the data model
Cons: builds our (possibly falacious?) domain expectations into the data model

## Acceptance Criteria
- [ ] Deprecate the current "wouldAcceptTemporary" field without totally deleting or breaking it (good dev practice)
- [ ] Add new fields to schema, according to accepted proposal. Note, input definitions need to be updated as well.
- [ ] Data migration to create new column matching new fields. Should default to null (for future rows) but then also run an sql operation which initializes values for existing rows according to the wouldAcceptTemporary column.
- [ ] Add filters to PoolCandidate and User models
- [ ] phpunit tests for using those filters
- [ ] Ensure that the frontend always uses the new fields instead of wouldAcceptTemporary

vd1992:
closes #4667 
introducing the new and improved duration field
lotta crevices and cracks to check, whew

### DATABASE
- migration is successful and maps over duration

### TALENT SEARCH
- profile works
- updating profile works
- no profile bugs were introduced (like completeness) 
- applying still works
- search + request creation still works

### ADMIN
- request created above works, you can open it up and the parameters match up
- user table still works
- user table filters correctly (specifically duration)
- CSV and print view look good
- candidate profiles look good


vd1992:
Chromatic run 
https://www.chromatic.com/pullrequest?appId=61e099a1bb1465003a73d3ce&number=4956

api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
```diff
@@ -70,8 +70,8 @@ public function __invoke($_, array $args)
             }
 
             // wouldAcceptTemporary
-            if (array_key_exists('wouldAcceptTemporary', $filters)) {
-                User::scopeWouldAcceptTemporary($userQuery, $filters['wouldAcceptTemporary']);
+            if (array_key_exists('positionDuration', $filters)) {
```
petertgiles, line 73: don't forget to update the comment

api/database/migrations/2022_11_25_232602_would_accept_temporary_to_position_duration.php
```diff
@@ -0,0 +1,68 @@
...
+use Illuminate\Support\Facades\DB;
+
+class WouldAcceptTemporaryToPositionDuration extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
```
petertgiles, line 16: I'd recommend actually removing the `would_accept_temporary` column after you've mapped the data across.  You can leave that deprecated column as a calculated field on the model dervied from position_duration.  That will make removing the field eventually much easier after the deprecation period.

api/database/migrations/2022_11_25_232602_would_accept_temporary_to_position_duration.php
```diff
@@ -0,0 +1,68 @@
...
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->jsonb('position_duration')->nullable()->default(null);
+        });
+        Schema::table('applicant_filters', function (Blueprint $table) {
+            $table->jsonb('position_duration')->nullable()->default(null);
+        });
+
+        DB::statement(
```
esizer, line 25: Should we drop the `would_accept_temporary` column after migrating the data? Or are we leaving it in for backwards compatibility?

frontend/common/src/components/UserProfile/ProfileSections/WorkPreferencesSection.tsx
```diff
@@ -64,7 +68,7 @@ const WorkPreferencesSection: React.FunctionComponent<{
   return (
     <Well>
       <div data-h2-flex-grid="base(flex-start, x2, x1)">
-        {wouldAcceptTemporary === null && (
+        {positionDuration === null && (
```
petertgiles, line 71: Do we need to check for an empty array here?

esizer (changes requested):
Nice, looking great. I do have a few questions/suggestions though :)

frontend/common/src/components/UserProfile/ProfileSections/WorkPreferencesSection.tsx
```diff
@@ -86,53 +90,55 @@ const WorkPreferencesSection: React.FunctionComponent<{
...
-              </li>
-            </ul>
-          </div>
-        )}
+              </p>
+              <ul data-h2-padding="base(0, 0, 0, x1)">
+                <li>
+                  {intl.formatMessage({
+                    defaultMessage:
+                      "any duration. (short term, long term, or indeterminate duration)",
```
petertgiles, line 109: We'll need new copy here, right?  And you'd want to combine the two blocks into a single `ul`?

petertgiles (changes requested):
This is looking great!  I've done a readthrough and flagged a few small things.  I'll try and build it to test soon.

api/database/migrations/2022_11_25_232602_would_accept_temporary_to_position_duration.php
vd1992: acceptance criteria was to deprecate without totally breaking it, so yeah I think it is staying for a bit

api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
```diff
@@ -70,8 +70,8 @@ public function __invoke($_, array $args)
             }
 
             // wouldAcceptTemporary
```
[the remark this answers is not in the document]
vd1992: sure 
https://github.com/GCTC-NTGC/gc-digital-talent/pull/4956/commits/3500ed3b81d9c0c2d4d3ccce5d3251b1bc7f0afe

api/app/Models/User.php
```diff
@@ -530,6 +532,25 @@ public static function scopeWouldAcceptTemporary(Builder $query, ?bool $wouldAcc
...
+                if ($index === 0) {
+                    // First iteration must use where instead of orWhere
+                    $query->whereJsonContains('position_duration', $duration);
+                } else {
+                    $query->orWhereJsonContains('position_duration', $duration);
+                }
+            }
+        });
+        return $query;
+    }
```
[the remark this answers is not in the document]
vd1992: https://github.com/GCTC-NTGC/gc-digital-talent/pull/4956/commits/3500ed3b81d9c0c2d4d3ccce5d3251b1bc7f0afe

api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
vd1992: https://github.com/GCTC-NTGC/gc-digital-talent/pull/4956/commits/3500ed3b81d9c0c2d4d3ccce5d3251b1bc7f0afe
done it

api/database/migrations/2022_11_25_232602_would_accept_temporary_to_position_duration.php
vd1992: I will ask Tristan

vd1992: @tristan-orourke 
For the acceptance criteria task of deprecating wouldAcceptTemporary without totally breaking it, should I do this recommendation?
Or stick with what I have done? (keep it on the database, mark as deprecated wherever possible and remove from Schema otherwise)

frontend/common/src/components/UserProfile/ProfileSections/WorkPreferencesSection.tsx
vd1992: I am hesitant to change copy, I think this change at the moment is more invisible and for ease-of-developer-understanding purposes than changing actual frontend

frontend/common/src/components/UserProfile/ProfileSections/WorkPreferencesSection.tsx
vd1992: oh, yeah I think that would be important

https://github.com/GCTC-NTGC/gc-digital-talent/pull/4956/commits/72629730fa3fc394d4560478082d0eb0a85ceb4b

frontend/common/src/components/UserProfile/ProfileSections/WorkPreferencesSection.tsx
tristan-orourke: I agree, lets leave this as a backend refactor for now. We have another ticket about improving the copy/inputs, which we should flag again for designers: https://github.com/GCTC-NTGC/gc-digital-talent/issues/4671

api/database/migrations/2022_11_25_232602_would_accept_temporary_to_position_duration.php
tristan-orourke: I was originally thinking of it exactly like you've implemented it @vd1992, but I like Peter's recommendation here. It may be safer to keep the database restricted to a single source of truth. We can keep the graphql wouldAcceptAttribute working by defining a [Laravel accessor](https://laravel.com/docs/8.x/eloquent-mutators#defining-an-accessor) for it which reverses the calculation you did in this migration.

vd1992: excellent, thanks!
I wanted to double check before dropping anything from the tables

vd1992: done in  https://github.com/GCTC-NTGC/gc-digital-talent/pull/4956/commits/a7fda4ba5843ef8e7e50732bd69e86f93442a572  and  https://github.com/GCTC-NTGC/gc-digital-talent/pull/4956/commits/7cf989b7a27ead986e991d47021457d5fb8c1f1b

api/database/migrations/2022_11_25_232602_would_accept_temporary_to_position_duration.php
```diff
@@ -0,0 +1,75 @@
...
+     */
+    public function down()
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->dropColumn('position_duration');
+        });
+        Schema::table('applicant_filters', function (Blueprint $table) {
+            $table->dropColumn('position_duration');
+        });
+    }
```
esizer, line 74: Since we are dropping the `would_accept_temporary` columns, they should be added back in the `down` method so we can step through rollbacks if necessary.

esizer (changes requested):
Looking good! Just need to update the `down` migration so we don't lose data if we rollback 😄

api/database/migrations/2022_11_25_232602_would_accept_temporary_to_position_duration.php
vd1992: ah snap, yeah!
done in  https://github.com/GCTC-NTGC/gc-digital-talent/pull/4956/commits/8ee7546b93c12ef53106a45829cdafcd9d73b908  and  https://github.com/GCTC-NTGC/gc-digital-talent/pull/4956/commits/1f98bb8b12f3ce2e918daf26e7449772fc57757f

couldn't get the reversal to work using `case`, error messages had me giving up, so I went with a bit of a different approach

esizer (approved):
Awesome! Everything seems to be working for me 😄

frontend/talentsearch/src/js/components/search/SearchForm.tsx
```diff
@@ -173,8 +174,10 @@ const SearchForm = React.forwardRef<SearchFormRef, SearchFormProps>(
           ...(values.languageAbility !== NullSelection
             ? { languageAbility: values.languageAbility as LanguageAbility }
             : {}), // Ensure null in FormValues is converted to undefined
-          wouldAcceptTemporary:
-            values.employmentDuration === "true" ? true : null,
+          positionDuration:
```
petertgiles, line 177: Technically this works but this PR is to make it not work like a boolean.  I think it' worth updating the radio selections below to use the new constants rather than continuing to use this like a boolean.  I think the page should send "PERMANENT" to the API, not null, when the permanent radio is selected.

frontend/talentsearch/src/js/components/request/CreateRequest.tsx
```diff
@@ -52,7 +53,7 @@ type FormValues = {
       sync?: Array<Maybe<Skill["id"]>>;
     };
     hasDiploma?: ApplicantFilterInput["hasDiploma"];
-    wouldAcceptTemporary?: ApplicantFilterInput["wouldAcceptTemporary"];
+    positionDuration?: ApplicantFilterInput["positionDuration"];
```
petertgiles, line 56: I noticed that the request page still shows "Indeterminate duration (permanent)" if you don't select anything on the search page.  Since we don't want to treat this as boolean any longer I think it would make more sense if it showed the "(None selected)" string.  And left that field as undefined in the mutation.
image
Similarily, I think the permanent radio should send "PERMANENT" and not null to the mutation.

petertgiles (changes requested):
It's looking solid!  I think everything is working as it should but if we really want this to stop acting like a boolean I think you need to go a little further.

api/database/migrations/2022_11_25_232602_would_accept_temporary_to_position_duration.php
```diff
@@ -0,0 +1,113 @@
...
+
+        DB::statement(
+            <<<SQL
+                UPDATE users
+                    SET would_accept_temporary = false
+                    WHERE ((position_duration ?? :duration) = false)
+            SQL, [
+                'duration' => ApiEnums::POSITION_DURATION_TEMPORARY,
+        ]);
+        // can't do a !?? operation
```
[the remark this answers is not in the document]
vd1992: don't want to wrestle with that since I spent a while struggling with SQL yesterday and gave up on using `case`
let's not risk bugs 😬

frontend/talentsearch/src/js/components/search/SearchForm.tsx
vd1992: changed search/request to send PERMANENT
changed message for no duration selection
https://github.com/GCTC-NTGC/gc-digital-talent/pull/4956/commits/078d061d253a5acaa85792422833caa090b75e14
https://github.com/GCTC-NTGC/gc-digital-talent/pull/4956/commits/65ea0fe52c5f4f11e249baeec3175caa2242c05c

frontend/talentsearch/src/js/components/search/SearchForm.tsx
```diff
@@ -482,13 +496,13 @@ const SearchForm = React.forwardRef<SearchFormRef, SearchFormProps>(
                   }),
                 },
                 {
-                  value: "true",
+                  value: "term",
```
petertgiles, line 499: I'd recommend using `EmploymentDuration.Term` and `EmploymentDuration.Indeterminate` instead of making up your own strings to use here and in `durationSelectionToEnum`.

vd1992: like this https://github.com/GCTC-NTGC/gc-digital-talent/pull/4956/commits/30df366b54ee314d1e41325b97f0a21cf4a6782f  ?

api/database/migrations/2022_11_25_232602_would_accept_temporary_to_position_duration.php
tristan-orourke: Thanks! I came back to check this was done 😛

PR #4957, opened 2022-11-29
Tweaks to handling of skills during application

### LINKED ISSUE #4886: Hide N/A from the skills in experience
## Description

We removed the requirement for the "Skills in Detail" to be filled out during My Profile page.

But reviewing the info now seems like something went wrong since N/A appears. We should hide N/A from the experience accordion when it's empty.

image

## Acceptance Criteria
- [x] skills definitions do not need to show up on the exp accordions , they are only useful when adding skills or editing the skills in detail, we should remove them from the accordions in both pages
- [x] Skills in detail should hide when empty instead of displaying N/A (works on application but not on Exp and skills page
- [x] Required skills should check for "skills in detail" being filled out NOT JUST whether it is on the profile or not.

JamesHuf:
Resolves #4886.

Removes skill descriptions from accordions and made MissingSkills check for "skills in detail" being set for the skill.

To test:
1. Build the site
2. Look at a published pool advertisement and it's required skills but don't start an application to it.
3. Navigate to your profile
4. Confirm the experience accordions don't include the skill descriptions.
5. Add a new experience with one of the required skills from step 2 you don't already have.
6. Return to your profile
7. Confirm your new experience displays nothing instead "N/A" under the skill where the "Skills in detail" text would be displayed.
8. Start an application to the pool advertisement in step 2.
9. On the apply part one page: Confirm the skill added in step 5 is listed as missing from your profile.
10. Edit your new experience and fill out the "skills in detail" for the skill.
11. Return to the apply part one page.
12. Confirm the skill is no longer listed as missing from your profile.

### DIFF
--- a/frontend/common/src/components/MissingSkills/MissingSkills.stories.tsx
+++ b/frontend/common/src/components/MissingSkills/MissingSkills.stories.tsx
@@ -53,7 +53,7 @@ MissingSomeSkills.args = {
   requiredSkills: fakeRequiredSkills,
   optionalSkills: fakeOptionalSkills,
   addedSkills: [
-    ...fakeRequiredSkills.splice(0, skills.length / 2),
-    ...fakeOptionalSkills.splice(0, skills.length / 2),
+    ...fakeRequiredSkills.slice(0, fakeRequiredSkills.length / 2),
+    ...fakeOptionalSkills.slice(0, fakeOptionalSkills.length / 2),
   ],
 };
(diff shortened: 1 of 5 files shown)


esizer (commented):
Looking great, I just don't fully understand the acceptance criteria 😅 

@Jerryescandon Can you confirm that the goal of this is to simple remove skill descriptions from the accordions and only show "skill in detail" if it exists?

Jerryescandon:
> Looking great, I just don't fully understand the acceptance criteria 😅
> 
> @Jerryescandon Can you confirm that the goal of this is to simple remove skill descriptions from the accordions and only show "skill in detail" if it exists?

That's right

esizer (approved):
Awesome! 😁

PR #4985, opened 2022-12-01
Remove `Radio` component

### LINKED ISSUE #4566: Unused `Radio` component
## Description
The `Radio` component is not used for anywhere and should not be since there is no use case for an element that a user can select and then not deselect. The `RadioGroup` component is used for when radio inputs are required.

## Acceptance Criteria
- [ ] remove `Radio` component from codebase

mnigh:
Resolves #4566.

## Steps to test
1. Confirm there are no instances of the `Radio` component in the codebase


esizer (approved):
✅ Tests passing, can't see it being used anywhere, file is gone.

Awesome! 😁

PR #4988, opened 2022-12-01
Bug -  4925 - Prevent caching chunk load error

### LINKED ISSUE #4925: Prevent over-caching causing ChunkLoadError
## 👋 Introduction

We are still seeing `ChunkLoadError` after prod releases and need to do some more work to prevent that.

REF: https://talent-cloud.slack.com/archives/C0259G6KXJ6/p1669398368036509

## 🕵️ Details

After doing some digging, we now believe it is related to `index.html` and `app.js` being over-cached causing the browser to look for the incorrect files.

## 🛠️ Possible Solution

There are two things that can help prevent this (possibly)

1. Add no-cache meta tags to `index.html` 
2. Move the hash to filename rather than search param

### 1. No Cache Meta Tags

This would simply be adding the following tags to our `index.html` files:

```html
<meta http-equiv="Cache-Control" content="no-cache, no-store, must-revalidate" />
<meta http-equiv="Pragma" content="no-cache" />
```

### 2. Move Hash to Filename

In our `webpack.base.js` files, we are adding the hash as `?id=[contenthash]`. Changing this may help prevent this.

```js
  output: {
    publicPath: "/talent/", // final path for routing
    filename: "[hash].[name].js", // file hashing for cache busting
    chunkFilename: "[hash].[name].js", // file hashing for cache busting
    ...
  },
```

REF: https://raphael-leger.medium.com/react-webpack-chunkloaderror-loading-chunk-x-failed-ac385bd110e0

## ✅ Acceptance Criteria

- [ ] No cache meta tags added to `index.html` files
- [ ] Move to file name hash in webpack config

esizer:
## 👋 Introduction

This does two things:

1. Adds a `Cache-Control: no-cache` header to reduce over-caching `index.html` files
2. Moves the `*.js` file hash to the file name rather than query param and changes it to `hash` instead of `contenthash`

The idea is that both of these will help reduce the chances old files will be cached in the client and users will not encounter `ChunkLoadErrors` after new changes are deployed.

## 🧪 Testing

@vd1992 is realy good at reproducing this issue locally but I haven't had much luck so I'm not sure how to test that it fixes `ChunkLoadError`. However, you can follow these steps to ensure the changes made were implmented:

1. Rebuild `webserver` docker image
    - `docker-compose stop webserver`
    - `docker-compose rm webserver`
    - `docker-compose up --detach --build`
2. Navigate to `/`
3. Check the response headers for `Cache-Control: no-cache`
4. Check the `*.js` files loaded and confirm they match the pattern `[hash].[name].js`

## 🤖 Robot stuff

Resolves #4925

### DIFF
--- a/frontend/admin/webpack.base.js
+++ b/frontend/admin/webpack.base.js
@@ -154,8 +154,8 @@ module.exports = {
   },
   output: {
     publicPath: "/admin/", // final path for routing
-    filename: "[name].js?id=[contenthash]", // file hashing for cache busting
-    chunkFilename: "[name].js?id=[contenthash]",
+    filename: "[hash].[name].js", // file hashing for cache busting
+    chunkFilename: "[hash].[name].js", // file hashing for cache busting
     path: path.resolve(__dirname, "dist"), // output folder
     clean: true, // delete existing files on recompile
   },
--- a/infrastructure/conf/nginx-conf-local/default
+++ b/infrastructure/conf/nginx-conf-local/default
@@ -85,6 +85,8 @@ server {
 
     # rewrite root-level possible lang prefix to talentsearch index.html
     location ~ "^(?:/[a-z]{2})?/?$" {
+        add_header Cache-Control no-cache;
+        expires 0;
         rewrite "^(?:/[a-z]{2})?/?$" /frontend/talentsearch/dist/index.html break;
     }
 
@@ -95,6 +97,8 @@ server {
 
     # rewrite possible lang prefix then "admin" to admin index.html
     location ~ "^(?:/[a-z]{2})?/admin(/?.*|$)" {
+        add_header Cache-Control no-cache;
+        expires 0;
         rewrite "^(?:/[a-z]{2})?/admin(/?.*|$)" /frontend/admin/dist/index.html break;
     }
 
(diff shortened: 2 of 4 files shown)


vd1992:
as well, there is a deprecation warning

image

esizer:
> as well, there is a deprecation warning

Good catch! I wasn't aware of this. Reverted in [change hash back to contenthash](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4988/commits/9e6e318f0b0a63dcdbb79de7e98aaa54e29e1361)

vd1992:
I think things look nice! Is the Bundlewatch failure a simple fix?
As well, I think someone else should also check out this PR so I am not going to approve it yet

esizer:
> I think things look nice! Is the Bundlewatch failure a simple fix?

Oops, yeah just had to add a `*` since we now have the hash in the filename 😅 .

Fixed in [wildcard bundlewatch to fix hash in filename](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4988/commits/0462d3aa9af0bf70da4dc61397d82e03b541217b)

frontend/admin/webpack.base.js
```diff
@@ -154,8 +154,8 @@ module.exports = {
   },
   output: {
     publicPath: "/admin/", // final path for routing
-    filename: "[name].js?id=[contenthash]", // file hashing for cache busting
-    chunkFilename: "[name].js?id=[contenthash]",
+    filename: "[contenthash].[name].js", // file hashing for cache busting
```
petertgiles, line 157: Might be easier to find files in alphabetical listing if it was [name].[contenthash].js instead, no?  Is there an advantage to starting with the hash?

infrastructure/conf/nginx-conf-local/default
```diff
@@ -85,6 +85,8 @@ server {
 
     # rewrite root-level possible lang prefix to talentsearch index.html
     location ~ "^(?:/[a-z]{2})?/?$" {
+        add_header Cache-Control no-cache;
```
petertgiles, line 88: Don't forget to mirror these changes in infrastructure/conf/nginx-conf-deploy/default

frontend/admin/webpack.base.js
```diff
@@ -154,8 +154,8 @@ module.exports = {
...
+    chunkFilename: "[contenthash].[name].js", // file hashing for cache busting
```
petertgiles, line 158: Don't forget the CSS hash on line 46

petertgiles (changes requested):
So I tried really hard but was not able to replicate the issue myself.  I'm hesistent to merge any changes to main until we're able to do that.  I'm a little worried about trying to fix it with headers, too.  The headers are really different between local and prod and I bet our reverse-proxy is modifying them.

frontend/admin/webpack.base.js
esizer: I can't think of any advantage, I was just using the referenced example. But, I do agree moving it to the end makes it more human friendly :)

Updated in [move content hash to end of filename](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4988/commits/06e5fcfb5330180c39eff076e76c647df73dd515)

infrastructure/conf/nginx-conf-local/default
esizer: Completely forgot about that 😞 

Added in [add no cache header to nginx deploy conf](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4988/commits/a44179d016416a6845b70c576d4d8c1c9236ec40)

frontend/admin/webpack.base.js
esizer: Oops! I didn't even see that line!

Updated in [update css output to have hash in filename](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4988/commits/f3c65517d91453369882fef5ac140f6a969dd741)

petertgiles:
I dunno how to replicate this.  I was playing with the webpackChunkName hints in the router to force new chunks with no names to be generated but every time I navigated back to the site the combination of  `If-Modified-Since` and `If-None-Match` headers were able to figure out if new index.html and app.js files were needed or could be cached.  I used FoxyProxy to make sure I wasn't getting modified "localhost" behavior, too.

Strangely, in production my browser isn't using the "if" headers so I wonder if that's we we're sometimes seeing this in production and not locally?  I'm not clear on why the header behavior is different in prod but I suspect it has to do with our upstream reverse proxy.  I thought @vd1992 was able to see this locally, though.

esizer:
> I dunno how to replicate this. I was playing with the webpackChunkName hints in the router to force new chunks with different names to be generated but every time I navigated back to the site the combination of `If-Modified-Since` and `If-None-Match` headers were able to figure out if new index.html and app.js files were needed or could be cached. I used FoxyProxy to make sure I wasn't getting modified "localhost" behavior, too.
> 
> Strangely, in production my browser isn't using the "if" headers so I wonder if that's why we're sometimes seeing this in production and not locally? I'm not clear on why the header behavior is different in prod but I suspect it has to do with our upstream reverse proxy. I thought @vd1992 was able to see this locally, though.

😭

vd1992:
> I dunno how to replicate this. I was playing with the webpackChunkName hints in the router to force new chunks with different names to be generated but every time I navigated back to the site the combination of `If-Modified-Since` and `If-None-Match` headers were able to figure out if new index.html and app.js files were needed or could be cached. I used FoxyProxy to make sure I wasn't getting modified "localhost" behavior, too.
> 
> Strangely, in production my browser isn't using the "if" headers so I wonder if that's why we're sometimes seeing this in production and not locally? I'm not clear on why the header behavior is different in prod but I suspect it has to do with our upstream reverse proxy. I thought @vd1992 was able to see this locally, though.

locally I had it happen very rarely I think, it was on prod I got it to happen consistently

petertgiles (approved):

PR #4993, opened 2022-12-02
Bug - 4921 - Validate UUID queries

### LINKED ISSUE #4921: Handle Invalid Input (UUID) Error in API
## 👋 Introduction

We are using UUIDs in the route parameters to access specific entities in the API. However, since users can put **_anything_** they want in the URL, they can put invalid UUIDs. Unfortunately, this causes our `graphql` server to throw a `500` with not much useful information. This makes it difficult to handle gracefully on the frontend and has lead us to show a generic `404` in the even to **_any_** error.

## 🛑 Error

This is what the error looks like.

```js
[
    {
        "message": "Internal server error",
        "path": [
            "poolAdvertisement"
        ],
        "extensions": {
            "category": "internal"
        }
    }
]
```

## ❓ What is Happening?

The real error is to do with invalid input, we expect a properly formatting UUID to be passed. Unfortunately, instead of sending a proper error code (`400` or `422`) we get a generic 500 error.

## ✅ Acceptance Criteria

- [ ] Determine appropriate status code (`400` or `422`)
- [ ] Handle bad UUIDs on server
- [ ] Send appropriate error to frontend
- [ ] Handle the error for pages

esizer:
## 👋 Introduction

This adds `uuid` validation rule to the input on our find queries as well as handling the errors on the frontend.

### Toast Error Message

~~We now get the error code `validation.uuid` when a user types in a non-uuid into the URL bar. Since we show a toast with localized messages for errors, what message would be a good catch all? I doubt users would understand/care about an invalid `uuid`. Should we simply say "Page not found"?~~

> **Note**
> We technically no longer error with the scalar so the toast does not happent

### Approach

~~We are now getting a proper error message and not just throwing a `404` on any error. Does this seem like a better approach that we are happy with?~~

> **Note**
> We moved to using a scalar (`UUID`) instead of laravel validation in the hopes that https://github.com/nuwave/lighthouse/issues/1751 is resolved at some point.

## 🧪 Testing

2. Navigate to a page that accepts a UUID in the route with an invalid uuid i.e(`/browse/pools/not-an-id`)
3. Confirm you are sent to the 404 page 

## 🤖 Robot stuff

Resolves #4921


api/graphql/schema.graphql
```diff
@@ -17,6 +17,8 @@ scalar KeyString
 "Arbitrary data encoded in JavaScript Object Notation. See https://www.json.org."
 scalar JSON @scalar(class: "MLL\\GraphQLScalars\\JSON")
 
+scalar UUID @scalar(class: "UUID")
```
petertgiles, line 20: Could you add a little comment here to introduce the scalar?

petertgiles (approved):
This is a decent solution for now.  Hopefully lighthouse improves their error reporting soon.  Don't forget to tidy up the phpunit tests.

PR #4998, opened 2022-12-02
Improve Search Workflows Cypress Test

### LINKED ISSUE #4357: Ensure tests are reproducible
## Summary

Most of our database seeders are using `fakerphp/faker` to generate random data for testing. However, for specific test cases this is resulting tests that are no reproducible. 

## Example

For the `search-workflows.spec.js` we are searching for users requiring that they have a diploma and expecting at least one result. However, we are randomly setting that value for all users and there is a small chance (which is occuring) where no user seeded has a diploma so it causes a false positive.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/c65ca74177ce2ae578b92e2c3965bbb6d2362412/api/database/factories/UserFactory.php#L107

https://github.com/GCTC-NTGC/gc-digital-talent/blob/c65ca74177ce2ae578b92e2c3965bbb6d2362412/frontend/cypress/integration/talentsearch/search-workflows.spec.js#L30-L33

https://github.com/GCTC-NTGC/gc-digital-talent/blob/c65ca74177ce2ae578b92e2c3965bbb6d2362412/frontend/cypress/integration/talentsearch/search-workflows.spec.js#L50

## Solution

There are some options to solve this, in order of preference(?):

1. Improve our tests so that we cover edge cases less and make fewer assumptions about the data
    - However unlikely, it is a valid that we have no users with diplomas
2. Update our seeders to ensure we cover our test cases
    - Undesirable since it presents a biased view of how we **_expect_** the data to look rather than what it **_could_** look like

petertgiles:
# :wave: Introduction
This branch improves the search workflows cypress test by removing assumptions about the data in the database and also more careful testing of the results.

The test assumes that the standard skills, classifications, etc are seeded to the database.  However, each run of the test creates new pools, users, and pool candidates to test with to avoid a situation where random changes has failed to create the right test context.  

This setup allows the test to know that exactly one pool candidate should match a given set of criteria.  It also allows the test to check that the candidate does not match a given set of criteria.

# :test_tube:  Testing
- [ ] The search workflows test consistently passes locally
- [ ] The search workflows test consistently passes in CI

# :dart: Scope
I spent way too much time on this single test.  I'll open new issues for the other tests that need to be refactored but not do them now.

# :robot: Robot stuff
closes #4357


frontend/talentsearch/src/js/components/search/SearchPools.tsx
```diff
@@ -32,8 +32,11 @@ const SearchPools: React.FunctionComponent<SearchPoolsProps> = ({
     pool.classifications as SimpleClassification[];
 
   return (
-    <div data-h2-padding="base(x1)">
-      <p data-h2-font-weight="base(700)">
+    <article
+      data-h2-padding="base(x1)"
+      aria-labelledby={`search_pool_${pool.id}`}
+    >
+      <p data-h2-font-weight="base(700)" id={`search_pool_${pool.id}`}>
```
vd1992, line 39: a doozy to wrap your head around
a question or two, is this for accessibility purposes and not testing purposes? enabling linking to a card?

vd1992: is there any point where test run time is considered a problem?

image

petertgiles: > is this for accessibility purposes and not testing purposes? enabling linking to a card

Yeah, it's for testing purposes.  It's probably good for a11y too but it's mostly there to help Cypress find the right card.

petertgiles: > is there any point where test run time is considered a problem?

Certainly there has to be a line somewhere.  This test could be broken up to test each filter individually but in aggregate it would take much longer since the test would have to be set up over and over instead of just once like it is now.  I'd suggest that the time is well spent on this test since this page has been a rich source of bugs so far.

patcon:
grrrr... hitting this locally, but it's def me to sort out. Gonna take dins and come back to sort!

Screen Shot 2022-12-06 at 7 09 33 PM

petertgiles:
> but it's def me to sort out

It's part of the Laravel 9 upgrade path.
https://talent-cloud.slack.com/archives/C0259G6KXJ6/p1669925782729109

patcon:
## patcon todo
- [x] ensure `search-workflows.cy.js` passing locally
- [x] ensure `search-workflows.cy.js` passing on CI
- [ ] review code itself
- [ ] create PR with small script/config fixes

frontend/talentsearch/src/js/components/search/SearchPools.tsx
patcon, line 39: why not use `data-test` attr, which is a cypress recommended way (that we mostly ignore).
https://docs.cypress.io/guides/references/best-practices#How-It-Works

The nice part about this is that there are already webpack plugins to filter any of this out of prod builds (if we want), whereas using this is hard to strip later.

```suggestion
      data-test={`search_pool_${pool.id}`}
    >
      <p data-h2-font-weight="base(700)" data-test={`search_pool_${pool.id}`}>
```

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
```diff
@@ -8,89 +11,308 @@ describe("Talent Search Workflow Tests", () => {
...
+
+  it("searches for a candidate with all the filters and submits a request", () => {
+    // have a unique ID to track the test objects with
+    const uniqueTestId = Date.now().valueOf();
+    cy.log(`Test run ${uniqueTestId}`);
+
+    const searchFindsMySingleCandidate = () => {
+      cy.findAllByRole("article", {
+        name: `Cypress Test Pool EN 1 ${uniqueTestId} (IT-01 Business Line Advisory Services)`,
+      }).within(() => {
```
patcon, line 43:
```suggestion
      cyget(`article[data-test=${uniqueTestId}]`).within(() => {
```

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
```diff
@@ -8,89 +11,308 @@ describe("Talent Search Workflow Tests", () => {
...
+          .should("exist")
+          .and("be.visible")
+          .and("not.be.disabled");
+      });
+    };
+
+    const searchRejectsMySingleCandidate = () => {
+      cy.findAllByRole("article", {
+        name: `Cypress Test Pool 1 EN ${uniqueTestId} (IT-01 Business Line Advisory Services)`,
+      }).should("not.exist");
```
patcon, line 56:
```suggestion
      cy.get(`article[data-test=${uniqueTestId}]`).within(() => {
```

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
```diff
@@ -8,89 +11,308 @@ describe("Talent Search Workflow Tests", () => {
...
+    });
 
-    cy.findAllByRole("button", { name: /Request Candidates/i }).then($button => {
-      cy.wrap($button)
-        .should("exist")
-        .and("be.visible")
-        .and("not.be.disabled");
+    cy.findAllByRole("article", {
+      name: `Cypress Test Pool EN 1 ${uniqueTestId} (IT-01 Business Line Advisory Services)`,
+    }).within(() => {
```
patcon, line 251:
```suggestion
      cy.get(`article[data-test=${uniqueTestId}]`).within(() => {
```

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
patcon: findAllByRole can return many, and within only works for one:

https://docs.cypress.io/api/commands/within#Requirements

> ## Requirements 
>
> `.within()` requires being chained off a previous command that yields exactly one DOM element.

frontend/admin/src/js/components/pool/EditPool/editPoolOperations.graphql
```diff
@@ -142,11 +142,13 @@ mutation updatePoolAdvertisement(
 
 mutation publishPoolAdvertisement($id: ID!) {
   publishPoolAdvertisement(id: $id) {
+    id
```
patcon, line 145: nice, i like the idea of always returning the ID 👍  seems like this was just a lone straggler, and that's the default :)

frontend/talentsearch/src/js/components/search/SearchPools.tsx
patcon: hm.... I'm not seeing "search_pool_" anywhere else. Did this end up being used in selectors?

frontend/cypress/support/applicationCommands.js
```diff
@@ -0,0 +1,72 @@
...
+    },
+    body: {
+      operationName: 'createApplication',
+      query: getGqlString(CreateApplicationDocument),
+      variables: {
+        userId,
+        poolId
+      },
+    },
+  })
```
patcon, line 23: Maybe package this into `cy.graphqlRequest()` and pack the method/url/auth in there?

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
```diff
@@ -8,89 +11,308 @@ describe("Talent Search Workflow Tests", () => {
...
+      .toString()
+      .padStart(2, "0")}`;
+
+  it("searches for a candidate with all the filters and submits a request", () => {
+    // have a unique ID to track the test objects with
+    const uniqueTestId = Date.now().valueOf();
+    cy.log(`Test run ${uniqueTestId}`);
+
+    const searchFindsMySingleCandidate = () => {
+      cy.findAllByRole("article", {
```
patcon, line 41: Confirmed this works :) Re: https://github.com/GCTC-NTGC/gc-digital-talent/pull/4998/files#r1042304305

```suggestion
      cy.findByRole("article", {
```

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
patcon: Gonna resolve this, as it works fine even if a little more fragile imho :)

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
```diff
@@ -8,89 +11,308 @@ describe("Talent Search Workflow Tests", () => {
...
+
+        cy.findByRole("button", { name: /Request Candidates/i })
+          .should("exist")
+          .and("be.visible")
+          .and("not.be.disabled");
+      });
+    };
+
+    const searchRejectsMySingleCandidate = () => {
+      cy.findAllByRole("article", {
```
patcon, line 54: findAllByRole can return many, and within only works for one:

https://docs.cypress.io/api/commands/within#Requirements

> ## Requirements 
>
> `.within()` requires being chained off a previous command that yields exactly one DOM element.

```suggestion
      cy.findByRole("article", {
```

Confirmed this works :)

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
```diff
@@ -8,89 +11,308 @@ describe("Talent Search Workflow Tests", () => {
...
+    cy.findByRole("heading", {
+      name: /Results: [1-9][0-9]* matching candidate/i,
+    });
 
-    cy.findAllByRole("button", { name: /Request Candidates/i }).then($button => {
-      cy.wrap($button)
-        .should("exist")
-        .and("be.visible")
-        .and("not.be.disabled");
+    cy.findAllByRole("article", {
```
patcon, line 249:
```suggestion
    cy.findByRole("article", {
```

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
```diff
@@ -8,89 +11,308 @@ describe("Talent Search Workflow Tests", () => {
...
+    );
 
     cy.findByRole("combobox", { name: /Department/i }).select(
       "Treasury Board Secretariat",
     );
 
-    // will actually exist twice in the DOM
+    // classification filter
+    cy.get("@testClassification1").then((classification) => {
+      cy.findAllByText(classificationName(classification)).should("exist");
```
patcon, line 281: why is findAll* needed here? I'm seeing the code below seem to match the find text as well, but it's hidden. And so I feel like it shouldn't be picked up? so wondering if we're doing something funny that screen-readers will also misread. Seems like it's screen width thing maybe? (a little above the context)

https://github.com/GCTC-NTGC/gc-digital-talent/blob/e0404a71ecbc8b126b64629694f2e3760cf97b4c/frontend/common/src/components/SearchRequestFilters/SearchRequestFilters.tsx#L69-L71

patcon: This comment applies in a few places where findAllByText is used

patcon (changes requested):
neglected the final approval, but LGTM

api/app/Models/Pool.php
```diff
@@ -121,13 +121,13 @@ public function getAdvertisementStatusAttribute()
         $expiryDate = $this->expiry_date;
         $currentTime = date("Y-m-d H:i:s");
         if ($expiryDate != null) {
-            $isExpired = $currentTime > $expiryDate ? true : false;
+            $isExpired = $currentTime >= $expiryDate ? true : false;
```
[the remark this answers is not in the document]
petertgiles: Yeah, Cypress can be _really_ fast.  :laughing:

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
petertgiles: Good catch!  I've changed this to findByRole in https://github.com/GCTC-NTGC/gc-digital-talent/pull/4998/commits/44f59935d54487c1cab50f88ab2dd90f738fdcc8.

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
petertgiles: Good catch!  I've changed this to findByRole in https://github.com/GCTC-NTGC/gc-digital-talent/pull/4998/commits/44f59935d54487c1cab50f88ab2dd90f738fdcc8.

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
petertgiles: Good catch!  I've changed this to findByRole in https://github.com/GCTC-NTGC/gc-digital-talent/pull/4998/commits/44f59935d54487c1cab50f88ab2dd90f738fdcc8.

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
petertgiles: Good catch!  I've changed this to findByRole in https://github.com/GCTC-NTGC/gc-digital-talent/pull/4998/commits/44f59935d54487c1cab50f88ab2dd90f738fdcc8.

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
petertgiles: Good catch!  I've changed this to findByRole in https://github.com/GCTC-NTGC/gc-digital-talent/pull/4998/commits/44f59935d54487c1cab50f88ab2dd90f738fdcc8.

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
petertgiles: Good catch!  I've changed this to findByRole in https://github.com/GCTC-NTGC/gc-digital-talent/pull/4998/commits/44f59935d54487c1cab50f88ab2dd90f738fdcc8.

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
petertgiles: LOL, there are two separate divs on the page, but only one is visible at a time.
image
@esizer This would need an aria-hidden as well to keep screenreaders from reading twice, right?

esizer: What is the purpose behind the hidden one? 🤔

esizer: lol nvm, i see that we switch them out depending on screen size. I don't think `aria-hidden` is necessary because `display: none;` effectively does that. You could remove it from the DOM entirely using the `useIsSmallScreen` instead but that isn't necessary.

frontend/cypress/support/applicationCommands.js
petertgiles: I've added it in https://github.com/GCTC-NTGC/gc-digital-talent/pull/4998/commits/10455e5728f9216afbbbde6532833e4ecefc0834.  It's blown out the size of the PR but I do like it!

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
```diff
@@ -8,89 +11,312 @@ describe("Talent Search Workflow Tests", () => {
...
+      cy.findByRole("article", {
+        name: `Cypress Test Pool 1 EN ${uniqueTestId} (IT-01 Business Line Advisory Services)`,
+      }).should("not.exist");
+    };
+
+    // create a test user to attach test candidates to
+    cy.loginByRole("admin").then(() => {
+      cy.getMe()
+        .its("id")
+        .then((adminUserId) => {
```
patcon, line 63: Ah shoot, I'm sorry, I didn't clue in that we were chaining login commands. There shouldn't be any reason for this, as cypress does things to run individual cypress commands sychronously

So if this is needed, then I think that means we're creating our cypress commands in a non-standard way.

```suggestion
    cy.loginByRole("admin")
    cy.getMe()
      .its("id")
      .then((adminUserId) => {
```

patcon: And we can keep chaining, but my recollection from looking at the docs is that tests usually look more serial, using what cypress provides to operate synchronously between blocks when possible

```suggestion
    # Setup at start of test.
    cy.getMe().as("me")

    cy.loginByRole("admin")
    cy.get("@me").then((adminUser) => {
```

patcon (approved):
LGTM!

Just small flag, but no concrete asks, just noting it:
- I feel maybe we're nesting more than cypress needs, and treating our code like async JS, when cypress is very intentionally not async. (I've also heard it said that nesting in general is a smell in tests)

frontend/cypress/e2e/talentsearch/search-workflows.cy.js
petertgiles: Yeah, I think I started totally synchronous and then went too far asynchronous and need to dial it back.  It's a head trip for sure.  I stripped out the `.then(() => {` chains ([reduce unnecessary chaining](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4998/commits/2d71a3e562dd5e1613212605718d0e29cbabf054)) and the tests are still passing so that seems to be fine.  I think that's as far as I want to go for now.

I saw [this page](https://glebbahmutov.com/blog/avoid-cypress-pyramid-of-doom/) about refactoring pyramids of doom. That might be a good refinement sometime.  I really wish Cypress had the ability to get an array of aliases at the same time like `Promise.all` instead of relying on the values to magically appear on the `this` context.

I'm wondering about seeding well-known uuids for the standard users (admin, applicant, etc) so we could put the IDs back in the user fixture file.  That would save a bunch of `me` queries in the tests.  Another possible improvement some day.

PR #5006, opened 2022-12-02
Tooling/4898 no seeding candidates draft

### LINKED ISSUE #4898: Seeder seeds candidates into draft pools
Our seeder seeds candidates into draft pools.  This causes some weirdness in testing, like going to "My applications" and seeing draft pools.

# Acceptance Criteria:
After a fresh seed, this query should return no results:
```
select p.id, count(*) 
from pools p 
join pool_candidates pc on pc.pool_id = p.id
where p.published_at is null
group by p.id
```

vd1992:
closes #4898 
updates seeder to only add pool candidates to published pools

test:
build and seed fresh
ensure the query suggested by Peter returns no results
```
select p.id, count(*) 
from pools p 
join pool_candidates pc on pc.pool_id = p.id
where p.published_at is null
group by p.id
```
navigate around pools stuff in /admin as well as /search and look around

### DIFF
--- a/api/database/seeders/DatabaseSeeder.php
+++ b/api/database/seeders/DatabaseSeeder.php
@@ -59,9 +59,10 @@ class DatabaseSeeder extends Seeder
                 $genericJobTitles = GenericJobTitle::inRandomOrder()->limit(2)->pluck('id')->toArray();
                 $user->expectedGenericJobTitles()->sync($genericJobTitles);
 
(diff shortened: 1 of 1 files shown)


vd1992:
> This looks right, but I think PoolCandidateFactory could be edited as well, so that any Pools created along with a PoolCandidate are guaranteed to be published.

Good call!

PR #5028, opened 2022-12-06
Feature -  4930 - Add ongoing publishing group

### LINKED ISSUE #4930: Add an IT_JOBS_ONGOING publishing group to schema
# Description
Add an IT_JOBS_ONGOING publishing group to schema enum

# Acceptance Criteria
- [x] ~Add ongoingRecruitment to PoolFactory~
- [ ] Make sure we can/do seed jobs in the new IT_JOBS_ONGOING publishing group.
- [ ] Add a seeder to DevSeeder which creates ongoing pools to match what we expect in production:
  - One ongoing pool for each stream/classification combo (for IT 1-4)
  - For now, add 2 random essential skills to each
  - Actually, two for each combo: one which is expired, one which expires in the future

esizer:
## 👋 Introduction

This adds the new publishing group for ongoing IT pools and adds a seeder that ensures we have them guaranteed in dev.

## ✅ TODO

- [x] Send off for translations
- [ ] Add translations

> **Note**
> Seeder for this is pretty slow (1min)

## 🧪 Testing

1. Refresh the database `docker-compose exec -w /home/site/wwwroot/api webserver sh -c "php artisan migrate:fresh --seed"`
2. Build admin `npm run production --workspace=admin`
3. Navigate to the pools index `/admin/pools`
4. Confirm there are pools seeded for IT-0(1-4) for both active and ongoing, expired/not expired
5. Confirm when editing a pool you can select "IT Jobs (ongoing)"

## 🤖 Robot Stuff

Resolves #4930

### DIFF
--- a/frontend/common/src/lang/fr.json
+++ b/frontend/common/src/lang/fr.json
@@ -1891,5 +1891,9 @@
   "pg1S01": {
     "defaultMessage": "Sauter la liste des compétences.",
     "description": "Link text to skip the list of add skill links"
+  },
+  "1YuZjR": {
+    "defaultMessage": "Emplois en TI (en cours)",
+    "description": "The publishing group called IT Jobs for ongoing recruitments"
   }
 }


tristan-orourke (commented):
Wait, this seriously takes a minute for the seeder to run? 😨

esizer:
> Wait, this seriously takes a minute for the seeder to run? 😨

Unfortunately, yes. The extra stuff we do at the end of the seeder has always been long, this just makes it longer. I wouldn't mind taking a look at it and seeing if we can do some optimizations.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/d0b43cb2d82f2baa0ff4e78ccf5663100eacea68/api/database/seeders/DatabaseSeeder.php#L51-L168

tristan-orourke (changes requested):
I found it difficult to really evaluate what pools I had through the admin UI, so I ran this query on Adminer instead:

[SELECT](https://www.postgresql.org/docs/12/static/sql-select.html) p.id, p.publishing_group, p.published_at, c.name, c.group, c.level, p.stream
FROM [pools](http://localhost:8080/?pgsql=postgres&username=postgres&db=gctalent&ns=public&table=pools) as p, [classifications](http://localhost:8080/?pgsql=postgres&username=postgres&db=gctalent&ns=public&table=classifications) as c, [classification_pool](http://localhost:8080/?pgsql=postgres&username=postgres&db=gctalent&ns=public&table=classification_pool) as pivot
where pivot.classification_id = c.id [AND](https://www.postgresql.org/docs/12/static/functions-logical.html) pivot.pool_id = p.id
ORDER BY p.publishing_group, p.stream

(That suggests we should add some new sortable columns to the pools table - work stream, publishing group, classifications, would probably all be useful.)

After running the seeder, almost all the IT_JOBS_ONGOING jobs are IT-01, instead of being at all four levels. Also, while you're setting expiry date manually, publishished_at seems to be left random. I feel like it might be better to set them all as published?

tristan-orourke:
Let's not worry about the seeder optimization in this PR, but let's make a ticket to try to optimize that.

esizer:
> After running the seeder, almost all the IT_JOBS_ONGOING jobs are IT-01, instead of being at all four levels. Also, while you're setting expiry date manually, publishished_at seems to be left random. I feel like it might be better to set them all as published?

Okay, so looks like this was just a poorly written seeder on my part. I have fixed it to properly seed each level and  made it so all pools are published in [fix pool publishing group seeder](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5028/commits/75000844c2fce258284536f375f26deb44353807)

tristan-orourke (approved):
Looks good now. And translations are added now too, right?

PR #5038, opened 2022-12-06
Set CSpell to ignore message IDs

petertgiles:
Low priority review.  :smile: 

This branch updates the CSpell settings to ignore the message descriptor IDs in the code and reduce false-positive errors.

Resources:
https://regex101.com/r/r9rbNY/1
https://stackoverflow.com/a/475217

Testing:
- [ ] Does it ignore your message descriptors?
- [ ] Does it flag other things wrongly?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -69,5 +69,8 @@
     ".eslintrc",
     "hydrogen.config.json"
   ],
-  "jest.jestCommandLine": "test"
+  "jest.jestCommandLine": "test",
+  "cSpell.ignoreRegExpList": [
+    "/id\\: \\\"[A-Za-z0-9+\\/]{6}\\\"/g"
+  ]
 }


mnigh (dismissed):
I like! 

1. Any ideas on how to get CSpell to ignore the IDs in the fr.json files?
2. What do you think of replacing all the static _lorem ipsum_ with some `faker.lorem` to avoid those warnings?

esizer:
> Low priority review. 😄

I'm going to argue this is high priority since it annoys me to see all the blue lines in text heavy components 😜

petertgiles:
> Actually, both English and French copy can co-exist and be checked.

Oh really?  That's very cool, I didn't know that could be done.  Yeah, I can loosen the regex to just ignore six characters between quotes.  What about a second regex to ignore six characters in quotes, colon, open curly bracket like:
```
 "AyP6Fr": {
 ```

mnigh (approved):
So good!👨‍🍳 

Testing this work allowed me to find a [typo](https://github.com/GCTC-NTGC/gc-digital-talent/issues/5057) so it's already paying dividends!

PR #5044, opened 2022-12-07
Spike - 4741 - Check broken links

esizer:
## 👋 Introduction

This is initial work to check the site for broken links. 

> **Note**
> Admittedly, it is not as thorough as I'd hope but that is proving difficult with cypress.

## 🕵️ Details

This uses a fixture of top level links we know to exist in the application. It loops over the URLs making a request and checking the status code equals `200`. It then visits the page, getting all links on the page and making the same request/assertion on each. Every time this runs, it adds the url to an array to prevent double checking links.

The issue is, we do have some links hidden in dialogs, tabs and accordion that are difficult to check. 🤔 

## 🧪 Testing

1. Reinstall dependencies `npm i`
2. Build talent search `npm run production --workspace=talentsearch`
3. Open cypress `npm run e2e:open`
4. Run the broken links spec

### DIFF
--- a/frontend/cypress/e2e/talentsearch/broken-links.cy.js
+++ b/frontend/cypress/e2e/talentsearch/broken-links.cy.js
@@ -1,0 +1,58 @@
+describe("Broken links", () => {
+
+  /**
+   * Make a request to a url and
+   * check that the status code is 200
+   *
+   * @param {string} url
+   */
+  const checkLink = (url) => {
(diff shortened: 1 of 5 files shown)


esizer:
> dude, that run time

Yeaahh, I would not consider this a solution unless we can get that time down significantly.

esizer:
This is a failed experiment 😢

PR #5043, opened 2022-12-07
Ongoing recruitment section

petertgiles:
# :wave: Introduction
This branch adds an Ongoing Recruitment section to the Browse Opportunities page.
# :monocle_face: Quirks and Features
- Accordion styling has been spun off to #5042
- No copy was provided for the _Information and Data Functions_ stream so it has been excluded
- The copy for the classifications is slightly different from what's on the Role and Salary page.  As per Gray, I've just copied-and-pasted what was provided rather than try to centralize it and make it consistent.

# :test_tube: Testing
- Rebuild the api schema and talentsearch site and visit `/browse/pools`
- ~We don't have seeders for the new pool stream [yet](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5028) so I suggest manually modifying seeded pools to set up testing scenarios.~
- The origin issue has a detailed set of AC to test against

# :robot: Robot Stuff
Closes #4931

PR #5048, opened 2022-12-08
Add quick filter to ongoing recruitments

### LINKED ISSUE #4952: Ongoing Recruitments quick filter
# Acceptance Criteria
- [ ] A dropdown which allows you to select a work stream, which hides the accordions that aren't the option they've selected
- [ ] Good accessibility

petertgiles:
> **Warning**
> This PR is in draft because it is waiting on translations and is blocked by #5043.  It will be rebased on main with a smaller diff once the blocker is merged.

# :wave: Introduction
This branch adds a quick filter to the ongoing recruitment section of the browse opportunities page.

# :monocle_face: Quirks and Features
There is currently no way to disable the filtering though I've asked designers if there should be.

# :test_tube: Testing
- Rebuild the talentsearch site and visit `/browse/pools`.
- We don't have seeders for the new pool stream [yet](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5028) so I suggest manually modifying seeded pools to set up testing scenarios.

# :robot: Robot Stuff
- closes #4952


frontend/talentsearch/src/js/components/Browse/OngoingRecruitmentSection.tsx
```diff
@@ -619,75 +644,128 @@ export const OngoingRecruitmentSection = ({
             "instructions for section with ongoing pool advertisements",
         })}
       </p>
-      <div data-h2-padding="base(x2, 0, 0, 0) p-tablet(x3, 0, 0, 0)">
+      <div data-h2-display="base(flex)">
+        <FormProvider {...methods}>
+          <form>
```
esizer, line 649: If we are going to do this `onChange` then I would drop the `<form>` wrapper. How it is now, it would fail an automated audit for [G80: Providing a submit a submit button to initiate a change of context](https://www.w3.org/TR/WCAG20-TECHS/G80.html). 

I believe the `aria-live` area you added below would be enough to pass a manual test but having the `<form>` wrapper might get flagged in any automated audit and isn't entirely necessary so it may be better to remove it 👍

frontend/talentsearch/src/js/components/Browse/OngoingRecruitmentSection.tsx
```diff
@@ -619,75 +644,128 @@ export const OngoingRecruitmentSection = ({
...
+              ]}
+              trackUnsaved={false}
+            />
+          </form>
+        </FormProvider>
+      </div>
+
+      <p
+        aria-live="polite"
+        aria-atomic="true"
```
esizer, line 682: I made a mistake when I recommended using `aria-atomic=true`, it results in a poorer UX for screen readers. Sorry about that! 😢 

```suggestion
```

esizer (changes requested):
Nice, this is looking great 😄 

Two very minor suggestions.

frontend/talentsearch/src/js/components/Browse/OngoingRecruitmentSection.tsx
petertgiles: Ah, good catch!  Fixed in [Fixes from Eric's review](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5048/commits/49988f14f6c07d06bd7000f1ceb32dc9f9d26e4a)

frontend/talentsearch/src/js/components/Browse/OngoingRecruitmentSection.tsx
esizer: It only announces the content that has changed so it just sounds like it is repeating the option you selected. I thought it would be better but it just sounds odd and the other text isn't that verbose so dropping it seems like a good option.

esizer (approved):
Awesome! 🌟

PR #5051, opened 2022-12-08
Allow admins to add users to published AND expired pools

### LINKED ISSUE #5049: Admins should be able to add users to "expired" Pools
We recently removed this functionality in https://github.com/gctc-ntgc/gc-digital-talent/issues/4875, but it turns out our clients are relying on this ability to add users to "expired" pools. 

(Worth noting that "expired" really just means "closed to direct intake applications", and should maybe be renamed.)

## Acceptance Criteria
- [ ] Admins can place users in expired pools
- [ ] Admins can still NOT place users in draft pools

tristan-orourke:
Resolves #5049 

Display Published _and_ Expired pools in the AddToPool dialog.
Note that we don't have a way of Archiving pools yet, but when we do, archived pools will NOT appear here.

Testing steps:
1. Log in as an admin.
2. Visit http://localhost:8000/en/admin/pools, and note the name of one Draft pool, one Published pool, and one Expired pool. (If any of these do not exist, you may need to run seeders or create them.)
3. Visit http://localhost:8000/en/admin/users
4. Click View on any user
5. On the User page, click "Add user to Pool"
6. In the dialog, confirm that the Draft pool does not appear
7. In the dialog, confirm that the Published pool appears and you can add the user to it
8. In the dialog, confirm the Expired pool appears and you can add the user to it

### DIFF
--- a/frontend/admin/src/js/components/user/GeneralInformationTab/AddToPoolDialog.tsx
+++ b/frontend/admin/src/js/components/user/GeneralInformationTab/AddToPoolDialog.tsx
@@ -153,7 +153,8 @@ export const AddToPoolDialog: React.FC<AddToPoolDialogProps> = ({
                   .filter(
                     (pool) =>
                       pool.advertisementStatus ===
-                      AdvertisementStatus.Published,
+                        AdvertisementStatus.Published ||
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
Great!

Issue #2365, opened 2022-03-31
Refactor graphql Schema to reduce code duplication


tristan-orourke:
# Description
We have a lot of duplication in our graphql schema, eg with User and Applicant, and with certain input types (eg CreateUser and UpdateUser). It would be much easier to maintain if there were some way to modularize and use chunks of these types.

Unfortunately, such features don't appear to be native to graphQL. We can make an effort to use *composition* over *inheritance*, but in some cases (eg User/Applicant) that won't work. 

I have found a library which allows you to inherit types in a way that would be super helpful: https://github.com/nicolasdao/graphql-s2s . We would need to figure out how to get it working along with lighthouse-php.

TIMEBOX: 1 day

## :question: Spike Question
- [ ] What are the best options to refactor out duplication in the schema?

tristan-orourke:
[git push --set-upstream origin feature/self-directive](https://github.com/GCTC-NTGC/gc-digital-talent/tree/feature/self-directive)

tristan-orourke:
Taking another quick look at the [graphql-s2s](https://github.com/nicolasdao/graphql-s2s) library I mentioned above... it's a js library, and won't be much help to our php implementation. 

An alternative approach I was considering would be to define our types in such a way that we can compose types, like User, out of other types, like Applicant. I did a bit of work on a `@self` directive that might aid this, on this unmerged feature branch: https://github.com/GCTC-NTGC/gc-digital-talent/tree/feature/self-directive

tristan-orourke:
## Assumptions
- Avoiding duplicate code is good. It reduces file size, it reduces work involved in maintaining similar types in different places, and reduces the risk of types drifting apart when they shouldn't.
- Reducing the number of directives in the schema is good. Directives reduce the simplicity and readability of the schema by adding hidden functionality as well as visual noise.
- Expressing entity relationships directly in the schema is good. This allows the schema to document the underlying relationships, and gives API clients more flexibility.

## Facts
- The GraphQL spec [does not allow](https://stackoverflow.com/questions/47523384/how-to-inherit-or-extend-typedefs-in-graphql) for inheriting/extending types or inputs as a way to reduce duplication. 
[...]

PR #5052, opened 2022-12-08
Debt - 4672 single search request table

### LINKED ISSUE #4672: List PoolCandidates instead of Users in SingleSearchPage requests
After #4623, when an Admin views a single search request, they see a list of Users. They really want to see a list of PoolCandidates, because they're interested in status and priority.

1. build a poolCandidates query which accepts a ApplicantFilter (we're almost there already, actually)
2. build a PoolCandidatesTable which uses an ApplicantFilter instead of PoolCandidateFilter (maybe modify the existing table? I think we want to stop using PoolCandidateFilter at all?)
3. use the PoolCandidatesTable here instead of UserTable.

See these two comments:
- https://github.com/GCTC-NTGC/gc-digital-talent/pull/4623#issuecomment-1307662346
- https://github.com/GCTC-NTGC/gc-digital-talent/pull/4623#issuecomment-1307859883

## Acceptance Criteria
- [ ] SingleSearchRequest page shows a table which matches the table on the PoolCandidates page.
- [ ] The pool candidates shown should follow this logic:
    - find all applicants who meet the requirements of the ApplicantFilter
    - get all PoolCandidates for those users
    - if ApplicantFilter includes a poolId filter, filter the PoolCandidates by those pools
    - Note: These steps are conceptual. the filter algorithm probably shouldn't work exactly like that, but it should return the same results as if it did.
- [ ] Any changes/additions to the PoolCandidates model filtering functionality should be tested in phpunit

yonikid15:
Eventually will resolve #4672 

(UPDATED)
## Notes
- Uses the combined efforts of @vd1992 and @tristan-orourke.
- Adds a new input `PoolCandidateSearchInput` for the `poolCandidatesPaginated` query.
```
input PoolCandidateSearchInput {
  applicantFilter: ApplicantFilterInput @spread
  priorityWeight: [Int] @scope(name: "priorityWeight")
  status: [PoolCandidateStatus] @scope(name: "poolCandidateStatuses")
  generalSearch: string
  name: string
  email: string
}
```
- Refactors the `ApplicantFilterInput` and ` PoolCandidateFilterInput` to use the scope directive instead of the builder. A scope directive will cause Lighthouse to look for the appropriatly named scope method on whatever Class its returning, in this case PoolCandidate.
- Adds new test page `PoolCandidateSearchTest.php` 

- @tristan-orourke I think we should ask DCMO what filters they want in the `PoolCandidateFilterDialog` before adding this.
- Question: Should we rename the `PoolCandidateSearchFilter` to something more accurate?

## Testing 
- The best way to test this is to go to the [Search page](http://localhost:8000/en/search).
- Create a new request.
- Go to [admin talent requests page](http://localhost:8000/en/admin/talent-requests) and view our recently created request.
- Ensure number of candidates match total requested.
- Ensure when clicking on Filters dialog that all `initialFilters` are set.


tristan-orourke:
After looking at this and @vd1992's experimental version #5061, here are some thoughts:

**1.**
Vachan's approach of creating a new input type with the same attributes but different directives will work great for the frontend, even if giving us more to maintain on the backend. Seems like a good fallback solution if we can't find something better.

**2.** 
One improvement might be to convert our `filterBy` methods to [scopes](https://laravel.com/docs/8.x/eloquent#local-scopes), which will just require renaming them I think. My understanding is that a scope directive will cause Lighthouse to look for the appropriatly named scope method on whatever Class its returning. As long as PoolCandidate.php and User.php have the same scopes, and ApplicantFilterInput uses all scope directives, it may be able to work on both.

**3.**
There's the problem of how to add additional parameters to this PoolCandidate search, parameters which aren't in AttributeFilter. eg status and priorityWeight. We don't want to add them the ApplicantFilterInput object directly, because we want admins to be able to use them in this search, but for those fields to not be accessible on the public search page. 

I previously solved a problem like this in UserFilterInput by including ApplicantFilterInput as a first field, and then adding any extra fields that only Admins have access to. Here it would look something like:
```
input PoolCandidateSearchInput {
  applicantFilter: ApplicantFilterInput @spread
  priorityWeight: [Int] @scope(name: "priorityWeight")
  status: [PoolCandidateStatus] @scope(name: "poolCandidateStatuses")
}
```
But I'm not sure now if that was the best approach. Maybe it would be better to simply create a brand new Input type, as @vd1992 did in #5061, and redefine all the values we want in the search, and allow the frontend to convert ApplicantFilter to our new PoolCandidateFilterInput as necessary. That's the most flexible, and maybe conforms better to the single-responsibility principle, though its less DRY as a downside...

Sigh... this really makes me want to take another look at https://github.com/GCTC-NTGC/gc-digital-talent/issues/2365 and try to add inheritance/extension to GraphQL types.

vd1992:
request table could still use some tweaks

submitting a request, 10 candidates, viewing that request, 29 candidates
likely due to not filtering for the two statuses (qualified available and placed casual)

image

selecting the two statuses in the filter dialog wipes the pool selection and returns candidates from all pools

image

vd1992 (changes requested):
- default filter for placed casual + qualified available for request table
- need to ensure candidate count matches on request submission and viewing request
- changes filter on request table shouldn't wipe pool selection

vd1992:
same problem exists on Pool Candidate table
adjusting a filter wipes the pool selection

image

tristan-orourke:
> need to ensure candidate count matches on request submission and viewing request

I just want to point out that this may not necessarily be the case if there any skill filters. Skill filters are ignored on the public search page when getting the count, but should be taken into consideration here. Admins can always remove the skills from the filter dialog.

api/app/Models/User.php
```diff
@@ -234,7 +238,7 @@ public static function filterByPools(Builder $query, ?array $poolFilters): Build
                         return function ($query) use ($filter) {
                             $query->where('pool_candidates.pool_id', $filter['poolId']);
                             $query->where(function ($query) use ($filter) {
-                                if ($filter['expiryStatus'] == ApiEnums::CANDIDATE_EXPIRY_FILTER_ACTIVE) {
+                                if (array_key_exists('statuses', $filter) && $filter['expiryStatus'] == ApiEnums::CANDIDATE_EXPIRY_FILTER_ACTIVE) {
```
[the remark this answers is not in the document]
yonikid15: Yes 👍

frontend/admin/src/js/components/poolCandidate/PoolCandidateTableFilterDialog.tsx
```diff
@@ -120,6 +121,18 @@ const PoolCandidateTableFilterDialog = ({
...
+                id="pools"
+                label={formatMessage({
+                  defaultMessage: "Pools",
+                  id: "mjyHeP",
+                })}
+                options={optionsData.pools}
+                isLoading={rawGraphqlResults.pools.fetching}
+                isDisabled
+              />
+            </div>
```
tristan-orourke, line 135: I'm happy this selection is added in - I'm not actually sure it should be disabled. Unless there's a technical reason, why should admins be barred from changing this?

api/app/Models/PoolCandidate.php
```diff
@@ -87,166 +87,109 @@ public function scopeClassifications(Builder $query, ?array $classifications): B
...
-            $query->select('id')
-            ->from('users')
-            ->whereColumn('users.id', 'pool_candidates.user_id')
-            ->whereJsonContains('accepted_operational_requirements', $operationalRequirements);
+        // point at filter on User
+        $query->whereHas('user', function ($query) use ($workRegions) {
+            User::scopeLocationPreferences($query, $workRegions);
         });
 
         return $query;
```
tristan-orourke, line 101: Copy paste error? (workRegions instead of operationRequirements)

api/app/Models/PoolCandidate.php
```diff
@@ -87,166 +87,109 @@ public function scopeClassifications(Builder $query, ?array $classifications): B
...
-        if (empty($pools)) {
+        if (empty($poolIds)) {
             return $query;
         }
 
         // Pool acts as an OR filter. The query should return candidates in ANY of the pools.
-        $poolIds = [];
-        foreach ($pools as $pool) {
-            array_push($poolIds, $pool['id']);
-        }
```
tristan-orourke, line 155: If you look at the schema, the data passed to filterByPools is in the format `{id: ID!}[]`, meaning some kind of step like this is required to flatten it into an array of ids instead of an array of objects which hold ids.

api/app/Models/PoolCandidate.php
yonikid15: Thanks!

api/app/Models/PoolCandidate.php
yonikid15: Hmm, this must of been an accident 😆 I don't remember removing this.

frontend/admin/src/js/components/poolCandidate/PoolCandidateTableFilterDialog.tsx
yonikid15: My initial thought was to eliminate the possibility of the admin from accidentally changing the pool and getting candidates not associated with what they are looking for. However, now I think your right it's probably better if they have that freedom to change it.

yonikid15: @tristan-orourke maybe we should ask DCMO what filters they would like to see on the PoolCandidateTableFilterDialog before deploying on Thursday?

vd1992: this will definitely be something to keep an eye on for Organizations, preventing managers from viewing candidates of pools they shouldn't have access to

vd1992:
the e2e test is probably failing for the same reason as below
image

vd1992:
solutions:

1)
transform this input to match the scope

image

2)
remove the `'id'` bit from the scope and ensure all queries that use that scope are adjusted accordingly

api/app/Models/PoolCandidate.php
yonikid15: Actually, I needed to update this since the ApplicantFilterInput uses the `@pluck(key: "id")` scope. I added it to the PoolCandidateInput scope as well.

yonikid15:
Okay, I believe I have got all of the requested changes from both of you. Please let me know if I have missed anything. 

> I just want to point out that this may not necessarily be the case if there any skill filters. Skill filters are ignored on the public search page when getting the count, but should be taken into consideration here. Admins can always remove the skills from the filter dialog.

Currently, there is no skills filter in the filter dialog. Should I add one in?

tristan-orourke:
I do think we'll need to add a filter for Skills, if not now then soon.

vd1992 (changes requested):
I'm going to have to disagree, since adding skills to the request doesn't impact creating a request, there is nothing stopping someone from stacking lots of skills in their request. So, the filter dialog will be needed really, really quickly. 

Like so

image

without the ability to remove skill filters, an admin is therefore unable to do anything with the request  below

image

api/graphql/schema.graphql
```diff
@@ -758,63 +758,53 @@ input UserPoolFilterInput {
...
-    @builder(method: "App\\Models\\PoolCandidate@filterByEquity")
-  expectedClassifications: [ClassificationFilterInput]
-    @scope(name: "classifications")
-  generalSearch: String
-    @builder(method: "App\\Models\\PoolCandidate@filterByGeneralSearch")
+  equity: EquitySelectionsInput @scope
+  expectedClassifications: [ClassificationFilterInput] @scope(name: "classifications")
   email: String @scope
-  name: String @builder(method: "App\\Models\\PoolCandidate@filterByName")
+  generalSearch: String @scope
```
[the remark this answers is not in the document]
tristan-orourke: I'll make a separate ticket for this. Looks like the only queries which use PoolCandidateFilterInput are either deprecated or could/should be deprecated, so there shouldn't be any problem with removing these fields.

tristan-orourke (approved):
Looks good to me now! I found a couple things I'll spin off into new issues, but I think we can go ahead with this PR.

tristan-orourke:
> @vd1992: I'm going to have to disagree, since adding skills to the request doesn't impact creating a request, there is nothing stopping someone from stacking lots of skills in their request. So, the filter dialog will be needed really, really quickly.

Since the PoolCandidateTable filter dialog lacks skill filter inputs, the skills in the request are ignored! It works for the moment, since the skills are also ignored in the frontend, and Admins can still see the skills in the Summary of Filters. But yes, we'll want to add them in ASAP.

ps @vd1992 I think your table is empty in your example because the languages filter is bugged, not because of skills. Try removing the language filter and seeing if you see any results.

vd1992:
Uh, I don't think that appears to be the case 😕 

image

api/graphql/schema.graphql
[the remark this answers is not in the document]
yonikid15: Okay sounds good thanks Tristan

yonikid15:
> Uh, I don't think that appears to be the case 😕

Ya @vd1992 is correct, the skills are being applied.. I can either add the filter in which shouldn't be to hard, or I can temporary remove the skills from the applicantFilter before getting passed into the dialog?

tristan-orourke:
Ah, I see. Skills are included in the initial request, but are lost after you change any other filters. Okay, lets add in the skills.

yonikid15:
Alright, I added in the skills filter [fc8c892](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5052/commits/fc8c892baa7fd2af989c98d9e0f4b513211838a7). Should be good to go now 👍

vd1992 (approved):

vd1992:
think that is all I got
awesome! this was definitely a tricky one

PR #5056, opened 2022-12-08
Feature - Add rate limits for api, graphql

### LINKED ISSUE #4185: Backend Rate Limiting
In theory it would be possible for a bad actor to flood our backend and bring it down and/or cost us a lot of money.  Do we have a way of rate limiting excessive usage?  With graphql this might be a bit tricky since not all requests are comparably costly.

## Acceptance Criteria
- [ ] Any API request trying to hit our endpoint gets throttled after 100 requests in a minute.
- [ ] Error message that users gets in a toast after they are throttled is:
> Too many requests, please wait a minute and try again.
> Trop de demandes, veuillez patienter une minute et réessayer.

mnigh:
Resolves #4185.

h/t to @esizer.

## Notes
Since we only have one "api" endpoint right now in `api/support/tickets, 100 requests per minute should not be a problem. However, since almost every page for **admin** and **talentsearch** run one or more GraphQL queries, 100 requests per minute for "graphql" might be too low a number.

- **Update 1**: default request limit set to 600.
- **Update 2**: added a throttle on the `createUser` GraphQL mutation of 100 requests per minute for future reference (and, since no one should be creating more than one user per second!)

## Screenshots
Screen Shot 2022-12-08 at 14 32 44
Screen Shot 2022-12-08 at 14 34 06
Screen Shot 2022-12-08 at 14 37 16
Screen Shot 2022-12-08 at 15 27 09

## Steps to test

1. Compile common `npm run intl-compile --workspace=common`
2. Build workspaces `npm run production --workspaces --if-present`
3. Add [`APP_RATE_LIMIT=10`] in (api/.env)
4. Change GraphQL me query in schema.graphql from `me: User @auth` to `me: User @auth @throttle(maxAttempts: 1)`
5. Refresh API (refresh_api.sh)
6. Navigate to /support
7. Verify singular toast appears in English with correct text
8. Switch to French
9. Verify singular toast appears in French with correct text
10. Navigate to /search
11. Verify singular toast appears in English with correct text
12. Switch to French
13. Verify singular toast appears in French with correct text
14. Navigate to /admin/dashboard
15. Verify singular toast appears in English with correct text
16. Switch to French
17. Verify singular toast appears in French with correct text
18. Navigate to /login
19. Verify Too Many Requests text


api/app/Providers/RouteServiceProvider.php
```diff
@@ -57,7 +57,14 @@ public function boot()
     protected function configureRateLimiting()
     {
         RateLimiter::for('api', function (Request $request) {
-            return Limit::perMinute(60)->by(optional($request->user())->id ?: $request->ip());
+            return Limit::perMinute(100)->by(optional($request->user())->id ?: $request->ip())->response(function () {
```
petertgiles, line 60: I agree that 100 is probably too low.  600 maybe?  Since this is in multiple places and we probably will want to tune it could it be factored out to a config and added to the .env file?

api/app/Providers/RouteServiceProvider.php
```diff
@@ -57,7 +57,14 @@ public function boot()
     protected function configureRateLimiting()
     {
         RateLimiter::for('api', function (Request $request) {
```
petertgiles, line 59: Our auth end points are in the web.php routes.  Does a limiter need to be added for them, too?

petertgiles (changes requested):
This is great!  I love the frontend work but I've got a few questions about the backend changes.

api/app/Providers/RouteServiceProvider.php
mnigh: Sure, 600 for both api and graphql, or 600 for graphql and api remains at 100? Should the Acceptance Criteria be updated for the attached issue as well?

I like the idea of environment variables. What do you think of:

- RATE_LIMIT_GRAPHQL
- RATE_LIMIT_API

api/app/Providers/RouteServiceProvider.php
mnigh: Good catch, I hadn't considered auth. What should the limits be for the web group?

I like the idea of environment variables from the other comment. What do you think of:

- RATE_LIMIT_WEB

api/app/Providers/RouteServiceProvider.php
```diff
@@ -57,7 +57,21 @@ public function boot()
...
+            });
+        });
+        RateLimiter::for('web', function (Request $request) {
+            return Limit::perMinute(config('app.rate_limit'))->by(optional($request->user())->id ?: $request->ip())->response(function () {
+                return response([
+                    'message' => 'Rate Limit Reached for Web'
+                ], 429);
+            });
+        });
+        RateLimiter::for('graphql', function (Request $request) {
```
petertgiles, line 73: Suggest a comment here to clarify that this limiter is for the throttle directive which can be set independently of the of the route-based limiters and can raise graphql style error messages.

petertgiles (approved):
Love it!  I'd probably combine the web and api limiters but probably the better thing would be to combine the two route groups anyway. :laughing:  This is great for now.  I left a few small suggestions for comments below.

petertgiles (approved):
:+1:

PR #5097, opened 2022-12-12
Bug - 4980 - Prevent unsaved changes while editing

esizer:
## 👋 Introduction

This updated the `<BasicForm />` to only show the `<UnsavedChanges />` alert on the initial render if it has unsaved changes by using internal `useState`. It also allows users to dismiss the alert if they wish.

## 🧪 Testing

1. Build talent search `npm run production --workspace=talentsearch`
2. Navigate to a form on the profile
3. Make some changes to the form
4. Confirm the alert at the top of the form does not appear
5. Refresh the page
6. Confirm the alert appears and can be dismissed
7. Dismiss the alert and make more changes
8. Confirm the alert does not reappear

## 🤖 Robot Stuff

Resolves #4980

PR #5098, opened 2022-12-12
Bug - 4809 - Prevent error summary while editing

### LINKED ISSUE #4809: Error message displays before form submission
## Description

The error message currently displays if there are any empty required fields BEFORE the form is submitted.  This causes confusion from a screen reader perspective since the placement of the field is at the top of the page, which the reader is unaware of (and it is constantly changing while the user is working).

## Wanted behaviour
The error message for required fields should only be displayed AFTER the form is submitted.  This allows the screen reader user to see a list of the fields that need to be fixed at the top of the page. 

 After the user attempts to save & the page loads, navigate the user to the top of the "Oops you have some errors" message (focus), so that the error message text is the first thing that is read.

See [WCAG Advisory technique G139](https://www.w3.org/WAI/WCAG21/Techniques/general/G139) for more information on this issue. Notice that, in all examples, the error is only shown after form submission.

## Screenshot
Capture

## Acceptance Criteria

- [x] Only show the big alert at the top of the page after the user hits the submit button. NOT when they're on the page making changes
- [x] Keep in-line errors contained to ONLY that specific field when the user made an error and moves away and not everything on the page at once if the user has errors before hitting the submit button.
- [x] Verify everything is marked all at once AFTER the user clicked submit and has many errors.

esizer:
## 👋 Introduction

This uses the same logic in #5097 to make the `<ErrorSummary />` only appear after a user attempts to submit a form. It also adds `role="alert"` back to `<InputError />` to announce errors to a user as they type (this was removed to prevent duplicate announcements when summary was also appearing while editing).

## 🧪 Testing 

1. Build talent search npm run production --workspace=talentsearch
2. Navigate to a form on the profile
3. Make some changes to the form
4. Confirm the alert at the top of the form does not appear
5. Attempt to submit the form with errors
6. Confirm the alert appears

## 🤖 Robot Stuff

Resolves #4809

### DIFF
--- a/frontend/common/src/components/form/BasicForm.tsx
+++ b/frontend/common/src/components/form/BasicForm.tsx
@@ -42,6 +42,8 @@ export function BasicForm<TFieldValues extends FieldValues>({
   cacheKey,
   labels,
 }: BasicFormProps<TFieldValues>): ReactElement {
+  const [showErrorSummary, setShowErrorSummary] =
+    React.useState<boolean>(false);
   const errorSummaryRef = React.useRef<HTMLDivElement>(null);
   const methods = useForm({
     mode: "onChange",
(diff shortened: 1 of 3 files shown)


mnigh (approved):
1. Built talent search
2. Navigated to a form on the profile (About me)
3. Made some changes to the form
4. Confirmed the alert at the top of the form does not appear
5. Attempted to submit the form with errors
6. Confirmed the alert appears

👍

Issue #4916, opened 2022-11-24
♻ Implement PKCE auth using `react-oidc` library


tristan-orourke:
# Description
In our current authentication strategy, the client makes a call to our API, which communicates with an OAuth 2.0 identity provider, exchanging a client secret for access tokens, and then passes tokens back to the client using url query variables.

We want to replace this with an OAuth authentication flow using PKCE, where the client can request tokens directly from the identity provider. To make this refactor easier, we want to use a frontend library to handle the auth flow in the background. 

@patcon previously researched auth module options in #2572, and recommended https://github.com/AxaGuilDEv/react-oidc. We have chosen this library because it has good active support, and unlike other options, it protects against XSRF attacks by storing tokens in a service worker instead of localStorage or a cookie.

The library suggest that the service worker can inject tokens directly into fetch requests, which may make it unnecessary to use auth-exchange in urql. That probably requires testing.

## Acceptance Criteria
- [ ] Authentication is handled through the https://github.com/AxaGuilDEv/react-oidc library
- [ ] If possible, simplify our urql client provider by avoiding use of auth-exchange
- [ ] User experience of logging in and out is unchanged.
- [ ] We can continue to use React Router to specify that certain routes require being authenticated with certain roles
- [ ] Refreshing access tokens continues to happen silently in the background
- [ ] Using the app in multiple tabs does not cause any issues (as in #4523) 
- [ ] Necessary communications with Sign-In Canada to keep auth working

petertgiles:
Currently blocked on localhost issues.
- Requests to mockauth go to different hostnames between the api (inside the docker bridge network) and the browser (outside of docker).  This fails out the key validation.  This might be solvable with mockauth configuration.
- Requests for the configuration data from the .well-known endpoints now come from the browser and are subject to CORS protections.  It should be possible to add CORS headers to mockauth but I haven't figured it out yet.  SIC should be able to add the headers, too.
- Chrome rejects CORS requests from localhost always.
- I tried a browser extension that hacks on the CORS headers but the SIC captcha is rejecting it.

brindasasi:
> Currently blocked on localhost issues.
> 
> * Requests to mockauth go to different hostnames between the api (inside the docker bridge network) and the browser (outside of docker).  This fails out the key validation.  This might be solvable with mockauth configuration.
[...]

PR #5100, opened 2022-12-12
Improve session-timeout logout experience

### LINKED ISSUE #4295: User redirected to logged-out page unnecessarily
When page has been opened long enough that auth token has gone stale, following any link forcibly redirects user to the logged-out page, instead of where they wanted to go.

a) If user was going to a page which doesn't require authentication, this redirect is disruptive and unnecessary.
b) If user was going to a page which does require being authenticated, it would be more convenient to just redirect directly to the login page, and bring them back to the intended url after.

In either case, a toast describing the situation would be preferable to a redirect to the logged-out page.

## Acceptance Criteria
- When the auth token has gone stale, and the user is going to a page which **DOESN'T require authentication**
  - [ ] **continue** to the page instead of logging out and
  - [ ] **display a toast** saying user has logged out.
- When the auth token has gone stale, and the user is going to a page which **DOES require authentication**
  - [ ] **redirect to login** and
  - [ ] include the intended page as the **post-login redirect**
  - [ ] do NOT show a toast

tristan-orourke:
Resolves #4295.

~NOT YET FULLY WORKING.~
It's working now! 👍 

Steps to test:
1. Reduce the expiry timeout in mock-ouath config json and then restart docker container. I recommend setting it to 15 seconds.

a) Test restricted page with browser refresh

1. Log in and go to a page that requires being authorized as a particular user, like Profile page.
2. In the browser inspector, go to the Application tab, open localStorage, and delete the refresh token. (This will ensure that token refresh fails when the access token expires.)
3. Wait long enough for the access token to expire.
4. Refresh the browser page. You should be redirected to the mock-auth login page, then sent back to the profile page when login is successful.

b) Test restricted page with internal routing. 
1. Log in and go to a page that requires being authorized as a particular user, like Profile page.
2. In the browser inspector, go to the Application tab, open localStorage, and delete the refresh token. (This will ensure that token refresh fails when the access token expires.)
3. Wait long enough for the access token to expire.
4. Click a link to another page which requires being logged in (like the My Applications page). You should be redirected to the mock-auth login page, then sent back to the My Applications page which you were trying to navigate to.

c) Test public page with browser refresh
1. Log in and go to a page that does NOT require being authorized as a particular user, like home page.
2. In the browser inspector, go to the Application tab, open localStorage, and delete the refresh token. (This will ensure that token refresh fails when the access token expires.)
3. Wait long enough for the access token to expire.
4. Refresh the browser page. You should be logged out (indicate by changes to the nav bar) and then land back on the page you were on.

d) Test public page with internal routing
1. Log in and go to a page that does NOT require being authorized as a particular user, like home page.
2. In the browser inspector, go to the Application tab, open localStorage, and delete the refresh token. (This will ensure that token refresh fails when the access token expires.)
3. Wait long enough for the access token to expire.
4. Click a link to another page which does not require being logged in (like the Search page). You should be logged out (indicate by changes to the nav bar) and then land back on the page you were trying to reach.

- [x] Test a) passes
- [x] Test b) passes
- [x] Test c) passes
- [x] Test d) passes
- [x] Test a) passes when starting from an admin page
- [x] test b) passes when starting from an admin page

TODO:
~Right now, expected behaviour happens when you refresh the browser. However, if you trigger an api call without refreshing the browser after the access token has expired (ie tests b) and d)) then you get 401 unauthorized responses from the API, and console errors in the browser, but you are never actually logged out. I'm not sure why. Furthermore, if you reload the browser AFTER getting these errors, you still get the errors instead of proper behaviour.~
Resolved this issue! It turned out that an error in refreshTokenSet was causing code to crash before expected code could execute. This is fixed in acfcfab99e01e738b459376000f7a3f668161654.

QUESTION:
Oh, what should happen if you are on a restricted page (either the profile, or an admin), your session expires, you get redirected to logout, and then you log in as a different user (who's not an admin)?

A: The default behaviour on admin makes sense - you ge an error notice saying you're not authorized to view this page (because you don't have the correct role). I found a bit of an odd case on the profile. After logging as someone new, you're redirected to the previous user's profile, it seems to work just fine! Upon inspection, you'll notice the date is all from the new user, even though the url reflects the previous user's id, because we just use `getMe` graphql queries. I'll record this as a bug in another issue.


tristan-orourke:
Resolved the main issue, and this is now working correctly! Leaving it as a draft because I still want to do a last bit of code cleanup.

tristan-orourke:
Okay, I think this is worth reviewing now. Here's how it works:

If your access token is stale, and refreshing doesn't work (probably because your refresh token is also stale - that means 4 hours afk) the AuthenticationContainer stores your current url in session storage, then triggers a logout. Logout may or may not send you to the SiC logout uri, and then sends you to our /logged-out page. Now the logged-out page checks for a stored url in session storage, and redirects you again if it exists.

It feels a little clumsy to me, and can be a little slow, since it involves a number of hard refreshes, and because the react app must load for the /logged-out page before the final redirect happens. However, I'm not sure how to improve this. I considered moving the redirect logic to the backend (like our /login api route), but we're already interested in moving the auth functionality more towards the frontend exclusively (https://github.com/GCTC-NTGC/gc-digital-talent/issues/4916).

petertgiles:
> It feels a little clumsy to me, and can be a little slow

I agree that this is worth revisiting after the auth library change since it might look quite different after that.  I believe there is an auth logout option that uses a hidden iframe rather than navigation there and back.  That would remove the navigation to SiC, the one back to logged-out, and the one back to the original location, too.

petertgiles:
> we just use `getMe` graphql queries. I'll record this as a bug in another issue.

Yeah, using `getMe` everywhere is not great, and now there's another reason to avoid it!  Should probably be relying on our context providers for this instead.

frontend/common/src/components/Auth/AuthenticationContainer.tsx
```diff
@@ -49,6 +52,16 @@ const logoutAndRefreshPage = (
   localStorage.removeItem(REFRESH_TOKEN);
   localStorage.removeItem(ID_TOKEN);
 
+  if (postLogoutUri) {
+    if (!postLogoutUri.startsWith("/")) {
+      defaultLogger.warning(
+        `Tried to set an unsafe uri as postLogoutUri: ${postLogoutUri}`,
+      );
+    } else {
+      sessionStorage.setItem(POST_LOGOUT_URI, postLogoutUri);
```
petertgiles, line 61: Nice!  Storing it in sessionStorage should help use with the two "go to login-info first" issues we put in the backlog yesterday!

petertgiles (changes requested):
This is working great for me!  It should really improve quality of life, I think.

Would you consider adding cypress tests to ensure we don't lose this in the upcoming auth change?  It looks like it is possible to manipulate localStorage from Cypress.  https://stackoverflow.com/a/50820733

I could be convinced to spin it off into another issue to avoid scope creep.  :smile:

tristan-orourke:
> > we just use `getMe` graphql queries. I'll record this as a bug in another issue.
> 
> Yeah, using `getMe` everywhere is not great, and now there's another reason to avoid it! Should probably be relying on our context providers for this instead.

Haha, I had the opposite response - I though `getMe` makes sense and we should remove the id from the url 😆. I've mentioned both options for discussion in a ticket https://github.com/GCTC-NTGC/gc-digital-talent/issues/5413.

petertgiles:
Will spin off an issue for testing.

petertgiles (approved):

PR #5055, opened 2022-12-08
Debt/2218 pool candidate obsolete backend

vd1992:
closes #2218 
- drops the content that was migrated by the `MigratePoolCandidatesToUser` command to remove duplication
- removes the fields that reference dropped content from front to back
- PoolCandidate.php tweaks 
        -> filterByCmoAssets updated to use the relations on user instead
        -> filterByClassificationToSalary updated to use user model instead of pool candidate (vital as profiles are live)

### TESTING
will need to be cautious with this one but finally some 🧹 

**DATABASE**
- test migration
- observe it works and rollbacks as it ought to

**BACKEND**
- model logic checks out
- schema changes, everything that should be removed was removed and nothing more

**FRONTEND**
- nothing should have really changed frontend wise, aside from the to-be-deprecated PoolCandidate create and update forms
- assert everything works as it should (talentsearch, poolcandidate tables, request creation, table filtering...)

Issue #4222, opened 2022-10-06
Missing PHP linter in GitHub workflow


mnigh:
## Description
The GitHub workflow for `[linting](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/.github/workflows/lint.yml)` does not run a linter for PHP even though our API is written in PHP.

## Possible Solution
Replace current workflow with [Super-Linter](https://github.com/marketplace/actions/super-linter) GitHub action.

patcon:
No block on super-linter, but I wonder if it might be simpler to just add a step/workflow for PHPStan rather than an omni-linter docker image with every linter installed, which might not be clear how to run locally

https://phpstan.org/

(when super-linter came out, it kinda struck me as a bit of an idea that come more from marketing team for GitHub Actions. The 8k stars for a GitHub-backed project that applies to ALL LANGUAGES aka every repo, that kinda corroborates for me that the developer response was a bit tepid)

vd1992:
related issue #3950

petertgiles:
I've never used any of these tools but I think PHPStan is more of an error checker than a linter.  It does sound like something we would want, too, though.

patcon:
bah, i had a huge comment here where I'd done the research, but lost it in The Great Chrome Crash of last night 😭 

Was basically summarizing these articles/posts:
- https://phpstan.org/blog/three-pillars-of-static-analysis-in-php
- https://www.reddit.com/r/PHP/comments/cv8cc3/why_did_you_choose_phpstan_or_psalm/

- rapid re-summary
  - @petertgiles is right about linters vs error checkers being diff
  - someone in comments suggested it's easier to just run all the comparable static analysis tools, rather than try to decide in advance what's best. if one gets annoying or never seems to be the one to raise issues, it can be removed later.
  - linting: https://github.com/JakubOnderka/PHP-Parallel-Lint is a good re-surface of php's native linter
  - Psalm and PhpStan are the error-checkers
    - Psalm is the new kid on the block. Seems ppl like it when they try. more features? but also maybe aggressive defaults that scare ppl, so dial them back
      - https://github.com/vimeo/psalm
    - PhpStan is the more mature one, with momentum and very few "problems" that drive anyone away
      - https://github.com/phpstan/phpstan
  - php codesniffer is more about codestyle: https://github.com/squizlabs/PHP_CodeSniffer
    - existing company/project-specific sniffs to choose from

vd1992:
[...]

PR #5101, opened 2022-12-12
Bug/5073 language ability

### LINKED ISSUE #5073: Admin - filtering users table by "English Only" filters out candidates with preference for English
I viewed a request on Prod. I modified the filters so the only filters were:
- pools: IT-04 software solutions
- Profile Complete: Yes
- Languages: English only

This resulted in zero users appearing in the table. When I removed the Languages filter, three users appeared. However, the Preferred language was English for all three of them. What is going on?

It didn't matter if I selected English only, French only, or Bilingual - any language filter at all would filter out all users.

Also note: when I remove the Profile Complete filter but leave the Languages filter, I see a number of admin_entry users. That means the Languages filter is only filtering out users who have completed their own profiles.

vd1992:
closes #5073 
`language_ability` is a field that really isn't used and was tripping up language filtering as it is not set anywhere on `/profile` side or for that matter anywhere on the frontend

- drop the field in a migration, conditionally update the `looking_for_X` set
- update model file `filterByLanguage` accordingly
- no frontend changes appeared needed as it appears the field truly was obsolete

Testing: 
- test the migration, forwards and back
- assert updating your language info isn't broken
- test submitting a request and viewing it admin-side while using language filtering
- test filtering pool candidate and/or user tables with language


api/graphql/schema.graphql
```diff
@@ -54,7 +54,9 @@ type User {
   armedForcesStatus: ArmedForcesStatus @rename(attribute: "armed_forces_status")
 
   # Language
-  languageAbility: LanguageAbility @rename(attribute: "language_ability")
+  languageAbility: LanguageAbility
+    @rename(attribute: "language_ability")
+    @deprecated(reason: "obsolete field")
```
tristan-orourke, line 59: Please add a more detailed/informative reason.

api/tests/Feature/ApplicantFilterTest.php
```diff
@@ -404,7 +403,7 @@ public function testFilterCanBeStoredAndRetrievedWithoutChangingResults()
                 'is_visible_minority' => $candidate->user->is_visible_minority,
                 'is_woman' => $candidate->user->is_woman,
                 'position_duration' => $candidate->user->position_duration,
-                'language_ability' => $candidate->user->language_ability,
+                'language_ability' => null, // user->language_ability is a computed field
```
tristan-orourke, line 406: You can delete this line.

api/app/Models/PoolCandidate.php
```diff
@@ -238,27 +238,22 @@ public function filterByLocationPreferences(Builder $query, ?array $workRegions)
         });
         return $query;
     }
+
```
tristan-orourke, line 241: The languageAbility field still exists on PoolCandidate in the schema (even if deprecated) so PoolCandidate.php should also have an accessor for it. It can reference the User accessor.

tristan-orourke (changes requested):
Looks pretty good, just a few suggestions.

api/app/Models/PoolCandidate.php
```diff
@@ -238,27 +238,22 @@ public function filterByLocationPreferences(Builder $query, ?array $workRegions)
...
-            $query->select('id')
-            ->from('users')
-            ->whereColumn('users.id', 'pool_candidates.user_id')
-            ->where(function ($query) use ($languageAbility) {
-                $query->where('language_ability', $languageAbility);
-                if ($languageAbility == ApiEnums::LANGUAGE_ABILITY_ENGLISH || $languageAbility == ApiEnums::LANGUAGE_ABILITY_FRENCH) {
-                    $query->orWhere('language_ability', ApiEnums::LANGUAGE_ABILITY_BILINGUAL);
-                }
+        // point at filter on User
+        $query->where(function ($query) use ($languageAbility) {
```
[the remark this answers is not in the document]
vd1992: I think it might be
https://github.com/GCTC-NTGC/gc-digital-talent/pull/5101/commits/59844071df205a570d4bb936103d0e2472ba2c84

api/graphql/schema.graphql
vd1992: aye aye
https://github.com/GCTC-NTGC/gc-digital-talent/pull/5101/commits/59844071df205a570d4bb936103d0e2472ba2c84

api/tests/Feature/ApplicantFilterTest.php
vd1992: no, I don't think I can without risking the test failing sometimes
the block as seen below attempts to match the filter to `$candidate`
what the accessor returns might not always be the appropriate result to filter with, so, to be safe and rather than randomly generate it in `factory()` I force it to null

the immediate problem case I can see is
```
lookingForEnglish = true
lookingForFrench = true
lookingForBilingual = false
```

in that case the accessor would assign BILINGUAL as best fit, but filtering for it wouldn't return the candidate

image

api/app/Models/PoolCandidate.php
vd1992: I didn't deprecate it on `PoolCandidate` though and didn't drop it from the pool candidate table
reason being the field is being taken care of in #5055 

image

tristan-orourke: Ah, that's right, you didn't actually drop it from the pool candidate table. Carry on. 😁

api/tests/Feature/ApplicantFilterTest.php
tristan-orourke: Hmmm... I see what you're saying, but this seems like a hack. I feel like there must be a better solution.

tristan-orourke: Actually, wait. This is on ApplicantFilter, which _does_ still have the language_ability column in the database, and it needs to still work when using it to filter. We can't delete this line - we shouldn't modify it at all, in fact.

tristan-orourke (changes requested):
Hmm. Actually running the migration results in the `looking_for_<language>` columns remaining null if they were before, even if `language_ability` was set. Not sure why yet 😕

vd1992:
> Hmm. Actually running the migration results in the `looking_for_<language>` columns remaining null if they were before, even if `language_ability` was set. Not sure why yet 😕

that's strange
time to 🔍 🕵️

vd1992:
can't seem to replicate 😕 

using this to dump out changes
image

results in this logged

image

vd1992:
Nevermind!
Bug found (accessor issue) in https://github.com/GCTC-NTGC/gc-digital-talent/pull/5101/commits/87c505a8aecee3c77ef57780b2ecfc301c181395

api/app/Models/User.php
```diff
@@ -667,4 +673,34 @@ public function getIsIndigenousAttribute() {
 
         return null; // if indigenousCommunities is null then so is isIndigenous
     }
+
+    /* accessor to maintain functionality of to be deprecated languageAbility field, its logic comes from migration drop_language_ability*/
+    public function getLanguageAbilityAttribute($languageAbility = null) {
+        // if the field exists, say for migration purposes, must stop accessor overriding
+        if($languageAbility !== null){
+            return $languageAbility;
+        }
```
tristan-orourke, line 682: Ahhh, I see. Actually, this highlights why its generally a bad idea to rely on the Eloquent class in a database migration. The class can change in ways that doesn't reflect the state of the database.

tristan-orourke (commented):
Hmm, how hard would it be to rewrite the migration to just use database operations, and not reference the Users class?

vd1992:
> Hmm, how hard would it be to rewrite the migration to just use database operations, and not reference the Users class?

Not too tricky after all
https://github.com/GCTC-NTGC/gc-digital-talent/pull/5101/commits/0de55328e09f80047af3eaa489d4754c6b8bedc8

api/database/migrations/2022_12_09_230812_drop_language_ability_users.php
```diff
@@ -0,0 +1,178 @@
...
+                    if ($languageAbility == ApiEnums::LANGUAGE_ABILITY_ENGLISH | $languageAbility == ApiEnums::LANGUAGE_ABILITY_BILINGUAL) {
+                        DB::statement(
+                            <<<SQL
+                                UPDATE users
+                                SET looking_for_english = true
+                                WHERE id = :userId
+                            SQL, [
+                                'userId' => $userId,
+                            ]
+                        );
```
tristan-orourke, line 37: For simple statements like this I recommend using the [Laravel DB methods](https://laravel.com/docs/9.x/queries#update-statements) since they include built-in sanitization and stuff like that.
```suggestion
                        DB::table('users')->where('id', $userId)->update(['looking_for_english' => true]);
```

api/tests/Feature/ApplicantFilterTest.php
tristan-orourke: Okay, so the problem with this is that when using the `language_ability` field to filter, we actually check the `looking_for_<language>` fields on the User. In that case, when generating a filter to match a specific user, we can use the `looking_for_<language>` fields instead of `user->language_ability`.

tristan-orourke (changes requested):

api/database/migrations/2022_12_09_230812_drop_language_ability_users.php
vd1992: so many ways to do things, neat, thanks
done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/5101/commits/da8d2eda2ea250476cc76e23c404c91708c9759d

as well, noticed a bug that was resolved by https://github.com/GCTC-NTGC/gc-digital-talent/pull/5101/commits/6e8ad5cc315b12b9736fa310dd7d36937a008ce3

api/tests/Feature/ApplicantFilterTest.php
vd1992: alright, I'll try something else in https://github.com/GCTC-NTGC/gc-digital-talent/pull/5101/commits/149a33af5e32b2f44d8fb50c4bbbfcb97d71c27d

api/database/migrations/2022_12_09_230812_drop_language_ability_users.php
vd1992: per my latter mistake, an issue sorta relevant exists here #4222

api/database/migrations/2022_12_09_230812_drop_language_ability_users.php
```diff
@@ -0,0 +1,103 @@
...
+
+                // only override looking_for_english if it is null
+                if ($user->looking_for_english === null) {
+                    // play it safe, ENGLISH and BILINGUAL map to true
+                    if ($languageAbility == ApiEnums::LANGUAGE_ABILITY_ENGLISH | $languageAbility == ApiEnums::LANGUAGE_ABILITY_BILINGUAL) {
+                        DB::table('users')->where('id', $userId)->update(['looking_for_english' => true]);
+                    }
+                    else {
+                        DB::table('users')->where('id', $userId)->update(['looking_for_english' => false]);
+                    }
```
tristan-orourke, line 33: At this point, I think it makes sense to skip the if statement and just update the DB with a variable.
```suggestion
                    $lookingForEnglish = ($languageAbility == ApiEnums::LANGUAGE_ABILITY_ENGLISH || $languageAbility == ApiEnums::LANGUAGE_ABILITY_BILINGUAL);
                    DB::table('users')->where('id', $userId)->update(['looking_for_english' => $lookingForEnglish]);
```
ps. You probably want to use `||` instead of `|`. See [difference between bitwise and logical operators](https://stackoverflow.com/questions/2376348/difference-between-and-in-php)

api/tests/Feature/ApplicantFilterTest.php
```diff
@@ -393,9 +394,18 @@ public function testFilterCanBeStoredAndRetrievedWithoutChangingResults()
...
+        $filterLanguage = null; // run through fields and assign the enum for the first one that is true
+        if ($candidate->looking_for_english) {
+            $filterLanguage = ApiEnums::LANGUAGE_ABILITY_ENGLISH;
+        }
+        elseif ($candidate->looking_for_french) {
+            $filterLanguage = ApiEnums::LANGUAGE_ABILITY_FRENCH;
+        }
+        elseif ($candidate->looking_for_bilingual) {
+            $filterLanguage = ApiEnums::LANGUAGE_ABILITY_BILINGUAL;
+        }
```
tristan-orourke, line 408: 👍

tristan-orourke (changes requested):

tristan-orourke (approved):

PR #5135, opened 2022-12-15
Fixed 404 routing for admin and talentsearch

### LINKED ISSUE #4822: Fix router for 404
## Description
On requests that I would expect to throw a 404, they instead show the **Sorry, we encountered an error.** message.

## Notes
I was using their [FAQ documentation for 404](https://reactrouter.com/en/main/start/faq#how-do-i-add-a-no-match-404-route-in-react-router-v6) but looks like that is not necessary with our `errorElement`! 

We can remove the following and it should work:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/b8227c9a847868c569e84e765eb580ff9182a43e/frontend/talentsearch/src/js/components/Router.tsx#L341-L344

_Originally posted by @esizer in https://github.com/GCTC-NTGC/gc-digital-talent/issues/4815#issuecomment-1320382831_

JamesHuf:
Resolves #4822.

Fixes routing for handing 404s in admin and talentsearch.

To test:
1. Refresh the site
2. Navigate to a non-existent route (ex. `/en/oops`)
3. Confirm the 404 page displays as expected.
4. Navigate to a non-existent admin route (ex. `/en/admin/oops`)
5. Confirm the admin 404 page displays as expected.

Notes:
- We don't currently have an error page for the IAP stuff so currently navigation to non-existent pages there (ex. `/en/indigenous-it-apprentice/oops`) will result in the default react router error page.

### DIFF
--- a/frontend/admin/src/js/components/Router.tsx
+++ b/frontend/admin/src/js/components/Router.tsx
@@ -226,12 +226,12 @@ const SingleSearchRequestPage = React.lazy(() =>
 
 const router = createBrowserRouter([
   {
-    path: "/:locale",
+    path: `/`,
     element: <Layout />,
     errorElement: <ErrorPage />,
     children: [
       {
-        path: "admin",
+        path: ":locale/admin",
         errorElement: <ErrorPage />,
         children: [
           {
@@ -541,6 +541,12 @@ const router = createBrowserRouter([
                 ],
               },
             ],
+          },
+          {
+            path: "*",
+            loader: () => {
+              throw new Response("Not Found", { status: 404 });
+            },
           },
         ],
       },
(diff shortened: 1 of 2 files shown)


frontend/admin/src/js/components/Router.tsx
```diff
@@ -542,6 +542,12 @@ const router = createBrowserRouter([
               },
             ],
           },
+          {
+            path: "*",
+            loader: () => {
+              throw new Response("Not Found", { status: 404 });
+            },
+          },
```
esizer, line 550: ooo, I like this. Very clever 🕵️

esizer (approved):
Very clever approach! 👍 

---

> **Note**
> I wouldn't worry about IAP, I think we intend to merge it with `talentsearch` in the upcoming sprint anyway 🤔

PR #5139, opened 2022-12-15
Feature - Change Dependabot update day to Thursday

### LINKED ISSUE #5096: Move dependabot to Thursday
I propose moving "dependabot day" to Thursday.  This would have two benefits:
1. The occasional instability it causes would only happen at the start of a sprint or in the middle of one.
2. We might avoid some of the timeouts in CI since we would not be competing with most of the other clients on the default Monday to download packages and container images.

## Acceptance Criteria
- [x] Dependabot runs every Thursday

Ref: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduleday

mnigh:
Resolves #5096.

## Steps to test
1. Verify documentation is still accurate in **documentation/dependabot.md**.
2. Approve this PR and see if Dependabot opens a bunch of PRs tomorrow 😉

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,6 +11,7 @@ updates:
     directory: "/frontend"
     schedule:
       interval: "weekly"
+      day: "thursday"
     open-pull-requests-limit: 15
 
   # This ignores all updates to tc-report, since it's now served as a static website.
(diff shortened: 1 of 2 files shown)


vd1992 (approved):
seems to look right to me, one way to find out I suppose

PR #5150, opened 2022-12-15
Fix token tests

petertgiles:
This branch updates the modulus values advertised in the mock jwks file.  It seems that this value should be base64url encoded and our new library is more strict about this.

Testing: 
- disabled tests are enabled
- PHPunit tests reliably pass


esizer (approved):
Awesome! ⭐ Working well for me 👍 

```sh
  PASS  OpenIdBearerTokenTest
  ✓ accepts good token and returns correct sub
  ✓ rejects empty token
  ✓ rejects nonsense token
  ✓ rejects incorrect issuer
  ✓ rejects expired token
  ✓ rejects future token
  ✓ rejects token signed with different key
  ✓ accepts expired token within allowable skew
  ✓ that introspection can reject a valid token
```

PR #5166, opened 2022-12-16
testing/pool-candidate-workflow-issue TEST FIX

### LINKED ISSUE #5165: 🧪 Tests pool-candidate.cy failing
## 🧪 Tests

E2E test failing on `pool-candidate.cy`

## 🕵️ Details

Test fails as sometimes there is no one to "view application" on due to seeding randomness

## 🙋‍♀️ Proposed Solution

Constrain pool selection

## ✅ Acceptance Criteria

- [ ] Tests pass consistently

vd1992:
closes #5165 
resolves an issue with the e2e workflow failing
the workflow would fail sometimes on `pool-candidates.cy`, this was due to there sometimes not being a candidate for "view application" to work on

solution was to always select candidates for Digital Careers, given it has the most candidates seeded, it should almost always have someone to "view application" on
in the long run it is another test to revisit, but for now this should suffice
and make reviewing doable again!!!

testing:
re-run e2e a few times to see if it is all good

### DIFF
--- a/frontend/cypress/e2e/admin/pool-candidate.cy.js
+++ b/frontend/cypress/e2e/admin/pool-candidate.cy.js
@@ -28,15 +28,11 @@ describe("Pool Candidates", () => {
   it("should update pool candidate status", () => {
     cy.wait("@gqlgetPoolsQuery");
 
-    cy.findAllByRole("link", { name: /view candidates/i })
-      .eq(0)
-      .click();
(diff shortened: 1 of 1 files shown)


vd1992:
clicking on "availability" removed as it wasn't needed, the default sort is perfect in this case

petertgiles (approved):
:fire:

PR #5169, opened 2022-12-16
Bug - Fix PR Template

esizer:
## 👋 Introduction

This moves the PR template to a new filename that should be picked up as the default to no longer need adding it as a search param.


mnigh (approved):
👍 ? I guess we shall see!

Issue #4949, opened 2022-11-28
Send custom events to AppInsights 


brindasasi:
**Problem**

We don't have enough telemetry about exceptions that are happening in Prod. 
We get the below telemetry but there is no trace for what api has been called in here. 
![Screenshot 2022-11-28 134556.png](https://images.zenhubusercontent.com/61a7bd49e20af80694dd845e/ff0b2954-c115-4ac2-97a2-c2c5a36e219d)

### A/C 

- Investigate what type of telemetry details we can get more out of the box like turning on FETCH tracking 
- How do we use custom events to track the user journey ? -> Add the ability to do send custom events

mnigh:
Sentry is a cool tool: https://docs.sentry.io/product/integrations/source-code-mgmt/azure-devops/

PR #5176, opened 2022-12-19
Add the ability to send custom events to Azure App Insights 

brindasasi:
## 👋 Introduction

Adding custom events to Azure AppInsights via react hooks 

## 🕵️ Details
This PR provides a generic custom event hook which can be used to track any event in the app. 
It takes ReactPlugin instance , event name and event data. ReactPlugin contains the connection info to Azure AppInsights. 

## 🧪 Testing
This is a foundation branch. Example usage of custom event could be  adding the below snippet in button click of submit button: 

```
 ** Step 1. Declare the tracking event** 
  const reactPlugin = useAppInsightsContext();

  const trackCandidateRequest = useCustomEvent(
    reactPlugin,
    "Candidate Request Submitted"
  );
  
  ** Step 2: use it where it needed **
trackCandidateRequest({
        candidateCount,
        poolId,
        selectedClassifications,
      });
```

For testing it requires logging into Azure AppInsights .  Submit a search request in dev environment and verify that it is tracked in the `AppInsights` -> `Transaction Search`

## 📸 Screenshot

image

## 🤖 Robot Stuff

Resolves {#4949}


brindasasi:
I removed the usage of custom event as we decided to merge in the base version only under this ticket. Let me know if I'm wrong on that.

frontend/common/src/components/AppInsights/AppInsightsContext.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { createContext } from "react";
+import { reactPlugin } from "./ReactPlugin";
+
+const AppInsightsContext = createContext(reactPlugin);
+
+const AppInsightsContextProvider = ({
+  children,
```
petertgiles, line 7: Is this context provider being used anywhere?  Shouldn't it be used in our context sandwich like the locale provider and client provider?

frontend/talentsearch/src/js/components/Layout.tsx
```diff
@@ -137,61 +124,63 @@ const Layout = () => {
   }
 
   return (
-    <>
-      <Helmet>
-        <html lang={locale} />
-        <title>
+    <AppInsightsContextProvider>
```
petertgiles, line 127: Hmm, I wonder if it would make sense to instead have this in the common context sandwich.  Then it would be along with the other context provders like the client provider and the locale provider.  It would make it available to the other two workspaces, too.

frontend/common/src/hooks/useCustomEvent.ts
```diff
@@ -0,0 +1,23 @@
+import { ReactPlugin } from "@microsoft/applicationinsights-react-js";
+import { ICustomProperties } from "@microsoft/applicationinsights-web";
+import { useState, useEffect, useRef } from "react";
+
+export default function useCustomEvent(
```
petertgiles, line 5: Could we give this a more descriptive name?  Maybe useAppInsightsEvent?

frontend/common/src/hooks/useCustomEvent.ts
```diff
@@ -0,0 +1,23 @@
...
+  reactPlugin: ReactPlugin,
```
petertgiles, line 6: Does reactPlugin have to be a parameter?  Could this hook just call `useAppInsightsContext();` itself?

petertgiles (changes requested):
Looking cool!  I've got a couple of quick questions.  Also, don't forget to commit your package-lock.json changes.

frontend/common/src/hooks/useCustomEvent.ts
brindasasi: hmm ..all the events triggered by this type will show up as "custom event" in App Insights. It is the type of the event.

frontend/common/src/hooks/useCustomEvent.ts
brindasasi: it has to call reactPlugin.trackEvent which is not available in that hook. I will give it another look on this

brindasasi: this has been resolved now

frontend/common/src/hooks/useCustomEvent.ts
petertgiles: Maybe useAppInsightsCustomEvent then?  It would make it clear that it is associated with then other AppInsights components.  I'm worried that useCustomEvent is too generic to have any meaning.

frontend/talentsearch/src/js/components/Layout.tsx
brindasasi: I hope I've addressed this. Let me know if that isn't what you meant.

frontend/common/src/hooks/useCustomEvent.ts
petertgiles: Uh oh, typo:
image

petertgiles (approved):
Yay!  I think we're good.

petertgiles (approved):

PR #5185, opened 2022-12-20
Feature/backend experiences skill filtering

### LINKED ISSUE #5030: Only show "nice to have" and "need to have skills" on the Snapshot
As an admin, I only want to see relevant information while screening candidates so I can quickly assess them.

## ⚙️ Summary

DCMO is getting application snapshots with 80+ skills on them. This is making the snapshot hard to parse.

We should remove unrelated skill information on the application snapshot.

## Acceptance Criteria
- [ ] Backend operation
- [ ] Only need to have and nice to have skills show up in the **application snapshot** 
- [ ] Only need to have and nice to have skills show up in saved-archived text version of the snapshot
- [ ] Other skills do not show up at all

vd1992:
🤖 Resolves #5030

## 👋 Introduction

Stop saving every user skill in the snapshot, instead, only save skills that are relevant to the pool. 
Thanks @petertgiles for giving some pointers. 

## 🕵️ Details

Experiences with zero skills left after filtering are to be kept and shown. 
Done by passing an array of skill ids into the UserResource class and then filtering a collection with it.

## 🧪 Testing

1. Get a profile ready to go, bunch of skills and experiences attached (seeding should take care of this)
2. Head to browse opportunities, note which pool it is and its associated skills, apply to it
3. Go to the admin dashboard and view candidates for the pool you applied to and find the submitted application
4. Verify the snapshot only contains skills associated with the pool

## 📸 Screenshot

image


api/app/Http/Resources/UserResource.php
```diff
@@ -32,6 +40,17 @@ public function toArray($request)
         $collection = $collection->merge($personalExperiences);
         $collection = $collection->merge($workExperiences);
 
+        // if poolSkillIds is valid, filter out skills that are not in the id array by stepping through each item in $collection
+        if ($poolSkillIds && count($poolSkillIds) > 0) {
```
petertgiles, line 44: Careful, you can't for test for null like that in PHP.  However, if you use the `empty` fuction instead of `count` it will handle null safely.
image

petertgiles (changes requested):
Looks great!  Works well.  Thanks for adding the test.

Just noticed one fix needed for null handling.

api/app/Http/Resources/UserResource.php
vd1992: ooh, got it, thanks!
https://github.com/GCTC-NTGC/gc-digital-talent/pull/5185/commits/24151228bbfb41ce7bea243955913350978b6313

vd1992: a case where habits from one language slide over to another 😃

petertgiles (approved):

PR #5194, opened 2022-12-21
2404 prep for react 18

petertgiles:
## 👋 Introduction

This branch is prep work for the React 18 upgrade.  The type definitions are a little more strict at that version and so this branch preemptively knocks out most of the compile-time errors.

## 🕵️ Details

90% of the changes are related to two things:
- The `children` prop must be explicitly declared. Ref: https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-typescript-definitions
- FormatMessage is more strict about its types.  Ref: https://github.com/formatjs/formatjs/issues/3633#issuecomment-1245103640

## 🧪 Testing

1. Rebuild the project
2. No changes to how the site works.

### DIFF
--- a/frontend/admin/src/js/components/DashboardContentContainer.tsx
+++ b/frontend/admin/src/js/components/DashboardContentContainer.tsx
@@ -1,8 +1,8 @@
 import React from "react";
 
-export const DashboardContentContainer: React.FC = ({
-  children,
-}): React.ReactElement => {
+export const DashboardContentContainer: React.FC<{
+  children?: React.ReactNode;
+}> = ({ children }): React.ReactElement => {
   return (
     <div data-h2-min-height="base(100%)">
       <div data-h2-container="base(center, full, x2)">
--- a/frontend/admin/src/js/components/Table/Table.tsx
+++ b/frontend/admin/src/js/components/Table/Table.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable react/jsx-no-useless-fragment */
 /* eslint-disable react/jsx-key */
 import "regenerator-runtime/runtime"; // Hack: Needed for react-table?
 import React, { HTMLAttributes, ReactElement } from "react";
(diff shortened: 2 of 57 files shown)


esizer:
> * The `children` prop must be explicitly declared. Ref: https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-typescript-definitions

Yeah, I know @patcon made the suggestion that we avoid `React.FC` typing since it implicitly adds stuff we sometimes don't want (specifically `children`). Looks like this was a great suggestion since they have [removed implicit children](https://solverfox.dev/writing/no-implicit-children/) 😄 

This may be a big one since we were using `const Component: React.FC<T>` a lot. 😬

tristan-orourke:
I went looking for the current definition of the React.FC type, found it here https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts

```
    type FC<P = {}> = FunctionComponent<P>;

    interface FunctionComponent<P = {}> {
        (props: P, context?: any): ReactElement<any, any> | null;
        propTypes?: WeakValidationMap<P> | undefined;
        contextTypes?: ValidationMap<any> | undefined;
        defaultProps?: Partial<P> | undefined;
        displayName?: string | undefined;
    }
```

Looks like they removed children, but there's still some other stuff that we never use. I'm not against removing it. Should we make that its own PR? There are 244 instances in 185 files of React.FC or React.FunctionComponent 😆

esizer:
> Looks like they removed children, but there's still some other stuff that we never use. I'm not against removing it. Should we make that its own PR? There are 244 instances in 185 files of React.FC or React.FunctionComponent 😆

I think so 👍 

See: https://github.com/facebook/create-react-app/pull/8177

tristan-orourke:
Okay, React.FC has its own ticket now: https://github.com/GCTC-NTGC/gc-digital-talent/issues/5214

esizer (approved):
Everything is working for me with no errors. Think this is safe to merge 👍

PR #5218, opened 2022-12-23
feature/5217-candidates-table-default-sorting

### LINKED ISSUE #5217: Sort candidates table by default (after Status and Priority)
### Description
There doesn't seem to be any sorting on the candidates table after Status and Priority. This seems to cause a random selection of any given status and priority to be shown when a single group spans more than one page.

Simply having the default sorting of date-received (after status and priority) should fix it.  Oldest applications (aka first received) should be shown at the top of the table.

vd1992:
🤖 Resolves #5217 

## 👋 Introduction
Change the default sort on the Pool Candidates Table component. Setting it to sort by submitted date, oldest applications first. 

## 🕵️ Details
Primary sort is by status
Secondary sort is by priority
Tertiary sorting is what is being changed here

## 🧪 Testing

1. go to `/en/admin/pools/:id/pool-candidates`
2. observe that candidates with the same status and priority are ordered by oldest applications first
3. ensure table still works fine

## 📸 Screenshot

image

### DIFF
--- a/frontend/admin/src/js/components/poolCandidate/PoolCandidatesTable.tsx
+++ b/frontend/admin/src/js/components/poolCandidate/PoolCandidatesTable.tsx
@@ -270,13 +270,14 @@ const PoolCandidatesTable: React.FC<{
         },
       };
     }
+    // input cannot be optional for QueryPoolCandidatesPaginatedOrderByRelationOrderByClause
+    // default tertiary sort is submitted_at,
     return {
-      column: "status_weight",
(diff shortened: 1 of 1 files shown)


esizer (approved):
LGTM! 👍 Kind of confusing since `status` and `priority` take priority over `submitted_at` but it looks like it is doing what it says it does!

yonikid15 (approved):
LGTM 🎉 nice work

PR #5242, opened 2022-12-29
Bug - 4814 - Reset notes field on submission

### LINKED ISSUE #4814: Notes showing as unsaved on admin request page
On `http://localhost:8000/en/admin/talent-requests/:requestId` you can type in a note and click save but the form still thinks the note is unsaved.

esizer:
🤖 Resolves #4814

## 👋 Introduction

This fixes an issue where the notes fields was not being cleaned after submission.

## 🕵️ Details

Since the request form is not a traditional form we are relying on a function to save notes rather than the submit action. This resets the field to the data it currently has as a way of cleaning it.

## 🧪 Testing

1. Build admin `npm run production --admin`
2. Navigate to `/admin/talent-requests`
3. View a request
4. Edit the notes field, noting that it marks it as unsaved
5. Click the Save Notes button
6. Confirm the unsaved message disappears once the data has been saved

### DIFF
--- a/frontend/admin/src/js/components/searchRequest/UpdateSearchRequest.tsx
+++ b/frontend/admin/src/js/components/searchRequest/UpdateSearchRequest.tsx
@@ -47,6 +47,13 @@ export const UpdateSearchRequestForm: React.FunctionComponent<
       adminNotes: data.adminNotes,
     })
       .then(() => {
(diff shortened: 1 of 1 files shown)


JamesHuf (approved):
Works for me!

PR #5256, opened 2023-01-03
Feature - 5131 - Add User to Multiple Pools

### LINKED ISSUE #5131: Batch "Add to pools"
## Description

Allow for a user to be added to multiple pools at once, change the all stats to qualified available and only put the expiration date once.

## To-do

Convert current selection for pool to a multi-value or checkbox in the modal when adding a user to a new pool so we can add them to many pools at once. 

Allow date of expiration to the same modal

## Screenshot: Prototype

image

## Acceptance Criteria
- [ ] Select pool in the "add to pool dialogue" is a multi-select
- [ ] User should be added to all selected pools

esizer:
🤖 Resolves #5131 

## 👋 Introduction

This allows admins to add a user to multiple pools at once.

## 🕵️ Details

Updates the `<Select />` to a `<MultiSelectV2 />`, loops through the selected IDs adding mutations to an array so we can utilize [`Promise.allSettled()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) to show toast once all mutations have completed.

## 🧪 Testing

1. Build admin `npm run build`
2. Navigate to `/admin/users`
3. View a user
4. Clikc "Add user to pool` dialog
5. Select more than one pool
6. Click "Add to new pool"
7. Observe the mutations running in the network tab
8. Confirm
    - [ ] user is added to pools
    - [ ] toast is shown
    - [ ] list of pools updates
9. Comment out the line where we filter out existing pools (157)
10. Attempt to add a user to pools they are already in
11. Confirm
    - [ ] Error toast appears with list of failed pools

## 📸 Screenshot

Screenshot 2023-01-03 at 1 56 59 PM


brindasasi (changes requested):
This is causing a network call for each pool id. While it works I'm skeptical about so many calls which would be performance hit. Can we do it in bulk ? May be the one Yoni has suggested ? 
Screen Shot 2023-01-04 at 2 48 18 PM

esizer:
> This is causing a network call for each pool id. While it works I'm skeptical about so many calls which would be performance hit. Can we do it in bulk ? May be the one Yoni has suggested in the issue ?

So, I did look into that approach and `urql` does not support batch requests right now. The only other way is if you know the number of mutations ahead of time (which we do not). The only other solution would be to create an entirely new mutation that accepts an array of IDs, would that be preferred?

brindasasi:
I have seen discussion going on this bulk mutation issue. If we couldn't make the bulk mutation work, I would prefer the new mutation. Even if we don't do more than 10 - 15 requests at a time , it would still be a better way of doing things and if we invest time on this now , it can be reused across the application like  updating batch status  and many other. Thats just my opinion on this.

brindasasi:
@Jerryescandon this reminds me should we have limit on the UI on how many pools can be added ? or we would leave that to unlimited ?

Jerryescandon:
> @Jerryescandon this reminds me should we have limit on the UI on how many pools can be added ? or we would leave that to unlimited ?

Unlimited, we have no reason to limit them right now, it's gonna be at DCM's discretion

frontend/admin/src/js/components/user/GeneralInformationTab/AddToPoolDialog.tsx
```diff
@@ -89,6 +139,20 @@ export const AddToPoolDialog: React.FC<AddToPoolDialogProps> = ({
   };
   const { handleSubmit } = methods;
 
+  const poolOptions = pools
+    // .filter((pool) => !currentPools.includes(pool.id))
```
brindasasi, line 143: could you remove commented code ?

brindasasi (changes requested):

frontend/admin/src/js/components/user/GeneralInformationTab/AddToPoolDialog.tsx
esizer: 😬 That wasn't supposed to be committed. Great catch!

frontend/admin/src/js/components/user/GeneralInformationTab/AddToPoolDialog.tsx
```diff
@@ -57,24 +63,68 @@ export const AddToPoolDialog: React.FC<AddToPoolDialogProps> = ({
...
+              description:
+                "Toast for successful add user to pool on view-user page",
+            }),
+          );
+        } else {
+          toast.error(
+            <>
+              {intl.formatMessage({
+                defaultMessage: "Failed to add user to the following pools:",
+                id: "6Y6hy9",
```
brindasasi, line 114: do we have the french translation for this?

esizer: I'm just waiting to get it back right now

brindasasi: probably you could do it with the existing message like this 
https://github.com/GCTC-NTGC/gc-digital-talent/pull/5295/files#diff-15a2f6f5b8b70c1bf926290de4dce397d0939765ca1959181f96c8a4b40cff24R131.


I believe the existing msg for this 
```
toast.error(
          intl.formatMessage({
            defaultMessage: "Failed adding user",
            id: "GZqEuI",
            description: "Toast for failed add user to pool on view-user page",
          }),
```

frontend/admin/src/js/components/user/GeneralInformationTab/AddToPoolDialog.tsx
brindasasi: ooh .. thats why I got the pools I have already been in .. 😆

esizer: Yeah, I commented it out to force and error while testing by attempting to add a user to a pool they already belonged to 😜

frontend/admin/src/js/components/user/GeneralInformationTab/AddToPoolDialog.tsx
esizer: Added in [add translation](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5256/commits/4db8d307d6603f80872be380c14695f06748b520)

brindasasi (approved):
LGTM!

Issue #5128, opened 2022-12-14
Only show "need to have" and "nice to have" skills during the application process (and add some copy)


gobyrne:
### Description
During the application process, users can only add the "need to have" and "nice to have" skills to their application. BUT if they go to their profile they can add as many as they want. Then, when they come back to the application, they see all the skills they added to their profile and can also add justifications for them.

This way more info than admins will be able to review during the initial application review. To save candidates time we can to be more transparent and consistent in how we collect this skills information.

![image.png](https://images.zenhubusercontent.com/5dc084854553760001149b53/3349d8f3-bf7e-408a-a5b4-eec058a9bbf1)

### Acceptance Criteria
- [ ] Only skills that are "need to have" or "nice to have" are visible on the experience page of the application.
- [ ] Only skills that are "need to have" or "nice to have" are visible on all 5 "add experience" forms. (this is already the case where you add skills, but you can see everything when you get to the justifications section)
- [ ] A new info bubble explains "Skills unrelated to this application will be hidden during the initial assessment of your application but will always be visible on your profile."  See screenshot, BUT please use the text here (I added "of your application").

patcon:
Potential refinement of A/C:
- [ ] Only skills that are "need to have" or "nice to have" are visible on the experience page of the application.
- [ ] Only skills that are "need to have" or "nice to have" are visible on all 5 "add experience" forms. (this is already the case where you add skills, but you can see everything when you get to the justifications section)
- [ ] hide unrelated skills from skill picker count (can’t see them to add them)
- [ ] experience page (all 3): hide skills unrelated to job poster
- [ ] experience page (all 3): only count skills relevant to job poster
- [ ] all skills (even unrelated) will be visible on profile
- [ ] A new info bubble explains "Skills unrelated to this application will be hidden during the initial assessment of your application but will always be visible on your profile." See screenshot, BUT please use the text here (I added "of your application").

PR #5260, opened 2023-01-04
Feature/5128 only show applicable skills during application

JamesHuf:
🤖 Resolves {issue #5128}

## 👋 Introduction

Hides unrelated skills on a users profile during an application.

## 🧪 Testing

1. Build talentsearch `npm run build`.
2. Navigate to a pool advertisement and take note of the required and asset skills.
3. Navigate to your profile.
4. Create a new experience on your profile with both some required/asset skills and some unrelated skills with a couple of the skills having justifications. 
5. Start an application to the pool advertisement viewed in step 2.
6. Scroll to the "Experience and Skills" section of the application and confirm the info bubble about unrelated skills being hidden exists.
7. Open the accordion for the experience created in step 4 and confirm only the required/asset skills are listed.
8. Select to edit the experience and confirm the 'Skills in detail' section only lists the required/asset skills.
9. Replace an existing skill with a new one with a different justification and submit.
10. Confirm the the accordion for the experience displays all relevant skills with the appropriate justifications.

## 📸 Screenshot

image


frontend/talentsearch/src/js/components/experienceForm/ExperienceForm.tsx
```diff
@@ -140,8 +132,43 @@ export const ExperienceForm: React.FunctionComponent<ExperienceFormProps> = ({
...
+          )
+        );
+      });
+      // eslint-disable-next-line no-param-reassign
+      experience.skills = experience?.skills?.filter(
+        (skill) =>
+          !irrelevantSkills?.find(
+            (irrelevantSkill) => irrelevantSkill.id === skill.id,
+          ),
+      );
```
esizer, line 153: Not sure if it counts as within scope of this issue but we don't really need to ignore this rule since it is recommend to use use a differnt config than we are currently using.

**Recommended**

```json
{
  "no-param-reassign": ["error", { "props": false }]
}
```

**Current**

```json
{
  "no-param-reassign": "warn"
}
```

REF: https://stackoverflow.com/a/42399879

Another option would to just use a new variable instead of modifying experience directly.

```tsx
const newExperience = {...experience};
newExperience.skills = ...;
```

esizer (changes requested):
Looking great! Everything is working as intened. I have a small concern about ignoring the eslint rule but not really a deal breaker since we aren't really defining the rule properly anyway.

Also, unrelated but im a little concerned about how much info we are frontloading the experiences with making adding/editing experiences daunting with all the information we are throwing at users.

Screenshot 2023-01-06 at 10 05 39 AM

frontend/talentsearch/src/js/components/experienceForm/ExperienceForm.tsx
petertgiles: I've run into this one before and just created a new variable like you suggested.  The linting rule change seems reasonable but  I'm not sure changing it would fix this case.  It's not a prop of skills that is being changed, it's skills itself, right?

esizer: I actually tested it and it does make this pass without ignoring the rule. However, I am not opposed to leaving the rule alone and just creating a new variable.

tristan-orourke: @petertgiles skills is a prop of experience, so I think the linting change would work.

However, even if it avoids the arg bugs described in that stackoverflow answer, I would avoid this kind of reassignment on principle 😅. Especially modifying something that was passed into the function/component as a prop, which is a case of [modifying something outside of this function's scope](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/documentation/style-guide.md#main-principles).

I'd much prefer `const skills = experience?.skills?.filter(...);`, or if you really need them in the full package,
```
const newExperience = {
  ...experience,
  skills = ...,
};
```

PR #5263, opened 2023-01-04
Bug - 5259 - Fix `deploy.sh` script

### LINKED ISSUE #5259: 🐛 Builds are failing in Azure pipelines since the turborepo merge
## 🐛 Bug

The Build Artifact process is failing in azure devops for new commits on main. This has been happening since Dec 30, likely when #5173, which included turborepo, was merged. 

## 🦋 Expected Behaviour

Every new commit to main should trigger a successful build.

## 🕵️ Details
[build error.txt](https://github.com/GCTC-NTGC/gc-digital-talent/files/10345549/build.error.txt)

## 📋 Steps to Reproduce

1. Log in to azure devops
2. Look at the recent builds, and see that they've failed.

## 📸 Screenshot

...

## 🙋‍♀️ Proposed Solution

I believe we need to update `infrastructure/bin/deploy.sh`. If turborepo works in the azure environment, then we should be able to replace most of the deploy file with a single line of `npm run build`. I think?

## ✅ Acceptance Criteria

- [ ] New commits to main trigger a successful build in azure devops.

esizer:
🤖 Resolves #5259 

## 👋 Introduction

Fixed the `deploy.sh` script with new commands.

## 🕵️ Details

When we started the process to merge projects, we changes the `npm` scripts up a bit and they are now called in the root folder only once instead of individually. This updates our `deploy.sh` script to change how the frontend is built.

## 🧪 Testing

1. ???

### DIFF
--- a/infrastructure/bin/deploy.sh
+++ b/infrastructure/bin/deploy.sh
@@ -34,43 +34,12 @@ sudo chmod -R 775 ./ ./storage
 php artisan lighthouse:print-schema --write
 
 ### Install all npm dependencies
-cd $ROOT_DIR/frontend
+cd $ROOT_DIR
 npm ci --include=dev
 
-# Run h2-build using workspace command from /frontend
-npm run h2-build --workspace common
-
-### Common
-
-cd $ROOT_DIR/frontend/common
-npm run codegen
-npm run intl-compile
-
-### Talentsearch
-
-cd $ROOT_DIR/frontend/talentsearch
-npm run codegen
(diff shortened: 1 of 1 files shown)


brindasasi (approved):
This looks good on Azure

tristan-orourke (approved):
Tested this on Azure devops - it works! I love how simple this makes the build script. I suppose the complexity has been moved elsewhere (package.json commands and turbo config) but its a great tradeoff.

PR #5266, opened 2023-01-04
a11y - 4860 - Use full month name for skill accordions

### LINKED ISSUE #4860: Align month abbreviations with GoC guidelines
We're using abbreviated month names in several places in the app, including experience accordion date ranges.  To render this we're using the date-fns format function with token MMM.
https://date-fns.org/v2.29.3/docs/format
This gets us pretty close to the GoC standard but is not quite right.
https://www.noslangues-ourlanguages.gc.ca/fr/cles-de-la-redaction/mois.html
In English, for example, June and July should not be abbreviated and abbreviated months should end with a a period.

Acceptance Critera:
- [ ] Show the full month word everywhere.
Ref: https://github.com/GCTC-NTGC/gc-digital-talent/pull/4819#pullrequestreview-1188713641

esizer:
🤖 Resolves #4860 

## 👋 Introduction

This changes the skill accordions to present months with their full name rather than abbreviations.

## 🕵️ Details

The month name abbreviations do not match the guidelines for Canadian month abbreviations and are less helpful than presenting the full month name. For this reason, we are no longer using abbreviations for month names.

## 🧪 Testing

2. Navigate to the user profile `/talent/profile`
3. Scroll to the accordions in the "My skills and experience" section
4. Confirm the months are the full name in both French and English

## 📸 Screenshot

Screenshot 2023-01-04 at 4 47 34 PM

### DIFF
--- a/frontend/common/src/components/UserProfile/accordionUtils.tsx
+++ b/frontend/common/src/components/UserProfile/accordionUtils.tsx
@@ -5,7 +5,7 @@ import { Maybe, Scalars } from "../../api/generated";
 export function formattedDate(date: Scalars["Date"], intl: IntlShape) {
   return formatDate({
     date: new Date(date),
-    formatString: "MMM RRRR",
+    formatString: "MMMM RRRR",
     intl,
   });
 }


mnigh (approved):
👍

PR #5289, opened 2023-01-05
a11y - 5007 - Remove `<br />` tags to improve a11y

### LINKED ISSUE #5007: ⌨ Accessibility - My Profile - screen reader reads out some text twice
Some of the text in the My Profile page is being read twice when using Voiceover on Chrome.  See video below:

https://user-images.githubusercontent.com/1794660/205418330-34bcd6aa-0553-4013-9d22-7c82896ae077.mp4

### How to reproduce
1. Environment (prod or UAT): prod, talent.canada.ca/en/users/.../profile
2. Which browser and/or Assistive Technology? Voiceover on Mac - Chrome browser

**Step-by-step instructions explaining how to reproduce the problem.**

1. Enable voiceover
2. Make sure that you've filled out the "about me" and "language information" sections of the page
3. navigate through the page until you get to the text in the "about me" section.
4. Use voiceover key + right arrow key to read out the text.

Notice that some of the same text is read twice.

### Acceptance Criteria

- [ ] Ensure text is read once and correctly.
- [ ] Remove `<br>` tags in between the text items and just encompass them in `<p>` tags.  This seems to make the text read correctly. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br#accessibility_concerns
- [ ] Check across components in My Profile for any similar `<br>` tags and replace these as well.
- [ ] Ensure there is no visual difference after changing any html/css structure.

esizer:
🤖 Resolves #5007 

## 👋 Introduction

Removes `<br />` tags from the application due to the issues with a11y.

## 🕵️ Details

Text was being read out twice on the profiles which was caused by the use of the `<br />` tags. This removes those tags to prevent any issues with screen readers.

## 🧪 Testing

1. Build the projects `npm run build`
2. Navigate to a user profile `/talent/profile`
3. Navigate through the page with a screen reader
4. Confirm no text is read out twice


yonikid15 (commented):
Nice work Eric! However, I'm using NVDA and it seems to be reading things incorrectly for me. For example, when navigating through the "About Me" section using the up and down arrow keys, it's reading all the text in the section 😕. It might be that I'm using NVDA incorrectly, however I tried doing what Maddie showed in her video, changing the label and the text to p tags, and that works for me. 

It's hard to explain in text, but I can go over it with you in a huddle if you want.

esizer:
> Nice work Eric! However, I'm using NVDA and it seems to be reading things incorrectly for me. For example, when navigating through the "About Me" section using the up and down arrow keys, it's reading all the text in the section 😕. It might be that I'm using NVDA incorrectly, however I tried doing what Maddie showed in her video, changing the label and the text to p tags, and that works for me.

What browser are you using when testing? I want to see if I can reproduce this.

yonikid15:
> What browser are you using when testing? I want to see if I can reproduce this.

I was using Chrome.

esizer:
> I was using Chrome.

Okay, so NVDA works best with Firefox. However, that doesn't mean we should ignore chrome so I looked into it and the functionality you described matches the expected outcome.

- NVDA + Firefox - Separates `<p>` and reads them individually.
- NVDA + Chrome - Reads all content (`<p>`) within a section (`<h*>`)

yonikid15:
If that's the expected outcome on Chrome, then this is good to go 👍 
I just tested it on Firefox and it's working as expected.

yonikid15 (approved):
🥇

Issue #5254, opened 2023-01-03
🧪 Pool Candidate Test Failing


esizer:
## 🧪 Tests

The `e2e` test `pool-candidate.spec.js` is consistently failing. 

## 🕵️ Details

The test looks for a **View application** link in the table that does not seem to exist. From testing, it was observed that the test only fails in github actions and is passing locally.

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the debt tests were properly updated and remains functional/passing.

- [ ] `pool-candidate.spec.js` consistently passes locally
- [ ] `pool-candidate.spec.js` consistently passes in github actions
- [ ] No changes to _expected_ application functionality

tristan-orourke:
Update the test to be reliable regardless of seeders. Create the data you need with API calls before running the test.

PR #5292, opened 2023-01-05
Feature - 5187 - Add PoolCandidateStatus `Removed`

### LINKED ISSUE #5187: Add new status for candidates in a pool
As an admin I want to flag when a candidates has been placed through another pool, so they no longer turn up in searches and I can track which pool they were placed in.

### Description

Add a new status for pools
**English:** Removed - placed through other pool
**French:** Retiré - placé à partir d'un autre bassin

Like the "placed" status, these candidates should no longer show up in searches or requests.

## Acceptance Criteria
- [ ] Add new migration for the "Removed" status (Question: What weight should the "Removed" status be attached too? Weight can be 160 (end of table))
- [ ] Add "Removed" status to PoolCandidateStatus enum in schema
- [ ] Add status to localized constants
- [ ] Update all search queries to remove candidates that have status = Removed
- [ ] Update sorting on the Pool Candidates table based on the new status weight (Question: Do we even want to display removed candidates? The option should be there to see them if you want, but ideally they would be filtered out by default)

mnigh:
🤖 Resolves #5187.

## 👋 Introduction

This PR adds `Removed` as a PoolCandidateStatus option.

## 🧪 Testing

1. Refresh All projects `docker-compose run --rm maintenance bash refresh_all.sh`
2. Navigate to `/admin/candidates/:id/application`
3. Verify **Removed** appears as an option in the **Candidate pool status** select field and can be saved
4. Navigate to `/admin/users/:id`
5. Verify **Removed** appears as an option in the **Pool status** select field under the **Pool status** section (click Change status) and can be saved


yonikid15 (approved):
Everything's working as expected! Nice work 🎉 

I found one unrelated bug which I'll make a issue for. If your on a table and going through the pages and decide to apply a filter (in this case REMOVED status) if the number of results (eg. only 3 candidates are returned) doesn't reach that page (eg. I clicked apply filters on page 3) then the table will be empty. We need to ensure the current page is updated to 1 after applying the filters.


https://user-images.githubusercontent.com/22059495/211079394-a31a97e8-7730-44c1-acd2-c11206573c00.mp4

mnigh:
Merging into `main` branch despite failing e2e-tests check since problem was pre-existing this branch and an [issue has already been created](https://github.com/GCTC-NTGC/gc-digital-talent/issues/5254).

PR #5308, opened 2023-01-10
Bug/5304 skill accordion justifications

### LINKED ISSUE #5304: 🐛 Bug - Skills appear on different experiences depending on sorting
## 🐛 Bug

When viewing applications on the admin site:
The skill justifications attached to a specific experience is different when sorting by date vs by skill.

Gray reproduced this with 2 different applicants on prod.

## 🦋 Expected Behaviour

Skill justifications should be attached to a specific skill/experience and not move around based on how they are sorted.

## 📸 Screenshot
Screenshots include personal information. Let Gray know if you need to see them.

## ✅ Acceptance Criteria

- [ ] Criteria 1
- [ ] Criteria 2

vd1992:
🤖 Resolves #5304 

## 👋 Introduction

Skill justifications do not match the skills they should for the skill accordion

## 🕵️ Details

The problem to solve can be seen here

image

## 🧪 Testing

1. Create an application to a pool
2. Fill out multiple different experiences, with the required skills, ensure the skill justification for each makes it obvious where they came from
3. Submit application and then open up the application view in the admin dashboard
4. Ensure skills and their justifications go where they should for the skill accordion

## 📸 Screenshot

Like so

image

### DIFF
--- a/frontend/common/src/components/UserProfile/SkillAccordion/SkillAccordion.tsx
+++ b/frontend/common/src/components/UserProfile/SkillAccordion/SkillAccordion.tsx
@@ -39,15 +39,15 @@ export interface SkillAccordionProps {
  * @returns justification for the skill that matches the accordion skill
  */
 const grabSkillJustification = (
-  skillsArray: Skill[] | undefined,
+  skillsArray: Skill[],
   accordionSkill: Skill,
 ): string => {
   // find the skill from the array that matches the accordion skill, this results in an array of one skill
-  const specificSkill = skillsArray?.filter(
+  const specificSkill = skillsArray.filter(
     (skillIterator) => skillIterator.id === accordionSkill.id,
   );
-  const justification = specificSkill?.[0].experienceSkillRecord?.details
-    ? specificSkill[0].experienceSkillRecord?.details
+  const justification = specificSkill[0].experienceSkillRecord?.details
+    ? specificSkill[0].experienceSkillRecord.details
     : "";
   return justification;
 };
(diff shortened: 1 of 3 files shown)


vd1992:
no complaints from Storybook

image

frontend/common/src/components/UserProfile/SkillAccordion/SkillAccordion.tsx
```diff
@@ -33,6 +33,29 @@ export interface SkillAccordionProps {
...
+ * @param skillsArray - array of skills attached to experience
+ * @param accordionSkill - single skill that is the focus of a skill accordion
+ * @returns justification for the skill that matches the accordion skill or return an empty string
+ */
+const grabSkillJustification = (
+  skillsArray: Skill[],
+  accordionSkill: Skill,
+): string => {
+  // find the skill from the array that matches the accordion skill, this results in an array of one skill
+  const specificSkill = skillsArray.filter(
```
petertgiles, line 46: If you're looking for a single item in the array, you should probably use [Array.find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead of Array.filter

petertgiles (approved):
Wow, it works!  Indeed, Storybook seems happy.  Nice job!  :1st_place_medal:

frontend/common/src/components/UserProfile/SkillAccordion/SkillAccordion.tsx
vd1992: like so https://github.com/GCTC-NTGC/gc-digital-talent/pull/5308/commits/9c578f2f801d1dffbdec0c32d8879fcc1da1d912 ?

petertgiles: perfect!

Issue #5127, opened 2022-12-14
Expand language preferences


Jerryescandon:
## Description

Expand Language preferences in user profiles to include:
- Working Ability
- ~~FOL~~
- Written Exam
- Correspondence
- Interview language

## Other Considerations

Note as requirement:  Check how this looks on Admin? Does it show on tables? etc.

## Screenshot: Prototype

![image](https://user-images.githubusercontent.com/16977254/209225471-a11a35ea-c628-481f-8b54-27e17a9f3c80.png)

Vachan says:
migration with two fields
User.php changes
update schema for new inputs (updateUser)
update frontend (language form, profile view, tables, csv, download)

## Acceptance Criteria
- [ ] Two new *Required* language select fields on the Personal Information page of the profile
- [ ] Selected answers show up on the profile page
- [ ] Validation ensures you have selected these before progressing with your profile (like the correspondence language)
- [ ] Translation for the copy
- [ ] Update on the application snapshot to show these
- [ ] Update to all the downloads to include these

gobyrne:
The word "first" shouldn't be there for "exam language preference"

Jerryescandon:
Just to reiterate the conversation we had on the design team, about "First Official Language" and we have several concerns:
- The FOL label as-is is confusing and biased, it will deter diverse applicants from applying and reinforces colonialist values.
- It adds no new value since it is the same information collected on the Language section but worded differently.

Proposed solution:
- Use the information collected on the Language section and relabel it on admin side for HR consumption.

substrae:
@Jerryescandon just to confirm - did we want to go with the dropdowns as well?

![image](https://user-images.githubusercontent.com/6960386/208679375-1cb2a907-0b66-4565-a459-9e9e93af8a50.png)

Jerryescandon:
> @Jerryescandon just to confirm - did we want to go with the dropdowns as well? I think as also wanted to modify the labels to:
> 
> * Select your preferred spoken interview language
> * Select your preferred written exam language
> 

Yeah I think so, this hasn't moved into the sprint yet, But I think we should use the dropdowns if we are dropping FOL.

brindasasi:
@Jerryescandon do we need to add this in the account creation page as well?

Jerryescandon:
Do we have everything that is on the "About me" on the "Getting started"?
[...]

PR #5315, opened 2023-01-10
Feature/5127 langugage preference

brindasasi:
## 👋 Introduction
🤖 Resolves {#5127}

## 🕵️ Details
This pr adds 2 new fields "preferred_language_for_interview" ,  "preferred_language_for_exam" into db and basically everywhere preferred language has been added , these 2 new fields are also added. 

Existing `Preferred Language` has been renamed to `Preferred Communication Language`
New field labels `Preferred Spoken Interview Language`  and `Preferred Written Exam Language`

## 🧪 Testing
Verify all the below forms showing the new drop downs as per the screen shot , able to edit , save and retain the saved values. 

User

- [ ] About Me Section
- [ ] About Me Form

Admin 
- [ ] Users table
- [ ] Pool Candidates table
- [ ] CSV download of Pool Candidates 
- [ ] Create User
- [ ] Update User

## 📸 Screenshot
Screen Shot 2023-01-10 at 10 49 54 PM
Screen Shot 2023-01-10 at 10 49 41 PM

Admin/users

Screen Shot 2023-01-11 at 1 57 33 PM


frontend/talentsearch/src/js/components/createAccount/createAccountOperations.graphql
```diff
@@ -30,6 +30,8 @@ mutation CreateAccount($id: ID!, $user: UpdateUserAsUserInput!) {
     firstName
     lastName
     preferredLang
+    preferredLanguageForInterview
```
brindasasi, line 33: - [ ] revert this as we don't need it

brindasasi:
translations strings are pending. But rest would be good for review now

tristan-orourke (commented):
The two new language fields are required, in the Personal Information form. Should we add them to the isProfileComplete check? And if we do that, should we add a line to the migration which sets the new field to match the existing preferred_lang field for existing users?

frontend/admin/src/js/components/user/UserTable.tsx
```diff
@@ -324,15 +324,39 @@ export const UserTable = ({ initialFilterInput }: UserTableProps) => {
...
+        label: intl.formatMessage({
+          defaultMessage: "Preferred Spoken Interview Language",
+          id: "iRJV64",
+          description:
+            "Title displayed on the Pool Candidates table Preferred Spoken Language column.",
+        }),
+        id: "preferredLang",
+        accessor: (user) =>
+          languageAccessor(user?.preferredLanguageForInterview, intl),
+        sortColumnName: "PREFERRED_LANGUAGE_FOR_INTERVIEW",
```
tristan-orourke, line 346: I'm not entirely sure how sortColumnName works, but is there any reason its in all-caps here and not on other columns?

tristan-orourke (commented):
Haven't tested it yet, but everything looks good so far on a read-through of the code 👍

frontend/admin/src/js/components/user/UserTable.tsx
brindasasi: no .. thats by mistake ... good catch!

brindasasi:
> The two new language fields are required, in the Personal Information form. Should we add them to the isProfileComplete check? And if we do that, should we add a line to the migration which sets the new field to match the existing preferred_lang field for existing users?

yes. for sure. My understanding is that it should be in the profile complete check .

For existing users , it wouldn't affect the existing applications. Only on the new application they would be demanded to fill in the new values. Wouldn't that be ok? It would also be another reminder to tell them they can choose these values.

If not liking that approach, we could force update these values with preferred_lang value with migration. I'm just wondering if that would be defeating the purpose of having them to choose. But you know lets ask @Jerryescandon  here 😅 .

brindasasi:
> Haven't tested it yet, but everything looks good so far on a read-through of the code 👍

thanks! btw, where am I missing that profileCompleteCheck .. could you point me out?

Jerryescandon:
Just add them to the check moving forward, dont worry about migrating or retrofitting, DCM can manually ask people if they need to. Let just focus on building things now for future applications.

brindasasi:
pool candidates csv 
[pool_candidates_2023-01-12T19_50_32.519Z.csv](https://github.com/GCTC-NTGC/gc-digital-talent/files/10405491/pool_candidates_2023-01-12T19_50_32.519Z.csv)

tristan-orourke:
I just want to note that if we add these new fields to the profileIsComplete check, and we do not set default values for existing users, then any existing users who already completed their profiles will no longer be considered complete.

Jerryescandon:
> I just want to note that if we add these new fields to the profileIsComplete check, and we do not set default values for existing users, then any existing users who already completed their profiles will no longer be considered complete.

:NotLikeThis:

api/app/GraphQL/Mutations/PoolCandidateSnapshot.graphql
```diff
@@ -10,7 +10,8 @@ query getProfile($userId: ID!) {
     email
     telephone
     preferredLang
-    currentProvince
```
vd1992, line 13: I don't think this should be removed?

vd1992:
> I just want to note that if we add these new fields to the profileIsComplete check, and we do not set default values for existing users, then any existing users who already completed their profiles will no longer be considered complete.

in the migration, after the new columns are created, I think it would be reasonable to just default assign the value of `preferred_lang` to them
users can change them to their liking afterwards if they prefer, but profile completeness won't be broken

something like this, or however the SQL needs to be formatted

```
        DB::statement(
            <<<SQL
                UPDATE users
                    SET preferred_language_for_interview =
                        case preferred_lang
                            when :english then :english
                            when :french then :french
                        end
            SQL, [
                'english' =>'en',
                'french' => 'fr,
            ]
        );
```

api/app/GraphQL/Mutations/PoolCandidateSnapshot.graphql
brindasasi: oops .. thank you!

brindasasi:
> > I just want to note that if we add these new fields to the profileIsComplete check, and we do not set default values for existing users, then any existing users who already completed their profiles will no longer be considered complete.
> 
> :NotLikeThis:



> > I just want to note that if we add these new fields to the profileIsComplete check, and we do not set default values for existing users, then any existing users who already completed their profiles will no longer be considered complete.
> 
> in the migration, after the new columns are created, I think it would be reasonable to just default assign the value of `preferred_lang` to them users can change them to their liking afterwards if they prefer, but profile completeness won't be broken
> 
> something like this, or however the SQL needs to be formatted
> 
> ```
>         DB::statement(
>             <<<SQL
>                 UPDATE users
>                     SET preferred_language_for_interview =
>                         case preferred_lang
>                             when :english then :english
>                             when :french then :french
>                         end
>             SQL, [
>                 'english' =>'en',
>                 'french' => 'fr,
>             ]
>         );
> ```



> > I just want to note that if we add these new fields to the profileIsComplete check, and we do not set default values for existing users, then any existing users who already completed their profiles will no longer be considered complete.
> 
> :NotLikeThis:

@Jerryescandon how do you want to go about this? For existing users, should we assign communication language to other two as well ?

Jerryescandon:
I agree, I feel like its a fair assumption at this point to inherit the communication language selection and not break things, later on we (and by we I mean DCM) can send emails to let people know about these fields and prompt them to update if they so wish to.

brindasasi:
> > I just want to note that if we add these new fields to the profileIsComplete check, and we do not set default values for existing users, then any existing users who already completed their profiles will no longer be considered complete.
> 
> in the migration, after the new columns are created, I think it would be reasonable to just default assign the value of `preferred_lang` to them users can change them to their liking afterwards if they prefer, but profile completeness won't be broken
> 
> something like this, or however the SQL needs to be formatted
> 
> ```
>         DB::statement(
>             <<<SQL
>                 UPDATE users
>                     SET preferred_language_for_interview =
>                         case preferred_lang
>                             when :english then :english
>                             when :french then :french
>                         end
>             SQL, [
>                 'english' =>'en',
>                 'french' => 'fr,
>             ]
>         );
> ```

I just gave the other column name for that .. It seems to be working fine when tested

brindasasi:
After the latest main merge I get this error. Not sure why it would remove the french translations. Looking into it 
Screen Shot 2023-01-18 at 10 56 10 AM

vd1992:
warning observed on the users table

image

vd1992:
table columns, both toggle despite only clicking on one of them

gif

vd1992:
for download csv and profile print
on **USERS** table
language is missing, only one that appears is preference for english/french/bilingual column

image

vd1992:
snapshot is not capturing the new fields or not retrieving from the snapshot, I submitted an application on this branch after migrating

image

vd1992 (changes requested):
please see the above series of comments and pictures

most of the problems seem localized to the _users_ table, _pool candidate_ table appears to work 
csv download and table columns for example are only throwing issues on the _users_ table

brindasasi:
> language preferences are not reflected in the about me for candidate profile tab at `/admin/users/:id` thus they are not shown in the print either, I think they ought to be
> 
> image
> 
> image

This form shouldn't be affected. There is no language preference shown already.

brindasasi:
> for download csv and profile print on **USERS** table language is missing, only one that appears is preference for english/french/bilingual column
> 
> image

There shouldn't be csv download change on users table as well as I mentioned in the check list above. The language that is showing here is not the langugage preference. It is about the jobs/opportunities that the applicant willing to work on.

brindasasi:
> 

 label id looks to be the same which is causing this issue. I think I messed up after trans or merging.

brindasasi:
> warning observed on the users table
> 
> image

Seems to be the root cause for below issue on the table

brindasasi:
> does the snapshot view work on your end? it doesn't for me and I don't get why since I can see it is passing in the new fields
> 
> image

is this a old snapshot?  🤔

brindasasi:
> does the snapshot view work on your end? it doesn't for me and I don't get why since I can see it is passing in the new fields
> 
> image
It should be good now. Could you check? 
Screen Shot 2023-01-19 at 4 33 01 PM

vd1992:
yep snapshots parse properly now!

vd1992 (approved):
not sure if @tristan-orourke is or will be reviewing this...

but, things look good to me now, my feedback has been addressed
great job!

brindasasi:
> not sure if @tristan-orourke is or will be reviewing this...
> 
> but, things look good to me now, my feedback has been addressed great job!

Thanks Vachan! Will leave it until tmo. for merge

tristan-orourke:
Thanks, I think @vd1992 has this covered, don't wait for my review! 😄

PR #5341, opened 2023-01-13
Feature - 4629 - Remove CMO Assets

### LINKED ISSUE #4629: Deprecate CMO Assets filter after FEATURE_APPLICANTSEARCH flag is turned on
https://github.com/GCTC-NTGC/gc-digital-talent/blob/9968de2525e4614e54d5a66147279df65539f738/api/app/Models/PoolCandidate.php#L452

mnigh:
🤖 Resolves #4629.

## 👋 Introduction

This PR removes any data for CMO Assets and code that reference them.

## 🧪 Testing

1. Rebuild docker `docker-compose down; docker-compose rm webserver; docker-compose up --detach --build`
3. Setup Project `docker-compose run --rm maintenance bash setup.sh`
4. Confirm _admin_, _talentsearch_ frontend works without any changes to the frontend
5. Search repository for case insensitive versions of the following strings (and related strings): **cmo_asset**, **cmoAsset**, **cmo asset**, **essentialCriteria**, **assetCriteria**
6. Confirm only references of the strings appear in laravel migrations


esizer (approved):
LGTM! 👍 

- ✅ CMO only appears in migrations
- ✅ Migrations ran successfully
- ✅ App build and runs with no errors

PR #5345, opened 2023-01-13
debt/3228-react-strict

### LINKED ISSUE #3228: Use React StrictMode

## Description
StrictMode is a feature that helps with a lot of things (but only when running in development mode), including:
- double mounts each component to compare state and find memory leaks from missing unmount logic, or other mis-use of lifecycle hooks (e.g., useEffect, useMemo, useState)
- warns of specific deprecated methods in the app _and_ any dependant packages. Finding these early helps react version upgrades go smoothly.

StrictMode is the default for apps created with create-react-app, which makes it seem like a decent practice.

Reference: https://reactjs.org/docs/strict-mode.html

## Acceptance Criteria
- [ ] all apps are wrapped in `<React.StrictMode />` component 
- [ ] Do not merge if there are errors (yet) - Bring to stand-up or have a discussion about it.

## Not acceptance criteria
- [ ] all apps run in development mode without error (console or otherwise)

vd1992:
🤖 Resolves #3228 

## 👋 Introduction

Activate strict mode

## 🕵️ Details

So far seems to work fine for the most part. One exception with a warning thrown everywhere as seen below 

This appears to originate from react-helmet and is an issue noted here https://github.com/nfl/react-helmet/issues/694
The suggested solution does not work and is the same problem as seen here https://github.com/staylor/react-helmet-async/issues/145
Given this is a Warning and not Error thrown, it isn't a huge issue at the moment. 

## 🧪 Testing

1. Ensure nothing broke
2. ?

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/main.tsx
+++ b/apps/web/src/main.tsx
@@ -11,11 +11,11 @@ const container = document.getElementById("app");
 if (container) {
   const root = createRoot(container);
   root.render(
-    <>
+    <React.StrictMode>
       <ContextContainer messages={messages}>
         <Router />
       </ContextContainer>
       <Toast />
-    </>,
+    </React.StrictMode>,
   );
 }
--- a/frontend/admin/src/js/main.tsx
+++ b/frontend/admin/src/js/main.tsx
@@ -12,11 +12,11 @@ const container = document.getElementById("app");
 if (container) {
   const root = createRoot(container);
   root.render(
-    <>
+    <React.StrictMode>
       <ContextContainer messages={messages}>
         <Router />
       </ContextContainer>
       <Toast />
-    </>,
+    </React.StrictMode>,
   );
 }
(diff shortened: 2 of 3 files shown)


esizer:
> This appears to originate from react-helmet and is an issue noted here [nfl/react-helmet#694](https://github.com/nfl/react-helmet/issues/694)
> The suggested solution does not work and is the same problem as seen here [staylor/react-helmet-async#145](https://github.com/staylor/react-helmet-async/issues/145)
> Given this is a Warning and not Error thrown, it isn't a huge issue at the moment.

Agreed. Also, while looking into this I noticed `react-helmet` doesn't seem to be maintained anymore. I think it would be a good idea to migrate to `react-helmet-async` regardless if this issue has been resolved. I'll throw an issue 👍

esizer (approved):
Awesome! ⭐ 

I put up an issue to migrate to `react-helmet-async` on the board. I think you are correct on the issue not being super urgent but def something we should look at.

PR #5386, opened 2023-01-18
Fix hidden column controls on User table

### LINKED ISSUE #5371: 🐛 Table columns checkboxes erratic behaviour
## 🐛 Bug

When I try to check all the boxes in the Table Columns modal on the Users table on the Admin site it unchecks some boxes when I attempt to check the last box on the list.

In addition, the toggle all seems to be broken, it might be related to this or be something else 🤷 

## 🦋 Expected Behaviour

It should allow me to check all boxes without unchecking anything.

## 🕵️ Details

- **Operating System**: Windows (TBS tablet)
- **Browser**: Chrome
- UAT / Release 2.3.0

## 📸 Screenshot

image

## ✅ Acceptance Criteria

- [ ] User can select all boxes
- [ ] User can unselect all boxes
- [ ] User can select some boxes
- [ ] User can toggle all boxes with the _Toggle all_ box
- [ ] Table state is preserved

tristan-orourke:
🤖 Resolves #5371

## 👋 Introduction

Fixes glitchy behaviour on Hidden Column controls when a table has hidden columns by default, as is the case on the Users table.

## 🕵️ Details

Unexpected behaviour was being caused by two things:
1. showing all columns means storing an empty list in the hidden columns param. But an empty list was being stored in the url as an empty string, which was hard to distinguish from the param being undefined at all, and the param being undefined was interpreted as the defaultValue, which in this case is NOT an empty list.
2. a new value could be interpreted as different from the default value if they had the same columns but in a different order. 

These problems were resolved by
1. encoding the hidden columns list as a json string, making the empty list very explicit
2. converting lists to Sets before comparison

## 🧪 Testing

1. Go to http://localhost:8000/en/admin/users
2. By default, there should be no hiddenColumnIds param in the url
3. Using the Columns dialog, make sure the following all results in expected behaviour:
    - [ ] User can select all boxes
    - [ ] User can unselect all boxes
    - [ ] User can select some boxes
    - [ ] User can toggle all boxes with the _Toggle all_ box
    - [ ] Table state is preserved if you follow a link then go Back
4. If you uncheck only Created At, Updated At, and Telephone, the hiddenColumnIds param should disappear from the url.
5. The above all still works on a PoolCandidates table as well (except hiddenColumnIds param should disappear when all columns are visible).

### DIFF
--- a/frontend/admin/src/js/components/apiManagedTable/useTableState.ts
+++ b/frontend/admin/src/js/components/apiManagedTable/useTableState.ts
@@ -44,10 +44,18 @@ const setArray = <T>(
   newValue?: Array<T>,
   defaultValue?: Array<T>,
 ) => {
-  const prevValue = params.get(key);
-  if (!prevValue || !isEqual(prevValue.split(","), newValue)) {
-    if (newValue && !isEqual(newValue, defaultValue)) {
-      params.set(key, newValue.toString());
+  const prevValueEncoded = params.get(key);
+  const prevValue = prevValueEncoded
+    ? JSON.parse(decodeURIComponent(prevValueEncoded))
+    : undefined;
+  // Compare sets so that order doesn't matter
+  const prevValueSet = new Set(prevValue);
+  const newValueSet = new Set(newValue);
+  const defaultValueSet = new Set(defaultValue);
+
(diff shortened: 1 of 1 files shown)


esizer (approved):
Awesome! 👍 

`Set`'s 😍

PR #5422, opened 2023-01-20
Feature - 5314 - Update Page Heroes

### LINKED ISSUE #5314: ✨ Apply new header component to all pages
## ✨ Feature

Apply the header components as described in #5215 to all public facing pages (not admin). The component should already be built in #5298.

## 🕵️ Details

- Standard page header goes everywhere (other are the exception)
- Photo header is applied to the Browse Jobs page and Support/Page
- Centered layout goes on the manager search page

This will apply to all pages, regardless of whether or not they currently are styled for the "new look". This will actually make the pages look closer to the "new look", even though buttons are still the wrong color (for example).

Some pages also don't have headers at all right now. For those, there should still be an H1 (the profile sub-pages, for example). We should use the H1 text in the header and remove any links or similar that are between the header and the current location of the H1 (the **Cancel and go back** links are what we are thinking about right now).

(The homepage uses something else entirely, so don't do home page)

## ✅ Acceptance Criteria

- [ ] Go through site map, make sure all non-admin pages use the new header component.
- [ ] On edit profile pages, move title and subtitle into new Hero, and remove Go Back buttons.

Site map: https://docs.google.com/spreadsheets/d/17A3n9kCyr0lELtXYfegI8tNxlN1-vJVUK37NExYNHUI/edit#gid=1116634425

esizer:
🤖 Resolves #5314 

## 👋 Introduction

This updates all public pages to use the new hero component.

## 🕵️ Details

All profile forms have been updated to use the new hero as well so make sure to check those.

> **Note**
> @substrae Mentioned he was working on redesigning the profile form pages so they may look a little odd temporarily.

## 🧪 Testing

1. Build frontend `npm run build`
2. Navigate to each page confirming the new hero is there and spacing looks okay

## 📸 Screenshot

Screenshot 2023-01-20 131624

### DIFF
--- a/apps/web/src/components/ApplicationPageWrapper/ApplicationPageWrapper.stories.tsx
+++ b/apps/web/src/components/ApplicationPageWrapper/ApplicationPageWrapper.stories.tsx
@@ -29,7 +29,10 @@ BasicApplicationPageWrapper.args = {
   title: "Basic Application Page Wrapper",
   subtitle: "Subtitle for Page Wrapper",
   closingDate: FAR_FUTURE_DATE,
-  crumbs: [{ title: "Pool Name" }, { title: "About Me" }],
+  crumbs: [
+    { label: "Pool Name", url: "#pool" },
+    { label: "About Me", url: "#about" },
+  ],
   navigation: {
     currentStep: 1,
     steps: [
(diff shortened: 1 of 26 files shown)


substrae:
This has made such a massive difference 😱 Great work @esizer 

**Couple small nitpicks:**

For graphic headers specifically, we can probably have the subtitle extend to the edge of the container, as the background interference is minimal.

image

I think this is only happening on the centered card variation, but the whitespace difference when the subtitle is missing is different:

image

For content pages, can we increase the distance between the header and the topmost content to match the distance from the bottom of the subtitle to the breadcrumbs at minimum?

image

esizer:
okay @substrae , This should solve the issues you highlighted: [adjust spacing](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5422/commits/b4d5417065fc3d81c81fa2e2f709e962601e26f8).

substrae (approved):
This looks a-okay to me, man! 🥳

Issue #5418, opened 2023-01-20
Redirect from legacy search URL


gobyrne:
### Description
Before we implemented our site-map the link to the old search page was provided to the PSC. They are using it on their site for the Digital Careers pools: **http://extranet.psc-cfp.gc.ca/sib-dgsi/pools/index-fra.html**. Unfortunately that now takes them to a 404. This was flagged a while back (because people actually do use it) but I thought it would fix on their end. Let's just do it on our end though so I don't have to keep worrying about it.

### Acceptance Criteria

- [ ] https://talent.canada.ca/en/talent/search permanent redirect to https://talent.canada.ca/en/search
- [ ] https://talent.canada.ca/fr/talent/search permanent redirect to https://talent.canada.ca/fr/search
- [ ] Cypress test to ensure these two continue to work

petertgiles:
We have a similar redirect in our nginx routing already.
![image.png](https://images.zenhubusercontent.com/6172cf6e898e11fd20507433/40c2b18b-9910-4ce0-bc5a-cfd6a147f2d5)

PR #5433, opened 2023-01-23
Feature/5418 url redirect

brindasasi:
🤖 Resolves {#5418}

## 👋 Introduction

## 🕵️ Details

## 🧪 Testing

1. ...
2. ...

## 📸 Screenshot

### DIFF
--- a/apps/e2e/cypress/e2e/talentsearch/redirects.cy.js
+++ b/apps/e2e/cypress/e2e/talentsearch/redirects.cy.js
@@ -2,7 +2,7 @@ import { aliasQuery } from "../../support/graphql-test-utils";
 
 const uuidRegEx = /[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[089ab][0-9a-f]{3}-[0-9a-f]{12}/;
 
-describe("Redirects", () => {
+describe.only("Redirects", () => {
   const expectToBeOnProfile = () => {
     cy.wait("@gqlgetMeQuery");
     cy.url()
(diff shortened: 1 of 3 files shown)


apps/e2e/cypress/e2e/talentsearch/redirects.cy.js
```diff
@@ -27,4 +27,26 @@ describe("Redirects", () => {
...
+  cy.request({
+    url: oldUrl,
+    followRedirect: false, // turn off following redirects
+  }).then((resp) => {
+    // redirect status code is 301
+    expect(resp.status).to.eq(301);
+    expect(resp.redirectedToUrl).to.eq(newUrl);
+  });
+});
+
```
vd1992, line 51: a little nitpicky, but I think it would be good to cover redirects in /fr a bit too, rather than just english in tests
the other redirect tests were just english so far

https://github.com/GCTC-NTGC/gc-digital-talent/blob/f605e02562826b1c90a3225061d560960fcfb8cd/apps/e2e/cypress/e2e/talentsearch/redirects.cy.js#L18-L29

vd1992 (approved):
I left a minor thing for you to consider above, but otherwise

- redirects work
- prior redirects to canada.ca still work
- tests, aside from php, pass

Nice!

infrastructure/conf/nginx-conf-local/default
```diff
@@ -31,6 +32,10 @@ server {
     location = /en/communities/digital-talent { return 301 https://www.canada.ca/en/government/system/digital-government/gcdigital-community.html; }
     location = /fr/communities/digital-talent { return 301 https://www.canada.ca/fr/gouvernement/systeme/gouvernement-numerique/collectivite-gcnumerique.html; }
 
+    # permanent redirect within the same domain with relative path
+    absolute_redirect off;
```
petertgiles, line 36: It looks like [this directive can be used in the context of a location block](http://nginx.org/en/docs/http/ngx_http_core_module.html#absolute_redirect).  It would make me more comfortable if you issued it in these two locations rather than at the top level where it could affect other rules.

petertgiles (approved):
Looks great!  Thanks!

PR #5456, opened 2023-01-24
Bug - 5428 - Fix Search Request Pool Candidates Default Filters

### LINKED ISSUE #5428: 🐛 Filters not populating table initially when opening a single search request
## 🐛 Bug

When you go to the SingleSearchRequest page on the admin dashboard, the Candidates table initially is only filtered by Pool. This means that you have to add all the request filters manually.

## 🦋 Expected Behaviour

When opening up a single search request on the Admin dashboard, the Candidates table should initially have filters set which match the request filters.

## 📋 Steps to Reproduce

1. Submit a request from the search page that includes some optional filters (eg skills, location, EE)
2. Go to the Requests table on the admin side. 
3. Open your newest request.
4. Open the filter modal. Are all the request filters already set?

## 📸 Screenshot

## 🙋‍♀️ Proposed Solution

I think the cause has to do with table state caching. The table is simultaneously trying to initialize its filters with values from the request, and values from url params. It seems like the values from url params (which are empty when it first opens) are winning out. What we actually want is for request values to take precedence, and for them then to be put into url params.

## ✅ Acceptance Criteria

- [ ] When you first open a SingleSearchRequest page, the table filters match the Request filters, without having to set them manually.

esizer:
🤖 Resolves #5428 

## 👋 Introduction

This fixes the default filters being populated on the `PoolCandidatesTable` component.

## 🕵️ Details

We were only populating pools previously, this simply adjusts it to use all filters.

## 🐛 Request Classifications

While fixing this bug, I was able to determine that classifications are not being passed to the request mutation and stored. They appear under "Request Information" because they are coming from the `Pool` rather than the `SearchRequest`.

Is this a bug? Or expected behaviour? If so, should the filter for classifications follow suit and be populated from the pool or should it come from the request when searching?

https://github.com/GCTC-NTGC/gc-digital-talent/blob/e4bc365ba09e14f325c4fe65d469c984396155f5/frontend/common/src/components/SearchRequestFilters/SearchRequestFilters.tsx#L131-L147

## 🧪 Testing

> **Note**
> Classifications are not working due to a possible bug. See the previous section in the description for more info.

1. Build the application `npm run build`
2. Navigate to `/search`
3. Fill out the form, taking note of the filters you use
4. Submit a request
5. Navigate to `/admin/talent-requests`
6. Click "View Request" on the request you created in step 4
7. Scroll down to the  "Candidate Results" section
8. Open the "Filters" dialog
9. Confirm the filters you submitted appear pre-populated in the form

## 📸 Screenshot

Screenshot 2023-01-24 134352

### DIFF
--- a/frontend/admin/src/js/components/poolCandidate/PoolCandidatesTable.tsx
+++ b/frontend/admin/src/js/components/poolCandidate/PoolCandidatesTable.tsx
@@ -236,7 +236,10 @@ const defaultState = {
       hasDiploma: undefined,
       languageAbility: undefined,
     },
-    poolCandidateStatus: [],
+    poolCandidateStatus: [
+      PoolCandidateStatus.PlacedCasual,
+      PoolCandidateStatus.QualifiedAvailable,
+    ],
     priorityWeight: [],
   },
 };
--- a/frontend/admin/src/js/components/poolCandidate/PoolCandidatesTable.tsx
+++ b/frontend/admin/src/js/components/poolCandidate/PoolCandidatesTable.tsx
@@ -258,12 +258,13 @@ const PoolCandidatesTable: React.FC<{
         ...defaultState.filters,
         applicantFilter: {
           ...defaultState.filters.applicantFilter,
-          pools: initialFilterInput?.applicantFilter?.pools,
+          ...initialFilterInput?.applicantFilter,
         },
       },
     }),
     [initialFilterInput],
   );
+
   const [tableState, setTableState] = useTableState<
     Data,
     PoolCandidateSearchInput
--- a/frontend/admin/src/js/components/poolCandidate/PoolCandidatesTable.tsx
+++ b/frontend/admin/src/js/components/poolCandidate/PoolCandidatesTable.tsx
@@ -236,10 +236,7 @@ const defaultState = {
       hasDiploma: undefined,
       languageAbility: undefined,
     },
-    poolCandidateStatus: [
-      PoolCandidateStatus.PlacedCasual,
-      PoolCandidateStatus.QualifiedAvailable,
-    ],
+    poolCandidateStatus: [],
     priorityWeight: [],
   },
 };
@@ -263,6 +260,7 @@ const PoolCandidatesTable: React.FC<{
           ...defaultState.filters.applicantFilter,
           ...initialFilterInput?.applicantFilter,
         },
+        poolCandidateStatus: initialFilterInput?.poolCandidateStatus,
       },
     }),
     [initialFilterInput],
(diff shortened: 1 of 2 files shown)


vd1992:
I think you missed the second acceptance criteria bit about the URL re-write?

as for classification, what I got was the classification loading a different one from the pool

image

esizer:
> I think you missed the second acceptance criteria bit about the URL re-write?

That would be a divergence from our current state persistence functionality. In order to keep the URL clean, we only modify it when the filters do not match the default. changing this behaviour would probably introduce more bugs and be an unwelcome change.

tristan-orourke:
> While fixing this bug, I was able to determine that classifications are not being passed to the request mutation and stored. They appear under "Request Information" because they are coming from the Pool rather than the SearchRequest.

Intended behaviour! Personally I would rather show the Pool name under "Request Information" - I'll propose that to Jerbo, but even if I get a go-ahead on that it would be out of scope for this.

tristan-orourke:
Ah, right, when I suggested adding modifying the url to Acceptance Criteria I forgot how that would interact with our use of default filters. Thanks @esizer!

tristan-orourke (approved):
Looks good! I love a one-line fix 😁

tristan-orourke:
Oh wait - actually, could we add one more thing? Could we add default statuses of Placed Casual and Qualified Available to initialFilterInput in SingleSearchRequestTableApi component?

tristan-orourke:
Or I could make another small ticket for it.

esizer:
> Oh wait - actually, could we add one more thing? Could we add default statuses of Placed Casual and Qualified Available to initialFilterInput in SingleSearchRequestTableApi component?

I'll add it in, seems simple enough 👍

tristan-orourke:
Thanks, but let's only make it the default on the SingleSearchRequest page. I expect they'll want to see everyone on the Pool Candidates page.

esizer:
> Thanks, but let's only make it the default on the SingleSearchRequest page. I expect they'll want to see everyone on the Pool Candidates page.

ohhh, oof. I did not realize we were repurposing that table 😢

esizer:
> > Thanks, but let's only make it the default on the SingleSearchRequest page. I expect they'll want to see everyone on the Pool Candidates page.
> 
> ohhh, oof. I did not realize we were repurposing that table 😢

This should look better 😅 
[only apply default status to search request candidates](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5456/commits/cbf277953902ca24077cf96d01b56db59acaf521)

tristan-orourke:
Thanks, double approved!

PR #5350, opened 2023-01-16
Feature - 31 - GC Notify Integration

esizer:
🤖 Resolves #31

## 👋 Introduction

This presents a solution to send emails via [GC Notify](https://notification.canada.ca/)

## 🕵️ Details

This binds a custom GC Notify Client (`App\Notify\Client`) to Laravel's  [Service Container](https://laravel.com/docs/9.x/container) using a [Service Provider](https://laravel.com/docs/9.x/providers) (`App\Providers\NotifyServiceProvider`) and [Facade](https://laravel.com/docs/9.x/facades) (`App\Facades\Notify`).

### Usage

This only implements a basic solution and a test api routes to show an example of usage. We expose the Notify client as a singleton to be interacted with directly.

```php
<?php

use App\Facades\Notify;

// Note: $repsonse is an instance of \Illuminate\Http\Client\Response

// Email Attachment
$attachment = [
  'file' => $file, // Base64 encoded file
  'filename' => 'file.txt' // Name of the file (can be anything)
];

// Sending email
$response = Notify::sendEmail(
  $to, // Email of receipient
  $templateId, // This is an ID of a template created in GC Notify
  $personalisation, // (optional) An array of template values to be replaced by
  $reference, // (optional) Add a reference key to identify the message
  $replyTo, // (optional) ID for a reply to email address
  $attachment, // (optional) A file attachment for the email
);

// Sending SMS
$response = Notify::sendSms(
  $phoneNumber, // Phone number of receipient
  $templateId, // This is an ID of a template created in GC Notify
  $personalisation, // An array of template values to be replaced by
  $reference, // (optional) Add a reference key to identify the message
  $sender // (optional) ID for a sender number
);

// Bulk Recipient Rows (email)
$rows = [
  ['email address', 'name'],
  ['email+1@domain.com', 'First User'],
  ['email+2@domain.com', 'Second User']
];

// Bulk Recipient Rows (phone)
$rows = [
  ['phone number', 'name'],
  ['+5555555555', 'First User'],
  ['+4444444444', 'Second User']
];

// Sending Bulk Email/SMS
$response = Notify::sendBulk(
  $name, // Unique Indentifier for bulk messages
  $rows, // Array of recipients, first index is headers, following is data
  $templateId, // This is an ID of a template created in GC Notify
  $scheduleFor, // (optional) \Carbon\Carbon date to schedule when messages send
  $replyTo, // (optional) ID for a reply to email address
);
```

### Configuration

[...]

PR #5463, opened 2023-01-24
Tests - 5283 - Replaces `ClearsSchemaCache` with `RefreshesSchemaCache`

### LINKED ISSUE #5283: 🧪 PHP Testing - ClearsSchemaCache deprecated
## 🧪 Tests

The `ClearsSchemaCache` trait is deprecated for PHP testing.

## 🕵️ Details

https://github.com/nuwave/lighthouse/pull/2076

## 🙋‍♀️ Proposed Solution

Seems like `RefreshesSchemaCache` is the preferred solution.

## ✅ Acceptance Criteria

- [ ] PHP tests continue to function normally.
- [ ] Any usage of the ClearsSchemaCache trait is removed.

mnigh:
🤖 Resolves #5283.

## 👋 Introduction

This PR replaces the deprecated Lighthouse PHP trait `ClearsSchemaCache` with `RefreshesSchemaCache`.

## 🧪 Testing

1. Run PHPUnit `docker-compose exec -w /home/site/wwwroot/api webserver sh -c "vendor/bin/phpunit --verbose"`
2. Verify tests run normally

## 📸 Screenshot

Screen Shot 2023-01-24 at 16 17 27

### DIFF
--- a/api/tests/Feature/ApplicantFilterTest.php
+++ b/api/tests/Feature/ApplicantFilterTest.php
@@ -15,7 +15,7 @@ use Database\Seeders\PoolSeeder;
 use Database\Seeders\SkillFamilySeeder;
 use Database\Seeders\SkillSeeder;
 use Illuminate\Foundation\Testing\RefreshDatabase;
-use Nuwave\Lighthouse\Testing\ClearsSchemaCache;
+use Nuwave\Lighthouse\Testing\RefreshesSchemaCache;
 use Nuwave\Lighthouse\Testing\MakesGraphQLRequests;
 use Tests\TestCase;
 
@@ -23,13 +23,13 @@ class ApplicantFilterTest extends TestCase
 {
     use RefreshDatabase;
     use MakesGraphQLRequests;
-    use ClearsSchemaCache;
+    use RefreshesSchemaCache;
 
     protected function setUp(): void
     {
         parent::setUp();
 
-        $this->bootClearsSchemaCache();
+        $this->bootRefreshesSchemaCache();
 
         // Create admin user we run tests as
         // Note: this extra user does change the results of a couple queries
--- a/api/tests/Feature/ApplicantTest.php
+++ b/api/tests/Feature/ApplicantTest.php
@@ -10,7 +10,7 @@ use App\Models\CommunityExperience;
 use App\Models\PersonalExperience;
 use App\Models\WorkExperience;
 use Illuminate\Foundation\Testing\RefreshDatabase;
-use Nuwave\Lighthouse\Testing\ClearsSchemaCache;
+use Nuwave\Lighthouse\Testing\RefreshesSchemaCache;
 use Nuwave\Lighthouse\Testing\MakesGraphQLRequests;
 use Tests\TestCase;
 use Database\Helpers\ApiEnums;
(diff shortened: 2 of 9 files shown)


vd1992 (approved):
no more instances of `ClearsSchemaCache` 👍 

an aside, these two tests fail locally for me
gc notify api key is not in the .env.example, so I assume it is something we don't want exposed to everyone, in which case the tests probably ought to be marked as Skip to avoid having tests relying on secrets

image

esizer:
> an aside, these two tests fail locally for me
> gc notify api key is not in the .env.example, so I assume it is something we don't want exposed to everyone, in which case the tests probably ought to be marked as Skip to avoid having tests relying on secrets

The key is in the example file, I can provide it if you need it 😄 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/c38bd6619801a1a4994bbf4014f6bc3fc4cc218a/api/.env.example#L88-L93

vd1992:
> > an aside, these two tests fail locally for me
> > gc notify api key is not in the .env.example, so I assume it is something we don't want exposed to everyone, in which case the tests probably ought to be marked as Skip to avoid having tests relying on secrets
> 
> The key is in the example file, I can provide it if you need it 😄
> 

I feel like we're misunderstanding one another, I am referring to this in particular. 
https://github.com/GCTC-NTGC/gc-digital-talent/blob/c38bd6619801a1a4994bbf4014f6bc3fc4cc218a/api/.env.example#L89

esizer:
> I feel like we're misunderstanding one another, I am referring to this in particular. This isn't in the example file, intentionally right?

Oh, yes. It is empty intentionally 😜

mnigh:
> an aside, these two tests fail locally for me gc notify api key is not in the .env.example, so I assume it is something we don't want exposed to everyone, in which case the tests probably ought to be marked as Skip to avoid having tests relying on secrets

You are right, the tests related to GC Notify will fail locally if the `GCNOTIFY_API_KEY` key in `api/.env` is missing a valid value. If you want to see it pass, @esizer can provide you with a valid value if you ask him ([related PR](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5350#:~:text=You%20can%20find%20configuration%20in%20the%20config/notify.php%20file.%20It%20is%20required%20to%20have%20an%20environment%20variable%20to%20be%20set%20for%20GCNOTIFY_API_KEY.%20You%20can%20get%20this%20from%20%40esizer.)). In the GitHub workflow, the tests should not fail (though, they will not even run at all as we are waiting on a new version of the setup-php GitHub action to be released with a [fix](https://github.com/shivammathur/setup-php/issues/688)) since there is currently a valid value for `GCNOTIFY_API_KEY` in the repository (or organization) GitHub secrets:

 https://github.com/GCTC-NTGC/gc-digital-talent/blob/badef5102dba52b6a30b3bf0172e24047d5c9d51/.github/workflows/phpunit.yml#L14

So, in conclusion, no need to skip the tests!

petertgiles:
> the tests related to GC Notify will fail locally if the `GCNOTIFY_API_KEY` key in `api/.env` is missing a valid value

That seems weird to me to have tests that we can't run locally, no?

vd1992:
> > the tests related to GC Notify will fail locally if the `GCNOTIFY_API_KEY` key in `api/.env` is missing a valid value
> 
> That seems weird to me to have tests that we can't run locally, no?

I agree

PR #5481, opened 2023-01-25
bug/5296-reset-page-number-filtering

### LINKED ISSUE #5296: 🐛 Table filters do not reset pagination to first page
## 🐛 Bug

In an admin table when applying a filter (e.g. `Removed` status) if the number of results (e.g. only 3 candidates are returned) doesn't reach that page (e.g. click apply filters on page 3) then the table will be empty. (_Originally posted by @yonikid15 in https://github.com/GCTC-NTGC/gc-digital-talent/pull/5292#pullrequestreview-1239249099_)

## 🦋 Expected Behaviour

The current page is updated to 1 after applying any filters.

## 📋 Steps to Reproduce

1. Visit `/admin/pools/:id/pool-candidates`
2. Click on page 2 in pagination
3. Click on **Filters** button
4. Select **Removed** in status field
5. Click **Show Results**

## 📸 Screenshot

https://user-images.githubusercontent.com/22059495/211079394-a31a97e8-7730-44c1-acd2-c11206573c00.mp4

## ✅ Acceptance Criteria

- [ ] Current page is updated to 1 after applying any filters

vd1992:
🤖 Resolves #5296 

## 👋 Introduction

On the API managed tables, updating filtering doesn't reset the current page. This creates odd situations where you might be on page 5 for results that only span 3 pages.

## 🕵️ Details

Updated both instances to reset current page to default upon submit of new filter values.

## 🧪 Testing

1. visit the api managed tables (users and pool candidates)
2. navigate to a later page 
3. submit new filters that would significantly drop results returned
4. confirm you are bumped back to page 1

### DIFF
--- a/frontend/admin/src/js/components/poolCandidate/PoolCandidatesTable.tsx
+++ b/frontend/admin/src/js/components/poolCandidate/PoolCandidatesTable.tsx
@@ -391,7 +391,10 @@ const PoolCandidatesTable: React.FC<{
       }),
     };
 
-    setTableState({ filters: transformedData });
+    setTableState({
+      filters: transformedData,
+      currentPage: defaultState.currentPage,
+    });
   };
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


esizer (approved):
Clean! 🧼

PR #5506, opened 2023-01-26
Hydrogen 2.0.0-beta.39

substrae:
## 👋 Introduction

Updates Hydrogen with a handful of bugfixes and a pinch of love.

## 🕵️ Details

The biggest change here is twofold:
- `data-h2-transition` takes standard CSS syntax now instead of the original comma separated values - this was because commas are used in native CSS transition to allow for transitioning multiple properties by comma separating a list
- No one has probably used this yet, but you used to be forced to use the opposite color modifier in dark mode if automated modifiers were enabled - this was because dark mode was swapping the value of the CSS variable, so in dark mode, the `lighter` variable was being swapped to `darker` automatically for you, but this also meant that using `:dark(lighter)` would spit out the darker color... which is not intuitive, so now, `:dark(lighter)` spits out the lighter color as you'd expect :)
- Small thing, but Hydrogen now tells you what version is running to help create clarity on the currently installed dependency

## 🧪 Testing

1. Update your dependencies using the setup script
2. Run dev or storybook and ensure Hydrogen compiles properly and displays `v2.0.0-beta.39`

### DIFF
--- a/.vscode/hydrogen.code-snippets
+++ b/.vscode/hydrogen.code-snippets
@@ -4,6 +4,11 @@
     "body": ["${1:query}${2::modifiers}(${3:css})"],
     "description": "Adds a query to an existing Hydrogen attribute."
   },
+  "hydrogen-property": {
+    "prefix": "h2-property",
+    "body": ["data-h2-${1:property}=\"${2:query}${3::modifiers}(${4:css})\""],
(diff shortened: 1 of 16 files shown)


esizer (approved):
Hard to see since the transitions are so fast but it generates valid css! 👍

PR #5515, opened 2023-01-26
Debt/5462 rename user roles

### LINKED ISSUE #5462: ♻️ Rename existing User Roles
## ♻️ Debt/Refactor

Rename existing user roles so they do not conflict with the new roles we will be adding with Laratrust (#5434).

## 🕵️ Details

We don't want User.roles to overlap with `roles` function added by Laratrust https://github.com/santigarcor/laratrust/blob/master/src/Traits/LaratrustUserTrait.php.

## 🙋‍♀️ Proposed Solution

Rename the Role enum in schema, the User.roles property in the schema, and the roles column in the User table, and any code anywhere else that references these. Rename to LagacyRole and legacy_roles.

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Database column name updated
- [ ] Schema updated
- [ ] Operations which references user.roles updated
- [ ] AuthorizationContainer continues to work the same as before.

vd1992:
🤖 Resolves #5462 

## 👋 Introduction

Renames database column user.roles to user.legacy_roles. As well as updating everywhere else as needed.
In preparation for Laratrust.

## 🕵️ Details

Lots of places to check, going to need to comb through things to make sure nothing broke anywhere. 
Nothing changed in a complex way, more so lots of things just got touched.  

## 🧪 Testing

1. migrate and seed fresh
2. run through the app and make sure authorization still works
3. tests succeeding, PHP test must be verified locally

### DIFF
--- a/frontend/admin/src/js/components/user/UpdateUser.tsx
+++ b/frontend/admin/src/js/components/user/UpdateUser.tsx
@@ -249,8 +249,8 @@ export const UpdateUserForm: React.FunctionComponent<UpdateUserFormProps> = ({
             />
             <div data-h2-margin="base(x1, 0)">
               <MultiSelectField
-                id="roles"
-                name="roles"
+                id="legacyRoles"
+                name="legacyRoles"
                 label={intl.formatMessage({
                   defaultMessage: "Roles",
                   id: "kwNyl6",
--- a/frontend/admin/src/js/components/user/UpdateUser.tsx
+++ b/frontend/admin/src/js/components/user/UpdateUser.tsx
@@ -19,7 +19,7 @@ import MultiSelectField from "@common/components/form/MultiSelect/MultiSelectFie
 import { useAdminRoutes } from "../../adminRoutes";
 import {
   Language,
-  Role,
+  LegacyRole,
   Scalars,
   UpdateUserAsAdminInput,
   UpdateUserAsAdminMutation,
@@ -35,7 +35,7 @@ type FormValues = Pick<
   | "email"
   | "firstName"
   | "lastName"
-  | "roles"
+  | "legacyRoles"
   | "preferredLang"
   | "preferredLanguageForInterview"
   | "preferredLanguageForExam"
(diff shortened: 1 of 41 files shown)


vd1992:
chromatic deployment
https://www.chromatic.com/pullrequest?appId=61e099a1bb1465003a73d3ce&number=5515
no stories broken (that weren't already having issues)

frontend/admin/src/js/components/user/UpdateUser.tsx
```diff
@@ -262,7 +262,7 @@ export const UpdateUserForm: React.FunctionComponent<UpdateUserFormProps> = ({
                   description:
                     "Placeholder displayed on the user form roles field.",
                 })}
-                options={enumToOptions(Role).map(({ value }) => ({
```
petertgiles, line 265: Could you take another look at this multiselect?  I think it's no longer properly.
image

petertgiles (changes requested):
Very nice!  The authorization container seems to be working well but I think the user editor is broken.

frontend/admin/src/js/components/user/UpdateUser.tsx
vd1992: needed to update the input name/id there, done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/5515/commits/879d85671237b5cbf4dee7636440f96b349f1680

petertgiles: Well, it's initializing properly now but the mutation isn't working yet.  Could you take another look at this page?

vd1992: ah, whoops
mutation also needed to be re-connected as it picks fields out by name and not by formvalues as a whole
done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/5515/commits/6ca00d9c547d2e47b3ef6ae391fb4a15fb88bc1b
create user I think isn't used anymore? but would've had the same issue with input naming but the mutation is fine https://github.com/GCTC-NTGC/gc-digital-talent/pull/5515/commits/ffe8b00b4a128a485060b4bbb555f96f2a65f694

petertgiles: Looks good to me. :+1:

petertgiles (approved):

PR #5518, opened 2023-01-27
Accessibility/3313 fix abbreviations sitewide

### LINKED ISSUE #3313: Screen readers read out abbreviations as words

yonikid15:
🤖 Resolves #3313 

## 👋 Introduction

The goal of this PR is to wrap all classification abbreviations in `<abbr>` tag, and add helper functions to make it easier from now on. After doing research it's important to "...use the element consistently throughout your web page. If you use an abbreviation in one part of the page, make sure to use the `<abbr>` element to define it, and provide the full expansion using the title attribute. This will help to ensure that the user has access to the full meaning of the acronym or abbreviation wherever it is used on the page". (https://webreference.com/html/tags/abbr/)

Talent Search Pages: 
- [x] Homepage
- [x] Search
- [x] Request Page
- [x] Browse Opportunities
- [x] My Applications
- [x] View Pool advertisement
- [x] My Profile
- [x] Government Info Form

Admin: 
- [x] Dashboard
- [x] Pools
- [x] Users* (Multi-Select input will be done in separate issue)
- [x] View User
- [x] Requests
- [x] View Request
- [x] All Candidates* (title in breadcrumbs not spacing correctly(??))
- [x] Classifications

IAP:
- [x] Homepage

## 🕵️ Details

There were a lot of changes to copy so I'm expecting some typos. Please be thorough in your review. 

## 🧪 Testing

- Go through the entire site and make sure all Abbreviations are wrapped in an `<abbr>` tag. They should have a dotted underline and show the full title when hovered over. 
- Ensure that AS classifications are read correctly on screen readers.


petertgiles (commented):
I feel really bad suggesting this because you've made a huge number of substitutions in this branch already...  I don't love the `{ITAbbr}` message variable approach.  :grimacing: 

I'm concerned that either our translators are really going to struggle with it or we're going to have to manually substitute them in for every translation.  The value that goes in there is known so conceptually it doesn't make as much sense to me either.

Did you explore using markup like we do with the frontend/common/src/helpers/format.tsx file?  I'm imagining some more like
```
intl.formatMessage(
  {
    defaultMessage: "Explore <abbreviation>IT</abbreviation> opportunities within the federal government",
    id: "j3WBqJ",
    description: "Button text to submit the Indigenous self-declaration form when not Indigenous.",
  },
  {
    abbreviation: (str) => wrapAbbr(str, abbreviations.it),
  },
)
```

And I'm OK if you disagree with this or want to get a second opinion.  :smile:

tristan-orourke:
I just want to second Peter's suggestion. Something like:
```
intl.formatMessage(
  {
    defaultMessage: "Explore <abbreviation>IT</abbreviation> opportunities within the federal government",
    id: "j3WBqJ",
    description: "Button text to submit the Indigenous self-declaration form when not Indigenous.",
  },
  {
    abbreviation: (str) => wrapAbbr(str, abbreviations.it),
  },
)
```

yonikid15:
I made a lot of changes again so I created a new branch to make it a little easier to review.
https://github.com/GCTC-NTGC/gc-digital-talent/pull/5599

PR #5548, opened 2023-01-31
Feature/5511 teams graphql query

### LINKED ISSUE #5511: ✨ Implement Teams graphql query
## ✨ Feature

Make a query which allows for getting the list of all Teams. Also add some fields to Teams. May want to work on this alongside https://github.com/GCTC-NTGC/gc-digital-talent/issues/5496.

## 🛑  Blocked By
- [x] #5434 

## 🕵️ Details

Create a Team type in schema, with the following definition:
```
type Team {
  name: String
  displayName: LocalizedString
  description: LocalizedString
  departments: Department[]
  contactEmail: Email
}
```

Add new fields to the Teams the in the database: display_name_localized, description_localized, contact_email; and a pivot table for Departments.
Note: the teams table already has display_name and description columns which are not localized, so ensure the attributes in the schema point to the correct columns (either with the `@rename` directive or an accessor on the Team model).

Create a query which returns all Teams, and is protected by a Team.viewAny policy.

## 🎨 Design File

## 📸 Screenshot

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] New fields added to Team model in database and php class
- [ ] Relationship with Departments added to Team and Departments models
- [ ] New Team type added to schema
- [ ] Teams query which returns all teams and is protected by a Team.viewAny policy (which can be implemented with isAdmin or checking view-any-team permission, depending on whether https://github.com/GCTC-NTGC/gc-digital-talent/issues/5438 is done yet)
- [ ] Team Factory exists which fills all fields
- [ ] A phpunit test which tests the query

vd1992:
🤖 Resolves #5511 

## 👋 Introduction

Tweaks the teams table as well as add a query to fetch all teams. 

## 🕵️ Details

Adjusted tables, connect departments with teams, introduced a query, added testing for it, and a new factory. 

## 🧪 Testing

1. ensure migration works (and seed)
2. ensure php tests succeed locally
3. play around with the new query teams
```
query teams {
  teams {
    id
    name
    contactEmail
    displayName {
      en
      fr
    }
    departments {
      id
    }
  }
}
```

4. look out for any potential problems


api/app/Policies/TeamPolicy.php
```diff
@@ -0,0 +1,23 @@
...
+
+    /**
+     * Determine whether the user can view any models.
+     *
+     * @param  \App\Models\User|null  $user
+     * @return \Illuminate\Auth\Access\Response|bool
+     */
+    public function viewAny(User $user = null)
+    {
+        return $user->isAdmin();
```
esizer, line 21: We should probably check if a user exists here before checking if they are admin. Otherwise we get a `Call to a member function isAdmin() on null`

api/app/Models/Team.php
```diff
@@ -25,4 +29,9 @@ class Team extends LaratrustTeam
     ];
 
     public $guarded = [];
+
+    public function departments(): BelongsToMany
+    {
+        return $this->belongsToMany(Department::class, 'team_department');
+    }
```
esizer, line 36: Is this right? A department can belong to more than one team? 🤔 I'm not aware of the requirements but this seems a little odd to me.

esizer (changes requested):
Looking pretty good so far! Just one question and a change 😸

api/app/Policies/TeamPolicy.php
vd1992: I believe that is the role `$user = null` serves, to avoid erroring for guests

it is the same snippet we use elsewhere like

https://github.com/GCTC-NTGC/gc-digital-talent/blob/073d8e1b970fede64705c7eac24b6650d6d3b645/api/app/Policies/DepartmentPolicy.php#L19-L22

esizer: The issue is that policy methods need to return a `boolean` so they can send a proper error message to the client. 
Right now we get 500 internal server error if they are anonymous instead of a proper 401.

vd1992: ah yes, yes, now I think I got it
nested it in https://github.com/GCTC-NTGC/gc-digital-talent/pull/5548/commits/01d0cb1b447cb15f3a705042fc176a44dab6e20e

api/app/Models/Team.php
vd1992: I just mirrored how another pivot table was represented in the model files

esizer: @tristan-orourke What should this relationship look like?

- One to One (how it appears in the table view)
- One to Many (how it appears in the proposed graphql model)
- Many to Many (how it is defined here)

tristan-orourke: I just confirmed with @Jerryescandon - it should be Many to Many. One team can cover multiple depts, and there's no restriction on multiple teams being in the same dept.

esizer (approved):
Awesome! LGTM 🌟

Issue #4362, opened 2022-10-17
♻️ Only `me` query should return the entire User object


tristan-orourke:
## ♻️ Debt/Refactor

We want to remove the need for Admins to access or edit the entire User object. To make this more robust, only the `me` query should return the entire User object. We should redesign other queries to return more limited aliases of the User, like Applicant or UserPublicProfile, depending on what data is required.

UPDATE: it seems like we want to keep the ability for Platform Admins to view almost all user info. If there are exceptions, we can guard those fields with `@can` directives.

## 🕵️ Details

This can be considered part of #4360, as it is part of our overhaul of permissions.

We need to look through the schema and find all cases where a query or mutation returns a full User object, or it is an attribute of another Type. In almost all of those cases, we should return a more limited version of the User instead, like Applicant or UserPublicProfile. This is because the User type includes much private information, including the users applications, any pools they own, their roles, etc. We must replace the type in the backend schema, as well as any frontend operations which call the updated endpoints.

## 🙋‍♀️ Proposed Solution

I propose three different slices/aliases for User:
- UserContactInfo, which would replace UserPublicProfile:
    ```
    type ContactInfo { 
        id: ID!
        email: Email
        firstName: String @rename(attribute: "first_name")
        lastName: String @rename(attribute: "last_name")
        preferredLang: Language @rename(attribute: "preferred_lang")
    }
    ```
- Applicant, which contains everything necessary for considering a candidate to a pool, including ContactInfo.
- The rest of the User object, which includes the previous two types, as well as other relationships:
```
type User {
  id: ID!
  sub: String
  contactInfo: ContactInfo @self
  applicantProfile: Applicant @self
  legacyRoles: [LegacyRole]
  roleAssignments: [RoleAssignment]
  poolCandidates: [PoolCandidate] #this one might not be accessible to platform admins even if the User is
  pools: [Pool] # pools a user owns. We can deprecate this once we're going through roleAssignments.teams
}
```

There are no existing types which include a User or User[] field, so that's good.

Queries/Mutations which return the full User type:
- me // this one makes sense
- user(id: UUID!) // The only frontend place which uses this is the admin UpdateUser page, which actually, we can continue to allow, protected by a view Any User policy.
- users // There is an AllUsers operation defined in the backend, but it doesn't appear to be used.
- usersPaginated // Used for the admin table. Can be replaced with a call which returns ContactInfo.
[...]

PR #5560, opened 2023-02-01
Feature - 5436 - Seed roles permissions

### LINKED ISSUE #5436: ✨ Implement Roles and Permissions
## ✨ Feature

Create the Roles and Permissions we need in a Seeder which can be run in Production.

## 🛑  Blocked By
- [x] #5434 

## 🕵️ Details

Our table of [Roles and Permissions](https://docs.google.com/spreadsheets/d/1RC7wufxwjLEvaEunHsfZfOzBefg218dvvzkDwKSiQjI/edit?usp=sharing).

We want a seeder which sets up all the roles and permissions we want. [Roles and Permissions](https://laratrust.santigarcor.me/docs/7.x/usage/roles-and-permissions.html#setting-things-up) only require a name, though they can optionally have a display name and description.

~QUESTION: we need to finalize the key-like names for our Roles and Permissions, and any Display Name or Description copy we want for them.~

~QUESTION: the description and display_name attributes attached to the default Laratrust models are probably just strings, and we use jsonb for translatable attributes. Should we add translatable versions with another mutation? Should we just use `name` and leave the rest for localizedConstants?~

Use the names in the spreadsheet for roles and permissions. Ignore display_name and description for Permissions. Use the display name and description from the table for Roles. They can be made translatable later.

Laratrust has [tool](https://laratrust.santigarcor.me/docs/7.x/usage/seeder.html#seeder-configuration-file) which can help generate a seeder file, after configuring a less verbose array. I would give it a try, though we might want to modify the resulting seeder file. We probably want to use the updateOrCreate method, indexed on `name`, so the seeder can be run multiple times without issue. 

They're tool says it doesn't support teams, but I think that's fine since we won't be creating the Teams in this seeder.

## 🌎 Localization

Display Names and Descriptions, maybe???

## ✅ Acceptance Criteria

- [ ] The full set of Permissions defined [here](https://docs.google.com/spreadsheets/d/1RC7wufxwjLEvaEunHsfZfOzBefg218dvvzkDwKSiQjI/edit?usp=sharing) exist in the database after seeding
- [ ] After seeding, the Guest, Base User, Applicant, Team Admin, and Super Admin roles exist, with the [correct permissions](https://docs.google.com/spreadsheets/d/1RC7wufxwjLEvaEunHsfZfOzBefg218dvvzkDwKSiQjI/edit?usp=sharing)

esizer:
🤖 Resolves #5436
 
## 👋 Introduction

This adds the `Role` and `Permission` seeders along with a config file for them.

## 🕵️ Details

I believed the [laratrust seeder](https://laratrust.santigarcor.me/docs/7.x/usage/seeder.html#seeder-configuration-file) would not work with teams enabled based on their docs so created a custom one based off of theirs.

## 📋 TO DO

- [x] Add real translations
- [x] Write tests

> **Note**
> All French display names were translated by a 🤖 robot. I will be sending them off to get updated translations.

## 🧪 Testing

1. Run the seeder
2. Inspect the database to check for permissions and roles

> **Note**
> My SQL is pretty bad so I would appreciate if someone could assist writing some sample queries to test that the relationships were created accurately 😆

### DIFF
--- a/api/config/rolepermission.php
+++ b/api/config/rolepermission.php
@@ -1,0 +1,521 @@
+<?php
+
+return [
+    /*
+    |--------------------------------------------------------------------------
+    | Actions
+    |--------------------------------------------------------------------------
+    |
+    | A map of all actions available within the application.
+    |
+    */
+    'actions' => [
+        'v' => 'view',
+        'c' => 'create',
+        'u' => 'update',
+        'd' => 'delete',
+        'as' => 'assign',
+        'arc' => 'archive',
+        's' => 'submit',
+        'p' => 'publish'
+    ],
+
+    /*
+    |--------------------------------------------------------------------------
+    | Scopes
+    |--------------------------------------------------------------------------
+    |
+    | A map of all scopes for different actions performed.
+    |
+    */
+    'scopes' => [
+        'o' => 'own',
+        'a' => 'any',
+        't' => 'team'
+    ],
+
+    /*
+    |--------------------------------------------------------------------------
+    | Resources
+    |--------------------------------------------------------------------------
+    |
+    | A map of all resources that a specific action can be taken on.
+    |
+    */
+    'resources' => [
+        'clss' => 'classification',
+        'dpt' => 'department',
+        'sk' => 'skill',
+        'sf' => 'skillFamily',
+        'usr' => 'user',
+        'ubi' => 'userBasicInfo',
+        'pl' => 'pool',
+        'ppa' => 'publishedPoolAdvertisement',
+        'dp' => 'draftPool',
+        'pcd' => 'poolClosingDate',
+        'app' => 'application',
+        'sapp' => 'submittedApplication',
+        'aprf' => 'applicantProfile',
+        'dapp' => 'draftApplication',
+        'apps' => 'applicationStatus',
+        'ac' => 'applicantCount',
+        'sr' => 'searchRequest',
+        't' => 'team',
+        'tusrbi' => 'teamUsersBasicInfo',
+        'rle' => 'role'
+    ],
+
+    /*
+    |--------------------------------------------------------------------------
+    | Permissions
+    |--------------------------------------------------------------------------
+    |
+    | A map of all permissions with $name => $display_name shape.
+    |
+    */
+    'permissions' => [
+        'view-any-classification' => [
+            'en' => 'View Any Classification',
+            'fr' => 'Afficher toute classification'
+        ],
+        'create-any-classification' => [
+            'en' => 'Create Any Classification',
+            'fr' => 'Créer n\'importe quelle classification'
+        ],
+        'update-any-classification' => [
+            'en' => 'Update Any Classification',
+            'fr' => 'Mise à jour de toute classification'
+        ],
+        'delete-any-classification' => [
+            'en' => 'Delete Any Classification',
+            'fr' => 'Supprimer toute classification'
+        ],
+
+        'view-any-department' => [
+            'en' => 'View Any Department',
+            'fr' => 'Voir n\'importe quel département'
+        ],
+        'create-any-department' => [
+            'en' => 'Create Any Department',
+            'fr' => 'Créer n\'importe quel département'
+        ],
+        'update-any-department' => [
+            'en' => 'Update Any Department',
+            'fr' => 'Mise à jour de n\'importe quel département'
+        ],
+        'delete-any-department' => [
+            'en' => 'Delete Any Department',
+            'fr' => 'Supprimer un département'
+        ],
+
+        'view-any-skill' => [
+            'en' => 'View Any Skill',
+            'fr' => 'Afficher n\'importe quelle compétence'
+        ],
+        'create-any-skill' => [
+            'en' => 'Create Any Skill',
+            'fr' => 'Créer n\'importe quelle compétence'
+        ],
+        'update-any-skill' => [
+            'en' => 'Update Any Skill',
+            'fr' => 'Mettre à jour n\'importe quelle compétence'
+        ],
+        'delete-any-skill' => [
+            'en' => 'Delete Any Skill',
+            'fr' => 'Supprimer toute compétence'
+        ],
+
+        'view-any-skillFamily' => [
+            'en' => 'View Any Skill Family',
+            'fr' => 'Voir n\'importe quelle famille de compétences'
+        ],
+        'create-any-skillFamily' => [
+            'en' => 'Create Any Skill Family',
+            'fr' => 'Créer n\'importe quelle famille de compétences'
+        ],
+        'update-any-skillFamily' => [
+            'en' => 'Update Any Skill Family',
+            'fr' => 'Mettre à jour n\'importe quelle famille de compétences'
+        ],
+        'delete-any-skillFamily' => [
+            'en' => 'Delete Any Skill Family',
+            'fr' => 'Supprimer toute famille de compétences'
+        ],
+
+        'view-any-user' => [
+            'en' => 'View Any User',
+            'fr' => 'Afficher tout utilisateur'
+        ],
+        'view-any-userBasicInfo' => [
+            'en' => 'View Any User Basic Info',
+            'fr' => 'Afficher les informations de base de tout utilisateur'
+        ],
+        'view-own-user' => [
+            'en' => 'View Own User',
+            'fr' => 'Voir son propre utilisateur'
+        ],
+        'update-any-user' => [
+            'en' => 'Update Any User',
+            'fr' => 'Mise à jour de tout utilisateur'
+        ],
+        'update-own-user' => [
+            'en' => 'Update Own User',
+            'fr' => 'Mettre à jour son propre utilisateur'
+        ],
+        'delete-any-user' => [
+            'en' => 'Delete Any User',
+            'fr' => 'Supprimer un utilisateur'
+        ],
+
+        'view-team-pool' => [
+            'en' => 'View Team Pool',
+            'fr' => 'Voir le pool d\'équipes'
+        ],
+        'view-any-publishedPoolAdvertisement' => [
+            'en' => 'View Any Published Pool Advertisement',
+            'fr' => 'Voir toute annonce publiée sur la piscine'
+        ],
+        'create-team-pool' => [
+            'en' => 'Create Team Pool',
+            'fr' => 'Créer un pool d\'équipes'
+        ],
+        'update-team-draftPool' => [
+            'en' => 'Update Team Draft Pool',
+            'fr' => 'Mise à jour du pool de sélection des équipes'
+        ],
+        'publish-team-pool' => [
+            'en' => 'Update Team Pool',
+            'fr' => 'Mise à jour du pool d\'équipes'
+        ],
+        'update-team-poolClosingDate' => [
+            'en' => 'Update Team Pool Closing Date',
+            'fr' => 'Mise à jour de la date de clôture de la réserve d\'équipes'
+        ],
+        'delete-team-draftPool' => [
+            'en' => 'Delete Team Draft Pool',
+            'fr' => 'Supprimer le pool de sélection d\'équipes'
+        ],
+
+        'view-own-application' => [
+            'en' => 'View Own Application',
+            'fr' => 'Afficher sa propre application'
+        ],
+        'view-team-submittedApplication' => [
+            'en' => 'View Team Submitted Application',
+            'fr' => 'Voir la demande soumise par l\'équipe'
+        ],
+        'view-team-applicantProfile' => [
+            'en' => 'View Team Applicant Profile',
+            'fr' => 'Voir le profil du candidat de l\'équipe'
+        ],
+        'create-own-draftApplication' => [
+            'en' => 'Create Own Draft Application',
+            'fr' => 'Créer son propre projet d\'application'
+        ],
+        'submit-own-application' => [
+            'en' => 'Submit Own Application',
+            'fr' => 'Soumettre sa propre demande'
+        ],
+        'update-team-applicationStatus' => [
+            'en' => 'Update Team Application Status',
+            'fr' => 'Mise à jour du statut de la demande de l\'équipe'
+        ],
+        'delete-own-draftApplication' => [
+            'en' => 'Delete Own Draft Application',
+            'fr' => 'Supprimer son propre projet de demande'
+        ],
+        'archive-own-submittedApplication' => [
+            'en' => 'Archive Own Submitted Application',
+            'fr' => 'Archiver sa propre demande'
+        ],
+        'view-any-applicantCount' => [
+            'en' => 'View Applicant Count',
+            'fr' => 'Voir le nombre de candidats'
+        ],
+
+        'create-any-searchRequest' => [
+            'en' => 'Create Any Search Request',
+            'fr' => 'Créer toute demande de recherche'
+        ],
+        'view-team-searchRequest' => [
+            'en' => 'Create Team Search Request',
+            'fr' => 'Créer une demande de recherche d\'équipe'
+        ],
+        'update-team-searchRequest' => [
+            'en' => 'Update Team Search Request',
+            'fr' => 'Mise à jour de la demande de recherche d\'équipe'
+        ],
+        'delete-team-searchRequest' => [
+            'en' => 'Delete Team Search Request',
+            'fr' => 'Supprimer une demande de recherche d\'équipe'
+        ],
+
+        'view-any-team' => [
+            'en' => 'View Any Team',
+            'fr' => 'Voir n\'importe quelle équipe'
+        ],
+        'view-any-teamUsersBasicInfo' => [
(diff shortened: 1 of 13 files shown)


api/config/rolepermission.php
```diff
@@ -0,0 +1,521 @@
+<?php
+
```
tristan-orourke, line 2: Could you add a comment explaining this is used by RolePermissionSeeder, and a link to the Laratrust Documentation this is based on?

api/config/rolepermission.php
```diff
@@ -0,0 +1,521 @@
...
+        'create-any-team' => [
+            'en' => 'Create Any Team',
+            'fr' => 'Créer n\'importe quelle équipe'
+        ],
+        'update-any-team' => [
+            'en' => 'Update Any Team',
+            'fr' => 'Mise à jour de n\'importe quelle équipe'
+        ],
+        'update-team-team' => [
+            'en' => 'Update Team Team',
```
mnigh, line 274: Will folks using this know what this means? Is there another way to write this?

api/config/rolepermission.php
```diff
@@ -0,0 +1,521 @@
...
+        'view-any-team' => [
+            'en' => 'View Any Team',
+            'fr' => 'Voir n\'importe quelle équipe'
+        ],
+        'view-any-teamUsersBasicInfo' => [
+            'en' => 'View Any Team Users Basic Info',
+            'fr' => 'Afficher les informations de base sur les utilisateurs de n\'importe quelle équipe'
+        ],
+        'view-team-teamUsersBasicInfo' => [
+            'en' => 'View Teams Team Users Basic Info',
```
mnigh, line 262: Will folks using this know what this means? Is there another way to write this?

api/config/rolepermission.php
```diff
@@ -0,0 +1,521 @@
...
+    |           'scope' => [ 'actions' ]
+    |       ]
+    |   ]
+    |
+    */
+    'seeders' => [
+        'guest' => [
+            'clss' => [
+                'a' => ['v']
+            ],
```
tristan-orourke, line 388: Let's use the full names here, instead of the short keys. Like:
```
            'classification' => [
                'any' => ['view']
            ],
```
 I think it will be a lot more readable. And should be easy now with a find-replace.

api/config/rolepermission.php
```diff
@@ -0,0 +1,521 @@
...
+return [
+    /*
+    |--------------------------------------------------------------------------
+    | Actions
+    |--------------------------------------------------------------------------
+    |
+    | A map of all actions available within the application.
+    |
+    */
+    'actions' => [
```
mnigh, line 12: Would it not make sense to use the same number of characters for each key for consistency? Maybe at least two characters in case another action is added in the future.

tristan-orourke (changes requested):
I have a couple suggestions for readability to start with.

api/config/rolepermission.php
```diff
@@ -0,0 +1,521 @@
...
+
+    /*
+    |--------------------------------------------------------------------------
+    | Resources
+    |--------------------------------------------------------------------------
+    |
+    | A map of all resources that a specific action can be taken on.
+    |
+    */
+    'resources' => [
```
mnigh, line 45: Would it not make sense to use the same number of characters for each key for consistency? Maybe at least three characters in case another action is added in the future.

api/config/rolepermission.php
```diff
@@ -0,0 +1,521 @@
...
+        'update-any-skill' => [
+            'en' => 'Update Any Skill',
+            'fr' => 'Mettre à jour n\'importe quelle compétence'
+        ],
+        'delete-any-skill' => [
+            'en' => 'Delete Any Skill',
+            'fr' => 'Supprimer toute compétence'
+        ],
+
+        'view-any-skillFamily' => [
```
mnigh, line 128: Is there a specific reason why a combination of cases (kebab-case and camelCase) is used for these keys instead of just kebab-case?

api/config/rolepermission.php
```diff
@@ -0,0 +1,521 @@
...
+                'en' => 'Applicant',
+                'fr' => 'Candidat'
+            ],
+            'description' => [
+                'en' => 'Can edit their own profile and apply to jobs.',
+                'fr' => 'Ils peuvent modifier leur propre profil et postuler à des emplois.'
+            ]
+        ],
+
+        'team_admin' => [
```
mnigh, line 347: Could this be just `admin` or are there plans to add other types of admin roles (other than `super_admin`)?

api/config/rolepermission.php
```diff
@@ -0,0 +1,521 @@
...
+    | A map of all roles with the shape:
+    |
+    |   $name => [
+    |       'display_name' => LocalizedString,
+    |       'description' => LocalizedString,
+    |   ]
+    |
+    */
+    'roles' => [
+        'guest' => [
```
mnigh, line 314: How could someone have any role if they are not logged in?

mnigh (commented):
Maybe this is too late for this PR, but for some simple predefined role names, I like what [WordPress](https://wordpress.org/documentation/article/roles-and-capabilities/#summary-of-roles) has done and find it easy to understand. Some of the language might not apply to this project, but might be helpful in keeping roles to one word.

tristan-orourke:
> I would appreciate if someone could assist writing some sample queries to test that the relationships were created accurately

Rather than write sql queries, I would suggest writing some PHP tests which [assign and check](https://laratrust.santigarcor.me/docs/6.x/usage/roles-and-permissions.html#setting-things-up) roles/permissions in the way we plan to do so in the app:

1. seed a user
2. assign a role
3. check if they have a permission
4. remove the role
5. check again

Then, check with [teams](https://laratrust.santigarcor.me/docs/6.x/usage/teams.html):
1. seed a user
2. assign a role with a team: `$user->attachRole($admin, $team);`
3. check if they have a team permission in that team context:  `$user->isAbleTo('view-team-pool', 'my-awesome-team');`
4. check if they have a team permission in a different team context: ` $user->isAbleTo('view-team-pool', 'other-lame-team');`
5. check that any-scoped permissions associated with that role still work: `$user->isAbleTo('view-any-userBasicInfo`);`

api/config/rolepermission.php
esizer: I'm all for suggestions! Display name is a nullable value but I though these has some value so added them. Honestly, I was just writing the keys out as is because of the number of them I needed to convert.

I would love some input on better names for any of the permissions.

api/config/rolepermission.php
esizer: I'm going to take the [suggestion to use full names](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5560#discussion_r1093699187) here.

api/config/rolepermission.php
tristan-orourke: Ahaha, I'm not sure what else to use as a key without making the general action-scope-resource nomenclature worse, but I'm open to suggestions. The display name can definitely be better. "Update my Teams"?

api/config/rolepermission.php
esizer: This is to avoid confusion. We have the pattern of `{action}-{scope}-{resource}`. So the hyphens are acting like spaces between camel case names. It just so happens that resources are the only multi-word names that require camel case.

api/config/rolepermission.php
tristan-orourke: "View the BasicInfo of Users in my Teams" is my more descriptive attempt.

If you're interested, I could use some feedback on https://github.com/GCTC-NTGC/gc-digital-talent/issues/4362, which also deals with some how to break down (and name) different versions of `User`.

api/config/rolepermission.php
tristan-orourke: It clearly demarcates the three sections (action, scope, resource) of each permission

api/config/rolepermission.php
esizer: I may be wrong on this, but this role is supposed to act like a façade to assign anonymous permissions, if the current user is null, we use this role to determine what they can do. 

We are also not using permission inheritance, rather assigning multiple roles to add permissions. So a user would have:

 - Guest
 - Base User
 - Applicant

And all permissions that come with that.

tristan-orourke: Possibly like this:
```
if ($user) {
  return $user->isAbleTo('view-something');
} else {
  Role::where('name', 'guest')->first()->hasPermission('view-something');
}
```
And if not that, then just for organizing and recording our own understanding of permissions.

tristan-orourke:
@mnigh I'm very open to suggestions on the role names!

api/config/rolepermission.php
esizer: Add docs in [update config with docs and new keys](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5560/commits/68fe9b46e66ce421900acf0b5a55799886c1bf77)

api/config/rolepermission.php
esizer: Updated the maps to work more like enums for readability in [use full words instead of abbrs for stubs](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5560/commits/d3fd102751dbe14e9927ac5bf4a83b231d04f103)

api/config/rolepermission.php
mnigh: 1. Is it likely that users will require multiple roles and if so, why?
2. Is there a scenario where having multiple roles that contain conflicting permissions could occur?

tristan-orourke: 1.Yes. For example, it is likely that a user with a high-level role (eg for creating teams) may also be able to act as a lower-level role (eg for creating pools in a team). As an alternative, we _could_ design high-level roles to be strictly a superset of low-level roles (all the same permissions, plus more) and give every user only one role, but that is more restrictive and less flexible.
2. Having more roles gives you more permissions in a purely additive way, so I don't think there's any way they could conflict.

esizer:
> Rather than write sql queries, I would suggest writing some PHP tests which [assign and check](https://laratrust.santigarcor.me/docs/6.x/usage/roles-and-permissions.html#setting-things-up) roles/permissions in the way we plan to do so in the app:

Excellent idea! Added some tests in [add role and permissions test](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5560/commits/40fe58987280d1d5e40cd30fffc771d4924dbbe3)

api/tests/Feature/RolePermissionTest.php
```diff
@@ -0,0 +1,195 @@
...
+     *
+     * @return void
+     */
+    public function test_applicant_role()
+    {
+        $applicantRole = Role::where('name', 'applicant')->sole();
+        $this->user->attachRole($applicantRole);
+
+        $this->assertTrue($this->user->hasRole('applicant'));
+        $this->assertTrue($this->user->isAbleTo([
```
tristan-orourke, line 97: This will return true if the user has any one of these permissions. Did you mean to check test if they have _all_ the permissions? If so, you can add `true` as a second arg to isAbleTo.
https://laratrust.santigarcor.me/docs/7.x/usage/roles-and-permissions.html#checking-for-roles-permissions

esizer: hehe, good catch! Yes, I wanted to check all 😅

api/config/rolepermission.php
```diff
@@ -0,0 +1,532 @@
...
+                'own' => ['create', 'delete']
+            ],
+            'submittedApplication' => [
+                'own' => ['archive']
+            ]
+        ],
+
+        'team_admin' => [
+            'user' => [
+                'any' => ['view']
```
tristan-orourke, line 467: I don't think team_admin should have any `any-user` permissions.

tristan-orourke: Oh, I think this is supposed to be userBasicInfo, not user

tristan-orourke (changes requested):
After playing around with roles a bit locally, I think I was wrong about [teams_strict_check](https://laratrust.santigarcor.me/docs/7.x/usage/teams.html#checking-roles-permissions). 😅 Now I think we have to change it to false if we want to give team-scoped roles _any_ any-scoped permissions. Please double-check my logic, but if I'm correct lets change the config in this PR.

I'll push a commit soon which updates permission Display Names.

api/tests/Feature/RolePermissionTest.php
esizer: okay, i added strict checks and the tests are still passing [add strict permission check to test](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5560/commits/4d0ba14bd69c545d4a53b7770c897364ac5f0fae)

api/config/rolepermission.php
esizer: That makes sense, updated in [fix team admin permissions](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5560/commits/f1cec7c1c76cef7ab4a0f9361e3f9e3664217582)

esizer:
> After playing around with roles a bit locally, I think I was wrong about [teams_strict_check](https://laratrust.santigarcor.me/docs/7.x/usage/teams.html#checking-roles-permissions). 😅 Now I think we have to change it to false if we want to give team-scoped roles _any_ any-scoped permissions. Please double-check my logic, but if I'm correct lets change the config in this PR.

Lol, im so confused now. I think you are right...maybe? 😅 It is passing with `true` and `false` so it's hard to tell 🤔 Maybe ill add a null team check as well to see the difference.

api/tests/Feature/RolePermissionTest.php
```diff
@@ -0,0 +1,195 @@
...
+            'update-team-team',
+            'view-any-role',
+            'assign-team-role',
+        ];
+
+        $this->assertTrue($this->user->hasRole('team_admin',  $this->ownedTeam));
+        $this->assertTrue($this->user->isAbleTo($permissionsToCheck, $this->ownedTeam, true));
+
+        $this->assertFalse($this->user->hasRole('team_admin',  $this->unownedTeam));
+        $this->assertFalse($this->user->isAbleTo($permissionsToCheck, $this->unownedTeam));
```
tristan-orourke, line 145: So basically what I'm worried about, in regards to the team_check_strict, is that we won't have a team to pass in when we check `view-any-role`. Like this:
```suggestion
        $teamPermissionsToCheck = [
            'view-any-userBasicInfo',
            'view-team-pool',
            'create-team-pool',
            'publish-team-pool',
            'update-team-draftPool',
            'delete-team-draftPool',
            'update-team-poolClosingDate',
            'view-team-submittedApplication',
            'view-team-applicantProfile',
            'update-team-applicationStatus',
            'view-team-searchRequest',
            'update-team-searchRequest',
            'delete-team-searchRequest',
            'view-team-teamMembers',
            'update-team-team',
            'assign-team-role',
        ];
        $otherPermissionsToCheck = [
            'view-any-role',
        ];

        $this->assertTrue($this->user->isAbleTo($otherPermissionsToCheck, null, true));

        $this->assertTrue($this->user->hasRole('team_admin',  $this->ownedTeam));
        $this->assertTrue($this->user->isAbleTo($teamPermissionsToCheck, $this->ownedTeam, true));

        $this->assertFalse($this->user->hasRole('team_admin',  $this->unownedTeam));
        $this->assertFalse($this->user->isAbleTo($teamPermissionsToCheck, $this->unownedTeam));
```
Here's the trouble... I thought this would fail with teams_strict_check=true, and pass with teams_strict_check=false. But I've tested this locally, changing the config file, and it seems to fail either way! 😧

esizer: haha yeah, that is part of my confusion! It seems to return the same value regardless of what you put in the strict check arg 😖

esizer: Essentially when it is:

-  `true` you _must_ pass in the team to the role/permission check if it comes from a team other wise, it will check permissions that are _not_ associated with that team
- `false` means it does not care if the permission is attached to a team or not, it will return true no matter if that role/permission is part of a team or not and doesn't even care which team that role/permission comes from

tristan-orourke: > It seems to return the same value regardless of what you put in the strict check arg 

I think you have to change teams_strict_check in the config file, I don't think there's an arg for it. The third arg to isAbleTo is about whether to do an ALL or ANY operation on arrays of permissions.

esizer: > > It seems to return the same value regardless of what you put in the strict check arg
> 
> I think you have to change teams_strict_check in the config file, I don't think there's an arg for it. The third arg to isAbleTo is about whether to do an ALL or ANY operation on arrays of permissions.

Oh yeah, poor word choice on my part. I was updating config. I was running into some permission issues on WSL where the config cache was not updating. I'm pretty sure I have it sorted out now.

Do we want it to be `false` still? This might cause problems when we start to mix team and non-team roles. We will need to be careful to _always_ pass a team in when checking elevated team permissions since a check with a `null` team checks all teams for that permission rather than all permissions that are not associated with a team.

tristan-orourke (approved):

PR #5561, opened 2023-02-01
Feature/5435 connect pools teams

### LINKED ISSUE #5435: ✨ Connect Pools to Teams
## ✨ Feature

Every Pool must have a belongsTo relationship to a single [Team](https://laratrust.santigarcor.me/docs/7.x/the-basics/models/team.html).

## 🛑 Blocked By
- [x] #5513
- [x] #5511 

## 🕵️ Details

Create a relationship between Teams and Pools. Its `belongsTo` on Pools, and `hasMany` on Teams. 

Existing pools must be attached to the Team created in #5513.

## ✅ Acceptance Criteria

- [ ] Pools have a belongsTo relationship to Teams
- [ ] Teams have a hasMany relationship with Pools
- [ ] Migration ensures all existing Pools are connected to the DCM pool.
- [ ] A Team Factory and seeder which adds a couple more random teams to database
- [ ] Pool seeder / factory attaches a team

vd1992:
🤖 Resolves #5435 

## 👋 Introduction

Connects teams and pools.

## 🕵️ Details

Migration, and updating model/schema. 
Touches factory and seeding too. 

## 🧪 Testing

1. migrate and seed off main before hopping to this branch
2. ensure migration works, that existing pools were attached to the dcm team
3. explore the database in adminer, and play around with values or seed fresh
4. query teams and see pools
```
query teams {
  teams {
    id
    name
    pools {
      id
    }
  }
}
```
5. ensure nothing /admin side broke (pools)

## 💻 Deployment

Migrations run automatically, however, this one is supposed to run after TeamSeeder.php has been called.
It duplicates the purpose of that seeder to ensure that the migration will be a success, but uses updateOrCreate to avoid creating duplicate teams.


vd1992:
for the time being, team is an optional field on the pool type
making it required breaks the pool type everywhere, so changing it to required might be something we want to do later but will be a little messy

api/database/migrations/2023_02_01_194314_connect_teams_pools.php
```diff
@@ -0,0 +1,63 @@
...
+        $dcmIdQuery = DB::select('SELECT id FROM teams WHERE name = :name', ['name' => 'digital-community-management']);
+        $dcmId = $dcmIdQuery[0]->id;
+
+        DB::statement(
+            <<<SQL
+                UPDATE pools
+                    SET team_id = :dcmId
+            SQL, [
+                'dcmId' => $dcmId,
+            ]);
```
esizer, line 45: I think we can use eqloquent here to:

```php
  Pool::all()
    ->team()
    ->save($dcmId);
```

api/database/migrations/2023_02_01_194314_connect_teams_pools.php
```diff
@@ -0,0 +1,63 @@
...
+            [
+                'display_name' => [
+                    'en' => 'Digital Community Management',
+                    'fr' => 'Gestion de la collectivité numérique',
+                ],
+            ],
+        );
+
+        Schema::table('pools', function (Blueprint $table) {
+            $table->uuid('team_id')->references('id')->on('teams')->nullable()->default(null);
```
esizer, line 33: I think we should have some on delete logic here in case a team is deleted. Unless we want to handle this in the controller?

```php
$table->uuid('team_id')
  ->references('id')
  ->on('teams')
  ->nullable()
  ->default(null)
  ->cascadeOnDelete();
```

api/database/seeders/PoolSeeder.php
```diff
@@ -25,6 +26,7 @@ public function run()
                 ],
                 'key' => 'digital_careers',
                 'user_id' => User::where('email', 'admin@test.com')->first()->id,
+                'team_id' => Team::where('name', 'digital-community-management')->sole()['id'],
```
esizer, line 29: For consistency, can we access the `id` property with small arrow syntax?

```suggestion
                'team_id' => Team::where('name', 'digital-community-management')->sole()->id,
```

esizer (changes requested):
Very cool! I just made some suggestions to lean on eloquent over accessing the DB facade directly 😄

api/database/migrations/2023_02_01_194314_connect_teams_pools.php
```diff
@@ -0,0 +1,63 @@
...
+        });
+
+        $dcmIdQuery = DB::select('SELECT id FROM teams WHERE name = :name', ['name' => 'digital-community-management']);
+        $dcmId = $dcmIdQuery[0]->id;
```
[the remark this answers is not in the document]
vd1992: honestly, just a whim (I feel like writing this bit in SQL this time)
is there any preference of what to use for these?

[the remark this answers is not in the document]
vd1992: okay, https://github.com/GCTC-NTGC/gc-digital-talent/pull/5561/commits/962c19e480eb1cff4a3a9b6c40abcff53bcd6188

api/database/migrations/2023_02_01_194314_connect_teams_pools.php
vd1992: https://github.com/GCTC-NTGC/gc-digital-talent/pull/5561/commits/962c19e480eb1cff4a3a9b6c40abcff53bcd6188

api/database/seeders/PoolSeeder.php
vd1992: just mimicked the line above instead https://github.com/GCTC-NTGC/gc-digital-talent/pull/5561/commits/883ecee07fdc65dc74ec0d7acc1625eed579b34e

api/tests/Feature/TeamsTest.php
```diff
@@ -26,6 +27,8 @@ protected function setUp(): void
     $newUser->sub = 'admin@test.com';
     $newUser->legacy_roles = ['ADMIN'];
     $newUser->save();
+
+    DB::table('teams')->delete(); // clear teams created in migrations before testing
```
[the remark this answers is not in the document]
vd1992: https://github.com/GCTC-NTGC/gc-digital-talent/pull/5561/commits/883ecee07fdc65dc74ec0d7acc1625eed579b34e

api/database/migrations/2023_02_01_194314_connect_teams_pools.php
vd1992: I don't think we do this anywhere else, so I'd rather spin off a discussion as to whether we want to implement this across the database

api/database/migrations/2023_02_01_194314_connect_teams_pools.php
vd1992: it didn't like that method, so I went back to the tried and true 🤷 
https://github.com/GCTC-NTGC/gc-digital-talent/pull/5561/commits/139c3d633dd8db903c3dea6cd1e6710cda7e8b46

vd1992:
a problem was created where pool creation would be broken due to a missing team to connect it to, two solutions came to mind
- make the team_id field nullable for some time, big drawback of potential pools falling through the crack if created after migrating
- stopgap of adding on the dcm team id in the pool advertisement creation behind the scenes, drawback ???

I chose to go with the latter option in https://github.com/GCTC-NTGC/gc-digital-talent/pull/5561/commits/8eecd8a33cd67abc1ecfe0ee8030dcb0e92dd9ff

once teams has matured to the point where non-dcm pools can be created, the auto-attach to dcm can be removed probably easily enough (pluck teamId as you would userId from the ME)

api/database/migrations/2023_02_01_194314_connect_teams_pools.php
esizer: Sorry, `all()` returns a collection so you would need to loop through them 😅 

```php
Pool::all()
  ->each(function ($pool) use ($dcmId) {
    $pool->team()->save($dcmId);
  });
```

Another option would to just set it as the default 🤔 Since we are removing the default later on.

```php
$dcmId = Team::where('name', 'digital-community-management')
    ->first()->id;

Schema::table('pools', function (Blueprint $table) use ($dcmId) {
    $table->uuid('team_id')->references('id')->on('teams')->nullable()->default($dcmId);
});
```

I just don't think it is good practice to rely on SQL. Eloquent is a safer way of writing these statements and has a lot of stuff happening in the background related to compatibility with postrgres and other SQL servers (in the event we decide to replace it in the future)

vd1992: I see, well, it is good to have a standard to use
wow, the set default directly was a way better idea thanks, https://github.com/GCTC-NTGC/gc-digital-talent/pull/5561/commits/8559089a2213c9bf70000b5702229ffe91261ffa

vd1992: actually nevermind lol
setting it as a default that require a migration to reverse down the line seems like it'd create an annoyance for someone, bugs possibly

lets not https://github.com/GCTC-NTGC/gc-digital-talent/pull/5561/commits/c22dc06ccf82edc04d180c5681642b43174a1073

esizer (approved):

vd1992:
@tristan-orourke 
could you please take a quick look at the files changed, make sure nothing causes you concern?

api/database/migrations/2023_02_01_194314_connect_teams_pools.php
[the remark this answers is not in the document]
tristan-orourke: In general, we prefer leaning on eloquent because it should theoretically make it harder to screw up the sql 😄

tristan-orourke:
Okay, some general advice on using Eloquent vs manually-written sql in migrations:
1. In general, we should prefer using Eloquent methods to writing any SQL. Theoretically, this should make it easier to avoid making mistakes in our SQL. We should only write out the SQL for complex cases where Eloquent is not sufficient, or much harder to read/write.
2. However, its also a good idea to _not_ use our Eloquent Models in _migrations_ specifically, because in the future the Model and its relationships may have changed, and we still need the migration to work when spinning up a fresh database!
3. So: a good compromise is to use statements `DB::table("pools")->get()` instead of `Pools::get()`, but to continue to use eloquent methods like `->where()`, `->get()`, `->update()`. 

Note: because you're interacting with the table directly in this case, not the Model, you won't be able to use Relationship magic, and will need to get linked rows by selecting on foreign-ids.

vd1992:
> Okay, some general advice on using Eloquent vs manually-written sql in migrations:
> 
>     2. However, its also a good idea to _not_ use our Eloquent Models in _migrations_ specifically, because in the future the Model and its relationships may have changed, and we still need the migration to work when spinning up a fresh database!
> 
>     3. So: a good compromise is to use statements `DB::table("pools")->get()` instead of `Pools::get()`, but to continue to use eloquent methods like `->where()`, `->get()`, `->update()`.
> 
> 
> Note: because you're interacting with the table directly in this case, not the Model, you won't be able to use Relationship magic, and will need to get linked rows by selecting on foreign-ids.

okay, yeah, feels reasonable that there is some downside to relying on model files 

I feel like my earlier snippet 
```
        DB::statement(
            <<<SQL
                UPDATE pools
                    SET team_id = :dcmId
            SQL, [
                'dcmId' => $dcmId,
            ]);
```

does the job while being very straightforward
should I swap it back in?

api/database/migrations/2023_02_01_194314_connect_teams_pools.php
tristan-orourke: Looks like Eloquent supports mass updates: https://laravel.com/docs/9.x/eloquent#mass-updates

tristan-orourke:
> should I swap it back in?
I'd prefer something like this, does this work?
```
DB::table('pools')->update(['team_id' => $dcmId]);
```

vd1992:
> > should I swap it back in?
> 
> I'd prefer something like this, does this work?
> 
> ```
> DB::table('pools')->update(['team_id' => $dcmId]);
> ```

yes that appears to work fine, thanks!
https://github.com/GCTC-NTGC/gc-digital-talent/pull/5561/commits/0b24d5c8a719456133ba31c6cb0b1a020ce408fa

@esizer 
changed the migration in the commit above
will merge after waiting a bit

api/graphql/schema.graphql
```diff
@@ -521,6 +523,7 @@ type Team {
   description: LocalizedString
   departments: [Department] @belongsToMany
   contactEmail: Email @rename(attribute: "contact_email")
+  pools: [Pool] @hasMany
```
tristan-orourke, line 526: This is a potential security leak, since a user who has permission to view a Team may not have permissions to view Pools. (Remember, we generally use Pool as a highly-restricted type, with PoolAdvertisement as the public version.) 
This may likely get changed during the course of the Roles and Teams epic, but for now, the quickest solution would be to slap `@can(ability: "viewAny")` on Team.pools field, as well as on any `team`/`teams` fields which appear in the schema.

tristan-orourke (changes requested):

api/graphql/schema.graphql
vd1992: okay, slapped the directive on in a couple places https://github.com/GCTC-NTGC/gc-digital-talent/pull/5561/commits/6fca42cdb5b89926cdc5af6f06768bfe48fb346c

tristan-orourke (approved):
Thanks, that was it

PR #5601, opened 2023-02-03
Feature - 5459 - Support empty request submissions

### LINKED ISSUE #5459: ✨ (Search page) Submit empty requests
## ✨ Feature

DCM can really benefit from knowing what Managers are looking for in their original requests, specially if their original request returns zero available candidates, this information can guide DCM to future pool advertisements. We want to allow some ways of capturing this information by creating a way to allow Managers to submit empty requests.

## 🕵️ Details

We have created several interventions, each one might require it's own ticket, but I will list them all here:

1.  Different state for the sticky box on the right. This box is the first indicator for (visual) users to know whether there are any candidates that match their request. We want to create a state in case the counter reaches 0, that lets managers know they can still submit their request as is for additional help. 

image

2. Change the interaction in the results from a link to a submit button 

### OLD
image

### NEW
image

3. Empty Request Content: We should let both the manager and DCM know when a request was created empty originally. Since candidates move a lot in and out of pools and their status change all the time, we should capture when a request was originally created as an empty request, regardless wether or not there are matching candidates at the moment its being viewed. To this effect, we want to add a line that gets passed in the request as a string if the request is empty.

image

image

## ✅ Acceptance Criteria

- [ ] Different state for the sticky box on the right with zero results
- [ ] Change the interaction in the results at the bottom from a link to a submit button
- [ ] On request page, different empty Request Content
- [ ] Ensure that the admin table and SSR page function OK with empty fields like pool-id on the request

esizer:
🤖 Resolves #5459 

## 👋 Introduction

This allows users to submit requests, even if no candidates were found.

## 🕵️ Details

We now store empty requests with the `was_empty` column in the database. This is past tense since new candidates could appear so it **_only_** says if the request was empty on the initial request. At least, this was my interpretation of the ticket description "DCM can really benefit from knowing what Managers are looking for in their original requests, specially if their original request returns zero available candidates,"

## 📋 TO DO

- [x] Add translations

## 🧪 Testing

1. Refresh the API `docker-compose run --rm maintenance bash refresh_api.sh`
2. Build the application `npm run build`
3. Navigate to `/search`
4. Fill out the filters so that there are no candidates
5. Confirm the estimated candidates box and candidate results box both have the new content
6. Confirm you can continue on to the request form
7. Confirm you can submit the request form
8. Login to admin and navigate to `/admin/talent-requests`
9. View the request you submitted in step 7
10. Confirm the message about no candidates appears at the top of the page
11. Confirm the classifications appear along with all pools matching

### Search Form

Screenshot 2023-02-03 122500

### Request Form

Screenshot 2023-02-03 122524

### Admin Search Request Page

Screenshot 2023-02-03 122535

### DIFF
--- a/api/database/migrations/2023_02_03_170212_add_empty_column_requests_table.php
+++ b/api/database/migrations/2023_02_03_170212_add_empty_column_requests_table.php
@@ -14,7 +14,7 @@ return new class extends Migration
     public function up()
     {
         Schema::table('pool_candidate_search_requests', function (Blueprint $table) {
-            $table->boolean('was_empty')->default(false)->after('pool_candidate_filter_id');
+            $table->boolean('was_empty')->default(false);
         });
     }
 
--- a/api/database/migrations/2023_02_03_170212_add_empty_column_requests_table.php
+++ b/api/database/migrations/2023_02_03_170212_add_empty_column_requests_table.php
@@ -1,0 +1,32 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('pool_candidate_search_requests', function (Blueprint $table) {
+            $table->boolean('was_empty')->default(false)->after('pool_candidate_filter_id');
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('pool_candidate_search_requests', function (Blueprint $table) {
+            $table->dropColumn('was_empty');
+        });
+    }
+};
--- a/apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
+++ b/apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
@@ -342,8 +342,8 @@ export const RequestForm: React.FunctionComponent<RequestFormProps> = ({
               <p data-h2-margin="base(x.5, 0, 0, 0)">
                 {intl.formatMessage({
                   defaultMessage:
-                    "If you are submitting a form that had zero estimated candidates, let us know moe about this request in the comments.",
-                  id: "Y79y47",
+                    "If you are submitting a form that had zero estimated candidates, let us know more about this request in the comments.",
+                  id: "adM1fA",
                   description:
                     "Instructions to provide additional details when submitting a request with no candidates",
                 })}
--- a/apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
+++ b/apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
@@ -337,6 +337,18 @@ export const RequestForm: React.FunctionComponent<RequestFormProps> = ({
                   "Blurb before additional comments textarea in the request form.",
               })}
             </p>
+            {candidateCount === 0 ? (
+              <p data-h2-margin="base(x.5, 0, 0, 0)">
+                {intl.formatMessage({
+                  defaultMessage:
+                    "If you are submitting a form that had zero estimated candidates, let us know moe about this request in the comments.",
+                  id: "Y79y47",
+                  description:
+                    "Instructions to provide additional details when submitting a request with no candidates",
+                })}
+              </p>
+            ) : null}
+
             <TextArea
               id="additionalComments"
               name="additionalComments"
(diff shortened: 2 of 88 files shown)


petertgiles (commented):
I built this and ran it through.  It seems to work well and the `was_empty` is an elegant solution to changing the copy.  Despite the gymnastics, the original classification selection is lost, though, right?  There's just no spot in the database to store it and this is supposed to a temporary solution?

esizer:
> I built this and ran it through. It seems to work well and the `was_empty` is an elegant solution to changing the copy. Despite the gymnastics, the original classification selection is lost, though, right? There's just no spot in the database to store it and this is supposed to a temporary solution?

Yeah, when we show the classification it is coming from the pool and not the request itself. I'm not sure if that is another ticket or within scope of this? I believe it would be important to store the requests classification regardless of what pool is attached to requests.

@tristan-orourke Do you have any thoughts about this and how we handle requests?

esizer:
> For now, one option could be to attach _all_ the pools which match classification and stream?

Hmm, that is possible. Though, it kind of shifts the acceptance criteria a bit from the original "Ensure that the admin table and SSR page function OK with empty fields like pool-id on the request" Since the purpose of the PR is to allow an empty pool rather than adding multiple pools 🤔

petertgiles:
From scrum discussion: assign empty requests to all pools with matching classification.

esizer:
Okay, I have updated this to [add all pools that have a matching classification to a search request](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5601/commits/47ee84b8e1faee823087c527d9f4a38aee229d10). Also, it [only shows unique classification](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5601/commits/f48a887a1e5dcea42875b6477290e6235c3bb218) instead of repeating the same item.

@petertgiles + @tristan-orourke

api/database/migrations/2023_02_03_170212_add_empty_column_requests_table.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('pool_candidate_search_requests', function (Blueprint $table) {
+            $table->boolean('was_empty')->default(false)->after('pool_candidate_filter_id');
```
vd1992, line 17: just as a heads up, I don't think `after(...)` does anything for postgres
column ordering as a whole isn't pleasant in it [see this](https://stackoverflow.com/questions/285733/how-do-i-alter-the-position-of-a-column-in-a-postgresql-database-table)

image

esizer: oh you are right, that is **_super_** annoying lol

vd1992: there is no table organization allowed here

apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
```diff
@@ -337,6 +338,18 @@ export const RequestForm: React.FunctionComponent<RequestFormProps> = ({
                   "Blurb before additional comments textarea in the request form.",
               })}
             </p>
+            {candidateCount === 0 ? (
+              <p data-h2-margin="base(x.5, 0, 0, 0)">
+                {intl.formatMessage({
+                  defaultMessage:
+                    "If you are submitting a form that had zero estimated candidates, let us know moe about this request in the comments.",
```
vd1992, line 345: "let us know moe"? 😛 
probably need to check the translation too

vd1992:
need a different blurb on admin side, one that doesn't request more information from you in the comments

image

esizer:
> need a different blurb on admin side, one that doesn't request more information from you in the comments

Seems weird, but that is actually what it is supposed to say:

214392335-f28f5bb1-31b3-4ac8-ab04-b15ac430c649

apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
esizer: weird, i totally fixed that. must have forgot to commit lol

[fix typo](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5601/commits/5bdbb8df80813a5e9e43b6d6123d8cf62e795556)

petertgiles:
Issue #5644 created to directly associate classifications and job streams to a request.

petertgiles (approved):
Nice, this seems to work well for me.  I noticed that when I reviewed my empty request in admin I got a candidate available in the table.  I'm going to chalk that up to the search page returning users and the admin page returning candidates and our seeded data sometimes being unrealistic.

PR #5602, opened 2023-02-03
a11y - 5466 - Focus Input on Error Link Click

### LINKED ISSUE #5466: Form inputs - field not focused when clicking error message
Clicking on the field name in the error block scrolls to the correct field but the field is not automatically focused so that input can begin.

Context: split off from #5416

### How to reproduce
1. Environment - any
2. Which browser and/or Assistive Technology?  I'm using Firefox without AT, not sure about Fable
3. Step-by-step instructions explaining how to reproduce the problem.
  1. Log in as a new user
  2. Leave the Create Account form blank and try to submit
  4.  Find the error block and click the name of the first name field
  5. Observe that the field has been scrolled to but the focus is not in that field.

### Questions
- What does it mean for a checkbox fieldset to be focused?
- Answer: focus the input itself regardless of input type

### Acceptance Criteria

When a field name in an error block has been clicked:
- [ ] The field is scrolled to
- [ ] The field is focused for interaction

esizer:
🤖 Resolves #5466 

## 👋 Introduction

This focuses the input when the link to the input is clicked within the error summary.

## 🧪 Testing

> **Note**
> For radio groups, you may need to temporarily set the default value to `undefined` to properly force an error.

2. Navigate to the profile
3. Purposefully create some errors on every form input type
4. Submit the form and confirm the error summary appears
5. Click a link to the field in the error summary
6. Confirm focus is moved to the input and you can immediately interact with that input

### DIFF
--- a/apps/web/src/pages/Auth/CreateAccountPage/CreateAccountPage.tsx
+++ b/apps/web/src/pages/Auth/CreateAccountPage/CreateAccountPage.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { useLocation, useNavigate } from "react-router-dom";
+import { useLocation, useNavigate, useSearchParams } from "react-router-dom";
 import { useIntl } from "react-intl";
 
 import Hero from "@common/components/Hero/Hero";
@@ -286,10 +286,10 @@ type LocationState = {
 
 const CreateAccount: React.FunctionComponent = () => {
   const intl = useIntl();
-  const location = useLocation();
   const navigate = useNavigate();
   const paths = useRoutes();
-  const { from } = location.state as LocationState;
+  const [searchParams] = useSearchParams();
+  const from = searchParams.get("from");
   const authContext = React.useContext(AuthorizationContext);
   const meId = authContext.loggedInUser?.id;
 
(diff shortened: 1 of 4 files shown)


yonikid15 (changes requested):
Nice work Eric, just found an issue with the CreateAccount form. However the forms on the profile seems to be working fine 👍

Here's a clip of what I'm seeing on the Create Account page. Instead of scrolling to the input it's trying to actually link to a new page:

https://user-images.githubusercontent.com/22059495/217020033-3e493245-716b-4c22-93ed-e2bdfe1e7dfb.mp4

The href and the input id's are matching up though.
image

image

esizer:
> Nice work Eric, just found an issue with the CreateAccount form. However the forms on the profile seems to be working fine 👍
> 
> Here's a clip of what I'm seeing on the Create Account page. Instead of scrolling to the input it's trying to actually link to a new page:

Great catch! Updated to persist the `from` state in the link so we don't lose it when scrolling 👍 

[persist location state in scroll to link](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5602/commits/0ef1e95f5739996dceb6a9761a5f5ad0fbf0fc7a)

yonikid15 (approved):
Nice fix looks like all the forms are working as expected now 🥇

PR #5605, opened 2023-02-03
Bug - 5541 - Fixes breadcrumbs in Application context

### LINKED ISSUE #5541: Breadcrumbs broken for application sub pages
## 🐛 Bug

When visiting the profile sub pages as part of an application the URL navigation works a bit differently to ensure that the page logic is aware of the application context.  Unfortunately the breadcrumbs aren't handling this well.

## 🦋 Expected Behaviour

The breadcrumbs should be aware of the application context and update accordingly

## 📋 Steps to Reproduce

1. Start an application
2. Visit the skills and experiences sub page as part of that applicatin
3. Test the segments of the breadcrumbs

## 📸 Screenshot
The _Step 1_ link is using the user ID instead of the application ID so it _404_s

The _Experience and skills_ link loses the application context and so you are ejected out of your application

## ✅ Acceptance Criteria

- [ ] Each segment of the breadcrumbs navigates to the expected locations in the subpages of the profile

mnigh:
🤖 Resolves #5541.

## 👋 Introduction

This PR fixes breadcrumbs for profile forms so that they link correctly in an Application context.

## 🕵️ Details

Also, `returnRoute` was standardized to use **applicationId** from `searchParams` instead of from a prop as it had been in some cases.

## 🧪 Testing

Test all segments of the breadcrumbs, specifically the following ones.

### Application context

1. Start an application
2. Click **Edit _Section Name_**  
3. Click on the last breadcrumb in the list
4. Verify the page is the same and the breadcrumbs are in an application context
5. Click on the **Step 1** breadcrumb in the list
6. Verify the page returns to the **My application profile** page
7. Repeat for all sections

### Profile context

1. Click **My Profile**
2. Click **Edit _Section Name_**  
3. Click on the last breadcrumb in the list
4. Verify the page is the same and the breadcrumbs are in a profile context
5. Repeat for all sections

### DIFF
--- a/apps/web/src/pages/Profile/EmploymentEquityPage/components/EmploymentEquityForm/EmploymentEquityForm.tsx
+++ b/apps/web/src/pages/Profile/EmploymentEquityPage/components/EmploymentEquityForm/EmploymentEquityForm.tsx
@@ -11,6 +11,7 @@ import ProfileFormWrapper, {
   ProfileFormFooter,
 } from "~/components/ProfileFormWrapper/ProfileFormWrapper";
 
+import { useSearchParams } from "react-router-dom";
 import EquityOptions from "./EquityOptions";
 import type { EmploymentEquityUpdateHandler, EquityKeys } from "../../types";
 
@@ -29,6 +30,11 @@ const EmploymentEquityForm: React.FC<EmploymentEquityFormProps> = ({
 }) => {
   const intl = useIntl();
   const paths = useRoutes();
+  const [searchParams] = useSearchParams();
+  const applicationId = searchParams.get("applicationId");
+  const applicationParam = applicationId
+    ? `?applicationId=${applicationId}`
+    : ``;
   const returnRoute = application
     ? paths.reviewApplication(application.id)
     : paths.profile(user.id);
@@ -59,7 +65,16 @@ const EmploymentEquityForm: React.FC<EmploymentEquityFormProps> = ({
         },
         {
           label: intl.formatMessage(navigationMessages.stepOne),
-          url: paths.reviewApplication(application.id),
+          url: paths.reviewApplication(applicationId ?? ""),
+        },
+        {
+          label: intl.formatMessage({
+            defaultMessage: "Diversity, equity and inclusion",
+            id: "pGTTrp",
+            description:
+              "Display Text for Diversity, equity and inclusion Page",
+          }),
+          url: `${paths.diversityEquityInclusion(user.id)}${applicationParam}`,
         },
       ]
     : [];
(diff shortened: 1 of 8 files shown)


apps/web/src/pages/Profile/EmploymentEquityPage/components/EmploymentEquityForm/EmploymentEquityForm.tsx
```diff
@@ -11,6 +11,7 @@ import ProfileFormWrapper, {
   ProfileFormFooter,
 } from "~/components/ProfileFormWrapper/ProfileFormWrapper";
 
+import { useSearchParams } from "react-router-dom";
```
esizer, line 14: nitpicky but, can we ensure third party imports are at the top 😅

apps/web/src/pages/Profile/EmploymentEquityPage/components/EmploymentEquityForm/EmploymentEquityForm.tsx
```diff
@@ -29,8 +30,13 @@ const EmploymentEquityForm: React.FC<EmploymentEquityFormProps> = ({
...
-  const returnRoute = application
-    ? paths.reviewApplication(application.id)
+  const [searchParams] = useSearchParams();
+  const applicationId = searchParams.get("applicationId");
+  const applicationParam = applicationId
+    ? `?applicationId=${applicationId}`
+    : ``;
+  const returnRoute = applicationId
+    ? paths.reviewApplication(applicationId)
     : paths.profile(user.id);
```
esizer, line 40: I'm tempted to say we make this a custom hook since this seems to be a patter we use in every profile form 🤔 Maybe like:

```tsx
const { param, id } = useCurrentApplication();
```
This would be out of scope for this PR but if you agree we could add a ticket.

esizer (changes requested):
Awesome! One super minor change for import order.

apps/web/src/pages/Profile/EmploymentEquityPage/components/EmploymentEquityForm/EmploymentEquityForm.tsx
mnigh: see: [Move third party import to top](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5605/commits/219f019d33eb2a0545bdb5e7940c282e47300e9d)

apps/web/src/pages/Profile/EmploymentEquityPage/components/EmploymentEquityForm/EmploymentEquityForm.tsx
mnigh: I thought about the same thing when doing this but wanted to stay within the scope. Feel free to create a new issue with the proposed solution 😄

esizer (approved):
Great! 🌞

apps/web/src/pages/Profile/EmploymentEquityPage/components/EmploymentEquityForm/EmploymentEquityForm.tsx
esizer: Added the issue, hopefully I summarized it well 😅 

https://github.com/GCTC-NTGC/gc-digital-talent/issues/5609

Issue #5352, opened 2023-01-16
🐛 Bug Report


petertgiles:
## 🐛 Bug

If you submit to a pool advertisement while not logged in the auto-login-and-apply logic in frontend/talentsearch/src/js/components/pool/CreateApplication.tsx submits the application twice and an error toast appears for the second application.

## 🦋 Expected Behaviour

The application should be submitted once and so no error toast should show.

## 🕵️ Details

I suspect this is due to the createApplication function living in a useEffect and Strict Mode fires useEffects twice.

- **Operating System**: Linux Pop_OS 22.04
- **Browser**: Firefox 108

## 📋 Steps to Reproduce

1. Start "not logged in"
2. Browse to a pool application and apply
3. Log in as usual
4. Observe a toast that the application was created and a second toast that there was an error

## 📸 Screenshot

![image](https://user-images.githubusercontent.com/8978655/212772932-6e80b99d-7be8-491a-920d-73d53aed545f.png)

## 🙋‍♀️ Proposed Solution

Some logic will need to be added to CreateApplication component to not send the mutation more than once, even when useEffect is run.  Additionally, the new navigate function of React Router will probably need to be moved out of the mutation promise and into a useEffect as well.

## ✅ Acceptance Criteria

- [ ] Applying to a pool while logged in fires the mutation to create the application only once
- [ ] Applying to a pool while not logged in while log in as usual and then fires the mutation to create the application only once

esizer:
Based on the discussion had on https://github.com/GCTC-NTGC/gc-digital-talent/pull/5614, we should be okay to move forward with converting this button to an [action ](https://reactrouter.com/en/main/route/action) and showing a login button that returns a user to the pool advertisement rather than automatically creating it.

If we do not have access to the `urql` client in our router already the PR mentioned **_should_** have the required info to do it 👍

PR #5607, opened 2023-02-03
Visit login-info during login redirect

### LINKED ISSUE #5378: Apply for this process should go to register/login-info
A screen reader user wouldn’t know to select the [Register link] if they didn’t catch it in the main utility navigation region on the Job Poster. As a result, they would not understand why they are directed to GC Key before they can apply to a recruitment/pool. 

### How to reproduce
1. Access the Home Page using the JAWS, NVDA or Narrator in Windows 10 or 11.
2. Skip the main navigation utility.
3. Activate the  Browse Jobs Page 
4. Apply to a recruitment 
5. Apply for the process
6. Understand why you have been ushered to the GC Key Welcome Page.
## ✔ Expected behaviour (fix)
Users are directed to the ["Register using GC Key"](https://uat-talentcloud.tbs-sct.gc.ca/en/register-info) or Log in (Sign in) using GC Key page when they activate the “Apply for this process” button  in a job poster. This will help them understand why they are being directed to GC Key. 
### Acceptance Criteria
A set of assumptions which, when tested, verify that the accessibility requirement was met.
- [ ] Users are directed to the ["Register using GC Key"](https://uat-talentcloud.tbs-sct.gc.ca/en/register-info) or the "Log in (Sign in) using GC Key" page
- [ ] After login, users are returned to the new application like before

petertgiles:
🤖 Resolves #5378 

## 👋 Introduction

This PR updates the router so that any protected route that requires auth will go to `/login-info` first.

## 🕵️ Details

This works by updating the RequireAuth component that guards protected routes.  Rather then navigating directly to the API login endpoint it navigates to login-info with a search param that indicates the eventual destination.  Login-info then passes that search param to the API login endpoint, which passes it to the auth provider, which passes it back to our app.  :dizzy: 

I've updated the RequireAuth component to take the login path as a prop since that route exists in `apps/web` and RequireAuth is in `common`.

I don't think we need to worry about an open redirect vulnerability here.  Any route that is protected with RequireAuth is effectively whitelisted by its presence in our router.  

## 🧪 Testing

Suggested test steps:
1. Directly logging in as a new user goes to create-account then to the profile as usual
2. Directly logging in as an existing user goes to the profile as usual
3. Not logged in, navigating in the browser to a protected path like `/en/admin/user` goes to login-info then goes to the admin users page after successful login
4. Not logged in, going to a pool advertisement, and applying goes to login-info, then starts the application after successful log in.
5. Logged in, navigating to a protected route works as usual

> **Warning**
> There is an outstanding bug being tracked in #5352 where creating a new application will trigger twice and show an error toast.

## 📸 Screenshot

image


esizer:
> I've updated the RequireAuth component to take the login path as a prop since that route exists in `apps/web` and RequireAuth is in `common`.

lol maybe we should consider moving `useRoutes` to a package when we split up common so it can be used anywhere in the projest 😅

petertgiles:
> maybe we should consider moving `useRoutes` to a package when we split up common

I dunno, this change seems OK to me.  RequireAuth could be useful on its own but our routes aren't really meaningful outside the app.  But, I don't doubt we'll want to rethink how common is being used and what's in there after the move.

petertgiles:
This is making me a bit crazy... I can't replicate the build failures locally.  Any ideas?

esizer:
> This is making me a bit crazy... I can't replicate the build failures locally. Any ideas?

If this latest run fails too, I'll take a look

apps/web/src/pages/Auth/LoginPage/LoginPage.tsx
```diff
@@ -9,6 +9,7 @@ import SEO from "@common/components/SEO/SEO";
 
 import useRoutes from "~/hooks/useRoutes";
 import useBreadcrumbs from "~/hooks/useBreadcrumbs";
+import { useSearchParams } from "react-router-dom";
```
esizer, line 12: a tiny nitpick but, can we move this up with the other external dependency imports? 😅

esizer (approved):
Awesome! ⭐

apps/web/src/pages/Auth/LoginPage/LoginPage.tsx
petertgiles: Yeah, for sure.
Lol, I just realized that there was a convention in how these imports were grouped.  :male_detective:

PR #5612, opened 2023-02-03
Feature/5585 create update team mutation

### LINKED ISSUE #5585: ✨ Implement createTeam & updateTeam mutation
## ✨ Feature

Implement mutations which allow for creating and updating Teams.

## 🕵️ Details

I don't think the updateTeam mutation needs to allow for adding members at the same. If it were to, remember that changing member roles may require different permissions than editing Team data.

Fields that can be edited on a team: name, display name (localized), description (localized), departments (probably a sync operation).

Create and Update Team mutations should be protected by TeamPolicy methods, which can check for isAdmin LegacyRole initially.

## 🎨 Design File

## 📸 Screenshot

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] API call exists to create new Team with name, display_name, description, and departments and contact_email
- [ ] API call exists to edit Team name, display_name, description, and departments and contact_email
- [ ] Admins can call these mutations, other users cannot
- [ ] PHPunit tests to ensure mutations work when called through graphql

vd1992:
🤖 Resolves #5585 

## 👋 Introduction

Add createTeam and updateTeam mutations.

## 🕵️ Details

Adds the new mutations, with policies, as well as unit testing.
I also threw in a query to grab a specific team since I imagine we'll need it soon. 

## 🧪 Testing

1. play around with the new queries/mutations

```
query team($id: UUID!) {
  team(id: $id) {
    id
    name
  }
}
```

```
mutation createTeam($team: CreateTeamInput!) {
  createTeam(team: $team) {
        name
        id
        displayName {
           en
           fr
         }
         description {
           en
           fr
         }
         contactEmail
         departments {
           id
         }
  }
}
```

```
mutation updateTeam($id: UUID!, $team: UpdateTeamInput!) {
     updateTeam(id: $id, team: $team) {
        name
        id
        displayName {
           en
           fr
         }
         description {
           en
           fr
         }
         contactEmail
         departments {
           id
         }
     }
}
```

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -1310,23 +1310,23 @@ input DepartmentBelongsTo {
   connect: ID!
 }
 
-input DepartmentsBelongsToMany {
-  sync: [ID]
+input DepartmentBelongsToMany {
+  sync: [ID!]
 }
 
 input CreateTeamInput {
   name: String!
-  displayName: LocalizedString @rename(attribute: "display_name")
-  description: LocalizedString
-  departments: DepartmentsBelongsToMany
+  displayName: LocalizedStringInput @rename(attribute: "display_name")
+  description: LocalizedStringInput
+  departments: DepartmentBelongsToMany
   contactEmail: Email @rename(attribute: "contact_email")
 }
 
 input UpdateTeamInput {
   name: String
-  displayName: LocalizedString @rename(attribute: "display_name")
-  description: LocalizedString
-  departments: DepartmentsBelongsToMany
+  displayName: LocalizedStringInput @rename(attribute: "display_name")
+  description: LocalizedStringInput
+  departments: DepartmentBelongsToMany
   contactEmail: Email @rename(attribute: "contact_email")
 }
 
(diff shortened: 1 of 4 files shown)


api/graphql/schema.graphql
```diff
@@ -970,6 +970,7 @@ type Query {
   skills: [Skill]! @all
   genericJobTitle(id: UUID! @eq): GenericJobTitle @find
   genericJobTitles: [GenericJobTitle]! @all
+  team(id: UUID! @eq): Team @find @can(ability: "viewAny")
```
esizer, line 973: For find queries like this, we should use the `view` policy so we can determine if the user can view this specific team. Right now, it will effectively be the same but when we implement the `view-team-team` and `view-own-team` permissions it will become important

esizer (changes requested):
Cool! Just one minor suggestion on the policy for getting a single team 😄

esizer (approved):

api/graphql/schema.graphql
esizer: I doubt the `e2e` tests failing are related but I re-ran to be safe

Issue #5361, opened 2023-01-17
Explore using urql without the hooks


petertgiles:
Since we have migrated to react-router, we may be able to use this as an opportunity to research using urql without the hooks so we can utilize loaders to simplify our page routes.

## :question: Spike Questions
- [ ] How much effort would it be to use urql without the hooks?
- [ ] What are the pros and cons of using urql without the hooks?

Timebox: 1.5 days

petertgiles:
Eric has already completed the research on this topic.

Issue #4545, opened 2022-11-01
When clicking Apply when not logged in, unclear to user that anything is happening for several seconds


tristan-orourke:
When I click "Apply for this process" when I'm not logged in, I get redirected to /create-application. But (on UAT) it took several seconds for the redirect to bring me to the login page. During that time, it looked to me as a user like the page was simply ignoring my click.

Acceptance Criteria:
 - [ ] Using dev tools to simulate a slow connection 
 - [ ] Clicking "Apply for this process" when not logged in shows a loading message/spinner almost immediately until the login page loads
 - [ ] Ensure the experience is clear for screen readers, too

tristan-orourke:
May be fixed by #5352? Blocking this for now, we can try to reproduce later.

tristan-orourke:
Alright, I'm closing this for now then.

PR #5614, opened 2023-02-04
Experiment - 5352 - Stabilize Create Application

esizer:
> **Note**
> For now, this is just an experiment to see what people think. It has some hardcoded values and could use a really good refactor to clean things up.

## 👋 Introduction

This doesn't fix the bug where applications are submitted twice when navigating to the create application page directly. 

## 🕵️ Details

 - #5352 This does not prevent the double application for some reason even with no `useEffect` 😖 
 - #4545 Possibly solve this but unsure 🤷‍♀️ 
 - #5361 Is one option to use `urql` in loaders without hooks

## ❗ Problem

I'm wondering if this should be moved to an `action` and we remove some of the magic around application creation. 

For example, we don't even show the button to apply and instead tell the user to login to apply and bring them back to the advertisement after logging in. We could then move the logic currently in the loader to an `action` and convert the button to a `Form` with hidden inputs for `poolId` and `userId`. Since it is in an action, there is less magic and it should stop the mutation running twice.

### Pass Client to Router

We start by getting the client and passing it into our router creation so we can access it in our loaders.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/4f46166f16ada0dd59785b2db361cb5d2235078a/apps/web/src/components/Router.tsx#L1175-L1177

### Create Loader with Client

Then, we can use a wrapper function to create a loader function that has access to the `client` and current `userId`.

> **Note**
> We don't really need to pass the `userId` here, we could include it in the params as well.

REF: [loader](https://reactrouter.com/en/main/route/loader)

https://github.com/GCTC-NTGC/gc-digital-talent/blob/4f46166f16ada0dd59785b2db361cb5d2235078a/apps/web/src/pages/CreateApplicationPage/CreateApplicationPage.tsx#L30-L32

### Authorize Operation

We can check if the user exists and boot them over to login if they do not. This means, we no longer need the `RequireAuth` wrapper for this.

> **Note**
> There is [an RFC to implement middleware to add context to a loaders](https://github.com/remix-run/react-router/discussions/9564). When this is released, we can check for auth there to reduce this type of boilerplate.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/4f46166f16ada0dd59785b2db361cb5d2235078a/apps/web/src/pages/CreateApplicationPage/CreateApplicationPage.tsx#L36-L50

### Create the Application

Once we know the user exists, we can create the application and `defer` the data. 

REF: [Deferred Data](https://reactrouter.com/en/main/guides/deferred)

https://github.com/GCTC-NTGC/gc-digital-talent/blob/4f46166f16ada0dd59785b2db361cb5d2235078a/apps/web/src/pages/CreateApplicationPage/CreateApplicationPage.tsx#L52-L60

### Get the Data

We retrieve the deferred data in the page component (essentially anything returned from the loader).

https://github.com/GCTC-NTGC/gc-digital-talent/blob/4f46166f16ada0dd59785b2db361cb5d2235078a/apps/web/src/pages/CreateApplicationPage/CreateApplicationPage.tsx#L70

### Render the Page

Here we are rendering the page. Since this is essentially middleware to create an application, we only ever render one of three things:

- Loading spinner while the deferred data resolves
- Redirect to pool advertisement on error
- Redirect to application on success

> **Note**
>  Since we are using partial errors with `graphql` the `createApplication` mutation returns the existing application (if it does) so we can redirect to it even with errors 😄 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/4f46166f16ada0dd59785b2db361cb5d2235078a/apps/web/src/pages/CreateApplicationPage/CreateApplicationPage.tsx#L82-L104

### No More API Wrappers

If we follow a similar pattern for all pages, we would no longer need those API wrappers and can essentially move all of that code to `loaders` and `actions`.

## 🧪 Testing

2. Ensure you are logged out
3. Navigate to a pool advertisement
4. Start an application
5. Note that the mutation still runs twice regardless of no `useEffect`

### DIFF
--- a/apps/web/src/components/Router.tsx
+++ b/apps/web/src/components/Router.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { createBrowserRouter, RouterProvider } from "react-router-dom";
+import { createBrowserRouter, Outlet, RouterProvider } from "react-router-dom";
 
 import { Locales } from "@common/helpers/localize";
 import Loading from "@common/components/Pending/Loading";
@@ -15,6 +15,12 @@ import IAPLayout from "~/components/Layout/IAPLayout";
 import { TalentRedirect, ProfileRedirect } from "~/components/Redirects";
 import CreateAccountRedirect from "~/pages/Auth/CreateAccountPage/CreateAccountRedirect";
 import { LegacyRole } from "~/api/generated";
+import { Client, useClient } from "urql";
+
+import useAuthorizationContext from "@common/hooks/useAuthorizationContext";
+import { PossibleUser } from "@common/components/Auth/AuthorizationContainer";
+
+import { loader as applicationLoader } from "~/pages/CreateApplicationPage/CreateApplicationPage";
 
 /** Home */
 const HomePage = React.lazy(() =>
@@ -209,6 +215,7 @@ const PoolAdvertisementPage = React.lazy(() =>
       ),
   ),
 );
+
 const CreateApplicationPage = React.lazy(() =>
   lazyRetry(
     () =>
@@ -496,7 +503,7 @@ const ViewSearchRequestPage = React.lazy(() =>
   ),
 );
 
-const createRoute = (locale: Locales) =>
+const createRoute = (locale: Locales, client: Client, user: PossibleUser) =>
   createBrowserRouter([
     {
       path: `/`,
@@ -733,6 +740,7 @@ const createRoute = (locale: Locales) =>
                         },
                         {
                           path: "create-application",
+                          loader: applicationLoader(client, user?.id),
                           element: (
                             <RequireAuth roles={[LegacyRole.Applicant]}>
                               <CreateApplicationPage />
@@ -1168,7 +1176,9 @@ const createRoute = (locale: Locales) =>
 
 const Router = () => {
   const { locale } = useLocale();
-  const router = createRoute(locale);
+  const client = useClient();
+  const { loggedInUser } = useAuthorizationContext();
+  const router = createRoute(locale, client, loggedInUser);
   return <RouterProvider router={router} fallbackElement={<Loading />} />;
 };
 
--- a/apps/web/src/components/Router.tsx
+++ b/apps/web/src/components/Router.tsx
@@ -741,11 +741,7 @@ const createRoute = (locale: Locales, client: Client, user: PossibleUser) =>
                         {
                           path: "create-application",
                           loader: applicationLoader(client, user?.id),
-                          element: (
-                            <RequireAuth roles={[LegacyRole.Applicant]}>
-                              <CreateApplicationPage />
-                            </RequireAuth>
-                          ),
+                          element: <CreateApplicationPage />,
                         },
                       ],
                     },
--- a/apps/web/src/pages/CreateApplicationPage/CreateApplicationPage.tsx
+++ b/apps/web/src/pages/CreateApplicationPage/CreateApplicationPage.tsx
@@ -30,6 +30,7 @@ export type RouteParams = {
 export const loader =
   (client: Client, userId: Scalars["UUID"]): LoaderFunction =>
   async ({ params, request }) => {
+    const url = new URL(request.url);
     const { poolId } = params;
 
     /**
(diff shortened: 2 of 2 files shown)


tristan-orourke:
This is a great example of how to use loaders! Overall, this definitely feels like an improvement for CreateApplicationPage.

> We can check if the user exists and boot them over to login if they do not. This means, we no longer need the RequireAuth wrapper for this.

The one caveat for me is the authorization. I liked having one method of authorizing our pages. At least, can the loader make use the existing Authorization contexts? We may be querying for `me` when that info is already loading. To be fair, this may not be a big deal, since urql should cache the `me` query. If we do have to reproduce some auth functionality in Loaders, seems like a good candidate for a reusable utility function.

> Render the Page

I also like the little React.Suspense example! I'm not entirely sure what advantages it offers over our existing loading strategy, but I know React has been working on it for a while, and we should probably think about using it where we can.

> No More API Wrappers

Personally, I still like the pattern of having one component which just receives everything it needs as props, and another component which produces the data and handler functions. It makes sense to me to replace ___API components which call the api, with ___Page components which get data from loaders. I'm less certain about ditching wrapper components entirely, though I'm willing to be outvoted.

> I'm wondering if this should be moved to an action and we remove some of the magic around application creation.
> For example, we don't even show the button to apply and instead tell the user to login to apply and bring them back to the advertisement after logging in. We could then move the logic currently in the loader to an action and convert the button to a Form with hidden inputs for poolId and userId. Since it is in an action, there is less magic and it should stop the mutation running twice.

I looked up react-router actions, and this sounds ideal to me! 🤩 The create-application route always made more sense as a POST action, it just didn't seem worth it to recreate a graphql mutation with a backend api endpoint like that. If react-router can simulate that for us, its perfect!

petertgiles:
Yeah, this is super cool.  I agree with everything Tristan said.  :laughing:  I also agree with you that a small design tweak like asking the user to log in before letting them apply would reduce the amount of magic.

esizer:
> The one caveat for me is the authorization. I liked having one method of authorizing our pages. At least, can the loader make use the existing Authorization contexts? We may be querying for `me` when that info is already loading. To be fair, this may not be a big deal, since urql should cache the `me` query. If we do have to reproduce some auth functionality in Loaders, seems like a good candidate for a reusable utility function.

Yeah, this is one of the things I didn't care for much in this approach. However, it may get better soon with the [middleware RFC](https://github.com/remix-run/react-router/discussions/9564).

> I also like the little React.Suspense example! I'm not entirely sure what advantages it offers over our existing loading strategy, but I know React has been working on it for a while, and we should probably think about using it where we can.

We are actually [using this right now](https://github.com/GCTC-NTGC/gc-digital-talent/blob/d6b66a44e70064592d1ddb69e089870012e5ca54/frontend/common/src/components/Pending/Pending.tsx#L42-L56)! However, this is not needed. This specific example makes use of [deferred data](https://reactrouter.com/en/main/guides/deferred), but we don't need to defer the data and have it load immediately 😄 The reason I used deferred here was because it was an intermediary route and a bit of a hack and I wanted the spinner to display while we were executing the mutation.

> Personally, I still like the pattern of having one component which just receives everything it needs as props, and another component which produces the data and handler functions. It makes sense to me to replace ___API components which call the api, with ___Page components which get data from loaders. I'm less certain about ditching wrapper components entirely, though I'm willing to be outvoted.

I more or less agree with this. I like the ___Page idea. Though, we will be able to do different things that deal with getting and mutating data in the wrapper component. I was thinking about stuff like using the [SEO component](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/frontend/common/src/components/SEO/SEO.tsx) and maybe the breadcrumbs and general layout then the children could be the actual content. Since we could make use of loaders and actions, those wrapper components don't need to handle the loading and mutating anymore.

> I looked up react-router actions, and this sounds ideal to me! 🤩 The create-application route always made more sense as a POST action, it just didn't seem worth it to recreate a graphql mutation with a backend api endpoint like that. If react-router can simulate that for us, its perfect!

Awesome! This was mostly my fault for doing this way when I was trying to meet the AC 😅 . I can update the description of https://github.com/gctc-ntgc/gc-digital-talent/issues/5352 to mention this approach as a possible fix :D

Issue #4716, opened 2022-11-10
Search page filters data needs to be preserved when going back from summary of filters page 


brindasasi:
### Problem
When clicking back from search request summary page , search page refreshed and lost all of the selected filters 
We need to cache the filters data on the search page  so that user can come back and make any changes to it 

![Screen Shot 2022-11-10 at 3.23.40 PM.png](https://images.zenhubusercontent.com/61a7bd49e20af80694dd845e/ebd402cd-575b-4bd6-80a3-2d527f50d2fb)

### Acceptance Criteria
- [ ] Navigating to search for the first time brings up default search filters
- [ ] Navigating back to search from a request page restores the previous search filters
- [ ] Navigating to search after successfully submitting a request gives default filters
- [ ] Should searching, not submitting, going somewhere else, should preserve filters if possible

### Recommended Implementation
- URL params

esizer:
## ❓ Question

Would we prefer to store this in `window.localStorage` or `URLSearchParams`? I like the idea of storing it in the url so it can be shared to pre-populate the search form wth some default values.

petertgiles:
Designer Convo:

Peter:
- Should navigating to search after successfully submitting a request give you the default filters or the just-submitted filters?
- Should searching, not submitting, going somewhere else, then returning to search restore the previous search filters?
Josh:
I think reset filters would be the expectation
Basically - if you've deliberately navigated away from the task, resetting filters shouldn't be a shock
But if you've navigated away by accident or want to edit a submission before sending it, then retaining them would be expected

I've updated AC

brindasasi:
[Open Requirement in TestQuality](https://web.testquality.com/site/brinda.sasikumar/project/15300?requirementId=15221)

brindasasi:
About this A/C 
>  Should searching, not submitting, going somewhere else, then returning to search gives default filters

currently if you navigate away from this page deliberately and come back using browser back button retains the filter data. Are we ok to keep that?  @Jerryescandon

PR #5617, opened 2023-02-06
pass the state while navigating

brindasasi:
🤖 Resolves {#4716}

## 👋 Introduction

Previously the search was assuming the filters on the page should always be default  while loading. This PR removes that assumption and add the ability to pass the state while navigating from another page or using browser back button. 

## 🧪 Testing

- [ ] Navigating to search for the first time brings up default search filters
- [ ]  Navigating  to search form a request page using back button or browser back button restores the previous search filters
- [ ] Navigating  to search form using browser back button from any page restores prev search filters
- [ ]  Navigating  to search form using `Search top navigation link button` from any page loads default search filters ( this is considered as a fresh navigation. This shouldn't reload the prev search filter ) 

## 📸 Screenshot


apps/web/src/pages/SearchRequests/SearchPage/components/SearchForm.tsx
```diff
@@ -38,6 +39,53 @@ import AdvancedFilters from "./AdvancedFilters";
 import AddSkillsToFilter from "./AddSkillsToFilter";
 import FilterBlock from "./FilterBlock";
 
+const positionDurationToEmploymentDuration = (
+  durations: Maybe<PositionDuration>[],
```
brindasasi, line 43: - [ ]  Move this out to a utility function

apps/web/src/pages/SearchRequests/SearchPage/components/SearchContainer.tsx
```diff
@@ -393,19 +397,31 @@ const SearchContainerApi = () => {
       error: searchFormDataError,
     },
   ] = useGetSearchFormDataAcrossAllPoolsQuery();
-
+  const applicantFilterFromBrowserHistory = state
+    ? state.applicantFilter
+    : undefined;
```
tristan-orourke, line 402:
```suggestion
  const applicantFilterFromBrowserHistory = state?.applicantFilter;
```

tristan-orourke (approved):
I'm very much looking forward to refactoring this properly 😆 . When we do that, I'd love to see the filters represent in url params like we do in the admin tables, but for now this meets the acceptance criteria and works well in testing. I left one minor suggestion, but this looks good to go.

PR #5618, opened 2023-02-06
Move tc-reports static files location block above apps/web static fil…

### LINKED ISSUE #5477: tc-report assets broken
## 🐛 Bug

The tc-report pages are broken right now.  The page loads but all assets are 404'ing.

## 📋 Steps to Reproduce

1. Visit http://localhost:8000/en/talent-cloud/report

## 🙋‍♀️ Proposed Solution

Probably an nginx routing issue.

## ✅ Acceptance Criteria

- [ ] tc-report pages can be viewed and all assets load correctly

yonikid15:
🤖 Resolves #5477 

## 👋 Introduction

Fixes the broken tc-report pages. Currently, the pages are loading but all assets are 404'ing.

## 🧪 Testing

1. Locally test all apps (admin, talentsearch, iap, tc-report) and make sure they load pages correctly, and ensure all static files are loading.
2. We should also test this on the dev server as this it makes changes to the nginx deploy config as well.

### DIFF
--- a/infrastructure/conf/nginx-conf-deploy/default
+++ b/infrastructure/conf/nginx-conf-deploy/default
@@ -92,6 +92,11 @@ server {
         alias /home/site/config-web.js;
     }
 
+    # rewrite tc-report static files to "tc-report/_site"
+    location ~ "^/tc-report/(.*\.(png|ico|gif|jpg|jpeg|svg|css|js|pdf|map|webmanifest))$" {
+        rewrite "^/tc-report/(.*\.(png|ico|gif|jpg|jpeg|svg|css|js|pdf|map|webmanifest))$" /tc-report/_site/$1 break;
+    }
+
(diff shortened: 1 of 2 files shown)


petertgiles (commented):
I've just tested locally and it's looking great. 

One thin I noticed that the favicons are still 404'ing for tc-report.  I can see that they're not under the `/tc-report` path.  I think that should be out of scope if you can put up a new issue to put those under the `/tc-report` path (or use the ones from the main app).

image

I'll test in Azure soon.

petertgiles (approved):
Looks good on Azure, too. :partying_face:   As long as there's an issue on the board to track the favicons I'm good with this as-is.

yonikid15:
Added new issue for the [favicon bug](https://github.com/GCTC-NTGC/gc-digital-talent/issues/5652)

Issue #5072, opened 2022-12-09
Move in Common


petertgiles:
> **Warning**
> As a team, we're not sure if we want to do this yet.

Now that admin, talentsearch, and iap are combined - do want to combine common in as well?
Would it make the project simpler?
Do we want to split out common to packages to use with turborepo?

PR #5628, opened 2023-02-06
Bug - 5551 - Updates `MissingSkills` component, stories

### LINKED ISSUE #5551: 🐛 `MissingSkills` stories all look the same
## 🐛 Bug

All of the story variations render the same thing regardless of which is selected:

- `MissingBothSkills`
- `MissingRequiredSkills`
- `MissingOptionalSkills`
- `MissingSomeSkills`

## 🦋 Expected Behaviour

Should render differently based on case. If not, then there is no apparent need for variation stories. 

## 📋 Steps to Reproduce

1. `npm run storybook`
2. Navigate to  **Missing Skills** stories

## 📸 Screenshot

Screen Shot 2023-01-31 at 09 31 23

## ✅ Acceptance Criteria

- [ ] Each of the four stories properly displays the component in the different states

mnigh:
🤖 Resolves #5551.

h/t @esizer for going through the logic behind the individual `MissingSkills` conditional blocks!

## 👋 Introduction

This PR updates the `MissingSkills` stories to reflect the current functionality and properties of `MissingSkills`.

## 🕵️ Details

The `MissingSkills` component itself had a bug in how `missingOptionalSkills` were calculated and was addressed in this PR in order for the different stories to be rendered correctly.

## 🧪 Testing

1. `npm run storybook`
2.  View five **Missing Skills** stories
3. Ensure each story render what its name proposes to do

## 📸 Screenshot
Screen Shot 2023-02-06 at 17 33 03

### DIFF
--- a/frontend/common/src/components/MissingSkills/MissingSkills.stories.tsx
+++ b/frontend/common/src/components/MissingSkills/MissingSkills.stories.tsx
@@ -1,12 +1,14 @@
 import React from "react";
 import { ComponentStory, ComponentMeta } from "@storybook/react";
+import { filterSkillsByCategory } from "@common/helpers/skillUtils";
+import { SkillCategory } from "@common/api/generated";
 import MissingSkills from "./MissingSkills";
 
-import { fakeSkills } from "../../fakeData";
+import { fakeSkillFamilies, fakeSkills } from "../../fakeData";
 
 type MissingSkillsComponent = typeof MissingSkills;
 
-const skills = fakeSkills();
+const skills = fakeSkills(10, fakeSkillFamilies(2));
 
 const fakeRequiredSkills = skills.splice(0, skills.length / 2);
 const fakeOptionalSkills = skills.splice(skills.length / 2, skills.length);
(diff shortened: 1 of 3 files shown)


frontend/common/src/components/MissingSkills/MissingSkills.stories.tsx
```diff
@@ -1,12 +1,14 @@
 import React from "react";
 import { ComponentStory, ComponentMeta } from "@storybook/react";
+import { filterSkillsByCategory } from "@common/helpers/skillUtils";
+import { SkillCategory } from "@common/api/generated";
```
petertgiles, line 4: I don't think we would use the `@common` alias from inside common.  I think this is what the linter doesn't like.

petertgiles (approved):
Great fix!  :clap:

frontend/common/src/components/MissingSkills/MissingSkills.stories.tsx
mnigh: Another reason to do #5072!

PR #5631, opened 2023-02-06
Feature/5534 pool table search

### LINKED ISSUE #5534: ✨ Add basic Search bar to Pools table
## ✨ Feature

We now have lots of Pools, and navigating the Pools table on Admin would be a lot easier with a search bar.

## 🕵️ Details

It's most important to be able to search by pool name (fully formatted pool name). We will also need to search by status and owner.

## 🎨 Design File

## 📸 Screenshot

This may be as simple as letting `search={true}` on the Table component, but we'll need to test that it works as expected. Some columns, like Pool Name, might benefit from being split into [accessor and cell](https://react-table-v7.tanstack.com/docs/api/useTable#column-options).

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Its possible to type Pool Name into a search bar on the Pools table to filter down the visible rows.
- [ ] Its possible to type a Pool Status into the search bar to filter to pools with that status (if it's not too difficult)
- [ ] Its possible to type an owner name into the search bar to filter to pools with that owner (if it's not too difficult)

vd1992:
🤖 Resolves #5534 

## 👋 Introduction

Adds searching to the pools table.

## 🕵️ Details

Rolled some other stuff in too. 
Refactored things to split into cells and accessors. Accessors only return primitives whereas cells are more complex and usually JSX.
Search and sort should all work, search uses the accessors. Sort default is case sensitive so doing a lot of toLowerCase() ☹️ 

## 🧪 Testing

1. head to /pools and play around
2. test the sort
3. test the search
4. test links

## 📸 Screenshot

gif

### DIFF
--- a/apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
+++ b/apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
@@ -292,22 +292,12 @@ export const PoolTable = ({ pools }: PoolTableProps) => {
         }),
         accessor: (d) => {
           const firstName =
-            d.owner && d.owner.firstName ? d.owner.firstName : "";
-          const lastName = d.owner && d.owner.lastName ? d.owner.lastName : "";
+            d.owner && d.owner.firstName ? d.owner.firstName.toLowerCase() : "";
+          const lastName =
+            d.owner && d.owner.lastName ? d.owner.lastName.toLowerCase() : "";
           return `${firstName} ${lastName}`;
         },
         Cell: ({ row }: IndividualCell) => fullNameCell(row.original, intl),
-        sortType: (rowA, rowB) => {
-          const a = rowA.original.owner?.firstName
-            ? rowA.original.owner.firstName.toLowerCase()
-            : "";
-          const b = rowB.original.owner?.firstName
-            ? rowB.original.owner.firstName.toLowerCase()
-            : "";
-          if (a > b) return 1;
-          if (a < b) return -1;
-          return 0;
-        },
         id: "ownerName",
       },
       {
@@ -317,7 +307,7 @@ export const PoolTable = ({ pools }: PoolTableProps) => {
           description: "Title displayed for the Pool table Owner Email column",
         }),
         accessor: (d) => {
-          return d.owner && d.owner.email ? d.owner.email : "";
+          return d.owner && d.owner.email ? d.owner.email.toLowerCase() : "";
         },
         Cell: ({ row }: IndividualCell) =>
           emailLinkAccessor(
@@ -326,17 +316,6 @@ export const PoolTable = ({ pools }: PoolTableProps) => {
               : "",
             intl,
           ),
-        sortType: (rowA, rowB) => {
-          const a = rowA.original.owner?.email
-            ? rowA.original.owner.email.toLowerCase()
-            : "";
-          const b = rowB.original.owner?.email
-            ? rowB.original.owner.email.toLowerCase()
-            : "";
-          if (a > b) return 1;
-          if (a < b) return -1;
-          return 0;
-        },
         id: "ownerEmail",
       },
       {
--- a/apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
+++ b/apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
@@ -163,43 +163,6 @@ export const PoolTable = ({ pools }: PoolTableProps) => {
         },
         Cell: ({ row }: IndividualCell) =>
           viewLinkAccessor(paths.poolView(row.original.id), row.original, intl),
-        sortType: (rowA, rowB, id, desc) => {
-          // passing in sortType to override default sort
-          let rowAName;
-          let rowBName;
-
-          if (locale === "en") {
-            rowAName = rowA.original.name?.en ?? "";
-            rowBName = rowB.original.name?.en ?? "";
-          } else {
-            rowAName = rowA.original.name?.fr ?? "";
-            rowBName = rowB.original.name?.fr ?? "";
-          }
-          const rowALevel =
-            rowA.original.classifications && rowA.original.classifications[0]
-              ? rowA.original.classifications[0].level
-              : 0;
-          const rowBLevel =
-            rowB.original.classifications && rowB.original.classifications[0]
-              ? rowB.original.classifications[0].level
-              : 0;
-
-          if (rowAName.toLowerCase() > rowBName.toLowerCase()) {
-            return 1;
-          }
-          if (rowAName.toLowerCase() < rowBName.toLowerCase()) {
-            return -1;
-          }
-          // if names identical then sort by level
-          // level sorting adjusted to always be ascending regardless of whether name sort is A-Z or Z-A
-          if (rowALevel > rowBLevel) {
-            return desc ? -1 : 1;
-          }
-          if (rowALevel < rowBLevel) {
-            return desc ? 1 : -1;
-          }
-          return 0;
-        },
       },
       {
         Header: intl.formatMessage({
--- a/apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
+++ b/apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
@@ -174,6 +174,7 @@ export const PoolTable = ({ pools }: PoolTableProps) => {
         accessor: (d) => {
           return d.id;
         },
+        disableGlobalFilter: true,
         Cell: ({ row }: IndividualCell) =>
           poolCandidatesLinkAccessor(
             paths.poolCandidateTable(row.original.id),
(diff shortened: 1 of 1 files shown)


apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
```diff
@@ -180,20 +215,41 @@ export const PoolTable = ({ pools }: PoolTableProps) => {
...
+                    color="primary"
+                    mode="outline"
+                  >
+                    {classification?.group}&#8209;{classification?.level}
+                  </Pill>
+                );
+              })
+            : null;
+          return classificationsCell(pillsArray);
+        },
```
esizer, line 250: Could this all be moved to the `classificationsCell` function?

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
```diff
@@ -153,11 +184,14 @@ export const PoolTable = ({ pools }: PoolTableProps) => {
           description:
             "Header for the View Candidates column of the Pools table",
         }),
-        accessor: (pool) =>
+        accessor: (d) => {
+          return d.id;
+        },
```
esizer, line 189: If we are retuning just the ID this might cause some confusion on sorting. I think it may be best to just disable it for this column with `disableSortBy`. Possible search as well, it is causing some issues where I get unexpected results.

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
```diff
@@ -67,6 +67,21 @@ function dateAccessor(value: Maybe<string>, intl: IntlShape) {
...
+    <span>
+      {getFullNameHtml(pool.owner?.firstName, pool.owner?.lastName, intl)}
+    </span>
+  );
+};
+
+const classificationsCell = (pills: JSX.Element[] | null) => {
+  if (pills) {
+    return <span>{pills}</span>;
+  }
```
esizer, line 81: `<Pill />` should be wrapped in `<Pills />` to show the proper styling when there is more than one.

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
```diff
@@ -67,6 +67,21 @@ function dateAccessor(value: Maybe<string>, intl: IntlShape) {
...
+  return <span />;
```
esizer, line 82: I think returning null would be a better option that returning an empty `<span />`

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
```diff
@@ -67,6 +67,21 @@ function dateAccessor(value: Maybe<string>, intl: IntlShape) {
   ) : null;
 }
 
+const fullNameCell = (pool: Pool, intl: IntlShape) => {
+  return (
+    <span>
+      {getFullNameHtml(pool.owner?.firstName, pool.owner?.lastName, intl)}
+    </span>
```
esizer, line 74: Does this need to be wrapped in a `span`? If not then I think we could remove it to reduce the tiny amount of DOM bloat.

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
```diff
@@ -276,11 +324,14 @@ export const PoolTable = ({ pools }: PoolTableProps) => {
           id: "tpzt/B",
           description: "Title displayed for the Pool table Edit column.",
         }),
-        accessor: (d) =>
+        accessor: () => {
+          return "Edit";
+        },
```
esizer, line 329: I think if we are just going to return the same value for all accessors on this column it would be best to disable sorting on this column. I would also disable search on this column as well since when I search for "Edi" all rows show up since they technically all satisfy that search string.

esizer (changes requested):
Looking good so far, couple of suggestions to go through.

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
vd1992: we don't appear to have such a component, it'll need to be made

image

and added to other places like here

image

esizer: Oh, you are right. Shoot! Umm, okay. I'd say out of scope. I can create an issue for that 👍

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
```diff
@@ -211,10 +267,10 @@ export const PoolTable = ({ pools }: PoolTableProps) => {
...
               : 0;
 
-          if (rowAGroup > rowBGroup) {
+          if (rowAGroup.toLowerCase() > rowBGroup.toLowerCase()) {
             return 1;
           }
-          if (rowAGroup < rowBGroup) {
+          if (rowAGroup.toLowerCase() < rowBGroup.toLowerCase()) {
             return -1;
           }
```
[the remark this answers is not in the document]
vd1992: I want to keep this one as it sorts classification in a particular way

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
vd1992: disableSortBy makes the header look not good 😬 


image

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
vd1992: it was to try and handle typing issues that kept popping up like

image

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
vd1992: done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/5631/commits/42fe014afa9588f87a12b3880dce81c6fd9c00f2

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
vd1992: done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/5631/commits/42fe014afa9588f87a12b3880dce81c6fd9c00f2

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
```diff
@@ -107,9 +128,19 @@ export const PoolTable = ({ pools }: PoolTableProps) => {
...
+            return `${d.name.en} ${d.stream ? d.stream : ""}`;
+          }
+          if (d.name && d.name.fr && locale === "fr") {
+            return `${d.name.fr} ${d.stream ? d.stream : ""}`;
+          }
+          return "";
+        },
+        Cell: ({ row }: IndividualCell) =>
+          viewLinkAccessor(paths.poolView(row.original.id), row.original, intl),
         sortType: (rowA, rowB, id, desc) => {
```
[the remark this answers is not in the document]
vd1992: done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/5631/commits/0152c7f4a9a70c360f3680432839dd79c9f26ba5

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
esizer: okay, I assumed there was something like that going on here. Would be nice to not need a span but def not in scope of this issue :P

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
esizer: Okay, not in scope of this issue so I'll create a ticket for that.

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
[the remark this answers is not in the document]
esizer: Ahh, okay that makes sense 👍

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
esizer: https://github.com/GCTC-NTGC/gc-digital-talent/issues/5642

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
vd1992: oh and this one https://github.com/GCTC-NTGC/gc-digital-talent/pull/5631/commits/e24b2faf554313e3d99be15bedb48b078532c32e

esizer (approved):

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
vd1992: thanks!

PR #5636, opened 2023-02-07
Feature - 5625 - Removes search request table pool column

### LINKED ISSUE #5625: ✨ Remove pool column from request table
## ✨ Feature

As an admin, I don't want the list of pool candidates to be easily reachable from the requests page so that I don't accidently end up confusing the pool candidates and the request search results.

## 🕵️ Details

This will go away anyway when we refactor requests to work without being associated to a specific pool. But in the meantime, the link to the pool is a little confusing on the admin UI. We could just remove that column right away. Hiding it by default would also be acceptable.

## 📸 Screenshot

## ✅ Acceptance Criteria

- [ ] Pool column no longer appears on the Request table. 
- [ ] Preference is to totally remove column (or hide if this is very difficult)

mnigh:
🤖 Resolves #5625.

## 👋 Introduction

This PR removes the **Pool** column from the search request table.

## 🧪 Testing

1. `npm run build`
2. Navigate to /admin/dashboard
3. Ensure **Pool** column is not rendered by default and is not in the list of Table columns
4. Navigate to /admin/talent-requests
5. Ensure **Pool** column is not rendered by default and is not in the list of Table columns

## 📸 Screenshots

Screen Shot 2023-02-07 at 11 50 03
Screen Shot 2023-02-07 at 11 49 51

### DIFF
--- a/apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
+++ b/apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo } from "react";
+import React, { useMemo } from "react";
 import { IntlShape, useIntl } from "react-intl";
 
 import { notEmpty } from "@common/helpers/util";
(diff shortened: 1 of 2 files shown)


vd1992 (approved):
/dashboard and /talent-requests appear to both work without issue
do not see pools as a header or column toggle
quite a chunk of code carved out by dropping that one column eh

looks good!

Issue #5642, opened 2023-02-07
🐛 `ClientManagedTable` sorting on some columns causes problems


esizer:
## 🐛 Bug

Right now we have sort enabled on all columns of the `ClientManagedTable`. This is a little odd when sorting by the "Edit" button and others. Also, when disabling sort, the visual display of the column header is a little odd.

## 🦋 Expected Behaviour

Some columns should not be able to be sorted since it does not make sense to sort them.

## 📋 Steps to Reproduce

1. Navigate to any `ClientManagedTable`
2. Sort by any column that includes a link/button
3. See that the sort doesn't really do much

## 📸 Screenshot

![217311252-52460791-1774-4829-916c-a8d4c2a8ca73](https://user-images.githubusercontent.com/4127998/217333408-0d082a44-e77c-4b07-9307-128456559e07.png)

## 🙋‍♀️ Proposed Solution

Review our tables and determine what columns should have sorting disabled. As well, we should have a visual indicator so users know what columns can be sorted.

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the bug was addressed.

- [ ] Sorting is disabled for the appropriate columns
- [ ] Column headers should be consistent in font between sortable and non-sortable columns
- [ ] Visual indicator for users to determine what columns can be sorted

tristan-orourke:
This table component is being replaced

PR #5649, opened 2023-02-08
Bug - 5600 - Update table accessors

### LINKED ISSUE #5600: 🐛 Update client tables to use `Cell` instead of `accessor` for rich text
## 🐛 Bug

Incorrect usage of the `accessor` property on `react-table` causing issues with sorting and filtering.

## 🦋 Expected Behaviour

The `accessor` attribute returns a primitive string for all tables to properly search, filter and sort.

## 🕵️ Details

Right now for most client side tables we use the `accessor` accessor property to format a columns values with rich text.  The docs specify this [should be a primitive, sortable value](https://react-table-v7.tanstack.com/docs/api/useTable#column-options:~:text=The%20data%20returned%20by%20an%20accessor%20should%20be%20primitive%20and%20sortable.). 

## 📋 Steps to Reproduce

1. Find a table that uses `accessor` to format with rich text
2. Attempt to search those values
3. Confirm that it does not work as expected

## 🙋‍♀️ Proposed Solution

Update all `accessors` to return a primitive value. If needed, use the `Cell` property to add rich text.

## ✅ Acceptance Criteria

- [ ] All `accessor` properties return a primitive value
- [ ] Rich text columns still function
- [ ] Searching works on all tables
- [ ] Sorting works on all tables
- [ ] Remove unneeded custom `sortType` from `react-table`'s on a case-by-case basis

esizer:
🤖 Resolves #5600 

## 👋 Introduction

This updates our `ClientManagedTable` components to use the `Cell` attribute instead of `accessor` to render non-primitive values.

## 🕵️ Details

We were using `accessor` to render any value (even JSX). However, the docs state that we should be using `Cell` instead and [make `accessor` a primitive value](https://react-table-v7.tanstack.com/docs/api/useTable#column-options:~:text=The%20data%20returned%20by%20an%20accessor%20should%20be%20primitive%20and%20sortable.).

## 🧪 Testing

2. Navigate to each table
3. Confirm they still function as expected
    - Search
    - Sort
    - Rendering values

### DIFF
--- a/apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
+++ b/apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
@@ -17,6 +17,7 @@ import useRoutes from "~/hooks/useRoutes";
 import Table, {
   ColumnsOf,
   tableViewItemButtonAccessor,
+  IndividualCell,
 } from "~/components/Table/ClientManagedTable";
 
 // callbacks extracted to separate function to stabilize memoized component
@@ -40,6 +41,8 @@ function dateAccessor(value: Maybe<string>, intl: IntlShape) {
   ) : null;
 }
 
+type Cell = IndividualCell<PoolCandidateSearchRequest>;
+
 interface SearchRequestTableProps {
   poolCandidateSearchRequests: Array<Maybe<PoolCandidateSearchRequest>>;
 }
@@ -60,16 +63,18 @@ export const SearchRequestTable = ({
             "Title displayed for the search request table edit column.",
         }),
         id: "action", // required when accessor is a function
-        accessor: ({ id, fullName }) =>
+        accessor: (d) => `Action ${d.id}`,
+        disableGlobalFilter: true,
+        Cell: ({ row: { original: searchRequest } }: Cell) =>
           tableViewItemButtonAccessor(
-            paths.searchRequestView(id),
+            paths.searchRequestView(searchRequest.id),
             intl.formatMessage({
               defaultMessage: "request",
               id: "gLtTaW",
               description:
                 "Text displayed after View text for Search Request table view action",
             }),
-            fullName || "",
+            searchRequest.fullName || "",
           ),
       },
       {
--- a/apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
+++ b/apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
@@ -11,6 +11,7 @@ import { formatDate, parseDateTimeUtc } from "@common/helpers/dateUtils";
 import {
   Maybe,
   PoolCandidateSearchRequest,
+  Scalars,
   useGetPoolCandidateSearchRequestsQuery,
 } from "~/api/generated";
 import useRoutes from "~/hooks/useRoutes";
@@ -29,14 +30,16 @@ const statusAccessor = (
     ? intl.formatMessage(getPoolCandidateSearchStatus(status as string))
     : "";
 
-function dateAccessor(value: Maybe<string>, intl: IntlShape) {
-  return value
-    ? formatDate({
-        date: parseDateTimeUtc(value),
+function dateCell(date: Maybe<Scalars["DateTime"]>, intl: IntlShape) {
+  return date ? (
+    <span>
+      {formatDate({
+        date: parseDateTimeUtc(date),
         formatString: "PPP p",
         intl,
-      })
-    : null;
+      })}
+    </span>
+  ) : null;
 }
 
 type SearchRequestCell = Cell<PoolCandidateSearchRequest>;
@@ -90,7 +93,10 @@ export const SearchRequestTable = ({
           description:
             "Title displayed on the search request table requested date column.",
         }),
-        accessor: ({ requestedDate }) => dateAccessor(requestedDate, intl),
+        accessor: ({ requestedDate }) =>
+          requestedDate ? parseDateTimeUtc(requestedDate).valueOf() : null,
+        Cell: ({ row: { original: searchRequest } }: SearchRequestCell) =>
+          dateCell(searchRequest.requestedDate, intl),
       },
       {
         Header: intl.formatMessage({
--- a/apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
+++ b/apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
@@ -17,7 +17,7 @@ import useRoutes from "~/hooks/useRoutes";
 import Table, {
   ColumnsOf,
   tableViewItemButtonAccessor,
-  IndividualCell,
+  Cell,
 } from "~/components/Table/ClientManagedTable";
 
 // callbacks extracted to separate function to stabilize memoized component
@@ -39,7 +39,7 @@ function dateAccessor(value: Maybe<string>, intl: IntlShape) {
     : null;
 }
 
-type Cell = IndividualCell<PoolCandidateSearchRequest>;
+type SearchRequestCell = Cell<PoolCandidateSearchRequest>;
 
 interface SearchRequestTableProps {
   poolCandidateSearchRequests: Array<Maybe<PoolCandidateSearchRequest>>;
@@ -63,7 +63,7 @@ export const SearchRequestTable = ({
         id: "action", // required when accessor is a function
         accessor: (d) => `Action ${d.id}`,
         disableGlobalFilter: true,
-        Cell: ({ row: { original: searchRequest } }: Cell) =>
+        Cell: ({ row: { original: searchRequest } }: SearchRequestCell) =>
           tableViewItemButtonAccessor(
             paths.searchRequestView(searchRequest.id),
             intl.formatMessage({
--- a/apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
+++ b/apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
@@ -30,15 +30,13 @@ const statusAccessor = (
     : "";
 
 function dateAccessor(value: Maybe<string>, intl: IntlShape) {
-  return value ? (
-    <span>
-      {formatDate({
+  return value
+    ? formatDate({
         date: parseDateTimeUtc(value),
         formatString: "PPP p",
         intl,
-      })}
-    </span>
-  ) : null;
+      })
+    : null;
 }
 
 type Cell = IndividualCell<PoolCandidateSearchRequest>;
@@ -92,8 +90,7 @@ export const SearchRequestTable = ({
           description:
             "Title displayed on the search request table requested date column.",
         }),
-        accessor: "requestedDate",
-        Cell: ({ value }) => dateAccessor(value, intl),
+        accessor: ({ requestedDate }) => dateAccessor(requestedDate, intl),
       },
       {
         Header: intl.formatMessage({
--- a/apps/web/src/components/Table/ClientManagedTable/types.ts
+++ b/apps/web/src/components/Table/ClientManagedTable/types.ts
@@ -1,0 +1,8 @@
+export type IndividualRow<T> = {
+  original: T;
+};
+
+export type IndividualCell<T> = {
+  row: IndividualRow<T>;
+  value: string;
+};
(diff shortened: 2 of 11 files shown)


vd1992:
I think the accessor for dates should also be localized so it is searchable by what you see rather than the stored date format

image

I think you'd just use the formatting function which returns a string

image

although, you'll need a custom sortType then to ensure sorting works (otherwise dates will be alphabetical)

vd1992:
problem in `SearchRequestCandidatesTable` I believe

image

apps/web/src/components/Table/ClientManagedTable/types.ts
```diff
@@ -0,0 +1,8 @@
+export type IndividualRow<T> = {
+  original: T;
+};
+
+export type IndividualCell<T> = {
```
petertgiles, line 5: This is a little strange but I'm guessing you had a good reason.  Why not just `Row` and `Cell`?

vd1992:
`SkillFamilyTable`

category's accessor is a jsx element, needs a cell

image

apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
```diff
@@ -29,17 +30,17 @@ const statusAccessor = (
...
-      })}
-    </span>
-  ) : null;
+      })
+    : null;
 }
 
+type Cell = IndividualCell<PoolCandidateSearchRequest>;
+
 interface SearchRequestTableProps {
```
petertgiles, line 44: Related to the other question, I'd expect this to be 
```tsx
type SearchRequestCell = Cell<PoolCandidateSearchRequest>;
```

esizer:
> category's accessor is a jsx element, needs a cell

Fixed in [update skill family category accessor to return a string](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5649/commits/5fa0b03a13b3dbfd6a0846b43883e9d051a9228e)

petertgiles (commented):
A quick question while working through this...

vd1992:
a little unsure of this, a string array is not a primitive right? arrays can be made of primitives but they aren't a primitive them self 

image

apps/web/src/components/Table/ClientManagedTable/types.ts
esizer: `react-table` uses `Cell` since it is for a component and I thought it could cause confusion. Also, just copied the exiting one defined in the `PoolTable`

https://github.com/GCTC-NTGC/gc-digital-talent/blob/518b6ee53199eed84530f000fada366eeec9ded1/apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx#L131-L135

However, I am **_totally_** open to changing it if we don't think that would be an issue 👍

apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
esizer: Actually, I should have read both before responding. I like this so I'll update all to match this pattern. Thanks! 😁

esizer: Thanks for suggestion, done in [simplify table type names](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5649/commits/3c4df795d3588458063b1490824a20c60ba39d6a)

esizer:
> a little unsure of this, a string array is not a primitive right? arrays can be made of primitives but they aren't a primitive them self

Oops, I meant to do a `.join(", ")` on those but forgot 😅 . Thanks for catching that! Fixed in [update array accessors to return sorted string](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5649/commits/97dc916ec9c21dc6949305af5cb623a7149043dd)

Also, I found another date access to update [update date accessor to return string](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5649/commits/3832161122d0cd5cb7d39ba7070ace987602c38c)

apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
```diff
@@ -87,8 +90,7 @@ export const SearchRequestTable = ({
           description:
             "Title displayed on the search request table requested date column.",
         }),
-        accessor: "requestedDate",
-        Cell: ({ value }) => dateAccessor(value, intl),
+        accessor: ({ requestedDate }) => dateAccessor(requestedDate, intl),
```
petertgiles, line 93: I'm not clear why this is happening or even if it pre-existed but the the url search params for sort state are really nasty.  Looks like's it's encoding a JSON object.
image

apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
```diff
@@ -87,8 +90,7 @@ export const SearchRequestTable = ({
...
       },
```
petertgiles, line 94: I think it's sorting the date alphabetically.
image

apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
```diff
@@ -60,16 +61,18 @@ export const SearchRequestTable = ({
             "Title displayed for the search request table edit column.",
         }),
         id: "action", // required when accessor is a function
-        accessor: ({ id, fullName }) =>
+        accessor: (d) => `Action ${d.id}`,
```
petertgiles, line 64: Is it possible to disable sort for a column?  It's meaningless to sort an action column.

vd1992:
well that's all my feedback addressed and didn't even need to hit Request Changes
handing things off to peter

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
```diff
@@ -65,15 +67,13 @@ function viewLinkAccessor(url: string, pool: Pool, intl: IntlShape) {
 }
 
 function dateAccessor(value: Maybe<string>, intl: IntlShape) {
-  return value ? (
-    <span>
-      {formatDate({
+  return value
+    ? formatDate({
         date: parseDateTimeUtc(value),
```
petertgiles, line 72: I don't think this should be formatting a string.  If it wants a primitive sortable value it should probably be `parseDateTimeUtc(value).valueOf()`.  This is probalby the source of the date sorting I noticed on SearchRequestTable.tsx.

petertgiles (changes requested):
Looking awesome!  A few last things to check...

apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
esizer: Yes! But that causes other issues. I created a ticket for that as well https://github.com/GCTC-NTGC/gc-digital-talent/issues/5642

apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
vd1992: wouldn't ascending (upwards arrow) be latest to to oldest, in which case April 2024 is newer than Feb 2023

apps/web/src/pages/SkillFamilies/components/SkillFamilyTable.tsx
```diff
@@ -76,12 +75,15 @@ export const SkillFamilyTable = ({ skillFamilies }: SkillFamilyTableProps) => {
           description:
             "Title displayed for the Skill Family table Edit column.",
         }),
-        accessor: (sf) =>
+        id: "edit",
```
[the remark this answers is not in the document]
petertgiles: Just figured out that nginx and asp.net are because it is sorting case-senstively.  I wonder if the accessor should run text through a normalize function to lowercase and remove diacritics for a primitive.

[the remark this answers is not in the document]
vd1992: asp.net and nginx is due to casing
upper case comes before lower case, hence why I liberally applied toLowerCase() in a bunch of places recently

apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
petertgiles: I think "arrow up" is smallest at the top.   For example, classification salaries:
image

apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
esizer: That is exactly what it is doing 😜 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/518b6ee53199eed84530f000fada366eeec9ded1/apps/web/src/components/Table/ClientManagedTable/Table.tsx#L208

petertgiles: Ah, maybe that's intended then.  :shrug: I guess I never noticed.

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
esizer: Thanks, this seemed to work for me!

[fix table date accessors](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5649/commits/c53c69ad78d8a7681c2f207e0b685405414868f1)

apps/web/src/pages/SearchRequests/ViewSearchRequestPage/components/SearchRequestCandidatesTable.tsx
```diff
@@ -71,6 +77,129 @@ function tableEditButtonAccessor(
...
+            })}
+        </Pill>
+      );
+    });
+
+  return equityItems ? <span>{equityItems}</span> : null;
+};
+
+const employmentEquityAccessor = (user: Applicant, intl: IntlShape) =>
+  getEmploymentEquityItems(user, intl)
```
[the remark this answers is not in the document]
esizer: oops, must have missed that one. good catch! 

[update equity accessor to return primitive](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5649/commits/a2256f1445579edc74495de689b0aec4623f83c5)

apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
esizer: Definitely ugly, I can't remember why but I did try to separate it out into each being a single value but it wasn't working. Maybe something we can revisit at some point 🤔

apps/web/src/pages/SkillFamilies/components/SkillFamilyTable.tsx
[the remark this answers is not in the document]
petertgiles: Fair enough.  I'll stub out an issue for that.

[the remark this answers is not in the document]
mnigh: I think **asp.net** and **nginx** are at the beginning (or end, if descending) because they begin with a lowercase letter. Same goes to why accented characters in French are treated differently.

One way to is to use localeCompare:

## Docs
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#using_options

## Example
https://github.com/GCTC-NTGC/TalentCloud/blob/17a4e0baa27dcd3f35f7d0958eb87202e256c277/resources/assets/js/components/JobBuilder/Skills/JobSkills.tsx#L275-L288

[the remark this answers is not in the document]
petertgiles: #5655

apps/web/src/components/SearchRequestTable/SearchRequestTable.tsx
petertgiles: :+1:

apps/web/src/pages/SearchRequests/ViewSearchRequestPage/components/SearchRequestCandidatesTable.tsx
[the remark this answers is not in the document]
petertgiles: Nice.  Probably don't need the map for an error message in the accessor but it works. :smile:

petertgiles (approved):

PR #5660, opened 2023-02-09
Feature/5216 view expired candidates

### LINKED ISSUE #5216: ✨ Admins need some way of viewing expired candidates in a pool
## ✨ Feature

As an Admin, I should be able to see all candidates in a pool who have expired.

## 🕵️ Details

The poolCandidatesPaginated query, which is used by the PoolCandidates table, locks the query to the ACTIVE expiryFilter scope, meaning any expired candidates will be filtered out. 

Remove the expiryFilter scope from the query, and instead add it to PoolCandidateSearchInput. Add an Expiry Status input to the filter dialog, with the options ACTIVE, EXPIRED, and ALL. Default this input to ACTIVE, so the default page behaviour is the same.

## ✅ Acceptance Criteria

- [ ] On the PoolCandidatesTable page, admins can change a filter to see all candidates (including expired ones), or only expired candidates
- [ ] By default, the PoolCandidates table continues to filter out expired candidates.
- [ ] Does not affect the behavior of the main search page
  - [ ] test verifies this

vd1992:
🤖 Resolves #5216 

## 👋 Introduction

Enables selectively filtering for candidates based off expiry, the expiry date and not the pool candidate status.

## 🕵️ Details

I decided to add a new scope `scopeExpiryStatus` because the old one was typed as array inputs, and trying to mess with it broke deprecated queries (`countPoolCandidates`). So rather than mess with other queries, I just added a new scope to be used with non-deprecated queries.
I am not sure the old one even worked right with the array since we usually just let the scope default to ACTIVE and never passed in actual arguments for it ever 🤔 

Old one

https://github.com/GCTC-NTGC/gc-digital-talent/blob/d59f2e2179f819c6d6a4c8ad61300bc61e838969/api/app/Models/PoolCandidate.php#L224-L236

## 🧪 Testing

1. play around with filtering, go to the profiles of user returned and ensure they are expired, or not, depending on the filter
2. observe the network requests
3. and whatever comes to mind

## 📸 Screenshot

image

image


vd1992:
### Translations

Pending

vd1992:
Chromatic run 
https://www.chromatic.com/build?appId=61e099a1bb1465003a73d3ce&number=4341

looks like the noise has been cleared @mnigh

brindasasi (approved):
LGTM!

PR #5683, opened 2023-02-09
Debt - 5609 - Create Application Parameter Hook

### LINKED ISSUE #5609: ♻️ Create hook for getting application from search params
## ♻️ Debt/Refactor

We should add a hook that gets the application id from the search params.

## 🕵️ Details

We have a pattern that is repeated where we look for the application ID in search params in the profile forms and return it with a return path. This ends up being the same 7 lines of code in every form. We could reduce this to 1 line and make it more maintainable by extracting this out into a hook.

## 🙋‍♀️ Proposed Solution

Naming is subjective but it would likely look something like this:

> **Note**
> No idea if this code works, but it could be something close 🤔 

```tsx
const useApplicationParam = (userId: Scalars["ID"]) => {
  const paths = useRoutes();
  const [searchParams] = useSearchParams();

  let id;
  let param;
  let returnRoute = paths.profile(userId);

  if(searchParams.has("applicationId")) {
    id = searchParams.get("applicationId");
    param = `?${searchParams.toString()}`;
    returnRoute = paths.reviewApplication(id)
  }

  return {
    id,
    param,
    returnRoute
  };
}


const {
    id: applicationId,
    param: applicationParam,
    returnRoute
} = useApplicationParam(userId);
```

## ✅ Acceptance Criteria

- [ ] A hook is created that gets relevant application information for profile forms
- [ ] All profile forms are updated to use the hook
- [ ] Breadcrumbs and back links still function properly in profile forms during an application

esizer:
🤖 Resolves #5609 

## 👋 Introduction

This extracts some re-used code out into a hook to keep the codebase DRY.

## 🕵️ Details

Returns:

- id: The current application id (if it exists)
- param: The existing search param (if it exists)
- returnRoute: The path any back buttons should return to
    - This is either the application,
    - The user profile
    - Generic profile route (if we do not have the current users id)

## 🧪 Testing

2. Navigate to each profile form and confirm they still function along with the back links
3. Create and application
4. Repeat step 3 within the context of an application

### DIFF
--- a/apps/web/src/hooks/useApplicationInfo.ts
+++ b/apps/web/src/hooks/useApplicationInfo.ts
@@ -19,21 +19,18 @@ const useApplicationInfo = (userId?: Scalars["ID"]) => {
   const [searchParams] = useSearchParams();
 
   let id;
-  let param;
   // If we do not have a userId, we can let the profile redirect handle it
   let returnRoute = userId ? paths.profile(userId) : paths.myProfile();
 
   if (searchParams.has("applicationId")) {
     id = searchParams.get("applicationId");
     if (id) {
-      param = `?${searchParams.toString()}`;
       returnRoute = paths.reviewApplication(id);
     }
   }
 
   return {
     id,
-    param,
     returnRoute,
   };
 };
--- a/apps/web/src/hooks/useApplicationInfo.ts
+++ b/apps/web/src/hooks/useApplicationInfo.ts
@@ -4,6 +4,16 @@ import { Scalars } from "~/api/generated";
 
 import useRoutes from "./useRoutes";
 
+/**
+ * Use Application Info
+ *
+ * Retrieves the current application context
+ * from search params returning its ID
+ * and a path to return to it.
+ *
+ * @param userId (optional) The ID of the applications user
+ * @returns {id: applicationID, returnRoute: Path to the current application}
+ */
 const useApplicationInfo = (userId?: Scalars["ID"]) => {
   const paths = useRoutes();
   const [searchParams] = useSearchParams();
--- a/apps/web/src/hooks/useApplicationInfo.ts
+++ b/apps/web/src/hooks/useApplicationInfo.ts
@@ -1,0 +1,31 @@
+import { useSearchParams } from "react-router-dom";
+
+import { Scalars } from "~/api/generated";
+
+import useRoutes from "./useRoutes";
+
+const useApplicationInfo = (userId?: Scalars["ID"]) => {
+  const paths = useRoutes();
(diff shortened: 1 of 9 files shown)


apps/web/src/hooks/useApplicationInfo.ts
```diff
@@ -0,0 +1,31 @@
+import { useSearchParams } from "react-router-dom";
+
+import { Scalars } from "~/api/generated";
+
+import useRoutes from "./useRoutes";
+
+const useApplicationInfo = (userId?: Scalars["ID"]) => {
```
petertgiles, line 7: You know me... could you add some documentation to this hook?  :laughing:

apps/web/src/hooks/useApplicationInfo.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  let id;
+  let param;
+  // If we do not have a userId, we can let the profile redirect handle it
+  let returnRoute = userId ? paths.profile(userId) : paths.myProfile();
+
+  if (searchParams.has("applicationId")) {
+    id = searchParams.get("applicationId");
+    if (id) {
+      param = `?${searchParams.toString()}`;
```
petertgiles, line 19: This isn't exactly the same behaviour as before.  Did you mean to return all the search params here?  It seems like it would match the name of the hook a little better if it just returned the applicationId or a key-value pair.  Also, if you didn't hardcode the question mark in the string you wouldn't run the risk of URL building a url with more than one question mark in the code that's using this hook.

petertgiles (changes requested):
Looking cool!  I just added a few thoughts.

patcon (approved):
I wanted to write a cypress test for this, but I think that was just tripping me up. LGTM!

apps/web/src/hooks/useApplicationInfo.ts
esizer: Thanks for reminding me 😄 Added in [add documentation](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5683/commits/7ad4ced1697a72c0cb71fabfc49dc965735ef029)

apps/web/src/hooks/useApplicationInfo.ts
esizer: Honestly, I felt a little weird about returning it at all because it can easily be done just by using the ID if needed. I removed it in [update to only return application related info](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5683/commits/cff5c72f60b64e56a0d629abe79acd2986ffa9e5)

petertgiles (approved):
Nice.  :+1:

Issue #4389, opened 2022-10-19
Replace Add Experience buttons with One Button + Modal


tristan-orourke:
One "Add Experience" button opens a modal, where selects what kind of Experience.

### Copy

```
Add a new experience to your resume
Get started by selecting the type of experience you would like to add.

Before you start describing your experience, first we would like to know what category it falls into. There are five options to choose from, you can add as many experiences to each category as you want.
```

```
Work experience
Share your experiences gained from full-time positions, part-time positions, self-employment, fellowships or internships. 
Add work experience
```

```
Education experience
Got credentials? Share your degree, certificates, online courses, trade apprenticeship, licences or alternative credentials. If you’ve learned something from a recognized educational provider, include your experiences here. 
Add education experience
```

```
Volunteer and community experience
Gained experience by being part of or giving back to a community? People learn skills from a wide range of experiences like volunteering or being part of non-profit organizations, Indigenous communities, or virtual collaborations. This could be anything from helping during events and ceremonies, DJ-ing at a friends wedding, to gaming and streaming.
Add community experience
```

```
Personal experience
People are more than just education and work experiences. We want to make space for you to share your learning from other experiences.
Add personal experience
```

```
Award
Did you get recognized for going above and beyond? There are many ways to get recognized, awards are just one of them. 
Add education experience
```

## EXP PAGE

![image](https://user-images.githubusercontent.com/16977254/216445164-d9a31522-9ef9-420c-8b4b-fa97367efda2.png)

## MODAL

![image](https://user-images.githubusercontent.com/16977254/217303458-1ed93f1b-fdbc-4c4c-a31b-4d995baf0467.png)

Acceptance Criteria:
- [ ] Replace series of five "add" buttons with single "add experience" button
- [ ] Button opens a modal with design shown up, also mobile view
- [ ] Clicking one of the "add" buttons in the dialog should navigate to the create page
- [ ] Cancel button closes dialog without navigation
- [ ] Use default behaviour of dialog initial focus on "X" in corner

petertgiles:
@Jerryescandon 
- Any differences for mobile sizes?
- "add" buttons in dialog should functionally be links?
- Could you provide the copy of the dialog as text?

Jerryescandon:
oh, I am unsure about this still being that relevant.

I would freeze this issue for now honestly, or ask @substrae if he has a different direction he wants to take Experience creation on the profile side.

substrae:
[...]

PR #5696, opened 2023-02-10
Allow IconButton to accept a ref

### LINKED ISSUE #4656: Find a way to not need to omit ref
https://github.com/GCTC-NTGC/gc-digital-talent/blob/9968de2525e4614e54d5a66147279df65539f738/frontend/common/src/components/Button/IconButton.tsx#L5

petertgiles:
🤖 Resolves #4656 

## 👋 Introduction

This adds the ability to forward a ref to underlying Button component for IconButton.

## 🕵️ Details

This turned out to be a blocker for #4389 so I pulled it in here.  I guess the Dialog component relies on passing a ref to the trigger button so it was emitting warnings when I tried to use an IconButton as the trigger.

Read more here: https://beta.reactjs.org/learn/manipulating-the-dom-with-refs#accessing-another-components-dom-nodes

## 🧪 Testing

1. Review the new test
2. Observe that the test passes
3. Feel good that I pulled this commit into my PR for #4389 and confirmed that it fixed the problem :smile:

### DIFF
--- a/frontend/common/src/components/Button/IconButton.tsx
+++ b/frontend/common/src/components/Button/IconButton.tsx
@@ -1,16 +1,17 @@
 import React from "react";
 import Button from "./Button";
-import type { ButtonProps } from "./Button";
 
-// NOTE: We should find a way to not need to omit ref
-export interface IconButtonProps extends Omit<ButtonProps, "ref"> {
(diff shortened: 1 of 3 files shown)


petertgiles:
Big thanks to Eric for helping me get started on this one!

esizer (approved):
Awesome! ⭐

Issue #3642, opened 2022-08-09
Make API user set per-unit-test


vd1992:
## Description
Policies aren't really being tested and probably ought to be
Implementing them is complicated by https://github.com/GCTC-NTGC/gc-digital-talent/blob/9bc07b61b5ec5461ccf6f7616f798ea604ada264/api/phpunit.xml#L30

Remove that and set env on a per test file basis where not required, enabling policy testing

## Acceptance Criteria
- [ ] remove the env lin in the .xml file
- [ ] resolve broken tests by adding "setenv" stuff per file

   [image]: https://user-images.githubusercontent.com/305339/166570935-6425a615-82c7-4b2f-9840-5c7d698fa2ab.png
   [image-link]: https://example.com
   [figma]: https://go.talent.c4nada.ca/figma
   [sitemap]: https://go.talent.c4nada.ca/sitemap

tristan-orourke:
The best way to run requests as a particular user in phpunit tests is to use the [`actingAs` helper](https://laravel.com/docs/5.2/testing#sessions-and-authentication).

For example:
```php
$adminResponse = $this->actingAs($admin, 'api')->graphQL($query);
$applicantResponse = $this->actingAs($applicant, 'api')->graphQL($query);
```
Note that we must specify that we're actingAs a user according to the 'api' guard. Our graphql requests use the 'api' guard, as specified in `api/config/lighthouse.php`, while our general Laravel app still uses the 'web' guard by default, as specified in `api/config/auth.php`. (Maybe we should just change it there, since we're only using Laravel for our API.)

PR #5698, opened 2023-02-10
Add `@canOnParent` custom directive

### LINKED ISSUE #5587: ✨Extend Lighthouse `@can` directive to resolve parent object
## ✨ Feature

Extend the Lighthouse `@can` directive so it can work based off the parent of a resolved field.

## 🕵️ Details

The [can](https://lighthouse-php.com/master/api-reference/directives.html#can) directive already allows us to guard a query, mutation, or with with Policy method. Using the `@can(resolved: true)` option allows for passing the resolved object to the Policy. For example, 
```
type PoolCandidate {
  user: User @can(ability: 'view', resolved: true)
}
...
UserPolicy {
  view(User $user, User $model) { ... }
}
```
In the above, trying to view PoolCandiate.user will pass the user in question to UserPolicy.view in the $model param.

However, this doesn't work for array fields. An easier way to do this would be a `@can` directive which resolves the Parent object of the resolved field. For example:
```
type User {
  applications: [PoolCandidate] @can(ability: 'viewApplications', resolveParent: true)
}
...
UserPolicy {
  viewApplications(User $user, User $model) { ... }
}
```
In this case, the parent of the resolved field (`User`) is used to resolve the policy, instead of the field itself (`[PoolCandidate]`).

## 🎨 Design File

## 📸 Screenshot

Creating a new directive instead of extending is also perfectly acceptable. 

Pros of extended - consistent use of only a single directive for all permission guards
Pros of new directive - documentation for the Lighthouse directive remains accurate

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] We have a directive which can guard a field, but resolve the policy based on the field's parent
- [ ] phpunit tests for this directive

tristan-orourke:
🤖 Resolves #5587 

## 👋 Introduction

Complete a new directive which allows for gating the properties of an object on the Policy of the object itself (instead of the Policy of the property).

## 🕵️ Details

Actually, the policy mostly existed already, but I changed the name, updated it slightly, and added a phpunit test.

This also can be an example of how to answer #3642.

## 🧪 Testing

This doesn't modify the app yet, only add a new tool.
Running phpunit tests should test the new directive.


vd1992 (changes requested):
for the test, will need to find a way to make the results returned in a more consistent way
right now the array of ids is returned randomly and so sometimes the check fails if the order doesn't line up

image

vd1992:
`actingAs` though, looks very neat indeed 🤔 

gave it a whirl just for fun and seems to work just fine, like this example here, it runs

image

petertgiles:
> will need to find a way to make the results returned in a more consistent way

You could try [assertSimilarJson](https://laravel.com/api/9.x/Illuminate/Testing/TestResponse.html#method_assertSimilarJson) instead, which allows arrays to be out of order.

api/tests/Feature/DirectivesTest.php
```diff
@@ -128,4 +135,113 @@ public function testPluckFromObject(): void
...
+                ['message' => 'This action is unauthorized.']
+            ]
+        ];
+
+        $isUser = function($user) {
+            return function($argument) use ($user) {
+                return $argument instanceof User && $argument->id == $user->id;
+            };
+        };
+        $mock = $this->mock(UserPolicy::class);
```
petertgiles, line 224: Why is it necessary to mock the user policy?  Couldn't you use the real policy in the test schema?

petertgiles (approved):
A lot of this is above my head but it is very cool!  Seems to work well.

api/tests/Feature/DirectivesTest.php
tristan-orourke: I originally let it use the real policy, but mocking it
1. makes this test independent of any changes to the UserPolicy, which is not what I'm trying to test here
2. and more importantly, allows me to confirm that UserPolicy was checked, not PoolCandidatePolicy, which is the key point of the directive.

tristan-orourke:
> right now the array of ids is returned randomly and so sometimes the check fails if the order doesn't line up

Ugh, this works 10/10 times locally, why does it fail on CI 😫

api/tests/Feature/DirectivesTest.php
petertgiles: Makes sense, thanks. :pray:

vd1992 (approved):
tests ✅

tristan-orourke:
@vd1992 @petertgiles I fixed the tests by avoiding checking the specific contents of the collection. Also may be worth highlighting that I used Laravel's [fluent JSON testing](https://laravel.com/docs/9.x/http-tests#fluent-json-testing) methods, which I think are new?

PR #5705, opened 2023-02-14
Feature - 5622 - add page break to candidate documents

### LINKED ISSUE #5622: ✨ Page break after each application in the candidate print/download view
## ✨ Feature

As the recruitment team I want a "Page Break" after each individual application in the PDF candidate download so I can easily separate applications.

## 🕵️ Details

It doesn't need to be super slick, but it would be helpful to have a page break after each application. It's probably fine if it results in occasional "double page breaks" or similar.

## ✅ Acceptance Criteria

- [ ] Page break after every application in print view
- [ ] OK if a bit glitchy - works best in Edge and Chrome

yonikid15:
🤖 Resolves #5622 

## 👋 Introduction

Adds a page break after each individual application in the PDF candidate and user profile downloads, so that they can be easily separated.

- [x] Page break after every application in print view
- [x] OK if a bit glitchy - works best in Edge and Chrome. (Also works well in Firefox)

## 🕵️ Details

When testing please check all major browsers (Chrome, Firefox, Edge, etc) to see if the page break works correctly. I was seeing Firefox was adding an extra page, even though it wouldn't on Chrome.

## 🧪 Testing

1. Go to Pool Candidates table and Users table
2. Select 1 or more users, ideally try different amounts for testing
3. Ensure there are page breaks after each candidate and there is not an extra blank page.


petertgiles (approved):
Great!  Works well in Firefox and Chromium on Linux.

apps/web/src/components/PoolCandidatesTable/PoolCandidateDocument.tsx
```diff
@@ -126,143 +126,157 @@ const PoolCandidateDocument = React.forwardRef<
       <div ref={ref}>
         <div className="print-container">
           {candidates &&
-            candidates.map((candidate) => {
+            candidates.map((candidate, index) => {
               return (
-                <div key={candidate.id} className="page-wrapper">
-                  <div className="page-section">
+                <>
+                  <div key={candidate.id} className="page-wrapper">
```
petertgiles, line 132: Oops, don't forget to move the key up the top element in the map.

petertgiles (changes requested):
small fix needed on both documents

apps/web/src/components/PoolCandidatesTable/PoolCandidateDocument.tsx
yonikid15: Nice catch! Should be fixed by [a28a984](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5705/commits/a28a9848a3844db615243ad9cdc13734873816f6)

petertgiles (approved):
Looks like there's a key problem with the breadcrumbs but this PR looks good.

PR #5729, opened 2023-02-16
Feature - 5595 - Adds sorting for select components

### LINKED ISSUE #5595: Make select and multiselects sort alphabetically by default
## ✨ Feature

We're hoping first-letter-navigation will "just work" if we have our selects and multiselects all sorted alphabetically.  This should help other users find their options more easily, too.

## 🕵️ Details

More context can be found in #5487.

## ✅ Acceptance Criteria

- [ ] Select and Multiselects sort their options alphabetically by default
- [ ] Auto-sort can be disabled with props if necessary
- [ ] If the user is using French then the options should be sorted by their French labels
- [ ] If there are option groups, sort groups first and items within the groups second

mnigh:
🤖 Resolves #5595.

## 👋 Introduction

This PR allows for alphanumeric sorting of `option` and `optgroup` in select components (`Select`, `SelectFieldV2`, `MultiSelectField`). Also, adds the ability to render `optgroup` in the `Select` component.

## 🪙  Gratuities
@tristan-orourke, @petertgiles, @esizer.

## 🧪 Testing

1. `npm run storybook`
2. View `Select`, `SelectFieldV2` stories
3. Ensure they all render as expected
4. Navigate to pages using select components (`Select`, `SelectFieldV2`, `MultiSelectField`)
5. Ensure they all render and submit as expected


frontend/common/src/components/form/MultiSelect/MultiSelectField.test.tsx
```diff
@@ -104,6 +104,7 @@ describe("MultiSelectFieldV2", () => {
           { value: "BAZ", label: "Baz" },
           { value: "BAM", label: "Bam" },
         ]}
+        doNotSort
```
esizer, line 107: I tried to get this to work but I have no clue how to test this component because the markup it generates is just non-sense lol. It is rendering `<div>`'s with no discernable role to select them with. Probably a more pressing concern for a11y as well 🤔 

I check [WAI Comobox Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/) and they clearly have a `<li>` with `role=option` which is what `<Combobox />` is doing.

frontend/common/src/components/form/Select/Select.tsx
```diff
@@ -31,6 +49,40 @@ export interface SelectProps
   nullSelection?: string;
   /** Determine if it should track unsaved changes and render it */
   trackUnsaved?: boolean;
+  /** Determine if it should sort options in alphanumeric ascending order */
+  doNotSort?: boolean;
+}
+
+function sortOptions(options: OptGroupOrOption[]) {
+  const tempOptions = options.map((option: OptGroupOrOption) =>
+    Object.prototype.hasOwnProperty.call(option, "options")
```
esizer, line 58: Was `"options" in options` not working here?  It does the same thing, but is a little more terse.

REF: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in

frontend/common/src/components/form/Select/Select.tsx
```diff
@@ -93,15 +150,37 @@ const Select: React.FunctionComponent<SelectProps> = ({
...
-            <option
-              aria-label={option.ariaLabel}
-              key={option.value}
-              value={option.value}
-            >
-              {option.label}
-            </option>
-          ))}
+          {optionsModified.map((option) =>
+            Object.prototype.hasOwnProperty.call(option, "options") ? (
```
esizer, line 154: See previous comment about the `in` operarator

frontend/common/src/components/form/Select/Select.tsx
```diff
@@ -93,15 +150,37 @@ const Select: React.FunctionComponent<SelectProps> = ({
...
+              <optgroup
+                key={`optgroup${option.label}`}
+                label={
+                  isNull(option.label) ||
+                  isNumber(option.label) ||
+                  isBoolean(option.label) ||
+                  isUndefined(option.label)
+                    ? ""
+                    : option.label.toString()
+                }
```
esizer, line 164: Would ` label={option.label?.toString() || ""}` get to the same end result? I think the `?` will omit `null`, `undefined` and `false`. Though it would result in "true" being printed 🤔

esizer (commented):
Looking good! Everything except that pesky `react-select` test was working for me 😄 

I just had a few questions before approving 👍 \

frontend/common/src/components/form/Select/Select.tsx
tristan-orourke: If this works, try `label={option.label?.toString() ?? ""}` instead. We almost always want to use [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing#assigning_a_default_value_to_a_variable) instead of logical or for setting default values. That should avoid accidentally turning the expression into a boolean 😉 . Also it allows for falsey values like `0`.

frontend/common/src/components/form/MultiSelect/MultiSelectField.test.tsx
mnigh: 🤷

frontend/common/src/components/form/Select/Select.tsx
mnigh: Unfortunately, using the  `in` operator here results in this:

## Screenshot
Screen Shot 2023-02-17 at 10 53 11

instead of:

## Screenshot
Screen Shot 2023-02-17 at 10 54 06


> Unlike the [in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in) operator, this method does not check for the specified property in the object's prototype chain.

REF: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty#description

frontend/common/src/components/form/Select/Select.tsx
esizer: Makes sense to me! Though, converting to a string would make `0` and `false` truthy 😜 

Screenshot 2023-02-17 105842

frontend/common/src/components/form/Select/Select.tsx
esizer: Dang, okay!

frontend/common/src/components/form/Select/Select.tsx
mnigh: This was another one that TypeScript was complaining about it and I ended up with the mess, sorry. "Resolved" in [Refactor label condition](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5729/commits/3748982f345014d205e69c490d494c8ab60f28a2).

frontend/common/src/components/form/Select/Select.tsx
mnigh: See https://github.com/GCTC-NTGC/gc-digital-talent/pull/5729#discussion_r1110012421

esizer (approved):
LGTM! 👍

PR #5740, opened 2023-02-17
Allow Heading size prop to be unset

### LINKED ISSUE #5790: 🐛 `<Heading />` Component is not explicit in its styling props
## 🐛 Bug

The `<Heading />` component could be more explicit when selecting a size to display at.

## 🦋 Expected Behaviour

`<Heading />` component uses the size of the `level` defined unless it is overridden by `size`.

## 🕵️ Details

Currently, our `<Heading />` component defaults to `size = "h2"` which can cause some confusion when using different levels. We could have no default size and always fallback to the `level` prop and force developers to explicitly override it with the `size` prop.

## ✅ Acceptance Criteria

- [ ] Remove `size` default
- [ ] Ensure no visual diff

patcon:
🤖 Resolves #5790.

## 👋 Introduction

`level` prop sets the element name.
`size` prop sets the style override of the level's style.

This change simply allows Heading to be used like:

```
<Heading level="h3">Foo</Heading>
```

instead of requiring this:

```
<Heading level="h3" size="h3">Foo</Heading>
```

`size` prop is then only used when the style specified via "size" should override the element's default style.

## 🕵️ Details

This should not affect any existing Heading component usage, but will allow future using of Heading to be more terse.

## 🧪 Testing

1. Confirm that the Chromatic visual diffs haven't changed in appearance.

## 📸 Screenshot

cc: @substrae


esizer (changes requested):
Seeing some problems in the chromatic diff where headers are changing in size and weight fairly dramatically:

https://www.chromatic.com/pullrequest?appId=61e099a1bb1465003a73d3ce&number=5740&view=changes

patcon:
It seems there are lots of places where previously, the `level` was set (e.g., `h1`), but the size is unspecified and so previously defaulted [silently] to `h2`.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/5ea9db248f9ad30899bdc622fbcadb7bd412e037/frontend/common/src/components/Hero/Hero.tsx#L78-L80

But with this PR change, in order for those existing components not to change visibly, it means what was previous `<Heading level="h1" />` (again, styled silently as `h2`) require explicitly setting size as `<Heading level="h1" size="h2" />`. This change will preserve the prior styling.

Example of diff right now: https://www.chromatic.com/pullrequest?appId=61e099a1bb1465003a73d3ce&number=5740&activeElementId=comparison-test-key-63c1b6caded639d48e86d61a-1200&view=changes

Screen Shot 2023-02-17 at 5 04 07 PM


Does this new behaviour seem expected @substrae -- the explicitness that this creates, that it's clear the `<Heading />` props when the element `level` doesn't match the styling `size`, that personally feels like a good change to create more clarity. But maybe others feel differently?

substrae:
@patcon @esizer I'll always lean towards explicitness in cases like this. I understand it's a bit more work to go in and find instances where we need to address the diff manually, but I absolutely prefer the size following the level unless overwritten.

substrae:
Gotcha. I guess it comes down to the scope of refactoring the change and whether it should be shelved until we have the appropriate time for it.

esizer:
> Gotcha. I guess it comes down to the scope of refactoring the change and whether it should be shelved until we have the appropriate time for it.

Yeah, we will need to go through each diff and confirm if the change is wanted. Some may intentionally be `h2` while others are appearing at `h2` accidently because of it implicitly being set at `h2` by default.

brindasasi:
We need an issue for it and needs to be planned for the sprint.

patcon:
Ok, the `<Heading />` component usage has been updated so that any `size` props are explicitly specified as `h2` now, instead of silently being set. So this means that the chromatic build shows no diffs, even when this "fixes" the unexpected setting of `h2` in the underlying component.

We can now spin out a full issue to review all the `<Heading />` components to see if the styling we had before is what we want. This original visual diff might be helpful to find candidates for fixing the `<Heading />`'s  `size` prop:
https://www.chromatic.com/test?appId=61e099a1bb1465003a73d3ce&id=63efaa4993ad88891aea18bd

patcon:
Spun out the work suggested by @esizer into https://github.com/GCTC-NTGC/gc-digital-talent/issues/5753

Ready for review.

esizer (approved):
LGTM! 👍 

I went ahead and created an issue to attach this too so you get some story points 😜 

Sorry, I was focused on getting the refactor merged yesterday and now you have a few merge conflicts to deal with 😢

patcon:
thanks so much @esizer! will resolve these conflicts

mnigh:
@patcon does this PR resolve [this issue](https://github.com/GCTC-NTGC/gc-digital-talent/issues/5753)? If so, would you mind updating the PR description from:
 
🤖 Resolves (no issue), just several character change.
to
🤖 Resolves #5753.

esizer:
> @patcon does this PR resolve [this issue](https://github.com/GCTC-NTGC/gc-digital-talent/issues/5753)? If so, would you mind updating the PR description from:
> 
> 🤖 Resolves (no issue), just several character change. to 🤖 Resolves #5753.

I connected this to #5790 back when we were using zenhub but it looks like it needs to be connected via github.

PR #5749, opened 2023-02-21
5563 assign query roles

### LINKED ISSUE #5563: ✨ Assignment and Querying of Users to Roles and Teams
## ✨ Feature

Admins need to be able to:
- See which Roles each user has
- If a user has team-based Roles, see which Teams those Roles are associated with
- Know whether a Role is intended to be team-based or not
- See which users are on a team, and which teams a user is on (determined by whether the user has any team-based Roles)
- Assign or remove Roles to Users (team-based and not)

## 🛑 Blocked By
- [x] #5511 
- [x] #5559 

## 🕵️ Details

## 🎨 Design File

## 📸 Screenshot

- Add roles and teams to the Users type. Use the `@can` directive to ensure related objects can only be viewed with the correct permissions.
```
type User {
  ...
  roles: [Role] @belongsToMany @can('view', resolved: true)
  teams: [Team] @belongsToMany(relation: "rolesTeams") @can('view', resolved: true)
}
```
- When retrieving a user's roles, it should be possible to see if the role is attached to a Team. I think this can be implemented by using Laravel `withPivot`, similar to what we did with Users, Experiences and Skills:
```
type Role {
  ...
  team: Team @can('view', resolved: true) # This will be null unless going through a user-role pivot table
}
```
- Add users (basic info only) to the Teams type. Use the `@can` directive to ensure they can only be viewed with the correct permissions.
```
type Team {
  ...
  members: UserPublicProfile @belongsToMany(relation: "users") @can('view', resolved: true, model:  "User")
}
```
- QUESTION: We have options in how we design the mutations to attach/detach roles. 
  - We could do it all in one _sync_ mutation, something like:
  ```
  input RoleInput {
    roleId: ID!
    teamId: ID # must be set for team-based Roles, and left null for other Roles
  }
  type Mutation {
    syncRoles($userId: ID!, $rolesInput: [RoleInput])
  }
  ```
  - We could split attach and detach and do them one at a time. This may force the client to run a number of mutations at once:
  ```
  attachRole($userId: ID!, $roleInput: RoleInput!)
  detachRole($userId: ID!, $roleInput: RoleInput!) # if a role was attached with a team, it must also be detached with it.
  ```
  - and, we could split up Roles and TeamRoles, whether using sync or attach/detach:
  ```
  attachRole($userId: ID!, $roleId: ID!)
  attachTeamRole($userId: ID!, $roleInput: TeamRoleInput!)
  ```
- We may want to add a `users` relationship visible on teams. This could also show a teamRoles pivot, if we really wanted, though I doubt that's necessary.
- QUESTION: The permissions on all this are a bit thorny, especially if we want to to enable something like `team.users`. Perhaps having a separate query, `teamMembers($teamId!): UserPublicProfile` would be easier. This should be considered carefully in relation to #4362 and #2266.

## 🌎 Localization

## ✅ Acceptance Criteria

Admins can...
- [ ] See which Roles each user has
- [ ] If a user has team-based Roles, see which Teams those Roles are associated with
- [ ] Know whether a Role is intended to be team-based or not
- [ ] See which users are on a team, and which teams a user is on (determined by whether the user has any team-based Roles)
- [ ] Assign or remove Roles to Users (team-based and not)

Non-admins can...
- [ ] do none of the above

And...
- [ ] phpunit feature tests to verify the above cases.

petertgiles:
🤖 Resolves #5563 

## 👋 Introduction

This branch adds the ability to query and assign teams and roles.

## 🕵️ Details

- To work with the three way pivot (user, role, team) with polymorphism, a new model was created called RoleAssignment.  This model represents the role_user pivot table.
- The user local seeders were updated to assign roles and teams to seeded users.
- I know @tristan-orourke and I disagreed about where the mutation should be (standalone vs. part of user).  I really like how it turned out as part of user.  :smiling_face_with_tear: If you still think it should be standalone let me know and I'll move it out.
- I'm a little fuzzy on exactly what needs to be protected and how.   I think it's good but let me know if I missed anything:
  - The mutation is only part of "updated as admin".
  - All references to user just use UserPublicProfile.
  - The top level teams and roles queries already have "admin only" policies.
  - The user fields already can only be accessed by admin or self.
  - I think we decided that a user that's part of a team or role can see who else has that same team or role?

> __Note__
> ~~Question, should we add a policy to ensure that updates to roles marked as "team" have a team and also the opposite?~~
A validator was added for the RoleAssignment input.

## 🧪 Testing

1. Rerun migrations and seed fresh
2. Run phpunit tests (and reseed afterwards)
3. Test using Graphql Playground
4. Check the AC on issue

<details>
  <summary>User query</summary>
  
  ```graphql
query user($id: UUID!) {
  user(id: $id) {
    sub
    roleAssignments {
    	team { name }
      user { email }
      role { name }
    }
  }
}
---
{
  "id": "xxx-yyy-zzz" 
}
  ```
</details>

<details>
  <summary>Roles query</summary>
  
  ```graphql
query roles {
  roles {
    name    
    isTeamBased
    roleAssignments {
    	team { name }
      user { email }
      role { name }
    }
  }
}
  ```
</details>

<details>
  <summary>Team query</summary>
  
  ```graphql
query team($id: UUID!) {
  team(id: $id) {
    name
    roleAssignments {
    	team { name }
      user { email }
      role { name }
    }
  }
}
---
{
  "id": xxx-yyy-zzz"
}

  ```
</details>

<details>
  <summary>Update user mutation</summary>
  
  ```graphql
mutation updateUserAsAdmin($id:ID!, $user:UpdateUserAsAdminInput!) {
  updateUserAsAdmin(id:$id, user:$user) {
    id
    roleAssignments {
      role {name}
      team {name}
    }
  }
}
---
{
  "id": "xxx-yyy-zzz",
  "user": {
    "roles": {
      "attach":  {
        "roles": ["xxx-yyy-zzz"],
        "team": "xxx-yyy-zzz"
      }
    }
  }
}

  ```
</details>

<details>
  <summary>role_user SQL query</summary>
  
  ```sql
select
  u.id user_id, u.email,
  r.id role_id, r."name" role_name,
  t.id team_id, t."name" team_name 
from users u 
left join role_user ru on u.id = ru.user_id 
left join roles r  on ru.role_id = r.id 
left join teams t on ru.team_id = t.id 
where u.sub = 'applicant@test.com'
  ```
</details>

## 📸 Screenshot

image


api/graphql/schema.graphql
```diff
@@ -532,6 +544,24 @@ type Role {
...
+  id: ID!
+  role: Role @belongsTo
+  team: Team @belongsTo
+  user: UserPublicProfile @morphTo
+}
+
+# stub for Laratrust
+type Permission {
+  id: ID!
 }
```
tristan-orourke, line 565: I don't think we need to add this to schema yet if we have no use for it.

api/graphql/schema.graphql
```diff
@@ -532,6 +544,24 @@ type Role {
   displayName: LocalizedString @rename(attribute: "display_name")
   description: LocalizedString
   isTeamBased: Boolean @rename(attribute: "is_team_based")
+  # teams and roles: built-in
+  users: [UserPublicProfile!] @morphToMany
+  permissions: [Permission!] @belongsToMany
```
tristan-orourke, line 549: I don't think we have any use for querying permissions yet. I think this could be deleted.

api/graphql/schema.graphql
```diff
@@ -532,6 +544,24 @@ type Role {
   displayName: LocalizedString @rename(attribute: "display_name")
   description: LocalizedString
   isTeamBased: Boolean @rename(attribute: "is_team_based")
+  # teams and roles: built-in
+  users: [UserPublicProfile!] @morphToMany
```
tristan-orourke, line 548: I have a few thoughts about the `users` and `roleAssignments` relationships here.
First: do we need both, or should we just rely on roleAssignments?
Second: they need guarding with `@can` or `@canOnParent` directives since the roles themselves are public but their users are not. But I'm not exactly sure what the guard should be. One easy option is to just remove these relationships from the Role schema, since we don't currently have anywhere that needs to get the list of users by Role. Otherwise, maybe we use something like `@can(ability: 'viewAnyPublicProfile')` and `@can(ability: 'viewAnyTeamMembers', model: 'Team') respectively? (I'm trying to think based on our [permissions](https://docs.google.com/spreadsheets/d/1RC7wufxwjLEvaEunHsfZfOzBefg218dvvzkDwKSiQjI/edit#gid=428263758), though we can always add more permissions if necessary.)

api/graphql/schema.graphql
```diff
@@ -524,6 +532,10 @@ type Team {
   departments: [Department] @belongsToMany
   contactEmail: Email @rename(attribute: "contact_email")
   pools: [Pool] @hasMany @can(ability: "viewAny")
+  # teams and roles: built-in
+  users: [UserPublicProfile!] @morphToMany
+  # teams and roles: custom pivot model
+  roleAssignments: [RoleAssignment!] @hasMany
```
tristan-orourke, line 538: I'm wondering if we want to expose the users relationship at all, since it's kind of redundant with rolesAssignements. (We could consider it to have some worth as a convenience method, since it saves the client from doing a bit of processing if they just want the list of unique users. Actually, question, does it handle filtering out duplicate users?)

The other thing is these fields definitely need `@can` directives, since we have different permissions for viewing teams, and viewing team members. I figure both of these would be protected with `@canOnParent(ability: 'viewTeamMembers')`.

api/graphql/schema.graphql
```diff
@@ -131,6 +131,14 @@ type User {
   isProfileComplete: Boolean
   expectedGenericJobTitles: [GenericJobTitle] @belongsToMany
   priorityWeight: Int @rename(attribute: "priority_weight")
+
+  # teams and roles: built-in
+  roles: [Role!] @morphToMany
+  rolesTeams: [Team!] @morphToMany
+  permissionsTeams: [Permission!] @morphToMany
+  permissions: [Permission!] @morphToMany
```
tristan-orourke, line 139: We don't need to expose all the relationships we have available. I kind of feel like we should just give them roleAssignments. But I'm open to leaving `roles` and adding a `teams` relationship (aliased to rolesTeams) if you feel its valuable.

api/app/Models/RoleAssignment.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        return $this->belongsTo(Role::class);
+    }
+
+    public function team(): BelongsTo
+    {
+        return $this->belongsTo(Team::class);
+    }
+
+    public function user()
```
tristan-orourke, line 37: Does this need a typehint?

tristan-orourke (changes requested):
> I know @tristan-orourke and I disagreed about where the mutation should be (standalone vs. part of user). I really like how it turned out as part of user.

So you're right, this looks good like this, and makes sense as part of an updateAsAdmin mutation. I'm still a bit worried about the future, since we've [considered](https://docs.google.com/spreadsheets/d/1RC7wufxwjLEvaEunHsfZfOzBefg218dvvzkDwKSiQjI/edit#gid=428263758) giving some roles permission to add people to a team (assign-team-role/assign-any-role) without giving them permission to change anything else about a user (update-any-user). I'm not sure how to handle that situation without redesigning the mutation, but I'm open to ideas. Also, we could cross that bridge when we come to it.

> questions about protecting fields

I left some specific comments about `@can` and `@canOnParent` directives, but to some of your specific points...

> All references to user just use UserPublicProfile.

Right, that's helpful, though we also want to be careful about access to that, as its less "public" than its name suggests... 😅 

> The top level teams and roles queries already have "admin only" policies.

I expect that to change, actually.

> The user fields already can only be accessed by admin or self.

👍 

> I think we decided that a user that's part of a team or role can see who else has that same team or role?

We decided that a user in a team can see the others in the team. I don't remember deciding to extend that to roles. I'll have to think about that...

> Question, should we add a policy to ensure that updates to roles marked as "team" have a team and also the opposite?

Yeah, I think we should, though maybe we should check that in a Validator instead of a policy?

api/graphql/schema.graphql
petertgiles: Yeah, let's not add things we don't need yet.  Removed in [remove extra relationships](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5749/commits/69500c02baa6a090cdde1213656bb8672c78251c).

api/graphql/schema.graphql
petertgiles: Yeah, let's not add things we don't need yet.  Removed in [remove extra relationships](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5749/commits/69500c02baa6a090cdde1213656bb8672c78251c).

api/graphql/schema.graphql
petertgiles: Removed in [remove extra relationships](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5749/commits/69500c02baa6a090cdde1213656bb8672c78251c).

api/graphql/schema.graphql
petertgiles: Removed in [remove extra relationships](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5749/commits/69500c02baa6a090cdde1213656bb8672c78251c).

api/graphql/schema.graphql
petertgiles: Removed in [remove extra relationships](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5749/commits/69500c02baa6a090cdde1213656bb8672c78251c).

api/graphql/schema.graphql
petertgiles: Added viewTeamMembers guard in [Update guards](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5749/commits/804da4ff4cf20bfa9513bcce20f8d8c1b32a86cc).

tristan-orourke:
> I'm still a bit worried about the future, since we've [considered](https://docs.google.com/spreadsheets/d/1RC7wufxwjLEvaEunHsfZfOzBefg218dvvzkDwKSiQjI/edit#gid=428263758) giving some roles permission to add people to a team (assign-team-role/assign-any-role) without giving them permission to change anything else about a user (update-any-user).

Hey, about the user mutation... field-level permissions might actually work on GraphQL input types as well! That might solve the problem I'm worried about.

petertgiles:
> Hey, about the user mutation... field-level permissions might actually work on GraphQL input types as well! That might solve the problem I'm worried about.

Nice!  I hope so.  :smile:  Since we're trying to simplify and deduplicate our schema this seems like a preferred approach.

petertgiles:
> Yeah, I think we should, though maybe we should check that in a Validator instead of a policy?

Validator added to input type along with tests to verify functionality.

api/app/Policies/TeamPolicy.php
```diff
@@ -68,4 +68,19 @@ public function update(User $user = null)
...
+     * @param  \App\Models\User|null  $user
+     * @return \Illuminate\Auth\Access\Response|bool
+     */
+    public function viewAnyTeamMembers(User $user = null)
+    {
+        if ($user) {
+            return $user->isAdmin();
+        }
+        return false;
+    }
```
tristan-orourke, line 85: Unless the policy needs to return true for guest users, you can leave off the `= null` in the argument, and assume its not null.
```suggestion
    public function viewAnyTeamMembers(User $user)
    {           
        return $user->isAdmin();
    }
```

petertgiles: :+1:  [Make user model not-null](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5749/commits/31fcab5dbbbd25710506262e70b2f6483ad95de5)

tristan-orourke (changes requested):
Mostly looks good! Need to change the seeder a bit, and I'm still thinking about the Role.roleAssignments guard, because I agree with you, viewAnyTeamMembers isn't quite right. 🤔

api/graphql/schema.graphql
tristan-orourke: Hmmm. Yeah, Role.roleAssignments is a tricky one. 🤔

api/graphql/schema.graphql
```diff
@@ -532,6 +537,17 @@ type Role {
   displayName: LocalizedString @rename(attribute: "display_name")
   description: LocalizedString
   isTeamBased: Boolean @rename(attribute: "is_team_based")
+  roleAssignments: [RoleAssignment!]
+    @hasMany
+    @can(ability: "viewAnyTeamMembers", model: "Team")
```
tristan-orourke, line 542: > I've added the guard to roleAssignments but this one is a bit strange since many roles have roleAssignements but there is no team to query a policy from. I could make a 'viewAnyRoleMembers' policy on Roles and use that instead?

I missed this comment on my last review - yes, I think you nailed it with your suggestion to just make a new policy method on Role 👍 . We'll make a new permission to go along with it and give it to Platform Admins. Call it either view-any-roleMembers or view-any-roleAssignments.

tristan-orourke (changes requested):
The seeder change is great, the last remaining thing I'd like to see is the new roleAssignments policy method.

api/graphql/schema.graphql
petertgiles: Sounds good.  I've added it in [Add role policy to view any assignment](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5749/commits/3ade4ec6d4cddb8d3436c52457874b539ada2f1e).

api/graphql/schema.graphql
petertgiles: :+1: Added in [Add role policy to view any assignment](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5749/commits/3ade4ec6d4cddb8d3436c52457874b539ada2f1e)

tristan-orourke (approved):
👏

PR #5750, opened 2023-02-21
Debt - 5171 - Fix TS Cypress Warning

### LINKED ISSUE #5171: ♻️ Resolve noise of "tsconfig.json missing" warning in logs of local cypress run
## ♻️ Debt/Refactor

Describe the problem with the existing implementation.
Running cypress commands throws an error:

```
Couldn't find tsconfig.json. tsconfig-paths will be skipped
```

## 🕵️ Details

This seems to have started appearing when the cypress v10 upgrade happened.
We are NOT using any ts files in cypress. patcon recalls this as being needed before in order to properly share some config files during cypress runs. aka there was a need, but perhaps it's not needed anymore.

## 📸 Screenshot

Screen Shot 2022-12-16 at 3 23 10 PM

## 🙋‍♀️ Proposed Solution

we can remove references to tsconfig.json in the cypress folder to remove the error :)

## ✅ Acceptance Criteria

- [ ] Cypress CI still passes
- [ ] All the local `npm run e2e:*` commands in `apps/e2e` still work fine.

esizer:
🤖 Resolves #5171 

## 👋 Introduction

This solves the `tsconfig.json` warning that shows up in the console while running cypress tests.

## 🕵️ Details

Turns out it was fairly simple, not sure I feel great merging this with a 13 story point estimation 😅 

## 🧪 Testing

1. Run `npm run e2e:run:all`
2. Confirm the warning does not appear
3. Confirm tests are still passing

## 📸 Screenshot

Screenshot 2023-02-21 at 3 04 14 PM

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -67,7 +67,7 @@ jobs:
         with:
           spec: "**/*.cy.js"
           browser: chrome
-          config-file: cypress.config.js
+          config-file: cypress.config.ts
           project: ./apps/e2e
           # See: https://github.com/cypress-io/github-action/issues/489#issuecomment-1021379037
           command-prefix: "--"
--- a/apps/e2e/cypress.config.ts
+++ b/apps/e2e/cypress.config.ts
@@ -1,8 +1,8 @@
-const { defineConfig } = require('cypress')
+import { defineConfig } from "cypress"
 
 const extendedTimeout = 60000;
 
-module.exports = defineConfig({
+export default defineConfig({
   defaultCommandTimeout: process.env.CYPRESS_EXTEND_TIMEOUTS ? extendedTimeout : 8000,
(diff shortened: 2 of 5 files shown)


esizer:
> The warning is gone, but a different error has arrived 😢

Looks like it was because we added TS without using any TS files. Fixed the issue you raised with [update cypress config to fix no inputs error](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5750/commits/a7578213db8be02cac04977d2a6393abfc61535f). However, maybe we should work on updating all of our cypress files to TS instead of JS? 🤔

mnigh (approved):
nice work! 👍

✅ warning does not appear
✅ tests are still passing

mnigh:
> However, maybe we should work on updating all of our cypress files to TS instead of JS? 🤔

That seems reasonable to me to move to TS since the rest of the project uses it. Perhaps a new **debt** issue?

PR #5792, opened 2023-02-23
Bug -  5791 - Fix watch script

### LINKED ISSUE #5791: 🐛 Watch command only running for `apps/web`
## 🐛 Bug

Our watch command is not watching for changes in `packages/*`

## 🦋 Expected Behaviour

Changes to `packages/*` trigger a rebuild when using `npm run watch`

## 📋 Steps to Reproduce

1. Run `npm run watch`
2. Make change in `packages/*`
3. Observe packing not being rebuilt

## ✅ Acceptance Criteria

- [ ] Watch command does not filter
- [ ] Changes in `packages/*` rebuild package

esizer:
🤖 Resolves #5791 

## 👋 Introduction

This allows us to `npm run watch` all packages instead of just `apps/web`

## 🕵️ Details

Now that the restructuring happened, we are able to watch all packages and apps.

## 🧪 Testing

1. Run `npm run watch`
2. Make change in any `package/*` or `apps/*`
3. Confirm the package is rebuilt


petertgiles:
Every time I run the watch operation I get a bunch of build errors - different every time.  The app ends up building and running fine but I wonder if maybe there are some missing turbo dependencies.

image

esizer:
> Every time I run the watch operation I get a bunch of build errors - different every time. The app ends up building and running fine but I wonder if maybe there are some missing turbo dependencies.

okay, some of the errors were coming from a weird dependency thing with `@testing-library/react` that was fixed with dependabot this morning but not part of the base. Merged main to resolve that.

This is technically running all watch commands in parallel, so if a package that depends on another are both running at the same time, it will cause issues 😢 However, it should still all work because eventually all packages will be bundled and only the one changed should rebundle. 

Though, this is kinda hacky and it is probably safer to run `npm run watch:package -- --filter=@gc-digital-talent/i18n` for example.

There is [a feature request](https://github.com/vercel/turbo/issues/986) to build this into turborepo that has quite a bit of support behind it.

petertgiles (approved):
Nice, seems to work properly. :+1:  A bit janky with the errors but hopefully Turbo picks up the feature to handle deps in watch soon.

PR #5795, opened 2023-02-23
Feature - 5437 - Assign new user roles

### LINKED ISSUE #5437: ✨ Assign new Roles to existing Users
## ✨ Feature

All existing users need the Base User and Applicant roles. All existing Admin users need the System Admin, Request Responder and Pool Operator (DCM) roles. When any new user is created, they must be assigned the Base User and Applicant roles.

## 🕵️ Details

When the Pool Operator role is attached to a user, it must be [assigned in the context of a team](https://laratrust.santigarcor.me/docs/7.x/usage/teams.html#roles-assignment-removal) (the DCM team, for existing users).

Write a command that syncronizes uses roles based on LegacyRoles. Any users with the Applicant LegacyRole get the new ApplicantRole, any users with the Admin LegacyRole get the new System Admin and Request Responder and Pool Operator ('digital-community-management') roles. Any users without the Admin LegacyRole should lose the system_admin, pool_operator and request_responder roles if they have them.

This command should be safe to run multiple times in production, as user LegacyRoles may change.

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] All existing users have the Base User and Applicant roles. 
- [ ] All existing Admin users need the System Admim, Request Responder, and Pool Operator (DCM) roles
- [ ] The Pool Operator role is only ever attached along with a Team
- [ ] The synchronize command is safe to run periodically in production.
- [ ] Users' LegacyRoles are not changed.

out of scope 

- [ ] When any new user is created, they will be assigned the Base User and Applicant roles.

esizer:
🤖 Resolves #5437 

## 👋 Introduction

This creates the `php artisan role:sync` command to assign new user roles to users.

## 🕵️ Details

This only assigns roles only if the user has no existing roles.

## 🧪 Testing

1. Run `docker-compose exec -w /home/site/wwwroot/api webserver sh -c "php artisan role:sync"`
2. Confirm the appropriate roles are assigned.
3. Confirm `legacy_roles` are untouched
4. Re-run command a few times and confirm no errors

You can check with the following:

>**Note**
> ~~I suck at SQL, so if you know a better way let me know 😅~~

h/t @petertgiles for improving the SQL! 🎉 

```sql
SELECT
    u.legacy_roles,
    count(*),
    string_agg(distinct r.name||coalesce('-'||t.name, ''), ', ')    
FROM users u
    LEFT JOIN role_user ON role_user.user_id = u.id
    LEFT JOIN roles r ON role_user.role_id = r.id
    LEFT JOIN teams t ON role_user.team_id = t.id
group by u.legacy_roles
ORDER by u.legacy_roles
```

### Expectations

- `APPLICANT` / empty - base_user, applicant
- `ADMIN` - `pool_operator` (with dcm team), `request_responder`, `platform_admin`

## 🚀 Deployment Instructions

1. Run seeders
  - `./artisan db:seed --class=TeamSeeder`
  - `./artisan db:seed --class=RolePermissionSeeder`
2. Run command
  - `./artisan role:sync`

## 📸 Screenshot

Screenshot 2023-02-23 141941

### DIFF
--- a/api/app/Console/Commands/RoleSync.php
+++ b/api/app/Console/Commands/RoleSync.php
@@ -46,6 +46,8 @@ class RoleSync extends Command
             // Everyone gets the base roles :)
             $rolesToSync = [$baseUserRole, $applicantRole];
 
+            $user->syncRoles($rolesToSync);
+
             // Add admin roles if user has the legacy admin role
             if (in_array(ApiEnums::ROLE_ADMIN, $user->legacy_roles)) {
                 array_push($rolesToSync, $requestResponderRole, $superAdminRole);
@@ -53,8 +55,6 @@ class RoleSync extends Command
                 // Attach the pool_operator role to DCM team
                 $user->attachRole($poolOperatorRole, $DCMTeam);
             }
-
-            $user->syncRoles($rolesToSync);
         }
 
         return Command::SUCCESS;
--- a/api/app/Console/Commands/RoleSync.php
+++ b/api/app/Console/Commands/RoleSync.php
@@ -47,13 +47,13 @@ class RoleSync extends Command
             $rolesToSync = [$baseUserRole, $applicantRole];
             $attachTeamRole = false;
 
-            $user->syncRoles($rolesToSync);
-
             // Add admin roles if user has the legacy admin role
             if (in_array(ApiEnums::ROLE_ADMIN, $user->legacy_roles)) {
                 array_push($rolesToSync, $requestResponderRole, $superAdminRole);
                 $attachTeamRole = true;
             }
+
+            $user->syncRoles($rolesToSync);
 
             if ($attachTeamRole) {
                 // Attach the pool_operator role to DCM team
--- a/api/app/Console/Commands/RoleSync.php
+++ b/api/app/Console/Commands/RoleSync.php
@@ -42,22 +42,19 @@ class RoleSync extends Command
 
         $users = User::all();
 
-        foreach($users as $user) {
-            // Only sync roles if they have none assigned
-            if(!$user->roles()->exists()) {
-                // Everyone gets the base roles :)
-                $rolesToSync = [$baseUserRole, $applicantRole];
+        foreach ($users as $user) {
+            // Everyone gets the base roles :)
+            $rolesToSync = [$baseUserRole, $applicantRole];
 
-                // Add admin roles if user has the legacy admin role
-                if(in_array(ApiEnums::ROLE_ADMIN, $user->legacy_roles)) {
-                    array_push($rolesToSync, $requestResponderRole, $superAdminRole);
+            // Add admin roles if user has the legacy admin role
+            if (in_array(ApiEnums::ROLE_ADMIN, $user->legacy_roles)) {
+                array_push($rolesToSync, $requestResponderRole, $superAdminRole);
 
-                    // Attach the pool_operator role to DCM team
-                    $user->attachRole($poolOperatorRole, $DCMTeam);
-                }
-
-                $user->attachRoles($rolesToSync);
+                // Attach the pool_operator role to DCM team
+                $user->attachRole($poolOperatorRole, $DCMTeam);
             }
+
+            $user->syncRoles($rolesToSync);
         }
 
         return Command::SUCCESS;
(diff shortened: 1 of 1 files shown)


api/app/Console/Commands/RoleSync.php
```diff
@@ -0,0 +1,65 @@
...
+        $requestResponderRole = Role::where('name', 'request_responder')->sole();
+        $superAdminRole = Role::where('name', 'platform_admin')->sole();
+
+        $DCMTeam = Team::where('name', 'digital-community-management')
+            ->sole();
+
+        $users = User::all();
+
+        foreach($users as $user) {
+            // Only sync roles if they have none assigned
```
petertgiles, line 46: I think we want to be able to run this periodically to ensure that the new roles stay in sync with the old roles as the old roles slowly change with use.  If my user gets migrated and then my legacy_role changes this script will not sync me back up, right?
> This command should be safe to run multiple times in production, as user LegacyRoles may change.

api/app/Console/Commands/RoleSync.php
```diff
@@ -0,0 +1,65 @@
...
+
+                // Add admin roles if user has the legacy admin role
+                if(in_array(ApiEnums::ROLE_ADMIN, $user->legacy_roles)) {
+                    array_push($rolesToSync, $requestResponderRole, $superAdminRole);
+
+                    // Attach the pool_operator role to DCM team
+                    $user->attachRole($poolOperatorRole, $DCMTeam);
+                }
+
+                $user->attachRoles($rolesToSync);
```
petertgiles, line 59: To make it possible to remove roles on a user who already has a roles array, I suggest using `syncRoles` instead of `attachRoles`.

petertgiles:
Also, might be worth putting the deployment tag on this with instructions.  It can only be run if the two seeders have been run, and I don't think they have in production yet:
```
./artisan db:seed --class=TeamSeeder
./artisan db:seed --class=RolePermissionSeeder
```

petertgiles:
FWIW, here's a modification to your query to roll it up.  :smile: 
```sql
SELECT
    u.legacy_roles,
    count(*),
    string_agg(distinct r.name||coalesce('-'||t.name, ''), ', ')    
FROM users u
    LEFT JOIN role_user ON role_user.user_id = u.id
    LEFT JOIN roles r ON role_user.role_id = r.id
    LEFT JOIN teams t ON role_user.team_id = t.id
group by u.legacy_roles
ORDER by u.legacy_roles
```

esizer:
> FWIW, here's a modification to your query to roll it up. 😄

Thanks! I know just enough SQL to get by and had no idea how to do this 🙏

api/app/Console/Commands/RoleSync.php
esizer: Okay, so we are okay with overwriting existing roles? I did this guard in case of the following scenario:

- User has legacy role `APPLICANT`, user was assigned the new role `request_responder`. 
- Command is run
- sees `APPLICANT` legacy role
- syncs new `base_user` and `applicant` role
- sync removes the `request_responder` role

My interpretation of "safe to run multiple times" was that we could run it without overwriting users with the new roles and only affect those with legacy roles, updating them to the new roles once.

petertgiles: My understanding is that this sync command is the only way a user will get new roles until we "cut over" so we won't have to worry about overwriting a new role that was manually added.  Worth bringing up in scrum today to confirm, though.

petertgiles: We discussed this in scrum today.  My understanding is correct - this command should overwrite whatever "new roles" the user has to sync them up from "old roles" freshly.  Once we cutover to managing new roles directly we will stop using this command.

esizer: Thanks, made that change in [update to overwrite roles on each run](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5795/commits/2ae3e4bb619289131d53f09e3264fce889bdfbf5)

api/app/Console/Commands/RoleSync.php
esizer: Updated in [update to overwrite roles on each run](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5795/commits/2ae3e4bb619289131d53f09e3264fce889bdfbf5)

PR #5801, opened 2023-02-24
Feature -  5776 - Update basic policies

### LINKED ISSUE #5776: Update Classification, Department, Skill and Skill Family Policy
## ✨ Feature

A sub-task of #5438.

## 🕵️ Details

This story only cares about Classification, Department, Skill and Skill Family Policy methods should use the following permissions respectively

- view-any-classification
- create-any-classification
- update-any-classification
- delete-any-classification
- view-any-department
- create-any-department
- update-any-department
- delete-any-department
- view-any-skill
- create-any-skill
- update-any-skill
- delete-any-skill 
- view-any-skillFamily
- create-any-skillFamily
- update-any-skillFamily
- delete-any-skillFamily

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ]  Classification, Department, Skill and Skill Family policy methods now work based on the new permissions, instead of the `isAdmin` role
- [ ] Roles follow the permissions in the excel sheet https://docs.google.com/spreadsheets/d/1RC7wufxwjLEvaEunHsfZfOzBefg218dvvzkDwKSiQjI/edit#gid=428263758 
- [ ] All top-level mutations and policies for Classification, Department, Skill and SkillFamily have a `@can` directive

esizer:
🤖 Resolves #5776

## 👋 Introduction

Updates Classification, Department, Skill and Skill Family policies to use new permissions.

## 🕵️ Details

This updates all of the policies to use the Laratrust permissions and adds some tests to ensure we get the proper responses back from the API. This also adds a department factory which was not available before.

> **Note**
> I started with the `view-any-{resource}` but that broke the search page since we can view those as anonymous users. Reverted in: [allow anonymous users to view basic resources](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5801/commits/e8eff41417f96c405cb539365ad25b921bca127d)

## 🧪 Testing

1. Run PHPUnit inside the docker container (cant run it normally)
2. Confirm each of the following passes:
    -  `./vendor/bin/phpunit ./tests/Feature/ClassificationTest.php"`
    -  `./vendor/bin/phpunit ./tests/Feature/DepartmentTest.php"`
    -  `./vendor/bin/phpunit ./tests/Feature/SkillTest.php"`
    -  `./vendor/bin/phpunit ./tests/Feature/SkillFamilyTest.php"`


api/app/Policies/ClassificationPolicy.php
```diff
@@ -41,7 +41,7 @@ public function view(User $user = null)
      */
     public function create(User $user)
     {
-        return $user->isAdmin();
+        return $user && $user->isAbleTo("create-any-classification");
```
petertgiles, line 44: Not a big deal but PHP 8 has null safe method operator. https://stitcher.io/blog/php-8-nullsafe-operator  Since this is the first PR of the set it will probably be copied for all the rest.

petertgiles (commented):
This looks really nice and clean.  I'd like to discuss your question in scrum before approving though.
> I started with the view-any-{resource} but that broke the search page since we can view those as anonymous users.

api/app/Policies/ClassificationPolicy.php
```diff
@@ -53,7 +53,7 @@ public function create(User $user)
      */
     public function update(User $user, Classification $classification)
     {
-        return $user->isAdmin();
+        return $user && $user->isAbleTo("update-any-classification");
```
brindasasi, line 56: can we remove these unused params? classification , department , skillFamily etc ?

petertgiles:
> I started with the view-any-{resource} but that broke the search page since we can view those as anonymous users. 
We discussed this in scrum today.  The team is OK with with simply return `true` and not having an explicit check against the guest role.

petertgiles (approved):
There are a few small tidying suggestions but this looks really good! :+1:

api/app/Policies/ClassificationPolicy.php
esizer: Which method do we prefer? Should I update to setting `null` by default or use the method documented?

api/app/Policies/ClassificationPolicy.php
esizer: Removed in [add note, remove unused params](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5801/commits/4af9141d0e361ce1acecfca2bf73f760613962cf)

esizer:
> > I started with the view-any-{resource} but that broke the search page since we can view those as anonymous users.
> 
> We discussed this in scrum today. The team is OK with with simply return `true` and not having an explicit check against the guest role. It's recommended to put in a comment, something like "This action is possible for everyone, including anonymous users".

Thanks! Added in [add note, remove unused params](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5801/commits/4af9141d0e361ce1acecfca2bf73f760613962cf)

brindasasi (approved):
LGTM

api/app/Policies/ClassificationPolicy.php
esizer: okay, it has been updated to use the documented method.

[update to allow guest users](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5801/commits/3dbc9e7d6d7edd0df4099987aedcc19995b5bb99)

PR #5843, opened 2023-03-02
bug/5837-cypress-test

### LINKED ISSUE #5837: Cypress test admin/pool-candidates often fails
## 🧪 Tests

The admin/pool-candidates test often fails.  Rerunning it frequently allows it to pass

## 🕵️ Details

image

## ✅ Acceptance Criteria

- [ ] Test reliably passes unless there is an actual problem

vd1992:
🤖 Resolves #5837 

## 👋 Introduction

That bug seems annoying, so lets get rid of it.
Hopefully resolves the pool-candidate spec issue.

## 🕵️ Details

It appears that the test pool isn't always on the first page of pools, which causes the test to fail due to not finding it.
So searching for it, along with the should("exist") to make the search run. 

## 🧪 Testing

1. cypress keeps passing

image

### DIFF
--- a/apps/e2e/cypress/e2e/admin/pool-candidate.cy.js
+++ b/apps/e2e/cypress/e2e/admin/pool-candidate.cy.js
@@ -96,7 +96,12 @@ describe("Pool Candidates", () => {
 
     cy.wait("@gqlgetPoolsQuery");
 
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
seems good!  :+1:

PR #5853, opened 2023-03-03
Feature - 5782 - Update role policy

### LINKED ISSUE #5782: Role Policy

esizer:
🤖 Resolves #5782

## 👋 Introduction

Updates the `UserPolicy` and `RolePolicy` with the new Laratrust permissions.

## 🕵️ Details

The `assign` permissions were added to the `UserPolicy` if the user is attempting update a user's roles using the [`injectArgs`](https://lighthouse-php.com/5/security/authorization.html#restrict-fields-through-policies:~:text=checks%20with%20the-,injectArgs,-argument%3A) option in lighthouse-php.

## 🧪 Testing

1. Run PHPUnit and confirm tests pass for `UserRoleTest`
2. Review the policies to confirm they match [the spreadsheet](https://docs.google.com/spreadsheets/d/1RC7wufxwjLEvaEunHsfZfOzBefg218dvvzkDwKSiQjI/edit#gid=428263758)

### DIFF
--- a/api/app/Policies/UserPolicy.php
+++ b/api/app/Policies/UserPolicy.php
@@ -3,7 +3,9 @@
 namespace App\Policies;
 
 use App\Models\User;
+use App\Models\Team;
 use Illuminate\Auth\Access\HandlesAuthorization;
+use Illuminate\Support\Facades\Log;
 
 class UserPolicy
 {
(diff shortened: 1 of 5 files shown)


api/app/Policies/UserPolicy.php
```diff
@@ -80,8 +92,8 @@ public function update(User $user, User $model)
      */
     public function delete(User $user, User $model)
     {
-        if($user->isAdmin()){
-            if($user->id !== $model->id){
+        if ($user->isAdmin()) {
```
brindasasi, line 95: I believe this needs to be changed to $user->hasRole(`platform_admin`)

esizer: Agreed, however this is being worked on in https://github.com/GCTC-NTGC/gc-digital-talent/pull/5858

The only reason this touches the `UserPolicy` is because the `assign-any-role` happens on a user rather than a role so it made more sense here.

brindasasi: ok got you 👍

brindasasi (approved):
LGTM!

PR #5858, opened 2023-03-06
5784 update user policy

### LINKED ISSUE #5784: Update User Policy (and Experience Policy) to use new permissions
## ✨ Feature

A sub-task of #5438.

## 🕵️ Details

[Permissions table](https://docs.google.com/spreadsheets/d/1RC7wufxwjLEvaEunHsfZfOzBefg218dvvzkDwKSiQjI/edit?usp=sharing)

This story only cares about protecting Users and user subtypes, like Applicant.

Note that in our listed permissions, Applicants are known as applicantProfiles. Accordingly, user policy methods should expect to check the following permissions:

- view-team-applicantProfile
- view-any-applicantProfile
- view-any-user
- view-any-userBasicInfo
- view-own-user
- update-any-user
- update-own-user
- delete-any-user
Also note that, since Experiences are strictly a part of the ApplicantProfile, the ExperiencePolicy methods for viewing should refer to the methods for viewing Users or Applicants, and creating/updating/deleting Experiences should depend on the ability to update the related user/applicant.

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] User and Experience policy methods now work based on the new permissions, instead of the `isAdmin` role
- [ ] All top-level mutations and queries for Users, Applicants and Experiences have a `@can` directive
- [ ] For any queries whose policy methods can't be implement with new permissions, deprecate the query/mutation and add a `TODO: delete me when my query/mutation gets removed` in the required policy method.

tristan-orourke:
🤖 Resolves #5784

## 👋 Introduction

Updates UserPolicy to use new permissions. Updates ExperiencePolicy to defer to UserPolicy, since creating/editing/deleting Experiences is essentially part of updating a User.

## 🕵️ Details

Updated the PoolCandidate notDraft scope to use submittedAt now.

## 🧪 Testing

1. Test that you can still update your profile as an applicant
2. Test that you can still view applicant profiles of the users who applied to your pools, as an admin

## 📸 Screenshot


api/tests/Unit/UserPolicyTest.php
```diff
@@ -0,0 +1,214 @@
...
+     * @return void
+     */
+    public function testViewAny()
+    {
+        $this->assertFalse($this->guest->can('viewAny', User::class));
+        $this->assertFalse($this->applicant->can('viewAny', User::class));
+        $this->assertFalse($this->poolOperator->can('viewAny', User::class));
+        $this->assertFalse($this->requestResponder->can('viewAny', User::class));
+        $this->assertTrue($this->platformAdmin->can('viewAny', User::class));
+    }
```
esizer, line 109: Just popping in to ask, does this work? 😅  I was trying this but it just kept returning `null` every single time 😖

tristan-orourke: Yep, this works! I had to add `use Authorizable;` to the User model.

esizer: 🤦‍♀️ Thanks! That was what I was missing all this time 😝

api/app/Policies/UserPolicy.php
```diff
@@ -92,62 +90,46 @@ public function update(User $user, User $model, array $injected)
...
+    public function viewApplicant(User $user, User $model)
     {
-        return $user->isAdmin();
+        return $user->isAbleTo('view-any-applicantProfile')
+            || (
+                $user->isAbleTo('view-team-applicantProfile')
+                && $this->applicantHasAppliedToPoolInTeams($model, $user->rolesTeams()->pluck('id')
+            )
+        );
     }
```
vd1992, line 124: this doesn't allow an applicant to view themselves
which is needed for queries such as 

image

resulting in an inability to access experiences page, and therefore not able to complete applications either
used in 

image

vd1992 (changes requested):

vd1992:
is this desired behaviour? you can see the application but not profiles for people in your pool
it's awkward like this, maybe it should always have a link to application?
(no platform_admin role, but pool operator)

https://recordit.co/q46vzVFnnA

this restriction probably comes from

image

api/app/Policies/UserPolicy.php
```diff
@@ -68,19 +65,20 @@ public function create(User $user)
...
+        if ($injected && isset($injected["roles"])) {
             if (!$user->isAbleTo("assign-any-role")) {
                 return false;
             }
         }
-
-        return $user->isAdmin() || $user->id === $model->id;
+        // TODO: Right now, for a user to assign-any-role they ALSO need to be able to update-any-user! That doesn't quite match the permissions table.
+        return $user->isAbleTo('update-any-user')
+            || ($user->isAbleTo('update-own-user') && $user->id === $model->id);
```
esizer, line 81: Just confirming, is this a future TODO or were you planning on getting to this in this PR?

api/app/Policies/UserPolicy.php
```diff
@@ -92,62 +90,46 @@ public function update(User $user, User $model, array $injected)
...
-        if ($user->isAdmin()) {
-            if ($user->id !== $model->id) {
-                return true;
-            } else {
-                return false; // Do not allow user to delete their own model.
-            }
-        } else {
-            return false;
-        }
+        return $user->isAbleTo('delete-any-user') && $user->id !== $model->id; // Do not allow user to delete their own model.
```
esizer, line 93: Question about data retention here; is it not a privacy concern that we do not allow users to delete their own data? 🤔

esizer (commented):
Cool! I just have some questions

api/app/Policies/UserPolicy.php
tristan-orourke: Hah! Fair question. But we don't really have a mechanism designed for users to delete their own data. This is to ensure we don't break the system by deleting every admin by accident.

api/app/Policies/UserPolicy.php
tristan-orourke: Done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/5858/commits/58d765d99a3f7be0df977e32c8900357b121170f

api/config/rolepermission.php
```diff
@@ -228,6 +228,10 @@
             'en' => 'View any submitted Applications',
             'fr' => 'Voir n\import quelle candidature soumises à n\'importe quel bassin.'
         ],
+        'view-own-applicantProfile' => [
+            'en' => 'View own Applicant Profile.',
+            'fr' => 'Voir son propre profil de candidat.'
+        ],
```
vd1992, line 234: should probably also update RolePermissionTest.php with this
ensuring we have 100% (ideally) coverage of every added role

tristan-orourke:
@vd1992 good catch on the user profile issue. I moved it into its own bug report #5926

vd1992 (approved):
that's all my feedback addressed, all the policy + role + permission layers can get confusing 😵‍💫

PR #5860, opened 2023-03-06
bug/5842-codegen-turbo overcaching

### LINKED ISSUE #5842: Turbo overcaching codegen
## 🛠️ Tooling

Turbo is overcaching codegen and not rebuilding the generated files when changing graphql files.

## 🙋‍♀️ Proposed Solution

Add inputs to the codegen step:
` "inputs": ["src/**/*.graphql", "../../packages/**/src/**/*.graphql"]`

## ✅ Acceptance Criteria

- [ ] Changing a graphql file and running codegen causes the generated files to be properly rebuilt

vd1992:
🤖 Resolves #5842 

## 👋 Introduction

Hopefully patch that graphql caching matter

## 🕵️ Details

Just implementing Peter's suggestion 

## 🧪 Testing

1. npm run build
2. make changes to a graphql file
3. npm run build
4. assert the generated file was fresh

I tested by adding another return field to

image

and ensured the type was updated where the query was used

image

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -9,6 +9,7 @@
         "outputs": ["src/assets/css/hydrogen.css"]
     },
     "codegen": {
+      "inputs": ["src/**/*.graphql", "../../packages/**/src/**/*.graphql"],
       "outputs": ["src/api/generated.ts", "src/index.ts"]
     },
     "dev": {


petertgiles:
Thanks for picking this up!

I think it's a undercaching a little now.  Certainly better than before!  If I change a graphql file and rerun `npm run dev` then it rebuilds and recompiles.  If I run `npm run dev` a second time I don't get "full turbo" as expected though - it rebuilds `web` again.  I have to run it a third time before everything runs fully cached.  Any idea why?

PR #5863, opened 2023-03-07
Debt - 4781 - Updates admin manifest files

### LINKED ISSUE #4781: Manifest file does not contain a name or short_name value in admin
## Description
Manifest file does not contain a `name` or `short_name` value in admin:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/e33baa6f5f82dcb459f7f3775f10b252a97afeb4/apps/web/public/admin.webmanifest#L1

## Acceptance Criteria
- [ ] add values for `name` or `short_name`

mnigh:
🤖 Resolves #4781.

## 👋 Introduction

This PR updates webmanifest files and meta images (i.e. favicon) for `admin`.

## 🕵️ Details

[Replaces CMO favicon images with Digital Talent ones](https://github.com/GCTC-NTGC/gc-digital-talent/commit/a928885a86233c85afb9187eadedcbb74b5c7929)
> **Note**
> `admin` is no longer just CMO and so the images no longer make sense.

## 🧪 Testing

1. Navigate to `/admin`
2. [Inspect](https://firefox-source-docs.mozilla.org/devtools-user/application/manifests/index.html#inspecting-web-app-manifests) web app manifests
3. Verify manifest has correct values

## 📸 Screenshot

Screen Shot 2023-03-07 at 14 06 00


esizer (approved):
Nice! I'm seeing the following error but it seems out of scope:

"Manifest 'start_URL' is not valid"

mnigh:
> Nice! I'm seeing the following error but it seems out of scope:
> 
> "Manifest 'start_URL' is not valid"

Weird, can you post a screenshot? What about for `digital-talent` and `iap`?

mnigh:
> > Weird, can you post a screenshot? What about for `digital-talent` and `iap`?
> 
> Seems to be all. Also, I guess it is only to allow installing the app which we don't yet so I don't think it is a real issue.

Thanks for the screenshot. Looks like Edge supports [start_url](https://developer.mozilla.org/en-US/docs/Web/Manifest/start_url) but Firefox does not. Manifest files are weird!

esizer:
> Thanks for the screenshot. Looks like Edge supports [start_url](https://developer.mozilla.org/en-US/docs/Web/Manifest/start_url) but Firefox does not. Manifest files are weird!

Yeah, it is a required field for installable PWA's which is not even fully supported and not something we even have on our radar yet so i don't think we need to worry about it.

Screenshot 2023-03-07 150445

PR #5899, opened 2023-03-09
Feature/5783 update team policies

### LINKED ISSUE #5783: Update Team Policy to use new permissions
## ✨ Feature

A sub-task of #5438.

## 🕵️ Details

[Permissions table](https://docs.google.com/spreadsheets/d/1RC7wufxwjLEvaEunHsfZfOzBefg218dvvzkDwKSiQjI/edit?usp=sharing)

This story only cares about Team policy methods. The policy should expect to check the following permissions:

- view-any-team
- create-any-team
- update-any-team
- delete-any-team
- view-any-teamMembers
- view-team-teamMembers

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ]  Team policy methods now work based on the new permissions, instead of the `isAdmin` role
- [ ] All top-level mutations and queries for Teams have a `@can` directive
- [ ] For any queries whose policy methods can't be implement with new permissions, deprecate the query/mutation and add a `TODO: delete me when my query/mutation gets removed` in the required policy method.

yonikid15:
🤖 Resolves #5783 

## 👋 Introduction

Adds new permissions to the Team policies (replacing the previous `$user->isAdmin`): 

- view-any-team
- create-any-team
- update-any-team
- delete-any-team
- view-any-teamMembers
- view-team-teamMembers

## 🕵️ Details

I don't see the `view-team-teamMembers` being used in the schema anywhere, so I didn't add a php test for that policy method. Is the permission getting called some other way I'm not sure about?

## 🧪 Testing

1. Look at TeamPolicy.php file and ensure all policy methods are correct.
2. Review that new `deleteTeam` mutation works correctly.
3. Review TeamTest.php file and ensure all unit tests are there and work correctly (not including a test for `view-team-teamMembers`).


yonikid15:
Looks like there is a test failing that ensures a `base_user` (applicant) cannot query teams. However, looking at the Roles and Permissions table and the `rolepermission.php` file, it looks like we are allowing the `base_user` access to view the teams. Maybe @tristan-orourke can clear this up?

yonikid15:
> 
> Screen Shot 2023-03-10 at 10 24 52 AM
> 
> Looks like per the table , we should allow it. Will wait for Tristan confirmation

I updated the old test to ensure base users cannot view team members instead of teams. The change is made in this [74c7334](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5899/commits/74c7334ba0d0d1d105efd550fb2aa1567ee838e7).

api/app/Policies/TeamPolicy.php
```diff
@@ -18,7 +18,7 @@ class TeamPolicy
      */
     public function viewAny(User $user)
     {
-        return $user->isAdmin();
+        return $user->isAbleTo("view-any-team");
```
brindasasi, line 21: viewAny & view  both functions are same?

brindasasi (changes requested):
nit pick: can we remove unused parameters from all these methods?

api/app/Policies/TeamPolicy.php
```diff
@@ -66,6 +77,18 @@ public function update(User $user)
...
+    /**
+     * Determine whether the user can view a specific teams, team members.
+     * Likely to be updated later to allow the team admin and teammates to view their own team
+     *
+     * @param  \App\Models\User  $user
+     * @return \Illuminate\Auth\Access\Response|bool
+     */
+    public function viewTeamMembers(User $user)
+    {
+        return $user->isAbleTo("view-team-teamMembers");
```
tristan-orourke, line 92: Okay, so this is a team-scoped permission (permissions are named like `action-scope-resource`). That means a user only has this permission in the context of specific teams. To check team-scoped permissions we must [add the relevant team](https://laratrust.santigarcor.me/docs/7.x/usage/teams.html#checking-roles-permissions) to the isAbleTo call. In this case, we must get the relevant team from a second argument passed into the function, like this:
```
    public function viewTeamMembers(User $user, Team $model)
    {
        return $user->isAbleTo("view-team-teamMembers", $model);
     }
```

api/tests/Feature/TeamsTest.php
```diff
@@ -255,4 +279,117 @@ public function testTeamUpdateMutation(): void
...
+    assertEquals($teamMembersCount, 3);
+
+    // assert every team member is present in the response
+    $query->assertJsonFragment(['id' => $poolOperator1->id]);
+    $query->assertJsonFragment(['id' => $poolOperator2->id]);
+    $query->assertJsonFragment(['id' => $poolOperator3->id]);
+  }
+
+
+  // TODO: Create another test policy for view-team-teamMembers where we fetch roleAssignments (users) from a specific team (?).
```
tristan-orourke, line 393: Yeah, this would be good. You can ask for some pairing help if you want.

tristan-orourke (changes requested):
Needs a change and a test for the team-scoped permission.

api/tests/Feature/TeamsTest.php
yonikid15: ... I think I got it 😅. Take a look at the Policy and the test and let me know! The test only covers the policy and not any mutation since it hasn't been used in the api yet.

api/app/Policies/TeamPolicy.php
yonikid15: Thanks Tristan I added this in as well as a another condition to ensure admins can view still view any team as well. Let me know what you think. [8d731ca](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5899/commits/8d731cab9564810270b7b705c6489b2624391666)

api/tests/Feature/TeamsTest.php
tristan-orourke: I'm not sure when it got merged into main, but it actually is used in the API! Take a look at Team.roleAssignments:
image

A good query to test would be 
```
teams {
  roleAssignments {
    role { name }
    user { 
      id
      email
    }
  }
}
```

tristan-orourke: Oh, nevermind, I just realized you're already doing that query in `testViewAnyTeamMembers`

tristan-orourke: Nevermind x2, I just realized that Team.roleAssignments checks view**Any**TeamMembers, not viewTeamMembers. 🤔

tristan-orourke: Okay, I think the solution is to change the Team.roleAssignments policy from viewAnyTeamMembersto the more specific viewTeamMembers, and then rename testViewAnyTeamMembers to testViewTeamMembers, while also checking that the query works when acting as a PoolOperator in the same pool (and not as a PoolOperator in a different pool).

api/app/Policies/TeamPolicy.php
```diff
@@ -66,6 +77,19 @@ public function update(User $user)
...
+
+    /**
+     * Determine whether the user can view a specific teams, team members.
+     * Likely to be updated later to allow the team admin and teammates to view their own team
+     *
+     * @param  \App\Models\User  $user
+     * @param  \App\Models\Team/null $team
+     * @return \Illuminate\Auth\Access\Response|bool
+     */
+    public function viewTeamTeamMembers(User $user, Team $team)
```
tristan-orourke, line 91: I would rename this to viewTeamMembers, similarly to how we name the policy function `view`, not `viewTeam`.

api/tests/Feature/TeamsTest.php
```diff
@@ -216,43 +244,120 @@ public function testTeamUpdateMutation(): void
...
+
+  public function testViewAnyTeamMembers(): void
+  {
+    // Attach all users to team one
+    $this->admin->attachRole("pool_operator", $this->team1);
+    $this->poolOperator1->attachRole("pool_operator", $this->team1);
+    $this->poolOperator2->attachRole("pool_operator", $this->team1);
+
+    // Assert all teams query contains expected results
+    $query = $this->actingAs($this->admin, "api")
```
tristan-orourke, line 318: In addition to testing with admin, try acting as poolOperators from the same or different teams.

api/tests/Feature/TeamsTest.php
```diff
@@ -216,43 +244,120 @@ public function testTeamUpdateMutation(): void
...
+    $teamMembersCount = count($data['team']['roleAssignments']);
+    assertEquals($teamMembersCount, 3);
+
+    // assert every team member is present in the response
+    $query->assertJsonFragment(['id' => $this->admin->id]);
+    $query->assertJsonFragment(['id' => $this->poolOperator1->id]);
+    $query->assertJsonFragment(['id' => $this->poolOperator2->id]);
+  }
+
+  public function testViewTeamTeamMembersPolicy()
```
tristan-orourke, line 345: After my other suggested changes, I think this test becomes unnecessary.

tristan-orourke (changes requested):
Looking good!

My remaining suggestions are basically [this](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5899#discussion_r1142747949)

api/tests/Feature/TeamsTest.php
yonikid15: > Nevermind x2, I just realized that Team.roleAssignments checks view**Any**TeamMembers, not viewTeamMembers. 🤔

😆 ya that's why I left that note about not testing any query. Okay I'll get that done thanks Tristan.

api/tests/Feature/TeamsTest.php
yonikid15: Okay cool, removed in [ba0a1e6](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5899/commits/ba0a1e678c05c58f7c2398c9c86c468c66cf68e1)

tristan-orourke (approved):
Looks great!

brindasasi (approved):
LGTM

PR #5902, opened 2023-03-09
[Feature] Stepper Component

### LINKED ISSUE #5875: Storybook for Stepper component
## ✨ Feature

Steppers display progress through a sequence of logical steps

## 🕵️ Details

This is a linear stepper allows the user to complete the steps in sequence.
User needs to complete a step without any error before unlocking the next step. 
Once the step is unlocked user can go back and forth on the unlocked steps. 
Each step will load the appropriate page for the step completion 

## 🎨 Design File
https://www.figma.com/file/BXYVsu1fo9LS2IySJDUINY/Application-process?node-id=281%3A11717)
## 📸 Screenshot
image
**Icons**
image

for reference
https://mui.com/material-ui/react-stepper

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] component accepts a list of items and each item has a link, label, icon and status
- [ ] status - locked , active , complete , alert
- [ ] link is not rendered if status is locked
- [ ] status determines icon colour and overrides the current icon 
- [ ] jest test to show the component is showing proper state as per the props 

### Accessibility criteria
 - [ ] Use a `<nav>` element labelled using `aria-labelledby` and the stepper title
 - [ ] Add SR-specific text to the stepper title that provides context (i.e. Step 1 of 5 [of the application])
 - [ ] The stepper items should be `<li>` elements nested inside of a `<ul`>
 - [ ] The currently active step should use `aria-current` to indicate it as such
 - [ ] The status bubble/icon should read out SR-specific text that explains other states - "Step 5 has an error" / "Step 1 is complete"
 
## :stop_sign: Blockers

List any issues which must be completed before this one.

```[tasklist]
### Blockers
```

esizer:
🤖 Resolves #5875 

## 👋 Introduction

This creates a generic `<Stepper />` component.

## 🕵️ Details

> **Note**
> I'd like some feedback on the API if possible. I am not super confident it makes sense 🤔 

<details>
<summary>API</summary>

```tsx
type StepState = "active" | "completed" | "disabled" | "error";

type IconType = React.FC<React.SVGProps<SVGSVGElement>>;

type StepType = {
  href: string;
  icon: IconType;
  label: React.ReactNode;
  completed?: boolean; // Results in "completed" state
  error?: boolean; // Overrides "completed" with the "error" state
};

interface StepperProps {
  currentIndex: number; // Overrides <Step /> state if it matches that index and sets any index higher to "disabled"
  headingLevel?: HeadingLevel;
  label: string;
  steps: Array<StepType>;
}
```
</details>

<details>
<summary>Computed State</summary>

There is essentially a priority system based on the `Step.completed`, `Step.error` and `Stepper.currentIndex` props.

1. disabled
    - `currentIndex` is less than the steps index
2. active
    - `currentIndex` is equal to the steps index
4. error
    - `currentIndex` is grater than steps index
    - step has an error or is not complete
5. completed
    - `currentIndex` is grater than steps index
    - step is complete and has no error
</details>

<details>
<summary>Accessibility</summary>

There are a few considerations related to accessibility for this.

REF: https://www.w3.org/WAI/tutorials/forms/multi-page/

1. Wrapped in a `<nav>` element labelled by the `label` prop (which is why it is typed as `string` and no `React.ReactNode`)
2. Links are in an `<ol>` since this is a part of a list of steps where the order is important
3. There is hidden text in each link based on the state:
    - "Completed: " for completed steps
    - "Error: " for steps with errors
    - "Current: " for the active step
4. The active step gets the `aria-current="step"` which will be announced to assistive technology
5. Since you cannot disable a link we add some hacky stuff:
    - `role=link` required by;
    - `aria-disabled=true` to disable the link and announce to assistive technology

</details>

## 📋 TO DO

- [ ] Add translations

## 🧪 Testing

1. Run `npm run test -- --filter=@gc-digital-talent/ui -- Stepper`
2. Confirm no failures
3. Run storybook `npm run storybook`
4. Click around and confirm the state updates accordingly 

## 📸 Screenshot

Screenshot 2023-03-09 144005


substrae:
This is looking 👨‍🍳 🤌 💕

petertgiles (approved):
Looks awsome!  No concerns from me.

packages/ui/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,128 @@
...
+        )}
+        <span
+          data-h2-position="base(absolute)"
+          data-h2-location="base(0, auto, auto, 0)"
+          data-h2-width="base(x1.5)"
+          data-h2-height="base(100%)"
+        >
+          {!last && (
+            <div
+              className="Step__Tail Step__Flair"
```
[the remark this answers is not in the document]
esizer: Yes, they are used to target using the hydrogen child selector.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/e5a6c2efdb0e46735fe645805f248574a2c70e04/packages/ui/src/components/Stepper/utils.ts#L32-L38

packages/ui/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,89 @@
...
+  if (stepIndex === currentIndex) {
+    return "active";
+  }
+
+  if (completed && !error) {
+    return "completed";
+  }
+
+  return "error";
+};
```
tristan-orourke, line 30: I don't think this is quite right. It's possible to complete a step, return to a previous step, and then go back to the later step, so there's no reason steps greater than current must be disabled.

tristan-orourke: Thinking through it a bit more, I think this is the logic:
- If `error ==  true`, it shows red and is not disabled.
- If `complete == true && error !== true`, it shows green and is not disabled.
- The first non-complete step is the active step. It shows red if `error == true`, otherwise purple.
- Other non-complete, non-error steps are disabled.
- The current step is unrelated to the active step. The current step is bolded.
- If all steps are complete, no active step(?)

When we initially place the Stepper component on the page, currentIndex will also likely be the active step, but that may change as you navigate around.

tristan-orourke (changes requested):
Others have reviewed most of it, but I just want to suggest a change to the step state logic.

packages/ui/src/components/Stepper/Stepper.tsx
substrae: Hmm, one thing here - the purple indicator is telling the user where they are not, the most recently completed step. If the user starts a step and navigates away to a previous step, the step they've landed on should be purple, and the last step with information entered should be green (if it was completed) or red (if there's a validation error, which there likely will be if the form is half filled).

esizer: Okay, I think I _may_ have gotten it right in [update stepper state logic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5902/commits/dfd0727cac007fc6417415e8c8d7b46ca8051459) based on this feedback.

packages/ui/src/components/Stepper/Stepper.tsx
```diff
@@ -14,19 +14,19 @@ const deriveStepState = (
...
   }
 
-  if (stepIndex === currentIndex) {
-    return "active";
+  if (error) {
+    return "error";
   }
 
-  if (completed && !error) {
+  if (completed && stepIndex !== currentIndex) {
```
tristan-orourke, line 25: Pedantic comment - just like its not necessary to check for `!error` because it would have triggered a return branch, its not necessary to check stepIndex here.

tristan-orourke (approved):
Yep, based on @substrae's correction to my logic, this works!

packages/ui/src/components/Stepper/Stepper.tsx
esizer: lol right, removed in [remove extraneous conditional in stepper state logic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5902/commits/68615f8bc93a13f551bba7a144ef7fe0dee87ecf)

PR #5904, opened 2023-03-09
Feature - 5878 - pool creation team selection

### LINKED ISSUE #5878: ♻️ Update pool creation with teams
## ♻️ Debt/Refactor

Need to update pool creation to use teams properly. 

## 🕵️ Details

As a placeholder, pool creation automatically connects to the DCM team. This will require updating to connecting a pool to the appropriate team.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/0021af526932cde208fc6dd42da9d31583c0f585/apps/web/src/pages/Pools/CreatePoolPage/CreatePoolPage.tsx#L254-L260

## 🙋‍♀️ Proposed Solution

This can be a bit tricky as a user could be a pool operator for more than one team, in which case,
 `me -> roleAssignments -> poolOperator -> team.id`  
wouldn't be a safe choice.

We don't have a figma design for this, so add the new select input to the form in a way that looks consistent.

## ✅ Acceptance Criteria

- [ ] Instead of automatically adding the dcm team id, Team is selected in a dropdown element on the Create Team page
- [ ] Dropdown is populated with the teams the user belongs to as a pool operator

vd1992:
🤖 Resolves #5878 

## 👋 Introduction

Incorporates team selection in the pool creation process. 

## 🕵️ Details

Just a matter of updating queries and passing in a teams array to the form. 
Overhauled a few things too, like the unused genericJobTitles collection. 

## 🧪 Testing

1. play around with pool creation, try to think of edge cases
2. confirm you can create pools and they attach to teams, and your teams are appearing correctly

a query to check your teams and their associated pools

```
query getMyPools {
  me {
    roleAssignments {
      team {
        id 
        name
        pools {
          id 
          name {
            en
          }
        }
      }
    }
  }
}
```

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/lang/fr.json
+++ b/apps/web/src/lang/fr.json
@@ -6786,10 +6786,6 @@
     "defaultMessage": "Rôles des membres",
     "description": "Label for the input to add roles to a user's team membership"
   },
-  "eW7I5E": {
-    "defaultMessage": "Sélectionner les rôles...",
-    "description": "Placeholder text for role selection input"
-  },
   "guzGyX": {
     "defaultMessage": "Le membre a été retiré de l’équipe avec succès",
     "description": "Alert displayed to user when a team member is removed"
@@ -6809,5 +6805,13 @@
   "1BEtoY": {
     "defaultMessage": "Nom du gestionnaire",
     "description": "Label for the user select field on team membership form"
+  },
+  "hr/i9h": {
+    "defaultMessage": "Sélectionnez une équipe...",
+    "description": "Placeholder displayed for team selection input."
+  },
+  "mOS8rj": {
+    "defaultMessage": "Équipe principale",
+    "description": "Label displayed for selecting what team a new pool belongs to."
   }
 }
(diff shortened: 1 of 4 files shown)


vd1992:
@Jerryescandon 
any opinions on the wording of the new stuff, this bit in particular
opinions as in concerns

image

petertgiles (approved):
Looks great!  :+1:

apps/web/src/lang/fr.json
```diff
@@ -6786,10 +6786,6 @@
     "defaultMessage": "Rôles des membres",
     "description": "Label for the input to add roles to a user's team membership"
   },
-  "eW7I5E": {
-    "defaultMessage": "Sélectionner les rôles...",
-    "description": "Placeholder text for role selection input"
-  },
```
vd1992, line 6792: @petertgiles 
translation added in https://github.com/GCTC-NTGC/gc-digital-talent/pull/5904/commits/268e08b374ad8589dee53eac5e916ee95ac5c5a6
this deletion appears to have been a duplicate that was removed by one of the npm intl commands, translation check is passing

vd1992: I ran a few intl things to try so that's why I am not sure which one removed it 😅

petertgiles: :+1:

PR #5909, opened 2023-03-10
Debt - 5900 - Fix turbo config

### LINKED ISSUE #5900: Full review of turbo.json
## 🛠️ Tooling

Feel like we need to do a full review of the turbo.json along with reading through config options to determine what tasks we want to cache and how. 

## 🕵️ Details

For me, the latest headache is coming from the llloonngggg webpack build that takes nearly a minute to run -_-

Also, we could move to internal packages instead of external 

I feel like it has really simplified the setup script but for interactive development it's often over caching or under caching and not "quite right".

## 🙋‍♀️ Proposed Solution

I believe we need to:
- Cache less
- Rely more on inputs rather than outputs since our outputs are largely not source controlled So don't affect the cache
- maybe move away from external packages are exclusively use internal packages
Internal packages do not need to be bundled so we could drop all of the "^build" depends on so hopefully less points of failure in the cache. I only initially did external packages because there was a discussion about making our packages publishable but I think that was a misstep and is causing me a lot of headaches developing.

The other thought I had was the setup script should probably build in the no cache flag.

## ✅ Acceptance Criteria

- [ ] Turborepo commands work well for setting up the project in CI
- [ ] Turborepo commands work well for setting up the project locally the first time
- [ ] Turborepo never overcaches when updating the project locally
- [ ] Turborepo never overcaches tests and intl commands

esizer:
🤖 Resolves #5900 

## 👋 Introduction

Updates the `turbo.json` and `package.json` to (hopefully) fix some of the issues around over/under caching of turborepo.

## 🕵️ Details

- Adds `inputs` to only look at changes for specific files for the task caching
- Reduces output for cached tasks
- Converts some normal tasks to now use turborepo
- Updates it to invalidate the cache on `package.json` file changes
- Forces a cache refresh on `setup.sh`
- Adds a `build:fresh` script to do a fresh build with no cache

## 🧪 Testing

1. Run each task twice to confirm cache works
2. Change a file from the tasks `inputs` and run again to make sure its cache was invalidated
3. Get a full turbo `npm run build`
4. Run the `setup.sh` script and confirm there is no cache


turbo.json
```diff
@@ -1,39 +1,54 @@
...
-      "outputs": ["dist/**"]
+      "outputs": ["dist/**"],
+      "outputMode": "new-only"
     },
     "build:css": {
-        "outputs": ["src/assets/css/hydrogen.css"]
+      "inputs": ["src/**"],
+      "outputs": ["src/assets/css/hydrogen.css"]
     },
     "codegen": {
```
petertgiles, line 15: I tried changing a graphql file and rerunning `npm run dev` and it used cached `codgen` but reran `intl-extract` and `intl-compile`?

turbo.json
```diff
@@ -1,39 +1,54 @@
...
     },
     "build:intl-cli": {
-      "outputs": ["dist/cli.js"]
+      "inputs": ["src/cli.ts"],
+      "outputs": ["dist/cli.js"],
+      "outputMode": "new-only"
     },
     "intl-extract": {
-      "outputs": ["src/lang/en.json"]
+      "inputs": ["src/**"],
```
petertgiles, line 37: Worth narrowing to just ts and tsx?

turbo.json
```diff
@@ -1,39 +1,54 @@
...
+      "outputs": ["src/lang/en.json"],
+      "outputMode": "new-only"
     },
     "intl-compile": {
+      "inputs": ["src/**"],
```
petertgiles, line 42: Yeah, this hurts that we can't use en.json here.

turbo.json
```diff
@@ -1,39 +1,54 @@
...
     },
     "check-intl": {
+      "inputs": ["src/**"],
       "dependsOn": ["build:intl-cli", "intl-compile", "intl-extract"],
-      "cache": false
+      "cache": false,
+      "outputMode": "new-only"
     },
     "storybook": {
       "dependsOn": ["codegen", "intl-compile"],
```
petertgiles, line 54: This should depend on build:css, too, right?

turbo.json
```diff
@@ -42,10 +57,7 @@
     "build-storybook": {
       "outputs": ["dist/**", "storybook-static/**"]
     },
-    "test": {
-      "outputs": ["coverage/**"],
-      "dependsOn": ["^build"]
-    },
+    "test": {},
```
petertgiles, line 60: Is it possible to do "outputMode" as "new-only" on this?

date-helpers:test refers to some packages and fails to run.  Maybe we need to depend on build:packages? Or avoid referencing packages in tests and refernce the files directly?

petertgiles (commented):
OK, did a first round of testing. :smile:  Let me know if you want to talk through any of this.

turbo.json
```diff
@@ -1,39 +1,54 @@
 {
+  "$schema": "https://turbo.build/schema.json",
+  "globalDependencies": ["package.json"],
```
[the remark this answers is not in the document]
esizer: This key is relative to the `turbo.json` so it references https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/package.json . Added it to invalidate it when that changes. So if you add a global dep or update the `"scripts"` key.

package.json
```diff
@@ -14,20 +14,21 @@
   ],
   "scripts": {
     "build": "h2-build && turbo run build --filter='./apps/*' --filter='!@gc-digital-talent/e2e'",
+    "build:fresh": "npm run build -- --force --no-cache",
     "build:all": "turbo run build",
     "build:packages": "turbo run build --filter='./packages/*'",
     "build:css": "h2-build",
```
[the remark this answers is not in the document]
esizer: I don't think so since we are running a script defined in a dependency and not in one of our `apps/*` or `packages/*`. But my confidence on that is not incredibly high either 😅

turbo.json
esizer: Kind of annoying because it isn't very intuitive but, storybook runs `h2-build` when compiling through webpack so it is not necessary here. We needed to do it this way to get it working with storybooks HMR. I wish we could just put it here and make it explicit but it is hidden away in:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/f5b4498fae499d4fe245645baf4ad07a14de1e80/packages/hydrogen-webpack-plugin/index.js#L18-L24

petertgiles: I'd really love to write a proper h2 webpack plugin someday.

turbo.json
```diff
@@ -1,39 +1,54 @@
...
   "pipeline": {
     "build": {
```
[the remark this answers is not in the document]
vd1992: I deleted turbo cache then ran build twice
🎉 

image

turbo.json
esizer: I looked into it with the packages refactor which resulted in this plugin but it wasn't part of the scope so I did not spend much time on it at all.  I'm kinda waning on webpack after using other bundlers. Its so complex and slow 😢

maintenance/scripts/setup.sh
```diff
@@ -23,5 +23,5 @@ cp .env.example .env --preserve=all
 git config --global --add safe.directory /var/www/html
 cd /var/www/html
 npm install
-npm run build
+npm run build -- --force
```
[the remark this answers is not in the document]
esizer: Good idea! I made that script after doing this change and didn't even think to use that here 😅 

Updated in [update build:fresh script and add to setup](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5909/commits/b5381f7e66782d5321b8a31059b382689b595fea)

package.json
```diff
@@ -14,20 +14,21 @@
   ],
   "scripts": {
     "build": "h2-build && turbo run build --filter='./apps/*' --filter='!@gc-digital-talent/e2e'",
+    "build:fresh": "npm run build -- --force --no-cache",
```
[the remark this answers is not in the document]
esizer: Yeah, that makes total sense 👍 

Updated in [update build:fresh script and add to setup](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5909/commits/b5381f7e66782d5321b8a31059b382689b595fea)

turbo.json
esizer: I think it may be working in [update inputs to make it more specific](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5909/commits/f0fe86dd9c7b0dd8711cd161a2668717d6cdb8a2) 🤔

turbo.json
esizer: Yeah, I think I missed the mark on all these scripts inputs. Updated in [update inputs to make it more specific](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5909/commits/f0fe86dd9c7b0dd8711cd161a2668717d6cdb8a2)

turbo.json
esizer: updated in [update test output mode](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5909/commits/92c11fda42f858b6740be10c75df43940b2bb599)

turbo.json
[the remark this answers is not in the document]
petertgiles: Yep, this works for me now.

turbo.json
petertgiles: worked OK this time :shrug:

turbo.json
petertgiles: You've added en.json but that's not a tracked file, right?

esizer: So, when I was reading the documentation very carefully I noticed it says ["by default all non-git-ignored files in the workspace folder"](https://turbo.build/repo/docs/core-concepts/caching#:~:text=by%20default%20all%20non%2Dgit%2Dignored%20files%20in%20the%20workspace%20folder).

The way Interpreted this is that adding untracked files to the inputs property should override this and allow us to watch for changes in untracked files. From my testing, it **_seems_** to work

petertgiles:
Looking good.  I think just add the hydrogen config to build:css and I think that's enough for this issue.  There might be other small things but this is already 200% better!

petertgiles (approved):
:partying_face:

Issue #5734, opened 2023-02-16
✨ PoolCandidate tables filter out suspended candidates by default


tristan-orourke:
## ✨ Feature

Add the suspendedAt column to PoolCandidate tables on the admin portal. Should be hidden by default.

PoolCandidate tables on the AdminDashboard filter can filter out candidates based on the suspendedAt status. By default, suspended candidates _should_ be filtered out, but this can be changed in the FancyFilter dialog.

## 🛑 Blocked By
```[tasklist]
- [ ] https://github.com/GCTC-NTGC/gc-digital-talent/issues/5732
- [ ] https://github.com/GCTC-NTGC/gc-digital-talent/issues/5732
```

## 🎨 Design File

Include a link to the design file (if it exists).

## 📸 Screenshot

Add new filter dropdown somewhere in here:
![image](https://user-images.githubusercontent.com/4313717/219489874-934611ef-4fbb-4ebc-8029-d30ce7d739d0.png)

## 🙋‍♀️ Proposed Implementation

This will involve an update to PoolCandidateSearchInput on the backend as well as changes to the frontend.

## 🌎 Localization

(optional) Provide any new copy along with translations available.

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the feature was properly implemented.

- [ ] By default, PoolCandidate tables on the admin portal exclude PoolCandidates where `suspendedAt !== null && suspendedAt < now()`
- [ ] In the Filter dialog, Admins can change the filter to include users who have suspended their application
- [ ] Admins can show a column which displays the suspendedAt status, and sort by that column
- [ ]

tristan-orourke:
We could consider different names for this. We could say `available=true` instead of `optOut=false`, for example.

PR #5918, opened 2023-03-10
Bug - 5913 - search page candidate counts

### LINKED ISSUE #5913: 🐛 Search page candidate counts mis-match
## 🐛 Bug

Search page issue again 🙃 

## 🦋 Expected Behaviour

Count per pool should sum to total

## 🕵️ Details

.

## 📋 Steps to Reproduce

1. head to /search, after building off main
2. start adding in filters
3. observe total candidate count and candidate count per pool card

## 📸 Screenshot

image

image

re-seeding

image

## 🙋‍♀️ Proposed Solution

countApplicants is higher in number due to not filtering out suspended candidates

Pool Candidate has a scope for suspended, which is then used in count by pool

image

User does not check for suspended in a scope, this would be done by count applicants

image

## ✅ Acceptance Criteria

- [ ] count per pool sums to total count
- [ ] queries return correct users, as in scopes are correctly assigned

vd1992:
🤖 Resolves #5913 

## 👋 Introduction

Resolve the counts not making sense on the search page.

## 🕵️ Details

Needed to incorporate filtering out suspended applications in the count applicant query chain.
Introduced an enum for this, for reuse in issue #5734 

## 🧪 Testing

1. head to /search
2. play around with filters and watch candidate count per pool and total candidate count

🚨 Note 🚨 
users can have multiple pool candidates so total count isn't always the sum of pool counts
but when you've only one pool left, the numbers should match

## 📸 Screenshot

image


api/graphql/schema.graphql
```diff
@@ -796,10 +796,17 @@ enum CandidateExpiryFilter {
...
+  ACTIVE
+  SUSPENDED
+  ALL
+}
+
 input UserPoolFilterInput {
   poolId: ID!
   expiryStatus: CandidateExpiryFilter = ACTIVE
   statuses: [PoolCandidateStatus!]
+  suspendedStatus: CandidateSuspendedFilter = ALL
```
vd1992, line 809: I was a bit iffy on this bit, wasn't sure if it was needed to match this up with the possible fields in `scopePoolFilters`

yonikid15 (commented):
Alright nice work Vachan! Took a while to review since the Search Page has so many parts to it. After testing it all out, it looks good to me, the candidate count is showing up as expected. However, regarding `CandidateSuspendedFilter` enum it looks good to me, but it's not part of the AC so I'll let @tristan-orourke approve that.

One issue I found that might be out of scope for this is that after I made a request. The number of estimated users never matched the number of candidates on the "View Request" page admin side. The request table always has more than the requested candidates. I'm not sure exactly why.. yet.

vd1992:
> Note to self - #5929 removes scopeAvailableForOpportunities, how do we ensure that doesn't conflict with this PR?

I don't think it should, the filtering happens in `scopePoolFilters` which is either called directly or by `scopeAvailableInPools`. This PR does not build upon `scopeAvailableForOpportunities`.

api/graphql/schema.graphql
tristan-orourke, line 809: I think this should be set to ACTIVE by default. We won't usually care about suspended candidates.

tristan-orourke (changes requested):
Hey, great job adding the suspended_at enums! Despite not being specified in the issue, this matches what we've previously done. I'd just change the default filter to ACTIVE.

api/graphql/schema.graphql
vd1992: oops, you are right https://github.com/GCTC-NTGC/gc-digital-talent/pull/5918/commits/fefb4a73624a021b24fce5f071c738d754e3107a

api/app/Models/User.php
```diff
@@ -260,6 +261,16 @@ public static function scopePoolFilters(Builder $query, ?array $poolFilters): Bu
                             if (array_key_exists('statuses', $filter) && !empty($filter['statuses'])) {
                                 $query->whereIn('pool_candidates.pool_candidate_status', $filter['statuses']);
                             }
+                            $query->where(function ($query) use ($filter) {
+                                if (array_key_exists('suspendedStatus', $filter) && $filter['suspendedStatus'] == ApiEnums::CANDIDATE_SUSPENDED_FILTER_ACTIVE) {
+                                    $query->where(function ($query) {
```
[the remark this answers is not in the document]
vd1992: think it is https://github.com/GCTC-NTGC/gc-digital-talent/pull/5918/commits/88bc4a68a7b507cf3fb52b5dca6e2742db536369

tristan-orourke (approved):

PR #5935, opened 2023-03-14
Feature - 5734 - pool candidate table suspended filter

### LINKED ISSUE #5734: ✨ PoolCandidate tables filter out suspended candidates by default

vd1992:
🤖 Resolves #5734 

## 👋 Introduction

Add filtering of candidates by whether they are set to suspended or not to the pool candidates table.

## 🕵️ Details

Just mimicking the structure of how the expiry filtering was added. 

## 🧪 Testing

1. head to the pool candidates page for a pool
2. play around with the new filter
3. observe queries in the network tab

## 📸 Screenshot

image


vd1992:
@Jerryescandon 

does the copy/wording for the new filter look fine to you? and not confusing?

image

api/app/Models/PoolCandidate.php
```diff
@@ -261,6 +261,20 @@ public static function scopeExpiryStatus(Builder $query, ?string $expiryStatus)
         return $query;
     }
 
+    public static function scopeSuspendedStatus(Builder $query, ?string $suspendedStatus)
```
petertgiles, line 264: I'm a bit uncomfortable with how this overlaps with `scopeAvailable` but I'm not sure if that can be avoided. :shrug:

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
```diff
@@ -52,6 +53,7 @@ import {
 import useTableState from "~/components/Table/ApiManagedTable/useTableState";
 import {
   stringToEnumCandidateExpiry,
+  stringToEnumCandidateSuspended,
```
petertgiles, line 56: I think this column needs to be added to the table, too, right? As per the AC:

>  Admins can show a column which displays the suspendedAt status, and sort by that column

petertgiles (changes requested):
Looking great!  Seems to work as intended.  I think missing one AC though.

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
vd1992: oh you're right, while I am at it, I might as well replace Availability with this while I am at it?
one less place to have to come back to later

image

petertgiles: If it's not too messy a change, sure.

api/app/Models/PoolCandidate.php
vd1992: 🤷

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
vd1992: okay, done in like, the last 6 commits 😃 

sorting is odd because the selected sort is the tertiary level one, status/priority are first and locked
since this is the API driven table, I had to sort by suspended_at values, which should be fine for our purposes

petertgiles (approved):
Very nice!  :smile:  :+1:

PR #5978, opened 2023-03-16
Explore routing simplification with subdomains

petertgiles:
>__Warning__
> This is an exploratory branch for review and evaluation but is incomplete and should not be merged to main.

## 👋 Introduction

This branch explorers how our server-side routing could be simplified and made more manageable with subdomains.

## 🕵️ Details

We have discussed as a team the possibility of using subdomains with our app to simplify our routing but there was doubt about how that might work or look.  This branch demonstrates how our one large config file could be broken down into three much smaller files. It sets up the _api_ and _report_ subdomains for our api and the tc-report.

### Status
- Main site :heavy_check_mark: 
- Graphiql :heavy_check_mark: 
- Login with mock auth :heavy_check_mark: 
- Login with SiC :x: Would need to whitelist the new URL
- Refresh and logout :question: not tested
- tc-report
   - The HTML pages work, but we removed the landing page so you'll need to navigate to a specific page like http://report.localhost:8000/en/talent-cloud/report/
   - The asset paths are broken.  We'd need to recompile the site, I think.
- Deploy conf file :x:    

## 🧪 Testing

1. Manually symlink the new config files for nginx inside the webserver container.  (Or just copy them to sites-enabled.)
image
2. Copy new .env files (or rerun setup)
3. Rebuild the app 

## 📸 Screenshot

image

### DIFF
--- a/api/config/cors.php
+++ b/api/config/cors.php
@@ -15,7 +15,7 @@ return [
     |
     */
 
-    'paths' => ['api/*', 'sanctum/csrf-cookie'],
+    'paths' => ['*'],
 
     'allowed_methods' => ['*'],
 
--- a/packages/auth/src/hooks/useApiRoutes.ts
+++ b/packages/auth/src/hooks/useApiRoutes.ts
@@ -5,7 +5,7 @@ export interface ApiRoutes {
   refreshAccessToken: () => string;
 }
 
-const apiRoot = (): string => "/";
+const apiRoot = (): string => "http://api.localhost:8000/"; // this should be an environment variable
 
 const apiRoutes = {
   login: (from?: string, locale?: string): string => {
@@ -15,7 +15,7 @@ const apiRoutes = {
     const searchString = searchTerms.join("&");
 
     const url =
-      path.join(apiRoot(), "login") + (searchString ? `?${searchString}` : "");
+      new URL("/login", apiRoot()) + (searchString ? `?${searchString}` : "");
 
     return url;
   },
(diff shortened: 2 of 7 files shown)


api/config/cors.php
```diff
@@ -15,7 +15,7 @@
     |
     */
 
-    'paths' => ['api/*', 'sanctum/csrf-cookie'],
+    'paths' => ['*'],
```
petertgiles, line 18: Oops, CORS has been broken on our API.  Do we want to restrict it or is a wildcard OK?

packages/auth/src/hooks/useApiRoutes.ts
```diff
@@ -5,7 +5,7 @@ export interface ApiRoutes {
   refreshAccessToken: () => string;
 }
 
-const apiRoot = (): string => "/";
+const apiRoot = (): string => "http://api.localhost:8000/"; // this should be an environment variable
```
petertgiles, line 8: This should be an environment variable.

petertgiles (commented):
Few comments.

tristan-orourke:
Okay, I admit, I did not realize how much easier to maintain this would be until seeing it like this 🤯 .

tristan-orourke:
This kind of division would also make it a lot easier to put the client code in something like a Next.js server one day...

petertgiles:
A side benefit is that it would open up /login and /logout for use by the frontend app.

PR #6001, opened 2023-03-17
Bug 5993 experience application return path

### LINKED ISSUE #5993: Add an experience drops out of the application
## 🐛 Bug

If you add an experience to your profile as part of an application, clicking the "save and go back" button returns to the profile, not the application.

## 🦋 Expected Behaviour

"Save and go back" should return back to the application in progress.

## 🕵️ Details

- **Browser**: Firefox 111.0

## 📋 Steps to Reproduce

1. Log in as an application
2. Start and application
3. Add an experience as part of the application and fill out the page
4. Click "save and go back" on the new experience - observe you are in your profile, not your application

## 📸 Screenshot
image
image

## 🙋‍♀️ Proposed Solution

It looks like maybe the query at the end of the URL got mangled.

## ✅ Acceptance Criteria

- [ ] "Save and go back" should return back to the application in progress.
- [ ] Fix the "cancel" button, too.
- [ ] Fix the breadcrumbs, too.

vd1992:
🤖 Resolves #5993 

## 👋 Introduction

Enable returning to the correct place after updating an experience in the application workflow. 

## 🕵️ Details

URL was slightly mangled.
`returnPath` was defined in two places, elected to define it in just the container and pass it in as props to the form component. 

## 🧪 Testing

1. apply to an opportunity 
2. update various experiences and create them, ensure you stay in the application workflow
3. links work as expected

### DIFF
--- a/apps/web/src/pages/Profile/ExperienceFormPage/ExperienceForm.stories.tsx
+++ b/apps/web/src/pages/Profile/ExperienceFormPage/ExperienceForm.stories.tsx
@@ -45,6 +45,7 @@ const TemplateExperienceFormForm: Story<ExperienceFormStoryArgs> = ({
     deleteExperience={() => {
       /* do nothing */
     }}
+    returnPath=""
   />
 );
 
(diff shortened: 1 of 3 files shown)


vd1992:
one thing to note, clicking the final crumb for the breadcrumbs, creates this oddity that knocks you out of the application
https://recordit.co/VLVQG1xCNr
but this occurs for other forms too, like About Me in the example recording, so not limited to the experience page

I don't think the final crumb should be clickable/interactive as it points at the current page

esizer (approved):
LGTM! 👍

petertgiles (approved):
Nice one!

Issue #5095, opened 2022-12-12
Upgrade to Cypress 12


petertgiles:
## Migration Guide
- [Migrating to Cypress 12.0](https://docs.cypress.io/guides/references/migration-guide#Migrating-to-Cypress-12-0)

## Acceptance Criteria
- [ ] Cypress 12 used locally
- [ ] Cypress 12 used in CI
- [ ] Tests behave as usual
- [ ] Educate your teammates on how to use the new version

petertgiles:
What can we run in Github?

PR #6002, opened 2023-03-17
Dependencies Cypress v11 update

vd1992:
🤖 Progresses #5095 

## 👋 Introduction

Trying Cypress updates, followed the guide https://docs.cypress.io/guides/references/migration-guide
Updates to latest in v11 

## 🕵️ Details

Not much in the way of conflicts, some stuff removed/to be removed but not things we were using. Needed to update auth spec as `within()` can only have one element passed into it now.
There are new things to play around with like component testing. 

Can't get things to run on github with v12, shooting for v11 for now. 

## 🧪 Testing

1. cypress passes repeatedly on github
2. cypress passes repeatedly locally

### DIFF
--- a/apps/e2e/cypress/e2e/admin/auth.cy.js
+++ b/apps/e2e/cypress/e2e/admin/auth.cy.js
@@ -1,202 +1,203 @@
 import { aliasQuery } from "../../support/graphql-test-utils";
 
-const AUTH_SERVER_ROOT = 'http://localhost:8000/oxauth';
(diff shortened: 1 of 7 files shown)


apps/e2e/cypress/e2e/admin/auth.cy.js
mnigh, line 1: @vd1992 any particular reason why the single quotes needed to be changed to double quotes?

vd1992: auto-formatting 🤷‍♂️

PR #6006, opened 2023-03-20
[Bug] Fix applicant pool candidates query

### LINKED ISSUE #6005: 🐛 Pool Operators can't view pool candidates profile when the pool candidate has any draft application
## 🐛 Bug

Pool Operators can't view pool candidates profile when the pool candidate has any draft application 

## 🦋 Expected Behaviour

Pool operators should be able to view the applicant profile regardless of whether they have draft application or not. 

## 🕵️ Details

**Error Received :** 
```
graphQLErrors":[
         {
            "message":"This action is unauthorized.",
            "path":[
               "applicant",
               "poolCandidates"
            ],
            "extensions":{
               "category":"authorization"
            }
         }
      ],
```
Query that fails : `GetViewUserData` 

**Root cause** 
public function view(User $user, PoolCandidate $poolCandidate)
    {

        // If the user owns the application, we do not care about status
        if ($user->id === $poolCandidate->user_id && $user?->isAbleTo("view-own-application")) {
            return true;
        }

        $isDraft = $poolCandidate->isDraft();

        // Exit early if user can view any draft application
        if ($isDraft) {
            return $user->isAbleTo("view-any-draftApplication");
        }

        if(!$isDraft) {
            if($user->isAbleTo("view-any-submittedApplication")) {
                return true;
            }

            $poolCandidate->loadMissing('pool.team');
            $candidatePoolTeam = $poolCandidate->pool->team;
            if ($user->isAbleTo("view-team-submittedApplication", $candidatePoolTeam)) {
                return true;
            }
        }

        // Noting passed for deny access
        return false;
    }
## 📋 Steps to Reproduce

1. Login as Applicant
2. Create 2 applications, Leave 1 as draft and submit the other application for a job poster 
3. Login as Pool Operator of that team that job poster belongs to 
4. Try to view the applicant profile 

## 📸 Screenshot

## 🙋‍♀️ Proposed Solution

add notDraft scope to Applicant.poolCandidates 

## ✅ Acceptance Criteria

- [ ] Pool operators should be able to view the applicant profile regardless of whether they have draft application or not. 
- [ ] Take a closer look at any queries which return or types which include PoolCandidates, and decide if they also need additional scopes.

## 🛑 Blockers

Blocked by {issue numbers}

esizer:
🤖 Resolves #6005 

## 👋 Introduction

Adds the `notDraft` scope to `Query.applicant.poolCandidates`.

## 🕵️ Details

No one but the owner of a draft may view draft pool candidates. This was causing a problem with `Query.applicant.poolCandidates` since it was exposing draft applications. This adds the scope to filter out draft applications.

## 🧪 Testing

1. Login as applicant
2. Submit an application and start another on as a draft
3. Login as an admin
4. View the user information page for the applicant user
5. Confirm there are no GraphQL errors
6. Confirm you only see the submitted application

## 📸 Screenshot

Screenshot 2023-03-20 113822

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -273,7 +273,7 @@ type Applicant {
   jobLookingStatus: JobLookingStatus
     @rename(attribute: "job_looking_status")
     @deprecated(reason: "Removing with applicantDashboard feature flag")
-  poolCandidates: [PoolCandidate] @hasMany @can(ability: "view", resolved: true)
+  poolCandidates: [PoolCandidate] @hasMany(scopes: ["notDraft"]) @can(ability: "view", resolved: true)
 
   hasDiploma: Boolean @rename(attribute: "has_diploma")
   locationPreferences: [WorkRegion] @rename(attribute: "location_preferences")


petertgiles (approved):
Yeah, this works!

I see two other relations to pool candidates:
- Pool -> poolCandidates
- User -> poolCandidates
Do we need the fix there?

I guess the first is OK since the relation is publishedPoolCandidates which uses the notDraft scope explicitly.

The second does have that problem (!) but as long as the admin site doesn't use it it should be OK.  But if we scope it I guess the user will never be able to list their own draft applications.

esizer:
> The second does have that problem (!) but as long as the admin site doesn't use it it should be OK. But if we scope it I guess the user will never be able to list their own draft applications.

Yeah, the reason I left it out here is because we use `User.poolCandidates` to show draft applications to users so it would break the page that lists user applications and the new applicant dashboard since `Query.me` returns a `User` and should be  covered by `@can(ability: "view", resolved: true)`.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/0e74d0486deb8f97c0913ee4405871d664e96fb3/api/graphql/schema.graphql#L879

https://github.com/GCTC-NTGC/gc-digital-talent/blob/0e74d0486deb8f97c0913ee4405871d664e96fb3/api/graphql/schema.graphql#L116-L117

tristan-orourke:
You guys are too fast! 😆 I come back from lunch to review this and its already in!

I was going to suggest adding the notDraft scope to countPoolCandidates and searchPoolCandidates, but countPoolCandidates is deprecated and I think searchPoolCandidates SHOULD be deprecated, so its not a huge deal.

PR #6009, opened 2023-03-20
[Bug] Fix application profile breadcrumbs

### LINKED ISSUE #6004: 🐛 Add/Edit Experience Breadcrumb Link invalid
## 🐛 Bug

We are passing an empty has to the final breadcrumb in the add/edit experience page, making it invalid.

## 🦋 Expected Behaviour

The final link in the breadcrumb should have a valid URL and be marked with `aria-current="page"`.

## 🕵️ Details

https://github.com/GCTC-NTGC/gc-digital-talent/pull/6001#issuecomment-1476185739

https://github.com/GCTC-NTGC/gc-digital-talent/blob/601529adc62dac6d3ccd629beb7d802eb126a15a/apps/web/src/pages/Profile/ExperienceFormPage/ExperienceFormPage.tsx#L103-L116

## 📋 Steps to Reproduce

1. Apply to a pool advertisement
2. Edit your experiences within the application context
3. Add or Edit an experience within the application context
4. Click the last breadcrumb
5. Confirm you are taken out of the application context

## 📸 Screenshot

## 🙋‍♀️ Proposed Solution

Pass a valid url to the breadcrumbs.

## ✅ Acceptance Criteria

- [ ] The edit/add experience page breadcrumbs have a valid URL
- [ ] Clicking a breadcrumb on the add/edit experience pages keeps you in the current context (application/profile).

esizer:
🤖 Resolves #6004 

## 👋 Introduction

Fixes URLs for current page breadcrumb links within the application context.

## 🕵️ Details

Adds the missing search param name as well as removes the trailing slash for edit links.

> **Note**
> Removing the trailing slash was not part of the acceptance criteria. However, we do not currently use trailing slashes anywhere else and, it was causing a navigation event where the back button would just add the trailing slash back.

## 🧪 Testing

1. Apply to a pool advertisement
2. Click the edit link for each section
3. Confirm trailing slash is gone
4. Confirm the current breadcrumb does not take you out of the application context

### DIFF
--- a/apps/web/src/pages/Profile/ExperienceFormPage/ExperienceForm.test.tsx
+++ b/apps/web/src/pages/Profile/ExperienceFormPage/ExperienceForm.test.tsx
@@ -12,7 +12,6 @@ import type { ExperienceQueryData, ExperienceType } from "./types";
 const mockUserId = "user-id";
 const mockSkills = fakeSkills(50);
 const mockExperiences = fakeExperiences(5);
-const returnPath = "";
 const mockCallback = jest.fn();
 
 const renderExperienceForm = (props: ExperienceFormProps) =>


apps/web/src/hooks/useRoutes.ts
```diff
@@ -9,33 +9,31 @@ const getRoutes = (lang: Locales) => {
   const adminUrl = path.join(baseUrl, "admin");
   const userUrl = (userId: string) => path.join(baseUrl, "users", userId);
   const applicationParam = (applicationId?: string) =>
-    applicationId ? path.join(`?applicationId=${applicationId}`) : "";
+    applicationId ? `?applicationId=${applicationId}` : "";
```
petertgiles, line 12: Could you check the Edit Skills and Experience link?  Looks like we got an extra slash.
image

petertgiles (changes requested):
Great tidying!  I think I found one missed spot for the links but the breadcrumbs looks great.

apps/web/src/hooks/useRoutes.ts
esizer: Good catch! Fixed in [remove trailing slash from edit skills link](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6009/commits/b224e58c4b100b86c731bcdc2cc4993313188be1)

Issue #5933, opened 2023-03-14
🐛 Request page - Empty requests should enforce required fields


Jerryescandon:
## 🐛 Bug

DCM has been receiving requests with multiple pools indicated. This should not be possible. When submitting this test request, I only selected group and level, but not a stream, even though it was marked as required, it wasnt enforced for empty requests.

Additionally: When reviewing the request on admin side, ONE applicant showed up. Which means something is off in the results. If there was at least one candidate, it should not have given me an empty request option. 🤔 (see 2nd screenshot) maybe a different issue.

## 🦋 Expected Behaviour

 Even when an empty request is submitted, we should enforce the Classification + level + Stream selection. This would vastly reduced the pools that show up on the request page

## 🕵️ Details

Add any additional details that could assist with troubleshooting/fixing the issue.

- **Operating System**: Name and version (if possible).
- **Browser**: Name and version (if possible).

## 📋 Steps to Reproduce

1. ...
2. ...

## 📸 Screenshot

![image](https://user-images.githubusercontent.com/16977254/225115090-7644bd31-a4a2-4fbe-98b8-1bad04e8e3c7.png)

![image](https://user-images.githubusercontent.com/16977254/225115381-ff949a6b-33c1-494c-a570-556e4e339d48.png)

## 🙋‍♀️ Proposed Solution

After #5644 is completed, we'll be able to associate the stream and classification to the request properly. At that point, we can simply avoid adding any pools to empty requests.

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the bug was addressed.

- [ ] Criteria 1
- [ ] Criteria 2

## 🛑 Blockers

Blocked by #5644

vd1992:
for the empty request part of the issue, I feel this can happen downstream of #5644 
the completion of that issue may solve this as a side-effect, or at least make this issue doable

esizer:
> for the empty request part of the issue, I feel this can happen downstream of https://github.com/GCTC-NTGC/gc-digital-talent/issues/5644
the completion of that issue may solve this as a side-effect, or at least make this issue doable

> so blocked by https://github.com/GCTC-NTGC/gc-digital-talent/issues/5644 would be my say

You are correct @vd1992 . The reason more than on pool is attached to a request here is because we do not associate classification or job stream with search requests. So, when an empty request gets submitted we attach all pools and classifications available with the information we do know.

We added #5644 so we can eventually do away with the work around noted previously 👍

Issue #5644, opened 2023-02-07
Associate search request directly with classification and job stream


petertgiles:
## ✨ Feature

Currently we associate search requests with a pool and infer the desired classification and job stream from that pool.  In the near future search requests will no longer be associated with a pool.  Therefore the classification and job stream of the request will need to be associated directly with the search request.

## 🕵️ Details

Add any additional details that could assist with the development of the new feature.

## 🎨 Design File

Include a link to the design file (if it exists).

## 📸 Screenshot

Add a screenshot of the new feature (if possible).

## 🙋‍♀️ Proposed Implementation

Add Classification and Stream to ApplicantFilter and CreateApplicantFilterInput as optional fields. Ensure they correctly filter query results.

## 🌎 Localization

(optional) Provide any new copy along with translations available.

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the feature was properly implemented.

- [x] When new search requests are submitted, they include Classification and Stream attached directly, _along with_ the pool
- [x] If classification and stream are defined, then show them in the SearchRequestFilters component
- [x] If applicantFIlter.pools is not empty, show the name of each in SearchRequestFilters, instead of just displaying pool classifications.
- [x] When an empty search request is submitted, don't include any Pools, but _do_ include Classification and Stream
- [x] On the Admin side, on PoolCandidate tables, allow for filtering by Classification and Stream. Ensure these filters are initialized correctly by the search request.

vd1992:
the means to connect classification to applicant filters exists, the pivot table exists, it just needs a frontend tweak to populate correctly (currently submitting a request doesn't attach the classification properly)

stream requires work, streams aren't objects but fields on pools, therefore may need a migration to add a stream field onto applicant filter model

![image](https://user-images.githubusercontent.com/40485260/224189976-573d8a3d-aca5-489c-8edc-b91040d8f1b4.png)


![image](https://user-images.githubusercontent.com/40485260/224189739-6f89cae5-a539-4bd7-aaa2-f911341d72e4.png)

brindasasi:
moving this back to in progress as we intend to fix the poolcandidate filter also in this PR.

Issue #5347, opened 2023-01-16
✨ Remove Roles and Salaries section of profile


tristan-orourke:
## ✨ Feature

Remove the Roles and Salaries section of the profile (again).

## 🕵️ Details

We did this once, and had to revert it. But at some point we'll be ready to do this again.

## 🙋‍♀️ Proposed Implementation

Revert `Revert "Debt - 5133 - Remove roles salaries"` #5343

## ✅ Acceptance Criteria

- [ ] Roles and Salaries section is gone from the Profile
- [ ] Roles and Salaries section is gone from the profile when viewed by Admins as well
- [ ] expectedClassifications, expectedSalary, and expectedGenericJobTitle fields are deprecated in types and removed from inputs in schema
- [ ] expectedClassifications, expectedSalary, and expectedGenericJobTitle fields are are removed from tests, seeders, and graphql operations
- [ ] expectedClassifications, expectedSalary, and expectedGenericJobTitle fields are removed from applicant snapshotting

tristan-orourke:
We'll remove data from the database later. For now let's just deprecate the fields in the schema.

mnigh:
@tristan-orourke do we have the go-ahead to do it again? who's the go-ahead giver? 😄

tristan-orourke:
@gobyrne is the go-ahead giver. What do you think, @gobyrne?

tristan-orourke:
Alright @gobyrne isn't around, lets just do it. I'm pretty sure no one will miss it.

Issue #5133, opened 2022-12-14
Hide, remove, deprecate, delete, destroy the "Role and salary expectations" section of the profile


gobyrne:
The roles and salary expectations page is not currently in use and even the intended future use will now be replaced by the "Ongoing Recruitment" feature. So now we can safely remove it.

At a minimum we could simply hide the section on the profile and remove associated validation for it. We also need to remove/hide the corresponding section on the admin view of the profile and the "print profile", and the CSV user & candidate downloads.

At a maximum we could seek out and destroy all references to this in the code.

### Acceptance Criteria
- [ ] Delete role and salary components / pages
- [ ] Delete routes
- [ ] Fix the completion check
- [ ] Deprecate field in api

vd1992:
@tristan-orourke 

this form connects to the `expectedGenericJobTitles` field

another thing to note is that there is also another related field that seems to have been redundant for a while, `expectedClassifications`
it does not appear to be defined anywhere from an applicant's perspective and should also be deprecated at some point?

as well, `expectedSalary` is the same, not controllable by the user anywhere

with the above three fields no longer useful, `scopeClassifications` is also no longer usable as there is nothing to return
this makes all filtering by classification then defunct, which is a filter present on the Users and the Pool Candidates table filter dialogs

so, it would be a good idea to clear up the state of the classifications/genericJobTitles/expectedSalary fields and which are for sure deprecated

tristan-orourke:
Nice catch @vd1992. Sounds like all of these fields should be deprecated.

esizer:
I may be missing something here, but it seems as though we are using this to filter users on the search page 🤔 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/6d1577d25eddee17812d21c2610c190f903ff2e2/api/app/GraphQL/Queries/CountPoolCandidatesByPool.php#L77-L80

https://github.com/GCTC-NTGC/gc-digital-talent/blob/6d1577d25eddee17812d21c2610c190f903ff2e2/api/app/Models/User.php#L413-L543

tristan-orourke:
@esizer @vd1992 References to `expectedClassifications` still exist in the code, but I think Vachan's right that they're not actually being _used_. For example, in line 160 of SearchForm.tsx, where the search page builds a filter object to send to the api, expectedClassifications is always set to the empty list. 

And in places where we might actually be using it, like on some Admin table filters, its not useful, because candidates have no way to add that info to their profile now.

[...]

PR #6010, opened 2023-03-20
[Feature] add classification and stream into search requests

brindasasi:
🤖 Resolves {issue #5644 }

## 👋 Introduction
This PR adds Classification & Stream into the search request. 
It doesn't change how existing search functionality works with pool id. 
When no candidate exist these classification & stream will still be stored with the search request. 
It fixes the pool candidate filter on the view request page 
It resolves  https://github.com/GCTC-NTGC/gc-digital-talent/issues/5933 ,https://github.com/GCTC-NTGC/gc-digital-talent/issues/5933 as well

## 🕵️ Details

This PR 

-  adds qualifiedClassifications, qualifiedStreams into ApplicantFilter ( user can filter by multiples of them in pool candidate filter; hence plural naming) 
-  search page and admin side view request page now refers to `qualifiedClassifications` instead of 
  `expectedClassifications`
- ApplicantFilter model still keeps the  `expectedClassifications` and passing an empty array on search and request pages. 
- leaves expectedClassifications as is and wherever it has been referred as `Classifications` has been changed to `expectedClassifications` to avoid confusion
- scopeClassifications got renamed to scopeExpectedClassifications
- added new scopes scopeQualifiedClassifications, scopeQualifiedStreams in order to use in the pool candidate table filtering

## 🧪 Testing

**Search Page**
1. Navigate to Search page, Fill up the options so that atleast 1 candidate exist 
2. Proceed to next page to submit request 
3. Verify Classification & Stream exist 

1. Navigate to Search page, Fill up the options so that 0 candidate exist 
2. Proceed to next page to submit empty request 
3. Verify Classification & Stream is sent along the request 

1. Navigate to Search page , Fill up Classification , Location Preference and leave Stream as empty
2. Verify validation error appears for Stream ( part of #5933) 

**View Requests Page**
1. Go to Requests page 
2. Verify the request you submitted with candidates exists
3. Click on the View Request
5. Verify Classification & Stream show up correctly on the page 
6. Verify the no. of pool candidates shown in the table match the no that was shown in the Search page

## 📸 Screenshot

Screenshot 2023-03-27 at 11 00 08 AM

Screenshot 2023-03-27 at 10 57 32 AM

** Empty Request **
1. Go to Requests page 
2. Verify the request you submitted with zero candidates exists
3. Click on the View Request
5. Verify Classification & Stream show up correctly on the page 
6. Verify the pool name is empty 
7. Verify the pool candidate table is empty


api/app/Models/ApplicantFilter.php
```diff
@@ -45,6 +46,10 @@ public function classifications(): BelongsToMany
     {
         return $this->belongsToMany(Classification::class, 'applicant_filter_classification');
     }
+    public function streams(): BelongsToMany
+    {
+        return $this->belongsToMany(PoolStream::class, 'applicant_filter_stream');
+    }
```
vd1992, line 52: This is just a new field on a model and not a relation between tables so this would be wrong I think

apps/web/src/components/SearchRequestFilters/SearchRequestFilters.tsx
```diff
@@ -273,6 +282,14 @@ const ApplicantFilters = ({
                 classificationsFromApplicantFilter,
             )}
           />
+          <FilterBlock
+            title={intl.formatMessage({
+              defaultMessage: "Stream",
+              id: "Ua/X9Q",
+              description: "Title for stream on summary of filters section",
+            })}
+            content={streamFromApplicationFilter || streamFromPool}
```
vd1992, line 291: should be localized

apps/web/src/components/SearchRequestFilters/SearchRequestFilters.tsx
```diff
@@ -503,6 +525,14 @@ const SearchRequestFilters = ({
               })}
               content={classifications}
             />
+            <FilterBlock
+              title={intl.formatMessage({
+                defaultMessage: "Stream",
+                id: "Ua/X9Q",
+                description: "Title for stream on summary of filters section",
+              })}
+              content={streams}
```
vd1992, line 534: should be localized

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
```diff
@@ -77,6 +77,9 @@ function transformPoolCandidateSearchInputToFormValues(
       input?.applicantFilter?.expectedClassifications
         ?.filter(notEmpty)
         .map((c) => `${c.group}-${c.level}`) ?? [],
+    stream: input?.applicantFilter?.stream
+      ? [input?.applicantFilter?.stream]
+      : [],
```
vd1992, line 82: the stream from the request does not get loaded in the filters dialog initial values, it should
nor does it seem to do any filtering (no API call is made)

https://recordit.co/oEiD9kzSdA

vd1992 (changes requested):

apps/web/src/components/SearchRequestFilters/SearchRequestFilters.tsx
vd1992: image

api/app/Models/ApplicantFilter.php
brindasasi: oh yes.. will remove that

brindasasi:
> it appears to sync classifications correctly now! which should make #5933 probably straightforward? the empty requests will be made to look better in that issue

Yes.. its  quite straight forward now to remove attaching all pool ids when there is empty request. I planned to do that right after this.

apps/web/src/components/Table/ApiManagedTable/useFilterOptions.tsx
```diff
@@ -33,6 +34,7 @@ import {
   CandidateExpiryFilter,
   CandidateSuspendedFilter,
 } from "~/api/generated";
+import { PoolStream } from "@gc-digital-talent/graphql";
```
petertgiles, line 37: This should probably be imported with all the other models above.  Either move them to this import or move this model to that import.

petertgiles (commented):
Looking cool!  I did a quick skim but I'll take a closer look once you've had a chance to work through Vachan's review.

apps/web/src/components/SearchRequestFilters/SearchRequestFilters.tsx
brindasasi: oo.. I missed this one.. fixing now

petertgiles:
Task: double-check that estimates match request table

api/app/Models/PoolCandidateFilter.php
```diff
@@ -57,6 +57,10 @@ public function poolCandidateSearchRequest(): HasOne
     {
         return $this->hasOne(PoolCandidateSearchRequest::class);
     }
+    public function stream()
+    {
+        return $this->stream;
+    }
```
vd1992, line 63: what is the purpose of this given pool_candidate_filters is obsolete and not being updated here?

api/graphql/schema.graphql
```diff
@@ -634,6 +635,9 @@ type ApplicantFilter {
   expectedClassifications: [Classification]
     @belongsToMany(relation: "classifications")
   skills: [Skill] @belongsToMany
+  classifications: [Classification]
+    @belongsToMany(relation: "classifications")
```
vd1992, line 639: type ApplicantFilter now has two classifications fields connected to it, this is kind of confusing, are they addressing the same thing because if so there should only be one

https://github.com/GCTC-NTGC/gc-digital-talent/blob/9990f8b224afba81f60e8008dde9ed86d80b92dc/api/graphql/schema.graphql#L635-L639

vd1992 (commented):
couple further questions, prior feedback appears addressed

api/graphql/schema.graphql
brindasasi: Agree its very confusing. Both are 2 different fields and both are live in production. However, some of the places we inject `classifications` into `expectedClassifications` and rename it back into `classifications` while storing into db. 
I started cleaning up and left with the bundle of tangled threads. So thought to stay away from that in this PR.

vd1992: that is a bit unfortunate
I supposed please include a comment then that quickly explains what each classification field does? to make it easier to understand what is going on in the schema 

"some of the places we inject classifications into expectedClassifications and rename it back into classifications"
😭

vd1992:
I think things seem to work, my feedback appears addressed, handing things over to @petertgiles now
One thing is, empty requests are rather funky now, but that is already a separate issue and ought to be pulled in ASAP after this is done. 

> Task: double-check that estimates match request table

This appears resolved on my end

apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
petertgiles, line 1: I don't think the empty request AC is being satisfied.
> When an empty search request is submitted, don't include any Pools, but do include Classification and Stream

Here's an empty request I submitted:
image

petertgiles (changes requested):
Oof, this is a tricky one.  I've highlighted a few things below.

api/graphql/schema.graphql
brindasasi: https://github.com/GCTC-NTGC/gc-digital-talent/issues/5133

brindasasi: Expected Classification was originally used meant to capture this classification. But it has been changed as above now
Screenshot 2023-04-04 at 10 53 32 AM

apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
petertgiles: The trouble I found is that admin page _does_ use them, and then you get different results on that page.

api/graphql/schema.graphql
petertgiles: Yeah, it's a mess. :cry: Whatever you end up doing, please add some good comments to the schema file to make it clear what's going on.

apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
brindasasi: I removed the pool ids from empty request. Unfortunately it doesn't seem to solve the problem of showing candidates in the pool table. I submitted an empty request for IT-04 Software Solutions. When I navigate to request page , pool candidate shows 4 candidates for that particular request like below :

Screenshot 2023-04-04 at 3 43 27 PM
But the query `GetPoolCandidatesPaginated` sends out the argument properly though
Screenshot 2023-04-04 at 3 43 18 PM

Resultant candidates are either:  **IT-04 OR Bilingual OR Woman** 🙈

tristan-orourke: I don't think there's anything wrong with the results of that query 🤔

brindasasi: but search page does **IT-04 AND SOFTWARE_SOLUTIONS AND BILINGUAL AND WOMAN** 
There is no such pool candidate available hence I submitted empty request.
When I open the request , pool candidate table shows like this with 4 candidates. 
Thats what the problem described in second paragraph of https://github.com/GCTC-NTGC/gc-digital-talent/issues/5933

tristan-orourke: Ah, oops, I see what you mean.

tristan-orourke: I think you're right, this is more in the scope of #5933.

api/app/Models/PoolCandidateFilter.php
patcon: Maybe if this is the expectation, we could use phpdoc to deprecate the class and explain that we're not expecting any more changes to them?

https://phpstan.org/writing-php-code/phpdocs-basics#deprecations

api/graphql/schema.graphql
patcon: Was wondering whether to spin out an issue for removing `expectedClassification` to remove opportunity for this confusion to surface again, and found this related issues: https://github.com/GCTC-NTGC/gc-digital-talent/issues/5347

brindasasi:
> Small wondering: Any support for renaming `expectedClassifications` to `classificationsExpected` to make these show up beside one another in lexicographical sorts, maybe allowing more coherent comments explaining what's up? Also, a little less visual noise around code working with them

these classifications are not in that context anymore. These classifications are just the user searched for in the search page. 
we are attaching it to the applicantfilter model in order to show that request.

api/graphql/schema.graphql
brindasasi: thats right Pat! I attempted to rename only related pages in this PR. The whole cleanup should be in separate PR as in some places we are using `expectedClassification`  name to store `GenericJobTitles` .. thats making it more worse.. 😩```
expectedClassification: getExpectedClassifications(
          expectedGenericJobTitles,
          intl,
        ),
```

**Comment by patcon** [comment]
Pls re-request when it's ready for a review :)

File: api/app/GraphQL/Queries/CountPoolCandidatesByPool.php [hunk 8]
```diff
@@ -73,7 +73,7 @@ public function __invoke($_, array $args)
 
             // expectedClassifications
             if (array_key_exists('expectedClassifications', $filters)) {
-                User::scopeClassifications($userQuery, $filters['expectedClassifications']);
+                User::scopeExpectedClassifications($userQuery, $filters['expectedClassifications']);
```
brindasasi, line 76: changed this for explicit naming

api/app/Models/PoolCandidate.php
```diff
@@ -84,7 +84,75 @@ public function screeningQuestionResponses(): HasMany
         return $this->hasMany(ScreeningQuestionResponse::class);
     }
 
-    public function scopeClassifications(Builder $query, ?array $classifications): Builder
+    public static function scopeQualifiedStreams(Builder $query, ?array $streams): Builder
+    {
+        if (empty($streams)) {
```
brindasasi, line 89: This and below are the actual meat that fixes the pool candidate filter.

apps/web/src/components/SearchRequestFilters/SearchRequestFilters.tsx
```diff
@@ -129,15 +130,7 @@ const ApplicantFilters = ({
...
-  const pools = applicantFilter?.pools?.filter(notEmpty);
-  const classifications: Classification[] =
-    pools?.reduce(
-      (previousValue: Classification[], currentValue: Maybe<Pool>) => {
-        const cls = currentValue?.classifications?.filter(notEmpty) || [];
-        return [...previousValue, ...cls];
-      },
-      [],
-    ) || [];
+  const classifications = applicantFilter?.qualifiedClassifications || [];
```
brindasasi, line 133: we shouldn't get the classification from pool any more.

brindasasi:
Failing test is on deprecated functionality .. I will discuss with @tristan-orourke on this .. but otherwise this should be ready to review. Tried some inline comments for PR for explaining stuff.

petertgiles (changes requested):

petertgiles (changes requested):
It seems to work!  I admit I don't feel I have a good handle on the search schema works at this point.  I think it may be worth having some new PHPUnit tests to prove that these changes work, since _search_ is such a hot part of the app.  I've left a few superficial suggestions below, as well.

petertgiles:
All the new comments are great!  :heart_eyes:

tristan-orourke (commented):
Looks good, except I think one of the acceptance criteria is incomplete:
> If applicantFIlter.pools is not empty, show the name of each in SearchRequestFilters, instead of just displaying pool classifications.

vd1992:
> I think it may be worth having some new PHPUnit tests to prove that these changes work, since _search_ is such a hot part of the app. I've left a few superficial suggestions below, as well.

Definitely good to be safe!

brindasasi:
> Looks good, except I think one of the acceptance criteria is incomplete:
> 
> > If applicantFIlter.pools is not empty, show the name of each in SearchRequestFilters, instead of just displaying pool classifications.

Pool name has been populated in the filter. Do you mean that ? 
Screenshot 2023-04-17 at 1 57 09 PM
Screenshot 2023-04-17 at 1 57 23 PM

brindasasi:
> > I think it may be worth having some new PHPUnit tests to prove that these changes work, since _search_ is such a hot part of the app. I've left a few superficial suggestions below, as well.
> 
> Definitely good to be safe!

agree. I will try my level best on php unit tests

tristan-orourke:
I think the Pool (if its selected) should appear in the Summary of Filters section, not the Manager Information section. And this way, it'll also appear then a user is making a request:

image

brindasasi:
> I think the Pool (if its selected) should appear in the Summary of Filters section, not the Manager Information section. And this way, it'll also appear then a user is making a request:
> 
> image

hmm are you sure? .. its not part of the filter and the heading says `Summary of Filters.` 
It is not hard to add it in there as the info is in there already.

tristan-orourke:
> hmm are you sure? .. its not part of the filter and the heading says Summary of Filters.

Doesn't pool id get submitted in the ApplicantFilter?
image

brindasasi:
> > hmm are you sure? .. its not part of the filter and the heading says Summary of Filters.
> 
> Doesn't pool id get submitted in the ApplicantFilter? image

Oh it is part of the data model. I meant its not part of the filter section on the UI where the user going thru bunch of dropdown filters. 
User doesn't filter by pool. But he selects it from the result section of those filters. 


> > hmm are you sure? .. its not part of the filter and the heading says Summary of Filters.
> 
> Doesn't pool id get submitted in the ApplicantFilter? image

It doesn't have the pool name for some reason..  :( .. will try to get it using that id from the list of pools  
Screenshot 2023-04-19 at 1 29 53 PM

tristan-orourke (approved):
Actually, never mind, my request can be spun off into its own issue, and I'll describe it better there.

Everything else looks good to me.

Issue #5214, opened 2022-12-22
♻️ Remove all uses of React.FC and React.FunctionComponent


tristan-orourke:
## ♻️ Debt/Refactor

As [we have discussed](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5194#issuecomment-1362992705) and we have seen [recommended elsewhere](https://github.com/facebook/create-react-app/pull/8177), use of the React.FC and React.FunctionComponent types has no actual benefit and slight potential downsides. We've tried to stop using them, but previous use remains everywhere in the code. 

Removing existing uses will ensure we don't accidentally promote further use via copying existing examples.

## 🕵️ Details

Add any additional details that could assist with the development of the new feature.

## 🙋‍♀️ Proposed Solution

As described [here](https://github.com/facebook/create-react-app/pull/8177), any references to React.FC/React.FunctionComponent can be deleted, the props Type/Interface can be attached directly to the function argument, and we can let Typescript infer the component function return type.

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the debt was addressed and expected functionality has not been affected.

- [ ] A global search of React.FC and React.FunctionComponent find zero results.

petertgiles:
We can break this up into groups of files if needed.

PR #6011, opened 2023-03-20
Remove FC from packages/forms

petertgiles:
🤖 Related to (but does not close) #5214 

## 👋 Introduction

This branch removes the use of React.FunctionComponent and React.FC from packages/forms.  Other subsequent PRs will remove it from other locations.

## 🧪 Testing

1. App builds
2. App runs OK (especially form components)
3. There are no more instances of FunctionComponent or FC in packages/forms.

### DIFF
--- a/packages/forms/src/components/CheckButton/CheckButton.tsx
+++ b/packages/forms/src/components/CheckButton/CheckButton.tsx
@@ -47,7 +47,9 @@ const CheckButton = ({
     onToggle();
   };
 
-  let Icon: string | React.FC<React.SVGProps<SVGSVGElement>> = "span";
+  let Icon:
+    | string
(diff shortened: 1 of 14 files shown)


packages/forms/src/components/CheckButton/CheckButton.tsx
```diff
@@ -47,7 +47,9 @@ const CheckButton = ({
     onToggle();
   };
 
-  let Icon: string | React.FC<React.SVGProps<SVGSVGElement>> = "span";
+  let Icon:
+    | string
+    | React.ForwardRefExoticComponent<React.SVGProps<SVGSVGElement>> = "span";
```
petertgiles, line 52: This is the one I was a little unsure how to handle.

esizer (approved):
Nice! LGTM 🏅 

I think the icon props make sense 👍

PR #6013, opened 2023-03-20
[Debt] Webpack asset management

### LINKED ISSUE #5356: ♻️ Allow webpack to process images (using imports instead of "public" dir)
## ♻️ Debt/Refactor

Allow webpack to process images (using imports instead of "public" dir) 

 @patcon - Could you please fill out this stub?

## 🕵️ Details

Should be as easy as using [webpack asset management](https://webpack.js.org/guides/asset-management/#loading-images) since we are on v5. The more time consuming part would be replacing all of our imageUrl usage with import statements.

## 🙋‍♀️ Proposed Solution

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Get rid of image helper and use import instead for all images in the app 
- [ ] All images in the app loads properly without any issues

esizer:
🤖 Resolves #5356 

## 👋 Introduction

Updates the webpack config to process static assets.

## 🕵️ Details

This does multiple things in order to function

- Moves all images to a directory under `src/` so they can be imported
- Adds an `index.d.ts` to declare modules for image asset imports in typescript
- Updates the webpack config to allow for processing of assets
- Updates all image use to be imports instead of `imageUrl` helper
- Removes the `imageUrl` helper
- Removes any images that were not being used

## 🧪 Testing

1. Build the site `npm run build`
2. Navigate to pages with images (all "marketing" pages along with employment equity)
3. Confirm images still appear as expected
4. Run storybook
5. Confirm images still appear


packages/webpack-config/webpack.base.js
```diff
@@ -93,6 +93,10 @@ module.exports = (basePath) => {
           test: /\.css$/i,
           use: [MiniCssExtractPlugin.loader, "css-loader"],
         },
+        {
+          test: /\.(png|svg|jpg|jpeg|gif)$/i,
```
petertgiles, line 97: These filenames are stable hashes, right?  The names aren't going to change unless we change the images themeselves?
image

petertgiles (approved):
Very cool!  Nice and clean.

packages/webpack-config/webpack.base.js
esizer: okay, so it is the same as the bundle filename. Meaning, it is a hash of the file itself so it should be stable. That being said, we can add the original filename if we want, like how we are with our chunks and bundles.

- https://webpack.js.org/configuration/output/#outputassetmodulefilename
- https://webpack.js.org/configuration/output/#outputfilename

esizer: Cool, I double confirmed that the hash remains the same by:

1. `rm -rf dist`
2. `npm run build:fresh`
3. Check hash for an image on the page
4. Repeat steps 1-3

Screenshot 2023-03-22 103305
Screenshot 2023-03-22 103556


- `background-image: url("/11dd19b67fa6015463ef.jpg");`
- `background-image: url("/11dd19b67fa6015463ef.jpg");`

PR #6014, opened 2023-03-20
Tooling 4461 seeding improvements

### LINKED ISSUE #4461: Update Seeders
There are a couple of missing things in the seeders that bother me.  Feel free to add more when you find them:
- Published pool advertisements should have a stream
- The "main" users like applicant@test.com should have a generic job title set to complete the profile

vd1992:
🤖 Resolves #4461 

## 👋 Introduction

Improve database seeding for quality of life purposes. 

## 🕵️ Details

- sometimes, users would be looking for neither english nor french nor bilingual, which needed a solution
- published pools now always have a stream
- digital careers pool accepts various, so any, languages
- dcm team would have one pool, test team zero pools, adjusted so the pool table isn't so empty
- applicant@test.com always has an experience that fulfills the essential skills of the digital careers pool

## 🧪 Testing

1. seed, bounce around pools, look at opportunities, apply to DCM pool as applicant@test.com
2. re-seed and and check again to be sure

Seeding should not error, applying as applicant should be quick, and published pools have streams.

### DIFF
--- a/api/database/seeders/DatabaseSeeder.php
+++ b/api/database/seeders/DatabaseSeeder.php
@@ -63,11 +63,13 @@ class DatabaseSeeder extends Seeder
         // Seed some expected values
         $this->seedPools();
 
+        $digitalTalentPool = Pool::where('key', "digital_careers")->sole();
+
         User::factory([
             'legacy_roles' => [ApiEnums::LEGACY_ROLE_APPLICANT]
         ])
             ->count(150)
-            ->afterCreating(function (User $user) use ($faker) {
+            ->afterCreating(function (User $user) use ($faker, $digitalTalentPool) {
 
                 $genericJobTitles = GenericJobTitle::inRandomOrder()->limit(2)->pluck('id')->toArray();
                 $user->expectedGenericJobTitles()->sync($genericJobTitles);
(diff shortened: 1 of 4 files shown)


api/database/seeders/DatabaseSeeder.php
```diff
@@ -63,11 +63,13 @@ public function run()
         // Seed some expected values
         $this->seedPools();
 
+        $digitalTalentPool = Pool::where('key', "digital_careers")->sole();
```
petertgiles, line 66: Good catch, pulling this out of the loop.

api/tests/Feature/PoolCandidateTest.php
```diff
@@ -545,18 +545,27 @@ public function testFilterByEmploymentEquity(): void
 
     public function testFilterByLanguageAbility(): void
     {
+        $noLanguageUser = User::factory()->create();
+        $noLanguageUser->looking_for_english = false;
```
petertgiles, line 549: To set intial values, pass an object with key-value pairs to `create`, like on line 555.

api/tests/Feature/UserTest.php
```diff
@@ -683,10 +683,16 @@ public function testFilterByCandidateExpiryDate(): void
 
     public function testFilterByLanguageAbility(): void
     {
-        User::factory()->count(5)->create([
-            'looking_for_english' => false,
-            'looking_for_french' => false,
-            'looking_for_bilingual' => false,
+        $noLanguageUser = User::factory()->create();
+        $noLanguageUser->looking_for_english = false;
```
petertgiles, line 687: To set intial values, pass an object with key-value pairs to `create`, like on line 693.

api/tests/Feature/PoolCandidateTest.php
vd1992: okay...
done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/6014/commits/661fc939f34e583a40b2311c26d43658e020db8b

petertgiles: Oh yeah, I like that much better!

PR #6019, opened 2023-03-21
Remove FC from IAP and Profile pages

petertgiles:
🤖 Related to (but does not close) #5214 

## 👋 Introduction

This branch removes the most (but not all!) of the uses of React.FunctionComponent and React.FC from the IAP and profile pages  Other subsequent PRs will finish removing it from this location and remove it from other locations.

## 🧪 Testing

1. App builds
2. App runs OK and looks normal

> **Note**
> If any diffs looks really big, review with whitespace changes hidden it will appear more familiar.


esizer (approved):
🚅 Let's keep this train going!

PR #6024, opened 2023-03-21
Seeding - add users with specific roles to local seeder

vd1992:
## 👋 Introduction

Seeding some users for convenience, small and quick addition.

## 🕵️ Details

While testing with roles in mind, it can be a bit slow if you don't immediately have a user with the specific permissions ready to go. And then rather than edit users with permissions in mind, it is easier to create some role specific ones ready to go.

platform@test.com -> platform-admin
request@test.com -> request-responder
pool@test.com -> pool-operator

All share base-user, without the applicant role you can immediately see login directs people to an error page. 

## 🧪 Testing

1. seed and login
2. policies aren't all sorted out yet so things will look off


petertgiles (changes requested):
This is cool!  :sunglasses: Could you give them all the applicant role, too?  Technically all new users get this role too.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/ca16b33f39ab42704b762e983ce8bf7d916a7ffb/api/app/Providers/AuthServiceProvider.php#L105

petertgiles (approved):

PR #6025, opened 2023-03-22
[Bug] Fix locale swticher

### LINKED ISSUE #6023: 🐛 Language toggle link doesn't work in Safari browser ( production )
## 🐛 Bug

Language toggle link doesn't work in Safari browser in our home page 

1. Navigate to talent.canada.ca in safari browser
2. Click on French link on the home page 
3. Observe the page is redirected back to english version

1. Navigate to talent.canada.ca/fr in safari browser
2. Click on English link on the home page 
3. Observe the page is redirected back to french version

## 🦋 Expected Behaviour

What should be happening?

## 🕵️ Details

- Environment - Production (local works fine ) 
- **Browser**: Safari 

## 📋 Steps to Reproduce

1. ...
5. ...

## 📸 Screenshot

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] Criteria 1
- [ ] Criteria 2

## 🛑 Blockers

Blocked by {issue numbers}

esizer:
🤖 Resolves #6023 

## 👋 Introduction

This fixes the language toggle in safari.

## 🕵️ Details

For some odd reason, safari was not deriving the locale from the context and would get stuck on the current locale. This updates the language toggle link to set the context as well as navigate to the new URL (which seems to be the source of truth for safari).

## 🧪 Testing

2. Open any page in **safari**
3. Click the language toggle
4. Confirm it switches languages with no issue

## 📸 Screencast

https://user-images.githubusercontent.com/4127998/226908647-cde56568-c7f3-458e-acd1-da5b3cdda5fe.mov

### DIFF
--- a/apps/web/src/components/Header/Header.tsx
+++ b/apps/web/src/components/Header/Header.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { useIntl } from "react-intl";
-import { useLocation } from "react-router-dom";
+import { useLocation, useNavigate } from "react-router-dom";
 
 import {
   localizePath,
@@ -16,6 +16,7 @@ export interface HeaderProps {
 
 const Header = ({ width }: HeaderProps) => {
   const intl = useIntl();
+  const navigate = useNavigate();
   const { locale, setLocale } = useLocale();
 
   const location = useLocation();
@@ -99,6 +100,7 @@ const Header = ({ width }: HeaderProps) => {
                 onClick={(e) => {
                   e.preventDefault();
                   setLocale(changeToLang);
+                  navigate(languageTogglePath);
                 }}
                 lang={changeToLang === "en" ? "en" : "fr"}
               >


petertgiles (approved):
It works!  I don't fully understand but it does the trick.  :laughing:   I tested locally on Firefox and Chromium and on my TBS iPhone in Safari.

Why can't we just navigate and use our regular Link component?  Why do we still need the setLocale?
```
Matched leaf route at location "/en/" does not have an element or Component. This means it will render an <Outlet /> with a null value by default resulting in an "empty" page. 
```

esizer:
> Why can't we just navigate and use our regular Link component? Why do we still need the setLocale?

Technically we could. `setLocale` is faster than navigating but I guess if it doesn't work in safari then we may as well just navigate 🤔

petertgiles:
> I guess if it doesn't work in safari then we may as well just navigate

It would be simpler.  But my naive quick test didn't work.  :cry:

mnigh (approved):
Tested and works on macOS 12.6.3:

- Safari Version 16.3
- Chrome Version 111.0.5563.64
- Firefox Version 111.0.1
- Edge Version 111.0.1661.51

PR #6026, opened 2023-03-22
Final FC removal PR with linting rule

### LINKED ISSUE #5214: ♻️ Remove all uses of React.FC and React.FunctionComponent

petertgiles:
🤖 Resolves #5214 

## 👋 Introduction

This branch removes the remaining uses of React.FunctionComponent and React.FC from the repo.  It also adds a linting rule to prevent future use.

## 🧪 Testing

1. App builds
2. App runs OK and looks normal
3. No more instances of FC or FunctionComponent remain in repo
4. Attempt to add a new one and observe linting warning

### DIFF
--- a/packages/ui/src/components/SideMenu/SideMenu.test.tsx
+++ b/packages/ui/src/components/SideMenu/SideMenu.test.tsx
@@ -5,6 +5,7 @@ import "@testing-library/jest-dom";
 import React from "react";
 import { screen } from "@testing-library/react";
 import { renderWithProviders } from "@gc-digital-talent/jest-helpers";
+import { MagnifyingGlassIcon } from "@heroicons/react/24/solid";
 import SideMenu from "./SideMenu";
 import type { SideMenuProps } from "./SideMenu";
 import SideMenuItem from "./SideMenuItem";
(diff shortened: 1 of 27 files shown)


packages/ui/src/components/SideMenu/SideMenu.test.tsx
```diff
@@ -15,7 +16,7 @@ const defaultProps = {
   onToggle: jest.fn(() => null),
 };
 
-const icon = () => <>Icon</>;
+const icon = MagnifyingGlassIcon;
```
esizer, line 19: Nice to see the icon type works 😛

esizer (approved):
Awesome! 😎

PR #6041, opened 2023-03-22
Bug 6035 address errors with manual candidate addition

### LINKED ISSUE #6035: 🐛 Manually added pool candidates are NOT showing up in the UI
## 🐛 Bug

Manually added pool candidates are not showing up in the UI 

## 🦋 Expected Behaviour

Manually added Pool candidates should show up 

## 🕵️ Details

submitted_at column defines whether the application is draft are not. 
We show the candidates only when the application is not draft 
When the user manually adds the candidate we currently don't set the submitted_at column. (It is set only when the user go through application submission process ). 

We need to set this column during manual addition as well. 

## 📋 Steps to Reproduce

https://recordit.co/lRkJxUCmBQ

## 📸 Screenshot
image (14)
image (13)

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] When added manually pool candidate should show up in candidate status page and pool candidates table

## 🛑 Blockers

Blocked by {issue numbers}

vd1992:
🤖 Resolves #6035 

## 👋 Introduction

Twofold, correct the bug and update the database with an artisan command.

## 🕵️ Details

Corrected the bug by updating the input type to also take `submittedAt`, this is filled during the submission handler.

Update the database with a simple command to fill in dates for candidates with the specific status and null dates. Manual submission automatically sets the status to new application, which makes it easier to pull out the right ones to update. 

## 🧪 Testing

Two parts to check

Bug

1. proceed to view user page for some user
2. attempt to add them to pools
3. this should be successful and the table updates

Database

1. build off main and then head to view user for some user
2. add that user to multiple pools successfully, you will not see the table populate
3. switch branches to this one, run the command below
4. refresh page, the table should now be filled out

```
php artisan update-submitted-at
```

### DIFF
--- a/api/app/Console/Commands/SubmittedAtApplicationDates.php
+++ b/api/app/Console/Commands/SubmittedAtApplicationDates.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Models\PoolCandidate;
+use Illuminate\Console\Command;
+use Carbon\Carbon;
+use Database\Helpers\ApiEnums;
+
+class SubmittedAtApplicationDates extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'update-submitted-at';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'For pool candidates that should have submitted dates, submit them';
+
+    /**
+     * Execute the console command.
+     *
+     * @return int
+     */
+    public function handle()
+    {
+        $dateNow = Carbon::now();
+        $candidatesToSubmit = PoolCandidate::where('pool_candidate_status', ApiEnums::CANDIDATE_STATUS_NEW_APPLICATION)
+            ->where('submitted_at', null);
+
+        foreach ($candidatesToSubmit as $candidate) {
+            $candidate->update(['submitted_at' => $dateNow]);
+        }
+
+        return Command::SUCCESS;
+    }
+}
--- a/api/app/Console/Commands/SubmittedAtApplicationDates.php
+++ b/api/app/Console/Commands/SubmittedAtApplicationDates.php
@@ -32,7 +32,7 @@ class SubmittedAtApplicationDates extends Command
     {
         $dateNow = Carbon::now();
         $candidatesToSubmit = PoolCandidate::where('pool_candidate_status', ApiEnums::CANDIDATE_STATUS_NEW_APPLICATION)
-            ->where('submitted_at', null);
+            ->where('submitted_at', null)->get();
 
         foreach ($candidatesToSubmit as $candidate) {
             $candidate->update(['submitted_at' => $dateNow]);
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -131,7 +131,9 @@ type User {
   pools: [Pool]
     @hasMany
     @can(ability: "view", resolved: true)
-    @deprecated(reason: "Pools are now associated with a Team more than a single User. Use User.roles.team.pools instead.") # Pools a user owns
+    @deprecated(
+      reason: "Pools are now associated with a Team more than a single User. Use User.roles.team.pools instead."
+    ) # Pools a user owns
   # Profile Status
   isProfileComplete: Boolean
   expectedGenericJobTitles: [GenericJobTitle] @belongsToMany
@@ -276,7 +278,9 @@ type Applicant {
   jobLookingStatus: JobLookingStatus
     @rename(attribute: "job_looking_status")
     @deprecated(reason: "Removing with applicantDashboard feature flag")
-  poolCandidates: [PoolCandidate] @hasMany(scopes: ["notDraft"]) @can(ability: "view", resolved: true)
+  poolCandidates: [PoolCandidate]
+    @hasMany(scopes: ["notDraft"])
+    @can(ability: "view", resolved: true)
 
   hasDiploma: Boolean @rename(attribute: "has_diploma")
   locationPreferences: [WorkRegion] @rename(attribute: "location_preferences")
@@ -434,7 +438,9 @@ enum PublishingGroup {
 type PoolCandidate {
   id: ID!
   pool: Pool! @belongsTo @can(ability: "view", resolved: true)
-  poolAdvertisement: PoolAdvertisement @belongsTo(relation: "pool") @can(ability: "viewAdvertisement", resolved: true)
+  poolAdvertisement: PoolAdvertisement
+    @belongsTo(relation: "pool")
+    @can(ability: "viewAdvertisement", resolved: true)
   user: Applicant! @belongsTo(relation: "user")
 
   # cmoIdentifier can be an arbitrary string used to relate this candidate to an external database.
@@ -899,8 +905,7 @@ type Query {
     @guard
     @can(ability: "viewAnyApplicants", model: "User")
   # countApplicants returns the number of candidates matching its filters, and requires no special permissions.
-  countApplicants(where: ApplicantFilterInput): Int!
-    @count(model: "User")
+  countApplicants(where: ApplicantFilterInput): Int! @count(model: "User")
   pool(id: UUID! @eq): Pool @find @can(ability: "view", query: true)
   poolAdvertisement(id: UUID! @eq): PoolAdvertisement
     @find(model: "Pool")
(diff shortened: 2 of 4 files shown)


api/app/Console/Commands/SubmittedAtApplicationDates.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Execute the console command.
+     *
+     * @return int
+     */
+    public function handle()
+    {
+        $dateNow = Carbon::now();
+        $candidatesToSubmit = PoolCandidate::where('pool_candidate_status', ApiEnums::CANDIDATE_STATUS_NEW_APPLICATION)
```
brindasasi, line 34: But wouldn't this condition make all draft application also to be submitted as well ? 🤔 
I would think this command should be scoped to only those manually added applications.
I'm unsure how though. Probably adding the pool name?

vd1992: should not
created applications have the draft status

https://github.com/GCTC-NTGC/gc-digital-talent/blob/f3cb8731f011704811d5df3c24cda6a935792738/api/app/GraphQL/Mutations/CreateApplication.php#L45-L48

brindasasi: ooooo ... I think we thought `CANDIDATE_STATUS_NEW_APPLICATION` is draft status for that production issue. 
image (15)
Hence we added notDraft scope ..thats why these candidates are not showing up for admins any more 😅

brindasasi: I think we should be fine after this. Just need to make sure both draft & new application status are working fine with this change.

tristan-orourke, line 34: Instead of looking for status == NEW_APPLICATION, I think you should check for status !== DRAFT. Its possible they've changed the status of some candidates since manually creating them.

api/app/Console/Commands/SubmittedAtApplicationDates.php
```diff
@@ -0,0 +1,43 @@
...
+            ->where('submitted_at', null)->get();
+
+        foreach ($candidatesToSubmit as $candidate) {
```
tristan-orourke, line 37: You can do a mass update without a for loop - probably a lot faster! https://laravel.com/docs/10.x/eloquent#mass-updates

api/graphql/schema.graphql
```diff
@@ -1307,6 +1308,7 @@ input CreatePoolCandidateAsAdminInput {
   status: PoolCandidateStatus = NEW_APPLICATION
     @rename(attribute: "pool_candidate_status")
   notes: String
+  submittedAt: DateTime @rename(attribute: "submitted_at")
```
tristan-orourke, line 1311: Instead of allowing the client to (optionally) set a submittedAt time, the mutation should automatically set this to the current time, with no option to override.

@petertgiles @esizer I thought we figured out a way to do this before, but I don't see any examples in the current schema. Do either of you remember anything about this?

tristan-orourke (changes requested):

api/graphql/schema.graphql
esizer: I think this is what you are looking for?...maybe 😅 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/198fe5a9a4b01bace0d4d0b66c7e1bf9cbe03629/api/app/GraphQL/Mutations/SubmitApplication.php#L33-L45

tristan-orourke: I was hoping there was a way of doing it with schema directives, but that might be it.

esizer: Maybe this SO answer could help?

https://stackoverflow.com/a/71115849

tristan-orourke: Yeah, I was just looking at that, but I don't love that solution.

I'm thinking a good solution would be to write a custom directive, based on InjectDirective, which just gets the current time instead of getting a value from context.

tristan-orourke:
```
<?php

namespace Nuwave\Lighthouse\Schema\Directives;

use Nuwave\Lighthouse\Exceptions\DefinitionException;
use Nuwave\Lighthouse\Execution\ResolveInfo;
use Nuwave\Lighthouse\Schema\Values\FieldValue;
use Nuwave\Lighthouse\Support\Contracts\FieldMiddleware;
use Nuwave\Lighthouse\Support\Contracts\GraphQLContext;

class InjectNowDirective extends BaseDirective implements FieldMiddleware
{
    public static function definition(): string
    {
        return /** @lang GraphQL */ <<<'GRAPHQL'
"""
Inject the current DateTime into the arguments.
"""
directive @injectNow(
  """
  The target name of the argument into which the current dateTime is injected.
  You can use dot notation to set the value at arbitrary depth
  within the incoming argument.
  """
  name: String!
) repeatable on FIELD_DEFINITION
GRAPHQL;
    }

    /**
     * @throws \Nuwave\Lighthouse\Exceptions\DefinitionException
     */
    public function handleField(FieldValue $fieldValue, \Closure $next): FieldValue
    {
        // $contextAttributeName = $this->directiveArgValue('context');
        // if (! $contextAttributeName) {
        //     throw new DefinitionException(
        //         "The `inject` directive on {$fieldValue->getParentName()} [{$fieldValue->getFieldName()}] must have a `context` argument"
        //     );
        // }

        $argumentName = $this->directiveArgValue('name');
        if (! $argumentName) {
            throw new DefinitionException(
                "The `inject` directive on {$fieldValue->getParentName()} [{$fieldValue->getFieldName()}] must have a `name` argument"
            );
        }

        $previousResolver = $fieldValue->getResolver();

        $fieldValue->setResolver(function ($rootValue, array $args, GraphQLContext $context, ResolveInfo $resolveInfo) use ($contextAttributeName, $argumentName, $previousResolver) {
            // $valueFromContext = data_get($context, $contextAttributeName);

            // $resolveInfo->argumentSet->addValue($argumentName, $valueFromContext);

            $resolveInfo->argumentSet->addValue($argumentName, now());

            return $previousResolver(
                $rootValue,
                $resolveInfo->argumentSet->toArray(),
                $context,
                $resolveInfo
            );
        });

        return $next($fieldValue);
    }
}
```

api/app/Console/Commands/SubmittedAtApplicationDates.php
vd1992: good point, done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/6041/commits/2b56645e966205e52093dc632e6e23c51ef514cd

api/app/Console/Commands/SubmittedAtApplicationDates.php
vd1992: implemented! https://github.com/GCTC-NTGC/gc-digital-talent/pull/6041/commits/2b56645e966205e52093dc632e6e23c51ef514cd

api/graphql/schema.graphql
vd1992: I removed it from the front-end in https://github.com/GCTC-NTGC/gc-digital-talent/pull/6041/commits/60772a75f5c969a12c34e051f1f30c6bf8b19b26
and added in the above directive in https://github.com/GCTC-NTGC/gc-digital-talent/pull/6041/commits/52adcf582d295c35df2726bfc54ff302e3290b56

I can't quite get it to actually work however, not too sure what the problem is

tristan-orourke: Try this?
```
createPoolCandidateAsAdmin(
    poolCandidate: CreatePoolCandidateAsAdminInput! @spread
  ): PoolCandidate @create @guard @can(ability: "create") @injectNow(name: "submitted_at")
```

vd1992: oh yeah... good call

fixed in https://github.com/GCTC-NTGC/gc-digital-talent/pull/6041/commits/bf59e0bba60a9373f6789a2aff539edaf7c78955
attempted a test in https://github.com/GCTC-NTGC/gc-digital-talent/pull/6041/commits/0c6feeab9dc904282ae37afe470a73af4ee703da
and cleared an outstanding to do on that input in https://github.com/GCTC-NTGC/gc-digital-talent/pull/6041/commits/effae10be212b964c745d0715e6f5597e4eef3f2

tristan-orourke (changes requested):
Good work! I'd just remove submittedAt from CreatePoolCandidateAsAdminInput entirely, since we don't want to allow it to be overridden.

tristan-orourke (approved):

PR #6064, opened 2023-03-23
[Bug] Fix user admin page auth

### LINKED ISSUE #5926: 🐛 Admin users/:id and users/:id/profile not available to Pool Operator and Request Responder
blocked by #5781 , #5783 

## 🐛 Bug

Request Responders and Pool Operators aren't able to view user profiles. Discovered by @vd1992 and first raised [here](https://github.com/GCTC-NTGC/gc-digital-talent/pull/5858#issuecomment-1464727216).

## 🦋 Expected Behaviour

Request Responders should be able to view Applicant profiles (meaning /users/:id and /users/:id/profile), and Pool Operators should be able to view those pages if the user has applied to a pool in their team.

## 🕵️ Details

Currently there are two problems: client-side role checking in Router.tsx, and backend permissions for the applicant.poolCandidates graphql query.

## 📋 Steps to Reproduce

1. Create two pools in two different teams
2. As an applicant, apply to both pools
3. As a user with request_responder role, but not platform_admin, try to go to en/admin/users/:id where :id is that of the applicant user. From there you should be able to view User Profile, but not Edit User account.
4. As a user with the pool_operator in on of the pool teams, you should be able to go to /pools, view candidates in the pool, accept the user into the pool, then follow the link to view the user's profile. In the Pool Status section, you should not see any pools not on your team. From there you should be able to view User Profile, but not Edit User account.

## 📸 Screenshot

Pool Operator gets unauthorized error if the user has applied to any pools in another team.
Pool Operators gets unauthorized error

Pool Operators should see pool status section, but for pools in their own team.
image

## 🙋‍♀️ Proposed Solution

In the Router.tsx route tree, a RequestAuth section cannot be more restrictive than its nested routes, or it will block them anyway. Therefore, `path: ":userId"` and its index route should allow all three kinds of admin roles.

The Pool Operators getting an auth error because of the pool status section is more complicated. Several possible approaches:
1. Change the query so that it only returns PoolCandidates in teams which the user is allowed to see. This might require a new graphql query defined, maybe one which accepts a user-id and a list of valid team-ids?
2. Maybe its not so bad if a PoolOperator can see which other pools a user has applied to? Maybe we could make an equivalent of UserPublicProfile for PoolCandidate, or use heavy use of field-level permissions on PoolCandidate, so that _some_ data is available to PoolOperators regardless of the pool.
3. Easiest thing would be to just not render that whole section, for Pool Operators, and not run that part of the query.

## ✅ Acceptance Criteria

- [ ] a user with request_responder role should be able to view all Applicant profiles (meaning /admin/users/:id and /admin/users/:id/profile)
- [ ] a user with pool_operator role should be able to view Applicant profiles for users who applied to pools in their team (meaning /admin/users/:id and /admin/users/:id/profile), even if the user also applied to pools in other teams.

esizer:
🤖 Resolves #5926 

## 👋 Introduction

Allows both `pool_operator` and `request_responder` to view user information.

## 🕵️ Details

This makes some temporary fixes to allow some expected functionality as a patch until we can provide a more permanent solution.

## 🧪 Testing

### Pool Operator

1. Login as an admin
2. Create a basic pool in the "Test Team"
3. Login as an applicant and apply to the pool
4. Login as pool@test.com
5. Confirm you can view the applicant in that pool

### Request Responder

1. Login as request@test.com
2. Go to a request
3. Change the filters so you see some candidates
4. Click "View profile"
5. Confirm you can see the user


vd1992:
I think I found an edge case

applicant applies to pool you own, as well as other pools that you cannot access
you click on "view profile" for that applicant
authorization error

gif

esizer:
> I think I found an edge case for pool-operator

Good catch! I did not realize that the `pool@test.com` was part of all available pools 😛 

Fixed in [scope pool candidates to prevent unauthorized](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6064/commits/1c4e50d0abf8f4e18fab4bb1a5f0182dc5d935d0)

tristan-orourke:
> I think I found an **edge case** for pool-operator

> applicant applies to pool you own, as well as other pools that you cannot access
you click on "view profile" for that applicant
authorization error

I'll be honest - this was the main issue I was trying to describe in #5926. Guess I didn't do a great job describing it, hehe 😅

api/app/Models/Pool.php
```diff
@@ -150,4 +149,23 @@ public function scopeWasPublished(Builder $query, ?array $args)
...
+    {
+        $userId = Auth::user()->id;
+        $user = User::find($userId);
+
+        if ($user->isAbleTo("view-team-pool")) {
+            $teamIds = $user->rolesTeams()->get()->pluck('id');
+            return $query->whereHas('team', function (Builder $query) use ($teamIds) {
+                return $query->whereIn('id', $teamIds);
+            });
+        }
```
tristan-orourke, line 163: I know we only have one team role right now, buuuut... this might not generalize, since _technically_ you might not have the view-team-pool permission on of your teams. You'd have to check `isAbleTo("view-team-pool", $teamId)`, and only add it to teamIds if true.

api/app/Models/PoolCandidate.php
```diff
@@ -411,4 +411,29 @@ public function isDraft()
...
+    /**
+     * Scope the query to PoolCandidate's the current user can view
+     */
+    public function scopeAvailableToQuery(Builder $query)
+    {
+        $userId = Auth::user()->id;
+        $user = User::find($userId);
+
+        if ($user->isAbleTo("view-any-submittedApplication")) {
+            return $query; //We filter out drafts in another scope
```
tristan-orourke, line 424: If we implement a scope like this which scope to items the user is authorized to view, I think it should cover all cases, including the logic around drafts.

api/app/Models/PoolCandidate.php
```diff
@@ -411,4 +411,29 @@ public function isDraft()
     {
         return is_null($this->submitted_at) || $this->submitted_at->isFuture();
     }
+
+    /**
+     * Scope the query to PoolCandidate's the current user can view
+     */
+    public function scopeAvailableToQuery(Builder $query)
```
tristan-orourke, line 418: Does this cover the case where a user is trying to view their own application?

api/app/Models/PoolCandidate.php
esizer: I think this makes sense since the `notDraft` scope was introduced for an incredibly similar use case 👍 

[update scope to filter draft, restrict viewing own](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6064/commits/4b95d1f03c3c21c8d35155b715741e7fcb796fbd)

api/app/Models/PoolCandidate.php
esizer: I had that and removed it for some reason 🤔 Added back in [update scope to filter draft, restrict viewing own](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6064/commits/4b95d1f03c3c21c8d35155b715741e7fcb796fbd)

api/app/Models/Pool.php
esizer: Okay, I hope this makes sense. There are two checks (in the interest of not looping through teams when unnecessary).

[make pool team scope more strict](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6064/commits/159c55cb87ebbab945b0756f067ad054cbf1b65c)

api/app/Models/PoolCandidate.php
```diff
@@ -411,4 +411,32 @@ public function isDraft()
...
+    {
+        $userId = Auth::user()->id;
+        $user = User::find($userId);
+
+        if ($user->isAbleTo("view-any-submittedApplication")) {
+            return  $query->where('submitted_at', '<=', Carbon::now()->toDateTimeString());
```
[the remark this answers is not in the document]
esizer: Ahh, good call! I didn't even think of that 👍 

Updated in [update scopes to not return early](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6064/commits/e4687dcd8587110f14bc106a22fd1401ae3a00b6)

tristan-orourke (changes requested):
Okay, I'm still a little nervous about adding an authorization-based scope while the rest of the schema uses very different patterns, but this seems like it should work for now.

Last couple suggestions - I don't think we need the new scope on the pools in the schema, and can we rename the scope to authorizedToView?

api/graphql/schema.graphql
```diff
@@ -589,7 +589,7 @@ type PoolCandidateFilter {
   operationalRequirements: [OperationalRequirement]
     @rename(attribute: "operational_requirements")
   workRegions: [WorkRegion] @rename(attribute: "work_regions")
-  pools: [Pool] @belongsToMany @can(ability: "view", resolved: true)
+  pools: [Pool] @belongsToMany(scopes: ["availableToQuery"]) @can(ability: "view", resolved: true)
```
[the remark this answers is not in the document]
esizer: Updated in [rename scope, remove from pool candidate filter](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6064/commits/830f56ab9a2d86511db9c5f33ce9c6ab1ed58a18)

api/app/Models/Pool.php
```diff
@@ -150,4 +149,37 @@ public function scopeWasPublished(Builder $query, ?array $args)
         $query->where('published_at', '<=', Carbon::now()->toDateTimeString());
         return $query;
     }
+
+    public function scopeauthorizedToView(Builder $query)
```
tristan-orourke, line 153: I think by convention the casing for this should be "scopeAuthorizedToView"

api/app/Models/PoolCandidate.php
```diff
@@ -411,4 +411,35 @@ public function isDraft()
     {
         return is_null($this->submitted_at) || $this->submitted_at->isFuture();
     }
+
+    /**
+     * Scope the query to PoolCandidate's the current user can view
+     */
+    public function scopeauthorizedToView(Builder $query)
```
tristan-orourke, line 418: Casing again - scopeAuthorizedToView

api/graphql/schema.graphql
```diff
@@ -608,7 +608,7 @@ type ApplicantFilter {
   expectedClassifications: [Classification]
     @belongsToMany(relation: "classifications")
   skills: [Skill] @belongsToMany
-  pools: [Pool] @belongsToMany @can(ability: "view", resolved: true)
+  pools: [Pool] @belongsToMany(scopes: ["authorizedToView"]) @can(ability: "view", resolved: true)
```
tristan-orourke, line 611: The authorizedToView scope is unnecessary here, since all the pools attached to a filter should already by published.

api/graphql/schema.graphql
```diff
@@ -912,7 +912,7 @@ type Query {
     @all(model: "Pool", scopes: ["wasPublished"])
     @can(ability: "viewAnyPublishedAdvertisement", model: "Pool")
   poolByKey(key: String! @eq): Pool @find @can(ability: "view", query: true)
-  pools: [Pool]! @all @can(ability: "viewAny")
+  pools: [Pool]! @all(scopes: ["authorizedToView"]) @can(ability: "view", resolved: true)
```
tristan-orourke, line 915: The authorizedToView scope is unnecessary here, since all the pools attached to a search request should already by published.

api/graphql/schema.graphql
tristan-orourke: Or... Hmm. I'm assuming we should avoid using the scope if we can, since it makes the query more complex. If we decide to rely more heavily on this kind of scope, we might have to take a different stance. Another point for discussion in https://github.com/GCTC-NTGC/gc-digital-talent/discussions/6038 I suppose.

tristan-orourke (changes requested):
I tested manually, and everything looks good! Just a couple casing suggestions now, and possible efficiencies in the schema.

api/graphql/schema.graphql
esizer: Sorry, was this commented in the correct line? This seems to be the general `Query.pools` query which could cause errors if called without the `platform_admin` role (which is the case for `request_responder` currently).

api/graphql/schema.graphql
esizer: Do we know if our seeders generate invalid data? 🤔 Removing the scopes just results in the same error we were already getting:

```js
[
    {
        "message": "This action is unauthorized.",
        "extensions": {
            "category": "authorization"
        },
        "path": [   "poolCandidateSearchRequests",  0,  "applicantFilter", "pools" ]
    }
]
```

api/graphql/schema.graphql
tristan-orourke: Oh! Yes, that's a mistake on my part, I was thinking this line was part of the PoolCandidateSearchRequest block, not the Query block 🤦‍♂️

api/graphql/schema.graphql
esizer: Okay, yeah. Seems like this could be an issue with seeders since we just attach a random pool in the factory.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/f40006c6da0a723bd7d35c6968d2c22c2e2b5393/api/database/factories/ApplicantFilterFactory.php#L85-L88

api/graphql/schema.graphql
tristan-orourke: Where are request_responders running the pools query?

esizer: On the Talent Requests page there is the "Candidate Results" table that queries pools right now

api/app/Models/PoolCandidate.php
esizer: Fixed in [fix scope case](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6064/commits/49c451143e61f09be2caa988e15c3aa140db8893)

api/graphql/schema.graphql
tristan-orourke: Ah, of course

api/graphql/schema.graphql
esizer: Updated seeders in [fix filter factories to only attach eligible pools](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6064/commits/fdee787f7911e28a51cb73398314ac55e07701a7) and removed scope in [remove unnecessary scope from applicant filter](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6064/commits/7576dfff980bff3d5f46615b3052a6f8beaf3f9f)

tristan-orourke (approved):
🎉

PR #6077, opened 2023-03-23
[Feature] Inline Toggle Component

### LINKED ISSUE #5940: ⚙️ Build `Inline Editable Form` component
# ⚙️ Summary
The pattern enables users to edit the form text directly on the same page instead of navigating them to another page for editing the text. This is useful when the user may want to make a single edit without the need to refresh the entire form to make it editable.

### [[Visit the Figma file](https://www.figma.com/file/BXYVsu1fo9LS2IySJDUINY/Applicants---Application?node-id=653%3A59471&t=1mqKGZguvjMWyJZF-0)]

**For all layout and spacing guidance, including how this component should respond on smaller devices, please see the Figma file.**

# 💡 Interaction states

## Initial mode (without data) 
image

## Editable mode
image

## Editable mode - Error state
image

## Display mode
Screen Shot 2023-03-15 at 5 53 09 PM

## Display mode - Error state 
Screen Shot 2023-03-15 at 6 01 39 PM

# 🛠️ Properties

## Mode 
Initial, Display, Editable 

## ✅ Acceptance Criteria
- [ ] Component has initial mode, display mode, editable mode 
- [ ] Component can have nested Inline Editable Form component as child as well 
- [ ] added in storybook 
- [ ] axe-linter check for accessibility
- [ ] focus goes to first editable input on the edit mode
- [ ] this is a generic component and so show off this component with simple field list ( doesn't have to be like in the mock up )  

## ✅ Accessibility checklist
- [ ] Tap targets are adequate on smaller screens
- [ ] Color contrast for all modes and states has been checked and cleared
- [ ] Zoom has been considered as layout changes occur
- [ ] `alt` text has been identified and labelled
- [ ] Non-standard states or interactions have been described and provided accessible design context

esizer:
🤖 Resolves #5940 

## 👋 Introduction

This adds a generic component that toggles between two different states.

## 🕵️ Details

This is based on `@radix-ui` pattern to match our other generic components.

## ⚛️ Component Implementation

```tsx
{/** Required: Provides the context for subcomponents */}
<ToggleSection.Root onOpenChange={(open) => console.log(open)}>

  {/** Optional styling (accepts HeadingProps as well as toggle */}
  <ToggleSection.Header 
    toggle={() => (
      <ToggleSection.Toggle>
        <Button>Open/Close</Button>
      </ToggleSecction.Toggle>
    )}
  >
      Toggle Section Heading
  </ToggleSection.Header>
  
  {/** Optional: Styled to wrap the content that will be toggled */}
  <ToggleSection.Content>

    {/** Required: The content that appears when `open=false` */} 
    <ToggleSection.InitialContent>
      <p>The section is closed.</p>
    </ToggleSection.InitialContent>


    {/** Required: The content that appears when `open=true` */} 
    <ToggleSection.OpenContent>
      <p>The section is open.</p>
    </ToggleSection.OpenContent>

  </ToggleSection.Content>

</ToggleSection.Root>
```

## :accessibility: Accessibility

- Applies `aria-controls` to the triggers that toggle/open/close the section.
- Applies `aria-selected` attribute to the content `div`s when they are open/closed
- No focus management

> **Note**
> The more I thought about this component, the more I believe that focus management was not needed. This is effectively a combination of collapse/accordion/tabs which do not implement any sort of focus management. So, I opted to leave it out and use the `aria-controls` and `aria-selected` to supplement the component instead.

## 🧪 Testing

2. Navigate to the "Toggle Section" component
3. Interact with it and confirm it works as expected

## 📸 Screenshot

Screenshot 2023-03-23 154935

### DIFF
--- a/packages/ui/src/components/ToggleSection/ToggleSection.stories.tsx
+++ b/packages/ui/src/components/ToggleSection/ToggleSection.stories.tsx
@@ -49,3 +49,48 @@ const Template: ComponentStory<typeof ToggleSection.Root> = (args) => (
 );
 
 export const Default = Template.bind({});
+
+const NestedTemplate: ComponentStory<typeof ToggleSection.Root> = (args) => (
+  <ToggleSection.Root
+    {...args}
+    onOpenChange={(open) => action("onOpenToggle")(open)}
+  >
+    <ToggleSection.Header Icon={AcademicCapIcon} toggle={<Toggle />}>
+      Toggle Section
+    </ToggleSection.Header>
+
+    <ToggleSection.Content data-h2-text-align="base(center)">
+      <ToggleSection.InitialContent>
+        <p>Initial Content Here</p>
+        <ToggleSection.Open>
+          <Button mode="inline">Open Main Content</Button>
+        </ToggleSection.Open>
+
+        <ToggleSection.Root>
+          <ToggleSection.Trigger>
+            <Button mode="inline">Toggle Nested Content</Button>
+          </ToggleSection.Trigger>
+
+          <ToggleSection.Content data-h2-text-align="base(center)">
+            <ToggleSection.InitialContent>
+              <p>Nested Initial Content Here</p>
+            </ToggleSection.InitialContent>
+
+            <ToggleSection.OpenContent>
+              <p>Nested Open Content Here</p>
+            </ToggleSection.OpenContent>
+          </ToggleSection.Content>
+        </ToggleSection.Root>
+      </ToggleSection.InitialContent>
+
+      <ToggleSection.OpenContent>
+        <p>Open Content Here</p>
+        <ToggleSection.Close>
+          <Button mode="inline">Close</Button>
+        </ToggleSection.Close>
+      </ToggleSection.OpenContent>
+    </ToggleSection.Content>
+  </ToggleSection.Root>
+);
+
+export const Nested = NestedTemplate.bind({});


packages/ui/package.json
```diff
@@ -36,9 +36,12 @@
     "@radix-ui/react-dropdown-menu": "^2.0.4",
     "@radix-ui/react-scroll-area": "^1.0.3",
     "@radix-ui/react-separator": "^1.0.2",
+    "@radix-ui/react-slot": "^1.0.1",
     "@radix-ui/react-switch": "^1.0.2",
     "@radix-ui/react-tabs": "^1.0.3",
+    "@radix-ui/react-toggle": "^1.0.2",
     "@radix-ui/react-toggle-group": "^1.0.3",
+    "@radix-ui/react-use-controllable-state": "^1.0.0",
```
petertgiles, line 44: I'm having trouble finding any documentation on this one.  Is this OK to use?

packages/ui/src/components/ToggleSection/ToggleSection.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+interface HeaderProps extends HeadingProps {
+  /** The toggle for the component (appears on right side of header) */
+  toggle: React.ReactElement<typeof Toggle>;
+}
+
+/**
+ * A styled header for the section
+ */
+const Header = React.forwardRef<HTMLHeadingElement, HeaderProps>(
```
petertgiles, line 264: I wonder if including things like a header and icon are a bit too specific for a reusable component?  Is open/closed even too specific?  I wonder about maybe having this really generic bare compoent with primary/secondary states and then moving the styling to more focused InlineEditableForm component on top of it?  Feel free to disagree.  :laughing:  The InlineEditableForm might have support for the error alert, too?

packages/ui/src/components/ToggleSection/ToggleSection.tsx
```diff
@@ -0,0 +1,294 @@
...
+import Heading, { HeadingProps } from "../Heading";
+
+import {
+  ToggleSectionProvider,
+  useToggleSectionContext,
+} from "./ToggleSectionProvider";
+
+// Names used for unique IDs
+const NAME = {
+  INITIAL_CONTENT: "InitialContent",
```
petertgiles, line 15: It seems like the design call for three state: Initial, Display, Editable

packages/ui/src/components/ToggleSection/ToggleSection.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <p>Open Content Here</p>
+        <ToggleSection.Close>
+          <Button mode="inline">Close</Button>
+        </ToggleSection.Close>
+      </ToggleSection.OpenContent>
+    </ToggleSection.Content>
+  </ToggleSection.Root>
+);
+
+export const Default = Template.bind({});
```
petertgiles, line 51: Could we get a story with a nested section?

packages/ui/src/components/ToggleSection/ToggleSection.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+  >
+    <ToggleSection.Header Icon={AcademicCapIcon} toggle={<Toggle />}>
+      Toggle Section
+    </ToggleSection.Header>
+
+    <ToggleSection.Content data-h2-text-align="base(center)">
+      <ToggleSection.InitialContent>
+        <p>Initial Content Here</p>
+        <ToggleSection.Open>
+          <Button mode="inline">Open</Button>
```
petertgiles, line 37: Is there a way to test this AC?
> focus goes to first editable input on the edit mode

It seems like currently there's no focus change on toggle.

petertgiles (commented):
This is looking very cool!  I have a few questions and comments.

packages/ui/package.json
esizer: Probably not. Though, it is exactly what I was looking for here. I could re-write it and link to original implementation as a reference.

packages/ui/src/components/ToggleSection/ToggleSection.tsx
esizer: Yeah, I was having a problem with making it generic enough while providing some built in stuff that would always be there. This is a totally optional part. You can compose this component excluding the header. Though, if we think this is too much it can be removed and implemented on a as needed basis as well.

packages/ui/src/components/ToggleSection/ToggleSection.tsx
esizer: I did see that and had a hard time understanding the real difference between initial and display. Let me know if my assumption was correct:

- **Initial**: Effectively a null state
- **Display**: Primary content that is displaced before "opening" the component (could be null or display some info)
- **Editable**: Secondary content that appears when the component is "opened"

So, my understanding was that initial and display are just two different states that could appear in the primary content and not something the user can control through a toggle.

packages/ui/src/components/ToggleSection/ToggleSection.stories.tsx
petertgiles: Oh yeah, missed that note, thanks.

packages/ui/package.json
esizer: Found this discussion specifically about this and think it may be more stable to re-implement this ourselves.

https://github.com/radix-ui/primitives/discussions/1146

petertgiles: Yeah, good find.  We have separate packages now so we've got a good home for it.  :smile:

esizer: Okay, I wasn't sure if they should be their own package or not since they are only being used in the one place 🤔 

But, I added them in [re-implement radix hooks](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6077/commits/8a0f4c9f0b4d8c2307a9259908558dce13f6814e). Let me know if we should make them separate packages or not or If you believe they should be placed elsewhere.

packages/ui/src/components/ToggleSection/ToggleSection.stories.tsx
esizer: Yup! Added in [add story for nested toggle sections](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6077/commits/bbf75830e5d65a132d866b5dff36cb24fc31b79f)

packages/ui/package.json
petertgiles: Oh yeah, I like this in the hooks package!

packages/ui/src/components/ToggleSection/ToggleSection.tsx
petertgiles: Well, let's leave it as is - this might be premature over-engineering anyway.  I wouldn't be surprised if we need to split this in half later but let's not worry about it for now.

packages/ui/src/components/ToggleSection/ToggleSection.tsx
petertgiles: Yeah, open/close is what we need in this component then.

packages/ui/src/components/ToggleSection/ToggleSection.stories.tsx
petertgiles: Love it!

petertgiles (approved):
Let's go!  :rocket:

PR #6080, opened 2023-03-23
Feature - 5999 - platform admin can view pools table

### LINKED ISSUE #5999: ✨ Make Pools table accessible to Platform Admins
## ✨ Feature

The Pools table page has client-side role checking that restricts it to Pool Operators. Open it up to Platform Admins, but allow them to view all pools instead of only team pools.

## 🕵️ Details

Platform Admins need to see all Pools so they can find Draft Pools which need publishing. (We ended up deciding that Platform Admins are the only role able to publish.) 

When a Pool Operator visits this page they should only see pools in their teams - that's the current behaviour. But Platform Admins should see all Pools, without being in all teams. 

## 🎨 Design File

## 📸 Screenshot

I had been trying to avoid this, but I think now that the only way to implement this is to use different queries depending on user role. All Pools if the user has Platform Admin role, and via roleAssignments otherwise. Note that hooks cannot be conditional, but the urql query hooks allow for [pausing execution](https://formidable.com/open-source/urql/docs/api/urql/).

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Platform Admins can visit the Pools table page
- [ ] If user is not a platform admin, they see only pools in their team.
- [ ] If the user is a platform admin, or a platform admin AND a pool operator, they see all pools.
- [ ] Jest tests or integration tests for the two different queries?

## 🛑 Blockers

Blocked by #5781

vd1992:
🤖 Resolves #5999 

## 👋 Introduction

Enables users with platform admin but not pool operator role to also view the pool table.

## 🕵️ Details

Simpler solution (I think) than the query pause idea. Just two wrapper components with different queries and conditionally render one or the other.
Platform admin can now also view a pool layout, info, and edit pages (need to in order to publish) but not candidates.

## 🧪 Testing

1. use adminer or whichever tool to keep reference of all pools
2. login as pool operator (pool@test.com), view pools table and note pools displayed
3. login as platform admin (platform@test.com), view pools table and note pools displayed
4. pool operator can only view dcm/test team pools whereas the other can view all so the table has more pools
5. platform admin can also see information for the pools but not view candidates

### DIFF
--- a/apps/web/src/pages/Pools/IndexPoolPage/IndexPoolPage.tsx
+++ b/apps/web/src/pages/Pools/IndexPoolPage/IndexPoolPage.tsx
@@ -9,7 +9,10 @@ import SEO from "~/components/SEO/SEO";
 
 import AdminContentWrapper from "~/components/AdminContentWrapper/AdminContentWrapper";
 import { checkRole } from "~/utils/teamUtils";
-import PoolTableApi from "./components/PoolTable";
+import {
+  PoolOperatorTableApi,
+  PoolAdminTableApi,
+} from "./components/PoolTable";
 
 export const PoolPage = () => {
   const intl = useIntl();
@@ -46,7 +49,7 @@ export const PoolPage = () => {
     <AdminContentWrapper crumbs={navigationCrumbs}>
       <SEO title={pageTitle} />
       <PageHeader icon={Squares2X2Icon}>{pageTitle}</PageHeader>
-      <PoolTableApi />
+      {isAdmin ? <PoolAdminTableApi /> : <PoolOperatorTableApi />}
     </AdminContentWrapper>
   );
 };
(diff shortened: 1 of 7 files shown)


apps/web/src/pages/Pools/IndexPoolPage/IndexPoolPage.tsx
```diff
@@ -42,7 +49,7 @@ export const PoolPage = () => {
     <AdminContentWrapper crumbs={navigationCrumbs}>
       <SEO title={pageTitle} />
       <PageHeader icon={Squares2X2Icon}>{pageTitle}</PageHeader>
-      <PoolTableApi />
+      {isAdmin ? <PoolAdminTableApi /> : <PoolOperatorTableApi />}
```
petertgiles, line 52: I suggest pulling in the isLoaded field from useAuthorization and maybe use Pending so you don't render either of these until auth is settled.  It's impossible to get here without auth since auth is also used for routing but instead of depending on that to never change let's defend this page.

petertgiles (approved):
Nice!  I understand whey you needed to seed thos new users.  :laugh:

apps/web/src/pages/Pools/IndexPoolPage/IndexPoolPage.tsx
vd1992: like so https://github.com/GCTC-NTGC/gc-digital-talent/pull/6080/commits/798bb235a890746d527b2ddaff07f97074fa7d98 ?

petertgiles: Nice, that looks great!

PR #6083, opened 2023-03-24
User edit page - team based roles

### LINKED ISSUE #5788: ✨ (Teams) Edit User - Include Team based roles section
## ✨ Feature

With the implementation of Teams, we also need to be able to control these roles from the Edit User page, this new sections and interactions allows Admin users to reproduce the interaction of "adding memberships" to users through this page.

## 🕵️ Details

Things we need to be able to do in this new "Team Based Roles" section:

- Assign team based roles (role + team)
- Remove team based roles

I have included a new "On this page:" to this page since we now have two sections that we might want to treat separately. Particularly keep the "Save information" contained within the section that has open fields, as opposed to a global save. 

### For consideration

I reused tables as component, but for the purposes of this issue, we don't need the "search table" or "columns" functionality to work right out of the gate, if removing these would speed this up significantly I would opt for that.

## 🎨 Design File

[Link to figma](https://www.figma.com/proto/UsrB15yicgDCLDQmx3snyl/User-Flows?page-id=0%3A1&node-id=3031%3A4790&viewport=370%2C4599%2C0.31&scaling=min-zoom&starting-point-node-id=3060%3A6599&show-proto-sidebar=1)

## 📸 Screenshot

UPDATE USER (ADMIN ONLY)

## Team Based Roles

image

image

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Add new membership  pop up should show up when the user click on "Add new membership"
- [ ] Selection of Team dropdown should pull up the values for Membership roles dropdown ( All roles belong to that selected team should come up ) 
- [ ] Membership roles dropdown is a multiselect and is required field for submission
- [ ] Remove membership roles show up when the user clicks on the trash icon on the Membership table 
- [ ] Save changes button dismiss the pop up and the table should be updated for add or remove action
- [ ] Cancel button dismiss the pop up and the table show up without any changes
- [ ] Organization name in the table should be clickable and upon clicking the user should be taken to  "View Team" page

petertgiles:
🤖 Resolves #5788 

h/t to pat and eric for most of this code!  :smile: 

## 👋 Introduction

This branch adds the team-based roles table with add and remove functionality to the edit user page.

## 🧪 Testing

1. Build the app, log into admin, go to edit a user
2. Scroll down to the team-roles table and try the functionality
  - [ ] list team roles
  - [ ] add team roles
  - [ ] remove team roles 

## 📸 Screenshot

image


apps/web/src/lang/fr.json
```diff
@@ -6857,5 +6857,45 @@
...
+    "description": "Heading for updating a users team roles"
+  },
+  "3IZ3mN": {
+    "defaultMessage": "Équipe",
+    "description": "Title displayed for the role table display team column"
+  },
+  "GaMSN8": {
+    "defaultMessage": "Équipe",
+    "description": "Label for the input to select team of a team role"
+  },
```
vd1992, line 6872: Not an issue strictly related to this PR but we need to exercise a little caution with re-translating things multiple times. The biggest issue with re-translating things that should be the same is different translations coming back resulting in inconsistent wording at the French site. 

image

apps/web/src/pages/Users/UpdateUserPage/components/TeamRoleTable.tsx
```diff
@@ -0,0 +1,146 @@
...
+    <RemoveTeamRoleDialog
+      role={roleAssignment.role}
+      user={user}
+      team={roleAssignment.team}
+      onUpdateUser={onUpdateUser}
+    />
+  ) : null;
+};
+
+const teamCell = (displayName: string, href: string) => {
```
vd1992, line 44: gave multiple team based roles a try

image

and it seems to work fine! nice!

https://recordit.co/A3n0sNEbfO

vd1992: things look good to me, I want to let Eric/Pat have the final say since they worked on this a lot and may have code improvement suggestions

apps/web/src/lang/fr.json
petertgiles: Yeah, I know this has come up a few times.  We tend to err on the side of caution and separately translate the same word multiple times and up with a lot of duplicates.  In theory, the same word in English could be different words in French in different contexts but you have to know the language to know if this is case.  :smile:   And you could up with contradictions, as you mentioned.

apps/web/src/pages/Users/UpdateUserPage/components/TeamRoleTable.tsx
```diff
@@ -0,0 +1,146 @@
...
+import RemoveTeamRoleDialog from "./RemoveTeamRoleDialog";
+import { UpdateUserFunc } from "../types";
+
+const roleCell = (displayName: string) => {
+  return (
+    <Pill color="neutral" mode="solid">
+      {displayName}
+    </Pill>
+  );
+};
```
esizer, line 27: This may be a better question for @Jerryescandon but, I beleive we want to list all roles associated with a specific team in a single row. Kind of like:

| Actions     | Team      | Roles                           |
| ----------- | --------- | ------------------------------- |
| edit/remove | Team Name | Pool Operator, Pool Admin, etc. |

esizer (commented):
Looking really good, I just had one question about the table before approval 👍

apps/web/src/pages/Users/UpdateUserPage/components/TeamRoleTable.tsx
petertgiles: I think you're right, though it exposes a weird workflow.  If you want to remove a single team role while preserving a second team role on the same team you'll need to first remove all the team roles then add the preserved ones back in.  It's not important right now since we only have a single team role but it could get weird in the future.

esizer: Yeah, I agree. We talked about this but I cannot remember the decision we came to 😢 

@tristan-orourke do you remember the outcome of that discussion we had the other week where we discussed this?

petertgiles: The design for remove team membership looks like all or nothing to me.  If we roll up all the roles for a team to a single row, that row has a single "remove" button.
image

petertgiles: Well, you can click the "add new membership" button and select the team name in the list.  :laughing:

petertgiles: OK, I've done the team rollup.  We'll probably need an "edit membership" dialog someday, but not today.  I'll send away for the updated translation.
ref: [roll up teams and fix EN copy](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6083/commits/2233ec4eaea6d1d02d70df77d7fbb714e37d4cde)

apps/web/src/pages/Users/UpdateUserPage/components/AddTeamRoleDialog.tsx
```diff
@@ -0,0 +1,207 @@
...
+                  required: intl.formatMessage(errorMessages.required),
+                }}
+                placeholder={intl.formatMessage({
+                  defaultMessage: "Select team...",
+                  id: "ZdOvlC",
+                  description: "Placeholder text for team selection input",
+                })}
+                isLoading={teamsFetching}
+                options={teamOptions}
+              />
```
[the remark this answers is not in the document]
petertgiles: I don't feel strongly about it.  I've rolled it back with [switch to Select component](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6083/commits/2e53b17717640fce0ee2e09646e17901d178c26a).

apps/web/src/pages/Users/UpdateUserPage/components/TeamRoleTable.tsx
tristan-orourke: Ah, I didn't realize the figma designs only had the delete button. In last week's discussion we decided there should be an edit button beside the delete button, just like on the current Team Membership table (using essentially the same dialogs) 
image

patcon:
Looks good! Though I'd added a new `UpdateUserPage.stories.tsx` and renamed the old one to `UpdateUserForm.stories.tsx`, which was actually the extent of the component it rendered (not the Page). This is only worth mentioning because the new `UpdateUserPage.stories.tsx` was what made these new interface elements visible in storybook, and otherwise they don't have stories, since they're outside the form :)

patcon:
Also, curious your thoughts on this change (not fully tested, but types all work out):
https://gist.github.com/patcon/9f5fb664a2359c27b9390bf45d99e7cc

Basically, I was seeing `availableRoles` passed through levels and being processed in a few places, and it seemed like this was a very hook-able thing. The graphql query should be cached (I recall someone mentioned we already had SWR in our tools?), and so there's no explicit need to be passing the product of the query down through prop layers, since it will only call it once, right? (I understood this to be a benefit of what SWR unlocked -- that the query hook can be called lots of times right from the place where it's needed, and there's no added cost)

apps/web/src/pages/Users/UpdateUserPage/components/TeamRoleTable.tsx
petertgiles: OK, edit dialog added in [Edit dialog](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6083/commits/40ad1e7b72fb9171cbf568a64f14b9143f79c7ec) and [tidy up edits](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6083/commits/002abbafbb9505a6f114fefc43700e8262851809)

esizer (changes requested):
Looking good! 👍 Just a little tweak to the action cell

patcon:
If we want to cover `TeamMemberPage.stories.tsx` with a story as well (to cover the components used here), they are pretty similar:
https://gist.github.com/patcon/aa3b5be8fdd0e02521de156c5a3637f5

esizer (approved):

PR #6092, opened 2023-03-27
Feature - 5874 - add any missing corresponding policies

### LINKED ISSUE #5874: Make sure we didn't miss any endpoints or schema relationship which should have a policy but not there yet

yonikid15:
🤖 Resolves #5874 

## 👋 Introduction

All mutations and queries should have a corresponding policy method, ie a @can directive.
If a mutation or policy didn't have a @can directive, I added one and connected the appropriate policy method.

`GenericJobTitle` was the only model missing a `@can` directive and policy methods. I add the model to the [roles and permissions](https://docs.google.com/spreadsheets/d/1RC7wufxwjLEvaEunHsfZfOzBefg218dvvzkDwKSiQjI/edit#gid=428263758) and created policy methods (similar to the more basic models like Classifications, Departments, etc).

## 🕵️ Details

- I went through the entire schema and policies for each model:
  - [x] Checked all `@can` directives have a corresponding policy method.
  - [x] Ensured all `ability` values are spelled correctly

## 🧪 Testing

1. Ensure all roles and permissions are correct (rolePermissions.php)
2. Ensure policy methods are using the correct permissions.
3. Ensure all queries and mutations have a `@can` directive

### DIFF
--- a/api/app/Policies/ClassificationPolicy.php
+++ b/api/app/Policies/ClassificationPolicy.php
@@ -19,7 +19,7 @@ class ClassificationPolicy
      */
     public function viewAny(?User $user)
     {
-        return $user->isAbleTo("view-any-classification");
+        return true;
     }
 
     /**
@@ -33,7 +33,7 @@ class ClassificationPolicy
      */
     public function view(?User $user)
     {
-        return $user->isAbleTo("view-any-classification");
+        return true;
     }
 
     /**
--- a/api/app/Policies/ClassificationPolicy.php
+++ b/api/app/Policies/ClassificationPolicy.php
@@ -2,7 +2,6 @@
 
 namespace App\Policies;
 
-use App\Models\Classification;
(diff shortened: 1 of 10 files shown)


api/app/Policies/ClassificationPolicy.php
```diff
@@ -20,7 +19,7 @@ class ClassificationPolicy
      */
     public function viewAny(?User $user)
     {
-        return true;
+        return $user->isAbleTo("view-any-classification");
```
esizer, line 22: For each of these, user can be null so we need some null safe operators for these:

```php
  return $user?->isAbleTo("view-any-classification");
```

esizer (changes requested):
Looking great! We just need to update some of the policy methods. These updates should sort out the `e2e` tests as well.

esizer (approved):
Noice 👍

PR #6099, opened 2023-03-28
[Tests] Add cypress axe

### LINKED ISSUE #5934: Add cypress-axe for a11y testing
https://github.com/GCTC-NTGC/gc-digital-talent/blob/a71c628088590022a8760ba83d18b1ed710282ec/documentation/accessibility.md?plain=1#L41-L48

esizer:
🤖 Resolves #5934 

## 👋 Introduction

This installs `cypress-axe` and sets it up to run on the public facing pages as a start.

## 🕵️ Details

> **Note**
> You must call `cy.injectAxe` after `cy.visit` to add `axe-core` to the `window`. 

There seems to be a bit of a pattern of always doing:

1.  `cy.visit('/path/');`
2. `cy.injectAxe();`
3. `cy.checkAlly(options);`

Would it be worth it to make a command like the following:

```tsx
cy.checkAccessibility('/path/', options = {});
```

## 🧪 Testing

1. Install new deps `npm i`
2. Run cypress `npm run e2e`
3. Run the following specs:
    - Direct Intake
    - Static Pages
    - Support Page
4. Confirm the "should have no accessibility errors" passes

## 📸 Screenshot

Screenshot 2023-03-28 at 12 33 32 PM


documentation/accessibility.md
```diff
@@ -40,7 +40,7 @@ In the following example, we have a `<nav>` element where we capture an `onKeyDo
 
 ### `cypress-axe`
 
-While it has yet to be implemented, we will be adding a11y to our E2E testing with [`cypress-axe`](https://www.npmjs.com/package/cypress-axe). In the future, when writing tests your cypress tests, it would be considered best practices to add `cy.checkA11y()` when changing the state of the application. Some examples (non-exhaustive) of when to check a11y with cypress are:
+We are doing automated a11y testing in our E2E testing with [`cypress-axe`](https://www.npmjs.com/package/cypress-axe). When writing tests your cypress tests, it is considered best practices to add `cy.checkA11y()` when changing the state of the application. Some examples (non-exhaustive) of when to check a11y with cypress are:
```
mnigh, line 43:
```suggestion
We are doing automated a11y testing in our E2E testing with [`cypress-axe`](https://www.npmjs.com/package/cypress-axe). When writing tests your cypress tests, it is considered best practice to add `cy.checkA11y()` when changing the state of the application. Some examples (non-exhaustive) of when to check a11y with cypress are:
```

esizer:
> it's failing in CI but not marking the spec as failed

So, thanks to @patcon I know what happened here. The test was flaky (we needed to wait until the doc loaded before we could check a11y). What happened was:

1. Visit page
2. Check a11y
3. Oh no, no `<h1>`! 😱 
4. Document loaded
5. Let's retry that a11y just to be sure
6. Oh, there is the `<h1>`, ✅ 

I have updated the test to always check for the `<h1>` (so we know the page is loaded) before checking a11y so it is more consistent. [make browse pools a11y check more reliable](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6099/commits/24b887c2a8e388b3989f199ae8ddfa0600e48bbd)

vd1992 (approved):
runs locally, and the source of the matter appears resolved, nice!
asterisk taken care of again https://github.com/GCTC-NTGC/gc-digital-talent/pull/6099/commits/e8c7e3cc8cc4c0febe51871f82d3b627b9d46d6d

PR #6102, opened 2023-03-29
Debt - 5859 - remove default user in phpunit tests

### LINKED ISSUE #5859: ♻️ Remove default user from phpunit tests
## ♻️ Debt/Refactor

Remove the default admin user which acts as "logged in" by default during all phpunit tests. Instead, individual tests should use `actingAs` if necessary.

## 🕵️ Details

With our implementation of a default user, we can use `actingAs` to override the default and act as another user if we want. However, its impossible to act as a "null" user, meaning we can't write proper tests for guest users. We're currently using `actingAs($guestUser)` as a workaround, but that's not really accurate. Therefore, we should remove the default user, add `actingAs` anywhere its required, and then remove `actingAs` anywhere we want to act as a guest user.

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] There is no default user for phpunit tests
- [ ] We never use `actingAs($guestUser)`. Instead, we just leave off the actingAs.
- [ ] All tests are passing.

JamesHuf:
🤖 Resolves #5859.

## 👋 Introduction

Updates the phpunit tests to no longer use a default user and instead run queries with actingAs when required.

## 🧪 Testing

1. Confirm the phpunit tests all still pass.
2. Look through the tests and confirm they no longer use a default user.


api/tests/Feature/SnapshotTest.php
```diff
@@ -61,7 +53,7 @@ public function testCreateSnapshot()
 
         // get the just-created snapshot
         $actualSnapshot = $this->graphQL(
-            /** @lang Graphql */
+            /** @lang GraphQL */
             '
             query getSnapshot($poolCandidateId:ID!) {
                 poolCandidate(id: $poolCandidateId) {
```
esizer, line 59: How is this passing? 🤔 I don't see a user being passed in 😅

esizer (approved):
Awesome! Love the fix for graphql syntax highlighting too 😁

PR #6111, opened 2023-03-29
[Bug] Fix user role from removing email

### LINKED ISSUE #6103: Edit user role table changes null out the user's email
## 🐛 Bug

The role tables on the edit user page use the same mutation from the edit user page itself.  However, this mutation is set to null out empty email addresses.  Since the role dialogs do not provide an email address they null out the email field for the user.

## 🦋 Expected Behaviour

Role edits should not affect the user's email address

## 🕵️ Details

It seems like this behaviour was added in https://github.com/GCTC-NTGC/gc-digital-talent/pull/4103.

## 📋 Steps to Reproduce

1. Navigate to http://localhost:8000/en/admin/users/{userId}/edit
2. Ensure the user has an email address in their profile
3. Make any role change
4. Observe that the email is gone

## 📸 Screenshot

## 🙋‍♀️ Proposed Solution

Perhaps the page should pass a separate mutation to the role tables?  Or the role tables should manage their own mutations?

## ✅ Acceptance Criteria

- [ ] Role changes no longer affect the email address
- [ ] Whatever PR #4103 fixed is not regressed

esizer:
🤖 Resolves #6103 

## 👋 Introduction

This prevents role updates from overriding the user email to `null`.

## 🕵️ Details

We were setting the users email to `null` when it did not exist (either undefined or empty string) to prevent our validation from seeing `""` as a duplicate email. This unsets the email key in our mutation variables if it is undefined so that it is only nulled when it is an empty string.

## 🧪 Testing

1. Build the application `npm run build`
2. Navigate to `/admin/users/{userId}/edit
3. **Add** a role, confirm email is still present
4. **Remove** a role, confirm email is still present
5. **Empty** out the email, confirm it saves as empty
6. Attempt to update to an **existing** email, confirm it fails with nice error message
7. Attempt to update to a **unique** email, confirm it succeeds

### DIFF
--- a/apps/web/src/pages/Users/UpdateUserPage/UpdateUserPage.tsx
+++ b/apps/web/src/pages/Users/UpdateUserPage/UpdateUserPage.tsx
@@ -311,7 +311,9 @@ const UpdateUserPage = () => {
       id,
       user: {
         id,
-        email: emptyToNull(data.email),
+        // Do not include email in the request if it is not part of form data
+        // to prevent accidentally setting it to null
+        email: data.email !== undefined ? emptyToNull(data.email) : undefined,
         ...pick(data, [
           "firstName",
           "lastName",


petertgiles (approved):
Perfection!  Nice simple solution.  :dart:

PR #6173, opened 2023-04-04
Feature - 5941 - Tree view

### LINKED ISSUE #5941: ⚙️  Build Tree view component
# ⚙️ Summary
A tree view widget presents a hierarchical list.

**For all layout and spacing guidance, including how this component should respond on smaller devices, please see the Figma file.**
### [[Visit the Figma file](https://www.figma.com/file/BXYVsu1fo9LS2IySJDUINY/Applicants---Application?node-id=653%3A63152&t=1mqKGZguvjMWyJZF-0)]

# Initial Mode 
image

# Display Mode   
image

# 🛠️ Properties

## Root Element
Text element to be displayed on the root level 

## Error Text
Error text to be displayed on the root level 

## Children
List of card components that to be displayed on the leaf level 

## 🌎 Localization

English Copy 

> This required skill must have at least 1 résumé experience associated with it.

French Copy

[Place the french copy here] 

English Copy 

> Connect a résumé  experience

French Copy
[Place the french copy here] 

## ✅ Acceptance Criteria

- [ ] Component should accept a generic root component which is displayed first
- [ ]  Child components are displayed in a list below the root component, with some space on the left
- [ ]  A line extends down from the left end of the root component down as far as the last child component
- [ ]  The line extends across to child components. This branch can be switched off for individual children
- [ ]  Work with @substrae on getting the lines CSS 

## ✅ Accessibility checklist
- [ ] Tap targets are adequate on smaller screens
- [ ] Color contrast for all modes and states has been checked and cleared
- [ ] Zoom has been considered as layout changes occur
- [ ] `alt` text has been identified and labelled
- [ ] Non-standard states or interactions have been described and provided accessible design context

yonikid15:
🤖 Resolves #5941 

## 👋 Introduction

- Adds a new `<TreeView />` component which presents a hierarchical list. The component was created is such a way where any basic component can be added to the tree and the styling will branch accordingly. 
- Also, this adds a `<ExperienceTreeView />` component which uses the `<TreeView />`, but is tailored to create the views below:
  - Initial Mode (null): It's the applicants first time viewing the Skill Requirements section OR they have not added any experiences.
  - Display Mode: The applicant has added attached an experience to the skill.

# Initial Mode 
image

# Display Mode   
image

## 🕵️ Details

- [ ] Play around with screen sizes to ensure the vertical and horizontal tree lines are positioned correctly. Nested tree views are still not perfect though. If you have any thoughts let me know. I'm currently trying to re-purpose this [example](https://iamkate.com/code/tree-views/).
- [ ] Also, it's waiting on translations.

## 🧪 Testing

1. Go to storybook and open `Components/TreeView` and view all the stories inside.


esizer (changes requested):
This is looking great so far! But, I might be about to ruin your day, sorry 😅 

I have some suggestions but I also have a question. The branch looks like it is connecting at a specific point and that looks a little odd for the button since it lands _almost_ in the centre but not quite. If it is a fixed point near the top, would it be possible to fix that point so it lines up with the middle of the button? Not really an issue with larger elements, just the single line ones.

packages/ui/src/components/TreeView/TreeView.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+const Root = ({
+  title,
+  subtitle,
+  error,
+  children,
+  ...rest
+}: TreeViewRootProps) => (
+  <ul className="tree" {...rest}>
```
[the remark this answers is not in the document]
yonikid15: Ya I guess it's not technically a list in our case. I removed the list element format.

.vscode/project-words.txt
```diff
@@ -58,6 +58,7 @@ recruitmentimit
 recruitments
 recrutementgiti
 resumé
```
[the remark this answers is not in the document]
yonikid15: Ya that makes sense! I removed it from the list 👍.

packages/ui/src/components/TreeView/TreeView.tsx
```diff
@@ -0,0 +1,65 @@
...
+const Item = ({ noBranch, children, ...rest }: TreeViewItemProps) => (
+  <div
+    className="tree-view-item"
+    data-h2-margin-left="base(x2)"
+    data-h2-padding-left="base(x1)"
+    data-h2-padding-top="base(x1)"
+    data-h2-border-left="base:selectors[:not(:last-child)](1px solid gray.light)"
+    data-h2-content="base:selectors[::before]('')"
+    data-h2-display="base:selectors[::before](block)"
+    data-h2-position="base:selectors[::before](absolute)"
```
esizer, line 20: You can double up properties here to remove the css completelty. This should work:

```suggestion
    data-h2-position="base(relative) base:selectors[::before](absolute)"
```

packages/ui/src/components/TreeView/TreeView.tsx
```diff
@@ -0,0 +1,65 @@
...
+import "./tree-view.css";
+
+interface TreeViewItemProps {
+  noBranch?: boolean;
+  children: React.ReactNode;
+}
+
+const Item = ({ noBranch, children, ...rest }: TreeViewItemProps) => (
+  <div
+    className="tree-view-item"
```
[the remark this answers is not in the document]
yonikid15: Yes that was the last thing I was just fixing up!

apps/web/src/components/ExperienceTreeView/ExperienceTreeView.tsx
```diff
@@ -0,0 +1,64 @@
...
+        </TreeView.Item>
+      )}
+      {experiences.length > 0 &&
+        experiences.map((experience) => (
+          <TreeView.Item key={experience.id}>
+            <Accordion.Root
+              type="single"
+              collapsible
+              data-h2-margin="base(0, 0)"
+            >
```
esizer, line 45: Ugh, sorry but I just noticed this. We should avoid adding margin since it will overwrite required styles, in this case it is causing problems with the margin making the border show a gap. You could wrap this in a `div` with negative margin to fix that issue.

```tsx
<div data-h2-margin="base(-x.5, 0)">
  <Accordion />
</div>
```
Screenshot 2023-04-11 153600

packages/ui/src/components/TreeView/TreeView.tsx
```diff
@@ -0,0 +1,64 @@
...
+interface TreeViewRootProps {
+  title: string;
+  subtitle?: React.ReactNode;
+  children: React.ReactNode;
+}
+
+const Root = ({ title, subtitle, children, ...rest }: TreeViewRootProps) => (
+  <div {...rest}>
+    {/* TODO: Should the card be kept in the tree view component or manually added? */}
+    <Card title={title} color="white" bold>
```
esizer, line 40: Just my opinion, but I'd prefer to just pass this in with the children. That way we don't **_have_** to use a `Card` and have more control over the content here.

If we are sticking with a `Card` by default here could we at least drill some `CardProps` here to have more control over it and remove the TO DO?

```tsx
interface TreeViewRootProps {
  title: string;
  subtitle?: React.ReactNode;
  children: React.ReactNode;
  cardProps?: Omit<CardProps, "title">
}

<Card title={title} color="white" bold {...cardProps}>
```

esizer (changes requested):
Sorry, I just caught some stuff on the third time through reviewing 😢

packages/ui/src/components/TreeView/TreeView.tsx
yonikid15: Sounds good I removed the Card so we can have more control over what gets added [bebc221](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6173/commits/bebc22173522ed79d6fb405a6a90d58aa28ca170)

apps/web/src/components/ExperienceTreeView/ExperienceTreeView.tsx
yonikid15: Thanks for that! I fixed it in [bebc221](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6173/commits/bebc22173522ed79d6fb405a6a90d58aa28ca170)

esizer (approved):
LGTM! 👍

PR #6199, opened 2023-04-05
Bug - 6194 - limit cache of config.js

### LINKED ISSUE #6194: 🐛 When feature flags change, user doesn't see changes until hard refresh
## 🐛 Bug

About 18 hours after UAT was updated with new feature flags - ONGOING_RECRUITMENT set to true - Ongoing Recruitments were still not visible for a user. After being instructed to hard-refresh the page, by holding CTRL while clicking refresh, the Ongoing Recruitments section appeared.

## 🦋 Expected Behaviour

When feature flags change, users should see new behaviour within a reasonable time, without having to know to hard refresh.

## 🕵️ Details

- **Operating System**: Windows (TBS Surface Table)
- **Browser**: Edge

## 📋 Steps to Reproduce

1. On dev server or UAT, change a feature flag in azure portal configuration
2. Go to the site, looking for the behaviour you expect to have changed. If you open the inspector, make sure to uncheck "disable caching".

## 📸 Screenshot

## 🙋‍♀️ Proposed Solution

Some kind of cache-busting on config.js in the post deployment script? Or set a no-cache header on this file?

## ✅ Acceptance Criteria

- [ ] Within, lets say, an hour of the feature flags being changed, users should see the new behaviour, without having to hard refresh.

vd1992:
🤖 Resolves #6194 

## 👋 Introduction

Solve the issue of out of date config files by never storing it. `no-cache` doesn't mean no cache, but rather check for updates before serving something from cache. Given the config file should never be large, I felt it was best to just always grab fresh rather than run through the validation and hope it works always.

## 🕵️ Details

Also mirrored things in the local file so the process works identically to deployed. 

## 🧪 Testing

Locally 
1. docker down then up
2. load the homepage, check network tab, observe that config.js is always downloaded upon refresh

Deployed
1. I don't know how this process works

## 📸 Screenshot

image


tristan-orourke (changes requested):
Okay, I've been some [guides](https://www.howtogeek.com/devops/how-to-configure-cache-control-headers-in-nginx/) on [cache controls](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control), and I think we can do better than no-store. This file is small, sure, but there's no _danger_ in cacheing it, we just don't want it cached forever. We could even set it to `public`, allowing CDNs to cache it, as long as its invalidated frequently enough. I propose we use `Cache-Control: public, max-age=60` for now.

vd1992:
> Okay, I've been some [guides](https://www.howtogeek.com/devops/how-to-configure-cache-control-headers-in-nginx/) on [cache controls](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control), and I think we can do better than no-store. This file is small, sure, but there's no _danger_ in cacheing it, we just don't want it cached forever. We could even set it to `public`, allowing CDNs to cache it, as long as its invalidated frequently enough. I propose we use `Cache-Control: public, max-age=60` for now.

Sure https://github.com/GCTC-NTGC/gc-digital-talent/pull/6199/commits/30acdf36898a746e28de12cd77b7bb76ddb7b90c

patcon:
This is great. To share another option, can alias any regex match for `/config.*.js` to the actual unhashed file, and generate that hash based on something that will def change when a new toggle is deployed. Any hash values gets directed to the file, but regular caching happens until the hash changes

vd1992:
aye aye https://github.com/GCTC-NTGC/gc-digital-talent/pull/6199/commits/ed3a8364bd5df1975e4369d25f8e7ea62e2d5594

tristan-orourke (approved):
Worked on the dev server! Good to go! 👍

PR #6255, opened 2023-04-12
Bug - 6081 - toast error for authorization

### LINKED ISSUE #6081: 🐛 Toast should inform user of actions they aren't allowed to do
## 🐛 Bug

Some actions users aren't allowed to do, and this results in an error. But the error toast does not inform you that you aren't allowed to do so, which makes it hard to comprehend what the issue is.

It would be better to prevent the user form firing these actions but that is out of scope for this issue.

## 🕵️ Details

Authorization stuff stems from policy files, but we do output error messages from the back-end in `apiMessages.ts`
Console has the authorization error logged. 

## 📋 Steps to Reproduce

1. attempt to do an action you aren't authorized to do (like update pools as platform admins)
2. observe toast

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

Toast alerts user of authorization issues.

## ✅ Acceptance Criteria

- [ ] when an action fails on the frontend due to policy, the user understands that something isn't broken, just that they aren't allowed to do that

vd1992:
🤖 Resolves #6081 

## 👋 Introduction

Toast message when an action fails due to authorization/policy. 

## 🕵️ Details

Translation in. 

## 🧪 Testing

1. Head to someplace to trigger a policy error
2. Example, attempt editing a pool as platform@test.com
3. Observe toast

## 📸 Screenshot

image

image

### DIFF
--- a/packages/client/src/components/ClientProvider/ClientProvider.tsx
+++ b/packages/client/src/components/ClientProvider/ClientProvider.tsx
@@ -23,8 +23,10 @@ import { toast } from "@gc-digital-talent/toast";
 import {
   buildRateLimitErrorMessageNode,
   buildValidationErrorMessageNode,
+  buildAuthorizationErrorMessageNode,
   extractRateLimitErrorMessages,
   extractValidationErrorMessages,
+  extractAuthorizationErrorMessages,
 } from "../../utils/errors";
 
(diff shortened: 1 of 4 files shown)


petertgiles (approved):
Nice!

PR #6283, opened 2023-04-13
Remove titles from templates

petertgiles:
## 👋 Introduction

This branch removes the placeholder issue titles from the templates.

## 🕵️ Details

We tend to get many issues on our board with very long titles.  For example, many titles start with ":keyboard:  Accessibility Issue - " but it's already clear that it's an issue with the accessibility label. I'm hoping that by removing the placeholder titles, submitters will not use them as prefixes and be encouraged to write short descriptive titles.


vd1992 (approved):
emojis preserved 👍

PR #6291, opened 2023-04-13
Dependencies - 6126 - update to Laratrust v8

### LINKED ISSUE #6126: 🛠️ Dependency - Update Laratrust from 7 to 8
## 🛠️ Tooling

Breaking changes with Laratrust 8.x that need to be addressed, update the dependency manually. 
https://github.com/GCTC-NTGC/gc-digital-talent/pull/6116

## 🕵️ Details

Consult the upgrading to 8.x guide here
https://laratrust.santigarcor.me/docs/8.x/upgrade.html

## ✅ Acceptance Criteria

- [ ] Laratrust updated to 8.x

vd1992:
🤖 Resolves #6126 

## 👋 Introduction

Updates Laratrust to v8 manually. 

## 🕵️ Details

Followed the guide for the most part. https://laratrust.santigarcor.me/docs/8.x/upgrade.html

Tricky bit was that the `UpdateUserAsAdmin` broke for updating `RoleAssignments`. My guess was that using `roles` as the field name didn't play nice with Laratrust the package, so I renamed it to something simple to represent it is calling a setter. 
The unit test kept failing as the mutation would just do nothing until I renamed from roles -> rolesSetter. 

Not the best solution, but I had been playing around things for a while now and finally seemed to get it to work again. 

## 🧪 Testing

1. Install
2. Bounce around /admin and ensure roles and permissions stuff works (can add and remove roles)
3. Tests pass


api/graphql/schema.graphql
```diff
@@ -1225,7 +1225,7 @@ input UpdateUserAsAdminInput
   educationExperiences: EducationExperienceHasMany
   awardExperiences: AwardExperienceHasMany
 
-  roles: RoleAssignmentHasMany
+  rolesSetter: RoleAssignmentHasMany
```
petertgiles, line 1228: Hmm, yeah this is a bit awkward.  It has bothered me ever since I set this up that this property was not called `roleAssignments` like it is on the User type.  Could you try that?  Or if you have to go with a non-standard method on the model could you use the rename directive to give it a nice name in the schema?

petertgiles (changes requested):
Seems to work great!  I left a few small things to check on.

api/graphql/schema.graphql
vd1992: done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/6291/commits/32dcf653157d2fc3572e45ea2aa50a2d038f92c5

vd1992:
Did some investigation regarding the Cypress test failure, repeatable locally. 
Failure happens when attempting to update user info, like phone numbers, and not the roles stuff. 
This appears due to the mutation/something in the frontend attempting to send `User.roleAssignments` to `UpdateUserAsAdmin.roleAssignments`

I think attempting to reuse terms we've already defined elsewhere just causes unforeseen issues. As both are fields named `roleAssignments` but very different type-wise. One is type `[RoleAssignment]` and the other `RoleAssignmentHasMany`

Pick a unique name Peter 😛 

image

vd1992:
There we go, checks passing
All looks good to still you Peter?

petertgiles:
Looks good.  Make it so.  :rocket:

PR #6322, opened 2023-04-19
Test - 6222 - user info page e2e tests

### LINKED ISSUE #6222: 🧪 Cypress tests for Admin "User Information" page
## 🧪 Tests

Test the behaviour of the Admin "User Information" page with different user roles.

## 🕵️ Details

Test that the page loads (or doesn't), and loads the user's applications in a table without errors.

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] If an applicant user is logged in, page shows the non-authorized warning
- [ ] If Pool Operator is logged in, the user's PoolCandidates to their teams pools are displayed
- [ ] If a Request Responder is logged in, all the user's PoolCandidates are displayed
- [ ] If a platform admin is logged in, all the user's PoolCandidates are displayed

## 🛑 Blockers

Blocked by {issue numbers}

yonikid15:
🤖 Resolves #6222 

## 👋 Introduction

- Tests the behaviour of the Admin "User Information" page with different user roles.
- Tests that the page loads (or doesn't), and loads the user's applications in a table without errors.

## 🕵️ Details

- [x] If an applicant user is logged in, page shows the non-authorized warning
- [x] If Pool Operator is logged in, the user's PoolCandidates to their teams pools are displayed
- [x] If a Request Responder is logged in, all the user's PoolCandidates are displayed
- [x] If a platform admin is logged in, all the user's PoolCandidates are displayed

## 🧪 Testing

1. All tests are in `.../cypress/e2e/admin/user-info.cy.js`


apps/e2e/cypress/e2e/admin/user-info.cy.js
```diff
@@ -0,0 +1,258 @@
...
+    cy.visit("/en/admin");
+
+    cy.findByRole("heading", {
+      name: /Sorry, you are not authorized to view this page./i,
+    })
+      .should("exist")
+      .and("be.visible");
+  });
+
+  it("Should show the pool operator a candidate's user information page on their team", () => {
```
petertgiles, line 124: The assertions on these tests seem a bit weak to me.  You're making sure the page loads (or doesn't) but it doesn't seem like you're asserting that certain pool candidates are visible or not.  For example, on this test I'd expect that you can see candidates that belong to the pool of the current pool operator but no candidates of other pools.

That type of assertion seems more appropriate (and easier to setup!) in a phpunit test though...  @tristan-orourke?

petertgiles (changes requested):
Awesome!  One small thing and one big thing to check on.  :smile:

apps/e2e/cypress/e2e/admin/user-info.cy.js
tristan-orourke: Hmm. I see your point @petertgiles that testing for exactly which pool candidates are returned by the query would be easier to do in a phpunit test. We should write that phpunit test if we haven't yet.

I would still like to see this cypress test - it can be simpler than my original acceptance criteria, but smarter than it is now. I would like it to create an applicant with applications to two different pools, and then a pool operator in one of those pools is able to view the profile without seeing a "GraphQL error".

tristan-orourke (changes requested):
We've decided that we want to make our cypress tests as atomic as possible. To keep them simpler and shorter. Basically, that means trying to test one thing at a time, and do as few clicks or interactions as possible.

For example, you don't need to navigate to admin, click on "pools", click on "view candidates", click on one candidate, and then test the page. You can just seed the specific user you want to look at with an api call, then navigate straight to that user's page, and test that the page loads correctly. No clicking necessary. It should run much faster.

@patcon led the push for more atomic feature tests, you can ask him for advice if you want.

apps/e2e/cypress/e2e/admin/user-info.cy.js
```diff
@@ -0,0 +1,258 @@
...
+
+    cy.findByRole("heading", { name: /pools/i })
+      .should("exist")
+      .and("be.visible");
+
+    cy.url().should("contain", "/admin/pools");
+
+    cy.wait("@gqlgetMePoolsQuery");
+
+    cy.findByRole("textbox", { name: /search/i })
```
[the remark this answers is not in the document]
yonikid15: Okay linting warnings should be gone now

apps/e2e/cypress/e2e/admin/user-info.cy.js
yonikid15: Alright, I simplified the testing by directly visiting the users info page. I stuck with one user throughout the testing:

- First testing viewing the user info page without creating an application. 
- Then tested viewing the user info page after applying to a pool that's owned by the pool operator. 
- Lastly, tested viewing the user info page after user applies to another pool not synced the pool operator.

tristan-orourke (changes requested):
I really like how you've written the test now! Just a couple suggestions to speed it up slightly.

Right now you run the assertions 3 times - before applying, after applying to one pool, then after applying to a second. The assertions should have the same results after the first and second application, so I think its okay to just run the assertions twice: once before applying, once after applying to both pools.

Similarly, I think you can easily get away with just testing the applicant role once, since it shouldn't be affected by the applications.

yonikid15:
> I really like how you've written the test now! Just a couple suggestions to speed it up slightly.
> 
> Right now you run the assertions 3 times - before applying, after applying to one pool, then after applying to a second. The assertions should have the same results after the first and second application, so I think its okay to just run the assertions twice: once before applying, once after applying to both pools.
> 
> Similarly, I think you can easily get away with just testing the applicant role once, since it shouldn't be affected by the applications.

Okay I ran assertions once before applying and once after applying to both pools. 👍 
And removed additional testing with applicant role.

tristan-orourke (approved):
👍

Issue #6040, opened 2023-03-22
✨ Application Revamp: Résumé page


tristan-orourke:
## ✨ Feature

Review your Resume step of the application, where experiences are shown but NOT skills. Slight changes to experience accordions. Null state if no experiences yet.

## 🕵️ Details

This lists your experiences. The content of the accordions is a little different than existing ones on our profile. Each shows something specific to the experience type, and then "Tasks and Responsibilities", which is the experience "details" in the API.

DO NOT show the "How I used x skill" section. Skills should not appear on this page at all.

Experiences can be sorted by date or by type.

## 🎨 Design File

https://www.figma.com/file/BXYVsu1fo9LS2IySJDUINY/Applicants---Job-application?node-id=636-59839&t=uYeh8QDrmOQsNpJQ-0

## 📸 Screenshot

### First time user (ie if now experiences exist yet)
![Step 3 - Resume review (first time user)](https://user-images.githubusercontent.com/4313717/227048278-a59efed6-404a-46c5-ad7a-4748f9540a9d.png)

### Review
![Step 3 - Resume review (complete)](https://user-images.githubusercontent.com/4313717/227048190-fc4f44ec-70d8-43d4-bcd2-2d14dd3125b0.png)

### Work accordion
![image](https://user-images.githubusercontent.com/4313717/227049805-7ffe4d99-085e-4471-bdc5-ce1e4f4fe116.png)

### Education accordion
![image](https://user-images.githubusercontent.com/4313717/227049838-f713f039-7e1e-4325-9064-f0a9951b49c5.png)

### Community accordion
![image](https://user-images.githubusercontent.com/4313717/227049900-f08e661d-f074-41ec-bb91-a4d6735d3e31.png)

### Personal accordion
![image](https://user-images.githubusercontent.com/4313717/227049969-e2fd0407-4206-4a0a-8fce-13fb18a5217d.png)

### Award accordion
![image](https://user-images.githubusercontent.com/4313717/227050015-d77c9cca-02ce-4bf3-94b9-c9b062e88c0f.png)

## 🙋‍♀️ Proposed Implementation

(optional) Do you have a proposed implementation?

## 🌎 Localization

| EN                                                                                                                                                                                                                                                                                                                                                         | FR                                                                                                                                                                                                                                                                                                                                                                                                                              | Description                                                                                                      |
[...]

PR #6384, opened 2023-04-25
Accordion refactor

petertgiles:
Related to #6040 but does not close it.

## 👋 Introduction

This branch refactors the accordion component to make it possible to created headers needed for #6040 

## 🕵️ Details

- moves the experience accordions out of User Profile to the components directory
- Factors out the current accordion header as a StandardHeader component to maintain current appearance
  - renamed prop from `headerAs` to `headingAs` to reduce confusion :smile: 
- Separates the Accordion Header from the Accordion Trigger so that they can be used independently
- New Accordion Storybook story `WithEditLink` with what will eventually become the new ExperienceHeader

## 🧪 Testing

1. Rebuild app
2. Check that experience accordions look good
3. Check that non-experience accordions look good
  - both regular and simple styles
4. Check Chromatic diffs
5. Check the new Accordion/WithEditLink story

## 📸 Screenshot

image


packages/ui/src/components/Accordion/Accordion.tsx
```diff
@@ -48,6 +48,26 @@ const StyledHeader = React.forwardRef<
...
+}
+
+const Header = React.forwardRef<
+  React.ElementRef<typeof AccordionPrimitive.Header>,
+  AccordionHeaderProps
+>(({ headingAs = "h2", children, ...rest }, forwardedRef) => {
+  const Heading = headingAs;
+  return (
+    <StyledHeader asChild ref={forwardedRef} {...rest}>
+      <Heading data-h2-line-height="base(1)">{children}</Heading>
```
esizer, line 66: We will need to pass `headingAs` to this underlying component. Right now everything shows up at `h2` regardless of what is passed in.

packages/ui/src/components/Accordion/Accordion.stories.tsx
```diff
@@ -35,30 +36,93 @@ const Template: ComponentStory<typeof Accordion.Root> = ({
...
+                data-h2-flex-direction="base(column)"
+                data-h2-gap="base(x.5 0)"
+              >
+                <span
+                  data-h2-display="base(block)"
+                  data-h2-font-size="base(h6, 1)"
+                >
+                  <span data-h2-font-weight="base(700)">Custom</span>
+                  {" Title"}
+                </span>
```
esizer, line 92: I understand this is just a story for examples sake, but an [accordion trigger should always contain an element with `role=heading`](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/#:~:text=Each%20accordion%20header%20button%20is%20wrapped%20in%20an%20element%20with%20role%20heading%20that%20has%20a%20value%20set%20for%20aria%2Dlevel%20that%20is%20appropriate%20for%20the%20information%20architecture%20of%20the%20page.).

petertgiles: Yeah, this is helpful.  It is more than an example - I plan to use this for the new experience accordions.

packages/ui/src/components/Accordion/Accordion.tsx
petertgiles: Are you sure?  I popped an h3 into Storybook and it seemed to work OK.  Am I missing something?   :sweat_smile: 
image

esizer: Wow, I'm sorry. I completely missed line 64 there and was looking at the wrong story! 🤦‍♀️ 😛

packages/ui/src/components/Accordion/Accordion.stories.tsx
petertgiles: The radix header adds this role itself, doesn't it?
image

petertgiles:
Alright, I feel pretty good about this Chromatic diff.  I think it's done!  Famous last words....  :sweat_smile:

packages/ui/src/components/Accordion/StandardHeader.tsx
```diff
@@ -0,0 +1,67 @@
+import { Accordion } from "@gc-digital-talent/ui";
+import { AccordionHeaderProps } from "@gc-digital-talent/ui/src/components/Accordion/Accordion";
```
esizer, line 2: We will need to update this import path so the package isn't importing itself.

```tsx
import { Accordion } from "./Accordion";
```

Though, i'm wondering if it is worth just moving this to `Accordion.tsx` and exporting it as part of the `Accordion` object? So we could do: `<Accordion.StandardHeader>`. But that can be saved for #6412

esizer (approved):
LGTM once those imports get updated! 😁

packages/ui/src/components/Accordion/StandardHeader.tsx
petertgiles: Good catch!  I've fixed the imports in [fix imports](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6384/commits/9c2a04f5e52f8a7437946ebc7664299e4f79ecca).

Yeah, I'll consider exporting the StandardHeader in the Accordion in the main PR, after seeing how the new ones integrates.   I'm not quite comfortable with the name though.  Is there a word for a header and trigger together?  I don't want to call it StandardHeaderAndTrigger.  :laughing:

esizer: heh, yeah... That is exactly what I was thinking. We could just make it obvious and say `Accordion.HeaderTrigger` 😅

PR #6415, opened 2023-04-26
[Bug] Fix suspended application test

### LINKED ISSUE #6381: 🐛 `ApplicantTest` `testCountApplicantsQuerySuspended` failing intermittently
## 🐛 Bug

PHPUnit `ApplicantTest` `testCountApplicantsQuerySuspended` fails intermittently.

## 🕵️ Details

```shell
FAILED  ApplicantTest > count applicants query suspended                     
  Unable to find JSON: 

[{
    "data": {
        "countApplicants": 5
    }
}]

within response JSON:

[{
    "data": {
        "countApplicants": 6
    }
}].


Failed asserting that an array has the subset Array &0 (
    'data' => Array &1 (
        'countApplicants' => 5
    )
).
   array (
     'data' => 
     array (
  -    'countApplicants' => 5,
  +    'countApplicants' => 6,
     ),
   )
  

  at tests/Feature/ApplicantTest.php:1009
    1005▕                 ]
    1006▕             ]
    1007▕         )->assertJson([
    1008▕             'data' => [
  ➜ 1009▕                 'countApplicants' => 5
    1010▕             ]
    1011▕         ]);
    1012▕     }
    1013▕

```

## 📸 Screenshots

Screen Shot 2023-04-25 at 09 35 50

Screen Shot 2023-04-25 at 09 15 58

## 📋 Steps to Reproduce

1. Run 20 times `for i in {1..20}; do docker-compose exec -w /home/site/wwwroot/api webserver sh -c "./vendor/bin/phpunit ./tests/Feature/ApplicantTest.php"; done`
2. Observe failing test

- [ ] `ApplicantTest``testCountApplicantsQuerySuspended` passes 20 out of 20 times

esizer:
🤖 Resolves #6381 

## 👋 Introduction

Fixes the `ApplicantTest::testCountApplicantsQuerySuspended` making it pass consistently.

## 🕵️ Details

We introduced a new `suspended_at` column on the `PoolCandidate` model to determine if users are available instead of the deprecated `job_looking_status` on the `User` model but never updated the test to reflect that change.

> **Note**
> There are some formatting changes as a result of the new format on save. Sorry if it complicates the diff 😬 

## 🧪 Testing

1. Run `for i in {1..20}; do docker-compose exec -w /home/site/wwwroot/api webserver sh -c "./vendor/bin/phpunit ./tests/Feature/ApplicantTest.php"; done`
2. Go make a ☕ or 🍵 
3. Come back and ensure there were no failures

## 📸 Screenshot
Screenshot 2023-04-26 133547

### DIFF
--- a/api/tests/Feature/ApplicantTest.php
+++ b/api/tests/Feature/ApplicantTest.php
@@ -978,12 +978,14 @@ class ApplicantTest extends TestCase
         ]);
         PoolCandidate::factory()->count(5)->availableInSearch()->create([
             'pool_id' => $pool1,
+            'suspended_at' => null,
             'user_id' => User::factory([
                 'job_looking_status' => ApiEnums::USER_STATUS_ACTIVELY_LOOKING,
             ])
         ]);
         PoolCandidate::factory()->count(4)->suspended()->create([
             'pool_id' => $pool1,
+            'suspended_at' => config('constants.past_date'),
             'user_id' => User::factory([
                 'job_looking_status' => ApiEnums::USER_STATUS_ACTIVELY_LOOKING,
             ])
@@ -1044,16 +1046,18 @@ class ApplicantTest extends TestCase
         // Assert candidate one returns 10
         $this->actingAs($this->adminUser, "api")
             ->graphQL(
-            /** @lang GraphQL */
-            '
+                /** @lang GraphQL */
+                '
             query applicant($id: UUID!) {
                 applicant(id: $id) {
                     priorityWeight
                 }
             }
-            ', [
-                'id' => $candidateOne->id,
-            ])->assertJson([
+            ',
+                [
+                    'id' => $candidateOne->id,
+                ]
+            )->assertJson([
                 "data" => [
                     "applicant" => [
                         "priorityWeight" => 10,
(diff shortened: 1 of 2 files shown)


api/tests/Feature/ApplicantTest.php
```diff
@@ -978,12 +978,14 @@ public function testCountApplicantsQuerySuspended(): void
...
             'user_id' => User::factory([
                 'job_looking_status' => ApiEnums::USER_STATUS_ACTIVELY_LOOKING,
             ])
         ]);
         PoolCandidate::factory()->count(4)->suspended()->create([
             'pool_id' => $pool1,
+            'suspended_at' => config('constants.past_date'),
             'user_id' => User::factory([
                 'job_looking_status' => ApiEnums::USER_STATUS_ACTIVELY_LOOKING,
             ])
```
vd1992, line 991: Currently running the tests. 

That is odd though, it is calling `suspended()` in the chain which does set a `suspended_at`. 😕 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/736c3c31fa2dfea262ce57d8098f25b77cc2201b/api/database/factories/PoolCandidateFactory.php#L104-L112

esizer: Hmm, I never even saw that. Could it be the `now` that was maybe causing issues? I see check `>= now` in the scope so maybe on the odd chance it generates the upper range they are equal and it is included? 🤔

vd1992: That seems like the most likely culprit in that case.

To rule out any issues with rounding we could generate two more pool candidates for the test. 
One with
`suspended_at = Carbon::now()->subMinute())`

and the other
`suspended_at = Carbon::now()->addMinute())`

The expected candidate count then would go from 5 to 6.

esizer: So, I updated the `suspended` logic to just set it to one min earlier and that also seems to work after 50 re-runs 🤷‍♀️ 

[update suspended logic on pool candidate factory](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6415/commits/2e3fe614ddccb7fc8e54aa005a026ba3243edeb2)

vd1992: weird
guess the factory was generating "now" much more frequently than you'd expect

esizer: I guess so 😬

vd1992 (approved):
Ran 20 times, all good. 
Truly as they say, "no time like the ~present~ now".

PR #6449, opened 2023-04-27
[Tooling] PHP bootstrap clear cache

### LINKED ISSUE #5989: Clear bootstrap cache on setup
## 🛠️ Tooling

The Laravel 10 upgrade was a bit painful for people because of the "Class not found" errors.  We should probably clear the bootstrap cache files on setup since people often use that script for a more thorough refresh of their environments.

## 🙋‍♀️ Proposed Solution

image

## ✅ Acceptance Criteria

- [ ] The bootstrap/cache files are cleared on setup

vd1992:
🤖 Resolves #5989 

## 👋 Introduction

Just implementing Peter's suggestion. 

## 🕵️ Details

Added it to refresh API, since it is such a small step you may want it to run more regularly. And not need to run setup just to clear the cache.

## 🧪 Testing

1. Refresh

### DIFF
--- a/maintenance/scripts/refresh_api.sh
+++ b/maintenance/scripts/refresh_api.sh
@@ -4,6 +4,7 @@ parent_path=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 source ${parent_path}/lib/common.sh
 
 cd /var/www/html/api
+rm ./bootstrap/cache/*.php
 composer install --prefer-dist
 php artisan migrate
 php artisan lighthouse:print-schema --write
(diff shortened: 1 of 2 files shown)


vd1992:
Attempting to remove it breaks setup as the directory doesn't exist yet. 
So, I think it should stay in just the refresh script. https://github.com/GCTC-NTGC/gc-digital-talent/pull/6449/commits/755e3d26b761370f19efe4b55d561e6750117578

image

petertgiles:
Good idea with refresh!  I really think it has to be in the setup script though.  `--force` should do the trick.
image

petertgiles (approved):
Thanks!  Looks great.

PR #6458, opened 2023-04-28
[Feature] feature flag education requirement check

### LINKED ISSUE #6286: ✨ Add education evidence to backend "application complete" logic
## ✨ Feature

We have some logic in the API to determine if an application is complete.  It is used for things like determining if a an application can be submitted.  The education evidence need to be a part of that.

## 🕵️ Details

This logic should be part of the APPLICATION_REVAMP feature flag so it does not get used if the flag is off.

## ✅ Acceptance Criteria

- [ ] An application without education evidence is not considered complete for the backend validation
- [ ] Turning the feature flag off disables the check

vd1992:
🤖 Resolves #6286 

## 👋 Introduction

Adds back-end validation for checking if education requirement is complete or not, contingent on the feature flag. 

## 🕵️ Details

Followed the nice format Eric set for adding further validation. 

Needed to introduce `@with` to prevent lazy loading. 
Added testing for it based off the flag, conditional too. 
Set the default for the flag to `false` to prevent it from tripping up e2e testing using the setup file.

## 🧪 Testing

1. Load things up, set flag to `false`
2. Confirm you can apply normally
3. Set flag to `true`
4. Confirm you cannot successfully submit due to API errors
5. Update via GraphiQL, then confirm you can submit normally

```
  updateApplication(id: "ID", application: $aaa) {
    status
    submittedSteps
    educationRequirementOption
    educationRequirementExperiences {
      id
    }
  }

  "aaa": {
    "educationRequirementOption": "EDUCATION",
    "educationRequirementEducationExperiences": {"sync": ["ID"]}
  }
```

## 📸 Screenshot

image


api/.env.example
```diff
@@ -94,5 +94,5 @@ GCNOTIFY_TEMPLATE_TEST_BULK_SMS="d662aac4-08bb-4917-860f-cc9999f4648d"
 # Feature flags
 FEATURE_ONGOING_RECRUITMENTS=true
 FEATURE_APPLICANT_DASHBOARD=true
-FEATURE_APPLICATION_REVAMP=true
+FEATURE_APPLICATION_REVAMP=false
```
esizer, line 97: Should we make a note to change this back to true for dev once we fix the `e2e` stuff?

api/app/Rules/HasEducationRequirement.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /**
+     * Get the validation error message.
+     *
+     * @return string
+     */
+    public function message()
+    {
+        return 'EducationRequirementIncomplete';
```
esizer, line 50: Can we make this part of the enum that handles pool candidate error codes?

https://github.com/GCTC-NTGC/gc-digital-talent/blob/9a93c576b3fa9d1f1a29e86f027d6531691b72a3/api/database/helpers/ApiEnums.php#L405-L427

esizer (changes requested):
Looking great! Couple of small suggestions. As well, I think it would serve us well to add the frontend message in [`apiMessages`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/packages/i18n/src/messages/apiMessages.ts) unless that is part of the AC for another ticket.

api/.env.example
vd1992: Yeah, sure, done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/6458/commits/f59cb6d872c147e5af07fabe7169d5fa7a08502d
Will comment in the issue itself with a code snippet later.

api/app/Rules/HasEducationRequirement.php
vd1992: Enum stuff done here https://github.com/GCTC-NTGC/gc-digital-talent/pull/6458/commits/cdcb50ab9d2d1d3811e35600c275845e372418b2
Message added too, I realized the validation error messages are probably never going to be seen since it relies on the front-end validation failing but back-ups are back-ups https://github.com/GCTC-NTGC/gc-digital-talent/pull/6458/commits/ce4549f0dfa65b36b041e9488dad007b88b89d5e

As a thrown in thing, I noticed one other PHP test was failing, looked at it and it seemed like a simple thing. Sometimes the clock would roll to the next second while it was running, so I fixed things by just comparing to the nearest minute rather than second. https://github.com/GCTC-NTGC/gc-digital-talent/pull/6458/commits/f6aefb8c02bb2c2b7b76207ec3c0322c9bd1e5c7

vd1992: And got the translation in for it quickly with https://github.com/GCTC-NTGC/gc-digital-talent/pull/6458/commits/cfcd70254a331fe13442e602ca5d635b7037dac5

esizer (approved):
Nice 🥇 !

PR #6468, opened 2023-05-02
[Feature] skill filter intersectional and additive

### LINKED ISSUE #6367: ✨ Convert skill filter in candidate results from intersection to additive
## ✨ Feature

When looking at the results of candidates inside individual request it would be helpful to know which candidates have which of the skills selected in the skill filter.

## 🕵️ Details

Change the way skill filters work instead of intersection (return only users who match ALL skills) return the summation instead. (return ANY user who matches ANY of the skills selected)

In addition to this, adding a column to the candidate results that shows users from MOST skill matches to LEAST skill matches would help DCM decide which candidates fit best to the request.

## 🎨 Design File

[Link to figma](https://www.figma.com/file/UsrB15yicgDCLDQmx3snyl/User-Flows?node-id=3045%3A4807&t=zbLzQ0dXanipnzTx-1)

## 📸 Screenshot

image

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Criteria 1
- [ ] Criteria 2

## 🛑 Blockers

Blocked by {issue numbers}

vd1992:
🤖 Resolves #6367 

## 👋 Introduction

Introduces OR matching of skills rather than just AND. 

## 🕵️ Details

Tricky bit to keep things playing nice with ApplicantFilter.php
ApplicantFilterInput.skills is the OR matching while skillsIntersectional is the other one.
Scopes skillsIntersectional for the classic, skillsAdditive is the new coke. The latter becomes the main skills scope so X.skills 

Additional scope was straightforward to add since I only had to recycle the existing mammoth and tweak it.

## 🧪 Testing

1. head to view pool candidates for some pool
2. view a user profile, note what skills they have and don't have
3. skill filter for one skill they have and don't have, observe they appear in results

As well, tests pass locally

image

image

image

image

image

image

### DIFF
--- a/api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
+++ b/api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
@@ -78,7 +78,7 @@ final class CountPoolCandidatesByPool
 
             // skills
             if (array_key_exists('skills', $filters)) {
-                User::scopeSkills($userQuery, $filters['skills']);
+                User::scopeSkillsIntersectional($userQuery, $filters['skills']);
             }
         });
 
--- a/api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
+++ b/api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
@@ -78,7 +78,7 @@ final class CountPoolCandidatesByPool
 
             // skills
             if (array_key_exists('skills', $filters)) {
-                User::scopeSkillsIntersectional($userQuery, $filters['skills']);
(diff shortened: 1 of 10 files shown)


vd1992:
It currently functions by sending an array of skills to the filter which executes `scopeSkillsAdditive` through `ApplicantFilterInput.skills`
Implementing AND filtering could be done through a toggle on the filter page that if true, nulls `ApplicantFilterInput.skills` and sends the array of skills to `ApplicantFilterInput.skillsIntersectional` instead.

tristan-orourke:
@Jerryescandon it works the same on both tables

Issue #6462, opened 2023-05-01
✨ IAP Application changes hydrogen theme


tristan-orourke:
## ✨ Feature

When on the IAP version of the application, the IAP Hydrogen theme is used.

## 🕵️ Details

This depends partly on how we solved #6461.

If the IAP application only exists below the `/indigenous-it-apprentice` url, then theming it should be easy: we simply use the IAP theme at that route and below.

Otherwise, there is an extra complication here: we are determining whether to behave as an IAP page at the level of the Application component, but the theme should affect the whole page (including headers and footers). We may need something like a clever custom hook for this.

## 🎨 Design File

[IAP Application](https://www.figma.com/file/8piNhFaBjcxssBB8LAGxEY/IAP-User-flows?node-id=269-76379&t=FM1b6mI8HeHY19YM-0)

## 📸 Screenshot

IAP Theme:
![Application introduction](https://user-images.githubusercontent.com/4313717/235505513-6769b563-d7fa-4915-994d-bec9dabd27d0.jpg)

Original Theme:
![Application introduction](https://user-images.githubusercontent.com/4313717/235505554-2107a38f-02ce-4b69-b2cf-1798733f7149.jpg)

## 🙋‍♀️ Proposed Implementation

Not yet.

## 🌎 Localization

(optional) Provide any new copy along with translations available.

## ✅ Acceptance Criteria

- [ ] When on an IAP Application page, the whole page uses the IAP Hydrogen theme
- [ ] When going to any other page or application (besides the IAP homepage) the theme switches back to base hydrogen theme
- [ ] Record issues for any problems that appear on the application pages, as result of using new theme

tristan-orourke:
useTheme hook already exists

PR #6521, opened 2023-05-05
[Feature] IAP publishing group, application theme

### LINKED ISSUE #6461: ✨ Application component can determine if it's for the IAP
## ✨ Feature

The Application process and UI changes slightly when applying to the IAP. The Application component needs some way to determine which version to show.

## 🕵️ Details

## 🎨 Design File

[Include a link to the design file (if it exists).](https://www.figma.com/file/BXYVsu1fo9LS2IySJDUINY/Applicants---Job-application?node-id=322-31431&t=B3DmGu4KMyi5po0D-0)

Several ideas:
1. Add an IAP publishing group. Switch behaviour behaviour based on the Pool's publishing group.
2. We expect there to only be one IAP pool, so theoretically we could just check if the pool matches the ID of our one IAP pool. Maybe  save this id in an environment variable?
3. Pass a flag into the Application component as a prop, and change behaviour based on the prop. This might make storybook stories easier, but might lead to some prop drilling. We could include a different version of the application, with its own routing, below the `/indigenous-it-apprentice` route, and give it this prop. Or, we could combine the prop strategy with 1) or 2) via a wrapper component.

## ✅ Acceptance Criteria

- [ ] Any step of the application can determine whether to behave as the IAP version or not

esizer:
🤖 Resolves #6461 

## 👋 Introduction

This adds the IAP publishing group and sets the theme to IAP on application pages for the new publishing group.

## 🕵️ Details

Something to note, the loading spinner may hop themes since it will fallback to the current theme until we know the publishing group.

## 🧪 Testing

> **Note**
> We currently only show IT pools so you will need to navigate directly to the job poster. The URL can be found on the admin "pool information" page

2. Create a pool, note the new IAP publishing group (use this)
3. Start an application for an IT pool
4. Confirm the new application pages use the digital talent theme
5. Start an application for an IAP pool
6. Confirm the new application pages use the IAP theme

## 📸 Screenshot

localhost_8000_en_applications_96a99aae-7b10-4e9f-a25e-ad34be43003d_welcome

### DIFF
--- a/api/database/helpers/ApiEnums.php
+++ b/api/database/helpers/ApiEnums.php
@@ -387,6 +387,7 @@ class ApiEnums
     /**
      * Publishing Groups
      */
+    const PUBLISHING_GROUP_IAP = 'IAP';
     const PUBLISHING_GROUP_IT_JOBS = 'IT_JOBS';
     const PUBLISHING_GROUP_IT_JOBS_ONGOING = 'IT_JOBS_ONGOING';
     const PUBLISHING_GROUP_EXECUTIVE_JOBS = 'EXECUTIVE_JOBS';
@@ -395,6 +396,7 @@ class ApiEnums
     public static function publishingGroups(): array
     {
         return [
+            self::PUBLISHING_GROUP_IAP,
             self::PUBLISHING_GROUP_IT_JOBS,
             self::PUBLISHING_GROUP_IT_JOBS_ONGOING,
             self::PUBLISHING_GROUP_EXECUTIVE_JOBS,
--- a/apps/web/src/pages/Applications/ApplicationLayout.tsx
+++ b/apps/web/src/pages/Applications/ApplicationLayout.tsx
@@ -8,6 +8,7 @@ import {
   Pending,
   ThrowNotFound,
 } from "@gc-digital-talent/ui";
+import { useTheme } from "@gc-digital-talent/theme";
 import { empty } from "@gc-digital-talent/helpers";
 
 import SEO from "~/components/SEO/SEO";
@@ -21,7 +22,7 @@ import {
   getFullPoolAdvertisementTitleHtml,
   getFullPoolAdvertisementTitleLabel,
 } from "~/utils/poolUtils";
-import { useGetApplicationQuery } from "~/api/generated";
+import { PublishingGroup, useGetApplicationQuery } from "~/api/generated";
 import {
   checkForDisabledPage,
   deriveSteps,
@@ -37,6 +38,7 @@ const ApplicationPageWrapper = ({ application }: ApplicationPageProps) => {
   const paths = useRoutes();
   const navigate = useNavigate();
   const { experienceId } = useParams();
+  const { setTheme, mode } = useTheme();
   const pages = getApplicationPages({
     intl,
     paths,
@@ -102,6 +104,8 @@ const ApplicationPageWrapper = ({ application }: ApplicationPageProps) => {
     application.submittedSteps,
   );
 
+  const publishingGroup = application.poolAdvertisement?.publishingGroup;
+
   // If we cannot find the current page, redirect to the first step
   // that has not been submitted yet, or the last step
   React.useEffect(() => {
@@ -111,6 +115,12 @@ const ApplicationPageWrapper = ({ application }: ApplicationPageProps) => {
       });
     }
   }, [currentPage, navigate, nextStepUrl]);
+
+  React.useEffect(() => {
+    if (publishingGroup === PublishingGroup.Iap) {
+      setTheme("iap", mode);
+    }
+  }, [setTheme, publishingGroup, mode]);
 
   return (
     <>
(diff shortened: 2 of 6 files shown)


packages/i18n/src/messages/localizedConstants.tsx
```diff
@@ -1576,6 +1576,11 @@ export const publishingGroups = defineMessages({
     id: "Mixlw/",
     description: "The publishing group called Executive Jobs",
   },
+  [PublishingGroup.Iap]: {
+    defaultMessage: "IAP",
+    id: "7ObbNV",
+    description: "The publishing group called IAP",
```
petertgiles, line 1582: Not sure if this gives the translators any helpful context for this acronym.  :laughing:

apps/web/src/pages/Applications/ApplicationLayout.tsx
```diff
@@ -112,6 +116,12 @@ const ApplicationPageWrapper = ({ application }: ApplicationPageProps) => {
     }
   }, [currentPage, navigate, nextStepUrl]);
 
+  React.useEffect(() => {
+    if (publishingGroup === PublishingGroup.Iap) {
+      setTheme("iap", mode);
```
petertgiles, line 121: This completes #6462 as well, right?

api/database/helpers/ApiEnums.php
```diff
@@ -387,6 +387,7 @@ public static function armedForcesStatuses(): array
     /**
      * Publishing Groups
      */
+    const PUBLISHING_GROUP_IAP = 'IAP';
```
petertgiles, line 390: Could you update the seeder for the IAP to use this?
https://github.com/GCTC-NTGC/gc-digital-talent/blob/935f1aa95af8b8b333ec3f8113899ccbbf67c6ac/api/database/seeders/PoolSeeder.php#L45

apps/web/src/pages/Applications/ApplicationLayout.tsx
```diff
@@ -112,6 +116,12 @@ const ApplicationPageWrapper = ({ application }: ApplicationPageProps) => {
...
+    }
+  }, [setTheme, publishingGroup, mode]);
```
petertgiles, line 123: When you leave the application process the theme doesn't switch back.  :smile:
image

petertgiles (changes requested):
This is very cool!  A few things to check on...

api/database/helpers/ApiEnums.php
esizer: Thanks for pointing that out, I didn't realize it existed 😜 Updated in [update IAP pool seeder](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6521/commits/0507690c1ef835ea8e3a43cfac06a51b3e0a97f6)

apps/web/src/pages/Applications/ApplicationLayout.tsx
esizer: Oops, looks like client-side routes don't actually re-render the layout. Made this somewhat hacky hook to fix that in [fix theme not changing on client side route changes](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6521/commits/d5de72cb11b1c34221cf7577a759411596a73d70)

packages/i18n/src/messages/localizedConstants.tsx
esizer: Updated in [improve context of publishing group description](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6521/commits/fa251abc4e0f1d69e71a284fea0417bdd07f1d72)

apps/web/src/pages/Applications/ApplicationLayout.tsx
petertgiles: Oh, it's almost perfect.  I'm getting this now, though.  
image

petertgiles: Hmm, the theme seems to dropping out on navigation now.

https://user-images.githubusercontent.com/8978655/236880988-ef83f0d0-c549-44c0-8a44-4b7deb401262.mp4

esizer: So, I updated the theme changing to be more stable in [stabilize theme changing](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6521/commits/a14cb8ea5a4986681477cd16bc66141256b84074)

However, it flashes very quickly while it sorts out the themes and settles on what it should be showing. This is happening because:

1. Layout loads
2. Sets theme to default
3. Application page loads
4. Retrieves application
5. Checks if it is IAP
6. Sets theme to IAP

The flash is happening in between steps 2-5. I'm trying to think of a way to make this more stable without having to set the theme on _every_ page instead of the layout itself 🤔 Which, as of right now, is the only way I can think of to prevent the flash.

---

Though, taking another look at our `ThemeProvider`, I think this could also use some work. Looks a little wonky and I'm not sure why I wrote it specifically like this. We could leverage our `useLocalStorage` hook and sync the DOM with that instead of all the craziness in this.

petertgiles: Oof, I'm in a loop now.  :cry: 
image

petertgiles (approved):

apps/web/src/pages/Applications/ApplicationLayout.tsx
petertgiles: Seems good now!

PR #6525, opened 2023-05-05
[Feature] Requested pool labelling for requests

### LINKED ISSUE #6366: ✨ SearchRequestFilters component should show Pool names, if the filter includes pools
## ✨ Feature

If a filter filters by pool(s), then the pool name(s) should be shown along with the other filter details. We used to show the pool classification instead of the pool name, but filters can now specify classifications seperately from pools, and it's no longer a good assumption that there is only one pool per classification.

If a filter *does* specify pools, I think we should be transparent about it.

## 🕵️ Details

1. Add "Pool requested" as the first item to the summary of filters before submitting a request
2. Remove the "Pool requested" from the Manager Info section of the admin view, where it doesn't really make sense for it to be.
3. Add it instead to the Summary of filters as the first item.

The SearchRequestFilters component is used on the Request form, when you're submitting a search request, and on the SingleSearchRequest page on the admin side, where you're viewing the submitted request.

## 🎨 Design File

## 📸 Screenshot
Summary of Filters
image

Manager Info (Admin view)
image

Request Information (Admin view)
image

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] If a filter includes pool ids, the pool names are displayed in the Summary of Filters on the Request Form page
- [ ] If a filter includes pool ids, the pool names are displayed in the Request Information section on the Admin SingleSearchRequest page
- [ ] Pool name removed from Manager Information on the Admin SingleSearchRequest page

vd1992:
🤖 Resolves #6366 

## 👋 Introduction

Adds a filter block for "Pool requested" to the filter summary component, as well as removes it from the manager info component. 

## 🕵️ Details

Nothing too fancy. I mimicked how classifications/skills were done to add pool naming to the request page.

On that note.
This feels sub-optimal and the query will get rather slow with more pools and any other new additions. Something to look at for refactoring search.
Is it possible to save to local storage the selected classification object, array of skills selected, and pool object rather than fetch all classifications + skills + pools on the following page to re-assemble the state on the page prior?

Like
```
objectStore = {
    classification : {group ...},
    skills: [Skill},
    pool: {name ...}
}
```

Then on the request form page, `{ group,  level } = { objectStore.classification }`  and so on. 

And then wipe it on request submit. 

The query in question. Departments definitely needs to stay but the other three are fetched on the prior page.

image

Search page -> fetch classifications, pools, skills -> carry-over ids of selected classification, pool. and skills -> Request page -> fetch classifications, pools, skills again -> re-assemble state of search page

Re-assembling example, using fetched all skills and array of skill ids to reproduce search page
image

## 🧪 Testing

1. Go through the search page, submitting a request, and viewing said request
2. Affirm things updated

## 📸 Screenshot

Observe

image

### DIFF
--- a/apps/web/src/components/SearchRequestFilters/SearchRequestFilters.tsx
+++ b/apps/web/src/components/SearchRequestFilters/SearchRequestFilters.tsx
@@ -15,6 +15,7 @@ import {
   getPoolStream,
 } from "@gc-digital-talent/i18n";
 
+import { getFullPoolAdvertisementTitleHtml } from "~/utils/poolUtils";
 import { wrapAbbr } from "~/utils/nameUtils";
 import {
   ApplicantFilter,
@@ -255,6 +256,21 @@ const ApplicantFilters = ({
     <section data-h2-flex-grid="base(flex-start, x2, x.5)">
       <div data-h2-flex-item="base(1of1) p-tablet(1of2)">
         <div>
+          <FilterBlock
+            title={intl.formatMessage({
+              defaultMessage: "Pool Requested",
+              id: "rz8uPO",
+              description:
+                "Title for the pool block in the manager info section of the single search request view.",
+            })}
+            content={
+              applicantFilter
+                ? applicantFilter?.pools?.map((pool) =>
+                    getFullPoolAdvertisementTitleHtml(intl, pool),
+                  )
+                : null
+            }
+          />
           <FilterBlock
             title={intl.formatMessage({
               defaultMessage: "Group and level",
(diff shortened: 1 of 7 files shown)


vd1992:
Also updated the request story, it was outdated, still using PoolCandidateFilter

tristan-orourke (approved):
I'm happy with this as-is. I appreciate the concerns about duplicated queries, but I'm not worried. URQL caching does a lot of work for us - since we loaded classifications, pools and pools on the Search page, they should still be immediately available from the cache on the Request page. As for having to reassemble the pool names, I expect that the number of pools won't be high enough for that to take an appreciable amount of time, for a long time. We'll have to come back to how we load and present pools on the Search page before that's a problem, I expect.

vd1992:
> I'm happy with this as-is. I appreciate the concerns about duplicated queries, but I'm not worried. URQL caching does a lot of work for us - since we loaded classifications, pools and pools on the Search page, they should still be immediately available from the cache on the Request page. As for having to reassemble the pool names, I expect that the number of pools won't be high enough for that to take an appreciable amount of time, for a long time. We'll have to come back to how we load and present pools on the Search page before that's a problem, I expect.

Alright, sounds good! Thanks for your input

PR #6528, opened 2023-05-08
Create a `@self` directive to help organize schema

### LINKED ISSUE #6527: ✨  Create a `@self` directive to help organize schema
## ✨ Feature

Create a custom lighthouse directive, @self, which allows you to create nested types in the schema which are actually aliases to the same underlying model.

## 🕵️ Details

For example, maybe `Applicant` and `User` are two different types in the schema, with different fields and different permissions, but they are actually two aliases for the same underlying database table and laravel model. Then, we could use the `@self` directive like so, and the query should work:
```
type User {
    profile: Applicant @self
}
```

## ✅ Acceptance Criteria

- [ ] Custom Lighthouse directive exists and works as expected

tristan-orourke:
🤖 Resolves #6527 

## 👋 Introduction

Adds the `@self` custom lighthouse directive.

## 🕵️ Details

I wanted to write some php unit tests for this directive, but I can't figure out how to do so. A proper unit test wouldn't be dependent on the schema or models in the app. I tried using a [test schema](https://lighthouse-php.com/6/testing/extensions.html#mock-resolvers), but couldn't figure out how to set it up in a way that properly made use of the directive.

NOTE: you will need to restart VS Code for IDE tools to recognize the `@self` directive.

## 🧪 Testing

1. Temporarily modify the schema: add `applicant: Applicant @self` to the `User` type.
2. Add `AUTH_DEFAULT_USER='admin@test.com'` to api/.env
3. go to http://localhost:8000/graphiql and run the following query:
```
query {
  me {
    applicant {
      email
    }
  }
}
```
4. You should get the email admin@test.com
5. Undo your changes to the schema

### DIFF
--- a/api/app/GraphQL/Directives/SelfDirective.php
+++ b/api/app/GraphQL/Directives/SelfDirective.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace App\GraphQL\Directives;
+
+use Nuwave\Lighthouse\Schema\Directives\BaseDirective;
+use Nuwave\Lighthouse\Schema\Values\FieldValue;
+use Nuwave\Lighthouse\Support\Contracts\FieldResolver;
+
+final class SelfDirective extends BaseDirective implements FieldResolver
+{
+    // TODO implement the directive https://lighthouse-php.com/master/custom-directives/getting-started.html
+
+    public static function definition(): string
+    {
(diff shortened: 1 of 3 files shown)


esizer (approved):
LGTM, just need to run a schema refresh 👍

Issue #6067, opened 2023-03-23
✨ Application Revamp: Skill Requirement page


tristan-orourke:
## ✨ Feature

Skill Requirement page, where existing experiences are linked to skills.

## 🕵️ Details

Links between experiences and skills should be soft-deleted so text is saved (I think that's already the case). 

Edit buttons and Connect a resume experience buttons open a dialog.

The dialog allows for experience to be selected from a dropdown, and once it is, you can describe how you used this skill in an open text field.

If the dialog opened via the Edit button, the experience selector should be locked to the existing one, and there should also be a delete button.

Asset technical skills CAN be added but are not required. 
Behavioral skills are listed but have no content entry. 

Experience Card accordion is the same as on the Resume page, but with additional How I Used this Skill section. If a skill section of a card is empty, hide that section (instead of showing a null state).

Can only continue to next step when all required technical skills have at least one experience.

## 🎨 Design File

https://www.figma.com/file/BXYVsu1fo9LS2IySJDUINY/Applicants---Job-application?node-id=636-59839&t=GWcZoH5WyLW4uwaG-0

## 📸 Screenshot

### First time user (Skills have no linked experiences yet)
![Step 5 - Skill justification (empty)](https://user-images.githubusercontent.com/4313717/227238002-0ecad560-59b0-4293-92a0-be40f399fb3c.png)

### When some experiences are linked
![Step 5 - Skill justification (complete)](https://user-images.githubusercontent.com/4313717/227238205-93320b16-84eb-4942-b0ed-e7345b73df55.png)

### Dialog when it first opens
![image](https://user-images.githubusercontent.com/4313717/227238454-c93863a4-12fa-4693-827e-f4a533a8d7db.png)

### Dialog after an experience has been selected
![image](https://user-images.githubusercontent.com/4313717/227238634-1fa11386-3588-451c-836f-21e466f1748c.png)

### Dialog when editing an experience
![image](https://user-images.githubusercontent.com/4313717/227242209-5dd676fc-1671-4a9a-a729-16a423cac99b.png)

## 🙋‍♀️ Proposed Implementation

(optional) Do you have a proposed implementation?

## 🌎 Localization

| English                                                                                                                                                                                                                                                                                                                               | French                                                                                                                                                                                                                                                                                                                                                   | Description                                                                                                   |
[...]

PR #6530, opened 2023-05-08
[Feature] Make application steps optional

### LINKED ISSUE #6321: ✨ Application Revamp: accommodate Pools without screening questions
## ✨ Feature

If a pool does not have screening questions, the Screening Questions step and any related UI should disappear from the application process.

## 🕵️ Details

If the application is for a Pool which has no screening questions:
- [ ] Stepper should not render a Screening questions, and the step count should be out of 6 instead of 7
- [ ] Where the breadcrumbs include step number, like  "Step 7", the number should be determined dynamically.
- [ ] Save and continue on the Skills step should go to Review page instead of Education step
- [ ] Screening Questions section is not rendered on the Review Application page
- [ ] Not having any responses doesn't block application validation or submission

## 🎨 Design File

## 📸 Screenshot

I think wherever we're using "Step #" we should try to make it dynamic, since in the future we maybe adding steps as well, like for the IAP.

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Application is unaffected if it Pool has any screening questions

If the application is for a Pool which has no screening questions:
- [ ] Stepper should not render a Screening questions, and the step count should be out of 6 instead of 7
- [ ] Where the breadcrumbs include step number, like  "Step 7", the number should be determined dynamically.
- [ ] Save and continue on the Skills step should go to Review page instead of Education step
- [ ] Screening Questions section is not rendered on the Review Application page
- [ ] Not having any responses doesn't block application validation or submission
- [ ] QUESTION: what happens if you manually go to the screening questions url, and application has no screening questions?

petertgiles:
> __Warning__
I will probably rebase and do some more testing after #6067 and #6462 are merged.

🤖 Resolves #6321

## 👋 Introduction

This branch makes it possible to have application steps (the screening questions specifically) optional in the flow.

## 🕵️ Details

The main thrust of this PR is to change the layout component from managing pages to manage steps and to dynamically build the list of steps.

## 🧪 Testing

1. Rebuild the app
2. Run through an application with screening questions
3. Run through an application with no screening questions

- [ ] Optional step does not appear
- [ ] Stepper counts stays accurate
- [ ] The "next" button in the preceding step correctly jumps over skipped step

## 📸 Screenshot

image


esizer:
Wow, this is great! I did notice something though, if screening questions are omitted then the breadcrumbs no longer make sense 🤔 . The Review page says it is  step 7 even though there are only 6 steps 😅 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/e925c826a1e15f24d940b38a80cdd95da91f6019/apps/web/src/pages/Applications/ApplicationReviewPage/ApplicationReviewPage.tsx#L26-L35

petertgiles:
> if screening questions are omitted then the breadcrumbs no longer make sense

Shoot, missed that AC.  :sweat_smile: OK, I think I have it now.

esizer (approved):
Nice! Super clean 🧼

PR #6534, opened 2023-05-08
[Feature] Pool candidate skill count

### LINKED ISSUE #6395: ✨ Add Skill Rank to candidate results table
## ✨ Feature

When looking at the results of candidates inside individual request it would be helpful to know which candidates have which of the skills selected in the skill filter.

## 🕵️ Details

Add a column to the candidate results that shows how many of the requested skills each user has claimed. This will allow DCM to list candidates from MOST skill matches to LEAST skill matches, and help DCM decide which candidates fit best to the request.

## 🎨 Design File

[Link to figma](https://www.figma.com/file/UsrB15yicgDCLDQmx3snyl/User-Flows?node-id=3045%3A4807&t=zbLzQ0dXanipnzTx-1)

## 📸 Screenshot

image

Instead of returning `[PoolCandidate]`, maybe we return a list of a wrapper object: `[PoolCandidateWithSkillCount]`
```
type PoolCandidateWithSkillCount {
  poolCandidate: PoolCandidate @self
  skillCount: number
}
```
We should be able to add a column to the query by using a scope which uses [addSelect](https://laravel.com/docs/10.x/queries#specifying-a-select-clause).

## ✅ Acceptance Criteria

- [ ] Modify the poolCandidatesPaginated query to include the matching-skill-count for each candidate
- [ ] Table has a "Number of skills matched" result column
- [ ] "Number of skills matched" result column can be sorted
- [ ] Order the results by skill rank, descending, by default

esizer:
🤖 Resolves #6395 

## 👋 Introduction

Adds a virtual `skillCount` column to `poolCandidatesPaginated` paginated query.

## 🕵️ Details

## 📋 TO DO

- [ ] Update with wrapping type
- [ ] Add translations

## 🧪 Testing

1. `npm run build:fresh`
2. Navigate to a request `/admin/talent-requests/{requestId}`
3. Clear filters
4. Add a lot of skills
5. Confirm the "Number of skills column" updates with matching skills
6. Confirm they sort (kinda weird since status and priority) are locked.

## 📸 Screenshot

Screenshot 2023-05-08 150519

### DIFF
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -12,7 +12,9 @@ use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 use Illuminate\Database\Eloquent\Relations\HasMany;
 use Illuminate\Database\Eloquent\Builder;
 use App\Http\Resources\UserResource;
+use Illuminate\Support\Facades\DB;
 use Carbon\Carbon;
+use Illuminate\Database\Query\JoinClause;
 use Illuminate\Support\Facades\Log;
 
 /**
@@ -148,31 +150,28 @@ class PoolCandidate extends Model
      * Number of skills the user has claimed that match
      * the essential and nonessential skills for the pool
      */
-    public function getSkillCountAttribute()
-    {
-        $this->load([
-            'pool.essentialSkills',
-            'pool.nonessentialSkills',
-            'user.awardExperiences.skills',
-            'user.communityExperiences.skills',
-            'user.educationExperiences.skills',
-            'user.personalExperiences.skills',
-            'user.workExperiences.skills',
-        ]);
-        $skillIds = collect();
-        $skillIds = $skillIds->merge($this->pool->essentialSkills->pluck('id'));
-        $skillIds = $skillIds->merge($this->pool->nonessentialSkills->pluck('id'));
+    // public function getSkillCountAttribute()
+    // {
+    //     $this->load([
+    //         'pool.essentialSkills',
+    //         'pool.nonessentialSkills',
+    //         'user.awardExperiences.skills',
+    //         'user.communityExperiences.skills',
+    //         'user.educationExperiences.skills',
+    //         'user.personalExperiences.skills',
+    //         'user.workExperiences.skills',
+    //     ]);
+    //     $skillIds = collect();
+    //     $skillIds = $skillIds->merge($this->pool->essentialSkills->pluck('id'));
+    //     $skillIds = $skillIds->merge($this->pool->nonessentialSkills->pluck('id'));
 
-        $count = 0;
-        foreach ($this->user->experiences as $experience) {
-            $skillCount = $experience->skills()->whereIn('skill_id', $skillIds)->count();
-            if ($skillCount) {
-                $count += $skillCount;
-            }
-        }
+    //     $count = 0;
+    //     $skillCount = $this->user->experiences->whereHas('skills', function(Builder $query) use($skillIds) {
+    //         $query->whereIn('skill_id', $skillIds);
+    //     })->count();
 
-        return $count;
-    }
+    //     return $skillCount;
+    // }
 
     public static function scopeQualifiedStreams(Builder $query, ?array $streams): Builder
     {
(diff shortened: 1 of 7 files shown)


api/graphql/schema.graphql
```diff
@@ -498,6 +498,12 @@ type PoolCandidate {
     @with(relation: "educationRequirementWorkExperiences")
 }
 
+type PoolCandidateWithSkillCount {
+  id: ID! @self(key: "id")
+  poolCandidate: PoolCandidate! @self
+  skillCount: Int @rename(attribute: "skill_count")
```
tristan-orourke, line 504: Could you add a comment noting that skillCount will be null unless the query is scoped by skillsAdditive or skillsIntersectional?

esizer: Great suggestion, added in [add note for skillCount field value](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6534/commits/e88e593256e5e6b4e9c8d63270282739358bfe35)

api/app/Models/PoolCandidate.php
```diff
@@ -536,6 +540,8 @@ public function scopeSkillsIntersectional(Builder $query, ?array $skills): Build
             return $query;
         }
 
+        $query = $this->addSkillCountSelect($query, $skills);
+
         // call the skillFilter off connected user
         $query->whereHas('user', function (Builder $userQuery) use ($skills) {
             User::scopeSkillsIntersectional($userQuery, $skills);
```
vd1992, line 547: There isn't much of a point to adding it to `scopeSkillsIntersectional` though is there? It will always return the number of skills selected as it is AND matching.

But it does make things consistent and easy to swap skill fields by having consistent typing/return value 😕

tristan-orourke: That is a very good point 😆

tristan-orourke (changes requested):
Manually tested, it works great, but a few comments:

1. I don't think the changes to the `@self` directive are necessary. Explained more in a comment.
2. If you query for skillCount without providing skills to the filter, you don't get null, you get an internal server error 😵 . It would be nice to avoid that, if possible.
3. This query definitely deserves a phpunit test now, ensuring skillCount keeps working as we expect.

api/graphql/schema.graphql
```diff
@@ -498,6 +498,12 @@ type PoolCandidate {
     @with(relation: "educationRequirementWorkExperiences")
 }
 
+type PoolCandidateWithSkillCount {
+  id: ID! @self(key: "id")
```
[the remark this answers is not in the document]
esizer: Thanks for the tip, reverted in [revert changes to self directive](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6534/commits/780060b158111822392d70c5361f84fe06e4cc8a)

api/graphql/schema.graphql
```diff
@@ -994,8 +1000,13 @@ type Query {
...
-    @paginate(defaultCount: 10, maxCount: 1000, scopes: ["notDraft"])
+  ): [PoolCandidateWithSkillCount]!
+    @paginate(
+      defaultCount: 10
+      maxCount: 1000
+      scopes: ["notDraft"]
+      model: "App\\Models\\PoolCandidate"
+    )
     @guard
     @can(ability: "view", resolved: true)
```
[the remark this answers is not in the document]
esizer: I think that makes sense, I wasn't having any errors but to be safe I added it in [add model to](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6534/commits/114049a2a1f568080e24a440fb0bb89eac27da32) @can [for paginated pool candidates](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6534/commits/114049a2a1f568080e24a440fb0bb89eac27da32)

esizer:
> 2\. If you query for skillCount without providing skills to the filter, you don't get null, you get an internal server error 😵 . It would be nice to avoid that, if possible.

Okay, this is kind of messy but I was able to get it to work in [hack, scope to avoid server error with skill count](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6534/commits/2beea872ae748bae2360a3f556f56aec7b0b4b0a)

Essentially, if the query contains `skill_count` only once (in order by) then we add a null value, otherwise we let the skill scope takeover. There could be a better way of doing this but I'm struggling finding a cleaner solution 😓

tristan-orourke (changes requested):
Great, I like the fixes. Could you just add a unit test which tests the query? The cases which should be tested, off the top of my head:
- querying for skillCount
- ordering by skillCount
- doing the above when no skills are in the filter
- doing the above with one skill in the filter
- doing the above with multiple skills in the filter

esizer:
> Great, I like the fixes. Could you just add a unit test which tests the query? The cases which should be tested, off the top of my head:
> * querying for skillCount
> * ordering by skillCount
> * doing the above when no skills are in the filter
> * doing the above with one skill in the filter
> * doing the above with multiple skills in the filter

Okay, I updated the tests to attempt to cover all of these cases. I think I got them all, let me know if I missed any or if you thought of others that make sense.

[improve tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6534/commits/d9e536d465450c33cf504ce5123e9781bbbe1cf5)

tristan-orourke (approved):
Looks great!

PR #6535, opened 2023-05-08
[Bug] GovernmentInfoForm disconnect BelongsTo relations 

### LINKED ISSUE #6413: 🐛 Government Information form `isGovEmployee` logic does not clear government values
## 🐛 Bug

When switching from Yes to No in the **Do you currently work for the government of Canada?** input, the **Which department do you work for?**, **Current Classification Group**, and **Current Classification Level** are not cleared.

## 🦋 Expected Behaviour

If No is selected for **Do you currently work for the government of Canada?** and the form is saved, **Which department do you work for?**, **Current Classification Group**, and **Current Classification Level** should be saved as `NULL`.

## 📋 Steps to Reproduce

1. Navigate to `/users/:userId/profile/government-info/edit`
2. Select **Yes, I am a Government of Canada employee**
3. Enter values in subsequent form fields
4. Save form
5. Select **No, I am not a Government of Canada employee**
6. Save form
7. Select **Yes, I am a Government of Canada employee**
8. Observe subsequent form fields still have saved values (except for **As an employee, what is your employment status?**)

## 📸 Screen Recording

https://user-images.githubusercontent.com/3046459/234625020-67bf5385-a85b-4f87-ab6c-111d6d0032fb.mov

## ✅ Acceptance Criteria

- [ ] government fields are cleared on save if user indicates they are not a government employee

vd1992:
🤖 Resolves #6413 

## 👋 Introduction

The removal of `BelongsTo` relation classification and department wasn't written how it should be used. Written to use `disconnect`. 

## 🧪 Testing

Update the government info section with user info as a government employee then update it as not an employee. 
Keep an eye on the network tab and ensure requests/responses are what they should be for the operation. 

## 📸 Screenshot

image

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -1411,6 +1411,7 @@ input PoolBelongsToMany {
 
 input DepartmentBelongsTo {
   connect: ID!
+  disconnect: Boolean
 }
 
 input DepartmentBelongsToMany {
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -1410,7 +1410,7 @@ input PoolBelongsToMany {
 }
 
 input DepartmentBelongsTo {
-  connect: ID!
+  connect: ID
   disconnect: Boolean
 }
 
--- a/api/storage/app/lighthouse-schema.graphql
+++ b/api/storage/app/lighthouse-schema.graphql
@@ -417,6 +417,7 @@ type Department {
 
 input DepartmentBelongsTo {
   connect: ID!
+  disconnect: Boolean
 }
 
 input DepartmentBelongsToMany {
--- a/api/storage/app/lighthouse-schema.graphql
+++ b/api/storage/app/lighthouse-schema.graphql
@@ -416,7 +416,7 @@ type Department {
 }
 
 input DepartmentBelongsTo {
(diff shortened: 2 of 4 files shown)


petertgiles (changes requested):
This fix works perfectly!
But could you fix the same bug in the profile page of the application revamp? :smile: 
http://localhost:8000/en/applications/{pool_candidate}/profile

petertgiles:
Note to self:
```sql
select
  u.is_gov_employee,
  d.name->>'en' department_name,
  u.gov_employee_type,
  c."group" || '-' || c."level" classification,
  u.has_priority_entitlement,
  u.priority_number 
from users u
left join departments d on u.department = d.id
left join classifications c  on u.current_classification = c.id
where u.id = :id
```

vd1992:
> This fix works perfectly! But could you fix the same bug in the profile page of the application revamp? 😄 [http://localhost:8000/en/applications/{pool_candidate}/profile](http://localhost:8000/en/applications/%7Bpool_candidate%7D/profile)

Oops 😳 forgot that 
Done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/6535/commits/d2b8667b7a466b6cda7c9c3adfde0c8326b42870

Issue #2041, opened 2022-02-09
Bug - Its possible after logging in for the client to get stuck in an infinite loop of using the refresh token


tristan-orourke:
After coming back to the client after logging in on SiC, a graphql request returns this:

```
[...]

PR #6547, opened 2023-05-10
[Dep] Bump auth exchange

### LINKED ISSUE #5981: ♻️ Dependency - urql/exchange-auth 2.0.0
## ♻️ Debt/Refactor

Version 2.0.0 has breaking changes to resolves
As part of this PR, upgrade urql to v4

## 🕵️ Details

https://github.com/urql-graphql/urql/releases/tag/%40urql%2Fexchange-auth%402.0.0

The urql breaking changes hopefully don't affect us.

## ✅ Acceptance Criteria

- [ ] update the dependency
- [ ] `urql` is bumped past v4
- [ ] App builds and functions properly

### LINKED ISSUE #7764: ♻️ `errorExchange` deprecated
## ♻️ Debt/Refactor

The `urql` `errorExchange` has been deprecated in favour of the `mapExchange`

## 🕵️ Details

It looks as though the new `mapExchange` covers everything the old `errorExchange` did so it has been deprecated and should be replaced.

REF: https://formidable.com/open-source/urql/docs/api/core/#errorexchange-deprecated

## 🙋‍♀️ Proposed Solution

The `mapExchange` contains an `onError` option which is called when there is an error. It should be as easy as removing `errorExchange` and moving all the code to the `onError` portion of the `mapExchange`.

## ✅ Acceptance Criteria

- [ ] Our API client is no longer using `errorExchange`
- [ ] The `mapExchange` is configured to work in the same way as the previous `errorExchange`
- [ ] The app still logs and notifies users of errors

esizer:
🤖 Resolves #5981, resolves #7764

## 👋 Introduction

Bumps `urql` to `v4` and `@urql/exchange-auth` to `v2`.

## 🕵️ Details

Bumps both deps and fixes breaking changes. As part of that, `urql` reccomends defeining the config inline so it was moved to be there.

`getOperationName` was removed from exports so the test needed to be updated to determine operation name with conditionals.

## 🧪 Testing

2. Login and confirm auth still functions
3. Confirm logging out works
4. Check expired token refresh


packages/client/src/components/ClientProvider/ClientProvider.tsx
```diff
@@ -204,19 +118,70 @@ const ClientProvider = ({
               }
             },
           }),
-          dedupExchange,
```
petertgiles, line 207: What happened to dedupExchange?  Is it part of the libaray now?

packages/client/src/components/ClientProvider/ClientProvider.tsx
```diff
@@ -204,19 +118,70 @@ const ClientProvider = ({
...
+                }),
+              didAuthError(error) {
+                return error && error.response
+                  ? error.response.status === 401 ||
+                      error.graphQLErrors.some(
+                        (e) => e.extensions?.category === "authentication",
+                      )
+                  : false;
+              },
+              async refreshAuth() {
```
petertgiles, line 154: Hmm, I don't think the refresh is working for me.  It looks like the refresh request was properly issued but the subsequent requested didn't use the new toke?
image
image

petertgiles (changes requested):
Mostly runs but I'm seeing trouble with the refresh.  Something to do with taking it out of useCallback, maybe?

packages/client/src/components/ClientProvider/ClientProvider.tsx
esizer: Yup! 

https://github.com/urql-graphql/urql/blob/1470a9c7474c75b65303b503c0a3a51e895f1a38/packages/core/src/exchanges/dedup.ts#L3-L6

packages/client/src/components/ClientProvider/ClientProvider.tsx
esizer: It looks like you were right about the `useCallback`. I moved it into that and it seems to work. 🤷‍♀️ 

[move auth refresh to useCallback](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6547/commits/dbf68c7eae57611f0c1e9c0c9f0d9a853a244dc7)

petertgiles: Nice, the refresh works!  It seems not-quite-right though.  I'm observing it taking two tries before it "sticks".  This is the sort of thing that "willAuthError" is supposed to avoid, I think.

image

I'm going to try with SIC to see if it does the same thing.

esizer: Thanks, yeah they replaced it with some `refreshAuth` and said to define it inline. I used what we used to have and merged it with their example code but I am not super sure about what i did 😅

petertgiles: Hmm, closer look at the timings.
image
It's actually attempting the second refresh before it "knows" that the graphql request errored out.  :confused:

esizer: I wonder if it has something to do with either the `return` or `await`? 🤔 I added the return to satisy the type and no other reason (I don't remember them returning anything from this function)

https://github.com/GCTC-NTGC/gc-digital-talent/blob/dbf68c7eae57611f0c1e9c0c9f0d9a853a244dc7/packages/client/src/components/ClientProvider/ClientProvider.tsx#L86-L93

petertgiles: I bet it is the new dependencies of the `internalClient`:
image
When we get a new refresh token we get a new `internalClient` with a reset state that doesn't know it has requests in flight.
This sounds an awful lot like  #2041 so it's probably a similar problem with a similar solution.  The reason this is dangerous is because if a client keeps getting 401's repeatedly they will get stuck in a refresh loop.  :fearful:

petertgiles: Or it's just plain not working.  :laughing: I tried SIC which is more strict, of course.  The refresh just plain fails.  I think the new tokens are not getting loaded into state properly.
image

petertgiles: SIC detail:
image
image

esizer: Hmm, okay. This should give me enough info to work off of. Thanks. I may be pinging you later if I need more assistance 😛

esizer: Not entirely sure if this works, still triggers two refreshes but no more 401's 😬 

- [move auth refresh to useCallback](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6547/commits/dbf68c7eae57611f0c1e9c0c9f0d9a853a244dc7)
- [update client memo dependencies](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6547/commits/8d6ac47237ce32606e8947f2c124b404add3aeec)

petertgiles: OK, I tested again with SIC.  I'm not seeing it try the refresh twice but it still isn't handling the state properly.  It seems to wait until the refresh is done but it tries again with the old token, fails, then tries with the new token.
image

packages/client/src/components/ClientProvider/ClientProvider.tsx
```diff
@@ -1,17 +1,15 @@
-import { authExchange } from "@urql/exchange-auth";
+import React, { useMemo } from "react";
+import { AuthUtilities, authExchange } from "@urql/exchange-auth";
```
mnigh, line 2: `AuthUtilities` is defined but never used

```suggestion
import { authExchange } from "@urql/exchange-auth";
```

esizer:
Blocking this on the PKCE ticket and converting to draft until we know what it happening there.

esizer:
> I found one last issue that could use a look, though. If an error occurs during a graphql request the user is put into a refresh loop. One way to do this is if you intentionally break the introspection query.

Thanks for your help with this. I believe I have it but it is kind of janky if you ask me. Meaning, when it attempts to refresh it seems to take a while with the spinner grinding a bit. Not sure how to smooth that process out though 😕 

I'm wondering if the check needs to be more strict than just checking for a 401 as well? 🤔 Should we be checking for specific error messages?

- [prevent client recreation](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6547/commits/eb2b6817e49945de42c9931a901a56241dc77b18)
- [logout on auth error if we cant validate token](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6547/commits/7ead9318e9530c0073ab5f230dd6d35fd04301fa)

petertgiles:
Yep, this broke the "user deleted" behaviour.  :disappointed:   If you try to log in with a deleted user you end up here instead of the "user deleted" page.
image

esizer:
> Yep, this broke the "user deleted" behaviour. 😞 If you try to log in with a deleted user you end up here instead of the "user deleted" page.

Okay, I pushed up my WIP as discussed that includes the refactor of our auth exceptions. This will return a partial graphql error that includes the exception along with a `null` user and `200` status code ([what we are supposed to do anyway](https://the-guild.dev/blog/graphql-error-handling-with-fp#:~:text=Contrary%20to%20REST%2C%20GraphQL%20does%20not%20use%20Status%20Code%20to%20differentiate%20a%20successful%20result%20from%20an%20exception.)).

The idea is, we can look at the error extension and act upon it rather than a less than useful `401` with no other information. The only thing I can't figure out is why it is sending the same error 3 times with the login 🤔 

[update to use custom exception for auth](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6547/commits/f312838a786d9f1a40391cda480f0dc886ee84d0)

petertgiles (approved):
I like it!  There's a test that will need to be updated but I like the way it works.

The three-times-error is strange but it's enough of a niche case that I don't mind too much. 

:partying_face:

esizer:
> The three-times-error is strange but it's enough of a niche case that I don't mind too much.

Okay, as mentioned, I have updated this to work a little bit better on the frontend using the new error codes. Also, I wanted to log the error and in order to do so and not duplicate code I went ahead of removed the `errorExchange` (deprecated) and moved the code to the `mapExchange`. All done in [update frontend with new authentication errors](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6547/commits/217c8aa893499cf67a43122c3e506c79e4829f2b)

As well, I[ updated the failing test with the new error codes](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6547/commits/e10c9db27e9fd4192887093a32966f6909e562a2) 👍 

We just need to solve where to send the user when we log them out due to an invalid token and what to say, in the interim I just sent them to `/logged-out`

esizer:
@petertgiles Created a ticket based on the discussion during stand up. Feel free to make any changes if I got it wrong or missed anything 😄 

- https://github.com/GCTC-NTGC/gc-digital-talent/issues/7811

petertgiles (approved):
Looks great!  :shipit:

esizer:
> Oh, just noticed that you probably need to commit your packagelock file too.

Ooooo, great catch! I merged main and forgot to run install again.

PR #6574, opened 2023-05-11
[Bug] Fix experience over caching

### LINKED ISSUE #6532: 🐛 Application Revamp: New experiences don't show in page until hard reload
## 🐛 Bug

If you add an experience to your resume you have to reload the page to see it.

## 🦋 Expected Behaviour

The new experience should show immediately after navigating back to the page.

## 📋 Steps to Reproduce

1. Start an application
2. Advance to the resume page
3. Add a new experience to the resume
4. Click save and go back
5. Observe that the new experience doesn't show up until you reload the page

## 📸 Screenshot

https://images.zenhubusercontent.com/364642496/1f6c5b92-76ed-4aad-82da-03f0aff84d3d/simplescreenrecorder_2023_05_08_14_38_08.mp4

## ✅ Acceptance Criteria

- [ ] The new experience should show immediately after navigating back to the page.

esizer:
🤖 Resolves #6532 

## 👋 Introduction

Fixes a caching issue with experiences within the application workflow.

## 🕵️ Details

The main query is called at the top and cached. The main query was only aware of and caching the top level (PoolCandidate and Applicant) and only invalidating them. This extracts the experiences out into their own query to avoid the main cache and refetch when new experiences are added.

## 🧪 Testing

> **Note**
> It first reads from the cache, then makes a request. This causes a slight pop in effect. Wondering if it would be worth it to see if we could implement [AnimatePresence](https://www.framer.com/motion/animate-presence/) to make this a little smoother 🤔 

2. Create an application
3. On the resume page, delete all of your experiences
4. Ensure they are removed from the list
5. Add an experience
6. Confirm it shows up in the list of experiences without needing to refresh


petertgiles (approved):
Looks and works great!  Sorry about hte merge conflicts. :-(

PR #6596, opened 2023-05-15
[Bug] Fixes pool candidates skill count

### LINKED ISSUE #6594: 🐛 Pool Candidates Number of skills matched column not working
## 🐛 Bug

Pool Candidates Number of skills matched column not working. `scopeWithSkillCount` logic seems to miss a few cases.

## 🦋 Expected Behaviour

**Number of skills matched** should show numbers when skill filter applied and candidates match.

## 📋 Steps to Reproduce

1. Visit a pool candidates table
2. Add skills in the filter
3. Observe no skills showing up for **Number of skills matched**

mnigh:
🤖 Resolves #6594.

h/t @esizer.

## 👋 Introduction

This PR fix the pool candidates **number of skills matched** column that was not working correctly and strengthens the tests associated with it.

## 🧪 Testing

1. Run `docker-compose exec -w /home/site/wwwroot/api webserver sh -c "./vendor/bin/phpunit ./tests/Feature/PoolCandidateTest.php"`
2. Ensure tests passed 10 times

### DIFF
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -604,10 +604,12 @@ class PoolCandidate extends Model
     public function scopeWithSkillCount(Builder $query)
     {
         // Checks if the query already has a skill_count select and if it does, it skips adding it again
-        $currentSql = $query->getQuery()->toSql();
+        $currentSql = strtolower($query->getQuery()->toSql());
         $skillCountAppearances = substr_count($currentSql, 'skill_count');
-        $orderedBySkillCount = str_contains($currentSql, 'order by "skill_count"');
-        if ($orderedBySkillCount && $skillCountAppearances === 2) {
+        $orderedBySkillCountAsc = str_contains($currentSql, '"skill_count" asc');
+        $orderedBySkillCountDesc = str_contains($currentSql, '"skill_count" desc');
+        $orderedBySkillCount = $orderedBySkillCountAsc || $orderedBySkillCountDesc;
+        if ((($orderedBySkillCount) && $skillCountAppearances === 2) || !($orderedBySkillCount && $skillCountAppearances === 1)) {
             return $query;
         }
 
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -14,6 +14,9 @@ use Illuminate\Database\Eloquent\Builder;
 use App\Http\Resources\UserResource;
 use Illuminate\Support\Facades\DB;
 use Carbon\Carbon;
+use Illuminate\Database\Query\Expression;
+use Illuminate\Support\Facades\Log;
+use Illuminate\Support\Str;
 
 /**
  * Class PoolCandidate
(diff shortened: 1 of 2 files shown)


api/app/Models/PoolCandidate.php
```diff
@@ -604,10 +604,12 @@ public function setInsertSubmittedStepAttribute($applicationStep)
...
         // Checks if the query already has a skill_count select and if it does, it skips adding it again
-        $currentSql = $query->getQuery()->toSql();
+        $currentSql = strtolower($query->getQuery()->toSql());
         $skillCountAppearances = substr_count($currentSql, 'skill_count');
-        $orderedBySkillCount = str_contains($currentSql, 'order by "skill_count"');
-        if ($orderedBySkillCount && $skillCountAppearances === 2) {
+        $orderedBySkillCountAsc = str_contains($currentSql, '"skill_count" asc');
+        $orderedBySkillCountDesc = str_contains($currentSql, '"skill_count" desc');
+        $orderedBySkillCount = $orderedBySkillCountAsc || $orderedBySkillCountDesc;
+        if ((($orderedBySkillCount) && $skillCountAppearances === 2) || !($orderedBySkillCount && $skillCountAppearances === 1)) {
```
vd1992, line 612: I have run the test in question locally 20 times successfully. 
I can also filter by skills, get a skill count back, sort skill count by ascending or descending. So things seem working!

I am not 100% sure about what is going on in the snippet so I don't know if there are any potential problems. Like, if `orderedBySkillCount` is false, like in the case of a non-sorting query calling the scope, couldn't it potentially keep appending the same `selects` on? Or will the scope never be called by query without an "order by" argument also present?

esizer: So, we are adding an aliased select in the `scopeAdditiveSkills` as `skill_count`. This means that the column only gets added when you have the scope applied. This bit of code is checking to see if you are ordering but have not used the scope to get `skill_count` in the query. If that is the case, it just adds `select null as skill_count` to the query so there is a column there to order on (even if they are all null).

This is required to prevent the query from crashing.

tristan-orourke: I'm having trouble testing this at the moment, but try looking at `query-getQuery()->columns`. It might allow for a less hacky way of checking which columns have been selected for already, based on this stackoverflow commend: https://stackoverflow.com/questions/71697537/laravel-eloquent-builder-get-selected-columns#comment126710976_71697537

esizer: lmao, my googling ability must suck. I tried finding exactly that for like an hour and couldn't come up with it 😆

esizer: Great find, @tristan-orourke . The check has been updated to use `columns` instead of all of the SQL. This should make things a little less hacky and more readable 😅 

Updated in [fallback now checks columns over all SQL](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6596/commits/59bb46744fa2f0cb9881e5b197268b097f503e2f)

petertgiles:
I took a look at the query as @vd1992 requested.  I came into this cold because I didn't look at the original PR so I'm not totally following what's going on.  The auto-generated query is.... also pretty gnarly.  I'm not finding any bugs but I'm not confident enough to approve this right now.

tristan-orourke (approved):
Tests pass, code looks good, @vd1992 ran it 20 times, I think it's good to go.

PR #6597, opened 2023-05-15
[Bug] Fix sorting of Award Experiences

### LINKED ISSUE #6456: 🐛 compareByDate for experience is duplicated and doesn't handle awards
## 🐛 Bug

I found that the compareByDate function for sorting awards by date is duplicated and doesn't handle awards.

## 🕵️ Details

Because award is the only type of experience without a start date and end date it gets ignored in sorting.

## 📋 Steps to Reproduce

1. Add a bunch of experiences of different types with different dates (or seed it)
2. Start an application and visit the skills and experiences page
3. Sort by date

## 📸 Screenshot

image

image

## ✅ Acceptance Criteria

- [ ] Only one copy of function should exist
- [ ] Function should sort awards using award date

vd1992:
🤖 Resolves #6456 

## 👋 Introduction

Make sorting experiences work with Award Experiences and removed the duplicate sort.  

## 🕵️ Details

Solved by treating them as the same as the other experience types with the special state that award_date = start_date = end_date

As well, resolved some sorting bugs I noticed:
`nonEmptyExperiences` was introduced in #6574, but needs to be passed into what is rendered for the sort to happen
`watchSortExperiencesBy` starts off as undefined, which makes the sort initially disconnected from the select, resolved with a default
Seeding allows for end dates before start dates for experiences, changed that

## 🧪 Testing

1. head to `/en/applications/:id/resume`
2. observe sorting
3. create more experiences and see sorting

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/utils/experienceUtils.ts
+++ b/apps/web/src/utils/experienceUtils.ts
@@ -320,10 +320,30 @@ export const isWorkExperience = (e: AnyExperience): e is WorkExperience =>
   e.__typename === "WorkExperience";
 
 export const compareByDate = (e1: ExperienceForDate, e2: ExperienceForDate) => {
-  const e1EndDate = e1.endDate ? new Date(e1.endDate).getTime() : null;
-  const e2EndDate = e2.endDate ? new Date(e2.endDate).getTime() : null;
-  const e1StartDate = e1.startDate ? new Date(e1.startDate).getTime() : -1;
-  const e2StartDate = e2.startDate ? new Date(e2.startDate).getTime() : -1;
+  // fit AwardExperience to startDate - endDate format
+  const e1Adjusted = e1;
+  const e2Adjusted = e2;
+  if (e1.__typename === "AwardExperience") {
+    e1Adjusted.startDate = e1.awardedDate;
+    e1Adjusted.endDate = e1.awardedDate;
+  }
+  if (e2.__typename === "AwardExperience") {
+    e2Adjusted.startDate = e2.awardedDate;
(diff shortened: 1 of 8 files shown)


vd1992:
The page currently on `main`

gif

apps/web/src/utils/experienceUtils.ts
```diff
@@ -320,10 +320,30 @@ export const isWorkExperience = (e: AnyExperience): e is WorkExperience =>
...
+  const e1Adjusted = e1;
+  const e2Adjusted = e2;
+  if (e1.__typename === "AwardExperience") {
+    e1Adjusted.startDate = e1.awardedDate;
+    e1Adjusted.endDate = e1.awardedDate;
+  }
+  if (e2.__typename === "AwardExperience") {
+    e2Adjusted.startDate = e2.awardedDate;
+    e2Adjusted.endDate = e2.awardedDate;
+  }
```
esizer, line 333: Since `AwardExperience` only has a single date would it make sense to choose either `startDate` or `endDate` instead of setting them both to the same date? 🤔

vd1992: If you only pick say a start date, and end date is null it gets treated as an ongoing experience wouldn't it?
If that is preferable then I can change that, but it would bubble award experiences all to the top then. 

I assigned both start and end to have them sort in the same way as the other experiences.

esizer: Okay, that makes sense to me 👍

esizer (approved):
LGTM! 👍

PR #6634, opened 2023-05-18
[Feature] Add restricted page

### LINKED ISSUE #5362: Secure Network Error Page
We've asked IMTD for a more informative error message when trying to access the admin site from off-network devices.  They can either display a static page that we provide or they can redirect to a page for us.  We've decided we want the redirect so we have more control over it.  So we need to provide a static error page for ourselves now.

Design: ???

Copy: ???

URL: `talent.canada.ca/external-network-error `

Acceptance:
- [ ] Static page exists
- [ ] Static page can be routed to
- [ ] IMTD has been requested to setup the error redirect

petertgiles:
🤖 Resolves #5362 

## 👋 Introduction

This branch adds a restricted notice page that the F5s can redirect to when necessary.

## 🧪 Testing

1. Rebuild site
2. Restart webserver container
3. Navigate to `/restricted.html`
4. Bonus - test on the dev server, too

## 📸 Screenshot

image

### DIFF
--- a/apps/web/public/restricted.html
+++ b/apps/web/public/restricted.html
@@ -1,0 +1,6 @@
+<!DOCTYPE html>
+<html>
+  <title>Unauthorized | Non autorisé</title>
+  This page can only be accessed on the secure network. | Cette page n'est
+  accessible que sur le réseau sécurisé.
+</html>
--- a/apps/web/public/restricted.html
+++ b/apps/web/public/restricted.html
@@ -1,6 +1,14 @@
 <!DOCTYPE html>
 <html>
-  <title>Unauthorized | Non autorisé</title>
-  This page can only be accessed on the secure network. | Cette page n'est
-  accessible que sur le réseau sécurisé.
+  <head>
+    <title>Unauthorized | Non autorisé</title>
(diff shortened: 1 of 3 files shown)


esizer (changes requested):
Works well 👍 Just added a suggestion to update to valid HTML.

esizer:
> I originally went for the smallest valid page possible (according to [the w3 validator!](https://validator.w3.org/)) but I guess now that we're not going to be storing it on the reverse proxy itself there's no point in trying to shrink it as much as possible. Thanks for the update.

Lol I guess I can't argue with w3 but it seems odd that they would say it was valid when [the HTML spec does not](https://html.spec.whatwg.org/multipage/semantics.html#the-title-element:~:text=In%20a%20head%20element%20containing%20no%20other%20title%20elements.) as well as [their own documentation of the spec](https://www.w3.org/TR/html4/struct/global.html#edef-TITLE:~:text=Every%20HTML%20document%20must%20have%20a%20TITLE%20element%20in%20the%20HEAD%20section.)   😆

esizer (approved):

PR #6643, opened 2023-05-18
[Debt] Update application links

### LINKED ISSUE #6609: Send users to new application
When the Application Revamp flag is true, then buttons to Apply to an Application, Continue an Application, or View a complete application, should take the user to the new revamped Application.

## Proposed Implementation
- change CreateApplicationPage to navigate to the new path after creating the application
- change the ContinueButton in PoolAdvertismentPage to the new link
- change the links in ApplicationCard
- Ensure that we always use the `/applications/:id` path, which figures out itself which step the user should be sent to 

## Acceptance Criteria
- [ ] If featureFlags.applicationRevamp is false, navigation is unchanged
- [ ] If featureFlags is true, then anywhere the user would have been navigated to the old application (`useRoutes().reviewApplication`), they're instead sent to the new application (`/applications/:applicationId`)

esizer:
🤖 Resolves #6609 

## 👋 Introduction

Updates links to application to point to the new application workflow if feature flag is enabled.

## 🕵️ Details

There are some `reviewApplication` left, however these only exist within the current application workflow. Not sure if makes sense for the profile forms to return to the new application workflow if you got there from the exiting one 🤔 

## 🧪 Testing

1. Search for `reviewApplication` and make sure it only exists within the old application workflow
2. Build `npm run build:fresh`
3. Navigate to "My applications"
4. Confirm all links to applications go to new application workflow
5. Navigate to "My dashboard"
6. Repeat step 4

### DIFF
--- a/apps/web/.env.example
+++ b/apps/web/.env.example
@@ -10,7 +10,7 @@ OAUTH_LOGOUT_URI="http://localhost:8000/oxauth/endsession"
 # Feature flags
 FEATURE_ONGOING_RECRUITMENTS=true
 FEATURE_APPLICANT_DASHBOARD=true
-FEATURE_APPLICATION_REVAMP=true
+FEATURE_APPLICATION_REVAMP=false
 FEATURE_PSAC_STRIKE=true
 
 # Logging - enable by setting the logging level. Eight severity levels from rfc5424.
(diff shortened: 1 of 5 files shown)


vd1992 (approved):
revamp = false
-> apply, continue, application card links point to old process, dashboard too

revamp = true
-> apply, continue, application card links point to new process, dashboard too

No, I think the flows should be kept separate, so no linking to the new workflow from within the old one. 

Alright, nice!
Just make sure to note maybe in the cypress application revamp issue to revert the `.env.example` change

PR #6665, opened 2023-05-19
[Bug] Add missing null states to view pool page

### LINKED ISSUE #6306: 🐛 Null state for sections on view pool page
## 🐛 Bug

It looks a bit odd to me to just have a blank space under sections, especially since some may be intentionally left empty and it isn't a case of something failing to load. 

## 🦋 Expected Behaviour

Null state for sections. 
Most are not going to stay empty, but some are optional and so may stay empty like asset skills and screening questions. 

## 📋 Steps to Reproduce

1. /en/admin/pools/:id
2. observe the blank spaces under sections with no data

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

Copy can be debated. 
Maybe something like a simple "None" underneath.
Like 
`pool?.screeningQuestions ? <QuestionsElement/> : <p>Not provided</p>`

## ✅ Acceptance Criteria

- [ ] Blank spaces no longer blank
- [ ] Placeholder copy: none but with brackets too: "(None)"

esizer:
🤖 Resolves #6306 

## 👋 Introduction

Adds missing null states to the admin view pool page.

## 🕵️ Details

This currently uses the `commonMessages.notProvided` message in favour of introducing a new string that says a similar thing. Currently, I am waiting to hear if this is acceptable from designers.

>**Note**
> This also lifts the skill family headings outside of the skill check to always display them so we can tell admins that they have no added them instead of hiding it completely.

## 🧪 Testing

1. Build `npm run build`
2. Login as admin and create an empty pool
3. Navigate to the the pool information page for that new pool
4. Confirm there are no empty null states but instead it says "Not provided"

## 📸 Screenshot
localhost_8000_en_admin_pools_7c5f8c1f-d215-4094-b348-f73a69aa9016

### DIFF
--- a/apps/web/src/pages/Pools/ViewPoolPage/ViewPoolPage.tsx
+++ b/apps/web/src/pages/Pools/ViewPoolPage/ViewPoolPage.tsx
@@ -49,6 +49,7 @@ export const ViewPool = ({ pool }: ViewPoolProps): JSX.Element => {
   const paths = useRoutes();
   const form = useForm();
   const [linkCopied, setLinkCopied] = React.useState<boolean>(false);
+  const notProvided = intl.formatMessage(commonMessages.notProvided);
 
   /** Reset link copied after 3 seconds */
   React.useEffect(() => {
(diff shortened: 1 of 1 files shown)


vd1992 (approved):
Empty sections show the message instead ✅

PR #6666, opened 2023-05-19
[Bug] Fix storybook component errors

### LINKED ISSUE #6595: 🐛 Pool Candidates Table storybook renders Unexpected Application Error
## 🐛 Bug

Pool Candidates Table renders table in storybook with **Unexpected Application Error**.

`TypeError: Cannot read properties of undefined (reading 'user')`

## 🦋 Expected Behaviour

Pool Candidates Table renders table in storybook without **Unexpected Application Error**.

## 🕵️ Details

## 📋 Steps to Reproduce

1. Run storybook
2. Navigate to **Pool Candidates Table**
3. Observe **Unexpected Application Error**

## 📸 Screenshots

Screen Shot 2023-05-15 at 10 31 34
Screen Shot 2023-05-15 at 10 33 46
Screen Shot 2023-05-15 at 10 33 19

## ✅ Acceptance Criteria

- [ ] Pool Candidates Table renders table in storybook without **Unexpected Application Error**
- [ ] Solved in the Review Resume story as well #6599

esizer:
🤖 Resolves #6595 

## 👋 Introduction

Updates the shape of args passed into 3 stories to fix errors with the data the components require.

## 🕵️ Details

We updated the shape of `poolCandidatesPaginated` but never updated the story, this fixes that along with adjusting the args for the reivew resume page story.

## 🧪 Testing

2. Confirm the following stories function as expected:
    - Pages / Application Revamp / Review Resume / No Experiences
    - Pages / Application Revamp / Review Resume / Has Experiences
    - Tables / Pool Candidates Table / Default

## 📸 Screenshot
Screenshot 2023-05-19 154425

### DIFF
--- a/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.stories.tsx
+++ b/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.stories.tsx
@@ -1,4 +1,5 @@
 import React from "react";
+import { faker } from "@faker-js/faker";
 import { ComponentStory, ComponentMeta } from "@storybook/react";
 
 import { fakePoolCandidates } from "@gc-digital-talent/fake-data";
@@ -6,6 +7,20 @@ import { fakePoolCandidates } from "@gc-digital-talent/fake-data";
 import PoolCandidatesTable from "./PoolCandidatesTable";
 
 const poolCandidateData = fakePoolCandidates();
+
+const mockPoolCandidatesWithSkillCount = poolCandidateData.map(
+  (poolCandidate) => {
+    const skillCount = faker.datatype.number({
+      min: 0,
+      max: 10,
+    });
+    return {
+      id: poolCandidate.id,
+      poolCandidate,
+      skillCount: skillCount || null,
+    };
+  },
+);
 
 const mockPaginatorInfo = {
   count: 1,
@@ -27,7 +42,7 @@ export default {
       GetPoolCandidatesPaginated: {
         data: {
           poolCandidatesPaginated: {
-            data: poolCandidateData,
+            data: mockPoolCandidatesWithSkillCount,
             paginatorInfo: mockPaginatorInfo,
           },
         },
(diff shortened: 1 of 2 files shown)


apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.stories.tsx
```diff
@@ -7,6 +8,20 @@ import PoolCandidatesTable from "./PoolCandidatesTable";
 
 const poolCandidateData = fakePoolCandidates();
 
+const mockPoolCandidatesWithSkillCount = poolCandidateData.map(
+  (poolCandidate) => {
+    const skillCount = faker.datatype.number({
+      min: 0,
+      max: 10,
+    });
```
vd1992, line 16: Won't this result in a random skillCount number for each Chromatic run resulting in diffs every time?
My suggestion would be the same number for all for simplicity, or array index number = skillCount maybe since the array will always be the same.

vd1992 (changes requested):

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.stories.tsx
esizer: Hmm, that is a good point 🤔 I thought that the seed ensured the same number every time since [we are using it elsewhere](https://github.com/search?q=repo%3AGCTC-NTGC%2Fgc-digital-talent%20faker.datatype.number&type=code) and it seems stable.

vd1992: I think that requires this line to be called first somewhere in the file

https://github.com/GCTC-NTGC/gc-digital-talent/blob/53cdfcd3e0898ccdc25eb88b2ccc1531dc587bf3/packages/fake-data/src/fakePoolAdvertisements.ts#L69

vd1992: Hm, guess you don't need to declare seed 0 in the story file. Locally it matches with the web version. 

image

vd1992 (approved):
Story issues resolved for the web published, as well as locally. Above concern about shifting diffs appears unfounded. 
👍

PR #6667, opened 2023-05-19
[Bug] Prevent matchStringCaseDiacriticInsensitive() from crashing for enormous strings

### LINKED ISSUE #6031: Guard search against long strings
## 🐛 Bug

The app crashes when a long string is entered into search input.

## 🦋 Expected Behaviour

App should not crash and should not allow such a long string to be entered.

## 📋 Steps to Reproduce

1. Generate a long string (e.g. 60 paragraphs worth of [lorem ipsum](https://www.lipsum.com/))
2. Navigate to https://talent.canada.ca/en/search
3. Paste long string into field with label **Search skills by keyword**

## 📸 Screenshots
Screen Shot 2023-03-22 at 13 17 20

Screen Shot 2023-03-22 at 13 17 43

## 🙋‍♀️ Proposed Solution

Limit inputs to certain amount of characters to avoid app crashing on the front end. For example, using the [`maxlength`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength) attribute for `input` and `textarea`. Back end validation might also be worthwhile.

## ✅ Acceptance Criteria

- [ ] inputs going to matchStringCaseDiacriticInsensitive have max length validation
- [ ] matchStringCaseDiacriticInsensitive has a guard against long strings

vd1992:
🤖 Resolves #6031 

## 👋 Introduction

Prevent crashing of the function, and subsequently the page if a very, very long string is input. 

## 🕵️ Details

The function is only used in one place, the `Skillpicker`, and it is a rather edge case since it requires someone to go out of their way to intentionally crash their browser. 
Since the input doesn't actually submit anything, rather, just searches through a skills array in the frontend. I decided against adding restrictions to inputs, there was only one place where this function applied anyway. The input in question doesn't have a rules prop either. 

So I elected for the simple cap to short-circuit the function in cases where the string is 1000 characters or more. 

## 🧪 Testing

1. Navigate to `/en/search`
2. Head to the skills selection section
3. Paste a very massive string into the search, page should not crash

## 📸 Screenshot

Only used in one component. 
There is another function that uses it, matchString**_s_**CaseDiacriticInsensitive(), that isn't used anywhere. 

image

image

image

### DIFF
--- a/packages/forms/src/utils.ts
+++ b/packages/forms/src/utils.ts
@@ -99,6 +99,10 @@ export function matchStringCaseDiacriticInsensitive(
   needle: string,
   compareString: string,
 ) {
+  if (needle.length > 1000) {
+    // short-circuit for very long needle cases, prevents RegExp crashing
+    return false;
+  }
   const escapedNeedle = escapeAString(needle); // escape certain characters for Regex purposes
   return (
     compareString
--- a/packages/forms/src/utils.ts
+++ b/packages/forms/src/utils.ts
@@ -3,6 +3,7 @@ import { IntlShape } from "react-intl";
 import { LocalizedString, Maybe, Scalars } from "@gc-digital-talent/graphql";
 import { getLocale } from "@gc-digital-talent/i18n";
 import { getId, notEmpty } from "@gc-digital-talent/helpers";
+import { defaultLogger } from "@gc-digital-talent/logger";
 
 /**
  * Filters out empty data from data response.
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
Hmm, the logging doesn't seem to work - probably something from importing from another package.  Oh well, not important here.  Ship it!

vd1992:
> Hmm, the logging doesn't seem to work - probably something from importing from another package. Oh well, not important here. Ship it!

That is odd, worked on my end. I dunno, maybe a case of `npm run dev`  vs  `npm run build`

image

petertgiles:
> hat is odd, worked on my end.

Oh weird - must be a bad env for me.  That's great!

Issue #6109, opened 2023-03-29
Repoint Admin Pool Table query at root query


petertgiles:
## ♻️ Debt/Refactor

We would like to start relying on relationship scoping and partial results from our API instead of complex queries to ensure proper authorization.  Our Admin pool table query relies on navigating through the role assignment grqph.  Let's repoint it a simple root-level pool query and rely on the partial results.

## ✅ Acceptance Criteria

- [ ] Admin pool table uses a simple root-level query with partial results
- [ ] Verify correct partial-results are returned from the query based on auth

Issue #5486, opened 2023-01-26
🐜 Accessibility Bug: Unsaved changes warning & form labels being read together


RM-1978:
## ❌ Current behaviour (issue)
Unsaved changes warning are being read out with the subsequent form label in the Skills and experience forms. They seem to read together like one long line of code. 
### AT Configuration
 JAWS, Windows, Chrome
###Fable Video Clip
[Skills and Experience Form label issue](https://app.makeitfable.com/share/63d1b5a084cfb4002e8a3b2f)
### How to reproduce
1.	Open JAWS
2.	Navigate using your keyboard to the Edit Skills and Experience link in the application profile 
3.	Add some values in the inputs in the Work experience form fields so you get the “unsaved changes” warning
4.	Read the warnings and  form labels in successive order.
5.	Distinguish which label goes with what field.
## ✔ Expected behaviour (fix)
 Form labels are clearly detached for warning messages and associated with input fields. 
### Potential WCAG Considerations (if any): 
[3.3.2 Labels or Instructions,](https://www.w3.org/WAI/WCAG22/Understanding/labels-or-instructions)  [4.1.3 Status Messages](https://w3c.github.io/wcag/understanding/status-messages)
### Acceptance Criteria
- [ ] Warning messages are separated from subsequent form labels. 
- [ ] Each form label is associated with the correct input field.

petertgiles:
Thanks @RM-1978!
I took a screenshot of the word changes getting looped in with the label of the next form element.
![image](https://user-images.githubusercontent.com/8978655/230411561-168f48a1-1201-40e9-85b2-5901f6043f77.png)

To me, this looks like a bug in the AT, not a bug in the site that we can fix.  @esizer do you have an opinion on this?

The only workaround I can think of is we could change the message to simply "unsaved" so that it can't get wrapped to a second line.  Deepl tells me that the French would be "non sauvé" so that wouldn't really help.

esizer:
> To me, this looks like a bug in the AT, not a bug in the site that we can fix. @esizer do you have an opinion on this?
> 
> The only workaround I can think of is we could change the message to simply "unsaved" so that it can't get wrapped to a second line. Deepl tells me that the French would be "non sauvé" so that wouldn't really help.

Yeah, I'm not sure but this may have been solved. We will need to do some testing but we fixed an issue recently with how our messages are being associated with the input which may help.

RM-1978:
[...]

Issue #5359, opened 2023-01-17
⌨ Accessibility Issue: Unsaved Changes Interruptions


RM-1978:
## ❌ Current behaviour (issue)
The screen reader keeps flagging that a user has ‘unsaved changes’ each time they add information in a new field for every form they encounter. 

### How to reproduce
**AT configuration:** JAWS/NVDA Screen Reader, Windows 10 & 11, Chrome/Edge 

1.	Open JAWS/NVDA in Windows in Chrome or Edge.  
2.	Go to the About Me component of the Application profile.
3.	Using MS Narrator or Voice Over, enter text in any form field. 
4.	Navigate using your keyboard and fill in text field information while listening to the programmatic feedback to the screen reader: repeated “unsaved changes” statement. 
**Screen Capture** 
## ✔ Expected behaviour (fix)
Users indicate they only need to be advised that they have unsaved information once per page. Notify the user once at the end of the page with a list missing entries or unsaved changes. Perhaps, listing unsaved changes as they go through the UI in a prominent container somewhere on the page before they encounter the Save and Go Back button?

### Notes
Providing an error notification in the page title and an alert dialog box at the top of the page identifying the mandatory fields which may have been omitted or unsaved would improve their experience. 

**Potential WCAG Considerations (if any)**  
[4.1.3 Status Messages](https://w3c.github.io/wcag/understanding/status-messages),  [3.3.1 Error Identification ](https://www.w3.org/WAI/WCAG22/Understanding/error-identification) 

### Acceptance Criteria
A set of assumptions which, when tested, verify that the accessibility requirement was met.
- [ ] Criteria
- [ ] Criteria 2

petertgiles:
Thanks for this issue, Rosalie!  I'm trying to figure out the acceptance criteria for this one but it's a little unclear to me.  Is the issue to:
1. Make sure screen reader only says "unsaved changes" once
2. Bring back the unsaved changes component at the top of the page (see #4980)
3. Add "unsaved changes" to the page title
Or some combination of these?

RM-1978:
Hi Peter, thanks for the question!  I think the issue is the same  one flagged in  #4980 (see https://github.com/GCTC-NTGC/gc-digital-talent/issues/4980), with one addition: Make sure screen reader only says "unsaved changes" once when reloaded.

petertgiles:
Thanks, I've updated the acceptance criteria.  Feel free to tweak if it's not quite right yet.

[...]

Issue #4981, opened 2022-12-01
Make the "unsaved changes" field annotation less trigger-happy


patcon:
# Context
https://github.com/GCTC-NTGC/gc-digital-talent/issues/4831

Might be resolved with #4980, so using as dependency.

# 🧪 Proposed changes
- [ ] Verify only true changes trigger the unsaved effect, not empty fields that have been visited.
- [ ] Important to verify when adding skills during an application this doesn't trigger everything as well.

brindasasi:
is this story still relevant? 
Team raised concerns on "un saved changes" functionality. 
Wondering is this still in sprint to do list?

esizer:
> is this story still relevant?
> Team raised concerns on "un saved changes" functionality.
> Wondering is this still in sprint to do list?

I believe this is still relevant. The changes that were tentatively decided were:

1. Make alert appear on page load/submission only
2. Change the input messages to be hidden from assistive technology

petertgiles:
Gray to take this back to the designers for clarity on direction.

mnigh:
@gobyrne was there any clarity on direction for this issue? this issue is the only incomplete issue remaining for the epic #4832.

gobyrne:
This is resolved with the new application process. Closing

PR #6675, opened 2023-05-23
[Feature] Duplicate pool

### LINKED ISSUE #6546: ✨ Add "Duplicate pool" button
## ✨ Feature

A new button at the bottom of the "Edit pool" page that creates a duplicate pool.

## 🕵️ Details

This button creates a new pool, and pre-populates MOST of the information with the data from the origin pool.

The only data not copied over is:

- **Status**: the status of the duplicated pool should be set to 'Draft'
- **Closing date**: the closing date should be left empty
- **Specific title**: specific title should be specific title + "(copy)" in both english and french "(copie)" e.g: "React specialist (copy)"

When pressing the button, a dialog will open up to verify the action before proceeding.

Duplicating successfully will navigate the user to the "Edit pool" page of the newly created pool.

If there are unsaved changes on the pool, the duplicate should fail and inform the user of unsaved changes.

### TOASTS
Green toast notification for success:

`"Success: This pool has been duplicated successfully."`

Red toast notification for UNSAVED CHANGES error:

`"Error: You have unsaved changes on this pool, please save all field before duplicating this pool."`

Red toast notification for OTHER errors:

`"Error: Something went wrong, please try again in a minute or contact your administrator."`

## 🎨 Design File

## 📸 Screenshot

image

image

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] If there are unsaved changes on the page, duplicate button throws a toast saying you have to save everything first
- [ ] Duplicate Pool button creates a new pool which is a duplicate of current pool, except for status, closing date, and specific title (detailed above)
- [ ] If pool creation is successful, throw a toast and navigate to edit page of new pool
- [ ] If pool creation fails, show error toast

esizer:
🤖 Resolves #6546 

## 👋 Introduction

Adds the ability for pool operators to duplicate their pools

## 🕵️ Details

Adds a new mutation `duplicatePoolAdvertisement`. This needs to pass `team.id` along with the `pool.id` since the policy requires it in the args. I thought about making a new `duplicate` policy, does that make more sense?

```php
public function duplicate(User $user, $request)
{
  $existing = Pool::findOrFail($request["id"]);
  
  // Confirm the user can create pools for the team
  if ($user->isAbleTo("create-team-pool", $existing->team)) {
    return true;
  } else {
    return Response::deny("Cannot create a pool for that team.");
  }
}
```

## 🧪 Testing

2. Login as `pool@test.com`
3. Edit a pool
4. Scroll to the bottom
5. Confirm duplicate button exists
6. Confirm it opens and duplicates the pool
7. Login as `admin@test.com`
8. Edit a pool
9. Confirm the duplicate button does not appear (only pool operators can duplicate right now?)

## 📸 Screenshot

Screenshot 2023-05-23 151216


vd1992:
Microscopic issue I noticed only by logging in as `pool@test.com`, Pool Operator runs a different query at /admin/pools, it also needs the `updatedDate` field otherwise you get updated as null for all pools. `GetMePools` in this case. 
If you want to roll this in you can, otherwise I can split off an issue. 

image

image

apps/web/src/pages/Pools/EditPoolPage/components/StatusSection.tsx
vd1992, line 1: I think the bit from the acceptance criteria regarding handling for unsaved changes was missed.

image

esizer: So, I have no idea how this can easily be done. Seems like the same ask (that was too convoluted to properly implement) was made for the application profile page. The problem is, our current way of determiniing unsaved changes is flawed and has issues:

 - https://github.com/GCTC-NTGC/gc-digital-talent/issues/5359#issue-1536728561
 - https://github.com/GCTC-NTGC/gc-digital-talent/issues/4981#issue-1471634956
 - https://github.com/GCTC-NTGC/gc-digital-talent/issues/5486#issue-1557508841

Then we would need to do this for 8 forms at once to block an action when a user can easily make the changes after duplicating? 🤔 Also, seems strange this would be the only action we block with unsaved changed on this page instead of also blocking:

 - Publishing
 - Extending
 - Archiving
 - Closing
 - Deleting

esizer:
> Microscopic issue I noticed only by logging in as `pool@test.com`, Pool Operator runs a different query at /admin/pools, it also needs the `updatedDate` field otherwise you get updated as null for all pools. `GetMePools` in this case. I noticed this as I was looking to see what happened for sorting of duplicated pools.
> 
> If you want to roll this in you can, otherwise I can split off an issue. Just letting you know first in case you had any thoughts.

We are actually planning on removing this query entirely so I may just leave as is 😅 

https://github.com/gctc-ntgc/gc-digital-talent/pull/issues/6109

patcon:
cc @tristan-orourke for feedback on policies
cc @Jerryescandon for feedback on flow of unsaved fields during dup

Jerryescandon:
I guess we can go ahead without the unsaved changes toast, it seems like a bigger issue altogether.

esizer:
> I guess we can go ahead without the unsaved changes toast, it seems like a bigger issue altogether.

We could also look into applying this block to publishing as well 🤔

Jerryescandon:
> > I guess we can go ahead without the unsaved changes toast, it seems like a bigger issue altogether.
> 
> We could also look into applying this block to publishing as well 🤔

I had tasked Elisa before she left to [re-design this page](https://www.figma.com/proto/tH8X3HEgXsQVGTa2eheH8P/Pool-Updates---Admin-(elisa)?page-id=0%3A1&type=design&node-id=5034-52683&viewport=12510%2C3611%2C0.42&scaling=min-zoom&starting-point-node-id=5034%3A46129) using the profile section cards to resolve this problem a while ago.

tristan-orourke (changes requested):
The current method of replicating the pool only copies fields in the Pool table. This misses any data from relationships - skills and screening questions.

esizer:
> I like the plan to add a `duplicate` method to the PoolPolicy.

👍 Added in [add pool duplicate policy](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6675/commits/87ca0ecd0473cb95e77d468cadfac012688eae27)

> I don't quite understand what the issue is that @vd1992 found, but unless its a difficult fix lets try to resolve it. It may go away with #6109, but #6109 isn't in this sprint.

Pretty easy fix, added in [add updated at to pools index query](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6675/commits/bd1e14110a67f9a1150660e53324b282f590695b)

> The current method of replicating the pool only copies fields in the Pool table. This misses any data from relationships - skills and screening questions.

🤦‍♀️ Not sure where my head was at while doing this. I knew `$model->replicate()` did not copy relationships but forgot to add them. Added in [update replicate with relationships](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6675/commits/1d9ee59b127edbfa3e4d0e1e2819c435d7113cc9)

tristan-orourke (approved):
Looks good!

PR #6676, opened 2023-05-23
[Debt] Remove PoolCandidateFilterInput and associated changes

### LINKED ISSUE #5193: ♻️ Delete PoolCandidateFilterInput from the graphQL Schema
## ♻️ Debt/Refactor

Lets get rid of `PoolCandidateFilterInput` and any queries or mutations which references, to clean up our schema and avoid confusion.

## 🕵️ Details

All the queries/mutations which reference PoolCandidateFilterInput are already deprecated, or should be deprecated. It contains fields that it probably shouldn't, allowing filtering by personal contact info (only accessible to the public via a count query, but still). And since #5052 it overlaps confusingly with PoolCandidateSearchInput, which replaces it anywhere it was still useful. 

**Note:** PoolCandidateFilter (not -Input) must remain! It represents a table of data associated with our oldest search requests, and we want to keep that data around (at least for now). We no longer have any reason to _add_ to that data, hence removing PoolCandidatFilterInput.

## 🙋‍♀️ Proposed Solution

1. In the graphql schema, delete the following:
- PoolCandidateFilterInput
- countPoolCandidates(where: PoolCandidateFilterInput)
- searchPoolCandidates(where: PoolCandidateFilterInput ... )
- CreatePoolCandidateFilterInput
- PoolCandidateFilterBelongsTo
- the poolCandidateFilter field of CreatePoolCandidateSearchRequestInput

2. Modify the legacy SingleSearchRequest table, which still uses the searchPoolCandidates query. 

QUESTION for @gobyrne: This could go a few ways: 
- we do the best we can to adapt fields in the PoolCandidateFilter to the new PoolCandidateSearchInput
- or we don't even bother with a table, and just show some text saying "This is from a legacy version of requests which can longer be accurately represented with current filters." 
- OR... maybe we show our normal table, without any initialized filters, and text explaining this is a legacy request, and admins can do their best to adjust the filters to match the Summary of Filters shown, if they want.
Note: this should only ever appear for long-dealt-with requests, so unless DCM has a habit of going back to look at resolved requests, it shouldn't really matter.

## 🌎 Localization

TBD

## ✅ Acceptance Criteria

- [ ] PoolCandidateFilterInput deleted from schema
- [ ] Queries using PoolCandidateFilterInput deleted from schema
- [ ] Other inputs references PoolCandidateFilterInput are deleted or modified
- [ ] Any tests referencing these queries are removed
- [ ] Legacy SingleSearchRequestTable dealt with. QUESTION: how?

vd1992:
🤖 Resolves #5193

## 👋 Introduction

Removes the `PoolCandidateFilterInput` and related types, deprecated queries `countPoolCandidates` and `searchPoolCandidates`, testing for the types and queries involved, and rendering a special table for `PoolCandidateFilter` objects. 

## 🕵️ Details

Removing the above was mostly straightforward. Change to request table is that now if it is a `PoolCandidateFilter` instead of `ApplicantFilter`, it loads `<PoolCandidatesTable>` with an empty filter input. The specifics of the request are still on the page, just not loaded into the Fancy Filter by default. 

**NOTE**
The one change that stretched the scope a bit that I threw in, is in easily reversed commit https://github.com/GCTC-NTGC/gc-digital-talent/commit/49746f582bb31c5395489c3ac6a6ca01343e3706.
Basically, there were two expiry scopes present that do the same thing, I had to add the second in the past because the original was entangled in deprecated queries that are now removed. The new one is in a similar pattern to other scopes whereas the old one had a different structure, and due to being stuck in deprecated queries I could not update it without breaking deprecated queries. 
Now that the queries in question are being removed, I thought to sort out that bit and condense to one expiry scope.

## 🧪 Testing

1. Seeder still has `PoolCandidateFilter` requests, I think we should keep those for a bit for testing purposes
2. Navigate to `/en/admin/talent-requests`
3. Open requests of both filter types, observe page behaviour

## 📸 Screenshot

image

image

image

image


api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
```diff
@@ -35,7 +35,7 @@ public function __invoke($_, array $args)
         PoolCandidate::scopeAvailable($queryBuilder);
 
         // expiry status filter (filter active pool candidates)
-        PoolCandidate::scopeExpiryFilter($queryBuilder, ['expiryStatus' => ApiEnums::CANDIDATE_EXPIRY_FILTER_ACTIVE]);
+        PoolCandidate::scopeExpiryStatus($queryBuilder, ApiEnums::CANDIDATE_EXPIRY_FILTER_ACTIVE);
```
vd1992, line 38: Form of the other expiry scope is the same as the suspended scope, not an array of one key/value pair. 

image

api/graphql/schema.graphql
```diff
@@ -1498,10 +1451,7 @@ input CreatePoolCandidateSearchRequestInput {
   department: DepartmentBelongsTo!
   jobTitle: String! @rename(attribute: "job_title")
   additionalComments: String @rename(attribute: "additional_comments")
-  poolCandidateFilter: PoolCandidateFilterBelongsTo
-    @rules(apply: ["required_without:applicantFilter"]) # TODO: remove this field when we've completed the transition from using poolCandidateCount to applicantCount
-  applicantFilter: ApplicantFilterBelongsTo
-    @rules(apply: ["required_without:poolCandidateFilter"]) # TODO: make this required when we've completed the transition from using poolCandidateCount to applicantCount
+  applicantFilter: ApplicantFilterBelongsTo @rules(apply: ["required"])
```
vd1992, line 1454: Think this To Do is done too

tristan-orourke (changes requested):
Looks good! The only other thing I'd suggest is that we can probably stop creating PoolCandidateFilters in DatabaseSeeder and PoolCandidateSearchRequestFactory. We've confirmed that the page can still work with them, but I think they're mostly cluttering our dev environments now.

vd1992:
And now with passing tests!

PR #6682, opened 2023-05-25
[Debt] Remove PoolAdvertisement type

### LINKED ISSUE #6106: Remove PoolAdvertisement type and combine with Pool type
## ♻️ Debt/Refactor

We would like to start reducing the duplication in our schema and try relying on field-level permissions instead.  Let's remove PoolAdvertisement and rely on the Pool type to interact with this model.  The same auth restrictions should apply.

## ✅ Acceptance Criteria

- [ ] All fields from PoolAdvertisement available on Pool
- [ ] The PoolAdvertisement type is removed from the schema
- [ ] Auth restrictions remain at the same level
- [ ] Frontend queries repointed to the Pool type
- [ ] Tests are updated

JamesHuf:
🤖 Resolves #6106 

## 👋 Introduction

Removes the PoolAdvertisement type and combines it with Pool.

## 🕵️ Details

The fields `operationalRequirements`, `createdDate`, and `updatedDate` were on Pool but not PoolCandidate - I'm not sure if we want field level permissions for any of them.

## 🧪 Testing

1. Ensure all the tests are passing
2. Manually visit all pages that use Pools and ensure they still work as expected.


apps/web/src/pages/Applications/ApplicantDashboardPage/applicantOperations.graphql
petertgiles, line 1: A change somewhere broke the validation in the side menu.  Maybe the query doesn't pull in enough to do validation properly?
image

petertgiles (changes requested):
Wow, this is massive.  :exploding_head:   I appreciate all the little tidy-ups and the simpler type-wrangling.  I was only able to find one small regression bug.  Nice work!  :clap:

vd1992:
@JamesHuf 
Please merge main when you've got a chance, it appears the duplicate pools PR has introduced conflicts

apps/web/src/components/Table/ApiManagedTable/useFilterOptions.test.tsx
```diff
@@ -142,7 +142,7 @@ describe("useFilterOptions", () => {
       await waitFor(() =>
         expect(result.current.optionsData.pools).not.toBeUndefined(),
       );
-      expect(result.current.optionsData.pools).toHaveLength(2);
+      expect(result.current.optionsData.pools).toHaveLength(10);
```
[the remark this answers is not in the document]
vd1992: https://github.com/GCTC-NTGC/gc-digital-talent/pull/6682#discussion_r1209491721

packages/fake-data/src/fakePools.ts
```diff
@@ -31,36 +45,70 @@ const generatePool = (
...
+          users,
+          skills,
+          classifications,
+          "IT Apprenticeship Program for Indigenous Peoples",
+          "Programme d’apprentissage en TI pour les personnes autochtones",
+        );
+      default:
+        return generatePool(users, skills, classifications);
+    }
+  });
```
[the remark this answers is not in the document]
vd1992: Git difference does not show the pool faking above it

image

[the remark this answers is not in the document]
petertgiles: Ah, that makes sense, thanks.

apps/web/src/pages/Applications/ApplicantDashboardPage/applicantOperations.graphql
JamesHuf: It looks like this doesn't check the pool at all. We should probably we make a new issue for this if we want it changed but I don't think it's used like this once the application revamp is live anyway.
image

petertgiles: Oh, you're right - this issue preexists on main.  Also agree that it's probably not worth fixing on a deprecated page.

petertgiles (approved):

api/graphql/schema.graphql
```diff
@@ -466,10 +448,7 @@ type ScreeningQuestionResponse {
 
 type PoolCandidate {
   id: ID!
-  pool: Pool! @belongsTo @can(ability: "view", resolved: true)
-  poolAdvertisement: PoolAdvertisement
-    @belongsTo(relation: "pool")
-    @can(ability: "viewAdvertisement", resolved: true)
+  pool: Pool! @belongsTo @can(ability: "viewAdvertisement", resolved: true)
```
tristan-orourke, line 451: I feel like if we are getting rid of the PoolAdvertisment type, we can also get rid of or rename our "advertisement" permissions and policy functions. Like here, I think `ability: "view"` should work equally well

tristan-orourke (changes requested):
Looks good, besides one point in `type ScreeningQuestion`. 

It looks like you tried to make as few changes to the schema as necessary while ditching the PoolAdvertisment type, which is a good instinct, but I wonder if we want to be even more ruthless and remove any mention of PoolAdvertisments alltogether, to avoid future confusion? To avoid the question in the future of "why is this called 'pool' here and 'poolAdvertisment' somewhere else".

Like we can rename the `poolAdvertisements: [Pool]!` query to `pools: [Pool]!`. Remove the current `updatePool` mutation (which is deprecated) and then rename `updatePoolAdvertisment` to `updatePool`. Rename our input types like `UpdatePoolAdvertisementInput`.

Likewise, we can remove `viewAdvertisement` from PoolPolicy, since it's redundant with `view`, and rename `viewAnyPublishedAdvertisement` and `closePoolAdvertisment`. We can also rename our `view-any-publishedPoolAdvertisment` permission to `view-any-publishedPool`.

api/graphql/schema.graphql
```diff
@@ -321,7 +321,7 @@ type UserPublicProfile {
 
 type ScreeningQuestion {
   id: ID!
-  poolAdvertisement: PoolAdvertisement @belongsTo(relation: "pool")
+  poolAdvertisement: Pool @belongsTo
```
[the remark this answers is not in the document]
JamesHuf: This was addressed as part of [2ff0e2d](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6682/commits/2ff0e2d8b691a0c224bcc3bd0b79258d2c7c54ee).

api/graphql/schema.graphql
JamesHuf: This was addressed as part of [2ff0e2d](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6682/commits/2ff0e2d8b691a0c224bcc3bd0b79258d2c7c54ee).

JamesHuf:
Ok, I have now gone through and removed the majority of references to advertisements. This should be ready for re-review.

tristan-orourke (approved):
Thanks @JamesHuf! This is really thorough!

PR #6708, opened 2023-05-25
[Feature] Adjust setup script for CI

### LINKED ISSUE #5592: Separate CI Setup.sh
## ✨ Feature

Right now we have a single `setup.sh` script that was written to help developers do a first-time setup of their development environments.  Later, it started being used to run CI in Github Actions, too.  However we're starting to see some diverging needs in the setup.

## 🕵️ Details

- Devs like to see a large amount of random, realistic data in their database.  PHPUnit needs no seeded data at all.  Cypress needs a minimal amount of dimensional data only.  Our full seeders can take a lot of time to run.
- Devs want a dev build from Webpack to get the most amount of debugging and sourcemap data.  Cypress errors out on console errors from a dev build.

## ❓️ Questions (and Answers)
-    Is it a risk to run a different setup locally and in CI?
    -    there's the risk of similar code diverging where it shouldn't, but that's not worse than what we have now
 -   Are there other opportunities to optimize CI with its own setup other than the seeders?
    -    seeders
     -   building in dev mode vs production
  -  Should we have separate scripts or change the behaviour of a single script with option parameters?
     -   Let's try one script with a flag?
  -  Can we bring deploy.sh into alignment with these, too?
     -   Let's not worry about deploy.sh right now

## ✅ Acceptance Criteria

- [ ] setup.sh has a flag to run in CI mode
- [ ] dev mode runs all the regular seeders and builds the app in Webpack dev mode
- [ ] CI mode runs minimal seeders (maybe uat or prod seeder?) and builds the app in Webpack prod mode
- [ ] VS Code build tasks run setup in dev mode

esizer:
🤖 Resolves #5592 

## 👋 Introduction

Adds an option to the `setup.sh` script to run different scripts in CI.

### Usage

You can no pass options to the command, you cna 

```sh
$ docker-compose run --rm maintenance bash setup.sh -h

Setup the project and your environment.

Syntax: setup.sh [-h|c]
   -h   Show this help message.
   -c   Setup the environment for CI.
```

## 🧪 Testing

1. Run `docker-compose run --rm maintenance bash setup.sh`
2. Confirm it setups the env as usual but in dev mode
3. Run `docker-compose run --rm maintenance bash setup.sh -c`
4. Confirm minimal seeders are run (`CiSeeder.php`) and app is build in prod mode

### DIFF
--- a/api/database/seeders/CiSeeder.php
+++ b/api/database/seeders/CiSeeder.php
@@ -15,6 +15,9 @@ class CiSeeder extends Seeder
         $this->call(RolePermissionSeeder::class);
         $this->call(ClassificationSeeder::class);
         $this->call(DepartmentSeeder::class);
+        $this->call(GenericJobTitleSeeder::class);
+        $this->call(SkillFamilySeeder::class);
+        $this->call(SkillSeeder::class);
         $this->call(TeamSeederLocal::class);
         $this->call(UserSeederLocal::class);
     }
--- a/api/database/seeders/CiSeeder.php
+++ b/api/database/seeders/CiSeeder.php
@@ -15,6 +15,7 @@ class CiSeeder extends Seeder
         $this->call(RolePermissionSeeder::class);
         $this->call(ClassificationSeeder::class);
         $this->call(DepartmentSeeder::class);
-        $this->call(TeamSeeder::class);
+        $this->call(TeamSeederLocal::class);
+        $this->call(UserSeederLocal::class);
     }
 }
--- a/api/database/seeders/CiSeeder.php
+++ b/api/database/seeders/CiSeeder.php
@@ -2,7 +2,6 @@
 
 namespace Database\Seeders;
 
-use Illuminate\Database\Console\Seeds\WithoutModelEvents;
 use Illuminate\Database\Seeder;
 
 class CiSeeder extends Seeder
--- a/api/database/seeders/CiSeeder.php
+++ b/api/database/seeders/CiSeeder.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Database\Seeders;
+
+use Illuminate\Database\Console\Seeds\WithoutModelEvents;
+use Illuminate\Database\Seeder;
+
+class CiSeeder extends Seeder
+{
+    /**
+     * Run the database seeds.
+     */
+    public function run(): void
+    {
+        $this->call(RolePermissionSeeder::class);
+        $this->call(ClassificationSeeder::class);
+        $this->call(DepartmentSeeder::class);
+        $this->call(TeamSeeder::class);
+    }
+}
--- a/maintenance/scripts/setup.sh
+++ b/maintenance/scripts/setup.sh
@@ -10,7 +10,7 @@ Help()
   echo
 }
 
-CI=false
+GCDT_CI=false
 
 while getopts ":hc" option; do
   case $option in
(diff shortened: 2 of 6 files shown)


maintenance/scripts/setup.sh
```diff
@@ -1,5 +1,30 @@
...
+
+CI=false
+
+while getopts ":hc" option; do
+  case $option in
+    h | help) # display Help
+      Help
+      exit;;
+    c | ci)
+      CI=true;;
```
patcon, line 21: I think we should prefix these internal vars with our own prefix to avoid collision with envvars that might be set outside script. For example, CI=true is always set in GitHub Actions and many CI environments:
https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

This might drive someone bonkers if they ever want to unset this on CI for some sanity check or something :)

patcon: Ah wait, I'm seeing that `CI` system value is being overridden with `false` before, but honestly, best to just use a prefix and not step on the toes of the CI environment, because CI is the var used in every system since TravisCI. (and we're not using this envvar as it was intended to detect system automatically, but rather as a var to store a script flag)

esizer: Thanks, I was not aware that  github actions already used that var. Prefixed in [prefix CI var](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6708/commits/90dcbb3290d091c60d6e25ee11ab62505eb22cbd)

maintenance/scripts/setup.sh
```diff
@@ -1,5 +1,30 @@
...
+  echo "   -h   Show this help message."
+  echo "   -c   Setup the environment for CI."
+  echo
+}
+
+CI=false
+
+while getopts ":hc" option; do
+  case $option in
+    h | help) # display Help
```
patcon, line 17: I put this through shellcheck and i don't think there's a long-form option for getopts -- maybe you're mixing up conventions for shell's getopts vs perl getopt?
https://stackoverflow.com/questions/402377/using-getopts-to-process-long-and-short-command-line-options

patcon:
Re: shellcheck. we could seemingly have it run on every PR quite easily: https://github.com/reviewdog/action-shellcheck

I'll create an issue for it!

maintenance/scripts/setup.sh
esizer: I'm a little confused by the comment 😕 This is just allowing people to spell help out if they prefer to be more verbose

```sh
docker-compose run --rm maintenance bash setup.sh -help
```

patcon:
Hm, I'm not seeing the frontend app being served after running `docker-compose run --rm maintenance bash setup.sh`, but maybe my environment is wonky. sorting it now

mnigh (approved):
Small PHP linting thing (I think) but otherwise, LGTM 👍 

✅ it setups the env as usual but in dev mode
✅ minimal seeders are run (CiSeeder.php) and app is build in prod mode

api/database/seeders/CiSeeder.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Database\Seeders;
+
+use Illuminate\Database\Console\Seeds\WithoutModelEvents;
```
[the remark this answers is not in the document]
esizer: Oops, dumb generator adding unused references 😅 Removed in [remove unused import](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6708/commits/15705d94b3c4e70771117bdfe903a705262013f6)

PR #6727, opened 2023-05-26
[Bug] Add charset to restricted page

### LINKED ISSUE #6705: 🐛 Character set issue on restricted page
## 🐛 Bug

The new restricted page has some character set issues that were missed during testing.

## 🦋 Expected Behaviour

The accented characters should render properly in the browser.

## 📋 Steps to Reproduce

1. Navigate to `/restricted.html`

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

Probably just need to add `    <meta charset="UTF-8">` to the head.

## ✅ Acceptance Criteria

- [ ] Accented characters in French message render properly in the browser.

petertgiles:
🤖 Resolves #6705 

## 👋 Introduction

This branch adds the charset to the head of the restricted page so that the accented characters are rendered properly.

## 🧪 Testing

1. Build the app or just manually copy the page from public to dist
2. Visit /restricted.html

## 📸 Screenshot

image

### DIFF
--- a/apps/web/public/restricted.html
+++ b/apps/web/public/restricted.html
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
+    <meta charset="UTF-8" />
     <title>Unauthorized | Non autorisé</title>
   </head>
   <body>
@@ -11,4 +12,3 @@
     </p>
   </body>
 </html>
-


apps/web/public/restricted.html
```diff
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
+    <meta charset="UTF-8" />
```
mnigh, line 4: See: https://validator.w3.org/nu/#textarea

Trailing slash on void elements [has no effect](https://github.com/validator/validator/wiki/Markup-%C2%BB-Void-elements#trailing-slashes-in-void-element-start-tags-do-not-mark-the-start-tags-as-self-closing) and [interacts badly with unquoted attribute values](https://github.com/validator/validator/wiki/Markup-%C2%BB-Void-elements#trailing-slashes-directly-preceded-by-unquoted-attribute-values).

```suggestion
    <meta charset="UTF-8">
```

mnigh (changes requested):

apps/web/public/restricted.html
petertgiles: Yeah, I saw that, too.   It's actually an info message, not an error.  VS Code automatically adds that on save so it will be automatically changed back every time that page is interacted with unless we want to change linting rules.  According to the validator help link "my formatting tools adds it" and "I work with JSX" are valid reasons for adding the slash.  Along with, apparently, "I like how it looks"?  :laughing:

mnigh (approved):

PR #6736, opened 2023-05-26
[Feature] Cypress calls Artisan command

### LINKED ISSUE #6581: Test out Cypress calling out to Laravel commands
### Discussed in https://github.com/GCTC-NTGC/gc-digital-talent/discussions/6239

<div type='discussions-op-text'>

<sup>Originally posted by **petertgiles** April 11, 2023</sup>
It was already getting pretty awkward to set up Cypress tests in situations like needing a new application.  The steps are something like:
1. Admin creates pool advertisement
2. Applicant creates draft application
3. Applicant submits application
4. Admin accepts application

This is going to get even more tricky as we split the ability to do actions up in to an increasing number of roles.

It would be much easier to set this up with Eloquent models directly like we do with PHPUnit tests instead of having to work around all the workflow limitations of the GraphQL API.  I know we've worked with Artisan commands before that could do this.  And Cypress [has the ability to call shell commands](https://docs.cypress.io/api/commands/exec).  Database seeding is one of the example uses listed on their doc page.  Perhaps this could make our Cypress tests simpler?</div>

## Acceptance Critera
- [ ] Build out a single use-case: the new application setup described in the discussion
  - [ ] Create a custom artisan command to create a new application
  - [ ] Update Cypress helper command to use this artisan command
- [ ] Evaluate if this approach is actually superior

petertgiles:
🤖 Resolves #6581 

## 👋 Introduction

This branch demonstrates how Cypress can call Artisan commands to (hopefully!) simplify test setup.  I'm actually not convinced this is much simpler. :cry: 
image

## 🕵️ Details

A few things to note:
- The changes around `dcmId` is a bug fix for a race condition that appeared while making changes.
- I've added states to the pool factory to make it easier to use the Artisan command
- The JSON object that is returned from the command is in snake case, not Pascal case like GraphQL
- The pool factory has been changed to try to reuse existing users and teams to try to speed it up a bit

## 🧪 Testing

1. PHPunit tests pass
2. Reseed the database
3. Cypress tests pass
4. Decide whether this is worth doing.


api/app/Rules/PoolClosed.php
```diff
@@ -30,7 +30,7 @@ public function passes($attribute, $value)
     {
         $pool = Pool::find($value);
 
-        return $pool->closing_date->isFuture();
+        return is_null($pool->closing_date) || $pool->closing_date->isFuture();
```
petertgiles, line 33: This file is named backwards, right?  It's a rule that the pool is NOT closed?

esizer: Oops! 😅 Yeah, that was my mistake. You are correct, `PoolNotClosed` is what it should be.

petertgiles: Opened #6770

api/database/factories/PoolFactory.php
```diff
@@ -28,25 +28,27 @@ class PoolFactory extends Factory
      */
     public function definition()
     {
+        $userId = User::inRandomOrder()
```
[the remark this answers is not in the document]
petertgiles: Yeah, this will make the seeded data a little more realistic.
[pool owner is admin](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6736/commits/16d7166ee9dac2459f4a7949897ef2a3db7104ea)

petertgiles:
OK, I've updated it to target this new scope.
>     Instead of having to write a command wrapping each factory, we make a single generic factory command, which takes--class, --state and --attributes args. The latter would have to be a named array, or json object?

Yeah, I like this a lot better, especially if we're going only use this occasionally.

>     To get around the awkwardness of the returned object now matching the graphQL type, maybe we just return the new id, and let Cypress query for the object via API if it needs more details?

I'm not sure I want to return less just because the field names will be different.  I wrote a new Cypress test to demonstrate how to use the command with an example and some commenting to hopefully make it clear what's coming back.

apps/e2e/cypress/e2e/api/artisan-commands.cy.js
```diff
@@ -0,0 +1,33 @@
...
+    ).then((jsonObject) => {
+      // fields in JSON object from Eloquent model are snake case, not pascal case
+      cy.wrap(jsonObject.id).as("poolId");
+      cy.wrap(jsonObject.name.en).as("poolName");
+      cy.wrap(jsonObject.key_tasks.en).as("poolKeyTasks");
+    });
+
+    cy.get("@poolId").then((poolId) => {
+      cy.visit(`/en/browse/pools/${poolId}`);
+    });
```
tristan-orourke, line 20: If this test is just to confirm that the factory created a pool correctly, it would probably be faster to confirm this with an api call, rather than a visit to a page.

tristan-orourke (changes requested):
Great job! I love the new CallFactory command, and having a cypress test for it which also illustrates its use is a great idea. My only suggestion is to test the factory with an api call instead of page visit, which should run a bit faster.

apps/e2e/cypress/e2e/api/artisan-commands.cy.js
petertgiles: Good idea!
ref: [just use API for test](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6736/commits/63d675662c0c7f265f62fd4ea0737827eeba2b1c)

api/app/Console/Commands/CallFactory.php
```diff
@@ -0,0 +1,58 @@
...
+            $attributes = json_decode($jsonAttributes, true);
+        } else {
+            $attributes = null;
+        }
+
+        $state = $this->option('state');
+        if (empty($state)) {
+            $result = $qualifiedClassName::factory()->create($attributes);
+        } else {
+            $result = $qualifiedClassName::factory()->$state()->create($attributes);
```
esizer, line 54: I'm wondering, could we make state accept more than one item?

> **Note**
> I have no idea if this will work 😛 

```php
$state = $this->option('state');
if(empty($state)) {
  ...
} else {
  $states = explode(",", $state);
  $factoryWithState = $qualifiedClassName::factory();
  foreach($states as $stateMethod) {
     $factoryWithState = $factoryWithState->$stateMethod();
  }
  $result = $factoryWithState->create($attributes);
}
```

So then I could call:

```sh
php artisan app:call-factory User --state=asApplicant,withExperiences,withSkills
```

esizer (changes requested):
This is very cool and works well. I had another suggestion I noticed about the state. It seems we can only pass a single state which could be a problem for some models (specifically users)

api/app/Console/Commands/CallFactory.php
petertgiles: Artisan commands accept arguments with more than one value natively so we wouldn't have to do the explode.  I'll try it out!

esizer: Good because I don't like that function name 😝

petertgiles: Cool, it works!
image
ref: [allow multiple states](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6736/commits/cdb884e23199d1deb9022953d6a5ec83be25c895)

tristan-orourke (approved):
Nice job, looks great!

esizer (approved):

PR #6748, opened 2023-05-29
[Tooling] Cypress test application revamp

### LINKED ISSUE #6078: Cypress test for application revamp
## 🧪 Tests

We should have a Cypress test that works through the new application process in the same way that a real user would, step-by-step, to avoid regression bugs in the future.

## ✅ Acceptance Criteria

- [ ] A Cypress test that works through the new application process, step-by-step
- [ ] The test exercises all the major sections of the flow
- [ ] The test is reliable and does not depend on the pre-existing state of the database.
- [ ] in scope
    - [ ] clicking stepper to jump ahead
    - [ ] hacking url to get jump ahead
- [ ] revert env change as seen in comment below
- [ ] not in scope
    - [ ] ?

vd1992:
🤖 Resolves #6078

## 👋 Introduction

Adds a Cypress test for the application revamp, replacing the old flow.

## 🕵️ Details

Things made easier by being able to recycle the set-up work from the prior application test.
Flags reverted to true and old test deleted. 

Had to make a tweak to a graphQL file as it was missing fields for the review page. Compressed that bit into a fragment. 
Need to condense API operations to a single-point to avoid drift and missing fields someday, as an aside. Case in point, the snippet I was editing was missing `details` too.

## 🧪 Testing

1. Tests pass in CI
2. Pass locally too
3. Own discretion, anything overlooked?


vd1992:
Not sure how to access the date picker, which we intend to replace soon I think, in Cypress. It isn't accessible. 
The checkbox for whether it is a current experience also seems iffy. 
Without solving the two issues above, I can't exercise adding an experience in the application. 

What appears for the two in Cypress with an element inspector. 

image

image

Regular elements that are easy to interact with Cypress look like this

image

esizer:
> Not sure how to access the date picker, which we intend to replace soon I think, in Cypress. It isn't accessible.

Yeah, `input[type=date]` has no default role so we can't target it properly in tests 👎 The next best way to target date inputs is using label text.

```tsx
cy.findByLabelText(/start date/i})
```

> The checkbox for whether it is a current experience also seems iffy.

I just took a look and using `<Checkbox boundingBox />` results in some not-so great HTML where we use the text on top of the bounding box and the text next to the checkbox is just static text not associated with the input.

Screenshot 2023-05-30 091020

So you could target this with the following (even if it seems a bit odd):

```tsx
cy.findByRole("checkbox", { name: /current education/i });
```

apps/e2e/cypress/e2e/talentsearch/submit-applicationRevamp.cy.js
esizer, line 1: Would it be possible to keep the filename the same? `*Revamp` will be quickly outdated 😅

apps/e2e/cypress/e2e/talentsearch/submit-applicationRevamp.cy.js
```diff
@@ -0,0 +1,369 @@
...
+      name: /Uh oh, it looks like you jumped ahead!/i, // message for skipping steps
+    })
+      .should("exist")
+      .and("be.visible");
+    cy.findByRole("link", {
+      name: /Return to the last step I was working on/i,
+    }).click();
+    cy.findByRole("heading", { name: /Create your résumé/i }) // returned to resume step
+      .should("exist")
+      .and("be.visible");
```
esizer, line 192: Could this be a separate test case? Something like `it("prevents users from skipping steps")`

esizer (changes requested):
Looking good! Some minor suggestions.

I think this is out of scope but it would be nice to have a test for both a pool with no screening questions as well as one that is under the IAP publishing group too. Do you know if we have tickets for those yet?

vd1992:
This came to mind with the above comments, should definitely test the stepper somehow too. In the hypothetical IAP workflow, as well as this spec too https://github.com/GCTC-NTGC/gc-digital-talent/pull/6748/commits/b68f9cb36ddc87e16fc720b4567b7bc83ebdb1d5

esizer (approved):

apps/e2e/cypress/e2e/talentsearch/submit-applicationRevamp.cy.js
vd1992: Yeah... if you aren't sure it'd work might be best to experiment with it first. It wouldn't effect things that much I imagine it'll be low priority

Issue #6745, opened 2023-05-29
🐛 Application Revamp - Submitting IAP application wipes self declaration step


petertgiles:
## 🐛 Bug

It seems like submitting an IAP application using the application revamp clears the "SELF DECLARATION" step submitted and triggers the "uh oh" page on successful application.

## 🦋 Expected Behaviour

Submitted steps should not be adjusted on application submission.

## 📋 Steps to Reproduce

1. Start an IAP application
2. Work all the way through to the "Review and submit" step
3. Submit the application
4. Observer and "Uh oh" page instead of the "success" page

## 📸 Screenshot

https://github.com/GCTC-NTGC/gc-digital-talent/assets/8978655/6c938e41-f923-47a7-ac9b-5e8e4ee1d7f5

## ✅ Acceptance Criteria

- [ ] Application submission does not change submitted steps
- [ ] Application submission can handle dynamically optionally steps
- [ ] IAP application can be fully submitted

PR #6750, opened 2023-05-30
[Feature] IAP dashboard notifications for application revamp

### LINKED ISSUE #6485: ✨ Links from IAP Application to Dashboard trigger special notifications
## ✨ Feature

Users get to the the IAP Application from the IAP Landing page. That means that when they get to the the Dashboard it's a totally new UI, even with a new theme. We want to add a notification to the top of the page when they arrive on the Dashboard to help contextualize the change for the user.

## 🕵️ Details

The notification should be slightly different depending on whether they clicked "Save and quit" or whether they got there after submitting the application.

## 🎨 Design File

Note: the notification should appear at the very top of the page, like https://github.com/GCTC-NTGC/gc-digital-talent/issues/6485#issuecomment-1533294169

[Dashboard, after submitting IAP Application](https://www.figma.com/file/8piNhFaBjcxssBB8LAGxEY/IAP-User-flows?node-id=269-78632&t=FM1b6mI8HeHY19YM-0)

[Dashboard, after leaving in-progress application](https://www.figma.com/file/8piNhFaBjcxssBB8LAGxEY/IAP-User-flows?type=design&node-id=269-76379&t=fgKSYtT4BVSNN2CW-0#436224345)

## 📸 Screenshot

Notification for after submitting application
image

Notification for Save and quit
image

I propose adding a url parameter when linking to the Dashboard, and checking for it on the Dashboard page. We could use two for the two possible notifications, possibly `from_iap=true` and `from_iap_success=true`.

If for some reason we don't want to show this in the url, we could put something in the [history state](https://developer.mozilla.org/en-US/docs/Web/API/History/state) when navigating, but this is less transparent and therefore more difficult to maintain.

## 🌎 Localization

The two notifications differ only by title.

| String | En | Fr |
|--------|----|----|
| Success Title | Thanks for applying to the IT Apprenticeship Program for Indigenous Peoples! | ??? |
| In Progress title | Your draft application for the IT Apprenticeship Program for Indigenous Peoples has been successfully saved! | ??? |
| Paragraph 1 | You’ve now landed on your Digital Talent profile page. | ??? |
| Paragraph 2 | The apprenticeship program is a part of a larger digital talent initiative in the Government of Canada that has the hopes of recruiting awesome digital talent like yourself. This profile provides you access to other digital opportunities and recruitments beyond the apprenticeship program, if you choose to apply. If you’d prefer to stick to the apprenticeship program exclusively, that’s okay too! | ??? |
| Paragraph 3 | When you log into the account you created, you’ll be taken to this page from now on. You can use this page to review your Indigenous Apprenticeship application and track your position in the program. | ??? |

## ✅ Acceptance Criteria

- [ ] When completing an IAP application, user is sent to Dashboard with a special notification at top of page
- [ ] When saving and quiting IAP application, user is sent to Dashboard with a different special notification

petertgiles:
🤖 Resolves #6485

## 👋 Introduction

This branch adds two notifications to the application dashboard when reached from an IAP application.

## 🧪 Testing

Ensure FEATURE_APPLICATION_REVAMP is set to true and rebuild the application.

- [ ] Visiting the dashboard normally triggers no notifications
- [ ] Returning to the dashboard from a non-IAP application triggers no notifications
- [ ] Returning to the dashboard from a page in an IAP application triggers the "IAP application draft saved" notification
- [ ] Returning to the dashboard from the "success" page in an IAP application triggers the "IAP application successful" notification

> __Warning__
It is currently not possible to get to the IAP success page without manipulating the database and URL.  I've submitted bug #6745 to fix it.

## 📸 Screenshot

image


apps/web/src/pages/Applications/ApplicantDashboardPage/components/DashboardHeading.tsx
```diff
@@ -100,6 +106,100 @@ const DashboardHeading = ({ user }: DashboardHeadingProps) => {
...
       })}
     >
+      {searchParams.get(FromIapDraftQueryKey) === "true" && (
+        <Alert.Root
+          type="info"
+          dismissible
+          onDismiss={() => {
+            searchParams.delete(FromIapDraftQueryKey);
+            setSearchParams(searchParams);
+          }}
```
esizer, line 116: This is subjective, but I'm leaning towards these not being live alerts. I don't think it would be helpful to interrupt the user with this information 🤔 

```suggestion
        <Alert.Root
          type="info"
          dismissible
          live={false}
          onDismiss={() => {
            searchParams.delete(FromIapDraftQueryKey);
            setSearchParams(searchParams);
          }}
```

apps/web/src/pages/Applications/ApplicantDashboardPage/components/DashboardHeading.tsx
```diff
@@ -100,6 +106,100 @@ const DashboardHeading = ({ user }: DashboardHeadingProps) => {
...
+      )}
+      {searchParams.get(FromIapSuccessQueryKey) === "true" && (
+        <Alert.Root
+          type="info"
+          dismissible
+          onDismiss={() => {
+            searchParams.delete(FromIapSuccessQueryKey);
+            setSearchParams(searchParams);
+          }}
+        >
```
esizer, line 164: Similar idea here with the `live={false}`. That is, if the team agrees these should not actually have `role="alert"` 😅

apps/web/src/pages/Applications/ApplicantDashboardPage/components/DashboardHeading.tsx
```diff
@@ -100,6 +106,100 @@ const DashboardHeading = ({ user }: DashboardHeadingProps) => {
         description: "Subtitle for applicant dashboard hero",
       })}
     >
+      {searchParams.get(FromIapDraftQueryKey) === "true" && (
+        <Alert.Root
+          type="info"
+          dismissible
+          onDismiss={() => {
+            searchParams.delete(FromIapDraftQueryKey);
+            setSearchParams(searchParams);
```
esizer, line 115: Should we add a `{replace: true}` as the second arg here? Right now if you hit the browser back button after dismissing the alert, it just shows up again. This may be desirable though 🤔

esizer (commented):
Looking great! Mostly just questions from me 😅

apps/web/src/pages/Applications/ApplicantDashboardPage/components/DashboardHeading.tsx
petertgiles: I didn't consider that this would add an entry to history.  I added the `replace` and I like how that works.  Honestly I'm not sure we need to mutate the search params at all.  I could be convinced to remove the onDismiss entirely.
Ref: [replace search state](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6750/commits/88bcd764e5c20cdede959910f90c6f09ca3579f0)

apps/web/src/pages/Applications/ApplicantDashboardPage/components/DashboardHeading.tsx
petertgiles: I don't have enough experience with AT to have an opinion on this so I'll take your suggestion.  :sweat_smile:  As far as I know, the main motivation for these notifications is the theme change which would not be as jarring to folks relying on narration, I imagine.
Ref: [remove live alert](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6750/commits/fd8567e2731c70204bf842b6d6e21cd792e8cb63)

esizer (approved):
Awesome! ⭐

PR #6756, opened 2023-05-30
[Debt] Enable revamp and test fixes

### LINKED ISSUE #6753: ♻️ Enable flags to true and resolve testing issues
## ♻️ Debt/Refactor

In the .env example files, revamp is set to false for set-up purposes. Set both to true. 

## 🕵️ Details

Switching to true is it for the flags. 
Tests break though, like any instance of `submitApplication` being used as now education validation is required for submitting. 

## ✅ Acceptance Criteria

- [ ] Revamp enabled in both environment issues
- [ ] Tests pass by refactoring to solve the validation issues
- [ ] Special cases, like the submit application workflow which will break in the revamp environment, marked as skipped as that workflow will be replaced

vd1992:
🤖 Resolves #6753

## 👋 Introduction

Enables revamp in both environment example files.

## 🕵️ Details

Test fixes were needed due to validation being enabled now, breaking existing patterns.
Updated to work off other commands to add in an update application step to meet validation using new Cypress commands.
A PHP unit test needed some tweaks to meet validation as well.

~Didn't see any sort of skip in the Cypress documentation so I commented out the one test~

Cypress doesn't seem to like chaining commands in some cases, linting plugin runs in those files now too. Just not the CI. 

## 🧪 Testing

1. Tests pass in CI
2. Tests pass in local environment


apps/e2e/cypress/e2e/talentsearch/submit-application-workflows.cy.js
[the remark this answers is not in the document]
vd1992: Works, could've sworn I tried it. Guess I must've used it wrong.

apps/e2e/cypress/support/userCommands.js
```diff
@@ -88,6 +89,16 @@ Cypress.Commands.add("getMe", () => {
   });
 });
 
+Cypress.Commands.add("getMeAllData", () => {
```
mnigh, line 92:
```suggestion
Cypress.Commands.add("getMe", () => {
```

vd1992: That was deliberate as there is already a "getMe" command present that gathers only a slice of data returned by query `me`.
So rather than change existing commands, I introduced a new one that gathers all data when required.

esizer: Would it be better to just update the `cy.getMe` to return all data instead of having two co-existing queries that do similar things with slightly different data? This smells of another `Pool` v `PoolAdvertisement` thing 😬

vd1992: `cy.getMe` runs query `me`

image

Which will only ever return the below

image

What I added, runs query `getMe`

image

Which returns all

image

vd1992: So really, they are two distinct queries that accomplish different things, if anything, `cy.getMe` should be `cy.me` to match the query to the command.

petertgiles: It's probably good to not overfetch a bunch of unneeded data.  Perhaps better naming of the queries would help clarify the difference?

petertgiles: Then again, for e2e tests, overfetching doesn't really matter.  I think I prefer Eric's recommendation of switching the single command to use the bigger query.

apps/e2e/cypress/e2e/admin/pool-candidate.cy.js
```diff
@@ -86,6 +87,19 @@ describe("Pool Candidates", () => {
       cy.getMe().then((testUser) => {
         cy.get("@publishedTestPool").then((pool) => {
           cy.createApplication(testUser.id, pool.id).then((poolCandidate) => {
+            cy.getMeAllData().then((me) => {
```
petertgiles, line 90: This block of code has been copy-pasted throughout the tests so it's probably a good candidate for a Cypress command.  Or maybe just roll it into `createApplication`.

petertgiles (changes requested):
I ran the suite and it works great!  I have a request and a suggestion below to check on.

apps/e2e/cypress/support/userCommands.js
vd1992: done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/6756/commits/72b329cc25ee5036c52e2d9c0e9729b77ca4b0c0

apps/e2e/cypress/e2e/admin/pool-candidate.cy.js
```diff
@@ -119,15 +131,13 @@ describe("Pool Candidates", () => {
     cy.wait("@gqlgetPoolCandidateSnapshotQuery");
     cy.wait("@gqlGetPoolCandidateStatusQuery");
 
-    cy.findByRole("combobox", { name: /candidate pool status/i })
-      .select("Screened In")
-      .within(() => {
-        cy.get("option:selected").should("have.text", "Screened In");
-      });
+    cy.findByRole("combobox", { name: /candidate pool status/i }).select(
```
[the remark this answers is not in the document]
vd1992: aye aye https://github.com/GCTC-NTGC/gc-digital-talent/pull/6756/commits/db7bd00f30e732ce79898de7f6176a5ef93ad340

apps/e2e/cypress/e2e/admin/pool-candidate.cy.js
vd1992: I was iffy on it as I didn't want to hide or abstract too much of what was going on
I put it into `createApplication` with some comments explaining things
https://github.com/GCTC-NTGC/gc-digital-talent/pull/6756/commits/cd25fec21d31a1b36d505041d6954b6e09fdda52

petertgiles: Yeah, nice. :+1:

apps/e2e/cypress/e2e/admin/pool-candidate.cy.js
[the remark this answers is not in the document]
petertgiles: Yeah, that works.  Doing the same selection twice isn't super-dry but I think it's fine.
I think the preferred solution is to use aliases like this:
image

petertgiles (approved):
:+1:

apps/e2e/cypress/e2e/admin/pool-candidate.cy.js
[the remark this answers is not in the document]
vd1992: Not a big difference for this test instance, but I will keep this in mind for cases where you'd act on a single entity many times. 
Thanks!

PR #6803, opened 2023-06-01
[Debt] Switch requests table to paginated

### LINKED ISSUE #6389: ♻️ Convert Requests table to a backend paginated table
## ♻️ Debt/Refactor

The Admin Requests table is getting big. We should convert the query and the table to use server-side pagination.

## 🕵️ Details

For now, do not worry about adding filters. However, ensure that the table can be ordered by any column. 

Remove the search bar, as it will no longer work.

Default ordering should be by creation date, most recent shown first.

## 🙋‍♀️ Proposed Solution

Use the `@paginate` directive, and follow the same pattern we've established with other server-paginated tables.

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] The Requests table uses a server-paginated query.
- [ ] By default, results are ordered by creation date, most recent shown first.
- [ ] Results can be ordered by any column available in the table.
- [ ] QUESTION: default count should be...? 10? 50?

vd1992:
🤖 Resolves #6389 

## 👋 Introduction

Switches the requests table to use the API managed tools. 

## 🕵️ Details

The query was added.
The table was created by copying `<PoolCandidates>` and carving it up to replace everything. Leveraging existing stuff was half the battle here. 

Additional small change to `fakeRequests()` to move to seeding `ApplicantFilter` objects. 

## ❗ Notes

Department sorting not operational, wasn't sure how to order them best given that departments are a connected table and the name itself comes from two different fields, the English and French. Not sure whether to split it off, use search bar for it or what.

Test files aren't exactly exhaustive. I wanted to get the ball rolling with things, so for the added jest file for example, as more features are added to the table hopefully people fill in the test file.

The query has the filter input set-up, I thought it would be easier to have it set and do nothing rather than need to update the query itself to add in a filter object at a later point. It should be thus easier to fill in `PoolCandidateSearchRequestInput` with follow-up work.

Search and filtering are examples of tasks down the line. And ideally soon because as is the table is lacking a bit in usefulness if you can't search through requests. 

## 🧪 Testing

1. Navigate to `/en/admin/talent-requests`
2. Observe 

## 📸 Screenshot

image


apps/web/src/components/SearchRequestTable/SearchRequestsTable.tsx
petertgiles, line 1: I noticed the email column is missing new in the updated table.  Intentional?
image

petertgiles, line 1: I noticed the left padding is missing and the columns are bolded in the updated table.  Unfortunately, the admin tables are not consistent but the previous style seems to be more common.

petertgiles, line 1: I believe the default sorting should be created_at DESC.
image

apps/web/src/components/SearchRequestTable/SearchRequestsTable.tsx
```diff
@@ -0,0 +1,320 @@
...
+      {
+        label: intl.formatMessage({
+          defaultMessage: "Job Title",
+          id: "8hee5d",
+          description:
+            "Title displayed on the search request table job title column.",
+        }),
+        id: "jobTitle",
+        sortColumnName: "job_title",
+        accessor: (d) => d.jobTitle,
```
petertgiles, line 213: I think this needs to be localized, too.
image

apps/web/src/components/SearchRequestTable/SearchRequestsTable.tsx
vd1992: Are you uh sure? The email is present in your screenshot there, screenshots below, and in the PR screenshot section.

vd1992: Um, not sure about that. The bolded is how we've been doing the API-managed tables, so pool candidates as well as users. And the left padding bug is something that has been present for a while and tricky to resolve. It arises for column headers that are not sortable. 
I've been mimicking and reusing the API-managed table stuff, so styling and quirks come downstream of it. 

I don't have on hand where I discussed the visual quirkness of non-sortable columns with someone else (Eric maybe) but you can see another example of it below as well as the formatting of pool candidates. 

image

vd1992: Ah, here we go
Think I found it

https://github.com/GCTC-NTGC/gc-digital-talent/pull/5631#discussion_r1098794112

apps/web/src/components/SearchRequestTable/SearchRequestsTable.tsx
vd1992: `jobTitle` is not an object with en and fr keys

image

apps/web/src/components/SearchRequestTable/SearchRequestsTable.tsx
vd1992: Oops, yep, fixed in https://github.com/GCTC-NTGC/gc-digital-talent/pull/6803/commits/8a982d19f13f1b6603d9029d8e1b79ee818af5af
Given all the seeded requests are created at the same time, I didn't even consider the ordering. Not sure if this is good practice, but I added this to get around that. Created now varies, and request completion is now no longer before creation in https://github.com/GCTC-NTGC/gc-digital-talent/pull/6803/commits/10ecb3181cabe113d0e1b98588eb263818eeccf8

petertgiles: Sorry, in my screenshot the top is now, the bottom is before.  Yeah, if you check Chromatic for the removed story you can see there's no email.  
image

petertgiles: Oh, so it's a client vs server table thing.  OK, I'm fine if this is out-of-scope for now.

apps/web/src/components/SearchRequestTable/SearchRequestsTable.tsx
petertgiles: Oh, you're right - this is a non-localized free-form field.

apps/web/src/components/SearchRequestTable/SearchRequestTableClient.tsx
```diff
@@ -167,18 +167,18 @@ export const SearchRequestTable = ({
...
+      <SearchRequestTableClient
         poolCandidateSearchRequests={data?.poolCandidateSearchRequests ?? []}
         title={title}
       />
     </Pending>
   );
 };
 
-export default SearchRequestTableApi;
+export default SearchRequestTableWrapper;
```
vd1992, line 184: Calling the wrapper around a component named `_____` the `_____Api` component is a pattern we use, but confusing in this case. Given this file will be deleted before long, it should be palatable. 😛

apps/web/src/components/SearchRequestTable/SearchRequestsTable.tsx
vd1992: Email hidden by default in https://github.com/GCTC-NTGC/gc-digital-talent/pull/6803/commits/c0e883435e3d6178f524e6625d9db563f87f4300

vd1992:
Okay, Chromatic re-run

https://www.chromatic.com/review?appId=61e099a1bb1465003a73d3ce&number=6803&type=linked

petertgiles:
Could you leave some instructions here to swap the api-driven component back in?  It could be helpful for testing and for the next PR.

vd1992:
> Could you leave some instructions here to swap the api-driven component back in? It could be helpful for testing and for the next PR.

Sure, added to the page in question https://github.com/GCTC-NTGC/gc-digital-talent/pull/6803/commits/27f56f5f1ca8b20c5b5a80eb7a76b6ef45afb9c1

petertgiles (approved):

PR #6812, opened 2023-06-02
[Accessibility] Remove aria-selected attribute from ToggleSelection

### LINKED ISSUE #6764: ⌨ `ToggleSection` has incorrect `aria-*` attribute
## ⌨ Accessibility Issue

We curently use the `aria-selected` attribute which is only allowed on elements with specific roles.

### WCAG 2.1 Criteria

[Elements must only use allowed ARIA attributes](https://dequeuniversity.com/rules/axe/4.7/aria-allowed-attr?application=axeAPI)

## 🦋 Expected Behaviour

I do not believe this should have any `aria-*` attribute at all but am a little uncertain 🤔 

## 🙋‍♀️ Proposed Solution

Remove the `aria-selected` attribute or, justify a role/aria attribute modification.

## ✅ Acceptance Criteria

- [ ] ARIA error no longer appears

And, one of the following:

- [ ] `aria-selected` removed
- [ ] New `role` added the content divs
- [ ] Different `aria-*` attribute selected

yonikid15:
🤖 Resolves #6764 

## 👋 Introduction

- Removes `aria-selected` attribute from div elements in `<ToggleSelection />, which is only allowed on elements with specific roles (eg. tab, row, etc.)

## 🧪 Testing

1. Use Storybook or go to `http://localhost:8000/en/applications/{applicationId}/profile` (You will need to turn on APPLICATION_REVAMP feature flag)
2. Ensure aria errors do not appear.

## 📸 Screenshot

image

### DIFF
--- a/packages/ui/src/components/ToggleSection/ToggleSection.tsx
+++ b/packages/ui/src/components/ToggleSection/ToggleSection.tsx
@@ -140,12 +140,7 @@ const InitialContent = React.forwardRef<
   const id = composeId(NAME.INITIAL_CONTENT, context?.contentId);
 
   return (
-    <div
-      id={id}
-      ref={forwardedRef}
-      aria-selected={!!context?.open || false}
-      {...props}
-    >
+    <div id={id} ref={forwardedRef} {...props}>
       {context?.open ? null : children}
     </div>
   );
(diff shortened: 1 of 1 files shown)


esizer (approved):
LGTM! 👍

PR #6838, opened 2023-06-08
[Feature] Add optional skills to snapshot view

### LINKED ISSUE #6769: ♻️ Make optional skills justifications visible on Admin Snapshot view
## ♻️ Debt/Refactor

Add optional skills justifications, if filled out, to snapshot view. 

## 🕵️ Details

Optional skills are listed in some posters and answers should be displayed somewhere if filled out by an applicant. 

(Jerbo here) I added a couple more small fixes to the snapshot in here

1 - Rename the section to simply "Application"
2 - Rename the skills sections on admin side to "HR" language and call them "Essential skills" and "Asset skills" respectively
3 - Add "Asset skills" section

[Link to figma](https://www.figma.com/file/BXYVsu1fo9LS2IySJDUINY/Applicants---Job-application?type=design&node-id=636%3A54957&t=eN9WQXne2pTQYuY1-1)

<details><summary>Screenshot</summary>
<p>

Frame 40

</p>
</details> 

## ✅ Acceptance Criteria

- [ ] Optional skills section added to where requested

JamesHuf:
🤖 Resolves #6769.

## 👋 Introduction

Adds optional skills to the admin snapshot view. Also includes a couple minor copy changes specified in the issue.

## 🧪 Testing

1. Rebuild the site.
2. Submit an application to a pool with some optional skills having experiences.
3. Navigate to admin and view the applications snapshot.
4. Confirm the snapshot displays both Essential skills and Asset skills sections filled out as expected.


esizer (commented):
Looks great! I have two questions, both out of scope for this issue but one may be easy enough to roll in if we want 🤷‍♀️ 

1. I think we should have a null state, even something simple like a `Well` with "no skills" message 🤔 
2. Should we just go ahead and show all skills? Right now we just show technical even though we also know what behavioral skills exist in the snapshot. Should be as easy as dropping the `categorizeSkill` and using the skills directly 

@Jerryescandon What do you think?

Jerryescandon:
So there's a couple behavioural things that I would like to avoid. If we show to the applicant a massage like "no skills" it tends to create anxiety since they start believing their application is not as strong as it could be. So they tend to lie and add skills they dont really have to their profiles.

In the cases where we didn't ask for any asset skills, then its just gonna confuse users that we have a section that has no skills in it, they might think they SAHOULD have added some somewhere.

HOWEVER. This is not an issue for the pool operators. In fact is better for pool operators to quickly identify what they can or cant assess. If we can add this ONLY to the Application snapshot on admin side, but not to the review page on the application side, it would be a nice addition for pool operators (show null asset skills list)

For behavioural skills is a similar thing, since we didn't really allowed applicants to add any info I think we should avoid telling the users that they have nothing there. Again, it is not a big of an issue on admin side. Pool operators might benefit from this.

TL:DR
If we can ONLY do it for Admin, and not for the applicant page review it might be worth it.

esizer (approved):
Looks good! Questions answered during standup 👍

PR #6896, opened 2023-06-08
[Fix] Don't apply more than once

### LINKED ISSUE #6751: 🐛 Error alert appears "You have already applied to pool" when the user has not applied yet
## 🐛 Bug

When the user tries to apply the position he already created draft an error alert appears. 
Its not preventing him to continue the draft application. 
However that error popup looks buggy 
(Another annoying side effect , It also fills up Application Insight error log ) 

## 🦋 Expected Behaviour

User should be able to continue the draft application without any error 

## 🕵️ Details

## 📋 Steps to Reproduce

1. Go to https://talent.canada.ca/en/browse/pools
2. Click on any job and create draft application
3. Log out 
4. Now repeat the step 1
5. Click Apply button for the same job you created draft 
6. You will be directed to login page .. enter your credentials
7. You will be redirected back to the job page 
8. Observe the error alert appears 

## 📸 Video

https://github.com/GCTC-NTGC/gc-digital-talent/assets/89482910/5b66f20d-9baf-429d-b5f5-40d63e6586a4

## 🙋‍♀️ Proposed Solution

Two options:
1. In CreateApplicationPage, check if the user already has an application _before_ running the createApplication mutation. Navigate to the existing mutation if it exists.
2. In CreateApplicationPage, ensure the APPLICATION_EXISTS error is ignored if that is the only error returned. (The CreateApplication mutation on the backend is set up to return the error, but also return the existing application id, so navigating to the application already works.) This has the advantage of only ever requiring a single api call, but you'd have to figure out how to override the core error handler in ClientProvider for this one case, and I don't know if that's possible.

## ✅ Acceptance Criteria

- [ ] There shouldn't be any error when clicking the job that the user has not applied yet.

petertgiles:
🤖 Resolves #6751 

## 👋 Introduction

This branch fixes the issues of the user accidentally applying more than once because of:
- Being logged out while already having an application
- useEffect running more than once on the page

## 🕵️ Details

For the first issue, I just added a query for existing applications and didn't try applying until we have confirmed that an application doesn't already exist.

For the second issue, I added a ref to count how many times we had applied and just didn't fire the mutation if it was greater than zero.  This worked well and it occurred to me we didn't need the useEffect and useCallback functions at all if we relied on that.

## 🧪 Testing

1. Rebuild the app

- [ ] When logged out, try starting an application to a pool with an already existing application.  Confirm that you are sent to the existing application after log in.  Confirm no errors.
- [ ] When logged in, try starting a new application.  Confirm no errors.
- [ ] When logged in, try manually navigating to the `create-application` URL of a pool with an already existing application. Confirm that you are sent to the existing application.  Confirm no errors

### DIFF
--- a/apps/web/src/pages/CreateApplicationPage/CreateApplicationPage.tsx
+++ b/apps/web/src/pages/CreateApplicationPage/CreateApplicationPage.tsx
@@ -37,14 +37,8 @@ const CreateApplication = () => {
   const navigate = useNavigate();
   const { applicationRevamp } = useFeatureFlags();
   const auth = useAuthorization();
-  const [
-    {
-      fetching: creatingNewApplication,
-      data: newApplicationData,
-      operation: newApplicationOperation,
-    },
-    executeMutation,
-  ] = useCreateApplicationMutation();
+  const [{ data: newApplicationData }, executeMutation] =
+    useCreateApplicationMutation();
   const [{ data: existingApplicationsData }] = useMyApplicationsQuery();
 
   // Path to display application (new or existing).
@@ -58,6 +52,9 @@ const CreateApplication = () => {
 
   // Store path to redirect to later on
   let redirectPath = paths.pool(poolId || "");
+
+  // We hae the mutation in a useCallback to run once only but it fires twice in strict mode.
+  const mutationCounter = React.useRef<number>(0);
 
   /**
    * Handle any errors that occur during mutation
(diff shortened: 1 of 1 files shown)


esizer (approved):
Nice! Works well and squashes out two annoying 🐛 's!

Though, I should note; it feels kind of disruptive to click "Apply..." then be shot right into the middle of an application in process with no feedback on why 🤔 I'd like if we could do an informational toast letting users know what is about to happen but I'm not sure how others feel 🤷‍♀️

petertgiles:
When I added the new toast I could see it was popping twice.  I added a ref to make sure it only ran once.

esizer (approved):
Sweet! 🍭 Works great and super clean solution 😄

PR #6908, opened 2023-06-09
[Bug] Fix text link font size

### LINKED ISSUE #6907: 🐛 Text should not set font size
## 🐛 Bug

With the updates to links, we set the the font size to the same size as copy. This is great when we use with next to copy but looks weird when links occur in headings.

## 🦋 Expected Behaviour

Links with `mode="text"` should render the same font size as its siblings.

## 📸 Screenshot

Screenshot 2023-06-09 123932

## 🙋‍♀️ Proposed Solution

Omit font size from link styles when mode is set to text.

## ✅ Acceptance Criteria

- [ ] Text links render the same size as adjacent siblings
- [ ] Other links are un changed

esizer:
🤖 Resolves #6907 

## 👋 Introduction

Prevents text links from override font size.

## 🧪 Testing

2. Confirm links that are used within text match their siblings
3. Confirm other links are unaffected

## 📸 Screenshot

Screenshot 2023-06-09 125020

### DIFF
--- a/packages/ui/src/hooks/useCommonButtonLinkStyles.ts
+++ b/packages/ui/src/hooks/useCommonButtonLinkStyles.ts
@@ -11,6 +11,7 @@ type StyleRecord = Record<string, string>;
 const styleExclusions: Record<string, Array<ButtonLinkMode>> = {
   background: ["inline", "text"],
   border: ["inline", "text"],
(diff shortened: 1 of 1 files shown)


vd1992 (approved):
Diff looks good
Built and it also looks more fitting now too

👍

PR #6940, opened 2023-06-13
[Feature] Qualified recruitment card

### LINKED ISSUE #6737: ⚙️ Create a QualifiedRecruitmentCard for the Resume page
image

## 🌟 Details ([Figma file](https://www.figma.com/file/2VF8cE4cctTBQUkPBcVxUG/3.-Patterns?type=design&node-id=703-60935))

This card is for representing a pool a user has qualified in, ie a "Qualified Recruitment", and is primarily intended for the Qualified Recruitments section of the the Resume page (#6773)

- Data types that have their own card on profiles are being given a theme color, in this case, recruitments are using `secondary`
- The card has been contains a simple accordion which lists the required skills of the pool
- The layout and language around card sub-actions has been updated, including the addition of `aria-label` text

### Accessibility

- Because experience cards are repeated in a list, it's important that interactive elements inside the card contain descriptive `aria-label`  values to ensure that link-navigation via AT reads things out properly
  - "Show this recruitment's skill assessments" >> "Show assessed skills for the IT-02 Web technician (Java) recruitment process"
  - "Hide this recruitment's skill assessments" >> "Hide assessed skills for the IT-02 Web technician (Java) recruitment process"
  - "Change your availability" >> "Change your availability for the IT-02 Web technician (Java) recruitment process"
  - "Copy recruitment ID" >> "Copy the IT-02 Web technician (Java) recruitment process' ID"

### Typography

- The card title should use the `h6` font-size while being whatever HTML heading is relevant for the page context
- The recruitment type should use the `secondary.darker` color
- The separator bullet should use `black`
- The "Awarded on" label and date should use `black.light`
- The details expansion should use the `copy` size variant`
- Label and value pairs inside the expansion all use the `copy` font size, with labels bolded 
  - The "Behavioural skills" section, if applicable, uses the `globe-americas` Heroicon from the `mini` set with the `secondary` color
  - The "Technical skills" section, if applicable, uses the `bolt` Heroicon from the `mini` set with the `secondary` color
- The availability message and copy ID link use the `caption` font-size; links are bolded 

### Size

Note that the top and bottom right corners use the `rounded` border-radius value and the card itself uses the `larger` box-shadow value.

image

### Card groups

image

### Targeted recruitment vs. Ongoing recruitment

The only distinction between the two is that their type is indicated on the card below the title. For both, the expansion contains the assessed skills that were linked to the job advertisement.

image

### Qualified states

- All qualified states use the `success` variant of the chip component and contain the `shield-check` Heroicon from the `mini` set

####  Open availability

- The "open availability" message uses the `check-circle` Heroicon from the `mini` set with the `success` color

image

##### Change availability dialog

image

#### Closed availability

- The "unavailable" message uses the `no-symbol` Heroicon from the `mini` set with the `error` color

image

#### Hired

- The "you've been hired" message uses the `star` Heroicon from the `mini` set with the `quaternary.dark` color

image

### The expired state

- Uses the `error` variant of the chip component
- The "expired recruitment" message that replaces the availability control uses the `lock-closed` Heroicon from the `mini` set with the `gray.darker` color

image

## Acceptance Criteria
These criteria apply to the card when isApplicantPlaced is true.
- [ ] The card has the described layout, colors, and typography
- [ ] The card displays whether the pool was a targeted or ongoing recruitment
- [ ] A simple accordion has been added to the card, which shows which required skills were part of the Pool
- [ ] A user can suspend their application, as long as the pool hasn't expired or they haven't been placed
- [ ] `Copy recruitment id` is a button which copies pool id to clipboard
- [ ] All interactive elements have aria-labels
- [ ] Hide the date for now - we do not currently record the "accepted-at" date
- [ ] For now, Accordion trigger text does not need to change based on open/close state, just use the "Show" version.
- [ ] This component is used on the Resume page (soft mutual dependency with #6773)

esizer:
🤖 Resolves #6737 

## 👋 Introduction

Adds a new component `QualifiedRecruitmentCard`.

## 🕵️ Details

Some minor adjustments have been made to the dialog (to be submitted to designers for review) to attempt to improve accessibility. Some changes that were made:

- Updated the title to include the pool title since it is a label for the dialog and should be unique
- Reduced the amount of bold text to meet[ the WebAIM readability suggestions for cognitive accessibility](https://webaim.org/articles/evaluatingcognitive/#readability:~:text=Overuse%20can%20result%20in%20the%20loss%20of%20differentiation.%20Do%20not%20use%20italics%20or%20bold%20on%20long%20sections%20of%20text.%20Be%20careful%20with%20ALL%20CAPS.)
- Similar, the word "NOT" was all uppercase which can be read incorrectly by certain ATs and makes it more difficult for users with disability such as dyslexia to read
- Button order has been swapped, this one is far more subjective but it seems (at least to me) as though "Cancel" is the least "destructive" action so it should lead. 

## 🧪 Testing

2. Navigate to the Qualified Recruitment Card story
3. Confirm the states match [the expected states in the mockup](https://www.figma.com/file/2VF8cE4cctTBQUkPBcVxUG/3.-Patterns?type=design&node-id=703-60935&t=urWlzAtSRrR080MT-0)

### Qualified - Available
Screenshot 2023-06-13 095834

### Qualified - Not Available
Screenshot 2023-06-13 100109

### Expired
Screenshot 2023-06-13 100115

### Placed
Screenshot 2023-06-13 100121

### Open
Screenshot 2023-06-13 100842

### Dialog 
Screenshot 2023-06-13 100101


substrae:
@esizer 

- Not sure I agree with the title change; this is now super difficult to read from a sighted perspective, and is immediately redundant given the next line of text
- I've updated the designs with the reduced bolding
- I've updated the designs with the lowercased "not"
- We've taken an approach with buttons recently based on research @Jerryescandon had done around both left-aligning actions and promoting the primary action to always be first, with secondary actions after

esizer:
> * Not sure I agree with the title change; this is now super difficult to read from a sighted perspective, and is immediately redundant given the next line of text
> * I've updated the designs with the reduced bolding
> * I've updated the designs with the lowercased "not"
> * We've taken an approach with buttons recently based on research @Jerryescandon had done around both left-aligning actions and promoting the primary action to always be first, with secondary actions after

Thanks, I have updated the title and button order in [update title, button order](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6940/commits/cf8b0ca590d60f9f9073a4b03ad4f15bc2ebe743).

Related to the button order, we should do that sooner rather than later. I noticed we have introduced some inconsistency in the order which could cause some confusion. I'll see if we can get that planned as part of the next sprint 👍

substrae:
Oh one small thing I noticed, I think the title font-size on the cards is off by one level. Thanks!

esizer:
> Oh one small thing I noticed, I think the title font-size on the cards is off by one level. Thanks!

Hmm, if that is correct then we may have a larger issue with font size 😅 Right now it appears at `h6` (`18.4px`) since that was [how the issue was defined](https://github.com/GCTC-NTGC/gc-digital-talent/issues/6737#:~:text=The%20card%20title%20should%20use%20the%20h6%20font%2Dsize%20while%20being%20whatever%20HTML%20heading%20is%20relevant%20for%20the%20page%20context). If it is not correct then the hydrogen config may be out of sync 😬 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/91711484d7e4e5374c44d8167bab670493973d2b/apps/web/src/components/QualifiedRecruitmentCard/QualifiedRecruitmentCard.tsx#L81-L83

Screenshot 2023-06-13 132813

apps/web/src/components/RecruitmentAvailabilityDialog/RecruitmentAvailabilityDialog.tsx
```diff
@@ -0,0 +1,192 @@
...
+                items={[
+                  {
+                    label: intl.formatMessage({
+                      defaultMessage:
+                        "I am <strong>available</strong> for hire and want to be contacted about opportunities from this recruitment process.",
+                      id: "cAOf3a",
+                      description:
+                        "Radio button label for available for opportunities option",
+                    }),
+                    value: "true",
```
vd1992, line 151: This works, but it is a little confusing to follow as `isSuspended` is essentially flipped for the form where `true` maps to not suspended.
Connects to this bit below where `isSuspended = true` sets `isSuspended = false`
Not a big deal as it appears to work fine, maybe just a comment or renaming the second variable/form to `isActive` if you want. 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/91711484d7e4e5374c44d8167bab670493973d2b/apps/web/src/components/RecruitmentAvailabilityDialog/RecruitmentAvailabilityDialog.tsx#L32-L33

apps/web/src/components/RecruitmentAvailabilityDialog/RecruitmentAvailabilityDialog.tsx
```diff
@@ -0,0 +1,192 @@
...
+  });
+  const {
+    handleSubmit,
+    formState: { isSubmitting },
+  } = methods;
+
+  const updateSuspendedAtStatus = async (values: FormValues) => {
+    await executeMutation({
+      id: candidate.id,
+      isSuspended: values.isSuspended === "true",
```
vd1992, line 43: Hang on, maybe the above will pose a problem after all. 
As "I am available" maps to "true", this passes in `isSuspended = true`, so the opposite of how it should play out.
Selecting "I am available" will suspend your pool candidate.

vd1992: Yep, imported the component in to try it out

https://recordit.co/EJ1GVuk9IK

apps/web/src/components/QualifiedRecruitmentCard/QualifiedRecruitmentCard.tsx
vd1992, line 1: Might be worth it to stick margins around it so a list of them don't need a containing div to space them out?
Not too sure about this comment, so disregard if you want 👍 

image

apps/web/src/messages/poolCandidateMessages.ts
vd1992, line 1: I think we are supposed to use "recruitment process" for live site copy rather than "recruitment", might need to double check at stand-up

vd1992 (changes requested):
Checked around some places, not exhaustive as I am poor at reviewing things with lots of moving parts. 
Found some stuff, take a look. Otherwise, looks awesome

apps/web/src/components/QualifiedRecruitmentCard/QualifiedRecruitmentCard.tsx
esizer: Spacing can be determined by the layout of a page and can change depending on where a specific component is being displayed. Baking in a basic margin complicates this so it is avoided in favour of flex gap. I'm willing to listen to other opinions though if others feel we should have spacing baked into the components.

apps/web/src/components/RecruitmentAvailabilityDialog/RecruitmentAvailabilityDialog.tsx
esizer: Oof, yeah. That is wrong 😓 I got confused because the form is a positive statement "I am available" so I was thinking positive statement === true 😜 

Fixed in [invert isSuspended logic for availability dialog](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6940/commits/f891ecb079c5be9bab6b770e56d911ff9b22f00e)

apps/web/src/components/QualifiedRecruitmentCard/QualifiedRecruitmentCard.tsx
```diff
@@ -0,0 +1,343 @@
...
+          ) : (
+            statusPill.text
+          )}
+        </Pill>
+      </div>
+      <p
+        data-h2-color="base(secondary.darker)"
+        data-h2-margin="base(x.5 0 x1 0)"
+      >
+        {isOngoing
```
brindasasi, line 106: oh am doing the samething .. I think we are overlapping ..will base mine of yours then.
Do you mind taking this out into a utility function something like getRecruitmentType?

esizer: Yeah, of course! I can do that and let you know when it is completed 👍

brindasasi:
Looks like we are missing **Awarded on** 
Screenshot 2023-06-14 at 10 27 34 AM

apps/web/src/components/QualifiedRecruitmentCard/QualifiedRecruitmentCard.tsx
vd1992: 👍

apps/web/src/messages/poolCandidateMessages.ts
vd1992: Non issue

brindasasi:
> > Looks like we are missing **Awarded on**
> 
> The acceptance criteria states to [not display awarded on](https://github.com/GCTC-NTGC/gc-digital-talent/issues/6737#:~:text=Hide%20the%20date%20for%20now%20%2D%20we%20do%20not%20currently%20record%20the%20%22accepted%2Dat%22%20date).
oh yeah .. forgot about that. Nvm! :)

apps/web/src/components/RecruitmentAvailabilityDialog/RecruitmentAvailabilityDialog.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+  return (
+    <Dialog.Root open={isOpen} onOpenChange={setIsOpen}>
+      <Dialog.Trigger>
+        <Button
+          mode="inline"
+          color="black"
+          aria-label={intl.formatMessage(
+            {
+              defaultMessage: "Change your availability for {title}",
```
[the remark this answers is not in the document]
esizer: Fixed in [Add period to aria-label](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6940/commits/db59c1f102f19cede6f26f14d355799c068cbd6b) so the ID would also update to match.

esizer:
> This is bold in figma

Good catch! 👍 Fixed in [make status pill bold](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6940/commits/e66adec870c6396ab585f8e3d354d1382eb3f730)

brindasasi (commented):
Are you doing this ? https://github.com/GCTC-NTGC/gc-digital-talent/pull/6940/files/f891ecb079c5be9bab6b770e56d911ff9b22f00e#r1229684097
Everything else looks good to me! 👍🥳

brindasasi:
Looks 

> Are you doing this ? https://github.com/GCTC-NTGC/gc-digital-talent/pull/6940/files/f891ecb079c5be9bab6b770e56d911ff9b22f00e#r1229684097 Everything else looks good to me! 👍🥳

oops .. just noticing ..thanks! 😅

brindasasi (commented):
minor thing: **getTargetedRecruitment**  name doesn't suit well as TargetedRecruitment is a type of recruitment and this function will also be used for ongoing recruitment as well. 

**getRecruitmentType** or something like that would be suitable

brindasasi (approved):
I think you are mind reading me 😆

Thank you and this is good to go with translations👏👏👏

PR #6959, opened 2023-06-14
[Fix] Disable transition for toast story

### LINKED ISSUE #6858: 🐛 Toast story goes in and out of Chromatic
## 🐛 Bug

I've observed the toast story going in and out of Chromatic

## 🦋 Expected Behaviour

The story should be stable so as to not trigger Chromatic diffs.  It's an animated story so maybe it's a timing thing.

## ✅ Acceptance Criteria

- [ ] Toast story is stable and does not trigger diffs

petertgiles:
🤖 Resolves #6858 

## 👋 Introduction

This branch disables the transitions for the toast story.  This hopefully will avoid empty snapshots being taken of it.

## 🧪 Testing

- [ ] In Storybook, the BasicToast story has no transition
- [ ] In the app, the Toasts still use the Slide transition
- [ ] In Chromatic, the snapshot is never blank


packages/toast/src/components/Toast/Toast.stories.tsx
```diff
@@ -31,7 +31,8 @@ const Template: Story<StoryArgs> = (args) => {
   toast.warning(text, { autoClose: false });
   toast.error(text, { autoClose: false });
 
-  return <Toast />;
+  // no transition to avoid issues with Chromatic snapshots
+  return <Toast disableTransition />;
```
esizer, line 35: May be worth [disabling auto close](https://fkhadra.github.io/react-toastify/autoClose#prevent-the-toast-from-closing-by-default) to so they don't disappear while the snapshot is taken 🤔

esizer (changes requested):
Not sure if this will improve it but I also don't think it will make it worse either 🤷‍♀️ 😆

packages/toast/src/components/Toast/Toast.stories.tsx
petertgiles: Good idea!
Ref: [disable autoclose for story](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6959/commits/c4c625ea3451aef73a57146695f9ab3abf12625f)

esizer (changes requested):
I have no idea what is going on here but now they disappear when you leave the story and you cant get them back unless you refresh the page 😵‍💫

petertgiles:
> I have no idea what is going on here but now they disappear when you leave the story and you cant get them back unless you refresh the page face_with_spiral_eyes

Yeah, I was seeing very weird behaviour before the change, too.  I figured it was just Storybook 7 being new and glitchy.  I'll take a closer look tomorrow.

petertgiles:
> I have no idea what is going on here but now they disappear when you leave the story and you cant get them back unless you refresh the page face_with_spiral_eyes

As far as I can tell, it was the preview on the autogenerated docs interfering.  It was happening on main, too.  I tried creating a placeholder docs page without the preview and it seems to be helping.

esizer (approved):
Awesome! Looks to be working now 👍 

Now we just need to go back and write the docs page 😆

PR #6994, opened 2023-06-15
[Feature] Searching on the requests table

### LINKED ISSUE #6465: ✨ Re-add search box to request table
## ✨ Feature

After the query is set to be backend-driven the search box will need to be added back in.

## ✅ Acceptance Criteria

- [ ] Search box re-added
- [ ] text search should filter out rows which don't have matching text in any of the following columns:
  - fullName
  - email
  - jobTitle
  - additionalComments
  - adminNotes
- [ ] Works in a similar way to other backend-driven tables such as _User_
- [ ] Changes to API are tested in phpunit

vd1992:
🤖 Resolves #6465

## 👋 Introduction

Enables using the search for the new requests table. 

## 🕵️ Details

Nothing too fancy going on here I think, mostly straightforward. Just enabling the search ala other paginated tables. 
I am treating `adminMessages.ts` as a sort of constants file now for anything likely to be re-used across the admin side. 

## 🧪 Testing

1. Swap in the new request table in `IndexSearchRequestPage.tsx`
2. Navigate to `/en/admin/talent-requests`
3. Play around with searching

## 📸 Screenshot

image

image

image


apps/web/src/messages/adminMessages.ts
vd1992, line 1: As mentioned in the PR itself. 
I think this is important to avoid many repetitions of the same translations. Looking at "Notes" for example

image

vd1992: Definitely probably ought to taken care of sometime. 
And here is the translation for just "email"

image

petertgiles (approved):
Bravo!  :clap:

Issue #6887, opened 2023-06-08
✨ IAP-specific Login and Registration pages


tristan-orourke:
## ✨ Feature

If you need to log in or register from the IAP, either because you clicked the login or register links, or because you tried to apply, we want to show an IAP-specific info page.

## 🕵️ Details

This info page should use the IAP nav bar, the IAP theme, and have slightly different copy.

## 🎨 Design File

[Figma](https://www.figma.com/file/8piNhFaBjcxssBB8LAGxEY/IAP-User-flows?type=design&node-id=269-79798&t=GzGkJJ3XJ9X2vY9l-0)

## 📸 Screenshot

Add a screenshot of the new feature (if possible).

## 🙋‍♀️ Proposed Implementation

(optional) Do you have a proposed implementation?

## 🌎 Localization

(optional) Provide any new copy along with translations available.

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the feature was properly implemented.

- [ ] If user clicks on Login or Register from IAP home page, they land on a page with the IAP Nav bar, the IAP theme, and IAP-specific copy. After logging in, user is sent back to the IAP home page.
- [ ] If user tries to apply without being logged in, they are sent to the same IAP-specific login-info page. After coming back from SiC, they end up directly in the application.

BrookeCameron:
Copy has been updated to include an email link to the IAP team. [updated figma link](https://www.figma.com/file/8piNhFaBjcxssBB8LAGxEY/IAP-User-flows?type=design&node-id=269-79798&t=vAD1HUYHCC9cpFL9-4)

PR #6999, opened 2023-06-16
[Feature] IAP login

petertgiles:
🤖 Progresses (but does not close) #6887 

> __Note__
I'll put a second PR through with the copy changes later.

## 👋 Introduction

This branch customizes the login experience for the IAP.

## 🕵️ Details

I added the "personality" search param that can activate IAP specific theme, copy, and behaviour.  This is used to change the login and nav experiences.  We may want to eventually consolidate the layout and nav components between the main site and IAP now that there's leakage between them.

## 🧪 Testing

Build the site then test that the AC (other than copy changes) are met.  Note that we decided to reuse the same pages and URLs and not create new pages.

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/components/Layout/LogoutButton.tsx
+++ b/apps/web/src/components/Layout/LogoutButton.tsx
@@ -1,0 +1,24 @@
+import React from "react";
+
+interface LogoutButtonProps extends React.HTMLProps<HTMLButtonElement> {
+  children: React.ReactNode;
+}
+const LogoutButton = React.forwardRef<HTMLButtonElement, LogoutButtonProps>(
+  ({ children, ...rest }, forwardedRef) => (
+    <button
+      data-h2-color="base(black) base:hover(primary) base:iap(primary) base:iap:hover(primary.darker)"
+      data-h2-font-size="base(normal)"
+      data-h2-text-decoration="base(underline)"
+      style={{
+        background: "none",
+      }}
+      ref={forwardedRef}
+      {...rest}
+      type="button"
+    >
+      {children}
+    </button>
+  ),
+);
+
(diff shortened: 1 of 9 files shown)


apps/web/src/components/Layout/LogoutButton.tsx
petertgiles, line 1: Moving this code to its own file was to prevent a dependency loop between IAPNavMenu.tsx and Layout.tsx.

apps/web/src/hooks/useBreadcrumbs.ts
vd1992, line 1: Should the second breadcrumbs link also be updated to acknowledge the state of the page?

image

petertgiles: Hmm, good question - I'm actually not sure.   We don't have breadcrumbs preserve search params anywhere else in the app.  For example, outside of the IAP context, let's say you wanted to go to your profile and were automatically redirected here by the _RequireAuth_ component.  If you click that link you'll lose your _from_ param and won't go to the page you wanted after login.  A good argument for not having breadcrumbs on this page or having the last breadcrumb not be a link.

vd1992:
Appears to work for me.
Logging into the other stuff and the login with apply application, as well as for the IAP stuff locally.
And then repeated with SiC enabled.

I don't 100% follow what's going on with the login handling process, so I want someone else to also take a look to approve

tristan-orourke (approved):
The code looks great! I like the use of `personality`. Feels like something we might want to expand on and put into a context.

petertgiles:
Added
Add personality to context #7002
Consolidate IAP Layout and Nav #7003

PR #7010, opened 2023-06-19
[Fix] Prevent candidate count interrupting title announcement

### LINKED ISSUE #6827: Focus for search page not on page title
Current behaviour: When navigating to the Search page, the focus is set to the estimated candidate region.
Expected behaviour:  When navigating to the Search page, focus is set to the page title.

>  After activating the search link, the platform did not indicate that the search page had loaded. I heard information about the number of candidates that match my profile.

Screen reader users do not understand why they are told an estimated number of candidates meet their criteria upon page opening.

## Proposed implementation

Add the role alert after the first search completes?

## Acceptance Criteria
- [ ] Candidate count should not be announced until after the first search completes. We only want to announce when it changes, not it's initial value.

esizer:
🤖 Resolves #6827 

## 👋 Introduction

This fixes an issue that caused the number of estimated candidates to interrupt the announcement of the current page title.

## 🕵️ Details

This was caused by two separate issues working together.

1. We removed the `tabIndex` attribute from the `h1` which meant it could not be programatically focused to be announced
2. The candidate count entered after the page title so it was interrupting the announcement

This re-adds the `tabIndex={-1}` so we can once again properly focus the page title to announce context changes. Also, it adds a `useAriaLive` hook that returns `"off"` until the first render so we don't accidentally overload the announce queue, announcing too much on initial load.

## 🧪 Testing

1. Build `npm run dev`
4. Turn on a screen reader
5. Navigate to the homepage `/`
6. Tab to the "Find talent" link
7. Click it and confirm the page title is announced in full to the user


esizer:
> I think this is working as excepted. Before I approve, can you confirm that the estimate candidates sentence is meant to be read out right afterwards (see screen recording)?

Hmm, that is a good question 🤔 The initial issue was related to confusion because it was the first thing announced, I'm not sure if that extends to it being read out at all. I'm not entirely sure how to avoid that entirely unless we use toasts to announce the new candidate account instead of making estimated candidates a live region.

@RM-1978 Are you able to help us out with this?

apps/web/src/pages/SearchRequests/SearchPage/components/SearchContainer.tsx
```diff
@@ -447,6 +455,39 @@ const SearchContainerApi = () => {
     ? candidatesData?.countPoolCandidatesByPool
     : [];
 
+  /**
+   * Announce the candidate count to users in a less verbose way
+   *
+   * Note: 3 is a magic number, our current candidate count is causing
```
mnigh, line 461: Not sure I understand what this comment means: what is 3 a magic number for?

mnigh (commented):
Sort of better than before, but now (sometimes) it announces the number of skills immediately after the heading.

## Screen recordings

https://github.com/GCTC-NTGC/gc-digital-talent/assets/3046459/3cdfd7d9-fb9a-4919-b5c2-3208d82e938e

https://github.com/GCTC-NTGC/gc-digital-talent/assets/3046459/7d8a304d-73c7-436f-ad94-4c76a38be2da

apps/web/src/pages/SearchRequests/SearchPage/components/SearchContainer.tsx
esizer: Okay, skills should now be announced only after searching and I attempted to make the comment clearer.

[wait on skills dirty before announcing, improve comments](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7010/commits/15ddda8ecce523a552917b63d2916c1008f48f44)

mnigh (approved):
Meets acceptance criteria 😄

PR #7015, opened 2023-06-20
[Feature] Pool candidate table skill match dialog

### LINKED ISSUE #6658: ✨ Add modal to request page "skill match" column
## ✨ Feature

To improve the speed at which DCM validates the experience of requested candidates against the skills in a request, we want to include a modal that opens up from the skill match column that includes the experiences tied to the skills requested.

This would make it easier to find the right candidate by allowing DCM to view the relevant skill information without leaving the request page, and limiting the experience that shows up to only the ones relevant to a specific request. (As opposed to experiences on profile which show all experiences).

## 🕵️ Details

On the request page, on the results table, make the "Skill matches" column a clickable column

Clicking on the skill match number opens up a dialog that showcases ONLY the relevant experiences to the skills selected for that individual candidate.

The modal is read only, no other interaction happens.

If a skill is linked to a profile, but has no experiences attached it shows a null state.

## 🎨 Design File

## 📸 Screenshot

https://github.com/GCTC-NTGC/gc-digital-talent/assets/16977254/3be61cce-b3ee-4d14-9cc3-235956378388

image

image

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] The skills that show on the modal are the skills used as filters applied to the table NOT the skills submitted on the request itself
- [ ] If a skill is linked to a profile, but has no experiences attached it shows a null state
- [ ] In the dialog use the skill UI from the profile snapshot view

yonikid15:
🤖 Resolves #6658 

## 👋 Introduction

- Adds a dialog that opens up from the skill match column that includes the experiences tied to the skills requested (the skills selected in the filter).
- This would make it easier to find the right candidate by allowing DCM to view the relevant skill information without leaving the request page.
- Clicking on the skill match number opens up a dialog that showcases ONLY the relevant experiences to the skills selected for that individual candidate.

## 🕵️ Details

- Adds a query to fetch all skills and creates a list of "filteredSkills" from the filters chosen skill ids. I went down a few different routes to complete this. but ended up choosing this way. The other being changing the `ApplicantFilter` type in the schema to have `skills` return `[Skill]` instead of `[InputId]`. That

## 🧪 Testing

1. Sign in as admin. 
2. Go to the Requests table on the Admin side.
3. View a request and select the "number of skills matched". Ensure dialog opens with skills selected.

## 📸 Screenshot

image

image


apps/web/src/components/PoolCandidatesTable/SkillMatchDialog.tsx
```diff
@@ -0,0 +1,95 @@
...
+};
+
+export default SkillMatchDialog;
+
+export function skillMatchDialogAccessor(
+  filteredSkills: Skill[],
+  experiences: Experience[],
+  skillCount: Maybe<number>,
+) {
+  if (filteredSkills.length === 0) return <p>0/0</p>;
```
esizer, line 87: @mnigh Should this be wrapped in an `intl.formatMessage`? I'm worried "0/0" is not universally correct.

mnigh: Not sure. If this format did not work in other languages, then how would `npcY4+` (`{skillsCount}/{filteredSkillsTotal}`)? Is there a language you are familiar with where this wouldn't work? A safer format would probably be `0 of 0` and `{skillsCount} of {filteredSkillsTotal}` though and then wrap that in `intl.formatMessage`.

esizer: I am not aware of any, was just checking to see if you knew. I agree that `0 of 0` reads a bit better. 

---

Oh! Even more problematic is the buttons are not unique 😬 We may want to rethink the text for these 😓 

"Show 1 user skill of 4 selected skills" as an `aria-label` perhaps? 😿

apps/web/src/components/PoolCandidatesTable/SkillMatchDialog.tsx
```diff
@@ -0,0 +1,138 @@
...
+              disclaimerMessage={intl.formatMessage({
+                defaultMessage:
+                  "There are no experiences attached to this skill",
+                id: "pJqoQF",
+                description:
+                  "Disclaimer displayed in the skill tree on the skill match dialog.",
+              })}
+              hideConnectButton
+              hideEdit
+            />
```
esizer, line 102: sorry, totally missed this on the first pass. Can we add `headingAs="h3"` right now we jump from `h2` in the title to `h4` in the tree 😬

esizer (changes requested):
Sorry 🙇

apps/web/src/components/PoolCandidatesTable/SkillMatchDialog.tsx
yonikid15: Nice catch! Fixed by [c32a80b](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7015/commits/c32a80b205504abe8258c4899130e17ae92fc41d)

apps/web/src/components/PoolCandidatesTable/SkillMatchDialog.tsx
mnigh: @yonikid15 I think that makes sense, but you might want to check with @Jerryescandon on this since the designs will need to be updated to reflect these changes.

esizer (changes requested):
Sorry, I did not catch this because it is somewhere else in the code and hasn't come up until now.

We will need to update `<SkillTree />` to increment the heading based on the `headingAs` prop. Peter made a handy little function to do this in the `ui` package.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/39f20da8a4443e9755987856f0ad1ff45b14740b/apps/web/src/pages/Applications/ApplicationSkillsPage/components/SkillTree.tsx#L117

https://github.com/GCTC-NTGC/gc-digital-talent/blob/39f20da8a4443e9755987856f0ad1ff45b14740b/packages/ui/src/utils.ts#L3-L8

An example of it in use can be found in the `ExperienceCard` component.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/39f20da8a4443e9755987856f0ad1ff45b14740b/apps/web/src/components/ExperienceCard/ExperienceCard.tsx#L61

apps/web/src/components/PoolCandidatesTable/SkillMatchDialog.tsx
yonikid15: I talked to Jerbo and he approved, and the designs will get updated 👍. Also, I should get the translations sometime today (hopefully).

yonikid15:
> Sorry, I did not catch this because it is somewhere else in the code and hasn't come up until now.
> 
> We will need to update `<SkillTree />` to increment the heading based on the `headingAs` prop. Peter made a handy little function to do this in the `ui` package.
>

Nice catch. I believe I have fixed what you mentioned in [4b354b0](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7015/commits/4b354b0b0e415494f8abdd986f744c3e74cd9d01)

esizer (approved):

PR #7020, opened 2023-06-20
[Fix] Hide recruitment flag from accessibility tree

### LINKED ISSUE #6952: Remove tags before heading 3 above job titles
### Description

On browse jobs page. The classification appears before the job title in the reading order. This leads to confusing reading from the screen reader.

Image

> Remove the tag s that read before the Heading 3 titles for each job title. They don't make any sense read before the job title for screen readers. ."WCAG criteria: 1.3.1 Info and Relationships, 1.3.2 Meaningful Sequence
> 
> They say things like "ai-01," abbreviations can sometimes be confusing or misinterpreted unless they are explained). A link to a help section with the means would improve the accessibility for those using screen readers as sometimes screen readers like VoiceOver will try to read the letters as actual wordsWhen tabbing through the page, sometimes odd behaviour would occur on the links. Something about a missing abbreviation. Apply to this recruitment (Abbreviation not found.) Currently, the apply link only tells you what section the job opportunity is in. It would be good to repeat the job title as well.
> 

### Acceptance Criteria
- [ ] **Aria-hidden = true** on the recruitment flag (that shows the classification - circled in red in image)

esizer:
🤖 Resolves #6952 

## 👋 Introduction

This adds `aria-hidden="true"` to the recruitment flag on the browse jobs page.

## 🕵️ Details

The flag was being read to screen reader users before the heading, causing some users confusion. Since the content of the flag also appears in the title, we have opted to hide it from the accessibility tree since no information is lost in doing so.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to `/browse/pools`
3. Inspect the flag on pool cards in the accessibility tree
4. Confirm it is ignored
5. Test using a screen reader
6. Confirm the contents of the flag are ignored

### DIFF
--- a/apps/web/src/pages/Pools/BrowsePoolsPage/components/PoolCard/PoolCard.tsx
+++ b/apps/web/src/pages/Pools/BrowsePoolsPage/components/PoolCard/PoolCard.tsx
@@ -70,6 +70,7 @@ const PoolCard = ({ pool, headingLevel = "h3" }: PoolCardProps) => {
       data-h2-radius="base(rounded)"
     >
       <div
+        aria-hidden="true"
         data-h2-position="base(absolute)"
         data-h2-location="base(0, auto, auto, x.5) p-tablet(0, auto, auto, x2)"
       >


mnigh (approved):
## 🧪 Test Details

- macOS 12.6.6
- Chrome 114.0.5735.133
- VoiceOver

### Screen recording
https://github.com/GCTC-NTGC/gc-digital-talent/assets/3046459/8c1edcac-dc61-4809-acb6-8349aad97d9b


### Screenshot
Screen Shot 2023-06-20 at 15 07 46

PR #7032, opened 2023-06-21
[Bug] AwardContent date displayed fix

### LINKED ISSUE #7000: 🐛 Date parser incorrectly parsing `Date`
## 🐛 Bug

Currently our date parser is parsing only `DateTime` causing issues with `Date`'s from the API. 

## 🦋 Expected Behaviour

Parser properly handles `Date` objects.

## 🕵️ Details

Since a `Date` has no time associated with it, the parser takes it as midnight then converts it from `UTC` to local time. Any negative timezone will convert the date to previous day.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/bb219bd81dd431194f7b426721fa08923a1290db/packages/date-helpers/src/index.ts#L150-L152

## 📋 Steps to Reproduce

1. Create an experience with date from the first day of the month
2. Check the card and confirm that date appears on the last day of the previous month

## 📸 Screenshot
246448138-a9d162ad-de07-418f-b9eb-d5a4e9459cd0

## 🙋‍♀️ Proposed Solution

1. Different parser for `Date`
2. Allow the parser to handle both `Date` and `DateTime`

## ✅ Acceptance Criteria

- [ ] `Date` from the API is parsed and formatted correctly
- [ ] `DateTime` still functions as expected

vd1992:
🤖 Resolves #7000 

## 👋 Introduction

Resolve the date inconsistency for award experiences. 

## 🕵️ Details

I just mimicked how dates were rendered for the other experiences on the same page, plus, since you couldn't pick the day itself the card shouldn't display a day. 
It also seems like we already had all the needed functions present. 

Used `formattedDate()` to align with other experiences

image

image

image

## 🧪 Testing

1. Navigate to `/resume-and-recruitments`
2. Create an award experience and assert the date displayed is correct 

## 📸 Screenshot

image

image

### DIFF
--- a/apps/web/src/components/ExperienceCard/AwardContent.tsx
+++ b/apps/web/src/components/ExperienceCard/AwardContent.tsx
@@ -1,7 +1,6 @@
 import React from "react";
 import { useIntl } from "react-intl";
 
-import { formatDate, parseDateTimeUtc } from "@gc-digital-talent/date-helpers";
 import {
   commonMessages,
   getAwardedScope,
@@ -9,6 +8,7 @@ import {
 } from "@gc-digital-talent/i18n";
 
 import { AwardExperience } from "~/api/generated";
+import { formattedDate } from "~/utils/dateUtils";
 import ContentSection from "./ContentSection";
 import { ContentProps } from "./types";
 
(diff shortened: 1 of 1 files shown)


esizer (approved):
Looks great! 👍

PR #7087, opened 2023-06-23
[Fix] Remove Pool keys, which weren't used and could be invalid

### LINKED ISSUE #7077: 🐛 Talent Search page is crashing on UAT
## 🐛 Bug

On UAT, Talent Search page is just showing that a GraphQL error happened. I've traced this to an error in the `publishedPools` query, which returns pools, but also returns an internal error.

## 🦋 Expected Behaviour

Talent Search page should work fine on all verticals.
publishedPools query should not cause an internal server error!

## 🕵️ Details

This has been reproduced on and off the TBS network.

The error:
```
  "errors": [
    {
      "message": "Internal server error",
      "extensions": {
        "category": "internal"
      },
      "locations": [
        {
          "line": 7,
          "column": 5
        }
      ],
      "path": [
        "publishedPools",
        14,
        "key"
      ]
    }
  ],
```
## 📋 Steps to Reproduce

1. Go to https://uat-talentcloud.tbs-sct.gc.ca/en/search

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

The error makes it seem like there's an illegal key string in one of the pools. We should look at the pool keys in the database, and if that's the case, determine how this happened so it doesn't happen again.

## ✅ Acceptance Criteria

- [ ] publishedPool does not cause an internal server error.
- [ ] We're confident this will not happen again on UAT or on Prod

tristan-orourke:
🤖 Resolves #7077

## 👋 Introduction

The issue was caused by the DuplicatePool mutation, which appended a timestamp to the end of the original pool's key to keep it unique. However, it did this even when the original key was null (which is bascally always), creating a key which starts with underscore. That isn't a valid KeyString, so GraphQL crashed if you requested the key value for Pools (which we did on the Talent Search page).

## 🕵️ Details

I initially fixed the bug in DuplicatePool.php. However... we're not using Pool.key at all. For most of our pools (in production and UAT) it's null anyway. We have no intention of making use of it - publishing group has replaced its original purpose.

So I've removed Pool.key entirely. Good riddance!

## 🧪 Testing

To be thorough, we want to test that the pool interactions still work without keys.

1. Create a pool on the admin side. Fill out the details.
2. Duplicate the pool. Publish the duplicate.
3. Apply to the duplicate pool.
4. On the admin side, qualifiy the application.
5. Go to Talent Search, do a search which you can be sure includes your application.
6. Make sure you never see any graphql errors through this process.

## 📸 Screenshot

### DIFF
--- a/api/app/GraphQL/Mutations/DuplicatePool.php
+++ b/api/app/GraphQL/Mutations/DuplicatePool.php
@@ -22,7 +22,6 @@ final class DuplicatePool
                 'en' => $pool->name['en'] . ' (copy)',
                 'fr' => $pool->name['fr'] . ' (copie)',
             ],
-            'key' => $pool->key ? $pool->key . '_' . time() : null, // If duplicating a pool with a non-null key, ensure the new key is unique.
             'closing_date' => null,
             'published_at' => null,
         ]);
--- a/api/app/GraphQL/Mutations/DuplicatePool.php
+++ b/api/app/GraphQL/Mutations/DuplicatePool.php
@@ -3,7 +3,6 @@
 namespace App\GraphQL\Mutations;
 
 use App\Models\Pool;
-use Database\Helpers\KeyStringHelpers;
 
 final class DuplicatePool
 {
--- a/api/app/GraphQL/Mutations/DuplicatePool.php
+++ b/api/app/GraphQL/Mutations/DuplicatePool.php
@@ -3,6 +3,7 @@
 namespace App\GraphQL\Mutations;
 
 use App\Models\Pool;
+use Database\Helpers\KeyStringHelpers;
 
 final class DuplicatePool
 {
@@ -21,7 +22,7 @@ final class DuplicatePool
                 'en' => $pool->name['en'] . ' (copy)',
                 'fr' => $pool->name['fr'] . ' (copie)',
             ],
-            'key' => $pool->key . '_' . time(), // Ensure unique key
+            'key' => $pool->key ? $pool->key . '_' . time() : null, // If duplicating a pool with a non-null key, ensure the new key is unique.
             'closing_date' => null,
             'published_at' => null,
         ]);
--- a/api/app/Models/Pool.php
+++ b/api/app/Models/Pool.php
@@ -18,7 +18,6 @@ use Illuminate\Support\Facades\Auth;
  *
  * @property string $id
  * @property array $name
- * @property string $key
  * @property int $user_id
  * @property array $operational_requirements
  * @property array $key_tasks
@@ -69,7 +68,6 @@ class Pool extends Model
         'closing_date',
         'published_at',
         'name',
-        'key',
         'key_tasks',
         'stream',
         'security_clearance',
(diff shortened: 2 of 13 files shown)


esizer (changes requested):
Looks like we need to update the `PoolSeeder.php` as well.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/d1e71e17984ead1a8d6ff4b311a8a53ee15e4783/api/database/seeders/PoolSeeder.php#L27

api/app/GraphQL/Mutations/DuplicatePool.php
```diff
@@ -3,6 +3,7 @@
 namespace App\GraphQL\Mutations;
 
 use App\Models\Pool;
+use Database\Helpers\KeyStringHelpers;
```
vd1992, line 6: This seems like an accidental addition?

vd1992 (commented):
Yeah, as Eric mentioned PoolSeeder references the key field multiple times.
I do believe, from my own poking about, that the duplicate pool function is the source of the issue.

I feel pool keys in prod may be something to keep an eye on until this goes through, either that or this is a strong candidate for pushing through as a patch to to prevent a live site problem.

api/app/GraphQL/Mutations/DuplicatePool.php
tristan-orourke: Leftover from an intermediate commit https://github.com/GCTC-NTGC/gc-digital-talent/pull/7087/commits/5479a16ee5c539c4058d7fa05a5f9fe215ebd2a1

esizer (changes requested):
These dang keys keep popping up all over the place 😅 I _think_  that is all of them.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/d4435ec018e663df77f409bbb97b9d2dc8d806f3/api/database/seeders/PoolSeederUat.php#L23

https://github.com/GCTC-NTGC/gc-digital-talent/blob/d4435ec018e663df77f409bbb97b9d2dc8d806f3/api/tests/Feature/ApplicantFilterTest.php#L291

https://github.com/GCTC-NTGC/gc-digital-talent/blob/d4435ec018e663df77f409bbb97b9d2dc8d806f3/api/tests/Feature/ApplicantFilterTest.php#L342

https://github.com/GCTC-NTGC/gc-digital-talent/blob/d4435ec018e663df77f409bbb97b9d2dc8d806f3/apps/web/src/pages/Applications/applicationOperations.graphql#L258

vd1992:
> These dang keys keep popping up all over the place 😅 I _think_ that is all of them.
> 

Whack-a-mole

I don't think 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/d4435ec018e663df77f409bbb97b9d2dc8d806f3/apps/web/src/pages/Applications/applicationOperations.graphql#L258

Is an issue, it is nested within the skills field

https://github.com/GCTC-NTGC/gc-digital-talent/blob/d4435ec018e663df77f409bbb97b9d2dc8d806f3/apps/web/src/pages/Applications/applicationOperations.graphql#L222C1-L224C8

tristan-orourke:
Thanks! I made sure to run testing locally this time 😓

esizer (approved):
Nice! Working well for me 👍

PR #7096, opened 2023-06-26
[Test] IAP application e2e

### LINKED ISSUE #6824: test
Looking good! Some minor suggestions.

I think this is out of scope but it would be nice to have a test for both a pool with no screening questions as well as one that is under the IAP publishing group too. Do you know if we have tickets for those yet?

_Originally posted by @esizer in https://github.com/GCTC-NTGC/gc-digital-talent/pull/6748#pullrequestreview-1463250234_

petertgiles:
🤖 Resolves #6824 

## 👋 Introduction

This branch adds an e2e test for the IAP application

## 🧪 Testing

1. Build the app
2. Run the e2e test in Cypress (apps/e2e/cypress/e2e/talentsearch/submit-application-iap.cy.js)
3. Test is reliable
4. Test covers AC

## 📸 Screenshot

image


apps/e2e/cypress/e2e/talentsearch/submit-application-iap.cy.js
```diff
@@ -0,0 +1,299 @@
...
+        .click();
+    });
+
+    // Welcome page - step one
+    cy.findByRole("heading", {
+      name: "Apply to Cypress Test Pool EN",
+      level: 1,
+    })
+      .should("exist")
+      .and("be.visible"); // IAP heading does not include job stream
```
esizer, line 123: We may want to wait on the create application mutation before this. Mine is checking before the page renders 😬

esizer (changes requested):
I couldn't get past the initial creation of the application so that is as far as I got. We might need to sprinkle some `cy.wait()`'s in for slower runs.

apps/e2e/cypress/e2e/talentsearch/submit-application-iap.cy.js
petertgiles: OK, adding waits back in.
Ref: [add query and mutation waits](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7096/commits/e4c6eb778853db4d481ba91d3adcdbbf1b702e5d)

apps/e2e/cypress/e2e/talentsearch/submit-application-iap.cy.js
```diff
@@ -0,0 +1,338 @@
...
+      name: /signature/i,
+    }).type(uniqueTestId);
+    cy.findByRole("button", { name: /Sign and continue/i }).click();
+    cy.wait("@gqlUpdateUserAndApplicationMutation");
+
+    // Review profile page - step three
+    cy.wait("@gqlGetApplicationQuery");
+    cy.findByRole("heading", { name: /Review your profile/i })
+      .should("exist")
+      .and("be.visible");
```
esizer, line 196: I got further 😅 Looks like this one depends on the `getMe` query so we should also wait on that.

esizer (changes requested):
Making progress, I think the specific queries we wait on need to be looked into now. This might be hard 🤔 

If you don't want to play whack-a-mole with the waits I can probably take a look since I can get it to reliably fail.

apps/e2e/cypress/e2e/talentsearch/submit-application-iap.cy.js
petertgiles: Hmm, I'm a bit confused now.  Is the regular submit-application.cy.js failing for you as well?  What is happening, exactly?  Is the `findByRole` timing out after x seconds while the page slowly loads?

esizer: Yeah, the `findByRole` times out because the `getMe` query is still resolving and the heading is actually not rendered until that query completes. 

I will check the other spec and let you know 👍

petertgiles: OK, I removed the waits and bumped the command timeout way up.  I think this is more realistic to the way a user would interact with the site, instead of waiting for each Graphql request one-by-one.

esizer (approved):
Awesome! ⭐ 

I like the change to the timeout instead of having to add wait's all over the place.

Issue #7099, opened 2023-06-26
♻️ Merge Aplicant type into User


petertgiles:
## ♻️ Debt/Refactor

We're trying to simplify our schema by removing duplicate types.  Let's remove the Applicant type and merge it into User.

## ✅ Acceptance Criteria

- [ ] Applicant type is gone
- [ ] User type used everywhere

petertgiles:
This issue is already pretty huge.  Could you take a quick looks while you work on it if it would be reasonable to remove `UserPublicProfile` at the same time?

esizer:
> This issue is already pretty huge. Could you take a quick looks while you work on it if it would be reasonable to remove `UserPublicProfile` at the same time?

That shouldn't be an issue with field level permissions. I'll hopefully have an answer for standup tomorrow 😄

PR #7102, opened 2023-06-27
[Fix] Separate query for experiences

### LINKED ISSUE #7014: 🐛 Application - have to refresh after adding an experience
## 📋 Steps to Reproduce

During an application:
1. Go to step 5 (Skill requirements)
2. Add ALL your available experiences to a skill
3. After adding all experiences the button to add another exp. disappears
4. Go to step 3
5. Add an experience
6. Go back to step 5
7. Button still missing even though we have more experiences.
8. Refreshing the page fixes this, but it would be nice to have it working without it.

## ✅ Acceptance Criteria

- [ ] Do not need to manually refresh the page after adding an experience to tie a skill

petertgiles:
🤖 Resolves #7014

## 👋 Introduction

This branch separates the query for the experiences on the skill requirements application page.  It mirrors the component on the resume page now.

## 🕵️ Details

It's not clear to me why cache invalidation isn't working on this page when the experiences are updated.  My guess is that it might be due to "User" and "Applicant" being different models.  I opened  #7099 to address that.

## 🧪 Testing

1. Build the app.
2. Run through the recreation steps in the issue.

### DIFF
--- a/apps/web/src/pages/Applications/ApplicationSkillsPage/ApplicationSkillsPage.tsx
+++ b/apps/web/src/pages/Applications/ApplicationSkillsPage/ApplicationSkillsPage.tsx
@@ -1,10 +1,17 @@
 import React from "react";
 import { useIntl } from "react-intl";
 import { FormProvider, useForm } from "react-hook-form";
-import { useNavigate } from "react-router-dom";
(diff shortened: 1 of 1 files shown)


esizer (approved):
Works great! 👍  Too bad about the cache 😢

PR #7109, opened 2023-06-27
[Dep] Hydrogen 2.0.2

substrae:
## ✨ Summary

Updates Hydrogen to patch `2.0.2`. Adjustments you'll notice right away:
- More concise console output
- No more timer settings (timers are inline with their respective output and are hidden if the output is hidden)
- Typography theme settings now have options for individually specified `line-height` values
  - This will cause a larger visual diff because titles will all adjust ever so slightly to use tighter line-heights

Bonus things:
- @petertgiles and @vd1992 mentioned properly returned exit codes, so you'll see the build fail if Hydrogen completes with errors now

You can learn more about the full release here: https://github.com/hydrogen-css/hydrogen/releases/tag/v2.0.2

## 🗃️ Requirements

- Hydrogen compiles same as ever
- Hydrogen's console output indicates v2.0.2
- There is a visual diff for headings

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -76,7 +76,7 @@
     "@graphql-codegen/typescript": "^3.0.4",
     "@graphql-codegen/typescript-operations": "^3.0.4",
     "@graphql-codegen/typescript-urql": "^3.7.3",
-    "@hydrogen-css/hydrogen": "2.0.1",
+    "@hydrogen-css/hydrogen": "2.0.2",
     "@mdx-js/react": "^2.3.0",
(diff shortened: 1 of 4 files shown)


esizer (approved):
The line height changes in the diff look good 👍 . I'll need to leave the exit codes to Peter to see if it was what they were looking for.

petertgiles (approved):
Nice!  Big Chromatic diff but nothing broke that I can see.  Exit code working well.  :+1:

substrae:
Woop woop, thanks gents!

PR #7154, opened 2023-06-30
[Feature] Nginx Amplify

### LINKED ISSUE #6833: 🛠️ Set Up Nginx Amplify
## 🛠️ Tooling

Nginx Amplify is a tool for monitoring the Nginx web server.  It has a free tier that we can use.

## 🕵️ Details

https://amplify.nginx.com/docs/

## ✅ Acceptance Criteria

- [ ] Create a shared account to administer the dashboard (free tier only allows 3 users)
  - Set the email address to a Slack channel, maybe?  Slack note or dollarama padlock for the password?
- [ ] Update the deploy process to start the agent in Azure, separately for the three verticals (free tier allows 5 agents)
- [ ] Set up the Slack integration
- [ ] Set up some reasonable alerts
- [ ] Create issues for any config analyzer problems

petertgiles:
🤖 Resolves #6833 

## 👋 Introduction

This PR updates our post-deploy script to install the Nginx Amplify agent for telemetry collection.  Check the Dollarama Padlock for the account details.

## 🕵️ Details

To enable this some changes were made:
- The nginx.conf file was customized with logging to provide extra telemetry for the Amplify agent
- A [stub_status location](http://nginx.org/en/docs/http/ngx_http_stub_status_module.html) was added to provide extra telemetry for the Amplify agent
- The Nginx config files were organized into a single directory
- A setup script was added for Nginx Amplify setup.

## :thinking: Thoughts
Installing this requires
  1. running apt-get update with `--allow-releaseinfo-change`
  2. installing sudo and python3
  3. installing a log rotation tool
  4. adding an internal location endpoint to Nginx

Running this requires
1. some amount of RAM, CPU, and network

I _think_ this is all secure but I am a bit uncomfortable with how heavy this is and how it increases deployment time by a minute or so.  Personally I think this telemetry is very valuable and is worth it but I recognize that it isn't a no-brainer.

I really we wish we had a better solution for storing shared accounts than Dollarama Padlock.  I put it there for now but it would be better if we had a team password manager.  Or maybe we'll get Slack Canvas soon and can store some passwords there.  If we find this tool useful we might get a paid account which would lift the 3 account limit.  We wouldn't have to share accounts then.

I'm not totally clear on how the log rotation works.  I'm going to check how dev looks in a few days.  I added this tool in because I didn't want to bring down our appsvc by filling it up with logs.  :grimacing: 

### Locally

It's possible but a bit tricky to test this locally.

1. Add the Amplify API key and a website name to your Docker Compose file.  Remove your webserver container and rebuild it to apply the changes.
2. Manually run the `setup_nginx.sh` and `setup_nginx_amplify.sh` scripts in the container, either by shelling in our running from docker-compose.
3. Undo the changes to the site-default-local file that the scripts make.
4. Log into Nginx Amplify and wait a few minutes for your telemetry to start appearing.
6. When you're done, remove your webserver container and then remove your host from Amplify so we don't get warnings that your site is down.  :smile: 

### In Azure

This is already deployed in dev but you can rebuild the branch and redeploy it if you like. Make sure the IaC of the appsvc is >= Release-459.

## :truck: Deployment
- A new API key was added to the IaC.  Make sure the appsvc is >= Release-459.

## 📸 Screenshot

image

## :fist_right:  Next Steps
- I was never able to get the PHP-FPM monitoring going. :cry: #7156
- The static analysis tool has identified a few issues with our configuration.  #7155


tristan-orourke (commented):
The code looks good, but as you point out, there's actually a lot going on here! Some thoughts:
- If we go through with this, is there any way we could we move the installation of python and other dependencies into the build artifact step (instead of the post-deploy script) to avoid lengthening the deploy downtime?
- I'm a bit surprised this just... works? I thought we'd need to talk to IMTD about whitelisting stuff in the F5 firewall. 
- Your caveats about how heavy this is makes me want to reconsider how much of this functionality we can get directly in Azure. I was looking at portal.azure.com this morning, and while it may not give us as much insight Nginx's inner workings, there's a lot of info we can find in the Alerts, the Metrics, and the Logs tools, if we take the time to learn how to use them. Looks like we could even set up email alerts for things like response times or error counts going above a threshold.

petertgiles:
Will investigate Azure tooling in  #7162 instead.

PR #7163, opened 2023-07-04
[Tooling] Sort fr.json files

### LINKED ISSUE #6964: 🛠️ Avoid merge conflicts on fr.json files
## 🛠️ Tooling

We often have simple merge conflicts on our fr.json files.  If two different branches add new translation messages to the bottom of the files then Github doesn't know how to resolve the conflict.

## 🙋‍♀️ Proposed Solution

Some ideas discussed:
- add a trailing comma to the last message in the array
- add a newline after the last message in the array
- add new message to a random position in the array
- sort the array

## ✅ Acceptance Criteria

- [ ] A branch with a new message, based before the head with a different message added at the end, does not cause a merge conflict

petertgiles:
🤖 Resolves #6964

## 👋 Introduction

This branch updates the `cli.js` file to output sorted JSON files which should reduce the number of merge conflicts we experience with the files.

## 🕵️ Details

I found [this Stackoverflow answer](https://stackoverflow.com/a/16168003) about sorting stringified JS objects.  Since it's a dev dependency it seemed reasonable to use [json-stable-stringify](https://www.npmjs.com/package/json-stable-stringify) as suggested.

I also updated the `cli.js` script (and associated docs and workflow) to always write the output file to ensure we're keeping up with the sorting.

> __Note__
The Indigenous language files do not use the `cli.js` files and so are out of scope.

> __Warning__
I was surprised to notice that new translations that match the original properly show the "add it to the whitelist" notice but are added to the `fr.json` file anyway.  It works like this on main as well.  Is this the intended behaviour or is it a bug?

## 🧪 Testing

Check that all the existing translation use cases work as expected, but with a sorted file.

- [ ] adding new translations
- [ ] finding orphaned translations
- [ ] handling whitelisted translations
- [ ] sorting an unsorted file
- [ ] quick edits of the fr.json files


yonikid15 (approved):
- [x] adding new translations 
- [x] finding orphaned translations
- [x] handling whitelisted translations
- [x] sorting an unsorted file
- [x] quick edits of the `fr.json` files

Tested everything a couple times and it's all working as expected 🎉 
Also, the new sorting is working great.

PR #7210, opened 2023-07-06
[Feature] Change request statuses

### LINKED ISSUE #6601: ✨ Request status upgrade
## ✨ Feature

DCM needs more granularity when tracking requests, they want to add more statuses to accurately track and manage the progress of any request.

## 🕵️ Details

Replace the status button "Pending > Done" for a dropdown and a save button that would allow different statuses to be selected.

We'll also need to update the filter added in #6365 

## 🎨 Design File

Include a[ link to the design file](https://www.figma.com/file/dDo0sUzYFyplp7DvlLHKYJ/Admin---Requests-details?type=design&node-id=104%3A493&mode=design&t=xdbyzY5g5fzcXaw3-1)

## 📸 Screenshot

Instead of "Other (see notes)", use "Waiting (see notes)" for the blocked status.

Image

Add startedAt and blockedAt fields to the request.

Conceptually, the status dates follow an order: startedAt -> blockedAt (optional) -> doneAt. The latest non-null value determines what we consider its status to be. If all three are null, the request is Pending. If blockedAt is non-null, the status is Waiting, regardless of whether startedAt is null or not. 

When the status changes, the date for that status is reset, and all 'later' dates are nullified.

We could have a computed column which keeps the final status up to date, as we've done elsewhere.

OR

Filtering and ordering by the status column might be vastly simpler if we switch to a single STATUS enum column, and track changes with some kind of [history table](https://medium.com/sammich-shop/simple-record-history-tracking-with-laravel-observers-48a2e3c5698b) if we really want.

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Request Responder users can change request status between the 4 shown statuses
- [ ] Status appears on the request table. Ordering by status column orders rows in intuitive order (pending, started, blocked, done; or the reverse)
- [ ] Requests can be filtered by all statuses, on the Requests table

vd1992:
🤖 Resolves #6601

## 👋 Introduction

Updates the available statuses for search requests. 

## 🕵️ Details

Adds status enums, as well as a column for sorting in a particular order. Migration and factory changes. 
Replaces `done_at` with `status_changed_at` 
Can filter by and sort on the new statuses, as well as update them on the view request page. 

## 🧪 Testing

1. Migrate and seed fresh
2. Submit request from `/en/search`
3. Confirm you can find your request at `/en/admin/talent-requests`
4. Sorting and filtering requests works here
5. View a request
6. Confirm updating works

## 📸 Screenshot

image

image


api/database/migrations/2023_07_05_214529_add_request_status_column.php
```diff
@@ -0,0 +1,63 @@
...
+                    GENERATED ALWAYS AS
+                        (case
+                            when request_status = 'NEW' then 10
+                            when request_status = 'IN_PROGRESS' then 20
+                            when request_status = 'WAITING' then 30
+                            when request_status = 'DONE' then 40
+                        end)
+                STORED;
+                SQL,
+        );
```
vd1992, line 46: Couldn't quite get variables to work here like the bit above it, got stuff about "could not determine data type of parameter". Maybe due to the `case` block being in parentheses.

apps/web/src/pages/SearchRequests/ViewSearchRequestPage/components/UpdateSearchRequest.tsx
```diff
@@ -78,112 +86,170 @@ export const UpdateSearchRequestForm = ({
       });
   };
 
-  const handleStatusChangeToDone: SubmitHandler<FormValues> = async (
+  const handleSaveStatus: SubmitHandler<FormValues> = async (
```
vd1992, line 89: I deleted `handleStatusChangeToDone` and added in `handleSaveStatus`, the git difference decided to make it look rough

apps/web/src/pages/SearchRequests/ViewSearchRequestPage/components/ViewSearchRequest.tsx
```diff
@@ -164,23 +164,22 @@ const ManagerInfo = ({
                 />
                 <FilterBlock
                   title={intl.formatMessage({
-                    defaultMessage: "Date done",
-                    id: "BAzKWq",
+                    defaultMessage: "Status last changed",
+                    id: "pAW6AU",
```
vd1992, line 168: Not entirely sure what to do here, should I drop this block altogether instead?

api/database/factories/PoolCandidateSearchRequestFactory.php
esizer, line 1: Should we mix in some `status_changed_at` dates here to make the data more diverse? Right now, at least for me, this column is always null.

apps/web/src/pages/SearchRequests/ViewSearchRequestPage/components/UpdateSearchRequest.tsx
```diff
@@ -4,9 +4,17 @@ import { FormProvider, SubmitHandler, useForm } from "react-hook-form";
 import { useIntl } from "react-intl";
 
 import { Button, Heading } from "@gc-digital-talent/ui";
-import { Submit, TextArea } from "@gc-digital-talent/forms";
+import {
+  SelectFieldV2,
```
esizer, line 8: Could we avoid using the `SelectFieldV2` if possible? It has many a11y problems and I'm not sure it serves any benefit here. A regular `Select` would be better imo.

apps/web/src/pages/SearchRequests/ViewSearchRequestPage/components/UpdateSearchRequest.tsx
```diff
@@ -78,112 +86,170 @@ export const UpdateSearchRequestForm = ({
...
+          }}
+          data-h2-margin="base(x1, x1, 0, 0)"
+        >
+          {intl.formatMessage({
+            defaultMessage: "Back to requests",
+            id: "O8nHiQ",
+            description:
+              "Button label displayed on the search request form which returns the user back to requests.",
+          })}
+        </Button>
```
esizer, line 250: I think this would be better as a link since all it does is navigate.

```tsx
<Link href={paths.searchRequestTable()} mode="inline" color="secondary">...</Link>
```

esizer (changes requested):
Looking great! Had a few minor suggestions. I can't speak to the SQL though since I'm not very strong in that area 😅

api/database/factories/PoolCandidateSearchRequestFactory.php
vd1992: Sure, why not https://github.com/GCTC-NTGC/gc-digital-talent/pull/7210/commits/dc5eaee35e842bebe47aa4016dd37bd110a4f88c

apps/web/src/pages/SearchRequests/ViewSearchRequestPage/components/UpdateSearchRequest.tsx
vd1992: I think it is better too in this case, no null selection https://github.com/GCTC-NTGC/gc-digital-talent/pull/7210/commits/e1fb95146b7d81f434e0e5e32ae17ee3cda3a52d

apps/web/src/pages/SearchRequests/ViewSearchRequestPage/components/UpdateSearchRequest.tsx
vd1992: Done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7210/commits/9ce5563e4df5b2b960fbf2533b0b8158f69b47f5

vd1992:
> Looking great! Had a few minor suggestions. I can't speak to the SQL though since I'm not very strong in that area 😅

I'll tag in @petertgiles too for thoughts on those

api/database/migrations/2023_07_05_214529_add_request_status_column.php
petertgiles: I think it's this way because you're making a column definition, not actually running the case statement immediately.  It's fine, I wouldn't worry about it.

api/database/migrations/2023_07_05_214529_add_request_status_column.php
```diff
@@ -0,0 +1,63 @@
...
+        Schema::table('pool_candidate_search_requests', function (Blueprint $table) {
+            $table->renameColumn('done_at', 'status_changed_at');
```
petertgiles, line 48: The related columns are _request_ status and _request_ status weight.  If it's not too painful it might be nice to call this `request_status_changed_at`.

apps/web/src/pages/SearchRequests/ViewSearchRequestPage/components/ViewSearchRequest.tsx
petertgiles: I recommend checking with @Jerryescandon  but I suggest moving the date right beside to the text label of the status to show they're strongly related.  Something like
image

petertgiles (approved):
Looks and works great!  I left a few thoughts but no blockers.

esizer (approved):

Jerryescandon:
thoughts? 🤔 

image

apps/web/src/pages/SearchRequests/ViewSearchRequestPage/components/ViewSearchRequest.tsx
vd1992: Adjusted to fit the screenshot by Jerbo below in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7210/commits/b27f175514c9a8d3e17754f93a563092b0e42926

One caveat, I preserved the datetime view since we use it elsewhere and I thought it should stay consistent

image

petertgiles: Love it!

api/database/migrations/2023_07_05_214529_add_request_status_column.php
petertgiles: Nice!

vd1992:
Various feedback addressed and translation added

petertgiles (approved):
Love it.  :+1:

PR #7215, opened 2023-07-07
[Feature] Add notes to pool candidates table

### LINKED ISSUE #6818: ✨ Notes on Requests and Pool Candidates Table
## ✨ Feature

Add a searchable Notes column to the Pool Candidates and Requests tables.

## 🕵️ Details

This is the PoolCandidates table that appears on the SingleSearchRequest page and Pools page.

Add a column for the `notes` field on the Pool Candidates. It should be hidden by default. QUESTION: do we wrap the text or cut it off with ellipses? ANSWER: cut off after a certain number of characters.

Add an input field to the filter dialog which searches on the notes column, filtering out candidates whose notes don't contain the input text. This filtering should be case insensitive.

## 📸 Screenshot

image

## ✅ Acceptance Criteria

- [ ] Notes column is added to PoolCandidates table, but is hidden by default. Notes cuts off with ellipses after a reasonable number of characters.
- [ ] Whether hidden or not, you can search by note text. This search is case insensitive.
- [ ] Any changes to the api, like a new scope, are tested in phpunit tests.
- [ ] Notes column is added to Requests table - behaves the same way as PoolCandidates table
- [ ] Add to the search bar column selector as well

esizer:
🤖 Resolves #6818 

## 👋 Introduction

Adds a notes column to the pool candidates table, in a new `<Spoiler />` component, allowing users to search by it.

### `<Spoiler />`

Added a new component to show partial content with a "read more" button to show the full text.

```tsx
export interface SpoilerProps {
  // Accessible name for the "read more" button
  linkSuffix: string;
  // Text to be truncated to read more
  text: string;
  // Character count before cutoff
  characterCount?: number;
}
```

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to the pool candidate table `/admin/pools/{poolID}/pool-candidates
3. Confirm "Notes" is hidden by default
4. Confirm you can show it using the columns dialog
5. Confirm you can open the notes field to read the full text
6. Confirm you can search for notes and results are accurate

## 📸 Screenshot

Screenshot 2023-07-07 095142

### DIFF
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -299,7 +299,7 @@ class PoolCandidate extends Model
 
         $query->whereHas('user', function ($query) use ($search) {
             User::scopeGeneralSearch($query, $search);
-        });
+        })->orWhere('notes', 'ilike', "%{$search}%");
 
         return $query;
     }
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -330,6 +330,16 @@ class PoolCandidate extends Model
         return $query;
     }
 
+    public static function scopeNotes(Builder $query, ?string $notes): Builder
+    {
+
+        if (!empty($notes)) {
+            $query->where('notes', 'ilike', "%{$notes}%");
+        }
+
+        return $query;
+    }
+
 
     public function scopePoolCandidateStatuses(Builder $query, ?array $poolCandidateStatuses): Builder
     {
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -297,9 +297,13 @@ class PoolCandidate extends Model
             return $query;
         }
 
-        $query->whereHas('user', function ($query) use ($search) {
-            User::scopeGeneralSearch($query, $search);
-        })->orWhere('notes', 'ilike', "%{$search}%");
+        $query->where(function ($query) use ($search) {
+            $query->whereHas('user', function ($query) use ($search) {
+                User::scopeGeneralSearch($query, $search);
+            })->orWhere(function ($query) use ($search) {
+                self::scopeNotes($query, $search);
+            });
+        });
 
         return $query;
     }
--- a/apps/web/src/components/SearchRequestTable/SearchRequestsTableApi.tsx
+++ b/apps/web/src/components/SearchRequestTable/SearchRequestsTableApi.tsx
@@ -114,7 +114,7 @@ const notesAccessor = (
             "Link text suffix to read more notes for a search request",
         },
         {
-          name: searchRequest.fullName,
+          name: searchRequest.jobTitle,
         },
       )}
     />
--- a/apps/web/src/components/SearchRequestTable/SearchRequestsTableApi.tsx
+++ b/apps/web/src/components/SearchRequestTable/SearchRequestsTableApi.tsx
@@ -3,7 +3,7 @@ import { IntlShape, useIntl } from "react-intl";
 import { SubmitHandler } from "react-hook-form";
 
 import { notEmpty } from "@gc-digital-talent/helpers";
-import { Pending } from "@gc-digital-talent/ui";
+import { Pending, Spoiler } from "@gc-digital-talent/ui";
 import {
   InputMaybe,
   Maybe,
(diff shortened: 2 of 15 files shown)


apps/web/src/lang/fr.json
```diff
@@ -5402,6 +5406,10 @@
     "defaultMessage": "À la prochaine fois!",
     "description": "Title for the page users land on after successful logout."
   },
+  "caSdGd": {
+    "defaultMessage": "Notes",
+    "description": "Title displayed on the Pool Candidates table notes column."
+  },
```
vd1992, line 5412: Could you maybe please use `adminMessages.notes` or some other recycling
We've already got "Notes" quite a bit already 🤣 

image

esizer: Yeah, good suggestion! @mnigh Is this a safe string to move to `adminMessages`?

vd1992: It already is in there but not used that many times I think

image

vd1992: For one word-ers, I think that description field's usefulness is debatable

esizer: Oops, did not notice that. Updated in [update to use common notes message](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7215/commits/268d4ef9433850ebef8ec7b5a398a0fb0628f101)

apps/web/src/lang/whitelist.yml
```diff
@@ -32,3 +32,4 @@
 - kvpRgN # 'Classifications' - Title for classifications
 - rwNxlI # 'Notes' - Notes label
 - EK+25s # 'Notes' - Notes in adminMessages constant
+- caSdGd # 'Notes' - Pool Candidate table column
```
vd1992, line 35: Please clean-up the whitelist too to match what translation ids exist

esizer: Oops, forgot about that 🤦‍♀️ Removed in [remove unused id's in whitelist](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7215/commits/5782e7e06d08a27e38c23c55d825f00801b16a7b)

vd1992:
The acceptance criteria mentions unit testing for any new scopes, but all you needed was appending an `orWhere` on an existing scope so I guess it really isn't needed?

It also needs to be added to the requests table, I think the search already works there making it just a matter of adding a hidden column.

image

esizer:
> The acceptance criteria mentions unit testing for any new scopes, but all you needed was appending an `orWhere` on an existing scope so I guess it really isn't needed?

Yeah, wasn't really sure about this one either 🤷‍♀️ 

> It also needs to be added to the requests table, I think the search already works there making it just a matter of adding a hidden column.

Oops, didn't see that at first. Thanks, added in [add hidden notes column to search request table](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7215/commits/e4000190db13fd83044f06803701d205ae66604e)

apps/web/src/components/SearchRequestTable/SearchRequestsTableApi.tsx
```diff
@@ -95,9 +96,30 @@ function dateCell(date: Maybe<Scalars["DateTime"]>, intl: IntlShape) {
...
+      text={searchRequest.adminNotes}
+      linkSuffix={intl.formatMessage(
+        {
+          defaultMessage: "notes for {name}",
+          id: "I73/lv",
+          description:
+            "Link text suffix to read more notes for a pool searchRequest",
+        },
+        {
+          name: searchRequest.fullName,
```
vd1992, line 117: I'm not sure if the invisible text should be the fullName field, I would've thought request job title

api/app/Models/PoolCandidate.php
vd1992, line 1: A little bit of being pedantic on my part, but I think the scope should be reused in the general search to ensure there isn't any accidental drift in how each performs

Like so `self::scope`

image

vd1992 (approved):
Consider the suggestions above, but they aren't critical.

Other than that, seems to work fine, nice! 🎉

api/app/Models/PoolCandidate.php
esizer: Makes sense! Hopefully I got this right 😅 

[update to reused preexisting scope](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7215/commits/e9500d58ff426e1833760532d0f6b1321dd1fac7)

apps/web/src/components/SearchRequestTable/SearchRequestsTableApi.tsx
esizer: Yeah, I think they offers better context. Good suggestion 👍 

[use job title for hidden notes text](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7215/commits/02c00f620e43f373545baf421aa6786349197824)

vd1992 (approved):
Still good

PR #7243, opened 2023-07-12
[Fix] Fix snapshot test

### LINKED ISSUE #7204: 🐛 SnapshotTest intermittently failing
## 🐛 Bug

The SnapshotTest PHPUnit test is intermittently failing.

## 🕵️ Details

```
  FAILED  Tests\Feature\SnapshotTest > create snapshot                         
  Failed asserting that two arrays are equal.
       'positionDuration' => Array (...)
       'experiences' => Array (...)
       'poolCandidates' => Array (
  -        1 => Array (...)
  +        0 => Array (...)
       )
       'isProfileComplete' => true
   )
  

  at tests/Feature/SnapshotTest.php:88
```

## 📋 Steps to Reproduce

1. Test tests/Feature/SnapshotTest.php

## ✅ Acceptance Criteria

- [ ] Test reliably passes

petertgiles:
🤖 Resolves #7204

## 👋 Introduction

This branch (hopefully) fixes the snapshot test that has been failing intermittently.  The solution was to use `array_values` to reindex the array back at zero after removing one of the elements.  I left in the more explicit pool and candidate initialization since I feel it makes the test more clear.  I don't think it actually fixed anything though.

## 🕵️ Details

I'm actually finding it really difficult to replicate this failure locally.  This is solution seems pretty stable.  You can check the previous commit before I removed the logging:
https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/5546440323
image

## 🧪 Testing

1. Review the code
2. Review the CI runs

### DIFF
--- a/api/tests/Feature/SnapshotTest.php
+++ b/api/tests/Feature/SnapshotTest.php
@@ -49,18 +49,16 @@ class SnapshotTest extends TestCase
         $pool1 = Pool::factory()->published()->create();
         $pool2 = Pool::factory()->published()->create();
 
-        $poolCandidate = PoolCandidate::factory()
-            ->create([
-                "user_id" => $user->id,
-                "pool_id" => $pool1->id,
-                "pool_candidate_status" => ApiEnums::CANDIDATE_STATUS_DRAFT
-            ]);
-        $poolCandidateUnrelated = PoolCandidate::factory()
-            ->create([
-                "user_id" => $user->id,
-                "pool_id" => $pool2->id,
-                "pool_candidate_status" => ApiEnums::CANDIDATE_STATUS_DRAFT
-            ]);
+        $poolCandidate = PoolCandidate::factory()->create([
+            "user_id" => $user->id,
+            "pool_id" => $pool1->id,
+            "pool_candidate_status" => ApiEnums::CANDIDATE_STATUS_DRAFT
+        ]);
+        $poolCandidateUnrelated = PoolCandidate::factory()->create([
+            "user_id" => $user->id,
+            "pool_id" => $pool2->id,
+            "pool_candidate_status" => ApiEnums::CANDIDATE_STATUS_DRAFT
+        ]);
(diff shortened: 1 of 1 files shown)


esizer (approved):
Amazing! ⭐ Thanks for playing detective on this one 😁 🕵️

api/tests/Feature/SnapshotTest.php
```diff
@@ -46,14 +46,23 @@ public function testCreateSnapshot()
...
+                "user_id" => $user->id,
+                "pool_id" => $pool1->id,
+                "pool_candidate_status" => ApiEnums::CANDIDATE_STATUS_DRAFT
+            ]);
+        $poolCandidateUnrelated = PoolCandidate::factory()
+            ->availableInSearch()
+            ->create([
+                "user_id" => $user->id,
+                "pool_id" => $pool2->id,
+                "pool_candidate_status" => ApiEnums::CANDIDATE_STATUS_DRAFT
```
vd1992, line 64: This creates an odd state.
`availableInSearch()` effectively submits it and then it is being reverted to draft. 
I don't think it impacts testing, it just creates a model state that is/should be impossible.

image

vd1992: Oh wait, hang on, should `availableInSearch` also defined a value for `submitted_at`? I think I've wandered into spinoff territory

petertgiles: Oh, you're right.  :facepalm: We don't have a factory state for draft yet but I see the factory has some conditional logic when the pool_candidate_status is DRAFT so I'll leave it there.
Ref: [remove pool candidate factory states](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7243/commits/574de73879e4f96029aeb2178c43890c0d8e90c7)

vd1992 (approved):
Ran it multiple times, seems good to me

PR #7280, opened 2023-07-13
[Fix] Remove invalid attributes from column visibility checkboxes

### LINKED ISSUE #7274: ⌨ "Toggle Column Visible" title read instead of checkbox state for admin tables
## 🕵️‍♀️ Details

**AT Configuration**
NVDA with Chrome on Windows

Tester Feedback

-"When in the columns dialogue to hide them, checkboxes are overly verbose if I jump from checkbox to checkbox with NVDA. every time I jump to the next or previous checkbox, it will read the label of the checkbox but add "Toggle Column Visible" which makes it overly verbose."

**Related WCAG criteria**
[2.4.6 Headings and Labels](https://w3c.github.io/wcag/understanding/headings-and-labels) 

## Screenshot

Image

**💻 Related WCAG Criteria**

## 🧪 Testing Steps

- Access the pools table from the admin UI using NVDA
- Navigate using the keyboard to the Columns button.
- When in the modal navigate to an unchecked (or hidden) box. 
- Observe what NVDA read.

## 🌟 Acceptance Criteria
- [ ] Remove the Title attribute from the checkboxes. The given title is currently getting read INSTEAD of the name and state of the actual checkbox.
- [ ]  This should be done for all similar tables.

esizer:
🤖 Resolves #7274 

## 👋 Introduction

Removes the `for`, `id` and `title` attributes from the client tables column visibly checkboxes.

## 🕵️ Details

These attributes were causing issues for AT users since:
 - `for` and `id` were invalid since they contained spaces which indicates multiple values
 - `title` attribute is discouraged and causes issues with ATs

Since the input is wrapped in a label, `for` and `id` were doing nothing but confusing the ATs. Since they were confused by multiple values in these attributes it started falling back eventually landing on the `title` attribute. Since the `title` was "Toggle Column Visibility" for all, the users had no idea what column they were toggling and never heard their current state.

All attributes were removed as unnecessary since wrapping an `input` with a `label` automatically associates them with each other also making the `title` attribute redundant.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to a table
3. Open the columns dialog
4. Confirm the checkboxes have the proper accessible name
5. Confirm state is announced to screen readers

### DIFF
--- a/apps/web/src/components/Table/ClientManagedTable/Table.tsx
+++ b/apps/web/src/components/Table/ClientManagedTable/Table.tsx
@@ -308,7 +308,9 @@ function Table<T extends Record<string, unknown>>({
                               </Field.Legend>
                               <div data-h2-margin="base(x.125, 0)">
                                 <IndeterminateCheckbox
-                                  {...(getToggleHideAllColumnsProps() as React.ComponentProps<
+                                  {...(getToggleHideAllColumnsProps({
+                                    title: undefined,
+                                  }) as React.ComponentProps<
                                     typeof IndeterminateCheckbox
                                   >)}
                                 />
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
Tested with Orca in Chromium.  Works great!

PR #7290, opened 2023-07-14
[Debt] Remove legacy roles

### LINKED ISSUE #5440: ✨Remove old User Roles
## ✨ Feature

Remove the LegacyRole enum and any reference to it.

## 🕵️ Details

After replacing the LegacyRole enum with our new and improved Roles and Permissions in the backend policies (#5438) and in frontend checks (#5439) we should be able to remove the enum Role entirely, and any references to it.

## 🎨 Design File

## 📸 Screenshot

Delete the enum.
Migration to remove the User.roles column.
Edits to User.php, and maybe seeders/tests?

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] The LegacyRole enum has been removed from schema
- [ ] The user.legacy_roles column has been removed from database
- [ ] There are no references in code to either of these.
- [ ] User seeder/factories assign new Roles instead of old Admin role

## :stop_sign: Blockers

List any issues which must be completed before this one.

```[tasklist]
### Blockers
- [ ] #5438 
- [ ] #5439 
- [ ] #5441 
```

vd1992:
🤖 Resolves #5440 

## 👋 Introduction

Removes legacy roles from everywhere. 

## 🕵️ Details

Migration, backend tweaks, and some frontend tweaks. 
Deletes the `RoleSync` command too, that was run months ago and I assume we definitely don't want to rerun it. 

## 🧪 Testing

1. Tests pass
2. Website builds and is useable
3. Legacy roles no longer present aside from migration files


api/database/migrations/2023_07_13_214737_drop_legacy_roles.php
vd1992, line 1: I wasn't sure which way to do this, so I did two methods. 
Eloquent and through the Facade, 

Facade I think is a little trickier to figure out, but I got it working and it performs better. 
Seeded 1000 users -> revert migration through Eloquent about 4.2 seconds, 0.05 seconds for Facade. Big difference!

packages/ui/src/components/SideMenu/SideMenuItem.tsx
vd1992, line 1: While removing legacy roles from the admin side menu stories, I noticed something off and went to dig around to find the source, turned out to be just a backwards conditional.

vd1992: Before (on main)

image

vd1992: On this branch

image

esizer:
> Deletes the `RoleSync` command too, that was run months ago and I assume we definitely don't want to rerun it.

This command was explicitly written in a way to be re-ran a few times. Should we run this one last time before deploying this change to confirm all users have new roles in prod?

vd1992:
> > Deletes the `RoleSync` command too, that was run months ago and I assume we definitely don't want to rerun it.
> 
> This command was explicitly written in a way to be re-ran a few times. Should we run this one last time before deploying this change to confirm all users have new roles in prod?

Per meeting, good to go (away)

api/database/migrations/2023_07_13_214737_drop_legacy_roles.php
esizer: Yeah, I think the facade makes sense here as well 👍

esizer (approved):
LGTM! 👍 Though, i'd delete the eloquent code commented out in the migration first 😅

vd1992:
> LGTM! 👍 Though, i'd delete the eloquent code commented out in the migration first 😅

easter egg removed in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7290/commits/f647f29472b4bd5a77a9cda1ea0ff0a56dcaa531

esizer (approved):
Nice! 🥇

PR #7313, opened 2023-07-19
[Fix] Make priority number required

### LINKED ISSUE #7220: Priority number conditionally required if "I have a priority entitlement" selected
Priority number should be required if applicant selects "I have a priority entitlement" on the "Government Information" section of the profile.  It is still optional when creating an account.

### Additional details

From Julie:

> Hi Jerbo! Can we please make the priority number field here mandatory if a candidate selects that they have priority entitlement? I think some candidates are confusing this field as EE priority entitlement (even though we link to the explanation) 

Image

### Acceptance Criteria
- [ ] "Priority number provided by the Public Service Commission of Canada" is required if applicant selects "I have a priority entitlement" on the "Government Information" section of the profile.
- [ ] Exists on profile page, application page
  - [ ] NOT a required field on create account page
- [ ] Backend and frontend validation (is step complete, is profile complete)

esizer:
🤖 Resolves #7220 

## 👋 Introduction

This makes the priority number required when the user has selected "Yes" to priority entitlement.

## 🕵️ Details

This change occurs in the application and profile pages but **_not_** create account.

## 🧪 Testing

1. Build `npm run build`
2. Login with a new account
3. Confirm priority number is not required when selected "yes" to priority entitlement
4. Navigate to your profile page
5. Confirm Confirm priority number is required when selected "yes" to priority entitlement 
6. Start and application and continue to the profile step
7. Confirm priority number is required when selected "yes" to priority entitlement 

### Create Account
Screenshot 2023-07-19 083934

### Profile / Application
Screenshot 2023-07-19 083957

### DIFF
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -178,6 +178,7 @@ class User extends Model implements Authenticatable, LaratrustUser
             ) or
             is_null($this->attributes['is_gov_employee']) or
             is_null($this->attributes['has_priority_entitlement']) or
+            ($this->attributes['has_priority_entitlement'] && is_null($this->attributes["priority_number"])) or
             is_null($this->attributes['location_preferences']) or
             empty($this->attributes['location_preferences']) or
             empty($this->attributes['position_duration'])  or
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -208,7 +208,13 @@ class User extends Model implements Authenticatable, LaratrustUser
                 $query->orWhereNotNull('looking_for_bilingual');
             });
             $query->whereNotNull('is_gov_employee');
-            $query->whereNotNull('has_priority_entitlement');
+            $query->where(function (Builder $query) {
+                $query->whereNotNull('has_priority_entitlement')
+                    ->orWhere(function (Builder $query) {
+                        $query->where('has_priority_entitlement', true)
+                            ->whereNotNull('priority_number');
+                    });
+            });
             $query->whereNotNull('location_preferences');
             $query->whereJsonLength('location_preferences', '>', 0);
             $query->whereJsonLength('position_duration', '>', 0);
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -209,7 +209,7 @@ class User extends Model implements Authenticatable, LaratrustUser
             });
             $query->whereNotNull('is_gov_employee');
             $query->where(function (Builder $query) {
-                $query->whereNotNull('has_priority_entitlement')
+                $query->where('has_priority_entitlement', false)
                     ->orWhere(function (Builder $query) {
                         $query->where('has_priority_entitlement', true)
                             ->whereNotNull('priority_number');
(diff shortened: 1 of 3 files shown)


vd1992:
Think you are missing bit 3, the validation updating

image

esizer:
> Think you are missing bit 3, the validation updating

Oops, thanks for catching that. Should be done as part of [add frontend and backend validation](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7313/commits/750f827b0e11e1dd7f05d0d552b2c4604d0dbc99)

api/app/Models/User.php
```diff
@@ -178,6 +178,7 @@ public function getIsProfileCompleteAttribute(): bool
             ) or
             is_null($this->attributes['is_gov_employee']) or
             is_null($this->attributes['has_priority_entitlement']) or
+            ($this->attributes['has_priority_entitlement'] && is_null($this->attributes["priority_number"])) or
```
vd1992, line 181: You've added it to `getIsProfileCompleteAttribute()` but how about `scopeIsProfileComplete`? The two seem quite in sync.

esizer: Huh, was no aware we had both :/ 

Added in [add priority number to profile completion scope](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7313/commits/ca738187731fed287f295ae0788361039ed76031)

vd1992:
@petertgiles @gobyrne 

This will cause some churn for the live site, users who selected have an entitlement but didn't fill in a number will now have their profiles count as incomplete suddenly. Does this merit some sort of alert, or communication in a PR Deployment section?

Although, you could also run

```
SELECT *
FROM USERS
WHERE has_priority_entitlement = true
AND priority_number is null
```

and find out if any users need to be contacted

api/app/Models/User.php
```diff
@@ -207,7 +208,13 @@ public static function scopeIsProfileComplete(Builder $query, ?bool $isProfileCo
...
             });
             $query->whereNotNull('is_gov_employee');
-            $query->whereNotNull('has_priority_entitlement');
+            $query->where(function (Builder $query) {
+                $query->whereNotNull('has_priority_entitlement')
+                    ->orWhere(function (Builder $query) {
+                        $query->where('has_priority_entitlement', true)
+                            ->whereNotNull('priority_number');
+                    });
+            });
```
vd1992, line 217: https://recordit.co/0crnaSC0G0
I'm not 100% sure on this bit 

`$query->whereNotNull('has_priority_entitlement')` 
will immediately return people with it as true but the number as null right?

Maybe instead something like
```
                $query->where('has_priority_entitlement', false)
                    ->orWhere(function (Builder $query) {
                        $query->where('has_priority_entitlement', true)
                            ->whereNotNull('priority_number');
```

esizer: You are completely right, thanks! Fixed in [fix profile complete scope](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7313/commits/426adc12dafb520421d265601f96ade9d3ace2ad)

vd1992 (approved):
- Not required on account creation
- Required for profile completeness 
- Application validation checks it (front end validation) 
- Profile completeness scopes updated (back end validation)

Things look good as far as I can tell. great!
Up to @petertgiles for what deployment stuff (if any) needs to be done

petertgiles:
If an account becomes incomplete they still appear in searches.  OK to merge.

PR #7350, opened 2023-07-20
[Dep] Bump `@radix/react-accordion`

### LINKED ISSUE #6792: ♻️ Bump @radix-ui/react-accordion from 1.1.1 to 1.1.2
## ♻️ Debt/Refactor

Manual bump of `@radix-ui/react-accordion` to `1.1.2`.

## 🕵️ Details

It appears as though there were some changes that broke our current implementation. This may be due to us rendering empty `Accordion.Root`. The error message is not very helpful however, it seems as though it is coming from the context provider creation.

https://github.com/radix-ui/primitives/blob/333eaf790a294c3af6092f2dcf870f27092d2a57/packages/react/context/src/createContext.tsx#L72

## 🙋‍♀️ Proposed Solution

Review each accordion implementation to make sure items exist before rendering. This is the possible solution:

```tsx
const items = []; // Empty array!

# Current (?)
<Accordion.Root>
   {items.map(item => <Accoridon.Item />)}
</Accordion.Root>

# Change To
{items.length ? (
  <Accordion.Root>
     {items.map(item => <Accoridon.Item />)}
  </Accordion.Root>
) : null}
```
## ✅ Acceptance Criteria

- [ ] `@radix-ui/react-accordion` on `1.1.2`
- [ ] App builds and functions with no console errors (run in dev mode)g

esizer:
🤖 Resolves #6792 

## 👋 Introduction

Bumps `@radix/react-accordion` to latest version and fixes breaking changes.

## 🕵️ Details

Seems as though there were a few issues here:

- Wrong import paths causing us to use an uncompiled version inside a compiled version
- Had an empty root which is an error now
- We were not passing the ref to the trigger which could lead to some issues

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to pages that contain accordions
3. Confirm there are no errors


esizer:
> Bonus, I cannot seem to replicate #7362 here? Was that fixed as a side effect?

I don't think so, I was able to reproduce it here as well. The difference being, it only disappears after tabbing into the accordion after closing it.

vd1992:
> > Bonus, I cannot seem to replicate #7362 here? Was that fixed as a side effect?
> 
> I don't think so, I was able to reproduce it here as well. The difference being, it only disappears after tabbing into the accordion after closing it.

Ah, okay, I was just pressing the space bar

PR #7362, opened 2023-07-21
[Fix] Accordions no longer disappear

### LINKED ISSUE #7322: 🐛 Track applications accordion disappears
## 🐛 Bug

If you use spacebar to activate the accordion on the profile page, it disappears.

## 🦋 Expected Behaviour

Hopefully, the accordion doesn't disappear ever!

## 🕵️ Details

- **Operating System**: Windows 10
- **Browser**: Chrome 114.0.5735.199

## 📋 Steps to Reproduce

1. Navigate to `/applicant/profile-and-applications`
2. Tab to the accordion and use the space bar to close it
3. Note that it disappears entirely instead

## 📸 Screenshot

Screenshot 2023-07-19 163856

## ✅ Acceptance Criteria

- [ ] The accordion always appears visible for both visual and AT users.

esizer:
🤖 Resolves #7322 

## 👋 Introduction

Fixes an issue where our accordions would disappear when closed.

## 🕵️ Details

It seems as though the animated accordions were causing some problems. In order to animate, we force the accordions to stay mounted then change opacity. For some reason, when they were hidden, users could tab in which caused a bug where the opacity was causing the entire accordion to disappear once you tabbed into it.

After exploring some fixes, I'm not sure it is straight forward. In the interest of getting this fixed, I have removed the animation from the accordions but left the code so we can revisit when we have time.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to:
    - `/seach`
    - `/applicant/profile-and-applications`
3. Make sure accordions are closed
4. Tab past the trigger
5. Confirm the next target is the next trigger and the previous accordion is still visible

### DIFF
--- a/apps/web/src/components/EmploymentEquity/EquityOptions.tsx
+++ b/apps/web/src/components/EmploymentEquity/EquityOptions.tsx
@@ -245,9 +245,7 @@ const EquityOptions = ({
                   },
                 )}
           </StandardAccordionHeader>
-          <Accordion.AnimatedContent
-            isOpen={accordionOpen === "available_options"}
-          >
+          <Accordion.Content>
             <Separator
               orientation="horizontal"
               decorative
@@ -359,7 +357,7 @@ const EquityOptions = ({
                 </p>
               </Well>
             )}
-          </Accordion.AnimatedContent>
+          </Accordion.Content>
         </Accordion.Item>
       </Accordion.Root>
     </>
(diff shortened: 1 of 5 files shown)


brindasasi:
This looks great to me.. could you update your branch @esizer ?  I just merged some changes to TrackApplicationsCard and just to make sure it works fine with that. :)

brindasasi (approved):
Works great! 🥳👏

PR #7374, opened 2023-07-24
[Bug] Application skills page validation

### LINKED ISSUE #7168: 🐛 Skill requirements validation gap
## 🐛 Bug

I was able to proceed through the submission process with a validation error showing on the Skill requirements step of the stepper.  I'm not sure exactly what was setting it off.

## 📋 Steps to Reproduce

1. Start a new application
2. Fill in the skill requirements
3. Submit application

## ✅ Acceptance Criteria

- [ ] Should not show a validation error if the step is actually OK

vd1992:
🤖 Resolves #7168

## 👋 Introduction

Resolving the validation quirk present on `/en/applications/:id/skills`

## 🕵️ Details

The validation done was different from the one used by the stepper, I changed the page to use that validator instead. 
That one ensured skill details was also filled. 

## 🧪 Testing

1. I managed to replicate the issue by doing the steps in the picture
2. Attempt to replicate the bug by doing that, it should not let you proceed until you update all the skills to have details. 

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/pages/Applications/ApplicationSkillsPage/ApplicationSkillsPage.tsx
+++ b/apps/web/src/pages/Applications/ApplicationSkillsPage/ApplicationSkillsPage.tsx
@@ -12,7 +12,6 @@ import {
   Separator,
   ThrowNotFound,
 } from "@gc-digital-talent/ui";
-import { notEmpty } from "@gc-digital-talent/helpers";
 import { toast } from "@gc-digital-talent/toast";
 import { Input } from "@gc-digital-talent/forms";
 import { useFeatureFlags } from "@gc-digital-talent/env";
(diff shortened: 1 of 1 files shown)


yonikid15 (approved):
Nice work Vachan👍 

Although this fixes the validation, it's not very clear which experiences are incomplete since at first glance all the experiences seems to be filled. It be much better to add a list of the incomplete experiences to the error message (or something more explicit). However, that's out of scope for this issue. I can make another issue for that.

vd1992:
> Nice work Vachan👍
> 
> Although this fixes the validation, it's not very clear which experiences are incomplete since at first glance all the experiences seems to be filled. It be much better to add a list of the incomplete experiences to the error message (or something more explicit).
> 
> However, that's out of scope for this issue. I can make another issue for that.

Yeah, sounds like a good idea, please do that. 
Adding a component or alert which is populated with what is registering as incomplete.

yonikid15:
> Yeah, sounds like a good idea, please do that. Adding a component or alert which is populated with what is registering as incomplete.

Okay will do 👍

PR #7398, opened 2023-07-26
[Debt] Add no unused modules lint rule

### LINKED ISSUE #7011: ♻️ Linting - No unused modules
## ♻️ Debt/Refactor

The eslint plugin _import_ has a rule for helping to detect dead code called _no-unused-modules_.  We already have this plugin installed but the rule is not configured.  Enabling it is easy but we have a lot of cleanup to do.  A lot of it is undoubtedly code that is unnecessarily exported.  Some of it is probably unused dead code.

## 🕵️ Details

Original find: https://blog.logrocket.com/how-detect-dead-code-frontend-project/
This page also references an approach that just uses TS itself.  It might be useful but the instructions here have it run as a separate CLI shortcut.

Rule reference: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-unused-modules.md

## ✅ Acceptance Criteria

- [ ] Enable the _no unused moduels_ rule.
- [ ] Do all the cleanup to get linting passing again.

esizer:
🤖 Resolves #7011 

## 👋 Introduction

This adds the eslint rule `import/no-unused-modules`  and fixes any warnings related to it.

## 🕵️ Details

This warns us when we export modules that are no being used anywhere. Combined with `no-unused-vars`, this should help surface dead code to be removed and prevent it in the future.

### `TALENTSEARCH_APP_DIR`

This was defined in a few places but we were not using it. It said it was part of storybook but I don't think it is anymore 🤔 . I see it still exists in out `azure-pipelines.yml` file. Is this necessary or can it be removed from here as well?

https://github.com/GCTC-NTGC/gc-digital-talent/blob/54bb6c750fa169aca3d80d3b8f36a8d4b4d98216/infrastructure/azure-pipelines.yml#L37

## 🧪 Testing

1. Rin lint `npm run lint -- --force`
2. Confirm no errors or warnings show up
3. Confirm the app can still be built and functions as expected


petertgiles:
Thanks for picking this up!

I'm wondering about files like utils, helpers, hooks, and types.  Other than lower order functions that are used by the main functions of those files, shouldn't everything in there be exported?  Should those files be exempt from this rule?  Or if the function is not being exported should it just be removed?  Or should we just leave them alone?  :smile:

petertgiles:
> This was defined in a few places but we were not using it. It said it was part of storybook but I don't think it is anymore thinking . I see it still exists in out azure-pipelines.yml file. Is this necessary or can it be removed from here as well?

I think you can remove it, yeah.  That path doesn't exist anymore.  If it does cause a problem with the app it'll get noticed really quickly on UAT.  :laughing:  If you want to make this change, I can deploy this branch to DEV for a look.

petertgiles (approved):
Well, it doesn't look like we got much change in bundle size like I secretly hoped.  :laughing:   The real value is here, though:
image
Looks great!  :trophy:

petertgiles:
Deployed to dev.  Seems good.  :+1:

PR #7434, opened 2023-07-27
[Dep] Bump `nuwave/lighthouse` to v6

### LINKED ISSUE #7321: ♻️ Upgrade Lighthouse `nuwave/lighthouse` to version 6
## ♻️ Refactor

Lighthouse `nuwave/lighthouse` released a new major version in March 2023. We should upgrade from version 5 to version 6!

## 🙋‍♀️ Proposed Solution

Upgrade guide: https://github.com/nuwave/lighthouse/blob/master/UPGRADE.md#v5-to-v6

## ✅ Acceptance Criteria

- [ ] application builds and runs as before

esizer:
🤖 Resolves #7321 

## 👋 Introduction

Bumps `nuwave/lighthouse` and `mll-lab/graphql-php-scalars` to v6.

### Notable Changes

- `FielMiddleware::handleField` now uses `wrapResolver` and returns `void`
- `FieldMiddleware::resolveField` now returns a callable instead of `FieldValue`
- `lighthouse.guard` is now an array and made plural (`lighthouse.guards`)
- `@delete`, `@resolve`, `@forceDelete` no longer uses `globalId` and requires a `@whereKey`
- Scalar descriptions must be typed as `?string`
- Directive methods should be typed to meet compatibility with the interfaces they implement

### Test Schema Cache

For some odd reason, schema caching was causing problems with phpunit. This was leading to two problems specifically.

1. Running out of memory
2. `UsesTestSchema` was not clearing itself on new tests

## 🧪 Testing

> **Note**
> Could be wise to target areas that use breaking changes (directives, 

1. Refresh api `docker-compose run --rm maintenance bash refresh_api.sh`
4. Rebuild frontend `npm run dev`
5. Confirm everything still works

## 🚚 Deployment

- re-running database seeders

> **Notes**
>


vd1992:
Probing around, things seem to work as far as I can tell. PHP testing and database commands working locally.
Can't weigh in on the actual changes made. 
One thing is, not exactly a showstopper, but it appears to break our API message toasts, it changes what gets returned in the graphQL error. I think that could be solved in another issue to update what message correspond to what graphQL result. 

Main

image

Branch

image

vd1992:
The API toasts would hook onto this part of the old result. 

image

image

esizer:
> One thing is, not exactly a showstopper, but it appears to break our API message toasts, it changes what gets returned in the graphQL error. I think that could be solved in another issue to update what message correspond to what graphQL result.

Looks like graphql removed `category` from the `extension`. It looks like we were relying on that but I couldn't make out how so I removed it. Does anyone know if this was crucial? 🤔 

[No longer require category extension](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7434/commits/c8cde6450b50764a4d7a11a8b3becbfb85ef3ba5)

api/config/lighthouse.php
```diff
@@ -29,84 +30,113 @@
...
+            Nuwave\Lighthouse\Http\Middleware\AcceptJson::class,
 
             // Logs in a user if they are authenticated. In contrast to Laravel's 'auth'
             // middleware, this delegates auth and permission checks to the field level.
-            \Nuwave\Lighthouse\Support\Http\Middleware\AttemptAuthentication::class,
+            Nuwave\Lighthouse\Http\Middleware\AttemptAuthentication::class,
 
             // Logs every incoming GraphQL query.
-            App\Http\Middleware\AuditQueryMiddleware::class,
+            Nuwave\Lighthouse\Http\Middleware\LogGraphQLQueries::class,
```
petertgiles, line 45: Did you mean to remove our AuditQueryMiddleware middleware?

api/config/lighthouse.php
```diff
@@ -29,84 +30,113 @@
...
+    |
+    */
 
+    'query_cache' => [
         /*
-         * The name of the cache item for the schema cache.
+         * Setting to true enables query caching.
          */
-        'key' => env('LIGHTHOUSE_CACHE_KEY', 'lighthouse-schema'),
+        'enable' => env('LIGHTHOUSE_QUERY_CACHE_ENABLE', true),
```
petertgiles, line 129: I wonder if this should be disabled locally, like the schema option.

api/config/lighthouse.php
esizer: Yeah, as far as I could tell `LogGraphQLQueries::class` (new to v6) was nearly identical so I preferred that over the custom `AuditQuryMiddleware`. If I was incorrect in this assumption, let me know and I will revert that change 👍

api/config/lighthouse.php
esizer: I would be in favour of this (cached queries in dev have bitten me before 😛 )

api/config/lighthouse.php
petertgiles: The main difference is that AuditQueryMiddleware only logs queries from admins.  If we start logging queries from everyone our log usage will shoot up, I think.

api/config/lighthouse.php
petertgiles: :100:

esizer: Updated to match other cache setting in [disable query cache locally](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7434/commits/7dda5341e2e5a8ad3fc6c60d76af0084059fa29b)

api/config/lighthouse.php
esizer: That makes total sense, thanks! Reverted back and added a comment to distinguish the two in [add query audit middleware back](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7434/commits/7ef5e9366ecb636cebdac7170c03b8f25dbf0f95)

PR #7440, opened 2023-07-28
[Feature] UserSkill add skill level and last used

### LINKED ISSUE #7380: ✨ Add skill level and currentlyUsed to UserSkills
## ✨ Feature

Add two fields to UserSkills: skill level and the "is this skill currently used or was used in the past" field (name pending).

## 🕵️ Details

We need these details to support #7378. 
Both these fields must be nullable.

- Four enums
1. BEGINNER
2. INTERMEDIATE
3. EXPERT
4. LEAD

- `currentlyUsed` should be an enum
- What should we call this second field? `WhenSkillUsed`

## ✅ Acceptance Criteria

- [ ] Fields added to schema
- [ ] Fields added to database
- [ ] Fields added to factory
- [ ] After seeding, applicant@test.com user has some UserSkills with these fields filled in, and some where they are null

vd1992:
🤖 Resolves #7380

## 👋 Introduction

Adds two new fields to the model. 

## 🕵️ Details

Migration, schema, and factory/seeding changes made. 

## 🧪 Testing

1. Migrate and seed successfully
2. Query for the applicant's seeded UserSkill with a query like below

```
query abc {
  user(id: "applicant@test.com-id") {
   userSkills {
    skill {
      name {
        en
      }
    }
    skillLevel
    whenSkillUsed
  }
  }
}
```


api/database/seeders/DatabaseSeeder.php
```diff
@@ -111,6 +111,23 @@ public function run()
         $pool = Pool::whereNotNull('published_at')->inRandomOrder()->first();
         $this->seedPoolCandidate($applicant, $pool);
         $this->seedAwardExperienceForPool($applicant, $digitalTalentPool);
+        $applicantUserSkills = DB::table('user_skills')
```
petertgiles, line 114: Is there a reason you're not using the ORM to get and update these models?
image

petertgiles (changes requested):
Looking great!  A small suggestion for the last AC.

api/database/seeders/DatabaseSeeder.php
vd1992: Facade seemed more straightforward to me in this case. I think this bit of seeding will probably be in flux, so I just rolled with what seemed simpler to me.

tristan-orourke: Nay. Sorry @vd1992 - for consistency, please use Eloquent models everywhere (except for in migrations).

vd1992: As an aside, I found it simpler, now I know why. 
I was originally trying to go `$applicant->userSkills()`, which obviously didn't work and gave up in favour of facade. 
I realized the issue looking at Peter's picture above. 

I just needed to not use the `()` 🤦 

Switched in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7440/commits/2c6b92625783fd3d2095d626d9ae06d2f18673c7

petertgiles: Looks great!  BTW, the screenshot is from [Tinker](https://laravel.com/docs/10.x/artisan#tinker), which I've started using more lately.

vd1992: Much to dig into 🤔 

image

PR #7447, opened 2023-07-31
[Feature] Snapshot test

### LINKED ISSUE #7323: 🧪 Test to ensure snapshot queries stay in sync
## 🧪 Tests

We currently have two queries that should stay in sync:
- api/app/GraphQL/Mutations/PoolCandidateSnapshot.graphql used to generate the snapshots
- apps/web/src/pages/Profile/ProfilePage/profileOperations.graphql used to query live data
 
Unfortunately they have drifted apart over time.  It's a bit weird to have a test that spans projects but I think it's worth it in this case.  The test will need to be a little clever since the two queries aren't totally identical - one is a `users` query and one is a `me` query.

## 🙋‍♀️ Proposed Solution

A PHPUnit test that loads up both files and compares them.  It might need to parse the JSON to do a proper comparison.

## :question: Questions
- Can PHPUnit read a file that's above the project root?  That might be tricky.

## ✅ Acceptance Criteria

- [ ] An automated test warns if the two files go out of sync
- [ ] Existing files put back in sync

petertgiles:
🤖 Resolves #7323 

## 👋 Introduction

This branch adds a PHPUnit test to ensure the that the query that makes the profile snapshots stays in sync with the regular query for displaying the profile.  It also syncs up the existing drift between the two queries by reordering fields and adding any missing between the two.  

I'm doubtful that _everything_ in the queries is necessary ~but I figured removing anything was out of scope for this PR~.  I needed to remove some extra user fields in [remove redundant user fields](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7447/commits/dca71623946f0366ebf2ba6cd40730a695817338) or the JSON would have a loop in it.  I'm not clear when those fields were added or why since it's older than the file history can retrieve.  They seem redundant since `User` is the top-level model of the query.  Nothing seems broken as far as I can see.

There was some gnarly type stuff needed to make Storybook happy.  I _think_ I fixed it correctly.  :crossed_fingers: 

## 🧪 Testing

Rebuild the app before beginning.

- [ ] Nothing removed from either query
- [ ] PHPUnit tests pass
- [ ] Applicant can submit an application which makes a new snapshot
- [ ] Admin can view the snapshot
- [ ] Nothing missing from the snapshot and it matches the user's profile

### DIFF
--- a/packages/fake-data/src/fakePoolCandidates.ts
+++ b/packages/fake-data/src/fakePoolCandidates.ts
@@ -10,9 +10,26 @@ import {
   Pool,
   User,
 } from "@gc-digital-talent/graphql";
-
+import {
+  GeneratedAwardExperience,
+  GeneratedCommunityExperience,
+  GeneratedEducationExperience,
+  GeneratedPersonalExperience,
+  GeneratedWorkExperience,
+} from "./fakeExperiences";
 import fakePools from "./fakePools";
 import fakeUsers from "./fakeUsers";
+
+export type GeneratedPoolCandidate = PoolCandidate & {
+  __typename: "PoolCandidate";
+  educationRequirementExperiences: Array<
+    | GeneratedAwardExperience
+    | GeneratedCommunityExperience
+    | GeneratedEducationExperience
+    | GeneratedPersonalExperience
+    | GeneratedWorkExperience
+  >;
+};
 
 const generatePoolCandidate = (pools: Pool[], users: User[]): PoolCandidate => {
   faker.setLocale("en");
--- a/packages/fake-data/src/fakePoolCandidates.ts
+++ b/packages/fake-data/src/fakePoolCandidates.ts
@@ -10,26 +10,8 @@ import {
   Pool,
   User,
 } from "@gc-digital-talent/graphql";
-import {
-  GeneratedAwardExperience,
-  GeneratedCommunityExperience,
-  GeneratedEducationExperience,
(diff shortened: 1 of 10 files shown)


packages/fake-data/src/fakePoolCandidates.ts
```diff
@@ -10,10 +10,27 @@ import {
...
+  __typename: "PoolCandidate";
+  educationRequirementExperiences: Array<
+    | GeneratedAwardExperience
+    | GeneratedCommunityExperience
+    | GeneratedEducationExperience
+    | GeneratedPersonalExperience
+    | GeneratedWorkExperience
+  >;
+};
+
```
esizer, line 33: I think this is the source of the dependency cycle. `fakeUsers.ts` imports this then `fakePoolCandidates` imports something from `fakeUsers.ts`

esizer (changes requested):
Looks great and is an excellent idea! Just some minor lint errors then we are good to :shipit:

packages/fake-data/src/fakePoolCandidates.ts
petertgiles: Thanks!  I missed this warning.  I just moved the type definition to its own file.
Ref: [move fake type out to own file](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7447/commits/2f65b30da36d599e100ec515337e842a2488b64a)

PR #7449, opened 2023-07-31
[Fix] Repoint pool table to root query

### LINKED ISSUE #6109: Repoint Admin Pool Table query at root query

esizer:
🤖 Resolves #6109 

## 👋 Introduction

Update the `PoolTable` to use the root pools query instead of querying through the users team roles.

### Question

All roles that can access the pools table have the `view-any-publishedPool` permission. This is a little odd since on the admin side, I feel like I would only be able to see pools I can edit. Is this an issue? Should we filter out non-team pools specifically on admin or is it fine if they see them and just can't edit them?

### Answer

We only want to show pools the user has access to for administration purposes. We are filtering the pools so that pool operators can only see those their team owns.

## 🧪 Testing

1. Build `npm run build`
2. Login as a platform admin
3. Add a pool on a specific team 
4. Log back in as a pool operator that is not on the team created in step 3
5. Navigate to the pools table
6. Confirm the pool created in step 3 does not exist


apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
```diff
@@ -23,11 +23,8 @@ import {
   Maybe,
   Pool,
   Scalars,
-  useGetMePoolsQuery,
-  RoleAssignment,
   LocalizedString,
   useAllPoolsQuery,
-  Team,
```
petertgiles, line 30: I think the team column is broken now.
image

petertgiles (changes requested):
Wow, so much cleaner!  One thing to check on, though.

apps/web/src/pages/Pools/IndexPoolPage/components/PoolTable.tsx
esizer: Oooo, good catch! 🙇‍♀️  Fixed in [query for team display name on pools table](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7449/commits/9e4fa79521cf734a8ac44e5d8b484fbf4660b95c)

petertgiles: Looks great!

petertgiles (approved):

PR #7462, opened 2023-08-01
[a11y] Replace select component on support form

### LINKED ISSUE #7355: ⌨ Support form subject component uses unreliable `MultiSelectFieldBase` component
## ⌨ Accessibility Issue

AT users can't reliably use the component `MultiSelectFieldBase` so it would be beneficial to avoid using it outside of admin. The only implementation outside of admin is in the support form.

REF: https://github.com/GCTC-NTGC/gc-digital-talent/pull/7314#discussion_r1270656926

## 🦋 Expected Behaviour

Field should be able to be used by AT users reliably.

## 📸 Screenshot

Screen Shot 2023-07-21 at 09 25 23

## 🙋‍♀️ Proposed Solution

Change `MultiSelectFieldBase` to `Select` component

https://github.com/GCTC-NTGC/gc-digital-talent/blob/d244993542fbe7aaef8d9d38ff6212345e02e8e7/apps/web/src/pages/SupportPage/components/SupportForm/SupportForm.tsx#L173-L213

## ✅ Acceptance Criteria

- [ ] support form subject field uses `Select` component

esizer:
🤖 Resolves #7355 

## 👋 Introduction

This replaces the component based on `react-select` with one based on the native `select` element on the support form.

## 🕵️ Details

The component based on `react-select` has a multitude of accessibility issues and does not provide much benefit here.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to `/support`
3. Confirm the select input works as intended
4. Confirm form still submits with proper information

## 📸 Screenshot
Screenshot 2023-08-01 100425

### DIFF
--- a/apps/web/src/pages/SupportPage/components/SupportForm/SupportForm.tsx
+++ b/apps/web/src/pages/SupportPage/components/SupportForm/SupportForm.tsx
@@ -3,13 +3,12 @@ import { FormProvider, SubmitHandler, useForm } from "react-hook-form";
 import { useIntl } from "react-intl";
 
 import { toast } from "@gc-digital-talent/toast";
+import { Input, Submit, TextArea, Select } from "@gc-digital-talent/forms";
 import {
-  Input,
-  Submit,
-  TextArea,
-  MultiSelectFieldBase,
-} from "@gc-digital-talent/forms";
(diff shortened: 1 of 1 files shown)


mnigh (approved):
👍

PR #7464, opened 2023-08-01
[Feature] Request table view [group+level] [stream] link

### LINKED ISSUE #7349: Request table changes for View Request link
## ✨ Feature

From UAT, we got the feedback that DCM team  don't like that View + Job Title for Action Link column 

## 🕵️ Details
We updated the links on the request table to use the submitted job title instead of just saying "view request". Unfortunately, these job titles are not informative to the request responders. Instead we want to change it to something that helps them understand what the request is.

Working with the recruitment team, we decided to use the classification and stream to generate this text.

Format: "View [ group-level] [stream] request"
Example: View IT-04 Software solutions request

Additionally since this repeats some info in the other columns (stream and classification) we can now hide those columns by default. We don't want to remove the columns completely though, because we still want to be able to sort by them if needed. But we can still hide Classification and stream columns by default on this table now.

### Current 
image

## ✅ Acceptance Criteria

- [ ] Column header changed back to "View request"
- [ ] View request link uses the text format "View [ classification-level] [stream] request" (FR: Voir la demande pour [group-level] [stream])
- [ ] "Stream" and "Group & Level" columns are hidden on the request table by default
-  [ ] Job title does not need to be a column at all

yonikid15:
🤖 Resolves #7349 

## 👋 Introduction

- Changes column header changed back to "View request"
- View request link uses the text format "View [ classification-level] [stream] request" (FR: Voir la demande pour [group-level] [stream])
- Hides the "Stream" and "Group & Level" columns on the request table by default
-  Removes the Job title column

## 🕵️ Details

- What should the link text format look like when there can be multiple classifications or multiple streams or both? Currently, a user can only select one on the Search page, so I set the link to display the first item in the array.

## 🧪 Testing

1. Login as admin@test.com
2. Go to Requests table`http://localhost:8000/en/admin/talent-requests` 
3. Ensure all criteria was met on the table view

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/components/SearchRequestTable/SearchRequestsTableApi.tsx
+++ b/apps/web/src/components/SearchRequestTable/SearchRequestsTableApi.tsx
@@ -3,7 +3,7 @@ import { IntlShape, useIntl } from "react-intl";
 import { SubmitHandler } from "react-hook-form";
 
 import { notEmpty } from "@gc-digital-talent/helpers";
-import { Pending, Spoiler } from "@gc-digital-talent/ui";
+import { Link, Pending, Spoiler } from "@gc-digital-talent/ui";
 import {
   InputMaybe,
   Maybe,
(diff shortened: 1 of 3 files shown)


apps/web/src/components/SearchRequestTable/SearchRequestsTableApi.tsx
```diff
@@ -255,18 +287,19 @@ const SearchRequestsTableApi = ({
...
+          defaultMessage: "View Request",
+          id: "kZGKwO",
           description:
-            "Title displayed for the search request table request job title column.",
+            "Title displayed for the search request table view column.",
         }),
-        id: "job_title",
-        sortColumnName: "job_title",
+        id: "view_request",
+        sortColumnName: "view_request",
```
esizer, line 296: I think we will need to remove this since it is a virtual column that we technically can't sort on through the API.

esizer (changes requested):
This is working well so far. Though, even if the ticket said to use this format for the link, it is troublesome. We should avoid links with the same name and a different "purpose" (i.e navigating to different pages).

Screenshot 2023-08-01 104911

Could we add some more unique value (either as part of an `aria-label` or sr-only text)? Maybe tacking on "from {fullName}"?

mnigh:
Change no longer requested by client.

PR #7470, opened 2023-08-02
[Fix] Change in hook order

### LINKED ISSUE #7397: 🐛 React has detected a change in the order of hooks called by `CareerTimelineSection`
## 🐛 Bug

React has detected a change in the order of hooks called by `CareerTimelineSection`.

## 🦋 Expected Behaviour

No console errors should appear.

## 📋 Steps to Reproduce

1. `npm run dev`
2. Navigate to **Profile > Career timeline and recruitment** page
3. Select **Sort experience by** field option
4. Select a different **Sort experience by** field option
5. Observe console error

## 📸 Screenshot

Screen Shot 2023-07-26 at 10 30 32

## ✅ Acceptance Criteria

- [ ] `console.error` no longer appears

esizer:
🤖 Resolves #7397 

## 👋 Introduction

Fixes an issue with sorting experiences calling hooks in the incorrect order.

## 🕵️ Details

This was caused by using a hook `useExperienceInfo` in a non-hook `sortAndFilterExperiences`. The hook was calling `useIntl` at different times because it was not being handled properly due to `useExperienceInfo` (which also contains `useIntl`) being used in a non-hook.

We were just deconstructing it for the title so it was replaced with `getExperienceName` which is what `useExperienceInfo` is using to get the title.

## 🧪 Testing

1. Build in dev mode `npm run dev`
2. Navigate to your career timeline (ensuring there are experiences there)
3. Sort the experiences in different ways
4. Confirm there are no errors in the javascript console
5. Confirm sorting still functions as expected

### DIFF
--- a/apps/web/src/components/ExperienceSortAndFilter/sortAndFilterUtil.ts
+++ b/apps/web/src/components/ExperienceSortAndFilter/sortAndFilterUtil.ts
@@ -1,7 +1,8 @@
 /* eslint-disable import/prefer-default-export */
+import { IntlShape } from "react-intl";
+
 import { Experience } from "@gc-digital-talent/graphql";
 import { notEmpty } from "@gc-digital-talent/helpers";
-import { ExperienceForDate } from "~/types/experience";
 import { PAST_DATE } from "@gc-digital-talent/date-helpers";
 
 import {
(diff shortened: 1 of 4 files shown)


vd1992 (approved):
Seeing no console error, sorting appears unchanged, and French still seems fine
💯

PR #7478, opened 2023-08-02
[Fix] Bad set state call in application career timeline

### LINKED ISSUE #7391: 🐛 Updating ApplicationCareerTimelinePage  - console warning
## 🐛 Bug

Console warning about updating `ApplicationCareerTimelinePage` and rendering a different component

## 📋 Steps to Reproduce

1. Run dev instead of build
2. Create an application and navigate to review career timeline stage
3. With console log open, click to edit an existing experience

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

A very similar issue here https://github.com/GCTC-NTGC/gc-digital-talent/issues/6146
And something else similar too https://github.com/GCTC-NTGC/gc-digital-talent/issues/4878

## ✅ Acceptance Criteria

- [ ] Console warning gone

esizer:
🤖 Resolves #7391 

## 👋 Introduction

Updates the application career timeline step to prefer cache when querying existing experiences.

## 🕵️ Details

We are calling a top level query on all these pages and queries in nested components seem to cause some issues when querying for similar data. We handle this by preferring cache (set in the parent components).

## 🧪 Testing

1. Build in dev mode `npm run dev`
2. Create and application
3. Continue to the career timeline step
4. "Edit" an experience
5. Confirm no console error appears
6. Complete creating the experience and return to the career timeline page
7. Confirm no console error appears
8. Confirm new experience appears

### DIFF
--- a/apps/e2e/cypress/e2e/talentsearch/submit-application-iap.cy.js
+++ b/apps/e2e/cypress/e2e/talentsearch/submit-application-iap.cy.js
@@ -15,6 +15,10 @@ import { aliasMutation, aliasQuery } from "../../support/graphql-test-utils";
 
 describe("Submit Application for IAP Workflow Tests", () => {
   beforeEach(() => {
+    cy.intercept("POST", "/graphql", function (req) {
+      aliasQuery(req, "getMyExperiences");
+    });
+
     cy.getSkills().then((allSkills) => {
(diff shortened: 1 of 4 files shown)


vd1992 (approved):
Can add, update, and delete experiences so it appears to work just right for me
Nice!

PR #7524, opened 2023-08-04
[Feature] ExperienceSkill pivots should soft-delete

### LINKED ISSUE #7363: ✨ Make ExperienceSkill soft-deletable
## ✨ Feature

ExperienceSkills, the pivots between experiences and UserSkills, should be soft-deleted instead of deleted by default. Connecting a skill/userSkill to an experience should restore the appropriate soft-deleted entry, if one exists.

## 🕵️ Details

UserSkills already use soft-delete, and the `syncSkills` and `connectSkills` methods of Experience will restore the appropriate UserSkill if a soft-deleted row exists, before linking to it.

Now we want the ExperienceSkill entry itself to also use soft-delete. The main goal is to preserve the details field on the pivot table in the following situation:

1. I link a skill to an experience, and write out details for it. This creates a UserSkill, and an ExperienceSkill which links to it.
2. I delete the skill from my library, which automatically (soft-)deletes any linked ExperienceSkills  (see #6942)
3. I re-add the skill to the experience. This restores the soft-deleted UserSkill _and_ ExperienceSkill.
4. I see the details text I previously wrote without having to re-write it.

The tricky thing here is that ExperienceSkill is implemented using [Laravel pivot table](https://laravel.com/docs/10.x/eloquent-relationships#defining-custom-intermediate-table-models) features, and...
> Pivot models may not use the SoftDeletes trait. If you need to soft delete pivot records consider converting your pivot model to an actual Eloquent model.

So if we want to use built-in soft-delete features we'll need to convert ExperienceSkill into an actual model, and make sure it doesn't affect the API. If that's too hard, we might be able to fudge it with some clever global scopes.

Besides that, this will mostly involve code changes to the `syncSkills`, `connectSkills` and `deleteSkills` methods of the Experience.php class.

## ✅ Acceptance Criteria

- [ ] ExperienceSkills deleted with the `deleteSkills` or `syncSkills` methods are soft-deleted
- [ ] When creating experienceSkills with `syncSkills` or `connectSkills`, if a soft-deleted experienceSkill already exists for the relevant UserSkill, restore it instead of creating a new record.
- [ ] The functionality of the API from an outside perspective is not affected at all.
- [ ] PHPUnit tests

tristan-orourke:
🤖 Resolves #7363.

## 👋 Introduction

This makes ExperienceSkills soft-deleted. That means that if you link an experience to a skill, add details, remove it, and re-link the skill, your previous details text will be restored.

## 🕵️ Details

This required some finagling to "simulate" soft-delete functionality, because [pivot models may not use the SoftDeletes trait](https://laravel.com/docs/10.x/eloquent-relationships#defining-custom-intermediate-table-models). In fact, I created an ExperienceSkill model which _can_ use the SoftDeletes trait and used that to make things a bit easier. But I can't use that model in the experience->skill relationships, because our API is currently designed around the `withPivot` functionality.

> Sidebar: Using withPivot is actually something I'd like to deprecate. Not only has it made this issue and #7218 **wayyyy** more complicated, I've also come to dislike how context-sensitive it makes the API. The value of something like `skill.experienceSkillRecord` for a given skill is entirely dependent on how you got to that skill. But that's an issue for another day. Especially since it's tied into our existing applicant snapshots. 😩

## 🧪 Testing

1. Create or go to an experience
2. Link that experience to a skill
3. Write details for how you used this skill
4. Remove the skill from the experience
5. Re-add the same skill
6. Check that the details you originally wrote are back.

### DIFF
--- a/api/app/Models/Experience.php
+++ b/api/app/Models/Experience.php
@@ -130,31 +130,6 @@ abstract class Experience extends Model
                 $this->userSkills()->attach($userSkillId, $detailsArray);
             }
         }
-
-        // ExperienceSkill::onlyTrashed()
-        //     ->where('experience_id', $this->id)
-        //     ->whereHas('userSkill', function ($query) use ($skillIds) {
-        //         $query->whereIn('skill_id', $skillIds);
-        //     })
-        //     ->restore();
-
-        // // I wanted to use syncWithoutDetaching, but it doesn't allow for updating pivot values like sync does.
-        // $skillsAlreadyAttachedToExperience = $this->userSkills()->pluck('skill_id');
-        // foreach($skills as $newSkill) {
-        //     $newSkill = collect($newSkill);
-        //     $userSkillId = $this->user->userSkills->firstWhere('skill_id', $newSkill->get('id'))->id;
-        //     $detailsArray = $newSkill->only('details')->toArray();
-        //     // If experienceSkill already exists
-        //     if ($skillsAlreadyAttachedToExperience->contains($newSkill->get('id'))) {
-        //         // And details is defined, then update details
-        //         if ($newSkill->has('details')) {
-        //             $this->userSkills()->updateExistingPivot($userSkillId, $detailsArray);
-        //         }
-        //     // Otherwise experienceSkill doesn't exist, so add it now
-        //     } else {
-        //         $this->userSkills()->attach($userSkillId, $detailsArray);
-        //     }
-        // }
         // If this experience instance continues to be used, ensure the in-memory instance is updated.
         $this->refresh();
     }
--- a/api/app/Models/Experience.php
+++ b/api/app/Models/Experience.php
@@ -96,7 +96,7 @@ abstract class Experience extends Model
         $skillIds = collect($skills)->pluck('id');
         $this->user->addSkills($skillIds);
 
-        $userSkills = UserSkill::where('user_id', $this->user_id); // Get this users UserSkills once, to avoid repeated db calls.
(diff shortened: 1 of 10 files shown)


esizer:
I'm not sure if I'm missing something here but, even though the soft-deletes work well, I still lose my details when attempting to "restore" on the frontend. 

Out frontend makes no API calls to retrieve the old data so when adding the skill back (restoring it), I get a fresh text area to type in and then it just updates the existing `ExperienceSkill` with new details.

Is this just prep work for the backend before we implement it on the frontend?

vd1992:
Running this strictly from the API, and it appears to work as expected for me.
Add -> remove -> add back and observe the details were not lost. So backend wise it appears to work 

https://recordit.co/tmrYhoewAo

api/app/Models/Experience.php
```diff
@@ -91,24 +96,65 @@ public function connectSkills($skills)
...
+        //     if ($skillsAlreadyAttachedToExperience->contains($newSkill->get('id'))) {
+        //         // And details is defined, then update details
+        //         if ($newSkill->has('details')) {
+        //             $this->userSkills()->updateExistingPivot($userSkillId, $detailsArray);
+        //         }
+        //     // Otherwise experienceSkill doesn't exist, so add it now
+        //     } else {
+        //         $this->userSkills()->attach($userSkillId, $detailsArray);
+        //     }
+        // }
```
vd1992, line 157: 🧹

tristan-orourke:
@esizer you're right, the testing instructions won't work because we don't have any UI where you add a skill without having a text box. Let's consider that out of scope for this issue. I'll create another ticket about allowing the frontend to retrieve old data so details can appear when you restore a skill.

PR #7532, opened 2023-08-05
[Fix] Mobile drawer navigation

### LINKED ISSUE #7107: 🐛 Admin menu on small screens functions poorly
## 🐛 Bug

The trigger to open the side menu on the admin side `/admin` shows at the very bottom of the screen and when toggled, it only opens it part of the way.

## 🦋 Expected Behaviour

Trigger should be at the top of the page and open the menu all the way.

## 📋 Steps to Reproduce

1. Navigate to `/admin`
2. Reduce your screen size
3. Note the button to open the menu is at the very bottom of the screen
4. Note that activating it only opens the menu partially

## 📸 Screenshot

Screenshot 2023-06-27 at 12 11 25 PM

## 🙋‍♀️ Proposed Solution

Implement the following mockup provided by @Jerryescandon 

image

## ✅ Acceptance Criteria

- [ ] Menu trigger appears under the header
- [ ] Opening the menu shows the entire menu and not just icons

esizer:
🤖 Resolves #7107 

## 👋 Introduction

This makes a number of improvements to the drawer (`SideMenu`) component.

- Moved the mobile trigger to be sticky below the header
- Added some animation to the open/close
- Added an overlay to the mobile menu (instead of push the content aside)
- Improved the focus management for keyboard users (return focus to trigger on close)

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to `/admin`
3. Open and close the menu and confirm it still works as expected
4. Reduce window size to narrower than 768px
5. Repeat step 3
6. Bonus, repeat 3 with only using keyboard and note the focus position change

## 📸 Screenshot
Screenshot 2023-08-05 082358
Screenshot 2023-08-05 082404


packages/ui/src/components/SideMenu/SideMenu.tsx
```diff
@@ -23,6 +24,7 @@ export interface SideMenuProps {
   header?: React.ReactNode;
   footer?: JSX.Element;
   children?: React.ReactNode;
+  trigger?: React.RefObject<HTMLButtonElement>;
```
vd1992, line 27: Could maybe do with a comment explaining what it is doing and what it references. Seems to read from the ref over in `AdminLayout` to control focus upon the overlay disappearing right?

vd1992 (approved):
Looks tricky but works great for me 🏆

PR #7533, opened 2023-08-08
[Debt] Replaces `getLocale` with `useLocale` hook

### LINKED ISSUE #3677: ♻️ Fully implement the useLocale hook throughout codebase

Re-ticketed from https://github.com/GCTC-NTGC/gc-digital-talent/pull/3249#discussion_r920089133

## Description
There was interest to try to use hooks more to encapsulate react logic that doesn't make sense to have in our vanilla helper functions as they exist. This is mostly because hooks are intended to allow _use of other hooks_, whereas our vanilla helpers instead accept args that are the _outputs_ of hooks. E.g...

```ts
// current way
const { intl } = useIntl();
const locale = getLocale(intl: IntlShape);

// more react-y way
const locale = useLocale();
// OR (more extensibly)
const { locale } = useLocale();
```

`useLocale()` was added in the above-linked PR, but only used within the scope of the PR. If we'd like to commit to the approach, we should fully migrate.

## Acceptance Criteria
- [ ] all usage of `getLocale` is migrated to `useLocale()`

mnigh:
🤖 Resolves #3677.

## 👋 Introduction

This PR replaces `getLocale` with `useLocale` hook.

## 🧪 Testing

1. Ensure no instances of `getLocale` exist in the codebase
2. Ensure app works as it did before


vd1992:
Tests are passing, it builds fine, no changes through Chromatic, hopping through the site and things appear unchanged. 
Appears to look good to me. 

Sure is a lot of files changed and you seemed hesitant about this being worth it in scrum, your thoughts now?

petertgiles:
Sorry, I'm still not seeing the value on this one.  :grimacing:   What is it we're trying to achieve with this?  We're not encapsulating reusable logic - the logic is already encapsulated in the helper function.  We're not abstracting away the formatjs library - we're still directly using the intl object all through our codebase.  We're not saving custom state or accessing a custom context in the hook.  We can't use the hook directly in a non-component function, which is the main uses of locale variables.  What does this solve?

petertgiles:
> We're not saving custom state or accessing a custom context in the hook

OK, we are accessing a custom context.   We're now accessing the state though this hook directly instead of through the intl hook (which has the state from this same hook).  Is that the advantage?

esizer:
> OK, we are accessing a custom context. We're now accessing the state though this hook directly instead of through the intl hook (which has the state from this same hook). Is that the advantage?

I don't believe there is any clear advantage except that the hook can also set the locale and consistency.

However, where it started getting a little ugly for me is when we pass in both `intl` and `locale`, which makes less sense. 

I'm wondering if we want to keep the non-hook version that simply reads locale and just refactor the functions that need a locale to using the `getLocale` function and `IntlShape` already passed in? 🤔

tristan-orourke:
Well... I kind of understand this, from a theoretical perspective. The Locale context is higher in the hierarchy than IntlProvider, so the Locale context can be considered the source of truth. 

But in practice, the `intl` object is always going to match the locale context anyway, and it grates to pass `locale` in as a separate value when it's already right there in `intl`.

If there's a component which needs to know the locale but has no use for the `intl` object otherwise, it might make sense to get it directly from context, instead of getting `intl` first and then passing it to another function. But let's keep the original `getLocale` function, and keep using it anywhere we already need the `intl` object.

Issue #5045, opened 2022-12-07
Improve database seeder performance


esizer:
## 👋 Introduction

Our `DatabaseSeeder.php` file is taking >1min to run right now and it is related the code following the normal `*Seeder.php` classes. We should look into moving this custom code into seeder classes and see about seeding more minimal data.

## ✅ Acceptance Criteria

- [ ] Improve seeder performance
- [ ] Reduce numbers of users seeded for basic dev seeder
- [ ] Reduce the amount of custom code in the main `DatabaseSeeder.php` file
- [ ] Hybrid approach: Some carefully created data and some random data

## 🕵️ Details

        > Wait, this seriously takes a minute for the seeder to run? 😨

Unfortunately, yes. The extra stuff we do at the end of the seeder has always been long, this just makes it longer. I wouldn't mind taking a look at it and seeing if we can do some optimizations.

## Update

It is specifically because we seed 150 users then for each user we:

- Generate 1-3 experiences for every category leading to creating 750-2250 experiences
- Create 1 pool candidate for each leading so seeding 150 pool candidates

I'm not sure we need that many users for dev, I bumped the number down to 20 and it only took 25s vs the 1min4s for the 150 users.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/d0b43cb2d82f2baa0ff4e78ccf5663100eacea68/api/database/seeders/DatabaseSeeder.php#L51-L168

_Originally posted by @esizer in https://github.com/GCTC-NTGC/gc-digital-talent/issues/5028#issuecomment-1340024387_

vd1992:
I'd say soft blocked by #4998 
as per the stand-up discussion, seeding has received multiple band-aids because of that test, ought to let that settle first

petertgiles:
One possible approach would be to figure out what we need to make each page look good and carefully make rows for each.  A bit more work to set up the seeder but you could get away far fewer entities and you could make them look more realistic.

esizer:
> One possible approach would be to figure out what we need to make each page look good and carefully make rows for each. A bit more work to set up the seeder but you could get away far fewer entities and you could make them look more realistic.

I was looking at picking this up but was having trouble knowing what we should actually be seeding. I think we may need to decide on this first. We would want to know what each page should look like and likely create some state methods for our factories before breaking up the `DatabaseSeeder` and make it more targeted rather that the shotgun method.

PR #7578, opened 2023-08-10
[Debt] Removes `expectedClassifications`, `expectedSalary`, `expectedGenericJobTitle`

### LINKED ISSUE #5347: ✨ Remove Roles and Salaries section of profile

mnigh:
🤖 Resolves #5347.

## 👋 Introduction

This PR removes the `expectedClassifications`, `expectedSalary`, and `expectedGenericJobTitle` from the application and deprecates the fields in the API.

## 🕵️ Details

We no longer use these anywhere in the application so they are now redundant.

> **Note**
> Some references were left for compatibility reasons.

## 🎩 Acknowledgements
👏 @esizer, @yonikid15 for past work used as a reference.

## 🧪 Testing

1. Run `refresh_api.sh`
2. Build frontend
    1. `npm run codegen`
    2. `npm run dev`
3. Navigate to `/admin/users`
    1. [ ] Confirm classifications are gone from filters
    2. [ ] Confirm role and salary does not exist on view users
    3. [ ] Print a user and confirm expected role and salary does not appear
    4. [ ] Export CSV and confirm expected role and salary does not appear
5. Navigate to `/users/:used-id/profile`
    1. [ ] Confirm role and salary section does not appear
    2. [ ] Confirm you can still complete your profile
6. Create an application
    1. [ ] Confirm role and salary section does not appear
    2. [ ] Confirm you can complete the application with no role and salary expectations
7. Ensure no references to `expectedClassifications`, `expectedSalary`, `expectedGenericJobTitle` (other than deprecated ones in graphql schema)


api/database/seeders/DatabaseSeeder.php
```diff
@@ -79,22 +74,12 @@ public function run()
 
                 // are they a government user?
                 if (rand(0, 1)) {
-                    // government users have a current classification and expected classifications but no salary
-                    $classification = Classification::inRandomOrder()->limit(1)->pluck('id')->toArray();
-                    $user->expectedClassifications()->sync($classification);
-                    $user->expected_salary = [];
+                    // government users have a current classification
```
esizer, line 77: Out of scope, but this part of the seeder looks odd. It seems as though we are just nulling the current classification for ~50% of users seeded. Wouldn't we also need to set the following:

```php
$user->is_gov_employee = false;
$user->department->null;
```

esizer (approved):
Looks great! 👏

api/database/seeders/DatabaseSeeder.php
mnigh: I was also confused by the logic within the "are they a government user?" conditional. I will ask about it in today's stand-up.

mnigh: See 511f7815dcbfcb3c9eeb5fbf43d22685f8bede1c

esizer: @petertgiles Do you have time to second check the state update? I paired up with Matt on this so we want to keep the review integrity in tact 😅

mnigh: @petertgiles do you mind reviewing this as well? 🙏

petertgiles: Looks nice!  Having the boolean parameter on the state function is a little unusual.  I haven't seen it before but I don't think it's wrong.  Would be nice to have a little comment above each of the two factory calls (L63 and L102) to explain at a glance why there are two different blocks.

I don't love having the big copy-paste afterCreating callback duplicated but I think that's a symptom of needing to refactor that and not a problem with the work in this PR.  Calling the factory twice with different chains shouldn't be a problem.

mnigh: @petertgiles is this what you were looking for in terms of the comments: e6eb1a8c42238910f94883f105ce1e237a8ac2a6

I agree with the duplicate `afterCreating` code not being great, but I think part of a larger issue: https://github.com/GCTC-NTGC/gc-digital-talent/issues/5045.

PR #7579, opened 2023-08-10
[Feature] Add UserSkill mutations

### LINKED ISSUE #6942: ✨ Add mutations for UserSkills
## ✨ Feature

We want to add mutations to create, edit, or delete UserSkill directly. This will mostly be straightforward, except for the fact that deleting a UserSkill has consequences for ExperienceSkillRecords.

## 🕵️ Details

A user can create a UserSkill with or without extra details (skill level and isCurrent[name pending]). The user must have the Policy ability to `update` the specified user. Validation should cause creation to fail if the User already has a record with the same Skill.

A user can update a UserSkill with a new details. It is not possible to change the user or skill id. The user must have the Policy ability to `update` the specified user.

A user can delete a UserSkill if they have the Policy ability to `update` the specified user.

Deleting a UserSkill also soft-deletes it from all experiences. 

If you add a skill to an experience, and it's already there but soft-deleted, then we restore the previous version, including justification text, and the matching UserSkill if necessary.

## ✅ Acceptance Criteria

- [ ] A user can create a UserSkill with or without nullable details. 
- [ ] To create a UserSkill The user must have the Policy ability to `update` the specified user. 
- [ ] Validation should cause creation to fail if the User already has a record with the same Skill (unless the existing skill is soft-deleted, in which case it is restored without error).
- [ ] A user can update a UserSkill with a new claimedLevel. It is not possible to change the user or skill id. 
- [ ] To update a UserSkill, he user must have the Policy ability to `update` the specified user.
- [ ] A user can delete a UserSkillRecord if they have the Policy ability to `update` the specified user.
- [ ] UserSkills are only soft-deleted.
- [ ] Deleting a UserSkill also soft-deletes that skill from all experiences. 
- [ ] Restoring a UserSkill does not automatically restore matching experiences.
- [ ] If you add a skill to an experience, and the ExperienceSkill is already there but soft-deleted, then we restore the previous version, including justification text, and the matching UserSkill if it is currently soft-deleted.
- [ ] PHPUnit tests created for soft deletion and restoring interactions
- [ ] PHPUnit feature tests for all of the mutations

vd1992:
🤖 Resolves #6942

## 👋 Introduction

Adds the three mutations: creating, updating, and deleting that act on `UserSkill` 

## 🕵️ Details

Mutations added and they also function with soft deletion.
Restoring models where applicable is present as well as delete/restore as needed on the `ExperienceSkill` model.
Unit testing added for all three mutation operations. 

## 🧪 Testing

1. Tests pass
2. Play with operations below and observe results

Creating 

```
mutation createUserSkill {
  createUserSkill(
    userId: ":user",
    skillId: ":skill",
    userSkill: {
      skillLevel: "",
      whenSkillUsed: ""
    }
  )
  {
    id
    user {
      id
    }
    skill {
      id
    }
    skillLevel
    whenSkillUsed
  }
}
```

Updating

```
mutation updateUserSkill {
  updateUserSkill(
    id: ":userSkill",	
    userSkill: {}
  ) {
    id
    user {
      id
    }
    skill {
      id
    }
    skillLevel
    whenSkillUsed
  }
}
```

Deleting 

```
mutation deleteUserSkill{
  deleteUserSkill(
    id:":userSkill"
) {
    id
  }
}
```


api/app/GraphQL/Mutations/DeleteUserSkill.php
```diff
@@ -0,0 +1,24 @@
...
+{
+    /**
+     * Soft delete a UserSkill model
+     * @param  null  $_
+     * @param  array{}  $args
+     * @return UserSkill
+     */
+    public function __invoke($_, array $args)
+    {
+        $existingModel = UserSkill::find($args['id']); // will never be null as policy check precedes this
```
tristan-orourke, line 18: > will never be null as policy check precedes this

That might be true right now, but what if someone calls this function elsewhere, forgetting to run the policy? Or changes the policy? Better to cover your back and handle edge cases here.

api/app/GraphQL/Mutations/CreateUserSkill.php
```diff
@@ -0,0 +1,42 @@
...
+                ExperienceSkill::where('user_skill_id', $existingModel->id)->restore(); // restore records on experience_skill
+                return $existingModel;
+            }
+            throw ValidationException::withMessages(["DuplicateUserSkill"]);
+        }
+
+        $createdModel = UserSkill::create([
+            'user_id' => $args['user_id'],
+            'skill_id' => $args['skill_id'],
+            'skill_level' => isset($args['skill_level']) ? $args['skill_level'] : null,
```
tristan-orourke, line 36: Is this necessary? Won't it default to null if it's not set?

api/app/GraphQL/Mutations/CreateUserSkill.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        // check if the model exists accounting for trashed and then restore or error out, if not found proceed to create
+        $existingModel = UserSkill::where('user_id', $args['user_id'])
+            ->where('skill_id', $args['skill_id'])
+            ->withTrashed()
+            ->first();
+        if ($existingModel !== null) {
+            if (($existingModel->deleted_at !== null)) {
+                $existingModel->restore();
+                ExperienceSkill::where('user_skill_id', $existingModel->id)->restore(); // restore records on experience_skill
```
tristan-orourke, line 27: This isn't necessary.
From the acceptance criteria: "Restoring a UserSkill does not automatically restore matching experiences."

tristan-orourke (changes requested):
Hey, good job with this, it looks like it works. However, I'd like to suggest an alternate approach. 

Take a look at [Laravel Eloquent events](https://laravel.com/docs/10.x/eloquent#events). Instead of writing a custom class for the mutations, we can let GraphQL treat them as standard Create and Delete operations, and trigger the code directly on the model lifecycle.

On the `creating` event, we can check if it already exists, and restore instead of creating (we cancel the create operation from return false in the event handler). In the `trashed` event, we also soft-delete the experience-skills.

api/app/GraphQL/Mutations/CreateUserSkill.php
vd1992: Okey, done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7579/commits/4ec7b77dbb28fed7af019891b237ce0e886fb690

api/app/GraphQL/Mutations/DeleteUserSkill.php
vd1992: Sure, I can do that 
https://github.com/GCTC-NTGC/gc-digital-talent/pull/7579/commits/e579e2fe9c3a60ac7ff588005e7036b2a39157f3

api/tests/Feature/UserSkillTest.php
```diff
@@ -33,4 +100,256 @@ public function testExperienceRelationshipSkipsSoftDeletedPivots(): void {
...
+                $this->createUserSkill,
+                [
+                    'userId' => $this->user->id,
+                    'skillId' => $skill->id,
+                    'userSkill' => null,
+                ]
+            )
+            ->assertGraphQLErrorMessage('This action is unauthorized.');
+
+        // assert user can successfully create with optional fields null
```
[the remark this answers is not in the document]
vd1992: Sure, I can split things up a little bit more 🔪 
Done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7579/commits/a81d13b7da58502dd141c297a218567e86608386

api/app/GraphQL/Mutations/CreateUserSkill.php
vd1992: I had errors popping up while writing testing that this change in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7579/commits/1a055ddf7f5c6a7fdef767a3b06622695e29e5b8 seemed to resolve
So, maybe, maybe not? 🤷

vd1992:
> Hey, good job with this, it looks like it works. However, I'd like to suggest an alternate approach.
> 
> Take a look at [Laravel Eloquent events](https://laravel.com/docs/10.x/eloquent#events). Instead of writing a custom class for the mutations, we can let GraphQL treat them as standard Create and Delete operations, and trigger the code directly on the model lifecycle.
> 
> On the `creating` event, we can check if it already exists, and restore instead of creating (we cancel the create operation from return false in the event handler). In the `trashed` event, we also soft-delete the experience-skills.

I spent some time and dug into it.
Deleting seemed pretty straightforward to implement and did so in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7579/commits/53ce90023000c9951d9a4564f47a7bf81ac986f1 and https://github.com/GCTC-NTGC/gc-digital-talent/pull/7579/commits/035065124b23cc9f6511efd0071c3aeb6973c782

However, the creation event was trickier and I ran into a wall. Returning false and cancelling doesn't work as the create operation expects something to be returned. So, I stuck to the custom class as in it I can conditionally return the restored model, which satisfies what the API is expecting. Something being returned is guaranteed and it can be either a new model or restored one.

api/app/Models/UserSkill.php
```diff
@@ -23,6 +23,26 @@ class UserSkill extends Model
...
+
+    /**
+     * model lifecycle methods
+     */
+    protected static function booted(): void
+    {
+        static::deleting(
+            function (UserSkill $userSkill) {
+                // soft delete all experience_skill records containing the model
+                ExperienceSkill::where('user_skill_id', $userSkill->attributes['id'])->delete();
```
tristan-orourke, line 41: When working with a laravel Eloquent model, it's conventional to access attributes directly, instead of through the attributes array.
```suggestion
                ExperienceSkill::where('user_skill_id', $userSkill->id)->delete();
```

tristan-orourke (approved):
Okay! All my comments are addressed. I see what you mean about the create event - you can cancel the create event and restore the existing model, but then the create method won't return the right object. The custom mutation works well.

I've made one style suggestion but I think this is ready to go!

api/app/Models/UserSkill.php
vd1992: Thanks, suggestion added in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7579/commits/4eefd5b3f5abcec389bd243de4e19295101a1be3
I can't remember where calling `->attributes` popped in first, I think from something I was playing around with 🤔

PR #7588, opened 2023-08-11
[Feature] Update table component

### LINKED ISSUE #6932: ⚙️ Enhance the table component
-- screenshot

### Focus and hover

esizer:
🤖 Resolves #6932 

## 👋 Introduction

Creates a new `Table` component that merges the initial two into a single component and makes it mobile responsive.

> **Warning**
> Right now `e2e` tests are failing because they were not updated to support the new tables. Once a tentative approval is given, those tables will be reverted and `e2e` tests should pass before merging.

### 🏷️ Names

I'm not super happy with the naming of API items. I don't feel like they are clear and describe what they do well. Any suggestions on names would be much appreciated 🙇‍♀️ 

### API

The API was designed to allow us to progressively adopt external control over the table. In order to allow `react-table` to manage state for different parts, we can use the features `internal: boolean` prop. Setting that to `true` will let `react-table` to control that portion of the state. If we set that to `false` then we must handle that portion externally (usually on the server) and pass the modified data into the table.

```tsx
interface TableProps<TData> {
  /** Accessible name for the table */
  caption: React.ReactNode;
  /** Data to be displayed within the table */
  data: TData[];
  /** Column definitions for `react-table` */
  columns: ColumnDef<TData>[];
  /** Column definitions for `react-table` */
  hiddenColumnIds: string[];
  /** Determine if any aspect of the table is loading (server side) */
  isLoading?: boolean;
  /** Override default null message with a custom one */
  nullMessage?: NullMessageProps;
  /** Enable row selection */
  rowSelect?: RowSelectDef<TData>;
  /** Enable the search form */
  search?: SearchDef<TData>;
  /** Enable sorting */
  sort?: SortDef;
  /** Enable pagination */
  pagination?: PaginationDef;
  /** Enable printing selected rows (requires rowSelect) */
  print?: DatasetPrint;
  /** Enable downloading selected rows and/or all data (requires rowSelect) */
  download?: DatasetDownload;
  /** Enable the "add item" button */
  add?: AddLinkProps;
  /** Filter component */
  filterComponent?: React.ReactNode;
}

interface NullMessageProps {
  /** Heading for the message */
  title?: React.ReactNode;
  /** Main body of the message */
  description?: React.ReactNode;
}

type RowSelectDef<T> = {
  /** Label for the "select all" checkbox in the header */
  allLabel?: string;
  /** Render method for the table cell (`td`) */
  cell: ({ row }: CellContext<T, unknown>) => JSX.Element;
  /** Callback method for when a row is (de)selected */
  onRowSelection?: (rows: T[]) => void;
};

type SearchDef<T> = {
  /** Allows the table to manage search */
  internal: boolean;
  initialState?: SearchState;
  onChange?: (newState: SearchState) => void;
} & Omit<SearchDefFormProps<T>, "onChange">;

type SortDef = {
  /** Allows the table to manage search */
  internal: boolean;
  initialState?: SortingState;
  /** Callback when sorting rule changes */
  onSortChange?: (sortState: SortingState) => void;
};

type PaginationDef = {
  /** Allows the table to manage search */
  internal: boolean;
  /** Callback for when the pagination changes */
  onPaginationChange?: (newPagination: PaginationState) => void;
  /** Initial pagination state */
  initialState?: PaginationState;
  /** Total number of pages */
  total?: number;
  /** Available page sizes */
  pageSizes?: number[];
};

type DatasetPrint = {
  onPrint: ButtonClickEvent;
  label?: React.ReactNode;
};

type DatasetDownloadItem = {
  /** The props required for CSV Download */
  csv: Csv;
  /** The label to display in the button */
  label?: React.ReactNode;
};

type DatasetDownload = {
  /** Props for the download button when items are selected */
  selection: DatasetDownloadItem;
  /** Props for the download button for all items */
  all?: DatasetDownloadItem;
};
```

#### Meta API

On the `ColumnDef` we have access to a generic `meta` field for a column. We are using this to control specific aspects about the table display (mostly for mobile responsiveness).

```
  interface ColumnMeta<TData extends RowData, TValue> {
    /** Mark this column as the row title (should be first column and only one) */
    isRowTitle?: boolean;
    /** Mark this column as the row selection column (do not use directly)  */
    isRowSelect?: boolean;
    /** Override the header for the mobile view */
    mobileHeader?: React.ReactNode;
    /** Hides the header for this column on the mobile */
    hideMobileHeader?: boolean;
    /** Header for the search column dropdown */
    searchHeader?: React.ReactNode;
  }
```

## 🔧 Usage

This does it's best to use the column definition to control specifics about the table itself. The new table has a new way to define columns. I recommend reading about [the new columns definitions](https://tanstack.com/table/v8/docs/guide/column-defs#column-def-types) if you would like to learn more.

There are certain features you can toggle on and off for each column using the column definition. These values are `true` by default unless otherwise noted.

- [`endableHiding`](https://tanstack.com/table/v8/docs/api/features/column-visibility#column-def-options) - Setting this to `false` will prevent users from hiding a column and not show it in the column visibility dialog.
- [`enableColumnFilter`](https://tanstack.com/table/v8/docs/api/features/filters#enablecolumnfilter) - This controls whether the column is available in the dropdown near the search form to search specifically on this column.
- [`enableGlobalFitler`](https://tanstack.com/table/v8/docs/api/features/filters#enableglobalfilter) - Includes this column in the general search (when no column is selected in the dropdown)
- [`enableSorting`](https://tanstack.com/table/v8/docs/api/features/sorting#enablesorting) - Determines if the column can be sorted

### Client Driven

> **Note**
> There is a temporary example of this in the `SkillTable`.

When defining different features for a client driven table, make sure to set `internal: true`. This allows `react-table` to take over and manage the data.

### API Driven

> **Note**
> There is a temporary example of this in the `ResponsiveUsersTable`.

These tables are a little more involved as you will see. The main thing is to set `interal: false` so you can control the data. Then, use the callback handlers for different features to sync state for the API call. Like the existing API table, there is a helper to convert internal sorting state to an `orderBy` clause that can safely be used with `graphql`.

#### `sortingStateToOrderByClause`

This is a simple function that takes the existing sorting state and an optional map and returns an `orderBy` clause for `graphql`.

The map being passed is for when the column ID does not match the column ID required by `graphql`. This could not really be needed since you can define a custom ID in the column definition and simply use that instead.

```tsx
const columns = [
  column.accessor("name", {
    id: "firstName",
  },
  column.accessor("createdAt",
    id: "createdAt",
  }
]

const sortColumnMap = new Map<string, string>([
  ["createdAt", "created_at"]
]);
```

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to `/admin/users`
3. Confirm the table functions as expected and matches the mockup
4. Navigate to `/admin/skills`
5. Confirm the table functions as expected and matches the mockup

## 📸 Screenshot

Screenshot 2023-08-22 090919
Screenshot 2023-08-22 091007


apps/web/src/components/Table/ResponsiveTable/ColumnDialog.tsx
```diff
@@ -0,0 +1,94 @@
...
+                <input
+                  ref={allColumnsRef}
+                  {...{
+                    type: "checkbox",
+                    checked: table.getIsAllColumnsVisible(),
+                    onChange: table.getToggleAllColumnsVisibilityHandler(),
+                  }}
+                />{" "}
+                {intl.formatMessage({
+                  defaultMessage: "Toggle All",
```
mnigh, line 63: Sentence case?

vd1992: So keen 😆

apps/web/src/components/Pagination/usePagination.ts
```diff
@@ -20,91 +35,59 @@ const range = (start: number, end: number): number[] => {
...
+  totalPages,
+  siblings = 1,
+  boundaries = 1,
+}) => {
+  const total = Math.max(Math.trunc(totalPages ?? 0), 0);
 
-    // Pages count is determined as siblingCount + firstPage + lastPage + currentPage + 2*DOTS
-    const totalPageNumbers = siblingCount + 5;
+  const pageRange = useMemo((): UsePaginationReturn => {
+    const totalPageNumbers = siblings * 2 + 3 + boundaries * 2;
```
tristan-orourke, line 47: Where does the math come from on this?

apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
```diff
@@ -0,0 +1,227 @@
...
+  download,
+  print,
+  add,
+  pagination,
+}: TableProps<TData>) => {
+  const id = React.useId();
+  const intl = useIntl();
+  const [, setSearchParams] = useSearchParams();
+  const isInternalSearch = search && search.internal;
+  const memoizedData = React.useMemo(() => data, [data]);
```
tristan-orourke, line 83: Is using `useMemo` like this any different from using data directly?

esizer: Huh, for some reason I thought the data needed to be memoized for `react-table` but I can't find anything to confirm that so maybe we can just pass it in directly! 🤔 I think they also removed the requirement for columns to memoized as well 🤷‍♀️

apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
```diff
@@ -0,0 +1,227 @@
...
+          return value ? memoizedData[index] : undefined;
+        })
+        .filter(notEmpty);
+
+      rowSelect.onRowSelection(selectedRows);
+    }
+    // Note: This is probably due to mis-use of useEffect but
+    // adding exhaustive deps causes this to run infinitely
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [rowSelection]);
```
tristan-orourke, line 115: I think you can avoid using `useEffect` here by running this code right when `setRowSelection` is called. Maybe by wrapping `setRowSelection`, like this:
```
const mySetRowSelection = (newRowSelectionState) => {
  // call the inner setRowSelection hook method. 
  setRowSelection(newRowSelectionState);
  // other stuff to do
  if (rowSelect?.onRowSelection) {
      const selectedRows = Object.values(newRowSelectionState)
        .map((value, index) => {
          return value ? memoizedData[index] : undefined; 
        })
        .filter(notEmpty);

      rowSelect.onRowSelection(selectedRows);
    }
}
```
Now, if the hook has giving you trouble, something must have an unstable value (memoizedData or rowSelect?) and that might still be a problem, I'm not sure. But avoiding useEffect is recommended where possible.

tristan-orourke:
At a preliminary view, this looks pretty good to me. Seems like it'll powerful and flexible. A bit heavy, but we have a lot we want it to do! 😂

apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
esizer: Sweet, that seems to have worked! Turned out to be a little more complicated so I moved it out to it's own hook as well.

[refactor row selection state to avoid useEffect](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7588/commits/f9a070fe195bbf2744e5903b2a75f158fc1b082e)

petertgiles (commented):
Few quick comments.

apps/web/src/components/Pagination/Pagination.tsx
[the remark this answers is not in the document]
esizer: Thanks for catching that! Fixed in [reduce font-weight for non-active pages](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7588/commits/9a5d26c4ea2cd488d5af10cbcc4ee40c9874e058)

apps/web/src/components/Table/ResponsiveTable/useControlledTableState.ts
```diff
@@ -0,0 +1,214 @@
...
+      if (callback) callback(newValue);
+      return newValue;
+    });
+  } else {
+    setter(updater);
+    if (callback) callback(updater);
+  }
+};
+
+const useTableSateFromSearchParams = (
```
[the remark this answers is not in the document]
esizer: 😅 Fixed in [fix typo](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7588/commits/d954a85ebbe37f9a3112fb27cc967f50e6fca984)

apps/web/src/components/Pagination/Pagination.tsx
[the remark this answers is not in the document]
vd1992: On other tables, there is a bit of a contrast issue now

image

apps/web/src/components/Table/ResponsiveTable/useRowSelection.ts
```diff
@@ -0,0 +1,52 @@
...
+  setter: OnChangeFn<RowSelectionState>,
+];
+
+const useRowSelection = <T>(
+  data: T[],
+  rowSelect?: RowSelectDef<T>,
+): UseRowSelectionReturn => {
+  const [rowSelection, setRowSelection] = React.useState<RowSelectionState>({});
+
+  const rowSelectionCallback = (newRowSelection: RowSelectionState) => {
```
tristan-orourke: Maybe the definitions of `rowSelectionCallback`, `handleRowSelection` and `setter` should use `useCallback`, to ensure the setter function returned by `useRowSelection` is as stable as possible?

esizer: Heh, somehow you were viewing an old file. This was already done 😅 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/5f38b168475ee5ad4c7611cb4f3d9ecb08047ae7/apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx#L333-L334

apps/web/src/components/Pagination/Pagination.tsx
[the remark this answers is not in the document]
vd1992: I'm not sure that did it

image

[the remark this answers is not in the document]
esizer: Huh, it should be showing up as yellow 😖 Where are you looking at this?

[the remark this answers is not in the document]
vd1992: `/en/admin/talent-requests`
same with other non-responsive tables like pools

image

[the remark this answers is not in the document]
esizer: I'm so confused :/ It was "quaternary" which is supposed to be yellow but showing as purple in your first screen shot. The update was still supposed top be yellow but showing up as blue. 😕 

I reverted the change because it was fine before. Can you try a `npm run build:fresh`?

https://webaim.org/resources/contrastchecker/?fcolor=F8A812&bcolor=1E1E23

vd1992, line 1: Is the paginator supposed to update the X of Y bit when flipping through pages? So the first 10 results, or 11-20 results, or adjusting to 4 if there is only 4 on the page. Or am I just misinterpreting. 

image

image

image

esizer: That is a good question. There was no specific AC for that line so I went with the page size.

@substrae what is that first number meant to represent?

vd1992, line 1: Results for pagination update in sync with results for responsive API

image

But not quite for the other one. I guess that makes sense, the first only has 1 result to show while this fetched all and proceeded to hide all but 1 

image

esizer: hmm, I'm actually not sure what one is the preferred option. Should the second number show all _available_ or only the _number of items we found_ after all filters have been applied @substrae ?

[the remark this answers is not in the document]
vd1992: Ah, sorry for the confusion, my browser highlighted the number and I didn't quite pay attention to it doing that. 
This is the above, but the number not highlighted 

image

apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx
petertgiles, line 1: This probably has nothing to do with your PR and you can ignore it.  :laughing:  It makes me crazy that this table makes a search request when there is nothing selected.
image

apps/web/src/components/Table/ResponsiveTable/SearchForm.tsx
petertgiles, line 1: This looks a little awkward in a narrow viewport.  I'm not sure how it's supposed to look though.
image
image

petertgiles (changes requested):
Wow, this is so impressive!  :exploding_head: I've noted a few small things but overall, this is :fire: .  The reorganization of state management is _tight_.

apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx
esizer: Oh! This is obviously just an example but we might be able to prevent that with the  [`urql` pause option](https://formidable.com/open-source/urql/docs/basics/react-preact/#pausing-usequery) 👀

apps/web/src/components/Table/ResponsiveTable/constants.ts
```diff
@@ -0,0 +1,27 @@
...
+  searchState: {
+    term: "",
+    type: "",
+  },
+};
+
+export const SEARCH_PARAM_KEY = {
+  SEARCH_TERM: "st",
+  SEARCH_COLUMN: "sc",
+  HIDDEN_COLUMNS: "cv",
```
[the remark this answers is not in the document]
petertgiles: No, not really.  A little comment on that key might satisfy curiosity.

apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx
petertgiles: Oh, of course!  If it bothers me enough I'll make an issue.  :laughing:

esizer: I think we should make an issue regardless 😛 I never noticed that until you pointed it out. Or we could make it part of the AC for the implementation ticket? 🤔 Can't imagine it would be anything more than:

```tsx
const results = usePaginatedQuery({
  pause: !rowsSelected.length
})
```

petertgiles: #7681 :wink:

apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
```diff
@@ -0,0 +1,335 @@
...
+    .map((column) => ({
+      label: getColumnHeader(column, "searchHeader"),
+      value: column.id,
+    }));
+
+  const sortingState = table.getState().sorting;
+  const columnFilterState = table.getState().columnFilters;
+  const globalFilterState = table.getState().globalFilter;
+  const columnVisibilityState = table.getState().columnVisibility;
+  const paginationState = table.getState().pagination;
```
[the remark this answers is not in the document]
esizer: I agree! I think this was a similar oddity I was doing as a way to prevent infinite re-renders. Seems like I was dead wrong.

Updated in [update state to use deconstruction](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7588/commits/ee07e87336ddab0aa3df8b407671dec041fa337f)

apps/web/src/components/Table/ResponsiveTable/constants.ts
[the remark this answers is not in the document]
esizer: Makes sense! Added in [note to column visibility search param key](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7588/commits/63dbf4f2c3882666fd29b685611f9027f188b9ba)

apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx
esizer: Thanks! Also, seems to work 👍 [pause select query when no rows selected](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7588/commits/22a7fe1f4f7b0ffbc73659c1417b617a763be21e)

Screenshot 2023-08-22 162835

apps/web/src/components/Pagination/Pagination.tsx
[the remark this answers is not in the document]
esizer: Okay, I finally figured it out! It is fine on black but that is not black! 😛 That is a slightly transparent black.

I updated the background colour and now we get a ratio of 5.1:1 and pass! 👏 

[fix contrast issues with new table pagination](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7588/commits/de087db46056e5f7f17da5cadf0a29a0c7adb5fe)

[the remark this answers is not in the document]
vd1992: Works for me too! 

image

apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
vd1992, line 1: I'm not 100% sure if this is an issue with the new table, or just an issue with the temporary responsive users table. 
Changing the filter state for `where` should reset to page 1. If it is the latter, then you can ignore this. 

https://recordit.co/e6b6Jms506

apps/web/src/components/Table/ResponsiveTable/SearchForm.tsx
petertgiles: The narrow version looks good now but the wide version still needs some love.  The create button is taller than the others.  Seem like unnecessary word-wrap, too.
image

esizer: Oops! Should be fixed after [fix add button alignment on larger screens](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7588/commits/c52c352b84827e8efbbb22752e89b286bf6f9746)

petertgiles: Looks great!

petertgiles (approved):
Looks great!  I know there are a few small things outstanding but this is really solid.  As discussed in scrum, I think this is good enough to merge and move forward with the skills showcase.  Please spinoff any pending todos to a subsequent issue.

apps/web/src/components/Pagination/Pagination.tsx
gobyrne: I think it's only after filters. @substrae ?

brindasasi: agree with @gobyrne 
when filters applied, second no. should represent the total count of filtered result

vd1992: It was answered here, and matches that 
https://github.com/GCTC-NTGC/gc-digital-talent/issues/7752#issuecomment-1697695961

Issue #7082, opened 2023-06-22
✨ Proof of concept asynchronous operation


petertgiles:
## ✨ Feature

We have a need to be able to handle asynchronous operations.  When a user needs to complete an operation that will take longer than the Nginx response time limit we need to be able to queue the operation and then allow the user to interact with the result afterwards.

## 🕵️ Details

The current use case is to request a big query and download the bulk result after but there may be others.

## 🙋‍♀️ Proposed Implementation

Some discussed ideas:
- Run the operation on a timer, like overnight, and then save the results to a file to be downloaded later
- Run the operation asynchronously then email the results to the user
- Queue the operation then use something like websockets to notify the client when the result is ready
- [Laravel Notifications](https://laravel.com/docs/10.x/notifications) is being set up in #6927 and has the concept of a [worker queue](https://laravel.com/docs/10.x/queues) and [event broadcasts](https://laravel.com/docs/10.x/notifications#broadcast-notifications)
- Graphql and Lighthouse has subscriptions. [read more](https://lighthouse-php.com/5/subscriptions/getting-started.html#setup)
- Possibly in Azure: https://learn.microsoft.com/en-us/answers/questions/1261058/pusher-com-service-equivalent-in-azure

## ✅ Acceptance Criteria

- [ ] A proof of concept to demonstrate the solution
- [ ] The solution is well suited to generating bulk data reports
- [ ] The solution is general enough to be used for other things

vd1992:
Might be unneeded due to Power BI

PR #7593, opened 2023-08-11
[Feature] Admin soft delete user

### LINKED ISSUE #6353: ✨ (Soft-)Delete user (Admin)
## ✨ Feature

This is Admin side. a follow up issue should be created on the User side. 

We need a way to cleanup the platform from garbage data created during migration. Ideally we can HARD delete. But many of these users are tied to pools and stuff. Probably best to figure out a soft delete first.

## 🕵️ Details

Deleting users should prevent that user to showing up EVERYWHERE. A deleted user should no longer show up in pools, requests, user table, etc. 

Ideally we could hard delete, so that another account can be created with the same email etc.

This should in turn, make the status "REMOVED" obsolete, or make it so that it can only be set through this UI (Admin only). For example, Pool operators and Request responders should not be able to remove users, or un-remove them either. Which means removed users should not show up for anyone other than admins.

## 📜  Policy considerations

In addition, we need to "Keep the data for 5 ~months~ years" according to GoC policy.

We should be able to recover a user within those 5 months.

We should destroy any data related to a user after 5 months. Including pools they qualified, data on their profile, etc.

## 🎨 Design File

[Link to the figma](https://www.figma.com/file/UsrB15yicgDCLDQmx3snyl/User-Flows?node-id=3031%3A4790&t=jMYEJgH0g8DMt9zk-1) 

## 📸 Screenshot

image

image

image

- Use Laravel soft deletes on User object
- Soft delete cascaded objects like pool candidate, experience, as well

## ✅ Acceptance Criteria

- [ ] Admin can soft delete user
- [ ] Admin can see soft-deleted users to edit them
- [ ] Soft-deleted users emails are modified by adding `+deleted-at-<date>` to the email name
- [ ] A user cannot log in if they have been soft deleted
  - [ ] User should get an error message

JamesHuf:
🤖 Resolves #6353

## 👋 Introduction

Allows admins to soft delete users.

## 🕵️ Details

Also adds an scheduled task which hard-deletes users soft-deleted for more than 5 years.

### Front end changes
1. Sign in as an admin.
2. Navigate to the users table and select to edit a user.
3. Use the new section at the bottom of the page to (soft) delete the user.
4. Return to the users table and confirm you can use the new filter to view the deleted user.
5. Sign in as the deleted user and confirm you are signed out and redirected to a page explaining the user has been deleted.

### Scheduled hard-delete
1. Manually change a users `deleted_at` to be more than 5 years ago.
2. Change "daily()" in line 55 of `api/app/Console.Kernel` to "everyMinute()".
3. Run `php artisan schedule:work` to run the scheduler locally.
4. Wait a minute for the event to be triggered.
5. Confirm the user and all associated data (poolCandidates, experiences, etc.) have been hard deleted.

## 📸 Screenshot
image
image


api/app/Console/Kernel.php
```diff
@@ -15,7 +19,40 @@ class Kernel extends ConsoleKernel
...
+                foreach ($user->roleAssignments()->get() as $role) {
+                    $role->forceDelete();
+                }
+                foreach ($user->userSkills()->withTrashed()->get() as $skill) {
+                    $skill->forceDelete();
+                }
+
+                $user->forceDelete();
+            }
+        })->daily();
```
vd1992, line 55: Not critical to the issue at hand, but I think scheduled functions should have their time set carefully and specifically. 
Ideally run at times of low server load, this execution should be fairly inexpensive as though it runs daily it would likely have zero people returned almost always and for sure for 5 years. But planning for possible further daily tasks like the other idea we floated with daily reporting #7082 

Laravel midnight would be 00:00 UTC or 20:00 EST or 17:00 PST so possibly executing these during possibly busy times. 
So I think we should establish a format for scheduled tasks to make it convenient to slot more in. 

Maybe `->dailyAt('08:00');` (so overnight in Canada)
And utilize `withoutOverlapping();` when adding further tasks

image

api/app/Models/User.php
```diff
@@ -251,6 +254,28 @@ public static function scopeIsProfileComplete(Builder $query, ?bool $isProfileCo
...
+        static::restoring(function (User $user) {
+            // Cascade restore to child models
+            foreach ($user->poolCandidates()->withTrashed()->get() as $candidate) {
+                $candidate->restore();
+            }
+
+            // Remove the deleted-at text from the end of the email
+            $newEmail = substr($user->email, 0, -22);
+            $user->update(['email' => $newEmail]);
+        });
```
vd1992, line 278: What if you try and restore an account and the email is being used by another account, say
Account made by user with email X. 
Account deleted for whatever reason, accidental or otherwise. Set to X-deleted. 
User tries to get account restored, in the meanwhile makes an account with freed email X. 
Admin cannot restore the account anymore as the -deleted can't be removed without violating uniqueness constraint

api/app/Console/Kernel.php
```diff
@@ -15,7 +19,40 @@ class Kernel extends ConsoleKernel
      */
     protected function schedule(Schedule $schedule)
     {
-        // $schedule->command('inspire')->hourly();
+        $schedule->call(function () {
+            $deleteDate = Carbon::now()->subYears(5);
+            $users = User::onlyTrashed()->whereDate('deleted_at', '<=', $deleteDate)->get();
+
+            foreach ($users as $user) {
+                // Cascade delete to child models
```
petertgiles, line 27: This is an interesting approach.  I think in other places we've tended to use database delete cascades instead of manually removing all the children first.  Is there a reason you went with this approach?

apps/web/src/components/Router.tsx
```diff
@@ -119,6 +120,14 @@ const SignedOutPage = React.lazy(() =>
       ),
   ),
 );
+const UserDeletedPage = React.lazy(() =>
+  lazyRetry(
+    () =>
+      import(
+        /* webpackChunkName: "tsSignedOutPage" */ "../pages/Auth/UserDeletedPage/UserDeletedPage"
```
petertgiles, line 127: Don't forget to update the chunk name.

petertgiles (commented):
This is looking cool! :medal_sports:  I did a quick code read-through and left a few comments.  Please rerequest a review when you're ready for a closer look.

api/app/Models/User.php
petertgiles: Agreed, the user can restore their own email after the account is restored.

api/app/Console/Kernel.php
petertgiles: Also, how do we find out if something goes wrong during a scheduled task?

JamesHuf: I've changed it to `->dailyAt('08:00');`. I don't think we need `withoutOverlapping` though since it looks like it prevents a single task from having multiple running instances at once which should be unnecessary in this case.

I'm not certain about server side but at least locally errors during a scheduled task show up in `storage/logs/laravel.log`. If that doesn't work on the server we could probably utilize [this](https://laravel.com/docs/10.x/scheduling#task-output).

api/app/Models/User.php
JamesHuf: I've updated this to specifically error out if the users email would be duplicated. This isn't actually used on the frontend currently though.

api/app/Console/Kernel.php
JamesHuf: No I didn't have any good reason for this, I've updated to use automatic database cascades with [ed961ae](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7593/commits/ed961ae4e258c556bf5a2b2bf7842fc5e127caee).

apps/web/src/components/Router.tsx
JamesHuf: Oops, this should be fixed now.

api/app/Console/Kernel.php
```diff
@@ -15,7 +19,40 @@ class Kernel extends ConsoleKernel
      */
     protected function schedule(Schedule $schedule)
     {
-        // $schedule->command('inspire')->hourly();
+        $schedule->call(function () {
```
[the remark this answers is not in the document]
JamesHuf: I've changed this into a Command and created a couple basic tests for it.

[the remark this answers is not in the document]
petertgiles: Oh yeah, I like this a lot better!

api/app/Console/Commands/HardDeleteOldUsers.php
```diff
@@ -0,0 +1,37 @@
...
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Hard-delete all users soft-deleted for at least 5 years';
+
+    /**
+     * Execute the console command.
+     */
+    public function handle()
```
petertgiles, line 28: Could you pull in the logging facade and add some logging to this?  Or maybe [regular command output](https://laravel.com/docs/10.x/artisan#writing-output) will automatically be logged if run from the scheduler?

I suggest a regular "command ran, x users hard deleted" message on every run along with a "user xxx-yyy-zzz (firstname lastname, trashed at Jan 1, 2000) hard deleted" for every deletion.

api/app/Models/User.php
petertgiles: If it would collide, maybe don't adjust the email instead of just cancelling the operation?

api/app/Providers/AuthServiceProvider.php
```diff
@@ -92,8 +92,12 @@ public function resolveUserOrAbort($bearerToken, $tokenService): ?User
 
             // By this point we have verified that the token is legitimate
 
-            $userMatch = User::where('sub', $sub)->first(); // 3. match "sub" claim to user 'sub' field.
+            $userMatch = User::where('sub', $sub)->withTrashed()->first(); // 3. match "sub" claim to user 'sub' field.
             if ($userMatch) {
+                if ($userMatch->deleted_at != null) {
```
petertgiles, line 97: Hmm, I'm worried about this.  Even though the app pages are guarded the deleted user can still interact with the API like before.  Could you return a 401 with an error message that is recognized by the app?

petertgiles (changes requested):
It works!  A few more things to check on but this is looking great.

apps/web/src/components/Router.tsx
petertgiles: :+1:

api/app/Models/User.php
tristan-orourke: I prefer peter's first instinct - let's not remove anything automatically. I'm worried about that being a bit brittle. If an admin restores a user, they can manually fix the email.

In fact - hear me out - what if we leave the the `deleted-at` section and stack on a new `restored-at` section as well! 😄

tristan-orourke:
Agreed with all @petertgiles' comments, but great work @JamesHuf! We'll not schedule any tasks yet, but the command makes sense and may give us a headstart when we decide how to deal with purging data for real.

api/app/Models/User.php
vd1992: Can't help but wonder how many toggles, deleted to/from restored it takes to break something. Just out of sheer curiosity than concern about someone actually trying it

api/app/Console/Kernel.php
vd1992: Nice stuff!
Yeah, the prevent overlapping would likely come into play for second and onward scheduled events

api/app/Models/User.php
JamesHuf: It looks like total email length can't exceed 255 characters.

apps/web/src/pages/Users/UpdateUserPage/components/DeleteUserSection.tsx
```diff
@@ -0,0 +1,44 @@
...
+        {intl.formatMessage({
+          defaultMessage: "Delete user",
+          id: "OaF7ls",
+          description: "Heading for section to delete a user",
+        })}
+      </Heading>
+      <p data-h2-padding="base(x1, 0)">
+        {intl.formatMessage({
+          defaultMessage:
+            'This will change the status of a user as "Removed" from the platform. This will prevent the user from appearing anywhere on the platform. <strong>This action cannot be undone</strong>.',
```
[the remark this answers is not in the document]
JamesHuf: @Jerryescandon Thoughts on this copy change to the text here?
image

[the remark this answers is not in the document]
Jerryescandon: we can drop "...from the platform"
and just leave it as:
`This will change the status of a user to "Removed".`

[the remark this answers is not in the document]
Jerryescandon: Although, following a conversation I had with @tristan-orourke today, maybe we should say
`This will change the status of a user to "Archived".`

api/app/Models/User.php
JamesHuf: I changed it to tack on a `restored-at` instead now.

api/app/Console/Commands/HardDeleteOldUsers.php
JamesHuf: Added some logging with [dff6504](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7593/commits/dff6504706f18d044ba6db75af13374243c55644).

api/tests/Feature/HardDeleteOldUsersTest.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+use App\Models\User;
+use App\Console\Commands\HardDeleteOldUsers;
+use Carbon\Carbon;
+use Illuminate\Support\Facades\Artisan;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Tests\TestCase;
+
+class HardDeleteOldUsersTest extends TestCase
```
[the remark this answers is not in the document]
JamesHuf: Added a test for this with [13f18c1](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7593/commits/13f18c190944dc2f130f84dadc2c7bda0ba32a1f).

api/app/Console/Kernel.php
```diff
@@ -15,7 +16,7 @@ class Kernel extends ConsoleKernel
      */
     protected function schedule(Schedule $schedule)
     {
-        // $schedule->command('inspire')->hourly();
+        $schedule->command(HardDeleteOldUsers::class)->dailyAt('08:00');
```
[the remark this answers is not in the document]
JamesHuf: Commented out this call for now with [df7075b](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7593/commits/df7075bf78f79a40b15e291c72864f93a50cfeb6).

apps/web/src/pages/Users/UpdateUserPage/components/DeleteUserDialog.tsx
```diff
@@ -0,0 +1,171 @@
...
+  const label = intl.formatMessage({
+    defaultMessage: "Delete this user",
+    id: "5jm/N7",
+    description: "Label on the button to open the delete user dialog",
+  });
+
+  return (
+    <Dialog.Root open={isOpen} onOpenChange={setIsOpen}>
+      <Dialog.Trigger>
+        <Button
```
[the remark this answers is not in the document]
JamesHuf: Changed this with [4c5a69c](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7593/commits/4c5a69c737357b41843654c83f154a643df965ad).

apps/web/src/pages/Users/UpdateUserPage/components/DeleteUserSection.tsx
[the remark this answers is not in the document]
JamesHuf: Copy updated with [de7cfd7](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7593/commits/de7cfd7720436ace421d55b06793c3dfc2f6aa53).

apps/web/src/hooks/useRoutes.ts
```diff
@@ -47,6 +47,7 @@ const getRoutes = (lang: Locales) => {
     register: () => path.join(baseUrl, "register-info"),
     login: () => path.join(baseUrl, "login-info"),
     loggedOut: () => path.join(baseUrl, "logged-out"),
+    userDeleted: () => path.join(baseUrl, "user-deleted"),
```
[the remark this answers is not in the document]
JamesHuf: This should be working now with [763d5d7](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7593/commits/763d5d7a5eb7171c2c43e0b09cdcbc7b5465a17f).

api/app/Providers/AuthServiceProvider.php
JamesHuf: Updated to prevent calls from deleted users with [27fce5b](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7593/commits/27fce5b8526556960ec66f77a6439dd478cb0a97).

api/app/Console/Commands/HardDeleteOldUsers.php
petertgiles: Wonderful!

api/tests/Feature/HardDeleteOldUsersTest.php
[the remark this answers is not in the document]
petertgiles: Beautiful!

api/app/Console/Kernel.php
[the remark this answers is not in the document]
petertgiles: Nice.  Leaving it as a comment gives us a head start in the future.  :+1:

apps/web/src/pages/Users/UpdateUserPage/components/DeleteUserDialog.tsx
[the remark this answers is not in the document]
petertgiles: :+1:

apps/web/src/hooks/useRoutes.ts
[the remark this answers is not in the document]
petertgiles: :+1:

api/app/Providers/AuthServiceProvider.php
petertgiles: Nice!

petertgiles (approved):
Love it!

PR #7596, opened 2023-08-14
[Bug] Prevent submitting both Status and Non-status

### LINKED ISSUE #7521: 🐛 Non-status and status First Nations values both true (backend validation only)
## 🐛 Bug

Both "I am Status First Nations" and "I am Non-Status First Nations" First Nations values are able to be saved in the profile.

## 🦋 Expected Behaviour

If one option is true, the other cannot be saved.

## 📋 Steps to Reproduce

1. Navigate to profile, Diversity, equity, and inclusion section
2. Click Edit this information for I affirm that I am First Nations (status or non-status), Inuk (Inuit), or a Métis person.
3. Check "I am Status First Nations" and "I am Non-Status First Nations" checkboxes for First Nations field
4. Submit form
5. Observe both values saved

## 📹 Screen recording

https://github.com/GCTC-NTGC/gc-digital-talent/assets/3046459/9b34018c-048f-465f-87a0-dc9ecfe3409a

## 🙋‍♀️ Proposed Solution

Use backend validation to allow for saving only one.

## ✅ Acceptance Criteria

- [ ] both "I am Status First Nations" and "I am Non-Status First Nations" First Nations values cannot be saved simultaneously in the database
- [ ] no frontend validation, backend only
- [ ] use default error message from Laravel, no copy provided from team yet

vd1992:
🤖 Resolves #7521

## 👋 Introduction

Adds backend validation to prevent setting communities as containing both Status and Non-status. 

## 🕵️ Details

Validation rule added. 
Small tweak to the toasting. 
API message returned that can be connected to a proper message to toast later on. 

## 🧪 Testing

1. Navigate to `/en/users/:id/profile#diversity-equity-inclusion-section`
2. Submit different states for communities in Self-declaration
3. Confirm it works as expected, but errors for when you try and pick Status and Non-status

## 📸 Screenshot

image


apps/web/src/components/EmploymentEquity/EquityOptions.tsx
```diff
@@ -106,13 +106,12 @@ const EquityOptions = ({
...
 
   const handleMultipleFieldSave = (data: UpdateUserAsUserInput) => {
-    onUpdate(data)
-      .then(() => {
-        toast.success(intl.formatMessage(profileMessages.userUpdated));
-      })
-      .catch(() => {
+    onUpdate(data).then((res) => {
+      if (res) toast.success(intl.formatMessage(profileMessages.userUpdated));
+      else {
```
vd1992, line 111: For the purpose of preventing the below

gif

petertgiles (dismissed):
Love the use of Laravel rules in the validator and the new test!  Thanks for fixing that response validation bug, too.  :medal_sports:

petertgiles:
> this leads to a worse UX

That's a fair point.  We must have code somewhere that extracts the validation error messages from a graphql response, right?  Even if it just displayed "BothStatusNonStatus" in a toast, that would be _something_.

esizer:
> That's a fair point. We must have code somewhere that extracts the validation error messages from a graphql response, right? Even if it just displayed "BothStatusNonStatus" in a toast, that would be _something_.

Yup, I believe it just needs to exist as a key in [`i18n/apiMessages`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/packages/i18n/src/messages/apiMessages.ts) and it will pick that up and display the value of that instead of the generic GraphQL error.

```tsx
{
  BothStatusNonStatus: 'BothStatusNonStatus',
}
```

api/app/GraphQL/Validators/UpdateUserInputValidator.php
```diff
@@ -45,6 +47,7 @@ public function messages(): array
         return [
             'email.unique' => 'EmailAddressInUse',
             'sub.unique' => 'SubInUse',
+            'indigenousCommunities' => 'BothStatusNonStatus',
```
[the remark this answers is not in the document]
vd1992: I think you are right, it does work
https://github.com/GCTC-NTGC/gc-digital-talent/pull/7596/commits/901af198c065a736edf83a7e9c2e760db32f74fa

api/app/Rules/IsStatusOrNonStatus.php
```diff
@@ -0,0 +1,25 @@
...
+     * Run the validation rule.
+     */
+    public function validate(string $attribute, mixed $value, Closure $fail): void
+    {
+        if (isset($value)) {
+            if (
+                in_array(ApiEnums::INDIGENOUS_STATUS_FIRST_NATIONS, $value) &&
+                in_array(ApiEnums::INDIGENOUS_NON_STATUS_FIRST_NATIONS, $value)
+            ) {
+                $fail('BothStatusNonStatus');
```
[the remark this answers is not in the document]
vd1992: Set up something in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7596/commits/5af44a06ea21b491a0d0042c4972bd9515039654

vd1992:
I attempted to toast the message to the client, it was a bit of a struggle. 
I got it to work in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7596/commits/d922d667601d96496f32a3a225d2e2b5031b2e7b

The issue being that the validation keys are a bit buried and not being pulled out currently, it outputs whatever exists for the field `graphQLErrors.messages` which seems autogenerated sometimes? 

image

packages/i18n/src/lang/whitelist.yml
```diff
@@ -17,3 +17,4 @@
 - z0JksZ # Inuit
 - gTB9r8 # Inuk (Inuit)
 - uNArwM # '+123243234' - Placeholder for form telephone field
+- ybzpGb # temporary output Api message to client
```
vd1992, line 20: Appears to have done nothing so I added it to `i18n/src/lang/fr.json`

vd1992:
To add on to the above point, the toasting issue is a thing elsewhere too. 
On main, updating a pool for instance. The closing date validation key isn't getting pulled out. It sticks to some autogenerated message. 

image

My change seems to have fixed that

image

packages/i18n/src/lang/whitelist.yml
petertgiles: Yeah, I've noticed this, too.  I'm not sure exactly what the expected behaviour of this file is.

packages/client/src/components/ClientProvider/ClientProvider.tsx
```diff
@@ -161,7 +162,13 @@ const ClientProvider = ({
               error: CombinedError,
               operation: Operation<unknown, AnyVariables>,
             ) => {
-              const errorMessages = extractErrorMessages(error);
+              let errorMessages = extractErrorMessages(error);
+
+              const validationMessageKeys = extractValidationMessageKeys(error);
+              if (validationMessageKeys) {
+                errorMessages = validationMessageKeys;
```
petertgiles, line 169: This will overwrite error messages from the first function, right?  Why not get the results of both and concatenate them?  Will there be duplicates?

packages/client/src/utils/errors.tsx
```diff
@@ -8,6 +8,23 @@ import { tryFindMessageDescriptor } from "@gc-digital-talent/i18n";
 export const extractErrorMessages = (combinedError: CombinedError) =>
   combinedError.graphQLErrors.flatMap((error) => error.message);
 
+// grab the validation error enums to map them to messages, guarding against possible missing fields
+export const extractValidationMessageKeys = (
+  combinedError: CombinedError,
+): string[] | null => {
+  const { extensions } = combinedError.graphQLErrors[0];
```
petertgiles, line 15: It makes me nervous to access this array without first checking the length.  Also, what if there are more than one error?  Maybe you could use a flatMap like the function above so it works regardless of the length?

packages/client/src/utils/errors.tsx
```diff
@@ -8,6 +8,23 @@ import { tryFindMessageDescriptor } from "@gc-digital-talent/i18n";
...
+  if (extensions && extensions.validation) {
```
petertgiles, line 16: Just a stylistic note but if you use optional chaining you can avoid the first check.
`if (extensions?.validation) {`

packages/client/src/utils/errors.tsx
```diff
@@ -8,6 +8,23 @@ import { tryFindMessageDescriptor } from "@gc-digital-talent/i18n";
...
+    const validationObject = extensions.validation as object;
```
petertgiles, line 17: Using `as` to override the type checker makes me nervous.  I'd prefer to see a real check like `typeof extensions.validation === 'object'`.

petertgiles (dismissed):
Oof, this one got a little more involved than we expected.  It works nicely!  I'd like to see a little refining on the extract function though.

packages/client/src/components/ClientProvider/ClientProvider.tsx
vd1992: Yeah, I did that intentionally. Only overrides if there are validation error enums. 
Sometimes it isn't a very useful message (Validation failed for the field ...) and sometimes it duplicates. 
For cases like authorization problems, there won't be a validation messages array so it returns the first function. 

If I concatenate  
`error.message` is returning what is in the array `error.extensions.validation`

image

packages/client/src/utils/errors.tsx
vd1992: Attempted in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7596/commits/a90863430b959908199ce039635c4ffd8a0ac476

packages/client/src/components/ClientProvider/ClientProvider.tsx
```diff
@@ -161,7 +159,13 @@ const ClientProvider = ({
               error: CombinedError,
               operation: Operation<unknown, AnyVariables>,
             ) => {
-              const errorMessages = extractErrorMessages(error);
+              let errorMessages = extractErrorMessages(error);
+
+              const validationMessageKeys = extractValidationMessageKeys(error);
+              if (validationMessageKeys) {
```
vd1992, line 165: This will return an array always, either with error enums or an empty array, in the case of the latter I do not think we want to replace errorMessages otherwise we lose things like error messages due to authorization

```suggestion
              if (validationMessageKeys.length > 0) {
```

vd1992:
Thank you Peter for guarding my types 🙇 
@esizer 
Could I please get a review from you?

esizer (approved):
Works well for me 👍 Hopefully we get a proper error message soon 😅

PR #3960, opened 2022-09-14
Backport v1.1.5 hotfixes to main

petertgiles:
This branch backports the fixes from v1.1.5 back to main

## Testing
- [ ] Pool candidates show with their exact status (not draft) even with not submitted_at date
- [ ] The pool candidates editor still exists and is functional in the admin site
- [ ] There is no db migration to move data from pool_candidates to users

Closes #3946

PR #7608, opened 2023-08-16
[Debt] Remove my applications page

### LINKED ISSUE #7594: ♻️ Remove unused `MyApplicationsPage`
## ♻️ Debt/Refactor

We no longer need this page (`MyApplicationsPage`) so it can safely be removed.

## 🕵️ Details

This page was used to display a list of applications for the user. The user applications now appear on their dashboard (`/applicant/profile-and-applications`) and we no longer link to this page anywhere.

https://github.com/GCTC-NTGC/gc-digital-talent/discussions/7590#discussioncomment-6706361

## 🙋‍♀️ Proposed Solution

Delete the page and move any components to the appropriate place (new page or central components).

## ✅ Acceptance Criteria

- [ ] `MyApplicationsPage` removed
- [ ] Route removed from the router
- [ ] Page components moved to appropriate location
- [ ] `nginx` config updated to redirect `/applications` to `/applicant/profile-and-applications` (see dashboard redirect)

esizer:
🤖 Resolves #7594 

## 👋 Introduction

Removes the "My applications" page and redirects the path to the new applicant dashboard.

## 🧪 Testing

2. Refresh the container `docker-compose down && docker-compose up --build --detach`
3. Confirm `/applications` redirects to `/profile-and-applications` for both locales and no locale


vd1992:
The redirects work
It builds right
The application process works, I can see the cards and statuses appear right. Using admin dashboard to update statuses. 

Except for one thing, which I am not sure if it was an existing issue or not, expiry dates seem to have no effect.
Tricky, because the card logic is a little complicated and there is more than one type. 

image

image

image

vd1992:
Hang on, could be due to the fact we have a pool candidate status called expired as well as an expiry date and they do not cooperate maybe. 

Ah, we do have an override to sort that out, it is still commented out

image

esizer:
> The application process works, I can see the cards and statuses appear right. Using admin dashboard to update statuses.
> 
> Except for one thing, which I am not sure if it was an existing issue or not, expiry dates seem to have no effect. Tricky, because the card logic is a little complicated and there is more than one type.

I'm confused, this is not part of the admin side at all 😕 . This was only being used on the applicant side as the old page where we listed applications for the user before we created `/profile-and-applications`

vd1992:
> > The application process works, I can see the cards and statuses appear right. Using admin dashboard to update statuses.
> > Except for one thing, which I am not sure if it was an existing issue or not, expiry dates seem to have no effect. Tricky, because the card logic is a little complicated and there is more than one type.
> 
> I'm confused, this is not part of the admin side at all 😕 . This was only being used on the applicant side as the old page where we listed applications for the user before we created `/profile-and-applications`

The second comment I squeezed in just before you replied, I think covers it.
I was exercising track application cards to make sure they were working right still with the updated imports where I surfaced an existing quirk. 

Expiry date and the expiry status do not mesh as that accessor override is still commented out. Then if logic based off expiry status is present, then changing the expiry date has no effect as expiry date and expired status are decoupled. 
I think the expected behaviour would be set expiry date -> past expiry date -> candidate functions like status is set to expired. Maybe. 

So in summary, I think I stumbled upon an existing case of something that may not be functioning as someone might expect and may need to have its own issue created.

vd1992:
@petertgiles 
You commented out the status override in https://github.com/GCTC-NTGC/gc-digital-talent/pull/3960/commits/a3aac35e49601b14a100007567d1ce4c38026175
Is that something we ever want to reverse? For cases whether expiry as a status and expiry due to date should converge in behaviour.

petertgiles:
> You commented out the status override in [a3aac35](https://github.com/GCTC-NTGC/gc-digital-talent/commit/a3aac35e49601b14a100007567d1ce4c38026175)

I remember commenting it out for old applications that went in before direct intake.  They should be live but have no submitted date.  I remember having a conversation later about whether to put it back in and we decided not to.  I don't remember clearly the reason, though.  Maybe that it was a bit too clever or that DCM wanted to manually manage statuses?

vd1992:
> > You commented out the status override in [a3aac35](https://github.com/GCTC-NTGC/gc-digital-talent/commit/a3aac35e49601b14a100007567d1ce4c38026175)
> 
> I remember commenting it out for old applications that went in before direct intake. They should be live but have no submitted date. I remember having a conversation later about whether to put it back in and we decided not to. I don't remember clearly the reason, though. Maybe that it was a bit too clever or that DCM wanted to manually manage statuses?

Ah, so I shouldn't sweat `expiryDate` vs `PoolCandidateStatus.Expired` then
Thanks!

vd1992 (approved):
Tangent aside, I think all is good 👍

PR #7678, opened 2023-08-21
[Feature] Backend for the skills showcase 

### LINKED ISSUE #7387: ✨ Build backend to support "Skill showcase"
## ✨ Feature

We need to build the backend which will allow us to build the "Skill showcase" page #7386.

## 🕵️ Details

For their Skill showcase, users can define the "top skills" and "skills you'd like to improve". They can do this separately for Behavioral and Technical skills, for a total of 4 lists. Order matters for each of these lists, since users can choose how to order skills.

## 🎨 Design File

Figma: https://www.figma.com/file/9Bq2VP39MW4hXKH7RfrovZ/Candidates---Skill-showcase?node-id=143%3A27334&mode=dev

## 📸 Screenshot

<details><summary>For convenience only - remember to work from figma to stay up-to-date!</summary>

</details>

I think we should just add two nullable Int fields to UserSkill. For sake of discussion lets say they're `topShowcaseNumber: Int` and `improveShowcaseNumber: Int` (better names decided by implementer).

If the number is null for a UserSkill, it's not in the list. Otherwise, it can be set to any number by the api mutation. We don't check for uniqueness, because Behavioural and Technical skills share the same field. We leave it up to the frontend to decide how many are allowed in each list - it will be easy for the frontend to sort them, whatever the numbers are, so I'm not worried about validating the values.

Update: actually, considering we'll want to set the order for [a number of skills at the same time](https://www.figma.com/file/LXvsSV76DcjExZZMCX0Iif/Candidates---Update-a-skill-showcase?type=design&node-id=2-3787&mode=design&t=nuKnPz9nmiUI8EgK-0) (#7378), it might be a pain to set the number with a migration for each skill individually. Maybe we need a mutation which simply takes an array of skill ids and syncs it up. It could even be the above implementation in the database if we handle it with migration class.

Even for querying, we _could_ do the ordering on the backend. Use a Laravel Accessor to put a field on the User object which gives the ordered lists straight up, like...
```
type User {
  userSkills: [UserSkill]
  topBehaviouralSkills: [UserSkill]
  topTechnicalSkills: [UserSkill]
}
```
I don't know if that's better than letting the client figure it out, but something to consider.

## ✅ Acceptance Criteria

- [ ] When querying UserSkills, the client will be able to determine which skills are in the showcase lists and in what order
- [ ] When updating UserSkills, the client will be able to change the showcase lists
- [ ] Factory and seeder are updated so that `applicant@test.com` user has some skills in each list.
- [ ] PHPUnit test

vd1992:
🤖 Resolves #7387 

## 👋 Introduction

Sets up the backend, so migrations, fields, and mutations , for the showcase. 

## 🕵️ Details

- Fields added
- Accessor to grab sorted rankings
- Mutation class to update rankings
- Seeding/factory updated.
- Testing added

## 🧪 Testing

1. Migrate/seed
2. Play with operations posted below

Querying the rankings:

```
  me {
    topTechnicalSkillsRanking {
      topSkillsRank
    }
    topBehaviouralSkillsRanking {
      topSkillsRank
    }
    improveTechnicalSkillsRanking {
      improveSkillsRank
    }
    improveBehaviouralSkillsRanking {
      improveSkillsRank
    }
  }
```

Updating the rankings:

```
mutation update
{
  updateUserSkillRankings(
    userId: ":userId",
    userSkillRanking: {
      topTechnicalSkillsRanked: [":userskillIdTechnical", ":userskillIdTechnical", ":userskillIdTechnical"],
      topBehaviouralSkillsRanked: [":userskillIdBehavioural", ":userskillIdBehavioural", ":userskillIdBehavioural"],
    }
  )
  {
		id
    }
}
```

## 📸 Screenshot

image


api/database/seeders/DatabaseSeeder.php
```diff
@@ -87,16 +87,60 @@ public function run()
...
+                        break;
+                    case 1:
+                        $applicantUserSkill->improve_skills_rank = 1;
+                        break;
+                    case 2:
+                        $applicantUserSkill->top_skills_rank = 2;
+                        break;
+                    case 3:
+                        $applicantUserSkill->top_skills_rank = 3;
+                        $applicantUserSkill->improve_skills_rank = 2;
```
esizer, line 115: I'm confused. Is this being marked as both "to improve" and "top skill" at the same time? 😕 I guess it isn't impossible but it seems unlikely that this would represent real data 😅

esizer (commented):
I just left some thoughts. Some like the `each` should be addressed but mostly I'm just trying to poke around and get a sense of the direction we are taking with this that I couldn't sort out from the ticket it is addressing.

api/app/GraphQL/Mutations/UpdateUserSkillRankings.php
```diff
@@ -0,0 +1,92 @@
...
+                ->each
+                ->update(['top_skills_rank' => null]);
+
+            // set the rankings using the input array of UserSkill ids
+            $rankIterator = 1;
+            foreach ($arrayUserSkillId as $userSkillId) {
+                $userSkillsCollection->where('id', $userSkillId)->each
+                    ->update(['top_skills_rank' => $rankIterator]);
+                $rankIterator++;
+            }
```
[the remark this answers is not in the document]
vd1992: 🤷 
I just went with the suggested idea using arrays of skill ids. 

image

api/database/seeders/DatabaseSeeder.php
vd1992: I thought it was possible for a skill you want to both showcase as a top skill and be something you also want to improve further. So intentionally both.

esizer: Okay, that sort of makes sense. It just seems odd that you would claim a skill as being one of your top skills _**and**_ wanting to improve it. 🤣  Shouldn't we all be striving to always improve all skills we have? 😛 

I guess it seems like we are easily offering users a silly cop out where they just fill up "skills to improve" with all of their "top skills" to make it seem like they have no weaknesses 😕

vd1992: > Shouldn't we all be striving to always improve all skills we have?

Life lessons brought to you by Eric 

> I guess it seems like we are easily offering users a silly cop out where they just fill up "skills to improve" with all of their "top skills" to make it seem like they have no weaknesses

"My weakness is I just can't think of any weaknesses among my endless strengths". 
Might be something to ask, not sure who to ask since that is more a question about the overall big picture

api/app/GraphQL/Mutations/UpdateUserSkillRankings.php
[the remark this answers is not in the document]
tristan-orourke: Pros and cons...
Individual skill update:
  `+` more intuitive/straghtforward API?
  `+` allows adjusting parts of the list (easily remove one skill, swap two skills, add new skill to top or bottom of list)
  `-` requires more work on the frontend (must keep track of skills removed from list, so they can be removed on API. Order numbers must be determined manually.)
  `-` simplest frontend implementation may end up updating all skills every time anyway

Sync-style update (current implementation):
  `+` simple frontend implementation (just current skills, in order)
  `+` setting numbers on backend means we never have duplicates
  `-` no partial list updates
  `-` less intuitive API?

api/app/GraphQL/Mutations/UpdateUserSkillRankings.php
```diff
@@ -0,0 +1,92 @@
...
+                ->each
+                ->update(['improve_skills_rank' => null]);
+
+            $rankIterator = 1;
+            foreach ($arrayUserSkillId as $userSkillId) {
+                $userSkillsCollection->where('id', $userSkillId)->each
+                    ->update(['improve_skills_rank' => $rankIterator]);
+                $rankIterator++;
+            }
+        }
```
tristan-orourke, line 86: If we do keep this approach...
You're running essentially the same code 4 times. Try putting it in a function. I think the function would take 4 args: $userSkillsCollection, $arrayUserSkillId, $category and $column (top_skills_rank/improve_skills_rank).

api/app/GraphQL/Mutations/UpdateUserSkillRankings.php
```diff
@@ -0,0 +1,92 @@
...
+        $userSkillRankingInput = $args['userSkillRanking'];
+
+        // execute blocks depending on whether the value at args.userSkillRankingInput.X is non-null
+        if (isset($userSkillRankingInput['topTechnicalSkillsRanked'])) {
+
+            // wipe existing technical top skills ranking
+            $arrayUserSkillId = $userSkillRankingInput['topTechnicalSkillsRanked'];
+            $userSkillsCollection->where('skill.category', 'TECHNICAL')
+                ->each
+                ->update(['top_skills_rank' => null]);
```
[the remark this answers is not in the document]
esizer: Oh, yeah I see why. It is because collection and not an eloquent relationship. I guess it is just another subject one 😅 

This does works as well but I have no idea if either is preferred. What you have is working so let's stick with it 👍 

```php
// wipe existing technical top skills ranking
$user->userSkills()->whereHas('skill', function ($query) {
  return $query->where('category', 'TECHNICAL');
})->update(['top_skills_rank' => null]);
```

api/app/GraphQL/Mutations/UpdateUserSkillRankings.php
vd1992: Okey, done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7678/commits/d3a41084d287930370066275049438204a3c6bd7

api/app/Models/User.php
```diff
@@ -810,6 +810,58 @@ public function getUnreadEnrichedNotificationsAttribute()
...
+
+    public function getTopBehaviouralSkillsRankingAttribute()
+    {
+        $behaviouralUserSkills = $this->userSkills()
+            ->whereNotNull('top_skills_rank')
+            ->whereHas('skill', function ($query) {
+                $query->where('category', 'BEHAVIOURAL');
+            })
+            ->get();
+        $sortedBehaviouralUserSkills = $behaviouralUserSkills->sortBy('top_skills_rank', SORT_NUMERIC);
```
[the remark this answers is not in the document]
vd1992: I think that might be nicer too, done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7678/commits/ecf0bce5c3192787886cfd29fe667160a479ace4

esizer:
Not sure if I am doing something wrong but running the queries does not update the ranks. 

Screenshot 2023-09-01 123132
Screenshot 2023-09-01 123304

api/app/GraphQL/Mutations/UpdateUserSkillRankings.php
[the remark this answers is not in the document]
vd1992: Sounds good to me!

api/app/GraphQL/Mutations/UpdateUserSkillRankings.php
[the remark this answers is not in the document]
esizer: Based on the discussion during stand up we are going to go with the current solution and update if needed. Resolving this conversation 👍

vd1992:
Executing

image

Gives me

image

vd1992:
Hm... Double check your data input maybe? Hard to tell from here

esizer:
> Gives me

@vd1992 you know what, that was my fault. I think I'm really misunderstanding the goal of this ticket/PR. 

I was under the assumption it accepted an array of `Skill` IDs and it would claim them and set the rank at the same time. Looks like you need to claim the skill before adding it to your showcase. I assume there is a good reason for this but as a user I would find it very convenient to do it all in one shot. Seems slow this way 🤔 

Would the workflow be:

1. Login
2. Go to "Skill Library"
3. Claim a bunch of skills
4. Navigate back to dashboard
5. Navigate to Showcase
6. Add claimed skills to one of the four areas?

I'm wondering if we should look at allowing users to do all skills based work on a single page rather than needing to jump around the application in the future.

vd1992:
Oh, I was assuming the goal was that it acts upon `UserSkill` models you "claimed" elsewhere like on the Skills Library. The page using the mutation would then let you order your `UserSkill` models and then submit them ordered as desired. Then when the page loads it fetches those models with metadata and attached skill. Which you can then continually re-sort.  

I think that would a big shift in design as claiming skills to fill your library a pretty distinct step as is then forming a showcase out of them right?
I could be misinterpreting goals of the overarching feature here.

esizer:
> I think that would a big shift in design as claiming skills to fill your library a pretty distinct step as is then forming a showcase out of them right?

Yeah, based on the AC you are right. It just seems odd to me that we can't do both at the same time. I could see someone wanting to do their showcase first and then set their proficiency at a later time. This way seems less flexible but maybe there is a reason for this that I'm just missing.

I was just assuming we would let users build their profile in any order they want as long as it is complete before applying but it seems like we want them to do it in a specific order.

1. Add experiences
2. Claim skills
3. Link Experiences
4. Set showcase skills

I don't see any reason why it needs to be this order except for that is just the way we set it up 🤷‍♀️

tristan-orourke:
I was making the same assumptions as @vd1992 when I wrote the Acceptance Criteria, but @esizer makes good points. Let's bring in the designer.

@substrae Does a user have to add skills to their "library" before adding to their showcase? I checked Figma for the [update showcase pages](https://www.figma.com/file/LXvsSV76DcjExZZMCX0Iif/Candidates---Update-skill-showcase?type=design&node-id=133-1454&mode=design&t=Rr9Mo7J40cPfZCuo-0), but its not shown there what interface users are actually picking skills from.

esizer:
> I was making the same assumptions as @vd1992 when I wrote the Acceptance Criteria, but @esizer makes good points. Let's bring in the designer.
> 
> @substrae Does a user have to add skills to their "library" before adding to their showcase? I checked Figma for the [update showcase pages](https://www.figma.com/file/LXvsSV76DcjExZZMCX0Iif/Candidates---Update-skill-showcase?type=design&node-id=133-1454&mode=design&t=Rr9Mo7J40cPfZCuo-0), but its not shown there what interface users are actually picking skills from.

Yeah, my understanding was this would work similar to experiences. You can claim a skill by linking it to an experience, you don't need to claim a skill _before_ linking it to an experience. Seems like that would be convenient for users if the skill showcase worked in a similar way.

tristan-orourke:
> Yeah, my understanding was this would work similar to experiences. You can claim a skill by linking it to an experience, you don't need to claim a skill before linking it to an experience. Seems like that would be convenient for users if the skill showcase worked in a similar way.

True. We might even use the same skill picker component. 

If that's the case, we need to modify the mutation quite a bit. It needs to take Skill ids, not UserSkill ids. And it needs to ensure those skills get added to the User if they aren't there yet. Thankfully we already have a method for that: `$user->addSkills($skillIds)`

tristan-orourke:
@substrae says @esizer has it right:
> If a user adds a skill to their showcase, it adds it to their library in the same way that it would via an application (without level)

Sorry for the extra work @vd1992!

vd1992:
😢

vd1992:
Okay... Alright... How does this look now?

https://github.com/GCTC-NTGC/gc-digital-talent/pull/7678/commits/0887cfb14266e1209123deec8023abeefd74205c
https://github.com/GCTC-NTGC/gc-digital-talent/pull/7678/commits/86a05e71a5b3f6c0d78daf92e0bd2c8ba33382d8
https://github.com/GCTC-NTGC/gc-digital-talent/pull/7678/commits/9226ebd68e5b4e16c1014e0e5a9090de1cb3dccc

api/app/GraphQL/Mutations/UpdateUserSkillRankings.php
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Update a user's UserSkill rankings
+     *
+     * @param  array{userId: UUID, userSkillRanking: UpdateUserSkillRankingsInput}  $args
+     * @return User
+     */
+    public function __invoke($_, array $args)
+    {
+        $user = User::find($args['userId'])->load(['userSkills', 'userSkills.skill']);
+        $userSkillsCollection = $user->userSkills;
```
tristan-orourke, line 18: You don't need to load anything yet, since you don't use userSkillsCollection until after you've refreshed user anyway

api/app/GraphQL/Mutations/UpdateUserSkillRankings.php
```diff
@@ -0,0 +1,95 @@
...
+
+        $user->refresh();
+
+        return $user;
+    }
+
+    private function syncRankings(object $userSkillsCollection, array $arraySkillIds, string $skillCategory, string $rankType)
+    {
+        // clear existing ranking for the category and type passed in
+        $userSkillsCollection->where('skill.category', $skillCategory)
```
tristan-orourke, line 83: Thinking about that `each` question again...
We could make this slightly more performant by doing it in one DB call by not using userSkillsCollection at all (meaning we could even skip loading them in the first place) and just doing `UserSkills::where('user_id', $userId)->whereHas('skill', function ($query) { $query->where('category', $skillCategory);})->update(...)`. Similarly for below, start with `UserSkills::where('user_id', ...)` instead of userSkillsCollection.

tristan-orourke (changes requested):
Left suggestions for slightly improving mutation performance, but other than that it looks good to me.

api/app/GraphQL/Mutations/UpdateUserSkillRankings.php
vd1992: Hm... Okay, sure. The collection was straightforward to remove, I don't know how much this changed performance
Done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7678/commits/e7425b61874691f1d8dae5820451beb401a42469

vd1992: Oh, didn't need a `whereHas` instance
https://github.com/GCTC-NTGC/gc-digital-talent/pull/7678/commits/bde556c404ee36ef33a2d36e61e209e42829b450

api/database/seeders/DatabaseSeeder.php
```diff
@@ -87,16 +87,60 @@ public function run()
...
+            if ($faker->boolean(75)) {
                 $applicantUserSkill->skill_level = $faker->randomElement(ApiEnums::skillLevels());
             }
-            if ($isWhenSkillUsedSet) {
+            if ($faker->boolean(75)) {
                 $applicantUserSkill->when_skill_used = $faker->randomElement(ApiEnums::whenSkillUsed());
             }
+            // consistent seeding of applicant's skill rankings that matches what is expected
+            // four types independent of each other, starting from 1, with no duplicates or gaps
+            if ($applicantUserSkill->skill->category === 'TECHNICAL') {
```
tristan-orourke, line 102: I find this use of switch/case a bit confusing and unexpected. Could we do all this imperitively outside the foreach loop?
Something more like...
```
foreach {
  ...
}
// Add skills to showcase lists
$applicantUserTechnicalSkills = < user skills where category is technical >
$applicantUserTechnicalSkills[0]->top_skills_rank = 1;
$applicantUserTechnicalSkills[0]->save()
$applicantUserTechnicalSkills[1]->top_skills_rank = 2;
$applicantUserTechnicalSkills[1]->save();
$applicantUserTechnicalSkills[2]->top_skills_rank = 3;
$applicantUserTechnicalSkills[2]->improve_skills_rank = 1;
$applicantUserTechnicalSkills[2]->save();
```

tristan-orourke (changes requested):
Just one suggestion about the seeder. Almost there!

api/database/seeders/DatabaseSeeder.php
esizer: Just ignore me if I'm out of the loop and misunderstanding this but could't we just use faker for this?

```php
foreach($skills as $skill) {
  $skill->top_skills_rank = $faker->unique()->numberBetween(1,5); // 5 or length of array 🤔 
  $skill->save():
}
```

api/graphql/schema.graphql
```diff
@@ -1909,6 +1922,10 @@ type Mutation {
   updateUserSkill(id: ID!, userSkill: UpdateUserSkillInput @spread): UserSkill
     @update
     @can(ability: "update", find: "id")
+  updateUserSkillRankings(
+    userId: UUID!
```
tristan-orourke, line 1926: Should we validate ids? Like, should we validate that this is a real userId, and that the skillRank arrays are all valid skill ids?
I know its not something we've been doing, but I just ran into an error during testing where I mis-pasted a uuid. If our client code did something like this in production, the user would get an opaque Internal Server Error.
@esizer @petertgiles thoughts?

```
"SQLSTATE[23503]: Foreign key violation: 7 ERROR:  insert or update on table \"user_skills\" violates foreign key constraint \"user_skills_skill_id_foreign\"\nDETAIL:  Key (skill_id)=(4526364b-196a-415a-a362-2305a9a14e02) is not present in table \"skills\". (Connection: pgsql, SQL: insert into \"user_skills\" (\"skill_id\", \"user_id\", \"updated_at\", \"created_at\") values (4526364b-196a-415a-a362-2305a9a14e02, 8259bf4b-4fb5-4f59-b55e-d79ec6dfd09a, 2023-09-05 14:30:03, 2023-09-05 14:30:03) returning \"id\")"
```

esizer: Yeah, I think this could be useful. I wonder if laravel has a built in rule for this, or we could make a `ModelExists` rule or something 🤔

api/database/seeders/DatabaseSeeder.php
vd1992: Okay, I just implemented the above in 
https://github.com/GCTC-NTGC/gc-digital-talent/pull/7678/commits/f1c8681b3d66d9bbd7e4a77eaea5f3c74a6f7cc0

tristan-orourke (approved):
🚀

PR #7685, opened 2023-08-23
[Debt] Migrate category from skill families to skills

### LINKED ISSUE #7575: Move skill type from Skill Family to Skill model

yonikid15:
🤖 Resolves #7575 

## 👋 Introduction

- Adds a new migration that moves the `SkillFamily` category column over to the `Skill` table. If a skill has two skill families with different categories, then the `category` is set to `TECHNICAL`.
- Updates the schema and all graphql operations including `skills { ... }`
- Updates the frontend to use the category from skill `skill.category`, instead of skillFamilies.
- Moves the category column to the Skill table on the admin dashboard.

## 🕵️ Details

## 🧪 Testing

1. Ensure the frontend gets the category from the skill, not skill family, anywhere category is used
2. Ensure the database migration sets skill category values according to existing SkillFamily relationships
3. Ensure the migration doesn't break if a skill is in two families with different categories

## 🚚 Deployment

- Running latest migration


api/database/migrations/2023_08_22_140400_move_category_from_skill_families_to_skill_table.php
```diff
@@ -0,0 +1,57 @@
...
+                ->default(ApiEnums::SKILL_CATEGORY_TECHNICAL);
+        });
+
+        // Iterate through all skills and update category using data from skill families
+        $skills = Skill::all();
+        foreach ($skills as $skill) {
+            $category = $skill
+                ->families()
+                ->firstWhere('category', '=', ApiEnums::SKILL_CATEGORY_TECHNICAL)
+                ->category;
```
vd1992, line 30: Fails to run for me

image

Adding the null coalesce should take care of it

image

vd1992:
After making the change in the comment above, the migration appears to work for updating the skills
There are 184 skills, 120 with technical, 65 with behavioural, 1 with both. So 119 technical only, 64 behavioural only, 1 with both adding to 184. All skills have a category. 
After migration the total skill count stays the same and the one skill with both categories shifts to technical 

BEFORE

image

AFTER

image

apps/web/src/pages/Skills/UpdateSkillPage.tsx
vd1992, line 1: Will need to update this operation/page to allow it to update the database with the selected category in the mutation.

apps/web/src/components/SkillDialog/SkillSelection.tsx
vd1992, line 1: The technical vs behavioural grouping for the skill picker may need some reconsideration.
Maybe an option to filter by all technical skills, an option to filter by all behavioural skills, and then by skill family. Rather than list skill family within categories. 

image

api/app/Models/SkillFamily.php
```diff
@@ -42,15 +39,4 @@ public function skills(): BelongsToMany
...
         return $this->belongsToMany(Skill::class);
     }
-
-
-    public static function scopeTechnical(Builder $query)
-    {
-        return $query->where('category', '=', ApiEnums::SKILL_CATEGORY_TECHNICAL);
-    }
-
-    public static function scopeBehavioural(Builder $query)
```
vd1992, line 52: Got one place still referencing these scopes rather than the new ones. 

image

api/database/seeders/SkillFamilySeeder.php
vd1992, line 1: I think we will need to add this to `SkillSeeder.php`
The migration runs before the seeder, so it seeds nothing for `skill.category` which defaults it to `"TECHNICAL"` per migration. The seeder isn't used for live site, but it is used for local testing. 

That would be a bunch of work though to manually add the category all 184 skills. Maybe spun to its own issue?
Or maybe there is some sort of Linux scripting to make it easier 🤔

yonikid15: Yes indeed,  I'm looking for a quicker way to carry over the category data into the SkillSeeder.php instead of manually adding it in.

petertgiles: If you seed this fresh into a database you could use SQL to make your data modifications and dump it out to a spreadsheet or directly to text.  I think I used Excel originally to make these files.  Let me know if you want to a hand on this.

api/database/migrations/2023_08_22_140400_move_category_from_skill_families_to_skill_table.php
yonikid15: Nice catch! Fixed by [28feb55](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7685/commits/28feb55d935f22fdfb9becf254f8edb7eb9c75cd)

apps/web/src/pages/Skills/UpdateSkillPage.tsx
yonikid15: Fixed by [3819600](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7685/commits/3819600abe86475f20efb500783fcc3c71c347aa)

apps/web/src/components/SkillDialog/SkillSelection.tsx
yonikid15: Ya that's good point. Also, are these getting replaced with the new skill picker? Either way, that'll need to be discussed with designers, and made into a new issue.

api/database/seeders/SkillFamilySeeder.php
vd1992: Have you found a path forward on this?

apps/web/src/components/SkillDialog/SkillSelection.tsx
vd1992: Deciding to not worry about the existing skill picker, instead prioritize implementing the new skill picker

yonikid15:
@vd1992 I paired with Peter and he created a SQL query which built our list of skills including the new `category` column from `skill_families` table.

vd1992:
Spun off observed bug while exercising branch in #7755

vd1992:
Attempting to investigate Cypress tests locally. Cannot. 
Test failing on my machine likely due to discrepancy between default and `CiSeeder`. The `GenericJobTitle` seeder was left in `CiSeeder`, and the application workflows depends on it. 

image

image

api/database/seeders/SkillSeeder.php
```diff
@@ -43,2324 +18,3718 @@ public function run()
...
+                    ],
+                    "keywords" => [
+                        "en" => [
+                            "Consideration",
+                            "Appreciation",
+                            "Professional",
+                            "Courteous",
+                        ],
+                        "fr" => [
+                            "Considération",
```
vd1992, line 3732: Something to clean-up

api/database/seeders/SkillSeeder.php
vd1992, line 1: After trying to mess with things to get the test to run locally, my guess is the test is fetching a behavioural skill but is written to expect a technical one. 

**Problem**
It appears to want a technical skill

image

Skill for testing comes from

image

image

**Suggestion**
Ensure the skill attached to test pool is technical and not behavioural

This aside, things look good to me.
We decided to replace the skill picker rather than extensively update the current one to match new categories.

api/database/seeders/SkillSeeder.php
vd1992: `SkillSeeder` otherwise matches my results from above 

image

petertgiles: You're looking at the SQL query?  We wanted to leave it in as a comment to make it easier to recreate the seeder again in the future if necessary.  We could leave it in a separate SQL or readme file or something if you prefer, though.

api/database/seeders/SkillSeeder.php
yonikid15: Yea I found the same thing to be the issue as well. Also the PublishPoolValidator.php should have an extra rule to ensure that the essentalSkills list has a "TECHNICAL" skill or else it shouldn't pass.

api/database/seeders/SkillSeeder.php
vd1992: Oh, if it is intentional then sure leaving it in is fine for reference. But I think somewhere a bit easier to find might help like a readme.

api/database/seeders/SkillSeeder.php
petertgiles: Decision from scrum: No, the essential skills list does not need to include a TECHNICAL skill.

vd1992:
I think the failing spec is just missing an import

vd1992 (approved):
Woohoo! 🏁

PR #7733, opened 2023-08-24
[Fix] Move experiences to `UserSkill` from `Skill`

### LINKED ISSUE #7688: ♻️  Experiences should be accessed view UserSkill instead of Skill
## ♻️ Debt/Refactor

Experiences can be retrieved from api schema via Skill type.

## 🕵️ Details

Experiences should be retrieved via UserSkill type, and not through Skills.

`experienceSkillRecord` can stay on Skill type, as it is only non-null when skill is nested in Experience.

`experience.experienceSkillRecord` should be non-null when nested in UserSkill instead of Skill.

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] `experiences`, and each experience subtype, are removed from Skill type in api schema
- [ ] `experiences`, and each experience subtype, are added to UserSkill type in api schema
- [ ] And client-side operations which retrieved experiences through skills go through UserSkills instead
- [ ] mutations to link experiences and skills are unaffected

esizer:
🤖 Resolves #7688 

## 👋 Introduction

Moves all experience types to the `UserSkill` model instead of the `Skill` model to better reflect the relationship.

## 🏅 Bonus

We had a component that used this relationship which was not being used so it was deleted instead of being fixed.

## 🕵️ Details

There is an intersection of `User`, `Skill` and `Experience` records with the owner being the `User`. 

Previously you accessed this through deeply nested queries (`user.userSkills.skills.experiences`). This made sense since experiences are linked to a user through skills. We want to flatten this relationship (if you need to know why, ask @tristan-orourke or @esizer ).

In order to do this, we are moving the `Experience` up to the `UserSkill` type instead to get it closer to the owner (`User`). So, you can now query experiences attached to a skill via `User.userSkills.experience`.

## 🧪 Testing

1. Navigate to `/graphiql`
2. Run some queries to get `UserSkill` records
3. Confirm you can get experiences off the user skill
4. Confirm you can no longer get experiences from the `Skill` model

<details>
  <summary>Test query</summary>

```graphql
query UserSkill {
  me {
    userSkills {
      id
      experiences {
        id
      }
    }
  }
}

```

</details>


api/graphql/schema.graphql
```diff
@@ -623,13 +623,6 @@ type Skill {
   keywords: SkillKeywords
   families: [SkillFamily!] @belongsToMany
 
-  # Experiences
```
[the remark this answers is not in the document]
esizer: @tristan-orourke What are your thoughts about this? Do we need these relationships in the actual model anymore?

api/graphql/schema.graphql
```diff
@@ -641,6 +634,19 @@ type UserSkill {
   skill: Skill! @belongsTo
   skillLevel: SkillLevel @rename(attribute: "skill_level")
   whenSkillUsed: WhenSkillUsed @rename(attribute: "when_skill_used")
+
+  # Experiences
+  experiences: [Experience!]
+    @with(relation: "awardExperiences")
```
tristan-orourke, line 640: Nice catch on the `@with` optimization

packages/fake-data/src/fakeUserSkills.ts
```diff
@@ -0,0 +1,61 @@
...
+const generateUserSkill = (
+  skill: Skill,
+  experiences: AnyGeneratedExperience[],
+  uniqueEnforcerId: UniqueEnforcer,
+) => {
+  const uniqueId = uniqueEnforcerId.enforce(() => {
+    return faker.string.uuid();
+  });
+  return {
+    __typename: undefined,
```
tristan-orourke, line 28: Why not set this to the real typename?

packages/fake-data/src/fakeUserSkills.ts
```diff
@@ -0,0 +1,61 @@
...
+    id: uniqueId,
+    skill,
+    user: mockUser,
```
tristan-orourke, line 31: I reccomend passing user in as an argument, with a default value of mockUser

tristan-orourke (approved):
Looks good! Just left a couple suggestions for the faker.

api/graphql/schema.graphql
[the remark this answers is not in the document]
esizer: Makes sense to me! Removed in [remove experience relationship from skill model](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7733/commits/d24608904ca4487da5dc1dde7b5c4a80974ec0a4)

packages/fake-data/src/fakeUserSkills.ts
esizer: That was just a copy and paste error 😅 Good catch!

Fixed in [improve fake user skill generation](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7733/commits/16a0f8fc5c1554e77b89e2ab22f99e60d521c47c)

packages/fake-data/src/fakeUserSkills.ts
esizer: Sounds good, I added it as an arg and set a default incase we don't care who the user is.

[improve fake user skill generation](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7733/commits/16a0f8fc5c1554e77b89e2ab22f99e60d521c47c)

PR #7740, opened 2023-08-24
[Bug] Remove default name for AuthService

### LINKED ISSUE #7684: 🐛 User ID rendered instead of first name
## 🐛 Bug

When reviewing profiles on Admin, some profiles (5+ seen recently) have a first name that is rendering the user id (uuid) instead of the first name value.

## 🦋 Expected Behaviour

The person's first name should be displayed properly.

## 📋 Steps to Reproduce

@gobyrne can provide specific accounts on Prod with this problem if you need it.

## 📸 Screenshot

> [!IMPORTANT]  
> I cropped out the rest of the info on the page, but it's otherwise a complete profile.

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdUV0IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--5cf8e1811a1d11c5463ab42402fb57ed13bb3823/image.png)

## 🙋‍♀️ Proposed Solution

Leave first name `NULL` for new accounts.

## ✅ Acceptance Criteria

- [ ] first name `NULL` for new accounts
- [ ] can't submit a new application without filling in first name field (even if create account is skipped)

vd1992:
🤖 Resolves #7684

## 👋 Introduction

No longer gives a default first name, forcing it to be updated before you can submit applications. 

## 🕵️ Details

Peter did the legwork with identifying the root of the problem. 

## 🧪 Testing

1. Be logged out and then navigate to a pool poster
2. Click apply, be directed to the mock auth
3. Fill in for credentials a completely new user, so fill in random letters for the sub for auth, say "tester1397"
4. Upon signing in you get bounced back to the application and not an account creation page
5. Observe the emptiness

## 📸 Screenshot

image

### DIFF
--- a/api/app/Providers/AuthServiceProvider.php
+++ b/api/app/Providers/AuthServiceProvider.php
@@ -98,7 +98,6 @@ class AuthServiceProvider extends ServiceProvider
             } else {
                 // No user found for given subscriber - lets auto-register them
                 $newUser = new User;
-                $newUser->first_name = $sub;  // displayed on the landing page so should help us find the user
                 $newUser->sub = $sub;
                 $newUser->save();
                 $newUser->syncRoles([  // every new user is automatically an base_user and an applicant


esizer (approved):
LGTM! 👍

PR #7641, opened 2023-08-17
[Debt] Deprecate non-paginated queries

esizer:
🤖 Resolves #4416 

## 👋 Introduction

Deprectaes some non-paginated queries in favour of their paginated versions.

## 🕵️ Details

We have some queries that contain large datasets that require paginated results. We are deprecating the non-paginated versions to be removed in the near future.

## 🧪 Testing

Assist reviewers with steps they can take to test that the PR does what it says it does.

1. Navigate to `/graphql`
2. Confirm the queries are marked as deprecated
3. Confirm the reason stated is accurate

Issue #5137, opened 2022-12-15
♻️ Remove deprecated graphql fields entirely


tristan-orourke:
Find all the fields which have been marked as `@deprecated` in the schema, and remove them from the backend entirely.
Some of these fields are computed, in which case its an Accessor which must be deleted.
In other cases, an entire database column can be dropped.
When removing a field, must also delete...
   - corresponding Inputs in schema
   - corresponding fields in Factory classes
   - instances of that field in tests
   - references in class methods (accessors, scopes, filter methods)

# Acceptance Criteria
1. [ ] Survey all the deprecated fields and decide if any should be kept around for now, for any reason
2. [ ] Delete deprecated fields in schema
3. [ ] Delete related fields in schema Inputs
4. [ ] Drop related columns from database
5. [ ] Remove related accessors in Eloquent classes
6. [ ] Remove references to field in tests
7. [ ] Remove field in Factory classes
8. [ ] _Seek out and destroy_ 🚀 💥  any remaining references to these fields in any other code.

petertgiles:
I wonder if we should close this and instead focus on targeted issues like #5290.   This would make a good issue template though.

petertgiles:
Maybe a spike?

petertgiles:
Current deprecated fields:

user.languageAbility
user.isIndigenous
user.wouldAcceptTemporary
applicant.languageAbility
applicant.isIndigenous
applicant.wouldAcceptTemporary
pool.status  #4183
applicantfilter.wouldAcceptTemporary
query.countPoolCandidates
query.lastestPoolCandidateSearchRequests

mnigh:
Current deprecated graphql queries and fields:

## queries

### `applicant`
- entire query

### `applicants`
- entire query

### `applicantFilter`
- entire query

### `applicantFilters`
- entire query

### `latestPoolCandidateSearchRequests`
- entire query

### `poolCandidates`
- entire query

### `poolCandidateFilter`
- entire query

### `poolCandidateFilters`
- entire query

### `poolCandidateSearchRequests`
- entire query

### `user`
- user.languageAbility
- user.isIndigenous
- user.expectedSalary (#6226)
- user.expectedClassifications (#6226)
- user.wouldAcceptTemporary
- user.pools
- user.expectedGenericJobTitles (#7645)

### `users`
- entire query

## types
### `ApplicantFilter`
- ApplicantFilter.wouldAcceptTemporary

[...]

PR #7744, opened 2023-08-24
[Bug] Remove education filtering of results

### LINKED ISSUE #5976: Search Page - Disable search filtering of Education Requirement
# Description
On the search page, it is possible to filter out users without a post-secondary education.

In the backend, this adds the `hasDipoma=true` filter to the request.

However, while we previously allowed DCM to set the `hasDiploma` field when manually creating Pool Candidates, it doesn't appear anywhere in the profile! This means that any user who signs up and completes their own profile will automatically be excluded from any requests which require a diploma.

We have decided to disable the search filtering so as to not exclude anybody until the new education experience logic is complete.  However, the choice should still get communicated to DCM so that can consider in their work.

# Acceptance Criteria
- [ ] Search page selection of "required diploma" does NOT filter out any results
- [ ] Update `countPoolCandidates` and `countUsers` queries (approximate names for queries, might be slightly different) to not include that filter in counts
- [ ] Search page selection of "required diploma" DOES get saved in request that DCM receives
- [ ] Admin site searches including "required diploma" does NOT filter out any results
- [ ] has_diploma field in schema marked as deprecated

*Cloned from #4970 as per [this](https://github.com/GCTC-NTGC/gc-digital-talent/issues/4970#issuecomment-1331338130)

vd1992:
🤖 Resolves #5976 

## 👋 Introduction

Prevents the selection of education option from having an effect on API operations. 

## 🕵️ Details

Disconnects the education field in the data massage function called before an operation is submitted to the API. Otherwise, the education option is unchanged in that it is select-able everywhere. 
Deprecates the field `hasDiploma` as well. 

## 🧪 Testing

1. Note results logged for search results off main at `/en/search` and `/en/admin/pools/:id/pool-candidates`
2. Switch to branch and build
3. Check the places from step one again, observe education filtering no longer changes result count
4. Submit a request with education selected, observe it is unchanged aside from no longer filtering results
5. Open the request from the admin dashboard, observe things are unchanged aside from the results count

## 📸 Screenshot

image

image

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -92,7 +92,9 @@ type User {
     @rename(attribute: "indigenous_declaration_signature")
 
   # Applicant info
-  hasDiploma: Boolean @rename(attribute: "has_diploma")
+  hasDiploma: Boolean
+    @rename(attribute: "has_diploma")
+    @deprecated(reason: "hasDiploma to be replaced")
   locationPreferences: [WorkRegion] @rename(attribute: "location_preferences")
   locationExemptions: String @rename(attribute: "location_exemptions")
   acceptedOperationalRequirements: [OperationalRequirement]
@@ -551,7 +553,9 @@ input EquitySelectionsInput {
 type PoolCandidateFilter {
   id: ID!
   classifications: [Classification] @belongsToMany
-  hasDiploma: Boolean @rename(attribute: "has_diploma")
+  hasDiploma: Boolean
+    @rename(attribute: "has_diploma")
+    @deprecated(reason: "hasDiploma to be replaced")
   equity: EquitySelections
   languageAbility: LanguageAbility @rename(attribute: "language_ability")
   operationalRequirements: [OperationalRequirement]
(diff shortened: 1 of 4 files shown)


api/graphql/schema.graphql
```diff
@@ -92,7 +92,9 @@ type User {
     @rename(attribute: "indigenous_declaration_signature")
 
   # Applicant info
-  hasDiploma: Boolean @rename(attribute: "has_diploma")
+  hasDiploma: Boolean
+    @rename(attribute: "has_diploma")
+    @deprecated(reason: "hasDiploma to be replaced")
```
mnigh, line 97: The standard is fairly new, but for writing the `reason` for a deprecated graphql entity, do you mind using the following pattern, creating a new issue for the removal, and adding it to the [epic](https://github.com/GCTC-NTGC/gc-digital-talent/issues/5137)?

`oldField` is deprecated. Use `newField` instead. Remove in `#issueNumber`.

REF: https://github.com/GCTC-NTGC/gc-digital-talent/pull/7641#discussion_r1297331161

vd1992: I would've, but it doesn't seem like a replacement has been decided or coming soon
https://github.com/GCTC-NTGC/gc-digital-talent/issues/5976#issuecomment-1679077223

brindasasi (approved):
LGTM!🥳

PR #7745, opened 2023-08-25
[Fix] Make schema permissions more strict

### LINKED ISSUE #7743: ♻️ Tighten up user permissions
## ♻️ Debt/Refactor

1. Ensure users can't see the notifications intended for other users, even if they can see other user's profile.
2. Add some `@can` directives to protect any instances of `user` in the api schema.

## 🕵️ Details

The `user.notifications` and `user.unreadNotifications` fields need some additional protection, to ensure that only the user themself can view the notifications. If I'm an admin, getting a user object to view their profile, I shouldn't be able to see their notifications.

And just to be extra cautious, add `@can(ability: "view", resolved: true)` to any fields which return User, like UserSkill has. (I think that's just PoolCandidate and all the Experiences, but double check.)

Do some manual testing of the site (particularly admin tables) to ensure the policy checks don't blow up the execution time on any queries.

## 🙋‍♀️ Proposed Solution

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Users (even admins) cannot see notifications intended for other users
- [ ] Any field returning `User` does a policy check.

esizer:
🤖 Resolves #7743 

## 👋 Introduction

This does two things:

- Updates `user.notifications` and `user.unReadNotifications` to only show those owned by the user regardless of role
- Adds `@can` to all user fields to protect accessing unauthorized users through other queries

## 🕵️ Details

The additional `@can` do impact query times for certain admin pages, I'm not sure if it has been impacted enough to be considered. Let me know if you find the queries are running too slow!

## 🧪 Testing

1. Login as admin
2. Trigger a notification by logging into admin and changing a pool candidates status
3. Query that users notifications
4. Confirm the array is empty
5. Authenticate as that user
6. Run the same query
7. Confirm you see the notifications

```graphql
query UserNotification($id: ID!) {
  user(id: $id) {
    id
    email
    firstName
    lastName
    notifications {
      id
    }
    unreadNotifications {
      id
    }
  }
}
```


api/app/Models/User.php
```diff
@@ -741,10 +745,42 @@ public function getEnrichedNotificationsAttribute()
...
         return $notifications;
     }
+
+
+    public function scopeAuthorizedToView(Builder $query)
+    {
+        $user = Auth::user();
+
+        if (!$user) {
+            return $query->where('id', null);
```
petertgiles, line 764: This is a short circuit, right?  I wonder if there's a better way that doesn't issue a db request at all?  :shrug:

petertgiles (approved):
Seems to work!  It doesn't seem too slow to me but it's hard to gauge how it will work on the prod server.

api/app/Models/User.php
tristan-orourke: I can't think of a better way to do it either 👍

tristan-orourke: We should consider back-porting this to PoolCandidate.scopeAuthorizedToView, which will just crash if the user is null

PR #7746, opened 2023-08-25
[Fix] `Pending` story stabilized for chromatic

### LINKED ISSUE #7742: 🐛 `Pending` story unstable in chromatic snapshots
## 🐛 Bug

Our `Pending` story has an infinite looped animation that is causing diffs in chromatic depending on at what point during the animation the snapshot is being captured.

## 🦋 Expected Behaviour

Diffs in the `Pending` story do not occur due to animation.

## 📸 Screenshot

Screenshot 2023-08-24 154601

## 🙋‍♀️ Proposed Solution

I believe you can check if it is chromatic in a story file using `chromatic/isChromatic`. Maybe we pause the animation if that returns `true`?

## ✅ Acceptance Criteria

- [ ] `Pending` component still shows animation while previewing in storybook
- [ ] `Pending` story is stable in chromatic and does not create a diff false positive due to animation

esizer:
🤖 Resolves #7742 

## 👋 Introduction

This is an attempt to stabilize the spinner in chromatic snapshots.

## 🕵️ Details

Also, @vd1992 noticed the spinner on the search page disappeared as a result of the table update.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to `/search`
3. Confirm the spinner in the sidebar shows up as expected
4. Check chromatic to see if the spinner is paused(?)


vd1992 (approved):
Pending works again for the search page. 
I think it will also be stabilized for `Pending.stories.tsx` with the added `pause` prop but I am concerned looking at the diff that other stories that reference the pending component will get noisy. 

I will mark as approved though.

PR #7748, opened 2023-08-25
[Fix] Uncheck disclaimer when form changes

### LINKED ISSUE #3709: Disclaimer agreement checkbox unchecks itself
## Description
When editing a previously submitted "Personal experience" the checkbox regarding permission to share automatically uncheckes itself without user intervention.  Strictly speaking, it should only uncheck if ANY changes have been made to the actual information.

## Screenshot: Prototype
image

## Acceptance Criteria

- [ ] Make disclaimer checked if editing an existing experience
- [ ] Uncheck the disclaimer if form is edited

esizer:
🤖 Resolves #3709 

## 👋 Introduction

This makes a couple of changes to the personal experience form:

- Makes the disclaimer checkbox checked by default if editing an experience
- Unchecks the disclaimer if a field other than it has changed from the default values

## 🕵️ Details

If users change their information, we want to make sure they re-agree to share the information since it is technically new info.

## 🧪 Testing

> **Note**
> I ran chromatic as part of this to test #7746

1. Build `npm run dev`
2. Create a personal experience
3. Check the disclaimer
4. Make a change
5. Confirm disclaimer unchecks
6. Save experience
7. Edit the same experience
8. Repeat steps 4-5

### DIFF
--- a/apps/web/src/components/ExperienceFormFields/PersonalFields.tsx
+++ b/apps/web/src/components/ExperienceFormFields/PersonalFields.tsx
@@ -17,7 +17,10 @@ import { SubExperienceFormProps } from "~/types/experience";
 const PersonalFields = ({ labels }: SubExperienceFormProps) => {
   const intl = useIntl();
   const todayDate = new Date();
-  const { setValue } = useFormContext();
+  const {
+    setValue,
+    formState: { defaultValues },
+  } = useFormContext();
   // to toggle whether End Date is required, the state of the Current Role checkbox must be monitored and have to adjust the form accordingly
   const isCurrent = useWatch({ name: "currentRole" });
   // ensuring end date isn't before the start date, using this as a minimum value
@@ -25,8 +28,10 @@ const PersonalFields = ({ labels }: SubExperienceFormProps) => {
   const watchDescription = useWatch({ name: "experienceDescription" });
 
   React.useEffect(() => {
-    setValue("disclaimer", false);
-  }, [setValue, watchDescription]);
+    if (watchDescription !== defaultValues?.experienceDescription) {
+      setValue("disclaimer", false);
+    }
+  }, [defaultValues?.experienceDescription, setValue, watchDescription]);
 
   return (
     <div
--- a/apps/web/src/components/ExperienceFormFields/PersonalFields.tsx
+++ b/apps/web/src/components/ExperienceFormFields/PersonalFields.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import { useIntl } from "react-intl";
 import { useFormContext, useWatch } from "react-hook-form";
-import { isEqual } from "lodash";
+import isEqual from "lodash/isEqual";
 
 import {
   Checkbox,
--- a/apps/web/src/components/ExperienceFormFields/PersonalFields.tsx
+++ b/apps/web/src/components/ExperienceFormFields/PersonalFields.tsx
@@ -32,9 +32,12 @@ const PersonalFields = ({ labels }: SubExperienceFormProps) => {
   React.useEffect(() => {
     const subscription = watch((values, { name }) => {
       if (
-        // Skills are not really part of the overall form
-        !["disclaimer", "skills"].includes(name ?? "") &&
-        !isEqual(values, defaultValues)
+        // Only uncheck when the description changes
+        name === "experienceDescription" &&
+        !isEqual(
+          values?.experienceDescription,
+          defaultValues?.experienceDescription,
+        )
       ) {
         setValue("disclaimer", false);
       }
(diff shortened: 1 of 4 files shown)


brindasasi:
It unchecks when the user makes any changes. But it is really frustrating when creating a new experience.  It keeps clearing for every additional info the user has added 
 @GCTC-NTGC/developers I don't think we should fix like this. 

https://github.com/GCTC-NTGC/gc-digital-talent/assets/89482910/451a1b2a-9c90-45ef-aa9d-4f71d3c3a52b

brindasasi (changes requested):
@GCTC-NTGC/designers could you review this solution again? It keeps clearing the checkbox when new changes are added. Can we move the checkbox down and make it enable only when all other required fields are filled in? So it will be the last field to update ?

esizer:
> If I'm remembering correctly, this disclaimer only applies to the preceding `textarea` element. In an ideal world, this box would only have to be checked once per visit to this page, and only if that specific `textarea` has changed.

Okay, I think that is where the confusion was. I assumed "this" referred to all information on the form 😅 . If this checkbox does only apply to the preceding text area I think we also need to tweak the copy and add some additional context. 

Do we need positive confirmation? If not, would a context description make this a little easier to understand and use? 

Also, are we not _technically_ sharing all information the user provides in their experience with HR advisors? If so, this checkbox is more confusing because it makes it seem like _only_ the description will be shared.

substrae:
@esizer 

I agree the copy could be clearer. This is something I think @Jerryescandon worked on originally, so he might have more insight into why it is the way it is. In this situation, I think the intent was to have positive confirmation deliberately, but to your point, we share everything with HR, so I agree that it might be confusing. As I understand it, we added this for this experience type specifically because the nature of the content being entered here could potentially be more personal, so we were hoping to curtail certain behaviours by forcing them to acknowledge the intent.

RM-1978:
@Jerryescandon 

- I'm curious about the placement of the disclaimer. Would appearing after all the form fields for user-entered data be more meaningful as a visual logic?
- I wonder if we can limit the disclaimer copy to: "I agree to share this information with verified Government of Canada managers and HR advisors.

Jerryescandon:
- **We normally share everything with managers and HR, so why is this the only form that has this disclaimer?**

Basically, because the long text field and the nature of "Personal experience" allow way more flexibility for users to share more than the usual hard data. (dates, titles, etc). 

This long text field is designed to not be prescriptive, we dont include any guidance for it. Other forms have specific "guidance" on what info to submit, for example "how did you apply javascript?" we attempt to guide users on how to submit certain type of info. So it is not really necessary to add any disclaimers, since the questions and the guidance are not personal in nature. 

This checkbox acts as both a disclaimer that these descriptions will be read by multiple people on the hiring team. As well as a reminder to **"keep-it-PG13"** if you will.

If this form had only the hard dates, there would be no need for this disclaimer. 

I agree we are only interested in reinforcing this checkbox ONLY if the text field info has changed, nothing else should really affect the disclaimer checkbox.

Jerryescandon:
> * I wonder if we can limit the disclaimer copy to: "I agree to share this information with verified Government of Canada managers and HR advisors.

Yeah copy seems a bit unnecessary over description, could benefit from some editing.

esizer:
> I agree we are only interested in reinforcing this checkbox ONLY if the text field info has changed, nothing else should really affect the disclaimer checkbox.

Thanks @Jerryescandon , @substrae and @brindasasi . I have went and made the change so that the disclaimer only unchecks itself when the description changes and nothing else.

[simplify by only watching description](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7748/commits/77f3bcf325d3d151f3260780fc32e31987c89a52)

RM-1978:
The Ui has changed signficantly since this issue was created

RM-1978:
- let[s update the disclaimer copy to: "I agree to share this **experience description** with verified Government of Canada managers and HR advisors."

RM-1978:
Great , I will add the text revisions to the Copy work

brindasasi (commented):
As per the recording above checkbox clears on edit page load. It should only be clearing when the description changes. 
If we don't store this field in DB, I think we need to set the disclaimer's default value as true if the description is not null.

esizer:
> It should clear only when the description changes , not on the page load.

Thanks for catching that. I added a conditional to check it against the default values _before_ unchecking it. 😅 

[check description before unchecking disclaimer](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7748/commits/82bced0c8a3e27c1527250515e1be04183f3db98)

brindasasi (approved):
awesome! works fine to me 🥇

PR #7760, opened 2023-08-29
[Fix] Reduce Laratrust cache time

### LINKED ISSUE #7759: ✨ Reduce Laratrust cache time
## ✨ Feature

Reduce Laratrust cache expiration time from 1 hour to 60 seconds. Also add `LARATRUST_ENABLE_CACHE=true` `api/.env.example` to encourage cacheing in local environments.

Also, add `@can` permission check to PoolCanadidate.user in schema, replacing authorizedToView scope, which was used mainly because laratrust was too slow without caching.

## 🕵️ Details

Reducing the cache time will be helpful on production because it will avoid a situation where we give admin users a new role, and they are confused about not being able to use it right away. One more database query every minute should cause a minimal performance hit.

Edit `api/config/laratrust.php` and `api/.env.example`.

## ✅ Acceptance Criteria

- [ ] Laratrust cache times out after a minute, on local and on production.
- [ ] When copying .env.example into a local environment, laratrust cache is enabled by default.
- [ ] `PoolCandidate.user` uses `can` permission check instead of `authorizedToView` scope.

tristan-orourke:
🤖 Resolves #7759 

## 👋 Introduction

Reducing the cache time will be helpful on production because it will avoid a situation where we give admin users a new role, and they are confused about not being able to use it right away. One more database query every minute should cause a minimal performance hit.

Also, changing PoolCandidate.user to use `@can` directive instead of the `authorizedToView` scope makes it more consistent with other queries.

## 🕵️ Details

**NOTE:** PoolCandidate.user is still fairly slow for PoolManagers, but it can complete for ~100 candidates without timing out. And it seems to be slow using the scope as well, so I'm switching to `@can` for consistency. I went down a short little rabbit hole trying to improve the performance, but that's out of scope for now.

### Testing prep
1. Add `LARATRUST_ENABLE_CACHE=true` to your local `api/.env` file.

### Tests
1. Log in as request@test.com.
2. Go to a admin site, requests table, and select a request.
3. Clear filters, allow table to load. Ensure request doesn't fail or take overly long.
4. Log out
5. Log in as pool@test.com
6. Go to admin site, go to pool candidates table for a pool you own (must be a pool with many applicants)
7. Ensure table loads without error and doesn't take overly long.
8. Log out
9. Run a query for poolCandidates.user where the user is owned by the authenticated user. Confirm the user appears in the results instead of null


esizer:
Everything seems to be working within the scope of the issue.  Except, I may have found a new bug.

The pool candidates paginated query crashes the application when it contains a candidate the user cannot view. I _think_ this is because `@can` returns an error rather than just removing (scoping) the item from the list🤔 

If there is no difference in runtime, would it be safer to use scopes on queries that return arrays rather than `@can`? Or should we just be protecting against possible auth errors when making the queries?


https://github.com/GCTC-NTGC/gc-digital-talent/assets/4127998/91307a53-7052-4273-b6af-ac8293a4ca3e

esizer (approved):
Awesome! 😄 

I found a bug but it is on main as well so I noted it and moved on 👍

tristan-orourke:
@esizer you make a great point, here's a new ticket for it: https://github.com/GCTC-NTGC/gc-digital-talent/issues/7761

PR #7815, opened 2023-09-05
[Fix] Prevent snapshot missing

### LINKED ISSUE #7217: 🐛 Snapshot not found
## 🐛 Bug

DCM has reported several instances in which they view a candidate in an ongoing recruitment pool but then gets a "Snapshot not found" error.

- The applicant has their status as "New"
- We have asked around and there seems to be **NO ONE** has placed this candidate directly into this pool.
- This is an ongoing recruitment pool
- We assume that this candidate applied using the application process.
- We assume they applied and no snapshot got generated, but they got placed into the pool.

To whoever picks up this ticket, ask jerbo for the name of the user.

## 🦋 Expected Behaviour

Snapshot should get generated and tied to every user that applies to a pool (Ongoing recruitment pools included)

## 🕵️ Details

MicrosoftTeams-image (3)

## 📋 Steps to Reproduce

Find examples in prod in this message:
https://talent-cloud.slack.com/archives/C0259G6KXJ6/p1690384537785269

## ✅ Acceptance Criteria

- [ ] Submission should fail if snapshotting fails. 
- [x] Figure out how to reproduce the issue
- [ ] Close the gap
- [ ] If unable to figure out what's wrong, implement a way to figure it out (extra logging, etc).

petertgiles:
🤖 Resolves #7217 

## 👋 Introduction

This branch prevents applications from being submitted without a snapshot.

## 🕵️ Details

The snapshots were being generated in an asynchronous event handler.  If something went wrong with the snapshot the error was not preventing the application from being submitted.  I've moved the snapshot logic to be a synchronous part of the submission process like the submission date property so that a failure will cause the submission to fail.  

I've also switched the start date properties to use null-safe method calls to avoid having the snapshots error out.  This what was causing the problem with the specific example user.  I'm not sure how they managed to save an experience without start dates but the snapshot is probably not the place to validate that.  This should prevent (or make it very unlikely that) the snapshot will fail and show the toast error message.

## 🧪 Testing

- [ ] Can complete a regular application submission .  Submitted with submission date, status, and snapshot.
- [ ] Can edit the database to remove an experience start date and still complete a submission. Submitted with submission date, status, and snapshot.
- [ ]  Undo the null-safe methods and now can't complete a submission when an exception happens in snapshot. No submission date, status, or snapshot.

## 📸 Screenshot

image


esizer (approved):
Awesome! ⭐ Works well 👍

PR #7824, opened 2023-09-05
[Fix] Pool, non-essential skill relationship typo

### LINKED ISSUE #7680: 🐛 Typo in `Skill` `poolsNonEssentialSkills` relationship method
## 🐛 Bug

When accessing the pools inwhich a skill is marked as an non-essential skill, we are returning the wrong values based on a minor typo.

## 🦋 Expected Behaviour

We should return pools in which the skill is marked as non-essential and not essential.

## 🗺️ Location

https://github.com/GCTC-NTGC/gc-digital-talent/blob/d989f6c340f69676824e275835f40cf545a301ba/api/app/Models/Skill.php#L101-L109

## 🙋‍♀️ Proposed Solution

Update the relationship methdo to return:

```php
return $this->belongsToMany(Pool::class, 'pools_nonessential_skills');
```

Or, we could just remove both methods. I'm not sure if we ever use that side of the relationship 🤔 

## ✅ Acceptance Criteria

- [ ] The `Skills -> Pool non-essential skills` returns the correct pools

esizer:
🤖 Resolves #7680 

## 👋 Introduction

Fixes a minor typo in the inverse relationship for pool nonessential skills.

## 🧪 Testing

> **Note**
> I'm not sure how to best test this since we aren't using the relationship yet 🤔

### DIFF
--- a/api/app/Models/Skill.php
+++ b/api/app/Models/Skill.php
@@ -66,7 +66,7 @@ class Skill extends Model
 
     public function poolsNonessentialSkills(): BelongsToMany
     {
-        return $this->belongsToMany(Pool::class, 'pools_essential_skills');
+        return $this->belongsToMany(Pool::class, 'pools_nonessential_skills');
     }
 
     public function getDetailsAttribute()


vd1992 (approved):
Works for me!

image

image

image

Issue #7752, opened 2023-08-28
♻️ Polish New Table Components


petertgiles:
## ♻️ Debt/Refactor

We were rushing a bit to merge [the new table components](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7588).  Now that they're in and the team can start building the dependent features we should go back and polish the base components.

Chromatic link: https://www.chromatic.com/component?appId=61e099a1bb1465003a73d3ce&csfId=tables-responsive-table&buildNumber=5556&k=64ecaeb96e6fc4f8630f355e-1200px-interactive-true&h=7&b=-1

## ✅ Acceptance Criteria

- [x] Update pagination (discussion: https://github.com/GCTC-NTGC/gc-digital-talent/pull/7588#discussion_r1302054232 solution: https://github.com/GCTC-NTGC/gc-digital-talent/issues/7752#issuecomment-1697695961)
- [x] [design review items](https://github.com/GCTC-NTGC/gc-digital-talent/issues/7752#issuecomment-1710190388)
  - [x] Font size on this checkbox group is different from our other checkbox inputs
  - [x] Font size on the "Show" select element should match the label next to it, and not be bolded
  - [x] change the label to `Filter by keyword`, and I'm not sure we need the icon in this context
  - [x] Column header bolding doesn't match the design - no need to bold unless an item is selected for sorting, in which case the underline should disappear
  - [x] The footer element's content should align with the first column's content
  - [x] Mobile row padding is a bit tight, designs had it set to `x1` by default, but if this is causing issues on smaller screens, we can reduce it before it increases to `x1` for larger mobile devices
  - [x] Bolding used in mobile content is using `800` instead of `700`
  - [x] Footer elements should centre on mobile
  - [x] The ability to select all objects is missing on mobile
  - [x] Padding between the two pagination rows is a bit tight compared to the design - I'm a bit worried about tap targets here
- [ ] [Mobile view has a "Sort by value" function](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7831#discussion_r1317880531)
- [ ] mobile viewport story for Chomatic to snapshot

vd1992:
@substrae 
There was also one other question that needs resolution please
https://github.com/GCTC-NTGC/gc-digital-talent/pull/7588#discussion_r1302098709

substrae:
@petertgiles @vd1992 

I think this is good to go based on Vachan's edit! 🥳

substrae:
[...]

PR #7831, opened 2023-09-06
[Feature] Skill library page

### LINKED ISSUE #6936: 📃 Create the "Skill library" page
image

<details><summary>📸 Full page (Empty)</summary>

</details>

<details><summary>📸 Full page (Complete)</summary>

</details>

# ✨ Details ([Figma file](https://www.figma.com/file/FhRhwczL8JI6JC8eVFlWOm/Candidates---Skill-library?type=design&node-id=141-19757&mode=design&t=IWy4rRdaWNUXQXLc-0))

Adds brand new functionality that allows users to link a skill directly to their profile without the need to link experiences. This is the first step in the addition of the skill showcase functionality. Subsequent pages will include a way to view a single skill (clicking its name in the tables), organize skills into showcases, and a way to manage those showcases.

# 🗃️ General requirements

```[tasklist]
## ⚙️ The skill table
- [ ] Ensure the table null state exists
- [ ] Ensure table headings are appropriately named and sort functionality is available if indicated
- [ ] Ensure font sizes and colors match
- [ ] Ensure skill names are clickable in preparation for the eventual skill view
- [ ] Ensure the alternate mobile table matches the design
- [ ] Allow users to add skills using the skill browser dialog
- [ ] Skill table has three columns: skill name, number of associated experiences, and skill level (which may be missing)
- [ ] The skill name in the table is a link to Individual Skill Page (#7378)
``` 

**Please note:** 
- The tables on this page leverage the table component originally described in #6932
- The placeholder example word in the search input changes depending on whether you're looking at the behavioural or technical skill table.

<details><summary>📸 Skill table - No skills</summary>

</details> 

<details><summary>📸 Skill table - With skills</summary>

</details> 

<details><summary>📸 Skill table - On mobile devices</summary>

</details>

### The skill browser dialog

When clicking "Add a new skill", the skill browser dialog should be triggered with the skill list limited to the correct category context the button was pressed (i.e. if they add a new skill to behavioural skills, only show behavioural skills in the list).

See issue #7394 for more details.

<details><summary>📸 Skill browser dialog</summary>

</details>

# 🗃️ Page section requirements

```[tasklist]
## 📚 Behavioural skills
- [ ] Ensure copy matches
- [ ] Ensure iconography is the correct size and color
- [ ] Ensure the skill table search placeholder matches
``` 

<details><summary>📸 Behavioural skills - No skills</summary>

</details>

<details><summary>📸 Behavioural skills - With skills</summary>

</details>

```[tasklist]
## 📚 Technical skills
- [ ] Ensure copy matches
- [ ] Ensure iconography is the correct size and color
- [ ] Ensure the skill table search placeholder matches
``` 

<details><summary>📸 Technical skills - No skills</summary>

</details>

<details><summary>📸 Technical skills - With skills</summary>

</details>

# 🗃️ State and other requirements

```[tasklist]
## 🔄️ When skills are missing a claimed level
- [ ] Ensure that skills can be added without a level through the application (and eventually showcase)
- [ ] Ensure the table indicates when a skill is missing
``` 

This state occurs when a candidate adds a skill to their profile through an application or the skill showcase. In those contexts, we don't ask them to specify their skill level as it would confuse the task at hand. When these skills are added without a level, the relevant table column should convert to a "Missing" link that takes the user to the skill view to update their level.

<details><summary>📸 Missing skill level</summary>

</details>

## Acceptance Criteria

- [ ] routing for page
- [ ] all of above criteria

esizer:
🤖 Resolves #6936 

## 👋 Introduction

Creates the skill library page for adding skills to a user profile.

## 🕵️ Details

There is some competition for the URL state on this. If you click one of the library links on the dashboard, you can see the category is added to the hash of the URL state so we can scroll to it. However, the tables search params overwrites that hash instead of adding onto it. I'm wondering if anyone can determine why it is overwriting instead of appending.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to `/applicant/profile-and-applications/skills`
3. Confirm look and feel match [the mockup](https://www.figma.com/file/FhRhwczL8JI6JC8eVFlWOm/Candidates---Skill-library?type=design&node-id=141-19754&mode=design&t=11EksoNlvAqLPVgX-0)
4. Confirm adding a skill works as expected
5. Confirm the tables function as expected

## 📸 Screenshot

localhost_8000_en_applicant_profile-and-applications_skills_cv= ps=10 p=1


apps/web/src/components/SkillDialog/SkillDialog.tsx
vd1992, line 1: Got an old typo here

image

"link one" instead

image

apps/web/src/components/Table/ResponsiveTable/Table.tsx
petertgiles, line 1: Out of scope for this PR, but what do you thing about pinning the size of the table somewhow?  I find the layout reflow when moving to the last page quite disorienting.

https://github.com/GCTC-NTGC/gc-digital-talent/assets/8978655/c5f00aba-7b6d-436c-9621-8fe40d3c9fef

apps/web/src/components/SkillDialog/SkillDialog.tsx
vd1992, line 1: Seems like there has been some copy drift since the dialog was first added and the issue now. Not sure how to best tackle this sort of thing 😕 Leave it to design review stage?

Current dialog

image

Dialog as seen in issue screenshot

image

apps/web/src/pages/Skills/SkillLibraryPage.tsx
petertgiles, line 1: It seems like adding a skill works but the table doesn't automatically update.
image
If I manually refresh it appears.  Maybe only happening with Behavioural?

petertgiles, line 1: This might be a backend bug and out of scope.  If I add a skill, delete it, then readd it at a different level, it appears in the list with the original level, not the latest one.

apps/web/src/components/SkillDialog/SkillDialog.tsx
esizer: The dialog was built a while ago and has went through different iterations since the design. We are using it else where and it was approved in its current state.

I wouldn't worry about how exact all the copy is. You are correct to let the design review to run its course to find small discrepancies like this.

I'd focus on new components and their functionality if you are reviewing. Though, if you find actual issues that need to be addressed with the dialog we should create new issues since the dialog is not specific to this PR.

vd1992: Sounds good!
Does adding proper toast messages to connect to API error keys fit in this issue or is it out of scope?

image

apps/web/src/components/Router.tsx
```diff
@@ -626,6 +626,14 @@ const UpdateSkillPage = React.lazy(() =>
       ),
   ),
 );
+const SkillLibraryPage = React.lazy(() =>
```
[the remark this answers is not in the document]
esizer: Uhh, that is a good question. We actually have two possible places that link could go and there was no acceptance criteria about that. 🤔 

@substrae Is this link for the library or the showcase?

apps/web/src/components/Table/ResponsiveTable/Table.tsx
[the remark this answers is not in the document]
esizer: Thanks, I didn't even notice that! I couldn't figure out why it was happening 😢

esizer: Yeah, I am also not a huge fan of layout shift. I think the trouble with this is that we need to know some crucial information to do this that would be difficult to determine.

- How many rows? - Easy because of page size
- Average row height - Hard because rows are not uniform

apps/web/src/pages/Skills/SkillLibraryPage.tsx
esizer: Does it change the database value? I'm wondering if it is stale `urql` cache or a bug? 🤔

apps/web/src/components/Router.tsx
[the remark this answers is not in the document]
vd1992: Will need to be fixed definitely, it is one link right now

image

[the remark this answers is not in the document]
petertgiles: Yeah, I remember building this hero and not having the skill link.  The solution was to combine them until the page existed.  Is this the page existing now?

apps/web/src/components/Table/ResponsiveTable/Table.tsx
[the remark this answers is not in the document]
vd1992: There is some more URL oddness afoot. The URL flickers when sorting. 
It also does the flicker when toggling sort by experience count too but capture compressed it to invisibility 

gif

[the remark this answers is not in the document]
esizer: Yeah, I mentioned that to be resolved. It is because you actually can't have two tables on thew same page since we sync state in the URL. So editing it updates both tables at once.

Unfortunately, it was never mentioned we would be having two tables on the same page when we were working on the new table component.

apps/web/src/pages/Skills/SkillLibraryPage.tsx
vd1992: Appears to update for me after adding

https://recordit.co/dGODPjsryR

esizer: It will if you have items, this is a known `urql` thing I just missed with `additionalTypenames`.

apps/web/src/components/Table/ResponsiveTable/Table.tsx
petertgiles: Opened issue:
https://github.com/GCTC-NTGC/gc-digital-talent/issues/7836

apps/web/src/pages/Skills/SkillLibraryPage.tsx
vd1992: Intentional logic, it restores soft-deleted if applicable

image

vd1992: From issue 

image

esizer: That seems wrong, when you re-add a skill you set the level at the same time (required) so it should use that rather than the previous level. 😕 

If that is a backend thing, I think that is a bug.

apps/web/src/components/Router.tsx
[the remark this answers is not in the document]
esizer: I guess it depends on that "this page" is. We have Skill Library and Skill Showcase, I'm not sure which one we should point this to since it just says "Skills" even though both pages show skills on them.

apps/web/src/components/Table/ResponsiveTable/Table.tsx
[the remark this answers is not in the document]
vd1992: Hm, what would be the best solution in your opinion then? No table params in the URL? At least, for the time being (could revisit later)

apps/web/src/components/SkillDialog/SkillDialog.tsx
esizer: It was not part of the AC and I was never supplied a message. If we can get whoever is in charge of copy to supply an error message it should be easy to add.

apps/web/src/pages/Skills/SkillLibraryPage.tsx
esizer: Fixed in [fix urql cache with no items](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7831/commits/79427249f019414070b60b36236e2387abd95273)

apps/web/src/components/Table/ResponsiveTable/Table.tsx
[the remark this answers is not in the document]
esizer: So, as a workaround I disabled the URL sync. I'm not sure what the real fix is we want to sync with multiple tables.

[add table state sync workaround](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7831/commits/ae223cb1e12b6a78c9d91115efde23ac92316e96)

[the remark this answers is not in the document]
esizer: I was able to fix the back button, it was because we weren't replacing the existing history so it was forcing you to back through all state changes. [fix back button on table search param](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7831/commits/a9e9ca09894bf475f684ab517c011a6e3eb177b1)

Also, it seems like the overwrite of a hash is intentional. I'm not sure I agree with their reasoning but they explained it on an issue someone submitted.

See: https://github.com/remix-run/react-router/issues/8393#issuecomment-979676684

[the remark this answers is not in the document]
vd1992: image

apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
vd1992, line 1: Mobile view has a "Sort by value" function in designs 
That would appear to be adding functionality to the ResponsiveTable and not strictly applicable to Skill Library right? If so, spinoff issue or should I not worry about it?

image

apps/web/src/components/SkillDialog/SkillDialog.tsx
vd1992: Sure 
@substrae 
For the above error toast, "DuplicateUserSkill", observed when attempting to claim the same skill twice, what would you like the message displayed to the user to read?

substrae: @vd1992 @esizer 

- To Vachan's point, the dialog input labels should match the design; the labels that are there now are incorrect
- I'm assuming `DuplicateUserSkill` applies in the scenario where you try to add a skill you already have? Should we not show this earlier (i.e. in the dialog) to prevent the behaviour in the first place? We could remove skills already applied to the contextual UI, or we could prevent the dialog from submitting with an error in situ that says something along the lines of "You've already added this skill to your ______."

apps/web/src/components/Router.tsx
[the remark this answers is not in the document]
substrae: @esizer 

The subtitle links to the library.

apps/web/src/components/Table/ResponsiveTable/Table.tsx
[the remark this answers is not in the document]
petertgiles: Nice, this works well for me, now.

apps/web/src/pages/Skills/SkillLibraryPage.tsx
petertgiles: Nice one.  :dart:

apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
esizer: Hmm, this is getting tricky since we moved on with a less than complete table since it was taking too long. Its kind of at the point where we need to decide when to do the work. I'm not sure if this PR is the correct place or if we should catch it in #7752 🤔

apps/web/src/components/SkillDialog/SkillDialog.tsx
substrae: @esizer 

Hmm, do you happen to have a link to the ticket that requested those changes?

Gotcha - then yes, a generic error is enough here.

esizer: Yeah, im looking for it right now but we had so many changes made over the past 10 months it is hard to pin point when they happened 😅 

In the meantime, these changes happened a while ago to say the least and were approved and the dialog is already in use on the site. I'm not sure that falls under the scope of this PR. Could we get another ticket to resolve any copy/design issues with the skill dialog to match what it is actually supposed to say/do? I feel like it has went through so many iterations there may have been small differences that were never caught and now it is out of sync with the current design 😢

apps/web/src/components/Router.tsx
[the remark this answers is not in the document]
esizer: Thanks, updated as part of [update dashboard hero skills link](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7831/commits/5b9f31eaa71d47c439c58b7a99d61c2421093754)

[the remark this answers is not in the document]
petertgiles: Careful, you've got two a3's.
image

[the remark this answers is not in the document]
esizer: 🤦‍♀️ Thankfully, you have working eyes 👀 

Fixed in [fix final link in hero](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7831/commits/b55a6ab125b4ea8285efc7a2fe0a2fa6fe1774ce)

apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
petertgiles: I vote for #7752

apps/web/src/components/SkillDialog/SkillDialog.tsx
substrae: Yep, I'll pop something up today!

petertgiles (approved):
I think all my review items are resolved.  Nice work!  :medal_sports:

PR #7873, opened 2023-09-07
[Fix] Custom validation for First Nations validation

### LINKED ISSUE #7360: 🐛 Non-status and status First Nations values both true
## 🐛 Bug

Both "I am Status First Nations" and "I am Non-Status First Nations" First Nations values are able to be saved in the profile.

## 🦋 Expected Behaviour

If one option is true, the other cannot be.

## 📋 Steps to Reproduce

1. Navigate to profile, Diversity, equity, and inclusion section
2. Click Edit this information for I affirm that I am First Nations (status or non-status), Inuk (Inuit), or a Métis person.
3. Check "I am Status First Nations" and "I am Non-Status First Nations" checkboxes for First Nations field
4. Submit form
5. Observe both values saved

## 📹 Screen recording

https://github.com/GCTC-NTGC/gc-digital-talent/assets/3046459/9b34018c-048f-465f-87a0-dc9ecfe3409a

## 🙋‍♀️ Proposed Solution

If one checkbox option is checked, the other cannot be.

## ✅ Acceptance Criteria

- [ ] both "I am Status First Nations" and "I am Non-Status First Nations" First Nations values cannot be saved simultaneously
- [ ] create regression test to avoid behaviour in future
- [ ] Error message from Gray: "Please select either Status First Nations or Non-Status First Nations."

petertgiles:
🤖 Resolves #7360 

## 👋 Introduction

This branch adds the new message for API validation and also adds custom for validation.

## 🕵️ Details

This is a bit of a hack, for sure.  The form treats the four different checklist groups as a single element so a regular validation rule causes all the groups to go into a visual error state with the message.  It also doesn't have the submit handler in the component so you can't hook in there.  I tried breaking the form up into separate elements (which is probably the best solution) but it quickly got out of hand with the chips and illustrations and broke everything.  I opted for an independent validation rule that only brings up one standalone validation message.  It looks better but behaves differently in that it doesn't focus the input element and validates onChange.

## 🧪 Testing

1. Rule prevents submission if both First Nations boxes are checked.
2. Can submit normally if one or neither boxes are checked.
3. Works on profile inline form IAP submission form.

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/pages/Applications/ApplicationSelfDeclarationPage/SelfDeclaration/CommunitySelection.tsx
+++ b/apps/web/src/pages/Applications/ApplicationSelfDeclarationPage/SelfDeclaration/CommunitySelection.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import { useFormContext } from "react-hook-form";
 import { useIntl } from "react-intl";
 
-import { FieldLabels, Checklist } from "@gc-digital-talent/forms";
+import { FieldLabels, Checklist, Field } from "@gc-digital-talent/forms";
 import { Alert, Chip, Chips } from "@gc-digital-talent/ui";
 
 import HelpLink from "./HelpLink";
@@ -18,7 +18,8 @@ export const CommunityList = ({ labels }: CommunityListProps) => {
   const [isAlertOpen, setIsAlertOpen] = React.useState<boolean>(false);
   const [hasDismissedAlert, setHasDismissedAlert] =
     React.useState<boolean>(false);
-  const { watch, setValue } = useFormContext();
+  const { watch, setValue, setError, clearErrors, formState } =
+    useFormContext();
 
   const communityLabels = getCommunityLabels(intl);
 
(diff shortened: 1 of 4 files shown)


apps/web/src/pages/Applications/ApplicationSelfDeclarationPage/SelfDeclaration/CommunitySelection.tsx
```diff
@@ -73,7 +95,13 @@ export const CommunityList = ({ labels }: CommunityListProps) => {
                 }),
               },
             ]}
+            aria-describedby={customAlertId}
           />
+          {formState.errors.firstNationsCustom && (
+            <Field.Error id={customAlertId}>
```
esizer, line 101: Could we add some margin here? 

```suggestion
            <Field.Error id={customAlertId} data-h2-margin-top="base(x.25)">
```

esizer (changes requested):
Great! One minor styling suggestion.

esizer (approved):

PR #7875, opened 2023-09-07
[Tooling] Add `/api` folder to lint workflow

### LINKED ISSUE #7871: 🛠️ Add `/api` to the lint workflow paths
## 🛠️ Tooling

We would like to run the linter on PRs with changes to the api.

## 🕵️ Details

We recently added a PHP linter (Pint) but are only running the workflow on changes to the frontend which is allowing some issues to be missed until they get into main.

## 🙋‍♀️ Proposed Solution

Add `/api` to the paths attribute in the workflow.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/ae223cb56a37ca82ebdf9f83d1001326329e9e6d/.github/workflows/lint.yml#L4-L14

## ✅ Acceptance Criteria

- [ ] Paths updated on lint workflow
- [ ] Lint workflow runs on changes to PHP files

esizer:
🤖 Resolves #7871 

## 👋 Introduction

Adds the `/api` folder to the triggers for the lint workflow.

## 🕵️ Details

Now that we have a PHP linter, we should be running it in CI when changes are made to PHP files.

## 🧪 Testing

> **Note**
> I'm not sure how we would test this since it also runs on changes to that file (which has happened in this PR).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,11 +7,13 @@ on:
       - .github/workflows/*lint*.yml
       - apps/**
       - packages/**
+      - api/**
   pull_request:
     paths:
       - .github/workflows/*lint*.yml
       - apps/**
       - packages/**
+      - api/**
   merge_group:
     branches: [main]
 jobs:


petertgiles (approved):
> I'm not sure how we would test this since it also runs on changes to that file (which has happened in this PR).

I agree.  LGTM.  :shrug:

PR #7876, opened 2023-09-07
[Fix] Hide "removed" applications from applicants

### LINKED ISSUE #7441: Removed status not intended to be shown to applicants
## Description

The purpose of the "REMOVED" status is intended to be a cleanup tool for "retrofit-recruitments".

Applicants whose status has been placed as removed are meant to signify that they have now properly created a real account and placed on a pool accordingly. OR move users from old generic "catch-all" recruitments into more specific recruitments. 

DCM were only meant to tag these users as "REMOVED" to prevent them from showing up twice in request results as they were being properly transitioned from fake accounts and pools and into real accounts and pools.

Meaning "REMOVED" was never meant to be a status that was intended to be shown on the applicant facing side, since it's meant to remove duplicates.

## To-do

Ideally removed status should actually prevent any card from appearing on the applicant facing side. Applicants tagged as removed should all ideally been placed into another more appropriate pool.

We should HIDE all cards tagged as removed on the applicant facing side.

## Screenshots

image

## Acceptance Criteria
- [ ] Profile and applications checked
- [ ] Experiences page checked
- [ ] Profile page checked

esizer:
🤖 Resolves #7441

## 👋 Introduction

This ensures users do not see applications with the "removed" status.

## 🧪 Testing

1. Build `npm run dev`
2. Create an application and set the status to "REMOVED"
3. Navigate to the following pages and ensure it does not appears
    - `/users/{userIs}/profile`, in the "Hide recruitment availability" accordion
    - `/applicant/profile-and-applications`, both current and past accordions
    - `/users/{userId}/profile/career-timeline-and-recruitment`, qualified recruitments section

### DIFF
--- a/apps/web/src/pages/ProfileAndApplicationsPage/components/TrackApplications/TrackApplications.tsx
+++ b/apps/web/src/pages/ProfileAndApplicationsPage/components/TrackApplications/TrackApplications.tsx
@@ -15,7 +15,7 @@ import {
 
 import { PoolCandidate, Scalars } from "~/api/generated";
 import useRoutes from "~/hooks/useRoutes";
-import { isApplicationInProgress } from "~/utils/applicationUtils";
+import { isApplicationInProgress, notRemoved } from "~/utils/applicationUtils";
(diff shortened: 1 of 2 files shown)


petertgiles (approved):
Noice.

PR #7878, opened 2023-09-07
[Fix] Respect reduced motion

esizer:
🤖 Resolves #7878

## 👋 Introduction

Adds config for reduced motion for `framer-motion` animations.

## 🕵️ Details

This allows us to reduce the amount of motion happing when a user requests it. It also assists in pausing said animations when running chromatic so they do not create false positive diffs.

> **Note**
> I'm wondering if a user requests reduced motion if we should completely stop the spinner or play a different more subtle animation that still respected reduced motion but does not trigger a diff in chromatic 🤔 

## 🧪 Testing

1. Set your accessibility settings to request reduced motion
2. Build `npm run dev`
3. Confirm spinner is static and does not animate
4. Login to admin and confirm menu is instant instead of animated


package-lock.json
```diff
@@ -27135,6 +27136,7 @@
         "@storybook/react": "^7.3.2",
         "@storybook/react-webpack5": "^7.3.2",
         "chromatic": "^7.1.0",
+        "framer-motion": "*",
```
petertgiles, line 27139: Did you intend to keep this one floating?

packages/ui/src/components/Loading/Loading.tsx
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import { motion } from "framer-motion";
+import { motion, useReducedMotion } from "framer-motion";
```
petertgiles, line 2: This was the big one for Chromatic.  But we do have motion in use in some other components like the accordion and repeater.  Worth updating those, too?

packages/ui/src/components/Loading/Loading.tsx
```diff
@@ -75,17 +76,18 @@ const Loading = ({
...
-              },
-              transition: {
-                ease: "linear",
-                duration: 1,
-                repeat: Infinity,
-                repeatDelay: 0,
-              },
-            })}
+            {...(!pause &&
+              !shouldReduceMotion && {
```
petertgiles, line 80: I like this.  If someone has requested reduced motion they probably appreciate having no spinner animation.  I think it's recognizable in context even when not spinning.

petertgiles (commented):
This is so cool!  I've never tried the reduced motion setting before.  A few questions...

package-lock.json
esizer: Definitely not 😅 Thanks for catching this! Fixed in [regenerate lock file](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7878/commits/a5f3af468a4ceca1f87cc3cfd900866c90cbcfd6)

packages/ui/src/components/Loading/Loading.tsx
esizer: OH right! I forgot about those as well as the repeater. We also have some opacity based animations but those should be fine since opacity isnt motion 👍 

[add reduced motion to remaining components](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7878/commits/0b725260555412f2e3d1fc443f9a8cbaab64bf52)

packages/ui/src/components/Loading/Loading.tsx
esizer: Sounds good, thanks for responding to that note🙏

petertgiles (approved):
Very cool!  :ice_cube:

PR #7880, opened 2023-09-07
[Feature] Allow filtering pool candidates by gov employee field

### LINKED ISSUE #7816: Add filter to request table - Employee of Government of canada
## ✨ Feature

As a Request Responder, I want to filter candidates by "Government Employee" so I can more easily fulfill requests from managers that only want to see internal candidates.

## 🕵️ Details

Add a "Government Employee" filter to the Candidate Results table (on Request page).

The "Government Employee" filter exists on our Users table, but is not currently on the Candidate Results table.

### 🧑‍🎨 Designer

Gray created this issue, but I'll ask Josh to review.

@substrae - This seems fine as is. Slot the filter in the same place for consistency, but otherwise the styles will reuse existing components.

## 📸 Screenshot

This is what the "Government employee" filter looks like on the USERS table.

Image

Implement in a similar way to the users table, but on the Request table!

## ✅ Acceptance Criteria

- [ ] "Government employee" filter added to Candidate Results table

vd1992:
🤖 Resolves #7816 

## 👋 Introduction

Enable filtering pool candidates by government employee field values, this applies to all instances utilizing the pool candidates table. 

## 🕵️ Details

Scope, schema tweak, tested, and then connected the pieces in the frontend. 

## 🧪 Testing

1. Test you can use the filter, go to a request and open the filter dialog
2. Test the filtering works, see below for an example

## 📸 Screenshot

Get values through your preferred method, here is the total candidate count excluding draft, and candidate count that are government employees. In this case, 49 and 18 respectively. 

image

Observe this lines up with filtering for pool candidates in that pool. Matches the numbers from above

image

Filter state for just government employee, everything else cleared aside from the needed pool selection

image

### DIFF
--- a/apps/web/src/components/PoolCandidatesTable/PoolCandidateTableFilterDialog.tsx
+++ b/apps/web/src/components/PoolCandidatesTable/PoolCandidateTableFilterDialog.tsx
@@ -33,6 +33,7 @@ export type FormValues = {
   expiryStatus: Option["value"][];
   suspendedStatus: Option["value"][];
   publishingGroups: Option["value"][];
+  govEmployee: Option["value"][];
 };
 
 const Footer = () => {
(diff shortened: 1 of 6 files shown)


apps/web/src/components/PoolCandidatesTable/PoolCandidateTableFilterDialog.tsx
```diff
@@ -256,6 +257,30 @@ const PoolCandidateTableFilterDialog = ({
...
+                  label={formatMessage({
+                    defaultMessage: "Skill Filter",
+                    id: "GGaxMx",
+                  })}
+                  options={optionsData.skills}
+                  isLoading={rawGraphqlResults.skills.fetching}
+                />
+              </div>
+              <div data-h2-flex-item="base(1of1) p-tablet(1of2) laptop(1of3)">
+                <MultiSelectFieldBase
```
vd1992, line 273: The issue mentioned to maintain placement consistency between filter dialogs. I wasn't 100% on this so I elected to match its filter buddy, move it so it next to Skill Filter.

vd1992: I was iffy on it as the dialogs do look quite different 

User filter dialog 

image

Pool candidates filter dialog with the changes

image

esizer (approved):

PR #7881, opened 2023-09-08
[New] Rich text editor form input

### LINKED ISSUE #7766: ✨ Markdown Editor in Storybook - for job poster
## ✨ Feature

On the job poster users are inputting lists and links, but they are not marked up properly. Instead the are just plain text which is a textbook accessibility fail.

## 🕵️ Details

This issue is to implement a markdown editor in storybook. Once we have it there we will pass to design to decide how we want to present the component.

Eric has used [TipTap](https://tiptap.dev/) in the past to good effect. (it's light weight and let's you add things in as you need them)

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Markdown editor working in Storybook with a save button that then presents the text back as it would appear (or similar)
- [ ] Markdown lists and links only
- [ ] TipTap is a suggestion but not the only option

esizer:
🤖 Resolves #7766  

## 👋 Introduction

Adds a new component for rich text content.

## 🕵️ Details

This is a story for us to test and find bugs before implementing it. Right now, we only allow:

- Unordered lists
- Links

> **Note**
> We can create a custom extensions as well so in the future we could add any of our custom components if we feel the need.

## 🔨 Usage

There is some examples in the story but this should use used sort of like the `Textarea` component minus some props. When submitting the form, this will emit a string of HTML.

> **Warning**
> Like all of our other inputs, this must be used within a `react-hook-form`.

### Input

```tsx
import { RichTextInput } from "@gc-digital-talent/forms";

<FormProvider>
   <RichTextInput name="description" id="description" label="Description" />
</FormProvider>
```

### Display

Since this gives use very basic HTML, a renderer has been created to apply some styles and more importantly, render our custom link component.

```tsx
import { RichTextRenderer, htmlToRichTextJSON } from "@gc-digital-talent/forms";

const node = htmlToRichTextJSON(html); // html represents the HTML we get back from the API

return (
  <RichTextRenderer node={node} />
);
```

## 🧪 Testing

1. Start storybook `npm run storybook`
2. Navigate to the "Rich text" story
3. Interact with the examples and confirm it works as expected

## 📸 Screenshot
Screenshot 2023-09-08 134405
Screenshot 2023-09-08 134419


packages/forms/src/components/RichTextInput/MenuButton.tsx
vd1992, line 1: Getting a console error due to refs off this file I believe 

image

packages/forms/src/components/RichTextInput/MenuBar.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <MenuButton
+          active={editor?.isActive("bulletList") ?? false}
+          onClick={() => editor?.chain().focus().toggleBulletList().run()}
+          disabled={!editor?.isEditable || !editor?.can().toggleBulletList()}
+          icon={ListBulletIcon}
+        >
+          {intl.formatMessage(richTextMessages.bulletList)}
+        </MenuButton>
+        <LinkDialog editor={editor} />
+      </div>
```
vd1992, line 39: Is it possible for the editable prop to make the Menu inactive? Currently you can click on buttons to make dialogs pop up it just won't update the field with the value, just throw console errors

image

image

packages/forms/src/components/RichTextInput/LinkDialog.tsx
vd1992, line 1: For some reason updating with links isn't working right for me, not sure why. Lists work 

https://recordit.co/KkOxJk9lAw

vd1992 (changes requested):
Looks pretty impressive, got a few things to run by you

packages/forms/src/components/RichTextInput/MenuBar.tsx
esizer: Oh, huh. I thought it was supposed to with `!editor?.isEditable`. I guess not 😅  Thanks for catching that. I'll need to take a look.

packages/forms/src/components/RichTextInput/LinkDialog.tsx
esizer: Oh, yeah. I was trying to figure this out but you need to have some text selected before adding the link. I'm not sure how to disable the button until you do. I was looking for sometihng like `editor?.getSelection()` or `editor?.hasSelection()`.

You can do `editor?.view.state.selection` but I'm not sure how to check the return value 🤔 

If you are bored, maybe you could take a look at that and see 😅

vd1992:
Tried to put it through its paces and plopped it into a form, my implementation was a bit janky and the input not working completely right in that context but just for trial's sake.  

Submits to API successfully, stores in database successfully, retrieves successfully, and renders what it calls successfully! 🥳 

image

image

Recording of attempt

https://recordit.co/3ORNCoqBBE

packages/forms/src/components/RichTextInput/LinkDialog.tsx
vd1992: Okey, I took a crack at it, does this approach seem useful? 

image

Testing it here

https://recordit.co/KmN8QcT953

esizer:
> my implementation was a bit janky

Just to confirm, did you mean janky as in you just "threw it together for testing's sake" or that you found the component API/usage janky? 😅

packages/forms/src/components/RichTextInput/MenuBar.tsx
esizer: WOW! I knew it was working at one point. Looks like I refactored it a bit and forgot to pass the `editable: boolean` field. That was why this wasn't working 🤦‍♀️ 

Fixed in [fix read only mode](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7881/commits/34a380466694cfe447b8d6534e0478ffb26d97c0)

vd1992:
> > my implementation was a bit janky
> 
> Just to confirm, did you mean janky as in you just "threw it together for testing's sake" or that you found the component API/usage janky? 😅

I just threw it together for testing sake so not the most elegant way. 
But I think something was a bit quirky for me, give me a few minutes to re-review

packages/forms/src/components/RichTextInput/RichTextInput.tsx
vd1992, line 1: I still get this error appearing, in this screenshot, it was opening up the story, then clicking "List" that made the error pop up. 

image

I think this is also causing a quirk elsewhere, like my test implementation where clicking "List" causes the page to submit for some reason. 

https://recordit.co/L9fwVCz9Xy

vd1992, line 1: The same appears to occur for "Link". The TypeError above. 
As well as the auto-submit. 

https://recordit.co/pUT5rKTI1V

Although in this case, the request doesn't acknowledge it is a link and passes in paragraph tags.

image

Circumventing the auto-submit for list by typing in "-" rather than clicking the button
List is submitted correctly and renders like so, I didn't have a workaround for links so can't tell what it would submit to API 

https://recordit.co/rCXNk9lVBE

image

esizer: Thanks for digging in and providing more info. Looks like this was caused because our buttons don't set their type by default (I thogught they did). So, when we used them without setting `type=button` they were treated like submit buttons.

I set their type explicitly and did some hacks in the link dialog to prevent it from submitting the parent form. While the form isn't nested in the DOM (illegal) it seems the providers are and it was propogating up to the parent 😢 

[fix auto submit issue](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7881/commits/f7973c6da4ef19e5df82d05329596537490ec804)

esizer: Fixed in [fix auto submit issue](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7881/commits/f7973c6da4ef19e5df82d05329596537490ec804)

vd1992 (approved):
Amazing, appears to work flawlessly 💯 

https://recordit.co/iLAicnxJzA

PR #7886, opened 2023-09-13
[Fix] Revert auth bump

### LINKED ISSUE #7885: 🐛 No bearer token sent in first graphql request after login
## 🐛 Bug

We're experiencing an issue where the first graphql request after logging in (getCurrentAuthorizedUser) is not sent with the bearer token.  The second (e.g. ApplicantInformation) and subsequent requests do have it.  What this looks like to the user is that they can log in but get a "Sorry, you are not authorized to view this page." error.  If they refresh the page they can proceed without further problems.

## 🕵️ Details

- Commit 93607e85e works but 02c92589c does not so I'm pretty confident that the issue started with https://github.com/GCTC-NTGC/gc-digital-talent/pull/6547.
- The issue only occurs with prod builds.  Dev builds do not exhibit the problem.
- Happens with both mock auth and SiC.

## 📋 Steps to Reproduce

1. Set up the project with urql V4 and auth-exchange V2.
2. Build in prod mode.
3. Log in.
4. Observe the "Sorry, you are not authorized to view this page." error.

## 📸 Screenshot

First request:
image
Second request:
image

## ✅ Acceptance Criteria

- [ ] With a prod build, every graphql request after log in has the "authorization" header.

petertgiles:
🤖 Resolves #7885 

## 👋 Introduction

There was a previously undetected issue with the auth bump branch where prod builds would fail to authenticate properly.  This is a quick fix to roll back that bump so we can try again later.

## 🧪 Testing

1. Rebuild in prod mode (`npm run build`), not dev.
2. Log in works.
3. First graphql request after login has authorization header and succeeds.
4. Token refresh succeeds.
5. Log in as deleted user redirects to "deleted user" page.
6. [Introspection corner case](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6547#pullrequestreview-1602788860) does not cause refresh loop.
7. Steps 2 - 6 work with SiC, too.

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -62,7 +62,7 @@
     "react-table": "^7.8.0",
     "react-to-print": "^2.14.13",
     "react-toastify": "^9.1.3",
-    "urql": "^4.0.2"
+    "urql": "3.0.4"
   },
   "devDependencies": {
     "@babel/core": "^7.22.15",
--- a/package-lock.json
+++ b/package-lock.json
@@ -126,7 +126,7 @@
         "react-table": "^7.8.0",
         "react-to-print": "^2.14.13",
         "react-toastify": "^9.1.3",
-        "urql": "^4.0.2"
+        "urql": "3.0.4"
       },
       "devDependencies": {
         "@babel/core": "^7.22.15",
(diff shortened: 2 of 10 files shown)


petertgiles:
Issue to try again: https://github.com/GCTC-NTGC/gc-digital-talent/issues/7887

vd1992 (approved):
Okay...
Did `npm run build:fresh` for testing

Cannot replicate bug. 

Local
- Login works normally 
- Waiting a bit before navigating to another page works normally
- Sign in as deleted user redirects
- Breaking Introspection as shown triggers deleted user redirect 

SiC
- Login works normally
- Waiting a bit before navigating to another page works
- Sign in as deleted user redirects after SiC login

PR #7925, opened 2023-09-15
[Fix] Fix unexpected user deleted page redirect

### LINKED ISSUE #7924: 🐛 User saw "This account has been deleted" message when it was not accurate
## 🐛 Bug

A user saw the "This account has been deleted" message when logging in. They were later able to log in properly.

## 🦋 Expected Behaviour

Should only see this message if account really has been deleted. Should never see this message by mistake, even if _something_ went wrong during login.

## 🕵️ Details

Ask Dasha for more details about the user who reported this.

## 📋 Steps to Reproduce

Unknown

## 📸 Screenshot

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] Even if something goes wrong during login, user should never see "account has been deleted" message (unless it is accurate)

petertgiles:
🤖 Resolves #7924 

## 👋 Introduction

This branch (hopefully!) prevents the redirect to the "user deleted" page unless the API actually indicates that the user was deleted.

## 🕵️ Details

Unfortunately when [I rolled back the auth-bump PR](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7886) I rolled back some fixes that were made in there, too.  I put those back in and added some more fixes as well.

I noticed some error messages pop a toast but immediately log out and reboot so the user can't see them.  Might need an issue to ensure the user can see what's going on, even though these are pretty niche.

## 🧪 Testing

1. :warning: Very important :warning: You must build in production mode.  In dev mode, the auth error stuff does not behave the same way.
2. Make sure normal behavour works
 - [x] Can log in normally
 - [x] First graphql request after log in works normally
 - [x] Refresh after 2 minutes works OK
3. Error handling behaviour works
- [x] Delete user and ensure login triggers "user deleted" error
- [x] [Break the introspection query](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6547#pullrequestreview-1602788860) and ensure logout but not "user deleted"
- [x] Force internal server error by renaming a database column and ensure logout but not "user deleted"
- [ ] Any other corner cases you can think of?

## 📸 Screenshot

image

### DIFF
--- a/packages/auth/src/components/AuthorizationProvider.tsx
+++ b/packages/auth/src/components/AuthorizationProvider.tsx
@@ -1,7 +1,9 @@
 import React from "react";
 
 import { useGetCurrentAuthorizedUserQuery } from "@gc-digital-talent/graphql";
+import { containsUserDeletedError } from "@gc-digital-talent/client";
 import { Pending } from "@gc-digital-talent/ui";
+import { useLogger } from "@gc-digital-talent/logger";
 
 import AuthorizationContainer from "./AuthorizationContainer";
 
@@ -13,8 +15,10 @@ const AuthorizationProvider = ({ children }: AuthorizationProviderProps) => {
   const [{ data, fetching, stale, error }] = useGetCurrentAuthorizedUserQuery();
   const isLoaded = !fetching && !stale;
   let deleted = false;
+  const logger = useLogger();
 
-  if (error) {
+  if (error && containsUserDeletedError(error)) {
+    logger.debug("Detected UserDeleted error in AuthorizationProvider");
     deleted = true;
   }
 
--- a/packages/auth/src/components/AuthorizationProvider.tsx
+++ b/packages/auth/src/components/AuthorizationProvider.tsx
@@ -1,11 +1,11 @@
 import React from "react";
 
 import { useGetCurrentAuthorizedUserQuery } from "@gc-digital-talent/graphql";
-import { containsUserDeletedError } from "@gc-digital-talent/client";
 import { Pending } from "@gc-digital-talent/ui";
 import { useLogger } from "@gc-digital-talent/logger";
 
 import AuthorizationContainer from "./AuthorizationContainer";
+import { containsUserDeletedError } from "../utils/errors";
 
 interface AuthorizationProviderProps {
   children?: React.ReactNode;
--- a/packages/auth/src/utils/errors.ts
+++ b/packages/auth/src/utils/errors.ts
@@ -2,7 +2,7 @@
 import { CombinedError } from "urql";
 
 /**
- * Contains auth error
+ * Contains user deleted error
  *
  * Determine if one of the errors we received
  * is related to the user being deleted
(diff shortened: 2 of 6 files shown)


petertgiles:
Shoot, I've got a dependency loop between auth and client.  :face_exhaling:

packages/auth/src/utils/errors.ts
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Determine if one of the errors we received
+ * is related to the user being deleted
+ */
+export const containsUserDeletedError = (
+  combinedError: CombinedError | undefined,
+) => {
+  return combinedError?.graphQLErrors.some((graphQLError) => {
+    return Object.values(graphQLError.extensions).some((extension) => {
+      return ["user_deleted"].includes(String(extension));
```
tristan-orourke, line 14: This seems like an odd way to compare strings, why did you write it like this?

petertgiles: Yeah, it's a bit weird.  I copied it from https://github.com/GCTC-NTGC/gc-digital-talent/blob/b08a0c7243a777195f38722a3058896983e5cf28/packages/client/src/utils/errors.tsx#L113 since it matches two different strings.  I left it in this form since they were originally beside each other in the same file but I can switch it to a regular string comparison if you prefer.

vd1992:
Renaming table columns, to force a code 500

image

Doesn't appear to trigger logout, but no looping or user deleted message 

http://recordit.co/0KvUjAQ9W5

packages/auth/src/components/AuthorizationProvider.tsx
```diff
@@ -13,8 +15,10 @@ const AuthorizationProvider = ({ children }: AuthorizationProviderProps) => {
   const [{ data, fetching, stale, error }] = useGetCurrentAuthorizedUserQuery();
   const isLoaded = !fetching && !stale;
   let deleted = false;
+  const logger = useLogger();
 
-  if (error) {
+  if (error && containsUserDeletedError(error)) {
+    logger.debug("Detected UserDeleted error in AuthorizationProvider");
     deleted = true;
```
vd1992, line 22: This feels like it could be the core part of the problem, _any_ error sets `deleted = true` in the context that appears everywhere  
Hopefully this cleans things up nicely 🤞

packages/auth/src/utils/errors.ts
petertgiles: I've gone ahead and switched this to a regular string comparison.
[better string comparison](https://github.com/GCTC-NTGC/gc-digital-talent/pull/7925/commits/cc54b642243fc27907f13b8eae88e5d21c79af70)

tristan-orourke:
I noticed that when I first log in as a deleted user, I see a flash of the "not authorized" error page before ending up on the logged out page. It's unintended, so I want to flag it, but I'm also not overly concerned.
https://recordit.co/MAFy4hjmFx

tristan-orourke:
I found the same thing as Vachan - when causing a server error by changing a DB column, the frontend shows "Oh no, graphql error", but doesn't log the user out. I think that's expected though, right?

tristan-orourke (approved):
Looks good to me. I didn't explore any corner cases besides the ones suggested in the PR description, though.

Issue #7920, opened 2023-09-14
✨ Update pool publishing validation


vd1992:
## ✨ Feature

Update pool publishing validation to account for the assessment plan

## 🕵️ Details

The assessment plan has its own validation in the frontend.  If the assessment plan is incomplete that should prevent publishing in the backend as well.

## 🙋‍♀️ Proposed Implementation

At a minimum update `PublishPoolValidator.php` with a status check 

## :question: Questions
Q: Do we want to tie this to a feature flag?
A: Yes, the assessment plan login should not activate if the RoD flag is false.

## ✅ Acceptance Criteria

- [ ] update validator
- [ ] uses the same logic as the frontend (fail if thereAreUnassessedPoolSkills or thereAreAssessmentStepsWithNoSkills)
- [ ] tied to RoD feature flag

tristan-orourke:
Blocked because we're not sure we actually want to do https://github.com/GCTC-NTGC/gc-digital-talent/issues/7918

petertgiles:
I've reworked this for RoD.

Issue #7918, opened 2023-09-14
✨ Ready for publishing


vd1992:
## ✨ Feature

Communicate that a pool is ready for publishing by a platform admin

## 🕵️ Details

Frontend, render ready to publish status for pool operator
Backend, added field and the ability to update it, status accessor to output new status
Policy, limited to pool operator 

Update other things as needed...

## 🎨 Design File

Include a link to the design file (if it exists).

### 🧑‍🎨 Designer

Include the author of the linked design file's name (or even better, their `@githubUsername`) for future reference.

## 📸 Screenshot

Add a screenshot of the new feature (if possible).

## 🙋‍♀️ Proposed Implementation

Datetime field added like `published_at`, maybe called `ready_at` introduced to pool model
Ready to published is `ready_at < now() && published_at === null`
Hit the submit for publishing sets the field to current datetime

Update draft status to now require `ready_at === null` 

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the feature was properly implemented.

- [ ] Criteria 1
- [ ] Criteria 2

brindasasi:
@Jerryescandon do we have a figma link for this ?

vd1992:
And for that matter, a goal? I remember we were unsure about even implementing this field hence the lack of acceptance criteria

brindasasi:
Its not higher priority now

brindasasi:
We can achieve the goal without introducing the new state by using notifications instead.

petertgiles:
Stale issue.  We never ended up developing " Ready for publishing"

PR #7926, opened 2023-09-16
[Feature] Pool completeness accessor 

### LINKED ISSUE #7917: ✨ Pool completeness variable
## ✨ Feature

The ability to check whether a pool is "complete", all required fields are complete and should be allowed to be published. 
Essential to start pool rework so high priority 

## 🕵️ Details

Backend check that emits a boolean through API enabling checking if it is complete on the frontend

Accessor on the pool model that works akin to profile completeness on the user model (validator in the accessor)
`isComplete`

## ✅ Acceptance Criteria

- [ ] Completeness field added to pool schema type, can be queried for a pool
- [ ] Basic test asserting incomplete pools emit the correct boolean

vd1992:
🤖 Resolves #7917 

## 👋 Introduction

Introduces an accessor that allows checking from API/frontend whether a pool is "complete", allowing checking without publishing. 

## 🕵️ Details

Adds an accessor, that is acting with a validator per issue body and leverages a reusable rule. 
Tests it too. 

## 🧪 Testing

1. Open up a draft pool in edit mode in one tab, graphiQL in another
2. Query with the below, while changing the pool around, checking validation logic

```
query pool {
  pool(id: :id) {
    isComplete
  }
}

```


petertgiles (changes requested):
Nice, this is very clean!  I love you've included a test, too.  The one thing I'm not fond of is that you've added a validation routine that is very similar to the one already there.  Is it possible to either reuse or replace the existing validation logic?
image

vd1992:
> Nice, this is very clean! I love you've included a test, too. The one thing I'm not fond of is that you've added a validation routine that is very similar to the one already there. Is it possible to either reuse or replace the existing validation logic?

Yeah, it is very similar because I used it as a guide for what a complete pool should look like 🤣 

But yeah, my intent with introducing a `PoolIsComplete` rule was to make it so an issue like #7920 or a debt issue could recycle the rule. Updating the pool publish validator to refactor to use the new rule seemed to really push the scope on this issue so I did not do it now.

vd1992:
@tristan-orourke 
Can I please request your opinion on the PR, and the conversation?

tristan-orourke:
@vd1992 I'd ask you reuse the PublishPoolValidator for this, for now. Not only does that avoid duplicated code, but by using a set of rules, instead of creating one new rule, the validation error will say which fields are missing.

tristan-orourke:
I can see your point about #7920 suggesting that we might need two different validators. This might be a use case for class extension or composition? Like maybe, 
```
class PublishPoolValidator {
  rules() {
    $completenessRules = (new PoolIsCompleteValidator)->rules();
    $extraRules = ...;
    return array_merge($completenessRules, extraRules);
  }
}
```

vd1992:
> @vd1992 I'd ask you reuse the PublishPoolValidator for this, for now. Not only does that avoid duplicated code, but by using a set of rules, instead of creating one new rule, the validation error will say which fields are missing.

Okay sure I can do that, my concern about all this was how there will be a gap between what is considered "complete" and what can be published, which will arise from issues #7918 and #7920 
Those issues are not set in stone, but I am a bit concerned as there will be a gap between a pool being "complete" and "ready to publish" so I feel hesitant about using the validator attached to "publish" for the purpose of "completeness" ☹️

vd1992:
Okay, I factored out a completeness validator and used it instead

petertgiles (approved):
Very nice!  I like this much better.  :+1: Works well for me.

tristan-orourke (approved):
Nice

petertgiles:
You're in a race with James' enum PR.  Whichever one is in second place will have some fixup to do.

PR #7931, opened 2023-09-18
add page index for the pagination

brindasasi:
🤖 Resolves {issue #}

## 👋 Introduction

## 🕵️ Details

## 🧪 Testing

1. ...
2. ...

## 📸 Screenshot

## 🚚 Deployment

- re-running database seeders

> **Notes**
>

### DIFF
--- a/apps/web/src/components/Pagination/Pagination.tsx
+++ b/apps/web/src/components/Pagination/Pagination.tsx
@@ -87,6 +87,8 @@ const Pagination = ({
   const lastPage = paginationRange[paginationRange.length - 1];
   const isLeftArrowDisabled = currentPage === 1;
   const isRightArrowDisabled = currentPage === lastPage;
+  const currentPageIndex = (currentPage - 1) * pageSize + 1;
+  const currentPageSize = currentPage * pageSize;
 
   const iconStyles = {
     "data-h2-width": "base(1rem)",
@@ -128,13 +130,15 @@ const Pagination = ({
         <span>
           {intl.formatMessage(
             {
-              defaultMessage: "Showing results {pageSize} of {total}",
-              id: "gMWuvj",
+              defaultMessage:
+                "Showing results {currentPageIndex} - {currentPageSize} of {total}",
+              id: "2QxibW",
               description:
                 "Description of how many items are being displayed out of the total value",
             },
             {
-              pageSize: pageSize > totalCount ? totalCount : pageSize,
+              currentPageIndex,
+              currentPageSize,
               total: totalCount,
             },
           )}


apps/web/src/components/Table/ResponsiveTable/ColumnDialog.tsx
```diff
@@ -61,7 +66,7 @@ const ColumnDialog = <T extends object>({ table }: ColumnDialogProps<T>) => {
                     }}
                   />{" "}
                   {intl.formatMessage(adminMessages.toggleAll)}
-                </label>
+                </Field.Label>
               </div>
             </Field.BoundingBox>
```
esizer, line 71: Something I noticed is that the bounding box only encompasses the first checkbox. @substrae Should we extend it to wrap all of the checkboxes?
Screenshot 2023-09-28 105318

apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx
```diff
@@ -201,45 +203,64 @@ const Actions = ({
...
+              >
+                {intl.formatMessage({
+                  defaultMessage: "Clear<hidden> row selection</hidden>",
+                  id: "VHG9Gm",
+                  description: "Button text to deselect all table rows",
+                })}
+              </Button>
+            </span>
+            <span>
+              {download?.selection && (
```
esizer, line 230: I think we move the conditional above the `span` so we don't have an empty `span` tags.

apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx
```diff
@@ -201,45 +203,64 @@ const Actions = ({
...
+                        id: "mxOuYK",
+                        description:
+                          "Text label for button to download a csv file of items in a table.",
+                      })}
+                  </DownloadCsv>
+                </>
+              )}
+            </span>
+            <span>
+              {print?.onPrint && (
```
esizer, line 249: See previous comment about moving the conditional.

esizer (changes requested):
This is looking great! 😁 

I just noticed a few tiny things.

esizer (changes requested):
This looks amazing! ⭐ 

One minor thing, for non-sortable columns, the labels appear to be bolded which is slightly confusing (made me think they were locked and sorted by default). Should be reduce the weight of non-sortable column headers @substrae ?

Screenshot 2023-10-03 095411

substrae:
@esizer @brindasasi 

Yeah, that sounds right to me - the bold should only be used for sorted columns, so reduce the font weight on non-sortable columns and ensure there's no underline.

esizer:
> I'm also noticing that bug from a while back creeping into this new table where some of the values are centered, while others are not (including table headers). Can we make sure that everything is left aligned always?

This is a known issue with our links. Somehow making them all centered snuck in with that massive update we did recently. We have been fixing them gradually but possibly your dark mode PR may fix it? 🤔

esizer (approved):
Great! 👏

PR #7937, opened 2023-09-19
[Fix] Experience cards surface single skill

### LINKED ISSUE #7818: ✨ `ExperienceCard` should surface skill details in a single skill context
image

## ✨ Summary

[Visit the figma file](https://www.figma.com/file/guHeIIh8dqFVCks310Wv0G/Style-library?type=design&node-id=42-9546&mode=design&t=6ArPOkkLP9wljVgr-4) • Contact @substrae with design questions

In instances where a user is viewing and/or editing a single skill's relationship with a given experience, it would be helpful to surface that skill's information to the top level of the card. Currently, the experience card simply shows a list of all linked skills, so it forces the user to click the details accordion, sift through the list to find the correct skill, and then read from there. This layout also makes it unclear whether the "Edit" button edits the experience as a whole or the target skill's relationship with that experience

Contexts where this should be implemented:
- Application skills page
- Update user skill page
- ...?

**Note:** This shouldn't change the featured skills layout that exists now, but rather, replace it for the contexts above. A full list of featured skills should still exist in cases where the user is viewing an experience as a full item (e.g. on their career timeline).

## 📸 Screenshot

image

This could be a little odd since we get the skill's and their details via the API `Experience` type returned in the schema. Maybe look into using the `showSkills` prop that accepts `boolean | Array<Skill>` and update it to accept `boolean | Skill | Array<Skill>`. Then if it is only a single skill render it in a different location? 🤔 

## ✅ Acceptance Criteria

- [ ] `ExperienceCard` can accept a single skill
- [ ] When a single skill is passed, the details are rendered outside of the collapsible element
- [ ] Appropriate areas are updated (see details)

esizer:
🤖 Resolves #7818 

## 👋 Introduction

Allows devs to pass a single skill to `ExperienceCard`.

## 🕵️ Details

When you pass a single skill, the card should show the `experienceSkillRecord.details` above the `Collapisble` component and not inside it.

## 🧪 Testing

2. Navigate to the "Single skill - Experience card" story
3. Confirm details are shown without opening the collapsible element
4. Confirm other card stories work as expected
5. Build `npm run dev`
6. Navigate to a application skills step
7. Confirm details are shown without opening the collapsible element
8. Navigate to the skill library and edit  a skill
9. Link an experience
10. Confirm details are shown without opening the collapsible element

## 📸 Screenshot
Screenshot 2023-09-19 134528

### DIFF
--- a/apps/web/src/components/ExperienceCard/ExperienceCard.stories.tsx
+++ b/apps/web/src/components/ExperienceCard/ExperienceCard.stories.tsx
@@ -1,9 +1,15 @@
 import React from "react";
 import { StoryFn } from "@storybook/react";
+import { faker } from "@faker-js/faker";
 
-import { experienceGenerators } from "@gc-digital-talent/fake-data";
+import {
+  experienceGenerators,
+  getStaticSkills,
+} from "@gc-digital-talent/fake-data";
 
 import ExperienceCard from "./ExperienceCard";
+
+faker.seed(0);
 
 export default {
   component: ExperienceCard,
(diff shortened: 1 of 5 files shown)


esizer:
> Very nice! I like the use of typing to use the same prop.

Oh good! I was a little worried that approach might seem odd. Glad it worked out! 👍

Issue #4302, opened 2022-10-13
Skill can be marked as both "Essential" and "Asset" for a pool


JamesHuf:
A skill can be marked as both an 'essential skill' and an 'asset skill' on a pool.

This isn't really a problem but it looks kind of odd.

## Acceptance Criteria
- [ ] Adding a skill as essential prevents it from being added as asset
- [ ] Added a skill as asset prevents it from being added as essential
- [ ] Removing a skill from one category makes it possible to add it as the other
- [ ] Enforced in the API as well as frontend validation

mnigh:
#7953 seems to have solved this in the API but not on the frontend. i've tested locally and can confirm, almost there. 

@petertgiles this issue could probably be re-estimated and brought down from a 5️⃣.

vd1992:
Closing in favour of #8013  
Will decide on some frontend stuff there

Issue #7833, opened 2023-09-06
✨ Restore details when re-adding a soft-deleted Experience-Skill


tristan-orourke:
## ✨ Feature

Allow frontend to restore "details" of soft-deleted Experience-Skill pivots.

## 🕵️ Details

When we unlink a skill from an experience (or delete the UserSkill, which cascades into the same thing) we soft-delete the Experience-Skill pivot. We can re-add the same skill to the same experience via API, and the "details" we previously wrote will be restored.

However, on the frontend, we always require users to write details when we add the skill to the experience. That means that even if the details exist on a soft-deleted pivot, the old details will be overwritten when we re-add it.

The way around this is to allow the frontend to query for soft-deleted Experience-Skills, and add the old "details" directly in the input form if available. Then the user will have the option of submitting it as-is or re-writing it.

This applies to anywhere we add skills to experiences.

### 🧑‍🎨 Designer

@substrae requested this behaviour, and is probably disappointed it's not already working like this 😅 

## 🙋‍♀️ Proposed Implementation

May need to add or modify a graphql query to allow frontend to get soft-deleted data.

## ✅ Acceptance Criteria

- [ ] When adding a skill to an experience, if a soft-deleted version of the pivot exists, the old "details" will appear in the input text before submitting.

vd1992:
notes, unsure about how to best implement
concern about different methods
 -> trash collection fields exposing unwanted stuff
-> tweaks to adding skill process, like add skill upon select

PR #7953, opened 2023-09-21
[Debt] Link Pools to Skills with a single pivot table

### LINKED ISSUE #7949: ♻️ Link Pools to Skills with a single pivot table (instead of two)
## ♻️ Debt/Refactor

Pools are linked to Skills via two pivot tables currently, one for essential skills and one for asset. Some future work (starting with #7922) would be easier if this used only a single pivot table, which held additional details about the connection (starting with whether its an optional or asset connection).

## 🕵️ Details

Current model
```mermaid
classDiagram

class Pool { }

class pools_essential_skills { }

class pools_nonessential_skills { }

Pool "1" o-- "0..*" pools_essential_skills
Pool "1" o-- "0..*" pools_nonessential_skills
pools_essential_skills "0..*" o-- "1" Skill
pools_nonessential_skills "0..*" o-- "1" Skill
```

Ideal data model:

```mermaid
classDiagram

class Pool { }

class pool_skill {
    PoolSkillType type (essential/asset)
}

Pool "1" o-- "0..*" pool_skill
pool_skill "0..*" o-- "1" Skill
```

## 🙋‍♀️ Proposed Solution

The question is, do we update the graphql schema at the same time as updating the database implementation?

I think it would be easiest to leave the schema as is, and rewrite the `Pool->essentialSkills` and `Pool->nonessentialSkills` to be `hasManyThrough` relationships. Later we can deprecate these fields in the schema in favour of a new `PoolSkill` type (which includes an essential/nonessential enum field) and gradually shift queries over to that.

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Pools are connected to Skills via a single pivot table in the database
- [ ] The migration properly moves all existing relationships to the new table before deleting the current pivot tables
- [ ] Tests, seeders, and current queries and mutations (like setting pool skills) continue to work.

vd1992:
🤖 Resolves #7949 

## 👋 Introduction

Merges the two pivot tables relating pools and skills into one. 

## 🕵️ Details

Executed a migration, and updated the pool functions to continue returning the same results. 
Subsequently, things really didn't change from the API endpoint, and some tweaks in how syncing is called in `/api`

The directive `resolveWith` Tristan added, helped.  

Think it also closes as a side effect https://github.com/GCTC-NTGC/gc-digital-talent/issues/4302

## 🧪 Testing

1. Seed and migrate in main, switch to branch, test migration preserves pool skill relations
2. Test normal workflows in branch, like pool operations, they should be unchanged


api/database/migrations/2023_09_20_211150_combine_pool_skills.php
```diff
@@ -0,0 +1,94 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        // create the combined table
+        Schema::create('pools_skills', function (Blueprint $table) {
```
petertgiles, line 16: By convention in Laravel, pivot tables are named with the singular model names in alphabetical order.  I think this should be `pool_skill`.

api/database/migrations/2023_09_20_211150_combine_pool_skills.php
```diff
@@ -0,0 +1,94 @@
...
+            $table->uuid('skill_id');
+            $table->foreign('skill_id')->references('id')->on('skills');
+            $table->uuid('pool_id');
+            $table->foreign('pool_id')->references('id')->on('pools');
+            $table->unique(['skill_id', 'pool_id']);
```
petertgiles, line 21: The good news is that this PR also fixes #4302.  The bad news is that #4302 means that this migration might fail in prod on deployment.  I suggest either confirming that it won't fail in prod ahead of time or updating this migration to gracefully handle it. Maybe by collecting both batches of skills and deduplicating them first? I guess if a skill is marked as both then it should go in as essential.

api/app/Models/Pool.php
```diff
@@ -113,12 +114,48 @@ public function publishedPoolCandidates(): HasMany
 
     public function essentialSkills(): BelongsToMany
     {
-        return $this->belongsToMany(Skill::class, 'pools_essential_skills')->withTrashed(); // always fetch all skills for a pool
+        return $this->belongsToMany(Skill::class, 'pools_skills')
+            ->withTrashed() // pool-skills always fetches soft-deleted skill models
```
petertgiles, line 118: I don't understand why this relationship needs to always return trashed skills?  I know we have to deal with this in #7833 but I'm not sure we want to make this change in here.

api/app/Models/Pool.php
```diff
@@ -113,12 +114,48 @@ public function publishedPoolCandidates(): HasMany
...
+            ->wherePivot('type', PoolSkillType::NONESSENTIAL->name);
+    }
+
+    /**
+     * Sync the essential skills in pools_skills
+     *
+     * @param $skills - array of skill ids
+     * @return void
+     */
+    public function setEssentialPoolSkills($skills)
```
petertgiles, line 135: These functions feel like what we're trying to get away from - treating the two sets of skills differently.  I'd suggest marking them (and the two mutations in the graphql schema they support) as deprecated and not using them in tests, factories, etc.

I suppose that means you would need to add a regular skills(): BelongsToMany relationship that returns both types and can sync both types together.  That might be a bit tricky but maybe details field on experience_skill can be a good example.

petertgiles (changes requested):
Very cool!  I think this a solid start.  I've flagged a few things that might be a bit tricky.  Feel free to get a second opinion if you disagree.

api/app/Models/Pool.php
tristan-orourke: Its because if a Pool has been soft-deleted (ie deprecated) but was already added to a pool in the past, we still want to return it

api/database/migrations/2023_09_20_211150_combine_pool_skills.php
tristan-orourke: @vd1992 is already ahead of us, @petertgiles! 😆 Deduplication happens at line 37: 
```
$nonessentialPoolSkills = DB::table('pools_nonessential_skills')
            ->whereNotIn('skill_id', $essentialPoolSkillsIds) // do not have repeat pool-skill entities with different type
            ->get();
```

api/app/Models/Pool.php
tristan-orourke: Hmm. I'm not sure if I fully agree. I think that eventually (when we want to add skill levels) we'll want to add a `poolSkill: BelongsToMany` relationship, and maybe deprecate the `essentialSkills` and `nonEssentialSkills` fields on graphql Pool type. 

But at the same time, I'm not against using mutations like this if they make writing mutations easier for us, or make using migrations easier for the client.

vd1992: I just approached with keeping things working the same per the proposed solution section, so the frontend doesn't change at all for example

> Later we can deprecate these fields in the schema in favour of a new PoolSkill type (which includes an essential/nonessential enum field) and gradually shift queries over to that.

It seemed like things might change a bunch so I attempted to maintain status quo.

api/app/Models/Pool.php
vd1992: There is no change though, it was `withTrashed()` before and after, spacing and comments is all that changed

petertgiles: Oh man, I missed that too.  Lol, I need to pay more attention in my reviews.  :laughing:

api/app/Models/Pool.php
petertgiles: Fair enough.  Since this is blocking work this sprint I'll go along with the quicker solution for now.

api/database/migrations/2023_09_20_211150_combine_pool_skills.php
vd1992: 🤦‍♂️ 

Unresolving because looking more closely at that code block, I need to update the filtering. 
It needs to filter for unique `skill_id` AND `pool_id`, as right now I think it just makes it so you can't have any skill be used as an asset and essential for any pool. Which is bad (only unique `skill_id` on the pivot)

api/database/migrations/2023_09_20_211150_combine_pool_skills.php
vd1992: Done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7953/commits/901217cd56479c1d72a27cc9881cce032affa221

As you can tell from this PR alone, removing `pools_essential_skills` and `pools_nonessential_skills`, does not appear that convention has been followed much 😛

api/database/migrations/2023_09_20_211150_combine_pool_skills.php
vd1992: Okay, now I think it should work as intended, testing and it migrates successfully while removing the actual duplicates

https://github.com/GCTC-NTGC/gc-digital-talent/pull/7953/commits/9609b5dc312a3e2de1f7ea9c4b292ac020eacdd2

api/database/migrations/2023_09_20_211150_combine_pool_skills.php
petertgiles: Yeah, it's true.  :disappointed: Some have definitely slipped through.

petertgiles: Ref: https://laravel.com/docs/10.x/eloquent-relationships#many-to-many-table-structure

api/tests/Feature/PoolApplicationTest.php
vd1992, line 1: Not sure why this test started failing due to "uniqueness violation in `pool_skill`" _after_ adding in the `id` field to table `pool_skill` (all CI passed three commits ago) and not before. But the adding essential skills step was redundant as the pool factory sets them already. 

Testing syncing skills in the frontend reveals no problems to me though?

https://recordit.co/i0FlqCWP1a

api/database/migrations/2023_09_20_211150_combine_pool_skills.php
petertgiles: :+1:

api/database/migrations/2023_09_20_211150_combine_pool_skills.php
petertgiles: :+1:

api/database/migrations/2023_09_20_211150_combine_pool_skills.php
petertgiles, line 1: This adds a new rough edge to the app.  If you try to add a skill as nonessential that is already set as essential the user gets a vague error message.  I think this is out of scope for this PR but I think we should add some sort of back-end (for better error) and front-end validation (to prevent it from happening at all) for this.  Perhaps even some design work.
image

petertgiles (changes requested):

petertgiles:
Nice, I think this is working well for me now.  Just one more small fix left, I think.

api/tests/Feature/PoolApplicationTest.php
petertgiles: I'm not seeing this fail.  Did you fix it already or is this an intermittent failure?

api/database/migrations/2023_09_20_211150_combine_pool_skills.php
tristan-orourke: That's a good point. My first thought was that the easiest quick fix would be a bit of validation on the updatePool mutation... but I don't know how to write this validation rule, so it might not be a quick fix at all! 😂 https://laravel.com/docs/10.x/validation

tristan-orourke: Spinning off another issue makes sense.

api/tests/Feature/PoolApplicationTest.php
vd1992: It was fixed by the changes to the file in question.

api/database/migrations/2023_09_20_211150_combine_pool_skills.php
vd1992: I decided to dabble in it for a bit, it seems like it'll be tricky. 

You are checking for uniqueness in a pivot, but you need to check two columns. Which is doable but then you are operating on a sync which feeds in ids, and you need to validate whether the sync would cause a uniqueness violation. So the validation would also need to dynamically ignore things in the array of ids if they were already present in the pivot, the sync passes in new things as well as what was already synced in one array. 

Issue made here #8013

api/app/Models/Pool.php
```diff
@@ -113,12 +114,48 @@ public function publishedPoolCandidates(): HasMany
 
     public function essentialSkills(): BelongsToMany
     {
-        return $this->belongsToMany(Skill::class, 'pools_essential_skills')->withTrashed(); // always fetch all skills for a pool
+        return $this->belongsToMany(Skill::class, 'pool_skill')
```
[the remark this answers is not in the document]
vd1992: Thanks! Done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7953/commits/ee055aeeb1e12911e5084fecbf9762a3ec2d5e6a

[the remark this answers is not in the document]
petertgiles: Nice one.  :+1:

petertgiles (approved):

vd1992:
All good from your perspective too, @tristan-orourke ?

api/database/migrations/2023_09_20_211150_combine_pool_skills.php
```diff
@@ -0,0 +1,104 @@
...
+                    return true;
+                }
+            }
+        });
+        foreach ($filteredNonessential as $instance) {
+            DB::table('pool_skill')->insert([
+                'skill_id' => $instance->skill_id,
+                'pool_id' => $instance->pool_id,
+                'type' => 'NONESSENTIAL',
+            ]);
```
tristan-orourke, line 54: It would be much faster to insert everything in a single DB operation. Use the loop to assemble an array of `pool_skill` rows, then you only need a single `insert` call. https://laravel.com/docs/10.x/queries#insert-statements

tristan-orourke: This goes for the essential skills too, of course

api/app/Models/Pool.php
```diff
@@ -113,12 +114,50 @@ public function publishedPoolCandidates(): HasMany
...
+            ->wherePivot('type', PoolSkillType::NONESSENTIAL->name);
+    }
+
+    /**
+     * Sync the essential skills in pool_skill
+     *
+     * @param $skills - array of skill ids
+     * @return void
+     */
+    public function setEssentialPoolSkills($skills)
```
tristan-orourke, line 137: I'd recommend renaming `$skills` to `$skillIds`, to make it obvious that the whole object shouldn't be passed in. Same for line 153.

api/app/Models/PoolCandidate.php
```diff
@@ -535,8 +535,8 @@ public function createSnapshot()
             'essentialSkills',
             'nonessentialSkills',
         ])->findOrFail($this->pool_id);
-        $essentialSkillIds = $pool->essentialSkills()->pluck('id')->toArray();
-        $nonessentialSkillIds = $pool->nonessentialSkills()->pluck('id')->toArray();
+        $essentialSkillIds = $pool->essentialSkills()->pluck('skills.id')->toArray();
```
tristan-orourke, line 538: Why this change? Doesn't essentialSkills return the same thing it did before?

tristan-orourke (changes requested):
Sorry, found few more suggestions and a question

api/app/Models/PoolCandidate.php
vd1992: Upon adding id to the new pivot table in this commit, tests started failing https://github.com/GCTC-NTGC/gc-digital-talent/pull/7953/commits/cfe5c212d7b70d75c1f5fad86ffc0493a0ebc158

The failing tests seemed similar to me, they related to the snapshot 

image

So I concluded that it was because calling

`pool->essentialSkills()->pluck('id')`

Was ambiguous on `id` due to `id` being present on `skills` and `pool_skill`
Specifying which table to pull id from resolved things.

api/database/migrations/2023_09_20_211150_combine_pool_skills.php
vd1992: Done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7953/commits/a54bbf43eff4167110264b0ff9a3cc8d13b4a006
Running migration forward and back even without a huge amount of rows and there is still a noticeable speed change

api/app/Models/Pool.php
vd1992: Aye aye https://github.com/GCTC-NTGC/gc-digital-talent/pull/7953/commits/f8af0abc4da9755572327b418b2c3941f2874036

api/app/Models/PoolCandidate.php
tristan-orourke: Huh, okay. I guess its because you're chaining methods on the relationship querybuilder directly, not on the resulting objects.

tristan-orourke (approved):
Thanks, I'm happy with this now

PR #8000, opened 2023-09-22
[Debt] Remove expectedGenericJobTitle from schema

### LINKED ISSUE #7645: ♻️ Remove deprecated `User.expectedGenericJobTitles` from graphql schema
## ♻️ Debt/Refactor

This field on the user was deprecated as part of the roles and salary expectations removal. We will need to remove it entirely eventually.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/40e4300c83d1fcb92f5a18cc29eea9b8aab351af/api/graphql/schema.graphql#L136-L138

## ✅ Acceptance Criteria

- [ ] `User.expectedGenericJobTitles` removed from schema file
- [ ] Queries using this field have been updated

yonikid15:
🤖 Resolves #7645

## 👋 Introduction

- Removes `User.expectedGenericJobTitle` from schema.
- The field was not found in any queries throughout the app.

## 🧪 Testing

1. Ensure app runs correctly and all tests pass

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -141,11 +141,6 @@ type User {
     ) # Pools a user owns
   # Profile Status
   isProfileComplete: Boolean
-  expectedGenericJobTitles: [GenericJobTitle]
-    @belongsToMany
-    @deprecated(
-      reason: "expectedGenericJobTitles is deprecated. Remove in #7645."
-    )
   priorityWeight: Int @rename(attribute: "priority_weight")
   # teams and roles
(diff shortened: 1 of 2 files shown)


mnigh (changes requested):
I think you need to run `npm run codegen` as well to generate the compiled schema. For docs, see:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/5ef65bbb26a87ce59cc2f78539f8743a5c8d46aa/api/README.md?plain=1#L30

PR #8011, opened 2023-09-25
[Feature] Add snapshot link to pool status table

### LINKED ISSUE #7372: Make snapshot downloadable always
As an admin, I want to be able to download the application snapshot even after candidates are qualified so that we can share the original application with HR for hiring them.

esizer:
🤖 Resolves #7372 

## 👋 Introduction

Adds a link to the application snapshot to the pool status table on a users profile.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to a users profile in admin
3. If they are not in a pool, use the "Add user to pool" button to add them to one
4. Confirm there is a new column for "Application" that links to the snapshot

## 📸 Screenshot

Screenshot 2023-09-25 111746

### DIFF
--- a/apps/web/src/components/PoolStatusTable/PoolStatusTable.tsx
+++ b/apps/web/src/components/PoolStatusTable/PoolStatusTable.tsx
@@ -4,7 +4,7 @@ import isEmpty from "lodash/isEmpty";
 import isPast from "date-fns/isPast";
 
 import { Link, Well } from "@gc-digital-talent/ui";
-import { getPoolCandidateStatus } from "@gc-digital-talent/i18n";
+import { getPoolCandidateStatus, getPoolStream } from "@gc-digital-talent/i18n";
 import { PoolCandidate } from "@gc-digital-talent/graphql";
 import { parseDateTimeUtc } from "@gc-digital-talent/date-helpers";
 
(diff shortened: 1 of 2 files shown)


apps/web/src/components/PoolStatusTable/PoolStatusTable.tsx
```diff
@@ -172,10 +179,31 @@ const PoolStatusTable = ({ user, pools }: UserInformationProps) => {
...
+                    {intl.formatMessage(
+                      {
+                        defaultMessage: "View {title} application",
+                        id: "+PCQ8p",
+                        description:
+                          "Link text to view a specific application of a candidate",
+                      },
+                      {
+                        title: intl.formatMessage(
+                          getPoolStream(candidate.pool.stream ?? ""),
```
vd1992, line 196: Just a small question for this 
This should never (probably) be null, but in this case an empty string in `getPoolStream()` will throw an error

image

Is that preferable to something like this to handle the null more gracefully? It'd sub in a empty space instead of erroring. 

image

vd1992 (approved):
Looking great, works for me. 
Small question about best way to handle a null case below, but not PR stopping

apps/web/src/components/PoolStatusTable/PoolStatusTable.tsx
esizer: Good call, added in a N/A message if it isn't there.

[check for pool stream before rendering](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8011/commits/3921cf95705fe0a83851186b0c52cef2360bad5b)

PR #8024, opened 2023-09-27
[Debt] Replaces "Cancel and go back" with single source

### LINKED ISSUE #8012: ♻️ Replace "Cancel and go back" with single source
## ♻️ Debt/Refactor

Replace all instances of the string "Cancel and go back" with the `formMessages` definition.

## 🕵️ Details

We currently have 33 instances of this string throughout the application. We also have it in `formMessages`. We should replace each instance (where it makes sense) with the single source of truth.

Screenshot 2023-09-25 113642

https://github.com/GCTC-NTGC/gc-digital-talent/blob/918c5226f1a4065292cba202e0469de8ebf22829/packages/i18n/src/messages/formMessages.ts#L111-L115

## ✅ Acceptance Criteria

- [ ] Only one reference to "Cancel and go back"

mnigh:
🤖 Resolves #8012.

## 👋 Introduction

This PR replaces all instances of the string "Cancel and go back" with the `formMessages` definition.

## 🧪 Testing

1. Search for instances of string "Cancel and go back"
2. Observe only one instance in codebase for "Cancel and go back"

### DIFF
--- a/apps/web/src/components/ExperienceSkillFormDialog/ExperienceSkillForm.tsx
+++ b/apps/web/src/components/ExperienceSkillFormDialog/ExperienceSkillForm.tsx
@@ -11,7 +11,7 @@ import {
   Well,
 } from "@gc-digital-talent/ui";
 import { Select, TextArea } from "@gc-digital-talent/forms";
-import { errorMessages } from "@gc-digital-talent/i18n";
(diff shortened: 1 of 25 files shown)


esizer (approved):
Sweet! 🍬

PR #8026, opened 2023-09-27
[Bug] Submit application essential skills

### LINKED ISSUE #8025: 🐛 Application submits without essential skills
## 🐛 Bug

User can submit application with no experience skills attached. As long as they have user skills of each essential skill then the submission will pass.

## 🦋 Expected Behaviour

Ensure application has an experience skill for all essential skills of the job poster.

## 📋 Steps to Reproduce

1. Log in as an applicant
2. Apply to a job. Continue the application as far as Review step
3. Go back to skill step and remove all experience connections to at least one skill.
4. Go back to review step and submit application. See it pass.
5. You can go to admin side and check the application and see there are no experience skills in essential skills column.

## 🙋‍♀️ Proposed Solution

In `HasEssentialSkills` rule, ensure that an experience skill exists for all essential skills in the pool.

## ✅ Acceptance Criteria

- [ ] User can only submit application if they have experience skill connection for all essential skills.

yonikid15:
🤖 Resolves #8025 

## 👋 Introduction

- Ensures user cannot submit application without having an experience skill for all essential skills.

## 🧪 Testing

1. Login as admin.
2. Create an application and complete all steps.
3. Remove experience skills either through the db or by going back to skill step.
4. Go to last step and try submitting application.
5. Ensure error message (toast) pops up.
6. Try adding experience skills and submitting again to see it pass.

### DIFF
--- a/api/app/Rules/HasEssentialSkills.php
+++ b/api/app/Rules/HasEssentialSkills.php
@@ -2,6 +2,7 @@
 
 namespace App\Rules;
 
+use App\Models\ExperienceSkill;
 use App\Models\Pool;
 use App\Models\Skill;
 use App\Models\User;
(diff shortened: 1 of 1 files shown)


esizer (commented):
Clean 🧼 
---
One suggestion that I have which is out of scope is, maybe we could have a more prominent way of displaying which skills or experiences are in error 🤔 It is not entirely obvious and I actually came across this error while testing when I wasn't expecting to because I missed one skill.

esizer (dismissed):

esizer (approved):
LGTM! 👏

PR #8113, opened 2023-10-03
[Fix] Hide user skills in skill dialog, showcase context

### LINKED ISSUE #8088: 🐛 Skills in your library should NOT prompt you for skill level when adding them to the showcase
image

## ✨ Summary

Chat with @substrae for more information

Right now, if you add a skill to your library and set a level and then go to your showcase and add that same skill to your showcase, the skill browser dialog prompts you for the level again (and worse, it doesn't remember your previous selection). On skill showcases, the only time a user should be prompted for skill level and current status in the skill browser dialog is if the skill doesn't yet exist in their skill library.

```[tasklist]
###  ✅ Requirements
- [ ] Ensure that skill level and "current skill" questions (and intro paragraph) don't appear in the dialog if the skill already exists in the user's skill library 
```

esizer:
🤖 Resolves #8088 

## 👋 Introduction

This hides any claimed skills when accessing the skill dialog in the context library.

## 🕵️ Details

The `SkillDialog` had an option to pass skills already in a user's library. This uses that, in combination with the context to prevent the details from appearing when the selected skill is in that array of skills.

## 🧪 Testing

1. Login and confirm a skill you do and do not have in your library
2. Navigate to the appropriate skill show case page
3. Activate "Add a new item" to open the skill dialog
4. First, select a skill you know you do not have, confirm the details section appear
5. Now, select the skill you know you do have, confirm the details do not appear
6. Bonus, do this with the library dialog and confirm the details always appear

### DIFF
--- a/apps/web/src/components/SkillDialog/SkillDialog.tsx
+++ b/apps/web/src/components/SkillDialog/SkillDialog.tsx
@@ -111,7 +111,10 @@ const SkillDialog = ({
     }
   }, [watchSkill, formTrigger]);
 
-  const shouldShowDetails = showDetails(context);
+  const skillInLibrary = !!inLibrary?.find(
+    (librarySkill) => selectedSkill?.id === librarySkill.id,
+  );
+  const shouldShowDetails = showDetails(skillInLibrary, context);
 
   return (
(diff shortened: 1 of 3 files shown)


petertgiles (approved):
Nice.  :+1:

PR #8116, opened 2023-10-03
[Fix] Handle firewall reject characters

### LINKED ISSUE #7812: 🐛Text cannot be saved in skills
## 🐛 Bug

The following pasted text cannot be saved in the skills box:

alongside business requirements . 

(This appears due to unicode characters being blocked by the F5 firewall, making this a prod-only problem)

## 🦋 Expected Behaviour

Unicode character issue. 
Should filter the character mentioned below by Brinda, and notify users about errors. 

## 🕵️ Details

Production only problem, due to firewall. 

## 📋 Steps to Reproduce

1. https://github.com/GCTC-NTGC/gc-digital-talent/issues/7812#issuecomment-1716456084

## 📸 Screenshot
image

## 🙋‍♀️ Proposed Solution

See copy for error message in comment below https://github.com/GCTC-NTGC/gc-digital-talent/issues/7812#issuecomment-1721677719

## ✅ Acceptance Criteria

- [ ] Error message to alert user for troublesome characters, not just the unicode space character, response to the F5 firewall 
- [ ] Toast message upon submitting with troublesome characters 
- [ ] Filter the one character from below and replace with regular non-breaking space, French friendly 

General solution to come later in a separate issue (ask IMTD, whitelist) ~TODO~ #7929 
https://stackoverflow.com/questions/12027137/javascript-trick-for-paste-as-plain-text-in-execcommand

petertgiles:
🤖 Resolves #7812 

## 👋 Introduction

This branch (attempts) to add an error message to help users identify when a request has been rejected due to banned characters.  It also adds a graceful handling for the narrow no-break space.

## 🕵️ Details

Since this behaviour only occurs on the production site behind the production firewall, I added a middleware that can be activated to simulate this behaviour locally.

For the narrow no-break space, I hooked into the whitespace trimming logic that already existed.  I did test that a regular no-break space is OK in production.

## 🧪 Testing

1. Uncomment the `SimulateFirewallMiddleware` class in kernel.php
2. Open a form and add a [narrow no-break space](https://symbl.cc/en/202F/) to an input and submit it.
3. Observe that it was converted to a regular no-break space onBlur.
image
4. Add another character to the banned character list in `SimulateFirewallMiddleware`.
image
5. Attempt to submit a form with that other banned character.
6. Observe that a descriptive toast message appears when the request is rejected.

## 📸 Screenshot

image

### DIFF
--- a/api/app/Http/Kernel.php
+++ b/api/app/Http/Kernel.php
@@ -21,6 +21,7 @@ class Kernel extends HttpKernel
         \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
         \App\Http\Middleware\TrimStrings::class,
         \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
+        // \App\Http\Middleware\SimulateFirewallMiddleware::class,
     ];
 
     /**
--- a/api/app/Http/Middleware/SimulateFirewallMiddleware.php
+++ b/api/app/Http/Middleware/SimulateFirewallMiddleware.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace App\Http\Middleware;
+
+use Closure;
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Log;
+
+/**
+ * Simulates the behavior of the production firewall
+ */
+class SimulateFirewallMiddleware
+{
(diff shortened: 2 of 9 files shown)


esizer (approved):
Sweet! 🍭 Works as expected for me 😄

PR #8121, opened 2023-10-03
[Bug] Skill dialog in skill showcases should save ranking

### LINKED ISSUE #8087: 🐛 Users expect the "Save and add this skill" to... save and add the skill to showcases
image

## ✨ Summary

Chat with @substrae for more information

When adding a skill to a showcase, the skill browser dialog prompts the user to "Save and add this skill", however this leads them to believe that it also saves the addition of the skill to the showcase itself, when this isn't the case. They are currently required to use the "Save and return" button pictured to save the skill showcase, and if they miss that button and leave the page, the showcase is then empty / doesn't reflect their changes.

```[tasklist]
### ✅ Requirements
- [ ] Save the addition of the skill to the showcase upon the click of "Save and add this skill" in the skill browser dialog
```

## 💡 Other thoughts

I think this expectation and behaviour is something we need to keep a keen eye on. It's definitely going to happen elsewhere on the site, if it's not already. In my opinion, we should research and implement navigation interruptions for any page where unsaved data can occur so that users are always prompted to delete the unsaved data before navigating away.

vd1992:
🤖 Resolves #8087

## 👋 Introduction

Updating/claiming a skill in the showcase context will update the ranking after the first mutation completes.

## 🕵️ Details

Elected to add another mutation via function that returns nothing, given the other is tightly coupled to the form it seemed better to not try and adjust the handler to fit two very different shapes. 
Removed the `append` call as updating the ranking reloads the parent component anyway. 

## 🧪 Testing

1. Pick a showcase type, open it up and add a skill
2. Note showcase is updated without requiring another save action on your part

Recording

https://recordit.co/imrrXJUohq

### DIFF
--- a/apps/web/src/pages/Skills/ImproveBehaviouralSkillsPage.tsx
+++ b/apps/web/src/pages/Skills/ImproveBehaviouralSkillsPage.tsx
@@ -141,6 +141,41 @@ const ImproveBehaviouralSkills = ({
       });
   };
 
+  const updateRankingsAfterAddingSkill = (
+    initialSkillRanking: string[],
+    newSkillId: string,
+  ) => {
+    const mergedSkillIds = [...initialSkillRanking, newSkillId];
+    executeMutation({
+      userId: user?.id,
+      userSkillRanking: {
+        improveBehaviouralSkillsRanked: mergedSkillIds,
+      },
+    })
+      .then((res) => {
+        if (res.data) {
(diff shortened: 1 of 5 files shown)


esizer (approved):
LGTM! 👍 

It now has to show a loading spinner but since it affects the currently logged in user, I can't see a way around that 🤷‍♀️

PR #8124, opened 2023-10-04
[Fix] Stabilize team members table

### LINKED ISSUE #8123: 🐛 Team members table infinite re-render
## 🐛 Bug

The actions columns for the team members table is stuck in an infinite re-render loop.

## 🦋 Expected Behaviour

Table should be stable and only re-render when user takes actions.

## 📋 Steps to Reproduce

1. Login as an admin
2. Navigate to `/admin/teams/{teamId}/members`
3. Inspect the actions column
4. Notice it keeps re-rendering

## 📸 Screenshot

Screenshot 2023-10-04 092909

## ✅ Acceptance Criteria

- [ ] Team members table is stable

esizer:
🤖 Resolves #8123 

## 👋 Introduction

This fixes an issue that was causing our team members table to infinitely re-render.

## 🕵️ Details

We were running three separate queries that go similar data. The data was changing causing re-renders to happen at different intervals. This combines all three queries into one and memoizes the data being passed into and used by the table to stabilize it.

## 🧪 Testing

1. Navigate to a team members page `/admin/teams/{teamId}/members`
2. Confirm the table is not re-rendering infinitely


mnigh (approved):
LGTM

PR #8173, opened 2023-10-06
[Feature] Split storybook scripts

### LINKED ISSUE #8128: 🐌 Storybook start times
### Discussed in https://github.com/GCTC-NTGC/gc-digital-talent/discussions/7948

<div type='discussions-op-text'>

<sup>Originally posted by **esizer** September 20, 2023</sup>
## 👋 Introduction

Use of full page stories is impacting startup times for storybook making it less useful for component development.

## 🕵️ Details

We currently load all stories into a single storybook which is helpful in a few ways but it is significantly slowing down the start time for storybook which is not great while developing.

We do not want to stop doing full page stories since they can be useful for design reviews and chromatic diffs.

### With Full Page Stories

Screenshot 2023-09-20 100942

### Only Design System

Screenshot 2023-09-20 100730

## 💡 Idea

We could split out our stories into 3 separate stories.

- Just full page stories (`/apps/web/**/*.stories`)
- Just the design system (`/packages/**/src/**/*.stories`)
- A merging of the previous two

We could have `npm` scripts that run each storybook and then one for chromatic.

- `npm run storybook:app`
- `npm run storybook:design-system`
- `npm run storybook:all`
- `npm run storybook:build`

We could organize this by creating new apps dedicated to storybook.

- `apps/storybook-app`
- `apps/storybook-design-system`
- `apps/storybook-all`</div>

esizer:
🤖 Resolves #8128 

## 👋 Introduction

splits up our storybooks to be more targeted and speed up start up times.

## 🕵️ Details

This adds two new commands to run specific story files. The more story files we load, the longer the start up times. This was growing too large so this splits up storybook into 3 different scripts:

 - `npm run storybook` - Runs all story files
 - `npm run storybook:web` - Runs just story files in `apps/web/**`
 - `npm run storyook:design-system` - Runs just story files in `packages/**`

We also had to refactor the new `santizeString` which was using a function that was not available in storybook.

## 🧪 Testing

1. Run the different commands noted in the details section
2. Confirm the expected story files are loaded
3. Confirm this actually improves start up times
4. Confirm the `sanitizeString` function still works for our textareas and inputs.

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -22,6 +22,8 @@
     "production": "npm run build",
     "build-storybook": "storybook build",
     "storybook": "storybook dev -p 6006",
+    "storybook:design-system": "SB_APP=design storybook dev -p 6006",
+    "storybook:web": "SB_APP=web storybook dev -p 6006",
     "test": "jest --detectOpenHandles --forceExit",
     "test:watch": "jest --watch",
     "tsc": "tsc --project tsconfig.json --noEmit",
--- a/package.json
+++ b/package.json
@@ -35,6 +35,8 @@
     "lint": "turbo run lint",
     "storybook:project": "npm run storybook",
     "storybook": "npm run storybook --workspace=@gc-digital-talent/web",
+    "storybook:design-system": "npm run storybook:design-system --workspace=@gc-digital-talent/web",
+    "storybook:web": "npm run storybook:web --workspace=@gc-digital-talent/web",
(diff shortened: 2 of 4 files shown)


petertgiles (approved):
This is great!  I'm seeing big perf improvements.

Combined: 3:27
Web: 2:03
Design System: 1:08

The sanitize on blur seems to still be working well, too.  :+1:

Issue #8161, opened 2023-10-05
⚙️ Card repeater


petertgiles:
## ⚙️ Summary

Our current repeater component is built to exist inside of forms.  We need a component that allows a display of generic non-form content with reorder, remove, and edit events.

## 🎨 Design File

[Include a link to the design file (if it exists).](https://www.figma.com/file/em5ArPr044YaIe1Vb1zCZz/Pools---Record-of-Decision?type=design&node-id=5437-34317&mode=design)

#### Screenshot

@Jerryescandon 
![image](https://github.com/GCTC-NTGC/gc-digital-talent/assets/8978655/e2c6f43e-44fb-4a51-b71a-73971ce88644)

## Acceptance Criteria
- [ ] Repeats content
- [ ] Has events for reorder, remove, edit
- [ ] No form state
- [ ] Can prevent different events from different items
- [ ] Context message
- [ ] Add button
- [ ] Implement in
  - [ ] ~Skills list in skill showcase~ #9035
  - [ ] Steps list in assessment builder
  - [ ] ~screening questions list in admin pool editor~ #8676

yonikid15:
After talking with the designers, it looks like there was some confusion on why we needed a separate Repeater component. #7916 The assessment plan page does have a form and save button so we can move forward with the regular Repeater with the additional features added there.

brindasasi:
We discussed to close it down as not planned

brindasasi:
we revisited this story and decided this approach would make more sense and easier than updating the repeater comp.

brindasasi:
Is there any accessibility concerns that we need to take care for the locked cards?  
Should we have the screen reader to read out about the locked card ? 
@GCTC-NTGC/designers

esizer:
I think we may need to re-estimate this one now that we have more context on it. The original ticket ( #6242 ) to create a repeater was based on the immense amount of heavy lifting `react-hook-form` was doing for us and that was a 5 😅 

Now that we need to write all that heavy lifting ourselves and add additional functionality, I'm thinking this needs to be higher than 3 and probably also higher than a 5 since it is more work than the original ticket 😬 

There will be a lot of accessibility concerns involved here as well:

- state management
- focus management
- keyboard shortcuts
- AT announcements

petertgiles:
@esizer if that's the case then maybe it's better to make the customizations to the regular repeater after all?  And just accept the weirdness that comes from leaning on react-hook-form to do that heavy lifting.

[...]

PR #8181, opened 2023-10-11
[Debt] Updates to Repeater component

### LINKED ISSUE #7928: 🏗️ Update the repeater component
image

## ✨ Summary

[Visit the Figma file](https://www.figma.com/file/guHeIIh8dqFVCks310Wv0G/Style-library?type=design&node-id=5-24&mode=design) • Contact @substrae with design questions

This update to the repeater component is aimed at making the element more legible, accessible, and usable in more contexts. This update is mostly a visual refresh, with one or two enhancements that provide the component with more flexibility. Please see the details section for more information.

```[tasklist]
## ✅ Requirements
- [x] Rearrange the repeater item controls to be above the content, included in the same card
- [x] Update the up/down icons to `arrow-small-up` and `arrow-small-down`
- [x] Update the up/down controls' disabled state to show a • character and disable any interactivity (hover, underlines, focus, etc.)
- [x] Add a bar to the top of the card that uses `base(secondary)` or `base(error)` when invalid
- [x] Ensure all controls have updated targetable whitespace around them
- [x] Ensure all controls have updated focus states
- [x] Ensure the up/down arrows are given the proper state when in the first/last/single item contexts
- [ ] Add an optional "edit" control using the `pencil-square` Heroicon (Moved to #8161)
- [x] Add a "locked" state that prevents the item from being moved up or down within the list (indicated by the `lock-closed` Heroicon to the left of the item's number)
- [x] Ensure the card content's whitespace matches the design
- [x] Update the whitespace between repeater items (but not between the button and the list of items) to `x.25`
- [ ] Update the button's border opacity and colors to `data-h2-border-color="base(gray.darker.30)"` when disabled
- [ ] Update the button's border opacity and colors to `data-h2-border-color="base(secondary.darker.30)"` when clickable
- [x] Add an optional, customizable "near limit" message well using the `warning` variant
- [x] Add an `error` well for validation errors and ensure that invalid items in the list are highlighted
- [x] Ensure error summaries in form contexts can link to the repeater list properly
``` 

### Control updates

A few issues we're hoping to solve:
- Eliminate awkward whitespace
- Provide larger clickable/tappable area for buttons
- Add the optional ability to include an edit control
- Create a better sense of division between repeater items
- Reduce height for shorter, single-line repeated items

The proposed solution moves individual repeated item controls above the item itself, creating a header. This layout is more flexible and reflects the mobile experience more consistently but putting the controls inline with each other, separating the order controls from other content controls, and eliminating the whitespace below the existing controls in the current design. The header also now provides a visual break between repeated items, creating a much stronger delineation of information.

<details>
  <summary>📸 General design changes</summary>
  
</details> 

<details>
  <summary>📸 Whitespace updates</summary>
  
</details> 

<details>
  <summary>📸 Locked state</summary>
  
</details> 

<details>
  <summary>📸 Single-line example</summary>
  
</details> 

<details>
  <summary>📸 Focus examples</summary>
  
</details> 

### New states

As a way to partially address #7888, repeater lists now have a few optional states.

The first state provides a unique and contextualized message (so make sure this is editable when importing the component) that explains the limit and that it is almost reached.

<details>
  <summary>📸 Approaching limit state</summary>
  
</details> 

The second state offers an error well that addresses errors with the list as a unit, as opposed to specific form errors. Items in the repeater are also now provided with an error state to make it clear where there is a problem. We'll also need the list to have an ID that can be linked to, and possibly a name, so that error summaries in form contexts have a place to point to.

<details>
  <summary>📸 Error state</summary>
  
</details> 

### Button updates

The "Add item" button has been modified ever-so-slightly to be less visually demanding, as it was competing with the rest of the content around it. This has been accomplished by reducing the border's opacity.

<details>
  <summary>📸 General design changes</summary>
  
</details> 

<details>
  <summary>📸 Focus examples</summary>
  
</details>

yonikid15:
🤖 Resolves #7928 

## 👋 Introduction

- Adds [new updates (view A/C)](https://github.com/GCTC-NTGC/gc-digital-talent/issues/7928) to the `<Repeater />` component.

## 🕵️ Details

I would recommend testing on storybook to view all variations. As well as the Edit pool page screening questions, and the showcase skills sub pages (e.g. TopBehaviouralSkillsPage)

## 🧪 Testing

1. Run `npm run storybook` and view Repeater stories
2. Go to Edit pool page: screening questions section and ensure everything works correctly.
3. Go to Showcase skills subpages and ensure everything works correctly.


packages/forms/src/components/Repeater/Repeater.tsx
```diff
@@ -145,95 +178,143 @@ const Fieldset = ({
...
+                      decrement
+                      animate={!shouldReduceMotion}
+                      aria-label={intl.formatMessage(
+                        formMessages.repeaterMove,
+                        {
+                          from: position,
+                          to: position - 1,
+                        },
+                      )}
+                      data-h2-font-weight="base(700)"
```
esizer, line 218: I think all the buttons are bold so adding this to the `ActionButton` might be a way to avoid repeating ourselves.

packages/forms/src/components/Repeater/Repeater.tsx
```diff
@@ -145,95 +178,143 @@ const Fieldset = ({
...
+                        </div>
+                      )}
+                    </ActionButton>
+                    {!hideIndex && (
+                      <span
+                        aria-hidden="true"
+                        data-h2-text-align="base(center)"
+                        data-h2-font-weight="base(700)"
+                        data-h2-margin="base(x.25, 0)"
+                      >
```
esizer, line 238: I'd add some padding here or use the gap property on the parent element to help spread some items out a big. They seem a little bit close right now.

Screenshot 2023-10-13 120150

packages/forms/src/components/Repeater/Repeater.tsx
```diff
@@ -290,7 +417,59 @@ const Root = ({
...
+            </>
+          )}
+        </Well>
+      )}
+      {hasError && (
+        <Field.Error id={name} data-h2-padding="base(0)">
+          <Well
+            data-h2-text-align="base(center)"
+            data-h2-margin-bottom="base(x1)"
+          >
```
esizer, line 455: Should be safe to drop this well and move the `text-align` to the error element dropping the `padding: 0;` at the same time. This should avoid the overlapping blue/red borders.

Screenshot 2023-10-13 120434

esizer (changes requested):
Looking good so far, noticed some small styling things.

packages/forms/src/components/Repeater/Repeater.tsx
```diff
@@ -89,12 +97,37 @@ const Fieldset = ({
...
+
   // Non-zero index position of the fieldset for humans
   const position = index + 1;
+  const isFirstItem = disabled || index <= 0;
+  const isLastItem = disabled || index === total - 1;
+
+  const isLocked = numOfLockedItems ? index + 1 <= numOfLockedItems : false;
+  const previousItemLocked = numOfLockedItems
+    ? numOfLockedItems + 1 === index + 1
+    : false;
```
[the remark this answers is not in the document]
yonikid15: Well I'm assuming were only locking things that take priority which would usually mean they would start at the top. Also my logic was you can't lock on item on an index 3 if the other 2 don't exist, and if there are two items above the 3rd locked item then they would probably be locked as well. However, I'm happy to hear your thoughts on it.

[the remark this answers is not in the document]
esizer: Okay, I guess maybe I don't really understand the locking part 😕 . If it has to be the first X items would we not just make that not part of the repeater? 🤔 My understanding would be a user could select to lock the position of any item. Take the following scenario,

User has a list of 5 items, they want to ensure one is always in position 3 so they "lock" it. They then move item 2 down which would make it skip over 3 and put it in position 4 and move position 4 up to 2.

packages/forms/src/components/Repeater/Repeater.tsx
yonikid15: Nice catch! That's new, I must have missed it with last changes. Fixed by [e51e6ab](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8181/commits/e51e6ab11f1085af2d7a5fa68c9f549426c83db3).

packages/forms/src/components/Repeater/Repeater.tsx
yonikid15: Sounds good 👍 Fixed by [e51e6ab](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8181/commits/e51e6ab11f1085af2d7a5fa68c9f549426c83db3).

packages/forms/src/components/Repeater/Repeater.tsx
yonikid15: Good point 😆 Fixed by [e51e6ab](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8181/commits/e51e6ab11f1085af2d7a5fa68c9f549426c83db3).

packages/forms/src/components/Repeater/Repeater.tsx
```diff
@@ -145,95 +178,143 @@ const Fieldset = ({
...
             <div
               data-h2-display="base(flex)"
-              data-h2-radius="base(s)"
-              data-h2-shadow="base(medium)"
-              data-h2-flex-direction="base(column)"
-              data-h2-align-items="base(center)"
-              data-h2-overflow="base(hidden)"
+              data-h2-justify-content="base(space-between)"
+              data-h2-width="base(100%)"
+              data-h2-margin-bottom="base(x.5)"
```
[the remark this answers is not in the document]
yonikid15: Okay 👍 Fixed by [e51e6ab](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8181/commits/e51e6ab11f1085af2d7a5fa68c9f549426c83db3).

packages/forms/src/components/Repeater/Repeater.tsx
```diff
@@ -145,95 +178,143 @@ const Fieldset = ({
...
-                >
-                  {legend}
-                </p>
-              )
-            }
-            {children}
-          </div>
-          <div
-            data-h2-flex-shrink="base(0)"
+            data-h2-margin="base(x.5 x1 x1 x1)"
```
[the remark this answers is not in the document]
yonikid15: Absolutely, Fixed by [e51e6ab](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8181/commits/e51e6ab11f1085af2d7a5fa68c9f549426c83db3).

packages/forms/src/components/Repeater/Repeater.tsx
```diff
@@ -145,95 +178,143 @@ const Fieldset = ({
...
+                      {...(isFirstItem
+                        ? { "data-h2-padding-left": "base(x.25)" }
+                        : {})}
+                    >
+                      {!isFirstItem && !previousItemLocked ? (
+                        <ArrowUpIcon data-h2-width="base(x.75)" />
+                      ) : (
+                        <div data-h2-width="base(x.75)">
+                          <p data-h2-padding-bottom="base(x.25)">.</p>
+                        </div>
```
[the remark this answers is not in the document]
yonikid15: Nice catch 👍 Fixed by [e51e6ab](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8181/commits/e51e6ab11f1085af2d7a5fa68c9f549426c83db3).

packages/forms/src/components/Repeater/Repeater.tsx
```diff
@@ -145,95 +178,143 @@ const Fieldset = ({
...
-                disabled={disabled || index === total - 1}
-                onClick={increment}
-                animate={!shouldReduceMotion}
-                aria-label={intl.formatMessage(formMessages.repeaterMove, {
-                  from: position,
-                  to: position + 1,
-                })}
+                {!isLocked ? (
+                  <>
+                    <ActionButton
```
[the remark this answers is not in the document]
yonikid15: Yea I missed that thanks, fixed by [4c3d573](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8181/commits/4c3d5733a34753ad7a7e517e9a430580a9fa25cb)

packages/forms/src/components/Repeater/Repeater.tsx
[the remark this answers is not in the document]
esizer: Sorry, I also meant the `div` in the code since it is also a block level element that shouldn't appear as a child of a `button`.

packages/forms/src/components/Repeater/Repeater.tsx
esizer: Looks great! Sorry, I also noticed we don't have a period, should we add one?

packages/forms/src/components/Repeater/Repeater.tsx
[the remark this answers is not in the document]
esizer: @substrae Seems as though we are having trouble understanding exactly how the lock feature should be working. Could you help us out?

packages/forms/src/components/Repeater/Repeater.tsx
yonikid15: Fixed by [6f56faa](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8181/commits/6f56faa8542558ac5c339179a8b4fe76b708c53a)

packages/forms/src/components/Repeater/Repeater.tsx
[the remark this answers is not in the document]
yonikid15: Fixed by [02dd9b6](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8181/commits/02dd9b68e2389ecb787cd08c3872d333901f3950)

packages/forms/src/components/Repeater/Repeater.tsx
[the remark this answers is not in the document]
substrae: @yonikid15 @esizer 

Ahh, sorry, this was a miscommunication on our part. The locking feature was implemented as a part of Jerbo's assessment record work, so it probably hasn't been fully explained. The intent here is for the system to be able to lock items, not the user. The reason this feature exists is because the locked items are still cognitively part of the list, but their order is uniquely constrained due to external factors. Locked items will, to my knowledge, always be at the top of the list, but they might not be added in that order.

For example, in Jerbo's design, you have a repeater of assessment types. The first type is **always** "Application", so that guy is locked in the first spot on page load. The user can then proceed to add other assessment types. If, however, they choose to add "Screening questions at the time of application", that assessment gets added and locked into the second spot, regardless of which place in the order they added it it.

After speaking with Jerbo, we don't foresee any need to lock items at specific indices, though it might be helpful down the road to reverse this and be able to lock items to the bottom of the list.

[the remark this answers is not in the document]
esizer: So this is part of the AC for the [form/card repeater](https://github.com/GCTC-NTGC/gc-digital-talent/issues/8161) and not the fieldset repeater? Or do both need the locking feature?

esizer:
I'm not sure this is in scope but the number of wells we have yelling at the user about number of items and saving seems excessive 😅 Also, it looks a little odd when one is center aligned and the others are left.

Screenshot 2023-10-17 141727

packages/forms/src/components/Repeater/Repeater.tsx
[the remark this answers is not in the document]
substrae: Are the not the same component with different children passed to them 🤔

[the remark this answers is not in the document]
esizer: No, they need to be separate because one represents repeating forms while the other one is repeating fieldsets so we can't re-use the same component 👎

yonikid15:
> I'm not sure this is in scope but the number of wells we have yelling at the user about number of items and saving seems excessive 😅 Also, it looks a little odd when one is center aligned and the others are left.
> 

Nice catch, that second well below the add button shouldn't be there! As for the total number of wells yelling that's a design question haha.

packages/forms/src/components/Repeater/Repeater.tsx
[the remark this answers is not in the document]
yonikid15: So the locking feature is only necessary for both components or just the CardRepeater? From the designs I've seen the locking feature only seems to be present in the soon to be built CardRepeater.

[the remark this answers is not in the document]
substrae: Okay, I'm a bit confused, or maybe I'm misunderstanding. The presentation layer and the functionality of the repeater component are generally indifferent to the context they exist in:
- look and feel of the list, cards, errors, add item button, etc.
- the functionality of moving an item up or down
- the functionality of locking an item in place
- the functionality to set a maximum or minimum number of items
- the functionality to accept child content for each card

The child content is passed when the repeater is imported into the UI, and how that content is handled (submission, etc.) should be handled by contextual variants, whether through props or other means.

If I was going to build this, it would look something like:
- Repeater
  - RepeaterCoreAndStyles
  - RepeaterDataHandling
    - FieldsetVariant
    - StaticVariant

To me this is more scaleable and reduces repetition by centralizing the majority of the interaction and UI while providing extensible options based on the context. Unless the intent is to break up the styles/core functionality and have your two unique components import them? Even then, that seems like it would be hard to follow for new devs.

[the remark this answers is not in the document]
esizer: The issue is, one is a form (or at least appears to be) and one is not (currently it is a `fieldset`) so it is impossible to make them the same component since the underlying HTML and functionality would be completly different even though it appears to be similar in the design. 

We are leveraging [`useFieldArray`](https://react-hook-form.com/docs/usefieldarray) from `react-hook-form` to handle things like inserting, removing and moving items. It is required to be used inside of a form. Since the new use case is not a repeatable `fieldset` being used in a form and instead a form itself, we can't use this component and the best we can do is share a few of the styles between them.

We need will need to create a component to repeat forms that emit an `onSave` event with the ability to toggle between a display and edit mode and act as it's own independent item.

Unless you are suggesting that the new changes are not forms, cannot be saved individually and instead are saved by a single submit handler.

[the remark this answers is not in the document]
substrae: Hmmm... maybe @Jerryescandon should chime in here, but I feel like they will always be a `fieldset`, because you're saving the list wrapper's form as a whole in order to save their order.

Jerryescandon:
> I'm not sure this is in scope but the number of wells we have yelling at the user about number of items and saving seems excessive 😅 Also, it looks a little odd when one is center aligned and the others are left.

I dont think a limit warning is necessary for the screening questions. We really only wanted to add a limit warning for the number of assessments.

Also, I think only one "unsaved changes" warning should suffice for the entire form, we don't need one for the fields and one for the repeater. 

(I'm assuming this is all inside the dialog right?)

yonikid15:
@esizer Okay, it's ready for a re-review. Unfortunately, I wasn't able to improve the lock feature, I left it so that you can lock the first `n` amount of items. However, this is blocking other issues. So I can make a new issue to make improvements to the locked feature, since it's not required for this PR.

apps/web/src/pages/Skills/components/UpdateSkillShowcase.tsx
```diff
@@ -309,84 +278,57 @@ const UpdateSkillShowcase = ({
...
-                              <p>
+                      {fields.map((item, index) => (
+                        <Repeater.Fieldset
+                          key={item.id}
+                          name="userSkills"
+                          index={index}
+                          total={fields.length}
+                          onMove={move}
+                          onRemove={remove}
+                          legend={
```
esizer, line 289: Should we create an issue to apply the `onSave` function to this component so users can update their skills here?

packages/forms/src/components/Repeater/Repeater.tsx
```diff
@@ -74,13 +78,22 @@ export interface RepeaterFieldsetProps {
...
   disabled?: boolean;
+  /** Locks item position (removes append and decrement actions) */
+  numOfLockedItems?: number;
   children: React.ReactNode;
+  /** Callback function when this item's index is changed' */
+  onMove: (from: number, to: number) => void;
+  /** Callback when the item is removed from the array */
+  onRemove: (index: number) => void;
+  /** Callback function when edit button is clicked */
+  onSave?: () => void;
```
esizer, line 89: I think it would be helpful to pass the index here as well. Since this (at least to start) will be opening a dialog with a form, it would be useful to get the appropriate info from the index.

packages/forms/src/components/Repeater/Repeater.tsx
```diff
@@ -40,12 +44,14 @@ const ActionButton = ({
...
+      data-h2-radius="base(50%)"
       data-h2-cursor="base(pointer)"
       data-h2-display="base(flex)"
       data-h2-align-items="base(center)"
       data-h2-padding="base(x.5)"
       data-h2-background-color="base(foreground) base:hover(gray.lightest) base:focus(focus)"
       data-h2-transition="base:children[svg](transform 200ms ease)"
+      data-h2-font-weight="base(700)"
       {...(disabled
         ? { disabled: true, "data-h2-color": "base(black.lightest)" }
```
esizer, line 56: We may also want to move the background color rules here. It looks a little odd when hovering disabled items due to the shape.

Screenshot 2023-10-23 104118

packages/forms/src/components/Repeater/Repeater.tsx
```diff
@@ -145,98 +184,158 @@ const Fieldset = ({
...
+                      decrement
+                      animate={!shouldReduceMotion}
+                      aria-label={intl.formatMessage(
+                        formMessages.repeaterMove,
+                        {
+                          from: position,
+                          to: position - 1,
+                        },
+                      )}
+                      data-h2-margin="base(0, x.25, 0, 0)"
```
esizer, line 229: Use of `gap` makes this unnecessary. I think we can remove most of the margin attributes now.

packages/forms/src/components/Repeater/Repeater.tsx
```diff
@@ -145,98 +184,158 @@ const Fieldset = ({
...
+                      {...(isFirstItem
+                        ? { "data-h2-padding-left": "base(x.25)" }
```
esizer, line 231: This is causing the button to have an odd shape. Could we use negative margin instead? Also, we need to determine the first unlocked item and apply it there as well.

Screenshot 2023-10-23 105306

packages/forms/src/components/Repeater/Repeater.tsx
```diff
@@ -145,98 +184,158 @@ const Fieldset = ({
...
+                          &bull;
+                        </span>
+                      )}
+                    </ActionButton>
+                    {!hideIndex && (
+                      <span
+                        aria-hidden="true"
+                        data-h2-text-align="base(center)"
+                        data-h2-font-weight="base(700)"
+                        data-h2-margin="base(x.25, 0)"
```
esizer, line 247: This margin could be removed as well

packages/forms/src/components/Repeater/Repeater.tsx
```diff
@@ -145,98 +184,158 @@ const Fieldset = ({
...
+                      onClick={increment}
+                      animate={!shouldReduceMotion}
+                      aria-label={intl.formatMessage(
+                        formMessages.repeaterMove,
+                        {
+                          from: position,
+                          to: position + 1,
+                        },
+                      )}
+                      data-h2-margin="base(0, x.25)"
```
esizer, line 263: Remove margin

packages/forms/src/components/Repeater/Repeater.tsx
```diff
@@ -145,98 +184,158 @@ const Fieldset = ({
...
+                        <span data-h2-width="base(x.75)" aria-hidden>
+                          &bull;
+                        </span>
+                      )}
+                    </ActionButton>
+                  </>
+                ) : (
+                  <>
+                    <LockClosedIcon
+                      data-h2-margin-left="base(x.5)"
```
esizer, line 277: I don't think this is necessary anymore either.

packages/forms/src/components/Repeater/Repeater.tsx
```diff
@@ -145,98 +184,158 @@ const Fieldset = ({
...
+                      data-h2-width="base(x.75)"
+                    />
+                    {!hideIndex && (
+                      <span
+                        aria-hidden="true"
+                        data-h2-text-align="base(center)"
+                        data-h2-font-weight="base(700)"
+                        data-h2-margin="base(x.25, 0)"
```
esizer, line 285: Should be safe to remove this as well.

apps/web/src/pages/Skills/components/UpdateSkillShowcase.tsx
yonikid15: Good idea! Added by [9bf4f0c](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8181/commits/9bf4f0c502671f42faec53630901ab7f55105235)

packages/forms/src/components/Repeater/Repeater.tsx
yonikid15: Fixed by [352db51](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8181/commits/352db510147821f32c820480e8561299b3f24f1c)

packages/forms/src/components/Repeater/Repeater.tsx
yonikid15: Nice catch. Fixed by [83cffcb](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8181/commits/83cffcb65189d8cfbe5e36e86e62495e1e28afc2)

packages/forms/src/components/Repeater/Repeater.tsx
yonikid15: Fixed by [83cffcb](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8181/commits/83cffcb65189d8cfbe5e36e86e62495e1e28afc2)

packages/forms/src/components/Repeater/Repeater.tsx
yonikid15: Hmm, I'm not seeing that on my end.

packages/forms/src/components/Repeater/Repeater.tsx
yonikid15: Fixed by [83cffcb](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8181/commits/83cffcb65189d8cfbe5e36e86e62495e1e28afc2).

packages/forms/src/components/Repeater/Repeater.tsx
yonikid15: Fixed by [83cffcb](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8181/commits/83cffcb65189d8cfbe5e36e86e62495e1e28afc2).

packages/forms/src/components/Repeater/Repeater.tsx
yonikid15: Removing this caused it to push the icon to the left edge on my end. Can you check again one more time and see if it's occurring on your end.

packages/forms/src/components/Repeater/Repeater.tsx
yonikid15: Fixed by [83cffcb](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8181/commits/83cffcb65189d8cfbe5e36e86e62495e1e28afc2).

packages/forms/src/components/Repeater/Repeater.tsx
esizer: Oh yup, you are right. I didn't notice the negative margin being used up the tree 👍

packages/forms/src/components/Repeater/Repeater.tsx
esizer: Oh, you can only notice it on hover but now that is gone you can't notice it anymore 👌

apps/web/src/pages/Skills/components/UpdateSkillShowcase.tsx
esizer: Huh, okay so this sounded like a good idea but I think it may be out of scope. It works however, the create mutation is still run so instead of editing the existing one you end up creating a new one.

Sorry, since this is higher priority I'd say just back this out for now and we can add it in another issue.

yonikid15: Let's just back this out then since this is high priority I can make an issue!

yonikid15: Done! [2c00151](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8181/commits/2c00151b2eaab2358759d1385cf6c557405382b1)

esizer (approved):
Awesome! 🌟

PR #8183, opened 2023-10-11
[Feature] Inline skill browser

### LINKED ISSUE #7112: ✨ Create inline version of `SkillDialog` component
We want the list of skills that appears on the Search page to **NOT appear by default**, we want it to appear only if:

A) The two dropdowns have have been filled out first, or
B) The user targets the list (dropdown, searchable) and starts typing

There has been some work done around this in issue #5054 but it's not there yet. Additionally, #5054 is meant to be a dialog, while this would include all 3 fields as part of the form.

### CURRRENT

<details><summary>Screenshot of current implementation</summary>
<p>

image

</p>
</details> 

### IDEAL 

<details><summary>Screenshot of ideal implementation</summary>
<p>

image

</p>
</details> 

## Acceptance Criteria

- [ ] rename `SkillDialog` component to `SkillBrowserDialog`
- [ ] create new inline version of component `SkillBrowserDialog` outside of a Dialog
- [ ] form not updated in current `SkillDialog` component, factor out contents
- [ ] create new storybook story for inline component

esizer:
🤖 Resolves #7112 

## 👋 Introduction

Creates an inline version of the skill browser.

## 🕵️ Details

This renames and refactors the existing `SkillDialog` component. It was renamed to `SkillBrowserDialog` to be more explicit. As well, some common functions were moved to a utility file so they can be reused by the inline version. Since the inline version is meant to be used within a form rather than be a form itself, it can accept rules unlike the dialog version that has predetermined rules.

## 🧪 Testing

> **Note**
> You can also test this on the search page with with the fallowing changes:

<details>

<summary>Skill Browser - Search Page Diff</summary>

```sh
diff --git a/apps/web/src/pages/SearchRequests/SearchPage/components/SearchForm.tsx b/apps/web/src/pages/SearchRequests/SearchPage/components/SearchForm.tsx
index 68984237e..5b0ded891 100644
--- a/apps/web/src/pages/SearchRequests/SearchPage/components/SearchForm.tsx
+++ b/apps/web/src/pages/SearchRequests/SearchPage/components/SearchForm.tsx
@@ -43,6 +43,7 @@ import {
 import AdvancedFilters from "./AdvancedFilters";
 import AddSkillsToFilter from "./AddSkillsToFilter";
 import FilterBlock from "./FilterBlock";
+import SkillBrowser from "../../../../components/SkillBrowser/SkillBrowser";
 
 const positionDurationToEmploymentDuration = (
   durations: Maybe<PositionDuration>[],
@@ -392,7 +393,7 @@ const SearchForm = React.forwardRef<SearchFormRef, SearchFormProps>(
               trackUnsaved={false}
             />
           </FilterBlock>
-          <AddSkillsToFilter allSkills={skills ?? []} linkId="skillFilter" />
+          <SkillBrowser skills={skills ?? []} name="skills" />
           <FilterBlock
             id="workingLanguageFilter"
```

</details>

2. Navigate to the `SkillBrowser` story
3. Interact with the inline version confirming it submits the expected data
4. Navigate to an existing skill dialog
5. Confirm it still functions as expected

## 📸 Screenshot

>

Screenshot 2023-10-11 130425

### DIFF
--- a/apps/web/src/components/SkillBrowser/utils.ts
+++ b/apps/web/src/components/SkillBrowser/utils.ts
@@ -213,18 +213,20 @@ export const getFilteredSkills: GetFilteredSkills = ({
 }) => {
   if (inLibrary && family && family === "library") {
     // If `inLibrary` was passed and selected, filter by those instead of family
-    return skills.filter((currentSkill) =>
-      inLibrary?.find(
-        (skillInLibrary) => skillInLibrary.id === currentSkill.id,
-      ),
+    return skills.filter(
+      (currentSkill) =>
+        inLibrary?.find(
+          (skillInLibrary) => skillInLibrary.id === currentSkill.id,
+        ),
     );
   }
 
   if (family && family !== "all") {
     // We only care about family if it is set
     // since we are filtering families by category
-    return skills.filter((currentSkill) =>
-      currentSkill.families?.some((skillFamily) => skillFamily.id === family),
+    return skills.filter(
+      (currentSkill) =>
+        currentSkill.families?.some((skillFamily) => skillFamily.id === family),
     );
   }
   if (category && category !== "all") {


vd1992:
@Jerryescandon 

Question, the issue body has the design using filtering by work streams. 
Since I don't think we can do anything like that, I just wanted to ensure that design-wise we are on the same page. That the filtering works as posted above in Eric's screenshot and everyone is on the same page.

image

Jerryescandon:
> @Jerryescandon
> 
> Question, the issue body has the design using filtering by work streams. Since I don't think we can do anything like that, I just wanted to ensure that design-wise we are on the same page. That the filtering works as posted above in Eric's screenshot and everyone is on the same page.
> 

Sounds good! I'll update the design for this. 

(Unrelated, there a spelling mistake on that screenshot: *nest > best)

apps/web/src/components/SkillBrowser/SkillBrowser.tsx
```diff
@@ -0,0 +1,146 @@
...
+          ? {
+              "data-h2-grid-template-columns": "base(1fr) p-tablet(1fr 1fr)",
+            }
+          : {
+              "data-h2-grid-template-columns": "base(1fr)",
+            })}
+      >
+        {showCategory && (
+          <Select
+            id={inputNames.category}
```
petertgiles, line 83: What's the reason for the dynamic IDs?  Is this going to make it tricky to do anything with the family or category in the save handler?
image

petertgiles (approved):
Seems good.  Nice one!  :+1:

apps/web/src/components/SkillBrowser/SkillBrowser.tsx
esizer: Oh, this is so we can use multiple on the same page. Yeah, it likely would have an impact if we actually wanted to reference these in the form submission. Though, I believe that we have no plans on doing this and these are just filters for the skill combobox.

@Jerryescandon do you know if we ever need to submit the category or family with this input when it is used?

yonikid15 (approved):
Looks great! Nice work 👍 

I have one suggestion that's out of scope, but I think a simple improvement would be moving the combo box options a bit lower so the user can see the list of pills getting added as they go down the list @Jerryescandon. However, that's a design suggestion and out of scope.

image

esizer:
> I have one suggestion that's out of scope, but I think a simple improvement would be moving the combo box options a bit lower so the user can see the list of pills getting added as they go down the list @Jerryescandon. However, that's a design suggestion and again out of scope.

The only problem with that is that the chip height is dynamic based on how many items you select and if you add say 30 skills that dropdown could be fairly far from the actual input 😅

yonikid15:
> The only problem with that is that the chip height is dynamic based on how many items you select and if you add say 30 skills that dropdown could be fairly far from the actual input 😅

True... maybe adding them above the select input then? Either way.. just a suggestion.

PR #8215, opened 2023-10-12
[Fix] Editable behavioral skills

esizer:
## 👋 Introduction

Fixes a bug where admins could edit a users top behavioral skills.


brindasasi (approved):
Thanks Eric! just confirmed its not api issue.. its only ui.. it was actually opening admin’s own skill .. hence the ui fix is sufficient for this

brindasasi:
I meant when I try to edit a user's skill via api , it didn't allow me.

PR #8226, opened 2023-10-13
[Feature] Automatic assessment step handling

### LINKED ISSUE #8034: ✨ Automatic handling of certain assessment step types
## ✨ Feature

Some assessment step types need handling automatically rather than be client driven. 

## 🕵️ Details

Creation of a draft pool should trigger the creation of an assessment step of type `APPLICATION_SCREENING`
This assessment cannot be destroyed unless the pool is deleted, syncs with all technical PoolSkills, and is locked to `sortOrder = 1`
This must be dynamic as the associated skills will change after this step is created. 

If screening questions are added to a pool, an assessment step must occupy `sortOrder = 2`  with type `SCREENING_QUESTIONS_AT_APPLICATION`
If screening questions are deleted so none exist, their assessment step record must be deleted. 

## 🎨 Design File

### 🧑‍🎨 Designer

## 📸 Screenshot

Laravel listener on `setEssentialSkills` and `setNonessentialSkills` methods that triggers re-syncing of pool skills to assessment step application screening. 

Boot methods running on creation and deletion of `ScreeningQuestion` models that check if the appropriate assessment step exists, and create/delete `AssessmentStep` as needed. 
ScreeningQuestion created -> check if assessment step exists -> create or do nothing
ScreeningQuestion deleted -> check if parent pool has more than zero screening question remaining -> delete assessment step if zero

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Creating a pool creates an application screening assessment
- [ ] Adding/deleting screening questions automatically updates the screening question assessment
- [ ] Screening questions can be added, with skills, in the frontend, too

JamesHuf:
🤖 Resolves #8034

## 👋 Introduction

This automatically creates/deletes a few assessment types in certain situations.

## 🧪 Testing

1. Create a new pool and confirm a new application screening assessment step is created for it.
2. Add some essential/nonessential skills to the pool and confirm the application screening assessment step is synced to all technical skills on the pool.
3. Add a screening question for the pool and confirm a screening question assessment is created for the pool.
4. Remove all the screening questions from the pool and confirm the screening question assessment is deleted as well.


brindasasi (changes requested):
`Remove all the screening questions from the pool and confirm the screening question assessment is deleted as well.`
This doesn't work properly if we shuffle the questions order before deleting. 

Root cause of this is 
It is creating a duplicate screening_questions record during the shuffle with the new value instead of updating it. 
Hence even if we delete all  the questions on the front end, there is still some stale questions on the back end which prevents your logic to delete the assessment step to work. 

Worst case is when you refresh the page now , all deleted questions does appear,  some times it is more than 3 on the front end. 

Possible fix would be fixing the existing bug of creating duplicate records while shuffling the questions order.

JamesHuf:
I fixed the bug causing duplicate questions to be created with [Fixed bug in screening questions form](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8226/commits/2b70a6b76faf8c33d06786849135adfae9be6561).

brindasasi (approved):
wonderful! works great 🥳

PR #8229, opened 2023-10-16
[Debt] Pause row selection queries

### LINKED ISSUE #7681: Pause "Selected" queries for tables unless at least one row selected.
# Description

Our tables which have selectable rows issue a request to the API even before a single row is selected.  This query should probably be paused until a row is selected.

# Context

              This probably has nothing to do with your PR and you can ignore it.  :laughing:  It makes me crazy that this table makes a search request when there is nothing selected.
image

_Originally posted by @petertgiles in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7588#discussion_r1302122700_
            
# Acceptance Criteria
- [ ] Tables with selectable rows do not issue a "selected items" query until at least one row is selected.

esizer:
🤖 Resolves #7681 

## 👋 Introduction

Pauses the row selection queries until users make at least one selection.

## 🕵️ Details

This was slightly tricker than I first assumed. I used a ref to track if any selection has been made so we still run the query once all items have been selected so we can empty out the in memory cache of the result. Maybe a better approach would be to try and empty that in memory cache instead of making a trip to the server when we know the response will be an empty array? 🤔 

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to `/admin/users`
3. Confirm the initial selected query is not run
4. Select a row
5. Confirm the query runs and printing works
6. Deselect a row
7. Confirm print is empty
8. Navigate to a pool candidates table
9. Repeat steps 4-7

### DIFF
--- a/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
+++ b/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useState } from "react";
+import React, { useEffect, useMemo } from "react";
 import { IntlShape, useIntl } from "react-intl";
 import LockClosedIcon from "@heroicons/react/24/solid/LockClosedIcon";
 import { SubmitHandler } from "react-hook-form";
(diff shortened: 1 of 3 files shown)


vd1992 (approved):
Seems to work well, no issues. 

Selecting a row then deselecting a row still sends an empty request. But I don't think it is a big deal, from a performance perspective, it triggering an API call with each selection seems the bigger fish. To elaborate, if you want to pick 5 people, that is 5 clicks and 5 requests.

PR #8234, opened 2023-10-16
[Debt] Server root default value

### LINKED ISSUE #2679: BearerTokenServiceProvider should not error due to missing env vars
During the build process, service providers may be registered (during `composer install` command). Our BearerTokenServiceProvider throws an error if oauth env vars aren't defined. This is happening on the azure server.

I think the simplest way to fix this is set default values in config files, even if those values are empty strings.

vd1992:
🤖 Resolves #2679

## 👋 Introduction

Just clearing a small issue from way back using the solution in the issue body, thanks Matt. 

## 🧪 Testing

1. ?  (Not sure)

### DIFF
--- a/api/config/oauth.php
+++ b/api/config/oauth.php
@@ -12,7 +12,7 @@ return [
     | further configuration.  Should not end in a trailing slash.
     |
     */
-    'server_root' => env('AUTH_SERVER_ROOT'),
+    'server_root' => env('AUTH_SERVER_ROOT', ''),
 
     /*
     |--------------------------------------------------------------------------


tristan-orourke:
@petertgiles, no that's the error I expect to see after this change, because it tries to connect to an invalid url. What I didn't expect was to get the same error, without the default empty string. For some reason, I thought that string concatenation in `BearerTokenServiceProvider`, line 21, would throw an exception if the env variable was left null. But it turns out nulls just convert to empty strings https://www.php.net/manual/en/language.types.string.php#language.types.string.casting. 

So... I guess this PR is unnecessary! I'm sorry @vd1992! We can close the PR and issue as won'tfix.

vd1992:
> So... I guess this PR is unnecessary! I'm sorry @vd1992! We can close the PR and issue as won'tfix.

Aye aye

PR #8240, opened 2023-10-17
[Debt] Add foreign key constraints to tables

### LINKED ISSUE #6709: ♻️ Tables missing foreign key constraints
## ♻️ Debt/Refactor

The following tables are missing foreign key constraints for the fields listed (as sub-items):

- applicant_filter_classification
    - applicant_filter_id
    - classification_id
- applicant_filter_pool
    - applicant_filter_id
    - pool_id
- applicant_filter_qualified_classification
    - applicant_filter_id
    - classification_id	
- applicant_filter_skill
    - applicant_filter_id
    - skill_id
- classification_user
    - classification_id
    - user_id
- generic_job_title_user
    - generic_job_title_id
    - user_id
- generic_job_titles
    - classification_id
- permission_user
    - user_id
- pools
    - team_id
- role_user
    - user_id
- team_department
    - team_id
    - department_id

## ✅ Acceptance Criteria

- [ ]  table have appropriate foreign key constraints via migrations

vd1992:
🤖 Resolves #6709

## 👋 Introduction

Adds constraints to various tables through a migration. 

## 🕵️ Details

Pretty straightforward. 
Since most were pivots, I ended up sticking to cascade on deletion for all. 

These tables from the issue body already had foreign keys. 

image

image

## 🧪 Testing

1. Migrate and rollback successful
2. Bounce through tables, check constraints
3. Checks pass

## 💻 Deployment

Check migration ran successfully


api/database/migrations/2023_10_17_183718_add_foreign_key_constraints.php
```diff
@@ -0,0 +1,104 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('applicant_filter_classification', function (Blueprint $table) {
```
petertgiles, line 14: I'm not sure I agree with all these delete cascades.  If a team is deleted then delete all their pools? :scream: I think we need to think these through.

When there's a very tight "composition" relationship that exists then it makes sense.  I'd agree with user-permission.  If the user is deleted it makes sense for their permissions to also disappear.  And some of the pivots.  If an applicant_filter is deleted it makes sense to remove its pivots to classifications.  But if a classification is deleted I'm not sure I want all the search filters that depend on it to silently disappear.

This seems pretty out of scope for this PR.  Unless you want to go through these carefully one-by-one I'd suggest not including the delete cascades at all unless it really breaks a delete use case that's in use in the app.  We can always add them later if we need to.

petertgiles (changes requested):
Great catches!  Just a bit of scope clarification needed.

api/database/migrations/2023_10_17_183718_add_foreign_key_constraints.php
vd1992: Hm, okay. I thought for things like pivots a delete cascade was a safe thing. But if it isn't I can wind it back a bit. 
I got rid of some in https://github.com/GCTC-NTGC/gc-digital-talent/pull/8240/commits/e08d3ba68de7973c02dc151029c9d1312dfbbf19

Basically only kept deletion cascading in cases of references to the users table. We do have user deletion so I thought that was the one cascade operation best to keep around.

petertgiles (approved):
Looks great!  I suggest adding a deploy note to this one to double-check that the migration goes though.  It is possible that these constraints won't be able to be added to the prod DB if the data is already inconsistent.

vd1992:
> Looks great! I suggest adding a deploy note to this one to double-check that the migration goes though. It is possible that these constraints won't be able to be added to the prod DB if the data is already inconsistent.

Sure. 
Deployment tag added to issue, notes added to this PR.

PR #8246, opened 2023-10-18
[Debt] Migrate e2e tests to use TypeScript

### LINKED ISSUE #5754: 🧪 Update `e2e` tests to use TypeScript
## 🧪 Tests

I like TypeScript and the safety it provides and think it would be a great idea to use it for our tests as well! 😄 

## 🕵️ Details

We recently[ updated our `cypress.config.js` to be TypeScript](https://github.com/GCTC-NTGC/gc-digital-talent/commit/f387927d33ee024fc8b315b0aa54ca402cde7154). I'm thinking it might be a good time to convert all of our tests to use typescript and reduce our reliance on JavaScript. 

## 🙋‍♀️ Proposed Solution

Follow[ the guide that cypress has in their documentation](https://docs.cypress.io/guides/tooling/typescript-support#Types-for-Custom-Commands).

## ✅ Acceptance Criteria

- [ ] Convert one spec to typescript
- [ ] All cypress tests still pass
- [ ] Very limited start - just convert one helper file or test to start with
- [ ] Make a new issue afterwards with remainder of work and lessons learned

esizer:
🤖 Resolves #5754 

## 👋 Introduction

Updates all code in `apps/e2e` to use TypeScript.

## 🕵️ Details

There are some important items to note for this change.

- Custom commands must be typed
- `cy.get` is a a generic where you must pass in the type you expect it to return
- `cy.graphqlRequest` is now a generic where you must pass in the query being run

### Custom Commands

This adds all the types for our existing custom commands to the [`cypress.d.ts`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/d3634b80cca62e9216a6b75a1168714fdf1804ee/apps/e2e/cypress.d.ts) declaration file. Items that wrap and need to be used should have a return type of `Chainable<T>` where `T` is the type of the item being wrapped.

### `cy.get()`

This method is a a generic where you can pass the type you expect it to return.

```ts
cy.wrap(someUser).as("testUser");
// User now has the correct type
cy.get<User>("@testUser").then(user => {});
```

### `cy.graphqlRequest()`

Similar to the `cy.get()` method, this is also now a generic that accepts the query or mutation type so we get the repsonse and variables typed properly.

```ts
// Data is typed as the response from the query
cy.graphqlRequest<GetClassificationsQuery>({}).then(data => {});

// Data is typed as the response from the mutation
cy.graphqlRequest<CreatePoolMutation>({}).then(data => {});
```

## 🧪 Testing

1. Install new types `npm i`
2. Build `npm run build`
3. Run all e2e tests `npm run e2e:all`
4. Confirm they pass with no errors


apps/e2e/cypress/e2e/admin/pools.cy.ts
```diff
@@ -229,33 +229,37 @@ describe("Pools", () => {
     cy.wait("@gqlallPoolsQuery");
 
     // Navigate to edit pool page
-    cy.findByRole("button", { name: /show 10/i }).click();
-    cy.get("[role=menuitemradio]").contains(/50/i).click();
+    cy.findByRole("button", { name: /show 10/i })
+      .click()
+      .then(() => {});
+    cy.get("[role=menuitemradio]").contains(/50/i).click({ force: true });
```
petertgiles, line 235: Ooh, these force clicks make my spidey sense tingle.  Were you finding something wasn't always ready here?  I'd rather see an explicit wait on something instead.

apps/e2e/cypress.d.ts
petertgiles, line 1: Could you help me understand why this file was necessary?

petertgiles (changes requested):
Wow, this turned out lovely!  :heart_eyes:   I highlighted a few things to check but overall this looks amazing.

apps/e2e/cypress.d.ts
esizer: Oh yeah for sure! So, since custom commands are chained off of `cy` you need to add them to the `Cypress.Chainable` interface to type them.

This does two things, types the actual definition of the new command, as well as provides the type for using the command as well. You will notice, that if you try to define or use a command that has not been typed, it complains.

You can read more about it in the Cypress documentation for TypeScript. Specifically, the [custom commands section](https://docs.cypress.io/guides/tooling/typescript-support#Types-for-Custom-Commands).

Screenshot 2023-10-19 095059
Screenshot 2023-10-19 095220

petertgiles: Very cool, thanks.  :+1:

apps/e2e/cypress/e2e/admin/pools.cy.ts
```diff
@@ -229,33 +229,37 @@ describe("Pools", () => {
     cy.wait("@gqlallPoolsQuery");
 
     // Navigate to edit pool page
-    cy.findByRole("button", { name: /show 10/i }).click();
-    cy.get("[role=menuitemradio]").contains(/50/i).click();
+    cy.findByRole("button", { name: /show 10/i })
+      .click()
+      .then(() => {});
```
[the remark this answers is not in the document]
esizer: Oops, just a mistake while trying to fix this test intermittently failing 😬 

Removed in [remove workarounds from dropdown menu](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8246/commits/26fcbea8c7f4475b8357a2139d2e56b9894db397)

apps/e2e/cypress/e2e/admin/pools.cy.ts
esizer: Huh, I'm not sure why but it was failing without this (I was just following the tests radix wrote to try and fix it). But just removing these seem to work fine locally 🤷‍♀️ 

Removed in [remove workarounds from dropdown menu](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8246/commits/26fcbea8c7f4475b8357a2139d2e56b9894db397)

apps/e2e/cypress/support/commands.ts
```diff
@@ -49,55 +49,12 @@ Cypress.Commands.add("setLocale", (locale) => {
...
-    case 2:
-    default:
-      [email, password] = args;
-      // See: https://www.cypress.io/blog/2021/08/04/authenticate-faster-in-tests-cy-session-command/
-      cy.session([email, password], () => {
-        loginByPassword(email, password);
-      });
-  }
+Cypress.Commands.add("loginByRole", (role) => {
+  loginByRole(role);
```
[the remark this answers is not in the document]
esizer: Ha, yeah. It made me glad to remove those switch statements 😆

apps/e2e/cypress/support/poolHelpers.ts
```diff
@@ -1,22 +1,32 @@
...
   classification,
   poolAlias,
-}) {
+}: CreateAndPublishPoolArgs) {
   cy.createPool(adminUserId, teamId, [classification.id]).then(
     (createdPool) => {
-      cy.get("@testSkill").then((skill) => {
-        cy.log(skill);
+      cy.get<Skill>("@testSkill").then((skill) => {
+        cy.log(JSON.stringify(skill));
```
esizer, line 29: Seems as though some debug code may have been left behind. Do we need this anymore? 🤔

PR #8251, opened 2023-10-18
[Debt] Remove deprecated 'users.pools'

### LINKED ISSUE #7664: ♻️ Remove deprecated `user.pools` field
## ♻️ Debt/Refactor

This field is deprecated.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/4000aacaaf392bc2a0b763d4d9dbd9f67cb2680f/api/graphql/schema.graphql#L136-L139

## ✅ Acceptance Criteria

- [ ] `user.pools` field removed from schema file
- [ ] References to this field have been removed or replaced
- [ ] There may be some code changes required to complete this

vd1992:
🤖 Resolves #7664

## 👋 Introduction

Remove the deprecated field. 

## 🕵️ Details

Wasn't actually used/referenced anywhere I could find other than the faker file. 

## 🧪 Testing

1. Builds fine, tests pass

### DIFF
--- a/packages/fake-data/src/fakeUsers.ts
+++ b/packages/fake-data/src/fakeUsers.ts
@@ -15,7 +15,6 @@ import {
   Department,
   CitizenshipStatus,
   ArmedForcesStatus,
-  GenericJobTitle,
   PositionDuration,
   IndigenousCommunity,
   Maybe,
@@ -30,7 +29,6 @@ import {
 } from "./fakeExperiences";
 import fakeClassifications from "./fakeClassifications";
 import fakeDepartments from "./fakeDepartments";
-import fakeGenericJobTitles from "./fakeGenericJobTitles";
 import { GeneratedPoolCandidate } from "./fakePoolCandidateTypes";
 
 type GeneratedUser = User & {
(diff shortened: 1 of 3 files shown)


packages/fake-data/src/fakeUsers.ts
```diff
@@ -51,7 +48,6 @@ type GeneratedUser = User & {
 const generateUser = (
   departments: Department[],
   classifications: Classification[], // all classifications
-  genericJobTitles: GenericJobTitle[], // all generic job titles
```
vd1992, line 54: While looking at `fakeUsers.ts` I noticed this was unused and didn't actually get used in the function return. 
Probably left floating around after #8000

esizer: Are we not using `genericJobTitle` at all anymore? I noticed a lot of unused refs to it while converting tests to TS.

esizer (approved):
LGTM! 👍

packages/fake-data/src/fakeUsers.ts
vd1992: Pretty much, ya, #7578

PR #8282, opened 2023-10-19
[Fix] Reduce width of row selection column

### LINKED ISSUE #8222: 🐛 Row selection column issues
## 🐛 Bug

We have two minor issues with the row selection column on the new table.

- The padding is off between the header and cells causing a misalignment
- It does not cancel out the min width causing some unnecessary whitespace

## 📋 Steps to Reproduce

1. Navigate to `/admin/users`
2. Observe the misalignment of checkboxes and extra whitespace on the row selection column 

### Current

Screenshot 2023-10-12 145452

### Expected

image

## 🙋‍♀️ Proposed Solution

- Override min width on row selection column
- Adjust padding for row selection column to align with header checkbox

## ✅ Acceptance Criteria

- [ ] Row selection column only takes up the necessary horizontal space
- [ ] Row selection cells line up with row selection header checkbox

esizer:
🤖 Resolves #8222 

## 👋 Introduction

This allows the row selection column to shrink. As well, you can force columns to shrink below the min width of `x8` with `meta.shrink?: boolean` in the column def.

## 🧪 Testing

2. Navigate to the  "Tables\Responsive Table\Row Selection" story
3. Confirm the column with checkboxes is smaller than the other columns

## 📸 Screenshot

Screenshot 2023-10-19 092246

### DIFF
--- a/apps/web/src/components/Table/ResponsiveTable/Table.tsx
+++ b/apps/web/src/components/Table/ResponsiveTable/Table.tsx
@@ -131,7 +131,6 @@ const HeadCell = <T,>({ header, ...rest }: HeadCellProps<T>) => {
       {...(!isRowSelect &&
         !shouldShrink && {
           "data-h2-min-width": "base(x8)",
-          "data-h2-width": "base(100%)",
         })}
       {...styles.cell}
       {...rest}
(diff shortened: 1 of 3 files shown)


petertgiles (approved):
Looks great!

PR #8294, opened 2023-10-19
[Fix] Remove unused delete action from teams table

### LINKED ISSUE #8233: 🐛 Cannot delete a team
## 🐛 Bug

Cannot delete teams in the teams table. 

## 🕵️ Details

Edit link works fine. 
Given it is a cell with two actions, the second action probably just is missing or not rendering right. 

## 📋 Steps to Reproduce

1. Navigate to teams table and attempt to delete a team

## 📸 Screenshot

image

## ✅ Acceptance Criteria

- [ ] Remove the delete button from the table

esizer:
🤖 Resolves #8233 

## 👋 Introduction

Removes the delete action from the teams table.

## 🕵️ Details

We has a delete button in the `cells.actions()` table helper but it wasn't doing anything. This removes that to reduce confusion for users.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to `/admin/teams`
3. Confirm only an edit button appears in the actions column
4. Confirm no other tables are impacted

## 📸 Screenshot

Screenshot 2023-10-19 151357

### DIFF
--- a/apps/web/src/components/Table/Actions.tsx
+++ b/apps/web/src/components/Table/Actions.tsx
@@ -1,9 +1,8 @@
 import React from "react";
 import { useIntl } from "react-intl";
 import PencilIcon from "@heroicons/react/24/outline/PencilIcon";
-import TrashIcon from "@heroicons/react/24/outline/TrashIcon";
 
-import { Link, Button } from "@gc-digital-talent/ui";
+import { Link } from "@gc-digital-talent/ui";
 
 import { Maybe } from "~/api/generated";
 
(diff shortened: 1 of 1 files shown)


petertgiles:
Are you sure you want to just remove the action?  We have some tables that will need it, no?
image
Maybe we just need a way to choose which actions to show?

esizer:
> Are you sure you want to just remove the action? We have some tables that will need it, no?

Those ones actually have different delete actions. I think maybe the next step would be to consolidate our action columns. I know @Jerryescandon has been wanting to do this as well.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/ba4036fa6eb230b51bc6b97963f4c2bfc383c4b1/apps/web/src/pages/Users/UpdateUserPage/components/helpers.tsx#L36-L48

https://github.com/GCTC-NTGC/gc-digital-talent/blob/ba4036fa6eb230b51bc6b97963f4c2bfc383c4b1/apps/web/src/pages/Users/UpdateUserPage/components/helpers.tsx#L50-L73

petertgiles:
> Those ones actually have different delete actions. I think maybe the next step would be to consolidate our action columns.

Fair enough.  We'll worry about a shared action when we're ready to consolidate them.

PR #8300, opened 2023-10-20
[Debt] Throw error with unexpected URL params

### LINKED ISSUE #4801: Throw errors from components when expected path param is missing
Re-ticketed from https://github.com/GCTC-NTGC/gc-digital-talent/pull/4597#pullrequestreview-1183113613

> - create ticket to throw errors when queries receive undefined from useParams https://github.com/GCTC-NTGC/gc-digital-talent/pull/4597#discussion_r1024327275

When we started using `useParams` hook from react-router in components, it might return either the ID (string) or a undefined. But our codegen/urql hooks are unable to accept an undefined variable, so we're passing in an empty string when undefined.

But we probably want to do something more clever when this happens, as it means that the component is being used incorrectly, or that something has gone wrong. We may want to consider throwing an error and catching it later. Worth thinking about whether this would prompt a redirect (with `<Navigate replace />`?) or maybe rendering something else (an error) in the place where the component is attempting to be rendered.

cc: @esizer

esizer:
🤖 Resolves #4801 

## 👋 Introduction

This is an attempt to constrain our URL params by throwing an error earlier in the code.

## 🕵️ Details

Currently, our URL params are very loosely typed as `Record<string, string | undefined>`. This has lead to a pattern of passing `param || ""` to the API and wait for it to return something. This will throw an error when we know a param will fail (it doesn't exist or is not an ID).

### API

First of all, does the type of the hook and its args make sense? 

```ts
type UseRequiredParams<T extends Record<string, string>>(
  keys: keyof T | Array<keyof T>, 
  enforceUUID: boolean = true
) => Record<string, string>;

type RouteParams = {
  userId: string
  poolId: string;
};

// Ok
const { userId } = useRequiredParams<RouteParams>("userId");

// Ok
const { userId, poolId } = useRequiredParams<RouteParams>(["userId", "poolId"]);

// Fails
const { skillId} = useRequiredParams<RouteParams>("skillId");
```

### Return Type

I couldn't figure out how to enforce the return type. Can anyone assist me there? It is not required but it would be nice to have the return type be enforced as well. Right now it just returns `Record<string, string>`.

So, something like the following is totally okay until runtime.

```ts
type RouteParams = {
  userId: string
};

// Currently this is fine (until runtime). pool ID will just be undefined.
const { poold } = useRequiredParams<RouteParams>("userId");
```

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to pages confirming they load properly
3. Mess with an ID in the URL
4. Confirm an error is thrown instead of making an API request and you see the error page

## 📸 Screenshot

Screenshot 2023-10-20 144106

### DIFF
--- a/apps/web/src/hooks/useRequiredParams.ts
+++ b/apps/web/src/hooks/useRequiredParams.ts
@@ -10,7 +10,7 @@ export const isUUID = (str: string): boolean => {
   return !!str.match(uuidRegEx);
 };
 
-const assertParam = (param?: string, enforceUUID: boolean = false) => {
+const assertParam = (param?: string, enforceUUID: boolean = true) => {
   invariant(
     notEmpty(param),
     `Could not find required URL parameter "${param}"`,
@@ -21,18 +21,28 @@ const assertParam = (param?: string, enforceUUID: boolean = false) => {
   }
 };
 
-const useRequiredParams = <T extends Record<string, string | undefined>>(
-  keys: string | string[],
-  enforceUUID: boolean = false,
+type Keys<T, K extends keyof T> = K | Array<K>;
+
+const useRequiredParams = <
+  T extends Record<
+    keyof {
+      [P in keyof T as T[P] extends string ? P : never]: string | undefined;
+    },
+    string
+  >,
+>(
+  keys: Keys<T, keyof T>,
+  enforceUUID: boolean = true,
 ): Record<string, string> => {
   const params: Params<string> = useParams<T>();
   const keyArray = Array.isArray(keys) ? keys : [keys];
 
   let nonEmptyParams = {};
   keyArray.forEach((key) => {
-    const param = params[key];
+    const param = params[String(key)];
     assertParam(param, enforceUUID);
-    nonEmptyParams = { ...nonEmptyParams, [key]: param };
+
+    nonEmptyParams = { ...nonEmptyParams, [key]: param ?? "" };
   });
 
   return nonEmptyParams;
--- a/apps/web/src/hooks/useRequiredParams.ts
+++ b/apps/web/src/hooks/useRequiredParams.ts
@@ -1,7 +1,13 @@
-import invariant from "tiny-invariant";
 import { Params, useParams } from "react-router-dom";
 
+import {
+  useLogger,
+  defaultLogger,
+  type Logger,
+} from "@gc-digital-talent/logger";
 import { notEmpty } from "@gc-digital-talent/helpers";
+
+import invariant from "~/utils/invariant";
 
 export const uuidRegEx =
   /[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[089ab][0-9a-f]{3}-[0-9a-f]{12}/;
@@ -10,14 +16,23 @@ export const isUUID = (str: string): boolean => {
   return !!str.match(uuidRegEx);
 };
 
-const assertParam = (param?: string, enforceUUID: boolean = true) => {
+const assertParam = (
+  param?: string,
+  enforceUUID: boolean = true,
+  logger: Logger = defaultLogger,
+) => {
   invariant(
     notEmpty(param),
     `Could not find required URL parameter "${param}"`,
+    logger,
   );
 
   if (enforceUUID) {
-    invariant(isUUID(param), `URL parameter "${param}" must be a UUID.`);
+    invariant(
+      isUUID(param),
+      `URL parameter "${param}" must be a UUID.`,
+      logger,
+    );
   }
 };
 
@@ -48,13 +63,14 @@ const useRequiredParams = <
   keys: Keys<T, keyof T>,
   enforceUUID: boolean = true,
 ): Record<string, string> => {
+  const logger = useLogger();
   const params: Params<string> = useParams<T>();
   const keyArray = Array.isArray(keys) ? keys : [keys];
 
   let nonEmptyParams = {};
   keyArray.forEach((key) => {
     const param = params[String(key)];
-    assertParam(param, enforceUUID);
+    assertParam(param, enforceUUID, logger);
 
     nonEmptyParams = { ...nonEmptyParams, [key]: param ?? "" };
   });
--- a/apps/web/src/hooks/useRequiredParams.ts
+++ b/apps/web/src/hooks/useRequiredParams.ts
@@ -25,6 +25,20 @@ type Keys<T, K extends keyof T> = K | Array<K>;
 
 const useRequiredParams = <
   T extends Record<
+    /**
+     * This is a way to self-reference the type to only allow string keys.
+     *
+     * Constrains keys of `T` to be the `P` alias which
+     * is constrained to type `string` and, a key of the passed in generic type.
+     *
+     * Right side could be `any` (we don't actually use this)
+     * but should be `string | undefined` to satisfy the type.
+     *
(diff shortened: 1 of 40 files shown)


apps/web/src/hooks/useRequiredParams.ts
```diff
@@ -0,0 +1,51 @@
...
+  keys: Keys<T, keyof T>,
+  enforceUUID: boolean = true,
+): Record<string, string> => {
+  const params: Params<string> = useParams<T>();
+  const keyArray = Array.isArray(keys) ? keys : [keys];
+
+  let nonEmptyParams = {};
+  keyArray.forEach((key) => {
+    const param = params[String(key)];
+    assertParam(param, enforceUUID);
```
esizer, line 43: I'm wondering if we start an error bag here instead of throwing immediately and throw all key errors at once? 🤔 

Right now it throws as soon as it hits the first error.

apps/web/src/hooks/useRequiredParams.ts
```diff
@@ -0,0 +1,51 @@
...
+    invariant(isUUID(param), `URL parameter "${param}" must be a UUID.`);
+  }
+};
+
+type Keys<T, K extends keyof T> = K | Array<K>;
+
+const useRequiredParams = <
+  T extends Record<
+    keyof {
+      [P in keyof T as T[P] extends string ? P : never]: string | undefined;
```
petertgiles, line 29: Wow, this is some dark TS voodoo.  Could you add some comments to help me out?

apps/web/src/hooks/useRequiredParams.ts
petertgiles: I don't feel strongly about this.  I imagine that hitting any errors here will be rare for the average user.

apps/web/src/hooks/useRequiredParams.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const uuidRegEx =
+  /[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[089ab][0-9a-f]{3}-[0-9a-f]{12}/;
+
+export const isUUID = (str: string): boolean => {
+  return !!str.match(uuidRegEx);
+};
+
+const assertParam = (param?: string, enforceUUID: boolean = true) => {
+  invariant(
```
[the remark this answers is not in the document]
esizer: Essentially, yeah. It is a super light weight library so we could just write a version of it and add logger to it to also log the error.

petertgiles:
Regarding the return type, I agree that it would be really nice though I think it's beyond my skill level.  :sob:

apps/web/src/hooks/useRequiredParams.ts
esizer: Hopefully this makes sense 😆 

[document generic useRequiredParams](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8300/commits/be2c7572fd40d2026aaf8e5661ada63e672f5ba6)

esizer:
> Regarding the return type, I agree that it would be really nice though I think it's beyond my skill level. 😭

Yeah, it took me longer than I wanted to figured out how to scope the keys to a generic passed in. I want to get the return type working as well to provide some autocomplete like the args but that proved to be much more challenging 😢

apps/web/src/hooks/useRequiredParams.ts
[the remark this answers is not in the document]
petertgiles: Very nice!

petertgiles (approved):
Love it!

tristan-orourke:
Hey @esizer, I spent a bit of time looking at this. I'm not sure if I understand the purpose of the voodoo typing in useRequiredParams 😅 but I wonder if something like this would work as a simplification:

```
function assertParam(
  param?: string,
  enforceUUID: boolean = true,
  logger: Logger = defaultLogger,
): asserts param is string {
  invariant(
    notEmpty(param),
    `Could not find required URL parameter "${param}"`,
    logger,
  );

  if (enforceUUID) {
    invariant(
      isUUID(param),
      `URL parameter "${param}" must be a UUID.`,
      logger,
    );
  }
}

const useRequiredParams = <
  T extends Record<string, string>,
  P extends keyof T = keyof T,
>(
  keys: P | Array<P>,
  enforceUUID: boolean = true,
): Record<P, string> => {
  const logger = useLogger();
  const params: Params<string> = useParams<T>();
  const keyArray = Array.isArray(keys) ? keys : [keys];

  return keyArray.reduce(
    (nonEmptyParams, key) => {
      const param: string | undefined = params[String(key)];
      assertParam(param, enforceUUID, logger);
      // eslint-disable-next-line no-param-reassign
      nonEmptyParams[key] = param; // param must be a string if assertParam didn't throw an error.
      // Typescript would already know this if we wrote a typeguard instead of asserting
      return nonEmptyParams;
    },
    {} as Record<P, string>,
  );
};

export default useRequiredParams;
```

The return type isn't quite perfect, but I think its an improvement. It gives:

```
type RouteParams = {
  userId: string
  poolId: string;
};

// Ok
const { userId } = useRequiredParams<RouteParams>("userId");

// Ok
const { userId, poolId } = useRequiredParams<RouteParams>(["userId", "poolId"]);

// Fails
const { skillId} = useRequiredParams<RouteParams>("skillId");

// Fails - this is an improvement over current version
const { skillId } = useRequiredParams<RouteParams>("userId");

// This is fine until runtime, but poolId will be undefined while TS thinks its a string!
const { poolId } = useRequiredParams<RouteParams>("userId");
// This is because TS can't infer the narrowed type of K (keys), and I had to set a default value of `keyof T`, ie it assumes we pass in all keys.
// You can override this with an explicit type for the keys...

// Fails
const { poolId } = useRequiredParams<RouteParams, "userId">("userId");
// Fails
key = "userId";
const { poolId } = useRequiredParams<RouteParams, typeof key>("userId");
```

We can force it to be consistent by removing the default type for K, but then we'd have to pass the explicit type for the keys each time :/
```

**Comment by tristan-orourke** [comment]
The problem above is that TS _could_ infer the string-literal type of the keys we pass in, but it can't infer the RouteParams type, and once we make one generic explicit we have to make them all explicit (or default). It's possible to make work nicely if useRequiredParams returns a second function which must be called. Something like this:

```
const useRequiredParams = <T extends Record<string, string>>(): {
  withKeys: <K extends keyof T>(
    keys: K | Array<K>,
    enforceUUID: boolean,
  ) => Record<K, string>;
} => {
  const logger = useLogger();
  const params: Params<string> = useParams<T>();
  return {
    withKeys: <K extends keyof T>(keys: K | Array<K>, enforceUUID = true) => {
      const keyArray = Array.isArray(keys) ? keys : [keys];
      return keyArray.reduce(
        (nonEmptyParams, key) => {
          const param: string | undefined = params[String(key)];
          assertParam(param, enforceUUID, logger);
          // eslint-disable-next-line no-param-reassign
          nonEmptyParams[key] = param; // param must be a string if assertParam didn't throw an error.
          // Typescript would already know this if we wrote a typeguard instead of asserting
          return nonEmptyParams;
        },
        {} as Record<K, string>,
      );
    },
  };
};

// Used like so...
type RouteParams = {
  userId: string
  poolId: string;
};

const { userId } = useRequiredParams<RouteParams>().withKeys(
    "userId",
    true,
  );

// This fails now!
const { poolId } = useRequiredParams<RouteParams>().withKeys(
    "userId",
    true,
  );
```

PR #8301, opened 2023-10-23
[Fix] Support nested fields in `ErrorSummary`

### LINKED ISSUE #7813: ✨ BasicForm error summary does not support field array subfields
## ✨ Feature

The BasicForm error summaries depend on supplying the component with the form labels through a _labels_ prop.  The keys of the label object must match the field name in React Hook Forms form object to link them up.  However, when using field arrays the field name has an array index in it, like "requirements.2.resource" which won't work for the labels object.  Therefore, fields in field arrays cannot be part of the error summary.

I suggest that he label lookup should just look at the last segment of the field name.  So the field name "requirements.2.resource"  should just look for a property named "resource".  It's simple but you can't have different labels for "requirements.2.resource" and "assets.0.resource" for example.

Alternatively, you could do something like use a map with asterisk keys, like "requirements.*.resource".  A little more flexible and complicated. (favour this one) 

## ✅ Acceptance Criteria

- [ ] BasicForm error summary can use labels in field arrays

esizer:
🤖 Resolves #7813 

## 👋 Introduction

This adds support for nested error messages in out `ErrorSummary`.

### Nested Labels

When defining labels for nested items, replace the index with an asterisk. The `ErrorSummary` will extract out the asterisk and replace it with the index of the specific error.

#### Example

```tsx
const labels: FieldLabels = {
  "nested.*.field": "Nest Field",
}

// Code omitted for brevity
<BasicForm {...formProps}>
  <Repeater.Root name="nested" {...repeaterProps}>
    {fields.map((field, index) => (
      <Repeater.Fieldset {...fieldsetProps}>
        <input
          name={`nested.${index}.field`}
          id={`nested.${index}.field`}
        />
      </Repeater.Fieldset>
    ))}
  </Repeater.Root>
</BasicForm>
```

## 🧪 Testing

2. Navigate to the "Basic Form" story
3. Trigger some errors on different inputs
4. Confirm you see errors for them in the summary
5. Confirm the links function

## 📸 Screenshot

Screenshot 2023-10-23 164135


yonikid15 (changes requested):
I found a `can't convert undefined to object` error (ErrorSummary.tsx line 34) 😢
The error is triggered by doing the following:

1. Click add to create a nested field
2. Click submit to reveal error summary block
3. Click add again and see error

Looking into a solution atm.


https://github.com/GCTC-NTGC/gc-digital-talent/assets/22059495/a9a88c1e-2228-4bb6-b88d-6ab91ee16e7f

image

packages/forms/src/components/ErrorSummary.tsx
```diff
@@ -1,18 +1,141 @@
...
+            ),
+          ];
+        });
+      }
+      // We have an error message so add it to the array (we don't want errors with no message)
+      if ("message" in fieldError) {
+        errorNames = [...errorNames, `${parentKey}${fieldName}`];
+      }
+    }
+  });
```
yonikid15, line 58:
```suggestion
   if (errors)
    Object.keys(errors).forEach((fieldName) => {
      const fieldError = errors[fieldName];
      if (fieldError) {
        // This is a root of a field array, so add it to the key so we are aware later
        if ("root" in fieldError) {
          errorNames = [...errorNames, `${fieldName}.root`];
        }
        // If it is a field array, loop through, hoisting up field names
        if (Array.isArray(fieldError)) {
          fieldError.forEach((subFieldError, index) => {
            errorNames = [
              ...errorNames,
              ...flattenErrors(
                subFieldError,
                `${parentKey}${fieldName}.${index}`,
              ),
            ];
          });
        }
        // We have an error message so add it to the array (we don't want errors with no message)
        if ("message" in fieldError) {
          errorNames = [...errorNames, `${parentKey}${fieldName}`];
        }
      }
    });
```

yonikid15: This stopped the error on my end. However, adding a new item doesn't trigger the errors until submit it clicked again which I'm not sure if that's correct behaviour or not.

esizer: That sounds like it would be ideal actually!

yonikid15 (changes requested):
It's almost there just need to fix the nested items when there's an error. The top border isn't changing color when it has an error. Luckily you already did all the work to fix it with `flattenErrors()`.

You'll need to change this line from:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/7813-support-field-array-error-summary/packages/forms/src/components/Repeater/Repeater.tsx#L123-L125

To: 

```js
const fieldErrors = flattenErrors(errors);
```

You'll also need to export that function and remove something other lingering imports and props.

esizer:
> It's almost there just need to fix the nested items when there's an error. The top border isn't changing color when it has an error. Luckily you already did all the work to fix it with `flattenErrors()`.

Good catch! Fixed in [fix repeater error styles](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8301/commits/29b411ea358b0644359d1d79acd0a9b5f39ad111)

packages/forms/src/components/ErrorSummary.tsx
```diff
@@ -1,17 +1,95 @@
...
 import type { FieldLabels } from "./BasicForm";
+import { flattenErrors } from "../utils";
+
+type FieldNameWithLabel = {
+  label: React.ReactNode;
+  name: string;
+  index?: number;
+};
+
+const numberRegex = /\d/g;
```
petertgiles, line 24: This will only match a single digit.  I found the error block won't appear for errors in the 11th field array item.

petertgiles (commented):
This is great, I can't wait to get it in.  A few small things to check on.

petertgiles (changes requested):
:star:

packages/forms/src/components/ErrorSummary.tsx
```diff
@@ -71,26 +129,31 @@ const ErrorSummary = React.forwardRef<
...
+                {fieldIndex ? ` (${fieldIndex})` : null}
+              </ScrollToLink>
+              {intl.formatMessage(commonMessages.dividingColon)}
+              <ErrorMessage name={field.name} />
+            </li>
+          );
+        })}
       </ul>
       <Alert.Footer>
         <p>{intl.formatMessage(errorMessages.summaryContact, { a })}</p>
```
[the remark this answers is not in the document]
petertgiles: @gobyrne 
image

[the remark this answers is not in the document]
esizer: That is actually a pretty good idea. I'll try to remember to bring that up today 👍

packages/forms/src/components/BasicForm.stories.tsx
```diff
@@ -0,0 +1,173 @@
...
+                type="text"
+                rules={{ required: "This array field must be filled out" }}
+              />
+              <Input
+                name={`fieldArray.${index}.optionalArrayField`}
+                id={`fieldArray.${index}.optionalArrayField`}
+                label={labels["fieldArray.*.optionalArrayField"]}
+                type="text"
+              />
+              <NestedFieldArray parentIndex={index} />
```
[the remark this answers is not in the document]
esizer: Looks like I was just careless in setting the field names. Should be better after [fix input names in story](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8301/commits/75d813bbfe3b79a9c26375f9c9b5fe592a985c54)

packages/forms/src/components/ErrorSummary.tsx
esizer: Fixed this as well as the weird issue with selecting even numbers only 😆 

[update error label logic for nested fields](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8301/commits/f2a9433d995a9237123c147b4a3a75900baaa24e)

packages/forms/src/components/ErrorSummary.tsx
```diff
@@ -29,33 +107,13 @@ const ErrorSummary = React.forwardRef<
...
-      }
+  if (!errors || !show || !labels) return null;
 
-      return undefined;
-    })
-    .filter(notEmpty);
+  // Flatten the error object and get the label
+  const invalidFieldNames = addLabelsToErrors(errors, labels);
 
   const handleErrorClick: ScrollLinkClickFunc = (e, target) => {
```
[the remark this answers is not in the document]
esizer: Total oversight 😅 I was meaning to come back and fix this but forgot to!

Fixed in [add id to repeater root for error scrolling](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8301/commits/12aa3a2e516650754def5ddd2b77b041b808cade)

yonikid15 (approved):
The repeater error styling looks great, nice work 👍

packages/forms/src/components/ErrorSummary.tsx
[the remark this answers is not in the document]
petertgiles: Nice!

packages/forms/src/components/ErrorSummary.tsx
petertgiles: Beautiful!

petertgiles (approved):

PR #8312, opened 2023-10-24
[Debt] Safer policy fall-throughs

### LINKED ISSUE #8220: ♻️ Rewrite policies with safer fall-through conditions.
## ♻️ Debt/Refactor

When a policy method fails early, particularly due to additional conditions based on the `$request` input, the condition may be missed if other parts of the code change or are renamed. This can lead to a situation where refactors may lead to a policy succeeding where it should fail - a security risk.

## 🕵️ Details

Review all the Policy classes. In cases where a method returns `false` early based on a condition, rewrite the method to return false as a default, and only return true in more specific cases. This should be more robust to future refactoring. 

## 🙋‍♀️ Proposed Solution

PoolCandidatePolicy.php has some examples of exactly what we want.

The following classes have some methods which could use this treatment:
- [ ] UserPolicy.php
- [ ] PoolPolicy.php
- [ ] AssessmentStepPolicy.php

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Less specific conditions should cause policy methods to fail, not to pass

Q: Are the three policy files the whole scope of this issue?  Or should all policy files be reviewed?
A: The above 3 policies are needed to be changed. Other policies are compliant right now.

JamesHuf:
🤖 Resolves #8220 

## 👋 Introduction

Rewrites some policies to have safer fall through conditions

## 🧪 Testing

1. Look over the policies and confirm the changes make sense.
2. Confirm all tests pass.

### DIFF
--- a/api/app/Policies/AssessmentStepPolicy.php
+++ b/api/app/Policies/AssessmentStepPolicy.php
@@ -27,14 +27,19 @@ class AssessmentStepPolicy
             $pool = Pool::with('team')->find($poolId);
 
             if (! is_null($pool)) {
-                return $pool->getStatusAttribute() === PoolStatus::DRAFT->name
-                    && $user->isAbleTo('update-team-draftPool', $pool->team);
+                if ($pool->getStatusAttribute() === PoolStatus::DRAFT->name
+                    && $user->isAbleTo('update-team-draftPool', $pool->team))
+                {
+                    return true;
+                }
             } else {
                 return Response::deny('Cannot find a pool matching pool_id.');
             }
         } else {
-            return Response::deny('AssessmentStep must be associated with a pool when it is created.');
+            return Response::deny('Assessment step must be associated with a pool when it is created.');
         }
+
+        return Response::deny('Cannot create that assessment step');
     }
 
     /**
--- a/api/app/Policies/AssessmentStepPolicy.php
+++ b/api/app/Policies/AssessmentStepPolicy.php
@@ -7,6 +7,7 @@ use App\Models\AssessmentStep;
 use App\Models\Pool;
 use App\Models\User;
 use Illuminate\Auth\Access\HandlesAuthorization;
+use Illuminate\Auth\Access\Response;
 
 class AssessmentStepPolicy
 {
@@ -25,11 +26,15 @@ class AssessmentStepPolicy
             $poolId = $request['pool_id'];
             $pool = Pool::with('team')->find($poolId);
 
-            return $pool->getStatusAttribute() === PoolStatus::DRAFT->name
-            && $user->isAbleTo('update-team-draftPool', $pool->team);
+            if (! is_null($pool)) {
+                return $pool->getStatusAttribute() === PoolStatus::DRAFT->name
+                    && $user->isAbleTo('update-team-draftPool', $pool->team);
+            } else {
+                return Response::deny('Cannot find a pool matching pool_id.');
+            }
+        } else {
+            return Response::deny('AssessmentStep must be associated with a pool when it is created.');
         }
-
-        return false;
     }
 
     /**
--- a/api/app/Policies/AssessmentStepPolicy.php
+++ b/api/app/Policies/AssessmentStepPolicy.php
@@ -28,8 +28,7 @@ class AssessmentStepPolicy
 
             if (! is_null($pool)) {
                 if ($pool->getStatusAttribute() === PoolStatus::DRAFT->name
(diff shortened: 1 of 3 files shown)


api/app/Policies/AssessmentStepPolicy.php
```diff
@@ -22,17 +21,15 @@ class AssessmentStepPolicy
...
         if (array_key_exists('pool_id', $request)) {
             $poolId = $request['pool_id'];
-        } else {
-            return Response::deny('Cannot find pool.');
-        }
+            $pool = Pool::with('team')->find($poolId);
 
-        $pool = Pool::with('team')->find($poolId);
+            return $pool->getStatusAttribute() === PoolStatus::DRAFT->name
+            && $user->isAbleTo('update-team-draftPool', $pool->team);
```
tristan-orourke, line 29: I know our linter doesn't seem to care, but I want to ask for this line to be indented  😅

api/app/Policies/PoolPolicy.php
```diff
@@ -71,26 +72,19 @@ public function viewAnyPublished(?User $user)
...
-            return true;
-        } else {
-            return Response::deny('Cannot create a pool for that team.');
+            // Confirm the user can create pools for the team
+            if ($user->isAbleTo('create-team-pool', $team)) {
+                return true;
+            }
         }
+
+        return Response::deny('Cannot create a pool for that team.');
```
tristan-orourke, line 87: I like that previously the error response was different between 'Pool must be associated with a team when it is created.' and 'Cannot create a pool for that team.'. In fact, I'd recommend throwing one other response if $team is null, something like "Cannot find a team matching team_id".

api/app/Policies/PoolPolicy.php
```diff
@@ -131,16 +125,14 @@ public function updateDraft(User $user, Pool $pool)
...
+            }
         }
 
-        // The closing date must be greater than today's date at the end of day.
-        if ($pool->closing_date && $pool->closing_date < Carbon::now()->endOfDay()) {
-            return Response::deny('Expiry date must be a future date.');
-        }
-
-        return $user->isAbleTo('publish-any-pool');
+        return false;
```
tristan-orourke, line 135: As above, its okay to keep using `else` statements to return a useful `Response::deny`

api/app/Policies/PoolPolicy.php
```diff
@@ -174,12 +166,13 @@ public function closePool(User $user, Pool $pool)
...
-        if ($pool->getStatusAttribute() !== PoolStatus::DRAFT->name) {
-            return Response::deny("You cannot delete a Pool once it's been published.");
+        if ($pool->getStatusAttribute() === PoolStatus::DRAFT->name) {
+            $pool->loadMissing('team');
+
+            return $user->isAbleTo('delete-team-draftPool', $pool->team);
         }
 
-        return $user->isAbleTo('delete-team-draftPool', $pool->team);
+        return false;
```
tristan-orourke, line 175: I miss the `Response::deny` here too

tristan-orourke (changes requested):
Good job. However, I'd like you to keep returning `Response::deny` in any of the cases we returned it before. It's okay to return `Response::deny` instead of false, and return it early from an `else` statement if necessary.

api/app/Policies/AssessmentStepPolicy.php
JamesHuf: I changed this as part of readding denial messages to the policy but then I changed this into an if statement and I'm not sure I like it anymore.

api/app/Policies/PoolPolicy.php
JamesHuf: I've added some error responses (back) to this now.

api/app/Policies/PoolPolicy.php
JamesHuf: I've (re)added some error responses with [Update pool publish policy with specific denial messages](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8312/commits/4d517a84e304ac1373deecf0cfc01031f30eaad7).

api/app/Policies/PoolPolicy.php
JamesHuf: I've (re)added some error responses with [Update pool delete policy with specific denial messages](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8312/commits/97cebccf478afceb30d5c4e86c40f75597bc8265).

tristan-orourke (approved):

PR #8314, opened 2023-10-25
[Debt] Reduce Tsup output

### LINKED ISSUE #8313: ♻️ Supress tsup output
## ♻️ Debt/Refactor

Tsup has a lot of output that is not entirely helpful. I think we can safely suppress it and turn it back on when needed.

## 🙋‍♀️ Proposed Solution

Add `slient: true` to the `tsup.config.ts` files.

## ✅ Acceptance Criteria

- [ ] Tsup output is reduced

esizer:
🤖 Resolves #8313 

## 👋 Introduction

Makes tsup cli output silent.

## 🕵️ Details

This output is rarely ever helpful, we only usually care about errors.

## 🧪 Testing

1. Run `npm run build:fresh` to get no cache
2. Observe the reduced amount of output in the console

### DIFF
--- a/packages/app-insights/tsup.config.ts
+++ b/packages/app-insights/tsup.config.ts
@@ -8,6 +8,7 @@ export const tsup: Options = {
   sourcemap: env === "prod", // source map is only available in prod
   clean: true, // rimraf dist
   dts: true, // generate dts file for main module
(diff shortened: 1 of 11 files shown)


yonikid15 (approved):
Nice ✅

PR #8317, opened 2023-10-25
[Fix] Improve `useRequiredParams` types

### LINKED ISSUE #8316: ♻️ Improve `useRequiredParams` return type
## ♻️ Debt/Refactor

We have the args typed to only accept expected route params, it would be nice if we enforced the return type as well to catch errors earlier.

## ✅ Acceptance Criteria

- [ ] `useRequiredParams` has a return type that matches the generic
- [ ] Typescript auto complete works in deconstruction

esizer:
🤖 Resolves #8316 

## 👋 Introduction

Updates the types for `useRequiredParams` to make the return type more strict while simplifying the generic. 

🎩  h/t @tristan-orourke for assisting with this and doing most of the work 🥇 

## 🧪 Testing

1. Find a use of `useRequiredParams`
2. Confirm args are typed and autocomplete still
3. Cofirm the return type works
   - Autocompletes return vales in record
   - Shows error when you try to use a key that does not exist in the generic
   - Confirm the values are `string` and not `string | undefined`

### DIFF
--- a/api/app/Http/Controllers/SupportController.php
+++ b/api/app/Http/Controllers/SupportController.php
@@ -18,6 +18,9 @@ class SupportController extends Controller
             'status' => 2, // Required by Freshdesk API. Status of the ticket. The default value is 2.
             'tags' => [config('freshdesk.api.ticket_tag')],
         ];
+        if ($request->input('previous_url')) {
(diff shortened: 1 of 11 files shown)


brindasasi (approved):
LGTM!

PR #8342, opened 2023-10-26
[Feature] Admin Skill table improvements

### LINKED ISSUE #8339: ✨ Usability improvements to admin Skills table
## ✨ Feature

A few minor changes to the Skills table and edit form to make it easier to update many Skills in one sitting.

> every time i navigate into a skill, the default table set up comes back. Cause I’m doing this on a surface, the edit buttons don’t appear because the screen is too small to see the right side. I have to turn off the columns every time I go back to the screen to see the edit button… or go to the bottom of the table, scroll right and come back up to the skill.

> Would be great if the skills table kept the columns I’ve turned off… basically, keep table preferences navigating in and out of the table during a single work session

## 🕵️ Details

- Hide keywords column by default (it's not generally useful)
- Remove column for Edit button, and make skill name clickable instead. (Or instead, put action column on the left? Goal is not to have to scroll for action button. @Jerryescandon?)
- When submitting skill form is successful, if previous url was the skills table, then go back to the table with filters/state intact

## 🎨 Design File

### 🧑‍🎨 Designer

Refer to @tristan-orourke and @Jerryescandon 

## 📸 Screenshot

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Keywords column is hidden by default on the skills table
- [ ] Link to edit a skill should be on the left of the table
- [ ] If I change table state (filters, search, columns, or page) then edit a skill, and successfully submit changes, I should return to the table in the state I left it.

tristan-orourke:
🤖 Resolves #8339

## 👋 Introduction

Makes it easier to navigate the skill table, edit a single skill, and return to where you were on the table.

## 🕵️ Details

I'm trying a new pattern here, where the table link cell components add a `from` url to the location state, and the edit/view/create page can navigate to that url (which includes table state) if its available. I really like the way this works! I'd be interested in adopting this on other edit/view/create pages.

## 🧪 Testing

1. Log in ad admin@test.com
2. Go to skills table
3. Add a search filter, change visible columns and navigate to a later page
4. Narrow your screen so you can't see all columns at once
5. Can you get to the skill edit page without horizontal scrolling?
6. Edit a skill, submit changes
7. Did you come back to the table in the state you left it in? (page and filters and columns)
8. Open a link to edit skill page in a new tab
9. Can you edit the skill in the new tab? It should send you to a table in the default state this time

### DIFF
--- a/apps/web/src/components/Table/Actions.tsx
+++ b/apps/web/src/components/Table/Actions.tsx
@@ -17,7 +17,7 @@ const Actions = ({ id, label, editPathFunc }: ActionsProps) => {
   const intl = useIntl();
   const editPath = editPathFunc(id);
   const { pathname, search, hash } = useLocation();
-  const currentUrl = `${pathname}${search}${hash}`;
+  const currentUrl = `${pathname}${search}${hash}`; // Passing the current url, including search params, allows the next page to return to current table state.
   return (
     <Link
       href={editPath}
--- a/apps/web/src/components/Table/Actions.tsx
+++ b/apps/web/src/components/Table/Actions.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import { useIntl } from "react-intl";
 import PencilIcon from "@heroicons/react/24/outline/PencilIcon";
+import { useLocation } from "react-router-dom";
 
 import { Link } from "@gc-digital-talent/ui";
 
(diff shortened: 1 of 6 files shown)


esizer (approved):
Works well! 👏 

I feel like this would be very beneficial for all of these small admin tables were users are bouncing back and forth between tables and edit pages 👍

PR #8354, opened 2023-10-30
[Feature] Add version link to footer

### LINKED ISSUE #4691: ✨ Add commit data to source code of builds
## Description
@Jerryescandon brought up that it would be great to easily see what features are deployed on the website

We could make this easier by embedding the current commit into the deployed build artifact. I've seen this done by either putting the link very small in the footer, or embedding it into a source code comment (so it's not visible). We could also do it in the console log, though that would be a little more advanced for someone to check.

Related StackOverflow question: **[In Git, how can I write the current commit hash to a file in the same commit](https://stackoverflow.com/questions/3442874/in-git-how-can-i-write-the-current-commit-hash-to-a-file-in-the-same-commit)**

## Proposed Resolution

My **personal preference** would be storing the git hash very small and subtle as `Version: f38fc06` in the visible footer, for all environments, and link it to the github repo file tree at that specific commit. For example (linked to https://github.com/GCTC-NTGC/gc-digital-talent/tree/f38fc065cbe6b58ebe71613f4de025e2494640cc):

### Screenshot of Proposed Change
Screen Shot 2022-11-10 at 1 08 25 PM

## Acceptance Criteria

- [ ] should appear in the footer after the **Date modified** value
- [ ] link to GitHub (releases notes or [index of repo](https://github.com/GCTC-NTGC/gc-digital-talent/))
- [ ] include commit hash or tag (if tag exists) (`tag ? {tag} : {commitHash}`)
- [ ] for every vertical
- [ ] happens automatically ([webpack](https://stackoverflow.com/questions/38400314/including-git-commit-hash-and-date-in-webpack-build)?)

esizer:
🤖 Resolves #4691 

## 👋 Introduction

Adds a link to the footer that shows the current tag or commit hash (if they exist).

## 🕵️ Details

This only works with annotated tags. If there is no annotated tag, it falls back to the commit hash.

## 🧪 Testing

1. Build `npm run dev`
2. Scroll to the footer and confirm you see the version link and it functions as expected

## 📸 Screenshot

Screenshot 2023-10-30 105155

### DIFF
--- a/apps/web/src/components/Footer/VersionLink.tsx
+++ b/apps/web/src/components/Footer/VersionLink.tsx
@@ -6,9 +6,14 @@ import { commonMessages } from "@gc-digital-talent/i18n";
 
 const VersionLink = () => {
   const intl = useIntl();
-  let content = process.env.COMMIT_HASH;
+  let content;
+  let url;
   if (process.env.VERSION) {
     content = process.env.VERSION;
+    url = `https://github.com/GCTC-NTGC/gc-digital-talent/releases/tag/${process.env.VERSION}`;
+  } else if (process.env.COMMIT_HASH) {
+    content = process.env.COMMIT_HASH;
+    url = `https://github.com/GCTC-NTGC/gc-digital-talent/commit/${process.env.COMMIT_HASH}`;
   }
 
   if (!content) {
@@ -25,13 +30,7 @@ const VersionLink = () => {
         description: "Label for the specific version number of the site",
       })}
       {intl.formatMessage(commonMessages.dividingColon)}
-      <Link
-        external
-        fontSize="caption"
-        href={`https://github.com/GCTC-NTGC/gc-digital-talent/releases${
-          process.env.VERSION ? `/tag/${process.env.VERSION}` : ``
-        }`}
-      >
+      <Link external fontSize="caption" href={url}>
         {content}
       </Link>
     </span>
--- a/apps/web/src/components/Footer/VersionLink.tsx
+++ b/apps/web/src/components/Footer/VersionLink.tsx
@@ -1,0 +1,41 @@
+import React from "react";
+import { useIntl } from "react-intl";
+
+import { Link } from "@gc-digital-talent/ui";
+import { commonMessages } from "@gc-digital-talent/i18n";
+
+const VersionLink = () => {
+  const intl = useIntl();
+  let content = process.env.COMMIT_HASH;
+  if (process.env.VERSION) {
+    content = process.env.VERSION;
+  }
+
+  if (!content) {
+    return null;
+  }
+
+  return (
+    <span>
+      {" "}
+      &mdash;{" "}
+      {intl.formatMessage({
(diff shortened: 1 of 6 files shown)


apps/web/src/components/Footer/VersionLink.tsx
```diff
@@ -0,0 +1,41 @@
...
+    return null;
+  }
+
+  return (
+    <span>
+      {" "}
+      &mdash;{" "}
+      {intl.formatMessage({
+        id: "C3fUwm",
+        defaultMessage: "Version",
```
mnigh, line 24: The French translation for **Version** is the same as in English, **Version**.

esizer: Awesome, thanks! Added in [add translation](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8354/commits/206e123a321baf7388e1e8fa5107dd6ed98e8e8b)

brindasasi (commented):
- Deployed this branch in dev and verified without tag and it shows the commit properly 
- Created an unannotated tag as usual and deployed this branch in dev. Still shows the commit sha.  
   This will enforce us to create annotated tags going forward. I guess that shouldn't be an issue. 
- yet to test with annotated tags

**Questions**
When there is no tag present ( local & dev), the page displays the latest commit sha. But it points to the release page. 
 
image (23)

1. Is there any issue linking the commit sha with our repo , to show the commit like [this](https://github.com/GCTC-NTGC/gc-digital-talent/commit/fa65294a2dd9fed232bb648f9756de3fd229395d) ? 
2. Is there any benefit/concern doing that?
3.  Could we just leave it to empty if tags not present? Showing the tag on UAT & Prod has the real value IMO. 
4. Could we point to https://github.com/GCTC-NTGC/gc-digital-talent/tags page instead where it shows commit sha & the tag name 

@GCTC-NTGC/developers

apps/web/src/components/Footer/VersionLink.tsx
```diff
@@ -0,0 +1,41 @@
...
+        description: "Label for the specific version number of the site",
+      })}
+      {intl.formatMessage(commonMessages.dividingColon)}
+      <Link
+        external
+        fontSize="caption"
+        href={`https://github.com/GCTC-NTGC/gc-digital-talent/releases${
+          process.env.VERSION ? `/tags/${process.env.VERSION}` : ``
```
brindasasi, line 32:
```suggestion
          process.env.VERSION ? `/tag/${process.env.VERSION}` : ``
```

brindasasi (changes requested):
annotated tag works fine in dev env. could you just fix the url as below ?

mnigh:
> * Could we point to https://github.com/GCTC-NTGC/gc-digital-talent/tags page instead where it shows commit sha & the tag name

Wouldn't it make sense to link to the commit since that is what is being displayed?

substrae:
Just a heads up that the Figma footer element was missing this, so it's been added an pushed as well :)

Issue #8343, opened 2023-10-26
🐛 Tables reset page number when navigating back


tristan-orourke:
## 🐛 Bug

If you change page of a table, new page is reflected in url search params. If you navigate to a different url, the table state (including page) is saved in the history stack. If you navigate _back_, most table state is restored (search, columns, etc) but page is always reset to 1.

## 🦋 Expected Behaviour

You should go _back_ to the same page you were on before.

## 🕵️ Details

Add any additional details that could assist with troubleshooting/fixing the issue.

- **Operating System**: Name and version (if possible).
- **Browser**: Name and version (if possible).

## 📋 Steps to Reproduce

1. Log in as admin@test.com
2. Go to skills table (or departments, or skill families...)
3. Change visible columns and add a search filter
4. Go to another page (eg results 11-20)
5. Navigate back in browser
6. See that search and columns are where you left, but you're back on page 1 of results!

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the bug was addressed.

- [ ] Navigating _back_ to a table with page set in url search params returns you to that page properly

PR #8360, opened 2023-10-31
[Debt] Refactor pool candidates table to use responsive table

### LINKED ISSUE #7693: ♻️ Refactor PoolCandidatesTable to use new table component
## ♻️ Debt/Refactor

We want to use new table component instead of old ApiManagedTable.

## 🕵️ Details

## 🙋‍♀️ Proposed Solution

## 🌎 Localization

n/a

## ✅ Acceptance Criteria

- [ ] PoolCandidatesTable.tsx uses new table component, does not reference ApiManagedTable, but functionality remains the same

yonikid15:
🤖 Resolves #7693 

## 👋 Introduction

- Replaces old table component with Responsive table for all pages using pool candidates table

## 🧪 Testing

1. Ensure new responsive table is working on all pages using the `<PoolCandidatesTable />`: All pool candidates page (`http://localhost:8000/en/admin/pool-candidates`), View candidates of single pool (`http://localhost:8000/en/admin/pools/{id}/pool-candidates` ), Single request page (`http://localhost:8000/en/admin/talent-requests/{id})`
2. Ensure all parts of table are working correctly:
  - Search bar and filter by
  - Custom filter dialog
  - Hidden columns
  - Sorting
  - Row selection
  - Download CSV
  - Print profile
  - Pagination


esizer:
> Last thing is getting pagination to work. I'm stuck on an infinite loop which is originating from

I think this might be related to #8343 🤔

apps/web/src/components/PoolCandidatesTable/tableMessages.ts
mnigh, line 1: @yonikid15 if these are new strings to be translated, might as well do them sentence case so we are not adding new strings that are not sentence case. For example, Candidacy Status should be Candidacy status.

yonikid15: Nah, I just moved these into a new file so it's a little less crowded.
However, If there are any other changes that need to be made to copy though let me know and I can send it off asap

mnigh: ok, don't worry about it then, out of scope for this issue.

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
```diff
@@ -39,18 +37,7 @@ import {
...
-import cells from "~/components/Table/cells";
-import {
-  ColumnsOf,
-  handleColumnHiddenChange,
-  handleRowSelectedChange,
-  rowSelectionColumn,
-  SortingRule,
-  TABLE_DEFAULTS,
-} from "~/components/Table/ApiManagedTable/helpers";
+import { TABLE_DEFAULTS } from "~/components/Table/ApiManagedTable/helpers";
```
mnigh, line 40: Shouldn't anything `ApiManagedTable` be factored out?

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
```diff
@@ -852,97 +370,220 @@ const PoolCandidatesTable = ({
...
+    <Table<PoolCandidateWithSkillCount>
+      caption={title}
+      data={filteredData}
+      columns={columns}
+      isLoading={fetching || fetchingSkills}
+      hiddenColumnIds={["candidacyStatus", "notes"]}
+      search={{
+        internal: false,
+        label: intl.formatMessage({
+          defaultMessage: "Search pool candidates",
```
mnigh, line 547: @yonikid15 not sure if you have already sent this string for translation, but the French would be: "Chercher les candidats du bassin".

apps/web/src/components/PoolCandidatesTable/helpers.tsx
```diff
@@ -0,0 +1,360 @@
...
+  ProvinceOrTerritory,
+  SortOrder,
+  User,
+} from "~/api/generated";
+import useRoutes from "~/hooks/useRoutes";
+import { getFullNameLabel } from "~/utils/nameUtils";
+
+import cells from "../Table/cells";
+
+export const statusAccessorNew = (
```
esizer, line 36: Does the "New" suffix mean anything, or is it just a way you avoided namespace collision while working on this ticket?

If the latter, could we remove that suffix for these functions?

Also, if this returns anything but a primitive value, it should be named "cell" since [accessors should only return primitive values](https://tanstack.com/table/v8/docs/guide/column-defs#:~:text=must%20be%20configured%20to%20extract%20primitive%20values).

---

Both comments seem to apply to all accessor named functions here outside of a few. I beleive these are being used as cells, just named incorrectly since they are not accessors.

I think renaming all of these to be `{field}Cell` instead should resolve both comments.

apps/web/src/components/PoolCandidatesTable/helpers.tsx
```diff
@@ -0,0 +1,360 @@
...
+) => {
+  if (status === PoolCandidateStatus.NewApplication) {
+    return (
+      <span
+        data-h2-color="base(tertiary.darker)"
+        data-h2-font-weight="base(700)"
+      >
+        {status
+          ? intl.formatMessage(getPoolCandidateStatus(status as string))
+          : ""}
```
esizer, line 48: Not sure if this is in scope. But, we check to see if status exists 3 times in this function. I think it would be an easy win if instead we just check that at the start and return `null` instead. Also, we won't have an empty span tag as well.

```tsx
if(!status) return null;

// Now we know it exists and can render different messages based on it
```

apps/web/src/components/PoolCandidatesTable/helpers.tsx
```diff
@@ -0,0 +1,360 @@
...
+        : ""}
+    </span>
+  );
+};
+
+export const priorityAccessorNew = (
+  priority: number | null | undefined,
+  intl: IntlShape,
+) => {
+  if (priority === 10 || priority === 20) {
```
esizer, line 82: Same as the status function, if we just check for the priority and return null, we won't need to do additional checks below and won't end up with an empty span if it doesn't exist.

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
```diff
@@ -852,97 +370,220 @@ const PoolCandidatesTable = ({
...
+    columnHelper.display({
+      // TODO: Should this be an accessor instead of a display type?
+      id: "candidacyStatus",
+      header: intl.formatMessage(tableMessages.candidacyStatus),
+      cell: ({
+        row: {
+          original: { poolCandidate },
+        },
+      }) => candidacyStatusAccessorNew(poolCandidate.suspendedAt, intl),
+    }),
```
esizer, line 418: If we ever want to sort on it, yeah. Also, this is dangerously close to already returning a primitive value. If you remove the wrapping spans we could easily convert this to an accessor and no longer need the cell.

```tsx
  columnHelper.accessor(({poolCandidate}) => candidacyStatusAccessor(poolCandidate.suspendedAt, intl), { ... }),
```

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
```diff
@@ -852,97 +370,220 @@ const PoolCandidatesTable = ({
...
+        id: "candidateName",
+        header: intl.formatMessage(tableMessages.candidateName),
+        sortingFn: normalizedText,
+        cell: ({
+          row: {
+            original: {
+              poolCandidate: { user },
+            },
+          },
+        }) => `${user?.firstName} ${user?.lastName}`,
```
esizer, line 440: This looks like it returns the same as the accessor. If a cell definition does not exist, it will render the result of the accessor by default so not need for the cell here 👍

esizer (changes requested):
Nice! This is look great. 

I know it looks like a lot but, some minor items I noticed while reviewing. I think its just renaming a lot of stuff to better describe the purpose of it.

## Accessor vs. Cell

So, this is a hang up from our previous implementation where we confused the two and I'm not sure if it makes sense to solve it here but this would be the last table using the old naming convention if we don't 😅 .

I took a look and I think for a lot of this we don't need to control markup or even return JSX at all so we could drop the `cell` definition for most of these and just update the accessors. I'd suggest the following changes:

 - `statusAccessorNew` -> Rename `statusCell`
 - `priorityAccessorNew` -> Rename `priorityCell`
 - `viewPoolCandidateAccessor` -> Rename  `viewPoolCandidateCell`
 - `candidacyStatusAccessorNew` -> Return string, move to accessor, rename `candidacyStatusAccessor`
 - `notesAccessorNew` -> Rename notesCell`
 - `userNameAccessor` -> We have a function to do this called `getFulllNameLabel`, also no need for a cell
 - `preferredLanguageAccessorNew` ->  return string, move to accessor, rename `preferredLanguageAccessor`


## Locked sort

So, something we did not take into consideration is that the first two rows are sorted but cannot be changed. This may require some changes to the base table component. We have no way of indicating locked sorting 🤔 I think a quick and dirty would be to add it to the meta of the column definition.

```tsx
// Define the type in `./src/table.d.tsx`
declare module "@tanstack/table-core" {
  interface ColumnMeta<TData extends RowData, TValue> {
    /** Mark this column as locked sorting (user cannot control) */
    sortingLocked?: boolean;
  }
}

// Column definition
const columns = [
  columnHelper.accessor("id", {
    meta: {
      sortingLocked: true
    }
  })
];

// Determine if the column is locked in `./src/Table.tsx:HeadCell` 
const HeadCell = <T,>({ header, ...rest }: HeadCellProps<T>) => {
  const sortingLocked = header.column.columnDef.meta?.sortingLocked;
  ...
  
  return <SortButton locked={sortingLocked}>
}

// Then, disable button and render lock icon if locked
const SortButton = <T,>({ column, locked, children }: SortButtonProps<T>) => {

  return <Button disabled={locked} icon={locked ? LockIcon : sortIcon}>
}
```

apps/web/src/components/PoolCandidatesTable/helpers.tsx
yonikid15: > If the latter, could we remove that suffix for these functions?
Ya just left over to avoid namespace, so it will be removed.

> Also, if this returns anything but a primitive value, it should be named "cell" since [accessors should only return primitive values](https://tanstack.com/table/v8/docs/guide/column-defs#:~:text=must%20be%20configured%20to%20extract%20primitive%20values).

Okay sounds good, thanks for clearing this up.

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
yonikid15: Fixed by [95daca8](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8360/commits/95daca8101e574241aa2f72decff7b9cc7c12cdd)

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
yonikid15: Fixed by [95daca8](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8360/commits/95daca8101e574241aa2f72decff7b9cc7c12cdd)

apps/web/src/components/PoolCandidatesTable/helpers.tsx
yonikid15: Fixed by [95daca8](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8360/commits/95daca8101e574241aa2f72decff7b9cc7c12cdd)

apps/web/src/components/PoolCandidatesTable/helpers.tsx
yonikid15: Fixed by [95daca8](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8360/commits/95daca8101e574241aa2f72decff7b9cc7c12cdd)

apps/web/src/components/PoolCandidatesTable/helpers.tsx
yonikid15: Fixed by [95daca8](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8360/commits/95daca8101e574241aa2f72decff7b9cc7c12cdd)

yonikid15:
@esizer Okay I believe it's ready for one last review. Let me know if I missed anything with the columns and locked sort. Also, I locked the status AND priority column.

esizer (approved):
This is great! 🥇 

Thanks for bearing with me on the reviews 😅

yonikid15:
> Thanks for bearing with me on the reviews 😅

Anytime! Thanks for reviewing it so well 🤝

PR #8361, opened 2023-10-31
[Fix] Prevent table resetting parameters

### LINKED ISSUE #8343: 🐛 Tables reset page number when navigating back

esizer:
🤖 Resolves #8343 

## 👋 Introduction

Fixes the new `<ResponsiveTable />` resetting some state variables on initial page load.

## 🕵️ Details

Looks like we forgot to supply the table with `initialState` when it exists. This meant that our table was starting out with an empty state. Then, when that state did not match our current state, it tried to overwrite the URL and use that to update our controlled state.

Simply providing the table with an initial state prevents that. 

## 🧪 Testing

1. Run `npm run dev`
2. Navigate to `/admin/skills`
3. Change the state, confirm URL updates
4. Navigate away from that page
5. Click the browser back button
6. Confirm the URL matches that in step 3
7. Confirm the table state matches the URL state with expected date

### DIFF
--- a/apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
+++ b/apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
@@ -135,6 +135,7 @@ const ResponsiveTable = <TData extends object>({
       ...state,
       rowSelection,
     },
+    autoResetPageIndex: false,
     enableGlobalFilter: isInternalSearch,
     enableRowSelection: !!rowSelect,
     enableSorting: !!sort,
@@ -185,7 +186,9 @@ const ResponsiveTable = <TData extends object>({
         (previous) => {
           const newParams = new URLSearchParams(previous);
 
-          if (isEqual(sortingState, sort?.initialState ?? [])) {
+          const initialSortState =
+            sort?.initialState ?? INITIAL_STATE.sortState;
+          if (isEqual(sortingState, initialSortState)) {
             newParams.delete(SEARCH_PARAM_KEY.SORT_RULE);
           } else {
             newParams.set(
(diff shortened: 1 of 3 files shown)


vd1992 (commented):
The back button preserves page but now I am noticing bugs that may be from this PR or #8342
For the skills table. 

Pagination count

image

Cannot search skills by name

image

esizer:
> Pagination count

This one is expected. That is actually the number of items it is displaying, not the page numbers.

> Cannot search skills by name

I'll take a look at this one to make sure this PR did not break that.

vd1992 (changes requested):
Searching not working right on the pools table either 😕 

Somehow you can have a bug where your pagination will get locked to a page, page two in this case. It keeps setting it to page two and this makes searching hard because stuff with results only on one page return with no results as their second page is empty.

Recording below, you can see the URL getting set back to page two when searching. At the 47 second mark I switch to a second tab, sorry for accidentally not including tabs in the recording.
Searching "IAP" then works, it only has enough results for one page.  

https://recordit.co/uAV8hPUKf1

esizer:
> Searching not working right on the pools table either 😕

okay, I think I finally figured it out 😅 . It appears as though, `react-table` attempts to reset pagination back to `initialState` when the filters change. Since we control the state, it was some wonky race condition where we were trying to set it to the first page while `react-table` was resetting back to the `initialState` which is being read from the search params.

I disabled this auto reset in [stabilize initial state, url updates](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8361/commits/a51e8719d4b14e4b726ca435380e92eaa70f6279) along with changing how we determine when to delete the search params.

vd1992 (approved):
Exercising different tables, searching, visiting items then returning; all appears good.
This appears to work like a dream now.

Amazing! Hope it wasn't too frustrating getting to the bottom of it.
🏆

Issue #8363, opened 2023-10-31
✨ Repeater updates for assessment step actions


petertgiles:
## ✨ Feature

The assessment step builder page has suggested a few changes for the assessment step mutations.

Or do we want to abandon this in favor of [a non-form component](https://github.com/GCTC-NTGC/gc-digital-talent/issues/8161)?

## ✅ Acceptance Criteria

- [ ] Allow preventing individual actions (move up, move down, delete, edit) independently for each fieldset
- [ ] Remove numOfLockedItems prop in preference for above
- [ ] Figure out solution for returning focus to edit and delete buttons from dialog (add custom delete button like custom edit button?)
- [ ] Figure out solution to maintain focus on a move button after a move operation

brindasasi:
we can close this issue in favour of implementing https://github.com/GCTC-NTGC/gc-digital-talent/issues/8161

brindasasi:
We re-open this issue in favour of closing #8161 
we realized it would be lot easier to update rather than the new component

Issue #8362, opened 2023-10-31
✨ Mutation updates for assessment step actions


petertgiles:
## ✨ Feature

The assessment step builder page has suggested a few changes for the assessment step mutations.

## ✅ Acceptance Criteria

- [x] If screening questions are added or removed, DONT automatically create an assessment step
- [x] If a screening question step is added/updated, force it sort order 2 regardless of the mutation parameters and reflow the sort order of other steps in the same pool
- [x] If a screening question step is added and there is already a screening question step, error out the mutation
- [x] If the screening question step is deleted, also hard delete all the screening questions
- [x] If the regular screening question delete mutation is called, do a HARD delete
- [x] Custom mutation to allow create/edit screening question assessment step with screening questions?  Also should have a sync method, not create/update/delete. Update the Assessment Plan builder page to use this new mutation.
- [x] Custom mutation to allow syncing multiple assessment steps in one mutation for reordering?  Or just for moving? 
  Update the Assessment Plan builder page to use this new mutation.

vd1992:
> If a screening question step is added/updated, force it sort order 2 regardless of the mutation parameters and reflow the sort order of other steps in the same pool

Might it be easier to go with Tristan's suggestion of just as Application screening is locked to `sortOrder = 1`, do the same for screening questions and lock `sortOrder = 2`, and then additional assessments get added at `sortOrder = 3` and above. So, a placement at 2 is not always present if a process has no screening

petertgiles:
> Might it be easier to go with @tristan-orourke 's suggestion of just as Application screening is locked to `sortOrder = 1`, do the same for screening questions and lock `sortOrder = 2`

Yeah, that would be better than reflowing the others.    We would have to make sure that any mutation calls for other step types have a floor of "3" for sort order then.

PR #8364, opened 2023-10-31
Finalize assessment plan page

### LINKED ISSUE #7916: ✨ Finalize Assessment Plan Page
## ✨ Feature

The assessment plan builder is the first step in the Record of decision (RoD) flow. It allows us to capture the list of assessments to be used during the staffing process, the order in which they will be performed, and the skills that each assessment will be targeting.

It features a repeater component with dialogs for details, and a table to overview how the skills and assessments are related to each other.

## 🕵️ Details

1. The title includes a chip/label to show the plan's current state
2. The plan is complete when ALL skills have been attached to at least one assessment
3. If a skill is missing an assessment, a waning will be displayed at the top
4. The list always include "Application screening" locked in the number one spot
5. IF "screening questions (at the time of application)" is selected, then:
     - It will always be listed on the second spot
     - the dialog does not offer optional fields for a different assessment title
     - the dialog changes to include a repeater (limit 6) to add the screening questions in english or french
6. IF the repeater has 4 or more assessments, a warning message will show underneath it
7. IF the repeater has only the application, display a warning for... too few assessments.
8. IF an assessment has no skills a warning will be displayed inside the repeater item (where skills would usually go)
9. There is a table at the end of the page that shows the information captured on the repeater but it is not editable

## 🎨 Design File and Designer

[Link to figma file](https://www.figma.com/file/em5ArPr044YaIe1Vb1zCZz/Pools---Record-of-Decision?type=design&node-id=5437%3A34317&mode=design&t=11Dob6JX6BOb3lle-1) by @Jerryescandon 

## 📸 Screenshots

<details><summary> Full page screenshot</summary>
<p>

Frame 40

</p>
</details> 

<details><summary>Dialog (regular)</summary>
<p>

Assessment details - Repeater dialog 01

</p>
</details> 
<details><summary>Dialog (with screening questions) </summary>
<p>

Assessment details - Repeater dialog

</p>
</details> 

## ✅ Acceptance Criteria

- [ ] Add accordions if copy is available
- [ ] Add card repeaters to "organize" section
- [ ] Wire up dialogs
- [ ] Add skill summary table
- [ ] Review with Jerbo
- [ ] "Save plan and go back" saves fieldset changes (reorder, delete)

petertgiles:
🤖 Resolves #7916

## 👋 Introduction

This branch pulls together all the pieces to finalize the assessment plan builder page.

## 🕵️ Details

~:warning: There are a few rough edges from where the pieces didn't quite line up.  :grin:   I've opened two issues:
 #8362
 #8363~
These issues are completed.

The following limitations exist:
- If a move occurs the focus does not remain on the move button - Added to #8491
- If two steps with the same type (but no custom titles) are added then we get a bunch of "identical key" warnings from React from the table.  Do we need to add validation to the frontend and backend for identical steps? Or should the table be updated?  Either way - probably out of scope.

:warning: The requirements have changed from the issue a bit - step deletes and moves occur immediately, not with a submit button.
:heavy_check_mark: Design review with Jerbo and Gray is complete.

## 🧪 Testing

1. Rebuild site, log in as admin, navigate to `/en/admin/pools/:poolId/plan`.
2. Ensure it's a pool in draft, for a team you belong to.
3. Or start a new draft pool.

Suggested steps:
- [ ] Acceptance criteria [from issue](https://github.com/GCTC-NTGC/gc-digital-talent/issues/7916) satisfied.
- [ ] A pool skill can be removed on the pool editor page which is already associated with an existing step.
- [ ] Application screening step is locked to step 1
- [ ] Screening question step is locked (if exists) to step 2
- [ ] Steps can be added, removed, and reordered

## 📸 Screenshot

image


tristan-orourke:
@petertgiles I'm not certain, but I suspect TypeScript doesn't like that AssessmentType can have AssessmentResults which have an AssessmentType.

But regardless, this is a good fix. I'm generally in favour of narrowing our types wherever we can, anyway. It makes a lot of tricky type shenanigans go away.

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/AssessmentDetailsDialog.tsx
```diff
@@ -16,179 +16,375 @@ import {
...
+        );
+      });
   };
 
   const canAddScreeningQuestions =
     fields.length < SCREENING_QUESTIONS_MAX_QUESTIONS;
 
-  const assessedSkillsItems: CheckboxOption[] = allPoolSkills
-    .map((poolSkill) => ({
+  const assessedSkillsItems: CheckboxOption[] = allPoolSkills.map(
```
esizer, line 350: Unsure of the best approach here but, if you have a pool in draft mode with no skills, it simply renders an empty skill selection check list which looks odd and never validates.

Screenshot 2023-11-06 151432

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/AssessmentPlanBuilderPage.tsx
esizer, line 1: I can get to this page even if I don't have permission to edit it. This may be okay because it is also the only place you can view it. However, it appears as though I can edit it and then I just get error toasts on every action.

This was very confusing to me! Should we disabled actions if users can't edit or should we prevent them from getting to the page entirely? 🤔

esizer, line 1: Probably not the place for this, but I was also very confused by the skill summary on the side. Specifically, the status for "Total assessment count".

Before reading the code, my assumption was:

- Warning: Not all skills are assessed
- Success: All skills assessed

I had no idea it was actually related to the number of steps you had added 😕

esizer (commented):
I got about as far as I could reviewing this but I'm having a lot of trouble understanding that intended functionality of this page. 😢 On the surface, it seems very straight forward, until I start reading all the instructions and edge case scenarios we outline! 😖 

Left some comments about items I struggled with but overall, I found this interaction to be very verbose and all the added instructions made it harder to understand for me and made me second guess myself!

Some smaller questions:

- Is there a plan to limit the number of "types" of steps? Right now I can just add "additional assessments" types that evaluate all skills and call it day!
- Some of the copy reads a bit odd
    - we use a lot of adjectives and adverbs i.e "very special" 
    - we don't consistently spell out digits (two to three, 6)
- This may be related to the limitations but we seem to rely heavily on users reading instructions and just hope they follow them. I think it would be far more suited to rely more on validation than instructions 😬 


Unrelated, but the skill summary table is hard to read. I can't really determine what it is trying to represent? Does a check mark mean there is a step to assess that skill? 🤔 Also, the accordion describing why behavioral skills are not on the initial application doesn't seem to belong here? I thought this page dealt specifically with after the initial application? My best guess was that it is referring to the "application screening" column on the table where most behavioral skills read "not applicable" If so, maybe we should align those terms

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/AssessmentDetailsDialog.tsx
Jerryescandon: We could probably use a null state here like:
`There are no skills selected. Please use the 'Advertisement information' page to add some skills.`

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/AssessmentPlanBuilderPage.tsx
Jerryescandon: This one is confusing, i'll give you that. 😆

Jerryescandon:
> I found this interaction to be very verbose

Ooff same, and we flagged this for copy editing but it still reads very heavy :( 

But overall I do have some answers but a lot of this could use improvements, specially copy. 

1. The types of steps are a building block for features down the road. The idea IS in fact to expand these to include stuff like "rating guides" and templates. Ideally selecting the type would be more useful to managers that are starting out and have little experience on creating assessments by giving them further options and guidance constructing their selected type if assessment.
2. Oh man I will accept all help with copy, I will try to bring this to the copy caucus again, cause that meet with Lauren was not super helpful in making things more readable.
3. I AM hoping we are adding all the validation necessary to enforce the instructions. HOWEVER If your idea is to remove instructions until the scenario happens... like the "I need more than six assessments" I dont hate that. We could remove the dropdown text and put it on the context dialog that appears when you reach the higher limit.

Jerryescandon:
> If so, maybe we should align those terms

Yup, good catch. These should be aligned.

Jerryescandon:
> I can't really determine what it is trying to represent? Does a check mark mean there is a step to assess that skill?

This has become more complex, and might be worth looking at again with the new requirements coming in. For now it was only supposed to show "at a glance" the entire plan and most importantly if there are any skills missing.

yonikid15 (commented):
image

I might have found an edge case 😢 When adding screening questions on the Edit Pool page, then jumping to the Assessment plan page the error message is shown on the screening questions type. Is this expected behaviour?

Jerryescandon:
> image
> 
> I might have found an edge case 😢 When adding screening questions on the Edit Pool page, then jumping to the Assessment plan page the error message is shown on the screening questions type. Is this expected behaviour?

This is actually correct. (For now)

This will be addressed during cleanup and future tickets. We should ideally only be able to add screening questions here. All good 👍

esizer:
> 1. The types of steps are a building block for features down the road. The idea IS in fact to expand these to include stuff like "rating guides" and templates. Ideally selecting the type would be more useful to managers that are starting out and have little experience on creating assessments, by giving them further options and guidance constructing their selected type of assessment.

Okay, so this is sort of an MVP we plan on building on top of an refining? Right now we are being a little loose on purpose until hammer things out?

> 2\. Oh man I will accept all help with copy, I will try to bring this to the copy caucus again, cause that meet with Lauren was not super helpful in making things more readable.

Writing copy is hard! 😖 Honestly, this was just nitpicking. There were some things that could be improved and a lot of it is hidden in accordions so I don't think it is much of a blocker. Just something for us to keep in mind.

> 3\. I AM hoping we are adding all the validation necessary to enforce the instructions. HOWEVER If your idea is to remove instructions until the scenario happens... like the "I need more than six assessments" I dont hate that. We could remove the dropdown text and put it on the context dialog that appears when you reach the higher limit.

I don't really have a problem with both since most of the instruction is in accordions so they can be ignored. I was more worried that they will be ignored and since our validation is a little lacking it would go through 😅 

Though, I think the current validation is more related to limitations than anything else 😢

yonikid15 (commented):
Great work Peter 👍 This was a tricky one. Went through all the acceptance criteria and other then the limitations you mentioned you nailed every one. 

Eric left a couple comments that I shared as well and I found a possible edge case. Other then that it's very close.

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/AssessmentPlanBuilderPage.tsx
esizer: Not sure if it makes sense to, but what if we moved that above the "Skills quick summary". I think the reason it was just below that heading was what was causing me to associate it with the skills being assessed rather than the actual assessments.

#### Assessment summary
Total assessment count    3

#### Skills quick summary
....

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/AssessmentDetailsDialog.tsx
petertgiles: Added the message in.
image
[error for no skills](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8364/commits/01ad45f1ca812a97353f602c6518f341e741fb8c)

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/AssessmentPlanBuilderPage.tsx
petertgiles: @Jerryescandon do you have thoughts about this?  Should we try to prevent users from interacting with this page if we think the requests will be rejected?  Warning at the top of the page?  Disable the form?  Fail the entire page with an "you are not authorized to view"?  Just don't show the edit link on the advertisement page?  Don't show the edit link in the pool list of the index page?

petertgiles:
> > If so, maybe we should align those terms
> 
> Yup, good catch. These should be aligned.

@Jerryescandon just to confirm, this paragraph should change two instances to the term "application screening"?
image

petertgiles:
> > I might have found an edge case 😢 When adding screening questions on the Edit Pool page, then jumping to the Assessment plan page the error message is shown on the screening questions type. Is this expected behaviour?
> 
> This is actually correct. (For now)

@yonikid15 note that the screening questions on the edit page will become hidden with the feature flag.  I added this in #8403.

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/AssessmentPlanBuilderPage.tsx
petertgiles: I spoke to Jerbo about this yesterday and the decision was just to error out the page if the permission wasn't there, for now.  I added that in [add custom team authorization](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8364/commits/bb24dd8d565931208087e6fbc5ee7e5e8da4f407).  It's a bit hacky since our RequireAuth component can't check teams before the entity is loaded and the team is known (cough, cough, loaders).  I suppose this type of check should eventually be hoisted up and apply to more than just this single page.

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/DeleteDialog.tsx
vd1992, line 1: If you delete an assessment that is middle of the group like say 3rd of 5, your sort order becomes 1, 2, 4, 5
Adding an additional assessment slots it at sort = 6 so it doesn't seem to break anything as far as I can tell. Just a little oddity that might be more trouble than its worth to fix, especially since it doesn't appear to duplicate sort order number

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/AssessmentPlanBuilderPage.tsx
```diff
@@ -97,13 +108,19 @@ export const AssessmentPlanBuilder = ({ pool }: AssessmentPlanBuilderProps) => {
...
+                mode="solid"
+                color="secondary"
+                href={routes.poolView(pool.id)}
+              >
+                {/* Doesn't actually save anything */}
+                {intl.formatMessage({
+                  defaultMessage: "Save plan and go back",
+                  id: "Rbp02p",
+                  description:
+                    "Text on a button to save the assessment plan and return to the pool page",
```
vd1992, line 121: Isn't save buttons that don't do anything just confusing for everyone involved?

vd1992: Particularly when you have a second button next to it, "cancel", that also does the same thing

vd1992:
Given that you cannot update the assessment plan after publishing, I assume publishing will be blocked upon having a completed plan? Therefore should be noted to add into issue #7920 ?

I also can't help but wonder if there should be a small blurb somewhere notifying the user that the assessment plan is finalized alongside the pool when published and isn't a separate modular piece intended to be revised while a pool is open or after closing. But not sure whether you need to be that explicit as a much smaller group of people will be interacting with it

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/DeleteDialog.tsx
petertgiles: Yeah, I agree that it's a little strange to have a sparse ordering.  We do already reserve spot 2 for screening questions so there will can be a gap there normally, too.  I wonder if we should add a unique index to the table to enforce no duplicates?

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/AssessmentPlanBuilderPage.tsx
petertgiles: Yeah, it's weird but we do it on the application, too.  I left that comment to acknowledge that it's not a mistake.

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/DeleteDialog.tsx
vd1992: I think that would be a good idea, but would need some testing to ensure it doesn't unintentionally break something like order swap somehow

yonikid15 (approved):
- [x] Acceptance criteria [from issue](https://github.com/GCTC-NTGC/gc-digital-talent/issues/7916) satisfied.
- [x] A pool skill can be removed on the pool editor page which is already associated with an existing step.
- [x] Application screening step is locked to step 1
- [x] Screening question step is locked (if exists) to step 2
- [x] Steps can be added, removed, and reordered

Excellent work Peter 🏅 looks like you've met all the A/C and nice work with the `moveDisabledIndexes` on the Repeater. I added in one small suggestion with one of the messages. Other then that looks good to go 🚀

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/DeleteDialog.tsx
petertgiles: Yeah, good point - it would prevent the swap from happening.  Now that James removed sortOrder from the mutation and only has swap it's much more strongly managed anyway.  I talked with Jerbo and he's OK with sparse numbering as long as the right order is preserved.

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/AssessmentPlanBuilderPage.tsx
petertgiles: Confirmed with Jerbo - it's a "behavioral save".

petertgiles:
> Given that you cannot update the assessment plan after publishing, I assume publishing will be blocked upon having a completed plan? Therefore should be noted to add into issue #7920 ?

I know you originally wrote this issue for "ready for publishing".  Since you suggested it, I took it over for RoD.  :smile: It's a good callout, though, we need to update the backend validation.

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/AssessmentPlanBuilderPage.tsx
```diff
@@ -173,27 +191,60 @@ export const AssessmentPlanBuilderPage = () => {
...
+        </NotFound>
+      );
+    }
+    if (!authorizedToSeeThePage) {
+      return intl.formatMessage({
+        description:
+          "Heading for the message saying the page to view is not authorized.",
+        defaultMessage: "Sorry, you are not authorized to view this page.",
+        id: "jPLaDk",
+      });
```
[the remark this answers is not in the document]
petertgiles: I did actually reuse it from the routing error messages.  I've refactored it so I could explicitly import it here.
[use shared message](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8364/commits/bd2d264c5033de8c991ae3bedecf48cd9407fcda)

vd1992 (approved):
At last, cross the 🏁

Issue #2651, opened 2022-05-04
Decide on device support


esizer:
## Summary

When developing the IAP homepage, @vd1992 noticed that it did not look great at 320px, as well @patcon found some issues with 4K screens. This started a discussion on what devices we are targeting for the user facing projects.

We should develop some sort of min-max range for device resolutions we are looking to support and add it to our definition of done.

---

## Discussion on Device Support

> Ah, it that case, yeah I wouldn't sweat the polish quite yet. No point worrying with placeholder
> 
> Are the mobile views as desired? Like with the overlaps? Small Mobile on Storybook is a bit unreadable for example

I would say yes but tentatively. This might be a discussion on device support since the storybook mobile small size is 320px pixels and most phone resolutions (even the smaller ones) start at 360px wide.

I believe 320px is comparable to an iPhone 5 (2012). If we do want to support devices between 320-360px then I can go back and do another pass at this width.

_Originally posted by @esizer in https://github.com/GCTC-NTGC/gc-digital-talent/issues/2539#issuecomment-1116369409_

mnigh:
> We should develop some sort of min-max range for device resolutions we are looking to support and add it to our definition of done.

@esizer Has this essentially been decided with the following?
https://github.com/GCTC-NTGC/gc-digital-talent/blob/8605ee3b8ee585c333bada7f98fa1afcfe11398b/frontend/indigenousapprenticeship/src/js/components/Home/Home.stories.tsx#L27-L30

esizer:
> @esizer Has this essentially been decided with the following?

I wouldn't say so. I did have a good reason for picking these resolutions though. Chromatic allows for viewports between 320 and 1800.

What I did, was I looked at screen resolution market share numbers and selected sensible ones between those two values. 

## Min

iPhones currently have the lowest value resolutions at 375-1242 and since it is a popular device on the low end I chose the low value. They do contain a device at 320 (iPhone 5) but tbh, that is really small by todays standards and likely does not have very high market share.

## Max

This one was simple, the most common large display is 1920 but since chromatic only goes to 1800, I picked the next highest which was 1366 with 17.89% desktop market share as of last month.

patcon:
Related: https://github.com/GCTC-NTGC/gc-digital-talent/issues/3576

JoshBeveridge:
[...]

PR #8399, opened 2023-11-02
[Debt] Utilize better image formats

### LINKED ISSUE #6165: Perf - serve images as AVIF and WebP
## ♻️ Debt/Refactor

Lighthouse recommends using AVIF and WebP instead of JPEG and PNG to further reduce image sizes. It notes that compatibility is more limited but there are ways to serve fallbacks. 

## 🕵️ Details

https://developer.chrome.com/docs/lighthouse/performance/uses-webp-images/?utm_source=lighthouse&utm_medium=devtools

## 🙋‍♀️ Proposed Solution

Eric's note:
> Yeah, these formats are just so much better than the alternatives. I wonder if we could create an incredibly rudimentary Image component that takes a file name with available extensions and spits out the <picture> element with the appropriate source sets and fallbacks.

## ✅ Acceptance Criteria

- [ ] Images, especially large ones service in AVIF and/or WebP
- [ ] Fallback to older formats for clients who can't accept newer ones

vd1992:
🤖 Resolves #6165

## 👋 Introduction

Trying out using new image formats. 

## 🕵️ Details

## 🧪 Testing

1. ...
2. ...

## 📸 Screenshot

npm run build main - home page

image

npm run build branch - home page

image

## 🚚 Deployment

- re-running database seeders

> **Notes**
>


apps/web/src/components/FeatureBlock/FeatureBlock.tsx
```diff
@@ -42,14 +56,15 @@ const FeatureBlock = ({ content }: FeatureBlockProps) => {
...
+      {/* <div
         data-h2-height="base(x10) desktop(x12)"
         style={{
           backgroundImage: `url('${content.img.path}')`,
           backgroundPosition: content.img.position || "center",
           backgroundSize: "cover",
         }}
-      />
+      /> */}
+      <ReactImage UrlArray={[content.img.path, content.img.fallback]} />
```
vd1992, line 67: For this instance of background image, I just tried replacing it with an image component, but in general not sure what is the best course of action for background images

vd1992: Like this is a background image and I haven't changed it yet

image

image

vd1992:
Question, thinking about it. Is it even worth trying to account for incompatibility?

My selection, WebP has an estimated compatibility of 96.86 %
The ones missing out are people using IE, Edge from 5 years ago 
https://caniuse.com/webp

Avif isn't supported by Edge at all, score of 85.41%
This one then isn't a good pick
https://caniuse.com/avif

packages/webpack-config/webpack.base.js
esizer, line 1: Looks like my formatter wasn't working when I added the git stuff 😅

package.json
```diff
@@ -54,5 +54,8 @@
     "jest-fail-on-console": "^3.1.1",
     "prettier": "^3.0.3",
     "turbo": "^1.10.16"
+  },
+  "dependencies": {
+    "react-image": "^4.1.0"
```
esizer, line 59: Did you mean to install this in `apps/web` instead?

`npm run i react-image --save --workspace=apps/web`

apps/web/src/components/FeatureBlock/FeatureBlock.tsx
```diff
@@ -18,6 +20,18 @@ interface FeatureBlockProps {
...
+  UrlArray: string[];
+}
+
+const ReactImage = ({ UrlArray }: ReactImageProps) => {
+  const { src } = useImage({
+    srcList: UrlArray,
+  });
+
+  return <img src={src} alt="" />;
+};
```
esizer, line 33: This may be handy as a reusable component in `packages/ui` as something like:

This would also allow us to enforce an `alt` attribute.

```tsx

type ImgProps = Omit<React.ImgHTMLAttributes<HTMLImageElement>, "src" | "alt"> & {
  sources: string[];
  alt: string;
}

const Img = React.forwardRef<HTMLImageElement, ImgProps>(
  ({ sources, ...rest }, forwardedRef) => {
    const { src } = useImage({
      srcList: UrlArray,
    });
    
    return <img ref={forwardedRef} {...rest} src={src} />;
  },
);
```

esizer:
> Question, thinking about it. Is it even worth trying to account for incompatibility?
> 
> My selection, WebP has an estimated compatibility of 96.86 % The ones missing out are people using IE, Edge from 5 years ago https://caniuse.com/webp
> 
> Avif isn't supported by Edge at all, score of 85.41% This one then isn't a good pick https://caniuse.com/avif
> 
> For reference, something like Flex-grow has a slightly lower compatibility score https://caniuse.com/mdn-css_properties_flex-grow

This is a good question and once again raises the question of [what devices/browsers should we support](https://github.com/GCTC-NTGC/gc-digital-talent/issues/2651)? Do we need IE 11 support? How far back does our need for specific versions go?

I don't hate providing a jpg fallback in case someone happens to be on ie11 but does seem to add to our time to paint (i think this is because it is trying to sort out what format to serve). So, while we are sending less data to the client, it seems our task execution time increased.

##Main
Screenshot 2023-11-03 084631

## This branch
Screenshot 2023-11-03 084950

esizer (changes requested):
I made some suggestions but it does seem like this is increasing load time to sort out fallback so I'm not sure how relevant it is 😅

petertgiles:
Fallback question aside, the image size reduction is very impressive!
image

brindasasi:
could we get our browser support documented somewhere?

brindasasi:
Could we test in safari?

vd1992:
Closing in favour of #8405

PR #8402, opened 2023-11-03
[Debt] Update accordion styles

### LINKED ISSUE #6734: ⚙️ Update the Accordion component
image

## ✨ Summary

[Visit the Figma file](https://www.figma.com/file/guHeIIh8dqFVCks310Wv0G/Style-library?type=design&node-id=5-20&mode=design) • Contact @substrae with design questions

Related to [#6735](https://github.com/GCTC-NTGC/gc-digital-talent/issues/6735)

As a part of the experience card updates, the accordion (AKA expansion, drawer, expand-collapse, details, etc.) has been updated to provide a bit more flexibility and usage contexts.

```[tasklist]
### ✅ Requirements
- [ ] Accordions can have two visual styles: simple and card
- [ ] Accordions can be grouped and are spaced as conveyed in the design
- [ ] Accordion groups should have the ability to be toggled all open/all closed (the button to this isn't part of the component, as it varies depending on the context)
- [ ] Accordion labels should NOT be underlined
- [ ] Accordions have the designed focus states
- [ ] All standard accordions on the site have been replaced/updated with this design
- [ ] The spacing has been updated component-wide
- [ ] The divider between trigger and content has been removed
``` 

### Accessibility

- Accordions should announce their interactivity and current state to AT
- Generally, accordion titles should be written as an action (e.g. Show more..., Learn more about..., Reveal information on...)
- Where applicable, accordion titles should be able to change to reflect their state (e.g. show >> hide)

<details>
  <summary>📸 Focus states</summary>
  
</details> 

### Typography

Accordion trigger elements should adapt based on the font-size applied to the title value. The Figma file provides designs for the 3 options currently in use: `copy`, `h6`, and `h5`. Note that the size and stroke on the chevron icon changes to match the heading level.

<details>
  <summary>📸 Typography examples</summary>
  
</details> 

### Grouping accordions

<details>
  <summary>📸 Simple group</summary>
  
</details> 

The older card-style accordion has been deprecated in favor of this more generic card-style that can be used in multiple contexts. 

<details>
  <summary>📸 Card group</summary>
  
</details>

esizer:
🤖 Resolves #6734 

## 👋 Introduction

Makes some minor updates to the accordions.

## 🕵️ Details

The API for accordions updated a bit, some of the changes include:

- Removed components (`StandardAccordionHeader`, `Accordion.Heading`)
- Added new props
- Default mode changed to `"simple"`

### Standard Accordion Header

 we have a `StandardAccordionHeader` that was being used to add some styles that no longer exist for the new accordions. So, it has been removed and no longer needs to be imported.

### Accordion Heading

This has been baked into the `Accordion.Trigger` component.

### New Props

Since we removed the standard header component, we have moved some of the props and around and added some.

- `headingAs` changed to `as` and now supports `p` as a value
- Props moved from `StandardAccordionHeader` to `Accordion.Trigger`
- **New** `size: "sm" | "md" | "lg" = "md";` prop on the `Accordion.Trigger` to control the heading/icon size
- **New** `spaced?: boolean = false` prop on the `Accordion.Root` which adds space between the card mode

### New default mode

We were almost exclusively using the simple mode so it has been made the default now. IF you want to 

## 🧪 Testing

1. Check chromatic diff for funky stuff
2. Build `npm run dev`
3. Check pages that include accordions and confirm they appear as they do in [the mockup](https://www.figma.com/file/guHeIIh8dqFVCks310Wv0G/Style-library?node-id=21%3A2839&mode=dev) 

### Default (simple)

```tsx
<Accordion.Root />
```

localhost_6006_iframe html_args= id=components-accordion--default viewMode=story

### Card

```tsx
<Accordion.Root mode="card" />
```

localhost_6006_iframe html_args= id=components-accordion--card viewMode=story

### Card (spaced)

```tsx
<Accordion.Root mode="card" spaced />
```

localhost_6006_iframe html_args= id=components-accordion--card-spaced viewMode=story


packages/ui/src/components/Accordion/Accordion.stories.tsx
petertgiles, line 1: Could you please add (or update) a story to demonstrate how to integrate an "expand/close all" button?

packages/ui/src/components/Accordion/Accordion.tsx
petertgiles, line 1: One of the requirements:
> Note that the stroke on the chevron icon changes to match the heading level, though its size remains constant.

It seems like the entire chevron size is changing right now?  We might need a map from Josh of all the weights.

image

petertgiles (changes requested):
Nice, this looks so good!  Love how you've rolled the standard header into the component.  I've flagged a few small things but overall it looks amazing.

Without access to NVDA or JAWS I can't really test the "announce" requirement.  Maybe Lee can help you with that testing?  I'd also recommend getting a design review from Josh before merging.

petertgiles:
> New spaced?: boolean = false prop on the Accordion.Root which adds space between the card mode

Is this prop necessary?  I figured if there is a space between the cards it would just be multiple accordions.

packages/ui/src/components/Accordion/Accordion.tsx
esizer: It should be changing but the values are so close you may not notice it much. These were the values I got from the figma. Maybe @substrae can confirm if they are correct?

https://github.com/GCTC-NTGC/gc-digital-talent/blob/9a74fe755c277c44607b9690c2b55c01ff1e66a4/packages/ui/src/components/Accordion/Accordion.tsx#L101-L124

petertgiles: Right, but you're changing the height and width of the icon, too, right?  The requirements say the size of the icon should remain constant.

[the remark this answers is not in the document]
esizer: Good catch! Fixed in [update chevron direction](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8402/commits/1f56232c5141193dd0d37fa9449b6689c47a383c)

packages/ui/src/components/Accordion/Accordion.stories.tsx
esizer: Excellent idea, added in [add expand/collapse all story](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8402/commits/8ab13534527b3f7a82915457fa967177ea2b7d53)

petertgiles: Nice!  That's actually a lot cleaner than I hoped.

packages/ui/src/components/Accordion/Accordion.tsx
substrae: @esizer @petertgiles 

Actually you know what, I think Eric's right in terms of size. Let's adapt them to the heading sizes as Eric described. I'll update the Figma file. Ideally the weight should still change though. Did you need a map for that?

substrae:
> > New spaced?: boolean = false prop on the Accordion.Root which adds space between the card mode
> 
> Is this prop necessary? I figured if there is a space between the cards it would just be multiple accordions. Probably a question for @substrae

Yeah, you're right - there's no need for this, because if there are gaps, they would be separate accordion groups entirely.

packages/ui/src/components/Accordion/Accordion.tsx
esizer: So this was that I used (based on the figma). Let me know if I got it right:

- SM: 2.5 
- MD: 3
- LG: 3.5

petertgiles (approved):
Love it.  I like that you've constrained the size to the root.  Hopefully that doesn't bite us later.  :laughing:

substrae:
@petertgiles 

If it makes you feel any better, we talked about it briefly and there should never really be a single group with different sibling accordions that differ in size. Visually, it would be way too confusing to have arbitrary sizes in amongst the list, and it would be better served to multiple lists, so thinking from that angle, having the size set on the group root should work well.

RM-1978:
> Nice, this looks so good! Love how you've rolled the standard header into the component. I've flagged a few small things but overall it looks amazing.
> 
> Without access to NVDA or JAWS I can't really test the "announce" requirement. Maybe Lee can help you with that testing? I'd also recommend getting a design review from Josh before merging.

@petertgiles  and @esizer,  totally happy to test the "announce" requirement it in NVDA or JAWS :)

Issue #7157, opened 2023-07-03
♻️ Refactor Experience tables to simplify queries


tristan-orourke:
## ♻️ Debt/Refactor

Create a single table which connects experiences to users. This will simplify database queries. This should be done without modifying the api schema.

## 🕵️ Details

Answers on stackoverflow suggest that our Experience model *should* be implemented in the database with a single Experience table for common data, 'extended' by subtype tables. Then the experience_skill table would reference the base table. This would make some of our sql queries simpler! 
- https://stackoverflow.com/questions/59183196/what-is-the-most-efficient-way-to-model-the-relationship-between-these-entities/59232905#59232905
- https://www.softwaregems.com.au/Documents/Article/Database/Relational%20Model/Subtype.pdf
- https://www.softwaregems.com.au/Documents/Documentary%20Examples/IDEF1X%20Introduction.pdf

## 🙋‍♀️ Proposed Solution

https://docs.google.com/drawings/d/1JUPihw8wqnHkir68lXzwZpFYNUUKQrtb2j1PI_kGMGQ/edit?usp=sharing

## 🌎 Localization

(optional) Provide any new copy along with translations available.

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the debt was addressed and expected functionality has not been affected.

- [ ] You can get all the experiences, and all the ExperienceSkills, of a single user with a single query.
- [ ] The api schema is unaffected.

vd1992:
@tristan-orourke 

Did you complete this issue as a side effect?

tristan-orourke:
No, this is different from what I've done.

vd1992:
This could be handy too, example implementation in Postgres
https://stackoverflow.com/questions/5466163/same-data-from-different-entities-in-database-best-practice-phone-numbers-ex/5471265#5471265

PR #8412, opened 2023-11-06
[Feature] Introduce Laravel logging

### LINKED ISSUE #8395: ✨ Log changes to key models
## ✨ Feature

We want to log when important models are created, updated or deleted. We can set this up to happen automatically for our chosen models.

## 🕵️ Details

At minimum, an activity log entry should include
- what entity was changed
- who initiated the change (if it was a logged-in user)
- what attributes were changed

For now, lets add this activity logging to the following models:
- User
- Pool
- PoolCandidate
- Requests

These will give us audit logs on changes to Users, and history on the flow through the pipeline for Pools, applications, and requests.

## 🎨 Design File

### 🧑‍🎨 Designer

## 📸 Screenshot

Use https://spatie.be/docs/laravel-activitylog/v4/introduction.

Set it up to [log model events](https://spatie.be/docs/laravel-activitylog/v4/advanced-usage/logging-model-events).

QUESTION: are there any attributes we should ignore logging? We don't have to worry about passwords, but is there anything else sensitive?
Ans: Log everything except sensitive info. 

QUESTION: do we want to add any special descriptions or values to certain events? For example, we could treat "change status" as a special event, with its own description or key/values, instead of just another UPDATE event. 
Ans: No

QUESTION: Its possible when setting up model-event logging to log changes to related models as well. We should consider if we want to do that now for any of these models. 
Ans: We could split that into sep ticket. No for now. Please log changes to applicant filter as part of search requests. 

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Basic model-event activity logging is set up for Users, Pools, PoolCandidates, and PoolCandidateSearchRequests
- [ ] Log everything and add the exception for sensitive columns ( password)
- [ ] Single activity_log table for all activity logs. Include Model name as part of each record so that we can filter that out. 
- [ ] Sanity tests to verify the logging as diff user ( test to log as per the logged in user , activities are logged properly etc)

vd1992:
🤖 Resolves #8395 

## 👋 Introduction

Add logging to four models: User, PoolCandidate, Pool, and PoolCandidateSearchRequest

## 🕵️ Details

Mostly just following instructions to implement with a few changes. 
One special addition being to store the related applicantFilter relation for requests. 
Added testing too

## 🧪 Testing

1. `php artisan migrate`
2. Run some mutations
3. Open up the activity_log table and find your actions

## 📸 Screenshot

image

## ❗ Note

Not sure what is going on with the pool model, its subject type doesn't match the model file unlike the rest. I think this is pretty automatic so not sure about the cause

image


esizer (commented):
Looking great! I just have a couple questions,

1. I can't seem to figure out why events related to pools are `poolSkills` and not `App\Models\Pool`. It is slightly confusing since I never actually changed the skills while testing 😕 
2. Do we want to setup a cron job to run clean periodically? Or is this a retention policy this?

esizer:
> 1. I can't seem to figure out why events related to pools are `poolSkills` and not `App\Models\Pool`. It is slightly confusing since I never actually changed the skills while testing 😕

I figured this out. The package uses the morph name as the subject and we are morphing it in our app service provider:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/158e85b63dc9ca6a355e7b380c658493e60b24e2/api/app/Providers/AppServiceProvider.php#L28-L35

I don't think this is a big deal as long as we are aware. 🤷‍♀️ What do you think?

api/tests/Feature/ActivityLogTest.php
petertgiles, line 1: Super-solid tests.  Nice.

documentation/activitylog.md
petertgiles, line 1: Love this.

petertgiles (commented):
This turned out really nice.  We're missing a lot of the pool candidate relations but this is a good way to start.  I hope we can figure out the weird pool model name.

petertgiles:
Actually, I wonder if any of these entries should be in the morph map at all?  It's not actually changing the name of anything - it's just shortening them?

esizer:
> Great find! I hate it. 😆 I can kind of understand why it is using the morph map - it makes sense. I wonder if the problem here is that pool shouldn't be in the morph map? There are no polymorphic relationships with pool, right?

Yeah, I don't really think it needs to be there 🤔 Also, isn't the goal to eventually not need any polymorphic relations since all experiences will be based on a single root table we could add the relationship to?

 - https://github.com/GCTC-NTGC/gc-digital-talent/issues/7157

vd1992:
Thank you Eric, that was a quick fix
https://github.com/GCTC-NTGC/gc-digital-talent/pull/8412/commits/889c499952bb8fab2082f84305722e937f859c24

Definitely will leave the experiences morphMap alone for that other issue, otherwise things get weird

image

petertgiles (approved):
Nice.  I'm excited to see this working in prod.

Issue #8419, opened 2023-11-07
✨ New stream for executives


gobyrne:
## ✨ Feature

Add a new stream

## 🕵️ Details

New stream with the following user-facing copy:
EN: Executive Group FR: Groupe de la direction

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the feature was properly implemented.

- [ ] New stream is available on the platform (for pool creation, for example)

PR #8422, opened 2023-11-08
add Executive Stream

brindasasi:
🤖 Resolves {#8419}

## 👋 Introduction

adds new stream "Executive Group"
## 🕵️ Details

Just added an enum and translation for it. 

## 🧪 Testing

1. Verify Executive Group shows up while pool creation and you can successfully save/edit the pool information with it.
2. Verify It also shows up in Search page stream dropdown and you can successfully do the search with it.
3. Check in the candidates filters and anything else you can think of. 

## 📸 Screenshot

Screenshot 2023-11-08 at 4 34 39 PM
Screenshot 2023-11-08 at 4 34 13 PM
Screenshot 2023-11-08 at 4 35 54 PM


vd1992 (approved):
Ran through things, creating pool -> apply -> add to qualified -> find candidate on search page. 
Works! Nice!

image

PR #8451, opened 2023-11-09
[Fix] Properly assign `aria-selected` to multi select combobox

### LINKED ISSUE #8424: ⌨ Multiselect combobox items set aria-selected on the highlighted item
## ⌨ Accessibility Issue

The manual accessibility check of the 'SkillBrowser' component (Issue [#8001](https://github.com/GCTC-NTGC/gc-digital-talent/issues/8001)) found that its Multiselect combobox items set aria-selected on the highlighted item, not on selected items. This is a bug with downshift.  We may need to hack around this. 

### Accessible Rich Internet Applications (WAI-ARIA) Criteria 

[6.5.1. Widget Attributes  - ARIA selected states](https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-selected)

## 🦋 Expected Behaviour

Only selected items are tagged as an [aria-selected (state)](https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-selected). 

## 🕵️ Details

- **Operating System**: Windows, MacOS.
- **Browser**: Chrome, Edge, Firefox and Safari.
- **Assistive Technology**: NVDA, Narrator and VoiceOver

## 📸 Screenshot

Image

## 🙋‍♀️ Proposed Solution

Our multi select list should have a aria-multiselectable=true on the ul[role=listbox], items in the list should have aria-selected values set to true when selected and false when not selected.

This is likely a high priority since screen readers announce that the item is selected regardless of its actual state since downshift adds aria-selected=true to the highlighted item which is very confusing.

## ✅ Acceptance Criteria

- [ ] Only selected items have ARIA selected state set to true.

esizer:
🤖 Resolves #8424 

## 👋 Introduction

This fixes a bug where downshift was applying `aria-selected=true` to the highlighted item and not the selected item.

## 🧪 Testing

2. Navigate to the multi combobox story
3. Open the dropdown and select some items
4. Confirm only the selected items in the list have `aria-selected` set to `true`

### DIFF
--- a/packages/forms/src/components/Combobox/Menu.tsx
+++ b/packages/forms/src/components/Combobox/Menu.tsx
@@ -115,6 +115,7 @@ const Item = React.forwardRef<HTMLLIElement, ItemProps>(
   ({ active, selected, children, ...rest }, forwardedRef) => (
     <li
       ref={forwardedRef}
+      role="option"
       data-h2-display="base(flex)"
       data-h2-align-items="base(center)"
       data-h2-gap="base(0 x.25)"
--- a/packages/forms/src/components/Combobox/Menu.tsx
+++ b/packages/forms/src/components/Combobox/Menu.tsx
@@ -3,12 +3,12 @@ import { useIntl } from "react-intl";
 import CheckIcon from "@heroicons/react/20/solid/CheckIcon";
 import ArrowPathIcon from "@heroicons/react/24/solid/ArrowPathIcon";
 import { motion, useReducedMotion } from "framer-motion";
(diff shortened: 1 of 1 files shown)


vd1992 (approved):
Think things are good. 
Only concern is, how or when do we revert this, if ever? Should we file a bug in their repo? 

image

esizer:
> Only concern is, how or when do we revert this, if ever? Should we file a bug in their repo?

Yeah, it is on the roadmap for v9. So we just need to wait until that.

PR #8471, opened 2023-11-10
[Debt] Add validators for unique values

### LINKED ISSUE #7239: ♻️ Validators for uniqueness check
## ♻️ Debt/Refactor

Properly implement uniqueness validation for applicable fields. 
The database will take care of things, but validators will stop failing inputs from ever annoying the database. 

## 🕵️ Details

Mimicking how it was done in

`api/app/GraphQL/Validators/UpdateUserInputValidator.php`

The mutation being acted on needs its id for the ignore function. 
The ignore is needed to avoid erroneous validation errors, like updating a model but not changing its unique fields. 

## 🙋‍♀️ Proposed Solution

This applies to field `department.department_number`

There are other unique fields present that may benefit from this too, they can be considered too. 
You can find them by searching, some hits for searching give irrelevant results like factory usage.

Another example would be `skills.key` 
`skill_families.key` already has validation and can be used as reference

image

image

image

## ✅ Acceptance Criteria

- [ ] Validation and validators added where applicable for unique fields
- [ ] Mutations not broken

esizer:
🤖 Resolves #7239 

## 👋 Introduction

This adds missing validators for unique values in our schema.

## 🕵️ Details

I believe I got all the ones that are possible through the API but let me know if I missed any!

I added them for:

- `Skill.key`
- `Team.name`

Also, we were missing a test case for creating duplicate skill keys (but not teams) so a test was only added for skills.

## 🧪 Testing

1. Build `npm run dev`
2. Attempt to create a skill with a duplicate key
3. Confirm error is returned
4. Attempt to create team with duplicate name
5. Confirm error is returned

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -1133,12 +1133,9 @@ input CreateTeamInput {
   contactEmail: Email @rename(attribute: "contact_email")
 }
 
-input UpdateTeamInput {
+input UpdateTeamInput
+  @validator(class: "App\\GraphQL\\Validators\\UpdateTeamInputValidator") {
   name: String
-    @rules(
-      apply: ["unique:teams,name"]
-      messages: [{ rule: "unique", message: "TeamNameInUse" }]
-    )
   displayName: LocalizedStringInput @rename(attribute: "display_name")
   description: LocalizedStringInput
   departments: DepartmentBelongsToMany
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -1103,10 +1103,7 @@ input CreateDepartmentInput {
   name: LocalizedStringInput
 }
 
-input UpdateDepartmentInput
-  @validator(
-    class: "App\\GraphQL\\Validators\\UpdateDepartmentInputValidator"
-  ) {
+input UpdateDepartmentInput {
   departmentNumber: Int @rename(attribute: "department_number")
   name: LocalizedStringInput
 }
@@ -1136,8 +1133,7 @@ input CreateTeamInput {
   contactEmail: Email @rename(attribute: "contact_email")
 }
 
-input UpdateTeamInput
-  @validator(class: "App\\GraphQL\\Validators\\UpdateTeamInputValidator") {
+input UpdateTeamInput {
   name: String
   displayName: LocalizedStringInput @rename(attribute: "display_name")
   description: LocalizedStringInput
@@ -1588,7 +1584,7 @@ type Mutation {
   updateDepartment(
     id: ID!
     department: UpdateDepartmentInput! @spread
-  ): Department @update @can(ability: "update", find: "id")
+  ): Department @update @can(ability: "update", find: "id") @validator
   deleteDepartment(id: ID! @whereKey): Department
     @delete
     @guard
@@ -1761,6 +1757,7 @@ type Mutation {
   updateTeam(id: UUID!, team: UpdateTeamInput! @spread): Team
     @update
     @can(ability: "update")
+    @validator
   deleteTeam(id: UUID! @whereKey): Team
     @delete
     @guard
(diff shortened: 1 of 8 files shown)


api/graphql/schema.graphql
```diff
@@ -1131,6 +1135,10 @@ input CreateTeamInput {
 
 input UpdateTeamInput {
   name: String
+    @rules(
+      apply: ["unique:teams,name"]
+      messages: [{ rule: "unique", message: "TeamNameInUse" }]
+    )
```
vd1992, line 1141: This blocks updating a team even if the name stays the same. 
Validating updates I believe requires the validation class with ignores like so. 

image

vd1992 (changes requested):
I think the scope of the issue involves validation classes for update operations, like the example department number in the issue body. 
Updating a department to a number in use for instance, gives a not very helpful error as it outputs the SQL error to console

image

esizer:
> I think the scope of the issue involves validation classes for update operations, like the example department number in the issue body.

Good catch, added in [validate unique department number](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8471/commits/288057245f3fca1db539e679eb4e6f9791fd15c6)

api/graphql/schema.graphql
vd1992: Ah, doesn't quite work yet

image

Will need to go on the mutation itself as you need the validator to be able to see the input ID. For this mutation the id is not part of the input object. 

image

While the other ones the id is part of the input object so you can stick the validator on that. This is the pattern I think we want going forward but refactoring existing mutations isn't something to worry about yet. 

image

vd1992: The same applies to the updateDepartment mutation

packages/i18n/src/messages/apiMessages.ts
vd1992, line 1: Please add a message for the department number too

esizer: There is already 😅 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/288057245f3fca1db539e679eb4e6f9791fd15c6/packages/i18n/src/messages/apiMessages.ts#L57-L63

vd1992: Oops, sorry 😆

api/graphql/schema.graphql
esizer: Oh, yeah. That makes sense. Moved in [move validators to top level mutations](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8471/commits/e3a876d13236fe2e475b1bdd1f62fc914344ffa1)

api/tests/Feature/SkillTest.php
```diff
@@ -179,6 +179,49 @@ public function test_create_skill()
             ->assertJsonFragment(['name' => $variables['skill']['name']]);
     }
 
+    /**
+     * Test that non-unique keys fail
+     *
+     * @return void
+     */
+    public function test_create_non_unique_skill_key()
+    {
```
vd1992, line 188: Question, why this casing, just nitpicking as camel-case seems to be our standard otherwise

esizer: I just copy and pasted from a test earlier in the file 🤷‍♀️ 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/e3a876d13236fe2e475b1bdd1f62fc914344ffa1/api/tests/Feature/SkillTest.php#L105

vd1992: Huh, oh weird, guess that file has its own convention going 🤷

vd1992 (approved):
Putting through its paces, the operations in question appear to work right and validate in correct situations. 

Nice!

api/tests/Feature/SkillTest.php
esizer: You could create a ticket to make our methods [PSR compliant](https://www.php-fig.org/psr/psr-1/#43-methods) 😅

vd1992: Done #8479

PR #8473, opened 2023-11-10
[Dependencies] Jwt-decode 4.0.0

### LINKED ISSUE #8396: ♻️ Bump `jwt-decode` to v4
## ♻️ Debt/Refactor

The `jwt-decode` hase updated with some breaking changes. We will need to do this manually.

## 🕵️ Details

Should be pretty easy, only one real [breaking change](https://github.com/auth0/jwt-decode/releases/tag/v4.0.0).

## ✅ Acceptance Criteria

- [ ] `jwt-decode` is at  v4+
- [ ] App builds and functions as expected

vd1992:
🤖 Resolves #8396

## 👋 Introduction

Updates the dependency from 3.1.2 to 4.0.0

## 🕵️ Details

The applicable breaking change was the import no longer being a default export. 
https://github.com/auth0/jwt-decode/releases/tag/v4.0.0

## 🧪 Testing

1. Login and logout, so auth tasks succeed still

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21157,8 +21157,12 @@
       }
     },
     "node_modules/jwt-decode": {
-      "version": "3.1.2",
-      "license": "MIT"
+      "version": "4.0.0",
+      "resolved": "https://registry.npmjs.org/jwt-decode/-/jwt-decode-4.0.0.tgz",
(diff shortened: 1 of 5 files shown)


vd1992:
The spec `admin-workflows.cy.ts` failed once, investigating and it appears to be a side effect of the test not always playing nice with the new table. When loading, the text is obscured but still detected by Cypress, so `cy.contain` does nothing to make it wait for loading to stop. 
Re-run was successful. I concluded the failure did not have to do with the PR. 

image

esizer (approved):
LGTM! 👍

PR #8482, opened 2023-11-15
[Feature] Screening questions and no skills

### LINKED ISSUE #8367: ✨ Make skills optional on assessment step
## User story
As a pool operator, I want to create a screening questions with no associate skills so I can ask relevant questions to application process that aren't part of the actual assessment. For example: "are you willing to submit yourself to a polygraph test" could be required for some high security positions but has nothing to do with demonstrating a skill.

## ✨ Feature

Make adding skills optional on the screening question section of the assessment plan builder.

## 🕵️ Details

Currently all sections of the assessment plan builder require that you choose at least one skill before you can save. We want to remove this requirement, but only for the screening questions assessment.

### 🧑‍🎨 Designer

@Jerryescandon 

## 📸 Screenshot

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Adding skill is optional for screening question step

vd1992:
🤖 Resolves #8367

## 👋 Introduction

Checks and adjusts for desired screening question behaviour. 

## 🕵️ Details

The assessment step for screening question mutation can successfully take no skills and that was checked via test.
Likewise asserted for assessment result, this is because the skill is only required for result type = skill. 

This made me think the most important change would be a third result type, right now there is only skill and education. For a hypothetical like polygraph need a third or other category. Education won't work and neither would skill without an actual skill to match it. 

Updated the assessment plan builder too. 

## 📸 Screenshot

image

## 🧪 Testing

1. I leave this up to you, @petertgiles to tell me what the ~people~ frontend need :salute


petertgiles:
Wow, impressively light touch with this one!  I think it's going to need a little more, though.

First off, you'll need to update the assessment step builder to allow you to create the step without skills attached.  I suggest waiting until #7916 is merged to do that.

Second, I'm not sure about adding the "OTHER" option to the enum.  More of a @Jerryescandon question though.  If a screening question step has no associated skills, does it still use the skill assessment decision dialog?  Or do we really treat it as a third type of assessment decision?
image

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/AssessmentDetailsDialog.tsx
```diff
@@ -607,16 +612,29 @@ const AssessmentDetailsDialog = ({
...
+                    idPrefix="assessedSkillsScreeningQuestions"
+                    id="assessedSkillsScreeningQuestions"
+                    name="assessedSkillsScreeningQuestions"
+                    legend={intl.formatMessage(labels.assessedSkills)}
+                    items={assessedSkillsItems}
+                  />
+                )}
+                {selectedTypeOfAssessment !==
+                  AssessmentStepType.ScreeningQuestionsAtApplication && (
+                  <Checklist
```
vd1992, line 627: One conditional input wasn't working, as far as I could tell the rules object somehow appeared to be persisted across both iterations.

petertgiles:
Decision from meeting with Jerbo, Gray, Vachan, and I: An assessment step with no skills to assess should automatically be assessed with a "success" when the application is submitted.  Should not even show up in the screening and assessment interface.

vd1992:
Enum renamed

petertgiles (changes requested):
Nice one, this works well.  I think we need to kick the enum down the road a bit now that we know that users won't be submitting assessment results for these but the frontend and tests work well.

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/AssessmentDetailsDialog.tsx
petertgiles: If it's acting weird, though, I'm fine to leave this structure.

api/tests/Feature/AssessmentResultTest.php
```diff
@@ -400,4 +401,30 @@ public function testSkillAssessmentRequiresSkill(): void
             )
             ->assertGraphQLValidationError('createAssessmentResult.poolSkillId', 'SkillAssessmentResultMissingSkill');
     }
+
+    // test screening question result and no skills
+    public function testScreeningQuestionResult(): void
```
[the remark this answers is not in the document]
vd1992: No it does not fit in anymore, at least for now. So I have removed it and the enum in this https://github.com/GCTC-NTGC/gc-digital-talent/pull/8482/commits/b0c46457ea70a3713ed884b592ec76f91655de02
And yeah, can see how the result side of things pans out

petertgiles (approved):

Issue #8423, opened 2023-11-01
🔤 Plain language copy updates to the `SkillBrowser`


RM-1978:
## 🔤 'SkillBrowser' copy issue 

Three copy updates to the skills browser. See acceptance criteria.

## 🎨 Design Files

> [Skills Library (Skill picker dialogue)](https://www.figma.com/file/5LcwA4AmOM8ebdJKuiPs8g/Skills-library?type=design&node-id=155-5147&mode=design&t=1E5MWAvwf6py5aYJ-0)
> [Talent Search](https://www.figma.com/file/9V8kBIWxOXRaN9dkcVCU1o/Manager---Talent-search?type=design&node-id=204-1593&mode=design&t=1E5MWAvwf6py5aYJ-0)
> [Job Application V6 (Select an experience) ](https://www.figma.com/file/BXYVsu1fo9LS2IySJDUINY/Candidates---Job-application---Version-1-to-6?type=design&node-id=547-49171&mode=design&t=1E5MWAvwf6py5aYJ-0)
> [Pools Record of Decision (Add Essential and Asset Skill modals)](https://www.figma.com/file/em5ArPr044YaIe1Vb1zCZz/Pools---Record-of-Decision?type=design&node-id=3513-11411&mode=design&t=1E5MWAvwf6py5aYJ-0)

### 🧑‍🎨 Designers

@RM-1978 

## 📖 Storybook component

> [SkillBrowser](https://61e099a1bb1465003a73d3ce-sdnauqrpyl.chromatic.com/iframe.html?id=components-skill-browser-inline--default&viewMode=story)

### 👩‍💻  Developer 

Eric Sizer (@esizer) 

## 🕵️ Details

Three copy instances where users: 

- "Select an option" null state label being read as: "disabled selection option 1." AT users cannot easily determine that they are being presented with filtering options.  
- Bracketed numbers at the end of each option are confusing  without context, and conflict with the selection option numbering read and announced by screen readers.
- Once you select this option, all skills from every family appear. The skills are not presented by family.

## ✅ Acceptance Criteria

- [ ] Update the null state with the component context: "Select a skill family"
- [ ] Update "All skill families" to read " All skills." 
- [ ] Add the word "skills" to the remaining menu options beside the count (i.e., 10 skills).

gobyrne:
@Jerryescandon Update this one too please :)

PR #8525, opened 2023-11-16
♻ [Debt] Replace search page skill filter

### LINKED ISSUE #7352: ✨ Implement `SkillBrowser` component inline version
## ✨ Feature

Implement inline version of `SkillBrowserDialog` called `SkillBrowser` component on `SearchPage`.

## 🎨 Design File

[Talent Search](https://www.figma.com/file/9V8kBIWxOXRaN9dkcVCU1o/Manager---Talent-search?type=design&node-id=204-1593&mode=design&t=1E5MWAvwf6py5aYJ-0)
[Job Application V6 (Select an experience)](https://www.figma.com/file/BXYVsu1fo9LS2IySJDUINY/Candidates---Job-application---Version-1-to-6?type=design&node-id=547-49171&mode=design&t=1E5MWAvwf6py5aYJ-0)
[Pools Record of Decision (Add Essential and Asset Skill modals)](https://www.figma.com/file/em5ArPr044YaIe1Vb1zCZz/Pools---Record-of-Decision?type=design&node-id=3513-11411&mode=design&t=1E5MWAvwf6py5aYJ-0)

## Component  Screenshot 

image

## ✅ Acceptance Criteria

- [ ] Implement inline `SkillBrowser` component on `SearchPage`
- [ ] Remove the old component from the code

esizer:
🤖 Resolves #7352 

## 👋 Introduction

This replaces the old skill picker on the search page with the new, combobox version.

## 🕵️ Details

This was in a separate component due to some handlers and state we needed since the old skill picker was not a form input. Now that we are using a form input, we no longer need that logic so it was moved to be more inline with other inputs on that page, using the `FilterBlock` component.

### Related ticket: 
#8423 

## 🧪 Testing

1. Build `npm run dev`
2. Navigatge to `/search`
3. Fill out the form
4. Confirm the following:
    - [x] Interacting with skills affects the API request
    - [ ] Filter advice functions as expected
    - [x] Skills submitted show up within the request

## 📸 Screenshot
Screenshot 2023-11-16 122659


mnigh (commented):
Cool! A few questions that might be out of scope since they deal with how the component itself works, but I thought are important.

## Order of options
The **All skill families** option does not appear at the beginning of the list in French because it is sorted alphabetically. The `getFamilyOptions` function is prepending the **All skill families** option to the beginning but when called in `SkillSelection` and `SkillBrowser`, the options are being sorted alphabetically. The list of options in English only renders the **all** option first because there aren't any options that are ahead of it alphabetically, but if something like Agile were added as a skill family, **all** would no longer appear at the beginning of the options list.
 
### Screenshot
skill_sort

## Selected skills visibility
Any selected skills, although persisted, are not prominent when returning to the component from the request page unless a **Skill family** option is selected (which it is not by default). Should **all** be selected by default?

### Screen recording

https://github.com/GCTC-NTGC/gc-digital-talent/assets/3046459/a03dead1-3816-471e-8e52-c2d522851360

esizer:
> The **All skill families** option does not appear at the beginning of the list in French because it is sorted alphabetically. The `getFamilyOptions` function is prepending the **All skill families** option to the beginning but when called in `SkillSelection` and `SkillBrowser`, the options are being sorted alphabetically. The list of options in English only renders the **all** option first because there aren't any options that are ahead of it alphabetically, but if something like Agile were added as a skill family, **all** would no longer appear at the beginning of the options list.

This was a good catch! I actually never noticed it was doing that 😅 Should be fixed in [fix sorting of skill browser family filter options](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8525/commits/a5e2d8f82cd9ae81fdee749d96c05133209e4980)

RM-1978:
@gobyrne, since we require the user to select a skill family before we present the skills, can we make the default selection "All skills or "All skills families" instead of "select an option". It would be a lot more efficient for the user.

petertgiles:
Decision from scrum: if there is a skill selected, have "all skill families" selected by default.

esizer:
> Any selected skills, although persisted, are not prominent when returning to the component from the request page unless a **Skill family** option is selected (which it is not by default). Should **all** be selected by default?

@mnigh , Updated to automatically set the family to "all" if the skills have been selected in [set family to "all" if skills have a value](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8525/commits/10c1330393e8c2f20550a0fc652c255787fab9f2).

Also, fixed sorting in dialog in [fix option sorting for skill dialog](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8525/commits/1c775332f5462be5cdbef21306f6041d24f114e5)

vd1992 (approved):
Not sure if Matt is re-reviewing, but this looks great to me. Works without any quirks I can notice and Matt's keen observations appear to be resolved

mnigh (approved):
👍

PR #8527, opened 2023-11-16
[Debt] Upgrade to PHP 8.2

### LINKED ISSUE #6003: Upgrade to PHP 8.2
> __Warning__
> PHP 8.2 will be unavailabe for the Ubuntu images until Mantic 23.10.  Then it can be used for the maintenance container.

## ♻️ Debt/Refactor

We are required to update our php App Service apps in azure to PHP 8.2 by Nov 25 2024. We should update our docker image at the same time, to confirm locally that everything will still work.

Let's upgrade to PHP 8.2.  It has been out for a few months and is available in Azure.

## 🕵️ Details

https://www.php.net/releases/8.2/en.php

image

## ✅ Acceptance Criteria

- [x] Webserver container uses PHP 8.2
- [x] Maintenance container uses PHP 8.2
- [x] Github CI uses PHP 8.2
- [ ] all three Azure servers use PHP 8.2
- [ ] Everything functions as before

petertgiles:
🤖 Resolves #6003

## 👋 Introduction

This branch updates our PHP platform to version 8.2.

Note that bumping the composer platform required running `update` which resulted in the following dep bumps:
- Upgrading lcobucci/clock (3.0.0 => 3.1.0)
- Upgrading phpunit/php-code-coverage (10.1.7 => 10.1.8)
- Upgrading webonyx/graphql-php (v15.7.0 => v15.8.0)

### Locally

1. `docker-compose down` to bring down your network
2. `docker-compose build maintenance && docker-compose build webserver` to update your images
3. `docker-compose up` to bring your network back up again

- [ ] Setup and refresh scripts work as before
- [ ] Can run tests in the maintenance container `docker-compose run --rm --workdir=/var/www/html/api maintenance bash -c "./artisan test"`

:warning: Note that Cypress can't run headless in our current maintenance container either.

### In Azure

1. Bring out your tablet

- [ ] Azure IaC changes looks good.
- [ ] A redeploy of the Dev app svc successfully brings up a PHP 8.2 instance.
- [ ] App functions properly on dev vertical.

## 🚚 Deployment

- [ ] Change PHP version to 8.2 in uat and prod IaC definitions
image
- [ ] Change default PHP version to 8.2 for shared vertical definition
image


infrastructure/maintenance-container/Dockerfile
```diff
@@ -2,14 +2,14 @@
 FROM node:18.16.0 as node
 
 # base
-FROM ubuntu:22.04
+FROM ubuntu:23.10
```
vd1992, line 5: For consistency purposes, I think all the workflows with set Ubuntu versions ought to be updated too

petertgiles: I don't think 23.10 is available for Github workflows.  The latest is 22.04.  They probably just stick to the LTS releases.  
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

It's not as big a deal since we use the shivammathur/setup-php@v2 action to specify the version of PHP we want to run in the workflow.

petertgiles:
Will move into the "ready to review" column when the dev database comes back up.

infrastructure/maintenance-container/Dockerfile
vd1992: Ah okay, something to come back to later then

petertgiles:
The dev database is back up now.

vd1992:
Locally

- Image build successful, need to run `docker-compose up --detach` instead
- API refresh succeeds
- Tests in the maintenance container succeeds
- Testing and migrate/seeding in the Docker interface also succeeds
- npm run build succeeds
- Cypress runs successfully
- Can login and bounce around the site without issue

Dev 

- Can visit site
- Bouncing around talent side and site loads
- I see an error at /en/search
"internal server error" in the console, `"path" : ["countPoolCandidatesByPool", 0, "pool", "essentialSkills"]`
Maybe data related and nothing to do with PHP 8.2?  
- See an error toast for internal server error at /en/browse too, "oh no graphql" 

Not sure whether to worry about the dev results?

petertgiles:
> Not sure whether to worry about the dev results?

OK, I've reset dev.  To do that I had to update the UAT seeder by adding some missing seeders.  I also had to remove the UAT pool seeder since you can't create a pool when there is no user to own it.  Ping me when you've logged in once and I'll promote your user to admin.

vd1992 (approved):
Can't see any issues, looks fine to me 👍

PR #8529, opened 2023-11-16
[Bug] Handle null dates better

### LINKED ISSUE #8528: 🐛 [Prod] The view candidates page crashes upon clicking on select all check box
## 🐛 Bug

This issue was found in prod on view candidates page , when the user clicks on download all checkbox. 
Reason is one of the candidate  is having award experience as null for awarded_date field. 

## 🦋 Expected Behaviour

Page should handle invalid date format exception properly and not let the page crash

## 🕵️ Details

Front end doesn't allow empty experience to be added. But there are 2 occurrences in the prod which has empty award experience. 

## 📋 Steps to Reproduce

1. Create an award experience manually in the db which has awarded_date value as null;
2. Add this candidate to any pool 
3. Open that pool candidates table 
4. Click on the top most check box to _select all_ 
5. Observe the page crashes 

## 📸 Screenshot

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK3JpQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f86bccc0f3d0f9ebfd96cb687dce00141d100a00/Capture.PNG)

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] Page should handle invalid date format exception properly and not let the page crash

vd1992:
🤖 Resolves #8528

## 👋 Introduction

Better handle null dates. 

## 🕵️ Details

`awardedAt` is nullable, so changed handling to pass into the function after confirming non-null
Also added a null check into the date function to double check

## 🧪 Testing

1. Cannot replicate the issue @brindasasi found


apps/web/src/utils/dateUtils.ts
```diff
@@ -5,12 +5,18 @@ import {
...
 
 import { Maybe, Scalars } from "~/api/generated";
 
 export function formattedDate(date: Scalars["Date"], intl: IntlShape) {
   let dateString = date;
 
+  // handle accidental passing in nulls
+  if (date === null) {
+    return intl.formatMessage(commonMessages.notAvailable);
+  }
```
vd1992, line 18: null isn't type-wise allowed by the function as a value for `date`, but clearly that is leaking

brindasasi: can we check for undefined also?

brindasasi: when the error happened in prod , it had `undefined` value for the experience.awardedDate.
It survived that ?? check in the above line , went into formattedDate where it blew up.
I think we can manually set the value undefined and test this.

vd1992: Add testing? Sure, here https://github.com/GCTC-NTGC/gc-digital-talent/pull/8529/commits/7b38364ee9c38d9dfb898c11a1be8be8e892b855

brindasasi: Awesome.. I was able to reproduce this error in my local. 
And verified this pr solved the issue. 
Before 
Screenshot 2023-11-20 at 10 36 50 AM

After 
Screenshot 2023-11-20 at 10 41 56 AM

brindasasi: Also career time line was also crashed without the date in award experience 
With this Pr , it can also load without date
Screenshot 2023-11-20 at 10 45 19 AM

brindasasi (approved):
Perfect! 👏🥳

PR #8531, opened 2023-11-17
[Fix] Change pool buttons from tertiary to secondary

### LINKED ISSUE #8460: 🐛 Buttons using positive actions are using negative colour
## 🐛 Bug

Create and save actions are using red buttons. Red buttons are usually interpreted as negative actions.

## 🦋 Expected Behaviour

They should not use red which is used for negative actions such as delete.

## 📋 Steps to Reproduce

1. Navigate to `/admin/pools/create`
2. Observe red button for creating a pool

## 📸 Screenshots

Screen Shot 2023-11-09 at 12 04 16
Screen Shot 2023-11-09 at 12 04 40
Screen Shot 2023-11-09 at 12 06 46

## ✅ Acceptance Criteria

- [ ] non-destructive buttons should not be red and replaced by`secondary` colour

petertgiles:
🤖 Resolves #8460 

## 👋 Introduction

Some buttons on the admin pool pages were using tertiary instead of secondary.  This branch fixes that.

## 🧪 Testing

1. Build and log into admin.
2. Check the buttons mentioned in the issue.

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/pages/Pools/CreatePoolPage/CreatePoolPage.tsx
+++ b/apps/web/src/pages/Pools/CreatePoolPage/CreatePoolPage.tsx
@@ -181,7 +181,7 @@ export const CreatePoolForm = ({
               />
               <div data-h2-align-self="base(flex-start)">
                 <Submit
-                  color="tertiary"
+                  color="secondary"
                   text={intl.formatMessage({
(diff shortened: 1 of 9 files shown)


esizer (approved):
LGTM! 👍

PR #8536, opened 2023-11-20
[Debt] Moved experiences from page query to SkillMatchDialogs

### LINKED ISSUE #8523: ✨ Pool Candidates Table - lazy load experiences for dialog
## ✨ Feature

Our pool candidate table is requesting a lot of data upfront.  This can lead to timeouts and poor performance.  We hope to stop requesting the pool candidate experience data upfront and only load it when the dialog is opened.

Context: https://talent-cloud.slack.com/archives/C0259G6KXJ6/p1700065749125969

## 📸 Screenshot

image

> I think we could just pass the user ID into the dialog and run the query on mount :+1: That would be a super easy way to reduce the size of the query. Honestly, we could be doing this a lot more with hidden data. Our current pattern of querying all the data at the page level may start to bite us (or already has) once we start scaling! :catshocked:
<Dialog userId={candidate.user.id} />

`const result = useExperienceQuery({ variables: { userId } });`

## ✅ Acceptance Criteria

- [ ] Pool candidate table query does not request experience data.
- [ ] Experience dialog loads experience data for the single candidate when it is opened.
- [ ] Make sure the query still returns the aggregated skill count , we just fetch the experience data on demand

JamesHuf:
🤖 Resolves #8523

## 👋 Introduction

Removes experiences from the main `PoolCandidatesTable` query moving them to a smaller query when opening the `SkillMatchDialog`.

## 🧪 Testing

1. Navigate to the pool candidates page and add a skill to the skill filter.
2. Confirm Experiences are not included in the main query.
3. Open the `SkillMatchDialog` for one of the candidates and confirm a smaller query runs for just that candidates experiences.

### DIFF
--- a/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
+++ b/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
@@ -713,8 +713,8 @@ const PoolCandidatesTable = ({
             allSkills?.filter(
               (skill) => filteredSkillIds?.includes(skill.id),
             ) ?? [],
-            user.experiences?.filter(notEmpty) ?? [],
             skillCount,
+            user.id,
             `${user.firstName} ${user.lastName}`,
           ),
       },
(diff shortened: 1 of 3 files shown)


esizer (approved):
Amazing! ⭐ 

Just incase anyone is interested I ran a test with default seeeders resulting in 50 candidates and got the following diff:

- **Request size**: 1.34MB -> 20KB = 98.5% decrease
- **Request time**: 5s -> 848ms = 83% decrease

Issue #8238, opened 2023-10-17
✨  Pool Candidate Keyword search


petertgiles:
## ✨ Feature

This issue is to complete the some initial backend work to support #7994.  It will only include one index to get started with.

## 🕵️ Details

We ran two spikes to investiagte some options:
#8090
#8089

As a team we decided that a standalone search service like Meilisearch was probably too much for our needs now but might be a good option in the future if we outgrow an integrated search tool.

We would like to start with using the [Postgresql Full Text Search](https://www.postgresql.org/docs/current/textsearch.html) feature to search against text indexes. There is lots of interest in using [Laravel Scout](https://laravel.com/docs/10.x/scout) as an interface to this tool to manage updates and searches. There's [a package](https://laravel-news.com/postgresql-full-text-search-for-laravel-scout) that might make this easy.

## Question and Answer

Q: I think it also needs to account for policy on search results, filter for can-view.
A: Using the existing search input should mean no changes to policy necessary

Q: we create an individual vectorized column for each field that needs to be indexed
A: Single column for entire user profile as a document

Q: Can we confirm the searchable fields from the list below ?

A:Users table
 email
first_name
last_name
telephone
current_city
Pool_Candiates
notes
All experiences table( work, personal, community, award )
  award - title, issued_by, details
  community - title, org, project, details
  education- areaofstudy, details, institution, thesis
  personal - description, details, title
  work - details, division, org, role
skill - justification

## 🙋‍♀️ Proposed Implementation

h/t @brindasasi 
>    Create a new column in the pool_candidates table to stored the vectorized data

>    Create a function
    ex.
    CREATE OR REPLACE FUNCTION update_profile_snapshot_vectorized() RETURNS TRIGGER AS $$
    BEGIN
    NEW.profile_snapshot_vectorized := to_tsvector('english', COALESCE(NEW.profile_snapshot::TEXT, ''));
    RETURN NEW;
    END;
    You can also combine multiple columns from different tables (ex. email id from users table)

>    Create a trigger for the above

>    Create GIN index for the column

>    To use the Laravel Scout driver , create the model like
    https://github.com/devNoiseConsulting/laravel-scout-postgres-tsvector

## ✅ Acceptance Criteria

- [x] Setup new column to hold index
Full profile = any open text fields the candidate has entered on their profile including screening questions, and notes written by request responders.  No numbers or enums.  Not snapshot.
- [x] Index keeps itself up to date
[...]

PR #8539, opened 2023-11-21
Implement keyword search backend

brindasasi:
🤖 Resolves {#8238}

## 👋 Introduction

This PR enables Full Text Search in Postgres and integrates it with Laravel Scout using the 
 package 
https://github.com/devNoiseConsulting/laravel-scout-postgres-tsvector

## 🕵️ Details

As discussed in the start of issue, we decided to use the existing api for pool candidate search 
GetPoolCandidatesPaginated , so we don't need to worry about any authorization issue. 

In this PR, I've used the `General Search` option which used to search on columns on users table like email, firstname, lastname and notes from pool candidates table. 
I've extended  this search to search on the entire profile but limited to the columns we have agreed on the issue. 

## 🧪 Testing

What Changed ? 
GetPoolCandidatesPagined Query has been modified to accept array of stings for `generalSearch`
Screenshot 2023-11-29 at 9 41 55 AM

How to Test ? 

1.. Grapiql 
2. via the search text box in the UI 
3. Separate each search term  by comma 
4. Space will be considered as a single phrase ( "UI Design") 

## 📸 Screenshot

Screenshot 2023-11-29 at 9 35 03 AM

What to Test? 

There is a lot you can test. 
There are lot of columns and many different tables. Try atleast 1 column for each table so we know its all covered. 

Also we are opening a new epic at this point. There is a lot to improve on ( ranking , highlighting). 
If it satisfies the core A/C that we decided on , it should get in and communicate any limitations with the user.

Screenshot 2023-11-29 at 9 56 51 AM

## Deployment Instructions
add env variable 
SCOUT_DRIVER=pgsql into Dev, UAT & Prod


vd1992:
> (I didn't add skill justification .. I don't know which column is that .. )

The justification is stored in the experience_skill pivot table under the details column.  It can be obtained by following the experience -> skills -> pivot path. For example, grabbing the justification for the first experience and its first skill in a user's experience collection. 

`workExperiences[0]->skills[0]->experience_skill->details`

composer.json
petertgiles, line 1: Did you mean to add a new composer file in apps/web?

api/database/migrations/2023_11_22_152329_update_users_table_add_searchable_column.php
petertgiles, line 1: I'm getting an error when I try to run setup from the UserSeederLocal.
image

api/app/Models/User.php
```diff
@@ -619,23 +644,27 @@ public static function scopeEquity(Builder $query, ?array $equity): Builder
...
-                $query->orWhere(function ($query) use ($search) {
-                    self::scopeTelephone($query, $search);
-                });
-            });
+        if ($searchTerms && is_array($searchTerms)) {
+            $searchResults = [];
+            foreach ($searchTerms as $searchTerm) {
+                // Use Scout's search method to perform the search and get results
+                $results = self::search($searchTerm)->usingPlainQuery()->get();
+                $searchResults = array_merge($searchResults, $results->all());
```
petertgiles, line 654: Running a search on each term serially and merging the results seems not ideal.  

The first problem is that it will result in an OR search, right?  From the original issue: 
> Example 1: Process engineering, Forestry would retrieve candidates that have both "process engineering" (entire term / phrase) and "forestry" on their profile.

The second problem is that it doesn't allow the database to parallelize the multiple searches and instead forces them to go one-by-one.

Is there any way to send all the terms to the database at once?

api/database/migrations/2023_11_22_152329_update_users_table_add_searchable_column.php
```diff
@@ -0,0 +1,42 @@
...
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        // add searchable column
+        // Blueprint doesn't work well with tsvector columns hence the raw statement
+        \DB::statement('ALTER TABLE users ADD COLUMN searchable TSVECTOR');
```
petertgiles, line 14: I think you need to `use Illuminate\Support\Facades\DB;` to use this.

api/database/migrations/2023_11_22_152329_update_users_table_add_searchable_column.php
```diff
@@ -0,0 +1,42 @@
...
+        \DB::statement('CREATE INDEX users_searchable_index ON users USING GIN(searchable)');
+
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        $table->dropColumn('searchable');
```
petertgiles, line 39: I think you need a `Schema::connection()->table()` closure to use this.

petertgiles (changes requested):
I'm excited to try this out - it looks very cool! :star_struck:  Unfortunately it doesn't build for me.  I did a read-through of the code and make some suggestions though.

api/database/migrations/2023_11_22_152329_update_users_table_add_searchable_column.php
brindasasi: This should have been resolved in [153a408](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8539/commits/153a40841bf150762ee554985c5cf734e0e6d016)

api/database/migrations/2023_11_22_152329_update_users_table_add_searchable_column.php
brindasasi: fixed in [69fafa2](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8539/commits/69fafa2538f97bb31015046bbceabbbd75ffb5b9)

api/app/Models/User.php
```diff
@@ -96,8 +98,90 @@ class User extends Model implements Authenticatable, LaratrustUser
...
+        $this->loadMissing([
+            'poolCandidates',
+            'workExperiences',
+            'educationExperiences',
+            'personalExperiences',
+            'communityExperiences',
+            'awardExperiences',
+        ]);
+
+        $notes = $this->poolCandidates->pluck('notes')->toArray();
```
petertgiles, line 124: Hmm, this might be a bit awkward.  If you're indexing all the pool candidate notes and you use this index to search pool canidates you might get hits for the notes on pool candidates that you're not able to see.

petertgiles (changes requested):
Very cool!  I can see it's working now.  :partying_face: I flagged a few more things to check on next...

api/database/migrations/2023_11_22_152329_update_users_table_add_searchable_column.php
```diff
@@ -0,0 +1,45 @@
...
+            CONCAT_WS(\' \', ARRAY(SELECT CONCAT_WS(\' \', role, organization, division, details) FROM work_experiences WHERE work_experiences.user_id = users.id)) ||
+            CONCAT_WS(\' \', ARRAY(SELECT CONCAT_WS(\' \', thesis_title, details) FROM education_experiences WHERE education_experiences.user_id = users.id)) ||
+            CONCAT_WS(\' \', ARRAY(SELECT CONCAT_WS(\' \', title, description, details) FROM personal_experiences WHERE personal_experiences.user_id = users.id)) ||
+            CONCAT_WS(\' \', ARRAY(SELECT CONCAT_WS(\' \', title, issued_by, details) FROM award_experiences WHERE award_experiences.user_id = users.id)) ||
+            CONCAT_WS(\' \', ARRAY(SELECT CONCAT_WS(\' \', title, details) FROM community_experiences WHERE community_experiences.user_id = users.id))
+            )'
+        );
+
+        // Create a GIN index on the tsvector column for efficient full-text search
+        \DB::statement('CREATE INDEX users_searchable_index ON users USING GIN(searchable)');
```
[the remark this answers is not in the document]
petertgiles: Could you check how many users are in prod and seed that many to see how big it gets?

api/database/seeders/UserSeederLocal.php
```diff
@@ -104,5 +105,9 @@ public function run()
                 'email' => 'noroles@test.com',
                 'sub' => 'noroles@test.com',
             ]);
+
+        // after seeding re-index the users in the search index
+        Artisan::call('scout:import', ['model' => User::class]);
```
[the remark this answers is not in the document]
brindasasi: when you update a model it will fire .. Initial deploy you need to import the existing records into into scout so that it can create the index.

api/app/Models/User.php
```diff
@@ -619,18 +704,17 @@ public static function scopeEquity(Builder $query, ?array $equity): Builder
...
-            $query->where(function ($query) use ($search) {
-                self::scopeName($query, $search);
-                $query->orWhere(function ($query) use ($search) {
-                    self::scopeEmail($query, $search);
-                });
-                $query->orWhere(function ($query) use ($search) {
-                    self::scopeTelephone($query, $search);
-                });
-            });
+        if ($searchTerms && is_array($searchTerms)) {
```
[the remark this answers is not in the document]
brindasasi: yeah I agree with the white space. Will trim it.

api/graphql/schema.graphql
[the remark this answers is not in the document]
brindasasi: oh this is a great review! Thank you.. will add some more tests and take a look at those suggestions. 😊

api/app/Models/Experience.php
```diff
@@ -147,4 +147,18 @@ public function disconnectSkills($skillIds)
         // If this experience instance continues to be used, ensure the in-memory instance is updated.
         $this->refresh();
     }
+
+    protected static function boot()
+    {
+        parent::boot();
+
+        // Eloquent event for saving
+        static::saving(function ($experience) {
```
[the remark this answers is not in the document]
brindasasi: Good point on delete. Its a bug. It didn't trigger when deleted. fixed now. 
Poolcandidate model doesn't need anything. When a user submits application he immediately shows up on candidate search. 
Added Created event for user creation. Might not need re-indexing on the seeder. Will test it out

api/database/migrations/2023_11_22_152329_update_users_table_add_searchable_column.php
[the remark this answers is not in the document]
brindasasi: this one is tricky. We have almost 16k users in prod. I can't seed that many in my local.
Discussed with @tristan-orourke. We need a separate work out to do this in Dev. Faker isn't prod build rt now. Even if create a sep build including with that , we don't want to run user seeder which will wipe out all existing users. But may be that could be fine.. This should be a sep discussion and issue to verify this. 
But I think very much needed

api/graphql/schema.graphql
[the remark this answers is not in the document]
brindasasi: I think Ive added all of them

api/app/Models/User.php
```diff
@@ -96,8 +98,90 @@ class User extends Model implements Authenticatable, LaratrustUser
...
+        $community_details = count($community_details) > 0 ? $community_details : [];
+        $award_titles = $this->awardExperiences->pluck('title')->toArray();
+        $award_titles = count($award_titles) > 0 ? $award_titles : [];
+        $award_details = $this->awardExperiences->pluck('details')->toArray();
+        $award_details = count($award_details) > 0 ? $award_details : [];
+        $issued_by = $this->awardExperiences->pluck('issued_by')->toArray();
+        $issued_by = count($issued_by) > 0 ? $issued_by : [];
+
+        return [
+            'id' => $this->id,
```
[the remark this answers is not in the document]
brindasasi: not really .. removed it

api/database/seeders/UserSeederLocal.php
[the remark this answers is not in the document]
brindasasi: aannd removed it after adding the re-index after saving the model. Seems to work without that now.

api/app/Models/Experience.php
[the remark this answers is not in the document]
brindasasi: yeah this worked out well.. thanks!

api/app/Models/User.php
```diff
@@ -618,18 +668,17 @@ public static function scopeEquity(Builder $query, ?array $equity): Builder
...
-            });
+        if ($searchTerms && is_array($searchTerms)) {
+            $combinedSearchTerm = implode('&', array_map('trim', $searchTerms));
+            $results = self::search($combinedSearchTerm)->usingWebSearchQuery()->get();
+
+            // Extract unique IDs from the search results
+            $uniqueIds = $results->pluck('id')->unique()->toArray();
+
+            // Use Eloquent builder to filter results based on unique IDs
+            $query->whereIn('id', $uniqueIds);
```
petertgiles, line 681: Hmm, this isn't really covered by the issue, I think, but we should probably consider limiting the list of search results that come back.  No point in pulling up 1,000 hits if the user is going to look at the top, say, 20 hits.

api/graphql/schema.graphql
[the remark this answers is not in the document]
petertgiles: Wow, these tests are great!  I love all of them.

`testUserSearchBySoftDeleted` is a good test but not quite what I was thinking of.  What I'd like to see, though, is if a model is hard-deleted that it leaves the index.  That's probably impossible with a graphQL test since it will always turn the IDs into models before returning them.  You'd have to call the `search` function on the model directly from your test.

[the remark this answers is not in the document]
petertgiles: Oh, thought of one more: multiple terms.  Could you make three models, one with term A, one with term B, one with both A and B.  Then assert that only the third one comes up when searching both terms.

api/app/Models/User.php
petertgiles: Oracle has a limit of 1000 terms that can be passed to an `in` clause.  I don't think Postgres has that limit but it would still probably burn a huge amount of ram to do that here.

brindasasi: I changed it to pick up only top 100 to get this going.
I believe I did this way because of the original discussion that we don't want to create a new end point, keep the existing and just change the GeneralSearch to accomodate this keyword search.
When we search inside the model there is really no field permission, role policies check going on there. All of the records will come and only when the query builder runs it make sure it returns only authorized records and skip any unauthorized.Also it is taking care of user's set pagination , so I don't have to deal with that separately.
When search have its own end point this can be optimized further. But at this point, I wouldn't worry about that.

api/app/Models/User.php
```diff
@@ -95,8 +97,52 @@ class User extends Model implements Authenticatable, LaratrustUser
...
+        foreach ($attributesToPluck as $relationship => $attributes) {
+            foreach ($attributes as $attribute) {
+                $values = $this->$relationship->pluck($attribute)->reject(function ($value) {
+                    return is_null($value) || $value === '';
+                })->toArray();
+
+                // Merge non-null and non-empty values
+                $entireIndex = $entireIndex->merge($values);
+            }
+        }
```
brindasasi, line 140: I reverted the changes         
```
return collect([$this->email, $this->first_name, $this->last_name, $this->telephone, $this->current_province, $this->current_city])
            ->merge($this->poolCandidates()->pluck('notes'))
            ->merge($this->workExperiences()->pluck('role'))
            ->merge($this->workExperiences()->pluck('organization'))
            ->merge($this->workExperiences()->pluck('division'))
            ->merge($this->workExperiences()->pluck('details'))
            ->merge($this->educationExperiences()->pluck('thesis_title'))
            ->merge($this->educationExperiences()->pluck('institution'))
            ->merge($this->educationExperiences()->pluck('details'))
            ->merge($this->educationExperiences()->pluck('area_of_study'))
            ->merge($this->personalExperiences()->pluck('title'))
            ->merge($this->personalExperiences()->pluck('description'))
            ->merge($this->personalExperiences()->pluck('details'))
            ->merge($this->communityExperiences()->pluck('title'))
            ->merge($this->communityExperiences()->pluck('organization'))
            ->merge($this->communityExperiences()->pluck('project'))
            ->merge($this->communityExperiences()->pluck('details'))
            ->merge($this->awardExperiences()->pluck('title'))
            ->merge($this->awardExperiences()->pluck('details'))
            ->merge($this->awardExperiences()->pluck('issued_by'))
            ->reject(function ($value) {
                return is_null($value) || $value === '';
            })->toArray();

```
because 
1. It broke the bunch of tests in AuthProvider as there are some nulls 
 Reject after all collection merge doesn't affect the nulls in the individual collection. It affects only the merged array.
Ex. If there are null or empty values in the notes, they won't be removed by the reject function in this context. 
Thats why it broke the test. 
2. Actually It is better practice to remove nulls before merging in, instead of removing it after the fact.

api/graphql/schema.graphql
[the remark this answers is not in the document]
brindasasi: testUserSearchByNameAndWorkExperienceTitle() should resolve this

api/app/Models/User.php
```diff
@@ -618,18 +668,18 @@ public static function scopeEquity(Builder $query, ?array $equity): Builder
...
-                $query->orWhere(function ($query) use ($search) {
-                    self::scopeEmail($query, $search);
-                });
-                $query->orWhere(function ($query) use ($search) {
-                    self::scopeTelephone($query, $search);
-                });
-            });
+        if ($searchTerms && is_array($searchTerms)) {
+            $combinedSearchTerm = implode('&', array_map('trim', $searchTerms));
+            $resultIds = self::search($combinedSearchTerm)->usingWebSearchQuery()
```
tristan-orourke, line 675: Does the Scout search function accept something like `take(100)` as well?

api/config/scout.php
```diff
@@ -0,0 +1,117 @@
...
+    | Database Transactions
+    |--------------------------------------------------------------------------
+    |
+    | This configuration option determines if your data will only be synced
+    | with your search indexes after every open database transaction has
+    | been committed, thus preventing any discarded data from syncing.
+    |
+    */
+
+    'after_commit' => false,
```
tristan-orourke, line 57: Any reason not to turn this on?

api/config/scout.php
```diff
@@ -0,0 +1,117 @@
...
+    |
+    | This option allows you to control whether to notify the search engine
+    | of the user performing the search. This is sometimes useful if the
+    | engine supports any analytics based on this application's users.
+    |
+    | Supported engines: "algolia"
+    |
+    */
+
+    'identify' => env('SCOUT_IDENTIFY', false),
```
tristan-orourke, line 101: Interesting! I don't think we should turn this on now, but it could allow some interesting auditing/analytics later.

api/database/migrations/2023_11_22_152329_update_users_table_add_searchable_column.php
```diff
@@ -0,0 +1,46 @@
...
+        DB::statement('CREATE INDEX users_searchable_index ON users USING GIN(searchable)');
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->dropColumn('searchable');
```
tristan-orourke, line 43: Does the index need to dropped as well?

api/app/Models/User.php
tristan-orourke: I just realized that [this comment](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8539#discussion_r1434665504) addresses this.

api/config/scout.php
brindasasi: Turning this on will sync data only when successful transaction and false will do every time.
I guess turning this on would be the right way. Will do it.

tristan-orourke:
@petertgiles I appreciate you being so quick with https://github.com/GCTC-NTGC/gc-digital-talent/pull/8539/commits/5c5c40830585d9aacf42dc0c37c8b390ff17ebde but I take back my request for that 😖 

Brinda's comment [here](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8539#discussion_r1434665504) actually explains that there was a very good reason she had it the way she did.

api/app/Models/User.php
petertgiles: Refactored in https://github.com/GCTC-NTGC/gc-digital-talent/pull/8539/commits/770be873d1d04173b29709365c728a3523444711

petertgiles (approved):
Official stamp of approval!

Issue #4227, opened 2022-10-06
Research Cypress for e2e testing with feature flags ON and OFF 


brindasasi:
### Problem

We need a testing mechanism to test ON and OFF side of feature flags as we constantly introduce bugs related to that. 
We need a automated way of catching those issues earlier. 

### :question: Spike Question
- How can we run e2e testing with feature flags on and off?

Timebox: 2 days

patcon:
Related slack convo: https://talent-cloud.slack.com/archives/C0259G6KXJ6/p1665007016304059?thread_ts=1665006944.546679&cid=C0259G6KXJ6

tristan-orourke:
https://github.com/GCTC-NTGC/gc-digital-talent/pull/4209#issuecomment-1271839207

patcon:
Example of how not running e2e tests on both states of feature flag can lead to issues with deprecated components breaking before we've totally removed the feature toggle that relies on them:

https://app.slack.com/client/T5ZFRSQ3V/threads/thread/C0259G6KXJ6-1669924724.249809?cdn_fallback=1 🔒

mnigh:
I think this is possible, taking a similar approach to what @tristan-orourke mentioned in his [comment](https://github.com/GCTC-NTGC/gc-digital-talent/pull/4209#issuecomment-1271839207) for when can use a feature introduced in Cypress 10 (which we can now). We can map our `.env` file in Cypress config and then within individual tests, set the environment variable as needed (REF: https://docs.cypress.io/guides/guides/environment-variables#Single-test-configuration).

## Screenshot
![Screen Shot 2023-11-08 at 10 51 24](https://github.com/GCTC-NTGC/gc-digital-talent/assets/3046459/9ece0533-db12-456e-8421-9a01ce7e681f)

The question is though for @GCTC-NTGC/developers: is there still an appetite to do this or should this issue be closed?

petertgiles:
> The question is though for @GCTC-NTGC/developers: is there still an appetite to do this or should this issue be closed?

I still think it's valuable.

vd1992:
[...]

PR #8546, opened 2023-11-21
[Feature] Update pool publishing validation for RoD

### LINKED ISSUE #7920: ✨ Update pool publishing validation

vd1992:
🤖 Resolves #7920

## 👋 Introduction

Adds assessment plan validation to the pool publishing validation file. 

## 🕵️ Details

Added validation, added testing, adjusted a factory function, and added frontend messages. Though, the frontend messages are impossible to see without breaking front validation. 

## 🧪 Testing

1. Have a pool complete except for assessment plan, ensure you cannot publish
2. Toggle off feature flag, can publish it

## 📸 Screenshot

Break the frontend and/or database manipulation to see
"skill" in the second message was pluralized after

image


api/.env.example
```diff
@@ -97,10 +97,9 @@ LARATRUST_ENABLE_CACHE=true
 FEATURE_SKILL_LIBRARY=true
 FEATURE_STATUS_NOTIFICATIONS=true
 FEATURE_DIRECTIVE_FORMS=true
-FEATURE_RECORD_OF_DECISION=true
+FEATURE_RECORD_OF_DECISION=false
```
vd1992, line 100: Couldn't quite get this resolved with Cypress config, might be due to there being two env files, web and api, that complicates things. The api might not be controllable from Cypress?

esizer: We have [a spike](https://github.com/GCTC-NTGC/gc-digital-talent/issues/4227) for this so I wouldn't worry about it.

packages/i18n/src/messages/apiMessages.ts
```diff
@@ -232,6 +232,21 @@ export const apiMessages: { [key: string]: MessageDescriptor } = defineMessages(
...
+      id: "rgxNWX",
+    },
+    PoolSkillsWithoutAssessments: {
+      defaultMessage:
+        "Please ensure the advertisement's skills are connected to one or more assessments",
+      description:
+        "Error message that the pool advertisement skills are lacking an assessment",
+      id: "ZWEzaw",
+    },
+
```
esizer, line 249: Tiny nitpick that you probably were not aware of but, the copy editors have recommended we stop using "Please" because it implies it is a request. Though, it is not a request, it is something the user _must_ do. Could we rephrase these to be statements rather than requests?

- Each assessment must include one or more skill for evaluation
- Each skill must be included in an assessment

api/.env.example
vd1992: Yeah, I tried a suggestion from that issue but couldn't get it to work

packages/i18n/src/messages/apiMessages.ts
vd1992: Politeness, dial down in https://github.com/GCTC-NTGC/gc-digital-talent/pull/8546/commits/b22057d0b835a0e4494380845dfd645b6bc1ea98

esizer: haha, that is an apt description 🤣

petertgiles (approved):
Nice, this works very well!  I love how you've split the logic into two rules and added tests.  :1st_place_medal: 

I actually found a bug in the frontend while testing the validation rules.  :grimacing:   I know it's technically out of scope for this PR but if you wouldn't mind, I'd be much obliged if you added a missing field to the graphql query.
image

vd1992:
> Nice, this works very well! I love how you've split the logic into two rules and added tests. 🥇
> 
> I actually found a bug in the frontend while testing the validation rules. 😬 I know it's technically out of scope for this PR but if you wouldn't mind, I'd be much obliged if you added a missing field to the graphql query. image

A hefty request https://github.com/GCTC-NTGC/gc-digital-talent/pull/8546/commits/d1f340efb8b8eb8406e7b80529d6dbbcf54fab9e

vd1992:
@petertgiles I will assume the rocket emoji means, approved x2

PR #8600, opened 2023-11-24
[Fix] Stabilize `urlSync` for `ResponsiveTable`

### LINKED ISSUE #8597: 🐛 `ResponsiveTable` infinite re-render on default filters
## 🐛 Bug

When default filters are applied to the `ResponsiveTable`, it results in an infinite re-render.

## 🦋 Expected Behaviour

Table can have default filters without continually re-rendering.

## 🕵️ Details

This is almost certainly an issue with updating the URL since the specific message is:

> Too many calls to Location or History

See @vd1992 [comment](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8592#issuecomment-1824977564) for more details.

## 📋 Steps to Reproduce

1. Navigate to `/admin/pools`
2. Click on "View candidates" on any row
3. Observe the table continually re-renders causing jittery scrolling and print/download broken

## ✅ Acceptance Criteria

- [ ] Default filters can be passed in still
- [ ] URL state is not updated on default filters

esizer:
🤖 Resolves #8597 

## 👋 Introduction

Fixes a bug that caused infinite re-renders when using `urlSync` on pages that need to read from url params. This also prevents the URL from updating with default filters.

## 🕵️ Details

It seems that the `setSearchParams` updater does not work like the default `setState` update in react. Using the updater, it causes a navigation regardless if there are any new params.

We fixed this by checking to see if the new params are different than our current params before setting them.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to `/admin/pools`
3. Click on "View candidates" for any row
4. Confirm the page is not infinitely re-rendering
5. Confirm the URL state sync still works

### DIFF
--- a/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
+++ b/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
@@ -542,6 +542,7 @@ const PoolCandidatesTable = ({
         initialState: defaultState.sortState,
       }}
       filter={{
+        initialState: initialFilterInput,
         state: filterRef.current,
         component: (
           <PoolCandidateTableFilterDialog
--- a/apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
+++ b/apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
@@ -241,7 +241,11 @@ const ResponsiveTable = <TData extends object, TFilters = object>({
         }
       }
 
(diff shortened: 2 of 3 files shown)


vd1992:
Impressive, it appears to have solved all the jank I was experiencing, the URL by default also looks nicer too. 
Definitely will want a second person to review this, I will keep playing around with things, but want to be extra careful for this page.

petertgiles (approved):
Code looks good, no problems with 500 pool candidates.

PR #8605, opened 2023-11-27
[Debt] Update `Switch` component

### LINKED ISSUE #8245: ✨ Update the switch component
image

## ✨ Summary

[Visit the Figma file](https://www.figma.com/file/guHeIIh8dqFVCks310Wv0G/Style-library?type=design&node-id=818-9584&mode=design) • Contact @substrae with design questions

As a part of upcoming assessment record designs, we'd like to implement a custom toggle component that leverages a single `checkbox` input. Toggles will have an optional visual label (but persistent AT-friendly label), an optional icon, and be available in all 5 brand colors, as well as our stoplight colors.

Toggles will only be used when an on/off format makes contextual sense (e.g. we wouldn't use one to confirm a choice).

```[tasklist]
### ✅ Requirements
- [ ] Implement a custom `checkbox` input that uses `:checked` to change the toggle's state
- [ ] Implement color variants as shown below
- [ ] Ensure that the input's label is visually optional while retaining AT-specific labelling at all times
- [ ] Allow for an optional icon, as well as for that icon to optionally change based on the toggled state
- [ ] Ensure the switch meets the criteria outlined in the [WAI switch pattern documentation](https://www.w3.org/WAI/ARIA/apg/patterns/switch/)
``` 

## 🔍 Details

<details>
  <summary>📸 Color variants</summary>
  
</details>

<details>
  <summary>📸 Focus state</summary>
  
</details>

<details>
  <summary>📸 Disabled state</summary>
  
</details>

<details>
  <summary>📸 Without labels</summary>
  
</details> 

<details>
  <summary>📸 With icons</summary>
  
</details>

esizer:
🤖 Resolves #8245 

## 👋 Introduction

Updates the `Switch` component styles/functionality.

## 🕵️ Details

This does a few important things outside of updating the styles.

- Moves to the `@gc-digital-talent/forms` package
- Modifies it to work in `react-hook-form`
- Updates the API to only need one component

### API

As always with new/changing APIs, let me know if anything doesn't make sense or suggest improvements! 🙏 

## ❓ Question

Since this component does not need to be inside a form, should be consider abstracting it out and make a `ui` one that the `forms` one is based on? I thought of doing this but backed away and am still uncertain about it 😕 

## 🧪 Testing

> [!WARNING]
> There is an a11y warning in storybook but it is a false negative! The button names are pragmatically determined. @substrae and I looked into this and think it may have something to do with browser support for `role=switch` and hope to get it sorted. But, for now, it is working as expected so you can ignore that specific error.

1. Update deps `npm i`
2. Run storybook `npm run storybook:design-system`
3. Navigate to the "Switch" stories
4. Confirm they match the mockup 

## 📸 Screenshot

Screenshot 2023-11-27 122413


petertgiles:
I don't think this is working for me.  :grimacing: Is it missing a commit or something?
image

esizer:
> I don't think this is working for me. 😬 Is it missing a commit or something?

Odd 😕  [Chromatic](https://61e099a1bb1465003a73d3ce-zgvlqripvu.chromatic.com/?path=/story/form-switch--default) seems to think it is working 😅 

Screenshot 2023-11-28 134533

petertgiles:
> Since this component does not need to be inside a form, should be consider abstracting it out and make a ui one that the forms one is based on? I thought of doing this but backed away and am still uncertain about it 😕

petertgiles:
> Since this component does not need to be inside a form, should be consider abstracting it out and make a ui one that the forms one is based on? I thought of doing this but backed away and am still uncertain about it 😕

I guess the question is, if I want to plop it on a page and use it to drive (for example) a useState boolean, can I do that easily?

esizer:
> I guess the question is, if I want to plop it on a page and use it to drive (for example) a useState boolean, can I do that easily?

Right now, no. It must be used within a `FormProvider`. So my thought would be to move all the primitive styling back into `packages/ui` and then import those into `packages/forms` and use those inside the `Controller` render prop 🤔 Does that make sense?

Then `Switch` just takes a `checked` and `onCheckedChange` prop. So it would look something like this:

```tsx
// Used anywhere!
import { Switch } from "@gc-digital-talent/ui";
  
const Component = () => {
  const [checked, setChecked] = React.useState<boolean>(false);
  
  return <Switch checked={checked} onCheckedChange={setChecked} />
} 


// ./packages/forms/SwitchInput.tsx
import { Switch } from "@gc-digital-talent/ui";

const SwtichInput = () => {
 ....
   return (
     <Controller
        render={() => <Switch checked={value} onCheckedChange={(newChecked) => setValue(name, newChecked)} />}
        ....
      />
   )
 }
 ```
 
 I was just worried about the confusion between `Swtich` and `SwitchInput`. Or do you think my concern is unwarranted?

**Comment by petertgiles** [comment]
> I was just worried about the confusion between `Swtich` and `SwitchInput`. Or do you think my concern is unwarranted?

I like this solution.  I can image the designers wanting to use this component outside of traditional forms more frequently.  With some good inline documentation the confusion could be eased.

Although, to argue, we use the dropdowns a lot and we just use a form provider with a watch... so maybe not necessary?  I could go either way to be honest.

**Review by petertgiles — APPROVED** [review]
I like it!  If you decide not to split it I think this is great!  If you do want to split it, just request another review.

**Comment by petertgiles** [comment]
Hmm, do we need a disabled state for these?  I'm thinking about if we want to use these outside of forms, for a pending state, too.

**Comment by substrae** [comment]
@petertgiles Yeah, I can absolutely add one to the ticket. In terms of the pending state, can you give me a more detailed example of the situation you're imagining?

**Comment by petertgiles** [comment]
> . In terms of the pending state, can you give me a more detailed example of the situation you're imagining?

It's part of the "using it outside a form" scenario.  When you click a submit button it gets disabled and the copy changes to "submitting" until the action resolves.  If we want to have an action occur immediately on switch we need a way to indicate that we're waiting for resolution.  Maybe it's a simple as just disabling the switch during the "submit"?

**Comment by substrae** [comment]
@petertgiles I want to account for this situation, but I can't think of an instance where we would want this. Maybe if they're acting as instant filters? We could change/show the icon as the loading arrows?

**Comment by petertgiles** [comment]
> I want to account for this situation, but I can't think of an instance where we would want this.

Fair enough.  No point in over-building for something we _might_ need.  Having a disabled state gets us 90% of the way there.

**Comment by esizer** [comment]
Okay, so @substrae had some minor tweaks to the styles as well as adding in the requested `disabled` portion. I figured, while I was in there, I may as well also [extract it out so it doesn't need to be a form inpu](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8605/commits/6bf59d3110f9c8633a4fa8346d38557b1dd04dcb)t as well! The following changes were made

 - Icon size: [reduce icon size](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8605/commits/103678616bd9a69e2a44e5f90cf77eaaf6b1aee4)
 - Colour styles: 
   - [lighten specific colours](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8605/commits/b969378ae1057681953bb8fc4ee00d13cbb894bb)
   - [fix unchecked background color](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8605/commits/2899e34e3cacb649e53c162aaea0451ea8658c4a)
 - Disabled: 
   - [add disabled styles](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8605/commits/29c14712ff16981af98dddc31d96141549399f19)
   - [adjust disabled styles](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8605/commits/12e24def0e8af4554860fb8ffd847fcf898d26d4)

File: packages/ui/src/components/Switch/styles.ts [hunk 1]
```diff
@@ -0,0 +1,66 @@
...
+  "black" | "white" | "blackFixed" | "whiteFixed"
+>;
+
+const getStyles = (
+  color: SwitchColor,
+  disabled?: boolean,
+): Record<string, string> => {
+  let colorStyle: Record<string, string> = {};
+
+  if (disabled) {
```
petertgiles, line 14: This doesn't seem to be working for me - they're still all coloured.  This time Chromatic agrees with me.  :laughing: 
image

petertgiles (changes requested):
Love the refactor!  Just need to fix up the disabled mode.

packages/ui/src/components/Switch/styles.ts
esizer: Oops, I guess actually passing the prop would help 😅 

Fixed in [pass down disabled prop](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8605/commits/2e2bc134f0808352a7170e66ef8c9fbdb540d2eb)

petertgiles: Oh yeah, that helps.  :laughing:   Looks great!

petertgiles (approved):

PR #8615, opened 2023-11-28
[Feature] Have more admin tables utilize `state.from` 

### LINKED ISSUE #8345: ✨ Allow more admin pages to make use of `state.from` to return to table as it was
## ✨ Feature

Basically repeat #8339 for as many admin pages as possible, it would be useful for many of them.

## 🕵️ Details

## 🎨 Design File

### 🧑‍🎨 Designer

## 📸 Screenshot

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Admin forms which return user to a table index page when submitted, will return them to the table in the state they left

vd1992:
🤖 Resolves #8345 

## 👋 Introduction

Implements the state solution of #8342 in more places. 

## 🕵️ Details

For the most part just finding places to copy paste. 

## 🧪 Testing

1. Go to effected tables, adjust the filters to change the URL then test the submit and return
2. Effected tables include classifications, departments, teams, skills, skill families, users, and requests
3. Submit and return, as well as go back buttons

## 📸 ~Screenshot~ Recordings

https://recordit.co/Orl9vqFytJ

https://recordit.co/CQf17DDn6u

### DIFF
--- a/apps/web/src/pages/Departments/components/DepartmentTable.tsx
+++ b/apps/web/src/pages/Departments/components/DepartmentTable.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import { ColumnDef, createColumnHelper } from "@tanstack/react-table";
 import { useIntl } from "react-intl";
+import { useLocation } from "react-router-dom";
 
 import { notEmpty } from "@gc-digital-talent/helpers";
 import { getLocalizedName } from "@gc-digital-talent/i18n";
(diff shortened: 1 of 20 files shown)


apps/web/src/pages/Departments/components/DepartmentTable.tsx
```diff
@@ -29,7 +30,7 @@ export const DepartmentTable = ({
   const columns = [
     columnHelper.accessor("departmentNumber", {
       id: "departmentNumber",
-      filterFn: "equals",
+      filterFn: "weakEquals",
```
vd1992, line 33: Department number searching wasn't working and this seemed to fix it. Seems it was a case of comparing a number to a stringified number, so needs to not match type

esizer (approved):
LGTM! 👍 

---

Out of scope but something that I noticed while reviewing is that our column visibility is not being stored in search params 😅 , I'll create a ticket for that 👍

PR #8617, opened 2023-11-28
[Debt] Make filter dialog generic

### LINKED ISSUE #3973: Make the filter dialog fully generic
Right now the "fancy filter" filter dialog is closely coupled to the User model and it's properties.  Let's create a new filter component based on the existing one that is fully generic.  Ideally it could be used to filter tables of users, pool candidates, requests, or any other model.

Acceptance Criteria:
- [ ] New, fully generic, filter dialog component
- [ ] Storybook story
- [ ] UserTableFilterDialog re-implemented on top of the new component
- [ ] PoolCandidateTableFilterDialog re-implemented on top of the new component

### LINKED ISSUE #4995: ⌨ a11y - react-select swallows keyboard input

The `react-select` library we are using to override the default `select[multiple=true]` input stops keyboard input from propogating outside of the component. This is particularly an issue when used in our `<Dialog />` component since we listen for `esc` to close the dialog. This forces keyboard users to tab through **_all_** inputs to close the dialog.

**WCAG 2.1 Criteria:**

> **Note**
> There is no specific criteria surrounding this, the closest would be [2.1.2 No Keyboard Trap](https://www.w3.org/TR/2008/REC-WCAG20-20081211/#keyboard-operation-trapping). However, while we never truly trap the user, it is an overall painful experience for users.

### How to reproduce
2. Navigate to `/admin/users`
3. Open the Filters dialog
4. Press `esc`
5. Notice the dialog does not close
6. Tab through all inputs so that one is not focus
7. Press `esc`
8. Notice the dialog now closes as expected

## ✔ Expected behaviour (fix)

Pressing keys in a `react-select` element must bubble up so we can use it to control other elements.

### Acceptance Criteria
A set of assumptions which, when tested, verify that the accessibility requirement was met.
- [ ] `react-select` keyboard input bubbles
- [ ] Pressing `esc` while focusing a `react-select` field in a dialog closes the dialog

## Additional information (optional)

1. This could be an issue in other areas such as the expected behaviour of `enter` submitting forms.
2. This traps users in the dialog while focusing `react-select` forcing them to tab through 10 items before being able to close the dialog

esizer:
🤖 Resolves #3973, Resolves #4995

## 👋 Introduction

Updates our filter dialogs on tables to be more generic but not entirely. 

## 🕵️ Details

The dialogs are fairly specific to the data they are filtering so it was difficult to make the full generifc. However, this gets us a lot closer to that. It also does much more than just that. 

## 🕷️ Scope Creep

Sadly, this ticket got a little out of hand but while working on making them generic, I noticed that the way we implemented the dialogs was much different than how we write other components. I tried to bring these more inline with how we write our components currently and bring them inline with our code style.

As well, we were using multi select for every input regardless of the data. So, this also changes the types of inputs we are using which will hopefully make them easier to use.  As part of updating the inputs, I also replaced the multiselect based off of `react-select` so we can close #4995 as well.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to the following pages and confirm the dialogs function
    - `/admin/pool-candidates`
    - `/admin/users`
    - `/admin/talent-requests`

### DIFF
--- a/apps/web/src/lang/fr.json
+++ b/apps/web/src/lang/fr.json
@@ -5331,6 +5331,10 @@
     "defaultMessage": "En apprendre davantage sur les tests de langue",
     "description": "Link text for language testing information"
   },
+  "Swq+OD": {
+    "defaultMessage": "Toute durée",
+    "description": "Option label for allowing any employment duration"
+  },
   "SxP9zE": {
     "defaultMessage": "Langue préférée de l'examen écrit",
     "description": "Label displayed on the user form preferred written exam language field."
@@ -9183,10 +9187,6 @@
     "defaultMessage": "Pour commencer",
     "description": "Main heading in create account page."
   },
-  "o0XhkM": {
-    "defaultMessage": "Possède un diplôme",
-    "description": "Label for the has diploma field"
-  },
   "o0Yt8Q": {
     "defaultMessage": "Établissement",
     "description": "Label displayed on education form for institution input"
@@ -9582,6 +9582,10 @@
   "qn77WI": {
     "defaultMessage": "Équipe, groupe ou division",
     "description": "Label displayed on Work Experience form for team/group/division input"
+  },
+  "qp68Mh": {
+    "defaultMessage": "Toute langue",
+    "description": "Option label for allowing any language"
   },
   "qrDCTV": {
     "defaultMessage": "Catégorie",


apps/web/src/components/FilterDialog/FilterDialog.stories.tsx
```diff
@@ -0,0 +1,89 @@
+import React from "react";
+import { StoryFn } from "@storybook/react";
+import { faker } from "@faker-js/faker";
+import { action } from "@storybook/addon-actions";
+
+// import { Combobox } from "@gc-digital-talent/forms";
```
mnigh, line 6: Is this needed?

esizer: Very much so! JK, removed in [remove commented import](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8617/commits/7ab363cf6fc06b55ca998bc7f02be5d3412d557a)

petertgiles (changes requested):
Wow, this turned out really nice!  I especially like the filter count on the trigger button.  I did flag a few things to check on before you merge.

petertgiles:
If you're closing #4995 should there be some changes to package.json files?

apps/web/src/components/FilterDialog/FilterDialog.tsx
```diff
@@ -0,0 +1,135 @@
...
+            defaultMessage: "Select filters",
+            id: "P9SZBZ",
+            description: "Candidate search filter dialog: title",
+          })}
+        </Dialog.Header>
+        <Dialog.Body>
+          <FormProvider {...methods}>
+            <form onSubmit={methods.handleSubmit(handleSubmit)}>
+              {children}
+              <Dialog.Footer>
```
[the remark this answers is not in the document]
esizer: That is a good question. Technically there is one in the top right so I'm not sure it is a huge deal right now but maybe we can bring it up and see if it makes sense to include one at the bottom along the other actions as well.

apps/web/src/components/PoolCandidatesTable/PoolCandidateFilterDialog.tsx
```diff
@@ -0,0 +1,313 @@
...
+            value="true"
+            label={intl.formatMessage({
+              defaultMessage: "Government employee",
+              id: "bOA3EH",
+              description: "Label for the government employee field",
+            })}
+          />
+        </div>
+        <div data-h2-grid-column="l-tablet(span 2)">
+          <Select
```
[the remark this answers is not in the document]
esizer: Oh yeah, that seems like something important 😅 Would adding an empty option like "Any language" work?

apps/web/src/pages/Users/IndexUserPage/components/UserFilterDialog.tsx
```diff
@@ -0,0 +1,223 @@
...
+} from "~/components/FilterDialog/FilterDialog";
+import adminMessages from "~/messages/adminMessages";
+import { getFullPoolTitleLabel } from "~/utils/poolUtils";
+
+export type FormValues = {
+  pools: string[];
+  languageAbility: string;
+  operationalRequirement: string[];
+  workRegion: string[];
+  employmentDuration: string;
```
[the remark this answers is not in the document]
esizer: So, this was another point of confusion for me. We had some conditional logic to only pass this if they select term. I have no idea why so I left it untouched. I'm not sure why we we would offer an option that does nothing so I updated it to work. 

If we decide that we actually want it to nothing, we can back it out 😆 

[allow to filter by indeterminate](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8617/commits/0853358b3d5cb9deaa00d51e96c5a8a5aec49723)

apps/web/src/components/PoolCandidatesTable/PoolCandidateFilterDialog.tsx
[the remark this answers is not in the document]
esizer: Added in [update filter select field null selections](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8617/commits/72b9f093bfe21be5c69866ce9f0512d298c3c803). If we find the label is not great, we can update it 👍

apps/web/src/components/PoolCandidatesTable/PoolCandidateFilterDialog.tsx
```diff
@@ -0,0 +1,317 @@
...
+              "isIndigenous",
+              getEmploymentEquityGroup("indigenous"),
+            ),
+            equityOption(
+              "isVisibleMinority",
+              getEmploymentEquityGroup("minority"),
+            ),
+          ]}
+        />
+        <Checklist
```
petertgiles, line 148: This is probably pre-existing as well so feel free to spin it off.  This seems like a very strange way to query priorities.  Why are we not querying the actual fields in `users` and instead using abstract weight values?

image

apps/web/src/components/PoolCandidatesTable/PoolCandidateFilterDialog.tsx
```diff
@@ -0,0 +1,317 @@
...
+            }))}
+          />
+        </div>
+        <div
+          data-h2-display="base(flex)"
+          data-h2-flex-direction="base(column)"
+          data-h2-gap="base(x.5)"
+          data-h2-align-self="l-tablet(center)"
+        >
+          <Checkbox
```
petertgiles, line 257: Oops, this is being sent as a null.
image

petertgiles (changes requested):
So close!  Almost there.

apps/web/src/components/PoolCandidatesTable/PoolCandidateFilterDialog.tsx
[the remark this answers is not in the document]
petertgiles: Good for me.  :+1:

apps/web/src/components/PoolCandidatesTable/PoolCandidateFilterDialog.tsx
esizer: Yeah, this was preexisting. I also found this a little odd.

I'll take a look and see if I can spin this off. I wasn't 100% sure how querying by priority worked so I left it as untouched as I could to be safe 😅

apps/web/src/components/PoolCandidatesTable/PoolCandidateFilterDialog.tsx
esizer: Looks like this is yet another filter that is purposely being left out of the query 😞 I'll try to remember to bring it up during stand up to see what we want to do.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/daf6a72d8a978b94fc140ab865c9e6c14d20be68/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx#L299

apps/web/src/components/PoolCandidatesTable/PoolCandidateFilterDialog.tsx
esizer: Okay, I tried my best at defining the ticket. Hopefully this makes sense, if not adjustments are welcome 👏 

 - https://github.com/GCTC-NTGC/gc-digital-talent/issues/8668

petertgiles: Thanks!  I added a bit more detail.

petertgiles (approved):

Issue #4529, opened 2022-10-31
Refactor `AuthenticationContainer` to sync tokens between tabs


tristan-orourke:
NOTE: if #2572 is already complete, this is likely unnecessary.

Based off concerns described in #4526.

Right now we keep our tokens in a useState hook, occasionally save/load to localStorage, and as of #4526 exclusively use refresh tokens directly from localStorage. We could make this code much simpler, and handle the concerns mentioned in #4526, by writing a custom hook which has an interface like useState but keeps values synced with localStorage (or sessionStorage, for that matter), and keeps them synced between tabs. We can follow the example here, and likely make use of our existing useStorage hook.

vd1992:
Appears not needed as of #8618

PR #8618, opened 2023-11-28
Bump auth and refactor

### LINKED ISSUE #7887: ♻️ Try Auth Bump Again
## ♻️ Debt/Refactor

We found an issue with the auth bump branch during deployment and had to quickly roll it back.  Let's try to bump these dependencies again.

## 🕵️ Details

Original PR: https://github.com/GCTC-NTGC/gc-digital-talent/issues/6547
PR Revert: https://github.com/GCTC-NTGC/gc-digital-talent/issues/7886

## ✅ Acceptance Criteria

- [ ] urql bumped back to version 4
- [ ] prod builds authenticate properly

petertgiles:
🤖 Resolves #7887

## 👋 Introduction

This branch starts with Eric's original "bump auth" branch and fixes the behavior when building in prod mode.  It also does some refactoring to clarify the flow.

## 🕵️ Details

I believe the original issue is that in prod mode React with only run the useEffect calls once.  Where we ran into problems was we set the tokens into local storage from the URL in a useEffect hook and I think that it was stored too late for the client to grab.  I've pulled  the local storage setting out of the useEffect and only mutated the URL in the useEffect.

I also found it very hard to manage synchronizing the state between the useState hook, the provider, and the local storage.  I've tried to simplify this by only using local storage and having the provider only pass down the loggedIn status.  Now the auth container can log in and out and it's up to the client to pull its own tokens when the state changes.

I've got a bunch more refactoring I want to do but this branch is already worryingly big....

I also had to rip out a bunch of tests.  Now that the functions are inline it's harder to test them.  The functions are so thin, I'm not sure how much value they were bringing anyway?  I could post an issue for better client tests?

## 🧪 Testing

Build the app __in prod mode__ and with __sign in canada__.

- [ ] Can log in.
- [ ] Can wait five minutes and then the token is refreshed on the next API call.
- [ ] When you have two tabs open, a refresh in one will allow the second tab to make an API call with the new tokens and no refresh.
- [ ] If [the introspection fails](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6547#pullrequestreview-1602788860) the user is immediately logged out.
- [ ] If you log in as a deleted user you end up on the "user deleted" page.


vd1992:
SiC testing

- Can login 
- Refresh token fetched after 5 minutes
- Second tab not affected by refresh action
- Can logout, affects all logged in tabs
- Login as deleted user redirects to deleted page

vd1992:
`npm run dev`

For the most part, things seem the exact same, aside from logging being duplicated sometimes. 
Had one instance where the refresh token with SiC failed after 5 minutes, redirecting to logged out page, but was unable to replicate it right after.

vd1992 (approved):
All seems to work reliably to me, amazing 🙇

esizer (approved):
Awesome! 🤩 

I noticed one thing with a conditional. I tihnk it could be simplified but I;m not 100% confident on that 😅 Just like 99% 😛 If I am wrong, this is good to go so I am approving 👍 

I do have a uqestion around the `/logged-out` experience (that is our of scope) in the event of an auth issue. I know we did a custom message for a deleted user and it seems like it would be beneficial to do something similar for this type of unintentional logout. Right now, if the auth fails for any reason (can't refresh, request failure, etc) we dsiplay a success message. Do we have a ticket to work on this? 🤔 

I honestly think we should fix this because as a user I may get a little worried and suspicious if I did not intend to logout and saw a success message. I'd an error message  to let the user know it was our fault and not theirs.

vd1992:
Think this makes #4529 fine with being closed?

petertgiles:
> I do have a uqestion around the `/logged-out` experience (that is out of scope) in the event of an auth issue. I know we did a custom message for a deleted user and it seems like it would be beneficial to do something similar for this type of unintentional logout. Right now, if the auth fails for any reason (can't refresh, request failure, etc) we display a success message. Do we have a ticket to work on this

I agree with this and I don't think we have a ticket for it.  I recommend getting rid of the "user deleted" page and just have a reason message on the logout page.
- You have been successfully logged out. (manual logout)
- Your session was ended.  (refresh failed, unknown auth error)
- Your account is inactive. (user deleted)

petertgiles:
Added two issues:
Detailed logout messages #8652
Improve login/logout tests #8654

esizer:
> I agree with this and I don't think we have a ticket for it. I recommend getting rid of the "user deleted" page and just have a reason message on the logout page.
> 
> * You have been successfully logged out. (manual logout)
> * Your session was ended.  (refresh failed, unknown auth error)
> * Your account is inactive. (user deleted)

Awesome, I love this idea! 👍

PR #8643, opened 2023-11-30
[Deps] Bump storybook to v7.6.1

esizer:
## 👋 Introduction

Groups the storybook dependabot updates for single merge.

## 🧪 Testing

1. Install deps `npm i`
2. Run storybook `npm run storybook`
3. Confirm it works


mnigh (commented):
Tried running storybook twice after `npm ci` and got stuck both times. Anyone else having the same problem?

```
64% building 2/2 entries 17917/18825 dependencies 5807/5811 modules
Cannot read properties of undefined (reading 'buildError')
```

## Screenshot

Screen Shot 2023-11-30 at 10 36 26

vd1992:
I see it too. But I can navigate through the stories still

image

`npm run storybook:design-system` has no error
`npm run storybook:web` does have it

I wonder if one story is having issues and we'll need to comb through and find it

esizer:
> I wonder if one story is having issues and we'll need to comb through and find it

I am seeing the same thing. I'm building right now and going through the stories one by one to see if anything pops up. In the meantime, they released 2 bug fixes so I bumped by one more patch release 😛

esizer:
Okay, so I went through each story one by one and outside of some duplicate key errors I couldn't find anything. Going back and reading the error, I tihnk it is of no concern (as long as we can all still run and open storybook). What it says seems like a minor issue in the storybook cli.

`Cannot read properties of undefined (reading 'buildError')`.

My hunch tells me they have something like this:

```ts
type Errors = {
 ...
 buildError?: Error;
}

const errors: Errors | undefined;

// Not checking to see if the key exists
log(errors.buildError);

// What they should do
log(errors?.buildError);
```

So, they are trying to access a key on an object that is `undefined` because there are no errors.

> [!IMPORTANT]
> I also ran [chromatic](https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/7051250933/job/19194130926?pr=8643) and it seems happy 🤷‍♀️

mnigh (approved):
Love me a good alias: all the cool [kids](https://backtothefuture.fandom.com/wiki/Marty_McFly) have 'em! 👍

## Screenshot

Screen Shot 2023-11-30 at 15 25 49

PR #8656, opened 2023-12-01
[Style] Update `Tab` styling

### LINKED ISSUE #8582: ♻️ Update the tab / tab group UI component
image

## ✨ Summary

[Visit the Figma file](https://www.figma.com/file/guHeIIh8dqFVCks310Wv0G/Style-library?type=design&node-id=912-23092&mode=design) • Contact @substrae with design questions

This ticket aims to update the visual styles of the tab UI component, as well as to ensure that it behaves consistently across the application. This is also a precursor update required by the new [admin header UI component](https://github.com/GCTC-NTGC/gc-digital-talent/issues/8584).

```[tasklist]
### ✅ Requirements
- [ ] Update the individual `Tab` UI component to match the designs
- [ ] Update the `TabGroup` UI component to match the designs
- [ ] Ensure all instances of tabs on the site are using this updated component
- [ ] Ensure all details outlined below are met
- [ ] Ensure tabs are accessible based on the [Tabs pattern documentation](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/)
- [ ] Apply dark mode styles and ensure Storybook has a side-by-side dark mode comparison
```

### Background

Note that the tab group should adopt the background the element is sitting on. Most of the time this will be `background`, but in some edge cases, it might sit on other colors, such as `foreground`.

<details>
  <summary>📸 `background` example</summary>
  
</details> 

<details>
  <summary>📸 `foreground` example</summary>
  
</details> 

### Dark mode

<details>
  <summary>📸 Dark mode styles</summary>
  
</details> 

### Hover

Hover effects should trigger only on tabs that are clickable (i.e. not the active tab).

<details>
  <summary>📸 Hover effect</summary>
  
</details> 

### Focus

Open to suggestions here if we feel the focus change isn't bold enough.

<details>
  <summary>📸 Focus effect</summary>
  
</details> 

### Mobile experience

I'm basing this approach on Material's tab component. There are conflicting opinions on the best way to handle tabs that exceed available screen space, but my gut tells me that this is the best approach based on the different implementations I've tried over the years. The key to this approach is ensuring that the tabs provide visual indication that they continue to the left and/or right by allowing them to be cut off by the edge of the screen (not the padding of their container). It's also important that when using keyboard navigation that the are scrolls as out-of-sight tabs are focused.

https://m3.material.io/components/tabs/guidelines#b0e3794c-059f-41e2-87a6-fe2498fc7fe8

<details>
  <summary>📸 On narrow screens (initial)</summary>
  
</details> 

<details>
  <summary>📸 On narrow screens (scrolled)</summary>
  
</details>

esizer:
🤖 Resolves #8582 

## 👋 Introduction

Updates the tab styles for mobile and dark mode.

## 🧪 Testing

2. Navigate to the "Tabs" story
3. Confirm it matches the mockup
4. Confirm tabs work on mobile

## 📸 Screenshot
image

### DIFF
--- a/packages/ui/src/components/Tabs/Tabs.tsx
+++ b/packages/ui/src/components/Tabs/Tabs.tsx
@@ -20,57 +20,73 @@ const List = React.forwardRef<
   React.ComponentPropsWithoutRef<typeof TabsPrimitive.List>
 >(({ children, ...rest }, forwardedRef) => (
   <TabsPrimitive.List
-    ata-h2-max-width="base(100%)"
-    data-h2-overflow-x="base(scroll)"
+    data-h2-max-width="base(100%)"
+    data-h2-display="base(flex)"
+    data-h2-gap="base(x.25)"
+    data-h2-overflow="base(auto visible)"
+    data-h2-width="base(100%)"
+    data-h2-padding="base(0 x1)"
+    data-h2-z-index="base(1)"
+    data-h2-overflow-x="base(auto)"
+    data-h2-overscroll-behavior-x="base(contain)"
+    data-h2-scroll-snap-align="base(start)"
+    data-h2-scroll-snap-type="base(x mandatory)"
+    data-h2-scrollbar-width="base:hover(none)"
     ref={forwardedRef}
     {...rest}
   >
-    <div
-      data-h2-display="base(flex)"
-      data-h2-gap="base(x.25)"
-      data-h2-overflow="base(auto visible)"
-      data-h2-width="base(100%)"
-      data-h2-padding="base(0 x1)"
-      data-h2-z-index="base(1)"
-    >
-      {children}
-    </div>
+    {children}
   </TabsPrimitive.List>
 ));
 
 const Trigger = React.forwardRef<
   React.ElementRef<typeof TabsPrimitive.Trigger>,
   React.ComponentPropsWithoutRef<typeof TabsPrimitive.Trigger>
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/Tabs/Tabs.tsx
```diff
@@ -8,52 +8,98 @@ const Root = React.forwardRef<
...
-    ref={forwardedRef}
-    {...props}
-  />
-));
+>(({ children, ...rest }, forwardedRef) => {
+  const handleFocus: React.FocusEventHandler<HTMLButtonElement> = (event) => {
+    event.currentTarget.scrollIntoView({
+      block: "center",
+    });
+  };
```
vd1992, line 50: Is the centring along the vertical axis intentional? 

https://recordit.co/s86dQvUTRi

esizer: Oops, no. That was to get the mobile views to scroll if you were using a keyboard but it seems to be scrolling the document and not the parent 😬 .

I may need to look into getting that working. Or maybe it is not needed 🤔

esizer: Alright, I believe I figured it out! Had to do some math but it seems to be working now 😅 

[fix mobile scroll logic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8656/commits/50499cf7c9ab3cb05938e11a8c5f57d45127f2ed)

Issue #8401, opened 2023-11-03
🐛 Gap between npm watch and codegen actions


petertgiles:
## 🐛 Bug

I've noticed that sometimes changes to a graphql file while running a top-level watch command don't end up in the packages/graphql/index.ts file but appear as expected in the apps/web/src/api/generated.ts.  If I manually run codegen in the packages/graphql directory the changes appear but then they (maybe?) get overwritten again on the next watch cycle.  I haven't pinned down exactly what's going on yet, though.   I'm not sure if this has always been the case but I first started noticing it about 1-2 months ago. I've just stared running npm run codegen -- --force before watching and it seems to work fine.

## 🙋‍♀️ Proposed Solution

There is probably a missing dependency in the turbo.json file

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the bug was addressed.

- [ ] When running top-level watch, changes to a graphql file should end up in both packages/graphql/index.ts and  apps/web/src/api/generated.ts files.

petertgiles:
Pinning this to help with #8673

PR #8673, opened 2023-12-05
[Debt] Migrate codegen to `client-preset`

### LINKED ISSUE #4662: `graphql-codegen` recommends `client-preset`
## 👋 Introduction

Currently, we are using the plugin `typescript-urql` to generate hooks and types for `urql`. However, `graphql-codegen` is now recommending the use of `client-preset`.

REF: https://www.the-guild.dev/graphql/codegen/plugins/typescript/typescript-urql

## 🕵️ Details

The library owners say that `client-preset` offers a better developer experience as well as smaller bundle size. It may be within our interest to migrate away from `typescript-urql`. 

At the same time, since we have migrated to `react-router`, we may be able to use this as an opportunity to research using `urql` without the hooks so we can utilize [loaders](https://reactrouter.com/en/main/route/loader) to simplify our page routes.

esizer:
🤖 Resolves #4662 

## 👋 Introduction

Migrates our current codegen implementation to the `client-preset`.

## 🕵️ Details

There is a lot of unpack here since it is a dramatic shift from the our current usage. We had a learning session where we talked about how it works with fragment masking but, that was a while ago now. I'm going to try and give some details about what is doing on here but of course, reach out if you have any question's and I'll do my best to answer!

> [!TIP]
> I highly recommend reading the article "[Unleash the power of Fragments with GraphQL Codegen](https://the-guild.dev/blog/unleash-the-power-of-fragments-with-graphql-codegen)"

### No more `*.graphql` files

The new preset uses function overloading to provide types for our documents. This means, we must define them in our TypeScript files rather than the GraphQL files. You will notice that we wrap these in a `graphql` function with the comment `/** GraphQL */`. The comment is similar to what we do in our PHP files to provide syntax highlighting so, while you can omit it, it is recommended to use that.

### No more hooks

The new present does not generate any hooks. We now use the hooks directly provided by `urql`. You may notice, if we change some values in our queries, or create a new one, TypeScript is not happy. This is due to the function overload. If you modify a query, you must re-run codegen for the types to be picked up.

### Fragment Masking

This is by far the most beneficla side effect of this migration. Yoiu may notice, we do not pass items in as their type anymore and instead use  new types for our props that come from the API then, we unwrap it in the component. The beneift of this is that codegen can infer and strongly type what we are passing. No more accidently not querying stuff, you will get a build error! 🎉 

```tsx
const Parent_ChildFragment {
  Parent_ChildFragment  on Child {
    id
    name
  }
}

type ChildComponentProps {
  fragment: FragmentType<typeof Parent_ComponentFragment>;
}

const ChildComponent = ({ fragment }: ChildComponentProps) => {
  const child = getFragment(Parent_ComponentFragment, fragment);
  const name = child.name; // This is okay
  const email = child.email; // Will flag an error since we did not query it.
  
  ...
}
```

#### Mocking Fragments

Since we have changed how props are passed in, this also affects how we mock data for tests and storybook. Thankfully, the new preset gives us a handy helper in `makeFragmentData`. This is a simple function that takes in the data you want to use in the mock and the fragment you are mocking.

```tsx
const mockData = makeFragment(user, UserPage_UserFragment);

const Template = () => <UserPage userFragment={mockDate} />
```

## 🎯 Scope

This ticket originally had a much larger scope but has been reduced to make it easier for reviewers. The new scope is:

- Install, configure and update codegen for `@gc-digital-talent/graphql`
- Update import paths for majority of `./apps/web` for incremental adoption
- Hardcode in `undefined` to helpers that use the `Maybe` type for incremental adoption
- Implement it on a small portion of pages as a concrete proof of concept
   - `./packages/**` -> Required since they depend on the graphql package
   - `AdminDashboardPage` -> Basic, no fragment example
   - `CreateAccountPage` -> Mutation, storybook, test and fragment example
   - `CreateAccountRedirect` -> Just happened to be easy and made sense to do both account pages
   - `artisan-commands.cy.ts` -> Cypress example

## 🧪 Testing

1. Install new deps `npm i`
2. Force codegen refresh `npm run codegen -- --force`
3. Build `npm run dev`
4. Bounce around and confirm the app still functions as expected

##  ❓ Questions

During development, there were some questions I had about the implementation and scope of the ticket. They all have answers now. I will leave this here in case you would like to see what they were and the decisions we made. They are also recorded in the comments on this PR.

### Incremental Adoption

<details>
<summary>Question</summary>

This is actually working well as is and, we could adopt this incrementally since we have codegen running in two places. Thanks to @petertgiles brilliant idea of applying the preset in `@gc-digital-talent/graphql` and leaving our `apps/web` config alone, I was able to get it working with just two pages.

We could, if we wanted to, ship this as is and create tickets to adopt it on a per-section basis then one final one to just remove it entirely from `apps/web` (this was only left for familiarity when switching to packages).

However, now that I believe I have this figure out, it is actually going incredibly smoothly. My concern is not time/effort of implementation, rather, it is that of the reviewer 😬 

</details>

<details>
<summary>Answer</summary>

Going to create an epic (TO DO: Add link here!) that includes issues to implement the new pattern on a per-section basis to keep work digestible by reviewers.

</details>

### Git check-in

<details>
<summary>Question</summary>

I'm not sure if it was for examples sake but, they seem to be checking the generated files into git so I followed suit. I'm not sold on the idea and it won't takle much convincing for me to re-add the `.gitignore` for these files. Though, if you are wondering why the diff has so many lines added, that is why 😅 

</details>

<details>
<summary>Answer</summary>

We never checked these in and they are easily recreated so we will continue to ignore them for now.

</details>

### `Maybe`

<details>
<summary>Question</summary>

All of our plugins we were using have essentially been rolled up into the preset with some strong opinions. One of these opinions (that we, seemingly, cannot override) is the type signature for `Maybe`. It has dropped the `undefined` portion and now looks type `type Maybe<T> = T | null;`.

I found this was a little nicer to work with but our code heavily depended on the `undefined` portion so I went and manually added it every where. It would be nice to do some refactoring to no longer need this but this is already a large diff and it seemed like a can of worms.

</details>

<details>
<summary>Answer</summary>

We are okay with the new type definition and manual addition of `undefined` and will add refactor as AC for epic tickets.

</details>

### Naming Conventions

<details>
<summary>Question</summary>

We have an eslint rule for camel case, however, it seems the convention is to use pascal case for defining query docs and fragments. I went ahead and added anything ending in `Query`, `Fragment`, `Mutation` to the allow list for this rule. 

Does that make sense? 🤔 

</details>

<details>
<summary>Answer</summary>

Yes, this makes sense 👌 

</details>

### Watch command

<details>
<summary>Question</summary>

I think we could maybe work `codegen` into our `npm run watch` command to avoid this...maybe 🤔 

</details>

<details>
<summary>Answer</summary>

Going to create a ticket and add it to the epic to investigate this.

</details>


petertgiles:
> We could, if we wanted to, ship this as is and create tickets to adopt it on a per-section basis then one final one to just remove it entirely from apps/web (this was only left for familiarity when switching to packages).

I agree with that, yeah.  If you can break this into logical chunks that will help with review and merge, I think.  Please make sure it's really clear what part of the app is updated in _this_ PR to help reviewers.

> I'm not sure if it was for examples sake but, they seem to be checking the generated files into git so I followed suit. I'm not sold on the idea and it won't take much convincing for me to re-add the .gitignore for these files.

I also think we should not check these in, unless we see the reason why.  If we hit that reason, checking them in should be easy, right.

> I think we could maybe work codegen into our npm run watch command to avoid this...maybe 🤔

It would be nice, but I don't see it as a blocker.  Maybe you could spin off a new issue?  I did pin #8401 to get that dependency fixed, though.

> No more accidently not querying stuff, you will get a build error! 🎉

Sweet cuppin cakes, I can't wait for this...

> It has dropped the undefined portion and now looks type type Maybe<T> = T | null;.

I like this but I'm OK with patching in the undefined to make a smaller diff and remove later, too.  Or maybe if you're working on a smaller part of the app at a time it can be done at the same time?  :shrug: 

> I went ahead and added anything ending in Query, Fragment, Mutation to the allow list for this rule.

Seems fine to me.

esizer:
> I agree with that, yeah. If you can break this into logical chunks that will help with review and merge, I think. Please make sure it's really clear what part of the app is updated in _this_ PR to help reviewers.

Okay, I spoke with @vd1992 about how we would break this down and, I went with something completely different from both of our ideas after looking into it more 😅 .

I created an epic and added the issues to it and added all the blockers. If this approach does not make sense let me know! I tried to keep the scope of each issue relevant and small but I may have missed the mark.

 - https://github.com/GCTC-NTGC/gc-digital-talent/issues/8681

documentation/graphq.md
```diff
@@ -0,0 +1,155 @@
...
+
+### 👺 Masking
+
+With the fragments, we can use what is referred to as "fragment masking". What this provides is a way to strongly type the specific fragment used in a query for our consumption. Previously, we had some problems where we added additional information to a component, but since our type system was very relaxed (it showed us what *could* be queried, not what *was* queried) we sometimes forgot to update the query and ended up with empty values.
+
+Now, we get a type error if we use data we never queried and will get a build error.
+
+## 🛠️ Implementation
+
+All examples are related and show how we would build a page and its operations bottom up.
```
mnigh, line 25:
```suggestion
All examples are related and show how to build a page and its operations from the bottom up.
```

documentation/graphq.md
```diff
@@ -0,0 +1,155 @@
...
+
+### ⚛️ Component
+
+When defining a component that needs to consume data from the API, you need to do a few things for the code generator to work.
```
mnigh, line 29:
```suggestion
When defining a component that needs to consume data from the API, there are a few things to do for the code generator to work.
```

mnigh (commented):
Nice work! My suggestions and comments are strictly related to copy (mostly style) in `documentation/graphq.md` and one comment in another file.

documentation/graphq.md
```diff
@@ -0,0 +1,155 @@
...
+      <p>{pool.stream}<p>
+    </>
+  );
+}
+```
+
+
+### 📖 Storybook
+
+Since our props require a specific type that comes from the code generator, we need to massage the mock data we pass into stories. Thankfully, we have a nice little helper to do this.
```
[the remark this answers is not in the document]
esizer: Sounds good! Hopefully this makes sense: [add reference to fragment helper](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8673/commits/e8bafb76c1db0ff2a230a7d45862453fca3e74cd)

documentation/graphq.md
```diff
@@ -0,0 +1,155 @@
...
+
+const mockPools = fakePools();
+const mockPoolFragment = makeFragmentData(mockPools[0], PoolCard__PoolFragment);
+
+const Template = () => <PoolCard pool={mockPoolFragment} />;
+```
+
+### 🧪 Testing
+
+Similar to storybook, we just need to create the fragment using the helper.
```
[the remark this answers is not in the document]
esizer: Added in [add reference to fragment helper](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8673/commits/e8bafb76c1db0ff2a230a7d45862453fca3e74cd)

packages/date-helpers/src/index.ts
```diff
@@ -145,13 +144,13 @@ export const convertDateTimeZone = (
 
 // Convert a DateTime scalar to a Date by stripping off the time
 export const convertDateTimeToDate = (
-  d: Scalars["DateTime"],
-): Scalars["Date"] => {
+  d: Scalars["DateTime"]["input"],
```
petertgiles, line 147: Huh, this is interesting.  The new codegen has separate input and output scalar types?

petertgiles (approved):
Very cool!  I'm looking forward to writing some queries of my own to really dig into this change.  Log in, log out, new account, and admin dashboard are all working well for me.

packages/date-helpers/src/index.ts
esizer: Yeah! And we can define them depending on what we need to pass in or get out of the API 😁 . So if we wanted to, i think we could do like:

```ts
{
  DateTime: {
    input: string;
    output: Date;
  }
}
```

PR #8680, opened 2023-12-06
[Feature] Updates support email

### LINKED ISSUE #8674: ✨ New support email at talent.canada.ca
## ✨ Feature

We have a new support email that we have more control over and shares the same domain as our site :)

Let's replace the current support email everywhere it appears with: support-soutien@talent.canada.ca

## ✅ Acceptance Criteria

- [x] support-soutien@talent.canada.ca is used with the support form
- [ ] ~support-soutien@talent.canada.ca is provided on job posters~
- [x] support-soutien@talent.canada.ca replaces the existing support email everywhere else as well
- [x] Update `AccessbilityStatementPage` to reference constant
- [ ] Deployment env change

mnigh:
🤖 Resolves #8674.

## 👋 Introduction

This PR updates `AccessbilityStatementPage` to reference the `TALENTSEARCH_SUPPORT_EMAIL` constant which itself gets its value from an identically named environment variable called `TALENTSEARCH_SUPPORT_EMAIL`. It also updates the fallback value for the TALENTSEARCH_SUPPORT_EMAIL constant to be [support-soutien@talent.canada.ca](mailto:support-soutien@talent.canada.ca) and updates a reference to the support email in the security policy file.

## 🧪 Testing

1. Update `TALENTSEARCH_SUPPORT_EMAIL` environment variable in `apps/web/.env` to `support-soutien@talent.canada.ca`
2. `npm run build`
3. Navigate to `/accessibility-statement`
4. Confirm email address rendered in feedback and contact information section is support-soutien@talent.canada.ca

## 🚚 Deployment

> [!IMPORTANT]  
> Update `TALENTSEARCH_SUPPORT_EMAIL` in UAT, DEV, and PROD from `gctalent-talentgc@support-soutien.gc.ca` to `support-soutien@talent.canada.ca`

### DIFF
--- a/apps/web/src/constants/talentSearchConstants.ts
+++ b/apps/web/src/constants/talentSearchConstants.ts
@@ -1,5 +1,5 @@
 export const TALENTSEARCH_SUPPORT_EMAIL =
   (process.env.TALENTSEARCH_SUPPORT_EMAIL as string) ??
-  "gctalent-talentgc@support-soutien.gc.ca";
+  "support-soutien@talent.canada.ca";
 export const API_SUPPORT_ENDPOINT =
   (process.env.API_SUPPORT_ENDPOINT as string) ?? "/api/support/tickets";
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,3 +1,3 @@
 # Security
 
-Do not post security vulnerabilities to our public repository. Security vulnerabilities should be reported by email to <gctalent-talentgc@support-soutien.gc.ca>.
+Do not post security vulnerabilities to our public repository. Security vulnerabilities should be reported by email to <support-soutien@talent.canada.ca>.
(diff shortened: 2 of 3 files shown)


apps/web/src/constants/talentSearchConstants.ts
```diff
@@ -1,5 +1,5 @@
 export const TALENTSEARCH_SUPPORT_EMAIL =
   (process.env.TALENTSEARCH_SUPPORT_EMAIL as string) ??
-  "gctalent-talentgc@support-soutien.gc.ca";
+  "support-soutien@talent.canada.ca";
```
vd1992, line 3: Less a request and more a question, why is this like it is?

image

image

image

There is no environment variable `TALENTSEARCH_SUPPORT_EMAIL` so all it does it return the fallback value in which case why refer to the environment variable in the first place rather than just return a constant from the start?

vd1992: Environment variables

image

mnigh: > Less a request and more a question, why is this like it is?
> 
> There is no environment variable `TALENTSEARCH_SUPPORT_EMAIL` so all it does it return the fallback value in which case why refer to the environment variable in the first place rather than just return a constant from the start?

The reason for the environment variable is so that different values can be uses in different environment. For example, if a user wanted to test having a different email address (ex., support-soutien-test@talent.canada.ca) to send to without having to alter the code on the UAT, we would be able to switch the value without committing code.

petertgiles (approved):
Nice.  :+1:

PR #8726, opened 2023-12-07
[Fix] Add all experience skills to sync method on experience card

### LINKED ISSUE #8727: 🐛 Prod Bug: Updating an experience-skill on Application can remove the experience from other skills
## 🐛 Bug

Multiple users reported having some of their experience-skills deleted when adding experiences to skills on the application to a live job.

## 🦋 Expected Behaviour

Users should be able add experiences to skills and update those connections without anything being deleted unexpectely!

## 🕵️ Details

We determined that a `sync` mutation was causing the problem.

## 📋 Steps to Reproduce

1. Start an application to a poster
2. Continue to the skills step
3. Add an experience to a skill
4. Add the same experience to another skill
5. Edit and save the write-up for one of those experience-skills
6. The experience will be removed from the other skill!

## 📸 Screenshot

## 🙋‍♀️ Proposed Solution

Replace the sync with a connect mutation?

## ✅ Acceptance Criteria

- [ ] Users can add and update experience-skills without any danger of removing unrelated experience-skills.

esizer:
🤖 Resolves #8727

## 👋 Introduction

The solution is not the best, but it works as a quick hot fix. We should revist this at some point.

## 🕵️ Details

We were filtering out un-related skills on the application skill page which means we only were syncing the skill details for that specific experience and it was being disconnected from the rest. This drills all skills down all the skills and merges the sync array.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/defd67584442596344b87382e568c4e430a5f830/apps/web/src/pages/Applications/ApplicationSkillsPage/components/SkillTree.tsx#L95

## 🧪 Testing

1. Build `npm run dev`
2. Start an application
3. Contine to the skills step
4. Connect and experience to two different skills
5. Update one of those experience details using the edit button on the experience card
6. Confirm it does not disconnect from the other skills on success

## 📸 Screenshot

https://github.com/GCTC-NTGC/gc-digital-talent/assets/4127998/7231de48-f699-439d-90ce-76d2f922bcbe

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -1257,6 +1257,8 @@ input AwardExperienceInput {
 
 input UpdateExperienceSkills {
   connect: [ConnectExperienceSkills!] @resolveWith(method: "connectSkills")
+  # This is an alias of connect since it is a custom command that can do both (insert and update)
+  # It is here to avoid confusion when using it on the frontend
   update: [ConnectExperienceSkills!] @resolveWith(method: "connectSkills")
   disconnect: [ID!] @resolveWith(method: "disconnectSkills")
   sync: [ConnectExperienceSkills!] @resolveWith(method: "syncSkills")


tristan-orourke:
Instead of passing down all the skills to sync, would it work if we just replaced `sync` with `connect`?

tristan-orourke:
If `connect` is specific to new additions, seems to me like adding an `update` [nested mutation](https://lighthouse-php.com/6/eloquent/nested-mutations.html#nested-mutations) to the API would be the best solution.

tristan-orourke:
However, fun fact, the mutation in this case isn't actually using _Lighthouse's_ `connect` function, it's using a custom function that I wrote, which DOES work fine with existing entities 😆 

```
input UpdateExperienceSkills {
  connect: [ConnectExperienceSkills!] @resolveWith(method: "connectSkills")
  disconnect: [ID!] @resolveWith(method: "disconnectSkills")
  sync: [ConnectExperienceSkills!] @resolveWith(method: "syncSkills")
}
``` 

(The custom function was necessary because of the complex way Experiences and Skills are related to each other in the database.)

tristan-orourke:
Maybe the best of all worlds is to add the `update` alias to the api, so we can keep following the lighthouse pattern, even if its using the same underlying function.

```
input UpdateExperienceSkills {
  connect: [ConnectExperienceSkills!] @resolveWith(method: "connectSkills")
  update: [ConnectExperienceSkills!] @resolveWith(method: "connectSkills")
  disconnect: [ID!] @resolveWith(method: "disconnectSkills")
  sync: [ConnectExperienceSkills!] @resolveWith(method: "syncSkills")
}
```

esizer:
> Maybe the best of both worlds is to add the `update` alias to the api, so we can keep following the lighthouse pattern, even if its using the same underlying function.

This is a much better implementation and was easy to do so I reverted my quick and ugly prop drill for this. Works well, thanks 👍 

[implement update mutation for experience skills](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8726/commits/1ae8c8169eeb12617dbcb39c8657179286ed3048)

package-lock.json
petertgiles, line 1: Why is there a package lock change in this branch?

api/graphql/schema.graphql
```diff
@@ -1257,6 +1257,7 @@ input AwardExperienceInput {
 
 input UpdateExperienceSkills {
   connect: [ConnectExperienceSkills!] @resolveWith(method: "connectSkills")
+  update: [ConnectExperienceSkills!] @resolveWith(method: "connectSkills")
```
petertgiles, line 1260: Could you please leave a comment for future explorers on why this is so weird?  😅

esizer: Added in [document aliased input item](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8726/commits/638cc68581f9032d09cbb6129fc3ba71c09db5e9)

petertgiles (approved):
It works!  And its nice and clean!

PR #8747, opened 2023-12-07
[Fix] Add codegen to watch command

### LINKED ISSUE #8401: 🐛 Gap between npm watch and codegen actions

esizer:
🤖 Resolves #8401 

## 👋 Introduction

This allows codegen to re-run on changes to our operations. It also prevents codegen from being cached since this seemed to be causing more problems than it was solving.

## 🕵️ Details

We were only running `codegen` on the start of the watch command but it never actually watched any files. This [uses concurrently](https://the-guild.dev/graphql/codegen/docs/getting-started/development-workflow#watch-mode) to have it run in watch mode alongside the `tsup` watch.

> [!NOTE]
> Something that is a tad annoying is that when you kill the process, you need to do it twice. This is [a known bug](https://github.com/open-cli-tools/concurrently/issues/452) and hopefully gets fixed.

## 🧪 Testing

1. Run watch `npm run watch`
2. Go add or delete a query in `apps/web`
3. Confirm the command is re-run and it affects the build


vd1992:
Cannot run on WSL

Running in a Linux environment
Running watch
-> off main, deleting a graphql operation causes no change 
-> off branch, deleting a graphql operation triggers codegen action and webpack compile but no errors are emitted

esizer:
> Cannot run on WSL

Hmm, odd. I am also on WSL and it seemed to work for me 🤔 

> -> off branch, deleting a graphql operation triggers codegen action and webpack compile but no errors are emitted

Which query did you delete? Do you remember? I'd like to see if I can reproduce.

petertgiles:
I'm not sure what to think here.  :laughing:  Codegen doesn't seem to be watching for changes to the .graphql files.  :-1:    But I can manually run codegen with having to do the force flag.  :+1: 

On the other hand, the new stuff with the client-preset is working great in watch mode, so maybe we don't care that much about the deprecated codegen stuff?

esizer:
> I'm not sure what to think here. :laughing: Codegen doesn't seem to be watching for changes to the .graphql files. :-1: But I can manually run codegen with having to do the force flag. :+1:

Ohh, that is a good point. Ever since we merged the `client-preset` (after I started this branch) the `.graphql` files are watched in a different place. I think I can resolve this, I had it working but merged main this morning which probably broke that 😅 .

petertgiles (approved):
Yay!  This is working great for me now.

esizer:
> Yay! This is working great for me now.

Awesome! This is going to make the `client-preset` migrations far more enjoyable. You have no idea how many times I have ran `npm run codegen -- --force` in the past 24 hours 💀

PR #8748, opened 2023-12-07
[Debt] Removes `user.isIndigenous` GraphQL field

### LINKED ISSUE #7662: ♻️ Remove deprecated `user.isIndigenous` field
## ♻️ Debt/Refactor

This field is deprecated.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/4000aacaaf392bc2a0b763d4d9dbd9f67cb2680f/api/graphql/schema.graphql#L83-L87

## ✅ Acceptance Criteria

- [ ] `user.isIndigenous` field removed from schema file
- [ ] References to this field have been removed or replaced

mnigh:
🤖 Resolves #7662.

## 👋 Introduction

This PR removes the deprecated `user.isIndigenous` GraphQL field.

## 🎩 Acknowlegments
@esizer for going down the 🐰 🕳️. 

### 🕵️ Details

> [!NOTE]  
> There were a few fixes out of the original scope of this issue but are important fixes that are directly related to the `isIndigenous` field.

- [Fix scopeEquity query for indigenous_communities](https://github.com/GCTC-NTGC/gc-digital-talent/commit/663e73d8a1c2fd8c9d5f8a8b829f08ca1bf2577e): had limited any search requests to **only** `LEGACY_IS_INDIGENOUS`
- [Fix isIndigenous reference](https://github.com/GCTC-NTGC/gc-digital-talent/commit/8e5fa65007dd06b5096ddb8072cf99e597eb5993): had limited rendering to **only** `LEGACY_IS_INDIGENOUS`

## 🧪 Testing

1. Ensure no references to `user.isIndigenous` GraphQL field

### DIFF
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -4,7 +4,6 @@ namespace App\Models;
 
 use App\Enums\CandidateExpiryFilter;
 use App\Enums\CandidateSuspendedFilter;
-use App\Enums\IndigenousCommunity;
 use App\Enums\LanguageAbility;
 use App\Enums\PoolCandidateStatus;
 use App\Enums\PublishingGroup;
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -605,11 +605,13 @@ class User extends Model implements Authenticatable, LaratrustUser
             array_push($equityVars, 'is_visible_minority');
         }
 
-        // 3 fields are booleans, one is a jsonb field, isIndigenous = LEGACY_IS_INDIGENOUS
+        // 3 fields are booleans, one is a jsonb field
         $query->where(function ($query) use ($equityVars) {
             foreach ($equityVars as $index => $equityInstance) {
                 if ($equityInstance === 'is_indigenous') {
-                    $query->orWhereJsonContains('indigenous_communities', IndigenousCommunity::LEGACY_IS_INDIGENOUS->name);
+                    foreach (array_column(IndigenousCommunity::cases(), 'name') as $indigenousOption) {
+                        $query->orWhereJsonContains('indigenous_communities', $indigenousOption);
+                    }
                 } else {
                     $query->orWhere($equityVars[$index], true);
                 }
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -588,8 +588,8 @@ class User extends Model implements Authenticatable, LaratrustUser
             return $query;
         }
 
-        // OR filter - first find out how many booleans are true, create array of all true equity bools
-        // equity object has 4 keys with associated bools
+        // OR filter - first find out how many booleans are true, create array of all true equity booleans
+        // equity object has 4 keys with associated booleans
         $equityVars = [];
         if (array_key_exists('is_woman', $equity) && $equity['is_woman']) {
             array_push($equityVars, 'is_woman');
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -693,22 +693,6 @@ class User extends Model implements Authenticatable, LaratrustUser
         return $query;
     }
 
-    /* accessor to maintain functionality of to be deprecated isIndigenous field */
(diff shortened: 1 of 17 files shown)


api/app/Models/User.php
```diff
@@ -605,11 +605,13 @@ public static function scopeEquity(Builder $query, ?array $equity): Builder
...
 
-        // 3 fields are booleans, one is a jsonb field, isIndigenous = LEGACY_IS_INDIGENOUS
+        // 3 fields are booleans, one is a jsonb field
         $query->where(function ($query) use ($equityVars) {
             foreach ($equityVars as $index => $equityInstance) {
                 if ($equityInstance === 'is_indigenous') {
-                    $query->orWhereJsonContains('indigenous_communities', IndigenousCommunity::LEGACY_IS_INDIGENOUS->name);
+                    foreach (array_column(IndigenousCommunity::cases(), 'name') as $indigenousOption) {
+                        $query->orWhereJsonContains('indigenous_communities', $indigenousOption);
+                    }
```
tristan-orourke, line 614: Can we simplify this and simply check that it's not empty? 
`$query->whereJsonLength('indigenous_communities', '>', 0)

vd1992 (approved):
- Can apply through the IAP context poster
- Update declaration normally
- Can filter for single enums or groups that don't include legacy

I think... It all works fine to me. Anything that comes up from probing prod can be a separate issue. 

Error alert when communities array is empty, as it should be

image

tristan-orourke (changes requested):
Looks good, I'd just like to see that scope query simplified.

api/app/Models/User.php
mnigh: cool, this reads as if it is the same logic as the current code but more concise, but i was having trouble testing this so i ran `dd($query->toSql());` before the return statement to see what the actual queries being built were.

## current
```
select * from "users" where (("indigenous_communities")::jsonb @> ? or ("indigenous_communities")::jsonb @> ? or ("indigenous_communities")::jsonb @> ? or ("indigenous_communities")::jsonb @> ? or ("indigenous_communities")::jsonb @> ? or ("indigenous_communities")::jsonb @> ?) and "users"."deleted_at" is null
```

## proposed
```
select * from "users" where (jsonb_array_length(("indigenous_communities")::jsonb) > ?) and "users"."deleted_at" is null
```

@tristan-orourke is the proposed query what you intended? i get an sql error trying to run either of these queries in postico or adminer.

tristan-orourke: That's what I was intending, yes. Not sure where why the error 😕

mnigh: the errors were my fault because i wasn't passing anything in to the question mark operator for the prepared statement (thanks @esizer!). 

### correct example
```
select * from "users" where (("indigenous_communities")::jsonb @> '"STATUS_FIRST_NATIONS"'::jsonb or ("indigenous_communities")::jsonb @> '"NON_STATUS_FIRST_NATIONS"'::jsonb or ("indigenous_communities")::jsonb @> '"INUIT"'::jsonb or ("indigenous_communities")::jsonb @> '"METIS"'::jsonb or ("indigenous_communities")::jsonb @> '"OTHER"'::jsonb or ("indigenous_communities")::jsonb @> '"LEGACY_IS_INDIGENOUS"'::jsonb) and "users"."deleted_at" is null;
```

@tristan-orourke the difference i can see in the logic is that the current one looks for the specific values in the `indigenous_communities` field whereas your proposal would look for any values in the `indigenous_communities` field. it leaves a small possibility that if there was a value like "N/A" or something other than the set enums was forced in, it would fulfill your proposed solution. are you ok with that? if so, i'll go ahead and commit it.

tristan-orourke: I'm okay with this because currently we simply pass in the full set of enum values anyway. Meaning that if we added a value which shouldn't be considered "true", we'd need to amend the code either way. So I prefer checking that it is not empty.

PR #8757, opened 2023-12-08
[Feature] Async operation proof of concept

### LINKED ISSUE #7082: ✨ Proof of concept asynchronous operation

petertgiles:
🤖 Resolves #7082 

> [!CAUTION]
> This is a proof of concept branch.  It should not be merged!

## 👋 Introduction

This branch shows a proof of concept async operation.  It provides a simple UI to request a magic word after a certain number of seconds.  After the delay has passed it responds with a notification that it has thought about your word.  :grin: 

## 🕵️ Details

The core of this PR is the setup of the [Laravel Queues](https://laravel.com/docs/10.x/queues) system.  It also exercises the rarely used notifications system.

This PR also shows how to set up [Supervisor](http://supervisord.org/) in the infrastructure to launch the queue workers and restart them as needed.  We don't run the post_deploy setup script locally so if testing queued operations locally it would probably be better to use the sync connection or shell into the webserver container and run `php artisan queue:listen` manually.

## 🧪 Testing

1. Rebuild the app frontend and backend, including migrations.
2. Shell into the web server, change to the api directory and run `./artisan queue:work` to start the worker.
3. Log into the app.  You can't receive notifications if you're not logged in.
4. Watch the app poll the notifications.
5. Request a word and wait for the notification to come back.

Or, it is deployed to the dev vertical right now for testing.

## 📸 Screenshot

image

image

## 🚚 Deployment
- Need to update QUEUE_CONNECTION to "database" in IaC.  I made a temporary setting change in DEV only.


apps/web/src/components/Layout/NotificationBar.tsx
```diff
@@ -0,0 +1,86 @@
...
+  // https://github.com/urql-graphql/urql/pull/1374
+  React.useEffect(() => {
+    if (!notificationsResult?.fetching) {
+      const id = setTimeout(
+        () => executeNotificationsQuery({ requestPolicy: "network-only" }),
+        5000,
+      );
+      return () => clearTimeout(id);
+    }
+  }, [notificationsResult.fetching, executeNotificationsQuery]);
```
esizer, line 31: I know this is in draft but I couldn't help but check it out! 😅 This is cool but what would be really cool is if we could get it working with subscriptions 😁 (obviously a different ticket 😛 )

 - [Lighthouse Subscriptions](https://lighthouse-php.com/master/subscriptions/getting-started.html#setup)
 - [URQL Subscriptions](https://formidable.com/open-source/urql/docs/advanced/subscriptions/#react--preact)
 - [Pusher subscription exchange example](https://gist.github.com/rmosolgo/a4c098de02af63ec1cdd09a3573a216b)
    - It says it is for the ruby implementation but [lighthouse was based on that](Much of the credit should be given to the [Ruby implementation (opens new window)](https://github.com/rmosolgo/graphql-ruby/blob/master/guides/subscriptions/overview.md)as they provided a great overview of how the backend should work.) so maybe it wouldn't be much different

vd1992: To chip in on Peter's behalf, the idea of API subscriptions was floated and decided against already 
https://www.apollographql.com/docs/react/data/subscriptions/#when-to-use-subscriptions

esizer: I'm not necessarily saying we shouldn't use polling here, I actually think it makes sense to start with it, especially for our current needs. But, did we decide to never use subscriptions? 

Unless we plan to never need or support low latency, real time updates I feel like it would be beneficial to explore in a different spike 😕 . Just like how you shouldn't use subscriptions for all async updates, we also shouldn't be using polling for all async updates. For what we need it for right now, polling is ideal. But, I remember we discussed having a notification system for things like status updates/etc.

Unless we expect notifications to happen frequently, I feel like a polling strategy could result in a lot of wasted round trips to the server. Polling is great when you expect the data to change frequently and does not need to be in real time. But, when the changes happen fairly infrequently but would benefit from real time updates (or need to be in real time) subscription can be more efficient since it is an open connection that is only used as needed and the resources used to keep that connection open are typically much lower than a request actually being processed. 

Now, this is obviously for examples sake but something like notifications are exactly when we should be using subscriptions (so we don't have all active users hitting the server every X seconds just to see if a notification is there or not) because most of the time, the answer will be "no new data" and we will have wasted a trip and unnecessarily used more resources on the server.

petertgiles: Yeah, I've never used graphql subscriptions and don't know that much about them.  When I started researching I saw the page that Vachan referenced and figured that notifications don't really qualify as "Small, incremental changes to large objects" or "Low-latency, real-time updates" so I switched back to polling.  5 seconds is probably excessive for prod, we could get away with several minutes, probably.

It's interesting though, and probably would be worth a spike to see how hard it would be to set up and what kind of gains we could get.  On [the Laravel broadcasting page](https://laravel.com/docs/10.x/broadcasting) they talk about integrating with a service called Pusher Channels.  Maybe there's a websockets service in Azure we could set up?  Or laravel-websockets?  I'm skeptical that our notifications have enough of a realtime requirement to make all this worth it but I'm not against someone doing some exploration for a possible future enhancement.  I popped up #8760 if you want to fill in anything else.

esizer: Yeah, I guess it really depends on how "real time" we want notifications to be. I guess our notifications are not necessarily critical so we could have them update every few mins (my past use cases were utility outage maps providing real-time updates to service disruptions which was far more critical than application updates 😛 ). 

Your suggestion of using several minutes for polling is typically what I see and what I would recommend we do when we actually implement it, since it can get out of hand really quickly as the number of active users increases.

Polling does make sense to start with though since it is simpler, we do not need a service to support it and we can get our feet wet and bring in subscriptions when we feel we need it. I just wanted to make sure we weren't totally ruling out subscriptions and using polling as a "one size fits all" solution 😅

api/config/queue.php
petertgiles, line 1: This file sourced from https://github.com/laravel/laravel/blob/10.x/config/queue.php

api/database/migrations/2023_12_01_195253_create_jobs_table.php
petertgiles, line 1: php artisan queue:table

api/database/migrations/2023_12_01_202238_create_failed_jobs_table.php
petertgiles, line 1: php artisan queue:failed-table

infrastructure/bin/setup_supervisor.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+# This script is run during the deployment to set up the supervisor service
+
+cp /home/site/wwwroot/infrastructure/conf/supervisord.conf /etc/supervisor/
+cp /home/site/wwwroot/infrastructure/conf/laravel-worker.conf /etc/supervisor/conf.d/
+supervisord -c /etc/supervisor/supervisord.conf
```
petertgiles, line 6: The workers are set as autostart so this launches them.  Run `supervisorctl status` to check on them.

infrastructure/conf/supervisord.conf
petertgiles, line 1: Sourced from `echo_supervisord_conf`.
http://supervisord.org/installing.html#creating-a-configuration-file

esizer: This is just the sample file mentioned in the link, correct? We don't actually use it, it is just for reference?

esizer (commented):
Wow! 🤯  This is works exactly as I'd expect it to, great work! 🎉 

I'd honestly say we could use much of the infrastructure/config/mutations from this. 

I'd like to see if we can get those in and move our CSV and print functions to this sooner rather than later 😄 Those seem to be one of the larger pain points on admin right now 😬

infrastructure/conf/supervisord.conf
petertgiles: It does actually get used.  The only modification I made was to add the [include] section at the end so that it can load the laravel-worker.conf file, which is totally custom.

esizer: Ohh okay. Thanks for that 👍 I thought it was because I saw it getting copied over but I also saw the big sample file comment so wanted to confirm 😛

vd1992:
Demo jobs fail and the worker queue is killed for operations > 60 seconds
I am assuming that is a timeout we can adjust, hopefully the queue being killed upon a failed job locally doesn't mean much

petertgiles:
> I am assuming that is a timeout we can adjust, hopefully the queue being killed upon a failed job locally doesn't mean much

Good question!  The workers are kicked off based on the options saved in the laravel-worker.conf file.  The timeout by default is 60 seconds [but can be adjusted](https://laravel.com/docs/10.x/queues#worker-timeouts).  If a job is killed it gets sent to the failed jobs table and can be retried, if configured.  If we're worried about small jobs getting starved out by big jobs we could put them in different queues with different workers with different timeouts.

apps/web/src/components/Layout/NotificationBar.tsx
tristan-orourke: So far, nothing we're considering Notifications for requires real-time updates. For things like updates to application status, submission reminders, or notifications of new requests (for admins), polling every few minutes would be fine.

However, that changes if we try to leverage the Notification system for other async operations like batch downloads, like this proof-of-concept suggests. If I submit a request for a large batch download, and it's ready in a handful of seconds, I'd want to know that as soon as possible. Maybe submitting a request like this can turn on some "waiting mode" in the client, causing it to pool more frequently?

I'm also a bit concerned about whether combining these async operations with the Notification system will confuse use of notifications a bit. Like, if they have different requirements when it comes to frequency of polling, will they have different needs in other ways too? As another example, we'll probably want to purge the download-files we've prepared from the server fairly frequently, whereas most notifications could be kept in the database for a long time.

petertgiles: Hmm, these are good points.  From just the "async operation" context, doing something other than a notification at the end of the job is easy enough.  And from the frontend perspective, making just one thing quickly poll is also easy.   But the "make a file available" thing is a whole other can of worms.  Do we store the file in the app service?  Or put it on a... whatever the Azure equivalent of an S3 bucket?  How do we ensure that only one person can download that file?  What is the retention period for files?

tristan-orourke: Oh, those are good questions about the file.

We _could_ store files in the /temp folder of the app service. There is also some kind of blob storage service IMTD mentioned but we never set up.

Laravel has utilities to help with [downloading files](https://laravel.com/docs/10.x/filesystem#downloading-files), but there are different ways we could use them. We could generate a temporary link to the file, which would expire quickly and be impossible to guess, but allow for anyone to access the file if the original user shares the link in the short timeframe. We could return a file response from a controller which does check for user authorization if we set up a database table which associates files with the users allowed to access them. Either way, we need to delete the files soon after (these kinds of generate-for-download files would have very low retention periods. 1 hour? 5 minutes?)

And also, maybe we don't need to go this route at all. [This post](https://laracasts.com/discuss/channels/laravel/using-responsedownload-for-large-files) has some links to help with streaming downloads from laravel. [This stackoverflow](https://stackoverflow.com/questions/15942497/why-dont-large-files-download-easily-in-laravel/15945626#15945626) post shows how to build a file stream response. Maybe we can do this synchronously, if we can generate the response in chunks and stream it back as we go?

petertgiles: Cool, this sounds like an interesting problem.  I'd be happy to expand the scope of this and try it out or spin it off into its own issue.

tristan-orourke: Yeah, this PR is a good blueprint for setting up queued jobs, but I think we need another spike for large file downloads, which was one of the original motivators behind this.

petertgiles: #8852

PR #8759, opened 2023-12-08
[Bug] Pool candidates table filter dialog and results out of sync

### LINKED ISSUE #8755: 🐛 [Found in Prod]Pool candidate table loads the candidates without the default filter
## 🐛 Bug
Pool candidates table loads the candidates without the default filter "Candidacy Status" AND "Expiry Status"
When you click on the Filters , Filter dialog opens up with these filters set. But initial query doesn't respect this filter. 

## 🦋 Expected Behaviour
Pool candidate table 

## 🕵️ Details

This bug caught in prod as suddenly new candidates appeared on the table as "New Application" for the pool where all the candidates are screened in already.

The query sets the flag properly when the filter dialog is opened and issued from there with any filters set or reset. 

## 📋 Steps to Reproduce

1. Go to any pool and click on view candidates link ( select CMO digital pool if you got local seeder run ) 
2. Observe the query `Candidacy Status` and `Expiry Status` is not in the query 
3. Click on the filters , set any filter and click "Show Results" 
4. Notice `Candidacy Status` and `Expiry Status` is in the query  
5. Clear the filter you set and click "show Results"
6. Notice `Candidacy Status` and `Expiry Status` filters are  still in the query  

## 🙋‍♀️ Proposed Solution

Set the default filter during the initial page load as well 

## ✅ Acceptance Criteria

- [ ] **Scenario 1: Pool candidate table loads candidates with default filter**
  - Given a pool candidate table
  - When the table loads
  - Then the default filter "Pool" , "Candidacy Status" and "Expiry Status" should be applied

- [ ] **Scenario 2: Filter dialog opens with default filters**
  - Given a pool candidate table
  - When the user clicks on the Filters
  - Then the filter dialog should open with the default filters  "Pool" , "Candidacy Status" and "Expiry Status" set

- [ ] **Scenario 4: Expired and Inactive Pool candidates should not appear in the table with default filters**
  - Given a pool candidate table
  - When candidates appear in the table
  - Then the candidates should not be in expired status or inactive status 

- [ ] **Scenario 5: Setting filters updates the query**
  - Given a pool candidate table
  - When the user clicks on the filters, sets any filter, and clicks "Show Results"
  - Then the query should always include the  default filters "Pool", "Candidacy Status" and "Expiry Status"

- [ ] **Scenario 6: Clearing filters updates the query**
  - Given a pool candidate table with filters set
  - When the user clears the set filters  above and clicks "Show Results"
  - Then the query should clear the filters but keep the default filters "Pool", "Candidacy Status" and "Expiry Status"

vd1992:
🤖 Resolves #8755  

## 👋 Introduction

Solve the discrepancy between the dialog and what actually happens filter-wise. 

## 🕵️ Details

The fallback condition in `transformPoolCandidateSearchInputToFormValues` checks in the desired state and this appears in the filter dialog, however, this is after filtering occurs and is not the earliest point in the flow of page/API. 

Fallback 

image

Adjusted by having the desired defaults fed in at the earliest state, the component props. 
Noticed this presented an issue for view candidates for a pool, but as well as the all candidates, and view search request pages.

## 🧪 Testing

1. Navigate to the view candidates page for a pool, observe the filter dialog and the submitted API operation
2. Repeat for all candidates table
3. Repeat for viewing a single search request
4. Ensure filter submits works as normal for all three

## 📸 Screenshot

View candidates

image

image

All candidates

image

image

For requests, adding onto fetched filters

image

image

### DIFF
--- a/apps/web/src/pages/PoolCandidates/AllPoolCandidatesPage/AllPoolCandidatesPage.tsx
+++ b/apps/web/src/pages/PoolCandidates/AllPoolCandidatesPage/AllPoolCandidatesPage.tsx
@@ -8,6 +8,10 @@ import SEO from "~/components/SEO/SEO";
 import AdminContentWrapper from "~/components/AdminContentWrapper/AdminContentWrapper";
 import adminMessages from "~/messages/adminMessages";
 import PageHeader from "~/components/PageHeader";
+import {
+  CandidateExpiryFilter,
+  CandidateSuspendedFilter,
+} from "~/api/generated";
 
 export const AllPoolCandidatesPage = () => {
   const intl = useIntl();
@@ -38,7 +42,13 @@ export const AllPoolCandidatesPage = () => {
     <AdminContentWrapper crumbs={navigationCrumbs}>
       <SEO title={pageTitle} />
       <PageHeader icon={IdentificationIcon}>{pageTitle}</PageHeader>
-      <PoolCandidatesTable title={pageTitle} />
+      <PoolCandidatesTable
+        title={pageTitle}
+        initialFilterInput={{
+          suspendedStatus: CandidateSuspendedFilter.Active,
+          expiryStatus: CandidateExpiryFilter.Active,
+        }}
+      />
     </AdminContentWrapper>
   );
 };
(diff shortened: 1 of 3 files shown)


esizer (approved):
Nice! 👏

PR #8761, opened 2023-12-11
[Tooling] Enforce GitHub workflow naming convention

### LINKED ISSUE #8310: 🛠️ Establish GitHub workflows naming convention
## 🛠️ Tooling

GitHub workflows do not follow a standard naming convention.

## 🕵️ Details

We have some that include a verb, the word test (aren't they all tests?), the word run (seems repetitive). Also, there are two jobs with the same name.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/232da2865854fb4fb2841a3220eff61c4ad49b0e/.github/workflows/lint.yml#L20

https://github.com/GCTC-NTGC/gc-digital-talent/blob/232da2865854fb4fb2841a3220eff61c4ad49b0e/.github/workflows/translations.yml#L18

## 🙋‍♀️ Proposed Solution

- filename should be `name.yml`
- workflow should be `Name`
- job should be `name`

## ✅ Acceptance Criteria

- [ ] GitHub workflows follow naming convention for filename, workflow names, and job names

mnigh:
🤖 Resolves #8310.

## 👋 Introduction

This PR creates and enforces a GitHub workflow naming convention as well as updates related (and some unrelated) file path references.

## 🧪 Testing

1. Verify naming convention followed for workflow names, jobs,


.github/workflows/codeql-analysis.yml
```diff
@@ -34,10 +23,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: ["javascript"]
-        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python' ]
+        language: ["javascript-typescript"]
```
vd1992, line 26: Think this is the source of this
https://github.com/GCTC-NTGC/gc-digital-talent/pull/8761/checks?check_run_id=19522047268

mnigh: Yes, this is a weird one because some of the old workflows will need to be disabled once this PR is merged. I am not exactly sure why but it might be that the CodeQL workflow only works when committed directly to the `main` branch.

> Warning: Code scanning cannot determine the alerts introduced by this pull request, because 1 configuration present on refs/heads/main was not found.

mnigh: this also might need one of the repo @GCTC-NTGC/admin to change any settings: https://docs.github.com/en/code-security/code-scanning/enabling-code-scanning/configuring-default-setup-for-code-scanning#configuring-default-setup-for-a-repository

vd1992 (approved):
Naming looks fine to me, all tests still run. Probably need to adjust the enabled workflows post merging?

tristan-orourke (approved):
The naming looks good to me too.

PR #8762, opened 2023-12-11
[Feature] Allow overriding feature flags in cypress

### LINKED ISSUE #4227: Research Cypress for e2e testing with feature flags ON and OFF

esizer:
🤖 Resolves #4227 

## 👋 Introduction

This adds a command that allows us to override our feature flags.

> [!NOTE]
> I was able to get a working example so I posted this PR. Though, I imagine there are other ways of accomplishing this and it does not need to result in a merge.

## 🕵️ Details

The approach taken is to intercept and override the `config.js` file on our server. We do this by:

1. Reading the `.env` file in `apps/web`.
2. Spreading any overridden flags (written as an object)
3. Then, we loop over it using the [`Map.prototype.set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/set) method to create the new config
4. Then, we return the new, custom `config.js` file

## 🛠️ Usage

It is pretty easy to do this, but, similar to aliasing requests, it matters where you use it. It is best to call this _before_ you visit a specific page.

```ts
// Override for all tests in a spec
describe("Override for all assertions", () => {
  beforeEach(() => {
    cy.overrideFeatureFlags({ FEATURE_FLAG: null });
  });
});

// Override for a specific assertion
describe("Override for specific assertions", () => {
  it("Is overridden", () => {
    cy.overrideFeatureFlags({ FEATURE_FLAG: null });
    cy.visit("/");
  });
  it("Is is default", () => {
    cy.visit("/");
  });
});
```

## 🧪 Testing

> [!TIP]
> I found using the `FEATURE_RECORD_OF_DECISION` on the `pool-candidates.cy.ts` is very noticeable to see it working.

### All asserations

1. Open a test you know is using a feature flag
2. Add the override to `beforeEach`
3. Confirm all assertions read the new value properly

### Specific assertion

1. Open a test you know is using a feature flag
2. Add the override to the top of a specific assertion
3. Confirm only that assertion uses the value

### DIFF
--- a/apps/e2e/cypress/support/featureFlagCommands.ts
+++ b/apps/e2e/cypress/support/featureFlagCommands.ts
@@ -1,0 +1,21 @@
+import * as dotenv from "dotenv";
+
+dotenv.config({ path: "../web/.env" });
+
+const overrideFeatureFlags = (flags: Object) => {
+  const env = {
+    ...process.env,
+    ...flags,
+  };
+
+  let map = `const data = new Map();`;
+  Object.keys(env).forEach((key) => {
+    map = `data.set(${key}, ${env[key]});`;
+  });
+
+  return `${map} window.__SERVER_CONFIG__ = data`;
+};
+
+Cypress.Commands.add("overrideFeatureFlags", (flags) => {
+  cy.intercept("GET", "/config.js", overrideFeatureFlags(flags));
+});
--- a/apps/e2e/cypress/support/featureFlagCommands.ts
+++ b/apps/e2e/cypress/support/featureFlagCommands.ts
@@ -10,7 +10,7 @@ const overrideFeatureFlags = (flags: Object) => {
 
   let map = `const data = new Map();`;
   Object.keys(env).forEach((key) => {
-    map = `data.set(${key}, ${env[key]});`;
+    map = `${map} data.set(${key}, ${env[key]});`;
   });
 
   return `${map} window.__SERVER_CONFIG__ = data`;
--- a/documentation/environment-variables.md
+++ b/documentation/environment-variables.md
@@ -20,3 +20,33 @@ For deployment in production, there needs to be a way to change variables in the
 
 To check what variables have been set in the app, open the console of your browser and enter:
 `window.__SERVER_CONFIG__`
+
+#### End-to-End Testing
+
+Since run-time variables are used to maintain the current feature set and can be changed, tests should be run against both versions of a specific feature set. In order to do this, the "feature flags" must be overridden in the test specifications.
+
+When writing test specifications, run-time variables can be modified for the entire spec, or specific tests using the `cy.overrideFeatureFlags(flags: Object): void` command.
+
+##### Override for entire spec
+
+```tsx
+describe("Override for all tests", () => {
+  beforeEach(() => {
+    cy.overrideFeatureFlags({ FEATURE_FLAG: null });
+  });
+});
+```
+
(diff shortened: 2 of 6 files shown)


petertgiles (approved):
Oh wow, this is so smart and works so well.  :star:

documentation/environment-variables.md
```diff
@@ -20,3 +20,33 @@ For deployment in production, there needs to be a way to change variables in the
 
 To check what variables have been set in the app, open the console of your browser and enter:
 `window.__SERVER_CONFIG__`
+
+#### End-to-End Testing
+
+Since run-time variables are used to maintain the current feature set and can be changed, tests should be run against both versions of a specific feature set. In order to do this, the "feature flags" must be overridden in the test specifications.
+
+When writing test specifications, run-time variables can be modified for the entire spec, or specific tests using the `cy.overrideFeatureFlags(flags: Object): void` command.
```
mnigh, line 28:
```suggestion
When writing test specifications, run-time variables can be modified for the entire spec, or for specific tests using the `cy.overrideFeatureFlags(flags: Object): void` command.
```

documentation/environment-variables.md
```diff
@@ -20,3 +20,33 @@ For deployment in production, there needs to be a way to change variables in the
 
 To check what variables have been set in the app, open the console of your browser and enter:
 `window.__SERVER_CONFIG__`
+
+#### End-to-End Testing
+
+Since run-time variables are used to maintain the current feature set and can be changed, tests should be run against both versions of a specific feature set. In order to do this, the "feature flags" must be overridden in the test specifications.
```
mnigh, line 26: any specific reason to have feature flags in quotes?
```suggestion
Since run-time variables are used to maintain the current feature set and can be changed, tests should be run against both versions of a specific feature set. In order to do this, the feature flags must be overridden in the test specifications.
```

mnigh (commented):
⭐ Awesome stuff! A few suggestions around copy but the documentation is very readable.

mnigh (approved):

apps/e2e/cypress/support/featureFlagCommands.ts
```diff
@@ -0,0 +1,21 @@
+import * as dotenv from "dotenv";
+
+dotenv.config({ path: "../web/.env" });
+
+const overrideFeatureFlags = (flags: Object) => {
```
tristan-orourke, line 5: Sorry, I didn't find time to review this before it was merged, but I'll leave this comment as a suggestion for next time...

@esizer My understanding is that we should basically never use `Object` in a typescript definition. I think a better typing for this would be `flags: Record<string, null | boolean>`.

esizer: Oh yeah, good point thanks for flagging this! 🙇‍♀️ . I totally just had this while testing (wasn't sure exactly what types we could accept) and forgot to go back and adjust to a more strict type.

esizer: Added an issue so this doesn't get buried and forgotten.

 - https://github.com/GCTC-NTGC/gc-digital-talent/issues/8780

PR #8771, opened 2023-12-12
Poster Deadline timezone fix

### LINKED ISSUE #8770: 🐛 Job Poster "apply on or before" date is misleading
## 🐛 Bug

In Pacific timezone, the date is fine. It says "Apply on or before Dec 12" if the poster closes at 11:59pm Dec 12, Pacific time. But the same poster, viewed in Eastern timezone, will say "Apply on or before Dec 13", since the poster closes at 2:59am Dec 13 local time.

This is misleading for users who believe they can come back any time on the day of Dec 13 to finish their application.

## 🦋 Expected Behaviour

The "apply on or before" should always show the date of the last full day the poster is open, ie the date according to Pacific Time.

## 📸 Screenshot

## 🙋‍♀️ Proposed Solution

image

image

## ✅ Acceptance Criteria

- [ ] The "Apply on or before" date always shows the date according to Pacific Time.

tristan-orourke:
🤖 Resolves #8770 

## 👋 Introduction

Ensures that the "Apply on or before date" is the same in all timezones - it is formatted as belonging to Pacific timezone.

## 🕵️ Details

This is to avoid a situation where it said "Apply on or before Dec 14" but the poster actually closes at 3am on Dec 14. The solution is just to format the date as it would appear in Pacific timezone, instead of local timezone.

## 🧪 Testing

1. Edit the database to ensure you have a job poster in your local environment which closes at 07:59:00 UTC on some day. (Or you can create a new job poster via admin portal.)
2. Look at the Job Poster. The "Apply on or before date" should match the Pacific time date which appears in the detailed modal.
3. The same date should appear on the Browse Jobs card, and on the "Track your applications" card if you start applying to it.
4. The same date should appear on the Process Information page on the admin portal.

## 📸 Screenshot

Main deadline display:
image

Detailed modal:
image

Admin portal process information:
image

### DIFF
--- a/apps/web/src/components/PoolCard/PoolCard.tsx
+++ b/apps/web/src/components/PoolCard/PoolCard.tsx
@@ -146,6 +146,7 @@ const PoolCard = ({ pool, headingLevel = "h3" }: PoolCardProps) => {
                       date: parseDateTimeUtc(pool.closingDate),
                       formatString: "PPP",
                       intl,
+                      timeZone: "Canada/Pacific",
                     }),
                   },
                 )
--- a/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx
+++ b/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx
@@ -422,6 +422,7 @@ export const PoolPoster = ({
                               date: parseDateTimeUtc(pool.closingDate),
                               formatString: "PPP",
                               intl,
+                              timeZone: "Canada/Pacific",
                             }),
                           },
                         )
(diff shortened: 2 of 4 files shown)


tristan-orourke:
So, regarding questions about how this looks outside of Canada...

In Japan, Josh's logic of "we should show the date of the last full 24 hour period" still holds true: 
image

The only way you could get into trouble is if you're _west_ of the Canadian/Pacific timezone; basically [Alaska or an island in the Pacific](https://www.timeanddate.com/time/map/). For example, if you're applying from Hawaii, and you assume that "Apply on or before" implies a midnight deadline, you'd be off by 2 hours:
image

However, I'm okay with this because:
- If we're only showing a date here and not a time, being off by 2 hours is a lot closer than being off by a full 22
- Our detail modal is still there with the exact details and reasoning behind the deadline
- We **do** cater to Canadians in particular, and we call that out in the detail modal

vd1992 (approved):
Working through the submit application workflow and it appears consistent to me
👍

PR #8774, opened 2023-12-12
[Feature] Add open graph image

### LINKED ISSUE #8614: ✨ New link previews
## ✨ Feature

When we share a link to our platform, currently there is no image that goes with it. This issue will fix that!

## 🕵️ Details

This is what we do not want:
image

## 🎨 Design File
These are the images to use (in two different sizes). Separate images are given for IAP and GC Digital talent... hope that's ok.

GCDT

Generic Facebook EN
Generic Facebook FR
Generic LinedIn FR
Generic LinkedIn EN

IAP 

Generic Facebook EN IAP
Generic Facebook FR IAP
Generic LinedIn FR IAP
Generic LinkedIn EN IAP

### 🧑‍🎨 Designer

The Brookster (@BrookeCameron)

We should be following the standards set out by Open Graph: https://ogp.me/

## ✅ Acceptance Criteria

- [ ] Image shows up when a link is shared.
- [ ] Different ones for IAP
- [ ] Titles show in share previews
- [ ] Subtitles show where appropriate in share previews

Start with main landing page, basic implementation. Additional variants to come later
Test with UAT, opened to public?

esizer:
🤖 Resolves #8614 

## 👋 Introduction

Adds an image for our open graph meta data. Also, fixes the webpack config for meta data.

## 🧪 Testing

> [!Note]
> This one will be tricky to properly test and we may need to do some odd dev deploy to properly test.
> In order to resolve the image, we must use a fully qualified domain name. Currently it is set to `https://talent.canada.ca/images/digital-talent/banner.jpg`. When testing you will need to change this value to point it to local host, or the dev instance.

If you'd like to test, I used the [OGraph Previewer](https://chromewebstore.google.com/detail/ograph-previewer/ggcfeakcnodgcmmllfdbmngekljbhiim) extension for chrome (unaware of other browser extensions that work 😢 )

1. Update the `packages/webpack-config/webpack.base.js` pointin the `meta.image` value to `http://localhost:8000/images/digital-talent/banner.jpg`
2. Do a fresh build `npm run dev:fresh`
3. Navigate to any page on the application
4. Use the extension to preview the results

## 📸 Screenshot

Screenshot 2023-12-12 154639


petertgiles (changes requested):
This works very well!  :clap: 

I don't love the hardcoded URL though.  Could we use an environment variable to build it?  In prod there is a variable called APP_URL with the value of "https://talent.canada.ca".  Perhaps you could set that up to use that locally?  This would allow it to be tested on UAT as well.

Also, in light of reducing bugs, is there any way a cypress test could verify this functionality works in the future?

esizer:
> I don't love the hardcoded URL though. Could we use an environment variable to build it? In prod there is a variable called APP_URL with the value of "https://talent.canada.ca". Perhaps you could set that up to use that locally? This would allow it to be tested on UAT as well.

Oh, thanks! That makes sense since laravel uses it 🤦‍♀️ I'll look into this and love the suggestion 👍 

> Also, in light of reducing bugs, is there any way a cypress test could verify this functionality works in the future?

Yeah, this sounds doable. Im sure it is as easy as querying for the meta tag and make sure it has the expected value 👍

petertgiles:
> Okay @petertgiles , I went ahead and applied your suggestion but took it a bit further. I moved all the meta config that had hardcoded values to use env vars with the previously hardcoded values as fallbacks (since I don't think we have them).

Nice, looks great!  One last ask, could you add them to the .env.example file?  At least just the keys commented out so that they're documented in a logic place.

esizer:
> Nice, looks great! One last ask, could you add them to the .env.example file? At least just the keys commented out so that they're documented in a logic place.

Great idea, added in [add new env to example file](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8774/commits/0af5502ca4e2d43a57b3925a179a9fa9ec192998)

petertgiles (approved):

PR #8775, opened 2023-12-12
[Bug] Application page reloads

### LINKED ISSUE #8357: 🐛 When linking experience to skills during an application, the page reloads (and scrolls) each time
## 🐛 Bug

During the skill requirements step of the application process, the user has to link experiences to their skills. Each time this happens, the dialog pops up, the user selects and describes the skill relationship, and then submits the addition. Annoyingly though, this causes the page to reload, scrolling the user back up to the top of the page (the skill is linked correctly). 

## 🦋 Expected Behaviour

Submitting the dialog to link the experience shouldn't reload the page, or if it must, it should at least scroll the user back to their last location. This is both a general usability and accessibility bug, because assistive tech users are forced to navigate back down to their last location over and over and over.

## 🕵️ Details

- **Operating System**: Windows 11
- **Browser**: Firefox

## 📋 Steps to Reproduce

1. Get to the skill requirements step in an application
2. Link a skill and experience
3. Scroll back down over and over

yonikid15:
🤖 Resolves #8357 

## 👋 Introduction

- Adds `useApplication` hook (with [useOutletContext](https://reactrouter.com/en/6.20.1/hooks/use-outlet-context)) to provide application data to child routes, from parent `<ApplicationLayout />`.
-  Removes unneeded application queries from child application pages.
- Refactors the rendering logic in the `<ApplicationLayout />` so that the main content doesn't re-render when the user executes a mutation. This change stops the page scrolling the user to the top and losing their tab location. However, the opacity of the loading icon needs to be stronger to hide any flash of errors.

## 🧪 Testing

1. Login as admin or applicant.
2. Go to browse jobs and apply to any job.
3. Go through application process and ensure everything still works correctly.
4. Ensure on the skills step that the user is returned to their scroll location and tab location after adding/editing/removing an experienceSkill.


brindasasi (commented):
I like the refactoring work! Looking concise.. 
Verified experience-skill connection step doesn't reload as well as the successful application submission.
Everything works fine. 
There are still test failures though.

vd1992:
Hopefully you can sort out your environment soon enough 😛 
But in the meanwhile to help, from what I can tell, the test appears to be catching on the following. 

- Navigate to `/en/applications/:id/skills` , so the skills step of applying
- Hit "save and continue", cannot continue due to incomplete page and an error alert appears at the bottom
- Complete the page, error alert does not disappear but if you hit "save and continue" it allows you to go through

I am guessing with the changes, updating an experience no longer is recalculating the value of `isSkillsExperiencesIncomplete` automatically. It holds the stale value unless you refresh the page

image

esizer:
> I am guessing with the changes, updating an experience no longer is recalculating the value of `isSkillsExperiencesIncomplete` automatically. It holds the stale value unless you refresh the page after updating the experiences to be all complete.

Yup, looks like this is what it is! That check only happens on clicking "Save and continue" but since @yonikid15 fixed the page reload issue the page never re-renders so the alert stays until you click the button again 😢 I think we have two options,

1. Bubble up a callback from the dialog (annoying because it is so far down the component tree) and set the error value in said callback, more work but better UX
2. Update the test to move the assertion to after we try to save and continue, easier but not great UX

### Option 1

```tsx
// ApplicationSkillsPage.tsx:256
<SkillTree
  key={optionalTechnicalSkill.id}
  skill={optionalTechnicalSkill}
  experiences={experiences}
  onConnectExperience={() => {
     setValue(
      "skillsMissingExperiences",
      isSkillsExperiencesIncomplete ? 1 : 0,
    );
  }}
/>
```

### Option 2

```ts
// submit-application.cy.ts:339
cy.findByRole("button", { name: /Save and continue/i }).click();
cy.wait("@gqlUpdateApplicationMutation");
cy.contains(
  /Please connect at least one career timeline experience to each required technical skill./i,
).should("not.exist"); // Experience and skill linked
cy.expectToast(/Successfully updated your skills!/i);
```

vd1992:
Not the best UX because as far as I can tell, in other forms, when you trigger a form validation thing the error alert goes away automatically when you've fixed it. No need to try submitting to see if it passes validation. 
So this gets a user accustomed to error alerts going away automatically once they have fixed things.

esizer:
> Not the best UX because as far as I can tell, in other forms, when you trigger a form validation thing the error alert goes away automatically when you've fixed it. No need to try submitting to see if it passes validation.
> So this gets a user accustomed to error alerts going away automatically once they have fixed things.

So I assume this is a vote in favour of option 1 then?

vd1992:
> > Not the best UX because as far as I can tell, in other forms, when you trigger a form validation thing the error alert goes away automatically when you've fixed it. No need to try submitting to see if it passes validation.
> > So this gets a user accustomed to error alerts going away automatically once they have fixed things.
> 
> So I assume this is a vote in favour of option 1 then?

Ideally yeah, unless it is really unwieldy 😕

yonikid15:
Wow y'all are awesome, thanks for the help. I'm up and running now, so I can add it in unless you already started?

vd1992:
> Wow y'all are awesome, thanks for the help. I'm up and running now, so I can add it in unless you already started?

Go on ahead, I just wanted to help push things along while you were stuck sorting out environment troubles 😃

esizer:
> Go on ahead, I just wanted to help push things along while you were stuck sorting out environment troubles 😃

Okay, we weren't able to get the `onConnect` thing working because the cache was updating asynchronously. So, instead we added a `useEffect` to update the value and revalidate. Should be working now 👍 

[update and revalidate on application change](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8775/commits/ac97ded5f373847fde14537a75e20468ea28222d)

vd1992 (approved):
Not sure if Brinda wants to add anything, but exercising the workflow in dev and build conditions and it appears to work without issue.
Great!

PR #8784, opened 2023-12-13
[Fix] Admin table filter dialog count and clearing

### LINKED ISSUE #8776: 🐛 Inconsistency in table filter modal
## 🐛 Bug

Unexpected behaviour when closing the table filter modal without clicking "show results".

## 🦋 Expected Behaviour

There are three ways to close the modal: clicking the "Show results" button, clicking the "Clear filters", and exiting the modal in some other way like clicking outside the modal or with the escape key.

### Show results
If I open the filter, change some filters, and click "Show results", then it works as expected. Three things happen:
  1. The modal closes
  2. The table updates to show the result of searching
  3. The filter button updates to show the number of active filters

### Clear filters
If I open the modal, change some filters, and click "Clear filters", then:
  1. The modal closes
  2. No new search is run, the table is unchanged
  3. The modal forgets about any filter changes I made. The number of active filters remains unchanged, and if I reopen the model, the previous filter changes I made have disappeared.

Our previous behaviour for "clear filters" was to clear all active filters and then rerun the search. This new behaviour would make sense to me, **if** we change the text from "clear filters" to "cancel".

### Other
If I open the modal, change some filters, and close the modal with the escape key, then:
  1. The modal closes
  2. No new search is run, the table is unchanged
  3. The filter button updates to show the number of active filters I left in the modal. **This no longer corresponds to the active filters reflected in the table results.**

It doesn't make sense to me that the number of active filters shown on the button or in the modal no longer reflects the "active filters" determining the results.

## 📋 Steps to Reproduce

See above

## 📸 Screenshot

Table with filter button with number of active filters:
image

"Clear filters" and "Show results" buttons on filter modal:
image

## 🙋‍♀️ Proposed Solution

a) Change "Clear filters" text to "Cancel", and ensure that exiting the modal without clicking "Show results" always triggers the same behaviour as clicking "Cancel"

b) Change "Clear filters" behaviour to remove any active filters, reflect that in the UI (the modal inputs and the number on the button) then re-run the search. Closing the modal in other ways should trigger the "cancel" behaviour currently triggered by the "clear filters" button.

## ✅ Acceptance Criteria

- [ ] Clear filters button restores default, and reruns search
- [ ] Closing the filter in other ways discards any staged filters
- [ ] count only updates when filters are submitted, when closed it should reset to previous number

esizer:
🤖 Resolves #8776 

## 👋 Introduction

This fixes the filter count for the dialog to only display the number of filters actually in use.

### Clear functionality

It appears as though the functionality we want on the clear button is not easily possible. This includes a solution but has been disabled while we revisit it. While it does clear the filters, we are unsure if this is actually what we want based on how we actually use the filters.

The odd stuff comes up when we indiscriminately clear filters, it affects the UI in an unexpected way. This is most notable on radio buttons where the true "empty" value is not empty. 

The current ideas are:

1. Force devs to pass in the expected "empty" value for specific inputs so the generic form is aware of them
2. Override the value for things like radio buttons to be an empty string (this will required revisiting the transforms)
3. Keep the existing functionality and just rename the button to reflect it (ie. "Reset filters")

#### Answer

REF: https://github.com/GCTC-NTGC/gc-digital-talent/issues/8776#issuecomment-1854523405

It seems we opted to maintain existing functionality (option 3) and just update the button text to better represent the function. Please be aware of this when testing.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to a table that includes the dialog
3. Apply some filters and submit
5. Confirm the number updates as expected
6. Open the dialog and use the "reset filters"
7. Confirm the number resets to the initial value
8. Open the dialog and set some filters
9. Close the dialog without submitting (X or <kbd>esc</kbd>)
10. Confirm the number remains constant

### DIFF
--- a/apps/web/src/components/FilterDialog/FilterDialog.tsx
+++ b/apps/web/src/components/FilterDialog/FilterDialog.tsx
@@ -15,17 +15,17 @@ import { notEmpty } from "@gc-digital-talent/helpers";
 // Used by specific dialogs
 export type CommonFilterDialogProps<TFieldValues extends FieldValues> = {
   onSubmit: SubmitHandler<TFieldValues>;
-  /** Defaults values, including from URL */
-  routeValues?: Partial<TFieldValues>;
-  /** Initial values to reset the form to, not including URL, */
-  initialValues: TFieldValues;
+  /** When the user resets filters they will return to these values. If initialValues is empty, resetValues is used to initialize the filters. */
+  resetValues: TFieldValues;
+  /** If initialValues is set, it will override resetValues when the filter form is first initialized. */
+  initialValues?: Partial<TFieldValues>;
 };
 
 type FilterDialogProps<TFieldValues extends FieldValues> = {
   onSubmit: CommonFilterDialogProps<TFieldValues>["onSubmit"];
   options?: UseFormProps<TFieldValues, unknown>;
   // Values to reset to (removing URL state)
-  initialValues: CommonFilterDialogProps<TFieldValues>["initialValues"];
+  resetValues: CommonFilterDialogProps<TFieldValues>["resetValues"];
   defaultOpen?: boolean;
   children: React.ReactNode;
 };
@@ -34,7 +34,7 @@ const FilterDialog = <TFieldValues extends FieldValues>({
   onSubmit,
   options,
   children,
-  initialValues,
+  resetValues,
   defaultOpen = false,
 }: FilterDialogProps<TFieldValues>) => {
   const [isOpen, setIsOpen] = React.useState<boolean>(defaultOpen);
@@ -69,8 +69,8 @@ const FilterDialog = <TFieldValues extends FieldValues>({
 
   // Reset form and submit
   const handleClear = async () => {
-    reset(initialValues);
-    setActiveFilters(initialValues);
+    reset(resetValues);
+    setActiveFilters(resetValues);
     await methods.handleSubmit(onSubmit)();
     setIsOpen(false);
   };
--- a/apps/web/src/components/FilterDialog/FilterDialog.tsx
+++ b/apps/web/src/components/FilterDialog/FilterDialog.tsx
@@ -16,7 +16,7 @@ import { notEmpty } from "@gc-digital-talent/helpers";
 export type CommonFilterDialogProps<TFieldValues extends FieldValues> = {
   onSubmit: SubmitHandler<TFieldValues>;
   /** Defaults values, including from URL */
-  defaultValues?: Partial<TFieldValues>;
+  routeValues?: Partial<TFieldValues>;
   /** Initial values to reset the form to, not including URL, */
   initialValues: TFieldValues;
 };
--- a/apps/web/src/components/FilterDialog/FilterDialog.tsx
+++ b/apps/web/src/components/FilterDialog/FilterDialog.tsx
@@ -72,7 +72,10 @@ const FilterDialog = <TFieldValues extends FieldValues>({
   // Reset the form with no submission on close
   const handleOpenChange = (newOpen: boolean) => {
     if (!newOpen) {
-      reset();
+      reset((currentValues) => ({
+        ...currentValues,
+        ...activeFilters,
+      }));
     }
 
     setIsOpen(newOpen);
--- a/apps/web/src/components/FilterDialog/FilterDialog.tsx
+++ b/apps/web/src/components/FilterDialog/FilterDialog.tsx
@@ -6,6 +6,7 @@ import {
   SubmitHandler,
   UseFormProps,
   useForm,
+  DefaultValues,
 } from "react-hook-form";
 import AdjustmentsVerticalIcon from "@heroicons/react/20/solid/AdjustmentsVerticalIcon";
 
@@ -39,10 +40,13 @@ const FilterDialog = <TFieldValues extends FieldValues>({
 }: FilterDialogProps<TFieldValues>) => {
   const [isOpen, setIsOpen] = React.useState<boolean>(defaultOpen);
   const intl = useIntl();
+  const defaultValues =
+    options?.defaultValues ??
+    (resetValues as DefaultValues<TFieldValues> | undefined);
   const methods = useForm({
     mode: "onSubmit",
     ...options,
-    defaultValues: options?.defaultValues,
+    defaultValues,
   });
   const {
     reset,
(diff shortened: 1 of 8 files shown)


apps/web/src/components/FilterDialog/FilterDialog.tsx
vd1992, line 1: I don't entirely follow along with what the goal is now, but exercising this page I see the filter dialog and filter count get decoupled. 

- Go to page
- Open dialog, set a filter, submit
- Open dialog, close it by clicking to side, then open again
- Note filter count and filter dialog out of sync

https://recordit.co/1M2uHQrxV8

esizer: Oh, that is a good find! 🤔 I'll investigate 🕵️

esizer: Okay, this was a lot easier than I assumed because of a simple mistake I made 😅 I was resetting to the default not the last submitted filters 😢 

Fixed in [reset to last active filters on close filter dialog](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8784/commits/0ede565a58b62e7e6448523e59f7ae042602e29d)

yonikid15: The filter count is still not changing for me.


https://github.com/GCTC-NTGC/gc-digital-talent/assets/22059495/3f8190da-50cc-4653-ba2d-43a4fafc2047

esizer: I tried my best to not need this, but it looks like we do since the `defaultValues` change when the URL updates 😢 

So we need to pass in the values from the URL if they exist (`defaultValues`) and the values we want to reset to (`intialValues`). The names aren't great but I am at a loss. Suggestions would be appreciated 😛 

[pass in initial values to reset to](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8784/commits/80b6f19c01e668a9715f63f89d76ba2dda9bfcfd)

yonikid15: I can't think of a better name 😆. I think it should be fine.
Looks to be working as expected now though 👍

apps/web/src/components/FilterDialog/FilterDialog.tsx
```diff
@@ -15,12 +15,17 @@ import { notEmpty } from "@gc-digital-talent/helpers";
 // Used by specific dialogs
 export type CommonFilterDialogProps<TFieldValues extends FieldValues> = {
   onSubmit: SubmitHandler<TFieldValues>;
+  /** Defaults values, including from URL */
   defaultValues?: Partial<TFieldValues>;
+  /** Initial values to reset the form to, not including URL, */
+  initialValues: TFieldValues;
```
yonikid15, line 21: Hmm, just thought of `pickedValues` or `chosenValues`?

yonikid15 (approved):
Looks great nice work 🥇

apps/web/src/components/FilterDialog/FilterDialog.tsx
esizer: That makes more sense! Since it is technically the ones a user selected and not the ones we hard code as default. I'll quickly mention at stand up and see what people think.

apps/web/src/components/FilterDialog/FilterDialog.tsx
```diff
@@ -39,12 +45,14 @@ const FilterDialog = <TFieldValues extends FieldValues>({
     defaultValues: options?.defaultValues,
```
tristan-orourke, line 45: If `options?.defaultValues` is undefined, I think we should use `initialValues` here.

apps/web/src/components/FilterDialog/FilterDialog.tsx
tristan-orourke: Okay, now that I've looked at the code, I wouldn't recommend `routeValues` 😅 . It should make sense more generally - those values don't _need_ to come from the route. Funnily enough, after understanding what's going on, I think "initialValues" (for routeValues) and "defaultValues" (for initialValues) actually make a lot of sense, but they're not helpful to _learn_ what's going on... :(

How about trying to be really explicit about how these values are used, instead of where they come from:
```
  /** When the user resets filters they will return to these values. If initialValues is empty, resetValues is used to initialize the filters. */
  resetValues: TFieldValues;
  /** If initialValues is set, it will override resetValues when the filter form is first initialized. */
  initialValues?: Partial<TFieldValues>;
```
Alternatively, `overrideValues` instead of `initialValues`?

tristan-orourke (changes requested):
Okay, it passes my testing, however... I'd like to retract my previous naming suggestion 😅

apps/web/src/components/FilterDialog/FilterDialog.tsx
esizer: I will try but I think this was intentional because whatever is passed in here is what is used when you call `reset()`. Though, there have been some tweaks since then so maybe that is no longer the case 👍

apps/web/src/components/FilterDialog/FilterDialog.tsx
esizer: I liked that so I just copy pasted 😄 

Updated in [update filter state props for even more clarity](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8784/commits/e3fefe9464c9aef41195540dc84a3127e65fa51c)

apps/web/src/components/FilterDialog/FilterDialog.tsx
tristan-orourke: Ugh. I spent way too long on trying to understand why this is. I'm tempted to cast it - I'm _pretty_ sure it should work. Or we leave it as is, and change the docstring describing resetValues and initialValues.

esizer: Okay, I also don't see a huge issue with a cast here since the full value should cover a partial value 😅 

[update default values to fallback to resetValues](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8784/commits/43d05be5f58a5dedc2b134c5811827f075375814)

tristan-orourke: Even a `Partial<TFieldValues>` doesn't work, it seems to specifically want `DefaultValues<TFieldValues>`. I found this old post which seems related, but unhelpful: https://github.com/orgs/react-hook-form/discussions/3713. Casting it should be fine.

tristan-orourke (approved):
Thanks Eric!

PR #8786, opened 2023-12-14
[Bug] Duplicate `experience_user_skill_unique`

### LINKED ISSUE #8778: 🐛 Prod logged error - experience_user_skill_unique constraint
## 🐛 Bug

We're getting occasional errors logged like this:
> SQLSTATE[23505]: Unique violation: 7 ERROR: duplicate key value violates unique constraint "experience_user_skill_unique"

Our app shouldn't be requesting an action that would cause a database constraint violation.

Is this a situation with soft-deletes preventing new records?

## ✅ Acceptance Criteria

- [ ] Find out why the errors are happening
- [ ] Prevent it from happening in the future

vd1992:
🤖 Resolves #8778

## 👋 Introduction

Find and resolve the constraint violation

## 🕵️ Details

Given the production environment has limited ways to edit the skill-userskill-experience grouping, where to look was narrowed. 
Setting skill-library to false and exploring I could not replicate the matter except for in one method, attempting to sync the same skill twice in one edit.
The soft deletion and restore seemed without issue to me. 

Solved then by preventing you from selecting the same skill multiple times. 

## 🧪 Testing

1. Set skill library to false
2. Be unable to replicate a duplication error from the career timeline

## 📸 Screenshot

Off main

Pick the same skill twice

image

Observe console error

image

Update fails

image

### DIFF
--- a/apps/web/src/pages/Profile/ExperienceFormPage/components/ExperienceSkills.tsx
+++ b/apps/web/src/pages/Profile/ExperienceFormPage/components/ExperienceSkills.tsx
@@ -49,6 +49,13 @@ const ExperienceSkills = ({
     }
   };
 
+  const watchedSkillIds = watchedSkills.map(
+    (watchedSkill) => watchedSkill.skillId,
+  );
+  const unclaimedSkills = skills.filter(
+    (skill) => !watchedSkillIds.includes(skill.id),
+  );
+
   const handleRemoveSkill = (id: string) => {
     const index = watchedSkills.findIndex(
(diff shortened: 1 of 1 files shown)


esizer (approved):
I think you nailed it! 🔨 

Maybe see if someone else wants to check to make sure there aren't anymore spots where this is occurring, but I poked around just to be safe and couldn't find any 👍

yonikid15 (approved):
Very nice work! Everything looks good to me!

PR #8812, opened 2023-12-14
[Fix] Prevent unauthorized users from accessing pool candidate notes

### LINKED ISSUE #8808: 🐛 Pool candidate notes visible to applicant
## 🐛 Bug

It is possible for the notes added by the recruitment team to be accessed by the applicant through the API.

## 🦋 Expected Behaviour

image

## 📋 Steps to Reproduce

1. Add the bearer token for an applicant to Graphiql
2. Run this query
```graphql
query MyInfo {
  me {
    firstName
    lastName
    poolCandidates {
      pool { name { en } }      
      notes      
    }  
  }
}
```

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

There should probably be a `@can` directive on this field.  Since this field will eventually be replaced by the record of decision assessments it might make sense to guard it with `@canOnParent(ability: "viewAssessmentResults")` like the other assessment results.

## ✅ Acceptance Criteria

- [ ] Applicants can not query for notes on their pool candidate model
- [ ] phpunit test to ensure candidates cannot access "admin only" fields on their candidate model

esizer:
🤖 Resolves #8808 

## 👋 Introduction

This adds the `@canOnParent` directive to the notes field on the `PoolCandidate` model in our GraphQL schema. It uses the `viewAssessmentResults` permission allowing access to only:

- Pool operator (who is part of the team that owns the pool)
- Platform admin

## 🕵️ Details

I hope the test I wrote is good enough, it basically checks that we get the notes field for the users who should see it as well as a auth error on those that cannot. I'm hoping it is not over-testing but I wanted to cover all the possibilities 🤷‍♀️ 

## 🧪 Testing

```graphql
query MyInfo {
  poolCandidate(id: "{id}") {
    id
    notes
  }
}
```

1. Login as an applicant `applicant@test.com`
2. Submit an application
3. Logout and log back in as an admin `admin@test.com`
4. Navigate to the application you just submitted and add some notes to it
5. Make sure to copy the ID for use in the following steps
6. Logout
7. Set the default auth user to applicant `AUTH_DEFAULT_USER=applicant@test.com` in `api/.env`
8. Navigate to `http://localhost:8000/graphiql`
9. Attempt to run the query noted previously replace `{id}` with the ID copied in step 5
10. Confirm you get an unauthorized error
11. Update the default auth user to `admin@test.com` from step 5
12. Re-run the query from step 8
13. Confirm you get the notes with no errors

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -257,7 +257,7 @@ type PoolCandidate {
 
   status: PoolCandidateStatus @rename(attribute: "pool_candidate_status")
   statusWeight: Int @rename(attribute: "status_weight")
-  notes: String
+  notes: String @canOnParent(ability: "viewAssessmentResults")
   archivedAt: DateTime @rename(attribute: "archived_at")
   submittedAt: DateTime @rename(attribute: "submitted_at")
   suspendedAt: DateTime @rename(attribute: "suspended_at")
--- a/api/tests/Feature/PoolCandidateTest.php
+++ b/api/tests/Feature/PoolCandidateTest.php
@@ -29,6 +29,8 @@ class PoolCandidateTest extends TestCase
 
     protected $teamUser;
 
+    protected $unAssociatedTeamUser;
+
     protected $team;
 
     protected $teamName = 'application-test-team';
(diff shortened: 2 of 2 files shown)


petertgiles (approved):
Beautiful!

yonikid15 (approved):
I think the tests are great! Nice work!

PR #8815, opened 2023-12-14
[Fix] Improve UUID enforcement on URL params

### LINKED ISSUE #8724: 🐛 Prod error logged: Expected type UUID
## 🐛 Bug

We occasionally see this error logged in prod:
> Expected type "UUID". {"exception":"[object] (GraphQL\\Error\\CoercionError(code: 0): Expected type \"UUID\". at /home/site/wwwroot/api/vendor/webonyx/graphql-php/src/Error/CoercionError.php:30)

This is error is pretty vague so it might be worth combing through the logs around this error to get some context.

## ✅ Acceptance Criteria

- [ ] Figure out where the graphql type error is coming from
- [ ] Fix the underlying issue

esizer:
🤖 Resolves #8724 

## 👋 Introduction

This solves an issue where malformed UUIDs in the URL were not throwing errors.

## 🕵️ Details

> [!NOTE]
> A big thanks to @petertgiles for helping play detective and getting to the root of the issue! 🙇‍♀️ 🕵️ 

We encountered an error in our logs where malformed UUIDs were making it to the API. With the help of @petertgiles we were able to determine it was attempting to retrieve a Pool with a specific ID, but there was a stray `]` at the end of the string. Reviewing our code, this did not seem to be a dev error.

Further testing, I found that if you added any extra data to the start or end of a UUID in the URL, it was not being caught by our `useRequiredParams` hook even when the `enforceUUID` option was set to true. Digging deeper, I noticed the regex was only checking to see if the string contained a UUID. So, you could change data inside the UUID to see the proper error, but adding data to the beginning or end caused the `CoercionError` since no error was thrown on the client and the malformed UUID was getting to the server😬 

So, this updates that regex to enforce start (`^`) and end (`$`) points on the string so we are now asserting the entire string is equal to a UUID rather than if it contains a UUID.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to any page that has a UUID in the URL and sent to the API
3. Add anything to the start of the string
4. Confirm you get the pug error page and not a blank page with `GraphQL` error
5. Confirm the graphql request was not sent
6. Confirm the laravel log does not contain a `CoercionError`
7. Repeat steps 3-6 but add the data to the end of the UUID

### DIFF
--- a/apps/web/src/hooks/useRequiredParams.ts
+++ b/apps/web/src/hooks/useRequiredParams.ts
@@ -1,4 +1,5 @@
 import { Params, useParams } from "react-router-dom";
+import { useIntl } from "react-intl";
 
 import {
   useLogger,
@@ -6,6 +7,7 @@ import {
   type Logger,
 } from "@gc-digital-talent/logger";
 import { notEmpty } from "@gc-digital-talent/helpers";
+import { commonMessages } from "@gc-digital-talent/i18n";
 
 import invariant from "~/utils/invariant";
 
@@ -61,7 +63,9 @@ const useRequiredParams = <
 >(
   keys: K | Array<K>,
   enforceUUID: boolean = true,
+  message: string | undefined = undefined,
 ): Record<K, string> => {
+  const intl = useIntl();
   const logger = useLogger();
   const params: Params<string> = useParams<T>();
   const keyArray = Array.isArray(keys) ? keys : [keys];
@@ -70,10 +74,20 @@ const useRequiredParams = <
   nonEmptyParams = keyArray.reduce((reducedParams, key) => {
     const param: string | undefined = params[String(key)];
     let newParams = { ...reducedParams };
-    assertParam(param, enforceUUID, logger);
+    try {
+      assertParam(param, enforceUUID, logger);
 
-    if (key) {
-      newParams = { ...newParams, [key]: param }; // param must be a string if assertParam didn't throw an error.
+      if (key) {
+        newParams = { ...newParams, [key]: param }; // param must be a string if assertParam didn't throw an error.
+      }
+    } catch (err) {
+      throw new Response(
+        message ?? intl.formatMessage(commonMessages.notFound),
+        {
+          status: 404,
+          statusText: "Not Found",
+        },
+      );
     }
 
     return newParams;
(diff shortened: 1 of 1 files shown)


apps/web/src/hooks/useRequiredParams.ts
```diff
@@ -70,6 +70,7 @@ const useRequiredParams = <
   nonEmptyParams = keyArray.reduce((reducedParams, key) => {
     const param: string | undefined = params[String(key)];
     let newParams = { ...reducedParams };
+    console.log(isUUID(param ?? ""));
```
mnigh, line 73: Can this be removed?

esizer: Absolutely! [remove console log statement](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8815/commits/2aa36b77fe1ac60ac2364a58257e5f0e3bec5711)

vd1992 (approved):
Appears to work fine to me. No graphQL errors fired, neither do I see them pop up in the Laravel log. 

Adding an additional character to end of the id

Before (main)

image

After (branch)

image

vd1992:
This error arising out of user error when copy-pasting a link actually makes a lot of sense to me and is entirely reasonable to have happen. 

Not within scope but a question: for the error page might it be good to mention to double check the URL? I guess that depends on if copy-pasting links is common enough and subsequent copy-pasting mistakes. 
A sort of UX consideration to think about

esizer:
> Not within scope but a question: for the error page might it be good to mention to double check the URL? I guess that depends on if copy-pasting links is common enough and subsequent copy-pasting mistakes.

Not a bad idea. I opted to throw a 404 error since that is really what is happening here and we can't really determine how the user got to this state, just that they did.

[update required params to throw 404](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8815/commits/197b4fa62ae7eb579511f8bf59a0fc54b65a3736)

vd1992:
Does not appear to respect the intl for some reason

image

esizer:
> Does not appear to respect the intl for some reason

Oops, I forgot we display status text there 😅 Fixed in [update status text for param 404](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8815/commits/eec8aadb11402cf5c321a70d995d28be1e03e1e3)

vd1992 (approved):
Approved x2

PR #8819, opened 2023-12-15
Dark mode - Applicant misc edition

substrae:
## ✨ Summary

Contact @substrae with design questions

The goal of this PR is to clean up any outstanding applicant-facing dark mode contrast issues. The scope of this PR includes all applicant facing pages, including the talent search/request pages, IAP pages, profile, application, authentication, etc.

```[tasklist]
### ✅ Requirements
- [ ] All remaining applicant dark mode UI passes AA contrast checks
- [ ] All remaining applicant dark mode UI has dark mode styles
```


petertgiles:
@JamesHuf Could you please browse through the Chromatic diff and leave a review for Josh with anything you notice?

substrae:
@JamesHuf I will also note that the Chromatic diff might not help much in terms of review; because this is dark mode specific stuff, you'll likely need to review the site manually by ensuring the `data-h2="default"` on the `<html>` element in your inspector is set to `data-h2="dark"`

vd1992:
I wasn't sure about how to check dark mode for it at the time and used a placeholder, close enough sort of, colour here. Might want to change it to a desired hexcode 

image

image

vd1992:
Search page (and probably elsewhere)

Contrast on the required asterisk

image

Page en/indigenous-it-apprentice

Switching to dark mode, and browser does not like this colour combo for contrast in a bunch of places

image

vd1992:
Page /personal-information

Contrast on the asterisks for the form as seen above, as well as

image

vd1992:
Application flow

Applicable to regular and dark, but noticed that really long experience titles kinda make the view funky, for your consideration

image

vd1992:
Submitting a request

Contrast

image

JamesHuf:
I'm not sure how to view the contrast like @vd1992 but the "contact us" link in the add indigenous community dialog.
image

JamesHuf:
Possible contrast issue with the placeholder text on failed fields.
image
image

JamesHuf:
Contrast on the 404 page:
image

substrae:
@vd1992 @JamesHuf Awesome, thanks for the thorough review gents! I'll have commits up later today for you to re-review :)

substrae:
@vd1992 @JamesHuf Okay a few comments to go along with my commits:

* For reviewing dark mode on IAP pages, make sure you change `data-h2="iap"` to `data-h2="dark iap"`, otherwise you end up with the GCDT theme in dark mode, which will definitely have contrast conflicts
* @vd1992 I think it makes sense to file a new bug for that weird width stuff you found with really long skills; it seems out of scope for this PR, but it's absolutely something we should fix
* I've adjusted input placeholders where I could, using `::placeholder`, but this might not be consistent across browsers or inputs where we're using plugins

vd1992:
Issue created https://github.com/GCTC-NTGC/gc-digital-talent/issues/8849

vd1992:
A bit of a harsh line appeared for the banner, but I think it should be a small change to match the shadow inset from https://github.com/GCTC-NTGC/gc-digital-talent/pull/8819/commits/a588f09ed7b7978890e99ffcd27e0bfbf21150dd to the new foreground in https://github.com/GCTC-NTGC/gc-digital-talent/pull/8819/commits/be167019f0e43361b49b20ff2996dd2a1e4b7cb7

image

substrae:
@vd1992 Hilarious... I told myself 3 different times not to forget to update that value and I STILL forgot 🤦‍♂️ Fixed

vd1992:
Contrast for the null case for form fields

image

substrae:
@vd1992 good catch - I don't think they were passing in light mode either 😱 I've fixed it on the profile, but are these error states handled over and over for other inline forms on the site?

vd1992:
> @vd1992 good catch - I don't think they were passing in light mode either 😱 I've fixed it on the profile, but are these error states handled over and over for other inline forms on the site?

That one appears for cases of inline forms AKA ToggleForm I believe, of which the other example I can quickly find is for pool editing over at the admin side, but that is out of scope for this. It will be something to fix later for sure. 

image

vd1992 (approved):
What I noted appears all addressed. Looks great!

JamesHuf:
The blue icon isn't clear in the iap application stepper
image

JamesHuf:
Also the iap experience accordions 
image

substrae:
Thanks for finding those @JamesHuf!! They should be fixed now.

JamesHuf (approved):
The new in progress color is pretty similar to the error one in the stepper but if you're fine with that I think this is good to go.
image

substrae:
@JamesHuf Yeah, I thought about this as well, but the difference in the icon's shape, along with it's color, is enough in my mind. Happy to change it if others find it too confusing.

JamesHuf:
Fair enough, I won't argue with it.

PR #8830, opened 2023-12-18
[Bug] Add character limits to inputs for experience forms

### LINKED ISSUE #8725: 🐛 Prod logged error: string data truncated for experiences
## 🐛 Bug

We occasionally see errors like this logged in prod:
> SQLSTATE[22001]: String data, right truncated: 7 ERROR:  value too long for type character varying(255) (Connection: pgsql, SQL: insert into "education_experiences" ("institution", "type", "status", "details", "user_id", "area_of_study", "thesis_title", "start_date", "end_date", "updated_at", "created_at") values... 

Worse, our app throws up a success and an error toast and closes the dialog when the error response comes back.

image

## ✅ Acceptance Criteria

- [ ] Ensure the dialogs do not close if the request comes back with an error (for example, text longer that 255 char).
- [ ] Ensure a success toast is not popped if the request comes back with an error
- [ ] Add a [maxlength](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength) or react-hook-form rules to the inputs backed by varchar(255) columns to avoid submitting the request at all
- [ ] Scope: all five experience dialogs

vd1992:
🤖 Resolves #8725

## 👋 Introduction

Uses `maxLength` rules for inputs, as well as correct toasting/action on errored updates. 

## 🕵️ Details

Elected to check if an error is returned in the response object, given the nested data object contains 10 possibles keys (5 experiences x 2 mutation types) the truthy check wasn't working right so I thought it would be easier to assert on `res.error` instead. 

## 🧪 Testing

1. Head to career timeline and the experiences section of the application flow
2. Check you cannot trigger the value too long error for VARCHAR database error
3. Check failed operations do not return a success toast

A suggestion to check the third option. Open an experience, then open and soft-delete it in adminer, then attempt to update it in app. 
Generate strings of a certain character count with https://loremipsum360.com/

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/components/ExperienceFormFields/AwardFields.tsx
+++ b/apps/web/src/components/ExperienceFormFields/AwardFields.tsx
@@ -34,7 +34,15 @@ const AwardFields = ({ labels }: SubExperienceFormProps) => {
             label={labels.awardTitle}
             name="awardTitle"
             type="text"
-            rules={{ required: intl.formatMessage(errorMessages.required) }}
+            rules={{
+              required: intl.formatMessage(errorMessages.required),
+              maxLength: {
+                message: intl.formatMessage(errorMessages.overCharacterLimit, {
+                  value: 256,
+                }),
+                value: 255,
+              },
+            }}
           />
         </div>
         <div data-h2-flex-item="base(1of1) p-tablet(1of2)">
@@ -68,7 +76,15 @@ const AwardFields = ({ labels }: SubExperienceFormProps) => {
             label={labels.issuedBy}
             name="issuedBy"
             type="text"
-            rules={{ required: intl.formatMessage(errorMessages.required) }}
+            rules={{
+              required: intl.formatMessage(errorMessages.required),
+              maxLength: {
+                message: intl.formatMessage(errorMessages.overCharacterLimit, {
+                  value: 256,
+                }),
+                value: 255,
+              },
+            }}
           />
         </div>
         <div data-h2-flex-item="base(1of1) p-tablet(1of2)">
--- a/apps/web/src/pages/Applications/ApplicationCareerTimelineEditPage/components/ExperienceEditForm.tsx
+++ b/apps/web/src/pages/Applications/ApplicationCareerTimelineEditPage/components/ExperienceEditForm.tsx
@@ -73,7 +73,16 @@ const EditExperienceForm = ({
     if (executeMutation) {
       executeMutation(args)
         .then((res) => {
-          if (res.data) {
+          if (res.error) {
+            toast.error(
+              intl.formatMessage({
+                defaultMessage: "Error: updating experience failed",
+                id: "WyKJsK",
+                description:
+                  "Message displayed to user after experience fails to be updated.",
+              }),
+            );
+          } else {
             toast.success(
               intl.formatMessage({
                 defaultMessage: "Successfully updated experience!",
--- a/packages/i18n/src/messages/errorMessages.ts
+++ b/packages/i18n/src/messages/errorMessages.ts
@@ -47,6 +47,12 @@ const errorMessages = defineMessages({
     description:
       "Error Message displayed on word counter when user passes the limit.",
   },
+  overCharacterLimit: {
+    defaultMessage: "This field must have less than {value} characters.",
+    id: "GiYTD4",
+    description:
+      "Error Message displayed when the user passes some given character or letter limit.",
+  },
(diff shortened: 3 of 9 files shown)


packages/i18n/src/messages/errorMessages.ts
```diff
@@ -47,6 +47,12 @@ const errorMessages = defineMessages({
     description:
       "Error Message displayed on word counter when user passes the limit.",
   },
+  overCharacterLimit: {
+    defaultMessage: "This field must have less than {value} characters.",
```
mnigh, line 51: In case this is the only translation needed, here is a suggestion: 

```
Ce champ doit contenir moins de {value} caractères.
``` 

We already have a nearly identical string for this purpose:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/62cbc97fa2e78ffda23f9a586b87226915780adf/packages/i18n/src/messages/errorMessages.ts#L44-L49

https://github.com/GCTC-NTGC/gc-digital-talent/blob/62cbc97fa2e78ffda23f9a586b87226915780adf/packages/i18n/src/lang/fr.json#L1830-L1833


Also, [Termium](https://www.btb.termiumplus.gc.ca/tpv2alpha/alpha-eng.html?lang=eng&i=1&srchtxt=CHARACTER&codom2nd_wet=1&index=alt) suggests _caractère_ as the translation for this type of **character**.

vd1992: Hm, I think that is reasonable, almost identical with the exception of the term at the end and we probably want these two translations in sync anyway since their purpose/phrasing is the same

vd1992: https://github.com/GCTC-NTGC/gc-digital-talent/pull/8830/commits/1fbf0da939ad61d83c93c2efca980c654f3b149a

petertgiles:
@mnigh Could you please either Approve or Request on this PR?

apps/web/src/components/ExperienceFormFields/AwardFields.tsx
```diff
@@ -34,7 +34,15 @@ const AwardFields = ({ labels }: SubExperienceFormProps) => {
...
-            rules={{ required: intl.formatMessage(errorMessages.required) }}
+            rules={{
+              required: intl.formatMessage(errorMessages.required),
+              maxLength: {
+                message: intl.formatMessage(errorMessages.overCharacterLimit, {
+                  value: 256,
+                }),
+                value: 255,
+              },
+            }}
```
yonikid15, line 45: Could we add this directly into the `<Input />` component so we can prevent this error across the app?

yonikid15 (changes requested):

apps/web/src/components/ExperienceFormFields/AwardFields.tsx
yonikid15: Something like this:
```tsx
// packages/forms/src/components/Input.tsx line 77
 {...register(name, {
          maxLength: {
            message: intl.formatMessage(errorMessages.overCharacterLimit, {
              value: 256,
            }),
            value: 255,
          },
          ...rules,
          onBlur: normalizeInput,
        })}
 ...
```

petertgiles: Decision from standup: we like this idea, will do it in a separate issue

vd1992:
Not sure why Snyk was permanently pending, but a merge main solved that

apps/web/src/pages/Applications/ApplicationCareerTimelineEditPage/components/ExperienceEditForm.tsx
```diff
@@ -73,7 +73,16 @@ const EditExperienceForm = ({
     if (executeMutation) {
       executeMutation(args)
         .then((res) => {
-          if (res.data) {
+          if (res.error) {
+            toast.error(
+              intl.formatMessage({
+                defaultMessage: "Error: updating experience failed",
```
mnigh, line 79: Not part of the acceptance criteria, but for some of the fields that don't have any character limit, this error doesn't seem descriptive enough. Is there a reason why we don't have a word limit on the Experience Description field, for example?

## Screenshot
Screen Shot 2023-12-20 at 09 02 24

mnigh (approved):
One comment that is likely out of scope, but otherwise, LGTM.

yonikid15 (approved):
LGTM! Nice work 👍

PR #8835, opened 2023-12-18
[Bug] Exception thrown on id token check

### LINKED ISSUE #8723: 🐛 Prod error logged: null token parse
## 🐛 Bug

We get this error logged occasionally in prod:
> Lcobucci\JWT\Token\Parser::parse(): Argument #1 ($jwt) must be of type string, null given, called in /home/site/wwwroot/api/app/Http/Controllers/AuthController.php on line 86 {"exception":"[object] (TypeError(code: 0): Lcobucci\\JWT\\Token\\Parser::parse(): Argument #1 ($jwt) must be of type string, null given, called in /home/site/wwwroot/api/app/Http/Controllers/AuthController.php on line 86 at /home/site/wwwroot/api/vendor/lcobucci/jwt/src/Token/Parser.php:26)

## ✅ Acceptance Criteria

- [ ] Find out where we try to parse the null
- [ ] Gracefully handle it to prevent the Lcobucci exception and stack trace

vd1992:
🤖 Resolves #8723 

## 👋 Introduction

Check the id token is a string, if it isn't throw an exception. 

## 🧪 Testing

Not entirely sure of the best way, an idea below

1. Set id token variable to a different type, like an integer value, right before the `throw_unless()` on `$idToken`
2. Attempt logging in, observe exception message

### DIFF
--- a/api/app/Http/Controllers/AuthController.php
+++ b/api/app/Http/Controllers/AuthController.php
@@ -83,6 +83,12 @@ class AuthController extends Controller
         $config = $this->fastSigner;
 
         assert($config instanceof Configuration);
+
+        throw_unless(
+            $idToken && is_string($idToken),
+            new InvalidArgumentException('id token is a '.gettype($idToken))
+        );
+
         $token = $config->parser()->parse($idToken);
         assert($token instanceof UnencryptedToken);
 
--- a/api/app/Http/Controllers/AuthController.php
+++ b/api/app/Http/Controllers/AuthController.php
@@ -5,6 +5,7 @@ namespace App\Http\Controllers;
 use App\Services\OpenIdBearerTokenService;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Http;
+use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Str;
 use InvalidArgumentException;
 use Lcobucci\JWT\Configuration;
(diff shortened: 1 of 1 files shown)


api/app/Http/Controllers/AuthController.php
```diff
@@ -83,6 +83,12 @@ public function authCallback(Request $request)
         $config = $this->fastSigner;
 
         assert($config instanceof Configuration);
+
+        throw_unless(
+            $idToken && is_string($idToken),
+            new InvalidArgumentException('id token is a '.gettype($idToken))
+        );
```
vd1992, line 90: Wasn't sure if outputting the type is really needed, and if just pointing out it isn't a string sufficies

petertgiles (approved):
Yay, this works. :partying_face:   We get a good error message with a good stack trace.
image  I used the debugger to change the value of $idToken and let it run to trigger this.

You've satisfied the AC so I'm approving.  I'd like to level this up a bit, though, if you're up for it.  

First, let's move our code up so it's right below `$idToken = $response->json('id_token');`.  There's no point in setting up other stuff if our token is bad.

Next, we already know that the id token is sometimes coming back null but we don't know why.  I'd like to dump the the entire raw response to log and see if there's an error message or something in there.
image
To do this you'll probably have to switch from the throw_unless convenience function and use a regular
```php
if(condition) {
  //extra logging
  // throw exception
}

vd1992:
I am not sure why, but it seems like if statements are not playing nice with throwing exceptions. 
The below doesn't seem to break login at all for example

image

petertgiles:
> The below doesn't seem to break login at all for example

You have to throw the exception. :grin:  Just instantiating it isn't enough.
https://rollbar.com/guides/php/how-to-throw-exceptions-in-php/#

vd1992:
> > The below doesn't seem to break login at all for example
> 
> You have to throw the exception. 😁 Just instantiating it isn't enough. https://rollbar.com/guides/php/how-to-throw-exceptions-in-php/#

Oops! I didn't think too hard about the "throw" part in `throw_unless()`  😅 
Changed in https://github.com/GCTC-NTGC/gc-digital-talent/pull/8835/commits/479701539e2f296566bc287a58e64e5bcd50c1bc

petertgiles (approved):
Looks great!

PR #8836, opened 2023-12-19
[Bug] Retrying call to auth endpoint

### LINKED ISSUE #8777: 🐛 Prod logged error - Empty reply from server
## 🐛 Bug

We occasionally see this error in the prod logs:
> cURL error 52: Empty reply from server (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for https://auth.id.canada.ca/oxauth/restv1/token {"exception":"[object] (Illuminate\\Http\\Client\\ConnectionException(code: 0): cURL error 52: Empty reply from server (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for https://auth.id.canada.ca/oxauth/restv1/token at /home/site/wwwroot/api/vendor/laravel/framework/src/Illuminate/Http/Client/PendingRequest.php:926)

This URL is used during the auth callback and refresh to get tokens.  That seems to be failing for some reason.

## 🙋‍♀️ Proposed Solution

Add a few retries to the request in case the issue is transient.

## ✅ Acceptance Criteria

- [ ] Add a few retries to the requests out to Sign In Canada endpoints.
- [ ] Add extra logging to failures so we have more details if we need to escalate to SIC admins

vd1992:
🤖 Resolves #8777 

## 👋 Introduction

Add retrying and a logging line to the auth callback function, the client request method. 

## 🕵️ Details

Attempted to just follow documentation for implementation, not 100% on things. 

https://laravel.com/docs/10.x/http-client

## 🧪 Testing

1. Can login successfully
2. ?

### DIFF
--- a/api/app/Http/Controllers/AuthController.php
+++ b/api/app/Http/Controllers/AuthController.php
@@ -83,7 +83,7 @@ class AuthController extends Controller
             'redirect_uri' => config('oauth.redirect_uri'),
             'code' => $request->code,
         ])->throw(function (Response $response, RequestException $e) {
-            Log::debug('Failed to get a response from POSTing to the token URI');
+            Log::error('Failed to get a response from POSTing to the token URI');
             Log::debug((string) $response->getBody());
         });
 
--- a/api/app/Http/Controllers/AuthController.php
+++ b/api/app/Http/Controllers/AuthController.php
@@ -3,8 +3,14 @@
 namespace App\Http\Controllers;
 
 use App\Services\OpenIdBearerTokenService;
+use Exception;
+use Illuminate\Http\Client\ConnectionException;
+use Illuminate\Http\Client\PendingRequest;
+use Illuminate\Http\Client\RequestException;
 use Illuminate\Http\Request;
+use Illuminate\Http\Response;
 use Illuminate\Support\Facades\Http;
+use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Str;
 use InvalidArgumentException;
 use Lcobucci\JWT\Configuration;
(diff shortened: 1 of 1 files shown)


petertgiles (changes requested):
Cool, I was hoping that you would find a nice clean way to do retries!

Sadly, I'm having trouble testing this one, too.  I tried to use [the Http::fake function](https://laravel.com/docs/10.x/http-client#faking-responses) in the facade but it's not making a response that will throw an exception. It seems like we can program the mockauth server to issue certain responses but that's something that I would have to spend some time on to figure out.  I guess this issue is already a "hope for the best" type of fix.

My only request is that we level-up the logging a bit.  Let's add an initial log message so that we know why the response is showing up in the log.  Something like "Failed to get a response from POSTing to the token URI".  Also, let's log the entire raw response instead of trying to parse out JSON, in case the response is a non-json error message.  Like I mentioned in #8835, try `(string)$response->getBody()`.

vd1992:
Sure, I made the logging changes in https://github.com/GCTC-NTGC/gc-digital-talent/pull/8836/commits/2815021ddcf1b169abccb9742a501c6ec2719489
I wasn't 100% on how much to retry or with how much time in-between, any opinions on that?

petertgiles (approved):
Nice, looks good to me.  The retry settings seem fine, too.
I suggest switching this one from "debug" to "error" though.
> Log::debug('Failed to get a response from POSTing to the token URI');

vd1992:
> Nice, looks good to me. The retry settings seem fine, too. I suggest switching this one from "debug" to "error" though.
> 
> > Log::debug('Failed to get a response from POSTing to the token URI');

Done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/8836/commits/049508b4fbbe53e859513bb5f482d60368b65d02

@brindasasi 
I believe you also wanted to take a look too, right?

brindasasi:
Not this one. I said about the spike. 
But looks good to me as well 👍

tristan-orourke:
Will `Log::debug` level messages get logged on production? Shouldn't the whole response be in an `error` level log? We are catching an exception, after all.

petertgiles:
> Will `Log::debug` level messages get logged on production? Shouldn't the whole response be in an `error` level log? We are catching an exception, after all. (I'm open to disagreement on that, just not sure what the argument against an error log would be.)

I figured the main error message would be at ERROR and the dump of the request would be at DEBUG.  It's a bit of an academic discussion right now though, because everything gets logged as ERROR in prod right now.  I'd like to tighten that up, though. https://talent-cloud.slack.com/archives/C0259G6KXJ6/p1703094110033009?thread_ts=1703092808.349839&cid=C0259G6KXJ6

petertgiles (approved):

PR #8842, opened 2023-12-19
[Fix] Table date sort

### LINKED ISSUE #8838: 🐛 Sorting on the Updated column of pools table is no longer working
## 🐛 Bug

Pools table can not be sorted by **Updated** ( Last modified date) column any more.
Also recently created pool is not appearing on the top of the table.
At this point it is very hard to find a pool since the name of the pools are almost same in production 
and sorting by name doesn't help at all. 

## 🦋 Expected Behaviour

1. Sorting by "Updated" column should work ( I would like a better name like "Last Modified Date" .. but you know :) ) 
2. It should be sorted that column by default 
 
## 🕵️ Details

## 📋 Steps to Reproduce

1. Go to pools table 
2. Click on "Updated" column to sort by it. 

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] Sorting by "Updated" column should work
- [ ] Make it default sorting for the table

esizer:
🤖 Resolves #8838 

## 👋 Introduction

Fixes the issue with tables not sorting date columns properly.

## 🕵️ Details

We had the `accessor` and `cells` mixed up 😅 . This updates it so the accessor returns a `Date` and adds a `cell` to handle the formatting of said date.

> [!NOTE]
> We have been sorting the `PoolTable`  by `ceatedDate` so I was hesitant to change this. Do we still want to be changing this to `updatedDate`.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to tables in the admin
3. Sort any date columns
4. Confirm they are sorted properly

## 📸 Screenshot

Screenshot 2023-12-19 124640

### DIFF
--- a/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
+++ b/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
@@ -509,12 +509,19 @@ const PoolCandidatesTable = ({
       },
     ),
     columnHelper.accessor(
-      ({ poolCandidate: { submittedAt } }) => accessors.date(submittedAt, intl),
+      ({ poolCandidate: { submittedAt } }) => accessors.date(submittedAt),
       {
         id: "dateReceived",
         enableColumnFilter: false,
         header: intl.formatMessage(tableMessages.dateReceived),
         sortingFn: "datetime",
+        cell: ({
+          row: {
+            original: {
+              poolCandidate: { submittedAt },
+            },
+          },
+        }) => cells.date(submittedAt, intl),
       },
     ),
   ] as ColumnDef<PoolCandidateWithSkillCount>[];
--- a/apps/web/src/components/PoolStatusTable/PoolStatusTable.tsx
+++ b/apps/web/src/components/PoolStatusTable/PoolStatusTable.tsx
@@ -166,7 +166,7 @@ const PoolStatusTable = ({ user, pools }: UserInformationProps) => {
         }),
       },
     ),
-    columnHelper.accessor((row) => accessors.date(row.expiryDate, intl), {
+    columnHelper.accessor((row) => accessors.date(row.expiryDate), {
       id: "expiryDate",
       enableHiding: false,
       sortingFn: "datetime",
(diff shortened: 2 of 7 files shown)


brindasasi:
I don't mind sorting by created date. But the thing is it is hidden by default 😅
Default visible column is "Updated" on that table. 
I would recommend doing the default sorting on default visible columns.

could we make "Created" column as visible by default? 
cc: @GCTC-NTGC/designers

brindasasi (changes requested):
Thank you for prioritizing this ticket. Sorting works nicely.
but on the pool candidates table , if we go to second or third pages, and then try to change the sorting order on the date column, it sorts randomly. 
If you look at the screen shot , I sorted by descending order on the 3rd page, it gets Oct ,Nov and then oct again.
Screenshot 2023-12-20 at 2 41 37 PM.
Could you check why it is happening?

esizer:
> Thank you for prioritizing this ticket. Sorting works nicely.
> but on the pool candidates table , if we go to second or third pages, and then try to change the sorting order on the date column, it sorts randomly.
> If you look at the screen shot , I sorted by descending order on the 3rd page, it gets Oct ,Nov and then oct again.

heh, yeah. So, this is _highly_ confusing but, it technically is sorting properly. We have multiple sorts going on here where two take precedence over the submitted date. It sorts like this:

1. Status
2. Category
3. Date (user controlled)

So what it happening is, when you sort by date received it first sorts by Status, then if two users have the same status it will sort by category, then if both the status and category are the same it finally sorts by date received.

Hopefully that clears things up 😅

brindasasi (approved):
you are right! It goes by that logic. After picking up specific status & category, sorting looks fine by the date. Thanks!👏

RM-1978:
Humm, I wonder if that logic might be a little complex for a user who simply wants to see the most recently updated pools? Do we need instructional text?

esizer:
> Humm, @esizer I wonder if that logic might be a little complex for a user who simply wants to see the most recently updated pools? Do we need instructional text?

We may need to ask DCM, they were the ones who originally requested this functionality so we can prioritize certain groups.

PR #8847, opened 2023-12-20
[Fix] Have memoized refresh function access localstorage directly

### LINKED ISSUE #8845: 🐛 Auth provider trying to reuse refresh tokens causing a refresh failure
## 🐛 Bug

We've had reports of users being kicked out of the app unexpectedly.  I believe the auth provider is not storing the the refresh token from the a refresh response.  When the next refresh is attempted it trying to reuse the old refresh token and the refresh is failing.

image

## 🦋 Expected Behaviour

The auth provider should store the new refresh token from a refresh and use that for the next refresh

## 📋 Steps to Reproduce

> [!IMPORTANT]
This can only be reproduced with SIC.  It does not happen with mock auth.

1. Log in, note the refresh token (A)
2. Wait for the access token to expire, observe that refresh token (A) is used to refresh and a new refresh token (B) is provided
3. Wait for the access token to expire, observe that refresh token (A) is reused and the refresh failes

## ✅ Acceptance Criteria

- [ ] User can refresh their session multiple times consecutively with SIC

petertgiles:
🤖 Resolves #8845

## 👋 Introduction

This branch updates the memoized refresh function to access the refresh token from localstorage instead of just using the first token frozen from when it was memoized.

## 🧪 Testing

> [!IMPORTANT]
Make sure you're testing with SIC and a prod build
https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/documentation/sign-in-canada.md

### Primary Focus

1. Rebuild the app with SIC and prod mode.
2. Log in, note the refresh token provided.
3. Wait until the access token expires and trigger a refresh.  Ensure that the first refresh token was used and a second refresh token was provided.
4. Wait until that access token expires and trigger a refresh.  Ensure that the second refresh token was used and a third refresh token was provided.

### Secondary Focus
 
It's probably  worth running through the original checklist again:

- [x] Can log in.
- [x] Can wait five minutes and then the token is refreshed on the next API call.
- [x] When you have two tabs open, a refresh in one will allow the second tab to make an API call with the new tokens and no refresh.
- [x] Can log out (with SiC endsession).
- [x] If [the introspection fails](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6547#pullrequestreview-1602788860) the user is immediately logged out.
- [x] If you log in as a deleted user you end up on the "user deleted" page.

### DIFF
--- a/packages/auth/src/components/AuthenticationContainer.tsx
+++ b/packages/auth/src/components/AuthenticationContainer.tsx
@@ -110,11 +110,6 @@ const AuthenticationContainer = ({
   children,
 }: AuthenticationContainerProps) => {
   const logger = useLogger();
-  const existingTokens = {
-    accessToken: localStorage.getItem(ACCESS_TOKEN),
-    refreshToken: localStorage.getItem(REFRESH_TOKEN),
-    idToken: localStorage.getItem(ID_TOKEN),
-  };
 
   const newTokens = getTokensFromLocation();
   logger.debug(`new tokens from location: ${JSON.stringify(newTokens)}`);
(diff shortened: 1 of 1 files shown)


tristan-orourke:
Out of curiousity, since it's not clear from the diff - what was it about the auth bump that made this start to fail _now_?

esizer (approved):
This may have taken almost 30 mins (mostly down time) to test but I am happy to report that I am not logged out every 5 mins now! 🎉 😅 

I also did the other regular tests as suggested and they all worked as expected as well 👍 

I'm going to give my approval but I'd say wait for @tristan-orourke to be absolutely sure.

vd1992:
> Out of curiousity, since it's not clear from the diff - what was it about the auth bump that made this start to fail _now_?

Yeah, definitely, why the perplexing lag time to failing

petertgiles:
> Out of curiousity, since it's not clear from the diff - what was it about the auth bump that made this start to fail _now_?

The refactor that I did in #8618 introduced this bug.  The refreshTokenSet function used to be a regular function that pulled the refresh token out of storage when called.  Then the AuthenticationContainer memoized its state variable with a reference to that function.  When I refactored it, I moved that refreshTokenSet function inline in the memoized state object.  I probably thought "Hey, now that I'm inline I can just grab the local token variables instead of pulling them out of storage."  I forgot that it was memoized so that local variable gets frozen and won't update after a refresh.  :face_exhaling: 

This bug has been there for two weeks but we don't notice it locally because mockauth doesn't care what you send as a refresh token.  And we all use mockauth locally.

tristan-orourke (approved):
Thanks for the explanation @petertgiles! I've tested all the scenarios you suggested and it looks good! 🚀

Issue #8898, opened 2023-12-22
🔤 N/A used inconsistently and ambiguously


mnigh:
## 🔤 Copy

The term **N/A** is used inconsistently on the site and is ambiguous since it can mean not applicable or not available.

### Screenshot

<img width="475" alt="Screen Shot 2023-12-22 at 09 47 44" src="https://github.com/GCTC-NTGC/gc-digital-talent/assets/3046459/e3655442-a02f-4903-9d14-fedf57feab71">

## 🙋‍♀️ Proposed Implementation

Replace all instances of **N/A** with **Not applicable** or **Not available** to avoid ambiguity of the abbreviation. Consolidate existing strings so that they use a single source of truth for each term. There are already strings for this but they need to be translated again.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/f6b0ce8cab1a021cd178995d2ee5bb03c4403fc0/packages/i18n/src/messages/commonMessages.ts#L56-L61

https://github.com/GCTC-NTGC/gc-digital-talent/blob/f6b0ce8cab1a021cd178995d2ee5bb03c4403fc0/packages/i18n/src/messages/commonMessages.ts#L138-L142

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the copy has been added or updated appropriately.

- [ ] no instances of **N/A** appear in the codebase
- [ ] **Not applicable** translated in French correctly
- [ ] **Not available** translated in French correctly

gobyrne:
I vote for just spelling out "Not applicable" everywhere

gobyrne:
@substrae makes the point that is a well defined dictionary abbreviation and there are potentially some places where the longer text will affect the designs.

Sorry for the last minute change @mnigh , but can you please wrap this in a abbreviation tag instead? Then we can use "not applicable" or "not available" as appropriate in the tag.

mnigh:
> the point that is a well defined dictionary abbreviation

Yes, n/a appears in the dictionary, but not applicable **and** not available are listed in the [cambridge dictionary](https://dictionary.cambridge.org/dictionary/english/n-a).

## Screenshot
<img width="755" alt="Screen Shot 2024-03-21 at 14 53 58" src="https://github.com/GCTC-NTGC/gc-digital-talent/assets/3046459/76c8a6d6-6221-4626-b711-b24d6afbec34">

> there are potentially some places where the longer text will affect the designs.

[...]

Issue #8896, opened 2023-12-22
🔤 View inconsistently translated in French


mnigh:
## 🔤 Copy

The string **view**, meaning the same thing everywhere, has been translated in different ways in French.

### View = Affichage
https://github.com/GCTC-NTGC/gc-digital-talent/blob/f6b0ce8cab1a021cd178995d2ee5bb03c4403fc0/apps/web/src/components/PoolCandidatesTable/tableMessages.ts#L15-L19

https://github.com/GCTC-NTGC/gc-digital-talent/blob/f6b0ce8cab1a021cd178995d2ee5bb03c4403fc0/apps/web/src/lang/fr.json#L5882-L5885

### View = Voir
https://github.com/GCTC-NTGC/gc-digital-talent/blob/f6b0ce8cab1a021cd178995d2ee5bb03c4403fc0/apps/web/src/messages/adminMessages.ts#L14-L18

https://github.com/GCTC-NTGC/gc-digital-talent/blob/f6b0ce8cab1a021cd178995d2ee5bb03c4403fc0/apps/web/src/lang/fr.json#L4030-L4033

##### Consulter

- View notifications = Consulter toutes les notifications
- View all notifications = Consulter toutes les notifications
- View and manage your notification history. = Consultez et gérez l'historique de vos notifications.
- View up-to-date profile = Consulter le profil à jour
- View experience = Consulter l’expérience
- View Notes = Consulter les notes
- ~View full summary = Consulter le résumé intégral~
- View assessment plan = Consulter le plan d’évaluation
- View the job poster for this recruitment process = Consulter l’affiche d’emploi pour ce processus de recrutement
- View experience for {experienceName} = Veuillez consulter l’expérience pour {experienceName}

> [!NOTE]  
> Not sure why in French _veuillez_ (please) is included in the French translate when it is not there in English.

##### Afficher
- View open pools = Afficher les bassins ouverts
- View team members = Afficher les membres de l’équipe
- View user = Afficher l’utilisateur

##### Voir
- View results = Voir les résultats
- View Candidates<hidden> for {label}</hidden> = Voir les candidats<hidden> pour {label}</hidden>
- View my application = Voir ma demande
- View advertisement = Voir l’annonce
- View {name} <hidden>{hiddenLabel}</hidden> = Vue {name} <hidden>{hiddenLabel}</hidden>

[...]

PR #8899, opened 2023-12-22
[Copy] Consolidates duplicate strings

### LINKED ISSUE #8822: 🔤 Consolidate English strings and update translations
## 🔤 Copy

> [!NOTE]  
> This issue is a refocused issue from #8463.  Refer back to that one for more context.  

We have many instances where multiple English strings with the same defaultMessage have been translated into different French strings.  Let's consolidate these English strings and use the updated translation.

## 🕵️ Details

h/t @RM-1978 for providing the updated translations.

The list of 86 new translations can be found in https://github.com/GCTC-NTGC/gc-digital-talent/files/13731311/Master_GCDT.translation.consistency_client.verified.12-20-2023.no.GCKey.xlsx .

## ✅ Acceptance Criteria

- [ ] Consolidate the multiple English strings to the same hash ID.
  - [ ] Syncing up the descriptions is fine.  Move to a shared messages file only where it's easy and makes sense.
- [ ] Upload the new translations from the workbook.

mnigh:
🤖 Resolves #8822.

## 👋 Introduction

This PR consolidates dozens of strings in order to eliminate duplicates.

## 🕵️ Details

There are several approaches that have been used to accomplish deduplication. 

The first was to use the same description for similar terms in order to generate a single hash id and therefore only one french translation. This approach was generally used when a term was referenced less than three times and not necessarily a common term to be reused. Some of these strings might actually benefit from larger refactoring to consolidate entire or pieces of components.

The second approach was to reference a shared string in one of the message files such as `commonMessages.ts`. This approach worked best with strings used more than twice and that were using the same string with the same meaning regardless of context.

A Google sheet is provided for reference with some notes: https://docs.google.com/spreadsheets/d/16ZPiM8yqiLeB0WhdTxlRikGgNRt6Iw9NOkiPGT772oQ/edit#gid=1089933606

> [!NOTE]
> Two strings are more complex and required individual issues:
>
> 1. https://github.com/GCTC-NTGC/gc-digital-talent/issues/8896
> 2. https://github.com/GCTC-NTGC/gc-digital-talent/issues/8898

## 🧪 Testing

1. Manually review site in English and French to make sure site still makes sense
2. Use chromatic diffs to observe some of the differences


esizer (approved):
LGTM! 👍

PR #8917, opened 2023-12-29
[Feature] Add Playwright

### LINKED ISSUE #8809: 🛠️ Test out Playwright
## 🛠️ Tooling

In the retro for the "Xenomorphs" sprint we decided we would like to try out Playwright by "rewriting an awful Cypress test" and see how it goes.

## 🕵️ Details

https://playwright.dev/

## ✅ Acceptance Criteria

- [ ] Rewrite an awful Cypress test in Playwright
- [ ] Make a recommendation to the team on whether they think it is worth a full migration

esizer:
🤖 Resolves #8809 

## 👋 Introduction

This is an example of [Playwright](https://playwright.dev/) implementation.

## 🕵️ Details

> [!NOTE]
> Unsure if this will get merged but it demonstrates how we could effectively implement playwright. I could take it further if the team is interested.

First off, this was very enjoyable and imo, much easier to work with and understand than Cypress. I'm going tot take my time to try and explain what is going on here but if you are reviewing and have questions, feel free to reach out.

### Async/Await

One of my largest gripes with Cypress is that it tries hard to hide async functionality from you. Playwright is different in that it embraces it and feels far less like magic. Instead of just assuming a navigation is waited on, Playwright makes you add `await` before to be explicit. 

### Fixtures

REF: https://playwright.dev/docs/test-fixtures

One of the largest changes was the idea of fixtures. In Cypress, we were just using them as static json. In Playwright, they are classes that encapsulates the current environment and functionality. So, in our setup we create an `adminPage` and `applicantPage` fixture that has that user logged in and isolated to that fixture. This means, we no longer need to keep logging in and out to do different things. As well, since these are just classes, you can add methods for common functionality for that user/page and extend them!

```ts
test("something", async ({ adminPage, applicantPage }) => {
  // This user can go to admin
  await adminPage.page.goto("/admin");

  // This user cannot (no need to logout/login)
  await applicantPage.page.goto("/admin");
});
```

### Projects

REF: https://playwright.dev/docs/test-projects

I'm not sure if cypress has this concept but it is neat. When you run `npx playwright test` it runs all projects at once in parallel. So we are testing chromium, firefox and webkit all that once with no extra work.

### Debugging

I never had any luck debugging tests in cypress but found playwright just worked. If you run `npx playwright test --debug` it opens the ui and a debugger console allowing you to step through your tests like a normal debugger. This made finding issues with my tests much easier.

## 🧪 Testing

> [!NOTE]
> I only added a single script and opted to pass flags for now since this was more exploratory (sorry 😅 ). We can add easier to run commands for ui and debugging if we decide we want to move forward.

1. Install `npm run install`
2. Build `bpm run build`
3. Run tests in headless `npm run e2e:playwright`
4. Confirm it works
5. Try out the ui `npm run e2e:playwright -- -- --ui`
6. Confirm it works
7. Try out the debug mode `npm run e2e:playwright -- -- --debug`

## 📸 Screenshot

Screenshot 2023-12-29 153319

### DIFF
--- a/apps/playwright/fixtures/AppPage.ts
+++ b/apps/playwright/fixtures/AppPage.ts
@@ -1,13 +1,15 @@
 import { type Page } from "@playwright/test";
 
+import { Test_MeQueryDocument } from "~/utils/user";
+
+/**
+ * App Page
+ *
+ * Common functionality, extended by other pages
+ */
 export class AppPage {
   constructor(public readonly page: Page) {}
 
-  /**
-   * Go to the homepage
-   *
-   * Common operation on all pages
-   */
   async gotoHome() {
     await this.page.goto("/");
   }
@@ -54,5 +56,11 @@ export class AppPage {
 
       return isGraphql && isOperation;
     });
+  }
+
+  async getMe() {
+    const res = await this.graphqlRequest(Test_MeQueryDocument);
+
+    return res.me;
   }
 }
--- a/apps/playwright/fixtures/AppPage.ts
+++ b/apps/playwright/fixtures/AppPage.ts
@@ -1,0 +1,49 @@
+import { type Page } from "@playwright/test";
+
+export class AppPage {
+  constructor(public readonly page: Page) {}
+
+  /**
+   * GraphQL Request
+   *
+   * Make a GraphQL request using the current page context
+   *
+   * @param query
+   * @param variables
+   * @returns
+   */
+  async graphqlRequest(query: string, variables?: Record<string, unknown>) {
+    const res = await this.page.request.post("/graphql", {
+      headers: {
+        "Content-Type": "application/json",
+      },
+      data: {
+        query,
+        variables,
+      },
+    });
+
+    const json = await res.json();
+
+    return json.data;
+  }
+
+  /**
+   * Wait for as specific GraphQL Response by operation name
+
+   * @param operationName
+   */
+  async waitForGraphqlResponse(operationName: string) {
+    await this.page.waitForResponse(async (resp) => {
+      let isOperation: boolean = false;
+      const isGraphql = await resp.url().includes("/graphql");
+
+      if (isGraphql) {
+        const body = await resp.text();
+        isOperation = body.includes(operationName);
+      }
+
+      return isGraphql && isOperation;
+    });
+  }
+}
--- a/apps/playwright/fixtures/AppPage.ts
+++ b/apps/playwright/fixtures/AppPage.ts
@@ -4,6 +4,15 @@ export class AppPage {
   constructor(public readonly page: Page) {}
 
   /**
+   * Go to the homepage
+   *
+   * Common operation on all pages
+   */
+  async gotoHome() {
+    await this.page.goto("/");
+  }
+
+  /**
    * GraphQL Request
    *
    * Make a GraphQL request using the current page context
--- a/apps/playwright/fixtures/AppPage.ts
+++ b/apps/playwright/fixtures/AppPage.ts
@@ -10,8 +10,8 @@ import { Test_MeQueryDocument } from "~/utils/user";
 export class AppPage {
   constructor(public readonly page: Page) {}
 
-  async gotoHome() {
-    await this.page.goto("/");
+  async gotoHome(locale: "en" | "fr" = "en") {
+    await this.page.goto(`${locale}`);
   }
 
   /**
@@ -46,15 +46,12 @@ export class AppPage {
    */
   async waitForGraphqlResponse(operationName: string) {
     await this.page.waitForResponse(async (resp) => {
-      let isOperation: boolean = false;
-      const isGraphql = await resp.url().includes("/graphql");
-
-      if (isGraphql) {
-        const body = await resp.text();
-        isOperation = body.includes(operationName);
+      if (await resp.url()?.includes("/graphql")) {
+        const reqJson = await resp.request()?.postDataJSON();
+        return reqJson.operationName === operationName;
       }
 
-      return isGraphql && isOperation;
+      return false;
     });
   }
 
--- a/apps/playwright/playwright.config.ts
+++ b/apps/playwright/playwright.config.ts
@@ -34,7 +34,8 @@ export default defineConfig({
   projects: [
     {
       name: "setup",
-      testMatch: /..\/global.setup\.ts/,
+      testDir: ".",
+      testMatch: /global.setup\.ts/,
     },
 
     {
--- a/apps/playwright/playwright.config.ts
+++ b/apps/playwright/playwright.config.ts
@@ -1,0 +1,85 @@
+import { defineConfig, devices } from "@playwright/test";
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// require('dotenv').config();
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default defineConfig({
+  testDir: "./tests",
+  /* Run tests in files in parallel */
+  fullyParallel: true,
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 1 : undefined,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: "html",
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    baseURL: "http://localhost:8000",
+
+    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+    trace: "on-first-retry",
+  },
+
+  /* Configure projects for major browsers */
+  projects: [
+    {
+      name: "setup",
+      testMatch: /..\/global.setup\.ts/,
+    },
+
+    {
+      name: "chromium",
+      use: { ...devices["Desktop Chrome"] },
+      dependencies: ["setup"],
+    },
+
+    {
+      name: "firefox",
+      use: { ...devices["Desktop Firefox"] },
+      dependencies: ["setup"],
+    },
+
+    {
+      name: "webkit",
+      use: { ...devices["Desktop Safari"] },
+      dependencies: ["setup"],
+    },
+
+    /* Test against mobile viewports. */
+    // {
+    //   name: 'Mobile Chrome',
+    //   use: { ...devices['Pixel 5'] },
+    // },
+    // {
+    //   name: 'Mobile Safari',
+    //   use: { ...devices['iPhone 12'] },
+    // },
+
+    /* Test against branded browsers. */
+    // {
+    //   name: 'Microsoft Edge',
+    //   use: { ...devices['Desktop Edge'], channel: 'msedge' },
+    // },
+    // {
+    //   name: 'Google Chrome',
+    //   use: { ...devices['Desktop Chrome'], channel: 'chrome' },
+    // },
+  ],
+
+  /* Run your local dev server before starting the tests */
+  // webServer: {
+  //   command: 'npm run start',
+  //   url: 'http://127.0.0.1:3000',
+  //   reuseExistingServer: !process.env.CI,
+  // },
+});
--- a/apps/playwright/tests/admin/auth.spec.ts
+++ b/apps/playwright/tests/admin/auth.spec.ts
@@ -1,53 +1,166 @@
 import { Page } from "@playwright/test";
 
 import { test, expect } from "~/fixtures";
-import auth from "../../constants/auth";
+import auth from "~/constants/auth";
+import { getAuthCookies, getAuthTokens, AuthCookies } from "~/utils/auth";
 
-const onLoginInfoPage = async (page: Page) => {
-  await page.waitForURL("**/login-info*");
-  await expect(page.url()).toContain("/en/login-info");
-};
+function expectAuthCookies(cookies: AuthCookies) {
+  expect(cookies).toEqual(
+    expect.objectContaining({
+      apiSession: expect.objectContaining({ name: "api_session" }),
+      xsrf: expect.objectContaining({ name: "XSRF-TOKEN" }),
+    }),
+  );
+}
 
-test.describe("Auth flows (development)", () => {
-  test.describe("Anonymous user", () => {
-    test("redirects restricted pages to sign in", async ({ browser }) => {
-      const pages = [
-        "/en/admin/dashboard",
-        "/en/admin/talent-requests",
-        "/en/admin/users",
-        "/en/admin/settings/classifications",
-        "/en/admin/pools",
-        "/en/admin/settings/departments",
-        "/en/admin/settings/skills",
-        "/en/admin/settings/skills/families",
-      ];
+const restrictedPaths = [
+  "/en/admin/dashboard",
+  "/en/admin/talent-requests",
+  "/en/admin/users",
+  "/en/admin/settings/classifications",
+  "/en/admin/pools",
+  "/en/admin/settings/departments",
+  "/en/admin/settings/skills",
+  "/en/admin/settings/skills/families",
+];
 
+test.describe("Anonymous user", () => {
+  test("Redirects restricted pages to sign in", async ({ browser }) => {
+    await Promise.all(
+      restrictedPaths.map(async (restrictedPath) => {
+        const context = await browser.newContext();
+        const page = await context.newPage();
+        await page.goto(restrictedPath);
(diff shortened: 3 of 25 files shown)


brindasasi:
This is cool. Love the parallel test runs. 
Debug mode opens up stack of windows at a time and lots of runners.
Do you know how to run single test or single suit?

brindasasi:
seems like we can automate this in Azure pipeline  as well 🥳
https://techcommunity.microsoft.com/t5/azure-architecture-blog/automating-end-to-end-testing-with-playwright-and-azure/ba-p/3883704

esizer:
> Do you know how to run single test or single suit?

Yeah, there should be accordions and you can expand them and press the play button. 

test-runner

brindasasi:
This looks good to me and my only concern is the effort for migration of all existing test cases into this. 
As Tristan said , probably we could start with new test cases run both and then slowly migrate & get rid of cypress. 

Also did you check how do we do feature flags in here? 

I would leave this here for other team members review & comments.

apps/playwright/playwright.config.ts
```diff
@@ -0,0 +1,86 @@
...
+    // {
+    //   name: 'Mobile Chrome',
+    //   use: { ...devices['Pixel 5'] },
+    // },
+    // {
+    //   name: 'Mobile Safari',
+    //   use: { ...devices['iPhone 12'] },
+    // },
+
+    /* Test against branded browsers. */
```
brindasasi, line 69: So far Edge is our top most used browser. We should keep it.

esizer:
> Also did you check how do we do feature flags in here?

Not yet but that does seem important. I can't imagine we will have issues so maybe I can take a look today? It looks like they have some pretty good documentation on [how to modify responses.](https://playwright.dev/docs/network#modify-responses)

apps/playwright/utils/user.ts
```diff
@@ -0,0 +1,58 @@
+export const defaultUser = {
+  // required
+  firstName: "Playwright",
+  lastName: "User",
+  preferredLang: "EN",
+  preferredLanguageForInterview: "EN",
+  preferredLanguageForExam: "EN",
+
+  // optional
+  telephone: undefined,
```
petertgiles, line 10: I think this stuff is a hangover from when Cypress was in plain JS.  It'd probably be better to just import the User type from graphql and type this object.

apps/playwright/playwright.config.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default defineConfig({
+  testDir: "./tests",
+  /* Run tests in files in parallel */
+  fullyParallel: true,
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
```
petertgiles, line 17: You don't anticipate any trouble running this in a GH workflow, right?

apps/playwright/playwright.config.ts
petertgiles, line 1: Is there Axe integration?

petertgiles (commented):
So far I like what I see!  The fixtures system looks cool but it's going to take a bit of head-wrapping since we never touch classes in JS. 

I think I need to mess around with this a bit more on Linux.  Most (but not all!) the tests failed for me.
image
image

apps/playwright/utils/user.ts
esizer: Oh, thanks! I was kind of just copy and pasting a lot of stuff from Cypress to this so that is good to know 👍

esizer:
> I think I need to mess around with this a bit more on Linux. Most (but not all!) the tests failed for me.

Oh shoot, I should have mentioned in the testing steps that would may need to run `npx playwright install-deps`. I also ran into this issue on WSL and that fixed it for me 👍

apps/playwright/playwright.config.ts
esizer: Yeah, so my intention was to do the minimum amount of work to see what people thought of the style of writing tests before exploring it further.

 It seems like the ones who have reviewed it are tentatively positive about it so we may be able to start filing in some gaps like the CI thing. Do you think it is worth exploring that now? This is obviously a requirement and they have some pretty good [documentation around it ](https://playwright.dev/docs/ci-intro)(one of the things I enjoyed about working with it).

apps/playwright/playwright.config.ts
esizer: Yup! There is an [axe package](https://www.npmjs.com/package/@axe-core/playwright) for it and playwright does a good job of [documenting how to use it](https://playwright.dev/docs/accessibility-testing) 😄

petertgiles: Nice!

apps/playwright/playwright.config.ts
petertgiles: No, that's fine.  It was a reasonable scope limit.  As long as it's well documented, that's enough for me right now.

apps/playwright/playwright.config.ts
petertgiles: I'm guessing this is disabled by default because it is assumed that Edge will behave exactly the same as Chromium.  Reasonable assumption?

esizer: Yeah, that is what I gathered as well.

esizer:
> Unfortunately it was still not able to find some packages. Bummer, it seems like Playwright is only supported on LTS versions of Ubuntu. We could probably run the CLI version in Docker but that negates the nice debugging UI, right?

oh darn, I forgot you were not on an LTS 😢 Yeah, we may need to run in docker then which may make the debugging non-functional. We could try to map the ports it uses 🤔 Though, I'm not super optimistic about that 👎

yonikid15 (commented):
Wow looks good! The code was easy to follow, and I'm excited to try it out myself. Reviewing the classes gave me a flashback to when I working with Java 😆.

I was able to run the tests headless, using the UI, and debug mode. However, some of the tests are failing on my end (still looking into it) 😢. I'll post screenshots just incase someone else experienced something similar.

Headless:
image

image

Debug mode:
Screenshot 2024-01-12 140838

esizer:
> I was able to run the tests headless, using the UI, and debug mode. However, some of the tests are failing on my end (still looking into it) 😢. I'll post screenshots just incase someone else experienced something similar.

I wasn't able to reproduce that exact error but I put some stuff in to slow them down. I think they were running too fast so the elements were hidden behind the loading spinner or something. Also noticed the skill library test name was not accurate 😛 

[improve tests, correct names](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8917/commits/a298357f19958b444c333d5f1b83f61cdd8813e6)

petertgiles:
I tried to get it running in our docker environment to see if I could get it going on my workstation.  Unfortunately I just got timeouts.  :cry: 
image
image
image

yonikid15:
> I tried to get it running in our docker environment to see if I could get it going on my workstation. Unfortunately I just got timeouts. 😢 ![image](https://private-user-images.githubusercontent.com/8978655/296797936-953a0122-33cb-4850-9f70-b996e9274af4.png?

I'm still seeing the same as Peter unfortunately 😭

yonikid15:
@esizer Okay so looks like the timeout changes and moving the `toBeVisible()` check fixed all the tests.
As requested, I attached a snapshot of all the tests and the duration for each browser. Here are the longest tests that would benefit from more fine grained timeout increases vs just a global increase:

1. Can submit application (~4min)
image

2. Skill library (~30sec)
image

3. Cannot access restricted paths (admin ~50sec + applicant ~50sec)
image

4. Can logout (~1min)
image

5. Redirects restricted pages to sign in (~1min)
image

yonikid15:
Here are list of the assertions that timeout:

## Authenticated:
image

image

image

image

image

## Skill Library

image

image

image

image

## Application

All browsers fail at same point (`Cannot read properties of null (reading 'id')` at u.id)
image

vd1992:
Trying it out
Running `npm run e2e:playwright`

Results: 

image

image

esizer:
> Trying it out

That one seems to be very troublesome 😛 There may be something wrong with the logic in it causing contexts to be closed too early 🤔 .

vd1992:
Trying the UI bit and it seems to work? Two tests appear but a little hard to follow what happened. Appears to flash through and you have to click on the frames in the timeline to see what was rendered? 

> > Trying it out
> 
> That one seems to be very troublesome 😛 There may be something wrong with the logic in it causing contexts to be closed too early 🤔 .

Reran and got basically an identical result with a slight runtime difference
Expanded the sections out

image

esizer:
> Here are list of the assertions that timeout:

Okay, I updated the timeout to be a little more reasonable (30 seconds). As well, I updated the restricted paths tests to wait for the URL instead of network idle. Hopefully this helps 🤷‍♀️ 

 - [tweak timeout to use lower value](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8917/commits/463429acad31d9f2b60513943d9429812c372508)
 - [improve waits for auth tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8917/commits/5dc9c85b9dfe6dc90a15cef5e101a53edbf5d623)

vd1992:
Got 5-6 tests failing now

Last run 

image

Application submit all fail for the same reason

image

And the auth fails are

image

image

vd1992:
For this run, noticed this little problem with it not failing correctly

https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/7644060225/job/20827486364

image

esizer:
> And the auth fails are

Do your pages typically take a long time to load on your local machine? That was the case with @yonikid15. Setting the timeout to 30s caused timeouts. I just bumped it to 60s ([increase timeout to 60 seconds](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8917/commits/a1b55d95dd7fc5b4256d06bbf4ce5ed5408d95bf)) which got them to consistently pass for him.

> For this run, noticed this little problem with it not failing correctly

Yeah, I noticed that. Trying to get it to fail properly in the CI right now 🤞

vd1992:
> > And the auth fails are
> 
> Do your pages typically take a long time to load on your local machine? That was the case with @yonikid15. Setting the timeout to 30s caused timeouts. I just bumped it to 60s ([increase timeout to 60 seconds](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8917/commits/a1b55d95dd7fc5b4256d06bbf4ce5ed5408d95bf)) which got them to consistently pass for him.
> 
> > For this run, noticed this little problem with it not failing correctly
> 
> Yeah, I noticed that. Trying to get it to fail properly in the CI right now 🤞

Not that long a load time locally. 
Reran, other than the Typescript error, this one persists

image

esizer:
> For this run, noticed this little problem with it not failing correctly

Got it to fail properly 😆 https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/7645913027/job/20833569267?pr=8917#step:9:241

> Not that long a load time locally.
> Reran, other than the TypeError, this one persists

Would you be willing to pair up so I can watch your screen and try and debug this problem in real time? I'm having trouble reproducing/debugging it 😢

vd1992:
Running that one test by itself in UI mode and it appears to succeed?

vd1992:
Running it a few times for each browser, that was a handy addition, chromium and firefox work. Webkit fails on "Can access restricted paths" only, but this test passes when I run it singled out in UI mode.  

Just putting that there as a point of comparison for others trying it out.

mnigh (approved):
❤️ this. 👋 🇨🇾!

yonikid15 (approved):

apps/playwright/fixtures/AppPage.ts
```diff
@@ -0,0 +1,86 @@
...
+   * GraphQL Request
+   *
+   * Make a GraphQL request using the current page context
+   *
+   * @param query
+   * @param variables
+   * @returns
+   */
+  async graphqlRequest(query: string, variables?: Record<string, unknown>) {
+    await this.gotoHome();
```
tristan-orourke, line 30: Why is this part necessary? I would think this step would just slow down graphql requests. If its to make sure the authoirzationQuery has run, would it be possible to try to get auth tokens, and if they don't exist yet, _then_ goToHome?

apps/playwright/tests/admin/auth.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    await page.goto("/login");
+    cookies = await getAuthCookies(page);
+    expectAuthCookies(cookies);
+  });
+});
+
+test.describe("Authenticated", () => {
+  test("Has tokens", async ({ applicantPage }) => {
+    await applicantPage.gotoHome();
```
tristan-orourke, line 77: So if you use the applicantPage fixture, it always comes authenticated? That's cool! Where does it actually happen? `fixtures/ApplicantPage.ts` doesn't look like it does anything...

tristan-orourke (commented):
Overall, the ways this is set up, with fixtures and util functions, makes a lot of sense to me! I did leave some questions about things I'd like to understand better.

apps/playwright/utils/graphql.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Make a GraphQL request using the
+ * admin auth state by default.
+ */
+export async function graphqlRequest(
+  query: string,
+  variables?: Record<string, unknown>,
+  storageState?: string,
+) {
+  const browser = await chromium.launch();
```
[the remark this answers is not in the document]
esizer: Yeah, that is a good thing to bring up. Originally, that was how things were happening and I really do agree that is how we should be doing it. This was an easy fix to get the auto refresh of our client since the access token is short lived and usually expires before the test is done. 

I should have written something about this so we could fix it but this was the easy fix. I think we could write something into the fixture to intercept refresh requests and update the auth storage state so it is always up to date. There are other ways but this is likely the best solution.

apps/playwright/fixtures/AppPage.ts
esizer: This is related to the other comment...sort of. So, our tokens were expiring before the test ended and this request was actually happening outside of the browser. So this just goes home, waits for the token to be refreshed and then sends off the request. I think we can make this better in one of three ways:

1. As mentioned in other comment, intercept refreshes in the in the fixture and keep our storage state up to date (probably best?)
2. Right a refresh to use here and in the other graphql utility to refresh if the token is expired. Basically, mimic what our client provider is doing
3. Just be lazy and extend our dev tokens to something like 30mins so this is unnecessary 😅 

If you'd like, I could see about getting one of these solutions working here before merging.

apps/playwright/tests/admin/auth.spec.ts
esizer: Yeah! I really liked this. We can also do other roles too! It happens in the `global.setup.ts` file before everything runs.

We log the user in then it takes a snapshot of the current app state (cookies, etc.) and stores it in a JSON file under `.auth`. Then when you use this fixture, it loads that into the browser. As mentioned before, it gets stale during testing since this seem to only happen once. I offered some solutions to attempt to keep this file up to date though.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/95c8c0d584e44f6f96dba759e8c6388b6137fd1d/apps/playwright/global.setup.ts#L6-L14

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,100 @@
...
+      - name: Run Firefox Tests
+        run: npm run e2e:playwright:firefox
+      - uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: firefox-report
+          path: apps/playwright/playwright-report/
+          retention-days: 30
+
+      - name: Run Webkit Tests
```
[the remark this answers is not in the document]
esizer: It does not, you can do that though by [passing the workers flag and setting it to 1](https://playwright.dev/docs/test-parallel#disable-parallelism).

I added the browser specific commands so we could still run all the tests but reduce the run time significantly.

tristan-orourke:
Decide whether to show errors in files changed tab
Show report on error
Remove the overlapping cypress test

apps/playwright/playwright.config.ts
```diff
@@ -0,0 +1,90 @@
...
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 1 : undefined,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: process.env.CI
+    ? [["github"], ["dot"]]
+    : [["line"], ["html", { open: "on-failure" }]],
+  timeout: 5 * 60 * 1000, // 5 mins
```
mnigh, line 26: Can we removed this?

apps/playwright/playwright.config.ts
```diff
@@ -0,0 +1,90 @@
...
+  expect: { timeout: 3 * 60 * 1000 }, // 3 mins
```
mnigh, line 27: Can we removed this?

tristan-orourke (approved):
🎉 🎉 🎉

PR #8918, opened 2024-01-02
[Fix] Error summary second focus

### LINKED ISSUE #8846: 🐛 Submit does not navigate to error section every time
## 🐛 Bug

When there are errors in the form and the form is submitted, the window should navigate to the error section every time the user submits.

## 🦋 Expected Behaviour

If there are errors in the form, the window should navigate to the error section every time.

## 📋 Steps to Reproduce

1. Navigate to career timeline
2. Add a new work experience
3. Fill out all required fields except **Additional details**
4. Enter **Save and return to my career timeline**
5. Observe the scroll
6. Fill out all fields except My role
7. Enter **Save and return to my career timeline**
8. Observe failure to scroll to error section

## 🎥 Screen recording

https://github.com/GCTC-NTGC/gc-digital-talent/assets/3046459/fa56cb91-2681-45b5-b5ad-34b5d53c1d92

## ✅ Acceptance Criteria

- [ ] when there are errors in the form, the window navigates to the error section every time

esizer:
🤖 Resolves #8846 

## 👋 Introduction

Fixes the focus of an error summary on a second attempt to submit.

## 🕵️ Details

The values we were using the effect dependencies array were not primitives. This meant the callback was never re-fired. This updates it to use the flattened array of input IDs which is primitive. This allows the callback to be called when it updates meaning we can focus the summary every time errors change on submit.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to a basic form that has labels (profile forms work well here)
3. Submit with errors, confirming the summary gets focus
4. Fix on error or introduce more
5. Re submit, confirming the summary is once again focused
6. Navigate to the experience form
7. Repeat steps 3-5

### DIFF
--- a/packages/forms/src/components/BasicForm.tsx
+++ b/packages/forms/src/components/BasicForm.tsx
@@ -17,6 +17,7 @@ import {
 
 import ErrorSummary from "./ErrorSummary";
 import UnsavedChanges from "./UnsavedChanges";
+import { flattenErrors } from "../utils";
 
 export type FieldLabels = Record<string, React.ReactNode>;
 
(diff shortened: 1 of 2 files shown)


packages/forms/src/components/BasicForm.tsx
```diff
@@ -98,10 +87,22 @@ function BasicForm<TFieldValues extends FieldValues>({
...
+  const handleInvalidSubmit: SubmitErrorHandler<TFieldValues> = (errors) => {
+    const flatErrors = flattenErrors(errors);
+    setShowErrorSummary(flatErrors.length > 0);
+
+    errorSummaryRef.current?.focus();
+  };
+
+  React.useEffect(() => {
+    errorSummaryRef.current?.focus();
+  }, [showErrorSummary]);
```
tristan-orourke, line 104: Why do we do it this way, where we set `showErrorSummary` state, and wait for that to trigger a useEffect hook? Why not replace `setShowErrorSummary(true)` with `errorSummaryRef.current?.focus();`? 

(Or more likely, with a `focusErrorSummary` function which does the focusing.)

tristan-orourke: Nevermind. I tried removing this useEffect since I thought this would work without it, but I see its necessary for focusing to work the first time.

tristan-orourke (approved):
Looks good and works as promised 👍

packages/forms/src/components/BasicForm.tsx
esizer: Yeah. I tried really hard to not need it but sadly it seems like it is required 😓

PR #8920, opened 2024-01-03
[Fix] Update skill showcase without refresh

### LINKED ISSUE #8814: 🐛 Skill rankings requires refresh to show changes
## 🐛 Bug

The pages related to ranking skills requires the user to refresh the page to see the items they have added.

## 🦋 Expected Behaviour

User should not need to refresh the page after adding a skill to the top/to improve skills pages.

## 🕵️ Details

This one seems a little odd and I can't quite tell what is causing it. It looks as though, when adding a new skill, we do two things:

1. Create/update the skill
2. Fire off a second mutation to update the ranking

I don't think that is the cause though. Since our repeater is a form field, but does not act like a form field here (something that we should stop doing or create a separate repeater as previously talked about) it never gets updated.

## 📋 Steps to Reproduce

1. Login
2. Navigate to `/applicant/profile-and-applications/skills/showcase/top-5-behavioural-skills`
3. Add a skill
4. Notice that the repeater is not updated with the new skill

## 📸 Recording

https://github.com/GCTC-NTGC/gc-digital-talent/assets/4127998/ba18b1f5-4741-4a69-8053-f20cbc5d8a2b

## 🙋‍♀️ Proposed Solution

I feel like we have two options, likely the "easy" option makes most sense but we should consider the more complete one 🤔 

1. **Easy**: Use the `react-hook-form` `setValue` function to update the skills when the API cache changes
2. **Complete**: Create a non-form version of the repeater so we don't need to hack around the fact that we are not using it like a form input

## ✅ Acceptance Criteria

- [ ] Adding a skill to your top/to improve skills appears without the need for a refresh
- [ ] Write a cypress test to confirm this previously mentioned behaviour
- [ ] Implement using the "easy" option for now.

esizer:
🤖 Resolves #8814 

## 👋 Introduction

Quick (not great) fix for for the skills showcase pages not updating without a full page refresh.

## 🕵️ Details

This uses the `useFieldArray` function `append` to add the skill after the mutation is successful. It is a little janky but works for our purposes until we can revisit the functionality for this type of repeatable item.

## 🧪 Testing

1. Build `npm run dev`
2. Login as an applicant and navigate to a skill showcase `/applicant/profile-and-applications/skills/showcase/*`
3. Add a skill in your library
4. Confirm it is added to the list without a refresh
5. Add a skill not in your library
6. Confirm it is added to the list without a refresh
7. Change the order
8. Save and confirm the expected values were saved

### DIFF
--- a/apps/web/src/pages/Skills/components/UpdateSkillShowcase.tsx
+++ b/apps/web/src/pages/Skills/components/UpdateSkillShowcase.tsx
@@ -81,7 +81,7 @@ const UpdateSkillShowcase = ({
     defaultValues: initialSkills,
   });
   const { control, watch, formState } = methods;
-  const { remove, move, fields } = useFieldArray({
+  const { remove, move, fields, append } = useFieldArray({
     control,
     name: "userSkills",
   });
@@ -134,6 +134,9 @@ const UpdateSkillShowcase = ({
         .then((res) => {
           handleSuccess();
           if (res.data?.updateUserSkill?.skill.id) {
+            append({
+              skill: res.data.updateUserSkill.skill.id,
+            });
             // having claimed a user skill in the modal and the mutation successful, update the ranking
             onAddition(
               existingSkillsRankingFiltered,
@@ -154,6 +157,9 @@ const UpdateSkillShowcase = ({
         .then((res) => {
           handleSuccess();
           if (res.data?.createUserSkill?.skill.id) {
+            append({
+              skill: res.data.createUserSkill.skill.id,
+            });
             onAddition(
               existingSkillsRankingFiltered,
               res.data.createUserSkill.skill.id,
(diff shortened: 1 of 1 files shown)


brindasasi:
When adding a skill , the skill level doesn't show up on the page for the first time.
It is showing when you go back for edit.
 I don't think it is caused by this Pr. 
Just wondering if it is easy enough to fix it along with this.

Screenshot 2024-01-04 at 10 59 55 AM

esizer:
> When adding a skill , the skill level doesn't show up on the page for the first time.
> It is showing when you go back for edit.
> I don't think it is caused by this Pr.
> Just wondering if it is easy enough to fix it along with this.

Great catch! I forgot to also append the skill level (thought it was part of the skill itself). Fixed in [also append skill level](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8920/commits/4f8e1b0ebeadb378cb92d7ee533774150a3c9c91)

brindasasi (approved):
works great!🥳

PR #8924, opened 2024-01-03
[Feature] Hide pool filter on specific pool candidate page

### LINKED ISSUE #5321: Disable pool filter in filter dialog on all candidates page
## ✨ Feature

Allowing managers to add or remove pools in the pool filter on the pool candidates page doesn't really make sense, since you are viewing a specific pool. Disable the pool filter (multi select field) on the pool candidates filter dialog (eg. http://localhost:8000/en/admin/pools/{id}/pool-candidates). 

There should already be a prop called `isDisabled` on the `<MultiSelectField />` component which you can set to true.

## ✅ Acceptance Criteria

- [ ] Pool filter is disabled in PoolCandidateTableFilterDialog only on the "All Candidates" page (eg. http://localhost:8000/en/admin/pools/{id}/pool-candidates).
- [ ] Pool filter should **still be** usable on the Request page tables.
- [ ] Watch out for the /admin/pool-candidates page, do  not break it

esizer:
🤖 Resolves #5321 

## 👋 Introduction

Hides the pool filter when on the "view candidates" page for a specific pool.

## 🕵️ Details

Since the admin is within the specific context of a pool, it does not make sense to allow them to filter by different pools. We also now has the "all pool candidates" page that allows this. So, we are hiding this filter. In order to do this a few things were done:

- Added a hidden input component that is registered with `react-hook-form`
- Allow modifying the filter count on the filter dialog button so we can omit hidden filters from the count
   - Used addition here to make it clear what is happening. I currently can't think of a reason why we would want to increase the number right now but seeing  a negative integer makes more sense to me 🤷‍♀️ 

## 🧪 Testing

1. Build `npm run dev`
2. Login as an admin
4. Filter by a specific pool
5. Confirm the filter count is correct
6. Make note of the candidates who appear and how many
7. Navigate to `/admin/pools`
8. Find the pool you filtered by in step 4
9. Click "View candidates" in that row
10. Confirm the table candidates matches the one from step 6
11. Confirm the filter count matches the **visible** filters
12. Confirm filters and count function as expected still

## 📸 Screenshot

Screenshot 2024-01-03 112820

### DIFF
--- a/apps/web/src/components/FilterDialog/FilterDialog.tsx
+++ b/apps/web/src/components/FilterDialog/FilterDialog.tsx
@@ -29,6 +29,8 @@ type FilterDialogProps<TFieldValues extends FieldValues> = {
   resetValues: CommonFilterDialogProps<TFieldValues>["resetValues"];
   defaultOpen?: boolean;
   children: React.ReactNode;
+  /** Modify the filter count in the button (most commonly used for hidden filters) */
+  modifyFilterCount?: number;
 };
 
 const FilterDialog = <TFieldValues extends FieldValues>({
@@ -36,6 +38,7 @@ const FilterDialog = <TFieldValues extends FieldValues>({
   options,
   children,
   resetValues,
+  modifyFilterCount,
   defaultOpen = false,
 }: FilterDialogProps<TFieldValues>) => {
   const [isOpen, setIsOpen] = React.useState<boolean>(defaultOpen);
@@ -99,7 +102,7 @@ const FilterDialog = <TFieldValues extends FieldValues>({
           type="button"
           icon={AdjustmentsVerticalIcon}
           {...(filterCount > 0 && {
-            counter: filterCount,
+            counter: filterCount + (modifyFilterCount ?? 0),
           })}
         >
           {intl.formatMessage({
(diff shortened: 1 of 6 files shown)


vd1992:
While playing around with the filter, specifically the combobox inputs, I noticed this warning appearing regularly in the console. Not sure if it is a concern or not, digging through the references lines and I think it stems from combobox bundle and maybe something to do with react-fast-compare. 

image

apps/web/src/components/FilterDialog/FilterDialog.tsx
```diff
@@ -29,13 +29,16 @@ type FilterDialogProps<TFieldValues extends FieldValues> = {
   resetValues: CommonFilterDialogProps<TFieldValues>["resetValues"];
   defaultOpen?: boolean;
   children: React.ReactNode;
+  /** Modify the filter count in the button (most commonly used for hidden filters) */
+  modifyFilterCount?: number;
```
vd1992, line 33: Addition with negative integers also makes sense to me, I think this also looks understandable how to use it.

vd1992 (approved):
Think it all looks good!

PR #8925, opened 2024-01-03
[Fix] Allow equal start/end dates

### LINKED ISSUE #8919: Start and end month the same (for trainings for example) will give error
### User story
As an applicant, I want to be able to have an experience with the same start and end month, so I can share experiences (including trainings) that were of a short duration.

### Details
Our validation on experiences does not allow for the same start and end date. Now that we are only collecting start and end month, this is blocking users from adding experiences that started and ended in the same month. Instead we should allow the end date to equal the start date, but it still cannot occur before the start date.

### Acceptance Criteria
- [ ] Experience validation allows for **end date** `equal or greater than` **start date**. (instead of strictly greater)

esizer:
🤖 Resolves #8919 

## 👋 Introduction

Adds an option to allow a min/max date match the exact date specified rather than forcing it to be greater/less than.

## 🕵️ Details

Opted to go with a prop so we can control this. Since it seems to be specific to the experience date ranges (and one on the directive form) it is opt-in. Hope this makes sense 😅 

**Update:** After discussing with Tristan, he suggested we just make this the default to fall inline with how the rules with with text/number inputs from `react-hook-form` and not need to modify the rule definition.

```ts
min: {
  value: startDate,
   // allowEqual: true,
  message: intl.formatMessage(errorMessages.mustBeGreater, {  value: startDate  },
},
```

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to add an experience page (`/users/{userId}/personal-information/career-timeline/create`
3. Enter the same value for start and end date
4. Confirm it passes validation
5. Make the end date be before the start date
6. Confirm validation fails
7. Make the end date be after the start date
8. Confirm it passes
9. Create a pool from `/admin/pools/create`
10. Edit the end date
11. Set the expiry date to the current date
12. Confirm it fails validation
13. Set the expiry date to a date in the past
14. Confirm it fails validation
15. Make the expiry date in the future
16. Confirm it passes validation

## 📸 Screenshot

Screenshot 2024-01-03 142818

### DIFF
--- a/apps/web/src/components/ExperienceFormFields/AwardFields.tsx
+++ b/apps/web/src/components/ExperienceFormFields/AwardFields.tsx
@@ -126,7 +126,6 @@ const AwardFields = ({ labels }: SubExperienceFormProps) => {
               max: {
                 value: strToFormDate(todayDate.toISOString()),
                 message: intl.formatMessage(errorMessages.mustNotBeFuture),
-                allowEqual: true,
               },
             }}
           />
--- a/apps/web/src/components/ExperienceFormFields/AwardFields.tsx
+++ b/apps/web/src/components/ExperienceFormFields/AwardFields.tsx
@@ -126,6 +126,7 @@ const AwardFields = ({ labels }: SubExperienceFormProps) => {
               max: {
                 value: strToFormDate(todayDate.toISOString()),
                 message: intl.formatMessage(errorMessages.mustNotBeFuture),
+                allowEqual: true,
               },
             }}
           />
--- a/apps/web/src/components/ExperienceFormFields/CommunityFields.tsx
+++ b/apps/web/src/components/ExperienceFormFields/CommunityFields.tsx
@@ -118,6 +118,7 @@ const CommunityFields = ({ labels }: SubExperienceFormProps) => {
                       required: intl.formatMessage(errorMessages.required),
                       min: {
                         value: startDate,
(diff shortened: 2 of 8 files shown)


tristan-orourke (changes requested):
Yeah, after taking a look at this, I'd rather have `allow equal` be the default behaviour. That's how `max` works, and that's how `min` works for other Inputs besides DateInput.

For cases where we do want it require it to be greater than a value, could we create a new rule, `greaterThan`, instead of `min`?

esizer:
> Yeah, after taking a look at this, I'd rather have `allow equal` be the default behaviour. That's how `max` works, and that's how `min` works for other Inputs besides DateInput.

Makes sense to me 👍 Updated in [make allow equal default for min/max rules](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8925/commits/6863be08721d4f99f2d9d406fc523cbf17fd46ef)

> For cases where we do want it require it to be greater than a value, could we create a new rule, `greaterThan`, instead of `min`?

Is this something you'd like to see in this PR or something we create once we need it? I can't imagine we have too many use cases for it at the moment 🤷‍♀️ As you mentioned yesterday, being one day out for something like an expiry date is almost as odd as making it the same day anyway 😆

tristan-orourke (approved):
Okay, I looked through the app and I'm happy with this as is!

PR #8972, opened 2024-01-09
[Feature] Admin side menu and page titles update

### LINKED ISSUE #8175: ✨ Update the admin sidebar menu
image

## ✨ Summary

[Visit the Figma file](https://www.figma.com/file/guHeIIh8dqFVCks310Wv0G/Style-library?type=design&node-id=652-29771&mode=design) • Contact @substrae with design questions

As a part of the larger admin-facing UI refresh, this ticket aims to reorganize and update the admin sidebar menu to make it clearer and easier to use.

The pin and popover elements have been spun off into #8902.

```[tasklist]
### ✅ Requirements
- [ ] Update the icons for each item
- [ ] Reorganize the items into categories as depicted
- [ ] Update the active item color to `base(secondary.lighter)` and `base:iap(secondary.lightest)`
- [ ] Ensure the menu's width is fixed to match
- [ ] Ensure that the menu's state is saved in local memory so that users can return to the same state next time
- [ ] Ensure mobile menu opens from the left
- [ ] Ensure mobile menu overlay appears and uses `black.light.9`
- [ ] Ensure the mobile menu, while open, is scrollable if the vertical height exceeds the viewport available
- [ ] Ensure these changes don't affect display of items based on permissions/roles
``` 

### On landscape tablets, laptops, and desktops

- We're updating the information architecture by categorizing menu items and giving them distinct icons

<details>
  <summary>📸 Menu closed</summary>
  
</details> 

<details>
  <summary>📸 Menu open</summary>
  
</details> 

### On mobile devices and portrait tablets

Not much is changing on mobile in terms of behaviour - the look and feel should be reflected. The bar remains sticky at the top, and triggers the menu to open from the left.

<details>
  <summary>📸 Menu closed</summary>
  
</details> 

<details>
  <summary>📸 Menu open</summary>
  
</details>

petertgiles:
🤖 Resolves #8175 

## 👋 Introduction

This branch has the side menu refresh.  It also updates page titles to match.  Note that the pinning and popover were deferred to a later issue.

## 🕵️ Details

Most of the diffs in this branch are to factor out the page titles and icons to the page files and referencing them directly instead of copy/pasting the changes which can go out of sync.

## 🧪 Testing

Rebuild the app and log in as an admin

- [ ] Full view matches design
- [ ] Mobile view matches design
- [ ] Open/close works
- [ ] Sign out works
- [ ] Links lead to the right pages
- [ ] Links appear only when the roles match

## 📸 Screenshot

image


esizer (changes requested):
Code looks good, design looks good (to me at least). 

Only thing I noticed was there is a bit of a "jump" when the menu opens. This does not appear to exist on `main`. I have been investigating to provide you with more detail but I'm having trouble determining how it got introduced 😖

packages/ui/src/components/SideMenu/SideMenuCategory.tsx
```diff
@@ -0,0 +1,67 @@
...
+        data-h2-border-bottom="base(1px solid)"
+        data-h2-font-weight="base(700)"
+        data-h2-height="base(x1)"
+      >
+        <motion.span
+          transition={transitionConfig}
+          animate={ctx?.open ? { opacity: 1 } : { opacity: 0 }}
+        >
+          {!!ctx?.open && title}
+        </motion.span>
```
esizer, line 59: Of course, the second I gave up and just submitted a review I found where the issue was coming from 🤦‍♀️ .

So, it is coming from the title popping in and out. It looks like height was added the parent to avoid this but it isn't 100% exact so it only reduces the effect. I'm not sure if this is the best way, but the following should resolve it. If you don't want the title to be announced to ATs you can just toggle the `aria-hidden` from `"true"` to `undefined` here instead.

```suggestion
        <motion.span
          data-h2-display="base(flex)"
          transition={transitionConfig}
          animate={ctx?.open ? { opacity: 1, width: "auto" } : { opacity: 0, width: 0 }}
        >
          <span>
            {title}
          </span>
        </motion.span>
```

substrae:
@petertgiles 

This is looking awesome! A few small things:
* Section titles (e.g. Recruitment) should be `caption` sized
* The divider line below titles should have a `margin-top` of `x.15`
* It's possible that this wasn't already in the mobile menu experience, but ideally clicking the non-menu overlay space when the menu is open should close it

packages/ui/src/components/SideMenu/SideMenuCategory.tsx
petertgiles: Good catch!  I found that this change had a weird wrap for titles with more than one word while the width was animating.
https://github.com/GCTC-NTGC/gc-digital-talent/assets/8978655/45b6aeb3-3254-4b98-90b8-d88b81687fde
I added a "nowrap" to the span to avoid this.  Hopefully none of the French titles are long enough to require this wrap.  :crossed_fingers: 
[fix title pop](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8972/commits/1fa0035ffc1366d5de1087d443e9f2bbd98a52fa)

esizer: oh yeah, that is a good call! LGTM 👍

petertgiles:
@substrae Thanks for the review!  I updated the first two items.
[review styling updates](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8972/commits/5e96d0fb3c6a40e2701229301900edc0b608e626)

For the third one, I can confirm that this is unchanged from the way it is now.  I checked, though, and "Ensure user can tap the overlay to close the mobile menu" is part of the AC for the popover and pinning issue.

esizer (approved):

PR #8975, opened 2024-01-10
[Debt] Remove applicant filters query

### LINKED ISSUE #7654: ♻️ Remove deprecated `applicantFilters` query
## ♻️ Debt/Refactor

This query is deprecated.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/4000aacaaf392bc2a0b763d4d9dbd9f67cb2680f/api/graphql/schema.graphql#L987-L994

## ✅ Acceptance Criteria

- [ ] `applicantFilters` query removed from schema file
- [ ] References to this query have been removed or replaced

yonikid15:
🤖 Resolves #7654 

## 👋 Introduction

- Removes `applicantFilter` query
- Updates tests

## 🧪 Testing

1. Ensure site works correctly and all tests pass.


api/tests/Feature/ApplicantFilterTest.php
```diff
@@ -98,73 +100,6 @@ protected function filterToCreateInput(ApplicantFilter $filter)
         return $input;
     }
 
-    /**
```
petertgiles, line 101: Is it worth repointing these tests at `poolCandidateSearchRequest.applicantFilter` instead of removing them?  Or am I misunderstanding how this works?

yonikid15: I noticed that the test `testFilterCanBeStoredAndRetrievedWithoutChangingResults()` was querying an applicant filter so I thought why not move the test there and save some space. However, I think your right looking back at it now since it's querying for all the fields which makes for a stronger test. I'll make the changes now 👍

petertgiles (approved):
Looks great!

PR #8976, opened 2024-01-10
[a11y] Increase touch target for checkbox/radio buttons

### LINKED ISSUE #8965: Maximizing clickable area for checkboxes and radio buttons across platform
## 🕵️‍♀️ Details (updated for dev)

Bigger clickable ares are easier to click, it's as simple as that - and it can make a big different for the ease of use for some users.

This issue aims to make the clickable area as big as possible for checkboxes and radio buttons with the existing component size. 

From Josh's proposed solution below:
> in order to improve the target area for checkboxes and radio circle specifically, we can apply a custom design that converts the existing whitespace to clickable area, expanding its target size. If we take the custom approach, we can also increase the checkbox/radio option height and width to x.75, effectively increasing the total target area to 34px tall and 40.5px wide.

**Updated clickable area:**
Image

**True clickable area (including label):**
Image

**Designer**
@substrae 

💻 Related WCAG Criteria
•	2.5.5 Target Size: The size of the [target](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html#dfn-target) for [pointer inputs](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html#dfn-pointer-input) is at least 44 by 44 [CSS pixels](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html#dfn-css-pixel).

## Acceptance criteria

- [ ] Line spacing converted to clickable area (see comments below)
- [ ] Padding around checkbox group converted to clickable area (see comments below)
- [ ] Visual appearance of radio button circle and checkbox box (height and width) increased to x.75
- [ ] Review from @substrae

esizer:
🤖 Resolves #8965 

## 👋 Introduction

Increases the total touch target of checkboxes and radio buttons. Also, improves the styling to better match our current design system.

## 🧪 Testing

>[!NOTE]
> This was tested in safari and working as expected.

1. Build `npm run dev`
2. Navigate to a form or, start storybook
3. Ensure the touch target is expanded in check lists and radio groups
4. Ensure they still function as expected

## 📸 Screenshot

Screenshot 2024-01-10 113838
Screenshot 2024-01-10 114044

### DIFF
--- a/packages/forms/src/components/RadioGroup/RadioGroup.tsx
+++ b/packages/forms/src/components/RadioGroup/RadioGroup.tsx
@@ -132,6 +132,7 @@ const RadioGroup = ({
                 >
                   <Field.Label
                     key={value}
+                    data-h2-cursor="base(pointer)"
                     data-h2-font-size="base(copy)"
                     data-h2-display="base(flex)"
                     data-h2-align-items="base(flex-start)"
--- a/packages/forms/src/components/RadioGroup/RadioGroup.tsx
+++ b/packages/forms/src/components/RadioGroup/RadioGroup.tsx
@@ -1,6 +1,7 @@
 import * as React from "react";
 import get from "lodash/get";
 import { FieldError, useFormContext } from "react-hook-form";
+import { useReducedMotion } from "framer-motion";
 
 import Field from "../Field";
 import type { CommonInputProps, HTMLFieldsetProps } from "../../types";
@@ -87,7 +88,8 @@ const RadioGroup = ({
   // To grab errors in nested objects we need to use lodash's get helper.
   const error = get(errors, name)?.message as FieldError;
   const baseStyles = useInputStyles();
-  const baseRadioStyles = getCheckboxRadioStyles();
+  const shouldReduceMotion = useReducedMotion();
+  const baseRadioStyles = getCheckboxRadioStyles(shouldReduceMotion);
   const stateStyles = useFieldStateStyles(name, !trackUnsaved);
   const fieldState = useFieldState(name, !trackUnsaved);
   const isUnsaved = fieldState === "dirty" && trackUnsaved;
--- a/packages/forms/src/components/RadioGroup/RadioGroup.tsx
+++ b/packages/forms/src/components/RadioGroup/RadioGroup.tsx
@@ -111,7 +111,10 @@ const RadioGroup = ({
         {...rest}
       >
         <Field.Legend required={!!rules.required}>{legend}</Field.Legend>
-        <Field.BoundingBox {...{ ...baseStyles, ...stateStyles }}>
+        <Field.BoundingBox
+          {...{ ...baseStyles, ...stateStyles }}
+          data-h2-padding="base(x.25 0)"
+        >
           <div
             data-h2-display="base(grid)"
             data-h2-gap="base(x.25)"
@@ -130,7 +133,7 @@ const RadioGroup = ({
                     data-h2-font-size="base(copy)"
                     data-h2-display="base(flex)"
                     data-h2-align-items="base(flex-start)"
-                    data-h2-gap="base(0 x.25)"
+                    data-h2-padding="base(x.25 x.5)"
                   >
                     <input
                       id={id}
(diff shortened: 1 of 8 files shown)


packages/forms/src/components/RadioGroup/RadioGroup.tsx
petertgiles, line 1: I actually don't know if this is a problem or not.  This isn't a multiline label but a label with a description.  Shouldn't the description be indented the same amount?  This might be out of scope since it probably isn't an issue with the component itself.  The indenting wasn't perfectly equal before, either.
image

petertgiles, line 1: It's probably nothing we can control but the dots don't always end up perfectly in the center for some reason.  
You can see that there are three pixels above and five below, so it's visibly off-center.
image
Most of the time it's better, so it's probably a subpixel rendering thing.
image

Chrome does a little better but now it's offset to the left.
image

I guess this is what we have to deal with when we deviate with the built-in widgets.  Or maybe it's just Linux.  :shrug:

packages/forms/src/utils/getCheckboxRadioStyles.ts
```diff
@@ -0,0 +1,20 @@
...
+  "data-h2-content": "base:selectors[::before]('')",
+  "data-h2-display": "base(grid)",
+  "data-h2-flex-shrink": "base(0)",
+  "data-h2-place-content": "base(center)",
+  "data-h2-height": "base(x.9) base:selectors[::before](x.5)",
+  "data-h2-line-height": "base(x1)",
+  "data-h2-margin": "base(x.1 0 0 0)",
+  "data-h2-transform":
+    "base:selectors[::before](scale(0)) base:selectors[:checked::before](scale(1))",
+  "data-h2-transition": "base:selectors[::before](120ms transform ease-in-out)",
```
petertgiles, line 16: Will this respect "prefer reduced motion"?

petertgiles (approved):
Wow!  Looks super nice.  :artist:

packages/forms/src/components/RadioGroup/RadioGroup.tsx
esizer: I'm going to guess this has to do with anti-aliasing. I'll take a look and see if I can reduce the offset a bit 👌

esizer: Okay, so it looks as though these all had custom spacing. I went and shifted those styles to be part of the base component and adjusted the spacing to match in [fix spacing for label, content below](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8976/commits/d7b90ff90d7418014e99df494e322fb7ef7aa1c9).

Also, a story was added so we can track this in chromatic as well.

esizer: Yeah, it was sub-pixels. I slightly increased the inputs to round off those pixels which should help avoid this.

[fix spacing for label, content below](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8976/commits/d7b90ff90d7418014e99df494e322fb7ef7aa1c9)

packages/forms/src/utils/getCheckboxRadioStyles.ts
esizer: Ooo, good call! I don't think it was. It should be now after [check reduced motion for animations](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8976/commits/ace246226fd7f8ba15ac6d59806c9d1a497aa4ef)

packages/forms/src/components/RadioGroup/RadioGroup.tsx
petertgiles: Nice!  This does look better.

packages/forms/src/utils/getCheckboxRadioStyles.ts
petertgiles: Great!

packages/forms/src/components/RadioGroup/RadioGroup.tsx
petertgiles: Nice!  It looks similar to before now.
image

petertgiles (approved):

substrae (approved):
SO so snazzy. Love it. My only thought was whether we should be using `cursor: pointer` on the options in both cases. It felt kinda weird to me when the cursor didn't change.

PR #9014, opened 2024-01-12
[Fix] Employment equity filters

### LINKED ISSUE #9010: Prod bug: EE Filters work one at a time, but not combined
## 🐛 Bug
In production, the EE filters seem to give inconsistent results whether one EE group is selected, or multiple.

## Expected behaviour
When multiple EE filters are selected, the result should be the union of the results for the individual EE groups.

## Steps to reproduce

1. Go to pool candidates page for any pool
2. Use filters to view only candidates in a single EE group (persons with disabilities for example)
3. See the total number of candidates and note a few of their names
4. Add all the other EE groups to the filter
5. You should have more candidates in your results now, but check to see if all the names you take note of are there.
6. In our tests, there were not all there

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] All individual EE filters work correctly
- [ ] Combining multiple EE filters returns the full union of all candidates returned by the filters individually
- [ ] Add php unit test to ensure query returns correct data
- [ ] Add jest test to ensure query built correctly

esizer:
🤖 Resolves #9010

## 👋 Introduction

Fixes the union (OR) function of our employment equity filters.

## 🕵️ Details

For the `isIndigenous` filter, we were using a `whereJsonLength`. Which means, when you included this filter with any of the others, it was looking for Indigenous and other equity groups. This has been updated to a `orWhereJsonLength` to treat them all like OR statements.

> [!NOTE]
> The functions we needed to do jest tests for were inside the component so this also needed to extract those out so they could actually be tested.

## 🧪 Testing

1. Confirm jest and PHP unit tests make sense and are passing
2. Build `npm run dev`
4. Open the filter dialog
5. Update the equity filter with different combinations and single selections
6. Confirm the users that appear have the expected equity declarations

### DIFF
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -662,7 +662,7 @@ class User extends Model implements Authenticatable, LaratrustUser
         $query->where(function ($query) use ($equityVars) {
             foreach ($equityVars as $index => $equityInstance) {
                 if ($equityInstance === 'is_indigenous') {
-                    $query->whereJsonLength('indigenous_communities', '>', 0);
+                    $query->orWhereJsonLength('indigenous_communities', '>', 0);
                 } else {
                     $query->orWhere($equityVars[$index], true);
                 }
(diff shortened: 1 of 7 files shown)


petertgiles (approved):
Nice.  :+1:   The new tests look great.

I tested with two pool candidates and the search seems to work well.  The query looks right, too.
image

PR #9015, opened 2024-01-15
[Fix] URL filter state not being read on initial render

### LINKED ISSUE #9013: 🐛 Table filters in url initially being applied
## 🐛 Bug

If you use the browser Back function to return to a table page, the filters and page number from when you were previously on the page will not be applied (even though they are visible in the url parameters).

## 🦋 Expected Behaviour

If you set table filters or change the page, and go to another page, then go "Back", the table should be in the state you left it.

## 📋 Steps to Reproduce

1. Log in as admin
2. Go to a pool candidates page
3. Change some filters and go to page 2
4. Click a candidate to go to the candidate page
5. Click "back" in the browser
6. The table should be on page 2 with the same filters applied, but it is not

## 📸 Screenshot

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] Table filters and page in the url parameters should be applied immediately when the page loads, if present
- [ ] Test solution in prod env

esizer:
🤖 Resolves #9013 

## 👋 Introduction

Fixes an issue where state for tables in the URL was not being applied on subsequent re-renders (navigating away and back).

## 🕵️ Details

We moved the initial getter out to stabilize some stuff with the URL. It seems like this was not needed (maybe fixed at some other point?). Since this was outside of the actual component, it only got called on initial loading. However, on subsequent re-renders (like navigating away and back with the same state) it never got called again so would only be reflected in a hard refresh.

Moving this call inside of the component is more expensive but forces it to be called on ever re-render which allows us to always have the most up to date initial state.

## 🧪 Testing

1. Build `npm run dev`
2. Login as admin navigate to `/admin/pool-candidates`
3. Apply some filters with the dialog
4. Change pages
5. Note the current list of users and state of the table
6. Navigate away from this page using an internal link on the page ("view profile" or "view application" both work well)
7. Click the back button to return to the page (do not hard refresh)
8. Confirm the current stable matches that in step 5


yonikid15 (approved):
Excellent 👍

PR #9021, opened 2024-01-15
[Fix] Tune PHP-FPM settings for web server

### LINKED ISSUE #8974: 🐛 Out of memory requests can make the server unresponsive
## 🐛 Bug

In some scenarios it is possible to make a request that runs out of memory and causes the web server to become unresponsive.

## 🦋 Expected Behaviour

The web server should be able to kill requests gracefully before it becomes unresponsive.

## 📋 Steps to Reproduce

> Attempting to download large numbers of candidates at a time is totally overwhelming the server :grimacing:
I tried loading 100 candidates at a time. Probably triggered a few requests - and not only did the requests fail, but then the entire site was stuck on a loading spinner for several minutes. Even the public homepage, and even on another device.
> It could also be that some users are writing a lot more or adding more skills than we've tested with
It seemed to be inconsistent; sometimes it worked with 100 candidates, sometimes not
I also wonder if it's only a problem when several requests hit the server at the same time

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

Tuning Laravel, FPM, Nginx, or the F5

## ✅ Acceptance Criteria

- [ ] The server can handle a reasonable number of "out of memory" requests simultaneously and remain responsive.

petertgiles:
🤖 Resolves #8974

## 👋 Introduction

This branch attempts to tune our PHP-FPM settings a bit to help prevent the server from becoming unresponsive.

## 🕵️ Details

PHP-FPM is the gateway used by Nginx to launch PHP processes.  When Nginx receives a request that needs to run PHP, it passes it off the PHP-FPM system which finds a worker job for it.  It manages a pool of worker jobs to balance quick response with resource usage.  When it runs out of workers it will queue up requests.

I've made a few changes:
- `listen.backlog = 128` Limit the number of requests that can get backed up in the queue.  I think it goes to 4096 by default on my system but it's not clear in the config file.  By limiting to 128 it will hopefully help the server recover faster from floods of requests.
- `pm.max_children = 10` How many workers can be spun up at the same time.  The config file help text notes 
"The below defaults are based on a server without much resources. Don't forget to tweak pm.* to fit your needs.". 
I've doubled it from 5 which is probably pretty conservative.  If the server stays happy after this change we might want to bump it again.  We have pages that will simultaneously fire 3 or 4 requests which can use up all almost all the workers by themselves!
- `pm.status_path = /fpm-status` I've enabled the FPM status endpoint so we can log into the server to check on it or maybe even use it for metrics in the future.  It will only be routable locally in production so it will stay secure.
- `request_terminate_timeout = 60s`  I found that requests could go past the 30 second PHP time limit if the PHP process was sleeping.  I think it's waiting for a database response when we see it go past 30 seconds in the logs.  This setting will cause FPM to kill the PHP process at 60 seconds as a safety belt for long-running queries.

As far as I can tell, our server is only a single core with 3.5 GB of RAM (virtually) so I don't want to go crazy with it yet.

## 🧪 Testing

Honestly, there's not a lot you can test locally unless you want to start making custom endpoints and setting up Locust.  :sweat_smile:   I'd love if @brindasasi could use her load-testing knowledge to validate that the new settings are at least no worse than before.

- "Down" then "Up" your Docker Compose network
- [ ] Can see the FPM status page at http://localhost:8000/fpm-status
- [ ] `listen queue len` is 128 

## 📸 Screenshot

image

## 🚚 Deployment

- Ensure the new PHP-FPM step is successful in the post-deploy script in Slack
- Ensure /fpm-status and /fpm-ping are NOT routable in your browser


brindasasi:
does listen.backlog means more than 128 requests at a time would result in closed connection error? 🤔 I hope not. I tried testing that with locust .
Simulated 500 concurrent users hitting on search page with 100 spawn rate. 
RPS went well above 128 and didn't find any failures which is good.
Screenshot 2024-01-25 at 10 07 51 AM

I also tried in Dev vertical with 100, 500 users candidates download , getting 500 error as expected but did not DDoS ourselves.

I want to check only 1 more than thing with the main branch to see the difference and then this will be good to go.

brindasasi (approved):
Tested with main branch .. I couldn't see big difference though. But the change itself seems to be working fine in Dev vertical with bulk data and under extreme pressure during load testing in local. 
Definitely worth the try 🥳

petertgiles:
> does listen.backlog means more than 128 requests at a time would result in closed connection error? 🤔 I hope not.

Sort of, yeah.  There can be up to 10 workers completing requests from the queue.  If the traffic increases beyond the rate that those 10 workers can satisfy then up to 128 requests will be queued.  After that, I assume HTTP 500s or something.  My intention there is to limit how far behind the server can fall and how long it will take to recover.  If we get a traffic flood we'll still be unable to respond to requests but I don't want to queue up, like, a billion requests that then take 24 hours to clear once the flood stops.

It really depends on how quickly the users are making requests and how long they take. If the 10 server workers can complete requests in 50 ms each and users are making one request per second then in theory we could support a sustained 200 requests per second?

Issue #8644, opened 2023-11-30
✨ Update Assessment Decisions


petertgiles:
## ✨ Feature

We need to update the assessment decisions that are possible with our application.  We are removing NOT_SURE and adding HOLD.

## 🕵️ Details

NOT_SURE is being removed in favor of using a null in the API to represent "no decision".

HOLD is being added as a sort of "soft successful".

## 🎨 Design File

@Jerryescandon 

Modals:
https://www.figma.com/file/em5ArPr044YaIe1Vb1zCZz/Pools---Record-of-Decision?type=design&node-id=6309-44439&mode=design&t=lVl0h2AgCTBVG24H-0

Step Tracker:
https://www.figma.com/file/em5ArPr044YaIe1Vb1zCZz/Pools---Record-of-Decision?type=design&node-id=6465-23869&mode=design&t=lVl0h2AgCTBVG24H-0

## 📸 Screenshot

![image](https://github.com/GCTC-NTGC/gc-digital-talent/assets/8978655/dfab7065-75fb-426c-a5dd-944831effa80)

![image](https://github.com/GCTC-NTGC/gc-digital-talent/assets/16977254/4f8fd5e8-d7ba-4de8-80ac-bcfb832bf2e4)

## :gear: The assessment step algorithm

```
Essential requirements = essential skills AND education requirements
Optional requirements = asset skills
Assessment Step Status =
  IF any essential requirements unsuccessful THEN "unsuccessful"
  IF any requirements unassessed THEN "to assess"
  IF essential requirements include at least one hold THEN "hold"
  IF all essential requirements successful THEN "success"

Also, on the step tracker, a step effectively only has the "to assess" status if all previous steps are "hold" or "success".
```

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the feature was properly implemented.

- [ ] Remove the old decision and add the new decision to the AssessmentDecision PHP enum.
- [ ] Ensure that the assessment decision mutation can set the three values or a null.
- [ ] Update the modals created in #8397
  - [ ] The "Not sure" selection sets the decision back to null
  - [ ] The "Not demonstrated (Hold for further assessment)" selection sets the decision to HOLD.
- [ ] Update the step tracker created in #8530
  - [ ] Add the hold status to the summary
  - [ ] Update the logic as shown in the algorithm above and make sure it can be easily shared with the Assessment Step List

PR #9022, opened 2024-01-15
[Fix] Grouping assessment results by candidate

### LINKED ISSUE #8957: 🐛 AssessmentStepTracker can list the same person multiple times
## 🐛 Bug

The AssessmentStepTracker component is meant to show one row per candidate (if they've reached the appropriate step). However, it's actually showing one row per AssessmentResult.

A candidate can have multiple AssessmentResults for a single Assessment Step (one for each skill), so the same users are appearing many times in each AssessmentStepTracker.

I think this means the numbers are off as well.

## 🦋 Expected Behaviour

The Assessment Step Algorithm (from #8644)
```
Essential requirements = essential skills AND education requirements
Optional requirements = asset skills
Assessment Step Status =
  IF any essential requirements unsuccessful THEN "unsuccessful"
  IF any requirements unassessed THEN "to assess"
  IF essential requirements include at least one hold THEN "hold"
  IF all essential requirements successful THEN "success"

Also, on the step tracker, a step effectively only has the "to assess" status if all previous steps are "hold" or "success".
```

## 🕵️ Details

https://www.chromatic.com/component?appId=61e099a1bb1465003a73d3ce&csfId=components-assessment-step-tracker&buildNumber=6320&k=65971820d6c07883395a1018-1200px-interactive-true&h=7&b=-1

## 📋 Steps to Reproduce

1. Look at the storybook story.
2. Notice the same names appearing many times per column.

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

The way we determine whether a candidate has passed a step needs to be more sophisticated, because passing a step requires passing for _each skill_, which have their own AssessmentResults.

We should have a reusable functions in a utility file which can determine the status for each step for a given candidate, and from that the current step for a given candidate. This functionality will likely need to be used in multiple places.

~I wonder if we should calculate this on the backend? Having a "stepsAssessed" count for a candidate available from the api would make this easier, and would also be useful for sorting the PoolCandidates table.~ For now let's compute this on the frontend.

## ✅ Acceptance Criteria

- [ ] A candidate appears at most one time per step in the AssessmentStepTracker
- [ ] We correctly determine whether a candidate has completed a step (see Expected Behaviour)
- [ ] If the sorting and step logic remains in the frontend, I'd like to see a Jest test for both the utilities functions and the component. ~If the logic gets moved to the backend, there should be a phpunit test.~

esizer:
🤖 Resolves #8957

## 👋 Introduction

Updates the `AssessmentStepTracker` component to group all assessments results by candidate so they only appear once in each step.

## 🕵️ Details

I think this may be better served as a API query eventually, but this seems to work as is. I think it got the tests right on this one. I added multiple results for the "to be assessed" candidate in the tests then confirmed it did not appear in the DOM twice and the result of the group function was one less than the results passed in.

## ❓ Question

I wasn't sure how to handle the expected behavior for "...step effectively only has the "to assess" status if all previous steps are "hold" or "success".". 

Does our backend handle this? They should only show up in a step column if an assessment result exists for that column. Or should this also artificially bump them over to the next step with a fake result for the next step? 🤔 Again, I feel like a query that returns a massaged type like this does may make this more effective and easier to work with.

```graphql
type CandidateAssessmentResult = {
  poolCandidate: PoolCandidate!;
  decision: AssessmentDecision;
  results: AssessmentResult[]!;
};

```

## 🧪 Testing

1. Build `npm run dev`
2. Login as admin
3. Navigate to CMO CMO Digital Careers screening page `/admin/pools/{poolId}/screening`
4. Confirm candidates only appear once in each step
5. Start storybook `npm run storybook:web`
6. Nsvigate to the "Assessment Step Tracker" story
7. Confirm candidates only appear once in each step

## 📸 Screenshot

Screenshot 2024-01-19 120425


tristan-orourke:
> They should only show up in a step column if an assessment result exists for that column. Or should this also artificially bump them over to the next step with a fake result for the next step?

I think the answer to this is that a user has to be able to exist in a column _without_ any results in that column yet.

apps/web/src/components/AssessmentStepTracker/utils.ts
```diff
@@ -3,17 +3,142 @@ import CheckCircleIcon from "@heroicons/react/20/solid/CheckCircleIcon";
...
+ * for a specific candidate
+ *
+ * @param results
+ * @returns CandidateAssessmentResult[]
+ */
+export const groupResultsByCandidate = (
+  results: AssessmentResult[],
+): CandidateAssessmentResult[] => {
+  const groupedAssessments = results.reduce<
+    CandidateAssessmentResultNoDecision[]
```
tristan-orourke, line 97: Maybe it would be easier (including conceptually) to put this in a Map (keyed by candidate id) instead of using an array and findIndex.

apps/web/src/components/AssessmentStepTracker/utils.ts
```diff
@@ -3,17 +3,142 @@ import CheckCircleIcon from "@heroicons/react/20/solid/CheckCircleIcon";
...
+        ...candidateResults.slice(candidateIndex + 1),
+      ];
+    },
+    [] as CandidateAssessmentResultNoDecision[],
+  );
+
+  return groupedAssessments.map((assessment) => ({
+    ...assessment,
+    decision: deriveResultsDecision(assessment.results),
+  }));
```
tristan-orourke, line 139: This feels to me like it should be separate from the `groupResultsByCandidate` function.

apps/web/src/components/AssessmentStepTracker/utils.ts
```diff
@@ -3,17 +3,142 @@ import CheckCircleIcon from "@heroicons/react/20/solid/CheckCircleIcon";
...
+ * Assessment Step Status =
+ *  IF any essential requirements unsuccessful THEN "unsuccessful"
+ *  IF any requirements unassessed THEN "to assess"
+ *  IF essential requirements include at least one hold THEN "hold"
+ *  IF all essential requirements successful THEN "success"
+ *
+ * @param results
+ * @returns NullableDecision
+ */
+const deriveResultsDecision = (
```
tristan-orourke, line 44: When deriving the decision, you also have to take into account some details from the AssessmentStep itself. The current function checks the status of existing results, but I don't see it checking if the AssessmentStep requires skills for which there exists no Result at all.

apps/web/src/components/AssessmentStepTracker/utils.ts
esizer: I started it out that way but quickly realized we just be calling it basically every where we use this group so that is why I included it in the return here. We use it in:

- Count on the summary of the step
- For the icon on the candidate
- For the sorting

So instead of calling it to get the decision in 3 different spots, it only needs to be called once here.

apps/web/src/components/AssessmentStepTracker/utils.ts
esizer: Hmm, okay I'll take a look. I assumed the fallback of `NO_DECISION` would be the catch all for missing assessment results though 🤔 Since, if a result does not exist it should fallback to "to assess", right?

tristan-orourke: Falling back on NO_DECISION is correct. I'm worried about the situation where a user only has a single Result, and its successful, but has no results at all related to other essential skills.

esizer: okay, that makes sense 👌 Thanks

tristan-orourke: My suggestion for this function would be to:
1. assemble the list of requirements from the Assessment Step (essential skills and possibly education)
2. iterate through the requirements
  a) try to find a matching Result
    - if matching result is UNSUCCESSFUL, return UNSUCCESSFUL
    - if matching result is ON HOLD then set a HAS_ON_HOLD flag to true and continue
    - if matching result is TO ASSESS then set HAS_TO_ASSESS flag to true and continue
    - if matching result cannot be found then set HAS_TO_ASSESS flag to true and continue
    - if matching result is SUCCESSFUL then continue
3. Return according to flags:
  - If HAS_ON_HOLD return ON_HOLD
  - else if HAS_TO_ASSESS return TO_ASSESS
  - else return SUCCESSFUL

NOTE: you could replace the HAS_TO_ASSESS flag with a ALL_SUCCESSFUL flag that gets switched from true to false. Also, I'm not 100% certain about how ON_HOLD interacts with TO_ASSESS and SUCCESSFUL, now that I think about it... which takes priority? 🤔

tristan-orourke: Went to look at #8644 to answer my own question. TO ASSESS takes priority over ON_HOLD:
```
3. Return according to flags:
  - If HAS_TO_ASSESS return TO_ASSESS
  - else if HAS_ON_HOLD return ON_HOLD
  - else return SUCCESSFUL
```  
If a user is ON_HOLD or SUCCESSFUL in one column, they should appear in the next column as well.

esizer: okay, thank-you so much for all that detail! It really helped a lot 😄 

I did a major refactor of the functionality and it should be fixed in [refactor grouping of candidates](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9022/commits/5df5503138a93af178dbd9956b446bd111f13323).

It is quite messy so I'd gladly take any suggestions 😅

tristan-orourke (changes requested):
I found some specific issues and left comments... I also found it pretty confusing to reason through overall 😓 . I'd love to find some time to pair and try to work out a better approach

apps/web/src/components/AssessmentStepTracker/utils.ts
```diff
@@ -326,3 +213,192 @@ export const sortResultsAndAddOrdinal = (
...
+          : false,
+      )
+      .filter(notEmpty);
+
+    assessments?.forEach((result) => {
+      if (!result) return;
+
+      const resultStep = result?.assessmentStep?.id
+        ? stepMap.get(result.assessmentStep.id)
+        : false;
```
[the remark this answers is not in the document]
esizer: The issue here is that `id` is undefined which you cannot pass to a map. Though I guess `result?.assessmentStep?.id ?? ""` would work since we never expect an empty string as key 🤔

apps/web/src/components/AssessmentStepTracker/utils.ts
```diff
@@ -188,11 +205,200 @@ export const sortResultsAndAddOrdinal = (
...
+  if (!results) return NO_DECISION;
+  let hasFailure: boolean = false;
+  let hasOnHold: boolean = false;
+  let hasToAssess: boolean = false;
+
+  const stepResults = results
+    ?.filter((result) => {
+      return result.assessmentStep?.id === step.id;
+    })
+    .filter(notEmpty);
```
tristan-orourke, line 230:
```suggestion
  const stepResults = results.filter((result) => {
    return result.assessmentStep?.id === step.id;
  });
```
Typescript is smart enough to realize that, if code has made it past line 221, then results must be `AssessmentResult[]`

apps/web/src/components/AssessmentStepTracker/utils.ts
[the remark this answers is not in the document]
tristan-orourke: Is that what typescript says? It's valid javascript in practice 😆 

```
const m = new Map();
m.set("one", 1);
  # Map(1) {'one' => 1}
m.get(undefined)
  # undefined
m.set(undefined, "test")
  # Map(2) {'one' => 1, undefined => 'test'}
m.get(undefined);
  # 'test'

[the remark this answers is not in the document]
Comment (esizer) [reply apps/web/src/components/AssessmentStepTracker/utils.ts:346]: Well, I guess I could type it to allow undefined, but I was trying to be safe 😆 I think `undefined` as a key should be fine so I'll just update the type 👍 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/5df5503138a93af178dbd9956b446bd111f13323/apps/web/src/components/AssessmentStepTracker/utils.ts#L305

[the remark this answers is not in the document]
Comment (esizer) [reply apps/web/src/components/AssessmentStepTracker/utils.ts]: Updated to allow `undefined` keys to reduce the logic here, thanks for the suggestion 👍 

[allow undefined keys for grouped steps](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9022/commits/3c19d968b263a0f711e6f38746558c3eba05cec5)

File: apps/web/src/components/AssessmentStepTracker/utils.ts [hunk 6]
```diff
@@ -326,3 +213,192 @@ export const sortResultsAndAddOrdinal = (
...
+  steps.forEach((step, index) => {
+    stepMap.set(step.id, {
+      step,
+      // Step one should show all candidates regardless of results existing
+      assessments: index === 0 ? allCandidatesMap : new Map(),
+    });
+  });
+
+  poolCandidates.forEach((poolCandidate) => {
+    // Only use assessments associated with this pool
```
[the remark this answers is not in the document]
esizer: You are right, I'm not sure what I was thinking 😅 I also accidentally lumped this in with a different commit but it was resolved in [ensure steps are in order](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9022/commits/379f8a3fbb9fe446b5f8bc8af67d0d9b831ebd30)

apps/web/src/components/AssessmentStepTracker/utils.ts
```diff
@@ -326,3 +213,192 @@ export const sortResultsAndAddOrdinal = (
...
+            ? [...stepCandidateAssessments.results, result]
+            : [result],
+        });
+      }
+    });
+  });
+
+  // Determine the decision for each assessment now that we have all the results
+  let previousStep: GroupedStep | undefined;
+  stepMap.forEach((step) => {
```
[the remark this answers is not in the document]
tristan-orourke: Hmm, okay. It still feels weird to me to rely on that. Map feels like the wrong structure to use if we care about order.

tristan-orourke (changes requested):
Not yet a complete review, but I want to submit these comments for now

esizer:
@tristan-orourke okay, I think I have it all plugged in and working. I used the utility functions to create a shape that works well and avoids a bunch of prop drilling in [update using new simplified functions](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9022/commits/6354b21603cc348c387a63a67da016e78245a3fc)

apps/web/src/components/AssessmentStepTracker/AssessmentStepTracker.tsx
```diff
@@ -17,29 +18,25 @@ export interface AssessmentStepTrackerProps {
...
+      {groupedSteps.map(({ step, resultCounts, results }, index) => (
+        <Board.Column key={step.id}>
+          <Board.ColumnHeader
+            prefix={intl.formatMessage(applicationMessages.numberedStep, {
+              stepOrdinal: index + 1,
+            })}
+          >
+            {getLocalizedName(step.title, intl)}
+          </Board.ColumnHeader>
+          <ResultsDetails {...{ resultCounts, step }} />
```
tristan-orourke, line 36: I've never seen this syntax in a component before. Is this the same as `<ResultsDetails resultCounts={resultsCounts} step={step} />`?

esizer: Yup, I can update it to use the more common syntax though. We do use it in places but not very frequently,

apps/web/src/components/AssessmentStepTracker/AssessmentStepTracker.test.tsx
```diff
@@ -167,24 +78,49 @@ describe("AssessmentStepTracker", () => {
...
     ).toBeInTheDocument();
   });
 
+  it("should group results by candidate", () => {
+    renderAssessmentStepTracker();
+
+    // Has two results but should only see one link for it
+    expect(
+      screen.getAllByRole("link", {
+        name: `1. ${unassessedCandidate.user.firstName} ${unassessedCandidate.user.lastName}`,
```
tristan-orourke, line 102: If the same user did appear multiple times, wouldn't the number be different?

apps/web/src/components/AssessmentStepTracker/utils.ts
```diff
@@ -184,11 +190,257 @@ export const sortResultsAndAddOrdinal = (
...
+        case AssessmentDecision.Hold:
+          hasOnHold = true;
+          break;
+        case AssessmentDecision.Unsuccessful:
+          hasFailure = true;
+          break;
+        default:
+      }
+    }
+  });
```
tristan-orourke, line 270: This only applies to the APPLICATION_SCREENING assessment type

tristan-orourke (changes requested):
Awesome, there's just one comment in the util file that still needs addressing, and I have a question about one test.

apps/web/src/components/AssessmentStepTracker/utils.ts
esizer: Oh I was basing this off of [the logic outlined](https://github.com/GCTC-NTGC/gc-digital-talent/issues/8957#:~:text=Essential%20requirements%20%3D%20essential%20skills%20AND%20education%20requirements) 😅 

So, should it be `step.type === AssessmentStepType.ApplicationScreening` instead?

esizer: Fixed in [fix step decision logic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9022/commits/1b0d12688ce81ae146853389288f1682058bc124)

apps/web/src/components/AssessmentStepTracker/AssessmentStepTracker.test.tsx
esizer: okay, updated to exclude the number and check for all instances to ensure only one appears in [update test for grouping](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9022/commits/10bd87a3ff2bd50980f2225a857c18c8cb9e485b)

apps/web/src/components/AssessmentStepTracker/utils.ts
```diff
@@ -184,11 +190,257 @@ export const sortResultsAndAddOrdinal = (
...
```
tristan-orourke, line 270: Actually, I meant like this:
```suggestion
  // Check for Education requirement if this is an ApplicationScreening step
  if (step.type === AssessmentStepType.ApplicationScreening) {
    const educationResults = stepResults.filter(result => result.assessmentResultType === AssessmentResultType.Education);
    if (educationResults.length === 0) {
      hasToAssess = true;
    }
    educationResults .forEach((result) => {
      // Any "to assess" should be marked
      if (result.assessmentDecision === null) {
        hasToAssess = true;
      }
      switch (result.assessmentDecision) {
        case null:
          hasToAssess = true;
          break;
        case AssessmentDecision.Hold:
          hasOnHold = true;
          break;
        case AssessmentDecision.Unsuccessful:
          hasFailure = true;
          break;
        default:
      }
    });
  }
```

tristan-orourke (changes requested):

tristan-orourke (approved):
Thanks for working with me on this!

PR #9025, opened 2024-01-15
[Deps] Bump Node and NPM

### LINKED ISSUE #8963: ♻️ Upgrade to Node.js 20 and NPM 10.
## ♻️ Debt/Refactor

The latest (as of now) LTS version of Node.js is 20 and it comes with NPM 10.  Since we're on Node.js 18 and NPM 8 we should consider upgrading.

image

## ✅ Acceptance Criteria

- [ ] package.json and lock files updated
- [ ] Maintenance container updated
- [ ] GH workflows updated
- [ ] Help your teammates update their environments

petertgiles:
🤖 Resolves #8963

## 👋 Introduction

This branch bumps Node and NPM to the latest LTS versions downloadable on nodejs.org.

## 🧪 Testing

1. Rebuild the maintenance container: `docker-compose build maintenance`
2. If you like to run scripts locally using nvm:
    1. `nvm install 20.11.0`
    2. `nvm use 20.11.0`

- [ ] Can setup and refresh the app using maintenance app
- [ ] Can run Cypress and Jest
- [ ] GH Workflows work
- [ ] Can build in Azure

> [!Note]
> It seems like this version of npm changes something about how it handles the permissions of files and current users.  The change to `packages/i18n/package.json` fixes a hangup in the cypress GH Action that runs after the docker container sets up the environment.  The change to `maintenance/scripts/setup.sh` aims to add writable permission to the directories where files might be created and overwritten when running scripts.  I was never able to get i18n to build locally, even though the `dist` directory is writable by everyone.
image
At least everything still runs fine in the maintenance container.


vd1992:
Combing through the log, I noticed this. Didn't appear to break build, and I also spotted it in another PR's successful Cypress run. Bringing it up still 

image

vd1992:
Doing some digging, not sure if this has any effect, but the Cypress Github action also has its own Node version
https://github.com/cypress-io/github-action#compatibility

petertgiles:
> Combing through the log, I noticed this.
Yeah, I've been seeing that locally for maybe a month or two now.  It hasn't gotten in my way so I haven't taken a close look at it but I'm fairly certain it doesn't have anything to do with this PR.  I'm guessing it has to do with the interaction of git and docker.

petertgiles:
> Doing some digging, not sure if this has any effect, but the Cypress Github action also has its own Node version

Oh, I didn't know this!  Another good reason to get to Node v20.

esizer (approved):
LGTM! 👍

PR #9033, opened 2024-01-17
[Fix] Errant new records in CSV download

### LINKED ISSUE #9018: 🐛 CSV download sometimes fails to escape quotes properly
## 🐛 Bug

When downloading candidates, the CSV sometimes contains quotes that are not escaped properly. The failure rate was fairly high (maybe 30%?) for the recent EX process.

## 🦋 Expected Behaviour

All quotes inside user submitted text in the CSV has quotes escaped (like this "" instead of ") always.

## 📋 Steps to Reproduce

Not sure the exact 'cause, but can show examples in real files if you need.

1. Just download large number of candidates as CSV

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] Quotes in CSV escaped properly always
- [ ] Jest test to ensure quotes are escaped properly

esizer:
🤖 Resolves #9018 

## 👋 Introduction

Applies some basic sanitization to strings used in CSV downloads to avoid accidently creating new records.

## 🕵️ Details

We had some sanitization on experience details but other user created data was not being sanitized properly. The data containing double quotes and line breaks were leading to CSV interpreting them as new columns and rows, breaking the CSV.

In order to fix this we have:

- Sanitized user created strings by 
  - replacing " with "" (way of escaping in CSVs) which was leading to new columns
  - removing line breaks `\r\n` which was leading to new rows (records)
  - added a column wrapper character (") which wraps each column of data to avoid accidental new columns

## 🧪 Testing

1. Build `npm run dev`
2. Login as admin
4. Edit some candidates adding double quote (") and line breaks into their records (notes is an easy field to accomplish this)
5. Navigate back to the table
6. Select those users
7. Activate "Download CSV"
8. Open the CSV and confirm the data is clean and rendered properly in any spreadsheet software
    - Preferably test with Microsoft Excel since that is likely the main software used by admins

### DIFF
--- a/apps/web/src/components/PoolCandidatesTable/poolCandidateCsv.tsx
+++ b/apps/web/src/components/PoolCandidatesTable/poolCandidateCsv.tsx
@@ -28,6 +28,7 @@ import {
   getExperienceTitles,
   getScreeningQuestionResponses,
   getIndigenousCommunities,
+  sanitizeCSVString,
 } from "~/utils/csvUtils";
 import { PoolCandidate, PositionDuration, Pool } from "~/api/generated";
 import adminMessages from "~/messages/adminMessages";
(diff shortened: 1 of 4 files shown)


vd1992:
Not relevant to the fields you've touched but just a curiosity whether these cells are an issue.
If you submit an application and have multiple experiences attached to a skill, it kinda warps the excel sheet view. It doesn't knock them out of order, just makes them really huge which I wasn't sure if it was a problem. It could also just be an excel setting for me, so a local issue. 

image

esizer:
> Not relevant to the fields you've touched but just a curiosity whether these cells are an issue.
> If you submit an application and have multiple experiences attached to a skill, it kinda warps the excel sheet view. It doesn't knock them out of order, just makes them really huge which I wasn't sure if it was a problem. It could also just be an excel setting for me, so a local issue.

Yeah, I agree it doesn't look great. I opted not to strip out new lines here to keep separation between experiences. I wonder if we would prefer to italicize the titles and striping new lines? 🤔 

Do you have any thoughts on this @gobyrne

vd1992:
And one more curiosity, it seems odd to me for 3/4 equity fields to output "no" when not applicable but the fourth just displays a blank space. But I don't know if that was intentionally desired as such or a big deal

image

vd1992 (approved):
Unrelated questions/observations aside, things look good to me. I could not break the column line up as was the case before. Can spin off issues for them if needed. 

Great job!

Issue #9037, opened 2024-01-17
🐛 Toasts with Links cause the app to crash


petertgiles:
## 🐛 Bug

Our app crashes if a toast attempts to call useIntl.  This is because the Toast component is outside of the IntlProvider.  This is a problem, for example, if you want to include a Link in a toast.

## 🦋 Expected Behaviour

You should not be able crash the app with a toast.

## 📋 Steps to Reproduce

1. Pop a toast with a <Link> component in it.

## 📸 Screenshot

![image](https://github.com/GCTC-NTGC/gc-digital-talent/assets/8978655/ed9fa74b-a4f0-4cbe-a6a9-8a39ce09fe94)
![image](https://github.com/GCTC-NTGC/gc-digital-talent/assets/8978655/27776c6c-9e02-49f1-9136-46e45df88725)

## 🙋‍♀️ Proposed Solution

1) Move the toast provider inside of IntlProvider.

:warning: It was probably moved out because rerenders would cause the toast to disappear.  Or maybe it was causing infinite rerenders?

2) Rig the compiler to prevent any components with intl calls in them?  :shrug: 

## ✅ Acceptance Criteria

- [ ] It is not possible to crash the app with a toast

Issue #9036, opened 2024-01-17
🐛 Support endpoint dumps out Freshdesk responses


petertgiles:
## 🐛 Bug

The support endpoint dumps out the JSON responses (both successful and not) from Freshdesk with no filtering.  Right now, everything is probably safe but we have no guarantee it will remain this way.  Someday it may leak our key.

Generally, in prod mode we don't reveal details of "internal server error" type errors to the user.

## 🦋 Expected Behaviour

Our endpoint should probably just return generic "everything OK" and "something went wrong" messages and log the details internally for ourselves.

## 📋 Steps to Reproduce

1. Setup the Freshdesk variables (correctly or incorrectly) in the .env file.
2. Attempt to submit a ticket from /support
3. Check the server response

## 📸 Screenshot

![image](https://github.com/GCTC-NTGC/gc-digital-talent/assets/8978655/492cee28-90a7-47f1-8615-5e65c0ee4ac3)

## ✅ Acceptance Criteria

- [ ] Success response is generic
- [ ] Failure response is generic
- [ ] Details logged internally
- [ ] Frontend page displays appropriate toasts

PR #9039, opened 2024-01-17
[Fix] Fix the blank white error on the support page

### LINKED ISSUE #8897: 🐛 Blank white page when submitting a ticket fails
## 🐛 Bug

If submitting a Freshdesk ticket fails, for example with invalid credentials, no error message is displayed to the user and the while page goes white.

## 🦋 Expected Behaviour

The site should stay with the filled in form and an error toast should display indicating there was an error submitting the ticket.

## 📋 Steps to Reproduce

1. Configure invalid freshdesk credentials
2. Try to submit a ticket

## 📸 Screenshot

image

## ✅ Acceptance Criteria

- [ ] the site should stay with the filled in form
- [ ] an error toast should display indicating there was an error submitting the ticket

petertgiles:
🤖 Resolves #8897 

## 👋 Introduction

This branch fixes the crash to a blank white screen on the support page.

## 🕵️ Details

The culprit of the app crash was that our toasts are not in the intl provider and our error message used the Link component which calls an intl function.  For a quick fix, I changed it a regular `a` tag.  I opened  #9037 for a more thorough review.

There was a desire to ensure that an email address was always displayed for errors.  So I removed the two other error messages in favour of the single generic one with the email.  I did a bit of refactoring on the submit handler to use `await` to make it easier to read, too.

There was also a desire for better logging around failed ticket creation so I added that as well.  I changed the response code to 500 which makes more sense, I think, and will make it more noticeable in analytics.

I noticed a security concern that I logged in #9036.

There was a small fix to a turbo bug I rolled in, too.  It was a duplicate script shortcut preventing running `check-intl` in the maintenance container.

## 🧪 Testing

- [ ] With Freshdesk configured correctly
  - [ ] Tickets go through with a success toast.
  - [ ] The form toggles to the success message
- [ ] With Freshdesk configured incorrectly
  - [ ] An error toast with an email address is always displayed
  - [ ] A detailed error message is logged server side
  - [ ] The form doesn't toggle or navigate

## 📸 Screenshot

image

### DIFF
--- a/api/app/Http/Controllers/SupportController.php
+++ b/api/app/Http/Controllers/SupportController.php
@@ -4,15 +4,18 @@ namespace App\Http\Controllers;
 
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Http;
+use Illuminate\Support\Facades\Log;
 
 class SupportController extends Controller
 {
     public function createTicket(Request $request)
     {
         if (! config('freshdesk.api.tickets_endpoint') || ! config('freshdesk.api.key')) {
+            Log::error('Attempted to create a ticket with missing config values.');
+
             return response([
                 'apiResponse' => 'Missing parameters',
-            ], 422);
+            ], 500);
         }
         $parameters = [
             'description' => $request->input('description'),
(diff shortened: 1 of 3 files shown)


esizer (approved):
LGTM! 👍

PR #9045, opened 2024-01-18
[Tooling] Slow query logging

### LINKED ISSUE #6834: 🛠️ Set up Laravel slow query monitoring with request text
## 🛠️ Tooling

We want to monitor when queries take a long time to complete in our API.

## 🙋‍♀️ Proposed Solution

One promising approach is to add a middleware that times the duration of the request and logs it if it exceeds a threshold.
https://github.com/GCTC-NTGC/gc-digital-talent/issues/5442#issuecomment-1570381952

There is also the [whenQueryingForLongerThan](https://laravel.com/docs/10.x/database#monitoring-cumulative-query-time) helper.

Logging these occurrences would allow us to monitor them in the Azure logs for dashboards, alerts, etc.

## ✅ Acceptance Criteria

- [ ] Laravel logs slow queries along with the request text
- [ ] Available for querying in Azure Logs

vd1992:
🤖 Resolves #6834

## 👋 Introduction

Adds logging of queries that take a long while to execute

## 🕵️ Details

Attempted more than one method, see below for each.
Winner in my opinion was the one that was Peter's suggestion that needed only a little adjustment. 

## 🧪 Testing

1. Play around with timing and look at the Laravel log
2. Consider what is best to track and time cutoffs?


api/app/Http/Middleware/AuditQueryMiddleware.php
```diff
@@ -36,6 +36,30 @@ public function handle(Request $request, Closure $next)
...
+        //     $this->logger->info(
+        //         'Slow query',
+        //         [
+        //             'request' => $request->json()->all(),
+        //             'elapsed_seconds' => $elapsedTime,
+        //         ]
+        //     );
+        // }
+
+        // return $result;
```
vd1992, line 61: The method suggested by Peter, appears to work quite fine. As you can see in the comments, tested timing with the logging on vs off and it appear to cause no noticeable slowdown. 
It also gives a nice logging result of the query and timing where it is easy to comprehend what is happening. 

Screenshot 2024-01-17 172454

api/app/Providers/AppServiceProvider.php
```diff
@@ -32,6 +34,27 @@ public function boot()
...
+        // DB::whenQueryingForLongerThan(50, function (Connection $connection, QueryExecuted $event) {
+        //     // Notify development team...
+        //     Log::debug(
+        //         [
+        //             'sql' => $event->sql,
+        //             'bindings' => $event->bindings,
+        //             'milliseconds' => $event->time,
+        //         ]
+        //     );
+        // });
```
vd1992, line 47: As seen in Laravel docs, did not work well as it doesn't output much info about the entire event, rather, after it hits some time cutoff it outputs the last statement to push it over the time. Even if that statement didn't take that long to execute

Screenshot 2024-01-17 170312

api/app/Providers/AppServiceProvider.php
```diff
@@ -32,6 +34,27 @@ public function boot()
...
+
+        // DB::listen(function ($query) {
+        //     if ($query->time > 20) {
+        //         Log::warning('Query exceeded 20 milliseconds -', [
+        //             'sql' => $query->sql,
+        //             'milliseconds' => $query->time,
+        //         ]);
+        //     }
+        // });
```
vd1992, line 56: An idea I found elsewhere to use the familiar listen method. This logs the specific SQL statements that take longer than a certain time threshold but not sure if it offers much insight as it might be confusing to comb through. As you don't know the operation context so you can't do much with the information. 

Screenshot 2024-01-17 172016

petertgiles (changes requested):
Nice, it works really well!  Thanks for going through the experiments of the other methods, too.  I think I'd like to see this get moved to its own middleware, though, and not rolled into the (very vaguely named  :sweat_smile: ) AuditQueryMiddleware.  This keeps everything single-responsibility and lets us independently turn them off and on.

api/app/Http/Middleware/AuditQueryMiddleware.php
```diff
@@ -36,6 +36,21 @@ public function handle(Request $request, Closure $next)
             );
         }
 
-        return $next($request);
+        $start = hrtime(true);
+        $result = $next($request);
+        $end = hrtime(true);
+        $elapsedTime = ($end - $start) / 1000000000;
```
petertgiles, line 42: Would recommend calling this elapsedSeconds to make the units clear.

api/app/Providers/AppServiceProvider.php
petertgiles: Good to know!  This could be useful for troubleshooting.  Could you leave this in the code, commented out?

api/app/Providers/AppServiceProvider.php
petertgiles: Too bad.  It's a nice looking facade.

api/app/Http/Middleware/AuditQueryMiddleware.php
vd1992: Aye, and moved into its own file in https://github.com/GCTC-NTGC/gc-digital-talent/pull/9045/commits/72ac439970b370ba6b497438ec2296f4ceed3845

api/app/Providers/AppServiceProvider.php
vd1992: Yeah sure, will do
https://github.com/GCTC-NTGC/gc-digital-talent/pull/9045/commits/28cb381993b69da19a5b5b5cceb8d323add30f5f

petertgiles (approved):
Love it!  :1st_place_medal:

brindasasi (commented):
aah .. I like what I'm seeing 👏 awesome work! 
15 secs is still a lot of time IMO. 
https://postgres.ai/blog/20210909-what-is-a-slow-sql-query
I remember public user facing screens used to have 3 seconds rule as the standard.
But atleast we could log anything more than 10 secs.

vd1992:
> aah .. I like what I'm seeing 👏 awesome work! 15 secs is still a lot of time IMO. https://postgres.ai/blog/20210909-what-is-a-slow-sql-query I remember public user facing screens used to have 3 seconds rule as the standard. But atleast we could log anything more than 10 secs.

Hm, I wasn't too sure about the timing either. 
Thoughts? @petertgiles @tristan-orourke

petertgiles:
> Hm, I wasn't too sure about the timing either.
> Thoughts?

Yeah, 3 seconds or 5, makes sense to me.

vd1992:
Okay, set to 5 in https://github.com/GCTC-NTGC/gc-digital-talent/pull/9045/commits/598fc21f901dab80bf0ec074b7be99520d17467e
@brindasasi @petertgiles

Issue #8491, opened 2023-11-15
♻️ Form Repeater refactor


petertgiles:
## ♻️ Debt/Refactor

The repeater has been updated and extended a few times.  It might be time for a refactor to freshen it up.

Two things come to mind:
1. Focus on the up and down buttons is lost when the cards move.   This is not good for a11y.
2. The API is very large and a lot of redundant data has to be passed down to the fieldsets.  It would be nice if there was an internal context that the fieldsets could access.

## Proposed Solution

Look at card repeater created in #8161 for inspiration.

## ✅ Acceptance Criteria

- [ ] Focus on the up/down buttons is not lost when the cards move.
- [ ] Fieldsets don't have to have redundant data passed into them that the root already knows.

petertgiles:
This component should become a lot simpler when #8161 is done.

tristan-orourke:
@petertgiles is there still work left to do on this since #8161?

petertgiles:
As far as I know, yes.  However, 8161 is a great example, now, of the changes to be made.

PR #9053, opened 2024-01-18
[Feature] Add card repeater component

### LINKED ISSUE #8161: ⚙️ Card repeater

esizer:
🤖 Resolves #8161 

## 👋 Introduction

Creates a `CardRepeater` component that allows us to repeat data. Similar to the `useFieldArray` from `react-hook-form` but without the need for it to be a form.

### Approach

I tried to handle most of the state in the actual context so it should handle most interactions for you. The only thing you should need to handle is adding and editing. The main reason for this is that we typically use dialogs for these actions and in order to keep focus management, those dialogs must be created for each card.

There is a hook (`useCardRepeaterContext`) you can use to update the internal state from these dialogs (see story example).

#### Async 

```tsx
const Add = () => {
  const [{fetching}, executeMutation] = useMutation(Mutation);
  const methods = useForm();
  const { append, total, remove } = useCardRepeaterContext();

  const handleError = () => {
    remove(total - 1); // Failed so remove the optimistic item
    toast.error("Failed!");
  };

  const handleSubmit = (data: FormData) => {
    // Optimistic(?)
    append(data);
    executeMutation({data}).then(result => {
      if(result.data) {
        // We may not need to update anything here
        // urql cache should update in the parent
        toast.success("Success!");
      } else {
        handleError();
      }
    }).catch(() => handleError());
  }

  retrun (
    <Dialog>
      <FormProvider {...methods} onSubmit={handleSubmit}>
        <FormFields />
        <Button type="submit">{fecthing ? "Adding..." : "Add"}</Button>
      </FormProvider>
    </Dialog>
  )
}

const Repeater = () => {
  const [{data, fetching}] = useQuery(Query);

  const items = data.items.filter(notEmpty) ?? [];

  return (
    <CardRepeater.Root items={items}add={<Add />}>
      {items.map((item, index) => (
        <CardRepeater.Card
          key={item.id}
          index={index}
        >
          {item.value}
        </CardRepeater.Card>
      ))}
    </CardRepeater.Root>
  );
}
```

### API

```tsx
// Force items to contain an ID for the key
export type ItemWithId<T> = T & { id: string };

type RootProps = {
   /** Default set of items (should never be needed but there if you find a use) */
  defaultItems?: ItemWithId<T>[];
  /** The current list of items */
  items: ItemWithId<T>[];
  /** Array of the indices to disable */
  moveDisabledIndexes?: number[];
   /** Array of the indices to disable editing */
  editDisabledIndexes?: number[];
   /** Array of the indices to disable removing */
  removeDisabledIndexes?: number[];
  /** Max number of items */
  max?: number;
  /** callback for when the list is updated (move, delete, add, edit) */
  onUpdate?: (newItems: ItemWithId<T>[]) => void;
};

type useRepeaterContextReturn<Item> = {
  /** Total number of items */
  total: number;
  /** Add item to end */
  append: (newItem: Item) => void;
  /** Add item to start */
  prepend: (newItem: Item) => void;
  /** Move an item to specific index */
  move: (from: number, to: number) => void;
  /** Remove item at specific index */
  remove: (index: number) => void;
 /** Update an item at a specific index */
  update: (index: number, Partial<Item>) => void
 /** Also contains props passed into the root */
...rest: RootProps
}
```

### Keyboard shortcuts

Added in very basic keyboard shortcuts. If you are within the list you can us up and down arrow keys to navigate the cards.

## 🧪 Testing

2. Navigate to the "Card Repeater" story
3. Confirm it functions as expected

## 📸 Screenshot

Screenshot 2024-01-23 161056


yonikid15 (commented):
Code is clean and concise, everything works as expected looking amazing 🎉!

Comments / Questions:
- One question is where should the focus be held after deleting an item? Should the user be placed back at the top of the list. Should the focus be placed on the next item instead?
- I believe we'll need a locked feature for this in the assessment builder. The [Repeater](https://github.com/GCTC-NTGC/gc-digital-talent/blob/8161-card-repeater-component/packages/forms/src/components/Repeater/Repeater.tsx#L47-L48) has this feature added so maybe looking into that might help.

esizer:
> * One question is where should the focus be held after deleting an item? Should the user be placed back at the top of the list. Should the focus be placed on the next item instead?

It should be the next item. That is just my opinion, I'd gladly take any suggestions. @RM-1978 What do you think?


> * I believe we'll need a locked feature for this in the assessment builder. The [Repeater](https://github.com/GCTC-NTGC/gc-digital-talent/blob/8161-card-repeater-component/packages/forms/src/components/Repeater/Repeater.tsx#L47-L48) has this feature added so maybe looking into that might help.

Yeah, that is one of the items I'm still working on. Should be the last thing I think 🤔

RM-1978:
@esizer and @yonikid15  It seems to me that there are three options we could explore. I am less enthused with the last option. 

1. Focus on the same spot in the list, but on the new item that has now "slid" into the spot after the deletion. Presumably, the list "collapses," and all of the items after the deleted one get moved up closer to the top of the list. If you leave the focus location the same, the new item that occupies that position in the list would be in focus.
2. Move the focus to the item before the deleted item. Instead of leaving the focus on the position in the list, move the focus back to the item BEFORE the deleted item.
3. Lose focus on the list altogether and declare focus on some higher-level page element.

esizer:
> It seems to me that there are three options we could explore. I am less enthused with the last option

I agree, last option is not great. Right now, we just keep focus on the current item (option 1). I think we should be safe leaving it as is. Thanks for the suggestions 🙇‍♀️

packages/ui/src/components/CardRepeater/CardRepeater.stories.tsx
```diff
@@ -0,0 +1,223 @@
...
+export const WithItems = Template.bind({});
+WithItems.args = {
+  items: defaultItems,
+};
+
+export const MaxItems = Template.bind({});
+MaxItems.args = {
+  ...WithItems.args,
+  max: 4,
+};
```
vd1992, line 203: Might be best to set the max to 4 since you are mapping 6 items so you don't have a story that should be impossible. 
Also, not sure if this is a problem but if you are over limit you can start adding again. It only prevents when you are at the limit. Not sure if it is a problem in that it _should_ be impossible to have happen. 

image

esizer: haha oops! I added two more for the locked story and forgot to update this. Thanks for catching that 😆 [fix max items story](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9053/commits/1b84350e1a0f3394632fce5fb9d041645055c181).

Also, should be hard to get over the max but I think it makes sense to be safe so I updated it to be `>=` instead of just `===` in [disabled add button is equal to or grater than max](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9053/commits/993ca973314b791a379237f27f467576b8d3d0fc)

packages/ui/src/components/CardRepeater/CardRepeater.tsx
petertgiles, line 1: I'm really not sure how doable this will be.  The move buttons become active for a brief period before the move animation runs and it becomes possible click it a second time causing an unwanted mutation to fire.  It's pretty minor so if it's too complicated, let's spin off an issue and not block merge.

https://github.com/GCTC-NTGC/gc-digital-talent/assets/8978655/57a47bda-c647-4326-9556-1faca5b9b7ac

petertgiles (changes requested):
Oh wow, this turned out so nicely. :heart_eyes:   The context provider should be a good guide for #8491.  Thanks for tidying up the assessment builder, too.  :sweat_smile:   I love it so much.

packages/ui/src/components/CardRepeater/CardRepeater.tsx
petertgiles: Hmm, yeah, I really prefer the current API to either alternative you mentioned.  If you can get a bug up on the board to track this I'm happy to approve as-is.

esizer: After having some time to think about it overnight, I was able to get this working without API changes. I just had to manage the state internally then update it with `useEffect` when the API data is changed.

[make card repeater optimistic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9053/commits/1cf4a494c5a2a1408d4f3b562bc5586a352a5353)

petertgiles: Looking good!  I think you need to add a way to back out the change if there's an error though.

https://github.com/GCTC-NTGC/gc-digital-talent/assets/8978655/4f2f523d-ace5-4d1d-9e30-ea5ae4e24b34

esizer: ohhh right, the cache is not affected on an error 🤦‍♀️

esizer: Thanks for catching that! Should be fixed as of [reset assessment steps on error](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9053/commits/b2ce20fb46e83a2f137d7307bd8b5fcb72e870ba)

petertgiles: Nice!  Works great!

petertgiles (approved):

PR #9059, opened 2024-01-19
[Bug] Don't output endpoint response

### LINKED ISSUE #9036: 🐛 Support endpoint dumps out Freshdesk responses

vd1992:
🤖 Resolves #9036 

## 👋 Introduction

No longer outputting the endpoint response into the API, just a simple success/error. 
Error toasting and logging of errors already present. 

## 🧪 Testing

1. Not sure, does it need to be pushed to dev vertical?
2. Messages appear instead of json

## 📸 Screenshot

image

### DIFF
--- a/api/app/Http/Controllers/SupportController.php
+++ b/api/app/Http/Controllers/SupportController.php
@@ -45,13 +45,13 @@ class SupportController extends Controller
             );
         if ($response->status() == 201) { // status code 201 = created.
             return response([
-                'serviceResponse' => $response->json(),
+                'serviceResponse' => 'success',
             ], 200);
         } else {
             Log::error('Error when trying to create a ticket: '.$response->getBody(true));
 
             return response([
-                'serviceResponse' => $response->json(),
+                'serviceResponse' => 'error',
             ], 500);
         }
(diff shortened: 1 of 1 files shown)


yonikid15 (approved):
Looks good! 
Success: 
image

Error: 
image

PR #9077, opened 2024-01-25
[Bug] Update tab active + focused styles

### LINKED ISSUE #8785: 🐛 Tab focus styles are a bit off
## 🐛 Bug

The new tab focus styles have a few minor styling oddities:
* The font weight of focused tabs shouldn't change - they should retain the styles of their current state
* The font color of a focused tab, regardless of its current state, should change to `black`

<details>
  <summary>📸 Tab state designs</summary>
  
</details> 

## 📋 Steps to Reproduce

1. Run Storybook
2. Navigate to the tab stories
3. Focus the tab components

```[tasklist]
### ✅ Requirements
- [ ] Focused tabs shouldn't modify font weight
- [ ] Focused tabs should always have a `black` font color
```

yonikid15:
🤖 Resolves #8785 

## 👋 Introduction

- Updates tab active + focused font color to black
290341557-404d0b3f-e5c0-4a6f-8b6a-992c686eddf2

## 🧪 Testing

1. Build app `npm run build`
2. Login as `admin@test.com`
3. Go to pools table and select a pool
4. Click and tab through TabNav
5. Ensure all states matches the states in the image above

## 📸 Screenshot

image

### DIFF
--- a/packages/ui/src/components/Tabs/utils.ts
+++ b/packages/ui/src/components/Tabs/utils.ts
@@ -39,6 +39,7 @@ export const commonTabStyles = {
     "data-h2-color": `
       base(black)
       base:children[a:focus-visible](black)
+      base:focus-visible:children[span](black)
       base:selectors[[data-state='active']](primary.darker)
     `,
     "data-h2-margin-top": "base(x.25) base:hover(0)",
--- a/packages/ui/src/components/Tabs/utils.ts
+++ b/packages/ui/src/components/Tabs/utils.ts
@@ -38,8 +38,8 @@ export const commonTabStyles = {
     `,
     "data-h2-color": `
       base(black)
+      base:focus-visible(black)
       base:children[a:focus-visible](black)
-      base:focus-visible:children[span](black)
       base:selectors[[data-state='active']](primary.darker)
     `,
     "data-h2-margin-top": "base(x.25) base:hover(0)",
--- a/packages/ui/src/components/Tabs/utils.ts
+++ b/packages/ui/src/components/Tabs/utils.ts
@@ -38,7 +38,8 @@ export const commonTabStyles = {
     `,
     "data-h2-color": `
       base(black)
-      base:selectors[[data-state='active']](black.darker)
+      base:children[a:focus-visible](black)
+      base:selectors[[data-state='active']](primary.darker)
     `,
     "data-h2-margin-top": "base(x.25) base:hover(0)",
     "data-h2-outline": "base(none) base:children[a](none)",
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/Tabs/utils.ts
```diff
@@ -38,7 +38,7 @@ export const commonTabStyles = {
     `,
     "data-h2-color": `
       base(black)
-      base:selectors[[data-state='active']](primary.darker)
+      base:selectors[[data-state='active']](black.darker)
```
esizer, line 41: I believe this was supposed to be applied to [the focused style only](https://github.com/GCTC-NTGC/gc-digital-talent/issues/8785#:~:text=Focused%20tabs%20should,black%20font%20color) (not active).  This should probably be:

```tsx
      base(black)
      base:focus-visible(black)
      base:selectors[[data-state='active']](primary.darker)
      base:selectors[[data-state='active']]:focus-visible(black)
```

esizer (changes requested):
I think this needs to be applied to the focus style, not the active. 

Am I correct @substrae ?

packages/ui/src/components/Tabs/utils.ts
substrae: @esizer Yeah, you've got it. The black text and blue focus color should be applied only on `focus-visible` but applies to both the generic and active states. The difference between them is that the active state, when focused, retains the bolding.

esizer (changes requested):
The fix works for `NavTabs` but not `Tabs`. This is because it only targets `a` tags and the underlying component for `Tabs` is a `button`.

Screenshot 2024-01-25 135915

yonikid15:
> The fix works for `NavTabs` but not `Tabs`. This is because it only targets `a` tags and the underlying component for `Tabs` is a `button`.

Nice catch, should be fixed by [2b1cfca](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9077/commits/2b1cfca691268d830f9fcb302f0a9fc7497f1b9a)

packages/ui/src/components/Tabs/utils.ts
```diff
@@ -39,6 +39,7 @@ export const commonTabStyles = {
     "data-h2-color": `
       base(black)
       base:children[a:focus-visible](black)
+      base:focus-visible:children[span](black)
```
esizer, line 42: Still purple for me 😢 
Screenshot 2024-01-26 112906

I think what you want is:
```suggestion
      base:focus-visible(black)
```

yonikid15: Are you sure? It's definitely black on my end 😕

yonikid15: That commit works too. Changed in: [dcf7d78](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9077/commits/dcf7d789cf4aaf4c3b103d40236f920a3798edc8)

esizer (approved):

PR #9084, opened 2024-01-25
[Bug] Add toast into context container

### LINKED ISSUE #9037: 🐛 Toasts with Links cause the app to crash

yonikid15:
🤖 Resolves #9037

## 👋 Introduction

- Adds `<Toast />` into ContextContainer, nesting it within the IntlProvider.

## 🧪 Testing

1. Build app `npm run build`
2. Add a component calling useIntl (e.g. Button) into a toast.error or toast.success message. 
3. Ensure component appears without crashing page.
4. Ensure other toast events still function correctly.

Example of change in SupportForm.tsx:
```jsx
      toast.error(
        <Button>
          {intl.formatMessage(
            {
              defaultMessage:
                "Sorry, something went wrong. Please email <anchorTag>{emailAddress}</anchorTag> and mention this error code: {errorCode}.",
              id: "rNVDaA",
              description: "Support form toast message error",
            },
            {
              anchorTag,
              emailAddress: TALENTSEARCH_SUPPORT_EMAIL,
              errorCode: errorMessage,
            },
          )}
        </Button>,
        { autoClose: 20000 },
      );
```
## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/components/Context/ContextProvider.tsx
+++ b/apps/web/src/components/Context/ContextProvider.tsx
@@ -16,6 +16,7 @@ import {
 } from "@gc-digital-talent/i18n";
 import { Announcer } from "@gc-digital-talent/ui";
 import { ThemeProvider } from "@gc-digital-talent/theme";
+import Toast from "@gc-digital-talent/toast";
 
 interface ContextContainerProps {
   messages: Messages;
@@ -28,6 +29,7 @@ const ContextContainer = ({ messages, children }: ContextContainerProps) => (
       <LocaleProvider>
         <AuthenticationProvider>
           <LanguageProvider messages={messages}>
+            <Toast />
             <ThemeProvider>
               <ClientProvider>
                 <AppInsightsProvider>
--- a/apps/web/src/main.tsx
+++ b/apps/web/src/main.tsx
@@ -1,8 +1,6 @@
 import React from "react";
 import { createRoot } from "react-dom/client";
 
-import Toast from "@gc-digital-talent/toast";
-
 import "@gc-digital-talent/forms/dist/index.css";
 import "@gc-digital-talent/ui/dist/index.css";
 import "@gc-digital-talent/toast/dist/index.css";
@@ -17,7 +15,6 @@ if (container) {
     <React.StrictMode>
       <ContextContainer messages={messages}>
         <Router />
-        <Toast />
       </ContextContainer>
     </React.StrictMode>,
   );
--- a/apps/web/src/main.tsx
+++ b/apps/web/src/main.tsx
@@ -17,8 +17,8 @@ if (container) {
     <React.StrictMode>
       <ContextContainer messages={messages}>
         <Router />
+        <Toast />
       </ContextContainer>
-      <Toast />
     </React.StrictMode>,
   );
 }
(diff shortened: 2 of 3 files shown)


esizer (changes requested):
Sadly, it is not this simple 😢 You will need to find a way to work it into our[ context sandwich](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/apps/web/src/components/Context/ContextProvider.tsx). 

You will need to make it a child of `LanguageProvider` but rework it in a way that prevents re-renders. Right now, when any of those contexts needs to re-render, the toast is removed from the screen. That would be why the cypress tests are failing.

yonikid15:
> Sadly, it is not this simple 😢 You will need to find a way to work it into our[ context sandwich](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/apps/web/src/components/Context/ContextProvider.tsx).
> 
> You will need to make it a child of `LanguageProvider` but rework it in a way that prevents re-renders. Right now, when any of those contexts needs to re-render, the toast is removed from the screen. That would be why the cypress tests are failing.
> 
> Also, the example shows a `Button` but the issue identified was with `Link` since we use `react-intl` for the "(opens in a new tab)" hidden text for ATs.

Ah okay thanks! Hmm, I got the page to crash with a Button component as well. Either way, I'll take another look.

esizer:
> Ah okay thanks! Hmm, I got the page to crash with a Button component as well. Either way, I'll take another look.

Sorry, I don't think the original issue wasn't the only problem. Turns out the reason `Button` works and `Link` does not is because `Link` needs to be a child of the `react-router` provider. I don't think this will be possible to fix as it stands. We would need to move `<Toast />` to the all of the main layout components and try and reduce the re-renders happening in the contexts 👎 

I think it would be easier if we just tried to make a rule (whether implicit or via `eslint`) to only use `<a>` in toasts rather than `Link`

yonikid15:
> > Ah okay thanks! Hmm, I got the page to crash with a Button component as well. Either way, I'll take another look.
> 
> Sorry, I don't think the original issue wasn't the only problem. Turns out the reason `Button` works and `Link` does not is because `Link` needs to be a child of the `react-router` provider. I don't think this will be possible to fix as it stands. We would need to move `<Toast />` to the all of the main layout components and try and reduce the re-renders happening in the contexts 👎
> 
> I think it would be easier if we just tried to make a rule (whether implicit or via `eslint`) to only use `<a>` in toasts rather than `Link`

Ya that's true I can add an eslint rule. That should prevent that from happening.
However, I think this fixes the rest of our in house components from crashing when using a toast containing usage of `useIntl()`, which was the scope of the issue.

PR #9086, opened 2024-01-25
[Fix] Dowgrade to `npm@9.9.2` 

### LINKED ISSUE #9085: 🐛 Dependabot does not support NPM 10.
## 🐛 Bug

It turns out that Dependabot does not support NPM 10.  This is why it hasn't been running automatically.  We jumped the gun a little, with #9025.

## 🦋 Expected Behaviour

Dependebot should be running automatically every week.

## 🕵️ Details

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

Downgrade NPM back to 8 or maybe 9.
Can we stay on Node 20 with a lower version of NPM?

## ✅ Acceptance Criteria

- [ ] Dependabot can run automatically again

esizer:
🤖 Resolves #9085 

## 👋 Introduction

This downgrades our `npm` version to 9 but keeps the node update intact.

## 🕵️ Details

Dependabot does not support `npm@10` yet so we are downgrading until it has support for the latest version.

## 🧪 Testing

1. Downgrade your down npm version `npm i npm@9 -g`
2. Install deps `npm i`
3. Build application `npm run dev`
4. Confirm no errors


vd1992:
Seems to work fine for me, though, might be good to check that all fine in Azure too right? @brindasasi

brindasasi:
> Seems to work fine for me, though, might be good to check that all is fine in Azure too right? @brindasasi

looks good to me on my end as well.. yeah good idea to deploy in dev. I can do that now

brindasasi (approved):
it looks good on local & dev vertical. Good to go ! 🥳

PR #9144, opened 2024-01-29
[Bug] Have filtered total display for client side table

### LINKED ISSUE #9061: Showing results values at the bottom of tables sometimes don't reflect the filters
## 🐛 Bug

When filters are applied, some tables to not update the number of "showing" results properly.

This currently works properly on the users table & candidates table, but not on the Pools table or Skills table for example.

## 🦋 Expected Behaviour

When you filter, the "showing" results should always adjust to reflect the filters.

## 🕵️ Details

## 📋 Steps to Reproduce

1. navigate to Pools page on admin UI
2. type something in the filter like "as"
3. Look at the "showing results 1-10..." copy at the bottom of the table. The total number of records does not reflect the filter.
4. try this on the users table to see how it should work

## 📸 Screenshot

Notice that it shows 11-20 of 261. But we are on "Show 10" and there are only 2 pages available for pagination.

Image

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] Filters consistently affect the "showing" results on all tables

vd1992:
🤖 Resolves #9061

## 👋 Introduction

Have the total displayed for client side tables be the result after filters were applied. 
Probed around and found a method to use and a place to plug it in. 

## 🧪 Testing

1. Ensure API-driven tables unaffected
2. Client side table pagination updated, counts in sync with filtering

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
+++ b/apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
@@ -295,6 +295,16 @@ const ResponsiveTable = <TData extends object, TFilters = object>({
       ? nullSearchMessage
       : nullMessage;
 
+  // manipulate pagination object prop as needed
+  // adjust total for client side only, to be post filtering
+  let paginationAdjusted: PaginationDef | undefined = pagination;
+  if (paginationAdjusted?.internal) {
+    paginationAdjusted = {
+      ...paginationAdjusted,
+      total: table.getFilteredRowModel().rows.length,
+    };
+  }
+
   return (
     <>
       <Table.Controls add={add}>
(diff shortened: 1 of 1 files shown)


apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
```diff
@@ -295,6 +295,16 @@ const ResponsiveTable = <TData extends object, TFilters = object>({
       ? nullSearchMessage
       : nullMessage;
 
+  // manipulate pagination object prop as needed
+  // adjust total for client side only, to be post filtering
+  let paginationAdjusted: PaginationDef | undefined = pagination;
+  if (paginationAdjusted?.internal) {
+    paginationAdjusted = {
```
vd1992, line 302: API tables appeared wonky to me if I didn't narrow it down to internal only.

esizer (approved):
LGTM! 👍

PR #9152, opened 2024-01-31
[Fix] Update user skill return path

### LINKED ISSUE #9062: Skills in the showcase bring you back to the library
## 🐛 Bug

When you click a skill in the **Skills showcase**, if you click "save", you are brought to the **skills library**.

## 🦋 Expected Behaviour

When you click a skill in the **Skills showcase**, if you save it, you are brought back to the **skills showcase**.

## 📋 Steps to Reproduce

1. Go to skills showcase
2. Click a skill
3. Click save
4. You are brought to the Skills library

## 🙋‍♀️ Proposed Solution

Use a "from" key parameter in url.

## ✅ Acceptance Criteria

- [ ] Saving a skill in skills showcase brings you back to the place you came from (ie showcase OR skills library, depending.)
- [ ] Check if you can get to this skill page from anywhere else, and unsure we send you there correctly as well.

esizer:
🤖 Resolves #9062 

## 👋 Introduction

This updates the update user skill page to return to the skill showcase when that is where the user came from. In addition, the breadcrumbs were not accurate so have been updated to also indicate that the skill showcase is included when coming from there.

## 🕵️ Details

It seems that the user can only get here from the library or showcase. So, instead of using a path I simply used "showcase" so we can update the return path and breadcrumbs based on that.

## 🧪 Testing

1. Build `npm run dev`
2. Login as an applicant
3. Navigate to your skill showcase `/applicant/profile-and-applications/skills/showcase`
4. If a skill does not exist yet, add one
5. Click on a skill title to edit it
6. Confirm the a `?from=showcase` appears in the URL
7. Confirm the breadcrumbs appear as expected
8. Save the form
9. Confirm you are returned to the showcase and not library
10. Navigate to the skill library `/applicant/profile-and-applications/skills`
11. Click on a skill to edit it
12. Save the form
13. Confirm you are returned to the library

### DIFF
--- a/apps/web/src/components/SkillRankCard/SkillRankListItem.tsx
+++ b/apps/web/src/components/SkillRankCard/SkillRankListItem.tsx
@@ -19,10 +19,12 @@ interface SkillLinkProps {
 
 const SkillLink = ({ id, children }: SkillLinkProps) => {
   const paths = useRoutes();
+  const searchParams = new URLSearchParams();
+  searchParams.set("from", "showcase");
 
   return (
     <Link
-      href={paths.editUserSkill(id ?? "")}
+      href={`${paths.editUserSkill(id ?? "")}?${searchParams.toString()}`}
       mode="text"
       color="black"
       data-h2-text-align="base(left)"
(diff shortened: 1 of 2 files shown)


petertgiles (approved):
:+1:  Works well

PR #9158, opened 2024-01-31
[Feature] Basic filters for `AssessmentStepTracker`

### LINKED ISSUE #9016: ✨ Add basic filters to Assessment Step tracker
## ✨ Feature

Add filters to the Assessment Step tracker which allows for filtering by name, and hiding various statuses.

## 🕵️ Details

Add a search bar which filters users by name, to quickly find the status of a particular candidate.

Add switches which allow for hiding users in certain statuses from each column.

The figma design includes a "More filters" button, but we can leave it out for now.

## 🎨 Design File

https://www.figma.com/file/em5ArPr044YaIe1Vb1zCZz/Pools---Record-of-Decision?type=design&node-id=6872-37861&mode=design&t=A1V8ZwHD6eu2g4XZ-0

### 🧑‍🎨 Designer

@Jerryescandon 

## 📸 Screenshot

image

The filters could be build as part of the Assessment Step Tracker component or in its parent component.

The search bar can filter out candidates before even passing them to the (rest of the) Assessment Step Tracker.

The statuses to hide (or show) can be passed into the individual Assessment Step Tracker columns as props, and they can simply avoid showing any candidates (and totals) in those statuses.

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Search bar above the Assessment Step Tracker. If any text is in the search bar, filter out any candidates whose names do not match (case insensitive)
- [ ] 4 switches for 4 statuses, default to ON. If any switch is turned OFF, then each assessment step hides any candidates in that status for that step. (Also hides the total count for that step.)

esizer:
🤖 Resolves #9016 

## 👋 Introduction

Adds some basic filters for the `AssessmentStepTracker` component.

- Candidate name search
- Status (decision) toggles

## 🕵️ Details

The test I wrote only test one filter at a time. Would it be worth it to try some combinations of filters? 🤔 

## 🧪 Testing

1. Build `npm run dev`
2. Open storybook `npm run storybook:web`
3. Navigate to the "Assessment Step Tracker" story
4. Play around with the filters
5. Confirm they act as you expect

## 📸 Screenshot

Screenshot 2024-01-31 143341

### DIFF
--- a/apps/web/src/components/AssessmentStepTracker/AssessmentStepTracker.tsx
+++ b/apps/web/src/components/AssessmentStepTracker/AssessmentStepTracker.tsx
@@ -14,8 +14,9 @@ import {
   ResultFilters,
   filterResults,
   groupPoolCandidatesByStep,
+  defaultFilters,
 } from "./utils";
-import Filters, { initialValues } from "./Filters";
+import Filters from "./Filters";
 
 export interface AssessmentStepTrackerProps {
   pool: Pool;
@@ -23,7 +24,7 @@ export interface AssessmentStepTrackerProps {
 
 const AssessmentStepTracker = ({ pool }: AssessmentStepTrackerProps) => {
   const intl = useIntl();
-  const [filters, setFilters] = React.useState<ResultFilters>(initialValues);
+  const [filters, setFilters] = React.useState<ResultFilters>(defaultFilters);
   const steps = unpackMaybes(pool.assessmentSteps);
   const candidates = unpackMaybes(pool.poolCandidates);
   const groupedSteps = groupPoolCandidatesByStep(steps, candidates);
--- a/apps/web/src/components/AssessmentStepTracker/AssessmentStepTracker.tsx
+++ b/apps/web/src/components/AssessmentStepTracker/AssessmentStepTracker.tsx
@@ -4,13 +4,18 @@ import { useIntl } from "react-intl";
 import { Board } from "@gc-digital-talent/ui";
 import { getLocalizedName } from "@gc-digital-talent/i18n";
 import { unpackMaybes } from "@gc-digital-talent/helpers";
+import { Pool } from "@gc-digital-talent/graphql";
 
-import { Pool } from "~/api/generated";
 import applicationMessages from "~/messages/applicationMessages";
 
 import ResultsDetails from "./ResultsDetails";
 import AssessmentResults from "./AssessmentResults";
-import { groupPoolCandidatesByStep } from "./utils";
+import {
+  ResultFilters,
+  filterResults,
+  groupPoolCandidatesByStep,
+} from "./utils";
+import Filters, { initialValues } from "./Filters";
 
 export interface AssessmentStepTrackerProps {
   pool: Pool;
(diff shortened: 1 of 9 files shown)


apps/web/src/components/AssessmentStepTracker/AssessmentStepTracker.tsx
```diff
@@ -4,40 +4,54 @@ import { useIntl } from "react-intl";
...
 const AssessmentStepTracker = ({ pool }: AssessmentStepTrackerProps) => {
   const intl = useIntl();
+  const [filters, setFilters] = React.useState<ResultFilters>(defaultFilters);
   const steps = unpackMaybes(pool.assessmentSteps);
   const candidates = unpackMaybes(pool.poolCandidates);
   const groupedSteps = groupPoolCandidatesByStep(steps, candidates);
 
+  const filteredSteps = React.useMemo(() => {
+    return filterResults(filters, groupedSteps);
+  }, [groupedSteps, filters]);
```
tristan-orourke, line 34: Is it worth memoizing this if groupedSteps isn't also memoized?

apps/web/src/components/AssessmentStepTracker/utils.ts
```diff
@@ -451,3 +426,72 @@ export const groupPoolCandidatesByStep = (
...
+  query: "",
+  [NO_DECISION]: true,
+  [AssessmentDecision.Successful]: true,
+  [AssessmentDecision.Hold]: true,
+  [AssessmentDecision.Unsuccessful]: true,
+};
+
+const statusIsAvailable = (filter?: boolean) => {
+  return typeof filter !== "undefined" && filter;
+};
```
tristan-orourke, line 448: Anywhere we need to check if `filter` is true, couldn't we just use `if (filter)` directly? And it not, then `!!filter`?

apps/web/src/components/AssessmentStepTracker/utils.ts
```diff
@@ -451,3 +426,72 @@ export const groupPoolCandidatesByStep = (
...
+          case AssessmentDecision.Unsuccessful:
+            available = statusIsAvailable(
+              filters[AssessmentDecision.Unsuccessful],
+            );
+            break;
+          default:
+            break;
+        }
+
+        return available;
```
tristan-orourke, line 488: I haven't tested, but I'm pretty sure this is equivalent. It just needs some comments explaining why it works. Here's my first attempt.
```suggestion
        // `filters` shows which decisions we want to show.
        // If `filters.decision` is false, that we means we want to filter out all candidates with that decision, so they don't appear in results.
        return filters[decision];
```

tristan-orourke (changes requested):
Looks good! Just one suggestion and a question.

apps/web/src/components/AssessmentStepTracker/AssessmentStepTracker.tsx
esizer: Yeah, that makes sense. Removed in [remove memoization](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9158/commits/903299c8eb91350d66ab89acef72b7415da7b512)

apps/web/src/components/AssessmentStepTracker/utils.ts
esizer: Oh yeah, that works as well 😅 I was trying to commit suggestion but since I touched that file it couldn't 👎 

Updated in [simplify status filter further](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9158/commits/37cb3868cc1277d953379c3d81725f723a6eaf38)

tristan-orourke (approved):

PR #9159, opened 2024-01-31
[Test] Login-Logout Tests

### LINKED ISSUE #8654: 🧪 Improve login/logout tests
## 🧪 Tests

Our login/logout process currently has poor test coverage.  A thorough test requires a lot of manual time.  It would be better if a test suite ran in CI, possibly a combination of unit tests and e2e tests.

A big batch of unit tests had to be removed when we moved a lot of the ClientProvider functions inline in the client setup.  It would be good to get the test coverage back and even expand some unit tests.
https://github.com/GCTC-NTGC/gc-digital-talent/pull/8618/commits/dddbd755bc6aac28cd66cd3dc834bf2fc51866cd

A pretty good set of manual tests that would be worth automating:
- [ ] Can log in.
- [ ] Can wait five minutes and then the token is refreshed on the next API call.
- [ ] ~When you have two tabs open, a refresh in one will allow the second tab to make an API call with the new tokens and no refresh.~
- [ ] Can log out (with SiC endsession).
- [ ] If [introspection fails](https://github.com/GCTC-NTGC/gc-digital-talent/pull/6547#pullrequestreview-1602788860) the user is immediately logged out.
- [ ] Breaking refresh by restarting the auth container -> leads to logged out page
- [ ] If you log in as a deleted user you end up on the "user deleted" page.
- [ ] ~Logout appears to affect all logged in tabs~
- [ ] ~can chain two refreshes~

## ✅ Acceptance Criteria

- [ ] Improve test coverage of the login/logout code

petertgiles:
🤖 Resolves #8654

## 👋 Introduction

This branch adds a number of new Cypress tests in a new login-logout test suite.  It also factors out a client function to restore some previously removed Jest tests.

## 🧪 Testing

> [!Tip]
You can run a Cypress spec multiple times using the [cypress-repeat](https://github.com/bahmutov/cypress-repeat) command.
`npx cypress-repeat run -n 10 --spec cypress/e2e/api/login-logout.cy.ts `

1. Tests reliably complete.
2. Tests test what they're supposed to be testing.
3. No tests missing.
4. Runtime login-logout functionality unaffected


docker-compose.yml
```diff
@@ -86,7 +86,7 @@ services:
   #
   mock-auth:
     # See: https://github.com/navikt/mock-oauth2-server#standalone-server
-    image: ghcr.io/navikt/mock-oauth2-server:0.4.8
+    image: ghcr.io/navikt/mock-oauth2-server:2.1.1
     volumes:
```
vd1992, line 90: I think it will be important to rebuild the image as well, so running
`docker-compose up --detach --build`

vd1992: I deleted the container/image, then ran the `--build` command to ensure I got it updated, I don't know if it was needed but that was what I did.

apps/e2e/cypress/e2e/api/login-logout.cy.ts
```diff
@@ -0,0 +1,416 @@
...
+        name: "Account deleted",
+        level: 1,
+      }).should("exist");
+    });
+  });
+
+  context("refresh", () => {
+    // will log in and refresh the token successfully
+    it.only("can refresh the token", () => {
+      cy.intercept({ pathname: "/refresh*" }).as("refresh");
```
vd1992, line 142: Accidental `only` ?

image

vd1992: Same of course happens locally

petertgiles: Doh!  It's too easy to do that.
[remove .only](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9159/commits/d74cb186d250b5b26e2f192e28578b58bf213136)

vd1992:
I see this sometimes locally too when running, but the specs are passing. Is this of value or noise?
From the test "can chain two refreshes"

image

vd1992:
> > From the test "can chain two refreshes"
> 
> Hmm, I can't reproduce this locally. Might be tricky to sus out. Also, how is this test passing? It is clearly bailing out part way through. 😵‍💫

Odd.
I can get it to appear somewhat reliably by slapping the `only` on the test in question then using the run 10 times command.

Multiple attempts happen sometimes, not always. 

image

But it still succeeds

image

petertgiles:
OK, my third attempt - time travel!  I've set cypress to change the clock to trigger a refresh.
image

vd1992:
Seems to all work fine to me! Nice work!
I'd like someone else to also take a look and try, so I will defer approval.

esizer:
> Turns out OAUTH is set up to make tampering with the tokens very, very hard. 😆

I guess that is a good thing, even if it is annoying for tests 😛

esizer (approved):
Nice! 🤩 

Love the new `ClientProvider` tests as well 👍

PR #9183, opened 2024-02-01
[Fix] Update position of skill error

### LINKED ISSUE #9027: 🐛 Skill requirement step on the application has weird error placement
## 🐛 Bug

The placement of the error on the skills step in the application makes it look like the error is specific to the "Optional behavioural skills" section rather than the page as a whole.

## 🦋 Expected Behaviour

The error should be positioned in a way that it is not visually associated with a specific section (top of the page, or below the divider line).

## 📸 

<details>
  <summary>📸 Current layout</summary>
  
</details> 

<details>
  <summary>📸 Proposed layout</summary>
  
</details>

esizer:
🤖 Resolves #9027 

## 👋 Introduction

Moves the error message for skill requirements on the application so that it no longer appears to be associated with the optional skills.

## 🧪 Testing

1. Build `npm run dev`
2. Login and an applicant `applicant@test.com`
3. Start an application
4. Continue through the process until the "Skill requirements" page
5. Force an error
6. Scroll to the bottom and confirm the error message appears below the `Separator`

## 📸 Screenshot

Screenshot 2024-02-01 081541

### DIFF
--- a/apps/web/src/pages/Applications/ApplicationSkillsPage/ApplicationSkillsPage.tsx
+++ b/apps/web/src/pages/Applications/ApplicationSkillsPage/ApplicationSkillsPage.tsx
@@ -322,27 +322,30 @@ export const ApplicationSkills = ({
       ) : null}
       <FormProvider {...methods}>
         <form onSubmit={methods.handleSubmit(handleSubmit)}>
-          <Input
-            id="skillsMissingExperiences"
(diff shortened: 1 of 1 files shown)


esizer:
> Question: Was the error message always displaying before the user even clicks the save button?

I'm actually not sure. Looking at the code, I imagine so 🤔

PR #9187, opened 2024-02-01
[Fix] Prevent `PoolCandidate` authorized to view scope error

### LINKED ISSUE #9153: 🐛 Prod logged error - Attempt to read property on null
## 🐛 Bug

We occasionally have exceptions logged in production that look like this:

> Attempt to read property "id" on null {"exception":"[object] (ErrorException(code: 0): Attempt to read property \"id\" on null at /home/site/wwwroot/api/app/Models/PoolCandidate.php:677)

image

It seems like this scope is being called with no authorized user present?

## 🦋 Expected Behaviour

Uncaught framework exceptions from Eloquent models should not be happening in production under normal circumstances

## 📋 Steps to Reproduce

_Unknown_

## 🙋‍♀️ Proposed Solution

1. Find out how to reproduce it
2. Decide on the solution
  - If it is a normal occurrence then prevent the exception from being thrown, or catch and deal with it
  - If it is due to a bug, fix the bug

## ✅ Acceptance Criteria

- [ ] It is no longer possible for this exception to occur in prod

esizer:
🤖 Resolves #9153 

## 👋 Introduction

This adds better protection to the `Pool.poolCandidates` field in our schema.

## 🕵️ Details

As far as Peter and I could tell, this was happening during our auth flakiness the other week(?). What was happening was, authenticated users were accessing the `Pool.poolCandidates` field in some weird time when they were not authenticated.

Since our `pools` query is not guarded, they could access that.  But we did not have a `@guard` directive on the `poolCandidates` field. The `@can` should have saw no user and rejected it but for some reason, the scope is evaluated first here 🤷‍♀️ .

So, this did two things:

- Added a `@guard` to `Pool.poolCandidates` (we only want authenticated users accessing this anyway)
- Did a quick check in the scope to make sure a user exists just in case (this is a little odd but extra careful)

## 🧪 Testing

```graphql
query {
  pools {
    id
    poolCandidates {
      id
    }
  }
}
```

1. Ensure you have no default auth user in the `api/.env` file
2. Navigate to `/graphiql`
3. Run the provided quiery
4. Confirm you get "unauthenticated" error
5. Remove the poolCandidates field from the query
6. Confirm you can still get pools

## 📸 Screenshot

Screenshot 2024-02-01 131303

### DIFF
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -675,8 +675,12 @@ class PoolCandidate extends Model
      */
     public function scopeAuthorizedToView(Builder $query)
     {
-        $userId = Auth::user()->id;
-        $user = User::find($userId);
+        $authUser = Auth::user();
+        if (! $authUser) {
+            return $query->where('id', null);
+        }
+
+        $user = User::find($authUser->id);
         if (! $user->isAbleTo('view-any-application')) {
             $query->where(function (Builder $query) use ($user) {
                 if ($user->isAbleTo('view-any-submittedApplication')) {
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -675,7 +675,9 @@ class PoolCandidate extends Model
      */
     public function scopeAuthorizedToView(Builder $query)
     {
-        $user = User::find(Auth::id());
+        /** @var \App\Models\User */
+        $user = Auth::user();
+
         if (! $user) {
             return $query->where('id', null);
         }
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -675,12 +675,11 @@ class PoolCandidate extends Model
      */
     public function scopeAuthorizedToView(Builder $query)
     {
-        $authUser = Auth::user();
-        if (! $authUser) {
+        $user = User::find(Auth::id());
+        if (! $user) {
             return $query->where('id', null);
         }
 
-        $user = User::find($authUser->id);
         if (! $user->isAbleTo('view-any-application')) {
             $query->where(function (Builder $query) use ($user) {
                 if ($user->isAbleTo('view-any-submittedApplication')) {
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -871,6 +871,7 @@ class User extends Model implements Authenticatable, LaratrustUser
 
     public function scopeAuthorizedToView(Builder $query)
     {
+        /** @var \App\Models\User */
         $user = Auth::user();
 
         if (! $user) {
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -173,6 +173,7 @@ type Pool {
   poolCandidates: [PoolCandidate]
     @hasMany(relation: "publishedPoolCandidates", scopes: ["authorizedToView"])
     @can(ability: "view", resolved: true)
+    @guard
   keyTasks: LocalizedString @rename(attribute: "key_tasks")
   yourImpact: LocalizedString @rename(attribute: "your_impact")
   whatToExpect: LocalizedString @rename(attribute: "what_to_expect")


api/app/Models/PoolCandidate.php
```diff
@@ -675,8 +675,12 @@ public function isDraft()
      */
     public function scopeAuthorizedToView(Builder $query)
     {
-        $userId = Auth::user()->id;
-        $user = User::find($userId);
+        $authUser = Auth::user();
+        if (! $authUser) {
+            return $query->where('id', null);
```
petertgiles, line 680: I don't think I understand `$query->where('id', null);`.  Is this like a "where true = false" expression?  Why not just throw an exception and abort the query?

esizer: I'm not sure, I kind of just assumed we would want the query to run on a happy path and just return an empty result. So if the rest of the query is okay it won't crash a page and we are still protecting data.

If we are okay with a panic response, I can throw an exception instead.

tristan-orourke:
> Since our pools query is not guarded, they could access that. But we did not have a @guard directive on the poolCandidates field. The @can should have saw no user and rejected it but for some reason, the scope is evaluated first here 🤷‍♀️ .

If the `@can` directive has `resolved: true` then it can only be evaluated after the DB query runs, which is why the scope was evaluated first.

esizer:
> If the `@can` directive has `resolved: true` then it can only be evaluated after the DB query runs, which is why the scope was evaluated first.

Right, that makes sense. Thanks 👏

api/app/Models/PoolCandidate.php
tristan-orourke: Yes, exactly, this is a quick way to force the query to return zero results. We do the same thing in `User.scopeAuthorizedToView`. We decided that scopes should not be concerned with doing authorization - if a user isn't permitted to see any results, the correct behaviour for the scope is to return no results. As eric says, that allows other parts of the query to return without crashing.

api/app/Models/PoolCandidate.php
```diff
@@ -675,8 +675,12 @@ public function isDraft()
...
+        }
+
+        $user = User::find($authUser->id);
```
tristan-orourke, line 683: Aren't $authUser and $user identical? What's the purpose of this line?

tristan-orourke (changes requested):

api/app/Models/PoolCandidate.php
esizer: `Auth::user()` was giving the error `Undefined method 'isAbleTo'`. Though I do see we have that in `User.php` and just ignore it. I assume it is safe to ignore that error? 🤔

petertgiles: Technically, what we get back is an Authenticatable, not a User.
image

In practice, for us, that's always a user but maybe it won't have the Eloquent relationships there?

tristan-orourke: Darn, I see! Does php actually throw an error, or is it just linting which _thinks_ it will?
My main concern is if this forces a repeat of the database call.
Looking around online, I haven't found any suggestions for how to change the return type of Auth::user. Some people have suggested that `$user = User::find(Auth::id())` _might_ be more efficient, if we happen to have the id in memory but not the user yet, but I doubt that's the case for us.

esizer: So, we _know_ we return a user from our provider. 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/c97149e361df11313767f153f06c0535b01b1112/api/config/auth.php#L66-L70

https://github.com/GCTC-NTGC/gc-digital-talent/blob/c97149e361df11313767f153f06c0535b01b1112/api/app/Providers/AuthServiceProvider.php#L98-L117

Could we just type hint this to make the errors go away? I think Intelephense is just flagging this because it is not aware of the middleware which could return anything that extends `Authenticatable` 

```php
/** @var \App\Models\User */
$user = Auth::user();
```

tristan-orourke: Yeah, if it doesn't throw an error at runtime, than I'm fine with typehinting it or ignoring the lint error, etc

esizer: Cool! Updated in [use type hint to resolve error, prevent additional db hit](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9187/commits/9452cdaf5f77bf630c3cf7ab924ee3ec038a66fe)

tristan-orourke (approved):

PR #9222, opened 2024-02-06
Enable hiding of status column if Record of Decision flag is true

### LINKED ISSUE #9221: 🐛 Status column can't be shown on Pool Candidate table post RoD
## 🐛 Bug

When the Record of Decision feature flag is true, the Status column of the Pool Candidate table is hidden by default but can't be un-hidden

## 🦋 Expected Behaviour

Status column should appear in the show/hide column menu

## 🕵️ Details

## 📋 Steps to Reproduce

1. ...
2. ...

## 📸 Screenshot

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] Status column should appear in the show/hide column menu when Record of Decision flag is true
- [ ] Status column should not be hideable if RoD is false (I assume we did that for a reason)
- [ ] Status column should be hidden by default if RoD is true.

tristan-orourke:
🤖 Resolves #9221

## 👋 Introduction

Status column wasn't able to be shown on the Pool Candidates table with Record of Decision flag on.

## 🧪 Testing

1. With Record of Decision flag OFF, status column is shown on the Pool Candidates table and locked, cannot be hidden
2. Turn Record of Decision flag ON and rebuild
3. Refresh Pool Candidates table page
4. Status column is hidden, but can be shown.

### DIFF
--- a/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
+++ b/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
@@ -426,7 +426,7 @@ const PoolCandidatesTable = ({
       {
         id: "status",
         header: intl.formatMessage(tableMessages.status),
-        enableHiding: false,
+        enableHiding: !recordOfDecision, // After record of decision is turned on, we can remove this property entirely (it defaults to true)
         cell: ({
           row: {
             original: { poolCandidate },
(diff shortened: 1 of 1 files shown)


apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
```diff
@@ -426,7 +426,7 @@ const PoolCandidatesTable = ({
       {
         id: "status",
         header: intl.formatMessage(tableMessages.status),
-        enableHiding: false,
+        enableHiding: !recordOfDecision, // After record of decision is turned on, we can remove this property entirely (it defaults to true)
```
vd1992, line 429: Should it not just be `enableHiding: recordOfDecision` ?
Now RoD as false resolves to true making it hide-able 

Does not meet goal
"Status column should not be hideable if RoD is false"

PR #9225, opened 2024-02-07
[Bug] Update countApplicants query base scopes

### LINKED ISSUE #9160: 🐛 countApplicants query missing "Active" scope
## 🐛 Bug

The countApplicants query counts users, not applicants. It automatically includes the `inITPublishingGroup` which restricts the count to users with at least application to an IT pool. However, this scope still allows for candidates who are expired, screened-out, suspended, or even still in draft.

## 🦋 Expected Behaviour

Even without filters, `countApplicants` should include the same base scopes as `CountPoolCandidatesByPool`: that is scopeAvailable, scopeExpiryStatus and scopeInITPublishingGroup.

Side note: why isn't checking for expiry integrated into scopeAvailable?

## 📋 Steps to Reproduce

1. Go to /graphiql
2. Run the query `countApplicants(where: {})`, and record the count
3. Go to /search
4. The number of applicants shown before adding any filters should match step 2.

## 📸 Screenshot

## 🙋‍♀️ Proposed Solution

```
# In User.php

  /**
     * Return users who have an available PoolCandidate in at least one pool.
     *
     * @param Builder $query
     * @return Builder
     */
    public static function scopeAvailableInSomePool(Builder $query): Builder
    {
        return $query->whereHas('poolCandidates', function ($innerQueryBuilder) {
            PoolCandidate::scopeAvailable($innerQueryBuilder);
            PoolCandidate::scopeExpiryStatus($innerQueryBuilder, CandidateExpiryFilter::ACTIVE->name);
            return $innerQueryBuilder;
        });
    }
```

```
# in schema.graphql

  countApplicants(where: ApplicantFilterInput): Int!
    @count(
      model: "User"
      scopes: ["inITPublishingGroup", "availableInSomePool"]
    )
```

## ✅ Acceptance Criteria

- [ ] countApplicants only ever counts users who are currently "available" in some pool
- [ ] regression test

JamesHuf:
🤖 Resolves #9160

## 👋 Introduction

Updates the base scopes for the countApplicants query to match the countPoolCandidatesByPool query.

## 🕵️ Details

The steps to reproduce in the issue will still return a different number of candidates due to candidates in archived pools. The base query includes those pools but the search page does not.

## 🧪 Testing

The easiest way to test this is probably just ensuring the tests match desired behavior and all pass. Removing any archived pools from your environment should also cause the steps to reproduce in the issue to pass.

### DIFF
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -452,6 +452,9 @@ class PoolCandidate extends Model
             ->where(function ($query) {
                 $query->where('suspended_at', '>=', Carbon::now())
                     ->orWhereNull('suspended_at');
+            })
+            ->where(function ($query) {
+                self::scopeExpiryStatus($query, CandidateExpiryFilter::ACTIVE->name);
             });
 
         return $query;
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -6,7 +6,6 @@ use App\Enums\CandidateExpiryFilter;
 use App\Enums\CandidateSuspendedFilter;
 use App\Enums\LanguageAbility;
 use App\Enums\PoolCandidateStatus;
-use App\Enums\PublishingGroup;
 use Carbon\Carbon;
 use Illuminate\Auth\Authenticatable as AuthenticatableTrait;
 use Illuminate\Contracts\Auth\Authenticatable;
@@ -595,15 +594,13 @@ class User extends Model implements Authenticatable, LaratrustUser
 
     /**
      * Return users who have an available PoolCandidate in at least one IT pool.
-     *
-     * @param Builder $query
-     * @return Builder
      */
     public static function scopeAvailableInITPool(Builder $query): Builder
     {
         return $query->whereHas('poolCandidates', function ($innerQueryBuilder) {
             PoolCandidate::scopeAvailable($innerQueryBuilder);
             PoolCandidate::scopeInITPublishingGroup($innerQueryBuilder);
+
             return $innerQueryBuilder;
         });
     }
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -595,7 +595,7 @@ class User extends Model implements Authenticatable, LaratrustUser
     /**
      * Return users who have an available PoolCandidate in at least one IT pool.
      */
-    public static function scopeAvailableInITPool(Builder $query): Builder
+    public static function scopeAvailableInITPublishingGroup(Builder $query): Builder
     {
         return $query->whereHas('poolCandidates', function ($innerQueryBuilder) {
             PoolCandidate::scopeAvailable($innerQueryBuilder);
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -594,22 +594,18 @@ class User extends Model implements Authenticatable, LaratrustUser
     }
 
     /**
-     * Scope is IT
+     * Return users who have an available PoolCandidate in at least one IT pool.
      *
-     * Restrict a query by pool candidates that are for pools
-     * containing IT specific publishing groups
-     *
-     * @param  \Illuminate\Database\Eloquent\Builder  $query  The existing query being built
-     * @return \Illuminate\Database\Eloquent\Builder The resulting query
+     * @param Builder $query
+     * @return Builder
      */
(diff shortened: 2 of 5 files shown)


api/tests/Feature/ApplicantTest.php
```diff
@@ -52,26 +52,53 @@ public function testCountApplicantsQuery(): void
     {
         // Get the ID of the base admin user
         $user = User::All()->first();
-        $pool1 = Pool::factory()->candidatesAvailableInSearch()->create([
+        $ITPool1 = Pool::factory()->candidatesAvailableInSearch()->create([
+            'user_id' => $user['id'],
```
tristan-orourke, line 56: I think you also need to set the classification of these pools

api/app/Models/User.php
```diff
@@ -594,22 +593,16 @@ public static function scopePublishingGroups(Builder $query, ?array $publishingG
...
-     * Scope is IT
-     *
-     * Restrict a query by pool candidates that are for pools
-     * containing IT specific publishing groups
-     *
-     * @param  \Illuminate\Database\Eloquent\Builder  $query  The existing query being built
-     * @return \Illuminate\Database\Eloquent\Builder The resulting query
+     * Return users who have an available PoolCandidate in at least one IT pool.
      */
-    public static function scopeInITPublishingGroup(Builder $query)
```
tristan-orourke, line 605: I think I'd rather see the `scopeInITPublishingGroup` left as is. There's no need for a scope which JUST calls two other scopes, with no additional information. I think overly nesting scopes like that may just obfuscate what's happening.

api/graphql/schema.graphql
```diff
@@ -684,7 +684,7 @@ type Query {
     )
   # countApplicants returns the number of candidates matching its filters, and requires no special permissions.
   countApplicants(where: ApplicantFilterInput): Int!
-    @count(model: "User", scopes: ["inITPublishingGroup"])
```
tristan-orourke, line 687: I think I'd rather see two scopes used here (inItPublishingGroup, and available) instead of one scope which just combines them.

api/app/Models/PoolCandidate.php
```diff
@@ -452,6 +452,9 @@ public static function scopeAvailable(Builder $query): Builder
             ->where(function ($query) {
                 $query->where('suspended_at', '>=', Carbon::now())
                     ->orWhereNull('suspended_at');
+            })
+            ->where(function ($query) {
```
tristan-orourke, line 456: This is a case where a scope calling another scope makes sense, I think 👍

tristan-orourke (changes requested):
In addition to my suggestion to reference both scopes (inIT and available) instead of combining them into one scope, we could also use tests for the countPoolCandidatesByPool query.

api/tests/Feature/ApplicantTest.php
JamesHuf: I could set a classification but the scope just checks the pools publishing group.

api/app/Models/User.php
JamesHuf: I think making this two separate scopes would cause users with an unavailable IT candidate but available other candidate to be counted here.

JamesHuf:
> we could also use tests for the countPoolCandidatesByPool query.

We already have [countPoolCandidatesPyPoolTest](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/api/tests/Feature/CountPoolCandidatesByPoolTest.php) the last 3 cover the scopes basic scopes relevant here.

api/tests/Feature/ApplicantTest.php
tristan-orourke: I see 👍

api/app/Models/User.php
```diff
@@ -594,22 +593,16 @@ public static function scopePublishingGroups(Builder $query, ?array $publishingG
...
+    public static function scopeAvailableInITPool(Builder $query): Builder
```
tristan-orourke, line 598: Okay, I see now why you combined the two scopes! Could you rename this to `scopeAvailableInITPublishingGroup`? The distinction between IT Classification and IT Publishing group is confusing enough (as shown by me mistaking them several times in my review!) so lets be explicit about using publishing group.

tristan-orourke (approved):
I see now why you did things this way! Please just rename the one scope for clarity before merging.

PR #9240, opened 2024-02-07
[Fix] Combox clear button resets options

### LINKED ISSUE #9239: 🐛 Search page skills clear doesn't reset options in dropdown
## 🐛 Bug

We fixed the issue with the "X" button not clearing the input but, it seems as though it is not resetting the options.

## 🦋 Expected Behaviour

Options are also reset when the field is cleared with the button.

## 📋 Steps to Reproduce

1. Navigate to `/search`
2. Type in the skills combobox
3. Click the clear
4. Without typing, open the dropdown again
5. Observe the dropdown is not reset to the default

## ✅ Acceptance Criteria

- [ ] Options are also reset when the field is cleared with the button.

esizer:
🤖 Resolves #9239 

## 👋 Introduction

Fixes the clear button on the mutli combobox to also reset the options in the dropdown.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to `/search`
3. Type in the skill combobox
4. Click the "X" to clear
5. Without typing, open the dropdown
6. Confirm the options were reset

### DIFF
--- a/packages/forms/src/components/Combobox/Multi.tsx
+++ b/packages/forms/src/components/Combobox/Multi.tsx
@@ -193,8 +193,7 @@ const Multi = ({
   }, [options]);
 
   const handleClear = () => {
-    setInputValue("");
-    onInputChange?.("");
+    handleInputChanged("");
     inputRef?.current?.focus();
   };
 


vd1992 (approved):
👍

PR #9242, opened 2024-02-07
[Fix] Step tracker search case insensitve

### LINKED ISSUE #9241: 🐛 Step tracker search is case sensitive
## 🐛 Bug

The new search form on the step tracker is case sensitive when it shouldn't be.

## 🦋 Expected Behaviour

Search field should be case insensitive.

## 📋 Steps to Reproduce

2. Navigate to the "Assessment Step Tracker" story
3. Type in a search that has differing cases to the results displayed
4. Confirm the result is filtered out 

## 🙋‍♀️ Proposed Solution

Use `string.toLowerCase()` on the values.

## ✅ Acceptance Criteria

- [ ] Search field returns results regardless of the query case

esizer:
🤖 Resolves #9241 

## 👋 Introduction

Fixes the search on the step tracker to be case insensitve

## 🧪 Testing

2. Navigate to the "Assessment Step Tracker" story
3. Use the search typing in queries that do not match the case of results
4. Confirm they appear in the results regardless of case

## 📸 Screenshot

Screenshot 2024-02-07 160407

### DIFF
--- a/apps/web/src/components/AssessmentStepTracker/utils.ts
+++ b/apps/web/src/components/AssessmentStepTracker/utils.ts
@@ -247,8 +247,9 @@ export const filterResults = (
         if (filters.query) {
           const fullName = [user.firstName, user.lastName]
             .filter(notEmpty)
-            .join(" ");
-          if (!fullName.includes(filters.query)) {
+            .join(" ")
+            .toLowerCase();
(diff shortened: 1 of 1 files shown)


vd1992 (approved):
Works for me

Issue #9051, opened 2024-01-18
🐛 Out of memory error on pool candidates csv download


brindasasi:
## 🐛 Bug

When trying to download 500 candidates CSV file , out of memory error occurs. This can be reproduced in Azure env. with large amount of data. 
Below is the complete error log. 

[outofmemoryError.txt](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNVBhQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--295ed8d00f2bfb6a44fb5bf41ea0c0e61ab512b1/outofmemoryError.txt)

## 🦋 Expected Behaviour

Download should work for large amount of data without any issue

## 🕵️ Details

Add any additional details that could assist with troubleshooting/fixing the issue.

- **Operating System**: Name and version (if possible).
- **Browser**: Name and version (if possible).

## 📋 Steps to Reproduce

1. Generate around 600 pool candidates
2. Try to download by 500s

## 📸 Screenshot

Add a screenshot (if possible).

## 🙋‍♀️ Proposed Solution

(optional) Do you have a proposed solution?

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the bug was addressed.

- [ ] - Poolcandidates query needs to be optimized 
- [ ] - Laravel queues needs to be implemented for high volume download
- [ ] - Bulk data download should work without any issues

brindasasi:
related #8974

brindasasi:
`PoolCandidatesTable_SelectPoolCandidates` triggers over 100 sql queries and finishes with 
[2024-01-25 20:14:54] local.ERROR: Maximum execution time of 30 seconds exceeded {"userId":"06be6b14-86ce-4585-b70f-36fa25fe7c94","exception":"[object] `(Symfony\\Component\\ErrorHandler\\Error\\FatalError(code: 0): Maximum execution time of 30 seconds exceeded at /home/site/wwwroot/api/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php:1939)`

for 10 candidates CSV download 😭

PR #9288, opened 2024-02-08
9051-poolcandidates-csv-download-optimization

brindasasi:
🤖 Resolves {issue #9051 }

## 👋 Introduction

This PR resolves out of memory error by using below techniques
1. Select only necessary columns instead of select all 
2. Avoid unnecessary trips to db to query the same info. 
3. Implement preloading technique using "With" directive in graphql 
4. Finally implement batch loader to batch the queries
 
Scope: I only worked on downloading the candidates query. Not the table loading. But certain changes might have benefited how the related tables are loaded/

## 🕵️ Details

 **1 .Select only necessary columns**
User table has tsvector column Searchable which is a huge one and going to continuously grow.
When we define relationship with tables via graphql , even though when we select only certain columns, under the hood it always do "Select * from Users" then return the queried columns to the client. 
This particular query itself has been duplicated every time we load a table that has relationship with Users table. 
There is no way of telling the graphql to `exclude` but only we can tell what to `include` in column list. 
So I opted to define the column list. 
This helped optimize the user table loads. Then I went ahead and did the same thing for other tables when it exceeds more than `3 min` per `select * from table` query. 

**2. Avoid unnecessary trips to db to query the same info**
I tried preloading the relationship info when possible like below : 
This avoided 4 separate `select * from userskills ` for a single candidate.
This set of 4 statements were multiplied for x no. of candidates. 
All those were simplified by the below one .. Kudos to Tristan actually to help with this.  

```
 topTechnicalSkillsRanking: [UserSkill] @with(relation: "userSkills")
  topBehaviouralSkillsRanking: [UserSkill] @with(relation: "userSkills")
  improveTechnicalSkillsRanking: [UserSkill] @with(relation: "userSkills")
  improveBehaviouralSkillsRanking: [UserSkill] @with(relation: "userSkills")
```
**3. When there is a relationship from Experience model to user, there is no need to define the samething for the inherited models. I added this during keyword search and I realize it causes extra select queries issued to DB. 
I removed it and search still works without that.

**4. Batch loader helps to chunk the queries and executes in parellel. Batch size can be customized and in my local batch of 10 itself is higher and 8 seems to work fine. But open to suggestions. In Dev, I didn't find any issues upto 50 per batch.
This helps to go further and download more than 500 candidates like select multiple pages of 500 candidates and download them together. It could also be used to download all.  But since I don't have that much data in Dev currently I tested only upto 1075 candidates. I can do moe if we want to 

## 🧪 Testing

1. Pool candidates CSV download should work without any error on 10, 50 ,100, 500 candidates. 
2. Since its batch, make sure to check the order of candidates on the csv is the same as you see in the table. 
3. Regression Test :Open up different pages like user table , applicant dashboard, search pages to make sure everything works without any error. 
4. Currently the latest build is on dev for testing 500 candidates download

## 📸 Screenshot

dev-1
dev-2
dev-03


apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
```diff
@@ -378,33 +379,69 @@ const PoolCandidatesTable = ({
...
-      });
+        return ids.map(
+          (id) => candidates.find((candidate) => candidate.id === id) ?? null,
+        );
+      } catch (error) {
+        return ids.map(() => null); // Return null for all IDs in case of an error
+      }
+    },
+    {
+      // Configure DataLoader to cache the results for each ID
```
brindasasi, line 417: This will enable cache the result so any subsequent query for the same user doesn't have to be repeated.

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
```diff
@@ -378,33 +379,69 @@ const PoolCandidatesTable = ({
...
-        return poolCandidates;
-      })
-      .catch(() => {
-        toast.error(intl.formatMessage(errorMessages.unknown));
-      });
+        return ids.map(
+          (id) => candidates.find((candidate) => candidate.id === id) ?? null,
+        );
+      } catch (error) {
+        return ids.map(() => null); // Return null for all IDs in case of an error
```
brindasasi, line 413: This will help complete the query in case of any error and download the profiles for successful batches.
I believe this is a better idea instead of giving an empty excel sheet on the error occurrence. 
Idk if I can still loop thru this failure batch and mine that candidates with profile  and return null only for errored candidates. 🤔

api/app/Models/Pool.php
```diff
@@ -105,7 +105,48 @@ public function getActivitylogOptions(): LogOptions
...
+                'position_duration',
+                'accepted_operational_requirements',
+                'gov_employee_type',
+                'priority_weight',
+                'indigenous_declaration_signature',
+                'indigenous_communities',
+                'preferred_language_for_interview',
+                'preferred_language_for_exam',
+                'deleted_at',
+            ]);
```
esizer, line 149: Sucks we can't exclude columns 😢 This is quite a large array that looks to be mirrored in `PoolCandidate`. Would it be worth making this a property of the `User` model so as we modify that model we only have to update in one place? Something like `User::relatedColumns`? Not sure about the name, but the concept could be useful.

petertgiles:
I really like optimization points 1, 2, and 3.  Nice!  :clap: 

I'm worried about the batch loader, though.  With multiple independent queries you open yourself up to a bunch of weird consistency issues that will happen when the data changes between queries that you don't get with a single proper ACID query.

For example, if you have two pages of 10 rows each.  Let's say that you download the first page, and before page two starts the first row from page one is removed from the database.  Row 11 now becomes row 10 but page one was already downloaded so row 11 will be missing from the batch download entirely.

Example 2, also with two pages of 10 rows, the other direction.  Between page one and two a new row 1 is added.  Now row 10 is downloaded twice and the new row 1 is missing from the document.

You can imagine that if we add more bulk operations like bulk expire or bulk qualify the consistency issues could get much greater, too.  I know that this is essentially what the users are doing manually with the tables right now, and I know these queries are what are driving the paginated tables already but I think we're going to cause our users and ourselves pain if we rely on this strategy for data exports.  I could be convinced that this is maybe better than nothing in the short term but then we should remove it as soon as possible later.

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
```diff
@@ -378,33 +379,69 @@ const PoolCandidatesTable = ({
...
+      if (filteredPoolCandidates.length === 0) {
+        toast.error(intl.formatMessage(adminMessages.noRowsSelected));
+      } else {
+        setSelectedCandidates(filteredPoolCandidates);
+      }
+
+      return filteredPoolCandidates;
+    } catch (error) {
+      toast.error(intl.formatMessage(errorMessages.unknown));
+      throw error;
```
esizer, line 440: Is this throw caught anywhere? I'm a little worried this will bubble up to the `ErrorBoundary` and display the error page.

package.json
```diff
@@ -54,5 +54,8 @@
     "jest-fail-on-console": "^3.1.2",
     "prettier": "^3.1.1",
     "turbo": "^1.12.4"
+  },
+  "dependencies": {
+    "dataloader": "^2.2.2"
   }
```
esizer, line 60: Could this be moved to the `apps/web/package.json` file?

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
```diff
@@ -378,33 +379,69 @@ const PoolCandidatesTable = ({
...
-        const poolCandidates: PoolCandidate[] = unpackMaybes(
-          result.data?.poolCandidates,
+  const isPoolCandidate = (
+    candidate: Error | PoolCandidate | null,
+  ): candidate is PoolCandidate =>
+    candidate !== null && !(candidate instanceof Error);
+
+  const batchLoader = new DataLoader<string, PoolCandidate | null>(
+    async (ids) => {
+      const batchSize = 8;
```
esizer, line 389: Is there a reason we settled on 8? This results in many requests that tend to block each other and create a very long waterfall. Is there any danger to increase this number so we can make fewer requests? Especially since we are on HTTP1 which has some trouble with multiple requests blocking.

Screenshot 2024-02-20 104627

esizer (changes requested):
Seems to do well, I just have some suggestions/questions.

 I was hoping we could avoid batching but it seems to be required since removing it still results in running out of memory 😢

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
brindasasi: yeah I explained in PR details. Dev was fine with 50 for safer and even good with 100 records per batch with little memory spike.

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
brindasasi: yeah .. I could remove this. unknown error toast will suffice

brindasasi:
> I really like optimization points 1, 2, and 3. Nice! 👏
> 
> I'm worried about the batch loader, though. With multiple independent queries you open yourself up to a bunch of weird consistency issues that will happen when the data changes between queries that you don't get with a single proper ACID query.
> 
> For example, if you have two pages of 10 rows each. Let's say that you download the first page, and before page two starts the first row from page one is removed from the database. Row 11 now becomes row 10 but page one was already downloaded so row 11 will be missing from the batch download entirely.
> 
> Example 2, also with two pages of 10 rows, the other direction. Between page one and two a new row 1 is added. Now row 10 is downloaded twice and the new row 1 is missing from the document.
> 
> You can imagine that if we add more bulk operations like bulk expire or bulk qualify the consistency issues could get much greater, too. I know that this is essentially what the users are doing manually with the tables right now, and I know these queries are what are driving the paginated tables already but I think we're going to cause our users and ourselves pain if we rely on this strategy for data exports. I could be convinced that this is maybe better than nothing in the short term but then we should remove it as soon as possible later.



> I really like optimization points 1, 2, and 3. Nice! 👏
> 
> I'm worried about the batch loader, though. With multiple independent queries you open yourself up to a bunch of weird consistency issues that will happen when the data changes between queries that you don't get with a single proper ACID query.
> 
> For example, if you have two pages of 10 rows each. Let's say that you download the first page, and before page two starts the first row from page one is removed from the database. Row 11 now becomes row 10 but page one was already downloaded so row 11 will be missing from the batch download entirely.
> 
> Example 2, also with two pages of 10 rows, the other direction. Between page one and two a new row 1 is added. Now row 10 is downloaded twice and the new row 1 is missing from the document.
> 
> You can imagine that if we add more bulk operations like bulk expire or bulk qualify the consistency issues could get much greater, too. I know that this is essentially what the users are doing manually with the tables right now, and I know these queries are what are driving the paginated tables already but I think we're going to cause our users and ourselves pain if we rely on this strategy for data exports. I could be convinced that this is maybe better than nothing in the short term but then we should remove it as soon as possible later.

I get your point. It depends on how frequently the table data would change. 
The download query might go lets say max 30 seconds. If the data change within that there is consistency issue for sure.
But wouldn't it be the same case if we store it in s3 bucket or laravel file storage  somewhere and access it from there?

petertgiles:
> But wouldn't it be the same case if we store it in s3 bucket or laravel file storage somewhere and access it from there?

No, if the entire dataset was retrieved with a single database request it wouldn't matter how long it took or how many concurrent changes occurred while it was executing.  The database engine would guarantee that the results would be exactly correct for the point that the isolated context started.  Now, if it took a long time and there were many changes the result might be quickly out of date but at least it wouldn't have duplicates or missing rows.

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
vd1992: That's strange, I set batch to 100. 
Pulling 221 candidates

Query 1: 2.13 MB in 20 seconds

Query 2: 2.14 MB in 19.7 seconds

Query 3: 0.5 MB in 7.3 seconds

Excel document of 221 candidates size 280 KB successful

Ten shouldn't be causing issues 😕

esizer: I beleive it is server dependant so when run locally it all depends on how much your computer can handle. I'm wondering if this calls for a env var so we can set it higher in prod but allow devs to set it to a reasonable number for local testing? 🤔 

Or is that overkill and we just set it at a prod number and put a comment to tell devs to lower manually if needed?

vd1992: Not sure, it can be handy for deployed too for fidgeting with the number much more quickly. Adjusting both up and down.

brindasasi: Exactly what I was thinking just now Eric!
Env variable would be easier for long run.

But if we think we are not keeping this batching for long run, not worth adding an env variable.

I will test again with 100 on dev. to make sure.

But I'm super impressed on Vachan's machine performance. 🫠

api/app/Models/Pool.php
brindasasi: This is a great suggestion.  Fixed.

api/app/Providers/AuthServiceProvider.php
```diff
@@ -95,7 +95,46 @@ public function resolveUserOrAbort($bearerToken, $tokenService): ?User
 
             // By this point we have verified that the token is legitimate
 
-            $userMatch = User::where('sub', $sub)->withTrashed()->first(); // 3. match "sub" claim to user 'sub' field.
+            $userMatch = User::where('sub', $sub)->withTrashed()->select(
```
esizer, line 98: These look like the same as `User::getSelectableColumns()`? Seems like we could also use that here too.

esizer (changes requested):
One last very minor suggestion then I think this is good to go! 👍

PR #9289, opened 2024-02-08
[Debt] Handle logging for tokens better

### LINKED ISSUE #9147: ♻️ Better handle rejected token requests
## ♻️ Debt/Refactor

We're not currently handling failed token requests very well.  We're ending up with unclear error messages and big stack traces in the logs.

image

That endpoint returns an HTTP 400 when it doesn't like the request.

image

## 🙋‍♀️ Proposed Solution

It looks like the Laravel HTTP client has a nice callback for handling 400 series errors.
https://laravel.com/docs/10.x/http-client#error-handling

Let's catch that and make sure that a clear message is logged (INFO level?).

## ✅ Acceptance Criteria

- [ ] The "invalid_grant" response is caught and a clear, concise message is logged instead of the "HTTP response is 400"
- [ ] Current login process is not impacted

petertgiles:
🤖 Resolves #9147

## 👋 Introduction

This branch turns off exception throwing when requesting a token and instead manually checks for a failed response.  This allows us to do clearer logging.

## 🧪 Testing

1. Configure your environment to use the Sign In Canada test server.
2. Break the code so that the authcallback token request to fail.  This is an easy way to do it:
image
3. Try to log in.

Optional:
Set 
```
APP_ENV=production
APP_DEBUG=false
```
so you can see what the users will see.

## 📸 Screenshot

image
image

### DIFF
--- a/api/app/Http/Controllers/AuthController.php
+++ b/api/app/Http/Controllers/AuthController.php
@@ -6,7 +6,6 @@ use App\Services\OpenIdBearerTokenService;
 use Exception;
 use Illuminate\Http\Client\ConnectionException;
 use Illuminate\Http\Client\PendingRequest;
-use Illuminate\Http\Client\RequestException;
 use Illuminate\Http\Request;
 use Illuminate\Http\Response;
 use Illuminate\Support\Facades\Http;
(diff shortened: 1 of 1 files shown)


vd1992 (approved):
Matched it 

image

image

And login normal when not intentionally breaking it

image

PR #9290, opened 2024-02-08
[Bug] Prevent duplicates in the showcase

### LINKED ISSUE #9194: 🐛 Able to add duplicate skills on Top skills and is mishandling the skill levels already added for same skill
## 🐛 Bug

On the top skills library, I'm able to add same skill twice. When added second time , the skill doesn't give level selection and stored as "N/A" . When you go to Skill showcase, now the prev skill has been replaced by the second skill and the level is lost.

## 🦋 Expected Behaviour

Top skill library can't have same skill more than once. 

## 🕵️ Details

## 📋 Steps to Reproduce

1. Go To Top skill library
2. Select a skill and add it with any level like "Entry Level"
4. Now click add and select the same skill from my library
5. Observe you have duplicate skills 
6. Go to Skill Showcase and come back via Top skills library , observe your prev selection is over written by the newly added one and the level is "undefined"

## 📸 Screenshot

Screenshot 2024-02-01 192100

Screenshot 2024-02-01 192243

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] Top skill library can't have same skill more than once. 
- [ ] To Improve skills can't have the same skill more than once
- [ ] BUT the same skill can appear once in both lists
- [ ] Backend validation (trying to add same skill twice results in errror)
- [ ] Test for backend validation
- [ ] Avoid showing the option in the client to add skill twice

vd1992:
🤖 Resolves #9194

## 👋 Introduction

Prevent duplication from occurring in the skill showcases. Prevention from both front and back. 

## 🧪 Testing

1. Head to the skills showcase
2. Be unable to have duplication
3. Showcase no different otherwise 

## 📸 Screenshot

image

image


esizer (approved):
Nice! Working great for me 👍

PR #9296, opened 2024-02-09
[Fix] User input in team members add dialog

### LINKED ISSUE #9064: 🐛 [Prod bug] Unable to add new member to Team from Team Members page
## 🐛 Bug

On the admin side, if you go to the Team page, go to View team members, and go to Add new member, you cannot add anyone because the dropdown of user names is empty. This is because the query to get all user public profiles returns with a 500 error!

## 🦋 Expected Behaviour

To work around this, we've decided to replace the query with a version which allows for searching by name or email, and which limits the results to 100 users. 

We will not worry about implementing pagination in the combobox. Instead, we will let users shorten the number of results by narrowing the search.

## 🕵️ Details

## 📋 Steps to Reproduce

I was unable to reproduce this exact issue locally, even with 20 000 seeded users. However, it is easy to reproduce in prod:

1. Log in as an admin
2. Go to the Teams table
3. Select a team
4. go to the View team members tab
5. Click "add new member"
6. Open the Manger name dropdown. No names appear.

## 📸 Screenshot

## 🙋‍♀️ Proposed Solution

Make the new filterable query paginated, to match the pattern of other queries and to make this query more flexible. Even if the UI doesn't make use of pagination here, we can just choose a page size of 100 and stick to the first page.

## ✅ Acceptance Criteria

- [ ] Community Managers can add people to teams through the Add new member dropdown
- [ ] The query will only return 100 users at a time, and can be filtered by name and email
- [ ] The combobox will repeat the query the value in the search bar changes, with some reasonable debouncing

esizer:
🤖 Resolves #9064 

## 👋 Introduction

Applies a fix to the users team member dialog crashing by using a paginated query.

## 🕵️ Details

This adds a new `userPublisProfilesPaginated` query that uses a limited search scope (`publicProfileSearch`) that only searches available fields on the public profile.

It also adds an `excludeIds` option so we return the proper page number (`100`) when filtering out existing team members.

## 🧪 Testing

1. Build `npm run dev`
2. Login as admin@test.com
3. Navigate to `/admin/teams`
4. Click on a team name
5. Navigate to the "View team members" tab
6. Active the "Add new member" dialog
7. Focus the "Manager name" field
8. Confirm it opens with only 100 results
9. Search for a user
10. Confirm search returns expected results
11. Confirm saving the form still adds the user

## 📸 Screenshot

Screenshot 2024-02-09 104755

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -713,7 +713,11 @@ type Query {
     orderBy: [OrderByClause!] @orderBy
   ): [UserPublicProfile]!
     @orderBy(column: "created_at", direction: DESC)
-    @paginate(defaultCount: 10, maxCount: 1000, scopes: ["authorizedToView"])
+    @paginate(
+      defaultCount: 10
+      maxCount: 1000
+      scopes: ["authorizedToViewBasicInfo"]
+    )
     @guard
     @can(ability: "viewBasicInfo", model: "User")
   # countApplicants returns the number of candidates matching its filters, and requires no special permissions.
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -614,6 +614,13 @@ input UserFilterInput {
   trashed: Trashed @trashed
 }
 
+input UserPublicProfileFilterInput {
+  telephone: String @scope
+  email: String @scope
+  name: String @scope
+  generalSearch: [String] @scope
+}
+
 # Changes to this input will require some manual updates to api/app/GraphQL/Queries/CountPoolCandidatesByPool.php since it doesn't use the directives for automatic resolution
 input ApplicantFilterInput {
   equity: EquitySelectionsInput @scope
@@ -703,6 +710,14 @@ type Query {
     @deprecated(
       reason: "We should avoid non-paginated queries when we know a query will return thousands of values. In this case, we must write an alternative first!"
     )
+  userPublicProfilesPaginated(
+    where: UserPublicProfileFilterInput
+    orderBy: [OrderByClause!] @orderBy
+  ): [UserPublicProfile]!
+    @orderBy(column: "created_at", direction: DESC)
+    @paginate(defaultCount: 10, maxCount: 1000)
+    @guard
+    @can(ability: "viewBasicInfo", model: "User")
   # countApplicants returns the number of candidates matching its filters, and requires no special permissions.
   countApplicants(where: ApplicantFilterInput): Int!
     @count(model: "User", scopes: ["inITPublishingGroup"])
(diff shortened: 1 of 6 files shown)


api/graphql/schema.graphql
```diff
@@ -703,6 +707,15 @@ type Query {
     @deprecated(
       reason: "We should avoid non-paginated queries when we know a query will return thousands of values. In this case, we must write an alternative first!"
     )
+  userPublicProfilesPaginated(
+    where: UserPublicProfileFilterInput
+    excludeIds: [ID!] @notIn(key: "id")
+    orderBy: [OrderByClause!] @orderBy
+  ): [UserPublicProfile]!
+    @orderBy(column: "created_at", direction: DESC)
+    @paginate(defaultCount: 10, maxCount: 1000, scopes: ["authorizedToView"])
```
vd1992, line 716: Question @tristan-orourke 
Community manager should be able to add people right? Currently they can remove members from a team only, and can only add themselves. I think this is occurring due to the scope being added, but I don't think it should be added as the permission/can and being a limited subset of User should suffice. 

image

esizer: Oh this is a good catch. I was actually wondering about the scope. I applied it because it was applied to the original query. Thanks for asking this!

esizer: @tristan-orourke do you mind taking a look at the rearranging of scopes? Not sure if this was the best way to do it or not 😅 

[fix user basic info query scope](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9296/commits/e650d5fda47399ba9f6edf2ac5b7a3dc0c009464)

tristan-orourke: I also don't know if its the best way or not, but it seems okay 😄

esizer: lol, okay thanks 👍 🤣

apps/web/src/pages/Teams/TeamMembersPage/components/useAvailableUsers.ts
```diff
@@ -1,44 +1,69 @@
...
 
 import { TeamMember } from "~/utils/teamUtils";
 
 const TeamMembers_AvailableUsersQuery = graphql(/* GraphQL */ `
-  query TeamMembers_AvailableUsers {
-    userPublicProfiles {
-      id
-      firstName
-      lastName
-      email
```
vd1992, line 15: Query now unneeded, so can delete it from schema and unused operations

image

vd1992 (changes requested):

apps/web/src/pages/Teams/TeamMembersPage/components/useAvailableUsers.ts
esizer: 🤦‍♀️ Forget we discussed that in standup. Removed in [remove old user public profiles query](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9296/commits/9591274888f921020a5ada2f4beae6cc6847ecce)

vd1992 (approved):
Things look good to me!

PR #9306, opened 2024-02-12
[Debt] New skill seeders for UAT

### LINKED ISSUE #9141: ✨ Update skills on UAT based on Prod data
## ✨ Feature

The skills on UAT are very out of date with Prod. We would like to update them to match again. This means pulling skills and skill families from Prod and using those to update UAT. (probably we should do this every year or so?)

Get live skill data from `/graphiql`. Use that to edit the skills seeder and fakeData file. Run skill seeder on UAT.

## ✅ Acceptance Criteria

- [ ] UAT skills and skill families mostly match what is in production
- [ ] New seeder has been run on UAT

petertgiles:
🤖 Resolves #9141

## 👋 Introduction

This branch updates the skill and skill family seeders to load from external JSON files.  It also includes JSON files to represent a snapshot of the data from prod.  Running these seeders will update any models which match by key but will not remove any models.

Our skill and skill family models are not sorted in our API so I've linked to a VS Code extension for sorting JSON.

## 🧪 Testing

1. `./artisan db:seed --class=SkillFamilySeeder`
2. `./artisan db:seed --class=SkillSeeder`

- [ ] Seeders work
- [ ] Seeders are not destructive
- [ ] Seeders represent prod data

## 📸 Screenshot

image

## :truck: Deployment Instructions

Run the seeders in UAT.

> [!CAUTION]
Do not run the seeders in prod!


petertgiles:
What do devs think?  Would you rather see the data stored in files as PHP arrays like before or is JSON like this better?

esizer:
> What do devs think? Would you rather see the data stored in files as PHP arrays like before or is JSON like this better?

I was going to ask about the possibility of somehow maintaining the JSON and our [storybook "fake" data](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/packages/fake-data/src/skills.json). If that would be possible, JSON would be preferred from me.

esizer (approved):
Nice! 👍 

 - Non-destructive
 - Updates existing
 - Seeded new

PR #9318, opened 2024-02-12
[Feature] Remove locked sorting for RoD

### LINKED ISSUE #9224: ✨ When Record of Decision is on, remove locked sorting on Pool Candidates table
## ✨ Feature

When Record of Decision feature flag is turned on, remove the locked sorting on the Pool Candidates table. The currently locked columns (Status, Category, Final Decision and Job Placement) should be sortable _if_ you click on them.

The only exception is that bookmarked candidates should continue to be locked to appearing at the top.

## 🕵️ Details

The locked columns are an old approach to when tables were used to guide assessment. With RoD, assessment has its own page and the locked sorting is not necessary.

## 🎨 Design File

### 🧑‍🎨 Designer

@Jerryescandon 

## 📸 Screenshot

image

In PoolCandidateTable.tsx
image

image

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] If RoD flag is true, PoolCandidates table no longer forces ordering by status or priority
- [ ] If RoD flag is true, Status, Category, Final Decision and Placement columns are not locked, and can be clicked on to sort
- [ ] If RoD flag is true, bookmarked candidates always still float to the top
- [ ] By default, candidates are sorted by name alphabetic

vd1992:
🤖 Resolves #9224

## 👋 Introduction

If RoD is active, default sort is less multi-layered, can sort by more columns. 
Final default sort is now alphabetical always. 

## 🧪 Testing

RoD OFF

1. Default final sort is alphabetical
2. Otherwise, unchanged

RoD ON

1. Default final sort is alphabetical
2. Priority and status no longer fed in by default
3. Can sort by priority and status columns now

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
+++ b/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
@@ -338,7 +338,12 @@ const PoolCandidatesTable = ({
       ),
       page: paginationState.pageIndex,
       first: paginationState.pageSize,
-      sortingInput: getSortOrder(sortState, filterState, doNotUseBookmark),
+      sortingInput: getSortOrder(
+        sortState,
+        filterState,
+        doNotUseBookmark,
+        recordOfDecision,
+      ),
     },
   });
 
--- a/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
+++ b/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
@@ -481,9 +481,7 @@ const PoolCandidatesTable = ({
             },
           },
         }) => finalDecisionCell(intl, status),
-        meta: {
-          sortingLocked: true,
-        },
+        enableSorting: false,
       },
     ),
     columnHelper.accessor(
@@ -501,9 +499,7 @@ const PoolCandidatesTable = ({
             },
           },
         }) => jobPlacementCell(intl, status),
-        meta: {
-          sortingLocked: true,
-        },
+        enableSorting: false,
       },
     ),
     columnHelper.accessor(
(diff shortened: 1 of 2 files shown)


apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
```diff
@@ -434,7 +439,7 @@ const PoolCandidatesTable = ({
           },
         }) => statusCell(poolCandidate.status, intl),
         meta: {
-          sortingLocked: true,
+          sortingLocked: !recordOfDecision,
           hideMobileHeader: true,
         },
       },
```
vd1992, line 445: I unlocked sorting for status and priority/category columns. I didn't unlock sorting for placement and final decision columns as I thought it would be confusing. Final decision and placement also access the status field, but then in the render step output something based off a function, therefore attempting to sort it wouldn't look sensible. 

Example

image

image

image

esizer: I'm wondering it is better to use [`enableSorting`](https://tanstack.com/table/v8/docs/api/features/sorting#enablesorting) instead. I believe the lock is a visual indication that the table is sorted by that column and cannot be changed which is technically not true.

esizer (commented):
Looking and working great! Glad to not be so confused by how it is sorted anymore 😅 Oddly, enough, it seems like we removed the logic to lock them earlier and just never removed the icons? 🤔 

Anyway, I just had one small question before approving.

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
vd1992: Sure, changed the other two columns to `enableSorting = false` for improved clarity in https://github.com/GCTC-NTGC/gc-digital-talent/pull/9318/commits/4541d41fa64882ae8b5c8fde164dc256c07d3d92

esizer (approved):
Nice! 👏

PR #9320, opened 2024-02-13
[Feature] Add process column to all candidates table

### LINKED ISSUE #9295: ✨ [User Requested Feature] Add new column "Process" (pool name)  in the ALL candidates table
## ✨ Feature
All candidates table consists candidates from all pools and it's hard to find which pool they belong to. 
So user always prefer going to pool and open up the table from there. Adding the pool name in here , will make this table more useful. 
## 🕵️ Details

Add new column "Process" ( to be consistent with the naming) to identify which pool each candidate belongs to. 

### 🧑‍🎨 Designer

Ask @Jerryescandon 

## 📸 Screenshot

image

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Add new column "Process" ( to be consistent with the naming) to identify which pool each candidate belongs to. 
- [ ] Hidden by default on the pool-specific page, visible by default on Candidate Search page.
- [ ] Value in column is a link to the pool page
- [ ] Use the full-template name (whatever format it is)
- [ ] Aria label `View process: <full title>`

esizer:
🤖 Resolves #9295 

## 👋 Introduction

Adds a "Process" column to the all candidates table, omitting it from specific pool candidate tables.

## 🧪 Testing

1. Build `npm run dev`
2. Login as admin
4. Confirm the process column exists and links to the proper location
5. Navigate to a specific pools "Talent placement" tab
6. Confirm the process column does not appear

## 📸 Screenshot

Screenshot 2024-02-13 123651


apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
```diff
@@ -419,6 +432,25 @@ const PoolCandidatesTable = ({
...
     ),
+    ...(currentPool
+      ? []
+      : [
+          columnHelper.accessor(
+            ({ poolCandidate: { pool } }) => getFullPoolTitleLabel(intl, pool),
+            {
+              id: "process",
+              header: intl.formatMessage(processMessages.process),
+              sortingFn: normalizedText,
```
vd1992, line 443: Pool candidates table is API driven, so the search and sort has the external factor. May need to be added into this function so sorting by process works. May also want to merge main first to incorporate my recent changes to this sort

image

vd1992 (changes requested):
One thing I got

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
esizer: This is proving to be incredibly tricky since it is a property of a `jsonb` column on a related table. I'm going to keep playing around to see if I can get it but we may need to spin sorting off into a separate ticket 😬

vd1992: Ah shoot, I forgot that it would be a bit trickier than just that. Thought it was just a string.
I think for now just disable the sort and leave it at that?

esizer: Okay, after some help from @tristan-orourke I settled on just creating a new scope for this query to work around the issues with ordering by related jsonb properties. This was done in two commits. I tried adding some documentation so we know why we did this weird thing 😅 Let me know if it does not make sense!

 - [add custom orderby scope for json relationship aggregates](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9320/commits/4e244574dcde4994f2701ee53515154a51d8b6d5)
 - [implement custom order by for pool name](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9320/commits/88aea5a7a6b4a278e6405b749cd11cdbe25f93e1)

esizer: And test added in [add test for new pool name sort scope](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9320/commits/39d26681dcb8017eeb0a8c684dd9b3ba50e072bb)

api/app/Models/PoolCandidate.php
```diff
@@ -761,6 +761,25 @@ private function addSkillCountSelect(Builder $query, ?array $skills): Builder
...
+    /**
+     * Custom sort to handle issues with how laravel aliases
+     * aggregate selects and orderBys for json fields in `lighthouse-php`
+     *
+     * The column used in the orderBy is `table_aggregate_column->property`
+     * But is actually aliased to snake case `table_aggregate_columnproperty`
+     */
+    public function scopeOrderByPoolName(Builder $query, ?array $args): Builder
+    {
+        extract($args);
```
vd1992, line 773: Not a logic concern but a naming one, scopes are used to refer to filtering right? In this case, it isn't really a scope or doing any scoping in the Laravel sense of the word. Not a big deal but just thinking about it

esizer: Scopes are just for query constraints in general. Imo, while it is a little odd to do ordering in a scope, it can be considered a constraint and be applicable.

vd1992 (approved):
Sorts well to me! Looks great!
Exercising it and nothing looked off. 

Sorting by French works as it should

image

It multi-layer sorts fine actually from what I can tell, here is the requests candidates page.

image

image

Same with the candidates table with RoD OFF, that is multilayered too

image

PR #9334, opened 2024-02-15
[Feature] Add tentatively placed status

### LINKED ISSUE #9076: ✨ New status for candidates again - "Tentatively placed (verbal or email confirmation)"
## ✨ Feature

New candidate status to help us with metrics on hiring. "Tentatively placed (verbal or email confirmation)"

## 🕵️ Details

In dropdowns on the admin side, use this copy:
- EN: Offer in progress
- FR: Offre en cours

The candidate should still show up in searches, ~so nothing special about this new status~ which requires it to be added to some functions in backend scopes.

## Proposed Implementation
 
Call the enum PLACED_TENTATIVE.

## ✅ Acceptance Criteria

- [ ] New status available for pool operators and request responders to set for candidates
- [ ] Applicants looking at their application status don't see this - the card still shows `Ready to hire` chip - same as QUALIFIED_AVAILABLE
- [ ] Tentatively Placed candidates DO still appear in searches. That means that any scopes in User.php or PoolCandidate.php which currently check for QUALIFIED_AVAILABLE and PLACED_CASUAL should now also allow PLACED_TENTATIVE
- [ ] Test coverage for ensuring PLACED_TENTATIVE candidates show up in searches

vd1992:
🤖 Resolves #9076

## 👋 Introduction

Adds the new status, copy called "Offer in Progress". Functions basically identically to qualified available. 

## 🕵️ Details

Migration to regenerate the weighing. Then finding all the places around the app to slot it in.
One test I found odd when adding testing so I basically replaced it

## 🧪 Testing

1. Candidate with new status rendered appropriately admin-side, like in tables
2. Appears in searches
3. Applicant side view sees "Ready to hire"

## 📸 Screenshot

Candidate with the new status and appropriate rendering. 

image

Can search for that candidate 

image

Applicant's perspective

image


petertgiles (approved):
Wow, this one was a bit tricky but it looks great!

packages/i18n/src/messages/localizedConstants.tsx
```diff
@@ -660,6 +660,12 @@ const poolCandidateStatuses = defineMessages({
     id: "yRxy3a",
     description: "The pool candidate's status is Qualified Withdrew.",
   },
+  [PoolCandidateStatus.PlacedTentative]: {
+    defaultMessage: "Offer in Progress",
```
mnigh, line 664: Should be sentence case (REF: https://github.com/GCTC-NTGC/gc-digital-talent/issues/9076#:~:text=EN%3A-,Offer%20in%20progress,-FR%3A%20Offre%20en)

French ID will also need to change based on different hash.
```suggestion
    defaultMessage: "Offer in progress",
```

mnigh (changes requested):

api/database/migrations/2024_02_13_234848_tentatively_placed_status_addition.php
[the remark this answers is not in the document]
vd1992: Sure https://github.com/GCTC-NTGC/gc-digital-talent/pull/9334/commits/f1d84eac76db491c6e13d67ae39862f57a54ee74

packages/i18n/src/messages/localizedConstants.tsx
vd1992: I'm going to push back, I was intentional with it. All the other statuses aren't sentence cased. 

image

But the French ones, except one, are it seems. Going to need an issue to decide what we want 

image

esizer: I think we should at least create a ticket to fix these. I'm pretty sure GC guidelines are that we always use sentence case rather than title case.

vd1992: Issue for reference https://github.com/GCTC-NTGC/gc-digital-talent/issues/9385

api/app/Models/PoolCandidate.php
```diff
@@ -222,7 +222,11 @@ public static function scopeQualifiedStreams(Builder $query, ?array $streams): B
         $query->where(function ($query) {
             $query->whereDate('pool_candidates.expiry_date', '>=', Carbon::now())->orWhereNull('expiry_date'); // Where the PoolCandidate is not expired
         })
-            ->whereIn('pool_candidates.pool_candidate_status', [PoolCandidateStatus::QUALIFIED_AVAILABLE->name, PoolCandidateStatus::PLACED_CASUAL->name]) // Where the PoolCandidate is accepted into the pool and not already placed.
+            ->whereIn('pool_candidates.pool_candidate_status', [
```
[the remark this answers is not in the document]
petertgiles: Nice!

petertgiles (approved):
Thanks for the tweaks!

PR #9399, opened 2024-02-19
[Debt] Add localized logout redirects

### LINKED ISSUE #9299: ♻️ Eliminate language redirect on logout
## ♻️ Debt/Refactor

Right now our logout process has a lot of redirects.  One of them is that when returning from ending the session the user is sent to /logged-out and then immediately has to do the language redirect again.  This causes the app to have to boot twice after ending the session.  If we sent the user back to the localized logged-out page then we could eliminate one of these redirects and reboots.

image

## 🙋‍♀️ Proposed Solution

1) Register the localized endpoints with Sign In Canada as valid redirects.
2) Send the localized endpoints as a post_logout_redirect when ending the session.

## ✅ Acceptance Criteria

- [ ] The user does not experience an unnecessary language redirect when returning from endsession

petertgiles:
🤖 Resolves #9299 

## 👋 Introduction

This branch improves the speed of the logout process by ensuring that after logout, the user is redirected back to the app in their preferred language and does not need to experience an unnecessary language redirect and app reboot.

## 🧪 Testing

1. Configure the app to use Sign In Canada (Test)
2. Log in
3. Log out
4. Confirm that you were sent back from SIC with a localized URL.

## 🚚 Deployment

1) Remove the OAUTH_POST_LOGOUT_REDIRECT variable from IaC
2) Add OAUTH_POST_LOGOUT_REDIRECT_EN and OAUTH_POST_LOGOUT_REDIRECT_FR variables with the correct "/en" and "/fr" prefixes. (Already done in IaC Release-511.)


brindasasi (approved):
Verified in Dev. Works just fine! 🥳

PR #9405, opened 2024-02-19
[Fix] Table landmark uniqueness

### LINKED ISSUE #9379: ⌨ Tables lack a label
## ⌨ Accessibility Issue

It looks like our tables have the `aria-labelledby` applied to the wrong `div` 😢 

### WCAG 2.1 Criteria

- https://dequeuniversity.com/rules/axe/4.8/landmark-unique?application=AxeChrome
- https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/region.html

## 🦋 Expected Behaviour

Table regions have a name in the accessibility tree and are unique.

## 🕵️ Details

Looks like we are applying the `aria-labelledby` to the main wrapping `div` but the `role="region"` is applied to a child `div`.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/a66841ef459952763f3b409238d2b2329f42cd00/apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx#L330-L331

https://github.com/GCTC-NTGC/gc-digital-talent/blob/a66841ef459952763f3b409238d2b2329f42cd00/apps/web/src/components/Table/ResponsiveTable/Table.tsx#L20-L21

## 📋 Steps to Reproduce

1. Navigate to the skill library
2. Open the accessibility tree
3. Inspect the region wrapping the tables
4. Observe that there is no name

## 📸 Screenshot

Screenshot 2024-02-15 131732

## 🙋‍♀️ Proposed Solution

I beleive we should be able to just move the `aria-labelledby` and `role="region"` so they are on the same element.

## ✅ Acceptance Criteria

- [ ] Table regions have programmatically discernible name
- [ ] Table region names are unique

esizer:
🤖 Resolves #9379 

## 👋 Introduction

Updates the the base table component to ensure landmarks are unique. 

## 🕵️ Details

We had an issue where we had a `div` with `role="region"` but applied the `aria-labelledby` attribute to the parent node instead. As well, the pagination had a hardcoded "Page navigation" label, we have prefixed it with the caption to make these unique.

## 🧪 Testing

1. Build `npm run dev`
2. Login as an applicant
3. Navigate to the skill library
4. Inspect the table accessibility tree
5. Confirm all landmarks (region, table, nav) have unique names across the tables

### DIFF
--- a/apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
+++ b/apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
@@ -327,8 +327,11 @@ const ResponsiveTable = <TData extends object, TFilters = object>({
       {hasNoData || hasNoVisibleRows ? (
         <NullMessage {...(nullStateMessage ? { ...nullStateMessage } : {})} />
       ) : (
-        <div aria-labelledby={captionId}>
-          <Table.Wrapper data-h2-position="base(relative)">
+        <>
+          <Table.Wrapper
+            data-h2-position="base(relative)"
+            aria-labelledby={captionId}
+          >
             <Table.Table>
               <Table.Caption id={captionId}>{caption}</Table.Caption>
               <Table.Head>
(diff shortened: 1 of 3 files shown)


petertgiles (approved):
Nice.  :+1:

Issue #9035, opened 2024-01-17
♻️ Skill Showcase list should use new non-form Card Repeater


tristan-orourke:
## ♻️ Debt/Refactor

Convert the Skill Showcase ui to use the new card repeater created in #8161

## 🕵️ Details

Ask @substrae if you have design questions

## 🙋‍♀️ Proposed Solution

(optional) Do you have a proposed solution?

## 🌎 Localization

(optional) Provide any new copy along with translations available.

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the debt was addressed and expected functionality has not been affected.

- [ ] Skill Showcase repeaters use new Card Repeater from #8161

PR #9424, opened 2024-02-21
[Debt] Refactor card repeater remove event

petertgiles:
🤖 Related to (but doesn't close) #9035

## 👋 Introduction

This branch adds the ability to use a dialog to remove an item from the repeater.  It removes the `onRemove` event from the card and instead adds a `remove` prop to inject a button, like `edit`.  It also moves the announcer into the repeater provider to make sure the announcements don't get missed.

## 🧪 Testing

1. Check the card repeater stories in Storybook
2. Check the card repeater in the assessment plan builder

- [ ] Button looks good, is accessible
- [ ] Button can be disabled
- [ ] Button fires event properly
- [ ] Dialog in story can cancel the event

## 📸 Screenshot

image


esizer (approved):
Works great! 👏

PR #9426, opened 2024-02-21
[Fix] General question remove

### LINKED ISSUE #9425: 🐛 General question repeater remove not working for `0` index
## 🐛 Bug

When attempting to delete the first general question on a pool, it does nothing.

## 🦋 Expected Behaviour

You can delete the first general question on a pool.

## 🕵️ Details

Just some poor conditional logic on my part. Since `0` is falsey, we should be checking for undefined or less than 0.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/99775d84875a2ebace531df2bd3f6aa535925b4a/apps/web/src/pages/Pools/EditPoolPage/components/GeneralQuestionsSection/GeneralQuestionDialog.tsx#L43-L44

## 📋 Steps to Reproduce

1. Navigate to an edit pool page
2. Make sure at least one general question exists
3. Attempt to delete
4. Confirm it does not trigger a request and remains in the list

## 🙋‍♀️ Proposed Solution

Use `if (typeof index !== "undefined") remove(index);` instead.

## ✅ Acceptance Criteria

- [ ] You can remove any item in general questions field

esizer:
🤖 Resolves #9425 

## 👋 Introduction

Fixes an issue when removing the first question in general questions section of edit pool page.

## 🧪 Testing

1. Build `npm run build`
2. Navigate to an edit pool page
3. Ensure there is at least one general question
4. Attempt to delete the first item
5. Confirm it gets deleted properly

### DIFF
--- a/apps/web/src/pages/Pools/EditPoolPage/components/GeneralQuestionsSection/GeneralQuestionDialog.tsx
+++ b/apps/web/src/pages/Pools/EditPoolPage/components/GeneralQuestionsSection/GeneralQuestionDialog.tsx
@@ -41,7 +41,7 @@ const GeneralQuestionDialog = ({
     const isDelete = values.action === "delete";
     const isAdd = values.id === "new";
     if (isDelete) {
-      if (index) remove(index);
+      if (typeof index !== "undefined") remove(index);
     } else if (values.id) {
       const newQuestion = {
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
Good fix.  :wrench:

PR #9427, opened 2024-02-21
[Debt] Reduce auth error redirects

### LINKED ISSUE #9300: ♻️ Eliminate language redirect on auth error
## ♻️ Debt/Refactor

Right now our auth error process has a lot of redirects.  One of them is that when detecting an auth error the user is sent to /logged-out and then immediately has to do the language redirect again.  This causes the app to have to boot two or three times after an error.  If we sent the user back to the localized logged-out page then we could eliminate one of these redirects and reboots.

image
image

## 🙋‍♀️ Proposed Solution

2) Send the user to a localized logged-out page when ending the session.

## ✅ Acceptance Criteria

- [ ] The user does not experience an unnecessary language redirect when an auth error happens

esizer:
🤖 Resolves #9300 

## 👋 Introduction

Adds the locale to our post logout redirect on auth errors to reduce the number of required redirects.

## 🧪 Testing

1. Build `npm run dev`
2. Login as any user
3. Open inspector and edit the access token to cause it to be malformed
4. Refresh the page
5. Confirm the logout redirects to the localized version of the logged out page

### Current

Screenshot 2024-02-21 114605

### Updated

Screenshot 2024-02-21 114333

### DIFF
--- a/packages/client/src/components/ClientProvider/ClientProvider.tsx
+++ b/packages/client/src/components/ClientProvider/ClientProvider.tsx
@@ -19,6 +19,7 @@ import {
 import { useLogger } from "@gc-digital-talent/logger";
 import { toast } from "@gc-digital-talent/toast";
 import { uniqueItems } from "@gc-digital-talent/helpers";
+import { getLocale } from "@gc-digital-talent/i18n";
 
 import {
   buildValidationErrorMessageNode,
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
Looks good!  FYI, you'll need to add locale to the useMemo dependencies, and I tested that way.  I was worried that you would lose your graphql client state on locale change but it seems to work fine.  I guess the whole app reboots on locale change anyway.

PR #9476, opened 2024-02-22
[Debt] Migrate to pnpm

### LINKED ISSUE #5338: ♻️ Switch from NPM to PNPM
## ♻️ Debt/Refactor

Right now our main motivation for this is that it would avoid the problem of missing integrity hashes. #2557 / #2558

## 🕵️ Details

It shouldn't be that hard to switch, but there may be some nuances. It makes sense to wait until we're done modifying how we use workspaces (#5065) to do this.

## ✅ Acceptance Criteria

- [ ] Ensure dependabot continues to work
- [ ] Ensure turbo continues to work
- [ ] PNPM used in maintenance scripts
- [ ] PNPM used in Github Actions
- [ ] PNPM used in Azure build
- [ ] PNPM detects missing integrity hashes
- [ ] PNPM detects malformed dependencies
- [ ] PNPM handles workspaces properly
- [ ] Help your teammates update their local environments

## 🛑 Blockers

```[tasklist]
### Blocked By
- [ ] #9702
```

esizer:
🤖 Resolves #5338 

> [!NOTE]
> I don't think this is high priority and still not sure if we want to merge at all but I'm putting this up as a draft to let people mull it over.

## 👋 Introduction

Migrates from `npm` to `pnpm`

## 🕵️ Details

I have been slowly working on this over the past few months every Thursday as dependencies update. Thanks to @mnigh we found out the reason we couldn't update was because we had an old version of `@urql/core` sitting in our lock file but not `package.json` so it was no visible. Matt got it updated (since we could after the auth exchange update).

### TypeScript

There is some weird stuff going on with Typescript. I have no idea why the package manager surfaces these issues but that is why there are some changes to the actual code and not just the package manager and dependencies.

### Missing packages

`npm` didn't really care if we actually required a dependency in a package as long as it was installed somewhere. `pnpm` is more strict about this so you will see there are some additions to `package.json` files. These dependencies were in use but not defined.

## 🧪 Testing

1. Install `pnpm` in [your preferred way](https://pnpm.io/installation)
2. Run a bunch of different commands
3. Confirm they succeed


packages/tsconfig/base.json
```diff
@@ -2,20 +2,21 @@
   "$schema": "https://json.schemastore.org/tsconfig",
   "compilerOptions": {
     "target": "es5",
-    "module": "esnext",
-    "lib": ["dom", "esnext", "esnext.intl", "es2017.intl", "es2018.intl"],
+    "module": "ES2020",
```
petertgiles, line 5: This seems like a pretty big change.  What's up?

package.json
```diff
@@ -33,29 +34,29 @@
     "dev:fresh": "h2-build && turbo run dev --filter='./apps/*' --filter='!@gc-digital-talent/e2e' --force",
     "dev:project": "turbo run dev",
     "lint": "turbo run lint",
-    "storybook:project": "npm run storybook",
-    "storybook": "npm run storybook --workspace=@gc-digital-talent/web",
-    "storybook:design-system": "npm run storybook:design-system --workspace=@gc-digital-talent/web",
-    "storybook:web": "npm run storybook:web --workspace=@gc-digital-talent/web",
-    "build-storybook": "npm run build-storybook --workspace=@gc-digital-talent/web",
+    "storybook:project": "pnpm run storybook",
```
petertgiles, line 37: Should this be just `pnpm storybook`?  Same with `e2e:open`?

package.json
petertgiles, line 1: I tried launching the maintenance container right after setup and running install again but it crashed out.
image
I haven't been able to repeat it though.

pnpm-workspace.yaml
petertgiles, line 1: Are you expecting some node_modules directories to be recreated on setup?
image

pnpm-lock.yaml
petertgiles, line 1: If I immediately run `pnpm i` in the maintenance container after a fresh setup it downloads all the packages again.  Why isn't it detecting that everything is already there?  If I run it a second time it doesn't download anything, as expected.

petertgiles (changes requested):
Very cool!  -48,000 lines :astonished: 

A few rough edges to check on, noted.

maintenance/scripts/setup.sh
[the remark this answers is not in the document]
esizer: lol oof, that was a bad merge conflict fix 😅

packages/tsconfig/base.json
esizer: I can change it back to `esnext` but that is a floating value so I locked it down to make troubleshooting a bit easier. If we do always want access to the latest specs I can change it back.

This honestly only got committed by mistake but I left it since everything was working with it set and the next module was making things harder to pin down.

maintenance/scripts/setup.sh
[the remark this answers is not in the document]
esizer: This is actually a bug in hydrogen that npm does not care about but pnpm does. Even though there is [no `bin/clean.js`](https://github.com/hydrogen-css/hydrogen/tree/main/bin) file, the [package.json still references it](https://github.com/hydrogen-css/hydrogen/blob/cff116ef9f05206fd01a942303248eefa948f621/package.json#L30).

@substrae want me to submit an issue? 😅

package.json
esizer: okay, this is actually a very good question. the second arg is assumed to be a script unless another built in command is added.

I did this to make things more simple but honestly, I already don't really like it and think we could explicitly add `run`.

So, we should make it consistent but I'm not sure what pattern to use. Any opinions?

pnpm-workspace.yaml
esizer: Was this related to the message we get on setup?

```sh
The modules directory at "/var/www/html/packages/{packageName}/node_modules" will be removed and reinstalled from scratch. Proceed? (Y/n)
```

packages/tsconfig/base.json
petertgiles: Oh, OK.  Yeah, I'm all for pinning down floating values.  Let's do that.

package.json
petertgiles: I flagged it because you seemed to be going with the other convention.  I'm guessing most of the team is unfamiliar with pnpm and we're using it because it's "drop in" so I'd recommend going with the familiar `pnpm run <script>` form.

pnpm-workspace.yaml
petertgiles: I was running `git clean -xdf ` to clean the workspace before running setup and I noticed those directories in the list of deletions.  I guess because I started off with a clean environment I never got those warnings.

esizer: okay, yeah. That is expected then. Those are just symlinks to the root `node_modules` to enforce only importing those specified in the `package.json` files.

`npm` only does this only for _unique_ packages in workspace but `pnpm` does this for _all_ packages so it can be more strict about importing and versions.

maintenance/scripts/setup.sh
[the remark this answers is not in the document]
esizer: Fixed in [remove duplicate key for swc](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9476/commits/d6d74f1e9636a11b1e686d1b8a87147da2410178)

packages/client/src/components/ClientProvider/ClientProvider.tsx
```diff
@@ -151,6 +151,8 @@ const ClientProvider = ({
 export default ClientProvider;
 
 // https://stackoverflow.com/questions/54116070/how-can-i-unit-test-non-exported-functions
-export const exportedForTesting = {
```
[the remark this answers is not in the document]
esizer: Nice! looks like it isn't needed after all 😄 Removed in [remove unnecessary explicit type](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9476/commits/13127f36804b3fae50e260c69212f370cb85f3b8)

package.json
esizer: I agree, updated in [update pnpm calls to avoid assumed run](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9476/commits/19ffbbc79635e7c4bfd24cf3db11f7c7558ef56e)

packages/date-helpers/package.json
```diff
@@ -8,26 +8,30 @@
     "dist/**"
   ],
   "scripts": {
-    "build": "tsc --declaration",
+    "build": "tsup src/index.ts --silent --dts --external react --minify",
     "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf dist",
-    "dev": "tsc --declaration",
-    "lint": "eslint \"src/**/*.ts*\"",
-    "test": "jest --detectOpenHandles --forceExit --passWithNoTests"
```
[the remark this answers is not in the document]
esizer: Good catch, this was a copy past error 🤦‍♀️ 

Fixed in [fix dev commands](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9476/commits/df2eb44ccab2bc86525a1c173a246b29d8a3d69f)

package.json
esizer: Yeah, I have been trying to reproduce and have been unable to 😢

maintenance/scripts/setup.sh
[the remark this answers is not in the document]
mnigh: I think you might have removed the wrong one as `swc/core` is now at 1.4.5: https://github.com/GCTC-NTGC/gc-digital-talent/pull/9668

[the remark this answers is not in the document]
esizer: Good catch! We just bumped that on Thursday and i forgot to merge main and just used the existing (old) version number 😓 

Fixed in [fix swc version mismatch](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9476/commits/45079cabde82029b9db198059cff04ce6c6b5a1e)

pnpm-lock.yaml
petertgiles: Hmm, the setup would run in a temporary maintenance container and then I'm rerunning `pnpm i` in a different maintenance container.  Perhaps the binaries are being saved in a temporary location that's not in a volume?
image

package.json
petertgiles: Looks like you're still missing some `run`s in the root package.json?

packages/date-helpers/package.json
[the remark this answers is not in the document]
petertgiles: Looks like dev is still missing in packages/fake-data/package.json

package.json
esizer: Hmm, I can't seem to find any commands that are missing it. Perhaps the [order of options](https://pnpm.io/cli/run#options) is confusing? 😅 If that isn't it do you mind helping me find the ones that are missing?

packages/date-helpers/package.json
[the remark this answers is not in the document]
esizer: oops! Added in [add dev command fake-data](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9476/commits/f1b057e1c37829cb4b0ec0270fb9e871aacabb1c)

package.json
petertgiles: Oops, sorry, you're right.  I missed the run command after the filter options.  :confounded:

pnpm-lock.yaml
esizer: I may have figured it out. When you install a package with `pnpm` actually stores your modules in a common location then just hard links to that location. So I'm guessing the hard link? If you jump back and forth between the maintenance container and local I'm pretty sure it is recreating the hard links and re-downloading binaries? 🤔 

Screenshot 2024-03-12 095837

maintenance/scripts/setup.sh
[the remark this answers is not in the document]
substrae: @esizer Yes please!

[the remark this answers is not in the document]
esizer: Done: https://github.com/hydrogen-css/hydrogen/issues/280

petertgiles (approved):
Deploy worked great!

mnigh (commented):
I followed the testing steps in the PR description but unfortunately, it errors out. Any suggestions?

## Screenshot

Screen Shot 2024-03-15 at 11 02 18

petertgiles:
> I followed the testing steps in the PR description but unfortunately, it errors out. Any suggestions?

Did it actually error out?  I noted the warnings about the missing hydrogen script but it looks like it installed successfully to me.

mnigh:
> > I followed the testing steps in the PR description but unfortunately, it errors out. Any suggestions?
> 
> Did it actually error out? I noted the warnings about the missing hydrogen script but it looks like it installed successfully to me.

Uh oh, it seems I interpreted 🟡 as error which is an error by me! 😆

I will run some scripts as instructed in the test steps and make sure they are working.

package.json
mnigh, line 1: Should `check-integrity` be removed and anything related to `package-lock.json` as well since it no longer exists?

## Screenshot
Screen Shot 2024-03-15 at 12 12 04

package.json
```diff
@@ -33,29 +34,29 @@
...
-    "e2e:playwright:chromium": "npm run e2e:playwright:chromium --workspace=@gc-digital-talent/playwright",
-    "e2e:playwright:firefox": "npm run e2e:playwright:firefox --workspace=@gc-digital-talent/playwright",
-    "e2e:playwright:webkit": "npm run e2e:playwright:webkit --workspace=@gc-digital-talent/playwright",
-    "e2e:playwright:report": "npm run e2e:playwright:report --workspace=@gc-digital-talent/playwright",
-    "e2e:open": "npm run e2e:open",
-    "e2e:run:all": "npm run e2e:run:all --workspace=@gc-digital-talent/e2e",
-    "e2e:run:inspect": "npm run e2e:run:inspect --workspace=@gc-digital-talent/e2e",
+    "e2e": "pnpm --filter @gc-digital-talent/e2e run e2e",
+    "e2e:playwright": "pnpm --filter @gc-digital-talent/playwright run e2e:playwright",
+    "e2e:playwright:chromium": "pnpm --filter @gc-digital-talent/playwright run e2e:playwright:chromium ",
```
mnigh, line 47:
```suggestion
    "e2e:playwright:chromium": "pnpm --filter @gc-digital-talent/playwright run e2e:playwright:chromium",
```

package.json
```diff
@@ -5,14 +5,15 @@
   "license": "AGPL-3.0",
   "engines": {
     "node": "20.11.0",
-    "npm": "9.9.2"
+    "pnpm": ">=8.0.0"
   },
-  "packageManager": "npm@9.9.2",
+  "packageManager": "pnpm@8.15.0",
```
mnigh, line 10: How do make sure that `pnpm` itself gets updated if there are issues with this version if it's pinned here?

mnigh (commented):
I really like the resolutions for binaries with `pnpm`: it just works! A few more questions, hopefully not as inane as the one about [errors](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9476#issuecomment-1999969904) that I made. 😄

Should all of these references be changed from `npm` to `pnpm`?

## Screenshots

Screen Shot 2024-03-15 at 12 05 41
Screen Shot 2024-03-15 at 12 05 23

package.json
esizer: That is a good question. We typically gravitate towards pinning versions but do we want to pin this at the patch or minor?

package.json
esizer: Hmm, I'm not sure about `checkIntegrity` thing. We technically don't need that anymore since `pnpm` should be handling that for us. What do you think @petertgiles ?

Also, I can't seem to find the following references to `package-lock`:

 - `docker-compose.yml`
 - `chomatic.yml`
 - `jest.yml`
 - `lint.yml`

esizer:
> Should all of these references be changed from `npm` to `pnpm`?

Good catch! I updated most references in [update references to npm to pnpm](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9476/commits/09876bee983606d7a8fb25dd10b32ab1a894e0e9). Some were intentionally not changed because we need it to be used there. ie:

- `pnpm-lock.yaml`
- `dependabot.yml`
- `deploy.sh`
- `Dockerfile`

package.json
petertgiles: > Hmm, I'm not sure about checkIntegrity thing. We technically don't need that anymore since pnpm should be handling that for us.

If it's really not needed it would be nice to remove it.  Out of scope for this PR, though, I think.

mnigh: Issue created https://github.com/GCTC-NTGC/gc-digital-talent/issues/9771.

package.json
petertgiles: Do you have a sense of how often breaking changes happen?  With npm, even major version changes were largely compatible with each other.  Pin minor is probably fine, right?

esizer: okay, updated to only care about minor version. [pin pnpm to minor version](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9476/commits/55dcd3b67bb133b255fbfacde4d31f7b69d60fc6)

PR #9480, opened 2024-02-22
[Fix] User information e2e test

### LINKED ISSUE #9366: 🐛 `user-info` test intermittently failing
## 🐛 Bug

The "User information" page on admin has a subtitle for the username. In the test, we routinely check for this to determine if thee current role can access that page. Sometimes, the assertion times out.

## 🦋 Expected Behaviour

The assertion does not timeout when the text exists.

## 🕵️ Details

The following location is the assertion that times out.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/475a0652fae649493b438398bace9fabab84cc51/apps/e2e/cypress/e2e/admin/user-info.cy.ts#L119-L121

https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/7910133858/job/21592236559#step:8:6672

## 📋 Steps to Reproduce

1. Build `npm run build`
2. Open cypress `npm run e2e`
3. Run the `user-info` spec many times
4. It may fail? Seems to happen only in GH actions

## 📸 Screenshot

User Information Page -- tests the behavior of admin user information page with different user roles (failed)

## 🙋‍♀️ Proposed Solution

Maybe add a wait on the slowest graphql request to give it time to think? 🤔 

## ✅ Acceptance Criteria

- [ ] Test still makes the same assertion
- [ ] Consistently passes locally and in GH actions

esizer:
🤖 Resolves #9366 

## 👋 Introduction

Fixes an issue where the `user-info` test could sometimes fail.

## 🕵️ Details

This also migrates the test to playwright. The items that were done to reduce flakiness may have gotten lost in the migration so I'll do my best to list them here.

1. Separated spec out into different test cases to better identify when one fails
2. Added a wait for the queries to run before we check for data on the page populated via the API to prevent the assertion timing out before the render

### Waiting

https://github.com/GCTC-NTGC/gc-digital-talent/blob/2cdabe16ea670f729971e3d74ba9a11b817b0ecf/apps/playwright/tests/admin/user-info.spec.ts#L20-L28

## 🧪 Testing

2. Run `npm run e2e:playwright -- -- user-info.spec.ts --project=chromium` a few times
3. Confirm no failures


mnigh (approved):
👍

PR #9481, opened 2024-02-22
[Tooling] Dependabot groups

### LINKED ISSUE #9474: ✨ Dependabot merge groups
## ✨ Feature

It would be amazing if we could implement merge groups for dependabot.

## 🕵️ Details

A large reason dependabot merging takes so long is the fact that a lot of our dependencies are grouped so all have version updates at the same time (even when only one package has changed). Merging them separately causes merge conflicts meaning they need to be rebased and re-run our CI pipeline. This also has a domino effect significantly slowing the entire process down.

Dependabot has released [merge groups](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups) allowing us to merge dependency bumps to similar packages together. I think it would be a great help to implement this feature.

> [!NOTE]
> This may have some growing pains while we get the implementation correct 😢 

The largest contributor to the issue is storybook so we could start with that and expand on it once we are happy with it. But I have identified some other candidates we could also do if we want to dive right in 🏊 

- `@storybook/*`
- `@graphql-codegen/*`
- `@tiptap/*`
- `@urql/*`
- `@types/*` (maybe?) 🤷‍♀️ 
- `@formatjs/*`
- More?

## ✅ Acceptance Criteria

- [ ] Merge groups have been created
- [ ] Confirmed they work

vd1992:
🤖 Resolves #9474 

## 👋 Introduction

Add merge groups.
Not sure how to confirm it works? Other than merge and wait until there are enough dependencies to update then trigger it.

Storybook, graphql, tiptap, and types were what I picked as they have multiples. Storybook I allowed major since that would probably have to be merged together anyway for it to not break. 

## 🧪 Testing

1. ?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,6 +17,7 @@ updates:
       storybook:
         patterns:
           - "@storybook*"
+          - "storybook*"
         update-types:
           - "major"
           - "minor"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,6 +13,32 @@ updates:
       interval: "weekly"
       day: "thursday"
     open-pull-requests-limit: 50
+    groups:
+      storybook:
+        patterns:
+          - "@storybook*"
+        update-types:
+          - "major"
+          - "minor"
+          - "patch"
+      graphql-codegen:
+        patterns:
+          - "@graphql-codegen*"
+        update-types:
+          - "minor"
+          - "patch"
+      tiptap:
+        patterns:
+          - "@tiptap*"
+        update-types:
+          - "minor"
+          - "patch"
+      types:
+        patterns:
+          - "@types*"
+        update-types:
+          - "minor"
+          - "patch"
 
   # This ignores all updates to tc-report, since it's now served as a static website.
   #


.github/dependabot.yml
```diff
@@ -13,6 +13,32 @@ updates:
       interval: "weekly"
       day: "thursday"
     open-pull-requests-limit: 50
+    groups:
+      storybook:
+        patterns:
+          - "@storybook*"
```
esizer, line 19: I think we will also need `storybook` here for the main package that is not namespaced behind the org

esizer (changes requested):
I think this is what we are looking for. Guess we just have to wait until one of these dependencies needs an update 😛 

Just one small suggestion for storybook.

.github/dependabot.yml
vd1992: Yes, yes you are right
https://github.com/GCTC-NTGC/gc-digital-talent/pull/9481/commits/954a1e0ce425c1a36b99b7edc26d26f2c0481811

esizer (approved):
LGTM! 👍 Now we pray 🙏 😅

PR #9500, opened 2024-02-23
[Bug] Pagination on filter changes

### LINKED ISSUE #9312: 🐛 If filters change, and current page is higher than max page, table appears to be empty
## 🐛 Bug

Changing the filters of the table doesn't reset which "page" you're on. Its possible to end up on a higher page than exists, and the table thinks there are no results when there should be some.

## 🦋 Expected Behaviour

If the filters change, table should be reset to the first page.

## 🕵️ Details

## 📋 Steps to Reproduce

1. Log is as admin
2. Go to a PoolCandidate table
3. Add some filters which decrease the number of results (and the number of pages) but still have some results.
4. Reset the filters
5. Go to the last page of the table
6. Add the same filters from before
7. Table appears empty

## 📸 Screenshot

## 🙋‍♀️ Proposed Solution

Reset the table page when filters are updated

## ✅ Acceptance Criteria

- [ ] When search or filters are updated, always go back to the first page of the table

yonikid15:
🤖 Resolves #9312 

## 👋 Introduction

- Thanks to @esizer for the help 🥇 
- Adds external state to ResponsiveTable
- Adds fix to all api tables: PoolCandidatesTable, UserTable, SearchRequestTable

## 🧪 Testing

1. Log is as admin
2. Go to a PoolCandidate/UserTable/SearchRequestTable 
3. Add some filters which decrease the number of results (and the number of pages) but still have some results.
4. Reset the filters
5. Go to the last page of the table
6. Add the same filters from before
7. Table appears on first page with correct number of rows

### DIFF
--- a/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
+++ b/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
@@ -277,6 +277,7 @@ const PoolCandidatesTable = ({
   const filterRef = React.useRef<PoolCandidateSearchInput | undefined>(
     initialFilters,
   );
+
   const [paginationState, setPaginationState] = React.useState<PaginationState>(
     initialState.paginationState
       ? {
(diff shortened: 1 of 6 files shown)


brindasasi:
looking good so far.. checking few more things

brindasasi (approved):
Reproduced issue on main and verified this branch resolved it.  LGTM! 🥳

PR #9504, opened 2024-02-26
[Fix] Log users out of all tabs

### LINKED ISSUE #9188: 🐛 Logging out in a second tab doesn't log out the first tab
## 🐛 Bug

When a second tab ends the user's session the first tab doesn't go to the logged-out page on navigate.  Instead the user gets a 404 type page or even the create account page.

## 🦋 Expected Behaviour

When the session has ended in a different tab, the first navigation of the first tab should redirect to logged-out.  This is the same behaviour that currently happens if the token expired.

## 📋 Steps to Reproduce

1. Log in in tab 1
2. Open app in tab 2
3. Log out in tab 1
4. Attempt to navigate in tab 2

## 📸 Screenshot

https://github.com/GCTC-NTGC/gc-digital-talent/assets/8978655/cb55e302-c222-43cf-b108-19ac5afa2f75

## ✅ Acceptance Criteria

- [ ] When the session ended in tab 1, tab 2 should redirect to logged-out on navigation
- [ ] Unskip cypress test "will affect all tabs when logged out" in login-logout.cy.ts

esizer:
🤖 Resolves #9188 

## 👋 Introduction

This updates our auth container to log users out of all active tabs at once.

## 🕵️ Details

This uses two different ways to accomplish it.

1. Broadcast channel affects all tabs and posts a message to tell other tabs to logout
2. An effect that watches for the access token to be set to null and logging out when doing so

### Broadcast channel

This is the common way of completing this task. On mount of the auth container, it creates a broadcast channel singleton. When the user is logged out, it sends a logout message to all tabs. Then, the tabs redirect when that specific message is received. 

### Storage effect

This method is not entirely necessary but since we cannot test multiple tabs in Cypress we need to manually remove the access token. This does actually have a use case though, if the access token is removed by any method outside of `logoutAndRefreshPage` (unlikely but possible) this will also log the user out of other tabs.

## 🧪 Testing

1. Build `npm run dev`
2. Login
3. Open a new tab
4. Logout of either tab
5. Confirm the other tab is automatically logged out
6. Repeat steps 2 and 3
7. Open local storage of one tab
8. Delete the access token
9. Confirm other tab is logged out

### DIFF
--- a/packages/auth/src/components/AuthenticationContainer.tsx
+++ b/packages/auth/src/components/AuthenticationContainer.tsx
@@ -132,6 +132,21 @@ const AuthenticationContainer = ({
     }
   }
 
+  // Logout if the access token is removed in another way other than
+  // the user logging out manually
+  useEffect(() => {
+    const logoutOnAccessTokenRemoved = (event: StorageEvent) => {
+      if (event.key === ACCESS_TOKEN && event.newValue === null) {
+        window.location.href = logoutRedirectUri;
+      }
+    };
+
+    window.addEventListener("storage", logoutOnAccessTokenRemoved);
+
+    return () =>
+      window.removeEventListener("storage", logoutOnAccessTokenRemoved);
+  });
+
   // We have saved it in local storage , then clear query parameters.
   useEffect(() => {
     if (newTokens?.accessToken) {
--- a/packages/auth/src/components/AuthenticationContainer.tsx
+++ b/packages/auth/src/components/AuthenticationContainer.tsx
@@ -11,6 +11,7 @@ import {
   ID_TOKEN,
   POST_LOGOUT_URI_KEY,
 } from "../const";
+import useLogoutChannel from "../hooks/useLogoutChannel";
 
 export interface AuthenticationState {
   loggedIn: boolean;
@@ -39,6 +40,7 @@ const logoutAndRefreshPage = (
   logoutUri: string,
   logoutRedirectUri: string,
   postLogoutUri?: string,
+  postLogoutMessage?: () => void,
 ): void => {
   defaultLogger.notice("Logging out and refreshing the page");
   // capture tokens before they are removed
@@ -68,6 +70,7 @@ const logoutAndRefreshPage = (
       authSessionIsCurrentlyActive = Date.now() < decodedAccessToken.exp * 1000; // JWT expiry date in seconds, not milliseconds
   }
 
+  postLogoutMessage?.();
   if (idToken && authSessionIsCurrentlyActive) {
     // SiC logout will error out unless there is actually an active session
     window.location.href = `${logoutUri}?post_logout_redirect_uri=${logoutRedirectUri}&id_token_hint=${idToken}`;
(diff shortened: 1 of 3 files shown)


packages/auth/src/hooks/useLogoutChannel.ts
```diff
@@ -0,0 +1,53 @@
...
+        channel.close();
+        isSubscribed.current = false;
+      }
+    };
+    // We only want this to run on mount
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  const postLogoutMessage = useCallback(() => {
+    channel?.postMessage(LOGOUT_MESSAGE);
```
petertgiles, line 45: Maybe we could log an error message if they try to post a logout message and the channel isn't ready?

packages/auth/src/components/AuthenticationContainer.tsx
```diff
@@ -68,6 +70,7 @@ const logoutAndRefreshPage = (
       authSessionIsCurrentlyActive = Date.now() < decodedAccessToken.exp * 1000; // JWT expiry date in seconds, not milliseconds
   }
 
+  postLogoutMessage?.();
```
petertgiles, line 73: Could you add a comment to indidate what this actually does?

petertgiles (changes requested):
Oh wow, this is super high-tech.  I didn't even know about broadcast channels.  :exploding_head:  Unfortunately it's not quite working for me.  I no longer get the weird "new user" behaviour but I'm getting a "not found" error instead of an actual logout.

https://github.com/GCTC-NTGC/gc-digital-talent/assets/8978655/2725a0fc-0b07-4185-ba58-25e07453b3fc

esizer:
> Oh wow, this is super high-tech. I didn't even know about broadcast channels. 🤯 Unfortunately it's not quite working for me. I no longer get the weird "new user" behaviour but I'm getting a "not found" error instead of an actual logout.

Oh this is weird! It is immediate (no navigation needed) in Chrome 😖 I'll do some more testing in firefox

petertgiles (approved):
After a full rebuild, it worked great!  :clap:

packages/auth/src/components/AuthenticationContainer.tsx
```diff
@@ -39,6 +40,7 @@ const logoutAndRefreshPage = (
   logoutUri: string,
   logoutRedirectUri: string,
   postLogoutUri?: string,
+  postLogoutMessage?: () => void,
```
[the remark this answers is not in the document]
petertgiles: Or even `broadcastLogoutMessage`?

mnigh (approved):
Works for me on macOS Safari, Chrome, Firefox, and Edge. Very nice work!

packages/auth/src/hooks/useLogoutChannel.ts
esizer: Added in [add error log when no channel on post](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9504/commits/f156c503ec8f686f8dcc69770ac9cf15f6ea1b73)

packages/auth/src/components/AuthenticationContainer.tsx
esizer: Added in [document post logout message in logout function](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9504/commits/e6a1af50871bbc6fbdd839e674f464e11bba36fd)

packages/auth/src/components/AuthenticationContainer.tsx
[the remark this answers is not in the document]
esizer: Love it! Updated in 
[update logout broadcast function name...](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9504/commits/6620e7e0ec877889ac2245d6a054125d4201f33f)

PR #9508, opened 2024-02-26
[Debt] Load mode specific images only

### LINKED ISSUE #6166: Perf - Don't double images for dark mode
## ♻️ Debt/Refactor

For pages with dark mode we're forcing every visitor to download both sets of images. We should avoid this doubling of data.

## 🙋‍♀️ Proposed Solution

Eric's note:
> Yeah, I'm pretty sure right now we are just hiding them with CSS but instead we should unmount it depending on the current mode.

## ✅ Acceptance Criteria

- [ ] Initial page load only requests assets needed for the current mode
- [ ] Assets for the opposite mode are only requested on mode switch

esizer:
🤖 Resolves #6166 

## 👋 Introduction

Updates images that have light/dark mode specific versions to only load the one for the active mode.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to pages that have mode specific images
    - Home
    - Sign in/Sign out
    - Error page
    - Browse pools
    - Support
3. Confirm there are no duplicate images for light/dark mode
4. Confirm the proper image loads

### DIFF
--- a/apps/web/src/components/FlourishContainer/FlourishContainer.tsx
+++ b/apps/web/src/components/FlourishContainer/FlourishContainer.tsx
@@ -1,5 +1,7 @@
 import React from "react";
 
+import { useTheme } from "@gc-digital-talent/theme";
+
 import desktopGraphicsLight2 from "~/assets/img/Desktop_Graphics_light_2.webp";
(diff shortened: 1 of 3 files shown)


petertgiles (approved):
Works great!  :clap: Chromatic doesn't handle this well, but I guess we already knew there were problems with Storybook + theme.

PR #9536, opened 2024-02-27
[Feature] Accessibility testing in Playwright

### LINKED ISSUE #7670: 🏈 a11y Axe testing on components
## ✍️ Details

We want to move away from Axe testing with Jest since it is slow and doesn't include Hydrogen styles. We would like to test in a browser, like Playwright, but want to run it on components, not full pages. 

This way, we will catch more WCAG conformance issues and accessibility/ usability issues 

REF: https://playwright.dev/docs/accessibility-testing

## ✅ Acceptance Criteria

- [ ] provide answer for how we can migrate Axe tests away from Jest

esizer:
🤖 Resolves #7670 

## 👋 Introduction

Solves the spike with testing accessibility without `jest-axe` by using `@axe-core/playwright`.

## 🕵️ Details

While component testing in Playwright is [still experimental](https://playwright.dev/docs/test-components), you can test full pages. The benefit of this is, we can interact with the app and test at different stages. Meaning, we could navigate to a page, open a dialog and run a test. This would catch violations that could be hiding behind not being mounted.

I still think `jest-axe` is a decent solution if we only use it on small, isolated components. I think we should probably move forward with this pattern for more complex testing and avoid doing any `jest-axe` testing inside of the `apps/**` folders. 

If we keep `jest-axe` to components in `packages/**` the speed should be less of an issue.

## 🧪 Testing

2. Run `npm run e2e:playwright -- -- -- accessibility.spec.ts`
3. Confirm it passes
4. Maybe force an error
5. Re-run command in step 2
6. Confirm it fails
7. View report to see violations

### ❓ How to force an error

You can edit the `LinkWell` component. 

- Add `data-h2-background-color="base(black)"` to the `Well`
- Add `data-h2-color="base(black.light)"` to the `Heading` inside the `Well`

## 📸 Screenshot

Screenshot 2024-02-27 161534

Screenshot 2024-02-28 121235

### DIFF
--- a/apps/playwright/fixtures/index.ts
+++ b/apps/playwright/fixtures/index.ts
@@ -43,9 +43,9 @@ export const test = base.extend<AppFixtures>({
     await context.close();
   },
 
-  makeAxeBuilder: async ({ page }, use, testInfo) => {
+  makeAxeBuilder: async ({ appPage }, use) => {
     const makeAxeBuilder = () =>
-      new AxeBuilder({ page }).withTags([
+      new AxeBuilder({ page: appPage.page }).withTags([
         "wcag21a",
         "wcag21a",
         "wcag2a",
--- a/apps/playwright/fixtures/index.ts
+++ b/apps/playwright/fixtures/index.ts
@@ -1,4 +1,4 @@
-import { Page, test as base } from "@playwright/test";
+import { test as base } from "@playwright/test";
 
 import auth from "~/constants/auth";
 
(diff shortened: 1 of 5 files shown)


mnigh (commented):
This is cool, but I can't get axe to fail. Can you provide instructions on exactly how to get the test to fail for a11y?

mnigh (approved):
🍭 Sweet!

PR #9548, opened 2024-02-28
[Debt] Reduce `framer-motion` initial bundle

### LINKED ISSUE #8404: ♻️ Reduce initial bundle size used by Framer Motion
## ♻️ Debt/Refactor

Currently, Framer Motion makes up a pretty big chunk of our initial bundle.  Can we reduce this?

image

Context:
https://github.com/GCTC-NTGC/gc-digital-talent/issues/7840#issuecomment-1789601147

## 🙋‍♀️ Proposed Solution

Simply moving Framer to a lazy loaded bundle might work only if that bundle isn't immediately required by a page load.  I know the IAP site uses Framer, for example.

On the Framer site they have a link discussing how to reduce bundle size: https://www.framer.com/motion/guide-reduce-bundle-size/.  Perhaps the LazyMotion component can defer the loading until after the page has loaded.  Or perhaps we can import only what we need (like we do with Lodash) to help tree shaking.

Or perhaps we can use react-spring instead of framer-motion.
    https://bundlephobia.com/package/react-spring@9.7.3
    https://bundlephobia.com/package/framer-motion@10.16.4

## ✅ Acceptance Criteria

- [ ] The initial bundle size used for animation is reduced for a home page load.
- [ ] If lazy loading is used, the additional bundle isn't immediately required for a home page load.

esizer:
🤖 Resolves #8404 

## 👋 Introduction

Reduces the initial bundle size of `framer-motion` by lazy loading the animations.

## 🕵️ Details

At first, this was causing an increased bundle size. After investigation, it turns out this was because we had `ui` setup to be an external package. Meaning, it was actually getting bundled twice (once in `apps/web` and again in `packages/ui` and using different contexts.

Since we don't actually need `ui` to be an external package and are not publishing it, I converted it to an internal package. Which means, it only gets bundled once in the app and uses the same context 🎉 

### Before

308590517-4b6319aa-3667-4336-a742-6d979e0121b1
308590367-31321a35-d58b-42fb-b9b6-050c9e937f59

### After

Screenshot 2024-02-28 134456
Screenshot 2024-02-28 134611

## 🧪 Testing

1. Build app `npm run build`
2. Confirm animations are working still
3. Confirm the bundle size is lower than on main

### DIFF
--- a/apps/web/src/components/Context/ContextProvider.tsx
+++ b/apps/web/src/components/Context/ContextProvider.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { HelmetProvider } from "react-helmet-async";
-import { MotionConfig } from "framer-motion";
+import { MotionConfig, LazyMotion } from "framer-motion";
 
 import { AppInsightsProvider } from "@gc-digital-talent/app-insights";
 import {
@@ -18,6 +18,8 @@ import { Announcer } from "@gc-digital-talent/ui";
 import { ThemeProvider } from "@gc-digital-talent/theme";
 import Toast from "@gc-digital-talent/toast";
 
+const loadFeatures = () =>
+  import("./motion-features").then((res) => res.default);
 interface ContextContainerProps {
   messages: Messages;
   children: React.ReactNode;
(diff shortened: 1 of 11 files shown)


mnigh (approved):
👍 

Screen Shot 2024-02-29 at 13 33 44

PR #9584, opened 2024-02-29
[Fix] Set admin theme on layout

### LINKED ISSUE #9309: 🐛 Navigating from IAP to Admin persists theme
## 🐛 Bug

If you start on the IAP with that colour theme and then navigate directly to any `/admin` path, the IAP theme persists and the GC Digital Talent theme is not toggled on.

## 🦋 Expected Behaviour

Theme properly toggles to the GC Digital Talent theme when navigating from an IAP page.

## 🕵️ Details

This was introduced in https://github.com/GCTC-NTGC/gc-digital-talent/pull/8117 when the `useLayoutTheme` was removed. This was toggling the theme back.

## 📋 Steps to Reproduce

1. Login as an admin
2. Navigate to a page that has the IAP theme (`/indigenous-it-apprentice`)
3. Navigate directly from there to `/admin`
4. Observe the theme does not change

## 📸 Screenshot

304176782-85760ffd-0b90-43a6-bd69-76b82ca37c20

## 🙋‍♀️ Proposed Solution

Add `useLayoutTheme("default");` to the `AdminLayout` component.

## ✅ Acceptance Criteria

- [ ] IAP theme does not persist on `/admin` pages

esizer:
🤖 Resolves #9309 

## 👋 Introduction

Fixes an issue where navigating to admin from IAP did not switch theme back.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to `/indigenous-it-apprentice`
3. Navigate directly to `/admin` with no intermediary step (must use the address bar, not links)
4. Confirm the theme switches back to the default colours

### DIFF
--- a/apps/web/src/components/Layout/AdminLayout/AdminLayout.tsx
+++ b/apps/web/src/components/Layout/AdminLayout/AdminLayout.tsx
@@ -21,6 +21,7 @@ import Footer from "~/components/Footer/Footer";
 import Header from "~/components/Header/Header";
 import SEO, { Favicon } from "~/components/SEO/SEO";
 import useRoutes from "~/hooks/useRoutes";
+import useLayoutTheme from "~/hooks/useLayoutTheme";
 import { checkRole } from "~/utils/teamUtils";
 import {
   pageTitle as indexPoolPageTitle,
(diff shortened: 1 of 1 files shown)


mnigh (approved):
🚢

Issue #8852, opened 2023-12-20
🏐 Generate a file for a user


petertgiles:
## ✨ Feature

We want to allow users to generate custom files for themselves.  For example, a bulk data download.  Let's figure out a good way to do that.

There is some interest in using this to generate rich reports (PDF, Excel) and zip multiple reports together, too.

## 🕵️ Details

Context: https://github.com/GCTC-NTGC/gc-digital-talent/pull/8757#discussion_r1433090442

## 🙋‍♀️ Proposed Implementation

Some thoughts:
- Generate the file and make it available from the storage in the Azure app service.
- Store it in the Azure version of an S3 bucket for download. (https://learn.microsoft.com/en-us/azure/architecture/aws-professional/storage#s3ebsefs-and-azure-storage)
- Some kind of blob storage service IMTD mentioned but we never set up.
- Laravel has lots of File Storage helpers.  https://laravel.com/docs/10.x/filesystem#downloading-files
- Laravel can stream responses to avoid using lots of memory. https://laracasts.com/discuss/channels/laravel/using-responsedownload-for-large-files
- Could use notifications to let user know file is ready and share download link
- downloaded files should be secured, only availabe for user who generated them
- Could hidden link (random filename) be sufficient? Maybe... if file/link is short lived. Better to have policy though.
- These generated files shouldn't last long in storage

## :beverage_box:  Timebox

3 days

## :question: Main question

What's the best way to generate a dynamic file and make it available to a user to download?

petertgiles:
I started playing around with [the Azure Storage emulator](https://github.com/Azure/Azurite) in a docker container and [the Azure PHP SDK](https://github.com/Azure/azure-sdk-for-php).  I posted a basic branch at https://github.com/GCTC-NTGC/gc-digital-talent/tree/8852-azure-blob-storage showing how to add, examine, and delete a container.  As the SDK is deprecated we would probably want to use the REST API directly in our code.  The SDK is a nice simple way to get started, though.  We might perhaps write [a custom Laravel Storage driver](https://laravel.com/docs/10.x/filesystem#custom-filesystems)

PR #9606, opened 2024-03-01
[Spike] User generated files from local storage

petertgiles:
🤖 Related to (but does not close) #8852 

> [!Warning]
> This PR should not be merged.  It is a proof of concept, not production-ready code.

## 👋 Introduction

This branch provides a sample solution to serving user generated files from the web server's local storage.

Pros:
- Fairly simple code.
- No extra infrastructure to set up.
- Can hook into Laravel auth.

Cons:
- Possible to fill up our storage.
- Files (probably?) lost when app service recycled.
- Ties up a PHP-FPM worker during download.
- (Probably?) not comparatively fast.

## 🕵️ Details

You can simulate the operation of a graphql response resulting in a file by running:
```graphql
mutation MyMutation {
  makeAFile
}
```
In reality, this just downloads an image from a lorem generator.  You must be logged in to do this.  It will save the file in a directory with your user ID.  The files go into a custom `api/storage/app/user-generated` directory locally but would probably go into '/tmp' somewhere in the app service.

To download the file you can use your browser, a tool like Postman, or utility like curl.  You must send the authorization header for the same user that made the file.

```bash
curl http://localhost:8000/api/user-generated-files/Synthesis_NASA_data_Vis_720p_YouTube.mp4  \
  -O \
  -H "Authorization: Bearer {accessToken}"
```

An alternative to enforcing a logged in user in the controller would be to provide a [signed URL](https://laravel.com/docs/10.x/urls#signed-urls) when generating the file.  This would then be shareable by the user.  You would still need to stream the file download through the controller, though.

To clean up old files, I added an Artisan command that would be run periodically, either manually or by a scheduler.
```bash
php artisan app:prune-user-generated-files
```
It's set to delete after 12 hours and the actual delete is commented out right now.  You can see a "1" on the rows that would be deleted.

## 🧪 Testing

- [ ] When authenticated you can make a file.
- [ ] When not authenticated you cannot make a file.
- [ ] When authenticated with the same user you can get the file.
- [ ] When authenticated as a different user you cannot get the file.
- [ ] When not authenticated  you cannot get the file.

## 📸 Screenshot

image

### DIFF
--- a/api/app/Http/Controllers/UserGeneratedFilesController.php
+++ b/api/app/Http/Controllers/UserGeneratedFilesController.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\Storage;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\HttpKernel\Exception\UnauthorizedHttpException;
+
+class UserGeneratedFilesController extends Controller
+{
+    /*
+    curl http://localhost:8000/api/user-generated-files/Synthesis_NASA_data_Vis_720p_YouTube.mp4  \
+     -O \
+     -H "Authorization: Bearer {accessToken}"
+   */
+
+    public function getFile(string $fileName)
+    {
+        // https://laravel.com/docs/10.x/authentication#accessing-specific-guard-instances
+        $userId = Auth::guard('api')->id();
+        throw_unless(is_string($userId), UnauthorizedHttpException::class);
+
+        $filePath = $userId.'/'.$fileName;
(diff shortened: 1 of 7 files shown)


api/app/Http/Controllers/UserGeneratedFilesController.php
```diff
@@ -0,0 +1,54 @@
...
+            $handle = Storage::disk('userGenerated')->readStream($filePath);
+            while (! feof($handle)) {
+                $buffer = fread($handle, 4096);
+                if (! is_bool($buffer)) {
+                    echo $buffer;
+                }
+            }
+            fclose($handle);
+        }, $fileName, $contentType);
+
```
brindasasi, line 52: I would think of deleting the file right after it is successfully downloaded. Might help save the storage space as well.

PR #9612, opened 2024-03-01
Create template for new notification issues

tristan-orourke:
We'll be creating a lot of notifications as part of https://github.com/GCTC-NTGC/gc-digital-talent/milestone/25 and after. I want to standardize the details we need to successfully add a new type of notification to the app.


esizer (approved):
LGTM! 👍

Issue #9463, opened 2024-02-22
🐛 Null state message on pool candidates table doesn't make sense


petertgiles:
## 🐛 Bug

If you're creating a new pool, the view candidates table asks you to use the "add a new item button" to add an item.  This is not possible in this table and there is no add button.  I'm not sure it even makes sense to have this tab with this table until the pool is published anyway.

## 📋 Steps to Reproduce

1. Create a draft pool
2. Navigate to the View Candidates tab

## 📸 Screenshot

![image](https://github.com/GCTC-NTGC/gc-digital-talent/assets/8978655/71e866c7-9148-4779-82d4-3bef4cbb20a2)

## 🙋‍♀️ Proposed Solution

(optional) Do you have a proposed solution?

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the bug was addressed.

- [ ] Criteria 1
- [ ] Criteria 2

petertgiles:
Same problem on skill tables.
![image](https://github.com/GCTC-NTGC/gc-digital-talent/assets/8978655/173f1140-0319-4f6a-9357-42e2db836268)
https://github.com/GCTC-NTGC/gc-digital-talent/pull/9614

mnigh:
@tristan-orourke any thoughts on my list of [possible solutions](https://github.com/GCTC-NTGC/gc-digital-talent/issues/9463#issuecomment-1995574122)?

petertgiles:
I think I'd like to see a combination of 2 and 3.  Have a super general default message like "There are no rows" and then have a prop to allow a customized, specific message.

tristan-orourke:
Looks like we already have a prop for customized, specific messages - we just never use it 😆. 
What if we make it a required, so anyone using the component at least has to think about it, but provide an extremely generic message (like @petertgiles suggested) in commonMessages?

tristan-orourke:
Huh... I went to refine the AC, but now I'm stuck on what a good generic description text would be, which isn't completely repetitive of "There aren't any items here yet".

brindasasi:
ok my turn now 😅

![Image](https://github.com/GCTC-NTGC/gc-digital-talent/assets/89482910/c62d8a04-d406-4e60-85f2-70d54e630966)

Keep the default message "There aren't any items here yet." 
Allow the description to be empty and only pass messages when needed.

This issue looks definitely awkward and very visible in many screens. 
Let's get this move on.

PR #9614, opened 2024-03-01
[Feature] Public skills page

### LINKED ISSUE #9540: ✨ New public skills page
## ✨ Feature

We're getting sooo many requests for people to browse our skills. We decided it's just best if we create a new page to view and browse all skills. Minimal dev effort for now please! We can iterate on it if needed.

1. New page should be at **talent.canada.ca/skills** or similar
2. Create a new page and add a filterable table with all the skills.
3. Descriptions (in description column) should wrap if they are long.
4. Default number of items on the table should be 20
5. No links to this page from other pages yet 🤷‍♀️

## 🎨 Design File

[Figma file](https://www.figma.com/file/f9Vpe5c2bAWXmtwlyueJDq/Candidates---All-skills?type=design&node-id=813%3A380&mode=design&t=jeBoGBIp48rvjy72-1) by @Jerryescandon 

## 📸 Screenshot

 
All skills page

image

## 🌎 Localization

EN: All skills
Explore all the skills on our site.

FR: Toutes compétences
Découvrez toutes les compétences sur notre site.

## ✅ Acceptance Criteria

- [ ] New page at new url `/skills`
- [ ] Skills table has columns for name, family, category and description
- [ ] Skills are filterable by category and family in filter dialog
- [ ] Search bar filters by name and/or description
- [ ] Default number of table items is 20
- [ ] Table uses client-side filtering
- [ ] Add cant find skill `<Well />` at the bottom of the page

yonikid15:
🤖 Resolves #9540 

## 👋 Introduction

- Adds public skills page with skills table
- Refactors the admin skill table, adding filtering, pagination override, and a prop to set add skill button above table (should only be seen on admin side).
- Public table should be set to a min of 20 items

## 🕵️ Details

- You will need to restart nginx to see the new page. Either restart the webserver container on docker or go into the shell of said container and run `nginx -s reload`

## 🧪 Testing

1. Restart nginx
2. Build app `npm run build`
3. Go to skills page `http://localhost:8000/en/skills`

## 📸 Screenshot

image


infrastructure/conf/nginx-conf-deploy/default
```diff
@@ -229,6 +229,11 @@ server {
         rewrite "^(?:/[a-z]{2})?/directive-on-digital-talent(/.*|$)" /apps/web/dist/index.html break;
     }
 
+    # rewrite possible lang prefix then "skills" to apps/web index.hstml
+    location ~ "^(?:/[a-z]{2})?/skills(/.*|$)" {
+        rewrite "^(?:/[a-z]{2})?/skills(/.*|$)" /apps/web/dist/index.html break;
```
petertgiles, line 234: Could you add the no-cache header to this match?  We want to make sure that clients are never caching index.html.

apps/web/src/pages/Skills/components/SkillTable.tsx
```diff
@@ -1,37 +1,102 @@
...
-export const SkillTable = ({ skills, title }: SkillTableProps) => {
+export const SkillTable = ({
+  skills,
+  title,
+  paginationState,
+  addButton,
+}: SkillTableProps) => {
   const intl = useIntl();
   const paths = useRoutes();
+  const searchParams = new URLSearchParams(window.location.search);
```
petertgiles, line 83: I think we usually use the React Router hook for search params, like https://github.com/GCTC-NTGC/gc-digital-talent/blob/66cec5433f1bb3bedbeffd1a8a3e2b7e6f3f22e4/apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx#L99 .

apps/web/src/pages/Skills/components/SkillFilterDialog.tsx
```diff
@@ -0,0 +1,95 @@
...
+import FilterDialog, {
+  CommonFilterDialogProps,
+} from "~/components/FilterDialog/FilterDialog";
+
+export type FormValues = {
+  skillFamilies?: string[];
+  skillCategories?: SkillCategory[];
+};
+
+const SkillFilterDialogData_Query = graphql(/* GraphQL */ `
```
petertgiles, line 24: I honestly don't know the answer to this.  With the new codegen preset is it possible to send this as a fragment at the same time as SkillTableSkills instead of two different requests?

apps/web/src/pages/Skills/SkillPage.tsx
petertgiles, line 1: In the show/hide columns (which isn't in the design?) you have keywords but it doesn't appear when checked.  I think Jerbo said keywords is deprecated anyway so maybe you just want to remove it from the dialog?

petertgiles: Keywords is also a "filter by..." choice.

apps/web/src/pages/Skills/components/SkillTable.tsx
petertgiles, line 1: Ugh, pagination.  Maybe out of scope for this issue but it is still possible go to page 14, apply a filter that would result in one page of results, and see nothing.

petertgiles (changes requested):
Wow, it turned out great!  :sunglasses: I flagged a few things to check on.

substrae:
@yonikid15

Well looks good to me!

apps/web/src/pages/Skills/SkillPage.tsx
yonikid15: > In the show/hide columns (which isn't in the design?) you have keywords but it doesn't appear when checked. I think Jerbo said keywords is deprecated anyway so maybe you just want to remove it from the dialog?

Hmm, it's showing up on my end 😕 
image

Either way we can remove it!

infrastructure/conf/nginx-conf-deploy/default
yonikid15: Sounds good! Fixed by [913b728](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9614/commits/913b72862967a8ea89abfeb63f60b97be4f8389c)

apps/web/src/pages/Skills/components/SkillTable.tsx
yonikid15: Nice catch. Fixed by [ef7aa26](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9614/commits/ef7aa26bad0ad98c9600244852920e36f8cc30f7)

apps/web/src/pages/Skills/components/SkillFilterDialog.tsx
yonikid15: Love it, fixed by [d9a1d63](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9614/commits/d9a1d633a370b466823792a8e697635996cf9655)

apps/web/src/pages/Skills/SkillPage.tsx
yonikid15: Removed by [8074e20](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9614/commits/8074e2032226701ecde69d20a14b6f6df04cdf56)

apps/web/src/pages/Skills/components/SkillTable.tsx
yonikid15: Adding a fix for this is trickier than I thought. Maybe pulling it out into another issue might be the best case if we want to role this out fast 🤷 

The reason is this is the only table where we are filtering client side, with no pagination data coming in from the api to setup the pagination state. Therefore, we would need to build that pagination logic client side.

petertgiles: Yeah, please create a ticket for this.  :+1:

petertgiles:
I noticed that the null state message doesn't make any sense here.
image
I added it on to https://github.com/GCTC-NTGC/gc-digital-talent/issues/9463 though.

petertgiles (approved):

apps/web/src/pages/Skills/components/SkillTable.tsx
brindasasi: @yonikid15 I see this issue in UAT. Do we have a sep ticket for this?

PR #9618, opened 2024-03-04
[Fix] In library skill count

### LINKED ISSUE #9602: 🐛 Filter skill counts for "My library" in select
## 🐛 Bug

When adding category-dependent skills from your library, skills are filtered but not the counts. This leads to a mismatch which can be confusing at first.

## 📋 Steps to Reproduce

1. Head to skill showcase for say, behavioural skills
2. Go to add skills to showcase
3. Note there may be discrepancy in option total and actual options

## 📸 Screenshot

309067117-523833cf-66a2-4b8b-b24f-1c45db013fc3

## 🙋‍♀️ Proposed Solution

Skill count reflects actual option count

## ✅ Acceptance Criteria

- [ ] Counts in sync

esizer:
🤖 Resolves #9602 

## 👋 Introduction

Fixes the skill count for "in library" skills when being filtered by category.

## 🕵️ Details

This applies a second filter when the category is selected to only get those skills in the users library and the category to more accurately reflect the number of skills that will be avilable.

## 🧪 Testing

1. Build `npm run dev`
2. navigate to a skill showcase
3. Click the add a new item button
4. Confirm the number of library skills in the family select matches those available in the combobox

## 📸 Screenshot

Screenshot 2024-03-04 081222
Screenshot 2024-03-04 081232

### DIFF
--- a/apps/web/src/components/SkillBrowser/utils.ts
+++ b/apps/web/src/components/SkillBrowser/utils.ts
@@ -363,6 +363,14 @@ export const getFamilyOptions = (
   category?: SkillCategory | "all",
   inLibrary?: Skill[],
 ): Option[] => {
+  const filteredSkills =
+    category !== "all"
+      ? skills.filter((skill) => skill.category === category)
+      : skills;
+  const filteredLibrary = inLibrary?.filter((librarySkill) =>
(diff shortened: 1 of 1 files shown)


yonikid15 (approved):
Looks good!
image

PR #9635, opened 2024-03-05
[Fix] Prevent combobox auto focus

### LINKED ISSUE #9598: 🐛 Combobox focused on mount
## 🐛 Bug

It seems as though the combobox (both single and multi) are being focus immediately on mount.

## 🦋 Expected Behaviour

Combobox does not recieve focus on mount.

## 📋 Steps to Reproduce

1. Navigate to a page with a combobox visible (admin create team form, or admin filter dialogs)
2. Observe the focus jumps to the last combobox on the screen

## 🙋‍♀️ Proposed Solution

I'm struggling with this one so don't have a proposed issue but I'd start be investigating the manual `.focus()` we have in those files 🤔 

## ✅ Acceptance Criteria

- [ ] Combobox is not focused immediately
- [ ] Other (intentional focuses) in the component function as expected:
  - [ ] On select Item
  - [ ] On clear
  - [ ]  On reset

esizer:
🤖 Resolves #9598 

## 👋 Introduction

Fixes an issue where the comobox input was receiving focus when it mounted.

## 🕵️ Details

We were using a `useEffect` to sync the options with the items passed in. The issue was, we initially set the options to an empty array so it was unnecessarily re-rendering on mount and getting focus. I'm still not sure why it was receiving focus on re-render but this solves that issue along with making the component more stable by avoiding that initial re-render.

To do this, we set the available options in the default state instead of the empty array. Then, we rely on `useMemo` to keep the items a bit more stable. Now, the `useEffect` does not detect an initial change in options so does not fire and cause unnecessary re-renders.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to any combobox (pool candidates filter dialog works well for this)
3. Confirm that no combobox receives focus immediately

### ◀️  Regression Test 

1. Navigate to a skill browser
2. Select a family
3. Open the skills combobox menu
4. Confirm the available options matches the items in the selected skill family
5. Repeat steps 2-4 with different families

### DIFF
--- a/packages/forms/src/components/Combobox/Multi.tsx
+++ b/packages/forms/src/components/Combobox/Multi.tsx
@@ -35,9 +35,12 @@ const Multi = ({
 }: MultiProps) => {
   const intl = useIntl();
   const [inputValue, setInputValue] = React.useState<string>("");
-  const [available, setAvailable] = React.useState<Option[]>([]);
+  const [available, setAvailable] = React.useState<Option[]>(options);
   const inputRef = React.useRef<HTMLInputElement | null>(null);
-  const items = isExternalSearch ? options : available;
+  const items = React.useMemo(
+    () => (isExternalSearch ? options : available),
+    [available, isExternalSearch, options],
+  );
 
   const handleInputChanged = (newQuery?: string) => {
     const query = newQuery ?? "";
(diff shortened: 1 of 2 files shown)


mnigh (changes requested):
The initial focus on the department combobox in `/admin/teams/create` is gone, but it seems now that when there are multiple comboboxes on the same page that the focus  switches to the next one after selection.

## 🎥 Screen recordings

https://github.com/GCTC-NTGC/gc-digital-talent/assets/3046459/a28e484c-418c-442c-befa-489ceac2d93a

https://github.com/GCTC-NTGC/gc-digital-talent/assets/3046459/2ff3f305-f478-4a8b-a9d6-067946340905

mnigh (approved):
LGTM 👍

PR #9654, opened 2024-03-06
[Debt] Improve Seeder Performance

### LINKED ISSUE #5045: Improve database seeder performance

petertgiles:
🤖 Resolves #5045 

## 👋 Introduction

This branch improves the seeder performance when seeding a local environment for development.

## 🕵️ Details

Several approaches were used:
- Refactor DatabaseSeeder code into separate seeders for better organization and visibility.
- Use naming convention to better organize larger number of seeders
  - *RandomSeeder for seeders using faker to make random data
  - *TestSeeder for seeders making expected data for local testing
  - *Seeder for seeders making "real" prod data
- Move database requests out of loops
- Reduce number of columns queried and use existing data instead of creating where possible
- Reduce number of rows created, especially User and PoolCandidate models
- use the Faker from the DI container instead of recreating it
- get rid of the extra database truncate: this is usually run for a fresh setup anyway

Out of scope - It is known that ROD data isn't seeded very well and I didn't fix that.

## 🧪 Testing
1. Reset your database and reseed `./artisan migrate:fresh --seed`

- [ ] Runs quicker than before
- [ ] Enough data to use the site somewhat realistically
- [ ] e2e tests run successfully

## 📸 Screenshot

image

```bash
#! /bin/bash

php artisan migrate:fresh
start_time=`date +%s`
php artisan db:seed && echo run time is $(expr `date +%s` - $start_time) s
```


esizer (approved):
Love it! 😍

PR #9658, opened 2024-03-07
[Bug] Associate ScreeningDecisionDialog skill levels with pool's

### LINKED ISSUE #9524: 🐛 ScreeningDecisionDialog should use skill level from Poster, not Applicant
## 🐛 Bug

The ScreeningDecisionDialog for skill requirements refers to a skill level. Currently, it shows the applicant's skill level (or "Not found" if the applicant hasn't added a level for this skill). However, it is intended to show the skill level required as part of the job poster (added in #8828 )

## 🦋 Expected Behaviour

The ScreeningDecisionDialog should use the skill level required for the PoolSkill.

## 🕵️ Details

Pool skill levels are added in #8828 

## 📋 Steps to Reproduce

I had to look at the code to check...

## 📸 Screenshot

Image

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] If a PoolSkill has a required skill level, it appears in any associated ScreeningDecisionDialogs

## 🛑 Blockers

```[tasklist]
### Blocked By
- [ ] #8828 
```

vd1992:
🤖 Resolves #9524 

## 👋 Introduction

Have the screening dialogs draw their skill level from the pool itself and not from the user.

## 🧪 Testing

1. Setup pool with poster, assessments, and a candidate
2. Go to screening dialog and observe skill level, should match the pool's skill level

## 📸 Screenshot

Dialog view

image

Matches poster

image

Storybook dialog

image

### DIFF
--- a/apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.tsx
+++ b/apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { useIntl } from "react-intl";
+import { IntlShape, useIntl } from "react-intl";
 import { SubmitHandler } from "react-hook-form";
 import { useMutation } from "urql";
 
@@ -18,7 +18,6 @@ import {
   Skill,
   SkillCategory,
   UpdateAssessmentResultInput,
-  UserSkill,
   graphql,
 } from "@gc-digital-talent/graphql";
 import {
@@ -43,7 +42,9 @@ import {
 import { toast } from "@gc-digital-talent/toast";
 import {
   getAssessmentDecisionLevel,
+  getBehaviouralSkillLevel,
   getTableAssessmentDecision,
+  getTechnicalSkillLevel,
 } from "@gc-digital-talent/i18n/src/messages/localizedConstants";
 
 import { getExperienceSkills } from "~/utils/skillUtils";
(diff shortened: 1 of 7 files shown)


apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.tsx
```diff
@@ -59,23 +60,36 @@ import useHeaders from "./useHeaders";
...
+  let skillLevel = "";
+  if (poolSkill?.requiredLevel && poolSkill.skill) {
+    skillLevel =
+      poolSkill.skill.category === SkillCategory.Technical
+        ? intl.formatMessage(getTechnicalSkillLevel(poolSkill.requiredLevel))
+        : intl.formatMessage(getBehaviouralSkillLevel(poolSkill.requiredLevel));
+  } else skillLevel = intl.formatMessage(commonMessages.notFound);
+
+  return skillLevel;
+};
```
esizer, line 79: I dont think this is part of the scope, unless you really feel like it 😅 

But I think we have this pattern for both name and definition of skills levels a few places now. Would be nice if we could create a small helper so we don't need to keep repeating ourselves. Do you think I should create a ticket for that?

esizer (approved):
Cool 😎

apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.tsx
vd1992: Yeah, I think it would be a good idea so please go ahead. I think this logic will probably appear in more places so there will be further repeating,

PR #9676, opened 2024-03-07
[Debt] Consolidate `Pill` and `Chip` components

### LINKED ISSUE #9651: ♻️ Consolidate and simplify the chip component
## ✨ Summary

[Visit the Figma file](https://www.figma.com/file/guHeIIh8dqFVCks310Wv0G/Style-library?type=design&node-id=5-14&mode=design) • Contact @substrae with design questions

The app currently has too many variations of the `Chip` / `Pill` component, only some of which actually match designs provided in the style library. The goal of this ticket is to consolidate the existing `Chip` and `Pill` components into a single `Chip` component, bring the styles inline with Figma, and address how and when to use each of the two styles.

There should only be 2 different styles of chip:
* The standard chip, which has color variations, an optional icon on the left, and an optional dismiss button on the right
* The `Combobox` specific chip, which has a single color option, no icon on the left, and always has a dismiss button on the right

For all intents and purposes, the standard chip should be what we use in every instance except for the `Combobox`. 

There are also issues with our use of chips in certain locations that don't warrant them (see tables), however that is out of scope for this ticket.

```[tasklist]
### ✅ Requirements
- [ ] Consolidate the existing `Chip` and `Pill` components into `Chip`
- [ ] Retain the separation between the standard `Chip` and the `Combobox` selection blocks
- [ ] Ensure the standard `Chip` has the correct color variations
- [ ] Ensure the standard `Chip` has the correct padding and gap values (`x.5, x.15`, `x.25` respectively)
- [ ] Ensure the standard `Chip` has an optional icon that is colored and sized properly (`x.5`)
- [ ] Ensure the standard `Chip` has an optional dismiss button that is colored and sized properly (`x.5`) and uses the `x-circle` mini Heroicon
- [ ] Ensure the icon alignment is centered with the text
- [ ] Ensure that dismissable chips have the correct `hover` and `focus-visible` effects
- [ ] Ensure the standard `Chip`'s text is bolded
- [ ] Triple check dark mode
- [ ] Triple check the IAP theme
```

##  🔍 Details

<details>
  <summary>📸 Chips (light mode)</summary>
  
</details> 

<details>
  <summary>📸 Chips (dark mode)</summary>
  
</details>

esizer:
🤖 Resolves #9651 

## 👋 Introduction

This merges the `Pill` into the `Chip` component and replaces all references to the `Pill`.

## 🕵️ Details

I really wanted to render `button` for the dismissible chips but that would have meant some polymorphism on the component and I didn't think it was worth the effort. Instead, I reimplement the `role=button` for it which is more code but still less than the polymorphic version.

With that mentioned, right now you can use delete and backspace to remove. I felt like this made sense, does anyone think this should change? 

## 🧪 Testing

1. Review chromatic for diffs
2. Run storybook `npm run storybook:design-system`
3. Confirm the chips match [the mockup](https://www.figma.com/file/guHeIIh8dqFVCks310Wv0G/Style-library?type=design&node-id=11-4252&mode=design&t=iLs3pAfD7fWrP0t5-0)
4. Confirm no more references to `Pill` component

## 📸 Screenshot

Screenshot 2024-03-07 093520
Screenshot 2024-03-07 093523


petertgiles (approved):
It works really well!  Love all the little tidy-ups too.

The dismiss keyboard interaction seems a little weird to me.  Since the entire chip can be clicked and that's the only possible interaction I would expect to be able to dismiss a focused chip with space or enter, like a button.

esizer:
> The dismiss keyboard interaction seems a little weird to me. Since the entire chip can be clicked and that's the only possible interaction I would expect to be able to dismiss a focused chip with space or enter, like a button.

Yeah, I tend to agree with that. I just wasn't absolutely sure 😅 Thanks for answering the question 👍 

I went ahead and added in them in [add space and enter to delete keys](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9676/commits/1b6407ce22b5f9eb85a8249c516756ca3a9080ca)

petertgiles (approved):
Nice, I like the dismiss.

PR #9678, opened 2024-03-07
[Tooling] Test for changes to list of duplicates

### LINKED ISSUE #9435: 🛠️ Warning when duplicate strings during `intl-extract` script
## 🛠️ Tooling

Display warning when there are duplicate strings with a list of the strings during `intl-extract` script. Possibly during `check-intl` as well.

## Proposed Implementation

- Add an existing duplicates to a whitelist to avoid worrying about them

## ✅ Acceptance Criteria

- [ ] warning displayed with list of strings when there are duplicate strings during `intl-extract` script
- [ ] can ignore duplicate strings by adding them to a whitelist

petertgiles:
🤖 Resolves #9435 

## 👋 Introduction

This branch adds a Jest test to flag when the list of duplicates changes.

## 🕵️ Details

The test uses Jest's ability to do [snapshot testing](https://jestjs.io/docs/snapshot-testing).  An initial snapshots is saved and then any time the actual result doesn't match the saved snapshot, the test fails.  When a change is expected (valid duplication or fixing duplication) then the user can save the updated snapshot and commit the change to the repo along with the change.

## 🧪 Testing

1. Run the test `npx jest ./src/lang/lang.test.ts --testNamePattern="should have no changes to duplicate strings"` 

- [ ] Test initially passes
- [ ] Add a duplicate translation and the test fails
- [ ] Save the new snapshot and the test passes again
- [ ] Add a non-duplicated translation and the test passes

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/lang/__snapshots__/lang.test.ts.snap
+++ b/apps/web/src/lang/__snapshots__/lang.test.ts.snap
@@ -1,0 +1,83 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`apps:web language should have no changes to duplicate strings 1`] = `
+"[
+    {
+        "en": "Education requirement",
+        "fr": [
+            "Exigence d’études",
+            "Exigences pédagogiques"
+        ]
+    },
+    {
+        "en": "Career timeline",
+        "fr": [
+            "Parcours professionnel",
+            "Échéancier de carrière"
+        ]
+    },
+    {
+        "en": "Screened in",
+        "fr": [
+            "Sélectionné(e)",
+            "Accepté"
+        ]
+    },
+    {
+        "en": "Language requirement",
+        "fr": [
+            "Exigence linguistique",
+            "Exigences relatives à la langue"
+        ]
+    },
+    {
+        "en": "N/A",
+        "fr": [
+            "S.O.",
+            "S. O."
+        ]
+    },
+    {
+        "en": "Minimum education",
+        "fr": [
+            "Niveau d'études minimal",
+            "Études minimales"
+        ]
+    },
+    {
+        "en": "View",
+        "fr": [
+            "Voir",
+            "Affichage"
+        ]
+    },
+    {
+        "en": "Required",
+        "fr": [
+            "Requises",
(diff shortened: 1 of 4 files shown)


mnigh (approved):
🆒, works for me! It would be nice to have this somewhere that wasn't in `packages/i18n` since there exists the `web` and `i18n` division but I don't know where else (noticed you were likely thinking about this with 599e564ac0d2638f5ca5851462827fabcc768745).

esizer:
> 🆒, works for me! It would be nice to have this somewhere that wasn't in `packages/i18n` since there exists the `web` and `i18n` division but I don't know where else (noticed you were likely thinking about this with [599e564](https://github.com/GCTC-NTGC/gc-digital-talent/commit/599e564ac0d2638f5ca5851462827fabcc768745)).

I mean, we _ccoouullddd_ start running tests from the root if the span across multiple packages 😅 But that also feels a bit awkward.

petertgiles:
>  It would be nice to have this somewhere that wasn't in packages/i18n since there exists the web and i18n division

petertgiles:
> It would be nice to have this somewhere that wasn't in packages/i18n since there exists the web and i18n division 
Yeah, I was thinking about this.  You could have two different identical tests in both just checking their own message files but then you couldn't catch duplicates between the two.

petertgiles:
> I mean, we _ccoouullddd_ start running tests from the root if the span across multiple packages 😅 But that also feels a bit awkward.

Yeah, that would be awkward for sure.  I think it's perhaps a smell that maybe these message files shouldn't be split?

esizer:
> Yeah, that would be awkward for sure. I think it's perhaps a smell that maybe these message files shouldn't be split? It's annoying that you have to run all the translation tooling on two sets of files, too.

I totally agree. We could do something like what we are doing with `graphql` and just run the tooling in `i18n` and then import everything from there instead? I imagine we could just do:

```sh
formatjs extract '../../**/src/*.{ts,tsx}' --ignore './**/*.d.ts' --out-file src/lang/en.json --id-interpolation-pattern [sha512:contenthash:base64:6]`
```

PR #9682, opened 2024-03-07
[Fix] Display of skill accordion subtitle

### LINKED ISSUE #9672: ⌨ Lists direct child must be `li`
## ⌨ Accessibility Issue

Our new lists of skill levels on the pool advertisement page have parentheses as direct children but should be `li`.

### WCAG 2.1 Criteria

- [1.3.1: Info and relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships)
- [Technique H48: Using ol, ul and dl for lists or groups of links](https://www.w3.org/WAI/WCAG21/Techniques/html/H48.html)
- [ul and ol must only directly contain li, script or template elements](https://dequeuniversity.com/rules/axe/4.8/list)

## 🦋 Expected Behaviour

the `ul` element should only have `li` as the direct child.

## 📋 Steps to Reproduce

1. Navigate to a pool advertisement
2. Open then inspector targeting the list of skill levels on the skill accordions
3. Confirm that parentheses exist as the first and last child of the list

## 📸 Screenshot

Screenshot 2024-03-07 080805

## 🙋‍♀️ Proposed Solution

1. This does not seem to be an actual list, we could just remove the `ul` and `li` elements and use `span` instead
2. The parentheses dont't seem to serve a purpose so we could remove them
3. If we need them there and there is a purpose, they must be moved outside of the `ul` element

## ✅ Acceptance Criteria

- [x] Option 1: `ul` and `li` elements have been replaced with `span`
- [x] Option 2: Parentheses have been removed
- [ ] Option 3: Parentheses moved outside of the `ul` element

esizer:
🤖 Resolves #9672 

## 👋 Introduction

Fixes the display of the skill accordion subtitle by removing the parentheses and replacing the list with `span` tags.

## 🧪 Testing

1. Build `npm run dev`
2. Navigate to a job poster
3. Confirm the parentheses are no longer present in the subtitle
4. Confirm the list has been replaced by spans

## 📸 Screenshot

Screenshot 2024-03-07 115744

### DIFF
--- a/apps/web/src/pages/Pools/PoolAdvertisementPage/components/SkillAccordion.tsx
+++ b/apps/web/src/pages/Pools/PoolAdvertisementPage/components/SkillAccordion.tsx
@@ -92,26 +92,17 @@ const SkillAccordion = ({ poolSkill, required }: SkillAccordionProps) => {
         });
 
   const accordionSubtitle = (
-    <ul
+    <span
       data-h2-color="base(black.light)"
       data-h2-font-size="base(caption)"
-      data-h2-padding-left="base(0)"
       data-h2-margin-top="base(x.5)"
+      data-h2-display="base(flex)"
+      data-h2-gap="base(x.5)"
     >
-      (
-      <React.Fragment key={poolSkill.skill.name.en}>
-        <li data-h2-padding-left="base(0)" data-h2-display="base(inline)">
-          {skillLevelItem}
-        </li>
-        <span data-h2-margin="base(0 x.5)" aria-hidden>
-          •
-        </span>
-        <li data-h2-padding-left="base(0)" data-h2-display="base(inline)">
-          {screeningTime}
-        </li>
-      </React.Fragment>
-      )
-    </ul>
+      <span>{skillLevelItem}</span>
+      <span>&bull;</span>
+      <span>{screeningTime}</span>
+    </span>
   );
 
   return (
--- a/apps/web/src/pages/Pools/PoolAdvertisementPage/components/SkillAccordion.tsx
+++ b/apps/web/src/pages/Pools/PoolAdvertisementPage/components/SkillAccordion.tsx
@@ -93,14 +93,16 @@ const SkillAccordion = ({ poolSkill, required }: SkillAccordionProps) => {
 
   const accordionSubtitle = (
     <span
+      data-h2-align-items="base(flex-start) p-tablet(center)"
       data-h2-color="base(black.light)"
       data-h2-font-size="base(caption)"
       data-h2-margin-top="base(x.5)"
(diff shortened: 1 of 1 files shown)


mnigh (commented):
This PR meets the acceptance requirements by removing the parentheses and replacing `ul` and `li` elements with a `span`. However, on smaller screen widths, the elements do not break down very well for readability. Would it make sense to force `skillLevelItem` and `screeningTime` on their own lines when there is less space?

## Screenshot

Screen Shot 2024-03-07 at 12 13 36

esizer:
> This PR meets the acceptance requirements by removing the parentheses and replacing `ul` and `li` elements with a `span`. However, on smaller screen widths, the elements do not break down very well for readability. Would it make sense to force `skillLevelItem` and `screeningTime` on their own lines when there is less space?

The layout of this trigger component is a little tricky to deal with and not great for mobile (even before the addition of the subtitle). I did my best in [improve mobile layout for skill accordion triggers](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9682/commits/3042a395a9bdfe2e80adb6ac89d825ece3d76a73)

mnigh:
> > This PR meets the acceptance requirements by removing the parentheses and replacing `ul` and `li` elements with a `span`. However, on smaller screen widths, the elements do not break down very well for readability. Would it make sense to force `skillLevelItem` and `screeningTime` on their own lines when there is less space?
> 
> The layout of this trigger component is a little tricky to deal with and not great for mobile (even before the addition of the subtitle). I did my best in [improve mobile layout for skill accordion triggers](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9682/commits/3042a395a9bdfe2e80adb6ac89d825ece3d76a73)

It's an improvement, yes, but the Required/Optional seems oddly place on a small screen and as a whole, is trying to fit a lot of information on an accordion trigger. Maybe worth a look at by @GCTC-NTGC/designers to refactor for a future version though?

## Screenshot
Screen Shot 2024-03-07 at 14 29 05

mnigh (approved):
Meets acceptance criteria 👍

substrae:
@esizer

I'm sure I could finesse this further, but this is where I would lean towards for now
image

esizer:
> What if you `absolute` the context, give it a `max-width` and then add padding to the title equal to the `max-width` value?

Sadly, absolutely positioned elements cause issues with ATs and reading order. We have encountered this before and are doing our best to avoid it. Since those icons can be informative it may be best to avoid that.

We may need to just revisit and refactor this component again to achieve that mobile layout. 👎 

As a side note, for components like this, how much effort would it be to get mobile designs? We kinda just guessed on this one and obviously missed the mark 😅

PR #9689, opened 2024-03-07
[Debt] Removes `MissingSkills` component

### LINKED ISSUE #9673: ♻️ `MissingSkills` component no longer in use
## ♻️ Debt/Refactor

The `MissingSkills` block conditional rendering logic is never satisfied so the component will never be rendered and should be removed entirely.

## 🕵️ Details

This component used to be used to show an alert when a user navigated away from an application to edit their experience and skills. Since we no longer do this and edit the skills directly in the application workflow, this block is no longer needed.

## 🙋‍♀️ Proposed Solution

We can delete the following:

- `MissingSkills` component
- Code in `CareerTimelineAndRecruitments` referencing it

## ✅ Acceptance Criteria

- [ ] `MissingSkills` component deleted
- [ ] `CareerTimelineAndRecruitment` component updaed to no longer reference it

mnigh:
🤖 Resolves #9673.

## 👋 Introduction

This PR removes the `MissingSkills` and several other files that are only referenced by that component as well as any related i18n strings.

## 🧪 Testing

1. Verify no references to `MissingSkills` component or any files only referenced within that file exist in codebase
2. Verify app functions as before


esizer (approved):
LGTM! 👍 

Conflict should be easy since you can just delete the file 😆

mnigh:
> LGTM! 👍
> 
> Conflict should be easy since you can just delete the file 😆

No as easy as that unfortunately, `fr.json` too!

PR #9694, opened 2024-03-07
[Feature] Adds new mutations for notifications

### LINKED ISSUE #9554: ✨ Add more API operations on notifications
## ✨ Feature

New base notification mutations added to API

## 🕵️ Details

Delete a notification -> soft delete
Mark as unread, so nulls out `read_at`
mark all as read -> marks all unread notifications of a given user as read

notifications and unreadNotifications fields are paginated. Not sure if we can add that right to the current fields on the User type, or if that means they need new top-level queries.

## ✅ Acceptance Criteria

- [ ] Three mutations above added (delete notification, mark as unread, mark all as read)
- [ ] Queries for notifications should be paginated. when querying notifications and unreadNotifications we can specify limit and page number. They should always be sorted chonologically descending.
- [ ] Some testing

esizer:
🤖 Resolves #9554 

## 👋 Introduction

Adds new mutations for reading all, unreading and deleting notifications.

## 🕵️ Details

I hope I did this correctly. Not sure if there is a better way to do the mutations but I just followed the exact patter laid out by the existing mutation for marking as read. 😅 

Also, for the tests, I created a new notification that just returns a simple string that is passed in for east testing. The annoying bit is I had to add it to the schema to test the new mutations. Do we hate this? I kind of do. Should I just do the extra work and use the existing notification with faker?

## 🧪 Testing

This one is a little tricky to test since we don't have any real notifications yet. I just inserted the test notification into the database manually. Use the following command but replace `{userId}` with the ID for the default auth user.

1. Login as admin `@admin@test.com`
2. Navigate to a pool candidate
3. Change their status a few times to generate notifications for them
4. Set your default auth to the sub of the user you changed statuses for (found it was easiest to just use `admin@test.com`)
5. Test the different operations in graphqil
    - `me.notifcations`
    - `me.unreadNotifications`
    - `deleteNotification`
    - `markAllNotificationsAsRead`
    - `markNotificationAsUnread`


api/app/Notifications/Test.php
```diff
@@ -0,0 +1,33 @@
...
+
+namespace App\Notifications;
+
+use Illuminate\Bus\Queueable;
+use Illuminate\Notifications\Notification;
+
+/**
+ * Test Notification
+ *
+ * Used for PPH unit testing graphql operations
```
mnigh, line 11:
```suggestion
 * Used for PHPUnit testing graphql operations
```

api/graphql/schema.graphql
```diff
@@ -135,6 +135,14 @@ interface Notification {
...
 }
 
+type TestNotification implements Notification {
+  id: ID!
+  readAt: DateTime @rename(attribute: "read_at")
+  createdAt: DateTime @rename(attribute: "created_at")
+  updatedAt: DateTime @rename(attribute: "updated_at")
+  data: String @rename(attribute: "data")
+}
+
```
vd1992, line 145: Yeah, I am not a fan of this either. I don't think it is a good idea to add to the schema.
Writing up a test schema would be tedious I think? So maybe faker it is.

esizer:
> I believe you are missing this part of the AC
> 
> > Queries for notifications should be paginated. when querying notifications and unreadNotifications we can specify limit and page number. They should always be sorted chonologically descending

oh wow, yeah that is a big miss. Thanks for identifying it. I had to do some refactoring to get pagination working but I think it all works 🤔 

[add enriched notification trait for graphql pagination](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9694/commits/8bc8be0d8eba2c7507bf56d913d55d95d7a57186)

api/graphql/schema.graphql
esizer: Agreed, turns out faker wasn't really needed not sure what I was thinking with that 😆 

[update tests to not need test notification](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9694/commits/d66dad8a0cf33f250a8fa76b3b3a7e189bf56617)

api/app/GraphQL/Mutations/MarkNotificationAsRead.php
```diff
@@ -22,7 +22,6 @@ public function __invoke($_, array $args)
 
         if (! is_null($notification)) {
             $notification->markAsRead();
-            User::enrichNotification($notification);
```
tristan-orourke, line 25: Why delete this? I believe we should run this wherever we return a notification, no?

api/app/Models/User.php
```diff
@@ -833,43 +833,6 @@ public function setRoleAssignmentsInputAttribute($roleAssignmentHasMany)
         }
     }
 
-    // reattach all the extra fields from the JSON data column
-    public static function enrichNotification(object $notification)
```
tristan-orourke, line 837: Don't we still want this enrich function? Maybe in Notification.php's boot function?

api/tests/Feature/NotificationTest.php
```diff
@@ -0,0 +1,182 @@
...
+                                'id' => $this->notification->id,
+                                'readAt' => null,
+                            ],
+                        ],
+                    ],
+                ],
+            ]);
+    }
+
+    public function testUnreadNotification(): void
```
tristan-orourke, line 99:
```suggestion
    public function testUnreadNotificationMutation(): void
```
Stupid english language where read is both a verb and an adjective >:(

api/tests/Feature/NotificationTest.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        $response = $this->actingAs($this->user, 'api')
+            ->graphQL(/** @lang GraphQL */ '
+                mutation unReadNotification($id: UUID!) {
+                    markNotificationAsUnread(id: $id) {
+                        id
+                        readAt
+                    }
+                }
+            ', ['id' => $this->notification->id]);
```
tristan-orourke, line 109: Doesn't this notification start out as unread? I'd set read_at to non-null before running the mutation.

api/app/Models/Notification.php
```diff
@@ -0,0 +1,50 @@
...
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Notifications\DatabaseNotification;
+use Illuminate\Support\Facades\Auth;
+
+/**
+ * Class Notification
+ *
+ * Note: Exists for lighthouse-php pagination
+ */
+class Notification extends DatabaseNotification
```
tristan-orourke, line 14: Let's make Notifications soft-deletable. Not sure if we can set that here or if we need to set it on the base class somehow.

api/tests/Feature/NotificationTest.php
```diff
@@ -0,0 +1,182 @@
...
+                }
+            ', ['id' => $this->notification->id])->assertJson([
+                'data' => [
+                    'deleteNotification' => [
+                        'id' => $this->notification->id,
+                    ],
+                ],
+            ]);
+
+        $this->assertDatabaseMissing('notifications', ['id' => $this->notification->id]);
```
tristan-orourke, line 133: Actually, I'd like to make the delete mutation a soft-delete.

api/tests/Feature/NotificationTest.php
```diff
@@ -0,0 +1,182 @@
...
+                $this->newStatus,
+                $this->pool->id,
+                $this->pool->name
+            )
+        );
+
+        $this->notification = $this->user->unreadNotifications()->first();
+    }
+
+    public function testQueryNotification(): void
```
tristan-orourke, line 81: Can you add tests for the query with filters? Especially onlyUnread.

api/tests/Feature/PoolCandidateUpdateTest.php
```diff
@@ -310,30 +310,30 @@ public function testCanQueryForNotifications()
                 /** @lang GraphQL */
                 '
                 query myNotifications {
```
tristan-orourke, line 312: Maybe we should move these tests into the new file?

tristan-orourke (changes requested):

api/app/Models/User.php
esizer: It is, it just isn't as explicit so I can add some comments. It uses the magic `getAttribute` method to enrich it.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/c7559b23cff31b3032e9b0c480e152e105aa63ee/api/app/Models/Notification.php#L16-L25

api/app/GraphQL/Mutations/MarkNotificationAsRead.php
esizer: See: https://github.com/GCTC-NTGC/gc-digital-talent/pull/9694#discussion_r1520347907

api/tests/Feature/NotificationTest.php
esizer: lol it was even worse than that... I was attempting to grab the wrong field out of the response so it was always going to be null 😅 

Fixed in [add read test, fix unread test](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9694/commits/4f77c45161b860733739143bc899a8a1d7c9da34)

api/tests/Feature/PoolCandidateUpdateTest.php
esizer: Yeah, I can do that. Were you thinking of moving it to the `NotificationTest` or creating a new `PoolCandidateStatusChangedNotifcationTest` file?

api/app/Models/Notification.php
esizer: Updated to be soft deletes. [soft delete notifications](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9694/commits/fd2f7757e5d0e82478f0ed6990dbd4792b0a74e2)

api/tests/Feature/NotificationTest.php
esizer: Updated to be soft deletes. [soft delete notifications](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9694/commits/fd2f7757e5d0e82478f0ed6990dbd4792b0a74e2)

api/tests/Feature/NotificationTest.php
esizer: Oh right, that makes sense 😛 Added in [add filter tests for notifications](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9694/commits/d9fcf8946883af68411c52da7259df7731d175d5)

api/app/Models/User.php
tristan-orourke: 😮

api/tests/Feature/PoolCandidateUpdateTest.php
tristan-orourke: I was thinking they belong in `NotificationTest`.

api/tests/Feature/NotificationTest.php
tristan-orourke: Okay, that's good, but this notification still _starts out_ as unread, doesn't it? 😆

tristan-orourke (changes requested):

api/tests/Feature/NotificationTest.php
esizer: It does but as it stands, our tests are ran sequentially (default unless you turn on parallel tests). So the previous test that marks it as read is ran first and then we mark it as unread here. So it goes like this where each case is testing one thing in sequential order.

1. Query notification
2. Mark as read
3. Mark as unread
4. Delete

I did this so we don't have to keep massaging data or create new data for each test case and test at the same time as creating the data for the next test. 

Though, we could assume they are not ran sequentially and just massage the data in each case or merge them into a single test that does read/unread.

tristan-orourke: Hmmm. Are you sure about that? I was under the impression that by using the `use RefreshDatabase;` trait, the database is reset and the `setUp` function is run before every single test function. :/

esizer: I have seriously been writing PHPUnit tests based off of a false assumption for years and I'm honestly shocked it hasn't bitten me yet 😆 I always thought

- Class = Test
- Method = Test case

But, it is the opposite! So when laravel told me that it refreshes on every test, I assumed after all methods in the class were called 😬 

I have updated the test to mark the notification as read at the start 😅 [ensure notifications is read before testing unread mutation](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9694/commits/14dd2688874139c07295849e501910c34b07b591)

api/tests/Feature/PoolCandidateUpdateTest.php
esizer: Sounds good 👍 Moved in [consolidate notification tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9694/commits/8a13a0a5cb7126067f4e3d734da1dd5f437b10ac)

api/tests/Feature/NotificationTest.php
```diff
@@ -281,4 +299,168 @@ public function testDateRangeFilter()
...
+                ],
+            ]);
+    }
+
+    /**
+     * Test that use can dismiss notifications
+     *
+     * @return void
+     */
+    public function testCanDismissNotifications()
```
tristan-orourke, line 409: Okay, sorry, I just took a closer look at this test, and... I think we can delete it? It seems like just a more complicated version of `testReadNotificationMutation`. Please let me know if I'm missing something.

api/tests/Feature/NotificationTest.php
```diff
@@ -0,0 +1,467 @@
...
+
+        }
+    }
+
+    /**
+     * Test that notifications are generated for status changes
+     *
+     * @return void
+     */
+    public function testStatusChangeCausesNotifications()
```
tristan-orourke, line 309: Sorry, after taking a closer look at this, I think your other suggestion was correct. This deserves to be in its own file, specific to the Status Change notification.

api/tests/Feature/NotificationTest.php
```diff
@@ -0,0 +1,467 @@
...
+        // check that notification was fired
+        NotificationFacade::assertSentTo([$this->candidateUser], PoolCandidateStatusChanged::class);
+    }
+
+    /**
+     * Test that use can query for notifications
+     *
+     * @return void
+     */
+    public function testCanQueryForNotifications()
```
tristan-orourke, line 334: This one, I'm not sure if it should be moved to the new test file or just be deleted. Does it test anything above and beyond the simpler tests in this file? 

Maybe it was for testing that the "enrich notification" method worked? Do we need to test that anymore?

tristan-orourke (changes requested):
Sorry, on second look, I think you're right and the specific notification should have its own test file. And some of those seem redundant now anyway. 

Besides that, everything looks good!

api/tests/Feature/NotificationTest.php
esizer: Hmm, yeah it is a little more comprehensive and strict by actually checking the fields exist and match. I guess I could move the assertions to the other query test and then delete this one?

api/tests/Feature/NotificationTest.php
esizer: Looks like it was just a different way of doing the same thing so I removed it as part of [reorganized notification tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9694/commits/179699d422207e642706fd900a0538072d03a175)

api/tests/Feature/NotificationTest.php
esizer: Oh yeah, that one seems a little too specific for this file. Moved out in [reorganized notification tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9694/commits/179699d422207e642706fd900a0538072d03a175)

api/graphql/schema.graphql
```diff
@@ -816,6 +811,9 @@ type Query {
   teams: [Team]! @all @canModel(ability: "viewAny")
   roles: [Role]! @all @canModel(ability: "viewAny")
 
+  notifications(where: NotificationFilterInput): [Notification!]
+    @paginate(defaultCount: 10, scopes: ["authorizedToView"])
+
```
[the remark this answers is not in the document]
esizer: hah yeah, I had the same thought while working on this 😅

api/app/GraphQL/Mutations/MarkNotificationAsUnread.php
```diff
@@ -0,0 +1,29 @@
...
+
+namespace App\GraphQL\Mutations;
+
+use App\Models\User;
+use Illuminate\Support\Facades\Auth;
+
+final class MarkNotificationAsUnread
+{
+    /**
+     * Mark a specific notification as read
```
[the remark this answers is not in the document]
esizer: good catch! 😆 Fixed in [fix typo in doc](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9694/commits/d4e1228757ad26d8f801fdc0d514f9669a4b8978)

api/graphql/schema.graphql
```diff
@@ -1570,6 +1568,17 @@ input UpdateApplicationInput {
     @rules(apply: ["App\\Rules\\HasEssentialSkills"])
 }
 
+input DateRange {
+  from: DateTime!
+  to: DateTime!
+}
```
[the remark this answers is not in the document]
esizer: oh yeah! That is an great idea. Added in [add date range validation](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9694/commits/b605b908998bdf463f4be25033756c987d9d7ad7)

api/graphql/schema.graphql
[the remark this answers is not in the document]
tristan-orourke: ahhhh... I see your point, but I'm not sure how to avoid that :/

api/tests/Feature/NotificationTest.php
tristan-orourke: I like it!

api/graphql/schema.graphql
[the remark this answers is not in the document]
esizer: Its annoying but there are ways to make it a little nicer with fragments and switches. I've had ones with like 30-40 different sub types before and it was....manageable 😛

PR #9703, opened 2024-03-09
[Bug] Handle missing `where` in `CountPoolCandidatesByPool`

### LINKED ISSUE #9674: 🐛 Prod Logs - Undefined array key "where"
## 🐛 Bug

An exception was logged in prod:
`Undefined array key "where" {"exception":"[object] (ErrorException(code: 0): Undefined array key \"where\" at /home/site/wwwroot/api/app/GraphQL/Queries/CountPoolCandidatesByPool.php:16)`

## 🦋 Expected Behaviour

The array element should be safely retrieved and no exception should be thrown if it is missing.

## 📋 Steps to Reproduce

(unknown)

## 📸 Screenshot

image

## ✅ Acceptance Criteria

- [ ] No exception should be thrown if the query is run without a "where" arg
- [ ] Query should still run without predicates

vd1992:
🤖 Resolves #9674 

## 👋 Introduction

Define the variable `filters` such that it handles the input not being present. 

## 🧪 Testing

1. Test query below
2. Tests succeed

```
query count {
  countPoolCandidatesByPool(where: null) {
    pool {
      id
      name {
        en
      }
    }
    candidateCount
  }
}
```

## 📸 Screenshot

image

### DIFF
--- a/api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
+++ b/api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
@@ -13,7 +13,7 @@ final class CountPoolCandidatesByPool
      */
     public function __invoke($_, array $args)
     {
-        $filters = $args['where'];
+        $filters = ! empty($args['where']) ? $args['where'] : [];
 
         // query counts pool candidate rows, so start on that model
         $queryBuilder = PoolCandidate::query();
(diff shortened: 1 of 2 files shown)


petertgiles (approved):
Very nice!  The replication isn't quite right, I think.  It has to be:
```graphql
query count {
  countPoolCandidatesByPool {
    pool {
      id
      name {
        en
      }
    }
    candidateCount
  }
}
```
without the parameter. It looks like that causes a different exception which your fix also corrects.  :clap:

PR #9715, opened 2024-03-11
[Debt] Integrate text search into builder

### LINKED ISSUE #9071: ✨ Integrate full text search into query builder
## ✨ Feature

Currently the full text search in the user model executes its own query first then integrates the IDs into the query builder separately.  This is working but has a couple of drawbacks.  The first is that the performance lessened since we're requesting the data with two operations serially with two separate queries.  The second is that any existing query scopes are not applied to the text search query and potentially many of the resulting IDs have to be thrown out.  The third is that the database cannot cache or optimize the query because the parameter count and values change every time.

Let's try to integrate the full text search feature directly into the query builder to optimize performance and flexibility.

## 📸 Screenshot

image

## ✅ Acceptance Criteria

- [ ] scopeGeneralSearch adds to the query builder without running its own query
- [ ] Index updating and results continue to function currectly

petertgiles:
🤖 Resolves #9071 

## 👋 Introduction

This branch removes the Scout search, which is not really made to be integrated into query builder, and uses regular query builder commands for text search.  I tried to essentially replicate the query that Scout was building.  Scout is still being used to maintain the indexes, though.

## 🕵️ Details

Does it perform better?  A little better, maybe.  I seeded 10,000 users into my database and the updated query returns in about 800ms while the old query returns in about 1,000.

## 🧪 Testing

- [ ] Users table general search works as before
- [ ] Pool candidates table general search works as before
- [ ] Searches perform at least as well as before

### DIFF
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -723,12 +723,17 @@ class User extends Model implements Authenticatable, LaratrustUser
             $combinedSearchTerm = implode('&', array_map('trim', $searchTerms));
 
             $query
+                // attach the tsquery to every row to use for filtering
                 ->crossJoin(DB::raw('websearch_to_tsquery(coalesce(?, get_current_ts_config()), ?)'.' AS "tsquery"'))
+                // add to_tsquery function parameters
                 ->setBindings(['english', $combinedSearchTerm])
+                // filter rows against the tsquery
                 ->whereColumn('searchable', '@@', 'tsquery')
+                // add the calculated rank column to allow for ordering by text search rank
                 ->addSelect(DB::raw('ts_rank(searchable, "tsquery") AS rank'))
+                // now that we have added a column, query builder no longer will add a "*" to the select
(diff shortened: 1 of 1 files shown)


brindasasi (approved):
This works just fine as full text search🙌
Tried different terms ..seems to be working well.

vd1992:
Given the performance remarks, I wanted to see how things shook out in my environment. 

Seeded 1500 users in `UserRandomSeeder`
Tested pool candidates table at 100 candidates per page view

Main
Search term "ROI", total results 11
At 0.9 - 1.1 seconds
Search term "Qui" total results 788
At 3.2 - 3.5 seconds

Branch
Search term "ROI", total results 11
At 0.85 - 0.95 seconds
Search term "Qui" total results 788
At 2.95 - 3.10 seconds

Conclusion: A small, but definitely present bump in speediness. To throw out a value, maybe ~10% improvement
Nice!

PR #9720, opened 2024-03-12
[Debt] Refactors `Repeater` component

### LINKED ISSUE #8491: ♻️ Form Repeater refactor

mnigh:
🤖 Resolves #8491.

## 👋 Introduction

This PR refactors the `Repeater` component.

## 🧪 Testing

1. `npm run storybook:design-system`
2. Verify `Repeater` stories work as expected
3. Verify app works as before


packages/forms/src/components/Repeater/Repeater.tsx
```diff
@@ -185,30 +142,12 @@ const Fieldset = ({
     );
   }
```
esizer, line 143: Since this is never re-assigned we could clean it up by moving it directly to where it is rendered how it is:

```tsx
{onEdit && (
  <ActionButton {....props} />
)}
```

Same with the remove button.

esizer (changes requested):
There is likely more we could do but this is a sparsely used component and we could always push more refactoring back. 

As it stands, this is a good start and makes it more manageable for the future in the unlikely event we need to make changes to it 🤷‍♀️ 

--- 

Nice to have: `onRemove` sends focus to the previous item if it exists. Right now it sits between two which is fine but could be better.

esizer (approved):
Cool 😎

PR #9727, opened 2024-03-13
[Debt] Consolidate skill level message getters

### LINKED ISSUE #9680: ♻️ Consolidate skill level message getters
## ♻️ Debt/Refactor

Consolidate all of the message getter functions for skill levels into a two helper functions.

## 🕵️ Details

We have different messages for skill level depending on the type of skill. Because of this we have a lot different places were we use conditional logic to use a different message getter based on that. It would be nice to have a help function to get either one so we don't repeat ourselves.

## 🙋‍♀️ Proposed Solution

Crere something like:

```tsx
function getSkillLevelName(skillLevel: SkillLevel, type: SkillType): string;

function getSkillLevelDefition(skillLevel: SkillLevel, type: SkillType): string;

function getSkillLevelMessages(skillLevel: SkillLevel: type: SkillType): { name: string, defintion: string };
```

## ✅ Acceptance Criteria

- [ ] Helper functions for skill level message getters
- [ ] Replace existing conditional logic with new helpers

JamesHuf:
🤖 Resolves #9680

## 👋 Introduction

Consolidates the getter functions for skill levels somewhat.

## 🧪 Testing

1. Visit various pages involving skills (ex. skill library) and ensure the names and descriptions of the different levels still show as before.


esizer (approved):
Love it 👏

packages/i18n/src/messages/localizedConstants.tsx
```diff
@@ -1943,14 +1916,53 @@ const behaviouralSkillLevelDefinitions = defineMessages({
...
+    `Invalid behavioural skill level '${skillLevel}'`,
+  );
+};
+
+export const getSkillLevelDefinition = (
+  skillLevel: string | SkillLevel,
+  skillCategory: SkillCategory,
+): MessageDescriptor => {
+  if (skillCategory === SkillCategory.Technical) {
+    getOrThrowError(
```
esizer, line 1943: We will need a return here or we will always get back behavioral ones.

```suggestion
    return getOrThrowError(
```

esizer (changes requested):
One last thing!

esizer (approved):

PR #9755, opened 2024-03-14
[Feature] Notifications page

### LINKED ISSUE #9558: ✨ All notifications page
## ✨ Feature

All notifications page created

## 🕵️ Details

- Update settings links to notification settings on account management page
- Ignore pinning and filtering for now
- "..." context menu allows to Mark as read/unread and Delete (delete requires confirmation)
- Clicking a notification sets it to "read" and navigates to the relevant page (depends on notification type)
- Mark all as read fires mutation for all user's unread notifications to be marked as read
- User can toggle between All and Unread notifications
- Loads all notifications n at a time (n is some hard-coded number). "Load more" loads the next n, like a manually triggered infinite scroll. Repeats until none left to paginate. Does not hide previously loaded notifications. 
  - Let's say load 10 at a time for now
- Delete opens a confirmation dialog:
image

## 🎨 Design File

https://www.figma.com/file/bIvDLraUxqGxDBuerPJlNI/Candidates---Notifications?type=design&node-id=813-380&mode=design

@esizer left a note for implementation: https://www.figma.com/file/bIvDLraUxqGxDBuerPJlNI?node-id=876:32811&mode=design#721297727

Component library designs, including hover and focus states: https://www.figma.com/file/guHeIIh8dqFVCks310Wv0G/Style-library?type=design&node-id=2184-5048&mode=design&t=Liav3sHrKuf2RNtr-0

### 🧑‍🎨 Designer

@substrae 

## 📸 Screenshot

Image

## Implementation Suggestions

Each notification from the backend doesn't necessarily come with descriptive text or a link. We should have some utility file for Notifications which can generate the frontend text and link for each type of notification. As we add more types of notifications, we can just add to this utility file.

## ✅ Acceptance Criteria

- [ ] Page created (at `/applicant/notifications`)
- [ ] On page load, loads n most recent notifications (n=10)
- [ ] Unread notifications appear different
- [ ] "Load more" button loads n more notifications and adds them to list without reloading page or hiding previously fetched
- [ ] "Mark all as read" button is functional
- [ ] Clicking a notification sets it to "read" (if unread) and links to related page. (not opening a new tab)
- [ ] Text on notification should not be formatted as link - hover state should indicate it is clickable
- [ ] Clickable area _is_ a link in html, though it will need to intercept the click action
- [ ] If "Unread" filter is active, then only unread notifications are loaded
- [ ] Clicking the "..." opens a small context menu which allows for marking the notification as read/unread (depending on current status) or Delete
- [ ] Clicking Delete notification opens a confirmation dialog
- [ ] Marking a notification as read/unread or deleting it should immediately be reflected in the UI
- [ ] "..." menu has an aria label: `Manage this [TYPE/SPECIFIC CONTEXT] notification` (TBD)

## 🛑 Blockers

```[tasklist]
### Blocked By
- [ ] #9554 
- [ ] https://github.com/GCTC-NTGC/gc-digital-talent/issues/9557
```

esizer:
🤖 Resolves #9558 

## 👋 Introduction

Adds a page to view notifications for the logged in user.

## 🕵️ Details

I opted to use links for the "load more" and "unread/all" actions so we can sync with URL. This means you can navigate to the following URLs to get the list:

- `/applicant/notifications?unread`
- `/applicant/notifications?page=2`

Also, this means the back button should restore state if you click a notification and then the back button.

Some things to consider when reviewing:

- Design does not match exactly because the files are ahead of where the code is at right now. More specifically:
    - Header/Hero
    - Sidebar width
    - Buttons in dropdown are underlined (built into button) but not in design
    - Missing the pinning and filters (being done in another ticket not yet created)

## 🧪 Testing

In order to set this up (seed notifications) I found it easy to login as admin `admin@test.com` and then go to that users user information page, add them to a pool and just change the status over and over until you feel like you have enough 🤷‍♀️ 

1. Build `npm run dev`
2. Login
3. Navigatge to `/en/applicant/notifications`
4. Confirm notifications appear and match [the mockup](https://www.figma.com/file/bIvDLraUxqGxDBuerPJlNI/Candidates---Notifications?type=design&node-id=813-380&mode=design) 
5. Test mutations
    - Marking as read
    - Marking as unread
    - Deleting
    - Marking all as read
6. Click on a notification
7. Confirm it takes you to the application
8. Confirm that notification is now marked as read
9. Confirm load more appends to the list
10. Confirm load more button disappears when all notifications are loaded
11. Confirm switching the tab from "All" to "Unread" updates the list as expected

## 📸 Screenshot

localhost_8000_en_applicant_notifications (1)
localhost_8000_en_applicant_notifications


apps/web/src/components/Router.tsx
```diff
@@ -916,6 +924,17 @@ const createRoute = (
                     </RequireAuth>
                   ),
                 },
+                {
```
petertgiles, line 927: Should this page be behind the feature flag?

apps/web/src/components/NotificationList/NotificationListPage.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+const NotificationListPage = ({
+  page,
+  onlyUnread,
+  isLastPage,
+}: NotificationPageProps) => {
+  const intl = useIntl();
+  const [searchParams] = useSearchParams();
+  searchParams.set("page", `${page + 1}`);
+  const [{ data, fetching, error }] = useQuery({
```
petertgiles, line 46: Has there been any discussion about whether this page should be "live"?  Like, should this be polling for new notifications in the background?

apps/web/src/components/NotificationList/NotificationListPage.tsx
```diff
@@ -0,0 +1,121 @@
...
+  });
+
+  const notifications = unpackMaybes(data?.notifications?.data);
+
+  if (notifications.length === 0 && page === 1 && !fetching) {
+    return (
+      <Well>
+        <p data-h2-font-weight="base(700)" data-h2-margin-bottom="base(x1)">
+          {intl.formatMessage({
+            defaultMessage: "There aren't any notifications here yet.",
```
petertgiles, line 64: @substrae All the copy here about the future strikes me as strange.  There aren't any notifications here probably because I just marked them as read or deleted them.  Not because I haven't received any yet.

image

Hopefully the amount of time for which a user will have never received a notification will be pretty short, right?

apps/web/src/components/NotificationList/RemoveDialog.tsx
petertgiles, line 1: Looks like there should be a date in the dialog.
image

apps/web/src/components/NotificationList/NotificationListPage.tsx
```diff
@@ -0,0 +1,121 @@
...
+  page,
+  onlyUnread,
+  isLastPage,
+}: NotificationPageProps) => {
+  const intl = useIntl();
+  const [searchParams] = useSearchParams();
+  searchParams.set("page", `${page + 1}`);
+  const [{ data, fetching, error }] = useQuery({
+    query: Notifications_Query,
+    variables: {
```
petertgiles, line 48: Should there be a sorting command here?  To ensure pages are read out coherently?

petertgiles (changes requested):
Oh yeah, this is cool.  :sunglasses:  I flagged a few things to check on.

apps/web/src/components/NotificationList/NotificationListPage.tsx
esizer: I think that is a good idea! Can we do that right now or is there something we need first to support it?

petertgiles: If we want to do simple polling I guess you could just add a repeating timeout and rerun the query.  If we want to do graphql subscriptions or websockets or something we'll probably need to add some plumbing.  This is more important for the nav bar notification, I think.  Worth bringing up in standup.

apps/web/src/components/NotificationList/NotificationListPage.tsx
petertgiles: Oh, I missed that.  No, this is good, at least for now.

apps/web/src/components/NotificationList/NotificationListPage.tsx
esizer: Yeah, thanks. I likely will. I'm trying to think how we could do this with `graphql` and pages 🤔 I guess we could have a general query that polls for the latest items since the last run and append them on and then set the offset for pagination to however many we loaded from that general polled version? 🤷‍♀️

esizer: okay, it is kinda messy but I got this working in [fix markup, add polling to list](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9755/commits/89d6dcf36810d302d2103c70fee6f6f7e72bd6f7) but have it disabled for now. Not sure we want it on the main page but it is there for the dialog since I plan on using the same list component. If you want to test:

1. Change the `live` prop to `true` for the list on the page
2. Change the delay in seconds to something short
3. Login as `admin@test.com`
4. Open two tabs one on `/applicant/notifications` and the other on `/admin/users/{userId}` where `{userId}` is the ID of `admin@test.com`
5. Add the user to a pool on the admin page
6. Change their status
7. Observe the notifications page updates within the delay you set

apps/web/src/components/NotificationList/RemoveDialog.tsx
esizer: Oh! Good catch, added in [add date to remove dialog](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9755/commits/dadf9058482e2f6473138d056128885b556b418e)

apps/web/src/components/Router.tsx
esizer: Ahh yeah, that makes sense 😆 Fixed in [move page behind feature flag](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9755/commits/4c5e9f3586c59900b31fdf83a9807e3f9d9c1f3d)

apps/web/src/components/NotificationList/NotificationListPage.tsx
petertgiles: Cool, I've never used portals manually before!  I think there's some trouble with the polling logic though.  I get the refresh requests coming through in chains of two or three.  For example, I have the interval set to 10 seconds here:
image  
Also, by only querying for "createdAt"  the live updating misses items that should be removed after delete and mark-as-read.  I'm pretty skeptical of trying to live update paged data.  How can you do an incremental update with incomplete data?  Unless we can query for some sort of changelog or pub-sub I think we need to accept that paged data is going to jump around when it is updated.

apps/web/src/components/NotificationList/NotificationListPage.tsx
substrae: 🤔 won't you only ever receive notifications in a future sense though? Just because you dismissed them doesn't mean the statements there aren't true. Happy to update it if y'all have better suggestions though!

apps/web/src/components/NotificationList/NotificationListPage.tsx
esizer: Yeah, I'm not too sure what to do with this. What I have there is super hacky. `urql` expects you to be using the graphcache to get polling w/ pagination working.

I think it might be easier if we only use polling without the pagination until we have graphcache working :thinking: 

Or, get subscriptions working but that seems like alot more work.

apps/web/src/components/NotificationList/NotificationListPage.tsx
petertgiles: I'd be happy if we just dropped the word "yet" from the title.  The rest would be fine, I think.

substrae: @petertgiles works for me!

apps/web/src/components/NotificationList/NotificationListPage.tsx
esizer: @tristan-orourke Does the main notifications page need to be live updated for now? It is technically disabled right now. We could update the dialog to be live updated and not paginated to keep this simple to start 🤔 

In regards the timeout running multiple times... I can't figure that out and am wondering if it is related to `useEffect` firing twice. Though, I did just use the suggestion that was mentioned when `urql` removed polling support.

https://github.com/urql-graphql/urql/pull/1374

tristan-orourke: It seems like we're running into known issues with offset-based pagination 
https://www.sitepoint.com/paginating-real-time-data-cursor-based-pagination/#identifying-issues-in-real-time-data-pagination

tristan-orourke: @esizer did you see this urql example of [infinite pagination](https://github.com/urql-graphql/urql/tree/main/examples/with-infinite-pagination)? It doesn't require graphecache. It uses recursive components, which is neat - I don't think I've ever seen that pattern before.

It does, however, use cursor-type pagination, which Lighthouse [doesn't really support yet](https://lighthouse-php.com/6/api-reference/directives.html#pagination-type).

tristan-orourke: For the record, I don't care if live updates or "Load more" removes items which have been removed in another tab. I think needing to refresh the page to see that would not be surprising to a user.

tristan-orourke: I'm more concerned about new notifications appearing properly, and not rendering duplicates because paging offsets have been messed up.

esizer: Yeah, that was the reason I did not do that. We need cursor based pagination 😢 Though it is nice to know I was wrong about that style requiring graphcache! In their docs they only show it with graphcache 😅

tristan-orourke: > Does the main notifications page need to be live updated for now? It is technically disabled right now. We could update the dialog to be live updated and not paginated to keep this simple to start 🤔

That's a great suggestion, let's go with that for now!

esizer: okay, so until we can get cursor pagination to make it a smoother scenario for live notifications I have updated the component too disable pagination for the dialog but enabled it on this page while keeping it static.

[allow unpaginated notification list](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9755/commits/d29cf8f7e01142643033a3708c5b4a4cb4281f15)

PR #9770, opened 2024-03-17
[Debt] Move admin dashboard to index route

### LINKED ISSUE #9604: ♻️ Remove intermediary main admin page
## ♻️ Debt/Refactor

It would be nice to remove this intermediary page `/admin` that just shows a link to `/login-info` and uses a one-off custom component.

## 🕵️ Details

The main `/admin` URL doesn't seem that necessary anymore, all it does is link you to login and the only people landing here are the ones with  a direct link since we never link to it unless you are already logged in. As well, this is the only place that uses the `CardLink` component so we could get rid of some unnecessary code and redirects.

## 🙋‍♀️ Proposed Solution

Move everything from `/admin/dashboard` up a level to `/admin` ensuring to keep the `RequireAuth`. This would remove an unnecessary redirect and any anonymous user navigating directly to `/admin` should get redirected to the login with `RequireAuth`. 

Also, you can remove the `CardLink` component since it will be orphaned after this change.

## ✅ Acceptance Criteria

- [ ] Removed `/admin/dashboard` path
- [ ] Admin dashboard page component moved to `/admin`
- [ ] Route is wrapped in `RequireAuth` in the `Router`
- [ ] `CardLink` removed
- [ ] Unauthenticated and unauthorized users cannot access `/admin`

esizer:
🤖 Resolves #9604 

## 👋 Introduction

Moves the admin dashboard (`/admin/dashboard`) to the index route (`/admin`).

## 🧪 Testing

1. Build `npm run dev`
2. Ensure you are not signed in
3. Navigate directly to `/admin`
4. Confirm you are redirected to `/login-info`
5. Login as an applicant
6. Confirm an unauthorized message
7. Logout
8. Repeat steps 3-5 but with admin `admin@test.com`
9. Confirm you are sent to the dashboard but at the path `/admin`
10. Confirm links referencing `paths.adminDashboard()` now go to `/admin`


petertgiles (approved):
Thanks for merging in the fix.  Looks great!

PR #9781, opened 2024-03-18
[Fix] Announce admin pool page navigations

### LINKED ISSUE #9728: ⌨ Process navigation not announced to ATs
## ⌨ Accessibility Issue

On page pages, we set focus to the `h1` so it is announced to ATs and orients them on context changes. However, since the `PoolLayout` never unmounts and is consistent across navigations, it is not clear to AT users when we navigate that section.

### WCAG 2.1 Criteria

- [4.1.3: Status Messages (Level AA)](https://www.w3.org/WAI/WCAG21/Understanding/status-messages.html)
- [Changes of Context - Single Page Applications](https://stevenmouret.github.io/web-accessibility-guidelines/accessibility/scripts/changes-of-context.html#:~:text=accessible%20name.-,Single%20page%20application,-In%20a%20%E2%80%9CSingle)

## 🦋 Expected Behaviour

Screen reader users should be notified of a page/content change.
`
## 📋 Steps to Reproduce

1. Load up any screen reader you have access to (this affects all)
3. Navigate to a specific process
4. Navigate using the tabs in the hero
5. Notice there is no announcement of the page change

## 📸 Screenshot

Screenshot 2024-03-13 080814

## 🙋‍♀️ Proposed Solution

~This will need some help from designers. I think we need some sub title that we can focus in the event that the actual title does not change. I think we could use some attribute like `data-a11y-status` on that subtitle and use it in the hero's focus to move it there instead of the `h1` if it exists.~

Another option is maybe trying to use out `useAnnouncer` hook and attempt to do some announcements from the layout when the current page changes so we don't need to change the design at all?

## ✅ Acceptance Criteria

- [ ] When navigating the process admin pages all navigations (changes in context) are announced to ATs

esizer:
🤖 Resolves #9728 

## 👋 Introduction

Fixes an issue where pool page navigations were not being announced to screen readers.

## 🧪 Testing

1. Build `npm run dev`
2. Login as admin `admin@test.com
3. Start a screen reader
4. Navigate to a pool in the admin
5. Change tabs
6. Confirm page title is being announced after the page change

### DIFF
--- a/apps/web/src/pages/Pools/PoolLayout.tsx
+++ b/apps/web/src/pages/Pools/PoolLayout.tsx
@@ -3,7 +3,12 @@ import { useIntl } from "react-intl";
 import { Outlet } from "react-router-dom";
 import { useQuery } from "urql";
 
-import { Pending, Chip, ThrowNotFound } from "@gc-digital-talent/ui";
+import {
+  Pending,
+  Chip,
+  ThrowNotFound,
+  useAnnouncer,
+} from "@gc-digital-talent/ui";
 import { getLocalizedName } from "@gc-digital-talent/i18n";
 import { graphql, Pool } from "@gc-digital-talent/graphql";
 
@@ -25,6 +30,7 @@ interface PoolHeaderProps {
 
 const PoolHeader = ({ pool }: PoolHeaderProps) => {
   const intl = useIntl();
+  const { announce } = useAnnouncer();
 
   const pages = useAdminPoolPages(intl, pool);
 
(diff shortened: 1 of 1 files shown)


yonikid15 (approved):
Awesome the pages are getting announced. I used NVDA.

This is probably out of scope, but I'll flag it here, I noticed that the nav link for "Talent Placement" is different than the heading/pageTitle. It ends up announcing "Talent Placement visited link" when focusing on the nav link, then when you click on it announces "Candidates.. current page". I feel like that might be confusing.

esizer:
> This is probably out of scope, but I'll flag it here, I noticed that the nav link for "Talent Placement" is different than the heading/pageTitle. It ends up announcing "Talent Placement visited link" when focusing on the nav link, then when you click on it announces "Candidates.. current page". I feel like that might be confusing.

Yeah, I noticed that as well. If you look in the browser tab it also says "Pools candidates". I tihnk we just missed some spots when we updated the name of that page :thinking:

PR #9784, opened 2024-03-19
[Bug] Screening dialog handle null education requirement

### LINKED ISSUE #9730: 🐛 Screening decision dialog does not hide skill accordion if info does not exist
## 🐛 Bug

When screening a user, in the event that the skill they are assessing does not contain a description, the accordion appears but is empty.

## 🦋 Expected Behaviour

When a skill does not have a description, we should show a null message or hide this entirely. 

## 📋 Steps to Reproduce

1. Find or create a skill with no description
2. Create an assessment for that skill
3. Create an applicant to assess for that step/skill
4. Open the screening dialog for that skill
5. Confirm the well and accordion appear empty

## 📸 Screenshot

Screenshot 2024-03-13 084303

## 🙋‍♀️ Proposed Solution

1. Show a null state instead of hiding the contents
2. Hide entire section

## ✅ Acceptance Criteria

- [ ] Empty skill in screening dialog has a null message or
- [ ] Is hidden entirely

vd1992:
🤖 Resolves #9730

## 👋 Introduction

The referenced is not skill-related but actually the education form of the dialog, so falsy handling added.
Tweaked seeding a bit as it was throwing me for a loop a first. 

## 🧪 Testing

1. Normal assessment view for education unchanged
2. Empty education view has null handling 

## 📸 Screenshot

Normal 

image

Falsy

image

Easy way to add applications with empty education

image

French for reference to compare against image in issue

image

### DIFF
--- a/api/database/factories/PoolCandidateFactory.php
+++ b/api/database/factories/PoolCandidateFactory.php
@@ -111,8 +111,9 @@ class PoolCandidateFactory extends Factory
                 ]);
             }
 
-            if ($poolCandidate->education_requirement_option ===
-            EducationRequirementOption::EDUCATION->name || EducationRequirementOption::PROFESSIONAL_DESIGNATION->name) {
+            // attach either a work or education experience to a pool candidate to meet minimum criteria
+            if ($poolCandidate->education_requirement_option === EducationRequirementOption::EDUCATION->name ||
+            $poolCandidate->education_requirement_option === EducationRequirementOption::PROFESSIONAL_DESIGNATION->name) {
                 //Ensure user has at least one education experience
                 $experience = EducationExperience::factory()->create([
                     'user_id' => $poolCandidate->user_id,
--- a/api/database/factories/PoolCandidateFactory.php
+++ b/api/database/factories/PoolCandidateFactory.php
@@ -49,9 +49,8 @@ class PoolCandidateFactory extends Factory
             'submitted_steps' => array_slice(
                 array_column(ApplicationStep::cases(), 'name'),
                 0,
-                $this->faker->numberBetween(0, count(ApplicationStep::cases()) - 1)
+                $this->faker->numberBetween(0, count(ApplicationStep::cases()) - 2)
             ),
-            'education_requirement_option' => $this->faker->randomElement(EducationRequirementOption::cases())->name,
             'is_bookmarked' => $this->faker->boolean(10),
         ];
     }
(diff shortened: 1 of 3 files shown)


api/database/factories/PoolCandidateFactory.php
```diff
@@ -49,9 +49,8 @@ public function definition()
             'submitted_steps' => array_slice(
                 array_column(ApplicationStep::cases(), 'name'),
                 0,
-                $this->faker->numberBetween(0, count(ApplicationStep::cases()) - 1)
+                $this->faker->numberBetween(0, count(ApplicationStep::cases()) - 2)
```
vd1992, line 52: Incomplete applications never have the full set, if the application is complete (see below) always have the full set

api/database/factories/PoolCandidateFactory.php
```diff
@@ -98,7 +98,22 @@ public function configure()
...
+            } else {
+                $requirementOption = $this->faker->boolean() ? EducationRequirementOption::APPLIED_WORK->name : EducationRequirementOption::EDUCATION->name;
+                $poolCandidate->update([
+                    'education_requirement_option' => $requirementOption,
+                ]);
+            }
+
+            // attach either a work or education experience to a pool candidate to meet minimum criteria
+            if ($poolCandidate->education_requirement_option === EducationRequirementOption::EDUCATION->name ||
+            $poolCandidate->education_requirement_option === EducationRequirementOption::PROFESSIONAL_DESIGNATION->name) {
```
vd1992, line 116: Three education options exist, `Professional_Designation` is for EX pool's only and so should not be seeded if that isn't the case. It creates copy bugs otherwise that'll confuse you 😛

api/database/seeders/UserRandomSeeder.php
```diff
@@ -136,26 +133,6 @@ public function run()
...
             ->create();
-
-        // attach either a work or education experience to a pool candidate to meet minimum criteria
-        PoolCandidate::all()->load('user')->each(function ($poolCandidate) {
-            $educationRequirementOption = $poolCandidate->education_requirement_option;
-            $user = $poolCandidate->user;
-
-            if ($educationRequirementOption === EducationRequirementOption::EDUCATION->name) {
-                //Ensure user has at least one education experience
-                $experience = EducationExperience::factory()->create([
```
vd1992, line 147: Repeats the logic in 

image

PR #9785, opened 2024-03-19
[Feature] Notification settings backend

### LINKED ISSUE #9555: ✨ Backend for user notification settings
## ✨ Feature

Controlling the notifications a user receives will be changeable, requiring storing this information

QUESTION? Is it worth storing and rendering in options a setting you cannot change?
- Tristan: I think it makes sense to store "always on" preferences in the same way, because we can then check them in the same way when deciding to send a notification or not. We will show the "System messages" option, which cannot be turned off, as a transparency effort.

## 🎨 Design File

https://www.figma.com/file/AQbFeF2lm4gZZJOwxOJTN2/Candidates---Account-settings?type=design&node-id=1-3&mode=design

## 📸 Screenshot

Image

QUESTION: what enums do we start with? 
- Start with enums representing System Messages, Application Updates, and Job Alerts
- QUESTION: How do we handle default preferences?
  - By default, users should accept all notifications, and have to turn them off.
  - We can store preferences as an array of types to _ignore_. Then, if we create new types, they will always have to be added manually. And if we want to disallow ignoring a type (ie system notifications) then we could control that with validation on the mutation.

Ensure all notifcations have "family" field.
Addition of `user.ignored_email_notifications` and `user.ignored_in_app_notifications` that are arrays of the notification families

## ✅ Acceptance Criteria

- [ ] Email preferences can be retrieved and mutated via api
- [ ] Testing of the mutation
- [ ] seeding of new users with notification preferences

JamesHuf:
🤖 Resolves #9555

## 👋 Introduction

Adds the `ignored_email_notifications` and `ignored_in_app_notifications` fields to user and a mutation for updating them.

## 🧪 Testing

1. Run the migration and seeders.
2. Query users and confirm you can view their ignored notifications.
3. Confirm the mutation works as expected.


api/graphql/schema.graphql
```diff
@@ -1872,6 +1877,13 @@ type Mutation {
   markNotificationAsUnread(id: UUID!): Notification @guard
   deleteNotification(id: UUID!): Notification @guard
   markAllNotificationsAsRead: [Notification] @guard
+  updateIgnoredNotifications(
+    id: UUID!
+    ignoredEmailNotifications: [NotificationFamily]
+      @rename(attribute: "ignored_email_notifications")
+    ignoredInAppNotifications: [NotificationFamily]
+      @rename(attribute: "ignored_in_app_notifications")
+  ): User @update @guard @validator
```
esizer, line 1886: It seems as though any user can update another users ignored notifications here. Should we remove the ID and make it so running this mutation always affects the authenticated user or at least add a permission like `update-own-ignoredNotifcations`? 

Dropping a mention to @tristan-orourke to see his thoughts on best approach.

tristan-orourke: Agreed - I think the query to get notifications only works with the logged in user, right? The mutations to modify them should work the same.

JamesHuf: Updated with [53fd40f](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9785/commits/53fd40f88865224c7800e4ee0e5720f954959ab5) to just update the requesting users data.

esizer (approved):
Cool! :sunglasses: 

I wish graphqil interface for array fields worked. For a second I thought I could only update one at a time :laughing:

esizer:
oops I forgot to switch branches! I was like "I swore i commit this file before" :laughing:

PR #9801, opened 2024-03-20
[Bug] Filter out already disqualified candidates in processes

### LINKED ISSUE #9789: 🐛 "Disqualified" candidates show up on tracker under "To assess"
## 🐛 Bug

Julie is running a process that had been started previously to RoD deploy, approximately half of those candidates had already been screened out. When looking at the list of candidates in the "Talent placement" table, it is easy to see which ones have been disqualified. When looking at the candidates through the Screening and assessment tab, ALL candidates show up as "To assess"

## 🦋 Expected Behaviour

Candidates that have been "Disqualified" should NOT show up as "To assess" on the Screening and assessment tab, regardless of their status on each step.

## 📋 Steps to Reproduce

1. Run the seeder on your local 
2. Open up any process that has candidates with "Screened out" status already 
3. Observe the Assessment Tracker 

https://github.com/GCTC-NTGC/gc-digital-talent/assets/89482910/53cfeded-fd97-4ebf-a189-d07fe6912f01

**Root cause** 
This can happen if the candidate has been disqualified before the ROD roll out. 
Reason is that final decision is empty on this candidate.
Once I went into the same candidate assessment screen and record the final decision as disqualified , the tracker show up correctly. 
 We have broken backward compatibility in here. Hence there needs to be some migration to update the existing assessments.

## 🙋‍♀️ Proposed Solution

If no assessment results
and a final decision
then don't show in tracker

## ✅ Acceptance Criteria

- [ ] Candidates that have been "Disqualified" should NOT show up as "To assess" on the Screening and assessment tab, regardless of their status on each step.

vd1992:
🤖 Resolves #9789

## 👋 Introduction

Filter out candidates that were already disqualified prior to RoD
I think I got the gist of what the goal was? 😅 

## 🧪 Testing

1. Add a candidate to a process, disqualify them, then ensure they are missing from Screening and assessment
2. Must have no assessment results 

## 📸 Screenshot

Pool candidate

image

Present

image

Missing

image

### DIFF
--- a/apps/web/src/components/AssessmentStepTracker/utils.ts
+++ b/apps/web/src/components/AssessmentStepTracker/utils.ts
@@ -23,6 +23,7 @@ import {
   determineCandidateStatusPerStep,
   determineCurrentStepPerCandidate,
   getDecisionCountForEachStep,
+  isDisqualifiedStatus,
 } from "~/utils/poolCandidate";
 
 export type CandidateAssessmentResult = {
@@ -268,4 +269,19 @@ export const filterResults = (
       results: filteredResults,
     };
   });
+};
+
+// filter out candidates who are disqualified AND have an empty assessment results collection
+export const filterAlreadyDisqualified = (
+  candidates: PoolCandidate[],
+): PoolCandidate[] => {
+  const filteredResult = candidates.filter(
+    (candidate) =>
+      !(
+        isDisqualifiedStatus(candidate.status) &&
+        (candidate.assessmentResults ? candidate.assessmentResults : [])
+          .length === 0
+      ),
+  );
+  return filteredResult;
 };
--- a/apps/web/src/components/AssessmentStepTracker/utils.ts
+++ b/apps/web/src/components/AssessmentStepTracker/utils.ts
@@ -12,7 +12,6 @@ import {
   AssessmentDecision,
   Maybe,
   AssessmentStep,
-  PoolCandidateStatus,
 } from "@gc-digital-talent/graphql";
 import { notEmpty } from "@gc-digital-talent/helpers";
 import { commonMessages } from "@gc-digital-talent/i18n";
@@ -24,6 +23,8 @@ import {
   determineCandidateStatusPerStep,
   determineCurrentStepPerCandidate,
   getDecisionCountForEachStep,
+  isDisqualifiedStatus,
+  isRemovedStatus,
 } from "~/utils/poolCandidate";
 
 export type CandidateAssessmentResult = {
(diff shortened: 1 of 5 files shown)


brindasasi:
The mentioned A/C works well. Disqualified candidates are not shown in the tracker. 
What about the below status ? Are we gonna still show this? 
Screenshot 2024-03-21 at 4 04 19 PM

brindasasi (changes requested):
just confirmed with Jerbo. Looks like we gotta filter out these as well 
            SCREENED_OUT_ASSESSMENT
            SCREENED_OUT_NOT_INTERESTED
            SCREENED_OUT_NOT_RESPONSIVE

vd1992:
> just confirmed with Jerbo. Looks like we gotta filter out these as well SCREENED_OUT_ASSESSMENT SCREENED_OUT_NOT_INTERESTED SCREENED_OUT_NOT_RESPONSIVE

Sure, done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/9801/commits/814212bbec0fef1a4901636b7f0e7604cff98f46
The other function only considered the first two statuses as "disqualified"

brindasasi (approved):
LGTM! 🥳

apps/web/src/components/AssessmentStepTracker/utils.ts
```diff
@@ -269,3 +270,30 @@ export const filterResults = (
...
 };
+
+const consideredDisqualifiedStatuses = [
+  PoolCandidateStatus.ScreenedOutApplication,
+  PoolCandidateStatus.ScreenedOutAssessment,
+  PoolCandidateStatus.ScreenedOutNotInterested,
+  PoolCandidateStatus.ScreenedOutNotResponsive,
+];
+
+const isConsideredDisqualifiedStatus = (
```
tristan-orourke, line 281: Instead of this, I recommend using the functions that already exist in utils/poolCandidate.ts. Basically, replace this is with `isDisqualifiedStatus(candidate.status) || isRemovedStatus(candidate.status)`

tristan-orourke (changes requested):
Just one recommendation to use an existing util function to make sure we don't miss other "removed" statuses

apps/web/src/components/AssessmentStepTracker/utils.ts
vd1992: That is a better idea, instead of doing what I did in https://github.com/GCTC-NTGC/gc-digital-talent/pull/9801/commits/814212bbec0fef1a4901636b7f0e7604cff98f46, I should have done that. Done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/9801/commits/f1cb5046b19f8aa222e451f4e8bf9d679f4ff71d

tristan-orourke (approved):

PR #9864, opened 2024-03-21
[Feature] Add better email error

### LINKED ISSUE #9733: 🐛 Email validation gap on support page
## 🐛 Bug

It is possible to enter an email address that will pass frontend validation but fail backend validation.  Unfortunately, the error message is very vague when this happens.

## 🦋 Expected Behaviour

Option 1 is just improve the error message so that the user knows what went wrong and can try to correct it.  We can't control Laravel's or different browser's email address validation.

Option 2 is that we could consider removing all validation on the email field.  It could be argued that the support form is too important to allow to fail or that users may want to intentionally obfuscate their email address in a way that a human might be able to interpret.

## 📋 Steps to Reproduce

The example I found in the server logs was `validemail*temp*@gmail.com` in Chrome, as shown in the screenshot.

## 📸 Screenshot

image

## ✅ Acceptance Criteria

- [ ] Improve backend validation error toast
- [ ] Respond with a 400 error response, not 500, if that's what we get from api

petertgiles:
🤖 Resolves #9733

## 👋 Introduction

This branch adds a better error to the support page when an "invalid email" code is being returned by Freshdesk.  I ran the suggested error message past Marc and got a "looks good".

## 🧪 Testing

1. Set up your environment for Freshdesk (either set up a junk account or copy the values from prod).

- [ ] An invalid email address triggers the new error message.
- [ ] Any other error (like bad API key) triggers the existing default error message.
- [ ] A successful ticket submission triggers the existing success toast.

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/lang/fr.json
+++ b/apps/web/src/lang/fr.json
@@ -2668,7 +2668,7 @@
     "description": "Note for when an assessment was unsuccessful but put on hold"
   },
   "DOn3Hm": {
-    "defaultMessage": "Adresse courriel non valide. Essayez de nouveau ou envoyez un courriel, o <anchorTag>{emailAddress}</anchorTag>.",
+    "defaultMessage": "Adresse courriel non valide. Essayez de nouveau ou envoyez un courriel à <anchorTag>{emailAddress}</anchorTag>.",
     "description": "Support form toast message error"
   },
   "DPfJ30": {
--- a/apps/web/src/lang/fr.json
+++ b/apps/web/src/lang/fr.json
@@ -2667,6 +2667,10 @@
     "defaultMessage": "Non démontré suffisamment dans le cadre de la présente méthode d’évaluation – faire passer à de plus amples mises à l’essai. (Ne s’applique qu’aux occasions où les points sont cumulatifs à l’échelle des méthodes d’évaluation. Ne s’applique pas dans les cas où les candidats doivent répondre à certains critères essentiels à chaque étape).",
     "description": "Note for when an assessment was unsuccessful but put on hold"
   },
+  "DOn3Hm": {
+    "defaultMessage": "Adresse courriel non valide. Essayez de nouveau ou envoyez un courriel, o <anchorTag>{emailAddress}</anchorTag>.",
+    "description": "Support form toast message error"
+  },
   "DPfJ30": {
     "defaultMessage": "La compétence <strong>sera associée aux expériences auxquelles elle était précédemment associée.</strong>",
     "description": "Notice that re-adding a skill will re-add it to any linked experiences"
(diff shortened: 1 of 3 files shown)


apps/web/src/lang/fr.json
```diff
@@ -2667,6 +2667,10 @@
     "defaultMessage": "Non démontré suffisamment dans le cadre de la présente méthode d’évaluation – faire passer à de plus amples mises à l’essai. (Ne s’applique qu’aux occasions où les points sont cumulatifs à l’échelle des méthodes d’évaluation. Ne s’applique pas dans les cas où les candidats doivent répondre à certains critères essentiels à chaque étape).",
     "description": "Note for when an assessment was unsuccessful but put on hold"
   },
+  "DOn3Hm": {
+    "defaultMessage": "Adresse courriel non valide. Essayez de nouveau ou envoyez un courriel, o <anchorTag>{emailAddress}</anchorTag>.",
```
mnigh, line 2671: Based on the English equivalent, I think this is a typo.

```suggestion
    "defaultMessage": "Adresse courriel non valide. Essayez de nouveau ou envoyez un courriel à <anchorTag>{emailAddress}</anchorTag>.",
```

petertgiles: Thanks!  Fix checks out on deepl.

apps/web/src/pages/SupportPage/components/SupportForm/SupportForm.tsx
```diff
@@ -309,13 +276,52 @@ const SupportFormApi = () => {
...
+    const errorCode = `${response.status} - ${response.statusText}`;
+    logger.error(`Failed to submit ticket: ${JSON.stringify(responseBody)}`);
+
+    // default error message if we don't recognize the error
+    let errorMessage = defineMessage({
+      defaultMessage:
+        "Sorry, something went wrong. Please email <anchorTag>{emailAddress}</anchorTag> and mention this error code: {errorCode}.",
+      id: "rNVDaA",
+      description: "Support form toast message error",
+    });
```
yonikid15, line 300: I think we're moving to defining messages like this outside the component (just above usually) so they aren't getting redeclared.

yonikid15 (approved):
Nice work looks good!

Invalid email:
image

Bad api key:
image

apps/web/src/pages/SupportPage/components/SupportForm/SupportForm.tsx
petertgiles: Good idea, thanks!
[pull messages out of component](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9864/commits/d136f73078ddec9942ca324e46b6a7afa55ed501)

PR #9841, opened 2024-03-20
[Bug] Use snapshot for screening dialog

vd1992:
🤖 Resolves #9812 

## 👋 Introduction

Rectify the dialog component to pull data to render out of the snapshot. Also noticed the screening question responses ought to do that too. 
Just recycled code from elsewhere for the most part ♻️ 

## 🧪 Testing

1. Submit an application
2. Retroactively update stuff, it should not be reflected in the dialog

## 📸 Screenshot

Dialog view

image

Updated experience

image

Screening questions still rendered

image

Issue #9861, opened 2024-03-21
🐛 `ScreeningDecisionDialog` storybook story error


mnigh:
## 🐛 Bug

The `ScreeningDecisionDialog` stories are broken: https://www.chromatic.com/component?appId=61e099a1bb1465003a73d3ce&csfId=components-screening-decisions-screeningdecisiondialog&buildNumber=6787&k=65fc5c05dfe14120961e7e32-1200px-interactive-true&h=9&b=-1

## 🦋 Expected Behaviour

`ScreeningDecisionDialog` stories should render without error.

## 📋 Steps to Reproduce

1. Run storybook
2. View `ScreeningDecisionDialog` stories

## 📸 Screenshot

<img width="1438" alt="Screen Shot 2024-03-21 at 13 27 03" src="https://github.com/GCTC-NTGC/gc-digital-talent/assets/3046459/c95d57af-0007-4e1f-bbb6-ddc9ca8301c6">

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the bug was addressed.

- [ ] `ScreeningDecisionDialog` stories renders without error

mnigh:
Possibly related to #9841?

Issue #9860, opened 2024-03-21
🐛 Assessment Step Table storybook story error


mnigh:
## 🐛 Bug

The **Assessment Step Table** story is broken: https://www.chromatic.com/component?appId=61e099a1bb1465003a73d3ce&csfId=components-assessment-step-table&buildNumber=6787&k=65fc5c05dfe14120961e7e0d-1200px-interactive-true&h=8&b=-1

## 🦋 Expected Behaviour

**Assessment Step Table** story should render without error.

## 📋 Steps to Reproduce

1. Run storybook
2. View Assessment Step Table story

## 📸 Screenshot

<img width="1419" alt="Screen Shot 2024-03-21 at 13 24 06" src="https://github.com/GCTC-NTGC/gc-digital-talent/assets/3046459/379eb2fb-392e-4b75-aa37-52e6f09b1143">

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the bug was addressed.

- [ ] Assessment Step Table story renders without error

PR #9865, opened 2024-03-21
[Fix] Assessment results related stories

vd1992:
## 👋 Introduction

Oops, I broke some stories with #9841 
Switched to rendering from snapshot field but need it either present but null, or present correctly typed. So I did that 

## 🧪 Testing

1. Stories fixed
2. Assessment results table, and screening decision dialog stories

### DIFF
--- a/apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.stories.tsx
+++ b/apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.stories.tsx
@@ -28,6 +28,140 @@ experience.skills?.push(skill);
 poolCandidate.user.experiences?.push(experience);
 poolCandidate.user.userSkills?.push(fakeUserSkills(1, skill)[0]);
 
+const profileSnapshot = {
+  id: poolCandidate.user.id,
+  firstName: poolCandidate.user.firstName,
+  experiences: [
+    {
+      id: "11e06444-19f3-41bc-a8c0-557f2391791a",
+      role: "Residential Defender",
+      skills: [
+        {
+          id: "b9d2f8c5-c24e-4993-8e6d-10dd046fd387",
+          key: "cobol",
+          name: { en: "COBOL", fr: "COBOL" },
+          keywords: { en: null, fr: null },
+          description: {
+            en: "Demonstrated ability to apply concrete knowledge of this language to develop a variety of projects.",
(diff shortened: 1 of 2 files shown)


vd1992:
UI Review -> https://www.chromatic.com/review?appId=61e099a1bb1465003a73d3ce&number=9865&type=linked
The dialog stories -> https://www.chromatic.com/component?appId=61e099a1bb1465003a73d3ce&csfId=components-screening-decisions-screeningdecisiondialog--education-requirement&buildNumber=6790&k=65fc7d13d9ef596ae575b355-1200px-interactive-true&h=4&b=-2

Appearing as a button that you have to click isn't new

apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.stories.tsx
```diff
@@ -28,6 +28,140 @@ experience.skills?.push(skill);
 poolCandidate.user.experiences?.push(experience);
 poolCandidate.user.userSkills?.push(fakeUserSkills(1, skill)[0]);
 
+const profileSnapshot = {
```
mnigh, line 31: Could you use generated fake data, specifically for the `experiences` and `poolCandidates` keys instead so that it can be reused? We already have `fakePoolCandidates` and `fakeExperiences`, based on variables set previously in this file.

mnigh (commented):
Is this PR meant to resolve one or both of:

- #9860
- #9861

apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.stories.tsx
vd1992: Oh ya, I could write it nicer https://github.com/GCTC-NTGC/gc-digital-talent/pull/9865/commits/7540f18e71ad852f022f3040f6063555a752a77d

vd1992:
> Is this PR meant to resolve one or both of:

I started this before those issues were created, I guess it resolves both

mnigh (approved):
One suggestion would be to delay the Chromatic snapshots for the five [`ScreeningDecisionDialog` stories](https://www.chromatic.com/component?appId=61e099a1bb1465003a73d3ce&csfId=components-screening-decisions-screeningdecisiondialog&buildNumber=6792) since they all look the same and are of little value if they only render the text **To assess** in the snapshot, but maybe worth another issue since it's not really part of the scope of fixing the erroring stories?

No errors on either story anymore though so 👍

PR #9874, opened 2024-03-21
[Debt] Migrate to internal packages

### LINKED ISSUE #9845: ♻️ Move to internal packages
## ♻️ Debt/Refactor

Migrate our `packages/**` to internal packages to simplify dev workflow.

## 🕵️ Details

> [!NOTE]
> We technically have two partial internal packages in `forms` and `ui` since getting context working in them was hard so I gave up and just let webpack transpile them

We initially setup our monorepo to use external packages in the hopes of publishing them. However, this has caused a lot of problems with the dev experience and complicated things a bit. Also, our packages are fairly specific to our repo and would be difficult to publish and re-use outside of it.

If we use internal packages, we shouldn't need to spin up 30 different processes all watching and sometimes crashing. Also, the packages should be more exposed to debuggers.

## 🙋‍♀️ Proposed Solution

It should be pretty easy since internal packages are far simpler than external packages. 

- Update `package.json`
    - Point `types` to the index file
    - Convert `main` to `"exports": { ".": "./src/index.tsx"` } (or whatever the index file is)
- Optional: If we really want we could remove the `tsup` dependency as well as `build`, `watch` and `dev` scripts
   - I would only do this if we really don't ever plan on publshing anything sicne this was a pain to setup :sweat_smile: 

## ✅ Acceptance Criteria

- [ ] `packages/**` are updated to be internal
- [ ] App build scripts (`watch`, `dev`, `build`) still function as expected, or better
- [ ] Main watch command updated to only run on a single workspace at a time (`@gc-digital-talent/web`)

esizer:
🤖 Resolves #9845

## 👋 Introduction

Updates our `packages/**` to be internal packages rather than external removing additional bundling tooling.

## 🕵️ Details

This had some very nice improvements as a side effect

- `pnpm run watch` is much improved
- Massive reduction in number of scripts we needed
- Bundle size reduced from `3.51 MiB` to `3.37 MiB` ~4% reduction in initial bundle size
- Build time seems to be quicker too `2m5s` to `1m17s` on my machine ~37% reduction in build time

## 🧪 Testing

> [!TIP]
> Delete command: `find . -name '{folderName}' -type d -prune -exec rm -rf '{}' +`

1. Delete all of the following directories reucrssivley:
    - `.turbo`
    - `node_modules`
    - `dist`
2. Reinstall packages `pnpm install`
3. Run through all of the commands confirming they work as intended


turbo.json
petertgiles, line 1: Watch mode works but I'm a little confused.  When I fire it off I see the watch mode firing off in parallel for all the packages, as usual.  However, when I make a change in a package, only the `web` app reruns, which is different.  But it works!  Is it still necessary for all the packages to fire watch in parallel?

petertgiles (changes requested):
Oh, this is cool!  I'm not seeing much of a performance difference myself, but being able to debug packages is the big W for me.  :heart:  I've left a few questions below.

Also, I can't test Playwright.  Hopefully someone else can.

package.json
```diff
@@ -14,33 +14,27 @@
...
     "lint": "turbo run lint",
-    "storybook:project": "pnpm run storybook",
     "storybook": "pnpm --filter @gc-digital-talent/web run storybook",
     "storybook:design-system": "pnpm --filter @gc-digital-talent/web run storybook:design-system",
     "storybook:web": "pnpm --filter @gc-digital-talent/web run storybook:web",
     "build-storybook": "pnpm --filter @gc-digital-talent/web run build-storybook",
     "test": "turbo run test",
-    "watch": "turbo run watch --concurrency=30",
-    "watch:project": "turbo run watch",
+    "watch": "turbo run watch --filter @gc-digital-talent/web --filter @gc-digital-talent/graphql",
```
mnigh, line 37: [Similar](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9874#discussion_r1535679122) to @petertgiles's comment, `watch` works but is very slow (10+ seconds) for webpack to compile when changing anything outside `apps/web`: is this different than before?

.github/workflows/translations.yml
```diff
@@ -36,19 +36,14 @@ jobs:
...
       - name: "Extract English files: all workspaces"
         working-directory: ./
         run: pnpm run intl-extract
 
       - name: "Generate untranslated files: all workspaces"
         working-directory: ./
         run: |
-          pnpm run check-intl:i18n
-          pnpm run check-intl:web
+          pnpm run check-intl --force
```
mnigh, line 46: Nice!

apps/web/.storybook/preview.js
```diff
@@ -8,10 +8,7 @@ import {
   VIEWPORTS
 } from "@gc-digital-talent/storybook-helpers";
 import { richTextElements as defaultRichTextElements } from "@gc-digital-talent/i18n";
-import "@gc-digital-talent/forms/dist/index.css";
```
mnigh, line 11: 🆒

packages/auth/package.json
```diff
@@ -35,13 +29,11 @@
   },
   "devDependencies": {
     "@gc-digital-talent/eslint-config": "workspace:*",
-    "@swc/core": "^1.4.8",
```
mnigh, line 38: 👋

mnigh (changes requested):
Nice work with all of the `import` consolidation, very clean. Playwright and Cypress both fail for me though, not sure why they are passing in CI.

## Screenshot
Screen Shot 2024-03-22 at 10 55 31

turbo.json
esizer: okay, so only some should because of the turbo depends on. What should be happening is:

1. CSS is built by h2
2. Strings are extracted
3. French is compiled
4. Codegen occurs
5. Web is built and settles into watch mode

After this it really should only build web when files are changed.

mnigh:
> Nice work with all of the `import` consolidation, very clean. Playwright and Cypress both fail for me though, not sure why they are passing in CI.

I tried running the `setup.sh` script and it failed 😕.

## Screenshot

Screen Shot 2024-03-22 at 11 00 44

esizer:
> I tried running the `setup.sh` script and it failed 😕.

Can you try `docker-compose down & docker-compose up --build --detach`?

package.json
esizer: Yeah, I just checked main and it takes roughly the same amount of time give or take. Only more reason to move on to vite :rofl:

pnpm-lock.yaml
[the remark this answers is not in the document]
esizer: oh cool, I didn't know that existed :laughing: 

Updated in [update debug path config](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9874/commits/e877675ed0281e796c198087917a57ab330ebb2a)

mnigh:
> > I tried running the `setup.sh` script and it failed 😕.
> 
> Can you try `docker-compose down & docker-compose up --build --detach`?

Sure. I still see the errors in Playwright and Cypress though 🤷

vd1992 (approved):
Seems to work great for me!

Build times down ~1/3
Scripts seem to work fine, Cypress tests all passed, and site still appears to work

vd1992:
Playwright has been flaky for me, but ran a few specs and it still seems to work

PR #9882, opened 2024-03-22
[Bug] Resolve missing roles from users table

### LINKED ISSUE #9862: 🐛 No filter for new "Community manager" role in "All users" table
## 🐛 Bug
On the All users page, the table allows us to filter by "Roles and permissions", but the new role "Community manager does not appear in the filters list. In addition, "Community manager" does not appear in the "Roles and permissions" column in the All users table.

## 🦋 Expected Behaviour

Community manager should be a filter on the All users table and show show up in the Roles and permissions column.

## 📸 Screenshot

image

image

## ✅ Acceptance Criteria

- [ ] Community manager added to filters on the All users table
- [ ] Community manager shows up in the Roles and permissions column of the All users table

vd1992:
🤖 Resolves #9862 

## 👋 Introduction

Have it so no roles are missing, that should be present. Did the recommended inversion of the filtering 
Quick fix 🔧 

The roles NOT present are guest, user, and applicant 

## 🧪 Testing

1. Navigate to users table
2. Can filter by the desired roles

## 📸 Screenshot

image

image

### DIFF
--- a/apps/web/src/pages/Users/IndexUserPage/components/UserFilterDialog.tsx
+++ b/apps/web/src/pages/Users/IndexUserPage/components/UserFilterDialog.tsx
@@ -22,7 +22,6 @@ import {
   enumToOptionsWorkRegionSorted,
 } from "@gc-digital-talent/forms";
 import { unpackMaybes } from "@gc-digital-talent/helpers";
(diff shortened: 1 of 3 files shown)


esizer (approved):
LGTM! :+1:

PR #9890, opened 2024-03-25
[Feature] Generate candidate profiles

### LINKED ISSUE #9686: Create Server-Side Pool Candidate Report Document Generator
## ✨ Feature

We want to start generating the Pool Candidate report documents on the server.  Let's set up the view to do that.

## 🕵️ Details

This issue only deals with the report document itself.  How the client requests it and how it gets back to the client are out of scope for now.

## 🎨 Design File

The server-side report should match the current client-side report design.

[Laravel-pdf](https://spatie.be/docs/laravel-pdf/v1/introduction)
[PHPOffice    /    PHPWord](https://github.com/PHPOffice/PHPWord)

## ✅ Acceptance Criteria

- [ ] It is possible to generate a document with one or more pool candidates on the server
  - [ ] A dev can trigger the request manually.  A regular user cannot.
  - [ ] No need to authenticate request yet
  - [ ] No need to save the file on the server, can just send back an HTTP response
- [ ] Report matches the report generated by the client

## 🛑 Blockers

```[tasklist]
### Blocked By
```

esizer:
🤖 Resolves #9686 

## 👋 Introduction

Adds the ability to generate word docs of pool candidate profiles.

## 🕵️ Details

I didn't end up doing the ZIP work because this ended up being quite large and didn't want to expand the scope further.

Also, right now it takes 3 steps to create the file and I'm not sure if that makes sense:

```php
$doc = new CandidateProfileDoc(["uuid"]);
$doc-generate(); // We could write here to get rid of the third step. 
$doc->write("filename.docx");
```

I did this so we could write in a different way and pass the doc on fully generated but I'm not sure if that is useful and we could just like less code :woman_shrugging: 

## :globe_with_meridians: Internationalization

We have no way of translating from within laravel. So, while I set it up so we can pass a language into the generator, it only translates the fields where we have the French in the DB. That means all other text only appears in English for now.

## 🧪 Testing

> [!TIP]
> You may need to link storage if you havent `php artisan storage:link`

1. Get either one or multiple pool candidate IDs
2. Drop down into tinker `php artisan tinker`
3. Run `$doc = new \App\Generators\ProfileCandidateDoc([]);` passing in an array of the UUIDs from step 1
4. Run `$doc->generate();` confirming no errors
5. Run `$doc->save('filename.docx');` confirming no errors
    - Filename can be anything, the important part is the extension
6. Navigate to `/api/storage/generated/` in the file system
7. Confirm your generated file is there and looks like [the sample](https://docs.google.com/document/d/13epIYbnfFdYc-T1_d8SqCKhaJcAVJEXDYkhGBmzLLdA/edit)

### DIFF
--- a/api/app/Generators/CandidateProfileDoc.php
+++ b/api/app/Generators/CandidateProfileDoc.php
@@ -27,13 +27,18 @@ class CandidateProfileDoc extends DocGenerator
         $candidates = PoolCandidate::with([
             'user' => [
                 'department',
-                'currentClassification'
+                'currentClassification',
+                'awardExperiences' => ['userSkills' => ['skill']],
+                'communityExperiences' => ['userSkills' => ['skill']],
+                'educationExperiences' => ['userSkills' => ['skill']],
+                'personalExperiences' => ['userSkills' => ['skill']],
+                'workExperiences' => ['userSkills' => ['skill']],
             ]
         ])
             ->whereIn('id', $this->ids)
             ->get();
 
-        if(empty($candidates)){
+        if (empty($candidates)) {
             throw new Exception("No candidates found.");
         }
 
@@ -63,7 +68,7 @@ class CandidateProfileDoc extends DocGenerator
 
             $this->addSubTitle($section, 'Government information');
             $this->addLabelText($section, 'Government of Canada employee', $candidate->user->is_gov_employee ? "Yes" : "No");
-            if($candidate->user->is_gov_employee) {
+            if ($candidate->user->is_gov_employee) {
                 $department = $candidate->user->department()->first();
                 $this->addLabelText($section, 'Department', $department->name[$this->lang] ?? "");
                 $this->addLabelText($section, 'Employee type', $candidate->user->getGovEmployeeType());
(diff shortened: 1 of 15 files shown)


petertgiles:
Some notes on the testing instructions:
- Should be "\App\Generators\CandidateProfileDoc"
- Should be "$doc->write('filename.docx');"

api/app/Generators/CandidateProfileDoc.php
petertgiles, line 1: Some changes from existing exports.
image
image
image

petertgiles (changes requested):
Looking very cool!  Some initial first impressions:

api/app/Models/User.php
vd1992, line 1: Not a request for any changes but a concern, the user model file is getting pretty big and hard to tell when something is no longer needed. Or for that matter, what some stuff is even for

esizer: Yeah, I agree. We may want to consider splitting these into traits or go full service layer with them :stuck_out_tongue:

api/app/Generators/CandidateProfileDoc.php
esizer: Okay, updated to match the existing download instead of the provided template in [update to match existing document](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9890/commits/ec01c164ebeed6b966a564d615838a3438a368c9)

api/config/filesystems.php
```diff
@@ -35,6 +35,11 @@
             'root' => storage_path('app'),
         ],
 
+        'generated' => [
```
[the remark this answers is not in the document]
esizer: I liked that yours was clear about the fact it is users generating the files and not the system. Aligned with that in [align generated file system dir](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9890/commits/67728cfd710c925d037ad53b9748634df3b4bad2)

api/app/Generators/CandidateProfileDoc.php
petertgiles: Looking a lot better!  A few more...
image
image
image
image

esizer: oh good eye! Okay, I think I finally got it with these two:

 - [update to add missing information](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9890/commits/d82d5f7ff6aca2726552b8a87638008fb4ae74db)
 - [add experience specific information](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9890/commits/0fcf1d945bbf61f680c0f5fc90fdfb7cc7ea8c4c)

api/app/Generators/DocGenerator.php
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    protected function addLabelText(Element\Section $section, string $label, string $text)
+    {
+        $run = $section->addTextRun();
+        $run->addText($label.': ', $this->strong);
+        $run->addText($text);
+    }
+
+    protected function santizeEnum(string $enum): string
```
petertgiles, line 50:
```suggestion
    protected function sanitizeEnum(string $enum): string
```

api/app/Generators/CandidateProfileDoc.php
petertgiles: Sorry, a few more.  :grimacing:
image
image

api/app/Generators/DocGenerator.php
esizer: oof, I'm blaming this one on copilot! It auto completed this definition for me because i had a similar one I was playing with and copilot propagated a previous mistake :confounded: 

Anyway, should be fixed in [fix typo in function name](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9890/commits/fa995facf3bb90b4eec569d5a526cfb591c6f7db)

api/app/Generators/CandidateProfileDoc.php
esizer: okay, that was surprisingly tricky. I was no aware we deprecated some operational requirements :sweat: I'm pretty sure this is right :confused: 

 - [fix bug with getting string value of citizenship](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9890/commits/3fdb5b5d139acca97b8c8adb820d153c81eb4bb0)
 - [do not show scheduled overtime](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9890/commits/46965229bcd12ce14fb515b3066f810989285456)

api/app/Models/User.php
petertgiles: A lot of the new code is User is enum formatters.  Those might be moved to the enum themselves.  For example, instead of ` $candidate->user->getLanguage('preferred_lang')` it could be `Language::toEnString($candidate->user->preferred_lang)`.  Or if they have logic specific to the document it could be in the document file or a document helper file.  `getLookingForLanguage($candidate->user)`

We might be overhauling how the localized constants and i18n stuff in general works soon so it probably isn't worth refactoring right now.

esizer: oooo, I like moving it to the enum itself. That might lend itself very well to moving localized constants to the server :eyes:

api/app/Generators/CandidateProfileDoc.php
petertgiles: Sorry, I'm still seeing extra operational requirements.  :sob:
image

petertgiles: Yeah, for sure.  Not great that this is so hard to find.  :disappointed:

petertgiles: OK, I think good to go!

petertgiles (approved):

PR #9899, opened 2024-03-26
[Debt] Experience table refactor

### LINKED ISSUE #7157: ♻️ Refactor Experience tables to simplify queries

petertgiles:
🤖 Resolves #7157

## 👋 Introduction

This branch refactors the experience tables.  It attempts to simplify querying by removing individual experience tables and moving the custom experience fields into a JSONB column.

## 🕵️ Details

There was an experience_type column added to the experiences table so that the `newFromBuilder` knows why type of model to build from the data.  There are also global scopes added to the individual experience models so that only the right types of rows are queried.

Pre-existing issues:
- Can't edit an experience that was just added #9902

## 🧪 Testing

1. Check out a commit before this branch starts (ed131f46a9)
2. Set up some good data
```php
$user = User::where('sub', 'admin@test.com')->sole();
$pool = Pool::select('id')->where('name->en', 'CMO Digital Careers')->sole();

AwardExperience::factory(20)->for($user)->create();
CommunityExperience::factory(20)->for($user)->create();
EducationExperience::factory(20)->for($user)->create();
PersonalExperience::factory(20)->for($user)->create();
WorkExperience::factory(20)->for($user)->create();
// unfortunately our seeders don't link skills to experience so you'll have to do that manually. #9964
``` 
3. Check out the branch HEAD and run migrations
4. Rebuild the app

- [ ] Migration can roll forward and back without losing data
- [ ] App functions normally
  - [ ] can add, edit, remove experiences to profile
  - [ ] can link and unlink skills to experiences
  - [ ] can go to skill profile and see experiences linked to skills
  - [ ] can apply to a pool and link education-requirement experiences
  - [ ] can edit education-requirement experiences (both education and applied-work)
  - [ ] can see application with experiences and education-requirement experiences
  - [ ] Can see application snapshot with the same
  - [ ] Can export user profile and pool candidate documents


esizer (approved):
Pure excellence :star_struck: 

I'd wait to see if @tristan-orourke has any thoughts before merging though :+1:

vd1992:
Playing around with things a bit, so far...

Fill out profile, then migrate -> rollback -> migrate ✅ 
Add experience and view experiences after migrations ✅ 
Applying and exercising RoD after migrations ✅ 

Things seem to be working fine! Wow! 🤩 
Will dig around a bit more as things to check come to mind

api/app/Models/AwardExperience.php
```diff
@@ -25,21 +27,86 @@ class AwardExperience extends Experience
...
+
+    /**
+     * Interact with the experience's title
+     */
+    protected function title(): Attribute
+    {
+        return Attribute::make(
+            get: fn (mixed $value, array $attributes) => $this::getJsonPropertyString($attributes, 'title'),
+            set: fn (mixed $value, array $attributes) => $this::setJsonPropertyString($value, $attributes, 'title')
+        );
```
tristan-orourke, line 66: This pattern is repeated so much I might even abstract it a bit more 😆 Not that this is a necessary change or anything.

```
function makeJsonStringAttribute($attribute_name)
{
    return Attribute::make(
        get: fn (mixed $value, array $attributes) => $this::getJsonPropertyString($attributes, $attribute_name),
        set: fn (mixed $value, array $attributes) => $this::setJsonPropertyString($value, $attributes, $attribute_name)
    );
}
```

tristan-orourke (approved):
Very nice! I learned some tricks with Eloquent here.

api/app/Models/AwardExperience.php
petertgiles: I gave it a try.  It's a little drier, though (IMO) a little harder to follow.  Let me know which way you prefer.
[make attributes drier](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9899/commits/5b07433e3d651b42780ac782d12e6c33fb681a22)

petertgiles:
> Screening questions appear to be missing from print application, but I doubt that has anything to do with this PR

You're right, this is pre-existing.  However, the issue is actually that the section in the print document is misnamed.  It's actually general questions, of which there aren't any.  I opened  #9997 .

vd1992:
> > Screening questions appear to be missing from print application, but I doubt that has anything to do with this PR
> 
> You're right, this is pre-existing. However, the issue is actually that the section in the print document is misnamed. It's actually general questions, of which there aren't any. I opened #9997 .

Ah, thanks for getting to the bottom of it and making the issue

vd1992 (approved):

PR #9909, opened 2024-03-26
[Bug] Skill table filter fail on later pages

### LINKED ISSUE #9677: 🐛 Changing filters might result in false empty results for Skill table
## 🐛 Bug

Changing the filters of the Skill table doesn't reset which "page" you're on. Its possible to end up on a higher page than exists, and the table thinks there are no results when there should be some. The Skill table handles the filtering client side (different from other api handled tables), so the paginated data needs to be filtered as well to match the filtered results.

## 🦋 Expected Behaviour

If the filters change, table should be reset to the first page.

## 🕵️ Details

## 📋 Steps to Reproduce

1. Go to the public Skills table
3. Add some filters which decrease the number of results (and the number of pages) but still have some results.
4. Reset the filters
5. Go to the last page of the table
6. Add the same filters from before
7. Table appears empty

## 🙋‍♀️ Proposed Solution

Reset the table page when filters are updated

## ✅ Acceptance Criteria

- [ ] When search or filters are updated, always go back to the first page of the table

JamesHuf:
🤖 Resolves #9677

## 👋 Introduction

Updates the ResponsiveTable component to reset to the first page when filters change.

## 🕵️ Details

Also fixes a bug I noticed with the category filter on the skills table not working correctly.

## 🧪 Testing

1. Go to the public Skills table `/en/skills`
2. Add some filters which decrease the number of results (and the number of pages) but still have some results
3. Reset the filters
4. Go to the last page of the table
5. Add the same filters from before
6. Confirm the table returns to to the first page and displays results as expected


esizer (approved):
Cool! :sunglasses:

PR #9916, opened 2024-03-27
[Fix] Disable add button as soon as limit reached

### LINKED ISSUE #9908: 🐛 Cards in General questions not saving correctly
## 🐛 Bug

The cards in the General questions contains two cards with the same number and the wrong values from what had been entered in the form. This might be related to the `CardRepeater` component.

## 🦋 Expected Behaviour

Cards numbers should increment by one and should render the values that were entered in the form.

## 📋 Steps to Reproduce

1. Navigate to `/admin/pools` page
2. Click **Create process** button
3. Navigate to General questions section
4. Click **Add a new question** button
5. Enter values for required fields
6. Click **Add a new question** button
7. Enter values for required fields that are different than previous question
8. Observe data that does not reflect what was saved

## 📹 Screen recording

https://github.com/GCTC-NTGC/gc-digital-talent/assets/3046459/7b4c81e8-db18-477b-8587-ffb501869de6

## ✅ Acceptance Criteria

- [ ] Cards numbers should increment by one and should render the values that were entered in the form

esizer:
🤖 Resolves #9908 

## 👋 Introduction

Fixes an issue where users could add more than the maximum general questions to a job poster.

## 🕵️ Details

This came down to two things:

1. We were spreading all props to the add button meaning the `disabled={maxReached}` could be overwritten by accident
2. We don't have server validation so we need to rely more on the frontend (disabling the form until we know if you can add more)

## 🧪 Testing

1. Build `pnpm dev`
2. Login as admin `admin@test.com
3. Create a process
4. Scroll down to the general questions section
5. Add questions as fast as you can
6. Confirm that you can;t add more than the max (10) no matter how quickly you add them

### DIFF
--- a/apps/web/src/pages/Pools/EditPoolPage/components/GeneralQuestionsSection/GeneralQuestionDialog.tsx
+++ b/apps/web/src/pages/Pools/EditPoolPage/components/GeneralQuestionsSection/GeneralQuestionDialog.tsx
@@ -21,11 +21,13 @@ const TEXT_AREA_MAX_WORDS = 200;
 interface GeneralQuestionDialogProps {
   question?: GeneralQuestion;
   index?: number;
+  disabled?: boolean;
 }
 
 const GeneralQuestionDialog = ({
   question,
   index,
+  disabled,
 }: GeneralQuestionDialogProps) => {
   const intl = useIntl();
   const [isOpen, setIsOpen] = React.useState<boolean>(false);
@@ -65,12 +67,13 @@ const GeneralQuestionDialog = ({
       <Dialog.Trigger>
         {isUpdate ? (
           <CardRepeater.Edit
+            disabled={disabled}
             aria-label={intl.formatMessage(formMessages.repeaterEdit, {
               index,
             })}
           />
         ) : (
-          <CardRepeater.Add>
+          <CardRepeater.Add disabled={disabled}>
             {intl.formatMessage({
               defaultMessage: "Add a new question",
               id: "uEqA50",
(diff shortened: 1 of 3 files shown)


mnigh (commented):
I could not add more than the 10 question limit which is great, but I did encounter problems with ordering and deleting items in quick succession (duplicate keys, see the end of the screen recording). I'm not sure if this issue should have its scope expanded and fix all at the same time. Thoughts?

## 🎥 Screen recording

https://github.com/GCTC-NTGC/gc-digital-talent/assets/3046459/9d5fd119-bdb3-47fa-adcf-954a509bfd83

esizer:
> I could not add more than the 10 question limit which is great, but I did encounter problems with ordering and deleting items in quick succession (duplicate keys, see the end of the screen recording). I'm not sure if this issue should have its scope expanded and fix all at the same time. Thoughts?

I made my best effort smooth that out :sweat_smile: It is still a little wonky but I think it is a bit better :thinking: 

[simplify general question handling to avoid use effect](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9916/commits/ed7b968cc198a8bc12fbcd7c9db7174c4ee542dd)

esizer:
> Is there any way to disable delete until after the animation of the order changing? That seems to be an issue.

Yup, this should do it: [disable remove while updating](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9916/commits/f0b1f56197dc9bdfbea7c578833a53ab25c30928)

mnigh (approved):
Nice work! This is definitely an improvement on what was there before and meets the acceptance criteria of the linked issue.

The ordering and deletion still seems a bit unstable. Any chance you could create an issue for that? If you need it, feel free to use my [screen recording](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9916#issuecomment-2023493426) in the issue.

PR #9915, opened 2024-03-27
[Fix] Remove merge groups

esizer:
## 👋 Introduction

This is a test to see if `pnpm` + merge groups is the issue with dependabot not updating the lock file :woman_shrugging:

PR #9937, opened 2024-03-27
Revert "[Fix] Remove merge groups"

esizer:
Reverts GCTC-NTGC/gc-digital-talent#9915

This was not the issue and we do want merge groups


mnigh (approved):
⏪

PR #9945, opened 2024-03-28
[Feature] Close process immediately

### LINKED ISSUE #9863: Close Early button on published Processes
## Feature
There is no way to close published processes immediately.

The only way to edit published processes is with the "Extend closing date" button. Technically, you can actually shorten date in the dialog opens, not just extend it - though that's a bit confusing and unintuitive. However, the earliest you can set the closing date to is "Today at midnight pacific time". 

We want the ability to close posters immediately. If only to fix mistakes.

## Details
Add a new option to the "Extend closing date" dialogue for "close immediately". Also prompt the user for a required justification and save it.

## Design 
[Design file](https://www.figma.com/file/em5ArPr044YaIe1Vb1zCZz/Admin---Processes%2FPools%2FRecruitments---Record-of-Decision?type=design&node-id=8738%3A33771&mode=design&t=UvdtFuhMYWPUDPox-1) by @Jerryescandon 

### Current dialogue
image

### New dialogue

image

### New functionality
https://github.com/GCTC-NTGC/gc-digital-talent/assets/16977254/515a0f80-aa5a-4fc6-a145-63c01bcc9c92

```[tasklist]
## Acceptance criteria
- [ ] Link on Process information tab changed from "Extend closing date" to "Change closing date".
- [ ] Extend closing date dialogue renamed to "Change closing date"
- [ ] Radio button added to "Change closing date" with 2 options (Extend closing date and Close immediately)
- [ ] Close immediately option Prompts user to add a required "Reason for closing". "Change closing date" option leaves the rest of the dialogue unchanged
- [ ] Button adjusts to radio button selection as well
- [ ] If close process immediately button is trigger, the process closing date is changed to the current time (or one minute in the past... whatever works best)
```

esizer:
🤖 Resolves #9863 

## 👋 Introduction

Adds the ability to close a process immediately with a reason.

## 🕵️ Details

I wasn't sure if the reason should be required in the mutation itself or not. Easy change if we want to do that but for now it only validates on the front end.

## 🧪 Testing

1. Migrate the data `make refresh-api`
3. Navigate to  a published pool
4. Confirm the "Extend" button has been updated to a "Change" button
5. Confirm dialog matches [the mockup](https://www.figma.com/file/em5ArPr044YaIe1Vb1zCZz/Admin---Processes%2FPools%2FRecruitments---Record-of-Decision?type=design&node-id=8738-33771&mode=design)
6. Confirm you can still extend the date
7. Confirm you can close the date and the reason is stored in the DB

## 📸 Screenshot

screenshot_28-Mar-2024_10-54-1711637647
screenshot_28-Mar-2024_10-54-1711637655


petertgiles (changes requested):
Looking great!  I noticed a few edge cases to take a look at.

apps/web/src/pages/Pools/ViewPoolPage/components/ChangeDateDialog.tsx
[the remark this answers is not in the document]
esizer: :thinking: How did you accomplish this? We do have a rule to prevent that, is it just not working for you? Because it seems to be working for me.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/526ba6963a694cf508cc0d5eb014cfc4a8b47702/api/graphql/schema.graphql#L1399-L1404

screenshot_02-Apr-2024_12-20-1712074824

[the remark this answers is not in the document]
esizer: oh my, I didn't even realize we had another mutation just to change the date :woman_facepalming: Do we prefer that one over just using the `updatePool` mutation? It find it  a bit confusing we have all these different mutations to update specific fields :confused:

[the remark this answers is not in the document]
esizer: Made the reason required in [make reason required in mutation](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9945/commits/5ec504eb0663413a67d5ae4214167f596640898b)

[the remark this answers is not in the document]
esizer: Okay, I went ahead and added some validation to the rule to prevent setting a date before the existing one (if there is an existing one) in addition to preventing it from being in the past.


[prevent reducing process closing date](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9945/commits/95156e1fd49c5a4f892dd8c9752e70fe4d9d3bba)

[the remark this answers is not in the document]
esizer: I think this should also be covered by [prevent reducing process closing date](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9945/commits/95156e1fd49c5a4f892dd8c9752e70fe4d9d3bba)

apps/web/src/pages/Pools/ViewPoolPage/components/ChangeDateDialog.tsx
```diff
@@ -0,0 +1,240 @@
...
+                    data-h2-margin-bottom="base(x.5)"
+                    label={intl.formatMessage({
+                      defaultMessage: "Reason for closing",
+                      id: "SQRzSl",
+                      description: "Label for the reason for closing the pool",
+                    })}
+                    name="reason"
+                    rules={{
+                      required: intl.formatMessage(errorMessages.required),
+                    }}
```
vd1992, line 190: String column maxes out at 256 characters I believe, so ought to squeeze in `wordLimit={256}`
Avoid rejected operations that way

esizer: oh jeeze, so I agree with a word limit but I just realized that we should be using `text` instead of `string` here :sweat_smile: Thanks for spotting this!

vd1992: That works too 😛

esizer: okay, updated with the new column type and word limit in [update reason to text column, set word limit](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9945/commits/a4fdb3daf3c5df55b0b11fabf7f35204d7a15387)

api/app/GraphQL/Validators/ChangePoolClosingDateValidator.php
```diff
@@ -0,0 +1,41 @@
...
+
+        $id = $this->arg('id');
+        $pool = Pool::find($id);
+
+        return [
+            'id' => ['uuid', 'required'],
+            'newClosingDate' => [
+                'after:today',
+                Rule::when(! is_null($pool) && $pool->closing_date, ['after_or_equal:'.$pool?->closing_date]),
+            ],
```
vd1992, line 30: I know `new_closing_date` and `newClosingDate` already existed but I find them confusing as there is no such field and it becomes a chain of 
closing_date -> new_closing_date -> closing_date
for the validation and update process

api/graphql/schema.graphql
```diff
@@ -1823,14 +1825,19 @@ type Mutation {
...
-    newClosingDate: DateTime!
-      @rename(attribute: "new_closing_date")
-      @rules(
-        apply: ["after:today"]
-        messages: [{ rule: "after", message: "UpdatePoolClosingDate" }]
-      )
-  ): Pool @guard @canFind(ability: "changePoolClosingDate", find: "id")
-  closePool(id: ID!): Pool @guard @canFind(ability: "closePool", find: "id")
+    newClosingDate: DateTime! @rename(attribute: "new_closing_date")
+  ): Pool
```
vd1992, line 1829: Same as the comment above, as far as I can tell, we have a habit of using the rename directive to change to a real database field. This is not the case here which is confusing
Pre-existing I know

api/app/GraphQL/Validators/ChangePoolClosingDateValidator.php
vd1992, line 1: This is a little confusing as we already have an identically named file 

image

referenced in the mutation itself
They validate different things but it is confusing to have a mutation execute two different validators with the same name

image

api/app/GraphQL/Mutations/ClosePool.php
```diff
@@ -15,7 +15,10 @@ final class ClosePool
     public function __invoke($_, array $args)
     {
         $pool = Pool::find($args['id']);
-        $pool->update(['closing_date' => Carbon::now()]);
+        $pool->update([
+            'closing_date' => Carbon::now(),
+            'closing_reason' => isset($args['closing_reason']) ? $args['closing_reason'] : null,
+        ]);
```
vd1992, line 21: Should the `ChangePoolClosingDate` mutation then null out `closing_reason`

vd1992 (changes requested):
Sorry, I failed to hit "Start a review" earlier 😵

api/app/GraphQL/Validators/ChangePoolClosingDateValidator.php
esizer: Oops, did not see that there :sweat_smile: I guess lighthouse wasn't aware during generation because it was nested in `Mutations` folder.

Merged them in [merge change closing date validators](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9945/commits/6a45f80cb07993c28b511cdcb2cbb6dc8ff00835)

api/graphql/schema.graphql
esizer: Yeah, that makes sense. Updated as part of [update field name for closing date, fix tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9945/commits/9c997aa5100d93589a7aa87eb6cea257846fff61)

api/app/GraphQL/Mutations/ClosePool.php
esizer: That makes sense to me, updated in [merge change closing date validators](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9945/commits/6a45f80cb07993c28b511cdcb2cbb6dc8ff00835)

apps/web/src/pages/Pools/ViewPoolPage/components/ChangeDateDialog.tsx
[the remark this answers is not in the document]
petertgiles: It prevents it from happening now but the experience isn't great.  The dialog closes and one vague error and one irrelevant message appears.
image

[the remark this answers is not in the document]
petertgiles: > It find it a bit confusing we have all these different mutations to update specific fields 😕

Yeah, unless there are special rules in play (like edit draft pool vs publish pool) it would probably be good to consolidate them.  Out of scope for here but it might be nice if you could open an issue for it.

tristan-orourke:
Maybe, to make multiple mutations less confusing, we rename them to `updateDraftPool` and `updatePublishedPool`, which can have very different validation.

apps/web/src/pages/Pools/ViewPoolPage/components/ChangeDateDialog.tsx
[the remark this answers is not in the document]
esizer: Okay, so I think I have this working a bit better now. I added some front end validation to try and prevent getting to the server. As well as, making the error codes more specific/explicit so we can display a better error message.

 - [update frontend validation for change date](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9945/commits/d699fa8d8ac51592dc6b31d17e71741968317070)
 - [update error codes to be more explicit](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9945/commits/a9a56428965412843079cdfabf0e9f88c27ae71c)

vd1992 (approved):
My concerns appear addressed and things are working. 👍 
Kicking it over to Peter now

apps/web/src/pages/Pools/ViewPoolPage/components/ChangeDateDialog.tsx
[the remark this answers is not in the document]
petertgiles: Seems like this is being discussed separately now.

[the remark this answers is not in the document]
petertgiles: Looks great!

petertgiles (approved):
Thanks for the fixups.  Looks great!

Issue #8668, opened 2023-12-04
♻️ `PoolCandidateSearchInput.priorityWeight` using `Int`


esizer:
## ♻️ Debt/Refactor

We currently use the `Int` type to filter pool candidates by priority which makes understanding the code a little difficult.

Priority weight is actually a generated column which is only supposed to be used for sorting, not filtering data.  Filtering should been done on the data fields themselves.

![image](https://github.com/GCTC-NTGC/gc-digital-talent/assets/8978655/e87c9312-3de2-4983-88da-4f012645a6d9)

## 🕵️ Details

Initially, this makes sense since we store them as `Int` in the database so passing them in makes the actual query easier to build (just use the value passed in). However, while developing it is not very intuitive to use. It would be nice if we could pass in the name of the priority status (entitlement, veteran, citizen) to make this more human readable.

## 🙋‍♀️ Proposed Solution

It would be nice if we maybe had an `enum` or something that mapped human readable values to the underlying weights. I'd love to be able to pass in:

```graphql
input PoolCandidateSearchInput {
  priorityWeight: [PRIORITY_WEIGHT] @scope(name: "priorityWeight")
}

query {
  poolCandidatesPaginated(where: {
    priorityWeight: [PRIORITY_ENTITELMENT, VETERAN, ...] 
  })
}
```

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the debt was addressed and expected functionality has not been affected.

- [ ] Priority filtering relies on the source data fields rather than the sorting field.

PR #9988, opened 2024-04-04
[Feature] Add candidate csv generation

### LINKED ISSUE #9687: Create Server-Side Pool Candidate CSV Report Generator
## ✨ Feature

We want to start generating the Pool Candidate CSV reports on the server.  Let's set up the view to do that.

## 🕵️ Details

This issue only deals with the CSV report itself.  How the client requests it and how it gets back to the client are out of scope for now.

## 🎨 Design File

The server-side report should match the current client-side report design.

## ✅ Acceptance Criteria

- [ ] It is possible to generate a CSV with one or more pool candidates on the server
  - [ ] A dev can trigger the request manually.  A regular user cannot.
  - [ ] No need to authenticate request yet
  - [ ] No need to save the file on the server, can just send back an HTTP response
- [ ] Report matches the report generated by the client

## 🛑 Blockers

```[tasklist]
### Blocked By
```

esizer:
🤖 Resolves #9687 

## 👋 Introduction

Adds ability to generate a csv of pool candidates on the server.

## 🧪 Testing

```php
// File generation commands for step 6
$file = new \App\Generators\CandidateProfileCsv([ 'uuid' ]);
$file->generate();
$file->save('candidates.csv');
```

2. Open the DB and find or create a pool candidate and get their ID (or multiple)
3. Open the maintencane container
4. `cd api`
5. Run tinker `php artisan tinker`
6. Create the file with the commands mentioned
7. Open the file (stored in `api/storage/app/user_generated/candidates.csv`)
8. Confirm it contains the candidates and matches the existing client side version of the CSV

## 🚚 Deployment

 - The `PhpOffice\PhpSpreadsheet` package needed the `php-gd` extension to be installed so this was added to the `post_deployment.sh` script. We should probably highlight this to change management


api/app/Models/User.php
```diff
@@ -475,6 +476,45 @@ public function getPositionDuration()
         return null;
     }
 
+    public function getPriority()
```
petertgiles, line 479: I'd recommend deriving this from the actual business data columns rather than the weight column.  Ref: https://github.com/GCTC-NTGC/gc-digital-talent/issues/8668

api/app/Generators/CandidateProfileCsv.php
```diff
@@ -0,0 +1,219 @@
...
+        $this->ids = $ids;
+        $this->lang = $lang;
+
+        parent::__construct();
+    }
+
+    public function generate()
+    {
+        $candidates = PoolCandidate::with([
+            'educationRequirementAwardExperiences',
```
petertgiles, line 72: On L145 and L191 you're using just educationRequirementExperiences and it looks like you're not using and of the  user->*experiences at all.

infrastructure/bin/post_deployment.sh
```diff
@@ -23,7 +23,7 @@ fi
 BLOCKS="{ \"type\": \"header\", \"text\": { \"type\": \"plain_text\", \"text\": \"Post-deployment script was run\" } }"
 
 # Install packages from repository
-if apt-get update && apt-get install -y supervisor cron; then
```
petertgiles, line 26: I deployed this to Azure.  No problems.  :+1:

petertgiles (changes requested):
Looking cool!  :sunglasses:

Unfortunately, I can't run this right now.

api/Dockerfile
```diff
@@ -15,6 +15,13 @@ FROM base as builder
 RUN apt-get update \
     && apt-get install -y libzip-dev \
     && docker-php-ext-install zip
+# Install gd ext required for document generation.
+# REF: https://hub.docker.com/_/php#:~:text=php%2Dsource%20delete-,PHP%20Core%20Extensions,-For%20example%2C%20if
+RUN apt-get update \
```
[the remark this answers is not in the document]
esizer: I may ask for your help if you can provide it. I did try and combine them but was getting errors. I can't remember exactly what errors but I'll try again and see :ok_hand:

api/app/Models/User.php
esizer: okay, so are you suggesting something like:

```php
$priority = [];

if($this->has_priority_entitlement) {
  $priority[] = 'Priority entitlement'; 
}

// etc.

return implode(', ', $priority);
```

api/app/Generators/CandidateProfileCsv.php
esizer: ugh, I gotta turn off copilot :face_exhaling: I fixed this like 3 times but every time I went to change this it kept auto-completing to this every time :angry: and since it is so close I rarely ever noticed it :cry:

api/Dockerfile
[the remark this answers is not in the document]
esizer: Awesome, seems to have worked this time! I must have had a typo or just did it wrong the first time :sweat_smile: 

[combine php ext install commands](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9988/commits/d9181fa90ca066a9a0a5589837e73775df152e69)

api/app/Generators/CandidateProfileCsv.php
esizer: Also noticed we removed the bilingual eval option so I fixed(?) that as well.

[fix attribute access errors](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9988/commits/8df48157d266a985e857c5f49ec93e290f1324f7)

api/app/Models/User.php
esizer: Updated in [update priority status column value](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9988/commits/5784d9472d15e0c5ab45cd50df48bea52c2d0122)

api/app/Generators/CandidateProfileCsv.php
petertgiles: This is working well now so I'll approve.  I still think the user->*experiences relationships don't need to be `with`'d at all.  And you have userSkills `with`'d twice.

petertgiles (approved):

api/app/Generators/CandidateProfileCsv.php
esizer: Oops, did not notice that :sweat_smile: I also didn't realize the education experience eager load wasn't doing anything. It was complaining before your merged all the morphs so I guess I just assumed it was still needed after :woman_shrugging: Anway! I removed both in [remove unnecessary eager loads](https://github.com/GCTC-NTGC/gc-digital-talent/pull/9988/commits/0622e3c7fca5de6a9566996a99a398174cdf166d) and it is still working :+1:

Issue #9866, opened 2024-03-21
✨ Restrict privileged api actions behind network firewall


tristan-orourke:
## ✨ Feature

We currently restrict admin pages behind the network firewall. However, the full API can be accessed from anywhere, and that is how any real interaction with the database happens. We want to be able to ensure that privileged roles and actions are restricted to government employees. The easiest way to do that is restrict those actions to the government network, which is already restricted to government employees.

## 🕵️ Details

The plan is to host a second API endpoint which behind the firewall. `/admin/graphql` would work. And then, anywhere we're checking for permissions, we somehow check which api endpoint is being used. If the request came from the public endpoint, then we only allow permissions which are available to Guest, Base, or Applicant users to pass authorization.

We also need to be very careful on any authorizedToView scopes.

We will also need to change the client-side admin pages to use the new admin endpoint.

Helpful documentation:
- https://laratrust.santigarcor.me/docs/8.x/usage/roles-and-permissions.html
- https://laravel.com/docs/11.x/requests#accessing-the-request

### How to check for allowed permissions/roles

From the public access point, we only want to consider permissions that come from Guest, Base User, or Applicant roles. I've been thinking how to implement this, and I think you'd have to check if the user has the permission (what already happens), and then additionally check if the Role has that permission, and the user has the role. It would look something like the following:

```
isAbleTo( $permission, $team, $requireAll) {
  $usingSecureApi= isUsingSecureApi() ; // Check if we're on the protected endpoint somehow
  if ($usingSecureApi) {
    return parent::isAbleTo($permission, $team, $requireAll);
  } else {
    return self::canUsePublicPermission($permission) && parent::isAbleTo($permission, $team, $requireAll);  // what order of checking is likely to be most performant?
  }
}

canUsePublicPermission($permission) {
  $publicRoles = Role::whereIn('key', ['guest', 'base_user', 'applicant'])->get();
  $user = Auth::user();
  foreach($publicRoles as $role) {
    if ($role->hasPermission($permission) && $user->hasRole($role)) { // what order of checking is likely to be most performant?
      return true;
    }
  }
  return false;
}

isUsingSecureApi(Request $request) { 
  // How do we get the request object? Can you inject the request into any method, or only route controller methods? Is there a facade for accessing it? https://laravel.com/docs/11.x/requests#accessing-the-request
  return $request->is('admin/*'); 
}
```

[...]

PR #10030, opened 2024-04-10
[Feature] Privileged api - backend only

petertgiles:
> [!Warning] 
The core functionality of this PR requires a manual change to test breaking features.  Manually back out the changes of [disable breaking changes](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10030/commits/8972a29259957da7c73f199ed888241e7f345f4b) to properly test the API changes.

🤖 Related to (but does not close) #9866

## 👋 Introduction

This branch adds a second route to getting to the graphql API and enforces using it for any operation requiring privileged permissions.  

## 🕵️ Details

It works by using custom routing for lighthouse to accept two different routes.  A middleware adds a request variable when it detects that the protected route is used.  And finally, a custom user checker ensures that Laratrust only allows the permission when the request variable is detected.

The `MakesGraphQLRequests` trait is unable to properly work when custom routing is enabled for lighthouse, like this PR uses.  The vast majority of the file diffs in this PR are just adding a trait to set which route to use in every test.

### Maintain Existing Functionality

- [ ] Building and running this branch "as-is" maintains current functionality without breaking anything.

### Add New Functionality

1. Back out the changes of [disable breaking changes](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10030/commits/8972a29259957da7c73f199ed888241e7f345f4b) to add new, breaking, functionality.
 
> [!TIP] 
Graphiql will only run on the regular, unprotected route.  To test the protected route you will have to use a different client like Altair, Postman, Edit&Resend, etc.

For the purpose of this PR: 
- a limited permission is one available to the base, guest, or applicant roles
- a privileged permission is one not available to the base, guest, or applicant roles
- the unprotected route is /graphql
- the protected route is /admin/graphql

- [ ] Any operation requiring a limited permissions only will succeed on either route
- [ ] Any operation requiring a privileged permission will fail on the uprotected route
- [ ] Any operation requiring a privileged permission can succeed on the protected route
- [ ] PHPUnit tests pass

### DIFF
--- a/api/app/Http/Middleware/ProtectedRequest.php
+++ b/api/app/Http/Middleware/ProtectedRequest.php
@@ -16,8 +16,7 @@ class ProtectedRequest
      */
     public function handle(Request $request, Closure $next)
     {
-        $path = $request->path();
-        if (str_starts_with($path, 'admin/')) {
+        if ($request->is('admin/*')) {
             $request->merge(['isProtectedRequest' => true]);
         }
 
--- a/api/app/Http/Middleware/ProtectedRequest.php
+++ b/api/app/Http/Middleware/ProtectedRequest.php
@@ -17,7 +17,7 @@ class ProtectedRequest
     public function handle(Request $request, Closure $next)
     {
         $path = $request->path();
-        if (str_starts_with($path, 'admin')) {
+        if (str_starts_with($path, 'admin/')) {
             $request->merge(['isProtectedRequest' => true]);
         }
 
--- a/api/app/Http/Middleware/ProtectedRequest.php
+++ b/api/app/Http/Middleware/ProtectedRequest.php
@@ -18,7 +18,7 @@ class ProtectedRequest
     {
         $path = $request->path();
         if (str_starts_with($path, 'admin')) {
-            $request->merge(['isSecureRequest' => true]);
+            $request->merge(['isProtectedRequest' => true]);
         }
 
         return $next($request);
--- a/api/app/Http/Middleware/ProtectedRequest.php
+++ b/api/app/Http/Middleware/ProtectedRequest.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace App\Http\Middleware;
+
+use Closure;
+use Illuminate\Http\Request;
+
+/**
+ * Notices if the request is a secure one that would have been
+ * protected by the VPN.
+ */
+class ProtectedRequest
+{
+    /**
+     * @return mixed Any kind of response
+     */
+    public function handle(Request $request, Closure $next)
+    {
+        $path = $request->path();
+        if (str_starts_with($path, 'admin')) {
+            $request->merge(['isSecureRequest' => true]);
+        }
+
+        return $next($request);
+    }
+}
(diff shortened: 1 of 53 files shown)


api/tests/Unit/ProtectedRequestUserCheckerTest.php
```diff
@@ -0,0 +1,97 @@
...
+        $privilegedPermission = 'create-any-classification';
+
+        assertTrue($checker->currentUserHasPermission($privilegedPermission));
+    }
+
+    public function testCanNotUsePrivilegedPermissionUnprotected()
+    {
+        $this->markTestSkipped('Enable to test the route protection logic.');
+
+        // simulate a protected request context
```
petertgiles, line 76:
```suggestion
        // simulate an unprotected request context
```

api/tests/Unit/ProtectedRequestUserCheckerTest.php
```diff
@@ -0,0 +1,97 @@
...
+
+        // a privileged permission can be used only in a protected request
+        $privilegedPermission = 'create-any-classification';
+
+        assertFalse($checker->currentUserHasPermission($privilegedPermission));
+    }
+
+    public function testCanNotUseUnknownPermissionUnprotected()
+    {
+        // simulate a protected request context
```
petertgiles, line 88:
```suggestion
        // simulate an unprotected request context
```

petertgiles (commented):
Copypasta to fix

api/app/Http/Middleware/ProtectedRequest.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+class ProtectedRequest
+{
+    /**
+     * @return mixed Any kind of response
+     */
+    public function handle(Request $request, Closure $next)
+    {
+        $path = $request->path();
+        if (str_starts_with($path, 'admin')) {
```
tristan-orourke, line 20: Maybe instead use `$request->is('admin/*')`. See https://laravel.com/docs/11.x/requests#inspecting-the-request-path. I'm just thinking about how our files like `adminPage.js` got flagged by the firewall 😅

tristan-orourke (changes requested):
Very cool! It looks good to me, and well tested. Just one minor thing about the url check which makes me nervous.

api/app/Http/Middleware/ProtectedRequest.php
petertgiles: Oh, very good catch!  That's a nice little matcher function.

tristan-orourke (changes requested):
I found two things!

1. **authorizedToView** scopes! We skip checking policies on queries which are protected aby `authorizedToView` and `authorizedToViewBasicInfo`. Those queries can now return privileged information on the limited graphql endpoint. The scopes also need to check for `isProtectedRequest` and handle it appropriately.
2. graphiql. I think it would be a good idea to copy the graphiql config file into our config folder. That would allow us attach it to the /admin/graphql endpoint if we want, for local devleopment.

tristan-orourke (approved):
Everything seems to work for me now! I must not have had the project built correctly when I thought I found issues with the scope.

PR #10048, opened 2024-04-12
[Fix] Responsive hero images

### LINKED ISSUE #9636: 🐛 Unstable Chromatic snapshots - Manager's Home Page, Executive's Home Page
## 🐛 Bug

The manager's home page and executive's home page for mobile have an unstable snapshots that are creating nuisance diffs in Chromatic.

## 📸 Screenshots

image

Screen Shot 2024-04-11 at 16 06 21

## ✅ Acceptance Criteria

- [ ] Stories create stable snapshots in Chromatic

esizer:
🤖 Resolves #9636 

## 👋 Introduction

Fixes an issue where the incorrect image was being displayed for the `<HomeHero>` component.

## 🕵️ Details

It seems as though there was an attempt at loading different images for specific media queries on the home page heroes. However, we used the `img[srcset]` attribute which is not quite correct.

The `srcset` attribute on `img` is for loading more compressed images with only the resolution changing, not the focus or crop of an image for mobile devices to reduce load time and bandwidth. Meaning, if you have already loaded the larger image and cached it, it will never load the image set for a smaller viewport.  I beleive this was causing an unstable snapshot because chromatic could be asynchronous in which story gets snapshotted first. So on some occasions it was loading the higher res one first so it never snapshotted the lower res one.

In order to fix this, we have migrated to use the `picture` element which can load entirely differently images for viewports and will always display those regardless of what you have in your local cache. This should allow chromatic to always take an accurate snapshot, stabilizing it in the process.

## 🧪 Testing

> [!IMPORTANT]
> In order to see how this works, it might be a good idea to repeat these tests on `talent.canada.ca` to see how it currently functions.

1. Build `pnpm run dev`
2. Load the homepage on a desktop viewport
3. Make note of the image that is loaded and displayed
4. Reduce the screen width slowly
5. Observe the image change as you reduce the width (if using the prod site, you will not see any changes)


packages/storybook-helpers/main.ts
```diff
@@ -7,7 +7,6 @@ const path = require("path");
 
 const HydrogenPlugin = require("hydrogen-webpack-plugin");
 const TsTransformer = require("@formatjs/ts-transformer");
-const PreloadWebpackPlugin = require("@vue/preload-webpack-plugin");
```
mnigh, line 10: You'll want to remove `vue/preload-webpack-plugin` from `/packages/storybook-helpers/package.json` and re-run `pnpm install` to regenerate the lockfile.

mnigh (changes requested):
Looks good, just one thing to address before it is ready for approval.

packages/storybook-helpers/main.ts
esizer: Weird, I totally did do this :confounded: Maybe a merge conflict resolution gone wrong :sweat_smile: 

Anyway, removed in [remove preload plugin deps](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10048/commits/f9b0744a099bff5708f31b2fddd96bea427db213)

mnigh (approved):
📸!

PR #10074, opened 2024-04-16
[Feature] Add "Other" Publishing group to the searchable list of groups

### LINKED ISSUE #10062: IT-05 not showing up in search page
## Details

The candidates for the IT-05 pilot were not showing up in the search page count. This appears to be because the "other" publishing group was used.

So, we can just add the "other" publishing group to the count and match queries.

The snippet in question:

Image

## Acceptance criteria

- [ ] Candidates qualified in "Other" publishing group processes are included in the count on the search page
- [ ] Candidates qualified in "Other" publishing group processes appear in matches for searches

brindasasi:
🤖 Resolves #10062 

## 👋 Introduction

This PR allows IT-05 group that is published under "OTHER" publishing group to be searchable via the Find Talent form along with existing IT_JOBS_ONGOING, IT-JOBS. 

Technically it will allow any group that is published under "Other" will be part of the count candidates query.
I'm unsure whether this will bite back us though. 

## 🕵️ Details
The scopeInITPublishingGroup has been modified to include "Other" publishing group.
However it is not ITPublishingGroup any more.
Should we rename this scope? 
Would like dev suggestion

## 🧪 Testing

1. Create a process with IT-05 and publish under "Other" group
3. Apply for the same process via `View Advertisement` and apply from there. ( This will NOT be shown on browse pools page) 
4. Go to the process and qualify yourself
5. Checkout the search form initial total estimated count should be increased by 1 and IT-05 should be visible at the bottom. 
6. Now checkout IT-05 is available for search in the classification drop down
7. Filter the candidate for IT-05 and submit the request
8. Go to Requests page and view the request. 

## 📸 Screenshot

Screenshot 2024-04-16 at 3 19 25 PM

Screenshot 2024-04-16 at 3 32 46 PM

### DIFF
--- a/api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
+++ b/api/app/GraphQL/Queries/CountPoolCandidatesByPool.php
@@ -32,8 +32,8 @@ final class CountPoolCandidatesByPool
         // available candidates scope (scope CANDIDATE_STATUS_QUALIFIED_AVAILABLE or CANDIDATE_STATUS_PLACED_CASUAL, or PLACED_TENTATIVE)
         PoolCandidate::scopeAvailable($queryBuilder);
 
-        // Only display IT candidates
-        PoolCandidate::scopeInITPublishingGroup($queryBuilder);
+        // Only display IT & OTHER publishing group candidates
+        PoolCandidate::scopeInTalentSearchablePublishingGroup($queryBuilder);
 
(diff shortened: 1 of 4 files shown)


petertgiles (approved):
Works great for me.  :+1:

Regarding: Should we rename this scope?
Yeah, I think so.  Something "scopeInTalentSearchablePublishingGroup"

petertgiles:
Nice, thanks for the update.  Works great.

PR #10083, opened 2024-04-17
[Feature] Refactor screening assessment dialog notes

### LINKED ISSUE #9900: ✨ Refactor notes on screening assessment dialogs
## ✨ Feature

This issue has two parts: 

1) The "Notes for this assessment" textarea is no longer required in the generic assessment modal because we made the "Decision notes" textarea always available and gave it some smart requirement logic.

We want to delete the "Notes for this assessment" text area from the generic assessment modal.

## Design

[Design file ](https://www.figma.com/file/em5ArPr044YaIe1Vb1zCZz/Admin---Processes%2FPools%2FRecruitments---Record-of-Decision?type=design&node-id=8738%3A32603&mode=design&t=UvdtFuhMYWPUDPox-1)by @Jerryescandon 

## Screenshot
Image

---------------------

2) Julie has been using the RoD feature, she likes the "other reason" notes field but wishes it could also be available during the other justification selections NOT JUST the "other reason..." selection.

It would be simpler if this field was just renamed to "Decision notes" and have it appear by default regardless of the type of decision (Demonstrated, Not demonstrated, etc...)

Have the field be OPTIONAL:
- If the decision is "Demonstrated"
- If the decision is "Not demonstrated (Disqualified)" AND the justification one or both of:
   - "Not sufficiently demonstrated to meet the requirement."
   - "Not enough information provided."

Have the field be REQUIRED:
- If the decision is "Not demonstrated (Disqualified)" AND the justification includes:
   - "Other reason for screening out." (even if other justifications are selected)
- If the decision is "Not demonstrated (Hold for further assessment)"

This change should apply to all type of assessments. Application, screening questions, generic assessment, etc.

## 🎨 Design File

Design file by @Jerryescandon 

### CURRENT

image

image

### PROPOSED

image

Question 1: How are we combining the "Notes for this assessment"? 
Question 2: "Other reason" is getting renamed to decision notes, is the data getting migrated to combine with decision notes as well.

```graphql
type AssessmentResult {
  id: ID!
  assessmentStep: AssessmentStep @belongsTo
  poolCandidate: PoolCandidate @belongsTo
  poolSkill: PoolSkill @belongsTo
  assessmentResultType: AssessmentResultType
    @rename(attribute: "assessment_result_type")
  assessmentDecision: AssessmentDecision
    @rename(attribute: "assessment_decision")
  justifications: [AssessmentResultJustification]
  otherJustificationNotes: String
    @rename(attribute: "other_justification_notes") # "Other reason"
  assessmentDecisionLevel: AssessmentDecisionLevel
    @rename(attribute: "assessment_decision_level")
  skillDecisionNotes: String @rename(attribute: "skill_decision_notes") # "Decision notes"
  assessmentNotes: String @rename(attribute: "assessment_notes") # "Notes for this assessment"
}
```

## ✅ Acceptance Criteria

- [ ] Delete/remove/destroy the "Notes for this assessment" and "other reason" textarea that appears at the top of the generic assessment modal.
- [ ] Migration to combine `otherJustificationNotes` and `assessmentNotes` notes into the "Decision notes" box discussed in so that no data is lost. When migrating data separate old data with label and notes. (eg. Notes for this assessment: "the previous notes") 
- [ ] "Other reason" text area renamed to "Decision notes"
- [ ] Decision notes text area is always available in this modal
- [ ] Decision notes is conditionally required based on logic above

yonikid15:
🤖 Resolves #9900 

## 👋 Introduction

- Removes `assessment_notes` and `other_justification_notes` from AssessmentResult model.
- Removes "Notes on assessment" and "Other reason" textareas from the ScreeningDecisionDialog.
- Removes all existence of `assessment_notes` and `other_justification_notes` on backend. 
- Creates a migration to move `assessment_notes` and `other_justification_notes` into `skill_decision_notes`. If the notes exist they should be combined in this structure:

```
Decision notes: ... (// hmm maybe I should remove this label🤔 Update: This label is removed)
Notes for this assessment: ...
Other justification notes: ...
```

## 🕵️ Details

Ensure you only execute the migration artisan command without seeding the data to ensure you don't wipe the previous notes fields that are getting dropped. All that requires is a simple `php artisan migrate` when switching to this branch.

## 🧪 Testing

1. Run new migration `php artisan migrate`
2. Build app `pnpm run dev`
4. Go to a process and view a pool candidate. Ensure the "Decision notes" text area has the combined the data correctly. (You might need to create a process that has the three types of assessments to ensure the note textareas were removed).


vd1992:
I am not 100% following the goal of this issue, but I think one last step of renaming `skill_decision_notes` to `decision_notes` ought to exist for better clarity

api/database/migrations/2024_04_10_200445_combine-assessment-result-note-fields.php
```diff
@@ -0,0 +1,37 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        DB::statement("UPDATE assessment_results SET skill_decision_notes = CONCAT(
```
petertgiles, line 15: I'm guessing most of the time not all these fields are filled and this will create a lot of unnecessary junk in the skill_decision_notes field.  I think it would be nice to only add these titles in if there's actually something in the other fields.  I'd be happy to help with the SQL if you want to try this.

api/database/migrations/2024_04_10_200445_combine-assessment-result-note-fields.php
```diff
@@ -0,0 +1,37 @@
...
+            $table->dropColumn('other_justification_notes');
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        Schema::table('pool_candidates', function (Blueprint $table) {
```
petertgiles, line 32: Oof, this is a bit tough.  This really doesn't revert the state to the way it was before but it would be tough to parse out the data from skill_decision_notes back to the original columns.  Probably not worth working it out unless we have to.  I guess this feature isn't widely rolled out yet, anyway.

apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.tsx
petertgiles, line 1: I think there's a "reset state issue" here.  If I open the dialog and choose "remove from process" the text box is not there.  If I switch to another option and come back it becomes visible.

https://github.com/GCTC-NTGC/gc-digital-talent/assets/8978655/a6653c42-d079-4bbb-80ea-4da8ea7cd15a

petertgiles (changes requested):
Very nice, the code is much simpler!  :relaxed:  I noted a few things to check on below.

api/database/migrations/2024_04_10_200445_combine-assessment-result-note-fields.php
yonikid15: Nice catch! Okay I separated the statements and added a where clause to check if it's not null. If you have a better solution I'm happy to try that as well. Fixed by [6c5e28f](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10083/commits/6c5e28fcf5f5c62f59ea6c2ea2e0e19881a76a3e)

apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.tsx
yonikid15: I pretty sure this bug already existed, but I added a reset so it shouldn't appear anymore 👍 Fixed by [3d3ab92](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10083/commits/3d3ab927dba69096d9308fd36e30aef3410cf93c)

api/database/migrations/2024_04_10_200445_combine-assessment-result-note-fields.php
petertgiles: Nice, thanks.  There's a small annoyance where if skill_decision_notes starts empty it gets an unnecessary newline when extra data is concatted in but I'm not too worried about that.

apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.tsx
petertgiles: Nice, thanks for fixing this bug!

petertgiles, line 1: This is something I'm unsure about, but from the issue it sounds like they want Decision Notes to be always visible - sometimes required and sometimes optional.  Right now though, it's hidden on Remove From Process for the first two justifications.  Shouldn't it be visible for that?

petertgiles (changes requested):
Looking great!  One last question from me.

apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.tsx
petertgiles: Maybe it's a question for @Jerryescandon

Jerryescandon: **Correct**, the idea is to make "Decision notes" always visible and it should switch between "required" and "optional" depending on the selected decision. AND the checkboxes too. 

If the decision is "Not demonstrated (Disqualified)" 
AND the justification checkbox includes **ONLY the first two otpions** of:
- "Not sufficiently demonstrated to meet the requirement."
- "Not enough information provided."
THEN have the field be OPTIONAL
   
If the decision is "Not demonstrated (Disqualified)"
AND the justification checkbox includes:
- "Other reason for screening out." 
(even if other justifications are selected)
THEN have the field be REQUIRED

In both cases, it should be visible.

yonikid15: Nice catch Peter thanks! Fixed by [a965b11](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10083/commits/a965b11ea5f3567edd0f2498414a1800ad8c8ae0)

yonikid15: Alright I demoed the dialog to @Jerryescandon and it looks good now👍

petertgiles: Nice, looks good to me, too.

PR #10109, opened 2024-04-17
[Tooling] Disables Dependabot from running Chromatic workflow

### LINKED ISSUE #10108: 🛠️ Disable chromatic for dependabot PRs
## 🛠️ Tooling

We don't really need dependabot to run chromatic  workflow since it is just bumps and it eats up time and snapshots.

## 🙋‍♀️ Proposed Solution

Update the workflow with the if statement to run the job.

## ✅ Acceptance Criteria

- [ ] Run dependabot and confirm chromatic does not run

mnigh:
🤖 Resolves #10108.

## 👋 Introduction

This PR adds a conditional for Chromatic workflow to only run if it is not a dependabot bot-created PR.

### Documentation
REF: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#fetch-metadata-about-a-pull-request

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -19,6 +19,8 @@ on:
 jobs:
   chromatic:
     name: Chromatic
+    # Only run if not dependabot,
+    if: github.actor != 'dependabot[bot]'
     runs-on: ubuntu-22.04
     env:
       PNPM_VERSION: "8.15.0"


esizer (approved):
:cool:

vd1992 (changes requested):
I'm not so sure on this one, we've had instances in the past of dependency updates breaking Chromatic

PR #10111, opened 2024-04-17
Highlight first visible section of ToC

### LINKED ISSUE #8029: ✨ Enhance the table of contents component to indicate current scroll position
## ✨ Summary

[Visit the Figma file]() • Chat with @substrae for design questions

Users have requested a quality of life improvement to the tables of contents that would provide a small visual indicator telling them where on the page they are. This will likely have to be coordinated using scroll position and heading position on the page.

```[tasklist]
### ✅ Requirements
- [ ] Add the functionality to highlight where on the page a user's viewport currently is
- [ ] Update the styles for the standard ToC element
- [ ] Update the styles for both states of the completion status ToC element
- [ ] Ensure the highlighted element is properly announced to ATs if the user is reading through the ToC
``` 

## 🔍 Details

<details>
  <summary>📸 Standard table of contents styles</summary>
  
</details>

<details>
  <summary>📸 Completion indicator table of contents style</summary>
  
</details>

tristan-orourke:
🤖 Resolves #8029

## 👋 Introduction

The table of contents highlights the link to the first visible section on the page. It highlights using bolded text and the [aria-current](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current) property.

## 🕵️ Details

This is accomplished with raw javascript which uses query selectors and attribute/classlist manipulation instead of react refs, to avoid having to modify the Table of Contents architecture. The important function is attached to a scroll event listener by the Navigation wrapper component.

## 🧪 Testing

Try a few pages with table of contents (like job posters and the applicant profile). Check the table of contents storybook.

### DIFF
--- a/packages/ui/src/components/TableOfContents/AnchorLink.tsx
+++ b/packages/ui/src/components/TableOfContents/AnchorLink.tsx
@@ -14,7 +14,15 @@ const AnchorLink = ({ id, children }: AnchorLinkProps) => {
   };
 
   return (
-    <ScrollToLink to={id} color="black" mode="text" onScrollTo={handleScrollTo}>
+    <ScrollToLink
+      data-is-toc-link // Used to find Table of Contents link elements in the Navigation component
+      id={`toc-link-for-${id}`}
+      data-h2-font-weight="base:selectors[.active](bold)"
+      to={id}
+      color="black"
+      mode="text"
+      onScrollTo={handleScrollTo}
+    >
       {children}
     </ScrollToLink>
   );
--- a/packages/ui/src/components/TableOfContents/Navigation.tsx
+++ b/packages/ui/src/components/TableOfContents/Navigation.tsx
@@ -8,6 +8,44 @@ import Sidebar from "../Sidebar/Sidebar";
 
 type NavigationProps = React.HTMLProps<HTMLDivElement>;
 
+const isInView = (element: HTMLElement) => {
+  const rect = element.getBoundingClientRect();
+  // top edge is in view, or bottom edge is in view, or both edges are out of view but element is partially in view
+  return (
+    (rect.top >= 0 && rect.top <= window.innerHeight) ||
+    (rect.bottom >= 0 && rect.bottom <= window.innerHeight) ||
+    (rect.top < 0 && rect.bottom > window.innerHeight)
+  );
+};
+
+const getNavSections = (): HTMLElement[] =>
+  Array.from(document.querySelectorAll("[data-is-toc-section]"));
+
+const getNavLinks = (): HTMLAnchorElement[] =>
+  Array.from(document.querySelectorAll("[data-is-toc-link]"));
+
+const getNavLinkForSection = (section: HTMLElement): HTMLAnchorElement | null =>
+  document.querySelector(`#toc-link-for-${section.id}`);
+
+const resetNavLinks = (): void => {
+  getNavLinks().forEach((link) => {
+    link.removeAttribute("aria-current");
+    link.classList.remove("active");
+  });
+};
+
+const highlightCurrentNavLink = (): void => {
+  resetNavLinks();
+  const currentSection = getNavSections().find(isInView);
+  if (currentSection) {
+    const currentLink = getNavLinkForSection(currentSection);
+    if (currentLink) {
+      currentLink.setAttribute("aria-current", "location");
+      currentLink.classList.add("active");
+    }
+  }
+};
+
 const Navigation = ({ children, ...rest }: NavigationProps) => {
   const intl = useIntl();
   const id = uniqueId();
@@ -15,6 +53,14 @@ const Navigation = ({ children, ...rest }: NavigationProps) => {
   const textAlignStyles = {
     "data-h2-text-align": "base(left)",
   };
+
+  React.useEffect(() => {
+    highlightCurrentNavLink(); // Make sure this runs when page is first initialized
+    document.addEventListener("scroll", highlightCurrentNavLink);
+    return () => {
+      document.removeEventListener("scroll", highlightCurrentNavLink);
+    };
+  }, []);
 
   return (
     <Sidebar {...textAlignStyles} {...rest}>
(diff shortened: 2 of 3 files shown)


tristan-orourke:
What do you think @esizer? If we don't insist on controlling everything the "React" way, its not that hard!

esizer:
> What do you think @esizer? If we don't insist on controlling everything the "React" way, its not that hard!

Yeah, this is great work and pretty close to how it can be done! :tada: 

Unfortunately, the hard part is deciding when to highlight the section :cry: As you can see, I have clearly scrolled past the previous section and the next section is in full view but the highlight has not moved with me.

screenshot_17-Apr-2024_17-47-1713390467

It also gets a bit weird when multiple sections are in view at the same time.

screenshot_17-Apr-2024_17-53-1713390787


You are pretty darn close though! I think we just need to tweak the `isInView` and try and decide when we want to highlight the associated link. Some thoughts (though very subjective)

1. If first section (none above) highlight as soon as it appears
2. Otherwise, highlight once it gets past the midway of the screen :thinking: 
3. If more than one is in view, highlight the second one (?) 

Something else we may want to consider but, I'm not sure if this applies to us, is if we need to support nested sections? I don't think we do need this but I could be wrong.

### `aria-current`

One thing I actually am not sure about is the use of `aria-current`. [Location](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current#values:~:text=Represents%20the%20current%20location%20within%20an%20environment%20or%20context%20such%20as%20the%20image%20that%20is%20visually%20highlighted%20as%20the%20current%20component%20of%20a%20flow%20chart.) seems like it could be correct but the way I interpret it, it may be safer to just avoid aria here. I'm kind of the mind set that if we aren't sure if we need `aria` it is safer to not use it.

tristan-orourke:
## On the isInView logic
The current logic is very simple: highlight the first section which is at all visible on the page. And the potential "damage" from awkward edge cases is so minor that I'm not too worried about getting this perfect. But yeah, we can probably do better. 
Eric's two screenshots show the two edge cases I'm aware of well:
- when a section is _barely_ in view, like the first screenshot 
- when the last section is small enough, it will never be the first section on screen. Especially if you click the ToC link to get to that section, you might expect it to be highlighted.

I'm not overly worried about either of those cases, but also happy to experiment. I'll give @esizer's rules a try and see if it feels better.

@substrae do you have any suggestions for the specific rules of highlighting a section?

tristan-orourke:
## on aria-current
> but the way I interpret it, it may be safer to just avoid aria here.

@esizer could you elaborate on how you interpret aria-current? Are you saying we should not have an AT equivalent to the visual highlighting of the ToC link? Would it be too noisy with little benefit?

tristan-orourke:
Fair. I suppose the highlighted link doesn't provide any additional information to a screen reader that the section header wouldn't already provide. 🤔

tristan-orourke:
Okay, I like this version more. It tries to find a section which covers the center of the screen, and if that doesn't work uses the one that's closest. I also removed aria-current.

tristan-orourke:
@esizer like this? https://github.com/GCTC-NTGC/gc-digital-talent/pull/10111/commits/185c765c334be6da7f3d4a42ca3949d4be7463a1

tristan-orourke:
Its a bit unrelated, but I've been using this new make command during development, and I thought it might be useful to push it up for others. I can revert if you disagree @esizer 

https://github.com/GCTC-NTGC/gc-digital-talent/pull/10111/commits/797bdef000239eb42e07cdde66b77e6e2acc385d

esizer:
> Its a bit unrelated, but I've been using this new make command during development, and I thought it might be useful to push it up for others. I can revert if you disagree @esizer

oh no, that is great. I'm glad someone is using the `Makefile` :sweat_smile: I kind of selfishly pushed it through for myself and wasn't sure if anyone used the or not :stuck_out_tongue:

esizer (approved):
LGTM! :+1:

tristan-orourke:
I'm using Make all the time now! I just recently switched from running pnpm commands natively to fully relying on `make refresh`. (Mostly because I kept using `npm` by mistake lol)

tristan-orourke:
Chromatic confirms this works with [nested sections](https://www.chromatic.com/component?appId=61e099a1bb1465003a73d3ce&csfId=components-table-of-contents&buildNumber=7081&k=6622cd13ef15c8b1379dfde8-1200px-interactive-true&h=8&b=-1) too! 

But it also shows how this strategy of picking the "current" section can feel a bit janky with [short sections](https://www.chromatic.com/component?appId=61e099a1bb1465003a73d3ce&csfId=components-user-profile--user-profile-story-1&buildNumber=7081&k=6622cd13ef15c8b1379dfcf7-1200px-interactive-true&h=6&b=-6). After clicking a link, if the section doesn't reach at least halfway down the page, the _next_ section is the one highlighted.

@esizer and @substrae, any further thoughts on the "current section" logic or should I stick with this for now?

substrae:
@tristan-orourke @esizer 

Ayy! This is looking great. There's a weird edge case for sections at the bottom of the page that refuse to highlight if you can't scroll them high enough... I'd also maybe recommend not transitioning the `font-weight`, as it's a bit slow and distracting.

tristan-orourke:
Hmmm. Maybe a good strategy would be...
- if there are any small sections which appear entirely on the screen, highlight the last one
- otherwise, highlight the section which covers the center of the screen (or closest to)

esizer:
> I'd also maybe recommend not transitioning the `font-weight`, as it's a bit slow and distracting.

@tristan-orourke So, to do this, you can do it in different ways. The ones I can think of are as follows:

## Only transition colours for all buttons

You would change all the `data-h2-transition` attributes in the [base button styles file](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/packages/ui/src/utils/button/getButtonBaseStyle.ts#L74).

Not exactly sure how to do this with hydrogen but the css would look like:

```css
button {
   transition-property: color, background-color, border-color;
   transition-duration: 200ms;
   transition-timing-function: ease;
}
```

## Just cancel out transition on TOC links

Should just need to add the followinjg to any links we bold  for TOC: `data-h2-transition="base(unset)"`

tristan-orourke:
Thanks @esizer, I've disabled transitions on just the ToC link for now.

@substrae I'll just wait on your approval now.

PR #10132, opened 2024-04-18
[Feature] Protected endpoint - frontend

### LINKED ISSUE #9866: ✨ Restrict privileged api actions behind network firewall

petertgiles:
🤖 Resolves #9866

## 👋 Introduction

This branch configures the frontend application to use the protected graphql endpoint when needed and allows the backend to block privileged requests on the unprotected endpoint.

## 🕵️ Details

This branch uses a new URQL exchange (middleware) to examine the current browser location and repoint the request to the protected endpoint when appropriate.

There is a feature flag, as well, that can be used to disable all the new functionality.

## 🧪 Testing

1. Copy new .env variables (frontend and backend) and rebuild the app.

- With the flag FEATURE_PROTECTED_API set to true
  - [ ] Requests from the admin site are routed to /admin/graphql
  - [ ] Requests from the non-admin site are routed to /graphql
  - [ ] Manually generated privileged requests to the /graphql endpoint are rejected as unauthorized
- With the flag FEATURE_PROTECTED_API set to false
  - [ ] All requests are routed to /graphql
  - [ ] Privileged requests to the /graphql endpoint can succeed

> [!TIP]
~This branch is currently deployed to the DEV vertical in Azure.~

## 🚚 Deployment

I've made these changes to DEV already, but double-check and make sure IaC is deployed to UAT and PROD verticals.

1. Add feature flag FEATURE_PROTECTED_API and have it set to false initially.
2. There is a new build-time variable as well API_PROTECTED_URI="/admin/graphql".


tristan-orourke (commented):
Wow, this code was super straightforward to read through, I love it

tristan-orourke (approved):
Seems to work as expected!

vd1992:
Setting the flags true and then building, I can still login to admin and the new endpoint for graphiql also works
No issues for me 👍

PR #9971, opened 2024-04-03
[Feature] Account settings page

yonikid15:
🤖 Resolves #9556 

## 👋 Introduction

- Adds account settings page

## 🕵️ Details

Add any additional details that could assist with reviewing or testing this PR.

## 🧪 Testing

Assist reviewers with steps they can take to test that the PR does what it says it does.

1. Go to http://localhost:8000/en/applicant/settings
2. Ensure page looks and functions correctly

PR #10138, opened 2024-04-18
[Feature] Promotional email set to opt in

### LINKED ISSUE #10086: ✨ "promotional" email notifications must be opt-in
## ✨ Feature

In #9555 I suggested that notification preferences would default to "accept all", and users would have to opt-out. It's been brought to my attention that "promotional" emails MUST require explicit consent, accordion to [legislation](https://ised-isde.canada.ca/site/canada-anti-spam-legislation/en/getting-consent-send-email). I think it makes sense to consider our "job alert" emails as a promotional notification, so we must ensure that notification type is ignored for new and existing users.

## 🕵️ Details

The `ignored_email_notifications` must include `JOB_ALERT` by default. This must be applied to all existing users (either with a migration, a script, or an interactive tinker session), and there must be code which applies it to all new users created after this PR.

In-app notifications can still default to On.

Reference legislation: https://ised-isde.canada.ca/site/canada-anti-spam-legislation/en/getting-consent-send-email

## ✅ Acceptance Criteria

- [ ] Any new users are created with JOB_ALERT added to `ignored_email_notifications`
- [ ] All existing users have JOB_ALERT added to `ignored_email_notifications` (this may require manual intervention during deployment)

## 🚀 Deployment

???

vd1992:
🤖 Resolves #10086

## 👋 Introduction

Sets JOB_ALERT as default ignored for emails and updates existing models to reflect that. 
Elected to do a migration as it covers both tasks in one go. 

## 🧪 Testing

1. Migrate back and forth, nothing broke
2. Create a new user, run `User::create()` in tinker for example
3. Query notification fields in `/graphiql`

## 🚚 Deployment

Nothing currently needs to be done

### DIFF
--- a/api/database/migrations/2024_04_18_210339_job_alert_default_ignored.php
+++ b/api/database/migrations/2024_04_18_210339_job_alert_default_ignored.php
@@ -1,0 +1,35 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\DB;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        DB::statement(
+            <<<'SQL'
+            ALTER TABLE users
+            ALTER COLUMN ignored_email_notifications
+            SET DEFAULT '["JOB_ALERT"]'::jsonb
+            SQL, []
+        );
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        DB::statement(
+            <<<'SQL'
+            ALTER TABLE users
+            ALTER COLUMN ignored_email_notifications
+            SET DEFAULT null
+            SQL, []
+        );
+    }
+};
--- a/api/database/migrations/2024_04_18_210339_job_alert_default_ignored.php
+++ b/api/database/migrations/2024_04_18_210339_job_alert_default_ignored.php
@@ -17,6 +17,10 @@ return new class extends Migration
             SET DEFAULT '["JOB_ALERT"]'::jsonb
             SQL, []
         );
+
+        DB::table('users')->update([
+            'ignored_email_notifications' => ['JOB_ALERT'],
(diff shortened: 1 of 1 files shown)


api/database/migrations/2024_04_18_210339_job_alert_default_ignored.php
```diff
@@ -0,0 +1,39 @@
...
+     */
+    public function up(): void
+    {
+        DB::statement(
+            <<<'SQL'
+            ALTER TABLE users
+            ALTER COLUMN ignored_email_notifications
+            SET DEFAULT '["JOB_ALERT"]'::jsonb
+            SQL, []
+        );
```
vd1992, line 19: Took a while to get something to work here 😭

api/database/migrations/2024_04_18_210339_job_alert_default_ignored.php
```diff
@@ -0,0 +1,39 @@
...
+
+        DB::table('users')->update([
+            'ignored_email_notifications' => ['JOB_ALERT'],
+        ]);
```
vd1992, line 23: Given no one should have anything updated from the default as the page for it isn't even merged yet https://github.com/GCTC-NTGC/gc-digital-talent/pull/9971
Can take the simple route?

tristan-orourke: This makes me squirm a bit, since it won't work for any future notification types we want to make default, but yes, this seems safe this time.

tristan-orourke (approved):
Yep, looks good, looks like the simplest possible implementation really.

api/database/migrations/2024_04_18_210339_job_alert_default_ignored.php
vd1992: Yeah, in the future we can't just rewrite the field for everyone 😛

PR #10141, opened 2024-04-19
[Fix] Pins storybook dependencies to version 8.0.6

mnigh:
## 👋 Introduction

This PR pins storybook dependencies to version 8.0.6 since version 8.0.8 seems to be somehow break the Chromatic workflow for `build-storybook`.

## 🧪 Testing

1. Verify Chromatic GitHub workflow builds storybook


mnigh:
Storybook was not the problem. Problem introduced with merging #10089. Fixed in #10142.

PR #10160, opened 2024-04-24
[Debt] Remove stored locale

### LINKED ISSUE #10145: ♻️ Remove `STORED_LOCALE` const
## ♻️ Refactor

Remove underused `STORED_LOCALE` const.

## 🕵️ Details

The `STORED_LOCALE` const is only ever imported in one place. Its value, "store_locale", is used elsewhere (two files) but not enough to warrant being imported.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/acdf8dfa8c40cc5af5f7e084a2a8b50eb57bd1df/packages/i18n/src/const.ts#L2

https://github.com/GCTC-NTGC/gc-digital-talent/blob/acdf8dfa8c40cc5af5f7e084a2a8b50eb57bd1df/packages/i18n/src/utils/localize.ts#L10

## 🙋‍♀️ Proposed Solution

- Remove `packages/i18n/src/const.ts` file
- Replace single reference to use "store_locale" string instead

## ✅ Acceptance Criteria

- [ ] `packages/i18n/src/const.ts` or references to it or its constants are no longer in repo

esizer:
🤖 Resolves #10145 

## 👋 Introduction

Removes a file for a const (`STORED_LOCALE`) that was only being used in one place.

## 🧪 Testing

1. Build `pnpm run dev`
2. Switch languages
3. Confirm the proper value is stored in local storage

### DIFF
--- a/packages/i18n/src/const.ts
+++ b/packages/i18n/src/const.ts
@@ -1,2 +1,0 @@
-// eslint-disable-next-line import/prefer-default-export
-export const STORED_LOCALE = "stored_locale";
--- a/packages/i18n/src/index.tsx
+++ b/packages/i18n/src/index.tsx
@@ -84,7 +84,6 @@ import {
   getPoolOpportunityLength,
 } from "./messages/localizedConstants";
 import getOrThrowError from "./utils/error";
-import { STORED_LOCALE } from "./const";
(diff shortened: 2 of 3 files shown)


vd1992 (approved):
👍

PR #10166, opened 2024-04-24
[Tooling] Jest always runs on merging

vd1992:
## 👋 Introduction

Jest does not always run upon merging a branch in, this prevents code coverage uploads. PHP Unit always runs so it always pushes up a new coverage report. 
As seen in devspam, this causes confusion with drastic coverage changes stemming from if either one or two workflows uploaded something

## 🕵️ Details

Just having Jest always run for merges. Shouldn't really run that much more often, but clear up the occasional confusion.

### DIFF
--- a/.github/workflows/jest.yml
+++ b/.github/workflows/jest.yml
@@ -3,10 +3,6 @@ name: Jest
 on:
   push:
     branches: [main]
-    paths:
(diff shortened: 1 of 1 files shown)


brindasasi (approved):
yeah I agree with that 👍

PR #10212, opened 2024-04-29
[Debt] Bump Azure Ubuntu to 22.04

### LINKED ISSUE #10202: ♻️ Bump build pipeline to use Ubuntu 22.04 image
## ♻️ Debt/Refactor

Our build pipeline is still using 20.04 images.  Those should probably be bumped 22.04 to line up with our maintenance container and GH Action workflows.

image

> [!IMPORTANT]
More complicated than first seems per #10212. May need to find a workaround using docs: https://learn.microsoft.com/en-us/azure/devops/pipelines/?view=azure-devops

## ✅ Acceptance Criteria

- [ ] Azure build pipelines set to 22.04

## 🛑 Blockers

```[tasklist]
### Blocked By
```

vd1992:
🤖 Resolves #10202 

## 👋 Introduction

Bumps the Azure pipelines version of Ubuntu to v22 to be kept in line with all other environments and instances. 

## 🧪 Testing

1. Test deployment and build I assume

## 🚚 Deployment

?

FAIL 💀

### DIFF
--- a/infrastructure/azure-pipelines-dev.yml
+++ b/infrastructure/azure-pipelines-dev.yml
@@ -13,7 +13,7 @@ jobs:
   - job: build_artifact
     displayName: Build artifact
     pool:
-      vmImage: ubuntu-20.04
+      vmImage: ubuntu-22.04
     steps:
       - checkout: self
         clean: true
--- a/infrastructure/azure-pipelines.yml
+++ b/infrastructure/azure-pipelines.yml
@@ -20,7 +20,7 @@ jobs:
   - job: build_artifact
     displayName: Build artifact
     pool:
-      vmImage: ubuntu-20.04
+      vmImage: ubuntu-22.04
     steps:
       - checkout: self
         clean: true
--- a/infrastructure/bin/set_php_versions.sh
+++ b/infrastructure/bin/set_php_versions.sh
@@ -10,6 +10,12 @@ set -o nounset
 # Assign PHP version from arg if supplied. Runs default PHP bin without.
 PHP_VERSION=$1
 
+sudo update-alternatives --install /usr/bin/php php /usr/bin/php${PHP_VERSION} 100
+sudo update-alternatives --install /usr/bin/phar phar /usr/bin/php${PHP_VERSION} 100
+sudo update-alternatives --install /usr/bin/phpdbg phpdbg /usr/bin/php${PHP_VERSION} 100
+sudo update-alternatives --install /usr/bin/php-cgi php-cgi /usr/bin/php${PHP_VERSION} 100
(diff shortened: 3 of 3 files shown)


vd1992:
I think I'll need your help to test the deploy/build please @brindasasi 😄

petertgiles (changes requested):
Uh oh, looks like you might have some more to do on this one.  I have no idea what this means - some googling I guess.  😁 
image

petertgiles:
Maybe PHP 8.2 isn't available in this image?  The later image has less options for PHP?  Very strange...
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2004-Readme.md#php-tools
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#php-tools

petertgiles:
OK, I found out more about it.
https://github.com/actions/runner-images/issues/6331
It looks like they removed any non-official PHP sources in the image.  Since Ubuntu 22.04 came with PHP 8.1 that's the only one available.

In the linked page they recommend using setup-php to get other versions of PHP.  Like we already do in our GH workflows, I guess.

vd1992:
No luck 🤕

mnigh:
> What are we going to do when this image becomes deprecated?

@petertgiles Great question, not sure there is much we can do if the new trend continues for runners where they only include support for one version of PHP. The newest Ubuntu runner, `[Ubuntu 24.04](https://github.com/actions/runner-images/issues/9848)`, only supports 8.3 which would mean we would be forced to upgrade the runner **and** the PHP version we use at the exact same time. Also, it would mean that we would be stuck with both until a new runner version is released. 

Using `setup-php` works in our GitHub workflows because it is a GitHub action, but there isn't an equivalent task (as they are known in Azure Pipelines) for Azure Pipelines, or at least that I could find. My guess that the [documentation](https://learn.microsoft.com/en-us/azure/devops/pipelines/ecosystems/php?view=azure-devops#use-a-specific-php-version) that shows how to set a different PHP version was relevant for runners that had multiple versions of PHP installed (for example, ubuntu-20.04) but not for anything >=22.04.

petertgiles:
> Using `setup-php` works in our GitHub workflows because it is a GitHub action

Yeah, I was looking at the `setup-php` action to see how it works and I think [it's just using the ondrej PPA](https://github.com/search?q=repo%3Asetup-php%2Fsetup-php%20ondrej&type=code) for Ubuntu.  It makes me a little nervous to use a third party PPA to build our software but of course we're using this in an action already and it's not an obscure source.  I think we'll just end up having to add this to our pipeline.

Kboxoxx (approved):

PR #10215, opened 2024-04-30
[Debt] Removes `Card` component

### LINKED ISSUE #10156: ♻️ `Card` component not in use
## ♻️ Debt/Refactor

We have `Card` and `CardBasic`  components that are nearly identical in real world usage.

## 🕵️ Details

The main difference is `Card` has a slot for title but, we never actually use that prop. There is also an identical component as a one off in `EducationRequirements`.

## 🙋‍♀️ Proposed Solution

Delete `Card` and rename `CardBasic` to just `Card` and update imports/props.

## ✅ Acceptance Criteria

- [ ] `Card` component removed
- [ ] Import statements updated
- [ ] `title=""` instances removed
- [ ] `EducationRequirements` refactored to use new `Card`

esizer:
🤖 Resolves #10156 

## 👋 Introduction

Replaces the `Card` component with the `CardBasic` component.

## 🕵️ Details

We were only using `Card` in one place and it wasn't doing much so it made sense to remove it.

## 🧪 Testing

1. Confirm `Card` is removed
2. Confirm chromatic snapshot diff is acceptable


mnigh (commented):
LGTM.

One question before approving: if we are getting rid of `Card` component and its storybook story, should we have one for the `CardBasic` component so we don't lose snapshot coverage?

esizer:
> One question before approving: if we are getting rid of `Card` component and its storybook story, should we have one for the `CardBasic` component so we don't lose snapshot coverage?

Added in  https://github.com/GCTC-NTGC/gc-digital-talent/pull/10215/commits/a88b46d6b5c50e07898fb4da51505333b33142e9

mnigh (approved):
:shipit:

Issue #9901, opened 2024-03-26
♻️ Remove snapshot workaround: general & screening questions


petertgiles:
> [!NOTE]
Blocked until Feb 20, 2027 (three years after workaround added)

## ♻️ Debt/Refactor

A workaround was added to the snapshot to allow pre-ROD screening questions in a snapshot to be viewed.  One year later, let's remove the workaround.

## 🕵️ Details

🐛 Snapshot Screening Questions Not Visible in pre-ROD page
[#9414](https://github.com/GCTC-NTGC/gc-digital-talent/issues/9414)
And https://github.com/GCTC-NTGC/gc-digital-talent/pull/9070/commits/2e3560f242c03d077460af966116784cd276d9dc

## ✅ Acceptance Criteria

- [ ] Workaround removed
- [ ] Current snapshots still render screening questions and general questions properly.

mnigh:
@petertgiles can this issue be iceboxed?

petertgiles:
Discussion in standup: sometimes recruitment team reviews year-old pools.  Gray to ask team about what a good horizon would be.

mnigh:
> Discussion in standup: sometimes recruitment team reviews year-old pools. Gray to ask team about what a good horizon would be.

@gobyrne has a horizon been decided upon?

gobyrne:
3 years was the request from recruitment team

(1 year to apply and get qualified, then 2 years to expire in the pool)

PR #10224, opened 2024-04-30
[Debt] Remove deprecated pool skills queries

### LINKED ISSUE #10151: ♻️ Replace deprecated fields `pool.essentialSkills` and `pool.nonEssentialSkills`
## ♻️ Debt/Refactor

The fields have been deprecated in favour of `pool.poolSkills` and then filtering that collection off `PoolSkillType` 
Removing from schema, remove obsolete backend code, and update frontend queries

## 📸 Screenshots

Non-exhaustive example of things that will need replacing

image

Example of utils added in #9503 to leverage

image

## ✅ Acceptance Criteria

- [ ] Deprecated fields destroyed
- [ ] Remove Eloquent relations made unneeded
- [ ] Replaced everywhere needed, so app functions unchanged

esizer:
🤖 Resolves #10151 

## 👋 Introduction

Removes the deprecated skills fields on `Pool` and replaces it with the new field.

## 🕵️ Details

Also added in some filters on the field so we can avoid over fetching and filtering out certain skills on the frontend.

## 🧪 Testing

1. Build `pnpm run dev`
2. Confirm skills are still queried and presented properly


apps/web/src/utils/skillUtils.ts
```diff
@@ -330,11 +330,31 @@ export const sortPoolSkillsBySkillCategory = <T extends PoolSkill[]>(
...
+    [PoolSkillType.Essential]: filterPoolSkillsByType(
+      nonEmptyPoolSkills,
+      PoolSkillType.Essential,
+    ),
+    [PoolSkillType.Nonessential]: filterPoolSkillsByType(
+      nonEmptyPoolSkills,
+      PoolSkillType.Nonessential,
+    ),
+  };
+}
```
tristan-orourke, line 356: What do you think about this alternative? Benefits:
- if the number of skill types ever expanded (unlikely, I know) this would still work
- subjective, but I've come to prefer Map objects when doing this kind of grouping. I think it helps indicate the purpose of the return value. I know this would require some minor re-writes elsewhere.
- this filters out any Maybes and undefineds
```suggestion
export function groupPoolSkillByType(
  poolSkills?: Maybe<Maybe<PoolSkill>[]>,
): Map<PoolSkillType, Array<Skill>> {
  return unpackMaybes(poolSkills).reduce((map, poolSkill) => {
    const { type, skill } = poolSkill;
    if (type && skill) {
      if (!map.has(type)) {
        map.set(type, []);
      }
      map.get(type)?.push(skill);
    }
    return map;
  }, new Map<PoolSkillType, Array<Skill>>());
}
```

apps/web/src/components/QualifiedRecruitmentCard/QualifiedRecruitmentCard.tsx
```diff
@@ -51,7 +50,7 @@ const QualifiedRecruitmentCard = ({
 
   // NOTE: Until we store assessed skills, we will just be displayed all essential skills
   const categorizedSkills = categorizeSkill(
-    candidate.pool.essentialSkills?.filter(notEmpty) ?? [],
+    poolSkillsToSkills(candidate.pool.poolSkills),
```
tristan-orourke, line 53: It was essentialSkills before, now its all poolSkills, is that okay?

apps/web/src/pages/Applications/ApplicationSkillsPage/ApplicationSkillsPage.stories.tsx
```diff
@@ -41,7 +44,13 @@ const hasExperiencesProps: ApplicationSkillsProps = {
     },
     pool: {
       ...fakePoolCandidate.pool,
-      essentialSkills: [...experienceSkills],
+      poolSkills: experienceSkills.map((skill) => ({
+        id: faker.string.uuid(),
+        ...faker.helpers.arrayElement(
+          fakePoolCandidate?.pool?.poolSkills ?? [],
+        ),
```
tristan-orourke, line 51: Are we not worried about adding the same skill multiple times here? I guess it doesn't really matter for a story?

apps/web/src/utils/skillUtils.ts
```diff
@@ -330,11 +330,31 @@ export const sortPoolSkillsBySkillCategory = <T extends PoolSkill[]>(
  * @returns Skill[]
  */
 export const filterPoolSkillsByType = (
-  poolSkills: PoolSkill[],
+  poolSkills: Maybe<Maybe<PoolSkill>[]> | undefined,
```
tristan-orourke, line 333: This is such an ugly type 🤣

tristan-orourke (changes requested):
Looks great. Just have a couple minor questions and minor suggestion.

apps/web/src/components/QualifiedRecruitmentCard/QualifiedRecruitmentCard.tsx
vd1992: I think we should keep it as before and narrow it to essential skills

apps/web/src/utils/skillUtils.ts
vd1992: I've seen worse, had to implement this as a workaround. Will be gone with https://github.com/GCTC-NTGC/gc-digital-talent/issues/9901

image

apps/web/src/components/QualifiedRecruitmentCard/QualifiedRecruitmentCard.tsx
esizer: So, this is where it gets tricky :sweat_smile: It technically is still essential because that is what we are querying. 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/e942c8e4a8caf92ded9d13c263eeceb403ac23a7/apps/web/src/pages/Profile/CareerTimelineAndRecruitmentPage/components/CareerTimelineAndRecruitment.tsx#L136

However, what I wanted to do was do a fragment mask for this component so we don't need to rely on the query.

Though, that makes things weird since this is a pool candidate thing that is getting masked in the other PR. But, I get where you are both coming from, it could be far more explict. I'm struggling where to do this work though, here or  in the other PR :thinking:

apps/web/src/pages/Applications/ApplicationSkillsPage/ApplicationSkillsPage.stories.tsx
esizer: Yeah, I'm not sure. I just did my best to show as much as possible in the story. I personally don't care if there are duplicates but can change things if others would like to see unique skills here.

apps/web/src/utils/skillUtils.ts
esizer: Yeah :sweat_smile: The more I work on these queries and masking the more I want to start re-writing everything to use the masksing. I really like how it basically writes your types for you :sunglasses:

apps/web/src/components/QualifiedRecruitmentCard/QualifiedRecruitmentCard.tsx
tristan-orourke: I think my suggestion for now would be to run this through `filterPoolSkillsByType`, even if we know it should only be essential from the query. That makes the code a bit clearer, and guards against changes to the query. And leave a comment explaining we're being deliberately redundant, so future us doesn't think it's just an oversight.

apps/web/src/pages/Applications/ApplicationSkillsPage/ApplicationSkillsPage.stories.tsx
tristan-orourke: Eh, seems fine, not worth the effort to change.

apps/web/src/components/QualifiedRecruitmentCard/QualifiedRecruitmentCard.tsx
esizer: okay, updated in [update to be explicit about essential skills](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10224/commits/c65ff37f5357b484f6e8627b2719e06432210fba)

PR #10237, opened 2024-05-01
[Bug] Fragments missing fields

### LINKED ISSUE #10234: 🐛 Missing fields from Pool fragments
## 🐛 Bug

Seems some places are missing fields

See assessment plan completion contrast in 

image

AND

image

## 📋 Steps to Reproduce

1. Prepare an assessment plan and attempt to publish 

## 📸 Screenshot

Function needs `pool.assessmentSteps.poolSkills`, but it isn't queried for

image

image

image

## ✅ Acceptance Criteria

- [ ] Criteria 1
- [ ] Criteria 2

vd1992:
🤖 Resolves #10234

## 👋 Introduction

## 🕵️ Details

## 🧪 Testing

1. ...
2. ...

## 📸 Screenshot

### DIFF
--- a/apps/web/src/pages/Pools/ViewPoolPage/fragment.ts
+++ b/apps/web/src/pages/Pools/ViewPoolPage/fragment.ts
@@ -21,9 +21,15 @@ const ViewPool_Fragment = graphql(/* GraphQL */ `
     }
     poolSkills {
       id
+      type
     }
     assessmentSteps {
       id
+      type
+      poolSkills {
+        id
+        type
+      }
     }
     poolCandidates {
       id
--- a/apps/web/src/pages/Pools/ViewPoolPage/fragment.ts
+++ b/apps/web/src/pages/Pools/ViewPoolPage/fragment.ts
@@ -1,0 +1,41 @@
+import { graphql } from "@gc-digital-talent/graphql";
+
+const ViewPool_Fragment = graphql(/* GraphQL */ `
+  fragment ViewPool on Pool {
+    id
+    publishingGroup
+    publishedAt
+    isComplete
+    status
+    closingDate
+    processNumber
+    stream
+    classification {
+      id
+      group
+      level
+    }
+    name {
+      en
+      fr
+    }
+    poolSkills {
+      id
+    }
+    assessmentSteps {
+      id
+    }
+    poolCandidates {
+      id
+      pool {
+        id
+      }
+      user {
+        id
+        email
+      }
+    }
+  }
+`);
+
+export default ViewPool_Fragment;
--- a/apps/web/src/validators/pool/assessmentPlan.ts
+++ b/apps/web/src/validators/pool/assessmentPlan.ts
@@ -10,11 +10,24 @@ import {
 
 import { PoolCompleteness } from "~/types/pool";
 import ViewPool_Fragment from "~/pages/Pools/ViewPoolPage/fragment";
+import AssessmentPlanBuilderPool_Fragment from "~/pages/Pools/AssessmentPlanBuilderPage/fragment";
+
+function isViewPool_Fragment(
+  poolQuery:
+    | FragmentType<typeof ViewPool_Fragment>
+    | FragmentType<typeof AssessmentPlanBuilderPool_Fragment>,
+): poolQuery is FragmentType<typeof ViewPool_Fragment> {
+  return "ViewPoolFragment" in poolQuery;
+}
 
 export function getAssessmentPlanStatus(
-  poolQuery: FragmentType<typeof ViewPool_Fragment>,
+  poolQuery:
+    | FragmentType<typeof ViewPool_Fragment>
+    | FragmentType<typeof AssessmentPlanBuilderPool_Fragment>,
 ): PoolCompleteness {
-  const pool = getFragment(ViewPool_Fragment, poolQuery);
+  const pool = isViewPool_Fragment(poolQuery)
+    ? getFragment(ViewPool_Fragment, poolQuery)
+    : getFragment(AssessmentPlanBuilderPool_Fragment, poolQuery);
 
   if (!pool || !pool.poolSkills || !pool.assessmentSteps) {
     return "incomplete";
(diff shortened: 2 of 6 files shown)


apps/web/src/pages/Pools/ViewPoolPage/fragment.ts
```diff
@@ -21,9 +21,15 @@ const ViewPool_Fragment = graphql(/* GraphQL */ `
...
       id
+      type
     }
     assessmentSteps {
       id
+      type
+      poolSkills {
+        id
+        type
+      }
```
vd1992, line 32: These fields added as a resolve of errors shown by fragmenting

yonikid15:
@vd1992 I'm sure it was an accident 😆 but can you fill out the PR introduction, testing, etc above

apps/web/src/validators/pool/assessmentPlan.ts
```diff
@@ -3,13 +3,32 @@ import { parseDateTimeUtc } from "@gc-digital-talent/date-helpers";
...
+}
+
+export function getAssessmentPlanStatus(
+  poolQuery:
+    | FragmentType<typeof ViewPool_Fragment>
+    | FragmentType<typeof AssessmentPlanBuilderPool_Fragment>,
+): PoolCompleteness {
+  const pool = isViewPool_Fragment(poolQuery)
+    ? getFragment(ViewPool_Fragment, poolQuery)
+    : getFragment(AssessmentPlanBuilderPool_Fragment, poolQuery);
```
esizer, line 30: So, while this works I beleive we decided to be more explicit about specific items we pass in rather than pass around fragments or full model types. So I'd probably change this function signature to look more like:


```tsx
type GetAssessmentPlanStatusArgs = {
   poolSkills: Maybe<Array<Maybe<Pick<PoolSkill, "id" | "type">>> | undefined;
   addessmentSteps: Maybe<Array<Maybe<Pick<AssessmentStep, "step">>> | undefined;
}

type GetAssessmentPlanStatusFunc = (args: GetAssessmentPlanStatusArgs): PoolCompleteness;
```

Or if that doesn't work, you could just define a fragment for this function:

```tsx
const AssessmentPlanStatus_Fragment = graphql(/* GraphQL */`
    fragment AssessmentPlanStatus on Pool {
        poolSkills {
          id
          type
        }
        assessmentSteps {
          step {
             poolSkills { id }
          }
        }
    }
`);

function getAssessmentPlanStatus(poolQuery: FragmentType<AssessmentPlanStatus_Fragment>) {
  const pool = getFragment(AssessmentPlanStatus_Fragment, poolQuery);
}
```

esizer (changes requested):
Thanks for finding this :pray: Just added some comments based on our team discussion this morning.

apps/web/src/validators/pool/assessmentPlan.ts
vd1992: Second recommendation does not seem to work for me as it complains about a fragment type mismatch

image

image

image

vd1992: First recommendation also doesn't seem to work for me 😕 

If I define the function like so with nullable `publishedAt`
I get this error

image

image

If I define it so it isn't nullable
I get this error instead

image

image

esizer: For the first one, I tihnk you may need to pass it in like `getAssessmentPlanStatus(pool);` and not the fragment itself. I assume that is where the issue is coming from.

On the second one, I think the type for `publishedAt` is just a tiny bit off. Could you try: `pulbishedAt?: Maybe<string>;`?

vd1992: Adding the question marks stopped errors. But now it does not warn about missing fields, like assessmentSteps not being queried at all.
I do not think mixing typescript typing with fragments is fruitful

image

image

image

vd1992: > For the first one, I tihnk you may need to pass it in like `getAssessmentPlanStatus(pool);` and not the fragment itself. I assume that is where the issue is coming from.
> 
> On the second one, I think the type for `publishedAt` is just a tiny bit off. Could you try: `pulbishedAt?: Maybe<string>;`?

That doesn't seem to work either 😢 

image

esizer: Yeah, I think we should be refactoring this to not pass the entire pool in to the function. We should only be passing in what the function needs to work. So something like:

```tsx
const assessmentStatus = getAssessmentStatus(
  pool.publishedAt,
  unpackMaybes(pool.poolSkills),
  unpackMaybes(pool.assessmentSteps)
);
```

That way we see the error here when passing items in and it requires them to exist. Now, this does not solve the issue with sub-fields. But, we could take it a step further and be very explicit about our types :thinking: but that seems overly verbose :sweat_smile: 

You will see, I removed the `assessmentSteps` entirely and it was able to pick up the error with the basic type. Then, if I removed the `PoolSkill.type` from the fragment and typed it explicitly, it was able to warn about a sub field missing.

screenshot_02-May-2024_15-45-1714679140

vd1992:
@esizer 

Okay, I got something working! One of the suggestions worked once I fixed a mistake of mine. 
See https://github.com/GCTC-NTGC/gc-digital-talent/pull/10237/commits/8027a590eabaaa3ebb1f44166736cd467b0ba70a

I like this approach, makes sure fields are present.

vd1992:
Don't need the separate fragment files in that case
https://github.com/GCTC-NTGC/gc-digital-talent/pull/10237/commits/b5fd322c6a493bd680ba6865ee4e91f5bbdbe444

apps/web/src/validators/pool/assessmentPlan.ts
vd1992, line 1: In the case of complex functions, ones with many sub-fields, define a fragment for type safety

apps/web/src/utils/poolUtils.tsx
vd1992, line 1: In the case of functions acting on minimal fields, specify fields rather than type the entire nebulous object

vd1992: Either with picking or the specific field like below

image

vd1992: Then if missing, you will see it, but not as involved as creating a fragment for a single field which would get really verbose

image

image

vd1992: This way, you defeat the `Maybe<>`

esizer (approved):
Nice! :sunglasses:

PR #10265, opened 2024-05-02
[Debt] Fix Nginx Warnings

### LINKED ISSUE #7155: 🐛 Nginx static analysis warnings
## 🐛 Bug

The Nginx Amplify static analysis tools raised a few warnings about ways we can improve our configuration.

### Missing default.conf from /etc/nginx/conf.d

The default.conf file is missing from /etc/nginx/conf.d directory. A new nginx package may automatically add default.conf without a warning. Since your nginx.conf contains the [include](http://nginx.org/en/docs/ngx_core_module.html#include) directive for conf.d/*.conf files, a new default.conf will be automatically read by nginx. This may result in an unpredictable nginx configuration.

Create an empty default.conf file in /etc/nginx/conf.d (or copy the original default.conf from the standard nginx package).

### Missing server_name in a virtual server configuration

When nginx accepts a request, the first step is to [determine](http://nginx.org/en/docs/http/request_processing.html) which virtual server the request should be routed to. Without a [server_name](http://nginx.org/en/docs/http/ngx_http_core_module.html#server_name) directive set for a virtual server, the request may hit the wrong one.

Always configure a unique [server_name](http://nginx.org/en/docs/http/ngx_http_core_module.html#server_name) directive for a [server](http://nginx.org/en/docs/http/ngx_http_core_module.html#server). Make sure the [server_name](http://nginx.org/en/docs/http/ngx_http_core_module.html#server_name) contains all possible site names.
Check the following file:
/etc/nginx/sites-enabled/default, line 13
Regex location has no regex pattern

### Regex [location](http://nginx.org/en/docs/http/ngx_http_core_module.html#location) has a valid modifier, but does not have a regex pattern.

Performance-wise it is more efficient to configure exact or prefix matching for locations that do not require regex matching. It is also less prone to errors. Please refer to the documentation describing [location](http://nginx.org/en/docs/http/ngx_http_core_module.html#location) directive to learn more.
Check the following file:
/etc/nginx/sites-enabled/default, line 66

### Regex location has no anchors in regex pattern

Regex [location](http://nginx.org/en/docs/http/ngx_http_core_module.html#location) has a regex pattern without ^ or $. This is a configuration style that is prone to errors. It may also lead to a situation when requests partially matching the regex pattern are incorrectly routed to this location.

Always use ^ or $ achor in a regex pattern.
Check the following file:
/etc/nginx/sites-enabled/default, line 66

## 📋 Steps to Reproduce

1. Log into amplify.nginx.com
2. Navigate to Analyzer \ Static Analysis

## ✅ Acceptance Criteria

- [ ] No warnings from static analysis tool

petertgiles:
🤖 Resolves #7155 

## 👋 Introduction

This fixes the config issues from the Nginx Amplify static analysis tool.

## 🕵️ Details

1. To fix the "missing a default.conf file" warning I added a `touch` command to the post-deploy script to create an empty file.
2. To fix the "missing a server_name directive" warning I added a substitution from an Azure variable that holds the domain name of the site.
3. To fix the two warnings about the git rule, I deleted it. :axe:  We no longer send the entire workspace to the app service so there's no way a real .git directory will be present.  If someone wants to see our git repo they can go to GH to get it.  :grin: 

## 🧪 Testing

1. Site can be built and visited locally.
2. Site can be built and visited in Azure.

## 📸 Screenshot

image

### DIFF
--- a/infrastructure/bin/post_deployment.sh
+++ b/infrastructure/bin/post_deployment.sh
@@ -85,10 +85,15 @@ else
 fi
 
 # Copy nginx config and reload
-if /home/site/wwwroot/infrastructure/bin/substitute_file.sh /home/site/wwwroot/infrastructure/conf/nginx-conf-deploy/default /etc/nginx/sites-available/default '$NGINX_PORT $ROBOTS_FILENAME' && nginx -s reload ; then
-    add_section_block ":white_check_mark: Config copy for Nginx *successful*."
+if
+    touch /etc/nginx/conf.d/default.conf && \
+    /home/site/wwwroot/infrastructure/bin/substitute_file.sh \
+        /home/site/wwwroot/infrastructure/conf/nginx-conf-deploy/default \
+        /etc/nginx/sites-available/default '$NGINX_PORT $ROBOTS_FILENAME' && \
+    nginx -s reload ; then
+    add_section_block ":white_check_mark: Set up Nginx *successful*."
 else
-    add_section_block ":X: Config copy for Nginx *failed*. $MENTION"
+    add_section_block ":X: Set up Nginx *failed*. $MENTION"
 fi
 
 # Copy custom PHP-FPM config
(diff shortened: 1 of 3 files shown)


brindasasi (approved):
Thanks for helping to review Peter! This looks great! 🥳

PR #10266, opened 2024-05-03
[Debt] Consolidate functions for determining candidate status

### LINKED ISSUE #10203: ♻️ Different functions for determining if candidate is qualified
## ♻️ Debt/Refactor

We have two, nearly identical, functions to determine if a pool candidate has a status that is considered qualified.

## 🕵️ Details

We have a constants file for pool candidates that contains arrays of different statuses that are used in the pool candidate utils in small functions to check if the pool candidate is in one of those statuses `is*`.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/e45e0cbec07433d66b44880908e9c5da0339db3a/apps/web/src/utils/poolCandidate.ts#L51-L53

https://github.com/GCTC-NTGC/gc-digital-talent/blob/e45e0cbec07433d66b44880908e9c5da0339db3a/apps/web/src/constants/poolCandidate.ts#L3-L11

As well, we have different helper functions in an application utils file that are nearly the same.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/e45e0cbec07433d66b44880908e9c5da0339db3a/apps/web/src/utils/applicationUtils.ts#L181-L192

## 🙋‍♀️ Proposed Solution

These functions are not exactly the same so we may need to figure out which one is the correct one and then update to only use that one and remove the other.

## ✅ Acceptance Criteria

- [ ] One of the two helper functions removed
- [ ] Only one way to determine if a status for a pool candidate is qualified

tristan-orourke:
🤖 Resolves #10203

## 👋 Introduction

Consolidates a couple functions which did essentially the same thing - determining the high-level candidate status.

## 🕵️ Details

Also added some comments explaining the difference between two utility files, ostensibly both related to applications.

Note: Applications with the REMOVED status are entirely hidden from applicants! I don't think that's right, but I'll create a new ticket to deal with it.

## 🧪 Testing

1. Log in as an Applicant
2. Go to http://localhost:8000/en/applicant/profile-and-applications
3. Ensure you have at least one application
4. Modify the status of the application (using database, admin side, or tinker) and ensure it continues to look right

The following is just a cautious smoke test. I don't think anything should have changed there:
1. Log in as an Admin
2. Look through the Pool Candidates table. Ensure the Final Decision badge looks right for everyone.

## 📸 Screenshot

image


esizer (approved):
LGTM! :+1:

PR #10267, opened 2024-05-03
[Debt] Pool candidate fragment masking

### LINKED ISSUE #9709: ♻️ Implement `PoolCandidate` fragment masking
## ♻️ Debt/Refactor

Update components and queries that access data from `PoolCandidate` to use fragment masking.

## 🕵️ Details

See the epic for more details:

 - https://github.com/GCTC-NTGC/gc-digital-talent/issues/9543

## ✅ Acceptance Criteria

- [ ] Queries that request data from the `PoolCandidate` type are using fragments
- [ ] Components that access data from the `PoolCandidate` type are refactored to use fragment masking

esizer:
🤖 Resolves #9709 

## 👋 Introduction

Adds fragment masking to queries using `PoolCandidate` model.

## 🕵️ Details

This does not fragment mask the assessment results tracker because we plan on reworking that soon and it does alot with the `PoolCandidate` model. In the interim, there was some light type casting added for the bookmarks.

## 🧪 Testing

1. Build `pnpm run dev`
2. Navigate to the pages that use pool candidates
3. Confirm the proper data is queried and displayed


vd1992:
Oh, appear to have a broken story

https://www.chromatic.com/review?appId=61e099a1bb1465003a73d3ce&number=10267&type=linked&view=changes

image

esizer:
> fix storybook args

Thankfully chromatic was working :stuck_out_tongue_winking_eye: Fixed in [fix storybook args](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10267/commits/1ba2fa27325aa3905f283c3c48d606da525ad2e9)

vd1992:
Would light typecasting some functions be a good addition? To build on the fragmentation. 

Like this change to a function

image

So a missing field like this

image

So it is flagged like so

image

Reverting the function changes...
Without that change to the function, the missing field is only caught at that level. So in this case, it is caught due to this bit

image

But not at the function that `submittedSteps` is missing

image

vd1992:
Thoughts on the above? Not demanding it, just putting it out there. 

Not too many places where this would be applicable, another example might be this function `candidateNameCell`
Separating the fields out of the `PoolCandidate` 

image

Used here

image

esizer:
> Thoughts on the above? Not demanding it, just putting it out there.

I think this is a good idea. I wasn't too sure if I was going to try and do this in this PR or see if we can do it in later tickets. I was worried about exploding the size of the PR and going too nuts with it though :sweat_smile: 

Do you think it would be a good idea to maybe get this last one in and then create a ticket to rework our function types to better utilize the masking since we kind of did the others like that as well and they will need to be revisited? :thinking:

apps/web/src/pages/Profile/CareerTimelineAndRecruitmentPage/components/CareerTimelineAndRecruitment.tsx
```diff
@@ -100,6 +100,7 @@ export const CareerTimelineExperience_Fragment = graphql(/* GraphQL */ `
 
 export const CareerTimelineApplication_Fragment = graphql(/* GraphQL */ `
   fragment CareerTimelineApplication on PoolCandidate {
+    ...QualifiedRecruitmentsCandidate
     id
     status
     archivedAt
```
vd1992, line 106: Oh, appears something broke

image

Looking through queries, with all the nested fragments, got a same field defined different ways. 

image

image

image

vd1992 (changes requested):

apps/web/src/pages/Profile/CareerTimelineAndRecruitmentPage/components/CareerTimelineAndRecruitment.tsx
esizer: Looks like come code snuck in with main since there was no conflict :woman_facepalming: Removed in [remove unused code](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10267/commits/705a088c24a44d18860efcae54a97f7228bce092)

vd1992 (approved):
> I think this is a good idea. I wasn't too sure if I was going to try and do this in this PR or see if we can do it in later tickets. I was worried about exploding the size of the PR and going too nuts with it though 😅
> 
> Do you think it would be a good idea to maybe get this last one in and then create a ticket to rework our function types to better utilize the masking since we kind of did the others like that as well and they will need to be revisited? 🤔

Okay, sounds good to me. 
I will let you create the issue and define its scope as you see fit.

Otherwise, seems to work well. Nice!
Noticed an unrelated bug while at it here https://github.com/GCTC-NTGC/gc-digital-talent/issues/10335

PR #10268, opened 2024-05-03
[Debt] Improve pool candidate count query

### LINKED ISSUE #10231: ♻️ Improve `ViewPoolPage_Query` performance
## ♻️ Debt/Refactor

The page to view a process could perform a bit better, specifically the operation `ViewPoolPage_Query`

## 🙋‍♀️ Proposed Solution

It appears we query a lot of data for a simple length check. If this is the case then a simple performance improvement would be to fetch the count in a simpler way. Either paginator -> paginatorInfo -> totalCount or an accessor virtual property that returns the total

Pagination would be through 
`PoolCandidateSearchInput.ApplicantFilterInput.pools`

I think the accessor might be best as you won't need to run the separate pagination query. 
Along the lines of 
`return count(PoolCandidate::select('id')->where('pool_id', $POOL_ID)->get());`

image

image

Link from Eric:
https://lighthouse-php.com/6/api-reference/directives.html#count

## ✅ Acceptance Criteria

- [ ] Page and query performance improved 📈 

## 🛑 Blockers

```[tasklist]
### Blocked By
- [ ] ticket number
```

esizer:
🤖 Resolves #10231 

## 👋 Introduction

Updates the candidate count query to use a new field that counts the rows of non-draft pool candidates on a pool.

## 🧪 Testing

1. Build `pnpm run dev`
3. Navigate to the main "Process information" page for a published pool with candidates
4. Confirm the total candidates count matches the non-draft pool candidates in the pool

## 📸 Screenshot

screenshot_03-May-2024_14-16-1714760187

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -183,7 +183,6 @@ type Pool {
     @guard
   poolCandidatesCount: Int
     @count(relation: "poolCandidates", scopes: ["notDraft"])
-    @guard
   keyTasks: LocalizedString @rename(attribute: "key_tasks")
   yourImpact: LocalizedString @rename(attribute: "your_impact")
   whatToExpect: LocalizedString @rename(attribute: "what_to_expect")
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -181,6 +181,9 @@ type Pool {
     @hasMany(relation: "publishedPoolCandidates", scopes: ["authorizedToView"])
     @canResolved(ability: "view")
     @guard
+  poolCandidatesCount: Int
+    @count(relation: "poolCandidates", scopes: ["notDraft"])
+    @guard
   keyTasks: LocalizedString @rename(attribute: "key_tasks")
   yourImpact: LocalizedString @rename(attribute: "your_impact")
   whatToExpect: LocalizedString @rename(attribute: "what_to_expect")
--- a/api/storage/app/lighthouse-schema.graphql
+++ b/api/storage/app/lighthouse-schema.graphql
@@ -328,6 +328,7 @@ type Pool {
   classification: Classification
   operationalRequirements: [OperationalRequirement]
   poolCandidates: [PoolCandidate]
+  poolCandidatesCount: Int
   keyTasks: LocalizedString
   yourImpact: LocalizedString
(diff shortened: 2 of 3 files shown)


vd1992:
I don't quite recall the decision from scrum, but I think it was desired to experiment with `poolCandidatePaginated` too?
@tristan-orourke

esizer:
> I don't quite recall the decision from scrum, but I think it was desired to experiment with `poolCandidatePaginated` too? @tristan-orourke

We could and I still can try that but, this is a sub-field of `Pool` so that would mean updating these to use paginated queries which would require either,

 - a new paginated version of the filed just to get a count or, 
 - modifying the existing field which would mean going and updating all queries that get pool candidates off of pool.

tristan-orourke:
Very good points from @esizer, this seems like a good solution. Let's just confirm who's allowed to see the number of candidates who have applied, or if it can be public.

vd1992:
In the meantime, testing it out with roughly 1000 pool candidates 
Operation `ViewPoolPage`

Main
Average time ~ 6.44 seconds

Branch
Average time ~ 1.33 seconds

Well that is about an 80% reduction in time!

api/graphql/schema.graphql
```diff
@@ -181,6 +181,9 @@ type Pool {
     @hasMany(relation: "publishedPoolCandidates", scopes: ["authorizedToView"])
     @canResolved(ability: "view")
     @guard
+  poolCandidatesCount: Int
+    @count(relation: "poolCandidates", scopes: ["notDraft"])
+    @guard
```
tristan-orourke, line 186: Designers are fine with this being public info, so we can even remove the `@guard` directive!

vd1992: I was probably a bit wrong on the slowness, I think policy checking probably also had a big part in cratering performance

image

For the purposes of counting, probably overkill to policy check each candidate of a possible thousand(s).
Important when doing things other than counting

esizer: Yeah, policy checks really slow down list queries since it needs to check each entity individually :cry:

vd1992 (approved):
Looks great to me, fantastic improvement. I prefer this solution over parallel queries anyway so good to me
🎉

PR #10273, opened 2024-05-03
Do not hide REMOVED poolCandidates from Applicant dashboard

### LINKED ISSUE #10272: 🐛 'REMOVED' applications don't appear on applicant dashboard
## 🐛 Bug

Any PoolCandidates with status of REMOVED do not appear anywhere on the applicant dashboard. Other removed-type statuses (eg SCREENED_OUT_NOT_RESPONSIVE) do still appear, with a "Removed" chip.

## 🦋 Expected Behaviour

I think the REMOVED status applications should appear as well. Otherwise, from the point of view of the applicant, the application just vanishes with no further context.

## 🕵️ Details

Talking to @Jerryescandon, it seems like we originally added the REMOVED status with the intention of it acting like a soft-delete. But now, we've adopted the language of "removed" to refer to all candidates who are taken out of a pool without necessarily being disqualified. 

And whatever the original intention, the REMOVED status _is_ being used, even on applications to a poster which is still live as this ticket is created. I'm worried that hiding the card from applicants is confusing, and that we should show them along with the other "removed" applications.

## 📋 Steps to Reproduce

1. ...
2. ...

## 📸 Screenshot

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] Criteria 1
- [ ] Criteria 2

tristan-orourke:
🤖 Resolves #10272

## 👋 Introduction

Allow applicants to see their applications even if they've been given the REMOVED status.

## 🕵️ Details

This was a deliberate choice, but I don't think it makes sense any longer.

## 🧪 Testing

1. Log in as an Applicant. Apply to at least one job, if you have no existing applications.
2. Update the application status to REMOVED (using adminer, or tinker, or admin user)
3. Go to http://localhost:8000/en/applicant/profile-and-applications and see if it appears in "past applications"

## 📸 Screenshot


vd1992 (approved):
Looks good!

image


image

Jerryescandon:
Any chance we can use the "black" colours on the chip as opposed to the red?

tristan-orourke:
@Jerryescandon hero you go
image

PR #10278, opened 2024-05-06
[Bug] Pool location validation

### LINKED ISSUE #10229: 🐛 Pool Incompleteness - when backend validator fails on a missing value front end doesn't show it
## 🐛 Bug

1. Create a process
2. For the Location field, enter English value and leave the French value empty
3. Enter all other required details
4. Observe Process stays as "Incomplete" without any red prompt

https://github.com/GCTC-NTGC/gc-digital-talent/assets/89482910/dfd4e37b-a411-45d8-85a7-13731bf39a9b

This pool can't be published and I had to go through the code to find out what I was missing

## 🦋 Expected Behaviour
Location for French should be shown on the Process page so that it can be given as red prompt when it is empty. 

## 📸 Screenshot

Image
Image

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] When the Location is left empty it should be shown in red prompt in UI as "Not Provided" like other fields

yonikid15:
🤖 Resolves #10229 

## 👋 Introduction

- Adds validation rule to ensure both values must be included to save changes.
- Adds adds specific location label for english and french values.

## 🧪 Testing

1. Login as `admin@test.com`
2. Create a new process
3. Ensure Core requirements section displays location fields correctly.
4. Ensure pool status at top of page is correct.

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/lang/fr.json
+++ b/apps/web/src/lang/fr.json
@@ -3599,6 +3599,10 @@
     "defaultMessage": "Êtes-vous certain de vouloir supprimer cette expérience de votre profil? Cette action ne peut être annulée.",
     "description": "Question displayed when a user attempts to delete an experience from their profile"
   },
+  "IhaKpH": {
+    "defaultMessage": "Lieu (anglais)",
+    "description": "Label for a process' english location requirement"
+  },
   "IjBtl5": {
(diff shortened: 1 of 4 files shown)


petertgiles (approved):
Works great!

PR #10279, opened 2024-05-06
[Feature] Batch candidate query for screening and evaluation

### LINKED ISSUE #10223: ✨ Use paginated query on Screening and Assessment page
## ✨ Feature

The screening and assessment page currently uses the `Pool.poolCandidates` field to fetch all candidates in a single query. We're aware that the number of candidates, and complexity of the PoolCandidate model, could the query to take a long time, and in worst case, crash. Also, we want to add filters to the page, which would further complicate the query.

The solution is to use pagination to load a manageable number of candidates at a time. 

## 🕵️ Details

Convert the ScreeningAndEvaluationPage.tsx component to use the `poolCandidatesPaginated` query. It should fetch candidates 100 at a time, and continue until all candidates are loaded. Don't show the results candidates until loading is complete. We have a new loading skeleton to show until then.

## 🎨 Design File

[Figma file](https://www.figma.com/file/em5ArPr044YaIe1Vb1zCZz/Admin---Processes%2FPools%2FRecruitments---Record-of-Decision?type=design&node-id=8738%3A34452&mode=design&t=DFf6UMQtRFQF0ln7-1) by @Jerryescandon 

## 📸 Screenshot

image

Repeatedly call the paginated query to load chunks of the results, until complete. Let's say 100 at a time, if that seems reliable?

QUESTION: should we modify the Pool type in the graphql schema to make it impossible to try to load all candidates without pagination?

## 🌎 Localization

?

## ✅ Acceptance Criteria

- [ ] Instead of trying to load all candidates at once, the Screening and Assessment page uses the poolCandidatesPaginated query
- [ ] The page displays a loading skeleton until all results are loaded

esizer:
🤖 Resolves #10223 

## 👋 Introduction

Updates the screening and evaluation page to use a batched, paginated query to prevent memory errors on large data sets.

## 🕵️ Details

## 🧪 Testing

> [!TIP]
> To test this effectively, it is best to seed a lot of candidates for a specific pool.
> `PoolCandidate::factory()->create(['pool_id' => '{poolId}']);`

1. Build `pnpm run dev`
2. Navigate to screening and decision page for a pool
3. Confirm the columns show loading message at first
4. Confirm data is batched in sizes of 100

### DIFF
--- a/apps/web/src/components/AssessmentStepTracker/AssessmentStepTracker.tsx
+++ b/apps/web/src/components/AssessmentStepTracker/AssessmentStepTracker.tsx
@@ -1,14 +1,10 @@
 import React from "react";
-import { IntlShape, useIntl } from "react-intl";
+import { useIntl } from "react-intl";
 
 import { Board, Link, Well } from "@gc-digital-talent/ui";
-import {
-  commonMessages,
-  getAssessmentStepType,
-  getLocalizedName,
-} from "@gc-digital-talent/i18n";
+import { commonMessages } from "@gc-digital-talent/i18n";
 import { unpackMaybes } from "@gc-digital-talent/helpers";
-import { AssessmentStep, Pool } from "@gc-digital-talent/graphql";
+import { FragmentType, getFragment, graphql } from "@gc-digital-talent/graphql";
 
 import applicationMessages from "~/messages/applicationMessages";
 import useRoutes from "~/hooks/useRoutes";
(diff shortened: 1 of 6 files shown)


vd1992:
I'm getting a discrepancy in candidate count. 

Main

image

Branch

image

vd1992 (changes requested):
Yeah, some candidates appear missing in the branch

image

image

esizer:
> Yeah, some candidates appear missing in the branch

oops, looks like pages are not 0 indexed so I was missing the last page :sweat_smile: Fixed in [add missing last page in batch loader](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10279/commits/3867f1beae31d2db8b106acb6e1cb306478e72ed)

apps/web/src/components/AssessmentStepTracker/AssessmentStepTracker.tsx
```diff
@@ -22,34 +18,88 @@ import {
...
+  const pool = getFragment(AssessmentStepTracker_PoolFragment, poolQuery);
+  const steps = unpackMaybes(pool?.assessmentSteps);
+  const poolCandidates = getFragment(
+    AssessmentStepTracker_CandidateFragment,
+    candidateQuery,
+  );
+  const candidates = unpackMaybes(poolCandidates ? [...poolCandidates] : []);
   // to handle partially completed processes, filter out candidates assessed and disqualified pre-RoD
   const filteredCandidates = filterAlreadyDisqualified(candidates);
   const groupedSteps = groupPoolCandidatesByStep(steps, filteredCandidates);
```
vd1992, line 105: I'd like to adjust the typing of some of these functions to be more definite to go along with your fragmenting addition. As this is some complicated stuff acting on quite a few fields, it is hard to catch a missed field in queries. 

image

Anything complicated but worth it can be spun off. 
I was thinking for example, the `filterAlreadyDisqualified` function as seen above could be narrowed to be type safe.

Instead of the whole PoolCandidate object, it just acts on the status field and that candidates had their assessment results fetched. Type safe would then like 
```
{
status: PoolCandidateStatus; 
assessmentResults: AssessmentResult [];
} []
```
I think. So a place with the lite version of typing we liked. 

image

vd1992 (changes requested):

vd1992:
Candidate count appears to be resolved. Testing performance...

Seeded a massive amount of data (Over 1.4 MB of JSON data)
Crashes/fails due to timeout limit of 30 seconds

Switch to branch and the page loads, eventually. Reliably. 

Data/batches for reference 😛 

image

image

apps/web/src/components/AssessmentStepTracker/AssessmentStepTracker.tsx
esizer: So, while this is a good suggestion. We discussed and I believe a lot of this code will be shifted to the server so I'm not sure how worth adjusting the types are right now. I think we plan on creating a query where you can do:

```graphql
query {
  pool(id: $id) {
     assessmentSteps {
       poolCandidates {
          ...AssessmentStepTrackerCandidate
       }
     }
  }
}
```

At the same time, I believe we will be moving a lot of the business logic (`filterAlreadyDisqualified`, `groupPoolCandidatesByStep`) onto the server instead. That way we only get the candidates that "matter".

esizer:
> Switch to branch and the page loads, eventually. Reliably.

"eventually" :laughing: Good to know!

apps/web/src/components/AssessmentStepTracker/AssessmentStepTracker.tsx
```diff
@@ -75,15 +125,27 @@ const AssessmentStepTracker = ({ pool }: AssessmentStepTrackerProps) => {
...
+                {fetching ? (
+                  <Well fontSize="caption" data-h2-margin="base(0 x.5)">
+                    <div
+                      data-h2-display="base(flex)"
+                      data-h2-align-items="base(center)"
+                    >
+                      <SpinnerIcon data-h2-width="base(x.75)" />
+                      <span>{intl.formatMessage(commonMessages.loading)}</span>
+                    </div>
+                  </Well>
```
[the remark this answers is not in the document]
esizer: Good catch! Added in [add top margin to loading well](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10279/commits/cfa31bc1e3c76999768049ce8d15e0b5de762313)

screenshot_07-May-2024_07-59-1715083194

apps/web/src/components/AssessmentStepTracker/AssessmentStepTracker.tsx
vd1992: Sure, if you think it will be unneeded, fine by me

vd1992 (approved):
Seems to work just fine for me, fantastic! 🏆 
Not sure what is up with Chromatic, storybook runs without issue for me locally though

Issue #9206, opened 2024-02-05
♻️ Subtitute environment variables into index.html on post-deploy


petertgiles:
## ♻️ Debt/Refactor

Let's run an `envsubst` pass on our index.html file on post-deploy to substitute environment variables in.  This will allow us to:
1) Get rid of the config.js file and one network round-trip during app startup.
2) Insert vertical-customized URLs into the head for things like Open Graph URLs.

## 🕵️ Details

Current config.js deployment: https://github.com/GCTC-NTGC/gc-digital-talent/blob/c97149e361df11313767f153f06c0535b01b1112/infrastructure/bin/post_deployment.sh#L71

## 🙋‍♀️ Proposed Solution

When the app is deployed it exists on a read-only mount.  Therefore, the substitution will need to copy the file to a writable location and the nginx conf will need to be able to find it in that location.

We probably don't want to run `envsubt` when building the app locally.  Can we get webpack to emit two versions of index.html?  One with environment variables already substituted and one that can be used as a template during deployment?

## ✅ Acceptance Criteria

- [ ] When deployed, runtime variables are substituted into the index.html file
- [ ] When deployed, vertical-specific head variables are substituted into the index.html file
- [ ] The app can still be easily built and run locally

petertgiles:
Most (but not all) of the scope will be completed by #10611

tristan-orourke:
#10611 gets rid of config.js for runtime env variables. What's left?

petertgiles:
It's just the opengraph URLs now, I think.  I've just trimmed down this issue to reflect that.  Due to the scope change, I've cleared the estimate of 3 and rolled it back to the estimation column.

PR #10285, opened 2024-05-06
[Debt] Allow `__typename` underscore dangle

### LINKED ISSUE #10206: ♻️ Add exception to no-underscore-dangle rule for __typename
## ♻️ Debt/Refactor

We add a lot of exceptions to the no-underscore-dangle rule to access the the built-in graphql __typename property.  We should probably just add an exception to the rule for it.

https://eslint.org/docs/latest/rules/no-underscore-dangle

## 🙋‍♀️ Proposed Solution

image

## ✅ Acceptance Criteria

- [ ] Add an exception to the eslint settings
- [ ] Remove all the exceptions that become unnecessary

## 🛑 Blockers

```[tasklist]
### Blocked By

```

esizer:
🤖 Resolves #10206 

## 👋 Introduction

Adds  `__typename` to list of allowed underscore dangle rule in eslint. 

## 🕵️ Details

This comes from the API and we need to use it some places so instead of ignoring the rule in all those places, we are just going to add it to the allow list.

## 🧪 Testing

1. Confirm `pnpm run lint` runs with no errors

### DIFF
--- a/apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.stories.tsx
+++ b/apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.stories.tsx
@@ -38,7 +38,6 @@ const profileSnapshot: User = {
     {
       ...poolCandidate,
       educationRequirementOption:
-        // eslint-disable-next-line no-underscore-dangle
         experience.__typename === "EducationExperience"
           ? EducationRequirementOption.Education
           : EducationRequirementOption.AppliedWork,
--- a/apps/web/src/components/SearchRequestFilters/SearchRequestFilters.tsx
+++ b/apps/web/src/components/SearchRequestFilters/SearchRequestFilters.tsx
@@ -281,7 +281,6 @@ const SearchRequestFilters = ({
 }: SearchRequestFiltersProps) => {
   const intl = useIntl();
   let poolCandidateFilter;
-  // eslint-disable-next-line no-underscore-dangle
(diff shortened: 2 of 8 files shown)


vd1992:
Not a common one, but perhaps one other place meriting a lint exclusion?

image

esizer:
> Not a common one, but perhaps one other place meriting a lint exclusion?

Yeah, I didn't do this one because this is the only place we do this and I think we are planning on trying to get rid of it in the future... maybe? @petertgiles Can you confirm my suspicion on this? :sweat_smile:

mnigh (approved):
✅ Lint exception works for typename dangling underscores
✅ Rule enforced on other dangling underscores

## Screenshot
Screen Shot 2024-05-06 at 14 42 13

petertgiles:
> Yeah, I didn't do this one because this is the only place we do this and I think we are planning on trying to get rid of it in the future

I don't remember how we were going to get rid of it, but yeah, it's only a single place so no point adding it to the rules.

mnigh:
> > Yeah, I didn't do this one because this is the only place we do this and I think we are planning on trying to get rid of it in the future
> 
> I don't remember how we were going to get rid of it, but yeah, it's only a single place so no point adding it to the rules.

@petertgiles Maybe #9206?

petertgiles:
> Maybe #9206?

Ooh, good idea, me-from-three-months-ago!  :laughing:

PR #10287, opened 2024-05-06
[Fix] Allow sorting candidates table by skill count

### LINKED ISSUE #10235: ✨ Add sorting to skill match column
## ✨ Feature

As a user, I find the skill match useful (inside individual requests, and on candidate tables) but I have no way of organizing the tables in order of Most matches to Least matches. I would like to be able to click on the column title the same way I order any of the other columns and have the list be reorganized based on this.

## 🕵️ Details

Add sorting to "Skills" column on all tables that display "Candidates" e.g. Candidates table, Inside individual requests, as well as "Talent placement" tab inside a process.

## 🎨 Design File

### 🧑‍🎨 Designer

## 📸 Screenshot

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Criteria 1
- [ ] Criteria 2

## 🛑 Blockers

```[tasklist]
### Blocked By
- [ ] ticket number
```

esizer:
🤖 Resolves #10235 

## 👋 Introduction

Adds the ability for sorting the pool candidates table by the skill count column

## 🕵️ Details

While working on this, I noticed we still had some old code for "locking" columns when record of decision flag is not present. Since we recently removed this flag, the locking was happening at the query level when it was not meant to be. So, this also removes that old code to no longer lock column sorting on this table.

## 🧪 Testing

1. Build `pnpm run dev`
3. Navigate to the pool candidates table (`/admin/pool-candidates`)
4. Add alot of skills tot he filters
    - Also, remove the government employee filter and set both status filters to "all" to maximize the candidates displayed
5. Sort by the "skill count" column
6. Confirm the results are sorted as expected

## 📸 Screenshot

screenshot_06-May-2024_15-28-1715023693

### DIFF
--- a/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
+++ b/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
@@ -663,7 +663,7 @@ const PoolCandidatesTable = ({
         ),
       },
     ),
-    columnHelper.display({
+    columnHelper.accessor("skillCount", {
       id: "skillCount",
       header: intl.formatMessage(tableMessages.skillCount),
       cell: ({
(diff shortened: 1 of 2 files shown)


vd1992 (approved):
Huh, well that was more straightforward then I thought. Nice!

image

PR #10295, opened 2024-05-07
[Fix] Pin chromatic to previous patch release

### LINKED ISSUE #10296: 🐛 Chromatic action not pinned
## 🐛 Bug

All of our other actions are pinned to a patch version so we can handle all updates through dependabot. Unfortunaltey, chromatic still had a floating version so it auto bumped a patch version with a bug :cry: 

## 🦋 Expected Behaviour

Action is only bumped by dependabot.

## 📋 Steps to Reproduce

Just go and look at all the failed chromatic pipelines.

## 🙋‍♀️ Proposed Solution

Pin action to the version before the but `11.3.0`

## ✅ Acceptance Criteria

- [ ] Chromatic action pinned to stable patch release
- [ ] Chromatic is passing in CI

esizer:
:robot: Resolves #10296

## 👋 Introduction

Pins chromatic action to a stable patch release to fix the workflow and allow dependabot to bump it in the future.

## 🧪 Testing

Check chromatic? :thinking:

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -56,7 +56,7 @@ jobs:
           pnpm build
 
       - name: "Publish Storybook: all"
-        uses: chromaui/action@v11
+        uses: chromaui/action@v11.3.0
         with:
           projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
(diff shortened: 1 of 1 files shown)


vd1992 (approved):
Fixed!

Well, this will be one speedy deny when dependabot puts out that PR this week

PR #10304, opened 2024-05-08
[Debt] Update applicant URL paths

### LINKED ISSUE #5413: ♻️ Profile page url includes user-id, but any current logged-in user is shown
## ♻️ Debt/Refactor

When the user goes to their own profile page, the url includes the User's id. However, the implementation of the profile page uses the `getMe` query, meaning the id in the url is entirely irrelevant. If a user logs in and goes to a url representing someone else's profile, they will still see their own profile.

## 🙋‍♀️ Proposed Solution

To make the behaviour of this page more consistent, we should either:
1. Use the `user(id)` query instead of `getMe`, and make use of the id in the url,
2. or change the url so it doesn't show an irrelevant and misleading id.

Solution 1 Advantages:
    - Don't need to change any routing
    - Purely a refactor, as expected behaviour does not change
    - A user with admin priveliges could visit other user profiles using currently set routes
Cons:
    - url is longer and more complicated, impractical to remember or enter by hand
    - almost all users will not have permissions necessary to view other profiles anyway
    - very few users, if any, have permissions necessary to view someone else's profile, and if they do we have a different page for that on the admin dashboard. So this could be maintaining unused & wasted functionality.

Solution 2 Advantages:
    - profile page url becomes shorter and simpler. It would now be easy to share the link that sends a user to their own profile, or for a user to go directly to their profile by entering the url by hand.
    - continuing to use `getMe` query and not having to pull id from id from url is simpler in some ways
Cons:
    - requires a tweak to routing
    - if we're being really thorough, we'll need to add permanent redirects for `users/*/profile` to the new url in our nginx config

**Tristan: Personally, I'm leaning toward solution 2 right now.**

## 🌎 Localization

Required if changing the url.

## ✅ Acceptance Criteria

- [ ] Change the URLs as in the list in Tristan's comment
- [ ] Create redirects for any URLs that don't currently have IDs in them (/applicant and /applicant/skills/showcase)
- [ ] Work with @petertgiles to make sure the notification emails still point to the right place

esizer:
🤖 Resolves #5413 

## 👋 Introduction

Updates the applicant URL paths to exclude user ID.

## 🧪 Testing

1. Rebuild nginx `make down && make up`
2. Build app `pnpm run dev`
3. Visit applicant pages and confirm URLs match [list outlined in the ticket discussion](https://github.com/GCTC-NTGC/gc-digital-talent/issues/5413#issuecomment-1979470197)
4. Visit old URLs and confirm redirects go to correct new paths

## :truck: Deployment

We will need to update the `OAUTH_POST_LOGIN_REDIRECT` to the new path: `/applicant`


apps/e2e/cypress/e2e/talentsearch/redirects.cy.ts
petertgiles, line 1: Worth adding some tests for your new redirects in here?

infrastructure/conf/nginx-conf-deploy/default
```diff
@@ -33,6 +33,15 @@ server {
     location = /fr/applicant/dashboard { return 301 /fr/applicant/profile-and-applications$is_args$args; }
     location = /en/privacy-notice { return 301 /en/privacy-policy$is_args$args; }
     location = /fr/privacy-notice { return 301 /fr/privacy-policy$is_args$args; }
+    location = /en/talent/profile { return 301 /en/applicant$is_args$args; }
```
petertgiles, line 36: These two are currently the fallback landing page for login redirect.  You should probably update that in this PR, too.  It will require a deployment step to update the value in Azure, as well.
https://github.com/GCTC-NTGC/gc-digital-talent/blob/9f1d5c16acb1fc20bdf6c290e97528acfec80308/api/app/Http/Controllers/AuthController.php#L122

apps/web/src/pages/Profile/CareerTimelineAndRecruitmentPage/components/CareerTimelineSection.tsx
petertgiles, line 1: I'm getting an error when I try to add a career timeline experience.
image

petertgiles (changes requested):
Very nice - looks like it really simplified some areas.  I flagged a few things to check on before merging.

infrastructure/conf/nginx-conf-deploy/default
esizer: oh yeah, thanks for nothing that :+1: Updated in [update login redirect path](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10304/commits/937e2f178eb5e755eaceb6e90330f7152aabb3ed)

apps/e2e/cypress/e2e/talentsearch/redirects.cy.ts
esizer: Yeah, that makes sense. I added them to playwright though instead of cypress :sweat_smile: 

[add redirect tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10304/commits/d45012f5987933076f659bdf68c290244253da63)

apps/web/src/pages/Profile/CareerTimelineAndRecruitmentPage/components/CareerTimelineSection.tsx
esizer: Oops! Didn't realize we were getting the user Id from the route here :sweat_smile: Good catch!

Fixed in [fix experience form mutations](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10304/commits/5d744b18a0342f581e49ce2dbc7c81e865de357e)

petertgiles (approved):

PR #10305, opened 2024-05-08
[Feature] Creating the community model

### LINKED ISSUE #10298: ✨ Create Community model
## ✨ Feature

Create new Community model.

## 🕵️ Details

For now, Community model can be very simple.
```
type Community {
  id: ID!
  key: String!
  name: LocalizedString
  description: LocalizedString
}
```
QUESTION: do we want to call it the more verbose `FunctionalCommunity` instead?

Create a query to view all Communities, but no mutations yet. Protected by a policy. Give base_user and guest roles the `view-any-community` permission.

Seeder which creates two prod-ready Communities, one for Digital, one for ATIP. These will be used by later migrations which need to populate existing entities with a mandatory Community relationship.

## 🌎 Localization

Copy for the default communities, until further notice:

field | Digital | ATIP
-- | -- | --
key | digital | atip
name | Digital Community, Communauté numérique | Access to Information and Privacy (ATIP), Demande d'accès à l'information et de protection des renseignements personnels (AIPRP)
description | `leave blank` | `leave blank`

## ✅ Acceptance Criteria

- [ ] New community model exists (database migration, eloquent model, schema type)
- [ ] Seeder which creates Digital and ATIP communities if they're not already in the database. Can be safely run in prod.
- [ ] Default database seeder runs the Communities seeder
- [ ] Communities can be queried, and base users and guest users have permission to view all communities.

## Deployment

Run the communities seeder.

vd1992:
🤖 Resolves #10298

## 👋 Introduction

Creation of the community model

## 🕵️ Details

Migration to add table, model file, schema, and factory for good measure. Query to view the collection too.
Role-permissions updated, as well as creating a community seeder.

## 🧪 Testing

1. Migrate fresh and seed
2. Use query below

```
query communities {
  communities {
    id
    key
    name {
      en
      fr
    }
    description {
      en
      fr
    }
  }
}
```

## 📸 Screenshot

image

## 🚚 Deployment

Run

`php artisan db:seed --class=RolePermissionSeeder`

`php artisan db:seed --class=CommunitySeeder`


api/app/Policies/CommunityPolicy.php
```diff
@@ -0,0 +1,26 @@
...
+     * @return \Illuminate\Auth\Access\Response|bool
+     */
+    public function viewAny(?User $user)
+    {
+        if ($user) {
+            return $user->isAbleTo('view-any-community');
+        }
+
+        return true;
+    }
```
vd1992, line 25: Was a bit iffy on this, is it even worth checking the permission? Or should it just return true by default? 
But that makes it sort of pointless having those permissions to begin with 🤔

esizer: @tristan-orourke can probably give a better answer but if anonymous (seems to be the case) users can view it, then I think just returning `true` makes sense. That is what we do for other things like `Skill` model.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/b160f92cb9b11b95cb2af66ab41514aeed35c18f/api/app/Policies/SkillPolicy.php#L19-L22

esizer (approved):
Looks good :+1: 

I'd just hold off merging until we hear about the policy method.

api/app/Policies/CommunityPolicy.php
tristan-orourke: Yeah, I would say follow what the Skill and Department models use, which is just returning true.

vd1992: Okay, will this be fine @esizer 
https://github.com/GCTC-NTGC/gc-digital-talent/pull/10305/commits/f01e627ff21a48f4f05bc6cc90c1d3d5599f642e

api/database/seeders/CommunitySeeder.php
vd1992, line 1: Safe to run seeder, @tristan-orourke  ?

tristan-orourke: Yep!

PR #10342, opened 2024-05-09
[Feature] Replace candidate status table

### LINKED ISSUE #9233: Update 'Candidate status' and 'Candidate search' table and filters
## Description

After RoD the status have been updated and split to better manage candidate flow through the platform. We can find these tables in several places throughout the platform:
1. Candidate search page on Admin
2. Candidate's Pools table on the Admin's User information tab.

After RoD, this table should be updated to reflect the new work on RoD statuses.

<details><summary>This is the new table that we like: 📸 Candidate search page</summary>
<p>

image

</p>
</details> 
<details><summary>The pool status table in this screenshot should be replaced by the Candidate search table above (but filtered for only this one candidate) 📸 Candidate status on User information tab</summary>
<p>

image

</details> 

### Candidate search table

1. Remove Status (old) column
2. Add functionality to the "job placement" column as seen in [#9140](https://github.com/GCTC-NTGC/gc-digital-talent/issues/9140)
3. Add "Application" column after 'Final decision' column (this will tell us the pool and will link to the application)

### Candidate status table on User information tab

1. Replace this table and reproduce the table with the one found on Candidate search but only for this user

## 🛑 Blockers
```[tasklist]
### Blocked By
- [ ] #9140 
```

vd1992:
🤖 Resolves #9233

## 👋 Introduction

## 🕵️ Details

## 🧪 Testing

1. ...
2. ...

## 📸 Screenshot

## 🚚 Deployment

- re-running database seeders

> **Notes**
>

### DIFF
--- a/apps/web/src/pages/Users/UserInformationPage/UserInformationPage.tsx
+++ b/apps/web/src/pages/Users/UserInformationPage/UserInformationPage.tsx
@@ -10,6 +10,7 @@ import { Pending, TableOfContents, ThrowNotFound } from "@gc-digital-talent/ui";
 import { unpackMaybes } from "@gc-digital-talent/helpers";
 import { commonMessages } from "@gc-digital-talent/i18n";
 import {
+  Department,
   FragmentType,
   Pool,
   Scalars,
@@ -256,9 +257,14 @@ const UserInfo_Fragment = graphql(/* GraphQL */ `
 interface UserInformationProps {
   userQuery: FragmentType<typeof UserInfo_Fragment>;
   pools: Pool[];
+  departments: Department[];
 }
 
-const UserInformation = ({ userQuery, pools }: UserInformationProps) => {
+const UserInformation = ({
+  userQuery,
+  pools,
+  departments,
+}: UserInformationProps) => {
   const intl = useIntl();
   const user = getFragment(UserInfo_Fragment, userQuery);
 
@@ -282,7 +288,13 @@ const UserInformation = ({ userQuery, pools }: UserInformationProps) => {
           "Title of the 'Candidate status' section of the view-user page",
       }),
       titleIcon: CalculatorIcon,
-      content: <CandidateStatusSection user={user} pools={pools} />,
+      content: (
+        <CandidateStatusSection
+          user={user}
+          pools={pools}
+          departments={departments}
+        />
+      ),
     },
     {
       id: "notes",
@@ -354,6 +366,14 @@ const UserInformation_Query = graphql(/* GraphQL */ `
       }
       status
     }
+    departments {
+      id
+      departmentNumber
+      name {
+        en
+        fr
+      }
+    }
   }
 `);
 
@@ -371,6 +391,7 @@ const UserInformationPage = () => {
 
   const user = data?.user;
   const pools = unpackMaybes(data?.pools);
+  const departments = unpackMaybes(data?.departments);
 
   return (
     <AdminContentWrapper>
@@ -383,7 +404,11 @@ const UserInformationPage = () => {
       />
       <Pending fetching={fetching} error={error}>
         {user && pools ? (
-          <UserInformation userQuery={user} pools={pools} />
+          <UserInformation
+            userQuery={user}
+            pools={pools}
+            departments={departments}
+          />
         ) : (
           <ThrowNotFound />
         )}
--- a/apps/web/src/pages/Users/UserInformationPage/UserInformationPage.tsx
+++ b/apps/web/src/pages/Users/UserInformationPage/UserInformationPage.tsx
@@ -56,6 +56,7 @@ const UserInfo_Fragment = graphql(/* GraphQL */ `
     govEmployeeType
     hasPriorityEntitlement
     priorityNumber
+    priorityWeight
     locationPreferences
     locationExemptions
     positionDuration
@@ -71,6 +72,15 @@ const UserInfo_Fragment = graphql(/* GraphQL */ `
       expiryDate
       notes
       suspendedAt
+      submittedAt
+      placedDepartment {
+        id
+        departmentNumber
+        name {
+          en
+          fr
+        }
+      }
       user {
         id
       }
--- a/apps/web/src/pages/Users/UserInformationPage/components/UserCandidatesTable/UserCandidatesTable.tsx
+++ b/apps/web/src/pages/Users/UserInformationPage/components/UserCandidatesTable/UserCandidatesTable.tsx
@@ -6,7 +6,7 @@ import {
   Department,
   FragmentType,
   PoolCandidate,
-  User,
+  UserInfoFragment,
 } from "@gc-digital-talent/graphql";
 import {
   commonMessages,
@@ -15,7 +15,7 @@ import {
   getPoolCandidatePriorities,
   getPoolCandidateStatus,
 } from "@gc-digital-talent/i18n";
-import { unpackMaybes } from "@gc-digital-talent/helpers";
+import { notEmpty, unpackMaybes } from "@gc-digital-talent/helpers";
 
 import Table from "~/components/Table/ResponsiveTable/ResponsiveTable";
 import { normalizedText } from "~/components/Table/sortingFns";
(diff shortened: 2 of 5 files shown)


apps/web/src/pages/Users/UserInformationPage/UserInformationPage.tsx
```diff
@@ -70,6 +72,15 @@ const UserInfo_Fragment = graphql(/* GraphQL */ `
...
       suspendedAt
+      submittedAt
+      placedDepartment {
+        id
+        departmentNumber
+        name {
+          en
+          fr
+        }
+      }
```
vd1992, line 83: My attempt at verbose typing let me catch what I needed to add 🧐 
Otherwise it is silent

image

packages/fake-data/src/fakeUsers.ts
```diff
@@ -193,4 +193,101 @@ export const fakeApplicants = (numToGenerate = 20): GeneratedUser[] => {
   return defaultGenerator(numToGenerate);
 };
 
+export const fakeUser = (): User => {
+  const departments = fakeDepartments();
```
vd1992, line 197: `GeneratedUser` type was not playing nice so I added this in for when you just want `User`

vd1992:
Okay @Jerryescandon 
I made some changes as the issue wanted, though, the issue is a few months old so some things have changed. Some stuff completed elsewhere or renamed for example. 

I created a story, please scroll down to the table and tell me what you would like to tweak, like what columns to hide by default. 

https://www.chromatic.com/component?appId=61e099a1bb1465003a73d3ce&csfId=pages-users-userinformationpage&specName=Default&buildNumber=7414&k=663e772a3278716464400d7f-1200px-interactive-true&h=3&b=-1

vd1992:
> Awesome work! I think for the default view we can keep the ones in green and hide the reds. 

Thanks for the answer, done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/10342/commits/b7a48f39a6bf9e05c374c45f29ca0cd4e05dad18

apps/web/src/pages/Users/UserInformationPage/components/UserCandidatesTable/UserCandidatesTable.tsx
```diff
@@ -0,0 +1,261 @@
...
+  | "notes"
+  | "submittedAt"
+  | "status"
+  | "pool"
+  | "placedDepartment"
+  | "suspendedAt"
+  | "isBookmarked"
+>;
+
+const columnHelper = createColumnHelper<PoolCandidateSlice>();
```
esizer, line 61: Just a question because I was struggling with this but did you try a fragment here?

`const UserCandidatesTableRow_Fragment = graphql()`

apps/web/src/pages/Users/UserInformationPage/components/UserCandidatesTable/UserCandidatesTable.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+const UserCandidatesTable = ({
+  user,
+  departments,
+  title,
+}: UserCandidatesTableProps) => {
+  const intl = useIntl();
+  const paths = useRoutes();
+
+  const poolCandidatesUnpacked = user.poolCandidates?.filter(notEmpty) ?? [];
```
esizer, line 77: This is exactly what `unpackMaybes` does so I think we should use that instead.

apps/web/src/pages/Users/UserInformationPage/components/UserCandidatesTable/UserCandidatesTable.tsx
```diff
@@ -0,0 +1,261 @@
...
+        id: "candidateName",
+        header: intl.formatMessage(tableMessages.candidateName),
+        sortingFn: normalizedText,
+        cell: ({ row: { original: poolCandidate } }) =>
+          candidateNameCell(user, poolCandidate, paths, intl, candidateIds),
+        meta: {
+          isRowTitle: true,
+        },
+      },
+    ),
```
esizer, line 123: I think this column is meant to be removed.

https://github.com/GCTC-NTGC/gc-digital-talent/pull/10342#issuecomment-2105265420

apps/web/src/pages/Users/UserInformationPage/components/UserCandidatesTable/UserCandidatesTable.tsx
```diff
@@ -0,0 +1,261 @@
...
+      cell: () => cells.email(user.email),
+    }),
+    columnHelper.accessor(
+      () =>
+        currentLocationAccessor(user.currentCity, user.currentProvince, intl),
+      {
+        id: "currentLocation",
+        header: intl.formatMessage(tableMessages.currentLocation),
+      },
+    ),
```
esizer, line 230: Same withe removing these. This information already appears on this page so it seems to just be duplicating it unnecessarily.

https://github.com/GCTC-NTGC/gc-digital-talent/pull/10342#issuecomment-2105265420

esizer (changes requested):
Looking good, just noticed a few things.

apps/web/src/pages/Users/UserInformationPage/components/UserCandidatesTable/UserCandidatesTable.tsx
vd1992: Um, I think it was to be hidden by default?

Jerryescandon: Yeah I agree with @esizer , but  I don't know what's best in terms of implementation, I just assumed that reusing a table and hiding anything that's not valuable would be the easiest solution. 

But I agree that information that repeats elsewhere is not necessary on the table itself. Mostly the table only needs to show information regarding the user/pool relationship like the final decision, status, expiry date, placement etc.

apps/web/src/pages/Users/UserInformationPage/components/UserCandidatesTable/UserCandidatesTable.tsx
vd1992: Makes sense to me! https://github.com/GCTC-NTGC/gc-digital-talent/pull/10342/commits/7aa94c753ce15bc6c17a91d252d1811834aa5bf5

apps/web/src/pages/Users/UserInformationPage/components/UserCandidatesTable/UserCandidatesTable.tsx
vd1992: Am experimenting. Tried something else in https://github.com/GCTC-NTGC/gc-digital-talent/pull/10342/commits/bb60ecac454518dc65f1fe5f5dc7c114ec3699f3

The tricky bit is not losing the typesafe of fragments, as we've seen with passing along into further functions or components you can easily lose the checking. If you pass your data into something with `type: PoolCandidate` you won't know if your fragment was missing something. 

An aside, I am finding it quite difficult to ensure all the RoD fields are present as it is a chain of functions for those and they often have the type PoolCandidate even when only interacting with 2 or 3 fields. Making it so I have to paste a lot of fields into the fragment just in case. So it'll be function A calling B, which calls function C and it passes along the entire pool candidate object and you have to dig around to see what fields it even needed.

apps/web/src/pages/Users/UserInformationPage/components/UserCandidatesTable/UserCandidatesTable.tsx
vd1992: Okay, fair. I removed some accessors. https://github.com/GCTC-NTGC/gc-digital-talent/pull/10342/commits/4218e22d60db862c7c4ba57d704e20a8aa81d9e5
Am keeping in the name one as that is more than just the name but also how you get to the specific application. It is the same way as for the pool candidates table and elected to have it work and appear the same.

esizer (approved):
Cool :sunglasses:

PR #10369, opened 2024-05-13
[Debture] Remove search comma separation

### LINKED ISSUE #9963: ✨ Remove comma separated search terms from candidate/user search
## ✨ Feature

We initially built the candidate/user general search to accept multiple search terms (which are joined by & on the backend), and on the frontend we split on commas to turn the search input into multiple search terms.

This functionality is not understood or used by users, and is made redundant by the default behaviour of the text search, which allows for boolean search terms, and treats each word as its own search term unless wrapped in quotes.

Let's just take out the behaviour to simplify things.

## 🕵️ Details

On the frontend, on user and candidate tables, don't split search terms on commas.
On the api, pass a single string to general search instead of an array of search terms.

## 🎨 Design File

None

### 🧑‍🎨 Designer

@Jerryescandon would be the best designer advisor for this.

I think this should be straightforward.

## ✅ Acceptance Criteria

- [ ] `UserFilterInput.generalSearch` and `PoolCandidateSearchInput.generalSearch` are turned from an array of strings into a single string
- [ ] Search bar in User and Candidate tables does not split on commas
and

vd1992:
🤖 Resolves #9963 

## 👋 Introduction

Remove the comma separation from the text search from front, API, and back. 

## 🧪 Testing

1. Commas will not split, ensure searching not broken for tables
2. Not much to actually test

### DIFF
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -405,15 +405,15 @@ class PoolCandidate extends Model
         return $query;
     }
 
-    public function scopeGeneralSearch(Builder $query, ?array $searchTerms): Builder
+    public function scopeGeneralSearch(Builder $query, ?string $searchTerm): Builder
     {
-        if (empty($searchTerms)) {
+        if (empty($searchTerm)) {
             return $query;
         }
 
-        $query->where(function ($query) use ($searchTerms) {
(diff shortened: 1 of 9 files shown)


tristan-orourke (commented):
Code looks great! Just gonna test manually now

PR #10370, opened 2024-05-13
[Feature] Admin table classification filter doesn't also filter status

### LINKED ISSUE #10081: ✨ In pool candidate tables, Classification filter should should not force QUALIFIED/AVAILABLE status
## ✨ Feature

In the public talent search, filtering by classification restricts the resulting count to candidates who are qualified and available in pools of that classification. That means, candidates who are still in assessment or screened out will be filtered out.

On the admin side, the Pool Candidates table includes a separate filter input for candidate status (eg screened out, qualified available, etc). Therefore, it doesn't make sense for the Classification filter to force certain status scopes behind the scenes.

## 🕵️ Details

The classification input on both the countCandidates queries and the Pool Candidates table are passed to the PoolCandidate `qualifiedClassifications` scope. This works for the public talent search but no longer makes sense for the admin tables.

I think the easiest solution would be to add a new scope to PoolCandidate.php which filters candidates by the classification of the pool they applied to.  Perhaps call it `appliedClassifications` instead of `qualifiedClassifications`. And add it to `PoolCandidateSearchInput`, which is only used on privileged queries, instead of `ApplicantFilterInput`, which is used by the public search queries. Then we make sure the PoolCandidates tables on Admin _don't_ use `qualifiedClassifications`, and use the other one instead (even when duplicating the results of a talent request).

As an alternate solution... The public countApplicants queries could also use the new scope, as long as those public queries have the `available` scope always added at the schema level.

> Side note: ApplicantFIlterInput vs PoolCandidateSearchInput is very confusing in hindsight. I'm pretty sure these are my fault. The main difference is that one can be used by the public and the other cannot. There's also the fact that ApplicantFilter(Input) is an object which can both be the result of a query and the input to one, which makes for some convenience on the frontend but adds some constraints to the backend... Perhaps this is worth a refactor at some point, if only to rename things more clearly.

## ✅ Acceptance Criteria

- [ ] On any Pool Candidates tables, admins can clear the 'status' filter and use the Classifications filter to see candidates in pools of that classification of any status
- [ ] When opening a Talent Request page on the admin side, the number of candidates in the table matches the number of candidates shown on the public talent search for the same request
- [ ] talent searches on the public side continue to filter out candidates who aren't qualified and available
- [ ] PHP unit testing
- [ ] Talent requests not broken

JamesHuf:
🤖 Resolves #10081

## 👋 Introduction

Updates the pool candidate tables classification filters to not also filter out unqualified candidates.

## 🕵️ Details

Adds a new appliedClassifications field to PoolCandidateSearchInput which won't also filter based on status.

## 🧪 Testing

1. Ensure you have a pool with at least on each of qualified, expired, and non-qualified status (ex. screened in) as well as at least one qualified candidate of a different classification.
2. Navigate to the search page `/search`.
3. Filter to the classification of the pool in step one and make note of the candidate count.
4. Fill out the necessary fields to submit and talent request for the classification.
5. Navigate to the talent requests page `/admin/talent-requests`.
6. Select to view your new request.
7. Scroll down to the candidate results section and clear any filters you added to submit the request.
8. Confirm the number of candidates matches the count in step 3. Note: "Removed" candidates won't be filtered out - ignore them in this count.
9. Change the "Expiry status" and "Candidacy status" filters to all and confirm they both increase the candidate count.

### DIFF
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -261,6 +261,33 @@ class PoolCandidate extends Model
     }
 
     /**
+     * Scopes the query to return PoolCandidates in a pool with one of the specified classifications.
+     * If $classifications is empty, this scope will be ignored.
+     *
+     * @param  array|null  $classifications  Each classification is an object with a group and a level field.
+     */
+    public static function scopeAppliedClassifications(Builder $query, ?array $classifications): Builder
+    {
+        if (empty($classifications)) {
+            return $query;
+        }
+
+        $query->whereHas('pool', function ($query) use ($classifications) {
+            $query->whereHas('classification', function ($query) use ($classifications) {
+                $query->where(function ($query) use ($classifications) {
+                    foreach ($classifications as $classification) {
+                        $query->orWhere(function ($query) use ($classification) {
+                            $query->where('group', $classification['group'])->where('level', $classification['level']);
+                        });
+                    }
+                });
+            });
+        });
+
+        return $query;
+    }
+
+    /**
      * Scopes the query to only return PoolCandidates who are available in a pool with one of the specified classifications.
      * If $classifications is empty, this scope will be ignored.
      *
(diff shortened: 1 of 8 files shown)


petertgiles (approved):
I think this is good.  :sweat_smile: It's pretty complicated but everything seems to be working properly.

Issue #6169, opened 2023-04-03
✨ Add 'Publishing group' filter to 'Pools' table


Jerryescandon:
## ✨ Feature

Include a filter on the pools table to be able to sort by "Publishing group"

## 🕵️ Details

DCM wants to be able to quickly get to their Continues Intake pools, which should be under their own publishing group. Adding this filter will allow for easier access to these pools.

## 📸 Screenshot

Pools page

![image](https://user-images.githubusercontent.com/16977254/229599808-5558b9b0-5b89-45f7-a9ca-276f716961c2.png)

Pools filter modal

![image](https://user-images.githubusercontent.com/16977254/229601436-3ec976be-850d-45c4-8ab5-4523032ccf8c.png)

## ✅ Acceptance Criteria

- [ ] Pools table filterable on publishing group
- [ ] Works in EN and FR

## 🛑 
```[tasklist]
### Blocked by
- [ ] #9718 
```

vd1992:
the pools table isn't set up like the pool candidate or users table, so added filtering won't be able to match them easily
(pool candidate and users table use the api-driven table, pools table does not)

petertgiles:
Note from Eric: https://react-table-v7.tanstack.com/docs/api/useFilters

esizer:
> the pools table isn't set up like the pool candidate or users table, so added filtering won't be able to match them easily (pool candidate and users table use the api-driven table, pools table does not)
> 
> so adding filtering should stay a separate task

Yeah, if we want to keep this as a client side table we may need to look at filtering. It may be a little tricky since the client side tables use the search dropdown for filters which would override any filters in the dialog (how it currently works) So it may require a bit of refactoring of the base table component to see how we can prevent the search from overriding other filters.

Or, we could create a paginated query and convert it to a server side table (_could_ be more beneficial in the long run).

--- 

Also, there is some work being done on the filter dialogs that will be impacting this so I an going to add #8776 as a blocker.

mnigh:
> Or, we could create a paginated query and convert it to a server side table (_could_ be more beneficial in the long run).

- @Jerryescandon @tristan-orourke is this feature still something the client is interested in? if so, should we create a paginated query and convert it to a server side table as mentioned by eric?

tristan-orourke:
You know what, it's probably time we make the pool query a paginated query, anyway.

PR #10383, opened 2024-05-14
[Debt] Replace `pools` query with `poolsPaginated`

### LINKED ISSUE #9718: ♻️ Replace the `pools` query with a `poolsPaginated` query
## ♻️ Debt/Refactor

We have a lot of pools at this point, and the number will continue to grow. We should deprecate the `pools` query and replace it with a paginated version. 

The client should switch to using the paginated version under the hood, but this change should be invisible to users.

## 🕵️ Details

Note: `publishedPools` query will remain non-paginated for now.

## 🙋‍♀️ Proposed Solution

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] `pools` query deprecated and not used anywhere client-side
- [ ] use of `pools` query replaced with a `poolsPaginated` query

## 🛑 Blockers

```[tasklist]
### Blocked By
```

esizer:
🤖 Resolves #9718

## 👋 Introduction

Replaces the existing `pools` query with a paginated version.

## 🧪 Testing

1. Build `pnpm run dev`
3. Navigate to areas affected:
    - Admin pools page
    - Pool candidate page (add to pool dialog, change status dialog)
    - User information page (add to pool dialog)
    - Pool candidates table (filter dialog)


api/app/Models/Pool.php
```diff
@@ -314,8 +314,139 @@ public static function scopeCurrentlyActive(Builder $query)
         return $query;
     }
 
+    public static function scopeName(Builder $query, ?string $name): Builder
+    {
+        if ($name) {
+            $query->where(function ($query) use ($name) {
+                $term = sprintf('%%%s%%', $name);
```
tristan-orourke, line 321: What exactly to the five % do? It thought we usually only have one on either side 😆

esizer: with `sprintf` the `%` indicates a replacement value so you add two `%%` to escape it :sweat_smile:

esizer: https://www.php.net/manual/en/function.sprintf.php#:~:text=With%20printf()%20and%20sprintf()%20functions%2C%20escape%20character%20is%20not%20backslash%20%27%5C%27%20but%20rather%20%27%25%27.

api/graphql/schema.graphql
```diff
@@ -628,6 +628,21 @@ input UserPublicProfileFilterInput {
   publicProfileSearch: String @scope
 }
 
+input PoolFilterInput {
+  generalSearch: String @scope
+  name: String @scope
+  ownerName: String @scope
+  ownerEmail: String @scope
```
tristan-orourke, line 635: We should probably deprecate these fields on the Pool, actually, and skip adding them here. I don't think we're using them anywhere.

tristan-orourke (commented):
Could you add some phpunit tests testing the filters on the new pool query?

apps/web/src/pages/Pools/IndexPoolPage/components/helpers.tsx
```diff
@@ -135,3 +147,118 @@ export function ownerNameAccessor(pool: Pool) {
...
+  return {
+    generalSearch: !!search.term && !search.type ? search.term : undefined,
+    name: search.type === "name" ? search.term : undefined,
+    ownerName: search.type === "ownerName" ? search.term : undefined,
+    ownerEmail: search.type === "ownerEmail" ? search.term : undefined,
+    team: search.type === "type" ? search.term : undefined,
+  };
+}
+
+export function getOrderByClause(
```
tristan-orourke, line 172: Setting up this sorting is so much more complicated than I remembered 🫣

api/graphql/schema.graphql
esizer: Oh yeah, I was wondering if we needed this or not. Removed in [remove owner scopes from pool](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10383/commits/ba6cf3f8d9b866fb1bf7839fb32ab203ac4e8a04)

Just working on the tests now :+1:

apps/web/src/pages/Pools/IndexPoolPage/components/helpers.tsx
esizer: Oh yeah, thankfully we had an example I could copy and paste from :stuck_out_tongue:

api/graphql/schema.graphql
tristan-orourke: ownerName and ownerEmail I mean

apps/web/src/pages/Users/UserInformationPage/components/ChangeStatusDialog.tsx
```diff
@@ -269,7 +268,13 @@ const ChangeStatusDialog = ({
...
               </p>
               <div data-h2-margin="base(x.5, 0, x.125, 0)">
-                <Combobox
+                <PoolFilterInput
+                  includeIds={userPoolIds}
+                  filterInput={{
+                    statuses: [PoolStatus.Closed, PoolStatus.Published],
+                  }}
+                />
+                {/* <Combobox
```
tristan-orourke, line 277: Is this meant to be removed?

api/graphql/schema.graphql
esizer: Oh, would you like them to be removed from the `PoolTable` as removing the scopes?

apps/web/src/pages/Users/UserInformationPage/components/ChangeStatusDialog.tsx
esizer: Oops, yeah I did :sweat_smile: Removed in [remove commented combobox](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10383/commits/0124c4b3d14607d1c7bd0bc146347bc5f4603774)

tristan-orourke:
@esizer could you take a look at the visual diffs on chromatic? Looks like some information is getting lost.

esizer:
> @esizer could you take a look at the visual diffs on chromatic? Looks like some information is getting lost.

Okay, I think I was able to fix the request form (I forgot ot add the mock query). However, the table diff was because I couldn't get the sort on the status because it is not actually a database column. I may need assistance on that because with the paginated queries, we can only sort by selected database columns since it is just doing a SQL `ORDER BY`.

I tried adding the `addSelect` but that was overriding the other selects :confounded: 

[fix request form story](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10383/commits/3842c7deab36192c932f01a507679ca9e8d74401)

api/graphql/schema.graphql
```diff
@@ -747,6 +760,26 @@ type Query {
...
+        relations: [
+          { relation: "user", columns: ["first_name", "email"] }
+          { relation: "classification", columns: ["group", "level"] }
+        ]
+      )
+  ): [Pool]!
+    @paginate(
+      defaultCount: 10
+      maxCount: 1000
+      scopes: ["authorizedToView", "notArchived"]
```
esizer, line 780: Working on the tests and was like "why is the archived scope failing?!" Then I remembered this...

Do we want this? :thinking: Or should I just not consider archived in the tests? @tristan-orourke

esizer:
> Could you add some phpunit tests testing the filters on the new pool query?

Added tests and the bugs it uncovered :stuck_out_tongue: [add tests, fix bugs from tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10383/commits/17fe2b71038821023abe405a503f0ba2aa8284dc)

api/graphql/schema.graphql
tristan-orourke: We can take off the built-in notArchived scope, though most places querying pools will want to add it. Could we make it enabled by default?

esizer: Hmm, I'll take a look. I'm pretty sure you can provide defaults and then when you pass something in, it overrides it :thinking:

tristan-orourke: I don't think there's an easy way to make it default. That's fine.

tristan-orourke (changes requested):
Found a couple funny things on the pools table.

If I filter by Publishing Group or Team, then the page controls don't notice the reduced results. (Is this filtering happening on the frontend?) 
image

Meanwhile, you can't search by publishing group unless you select the "Filter by Publishing Group" name, because the general search scope doesn't include it.

esizer:
> If I filter by Publishing Group or Team, then the page controls don't notice the reduced results. (Is this filtering happening on the frontend?)

Oh yeah, I actually did not intended to allow searching by publishing group since it is tricky due to being an enum and will be moving to filters once this gets merged. Does that sound right, or do you want me to try and figure out searching by enum value through the DB? :thinking:

tristan-orourke:
You're right, publishing group should be handled by filters, not the search bar. (Which is why we have https://github.com/GCTC-NTGC/gc-digital-talent/issues/6169!) Is there anything we can do to mitigate the buggyness in the meantime? 🤔

esizer:
> You're right, publishing group should be handled by filters, not the search bar. (Which is why we have #6169!) Is there anything we can do to mitigate the buggyness in the meantime? 🤔

I can just remove it from the search bar in this PR. Then, I'll pick the filter one up right away so we don't lose that ability in between deploys. Does that sound good?

tristan-orourke (approved):
That sounds great @esizer

PR #10385, opened 2024-05-14
[Fix] Fix snapshot test

### LINKED ISSUE #9499: 🧪 Update `testSnapshotQueryInSync` PHP Unit Test
## 🧪 Tests

This test will be broken after https://github.com/GCTC-NTGC/gc-digital-talent/issues/8716 and need to be updated.

## 🕵️ Details

We are removing our `*.graphql` files so there is nothing to compare against since the document is now in a Typescript file as a function argument.

## 🙋‍♀️ Proposed Solution

I'm not sure or I would have fixed it in my PR that removed the `profileOperations.graphql` 😅 

## ✅ Acceptance Criteria

- [ ] Test is no longer skipped
- [ ] Test sill passes and asserts the equality of operations
- [ ] Rewrite to compare query operation to snapshot? In which case just remove the skipped test, no need to fix it.

petertgiles:
🤖 Resolves #9499 

## 👋 Introduction

This branch replaces the old disabled PHPUnit test to make sure snapshot queries stayed in sync with a new Jest one.  It also syncs up the query differences.

## 🕵️ Details

I found that it was difficult for PHP to find the query in the TypeScript file but it was easy for TypeScript to load the query file for PHP.  So I rewrote the test in Jest instead.  I found a package from the Jest team called [jest-diff](https://www.npmjs.com/package/jest-diff) that made it easy to highlight differences between the files.

When I came to sync up the files I found that other than the fragment being inline, it was just a few new fields that needed to be added.  Also, they needed to be added to the Laravel resources to make sure they got stored.

## 🧪 Testing

- [ ] Jest test passed
- [ ] Jest test fails if you modify either query
- [ ] Snapshots are still saved and rendered properly by the app

## 📸 Screenshot

image


api/app/Http/Resources/UserStubResource.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace App\Http\Resources;
+
+use Illuminate\Http\Resources\Json\JsonResource;
+
+// Sometimes we have to query for relationship non-nullable fields when they're not needed because of codegen
```
esizer, line 7: hopefully we can finally not need this :cry:

esizer (approved):
Cool :sunglasses:

api/app/Http/Resources/UserStubResource.php
petertgiles: Yeah, this is a real thorn...

PR #10386, opened 2024-05-15
[Debt] Removes `OperationalRequirementV1`

### LINKED ISSUE #9917: ♻️ Remove `OperationalRequirementV1`
## ♻️ Debt/Refactor

Remove the old operational requirements.

## 🕵️ Details

We deprecated this a long time ago and feel like it is sage to fully remove it now. This will include deleting the old operational requirements and adding special handling on the front end where needed.

## 🙋‍♀️ Proposed Solution

1. Delete any operational requires that appear only in `OperationalRequirementV1` from the PHP enum `App\Enum\OperationalRequirement`
2. Remove references to `OperationalRequirementv1`
3. Maybe: You may need to wrap the snapshot render in an error boundary to catch instances where users have old operational requirements so it does not bubble up to the router error boundary necessarily

> [!IMPORTANT]  
> Check if values exist in Prod first to decide scope

## ✅ Acceptance Criteria

- [ ] References to old V1 operational requirements removed
- [ ] Snapshot renders properly when old values exist in the database

mnigh:
🤖 Resolves #9917.

## 👋 Introduction

This PR removes `OperationalRequirementV1` and associated code. It also renames `OperationalRequirementV2` to `OperationalRequirements` since there is no longer a reason to distinguish versions of the variable name.

## 🕵️ Details

It also adds a migration to remove the no longer used `OVERTIME_SCHEDULED` and `OVERTIME_SHORT_NOTICE` values from `pools.operation_requirements` and `users.accepted_operational_requirements` fields if they exist.

## 🧪 Testing

### `users` SQL query
```sql
SELECT accepted_operational_requirements FROM users WHERE accepted_operational_requirements @> '"OVERTIME_SHORT_NOTICE"' OR accepted_operational_requirements @> '"OVERTIME_SCHEDULED"'
```

### `pools` SQL query
```sql
SELECT operational_requirements FROM pools WHERE operational_requirements @> '"OVERTIME_SHORT_NOTICE"' OR operational_requirements @> '"OVERTIME_SCHEDULED"'
```

1. `pnpm build`
2. Add different combinations of `OVERTIME_SCHEDULED` and `OVERTIME_SHORT_NOTICE` to the `pools.operation_requirements` and `users.accepted_operational_requirements` fields
3. Run `users` SQL query and `pools` SQL query to ensure values exist in each table
4. Run migration `make migrate`
5. Run `users` SQL query and `pools` SQL query to ensure no values exist in either table
6. Ensure other values in `pools.operation_requirements` and `users.accepted_operational_requirements` remain the same


api/database/migrations/2024_05_14_194656_remove_operational_requirements_overtime_short_notice_overtime_scheduled.php
petertgiles, line 1: This migration is great for updating the database records but unfortunately they'll still be in profile snapshots which totally crash the app.
image

I don't think we want to update any snapshots - that kind of defeats the point of them.  I think in the past we talked about putting an error boundary around the component so that the worst-case scenario is that the application simply doesn't render but the app stays up.

The other thought that comes to me (but not discussed with the team) is that `getOrThrowError` in localizedConstants is pretty extreme.  It would probably be safer to just render the enum name (and log an error) if a description is not found and not take down the entire app.

petertgiles (changes requested):
Nice to finally get this cleaned up!  Everyone looks great, except for one hitch...

api/database/migrations/2024_05_14_194656_remove_operational_requirements_overtime_short_notice_overtime_scheduled.php
mnigh: @petertgiles should I block this on https://github.com/GCTC-NTGC/gc-digital-talent/issues/10395?

petertgiles: Yeah, please.  I don't think there is any more dev work to be done on this branch though.  Good-to-go once the app error handling is improved.

petertgiles: I just tested after the merge - looks good!

petertgiles (approved):

PR #10388, opened 2024-05-15
[Debt] Automatic JSX transform

### LINKED ISSUE #10367: ♻️ Use automatic JSX transform
## ♻️ Debt/Refactor

We should probably update to the automatic JSX transform.

## 🕵️ Details

It has beena  few years since React released an automatic JSX transform that has some [minor performance improvements](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#motivation).

## 🙋‍♀️ Proposed Solution

Follow the guide for our setup in [the blog post](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html).

## ✅ Acceptance Criteria

- [ ] React is using automatic JSX transform
- [ ] Explicit imports of `React` are removed
- [ ] Only import required items from react (hooks, types)

esizer:
🤖 Resolves #10367 

## 👋 Introduction

Updates to use the automatic JSX transform for React.

## 🕵️ Details

We should no longer be importing React and instead only use named imports.

## 🧪 Testing

1. Ensure all react imports are using named imports
2. Confirm app still builds without error


esizer:
> Is this supposed to remain the sole exception?

Definitely not! Good catch :clap: Removed in [remove missed react import](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10388/commits/3e6a238d18fdaa013ff5393ea67ecc851dc91808)

vd1992 (approved):
Well, can't find anymore mass imports. Branch builds, CI is good, and navigating briefly it appears to still work fine. 
The files changed tab lags my browser so I will pass on trying to review 800+ files changed. 

Don't know if anyone else wants to review this or has concerns

PR #10389, opened 2024-05-15
[Fix] Local seeding

vd1992:
## 👋 Introduction

Appears the addition of

https://github.com/GCTC-NTGC/gc-digital-talent/blob/76becb28ef7c963d633c6a65da819d10abbcc36a/api/app/Http/Resources/EducationExperienceResource.php#L29

Where the chain `user->experience->user` triggers lazy load error. So eager loading it fixes it 

## 🧪 Testing

1. Successfully seed locally

## 📸 Screenshot

Seed successfully

image


petertgiles (approved):
Thanks for the quick fix!

PR #10394, opened 2024-05-15
[Fix] Application e2e test

### LINKED ISSUE #10393: 🐛 Application e2e test intermittently failing
## 🐛 Bug

Our e2e test for submitting an application is failing on the add experience step but only some of the time :thinking: 

## 🦋 Expected Behaviour

Test passes consistently.

## 🕵️ Details

I think this has to do with our recent URL change since it fails on asserting a URL. However, I am no quite sure why it passes some of of the time :face_with_spiral_eyes: 

https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/9100001083/job/25013855166#step:13:42

## 🙋‍♀️ Proposed Solution

Find out the new URL and assert that, or remove the assertion since it is more there for safety and further assertions should work.

## ✅ Acceptance Criteria

- [ ] `application.spec.ts` consistently passes in all browsers

esizer:
🤖 Resolves #10393 

## 👋 Introduction

Removes a flaky assertion that was superfluous. 

## 🧪 Testing

1. Run the test a few times `pnpm run e2e:playwright:chromium application.spec.ts`
2. Confirm it passes consistently

### DIFF
--- a/apps/playwright/tests/applicant/application.spec.ts
+++ b/apps/playwright/tests/applicant/application.spec.ts
@@ -174,7 +174,6 @@ test.describe("Application", () => {
     await application.page
       .getByRole("link", { name: /add a new experience/i })
       .click();
-    await expect(application.page.url()).toContain("/career-timeline/add");
     await application.addExperience();
     await expect(application.page.getByRole("alert")).toContainText(
       /successfully added experience!/i,


vd1992 (approved):
Works locally 👍

PR #10432, opened 2024-05-21
[Feature] Associate requests with a community

### LINKED ISSUE #10331: ♻️ Associate every Request with a Community
## ♻️ Debt/Refactor

Create a many-to-one relationship between Requests and Communities - every Talent Request must belong to a single community. 
 (A community can have many requests.) This should apply retroactively too: every existing request in the database should belong to either the Digital community or the ATIP community, and we can use Pool Stream to determine which.

## 🕵️ Details

When submitting a talent request, it should be submitted to one of our communities.

For existing communities, they should belong to either the Digital community or the ATIP community, based on Pool Stream. 
- ACCESS_INFORMATION_PRIVACY means ATIP community
- any other stream means Digital
- a request without a stream (or with conflicting ones) defaults to Digital

The Community can be accessed from the ApplicantFilter and PoolCandidateSearchRequest types:
```
type ApplicantFilter {
   ...
   community: Community
}

type PoolCandidateSearchRequest {
   ...
   community: Community
}
```

For now, this community doesn't need to affect the results of the search. A future ticket will ensure that results are scoped to the community.

## 🙋‍♀️ Proposed Solution

Add `community_id` foreign key to both the PoolCandidateSearchRequest and the ApplicantFilter entities. (Schema type, eloquent model, and database tables.) It should a required non-nullable field on the search request, but optional in the filter. It is required on the search request because we will soon use it to control access. We will also use it to scope the results, hence being on the filter.

Note: we don't need to scope the results OR change access controls in this ticket. This is just setting us up for that.

An additional challenge:
If the migration adds a required (non-nullable) community_id field to the Requests table, what do we do about running the migration on systems with requests already in the database? We can do the matching in the mutation itself, but only if the Communities already exist. We likely need to blocking merging this until #10298 has been deployed and the seeder has been run.

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] All existing Requests associated with a Community in the database
- [ ] Any new Requests required a community
- [ ] The Community can be accessed from a Request in the api
- [ ] Creating new requests now requires selecting a Community
- [ ] The search page automatically sets the community on the request and filter based on the Pool Stream.
- [ ] Search request UI is unchanged

## 🛑 Blockers

```[tasklist]
### Blocked By
- [ ] https://github.com/GCTC-NTGC/gc-digital-talent/pull/10305
```

esizer:
🤖 Resolves #10331 

## 👋 Introduction

Updates the search request and applicant filter models to have a one to many relationship with a community.

## 🕵️ Details

Right now we are storing the community on both the applicant filter and search request. There is an outstanding discussion going on if we should store it on just one and pass the relationship along.

## 🧪 Testing

1. Migrate data `make refresh-api`
2. Submit a search request selecting the **Access to information and privacy** stream
3. Confirm the **Access to Information and Privacy (ATIP)** is added to the request
4. Submit a request selecting any stream other than **Access to Information and Privacy**
5. Confirm the **Digital community** is assigned to the request


api/graphql/schema.graphql
```diff
@@ -1203,6 +1205,11 @@ input DepartmentBelongsToMany {
   sync: [ID!]
 }
 
+input CommunityBelongsTo {
+  connect: UUID
+  disconnect: Boolean
+}
```
tristan-orourke, line 1211: I think this should be more like SkillBelongsTo, where connect is required and there's no disconnect. The search requests aren't really meant to be edited after creation, so disconnect is a bit meaningless. (We could name this to make it more specific to the talent request. That's not a pattern we've followed much, but maybe we should...)
```suggestion
input CommunityBelongsTo {
  connect: UUID!
}
```

api/database/migrations/2024_05_21_141338_add_community_id_column_search_request_applicant_filter_tables.php
```diff
@@ -0,0 +1,90 @@
...
+            ->select('id', 'key')
+            ->where('key', 'atip')
+            ->first();
+
+        $digital = DB::table('communities')
+            ->select('id', 'key')
+            ->where('key', 'digital')
+            ->first();
+
+        $rows->each(function ($row) use ($ATIP, $digital) {
```
tristan-orourke, line 42: I think this could all be done without a foreach loop, which will be very slow given many requests. You just need to  combine your UPDATE sql statements with the join statement from line 28, and you can add a `WHERE qualified_streams` clause.

tristan-orourke (changes requested):
Thanks for adding the community relationship to both models for now. I know its redundant and smelly, but I don't see an ideal solution here.

api/database/migrations/2024_05_21_141338_add_community_id_column_search_request_applicant_filter_tables.php
esizer: Ahh, that is a good idea! I think I got it in [update migration to avoid looping](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10432/commits/72ef8e23484231228958b516d624e70e1bd4e8af)

tristan-orourke (approved):
Looks good! Maybe leave a reminder for people in devspam that they'll need to run `php artisan db:seed --class=CommunitySeeder` before running migrations, or run `migrate:fresh`.

PR #10478, opened 2024-05-24
[Fix] `AssessmentResult` justification seeding

### LINKED ISSUE #10477: 🐛 `AssessmentResult` inaccurate justification seeding
## 🐛 Bug

The `AssessmentResultTestSeeder` is creating inaccurate database entries.

## 🦋 Expected Behaviour

Expected data is seeded.

## 🕵️ Details

The seeder uses `json_encode` before passing it in which is making laravel duplicate the serialization making the data a `string` instead of `json`.

## 📋 Steps to Reproduce

1. Run the seeder `make seed-fresh`
2. Observe the justification column has a mix of strings and json

## 📸 Screenshot

screenshot_24-May-2024_09-41-1716558077

## 🙋‍♀️ Proposed Solution

Avoid using `json_encode` in seeder and allow laravel to serialize the data for us.

## ✅ Acceptance Criteria

- [ ] Seeded justifications are `json` and not `string`

esizer:
🤖 Resolves #10477 

## 👋 Introduction

Fixes an issue where `AssessmentResult.justification` data was being serialized twice during seeding.

## 🕵️ Details

We were missing `->name` accessor on an enum in the justification array which required us to use `json_encode` since PHP cannot serialize non-backed enums (which ours are).

## 🧪 Testing

1. Seed fresh data `make seed-fresh`
2. Confirm the `justification` column only contains `json` or `null` and no `string`s.

### DIFF
--- a/api/database/seeders/AssessmentResultTestSeeder.php
+++ b/api/database/seeders/AssessmentResultTestSeeder.php
@@ -157,7 +157,7 @@ class AssessmentResultTestSeeder extends Seeder
             $publishedPool->poolSkills()->pluck('id')->toArray(),
             $assessmentStep,
             null,
-            [AssessmentResultJustification::SKILL_FAILED_INSUFFICIENTLY_DEMONSTRATED],
+            [AssessmentResultJustification::SKILL_FAILED_INSUFFICIENTLY_DEMONSTRATED->name],
             AssessmentDecision::HOLD->name,
             assessmentResultType::SKILL);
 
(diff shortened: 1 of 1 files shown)


mnigh (commented):
The justification column only contains json and no strings. Are there meant to be any rows that are seeded with `null` (there are not currently)?

esizer:
> The justification column only contains json and no strings. Are there meant to be any rows that are seeded with `null` (there are not currently)?

That i am not sure about :sweat_smile: I'll ask in standup :+1:

petertgiles:
Decision in standup: should be nullable and null in some situations

PR #10503, opened 2024-05-24
[Debt] Separate search index table

vd1992:
## 👋 Introduction

Create a separate table to store the `searchable` column for users. 

## 🕵️ Details

A mix of reading
- https://github.com/devNoiseConsulting/laravel-scout-postgres-tsvector/blob/master/README.md
- Postgres documentation
- Stack Overflow

With trial and error. 
Performance impact should be minimal, but the `selectableColumns()` work around is no longer needed. 

## 🧪 Testing

1. Navigate to general search
2. It still works
3. Test migration forward and back

## 📸 Screenshot

image

image

### DIFF
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -95,54 +95,6 @@ class User extends Model implements Authenticatable, HasLocalePreference, Laratr
     use Searchable;
     use SoftDeletes;
 
-    protected static $selectableColumns = [
-        'id',
-        'email',
-        'first_name',
-        'last_name',
-        'telephone',
-        'sub',
-        'preferred_lang',
-        'current_province',
-        'current_city',
-        'looking_for_english',
-        'looking_for_french',
-        'looking_for_bilingual',
-        'first_official_language',
-        'second_language_exam_completed',
-        'second_language_exam_validity',
-        'comprehension_level',
-        'written_level',
-        'verbal_level',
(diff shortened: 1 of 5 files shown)


api/app/Models/User.php
```diff
@@ -156,67 +109,32 @@ class User extends Model implements Authenticatable, HasLocalePreference, Laratr
...
+    public function searchableOptions()
     {
-        return self::$selectableColumns;
+        return [
+            // You may want to store the index outside of the Model table
+            // In that case let the engine know by setting this parameter to true.
+            'external' => true,
+            // If you don't want scout to maintain the index for you
+            // You can turn it off either for a Model or globally
+            'maintain_index' => false,
```
vd1992, line 124: Updating that runs automatically was pointing at `user.searchable` and failing
So I added `static::updated()` below

petertgiles (commented):
I've just done a general read of this.  I really like how the "searchable columns" stuff is gone but I don't love adding a new model for the search index.  It seems like this shouldn't be part of the data layer and should be an implementation detail of users.  Can we get away with not adding it?  It looks like you're not really using `searchableAdditionalArray` which is where I'd expect this to be implemented?  Let me know if you'd like me to dive into this a little more deeply.

vd1992:
Replaced by #10516

PR #10504, opened 2024-05-24
[Fix] Anonymous view of communities

### LINKED ISSUE #10476: 🐛 Can't submit talent request when not logged in
## 🐛 Bug

It's currently not possible to submit a talent request when not logged in.

## 🦋 Expected Behaviour

Should be able to submit a talent request when not logged in.

## 📋 Steps to Reproduce

1. Log out
2. Try to submit a talent request

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

Seems to have something to do with the canModel guard on communities.  If that guard is removed it works OK.

## ✅ Acceptance Criteria

- [ ] Can submit a talent request when not logged in.
- [ ] e2e test

esizer:
🤖 Resolves #10476 

## 👋 Introduction

This fixes the communities policy to allow anonymous users to view any.

## 🕵️ Details

The policy omitted the option user argument that lead to requiring the policy to be authenticated. It seems odd, but we need to include `?User $user` in the arguments to allow anonymous users (even if we never use it).

REF: https://laravel.com/docs/11.x/authorization#guest-users

## 🧪 Testing

1. Ensure you are not logged in
2. Confirm you can submit a search request

### DIFF
--- a/api/app/Policies/CommunityPolicy.php
+++ b/api/app/Policies/CommunityPolicy.php
@@ -15,7 +15,7 @@ class CommunityPolicy
      *
      * @return \Illuminate\Auth\Access\Response|bool
      */
-    public function viewAny()
+    public function viewAny(?User $user)
     {
         return true;
     }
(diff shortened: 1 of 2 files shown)


JamesHuf (approved):
It is kind of a weird fix but it works for me.

PR #10505, opened 2024-05-24
[Spike] Email Verification - Code

### LINKED ISSUE #10375: 🏐 Secure email validation
## 🏐 Spike

To perform email validation, we need to securely generate, send and validate codes sent to users by email.

(There are other ways to validate emails, it's possible for us to do something other than sending codes, but let's only change that approach if there's a good reason to)

## 🕵️ Details

We will eventually need a generator and validator for these, so if this spike produced those, it would be a bonus. We should be referencing some cyber standards in the solution for this.

Or maybe Laravel has it all built in :) That'd be nice.

## ❓ Questions to answer

- How do we build a secure email validation with Laravel?
- How much work would it take to use codes instead of built in Laravel links?
- How long are the links/codes Laravel generates valid for?
- If using links, how do we preserve the browsing state (send the user back to where they came from (or the next step).
- Are we secure enough?
- How do we throttle these? (does Laravel handle this?)
- How do we invalidate previous codes/links when requesting a new one or the email changes? (does laravel handle this)
- Can it handle multiple email addresses for a user?

## Output
Proof of concept email verification using Laravel + GC notify.

## 🕙 Timebox
1.5 days

petertgiles:
🤖 Resolves #10375 

## 👋 Introduction

This is a sample solution to the "Email Verification" spike using codes.  It is not very similar to the traditional Laravel way but is closer to the way our app normally works.

> [!CAUTION]
This is a spike and should not be merged.  It can be used as a reference for future PRs.

## 🕵️ Details

The branch works by storing something resembling a JWT in state and sending the key to that token to the user in an email.  If they are able to send the key back to us in a GraphQL mutation that we consider the token valid and mark the email address as valid.

I've used Laravel Cache as the data store for this since it is fun, lightweight, has built in expiry, and the data is transient by nature.  But we could easily store it in the database if we prefer, too.

I like that this method leverages our already-existing auth framework and GraphQL API.

This solution has a lot of copy-and-paste for clarity between work email and regular email.  A prod-ready solution would probably factor some of that out to make it more DRY.

This is one option.

> How much work would it take to use codes instead of built in Laravel links?

I think this is just as simple as links.  Perhaps more since it is more consistent with our other code.  If we wanted to _also_ send a link we could send it as a query param like talent.canada.ca/en/verify-email?code=1234

> How long are the links/codes Laravel generates valid for?

I set six hours on this example but you can make the code expire whenever you want.

> If using links, how do we preserve the browsing state (send the user back to where they came from (or the next step).

You could store the "next step" in the token and have it get returned from the mutation so that the app can navigate on success.

> Are we secure enough?

I think this is very secure since it uses our already-built auth framework.  If the user changes their email address or provides a code from another user the mutation will fail.  We could probably sign the token being saved but that's probably overkill since the user never sees it or touches it.

> How do we throttle these? (does Laravel handle this?)

This uses our existing GraphQL API which already has throttling in place.

> How do we invalidate previous codes/links when requesting a new one or the email changes? (does laravel handle this)

This does not invalidate previous codes sent.  However, when the code is redeemed it checks that the email address still matches the one in the request so it doesn't really matter.

> Can it handle multiple email addresses for a user?

Yes, this shown in the branch.

## 🧪 Testing

1. Set the GCNotify API key to the teams one and set the template ID:
   - GCNOTIFY_TEMPLATE_VERIFY_EMAIL_EN=c99fb24a-8e2d-4510-8569-6f9e39939323
2. Set the email values on your user.  Try tinker for this:
```php
$user = User::find('0129f316-eca9-4880-99fa-f810827d55c0')
$user->email = 'your email that exists in GCNotify'
$user->save()
$user->sendEmailVerificationNotification()
```
3. Get the code from your email and insert it using Graphiql
```graphql
mutation MyMutation {
  verifyEmail(code: "1234") {
    email
    emailVerifiedAt
    workEmail
    workEmailVerifiedAt    
  }
}
```
4. Observe that `emailVerifiedAt` is set
5. Repeat exercise for workEmail if desired
  
## 📸 Screenshot

image


petertgiles:
Discussed in standup.  Decided to go with this solution.
Suggested improvements:
- Use the user ID as the cache key.  Then a request for a new code will destroy a previous request.
- Use a larger codespace. Use more digits or alphanumeric characters.

petertgiles:
Was suggested that encoding a redirect link in the token was unnecessary.  Should just redirect back to profile landing page if needed.

PR #10506, opened 2024-05-24
 [Spike] Email Verification - Link

### LINKED ISSUE #10375: 🏐 Secure email validation

petertgiles:
🤖 Resolves #10375 

## 👋 Introduction

This is a sample solution to the "Email Verification" spike using links.  It is close to the traditional Laravel way but is less like the way our app normally works.

> [!CAUTION]
This is a spike and should not be merged.  It can be used as a reference for future PRs.

## 🕵️ Details

The branch works by creating a link with the user ID and hashed email address then sharing it as a temporary signed link.  It is pretty close to the regular Laravel way of working but I've had to remove User authentication from the solution since an email link can not add a bearer token to the HTTP GET request.

This solution uses REST since that's the traditional Laravel model.  I wasn't able to use that actual, built-in components from Laravel since it assumes there is only one email, that you want to verify it immediately on creation, that you're authenticating with cookies, and that you want to use the built-in email service.  However, the pieces I wrote are closely based on the Laravel ones.

This solution has a lot of copy-and-paste for clarity between work email and regular email.  A prod-ready solution would probably factor some of that out to make it more DRY.

This is one option.

> How much work would it take to use codes instead of built in Laravel links?

This wouldn't be easy with this solution - the Laravel solution is not really set up for this.  I suppose it might be possible to create a page on the app that builds the GET request if the user pastes in user ID, email hash, and signature.  :shrug: 

> How long are the links/codes Laravel generates valid for?

You can make the temporary signed links expire whenever you want.

> If using links, how do we preserve the browsing state (send the user back to where they came from (or the next step).

The Laravel route has a placeholder redirect to "/home" at the end but you could substitute in a query param from the link if you wanted.

> Are we secure enough?

It makes me uncomfortable that we had to remove the user authentication from this but I think it's still secure with the signed link.

> How do we throttle these? (does Laravel handle this?)

It's easy enough to add a throttle to a Laravel route.  We've done that before

> How do we invalidate previous codes/links when requesting a new one or the email changes? (does laravel handle this)

This does not invalidate previous links sent.  However, when the code is redeemed it checks that the email address still matches the one in the request so it doesn't really matter.

> Can it handle multiple email addresses for a user?

Yes, this shown in the branch.

## 🧪 Testing

1. Set the GCNotify API key to the teams one and set the template ID:
   - GCNOTIFY_TEMPLATE_VERIFY_EMAIL_EN=cad8b9c8-c009-4b88-9d3f-393886015b4f
2. Set the email values on your user.  Try tinker for this:
```php
$user = User::find('0129f316-eca9-4880-99fa-f810827d55c0')
$user->email = 'your email that exists in GCNotify'
$user->save()
$user->sendEmailVerificationNotification()
```
3. Get the link from your email request it in your browser.
4. Observe that the redirect happened and `emailVerifiedAt` is set
5. Repeat exercise for workEmail if desired
  
## 📸 Screenshot

image


petertgiles:
Discussed in standup.  Was decided to go with the "code solution".

PR #10514, opened 2024-05-27
[Fix] Removes superfluous spaces in French string

mnigh:
## 👋 Introduction

This PR removes superfluous spaces in a French string.


esizer (approved):
LGTM! :+1:

PR #10516, opened 2024-05-27
[Debt] Separate search index table second try

### LINKED ISSUE #9518: ♻️ Move the column used to textsearch out of Users table and into a new table
## ♻️ Debt/Refactor

To make textsearch work, we added a column to the Users table and use Laravel\Scout\Searchable to automatically update this column and index it for searching.

However, the way Laravel Eloquent works, if you ask it for a model it always retrieves all columns from the database. It always runs `select * from`, so that you can run the db query once and access all columns as properties. Since this new column is huge, this is probably slowing down every query which uses the User table.

## 🕵️ Details

If we can, lets move the textsearch column into a separate table, linked to users with a foreign id. Hopefully we can configure Laravel Scout in the User.php model to make this refactor invisible to the api.

## 🙋‍♀️ Proposed Solution

Not yet. Look into Laravel Scout.

## ✅ Acceptance Criteria

- [ ] The full text search via the api is completely unaffected.
- [ ] The column indexed for full text search is stored in its own table

vd1992:
🤖 Resolves #9518

## 👋 Introduction

Create a separate table to store the searchable column for users.

## 🕵️ Details

Thank you @petertgiles for pointing out how I should have actually been using `searchableAs()`

A mix of reading
-    https://github.com/devNoiseConsulting/laravel-scout-postgres-tsvector/blob/master/README.md
-    Postgres documentation
-    Stack Overflow

With trial and error.
Performance impact should be minimal, but the `selectableColumns()` work around is no longer needed.

## 🧪 Testing

1. Navigate to general search
2. It still works
3. Test migration forward and back

## 📸 Screenshot

image

image


api/database/migrations/2024_05_27_202609_user_search_index_table.php
```diff
@@ -0,0 +1,57 @@
...
+            $table->uuid('id')->primary('id')->default(new Expression('gen_random_uuid()'));
+            $table->foreign('id')->references('id')->on('users')->onUpdate('cascade')->onDelete('cascade');
+        });
+
+        DB::statement('DROP INDEX users_searchable_index');
+        DB::statement('ALTER TABLE user_search_indices ADD COLUMN searchable TSVECTOR');
+        DB::statement('CREATE INDEX users_searchable_index ON user_search_indices USING GIN(searchable)');
+
+        // copy existing values before dropping column
+        $users = DB::table('users')->select(['id', 'searchable'])->get();
```
petertgiles, line 26: Try to avoid using looping to work with database operations.  It's slow and since it's not atomic you might miss rows if there's databse activity while it's running.  Try to do something like 
```sql
insert into user_search_indices (id, searchable)
select id, searchable from users
```

vd1992: Okay, seems reasonable to me. Done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/10516/commits/604bfbe9b22c7b399995428eef7ccef0065141d9

petertgiles: Looks great!

petertgiles (approved):

PR #10518, opened 2024-05-28
[Debt] Migrate pool candidate test to Playwright

### LINKED ISSUE #10481: ♻️ Move e2e to Playwright - pool-candidate.cy.ts
## ♻️ Debt/Refactor

Time to move our Cypress tests to Playwright
apps/e2e/cypress/e2e/admin/pool-candidate.cy.ts

## ✅ Acceptance Criteria

- [ ] Test moved to Playwright: pool-candidate.cy.ts

## 🛑 Blockers

```[tasklist]
### Blocked By
```

esizer:
🤖 Resolves #10481 

## 👋 Introduction

Migrates the pool candidate test from Cypress to Playwright.

> [!NOTE]
> This also caught and fixed a bug with the expiry date formatting

> [!IMPORTANT]
> This test was actually skipped because it was testing the old page so it does not do exactly what the old test was doing.

## 🧪 Testing

1. Confirm test is passing
2. Confirm test covers a similar case as the previous cypress test


apps/playwright/tests/admin/pool-candidate.spec.ts
vd1992, line 1: I'd like a few more assertions to test the page a bit more thoroughly. Such as:
- Verifying the snapshot was loaded, so "Profile snapshot not found." does not appear or asserting snapshot values are present
- Updating candidate notes
- You assert the expiry date after recording final decision, but also assert status appears as "Qualified"
- There are quite a few mutation operations you can chain on that page, maybe after record final decision, test you can then place and then remove

vd1992 (changes requested):

apps/playwright/tests/admin/pool-candidate.spec.ts
esizer: This is a simple migration of a test that checked that we could change the status. We were not meant to add additional assertions. Can you create tickets to expand the coverage?

vd1992: Oh, I thought it was a rewrite

esizer: Oh no, we are just trying to get the same coverage from cypress to playwright. I like those assertions though and we should probably get them in eventually!

vd1992: Issue created here https://github.com/GCTC-NTGC/gc-digital-talent/issues/10523

vd1992 (approved):
Great catch with the date string formatting!

PR #10525, opened 2024-05-29
[Debt] Refactor pool candidate combined status

### LINKED ISSUE #10360: ♻️ Clean up poolCandidateCombinedStatus
## ♻️ Debt/Refactor

We have a utility file called poolCandidateCombinedStatus, which basically just generates the chips (colour and text) which show application status to applicants.

This is _exactly_ the same kind of thing we're doing in another file, `utils/poolCandidates`, except those status chips are shown on the admin side. Additionally, `poolCandidateCombinedStatus` uses its own set of statuses, called CombinedStatus, which only exists in this one place.

I'm sure there was good reason to do it like this initially, but I believe the use of these similar functions in different files, and the existence of the extra layer of CombinedStatus, now only confuses things and makes it easier for the two sides to evolve in different directions.

## 🕵️ Details

## 🙋‍♀️ Proposed Solution

Move the code in poolCandidateCombinedStatus into utils/poolCandidates. Rewrite the applicant-side chip generating functions to follow the same pattern as the admin ones - rename the two sets of functions so they clearly denote "applicantFacing" or "adminFacing"

## ✅ Acceptance Criteria

- [ ] `poolCandidateCombinedStatus.tsx` no longer exists
- [ ] `CombinedStatus` frontend type no longer exists
- [ ] Applicant- and admin-facing status chips are unchanged

yonikid15:
🤖 Resolves #10360 

## 👋 Introduction

- Removed CombinedStatus type
- Moves applicant facing code to poolCandidate utils file.

## 🧪 Testing

1. Build app `pnpm run dev`
2. Login as `admin@test.com`
3. Create an application
4. Switch btw the statuses of the pool candidate and isSuspended field
5. Ensure the status text and chip styling shows up correctly on the applicant card

## 📸 Screenshot

image
image
image


brindasasi (approved):
Verified all statuses. This is looking great to me. Awesome work!🙌

PR #10551, opened 2024-05-30
[Feature] Local-only makefile for Linux

### LINKED ISSUE #9838: 🛠️ Setup & refresh scripts that don't use the maintenance container
## 🛠️ Tooling

Some people might not want to use the maintenance container for setup and refresh.  Perhaps they want to avoid having all the files owned by `root` or find that Docker is slow or uses too much memory.  Let's figure out how they can set up and maintain their workspace without the extra container.

## 🙋‍♀️ Proposed Solution

- Parallel scripts that can be easily diffed with the main ones?
- Another command-line option to change the behavior of the existing scripts?
- Just document the script?

- Can be hooked into vscode build tasks?

## ✅ Acceptance Criteria

- [ ] It is easy to setup and refresh the project without the maintenance container
- [ ] Regular maintenance container setup continues to work locally and in CI

petertgiles:
🤖 Resolves #9838 

## 👋 Introduction

This branch adds a makefile especially for Linux environments where the right versions of php, composer, node, and pnpm are already installed.  It does not use the maintenance container and, as a result, does not have any files owned by root.  It also does not use any of the maintenance scripts.

Unfortunately, our environment requires setting the ownership to www-data on certain files so there is a sudo prompt in the "setup" script.

The scripts will fail if any files are already owned by root.  I recommend cleaning your workspace first with the `git_clean` script.  WARNING - it will delete all uncommitted files!  It also has a sudo prompt.

I found a VS Code extension (https://marketplace.visualstudio.com/items?itemName=carlos-algms.make-task-provider) that works nicely with it.  It allows you to pick the name of the makefile that you want to use with it.

If all this is too niche and no one else finds this useful I can close this and I'll submit a PR with just a gitignore entry and I'll keep using it out of my own stash.

## 🧪 Testing

`make --file Makefile.nix setup_all`
`ext install carlos-algms.make-task-provider`

## 📸 Screenshot

image

image

### DIFF
--- a/Makefile.nix
+++ b/Makefile.nix
@@ -1,6 +1,6 @@
 # Usage:
 # setup_all          # setup all the components
-# refresh_all        # setup all the components
+# refresh_all        # refresh all the components
 # make setup_api     # build the Laravel backend
 # make refresh_api   # refresh the Laravel backend
 # make setup_web     # build the React frontend
--- a/Makefile.nix
+++ b/Makefile.nix
@@ -1,0 +1,62 @@
+# Usage:
+# setup_all          # setup all the components
+# refresh_all        # setup all the components
+# make setup_api     # build the Laravel backend
+# make refresh_api   # refresh the Laravel backend
+# make setup_web     # build the React frontend
+# make refresh_web   # refresh the React frontend
+# make git_clean     # use git to remove all untracked files - DANGER!
+# make compose_up   # bring up the docker compose network
+# make compose_down # bring down the docker compose network
+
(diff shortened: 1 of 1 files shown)


Makefile.nix
```diff
@@ -0,0 +1,62 @@
+# Usage:
+# setup_all          # setup all the components
+# refresh_all        # setup all the components
```
petertgiles, line 3:
```suggestion
# refresh_all        # refresh all the components
```

petertgiles (commented):
self-review

esizer (approved):
I can't test because I have nothing installed locally but I'm happy to let this through to hopefully make DX better for people! :+1:

PR #10568, opened 2024-05-31
[Debt] API query assessment plan completeness

### LINKED ISSUE #10263: ♻️ Compute pool completeness on server
## ♻️ Debt/Refactor

We would like to move the pool completeness logic to the server to simplify the front end and reduce the amount of business logic contained there.

## 🕵️ Details

Right now we have some business logic in the client to determine if a pool is complete. It is semi-complicated and not easily determined just by looking at the code. It would be nice to be able to just query this from the API as a boolean.

## 🙋‍♀️ Proposed Solution

I'm not sure how discrete we want to get but something to start from would be:

```graphql
type Pool {
 # ...existingFields
 
 advertisementComplete: Boolean;
 assessmentPlanComplete: Boolean;

 # Maybe?
 advertismentBasicInfoComplete: Boolean; # etc.
}
```

## ✅ Acceptance Criteria

- [ ] Pool model updated with completeness attributes
- [ ] Business logic on front end moved to new attributes
- [ ] Front end queries update with new fields

vd1992:
🤖 Resolves #10263

## 👋 Introduction

Shift computing the plan completeness server-side. 
Basically just needed to pull out the validation that checked plan completeness before publishing into its own thing then use that. Straightforward for the most part. 

## 🧪 Testing

1. Process, assessment plan, and publishing unchanged
2. Work through creating and publishing a process to check this


api/app/Models/Pool.php
```diff
@@ -307,6 +308,24 @@ public function getIsCompleteAttribute()
...
+            $planCompletionValidation->messages()
+        );
+
+        if ($validator->fails()) {
+            return false;
+        }
+
+        return true;
+    }
+
```
esizer, line 328: It seems a little odd to use validation to compute a field. I think we should be reserving validation for its purpose of checking user input rather than computing an attribute :thinking:

vd1992: We've been using it for pool completeness too. 

As well, I think it makes sense as it is reusing shared logic. As in, pool completeness is required for publishing, therefore, the completion status as well validation in publishing both check the same thing. Makes sense to me to re-use rather than rewrite out the validator

yonikid15 (changes requested):
Oh no :cry: something broke the chip status. I think since you removed the other data being fetched in the `AssessmentPlanStatus` query, the assessmentSteps contain outdated data. I think we could add the AssessmentStep type (I think) to the [OperationContext to re-execute the query](https://commerce.nearform.com/open-source/urql/docs/api/urql/#usequery) in the `AssessmentPlanBuilderPage.tsx`




https://github.com/GCTC-NTGC/gc-digital-talent/assets/22059495/96726b38-c024-42e9-af41-ffdd7ebc1ef5

vd1992:
> Oh no 😢 something broke the chip status. Since you removed the other data being fetched in the `AssessmentPlanStatus` query, the assessmentSteps contain outdated data. I think we could add the AssessmentStep type (I think) to the [OperationContext to re-execute the query](https://commerce.nearform.com/open-source/urql/docs/api/urql/#usequery) in the `AssessmentPlanBuilderPage.tsx`

Thanks for the review
I couldn't quite replicate that locally, but added in your suggestion in https://github.com/GCTC-NTGC/gc-digital-talent/pull/10568/commits/fc77899b346fb538a74e918c13611d35d25a9f66
Hopefully that works

yonikid15 (changes requested):
It's still not perfect, if the assessment plan is complete then you add a new behavioural skill and go back to the assessment plan page it still says complete. I'm looking for a solution, but no luck yet.

vd1992:
> Update: Oh we missed adding the same thing to the PoolLayout.tsx page as well!

I literally just did that now 😆 
See https://github.com/GCTC-NTGC/gc-digital-talent/pull/10568/commits/c680e5626862d4db72ba80eb2c2716e8272b0b18

yonikid15 (approved):
Nice work :medal_military:

PR #10571, opened 2024-05-31
[Debt] Priority weight enum

### LINKED ISSUE #8668: ♻️ `PoolCandidateSearchInput.priorityWeight` using `Int`

yonikid15:
🤖 Resolves #8668 

## 👋 Introduction

- Adds PriorityWeight enum
- Updates PoolCandidateSearchInput to accept enum instead of the associated integer weights to help make it more readable on the frontend.
- Question: I've added a static function which contains a map of the enum value to it's associated weight, and returns the weight (int). Should we also update the `user.priorityWeight` field to be the PriorityWeight enum as well, and call that method when needed?

## 🧪 Testing

1. Build app `make setup` 
2. Login as `admin@test.com`
3. Go to pool candidates table and ensure the "Category" filter (priorityWeight) works correctly.

### DIFF
--- a/apps/web/src/utils/poolCandidate.ts
+++ b/apps/web/src/utils/poolCandidate.ts
@@ -30,6 +30,7 @@ import {
   Maybe,
   PoolCandidateStatus,
   PublishingGroup,
+  PriorityWeight,
 } from "@gc-digital-talent/graphql";
 import {
   getOrThrowError,
@@ -715,3 +716,19 @@ export const getCombinedStatusLabel = (
     statusLabelKey,
     `Invalid statusLabelKey '${statusLabelKey}'`,
   );
+
+export const getPriorityWeight = (priorityWeight: number): PriorityWeight => {
+  if (priorityWeight === 10) {
+    return PriorityWeight.PriorityEntitlement;
+  }
+
+  if (priorityWeight === 20) {
+    return PriorityWeight.Veteran;
+  }
+
+  if (priorityWeight === 30) {
+    return PriorityWeight.CitizenOrPermanentResident;
+  }
+
+  return PriorityWeight.Other;
+};
--- a/api/app/Enums/PriorityWeight.php
+++ b/api/app/Enums/PriorityWeight.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace App\Enums;
+
+enum PriorityWeight
+{
+    case PRIORITY_ENTITLEMENT;
+    case VETERAN;
+    case CITIZEN_OR_PERMANENT_RESIDENT;
+    case OTHER;
+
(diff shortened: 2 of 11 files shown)


apps/web/src/utils/poolCandidate.ts
```diff
@@ -715,3 +716,19 @@ export const getCombinedStatusLabel = (
...
+  if (priorityWeight === 20) {
+    return PriorityWeight.Veteran;
+  }
+
+  if (priorityWeight === 30) {
+    return PriorityWeight.CitizenOrPermanentResident;
+  }
+
+  return PriorityWeight.Other;
+};
```
esizer, line 734: Just to be clear your sugegstion of:

> Should we also update the user.priorityWeight field to be the PriorityWeight enum as well, and call that method when needed?

Would make this function not necessary, correct? Since when we query for `user.priorityWeight` we would get the enum back instead? :thinking: I like that since working with numbers directly is kind of confusing and the transformation should be in one spot to reduce error.

esizer (commented):
This looks great, thanks for making that change! I just had one comment based on your suggestion. But, if you want to break that off into another ticket then I'd also be okay with that since you did cover the AC in this.

apps/web/src/utils/poolCandidate.ts
yonikid15: Ya this wouldn't be needed anymore since we can get the `PriorityWeight` enum directly from the query. I'd rather make another PR since we want to run it by other devs first.

esizer (approved):
LGTM! :+1:

PR #10578, opened 2024-06-03
[Debt] Merge snapshot functions

### LINKED ISSUE #10392: ♻️ Merge PoolCandidate.php functions:  setApplicationSnapshot + createSnapshot
## ♻️ Debt/Refactor

We currently have two different functions for generating snapshots.

SubmitApplication calls setApplicationSnapshot but the seeder and test call createSnapshot.  It would really be good to merge them and only have one codepath.

## 📸 Screenshot

Image

## ✅ Acceptance Criteria

- [ ] One function to generate snapshots
- [ ] Snapshots still made on submission
- [ ] Tests still test
- [ ] Seeders still seed

## 🛑 Blockers

```[tasklist]
### Blocked By
```

vd1992:
🤖 Resolves #10392 

## 👋 Introduction

Merge/consolidate the snapshot functions. 

## 🕵️ Details

Effectively just deleted `createSnapshot` which was smooth sailing to remove. 

Seeding needed a tweak as it wasn't actually saving the snapshots so almost all candidates had null snapshots. Just needed the 
`->save()`
Added a simple test too, just to assert there is no overwriting. That explicitly is a null check to ensure no oddities with truthy evaluation. 

## 🧪 Testing

1. Data seeds fine
2. Application submission creates a snapshot normally


api/app/Models/PoolCandidate.php
```diff
@@ -585,59 +585,6 @@ public function getPoolCandidateStatusAttribute($candidateStatus)
...
-            'poolCandidates.generalQuestionResponses.generalQuestion',
-            'poolCandidates.screeningQuestionResponses',
-            'poolCandidates.screeningQuestionResponses.screeningQuestion',
-        ])->findOrFail($this->user_id);
-        $profile = new UserResource($user);
-
-        // collect skills attached to the Pool to pass into resource collection
-        $pool = Pool::with([
-            'poolSkills',
-            'classification',
```
esizer, line 630: Just a question, `classification` does not appear in the `setApplicationSnapshot` was it not needed or just an oversight?

api/app/Models/PoolCandidate.php
```diff
@@ -585,59 +585,6 @@ public function getPoolCandidateStatusAttribute($candidateStatus)
...
-        ])->findOrFail($this->pool_id);
-        $poolSkillIds = $pool->poolSkills()->pluck('skill_id')->toArray();
-
-        $profile = new UserResource($user);
-        $profile = $profile->poolSkillIds($poolSkillIds);
-
-        $this->profile_snapshot = $profile;
-        $this->save();
```
esizer, line 638: This is also missing from the `setApplicationSnapshot`. Is there a scenario where we need to set the snapshot without saving it?

esizer (commented):
Looks good to me, I just had two minor questions first.

api/app/Models/PoolCandidate.php
vd1992: Saving pool info deemed unneeded in snapshotting

api/app/Models/PoolCandidate.php
vd1992: Decided to be done
Completed in https://github.com/GCTC-NTGC/gc-digital-talent/pull/10578/commits/83edf4e434cd3d533dd9e30c53e3494501e4f98e

esizer (approved):
Nice! :sunglasses:

PR #10619, opened 2024-06-07
[Fix] Adds conditional for skill accordion without `skillLevel`

### LINKED ISSUE #10612: 🐛 Old job advertisements show skill level: Not Found on the production
## 🐛 Bug

We have introduced Skill Level on Job advertisement recently. We have some historic job advertisements on the production those are closed showing the Level: Not Found. This causes frustration between Applicant vs Hiring manager.

Image

## 🦋 Expected Behaviour

When SkillLevel is null don't show it in the UI.  Leave "Assessed during initial application" without any bullet point 

## 📋 Steps to Reproduce

1. Create a job advertisement & publish it 
2. Go to DB and manually make skill level as null for that
3. View the advertisement 

## ✅ Acceptance Criteria

- [ ] If skill level is null , Level is not shown on the UI
- [ ] If skill level is not null, Level is shown on the UI

mnigh:
🤖 Resolves #10612.

## 👋 Introduction

This PR adds a conditional statement around the **Level: {skillLevel}** and **Level definition** text for legacy processes that do not have skill levels.

## 🧪 Testing

### SQL query

```sql
SELECT skill_id, required_skill_level FROM pool_skill WHERE pool_id = $pool_id
```

1. `pnpm build`
2. Navigate to a job advertisement `/browse/pools/:pool-id#skill-requirements`
3. Find from the `:pool-id` in the `pool_skill` database table and set `required_skill_level` to `NULL` (see previous SQL query section)
5. Navigate to a job advertisement `/browse/pools/:pool-id#skill-requirements`
6. Verify changed skill accordion with  `NULL` skill level doesn't render **Level: {skillLevel}** text in the accordion trigger or **Level definition** paragraph in the accordion content

## 📸 Screenshot
Screen Shot 2024-06-07 at 16 10 09

### DIFF
--- a/apps/web/src/pages/Pools/PoolAdvertisementPage/components/SkillAccordion.tsx
+++ b/apps/web/src/pages/Pools/PoolAdvertisementPage/components/SkillAccordion.tsx
@@ -163,18 +163,21 @@ const SkillAccordion = ({ poolSkillQuery, required }: SkillAccordionProps) => {
             {getLocalizedName(poolSkill.skill.description, intl)}
           </p>
         )}
-        <p>
-          <span data-h2-font-weight="base(700)">
-            {intl.formatMessage({
-              defaultMessage: "Level definition",
-              id: "fqa45V",
-              description: "Label for the definition of a specific skill level",
-            }) + intl.formatMessage(commonMessages.dividingColon)}
-          </span>
-          {definitionAndLevel
-            ? intl.formatMessage(definitionAndLevel.definition)
-            : intl.formatMessage(commonMessages.notFound)}
-        </p>
+        {poolSkill.requiredLevel && (
+          <p>
+            <span data-h2-font-weight="base(700)">
+              {intl.formatMessage({
+                defaultMessage: "Level definition",
+                id: "fqa45V",
+                description:
+                  "Label for the definition of a specific skill level",
+              }) + intl.formatMessage(commonMessages.dividingColon)}
+            </span>
+            {definitionAndLevel
+              ? intl.formatMessage(definitionAndLevel.definition)
+              : intl.formatMessage(commonMessages.notFound)}
+          </p>
+        )}
       </Accordion.Content>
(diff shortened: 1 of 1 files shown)


apps/web/src/pages/Pools/PoolAdvertisementPage/components/SkillAccordion.tsx
```diff
@@ -45,11 +45,13 @@ const Context = ({ required }: ContextProps) => {
 };
 
 interface AccordionSubtitleProps {
+  skillLevel: string | null | undefined;
```
petertgiles, line 48: This is a little weird - passing in the enum in addition to the formatted string just to know whether to display the formatted string.  Maybe this first prop could be switched to a bool like "isSkillLevelAvailable" ?  Or change skillLevelItem to be string | null and just test for null before displaying it?

petertgiles (approved):
This works well, I'm approving.  I just added a quick thought on how it could be refactored to consider.

apps/web/src/pages/Pools/PoolAdvertisementPage/components/SkillAccordion.tsx
mnigh: I agree, it feels weird. I've taken your suggestion though and made the prop a boolean instead with [823a3a5](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10619/commits/823a3a5017aaa787da4ebf85304ee61e41c9889b). 

The problem with `skillLevelItem` is that it will never be null as it is always has a value regardless of `skillLevel`'s value is, which is really the crux of #10612.

```
  const skillLevelItem = `${`${
    intl.formatMessage({
      defaultMessage: "Level",
      id: "bVRixs",
      description: "Label displayed on the classification form level field.",
    }) + intl.formatMessage(commonMessages.dividingColon)
  } ${skillLevel}`}`;
```

PR #10622, opened 2024-06-08
[Spike] Example implementation of back-end localized enums

### LINKED ISSUE #10029: 🏐 How Can We Manage Localization On The Backend?
## ✨ Feature

With backend report generation, we'll need add localization so that reports can be generated in both English and French. In addition to the report template needing to be localized, there is all the "localized constants" logic that turns an enum to a string that needs to be available on the backend, too.

## ❓ Questions

Q: How hard will it be to maintain a totally separate localization pipeline for the backend?

Q: Is it possible to move all the localized constants logic to backend so that all the strings and mapping don't need to be maintained and synchronized between the two?

Q: Is there any way to share messages between the front and the back so we don't have to maintain duplicate message files?

Q: Should we use Laravel's localization library?

##  :beverage_box: Timebox

3 days

esizer:
🤖 Resolves #10029 

## 👋 Introduction

This is an example implmentation of how we could localized database enum fields on the back-end using laravels built in localization.

## 🕵️ Details

This uses a trait (`HasLocalization`) that can be implemented on enums that need localization. The implmentation has some rough edges around it but should work well, I think :sweat_smile: .

### How it works

The trait requires that the enum implements a `getLangFileName` method that returns a string for the localized file name (found in `app/lang/{locale}`.

The trait also provides a `localizedString` method that accepts a value (name of the enum case) and an optional intermediary key (for multiple translations of each key.

We can then use this in whatever way we want. I have an example of it being passed through graphql incase we would like to avoid duplicating this strings on the frontend. 

For stuff that does not exist in the database (csv column headers, PDF headings) we may need to bite the bullet for now and just duplicate them :cry: 

### Common strings

Some of our enums use a common string. This shows the citizenship status simply accessing the common string of "Other". It is a bit verbose but shouldn't need to be done that often so I am personally fine with this approach.

#### Fallback/Fault tolerance

I haven't figured out a way to ensure we have translations yet so if you do not provide a translation it will either fallback to the default locale (English) or, just display the "ID" (`citizenship_status.profile.citizen`).

#### Inflation

This may lead to some inflation of the schema, data we are retrieving from the API and model methods. Instead of retrieving just the enum value, we are now grabbing (possibly multiple) localized string representations of the data and need these to be defined on the model. 

However, this might be offset by the reduction of the compiled strings we get anyway when we load the app and make it more "on demand" :thinking: 

### How hard will it be to maintain a totally separate localization pipeline for the backend?

I don't see this as very difficult to maintain as a pipeline since it is mostly handled by laravel.

### Is it possible to move all the localized constants logic to backend so that all the strings and mapping don't need to be maintained and synchronized between the two?

Yes, I tihnk this clearly demonstrates that we can move all of the localized constant logic to the backend for localization. I picked an exmaple that shows both multiple strings for a single enum value as well as a common string that is shared.

### Is there any way to share messages between the front and the back so we don't have to maintain duplicate message files?

I am currently unable to see how we can share non-database backed strings easily. While we can translate them, sending them to the frontend in an elegant way eludes me at the moment.

### Should we use Laravel's localization library?

Yes, I beleive this demonstrates the ease of use that laravels built in localization provides for our use case.

## 🧪 Testing

```graphql
{
  me {
    profileCitizenship {
      value
      label {
        en
        fr
      }
    }
    adminCitizenship {
      value
      label {
        en
        fr
      }
    }
  }

 profileCitizenshipStatusStrings: localizedEnumStrings(
   enumName: "CitizenshipStatus",
  intermediaryKey: "profile"
  ) {
   value
   label {
     en
     fr
   }
 }
 
  adminCitizenshipStatusStrings: localizedEnumStrings(
    enumName: "CitizenshipStatus", 
    intermediaryKey: "admin"
 ) {
   value
   label {
     en
     fr
   }
 }
}
```

1. Update your `api/.env` file to set the default API user
2. Navigate to `/graphiql`
3. Run the provided query
4. Observe the localized versions of the string

## :camera_flash: Screenshot

image

### DIFF
--- a/api/app/Enums/CitizenshipStatus.php
+++ b/api/app/Enums/CitizenshipStatus.php
@@ -12,9 +12,9 @@ enum CitizenshipStatus
     case CITIZEN;
     case OTHER;
 
-    public static function getLangFilename(?string $fileKey = 'profile'): string
+    public static function getLangFilename(): string
     {
 
-        return 'citizenship_status.'.$fileKey;
+        return 'citizenship_status';
     }
 }
--- a/api/app/Enums/CitizenshipStatus.php
+++ b/api/app/Enums/CitizenshipStatus.php
@@ -2,9 +2,19 @@
 
 namespace App\Enums;
 
+use App\Traits\HasLocalization;
+
 enum CitizenshipStatus
 {
+    use HasLocalization;
+
     case PERMANENT_RESIDENT;
     case CITIZEN;
     case OTHER;
+
+    public static function getLangFilename(?string $fileKey = 'profile'): string
+    {
+
+        return 'citizenship_status.'.$fileKey;
+    }
 }
--- a/api/app/GraphQL/Queries/LocalizedEnumStrings.php
+++ b/api/app/GraphQL/Queries/LocalizedEnumStrings.php
@@ -1,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\GraphQL\Queries;
+
+final readonly class LocalizedEnumStrings
+{
+    /** @param  array{}  $args */
+    public function __invoke(null $_, array $args)
+    {
+
+        $enum = 'App\\Enums\\'.$args['enumName'];
+
+        return array_map(function ($case) use ($args, $enum) {
+            return [
+                'value' => $case,
+                'label' => $enum::localizedString($case, $args['intermediaryKey']),
+            ];
+        }, array_column($enum::cases(), 'name'));
+
+    }
+}
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -347,6 +347,16 @@ class User extends Model implements Authenticatable, HasLocalePreference, Laratr
         }
     }
 
+    public function getAdminCitizenshipAttribute()
+    {
+        return CitizenshipStatus::localizedString($this->citizenship, 'admin');
+    }
+
+    public function getProfileCitizenshipAttribute()
+    {
+        return CitizenshipStatus::localizedString($this->citizenship, 'profile');
+    }
+
     public function getLookingForLanguage()
     {
         if ($this->looking_for_bilingual) {
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -349,12 +349,18 @@ class User extends Model implements Authenticatable, HasLocalePreference, Laratr
 
     public function getAdminCitizenshipAttribute()
     {
-        return CitizenshipStatus::localizedString($this->citizenship, 'admin');
+        return [
+            'value' => $this->citizenship,
+            'label' => CitizenshipStatus::localizedString($this->citizenship, 'admin'),
+        ];
     }
 
     public function getProfileCitizenshipAttribute()
     {
-        return CitizenshipStatus::localizedString($this->citizenship, 'profile');
+        return [
+            'value' => $this->citizenship,
+            'label' => CitizenshipStatus::localizedString($this->citizenship, 'profile'),
+        ];
     }
 
     public function getLookingForLanguage()
--- a/api/app/Rules/LocalizedEnumExists.php
+++ b/api/app/Rules/LocalizedEnumExists.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace App\Rules;
+
+use Closure;
+use Database\Helpers\ApiErrorEnums;
+use Illuminate\Contracts\Validation\ValidationRule;
+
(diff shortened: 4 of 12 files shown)


petertgiles:
Very cool!  I love the use of traits.  Two questions come to mind as I read through it.
1) What does the failure path look like?  What happens if I request something that doesn't exist in the lang files?
2) How do I get a list of all of the strings of an enum on the frontend, like for a dropdown list?

esizer:
> 1. What does the failure path look like?  What happens if I request something that doesn't exist in the lang files?

You just get the "id" so for this instance assuming no key exists at all. So, you would just see `citizenship_status.profile.citizen`. Otherwise it would fallback to our defined locale (assuming English here like on `react-intl`. 

I feel like this is the largest area of improvement for this approach. I would like to provide a single fallback or throw something. I did not get that far yet though :sweat: 

> 2\. How do I get a list of all of the strings of an enum on the frontend, like for a drop down list?

I think we could do a maybe a whole graphql definition like:

```graphql
type CitizenshipStatusString {
   value: CitizenshipStatus;
   label: LocalizedString;
}

type EnumStrings {
   citizenshipStatuses: [CitizenshipStatus!];
  # ...
}

type Query {
  enumStrings: EnumStrings;
}
```

I think the PHP might look something like this: 

```php
final class EnumStrings
{
   public function __invoke()
   {
       return [
          'citizenshipStatus' => [
              [
                  'value' => CitizenshipStatus::CITIZEN->name,
                  'label' => CitizenshipStatus::localizedString($this->citizenship, 'profile');
              ]
          ]
       ];
   }
}
```

This may be a little verbose though so maybe there is a way to abstract it a bit? :thinking:

petertgiles:
> We could pass the enum we want into args and use that? 🤔

Oh, I like that updated idea a lot!  Much lighter weight.  I'd like to see if we can avoid another big map like we have in GraphQLServiceProvider.php already.  Since all our enums are in the same namespace I wonder if we can do something like `('App\Enums\'.$args['enumName'])::cases()`

esizer:
> Oh, I like that updated idea a lot! Much lighter weight. I'd like to see if we can avoid another big map like we have in GraphQLServiceProvider.php already. Since all our enums are in the same namespace I wonder if we can do something like `('App\Enums\'.$args['enumName'])::cases()`

Thanks for the input! I got a working example of how we could do this now. [add endpoint for getting all strings for localized enum](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10622/commits/e2401ce1baf56f5052385a758818ff96b93d964a)

It includes some validation so you must put in an existing enum name that has to implement the `localizedString` method. I was trying to ensure the trait was implemented but it seemed to only be working with classes and not enums :cry:

petertgiles:
> I got a working example of how we could do this now.

I love it!  It turned out so good.  :clap:

petertgiles:
I wonder if we might want to have the type fields return LocalizedEnumString instead of LocalizedString.  We might want the base enums in some cases on the frontend for business logic, like certain selections causing some form fields to appear.

esizer:
> I wonder if we might want to have the type fields return LocalizedEnumString instead of LocalizedString. We might want the base enums in some cases on the frontend for business logic, like certain selections causing some form fields to appear.

Oh yeah, I can see how that would be helpful :+1: 

Updated with that in mind [update to return value and label for localized enums](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10622/commits/5544817265b4ad979eda7f31ff80233ab61addf8)

tristan-orourke (commented):
Nice work! I love the HasLocalization trait. The LocalizedEnumStrings implementation is very clever, combined with the validation rule.

The one thing here which makes me nervous is the concept of the _intermediary key_. To me, this approach of having a backend-defined readable enum value makes the most sense if we only have a single canonical value. I like the example you picked, citizenship_status. The localized values for `citizenship_status.admin` seem straightforward, canonical. If you just showed me the enum values in a vacuum, that's how I'd translate them. But the copy for `citizenship_status.profile` is specific to a particular way of using the enum, in a particular place on the UI. Seems to me that in this case, the client code should ignore the backend values if it doesn't like them, and handle its own mapping of enum-to-copy for the specific component.

esizer:
> The one thing here which makes me nervous is the concept of the _intermediary key_.

Yeah, that makes sense. I don't think this is super common anyway. I was trying to show how we could provide alternative translations if we wanted to do. But, I think you are right, we can just do this when necessary.

PR #10626, opened 2024-06-10
[Debt] Make some assessment decisions null

### LINKED ISSUE #10498: ♻️ Seed some `null` `AssessmentResult.justifications`
## ♻️ Debt/Refactor

We would like to seed some null values for the `AssessmentResult.justifications` field since it is a realistic value to expect.

## 🕵️ Details

Right now we only seed results with justifications. Since it is feasible to have this field be `null`, we would like that represented in our seeded data.

## 🙋‍♀️ Proposed Solution

Update the `AssessmentResultFactory` to see `null` on `justifications` when appropriate.

## ✅ Acceptance Criteria

- [ ] At least one `null` value in the `justification` column of the `assessment_results` table when appropriate

esizer:
🤖 Resolves #10498 

## 👋 Introduction

Makes some justifications on `AssessmentResult` `null` when seeding fresh test data.

## 🕵️ Details

When the result decision is "unsure" or "on hold", the justification should be null in most cases. This updates the seeder to create more accurate data in that scenario instead of always setting a justification regardless of decision

## 🧪 Testing

1. Re-seed fresh data `make seed-fresh`
2. Login to the database at `http://localhost:8080`
3. Confirm that some `justifications` are `null` when the `assessment_decision` is either `null` or `HOLD`

## 📸 Screenshot

screenshot_10-Jun-2024_14-23-1718043791

### DIFF
--- a/api/database/factories/AssessmentResultFactory.php
+++ b/api/database/factories/AssessmentResultFactory.php
@@ -38,7 +38,7 @@ class AssessmentResultFactory extends Factory
             'pool_candidate_id' => PoolCandidate::factory(),
             'assessment_result_type' => $assessmentResultType,
             'assessment_decision' => $assessmentDecision,
-            'justifications' => $justifications,
+            'justifications' => $this->nullJustifications($assessmentDecision) ? null : $justifications,
             'assessment_decision_level' => $assessmentResultType === AssessmentResultType::SKILL->name &&
                 $assessmentDecision === AssessmentDecision::SUCCESSFUL->name ?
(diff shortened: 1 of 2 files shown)


esizer:
> This works for most rows but I'm seeing the users like [absent@test.com](mailto:absent@test.com) and [barely@test.com](mailto:barely@test.com) com still seeded with justifications. Could you take a look?

Good catch! I was able to find another seeder where we were overridiing the factory :+1: 

Fixed in [null justifications in results test seeder](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10626/commits/d70fa63d22a0a3563e1ba12c94b07d36212facd2)

petertgiles (approved):

PR #10642, opened 2024-06-11
[Feature] Compute candidate current step on back end

### LINKED ISSUE #8966: ✨ Compute "final decision" and "current assessment step" for Pool Candidates on backend
## ✨ Feature

While a Pool Candidate is in assessment, we can consider which steps have been completely assessed and which require further assessments. A step could also be unsucceessful.

It would be useful to compute this on the backend via database queries, since this would allow us to sort candidates by "Current step" and simplify some frontend components.

This will enable #8960.

## 🕵️ Details

Assessment step status is determined by:
```
Essential requirements = essential skills AND education requirements
Optional requirements = asset skills
Assessment Step Status =
  IF any essential requirements unsuccessful THEN "unsuccessful"
  IF any requirements unassessed THEN "to assess"
  IF essential requirements include at least one hold THEN "hold"
  IF all essential requirements successful THEN "success"

Also, on the step tracker, a step effectively only has the "to assess" status if all previous steps are "hold" or "success".
```

The "current step" is 
- null (not applicable) if all steps are successful or any steps are unsuccessful
- OR the number of the first "to assess"/"hold" step

### Sort order
The order of statuses should be as following:

Qualified
Qualified pending decision
To assess step 1
To assess step 2
....
Disqualified
Disqualified pending decision
Removed

## 🎨 Design File

### 🧑‍🎨 Designer

@Jerryescandon 

## 📸 Screenshot

Look at how we implemented Skill counts for pool candidate search. We can do something like that, I think.

For sorting, either sort like the "skill count" column, or an additional generated column for sorting.

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] When getting one or more pool candidates from the api, you can request a `currentStep: number` field
- [ ] New enum to represent: Final decision status 
- [ ] Way of sorting that includes both final decision and current step
- [ ] You can get the final decision status from the api

🛑 Blockers
```[tasklist]
### Blocked by
- [ ] #10110
```

esizer:
🤖 Resolves #8966 

## 👋 Introduction

Moves the computation of a candidates current step to the back end and updates the frontend to use new fields.

## 🕵️ Details

This creates and event that is fired when an `AssessmentResult` is saved (created or updated) that triggers a handler to compute and store a new column in the `PoolCandidate` database table for `AssessmentStatus`. 

This new field contains the candidates current assessment step (`null` indicates passed all steps), their current "final" decision and an array of the computed decision for each step.

We can now query those fields from the API and use them directly instead of needing to compute them every time on the client.

```graphql
query CandidateAssessmentStatus($id: UUID!) {
  poolCandidate(id: $id) {
    id
    assessmentStatus {
      finalDecision
      currentStep
      stepDecisions {
        step
        decision
      }
    }
  }
}
```

## 🧪 Testing

1. Login as `admin@test.com`
2. Create a pool with multiple assessment steps (including some essential skills)
3. Have a user apply to the pool
4. Test different scenarios for each step
5. Confirm that the values are updated in the database as expected


api/app/Models/PoolCandidate.php
```diff
@@ -834,4 +839,112 @@ public function setApplicationSnapshot()
...
+            'assessmentResults',
+        ]);
+
+        foreach ($this->pool->assessmentSteps as $step) {
+            $stepId = $step->id;
+            $hasFailure = false;
+            $hasOnHold = false;
+            $hasToAssess = false;
+
+            $stepResults = $this->assessmentResults()->where('assessment_step_id', $stepId);
```
tristan-orourke, line 857: Minor suggestion - I think this change will mean it only runs one database query, and filters the cached results. Since the list shouldn't be that long, I expect this to be faster than multiple DB queries.
```suggestion
            $stepResults = $this->assessmentResults->where('assessment_step_id', $stepId);
```

api/app/Models/PoolCandidate.php
```diff
@@ -834,4 +839,112 @@ public function setApplicationSnapshot()
...
+            if (! $stepResults->count()) {
+                $decisions[$stepId] = null;
+
+                continue;
+            }
+
+            // Check assessed essential skills on this step
+            $essentialSkillAssessments = $stepResults->whereHas('assessmentStep', function ($query) {
+                $query->whereHas('poolSkills', function ($query) {
```
tristan-orourke, line 866: Make sure to preload `assessmentResults.assessmentStep.poolSkills`

api/app/Models/PoolCandidate.php
```diff
@@ -834,4 +839,112 @@ public function setApplicationSnapshot()
...
+
+            if (! $essentialSkillAssessments) {
+                $decisions[$stepId] = null;
+
+                continue;
+            }
+
+            foreach ($essentialSkillAssessments as $essentialSkillAssessment) {
+                $decision = $essentialSkillAssessment->assessment_decision;
+                if (! $decision) {
```
tristan-orourke, line 879: Can this not be `case null` in the switch statement? (legitimate question - I'm not sure)

api/app/Models/PoolCandidate.php
```diff
@@ -834,4 +839,112 @@ public function setApplicationSnapshot()
...
+            }
+
+            // Check assessed essential skills on this step
+            $essentialSkillAssessments = $stepResults->whereHas('assessmentStep', function ($query) {
+                $query->whereHas('poolSkills', function ($query) {
+                    $query->where('type', PoolSkillType::ESSENTIAL->name);
+                });
+            });
+
+            if (! $essentialSkillAssessments) {
```
tristan-orourke, line 871: What if an assessment step doesn't _have_ any required skills?

Also, remember that the initial step (APPLICATION_SCREENING) also requires an education-type assessment result. 
EDIT: nevermind, I see you handle that later 😆

api/app/Listeners/ComputeFinalDecisionAndCurrentStep.php
```diff
@@ -0,0 +1,67 @@
...
+        $result->load([
+            'poolCandidate',
+            'poolCandidate.pool.assessmentSteps',
+        ]);
+
+        $candidate = $result->poolCandidate;
+
+        $decisionsPerStep = $candidate->computeDecisionsPerAssessmentStep();
+
+        $currentStep = 1;
```
tristan-orourke, line 33: Are we not zero-indexing? 😆 😕 😖

api/app/Listeners/ComputeFinalDecisionAndCurrentStep.php
```diff
@@ -0,0 +1,67 @@
...
+        $orderedSteps = $candidate->pool->assessmentSteps()->orderBy('sort_order', 'ASC')->get();
+        foreach ($orderedSteps as $step) {
```
tristan-orourke, line 35: Didn't we add some fancy logic for how HOLDs interacted with current step. I think we landed on "if the final step is HOLD, treat it like TO ASSESS". But I'll confirm...

api/storage/app/lighthouse-schema.graphql
```diff
@@ -434,6 +434,16 @@ type AssessmentResult {
   skillDecisionNotes: String
 }
 
+type AssessmentStepDecision {
+  step: UUID!
```
tristan-orourke, line 438: The more "GraphQL" version of this would be
```suggestion
  step: AssessmentStep!
```

api/app/Listeners/ComputeFinalDecisionAndCurrentStep.php
esizer: Wait, I can't tell if this is a joke or not :sweat_smile: Do you want me to start at 0 ? :stuck_out_tongue: 

I did 1 since that is how sort order words as well and it is easier for comparison.

api/app/Models/PoolCandidate.php
esizer: oph right, we changed it so that we can do asset only right? So instead of setting and continue I should be doing `$hasToAssess = true;` instead, right?

api/app/Models/PoolCandidate.php
esizer: `null` can, I need to check on `undefined`. Or is it impossible for this to be `undefined`? :thinking:

api/app/Models/PoolCandidate.php
esizer: Good catch! Added in [preload pool skills for decision computation](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10642/commits/4b521d5b631871b3e9e769578158241230ca1dea)

api/app/Listeners/ComputeFinalDecisionAndCurrentStep.php
tristan-orourke: It started as a joke, and then I realized I was actually confused about what to do here. On the one hand, my default assumption about any array-index-like number in code is that it starts from 0, so not doing that could be confusing and lead to misuse. On the other hand, matching sort order makes sense.

(Why didn't we start sort order from 0?)

I guess leave it starting from 1, but leave a comment in the schema that it starts from 1 not 0?

api/app/Models/PoolCandidate.php
tristan-orourke: I don't think PHP distinguishes between `null` and `undefined`, that's a javascript quirk 😆

api/app/Models/PoolCandidate.php
tristan-orourke: Actually... if a step has no essential skills, and its not APPLICATION ASSESSMENT (ie requires Education), then... is it automatically successful? If that's the case, you can do nothing here, and the function will return SUCCESSFUL correctly.

esizer: So just to confirm, you are suggesting we just delete this check all together and it should function as expected?

api/app/Models/PoolCandidate.php
esizer: I have been stuck in TS land too long :stuck_out_tongue: Should be fixed in [move null check to switch](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10642/commits/028f765344ac96c6b24be2168138a7ef894f977b)

api/app/Listeners/ComputeFinalDecisionAndCurrentStep.php
esizer: Alright, hopefully this is sufficient in communicating that.

[document current step in schema](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10642/commits/b83552a4213c25170d16fbff522655cb6be975b2)

api/app/Models/PoolCandidate.php
tristan-orourke: Yes.

tristan-orourke: This is exactly the kind of situation which should have a phpunit test

esizer: Cool, thanks! Yeah, my next task is writing tests to cover different scenarios now that I think the logic is worked out :+1:

esizer: Alright, tests added. I think I got all the logic covered but let me know if I am missing anything!

[add tests for logic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10642/commits/138c4be4cea3ba71a471e6e7938b98a45d8ffc1c)

esizer: I may need some assistance. The essential skill test is flaky and I really cannot tell why. I am guessing something to do with the seeded data but have not be able to locate it :confounded: 

https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/9504282927/job/26196583002?pr=10642#step:7:561

vd1992: Okay, I think I got it fixed in https://github.com/GCTC-NTGC/gc-digital-talent/pull/10642/commits/fc2ab6bbbe70f4bc9d71916f81f9ce3eb0d811f1
All successes on executing 
`for i in {1..20}; do docker-compose exec -w /home/site/wwwroot/api webserver sh -c "./vendor/bin/phpunit ./tests/Feature/CandidateAssessmentStatusTest.php"; done`

Issue being, the test skill would vary between technical and behavioural. Assessment step one I believe is always application screening, which ignores behavioural skills. So any time a behavioural skill was created, the screening step ignored it.

esizer: Thank-you! That makes sense now :pray:

api/app/Listeners/ComputeFinalDecisionAndCurrentStep.php
tristan-orourke: Okay, so here's the logic we currently have on the frontend. As I thought, we treat a HOLD on the final step as if it were TO ASSESS.

Oh, and we use zero indexing. 🤦‍♂️ Not sure what to make of that.

```
/**
 * Returns the "current step" a candidate should appear at in the assessment tracker.
 * A candidate should appear in all columns less than or equal to its current step.
 * A value of null means that a candidate has passed all steps.
 * @param assessmentToResult
 * @param assessmentOrdering
 * @returns
 */
const determineCurrentStep = (
  assessmentToResult: Map<AssessmentStepId, NullableDecision>,
  assessmentOrdering: string[],
): number | null => {
  for (let index = 0; index < assessmentOrdering.length; index += 1) {
    const assessmentStepId = assessmentOrdering[index];
    const result = assessmentToResult.get(assessmentStepId);
    if (result === AssessmentDecision.Unsuccessful || result === NO_DECISION) {
      return index;
    }
    // A candidate can be qualified with some Hold decisions, as long as they are followed by a Successful decision.
    // That means that if the final step is Hold, we treat it more like NO_DECISION.
    if (
      index === assessmentOrdering.length - 1 &&
      result === AssessmentDecision.Hold
    ) {
      return index;
    }
  }
  return null;
};
```

api/app/Providers/EventServiceProvider.php
```diff
@@ -14,8 +14,8 @@ class EventServiceProvider extends ServiceProvider
      * @var array<class-string, array<int, class-string>>
      */
     protected $listen = [
-        ApplicationSubmitted::class => [
```
tristan-orourke, line 17: Don't we need to keep the ApplicationSubmitted event?

esizer: It looks as though it actually wasn't in use. I'm not actually sure what it was removed though :thinking:

tristan-orourke (changes requested):
One additional change for the HOLD-on-last-step situation.

I'm also confused about the tests. The test pool is created with a single essential skill, right? So how does testSuccessPassesFirstStep work?

In terms of additional tests... here are some edge cases to check:
- A step tests only asset skills. It looks like our current behaviour is that the step is considered "to assess" until at least one skill has an assessment result, at which point it is considered Successful, even if the result is Unsuccessful or Not Sure. (This is a bit weird, but was intended to force assessors to at least interact with the step, even if it is for asset skills.)
- A step tests asset and required skills. The asset skills can be left empty, or unsure, or hold, or unsuccessful, and it doesn't affect the overall step result.
- The pool has no technical skills. This means only education assessment is required for the first step.
- Asset technical skills can also be ignored on the first step

apps/web/src/components/PoolCandidatesTable/helpers.tsx
```diff
@@ -206,13 +205,8 @@ export const currentLocationAccessor = (
...
   intl: IntlShape,
   poolCandidate: PoolCandidate,
-  poolAssessmentSteps: AssessmentStep[],
 ) => {
-  const { color, label } = getCandidateStatusChip(
-    poolCandidate,
-    unpackMaybes(poolAssessmentSteps),
-    intl,
-  );
+  const { color, label } = getCandidateStatusChip(poolCandidate, intl);
```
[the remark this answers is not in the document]
esizer: Yeah, that makes sense! Refactored in [refactor helper functions to only accept specific args](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10642/commits/ff18f9978d60c5047cfa0ea0b26dd09e5c83ee33)

api/tests/Feature/CandidateAssessmentStatusTest.php
```diff
@@ -0,0 +1,344 @@
...
+
+        AssessmentResult::factory()
+            ->withResultType(AssessmentResultType::EDUCATION)
+            ->create([
+                'assessment_step_id' => $stepOne->id,
+                'pool_candidate_id' => $this->candidate->id,
+                'assessment_decision' => AssessmentDecision::SUCCESSFUL->name,
+            ]);
+
+        $this->actingAs($this->adminUser, 'api')
```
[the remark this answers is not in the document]
esizer: Ah, yeah good catch. The essential skill assessment logic was not good. Fixed in [adjust essential skill assessment logic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10642/commits/526d35308dfb0126873f31285205e2a249ef4c24)

api/tests/Feature/CandidateAssessmentStatusTest.php
```diff
@@ -0,0 +1,344 @@
...
+                                    'step' => $steps[0]->id,
+                                    'decision' => AssessmentDecision::SUCCESSFUL->name,
+                                ],
+                                [
+                                    'step' => $steps[1]->id,
+                                    'decision' => AssessmentDecision::HOLD->name,
+                                ],
+                            ],
+                            'finalDecision' => AssessmentFinalDecision::TO_ASSESS->name,
+                            'currentStep' => null,
```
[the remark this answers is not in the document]
esizer: Okay, this should be fixed in [refactor to reduce number of loops needed](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10642/commits/199ce75cbbc528842d88785c32cefa52e872993e)

api/app/Listeners/ComputeFinalDecisionAndCurrentStep.php
```diff
@@ -0,0 +1,82 @@
...
+            if ($decision === AssessmentDecision::UNSUCCESSFUL->name) {
+                $containsDisqualified = true;
+            }
+        }
+
+        // Null step indicates the candidate passed all steps
+        if (is_null($currentStep)) {
+            $lastStepDecision = end($decisions);
+            if ($lastStepDecision['decision'] !== AssessmentDecision::HOLD->name) {
+                $finalDecision = AssessmentFinalDecision::QUALIFIED->name;
```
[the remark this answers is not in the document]
esizer: I think I was able to figure it out :sweat_smile: Refactored in [refactor to reduce number of loops needed](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10642/commits/199ce75cbbc528842d88785c32cefa52e872993e)

esizer:
> One additional change for the HOLD-on-last-step situation.
> 
> I'm also confused about the tests. The test pool is created with a single essential skill, right? So how does testSuccessPassesFirstStep work?
> 
> In terms of additional tests... here are some edge cases to check:
> 
> * A step tests only asset skills. It looks like our current behaviour is that the step is considered "to assess" until at least one skill has an assessment result, at which point it is considered Successful, even if the result is Unsuccessful or Not Sure. (This is a bit weird, but was intended to force assessors to at least interact with the step, even if it is for asset skills.)
> * A step tests asset and required skills. The asset skills can be left empty, or unsure, or hold, or unsuccessful, and it doesn't affect the overall step result.
> * The pool has no technical skills. This means only education assessment is required for the first step.
> * Asset technical skills can also be ignored on the first step

Okay, I like the ideas for different tests. Those might take a bit but in the mean time, I think I got all the other changes you requested.

api/app/Providers/EventServiceProvider.php
esizer: Oh, weird! I was a little confused as well. I guess we are not using event listeners to create snapshots anymore? I guess that sort of makes sense since we can just store it in the mutation itself since it can only happen from that one spot :woman_shrugging:

api/app/Models/PoolCandidate.php
```diff
@@ -911,4 +919,162 @@ public function setApplicationSnapshot()
...
+            $stepResults = $this->assessmentResults->where('assessment_step_id', $stepId);
+            if ($stepResults->isEmpty()) {
+                continue;
+            }
+
+            $hasEssentialSkillsToAssess = $step->poolSkills->contains(function ($poolSkill) {
+                return $poolSkill->type === PoolSkillType::ESSENTIAL->name;
+            });
+
+            if ($hasEssentialSkillsToAssess) {
```
tristan-orourke, line 949: So what happens if an assessment step has no essential skills (and doesn't require education)? See my earlier comment about that test case. Perhaps double check with Jerbo on what the desired behaviour is in that case.

api/app/Models/PoolCandidate.php
```diff
@@ -911,4 +919,162 @@ public function setApplicationSnapshot()
...
+
+        // While unlikely, current step could go over.
+        // So, set it back to total steps
+        if ($currentStep > $totalSteps) {
+            $currentStep = $totalSteps;
+        }
+
+        return [
+            'currentStep' => $currentStep,
+            'finalDecision' => $finalDecision,
```
tristan-orourke, line 1076: Hmm. It occurs to me now that "finalDecision" and "stepDecisions" are kindof misnomers... Neither of these fields is ever decided directly by a user - they're computed based on other fields. "finalDecision" especially is going to be confused with the "Final decision" that users actually make when they screen a candidate in or out, which is stored in poolCandidateStatus.

tristan-orourke (changes requested):
Looking good. Still have a couple specific concerns about the compute decision function.

Also, can we consider different name for finalDecision (and maaaybe stepDecisions)? Maybe put it to devspam?

api/app/Models/PoolCandidate.php
esizer: Yeah, I was kind of struggling with the naming here for the exact reason. Do you have any suggestions? I was thining of `assessmentDecision` but that exists on the `AssessmentResult` as well and could be conflated with that :disappointed: 

We just have so many layers to "decision" :sweat_smile: . Maybe the following works? Idk, I am at a loss :stuck_out_tongue: 

```
type AssessmentStatus {
   currentStep: Int
   qualificationStatus: AssessmentQualificationStatus! 
   stepAssessments: {
      step: UUID!
      decision: AssessmentDecision!
   }
}
```

tristan-orourke: Yeah its a tough one. 😅 
`assessmentStepStatus` and `overallAssessmentStatus`?
`computedDecision` and `overallComputedDecision`?

api/app/Models/PoolCandidate.php
```diff
@@ -911,4 +919,162 @@ public function setApplicationSnapshot()
...
+            });
+
+            if ($hasEssentialSkillsToAssess) {
+                // Check assessed essential skills on this step
+                $essentialSkillAssessments = $stepResults
+                    ->filter(function ($result) {
+                        return $result->poolSkill?->type === PoolSkillType::ESSENTIAL->name;
+                    });
+
+                if ($essentialSkillAssessments->isEmpty()) {
```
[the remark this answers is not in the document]
esizer: okay, updated in [adjust missing assessments and add test](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10642/commits/5f99183ce0d8438d7cd4e8ef984ed71dbae4edd9)

api/app/Models/PoolCandidate.php
```diff
@@ -911,4 +919,162 @@ public function setApplicationSnapshot()
...
+                }
+            }
+
+            // Check for education requirement if is application screening step
+            if ($step->type === AssessmentStepType::APPLICATION_SCREENING->name) {
+                $educationResults = $stepResults->where('assessment_result_type', AssessmentResultType::EDUCATION->name);
+
+                if (! $educationResults) {
+                    $decisions[] = [
+                        'step' => $stepId,
```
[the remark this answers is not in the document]
esizer: Test added in [adjust missing assessments and add test](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10642/commits/5f99183ce0d8438d7cd4e8ef984ed71dbae4edd9)

api/app/Models/PoolCandidate.php
esizer: Thanks for the suggestions, updated in [refactor names for accuracy](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10642/commits/b776ef4857c025be6fee39c47a26e58773fd719a)

api/tests/Feature/CandidateAssessmentStatusTest.php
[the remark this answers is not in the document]
tristan-orourke: Sorry, I'm still confused about this. The first step should be an APPLICATION_ASSESSMENT step, which always requires an education assessment as well as one for every required technical skill. Why is the skill assessment not required here?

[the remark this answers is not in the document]
esizer: Huh, yeah. I was able to get this to fail but I think things have changed since then. I'll nee to revisit it :confounded:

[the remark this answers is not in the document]
tristan-orourke: I just had a thought. When a Pool is created, the `created` event automatically adds an APPLICATION_SCREENING step. But I don't see the code which adds all technical skills to the APPLICATION_SCREENING step whenever they're added to the Pool. Are we missing a listener which does that, and ending up with an APPLICATION_SCREENING step which is missing some relevant related skills?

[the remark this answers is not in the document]
vd1992: That is done by this function

image

Called by

image

I believe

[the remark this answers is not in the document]
esizer: Okay, I found it.. even though the result is of type education, it still assesses the skill so it will pass. Since the UI does not allow for assessing skills on education I have added that as a qualifier for the essential skill assessments. 

[fix education with skill assessments affecting skill completion](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10642/commits/38abe59a092c9adc50821fa111dac87e03ad0618)

Now, if you comment out the skill assessment in this test, it will fail as expected.

api/app/Models/PoolCandidate.php
```diff
@@ -938,14 +938,28 @@ public function computeAssessmentStatus()
...
+                    'decision' => AssessmentDecision::SUCCESSFUL->name,
+                ];
+
+                $currentStep++;
+
+                continue;
+            }
+
+            if ($stepResults->isEmpty()) {
+                continue;
```
tristan-orourke, line 960: I don't think we can continue without adding _something_ to the decisions array. If we ended up here, that means there _are_ skill or education requirements, but no results, meaning we should add a TO_ASSESS decision, right?

api/app/Models/PoolCandidate.php
```diff
@@ -911,4 +919,167 @@ public function setApplicationSnapshot()
...
+                $decisions[] = [
+                    'step' => $stepId,
+                    'decision' => null,
+                ];
+
+                continue;
+            }
+
+            // Candidate has been assessed and was not unsuccessful so continue to next step
+            $currentStep++;
```
tristan-orourke, line 1038: I just realized, I think this will cause an incorrect currentStep in some cases.

Add a phpunit test for this case: There are two assessment steps. The first one is unassessed (or even unsuccessful). The second step is fully assessed, successfully. `currentStep` should be locked to step one. 

I'm worried that with the current implementation, the later steps in the loop will increment the `currentStep` past where it should get stuck.

tristan-orourke (changes requested):
Thought of another test case.

api/app/Models/PoolCandidate.php
esizer: You are correct. Added some logic to only increment when previous steps have been assessed and not been marked as unsuccessful

[do not increment step when previous steps not passed](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10642/commits/943049c819e067eddbd8a9a6a07310c5ca030780)

api/app/Models/PoolCandidate.php
esizer: Yeah, that makes sense. I think it was also required for the other comment you made so i rolled it into [do not increment step when previous steps not passed](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10642/commits/943049c819e067eddbd8a9a6a07310c5ca030780)

tristan-orourke (approved):
I think we're there!

PR #10680, opened 2024-06-13
[Debt] Squash DB migrations

### LINKED ISSUE #10187: ♻️ Squash database migrations
## ♻️ Debt/Refactor

As part of the Laravel 11 upgrade we're considering squashing all our existing database migrations.

https://laravel.com/docs/11.x/upgrade#modifying-columns
https://laravel.com/docs/10.x/migrations#squashing-migrations

This might have some performance and debt cleanup benefits as well.

## 🕵️ Details

Some considerations:
- Perhaps we want to practice a database restore again before attempting this?
- We'll need to carefully synchronize this change with our release schedule and any other PRs that might contain migrations before and after it gets squashed.
- We'll need to carefully test how it behaves, both with a fresh setup and an existing environment.

## ✅ Acceptance Criteria

- [ ] Migrations squashed
- [ ] Existing migrations no longer run for new environment setups
- [ ] Running `db:migrate` on existing environments (like UAT and prod) continues to be safe

## 🛑 Blockers

```[tasklist]
### Blocked By
- [ ] https://github.com/GCTC-NTGC/gc-digital-talent/issues/10204
- [ ] https://github.com/GCTC-NTGC/gc-digital-talent/issues/10205
```

petertgiles:
🤖 Resolves #10187 

## 👋 Introduction

This branch squashes the existing database migrations and creates a SQL file to recreate the database up to this point.

To maintain continuity, this branch is based right after the last release.  That means up-to-and-including migration 158 - 2024_06_03_173816_add_email_verification.  Any migrations after that should be applied normally.

## 🕵️ Details

According to [the docs](https://laravel.com/docs/10.x/migrations#squashing-migrations), this SQL file will only be applied if no migrations have been run in the environment yet, so it should have no effect on our Azure deployments.

The last section of the SQL file recreates the migration table up to the point of the dump so you can see which migrations are included.

Consider that we can't revert any migrations squashed in this operation.

## 🧪 Testing

- Rebuild the maintenance container `docker-compose build maintenance`

- [ ] Try effect on an existing setup
  1. Checkout and build a previous commit
  2. Checkout this branch
  3. `./artisan migrate`
- [ ] Try effect on a new setup
  1. `./artisan migrate:fresh`
- [ ] Can add subsequent migrations
  1. `./artisan make:migration test`
```php
    public function up(): void
    {
        Schema::table('users', function (Blueprint $table) {
            $table->string('junk')->nullable();
        });
    }
    public function down(): void
    {
        Schema::table('users', function (Blueprint $table) {
            $table->dropColumn('junk');
        });
    }
```
  2. `./artisan migrate`
  3. `./artisan migrate:fresh`


brindasasi (approved):
Perfect. This is working great in my local. Good to go 🥳

Issue #10675, opened 2024-06-13
✨ Claim rejection - remove indicator


petertgiles:
## ✨ Feature

If a priority or veteran claim is rejected then the row should no longer indicate that status in the pool candidate table or assessment tracker kanban.

## 📸 Screenshot

![image](https://github.com/GCTC-NTGC/gc-digital-talent/assets/8978655/b985a45a-e504-4bc7-b206-93ce82a75830)

## ✅ Acceptance Criteria

- [ ] If veteran claim is rejected they should no longer be indicated as veteran in the pool candidate table
- [ ] If veteran claim is rejected they should no longer be indicated as veteran in assessment tracker
- [ ] If priority claim is rejected they should no longer be indicated as priority in the pool candidate table
- [ ] If priority claim is rejected they should no longer be indicated as priority in assessment tracker

## 🛑 Blockers

```[tasklist]
### Blocked By
```

PR #10701, opened 2024-06-19
[Debt] Move enum localization to backend

### LINKED ISSUE #10634: ♻️ Move localized constants to backend
## ♻️ Debt/Refactor

Move our localized enum constants to the server.

## 🕵️ Details

We would like to start doing database value localization on the back end. So we are moving the `localizedConstants.tsx` file to laravel. A spike has been completed to figure out the approach to take 

## 🙋‍♀️ Proposed Solution

You can use the spike PR created to do the implementation. Keeping in mind, we do not want to use the `intermediaryKey` portion. Any custom translations should be done separately. 

 - https://github.com/GCTC-NTGC/gc-digital-talent/pull/10622

## ✅ Acceptance Criteria

- [ ] All (or most?) of the `localizedConstants.tsx` file has been moved to the server (laravel)
- [ ] Models that use the constants updated to return the new `LocalizedEnumString` instead of just the enum value
- [ ] Frontend updated to use the values from the server
- [ ] Add some PPHUnit tests to ensure the API validation and expected results occur

esizer:
🤖 Resolves #10634 

## 👋 Introduction

Moves the localization of specific enums to the backend.

## 🕵️ Details

This moves most of the strings found in localized constants to the backend using a few new tools. We have a new concept of a localized enum. This takes any PHP enum in the `App\Enums` namespace and adds a localized string to each case (assuming it exists).

In order to create a localized enum, there are a few things you must do:

1. Implement the `HasLocalization` trait on the eum (meaning defining a lang file)
2. Create the lang file with the name (in `/lang/{en|fr}.php` matching what you defined in the enum.
3. Add the strings in an associative array where the key is a lowercase version of the enum case you are localizing
4. Add the enum to the `$localizedEnums` array in the `GraphQLServiceProvider`
5. Mark the enum as localized in the graphql schema using the `@localizedEnum` directive
6. Update the frontend to use the new `Localized*` type

#### HasLocalization trait

This trait will force you to implement the `getLangFile` method which needs to return a string that represents the file name of the lang file. 

It then exposes a static function `localizedString` on the enum that takes a case of the enum and will return a localized string.

#### Lang files

These files are found in the `lang` folder and then organized again by locale. They simply contain associative arrays in a similar fashion to laravel config files. You just need to define a key that matches a lowercase version of an enum case and then the value is the string we want to render.

If a key for a speciific case does not exist in the lang file, it will first attempt to fallback to English and then, if that fails, just print out the expected key (i.e `citizenship_status.other`). Ideally we could find a way to provide a fallback message like "Not found" :thinking: 

##### Common strings

Of course, we have a few enums that share a common string like "Other". Thankfully, it is fairly simple to share these strings. We have a `common.php` file where you can store these and then access them using the `Lang` facade.

`Lang::get('common.key', [], 'locale');`

#### Graphql

We register these enums as a custom type in graphql in the `GraphQLServiceProvider`. Since these are all identical, all you need to do is define it in the array. We then loop through them and register it. The result is the following type (where `EnunName` is the name of the enum):

```graphql
type LocalizedEnumName {
   value: EnumName!
   label: LocalizedString!
}
```

##### Localized enum directive

Even though the type is registered, in order to actually use it, you will need to apply the `@localizedEnum` directive. This takes the value of the enum from the database and then converts it into the new type. If the key in graphql does not match the attribute on the PHP model, you can supply the `attribute` argument and it will behave in a similar way to the `@rename` directive.

##### Retrieving all strings for enum

Sometimes (usually in forms) we need to get all the strings of an enum for things like input options. In order to do this we have a simple query that just returns a basic version of the localized enum type. It has validation on the enum name which will fail if the enum does not exist in `App\Enums` or if it does not implement the `HasLocalization` trait.

```graphql
type LocalizedEnumString {
  value: String!
  label: LocalizedString!
}

  localizedEnumStrings(
    enumName: String!
      @rules(apply: ["required", "App\\Rules\\LocalizedEnumExists"]) <- ensures the enum name exists and is localized
  ): [LocalizedEnumString!]
  
# usage

query {
  # it is helpful to alias these so you can tell what it is in the client code
  citizenshipStatuses: localizedEnumStrings(enumName: "CitizenshipStatus") {
     value
     label {
       en
       fr
     }
  }
}
```

#### Mocking

Since we no longer have the strings for all these enums in the client code, we need to mock them for storybook and jest. We now have a helper in the `@gc-digital-talent/fake-data` package called `toLocalizedEnum`. This is basic helper that just takes the enum and converts it to the proper tpye signature making the label just the enum value replacing `_` with a space and making it lower case.

##### Forms

We often use all cases in an enum for input options. We have a helper `enumToOptions` that converts the cases to an options array. However, this will not work with localized enums. Now, we should use the new `localizedEnumToOptions` helper. Since we do not have strings in the client code, you will need to query them using the new query and pass them in.

```ts
<Select
   options={localizedEnumToOptions(dataFromApi, intl)}
   ...
 />
```

#### Others

We have some other helpers found in `@gc-digital-talent/i18n`. I'm not sure if this is the correct location but that is where they ended up. If we want to move them somewhere else, that can be done easily.

Sometimes, we have the value of an enum but not the localized version. We have two ways to convert them. One will return the entire localized enum type and the other just returns the localized string in the current locale.

```ts
const localizedEnum = getLocalizedEnumByValue(value, dataFromApi);

const label = getLocalizedEnumStringByValue(value, dataFromApi, intl);
```

Also, alot of these enums need specific sorting. We have all of these sort helpers there as well as a generic sorter that just takes an array for order and the data from the API.

### Maintaining the localized enum array

Having to maintain an array in the `GraphQLServiceProvider` is not great and I would like to find a way to "discover" any enum that is localized automatically. However, we do not add them that frequently so I think this could work for some time.

### Fallbacks

Right now if you do not provide a string it will just print out the key for the string. I would love if we could automatically fallback to either a "not found" string or just null :thinking: 

### Mocking

See the previous comment about mocking for details. tldr; We no longer have real enum strings in storybook or jest tests.

## 🧪 Testing

2. Clear cache `make artisan CMD="cache:clear"`
3. Build the app `pnpm run dev:fresh`
4. Review pages

## :mag: Review Plan

| Enum                                 | Reviewer | Done |
|--------------------------------------|----------|------|
| ArmedForcesStatus.php                | Peter         |      |
| AssessmentDecisionLevel.php          |    Peter      |      |
| AssessmentDecision.php               |        Peter  |      |
| AssessmentResultJustification.php    | Peter         |      |
| AssessmentResultType.php             | Peter         |      |
| AssessmentStepType.php               | Peter         |      |
| AwardedScope.php                     | Peter         |      |
| AwardedTo.php                        | Peter         |      |
| CandidateExpiryFilter.php            |     Peter     |      |
| CandidateRemovalReason.php           |    Peter      |      |
| CandidateSuspendedFilter.php         |     Peter     |      |
| CitizenshipStatus.php                |  VD        |      |
| EducationRequirementOption.php       | Matt          |      |
| EducationStatus.php                  | Matt          |      |
| EducationType.php                    | Matt          |      |
| EstimatedLanguageAbility.php         | Matt          |      |
| EvaluatedLanguageAbility.php         | Matt          |      |
| GenericJobTitleKey.php               | Matt          |      |
| GovEmployeeType.php                  | Matt          |      |
| IndigenousCommunity.php              | Matt          |      |
| LanguageAbility.php                  | Matt          |      |
| Language.php                         | Matt          |      |
| OperationalRequirement.php           | Yoni         |      |
| OverallAssessmentStatus.php          |  Yoni        |      |
| PlacementType.php                    | Yoni        |      |
| PoolCandidateSearchPositionType.php  | Yoni         |      |
| PoolCandidateSearchRequestReason.php | Yoni         |      |
| PoolCandidateSearchStatus.php        | Yoni         |      |
| PoolCandidateStatus.php              | Yoni         |      |
| PoolLanguage.php                     |  Yoni        |      |
| PoolOpportunityLength.php            | Yoni         |      |
| PoolSkillType.php                    | Yoni         |      |
| PoolStatus.php                       |     VD     |      |
| PoolStream.php                       |      VD|      |
| PriorityWeight.php                   |    VD      |      |
| ProvinceOrTerritory.php              |    VD      |      |
| PublishingGroup.php                  |   VD       |      |
| SecurityStatus.php                   |   VD       |      |
| SkillCategory.php                    |     VD     |      |
| SkillLevel.php                       |   VD       |      |
| WorkRegion.php                       |    VD      |      |


petertgiles:
I need to go through this more carefully but so far I really like it.  

Regarding the big array in GraphQLServiceProvider, I agree that it isn't great but isn't terrible either.  It should be possible to fix this - Laravel itself does a lot of "if a file exists I'll use it" magic.  At the same time, not terrible to have to do this manually for a bit.

Regarding the fallback for missing strings - I actually prefer the fallback to print the key.  I think that's a lot more useful for a user than a generic "not found" string.

brindasasi:
Just look for the enums usage in the code and test those pages for form submission , retrieving etc.,

api/app/Enums/ArmedForcesStatus.php
```diff
@@ -2,9 +2,18 @@
...
 {
+    use HasLocalization;
+
     case VETERAN;
     case MEMBER;
     case NON_CAF;
+
+    public static function getLangFilename(): string
+    {
+        return 'armed_forces_status';
```
petertgiles, line 17: Missing file?

api/app/Enums/AssessmentDecision.php
```diff
@@ -2,9 +2,18 @@
...
 {
+    use HasLocalization;
+
     case HOLD;
     case SUCCESSFUL;
     case UNSUCCESSFUL;
+
+    public static function getLangFilename(): string
+    {
+        return 'assessment_decision';
```
petertgiles, line 17: missing file?

api/lang/en/assessment_result_justification.php
```diff
@@ -0,0 +1,12 @@
...
+return [
+    'education_accepted_combination_education_work_experience' => 'Combination of education and work experience equivalency is accepted',
+    'education_accepted_information' => 'Education information is accepted',
+    'education_accepted_work_experience_equivalency' => 'Work experience equivalency is accepted',
+    'education_failed_not_relevant' => 'Not the right field or specialization',
+    'education_failed_requirement_not_met' => 'Not enough education or experience to meet the requirement',
+    'skill_failed_insufficiently_demonstrated' => 'Not sufficiently demonstrated to meet the requirement',
+    'failed_not_enough_information' => 'Not enough information provided',
+    'failed_other' => 'Other reason for screening out.',
+];
```
petertgiles, line 12: missing SKILL_ACCEPTED?

api/lang/fr/candidate_suspended_filter.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+use Illuminate\Support\Facades\Lang;
+
+return [
+    'active' => Lang::get('common.active', [], 'en'),
```
petertgiles, line 6: Shouldn't these be falling back to 'fr'?

api/app/Traits/HasLocalizedEnums.php
```diff
@@ -0,0 +1,30 @@
...
+
+trait HasLocalizedEnums
+{
+    protected function localizeEnum(?string $value, $enum)
+    {
+        if (! $value || is_object($enum) || ! method_exists($enum, 'localizedString')) {
+            return $value;
+        }
+
+        return [
```
petertgiles, line 13: Is this a fallback that, ideally, should not be hit?  Worth logging an error?

petertgiles (changes requested):
Part way through a review.  I'm going to stop now, just to let you know that Chromatic has flagged a bunch of things.  :grin:

esizer:
> Part way through a review. I'm going to stop now, just to let you know that Chromatic has flagged a bunch of things. 😁

Okay, I'm going to take a look but I do notice some of these are expected. Now that the actual strings are on the server we are essentially using fake data in storybook. We take the raw value generated from codegen and convert it from `PascalCase` to sentence case. I was trying to find a better solution but the consensus seemed to be "why are you trying to use data from your API in storybook, just mock it" :sweat_smile: 

I'm open to suggestions to make this better though!

api/app/Models/User.php
```diff
@@ -457,6 +458,20 @@ public function getPriority()
         return implode(', ', $priority);
     }
 
+    public function getPriorityAttribute()
```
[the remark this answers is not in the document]
esizer: I agree, it also makes me uneasy but we do use it on the applications :sweat: 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/a52c25ec4dfc8ba324ef8a6f618a1e81e1622bb7/apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx#L741-L742

api/app/Providers/GraphQLServiceProvider.php
```diff
@@ -71,15 +71,62 @@
...
 use GraphQL\Type\Definition\EnumType;
+use GraphQL\Type\Definition\ObjectType;
+use GraphQL\Type\Definition\Type;
 use Illuminate\Support\ServiceProvider;
 use Nuwave\Lighthouse\Schema\TypeRegistry;
 use Nuwave\Lighthouse\WhereConditions\Operator;
 
 class GraphQLServiceProvider extends ServiceProvider
 {
+    protected $localizedEnums = [
```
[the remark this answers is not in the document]
esizer: Thanks for reminding me! Done:

 - https://github.com/GCTC-NTGC/gc-digital-talent/issues/10861

api/app/Traits/HasLocalizedEnums.php
esizer: The inverse, this is the expected result. The if statement guards against converting non-localized enums. I'll add a log to that, just wondering if you have any input as the the severity we should log, I was thinking warning?

api/app/Enums/ArmedForcesStatus.php
esizer: Added in https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/e0143f0d1256e92bf43756ccdcb5397bdc95b781

api/app/Enums/AssessmentDecision.php
esizer: Added in https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/e0143f0d1256e92bf43756ccdcb5397bdc95b781

api/lang/en/assessment_result_justification.php
esizer: Hmm, we actually never had a translation for that one :thinking: 

@tristan-orourke or @Jerryescandon , do you know if this was intentional or did we just miss it? :sweat: 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/2c0c27f2e1ad20d3b4ed68ff26c548c9a464dba2/packages/i18n/src/messages/localizedConstants.tsx#L2082-L2134

api/lang/fr/candidate_suspended_filter.php
esizer: Good catch! I accidentally rolled it into this commit :sweat: 

https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/cd0f7f60fef51e493a7dd99455d905e856234927

api/app/Traits/HasLocalizedEnums.php
esizer: I'm not super happy with the guard, but we are now logging misuse :sweat_smile: 

[log attempt to localize non-localizable enums](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/1140f746cca83ff20d35e53c4144cab1285e451a)

api/app/Models/PoolCandidate.php
```diff
@@ -920,6 +932,94 @@ public function setApplicationSnapshot()
...
+
+        return $value;
+    }
+
+    /**
+     * Iterate through the snapshot and transform
+     * non-localized enum values into their localized
+     * version.
+     *
+     * NOTE: This is to handle legacy snapshots
```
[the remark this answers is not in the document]
esizer: Added in [add note on removing legacy enums in snapshot](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/fa2e3b429818a0d66dd9591e75bda70628dd0305)

I was going to add an optimistic date but wasn't confident we could even know that. Should be whenever we drop snapshots older than the merging of this PR though :ok_hand:

apps/web/src/utils/csvUtils.tsx
vd1992, line 1: Going to need to double check all the CSVs, appear to see a lot of blanks where there shouldn't be

image

vd1992 (changes requested):

apps/web/src/pages/Pools/EditPoolPage/components/CoreRequirementsSection/CoreRequirementsSection.tsx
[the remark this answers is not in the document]
esizer: :laughing: Matt actually found this when he was trying to publish a pool to test one of this enums. Should be fixed after [Fix SecurityStatus, PoolLanguage](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/a54324ebd4f769d43d11c337e21949759a0911f4)

api/lang/en/armed_forces_status.php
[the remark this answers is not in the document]
esizer: Good catch, turns out I was missing the files for translating this all along and @petertgiles noticed it :stuck_out_tongue: 

It should be fixed after [add missing files](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/e0143f0d1256e92bf43756ccdcb5397bdc95b781)

apps/web/src/utils/csvUtils.tsx
esizer: Good catch! Should be fixed after [fix csv values](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/b1a86864019ffb0e601fdc55693d750bb499cb14)[fix csv values](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/b1a86864019ffb0e601fdc55693d750bb499cb14)

vd1992:
Unrelated bugs found

Can't find archived pools https://github.com/GCTC-NTGC/gc-digital-talent/issues/10865

Not fetching skills library content in view candidate context https://github.com/GCTC-NTGC/gc-digital-talent/issues/10866

api/app/Enums/SecurityStatus.php
```diff
@@ -2,9 +2,18 @@
...
+    use HasLocalization;
+
     case RELIABILITY;
     case SECRET;
     case TOP_SECRET;
+
+    public static function getLangFilename(): string
+    {
+        return 'security_clearance';
+    }
```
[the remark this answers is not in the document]
vd1992: I see you went with SecurityStatus/security_status in https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/a54324ebd4f769d43d11c337e21949759a0911f4

apps/web/src/components/PoolCandidatesTable/helpers.tsx
vd1992, line 1: Something went a bit off the rails for bolding of `PriorityWeight` content

image

vd1992 (changes requested):

apps/web/src/components/PoolCandidatesTable/helpers.tsx
esizer: Okay, I figured out why it is not being bolded but I'm not sure what the expected result is here. The user has claimed "priority entitelement" but, it has been rejected on the application so we transofmr (on the frontend) the weight to 30 causing the bold to go away.

However, the label is on the server and is unaware of the new value. 

@tristan-orourke or @Jerryescandon  When a user has claimed a priority but rejected, do we still show that original claim or should we show something different here?

apps/web/src/utils/skillUtils.ts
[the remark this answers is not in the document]
esizer: ooo, looks like we were missing a field in the query (which was never there in the first place). This may also be a bug on main that we never noticed :sweat_smile: 

Fixed in [add missing query field](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/cafb2c96c9be3424b45946fbe0802252623baef1)

apps/web/src/components/PoolCandidatesTable/helpers.tsx
petertgiles: Yeah, the reject claim should change the value.  See https://github.com/GCTC-NTGC/gc-digital-talent/issues/10675 .
I didn't consider that that would be affected by this.  :grimacing:

esizer: Thanks :+1:  updated in [update priority column for candidates](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/b9045cf5032ae4060e8988cb4b4ac5f7ac16b644)

mnigh (changes requested):
This is a partial review to let you (@esizer) get ahead on some of these fixes while other parts are being reviewed.

api/lang/en/generic_job_title_key.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+return [
+    'analyst_it02' => 'Level 2: Analyst',
+    'executive_ex03' => '',
```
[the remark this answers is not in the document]
esizer: I don't think so but I just checked and we do not have an exisiting string for this. Do we know what it might be @tristan-orourke  or @Jerryescandon ?

https://github.com/GCTC-NTGC/gc-digital-talent/blob/1b1aef2f0caccf4bc440b2d81e8baa26fb2c53ab/packages/i18n/src/messages/localizedConstants.tsx#L892-L923

api/app/Enums/EducationRequirementOption.php
```diff
@@ -2,9 +2,18 @@
 
 namespace App\Enums;
 
+use App\Traits\HasLocalization;
+
 enum EducationRequirementOption
```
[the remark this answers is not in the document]
esizer: That is a good question. I did notice we have some enums that either do not match or have superfluous information added on see: `SecurityStatus`.

May be worth creating a ticket to resolve these inconsistencies :thinking:

api/app/Http/Resources/EducationExperienceResource.php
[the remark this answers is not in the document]
esizer: Fixed in [fix experience form default values](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/1482e56bf7bff6186ac3b96e952f30fbd48c53a7)

mnigh (changes requested):
I am finished with my list of enums. 

I did notice this loading behaviour for some slow loading. Is this new with moving enums to the backend or is it something I hadn't noticed before when it was coming from the frontend?

## Screen recording
https://github.com/GCTC-NTGC/gc-digital-talent/assets/3046459/10349086-dc65-4f06-8fdc-fd15007ab7ab

apps/web/src/components/PoolCandidatesTable/helpers.tsx
vd1992: Thanks, found another bug while I was at it 😛 https://github.com/GCTC-NTGC/gc-digital-talent/issues/10889

api/lang/en/generic_job_title_key.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
```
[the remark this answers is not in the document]
esizer: Honerslty, likely? I have always been confused by anything related to generic job titles :sweat_smile: @tristan-orourke Do you know if it is safe to remove these? :thinking:

vd1992 (approved):
Think I've run out of things to try and check, I leave it to the other reviewers now 🫡

esizer:
> I did notice this loading behaviour for some slow loading. Is this new with moving enums to the backend or is it something I hadn't noticed before when it was coming from the frontend?

Yeah, I was trying to avoid a bunch of prop drilling here and load the needed strings only when they were needed (opening the form). This can defintley be improved but I'm not sure how :thinking: 

1. Query on page load and prop drill or,
2. Show another spinner :cry:

api/lang/en/generic_job_title_key.php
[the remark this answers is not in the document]
tristan-orourke: These appear on the job poster, though it looks like we don't use this localizedConstants text directly.

image

See `apps/web/src/pages/Pools/PoolAdvertisementPage/components/ClassificationDefinition.tsx`

api/app/Enums/EducationRequirementOption.php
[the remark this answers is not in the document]
mnigh: Feel free to add any other suggestions to this issue or expand its scope:
- https://github.com/GCTC-NTGC/gc-digital-talent/issues/10892

api/lang/en/generic_job_title_key.php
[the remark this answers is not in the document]
esizer: Removed in [remove generic job title key localization](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/c537a19839f1395b955d0df990299a2cba30ed77)

api/lang/en/placement_type.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+return [
+    'place_tentative' => 'Tentatively placed (Verbal or email confirmation)',
+    'placed_casual' => 'Placed casual',
+    'placed_term' => 'Placed term',
+    'placed_indeterminate' => 'Placed indeterminate',
```
yonikid15, line 7: Need to fix a typo for `place_tentative` to `placed_tentative`

yonikid15 (changes requested):

api/lang/en/placement_type.php
yonikid15: Same with the placement.fr file!

esizer: Fixed in [fix typo](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/75256dc3df5c2b5dd95ae16fe44dba5c41f09b30)

api/lang/en/generic_job_title_key.php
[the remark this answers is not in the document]
esizer: Removed in [remove generic job title key localization](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/c537a19839f1395b955d0df990299a2cba30ed77)

apps/web/src/pages/SearchRequests/ViewSearchRequestPage/components/UpdateSearchRequest.tsx
```diff
@@ -28,6 +27,18 @@ import useRoutes from "~/hooks/useRoutes";
...
+    statuses: localizedEnumStrings(enumName: "PoolCandidateSearchStatus") {
+      value
+      label {
+        en
+        fr
+      }
+    }
+  }
+`);
+
```
yonikid15, line 41: I found a bug with this form. When the user refreshes the request page the `statuses` are fetched in this component, therefore it doesn't show the correct value in the select box. Since the options query hasn't completed yet. You'll need to turn this into a fragment and move it into `pages/SearchRequests/ViewSearchRequestPage.tsx`

esizer: Should be fixed in [fix statuses default value issue](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/f023ff33f5dafa2faeea7ce866bdc2538c746e9e)

yonikid15: Looks good :+1:

yonikid15 (approved):
Amazing work Eric :1st_place_medal:

mnigh (approved):
a lot of 🏋️ going on here, nice work. the issues i've raised have been addressed. [this](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701#issuecomment-2209424125) is not part of the acceptance criteria so i guess i approve 🤷.

api/lang/en/assessment_result_justification.php
petertgiles: Yeah, it's weird.  I suppose you never need to justifiy accepting a skill.  I recommend removing this enum value.
image

petertgiles: https://github.com/GCTC-NTGC/gc-digital-talent/issues/10898

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
petertgiles, line 1: Sorry, I don't this update is quite working.  I found "perfect priority" who was uverified priority and veteran.  They showed as "priority" in the list.  When I rejected the "priority" status they showed up as "Citizen or resident" and got unsorted.  I expected them to show as "Veteran" since they still had unverified veteran status.

petertgiles (changes requested):
Almost there! :sweat_smile:  A few small things and one big thing left.

packages/i18n/src/lang/fr.json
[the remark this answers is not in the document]
esizer: Yes, I think this is an excellent idea.

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
esizer: Yeah, I found that odd but it is how the code is currently working :face_with_spiral_eyes: 

If you claim and are rejected, you get set to `30` which is the citizen weight :thinking: 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/7a7e5e4697cc1ab5122d0c481408885502a438e2/apps/web/src/utils/poolCandidate.ts#L547-L569

petertgiles: Oh, it's probably a pre-existing bug then.  I'll open an issue.

petertgiles: On the bright side, this will be a lot easier to test when it's all on the backend.  :laughing:

packages/i18n/src/lang/fr.json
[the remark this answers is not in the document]
petertgiles: https://github.com/GCTC-NTGC/gc-digital-talent/issues/10900

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
petertgiles: https://github.com/GCTC-NTGC/gc-digital-talent/issues/10901

packages/i18n/src/utils/enum.ts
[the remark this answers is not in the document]
esizer: Of course, added in [document localized enum utility functions](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/e07d6b361ad369c4aedf443343a172b4f2711e4e)

packages/fake-data/src/fakeLocalizedEnum.ts
```diff
@@ -0,0 +1,41 @@
...
+
+function pascalToScreamingSnake<T extends string>(value: T): T {
+  return value.split(pascalSplitRegex).join("_").toUpperCase() as T;
+}
+
+type EnumCase = "pascal" | "screaming_snake";
+
+function toLocalizedEnum<T extends string>(
+  value: T,
+  delimeter?: string | RegExp,
```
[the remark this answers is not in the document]
esizer: Typos, my mortal enemy :sweat: Fixed in [fix typo](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/0c36d616cb54e717935a3ca1d9abc7ee3cc3a506)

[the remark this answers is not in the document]
esizer: I also took the time to document these as well.

[document fake enum data helpers](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10701/commits/0c6f0210c29628d82e59451d7df904da9046e20b)

petertgiles (approved):

PR #10739, opened 2024-06-20
[Debt] Removes Cypress

### LINKED ISSUE #10645: ♻️ Remove Cypress
## ♻️ Debt

Remove Cypress as it is no longer used.

## 🙋‍♀️ Proposed Solution

Remove all references to Cypress.

## ✅ Acceptance Criteria

- [ ] no references to Cypress in codebase

## 🛑 Blockers

```[tasklist]
### Blocked By
- [ ] #10499
```

mnigh:
🤖 Resolves #10645.

## 👋 Introduction

This PR removes Cypress and any references to it. It also updates a few references to using Cypress in the documentation with Playwright instructions.

## 🧪 Testing

1. Verify no references to Cypress in the codebase


infrastructure/azure-pipelines-dev.yml
```diff
@@ -45,9 +45,6 @@ jobs:
       - script: echo ##vso[task.setvariable variable=NPM_CONFIG_USERCONFIG]$(npm_config_user_config)
         displayName: "Set Custom .npmrc Location"
 
-      - script: CYPRESS_INSTALL_BINARY=/opt/cypress/13.10.0/cypress.zip pnpm install cypress
```
mnigh, line 48: @brindasasi I'm not sure if there is anything manual that needs to be changed or removed in Azure?

package.json
esizer, line 1: Should we update the commands to no longer include the `playwirght` specifier? :thinking: ie `e2e:playwright` to just `e2e` and then the projects like `e2e:chromium`

mnigh: I was thinking about that but wouldn't `playwright:chromium` be better?

esizer: Yeah, I think both work :+1: I have also gotten used to using playwright in the commands anyway :laughing:

infrastructure/azure-pipelines-dev.yml
brindasasi: This is the only place needs to be changed. Let me trigger the dev pipeline with this change

brindasasi (approved):
This is great! Verified dev pipeline passed the installation steps with this branch. Thanks a ton! 👏

PR #10741, opened 2024-06-20
[Tooling] Add no-unused-modules rule to eslint

### LINKED ISSUE #7453: 🛠️ Re-enable `no-unused-modules` eslint rule
## 🛠️ Tooling

The eslint rule for `no-unused-modules` has been disabled since it was causing eslint to no longer work in VSCode.

## 🕵️ Details

It seems as though this specific rule does not work with non-standard config names, we need to figure out how to work around this issue.

## 🙋‍♀️ Proposed Solution

The rule is too slow to run effectively in local environments. Run it only on CI.

- ~Wait until the library fixes the issue?~
- ~Define specific config for each package with this rule~

Maybe look into [combining configs](https://eslint.org/docs/latest/use/configure/combine-configs#apply-a-config-object-to-a-subset-of-files)? :thinking: 

## ✅ Acceptance Criteria

- [ ] eslint still functions in CLI and IDE
- [ ] eslint is checking for unused modules
- [ ] performance is not affected
- [ ] no-unused-modules rule runs in CI but not locally

yonikid15:
🤖 Resolves #7453 

## 👋 Introduction

- Add [no-unused-modules](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-unused-modules.md) rule to eslint. Rule runs on CI, but not locally

## 🧪 Testing

1. Run `pnpm run lint`
2. Ensure rule isn't applied locally


apps/web/.eslintrc.cjs
```diff
@@ -8,4 +8,19 @@ module.exports = {
       },
     },
   },
+  rules: {
+    "import/no-unused-modules": (() => {
+      // Trigger warning on CI only
```
petertgiles, line 13: Could you please update the comment that this is only CI because it is slow?

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/ApplicationInformation/types.ts
petertgiles, line 1: I'm not sure I like inlining this file.  This choice was made for organizational reasons, right?

packages/ui/src/components/Accordion/types.ts
petertgiles, line 1: Same comment about inlining.

petertgiles (approved):
Wow, a lot of dead code hacked out.  :clap: 
I left a few style comments but nothing blocking.

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/ApplicationInformation/types.ts
yonikid15: Ya I can move it back though. It was only being used on one page so I moved it there directly. Would you prefer recreating the types.ts and placing it there?

apps/web/.eslintrc.cjs
yonikid15: Yup! Made changes in [cbc6fe7](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10741/commits/cbc6fe71a631e39d92193c161a470e8953a43792)

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/ApplicationInformation/types.ts
petertgiles: Yeah, I like the separate types.ts file, even if it's only used for one page.

yonikid15: Oh I should also mention it only contained a single export for this and the comment below (Accordion). If this looks good I can push up the commit.

What is would look like with new change:


```ts
// ../ApplicationInformation/types.ts

// eslint-disable-next-line import/prefer-default-export
export const SECTION_KEY = {
  CONTACT: "contact",
  SCREENING: "screening",
  GENERAL: "generalQuestions",
  EDUCATION: "education",
  ESSENTIAL: "essentialSkills",
  ASSET: "assetSkills",
  LANGUAGE: "language",
  WORK_PREF: "workPref",
  GOV_INFO: "govInfo",
  DEI: "dei",
  SIGNATURE: "signature",
} as const;
```



```ts
// ../Accordion/types.ts
export type AccordionMode = "card" | "simple";
```

yonikid15: Fixed in [dff806e](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10741/commits/dff806ec488e15438dd87c816da59cd8ba289a4c)

packages/ui/src/components/Accordion/types.ts
yonikid15: Fixed in [dff806e](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10741/commits/dff806ec488e15438dd87c816da59cd8ba289a4c)

PR #10755, opened 2024-06-21
[Fix] Improve current location formatting

### LINKED ISSUE #10744: 🐛 When no current location is found for a user, it renders a comma
## 🐛 Bug

Old profiles including my own shows up current location as  empty with a single comma.
City and province are mandatory fields in profile currently. Probably it wasn't before 

## 🦋 Expected Behaviour

If there is no current location don't display just a comma. Leave it blank. 

## 📋 Steps to Reproduce

1. Manipulate any profile to wipe out city and province data. 
2. Check out the user profile 

## 📸 Screenshot

Screenshot_20-6-2024_215648_talent canada ca

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] If the current location is null , leave it blank. Don't append comma.

mnigh:
🤖 Resolves #10744.

## 👋 Introduction

This PR improves the formatting for when `user.currentCity` has no value.

## 🧪 Testing

1. `pnpm dev`
2. Navigate to `/admin/users/:user-id#about`
4. Set the `current_city` in the `users` table for the user to `NULL` (impossible in UI since the field is required)
5. Verify value after **Current Location:** does not have a comma

## 📸 Screenshot
Screen Shot 2024-06-21 at 12 02 19


petertgiles (approved):
This satisfies the AC so I'm approving.  It might be worth taking this a little bit further and handling the case where province is null.
image

mnigh:
> This satisfies the AC so I'm approving. It might be worth taking this a little bit further and handling the case where province is null.

@petertgiles You are right, I didn't take it far enough. A lot of the `UserInformationPage` components could use a refactor (replace a bunch of "key" strings where colons are included with existing "key" strings; improve conditional rendered; and more) but no reason not to improve the conditional checks for this.

## Screenshots

## Both null
Screen Shot 2024-06-24 at 10 17 25

## City
Screen Shot 2024-06-24 at 10 17 45

## Province
Screen Shot 2024-06-24 at 10 25 00

## City and Province
Screen Shot 2024-06-24 at 10 18 34

petertgiles (approved):
Nice, looks great!  :medal_sports:

PR #10765, opened 2024-06-21
[Fix] Refactors admin view user profile

### LINKED ISSUE #10745: 🐛 When viewing an user's profile as an admin , the null message for Work preferences isn't make sense
## 🐛 Bug

When viewing an user's profile as an admin , the null message for Work preferences inviting the admin to edit. Though the link isn't working , it shouldn't be like that.  
(I believe extra help messages makes maintenance of these pages little hard overall. We probably limit ourselves to generic messages like "No Information provided" . This will go with both user and admin. I'm suggesting it for all fields) 
 
## 🦋 Expected Behaviour
When the profile is viewed , the null message should be targeted for that audience. 
User viewing his own profile vs admin viewing user profile. 

## 📋 Steps to Reproduce

1. Make Work preferences empty in your user profile 
2. .View it as admin 

## 📸 Screenshot

Screenshot_20-6-2024_215427_talent canada ca

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] Keep only the last line "No information has been provided"
- [ ] Ensure null message for all other fields  on this page make sense for admin and user view

mnigh:
🤖 Resolves #10745.

## 👋 Introduction

This PR updates the null state messages for each admin view user profile section to remove messages related to editing the values. It also  removes messages and code that is no longer needed since its components are not used anywhere else.

## 🧪 Testing

1. `pnpm build`
2. Navigate to `/admin/users/:user-id/profile`
3. Verify sections still appear as before
4. Update the row in the users table for `:user-id` to have all DEFAULT values
5. Remove any skills, experiences related to that user
6. Navigate to `/admin/users/:user-id/profile`
7. Verify sections still appear as before but with more concise and accurate NULL state messages


apps/web/src/components/UserProfile/ProfileSections/AboutSection.tsx
```diff
@@ -1,235 +1,78 @@
...
-    currentCity,
-    currentProvince,
-    citizenship,
-    armedForcesStatus,
-  } = user;
+
   return (
-    <Well>
-      <div data-h2-flex-grid="base(flex-start, x2, x1)">
+    <div data-h2-flex-item="base(1of1) p-tablet(3of4)">
```
esizer, line 26: We can remove these wrapping `div`'s around each `Well`

apps/web/src/components/UserProfile/UserProfile.tsx
```diff
@@ -130,427 +43,227 @@ const UserProfile = ({
...
-                <TableOfContents.Heading
-                  as={headingLevel}
-                  icon={UserIcon}
-                  data-h2-margin-top="base(0)"
+            <div data-h2-padding="base(x2, 0, x1, 0)">
+              <div data-h2-flex-grid="base(center, x2, x1)">
+                <div
+                  data-h2-flex-item="base(1of1) p-tablet(fill)"
+                  data-h2-text-align="base(center) p-tablet(left)"
                 >
```
esizer, line 109: We can remove these extra `div`'s around each heading now

apps/web/src/components/UserProfile/UserProfile.tsx
```diff
@@ -130,427 +43,227 @@ const UserProfile = ({
...
-                  {intl.formatMessage(navigationMessages.aboutMe)}
-                </TableOfContents.Heading>
+                  <TableOfContents.Heading
+                    as={headingLevel}
+                    icon={UserIcon}
+                    data-h2-margin-top="base(0)"
```
esizer, line 113: Leave this but also add `data-h2-margin-bottom="base(x1)"` to all `TableOfContents.Heading`

esizer (changes requested):
I think we can :axe: even more.

In addition to comments, we should also be able to remove the following `div`:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/86cef21c7eaaf9390a747e9cb4375b12358be4db/apps/web/src/pages/Users/AdminUserProfilePage/AdminUserProfilePage.tsx#L255

apps/web/src/components/UserProfile/UserProfile.tsx
```diff
@@ -130,427 +43,168 @@ const UserProfile = ({
...
-            {sections.about?.override ? (
-              sections.about.override
-            ) : (
-              <AboutSection user={user} editPath={sections.about?.editUrl} />
-            )}
+            <TableOfContents.Heading
+              as={headingLevel}
+              icon={UserIcon}
+              data-h2-margin-top="base(0)"
+              data-h2-margin-bottom="base(x1)"
```
esizer, line 108: Sorry, this is my lack of understand of how hydrogen works but, this should be:

`data-h2-margin="base(0 0 x1 0)"` and all the others should be `data-h2-margin="base(x1.5 0 x1 0)"`

apps/web/src/components/UserProfile/UserProfile.tsx
```diff
@@ -130,427 +43,168 @@ const UserProfile = ({
...
+              </TableOfContents.AnchorLink>
+            </TableOfContents.ListItem>
+            <TableOfContents.ListItem>
+              <TableOfContents.AnchorLink id={PAGE_SECTION_ID.SKILL_SHOWCASE}>
+                {intl.formatMessage(navigationMessages.skillShowcase)}
+              </TableOfContents.AnchorLink>
             </TableOfContents.ListItem>
-          )} */}
           </TableOfContents.List>
         </TableOfContents.Navigation>
```
esizer, line 101: This is a larger change but, I am wondering if we want to move the print button here so we don't have all that extra  whitespace to accommodate it? :thinking: 

Not necessary so feel free to ignore.

screenshot_24-Jun-2024_14-18-1719253125

esizer (changes requested):

apps/web/src/components/UserProfile/UserProfile.tsx
mnigh: [b2b17d3](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10765/commits/b2b17d38c849ade58dda12db2e0512c22522168a)

esizer (approved):
LGTM! :+1:

PR #10789, opened 2024-06-25
[Fix] Updates table `nullMessage` descriptions

### LINKED ISSUE #9463: 🐛 Null state message on pool candidates table doesn't make sense

mnigh:
🤖 Resolves #9463.

## 👋 Introduction

This PR removes the fallback for the `nullMessage.description` prop for the `Table` component and adds custom messages for where there is an `add` prop.

## 🧪 Testing

1. Navigate to tables
2. Verify any table that has an add or create button has an instructional description when the table yields no results.

### `add` prop exists
Screen Shot 2024-06-25 at 15 38 47

### `add` prop does not exist
Screen Shot 2024-06-25 at 15 43 33

### DIFF
--- a/apps/web/src/components/Table/ResponsiveTable/NullMessage.tsx
+++ b/apps/web/src/components/Table/ResponsiveTable/NullMessage.tsx
@@ -3,6 +3,8 @@ import { ReactNode } from "react";
 
 import { Heading, Well } from "@gc-digital-talent/ui";
 
+import tableMessages from "../tableMessages";
+
 export interface NullMessageProps {
   /** Heading for the message */
   title?: ReactNode;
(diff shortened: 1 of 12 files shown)


petertgiles (approved):
Nice, very clean.  :broom:

Issue #10820, opened 2024-06-27
🐛 Notification API channel throttle starves legitimate jobs


petertgiles:
## 🐛 Bug

The exception throttle on the notification API shares a common "bucket".  This is probably intended so that our application doesn't hammer the Notify API if it runs into problems and start rejecting every request.
Ref: https://laravel.com/docs/10.x/queues#throttling-exceptions

However, I noticed that if you test the "team" API with only one of many addresses being valid the valid job will get starved out when all the jobs ahead of it throw exceptions and hit the exception throttle limit.  Eventually the 30 minute limit will arrive for the valid job before it has even been really attempted once and the job will fail out.

I'm worried this will happen in prod as well, if many users have entered invalid or junk email addresses in their profiles.

## 📋 Steps to Reproduce

1. Seed the database fresh.  You should have around 33 users.
2. Set the "team" api key and the system notification template IDs in your api/.env file.
3. Update one user to have a legitimate team email address.
4. `./artisan send-notifications:system notification_announcement `
5. Observe your 33 jobs in the database, of which only one has a valid team email address, the last one. (see query below)
6. Start the queue worker. `docker-compose exec webserver sh -c "runuser -u www-data -- php /home/site/wwwroot/api/artisan queue:work"`
7. Observe that 10 jobs run and fail and that your legitimate job never runs
8. Eventually, after 30 minutes, all jobs are kicked to the failed_jobs table and the legitimate email was never sent.

```sql

select
  job_id,
  substring(start_of_address, 0, strpos(start_of_address, '"')) address, 
  attempts, available_at, created_at
from (
	select
	    id job_id,
	    substring((payload::json->'data')->>'command', 199) start_of_address,
	    attempts, available_at, created_at
	from jobs j
) t

```

## 🙋‍♀️ Proposed Solution

I think we should not have the exception throttle limit shared between all notifications.  This means we may end up being "unkind" to the API if it has trouble but at least our jobs will not get starved out.

After running the first batch of notifications we will know what the common responses look like and we can stop throwing those as exceptions.  Maybe we can turn the shared throttle back on at that point.

## ✅ Acceptance Criteria

- [ ] Legitimate jobs still get run, even if they are behind 10 bad jobs

PR #10821, opened 2024-06-27
[Fix] Tune notification throttle

petertgiles:
🤖 Resolves : #10820 

## 👋 Introduction

This branch sets the notification throttle to bucket by job.  It also tunes the retry timing.  

## 🕵️ Details

This will behave differently now:
1. If a single job hits the exception throttle, all other jobs will continue in their own timing without changes.
2. The tuned exception throttle works like this: "Wait four minutes between tries.  Try again up to three times then blocked for 10 minutes.  Job fails if not complete in 30 minutes total."

Note: as far as I can tell, if a message is accepted by the API but is not deliverable it does not trigger an API error.  So this may not even be an issue in production.

## 🧪 Testing

1. Set up the "steps to reproduce" from the issue.
2. Observer that your legitimate email is properly sent.

### DIFF
--- a/api/app/Jobs/GcNotifyApiRequest.php
+++ b/api/app/Jobs/GcNotifyApiRequest.php
@@ -30,6 +30,7 @@ class GcNotifyApiRequest implements ShouldQueue
         return [
             (new GcNotifyRateLimited),
             (new ThrottlesExceptions(10, 5))
+                ->byJob()
                 ->backoff(5),
         ];
     }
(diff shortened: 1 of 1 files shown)


brindasasi:
Here is my test plan for throttle

1. Confirm Notification feature is turned on  (in app & email)
2. Post a job and trigger the mass notification 
3. Confirm throttle is happening
4. deploy the fix 
5. Post another job and trigger the notification again
6. confirm the bug is fixed

brindasasi (approved):
While having throttle issue these emails didn't go through.
After deploying the fix emails sent to the team's whitelisted emails! 🥳👏 
image (2)

brindasasi:
Verified failed_jobs table for the captured failed emails. Not all of them are in there. But I think its because its still retrying some of the jobs (3 times retry). I will check it out tmo again on that.

PR #10843, opened 2024-06-28
[Bug] Notification overlay fix

### LINKED ISSUE #10826: 🐛 Address underlying issue with framer-motion
## 🐛 Bug

We recently had a framer-motion dependency bump that broke our notification pane.  We rolled back the dependency but we need to solve the underlying issue.

## 🕵️ Details

Issue: https://github.com/GCTC-NTGC/gc-digital-talent/issues/10796
Rollback: https://github.com/GCTC-NTGC/gc-digital-talent/pull/10825

h/t @vd1992 for getting started on this already: https://github.com/GCTC-NTGC/gc-digital-talent/tree/experiment-notification-overlay-fix

## ✅ Acceptance Criteria

- [ ] Update framer-motion dependency beyond 11.2.10 
- [ ] Regression test to prevent the problem with the notification pane from occurring.
- [ ] Grey overlay appears and disappears when it's supposed to

vd1992:
🤖 Resolves #10826

## 👋 Introduction

Bump Framer up without breaking the notification overlay

## 🕵️ Details

Had a branch going for the issue, might as well let it be seen.

I think the issue was the interactions of a dialog nested in an animation. There are multiple parts involved
- The dialog
- The animation
- The elements with rendering to be removed after animation completion 

The elements would animate in, then out, but not actually disappear. Their removal after animation was not happening, the pane was still there but off-screen while the overlay still being present on the other hand was obvious. 

Refactored to change how parts should disappear by using `usePresence()` and preserved the animation in and out.
Used the existing Playwright spec to test. 

## 🧪 Testing

1. Framer updated
2. Pane slides in and out, overlay appears and disappears

## 📸 Screenshot

Open

image

Then closed

image


apps/web/src/components/NotificationDialog/NotificationDialog.tsx
petertgiles, line 1: I'm getting some errors from Radix in the console.   They seem to be related to a11y concerns.
image

petertgiles (commented):
I'll be honest, I don't understand the solution at all, though it's probably good.  :laughing: I can confirm that the AC is met and the test makes sense.

apps/web/src/components/NotificationDialog/NotificationDialog.tsx
vd1992: Not new to this PR, is pre-existing
Probably merits its own issue and discussion about whether something is actually wrong

esizer: I actually noticed this as well. It is a simple fix. I'll create a ticket + PR

vd1992: Thanks!

apps/web/src/components/NotificationDialog/NotificationDialog.tsx
```diff
@@ -44,9 +45,125 @@ const ellipsis = () => (
...
+}) => {
   const intl = useIntl();
   const paths = useRoutes();
+  const [isPresent] = usePresence();
+  const [render, setRender] = useState<boolean>(isPresent);
+  useEffect(() => {
+    if (isPresent) {
+      setRender(true);
+    } else {
+      setTimeout(() => setRender(false), 200); // let animation complete before removal
```
petertgiles, line 61: I noticed this because it came up in my PR, too.
https://github.com/GCTC-NTGC/gc-digital-talent/pull/10824#discussion_r1662609964

You should probably get the return from setTimeout and clean it up.

petertgiles (changes requested):
OK, I think I sort of understand now.  :laughing:  Just one thing left for me.

apps/web/src/components/NotificationDialog/NotificationDialog.tsx
vd1992: Like so https://github.com/GCTC-NTGC/gc-digital-talent/pull/10843/commits/8638d9427131be3ea2a7e316a67deb710426db79 ?

petertgiles: Nice.

petertgiles (approved):

PR #10899, opened 2024-07-05
[Perf] Lazy load site banner

### LINKED ISSUE #10897: ♻️ Lazy load `SitewideBanner`
## ♻️ Debt/Refactor

We should probably lazy load the component only if the query returns something we can render.

## 🕵️ Details

Currently, we always load the `SitewideBanner` component. The issue is, it contains a very heavy library (tiptap ~200kb) that we load, regardless if we ever intend to use it. We should lazy load this component only if it contains data we can render.

## 🙋‍♀️ Proposed Solution

I'm not 100% confident but this this could work? :thinking: 

```tsx
// BannerContent.tsx
const BannerContent = ({ data }) => ();

// SitewideBanner.tsx
const Content = React.lazy(() => import("./BannerContent.tsx");

const SitewideBanner = () => {
  const [{data, fetching}] = useQuery({ .. });
  
  if(!data || fetching) return null;
  
  return (
     <Suspense fallback={<Loading />}
        <Content />
     </Suspense>
  );
}
```

## ✅ Acceptance Criteria

- [ ] Tiptap lazy loaded only when needed
- [ ] Sitewide banner can still render rich text

esizer:
🤖 Resolves #10897 

## 👋 Introduction

Updates the site banner to be dynamically imported so it does not need to load when there is no message to be displayed.

## 🕵️ Details

Since the banner contains rich text, we need to load tiptap to render the data. Tiptap is one of our larger libraries (307kb) so we should only load it when we need it.

## 🧪 Testing

1. Build `pnpm run dev:fresh`
2. Navigate to the homepage
3. Confirm tip tap is not loaded
4. Enable the banner
5. Navigate home
6. Confirm tiptap is loaded and banner renders

### Not loaded

screenshot_05-Jul-2024_12-08-1720195696

### Loaded

screenshot_05-Jul-2024_12-11-1720195864

### DIFF
--- a/apps/web/src/components/Layout/BannerContent.tsx
+++ b/apps/web/src/components/Layout/BannerContent.tsx
@@ -1,0 +1,28 @@
+import { Alert } from "@gc-digital-talent/ui";
+import { RichTextRenderer, htmlToRichTextJSON } from "@gc-digital-talent/forms";
+
+interface BannerContentProps {
+  title: string;
+  message: string;
+}
+
+const BannerContent = ({ title, message }: BannerContentProps) => (
+  <div
+    data-h2-background-color="base(foreground) base:dark(white)"
+    data-h2-padding="base(x1, 0)"
+  >
+    <div data-h2-container="base(center, large, x1)">
+      <Alert.Root
+        type="warning"
+        live
+        data-h2-shadow="base(none)"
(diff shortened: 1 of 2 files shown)


petertgiles (approved):
Cool, clean, clever!  :1st_place_medal:

PR #10920, opened 2024-07-09
[Debt] Compute final decision on server

### LINKED ISSUE #10846: ✨ Compute Final Decision on backend and make it sortable
## ✨ Feature

The "Final decision" field, as we're currently calling it, is a property of PoolCandidates which combines the `assessmentStatus` added in #10642 with PoolCandidateStatus. We currently compute this on the frontend for display purposes, but we want to move it to the backend so that candidates can be sorted by this value before pagination.

## 🕵️ Details

The logic for "final decision" goes as follows. (This is based on `getCandidateStatusChip` in `apps/web/src/uitls/poolCandidate.ts`)
```


const TO_ASSESS_STATUSES = [
  PoolCandidateStatus.NewApplication,
  PoolCandidateStatus.ApplicationReview,
  PoolCandidateStatus.ScreenedIn,
  PoolCandidateStatus.UnderAssessment,
];

IF candidate.status is in TO_ASSESS_STATUSES:
  compute FinalDecision based based on candidate.assessmentStatus:
    IF candidate.assessmentStatus == null: // This should only be the case for pre-RoD candidates
      FinalDecision  is "To Assess" // "To Assess: Step 0" or "To Assess: Step 1" would also be fine
    IF candidate.assessmentStatus.overallAssessmentStatus == DISQUALIFIED:
      FinalDecision  is "Disqualified: Pending decision"
    IF candidate.assessmentStatus.overallAssessmentStatus == QUALIFIED:
      FinalDecision is "Qualified: Pending decision"
    IF candidate.assessmentStatus.overallAssessmentStatus == TO_ASSESS:
      FinalDecision is `To Assess: Step {candidate.assessmentStatus.currentStep}`

ELSE IF candidate.status is NOT in TO_ASSESS_STATUSES:
  FinalDecision is mapped from candidate.status:
    [ScreenedOutApplication, ScreenedOutAssessment] => "Disqualified"
    [QualifiedAvailable] => "Qualified"
    [PlacedCasual, PlacedTerm, PlacedIndeterminate, PlacedTentative] => "Qualified: Placed"
    [ScreenedOutNotInterested, ScreenedOutNotResponsive] => "Removed: To Assess"
    [QualifiedUnavailable, QualifiedWithdrew] => "Removed: Qualified"
    [Removed] => "Removed"
    [Expired] => "Expired: Qualified"
```

Another key feature is that when fetching paginated candidates, they can be ordered by FinalDecision. The default ordering should be like this:
```
Qualified
Qualified: Pending decision
To assess step 1
To assess step 2
....
Qualified: Placed
Disqualified: Pending decision
Disqualified
Removed: Qualified
Removed: To Assess
Removed
Expired: Qualified
```

## 🎨 Design File

### 🧑‍🎨 Designer

@Jerryescandon 

## 📸 Screenshot

image

I'm not sure. It could be a database column which sql is set up to automatically recompute (like priorityWeight), it could be generated on the fly by the sql query (like our skills count - and see also [this comment](https://github.com/GCTC-NTGC/gc-digital-talent/issues/8966#issuecomment-2158872097)), or it could be computed by laravel and stored in and event listener, like with #10642.

If any database columns are added which are computed automatically, prefix the name with `computed_`.

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] The PoolCandidates table can order candidate candidates by Final Decision, and they are sorted according to what's specified in Details.
- [ ] The final decision status is available from the graphql on the pool candidate object
- [ ] PHPunit tests
- [ ] Any frontend changes are out of scope

esizer:
🤖 Resolves #10846 

## 👋 Introduction

Adds a candidates final decision, with weight, to the database.

## :clipboard: TO DO

- [x] Create artisan command to seed data into existing applicants

## 🕵️ Details

There is an event we can dispatch that will update the candidate when items that impact this are changed. However, you can also just call the `computeFinalDecision` method on the `PoolCandidate` model. 

I didn't want to need to save twice when we know it is changing so in the mutations where we update the status, it simply computes the values and stores them just before saving.

> [!NOTE]
> I think I got all the places where we need to dispatch the event or compute manually. If reviewing, it would be helpful if people could look for and point out any other places it should happen :pray: 

## 🧪 Testing

1. Refresh DB `make seed-fresh`
2. Check in the DB to confirm that created candidates have the correct final decision and weight
3. Update some candidate statuses and submit some assessments
4. Confirm the final decision updates appropriately
5. Remove some weights and decisions from the database
6. Run sync command `make artisan CMD="app:sync-final-decision"`
7. Confirm it updates the candidates appropriatley

## :truck: Deployment

After deployment run the artisan command to sync new final decision columns

```sh
php artisan app:sync-final-decision
```

### DIFF
--- a/api/app/GraphQL/Mutations/DisqualifyCandidate.php
+++ b/api/app/GraphQL/Mutations/DisqualifyCandidate.php
@@ -21,10 +21,6 @@ final class DisqualifyCandidate
         $candidate->pool_candidate_status = $reason;
         $candidate->final_decision_at = $now;
 
-        $finalDecision = $candidate->computeFinalDecision();
-        $candidate->computed_final_decision = $finalDecision['decision'];
-        $candidate->computed_final_decision_weight = $finalDecision['weight'];
-
         $candidate->save();
 
         return $candidate;
--- a/api/app/GraphQL/Mutations/QualifyCandidate.php
+++ b/api/app/GraphQL/Mutations/QualifyCandidate.php
@@ -23,10 +23,6 @@ final class QualifyCandidate
         $candidate->expiry_date = $expiryDate;
         $candidate->final_decision_at = $now;
 
-        $finalDecision = $candidate->computeFinalDecision();
-        $candidate->computed_final_decision = $finalDecision['decision'];
-        $candidate->computed_final_decision_weight = $finalDecision['weight'];
-
         $candidate->save();
 
         return $candidate;
--- a/api/app/GraphQL/Mutations/ReinstateCandidate.php
+++ b/api/app/GraphQL/Mutations/ReinstateCandidate.php
@@ -40,10 +40,6 @@ final class ReinstateCandidate
         $candidate->removal_reason = null;
         $candidate->removal_reason_other = null;
 
-        $finalDecision = $candidate->computeFinalDecision();
-        $candidate->computed_final_decision = $finalDecision['decision'];
-        $candidate->computed_final_decision_weight = $finalDecision['weight'];
-
         $candidate->save();
 
         return $candidate;
--- a/api/app/GraphQL/Mutations/RemoveCandidate.php
+++ b/api/app/GraphQL/Mutations/RemoveCandidate.php
@@ -60,10 +60,6 @@ final class RemoveCandidate
                 break;
         }
 
-        $finalDecision = $candidate->computeFinalDecision();
-        $candidate->computed_final_decision = $finalDecision['decision'];
-        $candidate->computed_final_decision_weight = $finalDecision['weight'];
-
         $candidate->save();
 
         return $candidate;
--- a/api/app/GraphQL/Mutations/RevertFinalDecision.php
+++ b/api/app/GraphQL/Mutations/RevertFinalDecision.php
@@ -20,10 +20,6 @@ final class RevertFinalDecision
         $candidate->expiry_date = null;
         $candidate->final_decision_at = null;
 
-        $finalDecision = $candidate->computeFinalDecision();
-        $candidate->computed_final_decision = $finalDecision['decision'];
-        $candidate->computed_final_decision_weight = $finalDecision['weight'];
-
         $candidate->save();
 
         return $candidate;
--- a/api/app/GraphQL/Mutations/RevertPlaceCandidate.php
+++ b/api/app/GraphQL/Mutations/RevertPlaceCandidate.php
@@ -20,10 +20,6 @@ final class RevertPlaceCandidate
         $candidate->placed_at = null;
         $candidate->placed_department_id = null;
 
-        $finalDecision = $candidate->computeFinalDecision();
-        $candidate->computed_final_decision = $finalDecision['decision'];
-        $candidate->computed_final_decision_weight = $finalDecision['weight'];
-
         $candidate->save();
 
         return $candidate;
--- a/api/app/GraphQL/Mutations/SubmitApplication.php
+++ b/api/app/GraphQL/Mutations/SubmitApplication.php
@@ -47,10 +47,6 @@ final class SubmitApplication
             $application->priority_verification = ClaimVerificationResult::UNVERIFIED->name;
         }
 
-        $finalDecision = $application->computeFinalDecision();
-        $application->computed_final_decision = $finalDecision['decision'];
-        $application->computed_final_decision_weight = $finalDecision['weight'];
-
         // need to save application before setting application snapshot since fields have yet to be saved to the database.
         $application->save();
 
--- a/api/app/Listeners/ComputeCandidateFinalDecision.php
+++ b/api/app/Listeners/ComputeCandidateFinalDecision.php
@@ -26,6 +26,6 @@ class ComputeCandidateFinalDecision
         $candidate->computed_final_decision_weight = $finalDecision['weight'];
         $candidate->computed_final_decision = $finalDecision['decision'];
 
-        $candidate->save();
+        $candidate->saveQuietly();
     }
 }
--- a/api/app/Observers/PoolCandidateObserver.php
+++ b/api/app/Observers/PoolCandidateObserver.php
@@ -27,6 +27,8 @@ class PoolCandidateObserver
         $oldStatus = $poolCandidate->getOriginal('pool_candidate_status');
         $newStatus = $poolCandidate->pool_candidate_status;
 
+        CandidateStatusChanged::dispatchIf($poolCandidate->wasChanged('pool_candidate_status'), $poolCandidate);
+
         if (config('feature.notifications')) {
             if (
                 ($oldStatus != $newStatus) &&


api/tests/Feature/CandidateFinalDecisionTest.php
esizer, line 1: This actually seems more like a unit test :sweat_smile:

yonikid15 (changes requested):
I'd like to go over this with you if you have time. I believe it's working, but just want to double check some things.

api/app/Models/PoolCandidate.php
```diff
@@ -1229,4 +1233,92 @@ public function scopeProcessNumber(Builder $query, ?string $processNumber): Buil
...
+                // Giving a decent buffer to increase max steps
+                FinalDecision::QUALIFIED_PLACED->name => 200,
+                FinalDecision::DISQUALIFIED_PENDING->name => 210,
+                FinalDecision::DISQUALIFIED->name => 220,
+                FinalDecision::QUALIFIED_REMOVED->name => 230,
+                FinalDecision::TO_ASSESS_REMOVED->name => 240,
+                FinalDecision::REMOVED->name => 250,
+                FinalDecision::QUALIFIED_EXPIRED->name => 260,
+                default => null
+            };
```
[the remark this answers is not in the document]
esizer: Unfortunately not but it should follow what was [described in the ticket](https://github.com/GCTC-NTGC/gc-digital-talent/issues/10846#:~:text=Another%20key%20feature%20is%20that%20when%20fetching%20paginated%20candidates%2C%20they%20can%20be%20ordered%20by%20FinalDecision.%20The%20default%20ordering%20should%20be%20like%20this%3A) :+1:

api/app/GraphQL/Mutations/DisqualifyCandidate.php
```diff
@@ -22,6 +22,10 @@ public function __invoke($_, array $args)
         $candidate->final_decision_at = $now;
         $candidate->save();
 
+        $finalDecicion = $candidate->computeFinalDecision();
+        $candidate->computed_final_decision = $finalDecicion['decision'];
+        $candidate->computed_final_decision_weight = $finalDecicion['weight'];
```
yonikid15, line 27: Oops there's also a typo here. Is should be `$finalDecision`.

You'll need to fix that in the other files as well:

- [ ] PlaceCandidate.php
- [ ] QualifyCandidate.php
- [ ] RenistateCandidate.php
- [ ] RemoveCandidate.php
- [ ] RevertFinalDecision.php
- [ ] RevertPlaceCandidate.php
- [ ] SubmitApplication.php

api/app/Models/PoolCandidate.php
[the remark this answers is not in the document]
esizer: @tristan-orourke can you confirm this?

api/app/GraphQL/Mutations/QualifyCandidate.php
```diff
@@ -24,6 +24,10 @@ public function __invoke($_, array $args)
         $candidate->final_decision_at = $now;
         $candidate->save();
 
+        $finalDecicion = $candidate->computeFinalDecision();
+        $candidate->computed_final_decision = $finalDecicion['decision'];
+        $candidate->computed_final_decision_weight = $finalDecicion['weight'];
+
         return $candidate;
```
yonikid15: Need to move `candidate->save();` after the new changes to candidate.

Other files this needs to fixed in includes:

- [ ] DisqualifyCandidate.php

yonikid15 (changes requested):

api/app/GraphQL/Mutations/DisqualifyCandidate.php
esizer: Fixed in [fix typo](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10920/commits/2dea74fbe471ea8538bc6c5152faa655d710c48f)

yonikid15:
@tristan-orourke I we found everything, the code looks good to me. If you want to take a look as well that be great, specifically the logic that Eric followed and confirm it's correct.

```


const TO_ASSESS_STATUSES = [
  PoolCandidateStatus.NewApplication,
  PoolCandidateStatus.ApplicationReview,
  PoolCandidateStatus.ScreenedIn,
  PoolCandidateStatus.UnderAssessment,
];

IF candidate.status is in TO_ASSESS_STATUSES:
  compute FinalDecision based based on candidate.assessmentStatus:
    IF candidate.assessmentStatus == null: // This should only be the case for pre-RoD candidates
      FinalDecision  is "To Assess" // "To Assess: Step 0" or "To Assess: Step 1" would also be fine
    IF candidate.assessmentStatus.overallAssessmentStatus == DISQUALIFIED:
      FinalDecision  is "Disqualified: Pending decision"
    IF candidate.assessmentStatus.overallAssessmentStatus == QUALIFIED:
      FinalDecision is "Qualified: Pending decision"
    IF candidate.assessmentStatus.overallAssessmentStatus == TO_ASSESS:
      FinalDecision is `To Assess: Step {candidate.assessmentStatus.currentStep}`

ELSE IF candidate.status is NOT in TO_ASSESS_STATUSES:
  FinalDecision is mapped from candidate.status:
    [ScreenedOutApplication, ScreenedOutAssessment] => "Disqualified"
    [QualifiedAvailable] => "Qualified"
    [PlacedCasual, PlacedTerm, PlacedIndeterminate, PlacedTentative] => "Qualified: Placed"
    [ScreenedOutNotInterested, ScreenedOutNotResponsive] => "Removed: To Assess"
    [QualifiedUnavailable, QualifiedWithdrew] => "Removed: Qualified"
    [Removed] => "Removed"
    [Expired] => "Expired: Qualified"
```

api/app/GraphQL/Mutations/QualifyCandidate.php
yonikid15: Missed QualifyCandidate :sweat_smile: then should be good :+1:

api/app/Models/PoolCandidate.php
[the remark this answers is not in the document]
tristan-orourke: Based on @Jerryescandon's comment [here](https://github.com/GCTC-NTGC/gc-digital-talent/issues/10846#issuecomment-2211352110), could you just move Qualified Placed above To Assess?
The theory is that this table will be used more for matching, where Assessing Candidates is less relevant, and the "kanban" screen is where assessment happens.

api/app/GraphQL/Mutations/DisqualifyCandidate.php
```diff
@@ -20,6 +20,11 @@ public function __invoke($_, array $args)
 
         $candidate->pool_candidate_status = $reason;
         $candidate->final_decision_at = $now;
+
+        $finalDecision = $candidate->computeFinalDecision();
+        $candidate->computed_final_decision = $finalDecision['decision'];
+        $candidate->computed_final_decision_weight = $finalDecision['weight'];
```
tristan-orourke, line 26: Is running this in the Mutations necessary? Won't this be a duplication of the work triggered by the Candidate Status Changed event?

api/app/Models/PoolCandidate.php
```diff
@@ -1229,4 +1233,92 @@ public function scopeProcessNumber(Builder $query, ?string $processNumber): Buil
...
+            }
+
+            $decision = match ($overallStatus) {
+                OverallAssessmentStatus::QUALIFIED->name => FinalDecision::QUALIFIED_PENDING->name,
+                OverallAssessmentStatus::DISQUALIFIED->name => FinalDecision::DISQUALIFIED_PENDING->name,
+                default => FinalDecision::TO_ASSESS->name
+            };
+        } else {
+
+            $decisionMap = [
```
tristan-orourke, line 1258: Maybe use the `match` statement again, for consistency? But the logic looks right. 👍

tristan-orourke (commented):
The logic for the final decision looks good. Just please make that one change to the weighting 👍

api/app/Models/PoolCandidate.php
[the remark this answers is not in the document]
Jerryescandon: > The theory is that this table will be used more for matching, where Assessing Candidates is less relevant, and the "kanban" screen is where assessment happens.

Correct. On the "Talent placement" table its just important to have a consistent easy to understand organization, there are no actual weights for the purposes of priority. At this point we are not worried with "Who should we look at first?" Since that happens on the kanban. Here we just wanna be able to find the right person using the filters and column sorting. So we just wanna have tidy statuses.

api/app/GraphQL/Mutations/DisqualifyCandidate.php
esizer: No, that event is only fires on `AssessmentStatusChanged` and `PoolCandidate::created`. Since it calls a `save` event I figured we may as well update it here so we don't need to make another trip to the database.

api/app/GraphQL/Mutations/QualifyCandidate.php
esizer: Fixed in [fix save order when qualifying candidate](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10920/commits/070671adf3233accbf65abef486fd1d23a70f240)

api/app/Models/PoolCandidate.php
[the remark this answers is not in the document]
esizer: okay, I think it should be good now. Thanks for confirming :+1: 

[update decision map, weights](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10920/commits/c6609066373c1581e18c83e33bf5cf6ebc9d0225)

yonikid15 (approved):
Looks good on the things I addressed :+1:

api/app/GraphQL/Mutations/DisqualifyCandidate.php
tristan-orourke: Hmm. I'll be honest, I find it very confusing that updating the candidate status does not trigger the CandidateStatusChanged event. 😖 Unless there's some good reason not to, I recommend adding that to the PoolCandidateObserver.

tristan-orourke (changes requested):
The logic and weights look good. My only concern is that I find it confusing to have a CandidateStatusChanged event doesn't always trigger when the candidate status changes.

api/app/GraphQL/Mutations/DisqualifyCandidate.php
esizer: Moved to observer. I also now remeber why I moved it out. It was really bogging down but I realized the `save` was for some reason marking the status as changed and caused infiinte loop :face_with_spiral_eyes: 

Changing to `saveQuielty` seems to have fixed it :woman_shrugging: 

[compute final decision in observer](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10920/commits/ba6baaedb8021be36774dd2560d07d4491afed04)

tristan-orourke (approved):
Thanks @esizer !

PR #10921, opened 2024-07-09
[FIX] Add dev dependencies to Azure build pipeline

### LINKED ISSUE #10886: 🐛 Can no longer deploy artifact with dev dependencies
## 🐛 Bug

Our dev Azure pipeline was recently updated to run Playwright tests https://github.com/GCTC-NTGC/gc-digital-talent/pull/9652 .  However, it can no longer be used to deploy an artifact to an app service with dev dependencies https://github.com/GCTC-NTGC/gc-digital-talent/pull/9301 .  Let's restore that functionality or create a new pipeline just for that.

## ✅ Acceptance Criteria

- [ ] Can deploy an artifact to an app service with dev dependencies like Faker.

petertgiles:
🤖 Resolves #10886 

## 👋 Introduction

This branch adds the ability to include the dev dependencies when deploying to Azure.  Rather than creating a new pipeline, I added a runtime parameter to the existing pipeline.

The one thing I'm not sure about is how the trigger will work.  I hope the default value will be used but we'll find out when it gets merged.  :sweat_smile:  If it doesn't work we can manually trigger the pipeline when needed until the change gets backed out.

## 🧪 Testing

1. Run the Azure Pipeline in dev mode
2. Deploy to an app service
3. SSH in and try a factory in Tinker

## 📸 Screenshot

image

image

### DIFF
--- a/infrastructure/azure-pipelines.yml
+++ b/infrastructure/azure-pipelines.yml
@@ -1,4 +1,12 @@
 name: $(Build.DefinitionName)_$(SourceBranchName)_$(Date:yyyyMMdd)$(Rev:.r)
+parameters:
+  - name: buildMode
+    displayName: Build mode
+    type: string
+    default: default
+    values:
+      - default
+      - dev
 trigger:
   branches:
     include:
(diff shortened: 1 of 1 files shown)


brindasasi (approved):
yayy.. I'm able to run all the seeders in dev. 
Also it was easy to change it to usual prod build mode. 
Great work !👏
Thanks Peter!

PR #10925, opened 2024-07-09
[Fix] Assessment status logic

### LINKED ISSUE #10919: 🐛 Behavioral skills not considered in assessment step status
## 🐛 Bug

Currently, the step status is being reflected immediately after a single essential skill is assessed. We only want this to be reflected once all skills has some sort of assessment decision.

## 🦋 Expected Behaviour

Candidate only passes a step when the following criteria are met for steps other than the first step (application screening):

- All essential (regardless of category) have succeeded
- All asset skills have been assessed, regardless of whether they succeeded or not

## 📋 Steps to Reproduce

2. Navigate to a candidate in the admin
3. Assess all essential technical skills
4. Observer the overall status for the step (in the column header) updates regardless of essential behaviouirl skills have been assessed

## 📸 Screenshot

screenshot_08-Jul-2024_14-36-1720463788

## 🙋‍♀️ Proposed Solution

As well as getting a boolean value for essential skills, we should be getting a count of all skills to be assessed on the step. If the number of skills assessed is less than the total number of skills to be assessed, lock in the decision as "to assess". Only change this value and increment once all skills have been assessed.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/bd12fe758377517c4cdda3c84548f4e0de4cf84b/api/app/Models/PoolCandidate.php#L1042-L1044

## ✅ Acceptance Criteria

- [ ] Assessment step status only changes once all skills have been assessed
- [ ] All essential skills must succeed or be on hold before a step can be marked as successful

esizer:
🤖 Resolves #10919 

## 👋 Introduction

Fixes a few gaps in the logic when computing the assessment status.

## 🕵️ Details

The main issue was, we were computing it based on available results. So, as soon as a single success occurred, the user passed the step. We have rewrote the logic to do it based off of skills needing to be assessed. There is some frontend stuff that is a little odd but this is how it should work now:

1. On every step, all skills (except for unclaimed optional technical skills) must be assessed before the step is considered complete and the user can move on to the next step
2. On the first step (application screening), all technical skills and education requirement must pass
3. Every other step,
    1. If a skill is essential and failed, step automatically fails
    2. If the skill is unsure or has not been assessed, the step is marked as "to assess"
    3. If all essential skills succeed, step is marked with success
    4. All other skills only count towards the step completion but not the actual status of the step

## 🧪 Testing

1. Login as admin
2. Create a pool that has a few assessment steps making sure to mix up the types and categories of each skill type
3. Setup your user to have the following skills from the pool:
    - All essential technical skills
    - Missing at least one nonessential technical skill
    - Missing some essential and nonessential behavioural skills
4. Apply to the pool
5. Navigate tot he admin view of the application from step 4
6. Begin assessing the candidates skill, changing up the assessments to check the logic outlined
    - First step only passes once all skills + education have been assessed and are success or on hold
    - Other steps must have all essential skills and claimed (in the snapshot) non-essenital technical skills assessed before the step is incremented
    - Any essential skill failure is immediate disqualification
    - Step success is only determined after all skills (excluding unclaimed technical skills) have been asssessed

### DIFF
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -1093,11 +1093,15 @@ class PoolCandidate extends Model
                     }
                 } else {
                     if ($poolSkill->skill->category === SkillCategory::TECHNICAL->name) {
+                        $isClaimed = false;
                         $snapshot = $this->profile_snapshot;
-                        $claimedSkills = collect($snapshot['userSkills']);
-                        $isClaimed = $claimedSkills->contains(function ($userSkill) use ($poolSkill) {
-                            return $userSkill['skill']['id'] === $poolSkill->skill_id;
-                        });
+
+                        if ($snapshot) {
+                            $claimedSkills = collect($snapshot['userSkills']);
+                            $isClaimed = $claimedSkills->contains(function ($userSkill) use ($poolSkill) {
+                                return $userSkill['skill']['id'] === $poolSkill->skill_id;
+                            });
+                        }
 
                         if (! $isClaimed) {
                             continue;
@@ -1188,7 +1192,7 @@ class PoolCandidate extends Model
 
         if ($currentStep >= $totalSteps) {
             $lastStepDecision = end($decisions);
-            if ($lastStepDecision['decision'] !== AssessmentDecision::HOLD->name) {
+            if ($lastStepDecision['decision'] !== AssessmentDecision::HOLD->name && ! is_null($lastStepDecision['decision'])) {
                 $overallAssessmentStatus = OverallAssessmentStatus::QUALIFIED->name;
                 $currentStep = null;
             }
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -1084,7 +1084,7 @@ class PoolCandidate extends Model
 
                         continue;
                     }
-                } else {
+                } else { // $poolSkill is an ASSET skill
 
                     // We do not need to evaluate non-essential technical skills that are not on
                     // the users snapshot, so skip the result check
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -23,6 +23,7 @@ use App\Enums\PositionDuration;
 use App\Enums\PriorityWeight;
 use App\Enums\ProvinceOrTerritory;
 use App\Enums\PublishingGroup;
+use App\Enums\SkillCategory;
 use App\Enums\WorkRegion;
 use App\Http\Resources\UserResource;
 use App\Observers\PoolCandidateObserver;
@@ -1028,8 +1029,10 @@ class PoolCandidate extends Model
         $this->load([
             'pool.assessmentSteps',
             'pool.assessmentSteps.poolSkills',
+            'pool.assessmentSteps.poolSkills.skill',
             'assessmentResults',
             'assessmentResults.poolSkill',
+            'user.userSkills',
         ]);
 
         foreach ($this->pool->assessmentSteps as $step) {
@@ -1038,6 +1041,7 @@ class PoolCandidate extends Model
             $hasOnHold = false;
             $hasToAssess = false;
 
+            $isApplicationScreening = $step->type === AssessmentStepType::APPLICATION_SCREENING->name;
             $stepResults = $this->assessmentResults->where('assessment_step_id', $stepId);
             $hasEssentialSkillsToAssess = $step->poolSkills->contains(function ($poolSkill) {
                 return $poolSkill->type === PoolSkillType::ESSENTIAL->name;
@@ -1046,17 +1050,17 @@ class PoolCandidate extends Model
             // If no essential skills are to be assessed
             // and not on application screening (requires education assessment)
             // then it is an automatic pass
-            if (! $hasEssentialSkillsToAssess && $step->type !== AssessmentStepType::APPLICATION_SCREENING->name) {
+            if (! $hasEssentialSkillsToAssess && ! $isApplicationScreening) {
                 $decisions[] = [
                     'step' => $stepId,
                     'decision' => AssessmentDecision::SUCCESSFUL->name,
                 ];
 
-                $currentStep++;
-
                 continue;
             }
 
+            // Nothing has been assessed or, not all claimed skills have been assessed
+            // so we can't assign a decision yet
             if ($stepResults->isEmpty()) {
                 $decisions[] = [
                     'step' => $stepId,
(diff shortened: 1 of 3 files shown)


esizer:
@tristan-orourke Going to work on tests next but I think the logic is close enough for you to take a more in depth look now :+1:

api/app/Models/PoolCandidate.php
```diff
@@ -1046,17 +1050,17 @@ public function computeAssessmentStatus()
...
                     'step' => $stepId,
                     'decision' => AssessmentDecision::SUCCESSFUL->name,
                 ];
 
-                $currentStep++;
-
                 continue;
             }
 
+            // Nothing has been assessed or, not all claimed skills have been assessed
```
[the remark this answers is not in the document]
tristan-orourke: This situation could also have a test. An assessment step with no skills should always pass.

api/app/Models/PoolCandidate.php
```diff
@@ -1066,62 +1070,79 @@ public function computeAssessmentStatus()
...
-                        case AssessmentDecision::HOLD->name:
-                            $hasOnHold = true;
-                            break;
-                        case AssessmentDecision::UNSUCCESSFUL->name:
-                            $hasFailure = true;
-                            break;
-                        default:
+
+                            continue;
+                        }
```
tristan-orourke, line 1114: This should be outside the TECHNICAL if branch, since it needs to apply to asset behvioural skills.

api/app/Models/PoolCandidate.php
```diff
@@ -1066,62 +1070,79 @@ public function computeAssessmentStatus()
...
+
+                    if ($decision === AssessmentDecision::HOLD->name) {
+                        $hasOnHold = true;
 
-                    switch ($decision) {
-                        case null:
+                        continue;
+                    }
+                } else {
+                    if ($poolSkill->skill->category === SkillCategory::TECHNICAL->name) {
```
tristan-orourke, line 1095: Given how much trouble we had wrapping our heads around this, this block could use a comment explaining why its here, I think 😛

tristan-orourke: And/or maybe just leave this whole block of pseudocode in the function's docblock to summarize the logic?
```
foreach step in pool->assessmentSteps
  foreach skill in assessmentStep->skills:
    result = find matching assessment result
    if skill is essential:
      if result is UNSUCCESSFUL, THEN mark UNSUCCESSFUL and exit loop
      if result is HOLD THEN mark HOLD and continue loop (to look for failures)
      if result is null or undecided, THEN mark TO ASSESS and continue loop (to look for failures)
    else if skill is asset:
      if skill is Technical AND user did not claim skill, THEN skip and continue loop
      ELSE if null or undecided THEN mark TO ASSESS and continue loop (to look for essential failures)
      ELSE mark nothing and continue, since the result doesn't actually matter
  AND if step is Application Assessment then repeat the Essential switch statement education assessment result  
  stepStatus is first of UNSUCCESSFUL, TO ASSESS, HOLD, and else QUALIFIED  
```

api/tests/Feature/CandidateAssessmentStatusTest.php
```diff
@@ -504,4 +505,124 @@ public function testOutOfOrderAssessmentsDoNotIncrementStep()
...
+
+        // Non-essential skills must also be assessed if a user claims them
+        $poolSkillTwo = PoolSkill::create([
+            'pool_id' => $pool->id,
+            'skill_id' => $technicalSkills[1]->id,
+            'type' => PoolSkillType::NONESSENTIAL->name,
+        ]);
+
+        $user = User::factory()->create();
+        UserSkill::factory()->create([
```
tristan-orourke, line 533: Do you not need to give the user the first skill as well?

api/tests/Feature/CandidateAssessmentStatusTest.php
```diff
@@ -504,4 +505,124 @@ public function testOutOfOrderAssessmentsDoNotIncrementStep()
...
+            'pool_id' => $pool->id,
+            'skill_id' => $technicalSkills[0]->id,
+            'type' => PoolSkillType::ESSENTIAL->name,
+        ]);
+
+        // Non-essential skills must also be assessed if a user claims them
+        $poolSkillTwo = PoolSkill::create([
+            'pool_id' => $pool->id,
+            'skill_id' => $technicalSkills[1]->id,
+            'type' => PoolSkillType::NONESSENTIAL->name,
```
tristan-orourke, line 529: You called the test `testMultipleEssential` but the second skill is nonessential 😉

api/tests/Feature/CandidateAssessmentStatusTest.php
```diff
@@ -504,4 +505,124 @@ public function testOutOfOrderAssessmentsDoNotIncrementStep()
...
+        $candidate = PoolCandidate::factory()->withSnapshot()->create([
+            'user_id' => $user->id,
+            'pool_id' => $pool->id,
+            'submitted_at' => config('constants.past_date'),
+            'expiry_date' => config('constants.far_future_date'),
+        ]);
+
+        $stepOne = $pool->assessmentSteps->first();
+
+        $stepTwo = AssessmentStep::factory()
```
tristan-orourke, line 547: I don't know if it matters, but I feel like you should create the assessment plan before creating the application with snapshot.

api/app/Models/PoolCandidate.php
```diff
@@ -1046,17 +1050,17 @@ public function computeAssessmentStatus()
...
+            if (! $hasEssentialSkillsToAssess && ! $isApplicationScreening) {
                 $decisions[] = [
                     'step' => $stepId,
                     'decision' => AssessmentDecision::SUCCESSFUL->name,
                 ];
 
-                $currentStep++;
-
                 continue;
             }
```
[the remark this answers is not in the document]
esizer: Oh yeah, good call. That was in fact not needed it seems! Removed in [remove some guards, fix conditions, add tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10925/commits/fb446bf12a5a1196542241022b8aa873527da26b)

api/app/Models/PoolCandidate.php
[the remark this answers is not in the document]
esizer: Sounds good, removed in [remove some guards, fix conditions, add tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10925/commits/fb446bf12a5a1196542241022b8aa873527da26b)

api/app/Models/PoolCandidate.php
```diff
@@ -1066,62 +1070,79 @@ public function computeAssessmentStatus()
...
-                    $hasToAssess = true;
-                }
+                        continue;
+                    }
 
-                foreach ($essentialSkillAssessments as $essentialSkillAssessment) {
-                    $decision = $essentialSkillAssessment->assessment_decision;
+                    if ($decision === AssessmentDecision::UNSUCCESSFUL->name) {
+                        $hasFailure = true;
+                        break;
```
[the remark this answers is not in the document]
esizer: Added in [document some trickier parts of logic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10925/commits/eee4591cb9a605be75160add3f222e69a6f92905)

api/app/Models/PoolCandidate.php
esizer: Oops, yeah good catch! Moved in [remove some guards, fix conditions, add tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10925/commits/fb446bf12a5a1196542241022b8aa873527da26b)

api/tests/Feature/CandidateAssessmentStatusTest.php
esizer: We don't need to. The logic only cares that the skill was assessed, it doesn't really care if the user claimed it or not. We assume the user has claimed the necessary skills since it is required to submit an application.

api/app/Models/PoolCandidate.php
esizer: Yeah, that makes sense. I hope I got it right in [document some trickier parts of logic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10925/commits/eee4591cb9a605be75160add3f222e69a6f92905)

api/app/Models/PoolCandidate.php
```diff
@@ -1066,62 +1070,79 @@ public function computeAssessmentStatus()
...
-                            $result->assessment_result_type !== AssessmentResultType::EDUCATION;
-                    });
+                if ($poolSkill->type === PoolSkillType::ESSENTIAL->name) {
+                    if (! $result || is_null($result->assessment_decision)) {
+                        $hasToAssess = true;
 
-                if ($essentialSkillAssessments->isEmpty()) {
-                    $hasToAssess = true;
-                }
+                        continue;
```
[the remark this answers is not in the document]
esizer: I personally prefer `continue` statements over `else if` and find it easier to read and understand :sweat_smile: Unless this is incorrect, the continue statements are also avoiding the need to execute code that won't impact then end result.

api/tests/Feature/CandidateAssessmentStatusTest.php
esizer: Oof yeah, that test name should be more accurate after [adjust test names, order of creation](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10925/commits/db93cca5019117064fe22794f3e8f8da2267b2a3)

api/tests/Feature/CandidateAssessmentStatusTest.php
esizer: Yeah, not necessary but makes the test easier to follow so I updated the order in [adjust test names, order of creation](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10925/commits/db93cca5019117064fe22794f3e8f8da2267b2a3)

api/tests/Feature/CandidateAssessmentStatusTest.php
tristan-orourke: Right!

tristan-orourke (commented):
I'm totally happy with the code now! But I have meetings the rest of the day and won't have time to manually test this any time soon.

api/app/Models/PoolCandidate.php
vd1992, line 1: A user never needs to attach a justification to behavioural skills, therefore supporting evidence will likely be always empty making it best to maybe not render it for behavioural skills in the first place?
But this too would be a separate issue

image

vd1992, line 1: A bug maybe, I tested skipping a step like so

image

Step 3 correctly shows as still to assess but shouldn't step 4 appear as completed?

image

vd1992, line 1: Point of confusion
In the application workflow an applicant never needs to "claim" behavioural skills so they will almost always be unclaimed. Probably yet another thing to spin off 😛 

image

Essential behavioural skills don't have this text so no worries

image

Application process

image

vd1992, line 1: How should an assessment step connected to a sole asset skill be handled?

image

vd1992 (commented):
Giving Tristan a helping hand here

Sorta just did my own thing and did a scattershot approach of noting whatever stood out to me. Please see the comments and offer input/rebuttals

api/app/Models/PoolCandidate.php
vd1992: Weird, it auto-passed when I complete the Education requirement, probably not supposed to do that

image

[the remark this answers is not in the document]
esizer: Yeah, @Jerryescandon mentioned the copy needs to change and is working on a separate ticket for that.

esizer: Heh, yeah that is confusion but what you see there is correct. In the tracker view, you only show up in the column past your last successful step.

So, you could pass all the steps but #2 and you would still only show up in the second column.

esizer: Yeah, it is odd. The reason this happens is becuase right now the status is only computed when you create or update a result. So that is why you only see it after assessing one item :sweat_smile: 

I think maybe we need to dispatch the event on application submission as well. :thinking:

Jerryescandon: Yeah definitely

tristan-orourke (approved):

tristan-orourke:
Okay! Thanks @vd1992 for double checking everything. Seems like we have some ideas for copy improvements for a separate ticket, but this PR fixes what it needs to.

api/app/Models/PoolCandidate.php
vd1992: Ah... Okay 😵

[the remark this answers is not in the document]
Jerryescandon: Yup, new ticket will come in for that.

Jerryescandon: Yup. Ideally, we wouldnt allow advertisements to be created that include behavioural-non-essential skills. So the actual problem is on the advertisement creator. But for now, this is correct. All behavioural skills are assumed to be claimed.

Jerryescandon: Im also making a ticket to make it clear that these steps are instead "Not applicable" when non of the relevant skills have been claimed.

PR #10936, opened 2024-07-10
[Debt] Compress assets at built time

### LINKED ISSUE #10935: ♻️ Add compression to vite build
## ♻️ Debt/Refactor

Pre-compress files at build time.

## 🕵️ Details

I did not realize we were doing compression at build time so we lost it when moving to vite.

## 🙋‍♀️ Proposed Solution

Add the compression plugin to pre-compress files at build time so the server doesn't have to do it.

## ✅ Acceptance Criteria

- [ ] Assets are gzipped at build time
- [ ] Compressed versions are served instead of full versions

esizer:
🤖 Resolves #10935 

## 👋 Introduction

Compresses our assets at build time.

## 🧪 Testing

1. Build the project `pnpm run build:fresh`
2. Navigate to the homepage
3. Confirm the assets are not the full size
4. Confirm the header exists `Content-Encoding: gzip`

## 📸 Screenshot

screenshot_10-Jul-2024_13-09-1720631383

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -92,13 +92,14 @@
     "eslint": "^8.57.0",
     "jest": "^29.7.0",
     "jest-axe": "^9.0.0",
-    "jest-fail-on-console": "^3.3.0",
     "jest-diff": "^29.7.0",
+    "jest-fail-on-console": "^3.3.0",
     "prettier": "^3.3.2",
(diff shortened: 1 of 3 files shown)


petertgiles (approved):
Now that's more like it!
image

PR #10938, opened 2024-07-10
[Debt] Discover enums during registration

### LINKED ISSUE #10861: ♻️ Make localized enums discoverable
## ♻️ Debt/Refactor

Provide a way to discover all localized enums and implment it in the graphql provider registration.

## 🕵️ Details

With the move of localized enums to the server, we now need to register them a little differently in `lighthouse-php` graphql schema. Currently, we use a hard coded array of these enums to loop over. Unfortunately, this is going to be annoying to maintain and is open to human error. 

We would like a way to discover all the enums that have been localized and simply loop over those so they get registered automatically assuming they implement the `HasLocalization` trait.

## 🙋‍♀️ Proposed Solution

@petertgiles found a package ([`spatie/php-structure-discoverer`](https://github.com/spatie/php-structure-discoverer)) that seems like it would be a good fit for this.

## ✅ Acceptance Criteria

- [ ] Localized enums can be automatically discovered
- [ ] Hardcoded array removed from `GraphQLServiceProvider`
- [ ] All localized enums are registered and function as expected

esizer:
🤖 Resolves #10861 

## 👋 Introduction

Updates our graphql registration of enums to use discovery so we don't need to manually add them to registration.

## 🧪 Testing

1. Confirm enums still work as expected in our graphql server


esizer:
I can't figure out why the [schema check fails consistently](https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/9880283150/job/27288464119?pr=10938). I keep trying to run it locally and it doesn't change :face_with_spiral_eyes:

api/app/Providers/GraphQLServiceProvider.php
```diff
@@ -139,642 +33,62 @@ static function (): EnumType {
...
-                ]);
-            }
-        );
+        foreach ($enums as $enum) {
+            $name = class_basename($enum);
+            $typeRegistry->registerLazy(
+                $name,
+                static function () use ($name, $enum): EnumType {
+                    /** @disregard P1013 */
+                    $values = array_column($enum::cases(), 'name');
```
esizer, line 50: Does anyone know how to typehint this so we don't need to disregard the error about missing `cases()` method? :thinking:

api/app/Providers/GraphQLServiceProvider.php
```diff
@@ -139,642 +33,62 @@ static function (): EnumType {
...
+                            'value' => Type::nonNull($typeRegistry->get($name)),
+                            'label' => Type::nonNull($typeRegistry->get('LocalizedString')),
+                        ];
+                    },
+                    /** @disregard P1003 */
+                    'resolveField' => function ($value, array $args, $context, ResolveInfo $info) use ($enum) {
+                        switch ($info->fieldName) {
+                            case 'value': return $value;
+                                /** @disregard P1013, these enums do have the trait */
+                            case 'label': return $enum::localizedString($value);
```
esizer, line 85: Similar here, I wonder if there is a way to typehint this? I had a `method_exists` check and even then it was still not happy :confounded:

petertgiles (approved):
Nice, this is so much cleaner.  I was going to suggest looking at the caching but I suppose since this is in a service provider it's not necessary.

api/app/Providers/GraphQLServiceProvider.php
```diff
@@ -139,642 +33,62 @@ static function (): EnumType {
...
-            static function (): EnumType {
-                return new EnumType([
-                    'name' => 'NotificationFamily',
-                    'values' => array_column(NotificationFamily::cases(), 'name'),
-                ]);
-            }
-        );
+        $enums = Discover::in(app_path('Enums'))
+            ->enums()
+                // Language has a custom implementation
```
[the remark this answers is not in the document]
esizer: Yup! Created https://github.com/GCTC-NTGC/gc-digital-talent/issues/10964

api/app/Providers/GraphQLServiceProvider.php
petertgiles: I think this would do it?
image

api/app/Providers/GraphQLServiceProvider.php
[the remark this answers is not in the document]
esizer: okay, added some coments (let me know if they arentgood enough :sweat_smile: ). Alos, thanks to your comment about typehints, I was able to reduce the need for the `@disregards` in [reduce need for lsp disregards](https://github.com/GCTC-NTGC/gc-digital-talent/pull/10938/commits/d45d1a0facf4ba358b1ab99a8ba4f0d361103bdf) :tada:

api/app/Providers/GraphQLServiceProvider.php
esizer: Of course it was `UnitEnum` I was trying just `Enum` :woman_facepalming:

petertgiles: Yeah, it's a bit of a weird one.  Took some googling until I found https://www.php.net/manual/en/class.unitenum.php

PR #10962, opened 2024-07-11
[Fix] Add fallback to preferred locale

### LINKED ISSUE #10862: 🐛 Error when sending a notification to a user with no preferred lang
## 🐛 Bug

Our [User->PreferredLocale function](https://github.com/GCTC-NTGC/gc-digital-talent/blob/2c0c27f2e1ad20d3b4ed68ff26c548c9a464dba2/api/app/Models/User.php#L198) is hinted to only return strings.  However, it can return nulls and will error out when it does.

## 📋 Steps to Reproduce

1. Manually null out the preferred_lang field for a user. 
2. Attempt to send a notification to them.

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

The HasLocalePreference trait is actually hinted to return string|null.  We could align with that and then put in handlers for nulls whereever it is used.

Option 2 is to just fallback to one of the languages and continue to promise not to return nulls.

## ✅ Acceptance Criteria

- [ ] No errors when trying to send a notification to a user with no preferred lang.

esizer:
🤖 Resolves #10862 

## 👋 Introduction

Adds a fallback (English) to the users preferred locale

## 🧪 Testing

1. Null out a users `preferred_lang`
2. Attempt to send a notification to them
3. Confirm to null locale error

### DIFF
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -198,7 +198,7 @@ class User extends Model implements Authenticatable, HasLocalePreference, Laratr
      */
     public function preferredLocale(): string
     {
-        return $this->preferred_lang;
+        return $this?->preferred_lang ?? 'en';
     }
 
     public function pools(): HasMany


petertgiles (approved):
Thanks.  :+1:

PR #10968, opened 2024-07-11
[Debt] Merge pool candidates queries

### LINKED ISSUE #10926: ♻️ Too many Candidate Search Page inital requests
## ♻️ Debt/Refactor

Our Candidate search page (/admin/pool-candidates) currently fires 6 network requests after the standard first few, when starting up.

1. PoolCandidateFilterDialog_Query to get all the dimensions for the filter dialog
2. PoolCandidatesTableStrings to get the names of some of the enums
3. CandidatesTableCandidatesPaginated_Query to get the actual candidates
4. CandidatesTable_Query to get... more skills and departments?
5. JobPlacementOptions to get... more enum names?
6. JobPlacementOptions again?

Can we combine or remove some of these to lower the network chatter and improve the latency of the page?

## 🙋‍♀️ Proposed Solution

Some of these seem redundant, like the double JobPlacementOptions requests.  Can we find out why it is fired twice and lower it to once?

GraphQL has the ability to combine multiple queries into a single network request.  The first four are fired at exactly the same time which suggests to me that they could rolled into the same request.

## ✅ Acceptance Criteria

- [ ] The candidate search page has fewer initial network requests.

esizer:
🤖 Resolves #10926 

## 👋 Introduction

This merges a lot of the queries being run on the "Candidate search" page using fragments to run them all at the page level.

## 🕵️ Details

While working on this I noticed two things:

1. We are needed to drill a lot of props (see job placement) and I was wondering if we could work on this a bit :sweat_smile: 
2. We have a few other places that could probably be merged using fragments
3. We are grabbing some data before we actually need it (i.e all skills/pools for the comboboxes in the filter dialog) 

## 🧪 Testing

1. Build `pnpm run dev`
4. Inspect the network tab to see reduced query count. Should be 4 (minimum amount I could do)
    1. Auth query
    2. Site banner
    3. Candidates
    4. Additional data for filters/dialogs

## 📸 Screenshot

screenshot_11-Jul-2024_14-27-1720722449

### DIFF
--- a/apps/web/src/components/PoolCandidatesTable/JobPlacement.stories.tsx
+++ b/apps/web/src/components/PoolCandidatesTable/JobPlacement.stories.tsx
@@ -45,7 +45,7 @@ export default {
   decorators: [OverlayOrDialogDecorator, MockGraphqlDecorator],
   args: {
     defaultOpen: true,
-    jobPlacementOptions: makeFragmentData(
+    optionsQuery: makeFragmentData(
       {
         departments,
         placementTypes: fakeLocalizedEnum(PlacementType),
--- a/apps/web/src/components/PoolCandidatesTable/JobPlacement.stories.tsx
+++ b/apps/web/src/components/PoolCandidatesTable/JobPlacement.stories.tsx
@@ -18,6 +18,7 @@ import {
 
 import JobPlacementDialog, {
   JobPlacementDialog_Fragment,
+  JobPlacementOptions_Query,
 } from "./JobPlacementDialog";
 
 const fakedCandidate = fakePoolCandidates(1)[0];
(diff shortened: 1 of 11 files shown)


vd1992:
Out of scope, but looks like there will be more places where this could be done. For example the view a candidate page

Five on load

image

One for Removal options, which is an enum

image

Another for statuses, which is also an enum

image

vd1992:
For the above, you nestled one enum in already for the page

image

esizer:
> Out of scope, but looks like there will be more places where this could be done. For example the view a candidate page

Yeah, I am kind of hoping we can take a look at how many queries we are doing on each page and try and merge them where possible :sweat_smile:

vd1992:
Yeah, I think the best way would be to spin off an issue each time we find some place to improve. 

Could you write up an issue for my comments above please? I am not sure what the AC or details would look like since I am not entirely familiar yet with what is going on with the enum fetching arrangement

vd1992 (approved):
Seems to work well for me, not noticing any issues. Nice!

PR #10970, opened 2024-07-11
[Bug] Filtering for archived pools fix

### LINKED ISSUE #10865: 🐛 Cannot filter for archived pools
## 🐛 Bug

You can filter for archived pools in the pools table but this returns nothing even if you should see something

## 📋 Steps to Reproduce

1. Archive a pool
2. Attempt to find it in the pools table using the filters

## 📸 Screenshot

Archived

image

Filtering

image

Nothing

image

## 🙋‍♀️ Proposed Solution

Scope filters out archived pools always, should only do so when not searching for archived pools

image

## ✅ Acceptance Criteria

- [ ] When filtering for archived pools, you see archived pools
- [ ] By default the pool table should still load the unarchived pools
- [ ] All the community roles can see the unarchived pools

vd1992:
🤖 Resolves #10865

## 👋 Introduction

Enable filtering for archived pools to work per intent. Filtered out by default, but returned if wanted

## 🕵️ Details

Just kicking the scope off the query as a blanket application and into the fields scope functions.
Permissions was decided to not be a problem for archived pools

## 🧪 Testing

1. Navigate to pools table and also ensure you have a pool archived if not seeded
2. Default view returns all but archived pools
3. Play with the filters in the dialog, trying different combinations and check results
4. Ensure you can view archived pools when filtering for them

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -250,8 +250,7 @@ type Pool implements HasRoleAssignments {
   location: LocalizedString @rename(attribute: "advertisement_location")
   securityClearance: LocalizedSecurityStatus
     @rename(attribute: "security_clearance")
-  language: LocalizedPoolLanguage
-    @rename(attribute: "advertisement_language")
+  language: LocalizedPoolLanguage @rename(attribute: "advertisement_language")
   status: LocalizedPoolStatus @rename(attribute: "status")
   stream: LocalizedPoolStream
   processNumber: String @rename(attribute: "process_number")
@@ -898,11 +897,7 @@ type Query {
         ]
       )
   ): [Pool]!
-    @paginate(
-      defaultCount: 10
-      maxCount: 1000
-      scopes: ["authorizedToView", "notArchived"]
-    )
+    @paginate(defaultCount: 10, maxCount: 1000, scopes: ["authorizedToView"])
     @canResolved(ability: "view")
   poolCandidate(id: UUID! @eq): PoolCandidate
     @find
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -740,7 +740,7 @@ input PoolFilterInput {
   name: String @scope
   team: String @scope
   streams: [PoolStream!] @scope
-  statuses: [PoolStatus!] @scope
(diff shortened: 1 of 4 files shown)


api/graphql/schema.graphql
```diff
@@ -250,8 +250,7 @@ type Pool implements HasRoleAssignments {
   location: LocalizedString @rename(attribute: "advertisement_location")
   securityClearance: LocalizedSecurityStatus
     @rename(attribute: "security_clearance")
-  language: LocalizedPoolLanguage
-    @rename(attribute: "advertisement_language")
+  language: LocalizedPoolLanguage @rename(attribute: "advertisement_language")
```
vd1992, line 253: Format on save

api/graphql/schema.graphql
```diff
@@ -741,7 +740,7 @@ input PoolFilterInput {
   name: String @scope
   team: String @scope
   streams: [PoolStream!] @scope
-  statuses: [PoolStatus!] @scope
+  statuses: [PoolStatus!] = [] @scope
   processNumber: String @scope
```
vd1992, line 744: If the field is not present, the scope is not run. Given the scope always needs to be used, even if just to apply the archived filtering, a default must exist

esizer (changes requested):
Nice! Seems to do exactly what we want :+1: 

Can you also write a tests for this? I think three should assertions should it :thinking: 

- Create 2 pools (one archived, one not) 
   - Assert archived filtered by default
   - Assert passing in the non-archived pool returns only archived
   - Assert passing in an unrelated status returns no results

vd1992:
> Nice! Seems to do exactly what we want 👍
> 
> Can you also write a tests for this? I think three should assertions should it 🤔
> 
>     * Create 2 pools (one archived, one not)
>       
>       * Assert archived filtered by default
>       * Assert passing in the non-archived pool returns only archived
>       * Assert passing in an unrelated status returns no results

We already had a test for this scope, so I just added on to it a bit in https://github.com/GCTC-NTGC/gc-digital-talent/pull/10970/commits/b56c05954104bd2dc59118082351eb37e1cd8e92

esizer (approved):
Cool! :sunglasses:

PR #10973, opened 2024-07-12
Update to new blocking-issues version

tristan-orourke:
## 👋 Introduction

This should fix a bug with the blocking-issues action, where the `blocked: dependencies` label is not automatically added or removed from some tickets when it should be.

## 🕵️ Details

I think the root of the issue was that the bot was assuming that issues were always written as `#123`. In fact, they are often written as a full url (eg `https://github.com/org/repo/issues/123`). Thus, it was not parsing many of them correctly.

## 🧪 Testing

I'm not sure if this _can_ be tested? I believe its set up to only run on main. :/

### DIFF
--- a/.github/workflows/blocking-issues.yml
+++ b/.github/workflows/blocking-issues.yml
@@ -12,7 +12,7 @@ jobs:
     name: Checks for blocking issues
 
     steps:
-      - uses: tristan-orourke/blocking-issues@v1.1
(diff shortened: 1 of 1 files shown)


mnigh (approved):
> I'm not sure if this can be tested? I believe its set up to only run on main. :/

🧪 :shipit:

PR #10974, opened 2024-07-12
[Debt] Tweak notifications for performance

### LINKED ISSUE #10896: 🐛 System Notification spikes performance usage
## 🐛 Bug

We just sent our first system notification and it impacted the site's performance.  We should probably slow this down a bit.

## 📋 Steps to Reproduce

1. Send a system notification to a large number of users

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

- Slow down firing the notifications in the foreach loop in the Artisan command?
- Use [delayed dispatching](https://laravel.com/docs/10.x/queues#delayed-dispatching) for emails?
- Add a [sleep time](https://laravel.com/docs/10.x/queues#worker-sleep-duration) between jobs for queue workers?
- Use queued [events](https://laravel.com/docs/10.x/events) to generate the notifications (for pool publish)

## ✅ Acceptance Criteria

- [ ] System notification resource usage is spread out

esizer:
🤖 Resolves #10896 

## 👋 Introduction

Adjusts sending notifications to large number of users to save our poor CPU :cry: 

## 🕵️ Details

This attempts to spread out the load a bit so our CPU usage doesn't spike as hard when sending out notifications.

- Updates the system notification command to split the users into chunks so we are doing less work at once
- Moves the pool published notification to a queued job as well as doing the same chunking as system notifications

## 🧪 Testing

> [!NOTE]
> Testing the real benefits of this is difficult since it is sort of hardware dependent but, you can test to confirm it works. You can, if you'd like `top` , `sar -u 5` or `docker stats` while this is running to inspect CPU usage but note that it won't really be comparable to prod 

1. Seed a lot of users (~20,000 if you want to get close to prod)
2. Start up your queue worker `make queue-work`
3. Run the system notification `make artisan CMD="send-notification:system notification_announcement"`
4. Observe the queue being processed
5. Publish a pool
6. Observe the queue being processed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,4 +48,4 @@ lint-php:
 	$(DOCKER_API) "vendor/bin/pint --test"
 
 queue-work:
-	docker-compose exec webserver sh -c "runuser -u www-data -- php /home/site/wwwroot/api/artisan queue:work --timeout=120"
+	docker-compose exec webserver sh -c "runuser -u www-data -- php /home/site/wwwroot/api/artisan queue:work"
--- a/Makefile
+++ b/Makefile
@@ -48,4 +48,4 @@ lint-php:
 	$(DOCKER_API) "vendor/bin/pint --test"
 
 queue-work:
-	docker-compose exec webserver sh -c "runuser -u www-data -- php /home/site/wwwroot/api/artisan queue:work"
+	docker-compose exec webserver sh -c "runuser -u www-data -- php /home/site/wwwroot/api/artisan queue:work --timeout=120"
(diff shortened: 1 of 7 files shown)


petertgiles (approved):
I did deploy to Azure, seed about 15,000 users (before I hit the PHP memory limit), and set the GCNotify API key to "test" (no actualy email sent).

When I sent the system notification it took 3-5 minutes to run and didn't seem to impact site performance.  🥳   In these graphs the notification started around 8:20 PM - results seem uneventful:
image
I'm not sure why the app plan CPU usage started around 50%.  Maybe it's a weaker plan?  Maybe I don't really understand how the app plan CPU measurement works?  😁  Anyway, it didn't seem to spike when I was working.
Sending all the queued notifications took about 15 minutes before I killed the last 3000 or so.

I also published a pool and it happened almost instantly.
image

It looks like the event queued about 7000 notifications for itself.  🥳 


Nice one!

PR #10978, opened 2024-07-12
[Feature] Add intl command to api

### LINKED ISSUE #10900: ✨ Translations workflow for backend strings
## ✨ Feature

We now have a lot of localized strings in the backend which are invisible to any checking by the translations workflow.  Let's close that gap.  

## 🕵️ Details

- A new workflow or step to existing workflow? new workflow
- Have workflow launch a new artisan command?

## ✅ Acceptance Criteria

- [ ] Errors if any enums reference a non-existent file
- [ ] Errors if any enum cases don't exist in both EN and FR
- [ ] Errors if any strings exist in one but not the other language

```[tasklist]
### Blocked By
- [ ] #10634
- [ ] https://github.com/GCTC-NTGC/gc-digital-talent/issues/10861
```

esizer:
🤖 Resolves #10900 

## 👋 Introduction

Adds a command to verify strings on the api.

## 🕵️ Details

Similar to the `pnpm run check-intl:*` scripts, this checks our new api strings. It looks for:

 - Missing files for an enum
 - Missing strings for each localized enum string
 - Strings that match exactly

## :wrench: Usage

Run the command with: `make artisan CMD="app:check-intl"` which outputs a list of errors if there are any and stores the output json in `api/storage/app`.

If there are known missing strings (we have some) or known exact matches where the word is known to be the same in English and French, you can add them to the allow lists defined on the command.

If the allow lists get too large then maybe we can consider moving them to a config file? :thinking: 

## 🧪 Testing

1. Run `make artisan CMD="app:check-intl"`
2. Confirm success
3. Force errors by deleting some files in `api/lang` and commenting out items in the allow list
4. Confirm errors appear
5. Confirm [action runs as expected](https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/9913720763?pr=10978)

### DIFF
--- a/api/app/Console/Commands/CheckIntl.php
+++ b/api/app/Console/Commands/CheckIntl.php
@@ -57,8 +57,8 @@ class CheckIntl extends Command
      * to have an associated string
      */
     private $knownMissing = [
-        // 'assessment_result_justification.skill_accepted',
-        // 'indigenous_community.legacy_is_indigenous',
+        'assessment_result_justification.skill_accepted',
+        'indigenous_community.legacy_is_indigenous',
     ];
 
     /**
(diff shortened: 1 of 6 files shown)


api/app/Console/Commands/CheckIntl.php
```diff
@@ -0,0 +1,211 @@
...
+
+    /**
+     * Missing strings
+     */
+    private $missingStrings = ['en' => [], 'fr' => []];
+
+    /**
+     * Matching strings
+     *
+     * Any string that is the same in both locales
```
mnigh, line 46: What does it usually indicate? I think something is missing in the comment.

mnigh (commented):
Cool! Is there any way to check whether a key is missing a value or has an empty string as a value? Is there a way to check for orphan French strings and files? Do you think you could add some instructions for the new process to `documentation/translation.md`?

api/app/Console/Commands/CheckIntl.php
```diff
@@ -0,0 +1,202 @@
...
+        $keys = Arr::map(array_column($enum::cases(), 'name'), function ($case) {
+            return strtolower($case);
+        });
+
+        $existing = [
+            'en' => Lang::get(key: $fileName, locale: 'en'),
+            'fr' => Lang::get(key: $fileName, locale: 'fr'),
+        ];
+
+        $exisingKeys = [
```
mnigh, line 143:
```suggestion
        $existingKeys = [
```

api/app/Console/Commands/CheckIntl.php
```diff
@@ -0,0 +1,202 @@
...
+                // If a key is an int it means it was input incorrectly and the value is the expected key (probably)
+                $key = $keyIsInt ? $value : $key;
+
+                // Key is an int so not likely to have a proper $key => $value
+                if ($keyIsInt || ! $value) {
+                    $this->errors['empty_strings'][$locale][$fileName][] = $key;
+                }
+
+                // Key exists in one locale but not the other
+                if (! in_array($key, $exisingKeys[$oppositeLocale])) {
```
mnigh, line 164:
```suggestion
                if (! in_array($key, $existingKeys[$oppositeLocale])) {
```

mnigh (changes requested):

mnigh (approved):
👏 fantastique!

PR #10993, opened 2024-07-17
[Tooling] Turn off PR comments by codecov

brindasasi:
## 👋 Introduction

This PR turns off PR comments by codecov. 
I'll get a different PR for adding as status check but the comments are annoying currently. 
This can go. 

## Testing 
I think this PR itself should not get any comment from codecov.


esizer (approved):
:bow: Thanks!

Issue #10799, opened 2024-06-26
♻️ Avoid use of entire model types in Typescript


esizer:
## ♻️ Debt/Refactor

Migrate away from using the whole model types and prefer discrete types or fragments.

## 🕵️ Details

We currently use entire models as types (`Pool`, `Department`) to make passing data around easier. However, this causes issues when we encounter non nullable types. Our grapqhl codegen will not mark these as optional. So, when we use the entire type, it forces us to fetch that data to satisfy Typescript. 

This leads to either doing some typescript gymnastics or over fetching of data.

## 🙋‍♀️ Proposed Solution

We have two approaches we could take and likely will need to do both depending on the case:

1. Replace entire types like `pool: Pool` with more specific arguments `{ stream: PoolStream, publishingGroup: PublishingGroup }`
2. Utilize graphql fragments as the argument `poolQuery: FragmentType<typeof HelpFunctionPool_Fragment>`

I think we can follow a good rule of thumb to accomplish this. 

If it is a helper function that only accepts a couple of arguments, just separate them out and make us pass them in individually (option 1).

Then, if you start to see the number of arguments increase to something that seems to be hard to manage (4+?) then we use option 2 and write a fragment for it. Also, if it is a component we should drastically reduce the threshold in which we decide to use a fragment (2+?)

## ✅ Acceptance Criteria

- [ ] Entire model types are no longer used with few exceptions
- [ ] Exceptions are documented to explain why the entire model type was required
- [ ] Document the rule of thumb for how to use fragments

vd1992:
Sometimes all you need to do is import the fragment and use it to define type. 
Example for reference

https://github.com/GCTC-NTGC/gc-digital-talent/pull/10629/commits/0865ff9e2a9e4f04e1d6953468aea8956660be16

vd1992:
### Pull requests

- https://github.com/GCTC-NTGC/gc-digital-talent/pull/10994

PR #10994, opened 2024-07-17
[Debt] Constrain types classification, team, and department

vd1992:
🤖 Progresses #10799

## 👋 Introduction

Chips away at the issue by targeting types
`Team`
`Classification`
`Department`

## 🕵️ Details

Make the issue less daunting by electing to chip away at it over time.
There are many types to look at, but it is complicated by nesting of types too. For instance, may need to shake out `Classification` more once type `Pool` has been constrained as the former can be nested in the latter

Thoughts? @esizer 

## 🧪 Testing

1. All functionality related to the three types still works
2. Nothing left to change

### DIFF
--- a/apps/web/src/components/PoolCard/PoolCard.tsx
+++ b/apps/web/src/components/PoolCard/PoolCard.tsx
@@ -86,7 +86,7 @@ export const PoolCard_Fragment = graphql(/* GraphQL */ `
 
 const getSalaryRange = (
   locale: string,
-  classification?: Maybe<Classification>,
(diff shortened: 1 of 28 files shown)


esizer (approved):
Awesome! :grin: 

Shaved off 2kb gzipped of our grpahql package :laughing: I'm sure we will really start seeing benefits when we get to the larger ones!

```sh
# befpre
graphql-C5G_noJX.js    1,189.33 kB │ gzip:  56.48 kB

# after
graphql-DLDhSQlI.js   1,182.03 kB │ gzip:  56.25 kB 

```

PR #11013, opened 2024-07-18
[Debt] Update frontend with new final decision sort

### LINKED ISSUE #9792: ✨ Add sorting to "Final decision" column on Talent placement tab
## ✨ Feature

Users would like to be able to sort this column to easily group users by their final decision status.

## 🕵️ Details

The order of statuses should be as following:

Qualified
Qualified pending decision
To assess step 1
To assess step 2
....
Disqualified
Disqualified pending decision
Removed

## 📸 Screenshot

image

## ✅ Acceptance Criteria

- [ ] Final decision column is sortable on "Talent placement" table.
- [ ] Final decision column sorting follows the logic described above

## 🛑 Blockers
```[tasklist]
### Blocked By
- [ ] #8966 
- [ ] https://github.com/GCTC-NTGC/gc-digital-talent/issues/10846
```

esizer:
🤖 Resolves #9792 

## 👋 Introduction

Updates the tables with pool candidates use the new final decision weight column for sorting.

## 🧪 Testing

> [!NOTE]
> It may help to create pool candidates with different final decisions to better test this.
> You can find different options in the weight list following this.

2. login as admin `admin@test.com`
3. Navigate to the tables with final decision:
   - Candidate search
   - Process > Talent placement
   - User > User information
4. Confirm sorting follows the proper weights outlined in:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/2c48f60ec250d54805f1da0f14a07b6d8b0d62cb/api/app/Models/PoolCandidate.php#L1285-L1299


mnigh (changes requested):
👍 Sorts very nicely, respects weight and it pushes bookmarked candidates to top.

Not explicit in the acceptance criteria, but could you update the fake data to use final decision states instead of the old status so that storybook reflects the current fields?

https://www.chromatic.com/pullrequest?appId=61e099a1bb1465003a73d3ce&number=11013&activeElementId=comparison-test-key-663e7729d4fd93bc5377779a-1200px&view=changes

## Screenshot
Screen Shot 2024-07-18 at 15 57 48

esizer:
> Not explicit in the acceptance criteria, but could you update the fake data to use final decision states instead of the old status so that storybook reflects the current fields?

Yup, updated in: [add final decision to fake data](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11013/commits/41230df5f15d8eedf7a30c4327b494a13fd9976b)

Also, it is annoying that all of our candidates are the exact same. I tihnk it has something to do with the seed being 0 forcing the `faker.helpers.arrayElement` to pick the same user every loop iteration. :confounded: 

I created  a bug for this because right now it is not very helpful for some snapshots to just show the same data multiple times.

https://github.com/GCTC-NTGC/gc-digital-talent/issues/11017

mnigh (approved):
👍

PR #11020, opened 2024-07-19
[Fix] Indicate to user work is occurring on experience form

### LINKED ISSUE #10783: 🐛 Updating Experience with linked Skills can be confusingly slow
## 🐛 Bug

Certain users have reported needing to save experiences / skills several times before it works. 

We've seen in monitoring that UpdateWorkExperience requests have taken up to 18 seconds! We suspect that the operation is taking so long, without ongoing feedback, that user's think it finished but failed, instead of being slow.

The issue is probably in the `Experience.php connectSkills()`, which is looks complex and might very well be slow for a higher number of skills.

## 🦋 Expected Behaviour

1. update____Experience mutations should take no more than a second or two to complete.
2. It should be clear to users when the operation completes vs is still in progress.

## 🕵️ Details

## 📋 Steps to Reproduce

1. Find the log from the production that takes 18 secs and analyse it
2. ...

## 📸 Screenshot

## 🙋‍♀️ Proposed Solution

😬 I haven't looked at the connectSkills function since I wrote it - don't have thoughts yet on improving it.

~QUESTION: should we change the UI at all to better indicate when a request is in progress? Talk to designers.~

## ✅ Acceptance Criteria

- [ ] update____Experience mutations should take no more than a second or two to complete.
- [ ] Upon clicking the save button , button goes disabled and caption should say "Submitting" until the request returns. 
- [ ] Close the save dialog only when the request is completed successfully. 
- [ ] Have the dialog open when the request is failed and show the failure toast message.

esizer:
🤖 Resolves #10783

## 👋 Introduction

Updates the experience form to disable the submit button and indicate that the experience is being saved.

## 🕵️ Details

We were having a hard time replicating the actual issue of adding experiences being slow locally so this completes everything except fixing the query :sweat_smile: 

It would helpful if we had some more logs around the actual query from a deployed vertical to see how it performs in there since local is not great for finding it out.

## 🧪 Testing

2. Login as any user
3. Throttle yhour connection
4. Add an experience
5. Confirm the submit button indicates work is happening while it saves


vd1992 (approved):
Seems to work fine to me

PR #11030, opened 2024-07-22
[Fix] Set a max page for notifications paginatation

### LINKED ISSUE #11024: 🐛 No limit for notification page queries
## 🐛 Bug

There is no limit to how many notification page queries can be requested.

## 🦋 Expected Behaviour

There should be a max number of how many pages can be queried.

## 📋 Steps to Reproduce

> [!WARNING]  
> Be prepared for your machine to 💥 if you do this.

1. Navigate to `http://localhost:8000/en/applicant/notifications?page=10`
2. Change `page` query parameter to something like 10000
3. Observe your machine grind to a halt

## 📸 Screenshot

Screen Shot 2024-07-19 at 14 04 46

## ✅ Acceptance Criteria

- [ ] max number of how many pages can be queried for notifications pagination

esizer:
🤖 Resolves #11024 

## 👋 Introduction

Prevents users from requesting more than the max number of pages available.

## 🕵️ Details

While our throttle would prevent this from causing too much issue on the server, the user could request a max array size and crash their client. This helps avoid that mistake.

## 🧪 Testing

2. Login as any user
3. Navigate to `http://localhost:8000/en/applicant/notifications?pages=5000`
4. Confirm that the number of queries created is not 5000 but the max pages (should be 1 since no notifications exist)

### DIFF
--- a/apps/web/src/components/NotificationList/NotificationList.tsx
+++ b/apps/web/src/components/NotificationList/NotificationList.tsx
@@ -1,4 +1,5 @@
 import { useSearchParams } from "react-router-dom";
+import { useQuery } from "urql";
 
 import { graphql } from "@gc-digital-talent/graphql";
 import { unpackMaybes } from "@gc-digital-talent/helpers";
@@ -10,6 +11,16 @@ import NotificationActions from "./NotificationActions";
 import NotificationListPage from "./NotificationListPage";
 import NotificationItem from "./NotificationItem";
 import NotificationPortal from "./NotificationPortal";
+
+const MaxNotifcationPages_Query = graphql(/* GraphQL */ `
+  query MaxNotificationPages($where: NotificationFilterInput) {
+    notifications(where: $where, page: 1) {
+      paginatorInfo {
+        lastPage
+      }
+    }
+  }
+`);
 
 const NotificationPolling_Query = graphql(/* GraphQL */ `
   query NotificationPolling($where: NotificationFilterInput) {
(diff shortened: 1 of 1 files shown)


apps/web/src/components/NotificationList/NotificationList.tsx
```diff
@@ -11,6 +12,16 @@ import NotificationListPage from "./NotificationListPage";
 import NotificationItem from "./NotificationItem";
 import NotificationPortal from "./NotificationPortal";
 
+const MaxNotifcationPages_Query = graphql(/* GraphQL */ `
```
mnigh, line 15:
```suggestion
const MaxNotificationPages_Query = graphql(/* GraphQL */ `
```

apps/web/src/components/NotificationList/NotificationList.tsx
```diff
@@ -41,6 +52,14 @@ const NotificationList = ({
 }: NotificationListProps) => {
   const now = nowUTCDateTime();
   const [searchParams] = useSearchParams();
+  const onlyUnread =
+    searchParams.has("unread") && searchParams.get("unread") !== null;
+  const [{ data: maxPagesData }] = useQuery({
+    query: MaxNotifcationPages_Query,
```
mnigh, line 58:
```suggestion
    query: MaxNotificationPages_Query,
```

mnigh (changes requested):

mnigh (approved):
LGTM

PR #11036, opened 2024-07-23
[Fix] Adds `workingLanguageAbility` column

### LINKED ISSUE #10888: 🐛 Pool Candidate table `Languages` filter is confusing
## 🐛 Bug

When filtering by Languages, there is no Languages column rendered. 

> [!NOTE]  
> Although the Languages filter is not the using the same field as Preferred Communication Language, it is not clear to a user who wouldn't know this from the UI alone.

## 🦋 Expected Behaviour

A Languages column is rendered so that a user can observe the filtered values and avoid confusion with another language field, Preferred Communication Language.

## 📋 Steps to Reproduce

1. Navigate to pool candidates table
2. Filter by Languages French only
3. Observe Preferred Communication Language values in table with English

## 📸 Screenshots

Screen Shot 2024-07-04 at 12 24 42
Screen Shot 2024-07-04 at 12 23 56

## ✅ Acceptance Criteria

- [ ] Working language ability column is rendered on Pool Candidate tables by default beside Preferred Communication Language
- [ ] Languages field label changed to Working language ability in Pool Candidate table filter

mnigh:
🤖 Resolves #10888.

## 👋 Introduction

This PR adds changes the pool candidate and user table (not specifically part of the issue's scope but the same problem) filter dialog labels for **Languages** to **Working language ability** to bring it closer in sync with the field that it filters by. It also adds the **Working language ability** column to the aforementioned tables so that a user can see the result of filtering in the table. Finally, it consolidates some strings that were referring to the same thing and did not need to be separate. 

## 🧪 Testing

1. `pnpm build`
2. Navigate to pool candidates table
3. Filter by **Working language ability**
4. Verify results are as expected
5. Navigate to users table
3. Filter by **Working language ability**
4. Verify results are as expected

## 📸 Screenshots

Screen Shot 2024-07-23 at 16 18 19
Screen Shot 2024-07-23 at 16 17 39
Screen Shot 2024-07-23 at 16 25 28


apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
```diff
@@ -844,6 +844,29 @@ const PoolCandidatesTable = ({
...
+        if (lookingForBilingual) {
+          arr.push(intl.formatMessage(commonMessages.bilingualEnglishFrench));
+        }
+        return arr.join(", ");
+      },
+      {
+        id: "languageAbility",
+        header: intl.formatMessage(commonMessages.workingLanguageAbility),
+      },
+    ),
```
yonikid15, line 869: Could we pull this out into a utility file that can be used in both the PoolCandidatesTable and UserTable?

yonikid15 (approved):
Nice work Matt :+1:

yonikid15 (approved):

PR #11037, opened 2024-07-23
[Debt] Constrain types PoolSkill and UserSkill

vd1992:
🤖 Progresses #10799

## 👋 Introduction

Chips away at the issue by targeting types
`PoolSkill` 
`UserSkill`

## 🕵️ Details

Some of this was laying groundwork for later, like more fragments on the ViewPoolCandidate page to enable thorough shaking later. Some fields will need to be revisited once types `User` and `PoolCandidate` are handled as there is unneeded querying in places when something has to run through the generic type. 

I think I've had my fill of this issue for the time being 😛 

## 🧪 Testing

1.  All functionality related to the types still works
2.  Nothing left to change


apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/ViewPoolCandidatePage.tsx
vd1992, line 1: Aside from shaking out `userSkills`, I didn't remove fields intentionally. I added more fragments in so when it comes time to clear out type `PoolCandidate`, this query gets reduced massively.

esizer (approved):
Nice :sunglasses:

PR #11071, opened 2024-07-26
[Feat] Make candidate doc download async

### LINKED ISSUE #9692: ✨ Make Pool Candidate Word download async
## ✨ Feature

Let's convert the synchronous server-side generated PDF report to be run off-request asynchronously.

https://github.com/GCTC-NTGC/gc-digital-talent/pull/9606
https://github.com/GCTC-NTGC/gc-digital-talent/pull/8757

## ✅ Acceptance Criteria

- [ ] Client can request a report to be generated, and they are told to wait for completion
- [ ] Server executes the query and generates the report in an asynchronous worker
- [ ] Client is notified the report is ready with an in-app notification
- [ ] Client is notified if the report generation failed with an in-app notification
- [ ] Client can download the file from the server
- [ ] Only the user who requested the report can download the file (or use signed, sharable links?)
- [ ] generated file should be localized 
- [ ] A scheduled job running overnight deletes any files over 24 hours old
- [ ] Toast message while processing the download: "Preparing your file for download. You will receive a notification when it is ready."
- [ ] Success Notification: "`Your file is ready for download: <file name>`."
- [ ] Failure Notification: "`There was a problem on our end. <filename> failed to download. If you continue to receive this error, please get in touch with our support team`."
- [ ] File should be available for 24 hours for download.
- [ ] Clean up will be happening every night and will clean the files more than 24 hours old.

## 🛑 Blockers

```[tasklist]
### Blocked By
- [ ] #9688
- [ ] #9685
- [ ] #9684
- [ ] #10634
- [ ] #9693
```

esizer:
🤖 Resolves #9692 

## 👋 Introduction

Updates the downloading of candidate profile documents to be asynchronous. 

## 🕵️ Details

This contains a bit of refacoring of the setup work done for the CSV to make things a bit more flexible when adding new downloadable documents. The process should be a bit easier now.

```php
// Graphql Mutation Handler
{
  $user = Auth::user();
  // Probably do some auth checks here
  
  try {
  
    // Create the generator (this should be as thin as possible for serialization
    $generator = new SomeFileGenerator();
  
    // Dispatch job
    // This job accepts anything that implements FileGeneratorInterface
    GenerateUserFile::dispatch($generator, $user);
    return true; // So we can show success message
  } catch(\Exception $e) {
    return false;
  }
  
  return false;
}

// SomeFileGenerator.php
class SomeFileGenator implements FileGeneratorInterface {
   // Create the actual document to be saved
   public function generate(): self;
   
   // Save the document to disk
   public function save();
   
   // Gets the filename for saving, downloading, notification
   public function getFileName(): string;
}
```

## 🧪 Testing

1. Build `pnpm run dev:fresh`
2. Start queue `make queue-work`
4. Navigate to pool candidates table
5. Download some profile documents (not CSV)
6. Confirm you get notification with PDF download


api/app/Generators/PoolCandidateDocGenerator.php
petertgiles, line 1: It seems like there are a lot of differences in the titles of the new document.  Could you please double-check?
image

petertgiles (changes requested):
It's looking very cool.  I can't believe we're finally starting to use the async file capabilities!  I've done a first pass and flagged some stuff of check on.

api/app/Generators/PoolCandidateDocGenerator.php
esizer: This is a good catch. Some of them are blatantly incorrect and I'll work on fixing them. Some changed because I tried my best to re-use the same labels from the CSV. Though, I'm not sure how desired this is. :thinking:

petertgiles: Getting the profile doc and the CSV consolidated might be worth doing at some point but I'd suggest not trying to roll that into this PR.  :grin:

esizer: lol I agree but that actually makes this PR more work since we already have the CSV done with all the strings. Unless we go the other route and make the CSV match the DOC :stuck_out_tongue:

petertgiles: @Jerryescandon How do you feel about differing titles between the DOCX and CSV files?  Is that intentional?  Would you prefer one change to match the other?

infrastructure/conf/nginx-conf-deploy/default
```diff
@@ -120,7 +120,8 @@ server {
         include fastcgi_params;
         fastcgi_param SCRIPT_FILENAME /home/site/wwwroot/api/public/index.php;
     }
-    location ~ "^/api(/(.*))?$" {
+
+    location ^~ /api {
```
tristan-orourke, line 124: We shouldn't have any requests to _just_ `/api` right, it should always be `/api/something`? Let's change this to 
```suggestion
    location ^~ /api/ {
```

tristan-orourke: This should apply to both nginx files ;)

api/app/Generators/PoolCandidateDocGenerator.php
tristan-orourke: If the pdf labels are more thought out, let's use those, but keep a single value for both files.

infrastructure/conf/nginx-conf-deploy/default
esizer: Fixed in [update rule to require subpath](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11071/commits/05732d9865c5aaa06184404801fb4405cc5aa614)

api/app/Generators/PoolCandidateDocGenerator.php
esizer: Alright, I think I got them all in [update labels to match document](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11071/commits/751cd3ebf6c72840908f9c584288ba0bc5cfc494)

api/lang/fr/language.php
```diff
@@ -3,4 +3,5 @@
 return [
     'en' => 'Anglais',
     'fr' => 'Français',
+    'bilingual' => 'Billingual',
```
mnigh, line 6:
```suggestion
    'bilingual' => 'Bilingue',
```

api/app/Generators/CandidateProfileDoc.php
petertgiles, line 1: Looks like  "billingual" problem
image

petertgiles, line 1: The original document doesn't have all the field names bolded.
image

petertgiles, line 1: The work preferences section has some special copy that is missing, I think.
image

petertgiles, line 1: Missing English copy
image

petertgiles, line 1: Unlocalized date ranges.
image

petertgiles, line 1: Unlocalized string concatenation
image

petertgiles, line 1: Unlocalized skill levels
image

petertgiles (changes requested):
Coming along nicely!  A new batch of things to check on:

esizer:
@tristan-orourke I just noticed something... We have two different versions of this type of document. What I have done here is the pool candidates table version which is not the application but the user profile :grimacing: 

Since this ticket is already quite large and close to completion (hopefully :sweat_smile: ) Should I just create a ticket to move the actual application document over? I think we could just add some context arg to the mutation to conditionally render the appropriate sections.

api/app/Generators/CandidateProfileDoc.php
esizer: Not sure what happened here. I did move it but also updated the key. I think it is fixed but can you check to make sure?

esizer: Removed as part of [fix unlocalized items](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11071/commits/dbd4d2a6b13ea4d4cf9e1180e848c84476c5fcc8)

esizer: okay. I got the translation in [fix unlocalized items](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11071/commits/dbd4d2a6b13ea4d4cf9e1180e848c84476c5fcc8) but, I actually dont' think this is supposed to include questions :sweat_smile: Waiting to see where we land on that but either way, should be translated now :+1:

esizer: Should be fixed in [fix unlocalized items](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11071/commits/dbd4d2a6b13ea4d4cf9e1180e848c84476c5fcc8)

esizer: Should be fixed in [fix unlocalized items](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11071/commits/dbd4d2a6b13ea4d4cf9e1180e848c84476c5fcc8)

esizer: Thankfully Carbon made it easy to localize these :laughing: Fixed in [localize experience dates](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11071/commits/4f2d6ab03196bd3d9883cdb56714fb85fa83a63d)

tristan-orourke: I would say lets move the full strings to the backend, but separate from the main translation for the enum.

tristan-orourke:
> Since this ticket is already quite large and close to completion (hopefully 😅 ) Should I just create a ticket to move the actual application document over? I think we could just add some context arg to the mutation to conditionally render the appropriate sections.

@esizer I think your instinct to fix this in a different ticket is correct. Though User Profile files and Candidate files should probably get their own mutations. Is the main difference between them that the Candidate file should use snapshot data?

esizer:
> @esizer I think your instinct to fix this in a different ticket is correct. Though User Profile files and Candidate files should probably get their own mutations. Is the main difference between them that the Candidate file should use snapshot data?

Yeah, I think you are right about that. Alright, I can work on creating a ticket for that. Thanks :+1:

api/app/Generators/CandidateProfileDoc.php
petertgiles: Oops, not quite.  :laughing: 
image

petertgiles (changes requested):
Few more small things...

api/app/Generators/CandidateProfileDoc.php
esizer: Alright, the longer versions have been moved to the backend in [add longer version strings to server](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11071/commits/5d651c86392657c980eccb784cca26723e1a13cd)

api/app/Generators/PoolCandidateCsvGenerator.php
```diff
@@ -11,20 +11,15 @@
...
-class CandidateProfileCsv extends CsvGenerator
+class PoolCandidateCsvGenerator extends CsvGenerator implements FileGeneratorInterface
 {
-    protected array $ids;
-
-    protected string $lang;
-
     protected array $generatedHeaders = [
         'general_questions' => [],
         'skill_details' => [],
```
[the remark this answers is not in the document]
esizer: Good catch! Updated in [fix letter casing](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11071/commits/3968efbb2b7e7240dad40ff3813d3bb445dcfb16)

api/app/Generators/DocGenerator.php
```diff
@@ -2,43 +2,58 @@
...
-abstract class DocGenerator extends FileGenerator
+abstract class DocGenerator extends FileGenerator implements FileGeneratorInterface
 {
-    protected PhpWord $doc;
+    protected ?PhpWord $doc;
 
     protected array $strong;
 
-    public function __construct()
+    public function __construct(public string $fileName, protected ?string $dir)
```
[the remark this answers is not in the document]
esizer: Okay, they are now localized after [localize filenames](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11071/commits/60f48eb83564340dc43265fe54f1a343d8a0ea7a)

api/app/Generators/PoolCandidateDocGenerator.php
```diff
@@ -0,0 +1,333 @@
...
+            if ($user->looking_for_english) {
+                $section->addListItem($this->localize('language.en'));
+            }
+
+            if ($user->looking_for_french) {
+                $section->addListItem($this->localize('language.fr'));
+            }
+
+            if ($user->looking_for_bilingual) {
+                $section->addListItem($this->localize('common.billingual'));
```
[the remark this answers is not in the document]
esizer: oh oops, I accidentally staged this fix in [add longer version strings to server](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11071/commits/5d651c86392657c980eccb784cca26723e1a13cd). Should be good now though :sweat_smile:

petertgiles (approved):
Looks great!  Nothing else from me.  :clap:

api/app/Models/User.php
```diff
@@ -1,4 +1,4 @@
-<?php
```
petertgiles, line 1: Accidental change?

esizer: I just ran pint and put my confidence in it... guess my confidence was misplaced :shrug:

PR #11073, opened 2024-07-27
[Fix] GraphQL response content type

### LINKED ISSUE #10917: 🐛 Vite Watch Mode - `ProfileAndApplicationsApplicant` blocked
## 🐛 Bug

For some reason, the ProfileAndApplicationsApplicant query is CORS-blocked when using Vite watch mode.  The `Access-Control` headers are missing and the content type is set to `text/html`.

## 📋 Steps to Reproduce

1. Start Vite watch mode
2. Navigate to the site running on port 3000
3. Log in
4. Attempt to view your profile

## 📸 Screenshot

image

## ✅ Acceptance Criteria

- [ ] Can view profile when using watch mode

### LINKED ISSUE #10980: 🐛 `@self` directive causes graphql to return as `text/html`
## 🐛 Bug

Use of the `@self` directive is causing the response to be `text/html` instead of the expected `application/json`.

## 🦋 Expected Behaviour

All graphql queries return `application/json` and not `text/html`.

## 🕵️ Details

I cannot tell why this is occurring but believe it to be the cause of #10917 but not 100% certain. Though, it seems to be the commonality in the queries that are having issues. You don't even need to query the field that uses the `@self` directive to get the HTML :face_with_spiral_eyes: 

Currently we only use the `@self` directive on two types (`User` and `PoolCandidateWithSkillCount`). It just so happens, queries that use those types are returning as `text/html` and being blocked by CORs in watch mode.

```sh
HTTP/1.1 200 OK
Content-Type: text/html; charset=UTF-8
```

## 📋 Steps to Reproduce

1. Login as an admin `admin@test.com`
2. Open the network tab showing fetch tab
3. Navigate to your dashboard `/applicant`
4. Notice one query is HTML and not JSON
5. Navigate to `/admin/pool-candidates` 
6. Again, notice the HTML content type

## 📸 Screenshot

screenshot_12-Jul-2024_19-18-1720826337

## 🙋‍♀️ Proposed Solution

We only use this directive in two places so I'm wondering if we could maybe avoid it entirely and find a different way of doing the same thing? :thinking: 

#### Method

We could try the [`@method` directive](https://lighthouse-php.com/6/api-reference/directives.html#method) instead of a scope? :thinking: 

The thing is, the sorting may be tricky since we wouldn't have the query available in the method so we may need to also have 

```graphql
type PoolCandidate {
  skillCount(skills: [UUID!]) @method
}
```

```php
class PoolCandidate
{
  public function skillCount(array $skills)
  {
     return $this->user->whereHas('userSkills', function ($query) use($skills) {
        $query->whereIn('skill_id', $skills);
     })->count();
  }
}
```

#### Resolver

We could also just split this query off since skill count is sort niche to the pool candidate table and not used anywhere else.

### User

This is so we can split out the user auth info. Maybe instead of using the `@self` we just use a custom attribute? We could keep the sub type in graphql but instead of using self we could just do:

```graphql
type User {
  authInfo: UserAuthInfo
}
```

```php
class User
{
    public function authInfo()
   {
      return new Attribute(
          get: function () {
            if(!Auth::user()->isAbleTo("view", $this)) return null;
            
             return [
                'sub' => $this->sub,
                /...
             ];
          }
      )
   }
}
```

## ✅ Acceptance Criteria

- [ ] GraphQL queries/mutations only return as `application/json`

esizer:
🤖 Resolves #10980, resolves #10917 

## 👋 Introduction

Fixes an error where graphql could sometimes return a content type of `text/html`

## 🕵️ Details

I initially thought that this had to do with `@self` since the only queries causing an issue were ones that used that directive. However, I started building those queries out adding fields one at a time until I hit the unexpected content type.

Turns out they had a sneaky thing in common, they both use `ArmedForcesStatus` enum (via the user type). After looking into it, I noticed an empty line at the start of the lang file. Removing this fixed the problem. 

i'm not sure how we can catch this in the future since it technically isn't incorrect :face_with_spiral_eyes: 

## 🧪 Testing

2. Navigate to any page that queries for armed forces of assessment decision
3. Confirm the query returns with `Content-Type application/json`
4. Confirm no white space at beginning of `api/lang/**` files

### DIFF
--- a/api/lang/fr/armed_forces_status.php
+++ b/api/lang/fr/armed_forces_status.php
@@ -1,4 +1,3 @@
-
 <?php
 
 return [
--- a/api/lang/fr/assessment_decision.php
+++ b/api/lang/fr/assessment_decision.php
@@ -1,4 +1,3 @@
-
 <?php
 
 return [


petertgiles (approved):
Weird bug, but good sleuthing!  :detective:

esizer:
> Weird bug, but good sleuthing! 🕵️

Yeah, it was a head scratcher :face_with_spiral_eyes:

tristan-orourke:
Wow, weird. Glad we get to keep `@self`! 😆

PR #11076, opened 2024-07-29
[Fix] Mobile layout for step tracker filters

### LINKED ISSUE #11039: 🐛 Screening and assessment switches do not wrap on smaller screen sizes
## 🐛 Bug

The switches we have for assessment status on the screening and assessment page are not wrapping on smaller screen sizes and causes side scrollling.

## 🦋 Expected Behaviour

Switches wrap to avoid side scrolling.

## 📋 Steps to Reproduce

1. Login as admin
2. Navigate to a pools screening and assessment page
3. Reduce the screen size
4. Observe the switches remaining on the same line causing side scrolling

## 🙋‍♀️ Proposed Solution

I believe the group of switches should not be on the same line as the search bar at a certain width and then each switch should be on its own line when needed as well.

## ✅ Acceptance Criteria

- [ ] No side scrolling on smaller screen sizes due to switches
- [ ] Layout gracefully responds to screen size changes

esizer:
🤖 Resolves #11039 

## 👋 Introduction

Fixes the mobile layout for the switches on the assessment step tracker.

## 🧪 Testing

3. Navigate to a pool's assessment step tracker
4. Reduce window width to a phone size
5. Confirm the filters are all visible without need to side scroll

### DIFF
--- a/apps/web/src/components/AssessmentStepTracker/Filters.tsx
+++ b/apps/web/src/components/AssessmentStepTracker/Filters.tsx
@@ -49,8 +49,10 @@ const Filters = ({
       <form onSubmit={methods.handleSubmit(onFiltersChange)}>
         <div
           data-h2-display="base(flex)"
-          data-h2-align-items="base(flex-end)"
+          data-h2-flex-direction="base(column) p-tablet(row)"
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
Looks great, nice fix!

PR #11081, opened 2024-07-29
[Debt] Candidate status attribute

### LINKED ISSUE #11032: ♻️ Remove pool candidate status attribute
## ♻️ Debt/Refactor

We used to have an attribute getter for pool candidate status but no longer use it so we want to remove it.

## 🙋‍♀️ Proposed Solution

Delete the following code:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/42020d30b49b634c8fd0fdb29d0dff7953b78734/api/app/Models/PoolCandidate.php#L587-L615

## ✅ Acceptance Criteria

- [ ] Pool candidate status attribute getter removed
- [ ] Pool candidate status still returns the expected value (what is in the DB)

esizer:
🤖 Resolves #11032 

## 👋 Introduction

Removes the candidate status attribute that only contained commented out code and returned the unaltered attribute.

## 🧪 Testing

1. Confirm pool candidate status still returns the expected result

### DIFF
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -584,36 +584,6 @@ class PoolCandidate extends Model
         return $query;
     }
 
-    /* accessor to obtain pool candidate status, additional logic exists to override database field sometimes*/
-    // pool_candidate_status database value passed into accessor as an argument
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
Seems good.  :+1:

PR #11083, opened 2024-07-30
[Tooling] Type check on build

### LINKED ISSUE #10907: ✨ Add type checking to build workflow
## ✨ Feature

Right now, we do not do any type checking on build. I'd like to have that in place to prevent mistakes.

## 🕵️ Details

When we build our project, it only transpiles TypeScript files but does not perform type checks. You can do this before build by running `pnpm run tsc` and it will get caught in the pipeline but, it would be nice to get ahead of this and not need to remember to run it.

Add `pnpm run tsc` to the build + dev command in `apps/web` :thinking: Another option could be adding the [`vite-plugin-checker`](https://vite-plugin-checker.netlify.app/) plugin.

Because it can slow performance, we don't need to run it on `dev` or `watch` workflows.

## ✅ Acceptance Criteria

- [ ] Typechecks are ran as part of the build workflow automatically
- [ ] Type errors are caught by github actions

esizer:
🤖 Resolves #10907 

## 👋 Introduction

Adds `tsc` as a dependency to the `build` command so we type check on builds and fail if types are not correct.

## 🧪 Testing

1. Run build `pnpm run build:fresh`
2. Confirm tsc is run
3. Force a type error
4. Re-run build `pnpm run build:fresh`
5. Confirm build fails

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   ],
   "scripts": {
     "preinstall": "npx only-allow pnpm",
-    "build": "h2-build && turbo run build --filter @gc-digital-talent/web",
+    "build": "h2-build && turbo run build --filter @gc-digital-talent/web && turbo run tsc",
     "build:fresh": "h2-build && turbo run build --filter @gc-digital-talent/web --force",
     "build:css": "h2-build",
     "codegen": "turbo run codegen --filter @gc-digital-talent/graphql",
@@ -29,6 +29,7 @@
     "intl-extract": "turbo run intl-extract",
     "intl-compile": "turbo run intl-compile",
     "lint": "turbo run lint",
+    "lint:fix": "turbo run lint -- --fix",
     "storybook": "pnpm --filter @gc-digital-talent/web run storybook",
     "storybook:design-system": "pnpm --filter @gc-digital-talent/web run storybook:design-system",
     "storybook:web": "pnpm --filter @gc-digital-talent/web run storybook:web",
--- a/turbo.json
+++ b/turbo.json
@@ -8,6 +8,7 @@
   "tasks": {
     "build": {
       "dependsOn": [
+        "tsc",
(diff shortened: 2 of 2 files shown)


petertgiles:
I'm not sure about this solution.  Tagging it onto the end build seems muddy, since it's not actually building anything for us.  This seems like it would make more sense in test or lint?

Also, the test instructions say to run build:fresh but this is actually added to build, right?

esizer:
> I'm not sure about this solution. Tagging it onto the end build seems muddy, since it's not actually building anything for us. This seems like it would make more sense in test or lint?
> 
> Also, the test instructions say to run build:fresh but this is actually added to build, right?

I agree and, we actually already do run this as part of the lint pipeline. Maybe it is unnecessary? 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/a262328a50249fe4e5728eab0bb2ae67665f86cb/.github/workflows/lint.yml#L70-L74

petertgiles:
> Maybe it is unnecessary?

I kind of think so...  If you've ignored CI errors and want to build anyway, I don't think it makes sense to artificially force a failure after a successful build.  I'm open to be convinced otherwise, though.

vd1992:
> I kind of think so... If you've ignored CI errors and want to build anyway, I don't think it makes sense to artificially force a failure after a successful build. I'm open to be convinced otherwise, though.

I'm fine with not adding it to keep build faster 😎

esizer:
> > Maybe it is unnecessary?
> 
> I kind of think so... If you've ignored CI errors and want to build anyway, I don't think it makes sense to artificially force a failure after a successful build. I'm open to be convinced otherwise, though.

I agree, I'll let this sit to see if anyone feels strongly in the other direction but if not I think we are safe to just close this out as not planned.

PR #11084, opened 2024-07-30
[Debt] Reduce queries on view pool candidate page

### LINKED ISSUE #10987: ♻️ Reduce queries on view pool candidate page
## ♻️ Debt/Refactor

Our view pool candidate page is running an uncessary amount of queries and we would like to try and reduce that number.

## 🕵️ Details

See: https://github.com/GCTC-NTGC/gc-digital-talent/pull/10968#issuecomment-2229036362

Essentially, now that we have moved our enum strings to the server we increased the number of queries. We kind of just queried them in place when moving them there instead of attempting to surface them to the page level to keep the ticket scope from exploding. 

Now, we have the joy of going back and fine tuning it a bit more.

## 🙋‍♀️ Proposed Solution

Bubble the queries up as fragments so that we do the fewest number needed at the page level if possible.

## ✅ Acceptance Criteria

- [ ] Number of queries being called on the view pool candidate page has been reduced. In particular, enum strings don't each trigger their own network queries.

esizer:
🤖 Resolves #10987 

## 👋 Introduction

Moves enum queries in dialogs do they are only run once the dialog is open.

## 🕵️ Details

Seems like a lot of these queries were resolved already so I just cleaned up the remaining two.

## 🧪 Testing

1. Build `pnpm run dev:fresh`
2. Login as an admin `admin@test.com`
3. Navigate to a pool candidate
4. Confirm only one query is being made to get page content
5. Confirm the affected dialogs still function

### DIFF
--- a/apps/web/src/components/CandidateDialog/ChangeStatusDialog.tsx
+++ b/apps/web/src/components/CandidateDialog/ChangeStatusDialog.tsx
@@ -380,11 +380,11 @@ const ChangeStatusDialog = ({
                   {fetching
                     ? intl.formatMessage(commonMessages.saving)
                     : intl.formatMessage({
-                      defaultMessage: "Change status",
-                      id: "iuve97",
-                      description:
-                        "Confirmation button for change status dialog",
-                    })}
+                        defaultMessage: "Change status",
+                        id: "iuve97",
+                        description:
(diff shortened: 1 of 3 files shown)


petertgiles (approved):
Looks good.  Nice tidying, too.  :broom:

PR #11090, opened 2024-07-30
[Bugfix] Filter for `displayName` in teams dropdown 

### LINKED ISSUE #11080: 🐛 Teams dropdown throws in bunch of "Not available" options
## 🐛 Bug

While creating a process the Teams dropdown opens up like this 
Screenshot 2024-07-29 153433

## 🦋 Expected Behaviour

Team dropdown should show only available teams. This is happening only on UAT and Prod 

## 🕵️ Details
The query involved is "CreatePoolPage" 
The teams options are pulled from me.authInfo.RoleAssignments and the query returns the results like below as over 100 Roleassignments and all of them are having null value in DisplayName except the DCM team. 
Hence it displays "Not available" for all others
image

Screenshot 2024-07-29 153943

## 📋 Steps to Reproduce
I see this is happening in UAT as well as prod. Something going on here. 

1. Goto create process page in UAT or prod if have admin access 
2. Check out Teams dropdown

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] Teams dropdown should only display the teams that the user is part of as the process operator.

vd1992:
🤖 Resolves #11080

## 👋 Introduction

Filter for teams with a truthy `displayName` and then render them on the create process page. 

## 🧪 Testing

1. Ensure you have teams lacking a name first
2. Then ensure they are not rendered in the dropdown

## 📸 Screenshot

Off main

image

Off branch

image

### DIFF
--- a/apps/web/src/pages/Pools/CreatePoolPage/CreatePoolPage.tsx
+++ b/apps/web/src/pages/Pools/CreatePoolPage/CreatePoolPage.tsx
@@ -361,6 +361,7 @@ const CreatePoolPage = () => {
   const roleAssignments =
     unpackMaybes(data?.me?.authInfo?.roleAssignments) ?? [];
   const teamsArray = roleAssignmentsToTeams(roleAssignments);
+  const teamsArrayFiltered = teamsArray.filter((team) => !!team.displayName); // filter out lacking a display name as a proxy for new teams TODO #10368
 
   const [, executeMutation] = useMutation(CreatePoolPage_Mutation);
   const handleCreatePool = (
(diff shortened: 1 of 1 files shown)


esizer (approved):
LGTM! :+1:

brindasasi:
This is working great! Find another place happening like this. Should we fix it there as well or a sep issue? This is happening in admin/Teams. 
Screenshot 2024-07-31 at 10 56 49 AM

vd1992:
> This is working great! Find another place happening like this. Should we fix it there as well or a sep issue? This is happening in Edit user page. Below is in local 

I think those should be a separate issue as we might need to think about how we want to filter that out. If people somehow have new teamable roles then we probably want to show it anyways 🤔

brindasasi (approved):
LGTM!

brindasasi:
I put them into https://github.com/GCTC-NTGC/gc-digital-talent/issues/11099 so that a decision can be made. Thanks ! :)

PR #11095, opened 2024-07-31
[Tooling] Improve eslint performance

### LINKED ISSUE #11094: 🛠️ Improve eslint performance
## 🛠️ Tooling

ESlint seems to have decreased in performance and is hampering dev it seems. We would like to improve this.

## 🙋‍♀️ Proposed Solution

Implement some of the suggestions:

- https://typescript-eslint.io/troubleshooting/typed-linting/performance/

## ✅ Acceptance Criteria

- [ ] Eslint shows a noticeable improvement in run time
- [ ] We are happy with the result from any changes that occur to rules

esizer:
🤖 Resolves #11094 

## 👋 Introduction

Makes some changes to the eslint config to improve performance.

#### Local

- **Before**: ~1m13s
- **Now**: ~31s

```sh
 Tasks:    18 successful, 18 total
Cached:    0 cached, 18 total
  Time:    31.337s
```
#### CI

- **Before**: ~[2m31s](https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/10168167753/job/28122162999#step:10:1)
- **After**: ~[2m4s](https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/10172227019/job/28134481643?pr=11095#step:10:1)

### :question:  What does this do?

- Ignores generated graphql types which were flagged as taking ~800ms to parse
- Removes the eslint prettier plugin, I hope devs have prettier run on file save?
  - This duplicates the eslint run and is not necessary if we have `formatOnSave=true`
- Narrows the includes for `tsconfig.json` files to just `src` directories
  - We could narrow this further but things seem fine for now
- Narrows the TS eslint parser projects
- Moves some slow import rules to only run in CI
- Removes airbnb (we can add these back in if we want but they added some slow rules and conflicted with a lot of our other rules)
- Fix some rules that were missed before? :thinking: 
- Updates the vscode eslint working dir settings to work better

## 🧪 Testing

1. Run eslint, confirm it works and there is a noticeable speed improvement
2. Confirm code actions still work
3. Confirm no errors in eslint output
4. Confirm errors are still flagged as expected
5. Confirm prettier runs still


esizer:
> This is the only one I'm a little worried about. I'm not sure we should be relying on IDE settings for this. Can this be run only in CI with the other slow rules?

Actually, I forgot a step :sweat_smile: typescript eslint recommends [just running it as a separate task](https://typescript-eslint.io/troubleshooting/typed-linting/performance#eslint-plugin-prettier). Thanks for catching that, I'll go back in and add the task for `"prettier": "prettier --check"`.

Eslint is trying to [move away from formatting](https://eslint.org/blog/2023/10/deprecating-formatting-rules/) so separating the two now kind of makes sense...I think :sweat_smile:

packages/eslint-config-custom/index.js
```diff
@@ -1,24 +1,27 @@
...
     "plugin:import/warnings",
     "plugin:import/typescript",
-    "plugin:prettier/recommended",
     "prettier",
   ],
   ignorePatterns: [
     "index.js",
     ".eslintrc",
     ".eslintrc.cjs",
-    "vite.*.ts",
```
petertgiles, line 18: Did you mean to remove this?

turbo.json
```diff
@@ -154,6 +154,28 @@
...
     },
+    "lint:fix": {
+      "inputs": [
+        "../../packages/eslint-config-custom/index.js",
+        "../../packages/eslint-config-custom/react.js",
+        "**/*.{ts,tsx}",
+        "!node_modules"
+      ]
+    },
+    "prettier": {
```
petertgiles, line 165: Looks like we'll have a bunch of stuff that immediately starts failing when we merge this.
image

petertgiles (commented):
I like this! :1st_place_medal:   My eslint hasn't been too bad so I can't comment on speed but it all makes sense.  As you say, it's a bit controversial so we should probably wait to discuss it.

turbo.json
esizer: oh, yeah.... I guess we probably should ignore compiled files, eh? :sweat_smile:

packages/eslint-config-custom/index.js
esizer: Yeah, i'm not sure why I ignored linting our config files in the first place. Maybe because we were not linting webpack? :thinking: 

I can revert this change if we do actually want to ignore config files in linting though.

turbo.json
esizer: okay, updated to ignore any generated files and it seems to be passing now.

[update prettier ignore for generated files](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11095/commits/4bd5954bf13833a550cce4d572d903ab98732894)

petertgiles (approved):
Yay!  :fast_forward:

PR #11136, opened 2024-08-02
[Fix] Reduce unnecessary queries on screening and assessment page

### LINKED ISSUE #11135: 🐛 Assessment Tracker page issues tons of queries pulling down the site
## 🐛 Bug

Latest release Assessment tracker page has 223 requests while there are only 581 candidates. There must be some duplicate queries happening. 

## 🦋 Expected Behaviour

Assessment tracker page should be functioning properly and there shouldn't be any unnecessary queries 

## 📋 Steps to Reproduce

1. Have some few hundreds of candidates in a pool 
2. Open the assessment tracker page 

## 📸 Screenshot

Image

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] Assessment tracker page should be functioning properly and there shouldn't be any unnecessary queries

esizer:
🤖 Resolves #11135

## 👋 Introduction

Fixes and issue where we were running an unnecessary amount of queries on the screening and assessment page severely impacting performance.

## 🕵️ Details

We had a query at the top for pool candidates to determine the page size so we could run the correct number of queries in batches. However, this query had no filters applied to it at all, meaning it was running a count of all pool candidates, not just active ones on the current pool.

Because of this, the pager count was based on all candidates and not the ones we were looking for. 

This is applying the same filters to the top level query so it gets the proper page count.

## 🧪 Testing

> [!NOTE]
> You can also tweak the batch size for better testing.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/60a37565053b5ca25e5f8a34e76591c000270216/apps/web/src/pages/Pools/ScreeningAndEvaluationPage/ScreeningAndEvaluationPage.tsx#L65

1. Seed many, many candidates
2. Build `pnpm run dev`
3. login as admin `admin@test.com
4. Confirm number of candidates in a specific pool
5. Navigate to that pools screening and assessment page
6. Confirm the nummber of queries for candidates sent matches the expeted number based on `candidates.length / BATCH_SIZE`

### DIFF
--- a/apps/web/src/pages/Pools/ScreeningAndEvaluationPage/ScreeningAndEvaluationPage.tsx
+++ b/apps/web/src/pages/Pools/ScreeningAndEvaluationPage/ScreeningAndEvaluationPage.tsx
@@ -29,11 +29,11 @@ type RouteParams = {
 };
 
 const ScreeningAndEvaluation_PoolQuery = graphql(/* GraphQL */ `
-  query ScreeningAndEvaluation_Pools($poolId: UUID!, $first: Int!) {
+query ScreeningAndEvaluation_Pools($poolId: UUID!, $first: Int!, $where: PoolCandidateSearchInput) {
     pool(id: $poolId) {
       ...AssessmentStepTracker_Pool
     }
-    poolCandidatesPaginated(first: $first) {
+    poolCandidatesPaginated(first: $first, where: $where) {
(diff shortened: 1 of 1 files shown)


brindasasi (approved):
Working great! thanks for the quick fix as always :)

Issue #10609, opened 2024-06-06
✨ Give new roles access to Pool pages


tristan-orourke:
## ✨ Feature

Allow the new roles to view and work with the appropriate pools, and add a new tab where Process Operators can be assigned.

## 🕵️ Details

Community Admins, Community Recruiters and Process Operators can access the pools table. Ensure the authorizedToView scope works correctly for these new roles.

Community Admins, Community Recruiters and Process Operators can access the pool pages, including editing, assessment plan, viewing candidates, and assessing candidates.

There is also a new tab for each pool which lists Process Operators. Process Operators (and Platform Admins) can view this list, but only Community Admins and Community Recruiters can add or remove people.

Of the new roles, only Community Admins can publish pools, or edit published pools.

Process Operators can view and assess candidates, and can Qualify, Disqualify or Remove them, but cannot _Place_ candidates.

## 🎨 Design File

Include a link to the design file (if it exists).

### 🧑‍🎨 Designer

Include the author of the linked design file's name (or even better, their `@githubUsername`) for future reference.

## 📸 Screenshot

![image](https://github.com/user-attachments/assets/e6d596e0-00c1-4d25-9e11-d9984d298ffb)

![image](https://github.com/user-attachments/assets/a15d11a2-7e5d-4304-a245-5ecd41942d92)

## 🙋‍♀️ Proposed Implementation

A lot of this work will just involve updating policies so that the new roles can interact with the appropriate queries and mutations.

## 🌎 Localization

(optional) Provide any new copy along with translations available.

## ✅ Acceptance Criteria

- [ ] New roles (Community Admins, Community Recruiters and Process Operators) can access Pools table, edit pool page, assessment plan page, candidate assessment page, and individual candidate snapshot and candidate profile pages
- [ ] Search for `#10609` for places in the code base which need to be updated
- [ ] Pool and Candidate authorizedToView scope updated for new roles (Community Admins, Community Recruiters and Process Operators)
- [ ] New tab added for each pool which lists Process Operators. Process Operators (and Platform Admins) can view this list, but only Community Admins and Community Recruiters can add or remove people. "Add new member" button only appears for Community Admins and Recruiters.
- [ ] Only pool-based roles appear in the table. (eg if a user has Process Operator role AND community recruiter role, the recruiter role does not appear.)
- [ ] url for new tab is `/admin/pools/:id/manage-access`
- [ ] no filter dialog required for the membership table
[...]

Issue #10368, opened 2024-05-13
♻️ Remove Teams from UI


tristan-orourke:
## ♻️ Debt/Refactor

Remove Teams from all aspects of the UI.

## 🕵️ Details

- Remove Teams pages from admin UI
- Remove Teams links from admin dashboard and nav menu
- Remove Teams field from Create Pool form
  - also ensure its not required in DB and mutation
- Remove Team-based roles section from /admin Edit User page

## ✅ Acceptance Criteria

- [ ] Links to Teams pages do not appear in Nav and Dashboard
- [ ] Teams pages and components axed
- [ ] Teams field removed from Create Poster form (and poster can still be successfully created)
- [ ] Team not displayed on any pages associated with Job Poster
- [ ] Team-based roles section removed from Admin edit user page

tristan-orourke:
1. Remove UI that uses teams
2. Amend tests to use new roles instead of old ones
3. Actually remove data model

PR #11138, opened 2024-08-02
[Debt] Refactor authorizedtoview scopes

### LINKED ISSUE #11041: ♻️ Rewrite authorizedToView scopes so they fall-through to least permissive option
## ♻️ Debt/Refactor

Our authorizedToView scopes are written to add conditions to a query, and then return the query at the end. However, this means that if all of our conditionals are negative, the result is the least permissive query possible. It would be more difficult to make a mistake positive conditionals are required to return a non-empty query.

## 🕵️ Details

This is what we did for Policy methods a while back. Hopefully, it should make the logic flow of these scopes easier to understand and more difficult to mess up.

## 🙋‍♀️ Proposed Solution

Essentially, the final line of an authorizedToView query should be `return $query->where('id', null);`, which will force the query to return nothing. This in turn forces us to check for all relevant permissions and return early if the user has them. If we forget to check a permission, the scope will filter out more results, not fewer.

## ✅ Acceptance Criteria

- [ ] The final line of any `authorizedToView`-type scope is `return $query->where('id', null);`
- [ ] An `authorizedToView`-type scope returns all the items for which the matching policy method would return true.

petertgiles:
🤖 Resolves #11041

## 👋 Introduction

This branch refactors the AuthorizedToAction scopes to make them easier to maintain and safer.  It also adds tests to help document expected behaviour and guard against regression bugs.

## 🕵️ Details

The general strategy used here:
- Test for the most privileged conditions and return early if true
- Fall through to progressively less privileged conditions
- Don't test for permissions twice (hasSomePermission)
- [Type hint a void return](https://laravel.com/docs/10.x/eloquent#local-scopes) - Laravel will not use a different query if you return one

I also added some extra logging the privileged user checker.

I also split User::scopeAuthorizedToViewAsAdmin to (hopefully) make it a little clearer and easier to test.

I'm not sure which roles should be tested right now since we're in the middle of the communities feature.  I wrote tests for the "legacy" roles.  Hopefully they can be a good model for extending when the new roles become active.

I noticed two discrepancies from the role specs:
- Platform admins does not have `view-any-submittedApplication` in the spec but does in the seeder
- Request responder does not have `view-any-user` in the spec but does in the seeder

## 🧪 Testing

1. Rebuild app - no permissions changed so no need to run rolePermissionSeder.

- [ ] App behaves normally (especially access to notifications, pools, candidates, and users)
- [ ] Tests pass
- [ ] Assertions in new tests are correct


api/app/Checkers/ProtectedRequestUserChecker.php
```diff
@@ -56,6 +57,8 @@ public function currentUserHasRole(
         bool $requireAll = false
     ): bool {
         if (! $this->isSafeToUseRole($name)) {
+            Log::warning('Tried to unsafely use role '.json_encode($name));
```
petertgiles, line 60: This is not directly related to the AC - just that it would have saved me lots of time if it existed before.  :face_exhaling:

api/app/Models/Pool.php
```diff
@@ -541,43 +541,47 @@ public static function scopeClassifications(Builder $query, ?array $classificati
...
     /**
-     * Filter for pools the user is allowed to view admin information for, based around assessment plan permissions
+     * Filter for pools the user is allowed to admin, based on scopeAuthorizedToAdmin
      */
-    public static function scopeAuthorizedToViewAsAdmin(Builder $query, ?bool $canAdmin): Builder
+    public static function scopeCanAdmin(Builder $query, ?bool $canAdmin): Builder
     {
-        if (empty($canAdmin)) {
-            return $query;
+        if (! empty($canAdmin)) {
```
petertgiles, line 548: This maintains the existing behaviour - but maybe it should be `if ($canAdmin)` so it only runs if truthy?

api/tests/Unit/PoolCandidateAuthorizationScopeTest.php
```diff
@@ -0,0 +1,186 @@
...
+                User::factory()
+                    ->asCommunityManager()
+                    ->create()
+            );
+
+        $candidateIds = PoolCandidate::authorizedToView()->get()->pluck('id');
+        assertEqualsCanonicalizing([], $candidateIds->toArray());
+    }
+
+    // a platform admin should be able to view submitted candidates in any team
```
petertgiles, line 169: Technically, this goes against the spec.

api/tests/Unit/UserAuthorizationScopeTest.php
```diff
@@ -0,0 +1,263 @@
...
+
+        // can also see the user with an application submitted to team 1's pool
+        $userIds = User::authorizedToView()->get()->pluck('id');
+        assertEqualsCanonicalizing([
+            $poolOperator->id,
+            $this->user1->id,
+        ], $userIds->toArray());
+    }
+
+    // a request responder should be able to view any user
```
petertgiles, line 128: Technically, this goes against the spec.

petertgiles (commented):
Some notes.

tristan-orourke:
You're supposed to return void from scopes? Huh. And you can still chain them with other query builder methods?

petertgiles:
> You're supposed to return void from scopes? Huh. And you can still chain them with other query builder methods?

You _can_ return another query builder but I found out after much head-scratching that Lighthouse will not use it and wil stick with the original.  So probably safer not to do it at all.

esizer:
> > You're supposed to return void from scopes? Huh. And you can still chain them with other query builder methods?
> 
> You _can_ return another query builder but I found out after much head-scratching that Lighthouse will not use it and will stick with the original. So probably safer not to do it at all.

This is really good info to have :sweat_smile: Thanks for finding that out :+1:

api/app/Models/Pool.php
```diff
@@ -632,43 +636,45 @@ public function scopeOrderByPoolBookmarks(Builder $query, ?array $args): Builder
...
+            }
+        });
+        $filterCountAfter = count($query->getQuery()->wheres);
+        if ($filterCountAfter > $filterCountBefore) {
+            return;
         }
 
-        return $query;
+        // fall through - anyone can view a published pool
+        $query->where('published_at', '<=', Carbon::now()->toDateTimeString());
```
tristan-orourke, line 677: This is true, but it's already capturedby `view-any-publishedPool`, which is assigned to the baseUser role.

api/app/Models/Pool.php
```diff
@@ -632,43 +636,45 @@ public function scopeOrderByPoolBookmarks(Builder $query, ?array $args): Builder
...
+            }
 
-                return $query;
-            });
+            if ($user?->isAbleTo('view-any-publishedPool')) {
+                $query->orWhere('published_at', '<=', Carbon::now()->toDateTimeString());
+            }
+        });
+        $filterCountAfter = count($query->getQuery()->wheres);
+        if ($filterCountAfter > $filterCountBefore) {
```
tristan-orourke, line 672: Hmm. Won't this always be true, because of the `where` at line 651?

api/app/Models/User.php
```diff
@@ -1041,44 +1039,60 @@ public function scopeAuthorizedToViewSpecific(Builder $query)
...
+            return;
+        }
 
-        return $query;
+        // fall through - query will return nothing
+        $query->where('id', null);
     }
 
-    public function scopeAuthorizedToView(Builder $query)
+    public function scopeAuthorizedToView(Builder $query): void
```
tristan-orourke, line 1056: I don't remember why we separated this from `scopeAuthorizedToViewSpecific` and I don't see any advantage. Do you? 🤔

api/tests/Unit/PoolAuthorizationScopeTest.php
```diff
@@ -0,0 +1,289 @@
...
+            $this->poolPublished1->id,
+            $this->poolClosed1->id,
+            $this->poolArchived1->id,
+            $this->poolDraft2->id,
+            $this->poolPublished2->id,
+            $this->poolClosed2->id,
+            $this->poolArchived2->id,
+        ], $poolIds->toArray());
+    }
+}
```
tristan-orourke, line 289: Could you add a comment: `TODO: add tests for new roles as part of #10609`

api/tests/Unit/PoolCandidateAuthorizationScopeTest.php
tristan-orourke: Thanks, I've updated the spec in this case.

api/tests/Unit/PoolCandidateAuthorizationScopeTest.php
```diff
@@ -0,0 +1,186 @@
...
+            );
+
+        // there is one submitted candidate in each team's pool
+        $candidateIds = PoolCandidate::authorizedToView()->get()->pluck('id');
+        assertEqualsCanonicalizing([
+            $this->candidateSubmitted1B->id,
+            $this->candidateSubmitted2A->id,
+        ], $candidateIds->toArray());
+    }
+}
```
tristan-orourke, line 186: Could you add the reminder here as well? `TODO: add tests for new roles as part of #10609`

api/tests/Unit/UserAuthorizationScopeTest.php
```diff
@@ -0,0 +1,263 @@
...
+        $this->platformAdmin = User::factory()
+            ->asAdmin()
+            ->create();
+
+        $this->pool1 = Pool::factory()
+            ->for($this->platformAdmin)
+            ->published()
+            ->afterCreating(function (Pool $p) {
+                // normally this happens only when the process operator is added - might want to reconsider that logic
+                $p->team()->create(['name' => 'pool-'.$p->id]);
```
tristan-orourke, line 52: If you want to use this for the PoolOperator role, you actually want to use `$p->legacyTeam()`. The `team` relation going forward is used for the new ProcessOperator role. (This will be much less confusing after #10368.) Setting team_id, like you did in api/tests/Unit/PoolCandidateAuthorizationScopeTest.php, still works.

api/tests/Unit/UserAuthorizationScopeTest.php
```diff
@@ -0,0 +1,263 @@
...
+        assertEqualsCanonicalizing([
+            $this->user1->id,
+        ], $userIds->toArray());
+    }
+
+    // an pool operator should be able to view themselves and any users with submitted candidates in their pools
+    public function testViewAsPoolOperator(): void
+    {
+        $poolOperator = User::factory()
+            ->asPoolOperator($this->pool1->team->name)
```
tristan-orourke, line 115:
```suggestion
            ->asPoolOperator($this->pool1->legacyTeam->name)
```

api/tests/Unit/UserAuthorizationScopeTest.php
```diff
@@ -0,0 +1,263 @@
...
+        Auth::shouldReceive('user')
+            ->andReturn($this->platformAdmin);
+
+        $userIds = User::authorizedToViewBasicInfo()->get()->pluck('id');
+        assertEqualsCanonicalizing([
+            $this->platformAdmin->id,
+            $this->user1->id,
+            $this->user2->id,
+        ], $userIds->toArray());
+    }
```
tristan-orourke, line 262: Could you add the reminder here as well? `TODO: add tests for new roles as part of #10609`

api/app/Models/Pool.php
tristan-orourke: I guess not! I tested it in tinker. And `->where( function ($q) { ... } )` with an empty inner function doesn't actually increase the where count. Could you add a little comment mentioning that this has been considered and isn't a problem?

tristan-orourke (changes requested):
Looks good. The `filterCountAfter` check is a smart approach which avoids having to check permissions twice. I just have some minor questions and a note for a test.

api/tests/Unit/UserAuthorizationScopeTest.php
tristan-orourke: This one I'll scope out into another ticket.

api/app/Models/Pool.php
petertgiles: Oh, interesting!  I agree that if it confused us both it should change.
[clarify scopeCanAdmin test](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11138/commits/c81a8bc0e1d77913b9a145617ef493bf640dc555)

api/tests/Unit/UserAuthorizationScopeTest.php
petertgiles: [extra comments](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11138/commits/dbda192c713a308da0628b1501ca454a706f879a)

api/tests/Unit/PoolCandidateAuthorizationScopeTest.php
petertgiles: [extra comments](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11138/commits/dbda192c713a308da0628b1501ca454a706f879a)

api/tests/Unit/PoolAuthorizationScopeTest.php
petertgiles: [extra comments](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11138/commits/dbda192c713a308da0628b1501ca454a706f879a)

api/app/Models/Pool.php
```diff
@@ -554,43 +554,47 @@ public static function scopeClassifications(Builder $query, ?array $classificati
...
-                    $query->orWhereHas('legacyTeam', function (Builder $query) use ($teamIds) {
-                        return $query->whereIn('id', $teamIds);
-                    });
-                } else {
-                    return $query->where('id', null); // when the user can't see any assessment plans
                 }
-            }
-            );
+
+                $query->orWhereHas('legacyTeam', function (Builder $query) use ($teamIds) {
```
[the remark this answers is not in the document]
petertgiles: [extra comments](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11138/commits/dbda192c713a308da0628b1501ca454a706f879a)

api/app/Models/Pool.php
petertgiles: Even guests have `view-any-publishedPool`.  But how can you call `isAbleTo` on a null user object?

api/app/Models/Pool.php
petertgiles: [extra comments](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11138/commits/dbda192c713a308da0628b1501ca454a706f879a)

api/app/Models/User.php
petertgiles: It is an odd pattern.  I think it's just that we factored out the bulk of the logic and one scope has a quick return for the "any-basicinfo" permission.  Maybe we could find a better name for `scopeAuthorizedToViewSpecific`?
image

api/app/Models/Pool.php
tristan-orourke: Nope, that's a great point! This is necessary to accommodate guests.

api/tests/Unit/UserAuthorizationScopeTest.php
petertgiles: [fix user legacy team check](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11138/commits/aebb9a4a9f36ba5e51c932ba1f8977a1713dd48d)

api/app/Models/User.php
tristan-orourke: I see. The difference is that `authorizedToViewSpecific` skips checking `view-any-user` permissions, allowing it to be called by both `authorizedToView` and `authorizedToViewBasicInfo` without any duplication.

However... this still seems pointless and confusing to me now. I suggest ditching `authorizedToViewSpecific`, moving that logic into `authorizedToView`, and just calling `authorizedToView` from `authorizedToViewBasicInfo`. Any objection?

api/tests/Unit/UserAuthorizationScopeTest.php
tristan-orourke: Good catch!

tristan-orourke (commented):
Thanks, you've addressed pretty much everything, I'd just push more that refactoring of the user authorizedToViewSpecific scope before merging.

api/app/Models/User.php
petertgiles: Yeah, good call.
[refactor scopeAuthorizedToViewSpecific](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11138/commits/0f123831f9eea3959af35a3859d67b88cbc02f6e)

tristan-orourke (approved):

Issue #10785, opened 2024-06-25
✨ [Epic] Manager Dashboard View


brindasasi:
## ✨ Feature

As a Manager I would like to have a dashboard view of the job advertisements and talent requests. 
Also I would like to have the access to tool kit that helps me to do hiring activities effectively from the dashboard itself.

## 🕵️ Details

Add any additional details that could assist with the development of the new feature.

## 🎨 Design File

https://www.figma.com/design/oZhQe9ahidHFfaiqgREvxa/Dashboard-(All-users)?node-id=254-12798&t=fnEO58s7Od7SBz9s-4

### 🧑‍🎨 Designer

Include the author of the linked design file's name (or even better, their `@githubUsername`) for future reference.

## 📸 Screenshot

![image](https://github.com/GCTC-NTGC/gc-digital-talent/assets/89482910/0c0617ca-ad02-47bb-82b4-a8ca793bf20e)

## 🙋‍♀️ Proposed Implementation

(optional) Do you have a proposed implementation?

## 🌎 Localization

(optional) Provide any new copy along with translations available.

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the feature was properly implemented.

- [ ] Criteria 1
- [ ] Criteria 2

## 🛑 Blockers

Issues which must be completed before this one.

```[tasklist]
### Blocked By
- [ ] ticket number
```

brindasasi:
Does Manager create & publish job poster or just create only ? @substrae 
(Currently community manager has the permission to approve)

substrae:
@brindasasi 

So couple things:
* Manager =/= Community manager, just in case that was unclear (I think we're already on the same page there, but better safe than sorry 😂)
* Managers would only have the ability to create drafts, share those drafts with other people, and then submit those drafts for publishing; for now, a submitted draft would then be reviewed by the relevant community team who would also be the ones to publish it

brindasasi:
Phase 1  is only Talent Request History. 
Phase 2 will include job advertisements

tristan-orourke:
I would say...
1. View all searches => /manager/requests (doesn't exist yet)
2. Start a new search => /search (might be moved to /manager/search when it becomes exclusive to managers?)
3. Find talent => /search
4. The magnify icons are buttons to open up a dialog
5. Manager profile => Probably /manager/profile, but doesn't exist yet
6. Account and privacy page => /applicant/settings (we're going to want to move this, to /account/settings?)
[...]

PR #11161, opened 2024-08-07
[Feature] Task card component

### LINKED ISSUE #11075: ⚙️ Task card component
## ⚙️ Summary

- Create a new `<TaskCard/>` component. This component acts as a wrapper for whatever content needed (list of accordions, text, etc). The card header contains an icon (optional), title (required), heading color (optional), and a link in the top right corner (optional). This is very similar to the `<Card />` component so try using that and incorporating and new properties as needed.

image

## 🎨 Design File

[Figma design](https://www.figma.com/design/oZhQe9ahidHFfaiqgREvxa/Dashboard-(All-users)?node-id=1010-18071&t=5sabmafb44VqvrwG-0)

### 🧑‍🎨 Designer

@substrae 

> [!NOTE]  

### ✅ Acceptance criteria

- [ ] Component matches design
- [ ] Properties are generic and only accept what's needed for the component
- [ ] Title is the correct heading size relative to page
- [ ] Link takes user to new page. If link is undefined, it should not render
- [ ] Heading color shows up correctly
- [ ] Component is added to Storybook
- [ ] NOTE: The `<Well />` displayed in the design is only a placeholder

## 🛠️ Properties

- title - string, required
- icon - IconType, optional
- link - ReactNode, optional
- color - ColorType, optional, default -> "primary"

petertgiles:
🤖 Resolves #11075 

## 👋 Introduction

This branch adds a new task card component.

## 🕵️ Details

## 🧪 Testing

1. ` pnpm run storybook:design-system`
2.  `localhost:6006/?path=/story/ui-src-components-taskcard--default`

- [ ] AC met
- [ ] light mode / dark mode
- [ ] standard mode / mobile mode

## 📸 Screenshot

image

### DIFF
--- a/packages/ui/src/components/TaskCard/TaskCard.tsx
+++ b/packages/ui/src/components/TaskCard/TaskCard.tsx
@@ -1,0 +1,49 @@
+import { ElementType, ReactNode } from "react";
+import { IconType } from "../../types";
+import { CardColor as CardFlatColor } from "../Card/CardFlat/types";
+import { LinkProps } from "../Link";
+
+type CardColor = Extract<CardFlatColor, "primary" | "secondary">;
+
+export interface TaskCardProps {
+  icon?: IconType;
+  title: ReactNode;
+  headingColor?: CardColor;
+  link: ElementType<LinkProps>;
+  children?: ReactNode;
+}
+
+const colorMap: Record<CardColor, Record<string, string>> = {
+  primary: {
+    "data-h2-background-color": "base:all(primary.lightest)",
+    "data-h2-color": "base:all(primary.darker)",
+  },
+  secondary: {
+    "data-h2-background-color": "base:all(secondary.lightest)",
+    "data-h2-color": "base:all(secondary.darker)",
+  },
+};
+
+const TaskCard = ({
+  icon,
+  title,
+  headingColor = "primary",
+  link,
+  children,
+}: TaskCardProps) => {
+  const Icon = icon;
+  return (
+    <>
+      <div {...colorMap[headingColor]}>
+        {Icon && (
+          <Icon data-h2-height="base(auto)" data-h2-width="base(x.85)" />
+        )}
+        {title}
+        {link}
+      </div>
+      <div>{children}</div>
+    </>
+  );
+};
+
+export default TaskCard;
--- a/packages/ui/src/components/TaskCard/TaskCard.tsx
+++ b/packages/ui/src/components/TaskCard/TaskCard.tsx
@@ -60,7 +60,7 @@ export interface TaskCardProps {
   icon?: IconType;
   title: ReactNode;
   headingColor?: CardColor;
-  link: ReactElement<LinkProps>;
+  link?: ReactElement<LinkProps>;
   headingAs?: HeadingLevel;
   children?: ReactNode;
 }
@@ -74,9 +74,11 @@ const TaskCard = ({
   children,
 }: TaskCardProps) => {
   // prepare link
-  const headingBarLink = cloneElement(rawLink, {
-    color: headingColor,
-  });
+  const headingBarLink = rawLink
+    ? cloneElement(rawLink, {
+        color: headingColor,
+      })
+    : null;
 
   // prepare icon element
   const Icon = icon;
@@ -111,7 +113,11 @@ const TaskCard = ({
           data-h2-align-items="base(center)"
         >
           {Icon ? (
-            <Icon data-h2-height="base(x1)" data-h2-width="base(auto)" />
+            <Icon
+              data-h2-visibility="base(hidden) p-tablet(visible)"
+              data-h2-height="base(x1)"
+              data-h2-width="base(auto)"
+            />
           ) : null}
           <HeadingTextElement {...headingTextStyles}>
             {title}
--- a/packages/ui/src/components/TaskCard/TaskCard.tsx
+++ b/packages/ui/src/components/TaskCard/TaskCard.tsx
@@ -1,29 +1,66 @@
-import { ElementType, ReactNode } from "react";
+import { ReactNode } from "react";
 
 import { IconType } from "../../types";
-import { CardColor as CardFlatColor } from "../Card/CardFlat/types";
-import { LinkProps } from "../Link";
+import Link from "../Link";
 
-type CardColor = Extract<CardFlatColor, "primary" | "secondary">;
+export const colorOptions = [
+  "primary",
+  "secondary",
+  "tertiary",
+  "quaternary",
+  "quinary",
+] as const;
+
+type CardColor = (typeof colorOptions)[number];
 
 export interface TaskCardProps {
   icon?: IconType;
   title: ReactNode;
   headingColor?: CardColor;
-  link: ElementType<LinkProps>;
+  linkText?: string;
+  linkHref?: string;
+  // TODO: allow <Link> as prop with changing text colour
   children?: ReactNode;
 }
 
-const colorMap: Record<CardColor, Record<string, string>> = {
+const headingStyleMap: Record<CardColor, Record<string, string>> = {
   primary: {
-    "data-h2-background-color": "base:all(primary.lightest)",
-    "data-h2-color": "base:all(primary.darkest)",
-    "data-h2-border-bottom": "base:all(primary.darkest)",
+    "data-h2-background-color": "base(primary.lightest)",
+    "data-h2-border-bottom": "base(primary.darkest)",
   },
   secondary: {
-    "data-h2-background-color": "base:all(secondary.lightest)",
-    "data-h2-color": "base:all(secondary.darkest)",
-    "data-h2-border-bottom": "base:all(secondary.darkest)",
+    "data-h2-background-color": "base(secondary.lightest)",
+    "data-h2-border-bottom": "base(secondary.darkest)",
+  },
+  tertiary: {
+    "data-h2-background-color": "base(tertiary.lightest)",
+    "data-h2-border-bottom": "base(tertiary.darkest)",
+  },
+  quaternary: {
+    "data-h2-background-color": "base(quaternary.lightest)",
+    "data-h2-border-bottom": "base(quaternary.darkest)",
+  },
+  quinary: {
+    "data-h2-background-color": "base(quinary.lightest)",
+    "data-h2-border-bottom": "base(quinary.darkest)",
+  },
+};
+
+const wrapperStyleMap: Record<CardColor, Record<string, string>> = {
+  primary: {
+    "data-h2-color": "base(primary.darkest)",
+  },
+  secondary: {
+    "data-h2-color": "base(secondary.darkest)",
+  },
+  tertiary: {
+    "data-h2-color": "base(tertiary.darkest)",
+  },
+  quaternary: {
+    "data-h2-color": "base(quaternary.darkest)",
+  },
+  quinary: {
+    "data-h2-color": "base(quinary.darkest)",
   },
 };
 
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/TaskCard/TaskCard.tsx
```diff
@@ -0,0 +1,135 @@
...
+  quinary: {
+    "data-h2-color": "base(quinary.darkest)",
+  },
+};
+
+export interface TaskCardProps {
+  icon?: IconType;
+  title: ReactNode;
+  headingColor?: CardColor;
+  link?: ReactElement<LinkProps>;
```
petertgiles, line 63: I noticed that Storybook flags an a11y issue on colour contrast of the link - but only when you switch modes.  I'm not sure if that's legitimate?  Also, I don't want to create a custom style exception for the links here.  That always causes pain when the standard styles move.  I'm not sure what to do here.

petertgiles: The standard link colour is <color>.darker while the design uses <color>.darkest.  The standard design reserves the darkest colour variant for hover.

petertgiles: Axe jest doesn't seem to mind.  :shrug:

yonikid15: Ya I'm also dealing with a color contrast issue on my component, which I'm not sure is legitimate either. Axe jest doesn't mind, as well as axe dev tools.

I checked the first elements contrast ratio using [WebAIM](https://webaim.org/resources/contrastchecker/) and it passed all the checks :shrug

image

yonikid15 (changes requested):
Nice work Peter, it's almost there!

This needs a bit of tweaking to match the mobile view: centre the heading and link within the parent div, and maybe .25 more padding (but that's up to @JoshBeveridge).

There isn't a design with the link in mobile design we got so you'll to ask Josh what the spacing should be.

packages/ui/src/components/TaskCard/TaskCard.tsx
yonikid15: Ya I'm also dealing with a color contrast issue on my component, which I'm not sure is legitimate either. Axe jest doesn't mind, as well as axe dev tools.

I checked the first elements contrast ratio using [WebAIM](https://webaim.org/resources/contrastchecker/) and it passed all the checks :shrug

image

petertgiles:
> This needs a bit of tweaking to match the mobile view: centre the heading and link within the parent div, and maybe .25 more padding

Good catch!  I've updated it.
image

packages/ui/src/components/TaskCard/TaskCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+        color: headingColor,
+      })
+    : null;
+
+  // prepare icon element
+  const Icon = icon;
+
+  // prepare heading text element
+  const HeadingTextElement = headingAs;
+  const commonHeadingTextStyles = { ...headingStyles["h4"] };
```
yonikid15, line 88: We should change the "h4" in `...headingStyles["h4"]` to the prop value `headingAs`.

packages/ui/src/components/TaskCard/TaskCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+  link: rawLink,
+  headingAs = "h3",
+  children,
+}: TaskCardProps) => {
+  // prepare link
+  const headingBarLink = rawLink
+    ? cloneElement(rawLink, {
+        color: headingColor,
+      })
+    : null;
```
yonikid15, line 81: This is an [old practice](https://react.dev/reference/react/cloneElement) of React. To achieve the same result maybe change the prop to `link?: { href: string; label: string}`. Then you can use the Link component here directly, which should prevent the link from visual inconsistencies.

```tsx
<Link mode="text" href={link?.href} color={headingColor}>
    {link?.label}
 </Link>
```

yonikid15 (changes requested):
My bad Peter, I found some small things that need to be updated

packages/ui/src/components/TaskCard/TaskCard.tsx
```diff
@@ -0,0 +1,138 @@
...
+      data-h2-border-radius="base(rounded)"
+      data-h2-background-color="base(foreground)"
+    >
+      {/* heading bar */}
+      <div
+        {...headingBarStyleMap[headingColor]}
+        data-h2-border-radius="base(rounded rounded 0 0)"
+        data-h2-border-bottom-width="base(1px)"
+        data-h2-border-bottom-style="base(solid)"
+        data-h2-padding="base(x1) p-tablet(x1 x1.5 x1 x1.5)"
```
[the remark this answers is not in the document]
petertgiles: Right, updated in the same way as the other PR.
[terse styles](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11161/commits/e8c9653537f7a797ee03e7facaab01d405cb3dcd)

packages/ui/src/components/TaskCard/TaskCard.tsx
petertgiles: I'm not sure about this one.  I actually do want it to have all the styles of an h4.  
image

Notice that semantic level is still headingAs.
image

yonikid15: Oh actually you can use the <Heading /> component instead. That way you don't need to import the styles and delete the margin. It would look something like:

```tsx
<Heading
  level={headingAs}
  size="h4"
  data-h2-margin="base(0)"
  data-h2-text-align="base(center) p-tablet(left)"
>
  {title}
</Heading>
```

petertgiles: Great idaea!
[use Heading component](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11161/commits/a1c5b895e83a0bd9c657f91e00eaecc256f0ce93)

yonikid15 (approved):
Excellent work :+1: Looks ready for a last design review from @JoshBeveridge then it's ready to merge

JoshBeveridge (changes requested):
Awesome! Sorry again for taking so long 😭

* If a link is present, it should use the horizontal space it needs to stay on a single line (see first screenshot)
* This might be a can of worms if you're using the `<Heading />` component, but the way the heading scales with longer text is pretty wonky (see first screenshot); the icon should always remain the same size and should have a consistent stroke width as determined by the heading, while the alignment of the icon should always remain vertically center with the first line of text
* The border separating the title bar from the card content is missing its color value (should be `color.darkest`)
* We can probably increase the gap between the title and the link on mobile devices to `x1`
* The interior padding of the card content should match the title bar (see second screenshot), but be careful here, because items inside the card based on the design require a divider that spans the full width - it might be easier to apply the padding to the items individually in this case (see third screenshot)

<details>
  <summary>Link span and icon size/alignment</summary>
  
</details> 

<details>
  <summary>Item padding</summary>
  
</details> 

<details>
  <summary>Items requiring full-width borders</summary>
  
</details>

petertgiles:
Thanks for the review!

>    If a link is present, it should use the horizontal space it needs to stay on a single line (see first screenshot)

Updated in [don't wrap links](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11161/commits/617646020dd573e2c44eef0f56eda6f84540f47c)

>    This might be a can of worms if you're using the <Heading /> component, but the way the heading scales with longer text is pretty wonky (see first screenshot); the icon should always remain the same size and should have a consistent stroke width as determined by the heading, while the alignment of the icon should always remain vertically center with the first line of text

OK, this customization won't be possible with the existing header component.  I've switched back to a customized header and included importing the icon styling from that component.  [switch back to custom heading](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11161/commits/bc3f4625d11b9830885c6c463f4f663e99311533)

>    The border separating the title bar from the card content is missing its color value (should be color.darkest)

Good catch! [fix border colour](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11161/commits/91c18dee65698cb2303b713592ed23d461091114)

>    We can probably increase the gap between the title and the link on mobile devices to x1

Update in [increase mobile heading gap](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11161/commits/02897bf7b515d7de7e6688338023b5aaa188c3ea)

>    The interior padding of the card content should match the title bar (see second screenshot), but be careful here, because items inside the card based on the design require a divider that spans the full width - it might be easier to apply the padding to the items individually in this case (see third screenshot)

This was actually a bug in the story so I fixed that in [fix storybook story](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11161/commits/68a57f62837d289303624dfb84b731c719871ba3).  However, this will probably be common enough that I added a lightweight wrapper for easy use.  [TaskCardItem component](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11161/commits/a56a4d861bb68c17a201d271e7aae258c7b1dfdd)

packages/ui/src/components/TaskCard/TaskCardItem.tsx
```diff
@@ -0,0 +1,18 @@
...
+const TaskCardItem = ({ children }: TaskCardItemProps) => {
+  return (
+    <div
+      data-h2-padding="base(x1) p-tablet(x1 x1.5)"
+      data-h2-border-bottom="base:selectors[:not(:last-child)](1px solid gray.lighter)"
+    >
+      {children}
+    </div>
+  );
+};
```
yonikid15, line 16: I don't think this is what Josh meant, cause this component is only a card. The heading padding should match the interior padding, so you only needed to add the padding to the `<div>{children}</div>`. However, looking at the screenshot of the #10785 Dashboard view, this would keep things consistent so I like it :+1: !

Sorry to be nit picky, but could we make this and TaskCard into a compound component instead? Similar to the  [DescriptionList](https://github.com/GCTC-NTGC/gc-digital-talent/blob/11075-task-card-component/packages/ui/src/components/DescriptionList/DescriptionList.tsx#L1-L115). You just need to add this into the TaskCard, and rename this to Item, and rename the TaskCard to Root. Then we can use it similar to our other components: `<TaskCard.Root /><TaskCard.Item>{...}</TaskCard.Item></TaskCard.Root>`

yonikid15 (changes requested):
Just two small changes that I suggested, and then this looks good to go!

packages/ui/src/components/TaskCard/TaskCard.tsx
```diff
@@ -0,0 +1,141 @@
...
+        data-h2-padding="base(x1) p-tablet(x1 x1.5)"
+        data-h2-display="base(flex)"
+        data-h2-flex-direction="base(column) p-tablet(row)"
+        data-h2-gap="base(x1) p-tablet(x2)"
+        data-h2-align-items="base(center)"
+      >
+        {/* wrapper */}
+        <div {...wrapperStyleMap[headingColor]} data-h2-flex-grow="base(2)">
+          <CustomHeading
+            {...headingStyles["h4"]}
```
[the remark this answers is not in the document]
petertgiles: Yeah, I guess it is possible.  It makes me cringe that it has to customized so much - at what point is it still worth trying to use the existing component?
```tsx
<Heading
  level={headingAs}
  size="h4"
  data-h2-margin="base(0)"
  data-h2-text-align="base(center) p-tablet(left)"
  data-h2-text-wrap="base(balance)"
  {...(Icon && {
    "data-h2-display": "base(flex)",
    "data-h2-align-items": "base(start)",
    "data-h2-gap": "base(0 x.5)",
  })}
  Icon={icon}
>
  {title}
</Heading>
```

I guess I don't like it any _less_ this way.  :laughing: 
[use heading component again](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11161/commits/163ed1ac085179ed39ae3954fbe7402f05c2e12e)

packages/ui/src/components/TaskCard/TaskCardItem.tsx
petertgiles: I like consistency.  If we end up using this card for non-list content we might have to undo this though.
[refactor to compound component](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11161/commits/7b4c5222782220a76328aba072e307297ad58d2c)

packages/ui/src/components/TaskCard/TaskCard.tsx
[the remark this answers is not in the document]
yonikid15: Well that's true you were using the headingStyles anyways, so it's basically the same thing :laughing:

petertgiles:
Yikes, I forgot about that.  Yeah, one of the big reasons I have to do a custom header is that the icon is supposed to disappear at mobile widths.  I rolled back to my custom header but to make it a bit more palatable I factored it out to its own subcomponent with some extra commenting.

yonikid15 (approved):
:1st_place_medal:

PR #11200, opened 2024-08-08
Resource block component

### LINKED ISSUE #11078: ⚙️ Resource block component
## ⚙️ Summary

- Create the `<ResourceBlock />` component. These blocks will be used to give the user easy access to other parts of the site, specifically parts of the profile. Additionally, the status icon will give a quick view on if that resource if complete or not.
-  Use the TaskCard component as a wrapper. You might need to add a prop to change position of the title in the heading. 
- The content includes three parts: a link to the resource, a status (check mark or warning icon) representing if it's complete or not, and a short description of resource.
- In the design you can see a block may contain more than one resource. Therefore, we can have a prop which accepts a list of resources or create a compound component (See Dialog or Accordion, which has something like a `<ResourceBlock.Root />` holding the root level props (title, color) and `<ResourceBlock.Item />` (link, description, status).

image

## 🎨 Design File

[Figma design](https://www.figma.com/design/oZhQe9ahidHFfaiqgREvxa/Dashboard-(All-users)?node-id=1010-18071&t=5sabmafb44VqvrwG-0)

### 🧑‍🎨 Designer

@substrae 

> [!NOTE]  

### ✅ Acceptance criteria

- [ ] Component matches design
- [ ] Properties are generic and only accept what's needed for the component
- [ ] Status shows the correct icon
- [ ] Uses the TaskCard component as a wrapper (if possible/not to difficult). Heading is centred.
- [ ] Whole block is clickable
- [ ] Ensure icon is just decorative

## 🛠️ Properties

- color- ColorType, optional, default -> "primary"
- title - string, required
- resource item props
  - link - string, required
  - status - string union, optional
    - eg. status?: "complete" | "incomplete"
  - description - string, required (?)

## Screenshots
image

petertgiles:
🤖 Resolves #11078

## 👋 Introduction

This branch adds the the resource block component.  It's made up of a Root component that acts as a host for a collection of Item components.

I ended up not using the the task card component for the host.  I would have to force it into "mobile mode" for all widths and force the link to disappear.  It's only about 30-40 lines of code that are duplicated.  If we get a third component that shares the look and feel maybe we can consider consolidating at that point.

## 🧪 Testing

1. ` pnpm run storybook:design-system`
2.  `http://localhost:6006/?path=/story/ui-src-components-resourceblock--default`

- [ ] AC met
- [ ] light mode / dark mode

## 📸 Screenshot

image


packages/ui/src/components/ResourceBlock/ResourceBlock.tsx
```diff
@@ -0,0 +1,194 @@
...
+  return null;
+};
+
+const Item = ({ link: rawLink, description, state }: ItemProps) => {
+  // prepare link
+  const link = rawLink
+    ? cloneElement(rawLink, {
+        color: "black",
+        // yuck, style exception 😞
+        "data-h2-font-weight": "base(bold)",
```
petertgiles, line 140: I'm not sure what to do about this.  The design has a bolded link but our standard links don't look like that.  Style exceptions tend to bite us when it comes time to make site-wide style changes.

yonikid15: [cloneElement](https://react.dev/reference/react/cloneElement) is an old react feature. For this, you can just grab the `link: { href: string; label: string; }` as a prop. Then just use the `<Link />` component we have directly. Then remove the link variable.

```tsx
 <Link
      href={link.href}
      data-h2-font-weight="base(bold)"
      data-h2-position="base:selectors[::after](absolute)"
      data-h2-content="base:selectors[::after](' ')"
      data-h2-inset="base:selectors[::after](0)"
      data-h2-justify-self="base(end)"
      color="black"
    >
      {link.label}
 </Link>
 ```

File: packages/ui/src/components/ResourceBlock/ResourceBlock.tsx [hunk 2]
```diff
@@ -0,0 +1,197 @@
...
+      data-h2-padding="base(x1)"
+      data-h2-display="base(flex)"
+      data-h2-flex-direction="base(column)"
+      data-h2-gap="base(x0.15)"
+      data-h2-border-bottom="base:selectors[:not(:last-child)](1px solid gray.lighter)"
+      // make the containing block for status icon
+      data-h2-position="base(relative)"
+      {...extraStateStyles}
+    >
+      {getStateIcon(state)}
```
yonikid15, line 173: `{getStateIcon(state)}` could be refactored to it's own component within this file: `<StateIcon state={state} />`.

packages/ui/src/components/ResourceBlock/ResourceBlock.tsx
```diff
@@ -0,0 +1,197 @@
...
+      data-h2-border-radius="base(rounded)"
+      data-h2-background-color="base(foreground)"
+    >
+      {/* heading bar */}
+      <div
+        {...headingBarStyleMap[headingColor]}
+        data-h2-border-radius="base(rounded rounded 0 0)"
+        data-h2-border-bottom-width="base(1px)"
+        data-h2-border-bottom-style="base(solid)"
+        data-h2-padding="base(x1) p-tablet(x1 x1.5 x1 x1.5)"
```
yonikid15, line 91: This can be shortened to:
```css
data-h2-border-bottom="base(1px solid)"
data-h2-padding="base(x1) p-tablet(x1 x1.5)"
```

yonikid15 (changes requested):
Nice work Peter, everything looks good on storybook. Just needs some refactoring in the component.

packages/ui/src/components/ResourceBlock/ResourceBlock.tsx
petertgiles: Interesting, I didn't know that function was not recommended.  I like the new solution better anyway.
[refactor link props](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11200/commits/c7d7cdb5a3b7f3ee66175b0b02d018721671b3f0)

packages/ui/src/components/ResourceBlock/ResourceBlock.tsx
petertgiles: Good idea!
[refactor state icon](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11200/commits/796512574726a40ef07027a78c68432fd56e7802)

packages/ui/src/components/ResourceBlock/ResourceBlock.tsx
petertgiles: Good catch.
[styles more terse](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11200/commits/d79b516d023d69616768a08aa8d28dfa6d4cab2e)

packages/ui/src/components/ResourceBlock/ResourceBlock.tsx
```diff
@@ -0,0 +1,196 @@
...
+        data-h2-text-align="base(center)"
+      >
+        {/* wrapper */}
+        <div {...wrapperStyleMap[headingColor]}>
+          <HeadingTextElement
+            {...commonHeadingTextStyles}
+            data-h2-text-align="base(center) p-tablet(left)"
+          >
+            {title}
+          </HeadingTextElement>
```
yonikid15, line 100: Oh actually you can use the <Heading /> component instead. That way you don't need to import the styles and delete the margin. It would look something like:

```tsx
<Heading
  level={headingAs}
  size="h4"
  data-h2-margin="base(0)"
  data-h2-text-align="base(center) p-tablet(left)"
>
  {title}
</Heading>
```

petertgiles: Oh yeah, that's way better!
[use heading component](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11200/commits/0a7247dfa09fdbbf0cdbefd61e0a9626b74b704b)

yonikid15 (approved):
Looks great Peter. Again, just needs final design review from @JoshBeveridge then it's good to go :rocket:

JoshBeveridge (changes requested):
Thanks @petertgiles! Sorry for the delay in getting back to you. Couple small notes:

* I'm assuming this card is going to be used by both the main dashboard tool component and the sidebar information cards, but the title alignment for those differ depending on where they are; the tool card will need an icon on desktop and be left aligned, while the info cards won't ever have an icon and will always be centre aligned
* The arrow icon next to item titles should be `x.75` in height/width (as opposed to `20px` hard-coded, this was my fault in the Figma file)
* The arrow icon should also be using the `arrow-small-right` icon from the `mini` series
* The status icon should be `x.75` in height/width (as opposed to `20px` hard-coded, same dealio)
* The status icons should also be using their respective `base` color in `light` mode, but use their `lighter` variant in `dark` mode
* In certain circumstances with longer titles, things tend to wrap weird on different device sizes; this also pushed the arrow icon into a really weird spot - ideally the icon would be inline with the text so that it always appears immediately after the last line (see screenshots below)
  * Related, but we should also make sure there aren't any instances where the status icon overlaps the title text (this can be accomplished with a bit of `right` padding on the title element
* The `gray.lighter` borders on items in the card should be using the `:all` modifier so that the color persists into `dark` mode
* Item subtext should be using the `caption` `font-size`
* Item padding should be `base(x1)` on mobile and then `l-tablet(x1.5 x1)` on larger screens; this also applies to the title - when left aligned, the title should always align with the text inside each item (see screenshots below)
* The border on the title block (the colored bit) is missing color (should be using `color.darkest`)
* Cards are missing their corner radius at the bottom

<details>
  <summary>Arrow alignment in titles</summary>
  
</details> 

<details>
  <summary>Item padding</summary>
  
</details>

petertgiles:
> Just noticed https://github.com/GCTC-NTGC/gc-digital-talent/pull/11161 has the tool card, so in that case, the title on this fella should always be centred

Updated in [force centered titles](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11200/commits/8499c219181374f6bf54c149621a728c6d5e506d)

> The arrow icon next to item titles should be x.75 in height/width (as opposed to 20px hard-coded, this was my fault in the Figma file)
The arrow icon should also be using the arrow-small-right icon from the mini series

Updated in [fix arrow icon](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11200/commits/0aa8279c5d9170b944a322089769bfb77b557b37).  Note that this icon is deprecated so we'll need to change it at some point.

> The status icon should be x.75 in height/width (as opposed to 20px hard-coded, same dealio)
The status icons should also be using their respective base color in light mode, but use their lighter variant in dark mode

Updated in [fix state icons](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11200/commits/e51543f4c77490f17f45491335d03fc4018730ea).

> In certain circumstances with longer titles, things tend to wrap weird on different device sizes; this also pushed the arrow icon into a really weird spot - ideally the icon would be inline with the text so that it always appears immediately after the last line (see screenshots below)

Updated in [inline arrow](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11200/commits/e048a581b4acd81583b80ea368885afe5fcc5668).

>    Related, but we should also make sure there aren't any instances where the status icon overlaps the title text (this can be accomplished with a bit of right padding on the title element)

Updated in [extra padding](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11200/commits/ed218d0aabb71d8cc9b3e2cfc727c99fff129a89)
image


> The gray.lighter borders on items in the card should be using the :all modifier so that the color persists into dark mode
Item subtext should be using the caption font-size
Item padding should be base(x1) on mobile and then l-tablet(x1.5 x1) on larger screens; this also applies to the title - when left aligned, the title should always align with the text inside each item (see screenshots below)

Updated in [fix item padding](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11200/commits/01106b9dc02ebdd6bfb88581af9a35c9e7db9452) and [fix item formatting](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11200/commits/5fa80a049dd005e77e075e78889c6e9b030c5fec).

> The border on the title block (the colored bit) is missing color (should be using color.darkest)

Thanks, fixed in [fix border colour](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11200/commits/df896c57a7d3b53588b2e5cdd7adcaeb1228e7da).

> Cards are missing their corner radius at the bottom

Good catch! [bottom rounded corners](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11200/commits/cac80f969eb13eea67bc63f646a430081596a119)

packages/ui/src/components/ResourceBlock/ResourceBlock.tsx
```diff
@@ -108,19 +108,27 @@ const Root = ({
 // an icon pinned to the top-right to show the completion state
 const StateIcon = ({ state }: { state: ItemProps["state"] }) => {
   const commonStyles: HydrogenAttributes = {
-    "data-h2-width": "base(20px)",
-    "data-h2-height": "base(20px)",
+    "data-h2-width": "base(x0.75)",
+    "data-h2-height": "base(x0.75)",
     "data-h2-position": "base(absolute)",
     "data-h2-location": "base(x0.75, x0.75, auto, auto)",
```
yonikid15, line 114: The status icon isn't lining up with the title, also since the status icon is absolute positioned the text can still overlap with icon with long text (even if you added extra padding). The [absolute positioned](https://developer.mozilla.org/en-US/docs/Web/CSS/position#absolute) icon is pulled out of the normal document flow, and will end up just stacking on top of the text.
image

I would recommend moving away from absolute positioning in this case (so that the text and status never overlap), and move the location of the status icon after the link and use flexbox to line then up, followed by the description text. I tried this below.

```tsx
    <div
      data-h2-display="base(flex)"
      data-h2-background="base(foreground)"
      data-h2-padding="base(x1) l-tablet(x1 x1.5)"
      data-h2-flex-direction="base(column)"
      data-h2-gap="base(x0.15)"
      data-h2-border-bottom="base:all:selectors[:not(:last-child)](1px solid gray.lighter)"
      data-h2-border-radius="base:all:selectors[:last-child](0 0 rounded rounded)"
      // make the containing block for state icon
      data-h2-position="base(relative)"
      {...extraStateStyles}
    >
      // ADD WRAPPING DIV AROUND LINK AND STATUS - Yoni
      <div
        data-h2-display="base(flex)"
        data-h2-align-items="base(flex-start)"
        data-h2-justify-content="base(space-between)"
      >
        <Link
          href={href}
          color="black"
          // yuck, style exception 😞
          data-h2-font-weight="base(bold)"
          // big click target black magic 🧙
          data-h2-position="base:selectors[::after](absolute)"
          data-h2-content="base:selectors[::after](' ')"
          data-h2-inset="base:selectors[::after](0)"
          data-h2-justify-self="base(end)"
          // icon extends margin + icons size: x0.75 + x0.75 = x1.5
          // item margin is base(x1) l-tablet(x1.5)
          data-h2-padding-right="base(x0.5) l-tablet(0)"
        >
          {title}
          {/* issue #11284 */}
          <ArrowSmallRightIcon // eslint-disable-line deprecation/deprecation
            data-h2-width="base(auto)"
            data-h2-height="base(x0.75)"
            data-h2-margin="base(x0.15)"
            data-h2-vertical-align="base(top)"
            aria-hidden
          />
        </Link>
        <StateIcon state={state} />
      </div> // END OF WRAPPING DIV
      <p data-h2-color="base(black.light)" data-h2-font-size="base(caption)">
        {description}
      </p>
    </div>
```
Then you can remove the absolute positioning in the Status component above.

```tsx
const commonStyles: HydrogenAttributes = {
    // Adding the min and max height was only way I found to keep the icon steady
    "data-h2-min-width": "base(x.75)",
    "data-h2-min-height": "base(x.75)",
    "data-h2-max-width": "base(x.75)",
    "data-h2-max-height": "base(x.75)",
  };

  if (state === "incomplete") {
    return (
      <ExclamationCircleIcon
        data-h2-color="base(error) base:dark(error.lighter)"
        {...commonStyles}
      />
    );
  }
  if (state === "complete") {
    return (
      <CheckCircleIcon
        data-h2-color="base(success) base:dark(success.lighter)"
        {...commonStyles}
      />
    );
  }
  return null;
```

yonikid15 (changes requested):
I found one thing with the text and status icon, but I hope I made it clear how to change it. I added the changes I made that worked for me.

packages/ui/src/components/ResourceBlock/ResourceBlock.tsx
petertgiles: I don't think the status icon is supposed to line up with the text.  If you check Figma you can see that they don't line up and the icon is positioned from the top-right corner of the card.
image

I added some right padding to the text line to ensure it wraps before overlapping with the icon.
image

yonikid15: Oh wow my bad :+1: It is pushed up into the center in the design.. however the padding isn't going to effect the icon since it's absolute positioned and is basically stacked on top of the other content in the div and would overlap it. I did test it thoroughly and it never overlapped the text, so I'm fine leaving it like this.

yonikid15 (commented):
Oh! Last thing! I forgot we need an aria-label on the StateIcon since it's not decorative, but informative. It should tell the user it's "Complete" or "Incomplete". 

I would take the copy and translation from the [poolMessages.ts](https://github.com/GCTC-NTGC/gc-digital-talent/blob/11075-task-card-component/apps/web/src/messages/poolMessages.ts#L4-L15), and use them as the default value, but also provide props `ariaLabelIncomplete` (?) and `ariaLabelComplete` (?) to the `<StateIcon />`.

yonikid15 (commented):
Okay I found the [Icon](https://github.com/GCTC-NTGC/gc-digital-talent/blob/11078-resource-block-component/apps/web/src/components/AssessmentResultsTable/utils.tsx#L92-L93) I was looking for that that meets accessibility standards. 

I would add the aria-label to the StateIcon instead of the div

petertgiles:
OK, yeah, Eric told me the same thing - arial-hidde=false.  :laughing: 

He also said he prefers the custom title to labeling the icon though.  It's easier to understand a custom tile being read through the SR than have a graphic be read out in the item so I just tightened that up a bit.
image

packages/ui/src/components/ResourceBlock/ResourceBlock.tsx
```diff
@@ -0,0 +1,226 @@
...
+  children,
+}: RootProps) => {
+  const headingId = useId();
+  return (
+    <div
+      data-h2-shadow="base(larger)"
+      data-h2-border-radius="base(rounded)"
+      data-h2-background-color="base(foreground)"
+      role="list"
+      aria-labelledby={headingId}
```
yonikid15, line 84: image

Did you want to make this a list `li`? If so you'll need to add `role=listitem` to the children.

yonikid15 (changes requested):

packages/ui/src/components/ResourceBlock/ResourceBlock.tsx
petertgiles: Oh, good catch.  The children do have that role, further down.  I moved the list role down the tree a bit.
[move list role](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11200/commits/3a25f5f3605d7dc9585f3bdb991f310060847ac6)

yonikid15: Do you need to move the aria-labelledby as well? :sweat_smile:

petertgiles: OK, as discussed I left the list unlabelled and switched the top level element to a nav.
image
[switch to nav element](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11200/commits/eb7a2a012e75ed74a06d2c7e8692de70d50bac42)

PR #11206, opened 2024-08-09
[Bugfix] Remove id from verify email operations

### LINKED ISSUE #11155: 🐛 On protected endpoint, can send another user's email verification message
## 🐛 Bug

If you're on the protected endpoint it might be possible to send another user their email verification message.  I think this bug was masked due to using the unprotected endpoint in the test.

## 🦋 Expected Behaviour

It should only be possible to send the email verification for the currently logged in user.

## 📋 Steps to Reproduce

1. In api/tests/Feature/UserVerifyEmailTest.php, change `use UsesUnprotectedGraphqlEndpoint;` to `use UsesProtectedGraphqlEndpoint;`
2. Run `./artisan test tests/Feature/UserVerifyEmailTest.php  --filter=testUserCantGenerateNotificationForSomeoneElse`

## 🙋‍♀️ Proposed Solution

Maybe a policy or validator on that the `SendUserEmailVerification` mutation.

## ✅ Acceptance Criteria

- [ ] Test updated or new test written to expose bug
- [ ] Can no longer send email verification to another user, regardless of endpoint.
- [ ] Remove user id from email verification endpoints. Should just use authenticated user's id

vd1992:
🤖 Resolves #11155

## 👋 Introduction

Remove user id from arguments in favour of using `Auth`

## 🧪 Testing

1. Test the verification workflow and check it stills works
2. Can bypass email by checking the jobs table in database


esizer (approved):
Great! :+1: 

Thanks for the tip to grab the verification code out of the jobs table :pray:

PR #11225, opened 2024-08-13
[Debt] Remove unused data from snapshot

### LINKED ISSUE #11159: :recycle:  Snapshots storing too much information
## :recycle: Debt

Our snapshots are storing too much unnecessary information, particularly the pool candidates relationship.

## 🦋 Expected Behaviour

Snapshots only store the necessary information that needs to be frozen for a specific application and nothing else.

## 🕵️ Details

Right now, we are storing the user model along with all of the pool candidates relationship. We almost never need any of the pool candidates in the snapshot. 

The snapshot is stored in the `pool_candidates` table so it is redundant to store that info in the snapshot.

## 🙋‍♀️ Proposed Solution

Stop storing the user's pool candidates in the snapshot.
~Remove pool candidates from existing snapshots with an artisan command. (This will be long and slow because of many candidates in the database.)~ We'll wait for a future ticket to edit existing snapshots.

## :stop_sign:  Blockers

```[tasklist]
### Blocked By
- [ ] #11126 
```

## ✅ Acceptance Criteria

- [ ] Pool candidates no longer stored in snapshots
- [ ] Any code using the candidates from the snapshot updated to access the candidate directly
- [ ] ~Runnable artisan command which deletes pool_candidates from existing Pool Candidate snapshots.~

esizer:
🤖 Resolves #11159 

## 👋 Introduction

Updates snapshots to no longer store information we do not use.

## 🕵️ Details

The snapshot is really only necessary to get store a frozen version of the users profile including skills and experiences. We were storing all of their pool candidates full objects and never actually used them. This removes those from being stored in the snapshot along with some other redundant information.

## 🧪 Testing

1. Build `pnpm run dev:fresh`
2. Login as any user
3. Apply to a pool
4. Logout
5. Login as an admin `admin@test.com`
6. Navigate to the application you created in step 3
7. Confirm all application information still appears as expected
8. Confirm we no longer use `poolCandidates` from the snapshot


apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/ApplicationInformation/EducationRequirementsDisplay.tsx
```diff
@@ -10,14 +10,119 @@ import { notEmpty } from "@gc-digital-talent/helpers";
...
+      ... on WorkExperience {
+        role
+        organization
+        division
+        startDate
+        endDate
+      }
+    }
+  }
+`);
```
esizer, line 110: This is admittedly a lot of info. I kind of wish we did not need to query all this just to show the experience they selected for the education requirement :thinking: 

I kinda just ripped this from the career timeline page. Maybe we need to create a fragment for `ExperienceCard` so we know we are only ever querying the required data?

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/ApplicationInformation/EducationRequirementsDisplay.tsx
```diff
@@ -10,14 +10,119 @@ import { notEmpty } from "@gc-digital-talent/helpers";
...
+      value
+    }
+    educationRequirementExperiences {
+      id
+      details
+      user {
+        id
+        email
+      }
+      skills {
```
vd1992, line 30: I don't quite get what the desired state of the snapshot is, but I am not too sure about this bit. Doesn't this now query for the live experience result attached to `educationRequirementExperiences`? 
Which means if the user updates the experience(s) after application, this component will fetch different states? I think `educationRequirementExperiences` should be stored somehow in the snapshot

esizer: That is a good point. Though, it seems a little weird to hold application data in the profile snapshot :sweat_smile: 

But yeah, this actually will cause issues with the skills as well :woozy_face:

esizer: Okay, I think I found a solution. We store the users experiences so if we have the ID then we should be able to get it from those :thinking: Do you think this will work?

[use experiences from snapshot](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11225/commits/00b7d9703a32a6f4f235813dd1632e2aa47334d7)

vd1992: Hm... Maybe? It looks like a good idea to me

tristan-orourke: @tristan-orourke

tristan-orourke: Ohh... I see... It's confusing that you can query for nested details directly on the educationRequirementExperiences, but they're kindof the wrong details :/

I think you have a good solution for now though.

tristan-orourke: As a longer term solution... Make an accessor which actually gets values for educationRequirementExperiences from the snapshot? Essentially, moving this logic to the backend?

tristan-orourke (commented):
I haven't done a full thorough review, but the changes to PoolCandidate->setApplicationSnapshot and the Resource classes look right to me 👍

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/ApplicationInformation/EducationRequirementsDisplay.tsx
esizer: I'm a little confused, that is what this is doing right now :thinking: 

The education experience is just an array of IDs that point to the experiences so all we need to do is match it with the snapshotted experience (what this does). The reason it happens on the frontend is because if we did that on the backend it would just be duplicating the same data and sending unnecessary bytes does the pipe.

tristan-orourke: In the commit you linked, you deleted a bunch from the graphql request for educationRequirementExperiences and stuck to getting the ids. But, as your old code showed, you _can_ request all the experience details direct from graphql. But, for what's most relevant to the application (the PoolCandidate), those details are kindof wrong. Which is why you're now only getting the ids and making the connection to the snapshot on the frontend. I'm saying, it would probably be better if the query you deleted in [use experiences from snapshot](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11225/commits/00b7d9703a32a6f4f235813dd1632e2aa47334d7) just returned snapshot data in the first place. Moving this logic (connecting experience-id to snapshot-experience) from React to Laravel.

esizer: Yeah, that is why I am confused. It would be the exact same data though no? So we would be sending it down twice, once in the experiences array and then again in the education experiences. That was the main reason I deleted it, because the data was identical, we just needed the ID to connect the two. Maybe I am missing something but it seems the following is being proposed:

```json
{
  "experiences": [
     {
       "id": 1,
        "title": "Education experience",
     }
  ],
  "educationRequirementExperiences": [
    {
      "id": 1,
      "title": "Education experience"
    }
  ]
}
```

Unless I am missing something :sweat_smile:

vd1992: I feel like creating an accessor and schema addition would be more worth it if we needed to reuse it elsewhere

petertgiles (approved):
Smart, I love how much code you were able to remove.  Seems to work well for snapshots created before and after.  :clap:

PR #11235, opened 2024-08-14
[Debt] Removes unused dependencies

### LINKED ISSUE #11131: ♻️ Remove unused dependencies
## ♻️ Debt/Refactor

We have some depenendencies that do not seem to be in use and some that don't do a lot and maybe can be removed?

### Unused

- `core-js`
- `cpy-cli`

### Minimally used

- `path-browserify`
  - We only use this to stitch URL paths together, I feel like we could use a simple join instead :thinking: 
- ` js-yaml`
   - Seems like this is used to read intl whitelist files, could we just change the format of those files to JSON? :thinking: 
- `enforce-unique`
  -  We use this in 3 fake data generators, is this because `faker.string.uuid()` is not unique? 
  - NOTE turns out this is necessary, `faker.string.uuid()` is NOT unique

### `lodash`

This one gets its own section because we do use it in a lot of places but, most of the uses I see are either nativley supported or is pretty easy to implement in our `helpers` package. There are probably some that I missed but here aree some basic reimplementations of our uses of lodash. Not all are going to just work and we may have some edge cases but I think we could easily remove this depeendency since most helpers in it are pretty basic.

For example, we use `isEmpty` and `isArray` a lot and these should just be `!item.length` or `Array.isArray(maybeArray)`.

#### `pick`

```ts
// Instead of 
function (data) {
  const values = pick(data, ["name", "group"]);
}

// We can
function({ name, group});
```

#### `debounce`

This should work.

```ts
 function debounce<F extends (...args: Parameters<F>) => ReturnType<F>>(
  fn: T,
  delay: number,
): (...args: Parameters<F>) => void {
  let timeout: ReturnType<typeof setTimeout>;
  
  return (...args: Parameters<F>): void => {
    clearTimeout(timeout);
    timeout = setTimeout(() => func(...args), delay);
  };
}
```

#### `merge`

This is just:

```ts
const new = { ...default, ...override };
```

#### `random`

```ts
function random(min, max) {
  return Math.random() * (max - min) + min;
}
```

#### `uniqueId`

Use one of:
- `React.useId()`
- `faker.string.uuid()`

### `isEqual`

Only really use this for object comparison.

```ts
function isObject(value) {
  return value !== null && typeof value === 'object';
}

function isEqual(objA, objB) {
  const keysA = Object.keys(objA);
  const keysB = Object.keys(objB);
  
  if (keysA.length !== keysB.length) return false;
  
  for(let key of keysA) {
    const valA  = objA[key];
    const valB = objB[key];
    
    if bothObjs = Object.is(valA) && isObject(valB);
    
    if(
      (bothObjs && !isEqual(valA, valB)) ||
      (!bothObjs && valA != valB)
    ) {
      return false;
    }
    
    return true
}
```

#### `orderBy`

```ts
array.sort((a, b) => { });
```

#### `omit`

`delete obj.key`

#### `head`

`array[0]`

#### `reverse`

`arr.reverse();`

#### `compact`

`arr.filter((value) => !!val);`

#### `isEmpty`

For the most part, just use an empty check on object type.

```ts
!array.length;
!map.size;
!string.length;
```
#### `uppserCase`

`str.toUpperCase();`

## 🙋‍♀️ Proposed Solution

- Remove the unused dependencies
- Evaluate to see if we can imperilment the rarely used ones to be removed
- See about reducing our usage of lodash and maybe remove it entirely? 
- Maybe I missed some and you can :axe: more?

## ✅ Acceptance Criteria

- [ ] Unused dependencies removed
  -  `core-js`
  - `cpy-cli`
- [ ] remove `path-browserify`. Update code where it was used.
- Removing Lodash remains controversial. There might be a future ticket to reduce usage

mnigh:
🤖 Resolves #11131.

## 👋 Introduction

This PR removes the dependencies `core-js`, `cpy-cli`, and `path-browserify`. The `path-browserify` function for joining paths is replaced by [join](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join).

## 🧪 Testing

1. `pnpm i; pnpm build:fresh`
2. Navigate throughout site and ensure all links still work

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -57,7 +57,6 @@
     "framer-motion": "^11.3.24",
     "graphql": "^16.9.0",
     "lodash": "^4.17.21",
-    "path-browserify": "^1.0.1",
     "react": "^18.3.1",
     "react-dom": "^18.2.0",
     "react-helmet-async": "^2.0.5",
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -88,7 +88,6 @@
     "@types/react": "^18.3.3",
     "@types/react-dom": "^18.3.0",
     "@vitejs/plugin-react": "^4.3.1",
-    "core-js": "^3.37.1",
     "dotenv": "^16.4.5",
     "eslint": "^8.57.0",
     "jest": "^29.7.0",
--- a/apps/web/src/hooks/useRoutes.ts
+++ b/apps/web/src/hooks/useRoutes.ts
@@ -1,4 +1,3 @@
-import path from "path-browserify";
 import { useIntl } from "react-intl";
 
 import { Locales, getLocale } from "@gc-digital-talent/i18n";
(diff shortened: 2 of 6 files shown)


esizer (approved):
LGTM! :+1:

Issue #9483, opened 2024-02-22
✨ Add toggle to "Classification" to manage the filter on Search page


Jerryescandon:
## ✨ Feature

Currently the classification filter on the search page is managed by hard-coding the available classifications into the dropdown list, we would like to create a functionality that allows admin users to control this directly through the admin page.

## 🕵️ Details

An admin should be able to:
1. Navigate to the classifications page located on the admin portal
2. Select a classification from the table
3. Edit the information from that classification
4. Turn the toggle ON or OFF
5. Have that change affect whether or not a classification shows up on the search page.

## 🎨 Design File

[Figma file](https://www.figma.com/file/Ov3ZwbSJXh7Ckqk6cefPDo/Admin---Classifications?type=design&node-id=5115%3A20828&mode=design&t=LWU99DTRUaNGgRuA-1) by @Jerryescandon 

### Target filter list located on search page

location: `https://talent.canada.ca/en/search`

![image](https://github.com/GCTC-NTGC/gc-digital-talent/assets/16977254/273a3539-db14-403a-bb12-e41e1c19f586)

### New toggle to control visibility on the list

location: 

![Image](https://github.com/user-attachments/assets/2f7e2b39-b9dd-461e-a1bf-2ca341c7987a)

![Image](https://github.com/user-attachments/assets/59fe2849-a14a-466d-bcd8-728f16b1945f)

## 🙋‍♀️ Proposed Implementation

(optional) Do you have a proposed implementation?

## 🌎 Localization

(optional) Provide any new copy along with translations available.

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the feature was properly implemented.

- [ ] Platform Admins can control which classification-levels appear on the talent search page
- [ ] Search list uses the new display name field and fallback to regular name if empty
- [ ] Admins can assign a Display name for Group-level combinations that is displayed on the search page
- [ ] Classification admin table updated with new fields.
- [ ] Deployment instructions (seeder, artisan command, etc) to restore current list of classifications with their descriptions on the search page

mnigh:
The open-ended nature of the **Display name** seems a bit dangerous for formatting and could lead to yet another source of truth for the "Generic Job Title", not to mention it does not provide for the aria-label value used in search (and, like the regular labels themselves, is hard coded). If this is intended to be the "Generic Job Title", why not call it that and use it elsewhere?

## Screenshot

[...]

PR #11264, opened 2024-08-15
[Fix] Improve search page initial load

### LINKED ISSUE #11163: ♻️ Search page startup is slow
## ♻️ Debt/Refactor

Our search page often takes 5 - 10 seconds to fully load.  Can we improve this?

## 🕵️ Details

It seems like the two slowest queries are CandidateCount and SearchForm which both take over 2 seconds.  

## 🙋‍♀️ Proposed Solution

It makes sense that CandidateCount is a bit slow.  But the SearchForm query which populates the form and prevents the page from even showing is probably unnecessarily slow.  It's all public and pretty static data - can we cache it between users?  Or can we add some indexes?  Or are we overfetching?

Lets focus on the SearchForm query
1. Check for overfetching (are requesting more data than we need?)
2. Do we need to load pools at all, or is there an alternative? If we just need names of pools for the buttons at the bottom, would it be faster to run narrower queries after the candidate count comes back (or include pool name data in the candidate count query)
3. If we want to continue pre-fetching pools, could we filter out some (eg archived)
4. Is there a reason the query is slow, besides quantity of data (lots of policies running, complicated db query, etc?)
5. Finally, could consider [caching](https://lighthouse-php.com/6/performance/caching.html)

## ✅ Acceptance Criteria

- [ ] ~Improve speed of search page startup~
- [ ] Improve speed of SearchForm query used to render Search page

esizer:
🤖 Resolves #11163

## 👋 Introduction

Refactors the search page queries to avoid needing to query unnecessary data.

## 🕵️ Details

We were pulling down all published pools and cross referencing them with all classifications to determine which oncs to show in the drop down.

This simply hard codes in the classifications we want to show and allows empty results if you select one with no published pools to save us from pulling down 2mb of data.

## 🧪 Testing

1. Build `pnpm run dev:fresh`
2. Navigate to search `/search`
3. Interact with the form confirming you still get expected results back
4. Confirm that the data being queried is reasonable

### DIFF
--- a/api/tests/Feature/ApplicantTest.php
+++ b/api/tests/Feature/ApplicantTest.php
@@ -1923,7 +1923,6 @@ class ApplicantTest extends TestCase
 
     }
 
-
     public function testClassificationAndStreamsFilter()
     {
         $targetClassification = Classification::factory()->create();
@@ -1938,7 +1937,7 @@ class ApplicantTest extends TestCase
         ]);
         $targetStreamPool = Pool::factory()->candidatesAvailableInSearch()->create([
             'classification_id' => $excludedClassification,
-            'stream' => $targetStream
+            'stream' => $targetStream,
         ]);
         $targetStreamAndClassificationPool = Pool::factory()->candidatesAvailableInSearch()->create([
             'classification_id' => $targetClassification,
@@ -1946,7 +1945,7 @@ class ApplicantTest extends TestCase
         ]);
         $excludedPool = Pool::factory()->candidatesAvailableInSearch()->create([
             'classification_id' => $excludedClassification,
-            'stream' => $excludedStream
+            'stream' => $excludedStream,
         ]);
 
         $targetUser = User::factory()->create();
@@ -1957,43 +1956,43 @@ class ApplicantTest extends TestCase
         // Should show up for classification filter only
         PoolCandidate::factory()->availableInSearch()->create([
             'user_id' => $targetUser,
-            'pool_id' => $targetClassificationPool
+            'pool_id' => $targetClassificationPool,
         ]);
 
         // Should should up for stream filter only
         PoolCandidate::factory()->availableInSearch()->create([
             'user_id' => $targetUser,
-            'pool_id' => $targetStreamPool
+            'pool_id' => $targetStreamPool,
         ]);
 
         // Should show up for both stream + classification filters
         PoolCandidate::factory()->availableInSearch()->create([
             'user_id' => $targetUser,
-            'pool_id' => $targetStreamAndClassificationPool
+            'pool_id' => $targetStreamAndClassificationPool,
         ]);
 
         // Should show up for classification filter only
         PoolCandidate::factory()->availableInSearch()->create([
             'user_id' => $targetClassificationUser,
-            'pool_id' => $targetClassificationPool
+            'pool_id' => $targetClassificationPool,
         ]);
 
         // Should should up for stream filter only
         PoolCandidate::factory()->availableInSearch()->create([
             'user_id' => $targetStreamUser,
-            'pool_id' => $targetStreamPool
+            'pool_id' => $targetStreamPool,
         ]);
 
         // Should never show up
         PoolCandidate::factory()->availableInSearch()->create([
             'user_id' => $targetUser,
-            'pool_id' => $excludedPool
+            'pool_id' => $excludedPool,
         ]);
 
         // Should never show up
         PoolCandidate::factory()->availableInSearch()->create([
             'user_id' => $excludedUser,
-            'pool_id' => $excludedPool
+            'pool_id' => $excludedPool,
         ]);
 
         $query = /* GraphQL */ '
@@ -2010,8 +2009,8 @@ class ApplicantTest extends TestCase
                             [
                                 'group' => $targetClassification->group,
                                 'level' => $targetClassification->level,
-                            ]
-                        ]
+                            ],
+                        ],
                     ],
                 ]
             )->assertJson([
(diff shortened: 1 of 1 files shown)


vd1992:
While you wait, took a quick look and nothing seems broken. 

As an aside, do we want to consider gzip-ing API responses? 
Currently, we build zipped responses on deploy (I think) and send them off with `gzip_static` when requested
This makes the static assets not large, but the API responses do not get compressed. They'd need to be compressed by server at request time after being made and could shrink some responses considerably. But this could degrade performance a bit and we don't want to gzip already gzip-ed assets.... 🤔

esizer:
> As an aside, do we want to consider gzip-ing API responses? Currently, we build zipped responses on deploy (I think) and send them off with `gzip_static` when requested This makes the static assets not large, but the API responses do not get compressed. They'd need to be compressed by server at request time after being made and could shrink some responses considerably. But this could degrade performance a bit and we don't want to gzip already gzip-ed assets.... 🤔
> 
> Some queries remain beefy due to things like fetching all the skills and skills families like on `/en/skills` So that is why I asked

Hmm, we could try doing some comparisons on dev to see about that? :thinking: If we are more careful with our queries I don't think it would be necesary...yet but if it seems like an easy win then maybe :+1: 

I just tested our skills page and while it isn't bad, 2.1s added just for the skills query seems like we could have some easy wins there but I wouldn't say super high priority.

https://gtmetrix.com/reports/talent.canada.ca/fGB9VcCp/

apps/web/src/pages/SearchRequests/SearchPage/hooks.ts
```diff
@@ -74,22 +69,14 @@ export const useCandidateCount = (
     [filters],
   );
 
-  // The countApplicants query ignores the pool filter if it is an empty array, just like if it were undefined.
-  // However, we want to treat an empty pool filter as resulting in zero candidates.
-  // Therefore, we can skip the query and override the count results ourselves.
-  const hasPools =
```
tristan-orourke, line 80: How does removing this change the results?

api/app/Models/Classification.php
```diff
@@ -46,9 +47,23 @@ public function genericJobTitles(): HasMany
...
     protected function displayName(): Attribute
     {
+        /** @disregard P1003 Not using values */
         return Attribute::make(
             get: fn (mixed $value, array $attributes) => $attributes['group'].'-'.sprintf('%02d', $attributes['level']),
 
         );
     }
+
+    public static function scopeAvailableInSearch(Builder $query, bool $availableInSearch)
```
tristan-orourke, line 57: Let's add a comment explaining why this is here and what its for. Also, add a TODO to remove this in [#9483](https://github.com/GCTC-NTGC/gc-digital-talent/issues/9483).

tristan-orourke (commented):
@esizer I didn't do any in-depth testing, but this looks like a great approach to me!

apps/web/src/pages/SearchRequests/SearchPage/hooks.ts
esizer: This was preventing the initial query to get the count from firiring until we had all the published pools since otherwise it would retunr unexpected results.

Since we no longer use pools, it is no longer needed and we don't need to pause the query and just let it run on load now :+1:

apps/web/src/pages/SearchRequests/SearchPage/components/SearchForm.tsx
petertgiles, line 1: This is probably a pre-existing issue and should be spun off, but while the search is ongoing there is a message that there are 0 available candidates.  For a slow search, this isn't great UX.
image

petertgiles, line 1: I did some tests with the same dataset between main and this branch.  All the tests returned the same numbers except for one.  It was the one where I set a classification and a stream at the same time.  This branch reported many more candidates that `main` did.  It looks like the query from main has an additional `exists` clause that this branch is not adding:
image

petertgiles (changes requested):
Wow, this is a lot faster and cleaner!  :clap: 

I am noticing a difference in one search request, though.  :cry:

apps/web/src/pages/SearchRequests/SearchPage/components/SearchForm.tsx
petertgiles: https://github.com/GCTC-NTGC/gc-digital-talent/issues/11393

esizer: Alright, still working on a test but, I think I got this working? :thinking: 

It is not the best solution, but it should work :sweat_smile: If we continue to have issues with the search stuff, we may want to look at the queries again to see if we can simplify them further.

[attempt to fix applicant count](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11264/commits/160bab1af227b7b974ce2e2cb5751ca2cfa72adf)

petertgiles: Welp, it matches now.  :partying_face:   It looks like you've just manually added that pool ID filter back in?

esizer: Yeah, there is probably a better way to do that but this was the easiest. We are basically replicating work that was happening on the frontend but now on the backed instead :sweat_smile: 

I also got the test figured out... I think. [add regression test for filter changes](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11264/commits/cba675a6e8fa8efd1a5ab7a6c7f5723610c13379)

PR #11271, opened 2024-08-15
[Feature] Static document hosting

### LINKED ISSUE #10341: ♻️ Move static files in `apps/web/src/assets/documents` to blob storage
## ♻️ Refactor

These files are all pdf, docx, or pptx and are not actually rendered as part of the application and slows down builds.

## 🕵️ Details

Currently add up to ~15MB worth of files.

## 🙋‍♀️ Proposed Solution

1. Move all the files in `apps/web/src/assets/documents` to [blob storage in Azure](https://azure.microsoft.com/en-us/products/storage/blobs) and update the links accordingly
2. Delete dir and files in `apps/web/src/assets/documents`

Custom subdomain to point to Azure hosted static files?
Consider that we may move other things to this storage solution in the future, like tc-report.

## ✅ Acceptance Criteria

- [ ] files are still able to be accessed but not directly bundled as part of the application
- [ ] e2e tests to ensure that links to assets stay live
- [ ] redirect from old links
  - nginx?  F5 firewall?

petertgiles:
🤖 Resolves #10341 

## 👋 Introduction

This branch removes in-repo documents so that they can be statically hosted from Azure instead.

## 🕵️ Details

There's a redirect happening in the reverse proxy to send any requests to /static/ to the Azure static hosting.  There are three different static hosting containers - one for each vertical.

dev -> dev static hosting container
uat -> uat static hosting container
prd -> prd static hosting container

Since there's no static hosting container for local, I've added an Nginx rule to redirect those requests to the prod static hosting.

As discussed in standup - an e2e test for a file that doesn't exist in a vertical doesn't make much sense.  That's spun off here: https://github.com/GCTC-NTGC/gc-digital-talent/issues/11374

If you want to inspect the files in Azure, make sure you're using one of the Azure Virtual Desktops or access will be denied.

## 🧪 Testing

>[!NOTE]
There were 26 files moved.  Maybe pick a few different ones to check.  :sweat_smile: 

- [ ] Locally, requests for documents get redirected to the prod static hosting.
- [ ] In dev or uat, requests for documents come from the Azure static hosting
- [ ] For prod, try manually downloading some of the /static/ document links in your browser

## 📸 Screenshot

image


infrastructure/conf/nginx-conf-local/default
```diff
@@ -186,6 +214,11 @@ server {
         try_files /tc-report/_site/$uri /tc-report/_site/$uri/index.html /;
     }
 
+    # The reverse proxy is not present locally so this provides a temporty redirect to prod static hosting for testing
```
vd1992, line 217:
```suggestion
    # The reverse proxy is not present locally so this provides a temporary redirect to prod static hosting for testing
```

vd1992 (approved):
Okay

Tried jumping around on local, UAT, and prod downloading files. Files seemed to all download without issue and the correct file was served up. 
Could it be, blob is finally... done?

petertgiles:
I deployed this branch to dev and it looks good.

PR #11275, opened 2024-08-16
[Feature] Add user id to search request table

### LINKED ISSUE #10976: ✨ Add userId to Talent Requests if logged in
## ✨ Feature

If a user is logged in while submitting a talent search request, we should associated them with the request. 

## 🕵️ Details

This requires adding a user_id column to the talent requests database table. It does not require changing the api schema, since we can simply get the logged in user if it exists and save that to the request object.

For now, do not add userId to the Talent Request type in the schema - this userId need not be visible to anyone who is looking at talent requests.

## ✅ Acceptance Criteria

- [ ] If a talent request is submitted by a logged-in user, the user id is saved in the database
- [ ] The graphql public schema does not change
- [ ] PHPUnit test that userid is saved for logged in user, and not required for guest user

## 🛑 Blockers

```[tasklist]
### Blocked By
```

yonikid15:
🤖 Resolves #10976 

## 👋 Introduction

- Adds a user_id column to the pool_candidate_search_requests table
- If a user is logged in then set the user_id to the logged in users id, if a guest user is logged in then leave it as null

## 🕵️ Details

## 🧪 Testing

1. Ensure search requests still works correctly
2. Ensure user id is only set when a user is logged in and not a guest

## 🚚 Deployment

Run migration for new column.

### DIFF
--- a/api/app/Models/PoolCandidateSearchRequest.php
+++ b/api/app/Models/PoolCandidateSearchRequest.php
@@ -55,6 +55,22 @@ class PoolCandidateSearchRequest extends Model
     ];
 
     /**
+     * Boot function for using with PoolCandidateSearchRequest
+     *
+     * @return void
+     */
+    protected static function boot()
+    {
+        /** @var \App\Models\User */
+        $user = Auth::user();
+
+        parent::boot();
+        static::creating(function ($query) use ($user) {
+            $query->user_id = $user ? $user->id : null;
+        });
+    }
+
+    /**
      * The "booted" method of the model.
      */
     protected static function booted(): void
--- a/api/app/Models/PoolCandidateSearchRequest.php
+++ b/api/app/Models/PoolCandidateSearchRequest.php
@@ -65,8 +65,8 @@ class PoolCandidateSearchRequest extends Model
         $user = Auth::user();
 
         parent::boot();
-        static::creating(function ($query) use ($user) {
-            $query->user_id = $user ? $user->id : null;
+        static::creating(function ($searchRequest) use ($user) {
+            $searchRequest->user_id = $user?->id;
         });
     }
 
--- a/api/database/factories/PoolCandidateSearchRequestFactory.php
+++ b/api/database/factories/PoolCandidateSearchRequestFactory.php
@@ -32,7 +32,7 @@ class PoolCandidateSearchRequestFactory extends Factory
         $communityFetched = Community::inRandomOrder()->first();
         $community = isset($communityFetched) ? $communityFetched : Community::factory()->create();
         $guestUser = null;
-        $users = array(User::inRandomOrder()->first(), $guestUser);
+        $users = [User::inRandomOrder()->first(), $guestUser];
         $user = $users[array_rand($users)];
 
         return [
--- a/api/database/factories/PoolCandidateSearchRequestFactory.php
+++ b/api/database/factories/PoolCandidateSearchRequestFactory.php
@@ -9,6 +9,7 @@ use App\Models\ApplicantFilter;
 use App\Models\Community;
 use App\Models\Department;
 use App\Models\PoolCandidateSearchRequest;
+use App\Models\User;
 use Illuminate\Database\Eloquent\Factories\Factory;
 
 class PoolCandidateSearchRequestFactory extends Factory
@@ -29,8 +30,8 @@ class PoolCandidateSearchRequestFactory extends Factory
     {
 
         $communityFetched = Community::inRandomOrder()->first();
-
         $community = isset($communityFetched) ? $communityFetched : Community::factory()->create();
+        $user = User::inRandomOrder()->first();
 
         return [
             'full_name' => $this->faker->name(),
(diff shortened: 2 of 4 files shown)


api/app/Models/PoolCandidateSearchRequest.php
```diff
@@ -54,6 +54,22 @@ class PoolCandidateSearchRequest extends Model
...
+     *
+     * @return void
+     */
+    protected static function boot()
+    {
+        /** @var \App\Models\User */
+        $user = Auth::user();
+
+        parent::boot();
+        static::creating(function ($query) use ($user) {
```
petertgiles, line 68: The parameter name `$query` is a bit strange.  Maybe `$searchRequest` or even just `$model`?

api/app/Models/PoolCandidateSearchRequest.php
```diff
@@ -54,6 +54,22 @@ class PoolCandidateSearchRequest extends Model
...
+            $query->user_id = $user ? $user->id : null;
```
petertgiles, line 69: You could probably avoid this ternary with:
```suggestion
            $query->user_id = $user?->id;
```

api/database/factories/PoolCandidateSearchRequestFactory.php
```diff
@@ -49,6 +50,7 @@ public function definition()
             'position_type' => $this->faker->randomElement(PoolCandidateSearchPositionType::cases())->name,
             'reason' => $this->faker->randomElement(PoolCandidateSearchRequestReason::cases())->name,
             'community_id' => $community->id,
+            'user_id' => $user->id,
```
petertgiles, line 53:
```suggestion
            'user_id' => $user?->id,
```

api/database/factories/PoolCandidateSearchRequestFactory.php
```diff
@@ -29,8 +30,8 @@ public function definition()
     {
 
         $communityFetched = Community::inRandomOrder()->first();
-
         $community = isset($communityFetched) ? $communityFetched : Community::factory()->create();
+        $user = User::inRandomOrder()->first();
```
petertgiles, line 34: I'd recommend having some of these seed null for realism.

api/app/Models/PoolCandidateSearchRequest.php
petertgiles, line 1: I'm confused.  The issue says:
> For now, do not add userId to the Talent Request type in the schema - this userId need not be visible to anyone who is looking at talent requests.

How are we going to pull these for the manager homepage without a schema change?

api/database/factories/PoolCandidateSearchRequestFactory.php
petertgiles: This would probably fix the currently failed CI test.

petertgiles (changes requested):
Works great!  I left a few minor things to check on.

api/app/Models/PoolCandidateSearchRequest.php
yonikid15: Good catch, fixed in [21bdebf](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11275/commits/21bdebfc200f4d82b3cc1ab61a59aba94c4b2576)

api/database/factories/PoolCandidateSearchRequestFactory.php
yonikid15: Awesome thanks :+1: Fixed in [bb360d9](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11275/commits/bb360d9a772ebe3a4d79357f56c91a947f4a8127)

api/database/factories/PoolCandidateSearchRequestFactory.php
yonikid15: Good call! Fixed in [bb360d9](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11275/commits/bb360d9a772ebe3a4d79357f56c91a947f4a8127)

api/app/Models/PoolCandidateSearchRequest.php
yonikid15: Ya that's a good point. The currently logged in manager wouldn't need this to be queried, however if an admin was to look search requests if we want this data available we would need the user_id. @tristan-orourke what was the reasoning for hiding this from the schema?

tristan-orourke: My reasoning is that we shouldn't add fields to the schema that we don't have a use case for yet.

For the manager dashboard and request history page, we will need to modify the schema, but we're not there yet. And I think the best way to do that may be to add a `myRequests` field to the User object instead of adding a way to search requests by userId.

As for platform or community admins, seeing this user_id isn't useful to them as the page is right now. Later, if we decide to, we _could_ show some details from the manager's profile when viewing a request. Again, we can update the schema when we get to that, to make sure its the most appropriate change possible.

petertgiles (approved):
Looks great! :medal_sports:

PR #11283, opened 2024-08-19
[Debt] Bump `typescript-eslint` to v8

### LINKED ISSUE #11123: ♻️ Bump `typescript-eslint` to v8
## ♻️ Debt/Refactor

There is a new major version out for `typescript-eslint` with some breaking changes so we need to bump it manually.

## 🙋‍♀️ Proposed Solution

Follow [the blog post](https://typescript-eslint.io/blog/announcing-typescript-eslint-v8#as-an-existing-user) about fixing breaking changes.

## ✅ Acceptance Criteria

- [ ] Eslint plugins/configs for `typescript-eslint` are on >=8.0.0
- [ ] All breaking changes resolved

esizer:
🤖 Resolves #11123 

## 👋 Introduction

Bumps `typescript-eslint` to the latest version, updating the config and fixing any new errors.

## 🧪 Testing

1. Run lint `pnpm run lint --force`
2. Confirm it passes

### DIFF
--- a/packages/eslint-config-custom/react.js
+++ b/packages/eslint-config-custom/react.js
@@ -87,7 +87,7 @@ module.exports = {
         td: ["gridcell"],
       },
     ],
-    "@typescript-eslint/ban-types": [
+    "@typescript-eslint/no-restricted-types": [
       "warn",
       {
         types: {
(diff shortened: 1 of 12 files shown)


packages/eslint-config-custom/react.js
```diff
@@ -97,6 +97,18 @@ module.exports = {
...
+      {
+        args: "all",
+        argsIgnorePattern: "^_",
+        caughtErrors: "all",
+        caughtErrorsIgnorePattern: "^_",
+        destructuredArrayIgnorePattern: "^_",
+        varsIgnorePattern: "^_",
+        ignoreRestSiblings: true,
+      },
+    ],
```
esizer, line 111: Not sure why we never had this before but you can ignore this rule by prefixing vars we never intend to use with an underscore now `_`

vd1992: Is this not just replacing one ignore statement with another? Ignore unused var -> ignore underscore dangle
😕 

image

esizer: Only here since it is a proper var. Honestly this whole type gymnastics we are doing here is a bit odd and we should revisit it but for now, adding an ignore is probably sufficient.

vd1992: I think it is my type gymnastics 🥹

vd1992 (approved):
install ☑️ 
build ☑️ 
lint ☑️

PR #11294, opened 2024-08-20
[Feature] Nav context with nav roles

### LINKED ISSUE #11034: Certain pages can be associated with certain roles
## ✨ Feature

The new nav menu (#10793) will have different "modes", representing different roles. It will only show links relating to one role at a time. 

For the most part, the "nav role" depends on the page you're currently on. If you're applying to a job, you're acting as an Applicant. If you're updating skills, you're acting as an Admin. 

However, some pages make sense for multiple roles. The Account Settings page needs to be accessed by every role, for example, and Admins will need to view a lot of pages accessible to Manager or Community roles. In these cases, the nav should "remember" your "nav role" based on your previous pages.

## 🕵️ Details

The "roles" which are relevant to navigation each have their own Dashboard ([Figma link](https://www.figma.com/design/oZhQe9ahidHFfaiqgREvxa/Dashboard-(All-users)?node-id=1010-18639&t=SxW3YuDlMEIOt0dU-0)), but they don't match _exactly_ 1-to-1 with our backend Roles which group permissions. Some of them, like Community, may match with a few backend Roles. The navigation roles are: 
1. Applicant
2. Manager
3. Community
4. Admin

We should be able to map backend roles to navigation roles fairly easily:
```
[Applicant] => Applicant
[Manager] => Manager
[PoolOperator, RequestResponder, CommunityManager, ProcessOperator, CommnityRecruiter, CommunityAdmin] => Community
[PlatformAdmin] => Admin
```

Check the [nav designs](https://www.figma.com/design/oZhQe9ahidHFfaiqgREvxa/Dashboard-(All-users)?node-id=245-10680&t=SxW3YuDlMEIOt0dU-0) on Figma to help determine which pages are associated with which roles.

- The 4 Dashboards should obviously switch nav role to the matching navigation role.
- Application profile pages, including experience and skill library, should be Applicant pages.
- The job application pages should be Applicant pages.
- Account settings, and browse jobs should be neutral
- talent search / request should be Applicant or Manager
- Processes, Candidates, Requests, Teams, and Communities pages and subpages (currently on admin side) should be Community or Admin
- Users table should be Admin
- System settings (the data pages owned by Admin, like skills, skill families, departments, etc) should be Admin

## 🎨 Design File

https://www.figma.com/design/oZhQe9ahidHFfaiqgREvxa/Dashboard-(All-users)?node-id=1010-18639&t=SxW3YuDlMEIOt0dU-0

### 🧑‍🎨 Designer

@substrae 

## 📸 Screenshot

I imagine we'll want to build something into our reach router schema, which works similarly to how we check roles for permission to access a page.

Actually, we may be able to hijack the same system, as long as we're thorough with checking for permissions including Applicant role permissions.

1. We need some kind of context hook, which stores which of the nav roles (Applicant, Mangaer, Community, Admin) is active. 
2. When the user navigates to a new page, we check which backend roles can access it. We check if the user has an appropriate role - if not they get an authorization error.
3. If there's an intersection between the user's roles and the required roles, we consider the intersection and map those roles from backend role our four navigation roles. ex a page requires PlatformAdmin or CommunityAdmin to view, and the user has both. That means they could be "acting as" Community or Admin.
4. If the current "nav role" in our context hook (from step 1) matches one of their potential current roles, we keep it and end here.
5. If not, we pick the least privileged of their potential roles and make that the current role

For this to work, we also have to careful about checking for the Applicant role for pages which are only relevant to applicants, like Applicant Dashboard, Applicant Profile pages, and Job Application pages.

If a page is accessible to the public, or to any user, then the "nav role" should be left as is.

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Pages are aware of which Navigation roles they are associated with (they can be associated with one or more or none)
- [ ] When navigating to a new page, if it's associated with no roles, the nav role is unchanged (it defaults to Applicant)
- [ ] When navigating to a new page, if it's associated with a role which includes the nav role, nav role is unchanged
- [ ] When navigating to a new page, if it's not associated with the nav role, the nav role is updated to a role which matches the page and matches the user's permission roles
- [ ] nav role can be accessed via a context hook

petertgiles:
🤖 Resolves #11034 

## 👋 Introduction

This branch adds a new nav context containing nav role state.

## 🕵️ Details

As suggested in the issue, this hooks into the RequireAuth component to notify the nav context of what the authorized roles are for the current page.  If there is no RequireAuth for the current page (like on publicly accessible pages) then the nav role does not change.

I added a guest nav role for the situation where a user has not logged in yet.  I haven't seen any designs for a guest nav bar so maybe this should be removed and "applicant" is the fallback nav role?

## 🧪 Testing

There's no UI for this yet.  You can open the browser console and see the debug messages as the nav role changes.

1. Rebuild the app
2. Navigate through the app visiting pages with different authorization levels
3. Observe the nav role changing as expected

## 📸 Screenshot

image


apps/web/src/components/NavContext/NavContextContainer.tsx
```diff
@@ -0,0 +1,98 @@
...
+// this array is ordered by privilege to allow proper sorting
+const NAV_ROLES_BY_PRIVILEGE = [
+  "guest",
+  "applicant",
+  "manager",
+  "community",
+  "admin",
+] as const;
+
+export type NavRole = (typeof NAV_ROLES_BY_PRIVILEGE)[number];
```
mnigh, line 16:
```suggestion
type NavRole = (typeof NAV_ROLES_BY_PRIVILEGE)[number];
```

yonikid15 (approved):
Looks good! :rocket:

PR #11300, opened 2024-08-20
[Feat] Download all as CSV

### LINKED ISSUE #7021: ✨ Add 'Download All' state to tables with downloadable data
## ✨ Feature

Based on https://github.com/GCTC-NTGC/gc-digital-talent/issues/7018

DCM has pools with over 1500 applicants. They wish to download the entire list of applicants (CSV spreadsheet) for assessment tracking.

We have so far avoided implementing this out of fear of how much load this would put on the server.

This feature would aim to add a button option to the table that allows users to download an entire data set regardless of selections, pages or page-size.

Since this action might increase the wait time before the download can commence, we should also include a Toast notification that can set the right expectations for our users and instruct them not to navigate away.

## 🕵️ Details

<details><summary>Current design</summary>
<p>

image

</p>
</details> 

<details><summary>Proposed design</summary>
<p>

### OLD TABLE
image

### NEW TABLE
image

</p>
</details> 

<details><summary>Toast notification</summary>
<p>

Throw up the same toast we use for triggering the other backend file downloads.

</p>
</details> 

## 🎨 Design File

## 📸 Screenshot

1. Create mutations which accept the same args as poolCandidatesPaginated and usersPaginated but trigger file downloads.
2. The mutations create a queud job, passing the args and userId (of the requestor) to the job.
3. The job, when it runs, replicates the query. This will be the most annoying part of this ticket, since we need to replicate queries otherwise handled by Lighthouse directives. 
4. The job then creates saves the results in a CSV file and creates a notification for the user with a link to the file.

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] Button added to Candidate and User tables which triggers server-side generation of a CSV file which includes all items which match current filters.
- [ ] Use the same toast as in the #9693 
- [ ] Separate api that downloads all data without taking in any ids or pagination offset 
- [ ] Stress-test this on UAT with tens of thousands of candidates.

## Blocked By
```[tasklist]
### Blocked by
- [ ] #9683 
- [ ] https://github.com/GCTC-NTGC/gc-digital-talent/issues/11144
```

esizer:
🤖 Resolves #7021 

## 👋 Introduction

Adds the ability to download the full table dataset as a CSV for pool candidates and users.

## 🕵️ Details

This moves some stuff around to make the code easier to reason about (hopefully :sweat_smile:). Instead of inflating the number of mutations we create for these downloads I opted to reuse the current mutations and extended them by adding filters to them.

## 🧪 Testing

1. Rebuild `pnpm run dev:fresh`
2. Start queue `make queue-work`
4. Navigate to a pool candidates table
5. Mess around with filters and download using the "download full dataset" buttons
6. Confirm you get the expected results back in the file
7. Repeat steps 5+6 but with the users table


api/app/Jobs/GenerateUserFile.php
```diff
@@ -34,7 +34,7 @@ public function handle(): void
         } catch (\Throwable $e) {
             // Notify the user something went wrong
             $this->user->notify(new UserFileGenerationError($this->generator->getFileName()));
-            Log::error('Error generating file: '.$e->getMessage().' '.$e->getFile().':'.$e->getLine());
+            Log::error($e);
```
esizer, line 37: I hope this works for an error log... Do we need to say at what step the error occurred or do you think the stack trace makes it obvious?

api/app/GraphQL/Mutations/DownloadUsersCsv.php
```diff
@@ -24,26 +23,23 @@ public function __invoke($_, array $args)
...
-                fileName: $fileName,
+                fileName: sprintf('%s_%s.csv', Lang::get('filename.users', [], $locale), date('Y-m-d_His')),
                 dir: $user->id,
                 lang: strtolower($locale),
             );
 
+            $generator = $generator
+                ->setUserId($user->id)
+                ->setIds($ids)
+                ->setFilters($filters);
```
esizer, line 41: The constructors were getting kinda fat so I moved out some to builders and made them optional.

api/app/Traits/Generator/GeneratorFiltersById.php
esizer, line 1: Not super happy with the name of this but kinda ran out of ideas so went with this :woman_shrugging:

api/app/Traits/Generator/GeneratorIsFilterable.php
esizer, line 1: Same here, name could be better but ran out of ideas.

apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx
```diff
@@ -146,6 +146,55 @@ export const actionButtonStyles: Pick<
...
+          icon: SpinnerIcon,
+        })}
+      >
+        {label}
+      </Button>
+    );
+  }
+
+  return null;
+};
```
esizer, line 196: This is messy but hopefully we can clean it up if we move the skills download to the back end as well.

vd1992:
User CSV download can fail

image

I think it is from here

image

vd1992:
An aside, being able to download CSVs from multiple pools is odd since you have no idea which candidate is attached to what pool

esizer:
> User CSV download can fail

Oof, good catch. Im guessing that issue is actually on main :grimacing: Should be fixed in [fix users with null indigenous communities](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11300/commits/8798ec66b1c0b0aafed94b261c114b8d8f0f9b36)

esizer:
> An aside, being able to download CSVs from multiple pools is odd since you have no idea which candidate is attached to what pool

That is a good point, @tristan-orourke Should we add the pool title into the CSV download for pool candidates? :thinking:

api/app/Traits/Generator/Filterable.php
```diff
@@ -0,0 +1,90 @@
...
+
+        $filters = $this->flattenFilters($this->filters ?? []);
+        foreach ($filters as $key => $value) {
+            $scope = $scopeMap[$key] ?? $key;
+
+            if (method_exists($class, 'scope'.ucfirst($scope)) && $value) {
+                $query?->$scope($value);
+            }
+        }
+
```
vd1992, line 83: Appears to fail to apply skill filters, I think it might be due to the names

Adding logging

image

Logs - filters

image

Logs - skill filter

image

esizer: I'm pretty sure skills don't actually filter do they? It is a really weird scope that just tells you how many skills the user has that matches. I could be wrong though.

Does the document downloaded match the number of rows in the table?

esizer: Nvm, I just tested myself and it actually does filter them

esizer: Also, seems to work for me :sweat_smile:

vd1992: It does

image

image

But the CSV contains all

vd1992: Users table, I should've mentioned

esizer: oops, okay. Thansk, I'll take a look at that :ok_hand:

esizer: Fixed in [fix skills filter for user csv](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11300/commits/a94ffad1d8aef31bf0955a9dd45116ee2423a6cf)

vd1992:
So far seems to work well. 
Though, who is going to handle the stress test on UAT in the A/C?

tristan-orourke:
Okay, I've tested it on Dev with 2000+ candidates and it worked fine!

tristan-orourke:
18000 rows might be too much though 😓. I triggered a "download full dataset" on 18000 users, and five minutes later, I'm still waiting for the file.

esizer:
> 18000 rows might be too much though 😓. I triggered a "download full dataset" on 18000 users, and five minutes later, I'm still waiting for the file.

I suspect that will take a bit. It hasn't crashed or caused the CPU to melt yet has it? :sweat_smile:

api/app/Generators/PoolCandidateCsvGenerator.php
```diff
@@ -269,4 +261,35 @@ private function generatePoolHeaders()
...
+                'communityExperiences' => ['userSkills' => ['skill']],
+                'educationExperiences' => ['userSkills' => ['skill']],
+                'personalExperiences' => ['userSkills' => ['skill']],
+                'workExperiences' => ['userSkills' => ['skill']],
+            ],
+        ]);
+
+        $query = $this->applyFilters($query, PoolCandidate::class, [
+            'priorityWeight' => 'candidateCategory',
+            'poolcandidateStatus' => 'poolcandidateStatuses',
```
petertgiles, line 285: The casing seems wrong here.

api/app/Generators/PoolCandidateCsvGenerator.php
```diff
@@ -90,133 +94,123 @@ public function generate(): self
         $this->spreadsheet = new Spreadsheet();
 
         $sheet = $this->spreadsheet->getActiveSheet();
+
+        $currentCandidate = 1;
+        $query = $this->buildQuery();
+        $query->chunk(200, function ($candidates) use ($sheet, &$currentCandidate) {
+            foreach ($candidates as $candidate) {
+
+                if (! in_array($candidate->pool_id, $this->poolIds)) {
```
petertgiles, line 103: Could you please add a comment for what this special handling represents?  It's not really clear to me.

api/app/Generators/PoolCandidateCsvGenerator.php
```diff
@@ -269,4 +261,35 @@ private function generatePoolHeaders()
...
+                'userSkills' => ['skill', 'experiences' => ['skills']],
+                'awardExperiences' => ['userSkills' => ['skill']],
+                'communityExperiences' => ['userSkills' => ['skill']],
+                'educationExperiences' => ['userSkills' => ['skill']],
+                'personalExperiences' => ['userSkills' => ['skill']],
+                'workExperiences' => ['userSkills' => ['skill']],
+            ],
+        ]);
+
+        $query = $this->applyFilters($query, PoolCandidate::class, [
```
petertgiles, line 283: Feel free to ignore this if you don't agree, but it looks weird to me to reassign $query here.  At first glance it seems like the original value from L267 is being overwritten.  I think applyFilters is just returning the same Builder, right?

petertgiles: Same comment for api/app/Generators/UserCsvGenerator.php L160, api/app/GraphQL/Mutations/DownloadPoolCandidatesCsv.php L37, and api/app/GraphQL/Mutations/DownloadUsersCsv.php L38.

tristan-orourke:
I checked the failed_job table, and the job did fail, with a TimeoutExceededException. I'll create a new ticket to investigate the stacktrace and improve the capacity, but for now, I'm happy with this being able to handle a few thousand rows at a time.

api/app/Models/PoolCandidate.php
```diff
@@ -730,11 +730,16 @@ public function isDraft()
...
      */
-    public function scopeAuthorizedToView(Builder $query): void
+    public function scopeAuthorizedToView(Builder $query, ?array $args = null): void
     {
         /** @var \App\Models\User */
         $user = Auth::user();
 
+        $userId = $args['userId'] ?? null;
+        if ($userId) {
+            $user = User::find($userId);
```
petertgiles, line 740: Maybe findOrFail?  Is there a reason we would want to proceed if the given $userId doesn't exists?  Also, isn't `?? null` redundant?

petertgiles: Same comments for api/app/Models/User.php L1045.

api/graphql/schema.graphql
```diff
@@ -824,6 +824,7 @@ input PoolCandidateSearchInput {
   processNumber: String @scope(name: "processNumber")
 }
```
petertgiles, line 826: Stray newline

api/app/Traits/Generator/Filterable.php
petertgiles, line 1: The `flattenFilters` and `applyFilters` methods look a bit fiddly and error-prone.  Would you consider some tests?  It sounds like phpunit has a `getObjectForTrait` method to help with that. https://doeken.org/blog/testing-traits-in-phpunit

petertgiles (changes requested):
This is really nice and clean!  I did a code readthrough and had a few comments.

api/app/Generators/PoolCandidateCsvGenerator.php
esizer: Good catch, fixed in [fix casing for status scope map](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11300/commits/0f4cdcd48f6da194ce2e8568a608355c427f8329)

api/app/Generators/PoolCandidateCsvGenerator.php
esizer: For sure, hopefully this makes sense :sweat_smile: 

[document pool ids push](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11300/commits/60e6f1aaa5af23aab3b4b9433567a31a9d52a4c0)

api/app/Models/PoolCandidate.php
esizer: Oops, I was thinking of `array_key_exists`. This should do the same thing: [adjust userId access in authorized scopes](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11300/commits/4bcb96c1a12d4ccbb6627b23dbf8a24d994d2623)

api/graphql/schema.graphql
esizer: Removed in [remove stray line in schema](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11300/commits/d5fff4b35c2f2f8a2955a84dea33f6863f9ad982)

api/app/Generators/PoolCandidateCsvGenerator.php
esizer: Was this what you had in mind?

[remove reassignments](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11300/commits/bc77db820f0125b28dcce1a7d2860d25bf83504f)

petertgiles: Yeah, exactly.  You might still want to return the Builder from the function for chaining,

petertgiles: If you do keep the ": void" signature you'll need to get rid of the `return $query;` on L72.  It's causing an HTTP500 right now.

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
petertgiles, line 1: A fairly minor thing but the loading spinner appears for both CSV buttons at the same time when clicked.  If it's an easy fix then maybe go for it.  If it's a pain, don't worry about it.

https://github.com/user-attachments/assets/002f9ee2-f34f-4408-8241-7cfbb85a4ded

petertgiles (commented):
Seems to work great in practice.  Just noticed one thing to look at.

api/app/Generators/PoolCandidateCsvGenerator.php
esizer: Yeah, even though we are not chaining I like keeping that functionality since it is a common pattern for query builers.

[remove need to pass class in](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11300/commits/a2e00b2c8fa2d47bfe7412318a806a233ce7ec2c)

api/app/Traits/Generator/Filterable.php
esizer: Thanks for the link, I added some tests. I am probably missing some cases, if you can think of more to add let me know.

[add test for filterable trait](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11300/commits/e6a325f3f2b7cbe68d40a0a83715ba737f296ee3)

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
esizer: Hmm, maybe.... Let me think and come back. It is actually calling the same mutation so it is using the same loading bool for both buttons. Maybe I can check what triggered it :thinking:

api/app/Generators/PoolCandidateCsvGenerator.php
petertgiles: Yeah, I like this.  :+1:

api/app/Traits/Generator/Filterable.php
petertgiles: Whoah, nice!  I haven't seen tests written with providers like that before.  Seems nice.  :+1:

petertgiles (approved):

api/app/Traits/Generator/Filterable.php
esizer: Yeah, it is super nice to running the same test against multiple data sets

PR #11343, opened 2024-08-22
[Fix] Fix PreviewList title hover and focus color

yonikid15:
## 👋 Introduction

- Fixes the title hover and focus color. This was thanks to @esizer :muscle:!

## 🧪 Testing

1. Go to storybook and view PreviewList component


petertgiles (approved):
Wow, that's actually.... really smart.  I'm going to have to remember this one!

PR #11348, opened 2024-08-22
[Feat] Makes laravel aware of client locale

### LINKED ISSUE #11336: ✨ Make laravel aware of current locale
## ✨ Feature

We would like to make laravel aware of the current locale for each request so we can return localized messages instead of sending both languages and then extracting the appropriate one out in the client.

## 🕵️ Details

This will allow us to make full use of [laravels localization features](https://laravel.com/docs/11.x/localization), specifically the `__()` helper. Then, if we want, we can send just the localized version of dynamic strings and enums so we don't fetch both only to show one.

1. Pass a lang header in our `urql` client
2. Add a middleware that swaps laravel locale based on the header

## ✅ Acceptance Criteria

- [ ] Current locale from the `LocaleProvider` is sent as a header
- [ ] Middleware created to swap locale based on said header
- [ ] Maybe a unit test for the middleware to confirm we get the expected locale?

esizer:
🤖 Resolves #11336 

## 👋 Introduction

Uses the `Accept-Language` header to set the app locale for graphql requests.

## 🕵️ Details

REF: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language

For the server, we add a middleware that parses the `Accept-Language` header to get the preferred locale and set the app locale to one we support falling back to English.

On the client, we attach the header to all graphql requests basing it on the locale set in `react-intl`.

## 🧪 Testing

1. Hard to test but maybe just review the tests and ensure they pass and are comprehensive?

### DIFF
--- a/api/app/Http/Middleware/AcceptLanguageMiddleware.php
+++ b/api/app/Http/Middleware/AcceptLanguageMiddleware.php
@@ -25,9 +25,7 @@ class AcceptLanguageMiddleware
     public function handle(Request $request, Closure $next)
     {
         $locale = $request->getPreferredLanguage($this->availableLocales);
-        if ($locale == 'en' || $locale == 'fr') {
-            App::setLocale($locale);
-        }
+        App::setLocale($locale);
 
         return $next($request);
     }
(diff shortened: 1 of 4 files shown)


vd1992:
I see it set in the header

image

Added some logging

image

Updates the value appropriately 

image

vd1992:
Adding in 

image

Navigating to 

image

Yields

image

vd1992 (approved):
Per above

Playing around and it appears to all be in order. Does appear to work to me
Nice!

PR #11351, opened 2024-08-23
[Feature] Adds a content security policy in report mode

### LINKED ISSUE #10614: ✨ Deploy CSP headers with reporting
## ✨ Feature

A CSP header was designed in [#10420](https://github.com/GCTC-NTGC/gc-digital-talent/issues/10420).  Let's deploy it in reporting mode to see if there are any violations "in the wild".

## 🕵️ Details

https://github.com/GCTC-NTGC/gc-digital-talent/issues/10420#issuecomment-2147540521

How will we report it?
- Using the client-side logger into Application Insights?
- Building a Laravel endpoint to save into the database?

## ✅ Acceptance Criteria

- [ ] CSP header deployed with Nginx config
- [ ] violations reported back to our team

## 🛑 Blockers

```[tasklist]
### Blocked By
```

esizer:
🤖 Resolves #10614 

## 👋 Introduction

This adds a starting point for our content security policy in report mode along with an endpoint to log the report.

## 🕵️ Details

For those wondering how to read the policy [MDN has good documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) on it. I'll try my best to describe our specific policy:

- `default-src 'self'` This is the default, where self means anything coming from our domain
- `font-src fonts.gstatic.com 'self'` Is allowing google fonts
- `style-src-elem fonts.googleapis.com 'self' 'unsafe-inline' Allows styles from google fonts and inline.
    - NOTE: This is unsafe and I describe why later on
- `img-src 'self' data;` This is allowing images where the url starts with `data:` 
    - We may be able to get rid of this if we no longer process images through vite
    - We need this since vite can serve images as base64 if it means it is most optimal way of serving an image
- `report-uri` and `report-to` is telling the policy to report violations to our endpoint

### Unsafe inline styles

Sadly, we must allow unsafe inline styles since `@radix-ui` uses the `react-remove-scroll` library which injects a style tag we have no control over. There is an issue regarding this and hopefully we have a solution shortly but I tihnk having a policy with on unsafe is probably better than no policy at all :sweat_smile: 

 - https://github.com/radix-ui/primitives/issues/3063

## 🧪 Testing

> [!NOTE]
> Right now, I tihnk only firefox will report over localhost so it is best to test in that browser. This is because CSP will only report over a secure context and AFAIK only firefox considers localhost to be a secure context. 

1. Force violations by removing the nonce from the rule `'nonce-$cspNonce'` part
2. Reload nginx `docker exec $(docker ps | grep webserver | awk  '{ print $1 }') service nginx reload`
3. Maybe need to cache the new route `make artisan CMD="route:cache"`
4. Rebuild app `pnpm run dev:fresh`
5. Bounce around the site observing the console to confirm violations being reporting but stil allowed to run scripts
6. Open laravel logs and confirm the CSP Violation warnings
7. Undo change from step 1
8. Repeat steps 2-5 but check that the violations are gone

### DIFF
--- a/api/app/Http/Controllers/CspReportController.php
+++ b/api/app/Http/Controllers/CspReportController.php
@@ -11,9 +11,10 @@ class CspReportController extends Controller
     {
         $report = $request->json()->all();
         $message = sprintf(
-            'CSP violation, %s, %s, %s',
+            'CSP violation, %s, %s, %s, %s',
             $report['csp-report']['blocked-uri'] ?? '',
             $report['csp-report']['violated-directive'] ?? '',
+            $report['csp-report']['source-file'] ?? '',
             json_encode($report['csp-report'])
         );
 
--- a/api/app/Http/Controllers/CspReportController.php
+++ b/api/app/Http/Controllers/CspReportController.php
@@ -13,7 +13,7 @@ class CspReportController extends Controller
         $message = sprintf(
             'CSP violation, %s, %s, %s',
             $report['csp-report']['blocked-uri'] ?? '',
-            $report['csp-report']['column-number'] ?? '',
+            $report['csp-report']['violated-directive'] ?? '',
             json_encode($report['csp-report'])
         );
 
--- a/api/app/Http/Controllers/CspReportController.php
+++ b/api/app/Http/Controllers/CspReportController.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Log;
+
+class CspReportController extends Controller
+{
+    public function report(Request $request)
+    {
+        $report = $request->json()->all();
+        $message = sprintf(
+            'CSP violation, %s, %s, %s',
+            $report['csp-report']['blocked-uri'] ?? '',
+            $report['csp-report']['column-number'] ?? '',
+            json_encode($report['csp-report'])
+        );
+
+        Log::warning($message);
+    }
+}
--- a/apps/web/vite.config.ts
+++ b/apps/web/vite.config.ts
@@ -6,7 +6,6 @@ import react from "@vitejs/plugin-react";
 import { createHtmlPlugin } from "vite-plugin-html";
 import { compression } from "vite-plugin-compression2";
 import { Plugin, defineConfig } from "vite";
-
 import { hydrogen_watch } from "@hydrogen-css/hydrogen";
 
 dotenv.config({ path: "./.env" });
@@ -101,6 +100,9 @@ export default defineConfig(({ command }) => ({
   },
   server: {
     port: 3000,
+  },
+  html: {
+    cspNonce: "**CSP_NONCE**",
   },
   resolve: {
     extensions: [".ts", ".tsx", ".json", ".js"],
@@ -224,6 +226,6 @@ export default defineConfig(({ command }) => ({
         ],
       },
     }),
-    compression(),
+    compression({ exclude: /index\.html/i }),
   ],
 }));
--- a/apps/web/vite.config.ts
+++ b/apps/web/vite.config.ts
@@ -226,6 +226,13 @@ export default defineConfig(({ command }) => ({
         ],
       },
     }),
+    /**
+     * NOTE: We are not compressing the index.html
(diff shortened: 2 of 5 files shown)


petertgiles:
Quick tip - you can restart nginx with `nginx -s reload`

apps/web/vite.config.ts
```diff
@@ -224,6 +226,6 @@ export default defineConfig(({ command }) => ({
         ],
       },
     }),
-    compression(),
+    compression({ exclude: /index\.html/i }),
```
petertgiles, line 229: Could you please add a comment to remind us why it is excluded?

api/app/Http/Controllers/CspReportController.php
```diff
@@ -0,0 +1,22 @@
...
+
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Log;
+
+class CspReportController extends Controller
+{
+    public function report(Request $request)
+    {
+        $report = $request->json()->all();
+        $message = sprintf(
```
petertgiles, line 13: This works very well but I'm wondering if we're really highlighting the useful information in the fields.  I don't know that much about it but what would help us actually fix things?  violated-directive?  source-file?  We dont' have to take the first two fields - whatever number of whichever fields we want.

infrastructure/conf/nginx-conf-local/default
petertgiles, line 1: I can't get it to stop reporting violations.  :sweat-smile:  I reverted the nonce change, rebuilt the app, restarted the webserver.. and I'm still getting violations logged.  Any ideas?

petertgiles (changes requested):
It doesn't quite work for me, but it's probably my fault!

api/app/Http/Controllers/CspReportController.php
esizer: Hmm, what about:

- `blocked-uri`
- `violated-directive`

:thinking:

infrastructure/conf/nginx-conf-local/default
esizer: Hmm, can't seem to reproduce in either Chrome or Firefox :thinking_face: Can you check the following when you get a chance:

- The script tags had the proper nonce applied
- You did not have an `index.html.gz` in dist
- An extension wasn't injecting a script that was being blocked?

Since it is an inline script, my guess is it is the config script at the bottom of the page. That should be the only one :thinking:

apps/web/vite.config.ts
esizer: Good idea, added in [document reason for not compressing index.html](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11351/commits/3cfdab010ccc95e3c9dec6743c858423bb77e9d7)

api/app/Http/Controllers/CspReportController.php
esizer: Alright, updated in [update highlighted report fields](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11351/commits/808b0ce47e73abab6ce62ca5577e9e73b7927e8d)

infrastructure/conf/nginx-conf-local/default
petertgiles: Oh yeah, it was the React Developer Tools extension.  :grin:   It's coming through cleanly now.

So that makes me wonder, are we going to be able to determine if reports coming through are actionable or just noise?

petertgiles: Do we need the source-file field, maybe, for filtering?
image

esizer: It might take some trial and error but the `source-file` field seems to be a good indicator. I _just_ noticed in the dump you supplied that it is pretty obviious what is going on :sweat_smile: 

- `"source-file":"moz-extension",` - Clearly an extension caused this violation
- `"source-file":"sandbox eval code",` - I'm guessing some extensions are run in the sandbox?

esizer: Added it in [add source file to highlighted fields](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11351/commits/f58f893ede9c65cb2b72cd3e9f030d1bef662eb7)

petertgiles (approved):

PR #11352, opened 2024-08-23
[Fix] Personal information Indigenous identity

### LINKED ISSUE #11232: 🐛 Updates needed on the Personal information page under Equity options - Indigenous identity
## 🐛 Bug (+ copy)

Four points

1. On the personal information page, under equity options, if you select Indigenous identity and select both either of the first three options and the 'My community isn't listed' option, there's a warning that asks whether you're sure about selecting 'My community isn't listed' (see screenshot in details). This does not happen when only selecting the the option 'My community isn't listed'. Discussed with Gray and the warning should be removed all together.

GOB: Yes, this warning is very specific to the Apprenticeship Program and doesn't make sense on the main site. Ideally leave the code around though so we can use it in the unlikely event IAP decides to turn on their application process.

2. On the French Personal information page, under equity options - indigenous identity, the communities aren't listed alphabetically even though the description says so. We would like to change the order on the French page so it's also true here.

GOB: @mnigh rightly pointed out that the ordering was sensitive here. However, the alphabetical listing was how we felt we could best address the concerns. So it's important that we follow through with it. Let's get those sorted alphabetically in both official languages.

3. Also copy issue in the same section: We should move to using inclusive language for French and it's better for readability/accessibility to use full doublets (e.g. les étudiantes et étudiants) rather than abbreviated doublets (e.g. les étudiant(e)s). In this case that means, please change "Je suis métis(se)" to "Je suis métisse ou métis".  

## 🦋 Expected Behaviour

1. Warning should not pop up, no matter which combination of options you select.
2. Communities are ordered alphabetically on the French page as well: Inuk, Métis, Premières Nations

## 🕵️ Details

URL: https://talent.canada.ca/en/applicant/personal-information

## 📋 Steps to Reproduce

1. Go to add an indigenous community to you profile. Select any or all of the communities listed (First nations, Inuk, Metis) AND the option 'My community isn't listed'

2. Go to add an indigenous community to your profile on the French side.

## 📸 Screenshot

1. 
image

2. 
image

3.
image

## ✅ Acceptance Criteria

- [ ] The warning doesn't show anymore, no matter which combination of options you select.
- [ ] The French text for Métis reads "Je suis métisse ou métis"
- [ ] The French communities are listed alphabetically (Inuk, Métis, Premières Nations)

mnigh:
🤖 Resolves #11232.

## 👋 Introduction

This PR removes an alert that was not relevant to the profile in the Indigenous identity section of the user profile. It also updates some copy to use full doublets instead of abbreviated doublets. Finally, it orders listed communities in the `CommunitySelection` to be ordered alphabetically depending on the locale selected (en or fr).

## 🧪 Testing

1. `pnpm build:fresh`
2. Navigate to http://localhost:3000/en/applicant/personal-information#diversity-equity-inclusion-section
3. Engage the Edit this information button
4. Uncheck all the checkboxes except My community isn't listed
5. Verify the alert warning does not appear, no matter which combination of checkboxes you select
6. Switch to French
7. Engage the Edit this information button
8. Verify the text for Métis reads _Je suis métisse ou métis_
9. Verify the communities are listed alphabetically: _Inuk_, _Métis_, _Premières Nations_

## 📸 Screenshots

Screen Shot 2024-08-23 at 09 48 16

Screen Shot 2024-08-23 at 09 49 09

### DIFF
--- a/apps/web/src/components/SelfDeclaration/CommunitySelection.tsx
+++ b/apps/web/src/components/SelfDeclaration/CommunitySelection.tsx
@@ -1,5 +1,5 @@
-import { useFormContext } from "react-hook-form";
-import { useIntl } from "react-intl";
+import { FieldValues, FormState, useFormContext } from "react-hook-form";
+import { IntlShape, useIntl } from "react-intl";
 import { ReactNode, useEffect, useId } from "react";
 
 import {
@@ -21,6 +21,24 @@ interface RowProps {
   children: ReactNode;
 }
 
+interface FirstNationSectionProps {
+  labels: FieldLabels;
+  intl: IntlShape;
+  customAlertId: string;
+  formState: FormState<FieldValues>;
+  communitiesValue: string[];
+}
+
+interface InukSectionProps {
+  labels: FieldLabels;
+  intl: IntlShape;
+}
+
+interface MetisSectionProps {
+  labels: FieldLabels;
+  intl: IntlShape;
+}
+
 const Row = ({ children }: RowProps) => (
   <div
     data-h2-display="base(grid)"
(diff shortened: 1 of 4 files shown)


apps/web/src/components/SelfDeclaration/CommunitySelection.tsx
```diff
@@ -93,129 +80,150 @@ export const CommunityList = ({ labels }: CommunityListProps) => {
     }
   };
 
-  return (
+  const firstNations = (
```
esizer, line 83: Can we move these out into components? While this should be fine, it is dangerously close to [a common pitfall](https://react.dev/learn/your-first-component#:~:text=you%20must%20never%20nest%20their%20definitions) in react.

esizer (changes requested):
Looks good! One minor suggestion.

apps/web/src/components/SelfDeclaration/CommunitySelection.tsx
mnigh: a85d30f4bfad44c3007ac9fb762091bc606889fc

esizer (approved):
LGTM! :+1:

PR #11357, opened 2024-08-23
[Bugfix] Stabilize `PreviewList` story

### LINKED ISSUE #11308: 🐛 Unstable story: PreviewList
## 🐛 Bug

It looks like the example manager name in the PreviewList story is unstable

For example: https://www.chromatic.com/review?appId=61e099a1bb1465003a73d3ce&number=11294&type=linked&view=changes

## 📸 Screenshot

image

## ✅ Acceptance Criteria

- [ ] PreviewList story is stable

vd1992:
🤖 Resolves #11308  

## 👋 Introduction

Stabilize the story.
First I was going to set the seed value but noticed faker was being called to set just a single name while everything else was hardcoded. So I elected to just set the name as well to our dear superadmin's

## 🧪 Testing

1. Static name

## 📸 Screenshot

image

### DIFF
--- a/packages/ui/src/components/PreviewList/PreviewList.stories.tsx
+++ b/packages/ui/src/components/PreviewList/PreviewList.stories.tsx
@@ -1,5 +1,4 @@
 import { Meta, StoryFn } from "@storybook/react";
-import { faker } from "@faker-js/faker";
 
 import { allModes } from "@gc-digital-talent/storybook-helpers";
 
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/PreviewList/PreviewList.stories.tsx
esizer, line 1: Should we just use `faker.seed(0)` instead? :thinking: I'm fine either way but we usually use that rather than hard coding strings in stories.

vd1992: Everything else is hardcoded in the story so I figured why keep it 🤷

esizer (approved):
LGTM! :+1:

PR #11373, opened 2024-08-26
[Feature] ResourceBlock link menu

### LINKED ISSUE #11031: Role switcher dropdown
## ⚙️ Summary

Create a Dropdown that appears on the nav menu and the dashboards, allowing users to switch between different roles. Use the `<DropdownMenu />` component we already have. This component will be purely frontend work, so assume the `currentRole` is being passed in as a prop for now. As well as a switchRole method prop, which is triggered in the dropdown item `onClick`.

The component should fetch localized Role enums, and display them in the correct order. Since the drop down trigger is always the current role, we'll need detailed aria-labels to meet accessibility needs (e.g. "You are currently in the Applicant role. Select to switch roles.", and "Go to your applicant dashboard.").

## 🎨 Design File

https://www.figma.com/design/oZhQe9ahidHFfaiqgREvxa/Dashboard-(All-users)?node-id=1010-18639&t=j4U5JEvsDbOlCAf0-0

### 🧑‍🎨 Designer

@substrae 

> [!NOTE]  

#### Sub-components

> [!NOTE]
> The sub-menus described below are specific the desktop layout of the menu - the mobile version should handle sub-menus as simple `<details>` elements.

<details>
  <summary>Role sub-menu</summary>
  <p>This sub-menu only appears for users who have access to more than the applicant role on their account. Choosing an option from the role menu should take the user to that respective role's dashboard assuming they have the correct permissions.</p>
  <p>The role sub-menu is unique in that it's label changes to match the current selection. This is helpful because it allows the user to see, regardless of their location on the site, which role they're currently acting as.</p>
  
</details> 

## 🛠️ Properties

- currentRole: `string`
- switchRole: `() => void;`

### ✅ Acceptance criteria

- [ ] Component matches design for all views (mobile, dark + light themes, etc.)
- [ ] Create lang files for enum Role on backend. (Possibly re-use translations we have in other parts of the sight)
- [ ] Aria-labels added to drop down trigger and items/links
- [ ] Roles are displayed in the correct order
- [ ] The current role is highlighted when drop down is open
- [ ] Urls for each role is correct
- [ ] Only shows roles that owned by the current user

## 🛑 Blockers
```[tasklist]
### Blocked by
- [ ] https://github.com/GCTC-NTGC/gc-digital-talent/issues/11034
```

petertgiles:
🤖 Resolves #11031 

## 👋 Introduction

This branch refactors the resource block a bit and adds a link menu item.  The new hierarchy looks like this:

Root
└ BaseItem
&nbsp;&nbsp;&nbsp;├SingleLinkItem
&nbsp;&nbsp;&nbsp;└LinkMenuItem

## 🕵️ Details

The look is a little different from the Figma design because it uses links.  I think the design was completed before it was decided to use links instead of buttons.  We also have an existing dropdown component that handles things like the focus states and ensures that dropdowns look consistent across the site.  I also removed the code to make the entire block clickable.  It seems like that wasn't a hard requirement and it makes less sense when there's a block with a dropdown button in it.

If the static manager page is merged first I will update this PR to include it in the roles switcher block.  If this is merged first I'll update the manager page PR.

## 🧪 Testing

`pnpm run storybook:design-system`

- [ ] light/dark mode
- [ ] matches design

## 📸 Screenshot

image


yonikid15 (approved):
Nice work refactoring :+1: Everything works and looks good

PR #11388, opened 2024-08-26
[Fix] Prefix login path

### LINKED ISSUE #11387: 🐛 Fix login path for non-dev

esizer:
🤖 Resolves #11387 

## 👋 Introduction

Adds a prefix to the login path for non-local dev builds to build it off the root domain.

## 🧪 Testing

1. Set `apiHost` to false in `useApiRoutes`
2. Build `pnpm run dev:fresh`
3. Navigate to `/login-info`
4. Confirm the GCKey button links to the proper path

### DIFF
--- a/packages/auth/src/hooks/useApiRoutes.ts
+++ b/packages/auth/src/hooks/useApiRoutes.ts
@@ -22,7 +22,7 @@ const apiRoutes = {
 
     const url = apiHost
       ? new URL(loginPath, apiHost)
-      : ["login"].join("/") + (searchString ? `?${searchString}` : "");
+      : ["/login"].join("/") + (searchString ? `?${searchString}` : "");
 
     return url.toString();
   },


tristan-orourke (commented):
Is API_HOST even useful, then? Seems like we've been getting by without in production 😕

esizer:
> Is API_HOST even useful, then? Seems like we've been getting by without in production 😕

Yeah, it is needed for watch mode since vite runs on a different port. We are still investigating how to use some sort of reverse proxy for we can use the same port.

packages/auth/src/hooks/useApiRoutes.ts
```diff
@@ -22,7 +22,7 @@ const apiRoutes = {
 
     const url = apiHost
       ? new URL(loginPath, apiHost)
-      : ["login"].join("/") + (searchString ? `?${searchString}` : "");
+      : ["/login"].join("/") + (searchString ? `?${searchString}` : "");
```
petertgiles, line 25: I know you didn't write this code but it's a bit silly to call join on an array with one item.  :grin:

petertgiles: And why wouldn't you use `loginPath` with a slash like L33?

petertgiles (approved):
I deployed to dev and it worked.  :partying_face:

PR #11392, opened 2024-08-27
[Fix] Delete notification focus management

### LINKED ISSUE #10637: ⌨ Notifications focus after deleting a notification
## ⌨ Accessibility Issue

If you are tabbing through a bunch of notifications and delete one, focus goes back to the pane. 

### WCAG 2.1 Criteria

List any relevant criteria that is failing.

## 🦋 Expected Behaviour

Focus hops to another notification if present?

## 📋 Steps to Reproduce

1. Have multiple notifications
2. Tab to delete one
3. Observe focus

## 📸 Screenshot

image

image

## 🙋‍♀️ Proposed Solution

Focus goes to the next notification if it exists. And if it doesn't, let the focus do what it wants.

## ✅ Acceptance Criteria

- [ ] Focus goes to next notification after deletion if it exists, or to the pane.

esizer:
🤖 Resolves #10637 

## 👋 Introduction

Attempts to focus the next notification in the list on delete.

## 🕵️ Details

This stores a reference to the next item in the list on mount. If it exists when the component is unmounting (deleted) it will attempt to move focus to that next item.

## 🧪 Testing

> [!TIP]
> I found that the easiest way to seed some notifications is downloading a few files while the worker is running

2. Login as any user
3. Seed some notifications for the user
4. Using only your keyboard, delete a notification
5. Confirm that the focus is moved to the next item in the list

### DIFF
--- a/apps/web/src/components/NotificationList/NotificationButton.tsx
+++ b/apps/web/src/components/NotificationList/NotificationButton.tsx
@@ -29,6 +29,7 @@ const NotificationButton = forwardRef<HTMLButtonElement, NotificationLinkProps>(
         type="button"
         ref={forwardedRef}
         onClick={handleClick}
+        data-notification-link
         data-h2-text-align="base(left)"
         data-h2-background="base(none)"
         {...styles.link(isUnread, fetching)}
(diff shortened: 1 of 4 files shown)


vd1992 (approved):
Works great! Definitely convenient

PR #11395, opened 2024-08-27
[Bug/copy] Have "Other reason" require decision notes

### LINKED ISSUE #11391: 🐛 ROD Decision notes should be required with "Other" is the justification
## 🐛 Bug

In the ROD, if you select "Not demonstrated (Remove from process)", then select "Other reason for screening out", the Decision notes should be required. Right now it is not.

Also, what's with that period on the end of "Other reason for screening out." Let's clean that up too.

## 📸 Screenshot

image

vd1992:
🤖 Resolves #11391

## 👋 Introduction

Require decision notes if the justifications contains the "Other" option, as well as remove a bit of stray punctuation. 
One label was no longer used with the backend enums addition so it was 🪓 

## 🧪 Testing

1. Navigate to `/en/admin/candidates/:id/application`
2. Observe required conditional dependent on justifications checklist

## 📸 Screenshot

image

image


esizer (approved):
LGTM! :+1:

Issue #11153, opened 2024-08-06
🐛 New job notification has not been sent to all subscribers in the prod


brindasasi:
## 🐛 Bug

A new job posted on Aug. 2 Friday and we have around 22k users subscribed for email notifications.
I see only around 450 messages are sent out. 

## 🦋 Expected Behaviour

All subscribers should get the appropriate emails

## 🕵️ Details

I believe any error in the email queue messes up the whole remaining delivery. 

## 📋 Steps to Reproduce

Need further investigation based on prod incident. 

## 📸 Screenshot

Add a screenshot (if possible).

## 🙋‍♀️ Proposed Solution

(optional) Do you have a proposed solution?

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the bug was addressed.

- [ ] All subscribers should get the email notifications accordingly.

tristan-orourke:
Rate limiting and not retrying, or we hit rate limit, or something wrong with the queue? 
We can check logs for failures (are they all fake emails?)

PR #11396, opened 2024-08-27
Adjust tries and timeout

### LINKED ISSUE #11152: 🐛 New job posted email got sent 3 times to users about the same job
## 🐛 Bug

On Aug. 2 Friday around 1:30 pm we have sent 3 notification emails about the new job posted `IT security Analyst ` .

## 🦋 Expected Behaviour

We should not be sending duplicate emails 

## 🕵️ Details

We have 3 published processes in production. Would that align with why 3 times they have been sent out? 🤔
Here is the process details
https://talent.canada.ca/en/browse/pools/4a8e9f9f-56af-4c21-ad58-864351e088d4

## 📋 Steps to Reproduce
Couldn't reproduce in UAT.
Need more investigation on production why this is happened.  

## 📸 Screenshot

## 🙋‍♀️ Proposed Solution

## ✅ Acceptance Criteria

- [ ] New job notification email should be sent only once.

petertgiles:
🤖 Resolves #11152 , #11153 

## 👋 Introduction

I believe the the reason some people got three notifications and some got zero is that the job that queues the notification for each user was timing out and retrying.  This branch sets the timeout to a longer value and disables retries.

## 🕵️ Details

I can see the job that Brinda was referring to in the prod `failed_jobs` table with a TimeoutException.  The default timeout for a job is 60 seconds, the wait to retry is 90 seconds, and our workers are set to 3 retries.  This all lines up with the timeline Brinda logged in the issue.

I seeded 23,000 users locally which is about how many we have who received this notification in prod.  I confirmed that the job fails after 60 seconds, having queued about 7400 jobs.  When I upped the timeout, the full set took a little over 3 minutes.  I picked a timeout of 10 minutes for this job which should be about 3 times as much as we need - assuming that it runs a similar speed in prod.

## 🧪 Testing

1. Start the queue worker
2. Launch Tinker
3. `User::factory()->count(23000)->create(['enabled_email_notifications'=>['JOB_ALERT'], 'enabled_in_app_notifications'=>['JOB_ALERT']])`
4. `$p = Pool::factory()->draft()->create()`
5. `$p->published_at = Carbon/Carbon::now()`
6. `$p->save()`

- [ ] Over 23000 new jobs queued in the `jobs` table
- [ ] The App\Listeners\SendNewJobPostedNotification completes successfully

## 📸 Screenshot

image

### DIFF
--- a/api/app/Listeners/SendNewJobPostedNotification.php
+++ b/api/app/Listeners/SendNewJobPostedNotification.php
@@ -14,6 +14,23 @@ use Throwable;
 class SendNewJobPostedNotification implements ShouldQueue
 {
     /**
+     * The number of seconds the job can run before timing out.
+     * Locally, I can queue about 7400 jobs per minute.
+     *
+     * @var int
+     */
+    public $timeout = 60 * 10;
+
+    /**
+     * The number of times the job may be attempted.
+     * Duplicate runs of this job will result in
+     * duplicate notifications being sent.
+     *
+     * @var int
+     */
+    public $tries = 1;
+
+    /**
      * Create the event listener.
      */
     public function __construct() {}


esizer (approved):
Nice :sunglasses:

petertgiles:
I tested this on dev and it queued about 4600 jobs per minute, so we should have double the needed time right now.

tristan-orourke:
Could we set this up to break the task into multiple jobs, dividing up the list of users if its hits a certain threshold? This pushes back the problem but still won't scale.

petertgiles:
> Could we set this up to break the task into multiple jobs, dividing up the list of users if its hits a certain threshold? This pushes back the problem but still won't scale.

Well, this job is really already that operation - a job that just queues a bunch of other jobs.  :laughing:   Then we would have a job that queues a job that queues more jobs.  

I don't think we can disable the timeout on this job but we could effectively do that by setting the timeout to 12 hours of something.

PR #11421, opened 2024-08-29
[Fix] Candidate category label and sorting

### LINKED ISSUE #10977: 🐛 Pool candidate category sorting and display inconsistent
## 🐛 Bug

Right now our "Category" property for pool candidates is handled inconsistently.  Sometimes the sorting uses a scope that takes validation into account, sometimes it doesn't, sometimes it handles it on the frotnend only.  Usually the sorting happens on the backend and calculating the display name happens on the frontend.  The whole system uses a numeric weight value that gets introduced into the business logic, making it harder to maintain.

## 🦋 Expected Behaviour

Sorting and naming should be happening on the backend.  Backend sorting is mandatory to support paginated queries and the naming should happen with the same logic.  Now that we have localized enums it should make the naming easier on the backend and we may even be able to get rid of the numeric weights.  The three big pool candidate tables (pool candidate search, talent placement, screening and assessment) should be using shared logic for sorting and naming category.

## 🕵️ Details

https://github.com/GCTC-NTGC/gc-digital-talent/issues/10901#issuecomment-2225733446

## 📋 Steps to Reproduce

1. Do a "CI seed" to get a mostly empty database
2. Create a "regular user" who is a citizen with no veteran or priority status.
3. Have the regular user apply to the CMO pool
4. Create a "special user" who is not a citizen but claims veteran and priority status.
5. Have the special user apply to the CMO pool
6. Login as an admin and view the two pool candidates in the CMO pool
7. Try accepting and rejecting the veteran and priority claims on the special user to observer category naming and sorting
8. Compare the three pool candidate tables

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

1. Align all three table queries on the same backend sorting logic for category
2. Return a localized enum for the category from the same backend logic that is driving the sorting
3. Remove numeric weight column from database?  Refactor to a dynamically updated enum?  What does this do to the orderByClaimVerification scope?
4. Ensure that the backend logic for category properly handles claim verficiation
5. Fix any bugs found in that logic, eg "fallback to citizen"
6. Write some phpunit tests around this
7. Delete any frontend code for calculating effective category sort or name

## ✅ Acceptance Criteria

- [ ] Category naming comes from the backend
- [ ] Category sorting comes from the backend
- [ ] Tables (pool candidate search, talent placement, screening and assessment) share same category logic in backend
- [ ] No "fallback to citizen" bug
- [ ] No "one rejection cancels the claim other" bug
- [ ] PHPUnit tests
- [ ] Remove frontend category naming and sorting
- [ ] Sorting should go like this
Bookmarked
Priority (unless rejected)
Veteran (unless rejected)
Citizen/Permanent resident
Everyone else

esizer:
🤖 Resolves #10977 

## 👋 Introduction

Adjusts the pool candidates table and assessment step tracker to use verification in the label and sorting.

## 🕵️ Details

I actually have no clue if this is correct. I'm still really struggling with when things should be sorted and how. I feel like there are too many conditions on this field. Some questions I have are:

1. We were not considering the expiry date in the sort, should we?
   a. If we do, veteran expiry is not required, should we ignore that and only care about priority?
2. We were always sorting by bookmarked when doing priority, this was super confusing when not showing the bookmarked icon and wasted like 2 hours of my time wondering why the order was incorrect
    a. Should we add another arg to claim sort to add bookmarked in when on a specific process only?
    
### Answers

1. We do not care about expiry dates for now
2. I went ahead and made sorting by bookmark optional since it was confusing on tables that do not use the bookmark

## 🧪 Testing

2. Navigate to all pool candidate listings
    - Candidate search
    - Talent placement (under process)
    - Assessment step tracker
3. Confirm the labels for category are accurate
4. Confirm the sorting is accurate


api/app/Models/PoolCandidate.php
```diff
@@ -844,32 +881,47 @@ public function scopeOrderByPoolName(Builder $query, ?array $args): Builder
...
-                ->select('users.citizenship', 'pool_candidates.*')
-                ->orderBy('is_bookmarked', 'DESC')
-                ->orderByRaw($order);
+            if (isset($args['useBookmark']) && $args['useBookmark']) {
+                $query->orderBy('is_bookmarked', 'DESC');
+            }
+
+            $order = sprintf('%s %s', $orderWithoutDirection, $args['order']);
+
+            $query->orderByRaw($order);
```
esizer, line 924: This sorts randomly for candidates who match both. Should we add a tertiary sort to keep it stable? I was thinking `updated_at` :thinking:

api/app/Models/PoolCandidate.php
```diff
@@ -273,6 +273,43 @@ public function educationRequirementExperiences(): BelongsToMany
...
+                $category = PriorityWeight::OTHER;
+                break;
+
+            default:
+                return null;
+
+        }
+
+        return [
+            'weight' => $category->weight($category->name),
```
brindasasi, line 307: Both return statements are technically correct, but you need to ensure that $category is not null before accessing its properties in the final return.

api/app/Models/PoolCandidate.php
brindasasi: Right.. we should break the tie with additional critiera .. I would like to think as who applied first should go first when there is a tie.. like submitted_at ascending. May be @desfam ?

brindasasi (changes requested):
Overall looks good .. 
minor thing to address and a question to design team. 
few more cases to test from my side
then it should be good to go.

api/app/Models/PoolCandidate.php
esizer: Category will always set though unless I am missing something? In the switch statement we set category and in the default case we simply return `null`.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/1ccafa91f174e80a3cc0e2e56eb64266bd6fbe67/api/app/Models/PoolCandidate.php#L301-L302

So, once we get to this line, `$category` is set since otherwise it would have fall through to the default case and been returned before reaching this line.

brindasasi:
when there is no bookmark, sort order for candidates doesn't look right on this one 🤔
Shouldn't veteran come second as the score would be 30 ? Veteran claim here is _unverified_. 
Screenshot 2024-09-16 at 4 18 09 PM

esizer:
> when there is no bookmark, sort order for candidates doesn't look right on this one 🤔
> Shouldn't veteran come second as the score would be 30 ? Veteran claim here is _unverified_.

Which page is this?

brindasasi:
> #10977

> > when there is no bookmark, sort order for candidates doesn't look right on this one 🤔
> > Shouldn't veteran come second as the score would be 30 ? Veteran claim here is _unverified_.
> 
> Which page is this?

This was on Talent Placement page for the IT-01: CMO Digital Careers pool in the local

esizer:
> > #10977
> 
> > > when there is no bookmark, sort order for candidates doesn't look right on this one 🤔
> > > Shouldn't veteran come second as the score would be 30 ? Veteran claim here is _unverified_.
> > 
> > 
> > Which page is this?
> 
> This was on Talent Placement page for the IT-01: CMO Digital Careers pool in the local

Do you still have that data seeded? I was unfortunately not able to reproduce what you have in your screenshot :cry:

brindasasi:
sorry about the delay 
Responses from Jerbo 
1. When there is tie on the candidates inside the specific process,  sort by submitted_at asc 
2. Default sorting on the table of candidates inside a specific process - sort by final decision 
3. Default sorting on the table of ALL candidates - sort by DateReceived desc

esizer:
> sorry about the delay Responses from Jerbo
> 
> 1. When there is tie on the candidates inside the specific process,  sort by submitted_at asc
> 2. Default sorting on the table of candidates inside a specific process - sort by final decision
> 3. Default sorting on the table of ALL candidates - sort by DateReceived desc

okay, updatd sort order based on the information provided.

[update sort order](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11421/commits/22fc0d9b4905ab87a40a5c9147a56c9ef1d5cb84)

brindasasi:
just for the log , I'm putting it out here 
When there is a claim rejection, it messes up with the category sorting. 
First screen shot , admin had veteran claim and had that order in screenshot. I went ahead and rejected that claim. He still sits in the same position in between Veterans 
Screenshot 2024-09-19 at 3 26 17 PM

I've 2 candidates with rejected claims 
Screenshot 2024-09-19 at 3 35 04 PM
Screenshot 2024-09-19 at 3 34 51 PM
But they show up in different order 


Screenshot 2024-09-19 at 3 38 03 PM

esizer:
> just for the log , I'm putting it out here
> When there is a claim rejection, it messes up with the category sorting.
> First screen shot , admin had veteran claim and had that order in screenshot. I went ahead and rejected that claim. He still sits in the same position in between Veterans

Thanks for all the help with this, I think we finally got to the bottom of it and it should be working as expected now.

[fix category attribute](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11421/commits/83fd5fc13f3e697f98726523ccc59c4372ac8e55)

brindasasi (approved):
LGTM! 🥳👏🙌

PR #11428, opened 2024-08-30
[Debt] Move file generation auth to job

### LINKED ISSUE #11234: ♻️ Apply authorized scope in file generators
## ♻️ Debt/Refactor

In order to offload as much procesing to background jobs, we would like to move the authorized scope for downloaded files to the generators from the mutation request.

## 🕵️ Details

We could not do this previously since our user checker was returning false if it was run outside the context of a request. Now that this has been fixed in https://github.com/GCTC-NTGC/gc-digital-talent/pull/11228 we can move this check off to the generate so the query is only applied as part of a queued job.

## 🙋‍♀️ Proposed Solution

1. Update the `authorizedToView` scopes in `PoolCandidate` and `User` to accept args and override the user being checked if it is passed
2. Remove the scope queries in the download mutations
3. Move the scope, passing in the id of the user who triggered the mutation

> [!NOTE]
> Not sure if we should add user ID to any file generator or just specific ones :thinking: 

```php
// Generate query
->authorizedToView(['userId' => $this->userId])

// Authorized to view scope
$user = Auth::user();

$userId = $args['userId'] ?? null;
if($userId) {
  $user = User::find($userId);
}
```

## ✅ Acceptance Criteria

- [ ] File generator classes apply the `authorizedToView` scopes as part of their queries
- [ ] Authorized to view queries are not run in the mutation requests
- [ ] Confirm files still download all expected data

esizer:
🤖 Resolves #11234 

## 👋 Introduction

Moves the authorized to view scope to be run in the job rather than mutations for file generation.

## 🧪 Testing

3. Download users and candidates from the tables
4. Confirm you still get the expected results
5. Run a mutation with a reduced permission user and IDs for users you do not have access to
6. Confirm those users do not appear in the generated file

### DIFF
--- a/api/app/Generators/ApplicationDocGenerator.php
+++ b/api/app/Generators/ApplicationDocGenerator.php
@@ -36,6 +36,7 @@ class ApplicationDocGenerator extends DocGenerator implements FileGeneratorInter
             'generalQuestionResponses' => ['generalQuestion'],
         ])
             ->whereIn('id', $this->ids)
+            ->authorizedToView(['userId' => $this->userId])
             ->chunk(200, function ($candidates) use ($section) {
 
                 foreach ($candidates as $candidate) {
--- a/api/app/Generators/PoolCandidateUserDocGenerator.php
+++ b/api/app/Generators/PoolCandidateUserDocGenerator.php
@@ -29,6 +29,7 @@ class PoolCandidateUserDocGenerator extends DocGenerator implements FileGenerato
             'generalQuestionResponses' => ['generalQuestion'],
         ])
             ->whereIn('id', $this->ids)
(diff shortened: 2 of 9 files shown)


vd1992 (changes requested):
Downloads failing/erroring on view candidate page

image

Sample stack. think it has to do with changes to the mutation classes

image

esizer:
> Downloads failing/erroring on view candidate page

Fixed in [update with new user id setter](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11428/commits/fe6daf04a6f9a7bae058a657e61bb9701df1dd04)

vd1992 (approved):
Working through the pages and it all seems fine to me 💯

Issue #9905, opened 2024-03-26
♻️ Remove snapshot workaround: bilingual evaluation


yonikid15:
> [!Warning]
:stop_sign: Blocked until Mar 26, 2025 (one year after workaround added)

## ♻️ Debt/Refactor

A workaround was added to the snapshot to allow the deprecated bilingual evaluation in a snapshot to be viewed.  One year later, let's remove the workaround.

## 🕵️ Details

♻️ Rework the candidate facing language form
[#8031](https://github.com/GCTC-NTGC/gc-digital-talent/issues/8031)

## ✅ Acceptance Criteria

- [ ] Workaround removed
- [ ] Current snapshots still render new language profile changes correctly

petertgiles:
Oh yeah, probably not.  👍

PR #11433, opened 2024-09-03
[Feat] Version snapshots

### LINKED ISSUE #10913: ✨ Version snapshots
## ✨ Feature

Add some sort of versioning to our pool candidate snapshots to more easily handle legacy data.

## 🕵️ Details

Our snapshots are drifting pretty far as our schema changes and it is getting a little difficult to reason about that data in a snapshot is legacy and what we should expect it to contain. I think adding a version to these snapshots could make things much easier.

This way, we could indicate the version on places where the snapshot is rendered to at least make it a little more identifiable. Since recently we have had some items where bugs were reported due to missing or incorrect data in a snapshot actually being related to a legacy snapshot rather than an actual bug. With a version number, we could likely troubleshoot this quicker.

Also, with a version number, we can be more confident about what data each version should contain which should make rendering the actual information easier. 

### Rendering example

```tsx
const snapshot = parseSnapshot(poolCandidate.profileSnapshot);
const { version } = snapshot;

if(version === 1) return <SnapshotV1 snapshot={snapshot} />;

type ProfileSnapshotV1 = { ...expectedFields };

interface SnapshotV1Props { 
  snapshot: ProfileSnapshotV1;
}
```

This depends on how we want to do it. We first need to decide if we want to attempt to retroactivley version older snapshots. If so, we would need to find each PR that modified the snapshot and likely write an artisan command to find snapshots created between deployments of said PRs and inject a version somewhere.

Or, we can just treat everything without a version unstable and use our current approach for that and only start versioning forward.

As for storing the version, I'd just add it as a property in the actual JSON of the snapshot and we can maintain a config value of the current version and somehow enforce someway for bumping it as needed? :thinking: 

```php
$SNAPSHOT_VERSION = 2;

$candidate->profile_snapshot = [ 'version' => $SNAPSHOT_VERSION, ....$snapshot];
```

## ✅ Acceptance Criteria

- [ ] Profile snapshots contain a version number
- [ ] Version number is handled when rendering in:
    - [ ] Admin view pool candidate page
    - [ ] Application document download
    - [ ] Application CSV download

esizer:
🤖 Resolves #10913 

## 👋 Introduction

Adds a version number to snapshots and sets the initial value to 1.

## 🕵️ Details

Uses [value object casting](https://laravel.com/docs/10.x/eloquent-mutators#value-object-casting) to cast the snapshot for more complicated getters and setters and adds the version as a static property on it to be added in the setter.

## 🧪 Testing

1. Login as any user
2. Apply to a job completing and submitting to generate snapshot
3. Open the database record for the application that was submitted in step 2
4. View the snapshot field to confirm the JSON contains a version number

### DIFF
--- a/api/app/Http/Resources/UserResource.php
+++ b/api/app/Http/Resources/UserResource.php
@@ -22,6 +22,8 @@ class UserResource extends JsonResource
 
     protected $poolSkillIds;
 
+    public int $version;
+
     public function poolSkillIds($value)
     {
         $this->poolSkillIds = $value;
@@ -64,6 +66,7 @@ class UserResource extends JsonResource
         }
 
         return [
+            'version' => $this->version,
             'id' => $this->id,
             'sub' => $this->sub,
             'firstName' => $this->first_name,
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -2,7 +2,6 @@
 
 namespace App\Models;
 
-use App\Enums\ArmedForcesStatus;
 use App\Enums\AssessmentDecision;
 use App\Enums\AssessmentResultType;
 use App\Enums\AssessmentStepType;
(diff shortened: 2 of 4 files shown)


petertgiles:
I'm just reading through this - so cool!  I suppose the next step is make a new snapshot rendering component especially for version 1 snapshots?  It could essentially be the same as "version 0" but with any workarounds removed (https://github.com/GCTC-NTGC/gc-digital-talent/issues/9901 and https://github.com/GCTC-NTGC/gc-digital-talent/issues/9905 and maybe others?).  Is that too much scope creep for this PR?  Could you create a new issue for that?

esizer:
> I'm just reading through this - so cool! I suppose the next step is make a new snapshot rendering component especially for version 1 snapshots? It could essentially be the same as "version 0" but with any workarounds removed (#9901 and #9905 and maybe others?). Is that too much scope creep for this PR? Could you create a new issue for that?

Yeah, I was actually a little uncertain about that :sweat_smile: . Our current component that renders it is technically v1 since I'm not sure we can retroactively apply versions. Maybe we can set the version to `null` or `0` for now and create a new ticket to handle when we have a use case for versions? :thinking: 

I was going to do the frontend work right now but it would seem weird to do a single use case. I was assuming it would look something like:

```tsx
switch(snapshotVersion) {
  case 1:
     return <SnapshotV1 />;
 case 2:
    return <SnapshotV2 />;
 default:
   return <p>Some error message.</p>;
}
```

But as it stands, we only have one case. Though, maybe that isn't an awful idea? :thinking: Unless you saw this working some other way?

esizer:
> We do have essentially two versions that we can do, right? Anything with no version number (version 0) needs to be compatible with old snapshots as much as possible and needs the existing workarounds. But version 1 snapshots were created "today or later" and can have all existing workarounds removed.

Right. Sorry, I totally misread the original comment :sweat_smile: I think that is reasonable and can work on getting it in.

tristan-orourke:
I think the original ticket this is for, #10913, would seem to include frontend work, but yeah I recommend breaking that into a separate ticket. This is a very nice PR now as is.

petertgiles (approved):
Looks great!

PR #11437, opened 2024-09-03
[Feature] Last active at tracking

### LINKED ISSUE #11426: ✨ Track last sign-in for all users
## ✨ Feature

We want to know the number of active users on the site. The measure we intend to use for this is whether or not they have signed in to the site. So let's start collecting that!

## 🕵️ Details

Add a field that stores the timestamp of someone successfully signing into the site for all users.

vd1992:
🤖 Resolves #11426 

## 👋 Introduction

Adds a field to track last active and handles updating it through an authorization callback

## 🧪 Testing

1. Migration works
2. Open up database tool and check logging in updates the new field

## 📸 Screenshot

image

### DIFF
--- a/api/app/Providers/AuthServiceProvider.php
+++ b/api/app/Providers/AuthServiceProvider.php
@@ -21,6 +21,7 @@ use App\Policies\PoolCandidatePolicy;
 use App\Policies\PoolPolicy;
 use App\Policies\UserPolicy;
 use App\Services\OpenIdBearerTokenService;
+use Carbon\Carbon;
 use Illuminate\Support\Facades\Gate;
 use Illuminate\Support\ServiceProvider;
 use Lcobucci\JWT\Validation\RequiredConstraintsViolated;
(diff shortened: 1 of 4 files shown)


api/app/Providers/AuthServiceProvider.php
petertgiles, line 1: Hmm, I'm not sure this is the best place for the timestamp update.  This gets called for every request so the timestamp is continually being updated - not just when they log in.  I recommend moving this over to the auth callback so it only occurs on login.  You can see that the token that we receive and decode there does have the sub available, just like we grab the nonce.
image

petertgiles (changes requested):
Looking good!  I suggested a change to the implementation to get a little closer to the AC.

api/app/Providers/AuthServiceProvider.php
vd1992: Oh, there actually is something called auth callback specifically and it wasn't meant in the general sense 😊 
Tweaked in https://github.com/GCTC-NTGC/gc-digital-talent/pull/11437/commits/91130bdc427a6eddcc0a988b1cf990c7012849f1
Wasn't sure how to test it

petertgiles (commented):
Ooh, a bit stuck now...

api/app/Providers/AuthServiceProvider.php
petertgiles: OK, decision in standup is that this is good enough for now.  I'll create a new issue to adjust this.

PR #11474, opened 2024-09-06
[Bugfix] Fix application document generator on old snapshots

### LINKED ISSUE #11443: 🐛 Download of users application error
## 🐛 Bug

When downloading a specific user, we get an error message.

## 🦋 Expected Behaviour

A download of the application.

## 🕵️ Details

Will need to comb through Azure logs, find the error, and decide the solution appropriately. Could be an unexpected field nulled, could be a broken column value

May be causing https://github.com/GCTC-NTGC/gc-digital-talent/issues/11442 ? Same user at play possibly?

user email: Please ask Marc via MS Teams or email.

## 📸 Screenshot
image

## ✅ Acceptance Criteria

- [ ]  Problem identified and solution implemented

petertgiles:
🤖 Resolves #11443

## 👋 Introduction

This branch fixes a bug where "old style" snapshots with raw enum values in them couldn't be rendered by the document renderer expecting "new style" snapshots with the localized enums.

## 🧪 Testing

1. Check out the commit before server-side enums: 7a7e5e4697cc1ab5122d0c481408885502a438e2
2. Submit an application to generate an "old style" snapshot
3. Record the applicant "review" page of the application
4. Record the admin "application" page of the application
5. Record the "print application" pdf of the application
6. Check out the head of this branch
7. Compare the application "review" page to the one from before
8. Compare the admin "application" page to the one from before
9. Compare the generated application document to the printed pdf from before
10. Submit a new applicaton to generate a "new style" snapshot and make sure that can be rendered as a document as well


api/app/Traits/HydratesSnapshot.php
```diff
@@ -19,20 +20,46 @@ trait HydratesSnapshot
...
+    {
+        // validator for a single localized enum
+        $singleEnumValidator = Validator::make($snapshot, [
+            $snapshotField.'.value' => 'required|string',
+            $snapshotField.'.label.en' => 'required|string',
+            $snapshotField.'.label.fr' => 'required|string',
+        ]);
+        if ($singleEnumValidator->passes()) {
+            return true;
+        }
```
esizer, line 33: This may lead to unexpected behvaiour if we have null localized enum values (idk how likely this is) but:

```php
# Given
'key' => ['value' => null, 'label' => nulll]

# Validator will fail and hydrate to
 'key' => ['value' => null]
 
# instead of
'key' => null
```

Side note is there a reason we are using a validator here? We are really only checking keys so we don't need all the extra stuff that comes with a validator. Though, if we find it easier to reason about validators, we are using them in a similar way in other places so I guess it isn't a huge deal :woman_shrugging: 

```php
// There is probably also a better way to do this :sweat_smile: 
if(is_array($snapshotField) {

  $dot = Arr::dot($snapshotField);
  $expectedKeys = ['value' => '', 'label.en' => '', 'label.fr' => ''];
  
  if(array_is_list($snapshotField) {
     return array_reduce($dot, function ($hasKeys, $field) use($expectedKeys) {
        return $hasKeys && empty(array_diff_key($expectedKeys, $field));
     }, true);
  } else {
     return empty(array_diff_key($expectedKeys, $field));
  }
}
  
return false
```

esizer (commented):
Looking good! I had some questions about determining if something is localized though.

api/app/Traits/HydratesSnapshot.php
petertgiles: For the second point, yeah, it's readability.  It probably is overkill and a bit slow but I find it much much easier to read and reason about than the code you have above.

For the first point, yeah, let me see what I can do about it.

petertgiles: Right, then `en` and `fr` fields could be null, but `label` and `value` can't be.  I've updated the validators and tests to handle that.
[handle null label strings](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11474/commits/0908d468bf5be46f5ece909b72433401ebd48509)

esizer (approved):
Cool :sunglasses:

PR #11475, opened 2024-09-06
[Feature] Single document download filename

### LINKED ISSUE #11445: 🔤 Filename for single candidate profile & application
## 🔤 Copy

Update the Filename for the single candidate profile and application (snapshot) downloads.

For profile the new filename should be (regardless of the language of the file):

`firstname lastname - Profile - Profil`

Example: **Gray O'Byrne - Profile - Profil.docx**

For the application the new filename should be  (regardless of the language of the file):

`firstname lastname - Application - Candidature`

## ✅ Acceptance Criteria

- [ ] New filenames for individual Profile download
- [ ] New filenames for individual Application (snapshot) download
- [ ] New filenames for anonymous application downloads (last name replaced with initial)
- [ ] Sanitize name strings for legal filename characters
- [ ] Test resulting files with downloading and pruning - ensure spaces in filenames are okay - replace with underscores if not

vd1992:
🤖 Resolves #11445 

## 👋 Introduction

Adjust filename generation to desired format

## 🧪 Testing

1. Download some documents
2. Observe the cases: profile, anonymous profile, and application
3. Try and find edge cases to be cautious of?

## 📸 Screenshot

image

image

### DIFF
--- a/api/app/GraphQL/Mutations/DownloadApplicationDoc.php
+++ b/api/app/GraphQL/Mutations/DownloadApplicationDoc.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace App\GraphQL\Mutations;
 
 use App\Generators\ApplicationDocGenerator;
+use App\Models\PoolCandidate;
 use Illuminate\Support\Facades\App;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Log;
@@ -18,9 +19,19 @@ final readonly class DownloadApplicationDoc
         $user = Auth::user();
         throw_unless(is_string($user?->id), UnauthorizedException::class);
 
+        $targetApplicant = PoolCandidate::find($args['id'])->load(['user']);
+        $firstName = $targetApplicant?->user?->first_name;
+        $lastName = $targetApplicant?->user?->last_name;
+        if (isset($firstName)) {
+            $firstName = trim(filter_var($firstName, FILTER_SANITIZE_EMAIL));
+        }
+        if (isset($lastName)) {
+            $lastName = trim(filter_var($lastName, FILTER_SANITIZE_EMAIL));
+        }
+
         try {
 
-            $fileName = sprintf('%s_%s.docx', __('filename.candidate'), date('Y-m-d_His'));
+            $fileName = sprintf('%s - %s - Application - Candidature.docx', $firstName ? $firstName : '', $lastName ? $lastName : '');
 
             $generator = new ApplicationDocGenerator(
                 ids: [$args['id']],
--- a/api/app/GraphQL/Mutations/DownloadApplicationDoc.php
+++ b/api/app/GraphQL/Mutations/DownloadApplicationDoc.php
@@ -23,15 +23,19 @@ final readonly class DownloadApplicationDoc
         $firstName = $targetApplicant?->user?->first_name;
         $lastName = $targetApplicant?->user?->last_name;
         if (isset($firstName)) {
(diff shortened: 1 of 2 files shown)


api/app/GraphQL/Mutations/DownloadApplicationDoc.php
```diff
@@ -18,9 +19,19 @@ public function __invoke(null $_, array $args)
...
+        $targetApplicant = PoolCandidate::find($args['id'])->load(['user']);
+        $firstName = $targetApplicant?->user?->first_name;
+        $lastName = $targetApplicant?->user?->last_name;
+        if (isset($firstName)) {
+            $firstName = trim(filter_var($firstName, FILTER_SANITIZE_EMAIL));
+        }
+        if (isset($lastName)) {
+            $lastName = trim(filter_var($lastName, FILTER_SANITIZE_EMAIL));
+        }
+
```
esizer, line 31: I don't think `FILTER_SANITIZE_EMAIL` is sufficient for file names. Here is a [list of potentially illegal characters.](https://www.mtu.edu/umc/services/websites/writing/characters-avoid/)

The filter being used here [allows characters in that list](https://www.php.net/manual/en/filter.filters.sanitize.php#:~:text=Remove%20all%20characters%20except%20letters%2C%20digits%20and%20!%23%24%25%26%27*%2B%2D%3D%3F%5E_%60%7B%7C%7D~%40.%5B%5D.).

api/app/GraphQL/Mutations/DownloadUserDoc.php
```diff
@@ -18,8 +19,21 @@ public function __invoke(null $_, array $args)
...
+        if (isset($lastName)) {
+            $lastName = trim(filter_var($lastName, FILTER_SANITIZE_EMAIL));
+        }
+
         try {
-            $fileName = sprintf('%s_%s.docx', trim(__('filename.user')), date('Y-m-d_His'));
+
+            $fileName = $args['anonymous'] ?
+                sprintf('%s - %s - Profile - Profil.docx', $firstName ? $firstName : '', $lastName ? substr($lastName, 0, 1) : '') :
+                sprintf('%s - %s - Profile - Profil.docx', $firstName ? $firstName : '', $lastName ? $lastName : '');
```
esizer, line 36: I have a minor concern that may not be an issue. But, we no longer have any unique identifier so if an admin downloads a file twice within the window for pruning, it may overwrite the previously file if there were changes.

Do we care about this or is it expected that we will just maintain the most recent file in that window?

esizer (changes requested):

api/app/GraphQL/Mutations/DownloadUserDoc.php
vd1992: I have no idea

api/app/GraphQL/Mutations/DownloadApplicationDoc.php
vd1992: Huh, okay, how about https://github.com/GCTC-NTGC/gc-digital-talent/pull/11475/commits/196836af043d34227e9f1dee9e4222d82605909c

image

image

esizer: So, I know the link I sent said you shouldn't use spaces and while that is true, it was requested to put spaces in the file name :sweat_smile: So, we should keep those even if it is not best practice to do so

vd1992: 😆 
Okay, back in they go https://github.com/GCTC-NTGC/gc-digital-talent/pull/11475/commits/7baf764bf9702c92b5468fead72dab2454c587ea

PR #11483, opened 2024-09-09
[Debt] Add pool eloquent builder class

### LINKED ISSUE #11231: ♻️ Use custom query builders
## ♻️ Debt/Refactor

Implement some custom builders for larger models to make them more manageable. 

## 🕵️ Details

Some models are getting a little large (`PoolCandidate` is 1293 lines). While this isn't that large, they are getting a little disorganized. We sort of just stick code where ever and it is getting a bit difficult to reason about them.

A good first step to organize this a bit would be to implement some custom builders and move the query builder logic there (pretty much just scopes).

We don't need to do this for all models but `PoolCandidate` and `User` would make sense to do since they are our largest models.

## 🙋‍♀️ Proposed Solution

It is pretty straight forward to do and you can follow [this article](https://martinjoo.dev/build-your-own-laravel-query-builders) to do it.

Essentially,

1. Create builders (probably in the `App\Builders` namespace)
2. Move the model scopes there removing the `scope` prefix ensuring you still use camelCase
3. Set the custom builder on the model in the `newEloquentBuilder` method

## ✅ Acceptance Criteria

- [ ] Move scopes for larger models (`PoolCandidate`, `User` and maybe `Pool`?)
- [ ] Ensure they still function with `lighthouse-php` (I don't foresee any issues but pays to be careful)

esizer:
🤖 Resolves #11231 

## 👋 Introduction

Adds a specific eloquent builder for the pool model and moves scopes there.

## 🕵️ Details

I attempted to keep the naming as close as possible to what it is now but had to make some minor changes. Some scopes are prefixed with SQL like keywords, matching built in builder methods. I did this in an attempted to avoid name collisions with relationships (i.e `classifications`).

I made this changes only where needed to keep the scope smaller but maybe we want to use the keyword prefix for other methods in the future (`whereInPublishingGroup` vs `publishingGroups`). I think this makes the intention of the methods more clear.

## 🧪 Testing

1. Confirm existing scope tests are passing
2. Interact with places the scopes are used (mainly browse jobs and pools table)
3. Confirm they still function as expected

### DIFF
--- a/api/app/Models/Pool.php
+++ b/api/app/Models/Pool.php
@@ -404,14 +404,6 @@ class Pool extends Model
         return $this->team?->id;
     }
 
-    public static function scopeCurrentlyActive(Builder $query)
-    {
-        $query->where('published_at', '<=', Carbon::now()->toDateTimeString())
-            ->where('closing_date', '>', Carbon::now()->toDateTimeString());
-
-        return $query;
-    }
-
     public static function scopeName(Builder $query, ?string $name): Builder
     {
         if ($name) {
--- a/api/app/Models/Pool.php
+++ b/api/app/Models/Pool.php
@@ -404,110 +404,16 @@ class Pool extends Model
         return $this->team?->id;
     }
 
-    public static function scopeName(Builder $query, ?string $name): Builder
-    {
-        if ($name) {
-            $query->where(function ($query) use ($name) {
-                $term = sprintf('%%%s%%', $name);
-
-                return $query->where('name->en', 'ilike', $term)
-                    ->orWhere('name->fr', 'ilike', $term);
-            });
-        }
-
-        return $query;
-    }
-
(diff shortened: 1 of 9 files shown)


api/app/Models/Pool.php
```diff
@@ -162,6 +161,11 @@ protected static function boot()
         });
     }
 
+    public function newEloquentBuilder($query): PoolBuilder
```
petertgiles, line 164: This is what actually binds the new builder to the model, right?  Could you add a comment describing this?

petertgiles (approved):
I like this!  I think it does improve readability and maintainability.  I'm not sure how to thoroughly test it though.  I guess we find out how good our test suite is.  :crossed_fingers:

tristan-orourke:
Looks good to me

PR #11487, opened 2024-09-10
[Debt] Make process tests parallel

### LINKED ISSUE #10528: 🧪 Avoid need for serial mode in process actions test
## 🧪 Tests

In the `process-actions.spec.ts` test, we need to use serial mode since the tests rely on the previous step being completed. This causes the test to take along time since cannot be run in parallel. It would be nice to update this test so it can be run in parallel.

## 🙋‍♀️ Proposed Solution

We just need to make it so we are not updating/deleting the pool created in the first test and use one off pools for each test.

## ✅ Acceptance Criteria

- [ ] Test coverage remains the same
- [ ] Serial mode removed

esizer:
🤖 Resolves #10528 

## 👋 Introduction

Adjusts the process actions test to no longer require serial mode.

## 🧪 Testing

1. Run the test `pnpm run e2e:playwright:chromium -- --process-actions`
2. Confirm it passes
3. Confirm it still has the same coverage

### DIFF
--- a/apps/playwright/tests/admin/process-actions.spec.ts
+++ b/apps/playwright/tests/admin/process-actions.spec.ts
@@ -1,13 +1,13 @@
 import { test, expect } from "~/fixtures";
+import graphql from "~/utils/graphql";
+import { createPool } from "~/utils/pools";
+import { me } from "~/utils/user";
 
 const UPDATE_MUTATION = "UpdatePool";
 
 /**
  * Actions associated with processes
- *
- * Note: Separate so they can be run in serial mode
  */
-test.describe.configure({ mode: "serial" });
 test.describe("Process actions", () => {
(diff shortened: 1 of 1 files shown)


vd1992:
I don't think it had any impact on test speed? Just went a little slower maybe?

Run off main https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/10795600332/job/29942667797

image

Run off branch https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/10797837253/job/29949971520

image

esizer:
> I don't think it had any impact on test speed? Just went a little slower maybe?

That is because we don't use parallel mode in CI :sweat_smile:

vd1992 (approved):
image

Issue #11143, opened 2024-08-02
🐛 [PROD] Applicants cannot access their applications


tristan-orourke:
## 🐛 Bug

Applicants are unable to see their draft applications or their completed applications on their dashboard. 

## 🦋 Expected Behaviour

Applicants should be able to access their applications.

## 🕵️ Details

Reported by users this morning.
Reproduced on local.

## 📋 Steps to Reproduce

1. Log in as applicant@test.com, which should have seeded applicaitons.
2. See null state on application lists on dashboard

## 📸 Screenshot

Applicant Dashboard shows no applications, even if database and admin dashboard show that user does have poolCandidate objects.
![Image](https://github.com/user-attachments/assets/d97bd1a0-bbbe-489b-a7ad-fdec870bb870)

## 🙋‍♀️ Proposed Solution

The issue seems to be in PoolCandidate.scopeAuthorizedToView. It causes 0 results to be returned for candidates.

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the bug was addressed.

- [ ] Applicants can access their applications.
- [ ] This PR should be as minimal as possible, in order to be deployed quickly

PR #11491, opened 2024-09-11
[Tests] Regression tests for application permissions 

### LINKED ISSUE #11147: 🧪 Test that Applicants can access their pool candidates from API on public endpoint
## 🧪 Tests

Test that applicants can get their applications from the API on public endpoint. 

## 🕵️ Details

Essentially, this is a regression test for #11143. The test should fail if you revert #11145.

## 🙋‍♀️ Proposed Solution

I think a PHPUnit feature test is the right call here. We need to test that a user with only the Base and Applicant roles, can run the query `me { poolCandidates { id } }` on the public endpoint and get an array of all their poolCandidates.

## ✅ Acceptance Criteria

- [ ] Automated test which tests that user with Applicant role can query `me { poolCandidates { id } }` on public endpoint and see all of their results
- [ ] If you revert the commit from #11145, the test fails.
- [ ] Playwright test that applicant can see their applications on their dashboard.

esizer:
🤖 Resolves #11147 

## 👋 Introduction

Adds a regression test for permissions of applications on an unprotected endpoint.

## 🕵️ Details

We fixed an error where applications could not view their own applications (https://github.com/GCTC-NTGC/gc-digital-talent/issues/11143) and fixed it in https://github.com/GCTC-NTGC/gc-digital-talent/pull/11145. This adds tests to prevent any regression of that fix.

## 🧪 Testing

1. Confirm added tests meet AC
2. Confirm tests pass reliably
3. Add the code provided to the pool candidate authorization scope
4. Confirm tests fail

```php
// PoolCandidate::scopeAuthorizedToView
 $hasSomePermission = $user->isAbleTo([
   'view-own-application',
   'view-team-submittedApplication',
   'view-any-submittedApplication',
  ]);
  
  if (! $hasSomePermission) {
    $query->where('id', null);
    
    return;
}
```

### DIFF
--- a/api/tests/Feature/UnprotectedApplicationTest.php
+++ b/api/tests/Feature/UnprotectedApplicationTest.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Tests\Feature;
+
+use App\Models\PoolCandidate;
+use App\Models\User;
+use Database\Seeders\RolePermissionSeeder;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Nuwave\Lighthouse\Testing\MakesGraphQLRequests;
+use Nuwave\Lighthouse\Testing\RefreshesSchemaCache;
+use Tests\TestCase;
+use Tests\UsesUnprotectedGraphqlEndpoint;
+
+class UnprotectedApplicationTest extends TestCase
+{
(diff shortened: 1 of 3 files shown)


petertgiles (approved):
Nice!  :heart:

PR #11492, opened 2024-09-11
[Test] Update experience skill pivot assertions

### LINKED ISSUE #7483: 🧪 `ExperienceTest.php` intermittent failure
## ♻️ Debt/Refactor

Intermittent test failure. 

## 🕵️ Details

Test `testSkillRelationshipsWorkWithPivot` in `ExperienceTest.php`

image

## ✅ Acceptance Criteria

- [ ] Test suite passes reliably (5+ runs) in CI and locally

esizer:
🤖 Resolves #7483 

## 👋 Introduction

Updates the assertions for the test to check each fragment to avoid order causing errors.

## 🕵️ Details

I don't believe we care about the order or anything so this just checks to see if each skill fragment exists. This should stabilize the test a bit more.

## 🧪 Testing

`for i in {1..10}; do make artisan CMD="test --filter=testSkillRelationshipsWorkWithPivot ./tests/Feature/ExperienceTest.php"; done`

1. Run the test multiple times
2. Confirm no failures

### DIFF
--- a/api/tests/Feature/ExperienceTest.php
+++ b/api/tests/Feature/ExperienceTest.php
@@ -65,7 +65,7 @@ class ExperienceTest extends TestCase
             $userSkills[2]->id => ['details' => 'third skill'],
         ]);
 
-        $this->actingAs($this->platformAdmin, 'api')->graphQL(<<<'GRAPHQL'
+        $response = $this->actingAs($this->platformAdmin, 'api')->graphQL(<<<'GRAPHQL'
             query getUser($id: UUID!) {
                 user(id: $id) {
                     workExperiences {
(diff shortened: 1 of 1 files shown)


api/tests/Feature/ExperienceTest.php
```diff
@@ -66,9 +65,7 @@ public function testSkillRelationshipsWorkWithPivot(): void
             $userSkills[2]->id => ['details' => 'third skill'],
         ]);
 
-        $response = $this->actingAs($this->platformAdmin, 'api')->graphQL(
-            /** @lang GraphQL */
-            '
+        $this->actingAs($this->platformAdmin, 'api')->graphQL(<<<'GRAPHQL'
             query getUser($id: UUID!) {
```
petertgiles, line 69: just curious why you changed this formatting?

api/tests/Feature/ExperienceTest.php
```diff
@@ -82,18 +79,26 @@ public function testSkillRelationshipsWorkWithPivot(): void
...
+            'id' => $userSkills[1]->skill->id,
+            'experienceSkillRecord' => [
+                'details' => 'second skill',
+            ],
+        ])->assertJsonFragment([
+            'id' => $userSkills[2]->skill->id,
+            'experienceSkillRecord' => [
+                'details' => 'third skill',
+            ],
+        ]);
```
petertgiles, line 101: We've lost the assertion that there are three skills.  The test will now miss if, for example, four skills come back.

petertgiles: An alternate approach might be ot use assertEqualCanonicalizing which will sort the arrays before comparison.
https://www.geeksforgeeks.org/phpunit-assertequalscanonicalizing-function/

esizer: great suggestion! Updated in [update to use assert canonicalizing](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11492/commits/317403eeda9b769731b421e4da102f406723e622)

api/tests/Feature/ExperienceTest.php
petertgiles: Huh, nice.  I didn't even notice that the graphql formatting wasn't working.  With this new style I get errors at least.   :+1:

petertgiles (approved):

PR #11494, opened 2024-09-11
[Fix] Show skill details in document downloads

### LINKED ISSUE #11488: 🐛 Skill justifications do not appear in downloaded (word) profile
## 🐛 Bug

The skill justifications are not showing up in the word document download of the profile.

## 🕵️ Details

This is the profile on the site. See the skill justifications under each skill.
image

This is the downloaded profile. See NO skill justifications under each skill.
image

## 🙋‍♀️ Proposed Solution

Add skill justifications to the experience section of the profile download.

## ✅ Acceptance Criteria

- [ ] Skill justifications show up in the experience section of the application and profile downloads

esizer:
🤖 Resolves #11488 

## 👋 Introduction

Fixes an issue where skill details were not being displayed in profile and application downloads.

## 🕵️ Details

Looks like there were two issues,

1. We were trying to get details from the skill but they exist on the user skill
2. Not all skill pivots had the same alias so they would not appear

## 🧪 Testing

2. Apply to a process making sure you have one of each experience type with skill details
3. Navigate to the candidate admin page for the application in step 2
4. Download both a profile and application document
5. Confirm the details for skills on all experience types appear

### DIFF
--- a/api/app/Models/UserSkill.php
+++ b/api/app/Models/UserSkill.php
@@ -124,7 +124,7 @@ class UserSkill extends Model
             ->withTimestamps()
             ->withPivot(['details', 'deleted_at'])
             ->wherePivotNull('deleted_at')
-            ->as('experience_skill_pivot');
+            ->as('experience_skill');
     }
 
     public function experiences()
@@ -138,6 +138,6 @@ class UserSkill extends Model
             ->withTimestamps()
             ->withPivot(['details', 'deleted_at'])
             ->wherePivotNull('deleted_at')
-            ->as('experience_skill_pivot');
+            ->as('experience_skill');
     }
 }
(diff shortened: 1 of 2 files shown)


vd1992:
The user doc download now has them, but I checked the CSV too and it was missing from their. This might have been pre-existing and therefore unrelated

Missing skills details as well as general/screening responses

image

I think I found the issue, if it is the cause below then it is pre-existing and maybe this gets spun off into a separate priority issue? Or scope creep this a bit cause it might be an easy fix?
The wrong value is being accessed by the for loop. Note this also applies to screening questions. 

This

image

Results in 

image

This

image

Results in 

image

esizer:
> I think I found the issue, if it is the cause below then it is pre-existing and maybe this gets spun off into a separate priority issue? Or scope creep this a bit cause it might be an easy fix?

That is a good catch. I tihnk I found out what it is and it should probably be spun out into its own issue.

The issue is, for the three items:

- General questions
- Screening questions
- Pool skill details

We use the iteration of the candidates to build a list of pool ids. Then, later on we iteratate over that list of pool ids to build the headers and append them onto the end. However, while looping over the pool ids, we build a list of ids for the three items to make sure we get the right items in the candidate iterator.

So, as you noticed, we have an empty array because it hasn't been built yet.

I think we need to rearrange some stuff to rely less on the second array being built :thinking:

vd1992 (approved):
👍

PR #11526, opened 2024-09-13
[Fix] Add missing values to pool candidate csv

### LINKED ISSUE #11519: 🐛 Question and skills not appearing in pool candidate CSV
## 🐛 Bug

The question responses and skill justifications are not appearing in the pool candidate CSV document.

## 🦋 Expected Behaviour

The responses to questions along with skill justifications should be appearing under the appropriate columns and rows for candidates.

## 🕵️ Details

We have rearranged the order of operations to the point where expected darta that should exist no longer does:

1. Loop over candidates creating an array of their pool ids
2. Use array of ids to determine what column to place skills in (not built until step 3)
3. Loop over pool ids to get questions/skills while building array of their ids

## 📋 Steps to Reproduce

1. Stasrt worker queue `make queue-work`
3. Navigate to candidate table `/admin/pool-candidates`
4. Download CSV
5. Note the missing question responses and skill details

## 🙋‍♀️ Proposed Solution

I think if we get the pool skills in the candidate loop and then store an array where they key is the pool skill id and the value contains the candidate row number we can append them on after? :thinking: 

```php
// While creating values array
$this->poolSkills = [
  'pool_skill_id' => [
   [
       'candidate' => $currentCandidate,
       'value' => 'skill_details'
    ],
     [
       'candidate' => $currentCandidate,
       'value' => 'skill_details'
    ]
  ]
];

// While createing pool headers
foreach($pools as $pool) {
  $foreach($skills as $idx =>  $skill) {
     $rows = $this->poolSkills[$skill->id];
     foreach($rows as $row) {
       $values[$row['candidate']][$currentHeaders + $idx] = $row['value'];
     }
  }
}
```

## ✅ Acceptance Criteria

- [ ] Question responses and skill details appear in application download CSV

esizer:
🤖 Resolves #11519 

## 👋 Introduction

Updates the order we generate values in to add values that depend on the pool after the initial generation.

## 🧪 Testing

1. Start work queue `make queue-work`
3. Navigate to the candidates table `/admin/pool-candidates`
4. Download CSV of the candidates
5. Compare the download to the snapshots, confirming that the following appears in the expected cells:
    - General question responses
    - Screening question responses
    - Skill details

### DIFF
--- a/api/app/Generators/PoolCandidateCsvGenerator.php
+++ b/api/app/Generators/PoolCandidateCsvGenerator.php
@@ -169,10 +169,10 @@ class PoolCandidateCsvGenerator extends CsvGenerator implements FileGeneratorInt
                 });
                 $values[] = implode(', ', $userSkills->toArray());
 
-                $poolQuestionIds = $candidate->pool->generalQuestions->pluck('id')->toArray();
-                $candidateQuestionIds = $candidate->generalQuestionResponses->pluck('general_question_id')->toArray();
-                foreach ($poolQuestionIds as $questionId) {
-                    if (in_array($questionId, $candidateQuestionIds)) {
+                $poolGeneralQuestionIds = $candidate->pool->generalQuestions->pluck('id')->toArray();
+                $candidateGeneralQuestionIds = $candidate->generalQuestionResponses->pluck('general_question_id')->toArray();
+                foreach ($poolGeneralQuestionIds as $questionId) {
+                    if (in_array($questionId, $candidateGeneralQuestionIds)) {
                         $response = $candidate->generalQuestionResponses->where('general_question_id', $questionId)->first();
                         $this->generalQuestionResponses[$questionId][] = [
                             'candidate' => $currentCandidate,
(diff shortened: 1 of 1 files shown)


api/app/Generators/PoolCandidateCsvGenerator.php
```diff
@@ -162,38 +169,45 @@ public function generate(): self
                 });
                 $values[] = implode(', ', $userSkills->toArray());
 
+                $poolQuestionIds = $candidate->pool->generalQuestions->pluck('id')->toArray();
```
petertgiles, line 172: Can we call this $poolGeneralQuestionIds to differentiate it from $poolScreeningQuestionIds on L184?

petertgiles (approved):
Looks like it's working great.  :medal_sports:

api/app/Generators/PoolCandidateCsvGenerator.php
esizer: Great suggestion. I also updated the other reference in [clearer distinction between questions](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11526/commits/5d27b85d10f35c792665556fb9872db4f68b430f)

PR #11543, opened 2024-09-18
[Feature] Job template has nullable skill level

### LINKED ISSUE #11525: ✨ Job Template asset skills have no level
## ✨ Feature

Job template skill level is currently non-nullable. But that needs to change, since asset skills are intended to have no level.

## 🕵️ Details

Allow Job Template skill levels to have no level in the database.
In validation for mutations to create or update Job Templates, essential skills must have a skill level, but asset skills must not.

## ✅ Acceptance Criteria

- [ ] Job Template skill levels are nullable in the database and schema
- [ ] In validation for mutations to create or update Job Templates, essential skills must have a skill level, but asset skills must not.

## 🛑 Blockers

```[tasklist]
### Blocked By
```

JamesHuf:
🤖 Resolves #11525.

## 👋 Introduction

Makes the skill level for job templates nullable and attempts to validate only asset skills have a null skill level.

## 🧪 Testing

1. Run the migration and seeders
2. Confirm asset skills have a null skill level
3. Confirm the validation works as expected

### DIFF
--- a/api/app/Rules/SkillLevelRequiredIfEssential.php
+++ b/api/app/Rules/SkillLevelRequiredIfEssential.php
@@ -8,7 +8,6 @@ use Closure;
 use Illuminate\Contracts\Validation\DataAwareRule;
 use Illuminate\Contracts\Validation\ValidationRule;
 use Illuminate\Support\Arr;
-use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Str;
 
 /**
@@ -46,7 +45,6 @@ class SkillLevelRequiredIfEssential implements DataAwareRule, ValidationRule
     public function validate(string $attribute, mixed $incomingLevel, Closure $fail): void
     {
         $incomingType = Arr::get($this->data, Str::beforeLast($attribute, '.').'.type');
-        Log::debug($this->data);
 
         // If the template already has the skill take current values into account
         $templateId = Arr::get($this->data, 'jobPosterTemplate.id');
--- a/api/app/Rules/SkillLevelRequiredIfEssential.php
+++ b/api/app/Rules/SkillLevelRequiredIfEssential.php
@@ -54,10 +54,10 @@ class SkillLevelRequiredIfEssential implements DataAwareRule, ValidationRule
         $currentType = null;
         $currentLevel = null;
         if (! is_null($templateId)) {
-            $template = JobPosterTemplate::find($templateId);
+            $template = JobPosterTemplate::findOrFail($templateId);
             $model = $template->skills()->firstWhere('skill_id', $skillId);
-            $currentType = $model->pivot->type;
-            $currentLevel = $model->pivot->required_skill_level;
+            $currentType = $model?->pivot->type;
+            $currentLevel = $model?->pivot->required_skill_level;
         }
         $newType = isset($incomingType) ? $incomingType : $currentType;
         $newLevel = isset($incomingLevel) ? $incomingLevel : $currentLevel;
--- a/api/app/Rules/SkillLevelRequiredIfEssential.php
+++ b/api/app/Rules/SkillLevelRequiredIfEssential.php
@@ -1,0 +1,87 @@
+<?php
+
+namespace App\Rules;
+
+use App\Enums\PoolSkillType;
+use App\Models\JobPosterTemplate;
+use Closure;
+use Illuminate\Contracts\Validation\DataAwareRule;
+use Illuminate\Contracts\Validation\ValidationRule;
+use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Log;
+use Illuminate\Support\Str;
+
+/**
+ * This rule is for ensuring a skills requiredLevel in a job template is
+ * required (or prohibited) based on if it's essential
+ */
+class SkillLevelRequiredIfEssential implements DataAwareRule, ValidationRule
+{
+    /**
+     * Indicates whether the rule should be implicit.
+     *
+     * @var bool
+     */
+    public $implicit = true;
+
+    /**
+     * All of the data under validation.
+     *
+     * @var array<string, mixed>
+     */
+    protected $data = [];
+
+    /**
+     * Create a new rule instance.
+     *
+     * @return void
+     */
+    public function __construct() {}
+
+    /**
+     * Run the validation rule.
+     *
+     * @param  \Closure(string): \Illuminate\Translation\PotentiallyTranslatedString  $fail
+     */
+    public function validate(string $attribute, mixed $incomingLevel, Closure $fail): void
+    {
+        $incomingType = Arr::get($this->data, Str::beforeLast($attribute, '.').'.type');
+        Log::debug($this->data);
+
+        // If the template already has the skill take current values into account
+        $templateId = Arr::get($this->data, 'jobPosterTemplate.id');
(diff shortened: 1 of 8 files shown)


api/app/Rules/SkillLevelRequiredIfEssential.php
```diff
@@ -0,0 +1,87 @@
...
+        $templateId = Arr::get($this->data, 'jobPosterTemplate.id');
+        $skillId = Arr::get($this->data, Str::beforeLast($attribute, '.').'.id');
+        $currentType = null;
+        $currentLevel = null;
+        if (! is_null($templateId)) {
+            $template = JobPosterTemplate::find($templateId);
+            $model = $template->skills()->firstWhere('skill_id', $skillId);
+            $currentType = $model->pivot->type;
+            $currentLevel = $model->pivot->required_skill_level;
+        }
```
esizer, line 61: If attempting to connect a new skill, the `$model` can sometimes be null so we will need to handle that.

I think this should fix it.

```suggestion
        if (! is_null($templateId)) {
            $template = JobPosterTemplate::find($templateId);
            $model = $template->skills()->firstWhere('skill_id', $skillId);
            $currentType = $model?->pivot->type;
            $currentLevel = $model?->pivot->required_skill_level;
        }
```

api/app/Rules/SkillLevelRequiredIfEssential.php
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * Run the validation rule.
+     *
+     * @param  \Closure(string): \Illuminate\Translation\PotentiallyTranslatedString  $fail
+     */
+    public function validate(string $attribute, mixed $incomingLevel, Closure $fail): void
+    {
+        $incomingType = Arr::get($this->data, Str::beforeLast($attribute, '.').'.type');
+        Log::debug($this->data);
```
esizer, line 49: Left over debug we can remove now

api/tests/Feature/JobPosterTemplateTest.php
esizer, line 1: Can we add a test to ensure the rule works? I think it could be one to connect an essential skill with no level and expect the error and another to expect success on nonessential with no level

If we are worried about time, I'd also accept just a new issue :sweat_smile:

api/app/Rules/SkillLevelRequiredIfEssential.php
```diff
@@ -0,0 +1,87 @@
...
+     * Create a new rule instance.
+     *
+     * @return void
+     */
+    public function __construct() {}
+
+    /**
+     * Run the validation rule.
+     *
+     * @param  \Closure(string): \Illuminate\Translation\PotentiallyTranslatedString  $fail
```
petertgiles, line 44: I think this comment is out of date?

api/app/Rules/SkillLevelRequiredIfEssential.php
petertgiles, line 49: You probably don't need this anymore?

esizer (changes requested):
Great! Just two minor things I noticed and hopefully a test but don't see that as a blocker.

One thing we should note, and I'm not sure how to do this, but if you attempt to connect/sync two skills (one essential, one nonessential) and you do not pass a level to the nonessential, it passes validation but does not generate valid SQL.

The follwing passes validation but creates a  SQL syntax error.

```json
{
   "skills": {
     "connect": [
       {
          "id": "...",
          "type": "ESSENTIAL",
          "requiredLevel": "BEGINNER",
       },
       {
          "id": "...",
          "type": "NONESSENTIAL",
       }
     ]
   }
}
```

```sh
SQLSTATE[42601]: Syntax error: 7 ERROR: VALUES lists must all be the same length
```

This is easily fixed by passing null to the `requiredLevel` but maybe we want to document that somewhere until we can sort the validation out? :thinking:  Unless you know of a good way to do it here.

api/graphql/schema.graphql
```diff
@@ -2275,7 +2276,6 @@ type Mutation {
     jobPosterTemplate: CreateJobPosterTemplateInput! @spread
   ): JobPosterTemplate @create @guard @canModel(ability: "create")
   updateJobPosterTemplate(
```
petertgiles, line 2278: Not having the ID the first parameter of the update mutation is a bit out of convention.  What is this change working around?  Can we solve it a different way?

api/graphql/schema.graphql
```diff
@@ -1905,7 +1905,7 @@ input SyncJobPosterTemplateSkillInput {
 }
 
 input UpdateJobPosterTemplateSkillsInput {
-  attach: [AttachJobPosterTemplateSkillInput!]
+  connect: [ConnectJobPosterTemplateSkillInput!]
```
petertgiles, line 1908: I don't mind it, I'm just curious why you made this change?

petertgiles (commented):
Looking good!  I just finished a readthrough and I'm going to start testing it out.

api/app/Rules/SkillLevelRequiredIfEssential.php
JamesHuf: Fixed in [1d71812](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11543/commits/1d71812cb5b8a0a21d1297f50cfb1419aa53bdb3).

api/app/Rules/SkillLevelRequiredIfEssential.php
JamesHuf: Removed in [c288609](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11543/commits/c288609bbac5ce2a5ea65326889b9b8982c004aa).

api/graphql/schema.graphql
JamesHuf: Because we're running validation on the UpdateJobPosterTemplateInput type the id doesn't get passed in to validation using the other method. This means I can't access it to view current data and even that the unique check for referenceId will stumble on itself if you pass in it's current value when updating.
You can also see it in the update user mutations where we actually pass in the id in both places and if you miss passing the id in the input type it can cause the email unique check to fail on itself.

JamesHuf: For example this mutation
```
mutation UpdateUser($id: ID!, $user: UpdateUserAsUserInput!) {
    id
    email
  }
}
```

will fail if you fail sending the id a second time here
```
{
  "id": "16000f9d-41f7-4912-bd3d-a9b92f3f77d6",
  "user": {
    "id": "16000f9d-41f7-4912-bd3d-a9b92f3f77d6",
    "email": "admin@test.com"
  }
}
```

api/graphql/schema.graphql
JamesHuf: Attach is [not a proper operation](https://lighthouse-php.com/6/eloquent/nested-mutations.html#belongstomany) so no data was being saved for skills with it.

api/app/Rules/SkillLevelRequiredIfEssential.php
petertgiles: Weird.  :laughing: Oh well, not a big deal.

api/graphql/schema.graphql
petertgiles: Ah, OK.  Fair enough.

petertgiles (approved):
Working well for me now.  :clap:

esizer (approved):

PR #11553, opened 2024-09-18
[Bug] Fix skill category initial value in edit skill page

### LINKED ISSUE #11532: 🐛 [PROD] When editing an existing Skill as admin, Category isn't populated correctly
## 🐛 Bug

When you go to edit an existing skill in the admin UI, the Category (Technical vs Behavioural) select element isn't always initialized with the correct value. It may appear Technical in the table, but the form has Behavioural selected when you open the page to edit the skill.

## 🦋 Expected Behaviour

When editing a skill, the initial Category value in the form must match the skill's actual value.

## 🕵️ Details

A video shared in slack showing the error on prod: https://talent-cloud.slack.com/archives/C066AR2CUV7/p1726491525824419

## 📋 Steps to Reproduce

1. Create a skill with the Technical skill category.
2. Go back to edit it. Is the Category select element in the form set to Behavioural?
3. If it isn't, try editing other skills and seeing if the form matches the entry in the table.

## 🙋‍♀️ Proposed Solution

Check how the form is initialized. Maybe there's a typo or something?

Could also be related to localized enum changes.

## ✅ Acceptance Criteria

- [ ] When you open an Edit Skill page, the Category value in the form must match the actual category value of the skill.

vd1992:
🤖 Resolves #11532 

## 👋 Introduction

Resolve the bug.
Appears to stem from the enums change, now that enums are fetched from the API there is latency involved with the options dropdown. Order happening appeared to be something like:
Form renders, initial values applied -> API fetch of enums completes -> defaults to first in options list (alphabetically behavioural)

I resolved this by resetting the field to default/initial once the fetch completed. 

## 🧪 Testing

1. Create a skill with the Technical skill category
2. Go back to edit it, should have technical category, refresh page to double check
3. Check various other skills too


brindasasi (approved):
Neat! Fixes the problem 👏

PR #11569, opened 2024-09-19
[Fix] Reduce previous url to max length for freshdesk

### LINKED ISSUE #11484: 🐛 Page URL length causes failure to create a ticket
## 🐛 Bug

I've been seeing this error logged lately:
```
Error when trying to create a ticket: {"description":"Validation failed","errors":[{"field":"custom_fields.cf_page_url","message":"Has 2178 characters, it can have maximum of 255 characters","code":"invalid_value"}]}
```
## 🕵️ Details

https://github.com/GCTC-NTGC/gc-digital-talent/blob/8ba30db4179e9fbe73e2d718012c09f866436f78/api/app/Http/Controllers/SupportController.php#L38

## 🙋‍♀️ Proposed Solution

This field isn't one that the user can manually set in the form so maybe just automatically shorten it before submission?

## ✅ Acceptance Criteria

- [ ] It's not possible for this error to occur
- [ ] A user can always submit their help ticket

esizer:
🤖 Resolves #11484 

## 👋 Introduction

Fixes an issue where the previous url field was too long for freshdesk.

## 🕵️ Details

This does two things:

1. Only passes the path and query
    - We do not need any other parts like domain since we only set this on our pages so we can assume the domain is ours
2. Trims path + query to max length

## 🧪 Testing

1. Set anything for the freshdesk variable sin `api/.env` (they can be anything, we don't need to send a request but you can if you have that)
2. Confirm the previous URL is a max length of 255 characters

### DIFF
--- a/api/app/Http/Controllers/SupportController.php
+++ b/api/app/Http/Controllers/SupportController.php
@@ -6,8 +6,12 @@ use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Http;
 use Illuminate\Support\Facades\Log;
 
+use function Safe\parse_url;
+
 class SupportController extends Controller
 {
+    private static $MAX_URL_LENGTH = 255;
+
     public function createTicket(Request $request)
     {
         if (! config('freshdesk.api.tickets_endpoint') || ! config('freshdesk.api.key')) {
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
Works great in my testing!  I tried submitting a ticket to the Freshdesk service, too.  Thanks for fixing this.

PR #11576, opened 2024-09-19
[Feature] No queue for single row document downloads

### LINKED ISSUE #11534: ✨ When downloading user/candidate profiles from table, if only single item is checked, avoid queue+notification system
## ✨ Feature

This is essentially what we did in [#11401](https://github.com/GCTC-NTGC/gc-digital-talent/issues/11401), but we want to do the same thing from the table interface if only a single item is checked.

Filename should follow the pattern from 11401 too.

If only one row is checked, use one of the `download____Doc` mutations created in #11407 instead of the bulk download mutations.

## ✅ Acceptance Criteria

- [ ] If only a single user/candidate is checked on the table when Download is clicked, download the file directly instead of putting the file generation in the queue

vd1992:
🤖 Resolves #11534

## 👋 Introduction

Do not use queuing for downloads of a single row in the users or candidates table. 

## 🧪 Testing

1. Navigate to users table, have queue work running
2. Select one user and download, then select multiple users then download
3. Observe immediate download for the former, queuing for the latter
4. Go to candidates table and repeat

## 📸 Screenshot

image

### DIFF
--- a/api/app/GraphQL/Mutations/DownloadPoolCandidateDoc.php
+++ b/api/app/GraphQL/Mutations/DownloadPoolCandidateDoc.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace App\GraphQL\Mutations;
 
 use App\Generators\PoolCandidateUserDocGenerator;
+use App\Models\PoolCandidate;
 use Illuminate\Support\Facades\App;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Log;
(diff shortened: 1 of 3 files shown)


api/app/GraphQL/Mutations/DownloadPoolCandidateDoc.php
vd1992, line 1: Copied filename method from 
`app/GraphQL/Mutations/DownloadApplicationDoc.php`

apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx
```diff
@@ -986,11 +1014,16 @@ const PoolCandidatesTable = ({
...
-              disabled={downloadingDoc}
-              onClick={
-                hasSelectedRows ? handleDocDownload : handleNoRowsSelected
+              disabled={
+                !hasSelectedRows ||
+                downloadingDoc ||
+                downloadingSingleDoc ||
+                downloadingAsyncFile
+              }
+              onClick={handleDocDownload}
```
vd1992, line 1023: Rather than handle three cases, copied how it works in the users table where the download is disabled with no rows selected. Consistency?!

brindasasi (approved):
Nice! Just one question. Are we not including CSV file generation? I don't see it in A/C as well. Other than that this works fine as per A/C.

vd1992:
> Nice! Just one question. Are we not including CSV file generation? I don't see it in A/C as well. Other than that this works fine as per A/C.

I don't think single row CSV files makes sense

PR #11579, opened 2024-09-19
Store accordion states to local storage on manager dashboard

### LINKED ISSUE #11464: ✨ Store accordion states in localStorage
## ✨ Feature

We want to persist the accordion open and close state on the Manager dashboard page. Currently, there is only the Talent Request history accordion that needs to be implemented.

## 🕵️ Details

Once the other sections of the page that include accordions (e.g. Your job advertisements), we will need to implement the same functionality for them as well.

The implementation will be almost identical to the admin side menu's functionality:

```jsx
  // retain menu preference in storage
  const [accordionItems, setAccordionItems] = useLocalStorage(
    "accodion_items", // accordion key to access local storage (can be named anything)
    ["your_talent_requests", "your_job_advertisements"] // initial value to use if there is no item in the local storage with provided key
  );

(...)

<Accordion.Root
  type="multiple"
  value={currentAccordionItems}
  onValueChange={(newValue: AccordionItems) => {
    setCurrentAccordionItems(newValue);
  }}
>
(...)
</Accordion.Root>

(...)

<Accordion.Item value="your_talent_requests">(...)</Accordion.Item>
```

Don't forget you'll need to set the key of each accordion item

## ✅ Acceptance Criteria

- [ ] Accordions open and closed state persists on localStorage
- [ ] Accordions state is correct when the user visits the page
- [ ] Accordions should start open

## 🛑 Blockers

```[tasklist]
### Blocked By
- [ ] https://github.com/GCTC-NTGC/gc-digital-talent/issues/10785
```

yonikid15:
🤖 Resolves #11464 

## 👋 Introduction

- Stores accordion states to local storage on manager dashboard

## 🧪 Testing

1. Build app `pnpm run dev`
2. Login as `manager@test.com`
3. Go to `/manager/dashboard`
4. Ensure the accordion state persists on local storage and the state is correct

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/pages/ManagerDashboardPage/ManagerDashboardPage.tsx
+++ b/apps/web/src/pages/ManagerDashboardPage/ManagerDashboardPage.tsx
@@ -50,7 +50,7 @@ type AccordionItems = ("your_talent_searches" | "")[];
 const ManagerDashboard = ({ userQuery }: ManagerDashboardProps) => {
   const paths = useRoutes();
   const intl = useIntl();
-  const [accordionItems, setAccordionItems] = useLocalStorage(
+  const [accordionItems, setAccordionItems] = useLocalStorage<AccordionItems>(
     "accordion_items",
     ["your_talent_searches"],
   );
--- a/apps/web/src/pages/ManagerDashboardPage/ManagerDashboardPage.tsx
+++ b/apps/web/src/pages/ManagerDashboardPage/ManagerDashboardPage.tsx
@@ -51,7 +51,7 @@ const ManagerDashboard = ({ userQuery }: ManagerDashboardProps) => {
   const paths = useRoutes();
   const intl = useIntl();
   const [accordionItems, setAccordionItems] = useLocalStorage<AccordionItems>(
-    "accordion_items",
+    "managerDashboardPage:managerToolsAccordion:expandedItems",
     ["your_talent_searches"],
   );
   const user = getFragment(ManagerDashboardUser_Fragment, userQuery);
--- a/apps/web/src/pages/ManagerDashboardPage/ManagerDashboardPage.tsx
+++ b/apps/web/src/pages/ManagerDashboardPage/ManagerDashboardPage.tsx
@@ -15,6 +15,7 @@ import {
 import { navigationMessages } from "@gc-digital-talent/i18n";
 import { FragmentType, getFragment, graphql } from "@gc-digital-talent/graphql";
 import { ROLE_NAME } from "@gc-digital-talent/auth";
+import { useLocalStorage } from "@gc-digital-talent/storage";
 
 import SEO from "~/components/SEO/SEO";
 import profileMessages from "~/messages/profileMessages";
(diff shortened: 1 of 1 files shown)


apps/web/src/pages/ManagerDashboardPage/ManagerDashboardPage.tsx
```diff
@@ -44,9 +45,15 @@ interface ManagerDashboardProps {
...
 }
 
+type AccordionItems = ("your_talent_searches" | "")[];
+
 const ManagerDashboard = ({ userQuery }: ManagerDashboardProps) => {
   const paths = useRoutes();
   const intl = useIntl();
+  const [accordionItems, setAccordionItems] = useLocalStorage(
+    "accordion_items",
+    ["your_talent_searches"],
```
petertgiles, line 55: This doesn't seem to be stored on initial page load.  It only gets stored after you interact with it.  I guess it is smart enough to know that it doesn't need to store the default value because it works.  :shrug:

apps/web/src/pages/ManagerDashboardPage/ManagerDashboardPage.tsx
```diff
@@ -44,9 +45,15 @@ interface ManagerDashboardProps {
   userQuery: FragmentType<typeof ManagerDashboardUser_Fragment>;
 }
 
+type AccordionItems = ("your_talent_searches" | "")[];
+
 const ManagerDashboard = ({ userQuery }: ManagerDashboardProps) => {
   const paths = useRoutes();
   const intl = useIntl();
+  const [accordionItems, setAccordionItems] = useLocalStorage(
+    "accordion_items",
```
petertgiles, line 54: I think the key name of this accordion on this page is too generic since the storage is scoped for the entire origin.  
image
I wouldn't be surprised if we want to start storing other accordions on other pages, too.

You can see graphiql is using colons.  We could do that or periods.  :thinking:  We could do something like page:widget:function.  For example managerDashboardPage:managerToolsAccordion:expandedItems.

apps/web/src/pages/ManagerDashboardPage/ManagerDashboardPage.tsx
```diff
@@ -44,9 +45,15 @@ interface ManagerDashboardProps {
   userQuery: FragmentType<typeof ManagerDashboardUser_Fragment>;
 }
 
+type AccordionItems = ("your_talent_searches" | "")[];
+
 const ManagerDashboard = ({ userQuery }: ManagerDashboardProps) => {
   const paths = useRoutes();
   const intl = useIntl();
+  const [accordionItems, setAccordionItems] = useLocalStorage(
```
petertgiles, line 53: You can tighten the typing up a bit here.
```suggestion
  const [accordionItems, setAccordionItems] = useLocalStorage<AccordionItems>(
```

petertgiles (changes requested):
This is working great!   I have a few small refinements suggested.

apps/web/src/pages/ManagerDashboardPage/ManagerDashboardPage.tsx
petertgiles: Another approach is to store a load a single state object for a page.  Something like:

apps/web/src/pages/ManagerDashboardPage/ManagerDashboardPage.tsx
yonikid15: When the accordionItems value changes it shows up in local storage, so before that event it doesn't show up. The initial values are added to the Accordion.Root so it should be steady.

apps/web/src/pages/ManagerDashboardPage/ManagerDashboardPage.tsx
yonikid15: I changed it to your first suggestion. Very good catch. I think we use that accordion key on another page :laughing: . For the second suggestion, we could switch to that in the future if there are more states to store. Changed in [6151cf9](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11579/commits/6151cf91eda6fd317e4591f8bf6a7bf5b9048b2f)

petertgiles (approved):
Thanks!

PR #11645, opened 2024-09-26
[Fix] Increase uniqueness community key factory

### LINKED ISSUE #11644: 🐛 Community factory slug not unique enough
## 🐛 Bug

It is fairly rare, but the use of only 2 words in the key for our community factory is leading to collisions in our testing.

## 🦋 Expected Behaviour

No random key collisions for community factory.

## 📋 Steps to Reproduce

See: https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/11051828494/job/30702573849?pr=11634#step:7:46

## 🙋‍♀️ Proposed Solution

Increase the words used in the key to avoid collisions.

## ✅ Acceptance Criteria

- [ ] Key has increased uniqueness
- [ ] No collisions in tests

esizer:
🤖 Resolves #11644 

## 👋 Introduction

Increases the uniqueness of the community key factory by increasing the words in the result and using the `unique` method.

## 🕵️ Details

I'm not sure the `unique` key works when called separately so I also increased the words. We were using 2 which was sometimes causing collisions. 

## 🧪 Testing

1. Create a seeder to generates a lot of communities (1000?)
2. Run the seeder
3. Confirm unique constraint is not violated

### DIFF
--- a/api/database/factories/CommunityFactory.php
+++ b/api/database/factories/CommunityFactory.php
@@ -25,7 +25,7 @@ class CommunityFactory extends Factory
         $description = $this->faker->sentence();
 
         return [
-            'key' => $this->faker->slug(2),
+            'key' => $this->faker->unique()->slug(5),
             'name' => ['en' => $name.' EN', 'fr' => $name.' FR'],
             'description' => ['en' => $description.' EN', 'fr' => $description.' FR'],
         ];


vd1992 (approved):
Works for me, 750 created without issue

PR #11647, opened 2024-09-26
[Fix] Stabilize priority claim order 

### LINKED ISSUE #11586: 🧪 Pool candidate claim verification sort flaky
## 🧪 Tests

The test for sorting pool candidates by category (+ claim verification) is not consistently passing.

## 🕵️ Details

Seems as though the sort of the first two candidates comes back in a different order sometimes. As far as I can tell, it should be passing consistently. The candidates defined as:

1. Bookmark candidate that makes no positive claims about their priority (not veteran, not citizen, does not have priority entitlement) and submitted at am earlier date
2. Bookmarked candidate that claims citizenship and priority entitlement (and accepted) but does not claim to be a veteran who submitted at a later date.

The sorting should have the following weight:

1. Bookmarked
2. Has accepted priority claim
3. Is citizen
4. Submitted at an earlier date

So, even though both are bookmarked and the priority candidate submitted at a later date, since they are both a citizen and has an accepted priority claim, they should _always_ appear earlier in the order.

## 🙋‍♀️ Proposed Solution

It would be good to first confirm the assertion about the candidates in the details section is accurate first and that the factories are not creating candidates that do not meet the assertion. 

If the assertion about the candidates is accurate in the test factories, then we should look at the actual sorting to confirm there is nothing wrong there and that the candidates are consistency returned with the same order.

## ✅ Acceptance Criteria

- [ ] Test continues to assert the proper sort order
- [ ] Test consistently passes

esizer:
🤖 Resolves #11586 

## 👋 Introduction

Attempts to stabilize the ordering of candidate priority claims.

## 🕵️ Details

I think this was because the user who had unverified priority was not explicitly set to claim it so it was random and when they never claimed priority entitlement, they would naturally be sorted lower causing the error.

## 🧪 Testing

1. Confirm the `testOrderByClaimVerification` test passes consistently

### DIFF
--- a/api/tests/Feature/PoolCandidateTest.php
+++ b/api/tests/Feature/PoolCandidateTest.php
@@ -836,7 +836,10 @@ class PoolCandidateTest extends TestCase
         $unverifiedPriorityAndAcceptedVeteran = PoolCandidate::factory()->create(
             [
                 'pool_id' => $poolOne,
-                'user_id' => User::factory()->create(['citizenship' => CitizenshipStatus::CITIZEN->name]),
+                'user_id' => User::factory()->create([
+                    'has_priority_entitlement' => true,
+                    'citizenship' => CitizenshipStatus::CITIZEN->name,
+                ]),
             ],
         );
         $unverifiedPriorityAndAcceptedVeteran->update([


vd1992 (approved):
Only failures were from the other flaky test
Good to go

Issue #11379, opened 2024-08-26
♻️ No misused promises


esizer:
## ♻️ Debt/Refactor

Enable the `no-misused-promises` rule and fix associated errors.

## 🕵️ Details

https://typescript-eslint.io/rules/no-misused-promises/

This is a big one and almost exclusively comes from our form submit handlers. We may need to see about creating a pattern and update most of our forms to follow the new submit pattern that properly uses promises.

## ✅ Acceptance Criteria

- [ ] `no-misused-promises` enabled
- [ ] All errors fixed

PR #11658, opened 2024-09-27
[Debt] Fix require await errors

### LINKED ISSUE #11377: ♻️ Require await eslint rule
## ♻️ Debt/Refactor

Enable and fix `require-await`

## 🕵️ Details

https://typescript-eslint.io/rules/require-await/

## 🙋‍♀️ Proposed Solution

Remove the line disabling this rule and fix any errors associated with it.

## ✅ Acceptance Criteria

- [ ] `require-await` is enabled
- [ ] All errors fixed

esizer:
🤖 Resolves #11377 

## 👋 Introduction

Re-enables the require await eslint rule and fixes errors.

## 🕵️ Details

Turns out we were not missing awaits in a lot of places and rather were overeager about colouring our functions with the async keyword. This occurred most often in tests and form submissions that sent graphql requests.

## 🧪 Testing

1. Confirm no lint errors
2. Confirm forms submit as expected


petertgiles:
A lot of these changes are removing the async tag from submit handlers.  Shouldn't these be fixed by returning the promises instead?  That would help prevent things like closing a dialog before the mutation returns, right?

esizer:
> A lot of these changes are removing the async tag from submit handlers. Shouldn't these be fixed by returning the promises instead? That would help prevent things like closing a dialog before the mutation returns, right?

Not necessarily, the submissions themselves don't need to be async, they just trigger an async function (the mutation) that can close the dialog if successful. I actually think these changes are going to reduce the amount of [misused promises](https://github.com/GCTC-NTGC/gc-digital-talent/issues/11379) we have....maybe :sweat_smile:

apps/web/src/pages/Users/UserInformationPage/components/AddToPoolDialog.tsx
```diff
@@ -130,24 +130,20 @@ const AddToPoolDialog = ({ user, poolCandidates }: AddToPoolDialogProps) => {
...
-  ) => {
+  const submitForm: SubmitHandler<FormValues> = (formValues: FormValues) => {
     const poolsToUpdate = formValues.pools
       .map((poolId) => poolMap.get(poolId))
       .filter(notEmpty);
 
-    const promises = poolsToUpdate.map((pool) => {
-      return requestMutation({
+    const promises = poolsToUpdate.map(async (pool) => {
+      return await requestMutation({
```
petertgiles, line 139: Doesn't awaiting on the mutation here remove the benefit of using `allSettled` below?  Now they're forced to run serially?

apps/web/src/pages/Auth/RegistrationPages/GettingStartedPage/GettingStartedForm.stories.tsx
```diff
@@ -25,6 +25,8 @@ const Template: StoryFn<typeof GettingStartedForm> = () => {
     <GettingStartedForm
       cacheKey=""
       query={mockFragmentData}
+      // NOTE: Needed for function colouring
```
petertgiles, line 28: Could you please help me understand this change?

petertgiles (approved):
Looks great!  I've asked a few questions but nothing blocking.

apps/web/src/pages/Auth/RegistrationPages/GettingStartedPage/GettingStartedForm.stories.tsx
esizer: Yeah, this prop requires an async function be passed to it. However, since `action` is not async, we just give it the keyword to colour it to satisfy typescript. Just an annoying bit of JS/TS. Since this is a story, I opted to ignore it instead of hacking in some async code just to satisfy typescript.

apps/web/src/pages/Users/UserInformationPage/components/AddToPoolDialog.tsx
esizer: That is a good point. Instead, I have removed the async keyword so we don't need to await the code to satisfy the rule.

[remove async for add to pool submission](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11658/commits/45174e98fdad2516513f154743f9cf3549aced70)

PR #11661, opened 2024-09-27
[Fix] Only throw errors

### LINKED ISSUE #11378: ♻️ Only throw error
## ♻️ Debt/Refactor

Enable and fix the `only-throw-error` eslint rule

## 🕵️ Details

This one could be trickey since we are using the response to in a weird way since we don't have data loaders working in `react-router`.

REF: https://typescript-eslint.io/rules/only-throw-error

## 🙋‍♀️ Proposed Solution

I think there are few options here:

1. See if returning the response works
2. Extend the `Error` class in javascript to support status codes so we can render the appropriate error
3. Embed status code in the error message and try and parse it in our boundary

## ✅ Acceptance Criteria

- [ ] `only-throw-error` rule is enabled
- [ ] All errors fixed

esizer:
🤖 Resolves #11378 

## 👋 Introduction

Updates our throws to use custom errors instead of hijacking the response.

## 🕵️ Details

This had to tweak our error boundary to use Errors instead of Responses. Though, you could get an error response so we have to handle that as well (even though we don't directly use them).

## 🧪 Testing

1. Bujild `pnpm run dev:fresh`
2. Confirm explicit throws still render the appropriate error messages
3. Confirm no lint errors

### DIFF
--- a/apps/web/src/components/Router.tsx
+++ b/apps/web/src/components/Router.tsx
@@ -4,6 +4,7 @@ import { Locales, useLocale } from "@gc-digital-talent/i18n";
 import { POST_LOGOUT_OVERRIDE_PATH_KEY } from "@gc-digital-talent/auth";
 import { Loading } from "@gc-digital-talent/ui";
 import { defaultLogger } from "@gc-digital-talent/logger";
+import { NotFoundError } from "@gc-digital-talent/helpers";
 
 const createRoute = (locale: Locales) =>
   createBrowserRouter([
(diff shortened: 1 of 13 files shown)


apps/web/src/components/Router.tsx
```diff
@@ -35,7 +36,7 @@ const createRoute = (locale: Locales) =>
                 {
                   path: "dashboard",
                   loader: async () => {
-                    throw new Response("Not Found", { status: 404 }); // unfinished page
+                    throw new NotFoundError();
```
petertgiles, line 39:
```suggestion
                    throw new NotFoundError(); // unfinished page
```

petertgiles (approved):
Nice!  Works well.

PR #11662, opened 2024-09-27
[Fix] Remove base to string

### LINKED ISSUE #11380: ♻️ No base to string eslint rule
## ♻️ Debt/Refactor

Enable and fix the `no-base-to-string` eslint rule.

## 🕵️ Details

REF: https://typescript-eslint.io/rules/no-base-to-string

This is most used in our forms the turn form errors into strings to render. We may need to make our error getter a little more specific to get a string value, right now it returns a sum type where we just assume it is a string. 

There are are other instances but they should be handled in their own specific way.

## 🙋‍♀️ Proposed Solution

May need a recursive getter written for extracting error messages out for an input since they are assumed to be nested.

## ✅ Acceptance Criteria

- [ ] `no-base-to-string` rule enabled
- [ ] All errors fixed

esizer:
🤖 Resolves #11380 

## 👋 Introduction

Re-enables and fixes the `no-base-to-string` eslint rule.

## 🕵️ Details

This was used on field errors and react nodes. 

For the field errors, opted to use the already installed `ErrorMessage` component from `react-hook-form` that does what we were doing and removes the need for `lodash` and `toString`.

For react nodes, the [recommended approach](https://react.dev/reference/react-dom/server/renderToString#removing-rendertostring-from-the-client-code) was implemented.

## 🧪 Testing

1. Build `pnpm run dev:fresh`
2. Confirm form errors still work
3. Confirm responsive table headers still work
4. Confirm no lint errors


apps/web/src/utils/nodeToString.ts
petertgiles, line 1: Oh man, I could have used this so many times in the past.  :laughing:

apps/web/src/utils/nodeToString.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Convert a ReactNode to a string
+ *
+ * REF: https://react.dev/reference/react-dom/server/renderToString#removing-rendertostring-from-the-client-code
+ */
+function nodeToString(node: ReactNode): string {
+  if (typeof node === "string") return node;
+
+  const div = document.createElement("div");
+  const root = createRoot(div);
+  queueMicrotask(() => {
```
petertgiles, line 15: I've never used queueMicrotask before.  It looks like this is your own enhancement to the function?  How does this work?

petertgiles (approved):
Very cool!

apps/web/src/utils/nodeToString.ts
esizer: So, I was getting an error that `nodeToString` was being called in a life cycle hook (you cannot call `flushSync` in a  life cycle hook to [avoid the many caveats](https://react.dev/reference/react-dom/flushSync#caveats))

The `queueMicroTask` basically ripped this out of the life cycle by queuing it after it occurred. However, I can not reproduce the error so I removed it in [remove micro task](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11662/commits/5e735ca5b15f22e2b87189f32b569c9afd30d6bd)

PR #11663, opened 2024-09-27
[Fix] Fix consistent types for `useRequiredParams`

### LINKED ISSUE #11383: ♻️ Fix `useRequiredParams` generic type constraint
## ♻️ Debt/Refactor

Update `useRequiredParams` to accept interface type in generic.

## 🕵️ Details

The constraint for the generic type only works with `type` keyword but we would like it to also work with interfaces so we can prefer them.

REF: https://typescript-eslint.io/rules/consistent-type-definitions

## ✅ Acceptance Criteria

- [ ] `useRequiredParams` accepts interfaces for the generic type
- [ ] Removed ignore comments for `consistent-type-definitions`

esizer:
🤖 Resolves #11383 

## 👋 Introduction

Fixes the type signature for `useRequiredParams` so we can use consistent types (`interface`).

## 🕵️ Details

The issue was the extension of the generic requiring a string. Interfaces do not implement that interface. So, we need to extend it in our definitions.

## 🧪 Testing

1. Confirm no lint errors
2. Confirm `useRequiredParams` still functions

### DIFF
--- a/apps/web/src/hooks/useRequiredParams.ts
+++ b/apps/web/src/hooks/useRequiredParams.ts
@@ -58,7 +58,9 @@ function assertParam(
  * @param enforceUUID
  */
 const useRequiredParams = <
-  T extends Record<string, string>,
+  // Note: Constrained by the useParams generic also not knowing
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  T extends Record<PropertyKey, any>,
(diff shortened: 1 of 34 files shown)


brindasasi (approved):
Verified all the impacted pages loading properly and the eslint disable comment has been removed
LGTM! 👍

PR #11666, opened 2024-09-27
[Debt] Larastan - Call to function always...

### LINKED ISSUE #11593: ♻️ Larastan: Call to function X will always evaluate to {false|true}
## ♻️ Debt/Refactor

Fix all errors "Call to function X will always evaluate to {false|true}" and un-ignore them

## 🙋‍♀️ Proposed Solution

- Confirm the types are accurate
- If not, fix the type hinting
- If it is, fix the error mentioned

## ✅ Acceptance Criteria

- [ ] All errors fixed
- [ ] Ignores removed from config
- [ ] Larastan passes

vd1992:
🤖 Resolves #11593

## 👋 Introduction

Resolves "will always evaluate to ..." errors
Mostly type-hinting, an unnecessary check or two and one minor bug! Some other errors resolved again as a side effect. 

## 🧪 Testing

1. No errors left in
2. Things work still


api/app/Console/Commands/CheckIntl.php
```diff
@@ -158,7 +158,7 @@ private function checkStrings($enum)
 
         // Find exact matches of strings across locales
         $intersectLocales = array_filter(array_keys(array_intersect_assoc($existing['en'], $existing['fr'])), function ($item) {
-            return ! is_array($item) && ! in_array($item, $this->allowedMatch);
+            return ! in_array($item, $this->allowedMatch);
         });
```
vd1992, line 162: The variable apparently would never be an array making the deleted `! is_array($item)` always true

api/app/Models/PoolCandidate.php
```diff
@@ -284,11 +284,7 @@ public function getCategoryAttribute()
             $category = PriorityWeight::OTHER;
         }
 
-        return ! is_null($category) ? [
-            'weight' => $category->weight($category->name),
-            'value' => $category->name,
-            'label' => PriorityWeight::localizedString($category->name),
-        ] : $category;
```
vd1992, line 291: The fallback above made it so the variable was never null

api/app/Policies/PoolPolicy.php
```diff
@@ -172,7 +172,7 @@ public function updatePublished(User $user, Pool $pool)
      */
     public function publish(User $user, Pool $pool)
     {
-        if (! $pool->getStatusAttribute() === PoolStatus::DRAFT->name) {
+        if (! ($pool->getStatusAttribute() === PoolStatus::DRAFT->name)) {
             return Response::deny('Pool has already been published.');
         }
```
vd1992, line 178: An actual bug found! This would always evaluate to false before as it would evaluate it as
`( ! $pool->getStatusAttribute()) === PoolStatus::DRAFT->name`

Note the parentheses

esizer: We could even simplify this further if we want to make it not so confusing :sweat_smile: 

```php
if($pool->status !== PoolStatus::DRAFT->name) {}
```

esizer (approved):
LGTM! :+1: 

I made a tiny suggestion but it is technically expanding scope so only do it if you feel like it :stuck_out_tongue:

api/app/Policies/PoolPolicy.php
vd1992: Yeah, I think that looks better https://github.com/GCTC-NTGC/gc-digital-talent/pull/11666/commits/929b5acc0af4c8c846c15e1e740349f4cb923e8a

vd1992: https://github.com/GCTC-NTGC/gc-digital-talent/pull/11666/commits/da0eb9553261d013ed08d80830617af7c6720fac

PR #11671, opened 2024-10-01
[Debt] Larastan - nullsafe operator unneeded 

### LINKED ISSUE #11592: ♻️ Larastan: Using nullsafe method call on non-nullable type
## ♻️ Debt/Refactor

Fix all of the Using nullsafe method call on non-nullable type {type}. Use -> instead." errors.

## 🙋‍♀️ Proposed Solution

- Confirm the type is non-nullable
- If it is, do not use null safe method call
- If it is not, update the type to hint that it is nullable
- Remove all ignores

## ✅ Acceptance Criteria

- [ ] All nullsafe method call errors fixed
- [ ] No longer being ignored
- [ ] Larastan passes

vd1992:
🤖 Resolves #11592 

## 👋 Introduction

Updated typehinting or removed the nullsafe model operator where needed. 
Seemed straightforward to me?

## 🧪 Testing

1. No nullsafe errors missed
2. No logic/code in question was  functionally changed

### DIFF
--- a/api/app/Builders/PoolBuilder.php
+++ b/api/app/Builders/PoolBuilder.php
@@ -217,7 +217,7 @@ class PoolBuilder extends Builder
 
     public function authorizedToAdmin(): self
     {
-        /** @var \App\Models\User */
+        /** @var \App\Models\User | null */
         $user = Auth::user();
 
         // if they can view any, then nothing filtered out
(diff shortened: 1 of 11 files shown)


esizer (approved):
Seems to work well :+1:

PR #11667, opened 2024-09-27
[Fix] Update manager dashboard UI

petertgiles:
🤖 Resolves #11575

## 👋 Introduction

This branch corrects (almost) all the items from Josh's design review of the manager dashboard page.

## 🕵️ Details

Note that one AC was not done:
>  The final item in the request accordion shouldn't have bottom padding, as the card already provides it

I think it's already correct:
image

## 🧪 Testing

1. Uncomment the loader in router.tsx to enable the manager dashboard page
2. Visit /en/manager/dashboard

PR #11674, opened 2024-10-02
[Feature] Accordion middle partials

### LINKED ISSUE #11587: ✨ Add middle content to accordion component
## ✨ Feature

On the manager dashboard we need an accordion that has some space for interactive content between the trigger and the content.  Currently, the content is manually added on that page but the spacing is not configurable enough.

## 🎨 Design

Josh:

Component specification:
https://www.figma.com/design/guHeIIh8dqFVCks310Wv0G/Component-library?node-id=5-20&node-type=canvas&t=3VSWIj96B8VFWC0D-0

Accordion that needs middle content:
https://www.figma.com/design/oZhQe9ahidHFfaiqgREvxa/Dashboard-(All-users)?node-id=1010-18680&t=jCsiShy545bh9AqT-4

## 📸 Screenshot

image

From Josh:
It would be nice to build this into the component so we can reuse it elsewhere... a slot or something you can pass 0 or more actions to or something similar.

Should accept an array of items, like the preview list item.

## ✅ Acceptance Criteria

- [ ] Component updated
- [ ] Manager dashboard accordion updated to use new component feature

## 🛑 Blockers

```[tasklist]
### Blocked By
```

yonikid15:
🤖 Resolves #11587 & #11700

## 👋 Introduction

- Adds a new MetaData component to Accordion composition (still okay with changing the component name :laughing:).
- The MetaData component adds a list of buttons, links, text, and chips below accordion trigger.

## 🕵️ Details

- Will need to update styling after #11667 is merged.

## 🧪 Testing

1. Uncomment the loader in router.tsx to enable the manager dashboard page
2. Visit /en/manager/dashboard
3. AND open storybook:design-system and view Accordion component

## 📸 Screenshot

image


petertgiles:
> Adds a new MetaData component to Accordion composition (still okay with changing the component name 😆).

It looks like the name that Josh uses in Figma is "Metadata".  I think it's a good idea to stay in sync with designers for names so I'd suggest just correcting the capitalization.

packages/ui/src/components/Accordion/Accordion.stories.tsx
```diff
@@ -17,6 +18,38 @@ const Text = () => {
   return <p>{faker.lorem.sentences(5)}</p>;
 };
 
+export const metaData: AccordionMetaData[] = [
```
petertgiles, line 21: It seems like Storybook is picking this up as a story and failing to render anything.

packages/ui/src/components/Accordion/Accordion.tsx
petertgiles, line 1: Uh oh, something went wrong here.
image

petertgiles, line 1: Pre-existing, but this doesn't look right, to me.
image

petertgiles: Maybe related to switching the padding to margin?

petertgiles (changes requested):
Looks super nice!  :yum:   I flagged a few small things to look at.

petertgiles:
> Will need to update styling after https://github.com/GCTC-NTGC/gc-digital-talent/pull/11667 is merged.

It's merged now.

packages/ui/src/components/Accordion/Accordion.tsx
yonikid15: Hmm, ya that should be fixed. However, I believe the text was added as a reminder of the context of the icon. Normally, it would just be the icon if used across the site.

packages/ui/src/components/Accordion/Accordion.stories.tsx
yonikid15: Fixed in [a80fc37](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11674/commits/a80fc3707f6f724c93bccce227eb58cbbaf5d8df)

packages/ui/src/components/Accordion/Accordion.tsx
yonikid15: Nice catch! Fixed in [4148e2b](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11674/commits/4148e2b48c9421a918ed1cec81277417c69cccd5)

yonikid15: Created an issue for this https://github.com/GCTC-NTGC/gc-digital-talent/issues/11700

petertgiles (changes requested):
The code looks good!  Looking at the two styles side-by-side I think there might be two small tweaks left.
image

1) For "card style without subtitle", it seems like the gap between the title and the metadata is a little small.

2) For "simple style with and without subtitles", it seems like the meta data is pushed to the right a little too far.

image

petertgiles (approved):
Looks nice!  I recommend a Josh review before merging but it looks good to me.

JoshBeveridge (dismissed):
@yonikid15 sorry for the wait! Couple small notes:
* Chevron stroke weight should be thicker; in Figma it's set to `2.5`, but I don't think that translates to a usable unit for you. Just eye it.
* The subtitle text should be `caption` sized
* The optional icon on the right hand side, along with the context, should always be vertically aligned to the center of the title text (see screenshot 1)
* The actions list is a bit too far to the right; it should be left aligned with the title/subtitle/content (see screenshot 2)
* Bullets in the action list should use `black.lighter`
* Links in the action list should be bolded and use the `caption` font size
* Plaintext items in the action list should use `black.light`
* On desktop, the gap between the accordion title, subtitle, and action list should be a consistent `x.15`
* On mobile, the gap between the accordion title and subtitle should stay `x.15`, but the gap between those and the action list should increase to `x.5`; the gap between each item in the action list should also be `x.5` on mobile (this just gives them a bit more click space for thumbs)
* On mobile, the "additional context" text should be left aligned with the other elements (see screenshot 3) and also be `x.5` away from the title/subtitle group and action list

image
---
image
---
image

yonikid15:
- [ ] Chevron stroke weight should be thicker; in Figma it's set to 2.5, but I don't think that translates to a usable unit for you. Just eye it.
- [x] The subtitle text should be caption sized
- [x] The optional icon on the right hand side, along with the context, should always be vertically aligned to the center of the title text (see screenshot 1)
- [x] The actions list is a bit too far to the right; it should be left aligned with the title/subtitle/content (see screenshot 2)
- [x] Bullets in the action list should use black.lighter
- [x] Links in the action list should be bolded and use the caption font size
- [x] Plaintext items in the action list should use black.light
- [x] On desktop, the gap between the accordion title, subtitle, and action list should be a consistent x.15
- [x] On mobile, the gap between the accordion title and subtitle should stay x.15, but the gap between those and the action list should increase to x.5; the gap between each item in the action list should also be x.5 on mobile (this just gives them a bit more click space for thumbs)
- [x] On mobile, the "additional context" text should be left aligned with the other elements (see screenshot 3) and also be x.5 away from the title/subtitle group and action list

Thanks @JoshBeveridge. I got everything done, except for the Chevron stroke weight. It just wouldn't change no matter what I tried :cry:

yonikid15:
@petertgiles I think you'll have to confirm the changes are correct, whenever you have a chance.

petertgiles:
I can't figure out the stroke weight, either.  There are stroke-weight styles in the file but it seems like they're ignored.  Maybe this stopped working at some point?  :shrug: 

Everything else looks good except the third point.  It looks like the top of the context is aligned but it should be through the middle.
image
It looks like there's a medium width where it's worse.
image

yonikid15:
> I can't figure out the stroke weight, either. There are stroke-weight styles in the file but it seems like they're ignored. Maybe this stopped working at some point? 🤷
> 
> Everything else looks good except the third point. It looks like the top of the context is aligned but it should be through the middle.

Nice catch! I believe it's fixed now :crossed_fingers: [670fac6](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11674/commits/670fac6b49d0cfe088fc04d89cfd3444d7cd1189)

petertgiles:
> Nice catch! I believe it's fixed now

Looks nice!  :rocket:

PR #11676, opened 2024-10-02
[Fix] Prefer rejecting errors

### LINKED ISSUE #11382: ♻️ Prefer promise reject errors
## ♻️ Debt/Refactor

Enable and fix the `prefer-promise-reject-errors` rule.

## 🕵️ Details

https://typescript-eslint.io/rules/prefer-promise-reject-errors

We have some submit handlers that reject without raising an error to be handled. We should always be rejecting promises with a provided error to be handled at some point.

> [!NOTE]
> This may also be fixed with the changes that may be required in https://github.com/GCTC-NTGC/gc-digital-talent/issues/11379

## 🙋‍♀️ Proposed Solution

Adjust our promises to reject with an error.

## ✅ Acceptance Criteria

- [ ] `prefer-promise-reject-errors` enabled
- [ ] All errors fixed

esizer:
🤖 Resolves #11382 

## 👋 Introduction

Re-enables and fixes errors for preferring rejecting errors from promises.

## 🧪 Testing

1. Confirm lint rules pass
2. Confirm changes make sense


apps/web/src/components/NotificationList/mutations.ts
```diff
@@ -47,12 +47,12 @@ export function useMarkAsRead(
     return executeMutation({ id })
       .then((res) => {
         if (!res.data?.markNotificationAsRead) {
-          return Promise.reject();
+          return Promise.reject(new Error(res.error?.toString()));
```
brindasasi, line 50: can we handle the case when res.error is null or undefined ? 
Something like `new Error(res.error?.toString() || 'Unknown error')`

apps/web/src/components/NotificationList/mutations.ts
```diff
@@ -47,12 +47,12 @@ export function useMarkAsRead(
...
         }
 
         return Promise.resolve();
       })
-      .catch(() => Promise.reject());
+      .catch(() => Promise.reject(new Error()));
```
brindasasi, line 55: could we catch the err object here and read the message property from that if available?

apps/web/src/components/NotificationList/mutations.ts
esizer: Sounds good! Updated in [better error handling for notifications list](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11676/commits/f3db2605e544892fec2a88562a2b5d4ddef441b5)

apps/web/src/components/NotificationList/mutations.ts
esizer: Yeah, that makes sense :+1: Added in [better error handling for notifications list](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11676/commits/f3db2605e544892fec2a88562a2b5d4ddef441b5)

brindasasi (approved):
LGTM!

PR #11712, opened 2024-10-08
[Fix] Refactor flaky claim test

### LINKED ISSUE #11586: 🧪 Pool candidate claim verification sort flaky

petertgiles:
🤖 Resolves #11586 

## 👋 Introduction

This fixes the flaky claim sorting tests and performs some refactoring.

## 🕵️ Details

As far as I could tell, the two main issues with the tests were that the pool candidate factory would often change attributes of created models and that without bookmarks some of the items would change order in the sorted lists.

I solved the first problem by having the factor just change attributes if they ended up inconsistent.

I solved the second problem by just removing  the ambiguously sorted items from the unbookmarked test.

I also split the one giant test into three smaller tests in their own file to make it clearer what each test is doing and which tests fails, if one of them does.

## 🧪 Testing

1. Rerun the tests several times to ensure stability.
2. Ensure the tests accurately test what they're supposed to.


esizer (approved):
LGTM! :crossed_fingers:

PR #11716, opened 2024-10-08
[Debt][Security] Refactor updateRoles user policy

### LINKED ISSUE #11204: 🐛 UserPolicy->updateRoles has unsafe fall through
## 🐛 Bug

We're trying to ensure that all our security logic has a safe fall through for unrecognized states.  Our UserPolicy->updateRoles policy has an "else true" condition at the end which does not follow this.

## 🦋 Expected Behaviour

Policy methods should end with "else false" style logic to ensure that any unhandled scenario "fail safely".

## 📋 Steps to Reproduce

https://github.com/GCTC-NTGC/gc-digital-talent/blob/5b311f336a1336e640a77df58f3dc12c513dbde2/api/app/Policies/UserPolicy.php#L135

## ✅ Acceptance Criteria

- [ ] UserPolicy->updateRoles has an "else false" style ending
- [ ] policy check logic remains correct

yonikid15:
🤖 Resolves #11204 

## 👋 Introduction

- Refactors UserPolicy->updateRoles  to have an "else false" style ending. This will ensure that the security logic has a safe fall through for unrecognized states.

## 🧪 Testing

1. Build app
2. Login as different user roles and ensure updating roles works correctly. It can get confusing remembering which roles can update others. I'd suggest viewing the [UserPolicyTest.php](https://github.com/GCTC-NTGC/gc-digital-talent/blob/11204-fix-user-policy-update-roles/api/tests/Unit/UserPolicyTest.php#L290) and [rolePermissions.php](https://github.com/GCTC-NTGC/gc-digital-talent/blob/11204-fix-user-policy-update-roles/api/config/rolepermission.php#L795-L1224)

For example: A community recruiter can only update roles of pools in their community.


yonikid15:
> Hmm, this logic doesn't look quite right to me. The original looped through each role and completed only if all were true. The new one loops through each role and completes early if any of them are true.
> 
> I'd suggest using something like [every](https://laravel.com/docs/11.x/collections#method-every) to evaluate each array then end with something like
> 
> ```
> if ($attachRolesAreValid && $detachRolesAreValid) {
> return true;
> }
> //fallthrough
> return false;
> ```

Wow, good catch @petertgiles  :+1: Hopefully I got it right this time [8ca8d05](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11716/commits/8ca8d058be4e8c9f7faeec9aec04344972c7598b)

api/app/Policies/UserPolicy.php
```diff
@@ -89,50 +89,49 @@ public function updateRoles(User $user, $args)
...
         $detachRoles = isset($args['roleAssignmentsInput']) && isset($args['roleAssignmentsInput']['detach']) ?
-            $args['roleAssignmentsInput']['detach'] : null;
+            $args['roleAssignmentsInput']['detach'] : [];
 
         if (is_null($targetUserId)) {
             return false;
         }
 
-        if (! empty($attachRoles)) {
+        if (empty($attachRoles) && empty($detachRoles)) {
```
petertgiles, line 100: Having an early "return true" make me uncomfortable.  If we add more checks later in the funcation we might inadvertantly jump over them.  I like that you already coalesced the arrays to an empty array earlier.  I think that makes this check redundant.  According to the documentation: 

> If the collection is empty, the every method will return true:

api/app/Policies/UserPolicy.php
```diff
@@ -89,50 +89,49 @@ public function updateRoles(User $user, $args)
...
+            // loop through each element and check
+            if (isset($roleInput['teamId'])) {
+                return $this->teamAbleToCheck($user, $roleInput['roleId'], $roleInput['teamId']);
+            } else {
+                return $this->individualAbleToCheck($user, $roleInput['roleId']);
             }
-        }
 
-        if (! empty($detachRoles)) {
+            return true;
```
petertgiles, line 115: Falling through to "true" is what we're trying to get rid of.  :grin:

api/app/Policies/UserPolicy.php
```diff
@@ -89,50 +89,49 @@ public function updateRoles(User $user, $args)
...
             return false;
         }
 
-        if (! empty($attachRoles)) {
+        if (empty($attachRoles) && empty($detachRoles)) {
+            return true;
+        }
 
-            foreach ($attachRoles as $roleInput) {
+        $canAttachRoles = false;
```
petertgiles, line 104: This initialization is probably unnecessary.  We only do one unconditional assignment to these later.

petertgiles (changes requested):
Sorry, I feel like I'm being really picky.  But this is a security refactor...

api/app/Policies/UserPolicy.php
yonikid15: That's fair! I removed this in [17b2a82](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11716/commits/17b2a82d07fa6e5d12141d66e2f9b81e1cea83d0)

api/app/Policies/UserPolicy.php
yonikid15: :upside_down_face: Removed in [17b2a82](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11716/commits/17b2a82d07fa6e5d12141d66e2f9b81e1cea83d0)

api/app/Policies/UserPolicy.php
yonikid15: Ya that's true, removed in [17b2a82](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11716/commits/17b2a82d07fa6e5d12141d66e2f9b81e1cea83d0)

yonikid15:
> Sorry, I feel like I'm being really picky. But this is a security refactor...

No sorry needed, this is important and we should get it right :+1:

petertgiles (approved):
Nice, looks very clean.  I was able to log in as admin and update roles as usual.  Looks like we have some very thorough unit tests on this policy, too.

PR #11721, opened 2024-10-09
[Fix] PHP undefined property errors

### LINKED ISSUE #11590: ♻️ Larastan: Access to undefined property
## ♻️ Debt/Refactor

Fix all "Access to undefined property" errors.

## 🙋‍♀️ Proposed Solution

REF: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property

Do the following for each ignore line in the config file

 - Confirm the property exists
 - Add PHPDoc to type hint
 - Remove ignore in `phpstan.neon` file

## ✅ Acceptance Criteria

- [ ] No longer ignoring any errors for access to undefined property
- [ ] Larastan passes analysis with no errors

esizer:
🤖 Resolves #11590 

## 👋 Introduction

Fix PHP errors related to access undefined properties.

## 🕵️ Details

As with all PHPStan PRs, this is mostly just adjusting type hints. As a result, some `call to undefined methods` were also fixed.

## 🧪 Testing

1. Confirm no more instances of `undefined property` are ignored
2. Confirm PHPStan check passes


api/app/Models/CommunityExperience.php
```diff
@@ -17,11 +16,11 @@
  * @property string $title
  * @property string $organization
  * @property string $project
- * @property Illuminate\Support\Carbon $start_date
- * @property Illuminate\Support\Carbon $end_date
+ * @property ?\Illuminate\Support\Carbon $start_date
```
esizer, line 19: Pretty sure we should always have this but since we store this as json, we can never be sure so it is safter to mark as nullable

api/app/Rules/SkillLevelRequiredIfEssential.php
```diff
@@ -54,8 +54,8 @@ public function validate(string $attribute, mixed $incomingLevel, Closure $fail)
         if (! is_null($templateId)) {
             $template = JobPosterTemplate::findOrFail($templateId);
             $model = $template->skills()->firstWhere('skill_id', $skillId);
-            $currentType = $model->pivot->type;
-            $currentLevel = $model->pivot->required_skill_level;
+            $currentType = $model->getAttributeValue('type');
+            $currentLevel = $model->getAttributeValue('required_skill_level');
```
esizer, line 58: This is to avoid it since `pivot` is a magic getter. I can add it in the PHPDoc for `PoolSkill` but that also seemed weird :thinking:

api/app/Models/AwardExperience.php
```diff
@@ -15,12 +14,12 @@
...
  * @property string $issued_by
- * @property Illuminate\Support\Carbon $awarded_date
+ * @property ?\Illuminate\Support\Carbon $awarded_date
  * @property string $awarded_to
  * @property string $awarded_scope
  * @property string $details
- * @property Illuminate\Support\Carbon $created_at
- * @property Illuminate\Support\Carbon $updated_at
+ * @property \Illuminate\Support\Carbon $created_at
+ * @property \Illuminate\Support\Carbon $updated_at
```
brindasasi, line 22: Are we gonna use leading slash for all properties ? 
There are still left a lot this for $created_at  and probably for $updated_at too 
Screenshot 2024-10-10 at 1 19 18 PM

brindasasi:
looking great and working without any regression. Just a question about the left overs .. other than that good to go ! 👍

brindasasi (approved):
Looking great! 👍

PR #11718, opened 2024-10-09
[Debt] Laravel11 upgrade

brindasasi:
🤖 Resolves #9754 

## 👋 Introduction

Upgrade to Laravel 11 

## 🧪 Testing

1. Run Composer Update
2. Verify all tests pass 
3. Verify application behaves normally without any issues

PR #11739, opened 2024-10-10
[Fix] PHP parameter errors

### LINKED ISSUE #11596: ♻️ Larastan: Parameter #X {param} of  ... expects X, Y given
## ♻️ Debt/Refactor

Fix all "Parameter #X {param} of  ... expects X, Y given" errors and un-ignore

## 🙋‍♀️ Proposed Solution

- Confirm accurate type being passed in
- If it is, improve type hinting
- If not, update the type to be what the method accepts

## ✅ Acceptance Criteria

- [ ] All errors fixed
- [ ] No longer ignored in config
- [ ] Larastan passes

esizer:
🤖 Resolves #11596 

## 👋 Introduction

Fixes PHPStan errors related to method parameters.

## 🕵️ Details

Most of these were passing lists instead of assoc arrays to validation errors (key must be field with error).

A weird one was the callable signature for the graphql service provider. It is correct but PHPStan was not happy so I just cast it :woman_shrugging: 

## 🧪 Testing

1. Confirm no parameter errors ignored
2. Confirm linting passes
3. Confirm changes make sense


api/app/Models/Experience.php
```diff
@@ -35,13 +35,14 @@ class Experience extends Model
...
-     * @param  object  $attributes
+     * @param  mixed  $attributes
      * @param  string|null  $connection
      * @return static
      */
     public function newFromBuilder($attributes = [], $connection = null)
     {
-        $model = $this->newInstanceFromType($attributes->experience_type);
+        /** @disregard Even though it is typed as an array, it is actually a class */
+        $model = $this->newInstanceFromType(((object) $attributes)->experience_type);
```
esizer, line 45: This is the confusing bit. The `Model` class this extends expected `array $attributes = []` but we actually get an `object` so the cast works but is a bit hacky :cry:

brindasasi:
linting fails on the same error 
 Line   Http/Controllers/AuthController.php                                              
 ------ --------------------------------------------------------------------------------- 
         Ignored error pattern #^Call to an undefined method                              
         Illuminate\\Http\\Client\\Response\:\:getBody\(\)\.$#                            
         in path                                                                          
         /Users/admin/code/gc-digital-talent/api/app/Http/Controllers/AuthController.php  
         was not matched in reported errors.                                              
 ------ --------------------------------------------------------------------------------- 
which I removed in my [pr](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11718) .
May want to wait until it gets merged in ?

brindasasi (approved):
LGTM!

PR #11742, opened 2024-10-10
[Debt] Update to use `@date-fns/tz`

### LINKED ISSUE #11570: ♻️ Update `date-fns`, use first-class timezone support
## ♻️ Debt/Refactor

`date-fns` has a new major version that includes first-class support for timezones so we should probably bump to the major version and update our timezone stuff.

## 🕵️ Details

- [Release post](https://blog.date-fns.org/v40-with-time-zone-support/)
- [Documentation](https://date-fns.org/v4.0.0/docs/Time-Zones)

While I don't think the actual `date-fns` libarary has any breaking changes, I made this a separate ticket instead of doing through dependabot so we can implement the new timezone stuff since it is not clear that the current `date-fns-tz` is compatible.

Also, timezone stuff is tricky so we should pay close attention to it and go through a thorough review process.

## 🙋‍♀️ Proposed Solution

- Bump `data-fns` to latest version
- Remove `data-fns-tz`
- Install `@date-fns/tz` and possibly `@date-fns/utc`?
- Re-implement the `data-fns-tz` usage to use `@date-fns/tz` instead

## ✅ Acceptance Criteria

- [ ] `date-fns` is on latest version
- [ ] `data-fns-tz` replaced with `@date-fns/tz`
- [ ] Date utilities still function exactly as expected

esizer:
🤖 Resolves #11570 

## 👋 Introduction

Replaces the `date-fns-tz` library with `@date-fns/tz`

## 🕵️ Details

Hopefully our tests are good for this :sweat_smile: 

## 🧪 Testing

1. Confirm no references to `data-fns-tz`
2. Confirm helpers that use timezones still function as expected

### DIFF
--- a/packages/date-helpers/src/index.ts
+++ b/packages/date-helpers/src/index.ts
@@ -76,24 +76,18 @@ export const relativeClosingDate = ({
   timeZone,
   customFormat,
 }: relativeClosingDateOptions): string => {
-  const inTz = timeZone ? tz(timeZone) : undefined;
-  // A date formatting function that can use time zones optionally
-  const myFormatFunc = (
-    date: Date,
-    formatPattern: string,
-    options?: FormatOptions,
-  ) =>
-    format(date, formatPattern, {
-      ...options,
-      in: inTz,
-    });
+  const formatOpts: FormatOptions = {
+    in: timeZone ? tz(timeZone) : undefined,
+  };
 
   const strLocale = getLocale(intl);
   const locale = strLocale === "fr" ? fr : undefined;
-  const time = myFormatFunc(closingDate, `p`, {
+  const time = format(closingDate, `p`, {
+    ...formatOpts,
     locale,
   });
-  const dateTime = myFormatFunc(closingDate, customFormat ?? `PPP p`, {
+  const dateTime = format(closingDate, customFormat ?? `PPP p`, {
+    ...formatOpts,
     locale,
   });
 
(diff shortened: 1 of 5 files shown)


packages/date-helpers/src/index.ts
```diff
@@ -75,12 +76,17 @@ export const relativeClosingDate = ({
...
   customFormat,
 }: relativeClosingDateOptions): string => {
+  const inTz = timeZone ? tz(timeZone) : undefined;
   // A date formatting function that can use time zones optionally
-  const myFormatFunc = timeZone
-    ? (date: Date, formatPattern: string, options?: { locale?: Locale }) =>
-        formatInTimeZone(date, timeZone, formatPattern, options)
-    : (date: Date, formatPattern: string, options?: { locale?: Locale }) =>
-        format(date, formatPattern, options);
+  const myFormatFunc = (
```
petertgiles, line 81: You probably don't need `myFormatFunc` anymore since it is always the same function now.  You can just call `format` directly.

petertgiles (approved):
It works great!  I played with the pool closing stuff and it works as before, as far as I can tell.  I noted one refactoring possiblity, too.

petertgiles:
I'm noticing some "Unsafe assignment of an error typed value." warnings in my editor that GH isn't flagging.  Might be worth double-checking that.
image

packages/date-helpers/src/index.ts
esizer: Perfect, thanks. I was not 100% sure about removing it so I left it there initially. It has been removed in [remove format func](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11742/commits/b0ba8d4430b7d18910878f1a7cc2fafb99ce91fc)

esizer:
> I'm noticing some "Unsafe assignment of an error typed value." warnings in my editor that GH isn't flagging. Might be worth double-checking that.

Hmm, not sure. I could not get it showing up in my editor or running `pnpm run lint --force` :thinking:

packages/date-helpers/src/index.ts
petertgiles: Nice.

PR #11745, opened 2024-10-10
[🦔] Data loaders

esizer:
## 👋 Introduction

This is an example of how we could achieve `react-router` data loaders with `urql`.

## 🕵️ Details

Since loaders are regular functions, we cannot use hooks in them. So, we instead get the client in the router and pass it into a builder that returns the loader.

## 🧪 Testing

2. Login as any user
3. Confirm the dashboard loads correctly
4. Navigate away and then return
5. Notice the query is re-run every time even though it should be near instant read from cache


packages/client/src/components/ClientProvider/ClientProvider.tsx
```diff
@@ -69,7 +69,7 @@ const ClientProvider = ({
       client ??
       createClient({
         url: `${apiHost}${apiUri}`,
-        requestPolicy: "cache-and-network",
+        requestPolicy: "cache-first",
```
esizer, line 72: Without this change, the cache is never read. Not sure if we really want this :woman_shrugging:

tristan-orourke:
What about using deferal like in 1a464c7106eab8070a5e2ac38a0cf163a1c69e5b? I think this might give us the best of both worlds:
- query triggered by loader, before components are rendered
- component does not trigger a new query, it uses the loader's (this I'm most worried about, but looks good in my testing?)
- component can manage its own pending state
- continued use of "cache-first-but-refresh" strategy (we could consider switching, but I'm more nervous about combining refactors like that)

esizer:
Closing PR but leaving branch open for reference in future.

PR #11752, opened 2024-10-11
[Test] Add experience regression test

### LINKED ISSUE #11749: 🐛 Can't add a new work experience
## 🐛 Bug

After the update last night I can no longer add Work Experience to my profile. I get an error back from the GraphQL endpoint:

```
{
	"errors": [
		{
			"message": "Variable \"$id\" of non-null type \"ID!\" must not be null.",
			"locations": [
				{
					"line": 1,
					"column": 36
				}
			]
		}
	]
}
```

Specifically I tried to add a certification experience. I can still edit existing experience however.

## 🦋 Expected Behaviour

A new work experience should be added

## 🕵️ Details

- **Operating System**: Ubuntu 24.04
- **Browser**: FireFox 131.0.2

## 📋 Steps to Reproduce

1. Add a new work experience
2. Fill in details
3. Submit

## 📸 Screenshot

image

esizer:
🤖 Resolves #11749 

## 👋 Introduction

Adds a regression test for a previously fixed bug.

> [!NOTE]
> This does not fix the issue as it was already fixed. This is simply adding a regression test for that fix.
>
> Currently only tests work experiences, should we do all different types? :thinking: 

## 🧪 Testing

> [!TIP]
> Reverting the fix means setting the `experienceId` in the `ExperienceFormPage` to coalesce to an empty string
>
> `experienceId={experienceId ?? ""}`

https://github.com/GCTC-NTGC/gc-digital-talent/blob/62e91d475ab8d72281817bf7cb5a9f2c2b1fa64f/apps/web/src/pages/Profile/ExperienceFormPage/ExperienceFormPage.tsx#L685

1. Confirm test has decent coverage
2. Confirm test consistently passes
3. Revert the fix
4. Confirm test fails


vd1992:
Hmm, I think we should probably exercise the experience page a bit more. Not sure if we want to run it for each experience type.
I think updating a existing experience as well as linking skills should be tested too

brindasasi:
Test works fine but I agree with the comment that the we need to test more on 

- adding/editing of each type of experience 
- linking skills for each of the exp
- linking the same experience to different skills. ( we had prod bug with this ) 

I'm also ok to get this merged in and work on them as a sep tkt.

brindasasi (approved):
LGTM!

PR #11753, opened 2024-10-11
[Fix] Allow filling in screening question

### LINKED ISSUE #11748: 🐛 Can't submit an application to a pool with hidden screening questions.
## 🐛 Bug

On the IT-05 pool in prod there is a single screening question.  However, it is not visible during application and is preventing applications from being submitted.

## 📸 Screenshot

image

petertgiles:
🤖 Resolves #11748 

## 👋 Introduction

Fixes a bad null coalescing fix.  In this case, it was actually supposed to be an OR, not a null coalesce.

## 🕵️ Details

The way the logic for the "should I show the questions step" was being calculated was "only show if there are general questions".  This specific pool has no general questions but does have screening questions.  The screening questions were skipped, then.

## 🧪 Testing

1. Set up a pool with zero general questions but at least one screening question.
2. Submit an application to the pool, verifying that you filled in the screening questions.
3. Try with other pools using different permutations of questions.

### DIFF
--- a/apps/web/src/pages/Applications/utils.ts
+++ b/apps/web/src/pages/Applications/utils.ts
@@ -32,8 +32,8 @@ export const getApplicationSteps = ({
   experienceId,
 }: GetApplicationPagesArgs): ApplicationStepInfo[] => {
   const showQuestionStep =
-    application.pool.generalQuestions?.length ??
-    application.pool.screeningQuestions?.length;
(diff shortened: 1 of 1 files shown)


esizer (approved):
Nice :sunglasses: 

Tried with:

- No general, 1 screening question
- 1 of each general and screening
- 1 general, no screening


:heavy_check_mark: on all

PR #11758, opened 2024-10-15
[Bug] Notifications panel flashes double state

### LINKED ISSUE #11604: 🐛 Notification panel flashes doubled state
## 🐛 Bug

When the notification panel is opened it flickers through an intermittent state.  It appears that maybe it's showing double notifications before deduplicating them?

## 🦋 Expected Behaviour

It makes sense to show an old state then suddenly change to a new state but the middle state that represents neither is disorienting.

## 📋 Steps to Reproduce

1. Confirm you have more than one notification in the panel.
2. Generate a file and wait a few seconds so that it is ready.
3. Open the notification panel and watch how it changes before settling.

## 📸 Screenshot

https://github.com/user-attachments/assets/b1164fb7-f825-46a6-9df2-6afe6fc956aa

## ✅ Acceptance Criteria

- [ ] Panel shows only old state or new state.
- [ ] A doubled state is never shown.

yonikid15:
🤖 Resolves #11604 

## 👋 Introduction

- Pauses the live notification query till max pages query is finished. This should prevent the user from seeing any double notification list state.

## 🧪 Testing

1. Create a notification. The quickest way (i think) is to login as admin and create an application. Then on the edit user page, change the status.
image
2. Open up the notifications panel, and ensure no double flash occurs.


tristan-orourke:
Unfortunately, I'm still seeing it consistently in the following case:
1. Log in as admin, go to a candidates table
2. select multiple candidates
3. click download as CSV
4. immediately open notification pane
5. See doubled state as new notification appears

I tried to take a video, but recordit keeps crashing...

tristan-orourke (dismissed):
Unfortunately, I'm still seeing it consistently in the following case:

1. Log in as admin, go to a candidates table
2. select multiple candidates
3. click download as CSV
4. immediately open notification pane
5. See doubled state as new notification appears

I tried to take a video, but recordit keeps crashing...

JamesHuf (approved):
Works for me :+1:

PR #11759, opened 2024-10-15
[Debt] User creation shifted to `authCallback()`

### LINKED ISSUE #11466: ♻️ New user can't be tracked in auth callback
## ♻️ Debt/Refactor

Our current implementation has the new user initialization occurring in the AuthServiceProvider than runs whenever an authenticated request comes through.  Unfortunately, this means that when we record a successful login in the authcallback it is possible that the user row doesn't exist yet and we can't track it.

## 🕵️ Details

https://github.com/GCTC-NTGC/gc-digital-talent/pull/11437#discussion_r1743735146

## 🙋‍♀️ Proposed Solution

I think we should move the new user initialization to the auth callback.  This will allow us to track new users.  Conceptually, I think it makes more sense to potentially initialize new users on login rather than every request, anyway.

## ✅ Acceptance Criteria

- [ ] New users are initialized in the auth callback instead of the auth service provider
- [ ] the login timestamp is updated for new users as well as returning users
- [ ] Works with mock auth and sign in canada

## 🛑 Blockers

```[tasklist]
### Blocked By
```

vd1992:
🤖 Resolves #11466

## 👋 Introduction

Shift user creation to the callback in AuthController.php
Nothing fancy needed to be done

## 🧪 Testing

1. Can log in normally with existing user
2. User creation works
3. Sign in values updated or set appropriately
4. Test above with SiC

### DIFF
--- a/api/tests/Feature/AuthServiceProviderTest.php
+++ b/api/tests/Feature/AuthServiceProviderTest.php
@@ -7,7 +7,6 @@ use App\Models\Role;
 use App\Models\User;
 use App\Providers\AuthServiceProvider;
 use App\Services\OpenIdBearerTokenService;
-use Database\Seeders\RolePermissionSeeder;
 use Illuminate\Foundation\Testing\RefreshDatabase;
 use Lcobucci\JWT\Token\DataSet;
 use Lcobucci\JWT\Validation\ConstraintViolation;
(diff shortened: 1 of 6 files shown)


api/tests/Feature/AuthServiceProviderTest.php
vd1992, line 1: We've no testing for auth controller

vd1992:
Thank you Larastan, that was indeed something to fix, done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/11759/commits/14e866ebf30fa2f42ae1aac329b06ae3e985188b

petertgiles (changes requested):
This works great!  I tried it with mock auth and SIC and it works well.  :medal_sports:  I'm worried about losing that test, though.  This functionality it too important to not have good testing, IMO.  Could you rewrite the test for this?  Let me know if you want a hand.

vd1992:
> This works great! I tried it with mock auth and SIC and it works well. 🏅 I'm worried about losing that test, though. This functionality it too important to not have good testing, IMO. Could you rewrite the test for this? Let me know if you want a hand.

Had a nice learning experience and got something in https://github.com/GCTC-NTGC/gc-digital-talent/pull/11759/commits/2169efed5087c8df788a6a3869dc97cda416f80f

vd1992:
Not sure what is going on with the PHP test 😕

petertgiles (approved):
I love this test!  Very easy to read and understand.  It's a great base to use in the future for more testing, too.

I deployed this branch to Azure and it works well, too.

I'm trying a few things on the phpunit test to try and get that to complete in GH.

PR #11792, opened 2024-10-17
[PATCH] Release/2.33.3

brindasasi:
🤖 cherry pick only the fix for https://github.com/GCTC-NTGC/gc-digital-talent/issues/11749 
from https://github.com/GCTC-NTGC/gc-digital-talent/pull/11719 

## 🧪 Testing

Try add an experience to your profile


petertgiles (approved):
Seems good.  That looks like Eric's fix.  I confirm that I can add/edit/delete experiences both in the career timeline and in an application flow.

brindasasi:
This is in production now.

PR #11803, opened 2024-10-21
[Debt] Bumps `ts-jest`

### LINKED ISSUE #11364: ♻️ Bump `ts-jest` from 29.2.3 to latest
## ♻️ Debt/Refactor

For some reason this bump breaks

`src/lang/lang.test.ts`

## 🕵️ Details

PR https://github.com/GCTC-NTGC/gc-digital-talent/pull/11360

Test

image

## ✅ Acceptance Criteria

- [ ] Dependency updated
- [ ] Tests pass

mnigh:
🤖 Resolves #11364.

## 👋 Introduction

This PR bumps the version of `ts-jest` to its latest version and updates a file path for it to work.

## 🧪 Testing

1. `cd apps/web`
2. `pnpm test -- lang.test.ts`
3. Verify test passes

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -94,7 +94,7 @@
     "jest-diff": "^29.7.0",
     "jest-fail-on-console": "^3.3.0",
     "prettier": "^3.3.3",
-    "ts-jest": "^29.2.3",
+    "ts-jest": "^29.2.5",
     "ts-node": "^10.9.2",
     "tsconfig": "workspace:*",
     "typescript": "^5.6.3",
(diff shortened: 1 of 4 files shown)


petertgiles:
I don't love this solution, tbh.  Maybe adding the i18n package to the devDependencies would do the trick?

mnigh:
> I don't love this solution, tbh. Maybe adding the i18n package to the devDependencies would do the trick?

I like your solution, but it doesn't seem to work in CI (it does locally though). Thoughts?

petertgiles:
I posted my suggested fix

petertgiles:
... and it didn't work.  :laughing:

mnigh:
> ... and it didn't work. 😆

If I revert to my original solution, would that be acceptable?

petertgiles:
I don't really like it - it seems like a hack.  I think this patch version bump is not so urgent that we need to resort to that yet.

PR #11804, opened 2024-10-21
[Feature] Send 'new pool' emails overnight

### LINKED ISSUE #11709: ✨ Send new poster notifications overnight instead of immediately
## ✨ Feature

Instead of sending notifications about new posters immediately, send them overnight after they have been published. This will allow for a time window to unpublish a poster if a mistake has been made, before emails go out, and if multiple posters are published in a single day we can combine the notifications.

## 🕵️ Details

Instead of queuing a notification job when the job is initially published, set up a job that runs overnight and checks for newly published jobs. 

QUESTION: We talked about combining notifications if multiple posters are published the same day, but how would that look? Can GC Notify templates handle list variables? Would we need to assemble the list markup on the server side and send html as a single variable? What about our in-app notifications, is it easy to change the api for a notification type? Should we send one email but multiple in-app notifications?

ANSWER: We'll need a new GC Notify template, and the list of posters will need to be generated in markup on the server and passed as a single variable. We'll pull this out of scope and into its own ticket.

## ✅ Acceptance Criteria

- [ ] New Job Posted notifications are no longer sent when jobs are published
- [ ] When daily jobs run overnight, a job checks for posters published the previous day, and triggers notifications for those jobs
- [ ] ~QUESTION: if multiple jobs were published in one day, only send a single notification.~ Will split into separate ticket.
- [ ] Increase the time limit on queued jobs from 10 minutes to 30 minutes
- [ ] Still does not send for publishing group OTHER

petertgiles:
🤖 Resolves #11709 

## 👋 Introduction

Sends "new pool published" emails overnight instead of immediately.

## 🕵️ Details

One of the AC is:
>  Increase the time limit on queued jobs from 10 minutes to 30 minutes

This is no longer relevant since the script is running from the cron job and not a task scheduler worker.  It no longer has a time limit - I think.  :sweat_smile: 

## 🧪 Testing

1. Set the GC Notify "team" API key
2. Rebuild the app
3. Start the queue worker
4. Set your real work email address to a user and opt in to JOB_ALERT email alerts
5. Publish a pool with the IT_JOBS publishing group
6. Run `./artisan send-notifications:pool-published`
7. Ensure you get a notification email

Also:
- [ ] More than one published pool -> more than one email
- [ ] OTHER publishing group -> no email


brindasasi (commented):
image
22k notifications got processed in Dev without any issues. 
Everything looks good. This is good to go once the linting & failing tests are fixed.

api/app/Console/Commands/SendNotificationsPoolPublished.php
```diff
@@ -0,0 +1,103 @@
...
+        $this->info('SendNotificationsPoolPublished running at '.Carbon::now()->toDateTimeString().'.');
+
+        $successCount = 0;
+        $failureCount = 0;
+
+        $endOfSpan = Carbon::now();
+        $startOfSpan = Carbon::now()->subHours(24); // assuming the last reporting job ran about 24 hours ago
+        $this->info("Finding pools published between $startOfSpan and $endOfSpan.");
+
+        $poolsPublishedRecently = Pool::query()
```
brindasasi, line 57: one more question .. this pr checks the pools published within 24 hours but doesn't check for open pools. 
In the use case , they have published something accidentally and they closed it immediately, we would still send notification for that closed pool.

I'm wondering should we add that check for only active pools?

petertgiles: Good catch.  If I had remembered to update the tests first, this would have been caught.  :facepalm:

brindasasi (approved):
This is working great! Awesome 🥇

PR #11812, opened 2024-10-21
Avoid loading teamable in myAuth query

### LINKED ISSUE #11811: 🐛 Process Operators see toast error on non-admin pages
## 🐛 Bug

When a user with the new Process Operator role refreshes the page on a non-`/admin` route, they see a toast error saying "this action is unauthorized".

## 🦋 Expected Behaviour

No auth errors just for visiting home pages!

## 🕵️ Details

1. The auth error is triggered by the `myAuth` authorization query. This query runs when logged-in, and queries all of the user's roleAssignments, including the `teamable` objects they're associated with. 
2. The API checks if the user has permission to view these teamables before returning them. 
3. If the user has the processOperator role for a pool which is still in Draft, the only permission which lets them view it is from the ProcessOperator role.
4. _However_, this is a privileged role, and as such only works when on `/admin` routes!
5. Therefore, non-admin pages through an authorization error.

## 📋 Steps to Reproduce

1. Log in as community@test.com and create a draft pool.
2. Add a user to the pool using the Manage Access tab. (This gives the user a processOperator role.)
3. Sign out, and sign back in as the Process Operator role.
4. When refreshing the page on home page, or applicant dashboard, see error toast
5. To to /admin, and refresh. No toast error.

## 🙋‍♀️ Proposed Solution

Remove `teamable` from myAuth query? Do we need it for anything?

## ✅ Acceptance Criteria

- [ ] Process Operators of draft pools no longer see auth error on home and applicant pages.
- [ ] Regression browser test - process operator of draft pool can view home page without error toast

tristan-orourke:
🤖 Resolves #11811

## 👋 Introduction

Removes `teamable` from myAuth query.

## 🕵️ Details

The main thing to review and test here is that this teamable object isn't used anywhere else.

### DIFF
--- a/packages/auth/src/components/AuthorizationContainer.tsx
+++ b/packages/auth/src/components/AuthorizationContainer.tsx
@@ -6,8 +6,10 @@ import {
   UserAuthInfo,
 } from "@gc-digital-talent/graphql";
 
+type SimpleRoleAssignment = Exclude<RoleAssignment, "teamable">;
+
 export interface AuthorizationState {
-  roleAssignments: Maybe<RoleAssignment[]>;
+  roleAssignments: Maybe<SimpleRoleAssignment[]>;
   userAuthInfo?: Maybe<UserAuthInfo>;
   isLoaded: boolean;
 }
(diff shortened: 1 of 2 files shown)


brindasasi:
I've verified this solved the dashboard error for applicants and did some regression around adding work experience, linking skills , viewing applications and submitting applications. 
Leaving it to see if @GCTC-NTGC/developers has any concern with this approach.

brindasasi (approved):
LGTM!

PR #11818, opened 2024-10-22
[FIX] Close spreadsheet memory leak

### LINKED ISSUE #11760: ♻️ Improve memory usage in PoolCandidateCsvGenerator
## ♻️ Debt/Refactor

Recently, when trying to download a full pool of candidates as a CSV, the job would fail silently. We found that PoolCandidateCsvGenerator, when running in a job, would cause an out-of-memory error and crash. Currently, we have no way to detect and properly log this error or notify the user. However, we may be able to improve memory usage to avoid the problem entirely.

## 🕵️ Details

The CSV I was trying to generate had 70 columns and 270 rows (270 candidates), and we were using up all 128 Mb of memory. Bumping memory limit up to 256 Mb solved the issue - for now. 

## 🙋‍♀️ Proposed Solution

[According to PHPSpreadsheet](https://phpspreadsheet.readthedocs.io/en/latest/topics/memory_saving/), each cell takes up approximately 1 kb of memory. But 270*70=18 900kb=19mb, which is a lot but doesn't explain the crash. I also doubt holding 200 candidates in memory at a time would cause the error - though we should double check if reducing the chunk size helps. I suspect there's some kind of memory leak. We may need to make sure to [manually clear workbooks from memory](https://phpspreadsheet.readthedocs.io/en/latest/topics/creating-spreadsheet/#clearing-a-workbook-from-memory).

## ✅ Acceptance Criteria

- [ ] Downloading a spreadsheet of 300 real candidates shouldn't cause an out-of-memory error in UAT or Production (with 128 mb memory).

petertgiles:
🤖 Resolves #11760 

## 👋 Introduction

Closes a memory leak in the spreadsheet generator.

## 🕵️ Details

There is definitely a leak in the spreadsheet generator if you don't cleanup the worksheets when done.  You can observe this leak by deleting the new destructor and following the testing instructions below.

[The instructions for cleaning up the spreadsheets](https://phpspreadsheet.readthedocs.io/en/latest/topics/creating-spreadsheet/#clearing-a-workbook-from-memory) worked perfectly.  I added them to the destructor of the class.  Now the memory usage peaks at around 16 MB.

## 🧪 Testing

1. Check out the [testing code](https://github.com/GCTC-NTGC/gc-digital-talent/commit/79298ee8d57075a49608a002b0b456b701b99549) commit to have access to the testing tools I wrote.
2. Run `./artisan app:make-csv-files` to queue up 1000 CSV files.
3. Launch the queue worker and confirm that all 1000 files could be generated without running out of memory.
4. Open api/storage/logs/jobs.log to observe the memory usage climbing then dropping when a garbage collection runs.
5. Repeat with `app:make-docx-files` and `app:make-zip-files` to confirm that those generators are not leaking memory.

## 📸 Screenshot

image


brindasasi (approved):
I've got much longer time to execute each GenerateUserFile in my local. It's probably specific to my system. But I didn't get any out of memory error which is good. 
LGTM!

PR #11865, opened 2024-10-28
[Fix] Configure rate limiter based on App Insights user

### LINKED ISSUE #11852: 🐛 Prod - Regular users hitting rate limiter
## 🐛 Bug

After a firewall report from IMTD I found that regular users were hitting the rate limiter.  Their logs looking like regular browsing so this is almost certainly a bug.  

## 🦋 Expected Behaviour

They shouldn't be hitting the limiter unless they made 600 requests in a minute from the same IP address.
## 🕵️ Details

Query for 10/24/2024
```
AppServiceHTTPLogs 
| where Cookie contains "YAJYOZnKsMQ8eVUZL0VUrc" and CsUriStem =='/graphql'
| project TimeGenerated, CIp, ScStatus, Referer
```
Amost all traffic appears to come from the same IP address.  That address was confirmed to be the Azure gateway by IMTD.  Therefore, all unauthenticated traffic is being rate limited from the same bucket due to appearing to be the same IP address.

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

We'll need a new way to bucket unauthenticated users.  Perhaps from the ai_user cookie?

## ✅ Acceptance Criteria

- [ ] The rate limiter can bucket unauthenticated users by some other means than source IP address.

petertgiles:
🤖 Resolves #11852 

## 👋 Introduction

Configures the rate limiter to bucket users based on App Insights ID before IP address.

## 🧪 Testing

1. Check out the [locust test](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11865/commits/f7b68b42097fe8b3a1509cda0978b2137b720686) commit to have access to the Locust test.
2. In `api/.env` set "APP_RATE_LIMIT=60"
3. Restart the Compose network.
4. Visit http://localhost:8089/ and start a test with one worker pointed at http://webserver:8080
5. Observe how a stream of requests over 1.0 RPS can be sustained since the ai_user cookie is changing to simulate unique users.

More tests...
- Log in in a browser to get an auth token and run a test with that header.  Observe how the rate limiter kicks in since the unchanging user is preferred over the cookie.
- Remove the header and fix the cookie to a single string.  Observe how the rate limiter kicks in since the ai_user is not changing.
- Remove the header and the cookie.  Observer how the rate limiter kicks in with the unchanging IP address.

## 📸 Screenshot

image


brindasasi (approved):
- Verified the cookie in the request header has ai_user.
Screenshot 2024-11-01 at 1 18 46 PM

- Verified with different APP_LIMIT (30, 60) with single user and confirmed rps got maintained within the  limit 

Screenshot 2024-11-01 at 1 19 29 PM

- For the multiple users collectively it goes above the rate limit but should be expected I beleive
Screenshot 2024-11-01 at 1 15 24 PM

- verified there was no failures during this test 

LGTM! 👍

PR #11874, opened 2024-10-29
[Copy] Updates error message when a file no longer exists

### LINKED ISSUE #11098: 🐛 File download error is sometimes wrong
## 🐛 Bug

If a user tries to download a file that doesn't exist they get an error message saying that the file is over 24 hours old.  However, if some other unexpected reason (permissions issue, app service deployment, etc...) causes the file to not be available then the error message is wrong.  The app doesn't actually know if the requested file used to exist but was cleaned up.  The user probably does know how the old the file is and the false statement increases confusion.

## 🦋 Expected Behaviour

The error message should probably be worded as "_if_ the file is 24 hours old..." instead of confidently declaring it.

## 🕵️ Details

https://github.com/GCTC-NTGC/gc-digital-talent/blob/aa29935a1fea353eeb638051e990f2b88f3e71a9/apps/web/src/components/NotificationList/NotificationDownload.tsx#L47-L55

## 📋 Steps to Reproduce

1. Generate a file (eg profile document)
2. Manually delete the generated file in the backend
3. Click the notification

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

- This file is not available. If it's been more than 24 hours since you requested it, please try requesting the download again. If it's been less than 24 hours or the problem persists, contact support for assistance.

## ✅ Acceptance Criteria

- [ ] Error message update

mnigh:
🤖 Resolves #11098.

## 👋 Introduction

This PR updates the error message when a file no longer exists.

## 🧪 Testing

1. Generate a file (e.g. profile document)
2. Manually delete the generated file in the backend
3. Click the notification
4. Verify message uses updated copy

### DIFF
--- a/apps/web/src/components/NotificationList/NotificationDownload.tsx
+++ b/apps/web/src/components/NotificationList/NotificationDownload.tsx
@@ -47,8 +47,8 @@ const NotificationDownload = forwardRef<
               intl.formatMessage(
                 {
                   defaultMessage:
-                    "This file is older than 24 hours and is no longer available. Please request a new file. If the problem persists contact support for assistance.",
-                  id: "JFlQjb",
+                    "This file is not available. If it's been more than 24 hours since you requested it, please try requesting the download again. If it's been less than 24 hours or the problem persists, contact support for assistance.",
+                  id: "klyOZZ",
                   description: "Error message when a file no longer exists",
                 },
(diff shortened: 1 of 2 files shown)


petertgiles (approved):
Nice, thanks for grabbing that one.

PR #11938, opened 2024-11-13
[Fix] Remaining PHPStan errors

### LINKED ISSUE #11595: ♻️ Larastan: Call to undefined method
## ♻️ Debt/Refactor

Fix all "Call to undefined method" errors and un-ignore.

## 🙋‍♀️ Proposed Solution

- Confirm the method does not exist
- If it does, provide better type hinting
- If it does not, avoid calling the method

## ✅ Acceptance Criteria

- [ ] All errors fixed
- [ ] No longer ignored in the config
- [ ] Larastan passes

esizer:
🤖 Resolves #11595 

## 👋 Introduction

Fixes the remaining PHPStan errors.

## 🧪 Testing

1. Confirm no more ignored errors in phpstan
2. Confirm PHPStan check passes
3. Review changes made to ensure they make sense


api/app/GraphQL/Mutations/ArchivePool.php
```diff
@@ -14,8 +14,9 @@ final class ArchivePool
      */
     public function __invoke($_, array $args)
     {
+        /** @var Pool $pool */
         $pool = Pool::find($args['id']);
-        if ($pool->getStatusAttribute() !== PoolStatus::CLOSED->name) {
+        if ($pool->status !== PoolStatus::CLOSED->name) {
```
mnigh, line 19: What's the difference between `status` and `getStatusAttribute`? I searched through the codebase and see `getStatusAttribute` used in other places still: should those also be replaced with `status`?

api/app/Models/PoolCandidate.php
```diff
@@ -1178,27 +1177,21 @@ public function computeFinalDecision()
...
-                FinalDecision::DISQUALIFIED_PENDING->name => 200,
-                FinalDecision::DISQUALIFIED->name => 210,
-                FinalDecision::QUALIFIED_REMOVED->name => 220,
-                FinalDecision::TO_ASSESS_REMOVED->name => 230,
-                FinalDecision::REMOVED->name => 240,
-                FinalDecision::QUALIFIED_EXPIRED->name => 250,
-                default => null
-            };
-        } catch (\UnhandledMatchError $e) {
-            Log::error($e->getMessage());
```
mnigh, line 1198: Do we no longer need to log errors here?

api/app/Notify/Client.php
```diff
@@ -417,15 +417,4 @@ private function post(string $endpoint, array $payload, array $headers = []): Re
         return Http::withHeaders($this->buildHeaders($headers))
             ->post(self::BASE_URL.$endpoint, $payload);
     }
-
-    /**
-     * GET
-     *
-     * Make a GET request to the GC Notify API
```
mnigh, line 424: I guess this is not something we are going to use in the future?

api/app/GraphQL/Mutations/ArchivePool.php
esizer: Any method on a model that uses `get*Attribute` creates a magic accessor where you can use the `*` as a property.

Instead of adding the method to the model definition, I used the property directly since it is available. We may want to use the magic getter in the other places instead as well :thinking: I wonder why this is the only instance it flagged.

api/app/Notify/Client.php
esizer: Yeah, I originally added it since it was easy to add but I don't think we will ever use a get request with GCNotify. If we plan on using it, it will be super easy to add back in.

api/app/GraphQL/Mutations/UnarchivePool.php
```diff
@@ -13,8 +13,9 @@ final class UnarchivePool
      */
     public function __invoke($_, array $args)
     {
+        /** @var Pool $pool */
```
petertgiles, line 16: `find` could return null if the model is not found, right?  Then L18 will throw a null dereference?  Same with unarchive?  Maybe these could be switched to findOrFail?

api/app/Models/PoolCandidate.php
```diff
@@ -1178,27 +1177,21 @@ public function computeFinalDecision()
...
-                // Giving a decent buffer to increase max steps
-                FinalDecision::DISQUALIFIED_PENDING->name => 200,
-                FinalDecision::DISQUALIFIED->name => 210,
-                FinalDecision::QUALIFIED_REMOVED->name => 220,
-                FinalDecision::TO_ASSESS_REMOVED->name => 230,
-                FinalDecision::REMOVED->name => 240,
-                FinalDecision::QUALIFIED_EXPIRED->name => 250,
-                default => null
-            };
-        } catch (\UnhandledMatchError $e) {
```
petertgiles, line 1197: Can we still log an error if `default` is matched?  I don't have much experience with PHP pattern matching.  I think we still want to be notified if this falls through, though.

api/app/Rules/HasLanguageRequirements.php
```diff
@@ -35,7 +35,7 @@ public function passes($attribute, $value)
         $userLookingForBilingual = $thisUser->looking_for_bilingual;
         $poolNeedsBilingual = $this->pool->advertisement_language == PoolLanguage::BILINGUAL_INTERMEDIATE->name || $this->pool->advertisement_language == PoolLanguage::BILINGUAL_ADVANCED->name;
 
-        $passes = ! $poolNeedsBilingual || ($poolNeedsBilingual && $userLookingForBilingual);
+        $passes = ! $poolNeedsBilingual || $userLookingForBilingual;
```
petertgiles, line 38: While logically this is correct, I think it makes it harder to understand.  I can easily understand the business rules of the first form: "Either the pool doesn't need bilingual or the pool does need bilingual and the user is also bilingual".  But reading the shorter form, I can't easily see what it's trying to express.  There is an extra mental jump since the fact the the pool does need bilingual is just implied on the right side of the OR.

petertgiles (changes requested):
This is a nice tidy up!  I flagged a few things to check on.

api/app/GraphQL/Mutations/UnarchivePool.php
esizer: Oh yeah, you are right. For some reason I was assuming this was a `findOrFail`. Fixed in [fix pool type](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11938/commits/bfb90010b18263eeddb892704f9440d779f6d795)

api/app/Models/PoolCandidate.php
esizer: Yeah, added the log back in [add log for default match](https://github.com/GCTC-NTGC/gc-digital-talent/pull/11938/commits/d809c647353f02d518c63df783ce08ca53b6f4ca)

api/app/Rules/HasLanguageRequirements.php
esizer: Hmm, we might need a third person to come in and break the tie because I found the exact opposite. The double check of `$poolNeedsBilingual` made it more confusing for me which was why I was so uncertain about this change :sweat_smile:

petertgiles: Yep, we can get some other opinions. :+1:   Another options would be to assign the second part of the expression to its own variable to make it clearer.  Something like

```php
$userLookingForBilingual = $thisUser->looking_for_bilingual;
$poolNeedsBilingual = $this->pool->advertisement_language == PoolLanguage::BILINGUAL_INTERMEDIATE->name || $this->pool->advertisement_language == PoolLanguage::BILINGUAL_ADVANCED->name;
$poolAndUserAreBothBilingual = $poolNeedsBilingual && $userLookingForBilingual;

$passes = ! $poolNeedsBilingual || $poolAndUserAreBothBilingual;
```

petertgiles (approved):

Issue #11864, opened 2024-10-28
✨ Update to work experience - UI changes


gobyrne:
## ✨ Feature

We want to see a history of an employee's work in government for talent management. To this end, we want to collect group and level on the work experience. We also want to be able to match seniority in other orgs to those in gov, so we collect seniority when we do not collect group and level. Afterwards we will eventually deprecate collecting group and level in the "government employee" section.

## 🕵️ Details

There are a lot of things that show and hide, and display in various situations. Here's a list of all the changes and small interactions (but please also c[lick through the design](https://www.figma.com/proto/piRkRtWAHheoEX5Jq8d9u7/Create%2FEdit-an-experience-(Applicants)?node-id=487-58844&node-type=canvas&t=7ZVdFKtLK183qYms-0&scaling=min-zoom&content-scaling=fixed&page-id=472%3A70951)) :

### All experience
1. New question for all work experience "Employment category: External, Gov, CAF"
2. Start date and end date are moved to the bottom of form
3. Add a sentence to the description for "Highlight additional details" section stays at the bottom
4. In the designs the copy for "Current experience" is changed but honestly, we were just too lazy to revert all the designs back to match production. You do not need to change the labels for that checkbox.

### External
5. Add "size of org" question
6. Add "Seniority of the role" question (with "Self-employed" as an option)
7. **Read only state** - only needs to add those two fields, otherwise unchanged

### Gov
8. Instead of org, we ask for "Department, agency or other government organization" and it should be a Drop-down of the departments we have in the database.
9. Ask "Employment type"
10.  For Casual, term, or indeterminate we ask "Group" and "Level"
11. For indeterminate and term, we ask "Position type"
11.5 (**Gray added this late**) For term, we do **not** give the option for "substantive" in "Position type"
12. For Student, Casual, Term, Indeterminate-Acting, Indeterminate-Assignment, Indeterminate-Secondment, we ask for "Expected end date" if "Are you currently in this role?" is selected. After consultation with @petertgiles it was decided to store this data in the same field, just swap the label.
13. For Contractor we ask "Seniority of the role" (but no "Self-employed" option)
14. Contractor we ask "Contractor type" and if "firm or agency" is selected, we ask "Contracting firm or agency"
15. **Read only state** - The "Department, agency or other government organization" is shown where Organization normally shows up
16. **Read only state** - Add "Government of Canada" between Work experience and Dates
[...]

PR #11947, opened 2024-11-14
[Feature] New work experience fields

### LINKED ISSUE #11868: ✨ Backend - new fields for work experience
## ✨ Feature

We will need new fields added to the work experience model to support new data collection in the UI.

## 🕵️ Details

Refer to https://github.com/GCTC-NTGC/gc-digital-talent/issues/11864 for the UI that this issue will support.

Our work experiences are currently stored in a json object in the `experiences` table shared with other experience types.  This issue probably won't include adding actual columns to any tables.  Or database constraints for foreign references.

Note that organization used to be mandatory but will become conditionally mandatory based on employment category.

## 🎨 Design

@Jerryescandon 

https://www.figma.com/design/piRkRtWAHheoEX5Jq8d9u7/Create%2FEdit-an-experience-(Applicants)?node-id=472-70951&t=NbjSJSxQjUSo26FF-1

employment_category - enum
ext_size_of_organization - enum
ext_role_seniority - enum
department - relationship to Department model
gov_employment_type - enum
gov_position_type - enum
gov_contract_end_date - date
gov_contractor_role_seniority - enum
gov_contractor_type - enum
classification - relationship - Classification model
caf_employment_type - enum
caf_force - enum
caf_rank - enum

## ✅ Acceptance Criteria

- [ ] New fields input on new work experience creation
- [ ] New fields input on update existing work experience creation
- [ ] Validator to enforce consistency (eg, no CAF fields when external experience)
- [ ] Can continue to submit work experiences with current UI in place
- [ ] Unit tests for validation of different kinds of experiences
- [ ] New fields returned when querying experiences, same data protection as other experience fields
- [ ] New fields saved in snapshots on pool candidate submission
- [ ] New fields returned when querying snapshot
- [ ] French and English in localized enums
- [ ] Seeder and factory updated
- [ ] Old and new experiences continue to work in edit forms, viewing, downloads, snapshot viewing 

## 🛑 Blockers

```[tasklist]
### Blocked By
```

vd1992:
🤖 Resolves #11868

### DIFF
--- a/api/app/Enums/ExternalSizeOfOrganization.php
+++ b/api/app/Enums/ExternalSizeOfOrganization.php
@@ -4,7 +4,7 @@ namespace App\Enums;
 
 use App\Traits\HasLocalization;
 
-enum SizeOfOrganization
+enum ExternalSizeOfOrganization
 {
     use HasLocalization;
 
--- a/api/app/Enums/GovEmploymentType.php
+++ b/api/app/Enums/GovEmploymentType.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace App\Enums;
+
+use App\Traits\HasLocalization;
+
+enum GovEmploymentType
+{
+    use HasLocalization;
+
+    case STUDENT;
+    case CASUAL;
+    case TERM;
+    case INDETERMINATE;
+    case CONTRACTOR;
+
+    public static function getLangFilename(): string
+    {
+        return 'gov_employment_type';
+    }
+}
--- a/api/app/Models/Classification.php
+++ b/api/app/Models/Classification.php
@@ -8,6 +8,7 @@ use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\HasMany;
 use Illuminate\Database\Eloquent\SoftDeletes;
+use Staudenmeir\EloquentJsonRelations\HasJsonRelationships;
 
 /**
  * Class Classification
@@ -25,6 +26,7 @@ use Illuminate\Database\Eloquent\SoftDeletes;
 class Classification extends Model
 {
     use HasFactory;
+    use HasJsonRelationships;
     use SoftDeletes;
 
     protected $keyType = 'string';
--- a/api/app/Models/WorkExperience.php
+++ b/api/app/Models/WorkExperience.php
@@ -62,8 +62,8 @@ class WorkExperience extends Experience
         'start_date' => 'startDate',
         'end_date' => 'endDate',
         'employment_category' => 'employmentCategory',
-        'ext_size_of_organization' => 'externalSizeOfOrganization',
-        'ext_role_seniority' => 'externalRoleSeniority',
+        'ext_size_of_organization' => 'extSizeOfOrganization',
+        'ext_role_seniority' => 'extRoleSeniority',
         'gov_employment_type' => 'govEmploymentType',
         'gov_position_type' => 'govPositionType',
         'gov_contract_start_date' => 'govContractStartDate',
@@ -139,7 +139,7 @@ class WorkExperience extends Experience
     /**
      * Interact with the external organization size
      */
-    protected function externalSizeOfOrganization(): Attribute
+    protected function extSizeOfOrganization(): Attribute
     {
         return $this->makeJsonPropertyStringAttribute('ext_size_of_organization');
     }
@@ -147,7 +147,7 @@ class WorkExperience extends Experience
     /**
      * Interact with the external role seniority
      */
-    protected function externalRoleSeniority(): Attribute
+    protected function extRoleSeniority(): Attribute
     {
         return $this->makeJsonPropertyStringAttribute('ext_role_seniority');
     }
--- a/api/app/Models/WorkExperience.php
+++ b/api/app/Models/WorkExperience.php
@@ -7,6 +7,7 @@ use Illuminate\Database\Eloquent\Casts\Attribute;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\SoftDeletes;
 use Illuminate\Support\Facades\Lang;
+use Staudenmeir\EloquentJsonRelations\HasJsonRelationships;
 
 /**
  * Class WorkExperience
@@ -33,10 +34,13 @@ use Illuminate\Support\Facades\Lang;
  * @property string $caf_employment_type
  * @property string $caf_force
  * @property string $caf_rank
+ * @property string $classification_attached
+ * @property string $department_attached
  */
 class WorkExperience extends Experience
 {
     use HasFactory;
+    use HasJsonRelationships;
     use SoftDeletes;
 
     /**
@@ -45,6 +49,10 @@ class WorkExperience extends Experience
      * @var string
      */
     protected $table = 'experiences';
+
+    protected $casts = [
+        'properties' => 'json',
+    ];
 
     /**
      * Default values for attributes
@@ -73,6 +81,8 @@ class WorkExperience extends Experience
         'caf_employment_type' => 'cafEmploymentType',
         'caf_force' => 'cafForce',
         'caf_rank' => 'cafRank',
+        'classification_attached' => 'classificationAttached',
+        'department_attached' => 'departmentAttached',
     ];
 
     public function getTitle(?string $lang = 'en'): string
@@ -222,5 +232,37 @@ class WorkExperience extends Experience
     protected function cafRank(): Attribute
     {
         return $this->makeJsonPropertyStringAttribute('caf_rank');
+    }
+
+    /**
+     * Interact with the saved classification id
+     */
+    protected function classificationAttached(): Attribute
+    {
+        return $this->makeJsonPropertyStringAttribute('classification_attached');
+    }
+
+    /**
+     * Interact with the saved department id
+     */
+    protected function departmentAttached(): Attribute
+    {
+        return $this->makeJsonPropertyStringAttribute('department_attached');
+    }
+
+    /**
+     * Return the classification model from JSON
+     */
+    public function classification()
+    {
+        return $this->belongsTo(Classification::class, 'properties->classification_attached');
+    }
+
+    /**
+     * Return the department model from JSON
+     */
+    public function department()
+    {
+        return $this->belongsTo(Department::class, 'properties->department_attached');
     }
 }
--- a/api/app/Models/WorkExperience.php
+++ b/api/app/Models/WorkExperience.php
@@ -49,6 +49,18 @@ class WorkExperience extends Experience
         'division' => 'division',
         'start_date' => 'startDate',
         'end_date' => 'endDate',
+        'employment_category' => 'employmentCategory',
+        'ext_size_of_organization' => 'extSizeOfOrganization',
+        'ext_role_seniority' => 'extRoleSeniority',
+        'gov_employment_type' => 'govEmploymentType',
+        'gov_position_type' => 'govPositionType',
+        'gov_contract_start_date' => 'govContractStartDate',
+        'gov_contract_end_date' => 'govContractEndDate',
+        'gov_contractor_role_seniority' => 'govContractorRoleSeniority',
+        'gov_contractor_type' => 'govContractorType',
+        'caf_employment_type' => 'cafEmploymentType',
+        'caf_force' => 'cafForce',
+        'caf_rank' => 'cafRank',
     ];
 
     public function getTitle(?string $lang = 'en'): string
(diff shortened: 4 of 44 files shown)


api/app/GraphQL/Validators/Mutation/UpdateWorkExperienceValidator.php
```diff
@@ -0,0 +1,217 @@
...
+    {
+        $workExperienceModel = WorkExperience::findOrFail($this->arg('id'));
+
+        // new arguments take priority, otherwise fallback to model value
+        $employmentCategory = $this->arg('workExperience.employmentCategory') ?? $workExperienceModel->employment_category;
+        $extSizeOfOrganization = $this->arg('workExperience.extSizeOfOrganization') ?? $workExperienceModel->ext_size_of_organization;
+        $extRoleSeniority = $this->arg('workExperience.extRoleSeniority') ?? $workExperienceModel->ext_role_seniority;
+        $govEmploymentType = $this->arg('workExperience.govEmploymentType') ?? $workExperienceModel->gov_employment_type;
+        $govPositionType = $this->arg('workExperience.govPositionType') ?? $workExperienceModel->gov_position_type;
+        $govContractStartDate = $this->arg('workExperience.govContractStartDate') ?? $workExperienceModel->gov_contract_start_date;
```
vd1992, line 28: Will delete some of these unused ones after some reviewing is done

petertgiles: I dunno, this doesn't smell quite right to me.  Laravel validation (and Lighthouse) is all about validation input but really you want to validate the final state of the model.  This seems very clunky though.  I think I'll ask on Slack if anyone has ideas.

api/app/Models/Classification.php
```diff
@@ -25,6 +26,7 @@
 class Classification extends Model
 {
     use HasFactory;
+    use HasJsonRelationships;
```
petertgiles, line 29: Is this necessary?  This model doesn't have any JSON relationships does it?

api/app/ValueObjects/ProfileSnapshot.php
```diff
@@ -142,7 +172,9 @@ private function parseSnapshotEnum(mixed $value, $enum): mixed
     private function parseSnapshotRecursive(RecursiveArrayIterator $rai, array $accumulator, $enumMap)
     {
         foreach ($rai as $k => $v) {
-            if (array_key_exists($k, $enumMap)) {
+            if (($k === 'classification' || $k === 'department') && isset($v)) {
```
petertgiles, line 175: I don't love listing individual properties here - it's pretty brittle.  Do we have a programmatic way of identifying these?

api/database/factories/WorkExperienceFactory.php
```diff
@@ -26,15 +38,83 @@ class WorkExperienceFactory extends Factory
     public function definition()
     {
         $startDate = $this->faker->date();
+        $employmentCategory = $this->faker->randomElement(EmploymentCategory::cases())->name;
```
petertgiles, line 41: It might be possible to tidy some of this stuff out of the front of the `definition` function by using closures.  For example, if you check api/database/factories/DigitalContractingQuestionnaireFactory.php you can see how `authorities_involved` gets set to an enum and then `authority_involved_other` gets conditionally set in a closure based on that value without needing an _if_ structure before the attributes array.

api/graphql/schema.graphql
```diff
@@ -1571,6 +1593,24 @@ input WorkExperienceInput {
...
+  govContractStartDate: Date @rename(attribute: "gov_contract_start_date")
+  govContractEndDate: Date @rename(attribute: "gov_contract_end_date")
+  govContractorRoleSeniority: GovContractorRoleSeniority
+    @rename(attribute: "gov_contractor_role_seniority")
+  govContractorType: GovContractorType @rename(attribute: "gov_contractor_type")
+  cafEmploymentType: CafEmploymentType @rename(attribute: "caf_employment_type")
+  cafForce: CafForce @rename(attribute: "caf_force")
+  cafRank: CafRank @rename(attribute: "caf_rank")
+  classification: UUID @rename(attribute: "classification_id")
+  department: UUID @rename(attribute: "department_id")
```
petertgiles, line 1613: Maybe these two should be something like DepartmentBelongsTo?

api/app/Enums/ExternalSizeOfOrganization.php
petertgiles, line 1: These are probably going to need a custom sort since alphabetical won't make sense.  Do we have any examples or storing a sort order for a server-side API?  There might be other enums in this PR that need it, too.

petertgiles (commented):
Wow, a big chonky one!  I just did a code read-through before I start trying things out.

api/app/Enums/GovEmploymentType.php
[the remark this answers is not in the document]
vd1992: I'm hesistant in reusing enums as their functions could diverge, as well as how we also serve enum labels from the server too so the values and label could vary. I renamed it to make it clearer where it slots in https://github.com/GCTC-NTGC/gc-digital-talent/pull/11947/commits/a8b5106b68ef771f029d177857232764f696ed3a

api/app/Models/Classification.php
vd1992: Yeah, per the instructions on the can https://github.com/staudenmeir/eloquent-json-relations?tab=readme-ov-file#one-to-many-relationships

> Remember to use the `HasJsonRelationships` trait in both the parent and the related model.

api/database/factories/WorkExperienceFactory.php
vd1992: Oh, that's a pretty neat method to do this. Thanks. I think I got it all going in https://github.com/GCTC-NTGC/gc-digital-talent/pull/11947/commits/012e8e2d580e4ad6ab3d0ae290726b65b0268dc9

api/app/ValueObjects/ProfileSnapshot.php
vd1992: It was tricky for me to figure out how to get it to work at all 😆 
I've no clue for best practices

api/graphql/schema.graphql
vd1992: Hm, I think I should have "id" in the name of that, changed in https://github.com/GCTC-NTGC/gc-digital-talent/pull/11947/commits/371b58a401f249f4cac10d64b3ee9b20b29484a9
Not sure about sticking "BelongsTo" in there

api/app/Enums/ExternalSizeOfOrganization.php
vd1992: Do you mean database sorting or frontend rendering in dropdowns? If it the latter, I don't think we need to worry about that server-side, seems more like a frontend thing

api/app/Enums/GovEmploymentType.php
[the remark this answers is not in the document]
vd1992: Made a mistake in the above https://github.com/GCTC-NTGC/gc-digital-talent/pull/11947/commits/dba0687c6dedd8707146f56a8511e438a228fbf9

api/app/GraphQL/Validators/Mutation/UpdateWorkExperienceValidator.php
vd1992: So uh, what do we want?

api/app/Enums/ExternalSizeOfOrganization.php
petertgiles: Yeah, I was thinking of rendering dropdowns and option lists.  I looked through the enums and the only one that had anything like that was PriorityWeight.  I think you're right - leave it as a frontend concern for now.  I know we have a way of controlling that in TS.

api/database/factories/WorkExperienceFactory.php
petertgiles: Oh, so much nicer!  :chefs-kiss:

api/app/Enums/GovEmploymentType.php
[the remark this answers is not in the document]
petertgiles: A bit gnarly, but much clearer.  Let's go with it for now.

api/app/ValueObjects/ProfileSnapshot.php
petertgiles: OK, I took a closer look at this and I'm wondering if this logic needs to exist at all.  Why can't the recursive parser just treat these like other related models?  In the snapshot they don't look any different.  I wonder if perhaps you're working around a bug in the storing?
image
Shouldn't these look like?
```json
"department": {
     "id": "3241d6c5-2cdc-4c94-9ce4-28b610100098"
},
"classification": {
     "id": "1efbd3ea-ad55-4c7b-bfea-2feca5df1aca"
}
```

petertgiles: Kind of related to the question about DepartmentBelongsTo

api/app/GraphQL/Validators/Mutation/UpdateWorkExperienceValidator.php
petertgiles: Decision in scrum: don't query existing model, just expect that all the fields to check exist in the input.

vd1992: Aye aye, reusing the validator then in https://github.com/GCTC-NTGC/gc-digital-talent/pull/11947/commits/688a7a92e1aeb127ca32f955c276b728fce723ed

api/app/ValueObjects/ProfileSnapshot.php
vd1992: Okay, I think I got it! Seems like I needed to use the Resource files instead 
https://github.com/GCTC-NTGC/gc-digital-talent/pull/11947/commits/6f83723b5d8b7be9007a66ac1b368f210862c1a5

api/app/GraphQL/Validators/Mutation/CreateWorkExperienceValidator.php
```diff
@@ -0,0 +1,198 @@
...
+     * Return the validation rules.
+     *
+     * @return array<string, array<mixed>>
+     */
+    public function rules(): array
+    {
+        return [
+            // 'workExperience.employmentCategory' => [
+            //     'required',
+            // ],
```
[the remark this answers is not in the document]
vd1992: A/C in https://github.com/GCTC-NTGC/gc-digital-talent/issues/11864 
Makes mention of handling this

api/graphql/schema.graphql
vd1992: The reason I am hesitant about trying to keep it consistent is because it isn't the same at all. We're storing values in JSON and I had to install something to mimic normal Eloquent relations when querying the experience 😕

tristan-orourke: Given that you did install something to mimic normal eloquent relations... would DepartmentBelongsTo actually work?

api/app/ValueObjects/ProfileSnapshot.php
petertgiles: Nice, this is looks way better!

api/graphql/schema.graphql
vd1992: It doesn't seem like it'll work without issue/refactor

Schema change

image

Test input is then

image

Resulting in array to string error

image

Adding this logging

image

Reveals it is passing in the belongs to object and not handling it behind the scenes

image

vd1992: Hang on, hang on.
I got and tried an idea

Using the existing, handy `@pluck` in https://github.com/GCTC-NTGC/gc-digital-talent/pull/11947/commits/53cd3238adc16b2963b32a5c67843971f9d3a2da

petertgiles: Hmm, this doesn't really work the same way.  For example, you can't use disconnect anymore.  I might play with this a bit today to see if I can figure it out.

petertgiles: Bah, I can't get this to work either.  Oh well, I think you've taken this as far as we can.  I'm closing this.

api/app/Models/WorkExperience.php
```diff
@@ -21,10 +22,25 @@
...
  * @property \Illuminate\Support\Carbon $updated_at
+ * @property string $employment_category
+ * @property string $ext_size_of_organization
+ * @property string $ext_role_seniority
+ * @property string $gov_employment_type
+ * @property string $gov_position_type
+ * @property string $gov_contractor_role_seniority
+ * @property string $gov_contractor_type
+ * @property ?\Illuminate\Support\Carbon $gov_contract_start_date
+ * @property ?\Illuminate\Support\Carbon $gov_contract_end_date
```
petertgiles, line 33: Sorry, this is my bad.  We don't need extra date fields for this.  We'll store the dates in the regular start and end dates of the experience.

petertgiles (changes requested):
Almost done!  I think this is the last thing.

api/app/Models/WorkExperience.php
petertgiles: Nice.

petertgiles (approved):
:shipit:

PR #11995, opened 2024-11-15
[Dep] Bump `react-intl` to `6.8`

### LINKED ISSUE #11796: ♻️ Update react-intl from 6.7.0 to 6.8.0 or latest
## ♻️ Debt/Refactor

6.8.0 causes problems with linting

## 🕵️ Details

Relevant release notes here https://github.com/formatjs/formatjs/releases
Possibly due to the React 19 prep?

image

Problem

image

## ✅ Acceptance Criteria

- [ ] Dependency updated to 6.8.0 or later without issue

esizer:
🤖 Resolves #11796 

## 👋 Introduction

Bumps `react-intl` to the latest version.

## 🕵️ Details

Playwright `tsconfig` was checking lib's types and since `Intl` is a browser API it was overlapping and getting confused. Added `skipLibCheck` for playwright only so we avoid this. Also added excludes which should speed up the check...maybe :woman_shrugging: 

## 🧪 Testing

1. Confirm playwright still passes
2. Confirm `react-intl` on latest version

### DIFF
--- a/apps/playwright/tsconfig.json
+++ b/apps/playwright/tsconfig.json
@@ -7,8 +7,17 @@
     "moduleResolution": "Bundler",
     "esModuleInterop": true,
     "strictNullChecks": true,
+    "skipLibCheck": true,
     "paths": {
-      "~/*": ["./*"]
+      "~/*": [
+        "./*"
+      ]
     }
-  }
+  },
+  "exclude": [
+    "node_modules",
+    "!node_modules/@types",
+    "*/node_modules",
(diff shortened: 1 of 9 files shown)


esizer:
> Not sure if this is related, but some Playwright tests are timing out.

While not the exact test, it does seem that timeouts are rare but do happen in playwright :thinking: 

 - https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/11894490491/job/33141830399#step:12:126

PR #12006, opened 2024-11-18
[Fix] Save deleted work email in the right field

### LINKED ISSUE #11962: 🐛 User delete - work email overwriting contact email
## 🐛 Bug

When a user is deleted we modify the contact email to avoid unique constraint issues.  We tried to do that for work email as well but there's a bug causing it to overwrite the contact email and leave the work email unmodified.

## 🕵️ Details

https://github.com/GCTC-NTGC/gc-digital-talent/blob/3277311a77ef8a1a4931280f3c878c8b562699b2/api/app/Models/User.php#L476

## 📋 Steps to Reproduce

1. Populate both the contact email and work email of a user
2. Use the "delete user" button on the admin edit user page
3. Observe that the contact email contains the modified work email and the work email contains the unmodified work email

## 📸 Screenshot

image

## ✅ Acceptance Criteria

- [ ] When deleting a user, the modified contact email goes to the contact email field and the modified work email goes to the work email field
- [ ] Delete still works fine if work email is null

petertgiles:
🤖 Resolves #11962 

## 👋 Introduction

Saves the deleted work email in the right field.

## 🕵️ Details

I refactored the code a little bit to make one database update instead of two.

## 🧪 Testing

1. Set up a user with a contact email address and a different work email address.
2. Delete the user.
3. Confirm that the email addresses are modified correctly.
4. Repeat for a user with a null work email address.

### DIFF
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -469,12 +469,11 @@ class User extends Model implements Authenticatable, HasLocalePreference, Laratr
                 }
 
                 // Modify the email(s) to allow use by another user
-                $newContactEmail = $user->email.'-deleted-at-'.Carbon::now()->format('Y-m-d');
-                $user->update(['email' => $newContactEmail]);
+                $user->email = $user->email.'-deleted-at-'.Carbon::now()->format('Y-m-d');
                 if (! is_null($user->work_email)) {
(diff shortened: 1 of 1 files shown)


esizer (approved):
LGTM! :+1:

PR #12031, opened 2024-11-20
[Fix] Signout action in nav menu

### LINKED ISSUE #11981: ⌨ Signout button incorrect focus styles
## ⌨ Accessibility Issue

The signout button has no specific focus styles (blue background, black text) making it differ from all other interactive elements on the site.

Also you can't get to it with keyboard arrow navigation. Need to tab to it.

## 🦋 Expected Behaviour

Signout button styling is consistent with other interactive elements.

## 📋 Steps to Reproduce

1. Login as any user role
2. Using only keyboard, tab to the signout button in the account menu dropdown
3. Observe the default focus styles

## 🙋‍♀️ Proposed Solution

Use the same styles we use for other text links/buttons.

## ✅ Acceptance Criteria

- [ ] Signout button styles are consistent with other interactive elements
- [ ] Keyboard arrow key navigation within the sub-menu can get you to the sign out button

esizer:
🤖 Resolves #11981 

## 👋 Introduction

Updates the signout action to be a link to `/logged-out` to preserve keyboard shortcuts and styling.

## 🕵️ Details

This can act as an interim fix since it is better than what we currently have. The main issue right now is that the confirmation is still a dialog which comes with the expectation of focus management. Since we are visting a new page we lose focus entirely and it will shift back to the heading of the previous page when cancelling.

A solution could be, instead of using a dialog, we add the confirmation directly on the page like how GitHub does it currently.

## 🧪 Testing

1. Build `pnpm run dev:fresh`
2. Sign in as any existing user
3. Tab to the "Your account" dropdown
4. Confirm that arrow keys allow you to navigate all items in the dropdown (including sign out)
5. Confirm sign out action has the proper styles (specifically hover and focus)
6. Confirm activating the sign out link navigates you to `/logged-out` with confirmation
7. Confirm the cancel action returns you to your previous page
 -->

### DIFF
--- a/apps/web/src/components/NavMenu/useMainNavLinks.tsx
+++ b/apps/web/src/components/NavMenu/useMainNavLinks.tsx
@@ -301,6 +301,7 @@ const useMainNavLinks = () => {
       href={paths.loggedOut()}
       title={intl.formatMessage(authMessages.signOut)}
       state={{ from: pathname }}
+      subMenu
     />
   );
 
--- a/apps/web/src/components/NavMenu/useMainNavLinks.tsx
+++ b/apps/web/src/components/NavMenu/useMainNavLinks.tsx
@@ -1,8 +1,9 @@
 import { useIntl } from "react-intl";
 import uniqBy from "lodash/unionBy";
 import HomeIcon from "@heroicons/react/24/solid/HomeIcon";
+import { useLocation } from "react-router-dom";
 
-import { getNavLinkStyling, NavMenu } from "@gc-digital-talent/ui";
+import { LinkProps, NavMenu } from "@gc-digital-talent/ui";
 import { commonMessages, navigationMessages } from "@gc-digital-talent/i18n";
 import {
   hasRole,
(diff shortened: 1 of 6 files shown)


apps/web/src/components/NavMenu/useMainNavLinks.tsx
vd1992, line 1: It disappears in light mode

image

vd1992, line 1: Opening signout dialog logs a console warning

image

vd1992 (changes requested):
Found some things for your attention below

apps/web/src/components/NavMenu/useMainNavLinks.tsx
esizer: Fixed in [fix sign out link colour](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12031/commits/24a15b441ed629235b3c2fc12774672f714cbbdc)

esizer: Good catch! Added a description in [add description to signout confirmation](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12031/commits/51a767a9c6e400bf725aa280d067b69ec0c4fdd2)

vd1992, line 1: Got a little bit of a hiccup, look at the background. You aren't actually logged out, hitting cancel takes you back, but it is quite confusing

image

vd1992 (changes requested):
Might've been present for my last review, sorry I missed this

vd1992 (approved):
Looking good, nice 👍

apps/web/src/components/NavMenu/useMainNavLinks.tsx
esizer: Yeah, not my favourite solution but it works for now :sweat_smile:

PR #12060, opened 2024-11-21
[Debt] Removes the `experience.user` relationship

### LINKED ISSUE #11230: ♻️ Remove `Experience.user` relationship in schema
## ♻️ Debt/Refactor

We would like to remove the user relationship from experiences in graphql.

## 🕵️ Details

Currently I don't think we have any real use case for this relationship and we are trying to avoid adding deeply nested relationships unless we need them. I believe we should have a max depth/complexity set but theory we could do the following which is a little silly.

```graphql
query {
  users {
    experiences {
      user {
        experiences
      }
    }
  }
}
```

## 🙋‍♀️ Proposed Solution

- Remove the `user` field from all experience types in the schema
- Update any code that relied on this relationship (if we weren't being silly, this should be easy)

## ✅ Acceptance Criteria

- [ ] `user` field removed all experience types in the schema
- [ ] Frontend code updated to no longer use that relationship

esizer:
🤖 Resolves #11230

## 👋 Introduction

Removes the `experience.user` relationship in our graphql API and updates some queries that were using it.

## 🕵️ Details

The queries using it seemed to be legacy documents that qeuried it only to satisfy the codegen types so removing them should now impact functionality.

## 🧪 Testing

2. Confirm no errors during codegen step
3. Navigate to the pages affected
4. Confirm no errors

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -653,7 +653,6 @@ type SkillKeywords {
 
 interface Experience {
   id: ID!
-  user: User! @belongsTo(relation: "user") @canResolved(ability: "view")
   details: String
   skills: [Skill!] @hasMany
   # ExperienceSkillRecord is the pivot data and may only be queried when Experience is nested in a Skill (eg. skill->experiences->experienceSkillRecord). Querying for it otherwise will just return null.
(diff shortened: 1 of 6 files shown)


vd1992 (approved):
Works and the page's functionality appears unchanged

PR #12074, opened 2024-11-22
[Debt] Differentiate error boundaries

### LINKED ISSUE #11998: ♻️ Differentiate `ErrorBoundary` components
## ♻️ Debt/Refactor

We have two components named `ErrorBoundary` but they have subtle differences so we should differentiate them to avoid confusion.

## 🕵️ Details

One component (the function component) is meant for use with `react-router` and while it works in a very similar way to a proper error boundary, it is not one. This one is used exclusively with `react-router` to catch all page based errors.

The other, is a proper react `ErrorBoundary` that we use to catch non-page based errors.

## 🙋‍♀️ Proposed Solution

We should rename one of these but I'm really not sure which one. `react-router` expects to see a named export of `ErrorBoundary` so it might be odd to rename that one. However, that is the pseudo error boundary and the other one is a true `ErrorBoundary` but can be named anything so maybe we do that one? :thinking: 

Rename the react-router one to `RouterErrorBoundary`.

You may need to alias the imports in the `Router.tsx`.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/50a33f83410c2f094015229d5f964cd1f3088a9c/apps/web/src/components/Router.tsx#L19-L24

## ✅ Acceptance Criteria

- [ ] One of the `ErrorBoundary` components has been renamed

esizer:
🤖 Resolves #11998 

## 👋 Introduction

Renames the router based error boundary to differentiate it from a component based error boundary

## 🧪 Testing

2. Force a router error (404 is an easy one)
3. Confirm you still see the expected error boundary

### DIFF
--- a/apps/web/src/components/Layout/IAPLayout.tsx
+++ b/apps/web/src/components/Layout/IAPLayout.tsx
@@ -25,8 +25,8 @@ import micMessages from "~/lang/micCompiled.json";
 
 import SkipLink from "./SkipLink";
 import SitewideBanner from "./SitewideBanner";
-
-export { ErrorBoundary } from "./ErrorBoundary/ErrorBoundary";
+import ErrorBoundary from "./RouteErrorBoundary/RouteErrorBoundary";
+export { ErrorBoundary };
 
 const messages = new Map<string, Messages>([
   ["crg", crgMessages],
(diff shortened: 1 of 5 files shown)


vd1992 (approved):
👍 

image

PR #12079, opened 2024-11-22
[Tests] Add tests for adding each type of experience, linking a skill to an experience

### LINKED ISSUE #11809: 🧪 Add more tests for Experience Form
## 🧪 Tests

Add playwright tests for:

- adding/editing of each type of experience
- linking skills for each type of the experience
- linking the same experience to different skills (https://github.com/GCTC-NTGC/gc-digital-talent/issues/8727)
- 💡 think of any that could be added for experience form

## ✅ Acceptance Criteria

- [ ] Add tests for the above

mnigh:
🤖 Resolves #11809.

## 👋 Introduction

This PR adds tests for adding each type of experience in the create experience form; linking a skill to an experience in the create experience form; and linking the same experience to different skills.

## 🧪 Testing

1. `pnpm e2e:playwright:chromium`
2. Verify all tests pass


apps/playwright/tests/applicant/application.spec.ts
```diff
@@ -73,12 +86,163 @@ test.describe("Application", () => {
...
+      })
+      .click();
+    await application.connectExperience("QA Testing at Playwright University");
+
+    await expect(
+      application.page.getByText(
+        /please connect at least one career timeline experience to each required technical skill and ensure each skill has details about how you used it/i,
+      ),
+    ).toBeHidden();
+    await application.saveAndContinue();
```
yonikid15, line 218: This is the :moneybag: maker right here, and it looks good to me :+1: nice work. 
I was checking it out manually, I wonder if we should check if adding an experience to optional technical skills, won't remove it from other skills?

mnigh: @yonikid15 What would be the purpose of that check? The original [bug](https://github.com/GCTC-NTGC/gc-digital-talent/issues/8727) and its [fix](https://github.com/GCTC-NTGC/gc-digital-talent/pull/8726) were not related to whether the skill was optional or not. If you think this would an important test (I don't, but 🤷), perhaps a new issue would make more sense since it would be out of the scope of this work?

yonikid15: Oh if the bug was only related to the technical skills then this looks good!

yonikid15 (approved):
:star:

PR #12096, opened 2024-11-26
[Fix] Localized enums being available for form init

### LINKED ISSUE #12068: 🐛 Select inputs which fetch enum strings from API may not be initialized properly
## 🐛 Bug

When we render a form before having the enum data from the API, a select input will default to having the incorrect value selected.

#12033 has a well described example of this. However, this likely exists with other forms, and I want to solve them all at once.

## 🦋 Expected Behaviour

Select inputs should be initialized with the previously saved value, or with the nullSelection passed in to the component.

## 🕵️ Details

See this slack discussion: https://talent-cloud.slack.com/archives/C0259G6KXJ6/p1732137983034189

## 📋 Steps to Reproduce

For one example:
1. Log in as an applicant
2. Go to http://localhost:8000/en/applicant/personal-information
3. Edit Work Preferences, fill in form, use a current `Province, territory or region` other than Alberta, and save.
4. View Work Preferences. Your province selection should appear.
5. Click Edit work preferences to open the inline form.
6. The province dropdown now contains the value Alberta.

## 🙋‍♀️ Proposed Solution

For any form fields which require enum strings from the API, load query the data in a component one level up from the form, and use a Pending component to delay rendering the form component until all the data is ready.

Example: https://github.com/GCTC-NTGC/gc-digital-talent/blob/a691342f784ff92ede608f6bec8070877122d2e4/apps/web/src/pages/TrainingEvents/UpdateTrainingEventPage.tsx#L54

Besides the Work Preferences form, do a search for uses of the `localizeEnumToOptions` function to find other forms where this change should be done.

Also, redo #11553 to use the solution suggested here.

## ✅ Acceptance Criteria

- [ ] select inputs which rely on localized Enum data, including the Work Preferences province field, properly display previously saved values (or null selection)
- [ ] forms which rely on data from API aren't rendered until all required data is loaded

esizer:
🤖 Resolves #12068 

## 👋 Introduction

Fixes the issue where localized enums used as options in forms were causing issues with form default values.

## 🕵️ Details

Some of the localized enums that come from the API were not available until after the form initialized, causing problems with rendering the default values. This moves the queries for the options up 1-2 levels so we can delay form initialization until the options have been returned from the API.

## 🧪 Testing

2. Find all forms that use the `localizedEnumToOptions` helper
3. Confirm that their default values are initialized properly

### DIFF
--- a/apps/playwright/tests/admin/admin-workflows.spec.ts
+++ b/apps/playwright/tests/admin/admin-workflows.spec.ts
@@ -5,7 +5,9 @@ import { loginBySub } from "~/utils/auth";
 const USERS_PAGINATED_QUERY = "UsersPaginated";
 
 const searchForUser = async (appPage: AppPage, name: string) => {
-  await appPage.page.getByRole("textbox", { name: /search/i }).fill(name);
+  await appPage.page
+    .getByRole("textbox", { name: /search/i })
+    .fill(name, { timeout: 30000 });
 
   await appPage.waitForGraphqlResponse(USERS_PAGINATED_QUERY);
 };
@@ -45,7 +47,6 @@ test.describe("Admin workflows", () => {
 
     await appPage.page.getByRole("link", { name: /edit gul fields/i }).click();
     await appPage.waitForGraphqlResponse("UpdateUserData");
-    await appPage.waitForGraphqlResponse("UpdateUserOptions");
 
     await appPage.page
       .getByRole("textbox", { name: /telephone/i })
--- a/apps/playwright/tests/applicant/profile-workflows.spec.ts
+++ b/apps/playwright/tests/applicant/profile-workflows.spec.ts
@@ -10,9 +10,6 @@ test.describe("User Profile", () => {
       .getByRole("button", { name: /edit personal/i })
       .click();
 
-    await applicantPage.waitForGraphqlResponse(
-      "PersonalInformationFormOptions",
-    );
     await applicantPage.page
       .getByRole("textbox", { name: /telephone/i })
       .fill("123-456-7890");
@@ -29,7 +26,6 @@ test.describe("User Profile", () => {
     await applicantPage.page
       .getByRole("button", { name: /edit work preferences/i })
       .click();
-    await applicantPage.waitForGraphqlResponse("WorkPreferencesOptions");
     await applicantPage.page
       .getByRole("textbox", { name: /please indicate if there is a city/i })
       .fill("Test locations");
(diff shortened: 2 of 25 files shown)


petertgiles:
Could you please check the Chromatic diffs?  It looks like this has negatively affected the Storybook stories.

esizer:
> There are quite a few places that localizedEnumToOptions is used that you didn't update. For example, AssessmentResultsFilterDialog, ExperienceFormFields, JobPlacementDialog, etc. Did you come up with a more fine-grained criteria for what was affected or did you just test them all?

I manually went through all of them and tested them and those ones were okay. For the job placement dialog, I think it is using fragments and met the pattern already.

petertgiles:
Could you please check the assessment dialog?  It looks like the fancy radio option labels load in after the form is displayed.

https://github.com/user-attachments/assets/13d4792b-37f7-4ad7-96f8-ee25ff5711ab

petertgiles (changes requested):
Seems to work great, other than the small assessment dialog options thing.  It initializes properly so I could even be convinced that that is out of scope.  I'll take a look at that Playwright test and see if I can figure anything out.

esizer:
> Seems to work great, other than the small assessment dialog options thing. It initializes properly so I could even be convinced that that is out of scope. I'll take a look at that Playwright test and see if I can figure anything out.

I'll take a look. Hopefully I can fit it in the scope of this ticket. Idk what happened but I pushed up my e2e test changes and it works in CI even though it was flaky locally, if they pass for your locally maybe it is a me problem :sweat_smile:

petertgiles:
The Playwright tests pass for me.  Well, except for Webkit, but those always fail for me.

esizer:
> Seems to work great, other than the small assessment dialog options thing. It initializes properly so I could even be convinced that that is out of scope. I'll take a look at that Playwright test and see if I can figure anything out.

Alright, not sure if it is better but it shows a spinner now so it is a little less jarring. 

[show spinner when loading assessment options](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12096/commits/de635fef97adcd8417cfd1b93002927196b3362d)

> Could you please check the Chromatic diffs? It looks like this has negatively affected the Storybook stories.

Should be working now after [fix storybook enums](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12096/commits/6a8f3702ca25de6cce7d08a47050a45c537a4ee5)

petertgiles:
Nice.  At least it only does it the first time.  Thanks for the fix.

petertgiles (approved):

PR #12097, opened 2024-11-26
[Fix] Switch env constants to import.meta

### LINKED ISSUE #11578: 🐛 Some pages fail to render in Vite watch mode
## 🐛 Bug

Some pages on the site fail to render with a "We have encountered an unknown error while processing your request." error when using Vite watch mode.  However, if the site is built with a regular dev or prod build then they are fine.

## 🦋 Expected Behaviour

Pages in watch mode should behave exactly the same as pages in a regular build.

## 🕵️ Details

It seems related to the process variable. The second you introduce that in watch mode it fails.
While it does not fail on build...the process is not working.
image

## 📋 Steps to Reproduce

> [!Note]
This issue was found in PR #11577.  These steps require that PR to be merged.

1. Build the app in watch mode
2. Start working through an application
3. Add an experience (http://localhost:3000/en/applications/:applicationId/career-timeline/add)

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

image

## ✅ Acceptance Criteria

- [ ] Adding an experience to the career timeline works in a watch mode build
- [ ] Any changes to process.env work in Azure

petertgiles:
🤖 Resolves #11578 

## 👋 Introduction

Switches the talentsearch constants to plain constants to let watch mode work properly on those pages.

## 🕵️ Details

Vite imports environment variables [a little differently](https://vite.dev/guide/env-and-mode) than Webpack.  However, switching to the import.meta imports that Vite uses [breaks Jest](url).  Working around this would require the import of yet another dependency and mocking the variables for testing.  After [some discussion on Slack](https://talent-cloud.slack.com/archives/CB424V61J/p1732656221379149) I decided to switch these to plain constants.

## 🧪 Testing

1. Build in watch mode `pnpm run watch`
2. Log in as `admin@test.com`
3. Visit `http://localhost:3000/en/admin/communities/{communityId}/edit` and confirm it doesn't crash.

Make sure the two constants still work:
1. Visit `http://localhost:3000/en/accessibility-statement` to see the support email
2. Submit a ticket at `http://localhost:3000/en/support` to confirm it requests the correct endpoint (but probably won't properly submit without more setup).

### DIFF
--- a/apps/web/.env.example
+++ b/apps/web/.env.example
@@ -23,7 +23,3 @@ FEATURE_DIRECTIVE_FORMS=true
 # Logging - enable by setting the logging level. Eight severity levels from rfc5424.
 LOG_CONSOLE_LEVEL="debug"
 LOG_APPLICATIONINSIGHTS_LEVEL="warning"
-
-# Support
-VITE_TALENTSEARCH_SUPPORT_EMAIL="support-soutien@talent.canada.ca"
-VITE_API_SUPPORT_ENDPOINT="/api/support/tickets"
--- a/apps/web/src/constants/talentSearchConstants.ts
+++ b/apps/web/src/constants/talentSearchConstants.ts
@@ -1,6 +1,4 @@
-export const TALENTSEARCH_SUPPORT_EMAIL =
-  process.env.TALENTSEARCH_SUPPORT_EMAIL! ?? "support-soutien@talent.canada.ca";
-export const API_SUPPORT_ENDPOINT =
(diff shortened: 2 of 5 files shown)


esizer (commented):
I think your lint errors may be caused by these not being updated in the `vite-env.d.ts` file.

https://vite.dev/guide/env-and-mode#intellisense-for-typescript

petertgiles:
I've tested this on the Azure dev vertical.

esizer (approved):
Nice, love the switch to hard-coded values :+1:

PR #12162, opened 2024-12-02
[Fix] Removes `directive-on-digital-talent` paths from `privilegedPaths` in `protectedEndpointExchange`

### LINKED ISSUE #12070: 🐛 `directive-on-digital-talent` should not use privileged api endpoint
## 🐛 Bug

Any api requests on the Directive on Digital Talent page are currently routed through `/admin/graphql`, meaning they fail if not on the gc VPN. This is unnecessary and can cause some safe api calls to fail.

## 📋 Steps to Reproduce

1. Go to https://talent.canada.ca/en/directive-on-digital-talent
2. Open network tab in browser inspector
3. Refresh page
4. See benign api request fail (query SitewideBanner) 

## 📸 Screenshot

Image

## 🙋‍♀️ Proposed Solution

Remove `/directive-on-digital-talent` routes from `privilegedPaths` in `protectedEndpointExchange`.

## ✅ Acceptance Criteria

- [ ] API request on Directive page don't use privileged endpoint

mnigh:
🤖 Resolves #12070.

## 👋 Introduction

This PR removes `directive-on-digital-talent` paths from `privilegedPaths` in `protectedEndpointExchange`.

## 🧪 Testing

If I understand correctly, this would need to be tested on one of the Azure verticals.

### DIFF
--- a/packages/client/src/exchanges/protectedEndpointExchange.ts
+++ b/packages/client/src/exchanges/protectedEndpointExchange.ts
@@ -3,14 +3,7 @@ import type { Exchange } from "@urql/core";
 
 import { apiHost, protectedUrl } from "../constants";
 
-const privilegedPaths = [
-  "/admin",
-  "/en/admin",
-  "/fr/admin",
-  "/directive-on-digital-talent",
-  "/en/directive-on-digital-talent",
-  "/fr/directive-on-digital-talent",
-];
+const privilegedPaths = ["/admin", "/en/admin", "/fr/admin"];
 
 // A custom exchange that changes to the protected endpoint depending on the current location
 


brindasasi (commented):
Can't exactly go by the steps in the issue for verification as both Dev & UAT can be accessed only under VPN. 

However before deploying this branch SitewideBanner api is hitting the admin/graphql end point as it is from the previleged paths array.
image
After deploying this branch sitewidebanner api is hitting graphql endpoint directly which will work outside the vpn as well like in our other homepages 
image

Hence it should work in prod as well !

brindasasi (approved):
LGTM! 👍

PR #12177, opened 2024-12-03
[FIX] Add missing case for missing FinalDecision weighting

### LINKED ISSUE #12173: 🐛 Prod logs - No match for decision
## 🐛 Bug

We have the message "No match for decision" showing up in the prod logs a bunch.

## 🙋‍♀️ Proposed Solution

https://github.com/GCTC-NTGC/gc-digital-talent/blob/c78f46a7016e31c0c8d97e13c5adc3c2cc813530/api/app/Models/PoolCandidate.php#L1217
Since there's no decision specified after the error message I guess this is missing null handling?  Or the decision is missing when it shouldn't be?

## ✅ Acceptance Criteria

- [ ] Find out how this is happening
- [ ] Close the gap

tristan-orourke:
🤖 Resolves #12173

## 👋 Introduction

From what I can see, the only way to trigger thiswas to change the status of a PoolCandidate _to_ DRAFT or DRAFT_EXPIRED. Or to make an assessment on a DRAFT application. Neither of which should have happened... So further investigation into the prod logs might be worth it. Check the history table. Were any PoolCandidates updated when these logs happened? Why?

## 🕵️ Details

I like that it logs an error for unmatched decisions. This is a good red flag for us if we add a new status or FinalDecision value and forget to put it here. So for now, I allow the DRAFT statuses to skip the call to log an error, without changing the final values.

## 🧪 Testing

1. Log in as an admin
2. Change a submitted PoolCandidate status direction to DRAFT.
3. Check api/storage/logs/laravel.log. You should no longer see "local.ERROR: No match for decision"

### DIFF
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -1176,6 +1176,9 @@ class PoolCandidate extends Model
                 PoolCandidateStatus::REMOVED->name => FinalDecision::REMOVED->name,
                 PoolCandidateStatus::EXPIRED->name => FinalDecision::QUALIFIED_EXPIRED->name,
 
+                PoolCandidateStatus::DRAFT->name,
+                PoolCandidateStatus::DRAFT_EXPIRED->name => 'DRAFT',
+
                 default => null
             };
         }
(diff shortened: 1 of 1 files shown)


tristan-orourke:
> Could we not just short circuit all of this and do something like this at the top of the function?
> 
> ```
> if($status === DRAFT || $status === DRAFT_EXPIRED) {
>   return [ 'decision' => null, 'weight' => null ];
> }
> ```

Okay @esizer, updated to use this shortcut.

esizer (approved):
LGTM! :+1:

PR #12180, opened 2024-12-03
[Fix] Check for `BroadcastChannel` before using it

### LINKED ISSUE #11669: 🐛 Lack of BroadcastChannel causing exceptions
## 🐛 Bug

Some of our clients don't support the BroadcastChannel feature used to sync up tokens between tabs.  This results in exceptions being thrown.

## 🕵️ Details

https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel#browser_compatibility

## 📋 Steps to Reproduce

1. Use a browser that doesn't support the feature
2. Boot the app

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

I don't think it's worth doing a polyfill or anything.  Let's just fail gracefully without throwing an exception.

## ✅ Acceptance Criteria

- [ ] No exception thrown if the browser doesn't support BroadcastChannel

esizer:
🤖 Resolves #11669 

## 👋 Introduction

Updates our use of `BroadcastChannel` to check to make sure it exists in the window (supported by browser) before we use it. 

## 🕵️ Details

This was used to broadcast a logout event to all tabs so a user is not logged in on another tab accidentally.

> [!NOTE]
> I don't have access to any of the browsers that do not support this feature but this should work. Hopefully someone has access to one of the browsers that does not support `BroadcastChannel` to confirm the exception is not thrown. Of course Safari has only supported it for 2 years :woman_facepalming: 

## 🧪 Testing

2. Login as any user account
3. Open another tab that is logged in
4. Logout of one of the tabs
5. Confirm the other tab still logs out as well
6. Repeat steps 2-4 in one of the unsupported browsers (if possible)
7. Confirm no exception is thrown but the other tab stays logged in

### DIFF
--- a/packages/auth/src/hooks/useLogoutChannel.ts
+++ b/packages/auth/src/hooks/useLogoutChannel.ts
@@ -11,8 +11,8 @@ let singleChannel: BroadcastChannel | undefined;
  *
  * @returns
  */
-const getChannel = () => {
-  if (!singleChannel) {
+const getChannel = (): BroadcastChannel | undefined => {
+  if (!singleChannel && "BroadcastChannel" in window) {
     singleChannel = new BroadcastChannel("logoutChannel");
   }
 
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
Looks great!  :sunglasses: I don't have access to one of those older browsers either but at least this didn't break anything.

esizer:
> Looks great! 😎 I don't have access to one of those older browsers either but at least this didn't break anything.

I honestly doubt anyone on the team capable of testing this has access to them, you would need to have an apple device that has not been updated in 3 years :grimacing:

PR #12182, opened 2024-12-03
[Tests] Improve `pool-candidate.spec.ts`

### LINKED ISSUE #10523: 🧪 Improve `pool-candidate.spec.ts`
## 🧪 Tests

Improve `pool-candidate.spec.ts`

Right now it is pretty straightforward since it was based off a legacy page, the current view candidate page has had quite a lot of new functionality added. We've had some issues in the past with things not loading correctly like the snapshot, so would be good to be thorough here. 

## 🙋‍♀️ Proposed Solution

Assert snapshot loads correctly
Assert notes updating works
Assert Screening and assessment table, so can update an assessment to be passed
Test RoD mutation chain, so you load a candidate and test the right mutations can be done on it. So an under assessment candidate can be qualified then reverted then removed and status/headers appear appropriately for example

## ✅ Acceptance Criteria

- [ ] Test has suggestions added

vd1992:
🤖 Resolves #10523

## 👋 Introduction

The view an application page is pretty important with quite a few things going on. This exercises some functions and confirms the presence of the important snapshot

## 🧪 Testing

1. Passes CI
2. Run with `pnpm run e2e:playwright:chromium pool-candidate.spec.ts`


apps/playwright/tests/admin/pool-candidate.spec.ts
esizer, line 1: Looks great! I just have one minor thing to ask, can we prefer to use `getByRole` where possible? It just implictly tests the element has the correct role for ATs while also testing its functionality.

I just noticed a few spots where we use `getByLabel` and `getByPlaceholder` instead to interact with inputs.

vd1992: I was just going off what the Locator would return for selector statements. Revised in https://github.com/GCTC-NTGC/gc-digital-talent/pull/12182/commits/f7019937dbb4b626b07f6930d503201e0b525637
Though, I think I found an issue? See below, shortly

apps/playwright/tests/admin/pool-candidate.spec.ts
```diff
@@ -94,6 +96,208 @@ test.describe("Pool candidates", () => {
...
+      appPage.page.getByText("I meet the applied work"),
+    ).toBeVisible();
+    await expect(
+      appPage.page
+        .getByLabel("Assess the candidate's")
+        .getByText("Test Experience"),
+    ).toBeVisible();
+    await appPage.page.getByText("Demonstrated", { exact: true }).click();
+    await appPage.page
+      .getByText("Work experience equivalency is accepted", { exact: true })
```
vd1992, line 247: I cannot seem to get the radio by name to work here. Not sure why
This isn't a regular radio input, but from the CardItem/CardGroup component

image

apps/playwright/tests/admin/pool-candidate.spec.ts
```diff
@@ -94,6 +96,208 @@ test.describe("Pool candidates", () => {
...
+      .getByRole("button", { name: "Expand all application" })
+      .click();
+
+    // personal information
+    await expect(appPage.page.getByText(/Given namePlaywright/i)).toBeVisible();
+    await expect(
+      appPage.page
+        .getByLabel("Personal and contact")
+        .getByText(/playwright.sub./),
+    ).toBeVisible();
```
esizer, line 129: Can we also ensure this has the proper role? You can do something like:

```tsx
const personal = appPage.page.getByRole("region", { name: /personal and contact/i });
await expect(personal).toContainText(/given nameplaywright/i);
await expect(personal).toContainText(/playwright.sub./);
```

This will also work with the other accordions as well.

apps/playwright/tests/admin/pool-candidate.spec.ts
esizer: oof, that is super unfortunate and kind of raises a red flag. Looks like these tests were a good idea! I don't think this is an actual issue and probably just a downside of playwright but I'll do some extra testing of these components after this to confirm.

apps/playwright/tests/admin/pool-candidate.spec.ts
```diff
@@ -94,6 +96,208 @@ test.describe("Pool candidates", () => {
...
+    await expect(
+      appPage.page
+        .getByLabel("Assess the candidate's")
+        .getByText("Test Experience"),
+    ).toBeVisible();
+    await expect(
+      appPage.page.getByRole("button", {
+        name: "Demonstrated",
+      }),
+    ).toBeVisible();
```
esizer, line 259: These tests also uncovered something, not sure how great it is that all these buttons have the same text :cry: 

I'm going to do some testing but we may want to change these so the name is like "Education assessment, application screening demonstrated"

apps/playwright/tests/admin/pool-candidate.spec.ts
```diff
@@ -94,6 +96,208 @@ test.describe("Pool candidates", () => {
...
+    // skill result
+    await appPage.page
+      .getByRole("row", { name: `${technicalSkill?.name?.en}` })
+      .getByRole("button")
+      .click();
+    await expect(
+      appPage.page.getByText(`Test skill ${technicalSkill?.name?.en}`),
+    ).toBeVisible();
+    await appPage.page.getByText("Not demonstrated (Hold for").click();
+    await appPage.page.getByLabel("Decision notes *").fill("Reason");
```
esizer, line 270: This could probably be:

```tsx
await appPage.page.getByRole("textbox", { name: /decision notes/i }).fill("Reason");
```

esizer (changes requested):
Noticed a few more places we could use `getByRole`. 

I also just left some notes so I don't forget what I noticed while reviewing these tests :stuck_out_tongue:

apps/playwright/tests/admin/pool-candidate.spec.ts
vd1992: Aye aye

apps/playwright/tests/admin/pool-candidate.spec.ts
vd1992: Oh cool, I didn't know about the `region` thing
Done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/12182/commits/42cbc8fb1b83f07840605bcda6286c7adeac8815

apps/playwright/tests/admin/pool-candidate.spec.ts
vd1992: I think the `CardOption`s in a group all share the same `name`

apps/playwright/tests/admin/pool-candidate.spec.ts
esizer: In case you are wondering, I did do the testing and the issue was the table markup. 

https://github.com/GCTC-NTGC/gc-digital-talent/issues/12207

esizer (approved):
LGTM! 👍

PR #12186, opened 2024-12-04
[Fix] Sort nulls last for pool publish date

### LINKED ISSUE #11535: 🐛 When sorting Pools by Publishing Date, null dates should appear at bottom when sorting descending
## 🐛 Bug

Currently, when you sort Pools by Publishing Date descending, unpublished pools appear at the top, THEN recently published pools, THEN pools published a long time ago. This isn't very useful, since when sorting by published, I'm likely to want to know which ones were published most recently. All the unpublished pools get in the way of finding those.

## 🦋 Expected Behaviour

Unpublished pools (NULL date) should appear at the end of the list when sorting by publish date descending.

QUESTION: should NULLs _always_ appear at the end of the list, whether sorting by ascending _or_ descending?

## 📋 Steps to Reproduce

1. Log in as admin
2. go to pools table
3. Sort by publish date
4. Observe that you have to pass all the unpublished pools before getting to recently published pools

## ✅ Acceptance Criteria

- [ ] When sorting pools by publish_date descending, null dates appear at end of the list
- [ ] QUESTION: maybe? When sorting pools by publish_date descending, null dates still appear at end of the list

petertgiles:
🤖 Resolves #11535 

## 👋 Introduction

Sorts publish date with nulls last on the pool table.

## 🕵️ Details

I just added a new sorting scope to the pool paginated query.  It would be nice to generalize this to something we could use throughout the schema but for now it's a stopgap.

The new scope could be used for any physical column in the table but is not advanced enough to handle relationships like Lighthouse's directive can.  It also only accepts a single column at a time.

### Frontend

1. Rebuild and log in as admin@test.com
2. Navigate to /en/admin/pools
3. Play with sorting options
  - [ ] initial sort is still created date
  - [ ] bookmarks is still the overriding sort option
  - [ ] sorting by publishing date, ascending and descending, sorts nulls to the end
  - [ ] other sorting options unaffected

### Backend

1. Navigate to /admin/graphiql and add an `authorization` header with a bearer token for admin@test.com
2. Play with sorting options
```graphql
query Pools($where:PoolFilterInput, $orderByColumn: OrderByColumnInput) {
  poolsPaginated(first: 100, orderByColumn: $orderByColumn, where:$where) {
    data {
      id      
      publishedAt
    }    
  }
}
-----
{
  "where": {
    "statuses": [
      "DRAFT",
      "PUBLISHED"
    ]
  },
  "orderByColumn": {
    "column": "published_at",
    "order": "DESC",
    "nulls": "ORDER_LAST"
  }
}
```
- [ ] scope only accepts real column names
- [ ] scope only accepts real options for "order" and "nulls"
- [ ] scope handles undefined "nulls" option
- [ ] scope sorts as expected

## 📸 Screenshot

image

### DIFF
--- a/api/app/Builders/PoolBuilder.php
+++ b/api/app/Builders/PoolBuilder.php
@@ -211,7 +211,9 @@ class PoolBuilder extends Builder
         $nulls = $args['nulls'] ?? null;
 
         // verify if column name is valid
-        $selectableColumns = $this->model->getSelectableColumns();
+        /** @var \App\Models\Pool */
+        $model = $this->model;
+        $selectableColumns = $model->getSelectableColumns();
         if (! in_array($column, $selectableColumns)) {
             throw new \Exception('Invalid column');
         }
--- a/api/app/Builders/PoolBuilder.php
+++ b/api/app/Builders/PoolBuilder.php
@@ -235,6 +235,7 @@ class PoolBuilder extends Builder
             default => throw new \Exception('Invalid nulls option'),
         };
 
+        // SQL execution from user input!  Ensure sufficient sanitization.
         $this->orderByRaw("$columnSql $orderOptionSql $nullsOptionSql");
 
         return $this;
--- a/api/app/Builders/PoolBuilder.php
+++ b/api/app/Builders/PoolBuilder.php
@@ -203,6 +203,43 @@ class PoolBuilder extends Builder
         return $this;
     }
 
+    // A scope for a simple orderBy on a column.  Allows for nulls first or last.
+    public function orderByColumn(?array $args): self
+    {
+        $column = $args['column'];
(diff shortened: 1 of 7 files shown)


vd1992 (changes requested):
Looks and functions well, aside from linting my only thing to note it we should probably have that bespoke sort covered under some sort of PHP test.

petertgiles:
> we should probably have that bespoke sort covered under some sort of PHP test

Good call!
Added in [test suite](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12186/commits/c39656891ddd26cea95470a9092641bfd8739984).

vd1992 (approved):
Fantastic! Thanks for adding testing

PR #12219, opened 2024-12-05
[Fix] Updates `workEmailDomainRegex`

### LINKED ISSUE #12144: 🐛 Work email validation allows unexpected domains through
## 🐛 Bug

Our work email validation is pretty simple right now - just that it ends with, for example, "gc.ca".  This allows subdomains like "tbs-sct.gc.ca" to be properly approved. However, it also allows domains like dogc.ca to be approved.

## 📋 Steps to Reproduce

1. Enter a work email with a different top-level domain that ends with "gc.ca"

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

Our validation should be smart enough to allow the exact match of top-level domain and subdomains without matching our top-level domains that end with an expected string.  This probably means that the backend validation will needs to be switched to a regex as well.  This is probably better since the frontend can more easily be compared.

## ✅ Acceptance Criteria

- [ ] Exact top-level matches like "example@gc.ca" are approved
- [ ] Matches with a subdomain like "example@tbs-sct.gc.ca" are approved
- [ ] Different top-level domains that end with the same string like "example@dogc.ca" are rejected
- [ ] All of the above apply to frontend validation and backend validation

mnigh:
🤖 Resolves #12144.

## 👋 Introduction

This PR updates the `workEmailDomainRegex` to be more restrictive. It also updates the backend validator `UpdateUserAsUserInputValidator` to use the same regular expression as `workEmailDomainRegex`.

> [!NOTE]
> The regular expression is not perfect but it is better than what came before. 

## 🧪 Testing

> [!TIP]
> For testing regex against strings, https://regex101.com/r/T8IYJU/4.

1. `pnpm build`
2. Navigate to http://localhost:8000/en/applicant/personal-information#government-section
3. Edit Government employee information
4. Enter example@gc.ca in Work email address input
5. Save changes
6. Verify success toast Government information updated successfully!
7. Edit Government employee information
8. Enter example@tbs-sct.gc.ca in Work email address input
9. Save changes
10. Verify success toast Government information updated successfully!
11. Edit Government employee information
12. Enter example@dogc.ca in Work email address input
13. Verify frontend rule states not a valid government email

### DIFF
--- a/api/app/GraphQL/Validators/UpdateUserAsUserInputValidator.php
+++ b/api/app/GraphQL/Validators/UpdateUserAsUserInputValidator.php
@@ -46,7 +46,7 @@ final class UpdateUserAsUserInputValidator extends Validator
                 Rule::unique('users', 'email')->ignore($this->arg('id'), 'id'),
                 Rule::unique('users', 'work_email')->ignore($this->arg('id'), 'id'),
                 // Note: Should be kept in sync with the workEmailDomainRegex
-                'regex:/@([A-Za-z0-9-.]+\.)*(gc\.ca|canada\.ca|elections\.ca|ccc\.ca|canadapost-postescanada\.ca|gg\.ca)$/i',
+                'regex:/@([A-Za-z0-9-]+\.)*(gc\.ca|canada\.ca|elections\.ca|ccc\.ca|canadapost-postescanada\.ca|gg\.ca)$/i',
             ],
             'sub' => [
                 'sometimes',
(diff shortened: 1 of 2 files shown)


petertgiles (changes requested):
I like the unified regex approach!  There is a problem with this pattern, though.  It still allows for unexpected endings for the domains other than gc.ca.   For example, dogg.ca still goes through.

Here's a possible update to the pattern?
`@([A-Za-z0-9-.]+\.)*(gc\.ca|canada\.ca|elections\.ca|ccc\.ca|canadapost-postescanada\.ca|gg\.ca)$`

mnigh:
> @([A-Za-z0-9-.]+\.)*(gc\.ca|canada\.ca|elections\.ca|ccc\.ca|canadapost-postescanada\.ca|gg\.ca)$

Thanks for the suggestion. Updated in e9339690e14ec02bece7aed8e628c53401671bbb.

petertgiles (approved):
Nice, good stuff.

PR #12221, opened 2024-12-06
[Debt] Moves admin routes as children of root route

### LINKED ISSUE #12032: ♻️ Move admin routes as children of root route
## ♻️ Debt/Refactor

There is now no discernible difference between the root route and admin routes so we can likely move away from them being defined separately. 

## 🕵️ Details

We recently moved all the admin pages over to use a nearly identical layout to the root layout but did not move the routes over so we seem to have a redundant `AdminLayout` that simply mirrors the main layout.

This also means our route definition has to add in another top level definition for `${locale}/admin` that applies the same layout.

I think we should be safe to delete the `AdminLayout` now and move all the routes to the root route now.

## 🙋‍♀️ Proposed Solution

 - Reorganize the `createRoute` function to no longer need a separate definition for admin routes
 - Delete the unused `AdminLayout`
 - Possibly remove the admin specific error boundary as well :thinking: 

## ✅ Acceptance Criteria

- [ ] Routes under `/admin` have been moved to the root route definition
- [ ] No visual difference for admin routes
- [ ] Unused code removed

mnigh:
🤖 Resolves #12032.

## 👋 Introduction

This PR moves admin routes as children of root route. It removes the unused `AdminLayout` component and story.

## 🧪 Testing

1. Navigate to admin route
2. Ensure error boundary still renders
3. Navigate to non-admin route
4. Ensure error boundary still renders


petertgiles (approved):
Seems good.  Nice!

PR #12225, opened 2024-12-06
[Accessibility] Role menu aria text added

### LINKED ISSUE #11973: ⌨ Role menu purpose not apparent to screen readers
## ⌨ Accessibility Issue

The purpose of the new role dropdown menu lacks context for screen readers.

## 🦋 Expected Behaviour

The purpose of the menu is clear.

## 🕵️ Details

The new role dropdown menu uses the current role as the name which can be confusing. When using a screen reader, all you hear is something like "Button applicant collapsed, 3 items". For someone using the site who is not already aware of this, they may not understand what the dropdown does.

## 📋 Steps to Reproduce

1. Login with someone who has multiple roles `admin@test.com`
2. Start any screen reader
3. Tab focus until you land on the role dropdown
4. Observe what is read out

## 🙋‍♀️ Proposed Solution

~~We may want to add an `aria-label` and possibly an `aria-description` to notify users what the dropdown does 🤔 ~~

We took a look and unfortunately, `aria-description` is not fully supported yet. The interim solution would to add a label to the input like "Current role" and associate it with the button with an `aria-describedby` but that will require visual changes so we will need an updated design.

## Localized Copy

English: Select a different role to access its corresponding tools and information.
French: Sélectionnez un rôle différent pour accéder aux outils et informations correspondants.

## ✅ Acceptance Criteria

- [ ] Add "Current role" label (not a `label` element thought :sweat_smile: ) before the button (requires design change)
- [ ] Associated the current role specified text to the role menu as `aria-describedby`

vd1992:
🤖 Resolves #11973 

## 👋 Introduction

Added the context providing information.
I found aria-description did nothing so I went with label instead. I was using Microsoft Narrator 

## 🧪 Testing

1. Navigate to any page really
2. Test however you prefer

### DIFF
--- a/apps/web/src/components/NavMenu/MainNavMenu.tsx
+++ b/apps/web/src/components/NavMenu/MainNavMenu.tsx
@@ -157,6 +157,13 @@ const MainNavMenu = () => {
                     color={isSmallScreen ? "black" : "whiteFixed"}
                     mode="text"
                     block={false}
+                    aria-label={intl.formatMessage({
+                      defaultMessage:
+                        "Select a different role to access its corresponding tools and information",
+                      id: "sSbt7W",
+                      description:
+                        "Aria text to describe what the purpose of the role menu is",
+                    })}
                   >
                     {roleNames[navRole]}
                   </NavMenu.Trigger>
--- a/apps/web/src/lang/fr.json
+++ b/apps/web/src/lang/fr.json
@@ -11126,6 +11126,10 @@
     "defaultMessage": "Brève description de l'équipe (en anglais)",
     "description": "Label for team description in English language"
   },
+  "sSbt7W": {
+    "defaultMessage": "Sélectionnez un rôle différent pour accéder aux outils et informations correspondants",
+    "description": "Aria text to describe what the purpose of the role menu is"
+  },
   "sWBbdX": {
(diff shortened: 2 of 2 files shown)


apps/web/src/components/NavMenu/MainNavMenu.tsx
```diff
@@ -157,6 +157,13 @@ const MainNavMenu = () => {
                     color={isSmallScreen ? "black" : "whiteFixed"}
                     mode="text"
                     block={false}
+                    aria-label={intl.formatMessage({
+                      defaultMessage:
+                        "Select a different role to access its corresponding tools and information",
+                      id: "sSbt7W",
+                      description:
+                        "Aria text to describe what the purpose of the role menu is",
+                    })}
```
esizer, line 166: So, it looks like `aria-description` is not ready for production yet :cry: So not using it was the right call. However, with `aria-label`, we no longer know what the current role is. 

We should get both contexts for this and right now, I am not sure how we can do that :thinking:

vd1992: Tricky, tricky then huh

mnigh:
@vd1992 @esizer should this PR be closed?

vd1992:
> @vd1992 @esizer should this PR be closed?

🤷‍♂️

PR #12226, opened 2024-12-09
[Feature] Utility function to detect the VPN

### LINKED ISSUE #12071: ✨ Create a frontend utility function which can determine if browser is on VPN / protected network
## ✨ Feature

Parts of our site are restricted at the firewall level to government networks. We want to be able to pre-emptively detect if we are on those networks, to warn users about access issues or block off certain features which won't work anyway.

idea: Trigger a fetch request to `/admin`, and see if its redirected to `/restricted.html`

## ✅ Acceptance Criteria

- [ ] There exists a utility function which returns true if and only if the user is on a network which would allow them to visit `/admin` or use `/admin/graphql`.

petertgiles:
🤖 Resolves #12071 

## 👋 Introduction

Adds a utility function to detect if the protected endpoint is reachable (and therefore, on the VPN).  As a bonus, I also updated the special error exchange to better detect being off the VPN.

## 🕵️ Details

Unfortunately, we don't have a way to test for the VPN locally, or even on UAT.  I added an Nginx rule (commented out by default) to simulate it.  I hope it's an accurate simulation.  :crossed_fingers: 

## 🧪 Testing

1. Since the AC doesn't specify that this function should be used anywhere, plop this testing code with a button on a page somehwere:
```tsx
async function handleClick() {
  const result = await canAccessProtectedEndpoint();
  console.log(result);
}
```
2. Rebuild and test that your button returns `true`.
3. Enable the firewall simulator and test that the button returns `false`.
   1. Open `infrastructure/conf/nginx-conf-local/default`
   2. Comment out the block labelled "privileged access to the API behind the VPN"
   3. Uncomment the block labelled "simulate being off the VPN"
   4. Restart the webserver container
4. With the firewall simulator still enabled, visit the admin site and observe the "request wasn't completed" error toasts.
## 📸 Screenshot

image

### DIFF
--- a/infrastructure/conf/nginx-conf-local/default
+++ b/infrastructure/conf/nginx-conf-local/default
@@ -139,6 +139,13 @@ server {
         fastcgi_index index.php;
     }
 
+    # Simulate being off the VPN and the firewall redirecting to the restricted page.
+    # Don't forget to also comment out the location rule for /admin/graphql
+    # location ~* /admin(/|$) {
+    #     absolute_redirect on;
+    #     return 301 http://localhost:8000/restricted.html;
+    # }
+
     # api
     location = /graphql {
         access_log off;
@@ -153,12 +160,6 @@ server {
         include fastcgi_params;
         fastcgi_param SCRIPT_FILENAME $document_root/api/public/index.php;
     }
-
-    # simulate being off the VPN and the firewall redirecting to the restricted page
-    # location = /admin {
-    #     absolute_redirect on;
-    #     return 301 http://localhost:8000/restricted.html;
-    # }
 
     # graphql online web interface
     location = /graphiql {
--- a/infrastructure/conf/nginx-conf-local/default
+++ b/infrastructure/conf/nginx-conf-local/default
@@ -153,6 +153,13 @@ server {
         include fastcgi_params;
         fastcgi_param SCRIPT_FILENAME $document_root/api/public/index.php;
     }
+
+    # simulate being off the VPN and the firewall redirecting to the restricted page
+    # location = /admin/graphql {
+    #     absolute_redirect on;
+    #     return 301 http://localhost:8000/restricted.html;
+    # }
+
     # graphql online web interface
     location = /graphiql {
         fastcgi_pass 127.0.0.1:9000;
--- a/infrastructure/conf/nginx-conf-local/default
+++ b/infrastructure/conf/nginx-conf-local/default
@@ -155,7 +155,7 @@ server {
     }
 
     # simulate being off the VPN and the firewall redirecting to the restricted page
-    # location = /admin/graphql {
+    # location = /admin {
     #     absolute_redirect on;
     #     return 301 http://localhost:8000/restricted.html;
     # }
--- a/packages/client/src/exchanges/specialErrorExchange.ts
+++ b/packages/client/src/exchanges/specialErrorExchange.ts
@@ -18,11 +18,16 @@ const specialErrorExchange = ({ intl }: { intl: IntlShape }) => {
         tap((result) => {
           const err = result.error;
           if (err) {
-            const res = err.response as Response;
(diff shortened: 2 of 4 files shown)


infrastructure/conf/nginx-conf-local/default
```diff
@@ -153,6 +153,13 @@ server {
         include fastcgi_params;
         fastcgi_param SCRIPT_FILENAME $document_root/api/public/index.php;
     }
+
+    # simulate being off the VPN and the firewall redirecting to the restricted page
+    # location = /admin/graphql {
```
tristan-orourke, line 158: Shouldn't this be
```suggestion
    # location = /admin {
```
?

petertgiles: Oh yeah, that makes sense

petertgiles: Oh wait, the equals is an exact match.  This probably won't work anymore.

tristan-orourke: Ah, you're right

tristan-orourke: I think this might mirror what Prod is doing? When I use this, I see the same behaviour as on prod. Unfortunately, the error toast also seems to have stopped coming up...

```
    location ~ ^/(en|fr)?/admin {
        absolute_redirect on;
        return 301 http://localhost:8000/restricted.html;
    }
```

petertgiles: OK, I reworked this block to behave more like the prod one.
[reword firewall simulator](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12226/commits/3296a1247ebfa5cd6ce613e8b157cb6681390d29)

petertgiles: Since the regex version has a lower importance than an exact match I added a comment that you'll need to comment out the main rule for /admin/graphql since it will just silently override the firewall simulation rule otherwise.

packages/client/src/exchanges/specialErrorExchange.ts
```diff
@@ -18,13 +18,18 @@ const specialErrorExchange = ({ intl }: { intl: IntlShape }) => {
...
+            // I think this is old error condition from when the firewall responded directly when blocking
             if (
-              res?.status === 403 &&
+              errRes?.status === 403 &&
               err?.networkError?.message?.includes("Request Rejected")
             ) {
               toast.error(intl.formatMessage(errorMessages.requestRejected));
             }
+            // This is the new error condition from when the firewall responds a redirect to restricted.html
+            if (errRes?.url?.endsWith("/restricted.html")) {
```
tristan-orourke, line 30: So, if the current setup **is** working like the Prod firewall does, this still may not work. When I log errRes here, I see
```
redirected: false
status: 405
statusText: "Not Allowed"
type: "basic"
url: "http://localhost:8000/admin/graphql"
```

tristan-orourke (commented):
I believe canAccessProtectedEndpoint will work! I'm not sure about the specialErrorExchange, sadly, but I'm okay with approving this as is, since the toast was out of scope anyway.

packages/client/src/exchanges/specialErrorExchange.ts
petertgiles: That's weird... it works for me?
image

petertgiles: 405 "Not Allowed" is usually when you try to GET at the graphql endpoint.  How are you testing this?

tristan-orourke: Rebuilt and tried it again, and it works for me as well now. Maybe I had something set up wrong.

tristan-orourke (approved):
My previous issue went away. Probably I had something set up wrong. Looks good!

PR #12291, opened 2024-12-12
[Feat] Return locale aware string from `LocalizedString`

### LINKED ISSUE #11341: ♻️ Return already localized string with our `LocalizedString` type
## ♻️ Debt/Refactor

After completing the blocking tickets, we should have the ability to return an actually localized string from the API instead of both.

## 🕵️ Details

Right now, we send down both the English and French versions of dynamic strings (user generated). However, we really only need both when editing said strings. It would be nice if we could allow the consumers of our API to only get one of the strings so we don't over fetch.

## 🙋‍♀️ Proposed Solution

Since there are cases where we want both locales, this should add the localized version along side. That way you can get either in the query. the solution should be as simple as moving the type to a programmatic registration. We don't even need to add the scope of updating the client right now if we don't want to we can opt-in over time.

```php
$resolver = function ($value, $args, $context, ResolveInfo $info) use ($enum) {
  switch ($info->fieldName) {
    case 'en': return $value['en'];
    case 'fr': return $value['fr'];
    case 'localized': return $value[App::getLocale() ?? 'en'];
    default: return null;
  }
};

$typeRegistry->register(
  new ObjectType([
    'name' => 'LocalizedString',
    'fields' => function () use ($typeRegistry, $name): array {
      return [
        'en' => Type::nonNull(Type::string()),
        'fr' => Type::nonNull(Type::string()),
        'localized' => Type::nonNull(Type::string()),
      ];
    },
    'resolveField' => $resolver,
  ]);
);
```

## ✅ Acceptance Criteria

- [ ] Localized string contains a pre-localized version based on laravel locale

## 🛑 Blockers

```[tasklist]
### Blocked By
- [ ] https://github.com/GCTC-NTGC/gc-digital-talent/issues/11336
```

esizer:
🤖 Resolves #11341 

## 👋 Introduction

Updates the grpahql type for `LocalizedString` to return the localized string based on the app locale.

## 🕵️ Details

This is adding a custom cast that includes the current localized string along with both languages for the `LocalizedString` type. This does not update the frontend to use this new field.

## 🧪 Testing

1. Open graphqil `/graphiql`
2. Query for the models affected
3. Confirm `localized` is added to the return type
4. Switch the locale with the header `Accept-Language en|fr`
5. Confirm the localized message switches between the current locale as expected
6. Confirm setting values still functions

### DIFF
--- a/api/tests/Unit/PoolCandidatePolicyTest.php
+++ b/api/tests/Unit/PoolCandidatePolicyTest.php
@@ -159,7 +159,7 @@ class PoolCandidatePolicyTest extends TestCase
      *
      * @return void
      */
-    public function test_view_any()
+    public function testViewAny()
     {
         $this->assertFalse($this->guestUser->can('viewAny', PoolCandidate::class));
         $this->assertFalse($this->applicantUser->can('viewAny', PoolCandidate::class));
@@ -179,7 +179,7 @@ class PoolCandidatePolicyTest extends TestCase
      *
      * @return void
      */
-    public function test_view_draft()
+    public function testViewDraft()
     {
         $this->poolCandidate->submitted_at = null;
         $this->poolCandidate->save();
@@ -205,7 +205,7 @@ class PoolCandidatePolicyTest extends TestCase
      *
      * @return void
      */
-    public function test_view_submitted()
+    public function testViewSubmitted()
     {
         $this->poolCandidate->submitted_at = config('constants.past_date');
         $this->poolCandidate->save();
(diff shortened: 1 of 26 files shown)


api/tests/Unit/PoolCandidatePolicyTest.php
```diff
@@ -73,8 +73,8 @@ protected function setUp(): void
             ]);
 
         $this->team = Team::factory()->create(['name' => 'test-team']);
-        $this->community = Community::factory()->create(['name' => 'test-team']);
-        $this->otherCommunity = Community::factory()->create(['name' => 'suspicious-team']);
+        $this->community = Community::factory()->create(['name' => ['en' => 'test-team EN', 'fr' => 'test-team FR']]);
+        $this->otherCommunity = Community::factory()->create(['name' => ['en' => 'suspicious-team EN', 'fr' => 'suspicious-team FR']]);
```
yonikid15, line 77: Why not remove the `name` for these as well?

yonikid15 (approved):
Very cool! Nice work :+1: 
French:
image

English:
image

Issue #12164, opened 2024-12-02
✨ Applicant-facing PoolCandidate status chips


tristan-orourke:
## ✨ Feature

This task is to update the status chips used to represent the status of PoolCandidate objects to applicants.

Note: conceptuallly, PoolCandidate objects can be considered to represent two slightly different things at different points in their lifecycle. These are displayed separately, as Applications and as Recruitment Processes.
1. Initially, the PoolCandidate object represents an Application to a pool or job. This can be in draft, in assessment, or qualified/rejected.
2. If the Application is accepted (Qualified), the user can be considered be a candidate in a Recruitment Process. This can have statuses like "Looking for job", "suspended", "expired" or "removed".

## 🕵️ Details

Status Pill logic - I attempted to work through the whole logic [here](https://docs.google.com/spreadsheets/d/1ASOKPEV_FVbMM_cAs_e4o4i9eqn9RQiBsmRORYm6umE/edit?usp=sharing)

We only need to worry about showing a PoolCandidate object as a Recruitment Process if the user was Qualified in the process. ie finalDecisionAt is not null and finalDecision is one of
  - QUALIFIED;
  - QUALIFIED_EXPIRED;
  - QUALIFIED_PLACED;
  - QUALIFIED_REMOVED;

Note, QUALIFIED_PENDING indicates they're likely to be qualified, but finalDecision has not _actually_ been made yet, and finalDecisionAt will be null. From the application's perspective, the application is still in assessment.

## 🎨 Design

@JoshBeveridge 

[https://www.figma.com/design/oZhQe9ahidHFfaiqgREvxa/Dashboard-(All-users)?node-id=2086-63995&node-type=frame&t=1x50TV9VIcwOqroM-0](Figma link to where these appear in new dashboard designs.)

These statuses also appear on the old (current?) design, and thus this can be completed before #10788.

## 📸 Screenshot

Applications:
![image](https://github.com/user-attachments/assets/a7d4dc72-1e5f-40b3-a380-900720c1a964)

Recruitment processes:
![image](https://github.com/user-attachments/assets/66b2eb2b-a6de-4257-97cc-bc9aad20c379)

## 🙋‍♀️ Proposed Implementation

[...]

PR #12324, opened 2024-12-16
[Feat] Populate support form from search params

### LINKED ISSUE #12215: ✨ URL params for the support page to pre-populate the support form
## ✨ Feature

Create a way for us to pre-populate the support form based on a custom link that sent the user there.

## 🕵️ Details

When a user follows a link for "contact support" or similar, we want to help them out by adding some context to the form. This will help support and the user resolve things faster.

For example, we could add which process the user wants to follow up on, which community the ticket pertains to.

There are two relevant input fields on the support page: #subject and #description

The URL params should allow us to set either. For example: (use with caution, Gray wrote this)
`<a href="/en/support?subject=question&description=Add%20your%20question%20here:%0D%0A%0D%0AProcess%20number:%2012345678">`

Tristan says all the weird characters will probably just be handled gracefully by the browser. But regardless, this link should pre-populate the support form with "question" selected for the subject, and the following text in the description field:

```
Add your question here:

Process number: 12345678
```

## ✅ Acceptance Criteria

- [ ] URL params working for the support page

esizer:
🤖 Resolves #12215 

## 👋 Introduction

Uses search params to pre-populate the support form `subject` and `description` fields.

## 🧪 Testing

> [!TIP]
> When adding search params, make sure they are encoded for URIs.

2. Navigate to `/support`
3. Confirm the form is empty
4. Add search params for `subject` (only "bug", "feedback" and "question" are supported values) and `description`
5. Confirm the form is populated
6. Confirm unsupported subject fields leads to an empty "I'm looking to..." field
7. Confirm unsafe input to the `description` param is sanitized (i.e `<script>alert('unsafe');</script>` does nothing)

### DIFF
--- a/apps/playwright/tests/support-page.spec.ts
+++ b/apps/playwright/tests/support-page.spec.ts
@@ -16,6 +16,21 @@ test.describe("Support page", () => {
       const accessibilityScanResults = await makeAxeBuilder().analyze();
       expect(accessibilityScanResults.violations).toEqual([]);
     });
+    test("populates from search param", async ({ page }) => {
+      await page.goto("/en/support?subject=bug&description=test");
+      await expect(
+        page.getByRole("combobox", { name: /looking to/i }),
+      ).toHaveValue("bug");
+      await expect(page.getByRole("textbox", { name: /details/i })).toHaveValue(
+        "test",
+      );
+    });
+    test("does not populate invalid subject param", async ({ page }) => {
+      await page.goto("/en/support?subject=invalid");
+      await expect(
+        page.getByRole("combobox", { name: /looking to/i }),
+      ).toHaveValue("");
+    });
   });
   test.describe("Support form", () => {
     test("send POST request to existing API endpoint", async ({ request }) => {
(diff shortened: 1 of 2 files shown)


vd1992:
Question, where are the URLs supposed to be coming from? Do we have to generate them or is someone providing them to a user?

esizer:
> Question, where are the URLs supposed to be coming from? Do we have to generate them or is someone providing them to a user?

I think it can be both but likely that the majority will be generated on the client.

vd1992:
> > Question, where are the URLs supposed to be coming from? Do we have to generate them or is someone providing them to a user?
> 
> I think it can be both but likely that the majority will be generated on the client.

Okay, so there should be a separate ticket then to generate the URLs then? I was a bit confused as to whether the issue scope included generating the URLs too

petertgiles:
> Question, where are the URLs supposed to be coming from? Do we have to generate them or is someone providing them to a user?

This issue is part of the new applicant dashboard milestone.  You can see if reference to it here: https://github.com/GCTC-NTGC/gc-digital-talent/issues/12164 .  I think this part of the whole "if you've been screened out, reach out to us to chat about it" thing.

PR #12325, opened 2024-12-17
[Feature] Ad hoc email command

### LINKED ISSUE #11542: ✨ Improved laravel Command for sending arbitrary email notifications
## ✨ Feature

We sometimes want to be able to be able to define a new GC Notify email template and send that to many users, without having to deploy new code. This will look similar to SendNotificationsSystem.php, but be more flexible.

## 🕵️ Details

The new Laravel command should accept the following parameters:
- english template id (required)
- french template id (required)
- array of emails (optional)
- array of NotificationFamily enums (optional)
- boolean 'notifyAllUsers' (optional)

Behaviour:
The first step (conceptually) is to find all the users which should be notified. The method of doing so depends on which optional parameters have been supplied.
- if all optional parameters are null, or more than one are not null, then throw a readable error. Only one should be set at a time.
- if notifyAllUsers is true, notify ALL users
- If the list of emails is populated, and NotificationFamily is null, then look up users by their contact email. (Log any emails which cannot be found.)
- If NotificationFamilies have been supplied and emails are null, then notify all users who have _any_ of the listed NotificationFamilies in enabled_email_notifications. (It may be worth adding a scope to the User model for this.)

When the list of users has been retrieved, iterate through and create queued jobs to send the notification. Use the English or French GCNotify template based on the user's communication language preference. For [personalization](https://documentation.notification.canada.ca/en/send.html#sending-an-email), include the user's first and last names. We will assume that any notifications sent this way will be fairly basic, and only require the user's name.

Look at how `api/app/Console/Commands/SendNotificationsApplicationDeadlineApproaching.php` and `api/app/Console/Commands/SendNotificationsSystem.php` work.

May be instead of array of emails , filename param which contains the emails would be easy to deal with? 

## ✅ Acceptance Criteria

- [ ] Please refer to the Details section

petertgiles:
🤖 Resolves #11542 

## 👋 Introduction

Adds an Artisan command and Laravel notification for sending ad hoc emails with just the first name and last name.

## 🕵️ Details

The command will select users to notify by one of three criteria:
- specific email addresses
- specific notification families
- all users

## 🧪 Testing

1. Update your api/.env file with the "team" api key
2. Rerun ./artisan optimize:clear
3. Start a queue worker
4. Update one user to have your GCNotify registered email address
5. Test various forms of the Artisan command

- [x] Can use email criteria
    `./artisan send-notifications:ad-hoc-email cdaa82f1-7f76-4d9e-a82e-328c41a6ab62 0831e036-8dfb-4f5e-8b31-867b2c526d77 --emailAddress=email1@example.org --emailAddress=email2@example.org`
- [x] Can use notification family criteria
    `./artisan send-notifications:ad-hoc-email cdaa82f1-7f76-4d9e-a82e-328c41a6ab62 0831e036-8dfb-4f5e-8b31-867b2c526d77 --notificationFamily=JOB_ALERT --notificationFamily=APPLICATION_UPDATE`
- [x] Can use notify all criteria (OR logic)
    `./artisan send-notifications:ad-hoc-email cdaa82f1-7f76-4d9e-a82e-328c41a6ab62 0831e036-8dfb-4f5e-8b31-867b2c526d77 --notifyAllUsers`
- [x] Respects language preferences
- [x] Can not use 0 or 2 or 3 criteria types
- [x] Warns of missing email addresses
- [x] Errors on invalid notification families
- [x] first and last name passed to GC Notify

## 📸 Screenshot

image

### DIFF
--- a/api/app/Console/Commands/SendNotificationsAdHocEmail.php
+++ b/api/app/Console/Commands/SendNotificationsAdHocEmail.php
@@ -4,19 +4,20 @@ namespace App\Console\Commands;
 
 use App\Enums\NotificationFamily;
 use App\Models\User;
+use App\Notifications\AdHocGcNotifyEmail;
 use Illuminate\Console\Command;
 use Illuminate\Contracts\Console\PromptsForMissingInput;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Collection;
 
-class SendNotificationsEmail extends Command implements PromptsForMissingInput
+class SendNotificationsAdHocEmail extends Command implements PromptsForMissingInput
 {
     /**
      * The name and signature of the console command.
      *
      * @var string
      */
-    protected $signature = 'send-notifications:email
+    protected $signature = 'send-notifications:ad-hoc-email
                                 {templateIdEn : The template ID for the English email message}
                                 {templateIdFr : The template ID for the French email message}
                                 {--emailAddress=* : The list of contact email addresses to send the notification to}
@@ -29,7 +30,7 @@ class SendNotificationsEmail extends Command implements PromptsForMissingInput
      *
      * @var string
      */
-    protected $description = 'Send email notifications from given GC Notify templates';
+    protected $description = 'Send ad hoc email notifications from given GC Notify templates';
 
     /**
      * Execute the console command.
@@ -46,11 +47,12 @@ class SendNotificationsEmail extends Command implements PromptsForMissingInput
 
         if ($this->confirm('Do you wish to send notifications to '.$userCount.' users?')) {
             $progressBar = $this->output->createProgressBar($userCount);
+            $notification = new AdHocGcNotifyEmail($this->argument('templateIdEn'), $this->argument('templateIdFr'));
 
-            $users->chunk(200, function (Collection $chunkOfUsers) use (&$successCount, &$failureCount, $progressBar) {
+            $users->chunk(200, function (Collection $chunkOfUsers) use (&$successCount, &$failureCount, $progressBar, $notification) {
                 foreach ($chunkOfUsers as $user) {
                     try {
-                        // $this->sendNotification($user, $successCount, $failureCount);
+                        $user->notify($notification);
                         $successCount++;
                     } catch (\Throwable $e) {
                         $this->error("Failed to send notification to user $user->id: ".$e->getMessage());
@@ -83,12 +85,13 @@ class SendNotificationsEmail extends Command implements PromptsForMissingInput
 
         $options = $this->options();
 
-        $optionSelectedCount =
+        // How many types of options were used?
+        $optionTypesCount =
             (count($emailAddresses) > 0 ? 1 : 0) +
             (count($notificationFamilies) > 0 ? 1 : 0) +
             ($notifyAllUsers ? 1 : 0);
 
-        if ($optionSelectedCount != 1) {
+        if ($optionTypesCount != 1) {
             throw new \Error('Must filter users using exactly one of the option types');
         }
 
--- a/api/app/Console/Commands/SendNotificationsAdHocEmail.php
+++ b/api/app/Console/Commands/SendNotificationsAdHocEmail.php
@@ -93,7 +93,7 @@ class SendNotificationsAdHocEmail extends Command implements PromptsForMissingIn
             ($notifyAllUsers ? 1 : 0);
 
         if ($optionTypesCount != 1) {
-            throw new \Error('Must filter users using exactly one of the option types');
+            throw new \InvalidArgumentException('Must filter users using exactly one of the option types');
         }
 
         if (count($emailAddresses) > 0) {
@@ -108,7 +108,7 @@ class SendNotificationsAdHocEmail extends Command implements PromptsForMissingIn
             return User::query();
         }
 
-        throw new \Error('Unexpected function end point for options: '.json_encode($options));
+        throw new \InvalidArgumentException('Unexpected function end point for options: '.json_encode($options));
     }
 
     private function builderFromEmailAddresses(array $requestedEmailAddresses): Builder
@@ -134,7 +134,7 @@ class SendNotificationsAdHocEmail extends Command implements PromptsForMissingIn
         $allNotificationFamilies = array_column(NotificationFamily::cases(), 'name');
         foreach ($notificationFamilies as $notificationFamily) {
             if (! in_array($notificationFamily, $allNotificationFamilies)) {
-                throw new \Error('Invalid notification family: '.$notificationFamily);
+                throw new \InvalidArgumentException('Invalid notification family: '.$notificationFamily);
             }
         }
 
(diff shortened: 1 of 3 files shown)


api/app/Console/Commands/SendNotificationsAdHocEmail.php
```diff
@@ -0,0 +1,148 @@
...
+            $this->alert('The following email addresses were not found:');
+            foreach ($missingEmailAddresses as $missingEmailAddress) {
+                $this->warn($missingEmailAddress);
+            }
+        }
+
+        return $builder;
+    }
+
+    private function builderFromNotifictionFamilies(array $notificationFamilies): Builder
```
brindasasi, line 131: minor typo - builderFromNotificationFamilies

api/app/Console/Commands/SendNotificationsAdHocEmail.php
```diff
@@ -0,0 +1,148 @@
...
+        $options = $this->options();
+
+        // How many types of options were used?
+        $optionTypesCount =
+            (count($emailAddresses) > 0 ? 1 : 0) +
+            (count($notificationFamilies) > 0 ? 1 : 0) +
+            ($notifyAllUsers ? 1 : 0);
+
+        if ($optionTypesCount != 1) {
+            throw new \Error('Must filter users using exactly one of the option types');
```
brindasasi, line 96: This can simply use InvalidArgumentException .. but your choice :)

api/app/Console/Commands/SendNotificationsAdHocEmail.php
```diff
@@ -0,0 +1,148 @@
...
+    public function handle()
+    {
+        $successCount = 0;
+        $failureCount = 0;
+
+        $users = $this->getUsersToSendNotificationTo();
+        $userCount = $users->count();
+
+        if ($this->confirm('Do you wish to send notifications to '.$userCount.' users?')) {
+            $progressBar = $this->output->createProgressBar($userCount);
```
brindasasi, line 49: oo.. progressbar!!!

brindasasi (commented):
Code looks good so far. Will verify in dev

api/app/Console/Commands/SendNotificationsAdHocEmail.php
petertgiles: Good catch!
[fix function name typo](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12325/commits/cbbaab7789fbd70d449e37d6c4bee07528b1905e)

api/app/Console/Commands/SendNotificationsAdHocEmail.php
petertgiles: You're right - that's more descriptive.
[switch to InvalidArgumentException](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12325/commits/957001b96edad72e65c931b3421339f822a4f5f1)

api/app/Console/Commands/SendNotificationsAdHocEmail.php
petertgiles: Trying out some of the new Laravel 11 features.  :wink:

brindasasi:
Sorry for delaying on this  what I noticed so far 
* If I gave wrong template ID but valid email addresses , it gives me success message even though email doesn't get sent out 
* separtating the email id params  indivdiaully like. this are not very practical when we have to send bulk emails. 
--emailAddress=email1@example.org --emailAddress=email2@example.org 
It would be better to have an array with comma separated 

testing few more things

brindasasi (changes requested):
As mentioned, when the template is mistakenly given it is failing silently 
image

when there is multiple critieras given , I got different exception rather than the invalidargument .. but I'm not too worried about this .. as long as it breaks we know 
But if the above one is fixed , would be great.
image

Also if we can have comma separated emails for bulk email would be great.
Other than that this looks awesome!

petertgiles:
Thanks for the review!  Some feedback on your requests:

> If I gave wrong template ID but valid email addresses , it gives me success message even though email doesn't get sent out

This is because sending emails is an asynchronous process.  They aren't sent out when the command is run so we can't immediately display API errors like invalid template ID.  That will be logged in log files when the queue worker tries to send it.  I don't think we want to make this synchronous either.  We need to be able to slow it down to satisfy the API rate limiter and be able to retry later if the service goes offline.  This is easy to forget so updated the messages to try to be a bit clearer:
[use word queue in messages](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12325/commits/57dafe82737d2069796ea62e2ae631884ab53272)

> separtating the email id params indivdiaully like. this are not very practical when we have to send bulk emails.
--emailAddress=[email1@example.org](mailto:email1@example.org) --emailAddress=[email2@example.org](mailto:email2@example.org)
It would be better to have an array with comma separated

The AC of the issue didn't specify how the options should be sent so I went with [the built-in Laravel option](https://laravel.com/docs/11.x/artisan#option-arrays): When defining an option that expects multiple input values, each option value passed to the command should be prefixed with the option name.  If we want to manually parse a comma separated list we can do that but let's decide on a method and add it to the AC.  @tristan-orourke , preferences?

> when there is multiple critieras given , I got different exception rather than the invalidargument

In your screenshot it looks like your log file is not writable which is an unrelated error.  I think we will get the right exception for multiple criteria.
image

tristan-orourke:
Ah, I didn't realize Laravel had a built-in option like that for list inputs. Hmmm. When we need to use this command, we're likely to be starting from a comma-separated list of emails... However, it would only take a bit of find/replace in a text editor to modify that. Let's stick to the Laravel spec, and keep it as is.

brindasasi:
> Thanks for the review! Some feedback on your requests:
> 
> > If I gave wrong template ID but valid email addresses , it gives me success message even though email doesn't get sent out
> 
> This is because sending emails is an asynchronous process. They aren't sent out when the command is run so we can't immediately display API errors like invalid template ID. That will be logged in log files when the queue worker tries to send it. I don't think we want to make this synchronous either. We need to be able to slow it down to satisfy the API rate limiter and be able to retry later if the service goes offline. This is easy to forget so updated the messages to try to be a bit clearer: [use word queue in messages](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12325/commits/57dafe82737d2069796ea62e2ae631884ab53272)
> 
> > separtating the email id params indivdiaully like. this are not very practical when we have to send bulk emails.
> > --emailAddress=[email1@example.org](mailto:email1@example.org) --emailAddress=[email2@example.org](mailto:email2@example.org)
> > It would be better to have an array with comma separated
> 
> The AC of the issue didn't specify how the options should be sent so I went with [the built-in Laravel option](https://laravel.com/docs/11.x/artisan#option-arrays): When defining an option that expects multiple input values, each option value passed to the command should be prefixed with the option name. If we want to manually parse a comma separated list we can do that but let's decide on a method and add it to the AC. @tristan-orourke , preferences?
> 
> > when there is multiple critieras given , I got different exception rather than the invalidargument
> 
> In your screenshot it looks like your log file is not writable which is an unrelated error. I think we will get the right exception for multiple criteria. image

This error was on dev. I can try again. Could be related to the issue you solved yesterday

PR #12334, opened 2024-12-18
[Feat] Backend for employee profile

### LINKED ISSUE #12001: ✨ Backend data for Government Career Planning Employee Profile
## ✨ Feature

Add a EmployeeProfile model with a one-to-one relationship to User so that applicants can store the data associated with new Employee Profile section. 

Note: this does NOT include the community-specific data

## 🕵️ Details

QUESTION: how strict do we want to be about non-gov employees? Should we do any validation which requires user to be employees before updating this data? If its not important, would be easier not to. What if a user stops being a gov employee?
ANSWER: we won't worry about validating this on the backend.

Add a way for users to store their:
- talent mobility preferences
- dream job
- career goals and work style

## 🎨 Design

https://www.figma.com/design/Xnkie0dSq9RujvQRAZwrY4/Employee-profile-(Employees)?node-id=26-18365&node-type=frame&t=VeY3Kg8mOWlgGWXb-0

@JoshBeveridge 

> [!IMPORTANT]
> Some high-level details of the proposed implementation have changed a bit since estimating. 
> - In the graphql schema, we'll keep EmployeeProfile a separate model from User. This is because the permissions will differ, and different roles will need access to one or the other.
> - However, in the database, we will implement EmployeeProfile as additional fields on the users table. This will simplify the implementation of different relationships. This is similar to what we did with UserAuthInfo.
> - Also, the mutations to update the Employee Profile can be simplified to a single mutation. It is not necessary for our backend validation to directly match frontend validation, as long as the data saved to the backend remains sensible and valid.

### Enums

We will need some new enums to support the form options and storing them in the database. I think it makes sense to make these localized enums.

1. Organization types
    - Current department/agency/crown corp
    - Other departments
    - Other agencies
    - Other crown corps
2. Promotions/Lateral moves
    - Above level/promotion
    - At level/lateral
    - Below level/demotion
3. Mentorship status
    - Not participating
    - Mentee
    - Mentor
    - Both mentor/mentee
4. Mentorship interest
    - Mentor
    - Mentee
5. Exec coaching status
    - Not participating
    - Have a coach
    - Is a coach
    - Both coaching/learning
6. Exec coaching interest
    - Coaching
    - Learning

### Database

These fields will be added to the `users` table, backing the new EmployeeProfile graphql type.

#### Talent mobility options

 - `career_planning_organization_type_interest` - jsonb array, nullable (enum 1)
 - `career_planning_move_interest` - jsonb array, nullable (enum 2)
 - `career_planning_mentorship_status` - string, nullable (enum 3)
 - `career_planning_mentorship_interest` - jsonb array, nullable (enum 4)
 - `career_planning_exec_interest` - bool, nullable
 - `career_planning_exec_coaching_status` - string, nullable (enum 5)
 - `career_planning_exec_coaching_interest` jsonb array, nullable (enum 6)

#### Dream role

 - `dream_role_community_id` - foreign key, nullable 
 - `dream_role_title` - string, nullable
 - `dream_role_additional_information` - string, nullable
 - `dream_role_classification_id` - FK, classifications:id
 - `dream_role_work_stream_id` FK, work_streams:id

##### Dream job relationships

This relationship will require a new table. Ensure there is a user_id/department_id uniqueness constraint, and a user can't add the same department multiple times.
 - `user_dream_role_departments` - Table
     - `user_id` - FK, users:id
     - `department_id` - FK, departments:id

#### Goals, Work Style

 - `career_planning_about_you` - string, nullable
 - `career_planning_career_goals` - string, nullable
 - `career_planning_learning_goals` - string, nullable
 - `career_planning_work_style` - string, nullable

#### Other

The EmployeeProfile should also have a `userPublicProfile: UserPublicProfile` field, so users accessing the profile can see the associated name and email, without accessing more private data.

#### Queries

Create new EmployeeProfile model, with a one-to-one relationship with User. It should be protected by `@canResolved(ability: "view")`, because some managers who may have permission to view the applicant profile should not be able to see the employee profile.

```
type User {
  type employeeProfile: EmployeeProfile @self @canResolved(ability: "view")
}

type EmployeeProfile {
  ...
  userPublicProfile: UserPublicProfile @self
}
```
One implementation I'm not yet sure about is whether EmployeeProfile should have its own Laravel Model, which is backed by the 'users; table, or whether it should use the `@model(class: "\\App\\Models\\User")` directive like UserPublicProfile. 
- The former keeps EmployeeProfile-related relationships and Policies organized separately from the rest of the User, but maybe makes later relationship work more complicated (I'm not sure). 
- The latter keeps the User.php model growing, and means we need to add new policy methods to UserPolicy.php and do something like type `User {  type employeeProfile: EmployeeProfile @self @canResolved(ability: "viewEmployeeProfile") }`. 

#### Mutations

Create new mutation `updateEmployeeProfile`, where id is required (note that EmployeeProfile.id and User.id are the same) and all other fields are optional. We don't need a create mutation because users don't create Users, and if a user exists so does its "nested" EmployeeProfile! 

### Validation

All the new fields are nullable, and type constraints are handled by Lighthouse Graphql. The only extra validation I can think is possibly ensuring that the same department is not added to Dream Roles multiple times. Even that seems better implemented as a database constraint.

One of the frontend forms which updates EmployeeProfile requires all its fields before submitting, but I'm not concerned about validating that on the backend here. (I'm all for splitting up mutations into smaller, more focused ones, where it provides value, but upon reflection this case was becoming an anti-graphql-ish pattern.)

### Permissions

Permissions check if the authenticated user has permission to view or update their own employee profile (`view-own-employeeProfile`, `update-own-employeeProfile`). Create these permissions if they don't exist yet. Create EmployeeProfilePolicy if it doesn't exist yet, or depending on implementation details, add new methods to UserPolicy.

### Snapshot

This new data does NOT need to be saved in applicant snapshots

## ✅ Acceptance Criteria

- [x] New enums created
- [x] Database migrations created - mostly adding columns to users table, with one new pivot table
- [x] GraphQL schema updated with new EmployeeProfile type 
  - [x] include one new updateEmployeeProfile mutation, protected by new permission check
  - [x] User type includes link to EmployeeProfile type, protected by new permission check
- [x] New fields taken into consideration for seeders and factories
- [x] Create `view-own-employeeProfile` and `update-own-employeeProfile` permissions
```[tasklist]
### Tasks
```

esizer:
🤖 Resolves #12001 

## 👋 Introduction

Adds the `EmployeeProfile` and all supporting code to the backend.

## 🧪 Testing

1. Confirm matches requirements outlined in the ticket
2. Refresh DB to get an employee profile on `applicant@test.com` by running `make seed-fresh`
3. Open graphiql `/graphiql`
4. Make sure you are authenticated as `applicant@test.com` (from env or bearer token)
5. Query for your own users employee profile
6. Confirm it returns as expected
7. Query for another users employee profile
8. Confirm unauthorized
9. Repeat steps 5-8 but with the update mutation


api/app/Enums/Mentorship.php
```diff
@@ -0,0 +1,18 @@
...
+namespace App\Enums;
+
+use App\Traits\HasLocalization;
+
+enum Mentorship
+{
+    use HasLocalization;
+
+    case MENTOR;
+    case MENTEE;
```
mnigh, line 12: Might be worth adding this to our dictionary since mentee is a real word.

mnigh (commented):
I ran `make refresh-api; make seed-fresh; pnpm i; pnpm build:fresh`, navigate to http://localhost:8000/graphiql and received:

```
{
  "errors": [
    {
      "message": "Introspection must provide input type for arguments, but received: UserPublicProfile.",
      "stack": "Error: Introspection must provide input type for arguments, but received: UserPublicProfile.\n    at buildInputValue (http://unpkg.com/graphiql/graphiql.min.js:52187:13)\n    at keyValMap (http://unpkg.com/graphiql/graphiql.min.js:43886:27)\n    at buildInputValueDefMap (http://unpkg.com/graphiql/graphiql.min.js:52181:37)\n    at fields (http://unpkg.com/graphiql/graphiql.min.js:52153:21)\n    at resolveObjMapThunk (http://unpkg.com/graphiql/graphiql.min.js:48531:40)\n    at defineInputFieldMap (http://unpkg.com/graphiql/graphiql.min.js:49121:20)\n    at GraphQLInputObjectType.getFields (http://unpkg.com/graphiql/graphiql.min.js:49089:27)\n    at collectReferencedTypes (http://unpkg.com/graphiql/graphiql.min.js:50931:51)\n    at new GraphQLSchema (http://unpkg.com/graphiql/graphiql.min.js:50764:9)\n    at Object.buildClientSchema (http://unpkg.com/graphiql/graphiql.min.js:52014:10)"
    }
  ]
}
```

Is there a step I am missing?

api/app/Enums/Mentorship.php
esizer: Good call, added in [add mentee to dictionary](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12334/commits/9f63e141b759a402df2d5326f74a2d469b38e12a)

esizer:
> Is there a step I am missing?

Nope, I made a copy past error. Fixed in [fix graphql input for employee profile](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12334/commits/2a201f4deed1aa6a34320ddef2eff365cba3a6c5)

mnigh:
Having trouble with returning any values (this after having run `make refresh-api; make seed-fresh; pnpm i; pnpm build:fresh`).

> 4. Make sure you are authenticated as applicant@test.com (from env or bearer token)
> 5. Query for your own users employee profile
> 6. Confirm it returns as expected

## Screen recording
https://github.com/user-attachments/assets/3a9184cf-c870-4c4c-b841-ccd17e9e7a3f

api/database/migrations/2024_12_18_133424_create_employee_profiles_table.php
```diff
@@ -0,0 +1,90 @@
...
+            $table->dropForeign(['dream_role_community_id']);
+            $table->dropForeign(['dream_role_work_stream_id']);
+            $table->dropForeign(['dream_role_classification_id']);
+
+            $table->dropColumn('dream_role_community_id');
+            $table->dropColumn('dream_role_classification_id');
+            $table->dropColumn('dream_role_work_stream_id');
+        });
+
+        Schema::dropIfExists('user_dream_role_department');
```
mnigh, line 88: Shouldn't this be the same as the `department_user_dream_role` in the `up()`?

api/database/migrations/2024_12_18_133424_create_employee_profiles_table.php
```diff
@@ -0,0 +1,90 @@
...
+                ->constrained(table: 'classifications', column: 'id')
+                ->onDelete('cascade');
+            $table->foreignUuid('dream_role_work_stream_id')
+                ->nullable()
+                ->constrained(table: 'work_streams', column: 'id')
+                ->onDelete('cascade');
+
+        });
+
+        Schema::create('department_user_dream_role', function (Blueprint $table) {
```
mnigh, line 46: The [issue](https://github.com/GCTC-NTGC/gc-digital-talent/issues/12001#:~:text=user_dream_role_departments%20%2D%20Table) has a differently named table, was this name change on purpose?

api/graphql/schema.graphql
```diff
@@ -148,6 +151,35 @@ type User {
...
+  workStyle: String @rename(attribute: "career_planning_work_style")
+  dreamRoleTitle: String @rename(attribute: "dream_role_title")
+  dreamRoleAdditionalInformation: String
+    @rename(attribute: "dream_role_additional_information")
+  dreamRoleCommunity: Community @belongsTo
+  dreamRoleClassification: Classification @belongsTo
+  dreamRoleWorkStream: WorkStream @belongsTo
+  dreamRoleDepartments: [Department!] @belongsToMany
+
+  userPublicProfile: UserPublicProfile
```
mnigh, line 180: The [issue](https://github.com/GCTC-NTGC/gc-digital-talent/issues/12001#:~:text=It%20should%20be%20protected%20by%20%40canResolved(ability%3A%20%22view%22)%2C%20because%20some%20managers%20who%20may%20have%20permission%20to%20view%20the%20applicant%20profile%20should%20not%20be%20able%20to%20see%20the%20employee%20profile.) states: `userPublicProfile: UserPublicProfile @self`.

api/storage/app/lighthouse-schema.graphql
```diff
@@ -602,6 +623,7 @@ type User {
   locationExemptions: String
   acceptedOperationalRequirements: [LocalizedOperationalRequirement]
   positionDuration: [PositionDuration]
+  employeeProfile: EmployeeProfile
```
mnigh, line 626: The [issue](https://github.com/GCTC-NTGC/gc-digital-talent/issues/12001#:~:text=type%20User%20%7B%0A%20%20type%20employeeProfile%3A%20EmployeeProfile%20%40self%20%40canResolved(ability%3A%20%22view%22)%0A%7D) states: `type employeeProfile: EmployeeProfile @self @canResolved(ability: "view")`.

api/database/migrations/2024_12_18_133424_create_employee_profiles_table.php
esizer: I think I just wrote it wrong in the issue. This one follows the laravel convention of alphabetical table name ordering in pivot tables

api/graphql/schema.graphql
esizer: Does it work without self? Cause it should, `@self` is when you want to reference the root model which this does not 🤔

api/storage/app/lighthouse-schema.graphql
esizer: Same as the last one, this is a different model than `User` so `@self` does not make sense here. Unless I just do not understand the `@self` directive 😅

api/database/migrations/2024_12_18_133424_create_employee_profiles_table.php
esizer: Yup! Fixed in [fix down migration](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12334/commits/785c731048a96161a4ff8fee9c04fe103c7ab1cb)

mnigh (approved):
It would be nice to have another reviewer on this PR before merging.

api/app/Enums/ExecCoaching.php
```diff
@@ -0,0 +1,18 @@
...
+
+use App\Traits\HasLocalization;
+
+enum ExecCoaching
+{
+    use HasLocalization;
+
+    case COACHING;
+    case LEARNING;
+
```
brindasasi, line 13: **Exec coaching status**
- Not participating
- Have a coach
- Is a coach
- Both coaching/learning

same question as Mentorship status below 
Are we not storing  Not participating as sep enum value?

api/graphql/schema.graphql
brindasasi: I agree this field is about relationship between EmployeeProfile and UserPublicProfile, which are also different types.

brindasasi (commented):
Just those questions , others looking great to me

api/storage/app/lighthouse-schema.graphql
brindasasi: I agree this field is about relationship between EmployeeProfile and UserPublicProfile, which are also different types. @self is not needed in here

api/app/Enums/ExecCoaching.php
esizer: I envisioned it being an array.

- `[]` = not participating
- `null` = not provided

brindasasi: that works as well!

brindasasi (approved):
LGTM!

PR #12356, opened 2024-12-19
[Feat] Shard playwright tests in CI

### LINKED ISSUE #12187: 🛠️ Shard playwright tests
## 🛠️ Tooling

Update our playwright CI workflow to shard tests across multiple jobs.

## 🕵️ Details

Right now our playwright job is taking the longest to run in CI and it would be nice to find small wins in reducing the run time. I think using shards should be an easy way to speed up with workflow a bit.

## 🙋‍♀️ Proposed Solution

I'm hoping it is as easy as following the [guide within the playwright documentation](https://playwright.dev/docs/test-sharding#github-actions-example).

## ✅ Acceptance Criteria

- [ ] Playwright workflow shards test job
- [ ] Reports are merged upon completion
- [ ] Tests still consistently pass
- [ ] There is a noticeable reduction in run time of the total workflow

esizer:
🤖 Resolves #12187 

## 👋 Introduction

Updates the playwright workflow to utilize shards, reducing the amount of time needed to run.

## 🕵️ Details

This makes a few crucial changes:

- Updates the report to blob (required to merge reports)
- Runs all the tests in a single command instead of separate for each browser (makes balancing shards more even)
- Shards the tests into 4 different matrices
- Merges all reports into a single html report at the end instead of having a report per browser

### Time savings

We have went for 44mins to 22mins for total run time which is  a 50% reduction in total runtime. We can probably squeeze some more savings if we [split some larger tests files up](https://github.com/GCTC-NTGC/gc-digital-talent/issues/12357) to better leverage parallelism.

## 🧪 Testing

1. Confirm playwright workflow runs as expected

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -107,8 +107,9 @@ jobs:
     # Merge reports after playwright-tests, even if some shards have failed
     if: ${{ !cancelled() }}
     needs: [playwright]
-
     runs-on: ubuntu-latest
+    env:
+      PNPM_VERSION: "9.12.3"
     steps:
     - uses: pnpm/action-setup@v4
       with:
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -138,5 +138,5 @@ jobs:
       uses: actions/upload-artifact@v4
       with:
         name: html-report--attempt-${{ github.run_attempt }}
-        path: playwright-report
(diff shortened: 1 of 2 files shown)


vd1992 (approved):
Impressive! I reran it and it all still seems fine. 
The shards aren't quite equally distributed but that could be a pain to keep equal, so good enough

image


Great! 🏆

Issue #12223, opened 2024-12-06
🐛 Graphiql unavailable in prod


petertgiles:
## 🐛 Bug

It seems like Graphiql is unavailable in prod.

## 🕵️ Details

I see some logs saying `file_put_contents(/tmp/api/storage/framework/views...) Failed to open stream: Permission denied"` which I'm guessing is it.

## 📋 Steps to Reproduce

1. https://talent.canada.ca/graphiql

## 📸 Screenshot

![image](https://github.com/user-attachments/assets/bad029ac-7be6-4c5f-83df-76c020a0729f)

## ✅ Acceptance Criteria

- [ ] Graphiql is available in prod

PR #12359, opened 2024-12-19
[Debt] Remove sanitize string function

### LINKED ISSUE #11370: ♻️ Remove `sanitizeString`
## ♻️ Debt

Remove the `sanitizeString` function.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/970540aebe50ddfa0d33c440938b190c467ca1d1/packages/forms/src/utils.ts#L233

## 🕵️ Details

> [!IMPORTANT]
> There is a PHP function of the same name that should not be removed. 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/5996e758ea13fecba0b58ab63f95f2de0b31b5be/api/app/Traits/Generator/GeneratesFile.php#L76-L85

## 🙋‍♀️ Proposed Solution

Remove function `sanitizeString` and any implementations of it in the codebase.

## ✅ Acceptance Criteria

- [ ] no instances of JavaScript `sanitizeString` string in codebase

## 🛑 Blockers

```[tasklist]
### Blocked By
- [x] Meeting with IMTD about firewall
```

esizer:
🤖 Resolves #11370 

## 👋 Introduction

Removes our `sanitizeString` function in favour of adding characters to the firewall allow list.

## 🧪 Testing

2. Confirm you can submit forms with text inputs

### DIFF
--- a/packages/forms/src/components/Input/Input.tsx
+++ b/packages/forms/src/components/Input/Input.tsx
@@ -10,7 +10,6 @@ import useFieldState from "../../hooks/useFieldState";
 import useFieldStateStyles from "../../hooks/useFieldStateStyles";
 import useInputDescribedBy from "../../hooks/useInputDescribedBy";
 import useInputStyles from "../../hooks/useInputStyles";
-import { sanitizeString } from "../../utils";
 
 export type InputProps = HTMLInputProps &
   CommonInputProps & {
@@ -61,7 +60,6 @@ const Input = ({
     if (whitespaceTrim) {
       inputValue = inputValue.trim();
     }
-    inputValue = sanitizeString(inputValue);
     setValue(name, inputValue);
   };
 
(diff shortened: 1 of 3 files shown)


vd1992:
Hm, I think the important thing to check here is whether the Narrow No-Break Space was adjusted for in the firewall. Only safe to remove the function if it is the case 😆

petertgiles:
> whether the Narrow No-Break Space was adjusted for in the firewall

Try it out!  [Graphiql is down in prod](https://github.com/GCTC-NTGC/gc-digital-talent/issues/12223), unfortunately, but you could still try it with browser edit and resend or postman.

petertgiles (approved):
Thanks for cleaning this up!  I used postman to send a nnbsp directly to the prod API and it worked fine.

PR #12372, opened 2024-12-24
[Fix] Storage permissions after optimize

### LINKED ISSUE #12223: 🐛 Graphiql unavailable in prod

esizer:
🤖 Resolves #12223 

## 👋 Introduction

Re-runs the permission after optimize command to hopefully fix them.

## 🕵️ Details

I think this lined up with the addition of `php artisan optimize` after changing permissions. I think that deletes and/or rewrites some files that could have modified permissions on them to something undesirable. Since the graphiql error seems to be permissions related, this was my best guess.

## 🧪 Testing

> [!NOTE]
> This requires a deployment, or at least running `post_deployment.sh`

1. Run `post_deployment.sh`
2. Confirm `/graphiql` loads correctly with no error

### DIFF
--- a/infrastructure/bin/post_deployment.sh
+++ b/infrastructure/bin/post_deployment.sh
@@ -46,8 +46,6 @@ cd /home/site/wwwroot/api
 # Laravel local cache
 if
     mkdir --parents /tmp/bootstrap/cache /tmp/api/storage/framework/cache/data && \
-    chown www-data:www-data /tmp/bootstrap/cache && \
-    chown -R www-data:www-data /tmp/api/storage && \
     php artisan optimize && \
     chown www-data:www-data /tmp/bootstrap/cache && \
     chown -R www-data:www-data /tmp/api/storage;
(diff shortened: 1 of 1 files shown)


infrastructure/bin/post_deployment.sh
```diff
@@ -48,7 +48,9 @@ if
     mkdir --parents /tmp/bootstrap/cache /tmp/api/storage/framework/cache/data && \
     chown www-data:www-data /tmp/bootstrap/cache && \
     chown -R www-data:www-data /tmp/api/storage && \
```
brindasasi, line 50: Do we still need these lines if we are going to append it with optimize?

esizer: I'm actually not sure. We can test it to see?

esizer: Okay, removed to see if we can go without.

[remove possible redundant permission change](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12372/commits/14f8e53a6980fb4295f05543114a4c1496d2399f)

brindasasi: still works fine on Dev 👍 .  Good to go then!

brindasasi (approved):
LGTM!

PR #12373, opened 2024-12-24
[Debt] Migrate from `framer-motion` to `motion`

### LINKED ISSUE #12352: ♻️ Migrate from `framer-motion` to `motion`
## ♻️ Debt/Refactor

It might be a good idea to migrate to `motion`.

## 🕵️ Details

The author of `framer-motion` has [made his library independent of Framer](https://motion.dev/blog/framer-motion-is-now-independent-introducing-motion) since leaving the organization. The library is now completelty independent. It would likely be a good idea to move over?

## 🙋‍♀️ Proposed Solution

Hoping it will just be:

1. Install `motion`
2. Update imports
3. Uninstall `framer-motion`

REF: https://motion.dev/docs

## ✅ Acceptance Criteria

- [ ] Migrated from `framer-motion` to `motion`
- [ ] Animations still function

esizer:
🤖 Resolves #12352 

## 👋 Introduction

Moves from the old name `framer-motion` to the new independent `motion` library

## 🕵️ Details

This is the same library, it has just been decoupled from framer to be an independent library.

## 🧪 Testing

1. Install new deps `pnpm i`
2. Build app `pnpm run dev:fresh`
3. Confirm animations still run as expected


pnpm-lock.yaml
```diff
@@ -13306,6 +13320,14 @@ snapshots:
 
   motion-utils@11.14.3: {}
 
+  motion@11.15.0(react-dom@18.3.1(react@18.3.1))(react@18.3.1):
+    dependencies:
+      framer-motion: 11.15.0(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
```
mnigh, line 13325: Maybe I'm not reading this correctly, but why does `motion@11.15.0` still have a dependency of framer-motion 11.15.0?

mnigh (commented):
A question about dependency, otherwise looks good to me.

pnpm-lock.yaml
esizer: My hunch was correct, looks like it is just a re-export of `framer-motion` at the moment.

https://github.com/motiondivision/motion/blob/ecd97f7dce8954be300aa73ab6a96208437941c5/packages/motion/src/react.ts#L3

mnigh: Thanks for checking.

mnigh (approved):
LGTM 👍

PR #12392, opened 2024-12-30
[Debt] Split up e2e tests

### LINKED ISSUE #12357: ♻️ Address slow playwright tests
## ♻️ Debt/Refactor

We should probably address some of the slower tests in playwright.

## 🕵️ Details

As our test coverage expands (good thing) our run time is growing (bad thing). I think we may be doing too much in single test files. Since playwright runs our tests in parallel, writing smaller tests (with the same coverage) should help speed things up.

So, instead of trying to test a complete feature in a single test, we can still test the feature but break steps up into different tests.

## 🙋‍♀️ Proposed Solution

Look at the following slow tests being reported and see if breaking them up helps.

```sh
  Slow test file: [chromium] › applicant/experience.spec.ts (2.5m)
  Slow test file: [chromium] › admin/auth.spec.ts (2.3m)
  Slow test file: [chromium] › admin/pool-candidate.spec.ts (1.1m)
  Slow test file: [chromium] › admin/process-actions.spec.ts (59.9s)
  Slow test file: [chromium] › admin/admin-workflows.spec.ts (58.5s)
  Consider splitting slow test files to speed up parallel execution
```
### Experiences

Right now we test every experience type in a single test file. I think this one should be easy, just create different test files for each experience type.

### Auth

Similar with this one, we are testing every single role in a single test. We can create a test per role.

### Pool candidate

I think this one is the most disjointed. Each test, while related to the candidate, are not that similar. We could have a test for:
 - snapshot
 - assessment
 - "more actions"

### Process actions

This one is tricky because it is all related but, maybe we have separate files for the different crud operations? :thinking: 

 - create
 - update
 - delete

### Admin workflows

I feel like this one is easiest, move the download test out to its own file.

## ✅ Acceptance Criteria

- [ ] Long running tests are broken up
- [ ] Coverage remains the same (or higher :stuck_out_tongue: )
- [ ] Improvement on run time of playwright tests

esizer:
🤖 Resolves #12357 

## 👋 Introduction

Splits up slow playwright tests to hopefully better leverage shards.

## 🧪 Testing

1. Confirm tests pass
2. Confirm no loss in coverage


vd1992 (approved):
Seems to look good to me!

Still see some slow test logging but some of those tests really can't be split without becoming ugly so oh well. Flakiness in the CI and local, but that is nothing new 😛

mnigh (approved):
LGTM 👍

PR #12393, opened 2024-12-30
[Fix]try wait if the token is still refreshing

brindasasi:
🤖 Resolves


mnigh:
@brindasasi can this PR be closed?

PR #12411, opened 2025-01-02
[Feat] Community interest backend

### LINKED ISSUE #12253: ✨Backend model for Community Interest
## ✨ Feature

On the applicant/employee dashboard, government employees will be able to express interest in a community. When they do so, we collect some additional information about where there interests lie.

## 🕵️ Details

```
#I guess this is effectively a pivot table for user-community with a little extra info, we could rename it to UserCommunity or similar
type CommunityInterest {
  id: UUID!
  community: Community # belongsTo
  workstreamInterest: [WorkStream]  #workstreams from this community only
  
  jobInterest: Boolean
  trainingInterest: Boolean

  additionalInformation: String #input by user

  #developmentProgramInterest: [DevelopmentProgramInterest] # This doesn't exist yet, will be added in 12099
}

type EmployeeProfile {
  ...
  interestInCommunities: [CommunityInterest] @hasMany
}
```
**Notes:**
- Pivot table for development programs described in: #12099 
- EmployeeProfile added in #12001 
- In the database, each CommunityInterest item belongs to a user_id. EmployeeProfile is backed by the users table.

## 🎨 Design

https://www.figma.com/design/H1loB8WxGP6vyJTrVqcafd/Functional-communities-(Employees)?node-id=5-9&node-type=canvas&t=51PBxXT2GvAXqEZd-0

## 📸 Screenshot

Image

In terms of mutations, it may be easiest to add attach/detach/sync fields to EmployeeProfile.interestInCommunities the field.
Similarly, the CommunityInterest input should support attach/detach/sync operations for work streams.

Alternatively, we could add create and update mutations for CommunityInterest, where the input includes a userId or employeeProfileId. In this case the mutations should check the `update-own-employeeProfile` permission.

### Validation:
Work Streams attached to a CommunityInterest belong to the correct community.
For simplicity, we should probably not allow changing the community of a CommunityInterest entity. You'd have to create a new one.
A user should not be able to create two CommunityInterests with the same Community, or add the same stream multiple times to one CommunityInterest (these should _also_ be validated at database level, but we still want Laravel validation to return better error messages). 

## ✅ Acceptance Criteria

- [ ] Migration defines a community_interests table, with user_id and community_id foreign keys, and a pivot table with work streams. Constraints against the same user_id/community_id combo, or against work streams being added to to one Interest
- [ ] Laravel model for CommunityInterest, with relationships and factory and seeder
- [ ] CommunityInterest added to schema. It's queryable via EmployeeProfile.
- [ ] An employee can use a mutation to add or update their CommunityInterests. Validation ensures that work streams match the proper Community, and CommunityInterests and WorkStreams aren't duplicated.
- [ ] phpunit test for that validation

## 🛑 Blockers
```[tasklist]
### Blocked By
- [ ] #12001 
```

esizer:
🤖 Resolves #12253 

## 👋 Introduction

Adds the community interest model to the server along with supporting code.

## 🕵️ Details

Based on the UI, we will be adding these in their own forms so I went with their own mutations instead of a field on the employee profile. It seemed like it made sense to me, especially when it came to validating the input.

Though, I could be convinced to move it to a field on the `UpdateEmployeeProfileInput` as well.

## 🧪 Testing

1. Refresh the API `make refresh-api`
2. Confirm migrations runs as expected
3. Refresh the database `make seed-fresh`
4. Confirm no errors and expected data gets seeded into new tables
5. Test creating and updating community interests
6. Confirm test coverage is good (tried to cover edge cases but was not sure)


api/database/migrations/2025_01_02_144403_create_community_interest_table.php
```diff
@@ -0,0 +1,48 @@
...
+use Illuminate\Database\Query\Expression;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
```
petertgiles, line 13: Could you add some uniqueness constraints to these tables?  To prevent duplicates of user->community and interest->stream.

api/graphql/schema.graphql
```diff
@@ -1466,6 +1477,23 @@ input UpdateEmployeeProfileInput @validator {
...
+  userId: UUID! @rename(attribute: "user_id")
+  community: CommunityBelongsTo!
+  workStreams: WorkStreamBelongsToMany
+  jobInterest: Boolean @rename(attribute: "job_interest")
+  trainingInterest: Boolean @rename(attribute: "training_interest")
+  additionalInformation: String @rename(attribute: "additional_information")
+}
+
+input UpdateCommunityInterestInput @validator {
+  community: CommunityBelongsTo
```
petertgiles, line 1490: I'm not sure it makes sense to be able to change the community of an existing CommunityInterest model?

petertgiles: Especially if the set of workstreams is different on the new community.

petertgiles (changes requested):
Looks and works great!  I flagged a few small things to check on.

api/tests/Feature/CommunityInterestTest.php
```diff
@@ -0,0 +1,228 @@
...
+                        'community' => ['connect' => $this->communityId],
+                        'workStreams' => ['sync' => $this->workStreamIds],
+                    ],
+                ])
+            ->assertJson([
+                'data' => [
+                    'createCommunityInterest' => [
+                        ...$this->input,
+                        'community' => ['id' => $this->communityId],
+                        'workStreams' => Arr::map($this->workStreamIds, fn ($id) => ['id' => $id]),
```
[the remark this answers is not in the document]
esizer: oh yeah, that makes sense. I have updated it to just use the first one to avoid any ordering.

[improve create test stability](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12411/commits/d6b88bc801c3ec8c5a8fb1f1e9115e197323ddfd)

petertgiles (approved):

PR #12418, opened 2025-01-03
[Fix] Add token validation to didAuthError and allowable clock skew to willAuthError

### LINKED ISSUE #12369: 🐛 Do multiple tab interactions lead to early log out?
## 🐛 Bug

We have reports from users, including our own team. that SiC has been logging users out far earlier than we'd expect. 

We have some early hints that _maybe_ this happens when you're working in multiple tabs at the same time. Investigate this further, and fix if possible.

## 🦋 Expected Behaviour

Users sign-in session is expected to end after 4 hours.

## 📋 Steps to Reproduce

We're figuring that out.

## ✅ Acceptance Criteria

- [ ] Fewer unexpected logouts on Prod

petertgiles:
🤖 Resolves #12369 

## 👋 Introduction

We've been getting lots of reports of premature logouts with our app.  Our best explanation is that the Surface tablets appear to have over 5 seconds of clock skew which means that the client is attempting to use an expired token.  Also, URQL is not configured to recognize the token_validation error and so it does not attempt to refresh when it gets the error.  This PR attempts to fix both problems.

### DidAuthError

1. Rebuild and log in to app
2. Add "throw new Exception('Failed to get introspection');" to the top of `verifyJwtWithIntrospection` in api/app/Services/OpenIdBearerTokenService.php and to simulate an expired token.
3. Make an API call in your browser

- [ ] URQL will try to refresh ONCE and then force a log out.

### WillAuthError

1. Manually change allowableClockSkewSeconds in packages/client/src/constants.ts to something very large, like 8 * 60.
2. Rebuild and log into app.
3. Wait until you enter the skew zone, like 2 minutes and make an api call.

- [ ] URQL will try to refresh ONCE and and be successful

## 📸 Screenshot

image

### DIFF
--- a/packages/client/src/components/ClientProvider/ClientProvider.tsx
+++ b/packages/client/src/components/ClientProvider/ClientProvider.tsx
@@ -148,7 +148,9 @@ const ClientProvider = ({
                 const didError = res
                   ? res.status === 401 ||
                     error.graphQLErrors.some(
-                      (e) => e.extensions?.category === "authentication",
+                      (e) =>
+                        e.extensions?.category === "authentication" ||
+                        e.extensions?.reason === "token_validation", // the auth provider says the token is invalid - maybe just a refresh is needed
                     )
                   : false;
 
--- a/packages/client/src/components/ClientProvider/ClientProvider.tsx
+++ b/packages/client/src/components/ClientProvider/ClientProvider.tsx
@@ -33,18 +33,21 @@ import {
 } from "../../utils/errors";
 import specialErrorExchange from "../../exchanges/specialErrorExchange";
 import protectedEndpointExchange from "../../exchanges/protectedEndpointExchange";
-import { apiHost, apiUri } from "../../constants";
+import { allowableClockSkewSeconds, apiHost, apiUri } from "../../constants";
 
-const isTokenKnownToBeExpired = (accessToken: string | null): boolean => {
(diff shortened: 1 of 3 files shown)


packages/client/src/components/ClientProvider/ClientProvider.tsx
```diff
@@ -33,18 +33,21 @@ import {
...
-  let tokenIsKnownToBeExpired = false;
+const isTokenProbablyExpired = (accessToken: string | null): boolean => {
+  let tokenProbablyExpired = false;
   if (accessToken) {
     const decoded = jwtDecode<JwtPayload>(accessToken);
     if (decoded.exp) {
-      tokenIsKnownToBeExpired = Date.now() > decoded.exp * 1000; // JWT expiry date in seconds, not milliseconds
+      const tokenExpiryDateSeconds = decoded.exp;
+      const safeTokenExpiryDateSeconds =
+        tokenExpiryDateSeconds - allowableClockSkewSeconds; // allow for the client's machine to be a bit off
```
brindasasi, line 45: yeah .. this is what I did as well .. but I didn't see the refresh happening earlier than 10 mins .. may be let's try yours

brindasasi (approved):
This is looking like great to me and I didn't see any log out in UAT. I'm up for trying this out in prod tonight

PR #12427, opened 2025-01-07
[Feature] Autocomplete experience orgs

### LINKED ISSUE #12323: ✨ Autocomplete for experience orgs
## ✨ Feature

Update the experience editor to autocomplete "org fields" with a list of all the "orgs" from the user's experiences.

## 🕵️ Details

Work => Organization
Education => Institution
Community => Group/Organization/Community
Personal => none (there may be an equivalent Personal field coming later in (#12329)
Award => Issuing organization

Use a `datalist` element populated from a database query of the user's experiences.

https://stackoverflow.com/a/45551818

## ✅ Acceptance Criteria

- [ ] When adding a new experience of any type, the "org" field has autocomplete suggestions from all the user's previously saved orgs, of any type
- [ ] When editing an experience, the same autocomplete behaviour occurs
- [ ] When adding or updating an experience, the new org is immediately available to the next use of the form.  No app reboot is needed. 

## 🛑 Blockers

```[tasklist]
### Blocked By
```

vd1992:
🤖 Resolves #12323

## 👋 Introduction

Populate with suggestions, using `<datalist />`, orgs for all experiences except personal by pulling names from all experiences except personal.
There are two contexts for experiences, through timeline and within an application. A util function for reuse was added as well due to more than one context

## 🧪 Testing

1. Login and have experiences
2. Create/edit experiences and observe whether suggestions were populated for org name

## 📸 Screenshot

image

image

### DIFF
--- a/apps/playwright/fixtures/ExperiencePage.ts
+++ b/apps/playwright/fixtures/ExperiencePage.ts
@@ -62,7 +62,7 @@ class ExperiencePage extends AppPage {
       .click();
 
     await this.page
-      .getByRole("combobox", { name: /organization/i })
+      .getByLabel("Organization *", { exact: true })
       .fill(input.organization ?? "test org");
 
     await this.page
@@ -539,7 +539,7 @@ class ExperiencePage extends AppPage {
       .fill(input.title ?? "test role");
 
     await this.page
-      .getByRole("combobox", { name: /organization/i })
+      .getByLabel("Group / Organization / Community *", { exact: true })
       .fill(input?.organization ?? "test org");
 
     await this.page
@@ -577,7 +577,7 @@ class ExperiencePage extends AppPage {
       .selectOption({ label: "Me" });
 
     await this.page
-      .getByRole("combobox", { name: /organization/i })
+      .getByLabel("Issuing organization *", { exact: true })
       .fill(input?.issuedBy ?? "test org");
 
     await this.page
@@ -607,7 +607,7 @@ class ExperiencePage extends AppPage {
       .fill(input?.areaOfStudy ?? "test area of study");
 
     await this.page
-      .getByRole("combobox", { name: /institution/i })
+      .getByLabel("Institution *", { exact: true })
       .fill(input?.areaOfStudy ?? "test institution");
 
     await this.page
--- a/apps/playwright/fixtures/ExperiencePage.ts
+++ b/apps/playwright/fixtures/ExperiencePage.ts
@@ -62,7 +62,7 @@ class ExperiencePage extends AppPage {
       .click();
 
     await this.page
-      .getByRole("textbox", { name: /organization/i })
+      .getByRole("combobox", { name: /organization/i })
       .fill(input.organization ?? "test org");
 
     await this.page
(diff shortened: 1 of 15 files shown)


apps/playwright/fixtures/ExperiencePage.ts
vd1992, line 1: Seems `<datalist />` turns textboxes into comboboxes

🤷

apps/web/src/components/ExperienceFormFields/AwardFields.tsx
```diff
@@ -80,7 +83,15 @@ const AwardFields = ({ labels }: SubExperienceFormProps) => {
             name="issuedBy"
             type="text"
             rules={{ required: intl.formatMessage(errorMessages.required) }}
+            list="organizationSuggestions"
```
esizer, line 86: For each of these, I'd probably conditionally add this attribute only when there are suggestions.

apps/web/src/utils/experienceUtils.tsx
```diff
@@ -913,3 +914,37 @@ export const useExperienceInfo: UseExperienceInfo = (experience) => {
...
+      return exp.organization;
+    }
+    return undefined;
+  });
+  const organizationsForAutocompleteFiltered: string[] = unpackMaybes(
+    organizationsForAutocomplete,
+  );
+
+  return organizationsForAutocompleteFiltered;
+};
```
esizer, line 950: I think we will need to ensure the list returned is unique. If I have experiences with the same org, they show up multiple times in the list.

esizer (changes requested):
Looking good so far, just two minor suggestions.

apps/web/src/components/ExperienceFormFields/AwardFields.tsx
vd1992: Aye aye, done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/12427/commits/b90a9f7a8f30d7405976506ffcbfe14685358d8b
I noticed this makes the input type variable depending on whether `list` is defined, which can cause testing bugs https://github.com/GCTC-NTGC/gc-digital-talent/pull/12427/commits/36fc31fd0ce7f818922d271d0d4156d19c6eae41

apps/web/src/utils/experienceUtils.tsx
vd1992: Oh! Good spot!
Done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/12427/commits/de2f55124bd967bf884efe94cc88afcc0dc2e0a0

apps/web/src/components/ExperienceFormFields/AwardFields.tsx
esizer: Yeah, I saw that. Makes me wonder if we should create a specific input for this. Obviously out of scope for this issue though.

Maybe once we get this merged in we can sit and stew over it for a while and see if it makes sense :woman_shrugging:

esizer (approved):
LGTM! :+1:

PR #12469, opened 2025-01-10
[Fix] Date formatting in local timezone

### LINKED ISSUE #12312: 🐛 In-app notification date is wrong
## 🐛 Bug

Below is the notification I got on Dec.15 Sunday 10 pm and it shows up as Dec. 16 on the UI 

Image

## 🦋 Expected Behaviour

In-app notification date should be the receiving person's localized date 

## 🕵️ Details

I think this is the same happening in files download as well 

## 📋 Steps to Reproduce

1.  Create a pool closing in 3 days and start a application for it and leave it draft 
2. Run notification in the local 
3. Observe the in-app notification shows up with tmo's date 

## ✅ Acceptance Criteria

- [ ] The timestamp date shown for all in-app notifications should be converted from UTC to local timezone for display.
  - This should be the case for all types of In-app notification: Status Changed , New pool , Pool closing , file downloads  etc

### LINKED ISSUE #12458: 🐛 Showing UTC timezone in "Applied On:" span instead of users local time
## 🐛 Bug
This happens in the main Dashboard where applications in progress are listed. I did not notice if it also happens in "Show Past Applications" section.

## 🦋 Expected Behaviour

It should show the users application datetime in his timezone

## 🕵️ Details

N/A

## 📸 Screenshot

image

## AC
- [ ] Submitted times show in local time zone on application cards

esizer:
🤖 Resolves #12312 , resolves #12458 

## 👋 Introduction

Updates our date format function to use the local timezone.

## 🕵️ Details

Our date formatter was only formatting in a specific timezone instead of UTC when it was specified. The new method is not 100% but should work in +96% of browsers.

Now, we use the `Intl` api to get the users timezone and format in that timezone if not provided. We will then fallback to UTC if they do not support `Intl` (maybe we can do a better fallback? :thinking: )

## 🧪 Testing

1. Build `pnpm run dev:fresh`
2. Login as any user
3. Force some notification to be created
4. Change the notification created time to less than 5 hours after midnight in UTC (or what ever value would be the previous day in your timezone)
5. Veuiw the notification and confirm the date is in local time (the date is not tomorrow)
6. Submit an appllication
7. Navigate to applicant dashboard `/applicant`
8. Confirm the "Applied on" date is the correct local datetime


petertgiles:
Oh wow, this PR scares me.  :laughing:   What worries me is you're changing the default behaviour of a function that is used fairly widely throughout the app.  We would need to check all those uses to make sure nothing else was broken by this change.

Doesn't the format function already display dates in the local time zone by default?   Isn't the bug that the source page isn't using `parseDateTimeUtc` when receiving the data and is interpreting a UTC date as a local one?

As a side note, there's a slightly different method of getting the time zone in apps/web/src/pages/Pools/PoolAdvertisementPage/components/DeadlineDialog.tsxL34.  I have no idea which is better but it might be worth standardizing the logic in a helper function.

esizer:
> Doesn't the format function already display dates in the local time zone by default

Apparently not, it formats it in the time specified by the date (UTC). Both of the functions where we noticed the issue are using `parseDateTimeUTC` which does not transform the timezone so it is a UTC date that we are formatting.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/1a7f0a6fc7342886b4134fa314476402000fe3fc/apps/web/src/utils/poolCandidate.ts#L110

https://github.com/GCTC-NTGC/gc-digital-talent/blob/1a7f0a6fc7342886b4134fa314476402000fe3fc/apps/web/src/components/NotificationList/NotificationItem.tsx#L148

https://github.com/GCTC-NTGC/gc-digital-talent/blob/1a7f0a6fc7342886b4134fa314476402000fe3fc/packages/date-helpers/src/index.ts#L156-L158

petertgiles:
OK, here's my suggestion.  Let me know if you want to talk through it or discuss in standup.
1. Split into two functions with very clear names.  Like `formatDateInLocalZone` and `formatDateInSpecificZone`
2. Change all references to `formatDate` to one of those two.  There are 27 right now, so tedious but not too bad.
3. Add a Jest test for each of the two format functions.
  a. You might have to do some clever mocking of window.Intl to make the local one work.

packages/date-helpers/src/index.ts
```diff
@@ -34,6 +34,18 @@ export {
...
+  let localTz;
+  if (window.Intl && typeof window.Intl === "object") {
+    localTz = Intl.DateTimeFormat().resolvedOptions().timeZone;
+  }
+
+  return localTz;
+};
+
 /**
  * Format a date in given format and locale, optionally in a different time zone
```
petertgiles, line 50: Could you add a comment to make it clearer what the default is when no zone is provided?  Is the default always the local time zone or is it more nunanced than that?  I think it is, given the changes you made to parseDateTimeUtc.

apps/web/src/pages/Pools/PoolAdvertisementPage/components/DeadlineDialog.tsx
```diff
@@ -30,8 +30,7 @@ interface DeadlineDialogProps {
 const DeadlineDialog = ({ deadline }: DeadlineDialogProps) => {
   const intl = useIntl();
   const routes = useRoutes();
-  // https://stackoverflow.com/a/34602679
-  const localTimeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+  const localTimeZone = getLocalTimezone();
```
petertgiles, line 33: This isn't needed anymore, right?  Probably the other use of this helper function isn't either?

petertgiles (commented):
Overall, I like this approach, I think.  I left a few small comments.

apps/web/src/pages/Pools/PoolAdvertisementPage/components/DeadlineDialog.tsx
esizer: Yeah, rolled back in [remove local timezone helper](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12469/commits/2e43b02a4eb6e06d63ebff954da6011725fc6b7e)

packages/date-helpers/src/index.ts
esizer: It is very slightly more nuanced. I tried to explain it in [improve documentation for timezone behaviour](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12469/commits/50258bfaa9ba7aa710788f57947aebde7540ca3d)

Does that make sense? :sweat_smile:

petertgiles: That's better, yeah.  I wish the function didn't have a "it depends" behaviour but it is what is is.  :laughing:

esizer: haha yeah. We could if instead of dates we just used the timestamp though!

vd1992 (changes requested):
The UTC rendering issue was present in many other places, like looking at submitted date for a pool candidate in tables or the received date for a talent search request yikes. I guess with randomly seeded data you don't notice if the times look off 🤣 

I'd like to request the addition of just-in-case testing to `packages/date-helpers/src/date.test.ts`
Sample test cases for example would be input with a timezone, and without.

As an aside, I am worried people may have been confused by the timezone mix-up and were thus erroneously informed at some point in time or place

esizer:
> I'd like to request the addition of just-in-case testing to `packages/date-helpers/src/date.test.ts` Sample test cases for example would be input with a timezone, and without.

Hmm, we do have [pretty extensive tests](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/packages/date-helpers/src/date.test.ts) and I'm not really sure where the gap exists :thinking:

petertgiles (approved):
Nice, I think this is working well now.

petertgiles:
> Hmm, we do have [pretty extensive tests](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/packages/date-helpers/src/date.test.ts) and I'm not really sure where the gap exists 🤔

I wouldn't mind seeing some tests on `formatDate`.  I don't think we have any right now.
1) A DateTime passed in, no zone parameter,  that renders in the local time zone.  (Might be tricky to mock the local time zone)
2) A DateTime passed in, with zone parameter, that renders in that time zone.

esizer:
> > Hmm, we do have [pretty extensive tests](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/packages/date-helpers/src/date.test.ts) and I'm not really sure where the gap exists 🤔
> 
> I wouldn't mind seeing some tests on `formatDate`. I don't think we have any right now.
> 
> 1. A DateTime passed in, no zone parameter,  that renders in the local time zone.  (Might be tricky to mock the local time zone)
> 2. A DateTime passed in, with zone parameter, that renders in that time zone.

Okay @vd1992 and @petertgiles I got some tests in that should help out a bit:

 - [add timezone based format tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12469/commits/1e76582c8b7409014bae4d11d473f33c34a219db)
 - [add test same timezone when provided](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12469/commits/f66576f3d632b79d3e6c4eeaeaae048f567d63fb)

packages/date-helpers/src/date.test.ts
```diff
@@ -144,3 +147,49 @@ describe("parse DateTime UTC to native Date tests", () => {
...
+      date: parseDateTimeUtc("2022-01-01 00:00:00+00:00"),
+      formatString: DATETIME_FORMAT_STRING,
+      intl,
+    });
+
+    expect(actual).toBe("2021-12-31 19:00:00");
+  });
+
+  test("it formats with timzone Eastern provide", () => {
+    // NOTE: This is odd but +5 is actually -5
```
vd1992, line 173: Why

vd1992 (approved):
Excellent, thank you for adding some more testing. I thought it was imperative that `parseDateTimeUtc` doesn't sneak any surprises by us

PR #12472, opened 2025-01-10
[Bug] Resolve toast spam in create application path when an application already exists

### LINKED ISSUE #12314: 🐛 Multiple toast messages for Continue Application
## 🐛 Bug
Multiple toast messages for Continue Application  
Image

## 🦋 Expected Behaviour

Only one toast message is shown for Continue Application action 
## 📋 Steps to Reproduce

1. Create a draft for any opportunity 
2. Go back to browse pools page and click on the same opportunity 
3. Click on continue application 
4. Observe your page is full of above notification 
## ✅ Acceptance Criteria

- [ ] Continue Application toast message is shown only once 
- [ ] Confirm Application created toast message is also shown only once

vd1992:
🤖 Resolves #12314

## 👋 Introduction

As the title states, fix that annoyance. 
Followed the suggestion to handle the toast spam rather than try and diagnose why duplicate application creation was happening

## 🧪 Testing

1. Navigate to some process `/browse/pools/:id`
2. Apply
3. Apply again by visiting directly `/browse/pools/:id/create-application`
4. Observe toasts

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/pages/CreateApplicationPage/CreateApplicationPage.tsx
+++ b/apps/web/src/pages/CreateApplicationPage/CreateApplicationPage.tsx
@@ -85,10 +85,10 @@ const CreateApplication = () => {
     path: string,
     toastFunction: () => void,
   ): Promise<void> => {
-    if (navigateWithToastCounter.current > 0) return; // we've already started navigation
+    navigateWithToastCounter.current += 1;
+    if (navigateWithToastCounter.current > 1) return; // we've already started navigation
     await navigate(path, { replace: true });
     toastFunction();
-    navigateWithToastCounter.current += 1;
   };
 
   // If a "me" object came back then we've checked.


apps/web/src/pages/CreateApplicationPage/CreateApplicationPage.tsx
```diff
@@ -85,10 +85,10 @@ const CreateApplication = () => {
     path: string,
     toastFunction: () => void,
   ): Promise<void> => {
-    if (navigateWithToastCounter.current > 0) return; // we've already started navigation
+    navigateWithToastCounter.current += 1;
+    if (navigateWithToastCounter.current > 1) return; // we've already started navigation
     await navigate(path, { replace: true });
     toastFunction();
-    navigateWithToastCounter.current += 1;
   };
```
vd1992, line 93: Not entirely sure why or when this started, but it appeared this function in question kept firing many, many times. 
The counter incrementing after the async function might've caused the function to fire many times before the current actually incremented. Moving the counter up ensures it readily and immediately updates and doesn't get held up by the other functions.

petertgiles: Yeah, good find.  You can find the diff in the "Upgrade to React Router 7" PR.  As you said, it's that an `await` was added to the navigation so the execution never got to the counter increment.
image

PR #12532, opened 2025-01-20
[Feat] Log role assignment activity

### LINKED ISSUE #12491: ✨ Better audit trail for role assignments
## ✨ Feature

We need to generate a much better audit trail for role assignments.  We need to be able to answer questions about who granted a particular role assignment and when.

- add timestamps (created_at, updated_at) to the role_user table
  - ~maybe updated_by?~ I think its sufficient to rely on LogsActivity
- add the LogsActivity trait to the RoleAssignment model

## ✅ Acceptance Criteria

- [ ] We can find out who created or updated or deleted a role assignment and when

## 🛑 Blockers

```[tasklist]
### Blocked By
```

esizer:
🤖 Resolves #12491 

## 👋 Introduction

Adds timestamps and activity log to role assignments.

## 🕵️ Details

This is a little more involved because of traits coming from Laratrust.

- Had to extend the `roles` relationship that comes from the trait
- Added custom logs on the `roleAdded` and `roleRemoved` model events since the default made it difficult to determine what actually happened.

The logged properties are the ID of role and team (if present) removed or added. Hopefully this makes sense, if it does not, this is not too difficult to change.

## 🧪 Testing

3. Navigate to edit a user `/admin/users/{userId}/edit`
4. Add both a normal and team based role
5. Confirm the new records in the database for `role_user` table have the appropriate timestamps
6. Remove both roles added in step 4
7. Confirm the activity log contains `roleAdded` and `roleRemoved` activities with the expected properties

### DIFF
--- a/api/app/Models/RoleAssignment.php
+++ b/api/app/Models/RoleAssignment.php
@@ -23,7 +23,7 @@ class RoleAssignment extends Model
 
     protected $keyType = 'string';
 
-    public $timestamps = false;
+    public $timestamps = true;
 
     protected $fillable = [
         'role_id',
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -497,7 +497,7 @@ class User extends Model implements Authenticatable, HasLocalePreference, Laratr
      * @param  string  $eventName  The action taken `roleAdded` | `roleRemoved`
      * @param  User  $user  User being affected (subject)
      * @param  ?string  $roleId  Id of the role being added or removed
-     * @param  ?string|array  $team  IF team based role, the ID or array of then team ID
+     * @param  mixed  $team  IF team based role, the ID or array of then team ID
      */
     private static function logRoleChange(string $eventName, User $user, ?string $roleId, mixed $team)
     {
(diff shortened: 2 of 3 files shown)


petertgiles (commented):
Thanks for picking this up so quickly!  It works well but I'm not sure about the custom logging.  It is easy to read but it logs activity differently from the logging of the other models which is confusing in its own way.  I'd expect to see `create` and `update` rows for `Model\RoleAssignment` with role_id, user_id, and team_id inside of an attributes property.  Is it possible to make it more consistent?

esizer:
> Thanks for picking this up so quickly! It works well but I'm not sure about the custom logging. It is easy to read but it logs activity differently from the logging of the other models which is confusing in its own way. I'd expect to see `create` and `update` rows for `Model\RoleAssignment` with role_id, user_id, and team_id inside of an attributes property. Is it possible to make it more consistent?

Okay, did my best to get it looking more like the other logs based on your feedback.

[make shape of log more consistent](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12532/commits/54a2aeb81b09375b9b62f357ee8dc7027e780c95)

petertgiles (approved):
Awesome, thanks!!

Issue #12212, opened 2024-12-05
✨ Allow users to opt back into talent searches after they are hired


gobyrne:
## ✨ User story

As an employee that was hired through the platform, I want to start showing up in talent searches again, so I can move to a new job if the one I'm in isn't ideal.

## 🕵️ Details

In the new applicant dashboard dialogues #12214, we present UI for hired candidates to opt back in for being referred. This doesn't work yet though! So... we should probably fix that.

Right now only candidates with Qualified-Available and Placed-Casual (and Placed-Tentatative) statuses will show up in searches. We want to allow for more flexibility when candidates are "placed".

In this new model, we will "suspend" candidates when they get placed (unless it's placed-casual), but we continue to allow "Placed" candidates to show up in searches.

### Update the search
Currently Qualified available, Placed Tentative and Placed Casual show up in searches. To this we need to add:
- Placed term
- Placed indeterminate

Of course, anyone with a "suspendedAt" date, will not show up in talent searches.

### Suspend candidates when placed
Add new functionality:
When an admin sets Placed term or Placed indeterminate for a candidate - set the "suspendedAt" date (as well as the placedAt date).  Updating to ANY OTHER status sets suspendedAt date to null.

### Migration to make this work for all current users
To make this work for existing applications, we need to "suspend" all candidates that are currently Placed term, or Placed indeterminate

- Write a artisan command to set "suspendedAt"="placedAt" for all Placed term and Placed indeterminate candidates. Write this as a SQL command or Eloquent update command, not a loop of code, for efficiency. Ensure its wrapped in a transaction.  Run this at deployment.
- Create a ticket to delete your artisan command

### Update PoolCandidate.scopeAvailable
Currently Qualified available, Placed Tentative and Placed Casual show up in searches. To this we need to add:
- Placed term
- Placed indeterminate

(Probably update the qualifiedEquivalentGroup?)

And filter out anyone with "suspendedAt" date.

## ✅ Acceptance Criteria

- [ ] Search updated to show all "Placed" candidates (unless they are suspended)
- [ ] New functionality to suspend candidates when an admin sets them to Placed-term or Placed-indeterminate
- [ ] Artisan command to suspend currently Placed-term and Placed-indeterminate candidates
  - [ ] create ticket to delete command in the future
- [ ] phpunit feature tests that updating a user sets suspendedAt correctly.
[...]

Issue #12331, opened 2024-12-17
♻️ Delete Artisan command `UnsuspendPlacedCandidates`


vd1992:
## ♻️ Debt/Refactor

Delete the command once no longer needed

Likely safe to delete once it has been run, and waited a week or so to make sure nothing came up as an issue.

## ✅ Acceptance Criteria

- [ ] File is gone gone

## 🛑 Blockers

```[tasklist]
### Blocked By
```

petertgiles:
Could you please clarify "once no longer needed".  When is it no longer needed?

"File is gone gone".  What is the file name?

PR #12533, opened 2025-01-20
[Bug] Add `unsuspend-placed-candidates` command

### LINKED ISSUE #12479: 🐛 All placed candidates are not visible in any candidate table
## 🐛 Bug

Placed candidates are not visible in any candidates table ( All candidates , Candidates per pool ) 

Image

Image

Image

## 🕵️ Details

We ran a script to migrate all placed candidates into suspended status in order to support opt in feature. 
But we forgot that all candidate screens have a default filter for active status and any suspended candidate won't show up by default. 
https://github.com/GCTC-NTGC/gc-digital-talent/pull/12330

## 📋 Steps to Reproduce

1. Fresh seed the database in local 
2. Check you got some candidates with Placed Indeterminate or Placed Term 
3. Update any placed candidate into suspended status or run the `php artisan app:suspend-placed-candidates`
4. Open All candidates screen and look for the candidates that you had in step 2 
5. Observe they are missing
6. Now open the filter and filter by Suspended candidates 
7. observe the candidates are shown that you had in step 2 

## 🙋‍♀️ Proposed Solution

We've undone the change that automatically suspends candidates when they are placed (in #12486). Now let's also undo the migration. Create and run a new artisan script, `unsuspend-placed-candidates`.

The migration should use eloquent and go through the PoolCandidate model so that changes are logged. (Even though its slower than a mass update db query.)

Because it will be slow, should be run outside of a normal deployment.

## ✅ Acceptance Criteria

- [ ] Script can be run which unsuspends all placed candidates.
- [ ] Script should use eloquent and go through the PoolCandidate model so that changes are logged.

## 🚀 Deployment

Run `php artisan app:unsuspend-placed-candidates`
Run this outside of normal deployment process, because it may be slow, and its not strictly necessary for a proper deployement.

vd1992:
🤖 Resolves #12479

## 👋 Introduction

Adds a command for the purpose of reversing what was done database-wise in https://github.com/GCTC-NTGC/gc-digital-talent/issues/12212
Effectively the same command just reversed

## 🧪 Testing

1. Set candidates with the relevant status as suspended
2. Run command below
3. Check it worked

## 🚚 Deployment

Can be run before or after deployment, so at anytime

Run

`php artisan app:unsuspend-placed-candidates`

### DIFF
--- a/api/app/Console/Commands/UnsuspendPlacedCandidates.php
+++ b/api/app/Console/Commands/UnsuspendPlacedCandidates.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Enums\PoolCandidateStatus;
+use App\Models\PoolCandidate;
+use Illuminate\Console\Command;
+use Illuminate\Database\Eloquent\Collection;
+
+class UnsuspendPlacedCandidates extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'app:unsuspend-placed-candidates';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Set suspended at date to null for placed term and indeterminate candidates';
+
+    /**
+     * Execute the console command.
+     *
+     * @return int
+     */
+    public function handle()
+    {
+        $applicableStatuses = [PoolCandidateStatus::PLACED_TERM->name, PoolCandidateStatus::PLACED_INDETERMINATE->name];
+
+        PoolCandidate::whereIn('pool_candidate_status', $applicableStatuses)
+            ->whereNotNull('suspended_at')
+            ->with('user')
+            ->chunkById(100, function (Collection $candidates) {
+                foreach ($candidates as $candidate) {
+                    /** @var \App\Models\PoolCandidate $candidate */
+                    $candidate->suspended_at = null;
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
Nice, looks great.   OK to merge, but three suggestions:
- Could you add a counter so it outputs how many rows were changed?  Right now there's no feedback to the user at all that anything happened.
- Could you add a note to the deployment instructions that this can be run any time, before or after deployment.
- Could you add a small cleanup issue to the board to remove this command along with a reminder on _when_ it can be removed?

vd1992:
> Nice, looks great. OK to merge, but three suggestions:
> 
>     * Could you add a counter so it outputs how many rows were changed?  Right now there's no feedback to the user at all that anything happened.
> 
>     * Could you add a note to the deployment instructions that this can be run any time, before or after deployment.
> 
>     * Could you add a small cleanup issue to the board to remove this command along with a reminder on _when_ it can be removed?

1) Counter added in https://github.com/GCTC-NTGC/gc-digital-talent/pull/12533/commits/d29dd9a9ce32a9a748b6b4d1eae31125fe70b2c1

2) PR instructions updated

3) Reused existing issue that was no longer applicable https://github.com/GCTC-NTGC/gc-digital-talent/issues/12331

petertgiles (approved):
Thanks!  Looks great.

PR #12555, opened 2025-01-22
[Fix] Notify user of unknown support form errors

### LINKED ISSUE #12473: 🐛 No indication for failed support ticket submission
## 🐛 Bug

If the support ticket submission fails there is no indication to the user.

## 🦋 Expected Behaviour

A toast - I think there used to be one?  Something like "send an email instead"?

## 📋 Steps to Reproduce

1. Load up the support page
2. Stop the webserver container
3. Attempt to submit the ticket

## 📸 Screenshot

image

## 🙋‍♀️ Proposed Solution

Probably something wrong with the promise handling.

## ✅ Acceptance Criteria

- [ ] If support ticket fails to submit then the user is notified

esizer:
🤖 Resolves #12473 

## 👋 Introduction

Refactors the support form in order to handle unknown errors.

## 🕵️ Details

The form was handling all errors from the response except edge cases where we had no response to work with (500). This refactors it to handle that edge case and return the default error.

## 🧪 Testing

1. Build `pnpm run dev:fresh`
2. Navigate to support form `/support`
3. Turn off the web container
4. Submit the form
5. Confirm an error toast appears

### DIFF
--- a/apps/web/src/pages/SupportPage/components/SupportForm/SupportForm.tsx
+++ b/apps/web/src/pages/SupportPage/components/SupportForm/SupportForm.tsx
@@ -14,7 +14,7 @@ import {
   uiMessages,
 } from "@gc-digital-talent/i18n";
 import { Heading, Pending, Button } from "@gc-digital-talent/ui";
-import { useLogger } from "@gc-digital-talent/logger";
+import { Logger, useLogger } from "@gc-digital-talent/logger";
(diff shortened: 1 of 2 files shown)


mnigh (approved):
Tested happy path and Freshdesk ticket created successfully. Tested when webserver stopped and got expected error.

## Screenshot

Screen Shot 2025-01-22 at 10 22 18

PR #12559, opened 2025-01-23
[Feature] Never require asset behavioural skill assessment

### LINKED ISSUE #12552: 🐛 Asset behavioural skills should not require assessment
## 🐛 Bug

Currently we always expect asset behavioural skills to be assessed in computeAssessmentStatus on poolCandidates but only display them as needing assessment in the assessmentResultsTable if they have been 'claimed' by the applicant. Both of these should instead be updated to NEVER require asset behavioural skills to be assessed.

## 🦋 Expected Behaviour

Asset behavioural skills are not required to be assessed for a step to considered complete.

## 🙋‍♀️ Proposed Solution

Update both the computeAssessmentStatus function and assessmentResultsTable to never require asset behavioural skills to be assessed.

## ✅ Acceptance Criteria

- [ ] assessmentResultsTable never displays asset behavioural skills as needing assessment
- [ ] computeAssessmentStatus function does not require asset behavioural skills to be assessed for a step to be marked complete

JamesHuf:
🤖 Resolves #12552

## 👋 Introduction

Updates the site to prevent requiring assessment for non-essential behavioural skills. Essential skills and non-essential technical skills should remain the same.

## 🧪 Testing

1. Create a pool with both an essential skill and an asset behavioural skill.
2. Create an assessment step which assesses both skills.
3. Publish the pool filling in any other fields required.
4. Create and submit an application to the pool.
5. Navigate to view the application at `admin/candidates/{candidateId}/application`.
6. Confirm the essential skill appears as "To assess" and the asset skill as "Unclaimed" in the assessment step from 2.
7. Assess the essential skill as demonstrated.
8. Confirm the step appears as successful.

### DIFF
--- a/apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.tsx
+++ b/apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.tsx
@@ -19,6 +19,7 @@ import {
   PoolSkillType,
   Scalars,
   Skill,
+  SkillCategory,
   UpdateAssessmentResultInput,
   User,
   graphql,
@@ -484,7 +485,9 @@ export const ScreeningDecisionDialog = ({
             <>
               {intl.formatMessage(
                 poolSkill?.type?.value === PoolSkillType.Nonessential &&
-                  !experienceAttachedToSkill
+                  (!experienceAttachedToSkill ||
+                    poolSkill?.skill?.category.value ===
+                      SkillCategory.Behavioural)
                   ? poolCandidateMessages.unclaimed
                   : poolCandidateMessages.toAssess,
               )}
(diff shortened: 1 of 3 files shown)


apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.tsx
```diff
@@ -484,7 +485,9 @@ export const ScreeningDecisionDialog = ({
             <>
               {intl.formatMessage(
                 poolSkill?.type?.value === PoolSkillType.Nonessential &&
-                  !experienceAttachedToSkill
+                  (!experienceAttachedToSkill ||
+                    poolSkill?.skill?.category.value ===
```
brindasasi, line 489: this makes every behavioural skill is unclaimed even if it is connected to experience. But as per Desfam channel this is what they wanted seems like. Let's go with this. Thanks!

brindasasi (approved):
LGTM!

PR #12601, opened 2025-01-24
[Debt] Block frames for security

### LINKED ISSUE #12470: ♻️ Block iframe embeds
## ♻️ Debt/Refactor

We currently do not have a policy header for frames and should add one.

## 🕵️ Details

I do not think we use iframes anywhere on the site so we should probably just block them outright?

## 🙋‍♀️ Proposed Solution

Add the following headers (one we have and you just need to append to the CSP).

```
Content-Security-Policy: frame-ancestors 'none'
X-Frame-Options: DENY
```

## ✅ Acceptance Criteria

- [ ] Frame ancestors set to none in our CSP
- [ ] `X-Frame-Options` header set to `DENY`

esizer:
🤖 Resolves #12470 

## 👋 Introduction

Adds headers to block frames on our site.

## 🕵️ Details

Following [OWASP recommendations](https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Headers_Cheat_Sheet.html#recommendation), we are adding headers that block frames.

## 🧪 Testing

1. Rebuild your image `make down; make up;`
2. Navigate to a page on localhost
3. Confirm both `X-Frame-Options: Deny`  header and `frame-ancestors 'none'` option in CSP appear in headers

### DIFF
--- a/infrastructure/conf/nginx-conf-deploy/default
+++ b/infrastructure/conf/nginx-conf-deploy/default
@@ -1,6 +1,6 @@
 
 map $request_uri $csp_header {
-    default "default-src 'self'; font-src fonts.gstatic.com 'self'; style-src-elem fonts.googleapis.com 'self' 'unsafe-inline'; script-src-elem 'self' 'nonce-$cspNonce'; img-src 'self' data:; report-uri /api/csp-report/; report-to csp;";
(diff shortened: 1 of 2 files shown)


petertgiles:
> Works for me locally. @petertgiles, do you mind testing on one of the azure verticals?
Sorry, I've got the dev vertical tied up with the logging stuff right now.  I'll try to test this afterwards.

petertgiles:
If this is still "report only"  I'm not too worried though.

mnigh:
> > Works for me locally. @petertgiles, do you mind testing on one of the azure verticals?
> 
> Sorry, I've got the dev vertical tied up with the logging stuff right now. I'll try to test this afterwards.

I tested locally with Sign In Canada as the auth provider and was not blocked from signing in or signing out, so thankfully, no framing going on from what I could tell.

PR #12609, opened 2025-01-28
[Fix] Equity dialogs close on success

### LINKED ISSUE #11355: 🐛 Indigenous declaration dialog does not close on save
## 🐛 Bug

After a sucessful save, the Indigenous sef-declaration dialog does not close itself like our other dialogs.

## 🦋 Expected Behaviour

Dialog closes if saving is successful.

## 📋 Steps to Reproduce

1. Login as any user
2. Navigate to your profile `/applicant/personal-information`
3. Open the dialog for declaring Indigenous identity
4. Save the form
5. Observe that it is successful but dialog remains open

## 🙋‍♀️ Proposed Solution

You can find this pattern in other areas but I think we just need to control the dialog with the `isOpen` and `onOpenChange` props drilling them down and then setting `isOpen` to false on success of mutation

## ✅ Acceptance Criteria

- [ ] Dialog can still control itself
- [ ] Dialog closes on success
- [ ] Dialog remains open on error
- [ ] Fix behaviour on other EE dialogs as well
- [ ] Check both profile and application process

esizer:
🤖 Resolves #11355 

## 👋 Introduction

Fixes an issue where the equity dialogs were not closing on the user profile page. This also makes some adjustments to the queries on that page to avoid re-mounting it after each mutation.

## 🕵️ Details

We were storing sub in the snapshot and our tests were using the profile page query to confirm it matched. The issue here is twofold.

1. We likely should not store sub in the snapshot for security reasons
2. We moved sub to `UserAuthInfo`

We were querying `authInfo { id sub }` just to satisfy the snapshot test. Since we queried this, once the `User` became stale, it was also making its sub types as stale so the `AuthorizationContext` was re-rendering. This re-render caused the `Pending` component to unmount the children (basically the entire app) so the page "refreshed" and user lost their scroll position.

To fix this, we simply removed sub from the query and snapshot.

## 🧪 Testing

2. Login as any user
3. Go to your profile edit page `/applicant/personal-information`
4. Edit any form
5. Confirm the page does not refresh and you maintain your scroll position
6. Add an Indigenous identity to the equity section
7. Confirm the dialog closes on success
8. Edit that Indigenous identity but do not remove it
9. Confirm the dialog closes on success and you maintain your scroll position

### DIFF
--- a/api/app/GraphQL/Mutations/PoolCandidateSnapshot.graphql
+++ b/api/app/GraphQL/Mutations/PoolCandidateSnapshot.graphql
@@ -3,10 +3,6 @@
 query getProfile($userId: UUID!) {
   user(id: $userId) {
     id
-    authInfo {
-      id
-      sub
-    }
     firstName
     lastName
     email
--- a/apps/web/src/pages/Profile/ProfilePage/ProfilePage.tsx
+++ b/apps/web/src/pages/Profile/ProfilePage/ProfilePage.tsx
@@ -1,5 +1,5 @@
 import { defineMessage, useIntl } from "react-intl";
-import { useMutation, useQuery } from "urql";
+import { OperationContext, useMutation, useQuery } from "urql";
 
 import { TableOfContents, ThrowNotFound, Pending } from "@gc-digital-talent/ui";
 import { navigationMessages } from "@gc-digital-talent/i18n";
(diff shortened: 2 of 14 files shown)


api/app/GraphQL/Mutations/PoolCandidateSnapshot.graphql
```diff
@@ -3,10 +3,6 @@
 query getProfile($userId: UUID!) {
   user(id: $userId) {
     id
-    authInfo {
```
mnigh, line 6: > We likely should not store sub in the snapshot for security reasons

This makes sense to me, but can we confirm this as it is not technically part of the issue's scope?

apps/web/src/pages/Profile/ProfilePage/ProfilePage.tsx
```diff
@@ -27,88 +27,6 @@ const ProfileUpdateUser_Mutation = graphql(/* GraphQL */ `
   mutation UpdateUserAsUser($id: ID!, $user: UpdateUserAsUserInput!) {
     updateUserAsUser(id: $id, user: $user) {
       id
-      firstName
```
mnigh, line 30: This is great!

apps/web/src/pages/Profile/ProfilePage/ProfilePage.tsx
```diff
@@ -605,9 +519,16 @@ const ProfileUser_Query = graphql(/* GraphQL */ `
   }
 `);
 
+const context: Partial<OperationContext> = {
+  requestPolicy: "cache-first",
```
mnigh, line 523: What is this caching?

packages/auth/src/components/AuthorizationProvider.tsx
mnigh, line 1: 👍

mnigh (commented):
The acceptance criteria has one item "Check both profile and application process" and it looks like the application still reloads the page on submission of any of the equity dialogs there. Overall, the interactions are faster and and easier on the profile though.

## Screen recording
https://github.com/user-attachments/assets/d7779726-afe3-499e-84a4-ba2bd5ded73e

Another thing I noticed was on the dialogs (and I think this is most dialogs on the site and therefore likely out of scope), the submit button is disabled but the rest of the form elements are available to interact with, such as the checkbox and the Cancel link.

## Screen recording

https://github.com/user-attachments/assets/f0d02578-208f-4180-ab7b-0e34bc2422b3

esizer:
> The acceptance criteria has one item "Check both profile and application process" and it looks like the application still reloads the page on submission of any of the equity dialogs there. Overall, the interactions are faster and and easier on the profile though.

This one is a little different in that we have an inline spinner on all application pages (i think) so it looks like the page is refreshing but it is just hiding some jank. You will notice you don't actually lose the scroll position in the screen recording.

> Another thing I noticed was on the dialogs (and I think this is most dialogs on the site and therefore likely out of scope), the submit button is disabled but the rest of the form elements are available to interact with, such as the checkbox and the Cancel link.

Yeah, this is for all dialogs and forms in general I believe. I don't think we can actually disable form inputs because disabled inputs do not get submitted :cry:

mnigh:
> > The acceptance criteria has one item "Check both profile and application process" and it looks like the application still reloads the page on submission of any of the equity dialogs there. Overall, the interactions are faster and and easier on the profile though.
> 
> This one is a little different in that we have an inline spinner on all application pages (i think) so it looks like the page is refreshing but it is just hiding some jank. You will notice you don't actually lose the scroll position in the screen recording.

@petertgiles based on this, can we remove "Check both profile and application process" from the acceptance criteria or modify it to be "Check profile Equity section"? If so, I will approve this PR.

petertgiles:
> can we remove "Check both profile and application process" from the acceptance criteria or modify it to be "Check profile Equity section"? 

Yeah, if the dialog opens and closes when it should and the UX on update isn't terrible I think that's fine.

mnigh (approved):

PR #12653, opened 2025-01-30
[Feature] Rename `User.gov_*` columns to prefix with `computed_`

### LINKED ISSUE #12645: ✨ Add and rename gov_* user columns
> [!NOTE]
> This was split off from https://github.com/GCTC-NTGC/gc-digital-talent/issues/12217
 
## ✨ Feature

We're adding the collection of government information to experiences. This makes a lot of the info on the Government Employee section of the profile redundant. Instead of asking users about it twice (once in Experiences once in Profile) we'll ask it only in work experiences, and when any work experiences change we'll re-compute centralized values which are stored with the user model, making analytics and display easier.

## 🕵️ Details

In scope for this ticket:

- adding and renaming columns on the users table
- updated graphql schema so fields in User type continue to point to same underlying columns

Out of scope for this ticket:

- modifying UI or API
- compute and update gov_employee fields on users table when a 'current' work experience is saved or updated

Rename existing columns in users table to mark them as computed

    is_gov_employee -> computed_is_gov_employee
    gov_employee_type -> computed_gov_employee_type
    current_classification -> computed_current_classification
    department -> computed_department

Create new columns in users table (types to match corresponding experience fields)

    computed_gov_position_type
    computed_gov_end_date

## ✅ Acceptance Criteria

- [x] new computed gov employee columns added to user model
- [x] gov_employee columns renamed in database with computed_ prefix
- [x] graphql User fields pointing to renamed columns continue to work as they currently do
- [ ] graphql User fields pointing to new columns created for querying (but mutating not necessary)

## 🛑 Blockers

```[tasklist]
### Blocked By
```

esizer:
🤖 Resolves #12645 

## 👋 Introduction

Renames the fields related to being a government employee to be prefixed with `computed_` in preparation for the work experience event listener work.

## 🕵️ Details

This only renames and points existing fields to these for now. Actual computation will be done in separate tickets.

## 🧪 Testing

1. Refresh api `make refresh-api; make seed-fresh`
2. Confirm no errors
3. Confirm the following works:
     - Setting and displaying government info on profile/application
     - Downloading user government employee info

### DIFF
--- a/api/app/Generators/PoolCandidateCsvGenerator.php
+++ b/api/app/Generators/PoolCandidateCsvGenerator.php
@@ -173,9 +173,9 @@ class PoolCandidateCsvGenerator extends CsvGenerator implements FileGeneratorInt
                     $this->localizeEnum($candidate->user->written_level, EvaluatedLanguageAbility::class), // Writing level
                     $this->localizeEnum($candidate->user->verbal_level, EvaluatedLanguageAbility::class), // Oral interaction level
                     $this->localizeEnum($candidate->user->estimated_language_ability, EstimatedLanguageAbility::class),
-                    $this->yesOrNo($candidate->user->is_gov_employee), // Government employee
+                    $this->yesOrNo($candidate->user->computed_is_gov_employee), // Government employee
                     $department->name[$this->lang] ?? '', // Department
-                    $this->localizeEnum($candidate->user->gov_employee_type, GovEmployeeType::class),
+                    $this->localizeEnum($candidate->user->computed_gov_employee_type, GovEmployeeType::class),
                     $candidate->user->work_email, // Work email
                     $candidate->user->getClassification(), // Current classification
                     $this->yesOrNo($candidate->user->has_priority_entitlement), // Priority entitlement
--- a/api/app/Generators/PoolCandidateCsvGenerator.php
+++ b/api/app/Generators/PoolCandidateCsvGenerator.php
@@ -94,7 +94,7 @@ class PoolCandidateCsvGenerator extends CsvGenerator implements FileGeneratorInt
         'department',
         'employee_type',
         'work_email',
-        'current_classification',
+        'classification',
         'priority_entitlement',
         'priority_number',
         'accept_temporary',
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -364,11 +364,12 @@ class User extends Model implements Authenticatable, HasLocalePreference, Laratr
 
     public function getClassification()
     {
+
         if (! $this->computed_classification) {
             return '';
         }
 
-        $classification = $this->currentClassification()->first();
+        $classification = $this->currentClassification;
 
         $leadingZero = $classification->level < 10 ? '0' : '';
 
@@ -1354,6 +1355,10 @@ class User extends Model implements Authenticatable, HasLocalePreference, Laratr
 
         if (isset($snapshot['department'])) {
             $user->computed_department = $snapshot['department']['id'];
+        }
+
+        if (isset($snapshot['currentClassification'])) {
+            $user->computed_classification = $snapshot['currentClassification']['id'];
         }
 
         return $user;
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -377,7 +377,7 @@ class User extends Model implements Authenticatable, HasLocalePreference, Laratr
 
     public function getDepartment()
     {
-        if (! $this->department) {
+        if (! $this->computed_department) {
             return '';
         }
 
@@ -1353,7 +1353,7 @@ class User extends Model implements Authenticatable, HasLocalePreference, Laratr
         $user = self::hydrateFields($snapshot, $fields, $user);
 
         if (isset($snapshot['department'])) {
-            $user->department = $snapshot['department']['id'];
+            $user->computed_department = $snapshot['department']['id'];
         }
 
         return $user;
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -69,8 +69,8 @@ use Staudenmeir\EloquentHasManyDeep\HasRelationships;
  * @property ?\Illuminate\Support\Carbon $work_email_verified_at
  * @property ?bool $has_priority_entitlement
  * @property ?string $priority_number
- * @property ?string $department
- * @property ?string $current_classification
+ * @property ?string $computed_department
+ * @property ?string $computed_classification
  * @property ?string $citizenship
  * @property ?string $armed_forces_status
  * @property ?bool $is_woman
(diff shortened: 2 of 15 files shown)


api/database/migrations/2025_01_30_180016_rename_user_gov_columns.php
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->renameColumn('computed_is_gov_employee', 'is_gov_employee');
+            $table->renameColumn('computes_gov_employee_type', 'gov_employee_type');
```
mnigh, line 31:
```suggestion
            $table->renameColumn('computed_gov_employee_type', 'gov_employee_type');
```

mnigh (changes requested):

api/database/migrations/2025_01_30_180016_rename_user_gov_columns.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->renameColumn('is_gov_employee', 'computed_is_gov_employee');
+            $table->renameColumn('gov_employee_type', 'computed_gov_employee_type');
+            $table->renameColumn('current_classification', 'computed_classification');
```
vd1992, line 17: To be picky, it appears it should be `computed_current_classification`

image

api/database/migrations/2025_01_30_180016_rename_user_gov_columns.php
```diff
@@ -0,0 +1,38 @@
...
+            $table->renameColumn('department', 'computed_department');
+            $table->text('computed_gov_position_type')->nullable();
```
vd1992, line 19: On a tangent given sometimes we use `->text()` and sometimes `->string()`
Apparently it doesn't matter too much and text is usually recommended as a default choice 🤔 https://stackoverflow.com/a/4849030

api/database/migrations/2025_01_30_180016_rename_user_gov_columns.php
esizer: Okay, this was deliberate because of two things:

1. We almost never say current on any other fields (`current_email`, `current_phone_number`)
2. Current seems redundant since what else could it be? We don't store something like `past_classification` :sweat_smile:

api/database/migrations/2025_01_30_180016_rename_user_gov_columns.php
vd1992: Or maybe not, it might be heated debate 😅 

image

esizer: Yeah, I lean towards `text` as the default. I think `string` is better when you know the max length is only a few characters

api/app/Traits/Generator/GeneratesUserDoc.php
vd1992, line 1: Snapshot possesses all the info

image

Application doc download does not

image

api/database/migrations/2025_01_30_180016_rename_user_gov_columns.php
mnigh: > Okay, this was deliberate because of two things:
> 
> 1. We almost never say current on any other fields (`current_email`, `current_phone_number`)
> 2. Current seems redundant since what else could it be? We don't store something like `past_classification` 😅

I agree with @esizer in principle, but will other places need to be renamed as well?

## Screenshot
Screen Shot 2025-01-30 at 16 08 25

api/app/Traits/Generator/GeneratesUserDoc.php
vd1992: User doc download fails

image

api/app/Generators/PoolCandidateCsvGenerator.php
vd1992, line 1: CSV download fails too

image

vd1992 (changes requested):

api/database/migrations/2025_01_30_180016_rename_user_gov_columns.php
vd1992: I think some of those are labels/headers so would be in the realm of copy and not database naming

esizer: Okay, I left the actual copy alone but fixed the remaining references to `current_classification` which I beleive resolved the other issues @vd1992 found.

[remove current_ prefix from classification](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12653/commits/3d4be2bae197093bb9db964a9f2d050e174b35af)

vd1992: Might need to touch the other function too, see picture
Application doc download issue still persists

image

esizer: Oh weird! I must have got unlucky and tested the one application with no department :sweat_smile: 

I think it should be working now after [fix references to department](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12653/commits/f20c634c01fff406cfc51eb9d75080a8bd01b8b0)

petertgiles: Sorry, sloppiness in the issue.  The important thing is the "computed_" prefix.  There is no intention to change the column name other than the prefix (or fix other names) .

petertgiles: I'm not opposed to little fix-ups, as long as it doesn't risk adding bugs or blowing up the diff.

api/app/Models/User.php
```diff
@@ -1351,7 +1353,7 @@ public static function hydrateSnapshot(mixed $snapshot): User|array
         $user = self::hydrateFields($snapshot, $fields, $user);
 
         if (isset($snapshot['department'])) {
-            $user->department = $snapshot['department']['id'];
+            $user->computed_department = $snapshot['department']['id'];
         }
 
         return $user;
```
vd1992, line 1359: Department appears now, but classification is still missing from the application doc download. I'm not sure where the issue lies, maybe in `hydrateSnapshot`?
Profile doc download and application CSV however appear to work fine

Application doc download, government information
image

vd1992 (changes requested):

api/app/Models/User.php
esizer: Honestly a little confused as to how this was working on main :sweat_smile: Anyway, should be fixed after [fix classification hydration](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12653/commits/aa19177e76f032a86c034db8ab3a90455d8f2368)

vd1992 (approved):
I think it looks fine from what I can tell 👍

PR #12670, opened 2025-02-03
[Feature] Computed current gov work experience

### LINKED ISSUE #12643: ✨ Compute user gov_* fields based on experience
> [!NOTE]
> This was split off from https://github.com/GCTC-NTGC/gc-digital-talent/issues/12217
 
## ✨ Feature

We're adding the collection of government information to experiences. This makes a lot of the info on the Government Employee section of the profile redundant. Instead of asking users about it twice (once in Experiences once in Profile) we'll ask it only in work experiences, and when any work experiences change we'll re-compute centralized values which are stored with the user model, making analytics and display easier.

## 🕵️ Details

In scope for this ticket:

- compute gov_employee fields when a 'current' work experience is saved or updated
- phpunit test

Out of scope for this ticket:

- adding and renaming fields on the users table
- updating gov_employee fields on users table 
- updated graphql schema so fields in User type continue to point to same underlying fields
- modifying UI or API

## 🎨 Design

```
**experience selection logic**

1. First show the Group & Level from whichever government experience is marked as current
   1. Except if "employment type = contractor" or "employment type = student". These will be ignored. No group & level are collected for them anyway.
2. If multiple are marked current, show the one with the most recent start date
3. If multiple are marked current and have the same start date, prioritize the experience based on the employment type (though this probably should never happen):
   1. Show indeterminate first
   2. If not indeterminate, show Term
   3. Last show casual
4. If multiple are marked current, have the same start date and the same "employment type", then select based on the position type. (This actually is possible, for example you start a new substantive but are immediately put into an acting.)
   1. Show acting first
   2. If no acting, show secondment
   3. If no acting or secondment, show assignment
   4. Last show substantive
5. And if for some reason multiple experiences are marked current, have the same start date, have the same employment type and the same position type, then just pick whichever was created most recently by the user. (this should also never happen though)

```

- Create a laravel Event, ExperienceSaved, which dispatches [whenever a WorkExperience is saved](https://laravel-news.com/model-events). See AssessmentResultSaved as an example.
- Create a Laravel App Listener, ComputeGovEmployeeProfileData (open to other names). Similarly to how ComputeCandidateAssessmentStatus is set up to run in response to an AssessmentResultSaved event, set up the new listener to run in response to an ExperienceSaved event.
  - The ComputeGovEmployeeProfileData listener runs user->computeGovEmployeeProfile in the following cases:
    - The work experience is current (ie has no end date)
    - This is an update event, and before the update the experience was current, (ie end date was null but is updated to be non-null)

## ✅ Acceptance Criteria

- [ ] after user creates or updates a current work experience, or when updating a work experience previously marked as current, the computed gov employee fields are logged at a debug level
- [ ] computeGovEmployeeProfile is very readable, maintainable, testable

## 🛑 Blockers

```[tasklist]
### Blocked By
```

esizer:
🤖 Resolves #12643 

## 👋 Introduction

Computes and logs current work experience on created and updated work experience events.

## 🕵️ Details

This is just logging the expected values, storing them will be completed in a future ticket.

## 🧪 Testing

2. Login as any user role
3. Start be creating a few non-government experiences
4. Confirm that `computed_is_gov_employee` remains false
5. Add a few gov employee work experiences with varying position type, employment types and start dates
6. Confirm that the appropriate experience values are logged based on the logic outlined in the ticket

### DIFF
--- a/api/app/Listeners/ComputeGovEmployeeProfileData.php
+++ b/api/app/Listeners/ComputeGovEmployeeProfileData.php
@@ -68,14 +68,14 @@ class ComputeGovEmployeeProfileData
         });
 
         if ($sameStartDate->count()) {
-            $priortySortedExperiences = $sameStartDate
+            $prioritySortedExperiences = $sameStartDate
                 ->sortBy('created_at')
                 ->sortBy([
                     fn (WorkExperience $a, WorkExperience $b) => array_search($a->gov_position_type, $this->positionTypeOrder) <=> array_search($b->gov_position_type, $this->positionTypeOrder),
                     fn (WorkExperience $a, WorkExperience $b) => array_search($a->gov_employment_type, $this->employmentTypeOrder) <=> array_search($b->gov_employment_type, $this->employmentTypeOrder),
                 ]);
 
-            $latest = $priortySortedExperiences->first();
+            $latest = $prioritySortedExperiences->first();
         }
 
         Log::info([
--- a/api/app/Listeners/ComputeGovEmployeeProfileData.php
+++ b/api/app/Listeners/ComputeGovEmployeeProfileData.php
@@ -25,6 +25,7 @@ class ComputeGovEmployeeProfileData
         WorkExperienceGovEmployeeType::INDETERMINATE->name,
         WorkExperienceGovEmployeeType::TERM->name,
         WorkExperienceGovEmployeeType::CASUAL->name,
+        null,
     ];
 
     private $positionTypeOrder = [
@@ -32,6 +33,7 @@ class ComputeGovEmployeeProfileData
         GovPositionType::SECONDMENT->name,
         GovPositionType::ASSIGNMENT->name,
         GovPositionType::SUBSTANTIVE->name,
+        null,
     ];
 
     /**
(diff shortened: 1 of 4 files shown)


api/app/Models/WorkExperience.php
```diff
@@ -62,6 +63,13 @@ class WorkExperience extends Experience
         'experience_type' => WorkExperience::class,
     ];
 
+    /**
+     * Listeners for model events
+     */
+    protected $dispatchesEvents = [
```
petertgiles, line 69: I think we need to add the deleted event as well.  If someone deletes a work experience that could change what the "current value" is.

petertgiles (changes requested):
Very nice!  This works well.  I noted two small things, then should be ready to go.

I need to move the testing AC to the next issue, too.

api/app/Listeners/ComputeGovEmployeeProfileData.php
```diff
@@ -0,0 +1,90 @@
...
+
+        $latest = $currentExperiences->first();
+        $startDate = Carbon::parse($latest->start_date);
+        $sameStartDate = $currentExperiences->where(function ($experience) use ($startDate) {
+            // Is same month and year
+            return $experience?->start_date && $startDate->isSameMonth($experience->start_date, true);
+        });
+
+        if ($sameStartDate->count()) {
+            $priortySortedExperiences = $sameStartDate
```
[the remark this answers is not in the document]
esizer: oops! Fixed in [fix typo](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12670/commits/b82ca4f8ea9bd3cf72b31b6433471dc0919a1b6c)

api/app/Models/WorkExperience.php
esizer: oh yeah, that totally makes sense! Added in [add work experience deleted event to dispatch](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12670/commits/bce30cbc2937af21bbdf669af877fbd857a880cb)

petertgiles (approved):
Fixes look good, thanks!

Please note a small logic change requested by Gray before merging:
https://github.com/GCTC-NTGC/gc-digital-talent/issues/12643#issuecomment-2635101270

petertgiles:
Thanks for the quick update.  :heart:

PR #12524, opened 2025-01-16
[Feature] Community interest dialog

esizer:
🤖 Resolves #12285 

## 👋 Introduction

Adds a dialog component to view community interests.

## 🕵️ Details

This does not include the alert for missing information, that has been pushed until we need it.

## 🧪 Testing

1. Start storybook `pnpm run storybook`
2. Navigate to the `CommunityInterestDialog`
3. Confirm it matches the mock up

PR #12680, opened 2025-02-04
[Feature] Career development form

### LINKED ISSUE #12011: ✨ Career development form
## ✨ Feature

Inline form for editing a users career development, to be placed on the "Career planning" page.

## 🕵️ Details

This will be a new inline form using the mutations/queries defined in the [backend data for Career planning](https://github.com/GCTC-NTGC/gc-digital-talent/issues/12001) ticket.

The section subheading includes a status icon, showing `error` if any fields are empty, and `success` if all fields are filled.

## 🎨 Design

https://www.figma.com/design/Xnkie0dSq9RujvQRAZwrY4/Employee-profile-(Employees)?node-id=20-3397&t=gS6c1YiJVZrt1vWZ-0
@JoshBeveridge 

## 📸 Screenshot

Please see figma

Create a new component under `~/pages/Users/CareerPlanning/components` that uses the `ToggleSection` to include an inline form.

If the page stub is completed, include the component on the page under the second section as the first item.

## ✅ Acceptance Criteria

- [ ] Inline editing form component created
  - [ ] Radio buttons are required, checkboxes or checkbox groups are not
  - [ ] "Missing information" labels appear, the the overall status is incomplete, if any of the fields are _null_. Empty arrays are okay.
  - [ ] In the view mode, checkboxes are represented along with a green/grey icon, radio buttons are not.
- [ ] Connected to API (queries, mutations)
- [ ] If page exists, added to page in correct location
  - [ ] set the status of the Table of Contents entries.
- [ ] Design review by @JoshBeveridge 
- [ ] Language review by @NienkeBr  

## 🛑 Blockers

```[tasklist]
### Blocked By
- [ ] #12001
```

yonikid15:
🤖 Resolves #12011 

## 👋 Introduction

- Adds the career development form to the employee profile page
- Uses @esizer BoolCheckIcon component from the [Community interest dialogs](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12524), which may need tweaking
## 🕵️ Details

- There's going to be backend and frontend changes to this form in the near future. The figma design for this section has already been updated, so you'll need to view the images below to review everything.

## 🧪 Testing

1. Build app
2. Login as `applicant-employee@test.com`
3. Go to `en/applicant/employee-profile`
4. Ensure career development section matches design and functions correctly

## 📸 Screenshot

image (2)

image (4)

image (3)


brindasasi (changes requested):
* The work counter is working weird.  It doesn't count when I add the word in the next line. 
Screenshot 2025-02-06 at 1 37 36 PM

* For the larger text it looks like this . Could we wrap it ? 
Screenshot 2025-02-06 at 1 39 54 PM


* I'm unable to get the missing information screen shot that you have posted above.  As a new government employee I get these sections as below 

Screenshot 2025-02-06 at 1 54 16 PM

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/CareerDevelopmentSection.tsx
```diff
@@ -0,0 +1,503 @@
...
+  }: EmployeeProfile): FormValues => ({
+    organizationTypeInterest:
+      organizationTypeInterest?.map((x) => x.value) ?? [],
+    moveInterest: moveInterest?.map((x) => x.value) ?? [],
+    mentorshipStatus: mentorshipStatusToFormValues(mentorshipStatus),
+    mentorshipInterest: mentorshipInterest?.map((x) => x.value) ?? [],
+    execInterest: boolToYesNo(execInterest),
+    execCoachingStatus: execCoachingStatusToFormValues(execCoachingStatus),
+    execCoachingInterest: execCoachingInterest?.map((x) => x.value) ?? [],
+  });
```
esizer, line 243: Could this be defined outside of the component so it doesn't need to be recreated on every mount?

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/CareerDevelopmentSection.tsx
```diff
@@ -0,0 +1,503 @@
...
+                items={organizationTypeInterestOptions}
+              />
+              <Checklist
+                idPrefix="moveInterest"
+                legend={labels.moveInterest}
+                name="moveInterest"
+                id="moveInterest"
+                items={moveInterestOptions}
+              />
+              <Separator data-h2-margin="base(0)" />
```
esizer, line 426: Need the decorative prop here

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/CareerDevelopmentSection.tsx
```diff
@@ -0,0 +1,503 @@
...
+    value,
+    label: intl.formatMessage(getExecCoachingStatus(value)),
+  }));
+
+  const execCoachingInterestOptions: Radio[] = unpackMaybes(
+    careerDevelopmentOptions?.execCoaching,
+  ).map(({ value }) => ({
+    value,
+    label: intl.formatMessage(getExecCoachingInterest(value)),
+  }));
```
esizer, line 357: Seems like we have a lot of loops here for only 3-4 items. Would it be better to just write these arrays directly in the prop?

esizer (changes requested):

apps/web/src/pages/EmployeeProfile/EmployeeProfilePage.tsx
```diff
@@ -212,7 +226,10 @@ const EmployeeProfilePage = () => {
   return (
     <Pending fetching={fetching} error={error}>
       {data?.me ? (
-        <EmployeeProfile userQuery={data.me} />
+        <EmployeeProfile
+          userQuery={data.me}
+          careerDevelopmentOptionsQuery={data}
```
brindasasi, line 231: `data` already has me inside of it. I don't think you need it twice in here.  Remove `userQuery` from the props and you can it from `careerDevelopmentOptionsQuery` inside the `EmployeeProfile` component.

yonikid15:
> * The word counter is working weird.  It doesn't count when I add the word in the next line.
I couldn't recreate it. I might be doing it wrong. Either way I would say this is out of scope.

> * For the larger text it looks like this . Could we wrap it ?
We should fix this across the app. In the <FieldDisplay />.


> * I'm unable to get the missing information screen shot that you have posted above.  As a new government employee I get these sections as below

Isn't this normal behaviour? Its the same as all the other forms? I believe the design you are looking at hasn't been done yet.


> * Minor - Order of the section is different from figma you have posted above . Interest in promotions should come first

Okay! Nice catch!

brindasasi:
> > * I'm unable to get the missing information screen shot that you have posted above.  As a new government employee I get these sections as below
> 
> Isn't this normal behaviour? Its the same as all the other forms? I believe the design you are looking at hasn't been done yet.

>  It's in the A/C though.  @petertgiles  do you know we have a plan to build this in a sep story? 

Screenshot 2025-02-10 at 10 00 55 AM

petertgiles:
I think we did agree to use our existing inline form here so it might look a little different from the design because of that.

The way our forms work is if all fields are missing then a single message appears in the box instead of listing every missing field, like the language form here.  If at least one field is filled then the missing fields are listed, like the gov employee box.
image

Does this help?

brindasasi:
> I think we did agree to use our existing inline form here so it might look a little different from the design because of that.
> 
> The way our forms work is if all fields are missing then a single message appears in the box instead of listing every missing field, like the language form here. If at least one field is filled then the missing fields are listed, like the gov employee box. image
> 
> Does this help?

I get that . But this PR doesn't have that code for "Missing information" at all and Yoni's understanding is that , it will be done in different PR. 

Currently what's happening , when I very first time load the form I get this form 
410599498-810985f1-a7fa-4eed-bbc5-7c00ac279cb7

When I enter values for all the mandatory fields and leave out the optional fields as is , it displays like this 
Screenshot 2025-02-10 at 3 04 09 PM

There is no logic here to call the "Missing information" portion on any situation. 
I just want to make sure whether this is acceptable.

apps/web/src/pages/EmployeeProfile/EmployeeProfilePage.tsx
```diff
@@ -39,15 +47,20 @@ const EmployeeProfile_Fragment = graphql(/** GraphQL */ `
...
     employeeProfile {
       ...EmployeeProfileGoalsWorkStyle
+      ...EmployeeProfileCareerDevelopment
     }
   }
 `);
 
 interface EmployeeProfileProps {
   userQuery: FragmentType<typeof EmployeeProfile_Fragment>;
+  careerDevelopmentOptions?: EmployeeProfileCareerDevelopmentOptionsFragment;
```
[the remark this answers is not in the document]
yonikid15: This is outdated, fixed in [ae85322](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12680/commits/ae85322229bfc6af815887d60fae74f050e6e635)

apps/web/src/pages/EmployeeProfile/EmployeeProfilePage.tsx
```diff
@@ -209,10 +226,18 @@ const EmployeeProfilePage = () => {
     query: EmployeeProfilePage_Query,
   });
 
+  const careerDevelopmentOptions = getFragment(
+    EmployeeProfileCareerDevelopmentOptions_Fragment,
+    data,
+  );
```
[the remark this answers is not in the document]
yonikid15: This is outdated, fixed in [ae85322](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12680/commits/ae85322229bfc6af815887d60fae74f050e6e635)

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/CareerDevelopmentSection.tsx
```diff
@@ -0,0 +1,509 @@
...
+    moveInterest,
+    mentorshipStatus,
+    mentorshipInterest,
+    execInterest,
+    execCoachingStatus,
+    execCoachingInterest,
+  }: FormValues) => {
+    if (!userAuthInfo?.id) {
+      throw new UnauthorizedError();
+    }
```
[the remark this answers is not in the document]
yonikid15: Nice catch, I was following the format we had for the other form and didn't notice. Changed in [22c8483](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12680/commits/22c8483a97b8f6cf19dacc81c622f83c80124ce0)

Should I change it in the GoalsWorkStyleForm as well?

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/CareerDevelopmentSection.tsx
```diff
@@ -0,0 +1,509 @@
...
+            {
+              organizationTypeInterest,
+              moveInterest,
+              mentorshipStatus,
+              mentorshipInterest,
+              execInterest,
+              execCoachingStatus,
+              execCoachingInterest,
+            },
+            { keepDirty: true },
```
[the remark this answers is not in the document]
yonikid15: That's true! Changed in [6419c95](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12680/commits/6419c958e4b23b77362519784ae5bc821a2c7c67)

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/CareerDevelopmentSection.tsx
```diff
@@ -0,0 +1,509 @@
...
+                  legend={intl.formatMessage(
+                    employeeProfileMessages.execInterest,
+                  )}
+                  items={[
+                    ExecInterest.NOT_INTERESTED,
+                    ExecInterest.INTERESTED,
+                  ].map((value) => ({
+                    value,
+                    label: intl.formatMessage(getExecInterest(value)),
+                  }))}
```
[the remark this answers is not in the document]
yonikid15: Changed in [8854643](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12680/commits/8854643ad46681e9498b775d52efee6b84bab9c1)

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/CareerDevelopmentSection.tsx
```diff
@@ -0,0 +1,509 @@
...
+                  idPrefix="mentorshipInterest"
+                  name="mentorshipInterest"
+                  legend={intl.formatMessage(
+                    employeeProfileMessages.mentorshipInterest,
+                  )}
+                  items={mentorshipInterestOptions}
+                  rules={{
+                    required: intl.formatMessage(errorMessages.required),
+                  }}
+                />
```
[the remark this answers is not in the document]
yonikid15: This is outdated, it should have been changed in [896a341](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12680/commits/896a341aec2eb676e9133256bd9c1cbe69e918c1)

brindasasi:
> > There is no logic here to call the "Missing information" portion on any situation which is fine if we agreed to go with that. I just want to make sure whether this is acceptable. Also please remove it from A/C as well.
> 
> @brindasasi The design with the "Missing information" hasn't been done yet, so it should be acceptable to remove it from the A/C.

yeah ..  we had a separate story for that yesterday scrum. Could be removed in A/C here.

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/CareerDevelopmentSection.tsx
yonikid15: Moved in [22c8483](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12680/commits/22c8483a97b8f6cf19dacc81c622f83c80124ce0)

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/CareerDevelopmentSection.tsx
yonikid15: Always forget this 😅 [5d09538](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12680/commits/5d09538675cde5129e1d0480f266af18256951a4)

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/CareerDevelopmentSection.tsx
yonikid15: Moved in [6f560cb](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12680/commits/6f560cb4f502bfa0da754e1146a8c2fdb03d2855)

petertgiles:
There are changes to this form ready to be worked on now:
https://github.com/GCTC-NTGC/gc-digital-talent/issues/12745

I've asked Yoni to finish up this PR with the AC as-is, though perhaps the same level of "doneness" is not necessary here.  We'll immediately begin the changes after this PR is merged.

apps/web/src/pages/EmployeeProfile/EmployeeProfilePage.tsx
yonikid15: Hmm, I'm not sure what you mean here? I kept them separate since they are both fragments that will be used in the nested component.

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/utils.ts
```diff
@@ -0,0 +1,144 @@
...
+  ExecCoachingStatus,
+  getExecCoachingStatus,
+  getMentorshipStatus,
+  MentorshipStatus,
+} from "@gc-digital-talent/i18n";
+
+import { FormValues } from "./CareerDevelopmentSection";
+
+export const mentorshipStatusToFormValues = (
+  mentorshipStatus: EmployeeProfile["mentorshipStatus"],
```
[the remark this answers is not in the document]
yonikid15: I'm trying to keep it as closely typed to what is returned from the data. In this case:

`EmployeeProfile["mentorshipStatus"] = Maybe<LocalizedMentorship[]>`

Would you prefer I change it to `Maybe<LocalizedMentorship[]>` instead?

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/Display.tsx
```diff
@@ -0,0 +1,201 @@
...
+import {
+  displayExecCoachingStatus,
+  displayMentorshipStatus,
+  getLabels,
+} from "./utils";
+
+interface DisplayProps {
+  employeeProfile: EmployeeProfileCareerDevelopmentFragment;
+  careerDevelopmentOptions: EmployeeProfileCareerDevelopmentOptionsFragment;
+}
```
esizer, line 30: These should be something closer to `FragmentType<typeof EmployeeProfileCareerDevelopment_Fragment>`

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/Display.tsx
```diff
@@ -0,0 +1,201 @@
...
+                </li>
+              );
+            })}
+          </ul>
+        ) : (
+          notProvided
+        )}
+      </ToggleForm.FieldDisplay>
+      <ToggleForm.FieldDisplay
+        hasError={!!execInterest !== execInterest}
```
esizer, line 148: I could be wrong, but doesn't this always evaluate to the same thing? :sweat_smile:

apps/web/src/pages/EmployeeProfile/components/GoalsWorkStyleSection/GoalsWorkStyleSection.tsx
```diff
@@ -123,12 +122,8 @@ const GoalsWorkStyleSection = ({
     learningGoals,
     workStyle,
   }: FormValues) => {
-    if (!userAuthInfo?.id) {
-      throw new UnauthorizedError();
-    }
-
     return executeMutation({
-      id: userAuthInfo?.id,
+      id: userId,
```
esizer, line 126: Is there a reason we switched from just reading the user ID from the already queried auth data to just drilling it down?

esizer (changes requested):

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/CareerDevelopmentSection.tsx
```diff
@@ -0,0 +1,509 @@
...
+      initialData.mentorshipInterest?.map((x) => x.value) ?? [],
+    execInterest: initialData.execInterest
+      ? ExecInterest.INTERESTED
+      : ExecInterest.NOT_INTERESTED,
+    execCoachingStatus: execCoachingStatusToFormValues(
+      initialData.execCoachingStatus,
+    ),
+    execCoachingInterest:
+      initialData.execCoachingInterest?.map((x) => x.value) ?? [],
+  });
```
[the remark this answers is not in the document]
yonikid15: This is moved in [22c8483](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12680/commits/22c8483a97b8f6cf19dacc81c622f83c80124ce0)

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/Display.tsx
yonikid15: It's a little trick I saw recently :laughing:. ExecInterest type can be `execInterest: boolean | null | undefined`. If...
-  !!true === true
- !!false (or false) === false
- !!null (or false)  === null
- !!undefined (or false) === undefined

Since it's only false if it's null or undefined the little trick above works.
I forgot to revert this back to `empty(execInterest)` haha

apps/web/src/pages/EmployeeProfile/components/GoalsWorkStyleSection/GoalsWorkStyleSection.tsx
yonikid15: Short answer is typescript wants a string not a string | undefined. That's why someone added the check earlier.

```ts   
if (!userAuthInfo?.id) {
   throw new UnauthorizedError();
}
```

I can't remember if there's another way we were doing this?

esizer: You can just throw an error to prevent it and create a type guard

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/Display.tsx
yonikid15: Fixed in [81b268c](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12680/commits/81b268cfbfc291a2819e389ed49ff11c60235864)

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/Display.tsx
yonikid15: Refactored in [c3aaaf2](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12680/commits/c3aaaf22dc15805bb7efdf816ec2ec3e37dc76a1)

yonikid15:
@brindasasi 
- Fixed the text overflowing the page [787d7ff](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12680/commits/787d7ff8b1b3223ca91d209a799dc2f4ebc5aa73)
- Fixed the order of the form fields [50e162c](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12680/commits/50e162c5a3705ee4b70ebebbd37c7467afdd58ca)
- I wasn't able to recreate the word counter issue you found. However, that's not part of this issue haha. We should make a new issue to tackle that.

Anything else I might of missed?

brindasasi:
Will review in a moment ! Thanks !

apps/web/src/pages/EmployeeProfile/EmployeeProfilePage.tsx
```diff
@@ -39,15 +48,24 @@ const EmployeeProfile_Fragment = graphql(/** GraphQL */ `
...
+    typeof EmployeeProfileCareerDevelopmentOptions_Fragment
+  >;
 }
 
-const EmployeeProfile = ({ userQuery }: EmployeeProfileProps) => {
+const EmployeeProfile = ({
+  userId,
+  userQuery,
+  careerDevelopmentOptionsQuery,
+}: EmployeeProfileProps) => {
```
brindasasi, line 68: Its a small clean up. 
since userId , userQuery are retrieved from data , they can be assigned inside EmployeeProfile component. 
passing the data that you already have causes duplication. 
I meant this way , 

```
const EmployeeProfile = ({
careerDevelopmentOptionsQuery,
}: EmployeeProfileProps) => { 
const intl = useIntl();
 const paths = useRoutes();
 const userQuery = careerDevelopmentOptionsQuery.me;
 const userId = userQuery.id;
 
const user = getFragment(EmployeeProfile_Fragment, userQuery);
```
	
 ...

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/utils.ts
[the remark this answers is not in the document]
petertgiles: I like this type definition - I think it's very expressive about where the type comes from.

apps/web/src/pages/EmployeeProfile/EmployeeProfilePage.tsx
yonikid15: I removed the userId, but for the other two props, I think I should leave them since they are both fragments.

apps/web/src/pages/EmployeeProfile/components/CareerDevelopmentSection/CareerDevelopmentSection.tsx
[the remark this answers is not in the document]
petertgiles: Yep, this should never happen.  However, this check provides type narrowing to prove that we're not going to pass null into the mutation.

apps/web/src/pages/EmployeeProfile/EmployeeProfilePage.tsx
petertgiles: This makes sense.  One is a fragment on `User` and one is a fragment on `Query`, right?

apps/web/src/pages/EmployeeProfile/components/GoalsWorkStyleSection/GoalsWorkStyleSection.tsx
petertgiles: Thanks for switching back to the useAuth hook and using the `if` to narrow the type.

apps/web/src/pages/EmployeeProfile/EmployeeProfilePage.tsx
petertgiles: This makes sense. One is a fragment on User and one is a fragment on Query, right?

yonikid15: I could pass down the entire query in the props then split it up there, but it makes it easier with the typing to do it like this.

yonikid15: Nvm Brinda was right, I couldn't figure out this type error, but Brinda helped me see the light :bulb: 
Changed in [c0ec84c](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12680/commits/c0ec84c3f3e0afadb3e0556d80d7fd2811fa26af)

brindasasi (approved):
Looks great! Thank you for making the changes. I created sep issue for word counter component as that is out of scope.  Good to go! 🥳👏

PR #12734, opened 2025-02-10
[Fix] Timestamp updates on manual status change

### LINKED ISSUE #12726: 🐛  `removed_at` timestamp not getting cleared when status changes from Qualified Unavailable -> Qualified
## 🐛 Bug

Reverting a candidate from Qualified Unavailable to Qualified status still shows up as Unsuccessful on applicant dashboard side.

Reason for this is `removedAt` field is not getting cleared out.

Below is the screenshot of a candidate who became Qualified from Qualified Unavailable:

### Screenshot
Image

## 🦋 Expected Behaviour

Reverting a candidate from Qualified Unavailable to Qualified or any other positive status  (screened in, qualified available, etc.) should reflect on Applicant Dashboard properly.

## 🕵️ Details

Wipe out `removedAt` timestamp from the field when the candidate status is changed to any other not removed status (Qualified , Screened In, Placed, etc.).

## 📋 Steps to Reproduce

1. Make a candidate Qualified Unavailable on the admin side 
2. Observe the application card shows up on the dashboard with the pill "Unsuccessful" 
3. Make the same candidate Qualified on the admin side 
4. Observe the application card still shows up on the dashboard with the pill "Unsuccessful" 

## 📸 Screenshot

After reverting the status, application shows on the candidate table as Qualified
 
Image

Same application shows up unsuccessful on the applicant side 

Image

## ✅ Acceptance Criteria

- [ ] Reverting from unsuccessful status to any success status , the applicant side status pill should show up properly
- [ ] Confirm these changes are logged in to activity_log as well (This shouldn't impact.. just double confirm)

esizer:
🤖 Resolves #12726 

## 👋 Introduction

Sets application timestamps to null depending on manual status updates.

## 🕵️ Details

Three statuses in the test are commented out because I could not figure out permissions for those. Any assistance on who is allowed to set those statuses would be appreciated.

## 🧪 Testing

1. Login as a user who can change statuses on an application
2. Navigate to an application
3. Update the status
4. Confirm the appropriate time stamps get set to null


api/app/GraphQL/Mutations/UpdatePoolCandidateStatus.php
```diff
@@ -12,10 +12,17 @@
     public function __invoke(null $_, array $args)
     {
         $candidate = PoolCandidate::findOrFail($args['id']);
+        $values = [];
 
         if (isset($args['pool_candidate_status']) && $args['pool_candidate_status'] !== $candidate->pool_candidate_status) {
             $status = $args['pool_candidate_status'];
 
+            $values = [
```
petertgiles, line 20: This strategy of "set one timestamp and null all the rest" is not quite flexible enough, I think.  Some statuses, like PLACED_CASUAL, require having two set and one null.  Also, QUALIFIED_WITHDREW requires setting suspended_at, too.

api/tests/Feature/PoolCandidateUpdateTest.php
```diff
@@ -1229,4 +1229,143 @@ public function testManualStatusUpdatePolicy(): void
...
+
+    public static function nullTimeProvider()
+    {
+        $nullState = [
+            'removedAt' => null,
+            'finalDecisionAt' => null,
+            'placedAt' => null,
+        ];
+
+        $finalDeciscion = [
```
petertgiles, line 1266:
```suggestion
        $finalDecision = [
```

api/tests/Feature/PoolCandidateUpdateTest.php
```diff
@@ -1229,4 +1229,143 @@ public function testManualStatusUpdatePolicy(): void
...
+                PoolCandidateStatus::QUALIFIED_AVAILABLE->name,
+                $finalDeciscion,
+                'communityAdminUser',
+            ],
+            'qualified - available' => [
+                PoolCandidateStatus::QUALIFIED_AVAILABLE->name,
+                $finalDeciscion,
+                'communityAdminUser',
+            ],
+            // 'qualified - unavailable' => [
```
petertgiles, line 1329: It seems like these run OK, right?  They're just not giving you the result you expect?  It's not a user permission problem?

petertgiles (commented):
Thanks for picking this up so fast!  I think it needs to be expanded a bit, though.

api/tests/Feature/PoolCandidateUpdateTest.php
esizer: Hmm, I was actually getting an unauthorized error on them :thinking:

api/app/GraphQL/Mutations/UpdatePoolCandidateStatus.php
esizer: Huh, that seems confusing to me :sweat_smile: If final decision is getting set at the same time as placed than why differentiate them? If they are always the same, could we not just say they were placed on the final decision date? :thinking:

petertgiles: Yeah, this is the mismatch between the two status systems.  When manually changing the status it's one action but with the (preferred) state machine flow, qualifying with a final decision is one action and then placing is a second action.

esizer: Alright, updated the command to set multiple timestamps as discussed.

[allow setting multiple timestamps](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12734/commits/d061155b50ebf89aee02b22eef0609e86a31196f)

api/tests/Feature/PoolCandidateUpdateTest.php
esizer: Okay, not sure what was going on but they do work as expected. Added back  in [fix tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12734/commits/85a495b5af47cef5016356d9676659f6d599289f)

api/tests/Feature/PoolCandidateUpdateTest.php
esizer: Oops, fixed in [fix tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12734/commits/85a495b5af47cef5016356d9676659f6d599289f)

api/app/Console/Commands/UpdateApplicationTimestamps.php
JamesHuf, line 1: You don't appear to be handling QUALIFIED_WITHDREW at all here.

JamesHuf, line 1: I think we also want to ensure that certain timestamps **are** set.

api/app/GraphQL/Mutations/UpdatePoolCandidateStatus.php
```diff
@@ -12,41 +12,58 @@
...
+                PoolCandidateStatus::PLACED_INDETERMINATE->name => ['placed_at', 'final_decision_at'],
 
                 default => null// no-op
             };
 
-            if ($timestamp) {
-                $candidate->$timestamp = now();
+            if ($timestamps) {
+                foreach ($timestamps as $timestamp) {
+                    $values[$timestamp] = $now;
```
JamesHuf, line 49: I think if a timestamp is already set it would make sense to leave it with it's current value

api/tests/Feature/PoolCandidateUpdateTest.php
```diff
@@ -1229,4 +1229,123 @@ public function testManualStatusUpdatePolicy(): void
...
+                'candidate' => ['status' => $status],
+            ])->assertJsonFragment([
+                'status' => [
+                    'value' => $status,
+                ],
+                ...$expected,
+            ]);
+    }
+
+    public static function nullTimeProvider()
```
JamesHuf, line 1258: We don't just null timestamps we also set some. It might be nice to either update this or add a mirrored test that timestamps **are** set when desired.

esizer: We have that test already. However, it is now incorrect and I am working on trying to fix it.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/8d1557735efd095e1362407fc44c8720e9e11f17/api/tests/Feature/PoolCandidateUpdateTest.php#L990

JamesHuf: :facepalm: I clearly need to pay more attention to what else is in these files.

esizer: And it should be fixed now after [fix test for timestamp update](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12734/commits/42ed5114c7e80eda65e3e6d8fc8bd9bd2ee19c25)

api/app/Console/Commands/UpdateApplicationTimestamps.php
esizer: Oo good catch, added in [add missing status](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12734/commits/1dd4341ffe8ab06ec30853997eab6d4768d7dca0)

api/app/GraphQL/Mutations/UpdatePoolCandidateStatus.php
esizer: Hmm, I could be misunderstanding but the bug we are fixing is because we were not unsetting timestamps when they needed to so the intention is to fix that by unsetting those that are set when they should not be.

api/tests/Feature/PoolCandidateUpdateTest.php
esizer: It is a large file :laughing: I honeslty was about to add those tests in when I notice that test existed and the only reason I did notice is because it was failing :stuck_out_tongue:

api/app/GraphQL/Mutations/UpdatePoolCandidateStatus.php
JamesHuf: The goal of the issue is to ensure timestamps are null or not based on the status. I'm saying here that if we want the status to be set and it already is we could probably leave it alone.

For example if someone set a candidate to QUALIFIED_AVAILABLE a week ago their finalDecisionAt will be set to then. If we then set the user to placed casual we need to set placedAt to now but we could probably leave their finalDecisionAt as the date a week ago.

esizer: OH! okay, yeah I was just misunderstanding. That makes sense.

api/tests/Feature/PoolCandidateUpdateTest.php
esizer: That one actuall does seem to be impossible to set.. No matter what I get unauthorized both in the tests and UI. Now, I have no idea if that is expected or not so it may be another bug :woman_shrugging:

api/tests/Feature/PoolCandidateUpdateTest.php
JamesHuf: I think you're missing SCREENED_OUT_NOT_INTERESTED here.

api/app/GraphQL/Mutations/UpdatePoolCandidateStatus.php
petertgiles: Qualified withdrew also sets `suspended_at` in the spreadsheet.  That one is a little different since that timestamp never gets nulled out in a status change, only set.

api/app/GraphQL/Mutations/UpdatePoolCandidateStatus.php
esizer: Actually @petertgiles this is a good question. If they are changing a "final" decision to a different "final" decision should we record it as the last time the decision was set or leave the original.

Sadly, the "final" portion is not 100% true so it technically makes sense both ways :sweat:

api/app/GraphQL/Mutations/UpdatePoolCandidateStatus.php
esizer: That could be covered by:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/1dd4341ffe8ab06ec30853997eab6d4768d7dca0/api/app/GraphQL/Mutations/UpdatePoolCandidateStatus.php#L53-L55

api/app/GraphQL/Mutations/UpdatePoolCandidateStatus.php
petertgiles: It's not super obvious, is it?  :smile:  I think resetting a previously set field to "now" on a status change makes sense, like the "updated at" field.  If we need a log of the times all changes we should be referring to activity_log anyway.

api/app/GraphQL/Mutations/UpdatePoolCandidateStatus.php
petertgiles: Oh, it's underneath, sorry.  :+1:

api/app/GraphQL/Mutations/UpdatePoolCandidateStatus.php
JamesHuf: That argument works for me.

api/tests/Feature/PoolCandidateUpdateTest.php
esizer: Alright, sounds good. [add expired back with comment](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12734/commits/ce7f0fcdd2b09403cf3d8402c761dfe99454b9e1)

api/app/Console/Commands/UpdateApplicationTimestamps.php
esizer: Good call! Added in [update suspended at for withdrew](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12734/commits/f4f6d230044b7616b7e0037d4529f91a2af24401)

api/tests/Feature/PoolCandidateUpdateTest.php
JamesHuf: Sorry, I maybe shouldn't have added that here. I meant in testManualStatusUpdatesTimestamps.

api/app/Console/Commands/UpdateApplicationTimestamps.php
```diff
@@ -0,0 +1,131 @@
...
+    protected $signature = 'app:update-application-timestamps
+                            {--dry : Dry run to see how many candidates will be affected}
+    ';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Command description';
```
JamesHuf, line 26: If we are going to have a description it should at least say something useful.

api/app/GraphQL/Mutations/UpdatePoolCandidateStatus.php
JamesHuf, line 1: Application status label displayed to candidates for each status:
NEW_APPLICATION, APPLICATION_REVIEW, SCREENED_IN, UNDER_ASSESSMENT => "Received" / "Under review" / "Application reviewed" / "Under assessment" - this varies based on ROD results
DRAFT, DRAFT_EXPIRED, EXPIRED => Doesn't work, unauthorised error - Can we create another issue to fix/remove these?
PLACED_TENTATIVE, PLACED_CASUAL, PLACED_INDETERMINATE, PLACED_TERM, QUALIFIED_AVAILABLE, QUALIFIED_UNAVAILABLE, QUALIFIED_WITHDREW => "Qualified in process"
REMOVED, SCREENED_OUT_APPLICATION, SCREENED_OUT_ASSESSMENT, SCREENED_OUT_NOT_INTERESTED, SCREENED_OUT_NOT_RESPONSIVE => "Unsuccessful"

Qualified recruitment status label displayed to candidates for each status where it would appear:
PLACED_TENTATIVE, PLACED_CASUAL, PLACED_INDETERMINATE, PLACED_TERM => "Hired"
QUALIFIED_AVAILABLE, QUALIFIED_UNAVAILABLE => "Open to job offers" - will instead be "Not interested" if applicant previously suspended their application
QUALIFIED_WITHDREW => "Not interested" - applicant will be able to un-suspend themselves to instead be "Open to job offers"

petertgiles: Yeah, I did a similar exercise and sent this to Gray.
image

api/app/Console/Commands/UpdateApplicationTimestamps.php
esizer: Done! [update command description](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12734/commits/c80b60caac82d335a088f5f1ec6629fb9114a25f)

api/app/GraphQL/Mutations/UpdatePoolCandidateStatus.php
JamesHuf: Ok, that looks similar to mine. The one thing is PLACED_TENTATIVE which I don't think we should be setting placed_at for.

api/tests/Feature/PoolCandidateUpdateTest.php
```diff
@@ -1051,20 +1053,20 @@ public static function manualStatusProvider()
...
+            'qualified unavailable sets final decision' => [
+                PoolCandidateStatus::QUALIFIED_UNAVAILABLE->name,
+                'final_decision_at',
+            ],
+            'qualified withdrew sets final decision' => [
+                PoolCandidateStatus::QUALIFIED_WITHDREW->name,
+                'final_decision_at',
+            ],
 
             // Removed
```
JamesHuf, line 1065: I think you're missing SCREENED_OUT_NOT_INTERESTED here.

api/app/GraphQL/Mutations/UpdatePoolCandidateStatus.php
petertgiles: I left it as a question mark because I don't have it in my dropdown list.  Is it not possible to set in the UI?

JamesHuf: That one displays as "Offer in progress" in the list.

brindasasi: > DRAFT, DRAFT_EXPIRED, EXPIRED => Doesn't work, unauthorised error - Can we create another issue to fix/remove these?

No one should be able to set this status via manual status picker. Unauthorized error is correct.
Removing them from the status picker is even better solution.

brindasasi: Also please make sure Community Recruiters can not use these statuses 
PLACED_TENTATIVE
PLACED_CASUAL
PLACED_TERM
PLACED_INDETERMINATE

Community admins should be able to use these statuses without any problem

api/app/Console/Commands/UpdateApplicationTimestamps.php
petertgiles: Yes, from my analysis I think we do need to do this.

api/app/GraphQL/Mutations/UpdatePoolCandidateStatus.php
petertgiles: I'll ask Gray about PLACED_TENTATIVE/OFFER_IN_PROGRESS.

I think policy for who can make these changes is out of scope for now.

api/tests/Feature/PoolCandidateUpdateTest.php
petertgiles: Kind of out of scope.  Eric will complete this if there is time.

api/app/GraphQL/Mutations/UpdatePoolCandidateStatus.php
petertgiles: Good catch, James.  PLACED_TENTATIVE/OFFER_IN_PROGRESS does need to have `placed_at` null to work properly.  I've updated the Google Sheet.

api/app/Console/Commands/UpdateApplicationTimestamps.php
```diff
@@ -0,0 +1,158 @@
...
+
+        // Null non removed at fields
+        $removed = PoolCandidate::whereIn('pool_candidate_status', $this->removedStatuses)
+            ->where(function ($query) {
+                $query->whereNotNull('final_decision_at')
+                    ->orWhereNotNull('placed_at');
+            });
+
+        $removedCount = $this->update($removed, [
+            'final_decision_at' => null,
```
petertgiles, line 108: Sorry again for the last minute change I sprung on you yesterday.  Please let `final_decision_at` for REMOVED float and not update it.

esizer: okay, updated in [do not override final decsion for removed status](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12734/commits/313d17f85e20fb3f894e06a046e7e57c79f49f7d)

api/app/Console/Commands/UpdateApplicationTimestamps.php
```diff
@@ -0,0 +1,154 @@
...
+            [[$nullCount, $finalCount, $removedCount, $placedCount, $suspendedCount, $total]]
+        );
+    }
+
+    private function update(Builder $query, array $state)
+    {
+        if (! $this->dry) {
+            $query->chunk(100, function ($candidates) use ($state) {
+                foreach ($candidates as $candidate) {
+                    $candidate->update($state);
```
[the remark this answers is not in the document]
esizer: That is a very good point, updated in [update applicants quietly](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12734/commits/37f4dd10493b6b7fb33dd7e0a00c7e0212360eca)

petertgiles (approved):
Phew!  I think you got it.  :muscle:

brindasasi:
**Deployment Instructiions :** 

`app:update-application-timestamps`
                            {--dry  - to see how many candidates will be updated

PR #12741, opened 2025-02-11
[Feature] Placed casual displays as open to offers

### LINKED ISSUE #12738: ✨ Placed-casual leaves candidates as "Open to job offers / Not interested"
## User story

As a candidate, I want to see that I'm still "open to job offers" so that it's clear I'm still being referred for new opportunities.

## ✨ Feature

Currently, when a candidate is marked as place-casual, they show up as "Hired" in their career timeline. While this seems reasonable, casual placements are transient enough that we continue to refer the person for opportunities throughout. For this reason, it would be preferable to show the candidate that they are still "Open to job offers" unless the suspend themselves (so that it shows "Not interested").

## 🕵️ Details

Placed - casual displays to the candidates in the same way as "Qualified-available"

If `placed_at` is set and `Job placement status` is set to **Placed casual**, show on the career timeline: "**Open to job offers** or **Not interested**"

## ✅ Acceptance Criteria

- [ ] Placed-casual displays the same as qualified-available on the career timeline

vd1992:
🤖 Resolves #12738 

## 👋 Introduction

Have `PLACED_CASUAL` as an exception for the applicant's processes view, they see it as "open to offers" still

## 🧪 Testing

1. Submit an application
2. Head to application view for it and hop through actions
3. Set it to various states like qualified available, placed casual, and other placed fields
4. Observe status chip at `/en/applicant/career-timeline#qualified-recruitment-processes`

## 📸 Screenshot

image

image

### DIFF
--- a/apps/web/src/components/QualifiedRecruitmentCard/utils.ts
+++ b/apps/web/src/components/QualifiedRecruitmentCard/utils.ts
@@ -93,6 +93,7 @@ export const getQualifiedRecruitmentInfo = (
     statusChip: getQualifiedRecruitmentStatusChip(
       candidate.suspendedAt,
       candidate.placedAt,
+      candidate.status?.value ?? null,
       intl,
     ),
     availability: getAvailabilityInfo(candidate, intl),
--- a/apps/web/src/utils/poolCandidate.ts
+++ b/apps/web/src/utils/poolCandidate.ts
@@ -574,9 +574,11 @@ const qualifiedRecruitmentStatusDescriptions = defineMessages({
 export const getQualifiedRecruitmentStatusChip = (
   suspendedAt: PoolCandidate["suspendedAt"],
   placedAt: PoolCandidate["placedAt"],
+  status: PoolCandidateStatus | null,
   intl: IntlShape,
 ): StatusChipWithDescription => {
-  if (placedAt) {
+  // placed casual is an exception
+  if (placedAt && status !== PoolCandidateStatus.PlacedCasual) {
     return {
       color: "secondary",
       label: intl.formatMessage(qualifiedRecruitmentStatusLabels.HIRED),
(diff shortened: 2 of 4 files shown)


vd1992:
Extremely handy story

image

apps/web/src/components/QualifiedRecruitmentCard/utils.ts
```diff
@@ -37,7 +37,11 @@ const getAvailabilityInfo = (
     };
   }
 
-  if (isPlacedStatus(status?.value)) {
+  // placed casual is an exception, it can be suspended
+  if (
```
petertgiles, line 41: Because this logic is so confusing already, it might help if we name an intermediate variable.  Something like:
```ts
const isLongTermPlacedStatus =  isPlacedStatus(status?.value) &&
    status?.value !== PoolCandidateStatus.PlacedCasual;
```

apps/web/src/utils/poolCandidate.ts
```diff
@@ -80,7 +80,11 @@ export const isSuspendedStatus = (
 ): boolean => {
   const isSuspended = suspendedAt && new Date() > parseDateTimeUtc(suspendedAt);
 
-  return !!(isSuspended && status === PoolCandidateStatus.QualifiedAvailable);
+  return !!(
+    isSuspended &&
+    (status === PoolCandidateStatus.QualifiedAvailable ||
+      status === PoolCandidateStatus.PlacedCasual)
```
petertgiles, line 86: Same comment about naming this to make it easier to understand.  I think that "QualifiedAvailable and PlacedCasual" should probably be one of the categories in constants/poolCandidate, or a new category should get created.  SUSPENDABLE_STATUSES?

petertgiles (changes requested):
This seems to work great!  Thanks for the quick work on it.  I have some suggestions to make it a little easier to follow.

apps/web/src/components/QualifiedRecruitmentCard/utils.ts
vd1992: Sure https://github.com/GCTC-NTGC/gc-digital-talent/pull/12741/commits/9c65393a70a701de5761e915625d8e761c2826ea

apps/web/src/utils/poolCandidate.ts
vd1992: I think a category should be created when we've got more than one applicable entity for some purpose. Ease of consistency and reuse that way
https://github.com/GCTC-NTGC/gc-digital-talent/pull/12741/commits/a8e426ac7baf22ec1683662a5971e014b62dc172

petertgiles (approved):
Looks great.  Thanks!

PR #12817, opened 2025-02-20
Create bug_report.yaml

petertgiles:
## 👋 Introduction

Tasklist blocks are being sunset so we'll need to update our templates.  This is a test template we can try out that uses the new yaml syntax form builder.  It doesn't remove any templates, though.

## 🕵️ Details

It looks like the new sub-issue feature could handle our dependencies for us and allow us to get rid of the block tasklist and action.
image
image

## 🧪 Testing

1. You can submit an issue on my repo: https://github.com/petertgiles/yaml_issues/issues

> [!NOTE]
> I don't have the new issue type feature on my own repo yet. 

## 📸 Screenshot

image

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -1,7 +1,6 @@
 name: Bug Report (YAML)
 description: File a bug report.
 title: "🐛 "
-labels: ["bug"]
 projects: ["GCTC-NTGC/8"]
 type: "Bug"
 body:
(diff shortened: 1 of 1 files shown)


petertgiles:
I'm not sure sub-issues will be able to capture our dependency graph very well.  A subissue can only have one parent which means that an issue can only act as a blocker for one other issue.  :disappointed:

esizer (approved):
Even though we still have problems, at least this works :+1: I also really like how we have separate fields in the form now instead of one big description field! :clap:

PR #12824, opened 2025-02-20
[Bug] Add separator to "Goals and work style" form

petertgiles:
🤖 Resolves 

## 👋 Introduction

Adds the separators to the "Goals and work style" form.

> [!NOTE]
> I know that the spacing on the form should be x1.5 all around in desktop width but that's a whole other :worm: :can:.  I figure we can tackle that when we update the inline form component. 

## 🧪 Testing

1. Log in as `applicant-employee@test.com`
2. /applicant/employee-profile#goals-work-style-section

## 📸 Screenshot

image


vd1992 (approved):
👍

PR #12833, opened 2025-02-24
[Debt] Upgrade eslint

### LINKED ISSUE #10189: ♻️ Bumpy eslint to v9
## ♻️ Debt/Refactor

Looks like eslint v9 will require a manual update as it currently fails
Breaking changes https://github.com/eslint/eslint/releases/tag/v9.0.0

https://github.com/GCTC-NTGC/gc-digital-talent/pull/10178

## ✅ Acceptance Criteria

- [ ] Dependency updated

esizer:
🤖 Resolves #10189 

## 👋 Introduction

Upgrades eslint to v9 with flat config.

## 🕵️ Details

> [!NOTE]
> The `no-literal-string-in-jsx` [rule](https://formatjs.github.io/docs/tooling/linter/#no-literal-string-in-jsx) now works with flat config and lead to some fixes, thanks @esizer!

## 🧪 Testing

1. Install new deps `pnpm i`
2. Run lint `pnpm run lint`
3. Confirm no errors


esizer:
> So much for 30 line diff. 😆
> @mnigh Does this fix your environment?

lol I meant 30 lines more than what we already had :laughing: I'm going to do a write up but it comes down to basically one thing, we fell so far behind and have needed to close eslint tickets for the past 3-4 months that things just stopped working and it either failed silently or crashed. Most file touches were ignoring rules that were not working but are now.

We still have some performance issues but I believe that is mostly related to TS at this point.

petertgiles (changes requested):
Blocking until monday

yonikid15 (commented):
It was looking good, so I tried running it again after merging main on my local branch. Now seeing lint errors with the playwright assets:
image

Is this something I can fix on my end, or is it happening for anyone else?

esizer:
> It was looking good, so I tried running it again after merging main on my local branch. Now seeing lint errors with the playwright assets:

This is two things here:

- Some rules are not working on main so I need to keep this in sync to fix any errors coming from main which I did this morning in [fix lint warnings from main](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12833/commits/042fa35b34d896788f8ba89ee60344ef6b4f1689)
- This was an oversight when we were re-writing configs, fixed in [ignore playwright reports](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12833/commits/f50d11fdc177e642db9962850904b5af8444b026)

apps/web/src/pages/Home/IAPHomePage/components/Quote.tsx
vd1992, line 1: This one file seems to do a separate thing, rather than use the ignore line for string literals it defines a variable and uses it. Accident or desired method?

esizer: This was intentional because eslint was have a very hard time with this one and it was the only way I could get it to work :sob:

vd1992:
I can see that matt (probably) edited in the note, but it is still funny looking at this and seeing Eric thanking Eric

image

yonikid15 (approved):

yonikid15:
You might want to update the branch one more time just incase, but looks good to me

PR #12837, opened 2025-02-24
[Bug] Limit application edit after submission

### LINKED ISSUE #12138: 🐛 Applicant can modify application after submission
## 🐛 Bug

User shouldn't be able to modify any part of application after the submission. 
We should provide it only as read only. 

In this example I added an experience after the submission 
Image

## 🦋 Expected Behaviour

- Candidate should not edit their application 

## 📋 Steps to Reproduce

1. Submit an application 
2. Go back to the application and edit the skill requirements

## ✅ Acceptance Criteria

- [ ] Once the application is submitted it should be given as a link as downloadable on the applicant dashboard 
- [ ] Upon clicking that link application snapshot should be downloaded

vd1992:
🤖 Resolves #12138

## 👋 Introduction

Limit the things a user can attempt to edit. 
Application skills page, but also the edit button on the career timeline page. So the user can't edit prior experiences in both places

## 🧪 Testing

1. Go through the application process, submit it
2. Step through the submitted application
3. Be attentive

## 📸 Screenshot

image

image

image


esizer:
> You should be able to edit your previous experience in profile.
> But if you edit there does it show up in the application ?

Yes, it does show up in the application

PR #12847, opened 2025-02-26
[Feature] Government employee basic details

### LINKED ISSUE #12787: ✨ Basic Employee Details
## ✨ Feature

For GC employees, their name, email, department & job are not considered personal information. That's why GEDS can exist. Let's call these fields the "basic employee details".

For the nomination process we need to allow the user to search for government employees at several points.  We will expose this data like we do with UserPublicProfile.

## 🕵️ Details

One other piece needs to be added to make the feature work as well:
- Populate the current gov work experience `role` as `computed_gov_role` on the user, like we do with the computed_gov_employee_type.

## 📸 Screenshot

Image

Let's use UserPublicProfile as an example, here.

## ✅ Acceptance Criteria

- [ ] A "basic employee details" includes some fields:
  - [ ] first name
  - [ ] last name
  - [ ] role
  - [ ] department
  - [ ] work email
- [ ] A public employee profile only "exists" if the user is a "verified government employee"
  - [ ] based on https://github.com/GCTC-NTGC/gc-digital-talent/issues/12828
- [ ] Any authenticated user can search and view basic employee details
- [ ] Search can only by done by exact work email address
- [ ] 'computed_gov_role' added to the user model and updated in the same way as other "computed_gov" fields
- [ ] validation on search
  - [ ] recognizable error if work email address is not valid government address
  - [ ] ~recognizable error if work email address is in the system but not "verified government employee"~ #12831 

## 🛑 Blockers

```[tasklist]
### Blocked By
```

esizer:
🤖 Resolves #12787 

## 👋 Introduction

Adds query to get basic details for a verified government employee.

## 🕵️ Details

Adds the following:

- Government role to user model computed by latest government work experience
- Query to get employee details from exact work email for only verified government employees
- Validation rule for government email addresses

## 🧪 Testing

<details>
<summary>GraphQL query</summary>

```graphql
query GovEmployee($workEmail: String!) {
  govEmployee(workEmail: $workEmail) {
    id
    firstName
    lastName
    workEmail
    role
  }
}
````

```json
{
  "workEmail": "email@domain.com"
}
```

</details>

1. Refresh en `make refresh-api; make seed-fresh;`
2. Navigate to graphiql `http://localhost:8000/graphiql`
3. Authenticate via access token or setting default auth user
4. Run the provided query with different emails
5. Confirm the following cases:
    1. Work email with government tld for verified government employee returns details
    2. Work email with non-government tld returns validation error
    3. Any email for unverified government employee returns null


api/tests/Feature/GovEmployeeDetailsTest.php
```diff
@@ -0,0 +1,155 @@
...
+            ]);
+        } else {
+            $res->assertGraphQLValidationError('workEmail', ApiErrorEnums::NOT_GOVERNMENT_EMAIL);
+        }
+    }
+
+    public static function validationProvider(): array
+    {
+
+        $tlds = [
```
mnigh: Technically, none of these are top-level domains: `.ca` is a top-level domains. The array includes some second-level domains like `gc.ca`. Using a generic name like `$domains` is probably more accurate here.

api/app/Models/User.php
```diff
@@ -1171,6 +1178,12 @@ public function getImproveBehaviouralSkillsRankingAttribute()
             ->sortBy('improve_skills_rank');
     }
 
+    public function scopeIsVerifiedGovEmployee(Builder $query): void
+    {
+        $query->where('computed_is_gov_employee', true)
+            ->whereNotNull('work_email_verified_at');
```
petertgiles, line 1184: It's probably not possible to set `work_email_verified_at` without setting a work email so this probably works fine.  However, our definition includes `work_email` is not null and I think it is more understandable and secure if that's included here explicitly.

api/app/Rules/GovernmentEmail.php
```diff
@@ -0,0 +1,25 @@
...
+{
+    /**
+     * Run the validation rule.
+     *
+     * @param  \Closure(string, ?string=): \Illuminate\Translation\PotentiallyTranslatedString  $fail
+     */
+    public function validate(string $attribute, mixed $value, Closure $fail): void
+    {
+        $governmentEmailRegex =
+        '/@([A-Za-z0-9-]+\.)*(gc\.ca|canada\.ca|elections\.ca|ccc\.ca|canadapost-postescanada\.ca|gg\.ca|scics\.ca|scc-csc\.ca|ccohs\.ca|cchst\.ca|edc\.ca|invcanada\.ca|parl\.ca|telefilm\.ca|bankofcanada\.ca|banqueducanada\.ca|ncc-ccn\.ca|bank-banque-canada\.ca|cef-cce\.ca|cgc\.ca|nfb\.ca|onf\.ca|canadacouncil\.ca|conseildesarts\.ca|humanrights\.ca|droitsdelapersonne\.ca|ingeniumcanada\.org|cjc-ccm\.ca|bdc\.ca|idrc\.ca|museedelhistoire\.ca|historymuseum\.ca|cdic\.ca|sadc\.ca|scc\.ca|clc\.ca|clc-sic\.ca|cntower\.ca|latourcn\.ca)$/i';
```
petertgiles, line 19: Can we store this is one place in Laravel instead of copy/pasting it?  Right now it also exists in api/app/GraphQL/Validators/UpdateUserAsUserInputValidator.php , right?

api/graphql/schema.graphql
```diff
@@ -283,6 +284,15 @@ type UserPublicProfile @model(class: "\\App\\Models\\User") {
   lastName: String @rename(attribute: "last_name")
 }
 
+type BasicGovEmployeeDetails @model(class: "\\App\\Models\\User") {
```
petertgiles, line 287: Shoot, seeing this now, I see that it's confusing that it sounds like a plural name.  Maybe `BasicGovEmployeeProfile` so it aligns with the UserPublicProfile and it sounds singular?

api/graphql/schema.graphql
```diff
@@ -1069,6 +1079,12 @@ type Query {
   myAuth: UserAuthInfo @auth
   me: User @auth
   user(id: UUID! @eq): User @find @softDeletes @guard @canQuery(ability: "view")
+  govEmployee(
```
petertgiles, line 1082: Usually the query name matches (or is close to) the type it returns.  Could we get closer with this?

api/tests/Unit/GovernmentEmailValidationTest.php
```diff
@@ -0,0 +1,84 @@
...
+
+        $passes = [];
+        foreach ($domains as $domain) {
+            $passes[$domain.' passes validation'] = ['passes@'.$domain, true];
+        }
+
+        return [
+            'non government email fails validation' => ['email@domain.com', false],
+            ...$passes,
+        ];
```
petertgiles, line 82: Could we add a test for something like `endswithgc.ca` fails?

petertgiles (commented):
Just did a readthrough.  Looks great!  I'll run it soon.

api/app/Models/User.php
esizer: I agree, that makes sense to me! Added in [add work email to verified employee scope](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12847/commits/8689dad7492d46f5cbf1ae05b5f9063c62377c6b)

api/graphql/schema.graphql
esizer: Good call, renamed in [rename type and query](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12847/commits/0c60b716655f506fa9f05ed90e2614f00be3f783)

api/tests/Unit/GovernmentEmailValidationTest.php
esizer: Oh yeah, I forgot we ran into that edge case before :ok_hand: Added in [add additional test case](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12847/commits/dec178adaf33dbf7b5cfac24b6b48a1e03b8a97c)

api/graphql/schema.graphql
esizer: Right, good catch. Updated in [rename type and query](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12847/commits/0c60b716655f506fa9f05ed90e2614f00be3f783)

petertgiles (changes requested):
Oh, it works really well!  A few last things to consider.

api/graphql/schema.graphql
```diff
@@ -1069,6 +1079,12 @@ type Query {
   myAuth: UserAuthInfo @auth
   me: User @auth
   user(id: UUID! @eq): User @find @softDeletes @guard @canQuery(ability: "view")
+  govEmployeeProfile(
+    workEmail: String! @scope(name: "exactWorkEmail")
```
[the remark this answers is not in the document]
esizer: I'm not opposed to it if you strongly one way. I just did this since it was easy and thought we could extract out once needed.

api/app/GraphQL/Validators/UpdateUserAsUserInputValidator.php
```diff
@@ -45,8 +46,7 @@ public function rules(): array
                  */
                 Rule::unique('users', 'email')->ignore($this->arg('id'), 'id'),
                 Rule::unique('users', 'work_email')->ignore($this->arg('id'), 'id'),
-                // Note: Should be kept in sync with the workEmailDomainRegex
```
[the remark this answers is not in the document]
esizer: Oops, I totally meant to include that but forgot :sweat_smile: Brought back in [add comment back to gov email regex](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12847/commits/25d49dae5314f180ccacadc9b15b839466580985)

api/graphql/schema.graphql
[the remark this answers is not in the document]
petertgiles: Yeah, that's fair.  Don't worry about it.

api/app/Models/User.php
```diff
@@ -1080,7 +1080,7 @@ public static function scopeWorkEmail(Builder $query, ?string $email): Builder
 
     public static function scopeExactWorkEmail(Builder $query, string $email): Builder
     {
-        return $query->where('work_email', $email);
+        return $query->whereRaw('LOWER("work_email") like ?', [strtolower($email)]);
```
petertgiles, line 1083: If it's an exact match we can use `=` instead of `like` which will be faster, I think.

esizer: Good call! [user = over like for exact matching](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12847/commits/6e3feb94c8408ed14959bd9bf22489332b555882)

petertgiles (approved):
It's so good!  Thanks.  :bow:

Issue #12736, opened 2025-02-10
♻️ Avoid use of `User` type in validators


esizer:
## ♻️ Debt/Refactor

We use the full `User` type in our validators which is forcing us to query more data than we need to determine validity. Also, the types are fairly loose so we can forget to query something as well 👎 

## 🕵️ Details

Now that we have access to fragments, we no longer need to query non-nullable fields just to satisfy types. We can just query the required data.

## 🙋‍♀️ Proposed Solution

- Create a fragment for each validator
- Use specific types instead of `Pick<User, '...'>`

Also, it would be beneficial to review where we use these validators to determine if we can reduce the data queried to use them.

## ✅ Acceptance Criteria

- [ ] Validators do not use the generated `User` type
- [ ] Validator arguments are more specific to the data we need

PR #12850, opened 2025-02-26
[Debt] Move verified gov employee to backend

### LINKED ISSUE #12828: ♻️ Move "verified government employee" functionality to backend
## ♻️ Debt/Refactor

We have a function named `isVerifiedGovEmployee` in apps/web/src/utils/userUtils.ts .  Let's move this calculation to an attribute of the User model in the backend.

## ✅ Acceptance Criteria

- [ ] New calculated attribute added to the User model
  - [ ] The same calculation is used: isGovEmployee && workEmail && isWorkEmailVerified
- [ ] Attribute exposed through graphql
- [ ] The two places this calculation is used are switched over to the backend
- [ ] Frontend calculation removed

## 🛑 Blockers

```[tasklist]
### Blocked By
```

esizer:
🤖 Resolves #12828 

## 👋 Introduction

Moves the calculation of whether a user us a verified government employee to the backend instead of computing it on the server.

## 🧪 Testing

1. Rebuild `make refresh-api; make seed-fresh; pnpm dev:fresh;`
2. Login as a user who is not a verified government employee
3. Confirm the pages protected by this check cannot be accessed
4. Login as a verified government employee
5. Confirm pages protected by this check can be accessed

### DIFF
--- a/apps/web/src/pages/ApplicantDashboardPage/ApplicantDashboardPage.tsx
+++ b/apps/web/src/pages/ApplicantDashboardPage/ApplicantDashboardPage.tsx
@@ -13,7 +13,6 @@ import SEO from "~/components/SEO/SEO";
 import { getFullNameHtml } from "~/utils/nameUtils";
 import RequireAuth from "~/components/RequireAuth/RequireAuth";
 import Hero from "~/components/Hero";
-import { isVerifiedGovEmployee } from "~/utils/userUtils";
 import messages from "~/messages/profileMessages";
 import {
   aboutSectionHasEmptyRequiredFields,
@@ -32,8 +31,9 @@ export const ApplicantDashboardPage_Fragment = graphql(/* GraphQL */ `
     firstName
     lastName
     isGovEmployee
-    workEmail
-    isWorkEmailVerified
+    isVerifiedGovEmployee
+    hasPriorityEntitlement
+    priorityNumber
     employeeProfile {
       ...CareerDevelopmentTaskCard
     }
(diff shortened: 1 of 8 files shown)


apps/web/src/pages/ApplicantDashboardPage/ApplicantDashboardPage.tsx
```diff
@@ -32,8 +31,9 @@ export const ApplicantDashboardPage_Fragment = graphql(/* GraphQL */ `
     firstName
     lastName
     isGovEmployee
-    workEmail
-    isWorkEmailVerified
+    isVerifiedGovEmployee
+    hasPriorityEntitlement
```
JamesHuf, line 35: Is there a better way of typing helper functions to catch under-fetching like this?

JamesHuf (approved):
Work for me!

apps/web/src/pages/ApplicantDashboardPage/ApplicantDashboardPage.tsx
esizer: I was kind of thinking about this myself. I think if we can avoid using types like this:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/7c209c78367b62fb84f41a9ae8f799f768fe5212/apps/web/src/validators/profile/governmentInformation.ts#L4-L17

We might be better off. The problem is, unless the schema marks them as non nullable we get something like this:

```tsx
type User = {
  hasPriorityEntitlement?: Maybe<string>;
}|
```

So it would never get caught if we don't fetch it because it is totally valid to be `undefined.` I think something like this would be more accurate:

```tsx
type HasAllEmptyFieldsArgs = {
  hasPriorityEntitlement: string | null;
}
```

This way, we are forced to pass it in.

esizer: lol turns out I already noted this and created a ticket for it :laughing: 

https://github.com/GCTC-NTGC/gc-digital-talent/issues/12736

PR #12884, opened 2025-02-28
[Fix] Resetting government fields on work experience form

### LINKED ISSUE #12769: 🐛 Changing from Contractor to Indeterminate doesn't hide `contractorFirmAgencyName` field
## 🐛 Bug

When updating the experience from Government Contractor to any other employment type (like indeterminate or student) Contractor firm or agency field is not hiding.

Image

## 🦋 Expected Behaviour

Contractor firm or agency field should be hidden when updating the experience from contractor to some other type. 

## 📋 Steps to Reproduce

1. Create a government work experience with Contractor type with all required fields
2. Update that experience to Indeterminate for example
3. Observe the contractor field is not hidden 

## ✅ Acceptance Criteria

- [ ] Contractor firm or agency field should be hidden when updating the experience from contractor to some other type

esizer:
🤖 Resolves #12769 

## 👋 Introduction

Fixes an issue that occurred when changing the employment type was resetting fields to their default values rather than nulling them causing the form to go to unrealistic states.

## 🕵️ Details

When using `resetField` from `react-hook-form` it sets it back to the default value. So, when editing an existing experience it was resetting `govContractorType` to the one that was previously selected rather then setting it to `null`. Since this was the only condition to display that field, it remained visible and the user could submit a non-contractor experience with a firm or agency name.

Using `null` as the default value fro these resets fixes that issue.

## 🧪 Testing

1. Build `pnpm run dev:fresh`
2. Login as any role
3. Create a government contractor work experience that has a form or agency name
4. Edit that experience and select any other employment type
5. Confirm the contractor fields properly get set to `null` and are hidden

### DIFF
--- a/apps/web/src/components/ExperienceFormFields/WorkFields/GovFields.tsx
+++ b/apps/web/src/components/ExperienceFormFields/WorkFields/GovFields.tsx
@@ -210,26 +210,14 @@ const GovFields = ({ labels }: SubExperienceFormProps) => {
     if (watchGovEmploymentType !== WorkExperienceGovEmployeeType.Contractor) {
       resetDirtyField("govContractorRoleSeniority");
       resetDirtyField("govContractorType");
-    }
-
-    if (
-      watchGovContractorType === GovContractorType.SelfEmployed ||
-      watchGovEmploymentType !== WorkExperienceGovEmployeeType.Contractor
-    ) {
-      resetField("contractorFirmAgencyName", {
-        keepDirty: false,
-        defaultValue: undefined,
-      });
+      resetDirtyField("contractorFirmAgencyName");
     }
 
     // govPositionType field only applies to INDETERMINATE
     if (
       watchGovEmploymentType !== WorkExperienceGovEmployeeType.Indeterminate
     ) {
-      resetField("govPositionType", {
-        keepDirty: false,
-        defaultValue: null,
-      });
+      resetDirtyField("govPositionType");
     }
   }, [resetField, watchGovEmploymentType, watchGovContractorType]);
 
--- a/apps/web/src/components/ExperienceFormFields/WorkFields/GovFields.tsx
+++ b/apps/web/src/components/ExperienceFormFields/WorkFields/GovFields.tsx
@@ -210,14 +210,13 @@ const GovFields = ({ labels }: SubExperienceFormProps) => {
     if (watchGovEmploymentType !== WorkExperienceGovEmployeeType.Contractor) {
       resetDirtyField("govContractorRoleSeniority");
       resetDirtyField("govContractorType");
-      resetDirtyField("contractorFirmAgencyName");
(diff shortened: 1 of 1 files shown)


mnigh (changes requested):
It works better than before, but the value of **Contracting firm or agency** does not get reset to `null` after switching from Contractor to another option for **Employment type**. 

## 🎥  Screen recording

https://github.com/user-attachments/assets/58dff37b-54f8-4348-903e-599d70f45111

esizer:
> It works better than before, but the value of **Contracting firm or agency** does not get reset to `null` after switching from Contractor to another option for **Employment type**.

Good catch, I think I have it fixed now after [make reset more general](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12884/commits/4203ac88f7bb6e7e4dd6c41c6098fec4aab94dc7)

mnigh (commented):
This seems to work better now, thanks. Would another developer be able to run through the process as well? I am afraid changing the conditional logic in this complex form might have unintended consequences that I might have missed. Also, might a regression test (playwright) be helpful as well?

apps/web/src/components/ExperienceFormFields/WorkFields/GovFields.tsx
```diff
@@ -212,26 +210,14 @@ const GovFields = ({ labels }: SubExperienceFormProps) => {
...
-    }
-
-    if (
-      watchGovContractorType === GovContractorType.SelfEmployed ||
-      watchGovEmploymentType !== WorkExperienceGovEmployeeType.Contractor
-    ) {
-      resetField("contractorFirmAgencyName", {
-        keepDirty: false,
-        defaultValue: undefined,
-      });
```
vd1992, line 224: This shouldn't be removed entirely
Contractors can be self-employed or agency, but agency is the only one with a name field. So switching from agency to self-employed should also wipe the name field but not reset the other contractor fields

apps/web/src/components/ExperienceFormFields/WorkFields/GovFields.tsx
vd1992, line 1: I think we can get rid of one of the govPositionType resets, we accidentally had a redundancy 

image

apps/web/src/components/ExperienceFormFields/WorkFields/GovFields.tsx
esizer: Okay, this was adding back in [revert agency name condition, remove redundant type condition](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12884/commits/7f3a40104d3f00380959ad7df96ff8060b423f78)

brindasasi:
I need to check something with this branch in dev. Deploying it currently

brindasasi (approved):
Looks great! 👏🥳

PR #12956, opened 2025-03-06
[a11y] Include unread notification count in `aria-label`

### LINKED ISSUE #12891: ⌨ Notifications: Screen reader users are not informed of outstanding notifications
## ⌨ Accessibility Issue

When a screen reader user navigates to the notification button on the application, they are not informed of outstanding notifications. Visual users can see a small dot indicating a new notification, but this information is not conveyed to screen reader users, creating an unequal experience. Additionally, the dot could be made more prominent for sighted users.

### WCAG 2.1 Criteria

WCAG 1.3.1 Info and Relationships – Information, structure, and relationships conveyed visually must also be programmatically determined or available in text.

WCAG 4.1.2 Name, Role, Value – User interface components must provide meaningful information to assistive technologies.

## 🦋 Expected Behaviour

When a screen reader user navigates to the notification button, they should be informed of outstanding notifications (e.g., "Notifications (1 unread)"). This ensures equal access to information. Additionally, the visual indicator could be enhanced to make it more noticeable to sighted users.

## 🕵️ Details

- **Operating System**: MacOS
- **Browser**: Chrome
- **Assistive Technology**: Voiceover

## 📋 Steps to Reproduce

Tab to notification button from anywhere on the application.

## 📸 Screenshot

Image

## 🙋‍♀️ Proposed Solution

Update the `aria-label` to include a count of unread notifications when they exist.

```tsx
aria-label={intl.formatMessage({
  defaultMessage: "View notifications{count}",
  id: "ztx8xL",
  description: "Button text to open the notifications dialog",
}, {
  count: notificationCount > 0 ? intl.formatMessage({ 
    defaultMessage: " ({ count} unread)", { count: notificationCount }
  })  : ""
})}
```

## ✅ Acceptance Criteria

- [ ] Notification dialog indicates number of unread notifications to screen readers

esizer:
🤖 Resolves #12891 

## 👋 Introduction

Adds the number of unread notifications to the `aria-label` for the notification dialog trigger for screen readers.

## 🧪 Testing

1. Build `pnpm run dev:fresh`
2. Login
3. Confirim, with no unread notifications, the label remains the same
4. Create a notification for the logged in user
5. Confirm the label updates to include a count of new/unread notifications

### DIFF
--- a/apps/web/src/components/NotificationDialog/NotificationDialog.tsx
+++ b/apps/web/src/components/NotificationDialog/NotificationDialog.tsx
@@ -242,11 +242,26 @@ const NotificationDialog = ({
             mode={isSmallScreen ? "solid" : "icon_only"}
             color={isSmallScreen ? "black" : "secondary"}
             icon={notificationCount > 0 ? UnreadAlertBellIcon : BellAlertIconSm}
-            aria-label={intl.formatMessage({
-              defaultMessage: "View notifications",
-              id: "ztx8xL",
-              description: "Button text to open the notifications dialog",
-            })}
+            aria-label={intl.formatMessage(
+              {
+                defaultMessage: "View notifications{count}",
+                id: "l82MWI",
+                description: "Button text to open the notifications dialog",
+              },
+              {
+                count:
+                  notificationCount > 0
+                    ? ` ${intl.formatMessage(
+                        {
+                          defaultMessage: "({notificationCount} unread)",
+                          id: "Z22PPJ",
+                          description: "Number of unread notifications",
+                        },
+                        { notificationCount },
+                      )}`
+                    : "",
+              },
+            )}
             data-h2-margin-right="base:selectors[>*:first-child](-x.25) l-tablet:selectors[>*:first-child](0)"
             data-icon="true"
             {...(!isSmallScreen && linkColorStyling)}
--- a/apps/web/src/components/NotificationDialog/NotificationDialog.tsx
+++ b/apps/web/src/components/NotificationDialog/NotificationDialog.tsx
@@ -253,8 +253,11 @@ const NotificationDialog = ({
                   notificationCount > 0
                     ? ` ${intl.formatMessage(
                         {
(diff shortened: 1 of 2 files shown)


mnigh (commented):
This works for when the count is 0 or 1, but I can't test for more than 1 as the `count` that is using the `notificationCount` variable doesn't seem correct. The count should be 3 based on the number of unread messages rendered.

## Screenshot

Screen Shot 2025-03-07 at 13 40 08

Screen Shot 2025-03-07 at 13 44 54

mnigh:
@esizer I think the issue with count is that we only return the first one, `first: 1`. That will need to be removed and the comment updated.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/980ba5e8947095997a35e066125ab5ee74109738/apps/web/src/components/NotificationDialog/NotificationDialog.tsx#L29-L39

mnigh:
Also, noticed that when in French, VoiceOver reads out the `arial-label` in an English voice yet none of the other elements behave that way.

esizer:
> @esizer I think the issue with count is that we only return the first one, `first: 1`. That will need to be removed and the comment updated.

Huh... I'd really rather keep it to :sweat: @NienkeBr Do you tihnk it would be fine if we just said if there were any unread instead of the number?

If so, do you have copy to suggest? :sweat_smile: Maybe just:

> View notifications (unread)

NienkeBr:
If it's not possible to have the count, which is still what I'd prefer, I would suggest the following:
- View notifications (there are unread notifications)
- Voir les notifications (il y a des notifications non lues)

mnigh (approved):
LGTM 👍 

## Screenshots
<details><summary>Screenshots (4)</summary>
<p>

Screen Shot 2025-03-10 at 10 13 53

Screen Shot 2025-03-10 at 10 16 47

Screen Shot 2025-03-10 at 10 13 34

Screen Shot 2025-03-10 at 10 17 30

</p>
</details>

PR #12962, opened 2025-03-07
[Feat] Employee search input

### LINKED ISSUE #12910: ✨ Employee Search Component
## ✨ Feature

We have several forms that require a component to search for a user.  Let's create a reusable componet for it.

## 🎨 Design

@JoshBeveridge 

https://www.figma.com/design/k9lJ77FgLFvpO9L9tVyzID/Talent-nomination-(Employees)?node-id=118-46934&t=uChb4xPD86gGTmyr-4

## 📸 Screenshot

Image

## ✅ Acceptance Criteria

- [ ] Component can exist as part of a form
  - [ ] Contributes a user ID to a form state
  - [ ] Can be initialized in a form with a user ID
  - [ ] Has a configurable label
  - [ ] Icon button accessible description (Search work email)
  - [ ] multiple instances of this component can be part of a form together (consider uniqueness)
- [ ] Has several states:
  - [ ] (empty - no message underneath)
  - [ ] "Searching"
  - [ ] "We found a user with this email address"
  - [ ] "We couldn't find a matching profile"
  - [ ] "The email provided isn’t a valid Government of Canada email address"
  - [ ] "This user hasn’t verified their employee status" (not supported by the API yet)
- [ ] If a profile is found then it displays the employee profile name, role, and department
- [ ] light/dark and desktop/mobile styling
- [ ] Does search when clicking search button
- [ ] submit event does not propagate to outer form
- [ ] input and button disabled during search
- [ ] Changes to the component, with result messages, announced to AT

Out of scope:
- The "see information out of date or incorrect" well
- The "review" radiogroup

## 🛑 Blockers

```[tasklist]
### Blocked By
- [ ] #12787
```

esizer:
🤖 Resolves #12910 

## 👋 Introduction

Adds an employee search input to find employees and set the value to their user id.


apps/web/src/components/EmployeeSearchInput/ControlledInput.tsx
```diff
@@ -0,0 +1,226 @@
...
+            intl.formatMessage({
+              defaultMessage: "Search work email",
+              id: "pfwT4h",
+              description:
+                "Default button text to search for an employee by work email",
+            })
+          }
+          onClick={handleSearch}
+        >
+          <span data-h2-display="base(flex)" data-h2-align-items="base(cener)">
```
petertgiles, line 179:
```suggestion
          <span data-h2-display="base(flex)" data-h2-align-items="base(center)">
```

packages/fake-data/src/fakeUsers.ts
```diff
@@ -121,6 +121,7 @@ const generateUser = (
 
     // Gov info
     isGovEmployee: faker.datatype.boolean(),
+    workEmail: faker.internet.email(),
```
petertgiles, line 124: Maybe more realistic to do something like faker.userName() + "@gc.ca"

petertgiles: Same for L272

apps/web/src/components/EmployeeSearchInput/EmployeeSearchInput.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+interface WrapperProps
+  extends ComponentPropsWithoutRef<"div">,
+    HydrogenAttributes {}
+
+export interface EmployeeSearchInputProps
+  extends Omit<ComponentPropsWithoutRef<"input">, "id" | "name">,
+    Omit<CommonInputProps, "context"> {
+  buttonLabel?: string;
+  wrapperProps: WrapperProps;
```
petertgiles, line 24: Is this a mandatory prop?  It doesn't appear in your example.  It doesn't seem like it should be mandatory.

apps/web/src/components/EmployeeSearchInput/ControlledInput.tsx
```diff
@@ -0,0 +1,226 @@
...
+  formState: UseFormStateReturn<FieldValues>;
+  /** Current field state (to update styles) */
+  fieldState: FieldState;
+  inputProps?: Record<string, string>;
+  buttonLabel?: string;
+  describedBy?: string;
+  errorMessages?: Partial<ErrorMessages>;
+}
+
+const ControlledInput = ({
```
petertgiles, line 53: It looks like it's not possible to initialize one of these with a default value.
image

petertgiles: It also likes like it contributes a lot more to the form state than just the ID.  It's not necessarily a problem but it does make it awkward to figure out how to initialize the control with data, like above.  And a little trickier in the "formValuesToApiInput" function.
image

petertgiles (changes requested):
So cool!  Thanks for taking on this complicated control.  :bow: I started playing around with it and noticed a few things to check on.

apps/web/src/components/EmployeeSearchInput/ControlledInput.tsx
esizer: Okay, so I can :100: change this and had it working in a different way but I was not happy with it.

In order for it to just use ID and be initialized by only an ID, we need to make an API request on mount. I was'nt a huge fan of this but am very willing to go back because I can see how needing the email also has it's cons.

packages/fake-data/src/fakeUsers.ts
esizer: Great idea, updated in [improve fake work email](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12962/commits/2307d8f213863b7040510592055cd1080807eca0)

apps/web/src/components/EmployeeSearchInput/ControlledInput.tsx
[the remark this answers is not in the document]
esizer: Oops :sweat_smile: Fixed as part of [fix up styles](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12962/commits/3d5331c97da1e0bcc2e501fbccef67fab4f2f481)

apps/web/src/components/EmployeeSearchInput/ControlledInput.tsx
```diff
@@ -0,0 +1,226 @@
...
+          data-h2-flex-grow="base(1)"
+          data-h2-border-width="base(0)"
+          data-h2-radius="base(rounded 0 0 rounded)"
+          onChange={handleChange}
+          onKeyDown={handleKeyDown}
+        />
+        <Button
+          type="button"
+          mode="solid"
+          color="primary"
```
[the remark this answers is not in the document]
esizer: Oh, I could have sworn I saw purple! Anyway, adjust in [fix up styles](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12962/commits/3d5331c97da1e0bcc2e501fbccef67fab4f2f481)

apps/web/src/components/EmployeeSearchInput/ControlledInput.tsx
petertgiles: Hmm, good point.  Running a query on mount is not ideal.  I'm open to suggestions.  :sweat_smile:

esizer: Actually.. Looks like I couldn't get that to work and am running a query on mount :sweat: The problem came up that we need an email to run the query, not an ID so that is why it is included :sweat_smile: 

The only thing I can think of is accepting a fragment in addition to the default value from `react-hook-form` :thinking: 

```tsx
interface EmployeeSearchInputProps {
  defaultValueQuery?: FragmentType<typeof EmployeeSearchResult_Fragment>;
}

const employee = defaultValueQuery ?? data?.employeeProfile;

const methods = useForm({
  defaultValues: {
    nominee: data.talentNomination.nominee.id
  }
}):

<EmployeeSearchInput name="nominee" defaultValueQuery={data.talentNomination.nominee.employeeProfile} />
```

apps/web/src/lang/fr.json
```diff
@@ -823,6 +823,10 @@
     "defaultMessage": "Vous avez un profil sur Talents numériques du GC.",
     "description": "Second item in list of eligibility requirements section"
   },
+  "2ExR2X": {
+    "defaultMessage": "Il semble que {nom} ait déjà un compte chez nous. Nous avons rempli automatiquement ses renseignements sur la base de son profil. Une fois que vous aurez soumis cette candidature, nous l'informerons et il ou elle pourra consulter les détails de la mise en candidature et l'état d'avancement. »",
```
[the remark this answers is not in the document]
petertgiles: Oops, check again.  The variable is still in French for 2ExR2X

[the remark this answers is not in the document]
esizer: heh, idk what happened there :sweat_smile: I promise it is actually fixed now after [fix translation variable](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12962/commits/c092c01d36b6ed0fab56c40158574f6564c1d8bd)

apps/web/src/components/EmployeeSearchInput/ControlledInput.tsx
esizer: okay, I have something working after [update to use plain ID value](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12962/commits/d09b8120b88805a2c445806f685bc04c2e8c0d88)

Usage updates to be:

```tsx
import EmployeeSearchInput from "~/components/EmployeeSearchInput/EmployeeSearchInput";

// Now accepts only a UUID
type FormValues {
  nominee: Scalars["UUID"]["input"];
}

// Now you get the "options" with a fragment
const Form_Fragment = graphql(
  const Form_Fragment on User {
    id
    govEmployeeProfile {
      ...EmployeeSearchDefaultValue
    }
  }
)l

const Form = (...) => {
  const user = getFragment(Form_Fragment, formQuery);
  const methods = useForm<FormValues>({
     defaultValues: {
        nominee: user?.id, // Just the ID now
     }
  })
  
  const handleSubmit = (values: FormValues) => {
    // { nominee: "UUID" }
    onSubmit({ nominee: nominee.id });
  }
  
  // Now passes fragment into, like you select "options"
  return (
    <FormProvider {...methods}>
      <form onsubmit={methods.handleSubmit(handleSubmit)}>
         <EmployeeSearchInput 
              id="nomiee" 
              name="nomnee" 
              label="Nomiee" 
              employeeQuery={user.govEmployeeProfile}
            />
      </form>
    </FormProvilder>
  );
}
```

If you pass a user inot the fragment that does not match the default for value, it does not populate.

petertgiles: Very cool!  In general I think this is the right approach.  Three impressions from playing around with it:

1) Using a graphql fragment type for the new prop feels a bit specific and unusual when compared to other generic form components like Select.  I kind of want to see a prop that's just an object with a bunch of string fields.  But maybe this is totally unfair since the component literally fires off its own graphql queries?  :laughing: 

2) It would be nice if the prop accepted null since usually things coming back from the api are nullable.  It avoids having to do things like `employeeQuery={ applicantDashboardQuery.me?.govEmployeeProfile ?? undefined }`

3) If you initialize a value you get a weird in between state where it is telling you to search and that something was found at the same time.
image

apps/web/src/components/EmployeeSearchInput/EmployeeSearchInput.tsx
petertgiles, line 1: It seems like the work email validation error message isn't working now?
image

apps/web/src/components/EmployeeSearchInput/ControlledInput.tsx
petertgiles: This is a bit nuanced and open to opinion...  if the form is initialized with an ID and I do a search I would expect the form value to take the new ID if it succeeds or go to null if it fails.  Right now the old value sticks if a new search fails which is unexpected (at least to me).

esizer: Okay! Thanks for helping me with this. I have completed the small refactor we discussed. 

Usage wise, everything stays the same but the prop you pass in is now a plain object. I implemented the helper you suggested so we can still use the expected fragment but transform it into the expected shape.

```tsx
import EmployeeSearchInput, { fragmentToEmployee } from "~/components/EmployeeSearchInput/EmployeeSearchInput";

// Now you get the "options" with a fragment
const Form_Fragment = graphql(
  const Form_Fragment on User {
    id
    govEmployeeProfile {
      ...EmployeeSearchDefaultValue
    }
  }
)l

  // Now passes fragment into, like you select "options"
  return (
    <EmployeeSearchInput 
       // {...props}
       employee={fragmentToEmployee(user.govEmployeeProfile)}
    />
  );
}
```

apps/web/src/components/EmployeeSearchInput/EmployeeSearchInput.tsx
esizer: I tihnk the refactor fixed this.

apps/web/src/components/EmployeeSearchInput/ControlledInput.tsx
esizer: Also,

1. Thanks for the sugestion, I also tihnk the object makes things a bit more conventional compared to other inputs so it has been updated
2. It also now accepts null :+1: 
3. The null message should be fixed now as well
4. If an error or empty response comes back it now properly nulls out the value

petertgiles: Nice, this is working great now!  The last thing that doesn't feel quite right is the name of the prop.  It appears like this is going to set the result of control rather than just preloading a option.  Maybe `employeeOption` is a little clearer?

apps/web/src/components/EmployeeSearchInput/EmployeeSearchInput.tsx
petertgiles: It looks like the "not a government email" is being missed by the control and bubbling up to a toast now.
image

apps/web/src/components/EmployeeSearchInput/ControlledInput.tsx
petertgiles: Open to suggestions though... :laughing:

esizer: I was also struggling with this so I tihnk your suggestion works :stuck_out_tongue: 

Updated in [rename prop for employee](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12962/commits/2c6e08a62a01b7ac859ec3041359414f84bf1c2b)

apps/web/src/components/EmployeeSearchInput/EmployeeSearchInput.tsx
esizer: Good call, I went ahead and made some changes to fix both the error under the input and prevent the query if it is not a gov email in [fix errors](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12962/commits/1c6ab28c5e7ae8b8152e6f69e0203fe2f3167784)

petertgiles: OK, nice - I am getting the "not a government email" error now.  It doesn't clear out the form backing value but I don't mind that, tbh.  The UX makes it pretty clear a search never happened so it kind of makes sense.

A bigger problem is that I'm not getting the "profile not found" message any more.  If no user comes back I'm getting the "enter a work email" message.
image

esizer: Alright, hopefully this finally works :sweat_smile: 

I updated how the state and errors work in [fix state, error handling](https://github.com/GCTC-NTGC/gc-digital-talent/pull/12962/commits/f243da36b184e87b46dc8a7bc052af82dfc50011). Also, it now updates the form value every time you attempt a search.

petertgiles (approved):
Amazing!

PR #12977, opened 2025-03-12
[A11y] Reload document on current breadcrumb activation

### LINKED ISSUE #12901: ⌨ Breadcrumbs: Current page lacks feedback/functionality
## ⌨ Accessibility Issue

The current page in the breadcrumbs is coded as a clickable link (<a> tag). However, selecting the link does nothing, leading to a confusing user experience.

### WCAG 2.1 Criteria

WCAG 4.1.2 Name, Role, Value – Interactive elements must have a name, role, and value that is communicated to assistive technology. If an element is coded as a link, users expect it to perform a navigational action.

Heuristic: Consistency and Standards – Users expect links to be actionable. If a link is not intended to be clickable, it should not be coded as one.

## 🦋 Expected Behaviour

If the current page in the breadcrumbs is not meant to be navigable, it should be presented as plain text (e.g., a `<span>` or `<li>` without a link).

Alternatively, if the design choice is to keep it as a link, then the page should reload when the link is activated, maintaining consistent behavior expected by users.

Refer to [Breadcrumb Example | APG | WAI | W3C](https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/) for best practices on breadcrumb navigation.

## 🕵️ Details

This is likely to do with `react-router` `Link` element. Since we do not have true navigation being an SPA, `react-router` sees you are on the page and is probably "saving work/time" by not triggering a navigation event so it actually isn't doing anything.

Activating a link for the current page is another method of triggering a refresh so we should probably force a navigation on this event even if nothing may happen since that would be expected behaviour.

## 📋 Steps to Reproduce

Select the current page link on the breadcrumbs.

## 📸 Screenshot

Image

## 🙋‍♀️ Proposed Solution

Update the current page breadcrumb link to force a navigation event.
Eric thinks this is something we can force React Router to do.
We don't want to reboot the app when this happens.

## ✅ Acceptance Criteria

- [ ] Activating the current page link our breadcrumbs triggers a navigation event
- [ ] User is notified something happened

esizer:
🤖 Resolves #12901 

## 👋 Introduction

Forces the document to reload when activating the current page breadcrumb.

## 🧪 Testing

1. Build `pnpm run dev:fresh`
2. Navigate to a page with breadcrumbs i.e `/browse/pools`
3. Turn on a screen reader
4. Activate the current breadcrumb
5. Confirm the page is reloaded and the heading is announced

### DIFF
--- a/packages/ui/src/components/Breadcrumbs/Crumb.tsx
+++ b/packages/ui/src/components/Breadcrumbs/Crumb.tsx
@@ -21,6 +21,7 @@ const Crumb = ({ children, isCurrent, url }: CrumbProps) => (
             "data-h2-font-weight": "base(700)",
             "data-h2-text-decoration": "base(none)",
             "aria-current": "page",
+            reloadDocument: true,
           }
         : {})}
     >


mnigh (approved):
Tested using VoiceOver on Chrome and Firefox. Upon activating the current breadcrumb, the page is reloaded and the heading is announced.

PR #13006, opened 2025-03-13
[Task]  CreateDevelopmentProgramsSeeder 

brindasasi:
🤖 Resolves 

## 👋 Introduction
We got these development programs into UAT for finance community and eventually it is going to be in prod. 
So let's get that seeder into repo as well. 

[Development_program_and_designation_template - Complete.xlsx](https://github.com/user-attachments/files/19252522/Development_program_and_designation_template.-.Complete.xlsx)

## 🕵️ Details

This PR just creates a seeder for the attached excel file.


mnigh:
@brindasasi is this PR still being worked on or can it be closed?

brindasasi:
This should be reviewed .  Will put it up shortly

brindasasi:
I change my mind on this PR. This data is very specific to prod and finance community.
I don't think we should keep this in repo. We wouldn't re-use this for other communities.

PR #13033, opened 2025-03-19
[Feature] Create Talent Nomination Page

### LINKED ISSUE #13005: ✨ Improve create nomination experience
## ✨ Feature

The experience to create a nomination should be improved.  We will model this on the "start application" experience for visitors.

## 📸 Screenshot

Image

## ✅ Acceptance Criteria

- [ ] The "talent management events" page is visitable to any visitor (already the case)
- [ ] ~In the API, any logged-in user can create a new nomination as the submitter (need to remove verified employee restriction in the policy)~
- [ ] If a user clicks "start" when not logged in then they are redirected to login and nomination flow starts automatically when they return (like current application process)
- [ ] If a user clicks "start" when logged in then the nomination flow starts as expected
- [ ] If a user clicks "start" and there is an error from the API then an error toast is displayed and they remain on the talent events page (already the case)

## 🛑 Blockers

```[tasklist]
### Blocked By
```

mnigh:
🤖 Resolves #13005.

## 👋 Introduction

This PR adds a `CreateTalentNominationPage` component that works in a similar way to the `CreateApplicationPage` that is renders nothing but adds logic to force a redirect to the login page for non-logged in users.

## 🧪 Testing

1. `pnpm build:fresh`
3. Navigate to http://localhost:8000/en/communities/talent-events
4. Logout so no user is logged in
5. Press **Start a nomination** link
6. Verify redirect to login page
7. Login as applicant-employee@test.com
8. Verify nomination is created and nomination form instructions page is navigated to
9. Navigate to http://localhost:8000/en/communities/talent-events
10. Login as applicant-employee@test.com
11. Press **Start a nomination** link
12. Verify nomination is created and nomination form instructions page is navigated to

### DIFF
--- a/apps/web/src/components/Router.tsx
+++ b/apps/web/src/components/Router.tsx
@@ -66,10 +66,22 @@ const createRoute = (locale: Locales, newApplicantDashboard: boolean) =>
                 },
                 {
                   path: "talent-events",
-                  lazy: () =>
-                    import(
-                      "../pages/TalentManagementEventsPage/TalentManagementEventsPage"
-                    ),
+                  children: [
+                    {
+                      index: true,
+                      lazy: () =>
+                        import(
+                          "../pages/TalentManagementEventsPage/TalentManagementEventsPage"
+                        ),
+                    },
+                    {
+                      path: ":nominationEventId/create-talent-nomination",
+                      lazy: () =>
+                        import(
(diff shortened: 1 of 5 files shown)


esizer (approved):
LGTM! :+1:

PR #13028, opened 2025-03-18
[Feat] Add nominator section

esizer:
🤖 Resolves #12318 

## 👋 Introduction

Adds nominator fields to the talent nomination.

## 🧪 Testing

1. Build `pnpm run dev:fresh`
2. Login as any role
3. Start a nomination
4. Fill out the nominator page
5. Excersize the conditional logic of:
   - You being the nominator
   - Submit on behalf of a government employee in the system,
   - Submitting on behalf of a fallback government employee

PR #13019, opened 2025-03-18
[Feat] Nominee section of nomination

esizer:
🤖 Resolves #12319 

## 👋 Introduction

Adds the fields for the nominee section to the nomination wizard.

## 🧪 Testing

1. Build `pnpm run dev:fresh`
2. Start a nomination
3. Continue to the nominee section
4. Complete the form and confirm it has the proper result of conditional logic

PR #13053, opened 2025-03-20
[Fix] Display form errors for employee search input

### LINKED ISSUE #13018: 🐛 `EmployeeSearchInput` does not display required errors
## 🐛 Bug

The new employee search input can be set to required but does not display the error when the user attempts to submit the form without setting an employee properly.

## 🦋 Expected Behaviour

An error appears when a form submission is attempted with no value and the field is required.

## 📋 Steps to Reproduce

1. Go to a form with a required employee search input (talent nomination nomee section)
2.  Attempt to submit the form without filling out the input
3. Observe that it does not submit but no error is displayed

## 🙋‍♀️ Proposed Solution

There is some error handling but maybe we need to include a catch all since it only deals with specific errors. Look at our generic `Input` component and maybe use that approach? 🤔 

## ✅ Acceptance Criteria

- [ ] `EmployeeSearchInput` properly displays a required error

esizer:
🤖 Resolves #13018 

## 👋 Introduction

Updates the `EmployeeSearchInput` to display errors that come from the `rules` prop.

## 🧪 Testing

1. Run storybook `pnpm run storybook`
2. Navigate tot he "Required" story for employee searchng input
3. Try submitting with not filling it out
4. Confirm the required error appears

### DIFF
--- a/apps/web/src/components/EmployeeSearchInput/EmployeeSearchInput.stories.tsx
+++ b/apps/web/src/components/EmployeeSearchInput/EmployeeSearchInput.stories.tsx
@@ -82,6 +82,21 @@ export const NoResult: Story = {
   },
 };
 
+export const Required: Story = {
+  args: {
+    rules: { required: "This field is required" },
+  },
+  parameters: {
+    apiResponses: {
+      EmployeeSearch: {
+        data: {
+          govEmployeeProfile: null,
+        },
+      },
+    },
+  },
+};
+
 export const NotGovernmentEmail: Story = {
   parameters: {
     apiResponses: {
--- a/apps/web/src/components/EmployeeSearchInput/ControlledInput.tsx
+++ b/apps/web/src/components/EmployeeSearchInput/ControlledInput.tsx
@@ -11,6 +11,7 @@ import { useIntl } from "react-intl";
 import {
   FieldState,
   useCommonInputStyles,
+  useFieldStateStyles,
   useInputDescribedBy,
 } from "@gc-digital-talent/forms";
 import { graphql, Maybe } from "@gc-digital-talent/graphql";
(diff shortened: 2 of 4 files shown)


apps/web/src/components/EmployeeSearchInput/EmployeeSearchInput.stories.tsx
```diff
@@ -82,6 +82,21 @@ export const NoResult: Story = {
   },
 };
 
+export const Required: Story = {
```
mnigh, line 85: The [Chromatic snapshot](https://www.chromatic.com/component?appId=61e099a1bb1465003a73d3ce&csfId=components-employeesearchinput&specName=Required&buildNumber=10617&k=67dc0d957aad52dba614f403-1200px-snapshot-true&h=4&b=-2) doesn't capture the state with the error. I don't think we have used it yet, but would the [play function](https://storybook.js.org/docs/writing-stories/play-function#writing-stories-with-the-play-function) make sense for this one?

mnigh (commented):
Hard to tell if this works as intended with the story.

## Screenshot

Screen Shot 2025-03-20 at 09 57 30

apps/web/src/components/EmployeeSearchInput/EmployeeSearchInput.stories.tsx
esizer: Great suggestion. Should be better now after [submit form for error display](https://github.com/GCTC-NTGC/gc-digital-talent/pull/13053/commits/bd82436ab737a78f63f0d918b152ecef50c22a01)

esizer:
> @JoshBeveridge I looked for a design on this but couldn't find one. Are you able to mock something up for us or are the default error styles fine?

Response: Nothing custom for form errors. Use the default we have.

mnigh (approved):
@esizer This looks good to me in code. Since the component is not actually implemented anywhere but a storybook story, it is hard to test its functionality properly. I will approve but this work should be merged into the #13019 and #13028 and tested there as well and if there are any issues with it, an issue created to fix.

PR #13076, opened 2025-03-24
[Debt] Improve mock sign in template

### LINKED ISSUE #13075: ♻️ Improve mock sign in styles
## ♻️ Debt/Refactor

Our mock sign in page styles broke with our CSP. Would be nice to have it look not so awful.

## 🙋‍♀️ Proposed Solution

You can provide a template to use for sign in, do that but provide our own styles.

## ✅ Acceptance Criteria

- [ ] Mock auth doesn't look terrible

esizer:
🤖 Resolves #13075 

## 👋 Introduction

This just applies some styles to the mock auth now that it is no longer loading from a third party.

## 🧪 Testing

1. Rebuild docker `make down; make up;`
2. Go to the sign in
3. Confirm it looks not awful

### DIFF
--- a/infrastructure/mockauth/resources/login.html
+++ b/infrastructure/mockauth/resources/login.html
@@ -1,0 +1,101 @@
+<!doctype html>
+<html lang="en">
+
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+  <title>GC Digital Talent Mock Sign In</title>
+  <link rel="preconnect" href="https://fonts.googleapis.com" />
+  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+  <link
+    href="https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300;0,400;0,500;0,600;0,700;0,800;1,300;1,400;1,500;1,600;1,700;1,800&display=swap"
+    rel="stylesheet" />
+  <style nonce="**CSP_NONCE**">
+    :root {
+      --rounded: 0.375rem;
+      --black: 30, 30, 35;
+      --primary: 0, 195, 183;
+      --focus: 100, 146, 255;
+      --space: 0.625rem 1.3rem;
+    }
+
+    body {
+      background-color: rgba(245, 248, 253, 1);
+      font-size: 100%;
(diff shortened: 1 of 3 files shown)


infrastructure/mockauth/resources/login.html
```diff
@@ -0,0 +1,101 @@
...
+<html lang="en">
+
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+  <title>GC Digital Talent Mock Sign In</title>
+  <link rel="preconnect" href="https://fonts.googleapis.com" />
+  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+  <link
+    href="https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300;0,400;0,500;0,600;0,700;0,800;1,300;1,400;1,500;1,600;1,700;1,800&display=swap"
```
mnigh, line 11: Do we need all of these weights?

mnigh (commented):
I can confirm that it looks not awful. Do we care about other states like an error? There seem to be several templates that exist: https://github.com/navikt/mock-oauth2-server/tree/master/src/main/resources/templates.

## Screen recording
https://github.com/user-attachments/assets/0000d002-c078-4a41-9716-3642ec549238

infrastructure/mockauth/resources/login.html
esizer: Since it is a variabel font, it kind of comes with all the weights baked in. Though, I did make the call simpler in [reduce/fix fonts, add labels/required](https://github.com/GCTC-NTGC/gc-digital-talent/pull/13076/commits/f679f77539894f6009755e3bdce80ac72cf7fb75)

Also, added labels/required to hopefully prevent the error :thinking:

mnigh: The `required` attribute does stop someone from easily getting to the error template with an empty form value, nice work. I was able to get there with removing the `required` attribute manually, but it seems unlikely anyone will make it that far unwillingly.

mnigh (approved):
:shipit:

infrastructure/mockauth/resources/login.html
esizer: Yeah, sadly we do not have access to the server side validation :cry:

PR #13198, opened 2025-04-08
Fix talent events nested path

vd1992:
🤖 Resolves 

## 👋 Introduction

Fix the router issue enabling accessing event pages again

## 🧪 Testing

1. Can navigate to `/en/admin/talent-events/:eventId` from the table again
2. Table and subsequent nested paths accessible


vd1992:
@esizer 
This is basically the exact same router fix from the other PR split off into its own thing as was requested
I've tagged you as it should be basically the same as my other PR

brindasasi (approved):
Fixes the navigation issue.. Thanks!

PR #13244, opened 2025-04-14
[Debt] Bump `ts-jest` to latest

### LINKED ISSUE #11364: ♻️ Bump `ts-jest` from 29.2.3 to latest

esizer:
🤖 Resolves #11364 

## 👋 Introduction

This bumps `ts-jest` and fixes the issues we were having with newer version.

## 🕵️ Details

The trick was to turn `isolatedModules` on. I'm still not sure why, I'll update when I have more info but the rules most people seem to be following are:

 - Turn on `isolatedModules` which will warn about issues with transpiling
 - Avoid the `import * as` statements
 - Don't use `ts-jest` :sweat_smile: 

## 🧪 Testing

> [!IMPORTANT]
> Run these steps on main and this branch to compare!

1. install deps `pnpm i`
2. Run  jest with heap logging
3. `pnpm run test --force -- --runInBand --logHeapUsage`
4. Oberseve the heap growing

You should see that on main it is significantly higher and doesn't seem to ever get GC'd. After bumping `ts-jest` it should actually be a reasonable numvber and go up and down which indicates it is getting GC'd.

It still seems kind of ihgh for our use case and migrating to `vitest` is probably a good idea.


petertgiles (approved):
Wow, like night and day.  I salute you, sir.

PR #13247, opened 2025-04-14
[Hail Mary] Try to fix the talent nomination querying issue

vd1992:
🤖 Resolves 

## 👋 Introduction

Local experimentation and it appears the `@canQuery` breaks things for me for who knows what reason

## 🧪 Testing

1. Break local environment
2. Switch to branch
3. See if it fixes it

To break locally, I seeded 50,000 users

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -671,14 +671,14 @@ type TalentNomination {
   submittedAt: DateTime @rename(attribute: "submitted_at")
   submitter: BasicGovEmployeeProfile
     @belongsTo
-    @canQuery(ability: "viewAnyBasicGovEmployeeProfile", model: "User")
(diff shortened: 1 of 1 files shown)


brindasasi:
I don't understand how this fixes the problem but reproduced the issue on UAT with 5000 users and applying this fix fixed that. Also verified on 35k users. Might test more to match with prod.

petertgiles (approved):
Wow, another :dart: 

From the canQuery documention:
> Query for specific model instances to check the policy against, using arguments
with directives that add constraints to the query builder, such as `@eq`.

:face_exhaling: I guess it was querying every single user in the table?  :facepalm: I think canModel is perfect, since it doesn't depend on the specific User being returned.  The scope is in the relationship.  I suppose the other solution would be to move the scope into the canQuery directive, but this is good, too.

Works great for me. :1st_place_medal:

PR #13378, opened 2025-05-01
[Debt] Remove form watch type casts

### LINKED ISSUE #11371: ♻️ Avoid type casting for `react-hook-form` watchers
## ♻️ Debt/Refactor

Replace type casting of `react-hook-form` watchers to improve type safety.

## 🕵️ Details

We are currently type casting anywhere we use the `react-hook-form` watch methods (`watch`, `useWatch`). This was because we were pretty certain what they were and wanted to get in the typescript type linting in.

However, since `react-hook-form` is written in typescript, I feel like we could avoid this. 

## 🙋‍♀️ Proposed Solution

I believe these are generics so we should be able to just pass in the `FormValues` to restrict the types and not need a cast.

Methods to look for:

- `useFormContext`
- `useWatch`
- `watch`

### Example

```tsx
// From
const value = useWatch('name') as FormValues;

// To
const value = useWatch<FormValues>('name');
const methods = useFormContext<FormValues>();
const value = methods.watch('name');
```

## ✅ Acceptance Criteria

- [ ] No more type casting for `react-hook-form` watchers
- [ ] Watch return values have type safety

esizer:
🤖 Resolves #11371 

## 👋 Introduction

Updates the type casting for form watchers to prefer the use of the generic.

## 🧪 Testing

1. Confirm no more type casts around `useWatch` and `watch` form methods
2. Confirm typescript is still happy

### DIFF
--- a/apps/web/src/components/SkillBrowser/SkillBrowser.tsx
+++ b/apps/web/src/components/SkillBrowser/SkillBrowser.tsx
@@ -33,11 +33,11 @@ const SkillBrowser = ({
     category: `${id}-${INPUT_NAME.CATEGORY}`,
     family: `${id}-${INPUT_NAME.FAMILY}`,
   };
-  const { watch, resetField, setValue } = useFormContext();
-  const [family, skillValue] = watch([inputNames.family, name]) as [
-    FormValues["family"],
-    FormValues["skill"],
-  ];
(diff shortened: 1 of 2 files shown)


vd1992 (approved):
Seems to look fine to me

PR #13379, opened 2025-05-01
[Feat] New helper for sorting arrays of complex object alphabetically

### LINKED ISSUE #13063: ✨ Add alpha sort helper
## ✨ Feature

It could be nice to have a simple helper function to help sort our API responses which we seem to be doing more often.

I have something you can use, maybe implement an order too?

```tsx
// A good start
function sortAlphaBy<T extends object>(accessor: (value: T) => string): (a: T, b: T) => number {
    return (a, b) => accessor(a).localeCompare(accessor(b))
}

// Usage
type ApiResponse = {
    related?: {
        name?: {
            localized: string;
        };
    }
};

const data: ApiResponse[] = [ ... ];

const sorted = data.sort(
    sortAlphaBy(
        (item => data.related?.name?.localized ?? "")
    )
);

```

## ✅ Acceptance Criteria

- [ ] Add in the helper function to `@gc-digital-talent/helpers`
- [ ] Implement it in places we sort alphabetically

esizer:
🤖 Resolves #13063 

## 👋 Introduction

Adds a helper to sort arrays of objects alphabetically by a nested key.

## 🧪 Testing

1. Build `pnpm run dev:fresh`
2. Confirm replacements are still sorted properly

### DIFF
--- a/packages/helpers/src/utils/sort.test.ts
+++ b/packages/helpers/src/utils/sort.test.ts
@@ -1,0 +1,65 @@
+import { sortAlphaBy } from "./sort";
+
+const values = [
+  {
+    name: {
+      localized: "HHH",
+    },
+  },
+  {
+    name: {
+      localized: "ZZZ",
+    },
+  },
+  {
+    name: {
+      localized: "AAA",
+    },
+  },
+];
+
+describe("Sort array of objects alphabetically", () => {
+  it("sorts ascending", () => {
+    const newValues = values.sort(sortAlphaBy((x) => x.name.localized, "asc"));
+
+    expect(newValues).toEqual([
+      {
+        name: {
+          localized: "AAA",
+        },
+      },
+      {
+        name: {
+          localized: "HHH",
+        },
+      },
+      {
+        name: {
+          localized: "ZZZ",
+        },
+      },
+    ]);
+  });
+
+  it("sorts descending", () => {
+    const newValues = values.sort(sortAlphaBy((x) => x.name.localized, "desc"));
+
+    expect(newValues).toEqual([
+      {
+        name: {
+          localized: "ZZZ",
+        },
+      },
+      {
+        name: {
+          localized: "HHH",
+        },
+      },
+      {
+        name: {
+          localized: "AAA",
+        },
+      },
+    ]);
+  });
+});
--- a/packages/helpers/src/utils/sort.test.ts
+++ b/packages/helpers/src/utils/sort.test.ts
@@ -1,65 +1,34 @@
 import { sortAlphaBy } from "./sort";
-
-const values = [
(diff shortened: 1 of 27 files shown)


packages/helpers/src/utils/sort.ts
```diff
@@ -0,0 +1,25 @@
...
+  let aValue = a;
+  let bValue = b;
+  if (order === "desc") {
+    aValue = b;
+    bValue = a;
+  }
+
+  if (!aValue || !bValue) return 0;
+
+  return aValue.localeCompare(bValue);
```
yonikid15, line 17:
```suggestion
  if (!a || !b) return 0;

  const aValue = order === "desc" ? b : a;
  const bValue = order === "desc" ? a : b;

  return aValue.localeCompare(bValue);
```

packages/helpers/src/utils/sort.test.ts
```diff
@@ -0,0 +1,65 @@
...
+        },
+      },
+      {
+        name: {
+          localized: "AAA",
+        },
+      },
+    ]);
+  });
+});
```
yonikid15, line 65:
```suggestion
import { sortAlphaBy } from "./sort";
const HHH = {
  name: {
    localized: "HHH",
  },
};
const ZZZ = {
  name: {
    localized: "ZZZ",
  },
};
const AAA = {
  name: {
    localized: "AAA",
  },
};

describe("Sort array of objects alphabetically", () => {
  it("sorts ascending", () => {
    const newValues = [HHH, ZZZ, AAA].sort(
      sortAlphaBy((x) => x.name.localized, "asc"),
    );

    expect(newValues).toEqual([AAA, HHH, ZZZ]);
  });

  it("sorts descending", () => {
    const newValues = [HHH, ZZZ, AAA].sort(
      sortAlphaBy((x) => x.name.localized, "asc"),
    );

    expect(newValues).toEqual([ZZZ, HHH, AAA]);
  });
});
```

yonikid15 (changes requested):
Nice work so far. Just a couple suggestions to minimise code and keep things cleaner.  Also, I see a there's some other spots we could use this function:
image

esizer:
> Nice work so far. Just a couple suggestions to minimise code and keep things cleaner. Also, I see a there's some other spots we could use this function:

I updated one but the others were intentionally left out since we wanted to keep the scope smaller and only address really straightforward ones first.


[fix missed sort](https://github.com/GCTC-NTGC/gc-digital-talent/pull/13379/commits/2123260ab95d9b2017213fe78a85a6af69b17572)

yonikid15 (approved):
Looking mighty good :+1: My bad about that suggestion on the test, but looks like you fixed it

PR #13381, opened 2025-05-01
[Fix] Missing error messages on assessment plan skills inputs

### LINKED ISSUE #12540: 🐛 Error well duplicated on form Add new assessment
## 🐛 Bug

Error well appears twice for some reason, it should only render the one

## 📋 Steps to Reproduce

1. Navigate to `/en/admin/pools/:id/plan` for an unpublished pool
2. Attempt to add an assessment step, without selecting any skills, observe form error

## 📸 Screenshot

Image

## ✅ Acceptance Criteria

- [ ] Form error state works as it does elsewhere and does not render extra wells

esizer:
🤖 Resolves #12540 

## 👋 Introduction

Fixes an issue where empty error messages were appearing for the skill selection in the assessment plan dialogs

## 🕵️ Details

This is not a great solution but maybe works for now? :woman_shrugging: 

The issue is that we have two separate inputs that are essentially collecting for a single data point so at least one of the inputs is required.

To be valid you must select at least one skill from either input. The ideal fix would be to use a single input instead of two separate ones that are combined in the validation and submission, however that will need come designer input so I went with this solution for now :confused: 

> [!NOTE]
> The new error message was provided by @NienkeBr since "required" was not specific enough and could cause confusion. 

## 🧪 Testing

1. Build `pnpm run dev:fresh`
2. Login as an admin `admin@test.com`
3. Create a process
4. Select both essential and asset skills
5. Navigate to the assessment plan edit page
6. Add a new step
7. Do not check any skills
8. Attempt to submit the form
9. Observe the error messages appearing in the proper locations

### DIFF
--- a/apps/web/src/lang/fr.json
+++ b/apps/web/src/lang/fr.json
@@ -9446,6 +9446,10 @@
     "defaultMessage": "Dernière mise à jour du profil de la personne candidate : {lastUpdated}",
     "description": "Last time user's profile was updated"
   },
+  "gwpmTn": {
+    "defaultMessage": "Veuillez sélectionner au moins une compétence essentielle ou une compétence constituant un atout pour cette méthode d'évaluation.",
+    "description": "Error message for when a skill is not selected for assessment"
+  },
   "gyZV/3": {
     "defaultMessage": "Veuillez ajouter au moins une expérience.",
     "description": "Error message if there are no experiences"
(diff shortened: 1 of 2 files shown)


vd1992:
Not really a big deal, but just throwing out an opinion. The same error message appearing three times seems a bit much. Given it is truly only one error, check off one skill in either section to make it go away

esizer:
> Not really a big deal, but just throwing out an opinion. The same error message appearing three times seems a bit much. Given it is truly only one error, check off one skill in either section to make it go away

I agree, however we discussed this and decided this is what it needs to be for now because @Jerryescandon is working on a slightly different design right now anyway and it will be better addressed then

PR #13398, opened 2025-05-02
[Debt] Replace `lodash` functions with existing JavaScript functions

### LINKED ISSUE #12970: ♻️ Replace `lodash` functions with existing JavaScript functions
## ♻️ Refactor

Some `lodash` functions can be replaced by existing JavaScript functions.

## 🙋‍♀️ Proposed Solution

Replace these lodash functions with existing Javascript functions. Add ESLint `no-restricted-imports` rule for each of these functions to prevent use going forward.

1. `isArray`: https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore?tab=readme-ov-file#_isarray
3. `flatMap`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap (we actually use the native function in more places than the lodash version already)
4. `get`: https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore?tab=readme-ov-file#_get
5. `groupBy`: https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore?tab=readme-ov-file#_groupby
6. `upperCase`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase
7. `head`: https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore?tab=readme-ov-file#_head-and-_tail
8. `reverse`: https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore?tab=readme-ov-file#_reverse

REF: https://github.com/GCTC-NTGC/gc-digital-talent/issues/12949#issuecomment-2711629711

## ✅ Acceptance Criteria

- [ ] ESLint `no-restricted-imports` rule for each function to prevent use going forward
- [ ] no instances of lodash `isArray` in codebase
- [ ] no instances of lodash `flatMap` in codebase
- [ ] no instances of lodash `get` in codebase
- [ ] no instances of lodash `groupBy` in codebase
- [ ] no instances of lodash `upperCase` in codebase
- [ ] no instances of lodash `head` in codebase
- [ ] no instances of lodash `reverse` in codebase

JamesHuf:
🤖 Resolves #12970.

## 👋 Introduction

Replaces some lodash functions with existing Javascript functions and adds ESLint `no-restricted-imports` rule for each of them to prevent use going forward.

## 🧪 Testing

Just look over various pages confirming everything still works the same.


packages/forms/src/components/Combobox/utils.ts
```diff
@@ -1,6 +1,5 @@
 import { ValidationRule } from "react-hook-form";
 import orderBy from "lodash/orderBy";
-import isArray from "lodash/isArray";
 import isNumber from "lodash/isNumber";
 import isObject from "lodash/isObject";
 import isString from "lodash/isString";
```
esizer, line 5: I know it was not part of the scope but, we could probably get rid of these as well :thinking: 

```tsx
// isNumber
typeof x === "number"

// isObject
x instanceof Object

// isString
typeof x === "string"
```

esizer (changes requested):
Looking good. I made one minor suggestion to slightly expand the scope, if it is too much let me know and I'll create another ticket for it.

packages/forms/src/components/DateInput/ControlledInput.tsx
```diff
@@ -39,7 +38,7 @@ const ControlledInput = ({
   const intl = useIntl();
   const inputStyles = useInputStyles();
   const selectStyles = useInputStyles("select");
-  const rawDefaultValue: unknown = get(defaultValues, name);
+  const rawDefaultValue: unknown = defaultValues?.[name];
```
esizer, line 41: Oh shoot sorry, I think this is causing a bug with field arrays since it is nested.

```ts
// Looking for this
const defaultValues = {
  'date.0': '2020-01-02'
};

// Actually this
const defaultValues = {
  data: [ '2020-01-02' ]
};
```

If we want complex paths, you could use this: https://youmightnotneed.com/lodash#get 🤔

esizer (changes requested):
Sorry, noticed a bug when reviewing chromatic 👎

packages/forms/src/components/DateInput/ControlledInput.tsx
petertgiles: Please don't add any new helper functions in this PR.  If you can't accomplish it with built-in language features then it should be out of scope.

JamesHuf: Ok, I'm not sure how to do this without adding a helper function so I've [reverted](https://github.com/GCTC-NTGC/gc-digital-talent/pull/13398/commits/85b193c4ab5ea1a3e4d3180fdd40c312864ee1dd) to the lodash function here instead.

esizer (approved):
LGTM! :+1:

PR #13401, opened 2025-05-05
[Fix] Reloading employee profile on save

### LINKED ISSUE #12893: ⌨ Career planning: Inconsistent behaviour when saving form sections
## ⌨ Accessibility Issue

On the career planning page, after saving a section of the form, the page reloads inconsistently. Sometimes the user is returned to the exact spot they were previously at (e.g., "career development preferences"), while other times they are sent back to the top of the page (e.g., "your career objectives"). This inconsistent behavior disrupts the user experience and can cause confusion or frustration.

### WCAG 2.1 Criteria

- WCAG 3.2.3 Consistent Navigation – Navigational mechanisms that are repeated on multiple pages must occur in the same relative order each time they are repeated.
- Heuristic: Consistency and Standards – Users expect consistent behavior when performing similar actions, such as saving form data.
- Heuristic: User Control and Freedom – Users should easily be able to continue their task without unnecessary disruptions or having to re-navigate the page.

## 🦋 Expected Behaviour

After saving a section of the form on the career planning page, the user should consistently be brought back to the spot they were previously at. This maintains continuity in the user flow and avoids unnecessary scrolling or re-navigation.

## 🕵️ Details

N/A

## 📋 Steps to Reproduce

Tip: Use a keyboard to make this issue more obvious.

esizer:
🤖 Resolves #12893 

## 👋 Introduction

Fixes an issue where some forms on the employee profile page were causing the entire page to reload, moving the users focus.

## 🕵️ Details

This was because the mutation was returning fields that were causing the currently logged in user to go stale which forced the authorisation container to re-render meaning the entire page would

## 🧪 Testing

2. Login as an employee `applicant-employee@test.com`
3. Navigate to the employee profile `/applicant/employee-profile`
4. Make changes to and submit all forms on the page
5. Confirm none cause the entire page to reload

### DIFF
--- a/apps/web/src/pages/EmployeeProfile/components/CareerObjective/CareerObjectiveSection.tsx
+++ b/apps/web/src/pages/EmployeeProfile/components/CareerObjective/CareerObjectiveSection.tsx
@@ -150,7 +150,9 @@ const UpdateEmployeeProfile_Mutation = graphql(/* GraphQL */ `
     $employeeProfile: UpdateEmployeeProfileInput!
   ) {
     updateEmployeeProfile(id: $id, employeeProfile: $employeeProfile) {
-      ...EmployeeProfileCareerObjective
+      userPublicProfile {
+        id
+      }
     }
   }
 `);
--- a/apps/web/src/pages/EmployeeProfile/components/GoalsWorkStyleSection/GoalsWorkStyleSection.tsx
+++ b/apps/web/src/pages/EmployeeProfile/components/GoalsWorkStyleSection/GoalsWorkStyleSection.tsx
@@ -43,9 +43,9 @@ const UpdateEmployeeProfile_Mutation = graphql(/* GraphQL */ `
     $employeeProfile: UpdateEmployeeProfileInput!
   ) {
     updateEmployeeProfile(id: $id, employeeProfile: $employeeProfile) {
-      aboutYou
-      learningGoals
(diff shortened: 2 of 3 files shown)


mnigh (changes requested):
I created a new government user for the purposes of testing and although the page does not reload on submission of changes for each form, some of the values rendered are not reflecting the changes made on save. For example, the **Senior management status** checkbox and some other fields that are conditionally rendered based on that.

> [!CAUTION]
> Sorry for the poor quality screen recording. GitHub wouldn't allow over 100mb files so I exported it to 480p and overwrote the original file. 🤦

## Screen recording

https://github.com/user-attachments/assets/2815de4b-ceae-4c53-82d0-67d7f72d7987

esizer:
> I created a new government user for the purposes of testing and although the page does not reload on submission of changes for each form, some of the values rendered are not reflecting the changes made on save. For example, the **Senior management status** checkbox and some other fields that are conditionally rendered based on that.

I was able to fix the form stuff. The display stuff is a but harder and a known issue where things pop-in after a while and it is just the nature of how urql works I think :-1: 

[add missing fields to reset](https://github.com/GCTC-NTGC/gc-digital-talent/pull/13401/commits/1ddb075aa4359b1a5783f05486d334d3e7bbf6ab)

mnigh (approved):

PR #13449, opened 2025-05-08
[Debt] Migrate user scopes to builder

### LINKED ISSUE #13193: ♻️ Migrate `User` model scopes to `Builder`
## ♻️ Debt/Refactor

Migrate the scopes on the `User` model to a custom `UserBuilder`.

## 🕵️ Details

We have done this for the `Pool` model and this should follow a similar pattern.

1. Scopes moved to ` UserBuilder`
2. Drop the `scope` prefix since it is no longer needed
3. Look to use laravel keywords where appropriate for consistency (`where*`)

## 🙋‍♀️ Proposed Solution

Migrate scopes from `User` to a new `UserBuilder` using the `Pool` and `PoolBuilder` classes as an example.

## ✅ Acceptance Criteria

- [ ] `User` scopes contained in a custom builder
- [ ] Scopes still function in PHP and graphql

esizer:
🤖 Resolves #13193 

## 👋 Introduction

Moves all scopes from the user model to a custom builder.

## 🕵️ Details

Scopes were prefixed with `where` to meet laravel convention and keep PHPStan happy.

> [!NOTE]
> I did not understand the `scopeGeneralSearch` one so I left it. If I could get some help I would like to move it over but for now, I left it where it was.

## 🧪 Testing

1. Confirm no scopes in `User.php` model
2. Confirm scopes still function as expected


api/app/Builders/UserBuilder.php
```diff
@@ -0,0 +1,567 @@
...
+                $this->whereNot('last_name', 'ilike', $value);
+                $this->whereNot('email', 'ilike', $value);
+            }
+        }
+
+        return $this;
+    }
+
+    public function publicProfileSearch(?string $search): self
+    {
```
vd1992, line 409: Should this not have that "where" prefixing?

api/app/Models/User.php
```diff
@@ -1412,4 +797,64 @@ public function getFullNameAttribute()
     {
         return Str::trim($this->first_name.' '.$this->last_name);
     }
+
+    public static function scopeWhereGeneralSearch(Builder $query, ?string $searchTerm): Builder
+    {
```
vd1992, line 802: If you find this tricky to move, I'm not opposed to leaving it as is

esizer: Yeah, if you want to take a stab at it you can :sweat_smile: I tried but it kept returning nothing in the tests :cry:

api/app/Builders/UserBuilder.php
esizer: Good catch, now lets hope my change did not break naything :laughing: 

[add prefix for public profile search](https://github.com/GCTC-NTGC/gc-digital-talent/pull/13449/commits/b9db5b579b8963c1cf9518df5b6c86268292c7b5)

yonikid15 (approved):
Nice work looks good :+1:

api/app/Models/User.php
vd1992: Yeah, I'd rather spin it be spun off to another issue if desired to do. Could get a bit involved with refactoring

vd1992 (approved):
`User.php` less gargantuan 👍

Issue #13269, opened 2025-04-16
♻️ Fill in UAT with faked data


petertgiles:
### ♻️ Debt/Refactor

We have seen bugs that only express with a sufficient number of database rows.  Let's ensure that UAT has a realistic amount of data in it to show this.

### 🕵️ Details

Here's a query to find row counts:
```sql
SELECT table_name, (SELECT n_live_tup FROM pg_stat_user_tables WHERE relname = table_name) AS row_count
FROM information_schema.tables
WHERE table_schema = 'public'
and table_catalog = 'gctalent'
order by table_name;
```

Current prod row counts:
| Table                                           | Rows    |
|-------------------------------------------------|---------|
| activity_log                                    | 884480  |
| applicant_filter_classification                 | 0       |
| applicant_filter_pool                           | 3862    |
| applicant_filter_qualified_classification       | 4700    |
| applicant_filter_skill                          | 9086    |
| applicant_filter_work_stream                    | 4700    |
| applicant_filters                               | 5686    |
| assessment_results                              | 36115   |
| assessment_step_pool_skill                      | 2395    |
| assessment_steps                                | 356     |
| classification_development_program              | 40      |
| classification_pool_candidate_filter            | 0       |
| classifications                                 | 81      |
| communities                                     | 3       |
| community_interest_work_stream                  | 1189    |
| community_interests                             | 212     |
| department_user_career_objective                | 2181    |
| department_user_next_role                       | 2897    |
| departments                                     | 24      |
| development_program_interests                   | 867     |
| development_program_talent_nomination           | 5       |
| development_program_talent_nomination_event     | 4       |
| development_programs                            | 9       |
| experience_skill                                | 462357  |
| experiences                                     | 217804  |
| failed_jobs                                     | 3295    |
| general_question_responses                      | 53829   |
| general_questions                               | 155     |
| generic_job_title_user                          | 0       |
| generic_job_titles                              | 0       |
| job_poster_template_skill                       | 669     |
| job_poster_templates                            | 31      |
| jobs                                            | 0       |
| migrations                                      | 210     |
| notifications                                   | 1353988 |
| permission_role                                 | 233     |
| permission_user                                 | 0       |
| permissions                                     | 155     |
| personal_access_tokens                          | 0       |
| pool_candidate_education_requirement_experience | 155309  |
| pool_candidate_filters                          | 0       |
| pool_candidate_search_requests                  | 6010    |
| pool_candidates                                 | 66885   |
[...]

PR #13454, opened 2025-05-08
[Bug] Fix remaining `canQuery` call that fetches all

### LINKED ISSUE #13249: 🐛 Investigate the use of canQuery without eq
## 🐛 Bug

We had [a bad bug](https://github.com/GCTC-NTGC/gc-digital-talent/issues/13241) where we used canQuery without an eq directive.  It seems like this directive will load and evaluate every model in the database.  We need to check if this is a problem anywhere else.

## 🕵️ Details

It seems like the other place we do this is Pools.department.  It is likely that the departments table is small enough that it hasn't caused us grief yet.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/34d9f9942d8ada97a6f7fb7a3290f218e7979ab2/api/graphql/schema.graphql#L434

## ✅ Acceptance Criteria

- [ ] All uses of canQuery examined
- [ ] Any uses that load the entire table have been fixed.

vd1992:
🤖 Resolves #13249

## 👋 Introduction

Investigating, and it appears the one expected instance of `canQuery` fetching all does indeed fetch all. Once a high number of departments exists, queries that touch that policy degrade significantly.
I elected to delete the policy call as neither current nor near-future requires gating department, no context where you need to hide  the department from a pool you can view. 

## 🕵️ Details

Lotta seeding to do testing. Tangential, some possible performance issues become noticeable with large data volume, An issue already exists for this though https://github.com/GCTC-NTGC/gc-digital-talent/issues/13269

## 🧪 Testing

1. Not sure, at the minimum ensure nothing broke

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -431,7 +431,7 @@ type Pool implements HasRoleAssignments {
     @hasManyThrough
     @canRoot(ability: "viewTeamMembers")
   teamIdForRoleAssignment: ID @with(relation: "team")
-  department: Department @belongsTo @canQuery(ability: "view")
+  department: Department @belongsTo
   areaOfSelection: LocalizedPoolAreaOfSelection
     @rename(attribute: "area_of_selection")
   selectionLimitations: [LocalizedPoolSelectionLimitation!]


esizer (approved):
LGTM! :+1:

PR #13470, opened 2025-05-12
[Feat] Scroll to top on nomination form step transition

### LINKED ISSUE #13289: 🐛 Navigating to next page in nominations doesn't always scroll back to top of page

Hitting the "next" button should scroll back to the top after navigating to the new page.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Start a nomination
2. I tested on Step 4. Scroll down the page to the "next step" button and activate it
3. Notice that you are moved to Step 5, but you are not scrolled back to the top of the page.

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

Each "page" step tracks it's H2 heading at the top.  On mount, focus it.

### ✅ Acceptance Criteria

- [ ] Navigating to the Next (or previous) step in the nomination stepper scrolls you back to the top of the page.

esizer:
🤖 Resolves #13289 

## 👋 Introduction

Adds a focus and scroll to top on mount of the nomination talent step sub headings.

## 🧪 Testing

2. Login as an employee `applicant-employee@test.com`
3. Start a nomination
4. Fill out a step with lots of data to force a scroll that is noticable
5. Submit the page
6. Confirm the next page subheading is focus and scroll position is reset

### DIFF
--- a/apps/web/src/pages/TalentNominations/NominateTalent/components/SubHeading.tsx
+++ b/apps/web/src/pages/TalentNominations/NominateTalent/components/SubHeading.tsx
@@ -1,13 +1,33 @@
-import { Heading, HeadingProps } from "@gc-digital-talent/ui";
+import { useEffect, useRef } from "react";
 
-const SubHeading = (props: HeadingProps) => (
-  <Heading
-    level="h2"
-    color="primary"
-    data-h2-margin-top="base(0)"
-    data-h2-font-weight="base(400)"
-    {...props}
-  />
-);
+import { Heading, HeadingProps, HeadingRef } from "@gc-digital-talent/ui";
+
+interface SubHeadingProps extends HeadingProps {
+  preventAutoFocus?: boolean;
+}
+
+const SubHeading = ({ preventAutoFocus, ...rest }: SubHeadingProps) => {
+  const headingRef = useRef<HeadingRef>(null);
+
+  useEffect(() => {
+    if (headingRef.current && !preventAutoFocus) {
+      // Focus heading and scroll to top
+      headingRef.current.focus({ preventScroll: false });
+      window.scrollTo({ top: 0, left: 0, behavior: "instant" });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  return (
+    <Heading
+      ref={headingRef}
+      level="h2"
+      color="primary"
+      data-h2-margin-top="base(0)"
+      data-h2-font-weight="base(400)"
+      {...rest}
+    />
+  );
+};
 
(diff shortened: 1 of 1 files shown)


apps/web/src/pages/TalentNominations/NominateTalent/components/SubHeading.tsx
```diff
@@ -1,13 +1,33 @@
...
+  const headingRef = useRef<HeadingRef>(null);
+
+  useEffect(() => {
+    if (headingRef.current && !preventAutoFocus) {
+      // Focus heading and scroll to top
+      headingRef.current.focus({ preventScroll: false });
+      window.scrollTo({ top: 0, left: 0, behavior: "instant" });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
mnigh, line 19: What's the reason to omit `preventAutoFocus` from the dependency array?

mnigh (commented):
I believe the single item in the acceptance criteria "Navigating to the Next (or previous) step in the nomination stepper scrolls you back to the top of the page" is met, but I'm not sure about what constitutes "top of the page". Is main content considered top of the page or should it focus on the `h1` element or the `body` element? 

## 📹 Screen recording

https://github.com/user-attachments/assets/7616c05f-9e97-4e5d-933f-2e5954804004

apps/web/src/pages/TalentNominations/NominateTalent/components/SubHeading.tsx
esizer: I want it to run on mount which needs an empty dependency array.

petertgiles (commented):
Not part of the AC and not a blocker - but I feel like this is a pattern we may wan to deploy more widely.  Is there a way we can replicate this in a more seamless way?  Like maybe a "FocusOnMount" wrapper that handles the ref and the useEffect?

petertgiles:
> what constitutes "top of the page"

Gray's issue, but I'm guessing he just wants to replicate regular page navigation behaviour.

esizer:
> Not part of the AC and not a blocker - but I feel like this is a pattern we may wan to deploy more widely. Is there a way we can replicate this in a more seamless way? Like maybe a "FocusOnMount" wrapper that handles the ref and the useEffect?

Probably but we would need to make it generic/polymorphic since the element type is required for the ref. Also, we may need to do some handling so we don;t have two on a single page competing for focus :thinking:

petertgiles:
Sounds good.  I'll stub a new issue and we'll leave this one alone.

mnigh (approved):

PR #13473, opened 2025-05-13
[Feat] Environment variable for database SSL mode

### LINKED ISSUE #12333: 🐛 SSL not required for DB connection in prod
## 🐛 Bug

Right now our connection to the database has sslmode set to "prefer".  In the documentation, this is described as: 
>  	I don't care about encryption, but I wish to pay the overhead of encryption if the server supports it.

This is fine for local, but inappropriate for production.  At least, our production database server will reject the connection if SSL is not possible.

## 🦋 Expected Behaviour

We should be enforcing at least encryption in production.

## 🕵️ Details

https://github.com/GCTC-NTGC/gc-digital-talent/blob/e172ab60214c7c080654bbc1063dca054aa34b3f/api/config/database.php#L78

https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-CLIENTCERT

## 🙋‍♀️ Proposed Solution

Option 1: Make the sslmode option configurable and set it to "require" in production.
Option 2: Change the global option in the config file to "require" and figure out SSL in our postgres docker container.

I don't think we need to go all the way to "verify" since our traffic stays within Azure.

## ✅ Acceptance Criteria

- [ ] SSL for the DB connection is required in production

esizer:
🤖 Resolves #12333 

## 👋 Introduction

Adds an environment variable for configuring the database SSL mode.

## 🧪 Testing

1. Open `api/.env`
2. Add `DB_SSL_MODE=require` (or update if it exists)
3. Run `make refresh-api`
4. Confirm it fails with "does not support ssl"
5. Do the same in UAT and confirm it passes

## 🚚 Deployment

Set an environment variable `DB_SSL_MODE=require` and clear the config cache for it to get picked up.


brindasasi:
deploying this into dev

brindasasi:
> I've added env variable in dev but still keep getting this error in dev. Site is not loading. https://dev-talentcloud.tbs-sct.gc.ca/en/ image
> 
> I tried running prod version into Dev which worked fine. I tried this branch and the site is throwing this error

Oh this looks like a cache issue.. after struggling for some good minutes.. I got this back up and running! 
It's the samething in UAT. After deployment there is some cache issue going on with the React I guess

brindasasi (approved):
LGTM!

PR #13479, opened 2025-05-14
[Feat] Base tailwind config

### LINKED ISSUE #13461: ✨ Tailwind configuration and kitchen sink
### ✨ Feature Request

We want to set up our Tailwind configuration so it closely matches our Hydrogen config.  We'll need a "kitchen sink" demo to prove acceptance.

### 🕵️ Details

_No response_

### 🎨 Design file

_No response_

### 🎨 Designer

_No response_

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Implementation

_No response_

### 🌎 Localization

_No response_

### ✅ Acceptance Criteria

- [ ] Tailwind configuration completed
- [ ] Does not affect Hydrogen config (for example, CSS reset)
- [ ] A kitchen sink test to prove success
  - [ ] Not visible in prod (not committed?, storybook only?, not in router?, feature flag?)
- [ ] Supports all requirements
  - [ ] dark mode
  - [ ] states (focus, active, disabled...)
  - [ ] iap theme
  - [ ] targeting child attributes

### 🚢 Definition of Ready

- [x] This issue has met the criteria in our Definition of Ready

### 🛑 Blocking Issues

Blocked by #13459

esizer:
🤖 Resolves #13461 

## 👋 Introduction

Adds the base tailwind config includes colours, shadows, breakpoints and base font-size. It also installs the prettier plugin for sorting class names.

### Colours

We have transitioned all the colours to the new palette as well as some naming changes. Some notable changes:

 - No longer use `light` or `dark` modifiers but instead use a number scale that increments by `100` to leave room for middle colours where `100` is the lightest and current `700` is the darkest
 - The middle round for shades is `400` but has been aliased so you can omit that
 - We have removed `tertiary`, `quaternary` and `quinary`
 - `primary` and `secondary` have been swapped
 
Something I have been thinking about is the swapping of colours between light and dark mode. We have the auto swap on for gray so you can use it in background and text colour to keep contrast. However, I was wondering if we should leave it as manual for other colours so you need to be intentional about it :thinking: Example:

```tsx
<div className="bg-primary-100 dark:bg-primary-900" />
```

This way, it is explict that the colour is changing depending on modes and you can be specific about the shade to use. The auto switching is nice for base components but we rarely update those so it might not be so bad to force us to be explicit about the colour being used in each mode.

### Shadows

The shadows have been migrated one-to-one so you should have no issus with those. We have `inset` and then t-shirt sizes from `sm` to `2xl`.

 - `inset-shadow`
 - `shadow-sm`
 - `shadow-2xl`

### Breakpoints

We have all the same breakpoints as before but the naming has moved from target devices to t-shirt sizes scaling from `xs` to `lg`. These are used as modifiers before the class name with a colon separator.

```html
<div className="xs:grid-cols-2 sm:grid-cols-4 md:grid-cols-6 lg:grid-cols-8"></div>
````

### Font size

Other than maintintg the base font-size of `18px` I have not added anything else. The defualt contains a lot of variety that I think we can use to mirror our existing config. I think this is something that might get touched up once we start doing the `Heading` component though.

### Border radius

We only really ever use one border radius and that actually lines up nicely with tailwinds default so I left these as is.

## 🧪 Testing

> [!IMPORTANT]
> Right now storybook does not automatically switch the theme. You need to pick one and then refresh the page to have it switch properly. 

1. Install new deps `pnpm i`
2. Run storybook `pnpm run storybook`
3. Open the "Style guide" story and go to the tailwind one
4. Change the viewport and colour theme
5. Confirm The colours are correct and swap accordingly
6. Confirm the grid columns reduce by 2 each media query drop


petertgiles (commented):
Very cool!  I see I have some reading to do now.

There are two A/C that I'm not sure were covered:
- [ ] states (focus, active, disabled...)
- [ ] targeting child attributes

Could you please double-check that these were proved?

> We have the auto swap on for gray so you can use it in background and text colour to keep contrast. However, I was wondering if we should leave it as manual for other colours so you need to be intentional about it

Unfortunately, this is pretty theoretical for me.  In theory, this makes sense to me.  As you say, this is probably mostly used in base components where being explicit is most helpful.  Is this logical for people coming from a pure TW context?  Is it confusing for just gray to have a different behaviour?

petertgiles:
> This is possible, I just was not sure how to best represent that in the story

Yeah, nothing fancy.  That example you have seems good to me.

> They actually have this in their docs as an example where they calla colour "canvas" to commuicate that it is for the background. We could keep the background colour name and then have another for text?

Still pretty theoretical to me, but yeah, this makes a lot of sense in my mind.

esizer:
> > They actually have this in their docs as an example where they calla colour "canvas" to commuicate that it is for the background. We could keep the background colour name and then have another for text?
> 
> Still pretty theoretical to me, but yeah, this makes a lot of sense in my mind.

I'm actually interested to see what @JoshBeveridge thinks on this one actually :eyes:

PR #13511, opened 2025-05-16
[Debt] Move `PoolCandidate` scopes to builder

### LINKED ISSUE #13194: ♻️ Migrate `PoolCandidate` model scopes to `Builder`
## ♻️ Debt/Refactor

Migrate the scopes on the `PoolCandiate` model to a custom `PoolCandidateBuilder`.

## 🕵️ Details

We have done this for the `Pool` model and this should follow a similar pattern.

1. Scopes moved to ` PoolCandidate`
2. Drop the `scope` prefix since it is no longer needed
3. Look to use laravel keywords where appropriate for consistency (`where*`)

## 🙋‍♀️ Proposed Solution

Migrate scopes from `PoolCandidate` to a new `PoolCandidateBuilder` using the `Pool` and `PoolBuilder` classes as an example.

## ✅ Acceptance Criteria

- [ ] `PoolCandidate` scopes contained in a custom builder
- [ ] Scopes still function in PHP and graphql

esizer:
🤖 Resolves #13194 

## 👋 Introduction

This moves all the scopes from the `PoolCandidate` model to a custom builder to clean the model up a bit.

## 🧪 Testing

1. Test out pool candidate scopes (find talent and admin table are good spots)
2. Confirm they work as expected


api/app/Builders/PoolCandidateBuilder.php
```diff
@@ -0,0 +1,565 @@
...
+
+    public function whereOperationalRequirementsIn(?array $operationalRequirements): self
+    {
+        if (empty($operationalRequirements)) {
+            return $this;
+        }
+
+        return $this->whereHas('user', function ($query) use ($operationalRequirements) {
+            $query->whereOperationalRequirementsIn($operationalRequirements);
+        });
```
vd1992, line 111: I like the comment/pattern above. You can infer from the `whereHas` what model is being acted upon, but the comment makes it obvious and more importantly, where to find the method

image

api/app/Builders/PoolCandidateBuilder.php
```diff
@@ -0,0 +1,565 @@
...
+    }
+
+    public function whereGeneralSearch(?string $searchTerm): self
+    {
+        if (empty($searchTerm)) {
+            return $this;
+        }
+
+        return $this->whereHas('user', function ($userQuery) use ($searchTerm) {
+            $userQuery->whereGeneralSearch($searchTerm);
```
vd1992, line 163: Scrolling through and I see a phase of using `$userQuery`, an alternative to the above and neat way to name the variable off the builder in play

api/app/Traits/Generator/Filterable.php
```diff
@@ -11,6 +11,18 @@ trait Filterable
...
+    protected bool $debug = false;
+
+    public array $calledFilters = [];
+
+    public function debug(bool $value)
+    {
+        $this->debug = $value;
+
+        return $this;
+    }
```
vd1992, line 24: Is this something we intend to keep, or was this meant to be deleted after a problem was found?

api/app/Builders/PoolCandidateBuilder.php
esizer: That comment is only necessary because PHPStan cannot infer unless it is prefixed with `where` and I'd love to remove it eventually :sweat_smile:

api/app/Generators/ApplicationZipGenerator.php
```diff
@@ -20,7 +20,7 @@ public function generate(): self
             'generalQuestionResponses' => ['generalQuestion'],
         ])
             ->whereIn('id', $this->ids)
-            ->authorizedToView(['userId' => $this->userId])
+            ->whereauthorizedToView(['userId' => $this->userId])
```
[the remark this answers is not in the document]
vd1992: I think this might be causing a massive bug locally by silently not running the scope at all

api/graphql/schema.graphql
vd1992, line 1: An aside, I haven't used the candidates table much recently but I don't get what some of our non-default searches are even doing

image

image

???

vd1992, line 1: An aside, I haven't used the candidates table much recently but I don't get what some of our non-default searches are even doing

image

image

???

vd1992 (changes requested):
One typo/bug, one question

api/app/Traits/Generator/Filterable.php
esizer: Unfortunately we need to keep it so we can use it in the test :cry:

api/app/Generators/ApplicationZipGenerator.php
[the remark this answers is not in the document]
esizer: oof, good eye... wonder if we could get ti to fial loudly :sweat: 

Fixed in [fix typo](https://github.com/GCTC-NTGC/gc-digital-talent/pull/13511/commits/b4e1fe7d282507fef7fd2338f0af19234cf695eb)

api/graphql/schema.graphql
esizer: Are these two screenshots from main vs this branch?

vd1992: Occurs for both branch and main, I noticed it now while exercising the candidates table for testing.
Nothing in scope for the PR, just don't get what is going on with some of the searches

api/app/Generators/PoolCandidateCsvGenerator.php
vd1992, line 1: Typo fix didn't solve it, I am still seeing a discrepancy in candidate CSV download

Off main -> all candidates table -> download full dataset
Off branch -> repeat
I get a discrepancy in row count

vd1992 (changes requested):

api/app/Generators/PoolCandidateCsvGenerator.php
esizer: Hmm, I'm getting the same number that shows in the table for both main and this branch :confused:

vd1992: Sorry I meant a discrepancy in row count for for the spreadsheet
The Excel document has 68 rows for main, 71 rows for the branch. They should be identical
The all candidates table has 67 total for both, which makes sense, 67 + column headers row

esizer: Yeah, mine are identical for both spreadsheets 😕

vd1992: Seeded fresh, still got it. Main now 67 rows, 69 branch

esizer: Might need someone else to take a look because I keep getting identical numbers of 63 rows in each :confused:

vd1992: I suppose so, guess you get to voluntold someone

vd1992: Digging some more, the candidates that appear in the branch download and **not** main are either EXPIRED or SUSPENDED

api/app/Generators/PoolCandidateCsvGenerator.php
```diff
@@ -521,15 +521,15 @@ private function buildQuery()
...
 
         $this->applyFilters($query, [
-            'priorityWeight' => 'candidateCategory',
-            'poolCandidateStatus' => 'poolCandidateStatuses',
-            'pools' => 'availableInPools',
+            'priorityWeight' => 'whereCandidateCategoryIn',
+            'poolCandidateStatus' => 'whereStatusIn',
+            'pools' => 'whereAvailableInPools',
             'skills' => 'skillsAdditive',
             'community' => 'candidatesInCommunity',
```
vd1992, line 528: These should be updated too right?
`whereSkillsAdditive`
`whereHasPoolCandidateCommunity`

api/app/Generators/PoolCandidateCsvGenerator.php
esizer: Thanks for that info. That plus the other comment tipped me off. I forgot to add the new renames to the scope map! :confounded: I really wish it was easier to maintain that :sob: 

Should be fixed after [add missing scopes to map](https://github.com/GCTC-NTGC/gc-digital-talent/pull/13511/commits/660f9e37fedd5e714de8b9688430189b40e4b9da)

api/app/Generators/PoolCandidateCsvGenerator.php
esizer: Fixed as part of [add missing scopes to map](https://github.com/GCTC-NTGC/gc-digital-talent/pull/13511/commits/660f9e37fedd5e714de8b9688430189b40e4b9da)

vd1992 (approved):
Well, we're definitely going to need to have best practices if we're adding more downloads 😅 
That aside, I think it works well now. Glad to have it sorted out

PR #13516, opened 2025-05-16
[Debt] Add employee profile E2E test

### LINKED ISSUE #13250: 🧪 e2e for talent management
## 🧪 Tests

We should add some e2e tests for our new talent management feature.

## ✅ Acceptance Criteria

- [ ] Playwright tests that:
  - [ ] Fill in the career management section

## 🛑 Blockers

```[tasklist]
### Blocked By
```

JamesHuf:
🤖 Resolves #13250.

## 👋 Introduction

Adds e2e test for filling out the employee profile.

## 🧪 Testing

Confirm test passes with `pnpm run e2e:playwright employee-profile`.

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -1550,6 +1550,7 @@ input CreateUserInput {
       ]
       messages: [{ rule: "unique", message: "CreateUserEmailInUse" }]
     )
+  workEmailVerifiedAt: DateTime @rename(attribute: "work_email_verified_at")
   govEmployeeType: GovEmployeeType
     @rename(attribute: "computed_gov_employee_type")
   currentClassification: ClassificationBelongsTo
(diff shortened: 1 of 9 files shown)


petertgiles (changes requested):
Very nice!  This is very easy to read test.  I think you're missing one AC though: to check that the validation checkmarks are working on the page, too.

JamesHuf:
> I think you're missing one AC though: to check that the validation checkmarks are working on the page, too.

I've added validation for the icons but had to use testIds for it since they're aria-hidden and I'm not sure how else to locate them.

Also I broke up the test to make it a bit tidier hopefully.

JamesHuf:
After taking a look at this I feel it makes more sense to add labels in a separate PR.

It might also be a good idea to look for other areas of the site which have meaningful icons missing labels.

petertgiles (approved):
Thanks going the bit extra on this one.  Looks great!  You'll probably need to move that string from web to i18n now.

PR #13518, opened 2025-05-20
[Debt] Remove forms barrel file

### LINKED ISSUE #13248: ♻️ Remove `@gc-digital-talent/forms` barrel file
## ♻️ Debt/Refactor

We would like to test our removing barrel files to see if they can help out on project performance since we have had issues with typescript, linting and jest performance.

## 🕵️ Details

We have barrel files to assist in making imports more convenient. However, the convenience could be having some level of significance on our tooling performance.

The goal of this is to replace the barrel file in the forms package to see if there is any difference and see if we like it.

## 🙋‍♀️ Proposed Solution

- Remove the bare file at `./packages/forms/src/index.ts`
- Update the `./packages/forms/package.json` to include to component imports
- Update references to use the new subpath import

```json
{
  "name": "@gc-digital-talent/forms",
  "exports": {
     "./Input": "./src/components/Input.tsx",
     "./Textarea": "./src/components/Textarea.tsx",
  }
}
```

```tsx
import { Input } from "@gc-digital-talent/forms/Input";
import { Textarea } from "@gc-digital-talent/forms/Textarea";
```

## ✅ Acceptance Criteria

- [ ] Forms barrel file removed
- [ ] Forms has exports in `package.json`
- [ ] Imports from forms package updated to subpaths
- [ ] Some sort of improvement on performance (?)

esizer:
🤖 Resolves #13248 

## 👋 Introduction

Removes the forms package barrel file to hopefully help with linting/typescript performance.

## 🕵️ Details

Saw about 500ms difference. Not that exciting but there is a slight improvement and might get better as we do more packages? :thinking: 

```sh
# Original

 Tasks:    11 successful, 11 total
Cached:    0 cached, 11 total
  Time:    27.203s

# Without

 Tasks:    11 successful, 11 total
Cached:    0 cached, 11 total
  Time:    26.772s
```

## 🧪 Testing

1. Confirm `pnpm run tsc --force` works

### DIFF
--- a/packages/forms/eslint.config.js
+++ b/packages/forms/eslint.config.js
@@ -15,10 +15,7 @@ export default [
             1,
             {
               unusedExports: true,
-              ignoreExports: [
-                "src/index.{ts,tsx}",
-                "src/**/*.stories.@(js|jsx|ts|tsx|mdx)",
-              ],
+              ignoreExports: ["src/**/*.stories.@(js|jsx|ts|tsx|mdx)"],
             },
           ]
         : "off",


mnigh (dismissed):
In some ways, this moves us towards how we import with lodash and makes it more uniform. For all of the effort, I'm not sure it is worth it, especially since this would be the first part of a larger project. What do other @GCTC-NTGC/developers think?


As for performance, on my slower than @esizer's machine, I see a ~5 second improvement.

### Before
```
11 successful, 11 total
1m15.402s
```

### After
```
11 successful, 11 total
1m10.689s
```

petertgiles:
I got a 3 second improvement - about 5% faster.  To me, this doesn't seem worth the effort but if someone does want this I'm not opposed either.

It is interesting that it turned up a few "not used within other modules" lint warnings that were previously hidden by the barrel files.

esizer:
> I got a 3 second improvement - about 5% faster. To me, this doesn't seem worth the effort but if someone does want this I'm not opposed either.

It is really hard to say. This is just one of 14, I doubt we get 5% for every one but if we get 1-2% for 12 of them and then another 5-6% for the `ui` (our largest one) that could result in an overall oif >20% but it is a lot of work to do to maybe find out :sweat: 

> It is interesting that it turned up a few "not used within other modules" lint warnings that were previously hidden by the barrel files.

Yup, the linter counted it being imported in the barrel file as being "used" which is one of the down sides of a barrel file. You might accidentally include stuff you are not actually using.

petertgiles:
Discussion in standup: Eric will look at just removing the internal barrel files.

PR #13519, opened 2025-05-20
[Fix] Do not disable skill browser submit button

### LINKED ISSUE #13254: ⌨ Skill browser submit button loses focus when submitting
## ⌨ Accessibility Issue

The submit button in the skill browser dialog loses focus when you interact with it. This is a problem if you submit with an error, you lose focus on the current item and have to cycle back around to fix the error.

I believe this is because we are disabling the submit button on a submission attempt.

### WCAG 2.1 Criteria

Technically you are allowed to disable buttons, it is just [highly discouraged](https://adrianroselli.com/2024/02/dont-disable-form-controls.html#Submit).

## 🦋 Expected Behaviour

Your focus remains intact while submitting the form.

## 🕵️ Details

- **Browser**: Google chrome, safari, firefox
- **Assistive Technology**: None, just the keyboard

## 📋 Steps to Reproduce

1. Login as an applicant `applicant@test.com`
2. Navigate to create an experience
3. Open the "Add skill" dialog
4. Using only the keyboard, attempt to submit the form with an error
5. Observe the focus is lost on the submit button and you need to tab to get back

## 🙋‍♀️ Proposed Solution

Rather than disabling an input, prevent a mutation or callback from happening while one is in flight. 

## ✅ Acceptance Criteria

- [ ] Focus remains on submit button while form is submitting
- [ ] Duplicate submissions cannot occur

esizer:
🤖 Resolves #13254 

## 👋 Introduction

Updates the skill browser dialog to no longer disable the submit button.

## 🧪 Testing

1. Build `pnpm run dev:fresh`
2. Login as an applicant `applicant@test.com`
3. Go to the skill library
4. Using keyboard, select a skill and submit the form
5. Confirm the button is no longer disabled and focus remains on the submit button
6. Confirm you cannot trigger the submit multiple times

### DIFF
--- a/apps/web/src/components/SkillBrowser/SkillBrowserDialog.tsx
+++ b/apps/web/src/components/SkillBrowser/SkillBrowserDialog.tsx
@@ -67,6 +67,7 @@ const SkillBrowserDialog = ({
   const intl = useIntl();
   const paths = useRoutes();
   const [isOpen, setIsOpen] = useState<boolean>(defaultOpen);
+  const [adding, setAdding] = useState<boolean>(false);
   const [selectedSkill, setSelectedSkill] = useState<Skill | null>(null);
   const methods = useForm<FormValues>({
     defaultValues: initialState ?? defaultFormValues,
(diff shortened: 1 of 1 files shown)


JamesHuf (approved):
This seems to work for me

PR #13527, opened 2025-05-21
[Fix] formatjs eslint rules

### LINKED ISSUE #12832: ♻️ Renable and fix formatjs eslint rules
## ♻️ Debt/Refactor

After fixing our formatjs rules, we noticed there are some violations in our syntax. We will need to re-enable the rules we turned off and fix them.

## 🙋‍♀️ Proposed Solution

Following rules turned back on and fixed:

```
      "formatjs/no-multiple-plurals": "off",
      "formatjs/no-multiple-whitespaces": "off",
      "formatjs/prefer-pound-in-plural": "off",
      "formatjs/enforce-plural-rules": "off",
```

## ✅ Acceptance Criteria

- [ ] FormatJS eslint rules enabled
- [ ] All resulting errors fixed

esizer:
🤖 Resolves #12832 

## 👋 Introduction

Re-enables some of the rules and fixes instances of them.

## 🕵️ Details

I left off some because I lack the knowledge authority to enable them :sweat_smile: 

 - [`no-multiple-plurals`](https://formatjs.github.io/docs/tooling/linter#no-multiple-plurals)
     - We have strings with multiple plurals and I'm not sure how to modify them to not use it
 - [`enforce-plural-rules`](https://formatjs.github.io/docs/tooling/linter#enforce-plural-rules)
      - This requires we supply a value of `one` but some we don't and I'm not sure if that is valid or not since `one` is the same as `other` :confused:   

I did not change any of the  [`no-literal-string-in-jsx`](https://formatjs.github.io/docs/tooling/linter#no-literal-string-in-jsx) because it is primarily punctuaion and I don't feel condfident in modifying the French to fix this.

I'm willing to attempt fixing some of these but I'll need some help from someone who has moderate French ability to assist. Or someone else can pick this up? :pray: 

> [!NOTE]
> Spoke with @mnigh and @NienkeBr and agreed the `no-multiple-plurals` isn't a huge issue for us so we can take on the added complexity and have left this rule off. 

## 🧪 Testing

1. Confirm only `no-multiple-plurals` and `no-id` are the only disabled formatjs rules
2. Confirm lint passes
3. Confirm translations are intact

### DIFF
--- a/apps/web/src/lang/fr.json
+++ b/apps/web/src/lang/fr.json
@@ -4515,6 +4515,10 @@
     "defaultMessage": "Bons pour examens de certification",
     "description": "Title for the certification exam vouchers page"
   },
+  "JL6B1D": {
+    "defaultMessage": "Demande de candidats pour le bassin : <primary>{candidateCountNumber, plural, =0 {aucun candidat} one {# candidat estimé} other {# candidats estimés}}</primary>",
+    "description": "Total estimated candidates message in summary of filters"
+  },
   "JO2yLA": {
     "defaultMessage": "Lieu de travail",
     "description": "Work Location label, followed by colon"
@@ -6178,10 +6182,6 @@
   "R8Naqm": {
     "defaultMessage": "Filtrage et évaluation",
     "description": "Heading for the information of an application"
-  },
-  "R9Yk8Y": {
-    "defaultMessage": "Demande de candidats pour le bassin : <primary>{candidateCountNumber, plural, =0 {aucun candidat} =1 {1 candidat estimé} other {# candidats estimés}}</primary>",
-    "description": "Total estimated candidates message in summary of filters"
   },
   "RA4afJ": {
     "defaultMessage": "Approbation pour les programmes de perfectionnement",
@@ -8683,7 +8683,7 @@
     "description": "Page title for the individual user page"
   },
   "dkXjDY": {
-    "defaultMessage": "{monthCount, plural, =1 {# mois} other {# mois}}",
+    "defaultMessage": "{monthCount, plural, one {# mois} other {# mois}}",
     "description": "A duration of a certain number of months"
   },
   "dkYHTv": {
--- a/apps/web/src/lang/fr.json
+++ b/apps/web/src/lang/fr.json
@@ -111,10 +111,6 @@
     "defaultMessage": "Endroits exclus",
     "description": "Location specifics label"
   },
-  "+UTd4G": {
-    "defaultMessage": "{developmentProgramsNominationCount, plural, =0 {Perfectionnement } other {Perfectionnement ({developmentProgramsNominationCount, number})}}",
-    "description": "Development nominations, conditional rendered off count"
-  },
   "+UjXSM": {
     "defaultMessage": "le réseautage.",
     "description": "Seventh item in list of certification topics section"
@@ -1470,10 +1466,6 @@
   "4nP41q": {
     "defaultMessage": "Sauvegardez et ajoutez cette compétence",
     "description": "Button text to save a specific skill to a users profile"
-  },
-  "4ozix3": {
-    "defaultMessage": "{lateralMovementNominationCount, plural, =0 {Mutation latérale} other {Mutation latérale ({lateralMovementNominationCount, number})}}",
-    "description": "Lateral movement nominations, conditional rendered off count"
   },
   "4p0QdF": {
     "defaultMessage": "Fermer",
@@ -3995,10 +3987,6 @@
     "defaultMessage": "Poursuivre la demande<hidden> pour {poolName}</hidden>",
     "description": "Label for continue application link"
   },
-  "Gki8Ex": {
-    "defaultMessage": "{candidateCount, plural, one {<testId>{candidateCount}</testId> concordance approximative} other {<testId>{candidateCount}</testId> concordances approximatives} }",
-    "description": "Message for total estimated matching candidates in pool"
-  },
   "GndGRz": {
     "defaultMessage": "La mise à jour des informations sur votre prochain type de poste a échoué.",
     "description": "Message displayed when a user fails to updates employee profile information"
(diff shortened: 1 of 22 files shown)


vd1992:
For [no-multiple-plurals](https://formatjs.github.io/docs/tooling/linter#no-multiple-plurals)
There aren't many, I think I counted 3 or 4. Some of them really simple what is going on and could be split into two strings easily like

image

I don't think this is a problem we need to prevent via lint rule though. We don't have to worry about translation vendor support after all, the only concern is that too much nesting and it becomes too hard to parse or translate.

esizer:
> I don't think this is a problem we need to prevent via lint rule though. We don't have to worry about translation vendor support after all, the only concern is that too much nesting and it becomes too hard to parse or translate.

It would be nice if we could enforce this, cause I hate trying to parse even single plurals, let alone multiple :sweat_smile: IE:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/722356c5b162d900e5132900e225db2720110849/apps/web/src/pages/SearchRequests/SearchPage/components/SearchForm.tsx#L234-L253

apps/web/src/lang/fr.json
```diff
@@ -8790,6 +8798,10 @@
     "defaultMessage": "Cette bibliothèque contient une série de modèles pour des emplois courants dans les collectivités soutenues par Talents numériques du GC. Les modèles comprennent des informations suggérées et des recommandations pour des éléments tels que le titre du poste, les tâches courantes, les compétences essentielles et les compétences optionnelles. Parcourez la liste complète ou utilisez la barre de recherche et les filtres pour trouver des modèles qui correspondent aux spécificités du poste.",
     "description": "Description of how to use the form to filter job poster templates"
   },
+  "e5N39X": {
+    "defaultMessage": "Résultats : {totalCandidateCount, plural, =0 {<testId><b>#</b></testId> candidats correspondants} one {<testId><b>#</b></testId> candidats correspondants} other {<testId><b>#</b></testId> candidats correspondants à l'échelle } } de {numPools, plural, =0 {<b>#</b> bassins} one {<b>#</b> bassins} other {<b>#</b> bassins} }",
```
mnigh, line 8802: The `one` case seems incorrect, though out of scope for this issue. Worth creating an issue to re-examine French plural translations more generally.

apps/web/src/pages/TalentNominations/NominationGroup/components/NominatedForList.tsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable formatjs/enforce-plural-rules */
+// NOTE: Not really plural, I guess?
```
mnigh, line 2: They are not really plural. This probably should be two strings with a conditional wrapping it, no?

mnigh (commented):
A few questions, but otherwise, looks good.

apps/web/src/pages/TalentNominations/NominationGroup/components/NominatedForList.tsx
esizer: Hmm.. that seems reasonable :thinking: 

I think we have this as [part of the un-ignoring the literal strings ](https://github.com/GCTC-NTGC/gc-digital-talent/issues/13583#:~:text=Counts%20in%20parentheses%20after%20strings%20Iterable%20(x))so maybe that could be a ticket where we unignore and swap it for a conditional rather than plural string?

apps/web/src/lang/fr.json
mnigh: https://github.com/GCTC-NTGC/gc-digital-talent/issues/13585

apps/web/src/pages/TalentNominations/NominationGroup/components/NominatedForList.tsx
petertgiles, line 2: I agree, this is not a good use of the pluralization tool.  I recommend refactoring this - probably to conditionally concatenate the parens to the end of the message.  I feel like we often change the color, weight, or size of the font for these anyway.

packages/eslint-config-custom/react.mjs
```diff
@@ -29,6 +29,8 @@ export default [
       "react/jsx-uses-react": "off",
       "react/react-in-jsx-scope": "off",
       "formatjs/no-id": "off",
+      // We use this a bit and have not had issue with it...yet
+      "formatjs/no-multiple-plurals": "off",
```
petertgiles, line 33: I agree with the decision.  We don't do this much.

petertgiles (changes requested):
Very nice.  Just a few small notes.

petertgiles:
> I wonder if we could do like:

Oh, please don't.  I find that much harder to read.  :laughing:

esizer:
> > I wonder if we could do like:
> 
> Oh, please don't. I find that much harder to read. 😆

I'm so shocked :laughing: 

I have no idea how anyone could find this harder to read than that monstrosity we have now :stuck_out_tongue:

apps/web/src/pages/TalentNominations/NominationGroup/components/NominatedForList.tsx
esizer: Yeah, I created [this ticket ](https://github.com/GCTC-NTGC/gc-digital-talent/issues/13583)but I'm not sure if I should roll this into it or how to deal with it :thinking:

apps/web/src/lang/fr.json
```diff
@@ -8678,6 +8682,10 @@
     "defaultMessage": "Détails sur le candidat",
     "description": "Page title for the individual user page"
   },
+  "dkXjDY": {
+    "defaultMessage": "{monthCount, plural, =1 {# mois} other {# mois}}",
```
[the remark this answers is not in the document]
esizer: Fixed in [fix missing plural rule fixes](https://github.com/GCTC-NTGC/gc-digital-talent/pull/13527/commits/f72b484f40d1e51dd059da4355a5e1b9f247542d)

apps/web/src/pages/TalentNominations/NominationGroup/components/NominatedForList.tsx
petertgiles: Nice, that looks like a good ticket.  :+1:

petertgiles (approved):
lgtm

mnigh (approved):

PR #13688, opened 2025-05-30
[Debt] Tune app services

### LINKED ISSUE #12686: Tune PHP-FPM for new app services
We've updated our app services.  We should do some load testing and adjust things like number of workers to take advantage our new plan.

petertgiles:
🤖 Resolves #12686 

## 👋 Introduction

This PR doesn't actually tune the app services at all.  It turns out that we're not leaving any performance on the table.  :cry: It's actually quite easy to already pin the CPU to 100% in our UAT app service.  It might be better to take another shot at this, from the perspective of "how do we reduce the CPU usage of our app".  I did spend a bit of time checking that opcache was working and it seems that it is.

This PR just provides the Locust test and the opcache debugging to help with future efforts.

### Opcache

1. Restart the webserver container 
2. Visit http://localhost:8000/opcache.php
3. Observe that `opcache_enabled` = 1.
4. Observe that the various keys in `opcache_statistics` change as you use the site.

### Locust
1. I [installed Locust locally with PIP](https://docs.locust.io/en/stable/installation.html) and ran it with `locust -f apps/locust`
    - In the past, I used Docker.  You can see that here: https://github.com/GCTC-NTGC/gc-digital-talent/pull/11865/commits/f7b68b42097fe8b3a1509cda0978b2137b720686 if you want to try that instead.
2. Point it at your local installation, ramp to 100 users, and spin up your laptop's fans.  :laughing: 

## 📸 Screenshot

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ hydrogen-logs
 .turbo
 .env
 hostingstart.html
+__pycache__
--- a/api/opcache.php
+++ b/api/opcache.php
@@ -1,0 +1,11 @@
+<html>
+<pre>
+<?php
+
+$status = opcache_get_status();
+
+print_r($status);
+
+?>
+</pre>
+</html>
--- a/apps/locust/testSearchPage.py
+++ b/apps/locust/testSearchPage.py
@@ -1,0 +1,128 @@
+import string
+import random
+import time
+from locust import HttpUser, task, between
+
+class TestSearchPage(HttpUser):
+    @task
+    def candidate_count(self):
+        query = """
+query CandidateCount($where: ApplicantFilterInput) {
+  countApplicants(where: $where)
+  countPoolCandidatesByPool(where: $where) {
+    pool {
+      id
+      ...SearchResultCard_Pool
+      __typename
+    }
+    candidateCount
+    __typename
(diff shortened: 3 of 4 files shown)


brindasasi:
This is my first time doing this.. I'm curious here 
 I ran this in local and got the below stat 
 
On an avg. I got more than 80% of opcache hit 
Screenshot 2025-06-03 at 3 14 22 PM

I deployed it in dev and ssh to it and directly run opcache.php file from the cli

brindasasi:
> > I deployed it in dev and ssh to it and directly run opcache.php file from the cli & I got opcache hit rate is 0.
> 
> Hmm, my guess is that the PHP "instance" that is running in the CLI and the one running in PHP-FPM are different. You can see from the CLI screenshot that "hits" are 0 and "misses" are 1. Probably that one command you ran is the only activity logged in that instance at all.
> 
> If you really want to test this in Azure you could make a new branch off of this one and copy the "local" ngnix conf changes to the "deploy" nginx conf. As long as we don't merge that branch and don't deploy this branch to prod there's no security risk.

Ok I guess I can do that

brindasasi (approved):
LGTM!

Issue #13180, opened 2025-04-04
🐛 Add space between lists with icons


Jerryescandon:
## 🐛 Bug

Lists with icons (prominently on the GC employee profile) are missing some spacing between each item, causing them to be too close together which looks a little crowded.

![Image](https://github.com/user-attachments/assets/f18c9553-67cd-4890-a4ec-4d3ff59258c0)

## 🦋 Expected Behaviour

Add whitespace variable x-025 or 6.75px 

![Image](https://github.com/user-attachments/assets/349adc5e-2ba2-4b40-b45f-3fbdb8115125)

petertgiles:
I just saw the "NoList" component.  Perhaps this is a chance to standardize a component?  I feel like maybe we need a standardized bulleted list component, too.

petertgiles:
Discussion in standup.  Probably not NoList since that's supposed to have no styling.  But probably yes to standardizing in a component.

PR #13708, opened 2025-06-04
[Feat] Create list component

### LINKED ISSUE #13624: ♻️ Create `List` component
### ♻️ Debt/Refactor

Create a generic `List` component that has some opinionated styles and propagate the change throughout the site.

### 🕵️ Details

Right now, we typically just use `ul` and `ol` as needed and relying on browser default styles for these. However, we are migrating to tailwind which removes these styles in their reset.

We added some custom css for compatibility while we migrate but, this is getting in the way since we need to apply `!important` to override the defaults.

We should probably create a `List` component we can use instead of `ul`.

### 🙋‍♀️ Proposed Solution

Use the styles in the `app.css` for `ul` and apply them to a new `List` component using tailwind. Then, find all places we use `ul` and replace it with the new component.

You may need to accommodate variants here as well.

### ✅ Acceptance Criteria

- [ ] New `List` component with opinionated styles
- [ ] All `ul` replaced with new `List` (likely all with some variant handling and overrides)
- [ ] No significant chromatic diff
- [ ] Add a type prop to swap between ordered and unordered lists
- [ ] Storybook story

esizer:
🤖 Resolves #13624 

## 👋 Introduction

Creates a 2 new list components merging in the `NoList`

## 🕵️ Details

We have two new list components and have removed the `NoList` (explained later).

 - `Ul` -> unordered list
 - `Ol` -> ordered list

The props control the styling and they share the same props.

 - `space`
     - Spaces the items within the list vertically, omit to not add additional space
     - `sm`, `md`, `lg`, `xl`
- `unStyled`
    - Removes all list styling, this is what you would apply to recreate the `NoList`
- `noIndent`
    - Removes the left indent on the list items
- `inside`
    - By default, we use outside lists but you can set it to inside with this prop

> [!NOTE]
> This may also solve [#13180](https://github.com/GCTC-NTGC/gc-digital-talent/issues/13180) 

## 🧪 Testing

> [!IMPORTANT]
> Not all instances of `ul` were replaced. The ones that were not, had very custom implrementations and it didnt make sense to add a standard style just to remove it again. 

1. No significant chromatic diff
2. New component styles make sense


mnigh (approved):
👍

Issue #12492, opened 2025-01-14
✨ Improve UX of modifying user roles


tristan-orourke:
## ✨ Feature

On "Manage access" pages, or on the "Edit User account" page, adding/editing/removing a user role causes the entire page to refresh and re-render. This makes it tedious and frustrating to add multiple users to a resource (eg Community or Process), or to remove several roles from a user at once.

## 🕵️ Details

This is seen on the Community - Manage access page, the Process - Manage access page, and the User edit page (platform admin only).

## 🎨 Design

Consult @Jerryescandon if necessary

## 🙋‍♀️ Proposed Implementation

https://github.com/GCTC-NTGC/gc-digital-talent/issues/12492#issuecomment-3141216185

Old notes:
> Not sure. Ideas:
> - Pull query to a lower level component in the hierarchy?
> - Don't hide page under a spinner during request
> - How to handle sudden layout shift from table rows disappearing?
>   - spinner on or disabling of buttons in individual table row?
>   - animate layout shifts?
>   - don't worry about it?

## ✅ Acceptance Criteria

- [ ] When adding, editing or removing a role on Edit User page or Manage access pages, page doesn't re-render and lose user's place every time.

## 🛑 Blockers

```[tasklist]
### Blocked By
```

vd1992:
Well, looking around and noted the tab 'User information' does not have this behaviour when adding a user to pools. I also noticed, when you use a mutation on the tab in question, an `authorizationQuery` operation is fired immediately after. 

This leads me to suspect the issue is that when you update a user on this tab, the fields and `__typename = UserAuthInfo` get marked as stale which are also present on your authorization query. This triggers a fresh fetch which results in a whole page reboot.
Not sure how you can resolve this

vd1992:
A similar thing appears to be happening with the "Manage access" pages, updating a user's roles marks your authorization as stale

Issue #8309, opened 2023-10-24
🦔 Investigate switching to urql Graphcache in order to use optimistic updates


tristan-orourke:
## Background

When you save an inline form, it immediately goes back to the non-editable version, but showing the old data. Only when the request comes back successfully does the content update to what you just submitted. That delay can be a bit jarring, especially if it takes longer than a second. 

This ticket was originally looking for any solution to the UI delay. I've converted that into a new ticket, #9515, and made this one more specific to one solution we looked into.

## 🦋 Expected Behaviour

When you edit data in a form and submit, the app should immediately treat that data as true, unless/until the request fails.

This is exactly the promise of URQL's optimistic update feature! 
https://formidable.com/open-source/urql/docs/graphcache/cache-updates/#optimistic-updates

## 🕵️ Details

The main prerequisite to using the optimistic update feature is switching to urql graphecache -  we currently use the simpler and more naive document cache.

We don't fully understand yet the implications of this switch, but we assume they will be significant. The optimisic updates _could_ mean a significant overall improvement to UX, but we need to investigate and test the use of graphecache before we can confidently switch.

## Questions

- What are the risks of using graphecache instead of document cache?
- What we need to do to make graphecache work for our needs? How much new or changed code would it require?
- If we can switch to graphecache, how much new or changed code would Optimistic Updates require?
- Can it properly handle polymorphic relationships in graphql?  What are the limitations. 

Timebox: 2-3 days

gobyrne:
Optimistic updating using the data that was collected could solve this. In the event of an error, the data reverts.

esizer:
Looks like the optimistic updates should work. However, the documentation makes it seem like it requires using the graph cache, which requires urql@4.0.0, which we can't upgrade too until #7887 is done 😢 

Unless we consider this higher priority and want to install the previous version until we get a chance to circle back and bump auth, core and cache. Though, that does seem like some unnecessary effort even if [the changes are not significant](https://github.com/urql-graphql/urql/releases/tag/%40urql%2Fexchange-graphcache%406.0.0).

vd1992:
I don't think this is high priority so waiting until the dependencies are sorted would be best. Especially since it could make things messier down the road like reverting the auth bump revert.

esizer:
> I don't think this is high priority so waiting until the dependencies are sorted would be best. Especially since it could make things messier down the road like reverting the auth bump revert.

[...]

PR #13769, opened 2025-06-06
Try something - Improve UX of role modification

vd1992:
🤖 Resolves 

## 👋 Introduction

A possible way to address
issue #12492

It appears the issue was that modifying roles or the `UserAuthInfo` of someone marks your own authorization as stale forcing a page reload.
By having a distinct `__typename` for your auth object, avoiding fetching fields when mutating, and making page queries stale on `UserAuthInfo` changes you can resolve the problem with minimal complexity.

## 🧪 Testing

1. Navigate to `/en/admin/users/:id/edit`
2. Test mutating sub and roles for a user other than your own
3. Navigate to `/en/admin/pools/:id/manage-access`
4. Test adding and removing people
5. Navigate to `/en/admin/communities/:id/manage-access`
6. Test adding and removing people


vd1992:
> Hmm, I think you've narrowed it down nicely but this doesn't seem like a very clean solution. I don't love the necessity of creating a new Graphql type for our UserAuthInfo.
> 
> These UserAuthInfo models have an ID field to use as a key. Why is the URQL cache making our own info stale when we update someone else's info? I think this needs some more digging into the cache aspect.
> 
> I suggest starting with this page and leaning on @esizer 's experience with the URQL cache. https://nearform.com/open-source/urql/docs/graphcache/normalized-caching/

Yeah,  I figured you'd think that 😄 

That page has to do with `graphcache` which is a separate package. We've been considering/trying to add it but have yet to. For example
Attempted in https://github.com/GCTC-NTGC/gc-digital-talent/issues/8309 with https://github.com/GCTC-NTGC/gc-digital-talent/pull/11074
The spike had some discussion but got shot down too https://github.com/GCTC-NTGC/gc-digital-talent/pull/11074#issuecomment-2471374847

vd1992:
Not sure of another, not too complicated (finicky) idea

vd1992:
At a standstill for now. Did not delete for future reference if needed
Issue https://github.com/GCTC-NTGC/gc-digital-talent/issues/12492

PR #13770, opened 2025-06-06
[Feat] Improve type safety of localized enum query

### LINKED ISSUE #12978: ♻️ Improve localized enum type
## ♻️ Debt/Refactor

It would be nice to have stronger typing on the `localizedEnumStrings` query.

## 🕵️ Details

Currently the type returns the value as a string. This works but it would be nice to get it as the proper enum type. Also, it would cool if we could accept the enum as the argument instead of a string too.

## 🙋‍♀️ Proposed Solution

I don't know if this is possible but I tihnk we might be able to pull it off using interfaces or unions 🤔 .

### Interface

```graphql
interface LocalizedEnum {
  label: LocalizedString!
}

type LocalizedCSuiteRoleTitle implements LocalizedEnum {
  value: CSuiteRoleTitle! 
  label: LocalizedString!
}

type Query {
  localizedEnumStrings(enum: LocalizedEnum): [LocalizedEnum!]
}

query {
  localizedEnumStrings(enum: LocalizedCSuiteRoleTitle) {
    ...on LocalizedCSuiteRoleTitle {
       value
       label { localized }
    }
  }
}
```

### Union

Not sure  I like this because it will be a large union but we might be able to build it programatically like we already do with the actualy enum types.

```graphql
union LocalizedEnum = ....
```

## ✅ Acceptance Criteria

- [ ] Localized enum query has more accurate types
- [ ] Frontend updated with new types

esizer:
🤖 Resolves #12978 

## 👋 Introduction

Improves the type safety of our localized enum query.

## 🕵️ Details

Team decided to deprecate existing and only replace one instance to reduce the diff. This also cleans up our server provider and makes it a bit easier to understand (i hope).

The benefit is, now we just need to add a class in the `App\GraphQL\Types` namespace that implements `TypeRegistrarInterface` and it will automatically be registered.

Sadly, this is a little more boilerplate but avoids the type cast. Since the new query returns a union type, we need thta narrowed down. I created a helper that should help it out.

### Usage

```tsx
const EnumQuery = graphql(`
   query {
      localizedEnumOptions(enumName: "ArmedForcesStatus") { 
       ... on LocalziedArmedForcesStatus {
           value 
         }
      }
   }
`);

// Now, only armed forces types remain and value is typed properl
const armedForcedOptions = narrowEnumTypes(unpackMaybes(data?.localizedEnumOptions), "ArmedForcesStatus");
```

## 🧪 Testing

2. Confirm instances of localized enums still work
3. Confirm the work form stil works with the employment categories


yonikid15 (approved):
Looks good to me :tada: 
Hopefully we can get some more set of eyes to see if they approve of the solution.

PR #13776, opened 2025-06-09
[Bug] Add Null status message for Nominee when there is no current govt work experience

### LINKED ISSUE #13520: 🐛  Nominee Current position Null status - when the nominee is no longer govt employee

I had a nominee whose govt exp got expired by April 2025. 
Now his current position looks like this. 
I hope we can improve this for better communication like no longer govt employee or something like that 
Image

Also I'm unsure about letting the nomination process continue in this case 

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. If you have nominee already in your system, end their govt experience by last month or so. 
2. Go to Nominee profile page via Nomination event table
3. Observe the empty current position as above

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Current position section should communicate that the emp is no longer govt employee 
  - [ ] see message in https://github.com/GCTC-NTGC/gc-digital-talent/issues/13520#issuecomment-2912830779
- [ ] Nomination Evaluation button disabled ?

brindasasi:
🤖 Resolves #13520 

## 👋 Introduction

Currently the current position section checks only work experience & whether it is current. If doesn't it displays blank.
 
This PR changes it so that if the nominee doesn't have current **govt** work experience a message will show up requiring govt work experience. 

## 🧪 Testing

1. Test with having current valid govt work exp
2. expired govt work exp 
3. Current private work exp
 

## 📸 Screenshot

Screenshot 2025-06-09 at 10 53 49 AM


apps/web/src/utils/experienceUtils.tsx
```diff
@@ -528,6 +528,11 @@ export const isPersonalExperience = (
 export const isWorkExperience = (
   e: SimpleAnyExperience,
 ): e is Omit<WorkExperience, "user"> => e.__typename === "WorkExperience";
+export const isGovWorkExperience = (
+  e: SimpleAnyExperience,
+): e is Omit<WorkExperience, "user"> =>
```
petertgiles, line 533: I like this function but having it as a typeguard is a bit weird since there's no "GovWorkExperience" type in our schema that we're narrowing to.  I suggest removing the `e is Omit<WorkExperience, "user">` bit.

apps/web/src/pages/TalentNominations/NominationGroup/components/CurrentPositionExperiences.tsx
```diff
@@ -110,6 +157,20 @@ const CurrentPositionExperiences = ({
             data-h2-flex-direction="base(column)"
             data-h2-gap="base(x.5 0)"
           >
+            {sorted.length === 0 && (
+              <>
+                <p
```
petertgiles, line 162: This wasn't in the AC so not blocking, but we usually put these types of messages in a well.  I'd suggest a "warning" colour, but somethign like this:
image

petertgiles (approved):
Nice, works well!

apps/web/src/pages/TalentNominations/NominationGroup/components/CurrentPositionExperiences.tsx
mnigh: Reminder to use tailwind classes for anything new. I've made the commit: [30d352f](https://github.com/GCTC-NTGC/gc-digital-talent/pull/13776/commits/30d352f4fcc8b40b5c624d630801afa414772b9b)

apps/web/src/utils/experienceUtils.tsx
brindasasi: That does solve the typing error .. but that check is already inside the govWorkExperience

brindasasi: Ok I figure that I need this type guard in here `e is Omit<WorkExperience, "user">` so that it return as WorkExperience. Otherwise we would need that duplicate check for isWorkExperience just to return that type. 
Putting that back bit back.

Issue #12183, opened 2024-12-04
🐛 (Still) Slow Page - Screening


petertgiles:
## 🐛 Bug

Our logs are full of warnings about how slow the screening page is. This is causing poor UX, PHP worker shortage, and even worker timeout.

## ✅ Acceptance Criteria

- [ ] Performance of screening page is generally improved
- [ ] Criteria 2

esizer:
Do we have any insight on the DB layer of this? I feel like we keep revisiting this and nothing we do is working :sob:

tristan-orourke:
So, what did we change last time we looked at this?

tristan-orourke:
This could be slow:
1. in the database
2. in laravel
3. or transferring to the client

tristan-orourke:
Could use local logs to look at the database query, and run that query on prod database directly, to test timing there.

tristan-orourke:
```
<?php

namespace Database\Seeders;

use App\Enums\ApplicationStep;
use App\Enums\CandidateRemovalReason;
use App\Enums\PoolCandidateStatus;
use App\Models\Department;
use App\Models\Pool;
use App\Models\PoolCandidate;
use App\Models\User;
use Illuminate\Database\Seeder;

class MassivePoolSeeder extends Seeder
{
    /**
     * Seeds a large amount of candidates in a single pool for testing purposes.
     * These users are only useful for testing locally.
     *
     * @return void
     */
    public function run()
    {
        $pool = Pool::select('id', 'name')->whereNotNull('published_at')->firstOrFail();

        // Create many new applications
        User::factory()
            ->asApplicant()
            ->withSkillsAndExperiences()
            ->count(2000)
            ->afterCreating(function (User $user) use ($pool) {
                $this->applyToPool(
                    user: $user,
                    status: PoolCandidateStatus::NEW_APPLICATION->name,
                    pool: $pool,
                );
            })
            ->create();
    }

    private function applyToPool(
        $user,
        $status,
        $pool,
        $removalReason = null,
        $expiryDate = null,
        $priorityStatus = null,
        $priorityExpiry = null,
        $veteranStatus = null,
        $veteranExpiry = null,
        $placedAt = null)
    {
        // create a pool candidate in the pool
        PoolCandidate::factory()->for($user)->for($pool)
[...]

PR #13786, opened 2025-06-09
[Bug] Improve Talent placement performance

### LINKED ISSUE #13763: 🐛 Talent placement tab got really slow

Talent Placement tab will also take a while to load, and won't load more than 20 at a time. When table is set at 50 at a time will return zero result.

### 🕵️ Details

Poolcandidates pagninated query is timing out frequently in prod. Some incidents reported at 50 records , 100 records at a time. 
CandidatesTableCandidatesPaginated_Query

### 📋 Steps to Reproduce

I was able to reproduce it for 500s at a time in UAT and I'm sure if you adjust the network speed in browser this can be reproduced in local as well. 

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

https://github.com/GCTC-NTGC/gc-digital-talent/issues/12183#issuecomment-2573516555

### ✅ Acceptance Criteria

- [ ] Improve loading speed of Talent Placement table.

vd1992:
🤖 Resolves #13763

## 👋 Introduction

Try and improve the performance of `/en/admin/pools/:id/pool-candidates` AKA the "Talent placement" tab

## 🕵️ Details

Did two things

1. Shake down the page's query to clear out unneeded fields to reduce data fetched

2. Implement @esizer's suggestion https://github.com/GCTC-NTGC/gc-digital-talent/issues/12183#issuecomment-2573516555

Made some performance gains as a result

## 🧪 Testing

1. Have at least 500 candidates seeded for a pool
2. Navigate to the tab in question, set page size to 500
3. Compare and contrast main with the branch

Seeding
`make artisan CMD='db:seed --class=BigSeederPoolCandidateUser'`

## 📸 Screenshot

What I got

Transfer reduced, time not quite halved

image


api/graphql/schema.graphql
```diff
@@ -577,6 +603,10 @@ type PoolCandidate {
   priorityVerificationExpiry: Date
     @rename(attribute: "priority_verification_expiry")
     @canRoot(ability: "viewNotes")
+
+  # fields grouped by permission
+  viewNotes: PoolCandidateViewNotes @self @canRoot(ability: "viewNotes")
+  viewStatus: PoolCandidateViewStatus @self @canRoot(ability: "viewStatus")
```
vd1992, line 609: Wasn't sure if we want to deprecate and replace the individual fields or not everywhere, like for when accessed from single candidate operations. 
It would be a big diff I imagine, so it can be a separate issue if desired

vd1992: As seen with the Storybook struggles, it can easily get trickier than expected 😛

petertgiles:
@esizer could you please take a look at this, too?

Sigh, field level permissions burn us again. :grimacing:  In a perfect world, it would be nice if Lighthouse were smart enough to run the needed permission checks only once.  :face_exhaling:  Since it can't, it seems like this is the best way to optimize.  I guess any optimization is a performance tradeoff with simplicity.  I feel like we'll need to beef up the schema changes with a lot more explanatory documentation to help with that.

This is almost the same thing we've done with EmployeeProfile, right?  In that case we have a full Eloquent model and its own set of mutations and queries.  But it is just an overlay on the `users` table.  It also simplified permissions.

I wonder if we should check with @gobyrne on this before we finalize the solution.  It could be that we could better align this with business.  For example, we might be able to group these under one model instead of two.  With a more understandable name.

petertgiles:
Oh, also, I'm worried about leaving the fields in their original location as well.  That makes them available to be misused again.  It also could mess with the URQL cache.  But, as you said, will result in a much bigger diff.  :cry:

esizer:
Also, we need to keep in mind, every single time we check a permission here we have to load both the `pool.team` and `pool.community.team` to check for every `@can` for every item in the list so grouping similar items or avoid querying them in queries like this.

vd1992:
I mean, when you want to grab stuff from 8 tables, then fetch 500 records. You aren't getting great performance unless you skip Eloquent and Lighthouse to do DB queries. 
This is just a meaty table

PR #13876, opened 2025-06-17
[Fix] Mobile menu colour mismatch in safari

### LINKED ISSUE #13233: 🐛 Mobile menu colour mismatch
## 🐛 Bug

When in light mode, the mobile menu has white text on white background.

## 🦋 Expected Behaviour

Light mode menu should have dark text so it is legible.

## 🕵️ Details

- Operating System: Mac OS Ventura 13.0.1
- Browser: Safari version 16.1

## 📋 Steps to Reproduce

1. Navigate to the site
2. Ensure light mode is enabled
3. Open the mobile menu
4. reseize while the menu is open
5. Observe white text on white background (may not stay on Chrome or FF)

## 📸 Screenshot

Image

## 🙋‍♀️ Proposed Solution

M<aybe check the hydrogen breakpoints or theme colours? I am out of my depth here 😅 

## ✅ Acceptance Criteria

- [ ] Mobile menu text is legible/visible at any screen size and colour mode

esizer:
🤖 Resolves #13233 

## 👋 Introduction

Possible fix for the colours not matching when transitioning into the `sm` breakpoint.

## 🕵️ Details

We had a hook `useIsSmallScreen` that was using a dwebounced even listener watching for screen size changes and adjusting from `innerWidth`. This could have been causing issues in safari if the inner width was not updating in sync with the css media queries.

This refactors that hook to accept our tialwind sizes and react based on media query change rather than window size which should help it sync up with the CSS :thinking: 

## 🧪 Testing

1. Build the site `pnpm build:fresh`
2. Open the site at a size larger than 1080px width in safari
3. Reduce the window size until it collapses to a mobile menu
4. Confirm the background colour is corrext as well as links

### DIFF
--- a/packages/helpers/src/hooks/useIsSmallScreen.ts
+++ b/packages/helpers/src/hooks/useIsSmallScreen.ts
@@ -1,6 +1,6 @@
 import { useEffect, useState } from "react";
 
-export type Breakpoint = "xs" | "sm" | "md" | "lg";
+type Breakpoint = "xs" | "sm" | "md" | "lg";
 
 const breakpoints: Record<Breakpoint, string> = {
   xs: "48rem",
--- a/packages/helpers/src/hooks/useIsSmallScreen.ts
+++ b/packages/helpers/src/hooks/useIsSmallScreen.ts
@@ -1,22 +1,39 @@
-import { useLayoutEffect, useState } from "react";
-import debounce from "lodash/debounce";
+import { useEffect, useState } from "react";
 
-const useIsSmallScreen = (threshold = 768, wait = 250) => {
-  const [isSmallScreen, setSmallScreen] = useState(
-    window.innerWidth < threshold,
+export type Breakpoint = "xs" | "sm" | "md" | "lg";
+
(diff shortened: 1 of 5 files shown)


packages/helpers/src/hooks/useIsSmallScreen.ts
```diff
@@ -1,22 +1,39 @@
...
+export type Breakpoint = "xs" | "sm" | "md" | "lg";
+
+const breakpoints: Record<Breakpoint, string> = {
+  xs: "48rem",
+  sm: "67.5rem",
+  md: "80rem",
+  lg: "100rem",
+};
+
+export function useIsSmallScreen(threshold: Breakpoint): boolean;
```
mnigh, line 12: Why is the function `useIsSmallScreen` declared three times here and does it need to be exported?

mnigh (changes requested):
🥳 It works: the colours are correct and breakpoints break at the right place! Just a few small things to address.

mnigh (approved):
👏

PR #14133, opened 2025-07-10
[Feature] Improve the skill match dialog

### LINKED ISSUE #12876: ✨ Inaccurate skill matching results from Talent request candidates
## ✨ Feature

When looking at the list of candidates resulting in a talent request with multiple skills, the list often shows a number of skills matching, e.g. "3 out of 4".

But when clicking on the link to view the justifications, sometimes only a subset (2 out of 4) of those skills actually have a justification.

## 🕵️ Details

After reviewing the issue, it would appear that the 3 out of 4 is correct, but some of those skills have just been added through the skills portfolio section without adding a justification.

It would be good to make it clear in the dialog:
- Which skills are claimed with justification.
- Which skills are just claimed.
- Which skills are not claimed at all. 

1. We can do this by moving all NON claimed skills into a list below to separate them fully from the list the candidate has claimed. (Bonus points cause we use this pattern elsewhere).
2. We can then update the text in the yellow box to be more accurate like:

> {userFirstName} has added this skill to their skills portfolio, but it isn’t linked to a career experience.

## 🎨 Design

Figma file by @Jerryescandon 

Image

## 🛑 Blockers

```[tasklist]

### Blocked By
```

vd1992:
🤖 Resolves #12876

## 👋 Introduction

Improve the dialog through
- Ordering of skills; justified, claimed, and alphabetical sorting
- Unique well text for skills claimed in library but not connected to anything
- Basic list of skills selected for filtering that the user does **NOT** possess

## 🧪 Testing

1. Navigate to candidates table
2. Filter by skills, open and observe the dialog

## 📸 Screenshot

image


apps/web/src/components/Table/SkillMatchDialog.tsx
```diff
@@ -36,25 +59,110 @@ const SkillMatchDialogBody = ({
...
+    .map((skill) => skill.id);
+  const claimedButUnsupportedSkillIds = claimedSkillsIds.filter(
+    (skillId) => !supportedSkillsIds.includes(skillId),
+  );
+  const unclaimedSkills = filteredSkills.filter(
+    (skillToFind) => !userSkillsSkillIds.includes(skillToFind.id),
+  );
+
+  // sort, by claimed vs unclaimed, then by supported, lastly alphabetically
+  filteredSkills.sort((skillA, skillB) => {
```
vd1992, line 101: Order comes from 
https://github.com/GCTC-NTGC/gc-digital-talent/issues/12876#issuecomment-2996912789

apps/web/src/components/SkillTree/SkillTree.tsx
petertgiles, line 1: It doesn't seem right to me that skills that are not claimed appear in both the tree and the well at the bottom.  @Jerryescandon can you confirm?

petertgiles, line 1: I'm guessing we want some special handling for when there are no missing skills.  @Jerryescandon ? Hide the section?  Show a different message?

petertgiles (changes requested):
The code looks good.  :clap: I had a couple of question for Jerbo around some rough edges.

apps/web/src/components/SkillTree/SkillTree.tsx
Jerryescandon: Correct, if the candidate has all skills on their profile, then the unclaimed skills section should be hidden.

petertgiles (approved):
Looks great.  :+1:

PR #14151, opened 2025-07-16
[Fix] Announcer component redundant announcements

### LINKED ISSUE #13558: 🐛 `Announcer` announcing old message before new message

When you use the `useAnnouncer` hook to announce something to screen readers, if there is a stale message, it will announce that before the new message which is pretty confusing. 

### 🕵️ Details

**Operating system**: MacOS
**AT**: VoiceOVer

### 📋 Steps to Reproduce

This requires you have a page that does multipler announcemnets in a ro. I'm not sure where to go for that right now but here is how you can make some changed to reproduce this.

```tsx
const { announce } = useAnnouncer();
const [count, setCount] = useState(0);

const increment = () => {
  setCount((current) => {
    const newCount = count+1;
    announce(`${count} updates`);
    return newCount;
  });
}

return (
  <button onClick={increment}>Bump</button>
);
```

When you click the button you will hear:

1. 1 updates
2. 1 updates, 2 updates
3. 2 updates, 3 updates
4. etc.

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

Sadly, I'm not sure. Might take some investigation.

### ✅ Acceptance Criteria

- [ ] Announcer only announces the latest message

esizer:
🤖 Resolves #13558 

## 👋 Introduction

Fixes an issue where announcer was announcing old messages when new ones were added.

## 🕵️ Details

I think the issue was that we were just pushing out messages as they came in and letting it just announce them as is. So, the old message was still sitting in the announcer and some update happened and the DOM was modified and the container treated the old string moving out as relevant even though it was not.

The fix was to implement a queue where a message is added to the queue and then, after a timeout (1000ms) items are deuqued and sticthed together as a single message and announced in order of priority. This should prevent messages from being duplicated and also offers us the ability to prioritize messages.

> [!NOTE]
> This is not perfect, if the queued messages cannot be announced within the 1 second timeout and there are new incoming messages, the new messages with override the current messages. However, I tihnk this trade off makes sense for a few reason.
>
> Most messages should not take more than 1 second. Most uses of our announcer rarely ever send more than 1 message at a time.

The new flow should be:

Action - Message - Priority

 - queue "one" 0
 - Announce "one"
 - Queue cleared
 - Wait 1 second
     - queue "two" 0
     - queue "three" 1
     - queue "four" 2
 -  Announce "four. three, two"
 - Queue cleared
 - Wait another 1 second
     - No messages so remain idle

## 🧪 Testing

It is hard to get multiple messages to be read in short succession. What I did was add a button that forces announcements.

```tsx
const AnnouncerTest = () =>{
const { announce } = useAnnouncer();
const [val, setVal] = useState(0);

const handleClick = () => {
  setVal((old) => old + 1);
  announce("Message " + val + 1);
}

return <button onClick={handleClick}>Annoucne</button;
}
```

1. Build `pnpm dev:fresh`
2. Open the site with a screen reader on
3. Trigger multiple announcements
4. Confirm messages are announced in proper order and no redundant messages are announced


brindasasi (approved):
Announcer issue described in this issue seems to be fixed ! 👍
( I do see other issues with announcer .. will create sep issues for them )

PR #14153, opened 2025-07-16
[Fix] Attempt to improve performance of notification polling

### LINKED ISSUE #13869: 🐛 Polling can get slow with a many notifications

This is a reframing of ♻️ Prune notifications table [#12496](https://github.com/GCTC-NTGC/gc-digital-talent/issues/12496) after that got shot down in refinement.  😆   In that meeting it was suggested that some database optimization could help remediate.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

For a worst-case scenario, modify the big notification seeder to seed something like 10,000 notifications to a single user.

Image

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

- remove @orderyBy directive from endpoint
- separate query that just uses count instead of retrieving an actual notification

### ✅ Acceptance Criteria

- [ ] notification polling performance improved

esizer:
🤖 Resolves #13869 

## 👋 Introduction

This replaces the existing paginated query for notifications with a simple `@count`.

## 🕵️ Details

I made the changes we had but, it is actually slower :cry: I think we might need to dig in a find a better approach all together.

I don't think it is the database query itself but something inbetween :thinking: This is what is being generated and it seems straightforward. We also have indexs on `notifiable_type` and `notifiable_id` so that should help :woman_shrugging: 

 - Main: `~565ms`
 - PR: `~600ms`

```sql
# Runs in 0.01s locally
count(*) as aggregate from "notifications" where "read_at" is null and "notifiable_id" = ? and "notifications"."deleted_at" is null;
```

## 🧪 Testing

2. Seed a lot of notifications `make arisan CMD="db:seed --class=BigSeederNotifications`
3. Open the homepage
4. Confirm the time it takes for the `NotificationCount` query to run

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -1512,6 +1512,8 @@ type Query {
   ): [Notification!]
     @orderBy(column: "created_at", direction: DESC)
     @paginate(defaultCount: 10, scopes: ["authorizedToView"])
+  notificationCount(where: NotificationFilterInput): Int
+    @count(model: "App\\Models\\Notification", scopes: ["authorizedToView"])
 
   sitewideAnnouncement: SitewideAnnouncement
   talentNominationEvents(
--- a/api/storage/app/lighthouse-schema.graphql
+++ b/api/storage/app/lighthouse-schema.graphql
@@ -399,6 +399,7 @@ type Query {
   roles: [Role]!
   workStream(id: UUID!): WorkStream
   workStreams(talentSearchable: Boolean): [WorkStream]!
+  notificationCount(where: NotificationFilterInput): Int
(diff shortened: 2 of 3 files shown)


esizer:
> I'd recommend closing this PR and issue without merging. Sorry to waste the time. 😬

Yeah, thankfully the work was scoped out well and easy to complete to see if we would get any beenfits. :+1: 

> I wouldn't mind making it a little easier to get into this state though. Maybe a change `.env.example` that can be uncommented and some makefile magic?

That's probably a decent idea. Should make testing stuff like this a bit easier.

petertgiles:
Issue for testing tooling: https://github.com/GCTC-NTGC/gc-digital-talent/issues/14173

PR #14155, opened 2025-07-16
[Fix] Remove tiptap from manual chunk

### LINKED ISSUE #14077: 🐛 Tiptap bundle loaded on initial load

The tiptap bundle is loaded on the intial load, for example the homepage  This adds about 100 KiB of (usually) wasted data transfer.  Can we prevent it from loading initially if there is no banner message active?

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- Tiptap bundle not loaded intially unless there is an active banner message

esizer:
🤖 Resolves #14077 

## 👋 Introduction

Removes tiptap from from manual chunks to allow it to lazy load.

## 🕵️ Details

There is currently a bug in vite that forces all manual chunks to load before the entry point bundle. Since this is about 100kb and we don't really need to manual chunk it, removing it allows it to auto lazy load the resources.

```sh
# Branch
index.js   998.11 kB │ gzip: 308.87 kB │ map: 2,174.69 kB

# Main
tiptap.js  327.82 kB │ gzip: 104.56 kB │ map: 1,494.74 kB
index.js   998.43 kB │ gzip: 308.91 kB │ map: 2,174.69 kB
```

## 🧪 Testing

> [!TIP]
> Before testing, make sure you do not have a sitewide announcement. Having one will load tiptap dependencies since it needs it. 

1. Build main and take not of the bundle size for `index.js` and `tiptap.js`
2. Checkout branch and build `pnpm dev:fresh`
3. Confirm no `tiptap.js` bundle.
4. Confirm `index.js` has not increased in size too considerably
5. Load the home page
6. Check to make sure no tiptap deps were loaded

### DIFF
--- a/apps/web/vite.config.ts
+++ b/apps/web/vite.config.ts
@@ -80,12 +80,6 @@ export default defineConfig(({ command }) => ({
           graphql: ["@gc-digital-talent/graphql"],
           react: ["react", "react-dom"],
           router: ["react-router", "react-router"],
-          tiptap: [
-            "@tiptap/react",
-            "@tiptap/starter-kit",
-            "@tiptap/extension-link",
-            "@tiptap/extension-character-count",
-          ],
         },
       },
     },


esizer:
> This seems to work great! It does shave around 100KB off the load and that dependency bundle is gone.
> 
> So where did it go? If I browse into the announcement editor, the RichTextInput and RichTextRenderer bundles are only a little over 1KB each.

I think maybe there was a deeper issue with the manual chunks and this probably has more upfront savings :sweat_smile: 

My best guess is that grouping it as a chunk prevented it from tree-shaking? :thinking: Cause the renderer itself is only `1.3kb`.

I tried expermineting with removing some other chunks but did not see similar improvements (bytes just moved to `index.js`).

PR #14187, opened 2025-07-22
[Fix] Support arrays in protected user checker

### LINKED ISSUE #11149: 🐛 `ProtectedRequestUserChecker` has unexpected behaviour checking array of permissions
## 🐛 Bug

Our custom `ProtectedRequestUserChecker` class adds an extra layer to the Laratrust `isAbleTo` check. If a permission would require any role besides Applicant or Base roles, it fails unless the request came from the Protected endpoint.

However, in the case of an array of permissions being passed in, it seems to always fail. The default behaviour of this method is to pass as long as the user has _one_ of the permissions (unless a second parameter of `true` is passed in). 

## 🦋 Expected Behaviour

If an array of permissions is passed in, without a second parameter of `true`, then `ProtectedRequestUserChecker` should evaluate each permission separately. As long as any one permission belongs to the user and is unprivileged, then it should succeed even outside of a privileged request.

## 🕵️ Details

This is related to #11144 and #11143.

## 📋 Steps to Reproduce

1. Open a laravel tinker session `php artisan tinker`
2. Find applicant user `$applicant = User::where('email', 'applicant@test.com')->first()`
3. Compare the results of `$applicant->isAbleTo('view-own-application')` and `$applicant->isAbleTo(['view-own-application'])`

## 🙋‍♀️ Proposed Solution

Split the array internally (privileged and unprivileged) and check them separately.

## ✅ Acceptance Criteria

- [ ] If an array of permissions is passed in to `user->isAbleTo()`, without a second parameter of `true`, then `ProtectedRequestUserChecker` should evaluate each permission separately. As long as any one permission belongs to the user and is unprivileged, then it should succeed even outside of a privileged request.
- [ ] If the second param of `true` _is_ passed in, if should return true if and only if: 
  1. The user has every permission
  2. each permission is non-privileged OR the request came from the secure endpoint
- [ ] PHPUnit test that check that `isAbleTo` works as expected with arrays and single permission, on protected and non-protected endpoints

esizer:
🤖 Resolves #11149 

## 👋 Introduction

Updates the protected user checker to support array values.

## 🧪 Testing

1. Confirm new tests pass
2. Open a laravel tinker session `make artisan CMD="tinker"`
3. Find applicant user `$applicant = User::where('email', 'applicant@test.com')->first()`
4. Compare the results of `$applicant->isAbleTo('view-own-application') and $applicant->isAbleTo(['view-own-application'])`

### DIFF
--- a/api/tests/Unit/ProtectedRequestUserCheckerTest.php
+++ b/api/tests/Unit/ProtectedRequestUserCheckerTest.php
@@ -11,9 +11,6 @@ use Illuminate\Support\Facades\Request;
 use Illuminate\Support\Facades\Route;
 use Tests\TestCase;
 
-use function PHPUnit\Framework\assertFalse;
-use function PHPUnit\Framework\assertTrue;
-
 class ProtectedRequestUserCheckerTest extends TestCase
 {
     use RefreshDatabase;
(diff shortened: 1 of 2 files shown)


api/tests/Unit/ProtectedRequestUserCheckerTest.php
```diff
@@ -71,10 +77,9 @@ public function testCanUsePrivilegedPermissionProtected()
 
     public function testCanNotUsePrivilegedPermissionUnprotected()
     {
-        $this->markTestSkipped('Enable to test the route protection logic.');
-
         // simulate an unprotected request context
         Request::merge(['isProtectedRequest' => null]);
+        Route::expects('current')->andReturn($this->testRoute);
```
petertgiles, line 82: Could you add this to every test?  Otherwise, the logic isn't really being tested.

api/tests/Unit/ProtectedRequestUserCheckerTest.php
petertgiles, line 1: Could you please add a few more tests?  I think we're missing some cases.

Some ideas:
- mixed array, protected context
- mixed array, unprotected context
- mixed array, unprotected context, require all = true
- mixed array, protected context, require all = true
- privileged permission array, unprotected context, require all = true
- privileged permission array, protected context, require all = true

petertgiles (changes requested):
Thanks for taking this on!  :salute: Looking really, really good.  I think it just needs a bit more on the testing.

api/tests/Unit/ProtectedRequestUserCheckerTest.php
esizer: This was added in [refactor test to use data provider](https://github.com/GCTC-NTGC/gc-digital-talent/pull/14187/commits/ec8178674526af15f2ffe11aa3f3efaf8482813f)

api/tests/Unit/ProtectedRequestUserCheckerTest.php
esizer: Yeah, that makes sense. I think I got them all in [improve test structure, add additional tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/14187/commits/03779c673ee54c755c18b2f678feae2a5fb186d9)

PR #14196, opened 2025-07-22
[Bug] Adjust the scope used for filtering work streams

### LINKED ISSUE #14191: Combination of "Qualified Unavailable" status with any stream in candidates search tab produces no results.

When doing "Candidates" search in Community view and applying filters, the combination of "Qualified Unavailable" status and any stream will produce no results.
For example, if we don't select any stream, but select IT-02 and Qualified Unavailable, we will have a list of candidates in Software Solutions and Infrastructure Operations streams. But if we select Software Solutions or Infrastructure Operations (or both) we will have no results. 
This doesn't affect Qualified Available status. If we select Qualified Available, we can select any stream and get those results.

Expected behaviour - When choosing a stream, and Qualified Unavailable status, we should see candidates in that category (is there are any). 
In the case of Software Solutions and Infrastructure Operations, there are several. 

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1 - Go to Community
2 - Go to Candidates tab
3 - Go into search filters
4 - Apply any stream and Qualified Unavailable status.
5 - See zero results.
To check, remove stream, apply Qualified Unavailable, see the results (including the ones with the stream you previously selected). 

### 📸 Screenshot

Can provide screenshots if needed.
Please let me know. 

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

_No response_

vd1992:
🤖 Resolves #14191

## 👋 Introduction

Resolve the issue with filtering by stream

## 🕵️ Details

The scope being used was filtering by stream **_and_** status. This was probably due to being reused from somewhere/someplace where it was desired to do so. 
Another complicating factor is that the `ApplicantFilter` type is sensitive, not easy to tweak. Added stream filtering to a second place, the parent filter type. 

## 🧪 Testing

1. Test filtering in candidates tables
2. This includes multiple places, like a process, all candidates, and talent requests

Needed the filter type to populate the inputs but also act on the correct fields

## 📸 Screenshot

image


api/graphql/schema.graphql
vd1992, line 1: `ApplicantFilter.workStreams` would probably make more sense to be named `qualifiedWorkStreams` in line with `qualifiedClassifications`. Touching the filter type ripples elsewhere

As for the other type, I followed the pattern of `appliedClassifications`

image

brindasasi: I agree with that ... qualified workstreams makes more sense here

yonikid15 (approved):
LGTM

PR #14199, opened 2025-07-23
[Feat] New `Image` component

### LINKED ISSUE #14186: ♻️ `Image` component
### ♻️ Refactor

Create a new `Image` component to handle some common use cases.

### 🕵️ Details

The new image components goal is to allow for passing in a few props to control:

 - Lazy loading
 - Source sets

This should help with things like load time and layout shift.

### 🙋‍♀️ Proposed Solution

This new component should have an API similar to:

```tsx
type Breakpoint = "xs" | "sm" | "md" | "lg";
type PartialBreakpoints = Partial<Record<Breakpoint, string>>;

interface BaseImage {
  src: string; // Required base image
 sources: PartialBreakPoints;
}

// Width and height required by lazy loading
interface LazyImageProps {
  lazty: true;
  width: number;
  height: number;
}

interface EagerImageProps {
  lazy: false;
  width?: number;
  height?: number;
}
```

### ✅ Acceptance Criteria

- [ ] New image component created
- [ ] Implement is somewhere?

esizer:
🤖 Resolves #14186 

## 👋 Introduction

Adds a new `Image` component to handle source sets and lazy loading.

## 🕵️ Details

It looks like wdith and height are no longer a hard requirement for `loading=lazy` so I tweakerd this a bit. Instead of providing a `boolean` prop for the `lazy` properties, I simply use the `img` attributes to handle this to make it eaiser when passing in multiple sources.

We should still be providing these values when we can, however when you provide multiple sources, you can't really set them since each source would have a different width/height.

We can likely improve this over time if we feel the need.

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Open the network tab and filter by `img` requests
3. Navigate to the homepage
4. Observe, only those near the viewport are loaded
5. Scroll the page and notice that as images approach the viewport, they show up in the network tab and are loaded

### DIFF
--- a/apps/web/src/components/FeatureBlock/FeatureBlock.tsx
+++ b/apps/web/src/components/FeatureBlock/FeatureBlock.tsx
@@ -1,6 +1,6 @@
 import { ReactNode } from "react";
 
-import { Heading, Link } from "@gc-digital-talent/ui";
+import { Heading, Link, Image } from "@gc-digital-talent/ui";
 
 interface FeatureBlockProps {
   content: {
@@ -9,6 +9,8 @@ interface FeatureBlockProps {
     img: {
       path: string;
       position?: string;
+      width?: number;
+      height?: number;
     };
     link: {
       external?: boolean;
@@ -27,7 +29,8 @@ const FeatureBlock = ({ content }: FeatureBlockProps) => {
         </Heading>
       </div>
       <div>
-        <img
+        <Image
+          loading="lazy"
           src={content.img.path}
           alt={content.title}
           className="block h-auto w-full object-cover object-center"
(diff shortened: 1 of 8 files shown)


packages/ui/src/components/Image/Image.tsx
petertgiles, line 1: Could you please add a storybook story for this, to show how it can be used?

petertgiles (commented):
This is looking really cool!  I've read through it and left some thoughts.

apps/web/src/components/FeatureBlock/FeatureBlock.tsx
```diff
@@ -27,7 +29,8 @@ const FeatureBlock = ({ content }: FeatureBlockProps) => {
         </Heading>
       </div>
       <div>
-        <img
+        <Image
+          loading="lazy"
```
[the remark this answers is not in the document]
esizer: These are actualy below the fold on all screen sizes I tested. 

Depending on your veiwport, the lazy loading can be beneicial. If it happens to fall within the range of the loading taking place, it will just load it later (after an initial paint) and there shouldn't be much of a loss.

packages/ui/src/components/Image/Image.tsx
petertgiles: Looks so great, thanks.  BTW, axe-linter would like you to add alt text to it.  :laughing:

PR #14230, opened 2025-07-24
[Fix] Notification dialog poling removes focus

### LINKED ISSUE #13357: 🐛 Notifications panel pending spinner loses focus
> [!NOTE]
> This issue has been rewritten from the original based on later discussion.

Notification panel focus jumps to top of content whenever the pending spinner appears.  It should be unnecessary to cover the notifications list during periodic polling.

### 🕵️ Details

**Operating system**: Mac OS
**Browser**: Safari 
**Assistive Tech**: VoiceOver 

### 📋 Steps to Reproduce

1. Open notifications panel 
2. tab through notifications
3. wait for the focus to change to Top of notification panel (unprompted) 

### 📸 Screenshot

https://github.com/user-attachments/assets/0f9a453e-ea29-4925-bdb0-7d09d74fe859

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] After panel initially loads, no more pending spinners appear
- [ ] (Optional) If you remove the initial pending spinner, you can also close https://github.com/GCTC-NTGC/gc-digital-talent/issues/12295
- [ ] When the notification polling occurs, focus is not lost even if the list is updated

esizer:
🤖 Resolves #13357 

## 👋 Introduction

Fixes an issue where the notification dialog focus was reset when the polling query fired.

## 🕵️ Details

This results in a little pop-in since the query runs initially when opening but does improve the interactions after that.

I think we could reduce the pop-in by querying notifications outsiode the dialog and drilling them in, or using a cache-first strategy? :thinking: 

## 🧪 Testing

> [!TIP]
> I found it easiest to have two windows open, one to generate files for notifications and another with the dialog open to see it update after a poll.  You can aslo pause the queue, focus an item and then boot it back up so the next poll with have new notifications.

1. Build `pnpm dev:fresh`
2. Start queue so we can eaisly cvrete notifications `make queue-work`
4. Create some notifications by donwloading files
5. Open the notificaation dialog and focus an item
6. Create another notification
7. Confirm you do not lose your current focus position and the list just updates in place with no loading spinner

### DIFF
--- a/apps/web/src/components/NotificationList/NotificationListPage.tsx
+++ b/apps/web/src/components/NotificationList/NotificationListPage.tsx
@@ -5,7 +5,7 @@ import { useRef } from "react";
 
 import { Scalars, graphql } from "@gc-digital-talent/graphql";
 import { unpackMaybes } from "@gc-digital-talent/helpers";
-import { Link, Loading, Well } from "@gc-digital-talent/ui";
+import { Link, Well } from "@gc-digital-talent/ui";
 
 import NotificationItem from "./NotificationItem";
 import NotificationPortal, {
--- a/apps/web/src/components/NotificationList/NotificationListPage.tsx
+++ b/apps/web/src/components/NotificationList/NotificationListPage.tsx
@@ -78,9 +78,6 @@ const NotificationListPage = ({
 
   const notifications = unpackMaybes(data?.notifications?.data);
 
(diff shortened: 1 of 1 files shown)


bella77-69 (approved):
LGTM! 😎

PR #14243, opened 2025-07-28
[Fix] Handle null snapshots on download

### LINKED ISSUE #14150: 🐛 Having one of the application snapshot is null, crashes entire application download

It is possible to add an applicant directly to a pool in the system which will not generate snapshot. 
In such cases , trying to include these applications in any batch would crash the entire download process. 
This happens for Application Docx download only. 
Candidate CSV which downloads the same info into CSV file skips this null snapshot and produce the file without any error. 

It should be happening here as well 

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Manipulate the DB to make snapshot as null for any candidate. 
2. Try to download that with other candidates application docx 
3. Observe the process crashes 

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [] skip the application in the download batch  if the profile_snapshot is null

esizer:
🤖 Resolves #14150 

## 👋 Introduction

Updates the fiel downloads that require a profile snapshot to gracefully fail when it does not exist.

## 🕵️ Details

This simply throws an error and then is handled from the class that invokes the `generate` method.

- For a single document, we simply send an error message notification (toast)
- For a zip, we create a `failed.txt` file where each line is the profile that failed
  - To identify the user, we first attempt their name and then ID followed by "Unknown"

## 🧪 Testing

1. Open the database, find a few pool candidates and set their snapshot to null
2. Start the queue `make queue-work`
3. Attempt to download a single file and zip of the pool candidates
     - When doing the zip, include a few candidates that do have snapshots to confirm they still work
4. Confirm the operation fails gracefully


petertgiles (approved):
Works great!  Nice one.

PR #14261, opened 2025-07-28
[Bug] User profile download fixes

### LINKED ISSUE #14235: 🐛 Information on User profile download mismatching what's on the platform

1. In the user profile document there is a section like this which meant to list all the processes that user has been qualified on the platform

 **Recruitment processes**
Overview of recruitment processes in which <user> has been qualified.

This currently lists all applications of the user regardless of whether he is qualified or not 

2. Availability on the pool is shown as **Interest in job offers** in the document and currently it is mismatching with the platform 

This is in platform where it displays new application 
Image
Image

This is in the user profile document where it displays as qualified recruitment process and interest in job offers says No 
Image

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Login as applicant 
2. submit 2 applications 
3. Login as process operator
4. Qualify one of them 
5. Download user profile from the placement table 
6. Observe you get both applications under recruitment processes instead of only the qualified one 
7. Also notice the interest in job offers for the new application says No 

8. Now go and make the availablity for the qualified applciation as Inactive 
9. Now download the profile 
10. Observe the **Interest in job offer: yes** for that process in the download 
### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

[ ] Only qualified applications should be listed under Recruitment processes inside the user profile download
[ ] Availability should be reflected properly in the user profile download

vd1992:
🤖 Resolves #14235 

## 👋 Introduction

Makes some fixes to the doc download 

- Ensure only `OverallAssessmentStatus::QUALIFIED` candidacies are rendered
- Add a needed negation
- Expanded scope by resolving the querying issue

## 🧪 Testing

1. Pick or setup a user with multiple candidates, at least one of which is qualified
2. Check their availability too
3. Download and see what was rendered for the two points above

### DIFF
--- a/api/app/Traits/Generator/GeneratesUserDoc.php
+++ b/api/app/Traits/Generator/GeneratesUserDoc.php
@@ -585,7 +585,7 @@ trait GeneratesUserDoc
     /**
      * Generate all sections for a user
      */
-    protected function generateUser(Section $section, User $user, ?int $headingRank, User $authenticatedUser)
+    protected function generateUser(Section $section, User $user, ?int $headingRank)
     {
         if (is_null($headingRank)) {
             $headingRank = 2;
@@ -601,10 +601,10 @@ trait GeneratesUserDoc
             'workExperiences',
             'userSkills',
             'employeeProfile',
-            'poolCandidates' => function ($query) use ($authenticatedUser) {
+            'poolCandidates' => function ($query) {
                 /** @var \App\Builders\PoolCandidateBuilder $query */
                 $query
-                    ->whereAuthorizedToView(['userId' => $authenticatedUser->id])
+                    ->whereAuthorizedToView(['userId' => $this->authenticatedUserId])
                     ->whereQualified();
             },
             'poolCandidates.pool',
--- a/api/app/Traits/Generator/GeneratesUserDoc.php
+++ b/api/app/Traits/Generator/GeneratesUserDoc.php
@@ -26,6 +26,7 @@ use App\Enums\LearningOpportunitiesInterest;
 use App\Enums\Mentorship;
 use App\Enums\OperationalRequirement;
 use App\Enums\OrganizationTypeInterest;
+use App\Enums\OverallAssessmentStatus;
 use App\Enums\PositionDuration;
 use App\Enums\ProvinceOrTerritory;
 use App\Enums\SkillLevel;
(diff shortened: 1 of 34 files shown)


brindasasi (changes requested):
This restricts offer in progress applications as well . I'm not getting any of these applications in the user profile 
Screenshot 2025-07-29 at 10 56 29 AM
Screenshot 2025-07-29 at 10 56 36 AM

api/app/Traits/Generator/GeneratesUserDoc.php
```diff
@@ -597,8 +602,8 @@ protected function generateUser(Section $section, User $user, $headingRank = 2)
             'workExperiences',
             'userSkills',
             'employeeProfile',
-            'poolCandidates' => function ($query) use ($user) {
-                $query->whereAuthorizedToView(['userId' => $user->id]);
+            'poolCandidates' => function ($query) use ($authenticatedUser) {
+                $query->whereAuthorizedToView(['userId' => $authenticatedUser->id]);
```
vd1992, line 606: Needs to be the actor, not the target user

api/app/Traits/Generator/GeneratesUserDoc.php
```diff
@@ -585,8 +586,12 @@ protected function skillShowcase(Section $section, User $user, $headingRank = 3)
     /**
      * Generate all sections for a user
      */
-    protected function generateUser(Section $section, User $user, $headingRank = 2)
+    protected function generateUser(Section $section, User $user, ?int $headingRank, User $authenticatedUser)
     {
+        if (is_null($headingRank)) {
+            $headingRank = 2;
+        }
+
```
vd1992, line 594: Linting was being a little bit of an issue here, so I had to experiment

vd1992:
Got trickier than expected, good to go now I think?

brindasasi (approved):
Functionality looks good now .. thank you for addressing additional issues !👏⭐️
Although Eric has some suggestions that you might want to look into

esizer:
okay, I'll hopefuly be able to sum up my two comments here but it would affect a lot of this PR :sweat_smile: 

## Scope v Collection

Right now, we filter the collection which happens in memory. I think we could do this as a scope on `PoolCandidateBuilder` so that postgres only returns the candidates we care about rather than returning all of the authorized one and then we filter it some more in PHP.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/44992d4babe706209bf680c98bdf8315f2eb6062/api/app/Traits/Generator/GeneratesUserDoc.php#L682-L688

We could move this logic to the builder and call it on the `loadMissing` alongside `whereAuthorizedToView`So this line:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/44992d4babe706209bf680c98bdf8315f2eb6062/api/app/Traits/Generator/GeneratesUserDoc.php#L605-L607

So it would be:

```php
'poolCandidates' => function ($query) {
    $query
       ->whereQualified()
        ->whereAuthorizedToView(['userId' => $this->userIid]);
},

class PoolCandidateBuilder {
  public function whereQualified()
  {
     $this->where(function ($query) {
        $quety->whereNotNul('final_decision_at')
           ->whereIn('computed_final_decison', inalDecision::applicableToQualifiedRecruitment())
     });
  }
}
```

## Authenticated user (Actor)

This was an easy mistake becuase the name is very generic and is raisly confused. But, there was no need to pass this in anywhere since all file generators have a property `$userId` which is the currrent user passed in via the method `setUserId`.

You can see this in:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/44992d4babe706209bf680c98bdf8315f2eb6062/api/app/GraphQL/Mutations/DownloadUserDoc.php#L37

https://github.com/GCTC-NTGC/gc-digital-talent/blob/44992d4babe706209bf680c98bdf8315f2eb6062/api/app/Generators/ApplicationZipGenerator.php#L30

I would suggest, instead of passing it in a second time and having it twice as both `$authenticatedUser` and `$userId`, we instead just rename the property and method to something that is more descriptive of what it is. So we would replace:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/44992d4babe706209bf680c98bdf8315f2eb6062/api/app/Generators/FileGenerator.php#L101-L111

With something like:

```php
public functin setAuthneticatedUserId(string $id)
{
   $this->authenticatedUserId = $id;
   
   return $this;
}
```

vd1992:
Okay, made some big changes per Eric's big comment

esizer (approved):
Nice! 😎

PR #14263, opened 2025-07-29
[Debt] Reduce common image sizes

### LINKED ISSUE #14139: ♻️ Reduce some coomon image sizes
### ♻️ Refactor

_No response_

### 🕵️ Details

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

_No response_

esizer:
🤖 Resolves #14139 

## 👋 Introduction

Updates the images on some marketing pages to improve dimensions and compression.

## 🕵️ Details

I did not reduce the auth pages for the stesp and instead just removed the background colour and replaced it with transparency. Even though they have larger dimensions on desktop than how they are rendered, the esitmated size savings is incredibly negilible for a desktop site and it improved on mobile where they are closer to the actual size.

Due to this, I did not see the added complexity of `srcset` since we would only be saving about ~2kb per image at best.

swappy-20250729_115134

## 🧪 Testing

1. Build `pnpm build:fresh`
2. Navigate to each page with affected images
3. Confirm that lighthouse either does not mention the enw images or the size savings are negligble


apps/web/src/pages/Auth/SignUpPage/SignUpPage.tsx
vd1992, line 1: For a few places, like here, the total image transfer actually seems to have increased in size by ~ 100 KB
Transfer increased by a bit on `SignInPage` too. It appears to be the case for **_only_** light mode
Plus horizontal alignment broke

image

vd1992: Inspect view

image

vd1992: Storybook

image

vd1992 (changes requested):
For the most part, nothing appears broken and image transfer sizes were reduced. Got one or two places to double check though ⬇️

apps/web/src/pages/Auth/SignUpPage/SignUpPage.tsx
esizer: The file size is due to the introduction of transparency in the images. It added about 5-10kb to each image but, we have the added onus of not needing to match the colours exactly.

I was able to fix the imafge size though [fix no arrow image size](https://github.com/GCTC-NTGC/gc-digital-talent/pull/14263/commits/b7531c6142448da9e8d003a3f73f7290408e79de)

vd1992: Okay now that looks better

PR #14265, opened 2025-07-29
[Feat] Auto-register labels with `BasicForm`

### LINKED ISSUE #14228: ♻️ Improve `ErrorSummary` label detection
### ♻️ Refactor

Improve the way we detect input labels in the `BasicForm` and `ErrorSummary` components.

### 🕵️ Details

We have an `ErrorSummary` component we use in `BasicForm`. Since it is entirely decoupled from the actual for implementation, it is not aware of the input labels.

We need the labels in order to direct users to the input that has a specific error. Right now, we accomplish this by passing in an object where they key is the field name and the value is the label string.

The issue is, the label key must match the input name, if you make a mistake and name is anything else, the error summary will not show it,.

We would like to make it so it is very difficult or impossible for developers to make this mistake.

### 🙋‍♀️ Proposed Solution

We have two possible options,

 1. Using context to "register" labels long with the input
 2. Using some type constraints to enforce the match

## Context

We could employ some magic to help solve this via context. The idea would be to have our input components register their own labels with a context provider that wraps the `BasicForm` and therefore, the `ErrorSummary`. 

We could then access them anywhere inside the form context.

```tsx
import { createContext, ReactNode, useContext, useRef } from "react";

type FieldLabels = Record<string, ReactNode>;

interface LabelsContextState {
  labels: FieldLabels;
  setLabel: (name: string, label: ReactNode) => void;
}

const LabelsContext = createContext<LabelsContextState>({
  labels: {},
  setLabel: () => {
    // noop
  },
});

interface FormLabelsProviderProps {
  children: ReactNode;
}

export const FormLabelsProvider = ({ children }: FormLabelsProviderProps) => {
  const labelsRef = useRef<FieldLabels>({});

  const setLabel = (name: string, label: ReactNode) => {
    labelsRef.current[name] = label;
  };

  return (
    <LabelsContext.Provider value={{ labels: labelsRef.current, setLabel }}>
      {children}
    </LabelsContext.Provider>
  );
};

export const useFormLabels = () => useContext(LabelsContext);
```

Then, we could set the label in the inputs themsevles:

```tsx
const Input = (...) => {
  const { setLabel } = useFormLabels();
   setLabel(name, label);
   
   // ....
}
```

## Type constraints

```tsx
type BasicFormProps<TFieldLabels extends Record<string, ReactNode> {
  labels: FieldLabels;
}

type ObjectValues<T> = T[keyof T];
const FIELD_NAMES = {
  FIRST_NAME: "firstName",
}

type FieldName = ObjectValues<typeof FIELD_NAMES>;

type FieldLabels = Record<FieldName, ReactNode>;

const labels: FieldLabels = {
  [FIELD_NAMES.FIRST_NAME]: "First name",
}

const name = FIELD_NAMES.FIRST_NAME;
const label = labels[name];

<BasicForm labels={labels}>
```

### ✅ Acceptance Criteria

- [ ] Labels are derived semi or fully automatically reducing possibility for dev error
- [ ] review with the team to decide if it's worth merging

esizer:
🤖 Resolves #14228 

## 👋 Introduction

Adds context to the `BasicForm` to no longer need to supply specific labels via a prop.

## 🕵️ Details

Instead of requiring you to pass an object of `{ [name]: label }` where the `name` must match the input name exactly, we do this automatically in the underlying components.

This is a gap fill for when developers forget to supply or exactly match the input. It preserves the `labels` prop to allow overiding the auto-registered labels.

> [!NOTE]
> This is a sort of spike. If we do not like this, we can just close this and keep requiring developers to match the name/key exactly.  

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Login as any role
3. Go to your progile and create an error in "Personal information"
4. Attempt to submit and observe the error summary appears as expected
5. Start and application and continue to the questions page
6. Suybmit without entering anything
7. Observe the overridden labels for the form array still appear as expected

### DIFF
--- a/packages/forms/src/components/Checkbox/Checkbox.tsx
+++ b/packages/forms/src/components/Checkbox/Checkbox.tsx
@@ -8,7 +8,7 @@ import type { CommonInputProps, HTMLInputProps } from "../../types";
 import useInputDescribedBy from "../../hooks/useInputDescribedBy";
 import { checkboxRadioStyles, inputStateStyles } from "../../styles";
 import useFieldState from "../../hooks/useFieldState";
-import { useFormLabels, useRegisterFormLabel } from "../FormLabelsProvider";
+import { useRegisterFormLabel } from "../FormLabelsProvider";
 
 const checkbox = tv({
   extend: checkboxRadioStyles,
--- a/packages/forms/src/components/Combobox/Combobox.tsx
+++ b/packages/forms/src/components/Combobox/Combobox.tsx
@@ -18,7 +18,7 @@ import {
 import { BaseProps, ComboboxValue } from "./types";
 import Single from "./Single";
 import Multi from "./Multi";
-import { useFormLabels, useRegisterFormLabel } from "../FormLabelsProvider";
+import { useRegisterFormLabel } from "../FormLabelsProvider";
 
 export type ComboboxProps = Omit<HTMLInputProps, "ref"> &
   CommonInputProps & {
--- a/packages/forms/src/components/FormLabelsProvider.tsx
+++ b/packages/forms/src/components/FormLabelsProvider.tsx
@@ -2,7 +2,7 @@ import { createContext, ReactNode, useContext, useRef } from "react";
 
 const numberRegex = /(\d+)/g;
 
-export const normalizeName = (name: string): string => {
+const normalizeName = (name: string): string => {
   let normalized = name;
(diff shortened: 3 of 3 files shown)


yonikid15 (approved):
Looking good Eric!
This should help prevent any mismatching :+1: 

image


image

PR #14289, opened 2025-08-01
[Fix] Stabilize snapshot tests (hopefully)

### LINKED ISSUE #14253: 🧪 Stabilize file snapshot tests
### 🧪 Tests

We recently added a test for file snapshots. However, it does not seem to be stable.

### 🕵️ Details

It looks like dates being generated for experiences are not stable and can drift dramatically.

SEE:

https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/16568664211/job/46854876161

### 🙋‍♀️ Proposed Solution

We already set `$faker->seed(0)` so it could be somewhere else where the randomness is being introduced. Currently, I am not certain where this may be.

It might even be that a PR merged that affected this shortly after the snapshot was introduced and therefore was not failinfg so the base snapshot was not updated? 

#### State?

I think another good solution could be to add a `snapshot` state method that creates highly deterministic results from the factories.

```php
public function snapshot()
{
    return $this->state([
        'start_date' => '2020-01-01',
        'end_date' => '2050-01-01',
        'type' => EducationType::BACHELORS_DEGREE->name,
    ]);
}
```

### ✅ Acceptance Criteria

- [ ] Snapshot tests are stable

esizer:
🤖 Resolves #14253 

## 👋 Introduction

This is an attempt to fix the education requirement experiences for the application snapshot.

## 🕵️ Details

I'm hoping this does it but not sure. This is the easiest fix I could find that seemed to work without refacctoring more. 

I tihnk if this does not work, we might need to create state methods on factories to help create more deterministic results. It seems most of the instability is coming from the `afterCreating` and sadly, that method will always run unless we make some changes to that.

## 🧪 Testing

1. Run the appplication snapshot test a bunch (locally and CI)
2. Hope it passes consistently :crossed_fingers:

### DIFF
--- a/api/database/factories/PoolCandidateFactory.php
+++ b/api/database/factories/PoolCandidateFactory.php
@@ -167,16 +167,22 @@ class PoolCandidateFactory extends Factory
             if ($poolCandidate->education_requirement_option === EducationRequirementOption::EDUCATION->name ||
             $poolCandidate->education_requirement_option === EducationRequirementOption::PROFESSIONAL_DESIGNATION->name) {
                 // Ensure user has at least one education experience
-                $experience = EducationExperience::factory()->create([
-                    'user_id' => $poolCandidate->user_id,
-                ]);
+                $experience = EducationExperience::where('user_id', $poolCandidate->user_id)->first();
+                if (! $experience) {
+                    $experience = EducationExperience::factory()->create([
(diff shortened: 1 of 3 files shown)


petertgiles (approved):
Seems stable to me!  :+1:

PR #14297, opened 2025-08-05
Merge main into feature branch

vd1992:
## 👋 Introduction

Want to avoid feature branch falling too far behind, also want to get the infrastructure changes in because it causes inconvenience during branch hopping

I tried to merge main directly in at first in https://github.com/GCTC-NTGC/gc-digital-talent/pull/14296
but that clearly wasn't the right approach as I needed to handled merge conflicts too


vd1992:
> wow, had no idea it would be this large when I got the notitifcation 😆

Yeah, when I was hopping between branches I realized that handling all the merge conflicts at the end would possibly be horrible. That, and the infra/docker changes made branch hopping a pain

PR #14305, opened 2025-08-06
[Fix] Stabilize snapshots (again)

### LINKED ISSUE #14253: 🧪 Stabilize file snapshot tests

esizer:
🤖 Resolves #14253 

## 👋 Introduction

This is another attempt to stabilize the snapshots.

## 🕵️ Details

It looks like we had another date issue with the snapshots either from an un-updated snapdhot getting merged or the dates not being stable.

The one thing I noticed was the use of `$this->faker->boolean()` to optional set end date. I'm not sure if this is stable and it does seem like it is always end date having the issue.

Since this is a non-standard way of setting a field optionally, I updated this to use the `$this->faker->optional()` modifier instead. I think maybe this could be more deterministic? :thinking: 

## 🧪 Testing

1. Run snapshot tests multiple times
2. Confirm they consistnelty pass

### DIFF
--- a/api/database/factories/CommunityExperienceFactory.php
+++ b/api/database/factories/CommunityExperienceFactory.php
@@ -33,7 +33,7 @@ class CommunityExperienceFactory extends Factory
             'organization' => $this->faker->company(),
             'project' => $this->faker->bs(),
             'start_date' => $startDate,
-            'end_date' => $this->faker->boolean() ? $this->faker->dateTimeBetween($startDate) : null,
+            'end_date' => $this->faker->optional()->dateTimeBetween($startDate),
             'details' => $this->faker->text(),
         ];
     }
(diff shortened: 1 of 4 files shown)


vd1992 (approved):
Here is hoping

image

PR #14372, opened 2025-08-18
[Fix] Move `aria-label` for `ResourceBlock`

### LINKED ISSUE #14300: ⌨ Redundant link announcement in resource block

Our resource block has the list item role along with an aria label. The issue is, the aria label is mostly redundant causing the link test to be read out in duplicate.

### WCAG 2.1 Criteria

_No response_

### 📋 Steps to Reproduce

_No response_

### 🕵️ Details

The aria label is important becuase it adds information to the item (current state) however, since it is on a list item, it gets read out before the link and you hear the link text twice.

Eg. Link, Personal information, Personal information (complete), group

macOS

Safari

VoiceOver

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

I think it should be safe to move the `aria-label` to the link instead of the list item.

### ✅ Acceptance Criteria

- [ ] Link text is not duplicated

esizer:
🤖 Resolves #14300

## 👋 Introduction

This moves the `aria-label` for the `ResourceBlock` items from the list item to the link to avoid redundant announcements for ATs.

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Login as any user
3. Turn on a screen reader
4. Navigate to the resource blocks on the sidebar
5. Confirm the links do not have redundant announcements when tabbing to them


mnigh (approved):
I was able to replicate the issue on `main` with VoiceOver. This PR successfully eliminates the duplicate announcement: nice work!

PR #14449, opened 2025-08-26
[Feature] Combine email address save & verification into same dialog

### LINKED ISSUE #14092: ✨ Email verification dialog
### ✨ Feature Request

A new dialog that allows email input and verification from the same place.  This will NOT allow saving an unverified email address.

### 🕵️ Details

_No response_

### 🎨 Design file

https://www.figma.com/design/AQbFeF2lm4gZZJOwxOJTN2/Account-settings--All-users-?node-id=640-8379&m=dev

### 🎨 Designer

Josh

### 📸 Screenshot

_No response_

### 🌎 Localization

_No response_

### ✅ Acceptance Criteria

- [x] Dialog allows the email address to be updated only when the email is verified
  - [x] email address is populated in dialog if there is one already saved
- [x] Can be used for contact email and work email
- [x] Limit requests at the same rate as existing email verification pages
- [x] When entering a recognized domain for the contact email, it will also verify and update the work email address 
- [x] Existing verification logic for work email domains remains unchanged
- [x] when changing the email address after a code was requests, a message appears to request a new code 
- [x] make the dialog available on an unlinked URL (or url query parameter) to allow for testing

Out of scope:
- [ ] "locked" button variant.  The buttons should remain "normal".
- [ ] well styling, like the icon
- [ ] changes to uniqueness constraints

### 🚢 Definition of Ready

- [x] This issue has met the criteria in our Definition of Ready

### 🛑 Blocking Issues

Blocked by

brindasasi:
🤖 Resolves #14092 

## 👋 Introduction

Adds a dialog to save and verify email addresses at the same time.

## 🕵️ Details

Note that this doesn't officially live on any pages yet.  It can be accessed with a hidden flag on the account settings page.

Since we're now setting the email and verified_at simultaneously, I was worried about race conditions with the event handler that clears the verified_at when the email changes.  So I replaced the event handler with a custom mutator that clears the verified_at when the email changes.

I had to build all new backend features for this.  I started an issue to remove the original backend features for this when we're ready.
https://github.com/GCTC-NTGC/gc-digital-talent/issues/14578

There's a note in Figma that "ideally" the throttling message should show a live countdown.  Since that wasn't a firm AC for this issue I've moved it to a new issue.
https://github.com/GCTC-NTGC/gc-digital-talent/issues/14623

## 🧪 Testing

1. Add the GC Notify API key and notification IDs to the .env file.
2. Rebuild the app
3. Start the queue workers to send emails
4. Navigate to /en/applicant/settings?test-dialog-triggers=true (note the flag in the url query)

### Regression testing

- [ ] Can still save and verify email addresses in the "create user" onboarding pages
- [ ] Can still save and verify email addresses in the profile pages
- [ ] Changing an email address in the old way still clears the verified_at timestamp

### Dialog testing

- [ ] Satisfies all the AC in the original issue
- [ ] All the message wells appear and disappear as expected
- [ ] Unique constraints still enforced
- [ ] GC work email domains still enforced

## 📸 Screenshot

image


apps/web/src/components/EmailVerification/EmailVerificationDialog.stories.tsx
vd1992, line 1: I think you need the dialog to default to open 
Commented out bit intended to be kept?

image

apps/web/src/components/EmailVerification/EmailVerificationDialog.tsx
```diff
@@ -0,0 +1,540 @@
...
+    case EmailType.Contact:
+    default:
+      return intl.formatMessage({
+        defaultMessage:
+          "Add and verify a contact email that will be used for communication and notifications.",
+        id: "upuqPk",
+        description: "Contact email subtitle",
+      });
+  }
+};
```
yonikid15, line 100: Can these all be combined since they all repeat the same swtich case structure?

You could also just do something like 

```tsx
let description;
let label;
let subtitle;

if (emailType === EmailType.Work) {
  description = intl.formatMessage();
  label = intl.formatMessage();
  subtitle = intl.formatMessage();
} else {
  description = intl.formatMessage();
  label = intl.formatMessage();
  subtitle = intl.formatMessage();
}

```

apps/web/src/components/EmailVerification/EmailVerificationDialog.tsx
```diff
@@ -0,0 +1,540 @@
...
+              description:
+                "Body for a message informing the user that they changed the address and need to request a new code.",
+            })}
+          </p>
+        </Well>
+      );
+    default:
+      return assertUnreachable(message);
+  }
+};
```
yonikid15, line 167: Maybe this should go into it's own component.

apps/web/src/components/EmailVerification/EmailVerificationDialog.tsx
```diff
@@ -0,0 +1,540 @@
...
+              description:
+                "Body for a message informing the user that their contact email will be used as a work email.",
+            })}
+          </p>
+        </Well>
+      );
+    default:
+      return assertUnreachable(message);
+  }
+};
```
yonikid15, line 202: Maybe this should go into its own component as well

apps/web/src/components/EmailVerification/EmailVerificationDialog.stories.tsx
petertgiles: I had spun storybook out to https://github.com/GCTC-NTGC/gc-digital-talent/issues/14622 because it was broken while I wrote this.  It has since been fixed so I should come back to this.

apps/web/src/components/EmailVerification/EmailVerificationDialog.tsx
petertgiles: Subjective, but I usually prefer smaller functions with a single purpose that can be composed.  It is kind of ugly, though.  I refactored this to not need functions or case statements at all.  What do you think?
[refactor to message maps](https://github.com/GCTC-NTGC/gc-digital-talent/pull/14449/commits/c4eaf6d591bca9ec59d447c21548b62112d24566)

yonikid15 (approved):
Love the tests and the storybook additions :tada: 

One possible issue I noticed is the after successfully changing the email address in the dialog. The dialog closes. However, opening it up again shows the previous email.

I think getting the new email from the mutation result and displaying that would work (making sure to check the emailType). Over right the initialEmailAddress. I wrote some sudo below. It probably needs to be in a useEffect to work.

```tsx
// EmailVerificationDialog.tsx
...

let email = initialEmailAddress;
  if (data?.verifyUserEmails?.email || data?.verifyUserEmails?.workEmail) {
    email =
      dialogEmailType === EmailType.Contact
        ? data.verifyUserEmails.email
        : data.verifyUserEmails.workEmail;
  }

...

  const requestACodeFormMethods = useForm<RequestACodeFormValues>({
    defaultValues: {
      emailAddress: email ?? undefined,
      emailType: dialogEmailType,
    },
  });
  
```

This is definitely not a deal breaker on getting this through though so I'll Approve

petertgiles:
You were right, it needed a useEffect, not cache invalidation.  :laughing:   Would you mind taking a look?

apps/web/src/components/EmailVerification/EmailVerificationDialog.tsx
```diff
@@ -171,6 +170,16 @@ export const EmailVerificationDialog = ({
...
+  /*  Populate the initial email address in the form.
+   *  This can change if the email address is updated and the dialog is reopened.
+   *  Therefore, it is a useEffect instead of using ReactHookForms default values.
+   */
+  useEffect(() => {
+    if (initialEmailAddress) {
+      requestACodeFormMethods.setValue("emailAddress", initialEmailAddress);
+    }
+  }, [initialEmailAddress, requestACodeFormMethods]);
+
```
yonikid15, line 182: Okay this works on the initial dialog open. However, closing the dialog, then opening again, causes it to disappear.

I'm looking into it why now.

yonikid15 (changes requested):

yonikid15 (approved):
Looks good to go now :+1:

PR #14455, opened 2025-08-27
[Spike] Image attachments for support tickets

### LINKED ISSUE #13356: 🦔 How can we accept help desk screenshots.
### 🦔 Spike

This spike is to support https://github.com/GCTC-NTGC/gc-digital-talent/issues/12239

### 🕵️ Details

n/a

### ❓ Questions to answer

How can we accept a user screenshot when submitting a help ticket?

### 🕙 Timebox

TBD

petertgiles:
> [!WARNING]
> This is a spike.  It should not be merged to main.
 
🤖 Resolves #13356 

## 👋 Introduction

Adds the ability to attach a file to a Freshdesk support ticket.

## 🕵️ Details

To make this work, there are a couple of changes:
- a file input field added to the support form
- switching the form submission from a JSON object to a multi-part form.
- also switching to backend submission to Freshdesk to a multi-part form.

Considerations:
- One of the big original concerns was where to store the uploaded file.  It seems like in Laravel, the file comes in as an [UploadedFile](https://api.laravel.com/docs/11.x/Illuminate/Http/UploadedFile.html) stored in `/tmp` with a safe name, at least locally.  I'm not sure how this is managed or cleaned up.  It seems like it is part of Symphony and the file disappears after the request response is complete.  It seems to work fine in Azure as well.
- I didn't bother limiting by file type, mime type, or size.  The `UploadedFile` has `clientExtension()` and `getMimeType()` if we want server side filtering.  There's also `dimensions()` if we want to limit by image size or `getSize` to limit bytes.  
  - In [the Symphony docs ](https://symfony.com/doc/current/controller/upload_file.html)they suggest using mime type 
and `guessExtension()` instead of trusting the client.
- On the client side, there are many [possible limits](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/file) as well.
- It seems like the firewall has limits on upload size as well.  In dev, a 2MB image was rejected but a 500KB image was OK.
- I didn't bother figuring out multiple attachments.  It seems possible from the API docs and I don't think it would be hard.
- I didn't bother figuring out the array handling for multi-part forms so some properties are still commented out here.  It's probably not hard to figure out.

Next Steps:
- Do want to make a pretty/fancy file input component?
- Does the support team want to limit uploads by mime type, size, extension?
- Do we need to work with IMTD to up the firewall file size limit?
- It's not clear what the Freshdesk limitations are for attachments.
- If we wanted to go crazy with this we would probably handle the submission in an Azure Function instead of the app service with Laravel since there's no need for auth, graphql, the database, or really anything except the environment variables.

## 🧪 Testing

- Add the env variables for Freshdesk
- Rebuild and navigate to /en/support
- Submit a ticket with a file attached
    - Make sure you use your real email address and a testing note in the body so that the support team can respond to you.

## 📸 Screenshot

image

### DIFF
--- a/api/app/Http/Controllers/SupportController.php
+++ b/api/app/Http/Controllers/SupportController.php
@@ -5,7 +5,6 @@ namespace App\Http\Controllers;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Http;
 use Illuminate\Support\Facades\Log;
-use Illuminate\Support\Facades\Storage;
 
 use function Safe\parse_url;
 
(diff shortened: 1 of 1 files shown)


esizer:
> Do want to make a pretty/fancy file input component?

I would assume yes here. Up to the designers but I don't think we need to go wild here, just styling the browser one a bit should suffice.

> Does the support team want to limit uploads by mime type, size, extension?

I think we probably should. We don't want people uploading `.exe` files :sweat_smile: 

 > Do we need to work with IMTD to up the firewall file size limit?

If we are uploading screenshots we may want to. It will likely be a PNG which are usually larger in file size. Getting rejected at 2MB seems low for images (most very small screenshots I take are around 1MB) and I don't think we should expect users to be able to reduce file size. 

I tihnk if we could get it to 3MB we should be safe. That is still a pretty small limit and we can always go slow in increasing it.

> If we wanted to go crazy with this we would probably handle the submission in an Azure Function instead of the app service with Laravel since there's no need for auth, graphql, the database, or really anything except the environment variables.

That would be a nice long term goal. What you have done here so far looks great and something we can start with to get this up and running though :+1:

gobyrne:
Ok, Marc ensures me everything is in place to proceed.

PR #14489, opened 2025-08-29
Make email columns case sensitive

### LINKED ISSUE #13874: 🐛 Case-sensitive Email Uniqueness Check Allows Duplicate Emails

The email uniqueness check in the database allows duplicate email addresses when the case differs (e.g., test@example.com and TEST@EXAMPLE.COM are treated as different).
**Expected Result**: The system should prevent the creation of the second user, indicating that the email address is already in use.

**Actual Result**: The system allows the creation of both users, treating the email addresses as unique.

### 🕵️ Details

**Environment:**

Table: users
Column: email

### 📋 Steps to Reproduce

1. Create a user with the email test@example.com.
2. Attempt to create another user with the email TEST@EXAMPLE.COM

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ]  Duplicate email addresses should not be allowed regardless of their case

JamesHuf:
🤖 Resolves #13874.

## 👋 Introduction

Changes the email and work_email columns to the [citext](https://www.postgresql.org/docs/current/citext.html) type which is case insensitive.

## 🧪 Testing

1. Get a user with a capitalised email.
2. Attempt to set another users email to a lowercase version of that.
3. Confirm you get a uniqueness error.
4. Repeat for work emails.

## 🚚 Deployment

Any duplicate emails will need to be manually cleaned up **before** running the migration.

### DIFF
--- a/api/database/migrations/2025_08_29_220331_make_email_columns_case_insensitive.php
+++ b/api/database/migrations/2025_08_29_220331_make_email_columns_case_insensitive.php
@@ -1,0 +1,30 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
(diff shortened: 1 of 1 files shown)


bella77-69 (approved):
LGTM! 😁

PR #14508, opened 2025-09-03
[Spike] Server derived localized error messages

### LINKED ISSUE #14491: 🦔 Move a few API errors to localized enum
### 🦔 Spike

_No response_

### ❓ Questions to answer

Is this approach worth it?

### 🕙 Timebox

1 day

esizer:
🤖 Resolves #14491

## 👋 Introduction

This shows how we could use the new localized enums to return a localized error message. 

## 🕵️ Details

Honestly, we do not even need the enum backing this. We could simply use the built in localization. 

### Issues

I don't mind this approach because it could reduce our bundle since we don't need to have all messages bundled and shippped to the client and they are only sent when needed.

However,  we have a slight issue for when we re-use the server error message and the client one. So when we have matching validatio on the client as the server (we probably should for redundancy) than this gets less helpful because we would need that client message in the client and server so it severly diminishes the benefit here.

Sadly, we might just need to keep it returning just the error code since the message itself should probably also exist in the client and this would just result in duplication just so it gets caught in our checks.

Though, moving this to an actual enum and implementing a best practice in casing and location would be helpful.

### Solutions

Since I don;'t see a way out of not needing to bundle error messages in the client bundle I think the main benefit would now be from centralizing our error codes so we could:

 - Keep the consistent and centralized (right now we have a mix of cases and locations)
 - Have a defined list of expected errors so we could possibly use that in some check to confirm we are not missing any on the client (maybe :thinking: )

## 🧪 Testing

1. Test both mutations that were updated
2. Force an error on both
3. Confirm the error message returned is from the server
4. Change the language by setting the `Accept-Language` header to either `en` or `fr`
5. Confirm the message language changed

### DIFF
--- a/api/app/GraphQL/Validators/RevertFinalDecisionValidator.php
+++ b/api/app/GraphQL/Validators/RevertFinalDecisionValidator.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace App\GraphQL\Validators;
 
-use App\Enums\ErrorCode;
 use App\Enums\PoolCandidateStatus;
 use App\Models\PoolCandidate;
 use Nuwave\Lighthouse\Exceptions\ValidationException;
@@ -30,7 +29,7 @@ final class RevertFinalDecisionValidator extends Validator
         ];
 
         if (! (in_array($candidate->pool_candidate_status, $statusesArray))) {
-            throw ValidationException::withMessages(['id' => ErrorCode::INVALID_STATUS_REVERT_FINAL_DECISION->localized()]);
+            throw ValidationException::withMessages(['id' => __('error_code.invalid_revert_final_decision')]);
         }
 
         return [];


vd1992:
Yeah, things are definitely a bit scattershot. We've got
`errors.php`
`ApiErrorsEnum.php`
And sometimes the message is emitted in a class/method itself like in some validators and rules

Centralization alone would be a benefit

petertgiles:
I guess the two different validators are to show two different possible approaches?

I agree with the idea that we should probably standardize on an approach and bring some consistency to this topic.  I think where maybe I disagree is on the importance of having them perfectly aligned with the frontend.  I would guess around 90% of our errors are impossible to trigger from the UI and spending a bunch of time and ceremony on these is not super valuable.

I like the version you have in RevertFinalDecisionValidator.  Ditch the enum, ditch the error code.  Just send a localized message to the frontend that the UI can toast.  Let the client have its own error messages that it uses for client-side validation only.

PR #14542, opened 2025-09-04
[Fix] Remove migration to citext columns

### LINKED ISSUE #14539: 🐛 Undo migration to citext columns

We recently changed the email and work email columns to citext fields in #14489. We have since realised this will fail in prod since the citext extension is blocked by IMTD. We now need to back out this migration.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

Back out the bad migration to prevent issues when running it in production.

### ✅ Acceptance Criteria

_No response_

JamesHuf:
🤖 Resolves #14539.

## 👋 Introduction

Removes the migration to citext columns since it won't currently work in prod.

## 🧪 Testing

1. Refresh the site
2. Confirm viewing/updating emails still works as expected


petertgiles (approved):
Thanks for the quick fix.  ❤️

Issue #14545, opened 2025-09-04
🐛 Cannot use assessment step order swap mutation


vd1992:
### 🐛 Bug and Expected Behaviour

The step order swap mutation does not work, it returns a uniqueness constraint violation

### 📋 Steps to Reproduce

1. Have a process, open the assessment plan editor
2. Have 4 or more steps, try swapping order

### 📸 Screenshot

<img width="910" height="570" alt="Image" src="https://github.com/user-attachments/assets/8765d28c-df28-4c69-b014-3fc7e6d198a4" />

<img width="546" height="192" alt="Image" src="https://github.com/user-attachments/assets/b867a4b2-7740-45c9-b1e9-f71cae9de332" />

<img width="675" height="392" alt="Image" src="https://github.com/user-attachments/assets/9112d13f-1759-4649-ad84-c60cb4f131af" />

### 🙋‍♀️ Proposed Solution

- set one to zero while swapping
- postgres deferrable indexes

### ✅ Acceptance Criteria

- [ ] Mutation functions, swapable steps can swap again

PR #14546, opened 2025-09-04
[Bug] Assessment step order logic needed improvement

### LINKED ISSUE #14373: 🐛 Step name discrepancy in ROD

- ~~When in Step 2, the step name disappears~~
- The concatenated "Step number" + "Custom step name" is incorrect.
- This happens only on processes that do not have screening questions as an assessment

See video for walkthrough.

**Notes from design:** Jerbo and Gray think this might be related to the "Screening Questions" which have weird logic tied to them. And dont always appear on assessment plans. so they could be the culprit behind some of this.

### 📋 Steps to Reproduce
1. Create process with 3 assessment steps and atleast 3 essential skills
2. Submit application to this process
3. Assess Step 1 and observe step name changes
4. Assess Step 2 and before finishing all 3 skills observe the step name disappears

### 📸 Screenshot

https://github.com/user-attachments/assets/35cd8dee-10ec-4b20-837c-b853493c4b8d

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

_No response_

- [ ] When candidate is in Step 2, title should be visible, as it is visible in other steps.
- [ ] Step full title (number + name) should properly match step number to current step custom name

Blocked by: #14333

vd1992:
🤖 Resolves #14373

## 👋 Introduction

The automatic sort order value wasn't always set with the correct number, this stems from special cases like screening questions always needing to be second.
The incorrect step sort value would then cause trickle down confusion, like current step calculation 

## 🧪 Testing

1. Create a process and go to edit the assessment plan
2. Make many changes and keeps tabs on step sort order
3. One way to keep track of that is adminer -> assessment_steps table -> filter for the process in question

There is also another existing bug to be aware of https://github.com/GCTC-NTGC/gc-digital-talent/issues/14545

### DIFF
--- a/apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.tsx
+++ b/apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialog.tsx
@@ -8,6 +8,7 @@ import {
   graphql,
   Maybe,
   Scalars,
+  SkillCategory,
   User,
 } from "@gc-digital-talent/graphql";
 import { unpackMaybes } from "@gc-digital-talent/helpers";
@@ -52,6 +53,9 @@ export const ScreeningDecisionDialog_Fragment = graphql(/** GraphQL */ `
           id
           skill {
             id
+            category {
+              value
+            }
           }
           ...ScreeningTriggerPoolSkill
           ...ScreeningDialogHeaderPoolSkill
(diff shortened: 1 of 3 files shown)


api/app/Models/AssessmentStep.php
```diff
@@ -90,18 +90,27 @@ protected static function boot()
...
+                            $existingStep->save();
+                        }
+                    }
+                }
                 $sortOrder = 2;
             } else {
-                $highestSortOrderStep = AssessmentStep::where('pool_id', $step['pool_id'])->orderBy('sort_order', 'desc')->first();
-                $highestSortOrder = isset($highestSortOrderStep) ? $highestSortOrderStep->sort_order : 0;
-                if ($highestSortOrder < 3) {
-                    $sortOrder = 3;
```
vd1992, line 101: Probably the main thing, the second step being defaulted to sort order three. Likely the source for issues, as you'd have the application screening as step 1 then the next step as sort order 3 but nothing in between

brindasasi:
The original problem of current step calculation is not yet solved. 
Watch the video as below : 

https://github.com/user-attachments/assets/21e083c1-9e79-4fc5-b9b4-2c6141dca76b

PR #14572, opened 2025-09-09
[Debt] Reduce fetching on profile page

### LINKED ISSUE #13685: ♻️ Remove over fetched query data on Profile page
### ♻️ Debt/Refactor

- Remove over fetched data from the `UserProfile` fragment in the file [ProfilePage.tsx](https://github.com/GCTC-NTGC/gc-digital-talent/blob/13294-e2e-add-community-interest/apps/web/src/pages/Profile/ProfilePage/ProfilePage.tsx#L417-L789).
- The fragment is used in the [Application_PoolCandidate](https://github.com/GCTC-NTGC/gc-digital-talent/blob/13294-e2e-add-community-interest/apps/web/src/pages/Applications/fragment.ts#L3-L13) fragment. Therefore, this will require moving fields around / removing fields.

### 🕵️ Details

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Unused fields are removed from fragment
- [ ] Both UserProfile and Application_PoolCandidate fragment are consolidated

esizer:
🤖 Resolves #13685 

## 👋 Introduction

Updates the `ProfilePage` and its components to use fragments to ensure no under or over fetching is happening. 

## 🕵️ Details

Opted to use fragments which expanded the scope a bit but this was one area that we never got the chance to migrate due to the constraint around that profile snapshot test not allowing it. This seemed like the best approach to address the issue at the same time as updating this page  and its components to use our current approach.

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Login as `admin@test.com`
3. Navigate to your profile page
4. Edit each section confirming it still displays the correct info
5. Apply to a process
6. Cojnfirm the admin view of the application still renders the snapshot properly
7. Confirm the admin view of the profile still renders properly


apps/web/src/components/Profile/components/GovernmentInformation/Display.tsx
```diff
@@ -3,42 +3,60 @@ import { useNavigate } from "react-router";
...
-    isWorkEmailVerified,
-  },
+  query,
   showEmailVerification = false,
   readOnly = false,
 }: DisplayProps) => {
   const intl = useIntl();
   const navigate = useNavigate();
   const routes = useRoutes();
+  const user = getFragment(GovernmentInformationDisplay_Fragment, query);
```
petertgiles, line 55: You could make smaller diffs if you deconstructed user like you did in most of the other Display components.

apps/web/src/components/Profile/components/PersonalInformation/PersonalInformation.tsx
```diff
@@ -34,13 +39,44 @@ const PersonalInformationForm_Query = graphql(/* GraphQL */ `
   }
 `);
 
+const ProfilePersonalInformation_Fragment = graphql(/** GraphQL */ `
+  fragment ProfilePersonlaInformation on User {
```
petertgiles, line 43:
```suggestion
  fragment ProfilePersonalInformation on User {
```

apps/web/src/pages/Applications/ApplicationProfilePage/ApplicationProfilePage.tsx
```diff
@@ -84,7 +84,7 @@ export const ApplicationProfile = ({ application }: ApplicationPageProps) => {
   };
 
   const sectionProps = {
-    user: application.user as UserProfileFragment,
+    query: application.user as UserProfileFragment,
```
petertgiles, line 87: Since you're working on fragments... any chance we can get rid of this cast?

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/ApplicationInformation/ApplicationInformation.tsx
```diff
@@ -113,7 +125,7 @@ interface ApplicationInformationProps {
   applicationQuery: FragmentType<
     typeof ApplicationInformation_PoolCandidateFragment
   >;
-  snapshot: User; // recreated from Json
+  snapshot: User & { bilingualEvaluation?: BilingualEvaluation }; // recreated from Json
```
petertgiles, line 128: Is this because of snapshot versioning?  If so, I think we need to approach this in a more extensible way.  Maybe multiple copies of the component with different props for different versions?  Since the snapshot isn't actually coming from the API directly is it every worth doing fragments?  Or just using them as types?

petertgiles (commented):
Looking very cool  Just finished a readthrough and some comments.

apps/web/src/components/Profile/components/GovernmentInformation/Display.tsx
esizer: Ahh yes, thanks for catching that. I regretted this decison and forgot to come back and fix these :sweat_smile: 

Reverted in [revert to deconstructing](https://github.com/GCTC-NTGC/gc-digital-talent/pull/14572/commits/1241b6a894f0d64931e5408767aed0aec08b04dd)

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/ApplicationInformation/ApplicationInformation.tsx
esizer: Huh, I think we may need to dig deeper into the snapshot display again :cry: 

This is very similar to the other comment about the casting. The problem here is that the component that has the fragment is identical to the one we want to display here visually (unless of course you have an out of date snapshot).

So the fragments are very helpful when we can use them (showing the current profile) but get trickier when we cant (within the application).

Maybe we just need two versions, a set of components that render snapshot data with all the special handling for versions and legacy data and another version that is leaner and just hanldes the current user data model :thinking:

PR #14626, opened 2025-09-12
Update feature.yml

gobyrne:
## 👋 Introduction

Gray modifying feature issue template to suit himself.  I found myself doing this on multiple issues by hand, so might as well update the template. No reason we shouldn't use the H2 to organize either.

1. Adding in headings at level 2 
2. Reducing overall number of inputs
3. Reordering inputs to be a bit more logical too

## 🧪 Testing

Yeah, so I don't know how to actually view what the modified template will look like... maybe you do?


petertgiles (approved):
👍

PR #14671, opened 2025-09-19
[Feature] Log queries for more roles

### LINKED ISSUE #12581: ♻️  Add logging for all laravel queries with their payload info
## ♻️ Debt/Refactor

We are logging in all queries made by Platform admins. 
But we often need to know about the queries that are done by other people ( recruiters, coordinators). 
Can we just log all queries ? 

## 🕵️ Details

Should follow the same pattern as platform admin query logging

## ✅ Acceptance Criteria

- [ ] All graphql queries should be logged in with their payload info for users with roles higher than "applicant"

vd1992:
🤖 Resolves #12581

## 👋 Introduction

Log the queries for all elevated roles

I did a little tweaking. Rather than log per role, I combine all the roles into a single array so each request has all possessed roles. As well, I noticed notification query logging is pretty spammy, I don't think we should fill our logs with that

## 🧪 Testing

1. Login as various roles
2. Visit some admin pages
3. Observe logging

## 📸 Screenshot

image


brindasasi (approved):
Very useful! Thanks for this

PR #14686, opened 2025-09-24
[Fix] Resolve deleted education requirement experiences in snapshot

### LINKED ISSUE #14629: 🐛 Education requirement experiences can be deleted

After applying to a process with specific education requirement experiences selected, a user can delete those experiences meaning they can no longer show up for screening.

### 🕵️ Details

The issue here is that, while do snapshot the experience, we lose the reference to it on the application. We need that reference to perform the lookup in the snapshot for the appropriate experience.

 This is a many to many relationship and contains a pivot table. The row in that pivot table is not deleted, just the experience itself.  While the row exists, eloquent can't resolve the relationship because the experience has been removed and it never gets included in the query.

We will need to include some reference so we can use it as a lookup even after the experience has been removed.

### 📋 Steps to Reproduce

1. Submit an application
2. Delete any experience you attached in the education requirements
3. Login to view the application in the admin view
4. Open the education screening dialog
5. Observe the experience no longer show

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

We have the database row so we can add an additional attribute to `PoolCandidate that returns the IDs instead of getting the full experience.

I'm not sure if eloquent can do this or not but that would be ideal if it could. If not, you can use a raw query

```php
public function educationRequirementExperienceIds(): Attribute
{
  return DB::table('pool_candidate_education_requirement_experience')
    ->where('pool_candidate_id', $this->id)
    ->pluck('experience_id');
}
```

We can then update the `ScreeningDecisionDialog` to use this new attribute for a more stable lookup.

### ✅ Acceptance Criteria

- [ ] Deleted experiences still appear in supporting evidence for screening and assessment

esizer:
🤖 Resolves #14629 

## 👋 Introduction

Fixes an issue where the snapshot could not resolve the experiences marked as education requirements after they had been deleted.

## 🕵️ Details

This is accomplished by adding an attibute to get the IDs for those experiences from the pivit table since it still exists even when the referenced table row does not.

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Login as `admin@test.com`
3. Apply to a process taking note of the experiences you used in the educaiton requirement section
4. Delete all your experiences
5. View the  application as an admin
6. Confirm the experiences from the snapshot are still rendered even when the live versions do not exist

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -634,6 +634,8 @@ type PoolCandidate {
     @with(relation: "educationRequirementEducationExperiences")
     @with(relation: "educationRequirementPersonalExperiences")
     @with(relation: "educationRequirementWorkExperiences")
+  educationRequirementExperienceIds: [UUID!]
+    @rename(attribute: "education_requirement_experience_ids")
   # record of decision
   assessmentResults: [AssessmentResult]
     @hasMany
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -1,4 +1,4 @@
-"A date string with format `Y-m-d`, e.g. `2011-05-23`."
+"A date string with format `Y-m-d`, e.g. `2011-05-23`."schema
 scalar Date @scalar(class: "Nuwave\\Lighthouse\\Schema\\Types\\Scalars\\Date")
 
 "A datetime string with format `Y-m-d H:i:s`, e.g. `2018-05-23 13:43:32`."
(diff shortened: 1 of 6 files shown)


apps/web/src/components/ScreeningDecisions/SupportingEvidence.tsx
petertgiles, line 1: Could you please check Chromatic?  It looks like something might have broken in ScreeningDecisionDialog.

petertgiles: Seems to work OK in the app, though.

api/graphql/schema.graphql
```diff
@@ -634,6 +634,8 @@ type PoolCandidate {
     @with(relation: "educationRequirementEducationExperiences")
     @with(relation: "educationRequirementPersonalExperiences")
     @with(relation: "educationRequirementWorkExperiences")
+  educationRequirementExperienceIds: [UUID!]
```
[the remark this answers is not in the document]
esizer: Great idea, added some documentation in [document new field](https://github.com/GCTC-NTGC/gc-digital-talent/pull/14686/commits/ca9cafdd4305a140b92098f87f4725b7db2090ac)

PR #14713, opened 2025-09-26
[Fix] Application screening test assessment status computation

### LINKED ISSUE #14667: 🐛 Application Screening step is not getting completed even though all essential assessments are done in that step

In prod current Step is showing Step 1 -Application Screening even though that step has been completed

Image
Note the step 1 icon is also wrong as if it is not yet completed.

Expected Behaviour : In this case, current step should show Step 2 - Individual Interview

### 🕵️ Details

I think what is happening here is , there is a behavioural skill as essential and that will not be assessed in Application Screening section. 
We recently fixed this.
But we missed the step completion logic here. 

In this case , since we don't assess behavioural skill in the first step, this step should be considered complete. 

### 📋 Steps to Reproduce

1. Create an assessment with essential behavioural Skill along with other tech essential skills
2. Assess all essential tech skills in the first step 
3. Observe step has not been completed and still stays at 1 

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Application Screening assessment step should be completed once all essential tech skills are assessed

esizer:
🤖 Resolves #14667 

## 👋 Introduction

Updates the computation of the assessment status for application screening to skip essential skills if they are behavioural.

## 🧪 Testing

1. Login as `admin@test.com`
2. Create a process that has at least 2 assesment steps including essential behavioural skills
3. Apply to the process created in step 2
4. Assess the application created in step 3
5. Confirm you do _**not**_ need to assess any essential behavioural skills to pass the first step only

## 📸 Screenshot

swappy-20250926_083426

### DIFF
--- a/api/app/Models/PoolCandidate.php
+++ b/api/app/Models/PoolCandidate.php
@@ -14,6 +14,7 @@ use App\Enums\OverallAssessmentStatus;
 use App\Enums\PoolCandidateStatus;
 use App\Enums\PoolSkillType;
 use App\Enums\PriorityWeight;
+use App\Enums\SkillCategory;
 use App\Observers\PoolCandidateObserver;
 use App\Traits\EnrichedNotifiable;
 use App\ValueObjects\ProfileSnapshot;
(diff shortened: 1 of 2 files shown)


brindasasi (approved):
Screenshot 2025-09-26 at 3 59 48 PM
Verified essential behavioral skill doesn't need to be assessed in first step but required in other steps to complete the step.
Works great! 👍

PR #14725, opened 2025-10-01
[Debt] Create components for snapshot display

### LINKED ISSUE #14610: ♻️ Create snapshot display components
### ♻️ Refactor

We should probably create some standalone components for rendering snapshot data along with version handling.

### 🕵️ Details

We currently re-use the components for displaying a users current profile and their snapshotted profile. This has been fine but is starting to be a little more difficult since we are now introducing fragments to that area. This has forced us to make some awkward decisons in casting and increasing branching logic.

As the older snapshots diverge, we may want a better way of implementing the versioning and branching without poluting components that want to display the most up-to-date user model.

The idea here is that we have one set of components that deal with the snapshotted user model that would accept some sort of nebulous user type that comes from the snapshot JSON and then checks for fields with all the appropriate branching logic. The other could then drop all this and only deal with the current user model.

### 🙋‍♀️ Proposed Solution

We would want to take our existing `~/components/Profile/*` and extract out any special handling for snapshot data and add that into new snapshot components. Since users should never be able to edit a snapshjot, we can drop the logic for toggling a form and do a display only component.

```
if(version == null and bilingualEvaluation == null) <Version0>
else if (version == null and bilingualEvaluation != null) <Version0WithEvalution>
else if(version == 1) <Version1>
else if (version == 2) <Vesion2>
```

REF: https://github.com/GCTC-NTGC/gc-digital-talent/pull/14572#discussion_r2337616552

### ✅ Acceptance Criteria

- [ ] New snapshot based user profile components created
- [ ] Existing profile components no longer have special handling

esizer:
🤖 Resolves #14610 

## 👋 Introduction

This creates specific components for displaying snapshot data rather than using the current components.

## 🕵️ Details

This should hopefully simplify the base components for displaying profile data and hide the complexity of legacy data behind a set of versioned compoonents that deal with specific versions of the snapshot.

### How it works

Our snapshot not contains a version number, we use this to search a record (`componentMap`) for each snapshot component. It searches for the last supported version to display. 

Eg, the snapshot is on version 3, a component has a v1 and v2 support. If a candidate was snapshot at v1, it will use the v1 version of the component. If the canddiate is snapshotted at v3, it will fallback to v2.

The types for these components are fairly loose and specific to that version. This will hopefully avoid crashing the entire component if the snapshot is not exact and force more guards since we are elss confident in the data in the snapshot.

> [!NOTE]
> This is not very DRY but that is sort of on purpose. As the profile changes and breaking changes are created, these components will slowly drift. The latest version (onmly have 1 so far) should be very close tot he current profile and just be changed to not accept fragments since we cannot use them here. 

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Login as `admin@test.com`
3. View the information on different canadidates
4. Confirm the components display the approiate indformation


apps/web/src/components/Profile/components/LanguageProfile/Display.tsx
```diff
@@ -1,45 +1,11 @@
...
 import { Ul } from "@gc-digital-talent/ui";
 import { FragmentType, getFragment, graphql } from "@gc-digital-talent/graphql";
 
-import { getEvaluatedLanguageLevels } from "~/utils/userUtils";
 import FieldDisplay from "~/components/FieldDisplay/FieldDisplay";
-
-import { getExamValidityOptions, getLabels } from "./utils";
-
-// A workaround is required to show the deprecated bilingual evaluation in the profile snapshot.
-// These changes will be removed a year from now. (Remove snapshot workaround for the bilingual evaluation field #9905)
```
mnigh, line 14: Does this PR change anything related to #9905 and the deleted comment?

apps/web/src/components/ProfileSnapshot/utils.ts
```diff
@@ -0,0 +1,19 @@
...
+  version: number,
+): ComponentType<TProps> | null {
+  const availableVersions = Object.keys(components)
+    .map(Number)
+    .sort((a, b) => b - a); // descending
+
+  for (const v of availableVersions) {
+    if (v <= version) return components[v];
+  }
+  return null;
```
mnigh, line 18: Shouldn't it return `1` or is this because there are `profile_snapshot` entries without the `version` key at all?

mnigh (commented):
I have a few questions but I think it looks good.

apps/web/src/components/ProfileSnapshot/utils.ts
esizer: Hmm, good point. I agree, thisis confusing.

I was alaways pasing in 1 with `snapshot.version ?? 1` as the fallback but it was not super clear, as noted here, so I moved the `1` to the function args top holefully clear things. The `null` should make more sense basically being the case of the current version not existing in the map so we have no component to actually render. eg

```
const map = {
  3: Version3,
  4: Version3
}

const snapshot1 = {
  version: null
}

const snapshot2 = {
  version: 2
}

// Returns null becuase defualt `1` not found
const Comp1 = getSupportedVersionedComponent(snapshot1.version)

// Returns null becuase version 2 is not in the map and all fallbacks also do not exist
const Comp2 = getSupportedVersionedComponent(snapshot2.version)
```

mnigh (approved):

apps/web/src/components/Profile/components/LanguageProfile/Display.tsx
esizer: I'll have to see, the ticket is a little vauge and this one only deals with the frontend and not the backend (likle file downloads)

PR #14763, opened 2025-10-02
[Fix] Allow `-` in hashes for nginx long cache rule

### LINKED ISSUE #14717: 🐛 Cache lifetime short for some assets

Some of our assets with hashed names are still coming through with short cache lifetimes.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

It looks like they are all files with a hyphen in the hash.  Our regex is probably getting thrown by that.

### ✅ Acceptance Criteria

- [ ] All hashed assets get a long cache lifetime in prod.

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #14717 

## 👋 Introduction

Expands the location rule for our long (1 year) cache rule for assets.

## 🧪 Testing

> [!TIP]
>  This is disabled locally so you need to make some changes to test

1. open the local nginx config (`infrastructure/conf/nginx-conf-local/default`)
2. Uncomment the cache control header on L231
3. Reload nginx (i usually just do `make down; make up`)
4. Navigate to any page
5. Check the headers for files that have a `-` in their hash
6. Confirm it has a cache control for `31536000` (1 year)

## 📸 Screenshot

swappy-20251002_143405

### DIFF
--- a/infrastructure/conf/nginx-conf-deploy/default
+++ b/infrastructure/conf/nginx-conf-deploy/default
@@ -179,7 +179,7 @@ server {
     # serve other asset files
     location ~ "\.(png|ico|gif|jpg|jpeg|svg|css|js|pdf|map|doc|docx|webmanifest|webp|pptx)$" {
         # contenthashed asset files get cached for a year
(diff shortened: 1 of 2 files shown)


petertgiles (approved):
Thanks for the quick fix!

PR #14776, opened 2025-10-06
[Fix] Assessment plan form dialog not updating

### LINKED ISSUE #14756: 🐛 Editing a Screening Question in the Assessment Plan, causes the text box to clear the previous edit, unless the page is refreshed

**If you edit a screening question, the text box wipes out your last edit after you save it, unless the entire page is refreshed**
The original text saves with no issues. 
When attempting to edit or add to the text, whatever the last edit was made gets **replaced** with a new edit after saving, instead of adding to it. 
The possible workaround is to refresh the entire page after each edit. 
I created a test process on the UAT. https://uat-talentcloud.tbs-sct.gc.ca/en/admin/pools/ae701bc4-36d4-4719-b287-33d67c2fdf27/plan 

Expected behaviour - Edits should save without having to refresh the page. 

### 🕵️ Details

Tested and confirmed in production and UAT.
It is possible to work around the issue by refreshing the page after each edit and save. Inconvenient since after a refresh you have to search for a process again, open the assessment plan, and open the screening questions. 

### 📋 Steps to Reproduce

1. Go to the test poster. https://uat-talentcloud.tbs-sct.gc.ca/en/admin/pools/ae701bc4-36d4-4719-b287-33d67c2fdf27
2. Go to "Edit the assessment plan". 
3. Open screening questions.
4. Edit the text, click save.
5. Note that it seems to save the text in the assessment plan.
6. Open the edit field, note that the last edit is not there.
7. Try to add another edit (or simply save again) and watch the previous edit disappear.
8. Repeat the flow but refresh the page after the save. Note that the edits are saving properly.  

### 📸 Screenshot

Start here:

Image

Open edit mode:

Image

Add an edit (the word SECOND) :

Image

Seemingly saved in the assessment plan:

Image

Go back to the edit mode and note it (the word SECOND) is no longer there:

Image

Make another edit (the word THIRD) and watch your previous edit (the word SECOND)  disappear from the assessment plan after saving:

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- Screening questions can be saved without having to refresh the page

esizer:
🤖 Resolves #14756 

## 👋 Introduction

Fixes an issue where the assessment plan dialog was not updating without a page refresh.

## 🕵️ Details

Seems as though the `useForm` hook was only getting called when the form initiall enters the DOM. I beleive `react-hook-form` is pretty optimized to avoid re-renders so, even when `intialValeus` were updating, it avoided changing `defaultValues` since they are defaults, and not the live state. We had a reset but with no particular values so, it always reset to the defaults. 

This adds an effect to sync those states up whenever the `initialValues` change.

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Login as `admin@test.com`
3. Create a pool with atleast one skill
4. Go to the assessment plan
5. Add a screening question
6. Edit that screening question (without refreshing the page)
7. Go back to edit that screening question (again, without refreshing the page)
8. Confirm the new value from step 6 appears and not the one added in step 5

### DIFF
--- a/apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/AssessmentDetailsDialog.tsx
+++ b/apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/AssessmentDetailsDialog.tsx
@@ -175,6 +175,33 @@ interface InitialValues
   screeningQuestions?: ScreeningQuestion[];
 }
 
+const getDefaultValues = (initialValues: InitialValues): FormValues => {
+  const { screeningQuestions, ...restInitial } = initialValues;
+  if (screeningQuestions) {
+    screeningQuestions.sort((a, b) =>
+      (a.sortOrder ?? Number.MAX_SAFE_INTEGER) >
+      (b.sortOrder ?? Number.MAX_SAFE_INTEGER)
+        ? 1
+        : -1,
+    );
+  }
+
+  return {
+    ...restInitial,
+    screeningQuestionFieldArray: screeningQuestions?.map(
+      (initialScreeningQuestion) => ({
+        id: null, // filled by react-hook-form
+        screeningQuestion: {
+          id: initialScreeningQuestion.id,
+          sortOrder: initialScreeningQuestion.sortOrder,
+          en: initialScreeningQuestion.question?.en,
+          fr: initialScreeningQuestion.question?.fr,
+        },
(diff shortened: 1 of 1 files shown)


bella77-69 (approved):
LGTM! 😁

PR #14822, opened 2025-10-15
[Debt] Refactor pool completeness check

### LINKED ISSUE #14675: ♻️ `PoolIsCompleteValidator` should be an attribute and not a validator
### ♻️ Refactor

[Laravel's validator ](https://laravel.com/docs/12.x/validation)class is meant to be used for:

> Laravel provides several different approaches to validate your application's incoming data.

`PoolIsCompleteValidator` is not used for form validation.

### 🕵️ Details

_No response_

### 🙋‍♀️ Proposed Solution

For the computing "completeness" in getIsCompleteAttribute, use an [attribute](https://laravel.com/docs/12.x/eloquent-mutators#defining-an-accessor) like we have for `displayName`, for example.

### ✅ Acceptance Criteria

- [ ] `PoolIsCompleteValidator` removed and replaced by attribute that computes "pool completeness"

esizer:
🤖 Resolves #14675

## 👋 Introduction

This moves the pool completness attribute to use regular methods rather than a validator.

## 🕵️ Details

Validators are intended to validate incoming data to enture its integrity and safety before using or storing data. However, we had this one to check outgoing data which was a misuse of a validator and introduced unecessary overhead. We had extra rules that were not important to the completeness of the pool.

This is an attempt to refactor that so we can do mininmal checks and reduce the unecessary, albeit small, overhead fo doing the extra work and using a tool in the incorrect manner.

I hope this pattern makes sense and is a reasonable change.

## 🧪 Testing

1. Confirm the checks are accurate
2. Confirm the completness check remains intact and working as expected

### DIFF
--- a/api/app/Models/Pool.php
+++ b/api/app/Models/Pool.php
@@ -54,6 +54,7 @@ use Spatie\Activitylog\Traits\LogsActivity;
  * @property string $work_stream_id
  * @property ?string $area_of_selection
  * @property array $selection_limitations
+ * @property bool $is_remote
  * @property \Illuminate\Support\Carbon $created_at
  * @property ?\Illuminate\Support\Carbon $updated_at
  * @property ?\Illuminate\Support\Carbon $closing_date


brindasasi (approved):
I like this better! Works well ..verified the functionality remains the same

brindasasi:
oops my bad .. I just realized this check doesn't work properly when the pool's required fields are completed. Please fix it in a sep PR
Screenshot 2025-11-04 at 9 48 05 AM

PR #14845, opened 2025-10-15
[Debt] Enhance testing template

brindasasi:
## 👋 Introduction

This PR enhances the existing testing template to capture more details so that it can be developed without ambiguity and for better test case management.

**Test Type** can be 
  Manual , Unit Test, E2E 

Going forward we would create a test issue for release testing on UAT and so this should accommodate both.

**Scope/Component** 

Currently just a text box entry. But my plan is to group the similar and so we can organize the test cases.
Plan is to enhance this as a dropdown selection once we identified all components. 

## 🧪 Testing

Check it out here 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/testing-template/.github/ISSUE_TEMPLATE/testing.yml

### DIFF
--- a/.github/ISSUE_TEMPLATE/testing.yml
+++ b/.github/ISSUE_TEMPLATE/testing.yml
@@ -8,7 +8,7 @@ body:
     attributes:
       value: |
         Thanks for taking the time to fill out this testing issue!
- - type: input
+  - type: input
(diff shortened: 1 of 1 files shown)


mauli0110 (approved):
lgtm

PR #14858, opened 2025-10-17
[Refactor] Use department input type `DepartmentBelongsTo` always

### LINKED ISSUE #14097: ♻️ Schema, always use `DepartmentBelongsTo` for input
### ♻️ Debt/Refactor

At the schema level, it should be consistent for department inputs to function the same way.
Following the form 
` departmentId: DepartmentBelongsTo`

This is currently not the case in places like `DuplicatePoolInput.departmentId`

Image

Whats more, the field type varies

Image

The current three instances appear to be

- `PlaceCandidateInput`
- `WorkExperienceInput`
- `DuplicatePoolInput`

This applies to schema inputs and not general types

### 🙋‍♀️ Proposed Solution

Consistent usage of `DepartmentBelongsTo`
This will require additional work elsewhere as the input type change will break existing operations

### ✅ Acceptance Criteria

- [ ] Only use input type for department input
  - [ ] If, for whatever reason, we can't use DepartmentBelongsTo then create another generic/reusable input that has the archiving validation on it.
- [ ] Necessary work elsewhere to keep mutations working

vd1992:
🤖 Resolves #14097 

## 👋 Introduction

Updates the schema to use the input type always and ripple changes elsewhere

## 🧪 Testing

1. No places missed to update it
2. Affected operations still work (placing, duplication, work experience etc)

### DIFF
--- a/api/app/GraphQL/Validators/CreateUpdateWorkExperienceValidator.php
+++ b/api/app/GraphQL/Validators/CreateUpdateWorkExperienceValidator.php
@@ -165,12 +165,11 @@ final class CreateUpdateWorkExperienceValidator extends Validator
                 // This does not work without proper connect/disconnect in the schema
                 // Rule::exists('classifications', 'id'),
             ],
-            'workExperience.department' => [
+            'workExperience.department.connect' => [
                 Rule::requiredIf(
                     $this->arg('workExperience.employmentCategory') === EmploymentCategory::GOVERNMENT_OF_CANADA->name
                 ),
-                // This does not work without proper connect/disconnect in the schema
-                // Rule::exists('departments', 'id'),
+                Rule::exists('departments', 'id'),
             ],
             'workExperience.supervisedEmployeesNumber' => [
                 Rule::requiredIf(
--- a/api/app/Models/WorkExperience.php
+++ b/api/app/Models/WorkExperience.php
@@ -230,6 +230,7 @@ class WorkExperience extends Experience
     /**
      * Return the department model related to the experience
      */
+    /** @return BelongsTo<Department, $this> */
     public function department(): BelongsTo
     {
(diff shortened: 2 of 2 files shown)


api/graphql/schema.graphql
```diff
@@ -2603,7 +2603,9 @@ input WorkExperienceInput {
   cafForce: CafForce @rename(attribute: "caf_force")
   cafRank: CafRank @rename(attribute: "caf_rank")
   classificationId: String @rename(attribute: "classification_id")
-  departmentId: String @rename(attribute: "department_id")
+  departmentId: DepartmentBelongsTo
+    @pluck(key: "connect")
+    @rename(attribute: "department_id")
```
vd1992: For some reason this didn't work otherwise, don't know why this needed unique directives appended

api/app/GraphQL/Validators/CreateUpdateWorkExperienceValidator.php
```diff
@@ -165,7 +165,7 @@ public function rules(): array
                 // This does not work without proper connect/disconnect in the schema
                 // Rule::exists('classifications', 'id'),
             ],
-            'workExperience.departmentId' => [
+            'workExperience.department' => [
```
petertgiles: Should this be `workExperience.department.connect`?

api/app/Models/WorkExperience.php
```diff
@@ -230,7 +230,7 @@ public function classification()
     /**
      * Return the department model related to the experience
      */
-    public function department()
+    public function department(): BelongsTo
```
petertgiles: Could you please also add the type hint comment like `user()` has to help out Larastan?

petertgiles (approved):

PR #14868, opened 2025-10-20
[Fix] Allow `BoolCheckIcon` label to be read out by screen readers

### LINKED ISSUE #14857: 🐛 Bool check icon has an accessible hidden label

The BoolCheckIcon has props to add accessible labels to the icon.  However, the icon is aria-hidden so the labels are never presented.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Log in as applicant-employee@test.com
2. Browser to applicant dashboard -> career development -> functional communities -> view interest dialog
3. Observe the icons in the work stream list

### 📸 Screenshot

Image

Image

### 🙋‍♀️ Proposed Solution

https://www.a11y-collective.com/blog/svg-accessibility/#:~:text=other%20HTML%20content.-,Here%E2%80%99s%20an%20example%3A,-%3Cp%20class

### ✅ Acceptance Criteria

- [ ] Accessible labels in BoolCheckIcon are presented in the a11y tree

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #14857 

## 👋 Introduction

Fixes an issue where screen readers were having difficulty reading out the label for the `BoolCheckIcon`

## 🕵️ Details

All of our icons have `aria-hidden` by defualt, this is be cause for the most part, they are decorative. However, the `BoolCheckIcon` needs to convey information to the user. This does more than just set `aria-hidden="false"` since our current approach has a gap in support for firefox + JAWs (using `aria-label` on svgs).

The fix with the least amount of gaps is moving the label to a real element and associcating it with an image via `aria-labelledby`.

Why `hidden` instead of `className="sr-only`"? With `sr-only`, it creates a redundant announcment since `sr-only` only visually hides the element but leaves it in the acccessibility tree. So, but the element and the icon announce the message. By using `hidden` we remove the element from the accessibility tree, only exposing it via its relation to the svg with `aria-labelledby`.

## 🧪 Testing

> [!IMPORTANT]
> You should probably test Firefox since it has the most gaps in this specific scenario. 

1. Build `pnpm dev:fresh`
2. Navigate to a page that includes the `BoolCheckIcon` (employee profile has some)
3. Open a screen reader
4. Confirm the label is read out
    - E.g. "Not interested, image, {text}"

### DIFF
--- a/apps/web/src/components/BoolCheckIcon/BoolCheckIcon.tsx
+++ b/apps/web/src/components/BoolCheckIcon/BoolCheckIcon.tsx
@@ -1,6 +1,6 @@
 import XCircleIcon from "@heroicons/react/20/solid/XCircleIcon";
 import CheckCircleIcon from "@heroicons/react/20/solid/CheckCircleIcon";
-import { ReactNode } from "react";
+import { ReactNode, useId } from "react";
 import { tv } from "tailwind-variants";
 
 import { Maybe } from "@gc-digital-talent/graphql";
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
Sweet solution!  Looks good in FF.

Image

Unfortunately, no JAWS or NVDA on Linux.

PR #14872, opened 2025-10-20
[Spike] Router middleware

### LINKED ISSUE #14640: 🦔 Investigate `react-router` middleware
### 🦔 Spike

[Middleware](https://reactrouter.com/how-to/middleware) has recently went stable in `react-router`. We have a multitude of use cases that we think could be implemented using this new feature so it seems like investigating it would be a boon. Some of the use cases (that do not entirely need to be answered here) are:

 - Loading data so we can move fully to the [route modules](https://reactrouter.com/start/framework/route-module) and "framework" mode
 - Holding auth context leading to:
    - Removing the need to wrap every component in `RequireAuth` (it could instead use middleware)
    - Adding team related data maybe?
 - Being able to redirect users before a page loads for a multitude of reasons

### ❓ Questions to answer

 - Can we access `urql` client and use it in middleware? (this is crucial to bea ble to query data like the current user)
 - Can we access `intl` data in middleware to provide messages? (this is needed to provide the user language to the client)

### 🕙 Timebox

My guess is 2 days?

esizer:
🤖 Resolves #14640 

## 👋 Introduction

This is a proof of concept for `react-router` middleware.

## 🕵️ Details

This is an attempt to prove that we could begin using the middleware feature for `react-router` which should facilitate the move to loaders and actions (if we still want that). This should, hopefully, simplify a lot of process in our app.

Overall, I think this went well and could simplify things a bit and helps in co-locating code related to a specific route. We could almost entirely remove the following:

- `RequrieAuth.tsx` - replaced by the protected route middleware (not a huge deal)
- `Router.tsx` - Replaced by route config, same thing but easier to manage imo
- Some of `ContextProvider.tsx` Alot of our context providers can be moved to middleware

Also, eventually, we could move away from the API wrapper (`Pending.tsx`) pattern since this would be contained in loaders.

### Framework mode

> [!NOTE]
> This is optional but should drastically improve the migration and make it much nicer. We could do these in any order. However, I personally think we are closer to framework mode and it would save us time adding extra code for middleware (types) just to remove it later. 

Right now, I have a lot of types added here which seems cumbersome. However, if we take an important step prior to implementing real middleware, this should be much smoother.

That step would be to migrate from `RouteObject` to `RouteModule`. This would mean adopting the framework mode of `react-router`.

REF: https://reactrouter.com/upgrading/router-provider

This would simplify how our routes are defined. You can see, we are already close to supporting this and I already took the first step and wrote a simple "convert" function that takes out route objects and converts them into a framework friendly `RouteModule`.

This would give use two important peices to make this smoother.

1. Not needing to declare middleware in the `Router.tsx`
2. [Type safety](https://reactrouter.com/explanation/type-safety#type-safety)

The `LazyRouteObject` does not support middleware, so we currecntyl need to define it in the `Router`. If we move to framework mode, we could simply export `clientMiddleware` and it would work out of the box.

Also, type safety is only supported in framework mode. This is because framework mode requires a vite plugin that can infer and generate types (sort of like codegen). This would reduce the amount of manual types I added here.

### Authorisation

Since middleware runs before the loader and react is rendered, just throwing errors in middleware (like unauthorised) requires us to do that check in a loader which is the earliest we can show a nice error message.

We could do a factory (`makeProectedRouteMiddleware([ROLE_NAME.PlatformAdmin])`) but, currently it returns the opt level error (if we just throw the error) which is not nice and we do not have a nice page for it and does not include the layout.

This could take some work but, I think the simplest solution is to juse add it to the loader :woman_shrugging: 

### No more hooks

Since middleware runs outside of the react component lifecycle, we do not have access to hooks. You will see, that is not really an issue anmd the important parts should be woprking.

However, I took some liberties in migrating and hardcoded some stuff to avoid some hooks. Though, if you look deeper at the hooks I avoided (`useAuthenication`, `useApiRoutes`, `useLogger`), almost none of them need to be hooks. I simply avoided them to only touch necessary stuff.

I think this is easily done, it would just require small touches to these "hooks", essentiually removed the `use` prefix and prefer something like `get`.

### Complex code migration

This would mean making small refactor some very important parts of our app (`auth`, `i18n`). This would be non-trivial and likely require a lot of testing. 

Though, in my testing, it seems as though this could be done incrementally :thinking: We don't need the client for all pages,  we don't need `intl` for all pages. We could move pages over in chunks or on an "as needed" basis.

### Still no top-level error boundary

When testing the authorisation, you will notice the error message is really bad. This is because we lack a top level error boundary. 

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Navigate to any page
3. Observe the console log for `intl` related messages coming from middleware
4. Without logging in, navigate to `/applicant`
5. Observe you are still sent to login even without `RequireAuth`
6. Login as `applicant@test.com`
7. Confirm you can see the dashboard and see the data queried in the loader in the console
8. Update the `guardByRoles` call to require platform admin instead of applicant
9. Rebuild app `pnpm dev:fresh`
10. Reload the dashboard
11. Confirm you get the unauthorized error

### DIFF
--- a/api/app/Builders/PoolCandidateBuilder.php
+++ b/api/app/Builders/PoolCandidateBuilder.php
@@ -518,7 +518,9 @@ class PoolCandidateBuilder extends Builder
             return $this;
         }
 
-        return $this->whereIn('assessment_step', $sortOrder);
+        return $this->whereHas('assessmentStep', function (Builder $query) use ($sortOrder) {
+            $query->whereIn('sort_order', $sortOrder);
+        });
     }
 
     public function orderByClaimVerification(?array $args): self
--- a/api/app/Models/CommunityInterest.php
+++ b/api/app/Models/CommunityInterest.php
@@ -36,6 +36,8 @@ class CommunityInterest extends Model
 
     protected $keyType = 'string';
 
+    protected $touches = ['user'];
+
     /**
      * The attributes that should be cast.
      */
--- a/api/app/Models/Experience.php
+++ b/api/app/Models/Experience.php
@@ -26,6 +26,8 @@ abstract class Experience extends Model
 
     protected $keyType = 'string';
 
+    protected $touches = ['user'];
+
     protected static $hydrationFields;
 
     abstract public function getTitle(): string;
--- a/api/app/Models/UserSkill.php
+++ b/api/app/Models/UserSkill.php
@@ -33,6 +33,8 @@ class UserSkill extends Model
 
     protected $keyType = 'string';
 
+    protected $touches = ['user'];
+
     protected $casts = [
         'top_skills_rank' => 'integer',
         'improve_skills_rank' => 'integer',
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -41,6 +41,8 @@ type UserAuthInfo {
   roleAssignments: [RoleAssignment] @hasMany
   createdDate: DateTime @rename(attribute: "created_at")
   deletedDate: DateTime @rename(attribute: "deleted_at")
+  updatedDate: DateTime @rename(attribute: "updated_at")
+  lastSignInDate: DateTime @rename(attribute: "last_sign_in_at")
 }
 
 type User {
--- a/api/storage/app/lighthouse-schema.graphql
+++ b/api/storage/app/lighthouse-schema.graphql
@@ -708,6 +708,8 @@ type UserAuthInfo {
   roleAssignments: [RoleAssignment]
   createdDate: DateTime
   deletedDate: DateTime
+  updatedDate: DateTime
+  lastSignInDate: DateTime
 }
 
 type User {
--- a/api/tests/Feature/Generators/UserDocGeneratorTest.php
+++ b/api/tests/Feature/Generators/UserDocGeneratorTest.php
@@ -51,6 +51,8 @@ class UserDocGeneratorTest extends TestCase
             ->withCommunityInterests([$community->id])
             ->create();
 
+        $targetUser->loadMissing(['userSkills']);
+
         // Faker seed makes skill ranks the same.
         // This is not realistic data so we are forcing them
         // to be different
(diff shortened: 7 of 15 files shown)


apps/web/src/pages/ApplicantDashboardPage/ApplicantDashboardPage.tsx
```diff
@@ -474,9 +519,6 @@ export const ApplicantDashboardPageApi = () => {
   );
 };
 
-export const Component = () => (
-  <RequireAuth roles={[ROLE_NAME.Applicant]}>
-    <ApplicantDashboardPageApi />
-  </RequireAuth>
-);
+export default Component;
+
```
esizer: No more need for `RequireAuth`, we can handle all that in the loader now (see: `guardByRoles`)

apps/web/src/utils/routing.ts
esizer: This is simple renaming as a first step to framework mode (not really part of middleware but makes things easier).

SEE: https://reactrouter.com/upgrading/router-provider#1-move-route-definitions-into-route-modules

packages/client/src/utils/getClient.ts
petertgiles: Yikes, refactors on the auth system make me nervous.

apps/web/src/middleware/protectedRouteMiddleware.ts
```diff
@@ -0,0 +1,19 @@
+import { MiddlewareFunction, redirect } from "react-router";
+
+import { userContext } from "./authMiddleware";
+import { intlContext } from "./intlMiddleware";
+
+const protectedRouteMiddleware: MiddlewareFunction = ({ context }) => {
```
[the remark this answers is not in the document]
esizer: Yeah, the best thing about this is, all of this code runs before anything is rendered at all. We can provide a `HydrateFallback` (a loading spinner usually) that is rendered while this code runs.

REF: https://reactrouter.com/start/framework/route-module#hydratefallback

packages/client/src/utils/getClient.ts
esizer: Yeah, that is absolutley my largest concern with this :sweat:

apps/web/src/middleware/authMiddleware.ts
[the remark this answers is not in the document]
petertgiles: Yeah for sure.  I can imagine it already.  :laughing:

PR #14876, opened 2025-10-21
[Fix] Prevent redirect to logged out page on auth refresh error

### LINKED ISSUE #14536: 🐛 Links to public pages redirect to "logged out" page.

If a user has signed up for an account in the past and then let their session lapse, links to public pages will no longer work.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Applicant has created an account in the past and signed up for email notifications.
2. They didn't log out, just closed the browser.
3. They receive an email for a new pool with a link to it.
4. They click the link to the new pool
5. They arrive at the "You have been signed out page".
6. Where is the new pool?
### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

- Switch "you have been signed out" to a toast?

### ✅ Acceptance Criteria

- [ ] Clicking a link to a public page should go there without redirecting to "you have been signed out"
- [ ] Clicking a link to a private page should redirect to sign in and then redirect again to the private page

esizer:
🤖 Resolves #14536 

## 👋 Introduction

This should address issues where users have an expired session and try to navigate to any page.

## 🕵️ Details

> [!CAUTION]
> I'm not sure this is the best/correct solution but it does seem to address the issue. 

The idea here is simple. When we have a token refresh error, the logout occurs and, instead of redirecting to `/logged-out`, we send them to the same page they were trying to get to. This then allows the `RequireAuth` component to render and handle the guard.

So, if you are trying to get to a protected route, it should send you to the login flow that returns you to the protected page assuming the login was a successs. If you go to a public page, it simply logs you out and reloads the page to clear out any state (like being authenticated).

## 🧪 Testing

> [!NOTE]
> I assume the setup makes sense to emulate the issue. I hope I am correct in this :sweat_smile:  

1. Optional: Edit `mock-oauth2-server.json` to set the `tokenExpiry` to something short (so you don't need to sit around for 10 mins)
2. Edit `AuthController::refresh` to return some sort of error
3. Rebuild everything `make down; make up; pnpm dev:fresh`
4. Login as any user
5. Wait for the token to expire (10mins or what you set in step 1)
6. Navigate to a public page
7. Obseve the browser kinda reset (logout, refresh) but ultimatley land on the page you were trying to get to
8. Repeat steps 4+5
9. Navigate to a protected page
10. Observe you are redirected to login
11. Complete the login and confirm you end up on the expected page

### DIFF
--- a/apps/web/src/utils/utils.ts
+++ b/apps/web/src/utils/utils.ts
@@ -1,3 +1,9 @@
+/**
+ * Checks if the provided URL matches the current app's host name
+ *
+ * @param {string} url - The URL to test
+ * @returns {boolean} True if the URL matches the current host name
+ */
 export function urlMatchesAppHostName(url: string): boolean {
   const pattern = new URLPattern({ hostname: window.location.hostname });
   return pattern.test(url);


apps/web/src/components/Router.tsx
```diff
@@ -185,7 +185,14 @@ const createRoute = (locale: Locales, featureFlags: FeatureFlags) =>
             },
             {
               path: "logged-out",
-              loader: () => {
+              loader: ({ request }) => {
+                const url = new URL(request.url);
+                const from = url.searchParams.get("from");
+                if (from) {
+                  // eslint-disable-next-line @typescript-eslint/only-throw-error
+                  throw redirect(from);
```
petertgiles: I think we need to do some validation here.  Otherwise we've got an [open redirect vulnerability](https://learn.snyk.io/lesson/open-redirect/?ecosystem=javascript).  I could sneakily send a user anywhere, like http://localhost:8000/en/logged-out?from=https%3A%2F%2Fbing.com

[In the authcallback](https://github.com/GCTC-NTGC/gc-digital-talent/blob/0fe07853292afa8298184d86cb2fbd970e989a0c/api/app/Http/Controllers/AuthController.php#L145) we only allow relative URLs but I'm not sure that will work here.

apps/web/src/components/Router.tsx
```diff
@@ -185,7 +185,14 @@ const createRoute = (locale: Locales, featureFlags: FeatureFlags) =>
...
+                }
```
petertgiles: On line 202 we do a `window.location.href` set.  Is there a reason why you did a `redirect` here?

esizer: `redirect` sets the status code to `302` in the browser history which seemed like what we expect here. Though, if we do not want that behaviour, I can change it to the other method.

apps/web/src/components/Router.tsx
petertgiles: But this will never work now, right?  Because it's set from the full URL?
` searchParams.append("from", window.location.href);`

petertgiles: Even if we fix that, we want to be able to handle links in emails and things.  Instead of the very simple "does it start with a slash" logic, can we do something like check the host name or origin or something?

esizer: okay, this is not complete (still need to update the pipeline YAML) but what do you think fo this?

[add test for full URL](https://github.com/GCTC-NTGC/gc-digital-talent/pull/14876/commits/a4a2926ac1f223ee9ece04a05865575c5ae8e5b1)

It should work like:

 - `/en` -> true (starts with `/`)
 - `talent.canada.ca` -> false (no protocol)
 - `http://github.com` -> false (does not match our hostname)
 - `https://talent.canada.ca` -> true (matches our hostname and has protocol)

petertgiles: Hmm, I think this is on the right track but I don't think it will quite work.  If we use a build-time variable for this then the same artifact can't be used in dev, uat, and prod.  I don't love having to add new env variables, but if we wanted to do this it would have to be a deploy-time variable in index.html.  What if we just used a allow list that included dev and uat origins?   Or can we just check what the current location is and verify the it's either relative or to the same origin?

esizer: That makes a lot of sense. Do you think this will work?

[check against current hostname instead of env var](https://github.com/GCTC-NTGC/gc-digital-talent/pull/14876/commits/c8dbc32083d464c02b800c72c86b024098369d84)

apps/web/src/utils/utils.ts
petertgiles: Could you add a small comment on what this function does?

PR #14886, opened 2025-10-22
[Tests] Migrate to `vitest`

### LINKED ISSUE #12600: 🧪 Replace jest with vitest
## 🧪 Tests

The aim of this is to replace jest with vitest to hopefully simplify dependencies and config as well as speed up tests.

## 🕵️ Details

While jest is an amazing test runner, our setup is quite complex and somethings have held us back from upgrading specific dependencies. Vitest is created work alongside our existing vite config which should simplify configuration.

REF: https://vitest.dev/guide/comparisons.html#jest

## 🙋‍♀️ Proposed Solution

Follow the [migration guide](https://vitest.dev/guide/migration.html#jest).

## ✅ Acceptance Criteria

- [ ] Vitest installed and configured
- [ ] Jest and its related dependencies removed
- [ ] Same test coverage
- [ ] All tests passing

esizer:
🤖 Resolves #12600 

## 👋 Introduction

Replaces `jest` with `vitest`

## 🧪 Testing

1. Confirm the github workflow functions
2. Confirm the tests still pass/fail as expected

### DIFF
--- a/.storybook/preview.ts
+++ b/.storybook/preview.ts
@@ -9,7 +9,7 @@ import {
   VIEWPORTS,
 } from "@gc-digital-talent/storybook-helpers";
 import { ToastDecorator } from "@gc-digital-talent/toast";
-import { richTextElements as defaultRichTextElements } from "@gc-digital-talent/i18n";
+import defaultRichTextElements from "@gc-digital-talent/rich-text-elements";
 
 import frCommonCompiled from "@gc-digital-talent/i18n/frCompiled.json";
 import frCompiled from "../apps/web/src/lang/frCompiled.json";
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
   },
   "devDependencies": {
     "@gc-digital-talent/i18n": "workspace:*",
+    "@gc-digital-talent/rich-text-elements": "workspace:*",
     "@gc-digital-talent/storybook-helpers": "workspace:*",
     "@gc-digital-talent/toast": "workspace:*",
     "@gc-digital-talent/vitest-helpers": "workspace:*",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -11,6 +11,9 @@ importers:
       '@gc-digital-talent/i18n':
         specifier: workspace:*
         version: link:packages/i18n
+      '@gc-digital-talent/rich-text-elements':
+        specifier: workspace:*
+        version: link:packages/rich-text-elements
       '@gc-digital-talent/storybook-helpers':
         specifier: workspace:*
         version: link:packages/storybook-helpers


apps/web/src/utils/experienceUtils.test.ts
```diff
@@ -1,8 +1,9 @@
 /**
- * @jest-environment jsdom
+ * @vi-environment jsdom
```
petertgiles: All the other tests remove this comment.  This looks like a find-and-replace mistake?

apps/web/package.json
```diff
@@ -24,8 +24,9 @@
     "production": "pnpm run build",
     "build-storybook": "storybook build",
     "storybook": "storybook dev -p 6006",
-    "test": "jest --detectOpenHandles --forceExit",
-    "test:watch": "jest --watch",
+    "test": "vitest run",
```
petertgiles: Yes!  No more debug mode!

apps/web/package.json
```diff
@@ -24,8 +24,9 @@
...
+    "test:coverage": "vitest run --coverage",
+    "test:watch": "vitest",
```
petertgiles: Missing a watch flag?

packages/vitest-helpers/src/components/Providers.tsx
```diff
@@ -2,8 +2,26 @@ import { HelmetProvider } from "react-helmet-async";
 import { IntlProvider } from "react-intl";
 import { BrowserRouter } from "react-router";
 import { ReactNode } from "react";
+import { vi } from "vitest";
 
-import { richTextElements } from "@gc-digital-talent/i18n";
+const richTextElements = {
```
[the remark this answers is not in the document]
petertgiles: For most of them, it shouldn't matter, right?  It won't have CSS anyway.  The `hidden` one, though, I can imagine might mess up tests if it doesn't actually apply sr-only.  My second concern is, how will we remember to update this when we add new functions?

That's tricky about circular references.  I agree, a new package for this is a bit silly.  What about packages/helpers/utils?  packages/i18n already has a dependency on helpers.  And these are functions are not strictly about translation anyway.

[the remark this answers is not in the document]
esizer: Hmm, I have been looking and I think I might need to backtrack on merging the packages. We used to have `jest-helpers` and `jest-presets` to avoid the circular dependencies. I tried to solve that issue and merge them but unfortutley, it might not be possible unless we want to inject the rich tesct elements every time we need to use them :sweat:

apps/web/src/utils/experienceUtils.test.ts
esizer: Oh good catch! Removed in [remove comment](https://github.com/GCTC-NTGC/gc-digital-talent/pull/14886/commits/5f0daf761be94c1cb4fe03d535b99734344ca0fe)

vd1992:
Trying it out... In WSL

**Main**
Run at project root, crash, then 60 seconds
Running in just `apps/web` 50 seconds

**Branch**
Run at project root, crash, then crash again
Running in just `apps/web` 25 seconds

It works, WSL pain aside, when it runs it runs and halved the time

PR #14911, opened 2025-10-24
[Feat] Control notification polling interval from env var

esizer:
## 👋 Introduction

We are having problems with the norificatin count running frequelty causing users to logout after 1min from an intermittedn SiC issue. This should atleast alleviate that while we get a more permanent fix.

## 🧪 Testing

1. Add new env var to your `./apps/web/.env`
2. Set it to something low for testing
3. Build `pnpm dev:fresh`
4. Login as any user
5. Observe the polling happens at the interval set

## 🚚 Deployment

Add the `NOTIFICATION_POLLING_INTERVAL` variable to infra, setting as the interval in minutes.

### DIFF
--- a/apps/web/index.html
+++ b/apps/web/index.html
@@ -1,90 +1,130 @@
 <!doctype html>
 <html lang="en">
+  <head>
+    <meta charset="UTF-8" />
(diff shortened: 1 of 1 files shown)


brindasasi (approved):
Working fine to me! 👍

PR #14954, opened 2025-10-29
[Fix] `DropdownMenu` z-index and height

### LINKED ISSUE #14935: 🐛 Pool Candidates "Filter by" dropdown position

When using the "Filter by" dropdown in the `pool-candidates` table, the dropdown menu expands above the filter button on some screen sizes. The dropdown should open below the filter button.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Login as `admin@test.com`
2. Navigate to the `/pool-candidates` page
3. Click on the "Filter by" button and observe the dropdown position
4. Open dev tools and resize the view and observe the dropdown position in different screen sizes

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] menu is layered above the nav bar
- [ ] menu does not extend beyond the viewport

esizer:
🤖 Resolves #14935 

## 👋 Introduction

Fixes the `DropdownMenu` from showing below sticky nav and extending outside the bounds of the screen.

## 🕵️ Details

Pretty frsutrating that we actually do not have direct control over `z-index`. This lead to needing to add custom css with `!important` :-1: 

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Login as `admin@test.com`
4. Shrink height of screen to something snmaller than the content of the filter dropdown. Make sure that there is more room above the trigger than below
5. Open the dropdown
6. Confirm that it onyl extends to the top of the screen and shows up over top of the sticky nav

## 📸 Screenshot

swappy-20251029_135440

### DIFF
--- a/apps/web/src/assets/css/tailwind.css
+++ b/apps/web/src/assets/css/tailwind.css
@@ -273,3 +273,11 @@
 abbr:hover {
   cursor: help;
 }
+
+/*
+* Cannot directly style `DropdownMenu` content wrapper
+* Important required because `z-index: auto` set inline
+*/
+[data-radix-popper-content-wrapper] {
+  z-index: 50 !important;
+}


packages/ui/src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -17,15 +17,15 @@ const Trigger = forwardRef<
...
-  base: "max-w-screen rounded-md bg-white p-3 font-sans text-black shadow-md dark:bg-gray-600 dark:text-white",
+  base: "max-h-(--radix-dropdown-menu-content-available-height) max-w-screen overflow-y-auto rounded-md bg-white p-3 font-sans text-black shadow-md dark:bg-gray-600 dark:text-white",
 });
 
 const StyledContent = forwardRef<
   ComponentRef<typeof DropdownMenuPrimitive.Content>,
   ComponentPropsWithoutRef<typeof DropdownMenuPrimitive.Content>
 >(({ className, ...rest }, forwardedRef) => (
   <DropdownMenuPrimitive.Content
-    className={content({ class: ["z-10", className] })}
```
bella77-69: Learning question: Why couldn't we just change z-10 to z-50?? instead of using CSS? Is radix doing something behind the scenes that would not let that work?

esizer: So, this is actually a child of a component we do not control. That component has css attributes that make setting `z-index` here do effecivley nothing.

```html
<div style="position: fixed; z-index: auto;" />
```

The combination of these attributes make it so this controls all child element `z-index` keeping them as `auto`. The `z-index` set in child elements only apply to other siblings and not the overall document.

`z-index` can only go as far as the first ancestor element with a position of anything other than `static`.

Since we do not have contorl over this component, we could not change that `z-index: auto` to something like `z-index: 50;` (or use `className`). Also, since the `z-index` set is using the `style` attribute, it has the highest priority. 

Both of those points is why we had to set this in css and target it using the data attribute along with the `!important` specifier.

Not really happy with this but unfortunatley it was the only way we could get it to work with how our DOM is :-1: 

Hopefully that long rant made sense :sweat_smile:

PR #14996, opened 2025-11-03
[Fix] Missing or incorrect filter maps for document download

### LINKED ISSUE #14791: 🐛 Download dataset inconsistent across tables

Does not work the same way for the users and pool candidates table

### 🕵️ Details

Spun off from the discussion here
https://github.com/GCTC-NTGC/gc-digital-talent/pull/14683#discussion_r2413754829

### 📋 Steps to Reproduce

Go to users table observe users count, then set some filters, observe users count drop.
Download dataset, open file, observe row count returns more than the filtered table count. Approximately all users

Then

Go to all candidates table, observe candidates count, then set some filters, observe candidates count drop.
Download dataset, open file, observe row count (plus header) is equal to the filtered table count. _**Not**_ all candidates

The problem then is, which way should the download dataset work? The one that works with filtering or the one that ignores filtering. 

### 📸 Screenshot

Users table

Image

Image

.
Candidates table (23 is the row count _**including**_ the headings row)
.

Image

Image

### ✅ Acceptance Criteria

- [ ] fix the filter input map
- [ ] post an issue to try to come up with a solution that doesn't require the filter input map to be maintained with future changes.

esizer:
🤖 Resolves #14791 

## 👋 Introduction

Fixes an issue where the filter maps in document generators were not updated to make new or renamed scopes from graphql.

## 🕵️ Details

Since we do the filtering in a job, we don't have `lighthouse-php` applying scopes for us. This means, we need to do all the renaming ourselves in the form of maps. Unfortunaly, this is quitebrittle and we consistnetly forgot to do so. This addresses that.

However, a better solution would be to avoid renaming any scopes in graphq. I.e

```graphql
input CurrentFilterInput {
  name: String @scope(name: "whereName")
  email: String @scope(email: "whereEmail")
}

input ProposedFilterInput {
  whereName: String @scope
  whereEmail: String @scope
}
```

## 🧪 Testing

1. Login as `admin@test.com`
2. Navigate to the tables that are affected by the changes
3. Apply the filters that were missing
4. Download a document
5. Confirm the document contains the data that was represented in the table when the filters were applied

### DIFF
--- a/api/app/Generators/UserExcelGenerator.php
+++ b/api/app/Generators/UserExcelGenerator.php
@@ -322,7 +322,6 @@ class UserExcelGenerator extends ExcelGenerator implements FileGeneratorInterfac
             'name' => 'whereName',
             'generalSearch' => 'whereGeneralSearch',
             'roles' => 'whereRoleIn',
-            'skills' => 'skillsAdditive',
 
             // Applicant filter input renames
             'equity' => 'whereEquityIn',


esizer:
> Nice adding the filters in. However, I'm still noticing the filters not being applied to the Users table when downloading full dataset 😕.

Added some more missing filters to the map [add missing applicant filter scopes](https://github.com/GCTC-NTGC/gc-digital-talent/pull/14996/commits/0821de679b0ebca300747e727eceba8f157cb4e1)

Issue #14558, opened 2025-09-08
✨ Improve general search


petertgiles:
### ✨ Feature Request

The goal of this issue is to implement the suggestions that came out of a previous spike: 
https://github.com/GCTC-NTGC/gc-digital-talent/issues/13365

### ✅ Acceptance Criteria
- [ ] As much as possible, new search released in parallel with existing search to reduce disruption
- [ ]     Switch away from websearch_to_tsquery to plain to_tsquery. We're nitty-gritty building the query already so might as well not fight the library.
- [ ]     Start prefix-matching each term. This is enabled because of to_tsquery. This will allow us to partial match searches with the query.
- [ ]     Start using the result rank to order results. Right now we're ignoring the result rank which prevents better matches from bubbling to the top.
- [ ]     Use result rank as the default ordering instead of created_at. This is more useful by default. And if an explicit ordering is requested in the table then use result rank as the secondary ordering to help order results more usefully when the first ordering is low-cardinality.
- [ ]     Remove the ilike matching for first_name, last_name and email. With prefix matching and weighting it should be unnecessary. (👍 better ranking results 👎 only prefix matching, not middle or end) If a user specifically is looking in those fields they can use the other search methods.
- [ ] Menu item label: "New beta search" (bottom of the list)
- [ ] unit testing

petertgiles:
Reviewed in estimation:
- mixed response, no consensus

petertgiles:
Reviewed in refinement:
- Brinda and Gray: must move forward with search - very important
- Moving weighting out of A/C to reduce scope

PR #14998, opened 2025-11-03
[Refactor] Move search scope to model builder

### LINKED ISSUE #14585: ♻️ Move scopeWhereGeneralSearch to UserBuilder
### ♻️ Refactor

Most of our user scopes are in UserBuilder.  But not scopeWhereGeneralSearch.

### 🕵️ Details

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] scopeWhereGeneralSearch is in UserBuilder

petertgiles:
🤖 Resolves #14585 

## 👋 Introduction

Moves the general search scope to the user model builder.

## 🕵️ Details

I'm not totally sure why, but it seems like when the scope is moved from the model itself, Laravel stops wrapping the additions to the query in parenthesis.  Since this scope adds two clauses with an OR, that's a big problem.  See how the query changes when the parenthesis are dropped:
image
I just added a `$this->where(function($query) { ... }` to wrap the two clauses in the scope together.  You can see that change in the [wrap the OR clauses](https://github.com/GCTC-NTGC/gc-digital-talent/commit/83d7c310d493bf68e9bd261a8126f295771d0d1b) commit.

Unit testing for the win!  I totally would have missed this.  The test itself was way too long so I refactored it with a data provider as a bonus.

For what it's worth, the second clause with its `ilike`s and wrapper disappear in [#14558](https://github.com/GCTC-NTGC/gc-digital-talent/issues/14558) .  The scope drops to half its current size.

## 🧪 Testing

> [!IMPORTANT]
>  Quite a few places call the user general search other than the user table.

1. User table general search still works.
2. Pool candidate table general search still works.
3. Pool candidate general search + CSV download still works.
4. Community talent table general search still works.
5. WFA table general search still works.  (It doesn't work on main right now.  I included a fix in this PR.)
6. Tests complete.

### DIFF
--- a/api/app/Builders/UserBuilder.php
+++ b/api/app/Builders/UserBuilder.php
@@ -684,7 +684,7 @@ class UserBuilder extends Builder
                 });
 
                 if ($user->isAbleTo('view-own-employeeWFA')) {
-                    $query->orWhere('id', $user->id);
+                    $query->orWhere('users.id', $user->id);
                 }
             }
         });


esizer (approved):
Cool! 😎 Love the simplifcation of the tests 😍

PR #14999, opened 2025-11-03
[Feature] Null current step when removed or final decision

### LINKED ISSUE #14983: ✨ Change "Current step" to NULL for all Final decision statuses except "To assess"
### Details

Feedback following a demo from @brindasasi with recruitment team:

"It's very confusing (and tiresome) to have to select `To assess` and then `Current step` when trying to find candidates that are in a specific step. Everytime we select the step, it should only give us the candidates that are in the `To assess`, since all `Qualified`, `Disqualified `or `Removed`, should no longer BE in any step anymore"

### Design file & designer

_No response_

### 📸 Screenshots

Image

### 🌎 Localization

_No response_

### Proposed Implementation

_No response_

### ✅ Acceptance Criteria

- When a candidate is `Qualified`, `Disqualified `or `Removed`, the `Current step` should change to : `null`
- Migrate all 'Current step' to null for all candidates that are `Qualified`, `Disqualified `or `Removed`

vd1992:
🤖 Resolves #14983

## 👋 Introduction

Using `removed_at` and `final_decision_at` as equalling `Removed`, `Qualified`, or `Disqualified` 

Migration to null out `pool_candidate.assessment_step_id`
Have RoD mutations act upon the current step too. 
Added some testing to check updating current step when applying the mutations mentioned above

## 🧪 Testing

1. Test migration
2. Have a candidate, exercise the RoD mutations, check their current step

## 📸 Screenshot

image

image


api/app/Models/PoolCandidate.php
```diff
@@ -541,6 +541,11 @@ public function computeAssessmentStatus()
             }
         }
 
+        // override to set current step as NULL if removed_at OR final_decision_at are set
+        if (! is_null($this->removed_at) || ! is_null($this->final_decision_at)) {
+            $currentStep = null;
+        }
```
vd1992: I felt it was best to continue to allow this method/location to be the source of truth for current step and just refer to it from more places. The RoD mutations are reversible so need to keep both directions in mind

vd1992:
I think that is pre-existing and not a side effect from this PR.
To replicate on main

1) Have candidate set as qualified available
2) Manually set their `pool_candidate_status` to `EXPIRED`, need to use database tool for this

Observe their application page

image

Cause, I believe is due to this below snippet
The status `EXPIRED` returns true for both, and that was before this PR

image

Where 

image

image

And

image

image

brindasasi (approved):
Looking great 👍⭐️ I will create a sep bug for that issue

PR #15002, opened 2025-11-04
[Fix] Pool billingual field check

### LINKED ISSUE #14675: ♻️ `PoolIsCompleteValidator` should be an attribute and not a validator

esizer:
resolves #14675
## 👋 Introduction

Fixes the billingual field check for pool completeness.

## 🕵️ Details

We were only checking that either was empty, but it fails if one is empty and the other is not.

## 🧪 Testing

1. "Complete" a pool
2. confirm badge updates to complete

### DIFF
--- a/api/app/Models/Pool.php
+++ b/api/app/Models/Pool.php
@@ -548,7 +548,10 @@ class Pool extends Model
                 continue;
             }
 
-            if (empty($value['en']) || empty($value['fr'])) {
+            if (
+                (empty($value['en']) && ! empty($value['fr'])) ||
+                (! empty($value['en']) && empty($value['fr']))
+            ) {
                 return false;
             }
         }


yonikid15:
Is there an issue for this or it's a quick fix?

brindasasi (approved):
Looks good ! 👍  Thanks!

PR #15006, opened 2025-11-04
[Feature] Improve text search

### LINKED ISSUE #14558: ✨ Improve general search

petertgiles:
🤖 Resolves #14558 

## 👋 Introduction

Provides a "beta" option to use a different search scope for text searching on the user table.

## 🕵️ Details

It turned out to be pretty tricky to add an option to the filter menu in the table.  Jerbo provided an alternate design to switch off the new search under the table:
https://github.com/GCTC-NTGC/gc-digital-talent/issues/14558#issuecomment-3487435089

I copied the existing PHPUnit test file and repointed all the queries at the beta test.  I few of the tests had to be updated or removed due to some intentional changes to the way test works.  You can see those changes in [update tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/15006/commits/3ab1022a0f5f2cbf17e7aba5b925e5ac73fee53e).

After discussions with Jerbo and Gray, I did end up adding some simple index weighting to improve results.

## 🧪 Testing

> [!TIP]
> I suggest seeding many users into your DB with the big seeder for testing. 

1. If you are using existing database rows instead of newly seeded rows then run  `./artisan app:update-user-text-search-indexes` to update your indexes.

- [ ]     User table "beta" general search works.
- [ ]     User table "old" general search still works.
- [ ]     Pool candidate table general search still works.
- [ ]     Pool candidate general search + CSV download still works.
- [ ]     Community talent table general search still works.
- [ ]     WFA table general search still works. (It doesn't work on main right now. I included a fix in this PR.)
- [ ]     Tests complete.

## 📸 Screenshot

image

## :truck: Deployment

This doesn't actually have to be done during deployment.  It can safely be done later.

1. `./artisan app:update-user-text-search-indexes`

I estimate that this will take about 45 minutes in production.  It can safely be interrupted and restarted if necessary.

### DIFF
--- a/api/app/Utilities/PostgresTextSearch.php
+++ b/api/app/Utilities/PostgresTextSearch.php
@@ -26,7 +26,8 @@ class PostgresTextSearch
         '\('.'|'.       // ( anywhere
         '\)'.'|'.       // ) anywhere
         '<'.'|'.        // < anywhere
-        '>'.            // > anywhere
+        '>'.'|'.        // > anywhere
+        "'$".           // ' at the end
         '/';
 
     // convert a search string to query text with prefix matching
--- a/api/tests/Unit/PostgresTextSearchTest.php
+++ b/api/tests/Unit/PostgresTextSearchTest.php
@@ -97,23 +97,23 @@ class PostgresTextSearchTest extends TestCase
             ],
             'prefix, removes terms with invalid patterns' => [
                 'matchingType' => 'PREFIX',
-                'searchString' => 'term1& &term2 term3| |term4 :term5 te:rm6 term7: term!8 term9! ! ( ) < >',
+                'searchString' => "term1& &term2 term3| |term4 :term5 te:rm6 term7: term!8 term9! ! ( ) < > term10'",
                 'expectedOutput' => '',
             ],
             'exact, removes terms with invalid patterns' => [
                 'matchingType' => 'EXACT',
-                'searchString' => 'term1& &term2 term3| |term4 :term5 te:rm6 term7: term!8 term9! ! ( ) < >',
+                'searchString' => "term1& &term2 term3| |term4 :term5 te:rm6 term7: term!8 term9! ! ( ) < > term10'",
                 'expectedOutput' => '',
             ],
             'prefix, allows terms with valid patterns' => [
                 'matchingType' => 'PREFIX',
-                'searchString' => 'te&rm1 te|rm2 !term3',
-                'expectedOutput' => 'te&rm1:* & te|rm2:* & !term3:*',
+                'searchString' => "te&rm1 te|rm2 !term3 'term4 te'rm5",
+                'expectedOutput' => "te&rm1:* & te|rm2:* & !term3:* & 'term4:* & te'rm5:*",
             ],
             'exact, allows terms with valid patterns' => [
                 'matchingType' => 'EXACT',
-                'searchString' => 'te&rm1 te|rm2 !term3',
-                'expectedOutput' => 'te&rm1 & te|rm2 & !term3',
+                'searchString' => "te&rm1 te|rm2 !term3 'term4 te'rm5",
+                'expectedOutput' => "te&rm1 & te|rm2 & !term3 & 'term4 & te'rm5",
             ],
         ];
     }


apps/web/src/pages/Users/IndexUserPage/IndexUserPage.tsx
```diff
@@ -44,7 +50,38 @@ export const IndexUserPage = () => {
...
+                  description: "Body of a note around testing new search",
+                })}
+              </p>
+              <Checkbox
+                id="oldSearch"
+                label={intl.formatMessage({
+                  defaultMessage: "Use the old table search",
+                  id: "DDnn7i",
+                  description: "Check box to toggle new search",
+                })}
```
vd1992: Minor quibble. 

I think this should default to checked. Normally experimental features tend to be opt-in right? Defaulting to checked has the table search the way the user is familiar with as a default.

api/app/Builders/UserBuilder.php
vd1992: Before you could use operators like `||` and `&&`
But don't know if enough people used them to miss it hehe

image

image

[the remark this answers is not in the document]
petertgiles: Yeah, what you're seeing is that the search ranking comes first as the ordering.  And then whatever ordering set by the table comes second.

petertgiles: Probably not the way you think.  :grin: It looks like it was simply dropping those symbols.  So && is treated as & by default but so is ||.
image
I don't like that error though...  let me see what I can do with that.

petertgiles:
Reviewed in standup:
- Vachan would like more eyes on this one
- Vachan OK to review the code if we can put it in Azure for Jerbo, Gray, Jamie to test
- Peter will set up testing in Azure

api/app/Utilities/PostgresTextSearch.php
```diff
@@ -0,0 +1,131 @@
...
+    public static $invalidPatterns =
+        '/'.
+        '^&'.'|'.       // & at the beginning
+        '&$'.'|'.       // & at the end
+        '^\|'.'|'.      // | at the beginning
+        '\|$'.'|'.      // | at the end
+        ':'.'|'.        // : anywhere
+        '^.+?\!'.'|'.   // ! not at the beginning
+        '^\!$'.         // ! at the end
+        '/';
```
vd1992: Oh, seems like a few more characters to ignore to prevent errors. Unlikely to be used other than by accident, but prevents unneeded errors and logging. 
```
(
)
<
>
```

brindasasi: I get errors with single quote alone and something starting like this  '[

brindasasi: may be instead of adding invalid chars list , we could regex allowable chars and ignore others? upto you

petertgiles: The big problem is that I don't know what the list of allowed or disallowed characters are.  I'd prefer to err on the side of permissiveness.  Let them try whatever they want.  The stakes are low if they try something that doesn't work.  They'll just get an error message.

petertgiles: '[ is an error because it is essentially empty.  There's nothing to search for there.

api/app/Builders/UserBuilder.php
```diff
@@ -827,4 +829,45 @@ public function whereGeneralSearch(?string $searchTerm): self
...
+                // add the calculated rank column to allow for ordering by text search rank
+                // weighting the exact matches a little heavier
+                ->addSelect(DB::raw(<<<'SQL'
+                        (
+                            (ts_rank(user_search_indices.searchable, calculations.prefix_match_query) * 0.9) +
+                            (ts_rank(user_search_indices.searchable, calculations.exact_match_query) * 1.1)
+                        ) / 2 AS search_rank
+                        SQL))
+                // Now that we have added a column, query builder no longer will add a * to the select.  Add all possible columns manually.
+                ->addSelect(['users.*'])
```
brindasasi: do we need users.* in here? we have only handful of columns in these tables.. can't we just select required columns?

PR #15010, opened 2025-11-05
[Bug] No setting past expiry dates

### LINKED ISSUE #15004: 🐛 UAT | Admin updates Expiry date to the past date for the qualified candidates

Description: When the admin is submitting 'Final Decision' status for the candidates, the admin cannot put the past date as Expiry date which is valid but when for the same candidate, when admin tries to put the 'Past date' as an Expiry date, admin is able to do the same, which is kind of invalid, as the validation for both 'Expiry date' should be same.

Expected result should be - The expiry date cannot be selected to past date, and validation should be same on all the pages

### 🕵️ Details

Environment: UAT

### 📋 Steps to Reproduce

1. Access GC digital portal as an Admin user
2. Navigate to Candidates and search for the candidate who is Qualified
3. Open the candidate details and verify the Expiry date is in the future
4. Click on the Expiry date to update it in the past date, as the candidate needs to be marked as 'Expired: Qualified'
5. Verify admin user can update the expiry date to the past date
6. Go back to the candidates table and open any candidate who is in 'To Assess' status
7. Record final decision for that candidate
8. Try to select the past date as an Expiry date while recording the final decision for the candidate
9. Admin user receives an error while putting past date as an expiry date

### Admin tries to select Past date as an Expiry date

Image

### Admin changes the qualified candidate's expiration date to the past date 

Image

### 🙋‍♀️ Proposed Solution

- The Expiration date logic should work same for the Final Decision page as well as Candidates page

### ✅ Acceptance Criteria

- [ ] Admin should not be able to update the expiration date to the past date for Qualified Candidate

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

vd1992:
🤖 Resolves #15004

## 👋 Introduction

Prevent setting past expiry dates in the "Change expiry date" dialog. I copied the validation from the placement dialog to keep them synced

## 🧪 Testing

1. Navigate to an application
2. Cannot set expiry dates in the past

## 📸 Screenshot

image

### DIFF
--- a/api/app/GraphQL/Validators/QualifyCandidateValidator.php
+++ b/api/app/GraphQL/Validators/QualifyCandidateValidator.php
@@ -22,7 +22,7 @@ final class QualifyCandidateValidator extends Validator
     {
         $id = $this->arg('id');
         $candidate = PoolCandidate::findOrFail($id);
-        $endOfDay = Carbon::now()->endOfDay();
+        $startOfDay = Carbon::now()->startOfDay();
 
         $statusesArray = [
             PoolCandidateStatus::NEW_APPLICATION->name,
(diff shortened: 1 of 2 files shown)


JamesHuf:
How do we want to handle setting it to the current date? It looks like the front-end will allow it but the back-end won't. They should probably be in agreement.

PR #15016, opened 2025-11-05
 ⌨ Keyboard Accessibility for Nav Menu 

### LINKED ISSUE #11986: ⌨ Keyboard navigation of new menu is confusing
## ⌨ Accessibility Issue

Our new menu has sensible keyboard shortcuts, however it seems to be split into 3-4 different navs which makes it confusing due to proximity.

When navigating by keyboard, you can use arrow keys to switch between the different links. However, since it is split into different entities, you are forced to tab between the different navs and then can use arrow keys inside.

The issue is, visually it is not clear what consititutes a menu here so it seems arbitrary and adds to confusion.

## 🦋 Expected Behaviour

Different menus are cleary distinguised visually if they are different programatically.

## 📋 Steps to Reproduce

1. Usingf your keyboard, tab over to the menu
2. Attempt to use arrow keys to navigate menu items
3. Observe some items needing to tab to them

## 📸 Screenshot

You can only use arrow keys within each red box. You must tab between boxes.

## 🙋‍♀️ Proposed Solution

1. Combine everything in to a single menu
2. Discuss other options if that won't work

## ✅ Acceptance Criteria

- [ ] Nav groups are clearly distinguishable both visually and programmatically
- [ ] Arrow key nav works within everything on the nav bar, treated as one group
- [ ] Visually the menu should remain unchanged

bella77-69:
🤖 Resolves #11986

## 👋 Introduction

This PR fixes the keyboard navigation in the main navigation menu. The menu is split into separate sections and users had to tab between each section, now users can navigate all menu items using arrow keys. 

## 🕵️ Details

- Visually the menu is unchanged
- Users can use arrow keys to navigate the entire nav menu without having to tab between sections
- Nav groups are clearly distinguishable both visually and programmatically

## 🧪 Testing

1. Build app
2. Navigate to home page
3. Using your keyboard:
    - Tab to get the focus on the first navigation menu item 
    - Use the arrow keys to navigate through the menu items
    - Use the spacebar to open/close items that are expandable
    - Sign in as admin
    - repeat the steps above
4. Test in mobile view

### DIFF
--- a/apps/web/src/components/NavMenu/MainNavMenu.tsx
+++ b/apps/web/src/components/NavMenu/MainNavMenu.tsx
@@ -1,11 +1,4 @@
-import {
-  useEffect,
-  useState,
-  KeyboardEventHandler,
-  useRef,
-  ReactNode,
-  ReactElement,
-} from "react";
+import { useEffect, useState, KeyboardEventHandler, useRef } from "react";
 import FocusLock from "react-focus-lock";
 import { m, AnimatePresence } from "motion/react";
 import XMarkIcon from "@heroicons/react/20/solid/XMarkIcon";


apps/web/src/components/NavMenu/MainNavMenu.tsx
```diff
@@ -72,82 +72,98 @@ const MainNavMenu = () => {
...
-          <>
-            <NavMenu.Item className={borderItem({ borderRight: true })}>
+      <div
+        role="group"
+        aria-label={intl.formatMessage({
+          defaultMessage: "Main navigation",
+          id: "SY1LIh" /*  TODO: translate & update default message */,
+          description: "Label for the main navigation",
+        })}
+      >
```
esizer: I'm not sure that the correct answer here is because I have always been confused by this part of our nav but, I don't know main nav make sense (or if it being its own group even does).

I think we need to either:

 1. Try to group it with the other elements or,
 2. Find a different name for it

apps/web/src/components/NavMenu/Menu.tsx
```diff
@@ -95,9 +95,38 @@ const Menu = ({
...
+    event.preventDefault();
+    event.stopPropagation();
+    // calculate next index
+    const nextIndex = isNext
+      ? (currentIndex + 1) % navItems.length
+      : currentIndex > 0
+        ? currentIndex - 1
+        : navItems.length - 1;
+    // focus the element at next calculated index
+    navItems[nextIndex]?.focus();
```
esizer: This is really close to what I would expect! Though, not quite there yet. I think this might have to do with the fact that `NavMenu` primitice has keyboard shortcurts already and they might be competing. I havent looked too deep yet though.

 - When opening a dropdown, if you hit the next key, it skips the first item
 - When leaving the dropdown, it does not close itself
 - On mobile, you can't navigate past the main row (Home, theme, lang)
    - It requires you to press tab
    - Then you naturally cycle back and are stuck again

One other thing that I'm not sure are real issues or just unexpected and might need outside input.

whne you get to the end of the menu, you are cycled back to the beginning. This seems like an okay change but causes some odd behaviour. We present the menu as if the notfication button is part of it and you can get to it.

While you can get there via tab, I expect to use the arrow keys. Though, im not sure if that is reasonable. Just noting it to see if Im not alone in this.

apps/web/src/components/NavMenu/Menu.tsx
```diff
@@ -185,38 +214,48 @@ const Menu = ({
...
-                            class: "hidden before:mr-3 sm:inline-flex",
-                          })}
-                        >
-                          <NotificationDialog
-                            open={isNotificationDialogOpen}
-                            onOpenChange={setNotificationDialogOpen}
+                  <div
+                    role="group"
+                    aria-label={intl.formatMessage({
+                      defaultMessage: "Account & Notifications navigation",
```
mnigh: I would recommend not using an ampersand for i18n. @NienkeBr, could you advise please?

NienkeBr: I don't see where exactly this would show but in general I'm indeed in favour of avoiding the use of ampersands where possible (exceptions are official names for example)

yonikid15:
Issue completed in [#11986 ](https://github.com/GCTC-NTGC/gc-digital-talent/pull/15207)

PR #15041, opened 2025-11-06
[Fix] Document downloads not working

### LINKED ISSUE #15014: 🐛 Download excel sheet on users table is broken

There is an error caused by the location_preferences receiving null instead of an array.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Go to users table
2. Try downloading a excel list of users
3. Don't add any location preferences

Image

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

Check if the filter param is null before and return `$this` if true.

// UserBuilder.php

Image

### ✅ Acceptance Criteria

- [ ] Can download excel document on users table

### 🔄 Regression

Yes

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #15014 

## 👋 Introduction

Addresses an issue where documents could not download unless a location filter was applied.

## 🧪 Testing

1. Start the worker `make queue-work`
2. Login as `admin@test.com`
3. Go to different tables that download user data
4. Confirm you can download the data from the table without providing any location based filters

### DIFF
--- a/apps/playwright/tests/admin/user-excel.spec.ts
+++ b/apps/playwright/tests/admin/user-excel.spec.ts
@@ -9,8 +9,7 @@ test.describe("User Excel", () => {
     await loginBySub(userPage.page, "admin@test.com", false);
     await userPage.goToIndex();
     await userPage.waitForGraphqlResponse("UsersPaginated");
-    await userPage.page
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
Works great!  Thanks for the test refactor, too.

PR #15042, opened 2025-11-06
[Fix] Add GIN index on user `jsonb` columns

### LINKED ISSUE #15000: 🐛 Prod pages timing out

These two pages are hitting the 60 second timeout and crashing with a 504.

https://talent.canada.ca/en/admin/talent-requests/fb8be7cf-8b37-43af-a933-0eeec45f19bd
https://talent.canada.ca/en/admin/talent-requests/0f739521-8f32-4a22-b820-0c727be92fa4

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] pages don't time out

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #15000 

## 👋 Introduction

This is an attempt to improve the performance for SQL queries on jsonb columns for users.

## 🕵️ Details

You can read more about why this might be a good solution at the following: https://www.crunchydata.com/blog/indexing-jsonb-in-postgres

You see see the results of my tests at https://github.com/GCTC-NTGC/gc-digital-talent/issues/15000#issuecomment-3487629157

> [!IMPORTANT]
> If this does not result in a noticeable difference, we should not proceed. GIN indexes increase the upfront cost of the query as well as updates to these columns. 

## 🧪 Testing

1. Seed a lot of pool candidates (10_000+)
2. Run a query against the pool candidates using filters for the columns that we indexed here
3. Note the time it took (even better is to run the SQL directly weith `EXPLAIN ANALYZE`)
5. Run migration `make artisan CMD="migrate"
6. Repeat step 2-3
7. Compare the results to see if there is a noticeable difference

### DIFF
--- a/api/database/migrations/2025_11_04_191556_add_gin_index.php
+++ b/api/database/migrations/2025_11_04_191556_add_gin_index.php
@@ -1,0 +1,31 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\DB;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        DB::statement('CREATE INDEX accepted_operational_requirements_gin ON users USING GIN (accepted_operational_requirements)');
+        DB::statement('CREATE INDEX flexible_work_locations_gin ON users USING GIN (flexible_work_locations)');
+        DB::statement('CREATE INDEX location_preferences_gin ON users USING GIN (location_preferences)');
+        DB::statement('CREATE INDEX position_duration_gin ON users USING GIN (position_duration)');
+        DB::statement('CREATE INDEX indigenous_communities_gin ON users USING GIN (indigenous_communities)');
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        DB::statement('DROP INDEX IF EXISTS accepted_operational_requirements_gin');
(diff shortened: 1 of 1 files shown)


vd1992:
Also, an aside, to be on the same page. I am **NOT** using `admin@test.com` for testing this.

vd1992:
As an aside, to follow up on stand-up comments. I think I found out the cause for discrepancy between in-website querying and graphiQL querying. I was missing the sort argument, it has a sizable impact. My mistake overlooking it

Commenting and un-commenting the below was what I needed

```
		# orderBy: [
		# {
		# column: "submitted_at",
		# order: DESC
		# },
		# {
		# column: "id",
		# order: DESC
		# }
		# ]
```

vd1992:
I don't get what is going on with my pool candidates table but testing with the **USERS TABLE**
(Reference, as community admin with ~8000 users)

### Main

Gradually stacking on jsonb filters, so, flexible locations -> location preferences -> duration
Observing a rising SQL time and general slowness

image


### Branch

Repeating the above
I noticed a drastic reduction in time. Slow query logging never hit, query as whole takes about 1/3 the time, so any initial cost is not significant

image

petertgiles:
@brindasasi I think this is becoming more urgent.  On the dashboard I'm seeing more requests hitting that 60 second timeout and CPU spikes on the DB server.

PR #15048, opened 2025-11-07
[Fix] Improve pool candidate authorized to view permission checks

### LINKED ISSUE #14414: 🐛 Eloquent query performance poor for complex roles setup

It appears some queries or pagination stuff sometimes performs very poorly
One way to replicate is to assign yourself process operator, not admin, to many processes then trying to access the candidates table. Must have many candidates seeded first (thousands). The API request struggles

### 🕵️ Details

One of the standouts is that field `paginatorInfo.total` tanks performance something drastic
Creates one real gnarly SQL statement

### 📋 Steps to Reproduce

1. Seed many candidates (>2000)
2. Assign yourself many process operator roles (>10)
3. Open the candidates table

Running in GraphiQL, fetching 10 candidates
Adding in  `paginatorInfo.total` then commenting it out, comparing the difference, is quick to do

<details><summary>Full query, comment out as you see fit</summary>
<p>

```
query c {
  poolCandidatesPaginatedAdminView(first: 10, where: {}) {
    data {
      id
      poolCandidate {
        id
        notes
        status {
          value
          label {
            en
            fr
          }
        }
        placedDepartment {
          id
          name {
            en
            fr
          }
        }
        category {
          weight
          value
          label {
            en
            fr
          }
        }
        finalDecisionAt
        finalDecision {
          value
        }
        assessmentStep
        assessmentStatus {
          assessmentStepStatuses {
            step
          }
          overallAssessmentStatus
        }
        pool {
          id
          processNumber
          name {
            en
            fr
          }
          classification {
            id
            group
            level
          }
          workStream {
            id
            name {
              en
              fr
            }
          }
          publishingGroup {
            value
            label {
              en
              fr
            }
          }
          closingDate
          areaOfSelection {
            value
          }
          screeningQuestionsCount
          assessmentSteps {
            sortOrder
            title {
              localized
            }
            type {
              label {
                localized
              }
            }
          }
        }
        finalDecision {
          value
          label {
            en
            fr
          }
        }
        assessmentStep
        assessmentStatus {
          overallAssessmentStatus
        }
        user {
          id
          email
          firstName
          lastName
          preferredLang {
            value
            label {
              en
              fr
            }
          }
          lookingForEnglish
          lookingForFrench
          lookingForBilingual
          currentCity
          department {
            id
            name {
              localized
            }
          }
          currentProvince {
            value
            label {
              en
              fr
            }
          }
        }
        submittedAt
        suspendedAt
        removedAt
      }
      skillCount
    }
    paginatorInfo {
      total
    }
  }
}
```

</p>
</details> 

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] improved performance for many pool candidates when current user has many team roles

esizer:
🤖 Resolves #14414 

## 👋 Introduction

This pre-computes the team IDs to avoid looping in every scope which was the most signifcant contributor to slow down.

## 🕵️ Details

I don't know if this is even possible. @petertgiles pointed out, we have a custom user checker to handle protected requests. Since this skips over the `isAbleTo` for performance, we might be losing critical securiry.

Putting it up anyway to see what people thinkg. Maybe it will spark an idea? :woman_shrugging: 

---

The reason that the query got increasingly slow when having more team based roles was due to the laratrust `isAbleTo` method being slow when not cached (first run ever ~1min). Within the scope for the paginated pool candidates query, we were looping over the users team for each permission check calling `isAbleTo` in order to get the team ids. 

This was leading to a lot of calls to `isAbleTo` which was adding time per role (if you did not have platform admin).

This removes those calls and makes a direct database call to get the IDs to avoid some of the extra code/hydration in the `isAbleTo` call.

We still have some guards where we check a singular `isAbleTo` before using the team IDs so this should have the same security. But, extra documentation was added to try to emphasize why this exists.

Ideally, we do not avoid `isAbleTo` calls, unless it is causing performance issues due to many calls.

> [!NOTE]
> More information on this can be found in both this PR and the ticket dicusccion. 

## :question: Question

This only targets the pool candidates scope since we have had issues with it in the past. However, this pattern of looping over team based roles exists in other scopes.

Should we:

1. Apply this pattern to those now?
2. Create a new ticket to apply this pattern more broadly
3. Handle it on an as-needed basis?

## 🧪 Testing

1. Seed a lot of pool candidates (5000+)
4. Give a user a lot of team based roles (100+ process operator is good) 
5. Check bopth main and this branch
6. Compare the overall time
7. Confirm there is a noticable difference

### Before

before

### After

after

### DIFF
--- a/api/app/Builders/PoolCandidateBuilder.php
+++ b/api/app/Builders/PoolCandidateBuilder.php
@@ -720,6 +720,7 @@ class PoolCandidateBuilder extends Builder
             ->toArray();
     }
 
+    // main authorization scope for viewing PoolCandidateAdminView
     public function whereAuthorizedToViewPoolCandidateAdminView(): self
     {
         /** @var \App\Models\User | null */


esizer:
> I had a random thought the other day. What if we had a cache-warming command? It could iterate though users with team roles and all the teams. We could run it on deploy or nightly or something.

That may not be a bad idea. It would defintlety be a smaller touch than what I have here so far :thinking:  Though, doesn't it only cache for 1min? Seems like if we wan it nightly, it would be invalid by the time people needed it. Or am I reading this incorrectly? :sweat_smile: 

https://github.com/santigarcor/laratrust/blob/0077a654b107da5c535d382626b35e3a5472a81d/src/Checkers/User/UserDefaultChecker.php#L177-L179

esizer:
> Another thought I had: doesn't our checker support arrays? I remember we had that fix for it a while back. Does that trigger a single network request? Does that AND/OR behaviour of it match what we need?

I'm not sure if passing an array would be possible in this scenario.

The calls that are slowing it down (from what I could tell) are the ones to `isAbleTo` in the team check where we are just getting the team IDs where you have that permission.

This test PR just avoided that by getting them upfront directly from the database so we did not need to hydrate and do all the extra checks. 

I still cannot see why it inflates it so much. I did some profiling of that function and while the first check is 55-65ms, once the cached, the rest are only 3ms :thinking: 

My user has 6 roles, so we loop through each 6 times per request (plus 11 additional one off calls).

So... 

`55 + (6 * 3) + (11 * 3) =  106ms`

It should be 100ms but for whatever reason, it ends up being a 700ms difference :face_with_spiral_eyes: I must be missing something, there is no way the `$user->rolesTeams()->get()` is taking 700ms :stuck_out_tongue:

esizer:
okay, I added some more discrete server timings and it is quite clearly the `isAbleTo` calls :cry: Though, the cached result helps significantly. The problem being... the first result could time out :woman_shrugging: 

### Initial

swappy-20251118_154217


### Cached

swappy-20251118_154323

petertgiles:
That's so weird.  I'd be open to bypassing the checker with our own implementation if we really carefully signposted it.

1. comments around the bypass to say why it's there
2. comments on how we can easily check if it's still a problem and how we can remove it in the future
3. links to the bug on laratrusts' GH repo
4. a list started in our checker of all the places it's been bypassed

petertgiles:
It's not even that complicated a class.  smh

https://github.com/santigarcor/laratrust/blob/master/src/Checkers/User/UserDefaultChecker.php

vd1992:
Just skimming through and thought I'd bring up this issue since it was posted in that repository
[www.github.com/santigarcor/laratrust/issues/645](http://www.github.com/santigarcor/laratrust/issues/645)

PR #15049, opened 2025-11-07
[Fix] Only run pooling query when document has focus

### LINKED ISSUE #15036: 🐛 Multiple Tab Logout and Refresh Token Race Condition

Clients experienced unexpected logouts when working in multiple browser tabs. Multiple refresh requests using the same token were logged within 2-second windows, indicating race conditions. These overlaps were confirmed in both HTTP and console logs. 

Expected behaviour: Users should be able to work in multiple tabs without being logged out unexpectedly, even when tokens expire and refresh requests are triggered simultaneously.

### 🕵️ Details

**Operating System**: Multiple (user-reported)
**Browser**: Multiple (user-reported)

Included logs demonstrated that several refresh requests were made nearly simultaneously from different tabs, using the same refresh token. Backend appears to reject all but one refresh request, causing tabs with failed refresh attempts to log out.

### 📋 Steps to Reproduce

1. Open the application in multiple browser tabs under the same user account.
2. Wait until the authentication token expires (or simulate token expiry).
3. Observe refresh requests triggered from several tabs within a short time window.
4. Note that at least some tabs will be forcibly logged out after refresh attempts.

###  📎Attachment

[parallel_refresh_attempts (1).csv](https://github.com/user-attachments/files/23396455/parallel_refresh_attempts.1.csv)

### 🙋‍♀️ Proposed Solution

- Implement client-side coordination for refresh requests across tabs using BroadcastChannel, localStorage, or similar mechanism, so only a single refresh flow executes and other tabs re-use the new token.
- Consider adding a short window on the backend during which the previous refresh token remains valid after rotation to tolerate overlaps.
- Review session/token rotation logic to avoid invalidating concurrent requests from the same user/token session.

### ✅ Acceptance Criteria

- [ ] A user can use the app in multiple tabs without unexpected logouts when tokens expire.
- [ ] Only one refresh request is processed for a user/session at a time, even if triggered in several tabs.
- [ ] Consistent authentication session across all open tabs after a refresh.
- [ ] Logs confirm no overlapping/invalid refresh requests leading to logouts.

esizer:
🤖 Resolves #15036 

## 👋 Introduction

Updates our polling query hook to only execute a query when the current document has focus.

## 🧪 Testing

> [!TIP]
> It is easier if you increase the polling frequency. I did this by reducing the multiplier from 60 -> 10 for a 10 second poll. 

1. Build `pnpm dev:fresh`
2. Login as any user
3. Open two tabs
4. Have only one of those tabs focused
5. Confirm it runs the polling query at the interval
6. Change back to the other tab
7. Confirm no polling ocurred while it was inactive
8. Confirm the polling resums once it gains focus

### DIFF
--- a/apps/web/src/hooks/usePollingQuery.ts
+++ b/apps/web/src/hooks/usePollingQuery.ts
@@ -24,10 +24,17 @@ const usePollingQuery = <TData, TVariables extends AnyVariables>(
   useEffect(() => {
     remainingRef.current = delay * 1000;
     lastStartTimeRef.current = Date.now();
+    if (timeoutRef.current) {
+      clearTimeout(timeoutRef.current);
+      timeoutRef.current = null;
+    }
   }, [delay, disabled, queryArgs.variables]);
 
   useEffect(() => {
-    if (timeoutRef.current) clearTimeout(timeoutRef.current);
+    if (timeoutRef.current) {
+      clearTimeout(timeoutRef.current);
+      timeoutRef.current = null;
+    }
 
     // Do nothing if we are already fetching
     if (disabled || result.fetching) return undefined;
@@ -61,13 +68,6 @@ const usePollingQuery = <TData, TVariables extends AnyVariables>(
     disabled,
     isWindowActive,
   ]);
-
-  useEffect(() => {
-    if (!result.fetching && isWindowActive && !disabled) {
-      remainingRef.current = delay * 1000;
-      lastStartTimeRef.current = Date.now();
-    }
-  }, [result.fetching, isWindowActive, disabled, delay]);
 
   return [result, executeQuery];
 };


petertgiles:
This does work.  :+1:   However, it resets the polling timer every time the tab goes inactive.  I'm worried that if someone is working between two tabs, they will be switching back and forth faster than the polling timer and never get a notification.  I suggest keeping the polling timer stable in each tab and simply skipping the network query if the tab is inactive.  I think this will lessen the chances of never getting the notification.

petertgiles:
This is cool!  I like this strategy.  I think there's a bug though.  I started logging it and sometimes the number goes up when pausing the timer.
image

mnigh (approved):
LGTM in Safari.

PR #15054, opened 2025-11-10
[Fix] `Accordion` heading structure

### LINKED ISSUE #15011: ⌨ `Accordion` contain invalid heading structure

Our `Accordion` component contains both a `h3` and `h*` (* depends on the `Trigger` `as` prop).

### WCAG 2.1 Criteria

- https://www.w3.org/WAI/WCAG22/quickref/?versions=2.1#qr-content-structure-separation-programmatic

### 📋 Steps to Reproduce

1. Open our accordion component in storybook
2. Confirm that you see an `h3` and `h2` in the trigger

### 🕵️ Details

The markup for our trigger is:

- h3
  - button
    - h2

This has sevral problems with it. 

1. Heading ranks are out of order
2. You cannot put a heading inside of another heading

_No response_

_No response_

_No response_

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

By default, the `Accordion.Header` is a `h3`. You can provide the `asChild` prop to set the appropriate heading rank but, we do not.

I'm not sure what to do because the design requires we do some odd invalid HTML. Agian, I beleive we need to take a look at our accordions and simplify them significantly.

We have too much going on in the trigger element causing many accessibility issues and a very brittle component.

In the interim, I think maybe just making it a div? This will still be invlaid HTML but, it atleast will solve one major issue.

### ✅ Acceptance Criteria

- [ ] `Accordion.Trigger` only contains a single heading
- [ ] That heading is still the appropriate heading rank

esizer:
🤖 Resolves #15011 

## 👋 Introduction

Fixes an issue where our `Accordion` rendered nested headings.

## 🧪 Testing

1. Run storybook `pnpm storybook`
2. NAvigate to the accordion story
3. Confirm that the accordion does not have nested headings

### DIFF
--- a/packages/ui/src/components/Accordion/Accordion.tsx
+++ b/packages/ui/src/components/Accordion/Accordion.tsx
@@ -220,35 +220,37 @@ const Trigger = forwardRef<
     });
 
     return (
-      <AccordionPrimitive.Header className={header()} {...titleProps}>
-        <AccordionPrimitive.Trigger
-          ref={forwardedRef}
-          className={btn()}
-          {...rest}
-        >
-          <span className="-mt-0.5 rounded-full p-1 transition-colors duration-150 group-focus-visible/btn:bg-focus group-focus-visible/btn:text-black">
-            <ChevronRightIcon
-              className={iconSize({
-                class:
-                  "rotate-0 transform leading-none transition-transform duration-150 group-data-[state=open]/btn:rotate-90",
-              })}
-            />
-          </span>
-          <span className="flex grow flex-col">
-            <Heading className={heading()}>{children}</Heading>
-            {subtitle && (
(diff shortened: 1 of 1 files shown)


vd1992:
Was this feature around before 🫤 
Is Chromatic implying we're adding accessibility violations?  

image

https://www.chromatic.com/build?appId=61e099a1bb1465003a73d3ce&number=13320

image

image

esizer:
> Was this feature around before 🫤
> Is Chromatic implying we're adding accessibility violations?

@mnigh recently enabled it based on some discussions we had last retro. Though, looks like it is working and found an issue! :+1:

petertgiles (approved):
Nice.  👏

PR #15058, opened 2025-11-12
[Refactor] Update `NotificationList` to use env polling interval

### LINKED ISSUE #15057: ♻️ Update `NotificationList` to respect polling interval
### ♻️ Refactor

We have a polling interval set in env for notifications. Currently only the count for the badge respects this. We would like the list of notifications to also use that setting.

### 🕵️ Details

_No response_

### 🙋‍♀️ Proposed Solution

Implement the same logic for polling from `NotificationDialog` in the `NotificationList`

### ✅ Acceptance Criteria

- [ ] `NotificationList` uses the notiifcation polling interval

esizer:
🤖 Resolves #15057 

## 👋 Introduction

Updates the `NotificationList` to use the `NOTIFICATION_POLLING_INTERVAL` env var. 

## 🕵️ Details

Hoping this improves the race condition for polling too frequently. This also pauses the count query while the dialog is open so we have one fewer queries happening.

## 🧪 Testing

> [!TIP]
> Testing is easier if you reduce the multiple in the `~/constants/notifications` file so you arent waiting a full minute. 

1. Build `pnpm dev:fresh`
2. Login as any user
3. Open the `NotificationDialog`
4. Observe network traffic
5. Confirm that the polling query is fired at the expected interval from `NOTIFICATION_POLLING_INTERVAL`

### DIFF
--- a/apps/web/src/components/NotificationList/NotificationActions.tsx
+++ b/apps/web/src/components/NotificationList/NotificationActions.tsx
@@ -56,8 +56,6 @@ const NotificationActions = ({
   const [{ fetching: markingAllAsRead }, executeMarkAllAsReadMutation] =
     useMutation(MarkAllNotificationsAsRead_Mutation);
 
-  console.log(fetching);
-
   const handleMarkAllNotificationsAsRead = () => {
     executeMarkAllAsReadMutation({})
       .then(() => {


brindasasi (approved):
works fine .. polling has been reduced dramatically.. thanks for this. Please merge in after fixng the linting issues.. Thanks!

Issue #14851, opened 2025-10-16
🧪 Testing: Update applicant location preferences: Remote, Hybrid, On-site


mauli0110:
### 🎯 Target Bug(s) / Feature

https://github.com/GCTC-NTGC/gc-digital-talent/issues/9699

### 🛠 Test Type

e2e test

### 📍 Scope / Component

Update location preferences

### ✅ Test Objective

Objective of this test is to validate the update in 'Work location preferences'

1. New section 'Flexible work location options' under 'Work location preferences'
2. Telework is removed from work location preferences
3. New label 'Location exclusions' under 'Please indicate if there is a...' heading

### Scenario 1 - Verify the 'Work location preference' is updated for all existing users

1. Login as an existing user with a saved work preference.
2. Go to Applicant’s Profile -> Work Preferences section.
3. Verify Flexible work location options are present with selected options in green check mark
4. Verify the selected options based on below conditions: (Migration required to validate below conditions)
    - Remote work should be selected if, Telework was selected under work location preferences. 
    - Telework was selected AND at least 1 other region was selected -> Either Hybrid work should be selected or all three selected
    - At least one region was selected -> On-site work should be selected
5. Click Edit and confirm the heading appears under Flexible work location options.
6. Update any option (Remote, Hybrid, or On-site).
7. Confirm Telework option is no longer available.
8. Verify Location exclusions field appears under the “Please indicate if…” label.
9. Add or modify text in Location exclusions.
10. Click Save changes and verify the updates are reflected successfully.

### Scenario 2 - Verify the 'Work location preference' is present for new users

1. Register a new user account and log in for the first time.
2. Go to Applicant's profile -> Work Preferences section.
3. Get started by adding work preferences.
4. Verify none of the flexible work location options, work location preferences selected for new users
5. Select options under Flexible work locations options
6. Confirm Telework option is no longer available in any of those 3 sections
7. Add the details in work location preferences and location exclusions
8. Verify Location exclusions field appears under the “Please indicate if…” label.
9. Click Save changes and verify the updates are reflected successfully.
10. Ensure Flexible work location options are shown with the current selection (green check mark).

### Scenario 3 -Verify the admin view of 'Work location preference' for existing and new users

1. Login as Admin or Community user, who is having necessary privileges.
2. Search with any existing or new users from Users tab
3. Navigate to Work preferences
4. Validate all the necessary details are present in read only view such as - Flexible work location options, Job locations and Location exclusions
[...]

PR #15064, opened 2025-11-12
PW Test: Update applicant location preferences: Remote, Hybrid, On-site

mauli0110:
🤖 Resolves [14851](https://github.com/GCTC-NTGC/gc-digital-talent/issues/14851)

## 👋 Introduction

- Playwright test implementation for location preference update on Applicant and Admin side
- Implemented the teardown method where the new user will be deleted post testing

## 🕵️ Details

There are 3 tests which is automated: 

1. Validate work preferences update on existing user, update the work preferences and validate the change
2. New user should have pre-selected work preferences and validate the same after updating the work preferences
3. An admin view can validate the work preferences update for any user

## 🧪 Testing

1. pnpm exec playwright test tests/admin/location-preference-validation.spec.ts --ui --headed
2. pnpm exec playwright test tests/applicant/location-preference-update.spec.ts --ui --headed

### DIFF
--- a/api/composer.lock
+++ b/api/composer.lock
@@ -5272,37 +5272,35 @@
         },
         {
             "name": "spatie/php-structure-discoverer",
-            "version": "2.3.2",
+            "version": "2.3.3",
             "source": {
                 "type": "git",
                 "url": "https://github.com/spatie/php-structure-discoverer.git",
-                "reference": "6c46e069349c7f2f6ebbe00429332c9e6b70fa92"
+                "reference": "552a5b974a9853a32e5677a66e85ae615a96a90b"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/spatie/php-structure-discoverer/zipball/6c46e069349c7f2f6ebbe00429332c9e6b70fa92",
-                "reference": "6c46e069349c7f2f6ebbe00429332c9e6b70fa92",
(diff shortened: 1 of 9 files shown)


apps/playwright/fixtures/locationPreferenceUpdatePage.ts
```diff
@@ -0,0 +1,155 @@
...
+      [FIELD.LOCATION_EXCLUSIONS]: page.getByRole("textbox", {
+        name: /location exclusions/i,
+      }),
+      [FIELD.TELEWORK_OPTION_LIST_ITEM]: page.getByRole("listitem", {
+        name: /Telework/i,
+      }),
+      [FIELD.TELEWORK_OPTION]: page.getByRole("checkbox", {
+        name: /Telework/i,
+      }),
+    };
```
esizer: I wonder if it is possible to avoid `getByText` and raw `page.locator`.

The preference for `getByRole` ensures that the element we select has the appropriate role for assistive technologies.

I understand it is not always possible but we try to use it as much as possible.

mauli0110: I tried to use `getbyRole `in replacement of `getBytext`, it unables to find the element. The DOM looks like this as below:

DOM looks like this for validating 'Flexible work location options' heading is present: 
image

Execution result: image

I checked, the `getByRole` don't have a role that validates the text or heading present under any tags not only in <h.> but also in <span>
@esizer Would you be able to advice on this?

esizer (approved):
LGTM! 👍

PR #15111, opened 2025-11-18
[Feature] Upgrade to PostgreSQL 13

### LINKED ISSUE #14503: 🛠️ Upgrade to supported postgresql version
### 🛠️ Platform

The postgresql version being used, 12, is unsupported.

### 🕵️ Details

Upgrade to a supported version: https://www.postgresql.org/support/versioning/

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] postgresql is a supported version

petertgiles:
🤖 Resolves #14503 

## 🧪 Testing

For local testing, we're not upgrading, just replacing, the database

1. `docker compose down`
2. `docker volume ls` and find the pgdata volume
3. `docker volume rm 14503-postgres-13_pgdata` but replace with your volume name
4. `docker compose up --detach`

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,7 +25,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:12.22
+        image: postgres:13
         env:
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: password1234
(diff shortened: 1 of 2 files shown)


petertgiles:
> any reason you chose to upgrade to 13 instead of one of 14-18?

Yeah, that was Dan's recommendation.  The reason is that he is able to do an in-place upgrade if he goes version-by-version but if he does a jump then he'll have to do stand up a new instance and move all the data across.

That said, my Jira ticket is in security review right now and I'm guessing they're going to have a problem with upgrading to an eol version even if it very recently went eol.  :shrug:   If that's the case, we'll see what the next recommendation is.

mnigh:
@petertgiles can this PR be closed in favour of #15146?

Issue #15045, opened 2025-11-07
✨ Display two computed fields for current government employees


gobyrne:
### User story
As a talent manager, I want to quickly see if an individual is currently in a acting or substantive position, so I can understand what at-level mobility options are available.

As a recruiter, I want to easily see when an employee's term ends, so I know how much time is available to process recruitment options.

### Backend work
We have 2 computed fields that get populated based on current government experience.  We need to remove the ability for users to directly change them.

| schema field           | recommendation                     |
|------------------------|------------------------------------|
| govPositionType        | remove from mutation input |
| govEndDate    | remove from mutation input |

### Changes to Admin view of Government employee information

All community & admin views current government employee information should include the computed fieds:

1. govPositionType  (shown if employment type = indeterminate)
2. govEndDate   (shown if employment type != indeterminate)

Reminder that for employment type = contractor, the experience is not considered to be government experience.

**For Profile and Application (but normally only one of the two new fields is shown at a time):**
<img width="1039" height="1144" alt="Image" src="https://github.com/user-attachments/assets/b0a0791d-d13f-40bd-b945-6588e71153d4" />

Jerbo also made a nice gif:

![Image](https://github.com/user-attachments/assets/7a91c64b-9d2c-47a7-9a2e-f8e12ac25969)

**The fields should look in downloads (but normally only one of the two new fields is shown at a time):**
<img width="403" height="206" alt="Image" src="https://github.com/user-attachments/assets/d1b2b4d5-0bf3-4e74-9e92-92f41c8fe002" />

## Design file & designer

@Jerryescandon 

## ✅ Acceptance Criteria

- [ ] Two computed fields removed from mutation inputs
- [ ] Display in Government employee information section of User Profile and Candidate-application (possibly we need to add more info to snapshot?)
- [ ] Display in Government employee information section of User Profile and Application downloads
- [ ] Conditional show/hide for the 2 new fields based on employmentType

gobyrne:
this was broken off of #12140

PR #15137, opened 2025-11-20
[Feature] Retire government employee section

### LINKED ISSUE #12140: ✨ Retire the Government Employee section of the applicant profile
## ✨ Feature

The "government employee information" section of the applicant profile is outdated. It used to be the only place you could indicate that you are a government employee and add your government email, but now we want users to add government experience that directly in their experiences and the email is moved to account settings.

This feature is about retiring the government experience section EXCEPT for the `priority status` and `priority number`. Those absolutely need to get encountered during the application process and we don't collect them anywhere else for now.

Instead of entirely disappearing this section though, we want to put in an interim read-only view of it. 

To preserve the priority status inputs, we will break those out into their own section. 

### Changes to graphql schema

@petertgiles put this together. We can remove a number of mutations that are no longer needed.

| schema field           | in query | in mutation | database column            | recommendation                     |
|------------------------|----------|-------------|----------------------------|------------------------------------|
| isGovEmployee          | Yes      | Yes         | computed_is_gov_employee   | remove from mutation               |
| workEmail              | Yes      | Yes         | work_email                 | remove from mutation               |
| workEmailVerifiedAt    | Yes      | No          | work_email_verified_at     | -                                  |
| govEmployeeType        | Yes      | Yes         | computed_gov_employee_type | remove from mutation               |
| currentClassification  | Yes      | Yes         | computed_classification    | remove from mutation               |
| department             | Yes      | Yes         | computed_department        | remove from mutation               |
| hasPriorityEntitlement | Yes      | Yes         | has_priority_entitlement   | -                                  |
| priorityNumber         | Yes      | Yes         | priority_number            | -                                  |

### Breaking out the priority entitlement info and leaving read-only Government employee info

This breaks the current `Government employee information` section in two, BUT we don't think we need to set up a new query because one of the sections is read-only, so it makes it easier for them to share the same one.

We're also adding a new `well` to describe the change that's happening to users and where they can update the info.

Image

## 🎨 Design

Designer: mostly @JoshBeveridge 

### Backend changes
  - [ ] Several mutations input fields removed (can't edit the gov employee info directly anymore)

### Personal information page
- [ ] New priority entitlement section (uses existing query for government employee information)
  - [ ] Move priority status and priority number inputs here
- [ ] Government employee information now read-only
  - [ ] Section no longer editable (no "update" button)
  - [ ] New alert "this information is going away"
  - [ ] Section shows only computed values

> [!NOTE]
>  No changes to admin views or downloads are in scope for this issue

yonikid15:
🤖 Resolves #12140 

## 👋 Introduction

- Adds new priority entitlements section to profile page and application profile step
- Changes government information section to read-only
- Doesn't make any changes to the admin side or downloads (should be handled in #15045)

## 🕵️ Details

The issue asks to remove fields from their mutations that are no longer needed, but after checking again I wasn't sure which inputs no longer need them.

- CreateUserInput
- UpdateUserAsAdminInput 
- UpdateUserAsUserInput **(Update: removed from here)**

## 🧪 Testing

1. Build app
2. Login as anyone
3. Go to profile information page
4. Ensure everything matches the design and works correctly
5. Create an application and go to the second step (review profile)
6. Ensure everything matches the design and works correctly

## 📸 Screenshot

image

image

### DIFF
--- a/api/app/Builders/PoolCandidateBuilder.php
+++ b/api/app/Builders/PoolCandidateBuilder.php
@@ -13,6 +13,7 @@ use App\Enums\PriorityWeight;
 use App\Enums\PublishingGroup;
 use App\Models\Skill;
 use App\Models\User;
+use Database\Helpers\TeamHelpers as HelpersTeamHelpers;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Query\Expression;
 use Illuminate\Support\Carbon;
@@ -703,23 +704,6 @@ class PoolCandidateBuilder extends Builder
         return $this->where('id', null);
     }
 
-    /**
-     * Get the team ID for a team based permission where
-     * the user has that permission.
-     */
-    public function getTeamIdsForPermission(User $user, string $permission)
-    {
-        return DB::table('role_user')
-            ->join('roles', 'roles.id', '=', 'role_user.role_id')
-            ->join('permission_role', 'roles.id', '=', 'permission_role.role_id')
-            ->join('permissions', 'permission_role.permission_id', '=', 'permissions.id')
-            ->where('role_user.user_id', $user->id)
-            ->where('permissions.name', $permission)
-            ->pluck('role_user.team_id')
-            ->unique()
-            ->toArray();
-    }
-
     // main authorization scope for viewing PoolCandidateAdminView
     public function whereAuthorizedToViewPoolCandidateAdminView(): self
     {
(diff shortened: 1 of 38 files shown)


yonikid15:
> Hi Yoni 😁 I noticed this has changed how government employee information appears in the admin user table view
>

Nice catch @bella77-69 ! Should be fixed in [Fix admin view](https://github.com/GCTC-NTGC/gc-digital-talent/pull/15137/commits/77a28f9e0989a6108060e345b1718f4ebd490a63)

brindasasi (changes requested):
Confirmed #2 exists in main as well.. will create a  sep issue for that  .. but the #1 should be fixed in under this issue

yonikid15:
> Hey Yoni, Sorry I find one more issue on this flow.
> 
> 1. Work experience section in Personal Information page doesn't get refreshed with latest values when you navigate to that.
>    It only updates when you click refresh
> 
> Here in the video , I update a gov exp to non-gov exp and navigate back to Personal Information , still holds the old value until I click page refresh explicitly
> 

Okay I changed the request policy to cache and network which should now fetch the updated data. Changed in [Change request policy to cache and network](https://github.com/GCTC-NTGC/gc-digital-talent/pull/15137/commits/df6ea8a88664d3d294db43ec972e7097cba3c286)

brindasasi (approved):
LGTM!

PR #15149, opened 2025-11-20
[Bug] Allow removing work email

### LINKED ISSUE #14953: 🐛 UAT | Government work email cannot be removed from Applicant's profile even if Government work experience is not added

Description: An existing user, who has removed the government of canada experience or having non gov. of canada experience, that user cannot remove government of canada email from Account Settings

Expected outcome is - When the user is not government of canada employee then the user should have an opportunity to remove the government of canada email from Account Settings.

### 🕵️ Details

Environment - UAT

https://www.figma.com/design/AQbFeF2lm4gZZJOwxOJTN2/Account-settings--All-users-?node-id=792-8660&m=dev

### 📋 Steps to Reproduce

Pre-requisite: Choose the user who had verified Gov. work email

1. Access the GC Digital portal with the same user
2. Make sure to remove the government of Canada work experience or the government of Canada work experience is not added
3. Navigate to Account Settings to remove the gov. of Canada work email 
4. Observe the work email section still shows the old work email.

### 📸 Screenshot

Image

Image

Show the mandatory field error - 
Image

### 🙋‍♀️ Proposed Solution

- When the user is no more Government of Canada employee, that user should have an opportunity to remove the gov of canada work email
- Some documentation or information should be displayed when user tries to remove the gov. of canada work email
- The other solution can be, the Government of canada work email field should not be mandatory for the user to either fill it or remove it.

- Probably a new API mutation will be needed to clear the work email without triggering a verification email or using UpdateUserAsUser.

### ✅ Acceptance Criteria

- [ ] Users can remove a work email
  - [ ] verification status doesn't matter
  - [ ] has to click through an "are you sure" dialog
- [ ] The link to trigger the verification dialog changes for different states:
    -   Verify a GC work email (when null)
    -   Update work email (when verified and different from contact email)
    -   Verify a different work email (when verified and same as contact email)
    -   Re-verify work email (when unverified and non-null)
    

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

vd1992:
🤖 Resolves #14953

## 👋 Introduction

Adds the ability for a user to wipe their work email information. 

## 🧪 Testing

1. Navigate to `/en/applicant/settings`
2. Test varying state of the work email verification dialog's trigger 
3. Test remove mutation

## 📸 Screenshot

image

image

image

### DIFF
--- a/apps/playwright/tests/applicant/applicant-settings.spec.ts
+++ b/apps/playwright/tests/applicant/applicant-settings.spec.ts
@@ -1,0 +1,62 @@
+import { nowUTCDateTime } from "@gc-digital-talent/date-helpers";
+
+import { test, expect } from "~/fixtures";
+import { loginBySub } from "~/utils/auth";
+import graphql from "~/utils/graphql";
+import { generateUniqueTestId } from "~/utils/id";
+import { createUserWithRoles } from "~/utils/user";
+
+test.describe("Applicant settings page", () => {
+  let uniqueTestId: string;
+  let sub: string;
+
+  test.beforeAll(async () => {
+    uniqueTestId = generateUniqueTestId();
+    sub = `playwright.sub.${uniqueTestId}`;
+    const adminCtx = await graphql.newContext();
+
+    await createUserWithRoles(adminCtx, {
+      user: {
+        email: `${sub}@example.org`,
+        sub,
+        isGovEmployee: true,
+        workEmail: `${sub}@gc.ca`,
+        workEmailVerifiedAt: nowUTCDateTime(),
+      },
+      roles: ["guest", "base_user", "applicant"],
+    });
+  });
+
+  test("Work email removal", async ({ appPage }) => {
+    await loginBySub(appPage.page, sub);
+    await appPage.page.goto("/en/applicant/settings");
+    await appPage.waitForGraphqlResponse("AccountSettings");
+
+    // work email verified with correct verified message
+    await expect(
+      appPage.page
+        .locator("#personal-info div")
+        .filter({ hasText: "Government of Canada work" })
+        .nth(1)
+        .getByLabel("verified"),
+    ).toBeVisible();
+    await expect(
+      appPage.page.getByRole("button", { name: "Update work email" }),
+    ).toBeVisible();
+
+    // carry removal out
+    await appPage.page.getByRole("button", { name: "Remove" }).click();
+    await appPage.page
+      .getByRole("button", { name: "Remove work email" })
+      .click();
+    await appPage.waitForGraphqlResponse("RemoveUserWorkEmail");
+
+    // check changes
+    await expect(
+      appPage.page.getByText("No work email provided"),
+    ).toBeVisible();
+    await expect(
+      appPage.page.getByRole("button", { name: "Verify a GC work email" }),
+    ).toBeVisible();
+  });
+});


api/app/Policies/UserPolicy.php
```diff
@@ -205,6 +205,16 @@ public function restore(User $user, User $model)
...
 
+    /**
+     * Determine whether the user can wipe work email information.
+     *
+     * @return \Illuminate\Auth\Access\Response|bool
+     */
+    public function wipeWorkEmailInformation(User $user, User $model)
+    {
+        return $user->isAbleTo('update-own-user') && $user->id === $model->id;
+    }
```
vd1992: Mutation can easily be used by other users like say platform admins in the future, just need to update this

apps/web/src/components/WorkEmailCard.tsx/WipeWorkEmailDialog.tsx
```diff
@@ -0,0 +1,119 @@
...
+    await executeWipeWorkInformationMutation({ id })
+      .then((result) => {
+        if (result.data?.wipeUserWorkEmail) {
+          toast.success(
+            intl.formatMessage({
+              defaultMessage: "Work email deleted",
+              id: "3bVNFv",
+              description: "Successfully cleared work email information",
+            }),
+          );
```
esizer: Sort of nit-pciky but, we use the term "Remove" everywhere else so maybe we shiuld replace "deleted" in this and the error message with "removed".

Also, I tihnk we use periods in our toasts.

@NienkeBr Do you have any strong opinions about this?

apps/web/src/components/WorkEmailCard.tsx/WipeWorkEmailDialog.tsx
```diff
@@ -0,0 +1,119 @@
...
+        } else {
+          void handleError();
+        }
+      })
+      .catch(handleError);
+  };
+  const handleWipe = async () => {
+    await wipeWorkInformation().then(() => {
+      setIsOpen(false);
+    });
```
esizer: Hmm, usually when there is an error we leave the dialog open for easy retries. I think this would lead to it closing on error. Though, I dont have that strong of a feeling. 

However, we could just use `.finally()` so we don't need two functions to handle a single promise. If we want to leave the dialog open on error, the `setIsOpen` could jsut be moved to the success case.

esizer (changes requested):
Works well, just some minor suggestions.

Also, I'm not sure how I feel about the term "wipe". Maybe we keep it consistent and just use "remove" since that is the term we use in the UI.

api/app/Policies/UserPolicy.php
vd1992: Yeah I think it should be distinct, it is a very specific and special operation

apps/web/src/components/WorkEmailCard.tsx/WipeWorkEmailDialog.tsx
vd1992: Sure, why not. I relocated the close. https://github.com/GCTC-NTGC/gc-digital-talent/pull/15149/commits/c9be43224fe65ee18f336bb4d6935b431058b615

Renamed files too in  https://github.com/GCTC-NTGC/gc-digital-talent/pull/15149/commits/d22bdda71dfa2c5a8f0cb982e426accfaf2b3fc0

apps/web/src/components/WorkEmailCard.tsx/RemoveWorkEmailDialog.tsx
```diff
@@ -0,0 +1,119 @@
...
+    await executeWipeWorkInformationMutation({ id })
+      .then((result) => {
+        if (result.data?.wipeUserWorkEmail) {
+          toast.success(
+            intl.formatMessage({
+              defaultMessage: "Work email deleted",
+              id: "3bVNFv",
+              description: "Successfully cleared work email information",
+            }),
+          );
```
vd1992: Aye aye, see https://github.com/GCTC-NTGC/gc-digital-talent/pull/15149/commits/dd641ed11f5106938afed5a01577cd227d5b36a2

vd1992:
> Sorry, I forgot. We probably want an e2e test for this.

Should that be a separate issue or in this? We've been making issues to add Playwright testing
https://github.com/GCTC-NTGC/gc-digital-talent/issues?q=is%3Aissue%20state%3Aopen%20playwright

PR #15150, opened 2025-11-21
[Feat] Root error boundary

### LINKED ISSUE #14563: 🐛 No top level error boundary
## 🐛 Bug

We have an error boundary but it is one step below to root route. This is leading to users seeing raw errors

## 🦋 Expected Behaviour

Users should always see a proper error page with friendly messaging and not the raw dev errors.

## 🕵️ Details

This happens usually after a deploy since the `MainLayout` file hash changes and the cache might be looking for the incorrect file on the first load.

## 📋 Steps to Reproduce

I'm not sure it is easily reproducable. You'd need to check UAT or Prod right after a deploy with an existing cache.

You can force it by putting a blatant error in `MainLayout`

## 📸 Screenshot

Image

## 🙋‍♀️ Proposed Solution

We need to add another error boundary at the top level but it can;t contain the main nav since the layout that contains the main nav is what has the error.

## ✅ Acceptance Criteria

- [ ] Basic error boundary added at the root route
- [ ] Nice error page when an error occurs in `MainLayout`

### LINKED ISSUE #15009: 🐛 UAT - unexpected application error

I've observed this error multiple times 
Image

Image
Image

### 🕵️ Details

Let me know if there is any other details needed to diagnose this issue. 

### 📋 Steps to Reproduce

Not sure how to reproduce
I believe this is happening when the tab is left open and possibly a new build has been deployed .. 

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Figure out how to reliably replicate this error
- [ ] Display an error message to the user to refresh their browser
   - [ ] "We’ve made some updates to the platform. Please refresh the page to continue."
- [ ] If possible, an automated test

esizer:
🤖 Resolves #14563, resolves #15009 

## 👋 Introduction

This adds a root level error boundary along with a helpful message to attempt to reload the page.

## 🕵️ Details

This message it as the very top of our application. It is meant to handle extreme cases like the application failing to load becuase of an error.

The scenario in which this is most likely to ocurr is when a user has a cached page in their browser during a deploy and then attempts to navigate to it after a deploy. The reason this occurs is because in order to bust cache, we add a hash to the file name. The hash changes on deploy but the users browser is looking for the old file that no longer exists.

We added a helpful error message to tell users to most likely way of resolving the issue. 

> [!IMPORTANT]
> Since this is a top level error, the action buttons to navigate to home or support form have been removed since those will also not work. The only recourse here is to reload the page and hope it works :cry: 
>
> Because of this, we may want to host some external suppport form (maybe on freshdesk?) Or some better option that does not require the user to access our site. 

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Mangle the `MainLayout` file in `./apps/web/dist` (i just change one char in hash)
3. Try to load the app
4. Confirm you see the pretty error message and not the ugly dev errror
5. Restore the main layour file name to its original
6. Click the reload link on the page
7. Confirm everything goes back to normal

## 📸 Screenshot

localhost_8000_en_admin_candidates_541fc62a-6328-425d-89c8-2982998b9dfe_application

### DIFF
--- a/apps/web/src/components/Layout/RouteErrorBoundary/RootErrorBoundary.tsx
+++ b/apps/web/src/components/Layout/RouteErrorBoundary/RootErrorBoundary.tsx
@@ -1,3 +1,5 @@
+/* eslint-disable react/forbid-elements */
+// NOTE: `a` needed here to avoid styling, etc
 import { useIntl } from "react-intl";
 import { useLocation } from "react-router";
 import { ReactNode } from "react";


apps/web/src/components/Layout/RouteErrorBoundary/RootErrorBoundary.tsx
```diff
@@ -0,0 +1,135 @@
...
+export const RouteErrorBoundary = () => {
+  const intl = useIntl();
+  // eslint-disable-next-line no-restricted-syntax
+  const { locale } = useLocale();
+  const { mode } = useTheme();
+  const location = useLocation();
+  const error = useErrorMessages();
+  const logger = useLogger();
+
+  logger.notice(
```
petertgiles: I'd suggest bumping this to "error" level.  I think we really want to know if this is hit, right?  Also, can you update the message to specify that it's the RootErrorBoundary?

PR #15178, opened 2025-11-26
[Fix] Import resolver config

### LINKED ISSUE #15177: 🛠️ eslint not resolving Typescript path aliases
### 🛠️ Platform

Our path alias for `~/*` seems to not always be resolved by eslint in editor even though they work.

### 🕵️ Details

_No response_

### 🙋‍♀️ Proposed Solution

Fix the config for `eslint-plugin-import` so it can resolve the typescript path aliases

### ✅ Acceptance Criteria

- [ ] Eslint knows the path aliases when editing

esizer:
🤖 Resolves #15177 

## 👋 Introduction

Fixes an issue where path aliases were not properly being resolved by eslint.

## 🧪 Testing

1. Maybe need to run `pnpm i` and restart both Typescript and eslint in your editor
2. Confirm lint check still passes `CI=true pnpm lint --force`
3. Open a file that contains an import from an aliased path
4. Confirm no eslint errors appear

### DIFF
--- a/packages/eslint-config-custom/index.mjs
+++ b/packages/eslint-config-custom/index.mjs
@@ -54,7 +54,11 @@ export default tseslint.config(
         "@typescript-eslint/parser": [".ts", ".tsx"],
       },
       "import/resolver": {
-        typescript: {},
+        typescript: {
+          alwaysTryTypes: true,
+          project: ["./apps/*/tsconfig.json", "./packages/*/tsconfig.json"],
+        },
+        node: true,
       },
     },
     rules: {


esizer:
> Different problem?

Oof, okay. Unfortunatley I do not see that issue :cry: Though, seems like you are still having the resolver fail. I added a new step that is really 3 steps to the testing. If you did not do those can you try again and:

1. Run `pnpm i` (shouldnt be needed but just in case!)
2. Restart typescript server
3. Restart eslint server

petertgiles (approved):
Turns out I did have this problem.  😱 

Thanks for the fix.

PR #15207, opened 2025-11-27
[Bug] Fix keyboard navigation on nav menu

### LINKED ISSUE #11986: ⌨ Keyboard navigation of new menu is confusing

yonikid15:
🤖 Resolves #11986 

## 👋 Introduction

- Moves all links into one list to allow arrow navigation
- This doesn't include the notification menu since it's not a link. It seems like buttons can't be navigated to using the arrow key navigation, within the [list](https://www.radix-ui.com/primitives/docs/components/navigation-menu#list).

~~NOTE: Still fixing the role switcher sub menu not lining up with the trigger link~~
UPDATE: I got the role switcher sub menu to line up correctly

## 🧪 Testing

1. Build app
2. Tab to nav menu. Once focused, try using navigation arrows to flip through the menu

### DIFF
--- a/api/app/Enums/GovEmployeeType.php
+++ b/api/app/Enums/GovEmployeeType.php
@@ -12,6 +12,7 @@ enum GovEmployeeType
     case CASUAL;
     case TERM;
     case INDETERMINATE;
+    case CONTRACTOR;
 
     public static function getLangFilename(): string
     {
--- a/api/lang/en/gov_employee_type.php
+++ b/api/lang/en/gov_employee_type.php
@@ -5,4 +5,5 @@ return [
     'casual' => 'Casual',
     'term' => 'Term',
     'indeterminate' => 'Indeterminate',
+    'contractor' => 'Contractor',
 ];
--- a/api/lang/fr/gov_employee_type.php
+++ b/api/lang/fr/gov_employee_type.php
@@ -5,4 +5,5 @@ return [
     'casual' => 'Occasionnel',
     'term' => 'Durée déterminée',
     'indeterminate' => 'Durée indéterminée',
+    'contractor' => 'Entrepreneur ou entrepreneuse',
 ];
--- a/api/storage/app/lighthouse-schema.graphql
+++ b/api/storage/app/lighthouse-schema.graphql
@@ -3912,6 +3912,7 @@ enum GovEmployeeType {
   CASUAL
   TERM
   INDETERMINATE
+  CONTRACTOR
 }
 
 enum GovPositionType {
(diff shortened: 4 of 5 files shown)


yonikid15:
> Nice @yonikid15 learning question: For the notification menu could we use `NavMenu.Trigger`? for example wrapping the `NotificationDialog` with `NavMenu.Trigger` and hide the chevron?? 🤔

Maybe! I haven't tried that. It could be a work around to trick the navigation list. I'll try it and let you know :+1:

bella77-69 (commented):
I am not sure what the right way forward is with the notification button 🤷‍♀️There is still a gap, the arrow keys work on the menu links but the notification button is still not part of the navigation flow. @esizer thoughts?

bella77-69 (commented):
Question..... when I open a dropdown item with the spacebar and navigate through it I can use the escape key to exit out of the dropdown, but if I use the Tab to leave I need to press it twice to get back into the navbar, and the dropdown stays open until I open another dropdown. 

!! Ignore this question 👆.... I think this is Radix behaviour and the tab key will not close the dropdown (just doing some reading 🤓)

bella77-69:
Nice work Yoni 😁

I’m happy to approve this PR. It resolves the accessibility issue by consolidating the main navigation into a single nav list, which allows users to move across the entire navbar using the arrow keys (with the exception of the notification icon).

I did notice a couple of bugs that appear to be related to Radix dropdown behavior. They don’t block this PR, but it may be worth tracking them in a separate issue.

PR #15255, opened 2025-12-03
[Debt] Update most CSV downloads to Excel

### LINKED ISSUE #15018: ✨ Change all CSV download to excel format
### Details

Following the initial change of format from CSV to excel of the users download (https://github.com/GCTC-NTGC/gc-digital-talent/issues/13626) , we now want to change all CSVs to excel files.

### Design file & designer

_No response_

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

Only the file type will change, no change in content or styling for this issue.

### ✅ Acceptance Criteria

- [ ] All CSV downloads are now excel downloads
- [ ] Copy download buttons have been updated to reflect this change

esizer:
🤖 Resolves #15018 

## 👋 Introduction

This migrates all remaining server side generated CSV files to Excel files.

## 🕵️ Details

There are still some references to CSVs

1. I left the old class `CsvGenerator`. I see no issue keeping it for now in case we decide we want a CSV... Let me know if you think this was a mistake
2. The skills table uses a lib on the frontend so its not as easy to remove so I left it

### Skills table

Honestly, it would be really nice to make this server generated. I know for the ones that are queued, we need notifications. But, this could be a pretty static file. I wonder if it makes sense to just generate a file on `Skill` or `SkillFamily` change and just link to some static file so we can drop the dependency?

I'd be willing to do this in this PR if we don't think it pushes the diff too far. :woman_shrugging: 

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Login as `admin@test.com`
3. Go and download excels for:
  a. Pool candidates
  b. Community talent
  c. Nominations
4. Confirm each file downloads as an excel and still contains the proper content

### DIFF
--- a/apps/web/src/lang/fr.json
+++ b/apps/web/src/lang/fr.json
@@ -4963,6 +4963,10 @@
     "defaultMessage": "Profil linguistique",
     "description": "Title of the Language Information link section"
   },
+  "Ji6+E5": {
+    "defaultMessage": "Télécharger Excel sans évaluations",
+    "description": "Button label for download candidate excel without ROD data."
+  },
   "JjTGYe": {
     "defaultMessage": "Vous êtes sur le point de modifier la date d'expiration pour cet utilisateur :",
     "description": "First section of text on the change candidate expiry date dialog"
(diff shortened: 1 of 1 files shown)


apps/web/src/pages/Users/IndexUserPage/components/UserTable.tsx
petertgiles: Seems like you didn't touch the user generator much here, but I'm getting an empty file out the community talent table.  If it's pre-existing I can open a separate issue.

Image

api/app/Generators/NominationsExcelGenerator.php
petertgiles: Wouldn't be surprised if this was pre-existing as well but the talent nomination extract had only one row instead of two.

Image

PR #15366, opened 2025-12-15
[Feat] `PoolCandidate` custom events logging

### LINKED ISSUE #15359: ✨ Custom `PoolCandidate` activlity log events
### Details

Add some custom events for the different mutations for a `PoolCandidate` to the activity log. Events to be logged:

 - `submitted`
 - `disqualified`
 - `qualified`
 - `removed`
 - `placed`

### Design file & designer

https://www.figma.com/design/fn6LXNySdE5irUbvky9WRM/Activity-log---Pools--Communities-?node-id=249-10055&m=dev

### Proposed Implementation

In each mutation for a `PoolCandidate` log a custom event instead of the typical one.

```php
// PoolCandidate.php
public function remove()
{
  $this->disableLogging();

  // ...mutation code

 $causer = Auth::user();

  activity()
   ->causedBy($causer)
   ->performedOn($this)
   ->log('removed');
}
```

REF: 
 - https://spatie.be/docs/laravel-activitylog/v4/advanced-usage/logging-model-events#content-disabling-logging-on-demand
 - https://spatie.be/docs/laravel-activitylog/v4/basic-usage/logging-activity

### ✅ Acceptance Criteria

- [ ] `PoolCandidate` model logs mutations as custom events
- [ ] Original event is no longer logged

esizer:
🤖 Resolves #15359 

## 👋 Introduction

Adds logging for custom events on the pool candidate model.

## 🕵️ Details

This is meant to support the activity log to get more information about the specific events occuring for candidates in a pool.

## :question: Question

We are logging all model events (`created`, `updated`, `deleted`). These include, update the submitted steps at each time. Should we be keeping these logs or removing all events in favour of specific ones?

We should consider, that we may not want these events pre-submission appearing in the activity log so if we leave them around, they may need to be filtered out later.

### Answer

> > We could stop logging the draft events and only start logging the events on and after submission
> 
> This is my preference

## 🧪 Testing

1. Login as any user
2. Submit an application
3. Login as `admin@test.com`
4. Navigate to the admin view for the application submitting in step 2
5. Complete each mutation type on a pool candidate
     - Disqualify
     - Qualify
     - Place
     - Remove
     - Reinstate
6. Confirm that each event is logged with the appropriate event type

### DIFF
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -489,7 +489,7 @@ type Activity @model(class: "\\App\\Models\\Activity") {
   subjectId: UUID @rename(attribute: "subject_id")
   properties: ActivityProperties
   logName: String @rename(attribute: "log_name")
-  event: ActivityEvent!
+  event: ActivityEvent
   description: String
   createdAt: DateTime @rename(attribute: "created_at")
   updatedAt: DateTime @rename(attribute: "updated_at")
--- a/api/storage/app/lighthouse-schema.graphql
+++ b/api/storage/app/lighthouse-schema.graphql
@@ -1043,7 +1043,7 @@ type Activity {
   subjectId: UUID
   properties: ActivityProperties
   logName: String
-  event: ActivityEvent!
+  event: ActivityEvent
   description: String
   createdAt: DateTime
   updatedAt: DateTime


api/app/Enums/PoolCandidateEvent.php
```diff
@@ -0,0 +1,13 @@
...
+namespace App\Enums;
+
+enum PoolCandidateEvent: string
+{
+    case SUBMITTED = 'submitted';
+    case QUALIFIED = 'qualified';
+    case DISQUALIFIED = 'disqualified';
+    case PLACED = 'placed';
+    case REMOVED = 'removed';
+    case REINSTANTED = 'reinstated';
```
vd1992: REINSTATED*

api/app/Enums/PoolCandidateEvent.php
vd1992: Should you also touch the below files and add events for mutation classes 

`QualifyAndPlaceCandidate`
`RevertFinalDecision`
`RevertPlaceCandidate`

api/tests/Feature/ActivityLog/PoolCandidateEventTest.php
```diff
@@ -0,0 +1,143 @@
...
+        $activity = Activity::where('event', PoolCandidateEvent::QUALIFIED->value)->latest()->first();
+
+        $properties = $activity->properties;
+
+        $this->assertEquals(['expiry_date' => $date->format('Y-m-d H:i:s')], $properties['attributes']);
+    }
+
+    public function testDisqualifiedEvent()
+    {
+        $this->application->disqualify(PoolCandidateStatus::SCREENED_OUT_APPLICATION->name);
```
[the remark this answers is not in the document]
vd1992: Thoughts on this, @petertgiles

vd1992 (changes requested):
Just a read through first

api/app/Enums/PoolCandidateEvent.php
esizer: Fixed in [fix typo](https://github.com/GCTC-NTGC/gc-digital-talent/pull/15366/commits/a77aea7e55f9d39fbcaa7059e4453f2846964de4)

api/tests/Feature/ActivityLog/PoolCandidateEventTest.php
```diff
@@ -0,0 +1,143 @@
...
+use Tests\TestCase;
+
+class PoolCandidateEventTest extends TestCase
+{
+    use RefreshDatabase;
+
+    protected PoolCandidate $application;
+
+    protected User $user;
+
```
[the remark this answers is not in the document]
esizer: No, copy and paste error :stuck_out_tongue: Removed in [remove unused property](https://github.com/GCTC-NTGC/gc-digital-talent/pull/15366/commits/017fac8e98b9f46ae149b55de90b5ca04b2f1bc4)

api/app/Models/PoolCandidate.php
```diff
@@ -80,7 +80,6 @@ class PoolCandidate extends Model
 {
     use EnrichedNotifiable;
     use HasFactory;
-    use LogsActivity;
```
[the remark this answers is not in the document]
vd1992: Yeah, I think refining the pool candidate logging and what to include/exclude is still open-ended. Should be spun off into its own thing

api/tests/Feature/ActivityLog/PoolCandidateActivityEventTest.php
```diff
@@ -0,0 +1,143 @@
...
+        $activity = Activity::where('event', PoolCandidateEvent::QUALIFIED->value)->latest()->first();
+
+        $properties = $activity->properties;
+
+        $this->assertEquals(['expiry_date' => $date->format('Y-m-d H:i:s')], $properties['attributes']);
+    }
+
+    public function testDisqualifiedEvent()
+    {
+        $this->application->disqualify(PoolCandidateStatus::SCREENED_OUT_APPLICATION->name);
```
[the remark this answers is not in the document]
vd1992: >This actually is a "reason". It is DisqualificaitonReason enum that will be stored there eventually once we remove pool_candidate_status. It is just getting stored there for now which does seem odd

Yeah, my concern is storing a value there, intentionally or unintentionally, that does not the match the expected GraphQL return type. Say you call `disqualify($reason)` with `$reason = STATUS_WITH_TYPO`
That candidate model will crash anytime you call from API `poolCandidate.poolCandidateStatus`

That is why I suggested the extra mapping step, which would only save values from `PoolCandidateStatus::cases()` and fail to assign otherwise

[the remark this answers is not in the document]
petertgiles: Maybe adding a check for reason in DisqualifyCandidateValidator?

api/app/Enums/ActivityEvent.php
vd1992: I like having one event enum that encompasses all the possible values for the column. 

I think we should have a prefix or something to indicate model specific events, otherwise it could get confusing especially if we expand our activity log more. For example, "disqualified" you might realize it has to do with a pool candidate or may think it has to do with nominations.

I'd suggested `poolCandidate-` in front of every pool candidate specific event. That also has the added advantage of being able to fetch groups of events through matching `where('poolCandidate-*')`

esizer: Hmmm, I guess it depends on if there is overlap between the events. Like maybe it eventually would make sense that something else could be `submitted`, `reverted` or `removed`? :thinking: 

I think getting based on model might make sense to do:

`Activity::where('subject_type', PoolCandidate::model)` instead of by the event type.

api/tests/Feature/ActivityLog/PoolCandidateActivityEventTest.php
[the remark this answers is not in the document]
esizer: Just to confirm, this is future work we are talking about here related to the status migration right? Nothing for me to action? :sweat_smile:

api/tests/Feature/ActivityLogTest.php
```diff
@@ -223,8 +225,10 @@ public function testActivityLogByUser(): void
             ]
         )->assertSuccessful();
 
+        Log::debug(Activity::all());
+
```
vd1992: Something to remove

api/tests/Feature/ActivityLogTest.php
```diff
@@ -97,106 +100,105 @@ public function testLoggingMultipleEvents(): void
...
+        assertEquals($testClassiciation->id, $activityCreatePool->properties['attributes']['classification_id']);
+        assertEquals($this->community->id, $activityCreatePool->properties['attributes']['community_id']);
+        assertEquals($this->adminUser->id, $activityCreatePool->properties['attributes']['user_id']);
     }
 
     // test activity log actions undertaken by a specific user
     public function testActivityLogByUser(): void
     {
-        $testPool = Pool::factory()->published()->create();
+        $testClassiciation = Classification::factory()->create();
```
vd1992: testClassification *

api/tests/Feature/ActivityLogTest.php
```diff
@@ -97,106 +100,105 @@ public function testLoggingMultipleEvents(): void
...
-            /** @lang GraphQL */
-            '
-                mutation createApplication($userId: ID!, $poolId: ID!){
-                    createApplication(userId: $userId, poolId: $poolId) {
-                        id
-                    }
-                    }
-                ',
+
+        $testClassiciation = Classification::factory()->create();
```
vd1992: testClassification *

api/app/Models/PoolCandidate.php
vd1992: For reference, all the activity log records generated during an application
Count 9. Could vary for example if a process had no questions

Image

vd1992: Creating assessment results also creates activity records for pool candidate, those records consist of just the updated at change

Image

vd1992 (changes requested):
Okay, some things to fix as well as some observations about how there is some copious record creation

api/app/Enums/ActivityEvent.php
vd1992: I think this should still be considered, especially if the event enum could grow a bunch

api/app/Models/PoolCandidate.php
esizer: Yes, that will be going away though so should no longer touch the candidate.

api/graphql/schema.graphql
```diff
@@ -489,7 +489,7 @@ type Activity @model(class: "\\App\\Models\\Activity") {
   subjectId: UUID @rename(attribute: "subject_id")
   properties: ActivityProperties
   logName: String @rename(attribute: "log_name")
-  event: String
+  event: ActivityEvent!
```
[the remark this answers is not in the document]
esizer: Updated in [set event to nullable](https://github.com/GCTC-NTGC/gc-digital-talent/pull/15366/commits/35249b3887686b76ad993083fca9f2366283293f)

api/app/Enums/ActivityEvent.php
vd1992: I thought it might be good for a growing enum `ActivityEvent` where there will be groupings contained within, to make it more obvious. That way you can interact with just the event column without referring to another column. Example
log grouping -> manipulate where `event = poolCandidate-submitted`
This is more useful when tracking `submitted` on multiple models

Bu yeah, I'm open to changing my mind if other people don't see it as something to worry about. I'm don't feel that strongly about it

vd1992 (approved):
👍

PR #15466, opened 2025-12-29
[Feat] Aggregate pool activity

### LINKED ISSUE #15360: ✨ Aggregate Pool activity log to contain related models
### Details

Add related models into the response for a pools acitivty log.

> [!IMPORTANT]
> For `PoolCandidate` we only want to show custom events. You will need to filter out the base events of `created`, `updated` and `deleted`. 

### Design file & designer

_No response_

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

Update the scope to include the other subjects.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/dd8fda6665e8c92bdbea8f0188681b637619df24/api/app/Models/Activity.php#L27-L30

Include:

 - `PoolCandidate`
 - `PoolSkill`
 - `AssessmentStep`
 - ...?

### ✅ Acceptance Criteria

- [ ] Related models added to return of pool activity log

esizer:
🤖 Resolves #15360 

## 👋 Introduction

Updates the pool activity log to aggregate all the data from related models.

## 🕵️ Details

This adds a new log with a trait to use custom logs.

### Usage

```php
use App\Traits\LogsCustomActivity;

class PoolCandidate extends Model
{
    use LogsCustomActivity;

    // NOTE: If not defined, it will fallback to default
    protected $customLogName = ActivityLog::PROCESS->value;
    
    public function submit()
    {
       //...submit logic

       // This will log activity to the custom process log
       $this->logActivity(AcitivtyEvent::SUBITTED);
       
       // You can override the defined log name
       $this->logName('another-log')
         ->logActivity(...);
    }
    
    // You can add custom values on every log with
    protected function customizeActivityProperties(array &$properties, ActivityEvent $event): void
    {
        $properties['attributes']['user_name'] = $this->user?->fullName ?? null;
        $properties['attributes']['pool_id'] = $this->pool?->id ?? null;
    }

}
```

## 🧪 Testing

> [!TIP]
> When creating a process and assessing the candidate, do a lot of different events like adding/removing skills and assessment steps. Reverting and changing assesssments on candidate for variety. Ideally, look at the events being logged and trigger all of them. 

1. Refresh and build `make refresh-api; pnpm dev:fresh`
2. Login as `admin@test.com`
3. Create and publish a process
4. Logout
5. Login as `applicant@test.com
6. Apply to the process published in step 3
7. Logout
8. Login as `admin@test.com`
9. Make a bunch of different assessment changes to the applicant from step 6
10. Navigate to the activity log `/admin/pools/{poolId}/activity` for the pool publioshed in step 3
11. Confirm it contains all the expected log events

## 📸 Screenshot

swappy-20251229_124157

## :truck: Deployment

Sometime after deployment, we will want to run `artisan app:sync-process-log`


api/app/Console/Commands/SyncProcessLog.php
vd1992: So this duplicates all log records where the subject type is `App\Models\Pool` with the exception of setting `log_name = 'process'` right?
What exactly for, to avoid updating existing records?
Does it even touch related models?

api/app/Models/AssessmentStep.php
```diff
@@ -55,6 +63,11 @@ public function getActivitylogOptions(): LogOptions
             ->dontSubmitEmptyLogs();
     }
 
+    protected function customizeActivityProperties(array &$properties, ActivityEvent $event): void
+    {
+        $properties['attributes']['name'] = $this->name;
+    }
+
```
vd1992: Argument `$event` is hinted in the function but never used. Should it?

api/app/Enums/ActivityLog.php
vd1992: So every activity instance is going to be duplicated? One for each `ActivityLog` enum?

image

api/app/Models/PoolSkill.php
```diff
@@ -33,22 +37,37 @@ class PoolSkill extends Model
...
+
     protected static function boot()
     {
         parent::boot();
 
-        static::created(function (PoolSkill $poolSkill) {
+        $atts = ['pool_id', 'type', 'required_skill_level'];
+
+        static::created(function (PoolSkill $poolSkill) use ($atts) {
             $poolSkill->pool->syncApplicationScreeningStepPoolSkills();
```
vd1992: I think a method for updating is needed too. To add in the `pool_id` field into properties at the minimum so you can query it. 
Pool skills can be edited to changed the required skill level

api/app/Enums/ActivityEvent.php
```diff
@@ -9,12 +9,13 @@ enum ActivityEvent: string
     case UPDATED = 'updated';
     case DELETED = 'deleted';
 
-    // Pool candidate events
     case SUBMITTED = 'submitted';
     case QUALIFIED = 'qualified';
     case DISQUALIFIED = 'disqualified';
     case PLACED = 'placed';
+    case ADDED = 'added';
```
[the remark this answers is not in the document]
vd1992: Applies to other models

image

vd1992 (dismissed):
Okaaay, some thoughts below 👇

api/app/Enums/ActivityLog.php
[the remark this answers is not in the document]
esizer: Yeah, I figured this would be easier than converting all the cases to lower case.

api/app/Models/AssessmentStep.php
esizer: Yeah, it just needs to be there to match the expected function signature so our linters do not complain.

apps/web/src/components/Activity/Item.tsx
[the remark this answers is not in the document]
vd1992: This change doesn't appear to have gotten rid of the console warnings though

image

api/app/Enums/ActivityEvent.php
vd1992: Not sure where or when it should be done, but definitely need to expand this some more. You can update by extending a process's closing date, close it immediately, re-open, archive, and un-archive. Looking at the logging, you cannot tell what I did exactly.

I think updating the closing date is fine, but closing immediately, re-opening, archive, and un-archive need to be recognizable 

image

apps/web/src/components/Activity/Item.tsx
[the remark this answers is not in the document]
esizer: I moved the function call that triggers the warning to avoid it when we know we don't care about them in [move normalization to avoid errant warnings](https://github.com/GCTC-NTGC/gc-digital-talent/pull/15466/commits/d8fd8984e01cb191d88e8a3dfcd19afd4d15026f)

api/app/Console/Commands/SyncProcessLog.php
```diff
@@ -0,0 +1,71 @@
...
+namespace App\Console\Commands;
+
+use App\Models\Pool;
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\DB;
+
+class SyncProcessLog extends Command
+{
+    /**
+     * The name and signature of the console command.
```
petertgiles: Could you please add some comments:
- what this does
- when it should be run
- should it ever be rerun
- can it safely be rerun
- when can it be deleted

api/app/Models/Activity.php
```diff
@@ -31,9 +32,19 @@ protected function event(): Attribute
...
-        return $query->where('subject_type', Pool::class);
+        $poolId = $pool->id;
+        if (! $poolId) {
+            return $query->whereRaw('0 = 1');
+        }
+
+        return $query->where('log_name', ActivityLog::PROCESS->value)
+            ->where(function (Builder $subQuery) use ($poolId) {
+                $subQuery->where('subject_id', $poolId)
+                    ->orWhereJsonContains('properties->attributes->pool_id', $poolId);
```
petertgiles: This worries me a bit.  I feel like we could easily reveal more than we intend.  How many subject types would we need to list if we kept a list here?  I would feel better if we had to opt-in models to this scope.

api/app/Models/Activity.php
```diff
@@ -31,9 +32,19 @@ protected function event(): Attribute
...
     {
-        return $query->where('subject_type', Pool::class);
+        $poolId = $pool->id;
+        if (! $poolId) {
+            return $query->whereRaw('0 = 1');
+        }
+
+        return $query->where('log_name', ActivityLog::PROCESS->value)
+            ->where(function (Builder $subQuery) use ($poolId) {
+                $subQuery->where('subject_id', $poolId)
```
petertgiles: This first clause should probably also have subject_type, right?

api/app/Models/AssessmentStep.php
```diff
@@ -90,6 +141,8 @@ protected static function boot()
     {
         parent::boot();
 
+        $atts = ['pool_id', 'sort_order', 'type', 'title'];
```
petertgiles: Nitpick:
```suggestion
        $loggedAttributes = ['pool_id', 'sort_order', 'type', 'title'];
```

api/app/Models/Activity.php
esizer: Hmm, yeah. That might not be a bad idea. We would need to maintain the list of related subjects but it won't be long so that is probably a good thing to add.

api/app/Console/Commands/SyncProcessLog.php
petertgiles: Looks great!  Thanks.

PR #15472, opened 2025-12-30
[Fix] Add missing `user_id` index on experience tables

### LINKED ISSUE #15424: ♻️ Missing indexes - experience by user_id
### ♻️ Refactor

We've had the pg_stat_statements running for a bit now and I decided to take a look.  It turns out, a lot of db time is being wasted because we're missing user_id indexes on the experience tables.

### 🕵️ Details

Image

Image

### 🙋‍♀️ Proposed Solution

Add an index on user_id in all five experience tables.

### ✅ Acceptance Criteria

- [ ] Running this query does not cause a seq scan:
`select * from "work_experiences" where "work_experiences"."user_id" in ($1) and "work_experiences"."deleted_at" is null`

esizer:
🤖 Resolves #15424

## 👋 Introduction

Adds an index to the `user_id` column on the experience tables to avoid sequence scans.

## 🕵️ Details

```
QUERY PLAN
Index Scan using work_experiences_user_id_index on work_experiences  (cost=0.28..8.30 rows=1 width=544) (actual time=0.017..0.017 rows=1 loops=1)
  Index Cond: (user_id = 'd47419db-52c5-480f-ab91-d4cfd7f58878'::uuid)
  Filter: (deleted_at IS NULL)
Planning Time: 0.308 ms
Execution Time: 0.029 ms
```

## 🧪 Testing

> [!IMPORTANT]
> IKf you have a small table size, a sequence scan might actually be chosen instead of index so you will need to seed quite a few experiences (5000 should be enough).

```sql
EXPLAIN ANALYZE
SELECT * FROM work_experiences
WHERE user_id = '{UUID}'
  AND deleted_at IS NULL;
``` 

1. Run a qurey to an experience
2. Obersve the sequence scan
3. Run migration `make artisan CMD="migrate"`
4. re-run the query from step 1
5. Confirm it now runs an index scan

### DIFF
--- a/api/database/migrations/2025_12_30_161222_add_user_id_index_experience_tables.php
+++ b/api/database/migrations/2025_12_30_161222_add_user_id_index_experience_tables.php
@@ -1,0 +1,60 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('award_experiences', function (Blueprint $table) {
+            $table->index('user_id', 'award_experiences_user_id_index');
+        });
+
+        Schema::table('community_experiences', function (Blueprint $table) {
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
Thanks!  Works great.

PR #15570, opened 2026-01-08
[Bug] Ensure pool community and work stream consistent

### LINKED ISSUE #15415: 🐛 Invalid Community and Workstream combinations can be saved during Process creation

- While creating a process through `createPool` mutation, the default community has been assigned to the workstream selected to the same process.
- Ideally the logic is, the selected Communities' workstream list to be displayed and should get selected and vice versa (Not sure if we have that rule in place, when the workstream got selected, the respective community should get auto selected)
- For example if Access to Information and Policy workstream is selected through mutation, then the process created has been tagged to 'Digital Community' which is by default, Can we have the rule in place that this type of rare combination should not allowed in any case for process creation?

### 🕵️ Details

Env: Local

Note: In Mutation, the default community to be selected as 'Digital' only, but in ideal scenario, it shouldn't allow user to create a process with Community name as 'Digital Community' and work stream as 'Access to Information and Policy' or whatever user selects.
I would be updating the code which mandate the user to enter the community name while creating the pool, but wanted to highlight the mismatch in workstream and community, which leads to incorrect visibility of candidate in 'Talent Requests' 

### 📋 Steps to Reproduce

1. Create a new pool using `createPool` mutation through playwright with 'Access to Information and Policy' workstream - https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/apps/playwright/utils/pools.ts
2. Navigate to the Admin or Community admin portal and verify the process is created
3. Verify the workstream is same that we have used, but the community is assigned as 'Digital Community'

### 📸 Screenshot

Image

Image

### 🙋‍♀️ Proposed Solution

- The likelihood of this issue is really very rare, but if we can place a validation to avoid this mismatch data will be great.
- Through UI, this mismatch will not occur, but by using mutation, it still can be possible

### ✅ Acceptance Criteria

- [ ] A validation between the work stream & community and vice versa (Not on UI but in Mutation)

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

petertgiles:
🤖 Resolves #15415 

## 👋 Introduction

Ensures that a pool's community and work stream are consistent.

## 🕵️ Details

There was some discussion in the original issue about community ID missing from the create pool mutation.  It's there, just not in the CreatePoolMutation input.

### API

1. Log in as admin.
2. Navigate to the edit pool page of a draft pool.
3. The editor will only shows work streams in the community.  Choose one and save.  Verify success.
4. Try the same Graphql request with a work stream in a different community.  Either using Graphiql or Edit-And-Resend.  Verify failure.

### Seeder

1. Nuke all your pools. `truncate table pools cascade`
2. Run the pool seeder a few times. `./artisan db:seed --class=PoolRandomSeeder`
3. Verify you don't get any mismatches.

```sql
select p.community_id , ws.community_id 
from pools p
join work_streams ws on p.work_stream_id = ws.id
where p.community_id <> ws.community_id
```

### DIFF
--- a/apps/playwright/utils/pools.ts
+++ b/apps/playwright/utils/pools.ts
@@ -55,6 +55,9 @@ const Test_CreatePoolMutationDocument = /* GraphQL */ `
         en
         fr
       }
+      community {
+        id
+      }
     }
   }
 `;
@@ -231,10 +234,15 @@ export const createAndPublishPool: GraphQLRequestFunc<
     classificationId,
     departmentId,
   }).then(async (pool) => {
-    let workStream = workStreamId;
-    if (!workStream) {
+    let selectedWorkStreamId = workStreamId;
+    if (!selectedWorkStreamId) {
       const workStreams = await getWorkStreams(ctx, {});
-      workStream = workStreams[0].id;
+      selectedWorkStreamId = workStreams.find(
+        (ws) => ws.community?.id == pool.community?.id,
+      )?.id;
+    }
+    if (!selectedWorkStreamId) {
+      throw new Error("Failed to pick a work stream for the new pool.");
     }
 
     await updatePool(ctx, {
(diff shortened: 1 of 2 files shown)


vd1992 (approved):
Appears to work as intended

PR #15598, opened 2026-01-09
[Feature] Hero component tweak image displaying

### LINKED ISSUE #15349: ✨ Update browse jobs - header update
### Details

This child issue covers updates to the header featured in the parent issue.  This will update the shared Hero component and affect pages side-wide.

### Design file & designer

As part of the feature: https://www.figma.com/design/1ibVEEPTPTHbjifAgmcoGW/Browse-jobs--All-users-?node-id=355-4309&m=dev

Base component: https://www.figma.com/design/guHeIIh8dqFVCks310Wv0G/Component-library?node-id=1698-21757&m=dev

### 📸 Screenshots

Image

### 🌎 Localization

_No response_

### Proposed Implementation

_No response_

### ✅ Acceptance Criteria

- [ ] The page header image has been adjusted on wide screens and removed on narrow screens

vd1992:
🤖 Resolves #15349 

## 👋 Introduction

Adjust how images are rendered. Hide for narrowest view, move around a bit otherwise. 

## 🧪 Testing

1. Navigate to pages and view
2. Look at Chromatic

## 📸 Screenshot

Narrow view

image

Not so narrow view

image

### DIFF
--- a/apps/web/src/components/Hero/Hero.tsx
+++ b/apps/web/src/components/Hero/Hero.tsx
@@ -72,40 +72,40 @@ interface HeroSharedProps {
   additionalContent?: ReactNode;
 }
 
-interface HeroWithNavTabsProps extends HeroSharedProps {
+interface HeroNavAndImgProps extends HeroSharedProps {
   navTabs?: NavTab[];
   overlap?: never;
   imgPath?: string;
   centered?: never;
 }
 
-interface HeroWithOverlapProps extends HeroSharedProps {
-  navTabs?: never;
-  overlap: boolean;
-  imgPath?: string;
-  centered?: never;
-}
-
-interface HeroWithImageProps extends HeroSharedProps {
-  imgPath?: string;
-  centered?: never;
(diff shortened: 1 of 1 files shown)


vd1992:
Okay, how about now?

vd1992:
> If we are changing to a new layout that can't guarantee it, we may need to revisit the images we are using and how they are cropped 😢

Would that require punting this issue back to design then? To clarify some things

JoshBeveridge (approved):
This looks a lot better, thanks folks! @BrookeCameron do you mind also taking a quick look to make sure nothing has changed based on your image recommendations?

petertgiles (changes requested):
The code looks great and the updated pages look _way_ better.  I'm concerned about some possible unintentional changes flagged by Chromatic, though.  It looks like space is being reserved in the hero even when there is no image set.

Image

Image

Image

vd1992:
Okay, I think I got it to work now. To do what it should without breaking anything else. 
A little tricky when the TypeScript manipulation builds up

image

PR #15599, opened 2026-01-09
[Bug] Fix divider gap on `Notice`

### LINKED ISSUE #15315: 🐛 Tiny gap on dividing line

_No response_

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Submit an application

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] No gap on dividing line

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

vd1992:
🤖 Resolves #15315

## 👋 Introduction

Tweak the margin/padding to get rid of the tiny gap on the separator
Compound variant added and seems to have done the trick! 

## 🧪 Testing

1. Observe the summary
2. Two places I believe, application submitted page uses it as well as the error summary component

## 📸 Screenshot

image

image

### DIFF
--- a/packages/ui/src/components/Notice/Notice.stories.tsx
+++ b/packages/ui/src/components/Notice/Notice.stories.tsx
@@ -328,7 +328,6 @@ export const NoIcon: StoryObj<typeof Notice.Root> = {
 export const NonDismissible: StoryObj<typeof Notice.Root> = {
   args: {
     small: false,
-    dismissible: false,
     color: "gray",
   },
   render: (args) => (


petertgiles (approved):
To be honest, I didn't dive deeply enough into the layout of this component to understand the fix but the result in the app does look good!  👏   Two comments, though:
1. The other variants are expressed in terms of `mr` and `ml` but the new variant is expressed in terms of `mx` and `pl`.  Would it be possible to fix it with `mr` and `ml` instead, for consistency?
2. There's no Chromatic diff which makes me think there's no story for this variant.  Would you be willing to add one?

PR #15602, opened 2026-01-13
[Feat] Setup type safety for `react-router`

### LINKED ISSUE #14927: ♻️ Setup `react-router` type safety
### ♻️ Refactor

After migrating to framework mode, we can improve the type safety of our router.

### 🕵️ Details

You can read more about waht this means in [the documentation for type safety](https://reactrouter.com/explanation/type-safety).

### 🙋‍♀️ Proposed Solution

Follow [the setup guide](https://reactrouter.com/how-to/route-module-type-safety).

### ✅ Acceptance Criteria

- [ ] `.react-router` folder is ignored for git
- [ ] `tsconfig.json` includes generated types
- [ ] Type generation script added and comes before `tsc`
- [ ] Types are imported where needed (probably no where for now?)

esizer:
🤖 Resolves #14927 

## 👋 Introduction

Adds the support for type safety in `react-router` with the typegen command.

## 🕵️ Details

This is just preliminary work to get types to be generated. The types will eventually be used in routes and middleware.

For more detail see the [documentation on type safety](https://reactrouter.com/explanation/type-safety).

## 🧪 Testing

> [!TIP]
> The import path is `./+types/route-name` 

1. Run dev `pnpm dev` or `pnpm dev:fresh`
2. Confirm the `typegen` command runs
3. Confirm that it creates `./apps/web/.react-router` with types
4. Confirm that you can import those types in a route
5. Confirm the [expected types](https://reactrouter.com/explanation/type-safety#:~:text=The%20following%20types%20are%20generated%20for%20each%20route) exist

### DIFF
--- a/apps/web/src/pages/Classifications/UpdateClassificationPage.tsx
+++ b/apps/web/src/pages/Classifications/UpdateClassificationPage.tsx
@@ -22,7 +22,6 @@ import {
 } from "@gc-digital-talent/i18n";
 import {
   graphql,
-  Scalars,
   UpdateClassificationInput,
   FragmentType,
   getFragment,


esizer:
> I'm having trouble with this, too. Would you be willing to maybe update one page to illustrate how it works?

oh, looks like you identified a weird issue with param typegen :sweat_smile: I fixed that and added a basic exmaple in:

 - [fix typegen](https://github.com/GCTC-NTGC/gc-digital-talent/pull/15602/commits/c96bb1f78f6b451f9d899e0e129a12655861c78a)
 - [add example](https://github.com/GCTC-NTGC/gc-digital-talent/pull/15602/commits/e538ecdc4c9750452643f6058aed1f99b43d290c)
 
 I think the issue was two things, tsconfig had baseUrl set which made resolving some things difficult. Secondly, we were excluding types in hidden directories which is where `react-router` types live.
 
 The example added is not complete since we do not have middleware but it should show how the type safety works from params -> loader -> component

vd1992:
> > Not too sure about what exactly I ought to assert, but trying out the steps in testing.
> > Seeing a bunch of red in the generated files. Should I?
> 
> This im less sure about. I would not put much of an eye on the generated files, as long as they produce the proper types 🤔 Mine does not care about paths but hates the namespace but it still works.

I'm seeing much less red after https://github.com/GCTC-NTGC/gc-digital-talent/pull/15602/commits/c96bb1f78f6b451f9d899e0e129a12655861c78a

Now I just see
- use interface rather than type
- namespace complaining

PR #15603, opened 2026-01-13
[Feature] Accent-agnostic filter and search

### LINKED ISSUE #9957: ✨ accent-agnostic user/candidate search
## ✨ Feature

User and PoolCandidate name/email/general search treats accented characters uniquely. However, this can create confusion, especially when it comes to names, which people may choose to type with or without accents, somewhat arbitrarily. Like we've done in client-side search and sort features, it would be much better for users if we treated accented and non-accented characters the same.

## 🕵️ Details

In search, all accented versions of a character should be like the non-accented version. This should apply to the text being searched as well as the user input.

## 🎨 Design File

None

### 🧑‍🎨 Designer

@Jerryescandon 

If it is very difficult to apply this to the general search, at the very least it should apply to names and emails. That may be easier after #9956.

https://www.postgresql.org/docs/current/textsearch-dictionaries.html

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] search bar input in user and pool candidate tables ignores accents when matching names and emails
- [ ] Optional: search input also ingores accents for general text search matching

## 🛑 Blockers

Blocked by #9956.

petertgiles:
🤖 Resolves #9957 

## 👋 Introduction

Adds accent-agnostic filter and search to the user and pool candidates tables using the [unaccent](https://www.postgresql.org/docs/16/unaccent.html) dictionary.  Most of this PR is informed by https://stackoverflow.com/a/11007216 .

Once this PR gets approved, I'll work with Dan to replicate this in Azure.

## 🕵️ Details

I decided not to go with the collation strategy.  As far as I can tell, it doesn't work with the full text search feature we're using and it sounds like there are some performance penalties as well.

I opted to not include any indexes in this PR.  Since they're not (perf-wise) free, I figured I'd wait and see if they're actually needed in prod before putting them in.  I did put in all the support for them now so you can test them if you want.

## 🧪 Testing

1. Install all the database configuration with one of two options:
    1. Remove the `pgdata` docker volume and recreate like https://github.com/GCTC-NTGC/gc-digital-talent/pull/15440
    2. Manually run the CREATEs from infrastructure/db/2_template1.sql in your gctalent database
2. Log in as a user and update your first name, last name, email, and work_email to words with accents
3. Submit a pool candidate

> [!TIP]
>  Frontend validation may prevent email updates with accents.  You can use Tinker.  If you update it in the database directly you'll need to manually update the index.  You can use `artisan app:update-user-text-search-indexes`.

> [!TIP]
> Here are some words you can use if you want: Abigaëlle João Aglaé@example.org work@hôtel.gc.ca 

- [ ] On the user table, with "beta search":
    - [ ] general search matches accent-agnostic
    - [ ] first_name, last_name, email matches accent-agnostic
- [ ] On the user table, with "old search":
    - [ ] general search matches accent-agnostic
    - [ ] first_name, last_name, email matches accent-agnostic
- [ ] On the pool candidate table:
    - [ ] general search matches accent-agnostic
    - [ ] first_name, last_name, email matches accent-agnostic
- [ ] On a table with frontend search, like job poster templates
    - [ ] behaviour is unchanged, still accent sensitive
    - [ ] search still works

If you want to test indexes, you can add one manually with:
```sql
CREATE INDEX users_unaccent_name_trgm_idx ON users
USING gin (f_unaccent(first_name) gin_trgm_ops);
```
And then test it with:
```sql
explain analyze select * from users where f_unaccent(first_name) ilike f_unaccent('Abigaëlle')
```
Keep in mind that unless you have many rows, the optimizer might choose to do a sequential scan even with the index present.

## 📸 Screenshot

image

## 🚚 Deployment

We will need to update the user indexes again.
` ./artisan app:update-user-text-search-indexes`

This will take a while to complete in prod.  It doesn't necessarily have to happen exactly during deployment but text search won't work quite right until it is done.


apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
```diff
@@ -133,7 +133,7 @@ const ResponsiveTable = <TData extends object, TFilters = object>({
     },
     getRowId: rowSelect?.getRowId,
     autoResetPageIndex: false,
-    enableGlobalFilter: isInternalSearch,
+    manualFiltering: !isInternalSearch,
```
petertgiles: I'm not a tanstack table expert but with `enableGlobalFilter`, the `getFilteredRowModel` function was still front-end filtering.  This option seems to do what we expect.

infrastructure/db/2_template1.sql
```diff
@@ -0,0 +1,27 @@
+-- will apply to all new databases afterward, including temporary ones created for phpunit parallel testing
+\c template1;
+
+-- Add extensions to match what's in production
+CREATE EXTENSION IF NOT EXISTS plpgsql WITH VERSION '1.0';
+CREATE EXTENSION IF NOT EXISTS pgcrypto WITH VERSION '1.3';
+CREATE EXTENSION IF NOT EXISTS pg_stat_statements WITH VERSION '1.7';
+CREATE EXTENSION IF NOT EXISTS unaccent;
+CREATE EXTENSION IF NOT EXISTS pg_trgm;
```
petertgiles: The [pg_trgm](https://www.postgresql.org/docs/16/pgtrgm.html) extension is for use with possible indexes in the future.  It sounds like it can be used for fuzzy matching with full text search if we want to explore that, too.

PR #15611, opened 2026-01-14
[Debt] Merge status and final decision for pool candidates

### LINKED ISSUE #15518: ♻️ Replace `PoolCandidateStatus` and `FinalDecision`
### ♻️ Refactor

We need to replace both `PoolCandidateStatus` and `FinalDecision` with a new, simplified `ApplicationStatus`.

### 🕵️ Details

We currently have a few different ways to determine an applications current status where most are deprecated. This would be the work to remove the deprecated statuses with the new version.

The goal is to have a new, much more simple, `ApplicationStatus` that only covers a small set of both of the existing ones and store other details in different columns.

#### Application Status

 - `DRAFT` - Default status when created
 - `TO_ASSESS` - Change to this once submitted with the accompanying meta data
    - `screening_stage`
    - `assessment_stage` 
 - `REMOVED`
    - `removal_reason`
    - `removed_at`
 - `DISQUALIFIED`
    - `disqualification_reason`
    - `disqualified_at`
 - `QUALIFIED`
    - `placement_type`
    - `placed_at`
    - `expiry_date`
    - `qualified_at`

### 🙋‍♀️ Proposed Solution

A lot of this work should already be done but some has yet to be completed.

 - Create new `ApplicationStatus` enum with the values described
 - Create migration for:
    - Creating new columns: 
        - string `application_status`, default = `DRAFT`
        - string nullable `placement_type`
        - string nullable `disqualification_reason`
        - datetime nullable `disqualified_at`
        - datetime nullable `qualified_at`
    - Migrate current final decision / pool candidate status to new data
    - Remove columns `pool_candidate_status`, `computed_final_decision`, `final_decision_at`
 - Remove code to compute final decision
 - Update schema:
    - add and delete fields from migration to pool candidate type
    - Update mutations that accept `PoolCandidateStatus` to accept new types
    - Ensure validation exists for new fields

### Migration

SEE: 

 - [Status spreadsheet](https://docs.google.com/spreadsheets/d/1PDj4s4RGeDnZUmsQLaLSG3-_KD_jGaDLdXLsIwuQ7XQ/edit?usp=sharing) - This is the main authority
 - [Status flowchart](https://www.figma.com/board/FjnhpNxpCicPA3tejTTar9/Application-Status---Flowchart?node-id=0-1&p=f&t=uIWPij0gtna8ZPIT-0)

### ✅ Acceptance Criteria

- [ ] New `ApplicationStatus` enum and field on `PoolCandidate` model
- [ ] Data migrated to new status / meta data
- [ ] `PoolCandidateStatus` and `FinalDecision` removed
- [ ] Computation code removed
- [ ] Schema updated with new fields
- [ ] Validation for input and mutations updated with new status/meta data

esizer:
🤖 Resolves #15518 

## 👋 Introduction

This takes both the `pool_candidate_status` and `computed_final_decision` columns on the `pool_candidates` table and merges them into a new `application_status` column. No data should be lost, just transfered to new discrete columns.

## 🕵️ Details

This is quite a large PR that has a lot going on. But, it should be relativley straightforward. The main things this does:

 - Introduces a new `application_status` column that will be replacing both `pool_candidate_status` and `computed_final_decision`
 - Merges `final_decision_at` and `removed_at` into a single `status_updated_at`
 - Moves the extra data in the old statuses to their own columns i.e
   - `SCREENED_OUT_*` -> `disqualification_reason`
   - `QUALFIED_PLACED_*` -> `placement_type`
 - Removed `EXPIRED` based statuses, it is now derived solely based on the `expiry_date`
 - Adds a new `referring` boolean to support the transiaiton of `QUALIFIED_UNAVAILABLE`
     - This will be used in a new system for showing users on the search page
 - Adds some code to keep the deprecated info in sync for atleast 1 release

### Merging of statuses

This is the most complicated part of this. Thankfully, Jerbo made a very helpful flowchart describing exactly how it should function.

I would reccomend refernecing both of the following:

[Status spreadsheet](https://docs.google.com/spreadsheets/d/1PDj4s4RGeDnZUmsQLaLSG3-_KD_jGaDLdXLsIwuQ7XQ/edit?usp=sharing) - This is the main authority
[Status flowchart](https://www.figma.com/board/FjnhpNxpCicPA3tejTTar9/Application-Status---Flowchart?node-id=0-1&p=f&t=uIWPij0gtna8ZPIT-0)

### Timestamps

We don't care so much about the timestamps, these were just used to determining the status in some way. In order to simplify, we merged `removed_at` and `final_decision_at` into a single `status_updated_at`.

Honestly, I'm not even sure how needed this is :woman_shrugging: We only care about when something happened and this info exists in the activity log. So, we may be able to drop this as well. I just added it to keep parity for now.

### Metadata

We used to store additional information in the status itself (i.e a placed candidated would have a status of `PLACED_INDETERMINATE` and a final deciison of `QUALIFIED_PLACED`).

This was driving a significant amount of complexity trying to determine a simple status to display. The main reason for this PR is to move that and store it separatley. So not that same example would look like:

 - `application_status` = `QUALIFIED`
 - `placement_type` = `PLACED_INDETERMINATE`

### Referring

This is a little odd being here but, it is necessary. Right now, there is no way to "stop referring" a candidate without some weird work arounds with removing them. 

For a  while, admins were using the sort of weird `QUALIFIED_UNAVILABLE` status. Since this, currently, has no comparable way of representing that state, I added this column which will be used in the new system where and admin can toggle this to have a qualified candidate show in search requests or not without affecting their status.

### Backcompat

This adds a bit of code to keep the deprecated stuff (`pool_candidate_status` and `computed_final_decision`) in sync. Since this is pretty disruptive work, we don't want to delete the existing data, or have it go stale, while we have time to let this new work settle. This just allowing us to always revert this work if needed.

### Renaming

Something this does not do, is rename some components and functions. There were some things that were named specifically based on the legacy stuff (`CandidateStatusChanged`, `RevertFinalDecisionDialog`).

I did not rename them, mainly to keep the diff smaller since it does not impact functionality and can be done when we fully remove the deprecated code.

Also, the one was a notification and I was worried about it impacting the sending or displaying so I left it alone.

## 🧪 Testing

> [!TIP]
> While testing the frontened, it might make sense to have the applicant dashboard open as well as the admin side so you can see all the places statuses appear as you change them 

1. Take note of the excel and the existing candidate state before doing anything
2. Run the migration `artisan migrate`
3. Confirm the data migration leads to the expected state for candidates
4. Build `pnpm dev:fresh`
5. Login as any user
6. Apply to a process
7. Login as `admin@test.com`
8. Excersize all of the actions you can do with a candidate
9. Confirming you have the end result described in the excel and flowchart with each action
10. Confirm the candidate filters still work on different candidate tables
11. Confirm the search requests still function as expected

### DIFF
--- a/api/database/migrations/2026_01_13_145058_add_application_status_column_pool_candidates_table.php
+++ b/api/database/migrations/2026_01_13_145058_add_application_status_column_pool_candidates_table.php
@@ -110,18 +110,17 @@ return new class extends Migration
         DB::statement(<<<'SQL'
             UPDATE pool_candidates
             SET
+            assessment_step_id = CASE
+                WHEN application_status = 'TO_ASSESS' AND screening_stage = 'UNDER_ASSESSMENT' THEN assessment_step_id
+                ELSE NULL
+            END,
             screening_stage = CASE
                 WHEN application_status IN ('QUALIFIED', 'DISQUALIFIED', 'REMOVED') THEN NULL
                 WHEN application_status = 'TO_ASSESS' AND screening_stage IS NULL THEN 'NEW_APPLICATION'
                 ELSE screening_stage
-            END,
-            assessment_step_id = CASE
-                WHEN application_status IN ('QUALIFIED', 'DISQUALIFIED', 'REMOVED') THEN NULL
-                ELSE assessment_step_id
             END
             WHERE
-            application_status IN ('QUALIFIED', 'DISQUALIFIED', 'REMOVED')
-            OR (application_status = 'TO_ASSESS' AND screening_stage IS NULL);
+            application_status IN ('QUALIFIED', 'DISQUALIFIED', 'REMOVED', 'TO_ASSESS');
         SQL);
 
         // Make sure all `UNDER_ASSESSMENT` have an assessment step
@@ -179,7 +178,6 @@ return new class extends Migration
      */
     public function down(): void
     {
-
         Schema::table('pool_candidates', function (Blueprint $table) {
             $table->timestamp('final_decision_at')->nullable();
             $table->timestamp('removed_at')->nullable();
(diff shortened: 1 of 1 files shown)


api/database/migrations/2026_01_13_145058_add_application_status_column_pool_candidates_table.php
esizer: This will need a very high level of scrutiny :sweat_smile:

vd1992: For sure 😄 
Noticed off the top of my head
- Down migration needs to drop then regenerate `status_weight`
- Down migration also needs to drop `referring`

vd1992: Honestly finding things a little confusing due to the drift from what the issue contains. I can see why we held off on dropping the `pool_candidate_status` column for example, but otherwise
- Issue and Figma reference columns `disqualified_at` and `qualified_at` but they don't appear here or in the spreadsheet
- Keeping `computed_final_decision` though the issue says to get rid of it
- New columns `status_updated_at` and `referring`, though, the PR body mentions the changes, the flowchart uses old columns
- Flowchart refers to `removed_at` still
- For referring, the flowchart doesn't mention a boolean but rather two columns `StopReferralAtDate` and `StopReferralReason` 


image

image

image

esizer: Timestamps diff described in the PR:

> We don't care so much about the timestamps, these were just used to determining the status in some way. In order to simplify, we merged removed_at and final_decision_at into a single status_updated_at.

Keeping old code was also noted:

> This adds a bit of code to keep the deprecated stuff (pool_candidate_status and computed_final_decision) in sync. Since this is pretty disruptive work, we don't want to delete the existing data, or have it go stale, while we have time to let this new work settle. This just allowing us to always revert this work if needed.

Referring is compatibility only so we had somewhere to place `QUALIFIED_UNAVAILABLE` as described here:

> This is a little odd being here but, it is necessary. Right now, there is no way to "stop referring" a candidate without some weird work arounds with removing them.

api/database/migrations/2026_01_13_145058_add_application_status_column_pool_candidates_table.php
```diff
@@ -0,0 +1,190 @@
...
+     */
+    public function up(): void
+    {
+        Schema::table('pool_candidates', function (Blueprint $table) {
+            // Set nullable at first until we move data over
+            $table->string('application_status')->default(ApplicationStatus::DRAFT->name)->nullable();
+            $table->string('disqualification_reason')->nullable();
+            $table->dateTime('status_updated_at')->nullable();
+            $table->string('placement_type')->default(PlacementType::NOT_PLACED->name)->nullable();
+            $table->boolean('referring')->default(true);
```
vd1992: Per feature planning, I think it would be best to just add in `stop_referral_at` or some name as a datetime now rather than a boolean.
The other fields `stop_referral_reason` and whatever happens with pause length can come later I suppose

api/database/migrations/2026_01_13_145058_add_application_status_column_pool_candidates_table.php
```diff
@@ -0,0 +1,190 @@
...
+                removed_at = NULL
+        SQL);
+
+        Schema::table('pool_candidates', function (Blueprint $table) {
+            $table->dropColumn('application_status');
+            $table->dropColumn('disqualification_reason');
+            $table->dropColumn('status_updated_at');
+            $table->dropColumn('placement_type');
+            $table->dropColumn('referring');
+            $table->dropColumn('status_weight');
```
vd1992: I think status weight must be deleted first

image

api/database/factories/PoolCandidateFactory.php
```diff
@@ -40,22 +43,34 @@ class PoolCandidateFactory extends Factory
...
             ->pluck('id')
             ->first();
-        $removedStatuses = PoolCandidateStatus::removedGroup();
 
         return [
             'expiry_date' => $this->faker->dateTimeBetween('-1 years', '3 years'),
             'pool_candidate_status' => $this->faker->boolean() ?
                 $this->faker->randomElement([PoolCandidateStatus::QUALIFIED_AVAILABLE, PoolCandidateStatus::PLACED_CASUAL])->name :
                 $this->faker->randomElement(PoolCandidateStatus::cases())->name,
+            'application_status' => $this->faker->randomElement(ApplicationStatus::cases())->name,
```
vd1992: The factory should try to keep
`pool_candidate_status`
`application_status`
Synced in creation

api/app/Models/PoolCandidate.php
```diff
@@ -943,13 +943,14 @@ public function revertFinalDecision()
...
 
-        $atts = ['pool_candidate_status', 'expiry_date', 'final_decision_at', 'screening_stage'];
+        $atts = ['application_status', 'expiry_date', 'status_updated_at', 'screening_stage'];
         $old = $this->only($atts);
 
         $this->pool_candidate_status = PoolCandidateStatus::UNDER_ASSESSMENT->name;
+        $this->application_status = ApplicationStatus::TO_ASSESS->name;
         $this->expiry_date = null;
+        $this->status_updated_at = Carbon::now();
         $this->computed_final_decision = FinalDecision::TO_ASSESS->name;
```
vd1992: Method `revertFinalDecision` should also wipe out field `disqualification_reason`

api/app/Enums/PlacementType.php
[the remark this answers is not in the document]
esizer: Oh weird! Is that new? :sweat_smile: I see the excel seems to match that so I updated the groups to match the excel. Hopefully this looks good: [fix placement groups](https://github.com/GCTC-NTGC/gc-digital-talent/pull/15611/commits/b8048e084c4fb87fffde7249e817c2538a801427)

api/app/Models/PoolCandidate.php
```diff
@@ -943,13 +943,14 @@ public function revertFinalDecision()
     {
         $this->disableLogging();
 
-        $atts = ['pool_candidate_status', 'expiry_date', 'final_decision_at', 'screening_stage'];
+        $atts = ['application_status', 'expiry_date', 'status_updated_at', 'screening_stage'];
```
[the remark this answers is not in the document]
esizer: lol you both must take delight in forcing me to type out long variable names to try and force out more typos :sob: :stuck_out_tongue: 

Updated in [update atts var name](https://github.com/GCTC-NTGC/gc-digital-talent/pull/15611/commits/b183f97cdd9d8dd61d8acb7c1e646b2d3c7632aa)

api/app/Models/PoolCandidate.php
```diff
@@ -895,12 +893,13 @@ public function reinstate()
...
             default:
-                throw new Exception(ErrorCode::CANDIDATE_UNEXPECTED_STATUS->name);
+                // PASS: Do nothing
         }
 
-        $this->removed_at = null;
+        $this->status_updated_at = Carbon::now();
         $this->removal_reason = null;
         $this->removal_reason_other = null;
+        $this->application_status = ApplicationStatus::TO_ASSESS->name;
```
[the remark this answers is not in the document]
esizer: okay, after discussion we are going to go with the new system so reinstate puts both statuses intop `TO_ASSESS` and keeps them consistent.

api/app/Enums/PlacementType.php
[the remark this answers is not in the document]
esizer: ohhhh, huh. I understand now, it is to do with search resutls and not the cadndiate facing status. okay :+1: Let me look more into this :eyes:

api/app/Models/PoolCandidate.php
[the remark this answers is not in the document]
esizer: Updated to match what we discussed in [force reinstate to use new flow](https://github.com/GCTC-NTGC/gc-digital-talent/pull/15611/commits/aa365a4c7a8a6c1b2dbf176472e6a49b6e0ee4fd)

vd1992:
Okay, I do receive in-app notifications now! Nice!
I think I'll hand things off to Brinda now

brindasasi:
> Okay, I do receive in-app notifications now! Nice! I think I'll hand things off to Brinda now

Thanks Vachan! I still have 2 more things to check on this. Will update shortly.

brindasasi:
> > Okay, I do receive in-app notifications now! Nice! I think I'll hand things off to Brinda now
> 
> Thanks Vachan! I still have 2 more things to check on this. Will update shortly.

These 2 issues seems to be fixed in this PR. Its less confusing now.. great work!
Will let @mauli0110  close this after confirming in UAT

https://github.com/GCTC-NTGC/gc-digital-talent/issues/15662
https://github.com/GCTC-NTGC/gc-digital-talent/issues/15660

I've only 2 things now 

- `Final Decision` filter needs to be renamed to `Application Status`

Screenshot 2026-01-28 at 3 47 42 PM

- If we revert from Qualified or Disqualified , application goes to `Application Review` 
   Is this expected? 
   This application must have already completed Review & screening steps rt ? 
   @Jerryescandon

Jerryescandon:
> * If we revert from Qualified or Disqualified , application goes to `Application Review`
>   Is this expected?

Yes, this is by design.

PR #15635, opened 2026-01-15
[Refactor] Remove UsersThruPoolCandidatesExcelGenerator

### LINKED ISSUE #15530: ♻️ Remove api/app/Generators/UsersThruPoolCandidatesExcelGenerator.php
### ♻️ Refactor

We have a second user-excel generator that accepts a list of pool candidate IDs instead of a list of user IDs.  Unfortunately, there is a tremendous amount of code duplication in the second generator and it has fallen way behind the original one on updates.  We shouldn't be trying to maintain two different generators to produce identical files.

Image

### 🕵️ Details

_No response_

### 🙋‍♀️ Proposed Solution

I can see two different ways of doing this.

1. Have the frontend gather the user IDs from the selected pool candidates and call the same `DownloadUsersExcel` mutation as "by user ID".  I think I've done this in other places and is my preferred option as long as there are no blockers that I'm not thinking of.
2. On the backend, add functionality to accept pool candidates IDs instead of user IDs to `UserExcelGenerator` and still have a mostly unified code path.  I think this would probably be trickier and more effort.

### ✅ Acceptance Criteria

- [ ] The user profile excel file can be generated for a list of users or a list of pool candidates
- [ ] Regardless of how the query is built, there is only a single codepath to actually add render data to the Excel file

JamesHuf:
🤖 Resolves #15530.

## 👋 Introduction

Removes the UsersThruPoolCandidatesExcelGenerator replacing it's use with the UserExcelGenerator.

## 🧪 Testing

1. Rebuild the site.
2. Start the work queue.
3. Navigate to `admin/poolcandidates`.
4. Select some candidates.
5. Select Download Excel -> Download profiles Excel
6. Open the file and confirm it contains the profiles of the selected candidates.


bella77-69 (approved):
LGTM! 😁

PR #15638, opened 2026-01-16
[Feat] Client router middleware

### LINKED ISSUE #14929: ♻️ Create `react-router` middleware
### ♻️ Refactor

Intent is to create  the middleware and only add the global middleware to facilitate migrating to data loaders.

### 🕵️ Details

This work has previously been mostly completed in https://github.com/GCTC-NTGC/gc-digital-talent/pull/14872. In order to reach feature parity, we will need all the middleware in that PR.

 - `authMiddleware` - Adds authenticated user to context if present
 - `graphqlClientMiddleware` - Adds the `urql` client to context for use in loaders and actions
 - `intlMiddleware` - Adds the current locale info to context to be previded to the graphql API and return error messages
 - `protectedRouteMiddleware` - Guards specific routes when an authenticated user is not in the current context

### 🙋‍♀️ Proposed Solution

1. Create all middleware outlined in details, following the linked PR.
2. Add the global middleware (`authMiddleware`, `graphqlClientMiddleware`, `intlMiddlware`) to the root route

### ✅ Acceptance Criteria

 - [ ] Middleware created
 - [ ] Global middleware added to root route

### LINKED ISSUE #14930: ♻️ Add a role guard helper
### ♻️ Refactor

Add a helper function to guard specific pages by use role.

### 🕵️ Details

We currently guard certain pages/sections of the site using a `RequireAuth`. When we eventually have data loaders implemented, we can start using this function in loaders and drop the `RequireAuth` component entirely.

### 🙋‍♀️ Proposed Solution

Recrearte the function from https://github.com/GCTC-NTGC/gc-digital-talent/pull/14872 making an improvements you see fit.

Example:

```
export function guardByRoles(
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  context: RouteContext<any>,
  roles: RoleName[],
) {
  const user = context.get(userContext) as Maybe<Partial<UserAuthInfo>>;
  const isAuthorized = hasRole(roles, user?.roleAssignments);

  if (!isAuthorized) {
    throw new UnauthorizedError();
  }
}
```

### ✅ Acceptance Criteria

- [ ] Guard function created
- [ ] Mirrors the role guard in `RequireAuth`

esizer:
🤖 Resolves #14929 
:robot: Resolves #14930 

## 👋 Introduction

Adds client middleware for `react-router` and one example of it being implemented.

It also adds the role guard as part of the auth middleware.

### Middleware

This adds 3 middlewares that occur at the root of the application that add different data to a context for use in middleware, loaders and actions.

 1. `inltMiddleware` adds `react-intl` which has a redirect to the localized route from `/` to `/en` or `/fr`
 2. `graphqlClientMiddleware` adds our urql client to context
 3. `userMiddleware` adds auth info for the currently logged in user

This runs in order on the root route which gives all other routes access to it.

### Helpers

This adds a helper to protect routes. It is essentially our `RequireAuth` component. You can see an example of it in the `UpdateClassificationPage` but essentially it works like this:

- Add as a top middleware to run first
- Passes in context so it can know about the user
- Passes in request so we know that page we are trying to get to so we can return here aftter logging in if we need to
- Passes in an array of roles if we want to protect it by role. If nothing passed in, it will allow any role as long as you are authenticated

```ts
export const clientMiddleware: Route.ClientMiddlewareFunction[] = [
  async ({ context, request }, next) => {
    requireUser(context, request, [ROLE_NAME.PlatformAdmin]);
    return await next();
  },
];

```

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Login as `admin@test.com`
3. Navigate to edit a classification
4. Confirm everything works
5. Copy the URL for the classification edit
6. Logout
7. Paste the URL in
8. Confirm you can still go through the login redirect flow

### DIFF
--- a/apps/playwright/tests/login-logout.spec.ts
+++ b/apps/playwright/tests/login-logout.spec.ts
@@ -67,7 +67,9 @@ test.describe("Login and logout", () => {
     // stub the "user deleted" API response.
     // the auth response will indicate the user was deleted.
     await page.route("**/graphql", async (route) => {
-      const reqJson = (await route.request().postDataJSON()) as GraphQLOperation | null;
+      const reqJson = (await route
+        .request()
+        .postDataJSON()) as GraphQLOperation | null;
       if (reqJson?.operationName === "authorizationQuery") {
         const body = JSON.stringify({
           data: { myAuth: null },


yonikid15 (changes requested):
Exciting stuff so far! I'm still looking into what's causing it, but I can't get to the edit classification page. It redirects me to the sign in page, even if I'm already signed in.

The url I'm redirected to: `http://localhost:8000/en/login-info?from=%2Fen%2Fadmin%2Fsettings%2Fclassifications%2Fbb0be484-d796-409d-a05f-c27651499528%2Fedit`

image

PR #15701, opened 2026-01-28
[Feat] Add filters to process activity log

### LINKED ISSUE #15364: ✨ Add filters to pool activity log
### Details

Add filter scopes and dialog to the pool activity log.

### Design file & designer

_No response_

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

- Add scopes for the required filters to `Activty` model
- Create a dialog that applies scopes to query on frontend

### ✅ Acceptance Criteria

- [ ] Pool activity log has new filter scopes
- [ ] Dialog created to filter the activity log

esizer:
🤖 Resolves #15364 

## 👋 Introduction

Adds a dialog to filter the process acivity log by:

 - Date range
 - Assessment member actions
 - Candidates (subject)
 - Actions taken

## 🕵️ Details

We reduced the scope of the original ticket by omitting some filters that were slightly more challenging to implement.

## 🧪 Testing

> [!TIP]
> Before testing, it might be a good idea to create a pool and apply to it then make a bunch of decisions. This will help determine that the filters are functioning as expected 

1. Build `pnpm dev:fresh`
2. Login as `admin@test.com`
3. Navigate to a process activity page `/admin/pools/{id}/activity`
4. Use the filter dialog
5. Confirm it returns the expected results

## 📸 Screenshot

swappy-20260128_131556
swappy-20260128_131608

### DIFF
--- a/api/lang/fr/activity_event.php
+++ b/api/lang/fr/activity_event.php
@@ -1,0 +1,16 @@
+<?php
+
+return [
+    'created' => 'Créé',
+    'updated' => 'Mis à jour',
+    'deleted' => 'Supprimé',
+    'submitted' => 'Soumis',
+    'qualified' => 'Qualifié',
+    'disqualified' => 'Disqualifié',
+    'placed' => 'Placé',
+    'added' => 'Ajouté',
+    'removed' => 'Retiré',
+    'reinstated' => 'Réintégré',
+    'reverted' => 'Restauré',
+    'published' => 'Publié',
+];
--- a/apps/web/src/lang/fr.json
+++ b/apps/web/src/lang/fr.json
@@ -2943,6 +2943,10 @@
     "defaultMessage": "Outils d'assistance",
     "description": "Title of admin user support tools section"
   },
+  "9zjyN4": {
+    "defaultMessage": "Action",
+    "description": "Label for the action filter for activity log"
+  },
   "A+4huJ": {
     "defaultMessage": "Mettre à jour ou supprimer une expérience dans votre chronologie de carrière",
     "description": "Display text for edit experience form in breadcrumbs"
@@ -3442,6 +3446,10 @@
   "CFIG+8": {
     "defaultMessage": "Postes en français uniquement",
     "description": "French Positions message"
+  },
+  "CG07OF": {
+    "defaultMessage": "Membre de l'équipe d'évaluation",
+    "description": "Label for a process assessment team member input"
   },
   "CKsQyK": {
     "defaultMessage": "Accessibilité et mesures d’adaptation",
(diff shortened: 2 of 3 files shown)


apps/web/src/pages/Pools/PoolActivityPage/components/useAvailableAssessmentMembers.ts
```diff
@@ -0,0 +1,71 @@
...
+import { graphql } from "@gc-digital-talent/graphql";
+import { ROLE_NAME, RoleName } from "@gc-digital-talent/auth";
+import { unpackMaybes } from "@gc-digital-talent/helpers";
+
+const ASSESSMENT_MEMBER_ROLES: RoleName[] = [
+  ROLE_NAME.PlatformAdmin,
+  ROLE_NAME.CommunityRecruiter,
+  ROLE_NAME.CommunityTalentCoordinator,
+  ROLE_NAME.ProcessOperator,
+];
```
vd1992: I believe this is missing Community Admin. I recommend creating a const array in the below file, call the array maybe `ELEVATED_MEMBER_ROLES`
Defining it in a more central place makes it harder for roles to be wrong in single files

image

api/app/Models/Activity.php
```diff
@@ -51,6 +52,48 @@ public function scopeWhereIsAggregatePoolActivity(Builder $query, array $args, P
 
     }
 
+    public function scopeWhereDate(Builder $query, array $args)
+    {
```
vd1992: Unsurprisingly, timezones will make this more confusing than we wish. 
The rendered date is zone-localized, but the stored date which is what gets filtered is not. So filtering for January 28th will not show the action below despite it appearing to happening on the 28th. It happens on the 29th by UTC time so only appears for you if you filter for the following day

If I was to throw out a suggestion, you'd want to compute the start and end date to filter between locally then query with it. In example, for a user in central time CDT you'd take start of day to be 00:01 -> 05:01 UTC (5 hours ahead)
https://github.com/marnusw/date-fns-tz?tab=readme-ov-file#fromzonedtime

image

apps/web/src/pages/Pools/PoolActivityPage/components/useAvailableAssessmentMembers.ts
esizer: Yeah, that is a good suggestion. Updated in [update assessment memeber roles array](https://github.com/GCTC-NTGC/gc-digital-talent/pull/15701/commits/7bbfe5a2c418c5790ca668cf024b33a35f1e09cd)

api/app/Models/Activity.php
vd1992: Unfortunately, still not quite... 

image

PR #15753, opened 2026-02-04
[Fix] Form controls being announced twice

### LINKED ISSUE #15513: ⌨ Form labels repeating twice

In the add an experience and work preferences section all of the form field labels through out this form repeat twice when you go through them. For example, Voiceover will read "my role* my role*" when you are on the field.

Expected behaviour: Form fields should only be read once

### WCAG 2.1 Criteria

_No response_

### 📋 Steps to Reproduce

1. sign in 
2. navigate to either add an experience or work preferences
3. using VoiceOver and tab key navigate through the form fields

### 🕵️ Details

_No response_

macOS

Safari

VoiceOver

### 📸 Screenshot

https://github.com/user-attachments/assets/9a284649-649e-4246-8e1e-ea557680aef8

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Form fields are only read once

esizer:
🤖 Resolves #15513 

## 👋 Introduction

Fixes an issue where some of our inputs had their labels announced twice to screenreaders.

## 🕵️ Details

This is a safari specific bug that requires a work-around. That is, adding the `aria-labelledby` to override the actual label itself. Safari for some reason, treats the label and its inner text content as two separate nodes both assigned to the input. The `aria-labelledby` takes priority over that as a single text node.

## 🧪 Testing

> [!TIP]
> Storybook might be the best way to tesrt this so you can check all inputs there instead of searching the app for every type of input we have. 

1. Build `pnpm dev:fresh`
2. Open each input (even the ones not touched by this PR) in safari with voiceover
3. Focus the input
4. Confirm the label is only read a single time

### DIFF
--- a/packages/forms/src/components/TextArea/TextArea.tsx
+++ b/packages/forms/src/components/TextArea/TextArea.tsx
@@ -139,8 +139,8 @@ const TextArea = ({
           })}
           {...(readOnly
             ? {
-              readOnly: true,
-            }
+                readOnly: true,
+              }
             : {})}
           {...rest}
         />


mnigh (changes requested):
Still reading out the label twice in Safari.

## Screenshots
Image

Image

mnigh (approved):
LGTM!

PR #15801, opened 2026-02-06
[Bug]  Add first render condition to ResponsiveTable to prevent loop

### LINKED ISSUE #15799: 🐛 Going back in browser after applying filters in "Requests" stalls the page.

When in "Requests", we are applying filters to make selection of specific requests. Then we click on one of the selected requests. If we try to go back to the selected list by backing up in browser, the action causes the page to stall and freeze. 

### 🕵️ Details

This happens in both Edge and Chrome.
Also in both prod and UAT. 

To note - when trying to use the button "Back to requests" the list does not retain filters. We need to start the selection from scratch. 

### 📋 Steps to Reproduce

1. In the admin view go to requests.
2. Apply filters - let's say "New" and "Business line advisory". 
3. A list of requests that reflects the filters shows up.
4. Click on any of the requests in the list to open it up.
5. Try to back up in browser to return to the list of selected requests.
6. Observe the page stalling.

### 📸 Screenshot

Image

Image

Image

Image

Image

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

_No response_

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

yonikid15:
🤖 Resolves #15799 

## 👋 Introduction

- Adds a first render reference check to the responsive table `useEffect`, to prevent an infinite loop.
- NOTE: Fixed by @esizer :pray: 

## 🧪 Testing

1. Refresh app
2. Go to talent request page 
3. Add filters that provide at least one request
4. View the request in the table. 
5. Then try going back in the browser.
6. Ensure that the table renders.
7. Ensure that other tables aren't effected.

### DIFF
--- a/apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
+++ b/apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
@@ -92,6 +92,7 @@ const ResponsiveTable = <TData extends object, TFilters = object>({
 }: TableProps<TData, TFilters>) => {
   const id = useId();
   const intl = useIntl();
+  const isFirstRender = useRef(true);
   const { announce } = useAnnouncer();
   const hasUpdatedRows = useRef<boolean>(false);
   const [, setSearchParams] = useSearchParams();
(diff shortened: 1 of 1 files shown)


mnigh (commented):
I was able to reproduce the bug once but not again 🤷

PR #15852, opened 2026-02-13
[Debt] Improve `PoolCandidateFactory` accuracy

### LINKED ISSUE #15410: ♻️ Increase accuracy of `PoolCandidateFactory`
### ♻️ Refactor

Our `PoolCandidateFactory` can result in some unrealistic data, we should look into this to ensure the data it creates would be a realistic value we expect to see through normal app usage.

### 🕵️ Details

We use a quite large `afterCreating` in the `configure` method that seems to set some related data based on the state. I think this issue is, the state is pretty random so some of it can lead to the `afterCreating` creating some unrealistic states.

This is nice because the result is inferred from the state set and devs don't need to be explicit about it. However, because the state is random, it can lead to unexpected result.

Image

### 🙋‍♀️ Proposed Solution

I feel like there are two possible routes based on what we decide:

1. Keep using one large `afterCreating` to make usage implicit and easy and just fin out where it is incorrect
2. Refactor the one large `afterCreating` into more discrete, determined state methods like we do for `suspended`

### ✅ Acceptance Criteria

- [ ] `PoolCandidateFactory` produces realistic data

esizer:
🤖 Resolves #15410 

## 👋 Introduction

Refactors the `PoolCandidateFactory` to rely more on state methods.

## 🧪 Testing

1. Confirm that the expected states exist
2. Confirm seeders still create expected data
3. Confirm tests will function and have expected assertions

### DIFF
--- a/api/database/factories/PoolCandidateFactory.php
+++ b/api/database/factories/PoolCandidateFactory.php
@@ -71,16 +71,24 @@ class PoolCandidateFactory extends BaseFactory
             //  TO DO: Do we complete the user profile?
 
             // Education requirement
-            $eduRequirement = $this->faker->randomElement(EducationRequirementOption::classificationRequirements($pool->classification->group));
+            $eduRequirement = $this->faker->randomElement(
+                EducationRequirementOption::classificationRequirements($pool->classification->group)
+            );
+
             $updates['education_requirement_option'] = $eduRequirement;
             $isEdu = $eduRequirement === EducationRequirementOption::EDUCATION->name;
-            $exp = $isEdu
-                ? ($user->educationExperiences->first() ?? EducationExperience::factory()->for($user)->create())
-            : ($user->workExperiences->first() ?? WorkExperience::factory()->for($user)->create([
-                'employment_category' => $user->computed_is_gov_employee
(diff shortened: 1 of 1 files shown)


api/tests/Feature/PoolApplicationTest.php
```diff
@@ -746,12 +736,10 @@ public function testApplicationDeletion(): void
...
-        $newPoolCandidate = PoolCandidate::factory()->create([
-            'user_id' => $this->applicantUser->id,
-            'pool_id' => $newPool->id,
-            'application_status' => ApplicationStatus::DRAFT->name,
-            'expiry_date' => config('constants.far_future_date'),
-        ]);
+        $newPoolCandidate = PoolCandidate::factory()
+            ->for($this->applicantUser)
+            ->for($newPool)
+            ->create();
```
yonikid15: Might need to set the far future expiry date to prevent flaky test

```php
create(['expiry_date' => config('constants.far_future_date')]);

yonikid15 (commented):
Amazing work Eric 🎉 
I just have few comments I made to ensure we didn't miss anything

PR #15864, opened 2026-02-13
[Spike] Take the site fully offline

### LINKED ISSUE #15533: 🦔  How can we take the site fully offline and display a message without the database?
### Details

We recently took the database offline for an extended period.  We can put a banner up ahead of time to warn people but once the database is down, that banner no longer works and the site just looks really broken.  And there's no message to tell people what's happening. How can we display a message when we're expecting the database to go offline or we really need people to stop using the site entirely?

### Design file & designer

_No response_

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

We could try to detect a "no database" error from a graphql response and show a nice message but it doesn't help us if we want to take the app fully offline.

A couple of ideas:
- an app service "maintenance" deployment slot we can switch in
- update the nginx config on the live app service
- work with IMTD to do it with the F5
- another Azure feature I don't know about?

### ✅ Acceptance Criteria

- [ ] How can we take the site fully offline and display a message without the database?

petertgiles:
🤖 Resolves #15533 

> [!NOTE]
> This is not a complete solution.  It should not get merged to main. 

## 👋 Introduction

Proposes a solution for taking the production site offline in a more graceful way.

## 🕵️ Details

The idea is that we could have an nginx config file that we could hot-swap in to effectively take the site offline.  As shown, it could redirect all requests to a static page indicating that the site is offline.

Additionally, this allows us to provide a graphql error that will be shown to users who have the app already booted when the site goes offline.

In production, the nginx config is copied to the right location during post-deploy:
https://github.com/GCTC-NTGC/gc-digital-talent/blob/5195bc963e549986f6ac90047e9420acc043caeb/infrastructure/bin/post_deployment.sh#L102

With this solution, we could SSH into the app service, copy the new conf file over it, and run `nginx -s reload`.

## 🧪 Testing

1. Remove your webserver container and then `compose up` again.
2. Attempt to navigate to any URL on the site and observe how you are redirected to the offline.html page.

## 📸 Screenshot

image

image

## :thinking: Additional thoughts

It's annoying to have individual folders for each nginx config file.  I'd like to pull them out.

It's seems like the sites-available/enabled system is frowned upon now?  We could switch it to the conf.d directory.
https://serverfault.com/questions/527630/difference-in-sites-available-vs-sites-enabled-vs-conf-d-directories-nginx

With three conf files we might consider factoring out shared blocks into separate files.

If we really wanted to, we could probably make the offline.html file in the /tmp directory so we could customize it on a live system and point out to that.  :cowboy_hat_face:

### DIFF
--- a/apps/web/public/offline.html
+++ b/apps/web/public/offline.html
@@ -1,0 +1,13 @@
+<!doctype html>
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <title>Offline</title>
+  </head>
+  <body>
+    <h1>Offline</h1>
+    <p>
+      The site is currently offline. Please try again later.
+    </p>
+  </body>
+</html>
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,7 +27,7 @@ services:
     restart: always
     volumes:
       - ./:/home/site/wwwroot
-      - ./infrastructure/conf/nginx-conf-local:/etc/nginx/sites-available/
+      - ./infrastructure/conf/nginx-conf-offline:/etc/nginx/sites-available/
       - ./infrastructure/conf/xdebug.ini:/usr/local/etc/php/conf.d/xdebug.ini
       - ./infrastructure/conf/php-fpm-www.conf:/usr/local/etc/php-fpm.d/www.conf
       - ./infrastructure/conf/php-fpm-docker.conf:/usr/local/etc/php-fpm.d/docker.conf:ro
--- a/infrastructure/conf/nginx-conf-offline/default
+++ b/infrastructure/conf/nginx-conf-offline/default
@@ -1,0 +1,32 @@
+# Swap in this config to take the site offline
+error_log /var/log/nginx/error.log notice;
+
+server {
+    listen 8080;
+    listen [::]:8080;
+    server_name localhost;
+    gzip_static on;
+    absolute_redirect off;
+
+    location = /graphql {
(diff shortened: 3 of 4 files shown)


esizer (approved):
I think this is a pretty easy to accomplish solution! I have some very minor concerns but ont serious enough to block moving forward with this approach.

 1. This is a manual process, meaning we would ontl be prepared if we knew the site was going down. Though, maybe that is not a big issue. Though, it would be nice in the future to have this be something that could happen in unexpected outages
 2. Though there aren't that many steps, maybe we could have a simple bash script that copied the config/made the changes and reloaded nginx and the inverse of that? Something to think about when we do the real implementation

But, I think this is a great solution that doesnt overcomplicate things and we can build on when we feel the need or have spare time to do so 👍

esizer:
> Yeah, this would only be for a manual intervention. I'm really not sure how we would automate this. I think in most unexpected cases we would get a degradation of service and someone would need to make a decision to take it offline anyway. If we could define rules though, like "if the database stops responding, we always want to go offline" then it might be possible. In that example, we probably wouldn't need Nginx and could handle it in React I think.

Yeah, that makes a lot of sense to me! 

> Yeah, 100%. I didn't think it was necessary for this proof of concept but I agree that would we want to make this as foolproof as possible.

Yeah, I agree. Totally unecessary for a spike.

PR #15239, opened 2025-12-01
[Refactor] Set storage path directly

petertgiles:
🤖 Resolves #11471

## 👋 Introduction

Sets the `storage_path` for Laravel directly in Azure using the LARAVEL_STORAGE_PATH variable [introduced in Laravel 10](https://github.com/laravel/framework/pull/48115).  

## 🕵️ Details

I've settled on /var/site as the dir to place `storage` and `bootstrap/cache` in the app service, to mirror /home/site.  This is done to standardize the location, make it easier to use, and hopefully, pull into Azure soon.  It also remove's Tristan's "somewhat hacky solution" for storage. :laughing: 

Most of the diff in the PR is me changing the name of the disk from `userGenerated` to `user_generated`.  Sorry, it' was driving me crazy.  :stuck_out_tongue: 

## 🧪 Testing

1. Check locally
2. Check in azure

- [ ] Can log from a scheduled job

For example, set PruneUserGeneratedFiles to `->everyMinute()` in `api/app/Console/Kernel.php` and then run `php /home/site/wwwroot/api/artisan schedule:run`

- [ ] Can log from a queued job

For example, set up with the "team" api key, run `./artisan send-notifications:test admin@test.com`, then start a queue worker.

- [ ] Can generate, download, and prune user generated files.

## 🚚 Deployment

I haven't updated IaC for dev or uat.  So far I just made some manual changes to the dev app service directly.

1. Add: LARAVEL_STORAGE_PATH=/var/site/storage
2. Update:
    APP_CONFIG_CACHE=/var/site/bootstrap/cache/config.php
    APP_EVENTS_CACHE=/var/site/bootstrap/cache/events.php
    APP_PACKAGES_CACHE=/var/site/bootstrap/cache/packages.php
    APP_ROUTES_CACHE=/var/site/bootstrap/cache/routes.php
    APP_SERVICES_CACHE=/var/site/bootstrap/cache/services.php
3. Change (or remove to default): CACHE_DRIVER=file
4. Change (or remove to default): FILESYSTEM_DRIVER=local
5. Remove: STORAGE_USER_GENERATED_FILES_PATH
6. Add: LIGHTHOUSE_SCHEMA_CACHE_PATH=/var/site/bootstrap/cache/lighthouse-schema.php
7. Update appCommandLine=/home/site/wwwroot/infrastructure/bin/post_deployment.sh https://hooks.slack.com/services....

PR #15875, opened 2026-02-16
[Feature] New artisan command for manual pool emails

### LINKED ISSUE #15873: ✨ Artisan command to send notification for already published jobs
### Details

We have a use case that jobs are posted while email notifications are stopped for an investigation. 

Another time email notifications got malfunctioned and users didn't receive email for a specific job poster. 

We need to retro fit that notifications which are published more than 24 hours ago. 

### Design file & designer

_No response_

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

_No response_

### ✅ Acceptance Criteria

- [ ] New Job posted email template should be used
- [ ] Artisan command should accept a pool id
- [ ] **Email** and **in app** Notifications should be sent for the given pool id

petertgiles:
🤖 Resolves #15873 

## 👋 Introduction

Adds a new artisan command for sending emails for a specific pool.

## 🕵️ Details

Can send emails for a specific pool with a command like `./artisan send-notifications:pool-published-specific 818e3e2d-5a41-4cae-96a2-0cdfb94016ed`.

This PR is a "quick and safe" implementation.  It's pretty limited and not very DRY but given the time constraints and need for safety, I'm providing a fairly naive solution.   You can diff the two commands together to compare.

Some ideas for possible future improvements:

- combine the two artisan commands
- allow providing multiple pool IDs together
- allow running with a different time span than 24 hours

I did fix some logging in the existing command that was missed in https://github.com/GCTC-NTGC/gc-digital-talent/pull/15239 though.

## 🧪 Testing

1. Truncate the `jobs` and `notifications` tables
2. Pick a pool and note its ID
3. Find out how many users want notifications
4. Find out how many users want email notifications
5. Run the command
6. Check how many notifications and jobs are queued

## 📸 Screenshot

image

### DIFF
--- a/api/app/Console/Commands/SendNotificationsPoolPublishedSpecific.php
+++ b/api/app/Console/Commands/SendNotificationsPoolPublishedSpecific.php
@@ -1,0 +1,95 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Enums\NotificationFamily;
+use App\Models\Pool;
+use App\Models\User;
+use App\Notifications\NewJobPosted;
+use Illuminate\Console\Command;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\Log;
+use Throwable;
+
+// This is a copy of SendNotificationsPoolPublished designed to be run manually and choose a specific pool.
+class SendNotificationsPoolPublishedSpecific extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'send-notifications:pool-published-specific {poolId}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Send notifications to users about a specific pool being published.';
+
+    /**
+     * Execute the console command.
+     *
+     * @return int
+     */
+    public function handle()
+    {
+        $poolId = $this->argument('poolId');
+        $pool = Pool::find($poolId);
+
+        if (! $pool) {
+            $this->error('Could not find a pool with that ID.');
+
+            return Command::FAILURE;
+        }
+
+        $this->info('Found the pool "'.($pool->name['en']).'" published on "'.($pool->published_at).'" set to close on "'.($pool->closing_date).'".');
+        if (! $this->confirm('Do you wish to send notifications for this pool?')) {
+            $this->info('Cancelled command.');
+
+            return Command::FAILURE;
+        }
+
+        $notifications = collect([
+            new NewJobPosted(
+                $pool->name['en'],
+                $pool->name['fr'],
+                $pool->id
+            ),
+        ]);
+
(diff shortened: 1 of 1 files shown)


api/app/Console/Commands/SendNotificationsPoolPublishedSpecific.php
```diff
@@ -0,0 +1,95 @@
...
+            User::whereJsonContains('enabled_email_notifications', NotificationFamily::JOB_ALERT->name)
+                ->orWhereJsonContains('enabled_in_app_notifications', NotificationFamily::JOB_ALERT->name)
+                ->chunk(200, function (Collection $users) use ($notifications, &$successCount, &$failureCount) {
+                    foreach ($users as $user) {
+                        foreach ($notifications as $notification) {
+                            try {
+                                $user->notify($notification);
+                                $successCount++;
+                            } catch (Throwable $e) {
+                                // best-effort: log and continue
```
brindasasi: I dont think this catches the failure.
My emails are failed due to config reasons but the failure count is 0

Image
Image

petertgiles: Keep in mind, this command just _queues_ jobs, it doesn't actually send them.  You will only get errors logged here if something goes wrong when the notifications are getting queued into jobs and the in-app notifications.  Errors that occur when the queue workers are running jobs are logged later, in the jobs.log file.

brindasasi: yeah .. I don't know  if there is a better way to get notified about those errors unless we get into that log file.
Btw, the command works well .. good to merge in .. I''ll take this to prod tonight

brindasasi (changes requested):
didn't receive in-app notifications during testing and as well as in prod

brindasasi:
> > didn't receive in-app notifications during testing and as well as in prod
> 
> Are you sure? I got an in-app notification last night in prod. image

cross verified in-app notifications are fine 
Email notifications there are still some discrepancies.. but that would be outside the scope of this PR 
**Total Subscribers** 
```
select count(*) from users where users.enabled_email_notifications 
@> '["JOB_ALERT"]'::jsonb and users.deleted_at IS NULL;
```
**12403**

**Total Sent**
```
grep "https://talent.canada.ca/en/jobs/554a31d4-d480-4eec-be6a-02f2b5e8efff" jobs.log | wc -l
```
**10748**

**Successful Delivery**
`grep "Sent message: 201" emails.log | wc -l
`
**10292**

** Error code 429 - too many requests**
`grep "Sent message: 429" emails.log | wc -l
`
**453**

**Error code 400**
`grep "Sent message: 400" emails.log | wc -l
`
**3**

brindasasi (approved):
Thanks Peter for all the help diagnosing the issue ! LGTM

PR #15976, opened 2026-02-25
[Fix] Stabilize profile doc generator test

### LINKED ISSUE #14550: 🐛 testUserProfileDocSnapshot false positives

The test Tests\Feature\Generators\UserDocGeneratorTest::testUserProfileDocSnapshot causes many false positives when a factory changes its use of faker.  I guess the randomness sequence changes and causes a big cascade of differences.  This makes it difficult to noticed true positives.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Make sure the snapshot is up to date.
2. Add the line `$throwAway = $this->faker->boolean();` to UserFactory.php
3. `./artisan test tests/Feature/Generators/UserDocGeneratorTest.php`

### 📸 Screenshot

Image

Image

### 🙋‍♀️ Proposed Solution

Maybe running `$this->faker->seed(0);` between factory steps?
Maybe passing the seed into the factories?
Maybe something like this? https://stackoverflow.com/a/71052979
Maybe hand-building the models being tested?  😬 

### ✅ Acceptance Criteria

- [ ] Test remains stable even if a call to faker is added or remvoed.

esizer:
🤖 Resolves #14550 

## 👋 Introduction

Makes the user document snapshot test more deterministic.

## 🕵️ Details

Not a huge fan of this but honestly, nothing else was working. I wonder if it makes sense to have our factories define a `->deterministic()` flag so they can take control. But for now, I think it makes sense here :woman_shrugging: 

## 🧪 Testing

1. Make some non-impactful changes to some factories
2. Run the the test
3. Confirm snapshot is stable

### DIFF
--- a/api/tests/Feature/Generators/__snapshots__/UserDocGeneratorTest__testUserProfileDocSnapshot__1.html
+++ b/api/tests/Feature/Generators/__snapshots__/UserDocGeneratorTest__testUserProfileDocSnapshot__1.html
@@ -104,12 +104,12 @@ div > *:first-child {page-break-before: auto;}
 <p>C-suite role title: Chief Digital Officer</p>
 <p>Additional details: Did deterministic testing of policy.</p>
 <p>Work streams:</p>
-<p>Orn, Vandervort and Dach EN</p>
+<p>Snapshot community EN</p>
 <p>Snapshot work stream EN</p>
 <h3>Skill showcase</h3>
 <p>The skill showcase allows a candidate to provide a curated series of lists that highlight their specific strengths, weaknesses, and skill growth opportunities. These lists can provide you with insight into a candidate's broader skillset and where they might be interested in learning new skills.</p>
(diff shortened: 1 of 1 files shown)


esizer:
> hard code community name and skill category

Good catch! I have hardcoded those as well in [hard code community name and skill category](https://github.com/GCTC-NTGC/gc-digital-talent/pull/15976/commits/a3c3b8364647f3833e4ca7d261179254fcd9639b)

esizer:
> Peter commented on the issue that all the *DocGeneratorTest tests do the same thing. Do we want to create separate issues for each of them?

ohhh, that is a good point. I missed that. Do you think that it makes sense for me to do that work in this PR? Or would that be too much? 😅 

If you think its too much for a single PR, we can break it up into a few smaller issues, otherwise, im happy to do all of the tests in that folder to match this pattern.

PR #16044, opened 2026-03-03
[Deb]t Address `react-doctor` warnings

### LINKED ISSUE #16006: ♻️ Address `react-doctor` issues for `apps/web`
### ♻️ Refactor

I ran react-doctor and have noted some of the concerns it raised.

### 🕵️ Details

> [!NOTE]
> I have omitted a lot of the warnings since I don't necessarily agree with them. If we ever want to introduce this into our tooling we can configure them but for now, I just ignored them on a single run.

## react-doctor/rerender-memo-with-default-value

Default prop value [] creates a new array reference every render — extract to a module-level constant

**Suggestion**: Move to module scope: `const EMPTY_ITEMS: Item[] = []` then use as the default value

  - src/pages/Pools/AssessmentPlanBuilderPage/components/AssessmentDetailsDialog.tsx: 220
  - src/components/EmployeeSearchInput/EmployeeSearchInput.tsx: 32
  - src/components/SkillBrowser/SkillBrowser.tsx: 27
  - src/components/NotificationList/NotificationListPage.tsx: 60
  - src/components/PoolFilterInput/PoolFilterInput.tsx: 31

## react-doctor/no-array-index-as-key

Array index "i" used as key — causes bugs when list is reordered or filtered

**Suggestion**: Use a stable unique identifier: `key={item.id}` or `key={item.slug}` — index keys break on reorder/filter

  - src/pages/TalentNominations/NominationGroup/components/NominatorList.tsx:29

## react-doctor/no-cascading-set-state

3 setState calls in a single useEffect — consider using useReducer or deriving state

**Suggestion**: Combine into useReducer: `const [state, dispatch] = useReducer(reducer, initialState)`

  - src/pages/Pools/AssessmentPlanBuilderPage/components/AssessmentDetailsDialog.tsx:304
  - src/components/NotificationDialog/NotificationDialog.tsx:62

## react-doctor/no-usememo-simple-expression

useMemo wrapping a trivially cheap expression — memo overhead exceeds the computation

**Suggestion**: Remove useMemo — property access, math, and ternaries are already cheap without memoization

  - src/pages/TalentEvents/TalentEvent/TalentEventNominationsTable.tsx:103

## react-doctor/rerender-memo-with-default-value

Default prop value [] creates a new array reference every render — extract to a module-level constant

**Suggestion**: Move to module scope: `const EMPTY_ITEMS: Item[] = []` then use as the default value

  - src/pages/Pools/AssessmentPlanBuilderPage/components/AssessmentDetailsDialog.tsx:220
  - src/components/EmployeeSearchInput/EmployeeSearchInput.tsx:32
  - src/components/SkillBrowser/SkillBrowser.tsx:27
  - src/components/NotificationList/NotificationListPage.tsx:60
  - src/components/PoolFilterInput/PoolFilterInput.tsx:31

## react-doctor/no-cascading-set-state

Instead of doing individual `setValue` in effect, it might be better to use `reset`.

```tsx
useEffect(() => {
  const current = getValues();
  if(watchValue) {
     reset({
        ...current,
        newValue: null,
     });
  }
}, [watchValue, reset, getValues)
```

## react-doctor/rerender-memo-with-default-value

```tsx
// Don't do this
function Component({ default = [] })

// Do this
const EMPTY_LIST = [];
function Component({ default = EMPTY_LIST })
```

## react-doctor/no-array-index-as-key

Just use some stable unique value that is not the array index.

```tsx
// Don't do this
{items.map((item, index) => <Component key={index} />)}

// Do this
{items.map((item) => <Component key={item.id} />)}
```

## react-doctor/no-usememo-simple-expression

Just don't memoize this one. It is a net perf loss here.

### ✅ Acceptance Criteria

- [ ] All issues addressed
- [ ] No functionality changes

esizer:
🤖 Resolves #16006 

## 👋 Introduction

Fixes some warnings that came from a run of `react-doctor`.

## 🕵️ Details

- Removed all default values as they were unecessary
- Fixed some issues with `setState` calls in `useEffect`
- Removed an unecessary call to `useMemo`
- Replaced an array index as key

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Login as `admin@test.com`
3. Go to the pages that are afffected
4. Confirm they all still function as expected

### DIFF
--- a/apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/AssessmentDetailsDialog.tsx
+++ b/apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/AssessmentDetailsDialog.tsx
@@ -252,7 +252,7 @@ const AssessmentDetailsDialog = ({
     handleSubmit,
     control,
     watch,
-    getValues,
+    setValue,
     reset,
     formState: { errors },
   } = methods;
@@ -302,23 +302,14 @@ const AssessmentDetailsDialog = ({
   });
 
   useEffect(() => {
-    const values = getValues();
     if (dialogMode === "regular") {
-      reset({
-        ...values,
-        screeningQuestionFieldArray: [],
-        assessmentTitleEn: values.assessmentTitleEn,
-        assessmentTitleFr: values.assessmentTitleFr,
-      });
-    } else if (dialogMode === "screening_question") {
-      reset({
-        ...values,
-        assessmentTitleEn: null,
-        assessmentTitleFr: null,
-        screeningQuestionFieldArray: values.screeningQuestionFieldArray,
-      });
+      setValue("screeningQuestionFieldArray", []);
     }
-  }, [dialogMode, reset, getValues]);
+    if (dialogMode === "screening_question") {
+      setValue("assessmentTitleEn", null);
+      setValue("assessmentTitleFr", null);
+    }
+  }, [dialogMode, setValue]);
 
   // NOTE: Required to update form when initial values change
   useEffect(() => {


apps/web/src/pages/TalentNominations/NominationGroup/components/NominatorList.tsx
esizer: I'm not super happy with this. However, I was hesitant to anything more impactful.

While this is probably not impacting us much, I feel as though we could probably find a more idiomatic pattern to use here 🤔

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/AssessmentDetailsDialog.tsx
```diff
@@ -302,14 +302,23 @@ const AssessmentDetailsDialog = ({
   });
 
   useEffect(() => {
+    const values = getValues();
```
petertgiles: Sorry, I still think this one makes the code less maintainable.  Having to remember to include unrelated fields in a given reset call is a trap imo.  I think the original code is fine.

apps/web/src/pages/TalentNominations/NominationGroup/components/NominatorList.tsx
petertgiles: I like this!  :+1:

PR #16054, opened 2026-03-04
[Fix] Close pagination page size menu on click

### LINKED ISSUE #16043: 🐛 UAT | Performance latency and focus lost observed on candidate/user/process tables

Two issues were identified when adjusting the row count on the Candidate, User, and Process tables:
1. Focus Loss: Changing the "Show" result count (e.g., from 10 to 50 or 500) causes the table to lose focus, disrupting the user's navigation flow.
2. Performance Latency: Loading 500 candidate records in the UAT environment takes approximately 20 seconds, which may exceeds performance standards.

Expected outcome: The table should maintain focus during state changes, and datasets of up to 500 records should load within a few seconds to ensure a seamless user experience.

### 🕵️ Details

Env: UAT

### 📋 Steps to Reproduce

2. Navigate to Candidates table
3. Click on Showing result and select Show 500
4. Observe the page
5. The page takes around 20 seconds to load the 500 candidate data
6. The focus is lost somewhere in the page (Have to click on the page to get the focus back)
7. Repeat the steps on Users table and process table and observe the page

### 📸 Screenshot

https://github.com/user-attachments/assets/f768b8fc-142f-4d8b-ba4f-3b3662363ef1

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] The focus should not be lost when updating the show result on any table
- [ ] It should not take more time to load 500 candidates May be?!!

### 🔄 Regression

Yes

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #16043 

## 👋 Introduction

Updates the page size dropdown menu to close when an item inside is clicked. This is to avoid confusion when they dropdown gets pushed out of the viewport after a layout shift.

## 🕵️ Details

SEE: https://github.com/GCTC-NTGC/gc-digital-talent/issues/16043#issuecomment-3997352647

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Login as `admin@test.com`
3. Go to any table
4. Select a different page size
5. Confirm it closes on a single click and you are free to interact with the page

### DIFF
--- a/apps/web/src/components/Pagination/Pagination.tsx
+++ b/apps/web/src/components/Pagination/Pagination.tsx
@@ -168,11 +168,7 @@ const Pagination = ({
                   onValueChange={handlePageSizeChange}
                 >
                   {pageSizes.map((size) => (
-                    <DropdownMenu.RadioItem
-                      closeOnClick
-                      key={size}
-                      value={String(size)}
-                    >
+                    <DropdownMenu.RadioItem key={size} value={String(size)}>
                       {size}
                     </DropdownMenu.RadioItem>
                   ))}
--- a/packages/ui/src/components/DropdownMenu/DropdownMenu.tsx
+++ b/packages/ui/src/components/DropdownMenu/DropdownMenu.tsx
@@ -108,11 +108,13 @@ interface ItemProps extends ItemVariants, Omit<Menu.Item.Props, "className"> {
 const Item = ({
   className,
   color = "primary",
+  closeOnClick = true,
   disabled,
   ...rest
 }: ItemProps) => (
   <Menu.Item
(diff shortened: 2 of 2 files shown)


mauli0110:
> This is @mauli0110 's issue so I'd like her to confirm but it seemed like there was consensus in standup today to change the AC. I suggest something like this:
> 
> * [ ]  Original behaviour of dropdowns is restored to pre-baseui change.
>   
>   * [ ]  especially click-to-close
> * [ ]  API performance is out of scope

I agree with @petertgiles, I would like to verify the change in local now to confirm the updated ACs. Will be back to confirm in few mins

PR #16167, opened 2026-03-18
[Bug] Enable cache invalidation for flags and bookmarks

### LINKED ISSUE #16166: 🐛 UAT | Flagged candidate is not showing as flagged in candidate application page

### 🕵️ Details

Env: UAT

### 📋 Steps to Reproduce

### 📸 Screenshot

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

petertgiles:
🤖 Resolves #16166 

## 👋 Introduction

Fixes the issue of keeping a stale view of flags and bookmarks by returning a pool candidate from the API instead of a simple boolean.

:+1: Doesn't allow the flag and bookmark state to go stale
:-1: Causes rows to jump around the table

> [!CAUTION]
> If we go with this solution we should probably simplify the optimistic UI logic in useCandidateFlagToggle and useCandidateBookmarkToggle since it might become unnecessary.

### DIFF
--- a/api/app/GraphQL/Mutations/TogglePoolCandidateFlag.php
+++ b/api/app/GraphQL/Mutations/TogglePoolCandidateFlag.php
@@ -15,6 +15,6 @@ final class TogglePoolCandidateFlag
         $candidate->is_flagged = ! $candidate->is_flagged;
         $candidate->save();
 
-        return $candidate->is_flagged;
+        return $candidate;
     }
 }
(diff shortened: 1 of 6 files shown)


vd1992:
> I will demo for them sometime

Update?

petertgiles:
image

brindasasi:
@esizer @vd1992 any idea on this ..  has this been fixed ?

esizer:
> @esizer @vd1992 any idea on this .. has this been fixed ?

We needed further input, the fix is to have it update automatically. However, doing so means the rows will re-order after flagging something. However, a requirement was that the rows do not re-order.

So, I'm not sure what we can do here :sweat_smile:

tristan-orourke:
Replaced by a different approach, in #17774

PR #16218, opened 2026-03-23
Revert "[Fix]  Dialog.Footer is nested inside Dialog.Body"

mnigh:
Reverts GCTC-NTGC/gc-digital-talent#15688


mnigh:
> [!IMPORTANT]
>  I don't agree that the Footer should be a child of the Body and nor does AI as it is not a common pattern for UI and there is no documentation for using the component in this unique pattern, but 🤷

PR #16227, opened 2026-03-24
[Refactor] Playwright CI config, add more shards

esizer:
## 👋 Introduction

Updates the playwright CI config to spread workers out a bit more for better run times.

## 🕵️ Details

Currently we are only using 4 shards, 2 per project (chromium, webkit). This just doubles that so that we are running 4 per project for better balancing and almost half the runtime.

I beleive the downside is, each shard takes up an active action slot so it is a balancing act. Is taking up an extra 4 slots with the halved runtime? :woman_shrugging: 

## 🧪 Testing

1. Confirm CI passes
2. Determine if the runtime is worth the extra workers running or not


mauli0110 (approved):
Wow significant change in execution, thanks @esizer 
LGTM !

Image

PR #16382, opened 2026-04-07
[Fix] Legacy status down migration

esizer:
🤖 Resolves 

## 👋 Introduction

Missed adding `referring` column back in the down after dropping it in the up.

## 🧪 Testing

1. Ruyn migration a few steps fdoward and backward
2. Confirm it works


brindasasi (approved):
oops ..good to go

PR #16430, opened 2026-04-10
[Feat] Replace meta app url on post deploy

### LINKED ISSUE #9206: ♻️ Subtitute environment variables into index.html on post-deploy

esizer:
🤖 Resolves #9206 

## 👋 Introduction

Updates post deployment to substitute `APP_URL` in the `index.html` on post deployment.

## 🕵️ Details

This is to have our metadata such as open graph and twitter to use the appropriate vertical URL.

> [!NOTE]
> This actually breaks it locally since it required us to hardcode `$APP_URL` rather than use a fallback. But, that is likely an okay tradeoff? :thinking:  

## 🧪 Testing

> [!IMPORTANT]
> This requires verifying it on the dev server. 

1. Navigate to any page on the app in the dev version
2. Confirm the metadata in the head uses the dev URL and not the prod one


petertgiles:
Very weird.  I can't get this to deploy properly on the dev vertical.  The logging shows that it fails to prepare the cache directory and yet when I SSH in I can successfully run the same command.  I tried deploying IaC to an older version and that had the same problem so it doesn't seem like it's that.  I tried deploying the same release candidate that's on UAT and that was successful.  So either it's something in this PR or something in the base branch after the UAT cutoff?  :shrug: Do you have any idea what's going wrong?


Also, looking at this existing process again, I'm pretty uncomfortable with the security of allowing any env variable to get dumped into files that get shipped down to the browser.  What if we accidentally pasted in a var for the db password or the auth secret.  :skull: Not part of this issue but I really think we need to rethink this.  I like Vite's requirement to prefix the variable before it can get substituted in.  That might be an approach.

petertgiles:
Security concerns: https://github.com/GCTC-NTGC/gc-digital-talent/issues/16469

petertgiles (approved):
Nice, this is working for me now.  I'm not sure what I did wrong before, but thanks for double-checking anyway.  I'm not worried about these meta values not working outside of Azure.

PR #16439, opened 2026-04-10
[Feature] Store last sign in ISS

### LINKED ISSUE #16426: ✨ Track last sign in type
### Details

We have a number of use cases where we need to know if someone has moved over the CanadaLogin yet or not.  We can do this by storing which system the user last used to log in.

### Design file & designer

_No response_

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

When a user signs in we save the "last_signed_in_at" timestamp.  We can make a new column to tag at the same time. Would we store the current AUTH_SERVER_ROOT?  Or maybe `iss` from the token?  Or a new ENV?

### ✅ Acceptance Criteria

- [ ] We can track if a user has moved to CanadaLogin yet
- [ ] Also works with migrated users

petertgiles:
🤖 Resolves #16426 

## 👋 Introduction

Stores the iss (issuer) values for the last sign in to help us know whether someone has used CanadaLogin yet or not.

## 🕵️ Details

This is for future use, like a possible in-app migration tool or targeted emails.  For now, it is not exposed in the API at all.

I did check the tokens of a migrated account on UAT and the iss is `https://auth.login-connexion.alpha.canada.ca/oauth2`.

## 🧪 Testing

1. Run the migration
- [ ] Logging in with SIC stores a SIC value
- [ ] Logging in with CL stores a CL values

## 📸 Screenshot

image

### DIFF
--- a/api/app/Http/Controllers/AuthController.php
+++ b/api/app/Http/Controllers/AuthController.php
@@ -163,6 +163,7 @@ class AuthController extends Controller
 
         // update the user with values from logging in
         $userMatch->last_sign_in_at = Carbon::now();
+        $userMatch->last_sign_in_iss = $idToken->claims()->get('iss', null);
         if ($idToken->claims()->has('given_name')) {
             $userMatch->first_name = $idToken->claims()->get('given_name');
         }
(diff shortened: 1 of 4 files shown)


esizer (approved):
LGTM! 👍 

Image

PR #16442, opened 2026-04-10
[Bug] Make taken emails valid for API

### LINKED ISSUE #16428: 🐛 Taken emails are invalid

When we have an account take over the email from another account it mangles it in the older account.  Oops, that mangled version is no longer a valid email and crashes the API.

### 🕵️ Details

https://github.com/GCTC-NTGC/gc-digital-talent/blob/d8a13cd7e2fc224957d74cca31dc2746bbdcde62/api/app/Http/Controllers/AuthController.php#L164

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

Check how the archived user mangling happens.  Probably the mangling happens before the @ sign.

### ✅ Acceptance Criteria

- [ ] Users who have their email taken can still be sent through the API

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

petertgiles:
🤖 Resolves #16428 

## 👋 Introduction

Mangles the taken email addresses in a way that still allows them to pass through the API.

## 🧪 Testing

1. Configure your environment for CanadaLogin
2. Manually update some other user with the same email address you use with CL
3. Log in with your CL user
4. Grant your CL user admin privileges, maybe with Tinker :point_down: 
```php
$user = User::where('email', 'username@domain.tld')->sole();
$user->addRoles(['base_user', 'applicant', 'platform_admin']); 
```
5. Try to load the other user using the admin pages
- [ ] Your CL user has the email address properly set
- [ ] The other user can be loaded in the app without an error
- [ ] The email address is mangled with at "taken-at" timestamp

## 📸 Screenshot

image

### DIFF
--- a/api/app/Http/Controllers/AuthController.php
+++ b/api/app/Http/Controllers/AuthController.php
@@ -172,26 +172,28 @@ class AuthController extends Controller
         if ($idToken->claims()->has('email')) {
             $incomingEmailAddress = $idToken->claims()->get('email');
 
-            // if existing users have this email address then take it from them
-            try {
-                $existingUser = User::where('sub', '!=', $sub)
-                    ->where(fn ($subquery) => $subquery
-                        ->where('email', 'ilike', $incomingEmailAddress)
-                        ->orWhere('work_email', 'ilike', $incomingEmailAddress)
-                    )->first();
-                if (strcasecmp($existingUser->email, $incomingEmailAddress) == 0) {
-                    $existingUser->email = $existingUser->email.'-taken-at-'.Carbon::now()->timestamp;
+            // if email is not null, check if existing users have this email and take it from them
+            if (! empty($incomingEmailAddress)) {
+                try {
+                    $existingUser = User::where('sub', '!=', $sub)
+                        ->where(fn ($subquery) => $subquery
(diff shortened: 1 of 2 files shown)


esizer (commented):
Works well for me! Just one thing I noticed that might not be an issue; if the user who took your emial has a `null` sub, the existing user query doesn't return so the user is null and their email never gets updated.

I doublt anyone has a `null` sub, but maybe we should handle that case? 🤔

petertgiles:
Good catch, I forgot that sub can be null.  I changed it to disambiguate by ID.
[distinguish by ID](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16442/commits/7797087f7e029f2809c10853361cb30300065c65)

It also reminded me to add a check for null email addresses.  (The diff in GH makes it look like  a much bigger change than it actually is. :sweat_smile: )
[empty email check](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16442/commits/2f150958d5905dfed715156265573e2197023d3f)

esizer (approved):
Works great! 👍

PR #16447, opened 2026-04-13
[Debt] Enable verbatim module syntax packages

### LINKED ISSUE #16445: 🛠️ Enable `verbatimModuleSyntax` for `packages/*`
### 🛠️ Platform

Enables the `verbatimModuleSyntax` TS compiler option in preperation to migrate ot vite 8.

### 🕵️ Details

This setting is rerquired but has a large diff so we are doing it before vite 8 upgrade.

### 🙋‍♀️ Proposed Solution

Enabled `verbatimModuleSyntax` and the eslint rule. Then, run automated eslint fix.

### ✅ Acceptance Criteria

- [ ] `verbatimModuleSyntax` enabled for `packages/*`
- [ ] Consistent type imports eslint rule enabled
- [ ] All errors fixed

esizer:
🤖 Resolves #16445 

## 👋 Introduction

Enables the `verbatimSyntaxModules` in `packages/*` and fixes the errors.

## 🧪 Testing

1. Confirm no lint errors
2. Confirm everything builds


apps/playwright/tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "esModuleInterop": true,
     "strictNullChecks": true,
     "skipLibCheck": true,
+    "verbatimModuleSyntax": false,
```
mnigh, line 10: Why `false`?

apps/web/tsconfig.json
```diff
@@ -6,7 +6,8 @@
       "~/*": ["./src/*"],
       "./+types/*": ["./.react-router/types/src/+types/*"]
     },
-    "rootDirs": [".", "./.react-router/types"]
+    "rootDirs": [".", "./.react-router/types"],
+    "verbatimModuleSyntax": false
```
mnigh, line 10: Why `false`?

packages/app-insights/package.json
```diff
@@ -11,6 +11,7 @@
     "prettier": "prettier --check .",
     "prettier:fix": "prettier --write .",
     "lint": "eslint",
+    "lint:fix": "eslint --fix",
```
mnigh, line 14: Any reason we didn't have `lint:fix` before?

mnigh (commented):
LGTM, just a few questions.

apps/playwright/tsconfig.json
esizer: This is so we don't affect this folder to keep the diff isolated to `./packages/*`

apps/web/tsconfig.json
esizer: Same as the playwright one. We only want `./packages/*` afffected.

packages/app-insights/package.json
esizer: I'm not sure :thinking: Most packages were missing this so I went ahead and added them to make things eaiser :sweat_smile:

mnigh (approved):

PR #16457, opened 2026-04-13
Community admins can view the edit community page

### LINKED ISSUE #16458: 🐛 Community admin can't edit community info

As a community admin , I should be able to edit community information and manage roles as per this 

Image

But currently I get unauthorized error.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

 [ ] Community admin should be able to  update community info 
 [ ] process operator , community recruiter , community talent co-ordinator should not be able to update the community info 
 [ ] generate unit tests for each role

vd1992:
🤖 Resolves #16458 

## 👋 Introduction

Add the missing role to the roles array

## 🧪 Testing

1. As a community admin only
2. Can edit your community

### DIFF
--- a/apps/web/src/pages/Communities/UpdateCommunityPage/UpdateCommunityPage.tsx
+++ b/apps/web/src/pages/Communities/UpdateCommunityPage/UpdateCommunityPage.tsx
@@ -412,7 +412,7 @@ export const UpdateCommunity = () => {
 };
 
 export const Component = () => (
-  <RequireAuth roles={[ROLE_NAME.PlatformAdmin]}>
+  <RequireAuth roles={[ROLE_NAME.PlatformAdmin, ROLE_NAME.CommunityAdmin]}>
     <UpdateCommunity />
   </RequireAuth>
 );


brindasasi (changes requested):
fixes that now community admin can update their own community
Image
however this change opens up the info page for community admins in edit mode for all other communities they don't belong to.
It doesn't allow save though. So that is still ok. 
May be we create a sep issue to block the edit mode? 

another thing please add some unit tests as mentioned in A/C.

vd1992:
> however this change opens up the info page for community admins in edit mode for all other communities they don't belong to. It doesn't allow save though. So that is still ok. May be we create a sep issue to block the edit mode?

Yeah, constraining things by role _**and**_ team is a problem that pops up in many places. 
A recent refactor/feature introduces a new pattern to follow in https://github.com/GCTC-NTGC/gc-digital-talent/pull/15797 We have yet to implement it widely. 

So I have spun off this issue https://github.com/GCTC-NTGC/gc-digital-talent/issues/16461

vd1992:
I am not certain what use unit tests would provide at this point to be honest. They would be more useful added in https://github.com/GCTC-NTGC/gc-digital-talent/issues/16461 to assert the role + team checking works

brindasasi (approved):

PR #16482, opened 2026-04-15
[Feat] Dev docker container

### LINKED ISSUE #9844: 🛠️ Separate docker-compose environment for development
## 🛠️ Tooling

We have a docker container for our webserver that is very similar to our prod environment which is good for testing and verification.  However, it limits some exciting dev features like building the app in the container setup,  in-container watch, HMR, or GH dev containers.  Let's set up a separate docker-compose file for development.  This could have Laravel running in a container on its dev server and Webpack running in a different container on its dev server.  Perhaps this could remove the need for a separate maintenance container.

## 🕵️ Details

Some known challenges:
- CORs on API requests is not properly configured
- we have a lot of logic that assumes the API is running on the same host as the web server
- might need a reverse proxy container to handle routing from the browser

## ✅ Acceptance Criteria

- [ ] Separate docker-compose file for development
- [ ] Laravel serving the API from it's own container
- [ ] Webpack serving pages from it's own container
- [ ] Original docker-compose network remains unchanged

esizer:
🤖 Resolves #9844 

## 👋 Introduction

Adds a dev docker container for local development.

## 🕵️ Details

Adds a new docker compose file along with docker files and config. 

 - `api.Dockerfile` is a slim PHP container that has laravel running their dev server and starts background tasks
 - `web.Dockerfile` is a slim node container that runs vites dev server
 - `docker-compose.dev.yml` extends our base docker compose file but overwrites some values to work with the dev servers and adds the nginx proxy to route traffic properly

## 🧪 Testing

> [!TIP]
> This is a little janky. On initial start up you might see a 502 while the dev servers boot. Also, as awlays, watch modes arent super smooth when chaning graphql or importing new components. 

1. Shutdown existing container `make down`
2. Start up dev container `make up ENV=dev`
3. Navigate to app at `http://localhost:8000`
4. Confirm everything is sort of smooth (its a little janky :sweat: )


infrastructure/dev/default
esizer, line 1: I really don't like the idea of now needing to maintain 3 nginx confs 😢 

I wonder if we can use `include` to group similar rules across both local configs 🤔

infrastructure/dev/web.Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+FROM node:24.12.0-slim
+
+# Install git for version info in builds
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends git \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install pnpm globally
+RUN npm install -g pnpm@10.27.0
```
esizer, line 13: Also, not sure if we should be hardcoding values here but we were in the other one so 🤷‍♀️

infrastructure/dev/README.md
```diff
@@ -0,0 +1,34 @@
...
+make down ENV=dev    # Stop
+```
+
+Or using docker compose directly:
+
+```bash
+docker compose -f docker-compose.dev.yml up --build
+```
+
+Access the application at: http://localhost:8000
```
esizer, line 28: I stuck with the same port but maybe that isn't great? 🤔  You can't really run both at the same time (or shouldnt) so maybe not an issue?

infrastructure/dev/api.Dockerfile
```diff
@@ -0,0 +1,40 @@
+# Development Dockerfile for Laravel API
+# Uses PHP's built-in dev server with hot reloading
+
+FROM php:8.4-cli
```
esizer, line 4: This uses a diffferent base image. Not sure if this is an issue for dev work or if we should be using the full image from the other one 🤔

petertgiles:
I just started looking at this but I can't get it to build.  Am I doing something wrong?
image

infrastructure/dev/default
petertgiles: Agreed.  This new one is so different, though, that I'm not sure there's much consolidation that can be done.

infrastructure/dev/web.Dockerfile
petertgiles: This doesn't bother me.

infrastructure/dev/README.md
petertgiles: This doesn't bother me.  Actually I think I prefer it.

esizer:
> I just started looking at this but I can't get it to build. Am I doing something wrong? image

Do you have the other containers running? They defintley conflict so you might need to do: `docker compose down` first then bring the new ones up

infrastructure/dev/api.Dockerfile
petertgiles: I'd prefer if we could reuse an image we're already using.  The Ubuntu one that the maintenance container uses doesn't have the right version of PHP and the webserver is heavily configured.  :face_exhaling:   Maybe it doesn't matter since this is an optional compose network.

petertgiles:
No, nothing else running.
image

esizer:
> No, nothing else running. image

Okay, unfortunatley I was unable to reproduce that but I don't see a need for the adminer block at all so I removed it in [remove adminer overrides entirely](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16482/commits/bd90af96835f6bd1c565c8ac31ffc42d3c0e8530)

Maybe that addresses it? :sweat_smile: 

> [!NOTE]
> I also noticed that there was a hydration error due to the CSP nonce which imm looking into

infrastructure/dev/default
petertgiles: What if we really strip this one down?  I think we don't really need CSP or legacy redirects in this file.  It could be much simpler.

petertgiles:
> Maybe that addresses it? 😅

Sorry, same error but now it's complaining about the webserver service.
> services.webserver conflicts with imported resource

infrastructure/dev/default
esizer: Okay, that makes alot of sense to me! I removed quite a bit and cleaned it up in general in [clean up dev nginx config](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16482/commits/10c22d9d36c51e74aa519bdc302b009aec5af556)

petertgiles:
A different error!  :sweat_smile: 

> service "nginx-proxy" depends on undefined service "mock-auth": invalid compose project

petertgiles:
Here's where I'm at this morning.
> service "api" depends on undefined service "postgres": invalid compose project

petertgiles:
Hey, docker started this time!

It looks like there's an error related to Reverb in the api container but it did start.
image

The web container seems to be failing with errors from pnpm though.
image

esizer:
> It looks like there's an error related to Reverb in the api container but it did start.

Looks like [reverb requires pcntl installed](https://laracasts.com/discuss/channels/reverb/reverb-installation-issue). 

Added in [install missing dep](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16482/commits/f96ddea4ac22ff6d3bd4ca764be20d5f956fb911)
 
> The web container seems to be failing with errors from pnpm though.

I think this is because it is asking for confirmation (to overrwrite dependencies on install) but there is no TTY inside the container so I added a `--force` in [force install deps](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16482/commits/a6e8f224a9213b95fe0f665a302db6bc583177ab)

petertgiles (approved):
Nice, thanks for this!  I got it up and running and it seems good.  Yeah, a bit janky and funny but I'm looking forward to really trying it out on a PR.  🏅

PR #16512, opened 2026-04-16
[Platform] Increase FPM max children, spare servers

### LINKED ISSUE #16126: 🦔 How can we get more performance from PHP-FPM
### 🦔 Spike

PHP-FPM is a big program with a lot of configurability.  We haven't spent much time trying to optimize it though.

### ❓ Questions to answer

Are there any ways we can improve our performance on PHP-FPM through configuration?

### 🕙 Timebox

3 days

esizer:
🤖 Resolves #16126 

## 👋 Introduction

Increases our PHP-FPM max children and spare servers to better utilize resources.

## 🕵️ Details

We have more resources to spare on the server and while we rarely run out of workers, we have room to add more so it should never happen even under high loads.

## 🧪 Testing

> [!NOTE]
> Likely need to deploy this to truly test it.

1. I guess just reload FPM and run some load tests?

### DIFF
--- a/infrastructure/conf/php-fpm-www.conf
+++ b/infrastructure/conf/php-fpm-www.conf
@@ -124,22 +124,22 @@ pm = dynamic
 ; forget to tweak pm.* to fit your needs.
 ; Note: Used when pm is set to 'static', 'dynamic' or 'ondemand'
 ; Note: This value is mandatory.
-pm.max_children = 50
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
Nice.  I was able to deploy it to dev and, at least idling, resource usage seems about the same which is expected.

Image

PR #16519, opened 2026-04-16
[Platform] Increase auth cache times

### LINKED ISSUE #16518: ♻️ Increase auth caching
### ♻️ Refactor

We would like to increase the auth cache time for our oauth calls.

### 🕵️ Details

We started with very conservative values to not over cache these values but, we have ran into timeout issues due to too many outbound connections when we get a surge in traffic. We should be safe to increase these values to limit outbound connections.

 - Config (`getConfiguration`, `getConfiguationKey`)
   -  Some keys auto rotate but we currently only cache for 60s
   - auto rotation usually happens in terms of days to monts so,
   - we can increase this value singificantly
 - Introspection
   -  Our tokens live for 10mins but we only cache for 10s or the remaining lifetime, whatever is lower.
   - Since our tokens live for 10mins, we can safely increase this value

### 🙋‍♀️ Proposed Solution

Increase cache time in our token services:

 - `getConfiguration`: 30m - 1h
 - `getConfigurationKey`: 30m to 1h
 - `verifyJwtWithIntrospection`: 2m or remaining lifetime, whatever is lower

### ✅ Acceptance Criteria

- [ ] Cache times increased in both token services (CanadaLogin, SiC)

esizer:
🤖 Resolves #16518 

## 👋 Introduction

This significantly increases the time we cache auth checks for to limit the number of outbound connections.

## 🕵️ Details

This addes two env vars with default values:

 - `OAUTH_CONFIG_CACHE_TIME` 
   - Scope: Controls the cache for the open id config (should rarely change)
   - Default: 30m
- `OAUTH_INTROSPECTION_CACHE_TIME`
   - Scope: Controls the time we cache the introspection response
   - Default: 2m

Since the config values chould rarely change, we can set this to a much higher value. We are starting conservative at 30m but we can alwayus increase as needed via the env var.

The introspection was 10s burt we increased to 2m. These should be checked sort of frequently to confirm the token is still valid. However, our tokens should be living for 10m so we dont need to check every 10s. We also have logic to cache between the 10m and the remaining lifetime of the time, whatever is the smaller value. So, when we check, we basically say "don't check again for 2 mins or once the token expires." which should be safe. 
## 🧪 Testing

> [!TIP]
> You can test this locally but our cache does not work since the mock oauth does not return exp time in the introspection response.

1. Login
2. Hit some endpoints for more than 2 mins
3. Confirm the cache occures are there are not more than 3 introspection requests (initial, another after cache)

## 🚚 Deployment

Add the following env vars (optional):

```ini
OAUTH_CONFIG_CACHE_TIME=1800
OAUTH_INTROSPECTION_CACHE_TIME=120
```

### DIFF
--- a/api/app/Services/CanadaLoginBearerTokenService.php
+++ b/api/app/Services/CanadaLoginBearerTokenService.php
@@ -62,7 +62,7 @@ class CanadaLoginBearerTokenService implements BearerTokenService
     private function getConfigProperty(string $propertyName): string
     {
         // only get content every so often (default: 30min)
-        $jsonString = Cache::remember('openid_config_json_string', config('oauth.config_cache_timea'),
+        $jsonString = Cache::remember('openid_config_json_string', config('oauth.config_cache_time'),
             function () {
                 $response = Http::retry(times: config('oauth.request_retries'), sleepMilliseconds: 500, when: function (Exception $exception) {
                     return $exception instanceof ConnectionException;
--- a/api/app/Services/CanadaLoginBearerTokenService.php
+++ b/api/app/Services/CanadaLoginBearerTokenService.php
@@ -61,20 +61,22 @@ class CanadaLoginBearerTokenService implements BearerTokenService
     // get a configuration property from the openid configuration json document
     private function getConfigProperty(string $propertyName): string
     {
-        $jsonString = Cache::remember('openid_config_json_string', 60, function () { // only get content every minute
-            $response = Http::retry(times: config('oauth.request_retries'), sleepMilliseconds: 500, when: function (Exception $exception) {
-                return $exception instanceof ConnectionException;
-            }, throw: false)->get($this->configUri);
-            assert($response instanceof Response);
+        // only get content every so often (default: 30min)
+        $jsonString = Cache::remember('openid_config_json_string', config('oauth.config_cache_timea'),
+            function () {
+                $response = Http::retry(times: config('oauth.request_retries'), sleepMilliseconds: 500, when: function (Exception $exception) {
+                    return $exception instanceof ConnectionException;
+                }, throw: false)->get($this->configUri);
+                assert($response instanceof Response);
 
-            if ($response->failed()) {
-                Log::error('Failed when GETting the OpenID configuration in getConfigProperty');
-                Log::debug((string) $response->getBody());
-                throw new Exception('Failed to get config');
-            }
+                if ($response->failed()) {
+                    Log::error('Failed when GETting the OpenID configuration in getConfigProperty');
+                    Log::debug((string) $response->getBody());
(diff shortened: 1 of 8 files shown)


api/app/Services/CanadaLoginBearerTokenService.php
```diff
@@ -61,20 +61,22 @@ public function __construct(string $configUri, ClockInterface $clock, DateInterv
...
-            if ($response->failed()) {
-                Log::error('Failed when GETting the OpenID configuration in getConfigProperty');
-                Log::debug((string) $response->getBody());
-                throw new Exception('Failed to get config');
-            }
-
-            return $response->body();
-        });
+        // only get content every so often (default: 30min)
+        $jsonString = Cache::remember('openid_config_json_string', config('oauth.config_cache_timea'),
```
petertgiles:
```suggestion
        $jsonString = Cache::remember('openid_config_json_string', config('oauth.config_cache_time'),
```

api/app/Services/SignInCanadaBearerTokenService.php
```diff
@@ -59,7 +59,7 @@ public function __construct(string $configUri, ClockInterface $clock, DateInterv
     // get a configuration property from the openid configuration json document
     private function getConfigProperty(string $propertyName): string
     {
-        $jsonString = Cache::remember('openid_config_json_string', 60, function () { // only get content every minute
+        $jsonString = Cache::remember('openid_config_json_string', config('oauth.config_cache_time'), function () { // only get content every minute
```
petertgiles:
```suggestion
        $jsonString = Cache::remember('openid_config_json_string', config('oauth.config_cache_time'), function () { // only get content every so often (default: 30min)
```

api/app/Services/SignInCanadaBearerTokenService.php
```diff
@@ -90,7 +90,7 @@ private function getConfiguration(string $keyId): Configuration
         }
 
         $jwks_uri = $this->getConfigProperty('jwks_uri');
-        $jsonString = Cache::remember('jwks_json_string', 60, function () use ($jwks_uri) { // only get jwks content every minute
+        $jsonString = Cache::remember('jwks_json_string', config('oauth.config_cache_time'), function () use ($jwks_uri) { // only get jwks content every minute
```
petertgiles:
```suggestion
        $jsonString = Cache::remember('jwks_json_string', config('oauth.config_cache_time'), function () use ($jwks_uri) { // only get jwks content every so often (default: 30min)
```

petertgiles (changes requested):
Thanks for doing this so fast.  I flagged a few things to tidy then I'll try rolling it out on dev.

petertgiles (approved):
Deployed to dev and seems to behave normally.  👍

PR #16521, opened 2026-04-16
[Fix] `vite` version mismatch

### LINKED ISSUE #16520: 🛠️ Version mismatch in `vite`
### 🛠️ Platform

We need to update the vite version so they are both on the same version.

### 🕵️ Details

We recently upgraded to vite 8 but only updated the `app/web` version and not the `packages/vitest-helpers` version.

### 🙋‍♀️ Proposed Solution

Update `vitest-helpers` dependencies to use the same `vite` verison as `apps/web`

### ✅ Acceptance Criteria

- [ ] Same `vite` version everywhere

esizer:
🤖 Resolves #16520 

## 👋 Introduction

Fixes the version mismatch for vite.

## 🧪 Testing

1. Confirm only one version of vite is being used
2. Confirm `vitest` still functions as expected


packages/vitest-helpers/src/config/ui-config.mts
```diff
@@ -1,10 +1,9 @@
-import { defineProject, mergeConfig } from "vitest/config";
-
+import { defineConfig, mergeConfig } from "vitest/config";
 import baseConfig from "./base-config.mjs";
```
mnigh: Should this be `base-config.mts`?

packages/vitest-helpers/src/config/index.mts
mnigh, line 1: Should this be `base-config.mts`?

packages/vitest-helpers/src/config/ui-config.mts
esizer: Oops, I meant to make this a draft :sweat_smile: That should be addressed now after [fix lint issues](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16521/commits/bd8ba36bb00985676a39723a1172f2e2d7507696)

mnigh (approved):

PR #16532, opened 2026-04-20
[Fix] `ResponsiveTable` search param race condition

### LINKED ISSUE #15883: 🐛 Type Error on Talent Requests page in prod

**Uncaught TypeError: Cannot read properties of null (reading 'removeChild')**

Source URL : 
https://talent.canada.ca/en/admin/talent-requests/7eb5853c-9229-44f4-a795-76c31d784f36

### 🕵️ Details
I see a lot of these errors on Talent Requests page on prod. 

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

[ ] No type error on Talent Requests page
### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #15883 

## 👋 Introduction

Addresses an issue in the `ResponsiveTable` when navigating using the browser back button.

## 🕵️ Details

When you have search params in the browser history and the current page and navigated back, react was comparing the current URL params (which were changing) to determine if it should re-render.  The issue was, during navigatioin we had two separate search params (before and after navigaiton). They were racing eachother to replace thge URL state.

The losing one would see nothing to replace (old params were gone) hence the `removeChild` error.

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Login as `admin@test.com`
3. Navigate to `/admin/talent-requests`
4. Add some filters to updarte thr URLSearchParams
5. Click on a request
6. Scroll to the candidates table
7. Modify the filters tol up the URLSearchParams
8. Press the back button in the browser
9. Confirm no console errors for `removeChild`
10. Confirm the loading spinner does not hang

### DIFF
--- a/apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
+++ b/apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
@@ -95,7 +95,6 @@ const ResponsiveTable = <TData extends object, TFilters = object>({
 }: TableProps<TData, TFilters>) => {
   const id = useId();
   const intl = useIntl();
-  const isFirstRender = useRef(true);
   const lastWrittenTableParamsRef = useRef<Record<string, string | null>>({});
   const { announce } = useAnnouncer();
   const hasUpdatedRows = useRef<boolean>(false);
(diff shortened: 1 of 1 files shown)


petertgiles (approved):
I think I understand this fix?  It's whitelisting the keys to change also keeping the old values in a ref?  I tested it and the fix seems to work well.  Some extra comments in the code would be helpful though.

esizer:
> Some extra comments in the code would be helpful though.

Yeah, that makes sense! Added some in [document changes](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16532/commits/119ccf2340e7e672821d04c242751abe5c7dcbe2)

PR #16533, opened 2026-04-20
[Debt] Reduce can root usage

### LINKED ISSUE #16132: ♻️ Investigate reducing usage of `@canRoot`
### ♻️ Refactor

The `@canRoot` directive, when checking a complex policy, is one of the largest issues we have with queries that return lists.

### 🕵️ Details

This policy check runs for each time it is defined in the list. With simple policies, this is fairly inconseqential. The issue is, when:

 - The policy checks are complex
 - We over use the directive
 - There are many items in the list

We typically create two types (`PoolCandidate` v `PoolCandidateAdminView`). This is sort of annoying to maintain but, if it is necessary to maintain okay performance, we should keep doing that.

However, we might want to consider the following:

 - Do we need to use `@can`? Or can it be assumed
 - Are our policy checks overly complex?
 - Do policies cause an n+1 with `loadMissing`

### 🙋‍♀️ Proposed Solution

We care most about the `poolCandidatesPaginatedAdminView` query right now. We also already started this work here so, let's keep the scope small and start looking at the related types:

 - `Pool`
 - `User`
 - `AssessmentStep`
 - `Department`

If these types call `@canRoot`, we might get some noticeable improvements if we can apply the same logic to these types since they are used in the query.

Go through these types and identify any uses of `@canRoot`. For each case determine:

 - Is this required? If not just remove it
 - Is the policy check it is doing complex? If so, can we make it simpler?
 - Does the policy use `loadMissing`? If so, remove it and move it to `@with`
 - If all else fials, we might need a new type 😭 

It might not make sense to do any of these, if so, we can close this ticket. Don't feel like you _have_ to do something.

### ✅ Acceptance Criteria

- [ ] Each `@can` directive on the noted types has been addressed
   - [ ] Each question was at least asked and answered

esizer:
🤖 Resolves #16132 

## 👋 Introduction

This removes some instances of `@canRoot` and `@canResolved`.

## 🕵️ Details

These permission checks can hide n+1 queries which is detrimental when used in nested types. This PR removes 3 instances:

 - Removes `assessmentResults` from the `AssessmentStep` type since it was not being used
 - Removed `pools` from the `Community` type since it was not being used
 - Replaced `@canRoot` with a `whereAuthorizedToView` scope and tests

> [!IMPORTANT]
> The ticket comment mentions `PoolCandidate`. That was not done here due to how disruptive it would be. Separate scope for that refactor created in https://github.com/GCTC-NTGC/gc-digital-talent/issues/16534 

## 🧪 Testing

1. Confirm no instances where we query `AssessmentStep.assessmentResults`
2. Confirm no instances where we query `Community.pools`
3. Confirm `Pool.PoolSkill.assessmentSteps` can be queried
    a. Confirm tests check all happy and non-happy paths

### DIFF
--- a/api/app/Models/AssessmentStep.php
+++ b/api/app/Models/AssessmentStep.php
@@ -224,12 +224,12 @@ class AssessmentStep extends Model
 
             return $query->whereHas('pool', function (Builder $poolQuery) use ($teamIds) {
                 return $poolQuery->where(function (Builder $poolQuery) use ($teamIds) {
-                    $poolQuery->orWhereHas('team', function (Builder $poolQuery) use ($teamIds) {
-                        return $poolQuery->whereIn('id', $teamIds);
-                    })->orWhereHas('community.team', function (Builder $poolQuery) use ($teamIds) {
-                        return $poolQuery->whereIn('id', $teamIds);
-                    })->orWhereHas('department.team', function (Builder $poolQuery) use ($teamIds) {
-                        return $poolQuery->whereIn('id', $teamIds);
+                    $poolQuery->orWhereHas('team', function (Builder $teamQuery) use ($teamIds) {
+                        return $teamQuery->whereIn('id', $teamIds);
+                    })->orWhereHas('community.team', function (Builder $communityQuery) use ($teamIds) {
+                        return $communityQuery->whereIn('id', $teamIds);
+                    })->orWhereHas('department.team', function (Builder $deptQuery) use ($teamIds) {
+                        return $deptQuery->whereIn('id', $teamIds);
                     });
                 });
             });
--- a/api/app/Models/AssessmentStep.php
+++ b/api/app/Models/AssessmentStep.php
@@ -205,7 +205,7 @@ class AssessmentStep extends Model
         });
     }
 
-    public static function withPolicyEagerLoads(Builder $query): Builder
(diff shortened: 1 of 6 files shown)


yonikid15 (commented):
I think you can replace the [assessmentSteps](https://github.com/GCTC-NTGC/gc-digital-talent/blob/16132-reduce-can-root-usage/api/graphql/schema.graphql#L524) `@canRoot` in the Pool as well.

yonikid15 (commented):
Can confirm no instances where we query `AssessmentStep.assessmentResults`
Can confirm no instances where we query `Community.pools`
Can confirm `pool.poolSkills.assessementSteps` can be queried

yonikid15 (changes requested):

esizer:
> I think you can replace the [assessmentSteps](https://github.com/GCTC-NTGC/gc-digital-talent/blob/16132-reduce-can-root-usage/api/graphql/schema.graphql#L524) `@canRoot` in the Pool as well.

That one seemed harmless but I'm a-okay removing more :stuck_out_tongue: Removed in [remove canRoot on pool assessment steps](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16533/commits/40aae07837649be48ead44f1dde5349fd6c880b6)

yonikid15 (approved):
LGTM

PR #16539, opened 2026-04-20
[Feature] Inactivity dialog

### LINKED ISSUE #16301: ✨ Inactivity dialog
### Details

Dialog to warn users about being signed out due to inactivity. 

This is marked as a required UX design element from CanadaLogin

### Design file & designer

@BrookeCameron 

Design file: https://www.figma.com/design/S9DxxwhkrOXyfl54xUAnsF/Inactivity-alert?node-id=2-9 

### 📸 Screenshots

Image

Image

### ✍🏼 Illustrations

Image
Image

### Proposed Implementation

_No response_

### ✅ Acceptance Criteria

- [ ] Light and dark mode version 
- [ ] Mobile responsive design 
- [ ] design review 
- [ ] time countdown
- [ ] Controlled with a feature flag

petertgiles:
🤖 Resolves #16301

## 👋 Introduction

Adds an inactivity dialog after 55 minutes.

## 🕵️ Details

- Controlled by a new feature flag
- Uses a new dependency: [React IdleTimer](https://idletimer.dev/)

## 🧪 Testing

1. Enable the new feature flag
2. Shorten the timeouts in apps/web/src/components/ActivityContext/ActivityContainer.tsx (maybe 10 s and 5 s?)
3. Rebuild the app
- [ ] When not logged in, it never prompts
- [ ] When logged in and remaining active, it never prompts
- [ ] When logged in and idle, the prompt appears
- [ ] When prompted, can stay logged in by closing dialog
- [ ] When prompted, can stay logged in by clicking "stay logged in"
- [ ] When prompted, can logout by clicking "logout"
- [ ] When prompted, can logout by continuing to idle
- [ ] multiple tabs sync up
    - [ ] An idle tab will not force out an active tab
- [ ] When logged in and the feature flag is false, it never prompts

## 📸 Screenshot

image

## 🚚 Deployment

FEATURE_ACTIVITY_TIMER=false (or true to activate immediately)


apps/web/src/components/ActivityContext/InactivityDialog.tsx
```diff
@@ -0,0 +1,136 @@
...
+            </span>
+            <div className="order-2 flex flex-row gap-[--spacing(6*0.3)]">
+              <ClockIcon
+                className="h-6"
+                aria-hidden="false"
+                aria-label={intl.formatMessage({
+                  defaultMessage: "timer",
+                  id: "1Ooizo",
+                  description: "Accessible label for a clock icon",
+                })}
```
esizer: The `aria-label` here is probably not needed. This will read:

> timer time remaining 1 minute

apps/web/src/components/ActivityContext/ActivityContainer.tsx
```diff
@@ -0,0 +1,114 @@
...
+    throttle: 500,
+    crossTab: true,
+    leaderElection: true,
+    syncTimers: 200,
+  });
+
+  useEffect(() => {
+    const interval = setInterval(() => {
+      setRemainingMinutes(Math.ceil(getRemainingTime() / 1000 / 60)); // milliseconds to minutes
+    }, 500);
```
esizer, line 70: Rounding the minutes gets confusing when it is less than 1 min. I thought I had a full minute to continue by session but was logged out before I could click.

Do we want to show seconds once it is <60s remaining? 🤔

apps/web/src/components/ActivityContext/InactivityDialog.tsx
```diff
@@ -0,0 +1,136 @@
...
+    },
+    dark: {
+      xs: pugDarkSm,
+    },
+  };
+
+  return (
+    <Dialog.Root open={open} onOpenChange={onOpenChange}>
+      <Dialog.Content>
+        <Dialog.Header>
```
esizer, line 75: I feel like this should be `AlertDialog` since we require a response from the user and want to alert them something changed.

REF: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alertdialog_role#:~:text=The%20alertdialog%20role%20is%20to%20be%20used%20on%20modal%20alert%20dialogs%20that%20interrupt%20a%20user%27s%20workflow%20to%20communicate%20an%20important%20message%20and%20require%20a%20response.

apps/web/src/components/ActivityContext/InactivityDialog.tsx
```diff
@@ -0,0 +1,136 @@
...
+                  "Your session is about to end due to inactivity",
+                id: "eMMDjW",
+                description: "Title for the inactivity dialog",
+              })}
+            </span>
+          </div>
+        </Dialog.Header>
+        <Dialog.Body>
+          <div className="grid gap-(--text-sm) xs:grid-cols-[3fr_2fr]">
+            <span className="order-1 xs:col-span-2">
```
esizer: Should these `span` tags be `p` instead? The big difference is that `p` allows alterantive navigations to skip over them where as `span` acts more like plain text.

apps/web/src/components/ActivityContext/InactivityDialog.tsx
```diff
@@ -0,0 +1,136 @@
...
+              {intl.formatMessage({
+                defaultMessage:
+                  "If you do not continue your session you will be signed out automatically and unsaved changes will be lost.",
+                id: "PRaqZT",
+                description: "Body for the inactivity dialog",
+              })}
+            </span>
+            <div className="order-2 flex flex-row gap-[--spacing(6*0.3)]">
```
esizer: 🫠 This is so close to just being the `IconLabel` component from `PoolCard`.

The bolding is reveresed but maybe it could use similar size/spacing etc. from that component for consistency? 

https://github.com/GCTC-NTGC/gc-digital-talent/tree/main/packages/ui/src/components/IconLabel

apps/web/src/components/ActivityContext/InactivityDialog.tsx
```diff
@@ -0,0 +1,136 @@
...
+          <div className="flex flex-row gap-3 self-start">
+            <ExclamationTriangleIcon
+              className="h-9 text-warning lg:h-12"
+              aria-hidden="false"
+              aria-label={intl.formatMessage({
+                defaultMessage: "Warning",
+                id: "Yn2mVD",
+                description: "Accessible label for a warning icon",
+              })}
+            />
```
esizer, line 85: This looks a little large, the dialog header uses an `h4` size so maybe we could copy the layout/size styles from that for consistency? 🤔 

As an aside: If we want to add icons into dialog headings, it might make sense to accept it as a prop like `Heading` so we can contorl the size/colour etc there in the future.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/c6d3a00ea539312bce6fdc48a02dbca7f4e0bfc7/packages/ui/src/components/Dialog/Dialog.tsx#L167

https://github.com/GCTC-NTGC/gc-digital-talent/blob/c6d3a00ea539312bce6fdc48a02dbca7f4e0bfc7/packages/ui/src/components/Heading/Heading.tsx#L30-L33

esizer (changes requested):
Cool!

apps/web/src/components/ActivityContext/InactivityDialog.tsx
petertgiles: @BrookeCameron , what do you think of this suggestion?

apps/web/src/components/ActivityContext/ActivityContainer.tsx
petertgiles: This would be pretty easy but I'd need another string translated from Fay.  I might need to up the tick rate to make sure it looks stable as well.  @BrookeCameron , your thoughts?

apps/web/src/components/ActivityContext/InactivityDialog.tsx
petertgiles: Heh, Brooke flagged this as well.  This is one of those places where the component code is quite different from [the component design](https://www.figma.com/design/guHeIIh8dqFVCks310Wv0G/Component-library?node-id=3284-10750&m=dev) in Figma.  :disappointed:   In Figma the heading text is much smaller.  I tried to sync the icon to the actual font size but I made a mistake in the lg size: it should have been 10.5 instead of 12.  Brooke also requested that I make the icon 90% of the text size.
Anyway, that's here: [90% icon](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16539/commits/cafede90f33201b9ecf5a700f1494e00b51911e6)

I agree that this makes sense to incorporate into the base component.  I'd want to see it in Figma first, though.

apps/web/src/components/ActivityContext/InactivityDialog.tsx
petertgiles: Oh, [that component looks really different](https://www.chromatic.com/component?appId=61e099a1bb1465003a73d3ce&csfId=components-alertdialog&buildNumber=14285&k=69e7945b09e5dcac2020cc56-dark-interactive-true&h=4&b=-1).  Do we have a way to change the just the role on our dialog component?

apps/web/src/components/ActivityContext/InactivityDialog.tsx
petertgiles: Oh good catch, thanks.
[switch to Ps](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16539/commits/9f410ad493e7d718f5a28aea438e879c094eacf0)

apps/web/src/components/ActivityContext/InactivityDialog.tsx
esizer: I don't think so 😢 

I'm not sure why it looks so different either 😕  but this does seem like something that should be an `alertdialog`

apps/web/src/components/ActivityContext/InactivityDialog.tsx
petertgiles: It is extremely close.  The top one is the IconLabel component used on the pool card.  The bottom one is the custom one I built from the Figma design.  @BrookeCameron is it worth matching the pool card or do you have a strong preference for the custom one?

image

apps/web/src/components/ActivityContext/InactivityDialog.tsx
petertgiles: Hmm, I'm not sure there's a fix for this that I could reasonably include in the scope of this PR.

It seems like this is another one of those "the design and the code have deviated" situations.  In Figma, the Dialog component has two different variants while the code has two totally different components.  I could make an issue to consolidate the two into variants in a single Dialog component that can switch between "full" and "simple" styling and can switch the base primitive between "dialog" and "alertdialog".
image

apps/web/src/components/ActivityContext/InactivityDialog.tsx
BrookeCameron: ah good point @esizer , you can remove the Aria label @petertgiles

apps/web/src/components/ActivityContext/ActivityContainer.tsx
BrookeCameron: I think that makes sense, if it's rounding up and you really only have 10 seconds left that seems pretty annoying.

apps/web/src/components/ActivityContext/InactivityDialog.tsx
BrookeCameron: I think go ahead and keep things consistent @petertgiles , good catch @esizer

BrookeCameron (approved):

apps/web/src/components/ActivityContext/InactivityDialog.tsx
petertgiles: Sounds good.
[remove aria label for timer icon](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16539/commits/5a9696473226aeaa8b988a73f7e0d74cafe1bf15)

apps/web/src/components/ActivityContext/InactivityDialog.tsx
petertgiles: No problem.
[use IconLabel](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16539/commits/2fea1fdc8b703b91a06d2ec324fa6a70e709c004)

apps/web/src/components/ActivityContext/ActivityContainer.tsx
petertgiles: Here it is.
[minutes and seconds](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16539/commits/ef05cdc7a08ca5b6443637300ef2af65dd970252)

apps/web/src/components/ActivityContext/InactivityDialog.tsx
esizer: okay, unfortunatly that ended up with the icon being larger than the text still 😢 

I did the following which is easier to read but is only 86% of the text and not the requestd 90% 😖 

```tsx
<div className="flex flex-row gap-3 self-start">
  <ExclamationTriangleIcon
    className="mt-2 size-5 shrink-0 text-warning lg:size-6" />
```

petertgiles: I'm in design hell now.  :grimacing: Could you post a picture of that for Brooke?

apps/web/src/components/ActivityContext/InactivityDialog.tsx
```diff
@@ -0,0 +1,145 @@
...
+                defaultMessage:
+                  "Your session is about to end due to inactivity",
+                id: "eMMDjW",
+                description: "Title for the inactivity dialog",
+              })}
+            </span>
+          </div>
+        </Dialog.Header>
+        <Dialog.Body>
+          <div className="grid gap-(--text-sm) xs:grid-cols-[1fr_1fr]">
```
esizer: Oh, not a necessary change but I just noticed this and it could be simplifed since `1fr` is the default value for grid columns.


```suggestion
          <div className="grid gap-(--text-sm) xs:grid-cols-2">
```

Image

apps/web/src/components/ActivityContext/InactivityDialog.tsx
esizer: Yup, ohviously we should ignore the known bug of the close icon overlap :sweat_smile: 

swappy-20260423_090625
swappy-20260423_090633

petertgiles: @BrookeCameron how to do you feel about 86% size?  :point_up:

BrookeCameron: 86 is great, thank you!

petertgiles: Well, the styling code is a lot simpler.  :laughing: 
[86% icon](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16539/commits/8c7edf4c7b1621104e59b2de281864cad89b68c4)

apps/web/src/components/ActivityContext/InactivityDialog.tsx
esizer: Okay, maybe that will go on our list of components to address 😆

esizer (approved):
Very neat 🤓 Except when I forget to change the timeout numbers back and it keeps popping up every 10 seconds and agressivley logging me out 😆

PR #16546, opened 2026-04-21
[Fix] Remove gov employee check for profile completness

### LINKED ISSUE #16537: 🐛 Government employee status "not provided" in profiles with no work experience.

If only learning and community experiences are provided in a profile, the system does not capture GC employee status. The status should read - "No, I am not a Government of Canada employee." What we see - "not provided".
As the result, users cannot submit applications since their profile information is incomplete. Users do need to have work experiences to be able to apply to certain opportunities.  

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

Image

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] existing users with all experiences removed have "not gov" status
- [ ] new users with no experiences yet have "not gov" status
- [ ] existing behaviour around adding and removing gov experiences and gov status preserved
- [ ] validation for user profile accurate
- [ ] validation for application accurate
- [ ] existing users in the system fixed if needed

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #16537 

## 👋 Introduction

Removes the check for government employee status from profile completeness.

## 🕵️ Details

This is automatically set when work experiences are modified. However, since we don't require any work experience to appy, this check doesn't make much sense and can remain `null`.

## 🧪 Testing

1. Login as a brand new user
2. Complete your profile without creating any work experiences
3. Confirm you can submit an application while `computed_is_gov_employee` is `null`

### DIFF
--- a/api/app/Models/User.php
+++ b/api/app/Models/User.php
@@ -620,7 +620,6 @@ class User extends Model implements Authenticatable, HasLocalePreference, Laratr
                 is_null($this->attributes['looking_for_french']) &&
                 is_null($this->attributes['looking_for_bilingual'])
             ) or
-            is_null($this->attributes['computed_is_gov_employee']) or
             is_null($this->attributes['has_priority_entitlement']) or
             ($this->attributes['has_priority_entitlement'] && is_null($this->attributes['priority_number'])) or
             is_null($this->attributes['flexible_work_locations']) or


JamesHuf (approved):
Works for me!

Issue #16368, opened 2026-04-02
✨ Linking experience to development program interests


vd1992:
### Details

Enable a user to link an education experience to a development program interest during the community interest flow

The backend work for this should have been completed in https://github.com/GCTC-NTGC/gc-digital-talent/issues/16471

#### Linking experience to completed professionalization options

The user is presented with the option to link existing education experience to the professionalization option. They are also provided with a button that should take them to the education experience form in a new tab to avoid interrupting their flow.

> [!IMPORTANT]
> When the "Link existing experience" button is clicked, the UI should reload the user's experience to ensure anything new is loaded into the page. In a similar fashion, if the user uses the "Add new experience" button in the dialog, the list should be refreshed when the dropdown is focused.  

Once an experience is linked, the user can remove it (standard confirmation dialog) or swap it out for a different education experience.

<img width="5375" height="6280" alt="Image" src="https://github.com/user-attachments/assets/87022444-a918-4cc7-9239-6d29a4bbd112" />

### Design file & designer

https://www.figma.com/design/H1loB8WxGP6vyJTrVqcafd/Communities-of-Interest--Employees-?node-id=2032-14331&p=f&t=B6G78jmmnRYSG1Kx-0

### 📸 Screenshots

<img width="687" height="260" alt="Image" src="https://github.com/user-attachments/assets/e6b2ec46-4ae0-471a-8263-8427bfdcb5ec" />

<img width="785" height="380" alt="Image" src="https://github.com/user-attachments/assets/0edf754b-f244-4652-9aa7-22cd57be5dca" />

<img width="854" height="353" alt="Image" src="https://github.com/user-attachments/assets/7b1ad218-eaaf-4e59-9873-a87dcd96cd0c" />

<img width="399" height="479" alt="Image" src="https://github.com/user-attachments/assets/21e5bfb8-ef9d-4e0e-9e1d-bc67939fb89d" />

### ✅ Acceptance Criteria
- [x] When user selects completed for developement program , Link experience button appears 
- [x] upon clicking Link button all education experience should be shown 
- [x] when selecting a education experience it should be added into the form 
- [x] Once added education experience can be swapped or deleted 
- [x] Linking or de-linking an education experience doesn't trigger any mutation. It should be kept only in UI 
- [x] Submission should save community interest and development programs with experience link
- [x] When dialogue is opened, the list of experiences refreshes 
- [x] Double check validation for linked experience: Only education experience, Status = Completed only.

[...]

PR #16566, opened 2026-04-22
[Feature] professionalization component

### LINKED ISSUE #15941: ✨ Create the Professionalization Component (Talent Events and Professionalization)
Image

## 📚 Details

[Visit the Figma file](https://www.figma.com/design/3VPSAv5689GK51Us6BxbJm/Admin---Communities?node-id=10507-3173&t=lkB6gAzQ9yIZhJPP-4) • Contact @JoshBeveridge with questions

The goal of this component is to create a reusable element for displaying professionalization information. We're also hoping to test new placement for the action menu (the left hand side) to address visibility/findability issues noted by users using various zoom technologies.

Component data:

* Title (Required, EN/FR)
* Description (Required, EN/FR)
* Classification restrictions (Optional)

Other options:

* Striping (toggles whether the background is #F8F9FC or white)
* Divider (`1px` high, `gray.lighter` in light mode, `gray.darkest` in dark mode)
* Action menu (toggles whether the pencil icon button containing the edit/delete buttons is shown)

### Action Menu

Image

## ✔️ Requirements

- [ ] The component accepts and displays a title and description
- [ ] The component can optionally display classification restrictions
- [ ] The component can be candy-striped (striped background colour is #F8F9FC)
- [ ] The component has an action menu that allows the professionalization to be edited or removed (see #15934 for more details)
- [ ] The component contains an optional divider 
- [ ] Colours match Figma in both light and dark mode
- [ ] Spacing and layout of all elements matches Figma for both narrow and wide displays
- [ ] Design review has passed

yonikid15:
🤖 Resolves #15941 

## 👋 Introduction

- Adds the professionalisation component (a.k.a development program)
- Adds story

## 🧪 Testing

1. Run storybook
2. Ensure the component meets all A/C
3. Ensure component matches design

## 📸 Screenshot

image


apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
```diff
@@ -0,0 +1,87 @@
+import { useState } from "react";
+import type { ReactNode } from "react";
+import XMarkIcon from "@heroicons/react/24/outline/XMarkIcon";
+import PencilSquareIcon from "@heroicons/react/24/outline/PencilSquareIcon";
```
vd1992: Per issue screenshot, I think this should probably be Solid rather than Outline

vd1992 (approved):
I think the icon needs to be changed, other than that looks good.
Issue mentions design review so gonna tag @JoshBeveridge

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.stories.tsx
```diff
@@ -0,0 +1,51 @@
+import type { StoryFn, Meta } from "@storybook/react-vite";
+
+import { allModes } from "@gc-digital-talent/storybook-helpers";
+import { fakeClassifications } from "@gc-digital-talent/fake-data";
+
+import DevelopmentProgramCard from "./DevelopmentProgramCard";
+import fakeDevelopmentPrograms from "../../../../../packages/fake-data/src/fakeDevelopmentPrograms";
```
esizer: Should be `@gc-digital-talent/fake-data`

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
```diff
@@ -0,0 +1,87 @@
...
+            className="flex flex-col items-start gap-3 p-3"
+            portalProps={{ keepMounted: true }}
+          >
+            {edit && <DropdownMenu.Item>{edit}</DropdownMenu.Item>}
+            {remove && <DropdownMenu.Item>{remove}</DropdownMenu.Item>}
+          </DropdownMenu.Popup>
+        </DropdownMenu.Root>
+        <div className="flex flex-col items-start gap-6">
+          <span>
+            <p className="font-bold">{title}</p>
```
esizer: Since this is a title, shouldn't it be a `Heading`? 🤔

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
```diff
@@ -0,0 +1,87 @@
...
+  edit,
+  remove,
+  classificationRestrictions,
+}: DevelopmentProgramCardProps) => {
+  const intl = useIntl();
+  const [open, setOpen] = useState(false);
+
+  return (
+    <div className="border-b border-gray-200 p-6 last:border-b-0 odd:bg-gray-100/30 dark:odd:bg-gray-700/50 dark:even:bg-gray-700/30">
+      <div className="flex items-start gap-6">
```
esizer: This gap seems too large, figma is showing `gap-3`

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
```diff
@@ -0,0 +1,87 @@
...
+  iconLabel,
+  edit,
+  remove,
+  classificationRestrictions,
+}: DevelopmentProgramCardProps) => {
+  const intl = useIntl();
+  const [open, setOpen] = useState(false);
+
+  return (
+    <div className="border-b border-gray-200 p-6 last:border-b-0 odd:bg-gray-100/30 dark:odd:bg-gray-700/50 dark:even:bg-gray-700/30">
```
esizer: Just a small suggestion. Instead of removing the border for the last item, you might beable to use `not-last:border-b` instead.

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
```diff
@@ -0,0 +1,87 @@
...
+      <div className="flex items-start gap-6">
+        <DropdownMenu.Root open={open} onOpenChange={setOpen}>
```
esizer, line 46: I dont think you need to pass in any props here unless you plan on controlling it.

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
```diff
@@ -0,0 +1,87 @@
...
+              <IconButton
+                icon={open ? XMarkIcon : PencilSquareIcon}
+                color="primary"
+                label={iconLabel}
+                className="m-0 p-0"
+              />
+            }
+          />
+          <DropdownMenu.Popup
+            className="flex flex-col items-start gap-3 p-3"
```
esizer: I feel uneasy about making overrides to our dropdown. I feel like if we want these changes we should consider updating the underlying component 🤔

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
esizer: Yeah, also probably want `20` and not `24` I think.

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/design/guHeIIh8dqFVCks310Wv0G/Component-library?node-id=11-4252",
+    },
+  },
+} as Meta<typeof DevelopmentProgramCard>;
+
+const Template: StoryFn<typeof DevelopmentProgramCard> = () => {
+  return (
+    <div className="bg-white">
```
esizer: Won't this lead to unexpected results since you use transparency? Our background colour is `bg-gray-100` so setting this to white might be misleading.

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
esizer, line 1: This component is called `DevelopmentProgramCard` but the ticket describes it as "Professionalization"

While I do prefer development program, I feel like it might be too specific? Do we know that this component will only ever be used with development programs?

If we replace the `classificaitonRestrictions` with a slot that accepts any `ReactNode` this component can easily be reused in other places.

esizer (changes requested):

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
```diff
@@ -0,0 +1,87 @@
...
+    <div className="border-b border-gray-200 p-6 last:border-b-0 odd:bg-gray-100/30 dark:odd:bg-gray-700/50 dark:even:bg-gray-700/30">
+      <div className="flex items-start gap-6">
+        <DropdownMenu.Root open={open} onOpenChange={setOpen}>
+          <DropdownMenu.Trigger
+            render={
+              <IconButton
+                icon={open ? XMarkIcon : PencilSquareIcon}
+                color="primary"
+                label={iconLabel}
+                className="m-0 p-0"
```
esizer: Removing margin and padding is also giving this a really odd look like it is suffocating 😆

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
esizer: We might also want to consider rendering these in `li` components too fi they are going to be used as a list and not in isolation.

REF: https://inclusive-components.design/cards/

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
yonikid15: I need the open state to change the icons

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.stories.tsx
yonikid15: We need the white background to see the alternating colours. Also, in all the designs it's used on a white background.

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
yonikid15: I don't know if we plan to use this component else where, but today it's just being used with professionalization/development programs. In the future, if we plan to use it elsewhere, we can refactor.

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.stories.tsx
yonikid15: Fixed in [Fix fakeDevelopmentPrograms import](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16566/commits/7acb0e1ea660477ad5198814783b57c2b5c7da0d)

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
yonikid15: Changed in [Change icon import](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16566/commits/cabcde6f148d4d7c4ee8a82a2a262d9f6383f179)

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
yonikid15: Changed in [Change p tag to heading for title](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16566/commits/a171fc4d179e70ce71d85ccd7aab2433a5d8e971)

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
yonikid15: Fixed in [Shrink gap](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16566/commits/07b42e26c63e4f43447fad656bae17cf079f8ed1)

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
yonikid15: I tried this before, but didn't seem to work for some reason :shrug:

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
yonikid15: Removed in [Remove overrides on dropdown menu popup](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16566/commits/9cce2cc00ed4df936ca901c0d05935a099dd9811)

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
yonikid15: Removed in [Remove styling on icon](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16566/commits/fd31d0d5ea8f0eb9a17879e1696e14b1f86ebb60)

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
yonikid15: Changed to list in [Move cards into list](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16566/commits/df0ae54c8c23968465b307fea74fcc2ed30a4788)

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.stories.tsx
esizer: Yeah, im saying that is misleading. Unless it actually is used on a white background. which I'm not sure we can gaurantee 😅 Because now it looks like this in dakr mode:

Image

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
esizer: Right, oops 😛

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
esizer: so, if we don't plan on using this elswhere... my next quesiton is, why are we not using existing components to solve this identical issue? :sweat_smile: 

 - [Repeater](https://www.chromatic.com/component?appId=61e099a1bb1465003a73d3ce&csfId=components-repeater&buildNumber=14318&k=69ea316b800db26f80ab53b8-1200px-interactive-true&h=8&b=-1)
 - [CardRepeater](https://www.chromatic.com/component?appId=61e099a1bb1465003a73d3ce&csfId=components-cardrepeater&buildNumber=14318&k=69ea316b800db26f80ab52ef-dark-interactive-true&h=8&b=-1)

vd1992:
I guess Eric handled the design review

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
yonikid15: That's a good question for the designers. @JoshBeveridge thoughts on this?

One thing mentioned in the issue is:

> "The goal of this component is to create a reusable element for displaying professionalization information. We're also hoping to test new placement for the action menu (the left hand side) to address visibility/findability issues noted by users using various zoom technologies."

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.stories.tsx
yonikid15: Oops forgot to add dark mode background. Fixed in [Fix background color in storybook](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16566/commits/b110c9c06a4f66d35e1106135d75aa64d9b0b0e9).

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
JoshBeveridge: As Yoni pointed out - we do hope to extend the left-hand action menu pattern to other card patterns. This one is a test for us to see how it compares. If it works better, we'll move to migrate other cards that make sense. We need some flexibility to be able to A/B test components, and building them side-by-side is, to my knowledge, the only method we have available right now.

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.stories.tsx
esizer: We shouldn't have a background at all :laughing: Our storybook sets our site background colour so that it shows a realistic environment.

If we need this component to show up on a specific background colour, that should be part of the component, not something we need to remeber to set. As an example, our accordion requires a whirte background but it is baked into the `Root` of the accordion:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/5a3e4bb12bc0f44f278574e588eb2c6bffe399aa/packages/ui/src/components/Accordion/Accordion.tsx#L35

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
esizer: Yeah that makes sense however, I was asking if this should be a generalized component like the Repeater components. Becuase right now, it is specific to displaying development programs and I am already seeing [tickets pop up that look very similar to this](https://github.com/GCTC-NTGC/gc-digital-talent/issues/16368) which makes me think this is a generic component we plan on using elsewhere.

I just dont want us to end up with 2-3 different implementations of the same component pattern again if we already have plans to start making this a general component and not specific to this page. Such as:
 
  - Renaming it from `DevelopmentProgramCard` to something more generic
  - Putting it in the `ui` or `form` package
  - Changing the props to be more general or even making it a compound component so we can better control the contents for flexibilty. 


In the past, we just end up making a duplicate (sort of similar) component which is something we have been trying to avoid since they go out of sync over time.

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.stories.tsx
yonikid15: Oh I see, I added subcomponents for the root and item. Hopefully that works.

Changed in [Add root and item sub components](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16566/commits/a2ec4bf0434b9fdcd1d32593def87632eb800504).

apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
JoshBeveridge: Couple thoughts here:

Just so we're using the same language here, outside of the original card with an image found in the "component" library, design hasn't consistently treated individual card layouts as reusable, we've focused on content. Card styles are usually applied to a group of content elements based on need. I hear that dev is treating these whole card elements as reusable "components" though. If we look at the standard design system model - components/patterns/templates - we could establish three patterns here - a wrapper pattern and two content patterns.

**The action menu wrapper pattern**
This handles whether card styles should be applied (this can appear in a card list or other UI without the card styles), whether the item needs a divider, whether there should be striping between rows, and the placement and content of the action menu.

image

**Professionalization summary pattern**
This is a specific organization of text that conveys a professionalization. It may or may not be used by other models, but we shouldn't assume it will be. We CAN assume though that when we show a professionalization, this layout will be used everywhere on the site.

image

**Experience summary pattern**
This is a specific organization of data associated to an experience, so in a similar fashion, we can assume this will be reused when showing experience elsewhere, but not that it applies to other models.

image

Which, when combined, lead to:
image

Not sure if this makes sense, but if we have reusable content patterns for models that can then be put into different contexts (such as various cards or lists, etc.) I think we'll find things a bit easier to maintain.

esizer: Yeah, this makes a lot of sense :+1: 

The context is super important becuase in the story it is not displayed in a card. Which lead to some oddities because it required a white background and ended up [looking very odd without one](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16566#discussion_r3131651193).

I'm thinking of how we can build the building blocks so to say, so they they can be plugged together to construct the content for their specific needs. Thats where we can probably leverage compound components. With those layouts, I think it could be doable so that, while the content is different and flexible for what it displays, the layout, spacing and font-size are consistent with similar components.

## Example

This assumes it always appears in a card. If not, we might need to address somethings. Honestly, if we handle the `SummaryList` and other components properly, they dont need to be ain card, We can just use the `Card.SummaryList` as a wrapper over a more genric `Summary.List` component that intrcuts itself on how to be shown in any context and then the `Card.SummaryList` wrpaper would give it additional classes for how it should appear in a card.

Of course, each component accepts `className` to override any styles as needed for specific contexts.

So, the idea is that each component does minimal styling to get the layout, spacing and other base styles. But, it reamins agnostic about what you put into it for flexibility.

> [!NOTE]
> Naming can be improved, this was all just off the top of my head :sweat_smile:  

```tsx
{/** Applies the background, shadow, border radius and padding */}
<Card.Root>

  {/** Parent ul element that supplies item spacing, striping, separators? */}
  <Card.SummaryList
  
       {/** This could also be a `SummaryList.Add` that is a button with placeholder mode by defualt */}
       addBtn={{
         label: "Add Development Program",
         onAdd: (vals) => append(vals)
       }}
   >
       
    {/** Provides the layout of the item */}
    <SummaryList.SummaryItem>
    
        {/* Dropdown Menu that can contiane anything */ }
	    <Summary.Menu icon={EditSquareIcon} openIcon={XIcon}>
	      <Menu.Item><EditDialog /></Menu.Item>
		</Summary.Menu>
		
		{/* Contains content so it can be placed next to actions */}
		<Summary.Content>
		  
		  {/* Provides font-size, colour, etc? */}
		  <Summary.Heading as="h3">{title}</Summary.Heading>/
		  
		  {/** Maybe not needed? */}
		  <Summary.Description>{desc}</Summary.Description>
		  
		  {/** Can put any content here for flexibility */}
		  <p>{desc}</p>
		  
		  {/** Provides the horizontal list with bullet separators */}
		  <Summary.Meta>
		  
		      {/** Anything can go here for flexibility */}
			  <span>{type}</span>  
		  
		  </Summary.Meta>
		  
		  
		</Summary.Content>
    
    </SummaryList.Item>
  
  
  </Card.SummaryList>
  
</Card.Root>
```

esizer (approved):

Issue #16300, opened 2026-03-30
♻️ Implement `@advancedOrderBy` for `poolCandidatesPaginated` queries


esizer:
### ♻️ Refactor

We now have a more advanced order by directive (`@advancedOrderBy`). We should start using it where it makes sense. The first target would be the large `poolCandidatesPaginated*` queries.

### 🕵️ Details

We have the base `poolCandidatesPaginated` query deprecated for some time. Maybe we also just finally remove it and replace it with the current verison and drop the `AdminView` suffix since it is only ever using in `admin` and it does seem odd to name a query based on where it would be used rather than what it does.

### 🙋‍♀️ Proposed Solution

- Update the query to use the new directive
- Update client usage of the query
- Remove any scopes that are no longer necessary

Should be something like:

```graphql
  poolCandidatesPaginated(
    where: PoolCandidateSearchInput
  ): [PoolCandidateAdminViewWithSkillCount]!
    @guard
    @advancedOrderBy
    @paginate(
      defaultCount: 10
      maxCount: 1000
      scopes: ["whereNotDraft", "whereSkillCount", "whereAuthorizedToView"]
      model: "App\\Models\\PoolCandidate"
    )

```

### ✅ Acceptance Criteria

- [ ] Deprecated query replaced with new one
- [ ] Order scopes/args replaced by `@advancedOrderBy`
- [ ] Frontend updated to use updated query

vd1992:
I think this might be trickier to implement than first expected, I was looking through things to get a feel for the directive.
Pool candidates is complex, with all the customization. As well as different behaviour depending on where you are querying from

### Complicating factor
If queried from the scope of a single process, you get a triple layer sort. A default sort by column flag on the candidate as well as a more complex sort by pivot table (bookmarking). Then a third layer sort that is variable (name, department, etc).
Right off the bat it doesn't appear the new directive supports multi-layer sorting

esizer:
> If queried from the scope of a single process, you get a triple layer sort. A default sort by column flag on the candidate as well as a more complex sort by pivot table (bookmarking). Then a third layer sort that is variable (name, department, etc).
> Right off the bat it doesn't appear the new directive supports multi-layer sorting

It is supposed to, it accepts an array that should represent the priority. For the other sorts that are more complicated like bookmakr, it also supports scope based sorting. You just need to make sure we have something like `orderByBookmark` scope on `PoolCandidateBuilder`. The variables would look something like this:

```tsx
[...]

PR #16587, opened 2026-04-23
[Feat] Full next/prev candidates links

### LINKED ISSUE #10286: ✨ Get Next & Previous candidates buttons to work with full list of candidates
## ✨ Feature

In #10271, we are adding functionality to use the "next" and "previous" candidates buttons (on the candidate application page) with all candidate tables. However, it will be limited to only the candidates on the current page (from paginated query). This issue is to make it work with the full list of candidates instead.

## 🕵️ Details

The list of candidates being paged through _should_ depend on the filters applied to the original table but _should not_ be restricted to the current page or page size of the table.

## 🎨 Design File

### 🧑‍🎨 Designer

## 📸 Screenshot

We believe this will be easier after we've implemented cursor-based pagination. Then, instead of storing a list of candidate ids in the page state, we'd store the filters associated with the query (or the query itself?) along with the current cursor.

(Would cursor based pagination allow for retrieving the previous result? We might want to store the previous cursor as well.)

## 🌎 Localization

## ✅ Acceptance Criteria

- [ ] The list of candidates being paged through by next/previous buttons reflects the list of candidates who would appear in the original table (ie the same filters apply)
- [ ] The list of candidates being paged through by next/previous are not restricted to the candidates which appeared on the "current page" of the table.

## 🛑 Blockers

```[tasklist]
### Blocked By
- [ ] #10271
- [ ] #10827
```

esizer:
🤖 Resolves #10286 

## 👋 Introduction

Updates the candidate prev/next navigation links to support navigating the entire table.

## 🕵️ Details

Instead of passing all IDs in the current data set, we pass all the filters from the table along with the row index selected as the "current page".

We use this to then fire a query with two fields (previous, next) where they are single paged responses where the current page was the item index in the table paginated response +/- 1 to get the next/previous candidate.

## 🧪 Testing

> [!TIP]
> I used two side-by-side windows, one with the table another with the current candidate. This was helpful to ensure the navigation was properly following the table order. 

1. Build `pnpm dev:fresh`
2. Login as `admin@test.com`
4. Apply some filters/sorting
5. Click on a candidate in the middle
6. Confirm prev/next links work
7. Confirm it can navigate across table page boundaries


apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
```diff
@@ -281,7 +286,10 @@ const ResponsiveTable = <TData extends object, TFilters = object>({
         Object.fromEntries(managedKeys.map((key) => [key, newParams.get(key)]));
 
       if (!isEqual(desiredTableParams, lastWrittenTableParamsRef.current)) {
-        setSearchParams(newParams, { replace: true });
+        setSearchParams(newParams, {
```
petertgiles, line 289: Oh no, this state management in a useEffect was complicated enough.  Now we're trying to juggle state between the search params and a ref?  Now we've got two refs? I'm worried we're speeding towards "works but is unmaintainable".  😞

apps/web/src/hooks/usePoolCandidateNavigation.ts
```diff
@@ -0,0 +1,153 @@
...
+  Maybe,
+  PoolCandidatePoolNameOrderByInput,
+  PoolCandidateSearchInput,
+  PoolCandidatesBaseSort,
+  QueryPoolCandidatesPaginatedAdminViewOrderByRelationOrderByClause,
+  SortOrder,
+} from "@gc-digital-talent/graphql";
+
+// The filter/sort params shared between the table query and navigation queries
+export interface CandidateNavigationQueryParams {
```
petertgiles, line 17: Ugh, a hidden cost of having all those custom sorting scopes.

petertgiles (approved):
I've got mixed feelings on this one.  😅  On one hand, it works very well.  👏   On the other hand, the amount of complexity that is building in the ResponsiveTable and PoolCandidateTable components is starting to really worry me.  I'm at the point now where I'd cringe if I had to work on either one.

I think I'm OK to merge this given that we:
- are planning on splitting this table soon
- still have an open issue for per-edge cursors
- have an open issue to move some of the specialty scopes to the advanced ordering directive

apps/web/src/components/Table/ResponsiveTable/ResponsiveTable.tsx
esizer: Yeah, I would really love to solve this. It is way too wild :grimacing:

apps/web/src/hooks/usePoolCandidateNavigation.ts
esizer: Yeah, hopefully we get some buy in to [move the new directive forward](https://github.com/GCTC-NTGC/gc-digital-talent/issues/16300). I was hoping to wait for that one to make this eaiser but it has been sitting around stale with no traction for a month so I just gave up waiting and moved forward :stuck_out_tongue:

esizer:
> I've got mixed feelings on this one. 😅 On one hand, it works very well. 👏 On the other hand, the amount of complexity that is building in the ResponsiveTable and PoolCandidateTable components is starting to really worry me. I'm at the point now where I'd cringe if I had to work on either one.

Oh trsut me, I feel the exact same way :laughing: 

I have been thinking about how we can improve this. I do know designers are wokring on a new version of the table, we just havent gotten around to it. That might be a good way of cleaning this up.

I am starting to think it was a mistake to combine server and client tables :cry: I thought it would make things easier but its clearly just overcomplicating things :-1: 

We could keep all the internals the same and same API but the state managemight might make sense to split either by doing `Table.Server` and `Table.Client` or some hooks or something :woman_shrugging: 

Also, I would love to see if we can avoid the `useEffect` eniterly :sweat_smile:

petertgiles:
> Oh trsut me, I feel the exact same way 😆

Cool, cool.  To be clear, I still think we should merge this.  If you're thinking about this too, I'm sure we'll find ways to simplify.

PR #16606, opened 2026-04-24
[Feat] Restore soft deleted experience skills

### LINKED ISSUE #7833: ✨ Restore details when re-adding a soft-deleted Experience-Skill

esizer:
🤖 Resolves #7833 

## 👋 Introduction

Updates the experience form to restore details from a soft-deleted experience skill record.

## 🕵️ Details

This works by querying for a trashed experience skill after selecting a skill from the dialog. If it finds one, it is restored and the details are populated in the form.

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Login as `admin@test.com`
3. Create an experience, attaching a skill with some details
4. Delete the skill and save the experience
5. Go back and select the same skill
6. Confirm that the details from step 3 are populated

### DIFF
--- a/api/app/GraphQL/Queries/TrashedExperienceSkill.php
+++ b/api/app/GraphQL/Queries/TrashedExperienceSkill.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace App\GraphQL\Queries;
 
 use App\Models\ExperienceSkill;
+use App\Models\User;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Gate;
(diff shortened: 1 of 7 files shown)


yonikid15 (approved):
Everything works as expected 🚀 
Tests also look great!

PR #16617, opened 2026-04-27
[Refactor] Full `Announcement` model

### LINKED ISSUE #16373: ♻️ Make announcements full Laravel models
### ♻️ Refactor

When the announcements were originally implemented it was done in a hurry without any provided designs.  One of choices was to implement it without a backing Laravel model.  However, this results in:
- no automatic casting
- no automatic Lighthouse functionality
- no activity logging
- no automatic timestamps

Since the sitewide announcement continues to be used on a regular basis and hasn't had any significant design changes lets convert it to a full model.

### 🕵️ Details

_No response_

### 🙋‍♀️ Proposed Solution

It is currently living in a JSON blob in a settings table.  Perhaps it should be moved to an "announcements" table and identify the sitewide announcement with a key column.  The settings table was never used for anything else so it could get dropped.

### ✅ Acceptance Criteria

- [ ] The sitewide announcement is a full Laravel model
- [ ] No functional or visible differences to the sitewide announcement

esizer:
🤖 Resolves #16373 

## 👋 Introduction

Updates the announcements to use a full laravel model with table, factory and validation.

## 🕵️ Details

This adds a new table `announcements` migrating the current JSON to table columns. It also adds the corresponding factory and validation as well as updates the graphql schema.

## 🧪 Testing

1. Run migration `artisan migrate`
2. Confirm it runs, removes the `settings` table and creates new `announcments` table with the proper content
3. Confirm you can still edit and view the announcement as before (no regressions)


api/database/factories/AnnouncementFactory.php
```diff
@@ -0,0 +1,42 @@
...
+    protected $model = Announcement::class;
+
+    /**
+     * Define the model's default state.
+     *
+     * @return array
+     */
+    public function definition()
+    {
+        $publishDate = $this->faker->dateTimeBetween('-1 month', 'now');
```
petertgiles: Can we avoid floating dates?  They can cause false positive diffs.  Same for expiry_date.

petertgiles (approved):
Nice!  Thanks for picking this one up.  One minor recommendation but nothing blocking.

api/database/factories/AnnouncementFactory.php
petertgiles: I guess as long as the seed is steady, the `dateTimeBetween` is fine.  The floating `now` is where we've seen trouble before.

esizer: Oh yeah, for sure! Just so I'm clear on it, are you suggesting something like we did in the `UserFactory`?

https://github.com/GCTC-NTGC/gc-digital-talent/blob/3b2eba1fe892a99f13c2a5bd60403eaf700fec7b/api/database/factories/UserFactory.php#L224

petertgiles: Yeah, that's great. :+1:

esizer: Cool, updated in [use static dates in factory](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16617/commits/5b194a430a97a4e74172a4dc33b8265ecc473280)

Issue #16392, opened 2026-04-08
🐛 UAT | Department Admin is unable to close the published pool created by the same user


mauli0110:
### 🐛 Bug and Expected Behaviour

- A department admin creates a new pool and published it, when department admin trying to close the pool immediately, an error occurred. 

Expected outcome: The pool should be closed by department admin

### 🕵️ Details

Env: UAT

### 📋 Steps to Reproduce

1. Access to GC digital portal using Department admin role
2. Create a new pool and publish it
3. Click on Change closing date
4. Click on Close immediately and enter the reason
5. Observe the error.

### 📸 Screenshot

https://github.com/user-attachments/assets/64067852-af1e-4a8c-83ea-174807590191

### ✅ Acceptance Criteria

- [ ] Department Admin should be able to close the pool which is created and published by the same department admin

### 🔄 Regression

No

esizer:
I actually noticed that while working on the policies and thought it was intentional so added a coment 😅 

Sounds like it wasnt? 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/851d37d1d250b7f12d197b3d7cfa0441ae081fdf/api/app/Policies/PoolPolicy.php#L204-L208

vd1992:
Intentional as far as I can tell
Mutation -> `closePool`
Policy -> `closePool`
Relevant permission -> `update-team-publishedPool`

Only role that has that is Community admin

Now whether that should be added to department admin, is the question

petertgiles:
@gobyrne Right now there are no department roles that can update a published pool, like closing it early.  Should department admin have this permission?

<img width="993" height="584" alt="Image" src="https://github.com/user-attachments/assets/aea81571-d397-4dc7-8752-2d2a692083d6" />

gobyrne:
Yes, department admin should be able to update published pools. Thanks, I've updated the spreadsheet.

<img width="1315" height="696" alt="Image" src="https://github.com/user-attachments/assets/9c53deae-31cd-4bbe-9fcf-bb233a5e54ce" />

PR #16655, opened 2026-04-29
[Refactor] Centralize team checks

### LINKED ISSUE #16131: ♻️ Centralize policy team checks
### ♻️ Refactor

We have a very common pattern with team based permision checks in our policies. It might make it more maintainable to centralize it.

### 🕵️ Details

The pattern that is very common is mostly found in `PoolCandidate` and `Pool`. But, they are our most complicated ones and there is a lot of repetition right now.

```php
$teamPermission = ! is_null($model->team) && $user->isAbleTo('view-team-model', $mode->team);
$communityPermission = ! is_null($model?->community?->team) && $user->isAbleTo('view-team-model', $model->community->team);
$departmentPermission = ! is_null($model?->department?->team) && $user->isAbleTo('view-team-model', $model->department->team);

if ($teamPermission || $communityPermission || $departmentPermission) {
  return true;
}
```

### 🙋‍♀️ Proposed Solution

Create  trait for policies that plan on using a similar pattern.

If we want to get really wild, we could do a request based cache with a static variable so the team ids stay hot during the request 🤷‍♀️ 

```php
trait ChecksTeamPermissions
{
   protected static array $cache = [];

   protected function checkTeamPermissions(User $user, array $models, string $permission)
   {
     foreach($models as $model) {
        if(!$model) continue;
        
        $teamId = $model->team_id ?? $model->id;
        $cacheKey = "{$user->id}_{$teamId}_{$permission}";
        
        if (!isset(self::$cache[$cacheKey])) {
          self::$cache[$cacheKey] = $user->isAbleTo($permission, $model);
        }
        
        if (self::$cache[$cacheKey]) {
           return true;
        }
        
     }
     return false;
   }
}

class PoolCandidatePolicy
{
   use ChecksTeamPermissions;

   public function viewStatus(User $user, PoolCandidate $poolCandidate)
   {
     // ...

      return $this->checkTeamPermissions($user, [
            $poolCandidate->pool?->team,
            $poolCandidate->pool?->community?->team,
            $poolCandidate->pool?->department?->team,
        ], 'view-team-applicationStatus');
   }
}

class PoolPolicy
{
   use ChecksTeamPermissions;

   public function view(User $user, Pool $pool)
   {
      if ($user->isAbleTo('view-any-pool')) return true;

      return $this->checkTeamPermissions($user, [
        $pool->team,
        $pool->community?->team,
        $pool->department?->team
      ], 'view-team-pool');
   }
}
```

### ✅ Acceptance Criteria

- [ ] Team permission check centralized
- [ ] Optional: Static cache

esizer:
🤖 Resolves #16131 

## 👋 Introduction

Adds a trait to check team permissions, mainly for use in policies.

## 🕵️ Details

This trait is meant to accomplish the common pattern of checking the same permission across multiple teams. since this is commonly used for pools, it also includes a helper to just get all the different teams off a pool.

It also has a pre-request cache so if you pass it the same user, team and permission in a single request, it doesnt have to ask laratrust again and just returns the value in the static property.

## 🧪 Testing

> [!TIP]
> This might be easier just running the graphql operations directly in graphiql than trying toi find where they are used in the app. 

1. Login as different roles and confirm the policies still function


api/app/Traits/ChecksTeamPermissions.php
```diff
@@ -0,0 +1,85 @@
...
+ */
+trait ChecksTeamPermissions
+{
+    /**
+     * Per-request cache for permission checks.
+     * Static variable is reset on each new PHP request.
+     *
+     * @var array<string, bool>
+     */
+    protected static array $teamPermissionCache = [];
```
petertgiles: Laravel has an in-memory single-request caching mechanism already built-in.  I've never used it before but it sounds like we're you're doing here.
https://laravel.com/docs/12.x/cache#cache-memoization

api/app/Policies/PoolPolicy.php
```diff
@@ -204,10 +179,15 @@ public function publish(User $user, Pool $pool)
...
+        if ($user->isAbleTo('update-any-publishedPool')) {
+            return true;
+        }
+
+        // Note: this checks both pool team and community team, but not department team
         $teamPermission = ! is_null($pool->team) && $user->isAbleTo('update-team-publishedPool', $pool->team);
         $communityPermission = ! is_null($pool->community?->team) && $user->isAbleTo('update-team-publishedPool', $pool->community->team);
 
-        return $user->isAbleTo('update-any-publishedPool') || $teamPermission || $communityPermission;
+        return $teamPermission || $communityPermission;
```
petertgiles: You could still use the new helper here, right?  Just pass in the array of two teams instead of using `getPoolTeams`?

petertgiles: Same comment for closePool.

api/tests/Unit/ChecksTeamPermissionsTraitTest.php
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    #[DataProvider('permissionProvider')]
+    public function testCheckTeamPermissionLogic(string $userProp, bool $usePoolTeams, bool $expected): void
+    {
+        $user = $this->{$userProp};
+        $reflection = new \ReflectionClass($this->policy);
+        $checkMethod = $reflection->getMethod('checkTeamPermission');
+
+        if ($usePoolTeams) {
```
petertgiles: This conditional logic makes the test a little hard to follow.  Could you split out the one case with special handling to its own test?

petertgiles (commented):
I've read through the code now.  I really like it!  I've left a few comments but nothing blocking.

I'm trying to figure how to test this properly.  Since it's permissions related I don't want to wing it.  😵‍💫 

2^3 permission permutations * 20 or 30 policy methods touched = a lot!

api/tests/Unit/Policies/PolicyTestCase.php
```diff
@@ -0,0 +1,47 @@
...
+    protected function createContextualUser(string $method, Pool $pool): User
+    {
+        if (in_array($method, ['asAdmin', 'asGuest', 'asApplicant'])) {
+            return User::factory()->{$method}()->create();
+        }
+
+        $id = match ($method) {
+            'asProcessOperator' => $pool->id,
+            'asDepartmentAdmin', 'asDepartmentHRAdvisor' => $pool->department_id,
+            default => $pool->community_id,
```
petertgiles: I'd suggest defining all the community roles and an exception for default to fail loudly.

api/tests/Unit/Policies/PoolPolicyTest.php
```diff
@@ -0,0 +1,491 @@
...
+        $otherPool = Pool::factory()
+            ->for(Community::factory()->create())
+            ->for(Department::factory()->create())
+            ->create();
+        $user = $this->createContextualUser($factoryMethod, $otherPool);
+
+        $this->assertFalse($this->ensureBool($this->policy->view($user, $this->primaryPool)));
+    }
+
+    // --- create() ---
```
petertgiles, line 96: The combination of department and community is kind of tricky for creating pools. It's not clear to me if these use cases are covered (both positive and negative):
- Community is optional for new pools
- You can create a pool in any community or department AS LONG AS you have permissions for EITHER that community OR that department. Essentially, "you must be able to administer the pool you're creating"

petertgiles (approved):
I think this is good?  😅   I really like the changes to the policies.  I think it's a really nice, clean refactor.  The new tests _look_ really good.  They seem thorough but it's so big that I'm not sure I can confidently say everything is covered.

I'm going to mark as approved but I'd really like a second approval before it gets merged.

api/tests/Unit/Policies/PolicyTestCase.php
esizer: That sounds like a good idea, what should we do for platform admins though? They are allow to interfact with these policies and was meant to be caught by the defualt.

Should we just include it along side the community roles?

petertgiles: Yeah, add all the known cases and let the default just catch gaps.

api/tests/Unit/Policies/PoolPolicyTest.php
esizer: oh good catch, I hope I added the missing ones? :sweat_smile: 

Updated in [add some missing test cases](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16655/commits/92b60dd7a75a4c0f2a918bac899293ad224c92d3)

api/tests/Unit/Policies/PolicyTestCase.php
esizer: Cool! Updated in [update contexual user helper](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16655/commits/4e384cd976a478418757d2365f53e28874e98b16)

petertgiles: Nice!

api/tests/Unit/Policies/PoolPolicyTest.php
petertgiles: Looks good!

api/app/Policies/PoolPolicy.php
```diff
@@ -163,13 +143,8 @@ public function updatePublished(User $user, Pool $pool)
...
 
-        $communityPermission = ! is_null($pool->community?->team) && $user->isAbleTo('update-team-publishedPool', $pool->community->team);
-
-        if ($communityPermission) {
-            return true;
-        }
-
-        return false;
+        // For updatePublished, only community team permission is checked
+        return ! is_null($pool->community?->team) && $user->isAbleTo('update-team-publishedPool', $pool->community->team);
```
tristan-orourke, line 148: Is it necessary to only check community permissions now for this? I think it should already be governed by the roles and permissions matrix. I imagine a future where we'd want to check for this permission using Departments as well.

api/app/Policies/PoolPolicy.php
```diff
@@ -187,10 +162,7 @@ public function publish(User $user, Pool $pool)
             return true;
         }
 
-        if (
-            (! is_null($pool->community?->team) && $user->isAbleTo('publish-team-draftPool', $pool->community->team))
-            || (isset($pool->department->team) && $user->isAbleTo('publish-team-draftPool', $pool->department->team))
-        ) {
+        if ($this->checkTeamPermission($user, [$pool->community->team, $pool->department->team], 'publish-team-draftPool')) {
```
tristan-orourke: Same here, is there any reason not to use $this->getPoolTeams($pool) ?
```suggestion
        if ($this->checkTeamPermission($user, $this->getPoolTeams($pool), 'publish-team-draftPool')) {
```

api/app/Policies/PoolPolicy.php
```diff
@@ -204,10 +176,12 @@ public function publish(User $user, Pool $pool)
...
     {
-        $teamPermission = ! is_null($pool->team) && $user->isAbleTo('update-team-publishedPool', $pool->team);
-        $communityPermission = ! is_null($pool->community?->team) && $user->isAbleTo('update-team-publishedPool', $pool->community->team);
+        if ($user->isAbleTo('update-any-publishedPool')) {
+            return true;
+        }
 
-        return $user->isAbleTo('update-any-publishedPool') || $teamPermission || $communityPermission;
+        // Note: this checks both pool team and community team, but not department team
+        return $this->checkTeamPermission($user, [$pool->team, $pool->community->team], 'update-team-publishedPool');
```
tristan-orourke: Same question again, can we use the generic helper here as well?

api/app/Policies/PoolPolicy.php
```diff
@@ -217,10 +191,12 @@ public function changePoolClosingDate(User $user, Pool $pool)
...
```
tristan-orourke: Same question again, can we use the generic helper here as well?

api/app/Traits/ChecksTeamPermissions.php
```diff
@@ -0,0 +1,84 @@
...
+    protected function checkTeamPermission(User $user, array $teams, string $permission): bool
+    {
+        foreach ($teams as $team) {
+            if (is_null($team)) {
+                continue;
+            }
+
+            $cacheKey = $this->cacheKey($user, $team, $permission);
+
+            $hasPermission = Cache::memo('array')->remember(
```
tristan-orourke, line 39: great idea on the caching ❤️

api/tests/Unit/Policies/AssessmentResultPolicyTest.php
```diff
@@ -0,0 +1,93 @@
...
+            'community admin on team' => ['asCommunityAdmin'],
+            'department admin on team' => ['asDepartmentAdmin'],
+            'department hr advisor on team' => ['asDepartmentHRAdvisor'],
+        ];
+    }
+
+    #[DataProvider('authorizedTeamRoleProvider')]
+    public function testAuthorizedTeamRolesCanAccessResources(string $factoryMethod): void
+    {
+        Cache::store('array')->clear();
```
tristan-orourke: Does this need to be here? It's already part of the setUp method

api/tests/Unit/Policies/PoolSkillPolicyTest.php
```diff
@@ -0,0 +1,114 @@
...
+        $this->pool = Pool::factory()
+            ->for(Community::factory()->create())
+            ->for(Department::factory()->create())
+            ->withPoolSkills(1, 0)
+            ->create();
+        $this->poolSkill = $this->pool->poolSkills->first();
+        Cache::store('array')->clear();
+    }
+
+    public static function allTeamRolesProvider(): array
```
tristan-orourke: Might make sense to move this to PolicyTestCase as well

api/app/Policies/PoolPolicy.php
tristan-orourke: And if there's a reason not to, should probably use the null-safe getter for all of these cases: $pool->community?->team

api/tests/Unit/Policies/PoolSkillPolicyTest.php
tristan-orourke: Should this include the Community Talent Coordinator role?

tristan-orourke (changes requested):
Looks great, but have a few questions

api/app/Policies/PoolPolicy.php
esizer: So, most of these questions seem similar to this and honestly.. I can't answer it :sweat_smile: I was just following the existing checks which did not include departments.

api/tests/Unit/Policies/AssessmentResultPolicyTest.php
esizer: I don't think it is part of it is it? This was here because this method is specifically asserting the cache results. The other tests don't really care so I left it out of those.

esizer: Wait, nevermind. I totally read the file name wrong lol. Moved in [move cache clear to base test setup](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16655/commits/4fd9585af8c164173ba0fa1110e240ea0fd04704)

api/tests/Unit/Policies/PoolSkillPolicyTest.php
esizer: Makes sense, moved to base class in [move generic provider to base class](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16655/commits/9b4977338b3834cf971b0501b587899d9b55f047)

Howeever, community talent coordinator actually fails the tests these are used in. I added a not as to why it was not included though.

api/app/Policies/PoolPolicy.php
tristan-orourke: I figured. In that case, go ahead and switch all these cases to use the same generic helper as everywhere else.

tristan-orourke (changes requested):

api/app/Policies/PoolPolicy.php
esizer: Are we sure? I think  @vd1992 mentioned that we explictly did not include department based teams on most of these checks. We might want to confirm before giving them permissions since :sweat_smile:

vd1992: What-a-what
I'm not sure what's going on. But if it is permissions themselves you want to change, definitely not something to sneak into this issue. 

Related to
https://github.com/GCTC-NTGC/gc-digital-talent/issues/16392 
I believe?

esizer: oh thanks, yeah! That was what I was thinking of, I just couldnt remeber what ticket it was brought up in :stuck_out_tongue:

tristan-orourke: I'd say there's two related questions here:
1. should we change the permissions assigned to roles ? As @vd1992 says, not in this ticket. (But yes we should, separately, see [Gray's comment](https://github.com/GCTC-NTGC/gc-digital-talent/issues/16392#issuecomment-4441173080))
2. should we _check_ for all the permissions that _could_ be assigned to roles? I think we should. That's the change I'm suggesting.

esizer: oh okay, I think I understand the request now. Thanks for clarfying. You are suggesting, even though we know they will fail the check, we do it anyway so that when we do the permission chantges, we don't need to come back and add the new checks here? That makes sense :+1:

esizer: Updated in [use pool helper in additional methods](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16655/commits/c1cd814b7eb9f8fe4a0fabb8b9e806b364def102)

tristan-orourke (approved):
👍

PR #16678, opened 2026-04-30
[Dep] Bump eslint react hooks to 7.1.1

### LINKED ISSUE #16609: 🛠️ Bump `eslint-plugin-react-hooks` to `>= 7.1.0`
### 🛠️ Platform

The latest version of `eslint-plugin-react-hooks` adds support for eslint v10.

### 🕵️ Details

https://github.com/facebook/react/releases/tag/eslint-plugin-react-hooks%407.1.0

This has some breaking changes for us, specificcally the improvments to `set-state-in-effect` which catches more lint errors for us.

### 🙋‍♀️ Proposed Solution

1. Bump `eslint-plugin-react-hooks` to at least version `7.1.0`
2. Address any new lint errors that arise

### ✅ Acceptance Criteria

- [ ] `eslint-plugin-react-hooks` is on at least `7.1.0`
- [ ] No lint errors

esizer:
🤖 Resolves #16609 

## 👋 Introduction

Upgrades `eslint-plugin-react-hooks` and addresses some, but not all issues. Some were spun off as they were too disruptive.

## 🕵️ Details

The lint rules got much better and stricter so it caught alot of anti-patterns. I'll do my best to explain them here.

## 🕵️ Details

The upgrade to the latest version of `eslint-plugin-react-hooks` has introduced a significantly more rigorous analysis of the React rendering lifecycle.

> [!NOTE]
> This makes some pretty significant changes to some patterns we were using so I did my best to explain why it was worth doing this. Its mostly about ensurring predictibility and purity of the components. 

### Synchronous State and Derived Data

*   **Components**: `SkillTable.tsx`, `NotificationDialog.tsx`, and `OrganizeSection.tsx`.
*   **Problem**: These components previously utilized `useEffect` to synchronize local state with incoming props or other state changes.
*   **Fix**: Refactored to utilize **derived state** with `useMemo` or by **adjusting state during render**.

When a component updates its _own_ state inside a `useEffect` call without an asynchronous trigger, it forces the browser to complete a full render cycle before immediately triggering a second render to account for the new state. 

This cascading render pattern creates a performance bottleneck and can lead to visible flickering in the user interface as the component settles into its final state. By calculating these values during render, the component reaches its final state in a single, atomic pass, reducing the total work performed by the rendering engine.

### Ref Management and Render Purity

*   **Components**: `NavMenu.tsx` and various navigation utilities.
*   **Problem**: Logic accessed `ref.current` during the render phase to determine styling or conditional behavior.
*   **Fix**: Moved ref interactions into `useEffect` or event handlers to ensure the DOM is stable.

React expects the render function to be a **pure computation** that does not observe the state of the DOM or other mutable items outside of state. Reading from a ref while rendering is dangerous because the value of that ref may change during the render process without React being aware of the update, leading to a UI that is out of sync with the underlying document. By moving these checks into effects, we ensure that the component only interacts with the DOM once it has been successfully committed and the nodes are in a stable state.

### Deterministic Output and Idempotency

*   **Components**: `JobCard.tsx`.
*   **Problem**: Use of non-deterministic values like `Date.now()` within the render body.
*   **Fix**: Captured a stable timestamp once on mount using **lazy state initialization**.

Maintaining render purity requires that a component remains **idempotent**, meaning it must produce the same JSX output every time it is called with a specific set of props and state. Introducing values that change every millisecond causes a component to produce different outputs during a single render pass or fail to stay in sync with other components. This stabilization is critical for ensuring that the UI remains consistent across different rendering environments and re-renders.

Added a `useStableDate` hook  to hipefully make it easier to remeber to pattern. We chould now prefer this for getting the current timestamp rather than directly using `Date.now` in the main component body.

### Functional Data Transformations and Immutability

*   **Components**: `TalentEvent/Navigation.tsx`.
*   **Problem**: Reassignment of local variables within render loops to track state across iterations.
*   **Fix**: Refactored logic to use **pure functional transformations** like `slice` and `map`.

Reassigning variables declared in the component scope during a render pass is an impure side effect that makes the render output dependent on the order and execution of internal logic. This pattern leads to fragile code that can break if React interrupts and restarts a render pass. By deriving information purely from source data using immutable methods, we guarantee that the data flow remains predictable and the component output is always idempotent, making the code significantly more resilient to future changes in React.

### Dependency Stability and Closure Integrity

*   **Components**: `useDeepCompareEffect.ts`.
*   **Problem**: Unstable function or object references in dependency arrays causing excessive effect execution.
*   **Fix**: Stabilized references using `useMemo` and implemented **deep equality checks** for custom hooks.

Using functions or objects that are recreated on every render as effect dependencies causes those effects to run far too frequently. This is particularly problematic for components managing states where over-execution impacts performance and the user experience. By ensuring dependency arrays only contain stable references, we keep the application responsive and ensure that side effects only trigger when the underlying data has truly changed in a meaningful way.

## 🧪 Testing

> [!IMPORTANT]
> I did my best to give very clear instructions on what to test. It looks like a lot but I hope it is manageable. If you feel like something is risky or it is too much, I can try and break this up further. 

1. Build `pnpm dev:fresh`
2. Confirm it builds and no lint erors `CI=true pnpm lint --force`
3. Test the changes noted:

- [x] Process activity log renders proper component for activity type
- [x] Error summary still renders when expected and moves focus to it
- [x] JobCard still renders deadline properly
- [x] Sitewide banner publish and expiry date still work
- [x] Notification dialog opens and closes as expected
- [x] Job poster template filters still function (`useDeepCompareEffect`)
- [x] IAP theme still loads and changes back when going to main app
- [x] Talent nominations still scrolls to top when changing pages
- [x] Talent nomination navigation still functions as expected
- [x] Departments manage access roles dialog still works
- [x] Confirm the process assessment step builder still udpates when adding/removing
- [x] Confirm skills library filters still function
- [x] Appropriate menu (narrow/wide) appears on page load
- [x] Language swtiching still works
- [x] Storybook theme switcher works
- [x] Nominiations in the admin render to proper icon in the  "Nominated by" accordion
- [x] Table of contents links still scroll you to the section
- [x] Nav menu item still recieve proper active styles
- [x] Default notice icons are still renderd properly
- [x] Icon in application stepper still shows success and error states


apps/web/src/components/NotificationDialog/NotificationDialog.tsx
```diff
@@ -55,24 +55,26 @@ const DialogPortalWithPresence = ({
...
-    } else {
-      // let animation complete before removal
-      const timerId = setTimeout(() => {
-        setRender(false);
-      }, 200);
-      return () => clearTimeout(timerId);
-    }
+    if (isPresent) return undefined;
+
+    const timerId = setTimeout(() => {
```
petertgiles, line 74: Could you add a comment here about what this timeout is for?

packages/ui/src/components/Stepper/Step.tsx
```diff
@@ -52,6 +55,11 @@ interface StepProps
   label: ReactNode;
 }
 
+const ICON_MAP: Record<string, IconType | undefined> = {
```
petertgiles: Could you tighten up the key type and avoid a null coalesce on L71?

petertgiles (approved):
Wow, nice!  The code looks good and I was able to complete the testing checklist.

apps/web/src/components/NotificationDialog/NotificationDialog.tsx
esizer: Oh yeah, we had one but I accidentally removed it :woman_facepalming: 

Brought it back in [bring back comment for timeout](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16678/commits/e77167604ffd671534e3f6a197bfb63214493986)

packages/ui/src/components/Stepper/Step.tsx
esizer: Yeah, thats a good idea. Updated in [improve types](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16678/commits/2a4c8b80610754e4884dab69426bfb2a4abf9ccf)

PR #16684, opened 2026-05-01
[Feature] Back up emails instead of mangling

### LINKED ISSUE #16643: ✨ Backup email addresses instead of mangle
### Details

When a user signs in and their email address is already on an older account, the email address is mangled on the older account to prevent a unique constraint collision.  However, we need to be able to find that old account for migration and that is hard if the address is manged.  Instead, let's copy the email address to a separate field, unchanged, for matching.

### Design file & designer

_No response_

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

_No response_

### ✅ Acceptance Criteria

- [ ] new fields to hold backups of the email and work email
- [ ] new fields are not directly mutable or visible through the API
- [ ] when an account is deleted (deleted-at mangle) or taken (taken-at mangle):
    - [ ] email addresses are moved to backup instead of mangled
    - [ ] email address fields are blanked

petertgiles:
🤖 Resolves #16643 

## 👋 Introduction

Backs up email addresses instead of mangling when deleting or when a collision occurs on log in.

### Delete / Restore

> [!NOTE]
> You can use mock auth for this part

 Handy SQL query :point_down: 
```sql
select email, email_verified_at , email_backup, work_email, work_email_verified_at, work_email_backup  from users where sub = 'applicant-employee@test.com'
```

1. Log in as admin@test.com
2. Pick a user with email address and work email address et
3. Navigate to the admin advanced tools tab for that user

- [ ] Can archive user
- [ ] User's email addresses were backed up
- [ ] Can restore user
- [ ] User's email addresses were restored
- [ ] User's email addresses are unverified

### Login collisions

> [!IMPORTANT]
> You must use CanadaLogin for this part

1. Pick a secondary user and manually set the email address to match your primary CanadaLogin user
2. Log in as that CanadaLogin user

- [ ] Primary user has the verified email address
- [ ] Secondary user's email was backed up

### DIFF
--- a/api/app/Http/Controllers/AuthController.php
+++ b/api/app/Http/Controllers/AuthController.php
@@ -180,13 +180,16 @@ class AuthController extends Controller
                     $existingUser = User::where('id', '!=', $userMatch->id)
                         ->where(fn ($subquery) => $subquery
                             ->where('email', 'ilike', $incomingEmailAddress)
-                            ->orWhere('work_email', 'ilike', $incomingEmailAddress)
-                        )->first();
+                            ->orWhere('work_email', 'ilike', $incomingEmailAddress))
+                        ->withTrashed()
+                        ->first();
                     if (strcasecmp($existingUser->email, $incomingEmailAddress) == 0) {
-                        $existingUser->email = $existingUser->email.'-taken-at-'.Carbon::now()->timestamp;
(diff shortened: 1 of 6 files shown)


esizer (approved):
LGTM! 👍

PR #16692, opened 2026-05-04
[Feature] Introduce soft deleting of community development programs

### LINKED ISSUE #16654: ✨  Remove professionalization back-end work
### Details

Adding back-end functionality so community talent managers can remove development programs (professionnalisations) from their community.

This should hide the development program from the list on the community pages. It should also stop the development program from being encountered in the "Add community" flow and in the "nominations" flow.

However, we still want past nominations to show this development program AND the associated data submitted during the nomination.

The front-end work for this is described in #16612 

### Proposed Implementation

The challenge here is that we use a community-development-programs-talent-nomination pivot. So we can't simply delete the community-development-program record.

#### Option 1 - soft-delete/archive?
We could handle this by soft-deleting (or similar) the community-development-program, as long as it still shows up by default for past nominations.

#### Option 2 - refactor nominations to link directly to development programs
~­If we did a refactor to change community-development-programs-talent-nomination pivot to a development-programs-talent-nomination pivot, we might just be able to delete community-development-program without affecting nominations.~ Actually we would also have to address the talent-event-community-development-program pivot, so I don't think this is viable.

#### Option 3 - Archiving
Set an archive flag on the community-development-program model and just code all the functionality needed to meet the acceptance criteria below.

### ✅ Acceptance Criteria
- [ ] Removed professionnalisations are not encountered during community interest flow
- [ ] Removed professionnalisations are not encountered when creating a new talent nomination event
- [ ] Removed professionnalisations CONTINUE to show up for submitted talent nominations
- [ ] When viewing users (or downloading), the community roles should no longer see data about this professionnalisation in the community interest.

vd1992:
🤖 Resolves #16654

## 👋 Introduction

Change removal of "professionalizations" to soft deletion. 
Then make changes elsewhere as needed to preserve expected functionality. Including turning the create mutation into a create-or-restore 

## 🧪 Testing

For testing, rather than instructions. 
See intended state here: https://github.com/GCTC-NTGC/gc-digital-talent/issues/16654#issuecomment-4346610122
Then approach as you feel best


vd1992:
What do you think of this so far, @esizer

api/app/GraphQL/Mutations/CreateOrRestoreCommunityDevelopmentProgram.php
esizer, line 1: Feel free to ignore becuase it is a quite the nit-pick. But, I see we have the same code syncing the classifications in two different spots. We could rearragne this to only have it once like:

```php
if($existingRecord) {
  $existingRecord->restore();
  $record = $existingRecord;
} else {
  $record = CommunityDevelopmentProgram::create();
}

if(...) {
  $record->classifications()->sync(...);
}
```

vd1992: Okay, sure, why not
https://github.com/GCTC-NTGC/gc-digital-talent/pull/16692/changes/73379c0f6d48548d6f6ac067c0a3e4be0958b65c

vd1992:
@copilot 
Where is your review?

vd1992:
And the review summary it claims to post but also refuses to 

image

esizer (approved):
LGTM! 👍

apps/web/src/pages/TalentNominations/NominateTalent/components/Details.tsx
```diff
@@ -581,7 +581,7 @@ const NominateTalentDetails_Fragment = graphql(/* GraphQL */ `
       value
     }
     lateralMovementOptionsOther
-    communityDevelopmentPrograms {
+    communityDevelopmentPrograms(trashed: WITH) {
```
brindasasi, line 584: we shouldn't allow new nominations to select soft deleted dev. programs rt?
then here we shouldn't add this arg

apps/web/src/pages/TalentNominations/NominationGroup/components/TalentNominationAccordionItem.tsx
```diff
@@ -92,10 +94,12 @@ const TalentNominationAccordionItem_Fragment = graphql(/* GraphQL */ `
     }
     lateralMovementOptionsOther
 
-    developmentPrograms {
-      id
-      name {
-        localized
+    communityDevelopmentPrograms(trashed: WITH) {
```
brindasasi, line 97: Here also we shouldn't call soft deleted ones I believe .. check it out

apps/web/src/pages/TalentNominations/NominationGroupEvaluationDialog/components/DevelopmentProgramsSection.tsx
```diff
@@ -26,19 +26,24 @@ const NominationGroupEvaluationDialogDevelopmentPrograms_Fragment = graphql(
...
           }
         }
         developmentProgramOptionsOther
       }
       talentNominationEvent {
-        developmentPrograms {
-          id
-          name {
-            localized
+        communityDevelopmentPrograms(trashed: WITH) {
```
brindasasi, line 41: Here also not required

brindasasi:
@vd1992 double check my findings.. as far as I understand , we should preserve the history but prevent the new nominations with these deleted programs.

apps/web/src/pages/TalentNominations/NominateTalent/components/Details.tsx
vd1992: Disagree
For consistency in nominations, all potential nominations should have access to the same pool of programs attached to an event

apps/web/src/pages/TalentNominations/NominationGroup/components/TalentNominationAccordionItem.tsx
vd1992: This is fetching programs attached to a nomination, it absolutely should fetch soft deleted to be accurate

apps/web/src/pages/TalentNominations/NominationGroupEvaluationDialog/components/DevelopmentProgramsSection.tsx
vd1992: Disagree, without fetching soft deleted programs on the nominations. You miss content that was submitted

brindasasi:
> @vd1992 double check my findings.. as far as I understand , we should preserve the history but prevent the new nominations with these deleted programs.

nvm .. I was wrong on this .. Gray mentioned this 

> Yeah, the dev program should persist in a Talent Event that already has it, even if it is removed from the community.
> 
> That's why we carefully noted that it only affects the "create" flow for talent events.
>

brindasasi (approved):
LGTM!

PR #16696, opened 2026-05-05
[Bug] Set CSP on index.html only

### LINKED ISSUE #16690: 🐛 No CSP Headers in Prod

It seems like our CSP headers are missing in prod.
Also, CSP reports are being blocked by the firewall.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] CSP headers appear in prod
- [ ] CSP reports can be submitted in prod

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

petertgiles:
🤖 Resolves #16690 

## 👋 Introduction

Adjusts the Ngnix conf to send CSP headers in production.

## 🕵️ Details

Nginx conf is a little funny.  If you set headers at the server level (like the CSP headers) and also set headers at the location level (like the cache headers) then the server-level headers are wiped and totally replaced with the location-level headers.  

I've solved this by pulling the CSP and cache headers out of the server level and moving them to the location blocks.  The CSP header is just the single location for index.html. As a side effect, the CSP headers are no longer sent for assets, API requests, and robots.txt.  I think that's fine.

We had some fancy logic to try and preserve upstream cache headers from PHP.  I removed that as unnecessary now that there are no server-level cache headers to overwrite them.

I made a few cosmetic changes to try and reduce the diff between the conf files a bit more, too.

## 🧪 Testing

> [!TIP]
> As of 2026-05-05, this PR is deployed on the dev vertical. 

1. Test locally and in Azure.
2. CSP headers are sent only on index.html responses.
3. Different cache headers are sent on index.html, *.js, *.webp, and graphql respnses.

### DIFF
--- a/infrastructure/conf/nginx-conf-deploy/default
+++ b/infrastructure/conf/nginx-conf-deploy/default
@@ -1,9 +1,3 @@
-
-map $request_uri $csp_header {
-    default "default-src 'self'; font-src fonts.gstatic.com 'self'; style-src-elem fonts.googleapis.com 'self' 'nonce-$cspNonce' 'unsafe-inline'; connect-src 'self' https://canadacentral-1.in.applicationinsights.azure.com https://js.monitor.azure.com ws://$host wss://$host; script-src-elem 'self' 'nonce-$cspNonce'; img-src 'self' 'nonce-$cspNonce' data:; frame-ancestors 'none'; report-uri /api/csp-report/; report-to csp;";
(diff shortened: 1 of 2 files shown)


esizer (approved):
Awesome! 😎

PR #16712, opened 2026-05-06
[Fix] Make post deployment idempotent

### LINKED ISSUE #16646: ✨ Post_deployment.sh - Make safe to rerun
### Details

In some situations we may want to rerun [post_deployment.sh](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/infrastructure/bin/post_deployment.sh) in an app service.  We should make sure that it can be safely rerun.

We know the crontab can get loaded multiple times, for example:

Image

### Design file & designer

_No response_

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

_No response_

### ✅ Acceptance Criteria

- [ ] post_deployment.sh can be safely run multiple times in the app service

esizer:
🤖 Resolves #16646 

## 👋 Introduction

Updates the `post_deployment.sh` script idempotent in case we ever need to re-run it on demand.

## 🕵️ Details

Previously, we were appending onto the end of both the `crontab` and `php.ini`. This was fine for a single run, but when we do this a second time, it was duplicating things since it just kept appending items on the end.

This updates both those lines to use the drop-in directories with custom files that are loaded. We then overwrite the files entirely on runs since it should be safe to do so (only our contents go in them).

### Question/Future

I wonder if instead of writing these files, we just keep them in the `./infrastructure` folder and copy them into the correct location instead? :thinking: Anyway, this should work for now and be a lighter touch.

## 🧪 Testing

```sh
php -i | grep memory_limit
crontab -l -u www-data
```

1. Deploy to dev
2. SSH into dev machine
3. Confirm the settings exist
4. Re-run `./post_deployment.sh`
5. Confirm no duplicates and settings are the exact same as step 3

### DIFF
--- a/infrastructure/bin/post_deployment.sh
+++ b/infrastructure/bin/post_deployment.sh
@@ -98,6 +98,7 @@ add_section_block "$TRIPLE_BACK_TICK $CLEANED_STDOUT $TRIPLE_BACK_TICK"
 if cat > /etc/cron.d/gc-digital-talent << 'EOF'
   *  *  *  *  * www-data . /etc/profile ; php /home/site/wwwroot/api/artisan schedule:run
 EOF
+then
     add_section_block ":white_check_mark: Laravel Scheduler cron setup *successful*."
 else
     add_section_block ":X: Laravel Scheduler cron setup *failed*. $MENTION"
--- a/infrastructure/bin/post_deployment.sh
+++ b/infrastructure/bin/post_deployment.sh
@@ -94,7 +94,10 @@ add_section_block "$TRIPLE_BACK_TICK $CLEANED_STDOUT $TRIPLE_BACK_TICK"
 
 # Load Laravel Scheduler cron
 # For extra debugging you can add `>> /tmp/run_laravel_scheduler.log 2>&1` to the end of the cron'd command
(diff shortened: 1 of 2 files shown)


petertgiles (approved):
Nice, this works really well.  I think for single lines, I prefer the approach of an inline echo over copying a file.

Technically the AC is met since it is safe to rerun so I'm approving.  There's a gotcha, though.  It _looks_ like there was a failure with supervisor and it would be easy to be confused by that.  Would you consider updating the ✔️ / ❌  for that step?  Instead of just using the exit code of supervisord (which fails if it's already running) maybe check afterwards if it's running by using `supervisorctl status` or `ps aux | grep supervisord`?

esizer:
> Technically the AC is met since it is safe to rerun so I'm approving. There's a gotcha, though. It _looks_ like there was a failure with supervisor and it would be easy to be confused by that. Would you consider updating the ✔️ / ❌ for that step? Instead of just using the exit code of supervisord (which fails if it's already running) maybe check afterwards if it's running by using `supervisorctl status` or `ps aux | grep supervisord`?

oh yeah, good call. I think I got it in [fix supervisor for re-runs](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16712/commits/247dcd297f49eaac4a47ca2ab83a8c50c8c56614)

petertgiles (approved):
Very cool!  The reload of conf files is a nice touch.

PR #16755, opened 2026-05-12
[Tooling] Add codespace setup script for sharing with designers

brindasasi:
## Summary

- Adds `maintenance/scripts/setup-codespace.sh` to configure a GitHub Codespace for external sharing
- Adds `make codespace-setup` target to the Makefile

## What it does

Running `make codespace-setup` inside a Codespace terminal will:

1. Patch `api/.env` — updates `APP_URL` and all OAuth redirect URIs to the Codespace URL
2. Patch `apps/web/.env` — updates `API_HOST` and logout redirect URLs
3. Patch nginx — fixes the mock OAuth server's `Host` header so login works from an external browser
4. Start containers if not running
5. Rebuild the frontend (`API_HOST` is baked into the JS bundle by Vite at build time, so a rebuild is required)
6. Clear Laravel config cache and reload nginx
7. Set port 8000 to Public via `gh` CLI if available

## Why

By default the app is configured for `localhost:8000`. Sharing a Codespace with designers requires all URLs to point to the Codespace's forwarded-port URL (`https://{name}-8000.app.github.dev`). Several of these are runtime config but `API_HOST` is compile-time, which is why a frontend rebuild is needed.

## Testing

- [ ] Run `make codespace-setup` inside a Codespace
- [ ] Confirm the app loads at the printed URL
- [ ] Confirm GCKey sign-in reaches the mock auth login page
- [ ] Confirm login with `admin@test.com` works and lands on the admin dashboard and do the same for other logins as well

### DIFF
--- a/maintenance/scripts/setup-codespace.sh
+++ b/maintenance/scripts/setup-codespace.sh
@@ -1,0 +1,147 @@
+#! /bin/bash
+
+# Configures a GitHub Codespace for external sharing.
+#
+# Patches api/.env, apps/web/.env, and the nginx mock-auth proxy to use the
+# Codespace's forwarded-port URL instead of localhost:8000, then rebuilds the
+# frontend (API_HOST is baked into the JS bundle at Vite build time), reloads
+# nginx, and clears Laravel's config cache so the changes take effect.
+#
+# Usage (run from repo root inside the Codespace terminal):
+#   bash maintenance/scripts/setup-codespace.sh
+#
+# Or via make:
+#   make codespace-setup
+
+parent_path=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
+source ${parent_path}/lib/common.sh
+
+# ---------------------------------------------------------------------------
+# Resolve the Codespace app URL
+# ---------------------------------------------------------------------------
+
+if [ -z "${CODESPACE_NAME:-}" ]; then
+  echo "Error: CODESPACE_NAME is not set."
+  echo "This script must be run inside a GitHub Codespace terminal."
+  exit 1
+fi
+
+FORWARDING_DOMAIN="${GITHUB_CODESPACES_PORT_FORWARDING_DOMAIN:-app.github.dev}"
+APP_URL="https://${CODESPACE_NAME}-8000.${FORWARDING_DOMAIN}"
+APP_HOST="${CODESPACE_NAME}-8000.${FORWARDING_DOMAIN}"
+
+echo "Configuring Codespace for external access"
+echo "  App URL: ${APP_URL}"
+echo ""
+
+# ---------------------------------------------------------------------------
+# api/.env
+# ---------------------------------------------------------------------------
+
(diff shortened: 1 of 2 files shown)


brindasasi:
an attempt to make codespace setup easier for devs working around the limitations we have in the code

maintenance/scripts/setup-codespace.sh
```diff
@@ -0,0 +1,147 @@
...
+docker compose run --rm maintenance bash refresh_frontend.sh
+
+echo "✓ Frontend rebuilt"
+
+# ---------------------------------------------------------------------------
+# Clear Laravel config cache + reload nginx
+# ---------------------------------------------------------------------------
+
+docker compose exec webserver sh -c \
+  "php /home/site/wwwroot/api/artisan config:clear && php /home/site/wwwroot/api/artisan cache:clear"
```
esizer: This is failing for me, I think it needs to be run in the maintenance container? 🤔 

Actually, maube this should be `docker compose run --rm maintenance bash refresh_api.sh` and run before the frontend?

esizer (changes requested):
I think this is a good start! Though,. I can't actually get it working but it should be a simple fix (I think).

Though, im wondering if we could use  [devcontainers](https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/adding-a-dev-container-configuration/introduction-to-dev-containers) eventually since they seem to be doing something similar 🤔

maintenance/scripts/setup-codespace.sh
vd1992: This suggestion appears good

maintenance/scripts/setup-codespace.sh
vd1992, line 1: With Eric's suggestion, build completes

Unable to login however

Image

maintenance/scripts/setup-codespace.sh
vd1992: Just missing seeding after the API refresh I believe

brindasasi:
could you verify again? @vd1992

maintenance/scripts/setup-codespace.sh
vd1992: Build + seed resolved, this, still a problem

esizer: oh yeah, it is missing the app key. We will need to make sure we also run the secret update script. 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/maintenance/scripts/update_env_secrets.sh

I think we could follow what happens in setup :thinking: 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/832c65892de030c921acd10e8e847d25f462a013/maintenance/scripts/setup.sh#L33-L38

esizer (approved):
Wrosk for me! 🎉

vd1992 (approved):
1. Create a space
2. Run script (takes a while but it succeeds)
3. Can navigate site + login + execute a test mutation

Nice!

PR #16804, opened 2026-05-14
[Bug] Conditional opening date handling in event creation and updating

### LINKED ISSUE #16802: 🐛 Talent event open date timestamp is not correct

Creating an talent event with opening date today is saved as tmo in DB. 
I created this event today May 14 and it is set to open tmo 
Image

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Login as community admin 
2. Create an event and set the opening date as today 
3. Go to Talent Event Table 
4. Observe the event you just created is set to open tmo 

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

[ ] If the talent event opening date is today it should be live immediately
[ ] If the talent event opening date is in future , it should be live on that specific date 

_No response_

vd1992:
🤖 Resolves #16802

## 👋 Introduction

Adjust how opening dates are handled when submitting. 

-> Open date set to today, immediately open it via now in UTC
-> Open date set to future date, opens just after midnight Pacific equivalent

## 🧪 Testing

1. From `/en/admin/talent-events`
2. Exercise creating/updating events with varying opening dates

### DIFF
--- a/apps/web/src/pages/TalentEvents/CreateTalentEventPage.tsx
+++ b/apps/web/src/pages/TalentEvents/CreateTalentEventPage.tsx
@@ -17,7 +17,11 @@ import { graphql } from "@gc-digital-talent/graphql";
 import { toast } from "@gc-digital-talent/toast";
 import { commonMessages, navigationMessages } from "@gc-digital-talent/i18n";
 import { Submit } from "@gc-digital-talent/forms";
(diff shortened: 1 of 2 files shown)


brindasasi (approved):
LGTM!

PR #16815, opened 2026-05-15
[Feature] in app migration query

### LINKED ISSUE #16798: ✨ In-App Migration - Account Matching
### Details

We need an API call for users to find out if they can migrate an account.  This issue will result in that API call.  No actual migration or frontend work is in scope for this issue.

### Design file & designer

_No response_

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

_No response_

### ✅ Acceptance Criteria

- [ ] Users can contact the API to find if they can migrate an account
    - [ ] Can only request for "self", not any other user
- [ ] The API will return a boolean value (true/false) if a migration is possible,
    - [ ] no identifiable info of other account is returned
- [ ] A migration is possible if:
    - [ ] The `feature.auth_in_app_migration` flag is set to true
    - [ ] There is exactly one possible migration target
    - [ ] A possible migration target is defined as:
        - [ ] target account has the backup contact email matching (normalized)
        - [ ] target account has the telephone field matching (normalized)
        - [ ] target account was not last logged into with CanadaLogin
        - [ ] target account is not soft-deleted
- [ ] unit testing to confirm functionality

petertgiles:
🤖 Resolves #16798 

## 👋 Introduction

Adds a query for users to discover if it is possible to migrate.

## 🕵️ Details

Refer to the attached issue for the matching algorithm.

## 🧪 Testing

> [!IMPORTANT]
> There is no UI for this feature yet.  You can use this query:
query test { canMigrateMyAccount } 

- [ ] if feature flag is turned off then it always returns false
- [ ] if there are multiple matching users then it returns false
- [ ] if there are zero matching users then it returns false
- [ ] if there is one matching user then it returns true
- [ ] matching is case insensitive and trims for email 
- [ ] matching is digits-only, ignore leading 0s and 1s, and trims for phone
- [ ] having 'canada.ca' in last_sign_in_iss for target prevents matching
- [ ] soft deleting target prevents matching


bella77-69 (commented):
Sorry for the time it took me to review @petertgiles this was new learning experience for me so I don't feel that confident approving on my own 🙃 That being said would any @GCTC-NTGC/developers mind taking a look as well??

esizer (approved):
LGTM! 👍

PR #16761, opened 2026-05-12
[Refactor] Community pages middleware addition

vd1992:
🤖 Resolves #16461 

## 👋 Introduction

Use the new middleware stuff to handle frontend role checks, and a few changes alongside 

## 🕵️ Details

Issue A/C wanted some unit testing, used Claude in Agents to put together some stuff. Went decently, for the most part

## 🧪 Testing

Assert the roles behaviour per issue body

1. Platform admin access all
2. Community admin view all three tabs for own community and can edit
3. Recruiter views two tabs for own community
4. Coordinator view two tabs for own community
5. All four can view the communities table and the information view of any community

Edit restricted to platform admin and community admin (own community)
Only platform can create communities 

## 📸 Screenshot

Edit link now conditionally rendered 😮 

image

PR #16829, opened 2026-05-20
[tooling] whitelist localhost & 403 from canada login

brindasasi:
## 👋 Introduction

[This PR ](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16761) has added localhost links into the tests. This causes Link Checker to fail as it tries to reach them and fails.
Hence it has been added to the list of whitelisted hosts to not to verify 

Also Canada Login hosts are still rejecting bots even after the browser header retry.. hence we treat 403 from those specific hosts are still valid.  If they are down any other error we will get 5xx , 404 etc.


esizer (approved):
🤞

PR #16840, opened 2026-05-20
[Fix] Clear space for vitest workflow

### LINKED ISSUE #16806: 🛠️ Vitest action running out of space
### 🛠️ Platform

The action is fialing due to running out of space in the runner.

### 🕵️ Details

I assume this is due to the fact that we fairly recently added playwright browsers so we could run true accessiblity tests in storybook for vitest.

The browsers are large and take up a lot of space. We likely hit a point where we may have added more the space needed for coverage goes up.

### 🙋‍♀️ Proposed Solution

We actualy ran into this same issue with the playwright action so there is an example of how to solve it there:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/538a9baef023aad3afa551c755c58c9e17f5c8bb/.github/workflows/playwright.yml#L48-L50

### ✅ Acceptance Criteria

- [ ] Vitest action no longer runs out of space

esizer:
🤖 Resolves #16806 

## 👋 Introduction

Updates the vitest workflow to delete unused files from the base image so we have space for coverage reports.

## 🕵️ Details

This also appends on `df -h` for a human readable printout of the free space on the runner in the workflow so we have info to better diagnose issues in the future.

SEE: https://man7.org/linux/man-pages/man1/df.1.html

## 🧪 Testing

> [!NOTE]
> This may be difficult to test since the space error is a little flaky. 

1. Run the workflow multiple times
2. Confirm no space errors
3. Confirm free space with the `df -h` command

### DIFF
--- a/.github/workflows/vitest.yml
+++ b/.github/workflows/vitest.yml
@@ -28,11 +28,10 @@ jobs:
           rm -rf /usr/share/dotnet
           rm -rf /usr/local/share/boost
           rm -rf /opt/ghc
-          rm -rf /usr/local/share/powershell
-          rm -rf /usr/share/swift
-          rm -rf /usr/local/.ghcup
-          rm -rf /usr/lib/jvm
-          # Display free space for future diagnosis
+          sudo rm -rf /usr/local/share/powershell
+          sudo rm -rf /usr/share/swift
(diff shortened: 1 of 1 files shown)


vd1992 (approved):
👍 

Image

PR #16842, opened 2026-05-20
[Fix] Nominations filename sanitization

### LINKED ISSUE #16831: 🐛 UAT | An error occurred when downloading talent nominations submitted within 24 hours

- When Talent event and talent nominations are submitted on the same day when requesting for the talent nominations download, an error message pops up restricting user to download the file. 

**Expected Outcome:** Talent nominations should get downloaded without any issues

### 🕵️ Details

Env: UAT

### 📋 Steps to Reproduce

Prerequisite: Create a new talent event and submit a new talent nomination

1. Access GC digital portal using community admin role
2. Navigate to Talent management 
3. Open the talent event which has been created on the same day
4. Navigate to nominations
5. Check the nomination submitted and download excel 
6. Verify the notification received 
7. Click on notification to download the file
8. Observe the error

### 📸 Screenshot

https://github.com/user-attachments/assets/266a879a-8e39-4936-925b-7038bbd93f40

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Community admin can download the talent nominations anytime even within 24 hours

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #16831 

## 👋 Introduction

Updates the file path santitization helper to replace periods with underscores for a safer filename.

## 🕵️ Details

The santization was messing up and creating filenames that could not be safely retrieved. This did two things:

1. Replaces all periods but the extension delimiter with underscores in the filename
2. Updates the file generation to use santization to strip out unwanted characters

## 🧪 Testing

1. Login as `community@test.com`
2. Create a talent event with periods in its title
3. Seed some nominees for that talrent event
4. Confirm you can download the nominees without error

### DIFF
--- a/api/app/Support/FilePath.php
+++ b/api/app/Support/FilePath.php
@@ -12,13 +12,13 @@ class FilePath
     {
         // Remove dots to prevent directory traversal and double extensions
         if ($preserveExtension && str_contains($input, '.')) {
-            $parts = explode('.', $input);
-            $extension = array_pop($parts);
-            // Replace internal dots with spaces, then join back to the extension
-            $name = implode(' ', $parts).'.'.$extension;
+            $lastDot = strrpos($input, '.');
+            $extension = substr($input, $lastDot + 1);
+            // Replace internal dots (e.g. version numbers like 2.72.2) with underscores
+            $name = str_replace('.', '_', substr($input, 0, $lastDot)).'.'.$extension;
         } else {
             // No dots allowed at all (prevents double extensions)
-            $name = str_replace('.', ' ', $input);
+            $name = str_replace('.', '_', $input);
         }
 
         // Keep French letters, numbers, spaces, and dashes
--- a/api/app/GraphQL/Mutations/DownloadNominationsExcel.php
+++ b/api/app/GraphQL/Mutations/DownloadNominationsExcel.php
@@ -5,6 +5,7 @@ namespace App\GraphQL\Mutations;
 use App\Generators\NominationsExcelGenerator;
 use App\Jobs\GenerateUserFile;
 use App\Models\TalentNominationEvent;
+use App\Support\FilePath;
 use Illuminate\Support\Facades\App;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Lang;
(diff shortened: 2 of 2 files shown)


api/app/Support/FilePath.php
vd1992, line 1: Given the source of the problem appears to have been pre-existing, does that affect the issue's scope?
To say, whether sanitation should be adjusted to allow for more characters? Such as for example commonly used in French

esizer: @mauli0110 What do you think?

mauli0110: > @mauli0110 What do you think?

I think yes, we need to include more characters specially used in French. We can take our time to fix this issue as this is not release blocker for today.

esizer: Have we tested with more characters? We do allow most and I feel like we may have heard if it was an issue (but maybe not 😅 ).

https://github.com/GCTC-NTGC/gc-digital-talent/blob/8b67720b7d81c6feed74ecb013e93628c3ad6824/api/app/Support/FilePath.php#L24-L27

mauli0110: > Have we tested with more characters? We do allow most and I feel like we may have heard if it was an issue (but maybe not 😅 ).
> 
> https://github.com/GCTC-NTGC/gc-digital-talent/blob/8b67720b7d81c6feed74ecb013e93628c3ad6824/api/app/Support/FilePath.php#L24-L27

I believe mostly 'Underscore (_)' is used alot in talent events. The 'period (.)' was exceptional😅

vd1992: Hmm, yeah, good point. Given we've yet to hear about the problem, it must not have come up yet. Then again, there hasn't been that many nomination events yet either I believe

mauli0110: Yes, exactly. They mostly use underscores (_) between the names. I'll check and test the different characters especially with French names to be absolutely sure.

mauli0110: I confirmed, it's working for different characters [specially french], in local. 

Screenshot 2026-05-21 160108


Screenshot 2026-05-21 160019

vd1992: Seems you have reviewed this well enough to hit approve

mauli0110 (approved):
LGTM 🎉

PR #16846, opened 2026-05-21
[Fix] CSP reporting

### LINKED ISSUE #16819: 🐛 CSP violations not being tracked

I noticed in the logs that we're getting a lot of HTTP 400s and was able to recreate it.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Log into prod
2. Pop the email verification dialog
3. Observe to HTTP 400

(It seems like Edge doesn't show CSP reports in the network tab.  You might have to use a real browser like Firefox. 😉 )

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

It seems like our endpoint will strip out anything that isn't under the `csp-report` top level key.  I bet this report has an unexpected shape.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/9d575e31adaa979bdd26d53ebdc1b39a52febce8/api/app/Http/Controllers/CspReportController.php#L18

### ✅ Acceptance Criteria

- [ ] CSP report controller accepts this valid report
- [ ] CSP violation is either fixed or tracked in a separate issue on the board

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #16819 

## 👋 Introduction

Fixes an issue with CSP violations not being reported.

## 🕵️ Details

There were a few issues I was able to determrine which should solve this for both local and deployed versions:

- The `report-to` wants full absolute URLs and not the relative `/api/csp-report` we had
- The controller was only handling legacy shape from `report-uri`
- Locally, this would have never worked since `report-to` only logs over HTTPS

## 🧪 Testing

> [!NOTE]
> This will need to be tested on azure as well so do both steps locally and in dev. 

1. Rebuild `make down; make up`
2. Login as anyu user
3. Go to account settings and open an email verificaiton dialog
4. Confirm a CSP report is sent and logged


petertgiles:
Yo, I think I figured out the problem.
[use disguised host](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16846/commits/38bdfbb779293798cfe5f158b138ba191e27893b)

image

image

esizer:
> Yo, I think I figured out the problem. [use disguised host](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16846/commits/38bdfbb779293798cfe5f158b138ba191e27893b)
> 

oh my! That actually makes sense. I forgot we had that :face_exhaling:

petertgiles:
I'll do one more check on this then move it to review.

api/app/Http/Controllers/CspReportController.php
```diff
@@ -3,38 +3,68 @@
...
         }
 
         $data = $request->json()->all();
-        $report = $data['csp-report'] ?? null;
 
-        // Do not log anything if we don't have a report
+        // Reporting API format (report-to): a list of report objects
+        if (array_is_list($data) && $data !== []) {
+            /** @var array<int, array<string, mixed>> $data */
+            collect($data)
```
petertgiles, line 23: This data isn't getting logged anywhere yet, right?

api/app/Http/Controllers/CspReportController.php
```diff
@@ -3,38 +3,68 @@
 namespace App\Http\Controllers;
 
 use Illuminate\Http\Request;
+use Illuminate\Http\Response;
 use Illuminate\Support\Facades\Log;
 
 class CspReportController extends Controller
 {
-    public function report(Request $request)
+    public function report(Request $request): Response
```
petertgiles, line 11: The overall structure of this method is starting to get weird.

```
if (looks like an array) {
    // process like an array
    // return
}

if (doesn't look like csp-report) {
    // fail
}

process like csp report
```

I suggest reworking it as:
```
if (looks like an array) {
    // process like an array
    // return
}

if (looks like a CSP report) {
    // process like csp report
    // return 

}

// fail
```

petertgiles (changes requested):
One bug and one stylistic recommendation left.

api/app/Http/Controllers/CspReportController.php
esizer: It should be, it just happends on line 25 when we call `logViolation`

petertgiles: Oh, my bad.  I missed it.  Thanks.

api/app/Http/Controllers/CspReportController.php
esizer: oh yeah, that makes sense. Updated in [reorder controller to put fail statement last](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16846/commits/6e4246363f2ad0360aa2421bda90338c39e7d953)

petertgiles: Thanks, I like this better.

petertgiles:
Thanks, both forms of the CSP report are logging well for me.

petertgiles (approved):

Issue #16269, opened 2026-03-26
🦔 [RFC] Framework Mode Compatible Client Pattern


esizer:
### 🦔 Spike

> [!IMPORTANT]
> We should at least consider this before any more routes are migrated as it is highly disruptive. 

## Problem

When trying to implement the new `clientLoader` and `clientActions` we get with framework mode, we ran into some friction. From reading, there seems to be a few issues:

 - Document cache does not play nice with framework mode
 - URQL heavily relies on hooks and the react lifecycle for its cache in react. We are trying to eject the data handling outside of the react life cycle
 - We have had issues with graphcache which is supposed to make it play nicer with react router

It does work, but is not ideal. This is a proposal to try and smooth it out but developing a new pattern. 

## Scope

1. Top-level route middleware creates one request client per navigation and stores it in route context.
2. `clientLoader` uses `queryClient.ensureQueryData(...)` for prefetch.
3. `clientAction` executes mutation and explicitly invalidates query keys.
4. **Exchange parity is built into the request client pipeline** (not optional):
   - `protectedEndpointExchange`
   - `mapExchange`
   - `specialErrorExchange`

## New request pipeline (exchanges)

This would be the new request pipeline that replaces the current urql exchanges. You might note that this does not include a cache exchange, that is becuase the purposed new library (tanstack/query) is the cache exchange.

This would allow our cache to be more determinstic where a query is assigned a key and we can manually invalidate on object updates by key.

```ts
// packages/client/src/utils/routeRequestPipeline.ts
interface PipelineContext {
  intl: IntlShape;
  auth: AuthenticationState;
  endpoint: string;
  operationName: string;
}

// NOTE: This seems common and we could find something to do this? 🤔
// Or maybe this is included in the reuqest? Im not sure..
export function extractOperationName(document: string): string {
}

// protectedEndpointExchange parity
export function guardProtectedEndpoint(ctx: PipelineContext) {
  if (!canAccessProtectedEndpoint(ctx.endpoint)) {
    throw new Error(`Protected endpoint denied for ${ctx.operationName}`);
  }
}

// mapExchange parity
export function mapGraphqlError(error: unknown, ctx: PipelineContext) {
  const logger = getLogger();

  logger.error(
    JSON.stringify({
      message: "route request onError",
      operation: ctx.operationName,
      error,
    }),
  );

  if (containsUserDeletedError(error)) {
    const reason: LogoutReason = "user-deleted";
[...]

PR #16851, opened 2026-05-21
[Tooling] Run playwright tests in Azure

### LINKED ISSUE #14982: 🦔 How can we run Playwright against Azure?
### 🦔 Spike

We want to be able to run Playwright tests against an Azure vertical.  Our first attempt to make this happens hasn't turned out to be maintainable.  Some ideas:
- Playwright can run from a docker container: https://playwright.dev/docs/docker
  - Can we get Docker in the admin virtual desktop?
- Can we get it running in the app service itself?
- Can we get it running in a sidecar? https://learn.microsoft.com/en-us/azure/app-service/overview-sidecar

If we have it running with the webserver or with Laravel it will give us more options for setting up tests.

### ❓ Questions to answer

How can we run Playwright against Azure?

### 🕙 Timebox

3 days

brindasasi:
🤖 Resolves #14982 
## 👋 Introduction

This PR enables automated Playwright regression tests to execute against the UAT environment via Azure Pipelines. It introduces a test token mechanism that allows the test runner to obtain authenticated sessions without interacting with GCKey, and refactors the test suite to create and tear down test users dynamically.

## 🕵️ Details

**Test Token Mechanism**

**Endpoint selection**

The test token branch is implemented within the existing /refresh endpoint rather than as a new route. 
This decision was made on the following grounds:

**No change to the application's external surface**
  The /refresh route already exists, is already registered in WAF and firewall rules, and is already reachable from the public internet. Introducing a new route would require infrastructure rule changes and would expose a URL that is both discoverable by automated scanners and visually indicative of a testing facility.

**Activation is not URL-addressable**
The test branch is triggered exclusively by the presence of the X-Testing-Secret request header. A caller without the correct header receives the standard OAuth refresh response, unchanged. 

The header was chosen over a query parameter to prevent the secret from appearing in server access logs, proxy logs, or request traces.

**Guard conditions**

TESTING_TOKEN_ENABLED=true 
AND 
APP_ENV_LABEL is not production  
AND 
X-Testing-Secret header matches TESTING_ENDPOINT_SECRET 

**Cryptographic isolation**

Test tokens are signed with TESTING_JWT_SECRET, a 32-byte symmetric key stored in Azure Key Vault and provisioned on UAT only. 

TestBearerTokenService, the service responsible for validating test tokens, is not instantiated on any environment where TESTING_TOKEN_ENABLED is absent or APP_ENV_LABEL is production.

 A test token presented to any such environment would be rejected at the service layer for an unrecognised issuer, and rejected at the cryptographic layer for an unverifiable signature.

No path exists by which a test token issued on UAT could be accepted on production.

**TestBearerTokenService**

TestBearerTokenService wraps the real bearer token service. 

Tokens whose iss claim is not gc-digital-talent-testing are passed to the real validator without modification. GCKey-issued tokens on UAT are unaffected.

**Test User Lifecycle**

Tests no longer depend on a catalogue of pre-seeded accounts in the UAT database.

 Each test creates a dedicated user via `createUserWithRoles` in beforeEach and deletes it in afterEach.

 This is subject to change as per our need.

**Current Limitation and Path Forward — VPN-Protected UAT**

Current state
The pipeline runs on a Microsoft-hosted ubuntu-24.04 agent (pool: vmImage: ubuntu-24.04). Microsoft-hosted agents have no VPN access. This works today because UAT is publicly reachable, but it will break the moment UAT is moved behind the VPN — the agent will be unable to connect to BASE_URL and every test will fail at the token request step.

**When UAT goes under VPN**
We need **Docker on the self-hosted agent**

A self-hosted VM does not come with Playwright's browser system dependencies pre-installed (libatk, libgbm, libnss3, and several others). 
Installing and maintaining these directly on the VM is fragile and version-sensitive - they must match the exact Playwright release in use.

```
- script: |
    docker run --rm --network host \
      -v $(System.DefaultWorkingDirectory):/work \
      -w /work/apps/playwright \
      -e CI=true \
      -e BASE_URL=$(baseUrl) \
      -e TESTING_ENDPOINT_SECRET=$(TESTING_ENDPOINT_SECRET) \
      -e PLAYWRIGHT_PLATFORM_ADMIN_SUB=$(PLAYWRIGHT_PLATFORM_ADMIN_SUB) \
      -e PLAYWRIGHT_COMMUNITY_ADMIN_SUB=$(PLAYWRIGHT_COMMUNITY_ADMIN_SUB) \
      -e PLAYWRIGHT_APPLICANT_SUB=$(PLAYWRIGHT_APPLICANT_SUB) \
      mcr.microsoft.com/playwright:v$(PLAYWRIGHT_VERSION)-jammy \
      npx playwright test $(playwrightArgs) --retries=2
  displayName: "Run Playwright tests (UAT)"

```

The agent VM only needs Docker installed. 

**IMTD Approval**

I've created jira ticket for pipeline review and docker installation on self hosted VM.

## 🧪 Testing

Test token service locally :

1. Add to api/.env:
**TESTING_TOKEN_ENABLED=true
APP_ENV_VERTICAL=local
TESTING_JWT_SECRET=<generate below>
TESTING_ENDPOINT_SECRET=test123**

2. Generate a valid JWT secret (run inside the API container):
**docker compose exec api php -r "echo base64_encode(random_bytes(32));"**
Copy the output (44 chars, ends with =) as TESTING_JWT_SECRET.

3. Clear the config cache:
**docker compose exec api php artisan config:clear**

4. Hit the endpoint:
**curl "http://localhost:8000/refresh?sub=admin@test.com" \
  -H "X-Testing-Secret: test123"**

Observe you get the below in response
{
  "access_token": "<token>",
  "refresh_token": "<token>",
  "id_token": "<token>"
}

**Verify in UAT** 
This needs the branch code be deployed in UAT 
AND
Azure Dev pipeline be triggered against UAT 

Testing results will be show in a separate tab in Azure. 

## 🚚 Deployment

All these must be set in UAT env vars / IaC

TESTING_TOKEN_ENABLED=true
APP_ENV_VERTICAL=uat
TESTING_JWT_SECRET=<base64 32-byte key>
TESTING_ENDPOINT_SECRET=<64-char hex>

All these should be set in the Azure dev pipeline ( These are already set ..change only if needed a different one ) 

BASE_URL= <uat host name>
TESTING_ENDPOINT_SECRET=<same value as in env var>
PLAYWRIGHT_PLATFORM_ADMIN_SUB=<sub of a real platform_admin user in UAT DB>


.github/workflows/lint.yml
```diff
@@ -35,6 +35,18 @@ jobs:
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 3
 
     steps:
+      - name: Reclaim space on runner
```
brindasasi, line 38: The lint job was missing the disk-reclaim step that vitest.yml already has

api/app/Http/Controllers/AuthController.php
```diff
@@ -266,6 +266,11 @@ public function authCallback(Request $request)
 
     public function refresh(Request $request)
     {
+        // Test token branch — only active when TESTING_TOKEN_ENABLED=true and APP_ENV_LABEL != production
+        if (config('testing.token_enabled') && config('app.env_label') !== 'production' && $request->hasHeader('X-Testing-Secret')) {
+            return app(TestTokenController::class)->issue($request);
```
brindasasi, line 273: X-Testing-Secret is the password the pipeline has to present to even get a fake token issued

api/app/Services/TestBearerTokenService.php
```diff
@@ -0,0 +1,99 @@
...
+ * symmetric key — no GCKey JWKS fetch or introspection call is made.
+ *
+ * All other tokens (real GCKey JWTs) are passed through to the real service
+ * unchanged, so normal UAT logins continue to work while this is active.
+ *
+ * Only registered when TESTING_TOKEN_ENABLED=true. Never used in production.
+ */
+class TestBearerTokenService implements BearerTokenService
+{
+    public const TEST_ISSUER = 'gc-digital-talent-testing';
```
brindasasi, line 28: `gc-digital-talent-testing` is the label stamped on every fake token so the system can tell it apart from a real GCKey token and route it to the right validator.

petertgiles:
Very interesting!  I haven't tried running it yet but just finished a read-through.  I have some questions and concerns about the approach.  I know you've already arrived at a working solution but given that this is a spike issue I think it's appropriate to have more discussion on the PR.

Running it in a pipeline is an interesting idea.   There's nothing special about the pipeline runner?  It's just a docker host on the public internet.  We could do the same thing in a manually triggered Github Actions runner, right?

I know you've added a lot of extra security checks but having a codepath in the app that bypasses authentication checks makes me nervous.  For what it's worth, I still think we should remove this as well:
https://github.com/GCTC-NTGC/gc-digital-talent/blob/4ba3b90e761629d0de0466baf9a507c42d6f43c8/api/app/Providers/AuthServiceProvider.php#L112
Two possible approaches come to mind that could mitigate this.

First option - we're less than a month until the CanadaLogin launch and SIC will be gone anyway.  They've already told us they can turn off MFA in the staging service.  Maybe you could check if CanadaLogin will cooperate better with Playwright?  Given that we're in a "partner launch" posture with them they might be more receptive to requested changes right now.

Second option - I know we're already talking with the PSC about needing API integrations with external services.  I could see the possibility of taking the auth work that you've done here and formalize it in an API token system that we could use for other things in addition to testing.  I'm thinking of how Github has personal access tokens, for example.

I'd be interested in reactions from @tristan-orourke and @esizer , too.  I know we still need to discuss https://github.com/GCTC-NTGC/gc-digital-talent/issues/16269 , too.

esizer:
> Second option - I know we're already talking with the PSC about needing API integrations with external services. I could see the possibility of taking the auth work that you've done here and formalize it in an API token system that we could use for other things in addition to testing. I'm thinking of how Github has personal access tokens, for example.

Yeah, I have only really skimmed the code so far but, the idea of issuing personal access tokens like this seems like it could help with both playwright and the expected requirements coming from PSC :thinking:

api/.env.example
```diff
@@ -1,5 +1,8 @@
 APP_NAME=api
 APP_ENV=local
+# Tier label independent of APP_ENV — both UAT and production use APP_ENV=production.
+# Used as a second guard on test-only endpoints. Values: local | uat | production
+APP_ENV_LABEL=local
```
tristan-orourke: This defaults to `production` in  the config/app.php file. Maybe it should be the same here?

tristan-orourke: Maybe this would make more sense with a name like APP_ENV_SECURITY_LEVEL?

tristan-orourke:
@brindasasi you've done a great job explaining and justifying this approach in the PR details! It seems well thought-out, and it made it very easy to read through the PR.

@petertgiles I appreciate your concerns about bypassing authentication checks. I think you're right about it being past time to remove that `oauth.default_user` option. But thinking about it now, I think this PR is pointing to the right approach. The service that Canada Login provides is to prove that someone is who they claim to be. When Playwright is running tests, that's not really even true! It makes sense to bypass Canada Login for most tests. (It may make sense to have some integration tests, if possible.)

The other thing I like here is that we're technically not bypassing all authentication checks, we're just using a _different_ check (the shared secret token).

I do think @petertgiles makes a great point that we should think about extending this to work for non-testing purposes, and (though its a bit scary) on production itself. Basically, how do we handle integrating with trusted services? It doesn't make sense for those services to go through Canada Login, since they're not _people_.

petertgiles:
The google AI seems to think that IBM Verify (which CL is based on) supports service accounts.  I sent a message to that team to see.  Maybe we won't have to build this ourselves.  :crossed_fingers: 

image

petertgiles:
Reviewed in refinement:
- Peter: would like to see this running in Github instead - seems like there are firewall problems currently
- Peter: would like to see this generalized to an API key system we could use in production
- Tristan: this is useful now, we should proceed as-is

api/app/Http/Controllers/AuthController.php
```diff
@@ -266,6 +266,11 @@ public function authCallback(Request $request)
 
     public function refresh(Request $request)
     {
+        // Test token branch — only active when TESTING_TOKEN_ENABLED=true and APP_ENV_LABEL != production
+        if (config('testing.token_enabled') && config('app.env_label') !== 'production' && $request->hasHeader('X-Testing-Secret')) {
```
petertgiles: This branch seems strange here.  I think it would make more sense to have the branching logic in the routes file - web.php.  In the routes for /refresh you would call one controller or the other.

Why do we need to intercept refresh calls at all?  The test tokes last for 2 hours, no?

api/config/app.php
```diff
@@ -106,6 +106,22 @@
...
+    | A human-readable tier label independent of APP_ENV. Both UAT and production
+    | run APP_ENV=production, so this var is used where behaviour must differ
+    | between them (e.g. test-token endpoint guard).
+    |
+    | Values: local | uat | production
+    | Default: production — safe fallback so untagged environments are locked down.
+    |
+    */
+
+    'env_label' => env('APP_ENV_LABEL', 'production'),
```
petertgiles: I find "env_label" a bit too generic to understand what it is for.  Maybe "vertical" is more descriptive?  That's what is used in azure-pipelines-dev.yml

infrastructure/azure-pipelines-dev.yml
```diff
@@ -1,26 +1,35 @@
...
   serviceConnectionName: "Talent Cloud Service Connection"
   npm_config_user_config: "$(System.DefaultWorkingDirectory)/.npmrc"
-  dirResults: "$(System.DefaultWorkingDirectory)/apps/playwright/test-results"
-  resourceGroupName: "rg_tbs_talent_2_PaaS"
-  vertical: "dev"
-  storageAccountName: "sttalclddevcacstdl"
-  storageContainer: "playwright-results"
+  ${{ if eq(parameters.vertical, 'uat') }}:
+    baseUrl: "https://uat-talentcloud.tbs-sct.gc.ca"
+    playwrightArgs: "--project=uat-* --project=chromium --grep=@regression"
```
petertgiles: I'm not sure "regression" is a useful tag for this since it has so much meaning in testing already.  Maybe "runincloud" or "uatvertical" or something like that?

api/config/testing.php
```diff
@@ -0,0 +1,47 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | Base64-encoded 32-byte key used to sign and verify test tokens.
+    | No default — app will throw if TESTING_TOKEN_ENABLED=true and this
+    | is not set, preventing silent use of a guessable fallback.
+    |
+    | Generate: php -r "echo base64_encode(random_bytes(32));"
+    |
+    */
+    'jwt_secret' => env('TESTING_JWT_SECRET'),
```
petertgiles, line 32: Do we need a new secret?  Couldn't we use the Laravel app key as the encryption key?

api/routes/web.php
```diff
@@ -28,3 +29,10 @@
     Route::get('/refresh', [AuthController::class, 'refresh']);
     Route::get('/sector-identifier', [AuthController::class, 'sectorIdentifier']);
 });
+
+// Test-only token endpoint — only registered when TESTING_TOKEN_ENABLED=true.
+// Returns a short-lived JWT for a user matched by ?role= or ?sub= without
+// going through GCKey. Never available in production.
+if (config('testing.token_enabled') && config('app.env_label') !== 'production') {
+    Route::get('/testing/token', [TestTokenController::class, 'issue']);
```
petertgiles: Do you use this route?  I thought you were overloading /refresh because you didn't want to configure any new routes?

petertgiles (commented):
I did another read-through and have a few more questions.

infrastructure/azure-pipelines-dev.yml
brindasasi: sure .. not particular about the naming.
It is part of regression tests in UAT.

api/routes/web.php
brindasasi: oh yes ,, this is left over ..tried and abandoned

api/app/Http/Controllers/AuthController.php
brindasasi: > Why not web.php? 


The branch is in AuthController rather than web.php because Laravel routes cannot natively branch to different controllers based on a request header - you'd need middleware, which adds more moving parts for the same result. The current placement is the least invasive option.

> Why do we need to intercept refresh calls at all?

Also the test token isn't a token refresh. The endpoint was chosen purely because it already exists and is already reachable. The X-Testing-Secret header is what activates the test branch; without it the endpoint behaves exactly as before.

>  The test tokes last for 2 hours, no?

Yes. Playwright doesn't refresh them. The /refresh route is only called once per test user at the start of each test.

api/config/testing.php
brindasasi: Technically yes.. but I chose separate because a dedicated TESTING_JWT_SECRET can be rotated independently without touching anything else.
 It also keeps the test concern isolated so if this mechanism is ever removed, there's nothing to unpick from the app's core secrets.

api/.env.example
brindasasi: The .env.example intentionally shows local because developers copying it to .env are running locally , they'd want local, not production. 

The config default of production is the safe fallback for when the env var is absent entirely in the azure vertical

brindasasi: SECURITY_LEVEL sounds like some permission level .. but this merely environment differentiator 
could be APP_DEPLOYMENT_ENV something like that ?

api/routes/web.php
brindasasi: fixed in [9b1ea36](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16851/commits/9b1ea36301bac782f0ba825401b69d9f27ec87c8)

infrastructure/azure-pipelines-dev.yml
brindasasi: fixed in https://github.com/GCTC-NTGC/gc-digital-talent/pull/16851/commits/5c6cafff74ee0cf7915f7016886ccff1a9940dcf

api/config/app.php
brindasasi: fixed in https://github.com/GCTC-NTGC/gc-digital-talent/pull/16851/commits/be2bbb8959f5757bf7243e3bce0d567403389744

turbo.json
```diff
@@ -157,7 +157,14 @@
       "cache": false
     },
     "e2e:playwright": {
```
petertgiles, line 159: It seems like running this command locally no longer works.  Could we make sure we haven't added any friction for local testing?

Image

apps/playwright/tests/auth/setup-helper.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export async function authenticateAs(
+  name: string,
+  sub: string,
+  page: Page,
+  request: APIRequestContext,
+) {
+  const secret = process.env.TESTING_ENDPOINT_SECRET;
+  const response = await request.get(`/refresh?sub=${sub}`, {
```
petertgiles: Copilot flagged that this isn't urlencoded and can break on reserved characters.

api/app/Http/Controllers/TestTokenController.php
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    private function resolveUser(Request $request): ?User
+    {
+        if ($sub = $request->query('sub')) {
+            return User::where('sub', $sub)->first();
+        }
+
+        if ($role = $request->query('role')) {
+            return User::whereHas('roles', fn ($q) => $q->where('name', $role))->first();
```
petertgiles: This role-based lookup makes me nervous.  Besides being less secure if we have a flaw, it also makes the tests less deterministic and possibly more flaky.  Can we test by using sub only?

petertgiles (commented):
Some more comments.

api/app/Http/Controllers/AuthController.php
```diff
@@ -266,6 +266,11 @@ public function authCallback(Request $request)
 
     public function refresh(Request $request)
     {
+        // Test token branch — only active when TESTING_TOKEN_ENABLED=true and APP_ENV_VERTICAL != production
+        if (config('testing.token_enabled') && config('app.vertical') !== 'production' && $request->hasHeader('X-Testing-Secret')) {
+            return app(TestTokenController::class)->issue($request);
+        }
```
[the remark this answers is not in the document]
petertgiles: I'm not sure this is actually a problem.

.github/workflows/lint.yml
```diff
@@ -35,6 +35,18 @@ jobs:
...
+        run: |
+          rm -rf /usr/local/android
+          rm -rf /usr/share/dotnet
+          rm -rf /usr/local/share/boost
+          rm -rf /opt/ghc
+          sudo rm -rf /usr/local/share/powershell
+          sudo rm -rf /usr/share/swift
+          sudo rm -rf /usr/local/.ghcup
+          sudo rm -rf /usr/lib/jvm
+          df -h
```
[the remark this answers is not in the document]
petertgiles: I don't think this has been a problem.

api/.env.example
tristan-orourke: I liked @petertgiles's suggestion of naming it after verticals

api/app/Http/Controllers/TestTokenController.php
brindasasi: I don't think I've used it anymore .. another tried & abandoned method.. will clean up

turbo.json
petertgiles: I noticed that the three new testing variables aren't in the .env.example or filled by the setup script.  Are they necessary for local testing?  Even with them manually set it seems like the @uat tests ~are~ aren't passing for me.

image

brindasasi: No, not for local. They're only needed for running against UAT

petertgiles: Should the uat tests be passing locally though?  If not, shouldn't they be skipped when running locally?

brindasasi: I meant the additional vars doesn't need to exist in local .. they meant only for UAT. 
tests may need tweaks to run in local  as I used dynamic users in UAT instead of seeded ones in local ..let me check that closer

petertgiles: Nice!  Seems to be working now.

api/app/Http/Controllers/TestTokenController.php
```diff
@@ -0,0 +1,62 @@
...
+            'refresh_token' => $this->mintToken($user->sub),
+            'id_token' => $this->mintToken($user->sub),
+        ]);
+    }
+
+    private function mintToken(string $sub): string
+    {
+        $config = Configuration::forSymmetricSigner(
+            new Signer\Blake2b(),
+            InMemory::base64Encoded(config('testing.jwt_secret')),
```
petertgiles, line 51: I'm getting an exception on L51 here about invalid characters.

Image

My secret is `QI6OsOp46+EvdaKgBl/hmC3mDEjbaj8qnl7dZklkV4=` as generated with the instructions in the config file.  Am I doing something wrong?

petertgiles (commented):
I'm doing some testing with Postman but having trouble with the mint function.

api/app/Http/Controllers/TestTokenController.php
brindasasi: are you sure you are not missing any char? 
your string length is 43 chars and it needs to be 44
could you try this again ? 
php -r "echo base64_encode(random_bytes(32));"

brindasasi: updated testing steps above in the PR for testing token minting locally

petertgiles: You're right!  I must have dropped a character somewhere.  :facepalm:   Thanks for catching that.

petertgiles (approved):

PR #16864, opened 2026-05-22
[Feat] Add centralized permission checker

### LINKED ISSUE #16278: ✨ Centralized permission checker
## Details

We have some disparity when checking permissions and have made some attempts at having a single source of truth between the server and client. Thankfully, that single source of truth does exist, it's simply a matter of getting it to the client so it can be used there.

We can do this and leverage our graphql API and codegen to hopefully make it type safe as well.

## Proposed Implementation

I propose we use the `rolepermission.php` config to generate some graphql typoes and a query. We can then query them on app boot and store a stabel reference in context to be used in a permission checker.

### GraphQL Types

Similar to how we register enums, we can use the `rolepermnission.php` config file to register the types and the query so we end up with:

```graphql
enum Role {
  Applicant
  PlatformAdmin
  # ...
}

enum Permission {
  UpdateAnyApplicationPlacement
  UpdateTeamApplicationPlacement
  # ...
}

type RolePermission {
  role: Role!
  permissions: [Permission!]!
}

type Query {
  rolePermissions: [RolePermission!]!
}
```

### Client

On the client, we can cimply request these in the context provider and/or middleware. Then, we have it available for a permission checker.

```tsx
const UserMiddleware_Query = grpahql(/** GraphQL */`
  query UserMiddleware {
     # ... existing fileds
     rolePermissions {
        role
        permissions
     }
  }
`);

interface PermissionRequirement {
  permission: Permission.
  teamIds?: Maybe<string>;
}

// Helper fucntion that is similar to hasRequiredRoles
// This is for use in a hook and a function checker
const checkPermissions = (
   permissions: PermissionRequirement | PermissionRequirement[],
   userRoles: Maybe<Maybe<RoleAssignment>[]>;
   rolePermissionMap: RolePermissionMap[],
): boolean => {}

// Simplifeid utulity for checking a permissions in middleware
const requirePermissions = (
  ctx: AppContext,
  permission: PermissionRequirement | PermissionRequirement[],
) => {
  const roles = // Get roles from context
  const map = // Get map from context
  
 if(!checkPermission(permissions, roles, map, teamIds)) {
   // Throw unauth
 }
}

// Simplied hook for checks in components
const useHasPermissions = (
  permissions: PermissionRequirement | PermissionRequirement[],
): boolean => {
  const { roleAssignments, rolePermissionMap, isLoaded } = useAuthorization();
  return isLoaded && checkPermissions(
    permissions,
    roleAssignments,
    rolePermissionMap,
    teamIds
   );
};
```

## Usage

Then, we can use these to check the permissions which could be type safe by now 🤞 

```tsx
// In middleware or loader
requirePermissions(context, request, [
    { permission: Permission.UpdateAnyApplicationPlacement },
    { permission: Permission.UpdateTeamApplicationPlacement, teamId?: poolCandidate.pool.teamId }
]);

// In component
const canPlace = useHasPermissions(
  [
    { permission: Permission.UpdateAnyApplicationPlacement },
    { permission: Permission.UpdateTeamApplicationPlacement, teamId?: poolCandidate.pool.teamId }
  ]
);
```

### ✅ Acceptance Criteria

- [ ] GraphQL types and query for permission config
- [ ] Permission map in context (provider and middleware)
- [ ] Utilites created for asserting permissions
- [ ] Maybe one example?

esizer:
🤖 Resolves #16278 

## 👋 Introduction

Adds our roles and permissions to the graphql schema (enums, query) for use on the frontend.

## 🕵️ Details

We had a problem where we were trying to keep permission checks in sync between the client and server but had no single source of truth. This uses the existing config for roles a permissions to construct an enum for both `RoleName` and `Permission`. We can then use these on the frontend since they get generated as typescript types by codegen.

It also adds some helpers along with adding permissions a role has to the `Role` model so the client is aware of the permissions the logged in user has.

#### Generic checker

We have a generic central checker that is used in other higher level helpers. This can be used anywhere but is a little more involved.

```tsx
import { checkPermissions } from "@gc-digital-talent/auth";

const hasPermission = checkPermissions([
    { permission: Permission.UpdateAnyApplicationPlacement },
    { permission: Permission.UpdateTeamApplicationPlacement, teamId: "UUID" },
  ],
  unpackMaybes(dta?.user?.authInfo?.roleAssignments),
);
```

#### In Middleware

We already had a `requireUser` helper for enforcing a user is logged in, ti has been updated to accept both roles and permissions in an ooptions object.

```tsx
import { RoleName, Permission } from "@gc-digital-talent/graphql";

import { requireUser } from "~/routing/auth";

export const clientMiddleware: Route.ClientMiddlewareFunction[] = [
  async ({ context, request }, next) => {
    requireUser(context, request, { 
      roles: [{ name: RoleName.PlatformAdmin }],
      permissions: [{ permission: Permission.UpdateTeamProcess, teamId: "UUID" }], 
    });
    return await next();
  },
];

```

#### As a hook

We load the permission map into the auth context along with the current users role assignments. So, we now also have a handy hook that uses those to assert permissions.

```tsx
import { Permission } from "@gc-digital-talent/graphql";

const hasPermission  = useHasPermissions([
  { permission: Permission.UpdateAnyApplicationPlacement },
  { permission: Permission.UpdateTeamApplicationPlacement, teamId: "UUID" },
]);
```

## 🧪 Testing

1. Confirm schema changes match expectations
2. Rebuild `make refresh-api; pnpm dev:fresh`
3. Confirm query returns expected permission map
4. Login as various roles and confirm the implemented fix on the placement dialog for pool candidates is properly returning the right permissions


packages/auth/src/utils/checkPermissions.ts
```diff
@@ -0,0 +1,44 @@
...
+): boolean => {
+  const assignments = (roleAssignments ?? []).filter(
+    (a): a is RoleAssignment => !!a && !!a.role,
+  );
+
+  const reqs = Array.isArray(requirements) ? requirements : [requirements];
+
+  for (const req of reqs) {
+    for (const assignment of assignments) {
+      const permissions = assignment.role?.permissions ?? [];
```
petertgiles, line 30: I think the logic in this block is correct but I find it a bit hard to follow, especially the early continue and three different returns.  I suggest trying to restructure it more like:

```ts
for (...) {
    const permissions = ...
    const isTeamBased = ...

    if (isTeamBased) {
        if (/* business logic */) return true;
    }
    // non-team check
    if (/* business logic*/) return true;
}

File: apps/web/src/routing/auth.ts [hunk 2]
```diff
@@ -1,25 +1,30 @@
...
+ * Validates the user session, redirecting to login if missing.
+ * Optionally asserts role and/or permission requirements, throwing
+ * UnauthorizedError if neither set is satisfied.
  */
 export function requireUser<T extends AppContext>(
   context: T,
   request: Request,
-  requirements?: RoleRequirement | RoleRequirement[],
-  strict = false,
+  options?: RequireUserOptions,
```
petertgiles, line 27: I really like this change!  It makes for much more expressive calls.

apps/web/src/pages/Root/userMiddleware.ts
petertgiles, line 1: Probably not relevant to this PR but bringing it up since I'm looking at it.  Especially since these responses just got quite a bit bigger.

It looks like the app fires two identical requests (with different operation names) on boot?

Image

Image

petertgiles: If you agree it's weird, I'll post an issue.

apps/web/src/pages/Pools/EditPoolPage/EditPoolPage.tsx
petertgiles, line 1: I'm getting a lazy load error on teams on the view pool page but I can't tell if it's related to this PR.  If you want, I can make a separate issue for it.

Image

petertgiles (changes requested):
This turned out really great!  I've flagged a few things but I'm not sure any of it is actually relevant to this PR.  😆

apps/web/src/pages/Root/userMiddleware.ts
esizer: oh yeah, this is annoying but expected. We have this query running for the `ClientProvider` and `userMiddleware` so both clients aere aware.

It was not supposed to be long lived but we ran into some trouble with the middleware adoption so we still have both loading strategies in place :cry:

apps/web/src/pages/Pools/EditPoolPage/EditPoolPage.tsx
esizer: I think that is a pre-existing issue. It looks like it is coming from:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/449c48f97ffa01a898da92513407b2168cf27302/apps/web/src/pages/Pools/ViewPoolPage/ViewPoolPage.tsx#L506-L517

packages/auth/src/utils/checkPermissions.ts
esizer: okay, it should read better after [update permission checker logic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16864/commits/24dd4b5bd8e0f1dcab43b17879cadb9dfbbf2551)

petertgiles: Love it!

apps/web/src/pages/Pools/EditPoolPage/EditPoolPage.tsx
petertgiles: Thanks for checking.  I'll post a bug issue.

petertgiles (approved):

apps/web/src/pages/Pools/EditPoolPage/EditPoolPage.tsx
petertgiles: https://github.com/GCTC-NTGC/gc-digital-talent/issues/16882

PR #16917, opened 2026-05-28
[Bug] Referral status filter

### LINKED ISSUE #16390: 🐛 UAT | Disqualified and Removed candidate is present when filter with Referral status as Not referred

- When filtering the Pool candidate table or Talent request's candidate table with 'Referral status' as 'Not referred', the result shows the 'Disqualified' and 'Removed' candidates as well in the pool candidate table. 
- The requirement is those candidates whose Referral status is NULL or blank should not be present in the table when filtering with 'Not referred' 

**Expected Outcome** : 
1. Talent request should always return candidates that are:
Qualified
Available for referral
Active (pool expiry date)
Interested in opportunities
None of these should be changed when responding to a request

2. Candidate table filter result should contain the qualified candidate when filter with any of the referral status

### 🕵️ Details

Env: UAT

### 📋 Steps to Reproduce

Prerequisite: Find a process having multiple candidate applied to it with different application statuses. Create a talent request for the same pool

1. Access to GC digital portal using Recruiter role
2. Open the pool candidate table for above mentioned pool
3. Filter the table with 'Referral status' as 'Not referred'
4. Observe the result - Disqualified and removed candidates who is having referral status as BLANK, is also present in the candidate table

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] The candidate whose referral status is NULL or blank should not be present when filter with referral status as Not referred or Available for referral
- [ ] The referral status filter are exclusively for qualified candidate status only, hence filter result should contain the qualified candidate only.

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

yonikid15:
🤖 Resolves #16390 

## 👋 Introduction

- Fixes the referral status filtering business logic
- Adds comment to confirm and clarify logic 

## 🕵️ Details

- The candidate whose referral status is **NULL** or blank should **NOT** be present when filtering with: 'Not referred' or 'Available for referral'
- The referral status filters are exclusively for **QUALIFIED** candidates only

## 🧪 Testing

1. Build app
2. Login as admin
3. Go to the candidates table
4. Ensure the Referral status filter works correctly:
  - Only QUALIFIED candidates are present if one or both filter by's are selected
  - Ensure both 'Not referred' and 'Available for referral' still work when set individually

## 📸 Screenshot

image

image


api/app/Builders/PoolCandidateBuilder.php
```diff
@@ -360,7 +371,7 @@ public function whereNotBeingReferred(): self
                     $query->whereNull('resume_referrals_at')
                         ->orWhere('resume_referrals_at', '>', $now);
                 })
-                ->orWhere('application_status', '!=', ApplicationStatus::QUALIFIED->name);
+                ->where('application_status', ApplicationStatus::QUALIFIED->name);
```
petertgiles: Now that these three clauses are all being AND'd together, the subquery introduced on L367 should no longer be necessary.

api/app/Builders/PoolCandidateBuilder.php
```diff
@@ -333,7 +339,12 @@ public function whereReferralStatusIn(?array $referralStatuses): self
             return $this->whereNotBeingReferred();
         }
 
-        // none selected or both selected - no filtering
+        // both selected
+        if ($hasReferring && $hasNotReferring) {
+            return $this->where('application_status', ApplicationStatus::QUALIFIED->name);
```
petertgiles: I think this is correct but it's maybe not obvious why this condition represents "where being referred or not being referred".  I wonder if this could be refactored as something like:

```
return $this->where(function($subquery) {
    $subquery
        ->whereBeingReferred()
        ->orWhere(function(Builder $subsubquery) {
            $subsubquery->whereNotBeingReferred();
        })
});
```

I don't know... is that better or not?  😆  I guess it would be more resilient in case the logic in either of those two scopes changes in the future.

petertgiles (changes requested):
Nice, this seems to work well.  :clap:  I've left two suggestions but you can decide if you agree with them.  I would like to see some tests for this though.

api/app/Builders/PoolCandidateBuilder.php
petertgiles: Copilot says there might be a cleaner way to write orWhere closures... food for thought.
image

yonikid15: Ya it's more explicit, let's go with that :+1: 
Changed in [Refector to use explicit code](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16917/commits/20cf5bd87aa35c39c1789196111f0edd939619ab).

api/app/Builders/PoolCandidateBuilder.php
yonikid15: Nice catch, should be fixed in [Remove unecessary sub query](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16917/commits/cff5e9bdf6f903cf853afcaf2dcd63937654543a).

petertgiles (approved):
Nice one.  👏 👏

PR #16919, opened 2026-05-28
[Fix] Increase file generation job timeout

### LINKED ISSUE #16891: 🐛 UAT | Error occurred when downloading any document contains user's profile

- All downloads that contain user profile to be downloaded is failing, such as: 
   - talent nominations download
   - community talent download [Excel and docx] 
   - candidate's profile docx download
- Found it when multiple users selected to download the profile in docx, it fails. A single user download docx is working fine

### 🕵️ Details

Env: UAT

### 📋 Steps to Reproduce

2. Navigate to community talent and select some users to download 
3. Download excel and docx for those users
4. Observe the error
5. Follow same steps for talent nominations and candidates table
6. Observe the error

### 📸 Screenshot

- ### Community talent download

https://github.com/user-attachments/assets/775755b5-533c-4249-94f8-5d86cbe0138f

- ### Candidates download docx

https://github.com/user-attachments/assets/0d509684-3552-4735-b9d1-e879006730ae

- ### Talent nomination download

https://github.com/user-attachments/assets/3f76beb1-6e37-4447-a44e-b8d6e2973a13

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] All downloads should work without any error

### 🔄 Regression

Yes

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #16891 

## 👋 Introduction

Increass timeout for the user generated file job.

## 🕵️ Details

The download full dataset links were leading to timeouts in UAT. This is likely due to the increased complexity of the files we are generating as we add new items to them along wiht the growing number of rows it needs to process.

This was difficult to identify because timeouts are only added to the `failed_jobs` and not logged anywhere else. So, this adds some logging for that so we can more eaisly identify jobs that are timing out. 

## 🧪 Testing

> [!IMPORTANT]
> Timeout kills the process so it testing locally, you need to restart the queue to get the failed notififcation. This is unecessary when deployed since supervisor does this for us. 

1. Seed a lot of rows for users 50,000+
2. Try to download a full dataset of the seeded rows
3. Confirm it takes a while but still succeeds
4. Set the timeout to a value lower than how lontg it takes to generate that file
5. Confirm you see the timeout log and get a failure notification


brindasasi:
@esizer should we close this one ?

esizer:
> @esizer should we close this one ?

Yup, this one is good to close :+1:

PR #16924, opened 2026-05-29
[Fix] Large file generation

### LINKED ISSUE #16891: 🐛 UAT | Error occurred when downloading any document contains user's profile

esizer:
🤖 Resolves #16891 

## 👋 Introduction

Fixes an issue where large excel files were fialing to generate due to timeouts and out-of-memory errors and an issue with deployment slots stealing queued jobs.

## 🕵️ Details

This does two things,

1. Replaces `PHPSpreadsheet` with `OpenSpout`
2. Increaes job timeout
3. Add deployment slot name env for queue name

### OpenSpout

The decision to move to this library was due to the memory issue. `PhpSpreadhseet` works by creating the entire file in memory and the writing it. This was an issue becuase holding the large amount of rows in memory along with the laravel overhead, we were quickly running out of memoory.

`OpenSpout` solves this by taking a different approach, it creates an empty file and then writes to it row by row. So, it only needs to keep the current row in memory while it writes. This should significantly reduce the amount of memory used by this job.

### Timeout

This is a quick fix to increase the timeout of `GeneratesUserFile` job. This is not the best approach since it requires us to bump the retry after time to some value larger than the largest timeout in this queue.

The better fix would be to create a separate queue for longer jobs so we can keep shorter jobs on a lower retry after, but that is less of a concern imo.

### Deployment slot ENV var

We added a new env var for `DEPLOYMENT_SLOT_NAME`. This is used as the queue name that is stored in the DB. The reason this was introduced was because, while deployment slots share a database (they do in our setup but not required), they do not share the same filesystem. So, what happened was:

 1. Job inserted into DB on the default queue
 2. Any free worker on either deployment slot raced to pickup the job
 3. A job stores a file on that deployment slots filesystem
 4. The success notificaiton gets stored in the shared DB
 5. The current deployment slot accepting 100% of traffic saw the notificaiton in the DB so presents it to the user but the file exists somwhere esle (thje other deployment slot)

So, a user sees the file but cant access it. This is also why the logs were missing, they all got logged on the other machine.

### More issues of less concern

This should solve timeouts and memory issues but, what has not been solved is how long this perocess takes. Right now it is reasonble, only taking 3-5mins with our current dataset. However, that number will scale close to linearly with the number of rows we are generating.

We might want to keep an eye on the time it takes to generate files as we scale and pivot to a more effecient method. Continually bumping the timeout on the same queue as other jobs is not sustainable long-term as row counts may rise.

## :exclamation: Important

I'm wondering if we need to be able to download such large datasets at all. Before we spend time working on improving this more, we might want to consider how useful or necessary downloading all candidates, users, etc is. The main isuse here is that its just a lot of data we are parsing and we might start getting into more complex methods. 

If our users really are say, trying to download 50,000 users in an excel, it might benefit us to ask wyh and see if we can implement a better soluttion to avoid this use case.

## :woman_shrugging:  Possible Longer-term solution

If users do need to dopwnload this much data, we might want to do one or both of two things intially:

 1. Split this out into its own queue separate from other jobs so we can significanyl increase the timeout
 7. Write the file in chunks in parallel to save some time

### Separate queue

Personally, if someone is downloading this much data, they should expect it will take some time so I dont think the time is the issue here. As long as it is not timing out and has steady memory consumption, we can let it run for longer time.

Because we increased the job timeout, retry_after must be set to a value higher than that timeout. This means all jobs on the same queue (emails, notifications, subscriptions) can't retry until that window is exhausted, even though they're short-lived.

If they are on a separate queue, they are at least isolated from eachother and will complete when necessary without affecting normal, short-lived jobs.

### Parallel Chunks

If time is an actual concern, we could take the initial chunk and spread the file write across 2-5 jobs and then once they are all done, we stitch the files together. That would tie up more workers and require more resources but, the job will get done faster.

### Last report

If this is a critical part of the application and we expect to keep generating larger files more frequentyl, we might need to consider breaking this out as a service so we arent affecting anything in the application. But, I would not attempt this until we exhaust the other two, easier options first.

## 🧪 Testing

> [!IMPORTANT]
> We need to test for regressions so it would be a good idea to get a copy of each file from main before testing. Making sure that the  seeded data is the same for a true 1:1 comparisson. 

8. Login as `admin@test.com`
9. Donwload all three versinos of excel
10. Confirm they all funcitoin
11. Confirm it matches the original file exactly

## :truck: Deployment instructions

Add the following ENV var which can be anything but must be unique on each deploment slot.

`DEPLOYMENT_SLOT_NAME=`


api/app/Generators/CsvGenerator.php
```diff
@@ -17,37 +14,31 @@ public function __construct(public string $fileName, protected ?string $dir)
...
-
+        // File is written to disk during generate() — nothing to do here.
     }
 
-    public function write()
+    protected function openCsv(): void
     {
-        if (! $this->spreadsheet) {
+        $path = $this->getPath();
+        $this->handle = fopen($path, 'w');
```
brindasasi: this should throw exception on failure .. otherwise silent data loss will happen

api/app/Jobs/GenerateUserFile.php
```diff
@@ -17,32 +17,24 @@ class GenerateUserFile implements ShouldQueue
 {
     use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
 
-    /**
-     * The number of times the job may be attempted.
-     *
-     * @var int
-     */
-    public $tries = 3;
```
brindasasi, line 25: removed retry? is this intentional?

api/app/Generators/UserExcelGenerator.php
```diff
@@ -224,88 +223,83 @@ public function __construct(public string $fileName, public ?string $dir, protec
...
+        // Community Interest sheet
+        $interestSheet = $this->writer->addNewSheetAndMakeItCurrent();
+        $interestSheet->setName(Lang::get('headings.community_interest', [], $this->lang));
+        $this->generateCommunityInterestSheet();
 
-        // Generate data for all sheets
-        $this->generateUsersSheet($usersSheet);
-        $this->generateCareerExperienceSheet($careerSheet);
-        $this->generateCommunityInterestSheet($interestSheet);
+        $this->writer->close();
```
brindasasi: can we move this entire block under try ,catch and move this close() into finally block?

api/app/Generators/CsvGenerator.php
esizer: Good call, I assumed those were throwing but apparently they do not :woman_facepalming: I changed to the OOP version that does throw in [update CSV generator so it throws errors](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16924/commits/723d7f8e92401bc2c80e0756bc6c3bd78208ccd5)

api/app/Jobs/GenerateUserFile.php
esizer: It was because of how the timeouts changed but, I don't think that removing it was necessary so I added it back in with [add tries back](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16924/commits/658129bf68ae95188df76b87cd01ea543acfaff2)

api/app/Generators/UserExcelGenerator.php
esizer: Good call, I did the same with all excel generators in [wrap generates in try/finally block](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16924/commits/b0d65b1581e040f7c97eb753a5e39910b1f26ddc)

brindasasi:
I wasn't able to download more than 200 records in dev 
Download dataset with 6k records downloaded only 200 ( data loss)
Set the page size as 500 and tried download but it didn't come back

Let me try with this latest update again

brindasasi:
Separate queue - I think we should try this approach .. this job deserves its own queue

brindasasi (changes requested):
Recent update didn't work very well in the dev. I got partial data downloaded one time and rest of the times didn't get any file at all.
Let's revisit the approach. 
As you have mentioned I'm curious about separate queue - I think we should try this  .. this job deserves its own queue.

brindasasi (approved):
Amazing work! ⭐️ 
Works really good!

PR #16932, opened 2026-06-01
[Tooling] Adds a talent request sync script

### LINKED ISSUE #16928: 🛠️ Sync script to sync Talent Requests with PoolCandidateSearchRequests after go-live
### 🛠️ Platform

Write a tinker script which syncs PoolCandidateSearchRequests with TalentRequests. We did this in a migration when TalentRequests are created, but want to be able to sync any remaining requests created between the migration and overall go-live of the rest of the TalentRequests code.

### 🕵️ Details

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

_No response_

esizer:
🤖 Resolves #16928 

## 👋 Introduction

Adds a script to sync new `PoolCandidateSearchRequest` entities to the new `TalentRequest` entity.

## 🕵️ Details

We made an initial migration when the new table was introduced but, there will be a short period where new requests could come in that need to be synced over. This command was built to be ran as many times as needed and can be ran just before removing the old `PoolCandidateSearchRequest` mode.

## 🧪 Testing

1. Make sure all migrations ran `make refresh-api`
2. Login to DB and confirm the number and records in both `pool_candidate_search_requests` and `talent_requests` tables
3. Run the command `make artisan CMD="app:sync-talent-requests"`
4. Confirm that after running the command, the tables match (outside of deprecated fields)
5. Submit a new request for talent from `/search`
6. Run the same command from step 3
7. Confirm only the new record is synced and not others

### DIFF
--- a/api/app/Console/Commands/SyncTalentRequests.php
+++ b/api/app/Console/Commands/SyncTalentRequests.php
@@ -1,0 +1,76 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Attributes\Description;
+use Illuminate\Console\Attributes\Signature;
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+#[Signature('app:sync-talent-requests')]
+#[Description('Syncs search request rows to new talent requests table.')]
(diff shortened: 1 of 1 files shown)


yonikid15 (approved):
LGTM 👍

PR #16934, opened 2026-06-01
[Fix] Enable static react components eslint rule

### LINKED ISSUE #16676: ♻️ Renable and address `react-hooks/static-component`
### ♻️ Refactor

We disabled this rule due to how wide spread this issue was to keep the update scope small. We should renable this rule and address it though.

### 🕵️ Details

This is due to our pattern of using `Map` over `Record` when deriving components from props/state.

### Map vs Record

The choice between a Map and a Record often depends on the specific use case. The use case getting flagged by eslint is use checking small, static value lookups.

#### Why Record is better for this use case

**Static Traceability**: Map lookups are harder for the linter to analyze to ensure it is a static, stable lookup so it is cautius and flags it regardless. It's trying to avoid unecessary re-renders.

**Reference Stability**: React requires component references to be stable across renders. Accessing a property on a constant object is a direct reference to a static value. A Map lookup is treated as a dynamic operation, which triggers safety checks in the linter to prevent unmounting and remounting components.  

**Optimization**: Plain objects are highly optimized by JavaScript engines for fast property access when the keys are strings. For small, static sets of data like a list of icons, the overhead of a Map is generally unnecessary.

**Tree-shaking**: Similar to the linter, the bundler has a hard time determining if the lookup will be static at runtime so it includes every entry even if it is not needed. A record helps the bundler identify what records are needed and only bundles those.

#### Comparison

Feature | Record (Object) | Map
-- | -- | --
Key Types | Limited to strings or symbols. | Any type, including objects or functions.
Optimization | Optimized for fast property access in engines. | Optimized for frequent additions and removals.
Order | Not strictly guaranteed for all types. | Maintains insertion order of keys.
Analysis | Compatible with static linting rules. | Treated as dynamic runtime data.

</div></body></html>

### 🙋‍♀️ Proposed Solution

- [ ] Remove the rule disable line in `react.mjs` for the rule
- [ ] Fix errors that result

##### Before

```tsx
const elementMap = new Map([
  ["text", Text],
  ["heading", Heading],
]);

const Component = ({ as }) => {
  const El = elementMap.get(as ?? "text");

  return <El />;
}
```

##### After

```tsx
const ElementKey = "text" | "heading";

const elementMap: Record<ElementKey, JSX.Element> = {
  "text": Text,
  "heading": Heading,
};

const Component = ({ as }) => {
  const El = (as && as in elementMap) ? elementMap[as] : Text;

  return <El />;
}
```

### Strategy

 - **External lookups**: We should already be doing this but, make sure the records are defined outside of the component
 - **Strict types**: If possible, avoid primitites like (`string`, `number`) for autocomplete and avoid runtime errors
 - **Safety check**: Always verify the lookup exists before rendering it

### ✅ Acceptance Criteria

- [ ] Rule disabled line removed from config
- [ ] No errors related to the rule remain
- [ ] Strategy listed was followed

esizer:
🤖 Resolves #16676 

## 👋 Introduction

Renables the static components rule for react in eslint.

## 🕵️ Details

Turns out, the errors we were encountering were actually false positives. The rule is to avoid react components changing during render but, these are already static look ups. The issue is that, it is derived from a function that the rules cannot effectivley anaylse so it defaults to marking it as non-static and therefore not stable.

So, we simply ignored it with a comment as to why. The only other option would be to have an ever-growing if statement tree in every snapshot component which would be difficult to maintain over time so the disabling was added so we can keep the helper.

## 🧪 Testing

1. Run lint `pnpm lint --force`
2. Confirm no errors


mnigh (commented):
Can all of the `// eslint-disable-next-line react-hooks/static-components` lines be removed? eslint says: 

> Unused eslint-disable directive (no problems were reported from 'react-hooks/static-components').


## Screenshot
Image

esizer:
> Can all of the `// eslint-disable-next-line react-hooks/static-components` lines be removed? eslint says:
> 
> > Unused eslint-disable directive (no problems were reported from 'react-hooks/static-components').

Can you try `pnpm i` and then restarting VS Code eslint server? Cause these were fialing when I ran `pnpm lint --force` which is why I added those.

mnigh (approved):

PR #16942, opened 2026-06-01
[Bug] Fix column removal ordering

### LINKED ISSUE #16941: 🐛 Can't rollback migration

The `down` method of our api/database/migrations/2026_05_20_000000_add_talent_request_status_columns.php can't be run.  The database emits this error: `cannot drop column status of table pool_candidate_search_requests because other objects depend on it`.  This prevents rolling back the migration.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Run `./artisan migrate:status` to determine how many steps back `2026_05_20_000000_add_talent_request_status_columns.php` is.
2. Run `./artisan migrate:rollback --step=4` substituting in the the correct number of steps.
3. Observe that the rollback fails.

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

Reversing the order of the columns to drop should fix it.  The `status_weight` needs to be dropped first because it depends on the other columns and won't allow them to be dropped first.

### ✅ Acceptance Criteria

- [ ] Migration can be successfully rolled back.

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

petertgiles:
🤖 Resolves #16941 

## 👋 Introduction

Fixes the ordering of column removal to allow migration to properly rolllback.

## 🧪 Testing

1. Repeat the reproduction steps in the issue.
2. Observe successful rollback.

## 📸 Screenshot

image

### DIFF
--- a/api/database/migrations/2026_05_20_000000_add_talent_request_status_columns.php
+++ b/api/database/migrations/2026_05_20_000000_add_talent_request_status_columns.php
@@ -51,7 +51,7 @@ return new class() extends Migration
     public function down(): void
     {
         Schema::table('pool_candidate_search_requests', function (Blueprint $table) {
-            $table->dropColumn(['status', 'in_progress_details', 'closed_details', 'status_weight']);
+            $table->dropColumn(['status_weight', 'status', 'in_progress_details', 'closed_details']);
         });
     }
 };


brindasasi (approved):
Thanks for catching this ..LGTM!

PR #16951, opened 2026-06-02
[Fix] Excel user input crashing generation

### LINKED ISSUE #16565: 🐛 Prod logs: TypeError: array_intersect_key(): Argument #1 ($array)

In the job logs, it seems like the Excel is sometimes failing to perform a calculation and save the worksheet.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Open the log interface for the TalentAppInsights logging analytics workspace.
2. Run
> AppLogsPrd_CL
| where LogLevel == 'ERROR' and EventText startswith "TypeError: array_intersect_key(): Argument #1 ($array)"

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] No failures when saving Excel worksheets

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #16565 

## 👋 Introduction

Fixes an issue were user input could crash our excel file generation.

## 🕵️ Details

The isuse was some user input that was prefixed with an `=` character. `PhpSpreadsheet` sees this and infers that the cell should be a formula. It then tries to insert the forumla but, this was how the user submitted the data and was not a valid formula which is what resulted in the error.

This disables the ability for `PhpSpreadsheet` to try and auto-detect formulas and results in just the string being inserted rather that it trying to be converted into a formula.

```
Worksheet->writeCellFormula(Object(XMLWriter), '= To become mor...', Object(PhpOffice\PhpSpreadsheet\Cell\Cell))
```

## 🧪 Testing

2. Login as `admin@test.com`
3. Update some user input to start with `=`
4. Go and try and download an excel that contains that field
5. Confirm the download works with no error and the document contains the string version and not a formula

### DIFF
--- a/api/app/Generators/CommunityInterestUserExcelGenerator.php
+++ b/api/app/Generators/CommunityInterestUserExcelGenerator.php
@@ -17,7 +17,6 @@ use App\Traits\Generator\Filterable;
 use App\Traits\Generator\GeneratesFile;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Arr;
-use PhpOffice\PhpSpreadsheet\Spreadsheet;
 
 class CommunityInterestUserExcelGenerator extends ExcelGenerator implements FileGeneratorInterface
 {
@@ -68,7 +67,7 @@ class CommunityInterestUserExcelGenerator extends ExcelGenerator implements File
 
     public function generate(): self
     {
-        $this->spreadsheet = new Spreadsheet();
+        $this->spreadsheet = $this->newSpreadsheet();
 
         $sheet = $this->spreadsheet->getActiveSheet();
         $localizedHeaders = array_map(function ($key) {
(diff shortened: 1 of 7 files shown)


mnigh (commented):
There is no failure when generating the excel document, however it does format the formula. Are we ok with allowing this? Should we not be sanitizing the value?

## Screenshots

Image
Image

esizer:
> There is no failure when generating the excel document, however it does format the formula. Are we ok with allowing this? Should we not be sanitizing the value?
> 

Okay, thanks for that. Looks like it was working outside of the queue because it was not getting picked up. I made the change to apply it inside the queue which should solve that problem.

mnigh (approved):
LGTM

PR #16968, opened 2026-06-03
[Feat] Add tracked user model

### LINKED ISSUE #16861: ✨ New `TalentRequestTrackedUser` pivot model
### Details

As part of the new talent requests, we are going to be tracking more information about the users who have been identified and referred via a talent request. In order to accomplish that, we are creating a new pivot model/table between the new `TalentRequest` model and `User` model with some pivot data called `TalentRequestTrackedUser`

### Proposed Implementation

1. Create migration for new table `talent_request_tracked_users`
2. Add new enums:
   - `TalentRequestTrackedUserReferralDecision` (`REFERRED`, `NOT_REFERRED`)
   - `TalentRequestTrackedUserSelectionDecision` (`SELECTED`, `NOT_SELECTED`)
   - `TalentRequestTrackedUserNotReferredReason` (start with just `OTHER` until reason enums are solidified)
   - `TalentRequestTrackedUserNotSelectedReason` (start with just `OTHER` until reason enums are solidified)
3. Update `TalentRequestFactory` with method to add tracked users (something like `withTrackedUsers`) that creates realistic users who would appear in the talent request based on criteria in `ApplicantFilter`
4. Add this as a relationship on the `TalentRequest` model (maybe `trackedUsers`) along with a field in the schema that can filter based on status.

Mutations are tracked separately in #16963.

#### Schema

```graphql
type TalentRequestTrackedUser {
  id: UUID!
  user: UserWithSkillCount! @belongsTo
  referralDecision: LocalizedTalentRequestTrackedUserReferralDecision
  selectionDecision: LocalizedTalentRequestTrackedUserSelectionDecision
  notReferredReason: LocalizedTalentRequestTrackedUserNotReferredReason
  notSelectedReason: LocalizedTalentRequestTrackedUserNotSelectedReason
}

type TalentRequest {
  trackedUsers: [TalentRequestTrackedUser!] @hasMany
}
```

**Authorization note:** Use an `authorizedToView` scope on the user relationship rather than `@canResolved(ability: "view")` — `@canResolved` in list queries causes severe performance issues and can lead to page crashes. No changes to roles or permissions are needed at this time. 

Follow the same pattern as `PoolCandidateWithSkillCount`, i.e. the `UserWithSkillCount` type should run a `withSkillCount` scope, using the skills from the linked TalentRequest filters. 

### ✅ Acceptance Criteria

- [ ] New pivot model exists with all relevant fields (two decision fields, two reason fields)
- [ ] Migration created connecting this table to talent requests and users
- [ ] Factory updated so users can be seeded
- [ ] Queries return `UserWithSkillCount` via `authorizedToView` scope (not `@canResolved`)
- [ ] Tests for queries created, especially for skill count

esizer:
🤖 Resolves #16861 

## 👋 Introduction

Adds a new model `TalentRequestTrackedUser`.

## 🕵️ Details

This is the new pivot model between a `TalentRequest` and `User` with pivot information aobut the state of that tracked user. The contains:

 - Model
 - Factory
 - Migration
 - Schema changes
 - Scopes
 - Skill count matching
 - Tests

> [!NOTE]
> Notably, what this does not add is the actual queries and mutations for managing this. Those are to come in a different PR. 

## 🧪 Testing

1. Confirm the new model matches what is expected
2. Confirm tests cover the cases required (specficallty auth is important to have)

### DIFF
--- a/api/app/Models/TalentRequestTrackedUser.php
+++ b/api/app/Models/TalentRequestTrackedUser.php
@@ -4,6 +4,7 @@ namespace App\Models;
 
 use Database\Factories\TalentRequestTrackedUserFactory;
 use Illuminate\Database\Eloquent\Attributes\Table;
+use Illuminate\Database\Eloquent\Concerns\HasUuids;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Database\Eloquent\Relations\Pivot;
@@ -24,12 +25,13 @@ use Spatie\Activitylog\Support\LogOptions;
  * @property ?Carbon $updated_at
  * @property ?Carbon $deleted_at
  */
-#[Table(keyType: 'string', incrementing: false)]
+#[Table(name: 'talent_request_tracked_users', keyType: 'string', incrementing: false)]
 class TalentRequestTrackedUser extends Pivot
 {
     /** @use HasFactory<TalentRequestTrackedUserFactory> */
     use HasFactory;
 
+    use HasUuids;
     use LogsActivity;
 
     public function getActivitylogOptions(): LogOptions
--- a/api/app/Models/TalentRequestTrackedUser.php
+++ b/api/app/Models/TalentRequestTrackedUser.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace App\Models;
+
+use Database\Factories\TalentRequestTrackedUserFactory;
+use Illuminate\Database\Eloquent\Attributes\Table;
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+use Illuminate\Database\Eloquent\Relations\Pivot;
+use Spatie\Activitylog\Models\Concerns\LogsActivity;
+use Spatie\Activitylog\Support\LogOptions;
+
+/**
+ * Class TalentRequestTrackedUser
+ *
+ * @property string $id
+ * @property string $user_id
+ * @property string $talent_request_id
+ * @property string $referral_decision
+ * @property string $selection_decision
+ * @property string $not_referred_reason
+ * @property string $not_selected_reason
+ * @property Carbon $created_at
+ * @property ?Carbon $updated_at
+ * @property ?Carbon $deleted_at
+ */
+#[Table(keyType: 'string', incrementing: false)]
+class TalentRequestTrackedUser extends Pivot
+{
+    /** @use HasFactory<TalentRequestTrackedUserFactory> */
+    use HasFactory;
+
+    use LogsActivity;
+
+    public function getActivitylogOptions(): LogOptions
+    {
+        return LogOptions::defaults()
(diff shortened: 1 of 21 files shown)


api/app/Models/TalentRequestTrackedUser.php
```diff
@@ -0,0 +1,108 @@
...
+                    ->from('applicant_filter_skill')
+                    ->join('talent_requests', 'talent_requests.applicant_filter_id', '=', 'applicant_filter_skill.applicant_filter_id')
+                    ->whereColumn('talent_requests.id', 'talent_request_tracked_users.talent_request_id');
+            }),
+        ]);
+    }
+
+    /**
+     * @param  ?array<string>  $decisions
+     */
```
yonikid15: This can be removed or filled out to match the function

api/app/Models/TalentRequestTrackedUser.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        return $query->when(
+            $decisions,
+            fn (Builder $query) => $query->whereIn('referral_decision', $decisions)
+        );
+    }
+
+    /**
+     * @param  ?array<string>  $decisions
+     */
```
yonikid15: This can be removed or filled out to match the function

api/database/factories/TalentRequestTrackedUserFactory.php
```diff
@@ -0,0 +1,74 @@
...
+            'referral_decision' => TalentRequestTrackedUserReferralDecision::NOT_REFERRED->name,
+            'not_referred_reason' => $reason,
+            'selection_decision' => null,
+            'not_selected_reason' => null,
+        ]);
+    }
+
+    public function selected(): self
+    {
+        return $this->referred()->state(fn () => [
```
yonikid15, line 59: Design question: Just want to confirm that we always want to set the tracked user to "Referred" whenever they are selected or not selected? Is it possible to be selected (or not selected), and not referred?

yonikid15 (commented):
Wow nice work 🎉 Just a comment and question, otherwise it's looking good

api/database/factories/TalentRequestTrackedUserFactory.php
esizer: I don't think so but @Jerryescandon might be able to confirm that.

api/app/Models/TalentRequestTrackedUser.php
esizer: Removed both in [remove redundant PHPDoc](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16968/commits/a6585fd2fd1a35b5358fe8e4fad70fe6fad8d82a)

api/database/factories/TalentRequestTrackedUserFactory.php
Jerryescandon: Nope.
- You MUST be `Referred` in order to be `selected` or `not selected`.
- it should NOT be possible to be `Not referred` AND `selected`
- it should NOT be possible to be `Not referred` AND `not selected`

yonikid15 (approved):
Nice

PR #16973, opened 2026-06-03
[Feature] Better Xdebug experience

### LINKED ISSUE #16923: 🐛 PHP Debugging friction

We have a few minor but painful friction points when trying to do PHP debugging on the app:

1. The tokens expire after two minutes.  This makes it painful to use the API directly, like through GraphiQL or Altair.  The tokens in prod for CanadaLogin will live for one hour so I think we can safely bump that up at least that far.  We can always manually reduce it if it we need to test the refresh behaviour.
2. PHP-FPM will time a worker out after 60 seconds.  This matches prod but makes it difficult to use a step debugger, like Xdebug.  I think we can safely disable this locally.
3. In launch.json on L12 there is a setting specifically for WSL but this breaks the debugger for native Linux developers.  It would be good see if we can find a solution that works for both, or just duplicate the debug profile so that we have one for each.
4. There is no documentation to help new devs learn to debug PHP code.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] The four friction points above are improved

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

petertgiles:
🤖 Resolves #16923 

## 👋 Introduction

Improves the experience for using Xdebug.

Note that this PR is focused on the "native Linux" experience since that's what I use.  If you're using WSL or MacOS, this may not help you too much.  You're welcome to contribute to the PR to improve those experiences, too, though.

## 🕵️ Details

Main changes:
- Bumps the token lifetime to one hour.  This helps when using a GraphQL client like Graphiql or Altair that doesn't automatically refresh the token set.
- Disables the PHP-FPM request timeout.  This helps when using a step debugger in Xdebug.  Note that the Nginx gateway timeout will still occur after 60 seconds (HTTP 504) but the debugging session stays alive.
- Splits the Xdebug webserver debugging profile between WSL and native profiles.  This alleviates the need to manually edit the profile before using it on native Linux.
- Adds documentation to help users who are unfamiliar with Xdebug get started.

## 🧪 Testing

1. Read the Xdebug documentation
2. Try debugging on the webserver
3. Try debugging on the CLI

## 📸 Screenshot

image

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,7 +1,9 @@
 {
   "configurations": [
+    // For using Xdebug with the PHP runtime in the webserver container and your debugger is running in WSL.
+    // For webserver Xdebug configuration, see the infrastructure/conf/xdebug.ini file.
     {
-      "name": "Listen for PHP Xdebug from webserver",
+      "name": "PHP Xdebug webserver (WSL)",
       "type": "php",
       "request": "launch",
       "port": 9003,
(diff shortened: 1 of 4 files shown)


vd1992:
What are the steps to run debug in WSL? I remember trying it out quite a while ago but forgot. 
Seems to be more than just launching a session in "Run and debug"

petertgiles:
> What are the steps to run debug in WSL?

I don't have Windows on my work machine to test it but following the instructions in the "PHP running in the webserver container" section, using the WSL profile, should work.

yonikid15 (approved):
Everything is working as expected after going through the documentation 👍

PR #16990, opened 2026-06-04
[Refactor] Update talent request search

### LINKED ISSUE #16863: ♻️ Replace candidate search with user search for talent requests
### ♻️ Refactor

Add User-based talent-request matching to the API, all built on one clean `UserBuilder` matching scope:

1. New **count queries** (`countTalentRequestMatches`, `countTalentRequestMatchesByPool`) that search `User`, replacing the logic in the legacy `countApplicantsForSearch` / `countPoolCandidatesByPool` with no reference to the old `PoolCandidate` / search-request path.
2. A new paginated **list** query (`talentRequestMatches`) returning the matching Users, each with their **source** and the matching pools, as groundwork for the eventual results table.

Part of the `PoolCandidateSearchRequest` to `TalentRequest` migration.

### 🕵️ Details

Today the public SearchPage gets its counts from `countApplicantsForSearch` (total) and `countPoolCandidatesByPool` (per-pool cards), both driven by `ApplicantFilterInput` and rooted in the `PoolCandidate` / search-request world. We want fresh, User-based versions on a clean shared scope, plus the ability to list **which** users match and **why** (their `TalentRequestSource`). A user can match through more than one source:

1. **`PREQUALIFIED`**: a talent-searchable, available `PoolCandidate` (the only source wired in this ticket). The result points to the actual **Pool(s)** the user is qualified in.
2. **`AT_LEVEL`**: future; derived from community-interest / talent-nomination data. Extension point only.
3. **`ADVANCEMENT`**: future; extension point only.

Backend groundwork only: no front-end table.

> [!NOTE]
> The legacy `countApplicantsForSearch` / `countPoolCandidatesByPool` are left **unmodified** here. They are deleted later when the front end switches to the new queries (separate work). This ticket only adds the new equivalents alongside them.

### 🙋‍♀️ Proposed Solution

**Shared matching scope** (the heart of the ticket; everything else sits on it):

- `PoolCandidateBuilder::whereMatchesRequestFilters(?array $filters)`: the single definition of "a candidacy matching this request" (`whereAvailable` + `whereInTalentSearchablePublishingGroup`, plus filter-driven classification / work-stream / community / pool constraints). Used by the user match, the per-pool count, and the eager-load so they can't drift.
- `UserBuilder::whereMatchesRequest(?array $filters)`: joint location/flexible matching + a "1 or more source" OR group (`PREQUALIFIED` branch now, `AT_LEVEL` / `ADVANCEMENT` later) + the filter-constrained eager-load backing `qualifiedInPools` + the skill-count select when `skills` is supplied.
- `UserBuilder::addSkillCountSelect` / `withSkillCount`: `skill_count` correlated count, matching `TalentRequestTrackedUser::scopeWithSkillCount`'s shape (#16968).

**Count queries** (public, no `@guard`, auto-resolved resolver classes like `CountApplicantsForSearch`):

```graphql
countTalentRequestMatches(where: ApplicantFilterInput): Int!
countTalentRequestMatchesByPool(where: ApplicantFilterInput): [TalentRequestMatchPoolCount!]!

type TalentRequestMatchPoolCount {
  pool: Pool!
  count: Int!
}
```

- Total: `User::query()->whereMatchesRequest($filters)->count()`.
- Per-pool: matching users per pool, from `PoolCandidate::query()->whereMatchesRequestFilters($filters)` whose `user` also satisfies `whereMatchesRequest($filters)`, grouped by `pool_id`, `count(distinct user_id)`.

**List query** + flat wrapper type (no nested/polymorphic source objects; `@self` wraps the User model so the shared `User` type is untouched):

```graphql
talentRequestMatches(
  where: ApplicantFilterInput @scope(name: "whereMatchesRequest")
): [TalentRequestResult]!
  @advancedOrderBy
  @paginate(
    defaultCount: 10
    maxCount: 1000
    scopes: ["whereAuthorizedToView", "withSkillCount"]
    model: "App\\Models\\User"
  )
  @guard
  @canModel(ability: "viewAny")

type TalentRequestResult {
  id: UUID!
  user: User! @self
  sources: [TalentRequestSource!]!   # [PREQUALIFIED] for now
  qualifiedInPools: [Pool!]!         # ONLY pools matching the applicant filter
  skillCount: Int @rename(attribute: "skill_count")
}
```

**Enum**: `App\Enums\TalentRequestSource` (`PREQUALIFIED` wired; `AT_LEVEL`, `ADVANCEMENT` declared only), with `lang/en|fr/talent_request_source.php`. Canonical shared definition the future `source` column on `TalentRequestTrackedUser` (#16968) will cast to.

### ✅ Acceptance Criteria

- [ ] `UserBuilder::whereMatchesRequest` and `PoolCandidateBuilder::whereMatchesRequestFilters` define matching once, reused by the counts, the list query, and the eager-load.
- [ ] `countTalentRequestMatches(where)` returns the User-based total and agrees with the `talentRequestMatches` total for the same filter.
- [ ] `countTalentRequestMatchesByPool(where)` returns matching-user counts per pool (a user qualified in two matching pools contributes to both; non-matching pools absent). Both count queries are public and make no reference to the legacy `PoolCandidate` / search-request path.
- [ ] `talentRequestMatches(where)` returns `TalentRequestResult` rows, one per matching user, paginated, `@advancedOrderBy`, guarded with `viewAny`.
- [ ] A user is returned only with an available, talent-searchable qualified candidacy matching the filter; non-searchable/unavailable-only and no-candidacy users are excluded.
- [ ] `sources` returns `[PREQUALIFIED]`; `qualifiedInPools` contains **only** the pools matching the filter; `skillCount` matches the count of the user's matching skills and is null with no `skills` filter.
- [ ] `App\Enums\TalentRequestSource` exists with `PREQUALIFIED` / `AT_LEVEL` / `ADVANCEMENT`, registered in the schema, en/fr lang strings, usable as-is by a future `TalentRequestTrackedUser.source` column.
- [ ] The shared `User` GraphQL type is unchanged; the `User` model gains no feature-specific accessors.
- [ ] Legacy `countApplicantsForSearch` / `countPoolCandidatesByPool` are untouched.
- [ ] Feature tests cover the counts (total + per-pool), inclusion/exclusion, `sources`, filter-constrained `qualifiedInPools`, `skillCount`, and authorization.

esizer:
🤖 Resolves #16863 

## 👋 Introduction

Adds 3 new queries to replace the existing talent search.

## 🕵️ Details

This adds three new queries that are going to eventually replace the existing ones we use:

- `talentRequestMatches` retrieves users that match the filters only showing those the authenticated user has access to view
- `countTalentRequestMatches` is the raw count of users that is a public query
- `countTalentRequestMatchesByPool` is the number of users who match the filters grouped by the pools that also match the filters.

## :exclamation: Outstanding TOODs

1. API response shape
2. Additional filters
3. Excluding tracked users

### API response shape

We never landed on the shape of the response, this contains the flat version I wrote in the ticket but we were discussing a nested or union type. Might need more discussion before this is ready for review.

### Additional filters

We spent some time disucssing the filters we want but never officially landed on the specific filters. This chose to omit them for now but allows for adding them. Do we want to handle the additional filters in another PR scoped to just that, or do it here?

### Excluding tracked users

This accepts the `talentRequestId` so we could do that but, the tracked user code does not exist yet so we can't. We could wait for that PR to get merged or, move that to a new ticket with the scope of just adding those filters in.

## 🧪 Testing

> [!TIP]
> These are not wired up to any UI so you will need to test this with graphiql 

1. Refresh and build `make refresh-api; make seed-fresh;`
2. Excersize the queries and filters and confirmt they work as expected
3. Confirm test coverage seems good


api/app/GraphQL/Queries/CountTalentRequestMatchesByPool.php
```diff
@@ -0,0 +1,29 @@
...
+     */
+    public function __invoke($_, array $args): Collection
+    {
+        $filters = $args['where'] ?? [];
+
+        // count the matching users per pool: a candidacy that matches the request whose user
+        // also matches it overall. Same scopes as talentRequestMatches, so the totals agree.
+        return PoolCandidate::query()
+            ->whereMatchesTalentRequest($filters)
+            ->whereHas('user', fn ($user) => $user->whereMatchesTalentRequest($filters))
```
tristan-orourke, line 25: `user->whereMatchesTalentRequest` will call `poolCandidate->whereMatchesTalentRequest` again. Unfortunately, I think we do need to repeat it here, because other talent sources will mean it could fail to match a pool candidate at all in the future. But I think a comment acknowledging that here would be good.

esizer: I'm a little confused about this comment but let's see if I understand correctly :sweat_smile: Are you sugested we just expand the comment a bit to something like:

```php
// Count the matching users per pool: a candidacy that matches the request whose user
// also matches it overall. Same scopes as talentRequestMatches, so the totals agree.
// NOTE: $filters are intentionally applied to both the PoolCandidate and the User.
// While redundant now, this prevents false positives when alternative talent sources 
// are introduced in the future.
```

api/graphql/schema.graphql
```diff
@@ -1531,6 +1531,24 @@ type CandidateSearchPoolResult {
   candidateCount: Int! @rename(attribute: "candidate_count")
 }
 
+input TalentRequestMatchFilterInput {
+  applicantFilter: ApplicantFilterInput
+  talentRequestId: UUID # accepted but unused until the tracked-user pivot (#16968) lands
```
tristan-orourke: What do you think about making this attribute name more functionally descriptive, something like `excludeTrackedUsersFromTalentRequest: UUID`?

esizer: Oh, I never intended this to be the actual scope we use to exclude tracked users. Just there to support it. I was thinking, once we have the tracked users just a simple bool scope for `includeTrackedUsers: Boolean @scope` since I tihnk the default of not showing them is going to cover the majority of use cases, right?

api/graphql/schema.graphql
```diff
@@ -1665,6 +1683,23 @@ type Query {
...
+  ): [TalentRequestResult]!
+    @advancedOrderBy
+    @paginate(
+      defaultCount: 10
+      maxCount: 1000
+      scopes: ["whereAuthorizedToView"]
+      model: "App\\Models\\User"
+    )
+    @guard
+    @canModel(ability: "viewAny", model: "User")
```
tristan-orourke: Won't this restrict this to only admins? Shouldn't we be covered by the scope?

esizer: Oh, I tihnk our `viewAny` is incorrect then. That is not what that policy method is for :sweat_smile: Its supposed to be used if you can view any of the possible results. Looks like ours is more like "view every" not "view any" :confounded:

api/app/GraphQL/Queries/CountTalentRequestMatchesByPool.php
tristan-orourke: Close. More like
```
// NOTE: PoolCandidateBuilder->whereMatchesTalentRequest will run twice, since it is called here and again from within the call to $user->whereMatchesTalentRequest. While redundant now, this prevents false positives when alternative talent sources are introduced in the future.
```

api/graphql/schema.graphql
tristan-orourke: Lol I remember the two of us having this exact conversation before. 

Yes, in our permission system, the 'any' in our permissions could be considered to be 'every'. 
The alternative interpretation, using 'any' to mean 'some' or 'at least one', just hasn't been useful for a permission.

esizer: Yeah, ill remove this. I onyl added it becuase that was the intetion the laravel authors had for the `viewAny` and I just keep forgetting we treat it as an "every" rather than an "any" :stuck_out_tongue:

api/graphql/schema.graphql
tristan-orourke: I just want to be clear about how talentRequestId is being used here. Given the name of the query, `talentRequestMatches`, and the input `TalentRequestMatchFilterIntput`, it would be natural for someone reading the api (ie us, in the future :laughing: ) to assume that the talent request id we pass in is _the source of the filters_. Given that this is not the case at all, I think we should make that really clear.

esizer: Yeah, that would be very confusing. I agree :stuck_out_tongue_closed_eyes:

tristan-orourke:
Okay, outstanding questions:
> 1. API response shape
> 2. Additional filters
> 3. Excluding tracked users

For 2 and 3, I like your suggestion of creating new tickets for those.

For 1) the api response shape...

I've been thinking about my earlier suggestion of using unions, or something like that, and I think you're right that its overkill. We can have one field to show which talent sources were used here, and then other fields to show the instances of talent source evidence.

You currently have it returning the Pools the user has qualified in, but I think it should be PoolCandidate instead. Probably with an authorizedToView scope, just for safety.

Thinking ahead to more talent sources, it might look like this:
```
type TalentRequestResult {
  id: UUID!
  user: User! @self
  sources: [TalentRequestSource!]!   # [PREQUALIFIED, AT_LEVEL, and ADVANCEMENT]
  matchingPreQualifiedSources: [PoolCandidate!]         # ONLY for pools matching the search filters
  matchingAtLevelSources: [CommunityInterest!]         # ONLY for communities matching the search filters
  matchingAdvancementSources: [TalentNominationGroup!]    # ONLY for nominations matching the search filters
  skillCount: Int @rename(attribute: "skill_count")
}
```

> [!NOTE]
> Actually, here's a thought. There might be cases where we want to show badges for talent sources to users who don't necessarily have permission to view the whole PoolCandidate object. We've been talking about partial visibility in the eventual case of using Nominations for Advancement. What would be the easiest way to handle that. A thin type, wrapping the main source?

```
   matchingPreQualifiedSources: [PreQualifiedSource!]
}

type PreQualifiedSource {
   pool: Pool!
   poolCandidate: PoolCandidate # require view permission
   # other data to surface if we want, like maybe qualification status?
}
```

api/graphql/schema.graphql
esizer: okay, really quick before I make a change, what about `excludeTrackedByRequestId` just to cut a bit down on the verbosity? :sweat_smile:

esizer:
> Thinking ahead to more talent sources, it might look like this:

This looks good, I'll implement as-is :+1: As for adding in additional info for partial visiiblity, I think this makes sense but maybe as a future thing? Right now I tihnk we can be pretty confident that if you can see the user in the request, you should have access to the candidate, right? So, while it should make sense in the near future, I think starting out simple first and then adding in that extra layer later when needed seems reasonable. Waht do you think?

api/graphql/schema.graphql
tristan-orourke: Sure - maybe add a comment as well though?

api/app/Builders/PoolCandidateBuilder.php
```diff
@@ -105,6 +105,23 @@ public function whereInTalentSearchablePublishingGroup(): self
...
+
+        // skills/pools/community are already plain ids (ApplicantFilterInput @pluck); workStreams
+        // and classifications are not plucked, so workStreams still needs its id pulled out.
+        return $this->whereAvailable()
+            ->whereInTalentSearchablePublishingGroup()
+            ->whereAppliedClassificationsIn($filters['qualifiedInClassifications'] ?? null)
+            ->whereWorkStreamsIn(array_column($filters['qualifiedInWorkStreams'] ?? [], 'id'))
+            ->whereHasPoolCandidateCommunity($filters['community'] ?? null)
+            ->when($filters['pools'] ?? null, fn ($query, $pools) => $query->whereIn('pool_id', $pools));
+    }
```
yonikid15: This is missing filtering by skills (mentioned in your comments)

api/tests/Feature/TalentRequestMatchesTest.php
```diff
@@ -0,0 +1,387 @@
...
+            'user_id' => $user->id,
+            'pool_id' => $poolB->id,
+        ]);
+
+        $this->runMatches()
+            ->assertJsonPath('data.talentRequestMatches.paginatorInfo.total', 1)
+            ->assertJsonPath('data.talentRequestMatches.data.0.user.id', $user->id)
+            ->assertJsonCount(2, 'data.talentRequestMatches.data.0.matchingQualifiedInPoolSources')
+            ->assertJsonFragment(['id' => $poolA->id])
+            ->assertJsonFragment(['id' => $poolB->id]);
```
yonikid15: In file `TalentRequestMatchesTest.php` at line 240:

`assertJsonFragment(['id' => $poolA->id])` is too generic and can match unrelated `id` fields (e.g., `user.id`), making the test pass even if the pool IDs are missing. Assert on a more specific structure/path, such as the nested `pool.id` inside `matchingQualifiedInPoolSources`, or assert the exact list of pool IDs under `data.talentRequestMatches.data.0.matchingQualifiedInPoolSources.*.pool.id`.

api/tests/Feature/TalentRequestMatchesTest.php
```diff
@@ -0,0 +1,387 @@
...
+    {
+        $this->graphQL($this->query, ['where' => []])
+            ->assertGraphQLErrorMessage('Unauthenticated.');
+    }
+
+    public function testCountTotalsMatchingUsersAndIsPublic(): void
+    {
+        $pool = Pool::factory()->candidatesAvailableInSearch()->create();
+        $this->matchingUser($pool);
+        $this->matchingUser($pool);
```
yonikid15, line 343: The count queries are embedded repeatedly as raw strings. Pulling them into class constants or helper methods (runCountMatches, runCountByPool) would improve maintainability and reduce copy/paste drift.

api/tests/Feature/TalentRequestMatchesTest.php
```diff
@@ -0,0 +1,387 @@
...
+                                'sources' => [TalentRequestSource::QUALIFIED_IN_POOL->name],
+                                'matchingQualifiedInPoolSources' => [['pool' => ['id' => $pool->id]]],
+                                'skillCount' => null,
+                            ],
+                        ],
+                        'paginatorInfo' => ['total' => 1],
+                    ],
+                ],
+            ]);
+    }
```
yonikid15, line 127: assertJson is useful, but in tests expecting a single exact shape, assertExactJson or more targeted assertJsonPath checks can make failures easier to diagnose and reduce accidental pass-through.

yonikid15 (changes requested):
Looking pretty good just a few things that can be improved

api/app/Builders/PoolCandidateBuilder.php
esizer: @Jerryescandon in order for a user to have the "qualified in pool" talent source, what part of the filters do they need to match in the pool? Right now we are checking:

 - Publishing group is IT or OTHER
 - Classification matches requested
 - Work stram matches requested
 - Community matches requested
 - Pool matches requested
 
 Should this also confirm the pool was requesting:
 
  - At least one of the talent requested skills
  - All the talent requested skills
  - Or, do we not care about skills at all?

api/tests/Feature/TalentRequestMatchesTest.php
esizer: All test changes should be covered in [update test](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16990/commits/f653ce5fbdef219c554af16ab01751d7a3c41201)

api/app/Builders/PoolCandidateBuilder.php
esizer: Spoke with @Jerryescandon and this code is functioning as expected. The comment was just confusing things so I updated it in [update comment to be more descriptive/accurate](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16990/commits/f0ecc3def52e8c5df6c64c3a20f4f760ac1aa9b3)

api/tests/Feature/TalentRequestMatchesTest.php
yonikid15: @esizer You can change this to use assertExactJson

esizer: Done in [use exact json](https://github.com/GCTC-NTGC/gc-digital-talent/pull/16990/commits/196cf4d918a3c883ea56ce54dce8b2325fe27992)

yonikid15 (approved):
Hopefully we covered everything. Looks good now 👍 

Image

PR #17009, opened 2026-06-05
[Refactor] Reduce spacing around `<Card />`

### LINKED ISSUE #16051: ♻️ Reduce mobile padding
Image

### ♻️ Refactor

A small, if not widespread, change that reduces the padding (highlighted in yellow below) on either side of card elements on the site in narrow viewports to provide even more reading space while still maintaining the aesthetic.

Image

### ✅ Acceptance Criteria

- [ ] Update mobile padding to `x.5` from `x1` across the site

vd1992:
🤖 Resolves #16051

## 👋 Introduction

Reduce the spacing around the component. In the smallest viewport only

Tricky bit as the padding came from the parent `<Container />` . As the issue dealt with the cards in specific, I decided to solve by negative margins on the cards themselves rather than ripple out a much larger change

## 🧪 Testing

1. Navigate to pages with `<Card />`
2. Shrink window
3. Also, look at Chromatic

#### Other (Unchanged)

image

#### Smallest viewport 

image

### DIFF
--- a/packages/ui/src/components/Card/Card.tsx
+++ b/packages/ui/src/components/Card/Card.tsx
@@ -7,7 +7,7 @@ import type { SeparatorProps } from "../Separator/Separator";
 import { Grid, GridItem } from "./CardGrid";
 
 const card = tv({
-  base: "rounded-md bg-white text-black dark:bg-gray-600 dark:text-white",
+  base: "-mx-3 rounded-md bg-white text-black xs:mx-0 dark:bg-gray-600 dark:text-white",
   variants: {
     space: {
       xs: "p-3",


esizer (changes requested):
I feel like using negative margins will lead to unintended consquences here 😅 

We might need to get more information to solve this one rather than baking in negative margins to a component

vd1992:
Hm... Any opinions, @JoshBeveridge ?

JoshBeveridge:
While I applaud the creativity, I'm generally inclined to agree with @esizer here on the use of negative margins. They tend to be... unruly 😂. If there's any info I can provide to help move different solutions forward, let me know!

vd1992:
Well it is the lightest solution, probably. 
Other possible ideas, a prop on `<Container />` to switch to card spacing then updating to add it in a bunch of places
Or, accept shorter padding on anything nested in a container (for the smallest viewport)

vd1992:
Shelved, for now

PR #17029, opened 2026-06-09
[Feature] Talent request tracked user mutations

### LINKED ISSUE #16963: ✨ `TalentRequestTrackedUser` mutations
### Details

Adds the mutation layer for the `TalentRequestTrackedUser` pivot model created in #16861. Six separate mutations support bulk state changes (refer/not-refer, select/not-select) with validation against the defined state transition rules.

### Proposed Implementation

Six separate mutations, each operating on lists to support bulk actions. Each mutation must validate state transitions according to the [state transition rules (Figma board)](https://www.figma.com/board/xKPMW7MbDKADeADy6QgmZk/User-tracking-status?node-id=0-1&p=f&t=weNBTptCtbMML1I4-0).

- `createTrackedUsersReferred` — takes a list of user IDs
- `createTrackedUsersNotReferred` — takes a list of user IDs and a single `NotReferredReason` enum
- `updateTrackedUsersReferred` — takes a list of TrackedUser IDs
- `updateTrackedUsersNotReferred` — takes a list of TrackedUser IDs and a `NotReferredReason` enum
- `updateTrackedUsersSelected` — takes a list of TrackedUser IDs
- `updateTrackedUsersNotSelected` — takes a list of TrackedUser IDs and a single `NotSelectedReason` enum

For illegal state transitions, use GraphQL partial errors — update the legal ones and return errors for failures.

### ✅ Acceptance Criteria

- [ ] Six mutations added with state transition validation
- [ ] Illegal state transitions return partial errors (legal updates succeed, failures reported)
- [ ] Tests covering happy-path and illegal-transition scenarios for each mutation

yonikid15:
🤖 Resolves #16963 

## 👋 Introduction

- Adds mutations (see issue)
- Adds validation
- Adds tests

## 🧪 Testing

1. Ensure mutations work. I recommend using `admin/graphiql`.
2. Go through A/C and figma board. Ensure code reflects the business logic required.


api/app/GraphQL/Mutations/TalentRequestTrackedUserMutator.php
```diff
@@ -0,0 +1,109 @@
...
+                    [
+                        'referral_decision' => TalentRequestTrackedUserReferralDecision::REFERRED->name,
+                        'not_referred_reason' => null,
+                    ],
+                );
+            }
+        });
+
+        return true;
+    }
```
esizer, line 34: For both of these create bulk actions, I think using sync would be better than looping 🤔 

```php
$talentRequest = TalentRequest::findOrFail($args['talentRequestId']);
$userIds = $args['userIds'];

$syncData = collect($userIds)->associateWith(fn () => [
  'referral_decision' => TalentRequestTrackedUserReferralDecision::REFERRED->name,
  'not_referred_reason' => null,
])->toArray();

DB::transaction(function () use ($talentRequest, $syncData) {
  $talentRequest->trackedUsers()->syncWithoutDetaching($syncData);
});

return true;
```

api/app/Models/TalentRequestTrackedUser.php
```diff
@@ -99,4 +101,40 @@ public function scopeWhereSelectionDecisionIn(Builder $query, ?array $decisions)
             fn (Builder $query) => $query->whereIn('selection_decision', $decisions)
         );
     }
+
+    public function referred()
```
esizer: Sicne we call `referred` in other methods, should we pass in an arg to run save optionally so we don't hit the DB twice?

```php
public function referred(bool $save = true) {
  // ...
  if($save) {
    $this->save();
  }
}

public function selected()
{
  $this->referred(save: false);
  
  // ...
}
```

api/graphql/schema.graphql
```diff
@@ -3710,7 +3712,36 @@ type Mutation {
     @guard
     @canFind(ability: "update", find: "id", scopes: ["withPolicyEagerLoads"])
     @validator
-
+  createTrackedUsersReferred(userIds: [UUID!]!, talentRequestId: UUID!): Boolean
+    @field(resolver: "TalentRequestTrackedUserMutator@bulkCreateReferred")
+    @validator(class: "App\\GraphQL\\Validators\\Mutation\\CreateTrackedUsersReferredValidator")
```
esizer: For each of these mutations, lighthouse-php has some auto-resolution of the valdiators. So, you should be save to remove these as long as the validators are in the correct namespace.

api/tests/Feature/TalentRequestTrackedUserTest.php
esizer, line 1: Can you get the 🤖 to write some tests for the failure paths? I only see the happy paths here. We will need to make sure we test authorization and validation errors as well.

api/app/GraphQL/Mutations/TalentRequestTrackedUserMutator.php
esizer, line 1: We should probably wrap all of the loop mutations in a transaction as well.

esizer, line 1: This is a minor suggestion that I don't really care if we do or not. But, I notice we have a pattern here where we would do a helper method. You could probably get a 🤖 to do this.

```php
private function bulkUpdateTrackedUsers(array $ids, callable $callback): Collection
{
    if (empty($ids)) {
        return collect();
    }

    $trackedUsers = TalentRequestTrackedUser::whereIn('id', $ids)->get();

    DB::transaction(function () use ($trackedUsers, $callback) {
        foreach ($trackedUsers as $trackedUser) {
            $callback($trackedUser);
        }
    });

    return $trackedUsers;
}

public function bulkUpdateReferred($_, array $args): Collection
{
    return $this->bulkUpdateTrackedUsers($args['ids'], fn ($user) => $user->referred());
}
```

esizer (changes requested):

api/app/GraphQL/Mutations/TalentRequestTrackedUserMutator.php
yonikid15: Discussed: Current logic is fine

api/app/Models/TalentRequestTrackedUser.php
yonikid15: [Improve referred method logic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17029/commits/9592ab862428bac55ab5d7efd809ae553c2406b8)

api/graphql/schema.graphql
yonikid15: [Move validator files](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17029/commits/11cd679ef0b0bcb157994dc0ebea24e01b629b0e)

[Remove old files](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17029/commits/682333aa51b706ea2561ede892263deba6c0ab08)

api/tests/Feature/TalentRequestTrackedUserTest.php
yonikid15: [Add failure tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17029/commits/c35f3f3fb6e70c88b437ce7b5509c47600d86934)

api/app/GraphQL/Mutations/TalentRequestTrackedUserMutator.php
yonikid15: [Use transaction for other mutations](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17029/commits/f76628f44d74eec3d1e3dc684f73200bee865218)

yonikid15: [Use helper method](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17029/commits/e9125a89dacaf51cb2e47899900db07b896acd19)

esizer (changes requested):
Very close, major issue though. We don't have any `guard` or `@can*` directives here. We will want to add those for auth and update the tests with that.

api/app/Policies/TalentRequestPolicy.php
```diff
@@ -72,6 +73,34 @@ public function update(User $user, TalentRequest $talentRequest)
...
+            ->get()
+            ->pluck('talentRequest')
+            ->filter();
+
+        if ($talentRequests->isEmpty()) {
+            return false;
+        }
+
+        return $talentRequests->every(fn (TalentRequest $talentRequest) => $this->update($user, $talentRequest));
+    }
```
esizer, line 118: This has the potential to be pretty expensive if called on many IDs. I think we should use our team checker helper. Maybe something like:

```php
// If user can update any, do an early return
if ($user->isAbleTo('update-any-talentRequest')) {
  return true;
}

// Get teams, not talent requests since
// if bulking across many talent requests,
// should be much smaller count to loop over
$teams = TalentRequestTrackedUser::query()
  ->whereIn('id', $args['ids'] ?? [])
  ->with('talentRequest.community.team')
  ->get()
  ->pluck('talentRequest.community.team')
  ->filter()
  ->unique('id'); // distinct teams only

return $teams->isNotEmpty() &&
  $teams->every(fn ($team) =>
  // Has caching per-team so,
  // helps reduce calls to `isAbleTo` which can be expensive
  $this->checkTeamPermission($user, [$team], 'update-team-talentRequest'));
```

Also, still need the tests to prove users who should NOT be able to, cannot.

esizer (changes requested):

api/app/Policies/TalentRequestPolicy.php
yonikid15: How is this?

```php
/**
     * Determine whether the user can update tracked users for one or more talent requests.
     *
     * @param  array{ids?: array<int, string>}  $args
     * @return Response|bool
     */
    public function updateTrackedUsers(User $user, array $args)
    {
        if ($user->isAbleTo('update-any-talentRequest')) {
            return true;
        }

        $ids = $args['ids'] ?? null;

        if (! is_array($ids) || empty($ids)) {
            return false;
        }

        $uniqueIds = array_values(array_unique($ids));

        if (empty($uniqueIds)) {
            return false;
        }

        $trackedUsers = TalentRequestTrackedUser::query()
            ->whereIn('id', $uniqueIds)
            ->with('talentRequest.community.team')
            ->get();

        // Deny when any requested tracked user id cannot be resolved.
        if ($trackedUsers->count() !== count($uniqueIds)) {
            return false;
        }

        // Preserve update() behavior: deny when related talent request or team is missing.
        if ($trackedUsers->contains(fn (TalentRequestTrackedUser $trackedUser) => ! isset($trackedUser->talentRequest?->community?->team))) {
            return false;
        }

        $teams = $trackedUsers
            ->pluck('talentRequest.community.team')
            ->filter()
            ->unique('id');

        if ($teams->isEmpty()) {
            return false;
        }

        return $teams->every(fn ($team) => $user->isAbleTo('update-team-talentRequest', $team));
    }
```

esizer: that has all of the same problems as the first iteration sadly, just with more code :cry: 

The big thing is, we only care about the teams and should use the team helper we already have which is the big difference.

All of this is fairly unecessary imo:

```php
        // This can probably stay but is not really necessary
        // since passing `[]` will return 0 rows and still fail
        // but, it does avoid a trip to the DB so fine to keep
        if (! is_array($ids) || empty($ids)) {
            return false;
        }

        // Not sure if we care about uniqeness, if it shows up
        // twice, it sohuld result in the same response
        $uniqueIds = array_values(array_unique($ids));
        if (empty($uniqueIds)) {
            return false;
        }

       // This is entirely unecessary and does a second redundant trip to the DB
        $trackedUsers = TalentRequestTrackedUser::query()
            ->whereIn('id', $uniqueIds)
            ->with('talentRequest.community.team')
            ->get();

        // Deny when any requested tracked user id cannot be resolved.
        if ($trackedUsers->count() !== count($uniqueIds)) {
            return false;
        }

        // This is also unecessary and expensive, we only care about the team
        if ($trackedUsers->contains(fn (TalentRequestTrackedUser $trackedUser) => ! isset($trackedUser->talentRequest?->community?->team))) {
            return false;
        }
```

esizer (approved):

PR #17037, opened 2026-06-10
[Feat] Arbitrary table actions

### LINKED ISSUE #17036: ✨ Support arbitrary actions in table
### Details

We need the ability to pass arbitrary actions to the table. Currently, it only accepts the "download" options but we need to be able to pass annything in to support the bulk actions on talent requests.

### Design file & designer

_No response_

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

_No response_

### ✅ Acceptance Criteria

- [ ] Table component accepts any number of different actions
- [ ] Actions are applied to each selected row
- [ ] Action functions similar to how downloads work

esizer:
🤖 Resolves #17036 

## 👋 Introduction

Adds the option to add arbitrary actions to our table component.

## 🕵️ Details

This is rerquired to support actions for setting status for multiple talent request users.

## 🧪 Testing

1. Run storybook `pnpm storybook`
2. Navigate to the table with actions story
3. Confirm it does what you expect

## 📸 Screenshot

swappy-20260610_091903

### DIFF
--- a/apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx
+++ b/apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx
@@ -345,12 +345,7 @@ const Actions = ({
                   <Bullet />
                   <DropdownMenu.Root>
                     <DropdownMenu.Trigger btnProps={actionButtonStyles}>
-                      {intl.formatMessage({
-                        defaultMessage: "Actions",
-                        id: "PtnJL7",
-                        description:
-                          "Trigger label for the bulk actions menu on selected table rows",
-                      })}
+                      {intl.formatMessage(tableMessages.actions)}
                     </DropdownMenu.Trigger>
                     <DropdownMenu.Popup
                       positionerProps={{ align: "end", collisionPadding: 2 }}
--- a/apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx
+++ b/apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx
@@ -353,7 +353,6 @@ const Actions = ({
                       {actions.map((action, index) => (
                         <DropdownMenu.Item
                           key={index}
-                          disabled={action.disabled}
                           onClick={() => handleActionClick(action)}
                         >
                           {action.label}
--- a/apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx
+++ b/apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx
@@ -340,7 +340,7 @@ const Actions = ({
                   )}
                 </span>
               )}
-              {actions?.length ? (
+              {actions?.length && selectedRowIds?.length ? (
                 <>
                   <Bullet />
                   <DropdownMenu.Root>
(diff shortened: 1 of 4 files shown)


apps/web/src/components/Table/ResponsiveTable/types.ts
```diff
@@ -114,6 +114,15 @@ export interface DownloadDef {
       };
 }
 
+export interface TableAction {
+  /** Item label */
+  label: ReactNode;
+  /** Disable the item */
+  disabled?: boolean;
```
yonikid15: I don't think we should give the option to disable an action. It just creates confusing UI for the user. If the action isn't authorized then it shouldn't make it into the list of actions at all.

apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx
```diff
@@ -220,13 +225,23 @@ const Actions = ({
...
     toast.warning(intl.formatMessage(tableMessages.noRowsSelected));
   };
 
+  const handleActionClick = (action: TableAction) => {
+    if (count <= 0) {
+      handleNoRowsSelected();
+      return;
+    }
+    action.onClick?.(selectedRowIds);
+  };
```
yonikid15: Why don't you inline this method like the other buttons?

apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx
```diff
@@ -325,6 +340,29 @@ const Actions = ({
                   )}
                 </span>
               )}
+              {actions?.length ? (
```
yonikid15:
```suggestion
              {actions?.length && selectedRowIds?.length ? (
```
This will hide the actions options until they have selected a row. This is the behaviour we talked about in a previous meeting.

yonikid15 (changes requested):
Just a few things to address.

apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx
esizer: I forget what we talked about. Are you saying I should make it always show the actions menu?

apps/web/src/components/Table/ResponsiveTable/types.ts
esizer: Thats a good call, removed in [remove ability to disable actions](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17037/commits/445950ecdf0614131fb2488c16448a9da34fb185)

apps/web/src/components/Table/ResponsiveTable/RowSelection.tsx
esizer: Makes sense to me, moved in [inline function](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17037/commits/de93670e7167ee29725b9a9723c2825e7e1585c2)

yonikid15 (approved):
LGTM 👍

PR #17045, opened 2026-06-10
[Fix] Updates `whereInTalentSearchablePublishingGroup` conditions

### LINKED ISSUE #16991: 🐛 Executive jobs publishing group filtered out of Talent Request estimates

On the find talent page, we allow for the selection of EX classifications, but filter out all processes from the "Executive jobs" publishing group. All four EX processes on our site have been run under the "executive jobs" publishing group, so this means the number estimate is always 0 on the find talent page.

This is no longer true once you actually look at the candidate matches for requests though. There you can actually get matches for EX classifications.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

Do not filter "executive jobs" publishing group out of the find talent estimates.
Update  `PoolCandidateBuilder->whereInTalentSearchablePublishingGroup` to filter out IAP group instead of filtering _in_ a whitelist.

### ✅ Acceptance Criteria

- [ ] It is possible to have a non-zero estimate count for EX classifications on the find talent page
- [ ] IAP posters are still filtered out of search

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

mnigh:
🤖 Resolves #16991.

## 👋 Introduction

This PR updates the `whereInTalentSearchablePublishingGroup` conditions to exclude the IAP PublishingGroup instead of including the IT and OTHER PublishingGroups.

## 🧪 Testing

1. `make refresh-api; make seed-fresh; pnpm i; pnpm run build:fresh`
2. Delete all pools and pool candidates
3. Create an IAP pool (publishing group is IAP)
4. Create a single pool candidate in the previously created IAP pool
5. Navigate to http://localhost:8000/en/search
6. Verify estimated candidates count does not include any candidates


esizer (approved):
LGTM! 👍

PR #17046, opened 2026-06-10
Talent coordinators can submit nominations to past events

### LINKED ISSUE #16995: ✨ Talent Coordinators can submit nominations for past talent events
### Details

The digital community wants to retroactively add all their talent event nominees to the platform. The easiest way for us to enable this is to allow them to create nomination events in the past (#16994), then allow community talent coordinators to submit nominations to those events (this ticket). So that's what we're doing.

And for them to submit noms, they need the link even after the event is closed. We're addressing that in #17001.

### Design file & designer

I would like to see a confirmation dialogue intercept the user before they begin a talent nomination for a talent event in the past. Only users who are allowed to submit talent nominations to talent events in the past should encounter this.

Here some proposed text:
This event is already closed, are you sure you want submit a nomination?
(Start nomination) - (Close)

### Proposed Implementation

So we don't want to remove the existing validation for submitting talent nominations, because most users will still need it. Only Community Talent Coordinators (and community admins) should be able to submit Talent Nominations to Talent Nomination Events where the close date is in the past.

I don't know what the best implementation would be, but if you need to create a new permission, I propose this one:
Image

### ✅ Acceptance Criteria

- [ ] Community talent coordinators (and community admin) can submit talent nominations to closed (past) talent nomination events.
- [ ] If allowed to submit talent nomination past talent even, user encounters confirmation dialogue before they start the nomination

### LINKED ISSUE #17001: ✨ Remove logic to hide nomination link for "closed" nomination events (always show it instead)
### Details

We recently added a link to submit nominations to the Talent Nomination Events detail page (see image, it's on the right side of the page, near the top of the "event details" tab, but in the bottom right of the screen capture.)  This link get hidden after the event is closed. That was fine at the time, but now we want to allow Community Talent Coordinators to submit nominations after the event is closed (#16995), so they will need to get this link somewhere! Let's just remove the logic that hides it from here.

Image

### Design file & designer

_No response_

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

Remove ternary.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/a99bad7b48b0bcae63b52ce029e080d9660c6038/apps/web/src/pages/TalentEvents/TalentEvent/DetailsPage.tsx#L140

### ✅ Acceptance Criteria

- [ ] "Copy nomination link" always shown on Event details page

bella77-69:
🤖 Resolves  #16995
🤖 Resolves #17001

## 👋 Introduction

This PR allows Community Talent Coordinators & admins to submit nominations to past talent nomination events. 

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. log in as `talent-coordinator@test.com` or `community@test.com`
3. Navigate to `admin/talent-events`
4. Select a event that has a status "Past"
5. Click the "Copy nomination link"
6. Open a new tab and paste nomination link
7. Verify the dialog appears "This event is already closed. Are you sure you want to submit a nomination?"
8. Verify the "Cancel" and close modal buttons take the user back to `admin/talent-events`
9. Verify the "Start nomination" button allows you to start the nomination process.

## 📸 Screenshot

image


api/app/Rules/TalentEventOpenForCreatingNominations.php
```diff
@@ -25,6 +26,13 @@ public function validate(string $attribute, mixed $value, Closure $fail): void
                 $now = Carbon::now();
 
                 if ($now > $eventClosing) {
+                    $user = Auth::user();
+
+                    // Allow community coordinators and admins to nominate for past events
+                    if ($user->roles()->whereIn('name', ['community_talent_coordinator', 'community_admin'])->exists()) {
```
petertgiles: We try to avoid "does a user have a role" check since we can and do sometimes shift around what permissions a particular role includes.  I recommend changing this to the form `if($user->isAbleTo('some-permission'))` like you'll see in most policy files.

petertgiles: Same for TalentEventOpenForUpdatingNominations

apps/web/src/pages/TalentNominations/NominateTalent/components/Instructions.tsx
```diff
@@ -1,28 +1,155 @@
...
 
 import useRoutes from "~/hooks/useRoutes";
 
-import type { BaseFormValues } from "../types";
 import useCurrentStep from "../useCurrentStep";
 import SubHeading from "./SubHeading";
 import UpdateForm from "./UpdateForm";
 
-const Instructions = () => {
+export const NominationInstructions_Query = graphql(/* GraphQL */ `
```
petertgiles: Could you please switch this to a fragment that is loaded by the parent page?  This will avoid the "waterfall of queries" pattern.  You can see this pattern on all other page components.

apps/web/src/pages/TalentNominations/NominateTalent/components/Instructions.tsx
```diff
@@ -1,28 +1,155 @@
...
+  const [showForm, setShowForm] = useState(false);
+
+  const [result] = useQuery<InstructionsQueryData>({
+    query: NominationInstructions_Query,
+    variables: { id: nominationId },
+    pause: !nominationId,
+  });
+
+  const canNominatePast = hasRequiredRoles({
+    toCheck: [
```
petertgiles: Can we check for the permission here as well, instead of the role?  I think Eric very recently added the ability to see what roles the current user has in the authorization context?

petertgiles (changes requested):
I just did a code readthrough so far.  Look good!  Could you look at these few things?

api/app/Rules/TalentEventOpenForCreatingNominations.php
```diff
@@ -25,6 +26,13 @@ public function validate(string $attribute, mixed $value, Closure $fail): void
...
+                        return;
+                    }
```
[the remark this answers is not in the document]
petertgiles: If it's safe to fall back to falsey (as it is here) you can also do `$user?->isAbletTo(...)`

api/app/Rules/TalentEventOpenForCreatingNominations.php
bella77-69: Hi @petertgiles sorry for the delay I was having some issues with the validation when I tried updating it to: 
```
if ($user?->isAbleTo('create-own-pastTalentNomination')) {
    return;
}
```

I took what you talked about in learning fam and used copilot chat in vs code to help me understand what was going on and to find a working solution. I updated it to use the `TeamHelpers::getTeamIdsForPermission` helper instead. So when an event is closed only users with the right permission (talent coordinators/admins) can still create nominations and it also makes sure they have that permission for the same team that owns the event. Happy to chat if you have any thoughts, insights or other ideas. 🙃[314f46e](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17046/commits/314f46eb72c122b8e96213394780c8f1272f871b)

petertgiles (approved):

api/app/Rules/TalentEventOpenForCreatingNominations.php
petertgiles: The regular "isAbleTo" is failing because this is a privileged permission being used on a non-privileged URL.  We will fix this in a followup issue.

PR #17082, opened 2026-06-11
[Refactor] Group permission checking on `PoolCandidate`

### LINKED ISSUE #16534: ♻️ Group permission checks on `PoolCandidate` model
### ♻️ Refactor

We check the same permissions many times in the `PoolCandidate` graphql type, it might make sense to group these for readaiblity and slight performance.

### 🕵️ Details

See: https://github.com/GCTC-NTGC/gc-digital-talent/issues/16132#issuecomment-4270669314

After investigating the usage of `@canRoot` and `@canResolved`, it was identified that we call these excessivley in the `PoolCandidate` type for the same permissions (`viewNotes` and `viewStatus`).

Each time we query these fields, it has to check the permission during resolution. We should group these so common permission checks are together and lighthouse only needs to check that permission once per candidate rather than potentially 15 times in the case of status.

### 🙋‍♀️ Proposed Solution

See: https://github.com/GCTC-NTGC/gc-digital-talent/issues/16132#issuecomment-4270669314

Effectivley, we align these `@canRoot` with the underlyuing permissions as noted in https://github.com/GCTC-NTGC/gc-digital-talent/issues/16132#issuecomment-4280987715 and then add them to groups like:

```graphql
type PoolCandidateStatusMeta { }

type PoolCandidateNotesMeta { }

type PoolCandidate { 
  statusMeta: PoolCandidateStatusMeta @canRoot(ability: "viewStatus") @self
  notesMeta: PoolCandidateNotesMeta @canRoot(ability: "viewNotes") @self
}
```

### ✅ Acceptance Criteria

- [ ] Permission checks are grouped an unique on `PoolCandidate` type
- [ ] Tests implemented
- [ ] Frontend updated with new type shape

vd1992:
🤖 Resolves #16534

## 👋 Introduction

Groups fields into two chunks by policy reference

`PoolCandidateAssessmentData `
`PoolCandidateStatusData `

## 🕵️ Details

Some discussion has lead to spinoff issues


tristan-orourke:
Here's my suggestion:

1. We move candidateStatus and candidateInterest, and maybe statusUpdatedAt, back into the root PoolCandidate type and remove the special permissions. 
2. We add comments to those fields explaining how they summarize the other fields in a way that's meant to be visible to anyone who can see the PoolCandidate object. 
3. We remove the view-own-applicationStatus permission from the applicant role

The biggest risk here is removing the permission. I _think_ the only fields the applicant actually uses are candidateStatus, candidateInterest, and statusUpdatedAt, but that will require more careful confirmation than I can manage right now.

vd1992:
> The biggest risk here is removing the permission. I _think_ the only fields the applicant actually uses are candidateStatus, candidateInterest, and statusUpdatedAt, but that will require more careful confirmation than I can manage right now.

Candidate does interact with other fields. Some more of these fields can be less guarded like placement. 
As for `status`, I don't know if it ought to reference `candidateStatus` instead

#### `status`

image

#### `placedAt`

image

tristan-orourke:
Hmm. In CreateApplicationPage.tsx, it looks like the only fields actually used in that query are `id` and `pool.id`. I think we can delete status, archivedAt, and submittedAt from the query. _But._ This file is getting removed entirely in https://github.com/GCTC-NTGC/gc-digital-talent/pull/17084! So maybe coordinate with @esizer, and let that PR block this one.

RecruitmentProcessesPreviewList uses status to filter on the Qualified status, which could easily be replaced with CandidateStatus.Qualified. And similarly to CreateApplicationPage, it looks like it queries for placedAt but doesn't use it anywhere. RecruitmentProcessDialog uses statusUpdatedAt, but also doesn't seem to use status or placedAt.

Incidentally, I think I found a bug? Line 90 of RecruitmentProcessesPreviewList seems to filter processes by communityId if communityId is present - but it doesn't do anything with the result! Array.filter doesn't filter in place, it returns a filtered copy, and we don't save that copy to a variable here.

esizer:
> Hmm. In CreateApplicationPage.tsx, it looks like the only fields actually used in that query are `id` and `pool.id`. I think we can delete status, archivedAt, and submittedAt from the query. _But._ This file is getting removed entirely in #17084! So maybe coordinate with @esizer, and let that PR block this one.

Yeah, we are in the process of deleting that page. These fields were used to confirm they did not have any existing applications but, the [backend already validates that](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e8da8de904d0ad294fc46adc0eb570ea0c790e4e/api/app/GraphQL/Validators/Mutation/CreateApplicationValidator.php#L24) so, we no longer needed them.

You can look at that branch and see, we removed the query that contained those fields and it did not get copied over to the new hook created that replaced that page.

vd1992:
That was not an exhaustive list of every place, rather, examples of the fields being accessed by applicant
Other places include
`RecruitmentProcessDialog`
`ApplicationsProcessesTaskCard`
`ReviewRecruitmentProcessPreviewList`
`pages/Applications/fragment.ts`
`DeleteApplicationDialog`

>Incidentally, I think I found a bug? Line 90 of RecruitmentProcessesPreviewList seems to filter processes by communityId if communityId is present - but it doesn't do anything with the result! Array.filter doesn't filter in place, it returns a filtered copy, and we don't save that copy to a variable here.

Hm, trying to investigate (doesn't help that we have multiple components named "RecruitmentProcesses") and it appears this would be rendered within `TalentNominationGroupProfilePage`

tristan-orourke:
I've been thinking about my previous suggestion, and I think it expands the scope of this ticket too much. I'll create a new ticket for cleaning up our pool candidate queries (some can drop the querying for status) and considering whether candidates should have permission to view all the status data.

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/Dialog/ApplicationStatusDialog/ApplicationStatusDialog.tsx
```diff
@@ -115,19 +119,21 @@ const ApplicationStatusDialog = ({ query }: ApplicationStatusDialogProps) => {
...
         )}
 
-        {application.status.value === ApplicationStatus.Disqualified && (
+        {application.applicationStatusData?.status.value ===
+          ApplicationStatus.Disqualified && (
           <DisqualifiedStatusForm {...commonProps} query={application} />
         )}
 
-        {application.status.value === ApplicationStatus.Removed && (
+        {application.applicationStatusData?.status.value ===
```
tristan-orourke: These ones are missing the `?.value` null check

api/graphql/schema.graphql
```diff
@@ -619,6 +619,44 @@ type PoolCandidateCategory {
   label: LocalizedString!
 }
 
+type PoolCandidateViewApplicationAssessmentData {
```
tristan-orourke: This name seems a bit verbose. I know you're matching the permission name, but I'm not sure that's necessary, and "Application" is basically a synonym for "PoolCandidate". How about `PoolCandidateAssessmentData`?

api/graphql/schema.graphql
```diff
@@ -619,6 +619,44 @@ type PoolCandidateCategory {
...
+    @rename(attribute: "veteran_verification")
+  veteranVerificationExpiry: Date
+    @rename(attribute: "veteran_verification_expiry")
+  priorityVerification: ClaimVerificationResult
+    @rename(attribute: "priority_verification")
+  priorityVerificationExpiry: Date
+    @rename(attribute: "priority_verification_expiry")
+}
+
+type PoolCandidateViewApplicationStatusData {
```
tristan-orourke: How about `PoolCandidateStatusData`

api/app/Policies/PoolCandidatePolicy.php
```diff
@@ -166,7 +166,7 @@ public function delete(User $user, PoolCandidate $poolCandidate)
         return $user->id === $poolCandidate->user_id && $user->isAbleTo('delete-own-draftApplication');
     }
 
-    public function viewStatus(User $user, PoolCandidate $poolCandidate)
+    public function viewApplicationStatus(User $user, PoolCandidate $poolCandidate)
```
tristan-orourke, line 182: Is making the permission names more explicit in this way a pattern we've used elsewhere?

api/graphql/schema.graphql
```diff
@@ -1767,50 +1749,6 @@ type Query {
     @find
     @guard
     @canQuery(ability: "view", scopes: ["withPolicyEagerLoads"])
-  poolCandidatesPaginated(
```
tristan-orourke: Seems a bit odd to do this in this ticket?

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/ApplicationSidebar.stories.tsx
```diff
@@ -33,20 +33,7 @@ import ApplicationSidebar, {
...
-  | "disqualificationReason"
-  | "removalReason"
-  | "expiryDate"
-  | "screeningStage"
-  | "assessmentStep"
-  | "pauseReferralsAt"
-  | "resumeReferralsAt"
-  | "pauseReferralsReason"
->;
+type ApplicationSidebarData = Pick<PoolCandidate, "applicationStatusData">;
```
tristan-orourke, line 36: That's a nice simplification 👍

tristan-orourke (changes requested):

tristan-orourke:
Created #17160 and #17161 for further permissions cleanup

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/Dialog/ApplicationStatusDialog/ApplicationStatusDialog.tsx
vd1992: Strange nothing appeared for linting
https://github.com/GCTC-NTGC/gc-digital-talent/pull/17082/changes/038680465bfafbdc0d4d47aada19e5f0b6abe46a

api/graphql/schema.graphql
vd1992: Sure https://github.com/GCTC-NTGC/gc-digital-talent/pull/17082/changes/b1f006db914d49abae694aae90516e2b5dc0828a

api/graphql/schema.graphql
vd1992: Aye aye https://github.com/GCTC-NTGC/gc-digital-talent/pull/17082/changes/b1f006db914d49abae694aae90516e2b5dc0828a

api/app/Policies/PoolCandidatePolicy.php
vd1992: I think we've got an odd pattern going, in some cases it is explicit naming and in some cases the opposite
I'm pushing for more obvious naming here because it is getting hard to know what exactly "status" means for a pool candidate anymore

api/graphql/schema.graphql
vd1992: I was manipulating the `PoolCandidate` type which this query returns, would've needed to check this query since it could have been affected. But since it is deprecated and no longer used... 🤷

vd1992: Oh, and this type not needed anymore
https://github.com/GCTC-NTGC/gc-digital-talent/pull/17082/changes/b9296b4c2111bb9ad7f06e639172815f2c20d368

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/Dialog/ApplicationStatusDialog/StatusContent.tsx
```diff
@@ -76,10 +78,14 @@ export const Content = ({ query, reason, submitProps }: ContentProps) => {
...
-            date: application.statusUpdatedAt
+            status:
+              application.applicationStatusData?.status.label.localized ??
+              notAvailable,
+            date: application.applicationStatusData?.statusUpdatedAt
               ? formatDate({
-                  date: parseDateTimeUtc(application.statusUpdatedAt),
+                  date: parseDateTimeUtc(
+                    application.applicationStatusData?.statusUpdatedAt,
+                  ),
```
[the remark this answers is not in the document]
vd1992: >use a non-null assertion

Let us not

apps/web/src/pages/ApplicantDashboardPage/components/ReviewApplicationDialog.tsx
```diff
@@ -260,13 +266,14 @@ const ReviewApplicationDialog = ({
...
+            {application?.applicationStatusData?.candidateStatus?.value ===
+            CandidateStatus.Expired ? (
               <FieldDisplay
                 label={intl.formatMessage(commonMessages.deadlineToApply)}
                 className="xs:col-span-2"
                 hasError={deadlineToApply(
                   pool.closingDate,
-                  application.candidateStatus.value,
+                  application.applicationStatusData.candidateStatus.value,
                 )}
```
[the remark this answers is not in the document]
vd1992: That's actually a good suggestion https://github.com/GCTC-NTGC/gc-digital-talent/pull/17082/changes/cce5cab083490781a6138bac55e9eeb25da2386e

api/graphql/schema.graphql
tristan-orourke: Fair... I'm going to be a little picky here and request you pull these deletions into their own ticket/PR, to ensure they're well documented.

vd1992: Okay
Issue: https://github.com/GCTC-NTGC/gc-digital-talent/issues/17203
Commit: https://github.com/GCTC-NTGC/gc-digital-talent/pull/17082/changes/9e415b51ac95c2bf5681ced2425e322939f4358d

tristan-orourke (approved):
Sorry this take so long to review. Looks good!

PR #17083, opened 2026-06-12
[Refactor] Localized string faker provider

### LINKED ISSUE #17033: ✨ Move faker localizedString to provider
### Details

We have a "fake localized string" function in api/database/factories/BaseFactory.php for use in factories.  I saw that faker actually has the ability to be extended https://fakerphp.org/#faker-internals-understanding-providers.  Doing this for the localizedString might be more idiomatic.  Additionally, it would allow things like chaining with `optional`  or chaining like `->localizedString()->word()`.  Also, allows use outside of factories.

There maybe be other opportunities for custom providers, too, like a list of Canadian cities and town.

### Design file & designer

_No response_

### 📸 Screenshots

Image

### 🌎 Localization

_No response_

### Proposed Implementation

_No response_

### ✅ Acceptance Criteria

- localizedString exists as a Faker provider

esizer:
🤖 Resolves #17033 

## 👋 Introduction

Moves our factory `localizedString` method to a faker provider.

## 🕵️ Details

This went really well, great find by @petertgiles . Since this is so easy, I wonder if it makes sense to move the other methods in that class to providers :thinking: 

## 🧪 Testing

1. Run seeders `make seed-fresh`
2. Confirm they run and seed localized strings properly

### DIFF
--- a/api/app/Faker/ExtendedFormatsProvider.php
+++ b/api/app/Faker/ExtendedFormatsProvider.php
@@ -4,7 +4,7 @@ namespace App\Faker;
 
 use Faker\Provider\Base;
 
-class LocalizedStringProvider extends Base
+class ExtendedFormatsProvider extends Base
 {
     /**
      * Generate a localized string array for 'en' and 'fr' locales.
--- a/api/app/Faker/ExtendedFormatsProvider.php
+++ b/api/app/Faker/ExtendedFormatsProvider.php
@@ -46,4 +46,45 @@ class ExtendedFormatsProvider extends Base
             'fr' => $suffixLocale ? "{$base} (fr)" : $base,
         ];
     }
+
+    /**
+     * Pick a single random name from a pure Enum case list.
+     *
+     * @template TEnum of \UnitEnum
+     *
+     * @param  class-string<TEnum>  $enumClass
+     *
+     * @phpstan-template TEnum of \UnitEnum
+     *
+     * @phpstan-param class-string<TEnum> $enumClass
+     *
+     * @phpstan-return string
+     */
+    public function enum(string $enumClass): string
+    {
+        $names = array_map(fn ($case) => $case->name, $enumClass::cases());
+
+        return $this->generator->randomElement($names);
+    }
+
+    /**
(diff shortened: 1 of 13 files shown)


petertgiles:
Nice!  I was hoping this would work out.

> I wonder if it makes sense to move the other methods in that class to providers

I could definitely see the enum one moving there.  It could be modeled on [the array element](https://fakerjs.dev/api/helpers.html#arrayelements) and [array elements](https://fakerjs.dev/api/helpers.html#arrayelements) helpers.

The firstOrCreate one I'm not sure about.  I'm not sure it makes sense to have a faker method running Eloquent create commands.  I'm actually not sure about the value of the helper itself - it's just hiding a one-liner anyway.  :laughing:   If we do move it over, I'd suggest a name change.  Maybe `randomOrCreate`.  The "first" part is not really the important part of the behaviour.

esizer:
> The firstOrCreate one I'm not sure about. I'm not sure it makes sense to have a faker method running Eloquent create commands. I'm actually not sure about the value of the helper itself - it's just hiding a one-liner anyway. 😆 If we do move it over, I'd suggest a name change. Maybe `randomOrCreate`. The "first" part is not really the important part of the behaviour.

Yeah, that one does seem like it doesn't belong there. I created it initially because even though it is one line there, it was 3-4 in other places and we call it so darn frequently :sweat_smile:

petertgiles:
> even though it is one line there, it was 3-4 in other places and we call it so darn frequently

That's fair.  I think we've had similar discussions before.  :laughing:   I'll tend to lean towards what's most expressive over what's shortest.  This isn't a big deal though - I'll go with whatever you decide.

petertgiles:
Let me know if you plan on moving over other things from the base class in this PR or would like me to review as-is.

esizer:
> Let me know if you plan on moving over other things from the base class in this PR or would like me to review as-is.

I'll move over the enum one but leave the other function as-is since I tihnk we both agree it probably does not belong in faker :+1:

petertgiles (approved):
Turned out really nice.  Good stuff!

PR #17084, opened 2026-06-12
[Refactor] Convert apply action to form

### LINKED ISSUE #16944: ♻️ Create job applications from a form submission instead of a page load
### ♻️ Refactor

Refactor how job applications get created as a form submission rather than navigation

### 🕵️ Details

Creating a record is an action, but `CreateApplicationPage` was built as a page whose only job is to run a create mutation and then redirect (it even says "this is not a real page" in a comment). Because it is a component, the mutation runs while the component renders, and the page needs a `useRef` counter to stop it running more than once. This is what the new `react-hooks/refs` rule flags, and the file only passes linting because the rule has been disabled on it.

The fix is to create the application from a form submission, so it only runs when someone submits instead of when a page loads.

### 🙋‍♀️ Proposed Solution

- Move `CreateApplication` into a submit handler. Change the "Apply now" trigger in `ApplicationLink` and `ApplyLink` to a `<form>` with a submit button that runs the
  mutation, then navigates.
- Remove `CreateApplicationPage` and its route.

### ✅ Acceptance Criteria

- [ ] The application is created on submit, not during render.
- [ ] `CreateApplicationPage` and its route are deleted.
- [ ] Applying from `ApplicationLink` and the IAP `ApplyLink` still creates a draft and lands on the application.

esizer:
🤖 Resolves #16944 

## 👋 Introduction

Updates the job application link to be a form submission.

## 🕵️ Details

This is a more idiomatic approach to this flow and has the added benefit of no longer misusing refs.

### Unauthenticated new

1. Rebuild `pnpm dev:fresh`
2. Make sure you are logged out
3. Navigate to a job advertisement
4. Click "Apply now"
5. Confirm you are taken to login and on return, it cereate the application and redirects you

### Unauthenticated existing

1. Ensure you have an existing application and remeber the process
2. Make sure you are logged out
3. Navigate to the job advertisement of the process in step 1
4. Click "Apply now"
5. Confirm you are taken to login and on return, you are redirected to your application

### Authenticated new

1. Login as any user
2. Navigate to a job advertisement
3. Confirm "Apply now" creates and application and redirects you

### Authenticated existing

1. Login as any user
2. Navigate to a job advertisement that you already created an applicaiton for
3. Confirm the link says "Continue..." and takes you to the existing application

### IAP

1. Repeat all steps with an IAP pool from the IAP homepage


yonikid15 (changes requested):
Wow nice work, everything works as expected after testing.
One thing we can improve is changing the text on the Job Card to say "Continue to application" if they have already applied (Authenticated new, or Authenticated existing.

Image

esizer:
> Wow nice work, everything works as expected after testing. One thing we can improve is changing the text on the Job Card to say "Continue to application" if they have already applied (Authenticated new, or Authenticated existing.

Did this PR change how that works? It sounds out of scope but if it used to do that and this PR broke it, then i can fix it.

yonikid15:
@GCTC-NTGC/developers I think this should be thoroughly tested before merging. Can we get others to test this as well?

yonikid15 (approved):
I'll make an issue for changing the test for the apply now text to "Continue to my application" since it's out of scope 😛

tristan-orourke (approved):
Finally had a chance to look through this and test it. LGTM!

PR #17088, opened 2026-06-12
[Feature] Show "Copy nomination link" on all nomination events

### LINKED ISSUE #16995: ✨ Talent Coordinators can submit nominations for past talent events

### LINKED ISSUE #17001: ✨ Remove logic to hide nomination link for "closed" nomination events (always show it instead)

bella77-69:
🤖 Resolves #17001 
🤖 Resolves #16995

## 👋 Introduction

This PR removes the code that was hiding the "Copy nomination link" button when a nomination event is "closed"/"past"

## 🧪 Testing

1. Build `pnpm dev:fresh`
3. Navigate to `admin/talent-events` and select a event that has a status "Past"
4. Verify the "Copy nomination link" is visible 

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/pages/TalentEvents/TalentEvent/DetailsPage.tsx
+++ b/apps/web/src/pages/TalentEvents/TalentEvent/DetailsPage.tsx
@@ -8,11 +8,7 @@ import type {
   FragmentType,
   LocalizedTalentNominationEventStatus,
 } from "@gc-digital-talent/graphql";
-import {
-  getFragment,
-  graphql,
-  TalentNominationEventStatus,
-} from "@gc-digital-talent/graphql";
+import { getFragment, graphql } from "@gc-digital-talent/graphql";
 import {
   Button,
   Card,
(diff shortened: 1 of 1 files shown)


bella77-69:
Will be resolved by https://github.com/GCTC-NTGC/gc-digital-talent/issues/16995

PR #17107, opened 2026-06-16
[Refactor] Pool fields have access restricted 

### LINKED ISSUE #16992: ♻️ Move closing reason out of pools model
### ♻️ Refactor

Our pools model is public. This is very much intended, but recently we have started collecting justifications for changing published pools, or closing them early. This information is also being saved to the pool model, thus is still public. We don't consider this an important vulnerability, because the info is certainly still ATIP-able. But having these available on the public internet is certainly not meeting the expectations of the Recruiters filling them out.

So let's ensure the pool justifications are not public.

### 🕵️ Details

_No response_

### 🙋‍♀️ Proposed Solution

~Maybe we could just record these in the activity log instead? Regardless, we should ensure they are restricted to community/department roles.~

### ✅ Acceptance Criteria

- [ ] Justifications for editing published pool, or closing published pool early are only available to the users who can actually set them.
- [ ] Implement with a policy check, typical of the schema
    - [ ] Should be the same roles as view-team-draftPool have "view-team-poolNotes"
    - [ ] platform admin new permission named "view-any-poolNotes"

vd1992:
🤖 Resolves #16992

## 👋 Introduction

Implements the suggested changes
Permission added, policy method added, attribute created, and frontend adjusted

## 🕵️ Details

Specified fields no longer publicly accessible

## 🧪 Testing

1. Approaching how you prefer assert the fields are not publicly available
2. Public pool view not broken 

## 🚚 Deployment

Run roles and permissions seeder


api/graphql/schema.graphql
```diff
@@ -522,7 +522,10 @@ type Pool implements HasRoleAssignments {
     @hasMany
     @orderBy(column: "sort_order", direction: ASC)
   closingDate: DateTime @rename(attribute: "closing_date")
-  closingReason: String @rename(attribute: "closing_reason")
+  closingReason: String
+    @rename(attribute: "closing_reason")
+    @canRoot(ability: "viewPoolNotes")
+  wasClosedEarly: Boolean
```
petertgiles: This should be safe to be Boolean!, right?

api/tests/Unit/Policies/PoolPolicyTest.php
```diff
@@ -235,6 +241,16 @@ public function testTeamRolesCanUpdateDraftPool(string $factoryMethod): void
...
 
+    // --- viewPoolNotes() ---
+
+    #[DataProvider('allTeamRolesProvider')]
+    public function testTeamRolesCanViewPoolNotes(string $factoryMethod): void
+    {
+        $user = $this->createContextualUser($factoryMethod, $this->primaryPool);
+        $this->assertTrue($this->ensureBool($this->policy->updateDraft($user, $this->primaryPool)));
+        $this->assertFalse($this->ensureBool($this->policy->updateDraft($user, $this->unrelatedPool)));
+    }
```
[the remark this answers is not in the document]
petertgiles: Good catch, bot.

api/tests/Unit/Policies/PoolPolicyTest.php
```diff
@@ -235,6 +241,16 @@ public function testTeamRolesCanUpdateDraftPool(string $factoryMethod): void
         $this->assertTrue($this->ensureBool($this->policy->updateDraft($user, $this->primaryPool)));
     }
 
+    // --- viewPoolNotes() ---
+
+    #[DataProvider('allTeamRolesProvider')]
```
petertgiles, line 246: Shouldn't there be a test for platform admin as well?

api/tests/Unit/Policies/PoolPolicyTest.php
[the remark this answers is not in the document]
vd1992: 🤦

api/graphql/schema.graphql
```diff
@@ -522,7 +522,10 @@ type Pool implements HasRoleAssignments {
     @hasMany
     @orderBy(column: "sort_order", direction: ASC)
   closingDate: DateTime @rename(attribute: "closing_date")
-  closingReason: String @rename(attribute: "closing_reason")
+  closingReason: String
```
petertgiles, line 525: I'm getting "unauthorized" when I try to query this as admin@test.com.  I did rerun the permission seeder, too.  Does it work for you?

petertgiles (changes requested):
Looking cool!  Just a few small things to check on.

tristan-orourke:
@claude review

tristan-orourke:
Looks like the github Claude reviews are out of our budget for now :( 
We could look into running it in a GitHub action, alternatively

api/graphql/schema.graphql
vd1992: Sure, why not https://github.com/GCTC-NTGC/gc-digital-talent/pull/17107/changes/d23cf7c71bdce45c01ec2bdbece3c6fc904aba14

api/tests/Unit/Policies/PoolPolicyTest.php
vd1992: I just copied the test above, but it is worth asserting too https://github.com/GCTC-NTGC/gc-digital-talent/pull/17107/changes/3c1af38a7604e9248cdcc206bd3aa35434d1c24f

api/graphql/schema.graphql
vd1992: Test passes? https://github.com/GCTC-NTGC/gc-digital-talent/pull/17107/changes/3c1af38a7604e9248cdcc206bd3aa35434d1c24f

api/graphql/schema.graphql
petertgiles: Oof, sorry.  I was hoping that would make code usage simpler.  I didn't anticipate that it would make you do a bunch of extra type gymnastics in TS.  :face_exhaling:   Well, thanks anyway.  :smiling_face_with_tear:

api/graphql/schema.graphql
vd1992: Testing it, without Graphiql haven't gotten Altair setup, by plopping it into a page. It works...?

image

petertgiles: The AI suggested this:

> @canRoot resolves the policy check against the root of the entire GraphQL operation, not the parent model. For a query like pool(id: 1) { closingReason }, the root is the Pool, so this works. But for a nested query like application { pool { closingReason } } (if that path exists), the root would be the Application — and $pool would either be null or the wrong type, causing a runtime error.

> @can would be the safer directive here since closingReason is a field on Pool and you want to check against the specific Pool instance. Verify that @canRoot is intentional and that no nested query paths expose closingReason.

This might be why some queries are working and some not.  @can is deprecated but maybe @canResolved is what we want to ensure that we're always running the check against the exact pool that has the field?

vd1992: As an aside, a little confusing, shouldn't other instances break? Like when querying user -> poolcandidates for these

image

petertgiles: Well, let me try it out.  I'll let you know.

vd1992: Still seems to work

image

petertgiles: OK, both work fine for me.  I was trying to use the unprivileged endpoint.  :facepalm:  Tbh, I'm not sure the difference between the two can directives.  I would expect the canRoot one to not work on the nested one but it's fine.  :shrug: 
image

petertgiles (approved):

PR #17111, opened 2026-06-16
[Bug] Fix email take

### LINKED ISSUE #17106: 🐛 Failed to take email address on log in.

Brinda observed this bug in UAT:

production.ERROR: Failed to take email address on log in.Attempt to read property "email" on null {"sub":"xxxxxxxxxx","email address":"[xxxxxxxxxx@gmail.com](mailto:xxxxxxxxxx@gmail.com)"}

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Log in with CanadaLogin on UAT

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

It looks like AuthController is trying to update an existing user who is sharing the email, even when there is not an existing user.  The error message is coming from an exception handler so the login can continue.  We just need to check if the existing user exists to avoid the logging noise from the exception handler.

### ✅ Acceptance Criteria

- [ ] Login can proceed without the exception logging

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

petertgiles:
🤖 Resolves #17106 

## 👋 Introduction

Fixes an exception (that is already being handled) to reduce error logging.

## 🕵️ Details

- checks if there is an existing user with the address before trying to take over the address
- checks if the email is not null before trying to take over the address

> [!TIP]
> I recommend viewing the code diff without whitespace changes.

## 🧪 Testing

1. Configure environment for CanadaLogin

- [ ] Logging in as a new user works without logging errors
- [ ] Logging in as an existing user works without logging errors
- [ ] Logging in as a user which takes over an email address from another user works without logging errors

### DIFF
--- a/api/app/Http/Controllers/AuthController.php
+++ b/api/app/Http/Controllers/AuthController.php
@@ -183,15 +183,17 @@ class AuthController extends Controller
                             ->orWhere('work_email', 'ilike', $incomingEmailAddress))
                         ->withTrashed()
                         ->first();
-                    if (strcasecmp($existingUser->email, $incomingEmailAddress) == 0) {
-                        $existingUser->email_backup = $existingUser->email;
-                        $existingUser->email = null;
+                    if ($existingUser) {
+                        if (strcasecmp($existingUser->email, $incomingEmailAddress) == 0) {
+                            $existingUser->email_backup = $existingUser->email;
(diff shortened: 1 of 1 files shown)


esizer (approved):
LGTM! 👍

PR #17114, opened 2026-06-16
[Feat] Add talent sources to tracked users in talent request

### LINKED ISSUE #17110: ✨ Add talent sources to `TalentRequestTrackedUser`
### Details

We have them on the graphql schema type `TalentRequestSource` but we also need them on the `TalentRequestTrackedUser` so we can show the sources from the current filters.

### Design file & designer

_No response_

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

We will need these to be computed fields on the model so they always match the talent requests applicant filter for now.

Likely,

- add some attributes to the model
- a scope or something to moidfy and add the subqueries

### ✅ Acceptance Criteria

- [ ] Talent sources added to `TalentRequestTrackedUser`
- [ ] Matches the existing implementation shape on the `TalentSource`

esizer:
🤖 Resolves #17110 

## 👋 Introduction

Adds the computed talent sources for a tracked user on a talent request.

## 🕵️ Details

These filters are intended to always match the current filters so we added a scope that adds the selects in. Unfortunatley, this means baking in a very light n+1. We may want to consider making these stored computed values when the following change:

 - One of the potential sources for a user are created or updated
 - The talent request applicant filter changes

## 🧪 Testing

1. Confirm the talent sources exist
2. Confirm the test coverage is accurate and tests pass

### DIFF
--- a/api/app/Models/TalentRequestTrackedUser.php
+++ b/api/app/Models/TalentRequestTrackedUser.php
@@ -86,10 +86,19 @@ class TalentRequestTrackedUser extends Pivot
     /** @return Attribute<array<string>, never> */
     protected function sources(): Attribute
     {
-        return Attribute::get(fn (): array => $this->has_prequalified_source
+        return Attribute::get(fn (): array => $this->hasQualifiedInPoolSource()
             ? [TalentRequestSource::QUALIFIED_IN_POOL->name]
             : []
         );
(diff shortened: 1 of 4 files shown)


yonikid15 (approved):
LGTM

PR #17115, opened 2026-06-16
[Debt] Updates `ExecutiveHomePage_Query` to use classification instead of publishing group

### LINKED ISSUE #17096: ♻️ `ExecutiveHomePage_Query` uses classification
### ♻️ Refactor

Update `ExecutiveHomePage_Query` to use classification instead of publishing group.

### 🕵️ Details

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] `ExecutiveHomePage_Query` use classification instead of publishing group for filter

mnigh:
🤖 Resolves #17096.

## 👋 Introduction

This PR update `ExecutiveHomePage_Query` to use classification instead of publishing group as filter.

## 🧪 Testing

1. `pnpm build:fresh`
2. Navigate to http://localhost:8000/en/executive
3. Verify all pools returned have `classification.group` equal to `EX`


esizer (commented):
This works well and I think the approach makes a lot of sense. However, this could lead to some confusion if we still have the EX publishing group 🤔 

I wonder if we could also remove that since it seems to be doing nothing now and users might think it still functions as it used to.

mnigh:
> This works well and I think the approach makes a lot of sense. However, this could lead to some confusion if we still have the EX publishing group 🤔
> 
> I wonder if we could also remove that since it seems to be doing nothing now and users might think it still functions as it used to.

I would guess there is already confusion as there are three ways a pool could be an executive job (though of course only one is used for having a pool render on the executive home page):

1. Executive Jobs publishing group
2. EX group as part of a classification
3. Executive Group work stream

The entire `PublishingGroup` enum should be removed in the near future which would make this obsolete.

esizer (approved):
LGTM! 👍

PR #17158, opened 2026-06-19
[Task] Claude Code configuration

tristan-orourke:
## 👋 Introduction

Adds some shared configuration for Claude Code, and any AI agent.

## 🕵️ Details

`AGENTS.md` is a standard file which is read by most AI agents (eg Copilot) at the beginning of any interaction. Claude Code will read `.claude/CLAUDE.md`, which current just references `AGENTS.md` to avoid duplication. 

> [!TIP]
> In conversation with Claude or in any of its markdown config files, prefixing a file name with @ (eg `@folder/file.txt`) will cause it to immediately read that file in full.

`.claude/settings.json` includes a whitelist for some bash commands which I've needed often and, as read-only commands, should be very safe. There are other settings we could share here if we want. If you want to configure your own local settings, you can add them to the repo at `.claude/settings.local.json` or in your user-level claude settings, usually at `$HOME/.claude/settings.json`.

I've also in included a `write-ticket` [skill](https://code.claude.com/docs/en/skills). Skills are essentially files Claude reads as additional context/instruction. You can invoke this manually as a backslash command `/write-ticket`, but I've also added a description in the file frontmatter which _should_ cause it to automatically invoke it if you tell claude to something related to github issues or tickets.

## 🧪 Testing

1. Restart claude in the project directory
2. ask it what's in its CLAUDE.md file
3. Try the `write-ticket` command: `/write-ticket <describe a ticket or bug>`


petertgiles:
This looks cool!  I'm not sure about the commands part, though.  I have PHP installed on my machine and so I don't use the maintenance container.  It's faster, simpler for the AI, lets me click on file links directly.  :laughing:   Can the makefile instructions be configurable somehow?

tristan-orourke:
> This looks cool! I'm not sure about the commands part, though. I have PHP installed on my machine and so I don't use the maintenance container. It's faster, simpler for the AI, lets me click on file links directly. 😆 Can the makefile instructions be configurable somehow?

Hmm. Maybe I just revert e7ccd0233e85391f2e1c805ba67bc7909ee494ff !

I don't know one way or the other about configuring makefile commands. But we could probably add notes about multiple setup configurations in the AGENTS file. Something like: "If user is running dev environment in docker container, see this section. If running directly on linux, see this section. If running on a mac, ..." etc

petertgiles:
Can we easily turn skills on and off?  Something like "enable the maintenance container skill if you use that..."?

esizer:
Okay, so somethings I have had to instruct claude on that I did not notice here and could be helpful:

 - Not allowing it to provide French translations (or any copy for that matter). Mine would always add them and I do not trust them to follow our internal lexicon and translations would seem to be done but they were poor or inaccruate translations.
 - Claude also loved to put conversation meta in the comments about things we discussed (it loves commenting code). I told it to keep all the meta discussion out of comments and prefer terse comments that describe obscure or more magical code. This is pretty subjective though so might not be worth adding at all
 - We might want to reassess assigning specific version numbers in the file. This sounds helpful but might go stale? (I think it already is, arent we on laravel 13? :sweat_smile: )
 - The write ticket referenses a memory, I think those are local or account specific so we might want to replace that with the actual memory


Other than that, this seems like a really solid baseline we can add on to as we start using it more frequently :+1:

petertgiles:
I really like the AI guide.  Very clear and common sense.  :clap:

tristan-orourke:
I think this is ready to merge in now, if I can get a couple approvals. This files should be living documents anyway.

esizer (approved):

petertgiles (approved):
Seems good!

Out of curiosity, what kind of things would you put in your .claude/settings.local.json file?

tristan-orourke:
Good question Peter. 

Right now the settings file just has allowed commands, which if we stay conservative with our choices (ie read-only commands) are probably good to share. But you might have other allow/deny commands to put in your own settings.

Skimming through the [settings docs](https://code.claude.com/docs/en/settings#settings-files)...

You might have preferred defaults, like a default model or effort level. Turning on voice dictation. Visual theme. Caching behaviour.

PR #17163, opened 2026-06-19
[Bug] Use client credentials for ping

### LINKED ISSUE #17159: 🐛 CL Logging Noise.

The CL team has complained that we're creating a lot of noise for them with our failed requests in the auth ping:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/api/app/Console/Commands/AuthPing.php

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

Andrew Smith said this:

>  Let me check with the IBM Team supporting CanadaLogin to see if they have a better way to confirm if the endpoint is up and running. To the best of my knowledge, there is no healthcheck API endpoint but what might work better is to actually use the client_credentials grant type in the request to the token endpoint. If we set up API access without granting any permissions, you should still be able to get an access token but that token wouldn't be good for anything. That would both show that the endpoint is up and functioning. 

```php
$url = config('oauth.token_uri'); 

// Updated payload for machine-to-machine Client Credentials flow 
$data = [ 
 'grant_type' => 'client_credentials', 
 'client_id' => config('oauth.client_id'), 
 'client_secret' => config('oauth.client_secret'), 

$results = $this→pingTokenUri($url, $data);
```

### ✅ Acceptance Criteria

- [ ] AuthPing doesn't generate errors when it runs

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

petertgiles:
🤖 Resolves #17159 

## 👋 Introduction

Updates the auth ping command with the client credentials flow provided by the CanadaLogin team.

## 🕵️ Details

It only sort-of works with Sign In Canada.  We still get timing but the response is unsuccessful.  We should be live with CL by the time this deploys so I'm not too worried.

## 🧪 Testing

1. Configure your environment for CanadaLogin.
2. `./artisan auth:ping`

## 📸 Screenshot

image

### DIFF
--- a/api/app/Console/Commands/AuthPing.php
+++ b/api/app/Console/Commands/AuthPing.php
@@ -31,13 +31,11 @@ class AuthPing extends Command
     public function handle()
     {
         $url = config('oauth.token_uri');
-        // typical payload for requesting a login in the auth callback
+        // client credentials payload for OIDC token endpoint
         $data = [
-            'grant_type' => 'authorization_code',
+            'grant_type' => 'client_credentials',
             'client_id' => config('oauth.client_id'),
             'client_secret' => config('oauth.client_secret'),
-            'redirect_uri' => config('oauth.redirect_uri'),
-            'code' => '1', // not a real code, of course
         ];
 
         $results = $this->pingTokenUri($url, $data);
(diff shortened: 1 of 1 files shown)


tristan-orourke:
Okay, works for me.
What are the consequences for having this deployed without CL being turned on?

petertgiles:
> What are the consequences for having this deployed without CL being turned on?

Good question.  The success % will drop to 0 but we'll still get timing metrics from SIC.  We're not firing any alerts or anything from this so 0% success won't affect anything.

tristan-orourke (approved):

PR #17206, opened 2026-06-24
[Chore] Remove too many candidates reason

### LINKED ISSUE #17205: ♻️ Remove "Too many candidates" from tracked users
### ♻️ Refactor

We have a reason for not referring a user that statess "Too many candidates". We do not want it so, we should remove it.

### 🕵️ Details

This is not live so should be safe to remove without need to resolve any database concerns.

### 🙋‍♀️ Proposed Solution

Remove that enum case and its associated translations.

### ✅ Acceptance Criteria

- [ ] Too many candidates is not longer an option/reason
- [ ] No references or translations remain

esizer:
🤖 Resolves #17205 

## 👋 Introduction

This removes the "Too many candidates" reason for a user not being referred.

## 🧪 Testing

1. Confirm it is no longer present
2. Confirm submitting other reasons still works.


yonikid15 (approved):
🪓

PR #17207, opened 2026-06-24
[Bug] Fix "Job placement = Not placed" filter excluding reverted candidates

### LINKED ISSUE #17071: 🐛 Filtering issue: "Job placement = Not placed" returns no results and filter counter is inconsistent

This issue was identified while supporting a recruitment team investigating why certain candidates were not appearing in the results of a Talent Request.
The candidates in question meet all of the following criteria:

Application status: Qualified
Job placement: Not placed
Referral status: Available for referral
Interested in job offers: Open to job offers

However, these candidates were not appearing in Talent Request results when expected.

### 🕵️ Details

Candidates who appear as "Not placed" in the results table are excluded when the filter Job placement = Not placed is applied.

This creates a clear inconsistency between the values displayed in the table and the filtering results.

**Impact**

- Recruiters may incorrectly assume that no candidates meet their criteria
- Valid candidates are excluded from Talent Request results
- This affects hiring workflows and trust in the filtering system

### 📋 Steps to Reproduce

- Navigate 26-CEO-EA-DT-30599 process candidate list

- Apply the following filters:

      - **Application status:** _Qualified_
      - **Referral status:** _Available for referral_
      - **Interested in job offers:** _Open to job offers_

   Result: 7 candidates are returned

- Observe in the results table: All 7 candidates show "_Not placed_" under the **"Job placement"** column

- Add the filter: **Job placement:** _Not placed_

  Result: 0 candidates

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

Applying Job placement = Not placed should return candidates who are shown as "Not placed" in the results table.

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

tristan-orourke:
Closes #17071

## Summary

Fixes the "Job placement = Not placed" filter excluding reverted candidates by implementing the intended placement lifecycle: placement fields are **subfields of the QUALIFIED status**, only meaningful when a candidate is qualified (similar to how `removal_reason` is a subfield of REMOVED and `screening_stage` is a subfield of TO_ASSESS).

Ref: [Application Status Flowchart (Figma)](https://www.figma.com/board/FjnhpNxpCicPA3tejTTar9/Application-Status---Flowchart?node-id=0-1&p=f&t=AqAqtpPxfVaoyw78-0)

### What changed

**Backend lifecycle** — `placement_type` is now managed as part of status transitions:
- `qualify()` initializes `placement_type = NOT_PLACED`
- `revertFinalDecision()` nullifies all placement fields (candidate is leaving QUALIFIED)
- `revertPlacement()` sets `placement_type = NOT_PLACED` (returns to qualified-but-not-placed)

**Query scope** — `wherePlacementTypeIn()` uses a simple `whereIn` with no special NULL handling. NULL means "not qualified" and should not match any placement filter.

**Data migration** — Backfills existing data to match the intended lifecycle:
- Qualified candidates with NULL `placement_type` → `NOT_PLACED`
- Non-qualified candidates with stale placement data → NULL
- Column default changed from `NOT_PLACED` to `NULL`

**Frontend** — Removed the "Not placed" fallback for null `placement_type`. Non-qualified candidates now show an empty cell, matching how `screeningStage` displays for non-TO_ASSESS candidates.

**CanRemoveCandidate validation** -- Also updated validation for CanRemoveCandidate at the same time. This is tangential to the original ticket, but while we're on the topic of matching the intended state transitions more closely, validation was updated so candidates can only be "removed" from the "In assessment" state, not if they are qualified or disqualified or anything else.

## Root cause

The original implementation diverged from the intended design:
1. The DB defaulted `placement_type` to `NOT_PLACED`, so new applications started with a placement value before being qualified
2. Status transitions (`qualify`, `revertFinalDecision`) didn't manage placement fields
3. `revertPlacement()` set `placement_type = null` instead of `NOT_PLACED`

This created a mix of NULL and NOT_PLACED values with no consistent meaning, which broke the filter.

## Test plan

- [x] `PoolCandidateUpdateTest` passes (19 tests, 145 assertions)
- [x] `qualify()` sets `placement_type = NOT_PLACED`
- [x] `revertFinalDecision()` nullifies `placementType`
- [x] `revertPlacement()` sets `placement_type = NOT_PLACED`
- [ ] Run migration on dev, verify qualified candidates have NOT_PLACED and non-qualified have NULL
- [ ] Verify "Not placed" filter returns only qualified candidates with NOT_PLACED
- [ ] Place a candidate, revert placement, verify NOT_PLACED in DB
- [ ] Revert qualification, verify placement_type is NULL in DB

🤖 Generated with [Claude Code](https://claude.com/claude-code)


esizer:
We might need to loop @Jerryescandon in. iirc, it was an intentional decision to set this to `null` instead of `NOT_PLACED. I just can't remeber why :sweat_smile: 

swappy-20260625_082443

REF: https://www.figma.com/board/FjnhpNxpCicPA3tejTTar9/Application-Status---Flowchart?node-id=0-1&p=f&t=AqAqtpPxfVaoyw78-0

Jerryescandon:
> We might need to loop @Jerryescandon in. iirc, it was an intentional decision to set this to `null` instead of `NOT_PLACED. I just can't remeber why 😅
> 

I am not fully understanding the issue here. But reverting a candidate from `Qualified ` to `ToAssess ` should definitely mean that the placement should be set to `null`.

Otherwise we would have `ToAssess` candidates showing up when using the placement filter to find `Not placed` candidates.

Hit me with a huddle tho, im happy to chat more about this.

the `Not placed` attribute is exclusive to `Qualified` candidates

tristan-orourke:
@esizer Okay looking at that state diagram in Figma, I think I see what we were thinking. It looks like we thought of the placement fields as only being meaning for Qualified candidates. The placement fields start out as null, are initialized as NOT_PLACED _only_ when a candidate is qualified, and are nullified again if you revert qualification.

The problem is, that's not what we implemented :sweat_smile:. 
1. We have `placement_type` defaulting to NOT_PLACED in the database, so new applications start out with `placement=NOT_PLACED`, _not_ `placement=NULL`.
2. We don't touch placement fields at all when qualifying or reverting qualification on a candidate.
3. Instead, we set `placement_type=NULL` when reverting a _placement_ decision, instead of setting it back to NOT_PLACED. This seems like a bug, and is fixed by this PR.

tristan-orourke:
@Jerryescandon I'll hit you with that huddle later to discuss the original intent more.

esizer:
> @esizer Okay looking at that state diagram in Figma, I think I see what we were thinking. It looks like we thought of the placement fields as only being meaning for Qualified candidates. The placement fields start out as null, are initialized as NOT_PLACED _only_ when a candidate is qualified, and are nullified again if you revert qualification.
> 
> The problem is, that's not what we implemented 😅.
> 
> 1. We have `placement_type` defaulting to NOT_PLACED in the database, so new applications start out with `placement=NOT_PLACED`, _not_ `placement=NULL`.
> 2. We don't touch placement fields at all when qualifying or reverting qualification on a candidate.
> 3. Instead, we set `placement_type=NULL` when reverting a _placement_ decision, instead of setting it back to NOT_PLACED. This seems like a bug, and is fixed by this PR.

Yeah, and that is why we have the bug. I was just saying that the fix might not be setting the value to `NOT_PLACED` and instead look at the default value and scopes instead. The issue is, `NOT_PLACED` is a decision the assessor makes so it seems weird reverting a decsion would be the same as making an explicit decision to not place someone.

tristan-orourke:
@Jerryescandon @esizer, I made it match the original data model plan!

esizer:
Code seems good, do you mind resolving those conflicts before I approve though? I think they directly impact this fix.

esizer (changes requested):
okay, I noticed a few things on a more thorough review after those merge conflicts.

1. the `remove()` method does not `null` out `placement_type`. The figjam doesn't show that but, I think it likely makes sense to do.
2. Do we think adding a regression test for this makes sense? We can likely get away with a PHPUnit test (I think)

tristan-orourke:
> the remove() method does not null out placement_type. The figjam doesn't show that but, I think it likely makes sense to do.

I'm not sure. These state-transition methods seem to assume that we're always going through a legal transition. We could make them more robust, but then I would go all in:
- nullify all placement fields in submit(), disqualify() and remove()
- nullify `removal_reason` in submit(), disqualify(), and qualify(), and place()
- nullify `disqualified_reason` in submit(), qualify(), remove() and place()
- nullify `expiry_date` in submit(), disqualify(), and remove()

Its more robust to mistaken transitions, but the downside would be some added noise in all these methods. What do you think @esizer?

esizer:
> > the remove() method does not null out placement_type. The figjam doesn't show that but, I think it likely makes sense to do.
> 
> I'm not sure. These state-transition methods seem to assume that we're always going through a legal transition. We could make them more robust, but then I would go all in:
> 
> * nullify all placement fields in submit(), disqualify() and remove()
> * nullify `removal_reason` in submit(), disqualify(), and qualify(), and place()
> * nullify `disqualified_reason` in submit(), qualify(), remove() and place()
> * nullify `expiry_date` in submit(), disqualify(), and remove()
> 
> Its more robust to mistaken transitions, but the downside would be some added noise in all these methods. What do you think @esizer?

I'm a little confused. I'm not sure what is being asked here. It's just that, `placement_type` only makes sense if your status is qualfied. So as it stands, if you are qualified thebn removed, you still appear when the filter is set to not placed since that can still be in the database right? I'm not sure about the others right now, I was just looking at the issue with the not placed filter.

tristan-orourke:
Okay, I see the confusion now :smile: 

The opposite of qualify() isn't remove(), its revertFinalDecision(), which does nullify the placement fields.

remove() is a state similar to disqualified, and it can only be set when the candidate is still IN ASSESSMENT, when the placement fields will already be null.

image

esizer:
> Okay, I see the confusion now 😄
> 
> The opposite of qualify() isn't remove(), its revertFinalDecision(), which does nullify the placement fields.
> 
> remove() is a state similar to disqualified, and it can only be set when the candidate is still IN ASSESSMENT, when the placement fields will already be null.

Okay, the UI limits it but we should probably update the validator because it does allow this operation to be run right now.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/ec5178a8493df1326a8f9ca43fd18507e4a5ce6a/api/app/Rules/CanRemoveCandidate.php#L39-L42

tristan-orourke:
Thanks, good catch @esizer.

Jerryescandon:
Got your crash course into application statuses after all 🤣

esizer (approved):
LGTM! 👍

PR #17208, opened 2026-06-24
[Feature] Allow special applications to bypass some validation

### LINKED ISSUE #17197: ✨ Allow submitting special applications
### Details

Complete the likely remaining backend work to allow submitting a special application. This will involve bypassing some parts of validation

### Design file & designer

Design file & designer

@Jerryescandon
https://www.figma.com/design/vZEg5z6aXiJPNB07LamrgY/Admin---Users?node-id=1165-30703&t=vBD6Tz5Mn3ojfa0N-1

### Proposed Implementation

Bypass validation conditionally if special application fields are set
Some validation changes if those fields are set, like which closing date is relevant 

#### Validation 

Allow submitting special applications thru normal submission process

### ✅ Acceptance Criteria

- [ ] Special applications can be submitted like other applications
- [ ] Validation bypassed or is pointed at other field(s) as appropriate. Validation may need to be modified when:
  - [ ] creating an application
  - [ ] updating an application
  - [ ] submitting an application
  - [ ] deleting an application (User should still be able to delete an unsubmitted draft, if they really want)
- [ ] Add some testing
- [ ] Use cases that must be possible in the backend:
    - [ ] Veterans are allowed to apply to internal processes (for 2 or 5 years) after they leave the forces
    - [ ] Ministers staff are allowed to apply to internal processes (for a year or two) after they leave that job
    - [ ] Priorities are allowed to come forward at any time and join a process even after the closing date
    - [ ] The PSC could determine that someone should have been allowed to apply after an investigation, but the poster is now closed
    - [ ] Employees on leave who don't have access to their email should still be allowed to apply to internal jobs (maternity leave for example)

vd1992:
🤖 Resolves #17197

## 👋 Introduction

Enable special applications to interact with validation differently provided the special application check is true
- Ignore verified work email check
- Ignore normal closing date in favour of special closing date

## 🧪 Testing

1. You can apply to a closed poster with the two conditions we are allowing

- no verified work email
- pool is closed


api/app/Policies/PoolCandidatePolicy.php
```diff
@@ -173,7 +173,10 @@ public function count()
      */
     public function delete(User $user, PoolCandidate $poolCandidate)
     {
-        return $user->id === $poolCandidate->user_id && $user->isAbleTo('delete-own-draftApplication');
+        return
+            is_null($poolCandidate->submitted_at) &&
```
vd1992: I realized after that we do have validation preventing deletion of submitted applications making this change not important. The validator is called in a class rather than the schema so I missed it initially

api/tests/Feature/SpecialApplicationTest.php
```diff
@@ -160,4 +162,78 @@ public function testSpecialClosingDateNotBeforePoolClose(): void
                 'The pool candidate.special application closing date field must be a date after '.$this->pool->closing_date.'.'
             );
     }
+
+    // test accessor PoolCandidate::isSpecialApplication()
+    public function testIsSpecialApplication(): void
+    {
```
vd1992, line 168: The accessor is simple so this is less about testing it works and more about setting something up to trip if the accessor is adjusted

api/tests/Feature/SpecialApplicationTest.php
```diff
@@ -160,4 +162,78 @@ public function testSpecialClosingDateNotBeforePoolClose(): void
...
+            $setCandidate->is_special_application,
+            true,
+        );
+    }
+
+    // a special application can bypass standard validation when submitting
+    public function testSpecialApplicationBypassesNormalValidation(): void
+    {
+        // pool is internal and closed
+        $this->pool->area_of_selection === PoolAreaOfSelection::EMPLOYEES->name;
```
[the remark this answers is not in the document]
vd1992: TypeScript and its linting would've been on this unlike PHP...
https://github.com/GCTC-NTGC/gc-digital-talent/pull/17208/changes/3635719d94d2434d48c9d32d1715de135e35d21c

api/app/Policies/PoolCandidatePolicy.php
```diff
@@ -173,7 +173,10 @@ public function count()
...
+            $user->id === $poolCandidate->user_id &&
+            $user->isAbleTo('delete-own-draftApplication');
```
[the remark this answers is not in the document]
vd1992: https://github.com/GCTC-NTGC/gc-digital-talent/pull/17208/changes/3635719d94d2434d48c9d32d1715de135e35d21c

api/graphql/schema.graphql
```diff
@@ -734,6 +734,7 @@ type PoolCandidate {
     @canRoot(ability: "viewAssessment")
   specialApplicationClosingDate: DateTime
     @rename(attribute: "special_application_closing_date")
+  isSpecialApplication: Boolean
```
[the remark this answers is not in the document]
vd1992: Not true
Field `specialApplicationClosingDate` is also not specially guarded
Allowed to see if allowed to see the base pool candidate

api/app/Rules/EmployeeWorkEmailVerified.php
```diff
@@ -11,23 +11,32 @@
...
+        $this->isSpecialApplication = $isSpecialApplication;
     }
 
     /**
      * Determine if the validation rule passes.
      */
     public function validate(string $attribute, mixed $value, Closure $fail): void
     {
+        // do not run this check if this is a special application
+        if ($this->isSpecialApplication === true) {
```
petertgiles: This works but it seems strange to run a validation rule with a killswitch.  I think it would make more sense to conditionally run the rule in the calling validator.

api/app/GraphQL/Validators/Mutation/SubmitApplicationValidator.php
```diff
@@ -44,8 +53,8 @@ public function rules(): array
                 new HasLanguageRequirements($this->application->pool),
             ],
             'pool_id' => [
-                new EmployeeWorkEmailVerified($this->application->user),
-                new PoolNotClosed(),
+                new EmployeeWorkEmailVerified($this->application->user, $this->isSpecialApplication),
+                new PoolNotClosed($this->isSpecialApplication, $this->specialClosingDate),
```
petertgiles: Rather than increasing the scope of this rule, I suggest encompassing the new special logic in its own rule and OR'ing them together with a `Rule::anyOf()`.

api/app/Rules/EmployeeWorkEmailVerified.php
petertgiles: Perhaps something like `Rule::when()`

api/app/Models/PoolCandidate.php
```diff
@@ -92,6 +93,7 @@
  * @property ?string $special_application_type
  * @property ?string $special_application_justification
  * @property ?Carbon $special_application_closing_date
+ * @property bool $is_special_application
```
petertgiles, line 96: I don't think this should be here if you have it declared as an attribute explicitly below.

api/app/Policies/PoolCandidatePolicy.php
```diff
@@ -173,7 +173,10 @@ public function count()
      */
     public function delete(User $user, PoolCandidate $poolCandidate)
     {
-        return $user->id === $poolCandidate->user_id && $user->isAbleTo('delete-own-draftApplication');
+        return
+            $poolCandidate->isDraft() &&
```
petertgiles, line 177: Good catch!

petertgiles (changes requested):
I just finished a code readthrough.  Could you please check these suggestions?

api/app/Rules/EmployeeWorkEmailVerified.php
vd1992: Yeah, I agree. Shouldn't run the Rule if it isn't actually going to be evaluated 
https://github.com/GCTC-NTGC/gc-digital-talent/pull/17208/changes/526d10127f85dda7234afc7163ea965a00717d0b

api/app/Models/PoolCandidate.php
vd1992: It complains if not present 🤷 

image

petertgiles: Oh, I see, Lighthouse would access the pascal case attribute but other PHP code needs the snake case property.  You're right, nvm.

api/app/GraphQL/Validators/Mutation/SubmitApplicationValidator.php
vd1992: I'm not sure about this one, should they be spread out? It does look simpler this way in one file

petertgiles: The reason I suggested this is because the name of the rule is no loner accurate.  It's not just checking if the pool is closed.  So, do we rename it something like PoolNotClosedOrAllowedToApplyAnyway?  If the name of your function has an AND or an OR in it then that's a strong signal that it's doing too much.  If you really don't want to separate it, I think we need to refactor the rule a bit more to make it clearer what it's actually doing.

vd1992: Hm... Yeah, it could be made clearer, so given the rule is only used for the one validation case I elected to rename it to fit what role it serves. It it used to assert a pool is open for application and no other uses

https://github.com/GCTC-NTGC/gc-digital-talent/pull/17208/changes/c66a1f618e2c5cf836204b785d0d1db7b5f2800b

image

petertgiles: Oh, I like that!  Having a negated operation in the name was awkward anyway.

petertgiles (approved):
Works great:
✔️ Can apply to closed pool when special
✔️ Can't apply to closed pool when special expired
✔️ Can apply to employee pool as public when special

PR #17223, opened 2026-06-25
[Bug] Issue 17204

### LINKED ISSUE #17204: 🐛 Issue with updateApplication mutation

Details in CVD-RFI-020 report provided by Hackerone.
Ask @tristan-orourke for details.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] issue described in report no longer present

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

vd1992:
🤖 Resolves #17204 

## 👋 Introduction

Resolve the issue

## 🧪 Testing

1. ...
2. ...


api/app/GraphQL/Validators/Mutation/UpdateApplicationValidator.php
```diff
@@ -0,0 +1,93 @@
...
+     * Return the validation rules.
+     *
+     * @return array<string, array<mixed>>
+     */
+    public function rules(): array
+    {
+        $applicationId = $this->arg('id');
+
+        $poolCandidate = PoolCandidate::query()->select('user_id')->findOrFail($applicationId);
+
```
vd1992: Been trying to be more mindful of performance, like in this case trying to keep validation lightweight by being very deliberate in model/field fetching. No more, no less than what is needed

esizer: Hmm, I see a small issue with this if we want to optimize it. Right now it runs those queries regardlesss if someone is attempting to sync any of the experience types. I think we chould use the `exists` rule.

Soemthing like:

```php
$userId = PoolCandidate::query()
  ->select('user_id')
  ->findOrFail($this->arg('id'))
  ->user_id;

'application.educationRequirementAwardExperiences.sync.*' => [
  Rule::exists('award_experiences', 'id')->where('user_id', $userId),
],
```

This way, the query only runs if the user supplies one of the experiences sync. Of less concern but also true, this means we no longer need to pull all of the users experience IDs in.


SEE: https://laravel.com/docs/13.x/validation#specifying-a-custom-column-name

esizer (changes requested):
I think we should also write a regression test for this.

api/app/GraphQL/Validators/Mutation/UpdateApplicationValidator.php
vd1992: Yeah I actually was thinking about how to make that bit better but wasn't sure how to do so best. Thanks for the suggestion, it fits nice and neat
https://github.com/GCTC-NTGC/gc-digital-talent/pull/17223/changes/8e3a289cde676895fb7f3f92bc9055dbec8647a2
https://github.com/GCTC-NTGC/gc-digital-talent/pull/17223/changes/aa1b3ced3e6ef2d862786077a551dc4c332d7db1

esizer (approved):
LGTM! 👍

PR #17237, opened 2026-06-26
[Feature] Add `talentSources` field to `ApplicantFilterInput`

### LINKED ISSUE #17184: ✨ Add `talentSources` field to `ApplicantFilterInput`
### Details

The Talent Request feature needs the ability to filter matched users by talent source. This ticket adds the `talentSources` field to `ApplicantFilterInput` using the existing `TalentRequestSource` enum, along with a database migration.

- Add `talentSources: [TalentRequestSource]` to `ApplicantFilterInput` in `api/graphql/schema.graphql` (line ~1487)
- Create a database migration for the new column. Existing rows default to `null`
- Existing queries like `countPoolCandidatesByPool` can ignore the new field
- Talent request queries (`talentRequestMatches`, `countTalentRequestMatches`, `countTalentRequestMatchesByPool`) should accept the new field
  - If `talentSources` is null or empty, assume all possible options are selected (no filtering applied)

> [!NOTE]
> There is currently only one valid `TalentRequestSource` value (`QUALIFIED_IN_POOL`), so query modifications may be minimal at this stage. The filtering logic becomes meaningful once `AT_LEVEL` is added in #17185.

Related: #17185, #17186, #17187

### Design file & designer

_No response_

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

- Add the field to `ApplicantFilterInput` in `api/graphql/schema.graphql`
- The `TalentRequestSource` enum already exists at `api/app/Enums/TalentRequestSource.php`
- Add migration for the new column on the relevant table
- Wire the field into the talent request query scopes/resolvers (see schema.graphql lines ~1753–1766)

### ✅ Acceptance Criteria

- [ ] `talentSources: [TalentRequestSource]` field added to `ApplicantFilterInput`
- [ ] Migration created and can be rolled back; existing data defaults to null
- [ ] Existing queries (e.g. `countPoolCandidatesByPool`) are unaffected
- [ ] Talent request queries accept the field and treat null/empty as "all sources"

yonikid15:
🤖 Resolves #17184 

## 👋 Introduction

- Adds a migration which adds the `talent_sources` column to the `applicant_filters` table.
- Adds the `talentSources: [TalentRequestSource]` field added to `ApplicantFilterInput` 
- Adds filtering by AT_LEVEL source

## 🧪 Testing

1. Run `php artisan migrate` to apply the new column.
2. Manually send a `talentRequestMatches` query via GraphQL Playground with `applicantFilter: { talentSources: [AT_LEVEL] }` and confirm only qualified-at-level sources are returned.
3. Manually send a `talentRequestMatches` query with `applicantFilter: { talentSources: [QUALIFIED_IN_POOL] }` and confirm only qualified-in-pool sources are returned.
4. Send the same query with `talentSources: []` or omitting the field, and confirm both AT_LEVEL and QUALIFIED_IN_POOL sources appear in results.


api/app/Builders/UserBuilder.php
```diff
@@ -291,8 +291,19 @@ public function whereQualifiedInClassificationsIn(?array $classifications): self
...
+            })->orWhereHas('currentClassification', function ($query) use ($classifications) {
+                $query->where(function ($q) use ($classifications) {
+                    foreach ($classifications as $classification) {
+                        $q->orWhere(function ($q) use ($classification) {
+                            $q->where('group', $classification['group'])
+                                ->where('level', $classification['level']);
+                        });
+                    }
+                });
+            });
```
esizer: I think this query is still used by the `usersPaginated` query. We might need to back this change out.

esizer (changes requested):
I noticed a few things here:

1. The changes to `whereQualifiedInClassificationsIn` should be bacjked out since it is used in other queries
2. We would need to update the new count queries to use the input `TalentRequestMatchFilterInput` instead

Also, we might want to add some tests for when the user passes in the new sources and stuff

api/app/Builders/UserBuilder.php
yonikid15: [Fix count query input type and revert classification scope expansion](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17237/commits/0d32fc574e6368c45b2b0268450afc2b85b52585)

- Change `countTalentRequestMatches` and `countTalentRequestMatchesByPool`
  to accept `TalentRequestMatchFilterInput` instead of `ApplicantFilterInput`,
  making them consistent with `talentRequestMatches`
- Update `PoolCandidateBuilder::whereMatchesTalentRequest` to extract
  applicantFilter from the input shape, matching UserBuilder's behaviour
- Revert `UserBuilder::whereQualifiedInClassificationsIn` to pool-candidates
  only — the orWhereHas(currentClassification) expansion incorrectly
  broadened the usersPaginated classification filter
- Remove `testAtLevelClassificationFilterNarrowsResults` which relied on
  the reverted expansion
- Add `talentSources` filter tests: AT_LEVEL only, QUALIFIED_IN_POOL only,
  and both sources together

api/tests/Feature/TalentRequestMatchesTest.php
```diff
@@ -589,4 +590,137 @@ public function testCountAgreesWithTheListTotal(): void
...
+                    data {
+                        user { id }
+                        sources { value }
+                        matchingAtLevelSources { id }
+                    }
+                    paginatorInfo { total }
+                }
+            }
+            GRAPHQL;
+    }
```
esizer: nit: this could be a class property instead of a method sincve its only returning a string.

apps/web/src/pages/SearchRequests/SearchPage/hooks.ts
```diff
@@ -86,6 +86,11 @@ export const useCandidateCount = (
     [filters],
   );
 
+  const talentRequestQueryArgs = useMemo(
+    () => ({ where: { applicantFilter: queryArgs.where } }),
+    [queryArgs.where],
+  );
```
esizer: Is the useMemo necessary here? I don't think we need this and it's likely more expensive than just passing the args directly to the query.

esizer (changes requested):

apps/web/src/pages/SearchRequests/SearchPage/hooks.ts
yonikid15: We were doing it previously for the pool candidate search request query, so I just mimicked it. I'll take a closer to look to see if we can remove it.

api/tests/Feature/TalentRequestMatchesTest.php
yonikid15: [Change to string property](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17237/commits/3d8d74915b92aa0ae8c262a84be304c2fffe80b0)

apps/web/src/pages/SearchRequests/SearchPage/hooks.ts
yonikid15: [Remove useMemo](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17237/commits/d395adbbb507c9b7fec66ff04af9753ac6056cff)

esizer (approved):

PR #17251, opened 2026-06-30
[Refactor] Reenable permission gate

### LINKED ISSUE #17118: ♻️ Reenable permission gate
### ♻️ Refactor

In https://github.com/GCTC-NTGC/gc-digital-talent/pull/17046 we had to put in a workaround to skip the security gate for using privileged permissions on a non-privileged page.  Let's go back and fix that.

- api/app/Rules/TalentEventOpenForCreatingNominations.php
- api/app/Rules/TalentEventOpenForUpdatingNominations.php

> In [#16995](https://github.com/GCTC-NTGC/gc-digital-talent/issues/16995) we're supposed to let certain community roles create new nominations after the closing date of the event.  Problem is, the URL for creating nominations is not under /admin so the elevated permission check is failing. This is as intended, to make sure that privileged activity only happens on the secure network.  The URLs are in the form /en/communities/talent-events/:event_id/create-talent-nomination.  We'll have the same problem filling in the nomination.  I'm not sure what the best solution is though.  We've never had this type of hybrid privileged/not privileged page before, I think.

### 🕵️ Details

_No response_

### 🙋‍♀️ Proposed Solution

It seems reasonable to me to create a duplicate set of /create-talent-nomination  routes under the admin talent-event page.

### ✅ Acceptance Criteria

- [ ] Permission check uses the regular `$user->isAbleTo('some-permission')` call.

petertgiles:
🤖 Resolves #17118 

## 👋 Introduction

Switches back to the regular permission check for creating/updating/submitting nominations.

## 🕵️ Details

Adds the ability for individual queries to request usage of the protected endpoint and uses this on the create/update/submit nomination pages.

A lot of the diff of this PR is pulling down the event close date on every page so it knows which context to submit mutations in.

## 🧪 Testing

I recommend testing the nomination flow in three contexts:

1. A regular employee can complete the flow "off network", submiting to /graphql.
2. An admin can complete the flow "on network", submitting to /admin/graphql.
3. An admin gets understandable errors "off network", submitting to /amdin/graphql.

To simulate being "off network", you can adjust the nginix_local.conf file like this:
image

Don't forget to restart you webserver container after adjustments.


tristan-orourke:
So, the backend still requires the mutation to happen via the admin api endpoint, but the frontend can now force itself to route through the admin endpoint? Makes sense to me.

apps/web/src/pages/CreateTalentNominationPage/CreateTalentNominationPage.tsx
```diff
@@ -158,9 +153,58 @@ const CreateTalentNominationPage = () => {
...
+    talentNominationEvent(id: $nominationEventId) {
+      id
+      closeDate
+    }
+  }
+`);
+
+interface RouteParams extends Record<string, string> {
+  id: string;
+}
```
vd1992, line 167: Milord, I beg thy pardon for pointing this out, but the `RouteParams` interface declares `id: string` while the parameter actually read from the route is `nominationEventId` (line 173). The `Record<string, string>` extension hides the mismatch from the compiler. Might it please thee to declare `nominationEventId: string` instead?

I say this with trembling hands and only the deepest respect, praying it does not earn me the stocks.

🤖 *Comment generated by Claude Code*

apps/web/src/pages/CreateTalentNominationPage/CreateTalentNominationPage.tsx
```diff
@@ -32,14 +31,20 @@ const CreateTalentNomination_Mutation = graphql(/* GraphQL */ `
   }
 `);
 
+interface CreateTalentNominationProps {
+  nominationEventId: string;
+  nominationEventCloseDate: string;
```
vd1992, line 36: Milord, a small typing quibble from thy lowliest of subjects: `nominationEventCloseDate` is typed as `string`, yet the schema's `closeDate` may be null (an event without a close date). Since `isPastDateTime` already accepts `string | null | undefined`, widening this prop to match would be more honest. Relatedly, inside the truthy branch on line 182, the optional chains on lines 184–185 (`data?.talentNominationEvent?.id`) are redundant.

I prostrate myself before thy superior judgment and beg thee not to cast me from the realm.

🤖 *Comment generated by Claude Code*

apps/web/src/pages/TalentNominations/NominateTalent/useMutations.ts
```diff
@@ -42,7 +46,11 @@ interface TalentNominationMutations {
 
 type UseMutationsReturn = [boolean, TalentNominationMutations];
 
-const useMutations = (): UseMutationsReturn => {
+interface useMutationsOptions {
```
vd1992: Milord, forgive my impertinence, but the interface name `useMutationsOptions` breaks with the PascalCase convention for types — `UseMutationsOptions` would sit better among its peers. Thou might also consider making the parameter optional (`options?: UseMutationsOptions`) so future callers who need no protected endpoint are not forced to pass `{ forceProtectedEndpoint: false }`.

I remain thy faithful vassal, hoping this trifle does not bring thy wrath upon my humble hovel.

🤖 *Comment generated by Claude Code*

apps/web/src/utils/protectedUrqlContext.ts
```diff
@@ -0,0 +1,18 @@
...
+// Both helpers set the forceProtectedEndpoint flag so protectedEndpointExchange routes
+// the operation to the protected API endpoint, even when the current route is not /admin.
+// Two helpers exist because URQL has two different calling conventions:
+
+// Use when passing context as a positional argument, e.g. executeMutation(variables, getProtectedOperationContext())
+export const getProtectedOperationContext = (): Partial<OperationContext> =>
+  protectedEndpointContext();
+
+// Use when spreading into a useQuery options object, e.g. useQuery({ query, ...getProtectedQueryContextOptions() })
+export const getProtectedQueryContextOptions = (): {
```
vd1992: Milord, if I may be so bold: `getProtectedQueryContextOptions` is exported but, as far as mine unworthy eyes can see, never used anywhere in this PR — the queries all remain on the regular endpoint. Perhaps it should be added only when a query first needs it, lest dead code accumulate in thy fine estate.

I beseech thee, take no offence — I am but a simple peasant who fears the gallows.

🤖 *Comment generated by Claude Code*

packages/i18n/src/messages/apiMessages.ts
```diff
@@ -601,9 +601,20 @@ export const apiMessages: Record<string, MessageDescriptor> = defineMessages({
...
+  defaultMessage: "This page can only be accessed on the secure network.",
+  id: "Gr4YoX",
+  description: "Firewall block",
+});
+
 export const tryFindMessageDescriptor = (
   defaultMessage: string,
 ): MessageDescriptor | null => {
+  // This is the title of the restricted.html page that is triggered by trying to use the /admin pages off-network
+  if (defaultMessage.includes("Unauthorized | Non autorisé")) {
```
vd1992, line 615: Milord, a note of caution from thy watchful servant: matching the firewall block by the substring `"Unauthorized | Non autorisé"` couples this code to the exact title of the `restricted.html` page. Should the firewall page's title ever change, this mapping would silently fall back to the generic error. A shared constant, or at least a test asserting the mapping, would guard against such treachery.

Ever thy devoted subject, who wishes only to serve and never to displease.

🤖 *Comment generated by Claude Code*

api/tests/Feature/TalentNominationUnprotectedTest.php
```diff
@@ -0,0 +1,73 @@
...
+use Nuwave\Lighthouse\Testing\RefreshesSchemaCache;
+use Tests\TestCase;
+
+class TalentNominationUnprotectedTest extends TestCase
+{
+    use MakesGraphQLRequests;
+    use RefreshDatabase;
+    use RefreshesSchemaCache;
+
+    protected $communityTalentCoordinator;
```
vd1992, line 23: Milord, the humblest of observations: `protected $communityTalentCoordinator` is only used within the single test method — a local variable would suffice, as its sibling test in `TalentNominationTest.php` does. A trifle, truly.

I kiss the hem of thy garment and pray this nitpick is received with mercy rather than a summons to the tower.

🤖 *Comment generated by Claude Code*

vd1992 (commented):
## Review (🤖 generated by Claude Code)

Milord, thy humble servant has surveyed the whole of this noble diff and found the estate to be in fine order.

### Overview
- Backend rules (`TalentEventOpenForCreatingNominations`, `TalentEventOpenForUpdatingNominations`) return to the proper Laratrust `isAbleTo` team-scoped permission check instead of the `TeamHelpers` workaround — this is the heart of the refactor and it reads correctly.
- A new `forceProtectedEndpoint` operation-context flag lets individual mutations opt into `/admin/graphql`, used by the nomination create/update/submit flows when the event is already closed.
- `closeDate` is now fetched on every nomination step so each form knows which endpoint to use; `useHasPermissions` in `Instructions.tsx` is now correctly team-scoped, matching the backend check.
- Error handling is improved: the off-network firewall response is mapped to a translated, understandable message.

### Strengths
- The null-safe rewrite in `TalentEventOpenForUpdatingNominations` (`TalentNomination::with(...)->find($value)` + `?->`) removes a potential 500 on a missing nomination and lets `@canFind` produce the proper error. 👍
- `shouldUseProtectedEndpoint` extracted as a pure function with unit tests — good testability.
- Both positive (protected endpoint) and negative (unprotected endpoint) feature tests for the coordinator flow.

### Risks (minor)
- Small TOCTOU window: if an event closes between page load and submit, the mutation goes to the unprotected endpoint and fails with `TALENT_EVENT_IS_CLOSED` — but the improved error mapping surfaces this understandably, so it's acceptable.
- A handful of small nits left as inline comments, if it please the crown.

I remain, as ever, thy most obedient and easily frightened subject, hoping this review finds favour and that no dungeon awaits its author.

apps/web/src/pages/CreateTalentNominationPage/CreateTalentNominationPage.tsx
petertgiles: Yep, good catch.
[nit 1](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17251/commits/578b712a2edd1ffb431c8c14f8aa2915a9260099)

apps/web/src/pages/CreateTalentNominationPage/CreateTalentNominationPage.tsx
petertgiles: I'm confused - I don't think event close date is can be null?
https://github.com/GCTC-NTGC/gc-digital-talent/blob/dd96b0d69933beb5cfd40a4603ec248c12bb0d1f/api/graphql/schema.graphql#L871

It's true that the optional chains are unnecessary, though.
[nit 2](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17251/commits/baeee59236144836e847e2f56bb47ea6d88901e4)

apps/web/src/pages/TalentNominations/NominateTalent/useMutations.ts
petertgiles: Fair enough about the casing.
[nit 3](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17251/commits/5e0433421d42ffe434bd4949fe458ecea7500159)

I want to keep the options mandatory, though.  The function is only for this set of pages.

apps/web/src/utils/protectedUrqlContext.ts
petertgiles: Fair enough.  I guess we can add it back if/when we actually need it.
[nit 4](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17251/commits/c22a6ab255c7346dac42901b207b81afa7dc776d)

packages/i18n/src/messages/apiMessages.ts
petertgiles: Yeah, it's not ideal.  At least it's clearly linked and there's only one place to update.  Hopefully we'll get proper Graphql errors from the F5 some day.  :roll_eyes:

api/tests/Feature/TalentNominationUnprotectedTest.php
petertgiles: I'd rather keep the pattern.

vd1992:
Merge conflicts 👇

apps/web/src/pages/CreateTalentNominationPage/CreateTalentNominationPage.tsx
vd1992: Yeah, not sure about that comment. Field `closeDate` is non-nullable on `TalentNominationEvent` and does not exist on `TalentNomination`
Wonder if it is going on `UpdateTalentNominationEventInput.closeDate` where `closeDate` is optional but that is due to you optionally updating an already set field

In any case, the bot might be hallucinating and fortunately apologized in advance!

apps/web/src/utils/protectedUrqlContext.ts
vd1992, line 1: From changes in Nit 4, you just have the one helper function so "Both helpers..." is misleading 😛

vd1992 (approved):
Off-Network
- Non-admin can interact with nominations for open event only
- Admin can interact with nominations for open events only

On-Network
- Same for non-admin
- Admin can interact with closed events

An aside, couldn't submit a nomination due to 
`NINE_BOX_RATINGS_REQUIRED_FOR_EVENT`
Event seeding issue?

petertgiles:
> An aside, couldn't submit a nomination due to NINE_BOX_RATINGS_REQUIRED_FOR_EVENT Event seeding issue?

Ah, yeah, we don't have UI to add nine box ratings so if a seeded event requires them you won't be able to submit a nomination.  We're building that soon so hopefully it won't be a trap for long.

apps/web/src/utils/protectedUrqlContext.ts
petertgiles: Good catch, thanks.
[update comment after nit 4 fix](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17251/commits/1e3f4b4faae9f5859e62b756c9c1da36a5a85cc2)

Issue #17262, opened 2026-06-30
✨ Create special application dialog


vd1992:
### Details

Create and implement the dialog for special application creation 

### Design file & designer

@Jerryescandon
https://www.figma.com/design/vZEg5z6aXiJPNB07LamrgY/Admin---Users?node-id=1165-30703&t=vBD6Tz5Mn3ojfa0N-1

### 📸 Screenshots

<img width="485" height="900" alt="Image" src="https://github.com/user-attachments/assets/5ec7e5b9-2027-4179-9815-799599ffff5f" />

#### Goes here

<img width="846" height="316" alt="Image" src="https://github.com/user-attachments/assets/ee805e9b-5512-4528-a149-7c7fb46c0ade" />

### ✅ Acceptance Criteria

- [ ] Create dialog
- [ ] Dialogue replaces the current "add user to process" dialogue on the recruitment tab of user profile
- [ ] Only published pools are listed in dialogue
- [ ] This "add user to process" button remains only visible to platform admins
- [ ] Submitting the form in the dialogue triggers a special application

PR #17295, opened 2026-07-02
Add validation rule to special application creation

vd1992:
🤖 Resolves 

## 👋 Introduction

To assist with https://github.com/GCTC-NTGC/gc-digital-talent/issues/17262
Spun off a little work, a validation tweak. Enforce pool is not draft and not archived

## 🧪 Testing

1. Test special application creation

For example mutation

```
mutation A {
  createSpecialApplication(
    poolCandidate: {
      pool: { connect: ":id" }
      user: { connect: ":id" }
      specialApplicationType: ACCOMMODATION
      specialApplicationJustification: "reasons"
      specialApplicationClosingDate: "2027-06-30 00:00:01"
    }
  ) {
    id
  }
}
```


api/app/GraphQL/Validators/CreateSpecialApplicationValidator.php
```diff
@@ -20,7 +23,18 @@ public function rules(): array
         $poolId = $this->arg('poolCandidate.pool.connect');
         $userId = $this->arg('poolCandidate.user.connect');
 
-        $pool = Pool::query()->select('closing_date')->findOrFail($poolId);
+        $pool = Pool::query()->select(['published_at', 'closing_date', 'archived_at'])->findOrFail($poolId);
+
+        if (
+            // block un-published or archived, including blocks of future published_at and allowing future archived_at
```
vd1992: Didn't want to split this off into a rule because 
1. We already have a `PoolPublished` rule, where closed and draft are excluded whereas here closed is fine
2. One off rules classes that are slightly different from another rule makes it confusing to reuse them

petertgiles: Ugh, I really dislike having validation logic happening above the validation array, in the setup section.  I think a rule would be fine, like `PoolOpenForSpecialApplications`.  I think it would be clear how it relates to the existing PoolOpenForApplication rule.  If you really don't want to factor the logic out into a Rule, though, I would like to see the logic at least get moved to the relevant section in the validation array.
image

vd1992: >Ugh, I really dislike having validation logic happening above the validation array

Heh 
Moved it around in https://github.com/GCTC-NTGC/gc-digital-talent/pull/17295/commits/f2c884b041ffd5b07b1d066a11b13014ca6dd359

petertgiles (approved):
Thanks for the update.  Works great.
✔️ draft blocked
✔️ archived blocked
✔️ closed still works

PR #17297, opened 2026-07-02
Change dependabot `rebase-strategy`

vd1992:
🤖 Resolves 

## 👋 Introduction

Set it to disabled
Docs for reference https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#rebase-strategy--


esizer (approved):
Guess we will see how it goes on Thursday! 😆

PR #17301, opened 2026-07-03
[Tooling] Add production pilot slot architecture reference doc

brindasasi:
## Summary

- Adds a self-contained HTML reference document for the Canada Login phased migration via Azure App Service deployment slots
- Intended for sharing with other government organizations replicating this pattern
- Covers slot configuration, post-deployment lifecycle, CL OAuth requirements, shared infrastructure analysis, replication playbook, and the phased migration model (why this is not blue-green)

## Preview

👉 [View rendered document](https://claude.ai/code/artifact/97450006-f7c0-49a0-bba7-a57afc3b7e5d)

## Notes

- File lives in `docs/` for GitHub Pages hosting
- No build step — self-contained HTML with inline CSS and vanilla JS animations
- To publish via GitHub Pages: Settings → Pages → Source: `main` / `/docs`


docs/canada-login-slot-architecture.html
```diff
@@ -0,0 +1,876 @@
...
+      </div>
+
+      <hr class="arch-divider">
+
+      <!-- Two slots -->
+      <div class="arch-slots">
+        <div class="arch-node amber">
+          <div class="arch-node-eyebrow">Pilot Slot · Deployed First</div>
+          <div class="arch-node-name">Admins &amp; Power Users</div>
+          <div class="arch-node-detail amber">talentcloud-pilot.canada.ca</div>
```
petertgiles: I don't think this is the right domain name.

docs/canada-login-slot-architecture.html
petertgiles, line 1: This seems more like a "lessons learned report" than app documentation.  It has more in common with the tc-report than other pieces of documentation we have in the repo at the moment.  If we decide to keep this, maybe we should put it in a "reports" directory?

docs/canada-login-slot-architecture.html
```diff
@@ -0,0 +1,876 @@
+<title>Canada Login Migration — Deployment Slot Architecture</title>
```
petertgiles: This seems like it will be quite difficult to maintain and keep up to date.  Up til now we have been writing documentation using markdown and mermaid (and keeping it in the `documentation` directory), partly because of ease of writing and maintenance.

petertgiles (commented):
Interesting read!  I've added some comments.

docs/canada-login-slot-architecture.html
brindasasi: documentation seems to be the right choice.. will convert as .md file

brindasasi: actually I'm not sure now .. this is meant to share with other teams as needed and also to keep ourselves  a log on what we did and how to replicate next time. 
@tristan-orourke any suggestion ?

tristan-orourke:
Seems like a useful document. I agree with @petertgiles that this is very different from our existing documentation, so probably belongs somewhere else. Its different not just in style but purpose. As you said @brindasasi it seems like a document to share with others (though I'd want to refine it a bit more before we share it outside the team, I think). 

If we want to document for ourselves the process of setting up deployment slots, and a Canada Login integration, then maybe Google Docs is the place for it? (That's where we've documented other infrastructure processes.)

As for where/how to document a retrospective look on lessons learned from this process... I'm not sure. We've never landed on a good way or place to do that! Open to thoughts on that one.

tristan-orourke:
Thoughts on some manual refinement to make:
- how does `OAUTH_REDIRECT_URI` store /en and /fr callback uris?
- `02 — Deployment: Post-deployment lifecycle` illustrates our particular deployment process but seems unrelated to CanadaLogin in particular
- A lot of this, like the issue we hit with queued jobs, is probably very specific to our Laravel setup, not just an azure mult-slot plus CanadaLogin configuration. That's fine, just maybe want to emphasize that more?
- In section 05, Phased migration, there are some other issues we hit that could be expanded on, like ensuring we started a clean session with SiC and never using the skip_migration for returning users.

PR #17336, opened 2026-07-08
[Fix] Sending emails to null

### LINKED ISSUE #17032: 🐛 SendNotificationsAdHocEmail tries to send email to null

We're using the SendNotificationsAdHocEmail command to send CanadaLogin comms emails.  The "Notify All Users" tries to send messages to users with a null email address.  It's not a big problem but does slow down the sending process.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] The SendNotificationsAdHocEmail command does not try to send messages to null email address.

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #17032 

## 👋 Introduction

Fixes and issue where the adhoc email was attempting to send to users who had a `null` email.

## 🧪 Testing

> [!NOTE]
> I'm not sure how to get info to tell this happened during manual testing but there is a test for it. I'll do my best guess in the testing instructions

1. Make sure a user has a `null` email
2. Run the adhoc email command
3. Confirm no errors or attempts to send to the user with the null emil
    - Maybe by chekcing the send couint to make sure it only countsd those with an email?

### DIFF
--- a/api/app/Console/Commands/SendNotificationsAdHocEmail.php
+++ b/api/app/Console/Commands/SendNotificationsAdHocEmail.php
@@ -138,10 +138,13 @@ class SendNotificationsAdHocEmail extends Command implements PromptsForMissingIn
             }
         }
 
-        $builder = User::whereJsonContains('enabled_email_notifications', $notificationFamilies[0]);
-        for ($i = 1; $i < count($notificationFamilies); $i++) {
-            $builder->orWhereJsonContains('enabled_email_notifications', $notificationFamilies[$i]);
-        }
+        $builder = User::whereNotNull('email')
+            ->where(function (Builder $query) use ($notificationFamilies) {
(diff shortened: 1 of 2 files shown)


petertgiles (changes requested):
Nice, the fix works great for "all users".  Would you mind adding it to the `builderFromNotificationFamilies` function, too?

esizer:
> propogate fix to family method

Oops, didn't see that one :sweat_smile: Should be done in: [propogate fix to family method](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17336/commits/a6b3532158762dc0e29513fc0f57fde323bc60d0)

petertgiles (approved):
Awesome, thanks.

PR #17338, opened 2026-07-08
Fix false positive 404 from bot-blocking hosts in link checker

### LINKED ISSUE #17337: Fix false positive broken links from bot-blocking hosts in link checker
## Problem

The external link checker workflow incorrectly flags valid URLs from certain Canadian government hosts (e.g. `connexion.canada.ca`) as broken (404) on every run.

**Example false positive:**
```json
{
  "file": "apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx",
  "url": "https://connexion.canada.ca/fr/utilisateurs/nous-contacter/",
  "status": 404
}
```

## Root Cause

`connexion.canada.ca` is already listed in `KNOWN_AUTH_REQUIRED_HOSTS` (which was added to handle 403 bot-blocking responses), but this host returns **404** — not 403 — when it detects a non-browser User-Agent.

In `packages/linkChecker/src/checkExternalLinks.ts`, the retry-with-browser-headers logic only triggers on 403:

```ts
if (res.status === 403 && isKnownAuthRequiredHost(url)) {
  return fetchLinkWithBrowserHeaders(url, timeoutMs);
}
```

So 404 responses from these hosts are treated as genuine broken links, even though the page is reachable via a real browser.

## Fix

Extend the browser-header retry to also trigger on 404 from known auth-required hosts:

```ts
if ((res.status === 403 || res.status === 404) && isKnownAuthRequiredHost(url)) {
  return fetchLinkWithBrowserHeaders(url, timeoutMs);
}
```

If the page is genuinely missing, `fetchLinkWithBrowserHeaders` will still return 404 and correctly fail. If the host was just blocking the bot, it returns 200 and passes.

brindasasi:
Fixes #17337

## Summary

- `connexion.canada.ca` returns 404 (not 403) to block bot User-Agents, but the browser-header retry in the link checker only triggered on 403
- Extended the retry condition to include 404 from `KNOWN_AUTH_REQUIRED_HOSTS`, so these URLs are re-checked with real browser headers before being reported as broken
- Genuine 404s still fail correctly — `fetchLinkWithBrowserHeaders` will also return 404 if the page truly doesn't exist

## Test plan

- [ ] Trigger the [LinkChecker workflow](https://github.com/GCTC-NTGC/gc-digital-talent/actions/workflows/external-link-checker.yml) manually  and confirm `connexion.canada.ca/fr/utilisateurs/nous-contacter/` no longer appears in the broken links report


esizer (approved):
🤞

PR #17358, opened 2026-07-09
[Bug] Fix talent sources in edit dialog

### LINKED ISSUE #17355: 🐛 Talent sources always show not available

The source of talent section always shows "Not available" even if user has 1 or more talent sources. This is a regression that happened probably during resolving merge conflicts.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Login as admin
2. Go to candidates tab on a talent request
3. Open dialog by clicking on user name from table/inbox
4. See "Source of talent" is always null (Not available)

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

1. Fix `sourceLabels` always showing "N/A" in `TalentRequestEditReferralDialog.tsx` by reading `source.label.localized` directly (matching `TrackedUserListItem`) instead of looking it up in a `sourceOptions` list that was never actually populated, since the `sourceOptionsQuery` prop was never passed by the real caller. 

2. Remove the now-dead fragment/prop from the dialog, its Storybook mock, and the unused fragment spread in `Tracking.tsx`.

### ✅ Acceptance Criteria

- [ ] Dialog shows talent sources if user has any

### 🔄 Regression

Yes

### 🔗 Root Cause Pull Request

_No response_

yonikid15:
🤖 Resolves #17355 

## 👋 Introduction

- Adds fix for "Source of talent" section in talent request referral edit dialog.

## 🧪 Testing

1. Build front-end
2. Login as admin
3. Go to talent request with tracked users
4. Open dialog by clicking on user in the inbox/table
5. Ensure talent sources are showing

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/pages/TalentRequests/Tracking.tsx
+++ b/apps/web/src/pages/TalentRequests/Tracking.tsx
@@ -29,7 +29,6 @@ const TalentRequestTracking_Query = graphql(/** GraphQL */ `
     }
 
     ...TalentRequestReferralDialogOptions
-    ...TalentRequestEditReferralDialogSourceOptions
   }
 `);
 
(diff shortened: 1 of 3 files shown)


esizer (approved):
LGTM! 👍 Thanks!

PR #17360, opened 2026-07-09
Fix Playwright tests to work against UAT environment

### LINKED ISSUE #17241: 🐛 UAT | Fix Playwright failure in UAT

- Some playwright tests are failing in regression automation run:

Image

 

### 🕵️ Details

Env: UAT

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

_No response_

### 🔄 Regression

Yes

### 🔗 Root Cause Pull Request

_No response_

brindasasi:
resolves #17241

## Summary

- Added `override: true` to `dotenv.config()` in `playwright.config.ts` so `.env.uat` values (especially `BASE_URL`) override any pre-existing values from the local `.env`
- Switched the UAT token endpoint call in `getTokenForSub` from Playwright's `request.newContext()` to native Node.js `fetch` — custom headers (`X-Testing-Secret`) were not being sent correctly with the Playwright request context
- Fixed `resolveFixtureSub` in `utils/auth.ts`: dynamic email-format subs (e.g. `playwright.xxx@gc.ca`) now pass through unchanged instead of throwing a "no remote sub configured" error, so integration tests that create users dynamically work on UAT
- Added `tests/admin/platform-admin-smoke.spec.ts`: read-only regression smoke test that logs in as a real platform admin (via `PLAYWRIGHT_PLATFORM_ADMIN_SUB`) and asserts the admin dashboard loads
- Added `tests/applicant/applicant-smoke.spec.ts`: read-only regression smoke tests that log in as a real applicant (via `PLAYWRIGHT_APPLICANT_SUB`) and verify all dashboard sections are visible

## Test plan

- [ ] Run `PLAYWRIGHT_ENV_FILE=.env.uat pnpm exec playwright test tests/admin/platform-admin-smoke.spec.ts --project=chromium` — expect pass
- [ ] Run `PLAYWRIGHT_ENV_FILE=.env.uat pnpm exec playwright test tests/applicant/applicant-smoke.spec.ts --project=chromium` — expect pass
- [ ] Run an existing integration test (e.g. `applicant-dashboard.spec.ts`) against UAT — dynamic user creation should still work with `resolveFixtureSub` fix
- [ ] Confirm local runs (without `.env.uat`) are unaffected
## Screenshot 
image


api/database/migrations/2026_05_01_212040_community_development_program_add_soft_deletes.php
```diff
@@ -12,9 +12,11 @@
...
     {
         //
-        Schema::table('community_development_program', function (Blueprint $table) {
-            $table->softDeletes();
-        });
+        if (! Schema::hasColumn('community_development_program', 'deleted_at')) {
+            Schema::table('community_development_program', function (Blueprint $table) {
+                $table->softDeletes();
+            });
+        }
```
brindasasi: I've no idea why would this commit come up here .. will revert

apps/playwright/tests/auth/setup-helper.ts
brindasasi, line 1: These files are removed because these implemented the old flow: authenticate once in a Playwright "setup" project, write the resulting cookies/localStorage to .auth/*.json, then have uat-admin/uat-applicant projects load that file via storageState.

 This is replaced by calling loginBySub() directly in each spec's beforeEach.

Motivation behind this ..

 The setup step used Playwright's request context to hit /refresh, and that context wasn't sending X-Testing-Secret correctly against UAT - so the setup-then-reuse pattern was failing outright, not just redundant.

 Doing the login per-spec removes the dependency on a fragile shared setup step and lets getTokenForSub (now using native fetch) do the auth for every test uniformly.

mauli0110 (approved):
LGTM !! Smoke tests files are really helpful to know the env is ready for the execution, but I wonder if we can add additional tag just for differentiate in the filing system, rest all are perfectly working. 
I can make this change in my upcoming PR for running all smoke tests in UAT

PR #17362, opened 2026-07-09
[Feat] Add talent source filters

### LINKED ISSUE #17039: ✨ Add additional filters to new Talent Request matching users table
### Details

- Talent Source (?)

Image

- Applicant filters (?)
  - Default filters that can't be changed

Still needs more discussion on how exactly it will work.

### Design file & designer

Designed by @Jerryescandon: [Figma](https://www.figma.com/design/OcJlqTARk7UEG2aK7H0AG5/Talent-requests--Communities-?node-id=1666-18782&m=dev)

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

_No response_

### ✅ Acceptance Criteria

_No response_

esizer:
🤖 Resolves #17039 

## 👋 Introduction

Adds a filter for talent sources on the matches table for a talent request.

## 🧪 Testing

> [!NOTE]
> Our factory only created tracked users. This also updates the factoty to ensure some mathcing users as well.  So, you will need to re-seed.

1. Refresh, seed and build `make refresh-api; make seed-fresh; pnpm dev:fresh`
2. Login as `admin@test.com`
3. Navigate to a talent requests tracking tab
4. Confirm the new filter appears in the matches table
5. Confirm it functions as expected

## 📸 Screenshot

swappy-20260709_152445


api/database/factories/TalentRequestFactory.php
```diff
@@ -68,57 +71,92 @@ public function completed(): self
...
+
+        if ($matchesQualifiedInPool) {
+            $pool = Pool::factory()->candidatesAvailableInSearch()->create(array_filter([
+                'community_id' => $filter->community_id,
+                'classification_id' => $filter->qualifiedInClassifications->shuffle()->first()?->id,
+                'work_stream_id' => $filter->qualifiedInWorkStreams->shuffle()->first()?->id,
+            ]));
+
+            if ($filter->pools()->exists()) {
+                $filter->pools()->attach($pool->id);
```
tristan-orourke, line 114: **Correctness (CONFIRMED):** `$filter->pools()->attach($pool->id)` progressively mutates the shared `ApplicantFilter` across loop iterations.

When `withTrackedUsers(N)` or `withMatchingUsers(N)` calls `createUser` N times, each iteration that enters the `$matchesQualifiedInPool` branch creates a new Pool and attaches it to the filter. The filter's pool list grows with each iteration.

**Trigger:** `withTrackedUsers(3)` on a filter with 2 pools → after 3 iterations the filter may have 5 pools instead of 2.

**Suggestion:** Either skip the attach (the pool candidate is already discoverable via search), or create/share the pool outside the loop.

:robot: Generated with Claude Code

api/graphql/schema.graphql
```diff
@@ -1506,6 +1506,23 @@ input ApplicantFilterInput {
   talentSources: [TalentRequestSource!]
 }
 
+input TalentRequestApplicantFilterInput {
```
tristan-orourke, line 1510: **Maintenance risk (CONFIRMED):** `TalentRequestApplicantFilterInput` is a field-for-field copy of `ApplicantFilterInput` with `@scope` directives stripped. There is no comment, test, or lint rule enforcing sync between the two.

The duplication is structurally necessary (Lighthouse's `@scope` directives would fire against the wrong model), but if a field is added to `ApplicantFilterInput`, a developer must independently remember to mirror it here — or the talent-request filter silently ignores it.

**Suggestion:** Add a comment on both types cross-referencing the other (e.g. `# Keep in sync with TalentRequestApplicantFilterInput`). A test asserting field parity (minus directives) would be even better.

apps/web/src/pages/TalentRequests/components/TalentRequestMatchesTable/TalentRequestMatchesFilterDialog.tsx
```diff
@@ -185,7 +196,21 @@ const TalentRequestMatchesFilterDialog = ({
...
-      <Heading level="h3" size="h5" className="mt-0 mb-6 font-bold">
+      <Checklist
+        idPrefix="talentSources"
+        name="talentSources"
+        legend={intl.formatMessage(talentRequestMessages.sourceOfTalent)}
+        items={narrowEnumType(
+          unpackMaybes(options?.talentSources),
+          "TalentRequestSource",
+        )
+          .filter((source) => source.value !== TalentRequestSource.Advancement)
```
tristan-orourke, line 208: **Maintenance risk (CONFIRMED):** This hardcodes which `TalentRequestSource` values to exclude. The backend already tracks which sources are implemented (`matchRelation() !== null`) via `TalentRequestSource::selected()`, but this info is not exposed to the frontend.

When Advancement is eventually implemented, a developer must remember to remove this filter or users will never see the option.

**Suggestion:** Either add a `// TODO: remove when Advancement is implemented` comment, or expose an `implemented` flag from the backend so the frontend can filter dynamically.

api/app/Builders/UserBuilder.php
```diff
@@ -394,11 +394,15 @@ public function whereMatchesTalentRequest(?array $args): self
         return $this;
     }
 
-    // eager-load each selected source's matched, view-authorized records onto the user
     public function withTalentRequestMatches(array $filters): self
     {
-        foreach (TalentRequestSource::selected($filters['talentSources'] ?? null) as $source) {
-            $this->with([$source->matchRelation() => fn ($r) => $r
+        foreach (TalentRequestSource::cases() as $source) {
```
tristan-orourke, line 399: **Efficiency (PLAUSIBLE):** This now iterates `TalentRequestSource::cases()` instead of `::selected()`, eager-loading all implemented source relations regardless of the user's filter. When filtering by 1 source, this adds 1 extra constrained query per page load.

With only 2 implemented sources the impact is minimal, and this appears intentional so the UI can display match info for all source columns. Worth a brief comment documenting the intentional divergence from `whereMatchesTalentRequest` (which still uses `::selected()`).

tristan-orourke (commented):
Code review (medium effort): 4 findings — 1 correctness, 2 simplification, 1 efficiency.

api/graphql/schema.graphql
esizer: The old  one is going to go away when we dso the cleanup so im not sure if we should be trying to keep them in sync. do we plan on changing the shape of applicant filter in between now and when we remove the legacy code? :thinking:

apps/web/src/pages/TalentRequests/components/TalentRequestMatchesTable/TalentRequestMatchesFilterDialog.tsx
esizer: Do we have a ticket number for this? I'll addd the note but i do like adding ticket numbers so we know its planned and where to look.

api/app/Builders/UserBuilder.php
esizer: This is enitrely intentional for the reason stated. i'm confused why we would add a comment. WE are looping thourhg all so we can show all sources of talent.

Or was there somerthing else about this you found unclear?

api/database/factories/TalentRequestFactory.php
esizer: I'm a little confused. This is to make sure the attached usaers satisfy the criteria (multiple pool sources). We can't skip it.

I'm up for doing it a different way if you have a suggestion but skipping it will lead to incorrect data.

tristan-orourke:
Sorry @esizer, I was experimenting with Claude code reviews, and I don't know how useful it was in this case. I'll do a real review now :sweat_smile:

api/database/factories/TalentRequestFactory.php
tristan-orourke: Changing the original talent request filters when creating users that match it could be considered an unexpected side-effect. I can imagine a test that creates a TalentRequest for a particular pool, runs `withMatchingUsers`, and expects them to be in that pool.

I see why you did it like this though. I think the issue is that our current ApplicantFilterFactory doesn't guarantee consistency in its filters? There's no guarantee that its pools will match its other filters like classification and stream.

Ideally we'd have consistent ApplicantFilters, and then if `filter->pools` is populated we'd create a PoolCandidate from one of those pools here. But that might be more effort than its worth. Could you add a comment to `withMatchingUsers` and `withTrackedUsers` noting that it may modify `$talentRequest->filter` to ensure created users match it?

api/graphql/schema.graphql
tristan-orourke: Good point. I don't think we have a cleanup ticket for this yet though. Since you've been knee-deep in this code for a while, could you create one for us?

apps/web/src/pages/TalentRequests/components/TalentRequestMatchesTable/TalentRequestMatchesFilterDialog.tsx
tristan-orourke: Good call. Here, its a bit of a stub, but I just created a ticket for you: https://github.com/GCTC-NTGC/gc-digital-talent/issues/17382

api/database/factories/TalentRequestFactory.php
tristan-orourke: I think better ApplicantFilter factories might actually be worth a refactor https://github.com/GCTC-NTGC/gc-digital-talent/issues/17380

api/app/Builders/UserBuilder.php
tristan-orourke: I believe you that this is intentional, and the bot was overzealous here.

I _do_ find this function confusion though. I'm trying to trace through how it works. 
1. It uses this iteration on the Enum to find all the possible implemented talent sources. Each of them return a 'matching____Sources' property.
2. Those properties are implemented in the HasTalentRequestSources trait which User uses
3. those `matching___Sources` properties turn into Relationships (returning all of the users PoolCandidates or CommunityInterests)
4. except its not _all_ the related PoolCandidates or CommunityInterests, because those models also implement `whereMatchesTalentRequest` that we then call here

I think that's it? This took me a while to follow through, I do feel like we could use some comments here explaining the whole sequence :sweat_smile: 

Could it be worth it to add a talent source interface or something, to guarantee they have the `whereMatchesTalentRequest` function?

api/database/factories/TalentRequestFactory.php
esizer: Added in [document factory methods](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17362/commits/b98ec2c618a51ff8d9ac09cc7d10f9be65904b04)

api/app/Builders/UserBuilder.php
esizer: okay, here is an interface implement. Does this help? [add and implement interface](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17362/commits/bc32885ef9b09780c305a329f2abf8e484538103)

apps/web/src/pages/TalentRequests/components/TalentRequestMatchesTable/TalentRequestMatchesFilterDialog.tsx
esizer: Added in [comment talent request matches filter out advancment](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17362/commits/6347551cd06f0a9e7cec575e35fc90a132c2d872)

api/app/Builders/UserBuilder.php
```diff
@@ -362,8 +362,8 @@ public function whereMatchesTalentRequest(?array $args): self
         $filters = $args ? ($args['applicantFilter'] ?? $args) : [];
         $skillIds = $filters['skills'] ?? []; // already plain ids via ApplicantFilterInput @pluck
 
-        // keep only users who match at least one of the request's selected sources
         $this->where(function ($query) use ($filters) {
+            $query->whereRaw('1 = 0'); // false starting point for the orWhereHas chain
```
tristan-orourke, line 366: So this would mean that if you pass in filters with no talent source, you match zero users, right? I guess that makes sense.

api/graphql/schema.graphql
tristan-orourke: I was taking a closer look at this - I don't know if this ApplicantFilterInput can go away entirely, since its also used by the usersPaginated query. If that's the case then we should leave a comment here. Something like 
`# Identical to ApplicantFilterInput except without scopes. Matching is handled by `whereMatchesTalentRequest` instead of by `UserBuilder` scopes.`

tristan-orourke (changes requested):

api/app/Builders/CommunityInterestBuilder.php
```diff
@@ -0,0 +1,93 @@
...
+namespace App\Builders;
+
+use App\Contracts\TalentRequestMatchable;
+use App\Models\User;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Support\Facades\Auth;
+
+class CommunityInterestBuilder extends Builder implements TalentRequestMatchable
+{
+    public function whereMatchesTalentRequest(?array $filters): self
```
tristan-orourke, line 19: **Correctness:** `whereMatchesTalentRequest` doesn't enforce `consent_to_share_profile`

`PoolCandidateBuilder::whereMatchesTalentRequest` calls `whereAvailable()` to enforce availability constraints, but the CommunityInterest equivalent has no consent check. The `orWhereHas` existence check at `UserBuilder.php:368` can therefore match users whose community interest has `consent_to_share_profile = false`, surfacing them in search results. The `whereAuthorizedToView` on the eager-load path hides the *records*, but the user still appears in the result set — a user who hasn't consented could show up as a match with an empty `matchingAtLevelSources` array.

**Trigger:** User has a CommunityInterest with `consent_to_share_profile = false`. An AT_LEVEL talent request search returns them as a match, leaking the fact that they exist and match the criteria.

**Suggestion:** Add `$this->where('consent_to_share_profile', true);` to `CommunityInterestBuilder::whereMatchesTalentRequest`, mirroring how `PoolCandidateBuilder` calls `whereAvailable()`.

---
<sub>🤖 Generated with [Claude Code](https://claude.ai/code)</sub>

api/app/Builders/CommunityInterestBuilder.php
```diff
@@ -0,0 +1,93 @@
...
+                    // all community teams that the user is a member in
+                    $allCommunityTeams = $user->rolesTeams()
+                        ->where('teamable_type', "App\Models\Community")
+                        ->get();
+
+                    // filter community teams down to those where the user also has permission to see the interests
+                    $viewPermissionCommunityTeams = $allCommunityTeams
+                        ->filter(fn ($team) => $user->isAbleTo('view-team-communityInterest', $team));
+
+                    $query->whereIn('community_id', $viewPermissionCommunityTeams->pluck('teamable_id')->toArray());
```
tristan-orourke, line 87: **Conventions:** Hardcoded model class string instead of `::class` constant

`"App\Models\Community"` is used as a string literal for filtering `teamable_type`. This existed in the original code, but since it's being extracted into a new file this is a good opportunity to use `Community::class` instead for refactor safety.

**Trigger:** If the `Community` model is ever moved or renamed, this string silently stops matching.

**Suggestion:** Replace `"App\Models\Community"` with `Community::class` (and add the import).

---
<sub>🤖 Generated with [Claude Code](https://claude.ai/code)</sub>

api/tests/Feature/TalentRequestMatchesTest.php
```diff
@@ -694,6 +698,52 @@ public function testTalentSourcesAtLevelOnlyExcludesPoolOnlyUsers(): void
             ->assertJsonPath('data.talentRequestMatches.data.0.matchingAtLevelSources.0.id', $interest->id);
```
tristan-orourke, line 720: **Test coverage:** No test for consent filtering on AT_LEVEL source

The two new tests cover AT_LEVEL filtering by work stream and classification, but there's no test confirming that a user with `consent_to_share_profile = false` is excluded from AT_LEVEL results. Given the missing consent check noted in another comment, a test would both catch the current bug and prevent regressions.

**Trigger:** A future change accidentally removes or bypasses the consent check with no test failure.

**Suggestion:** Add a test creating a user with a CommunityInterest where `consent_to_share_profile = false`, run an AT_LEVEL search, and assert the user does *not* appear.

---
<sub>🤖 Generated with [Claude Code](https://claude.ai/code)</sub>

tristan-orourke (commented):
Code review: 3 finding(s) posted as inline comments.

---
🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>If this review was useful, react with 👍. Otherwise, react with 👎.</sub>

api/app/Builders/UserBuilder.php
esizer: No, the opposite. If you pass in with no filters, you match all of them. The `orWhere` clauses after attach to this. It is just a base query to work from.

```
SELECT * FROM users WHERE (
  (1 = 0) OR (qualified) OR (at level)
)
```

api/app/Builders/CommunityInterestBuilder.php
esizer: I'm confused, isnmt that a concern of authorization and not matching? :thinking:

api/app/Builders/CommunityInterestBuilder.php
esizer: I'm really confused here. Where in this code is a class name hardcoded in a string? :face_with_spiral_eyes:

api/tests/Feature/TalentRequestMatchesTest.php
esizer: Why would they be excluded? consent to share is an authorization concern. I'm a little confused by this comment.

api/graphql/schema.graphql
esizer: That raises a different question. Why does that query use `ApplicantFilter`? I thought those were specific to talent requests? :thinking:

tristan-orourke: Haha fair question! I think it was for convenience? Like, we wanted to filter users by all those fields and we already had a type set up with the right scopes, so we reused it. I'm definitely open to refactoring it.

api/app/Builders/UserBuilder.php
tristan-orourke: Are you saying that `orWhere` only works if theres an initial clause before it?

In either case, if there's no filter talent sources, you end up with `Select * from users WHERE 1=0`. Isn't that an impossible condition that will return zero rows?

api/app/Builders/CommunityInterestBuilder.php
tristan-orourke: Ugh, it got the line for the comment wrong. This should be at line 77.

```
 $allCommunityTeams = $user->rolesTeams()
                        ->where('teamable_type', "App\Models\Community")
                        ->get();
```

api/app/Builders/CommunityInterestBuilder.php
tristan-orourke: I think arguably its a concern for both. If a user has said they don't want to be matched, we should probably avoid matching them, even before we double check authorization.

tristan-orourke (approved):

PR #17368, opened 2026-07-10
[Fix] Default value reset for select inputs

### LINKED ISSUE #17363: 🐛 UAT | Filter values are not erased when resetting the filters in Find matching candidates table

- When user reset the filters applied to 'Find matching candidates' table, the filter value especially in profile filters are not erased, the selected data stays there
- I rechecked it is for all tables - Candidates, Users, and Community talent table

**Expected outcome:** Reset filters should reset or erased applied filters in 'Find matching candidates' table

### 🕵️ Details

Env: UAT
Version: 2.76.1

### 📋 Steps to Reproduce

1. Access to GC digital talent using community admin role
2. Navigate to the any open talent requests
3. Navigate to Find matching candidate table
4. Apply some filters such as Working language, Employment duration in profile filters and other filters
5. Verify the filter is applied correctly
6. Go to filters again and reset the filter
7. Verify the result is showing correctly post resetting the filters
8. Click on Filters
9. Verify the values or filter selected under profile filters remains there, it is not erased

### 📸 Screenshot

https://github.com/user-attachments/assets/e6686b36-40da-4e0f-b263-70afdb3def91

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Reset filters should erase the values in profile filters 

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #17363 

## 👋 Introduction

Fixes an issue where some select inputs were not resetting correctly.

## 🕵️ Details

This was due to the selects having `undefined` as their default value. It seems `react-hook-form` silently ignores those when resetting. This updates them to be an empty string instead which is honoured at the time of reset reliably.

## 🧪 Testing

> [!NOTE]
> This seemed to only be an issue when the default vlaue was empty so make sure those are included in the test. 

1. Build `pnpm dev:fresh`
2. Login as `admin@test.com`
3. Navigate to some tables with filter dilaogs
4. Edit some of the inputs (all if possible but foxcus on empty selects)
5. Apply the filters
6. Re-open the dialog and reset the values
7. Confirm all inputs return to their original value before any changes were made

### DIFF
--- a/apps/web/src/components/PoolCandidatesTable/helpers.tsx
+++ b/apps/web/src/components/PoolCandidatesTable/helpers.tsx
@@ -560,7 +560,7 @@ export function transformFormValuesToFilterState(
 ): PoolCandidateSearchInput {
   return {
     applicantFilter: {
-      languageAbility: data.languageAbility,
+      languageAbility: data.languageAbility || undefined,
       operationalRequirements: data.operationalRequirement,
       locationPreferences: data.workRegion,
       flexibleWorkLocations: data.flexibleWorkLocations,
--- a/apps/web/src/components/PoolCandidatesTable/helpers.tsx
+++ b/apps/web/src/components/PoolCandidatesTable/helpers.tsx
@@ -560,6 +560,8 @@ export function transformFormValuesToFilterState(
 ): PoolCandidateSearchInput {
   return {
     applicantFilter: {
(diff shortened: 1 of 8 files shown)


petertgiles (approved):
Looks great!  🏅

PR #17369, opened 2026-07-10
[Fix] Follow up date overdue state

### LINKED ISSUE #17365: 🐛 UAT | Future follow-up date incorrectly shows "0 days overdue" badge on talent request

- On the talent request details page, a follow-up date set in the future [even just one day after today] displays a `"0 days overdue"` badge next to it.

**Expected outcome**: The overdue badge should only appear when the follow-up date is in the past (i.e., at least one day overdue). For today's date or any future date, no overdue indicator should be shown.

### 🕵️ Details

Env: UAT
Version: 2.76.1

### 📋 Steps to Reproduce

1. Access GC digital portal using community admin role
2. Navigate to talent requests
3. Open the talent request which is in progress or new
4. Update the talent request and add follow-up date to next day 
5. Verify the 0 days overdue badge is present

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] A follow-up date set to a future date (e.g., today + 1 day) displays no overdue badge

### 🔄 Regression

Yes

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #17365 

## 👋 Introduction

Fixes an issue with the overdue bagde being set too early on a talent requests follow up date.

## 🕵️ Details

The issue was, we were treating 0 as overdue. We swapped the underlying function to calendar days rather than 24 hour period since the date has no time associated with it. We also changed the `<=` to a simple `<` so that a 0 value ddi not trigger "overdue"

## 🧪 Testing

1. Find a talent request in the database and set its overdue date to less than 24 hours in the future
2. Build `pnpm dev:fresh`
3. Login as `admin@test.com`
4. Navigate to the talent request
5. Confirm the followup date does not show "overdue"
6. Set the date to less than 24 hours in the past
7. Refresh the page and confirm irt now says overdue by a single day

### DIFF
--- a/apps/web/src/utils/searchRequestUtils.test.ts
+++ b/apps/web/src/utils/searchRequestUtils.test.ts
@@ -8,14 +8,16 @@ describe("followUpDateOverdueInfo", () => {
     expect(followUpDateOverdueInfo(followUpDate, now)).toEqual({
       daysOverdue: -1,
       isOverdue: false,
+      isDueToday: false,
     });
   });
 
-  test("is not overdue when follow-up date is today", () => {
+  test("is due today, not overdue, when follow-up date is today", () => {
     const followUpDate = new Date("2026-07-10T04:00:00Z");
     expect(followUpDateOverdueInfo(followUpDate, now)).toEqual({
       daysOverdue: 0,
       isOverdue: false,
+      isDueToday: true,
     });
   });
 
@@ -24,6 +26,7 @@ describe("followUpDateOverdueInfo", () => {
     expect(followUpDateOverdueInfo(followUpDate, now)).toEqual({
       daysOverdue: 1,
       isOverdue: true,
+      isDueToday: false,
     });
   });
 
@@ -31,6 +34,7 @@ describe("followUpDateOverdueInfo", () => {
     expect(followUpDateOverdueInfo(null, now)).toEqual({
       daysOverdue: -1,
       isOverdue: false,
+      isDueToday: false,
     });
   });
 });
--- a/apps/web/src/utils/searchRequestUtils.test.ts
+++ b/apps/web/src/utils/searchRequestUtils.test.ts
@@ -1,0 +1,36 @@
+import { followUpDateOverdueInfo } from "./searchRequestUtils";
+
+describe("followUpDateOverdueInfo", () => {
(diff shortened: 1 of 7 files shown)


apps/web/src/utils/searchRequestUtils.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    });
+  });
+
+  test("is not overdue when follow-up date is today", () => {
+    const followUpDate = new Date("2026-07-10T04:00:00Z");
+    expect(followUpDateOverdueInfo(followUpDate, now)).toEqual({
+      daysOverdue: 0,
+      isOverdue: false,
+    });
+  });
```
tristan-orourke: Intended behaviour is actually to show the badge when follow-up date is today

tristan-orourke (changes requested):
I like the addition of the tests, but showing the badge when follow up date is today was intended behaviour 😆. The idea is, if you're a recruiter looking at the list, the best time to follow up with a request is _on_ the follow up date. Therefore, we want to bring it to their attention on that day, not a day later.

Given all that, we should add a comment making that explicit.

apps/web/src/utils/searchRequestUtils.test.ts
esizer: ohh, so what should it say? Was the message of "0 days overdue" correct and it was just when it was appearing? :sweat_smile:

@Jerryescandon maybe you cna help out, I changed this mainly because the message seemed incorrect. It seemed odd to me to see "0 days overdue" becuase that seemed nonsensical to me :stuck_out_tongue_closed_eyes:

tristan-orourke: What if we had a special "Due today" message for 0 days overdue?

Jerryescandon: I like it. Simple. Ship it! 
image

esizer: Thanks to both, got it in with update message when due to today

tristan-orourke (approved):

tristan-orourke (approved):

Issue #17167, opened 2026-06-19
✨ CanadaLogin - PKCE


petertgiles:
### Details

The CanadaLogin team recommends implementing PKCE.

We have several ancient issues on this topic:
- https://github.com/GCTC-NTGC/gc-digital-talent/issues/4916
- https://github.com/GCTC-NTGC/gc-digital-talent/issues/2572

### 📸 Screenshots

<img width="657" height="309" alt="Image" src="https://github.com/user-attachments/assets/7c8e2310-0769-4ecf-bc2f-96cd39956f4e" />

### ✅ Acceptance Criteria

- [ ] PKCE feature is supported

petertgiles:
Reviewed in standup:
- worth starting another spike for this

tristan-orourke:
Spike completed!

> Do we want to try adopting PKCE?

Yes, this should actually be quite easy.


> Can we build the functionality ourselves without integrating a new auth library?

Yes!
1. We generate a random string ("code_verifier") and store it in the session. 
2. We hash it to create a "code_challenge" and pass the challenge with our original authorization code request. 
3. We then send the original code_verifier along when we request the access tokens, and the CanadaLogin server can verify that it was used to create the original code_challenge, thereby proving we are the same client in both cases.

> Effort vs value?

Fairly low value I think, but even lower effort, so lets go ahead.


See #17376 for a demo implementation.

PR #17376, opened 2026-07-10
[Feature] Add PKCE to OAuth authorization code flow

### LINKED ISSUE #17342: 🦔 Plan for PKCE auth change
### 🦔 Spike

CanadaLogin strongly recommends confidential clients (like us) adopt PKCE in our integration.  

https://github.com/GCTC-NTGC/gc-digital-talent/issues/17167

We have tried this before by integrating a whole new auth library and ended up abandoning it.  Let's take time to plan before adopting this again.

### ❓ Questions to answer

- Do we want to try adopting PKCE?
- Can we build the functionality ourselves without integrating a new auth library?
- Effort vs value?

### 🕙 Timebox

- 5 days

tristan-orourke:
Adds Proof Key for Code Exchange (RFC 7636) to the CanadaLogin auth flow as defense-in-depth alongside the existing client_secret, aligning with OAuth 2.1 and iGov government profile recommendations.

🤖 Resolves #17342

## 👋 Introduction

This is a quick proof of concept for a spike, answering questions related to PKCE: should we adopt it, would it require a library, how much work would it be, would it be worth it?

## 🕵️ Details

I think I finally understand PKCE! Its actually just a simple way of ensuring that the client who requested an Authorization Code is the same client who exchanges that Authorization Code for access tokens. 

I found this to be a good explanation: https://www.descope.com/learn/post/pkce

> Do we want to try adopting PKCE?

Yes, this should be quite easy.

> Can we build the functionality ourselves without integrating a new auth library?

Yes, this should be quite easy. 
1. We generate a random string ("code_verifier") and store it in the session. 
2. We hash it to create a "code_challenge" and pass the challenge with our original authorization code request. 
3. We then send the original code_verifier along when we request the access tokens, and the CanadaLogin server can verify that it was used to create the original code_challenge, thereby proving we are the same client in both cases.

> Effort vs value?

Fairly low value I think, but even lower effort, so lets go ahead.

## 🧪 Testing

1. Configure your environment to connect to a CanadaLogin auth server, and ensure you can still log in.
4. Ensure you can still log in with the mock auth system as well.

## 🚚 Deployment

I don't think there's any real work to do during a deployment (no env variables changed, no scripts to run) but I want to flag that this will need extra care when it comes to testing.

### DIFF
--- a/api/app/Http/Controllers/AuthController.php
+++ b/api/app/Http/Controllers/AuthController.php
@@ -15,6 +15,7 @@ use Illuminate\Support\Facades\Http;
 use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Str;
 use InvalidArgumentException;
+use Jose\Component\Core\Util\Base64UrlSafe;
 use Lcobucci\JWT\Configuration;
 use Lcobucci\JWT\UnencryptedToken;
 use Throwable;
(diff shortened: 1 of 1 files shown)


petertgiles:
Wow, this looks great.  I feel like I'm starting to understand it finally, too.  Turns out this is more manageable when you're not trying to swap out your full auth client.  :laughing: 

With this reference I feel like I'd be confident to try to add this feature. :+1:

tristan-orourke:
Okay, I'll close this pull request and reference it in the original feature which should be unblocked now (and no longer 13 points!) #17167

Issue #17181, opened 2026-06-22
🐛 Prod - lots of 500s on /app/gcdt


petertgiles:
### 🐛 Bug and Expected Behaviour

In production we have many requests to /app/gcdt resulting in HTTP 500.

### 📋 Steps to Reproduce

```kql
AppServiceHTTPLogs
| where ScStatus >= 500
```

### 📸 Screenshot

<img width="1553" height="646" alt="Image" src="https://github.com/user-attachments/assets/3ea40149-997b-49e4-a1ac-055a5ec0da2f" />

### 🙋‍♀️ Proposed Solution

~I noticed the LIGHTHOUSE_BROADCASTER variable wasn't set in prod.  Maybe that's it?~ The env variables have been fixed, at least the issues I found.

Some strategies and places to start:
- Try some more to reproduce this in prod.  Is it possible to trigger the 500s somehow?  Logging out in one tab maybe?  Logging out in a different device?  Try some different browsers?  Turn on security modes or ad blocking?
- Check the log file again.  Do we have more now that the env variables are fixed?  Is [the redirect](https://github.com/GCTC-NTGC/gc-digital-talent/blob/bec5811ac30e6ae2a17dee6b8978b3f92d933dbc/infrastructure/conf/reverb.conf#L6) not working and are we losing logs?  Is there logging going to a different file that we didn't check?
- Application insights - can we find telemetry on what is happening in the users' browsers when this happens?
- Azure logs - are there clues in either the HTTP logs or the app console logs at the exact same timestamps that would give us a clue to what's happening?
- Documentation - are there any clues in the Reverb documentation that we missed?

### ✅ Acceptance Criteria

- [ ] No more HTTP 500s on  /app/gcdt
- [ ] Solution tested carefully in UAT

### 🔄 Regression

No

petertgiles:
I couldn't reproduce the HTTP 500 but I did notice that I never got the toast that my file was done.

<img width="1458" height="836" alt="Image" src="https://github.com/user-attachments/assets/8e3b56d4-6f99-4ce9-bdad-41a404f1393d" />

esizer:
Do we have a stack trace?

petertgiles:
> Do we have a stack trace?

Not that I know of.  I haven't seen anything in the app console logs.  These would come from Reverb anyway, right?  There's nothing there.

[...]

PR #17379, opened 2026-07-10
[Feat] Reverb logging

esizer:
## 👋 Introduction

This is an attempt to see if we can get some errors out of reverb.

## 🕵️ Details

I'm not sure if it is reverb or nginx that is causing our 500s in https://github.com/GCTC-NTGC/gc-digital-talent/issues/17181 but, im hoping this might add some mroe information? :woman_shrugging: 

Just throwing this up to see what people think, feel free to close if this is unecessary or just noise :sweat_smile: 

## 🧪 Testing

Honestly, i'm not sure.

### DIFF
--- a/api/app/Providers/AppServiceProvider.php
+++ b/api/app/Providers/AppServiceProvider.php
@@ -10,6 +10,7 @@ use Illuminate\Support\Facades\Event;
 use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Facades\RateLimiter;
 use Illuminate\Support\ServiceProvider;
+use Laravel\Reverb\Contracts\Logger;
 use PhpOffice\PhpWord\Settings;
 
 class AppServiceProvider extends ServiceProvider
@@ -27,7 +28,7 @@ class AppServiceProvider extends ServiceProvider
     public function boot()
     {
         $this->app->instance(
-            \Laravel\Reverb\Contracts\Logger::class,
+            Logger::class,
             new ReverbLogger(),
         );
 
--- a/api/app/Providers/AppServiceProvider.php
+++ b/api/app/Providers/AppServiceProvider.php
@@ -2,6 +2,7 @@
 
 namespace App\Providers;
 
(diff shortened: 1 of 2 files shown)


petertgiles:
I like the idea but I don't think it will work as-is.  Reverb is a separate process so inside of Azure it can't use the default logger.  Since we're not sure how this will work or what we'll get, I suggest maybe starting off with an on-demand logger to a local file?
https://laravel.com/docs/13.x/logging#on-demand-channels

Maybe once we get the error fixed (whatever it is) we can switch it to Azure logging?
https://github.com/GCTC-NTGC/gc-digital-talent/blob/79969c8e090bb871c6326b7050c84812f06a9fa8/api/config/logging.php#L135

petertgiles:
I think there's a good chance we'll be able to replicate it in dev or uat, so maybe we won't even have to go to prod with on demand logging.

esizer:
> update reverb logger to be local

Thanks! This is the exact feedback i ws looking for. I think I got it in [update reverb logger to be local](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17379/commits/fce55e3b54199e4fdf04be3664dcacc7abc5002e) :sweat_smile:

petertgiles:
> I think I got it in [update reverb logger to be local](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17379/commits/fce55e3b54199e4fdf04be3664dcacc7abc5002e) 😅

Nice!  It looks pretty good to me.  I suggest deploying this branch to dev and seeing what you get it.  If the results are inconclusive then we'll do a formal review to merge it to main?

brindasasi:
@esizer do you mind if I try this PR on dev ?

esizer:
> @esizer do you mind if I try this PR on dev ?

I dont mimd at all! I have been meaning to try it but just haven't gotten around to it :P

esizer:
Includeds in https://github.com/GCTC-NTGC/gc-digital-talent/pull/17668

PR #17387, opened 2026-07-13
[Fix] Full dataset downloads

### LINKED ISSUE #17188: 🐛 Download full dataset still not working for digital community

We've been working to get the downloads more reliable. I believe some fixes were included in the latest release.

And my dream, is to be able to download all the profiles in the digital community at once. Unfortunately the file is still failing to download.

Steps to reproduce:

1. Go to Community talent
2. Filter by the "Digital community"
3. Click "download full dataset"
4. Wait for a few minutes until the "failed" notification appears

Image

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] able to download all the profiles in the digital community at once

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #17188 

## 👋 Introduction

Fixes an issue where full dataset downloads were not working when specifying a community.

## 🕵️ Details

The issue ended up being a typo in the scope map for both user and candidate downloads. The fact nominations were working was the clue. That download does not apply filters. 

We used to have a scope `hasPoolCandidatesInCommunity` but it got renamed to `whereCommunityIn`. You'll notice, neither of these existed so, I think this was never working.

This also noticed we had an issue where this was attemting to flatten them. This also updates the trait method to only fltatten when a scope is notadded to the map.

## 🧪 Testing

1. Start worker `make queue-work`
2. Login as `admin@test.com`
3. Go to `/admin/users`
4. Fitler by community
5. Attempt to download the full dataset
6. Confirm you get the file and only the expected user rows


api/app/Traits/Generator/Filterable.php
```diff
@@ -53,13 +53,13 @@ public function setIds(?array $ids)
...
-    private function flattenFilters($filters): array
+    private function flattenFilters($filters, array $scopeMap): array
     {
         $flattened = [];
         foreach ($filters as $k => $v) {
-            // NOTE: Equity is expected to be assoc array so do not flatten it
-            if (is_array($v) && Arr::isAssoc($v) && $k !== 'equity') {
-                $flattened = array_merge($flattened, $this->flattenFilters($v));
+            // NOTE: keys already in $scopeMap are left alone, not flattened
+            if (is_array($v) && Arr::isAssoc($v) && ! array_key_exists($k, $scopeMap)) {
```
vd1992: One heads-up on this change: the "leave equity alone" behavior now depends on `equity` being listed in the scope map, rather than being hard-coded. That's fine today because both generators that actually set filters include `equity` (and now `community`). But the protection is now implicit — if someone later adds a generator that sets filters and forgets to register an assoc-valued key like `equity` in its scope map, that filter would quietly start getting flattened again and stop working. Might be worth a one-line note here saying that assoc-valued filters have to be in the scope map to survive flattening.

_🤖 Generated by Claude_

api/tests/Feature/Generators/PoolCandidatesExcelTest.php
```diff
@@ -76,6 +79,60 @@ public function testCanGenerateFile(): void
...
+        $generator
+            ->setAuthenticatedUserId($adminUser->id)
+            ->setIds(null)
+            ->setFilters(['community' => ['id' => $community->id]]);
+        $generator->generate()->write();
+
+        // assert: only the candidate inside the filtered community appears (email is column index 13)
+        $rows = $this->readSheetRows($fileName, sheetIndex: 0, rowCount: 10);
+        array_shift($rows); // remove header row
+        $emails = array_column($rows, 13);
```
vd1992: Small thing: the email column index (13) is hard-coded here. If the column order in the generator ever changes, this will quietly read the wrong column. The `assertNotContains` check would likely still pass in that case, so the test could keep giving a green light while actually checking the wrong data. Since you're already reading the header row just above, one option is to look up the email column's position from the headers instead of hard-coding 13.

_🤖 Generated by Claude_

api/tests/Feature/Generators/PoolCandidatesExcelTest.php
```diff
@@ -120,4 +177,40 @@ public function testNeutralizesFormulaInjection(): void
...
         $this->assertStringNotContainsString('<f>', $xml, 'Notes value was written as a formula element');
     }
+
+    /**
+     * Read rows from a sheet in the generated file.
+     * Returns an array of rows, each row being an array of cell values.
+     *
+     * @return array<int, array<int, mixed>>
+     */
+    private function readSheetRows(string $fileName, int $sheetIndex, int $rowCount): array
```
vd1992: This `readSheetRows` helper is an exact copy of the one already in `UserExcelGeneratorTest`. It might be worth moving it into a shared trait or a base test class so the two copies don't drift apart down the road.

_🤖 Generated by Claude_

vd1992 (commented):
## Overview

This fixes full-dataset downloads when a community filter is applied. Two separate things were broken:

1. The scope map pointed `community` at `whereCandidatesInCommunity`, which doesn't exist on either builder, so the filter was silently skipped.
2. `flattenFilters` was flattening the `community => ['id' => ...]` value into a bare `id` key, dropping the community filter entirely.

Both fixes look correct to me. `whereInCommunity` does exist on both `UserBuilder` and `PoolCandidateBuilder`, and each accepts the `['id' => ...]` shape the generators pass. The flatten change also reads as more correct than before: instead of hard-coding `equity` as the one key to leave alone, it now leaves any key that's registered in the scope map intact.

A few things I checked while reviewing:

- `equity` is still protected, since both generators list it in their scope map.
- `NominationsExcelGenerator` (the third user of the trait) never sets filters, so the empty scope map it passes is harmless.
- `poolFilters` is a list rather than an assoc array, so it was never being flattened anyway — no behavior change there.

The new tests are a solid reproduction of the bug. Just a few small, non-blocking notes inline.

_🤖 Generated by Claude_

vd1992 (commented):
The issue mentions "Community talent" not "Users" page, don't know if that makes a big difference.

Also, doesn't seem to quite work right for me. So I am definitely a bit confused
1. Go to `/en/admin/community-talent`
2. Selected a community, I observed results count drop so table successfully filtered
3. Click to download the full dataset, then opened the file
4. File does not have any filtering by community, further more, it has every user (changing sheet to user) present in the spreadsheet

esizer:
> The issue mentions "Community talent" not "Users" page, don't know if that makes a big difference.
> 
> Also, doesn't seem to quite work right for me. So I am definitely a bit confused
> 
> 1. Go to `/en/admin/community-talent`
> 2. Selected a community, I observed results count drop so table successfully filtered
> 3. Click to download the full dataset, then opened the file
> 4. File does not have any filtering by community, further more, it has every user (changing sheet to user) present in the spreadsheet

okay, this mightbe another issue entirely sadly. The issue is, community talent supports filtering by multiple communites but, we changed this to download users and not community talent. The users fitler accepts enitrely different filters so we oinly forward the search filter.

We might need to make a ticket for this one.

api/app/Traits/Generator/Filterable.php
esizer: Honestly, claude suggested this and I was not very enthused but it kept pushing back so i caved :sweat_smile: 

I have reverted this change and went back to my original change in [change fix to filterable scope](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17387/commits/3c3add8bedd2d661d67d9d547eeb2433e122ccc3)

api/tests/Feature/Generators/PoolCandidatesExcelTest.php
esizer: This makes a lot of sense. This verison is maybe still a little flaky but probably better?

[fix and update tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17387/commits/8e5ebbcd632c498f24e7a0037b64456bf3c21cbc)

api/tests/Feature/Generators/PoolCandidatesExcelTest.php
esizer: oh, nice. That makes sense, updated in [fix and update tests](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17387/commits/8e5ebbcd632c498f24e7a0037b64456bf3c21cbc)

petertgiles:
Reviewed in standup
- separate concerns will be handled in a different issue
- the AC of this issue is just that the downloads complete successfully

vd1992 (approved):
Download of full dataset is successful, "Community Talent" page
Other issues noted and to be addressed separately

esizer:
> Download of full dataset is successful, "Community Talent" page Other issues noted and to be addressed separately

Ticket created: https://github.com/GCTC-NTGC/gc-digital-talent/issues/17448

PR #17389, opened 2026-07-13
[Feat] Log forwarded ip

### LINKED ISSUE #16993: ✨ Add last X-Forwarded IP of session to logs
### Details

IT security has recommended we store the "last X-Forwarded IP of session" so that we can use it "for incident response and risk based decisions".

So lets do it... though unfortunately this field was being masked by the F5.

Let's reconfirm if the "X-Forwarded IP field" is available in production (it might be now). If it is available, just log it. If not available, we need to work with IMTD to have them pass it to us.

### ✅ Acceptance Criteria

- [ ] Last X-Forwarded IP is being logged

esizer:
🤖 Resolves #16993 

## 👋 Introduction

Adds a middleware to log `X-Forwaded-For` header for every session.

## 🧪 Testing

> [!IMPORTANT]
> Local testing may not work since you likely are not being forwarded. This will likely need to be deployed to dev to be tested properly.

1. Login as any user
2. Check logs to make sure you see the IP log line

### DIFF
--- a/api/app/Http/Kernel.php
+++ b/api/app/Http/Kernel.php
@@ -6,6 +6,7 @@ use App\Http\Middleware\Authenticate;
 use App\Http\Middleware\EncryptCookies;
 use App\Http\Middleware\PreventRequestsDuringMaintenance;
 use App\Http\Middleware\ProtectedRequest;
+use App\Http\Middleware\RequestOriginLoggerMiddleware;
 use App\Http\Middleware\TrimStrings;
 use App\Http\Middleware\TrustProxies;
 use BeyondCode\ServerTiming\Middleware\ServerTimingMiddleware;
@@ -45,6 +46,7 @@ class Kernel extends HttpKernel
         TrimStrings::class,
         ConvertEmptyStringsToNull::class,
         ProtectedRequest::class,
+        RequestOriginLoggerMiddleware::class,
     ];
 
     /**
--- a/api/app/Http/Middleware/RequestOriginLoggerMiddleware.php
+++ b/api/app/Http/Middleware/RequestOriginLoggerMiddleware.php
@@ -4,6 +4,7 @@ namespace App\Http\Middleware;
 
 use Closure;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Log;
 use Psr\Log\LoggerInterface;
 
 class RequestOriginLoggerMiddleware
@@ -18,6 +19,10 @@ class RequestOriginLoggerMiddleware
     public function handle(Request $request, Closure $next)
     {
         $result = $next($request);
+
+        Log::info('TEMP DEBUG azure test', [
+            'XForwardedIP' => $request->header('X-Forwarded-For'),
+        ]);
 
         if ($request->hasSession()) {
             $xForwardedFor = $request->header('X-Forwarded-For');
--- a/api/app/Http/Middleware/RequestOriginLoggerMiddleware.php
+++ b/api/app/Http/Middleware/RequestOriginLoggerMiddleware.php
@@ -33,6 +33,7 @@ class RequestOriginLoggerMiddleware
             if (! is_null($xForwardedFor) && $xForwardedFor !== $previous) {
                 $this->logger->info('Session IP changed', [
                     'XForwardedIP' => $xForwardedFor,
+                    'UserId' => $user->id,
                 ]);
                 $this->cache->put($cacheKey, $xForwardedFor, now()->addMinutes((int) config('session.lifetime')));
             }
--- a/api/app/Http/Middleware/RequestOriginLoggerMiddleware.php
+++ b/api/app/Http/Middleware/RequestOriginLoggerMiddleware.php
@@ -4,7 +4,6 @@ namespace App\Http\Middleware;
 
 use Closure;
 use Illuminate\Http\Request;
-use Illuminate\Support\Facades\Log;
 use Psr\Log\LoggerInterface;
 
 class RequestOriginLoggerMiddleware
@@ -19,10 +18,6 @@ class RequestOriginLoggerMiddleware
     public function handle(Request $request, Closure $next)
     {
         $result = $next($request);
-
-        Log::info('TEMP DEBUG azure test', [
-            'XForwardedIP' => $request->header('X-Forwarded-For'),
-        ]);
 
         if ($request->hasSession()) {
             $xForwardedFor = $request->header('X-Forwarded-For');
--- a/api/app/Http/Middleware/RequestOriginLoggerMiddleware.php
+++ b/api/app/Http/Middleware/RequestOriginLoggerMiddleware.php
@@ -3,6 +3,7 @@
 namespace App\Http\Middleware;
 
 use Closure;
+use Illuminate\Contracts\Cache\Repository as Cache;
 use Illuminate\Http\Request;
 use Psr\Log\LoggerInterface;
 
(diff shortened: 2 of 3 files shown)


esizer:
@tristan-orourke I have this working in azure but was wondering if you could provide some guidance one point.

Logging it on request was so noisy and drowned out more important logs so I made it only log on changed IPs per session. I think this was suggested from my interpretation of the conversation

tristan-orourke:
This sounds reasonable to me :+1:

brindasasi:
> This sounds reasonable to me 👍

I see this is working during logging in.
But I think CL would still need logging on the requests since they would like to avoid someone stealing the session mid of the app usage , not just during log in

api/app/Http/Middleware/RequestOriginLoggerMiddleware.php
```diff
@@ -0,0 +1,36 @@
...
+    public function __construct(LoggerInterface $logger)
+    {
+        $this->logger = $logger;
+    }
+
+    public function handle(Request $request, Closure $next)
+    {
+        $result = $next($request);
+
+        if ($request->hasSession()) {
```
brindasasi: a Laravel session only exists during the login screen everywhere else it is false. This may not solve ask the CL team ask for.

api/app/Http/Middleware/RequestOriginLoggerMiddleware.php
```diff
@@ -0,0 +1,36 @@
...
+            $xForwardedFor = $request->header('X-Forwarded-For');
+            $previous = $request->session()->get('last_x_forwarded_ip');
```
brindasasi: instead of  "remember the last IP in the session" we may need to use  "remember the last IP for this logged-in user," using something that exists on every request (not just login)  like a small cache entry keyed by user ID

api/app/Http/Kernel.php
```diff
@@ -6,6 +6,7 @@
 use App\Http\Middleware\EncryptCookies;
 use App\Http\Middleware\PreventRequestsDuringMaintenance;
 use App\Http\Middleware\ProtectedRequest;
+use App\Http\Middleware\RequestOriginLoggerMiddleware;
```
brindasasi, line 9: I also think it needs to be plugged in into graphql.php the same spot AuditQueryMiddleware sits

brindasasi (changes requested):
I dug into why this only fires at login, so here's the finding

Why it's login-only: $request->hasSession() only ever returns true on the routes/web.php flow 
(login → auth-callback → refresh). 

The /graphql routes - which is where all actual app usage happens never start a Laravel session at all, since auth there is Bearer-token based, not cookie/session based. 

So right now this middleware is really only watching the sign-in moment, not the ongoing use of a session, which is the case CL specifically wants covered (a stolen/reused session being used mid-app-usage).


**Proposed change:**

- Replace the session-based "last seen IP" with something keyed on the authenticated user, e.g. Cache::get("last_x_forwarded_ip:{$user->id}") instead of $request->session()->get('last_x_forwarded_ip').

- Register the middleware in routes/graphql.php (alongside AttemptAuthentication / AuditQueryMiddleware) instead of globally in Kernel.php, so it actually sits on the traffic that matters.

That keeps the "only log on change" behaviour while covering what the ticket was actually opened for -  detecting an IP change during ongoing app usage, not just at login.

api/app/Http/Kernel.php
esizer: Hmm, I added this to kernel so it happens on every request, even grpahql ones. Do we want this to only run on graphql?

esizer:
@tristan-orourke and @brindasasi As promised, I have found some logging code that seems to attach this header to logs. I'm not sure how to confirm this in azure and im a little uncertain about this. 

From everyrhing I gather, this should be attaching additional informaiton to logs in azure but since I can't confuirm, im unsure if this is sufficient.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/69c9cf9961877804958dfe1b3a045e5bf790085c/api/app/Logging/TbsLoggingStandardProcessor.php#L24

brindasasi:
> @tristan-orourke and @brindasasi As promised, I have found some logging code that seems to attach this header to logs. I'm not sure how to confirm this in azure and im a little uncertain about this.
> 
> From everyrhing I gather, this should be attaching additional informaiton to logs in azure but since I can't confuirm, im unsure if this is sufficient.
> 
> https://github.com/GCTC-NTGC/gc-digital-talent/blob/69c9cf9961877804958dfe1b3a045e5bf790085c/api/app/Logging/TbsLoggingStandardProcessor.php#L24

this is a great find seriously.. let me look into this deeper

api/app/Http/Kernel.php
brindasasi: oh rt .. nevermind ..we need coverage on all..no need to change this

brindasasi:
> > @tristan-orourke and @brindasasi As promised, I have found some logging code that seems to attach this header to logs. I'm not sure how to confirm this in azure and im a little uncertain about this.
> > From everyrhing I gather, this should be attaching additional informaiton to logs in azure but since I can't confuirm, im unsure if this is sufficient.
> > https://github.com/GCTC-NTGC/gc-digital-talent/blob/69c9cf9961877804958dfe1b3a045e5bf790085c/api/app/Logging/TbsLoggingStandardProcessor.php#L24
> 
> this is a great find seriously.. let me look into this deeper

Checked this but  that processor is a dead end for this specific ask, unfortunately. It's only wired into the azure log channel, which is only used by queued jobs/artisan commands , not HTTP requests at all. 
Ordinary web/API traffic goes through LOG_CHANNEL=stack → single, which never touches this processor. 

On top of that, nothing currently logs anything at all for a normal successful request except   slow queries >5s, elevated-role audits, and GraphQL errors.

So I think your middleware is still the right approach.

keep the "only log when it changes" logic , key it by user_id instead of session and include user_id in the log line.

esizer:
> > > @tristan-orourke and @brindasasi As promised, I have found some logging code that seems to attach this header to logs. I'm not sure how to confirm this in azure and im a little uncertain about this.
> > > From everyrhing I gather, this should be attaching additional informaiton to logs in azure but since I can't confuirm, im unsure if this is sufficient.
> > > https://github.com/GCTC-NTGC/gc-digital-talent/blob/69c9cf9961877804958dfe1b3a045e5bf790085c/api/app/Logging/TbsLoggingStandardProcessor.php#L24
> > 
> > 
> > this is a great find seriously.. let me look into this deeper
> 
> Checked this but that processor is a dead end for this specific ask, unfortunately. It's only wired into the azure log channel, which is only used by queued jobs/artisan commands , not HTTP requests at all. Ordinary web/API traffic goes through LOG_CHANNEL=stack → single, which never touches this processor.
> 
> On top of that, nothing currently logs anything at all for a normal successful request except slow queries >5s, elevated-role audits, and GraphQL errors.
> 
> So I think your middleware is still the right approach.
> 
> keep the "only log when it changes" logic , key it by user_id instead of session and include user_id in the log line.

Thanks @brindasasi ! That was my hunch but I wasn't too sure. That makes sense why I was not finding it in the azure logs I was looking at :laughing: 

I think I got all your requests complete in:

 - [attach user Id to log](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17389/commits/f72c04aa676b7de4436d76dbbf2e2a68307aa97d)
 - [update to log based on user auth identifier rather than session](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17389/commits/6ae092bf025d4398f895e1cda9ad1b165ae05603)

brindasasi (approved):
Looking great!  well done⭐️ 🥳

PR #17393, opened 2026-07-13
[Feature] Additional tabs in nomination excel download

### LINKED ISSUE #17108: ✨ Nominations excel download - add additional profile tabs
### Details

The nominations download includes a tab with basic profile info for nominees, but it does not include the additional tabs for career experience and community interests. Here we want to add those tabs to the nominations download.

OCG currently downloads the info in two places, possibly to get around this (though I think they would need it anyway since they want the data for their whole EX population).

Digital community has asked if it could be combined for their nominees, just to save them the trouble of downloading in both places and filtering by nominees. 

### Design file & designer

_No response_

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

1. Add the career experiences tab (#15156) to the nominations download
2. Add the community interest tab (#15164) to the nominations download

### ✅ Acceptance Criteria

- [ ] career experiences tab is in the nominations download for each nominee
- [ ] community interest tab is in the nominations download for each nominee

bella77-69:
🤖 Resolves #17108

## 👋 Introduction

This PR updates the talent nominations excel download to include additional tabs for nominees (Career Experience & Community Interests) 
It also refactors the Excel generators (`UserExcelGenerator` & `NominationsExcelGenerator`) by extracting duplicated logic into reusable traits, reducing code duplication and improving maintainability.

## 🧪 Testing

1. Build `pnpm dev:fresh`
3. Navigate to `/admin/talent-events` and select an event with a nominee
4. In a second terminal run `make queue-work`
5. Download the excel file
6. Confirm:
    - File downloads 
    - Confirm the Career experience tab
    - Confirm the Community interest tab (** if the user is not part of the community there will be no data shown )

## 📸 Screenshot

In the community:
[TestTalentNominationEventactiveEN1_nominations_2026-07-13_194644.xlsx](https://github.com/user-attachments/files/29983027/TestTalentNominationEventactiveEN1_nominations_2026-07-13_194644.xlsx)

Not in the community:
[Past_nominations_2026-07-13_194102.xlsx](https://github.com/user-attachments/files/29983059/Past_nominations_2026-07-13_194102.xlsx)


petertgiles:
Hmm, this looks like a lot duplicated code to me.  From Gray's description it sounds like we want "a tab" to be a block of reusable functionality that can easily be put in different reports.  Do you think it might be possible to factor out the code to use like this?

petertgiles:
I haven't done a deep dive in the code but here are some things flagged by the AI that you could check on.  I haven't checked if any of these are valid, though.

1. **Privacy leak** — buildCommunityInterestRow() never applies the canShare() consent gate that every other tab uses, so non-consenting nominees' PII still gets exported.
2. **Wrong data** — Award experience rows always show is_current = "Yes" (reads a non-existent end_date attribute) and number_of_months as time-elapsed-since-award instead of 0, both confirmed against the correct sibling implementation in UserExcelGenerator.php.
3. **N+1 performance bug** — a per-skill-per-row DB query that's already available via an eager-loaded pivot, multiplying by 7 competencies × 5 experience tables.
4. Plus: a second consent-gating gap (award number_of_months), ~700 lines duplicated from UserExcelGenerator.php instead of extracted into a shared trait, a missing header row on an empty Community Interest sheet, and a missing null fallback on a localized program name.

petertgiles (approved):
Seems to work well!

petertgiles:
The AI is flagging a missing consent check still.  I'm guess this was a pre-existing problem so I'm factoring it out for a later closer check.
https://github.com/GCTC-NTGC/gc-digital-talent/issues/17462

petertgiles:
Since @bella77-69 is on vacation and today is code cutoff I will merge it myself.

PR #17395, opened 2026-07-13
[Bug] Don't disregard the filtering operation

### LINKED ISSUE #17162: Community filter in RecruitmentProcessesPreviewList has no effect
## Bug

In `RecruitmentProcessesPreviewList.tsx`, the community-based filtering on line 90 calls `Array.filter()` but discards the returned array. Since `Array.prototype.filter` returns a new array and does not mutate in place, the community filter is silently ignored.

```tsx
// Line 89-93
if (communityId) {
  recruitmentProcessesFiltered.filter(
    (recruitment) => recruitment.pool.community?.id === communityId,
  );
}
```

## Expected behavior

When a `communityId` is provided, the list should only show recruitment processes belonging to that community.

## Actual behavior

The `communityId` filter has no effect — all qualified recruitment processes are shown regardless of community.

## Suggested fix

Assign the result of `.filter()` back (switching `const` to `let`), or chain the community filter into the initial filter expression on line 81.

vd1992:
🤖 Resolves #17162

## 👋 Introduction

Don't discard the returned filtered array

As far as I could tell, the only place that the optional community id value was set was `TalentNominationGroupProfilePage` which I think matches to 
`/en/admin/talent-events/:eventId/nominations/:groupId/profile`
Which might explain why this wasn't a quickly noticed bug

## 🧪 Testing

1. Not sure what to recommend, assess the above mentioned page?

### DIFF
--- a/apps/web/src/components/RecruitmentProcesses/RecruitmentProcesses.tsx
+++ b/apps/web/src/components/RecruitmentProcesses/RecruitmentProcesses.tsx
@@ -50,7 +50,7 @@ const RecruitmentProcesses = ({
   const recruitmentProcesses = unpackMaybes(
     recruitmentProcessesFragment.poolCandidates,
   );
-  const recruitmentProcessesFiltered = recruitmentProcesses
+  let recruitmentProcessesFiltered = recruitmentProcesses
     ? recruitmentProcesses.filter(
         ({ applicationStatusData }) =>
           applicationStatusData?.candidateStatus?.value ===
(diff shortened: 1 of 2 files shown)


apps/web/src/components/RecruitmentProcesses/RecruitmentProcesses.tsx
vd1992, line 1: While changing `RecruitmentProcessesPreviewList` I checked its parent file and saw the same issue

esizer (approved):
LGTM! 👍

PR #17402, opened 2026-07-14
[Feature] Rendering English and French in the skills table

### LINKED ISSUE #15956: ✨ Link English and French corresponding skills in the skills library.
### Details

Currently skills are listed in either English or French, in separate lists, in alphabetical order (which doesn't match). 
It would be very helpful for community users, who create posters and add skills, to be able to see the skills in both languages at the same time. Many community admins are bilingual and work in both languages. 
It would be beneficial for a skill to be identified visually in both languages.
When we create a skill, we input the information in English and French in the same form. So that link is already registered. 
Also, when we download a list of skills, we see English and French side by side for each skill. This again suggests that we have this link in the background, and hopefully we can bring it forward. 
Thank you very much for looking into it :)

### Design file & designer

_No response_

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

_No response_

### ✅ Acceptance Criteria

_No response_

vd1992:
🤖 Resolves #15956

## 👋 Introduction

Add the requested simultaneous English and French display 

## 🧪 Testing

1. I believe the component is used in two places, public skills library page and the admin version of the table
2. Navigate to it, observe English and French views

#### English

image

#### French

image

### DIFF
--- a/packages/i18n/src/utils/utils.ts
+++ b/packages/i18n/src/utils/utils.ts
@@ -1,3 +1,5 @@
+import { createIntl, createIntlCache, type IntlShape } from "react-intl";
+
 import type { Messages, Locales } from "../types";
 import CommonFrench from "../lang/frCompiled.json" with { type: "json" };
 import { isLocale } from "./localize";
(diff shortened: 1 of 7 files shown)


packages/i18n/src/utils/utils.ts
vd1992, line 1: Requested this helper from Claude, hence the wordy comments. Noticed afterwards this implementation already existed in the file I was looking at so then consolidated stuff

yonikid15 (approved):
LGTM!

PR #17413, opened 2026-07-15
Validate that question response ids belong to user during update

### LINKED ISSUE #17409: 🐛 Question response update/delete on updateApplication not scoped to parent (IDOR)
### Bug

Lighthouse's `NestedOneToMany` resolver uses globally-scoped queries for `update` and `delete` operations on `@hasMany` nested inputs. On `updateApplication`, the `generalQuestionResponses` and `screeningQuestionResponses` inputs accept `update` (by ID) and `delete` (by ID) operations that are **not scoped to the parent application**.

- **delete** calls `GeneralQuestionResponse::destroy($ids)` — a global static delete.
- **update** calls `$model->newQuery()->findOrFail($id)` — a global find, not through the parent relationship.

This means a user who owns Application A (in DRAFT) can update or delete question responses belonging to a different user's Application B, provided they know the response UUID.

The parent `@canFind` directive only authorizes ownership of the application — it does not validate that child IDs belong to it. The existing `UpdateApplicationValidator` covers education requirement experience IDs (#17223) but not question response IDs.

This is the update/delete counterpart to #17405 (which covers the create path). Split from #17261.

### Steps to Reproduce

**Delete scenario:**
1. User A has Application A1 (DRAFT). User B has Application B1 with response R_B.
2. User A calls `updateApplication(id: "A1", application: { generalQuestionResponses: { delete: ["R_B"] } })`.
3. R_B is permanently deleted from User B's application.

**Update scenario:**
1. Same setup. User A calls `updateApplication(id: "A1", application: { generalQuestionResponses: { update: [{ id: "R_B", answer: "tampered" }] } })`.
2. R_B's answer is overwritten on User B's application.

### Proposed Solution

Add validation to `UpdateApplicationValidator` checking that update/delete IDs belong to the parent application:

```php
$applicationId = $this->arg('id');

'application.generalQuestionResponses.update.*.id' => [
    Rule::exists('general_question_responses', 'id')
        ->where('pool_candidate_id', $applicationId),
],
'application.generalQuestionResponses.delete.*' => [
    Rule::exists('general_question_responses', 'id')
        ->where('pool_candidate_id', $applicationId),
],
```

Same pattern for screening question responses.

### Acceptance Criteria

- [ ] `updateApplication` rejects `generalQuestionResponses.update` when the response ID does not belong to the application
- [ ] `updateApplication` rejects `generalQuestionResponses.delete` when the response ID does not belong to the application
- [ ] Same for `screeningQuestionResponses` update and delete
- [ ] Tests cover cross-application access attempts
- [ ] Existing question response functionality is unaffected

tristan-orourke:
🤖 Resolves #17409

## 👋 Introduction

In the `updateApplication` mutation, it was possible to update or delete a screening/general question response which didn't belong to to application, if you knew the full id. This adds validation to ensure the question responses involved belong to the parent application.

## 🕵️ Details

Followed TDD. Regression tests were added first, confirmed to fail, fix was added, then tests pass.
Code was produced with Claude Code.

## 🧪 Testing

The following scenarios should no longer be possible:

**Delete scenario:**
1. User A has Application A1 (DRAFT). User B has Application B1 with response R_B.
2. User A calls `updateApplication(id: "A1", application: { generalQuestionResponses: { delete: ["R_B"] } })`.
3. R_B is permanently deleted from User B's application.

**Update scenario:**
1. Same setup. User A calls `updateApplication(id: "A1", application: { generalQuestionResponses: { update: [{ id: "R_B", answer: "tampered" }] } })`.
2. R_B's answer is overwritten on User B's application.


esizer (approved):
LGTM! 👍 

I don't think this is a blocker but, we may want a human readablemessage for that error code on the frontend as well.

I think the fix is more important that a pretty message on the frontend, I'll leavbe that up to you 😅 

SEE: https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/packages/i18n/src/messages/apiMessages.ts

Aslo, I think that fialing PHPUnit test is unrelated and maybe just a flaky test? :face_with_spiral_eyes:

petertgiles:
Reviewed in standup: both CI failures are false positives

PR #17414, opened 2026-07-15
[Fix] Placement changed status notification rules

### LINKED ISSUE #17256: 🐛 Job placement changes should not trigger notifications

Recruitment team reached out to us. A candidate that had been placed as "Under consideration" in their `job placement` received a notification that something on their application had changed.

Most of the `job placement` options are an internal data tracking tool, most of them have no effects on the candidate status, therefore most of them should NOT send out notifications, when there are no new changes to show the candidate it leads to confusion.

### 🕵️ Details

Send a notification only if `placement_type` changes:

- To `PLACED_INDETERMINATE` 
- From `PLACED_INDETERMINATE` to something else

Otherwise, do not send a notification.

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Changes to job placement no longer send notifications
    - [ ] except for PLACED_INDETERMINATE

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #17256 

## 👋 Introduction

Updates the placement check to only send if the new or original placement was indeterminate.

## 🧪 Testing

> [!IMPORTANT]
> This is likely difficult to manuyally test since you need to be added as a valid user in our GC notify trst, i think :thinking: Or maybe we can test with live keys? There is a PHPUbnit test though so :woman_shrugging: 

1. Login as any user (`admin@test.com` might make it easier to have all roles)
2. Make sure your contact email is one you can access to confirm
3. Start queue (I think emails send on the queue :sweat_smile: anywya, cant hurt ) `make queue-work`
4. Submut an application
5. Change placem,enmt type to indeterminate
6. Confirm emial sent
7. Change placement to anynthing esle
8. Conmfirm emial sent
9. Change placmenet to another placment thaty is _not_ indeterminate
10. Confirm email was _not_ sent


yonikid15 (approved):
Works as expected 👍 nice work

PR #17420, opened 2026-07-15
[Fix] Restore auth login-initiated tracking on the CanadaLogin sign-in page

### LINKED ISSUE #17298: 🐛 Auth tracking in App Insights broken

It looks like the tracking for "GCKey Login Initiated" is not working properly after the CL transition.  We should probably update the name of these metrics anyway.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Rename metrics to not use the "GCKey" name.  Maybe just "Auth"?
- [ ] Start tracking "GCKey Login Initiated" properly again.

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

vd1992:
Resolves #17298

## Introduction

After the site moved to the new CanadaLogin sign-in experience, we stopped recording the analytics event that tells us when someone begins signing in. The tracking had only been attached to the old GCKey sign-in buttons, which are no longer shown to users, so nothing fired on the new screen. This PR moves the tracking into a single shared place that runs regardless of which sign-in screen is displayed, so the metric works again, and renames the event so it is no longer tied to the "GCKey" name.

## Details

The sign-in page ([SignInPage.tsx](apps/web/src/pages/Auth/SignInPage/SignInPage.tsx)) renders two different layouts depending on the `canadaLogin` feature flag. The `appInsights.trackEvent` calls for the login-initiated event previously lived inline on the two "Sign in with GCKey" buttons, which only render in the legacy (flag-off) branch; once `FEATURE_CANADALOGIN` was enabled the flag-on CanadaLogin branch renders instead, and its "Proceed to CanadaLogin" button had no `onClick`, so the event never fired. The fix extracts a single `trackLoginInitiated` handler and wires it to all three sign-in buttons across both branches, so instrumentation can no longer be stranded by a render-branch change. The handler also flushes App Insights before the external redirect so the buffered event is not dropped on unload, and the event was renamed from `GCKey Login Initiated` to `Auth Login Initiated` (with the `gcKeyStatus` property renamed to `loginStatus`).

## Testing

With the CanadaLogin flag enabled, open the sign-in page, click "Proceed to CanadaLogin", and confirm an `Auth Login Initiated` custom event is sent to Application Insights (visible as a request to the App Insights ingestion endpoint in the browser network tab, or in the App Insights portal). Repeat with the flag off to confirm the legacy "Sign in with GCKey" buttons still emit the same event.

## Deployment

No environment variable or database changes are required. However, the analytics event was renamed from `GCKey Login Initiated` to `Auth Login Initiated` — any Application Insights dashboards, workbooks, or alerts that query the old event name will need to be updated to the new name to preserve reporting continuity.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/web/src/pages/Auth/SignInPage/SignInPage.tsx
+++ b/apps/web/src/pages/Auth/SignInPage/SignInPage.tsx
@@ -808,6 +808,7 @@ export const Component = () => {
                 color="primary"
                 utilityIcon={ChevronDoubleRightIcon}
                 external
+                onClick={trackLoginInitiated}
               >
                 {selectedMethod === "canadaLogin"
                   ? intl.formatMessage({
--- a/apps/web/src/pages/Auth/SignInPage/SignInPage.tsx
+++ b/apps/web/src/pages/Auth/SignInPage/SignInPage.tsx
@@ -145,7 +145,7 @@ export const Component = () => {
     );
     // The click triggers a full-page navigation off-site, so flush the buffer
     // to avoid the event being dropped on unload.
-    appInsights.flush();
+    void appInsights.flush();
   };
 
   const InstructionCards = () => {
--- a/apps/web/src/pages/Auth/SignInPage/SignInPage.tsx
+++ b/apps/web/src/pages/Auth/SignInPage/SignInPage.tsx
@@ -1049,23 +1049,7 @@ export const Component = () => {
           mode="solid"
           color="primary"
           external
-          onClick={() => {
-            if (appInsights) {
-              const userId = appInsights.context?.user?.id;
-              appInsights.trackEvent(
-                { name: "GCKey Login Initiated" },
-                {
-                  aiUserId: userId,
-                  pageUrl: window.location.href,
-                  path: window.location.pathname,
-                  timestamp: new Date().toISOString(),
-                  userAgent: navigator.userAgent,
-                  referrer: document.referrer || "none",
-                  gcKeyStatus: "initiated",
-                },
-              );
-            }
-          }}
+          onClick={trackLoginInitiated}
         >
           {intl.formatMessage({
             defaultMessage: "Sign in with GCKey",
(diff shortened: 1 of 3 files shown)


apps/web/src/pages/Auth/SignInPage/SignInPage.tsx
```diff
@@ -785,6 +808,7 @@ export const Component = () => {
                 color="primary"
                 utilityIcon={ChevronDoubleRightIcon}
                 external
+                onClick={trackLoginInitiated}
```
vd1992, line 813: Appears this was the core of the problem, a missing `onClick`
If not for the function consolidation, a one-liner

petertgiles:
Good find!  I recommend either renaming all three metrics or renaming none of them.  Just renaming one seems like a recipe for future confusion.

packages/auth/src/utils/logout.ts
```diff
@@ -89,7 +89,7 @@ function logoutAndRefreshPage({
   if (appInsights) {
     const aiUserId = appInsights?.context?.user?.id || "unknown";
     appInsights.trackEvent?.(
-      { name: "GCKey Logout" },
+      { name: "Auth Logout" },
```
petertgiles, line 92: Could you rename the property on L99 like you did on SignInPage as well?

apps/web/src/pages/Auth/SignInPage/SignInPage.tsx
```diff
@@ -125,6 +125,29 @@ export const Component = () => {
...
+  // Fires when a user initiates sign-in, just before the external redirect to
+  // the IdP. Shared by both the CanadaLogin and legacy GCKey layouts so the two
+  // can't drift apart.
+  const trackLoginInitiated = () => {
+    if (!appInsights) return;
+
+    appInsights.trackEvent(
+      { name: "Auth Login Initiated" },
+      {
+        aiUserId: appInsights.context?.user?.id,
```
petertgiles: Could you please add a fallback like there was before?

petertgiles (changes requested):
I deployed this to Azure and it seems to be working.  :clap:  Just a few small things to tidy up before merging.

The AI reviewer found some stuff related to flushing events but I've spun that off into https://github.com/GCTC-NTGC/gc-digital-talent/issues/17454

packages/auth/src/utils/logout.ts
vd1992: Okay, sure https://github.com/GCTC-NTGC/gc-digital-talent/pull/17420/changes/598a5abb547ad6e7628aadb915f4855746744f9f

apps/web/src/pages/Auth/SignInPage/SignInPage.tsx
vd1992: Yep, can do https://github.com/GCTC-NTGC/gc-digital-talent/pull/17420/changes/598a5abb547ad6e7628aadb915f4855746744f9f

petertgiles (approved):
Thanks!

PR #17457, opened 2026-07-16
[Tooling] Persist migration error logs in Azure between container restarts

### LINKED ISSUE #16961: 🛠️ Migration errors don't reach Azure Monitor
### 🛠️ Platform

We have this gem implemented recently 
https://github.com/GCTC-NTGC/gc-digital-talent/pull/15940

I would like to extend this for migration errors

When post_deployment.sh hangs or crashes during php artisan migrate, there is no persistent record of what went wrong. 
The Azure Log Analytics channel (Custom-AppLogsPrd_CL) is configured in production but is only included in the jobs log stack (queue workers).

 The default stack channel used by artisan commands writes only to the container files, which is wiped on every restart.

During the 2.73.0 release deployment failure incident , the container crashed and restarted 8+ times over 22 minutes, no Slack notification arrived, and post-incident there was no log traces anywhere. 

### 🕵️ Details

**Problem 1 :**
Artisan migrate logs  does not get into Azure Monitor.

**Problem 2:** 
post_deployment.sh migration stdout to Slack.
If the script hangs mid-migration, the output is discarded when the container is killed. 
There is no file-based fallback at all

### 🙋‍♀️ Proposed Solution

I believe we just need to copy this pattern here 
```
 'channels' => array_filter([
                'jobs_file',
                // only try to include the azure logger in the stack if configured
                ! empty(env('AZURE_LOG_INGESTION_ENDPOINT')) ? 'azure' : null,
            ]),
```
into here 
https://github.com/GCTC-NTGC/gc-digital-talent/blob/430a62214ca4f6960a454cc0dc5a136defe12eed/api/config/logging.php#L54

### ✅ Acceptance Criteria

- [ ]  A PHP exception thrown during php artisan migrate appears in Custom-AppLogsPrd_CL in Log Analytics 
- [ ]  post_deployment.sh  writes to /home/LogFiles/post_deployment.log so that we have some back up when the migration is hung or killed

brindasasi:
🤖 Resolves #16961
## 👋 Introduction

Migration errors during deployment didn't reach Azure Monitor and had no persistent record if the container crashed mid-migration — during the 2.73.0 deployment incident, the container crash-looped for 22+ minutes with no Slack notification and no log trace anywhere afterward.

This PR adds a cli log channel that forwards CLI/migration exceptions to Azure Monitor, and a persistent log file (/home/LogFiles/post_deployment.log) that survives container restarts, so a failed deployment leaves a record even if nothing else does.

## 🕵️ Details

`migrate` normally logs to `stack`, which is local-file-only and wiped on restart. 

We didn't add `azure` to `stack` directly since that's also the default for web-request logs, and would send all of that to Azure too. 

`cli` is a separate opt-in channel for exactly this case - used only when `LOG_CHANNEL=cli` is set, currently just the `migrate` step.

`post_deployment.sh` also wraps `migrate` in `config:clear` / `config:cache`, since the config was already cached earlier in the script — without clearing it first, `LOG_CHANNEL=cli` is silently ignored.

### Local

A throwaway commit already exists with a deliberately failing migration on top of this fix: `b9583e9f78`.

1. On the **host**, check out that commit:
   ```bash
   git checkout b9583e9f78
   ```
   If your local API container mounts this repo (e.g. `docker-compose.dev.yml`), the failing migration is now visible inside it immediately. Otherwise, rebuild/restart the container so it picks up the checked-out code.
2. Log into the container and run the migration:
   ```bash
   docker exec -it <api-container> bash
   cd /var/www/html/api
   php artisan migrate --no-interaction --force --no-ansi
   tail -30 storage/logs/laravel.log
   ```
   Expect the migration to fail and a `QueryException` (SQLSTATE 42P01) to show up in `laravel.log`.
3. Clean up: exit the container, then on the host:
   ```bash
   git checkout 16961-migration-error
   ```
   (rebuild/restart the container again if needed), then re-run `php artisan migrate` inside it — expect `Nothing to migrate`.

### Azure

A throwaway tag already exists on this branch with both the fix and a deliberately failing migration: `scratch-cli-log-test-v2` → commit `b9583e9f78`.

1. Azure DevOps → Pipelines → **Run pipeline** → set ref to `scratch-cli-log-test-v2` → deploy the resulting artifact to a dev/review slot.
2. Check the Slack post-deploy notification: the "Database migration" block should say `failed`, with the SQL error text in the code block underneath.
3. Kudu SSH into the slot and check the persistent log:
   ```bash
   cat /home/LogFiles/post_deployment.log
   ```
   Expect a run header plus the migration exception text, even if the container is killed mid-migration.
4. Azure Portal → the app's Log Analytics workspace → **Logs**:
   ```kql
   AppLogsDev_CL   // Custom-AppLogsPrd_CL in prod
   | where TimeGenerated > ago(2h)
   | where * has "this_table_does_not_exist_xyz"
   | order by TimeGenerated desc
   ```
   Expect a row. Without the `config:clear` this PR adds before `migrate` in `post_deployment.sh`, nothing shows up here even though `LOG_CHANNEL=cli` is set — that's the bug being fixed.
5. Redeploy the branch tip (`d56f0d8e39`, no broken migration) to the same slot to leave it clean. Once everyone's done testing, delete the tag: `git push origin --delete scratch-cli-log-test-v2`.

## 📸 Screenshot
**Slack**
Screenshot 2026-07-27 at 10 39 50 AM

**Azure**
image


brindasasi:
yayyy!
image

infrastructure/bin/post_deployment.sh
```diff
@@ -72,9 +79,10 @@ else
     add_section_block ":X: Laravel cache setup *failed*. $MENTION"
 fi
 
-# Laravel database migrations
-MIGRATION_STDOUT=$(php artisan migrate --no-interaction --force --no-ansi)
+php artisan config:clear
```
brindasasi: config:clear is required because the earlier config:cache would otherwise freeze LOG_CHANNEL, silently dropping the cli override below. 

stderr is captured too (exceptions render there) and streamed through tee so the persistent log has partial output even if killed mid-migration

brindasasi (commented):
Moved the rationale for the config:clear/cache bracketing here instead of keeping it as a code comment.

infrastructure/bin/post_deployment.sh
petertgiles: I find it a little weird to run optimize on L75, clear it on L83, then cache again on L86.  Maybe just move the optimize command on L75 below the migrations?  Maybe down to L123 so it's the last Laravel command run?

petertgiles (approved):
This seems to be working well.  It seems a little limited:
- Only the migration step is written to the durable log file, no other steps
- Only migration exceptions are sent to Azure logging, no other output

~I still think the cache/clear/cache pattern is strange.~  Lol, you just changed it, nice.  :laughing: 

That said, I think this is the start of something good and if the scope is sufficient for you for now then let's do it.

PR #17460, opened 2026-07-17
[fix] down migration error

brindasasi:
The `down()` method at api/database/migrations/2026_07_03_173110_add_classification_at_approval.php 
drops the column without first dropping the foreign key constraint. 

You should call `$table->dropForeign(['classification_at_time_of_advancement_approval_id'])` before `dropColumn`.

_This PR applies 1/1 suggestions from code quality [AI findings](https://github.com/GCTC-NTGC/gc-digital-talent/security/quality/ai-findings)._


petertgiles:
I still think this change is unnecessary.  The migration and rollback work fine from my testing.
image

I asked Claude about it and it suggested that this is a MySQL pattern.  That's fine but we abandoned cross-database compatibility a long time ago.
image

PR #17471, opened 2026-07-20
[Fix] Improve tracked users query

### LINKED ISSUE #17468: 🐛 Prod | No candidates found when removing workstream filter

- Julie found it out that the candidate tracking table is taking long time to load and showing no candidates found. 
- In Find matching candidate table, remove the workstream filter, then select some candidates and change their status to 'Referred'. Those candidates will move to `'Candidate Tracking'` table `[total 25 candidates were referred]`, she then show more than 10 candidates by clicking on `'Show 50'`, the table is continuously loading for about `30-35 seconds`, then both tables were empty and show no candidates found message

**Expected outcome:** Referred candidate should be loaded successfully without any performance issue.

### 🕵️ Details

Env: Prod
Version: 2.76.1

### 📋 Steps to Reproduce

Prerequisite: Make sure one talent request is created having multiple candidates [at least more than 10] in it and should be from different work streams 

1. Access to GC digital portal using community admin role
2. Open newly created talent request
3. Verify that there are some candidates in find matching candidate table
4. Click on filters and remove the work stream filter and show results
5. Observe you have more candidates present in the table
6. Select some candidates and apply bulk update on them such as change status to referred
7. Verify those selected candidates are moved to Candidate tracking table
8. On candidate tracking table, click on show 50 to see more than 10 candidates
9. Observe the page [table is loading for 30-35 seconds]
10. Both tables got empty once loading is completed and showing no candidates found

### 📸 Screenshot

- Brinda can add prod logs here.

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Referred candidate should be loaded successfully without any performance issue.

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #17468 

## 👋 Introduction

Fixes an issue with tracked users query being slow.

## 🕵️ Details

There was a hidden n+1 in this query where we loaded the mathcing user rows and then had to go get the sources as well. This did two main things:

1. Replace a real load for a simple exists for the sources so we don't need to hydrate the entire model
2. Update the query to batch load the sub queries to reduce this from n+1 to just 2 queries

## 🧪 Testing

1. Confirm everything still functions
2. Confirm test assertion that only 2 queries are sent
3. Seed many tracked users
4. Select a larger page size
5. Confirm the query loads in a reasonable time


vd1992 (commented):
🤖 _Superseded — I re-posted this as an [inline review](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17471#pullrequestreview-4737891580) so each point sits on the relevant line. See that review for the findings._

api/app/GraphQL/Types/TalentRequestTrackedUser/MatchingSources.php
```diff
@@ -18,9 +21,14 @@ public function __invoke(TalentRequestTrackedUser $trackedUser, array $args, $co
...
 
-        return $trackedUser->user->talentRequestSourceMatches(
-            $info->fieldName,
-            $trackedUser->talentRequest->applicantFilter?->toMatchFilters() ?? []
+        $filters = $trackedUser->talentRequest->applicantFilter?->toMatchFilters() ?? [];
+
+        // Batch every row's lookup for this field into one query (issue #17468).
+        $loader = BatchLoaderRegistry::instance(
+            $info->path,
+            fn () => new MatchingSourcesBatchLoader($info->fieldName, $filters),
```
vd1992, line 29: 🔴 **High — this is the one I'd want addressed before merge.** The loader is keyed on `$info->path`, and Lighthouse runs that through `FieldPath::withoutLists()` (in the vendor `BatchLoaderRegistry`), which strips *every* numeric list index before using it as the cache key — and it only ever builds one instance per key (`$instances[$key] ??= …`).

That's fine for the top-level `talentRequestTrackedUsers(talentRequestId:)` query, where the whole page is a single request. But `matching*Sources` is also reachable under a *list* of requests:

```graphql
talentRequests { data { trackedUsers { matchingQualifiedInPoolSources { id } } } }
```

There the path collapses to one shared key (`talentRequests|data|trackedUsers|matchingQualifiedInPoolSources`) for every tracked user across every request on the page. Whichever row resolves first constructs the loader, and its `applicantFilter` filters get frozen in and applied to everyone — so users tracked under any *other* request end up matched against the wrong request's filter. The old resolver computed each row with its own `$trackedUser->talentRequest` filters, so this is a behavioural regression, not just a perf change.

I couldn't find a frontend query that actually hits this shape today, so it reads as latent rather than live — but it's reachable through the public schema and it fails silently with wrong data, so I'd guard it now. Simplest fix I can see is to fold the request id into the key so loaders are partitioned per request (a UUID survives `withoutLists()`):

```php
$loader = BatchLoaderRegistry::instance(
    [...$info->path, $trackedUser->talent_request_id],
    fn () => new MatchingSourcesBatchLoader($info->fieldName, $filters),
);
```

You keep the batching (one query per request per source), just correctly scoped.

api/app/GraphQL/BatchLoaders/MatchingSourcesBatchLoader.php
```diff
@@ -0,0 +1,70 @@
...
+use Illuminate\Database\Eloquent\Collection;
+use Illuminate\Database\Eloquent\Model;
+
+/**
+ * Batches the per-source match lookups for a page of tracked users into one query per
+ * source, so the tracking list does not run a match query per row (issue #17468).
+ *
+ * One instance is created per GraphQL field path (see BatchLoaderRegistry), so every row's
+ * matching*Sources field for the same list shares this loader. All rows on a page belong to
+ * the same talent request, so they share one set of match filters.
```
vd1992: Tied to the resolver comment: this docblock states the invariant that actually gets violated — *"All rows on a page belong to the same talent request."* That's true for the paginated `talentRequestTrackedUsers` query, but not for `talentRequests { trackedUsers { … } }`, where a single loader instance spans rows from several requests (each with its own filters). Once the keying is fixed, worth rewording this so it doesn't reassure a future reader that the shared-filter assumption is safe.

api/app/GraphQL/BatchLoaders/MatchingSourcesBatchLoader.php
```diff
@@ -0,0 +1,70 @@
...
+     */
+    public function __construct(
+        protected string $relation,
+        protected array $filters,
+    ) {}
+
+    public function load(TalentRequestTrackedUser $trackedUser): Deferred
+    {
+        $userId = $trackedUser->user_id;
+        $this->userIds[$userId] = $trackedUser;
```
vd1992: Nit: `resolve()` only ever reads `array_keys($this->userIds)`, so storing the whole `TalentRequestTrackedUser` model as the value here — and holding a reference to every hydrated row for the life of the loader — isn't buying anything. `$this->userIds[$userId] = true` would do, and the `@var` on line 22 would then actually match what's stored.

api/app/Models/TalentRequestTrackedUser.php
```diff
@@ -148,6 +176,57 @@ public function scopeWithSkillCount(Builder $query): Builder
...
+            'applicantFilter.qualifiedInWorkStreams',
+        ])->find($talentRequestId)?->applicantFilter;
+
+        if (! $filter) {
+            return $query;
+        }
+
+        $filters = $filter->toMatchFilters();
+
+        foreach (TalentRequestSource::selected($filters['talentSources'] ?? null) as $source) {
```
vd1992: 🟠 **Medium — the two `sources()` branches disagree on whether `talentSources` selection matters.** Here (the fast path) you only add `source_exists_*` columns for `TalentRequestSource::selected(...)`, so the paginated list reports *only selected* sources. But the fallback on lines 115–116 calls `talentRequestSources()`, which loops over *all* implemented sources and ignores `talentSources` entirely (`whereMatchesTalentRequest` never reads it). So the same user under the same request can come back differently depending on the path — e.g. with `talent_sources = ['QUALIFIED_IN_POOL']` and a user who matches both sources:

- `talentRequestTrackedUsers(...) { data { sources } }` → `[QUALIFIED_IN_POOL]`
- `talentRequest(id) { trackedUsers { sources } }` → `[QUALIFIED_IN_POOL, AT_LEVEL]`

It's also a change from the pre-PR paginated behaviour (which went through the all-implemented path). Could you confirm which semantics is the intended one and make both branches agree?

api/app/Models/TalentRequestTrackedUser.php
```diff
@@ -148,6 +176,57 @@ public function scopeWithSkillCount(Builder $query): Builder
...
+    public function scopeWithSourceExists(Builder $query, string $talentRequestId): Builder
+    {
+        $filter = TalentRequest::with([
+            'applicantFilter.qualifiedInClassifications',
+            'applicantFilter.qualifiedInWorkStreams',
+        ])->find($talentRequestId)?->applicantFilter;
+
+        if (! $filter) {
+            return $query;
+        }
```
vd1992: 🟠 **Medium — silent fallback to the per-row N+1.** This early return (and the case where `selected()` comes back empty — e.g. `talent_sources = ['ADVANCEMENT']`, only an unimplemented source) means no `source_exists_*` columns get added, so `sources()` drops through to the per-row `talentRequestSources()` branch. And since the schema `@scope` moved from `withTalentRequestMatches` to `withSourceExists`, the user match relations are no longer eager-loaded on that path — so a 500-row page would quietly go back to a query-per-row, which is exactly the timeout this PR is fixing (and a regression vs. the old eager-load for those cases). Worth making sure the list can't silently revert to per-row here.

api/app/Models/TalentRequestTrackedUser.php
```diff
@@ -148,6 +176,57 @@ public function scopeWithSkillCount(Builder $query): Builder
...
+
+        $filters = $filter->toMatchFilters();
+
+        foreach (TalentRequestSource::selected($filters['talentSources'] ?? null) as $source) {
+            $builder = match ($source) {
+                TalentRequestSource::QUALIFIED_IN_POOL => PoolCandidate::query(),
+                TalentRequestSource::AT_LEVEL => CommunityInterest::query(),
+                default => null,
+            };
```
vd1992: 🟡 **Low / maintainability.** Adding a new implemented source now takes three edits that nothing ties together: `matchRelation()` on the enum, the relation in `HasTalentRequestSources`, and this `match`. If someone does the first two but forgets this one, `selected()` will include the source but no column gets emitted — and the fast-path loop just `continue`s past the missing column while `$hasExistsColumns` is already `true`, so the source silently disappears from `sources` on the paginated list (while still showing up on the nested path and in its own `matching*Sources` field). Deriving this mapping from `matchRelation()`, or at least a comment here plus updating the `HasTalentRequestSources` docstring to name all three spots, would de-risk it.

api/tests/Feature/TalentRequestTrackedUserTest.php
```diff
@@ -1859,4 +1860,73 @@ public function testSourcesCorrectOnNestedPathWithoutPaginatedScope(): void
...
+        $batchedLookups = collect($log)
+            ->filter(fn (array $entry) => str_contains($entry['query'], 'from "pool_candidates"')
+                && str_contains($entry['query'], '"user_id" in ('))
+            ->count();
+
+        $this->assertSame(
+            1,
+            $batchedLookups,
+            'Matching pool sources must load in one batched query, not one per row.',
+        );
```
vd1992, line 1908: 🟠 **Medium — coverage gap.** This is a solid guard for the query *count*, but it never asserts the batched results are actually *correct* — a bug that mis-distributed matches across users (including the cross-request one above) would still pass this test. A few gaps worth closing while you're in here: `AT_LEVEL`/`CommunityInterest` batching (the loader serves `matchingAtLevelSources` too and is currently untested end-to-end), the multi-request nested path, `sources` with a `talentSources` subset, and that the fallback path still returns correct `sources`.

api/programmatic-types.graphql
```diff
@@ -2105,6 +2105,7 @@ type __Type {
   kind: __TypeKind!
   name: String
   description: String
+  specifiedByURL: String
```
vd1992, line 2108: Nit: this looks unrelated to the N+1 fix. `specifiedByURL` is a standard graphql-php introspection field, and this is the auto-generated Lighthouse ide-helper file, so it came in from regenerating rather than from your changes (there's no `composer.lock` change in the PR). Harmless, just diff noise — could be split out to keep this PR focused.

vd1992 (commented):
🤖 Claude-generated review (posted through Claude Code). Everything below — this summary and every inline comment — was written by Claude; please sanity-check before acting on it.

I'm re-posting my earlier top-level review as inline comments so each point sits on the code it's about.

Since you asked how deep this went — in technical terms: this wasn't a single read-through. I traced the batch-loader keying down into the Lighthouse vendor source (`BatchLoaderRegistry::instance` → `FieldPath::withoutLists`) to pin down exactly how loader instances are keyed and when `forgetInstances()` runs, then walked the reachable GraphQL paths in the schema (`talentRequests` `@paginate` → `TalentRequest.trackedUsers` `@hasMany` → the `MatchingSources` resolver) and diffed the new batch/exists logic against the pre-PR per-row resolver and the `HasTalentRequestSources` trait to isolate behavioural deltas.

Then I ran the candidate findings through an adversarial-verification pass rather than trusting the first read: each one was independently checked by two reviewers with different lenses — one on Eloquent/query-builder semantics (`getForeignKeyName`, `HasMany` constraints, `addSelect` subqueries, `groupBy` keying), one on GraphQL/Lighthouse execution (Deferred resolution order, loader lifetime, `@field` paths) — both told to try to *refute* the finding, not confirm it. In parallel, five independent sweeps (correctness, GraphQL/Lighthouse, security/authz, performance, style/tests) looked for anything I'd missed. One of my initial candidates (an AT_LEVEL SQL-ambiguity worry) got refuted that way and dropped; the survivors are inline below with severities.

Net: the fix is the right shape, and the path it targets — the single-request `talentRequestTrackedUsers` query — looks correct and should drop to ~2 queries as intended. The issues are all on the *other* paths that now share this code. The one I'd most want resolved before merge is the cross-request filter mix-up on the multi-request nested path (first comment).

api/app/GraphQL/Types/TalentRequestTrackedUser/MatchingSources.php
esizer: Added in [add more context to batch loader key](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17471/commits/b24d62aa2bfb7c5882b373aad835165f3a7e3beb)

api/app/GraphQL/BatchLoaders/MatchingSourcesBatchLoader.php
esizer: I think this was just a comment issue, fixed in [improve batch loader comment and resolve key](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17471/commits/3735cff2cdd832cf66448d89a7367a072e26c90d)

api/app/GraphQL/BatchLoaders/MatchingSourcesBatchLoader.php
esizer: Fixed in [improve batch loader comment and resolve key](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17471/commits/3735cff2cdd832cf66448d89a7367a072e26c90d)

api/app/Models/TalentRequestTrackedUser.php
esizer: Fixed in [unify talent request sources logic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17471/commits/0a49096cc196aa1d16052b4c71c4f45a15e4bc8d)

api/app/Models/TalentRequestTrackedUser.php
esizer: Fixed in [unify talent request sources logic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17471/commits/0a49096cc196aa1d16052b4c71c4f45a15e4bc8d)

api/app/Models/TalentRequestTrackedUser.php
esizer: Fixed in [unify talent request sources logic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17471/commits/0a49096cc196aa1d16052b4c71c4f45a15e4bc8d)

api/programmatic-types.graphql
esizer: This was aut generated from re-writing the schema. Not sdure why it got added but it keeps getting added in so i just comitted it :sweat_smile:

vd1992:
I just aimed Claude at this to help out Yoni in reviewing

yonikid15 (changes requested):
Since Claude is doing such a great job, I decided to double down...

## **EDITED**

### Why this is worth doing, not just patching around

Right now the codebase has **two different mechanisms** solving the same "don't N+1
across a page of tracked users" problem:

1. `matchingQualifiedInPoolSources` / `matchingAtLevelSources` / `matchingAdvancementSources`
   — resolved via a custom field resolver (`MatchingSources.php`) backed by a batch
   loader (`MatchingSourcesBatchLoader.php`). Works correctly for both the paginated
   list *and* the nested `talentRequests { trackedUsers { ... } }` path, because the
   batch loader is keyed per-row by `talent_request_id`.
2. `sources` — a plain Eloquent attribute (`TalentRequestTrackedUser::sources()`) that
   depends on SQL "exists" columns (`scopeWithSourceExists`) being present on the
   loaded row. Those columns only get added on the top-level paginated query (via a
   `@scope` schema directive); the nested path has no way to add them safely, so it
   falls back to slow per-user queries there.

Option B replaces mechanism #2 with mechanism #1 — give `sources` its own resolver +
batch loader, exactly like the three fields next to it already have. Once that's done,
`scopeWithSourceExists`, `sourceExistsColumn()`, and the two-path branching inside
`sources()` all become unnecessary and can be deleted. Net effect: **one** batching
mechanism for all four fields, instead of two, and the nested-path regression goes
away because it's the same code path as the (already correct) other three fields.

This is the more invasive option, but it removes complexity rather than adding it.

---

### Step 1 — Add a resolver-backed batch loader

New file: `api/app/GraphQL/BatchLoaders/SourcesBatchLoader.php`

```php
<?php

namespace App\GraphQL\BatchLoaders;

use App\Enums\TalentRequestSource;
use App\Models\TalentRequestTrackedUser;
use App\Models\User;
use GraphQL\Deferred;

/**
 * Batches the "which sources did this user match" lookup for a page of tracked users
 * into one query per selected source, instead of one query per (row × source).
 * All tracked users passed to one instance must belong to the same talent request.
 */
final class SourcesBatchLoader
{
    /** @var array<string, true> keyed by the tracked user's user_id */
    protected array $userIds = [];

    /** @var array<string, array<int, string>> keyed by user_id, value is matched source names */
    protected array $results = [];

    protected bool $hasResolved = false;

    /** @param  array<string, mixed>  $filters  match filters from the request's applicant filter */
    public function __construct(
        protected array $filters,
    ) {}

    public function load(TalentRequestTrackedUser $trackedUser): Deferred
    {
        $userId = $trackedUser->user_id;
        $this->userIds[$userId] = true;

        return new Deferred(function () use ($userId) {
            if (! $this->hasResolved) {
                $this->resolve();
            }

            return $this->results[$userId] ?? [];
        });
    }

    protected function resolve(): void
    {
        $selected = TalentRequestSource::selected($this->filters['talentSources'] ?? null);

        // one query per selected source (not per row): user_ids that matched it
        $matchedUserIdsBySource = [];
        foreach ($selected as $source) {
            $relation = (new User())->{$source->matchRelation()}();
            $foreignKey = $relation->getForeignKeyName();

            $matchedUserIdsBySource[$source->name] = $relation->getRelated()->newQuery()
                ->whereIn($foreignKey, array_keys($this->userIds))
                ->whereMatchesTalentRequest($this->filters)
                ->whereAuthorizedToView()
                ->pluck($foreignKey)
                ->flip();
        }

        foreach (array_keys($this->userIds) as $userId) {
            $this->results[$userId] = array_values(array_filter(
                array_keys($matchedUserIdsBySource),
                fn (string $sourceName) => $matchedUserIdsBySource[$sourceName]->has($userId),
            ));
        }

        $this->hasResolved = true;
    }
}
```

This runs **one query per selected source** (currently 1–2: `QUALIFIED_IN_POOL`,
`AT_LEVEL`), regardless of page size — the same batching shape as
`MatchingSourcesBatchLoader`, just returning source *names* per user instead of
hydrated records per relation.

---

### Step 2 — Add the field resolver

New file: `api/app/GraphQL/Types/TalentRequestTrackedUser/Sources.php`

```php
<?php

namespace App\GraphQL\Types\TalentRequestTrackedUser;

use App\GraphQL\BatchLoaders\SourcesBatchLoader;
use App\Models\TalentRequestTrackedUser;
use GraphQL\Deferred;
use GraphQL\Type\Definition\ResolveInfo;
use Nuwave\Lighthouse\Execution\BatchLoader\BatchLoaderRegistry;

final class Sources
{
    public function __invoke(TalentRequestTrackedUser $trackedUser, array $args, $context, ResolveInfo $info): Deferred
    {
        $filters = $trackedUser->talentRequest->applicantFilter?->toMatchFilters() ?? [];

        // keyed by field path + talent_request_id, same pattern as MatchingSources,
        // so a single loader instance never spans more than one talent request
        $loader = BatchLoaderRegistry::instance(
            [...$info->path, $trackedUser->talent_request_id],
            fn () => new SourcesBatchLoader($filters),
        );

        return $loader->load($trackedUser);
    }
}
```

This is a near-exact copy of `MatchingSources.php`'s structure — same cache-key
pattern, same reasoning for why it's safe across multiple talent requests.

---

### Step 3 — Rewire the schema field

In `api/graphql/schema.graphql`, the `sources` field on `TalentRequestTrackedUser`
currently reads:

```graphql
sources: [LocalizedTalentRequestSource!]!
    @with(relation: "user")
    @with(relation: "talentRequest.applicantFilter.qualifiedInClassifications")
    @with(relation: "talentRequest.applicantFilter.qualifiedInWorkStreams")
```

Change it to point at the new resolver, keeping the `@with` directives that prevent
the lazy-loading exception on `applicantFilter`'s relations:

```graphql
sources: [LocalizedTalentRequestSource!]!
    @field(resolver: "Sources")
    @with(relation: "talentRequest.applicantFilter.qualifiedInClassifications")
    @with(relation: "talentRequest.applicantFilter.qualifiedInWorkStreams")
```

(`@with(relation: "user")` can drop off this field specifically — the new resolver
doesn't touch `$trackedUser->user`.)

> ⚠️ **Verify before shipping:** confirm how `LocalizedTalentRequestSource` currently
> gets its localized label/value wrapper when the resolver returns plain enum-name
> strings (check how the existing `TalentRequestResult.sources` field — resolved via
> `@method(name: "talentRequestSources")` on `User` — produces the same shape, since
> that one also returns `array<string>` from PHP). It's likely handled automatically by
> the `LocalizedTalentRequestSource` type/directive, but worth a quick check rather than
> assuming.

---

### Step 4 — Clean up what's now dead

Once `sources` resolves through the batch loader for *both* the paginated and nested
paths, these become unused and should be deleted:

- `TalentRequestTrackedUser::scopeWithSourceExists()` (`api/app/Models/TalentRequestTrackedUser.php`)
- `TalentRequestTrackedUser::sourceExistsColumn()` (same file)
- The `sources()` Eloquent attribute itself (same file) — replaced entirely by the resolver
- The `@scope(name: "withSourceExists")` directive on the `talentRequestId` argument of
  the `talentRequestTrackedUsers` query in `schema.graphql`

This is the payoff: the two-path "fast SQL columns vs. slow fallback" branching that's
been the source of every bug in this PR's review history goes away completely, replaced
by one mechanism already proven correct for the other three fields.

---

### Step 5 — Tests to update

- The existing N+1 regression test on the top-level `talentRequestTrackedUsers` query
  (added early in this PR, per the PR description: *"Confirm test assertion that only 2
  queries are sent"*) will need its query-count expectations re-checked — it's now
  hitting the batch loader's per-source queries instead of SQL exists-columns, so the
  *count* may shift even though the *behavior* (no N+1) stays correct.
- `testTrackedUsersListBatchesMatchingAtLevelSources` (added in the most recent commits)
  can be mirrored with a `sources`-specific version, run against **both** the
  top-level `talentRequestTrackedUsers` query and the nested
  `talentRequests { data { trackedUsers { sources } } }` query — proving the fix
  actually closes the gap this whole discussion started from.
- `testSourcesCorrectOnNestedPath` and `testSourcesRespectsTalentSourcesSelection`
  should still pass unchanged (they assert on *values*, not query count), but worth
  re-running to confirm the resolver produces identical output to the old accessor.

---

### Honest scope note

This is a real, contained refactor — two new small classes, one schema line, one
cleanup pass, a few test updates. It's a good half-day-ish task, not a five-minute
patch. Flag it to the PR author as a possible **follow-up PR** rather than something
to squeeze into this one, if #17468's urgency doesn't require the nested path to be
fast right now.

yonikid15 (changes requested):
Nice you resolved everything 👍 

One last thing, via Claude, is we can add a test to prevent any regressions:

Gap:  there's no regression test asserting sources actually batches on the nested path (only `matchingAtLevelSources`/`matchingQualifiedInPoolSources` have one, e.g. `testTrackedUsersListBatchesMatchingAtLevelSources`). Values are covered (`testSourcesCorrectOnNestedPath`), but query count isn't, so nothing in CI would catch a future regression here. Suggested test below:

---

## Suggested test

**File:** `api/tests/Feature/TalentRequestTrackedUserTest.php`

**Where:** right after `testTrackedUsersListBatchesMatchingAtLevelSources` — the last
of the existing batching-regression tests, near the end of the class.

```php
// Regression guard: `sources` must batch on the nested talentRequests { trackedUsers { sources } }
// path too, not just the paginated talentRequestTrackedUsers query.
public function testNestedTrackedUsersListBatchesSources(): void
{
    $request = $this->seedReferredMatchingUsers(25);

    $listQuery = <<<'GRAPHQL'
        query {
            talentRequests {
                data {
                    trackedUsers {
                        sources { value }
                    }
                }
            }
        }
        GRAPHQL;

    DB::enableQueryLog();
    $response = $this->actingAs($this->admin, 'api')->graphQL($listQuery);
    $log = DB::getQueryLog();
    DB::disableQueryLog();

    $trackedUsers = collect($response->json('data.talentRequests.data'))
        ->firstWhere('id', $request->id)['trackedUsers'];
    $this->assertCount(25, $trackedUsers);

    // The batched lookup queries pool_candidates for a set of user ids at once. A per-row
    // regression would instead query one user id at a time ("user_id" = ?).
    $batchedLookups = collect($log)
        ->filter(fn (array $entry) => str_contains($entry['query'], 'from "pool_candidates"')
            && str_contains($entry['query'], '"user_id" in ('))
        ->count();

    $this->assertSame(
        1,
        $batchedLookups,
        'Sources must batch on the nested trackedUsers path too, not one query per row.',
    );
}
```

**Why this shape:** it reuses the existing `seedReferredMatchingUsers(25)` helper
(already in the file, defined right below `testTrackedUsersListBatchesMatchingPoolSources`)
and mirrors the same "count batched `IN (...)` queries" assertion style already used
for the other two batching tests — just pointed at the nested query shape instead of
the paginated one.

yonikid15 (approved):
Thanks for bearing through that review 😄 everything looks good now

PR #17473, opened 2026-07-20
[Feature]  Add missing columns to find candidates table

### LINKED ISSUE #17388: ✨Add missing columns to 'Find candidates' table in requests
### Details

Got some feedback from Julie on the "Find matching candidates" table"

> "it's important to be able to verify at a glance if the candidate is priority or veteran" Proposed solution:
- Add 'Category' column to 'table

> It also helps to know at a glance "Which specific pool" caused this candidate to be 'Qualified in pool', the expectation is that most candidate will only be qualified in one or two pools, we dont expect candidates to be qualified in many. Proposed solution:
- Add 'Qualified pool' column to table
- Comma separated
- Static info only (not clickable)

### Design file & designer

_No response_

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

_No response_

### ✅ Acceptance Criteria

to the 'Find matching candidates' table:
- Add 'Category' column (veteran, priority, citizen, other)
- Add 'Qualified pool' column (comma separated, static info)

yonikid15:
🤖 Resolves #17388 

## 👋 Introduction

- Adds two columns to the "Find matching candidates" table:
  - Adds the category column  (veteran, priority, citizen, other)
  - Adds a "Qualified pool" column, which shows the a comma separated list of matching pools that the candidate is in

## 🧪 Testing

1. Build front-end `pnpm dev:fresh`
2. Login as `admin@test.com`
3. Go to talent requests and find one with user(s) that are qualified in pool
4. Ensure columns are correct

## 📸 Screenshot

image


apps/web/src/pages/TalentRequests/components/TalentRequestMatchesTable/TalentRequestMatchesTable.tsx
```diff
@@ -98,13 +99,29 @@ const TalentRequestMatchingUsers_Query = graphql(/** GraphQL */ `
...
         ...TalentRequestAddReferralDialog
         sources {
           label {
             localized
           }
         }
+        matchingQualifiedInPoolSources {
+          pool {
+            id
+            name {
```
esizer: You might need to talk to @Jerryescandon but I think this should be `displayName` instead? 🤔 

Also, for both, can we prefer `localized` if it makes sense to do so?

esizer (changes requested):

apps/web/src/pages/TalentRequests/components/TalentRequestMatchesTable/TalentRequestMatchesTable.tsx
```diff
@@ -282,6 +299,21 @@ const TalentRequestMatchesTable = ({
...
+        cell: ({
+          row: {
+            original: { user },
+          },
+        }) =>
+          user.priority
+            ? priorityCell(user.priorityWeight ?? 0, user.priority.label, intl)
+            : null,
+      },
+    ),
```
esizer: Sorry, might also want to disable sorting and filtering on this. I don't think its actually doing either so this would just make the column more honest.

apps/web/src/pages/TalentRequests/components/TalentRequestMatchesTable/TalentRequestMatchesTable.tsx
yonikid15: Good catch! Fixed in [Fix pool display name](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17473/commits/85e083ff062db53de19bad26853e69f4975e43c6)

apps/web/src/pages/TalentRequests/components/TalentRequestMatchesTable/TalentRequestMatchesTable.tsx
yonikid15: Nice catch :+1: Fixed in [Refactor priority column](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17473/commits/e10a0982c387ff64aba40ae8f360b4df470bd9b4)

esizer (approved):
LGTM! 👍

PR #17487, opened 2026-07-21
[Fix] Only verified gov employees match at level talent source

### LINKED ISSUE #17465: 🐛 CommunityInterestBuilder.whereMatchesTalentRequest returns non gov employees

`CommunityInterestBuilder->whereMatchesTalentRequest()` can return users who are _not_ a verified government employee.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Go to find talent page, do not change default filters
2. Scroll down to Community Result cards, note the number of At-level matching candidates in Digital Community
3. Log in as community@test.com
4. Go to Community Talent Page, filter table by Digital Community
5. With our current seeders, the number of candidates there is unlikely to match what you saw on Find Talent page.

### 📸 Screenshot

Image

24 candidates in Digital Community, allegedly...

Image

Only one candidate in Community Talent table!

### 🙋‍♀️ Proposed Solution

I think the discrepancy is due to us using the `whereIsVerifiedGovEmployee` scope for the Community Talent table, and not in the find talent results.

The discrepancy is unlikely to be so large in real data, since you can only add Community Interest if you're a verified government employee... but you can _stop_ being a verified employee and leave the Community Interest data hanging around, which is why its so important to use the isVerifiedGovEmployee scope.

Suggested change at bottom of CommunityInterestBuilder.php, whereMatchesTalentRequest:
```
        $this->workStreams(array_column($filters['qualifiedInWorkStreams'] ?? [], 'id'));
        $this->communities($communityId ? [$communityId] : null);
        $this->where('consent_to_share_profile', true);
        $this->whereHas('user', fn (Builder $user) => $user->whereIsVerifiedGovEmployee()); // THIS LINE IS NEW

        return $this;
```

### ✅ Acceptance Criteria

- [ ] the four graphql queries that use TalentRequestMatchFilterInput do NOT return/count non-verified gov employees as part of the AT_LEVEL talent source
- [ ] phpUnit regression tests

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #17465 

## 👋 Introduction

Updates the at level talent source match scope to only include verified government employees.

## 🧪 Testing

> [!NOTE]
> Hard to manually test since you should not be able to create a community interest (requried to match) unless you are verified. This PR is about redundancy.

1. I guess confirm test coverage is good around this or,
2. Remove a matches `work_email_verified_at` by setting it to null to make then unverified
3. Confuirm they no longer appear in the results as an "At-level" match

### DIFF
--- a/api/app/Builders/CommunityInterestBuilder.php
+++ b/api/app/Builders/CommunityInterestBuilder.php
@@ -22,7 +22,8 @@ class CommunityInterestBuilder extends Builder implements TalentRequestMatchable
         $qualifiedInClassifications = $filters['qualifiedInClassifications'] ?? null;
 
         if (! empty($qualifiedInClassifications)) {
-            $this->whereHas('user', function (UserBuilder $userQuery) use ($qualifiedInClassifications) {
+            $this->whereHas('user', function (Builder $userQuery) use ($qualifiedInClassifications) {
+                /** @var UserBuilder $userQuery */
                 $userQuery->whereHas('currentClassification', function (Builder $classQuery) use ($qualifiedInClassifications) {
                     $classQuery->where(function (Builder $q) use ($qualifiedInClassifications) {
                         foreach ($qualifiedInClassifications as $classification) {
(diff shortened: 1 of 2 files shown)


yonikid15 (approved):
LGTM!

PR #17488, opened 2026-07-21
[Fix] Always display nominee's name in excel download

### LINKED ISSUE #16056: 🐛 UAT | Nominee's name present as 'Not Available' in excel download

- When Nominee has not agree to share their information with the community, the firstName and lastName column showing 'Not Available' in the excel download of Talent nominations

Expected outcome: May be a better way to showcase that information rather than 'Not Available'

### 🕵️ Details

Env: UAT

### 📋 Steps to Reproduce

2. Navigate to Talent Management 
3. Open any Talent events and navigate to Nominations
4. Select some or all and click on download excel
5. Open the downloaded excel file
6. Verify that some data specially first name and last name showing as 'Not Available'

### 📸 Screenshot

Image

Image

### 🙋‍♀️ Proposed Solution

- May be better way to show the details when some Nominee is not willing to share their profile data

### ✅ Acceptance Criteria

- [ ] The Nominee is not willing to share their profile, some better way to showcase in excel download

### 🔄 Regression

Yes

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #16056 

## 👋 Introduction

Updates the nominaitons excel download to no longer guard first and last name behind a consent to share profile check.

## 🧪 Testing

2. Make sure a talent managemenr event has some nominations
3. Remove the consent to share profile from one of the users who were nominated
4. Download the excel
5. Confirm you can view first and last name
6. Confirm all other profile information remains redacted

## 📸 Screenshot

swappy-20260721_114258

### DIFF
--- a/api/app/Generators/NominationsExcelGenerator.php
+++ b/api/app/Generators/NominationsExcelGenerator.php
@@ -251,8 +251,8 @@ class NominationsExcelGenerator extends ExcelGenerator implements FileGeneratorI
 
                 $values = [
                     $user->id,
-                    $this->canShare($consentToShare, $user->first_name),
-                    $this->canShare($consentToShare, $user->last_name),
+                    $user->first_name,
+                    $user->last_name,
                     $this->localizeEnum($talentNominationGroup->status, TalentNominationGroupStatus::class),
                     $nominators->join(', '),
                     $optionsStr,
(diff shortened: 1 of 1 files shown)


bella77-69 (approved):
That was an easy one 🤪 LGTM!

PR #17491, opened 2026-07-21
[Fix] Backfill talent sources for talent requests

### LINKED ISSUE #17490: ♻️ Migration old talent requests to select pool candidates talent source
### ♻️ Refactor

With the inclusion of talent sources, we should also do a migration, so that old requests have only one talent source selected: Pool candidates.

### 🕵️ Details

Image

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] All talent requests submitted before 2026-07-24 have the `Pool candidates` value selected in the `Talent source` field

esizer:
🤖 Resolves #17490 

## 👋 Introduction

Backfills the talent source options for past talent requests.

## 🕵️ Details

We recently enabled a new talent source. Unfortunatleyt, old requests were created without this new talernt source so the column was `null`. The issue here is, `null` means "all" talent sources. Since these were created before we had more than one, it unintentinally expanded the candidate pool.

This backfills those to set this valued to `QUALIFIED_IN_POOL` whiwhc was the default assumption before we added talent sources.

## 🧪 Testing

1. Seed a talent request that was created before 2026-07-24 with `null` `talent_sources`
2. Run the command `make artisan CMD="php app:backfill-talent-sources`
3. Conmfirm the talent sources are set to only `QUALIFIED_IN_POOL`
4. Rollback that migration
5. confirm the talernt sources gets reset to null or empty

## :truck: Deployment

SEE: https://github.com/GCTC-NTGC/gc-digital-talent/issues/17533


api/database/migrations/2026_07_21_160904_backfill_talent_sources_for_pre_existing_talent_requests.php
```diff
@@ -0,0 +1,41 @@
...
+    public function down(): void
+    {
+        DB::statement(<<<'SQL'
+            UPDATE applicant_filters
+            SET talent_sources = '[]'
+            WHERE id IN (
+                SELECT applicant_filter_id FROM talent_requests
+                WHERE created_at < '2026-07-24'
+            )
+            AND talent_sources = '["QUALIFIED_IN_POOL"]'::jsonb
```
vd1992: 🤖 _Claude-generated comment — produced with Claude Code (Opus 4.8, 1M context)._

**Where this came from:** surfaced by an adversarial verification workflow — 8 parallel Claude subagents cross-checking a first-pass review. One agent ran this migration against a live `gctalent` database and observed the asymmetry directly: `up()` reported `UPDATE 5`, but `down()` reported `UPDATE 6`.

**The problem — `up()` and `down()` are not inverses.** `up()` selects rows by their *prior* state: it only writes rows whose `talent_sources` was `NULL` or `[]`, deliberately leaving already-populated rows untouched. `down()` instead selects by the *resulting* value — it reverts every pre-`2026-07-24` talent-request filter whose `talent_sources` currently equals `["QUALIFIED_IN_POOL"]`. Nothing distinguishes a value this backfill wrote from one a user genuinely chose: jsonb normalizes both to the identical value, and applicant filters are immutable after creation (the GraphQL relation exposes only `create`). The required source checklist shipped to `main` on 2026-07-17, so real users can submit a request selecting only "Qualified in pool", producing exactly `["QUALIFIED_IN_POOL"]` on a pre-cutoff request. On rollback, `down()` silently rewrites those to `[]`, which `TalentRequestSource::selected()` treats as "all implemented sources" (it adds `AT_LEVEL`) — the opposite of the user's intent, and unrecoverable. The extra 6th row in the live test is exactly one of these genuine user selections that `up()` never touched. This is rollback-only, but the PR's own test plan (steps 4–5) exercises the rollback.

**Suggestion:** make `down()` non-destructive. Either make it an explicit no-op with a comment noting that `up()`'s `NULL`/`[]` → `["QUALIFIED_IN_POOL"]` collapse cannot be cleanly reversed, or capture the ids `up()` actually changed (e.g. into a temp/audit table) and revert only those — so a rollback can never clobber rows the backfill didn't create.

api/database/migrations/2026_07_21_160904_backfill_talent_sources_for_pre_existing_talent_requests.php
```diff
@@ -0,0 +1,41 @@
...
+use Illuminate\Support\Facades\DB;
+
+return new class() extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        DB::statement(<<<'SQL'
```
vd1992: 🤖 _Claude-generated comment — produced with Claude Code (Opus 4.8, 1M context)._

**Where this came from:** flagged by the "migration-conventions" lens of a Claude verification workflow that compared this file against other recent data migrations under `api/database/migrations`.

**The problem — no in-code rationale for a production data change.** This migration mutates production data, but carries no comment explaining why. The repo's other data-mutating migrations consistently include a short `//` note stating the rationale — e.g. `fix_placement_type_lifecycle.php`, `null_assessment_step_when_decided.php`, and `create_talent_requests_table.php`. Here the reasoning — that a `NULL`/`[]` `talent_sources` historically meant "all sources", which unintentionally widened the candidate pool once multiple sources existed — lives only in the PR description, which isn't visible in `git blame` at the migration site. A future reader inspecting this file has no context for why legacy rows are being pinned to `QUALIFIED_IN_POOL`.

**Suggestion:** add a one-line comment above the `up()` statement explaining that an empty/`null` `talent_sources` historically meant "all sources", and this pins pre-feature rows to the original single-source default (`QUALIFIED_IN_POOL`). That matches the established convention in the neighbouring migrations.

yonikid15 (approved):
LGTM!

PR #17498, opened 2026-07-22
[Fix] Invalid session state exception

### LINKED ISSUE #17340: ♻️ Deal witih invalid session state in prod
### ♻️ Refactor

We have an "invalid session state" error that appears fairly frequently in the prod logs.  With mock auth it's very easy to reproduce:
1. Log in
2. Use the browser's back button to return to the "enter the sub" screen
3. Log in a second time

Image

I'm not sure how to reproduce it using CanadaLogin, though.

### 🕵️ Details

_No response_

### 🙋‍♀️ Proposed Solution

If it's a normal occurrence then we probably want to return a HTTP 400-series error 

### ✅ Acceptance Criteria

- [ ] Figure out how to reproduce it
- [ ] If it's a normal occurrence 
    - [ ] don't handle with an exception or error logging
    - [ ] Nice error screen or automatic redirect
- [ ] If it's not a normal occurrence
    - [ ] fix the error

esizer:
🤖 Resolves #17340 

## 👋 Introduction

Updates the invalid session state exception to redirect to logout with an invalid session message.

## 🕵️ Details

While this could represent an attack and is a good guard to have, it can also just be normal user behaviour. So, rather than throwing an exception, we are updating this to be handled more gracefully by redirecting to the login page with the invalid sesssion reason.

## 🧪 Testing

1. Add the new env variable (`OAUTH_LOGGED_OUT_REDIRECT`) from the `.env.example` file to your own `.env` file
2. Refresh `make refresh-api`
3. Login as any user
4. During the login, activate the browser back button
5. Confirm you are redirected to `/logged-out` rather than seeing an exception error page

## 🚚 Deployment

Add the following environment variable:

`OAUTH_LOGGED_OUT_REDIRECT=https://talent.canada.ca/logged-out`

SEE: https://github.com/GCTC-NTGC/gc-digital-talent/issues/17532


petertgiles:
Instead of including deployment instructions inline, please add an issue with the type "Deployment" link to it from here.  You can assign in to Mauli.

api/config/oauth.php
```diff
@@ -89,6 +89,12 @@
     'dev_post_login_redirect' => env('DEV_OAUTH_POST_LOGIN_REDIRECT'),
     'dev_post_login_registration_redirect' => env('DEV_OAUTH_POST_LOGIN_REGISTRATION_REDIRECT'),
 
+    /**
+     * Where is the user redirected to retry logging in when it fails
+     */
+    'logged_out_redirect' => env('OAUTH_LOGGED_OUT_REDIRECT'),
+    'dev_logged_out_redirect' => env('DEV_OAUTH_LOGGED_OUT_REDIRECT'),
```
petertgiles: Does the DEV version actually get used anywhere?

api/app/Http/Controllers/AuthController.php
```diff
@@ -82,10 +82,11 @@ public function authCallback(Request $request)
         $state = $request->session()->pull('state');
         $nonce = $request->session()->pull('nonce');
 
-        throw_unless(
-            strlen($state) > 0 && $state === $request->state,
-            new InvalidArgumentException('Invalid session state')
-        );
+        // Session state does not match or is empty, do not login.
+        if (! (strlen($state) > 0 && $state === $request->state)) {
+            return redirect(config('oauth.logged_out_redirect').'?'.http_build_query(['reason' => 'invalid-session']));
```
petertgiles: OK, I guess we'll need some copy for this now.  I'll post something on desfam.

petertgiles (changes requested):
Looking cool!

api/app/Http/Controllers/AuthController.php
petertgiles: https://talent-cloud.slack.com/archives/CFCE2FSR4/p1784742388767029

api/config/oauth.php
esizer: no, I could read to see if it is the dev server and change the redirect though. I was just following the other patterns. Do you think we 

1. can remove this or,
2. do you want me to implement it so it works in watch mode as well?

api/app/Http/Controllers/AuthController.php
petertgiles: Decision in #desfam is to just reuse the error message for "session expired".

esizer: Cool, updated to use the same message as session expired.

[update message to use session expired](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17498/commits/a8acc7baedefaf574bde22153ee36947b3fc7bbf)

api/config/oauth.php
petertgiles: I don't feel strongly but I'd probably remove it.  It's an exception case so probably not worth putting in special logic for watch mode.

esizer: Remvoed in [remove dev path env var](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17498/commits/2135bc19f8c36dc40728b97ebe22fd7261479dcc)

petertgiles (approved):
Works great!  Thanks for solving this one.

Issue #16203, opened 2026-03-19
♻️ Remove SIC


petertgiles:
### ♻️ Refactor

> [!IMPORTANT]
> Blocked until the transition to CanadaLogin is complete

### ✅ Acceptance Criteria

- [ ] FEATURE_CANADALOGIN flag
- [ ] api/app/Services/SignInCanadaBearerTokenService.php
- [ ] apps/web/src/pages/Auth/RegistrationPages/RegistrationRedirect.tsx
- [ ] api/app/GraphQL/Mutations/VerifyUserEmails.php - CONTACT type
- [ ] apps/web/src/pages/Auth/RegistrationPages/GettingStartedPage.tsx
- [ ] previous registration info page (and associated images, & gckeyMessages)
- [ ] previous sign in info page (and associated images & gckeyMessages)
- [ ] apps/web/src/components/Profile/components/PersonalInformation
- [ ] apps/web/src/pages/Profile/AccountSettings/AccountManagement.tsx
- [ ] documentation/sign-in-canada.md

petertgiles:
Reviewed in refinement:
- Tristan: OK to proceed

PR #17499, opened 2026-07-22
[Refactor] Remove SIC - Update PW tests

petertgiles:
Related to (but does not close) #16203 

## 👋 Introduction

This is the first in a series of "remove SIC" PRs.
This updates the Playwright tests for the new CanadaLogin version of the app.

## 🕵️ Details

1. Flips the feature flag to default true
1. Updates the employee verification section of the make it easier to check the verification checkmarks (and improves a11y)
2. Allows login fixture to provide name and email on login (like CanadaLogin does)
3. Some tests moved around since email verification moved from settings to employee profile
4. Can no longer manually update name or contact email

## 🧪 Testing

- [ ] All tests reliably pass
- [ ] Changes make sense
- [ ] Employee verification section of employee profile still works

### DIFF
--- a/apps/playwright/fixtures/Registration.ts
+++ b/apps/playwright/fixtures/Registration.ts
@@ -1,15 +1,9 @@
-import { execSync } from "child_process";
-
 import type { Locator, Page } from "@playwright/test";
 import { expect } from "@playwright/test";
 
 import { loginBySub } from "~/utils/auth";
-import { deleteUser, me } from "~/utils/user";
-import {
-  fetchIdentificationNumber,
-  generateUniqueTestId,
-  uuidRegEx,
-} from "~/utils/id";
+import { deleteUser } from "~/utils/user";
+import { fetchIdentificationNumber, generateUniqueTestId } from "~/utils/id";
 import graphql from "~/utils/graphql";
 import testConfig from "~/constants/config";
 
@@ -18,14 +12,6 @@ import UserPage from "./UserPage";
 
 const FIELD = {
   GETTING_STARTED_HEADING: "gettingStartedHeading",
-  EMAIL_ADDRESS: "emailAddress",
-  SEND_VERIFICATION_EMAIL_BUTTON: "sendVerificationEmailButton",
-  VERIFICATION_EMAIL_SENT_HEADING: "verificationEmailSentHeading",
-  COUNTER_MESSAGE: "counterMessage",
-  VERIFICATION_CODE: "verificationCode",
-  FIRST_NAME: "firstName",
-  LAST_NAME: "lastName",
-  PREFERRED_CONTACT_LANGUAGE: "preferredContactLanguage",
   SAVE_AND_CONTINUE_BUTTON: "saveAndContinueButton",
   SAVE_AND_CONTINUE_LINK: "saveAndContinueLink",
   ADD_MOST_RECENT_WORK_EXPERIENCE: "addMostRecentWorkExperience",
@@ -44,9 +30,7 @@ export type Field = ObjectValues<typeof FIELD>;
 
 class Registration extends AppPage {
   readonly locators: Record<Field, Locator>;
-  readonly baseUrl: string = "/en/applicant";
   uniqueTestId = generateUniqueTestId();
-  context = graphql.newContext(this.uniqueTestId);
   uniqueEmailAddress = `${this.uniqueTestId}@gc.ca`;
   readonly firstName = "Playwright";
   readonly lastName = "Test";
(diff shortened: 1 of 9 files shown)


vd1992 (commented):
🤖 **Claude Opus-generated review** (Opus 4.8, run by @vd1992) — please verify before acting.

A few test-isolation and cleanup notes from reviewing the CanadaLogin test migration. The one worth blocking on is the shared-`beforeAll` coupling flagged in `employee-profile.spec.ts`; the other two are minor cleanups.

apps/playwright/tests/applicant/employee-profile.spec.ts
```diff
@@ -29,6 +29,30 @@ test.describe("Employee Profile", () => {
     });
   });
 
+  test("Work email removal", async ({ appPage }) => {
```
vd1992, line 34: 🤖 **Claude Opus-generated comment** — please verify before acting.

⚠️ **Destructive test sharing the `beforeAll` fixture user.**

`beforeAll` (line 15) creates a single verified-gov-employee user for the whole `describe`, and this test removes that user's work email — which nulls `work_email_verified_at` and flips `isVerifiedGovEmployee` to `false` for the shared user.

Under the project's `fullyParallel: true`, `beforeAll` runs once **per worker** and tests keep file order within a worker. So whenever the scheduler co-locates this test with a later one in the same worker (near-certain given `workers: "25%"` vs. 6 tests here), it runs first and every later test inherits the de-verified user:
- `"Existing User with Verified Emails"` (line 48) asserts `"Verified"` → gets `"Not provided"`.
- the four employee-profile form tests below may also break if their sections are gated on verified-employee state.

It surfaces as flakiness — `retries: 1` re-runs the victim alone in a fresh worker with a fresh `beforeAll` user, so it "passes on retry." That's why a single green run isn't enough to trust here.

This test used to live in `applicant-settings.spec.ts` under `beforeEach` (fresh user per test); moving it under `beforeAll` reintroduced the shared-state coupling. Please give the removal test its own isolated user — a dedicated `describe` with its own `beforeEach`, or create/login a one-off `sub` inside the test — so it never mutates the fixture the sibling tests read.

apps/playwright/tests/applicant/applicant-settings.spec.ts
```diff
@@ -68,25 +52,10 @@ test.describe("Applicant settings page", () => {
...
     // Verify the 'Green Check mark' is displayed for personal and work email contact card
-    const settingsPage = new AccountSettings(page);
-    await settingsPage.goToSettings();
-    await expect(
-      settingsPage.page.getByRole("img", { name: /verified/i }).last(),
-    ).toBeVisible();
+    const profilePage = new EmployeeProfile(page);
+    await profilePage.goToEmployeeProfile();
+    expect(await profilePage.workEmailVerificationLabel()).toBe("Verified");
     // Update the contact email address and verify throttling message
```
vd1992: 🤖 **Claude Opus-generated comment** — please verify before acting.

Stale comment — the `updateContactEmailAddress()` / throttling / Cancel lines it described were removed, so this no longer refers to any code. Safe to delete. (This test no longer touches the contact email; consider renaming it too, since it's now really a registration + work-email-verified check rather than an "Account Settings update.")

apps/playwright/fixtures/Registration.ts
```diff
@@ -108,30 +113,34 @@ class Registration extends AppPage {
...
-      this.locators[FIELD.VERIFICATION_EMAIL_SENT_HEADING],
-    ).toBeVisible();
-    await this.verifyThrottlingMessageForVerificationCode();
-    const verificationCode = this.getVerificationCode();
-    await this.locators[FIELD.VERIFICATION_CODE].fill(await verificationCode);
-    await this.locators[FIELD.FIRST_NAME].fill(this.firstName);
-    await this.locators[FIELD.LAST_NAME].fill(this.lastName);
-    await this.locators[FIELD.PREFERRED_CONTACT_LANGUAGE]
-      .getByRole("radio", { name: /english/i })
+    await this.locators[FIELD.GOVERNMENT_EMPLOYEE_STATUS]
```
vd1992, line 98: 🤖 **Claude Opus-generated comment** — please verify before acting.

Dead code left over from the SIC verification flow. Now that this method no longer requests/enters a verification code and the last caller (`applicant-settings.spec.ts`) was removed, these are unused on this branch:
- `getVerificationCode()` (line 181) and `verifyThrottlingMessageForVerificationCode()` (line 198)
- the locators they relied on: `EMAIL_ADDRESS` (21), `SEND_VERIFICATION_EMAIL_BUTTON` (22), `VERIFICATION_EMAIL_SENT_HEADING` (23), `VERIFICATION_CODE` (25)

Suggest removing them here, or leaving a note that a later "remove SIC" PR will, so the fixture doesn't carry unused verification plumbing.

apps/playwright/tests/applicant/employee-profile.spec.ts
vd1992: >It surfaces as flakiness — retries: 1 re-runs the victim alone in a fresh worker with a fresh beforeAll user, so it "passes on retry." That's why a single green run isn't enough to trust here.

It using the word "victim" here is funny to me

petertgiles: Good catch!
[new user for removal test](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17499/commits/67bccec1aec8ac85f907daea1de1789d76b4ead4)

apps/playwright/fixtures/Registration.ts
petertgiles: I like it.
[remove dead code](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17499/commits/1ba713566d945d0986a5a940ea38b5b5d7ada4fd)

apps/playwright/tests/applicant/applicant-settings.spec.ts
petertgiles: Makes sense.
[update test name](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17499/commits/81b843491c2be81beeed1b8ca1aa476762ce5936)

apps/playwright/tests/applicant/employee-profile.spec.ts
```diff
@@ -29,6 +31,45 @@ test.describe("Employee Profile", () => {
...
+    await createUserWithRoles(adminCtx, {
+      user: {
+        email: `${sub2}@example.org`,
+        sub: sub2,
+        isGovEmployee: true,
+        workEmail: `${sub2}@gc.ca`,
+        workEmailVerifiedAt: nowUTCDateTime(),
+      },
+      roles: ["guest", "base_user", "applicant"],
+    });
```
vd1992, line 48: Only thing to note. Claude mentioned this user persists after the test which was inconsistent with how we run things in some other places. Which would be deleting things upon test completion.
I don't know how the Playwright in UAT/Azure/something is going. This wouldn't be a issue locally but may be a pain in other environments from users accumulation

vd1992 (approved):
Good! 

Image

apps/playwright/tests/applicant/employee-profile.spec.ts
petertgiles: Yeah, I noticed that too.  None of the tests in this .spec file clean up their user while the ones in, say, registration.spec.ts do.  Cleaning up afterwards is probably desirable but a much bigger change than I was intending.

PR #17502, opened 2026-07-22
[Fix] Slow community interest query

### LINKED ISSUE #17496: 🐛 `CommunityInterestBuilder` match query scales badly on large tracked-users pages

While stress-testing PR #17471 (the fix for the tracked-users N+1, issue #17468) at
larger page sizes, the batched match query for the **AT_LEVEL** source
(`community_interests`) scales very poorly with page size — up to ~8.6 seconds for a
single query at 1000 rows. The **QUALIFIED_IN_POOL** source's equivalent query
(`pool_candidates`) stays fast (under 20ms) across the same range.

Expected: all of the `TalentRequestTrackedUser` match-lookup queries (`sources`,
`matchingQualifiedInPoolSources`, `matchingAtLevelSources`) should scale roughly
linearly (or better) with page size, the way `pool_candidates` already does, so a
large tracked-users page loads in a reasonable time regardless of which sources the
matched users happen to use.

### 🕵️ Details

**Where:** `CommunityInterestBuilder::whereMatchesTalentRequest()` in
`api/app/Builders/CommunityInterestBuilder.php` (~line 25). Classification matching
is done via a nested correlated subquery:

```php
$this->whereHas('user', function (Builder $userQuery) use ($qualifiedInClassifications) {
    $userQuery->whereHas('currentClassification', function (Builder $classQuery) use ($qualifiedInClassifications) {
        // ...
    });
});
```

This is an EXISTS-inside-EXISTS pattern, evaluated per row. It's fine when the query
is scoped to a single user (the pattern the old per-row N+1 code used), but once
PR #17471's batch loader runs it against a large `WHERE user_id IN (...)` list, it
does not scale the way a join would.

By contrast, `PoolCandidateBuilder::whereMatchesTalentRequest()`
(`api/app/Builders/PoolCandidateBuilder.php`, ~line 132) matches classification via
joins/scopes on the pool itself, and stays fast at every scale tested.

**Query execution time** (not query count — see below), measured against an isolated
Postgres test database via `DB::enableQueryLog()`, batched query for each source:

| Rows | `pool_candidates` query (QUALIFIED_IN_POOL) | `community_interests` query (AT_LEVEL) |
|---|---|---|
| 50 | ~5ms | ~4-5ms |
| 500 | ~11ms | ~435-476ms |
| 1000 | ~15-17ms | ~1,800-8,600ms |

**Query count stayed flat** across the same runs, confirming PR #17471's N+1 fix
itself is working correctly — this is a separate, execution-time-only issue:

| Rows seeded | Total SQL queries (paginated `talentRequestTrackedUsers`) | Total SQL queries (nested `talentRequests.trackedUsers`) |
|---|---|---|
| 50 | 20 | 16 |
| 500 | 20 | 15 |
| 1000 | 19 | 16 |

**Environment:** Local, tested against an isolated test database seeded via
`TalentRequestTrackedUser`/`CommunityInterest`/`PoolCandidate` factories, alternating
sources 50/50 across seeded rows.

### 📋 Steps to Reproduce

1. Check out PR #17471 (branch `17468-improve-tracked-users-query`).
2. Seed one `TalentRequest` with ~1000 `TalentRequestTrackedUser` rows, where roughly
   half the underlying users have a matching `CommunityInterest` (AT_LEVEL) and half
   have a matching `PoolCandidate` (QUALIFIED_IN_POOL) for the request's applicant
   filter.
3. Enable the query log (`DB::enableQueryLog()`) and run the
   `talentRequestTrackedUsers` GraphQL query for that request with a page size of
   ~200-1000, requesting `sources`, `matchingQualifiedInPoolSources`, and
   `matchingAtLevelSources`.
4. Inspect `DB::getQueryLog()` for the `community_interests` batched query's `time`
   value.
5. Observe that it takes on the order of seconds, while the equivalent
   `pool_candidates` query for the same page stays under 20ms.

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

Rewrite `CommunityInterestBuilder::whereMatchesTalentRequest()`'s classification
check to match `PoolCandidateBuilder`'s approach — a join against `users` (and
`classifications` via `users.computed_classification`) instead of nested
`whereHas()` calls. That should let Postgres use a single index-backed join instead
of a per-row correlated subquery.

> [!NOTE]
> If a join-based rewrite isn't straightforward given how `currentClassification`
> is modeled (`users.computed_classification` belongsTo `Classification`), an index
> on `users.computed_classification` may be enough on its own to let the existing
> `whereHas` pattern use an index-nested-loop instead of degrading at scale — worth
> checking with `EXPLAIN ANALYZE` before committing to a full rewrite.

### ✅ Acceptance Criteria

- [ ] The `community_interests` batched match query (AT_LEVEL source) scales
      roughly linearly with page size, comparable to the `pool_candidates` query
      (QUALIFIED_IN_POOL source)
- [ ] A `talentRequestTrackedUsers` page of ~1000 rows with a mix of AT_LEVEL and
      QUALIFIED_IN_POOL matches loads in under [X]s (threshold to be agreed with the
      team)
- [ ] Existing `TalentRequestTrackedUserTest` coverage for `sources` /
      `matchingAtLevelSources` values still passes unchanged
- [ ] No change in query *count* (the N+1 fix from #17468/#17471 is preserved)

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #17496 

## 👋 Introduction

Fixes an issue where the query for matching community interests to a talent request was slow.

## 🕵️ Details

The main fix was adding the index to `computed_classification`. I was refactoring the scope to test but that did not do much for us. However, i did leave some of refactoring in since it made it easier to read so, hopefully that is okay :sweat_smile: 

## 🧪 Testing

1. Run migration `make artisan migrate`
2. Seed alot of matching community interests tracked users (1000 or so should be fine)
3. Login as `admin@test.com`
4. Navigate to the candidate tracking tab for that talent request
5. Increase page size of the tracked users and matching candidates components
6. Confirm the query does not time out and completes in a reasonable time (hopeulf under 10s :crossed_fingers: )

### DIFF
--- a/api/app/Builders/CommunityInterestBuilder.php
+++ b/api/app/Builders/CommunityInterestBuilder.php
@@ -20,31 +20,29 @@ class CommunityInterestBuilder extends Builder implements TalentRequestMatchable
     {
         $filters ??= [];
         $qualifiedInClassifications = $filters['qualifiedInClassifications'] ?? null;
-
-        if (! empty($qualifiedInClassifications)) {
-            $this->whereHas('user', function (Builder $userQuery) use ($qualifiedInClassifications) {
-                /** @var UserBuilder $userQuery */
-                $userQuery->whereHas('currentClassification', function (Builder $classQuery) use ($qualifiedInClassifications) {
-                    $classQuery->where(function (Builder $q) use ($qualifiedInClassifications) {
-                        foreach ($qualifiedInClassifications as $classification) {
-                            $q->orWhere(function (Builder $q) use ($classification) {
-                                $q->where('group', $classification['group'])
-                                    ->where('level', $classification['level']);
-                            });
-                        }
-                    });
-                })->whereIsVerifiedGovEmployee();
-            });
-        }
-
         $community = $filters['community'] ?? null;
         $communityId = is_array($community) ? ($community['id'] ?? null) : $community;
 
-        $this->workStreams(array_column($filters['qualifiedInWorkStreams'] ?? [], 'id'));
-        $this->communities($communityId ? [$communityId] : null);
-        $this->where('consent_to_share_profile', true);
-
-        return $this;
+        return $this
+            ->whereHas('user', function (Builder $userQuery) use ($qualifiedInClassifications) {
+                /** @var UserBuilder $userQuery */
+                $userQuery->whereIsVerifiedGovEmployee()
+                    ->when($qualifiedInClassifications, function (UserBuilder $query, array $classifications) {
+                        $query->whereHas('currentClassification', function (Builder $classQuery) use ($classifications) {
+                            $classQuery->where(function (Builder $q) use ($classifications) {
+                                foreach ($classifications as $classification) {
+                                    $q->orWhere(function (Builder $q) use ($classification) {
+                                        $q->where('group', $classification['group'])
+                                            ->where('level', $classification['level']);
+                                    });
+                                }
+                            });
+                        });
+                    });
+            })
+            ->workStreams(array_column($filters['qualifiedInWorkStreams'] ?? [], 'id'))
+            ->communities($communityId ? [$communityId] : null)
+            ->where('consent_to_share_profile', true);
     }
 
     // scope the query to CommunityInterests the current user can view
--- a/api/app/Builders/CommunityInterestBuilder.php
+++ b/api/app/Builders/CommunityInterestBuilder.php
@@ -19,30 +19,40 @@ class CommunityInterestBuilder extends Builder implements TalentRequestMatchable
     public function whereMatchesTalentRequest(?array $filters): self
     {
         $filters ??= [];
-        $qualifiedInClassifications = $filters['qualifiedInClassifications'] ?? null;
         $community = $filters['community'] ?? null;
         $communityId = is_array($community) ? ($community['id'] ?? null) : $community;
 
+        $userIds = $this->atLevelUserIds($filters['qualifiedInClassifications'] ?? null);
+
+        // Match the ids as one Postgres array value, so the number of ids has no limit.
+        $userIdArray = '{'.$userIds->implode(',').'}';
+
         return $this
-            ->whereHas('user', function (Builder $userQuery) use ($qualifiedInClassifications) {
-                /** @var UserBuilder $userQuery */
-                $userQuery->whereIsVerifiedGovEmployee()
-                    ->when($qualifiedInClassifications, function (UserBuilder $query, array $classifications) {
-                        $query->whereHas('currentClassification', function (Builder $classQuery) use ($classifications) {
-                            $classQuery->where(function (Builder $q) use ($classifications) {
-                                foreach ($classifications as $classification) {
-                                    $q->orWhere(function (Builder $q) use ($classification) {
-                                        $q->where('group', $classification['group'])
-                                            ->where('level', $classification['level']);
-                                    });
-                                }
-                            });
-                        });
-                    });
-            })
+            ->whereRaw('community_interests.user_id = any(?::uuid[])', [$userIdArray])
             ->workStreams(array_column($filters['qualifiedInWorkStreams'] ?? [], 'id'))
             ->communities($communityId ? [$communityId] : null)
             ->where('consent_to_share_profile', true);
+    }
+
+    // Ids of verified gov employees whose current classification is one of the requested pairs.
+    private function atLevelUserIds(?array $qualifiedInClassifications)
+    {
+        return User::query()
+            ->whereIsVerifiedGovEmployee()
+            ->when($qualifiedInClassifications, function (Builder $query, array $classifications) {
+                /** @var UserBuilder $query */
+                $query->whereHas('currentClassification', function (Builder $classQuery) use ($classifications) {
+                    $classQuery->where(function (Builder $q) use ($classifications) {
+                        foreach ($classifications as $classification) {
+                            $q->orWhere(function (Builder $q) use ($classification) {
+                                $q->where('group', $classification['group'])
+                                    ->where('level', $classification['level']);
+                            });
+                        }
+                    });
+                });
+            })
+            ->pluck('id');
     }
 
     // scope the query to CommunityInterests the current user can view
--- a/api/app/Builders/CommunityInterestBuilder.php
+++ b/api/app/Builders/CommunityInterestBuilder.php
@@ -22,7 +22,7 @@ class CommunityInterestBuilder extends Builder implements TalentRequestMatchable
         $community = $filters['community'] ?? null;
         $communityId = is_array($community) ? ($community['id'] ?? null) : $community;
 
-        $userIds = $this->atLevelUserIds($filters['qualifiedInClassifications'] ?? null);
+        $userIds = $this->atLevelUserIds($filters);
 
         // Match the ids as one Postgres array value, so the number of ids has no limit.
         $userIdArray = '{'.$userIds->implode(',').'}';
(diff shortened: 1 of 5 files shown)


api/app/Builders/CommunityInterestBuilder.php
```diff
@@ -20,31 +20,29 @@ public function whereMatchesTalentRequest(?array $filters): self
...
+        return $this
+            ->whereHas('user', function (Builder $userQuery) use ($qualifiedInClassifications) {
                 /** @var UserBuilder $userQuery */
-                $userQuery->whereHas('currentClassification', function (Builder $classQuery) use ($qualifiedInClassifications) {
-                    $classQuery->where(function (Builder $q) use ($qualifiedInClassifications) {
-                        foreach ($qualifiedInClassifications as $classification) {
-                            $q->orWhere(function (Builder $q) use ($classification) {
-                                $q->where('group', $classification['group'])
-                                    ->where('level', $classification['level']);
+                $userQuery->whereIsVerifiedGovEmployee()
```
vd1992: **This makes `whereIsVerifiedGovEmployee()` unconditional** — it previously applied only when `qualifiedInClassifications` was non-empty. So when a request has no classification filter, AT_LEVEL now additionally requires `computed_is_gov_employee = true AND work_email IS NOT NULL AND work_email_verified_at IS NOT NULL`.

That's reachable and drops real rows: `work_email_verified_at` is independent state (creating a GC work experience only *prompts* verification, it doesn't set it), so "gov employee, verification pending" users fall out — as do the `atLevelCount`/`count` values, since `CountTalentRequestMatchesByCommunity` runs the same scope.

I suspect this is intended. If so, please pin it with a test (see the test-file comment). If not, move the call back inside the `when($qualifiedInClassifications, …)` closure.

api/database/migrations/2026_07_22_150810_add_index_to_users_current_classification.php
```diff
@@ -0,0 +1,22 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->index('computed_classification');
```
vd1992, line 12: Filename/class says `current_classification`, but the column is `computed_classification` (`currentClassification` is the relation, not a column). Worth renaming so it doesn't mislead later.

api/tests/Feature/TalentRequestMatchesTest.php
```diff
@@ -1122,7 +1122,11 @@ public function testAtLevelSourceMatchesUserWithCommunityInterest(): void
     {
         $community = Community::factory()->create();
 
-        $user = User::factory()->create();
+        $user = User::factory()->create([
```
vd1992, line 1125: These keep the assertions green but mask the change rather than assert it — this test previously passed with a plain, non-verified user, which is the evidence that the AT_LEVEL row set changed. Worth a dedicated test that pins the new behaviour: a consenting, non-verified community-interest user should be excluded when there's no `qualifiedInClassifications`.

vd1992 (commented):
> 🤖 **Generated by Claude (Opus 4.8) at a maintainer's request.** I couldn't run the DB here, so treat the perf note as reasoning, not measurement.

**One thing worth resolving before merge:** the refactor makes `whereIsVerifiedGovEmployee()` unconditional — it was previously gated behind `! empty($qualifiedInClassifications)`. So whenever a request has no classification filter, AT_LEVEL now returns fewer rows (and `atLevelCount`/`count` in `CountTalentRequestMatchesByCommunity` drop with it). I suspect requiring a verified gov employee is the *intended* semantics — the ask is just to make it deliberate: note it in the PR and add a test that pins it, rather than the fixture edits that only keep the old assertions green. Details inline.

Minor: the migration filename says `current_classification`, but the column is `computed_classification` — worth a rename.

Index choice looks good (no prior index on that column). 👍

yonikid15:
https://claude.ai/code/artifact/06441725-36c0-4700-8187-ffadebd72973?via=auto_preview

esizer:
> https://claude.ai/code/artifact/06441725-36c0-4700-8187-ffadebd72973?via=auto_preview

Cannot reproduce, I seed 1000 users, ran the query in 2.195ms with an index scan and logged the sql to analyze.

```SQL
Nested Loop Semi Join  (cost=0.41..50.58 rows=1 width=681) (actual time=0.047..2.153 rows=143 loops=1)
  Buffers: shared hit=4779
  ->  Nested Loop Semi Join  (cost=0.41..49.49 rows=1 width=681) (actual time=0.045..2.020 rows=143 loops=1)
        Buffers: shared hit=4636
        ->  Seq Scan on community_interests  (cost=0.00..40.95 rows=1 width=681) (actual time=0.005..0.244 rows=1000 loops=1)
              Filter: (consent_to_share_profile AND (community_id = 'e3736451-b7ec-4e25-afc3-32ab8aeb6d82'::uuid))
              Buffers: shared hit=36
        ->  Nested Loop  (cost=0.41..8.53 rows=1 width=16) (actual time=0.002..0.002 rows=0 loops=1000)
              Buffers: shared hit=4600
              ->  Index Scan using users_pkey on users  (cost=0.28..8.29 rows=1 width=32) (actual time=0.001..0.001 rows=1 loops=1000)
                    Index Cond: (id = community_interests.user_id)
                    Filter: (computed_is_gov_employee AND (work_email IS NOT NULL) AND (work_email_verified_at IS NOT NULL) AND (deleted_at IS NULL))
                    Rows Removed by Filter: 0
                    Buffers: shared hit=3000
              ->  Index Scan using classifications_pkey on classifications  (cost=0.14..0.19 rows=1 width=16) (actual time=0.001..0.001 rows=0 loops=800)
                    Index Cond: (id = users.computed_classification)
                    Filter: ((deleted_at IS NULL) AND (("group")::text = 'FM'::text) AND (level = 1))
                    Rows Removed by Filter: 1
                    Buffers: shared hit=1600
  ->  Seq Scan on communities  (cost=0.00..1.07 rows=1 width=16) (actuoops=143)
        Filter: (id = 'e3736451-b7ec-4e25-afc3-32ab8aeb6d82'::uuid)
        Rows Removed by Filter: 1
        Buffers: shared hit=143
Planning:
  Buffers: shared hit=21
Planning Time: 0.341 ms
Execution Time: 2.195 ms

Total: 4,779 buffer reads, 2.195 ms execution time, for 1,000 seeded rows.
```

esizer:
@yonikid15 okay, I made even more changes. I think I was testing something incorreclty which lead to such a large discrepancy. I tested every single path that uses this query and think we have something reasonable now.

- **Total count**: 5.9 ms -> 6.0 ms
- **Count per community**: 43.6 ms -> 5.7 ms
- **Match list**: 621.4 ms -> 19.1 ms

yonikid15 (approved):
AWESOME WORK 🎉 You improved it better than Claude did 😆

yonikid15:
@tristan-orourke when query at level users, we want to apply the same filters (e.g. filtered by diploma/equity/language/operational-requirement/skills/location filters) to them as qualified in pool sources, correct?

See https://github.com/GCTC-NTGC/gc-digital-talent/pull/17502/changes/87a308c7b7cfb2e49724a4c5aad804164ddfd16f

brindasasi:
@esizer @vd1992 is this still in works? or can be merged in ?

esizer:
> @esizer @vd1992 is this still in works? or can be merged in ?

Just waiting on a reply from @tristan-orourke

tristan-orourke:
> @tristan-orourke when query at level users, we want to apply the same filters (e.g. filtered by diploma/equity/language/operational-requirement/skills/location filters) to them as qualified in pool sources, correct?
> 
> See [87a308c](https://github.com/GCTC-NTGC/gc-digital-talent/commit/87a308c7b7cfb2e49724a4c5aad804164ddfd16f)

Yep, that's right.

PR #17503, opened 2026-07-22
[Fix] Hide talent source input

### LINKED ISSUE #17494: 🐛 Pre-select and hide talent source in Find Talent page

Seems we missed the mark on the Talent Source feature. While DCM wants the ability to see community candidates inside a talent request, they do NOT want users to be able to pick and choose where they want their talent source from.

For now, we will leave this only accessible for request responders only.

WHY?

> Most people in the community are bad at filling their profiles, and since there is no verification method, the amount of people that appear as "estimated candidates" is severely inflated, therefore estimate is really bad. (According to DCM this is CURRENTLY an issue that they are actively solving by keeping an additional spreadsheet of their own validated people)

> Managers will opt-in for the perceived "faster" route "Oh getting a GC employee is obviously better and faster, why would I ever choose from pools when community is an option". This compounded with bad estimates makes problem worse. DCM foresees that they will get ONLY requests for internal people. Resulting in then having to convince managers that there are in fact NO people from the community that match, and then convince them to instead consider pool qualified people. Which is simply adding more friction to their workflow.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Talent source is preselected to be "Pool candidates"
- [ ] Talent source is hidden from the find talent page
- [ ] Hide it from the summary in the submission form

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #17494 

## 👋 Introduction

Hides the talent source selection and hardcodes the value to the "Qualified in pool" talent source only.

## 🕵️ Details

The users indicated that while having community interests can be helpful, its not exactyl right and don't need to give the users the ability to select one. This hardcodes the talent source so that the new talent source is only used by the discretion of the recruiter.

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Navigate to `/search`
3. Confirm talent sources input no longer exists
4. Confirm request gets subnmitted with only qualifed in pools talent source
5. Confirm no references to community talent/users/at level appears on the search page

### DIFF
--- a/apps/playwright/fixtures/TalentSearch.ts
+++ b/apps/playwright/fixtures/TalentSearch.ts
@@ -67,8 +67,6 @@ class TalentSearch extends AppPage {
   ) {
     const poolCard = await this.poolCardVisibility(poolName);
 
-    await this.page.getByRole("checkbox", { name: /pool candidates/i }).click();
-
     const selectedClassification = classification.groupAndLevel;
     const classificationFilter = this.page.getByRole("combobox", {
       name: /classification/i,
--- a/apps/web/src/components/SearchRequestFilters/SearchRequestFilters.tsx
+++ b/apps/web/src/components/SearchRequestFilters/SearchRequestFilters.tsx
@@ -1,4 +1,4 @@
-import { useIntl, type MessageDescriptor } from "react-intl";
+import { useIntl } from "react-intl";
 
 import {
   notEmpty,
@@ -18,10 +18,7 @@ import type {
   LocalizedEnumString,
   LocalizedTalentRequestSource,
 } from "@gc-digital-talent/graphql";
-import {
-  FlexibleWorkLocation,
-  TalentRequestSource,
-} from "@gc-digital-talent/graphql";
+import { FlexibleWorkLocation } from "@gc-digital-talent/graphql";
 
 import { getShortPoolTitleHtml } from "~/utils/poolUtils";
 import { wrapAbbr } from "~/utils/nameUtils";
@@ -44,7 +41,6 @@ const ApplicantFilters = ({
   applicantFilter,
   selectedClassifications,
   flexibleWorkLocationOptions,
-  talentSourceOptions = [],
 }: {
   applicantFilter?: PartialApplicantFilter | null;
   selectedClassifications?: (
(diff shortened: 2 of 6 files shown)


yonikid15 (changes requested):
Nice work so far 👍 

I think we still need to change a couple things (need to double check with @Jerryescandon):

1. Remove the talent source from the selected filters on the request page

Image

---

2. By default the applicantFilters should include the TalentRequestSource.AT_LEVEL on submit (I think this is the quickest way to fix this issue)

esizer:
> Nice work so far 👍
> 
> I think we still need to change a couple things (need to double check with @Jerryescandon):
> 
> 1. Remove the talent source from the selected filters on the request page
> 
> Image
> 2. By default the applicantFilters should include the TalentRequestSource.AT_LEVEL on submit (I think this is the quickest way to fix this issue)

@Jerryescandon can you hlep out here? I thought we said we were only updating the search page and leaving the admin alone?

Jerryescandon:
> > Nice work so far 👍
> > I think we still need to change a couple things (need to double check with @Jerryescandon):
> > 
> > 1. Remove the talent source from the selected filters on the request page
> > 
> > Image
> > 2. By default the applicantFilters should include the TalentRequestSource.AT_LEVEL on submit (I think this is the quickest way to fix this issue)
> 
> @Jerryescandon can you hlep out here? I thought we said we were only updating the search page and leaving the admin alone?

this isnt admin, this is the submission form when you hit the request pools button, we should also hide it here

yonikid15 (approved):

yonikid15 (changes requested):
Need to remove the input from playwright

esizer:
> fix test

Removed in [fix test](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17503/commits/c54e5f9c13ef74d0b37f96e490db936b8b8d9b6d) Should pass now :crossed_fingers:

yonikid15 (approved):
Noice

PR #17531, opened 2026-07-23
[Fix] Talent request tracked users flaky test

### LINKED ISSUE #17530: 🐛 Talent request tracked user test flaky

Seems like this is failing about 15% of the time.

### 🕵️ Details

https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/30004956202/job/89198661729#step:7:53

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Test passes consistently

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #17530 

## 👋 Introduction

Fixes an issue where the test was failing somer of the time.

## 🕵️ Details

We had random talent sources but, were using pool candidates to run the tests. So, if they random sources did not include `QUALIFIED_IN_POOL` then the test would fail. This hardcodes that talent source to keep the test consistent.

## 🧪 Testing

1. Run the test atleast 30 times (doesnt fail too often)
2. Confirm all runs pass

### DIFF
--- a/api/tests/Feature/TalentRequestTrackedUserTest.php
+++ b/api/tests/Feature/TalentRequestTrackedUserTest.php
@@ -1911,7 +1911,9 @@ class TalentRequestTrackedUserTest extends TestCase
     private function seedReferredMatchingUsers(int $count): TalentRequest
     {
         $classification = Classification::factory()->create();
-        $filter = ApplicantFilter::factory()->for($this->community)->create();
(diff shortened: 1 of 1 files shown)


yonikid15 (approved):
LGTM 👍

PR #17545, opened 2026-07-23
[Feature] Add logging of special applications

### LINKED ISSUE #17449: ✨ Add special application events to activity log
### Details

Add event to process activity log when:
- Special application is created
- Special application is submitted

With the following information when created
- Name of user creating the application
- Special application type
- Special application deadline date
- Name of user who will be applying/added to the process

With the following information when submitted
- Name of user creating the application
- Special application type
 

### Design file & designer

[Figma file](https://www.figma.com/design/fn6LXNySdE5irUbvky9WRM/Activity-log---Pools--Communities-?node-id=91-3248&t=ut7tQiZCnDhaMuY8-1) by @Jerryescandon 

### 📸 Screenshots

Image

### 🌎 Localization

_No response_

### Proposed Implementation

_No response_

### ✅ Acceptance Criteria

- [ ] Even is added to activity log when a special application is created
   - [ ] Creator name
   - [ ] Type
   - [ ] Extended deadline
   - [ ] Applicant name
- [ ] Event is added to activity log when a special application is submitted
   - [ ] Applicant name
   - [ ] Type

vd1992:
🤖 Resolves #17449

## 👋 Introduction

Have the process activity log also handle special application creation + submission in the design format

## 🧪 Testing

1. Create a special application
2. Submit that special application
3. Navigate to that processes' activity log 

### Logging 

image

### Filter options

image

### Filtered

image


api/app/GraphQL/Mutations/CreateSpecialApplication.php
```diff
@@ -30,8 +33,20 @@ public function __invoke($_, array $args)
...
+                    'special_application_justification' => $poolCandidateInput['special_application_justification'],
+                    'special_application_closing_date' => $poolCandidateInput['special_application_closing_date'],
+                ]);
+            });
+
+            $existingPoolCandidate->logActivity(ActivityEvent::SPECIAL_APPLICATION_CREATED, [
+                'user_id' => $userId,
                 'special_application_type' => $poolCandidateInput['special_application_type'],
+                'special_application_type_en' => $specialApplicationLocalizedString['en'],
+                'special_application_type_fr' => $specialApplicationLocalizedString['fr'],
```
vd1992: How to approach being able to render the log contents in English and French was something I had to scratch my head about for a bit. Of three strategies considered, this is the one I liked more and stuck with. The other two

1. Frontend localization. What we used to do, mapping enums to a frontend collection of translation. Decided against as we phased out that approach and it has the downside of two sets of translations pointed at one enum grouping
2. Pull down enums from API on the base page to funnel to relevant components then match. This just seemed like it would be messy, hard to follow, especially if more enums came into play for the activity page

api/lang/fr/activity_event.php
```diff
@@ -13,4 +13,6 @@
     'reinstated' => 'Réintégré',
     'reverted' => 'Restauré',
     'published' => 'Publié',
+    'special_application_created' => 'Abc',
+    'special_application_submitted' => 'Def',
```
vd1992: TODO: Add translations

api/app/Models/PoolCandidate.php
```diff
@@ -815,9 +815,22 @@ public function submit(?string $signature)
...
+        $isSpecialApplication ?
+                $this->logActivity(ActivityEvent::SPECIAL_APPLICATION_SUBMITTED, [
+                    'signature' => $signature,
+                    'special_application_type' => $this->special_application_type,
+                    'special_application_type_en' => $specialApplicationLocalizedString['en'],
+                    'special_application_type_fr' => $specialApplicationLocalizedString['fr'],
+                ]) :
+                $this->logActivity(ActivityEvent::SUBMITTED, [
+                    'signature' => $signature,
+                ]);
```
esizer: I'm not seeing any usage of the result of this so we could avoid assigning this to a variable and just use a simple if/else statement.

api/app/GraphQL/Mutations/CreateSpecialApplication.php
esizer: hmm, we typically do not log the full enum here. The current log has activtyMessages. Did you look into that for this? 

https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/apps/web/src/messages/activityMessages.ts

api/app/GraphQL/Mutations/CreateSpecialApplication.php
esizer, line 1: This file contains a lot of duplicate values in the arrays. Are they always intended to be the same? If so, it might be worth assinging the duplicated values to a single array and jujst spreading it as needed intot he other arrays.

esizer (changes requested):

api/app/GraphQL/Mutations/CreateSpecialApplication.php
vd1992: What I am doing is plugging the value into the boxed area pictured below. 
I need to refer to the English and French of `special_application_type.php`
Duplicating the copy in the frontend is what I decided against in bullet 1

image

esizer: oh, our activty log was not setup for this yet :sweat_smile: thats why you found it so odd tyo do. If you look at the others, we just listt the keys. Hmm, okay :thinking:

vd1992: Yeah, so I did so thinking about a good approach. Storing a few enum labels seems fine, simple, and won't break. The other ideas I mused on above 
The others indeed just list what the touched field was, not a localized value

esizer: okay, that seems  good then. One suggestion though, I think this is duplicating a similar shape. We could just do:

```php
[
  'special_application_type' => SpecialApplicationType::localizedString(...)
]
```

Which would reuslt in:

```json
{
  "special_application_type": {
     "value" : "...",
     "en": "...",
     "fr": "...",
   }
}
```


But honestly, im tno too concered about this so its up to you if you want to do this :stuck_out_tongue:

vd1992: I think I will stick with as is to keep the nice and simple single layer. No nesting
Hmm, I feel we should avoid introducing nesting. Makes it simpler when accessing properties elsewhere when you can expect everything is a key that maps to a string

api/app/Models/PoolCandidate.php
vd1992: 🏈 [`c4feaf2a48` — Simplify special-application submit logging with if/else](https://github.com/GCTC-NTGC/gc-digital-talent/commit/c4feaf2a48f930b63d06f5be61147519e6556206), brought to you by **Brock Purdy** — no wasted motion: he never stashes a result he's never going to use, he just reads it and takes the clean if/else.

<sub>🤖 Claude Opus 4.8 (1M context)</sub>

api/app/GraphQL/Mutations/CreateSpecialApplication.php
vd1992: 🏈 [`7167a118ef` — Extract shared special-application fields into a spread array](https://github.com/GCTC-NTGC/gc-digital-talent/commit/7167a118efc779698634da24c28130bd7ea1be0e), brought to you by **Deebo Samuel** — the do-it-all weapon you *spread* all over the formation instead of fielding the same duplicate values in every array.

<sub>🤖 Claude Opus 4.8 (1M context)</sub>

api/lang/fr/activity_event.php
vd1992: https://github.com/GCTC-NTGC/gc-digital-talent/pull/17545/changes/f2f4f04dfef869595848b993e3d810cabd0d023c

esizer (approved):

PR #17550, opened 2026-07-24
[Fix] Validate response to pool question

### LINKED ISSUE #17405: 🐛 Question responses can connect to questions from a different pool
### Bug

When calling `updateApplication`, `GeneralQuestionResponse` and `ScreeningQuestionResponse` records can be created via `generalQuestionResponses.create` / `screeningQuestionResponses.create` with a question ID from a *different* pool. No validation exists to prevent this.

Related to the class of bug fixed in #17223.

### Steps to Reproduce

1. Authenticate as any applicant.
2. Call `updateApplication` with `generalQuestionResponses.create` containing a `generalQuestion.connect` ID from a different pool's question.
3. The response is created and linked to the wrong pool's question — no error.

### Proposed Solution

Add validation rules to `UpdateApplicationValidator` checking that question IDs belong to the application's pool:

```php
$poolId = PoolCandidate::find($this->arg('id'))->pool_id;
$generalQuestionIds = GeneralQuestion::where('pool_id', $poolId)->pluck('id')->toArray();

'application.generalQuestionResponses.create.*.generalQuestion.connect' => [
    Rule::in($generalQuestionIds),
],
```

Same pattern for screening questions.

### Acceptance Criteria

- [ ] `updateApplication` rejects `generalQuestionResponses.create` when the connected question does not belong to the application's pool
- [ ] Same for `screeningQuestionResponses.create`

esizer:
🤖 Resolves #17405

## 👋 Introduction

Adds validation to ensure users are responding to questions on the exact pool they are replying to.

## 🕵️ Details

Previously, a user could create a response to a question from a different pool. This adds the validaiton to prevent that by checking that the ID with the mathcing pool ID of the applicaiton exists.

## 🧪 Testing

1. Login as any user
2. Start an application
3. Manually submit an `updateApplication` mutation to that application attempting to respond to a question on a pool different than the one you were applying to
4. Confirm that you get a validation error

### DIFF
--- a/api/app/GraphQL/Validators/Mutation/UpdateApplicationValidator.php
+++ b/api/app/GraphQL/Validators/Mutation/UpdateApplicationValidator.php
@@ -18,10 +18,12 @@ final class UpdateApplicationValidator extends Validator
     {
         $applicationId = $this->arg('id');
 
-        $userId = PoolCandidate::query()
-            ->select('user_id')
-            ->findOrFail($applicationId)
-            ->user_id;
+        $application = PoolCandidate::query()
+            ->select(['user_id', 'pool_id'])
+            ->findOrFail($applicationId);
+
+        $userId = $application->user_id;
+        $poolId = $application->pool_id;
 
         return [
             'application.educationRequirementAwardExperiences.sync.*' => [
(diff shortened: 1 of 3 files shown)


petertgiles (approved):
Nice, clean fix.

Issue #17227, opened 2026-06-26
♻️ Remove instances of full object types


esizer:
### ♻️ Refactor

Codegen is [removing the generation of full object types in the next version](https://the-guild.dev/graphql/codegen/docs/migration/operations-and-client-preset-from-5-0#:~:text=1.-,Object%20types%20are%20no%20longer%20generated,-Previously%2C%20Object%20types), so we need to get rid of them too.

### 🕵️ Details

Codegen currently generates full object types for the graphql schema types and we use them quite extensively. Those would be the types like `User`, `Pool`, `PoolCandidate`.

These can cause problems since they contain the full object and non-nullable fields become required (I'm sure everyone has seen the error `X is required in type Y` when using them before which forces you to query it even though you dont need to use it.).

Since they are removing it, we will need to replace the instaces with more specific types and in most cases, likely reach for a fragment.

I expect this work to be quite intensive since we do use those types frequently. This will require many different small decisions on the approach we take for each one.

### 🙋‍♀️ Proposed Solution

Find all full object types (`Pool`, `CommunityInterest`, `DevelopmentProgram`, etc) and replace it with either:

- A specific interface defining only the fields needed rather than using `Pick`, `Extract`, `Omit` helpers to pull fields out
- A graphql fragment with a mask

The approach will depend on the usage.

### When to use a fragment

Use a GraphQL fragment when:

- The data comes directly from a query or mutation response
- The component or function receives the data as a prop from a parent that owns the query
- You want the type to stay in sync with the schema automatically (schema changes update the type without touching TS)
- The same fields are needed by multiple sibling components — share the fragment instead of duplicating interfaces
- The component is "slottable" — it will be spread into different parent queries across the app

**The test**: if you can write `...MyComponentFragment` in a GraphQL operation to satisfy it, use a fragment.

### When to use an interface

Use a TypeScript interface when:

- The function is a utility or formatter with no direct connection to a query (e.g. `formatUserName(firstName: string; lastName: string)`
- You are writing tests or factories and need to construct mock objects without a real query
- The type crosses a boundary where GraphQL has no reach (localStorage, URL params)
- The fields needed are so minimal and universal that attaching a fragment would force every consumer to spread it in their query even when the fields are already present

**The test**: if writing a fragment would require callers to add a spread they'd never otherwise need, use an interface.

### Other

[...]

PR #17552, opened 2026-07-24
[Refactor] Remove full graphql object types #1

esizer:
## 👋 Introduction

Partial resolution for #17227.

This contains setup work and a few small example changes for removing full object types.

## 🕵️ Details

This does a few things I feel is important to explain:

 - Adds the typescript plugin to generate full object types
 - Restricts full object import types to `fake-data`
 - Adds documentation on how to use the graphql package
 - Updates a few packages to remove full object types

### Why are we still generating full object types?

It might seem counterintuitive to add the full objecvt type generation in a PR that is intended to remove them. However, fully removing them is exceedingly tricky for a few reasons:

- Fake data relies on them heavily and it is not so straight forward to remove them
- We have some small utility types that are genuinly useful to use and it would seem silly to hand write them

The fake data package is a legitimate use case for these full object types and I felt like converting them to either fragments or interfaces is just a lot of work and a little odd. So, we still generate all of them but, we only allow fake data to import them.

In order for any other package to use them, you must add them as a named export in `grpahql/index.ts` so that we don't unintentionally import somerthing like `Pool` or `User` in `apps/web`.

I added some documentaion in `packages/graphql` to hopefully make this clear.

## 🧪 Testing

1. Install and build `pnpm i; pnpm dev:fresh`
2. Confirm no lint errors
3. Confirm `fake-data` package functions (tests, storybook)
4. Cofirm main app still builds and functions as expected


yonikid15 (approved):
Tests all pass ✔️ 
Storybook working as expected ✔️ 
Error message working as expected ✔️

Image

Issue #5645, opened 2023-02-07
New Date Picker


petertgiles:
## ✨ Feature

We've had various issues with the existing strategy of using `input type=date` for date inputs.  This is a placeholder issue to update this with a new date picker.

## 🕵️ Details

Add any additional details that could assist with the development of the new feature.

## 🎨 Design File

Include a link to the design file (if it exists).

## 📸 Screenshot

Add a screenshot of the new feature (if possible).

## 🙋‍♀️ Proposed Implementation

(optional) Do you have a proposed implementation?

## 🌎 Localization

(optional) Provide any new copy along with translations available.

## ✅ Acceptance Criteria

A set of assumptions which, when tested, verify that the feature was properly implemented.

- [ ] Criteria 1
- [ ] Criteria 2

substrae:
Added some notes above. @RM-1978 @esizer can y'all have a think about this approach from an accessibility perspective? I'm convinced that rather than messing around with libraries, having three standard, accessible inputs is going to provide the best experience across browsers, ATs, and devices.

esizer:
> Added some notes above. @RM-1978 @esizer can y'all have a think about this approach from an accessibility perspective? I'm convinced that rather than messing around with libraries, having three standard, accessible inputs is going to provide the best experience across browsers, ATs, and devices.

Makes sense to me! We will need to be careful on validation here since, while unlikely, it could allow users to enter an invalid date. Should we consider a min date and max date for validation as well? I'm thinking of things like:

 - Birthday should have max date of today
 - End date should have min date of start date
 - Start date should have max date of end date
 - etc.

RM-1978:
> Thanks @substrae I agree with @esizer 100%.  I also think that adding min and max dates would provide good safeguards  against consequences for the user resulting from potential mistakes (e.g. unintended keystrokes).

mnigh:
This would solve https://github.com/GCTC-NTGC/gc-digital-talent/issues/4314.

brindasasi:
can we have this fable tested from storybook? @RM-1978

RM-1978:
We have had this Fable Tested this component. See results in [Fable Date Picker Test results and requirements #6652](https://github.com/GCTC-NTGC/gc-digital-talent/discussions/6652) for a discussion of each of the user requirements coming out of this work with the Lab.

PR #17555, opened 2026-07-24
[Fix] Date input reset

### LINKED ISSUE #17459: 🐛 DateInput component doesn't properly reset

Our DateInput component doesn't get reset properly by React Hook Form. After calling resetField on a DateInput the field can visually appear set but throw errors when submitting the form.

### 🕵️ Details

This doesn't appear to currently be a problem anywhere in the site because it gets masked by hiding the field causing it to re-render with the new value. However I encountered this when working on the education experience form rework which merely replaces one DateInput with a different one and isn't completely re-rendered.

### 📋 Steps to Reproduce

To replicate this on main we have to edit a form to reset a DateInput without hiding it. I went with the pause referrals form since it's small.

1. Comment out the logic for hiding the date field in PauseReferralsFormFields.tsx
2. Navigate to view a candidate currently being referred
3. Select to open the pause referrals form
4. Fill out the fields selecting "Other" for pause length
5. Change the pause length to something else and back to "Other" to reset the date field
6. Confirm the date field still appears set
7. Select to submit the form
8. Confirm the form throws an error that the date field is required

### 📸 Screenshot

Image

https://github.com/user-attachments/assets/bb93d53d-f4b3-4452-a22c-8e8c6fc599e3

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] DateInput fields can be properly reset without being hidden.

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #17459 

## 👋 Introduction

Fixes an issue where the date input was not being reset.

## 🕵️ Details

This was because there was a default value set on the input. I beleive when we reset it, the actual value was removed but it would then fallback to the default value. Removing default value allowed the actual value to be set/reset.

In order to confirm that default value still worked, a test was added for that as well.

## 🧪 Testing

1. Comment out the logic for hiding the date field in PauseReferralsFormFields.tsx
2. Navigate to view a candidate currently being referred
3. Select to open the pause referrals form
4. Fill out the fields selecting "Other" for pause length
5. Change the pause length to something else and back to "Other" to reset the date field
6. Confirm the date field still appears to be emtpty
7. Select to submit the form
8. Confirm the form submits without error


packages/forms/src/components/DateInput/ControlledInput.tsx
```diff
@@ -145,7 +134,7 @@ const ControlledInput = ({
             name={ID.DAY}
             type="number"
             onChange={handleDayChange}
-            defaultValue={day}
+            value={day ? Number(day) : ""}
```
petertgiles: It looks like we have an unexpected behaviour change with this bug fix.

Image

petertgiles (changes requested):
Seems to work well but I think there's an unexpected behaviour change visible in Chromatic.

packages/forms/src/components/DateInput/ControlledInput.tsx
esizer: This is part of the fix. It was having issues due to a string being passed in. I added a `Number(day)` to properly set the value but, it also stripped the leading 0.

I guess we need to determine if we need the leading zero or not :thinking:

petertgiles: I went back and dug up in the original issue.
https://github.com/GCTC-NTGC/gc-digital-talent/issues/5645
Yeah, it looks like it should be two digit formatted.

packages/forms/src/components/DateInput/ControlledInput.tsx
```diff
@@ -145,7 +137,7 @@ const ControlledInput = ({
             name={ID.DAY}
             type="number"
             onChange={handleDayChange}
-            defaultValue={day}
+            value={day ?? ""}
```
vd1992, line 140: The day segment can't be cleared with Backspace after this change — it sticks on `00`. Editing a pre-filled date like `2023-01-15`, the day field goes `15 → 01 → 00 → 00 → 00…`, forever. `Delete` and `Home`+`Delete` don't help either; only `Ctrl+A`+Delete gets out.

Year is fine, and that's the clue — year passes a number, day passes a string. React's controlled update for `type="number"` uses a loose `!=`, so `"1" != 1` is false and year keeps the user's text, but `"1" != "01"` is true and React writes the padded value back on every keystroke. Number inputs don't support the selection API, so React can't restore the caret — it lands at the end, Backspace deletes the trailing character, and the handler re-pads `"0"` straight back to `"00"`.

Worth noting `min={1}` makes the stuck `00` a range underflow, so native validation blocks the submit before RHF ever runs — our own error message never renders and Submit just appears dead.

```suggestion
            value={day ? Number(day) : ""}
```

Makes day match year — Backspace goes `15 → 1 → ""`. I also checked forward typing, single digits, the spinner arrows, and `reset()` to both empty and non-empty values; all still behave, so `handleDayChange` can stay exactly as you wrote it.

<sub>🤖 Claude Opus 5 (1M context). Reproduced in Chromium against this branch's built component with real react-hook-form; the same harness on `main` clears fine, so it's new here.</sub>

vd1992 (dismissed):
Requesting changes on one item — the rest of this is good.

Moving to controlled inputs is the right fix for #17459, and I confirmed it does more than the description claims: `reset()`, `resetField(..., { defaultValue: undefined })` and `setValue()` all now update the visible segments, where previously none of them did.

The reason I am requesting changes rather than just leaving a note is that **this also introduces a regression, and one with some real significance.** The day segment can no longer be cleared with Backspace — it gets stuck on `00`, and since `min={1}` makes that a native range underflow, the browser blocks submission before react-hook-form runs, so Submit appears to do nothing and no error message is shown. The same key sequence on `main` clears fine, so this is new behaviour, not something pre-existing.

Nothing invalid can actually be saved, so it is not a data-integrity problem — but it is a dead-end in a shared primitive with roughly 24 day-visible usages, including the prefilled dialogs where people edit an existing date (expiry date, change date, claim verification, job placement). That feels like more than we should ship.

The fix is one line and I have left it as a suggestion inline, along with the mechanism and what I verified against it.

<sub>🤖 Review by Claude Opus 5 (1M context), run by @vd1992.</sub>

packages/forms/src/components/DateInput/ControlledInput.tsx
esizer: I already explained that this was already in and working but was told we do not want that.

Now I'm unsure what to do. That strips the leading zero that is as told was a requirement.

petertgiles: It seems good to me.  On the left is main and on the right is this branch.  They behave identically as far as I can tell.

https://github.com/user-attachments/assets/7b7cf69f-68e5-4e9c-a787-64a2e3fb2868

packages/forms/src/components/DateInput/ControlledInput.tsx
esizer: okay, updated to left pad [left pad default date value](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17555/commits/b5e80a011140ea0df824333d77168aa3912d0739)

Looks as though you confirmed this change did not include the regression claude thought they found?

petertgiles: Diff is gone!

petertgiles: Diff is gone!

petertgiles (approved):
Works well for me.  👏   Reset works on the pause referrals form.

packages/forms/src/components/DateInput/ControlledInput.tsx
vd1992: @esizer — my suggestion was wrong and I'm withdrawing it, with apologies. `value={day ? Number(day) : ""}` is precisely what you had in `5d37e7a1bf`, and precisely what was rejected against the two-digit requirement in #5645. I reviewed the head commit without reading this PR's own review history — so the answer was sitting in the thread I posted into, and I still sent you back to work you'd already been told to undo. That was avoidable and it's entirely on me; you were right on both counts, and it's my fault you were left unsure what to do. The leading-zero constraint should have been the starting point of my review instead of something I discovered afterwards.

@petertgiles — the behaviour does depend on how you enter the field, which is probably why they looked identical. **Tabbing** into the day input selects its contents, so the first Backspace wipes it and both branches clear. The difference only shows when nothing is selected:

1. **Click** into the day field of a pre-filled date (or press End) — don't Tab in
2. Press Backspace twice

`main` gives `15 → 1 → ""`. This branch gives `15 → 01 → 00`, and further Backspaces stay on `00` — the padded value is written back on each keystroke, so it re-pads `0` to `00` faster than you can delete it. `Ctrl+A` then Delete still clears it. I measured this in Chromium against builds of both branches; it reproduces in 2 of 7 entry flows, all of them the caret-at-end ones.

That's narrower than my original comment, which stated it flatly with no precondition at all. I overstated it, and I should have pinned down the entry-flow dependency before raising it as a blocker.

**On the actual dilemma of leading-zero vs. clearable:** you can have both by padding on blur instead of on every keystroke — render the user's raw text while the field is focused, fall back to the padded form value when it loses focus. I tested this: Backspace clears normally, typing `7` still settles to `07`, and `reset()` to `1999-12-05` still shows `05`. One wart worth knowing — if the form is reset while the day field is still focused, the field shows the stale typed text until blur.

I'm dismissing my blocking review; this shouldn't hold up a fix that works. Whether the blur change belongs here or in a follow-up ticket is your call, not mine.

<sub>🤖 Claude Opus 5 (1M context). Measured in Chromium against built copies of `main` and `b5e80a0`; I could not view the video attachment, so the Tab-vs-click explanation is my best inference of the difference rather than a reading of your test.</sub>

vd1992:
The bot is super sorry Eric and it is committing what you said to memory 😄 

image

PR #17557, opened 2026-07-25
[Fix] Ul/Ol add bottom margin to the last item in a list

### LINKED ISSUE #17232: 🐛 Ul component adds bottom margin to last item in list

The Ul component controls the bottom margin of its list items using the space prop.  It also adds it to the last item in the list which results in excessive margin below the list.

[Ref: packages/ui/src/components/List/styles.ts](https://github.com/GCTC-NTGC/gc-digital-talent/blob/1f4f8a504ed893da61a06555fd56d32a6a80d312/packages/ui/src/components/List/styles.ts#L11)

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

_No response_

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

Something like `[&_li:not(:last-child)]` ?

### ✅ Acceptance Criteria

- [ ] Last item in the list does not get bottom margin
- [ ] Look for places where the parent component may be relying on this extra space and fix them
- [ ] Review Chromatic for layout issues this may introduce

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

vd1992:
🤖 Resolves #17232

## 👋 Introduction

The `Ul` and `Ol` components applied a bottom margin to every list item, including the last one, so every list using the `space` prop rendered with surplus whitespace underneath it. This scopes that margin to all but the last item. I checked all 108 call sites that pass `space` for anything leaning on the extra space, and only one was — the comments list on the Terms and Conditions page, which now carries its own bottom margin. I also added a Storybook story, since the List components had none. A final commit fixes three unrelated list defects I ran into while auditing the call sites; it's self-contained and can be dropped if you'd rather see it separately.

## 🧪 Testing

Nearly every diff should look the same in kind: slightly less whitespace *below* a list — 3px at `space="sm"`, 6px at `md`, 12px at `lg`, 24px at `xl`. Anything that isn't that is worth a second look.

- **The failure mode to hunt for is over-tightening** — list text now flush against a container's bottom edge or the paragraph below it.
- **Lists followed by a heading should be unchanged.** The stray margin was already collapsing into the heading's larger `mt-9`, so it was never visible there. A diff on one of those means something is off.
- Densest list usage, worth a direct look: Accessibility Statement, Terms and Conditions, Roles and Permissions, IT Training Fund, Application Review, and the Applicant / Community / Admin dashboards.

For the `adjacent defects` commit:

1. **Certification Exam Vouchers** and **Instructor-Led Training** should render *identically* — only the DOM changed. The check is the console: React no longer warns `<ul> cannot appear as a descendant of <p>`.
2. **DND Digital Careers** skills grid should be unchanged; row spacing now comes from `gap-y-1.5` instead of item margins.
3. **Table of contents sidebar** should show no diff — that margin was already collapsing away. Fixed as the same latent bug, not because anything looked wrong.

## 📸 Screenshot

The new List story at the same viewport. The "before" build reverts only the `styles.ts` variant, so the two are directly comparable.

### Before

Gap below the last item, growing with the `space` value — clearest at `space="xl"`.

### After

Flush at every `space` value, with the spacing *between* items unchanged.

### Measured

A trailing `li` margin collapses out of the `ul`'s box, so the gap between the list and the block below it is exactly the stray margin:

| `space` | Before | After |
| --- | --- | --- |
| `sm` | 3px | **0px** |
| `md` | 6px | **0px** |
| `lg` | 12px | **0px** |
| `xl` | 24px | **0px** |
| nested | 6px | **0px** |

Same for `Ul`, `Ol` and `unStyled` — 15 cases, all 0px after.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Claude Opus 5 (1M context)

### DIFF
--- a/apps/web/src/pages/DNDDigitalCareersPage/DNDDigitalCareersPage.tsx
+++ b/apps/web/src/pages/DNDDigitalCareersPage/DNDDigitalCareersPage.tsx
@@ -217,7 +217,7 @@ export const Component = () => {
             description: "Lead-in text for specific skills DSG is looking for",
           }) + intl.formatMessage(commonMessages.dividingColon)}
         </p>
-        <Ul space="md" className="grid max-w-md sm:grid-cols-2 sm:gap-x-6">
+        <Ul className="grid max-w-md gap-y-1.5 sm:grid-cols-2 sm:gap-x-6">
           <li>
             {intl.formatMessage({
               defaultMessage: "cyber security",
--- a/packages/ui/src/components/List/List.stories.tsx
+++ b/packages/ui/src/components/List/List.stories.tsx
@@ -1,0 +1,104 @@
+import type { Meta, StoryFn } from "@storybook/react-vite";
+
+import { allModes } from "@gc-digital-talent/storybook-helpers";
+
+import type { ListVariants } from "./styles";
+import Ol from "./Ol";
+import Ul from "./Ul";
+
+const spaces: ListVariants["space"][] = ["sm", "md", "lg", "xl"];
+
+// A visible sibling so the gap below the last item is legible in a snapshot.
+const Boundary = () => (
+  <div className="border-t border-error bg-error/10 p-1.5 text-sm">
+    Following content
+  </div>
+);
+
+export default {
+  component: Ul,
+  parameters: {
+    chromatic: {
+      modes: {
+        light: allModes.light,
+        dark: allModes.dark,
+      },
+    },
+  },
+} as Meta<typeof Ul>;
+
+const Template: StoryFn<typeof Ul> = () => (
+  <div className="flex flex-col gap-12">
+    {spaces.map((space) => (
+      <div key={space}>
+        <p className="mb-3 font-bold">{`space="${space}"`}</p>
+        <div className="grid gap-6 sm:grid-cols-3">
+          <div>
+            <Ul space={space}>
+              <li>Unordered one</li>
+              <li>Unordered two</li>
+              <li>Unordered three</li>
+            </Ul>
+            <Boundary />
+          </div>
+          <div>
+            <Ol space={space}>
+              <li>Ordered one</li>
+              <li>Ordered two</li>
+              <li>Ordered three</li>
+            </Ol>
+            <Boundary />
+          </div>
+          <div>
+            <Ul space={space} unStyled>
+              <li>Unstyled one</li>
+              <li>Unstyled two</li>
+              <li>Unstyled three</li>
+            </Ul>
+            <Boundary />
+          </div>
+        </div>
+      </div>
+    ))}
+    <div>
+      <p className="mb-3 font-bold">Nested</p>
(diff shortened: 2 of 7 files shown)


vd1992:
D'aww, Claude needs a hand posting images 🥲 

#### BEFORE

01-before

#### After

02-after

apps/web/src/pages/DNDDigitalCareersPage/DNDDigitalCareersPage.tsx
```diff
@@ -217,7 +217,7 @@ export const Component = () => {
             description: "Lead-in text for specific skills DSG is looking for",
           }) + intl.formatMessage(commonMessages.dividingColon)}
         </p>
-        <Ul space="md" className="grid max-w-md sm:grid-cols-2 sm:gap-x-6">
+        <Ul className="grid max-w-md gap-y-1.5 sm:grid-cols-2 sm:gap-x-6">
```
petertgiles, line 220: This change is a little funny.  Why did you stop using the space prop?

packages/ui/src/components/List/List.stories.tsx
petertgiles, line 1: Cool story!  I like the use of colour to show borders.

petertgiles:
The AI review found an unrelated bug: https://github.com/GCTC-NTGC/gc-digital-talent/issues/17571

packages/ui/src/components/List/styles.ts
petertgiles, line 1: The AI review flagged that the SalaryRange dialog changed as well but it looks fine to me.

Image

petertgiles (approved):
Nice one!

apps/web/src/pages/DNDDigitalCareersPage/DNDDigitalCareersPage.tsx
vd1992: Not sure, let us get an answer from the source

vd1992: Because `space` is a per-item `margin-bottom`, and grid doesn't collapse margins — so on this list the fix wouldn't actually remove the surplus space, it would just relocate it into the last row. I measured this list both ways at `sm` and up (6 items, 2 columns): with `space="md"` the `<ul>` stays 90px tall and the final item stretches to 30px against its siblings' 24px, whereas `gap-y-1.5` gives 84px with every item at 24px. The row gap is an identical 6px either way, so nothing moves visually today — but the uneven last row would show the moment those items get a background or a border. `gap-y-1.5` is just that same 6px expressed the way the grid already expresses its column spacing (`sm:gap-x-6`). Happy to put the prop back if you'd rather keep it for consistency; it just wouldn't be doing anything useful on a grid.

<sub>🤖 Claude Opus 5 (1M context)</sub>

PR #17563, opened 2026-07-27
[Bug] Fix complete nomination history for nominee

### LINKED ISSUE #17542: 🐛 Nomination History only displays one talent event instead of complete history

When viewing the Nomination History tab for a nominee only a single talent event is being displayed. The page should display the nominee's complete nomination history across all talent management events they have been nominated in.

https://github.com/GCTC-NTGC/gc-digital-talent/pull/17322

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Login as `talent-coordinator@test.com`
2. Navigate to admin/talent-events
3. Create nominations for the same user in multiple talent management events
5. Open that user's nominee's profile
6. Select the Nomination History tab
7. Observe that only one talent event is displayed even when the nominee has nominations from multiple events

### 📸 Screenshot

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ]  The Nomination History tab displays **all** talent events associated with the nominee

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

bella77-69:
🤖 Resolves  #17542

## 👋 Introduction

This PR fixes the Nomination History tab to display all talent events associated with the nominee.

## 🧪 Testing

1. Build `pnpm dev:fresh`
3. Navigate to admin/talent-events and select an event with nominations & a nominee with multiple talent events
4. Verify:
    - Multiple talent events are displayed in the history tab for the nominee

## 📸 Screenshot
image

image


petertgiles:
Lol, Copilot just won't leave you alone.  :laughing: Do you have anything in this menu you can turn off?
image

bella77-69:
> Lol, Copilot just won't leave you alone. 😆 Do you have anything in this menu you can turn off?>

Yep I did that. Its so weird because it never auto did a review for me until we switched to Claude.  I will have to do some more digging 🤔

api/graphql/schema.graphql
```diff
@@ -1964,6 +1963,14 @@ type Query {
     @guard
     @canQuery(ability: "view", scopes: ["withPolicyEagerLoads"])
 
+  talentNominationGroupsByNominee(
+    nomineeUserId: UUID! @eq(key: "nominee_id")
+    communityId: UUID @scope(name: "whereCommunity")
+  ): [TalentNominationGroup!]!
+    @all(scopes: ["withPolicyEagerLoads", "authorizedToView"])
+    @guard
+    @canResolved(ability: "view", scopes: ["withPolicyEagerLoads"])
```
esizer: We should try and avoid `@canResolved` it has given us performance and memory issues in the past.

The `whereAuthorizedToView` scopes should cover any auth that we would would otherwise do in the `view` method on the policy.

api/graphql/schema.graphql
```diff
@@ -1964,6 +1963,14 @@ type Query {
     @guard
     @canQuery(ability: "view", scopes: ["withPolicyEagerLoads"])
 
+  talentNominationGroupsByNominee(
+    nomineeUserId: UUID! @eq(key: "nominee_id")
+    communityId: UUID @scope(name: "whereCommunity")
```
esizer: I think we can remove the community arg. I dont see it being used anywhere and there is not accompanying method on the model.

apps/web/src/pages/TalentNominations/NominationGroup/History.tsx
```diff
@@ -44,30 +50,56 @@ const TalentNominationGroupHistoryOptions_Fragment = graphql(/* GraphQL */ `
...
-  const talentNominationGroup = getFragment(
-    TalentNominationGroupHistoryNominationGroup_Fragment,
-    nominationGroupQuery,
-  );
+
+  const nominationGroups = nominationGroupsQuery
+    .map((group) =>
+      getFragment(TalentNominationGroupHistoryNominationGroup_Fragment, group),
+    )
+    .filter(Boolean);
```
esizer: I might be wrong but I dont think the `Boolean` filter or the map are doing anything here and it can be written like the following:

```tsx
const nominationGroups = getFragment(..., nominationGroupsQuery);
```

esizer (changes requested):
Cool! I noticed a few minor things, hopefulkly it all makes sense 😅

api/graphql/schema.graphql
petertgiles: I suggested it because most users can only see talent nomination groups in their own community.
image

I suppose that it would probably take some extra work to make it do anything though, since community ID isn't a field on TalentNominationGroup.

If the authorized to view scope achieves that same thing, I'm fine to leave it out though.

esizer: ahh okay, thanks for that explination! This does sound like authorization, unless we do intended to allow users to filter the list by communities they are a part of.

esizer (approved):
LGTM! 👍

PR #17572, opened 2026-07-28
[Fix] `ScrollToLink` to scroll nearest scrollable parent

### LINKED ISSUE #17246: 🐛 UAT | Clicking validation error link scrolls down the background page instead of dialog

- Background page scrolls down when user clicks on field validation error when trying to select any actions under Candidate tracking table

- **Expected Outcome:** When user clicks on the field validation error, the dialog scroll should down

### 🕵️ Details

Env: UAT
Version: 2.75

### 📋 Steps to Reproduce

1. Access GC digital portal as Community admin role
2. Navigate to the Talent requests 
3. Open the talent request which is already submitted and having matching candidates
4. Make sure any one or two candidates are present in the candidate tracking table
5. select at least 2 candidates from candidate tracking table
6. Click on Actions -> Change status to not selected
7. Without selecting any reason, click on Save changes
8. Verify an error occurred
9. Click on that field validation error multiple times
10. Verify that the background or outer scrolls down to the page

### 📸 Screenshot

https://github.com/user-attachments/assets/05e3ef13-43e4-4f19-ad3f-0a27170ebdfa

### 🙋‍♀️ Proposed Solution

Eric:
> This is becau we use scrollTo and not scrollIntoView
scrollTo does the page and not nearest scrolable elemtn it hink :thinking_face:

### ✅ Acceptance Criteria

- [ ] The inner scroll or dialog scrolls down when user clicks on the field validation error

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #17246 

## 👋 Introduction

Updates the `ScrollToLink` component to prefer scrolling the nearest scrollable element rather than always the window.

## 🕵️ Details

This replaces `scrollTo` with `[scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView)` which scrolls the nearest scrollable ancestor.

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Login as `admin@test.com`
3. Navigate to view a candidates application
4. Open the skill assessment dialog
5. Attempt to submit the form with an error
6. Activate the link in the error summary
7. Confirm the dialog gets scrolled and not the page behind it
8. Navigate to any page with a table of contents
9. Confirm no regression in the scrolling on the table of contents links

### DIFF
--- a/packages/ui/src/components/Link/ScrollToLink.tsx
+++ b/packages/ui/src/components/Link/ScrollToLink.tsx
@@ -20,25 +20,18 @@ const scrollToSection = (
   offsetEl?: HTMLElement | null,
 ) => {
   if (section) {
-    if (offsetEl) {
-      setTimeout(() => {
-        window.scrollTo({
-          behavior: "smooth",
-          top:
-            section.getBoundingClientRect().top -
-            document.body.getBoundingClientRect().top -
-            // NOTE: 16 adds a bit of a gap so text doesn't touch element
-            (offsetEl.getBoundingClientRect().height + 16),
-        });
-      }, 10);
-    } else {
-      setTimeout(() => {
-        section.scrollIntoView({
-          behavior: "smooth",
-          block: "start",
-        });
-      }, 10);
-    }
+    setTimeout(() => {
+      if (offsetEl) {
+        // NOTE: 16 adds a bit of a gap so text doesn't touch element
+        const offset = offsetEl.getBoundingClientRect().height + 16;
+        section.style.scrollMarginTop = `${offset}px`;
+      }
+
+      section.scrollIntoView({
+        behavior: "smooth",
+        block: "start",
+      });
+    }, 10);
   }
 };
 


vd1992 (approved):
- Tested manually
- Tested Claude's suggestions
- Background scroll also gone

Will paste this line from Claude due to the "Worth recording" line
>Worth recording why the old path failed, since it's slightly worse than "scrolled the wrong element" — the offset math wasn't idempotent for a target inside a fixed overlay. `section.getBoundingClientRect().top` is stable there while `document.body.getBoundingClientRect().top` moves, so each activation adds the current scroll position back on top of the intended destination. In a reproduction of the overlay CSS the page jumped to 1009 from a standing start, and to 1509 from a starting scroll of 500; successive clicks therefore compound — 0 → 1009 → 2018 — and the target never came on screen. That's the "click it multiple times" symptom in #17246.


Problem was resolved, looking good. Great!

PR #17577, opened 2026-07-28
[Fix] Assessment step mutation validation

### LINKED ISSUE #17408: 🐛 Assessment step pool skill sync does not validate pool ownership
### Bug

`createAssessmentStep` and `updateAssessmentStep` sync pool skills via Lighthouse's `@create`/`@update` directives without validating that the provided pool skill IDs belong to the assessment step's pool. An admin with access to Pool A could attach pool skills from Pool B to an assessment step in Pool A.

The sibling mutation `createOrUpdateScreeningQuestionAssessmentStep` already has the correct check (`CreateOrUpdateScreeningQuestionAssessmentStep.php:33-40`) — it validates `$skill->pool_id !== $pool->id` before syncing. This check was missed on the other two mutations.

Related to the class of bug fixed in #17223. Split from #17261.

### Steps to Reproduce

1. Authenticate as a pool admin with access to Pool A (in DRAFT status).
2. Obtain a pool skill UUID belonging to Pool B.
3. Call `createAssessmentStep` with `poolId` for Pool A and `poolSkills: { sync: ["<pool-B-skill-id>"] }`.
4. The assessment step is created with a cross-pool pool skill attached — no error.

### Proposed Solution

Replace the Lighthouse `@create`/`@update` directives with a custom mutation resolver (or add a `@validator`) that checks each pool skill ID belongs to the step's pool before syncing — matching the pattern in `CreateOrUpdateScreeningQuestionAssessmentStep`:

```php
foreach ($poolSkillIds as $skillID) {
    $skill = PoolSkill::find($skillID);
    if ($skill === null || $skill->pool_id !== $pool->id) {
        throw new Exception(ErrorCode::POOL_SKILL_NOT_VALID->name);
    }
}
```

### Acceptance Criteria

- [ ] `createAssessmentStep` rejects pool skill IDs that do not belong to the step's pool
- [ ] `updateAssessmentStep` rejects pool skill IDs that do not belong to the step's pool
- [ ] Existing assessment step functionality is unaffected

esizer:
🤖 Resolves #17408 

## 👋 Introduction

Updates the create and update mutations for assessment steps to use validation ensuring skills are part of the pool.

## 🧪 Testing

1. Attempt to run mutation with skills that are not part of the pool
2. Confirm validation error is returned
3. Confirm test coverage exists


api/tests/Feature/AssessmentStepTest.php
vd1992, line 1: Only thing missing is the positive cases tests mentioned in the A/C. 
Good to add I think, help ensure it is blocked when it should be and allowed when it should be

vd1992 (changes requested):
Nothing significant, just a bit of test coverage to match the issue's A/C

api/tests/Feature/AssessmentStepTest.php
esizer: okay, just to confirm. Do tyou consider this existing test enough to cover that? That was why I didn't explictly add a positive test since it successfully creates an assessment step with the pool skill. If not, I can try and add a more explictit positive case.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/e8bc10ffa3bc42a5d48070d08c99ef0256e2509d/api/tests/Feature/AssessmentStepTest.php#L209-L275

vd1992: Hm, yes that addresses `createAssessmentStep`
But I don't see update. I think that should be created, and probably placed next to that one

esizer: oh yeah! That makes sense. Added in [add positive update case](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17577/commits/24de9d0b243c643f2cddb8d1d6ce9da26b9c1032)

vd1992: Thanks

vd1992 (approved):
Good stuff!

PR #17579, opened 2026-07-28
[Feature] Improve bulk actions on inbox component

### LINKED ISSUE #17331: ✨ Improve bulk actions on Inbox component
### Details

After some research on how some popular email handle bulk action it is clear there is some best practices around hiding showing which actions can be performed based on whether the user has selected an item or not.

### Design file & designer
Figma file by @Jerryescandon 

When **NONE** are selected (will affect the entire results, except filtered ones):

- Show 'More actions' menu
  -  Mark all as not selected
  -  Download all as spreadsheet (.xslx)

When **ONE OR MORE** are selected (will affect ONLY the selected items):

- Show 'Mark as referred' button
- Show 'Mark as not referred' button
- Show 'Mark as not selected' button
- Show 'Download actions' menu
   - Download profiles
   - Download profiles without contact information
   - Download as spreadsheet

ADDITIONAL CLEANUP 🧹
- update language from "Change status to" to "Mark as"
- remove icons from all drop-down menus
- replace icon for "referred" with `bookmark-square-mini` on the status

OUT OF SCOPE

Remove 'Change status to selected' from the bulk actions.
This specific status should be set more intentionally.
For now, lets leave it only inside the dialogue when selecting a candidate.

### 📸 Screenshots

When no item has been selected:

Image

Image

When one or more items have been selected

Image

Image

Image

### 🌎 Localization

_No response_

### Proposed Implementation

_No response_

### ✅ Acceptance Criteria

- [ ] When **NONE** items are selected then we only show 'More actions' menu:
- [ ] 'More actions' menu affect all pages but exclude filtered items
- [ ] Menu includes all following options:
  - [ ] Mark all as not selected
  - [ ] Download all as spreadsheet (.xslx)

- [ ] When **ONE OR MORE** are selected show different actions
- [ ] Show 'Mark as referred' button
- [ ] Show 'Mark as not referred' button
- [ ] Show 'Mark as not selected' button
- [ ] Show 'Download actions' menu which includes the following:
  - [ ] Download profiles
  - [ ] Download profiles without contact information
  - [ ] Download as spreadsheet

ADDITIONAL CLEANUP 🧹
- [ ] update language from "Change status to" to "Mark as"
- [ ] remove icons from drop-down menus
- [ ] replace icon for "referred" with `bookmark-square-mini` on the status

yonikid15:
🤖 Resolves #17331 

## 👋 Introduction

- Updates the actions bar logic depending on how many items are selected:
  - If none are selected: Show the Actions button icon (ElipsesIcon)
  - If one ore more are selected: 
    - Show an icon button for each status (Referred, Not referred, Not selected, Selected)
    - Followed by the download button with dropdown caret utility icon
- Makes changes to the copy from "Change status" to "Mark as", or "Mark all as"
- Replaces referred icon with BookmarkSquareMiniIcon
- Adds outline icons for the actions 
- Removes marking selected items as "Selected" in the actions bar. It says it's out of scope, but since it was a simple change I added it in.

## 🧪 Testing

1. Build app `pnpm dev:fresh`
2. Login as admin
3. Go to a talent request
4. View new changes to the inbox component
5. Ensure they fulfill the acceptance criteria

## 📸 Screenshot

image

image


apps/web/src/pages/TalentRequests/components/TalentRequestTrackedUsersInbox/TalentRequestTrackedUsersInbox.tsx
esizer, line 1: Hmm, this works but the icons are very confusing. I think we need something esle since we don't exlusibly use these icons here and its not clear waht they are indicating without additional context.

Also, it seems the icons in the card changed, was that inentional?

esizer (commented):
Seems to work but I had a hard time testing this and found it a confusing interaction. We might need some context to help understand the icons.

To a lesser degree, I found the copy change was a ltitle confusing as well but that could just be me 😅

apps/web/src/pages/TalentRequests/components/TalentRequestTrackedUsersInbox/TalentRequestTrackedUsersInbox.tsx
yonikid15: Okay, should we push this back to design then? 

> Also, it seems the icons in the card changed, was that inentional?

Ya it's in the A/C (replace icon for "referred" with bookmark-square-mini on the status), and also you can see the [designs](https://www.figma.com/design/OcJlqTARk7UEG2aK7H0AG5/Talent-requests--Communities-?node-id=1308-22152&m=dev) for a full view of all the changes.

yonikid15:
@Jerryescandon @esizer 

The icons can be confusing without something like a Tooltip. We don't have a Tooltip component yet. I see that your design has a one though. 

My thinking to move this forward is: 

1. Use the actions button (with the EllipsesIcon), and include the bulk actions there. Similar to what we had before.
2. We build a Tooltip component, while this PR is set to draft. Once completed we can update the action icons with tooltips.
3. Other suggestions?

Jerryescandon:
1. I would prefer if we build the tool-tip component, otherwise there is little point in doing all this work. The fact that that it changes from a drop-down group to a single button is a big signal to the user that the actions have changed from bulk-everything to only bulk-selected.

2. On a different note, I forgot to update the screenshot but I was thinking it over and I think it would be best if we remove: `Mark all as referred` and `Mark all as not referred` from the drop-down. These two actions at this point in time should really be done on a case-by-case basis, they make sense when initially finding people to refer. But after they have been referred they seem dangerous to have as a bulk option.

3. Also `Mark all as selected` shouldn't even be there either. Neither in the bulk actions nor as an individual button. `Mark as selected` should only be accomplished by clicking on a user and changing it in the dialogue.

image


4. Ellipsis icon should be `ellipsis-horizontal` not vertical

5. The stroke width used in the icons is different in the individual actions vs the dropdown one. I'm not sure what that's about. but we should have the same width lines.

image

apps/web/src/pages/TalentRequests/components/TalentRequestTrackedUsersInbox/TalentRequestTrackedUsersInbox.tsx
```diff
@@ -266,96 +269,155 @@ const TalentRequestTrackedUsersInbox = ({
...
+            <Tooltip.Provider>
+              <Tooltip.Root>
+                <Tooltip.Trigger
+                  render={
+                    <IconButton
+                      color="black"
+                      icon={
+                        statusIcons.outline[
+                          TalentRequestTrackedUserStatus.Referred
+                        ]
```
esizer: Chekc firs tot confirm but, im not sure this needs to be an array,. I think we can pass the icon in directly.

esizer (changes requested):
Thanks for doing the tooltips. This is much better now! 

However, I found an issue. If you do the flow with a keyboard, once the dialog closes, you lose your focus and it is very disorrienting for a screen reader user.

We will need to apply some focus management once the dialog closes. I think maybe just send it to the "select all" checkboxc if possibler? 🤔 Or, you can maybe see if @Jerryescandon has any opinions on the matter.

Jerryescandon:
> However, I found an issue. If you do the flow with a keyboard, once the dialog closes, you lose your focus and it is very disorrienting for a screen reader user.

My instinct is to go to the next item. But I could be wrong, But I would imagine, If I start going down a list, Person A, then Person B, then I click on person B open up a dialogue and then it closes, and person B is gone, I should just go down to person C next?

esizer:
> > However, I found an issue. If you do the flow with a keyboard, once the dialog closes, you lose your focus and it is very disorrienting for a screen reader user.
> 
> My instinct is to go to the next item. But I could be wrong, But I would imagine, If I start going down a list, Person A, then Person B, then I click on person B open up a dialogue and then it closes, and person B is gone, I should just go down to person C next?

Oh sorry, this was referring specifically to the bulk actions. Though, I think we could improve the individual rows as well. When that happens your focus gets lost but it gets placed in some phantom spot where the old user was :ghost: :sweat_smile: 

I can see about creating a ticket to do it the way you descrbied since that makes sense to me (its how our notifications work when we delete them) :+1:

Jerryescandon:
Well, I guess we can kinda extrapolate a general rule. like, If a dialogue interrupts a flow when that dialogue closes it either:
 A) goes back to the item, if it still exists, or 
 B) goes to the next one in the tab order.

yonikid15:
@esizer I made a fix, let me know what you think

[Fix focus management with bulk mutations](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17579/commits/400a69211c0717374227becdb66737cdbf749887)

apps/web/src/pages/TalentRequests/components/TalentRequestTrackedUsersInbox/TalentRequestTrackedUsersInbox.tsx
yonikid15: Fixed in [Refactor icons](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17579/commits/5159cfb7d6fabbc225fb5c9a6243147d31c177a8)

packages/forms/src/components/CheckButton/CheckButton.tsx
```diff
@@ -37,43 +38,49 @@ export interface CheckButtonProps extends CheckButtonVariants {
...
-  checked,
-  label,
-  onToggle,
-  indeterminate = false,
-  color = "black",
-  className,
-}: CheckButtonProps) => {
-  const intl = useIntl();
-  const { btn, icon: iconStyles } = checkBtn({ color });
+const CheckButton = forwardRef<HTMLButtonElement, CheckButtonProps>(
```
esizer, line 41: Ref can now be passed as a prop. We should no longer be using `forwardRef` since it has been deprecated.

```tsx
interface IconButtonProps {
  ref: Ref<HTMLButtonElement | null>;
}
```

SEE: 

 - https://react.dev/blog/2024/12/05/react-19#ref-as-a-prop
 - https://react.dev/reference/react/forwardRef

esizer (changes requested):

esizer:
@Jerryescandon Do you think it woukld make sense if, once we complete this ticket, we update this to be [a toolbar](https://base-ui.com/react/components/toolbar)? 

This would allow arrow keys to navigate the controls which I intuitvlry did only to find out that did not work and had to use tab instead.

Jerryescandon:
> @Jerryescandon Do you think it woukld make sense if, once we complete this ticket, we update this to be [a toolbar](https://base-ui.com/react/components/toolbar)?

I didn't even know this was a thing, but yeah it makes sense. It seems completely like an implementation change right? or should it have any distinctive design changes?

packages/forms/src/components/CheckButton/CheckButton.tsx
yonikid15: Fixed in [Use ref prop](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17579/commits/630d919ee69ceb7011363e26bd37ef1642cf0403)

esizer (approved):

PR #17580, opened 2026-07-28
[Fix] Add skill rankings update validation

### LINKED ISSUE #17407: 🐛 updateUserSkillRankings accepts nonexistent skill IDs
### Bug

`updateUserSkillRankings` passes bare UUID arrays through `User::addSkills()` (`User.php:447`) which creates `UserSkill` records without checking that the skill IDs exist in the `skills` table. The PostgreSQL foreign key constraint catches the error, but it surfaces as an unhandled `QueryException` instead of a clean GraphQL validation error.

Same class of bug as #17406.

Split from #17261.

### Steps to Reproduce

1. Authenticate as any user.
2. Call `updateUserSkillRankings` with a `topTechnicalSkillsRanked` array containing a well-formed but nonexistent UUID.
3. The request fails with an internal server error (foreign key violation) instead of a validation error.

### Proposed Solution

Add `exists:skills,id` validation to the individual items in `UpdateUserSkillRankingsInput`. For example, using Lighthouse's `@rulesForArray` on each field:

```graphql
topTechnicalSkillsRanked: [UUID]
  @rulesForArray(apply: ["nullable", "App\\Rules\\ArrayIsUnique"])
  @rules(apply: ["exists:skills,id"])
```

Or add the check in the resolver / `addSkills()` method.

### Acceptance Criteria

- [ ] `updateUserSkillRankings` returns a validation error when any skill ID does not exist in the `skills` table
- [ ] Existing skill ranking functionality is unaffected

esizer:
🤖 Resolves #17407 

## 👋 Introduction

Adds validation for the mutation to update skill rankings. Adds a rule to ensure skills exit as well.

## 🧪 Testing

1. Attempt to update a skill ranking for a non-existant skill
2. Confirm you get a validation error
3. Confirm test coverage exists

### DIFF
--- a/api/app/GraphQL/Validators/UpdateUserSkillRankingsInputValidator.php
+++ b/api/app/GraphQL/Validators/UpdateUserSkillRankingsInputValidator.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\GraphQL\Validators;
+
+use App\Enums\ErrorCode;
+use App\Rules\ArrayIsUnique;
+use Nuwave\Lighthouse\Validation\Validator;
+
+final class UpdateUserSkillRankingsInputValidator extends Validator
+{
+    /**
+     * Return the validation rules.
+     *
+     * @return array<string, array<mixed>>
+     */
+    public function rules(): array
+    {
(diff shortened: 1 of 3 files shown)


vd1992 (approved):
Also tested in showcase to make sure mutation still worked
Nice

PR #17587, opened 2026-07-29
[Feat] Require sign in talent request submission

### LINKED ISSUE #17456: ✨ Require sign-in to submit talent request
### Details

We would like to require users to be authenticated before they can submit a talent request. They will still be able to access and use the search page but, they must authenticate before accessing the page to submit the request.

The important part to consider will be to persist the values during the page transitions from search -> login flow -> request.

Sionce the user will be authenitcated, you can prepopulate:

 - Government of canada email (if they have a verified work email)
 - Job title (likely from current substantive experience?)
 - Full name
 - Department (likely from current substantive experience?)

We are not removing these fields due to the requirement to support the legacy data and offer the person a change to submit differing information.

### Design file & designer

[Figma file](https://www.figma.com/design/oUIFoHHAZ5FwFcGHc8p7Yt/Manager---Request-form?node-id=1125-9315&t=qpk7cSSipyKz2jBW-1) by @Jerryescandon 

### 📸 Screenshots

Image

### 🌎 Localization

_No response_

### Proposed Implementation

This might be tricky due to the fitlers, currently we use router state. However, you will be navigating away from the application during authentication so another mechanism must be used.

Thankfully, the "Apply now" button on the job posters has a solid example where we use session storage to store the pool ID. I reccomend following this example.

The main difference here is, this is a true navigation and not a submision yet or running a mutation so, while the apply button is a form submission, this can be a link which should simplify that portion.

### ✅ Acceptance Criteria

- [ ] Attempting to submit a talent request while unauthenticated, redirects to login
- [ ] Filters applied persist to the request page (no regression)
- [ ] Full name field replaced with static version
- [ ] Request form pre populated with best guess information from known user

esizer:
🤖 Resolves #17456 

## 👋 Introduction

Updates the talent request submission form the require an authenticated user.

## 🕵️ Details

Since this requires us to navigate away from the site, we had to change how we persist the filters applied. Instead of using location state, we now store it in session storage.

The downside to this is, session storage is more persistent so, I did two things to try and keep parity with the old location state method. We now clear this specific session storage when"

- The submission is successful
- The user is navigating to a page that is not the search page
  - The search page exception is here so that we maintain the old functionality of navigating backwards to edit the request before submission wihtout starting over 

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Ensure you are not authenticated
3. Navigate to `/search`
4. Fill out the form keeping in mind what filters you selected
5. Attempt to submiot the filters to continue to the request form
6. Confirm you are redirected to login info
7. Login as any user
8. Confirm you return to the request page
9. Confirm the fields are prepopulated with your information
10. Confirm the filters you set in step 4 are persisted here
11. Attempt to navigate back to the search form
12. Confirm the form is not reset but keeps your filters
13. Confirm you can still submit the request successfully
14. Confirm going back to `/search` after submissions presents with an empty form


apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
```diff
@@ -293,8 +313,27 @@ export const RequestForm = ({
     "TalentRequestSource",
   );
 
+  // The search form repopulates from the stored request state, so only clear
+  // that state when leaving for anywhere else.
+  const blocker = useBlocker(
```
vd1992, line 313: **The problem.** Since the state is cleared on every navigation except the one to `/search`, pressing Back onto this page gives you a form that looks complete but can't be submitted: the summary reads `Group and level: (None selected)` / `no candidates`, while the contact fields survive in the separate `ts-createRequest` key. Submitting from there fails server-side on `applicant_filters.community_id`, and all the user sees is "Error: creating request failed". There's no way out from the page either, because the Back link reads the same cleared key, so the search form comes up empty and the search has to be redone from scratch. The same thing happens after a successful request: Back onto this page shows a freshly prefilled form above an empty filter summary.

**How it was verified.** I ran identical steps in Chromium against the local stack, once on this branch and once on a `main` build: IT-01 / Business Line Advisory Services / NCR -> *CMO Digital Careers* -> nav link to Browse jobs -> browser Back. `main` comes back with `IT-01` and `1 estimated candidate`; this branch shows `(None selected)` / `no candidates`, and `talentRequestState` was `null` by then. The submit failure was traced to `laravel.log` (`SQLSTATE[23502]` on `community_id`), and a second reviewer reproduced the whole sequence. Two nearby worries turned out fine: the language toggle is a plain `<a>`, so switching locale doesn't trip the blocker, and an apparent "Back needs two presses" effect was just timing in my test harness.

**Suggestion.** Clear on successful submission — an explicit `removeFromSessionStorage(TALENT_REQUEST_STATE_KEY)` in `onSubmit` — rather than inferring it from the user leaving. Then guard the page: with no stored filter state, send them to `/search` (or show an explicit "start a search" state) so a stripped form is never reachable. That also covers a direct hit on `/en/search/request`, which today renders an empty summary above a form that looks ready to submit. If the blocker stays as the mechanism, exempting the confirmation route only fixes the post-submit half and leaves nav-away -> Back broken.

apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
```diff
@@ -293,8 +313,27 @@ export const RequestForm = ({
...
+  }, [blocker]);
+
   const formMethods = useForm<FormValues>({
-    defaultValues: getFromSessionStorage(cacheKey, {}),
+    defaultValues: {
+      fullName: getBasicFullNameLabel(user?.firstName, user?.lastName),
+      email: user?.workEmail ?? "",
+      managerJobTitle: user?.role ?? "",
+      department: user?.department?.id,
+      ...getFromSessionStorage(cacheKey, {}),
```
vd1992: **The problem.** The cache is spread after the identity defaults, so it wins for every field it holds — and because `watch()` writes the whole form object on the first change, a single keystroke freezes all four prefilled fields for the rest of the tab session. Sign-out only clears the localStorage tokens, so both session-storage keys survive it, and the next person to sign in on that tab inherits a stranger's name, work email, job title and department. So the prefill really only works in a tab where nobody has touched this form yet. The cache could hold stale text before, but it now holds identity the user never typed, and that's what turns a stale-value annoyance into someone else's details.

**How it was verified.** Signed in as Dale Monroe, confirmed all four fields prefilled with `ts-createRequest` still absent on arrival (so the cache is written on interaction, not on load), typed one field, then signed out and signed in as Dara Kennedy in the same tab — the form still showed Dale's four values. Running `sessionStorage.removeItem('ts-createRequest')` and reloading gave Dara's, which points at the cache rather than the query. A second reviewer reproduced this, including the `removeItem` check.

**Suggestion.** Either key the cache per user (`ts-createRequest:<userId>`) or clear app-owned session keys on logout; the logout route fixes two things at once, since `useApplyToPool`'s `APPLY_ON_RETURN_KEY` has the same exposure today. One thing to watch: clearing the cache inside the blocker effect looks like a fix but isn't, because sign-out is a full page load via `window.location.href` and the blocker never runs. Either option also restores AC #4 for returning users, since a fresh sign-in then starts from the prefilled defaults.

vd1992 (commented):
Five findings from a QA pass on this branch, posted inline: filter state lost on Back navigation (1), the request-form cache overriding the new prefill and carrying across users in a tab (2), full name still editable against AC #3 (3), unverified work emails being prefilled (4), and the sign-in requirement not enforced at the API (5).

Highest priority is (1) — it's a confirmed regression against `main` on the main flow, and there's no in-page recovery once it happens. Next is (2), which silently defeats the prefill and can carry a previous user's name and work email into a submitted request. (3) and (4) are AC gaps, and (5) is a scope call rather than something this PR has to absorb.

Posting this to help @yonikid15's review.

apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
```diff
@@ -293,8 +313,27 @@ export const RequestForm = ({
...
+    if (blocker.state === "blocked") {
+      removeFromSessionStorage(TALENT_REQUEST_STATE_KEY);
+      blocker.proceed();
+    }
+  }, [blocker]);
+
   const formMethods = useForm<FormValues>({
-    defaultValues: getFromSessionStorage(cacheKey, {}),
+    defaultValues: {
+      fullName: getBasicFullNameLabel(user?.firstName, user?.lastName),
```
vd1992, line 326: **The problem.** AC #3 and the design on #17456 — where the arrow points right at it — show full name as static text, but the Full name `Input` below is still editable. The design keeps job title, email and department editable, so the issue's "chance to submit differing information" rationale doesn't really extend to this field.

**How it was verified.** Compared the rendered form field-by-field against the screenshot in the issue.

**Suggestion.** Worth knowing before swapping in `FieldDisplay`: `fullName` is `required` in both the form rules and `CreateTalentRequestValidator`, and it can legitimately come out empty. `first_name`/`last_name` are nullable, `AuthController` grants the applicant role at auto-registration, and names only get set from `given_name`/`family_name` ID-token claims — so someone with no name on record still passes this route's `RequireAuth`. A straight swap would leave them with a blank required field and no way forward. A fallback works (editable input when empty, or the static value plus a link to their profile), though which one is probably a question for @Jerryescandon.

apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
```diff
@@ -293,8 +313,27 @@ export const RequestForm = ({
...
+      email: user?.workEmail ?? "",
```
vd1992, line 327: **The problem.** The AC scopes this to "if they have a verified work email", and prefilling regardless makes an address the user hasn't proven they own the default contact address on a submitted request. Everywhere else the app treats that state as meaningful — `WorkEmailCard` offers "Re-verify work email", and internal-jobs applications require verification through `hasEmptyRequiredFields` — so this form ends up being the one place it doesn't matter. And since `TalentRequestObserver` stamps the request with the signed-in user, the account and the contact email can quietly disagree.

**How it was verified.** `applicant@test.com` has `work_email_verified_at` NULL in the local DB, and the field still prefills `Gul_Fields_Applicant@gc.ca`. Format isn't a safety net here, since `GovernmentEmailRegex` only checks for a gov domain, which unverified addresses usually satisfy.

**Suggestion.** `isWorkEmailVerified` already exists on `type User`, so no schema or resolver work is needed — add it to `RequestFormDefaultUser` and gate the default with `user?.isWorkEmailVerified ? user.workEmail ?? "" : ""`.

apps/web/src/pages/SearchRequests/RequestPage/RequestPage.tsx
```diff
@@ -45,17 +27,19 @@ export const Component = () => {
...
-        />
+        <CreateRequest />
       </Card>
     </Hero>
   );
 };
 
+// NOTE: Require authentication but allow any user role
+export const Component = () => (
+  <RequireAuth roles={[ROLE_NAME.Applicant]}>
```
vd1992, line 38: **The problem.** This gate controls who can reach the form, not who can submit one: `createTalentRequest` has no `@guard` and `TalentRequestPolicy::create` returns `true`, so an unauthenticated POST still creates a request. Those rows land with `user_id` NULL and `status = NEW`, so they're indistinguishable in the admin queue — which also means the new attribution only holds for requests that come through the SPA.

**How it was verified.** An unauthenticated `curl` to `/graphql` with a minimal valid input returned an id, and the row is there with `user_id` NULL. I also read the policy and the mutation's directives to confirm nothing guards that path.

**Suggestion.** If anonymous submission isn't worth keeping, `@guard` on the mutation plus a policy requiring a user closes it. If it is deliberate, then `user_id` shouldn't be treated downstream as "who submitted this". Either way this feels like a scope call for the team rather than something this PR has to absorb.

apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
Jerryescandon: I am not sure I understand.

vd1992: The bot is getting ahead of itself 🤣 
Its suggestion is probably wordier than it needs to be, needs refinement still. 

Basically, whether there should be some fallback case where someone tries to submit a request with no name. 
In which case if the field display is locked on that page by being a static full name field, then they can't edit it to fix it

To test, I manually nulled out my name then navigated to this page. Name is not prefilled as a result

image

vd1992: The question then becomes, should it match the design in the below screenshot?
If so, should it special handle and if so how, cases where a user with a null name tries to submit a request

image

yonikid15 (commented):
Looking really good Eric 👍 

1. This is a design question... When a user navigates away from the request page  `/request`, and then goes back. What should the request form look like? Currently, the session only holds the contact information, a.k.a the UserGovInfo. Unless it's been manually changed by the user (e.g. they change the department). However, the rest of the data is removed (all the fields below the contact information section).

2. I can't remember 😅 .. does the users work email NEED to be verified to be able to submit a talent request (I think we said it doesn't)? If not, then we can ignore a couple of the Vachan + Claude comments https://github.com/GCTC-NTGC/gc-digital-talent/pull/17587#discussion_r3677358708, https://github.com/GCTC-NTGC/gc-digital-talent/pull/17587#discussion_r3677358711. If it is, then we need to add a guard and update the create policy.

3. I wasn't able to verify this new flow works, after creating a new user, and all the fields are still populated. I might have found a separate bug with the email verification. (Looking into it now)

apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
yonikid15: I'm not sure I understand? The front and back end validation prevents creating a talent request without a full name. 

One question I do have @Jerryescandon is,.. does the user have the freedom to change the contact information as they choose? Or does it have to match their database information? For example, if I'm a verified user, can I submit a request on behalf of someone else, and just change the name, email, etc to that persons info.

apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
yonikid15: @vd1992 have you manually tested this yourself?

esizer:
> 1. This is a design question... When a user navigates away from the request page  `/request`, and then goes back. What should the request form look like? Currently, the session only holds the contact information, a.k.a the UserGovInfo. Unless it's been manually changed by the user (e.g. they change the department). However, the rest of the data is removed (all the fields below the contact information section).

This is how our forms work, It seems odd here becaue we are pre-populating those fields from the authed user which is in the AC. The others only cache when edited which is expected with our `UnsavedChanges` component. So, im geussing this is intended and expected.

> 2\. I can't remember 😅 .. does the users work email NEED to be verified to be able to submit a talent request (I think we said it doesn't)? If not, then we can ignore a couple of the Vachan + Claude comments [#17587 (comment)](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17587#discussion_r3677358708), [#17587 (comment)](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17587#discussion_r3677358711). If it is, then we need to add a guard and update the create policy.

It never was in the past, all we cared about is that you were an employee. Also, this would be trivial to add in the future if we deem it necessary. I think we can safely keep this oujt of the scope for now.


> 3\. I wasn't able to verify this new flow works, after creating a new user, and all the fields are still populated. I might have found a separate bug with the email verification. (Looking into it now)

Any update on this? Or maybe we can take a look together to see?

apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
vd1992: Yep

apps/web/src/pages/SearchRequests/RequestPage/RequestPage.tsx
vd1992: I think this one is actually important-ish

apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
Jerryescandon: Yeah, unfortunately thats one of the dangers of simply leaving it as a field, is that the user can change it before submitting. I would not add any additional validation to this. We will simply pre-populate it if they have anything, or leave it blank if they dont.

apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
yonikid15: Can you explain what this means more concisely because it's hard to follow this convoluted explanation?  Is the problem you are stating that the data is missing after leaving the search page and then returning back? Because that's what was intended to happen

apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
yonikid15: I was never able to recreate this.

vd1992: Observe. 

1. Logged in as Dale first 
2. Create but don't submit the request
3. Logout then login as Dara
4. Create request, go to submit page, Dale's information is pre-populated


https://github.com/user-attachments/assets/d8b3ce62-ed5f-4293-bce8-4905815ec245

apps/web/src/pages/SearchRequests/RequestPage/components/RequestForm.tsx
vd1992: If it is really convoluted, maybe just something to not worry about

yonikid15 (approved):
Looks like everything has been addressed 🎉

PR #17588, opened 2026-07-29
[Fix] Prevent status item badge from rendering a 0 count

brindasasi:
## Introduction
Fixes the status item count badge so an `itemCount` of 0 doesn't render a stray "0" (previously `0 && <span>` output the literal 0 instead of nothing).

## Testing
View a status item with a 0 count and confirm no badge/number is shown.


mnigh (commented):
i'm not sure why we wouldn't want to render 0. in any case, the optional prop `itemCount` on the `StatusItem` component is never used. we might consider removing the prop altogether.

## Screenshot

Image

mnigh:
> i'm not sure why we wouldn't want to render 0. in any case, the optional prop `itemCount` on the `StatusItem` component is never used. we might consider removing the prop altogether.
> 
> ## Screenshot
> Image

@GCTC-NTGC/developers anyone know a reason to keep the unused optional prop `itemCount` on the `StatusItem` component?

esizer:
> @GCTC-NTGC/developers anyone know a reason to keep the unused optional prop `itemCount` on the `StatusItem` component?

:axe:

mnigh:
> > @GCTC-NTGC/developers anyone know a reason to keep the unused optional prop `itemCount` on the `StatusItem` component?
> 
> 🪓

@brindasasi i suggest closing this PR and for someone to review (and approve 😜) https://github.com/GCTC-NTGC/gc-digital-talent/pull/18010.

PR #17613, opened 2026-07-30
[Platform] Free more space in vitest flow

### LINKED ISSUE #17401: 🛠️ Vitest workflow out of space
### 🛠️ Platform

I've started seeing the Vitest workflow occasionally run out of space.

Image

### 🕵️ Details

https://github.com/GCTC-NTGC/gc-digital-talent/actions/runs/29356124127/job/87164104968

### 🙋‍♀️ Proposed Solution

Claude think we can free up at least 14GB more.

Image

### ✅ Acceptance Criteria

- [ ] Workflow frees up more space before starting testing.
   - [ ] Bonus: if this works I think we have other workflows that try to free up space, too

petertgiles:
🤖 Resolves #17401 

## 👋 Introduction

Updates the "reclaim runner space" step to reclaim more space.  Available space went from 103GB to 116GB. (29% used down to 21% used.)  However, the run time goes up from 9s to 40s.  Not a clear value-add.  :face_with_diagonal_mouth: 

## 🕵️ Details

Rather thank copy-pasting these changes to our various workflow files I factored it out to a composite action that can be shared.
https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action

Note that the order of the steps change.  The composite action has to run after "checkout" since the action is in the codebase.

## 🧪 Testing

- [ ] Code looks good
- [ ] All workflows pass

### DIFF
--- a/.github/actions/reclaim-space/action.yml
+++ b/.github/actions/reclaim-space/action.yml
@@ -15,4 +15,4 @@ runs:
           sudo rm -rf /opt/hostedtoolcache/CodeQL
           docker system prune -af --volumes || true
           df -h
-      shell: bash
+        shell: bash
--- a/.github/actions/reclaim-space/action.yml
+++ b/.github/actions/reclaim-space/action.yml
@@ -2,17 +2,17 @@ name: "Reclaim space on runner"
 runs:
   using: "composite"
   steps:
-    - name: Reclaim space on runner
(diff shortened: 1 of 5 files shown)


esizer (approved):
oh cool! I like this. Now, hopefully it works 🤞

PR #17618, opened 2026-07-30
[Refactor] Update playwright graphql imports

esizer:
Related: #17227

## 👋 Introduction

This is part two of removing the full object types from graphql codegen.

## 🕵️ Details

We heavily rely on them in playwright and since it is not application code, I added this as another exception to the rule like we did with `fake-data`. 

If you object, let me know and I can see about other options. The other one I had in mind is to also generate fragments for this but then they would be in the client-preset generarted code which sort of seemed less desirable? I'm not sure :sweat_smile: 

## 🧪 Testing

1. Confirm playwright still works
2. Confirm no imports from the base graphql package for types

### DIFF
--- a/apps/playwright/fixtures/Department.ts
+++ b/apps/playwright/fixtures/Department.ts
@@ -4,8 +4,8 @@ import type {
   CreateDepartmentInput,
   LocalizedStringInput,
   UpdateDepartmentInput,
-} from "@gc-digital-talent/graphql";
-import { DepartmentSize } from "@gc-digital-talent/graphql";
+} from "@gc-digital-talent/graphql/schema-types";
+import { DepartmentSize } from "@gc-digital-talent/graphql/schema-types";
 
 import dConfig from "~/constants/config";
 import { loginBySub } from "~/utils/auth";
--- a/apps/playwright/fixtures/ExperiencePage.ts
+++ b/apps/playwright/fixtures/ExperiencePage.ts
@@ -7,7 +7,7 @@ import type {
   CommunityExperienceInput,
   AwardExperienceInput,
   EducationExperienceInput,
-} from "@gc-digital-talent/graphql";
+} from "@gc-digital-talent/graphql/schema-types";
 
 import AppPage from "./AppPage";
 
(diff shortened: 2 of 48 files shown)


petertgiles:
Related thoughts while looking at this PR: How are we going to enforce only using the full types in appropriate places (just PW and faker)?  What's the goal of this anyway?  Keep the bundles smaller?  Why don't we just update our uses of the full types to the second file everywhere and not worry about a refactor?

petertgiles (approved):
It's all type import changes and TS and PW are happy. 👍

esizer:
> How are we going to enforce only using the full types in appropriate places (just PW and faker)?

We have an eslint rule to restrict imports for this. It will error when importing from that file and then we simply disabled it in the fake data package. Playwright does not extend this config so no disabled was needed here.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/98c7eef2a68b1d17275af691d16f6255bfcb7c39/packages/eslint-config-custom/react.mjs#L208-L215

> What's the goal of this anyway?

[The migration guide](https://the-guild.dev/graphql/codegen/docs/migration/operations-and-client-preset-from-5-0#1-object-types-are-no-longer-generated) outlines why this change is being made. 

The bundle size will not change since type info is srtripped at build time anyway. 

The main imprvoement is encourage use of fragments and interfaces to avoid the need to overfetch. Since generating these types requires that non-nullable types become required, it forces us to query them or omit them and do a lot type wrangling. Fragments were the first major step to avoid these types of DX annoyances (`X is required in type Y` errors).

This allows the fragments to define the shape of the type for us rather than needing to constantly do `Pick<User, "firstName" ...>` or fetch fields we don't need just to satisfy the type.

If we simply side step this, we avoid this pain now but we keep both the type wrangling and overfetching issues the changes are attemping to avoid.

Also, I tried to do a smal wrtie yup in [the parent issue](https://github.com/GCTC-NTGC/gc-digital-talent/issues/17227) as well about why this is happening.

---

I hope that makes sense :sweat_smile:

petertgiles:
Ah, right, right.  Thanks for bearing with me.  :sweat_smile: 

Carry on...

esizer:
> Ah, right, right. Thanks for bearing with me. 😅
> 
> Carry on...

It was a good question, don't want to be refactoring/merging things for no good reason. Always good to question why things are being done :stuck_out_tongue:

PR #17620, opened 2026-07-30
[Bug] Apply CSP nonce to dialog scroll-lock styles

### LINKED ISSUE #17182: 🐛 Prod - CSP Violation on dialogs

We're getting a CSP violation every time a dialog is opened.

Now that we've fixed CSP reporting 🤞we should fix the actual source of the error.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

Open a dialog anywhere in the app.

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] No CSP errors when opening a dialog.

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

vd1992:
🤖 Resolves #17182

## 👋 Introduction

Opening any dialog produced a content security policy error in the browser console. The policy only permits stylesheets carrying a one-time token issued per page load, and the library behind our dialogs was inserting one without it, so the browser blocked it.

That blocked stylesheet is also what holds the page still while a dialog is open, so it had been failing silently too. With the token applied, the background no longer scrolls behind an open dialog, and scrolling inside a dialog no longer carries over to the page underneath once it hits the end.

## 🕵️ Details

**Solution**

At client startup the nonce is read off the document and handed to `get-nonce`, the module `react-style-singleton` consults when it builds its `<style>` tag. Radix's dialog reaches that code through `react-remove-scroll`, so one call at the entry point covers every dialog in the app. It has to be read from the DOM because nginx only substitutes the nonce into `index.html`, never into the hashed JS bundles, so there is nothing to bake in at build time.

**Justification**

The Radix PR linked from the issue adds a `nonce` prop, but only to `ScrollArea.Viewport` and `Select.Viewport`; neither package is in our tree, and `react-remove-scroll` exposes no equivalent prop, which leaves the `get-nonce` singleton as the supported hook. Copying `setNonce` locally would not work — the value has to land in that module's own variable, because that is the one the library reads back. `get-nonce` was already installed transitively, so declaring it directly adds no install weight and stops the import depending on hoisting. The one dependency-free alternative, setting `window.__webpack_nonce__`, was rejected as it needs a global declaration for a webpack-named hook that Vite never sets.

**Investigation and testing**

The stack trace in the issue pointed at `insertStyleTag` in `react-style-singleton`, which reads the nonce at tag-creation time, so the fix only has to run before the first dialog opens. In Chromium against the local stack, every dialog trigger on a job advertisement page plus an open/close/reopen cycle produced zero `securitypolicyviolation` events, while a control build with only the nonce lookup broken reproduces the error and leaves body `overflow: visible` instead of `hidden`. Nonce hiding still applies to the injected tag — `getAttribute("nonce")` returns `""`, matching the parsed elements — so the nonce is no more exposed to CSS-based exfiltration than it was before. Retested by hand in Firefox, including an admin form dialog closing on submit.

## 🧪 Testing

1. Open `/en/jobs`, click any job, and open the browser console.
2. Click **Learn more about skill levels**. No `style-src-elem` error should appear.
3. With the dialog open, run `[...document.querySelectorAll('style')].map(s => s.nonce)` — it should print one non-empty value. (The Elements panel shows `nonce=""`; only the JS property carries the real value.)
4. Still with the dialog open, scroll the page behind it. It should not move.
5. Close the dialog and scroll again. Normal scrolling should return, on this and on any dialog that closes by submitting a form.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Claude Opus 5 (1M context)

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -56,6 +56,7 @@
     "@tailwindcss/vite": "^4.3.3",
     "@tanstack/react-table": "^8.21.3",
     "date-fns": "^4.4.0",
+    "get-nonce": "^1.0.1",
     "graphql": "^16.14.2",
     "lodash": "^4.18.1",
     "motion": "^12.42.2",
--- a/apps/web/src/entry.client.tsx
+++ b/apps/web/src/entry.client.tsx
@@ -2,6 +2,10 @@ import { StrictMode } from "react";
 import { hydrateRoot } from "react-dom/client";
 import { HydratedRouter } from "react-router/dom";
 
+import { applyCspNonce } from "./utils/csp";
+
+applyCspNonce();
+
 hydrateRoot(
   document,
   <StrictMode>
--- a/apps/web/src/utils/csp.test.ts
+++ b/apps/web/src/utils/csp.test.ts
@@ -2,6 +2,14 @@ import { getNonce, setNonce } from "get-nonce";
 
 import { applyCspNonce } from "./csp";
 
+/**
+ * Limitation: jsdom has no CSP, so it never blanks the `nonce` content
+ * attribute the way a real browser does. Reading the attribute instead of the
+ * `nonce` property would therefore still pass here while silently breaking in
+ * production. That swap can only be caught in a browser served the CSP header
+ * — see the manual steps in the pull request.
+ */
+
 const addElementWithNonce = (nonce: string) => {
   const script = document.createElement("script");
   script.setAttribute("nonce", nonce);
--- a/apps/web/src/utils/csp.test.ts
+++ b/apps/web/src/utils/csp.test.ts
@@ -1,0 +1,30 @@
+import { getNonce, setNonce } from "get-nonce";
+
+import { applyCspNonce } from "./csp";
+
+const addElementWithNonce = (nonce: string) => {
+  const script = document.createElement("script");
+  script.setAttribute("nonce", nonce);
+  document.head.append(script);
+};
+
+describe("applyCspNonce", () => {
+  beforeEach(() => {
+    document.head.innerHTML = "";
+    setNonce(""); // falsy, so `getNonce` falls back to reporting no nonce
+  });
+
+  it("shares the document nonce with get-nonce consumers", () => {
+    addElementWithNonce("abc123");
+    applyCspNonce();
+
+    expect(getNonce()).toBe("abc123");
+  });
+
+  it("ignores the un-substituted placeholder", () => {
+    addElementWithNonce("**CSP_NONCE**");
+    applyCspNonce();
+
+    expect(getNonce()).toBeUndefined();
+  });
+});
--- a/apps/web/src/pages/SearchRequests/RequestConfirmationPage/RequestConfirmationPage.tsx
+++ b/apps/web/src/pages/SearchRequests/RequestConfirmationPage/RequestConfirmationPage.tsx
@@ -2,6 +2,7 @@ import type { ReactNode } from "react";
 import { useRef } from "react";
 import { defineMessage, useIntl } from "react-intl";
 import { useReactToPrint } from "react-to-print";
+import { getNonce } from "get-nonce";
 
 import {
   Notice,
@@ -76,6 +77,9 @@ export const Component = () => {
   const handlePrint = useReactToPrint({
     contentRef: componentRef,
     pageStyle: printStyles,
+    // Printing injects `pageStyle` as a <style> tag, which our CSP blocks
+    // without the nonce the app shares through `get-nonce` at startup.
+    nonce: getNonce(),
     documentTitle: intl.formatMessage({
       defaultMessage: "Request submitted",
       id: "0zo274",
--- a/apps/web/src/utils/csp.ts
+++ b/apps/web/src/utils/csp.ts
@@ -1,0 +1,26 @@
+import { setNonce } from "get-nonce";
+
+/**
+ * Placeholder that nginx swaps for the per-request nonce when it serves
+ * `index.html`. If it reaches the browser un-substituted (dev server, tests)
+ * there is no real nonce to share.
+ */
+const NONCE_PLACEHOLDER = "**CSP_NONCE**";
+
+/**
+ * Hand the nonce to `react-style-singleton`, which `react-remove-scroll` uses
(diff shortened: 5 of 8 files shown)


apps/web/package.json
```diff
@@ -56,6 +56,7 @@
     "@tailwindcss/vite": "^4.3.3",
     "@tanstack/react-table": "^8.21.3",
     "date-fns": "^4.4.0",
+    "get-nonce": "^1.0.1",
```
vd1992, line 59: As can be seen in PR body and commits, actually useful to import specified. I pushed to make sure this was actually the case

vd1992: And quoting 
>package pnpm was already installing for react-style-singleton

apps/web/src/utils/csp.test.ts
petertgiles, line 1: The AI flagged this for me:

> csp.test.ts can't catch a .nonce → .getAttribute("nonce") regression — jsdom doesn't blank the nonce attribute the way real browsers do, so a future refactor swapping the IDL property for getAttribute would pass CI while silently reintroducing the exact CSP scroll-lock bug this PR fixes. This is hard to close purely in jsdom (verified: even the "should differ" assertion would be meaningless there) — likely needs a real-browser (Playwright) check or at least a comment flagging the gap.

It seems like the easier option is to add a comment to highlight this limitation and the more complete option would be to move this test to Playwright.

apps/web/src/entry.client.tsx
petertgiles, line 1: Unfortunately this doesn't seem to fix every area.  I went to the admin talent nomination "evaluate a nomination" dialog because it is a nice long one that can scroll and I'm still getting the error.  Something still missing?

Image

petertgiles: OK, it seems like it's Tiptap that also needs the same fix.  The AI suggests adding get-nonce to the forms package and using it like this:

Image

> Root cause: the earlier PR only wired the CSP nonce into get-nonce, which react-remove-scroll reads for its dialog scroll-lock <style> tag. But the "Submit the evaluation" dialog also renders a Tiptap rich-text editor (for the rejection-reason field), and Tiptap injects its own <style> tag (for the gapcursor extension, etc.) through a completely separate mechanism — @tiptap/core's injectNonce editor option — which nothing was setting. That untagged <style> element was the one getting blocked.
> 
> Fix: [packages/forms/src/components/RichTextInput/ControlledInput.tsx](vscode-webview://14nu13bupmt3msjn628qsg4hqs0qtafvm5bho30reajlsd0vmmqf/packages/forms/src/components/RichTextInput/ControlledInput.tsx) now passes injectNonce: getNonce() (from the same get-nonce package applyCspNonce() already populates) into useEditor(). Added get-nonce as a direct dependency of packages/forms since it now imports it directly.
> 
> Verified live against your local nginx + built dist: before the fix, opening the dialog threw the style-src-elem violation from the Tiptap bundle chunk every time (including on close/reopen); after rebuilding, it's clean on first open and on close/reopen, and the editor toolbar/content still render correctly. Forms package tests and tsc pass.

petertgiles (changes requested):
Very cool.  I'm not super familiar with CSPs so this is a little above my head but AI to the rescue.

apps/web/src/utils/csp.test.ts
vd1992: Went with the comment. The gap is real: jsdom has no CSP, so it never blanks the `nonce` attribute, and a `.nonce` → `getAttribute("nonce")` swap would pass CI while breaking production.

`csp.test.ts` now says exactly that and points at the manual browser steps in the PR description.

For the record on the other option: a Playwright spec asserting zero `securitypolicyviolation` events would be the stronger guard, since that suite already runs against `localhost:8000` with the real CSP header — it would have caught the Tiptap gap below, not just an IDL-property swap. That's more than this PR should carry, but it's a good candidate for its own ticket.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Claude Opus 5 (1M context)

apps/web/src/entry.client.tsx
vd1992: Confirmed and fixed in d3c58cd. `@tiptap/core`'s `createStyleTag` only attaches a nonce when the editor is given `injectNonce`, so `ControlledInput` now passes `getNonce()` from the same singleton `applyCspNonce()` populates, and `packages/forms` declares `get-nonce` directly. One reassurance on scope: `createStyleTag` has exactly one caller, so there are no other injection points inside Tiptap to chase.

Verified on your repro — nomination group → "In progress" → "not supported" to reveal the rich text field. Clean on open *and* on close/reopen, which matters because `Editor.destroy()` removes the style tag, so it gets re-injected every time. A control build with `injectNonce` disabled reproduces the violation.

Rather than fix these one dialog at a time, I swept the built bundle for every runtime `<style>` injector that ships. Five:

- **react-style-singleton** (dialog scroll-lock) — fixed earlier in this PR
- **Tiptap** — fixed here
- **React DOM `<style precedence>`** — React-internal, not reachable from our code
- **motion `AnimatePresence mode="popLayout"`** — `MotionConfig` accepts a nonce, but `popLayout` appears nowhere in the repo
- **react-to-print** — accepts a `nonce` option nothing was passing. Fixed in 5807edc: printing a search request confirmation was throwing the same violation. Verified before/after, and the print preview now renders with its `pageStyle` applied.

So dialogs are fully covered, and the print flow was the one non-dialog instance of the same bug.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Claude Opus 5 (1M context)

apps/web/src/utils/csp.test.ts
petertgiles: Fair enough.  Please make an issue that we can action if we ever find it high enough priority.

apps/web/src/entry.client.tsx
petertgiles: Wow, I never would have found that react-to-print one.  Nice.

petertgiles (approved):

apps/web/src/utils/csp.test.ts
vd1992: Okay, sure https://github.com/GCTC-NTGC/gc-digital-talent/issues/17629

PR #17621, opened 2026-07-30
[Feature] PHP link checker

### LINKED ISSUE #16559: ♻️ Reduce false positives from Link Checker
### ♻️ Refactor

**Improvements to be done**  
1. isBrokenStatus only accepts HTTP 200. Any valid 2xx response other than 200 (e.g. 204 No Content, 206 Partial Content) is reported as a broken link.
2. connexion.canada.ca returns false positive 403
3. Legacy TLS single-URL spawn is dead code that re-run the entire checker
4. GET downloads full response bodies unnecessarily just to check status code
5. URLs in commented-out code is being checked

**Out of Scope** 
Not touching concurrency yet 

### 🕵️ Details

_No response_

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] isBrokenStatus - Accepts any 2xx (200–299), not just 200
- [ ] treat 403 from a known bot-protected host as non-broken 
- [ ] Dead spawn removed
- [ ] Use HEAD before GET 
- [ ] Skip commented urls

petertgiles:
🤖 Resolves #16559 

## 👋 Introduction

I know we've been having some trouble with our link checker.  Almost more of a spike but I worked with Claude to reimagine it as an Artisan command to take advantage of all the nice Laravel tooling.  I'm pretty happy with how it turned out.

## 🕵️ Details

- The main script goes from 402 lines of code down to 155.
- No separate respawning.  Laravel's HTTP client (built on curl) handles TLS and has builtin retry functionality.
- Laravel's HTTP client also has built-in concurrency. The script time goes from 20 seconds down to 4.8 seconds.
- The two links that require a browser UA are on a list to trigger that.  
- The two links that are protected by Cloudflare and the two that refuse connections from Github are moved to the block list.

## 🧪 Testing

1. `php artisan app:check-external-links`
2. Check for the two files in api/storage/app

I tried running this on my own fork of the repo.  You're welcome to do the same.  My fork doesn't have issues enabled to test that part, though.

https://github.com/petertgiles/gc-digital-talent/actions/workflows/external-link-checker-php.yml

## 📸 Screenshot

image

### DIFF
--- a/api/app/Console/Commands/CheckExternalLinks.php
+++ b/api/app/Console/Commands/CheckExternalLinks.php
@@ -5,23 +5,25 @@ namespace App\Console\Commands;
 use Illuminate\Console\Attributes\Description;
 use Illuminate\Console\Attributes\Signature;
 use Illuminate\Console\Command;
+use Illuminate\Http\Client\ConnectionException;
+use Illuminate\Http\Client\Pool;
+use Illuminate\Support\Facades\Http;
 use Illuminate\Support\Facades\Storage;
 use Illuminate\Support\Str;
 use Symfony\Component\Finder\Finder;
+use Throwable;
 
-#[Signature('app:check-external-links {--list-only : Only build the list of external links; do not check them (checking is not yet implemented)}')]
+#[Signature('app:check-external-links {--list-only : Only build the list of external links; do not check them}')]
 #[Description('Scan apps/web/src for external links and check them for broken responses.')]
 class CheckExternalLinks extends Command
 {
+    private const USER_AGENT = 'Mozilla/5.0 (compatible; LinkChecker/1.0; +https://github.com/GCTC-NTGC/gc-digital-talent)';
+
     /**
      * Execute the console command.
      */
     public function handle()
     {
-        if (! $this->option('list-only')) {
-            $this->warn('Link checking is not yet implemented; only building the list of external links.');
-        }
-
         $webSrcPath = config('linkchecker.web_src_path');
 
         $finder = new Finder();
@@ -55,7 +57,52 @@ class CheckExternalLinks extends Command
 
         $this->info("Scanned {$fileCount} files, found ".count($links).' unique external links.');
 
-        return Command::SUCCESS;
+        if ($this->option('list-only')) {
+            return Command::SUCCESS;
+        }
+
+        $brokenLinks = $this->checkLinks($links);
+
+        Storage::disk('local')->put('external-broken-links.json', json_encode($brokenLinks, JSON_PRETTY_PRINT));
+
+        $this->info(count($links).' links checked, '.count($brokenLinks).' broken.');
+
+        return count($brokenLinks) > 0 ? Command::FAILURE : Command::SUCCESS;
+    }
+
+    /**
+     * Check each link and return the ones that did not respond successfully.
+     *
+     * @param  array<int, array{file: string, url: string}>  $links
+     * @return array<int, array{file: string, url: string, status: int|string}>
+     */
+    private function checkLinks(array $links): array
+    {
+        $responses = Http::pool(fn (Pool $pool) => collect($links)->mapWithKeys(
+            fn ($link) => [$link['url'] => $pool->as($link['url'])
+                ->timeout(config('linkchecker.timeout'))
+                ->withUserAgent(self::USER_AGENT)
+                ->retry(2, 500, fn ($e) => $e instanceof ConnectionException, throw: false)
+                ->get($link['url'])]
+        )->all(), config('linkchecker.concurrency'));
+
+        $brokenLinks = [];
+
+        foreach ($links as $link) {
+            $response = $responses[$link['url']];
+
+            if ($response instanceof Throwable) {
+                $brokenLinks[] = ['file' => $link['file'], 'url' => $link['url'], 'status' => $response->getMessage()];
+
+                continue;
+            }
+
+            if (! $response->successful()) {
+                $brokenLinks[] = ['file' => $link['file'], 'url' => $link['url'], 'status' => $response->status()];
+            }
+        }
+
+        return $brokenLinks;
     }
 
     /**
@@ -75,7 +122,7 @@ class CheckExternalLinks extends Command
     }
 
     /**
-     * Mirrors the whitelisting logic from the original Node script.
+     * A link is worth checking if it's external and not blocklisted.
      */
     private function isValidExternalLink(string $url): bool
     {
--- a/api/app/Console/Commands/CheckExternalLinks.php
+++ b/api/app/Console/Commands/CheckExternalLinks.php
@@ -17,8 +17,6 @@ use Throwable;
 #[Description('Scan apps/web/src for external links and check them for broken responses.')]
 class CheckExternalLinks extends Command
 {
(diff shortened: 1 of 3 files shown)


petertgiles:
I tried the HEAD change.

> HEAD-only would cause false positives — plenty of servers 405 on HEAD or return different results than GET even when the page is fine. I'll do HEAD first, then only fall back to GET for links that failed, so most requests stay cheap but flaky ones get a fair check.

It adds quite a bit of complexity to the script and only saved 0.1s in my test.  Not worth it IMO.

brindasasi:
Really like the direction here, way simpler than the old script. Tested it locally
`Http::pool()` crashes intermittently, and the crash gets silently reported as "0 broken links" and a subsequent call succeeds .. 

```
php artisan app:check-external-links 

Scanned 1180 files, found 150 unique external links.

   GuzzleHttp\Promise\RejectionException 

  The promise was rejected with reason: Invoking the wait callback did not resolve the promise

  at vendor/guzzlehttp/promises/src/Create.php:59
     55▕         if ($reason instanceof \Throwable) {
     56▕             return $reason;
     57▕         }
     58▕ 
  ➜  59▕         return new RejectionException($reason);
     60▕     }
     61▕ 
     62▕     /**
     63▕      * Returns an iterator for the given value.

      +4 vendor frames 

  5   app/Console/Commands/CheckExternalLinks.php:79
      Illuminate\Support\Facades\Facade::__callStatic("pool")

  6   app/Console/Commands/CheckExternalLinks.php:62
      App\Console\Commands\CheckExternalLinks::checkLinks()

➜  api git:(62ae8b7c45) php artisan app:check-external-links

Scanned 1180 files, found 150 unique external links.
150 links checked, 0 broken.
➜  api git:(62ae8b7c45) php artisan app:check-external-links

Scanned 1180 files, found 150 unique external links.
150 links checked, 0 broken.

```

api/app/Console/Commands/CheckExternalLinks.php
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+     * Check each link and return the ones that did not respond successfully.
+     *
+     * @param  array<int, array{file: string, url: string}>  $links
+     * @return array<int, array{file: string, url: string, status: int|string}>
+     */
+    private function checkLinks(array $links): array
+    {
+        $responses = Http::pool(fn (Pool $pool) => collect($links)->mapWithKeys(
```
brindasasi, line 79: this needs to be wrapped in try catch and handle that crash something like "check failed to run" instead of silently leaving external-broken-links.json unwritten.

brindasasi (changes requested):
I don't know what's causing the crash .. probably concurrency ?

Image

I get 2 crashes in 4 runs.. 
I would prefer more to get that error logged using try catch  instead of leaving it unnoticed .. and then we can narrow down the root cause

api/app/Console/Commands/CheckExternalLinks.php
petertgiles: Agreed, the silent failure is not good.  Instead of trying to handle an an exception I just updated the workflow to crash on failures and not treat missing links as failures.
[fail on failure](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17621/changes/c8a150425e1e7c284992d30937fd60fd11c35341)

petertgiles:
> I don't know what's causing the crash .. probably concurrency ?

That's weird, I've never seen that but it's happening so often for you.  The AI thinks concurrency too.  I've removed the retry from in the pool and just added a very simple single retry for failures:
[fix retry logic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17621/commits/3fd5d79a5c5770cc6f9eb96fea1115acb241e524)

brindasasi:
> > I don't know what's causing the crash .. probably concurrency ?
> 
> That's weird, I've never seen that but it's happening so often for you. The AI thinks concurrency too. I've removed the retry from in the pool and just added a very simple single retry for failures: [fix retry logic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17621/commits/3fd5d79a5c5770cc6f9eb96fea1115acb241e524)

The fix actually increased crash % .  I tried different concurrency levels and this is what I get 

| Concurrency | Commit tested | Runs | Crashed | Rate |
|---|---|---|---|---|
| 10 (default) | `62ae8b7c45` (original) | 14 | 6 | ~43% |
| 10 (default) | `c8a150425e` ("fix retry logic" + "fail on failure") | 13 | 11 | ~85% |
| 5 | `c8a150425e` | 5 | 1 | 20% |
| 3 | `c8a150425e` | 5 | 5 | 100% |
| 1 | `c8a150425e` | 9+ | 0 | 0% |

Looks like it wouldn't perform well at concurrent levels.. concurrent level 3 is worse than 5. So there is no better level on that.. even if we remove the concurrency completely , I think this PR is still worth the try. No one is waiting for this to be finished faster

brindasasi:
> > > I don't know what's causing the crash .. probably concurrency ?
> > 
> > 
> > That's weird, I've never seen that but it's happening so often for you. The AI thinks concurrency too. I've removed the retry from in the pool and just added a very simple single retry for failures: [fix retry logic](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17621/commits/3fd5d79a5c5770cc6f9eb96fea1115acb241e524)
> 
> The fix actually increased crash % . I tried different concurrency levels and this is what I get
> 
> Concurrency	Commit tested	Runs	Crashed	Rate
> 10 (default)	`62ae8b7c45` (original)	14	6	~43%
> 10 (default)	`c8a150425e` ("fix retry logic" + "fail on failure")	13	11	~85%
> 5	`c8a150425e`	5	1	20%
> 3	`c8a150425e`	5	5	100%
> 1	`c8a150425e`	9+	0	0%
> Looks like it wouldn't perform well at concurrent levels.. concurrent level 3 is worse than 5. So there is no better level on that.. even if we remove the concurrency completely , I think this PR is still worth the try. No one is waiting for this to be finished faster

oh waait .. concurrency 0 gave a surprising result ..out of 15 runs there wasn't a single crash at all .. 
Time taken to run are the same as concurrency 10 . 


| Concurrency | Runs | Crashed | Rate | Avg. time (successful runs) |
|---|---|---|---|---|
| 0 (unlimited*) | 15 | 0 | **0%** | ~6.7s |
| 1 | 9+ | 0 | 0% | slower (sequential) |
| 3 | 5 | 5 | 100% | n/a |
| 5 | 5 | 1 | 20% | n/a |
| 10 (default) | 23 | 18 | ~78% | ~6.7s |

here is Claude's justification for the above result 

**Why:** `EachPromise`'s throttled concurrency window (used whenever `concurrency` is a number less than the total request count) has to add new promises to the shared curl_multi handle mid-flight as others complete. That incremental refill is what appears to trip a race in Guzzle's `Promise::waitIfPending()` safety check. At `concurrency=0`, every request is added up front with no refill needed; at `concurrency=1`, requests are fully sequential with no overlap. Both dodge the refill path entirely — only the throttled middle ground (3, 5, 10) exercises it.

**Suggested fix:** change `api/config/linkchecker.php`:
```php
'concurrency' => env('LINK_CHECKER_CONCURRENCY', 0),
```

brindasasi (approved):
LGTM!

PR #17622, opened 2026-07-31
[Refactor] Remove full graphql object types 3

esizer:
Related: https://github.com/GCTC-NTGC/gc-digital-talent/pull/17552

## 👋 Introduction

Removes full object types from `apps/web/src/pages/Pools`

## 🕵️ Details

I really hope this is the largest one :sweat_smile: This turned out being muhc larger than I anticipated so I apologize in advanced. I guess this goes to show how heavily we leaned on these large types instead of writing simpler ones by hand :sweat: 

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Test the different pools pages
3. Confirm no errors

### DIFF
--- a/apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/SkillSummaryTable.tsx
+++ b/apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/SkillSummaryTable.tsx
@@ -5,11 +5,9 @@ import type { IntlShape } from "react-intl";
 import { useIntl } from "react-intl";
 import CheckIcon from "@heroicons/react/20/solid/CheckIcon";
 
-import { commonMessages, getLocalizedName } from "@gc-digital-talent/i18n";
+import { commonMessages } from "@gc-digital-talent/i18n";
 import type {
   FragmentType,
-  LocalizedAssessmentStepType,
-  Skill,
   SkillSummaryPoolSkillFragment as SkillSummaryPoolSkillFragmentType,
   SkillSummaryTableAssessmentStepFragment as SkillSummaryAssessmentStepFragmentType,
 } from "@gc-digital-talent/graphql";
@@ -33,8 +31,7 @@ export const SkillSummaryTablePoolSkill_Fragment = graphql(/* GraphQL */ `
     type {
       value
       label {
-        en
-        fr
+        localized
       }
     }
     skill {
@@ -43,13 +40,11 @@ export const SkillSummaryTablePoolSkill_Fragment = graphql(/* GraphQL */ `
       category {
         value
         label {
-          en
-          fr
+          localized
         }
       }
       name {
-        en
-        fr
+        localized
       }
     }
   }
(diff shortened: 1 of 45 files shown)


apps/web/src/pages/Pools/EditPoolPage/components/SkillTable.tsx
petertgiles, line 1: Oops, looks like we lost the skill family names.

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/SkillSummaryTable.tsx
```diff
@@ -87,19 +80,13 @@ interface SkillSummaryTableProps {
...
 
 const CheckIconElement = (
-  skill: Skill | null | undefined,
-  assessmentStepType: LocalizedAssessmentStepType | null | undefined,
+  localizedName: string | null | undefined,
+  assessmentStepTypeLocalized: string | null | undefined,
 ): JSX.Element | null => {
   const intl = useIntl();
-  if (!skill) {
+  if (!localizedName) {
```
petertgiles: Before, if there was no name, it would still render will a fallback.  Now the icon is just gone from the table which seems much worse.

apps/web/src/pages/Pools/IndexPoolPage/components/helpers.tsx
```diff
@@ -32,31 +30,30 @@ import type { PoolBookmark_Fragment } from "./PoolBookmark";
...
     </Link>
   );
 }
 
-export function classificationCell(
-  classification: Pick<Classification, "groupAndLevel"> | null | undefined,
-) {
-  if (!classification) return null;
+export function classificationCell(groupAndLevel: string | undefined) {
+  if (!groupAndLevel) return null;
```
petertgiles, line 54: Again, do we want to hide the chip entirely when the field is empty?

petertgiles (changes requested):
Oof, a big PR that makes me a little nervous.  :sweat_smile:   Well, I guess we'll find out how good our e2e tests and AI are.  :laughing:.  A few small things to fix before merging.

apps/web/src/pages/Pools/EditPoolPage/components/SkillTable.tsx
esizer: Fixed in [query missing fields](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17622/commits/53f0ebbb0f822b95136d2d3321ad0050da9f426d)

apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/SkillSummaryTable.tsx
esizer: Oh I see what went wrong :confounded: Fixed in [render fallback instead of returning null](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17622/commits/7d2e99745e46bb288fbee4fa53c7d8106ed60c4d)

apps/web/src/pages/Pools/IndexPoolPage/components/helpers.tsx
esizer: This is a non-nullable field so it should always exist.

https://github.com/GCTC-NTGC/gc-digital-talent/blob/9689627549f9d7d7d644d73b419bafd37db87b68/api/graphql/schema.graphql#L761

petertgiles (approved):

PR #17624, opened 2026-07-31
[Feature] Templates filtering streams with multiselect

### LINKED ISSUE #15138: ✨ Adjust Work Stream menu on the Job Template Search page, or let admins modify it.
### Details

On the Job Advertisement Templates page (under Find a template), on the lefthand side, we have a Work Stream selection menu. 
The menu is very long and half of these work streams are not relevant since we have no job templates for them.

Suggestions from Jamie:
1. Adjust the list to include only the work streams for which we have templates (maybe the platform can pull that data).
2. Create a way for admins to modify that list. We can reduce the menu to only the streams we currently have, and when we create a new template and use a new stream we can add it on our end. 
3. Perhaps a drop down menu would be nice as well.

### Design file & designer

_No response_

### 📸 Screenshots

From Jamie - "While I actually like the design of this filter by workstream, it’s just SO long and not applicable since half these are not represented within the library."

Image

From Jamie - "Perhaps a workstream drop down like on the find talent page would be nice. Under admin access, I can add or remove workstreams from the find talent page, and it could be great to have another option to add or remove workstreams from the job template page."

Image

### 🌎 Localization

_No response_

### Proposed Implementation

_No response_

### ✅ Acceptance Criteria

_No response_

vd1992:
🤖 Resolves #15138

## 👋 Introduction

Switched to a multiselect
Some other changes included, details in code comments and replies below

## 🧪 Testing

1. Navigate to `/en/job-templates`
2. Trying filtering by streams

## 📸 Screenshot

image


apps/web/src/pages/JobPosterTemplates/JobPosterTemplatesPage/JobPosterTemplatesPage.tsx
```diff
@@ -333,7 +339,7 @@ const JobPosterTemplatesPage = () => {
         <Sidebar.Wrapper>
           <Sidebar.Sidebar>
             <FormProvider {...methods}>
-              <div className="flex flex-col gap-6">
+              <div className="flex flex-col gap-6 pb-18">
```
vd1992: Change is a separate commit for easy revert, explanation to follow

vd1992: Without this added space
The added input is a bit squished, once expanded it irked me

image

image

vd1992: _With_ the added space
A little more breathing room

image

image

packages/forms/src/components/Combobox/Multi.tsx
vd1992, line 1: While working on this, noticed a problem where clicking "Reset all filters" would not fully reset the multiselect. The chips internal state wouldn't fully clear, and selected chips persisted.
Claude flagged a problem, I could confirm a replication. Claude comment to follow

vd1992: The multiselect keeps its own copy of the selection in component state, and it only reads the form value once, when the options first load and it needs to resolve any pre-selected values. From that point on its internal copy is the source of truth, and nothing tells it when the form value changes underneath it. "Reset all filters" clears the form value, so the results list updates correctly, but the field never hears about it and keeps its chips on screen. That leaves the page contradicting itself: no filters are applied, yet the input still says three work streams are selected. Nothing had run into this before because everywhere else we reset a form we also unmount it at the same moment, since filter dialogs close and toggle sections collapse, so the component gets thrown away and rebuilt from the form value. This page keeps the form mounted the whole time, which is why it is the first place the gap shows.

The single select half of the same component already handles this, clearing itself whenever the form value empties, so the multiselect now does the same. The only subtlety is which value it watches. What the multiselect receives is the form value already matched against the currently loaded options, and several of our call sites pair it with a server side search that re-queries as the user types, so their selected items can legitimately fall out of that options list mid search. Clearing on that signal would wipe someone's selection while they were still searching for the next one. So the parent passes down a simple flag for whether the form itself still holds a value, which only goes false when the form has genuinely been emptied. A new test covers the reset case, and it fails without this change.

packages/forms/src/components/Combobox/Multi.tsx
```diff
@@ -224,6 +227,14 @@ const Multi = ({
     inputRef?.current?.focus();
   };
 
+  // Clear the selection when the form value is emptied from the outside, such
+  // as a form reset. Mirrors the single select, which does the same thing.
+  useEffect(() => {
+    if (!hasSelection) {
+      reset();
+    }
+  }, [hasSelection, reset]);
```
esizer: I'd see if we can avoid using use effect. Its a pretty big code smell. I'd take a look at [You Might Not Need An Effect](https://react.dev/learn/you-might-not-need-an-effect).

apps/web/src/pages/JobPosterTemplates/JobPosterTemplatesPage/JobPosterTemplatesPage.tsx
```diff
@@ -290,12 +291,17 @@ const JobPosterTemplatesPage = () => {
...
+  // filter changes are detected by comparing the values themselves. The reset
+  // keeps the fields from being marked as having unsaved changes.
+  const previousValues = useRef<FormValues>(defaultValues);
   useDeepCompareEffect(() => {
-    if (methods.formState.isDirty) {
-      handleSubmit();
-      reset(formData);
-    }
-  }, [formData, handleSubmit, methods, reset]);
+    if (isEqual(previousValues.current, formData)) return;
```
esizer: We could just pass the `shouldDirty` to avoid pulling in a dependency and complicating the code.

REF: https://react-hook-form.com/docs/useform/setvalue#:~:text=form%27s%20touched%20fields.-,shouldDirty,-boolean

apps/web/src/pages/JobPosterTemplates/JobPosterTemplatesPage/JobPosterTemplatesPage.tsx
esizer: Personal opinion, this looks like too much space to me 😅 But i'd as a designer.

esizer (changes requested):
Not sure if my comments are something we need to change but just things to look into.

apps/web/src/pages/JobPosterTemplates/JobPosterTemplatesPage/JobPosterTemplatesPage.tsx
vd1992: Fair. 
Will probably need to wait a few days for a response

apps/web/src/pages/JobPosterTemplates/JobPosterTemplatesPage/JobPosterTemplatesPage.tsx
vd1992: Thanks for the link
Done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/17624/changes/037bb0c02707d4b6926d6f9416bf984570e16318

packages/forms/src/components/Combobox/Multi.tsx
vd1992: Thanks for the `useEffect` scrutiny
Done in https://github.com/GCTC-NTGC/gc-digital-talent/pull/17624/changes/a7ab0093e07476fb3ba601c83413893c2adb7b1f

apps/web/src/pages/JobPosterTemplates/JobPosterTemplatesPage/JobPosterTemplatesPage.tsx
vd1992: Josh recommended looking into z-index values instead, the footer should be lower then the menu
Claude suggested the problem was the dropdown z-index was constrained within the sidebar only. So the solution was to try and elevate the sidebar as a whole

Reverted the spacing then adjusted the z-index https://github.com/GCTC-NTGC/gc-digital-talent/pull/17624/commits/34acbb255614dd3abcf568e31bf5281eb72304d3

image

esizer (approved):

PR #17625, opened 2026-07-31
[Refactor] Advanced order by pool candidates paginated

### LINKED ISSUE #16300: ♻️ Implement `@advancedOrderBy` for `poolCandidatesPaginated` queries

esizer:
🤖 Resolves #16300 

## 👋 Introduction

Updates the pool candidates paginated query to use the new advanced order directive.

## 🧪 Testing

1. Refresh and build `make refresh-api; pnpm dev:fresh`
2. Login as a user who can view the candidates table (`admin@test.com` or `community@test.com`)
3. Navigate to `/admin/candidates`
4. Confirm all columsn sort as expected
5. Confirm navigating to a candidate still shows the next/prev linsk that match the tabel order
6. Navigate to a process candidates table
7. Reconfirm both steps 4+5


api/app/Builders/PoolCandidateBuilder.php
```diff
@@ -668,132 +669,105 @@ public function whereScreeningStageIn(?array $screeningStages): self
...
-
-        if (isset($order) && isset($locale)) {
-            return
-            $this->withMax('pool', 'name->'.$locale)
-                ->orderBy('pool_max_name'.$locale, $order)
-                ->orderBy('submitted_at', 'asc');
-        }
-
-        return $this;
+        $citizenship = '(SELECT users.citizenship FROM users WHERE users.id = pool_candidates.user_id)';
```
petertgiles: You might disagree but I think this this additional substitution makes the SQL harder to understand.

api/graphql/schema.graphql
```diff
@@ -1764,49 +1750,20 @@ type Query {
     @find
     @guard
     @canQuery(ability: "view", scopes: ["withPolicyEagerLoads"])
-  poolCandidatesPaginatedAdminView(
+  poolCandidatesPaginated(
```
petertgiles, line 1753: Wow, what a big difference!

api/app/Builders/PoolCandidateBuilder.php
petertgiles: I hope the database is smart enough to run this subquery only once but if you switch the OR to an IN you can probably write the subquery only once.

api/app/GraphQL/Directives/AdvancedOrderByDirective.php
petertgiles, line 1: I asked the AI to do a check for vulnerabilities.  It flagged that we're not sanitizing what comes after the arrow properly, so something like this `email->x' , (SELECT CASE WHEN (1=1) THEN 1 ELSE (SELECT 1/0) END) --` is valid.  This predates your PR but we should probably fix it before we start using the directive.

Suggested fix: add a strict format rule in AdvancedOrderByInputValidator for column and relation.column

petertgiles (changes requested):
This seems to work well!  :clap:  I am worried about a vulnerability before we merge it though.  I could see it being fixed in a separate PR, too, if you wanted to do that.

api/app/GraphQL/Directives/AdvancedOrderByDirective.php
esizer: Hmm, I was not sure how to do this with a vlaidator but if you have auggestions im open to adding that as well. I was able to update the directive to pass them in as paratmeters in postgres which  should avoid the issue you described.

[paramterized json keys in directive](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17625/commits/f7a5822b59a4d327cd3a28d8e525743eb7d605ba)

api/app/Builders/PoolCandidateBuilder.php
esizer: Hopefully I improved this in [update SQL for claim ordering](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17625/commits/48ee8db67370852022018887c0b3cfff20adfce7) :sweat_smile:

petertgiles: Nice, I like that a lot better.  :heart:

api/app/GraphQL/Directives/AdvancedOrderByDirective.php
petertgiles: Nice, this looks much stronger to me and the table seems to still be working well.

The AI suggested that some injection-specific unit tests wouldn't hurt.

It also mentioned that the `nulls` parameter could use the same hardening that the `direction` parameter gets in AdvancedOrder.php.

```php
$nulls = strtoupper($input['nulls'] ?? '');
$this->nulls = in_array($nulls, ['FIRST', 'LAST']) ? $nulls : null;
```

petertgiles (approved):

PR #17637, opened 2026-08-04
[Fix] Candidates table filter count

### LINKED ISSUE #17483: 🐛 UAT | Filter counter is not present on pool candidate table

- When applying any filters in the pool candidate table, the filter counter is not present

**Expected outcome:** The filter counter should be present when filtering pool candidate table

### 🕵️ Details

Env: UAT
Version: 2.77.0

### 📋 Steps to Reproduce

Prerequisite: Create one pool and make sure having 5-10 candidates in it

1. Access to GC digital portal using community admin role
2. Navigate to any pool's candidates table
3. Apply some filter and show results
4. Observe that the filter count is not present on pool candidate table

### 📸 Screenshot

https://github.com/user-attachments/assets/b6da5365-763e-42cf-b8e0-6697bc814745

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Filters count should be present on the pool candidate table to know how many filters have been applied

### 🔄 Regression

Yes

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #17483 

## 👋 Introduction

Fixes an issue with the filter count not appearing on the trigger for the filter dialog on the candidate table.

## 🕵️ Details

We have a magic number that decrements the count by the hard-coded present filters. The issue was, we had this set to `-5` whish is then number of hidden filters, not the number being applied.

so, the filter count started as `-2` meaning you would need to apply 3 filters to get it to appear as `1`.

This updates it to be `-3` to fix this. Ideally, we would not need a magiv number here but that can be sovled if we eventually split these tables up.

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Login as `admin@test.com`
3. Navigate to any process in the admin area
4. Navigate to the candidates tab
5. Apply a number of filters
6. Confirm the value matches the number of filters you changed/applied

### DIFF
--- a/apps/web/src/components/PoolCandidatesTable/PoolCandidateFilterDialog.tsx
+++ b/apps/web/src/components/PoolCandidatesTable/PoolCandidateFilterDialog.tsx
@@ -236,8 +236,9 @@ const PoolCandidateFilterDialog = ({
     <FilterDialog<FormValues>
       options={{ defaultValues: initialValues }}
       // Remove hidden pools filters from count
+      // Default hard-coded filters: expiryStatus, suspendedStatus and poolId
       {...(hidePoolFilter && {
-        modifyFilterCount: -5,
+        modifyFilterCount: -3,
       })}
       {...{ resetValues, onSubmit }}
     >


vd1992 (approved):
Does the job!
Definitely not an ideal approach. It also isn't consistent across filter tables, like this one below at `/en/admin/pool-candidates`
But consistent filter count across all tables is out of scope

Image

PR #17639, opened 2026-08-04
[Refactor] Remove full graphql object types 4

esizer:
Related: #17227 

## 👋 Introduction

Updates pool candidates sections (application snapshot and table) to avoid using full object tyles

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Login as `admin@test.com`
3. Confirm the pool candidates table functions as expected
4. Navigate to the admin candidate view for multiple candidates
5. Confirm the application snapshot renders as expected


apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.stories.tsx
```diff
@@ -3,14 +3,13 @@ import type { StoryFn, Meta } from "@storybook/react-vite";
...
-import type { PoolCandidateAdminViewWithSkillCount } from "@gc-digital-talent/graphql";
 
 import PoolCandidatesTable from "./PoolCandidatesTable";
 
 const poolCandidateData = fakePoolCandidates();
 
-const mockPoolCandidatesWithSkillCount: PoolCandidateAdminViewWithSkillCount[] =
-  poolCandidateData.map((poolCandidate) => {
+const mockPoolCandidatesWithSkillCount = poolCandidateData.map(
+  (poolCandidate) => {
```
vd1992, line 12: Recommend keeping the type hint by importing type
`CandidatesTableCandidatesPaginatedQueryDataType`
from 
`apps/web/src/components/PoolCandidatesTable/PoolCandidatesTable.tsx`
That should be a small change that in turn helps type protect this story. I think it is an important story to merit some extra carefulness but up to you

Quoting

>  The mock goes into parameters.apiResponses, which is untyped — so the PoolCandidateAdminViewWithSkillCount[] annotation was the only thing keeping the mock's row shape in sync with what the table's columns read. After removal, a column that starts reading a new field renders blank in Chromatic instead of failing to compile.

Claude

apps/web/src/components/ApplicationSnapshot/ApplicationSnapshot.tsx
vd1992, line 1: Claude wanted to flag as a finding for files
`apps/web/src/components/ApplicationSnapshot/ApplicationSnapshot.tsx`
`apps/web/src/components/ApplicationSnapshot/Sections/types.ts`
`.../Sections/EducationRequirment/EducationRequirementV1.tsx`

Bringing it up for the sake of completion. The following as quoted: 

> ParsedSnapshot picks up experiences only because EducationRequirementSnapshotV1 happens to declare it — it's the one interface out of the seven in the extends list that does. But experiences is top-level snapshot data, not education-requirement data: ApplicationSnapshot.tsx:99 unpacks it and hands it to CareerTimelineSnapshot and both skills sections, none of which are anywhere in that chain. So a field four sections depend on is now owned by one section's version interface, and nothing in EducationRequirementV1.tsx hints at it.

> Low severity; nothing is broken today and extends User didn't say it any better. It comes due at the next version bump: an EducationRequirementSnapshotV2 that reshapes or drops experiences will either break ApplicationSnapshot plus three unrelated sections, or keep working by accident — and in both cases the person making the change gets no local signal. The fix is one shared interface in Sections/types.ts referenced by both ParsedSnapshot and the education section, so the field has a single definition and the top-level read stops routing through a section.

#### With recommended changes 

```diff
apps/web/src/components/ApplicationSnapshot/Sections/types.ts — add the shared declaration:


 import type { LocalizedString } from "@gc-digital-talent/graphql";
 
+import type { SnapshotExperience } from "~/utils/experienceUtils";
+
+// Top level of the snapshot JSON, read by the education, skills and career
+// timeline sections.
+export interface SnapshotExperiences {
+  experiences?: (SnapshotExperience | null | undefined)[] | null;
+}
+
 export interface SnapshotProps<TSnapshot extends object> {
   snapshot: { version?: number } & Partial<TSnapshot>;
 }
```

```diff
.../Sections/EducationRequirment/EducationRequirementV1.tsx — point at it instead of redeclaring. The alias avoids an empty-interface body, and every existing reference to EducationRequirementSnapshotV1 keeps working:


-import type { SnapshotExperience } from "~/utils/experienceUtils";
-
-import type { SnapshotProps } from "../types";
+import type { SnapshotExperiences, SnapshotProps } from "../types";
 
-export interface EducationRequirementSnapshotV1 {
-  experiences?: (SnapshotExperience | null | undefined)[] | null;
-}
+export type EducationRequirementSnapshotV1 = SnapshotExperiences;
```

>(SnapshotExperience is imported at line 19 solely for that field, so it goes with it.)

```diff
apps/web/src/components/ApplicationSnapshot/ApplicationSnapshot.tsx — declare it directly. Note this is ./Sections/types, not the ./types the file already imports SECTION_KEY from:


+import type { SnapshotExperiences } from "./Sections/types";
 import type { EducationRequirementSnapshotV1 } from "./Sections/EducationRequirment/EducationRequirementV1";
 
 interface ParsedSnapshot
-  extends Partial<PersonalInformationSnapshotV1>,
+  extends SnapshotExperiences,
+    Partial<PersonalInformationSnapshotV1>,
     Partial<LanguageProfileSnapshotV1>,
```

> SnapshotExperiences isn't wrapped in Partial because its field is already optional. Keeping Partial<EducationRequirementSnapshotV1> in the list is redundant right now but costs nothing and stays correct if that interface later grows fields of its own.

apps/web/src/components/PoolCandidatesTable/helpers.tsx
```diff
@@ -107,21 +102,14 @@ export const candidateNameCell = (
 };
 
 export const processCell = (
-  pool: Pick<Pool, "id" | "workStream" | "name" | "publishingGroup"> & {
-    classification?: Pick<Classification, "groupAndLevel"> | null;
-  },
+  poolId: string,
+  poolName: string,
```
vd1992, line 99: Documenting finding for completeness 
Claude didn't like this change as quoted

> Flattening to (poolId, poolName) puts two same-typed positional params next to each other, so processCell(poolName, pool.id, paths, intl) compiles clean and silently renders a link to paths.poolView("IT-03 Software Developer") with the UUID sitting in the aria-label — no type error, no test failure, just a 404.

As passing in arguments in either order throws no alerts. Just one causes problems silently
`processCell(pool.id, poolName, paths, intl)` 
`processCell(poolName, pool.id, paths, intl)` 

Personally I don't think anything needs to change here, we've used adjacent fields with the same type plenty. But I do think going forward for heavily reused functions this is a possible error to guard against. Someone calling a function with arguments in the wrong order.

It recommends the simplest solution as nesting in an object like so 
`processCell({ poolId, poolName }, paths, intl)`

vd1992 (approved):
Alright, gave this a good look through and manually tested some things on pages
- all candidates table
- single pool's candidates table
- view application

Things appear to look great to me!

However, Claude did flag some things of **low** severity. I will include the comments, and leave address or ignore them to your discretion. 
I do want to mention that I think finding 3, the possibility of a user putting in arguments in the wrong order as something to keep in mind. Not in one off function but in reused helpers type guarding against user error is important

That aside, how does this format of review feel?

PR #17642, opened 2026-08-04
[Bug] Fix the flag toggling

### LINKED ISSUE #17636: 🐛 Pool Canddiates Table - Flag icon not coloured on load

When flagging a user on the pool candidates table, the icon initially appears coloured but when reloading the page it is not visible.  Navigating into the candidate shows that it is flagged in the backend - it is only a frontend issue.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

1. Log in as admin@test.com
2. Navigate into a pool, select the candidates tab
3. Click the flag icon and observe that it becomes coloured
4. Reload the page and observe that it is no longer coloured

### 📸 Screenshot

https://github.com/user-attachments/assets/eb37cdb8-5c6a-42f2-9ce7-48379f898564

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Flag icon in the candidates table accurately reflects the state after a reload

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

vd1992:
🤖 Resolves #17636

## 👋 Introduction

Resolve the issue with candidate flags
The incorrect field was being targeted in the query/props

The component is not widely used so it does not need to worry about handling multiple pool candidate slices

## 🧪 Testing

1. Go to a pool's candidates table
2. Exercise toggling

### DIFF
--- a/apps/web/src/components/CandidateFlag/CandidateFlag.tsx
+++ b/apps/web/src/components/CandidateFlag/CandidateFlag.tsx
@@ -11,11 +11,9 @@ import { getFullNameLabel } from "~/utils/nameUtils";
 import useCandidateFlagToggle from "~/hooks/useCandidateFlagToggle";
 
 export const PoolCandidate_FlagFragment = graphql(/* GraphQL */ `
-  fragment PoolCandidate_Flag on PoolCandidate {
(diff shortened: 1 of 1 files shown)


esizer (approved):
LGTM! 👍

PR #17645, opened 2026-08-05
[Feat] Add community interest referral status

### LINKED ISSUE #17590: ✨ Add referral status to community interests
### Details

Expressing interest in a community has the ability to opt a user into appearing in talent requests based on classification. Since users can claim a specific classification in their work experience without an explicit validation of that claim, we need a way for recruiters to review these users.

In order to do so, we are adding some fields to `CommunityInterest` so that commmunity admins and recruiters can update referral status and have more control over how they find the correct talent to match a request based on classification.

### New fields

- `referral_status` (enum): Controls if and how a user is referred in a talent request
    - `NEW` (default): User has expressed interest but classification has not been reviewed
    - `PENDING`: The users classification is currently being reviewed
    - `AVAILABLE_FOR_REFERRAL`: Calssification has been reviewed and saved, allowing them to be referred at the saved classification
    - `NOT_REFERRED`: Classfication was reviewed but could not be confirmed so will not be referred, yet
- `community_classification_id` (foreign UUID to `classifications`): The stored classification to be used in matches 
- `follow_up_date` (date): A date to remind admins to follow up in their pending review or, to check if the stored classification is still valid
- `referral_notes` (string): Additional notes related to the referral review
- `community_referral_status_data_updated_at` (timestamp) - logs timestamp of last change to these new, recruiter-set fields. For now, does not need to be returned 

### Schema

These will be guarded using a `@can*` directive so, we should group these in a `CommunityReferralStatusData` type. Purposed:

```graphql
type CommunityReferralStatusData {
  status: LocalizedCommunityReferralStatus!
  followUpDate: Date
  classification: Classification
  notes: String
  updatedAt: DateTime #(note, can be null)
}
```

### Business rules

Since these new fields control how a user can be referred, we have some rules that go along with them. While these do not explicty forbid a user from being referred, they do change how they show up as a match depending on the referral status.

- `NEW` and `PENDING`: Uses the users current substantive experience classification in a match
- `AVAILABLE_FOR_REFERRAL`: Uses the `community_classification_id` in a match
- `NOT_REFERRED`: Will not appear as a match

### Talent request matches

By default, the only source being matched is `QUALIIFED_IN_POOL`. However, admins can expand the scope to include `AT_LEVEL` (Community employees). This is done using the filter dialog on the talent request matches table.

If a user selects "Community talent" as a source in the filter dialog, a new section for "Community employees" will appear with the referral status filter. This filter should default to the `AVAILABLE_FOR_REFERRAL` status but any status can be selected. No status values is equivalent to selecting all statuses.

### Validation

 - `referral_status`: Required, in enum, not `NEW`(?)
 - `follow_up_date`: Required if not `NOT_REFERRED`, date
 - `community_classification_id`: Required if `AVAIALABLE_FOR_REFERRAL`, uuid, exists in classifications
 - `referral_notes`: Required if `NO_REFERRED`, string

### Business rules

 - If `PENDING`, `null` out `community_classification_id`
 - If `NOT_REFERRED`, `null` out `community_classification_id`, `follow_up_date`

### Permissions

You should be adding a few new permissions:
- `view-any-commuintyInterestReferralStatus` assigned to `platform-admin`
- `view-team-communityInterestReferralStatus` assigned to Community recruiter, Community talent coordinator, and Community admin
- `update-team-communityInterestReferralStatus` assigned to Community recruiter, Community talent coordinator, and Community admin

### Design file & designer

- **Design**: [Figma link](https://www.figma.com/design/uZSzaV8ygi2SA1LwQlBIhh/Community-Talent--Communities-?node-id=1715-7276&p=f&m=dev)
- **Designer**: @Jerryescandon 

### Community talent table

Image

### Edit dialog

Image

### 🌎 Localization

_No response_

### Proposed Implementation

 - Add migration for new fields
 - Likely add new mutation for editing these fields
   - Guard mutation to authenticated, update team community interest permission
   - Validates all input
 - Add edit dialog for updating fields fields
 - Add following columns to community talent table:
   - Community referral status
   - Follow-up date
 - Update the `AT_LEVEL` match logic to consider new fields 
 - Add text coverage for:
   - Updateiung fields
   - Validation logic
   - Match filter logic 

### ✅ Acceptance Criteria

- [ ] New fields added to community interests
- [ ] new mutation to edit CommunityReferralStatusData
  - [ ] inputs validated
  - [ ] guarded by new permissions
  - [ ] auto-updates new `community_referral_status_data_updated_at` field 
- [ ] Community interest table updated with new fields
- [ ] Seeders populate the new fields
- [ ] Edit dialog created
- [ ] At-level match logic considerds new fields as described
- [ ] Talent request matches filters updated with new referral status
- [ ] New permissions added to role-permission seeder and [spreadsheet](https://docs.google.com/spreadsheets/d/1uDIX0Srb4p5Xq03A4BpXpvmnw5yRCZ0__DIdXpMt01k/edit?gid=851474736#gid=851474736)

esizer:
🤖 Resolves #17590 

## 👋 Introduction

Adds a new group of metadata to community intererst that contain information about if and how a user in a community are referred in talent requests.

## 🕵️ Details

Adds the new fields along with some new permissions and mutations to manage them.

 - All community roles can
   - update their own teams
   - view their own teams
 - Platform adming can view any

### Business rules

This modifies how users are matched agianst classifications.

- If status is `NOT_REFERRED` users do not appear in results
- If is `AVAILABLE_FOR_REFERRAL` it should match classificaitons afgianst the referral classificaiton on commuinity interests
- If anything else, it uses the existing substantive experience of the user

## 🧪 Testing

1. Confirm matchiong logic matches rules outline in issue
2. Confirm test coverage exists
3. Confirm permissions crearted and match described
4. Confirm new filter logic exists and functions as expected


api/app/GraphQL/Mutations/UpdateCommunityInterestReferralStatus.php
```diff
@@ -0,0 +1,41 @@
...
+final class UpdateCommunityInterestReferralStatus
+{
+    /**
+     * Handle updating a community interest referral status
+     */
+    public function __invoke($_, array $args)
+    {
+        $communityInterest = CommunityInterest::findOrFail($args['id']);
+        $status = $args['status'] ?? CommunityReferralStatus::NEW->name;
+        $isReferred = $status === CommunityReferralStatus::AVAILABLE_FOR_REFERRAL->name;
```
yonikid15: This variable seems off, shouldn't it be something like `$isAvailableForReferral`?

api/app/GraphQL/Validators/UpdateCommunityInterestReferralStatusInputValidator.php
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Return the validation rules.
+     *
+     * @return array<string, array<mixed>>
+     */
+    public function rules(): array
+    {
+
+        $isNotReferred = $this->arg('status') === CommunityReferralStatus::NOT_REFERRED->name;
+        $isReferred = $this->arg('status') === CommunityReferralStatus::AVAILABLE_FOR_REFERRAL->name;
```
yonikid15: This variable seems off, shouldn't it be something like `$isAvailableForReferral`?

api/app/GraphQL/Validators/UpdateCommunityInterestReferralStatusInputValidator.php
```diff
@@ -0,0 +1,39 @@
...
+
+        return [
+            'id' => ['required', 'uuid'],
+            'status' => ['required', Rule::in(array_column(CommunityReferralStatus::cases(), 'name'))],
+            'followUpDate' => [Rule::requiredUnless($isNotReferred), 'date'],
+            'classification.connect' => [Rule::requiredIf($isReferred), 'uuid', 'exists:classifications,id'],
```
yonikid15: `classification.connect` is only required for `AVAILABLE_FOR_REFERRAL`, but nothing stops it from being applied on other statuses too. Suggest adding `prohibitedIf` so it's rejected outside that status:

```php
'classification.connect' => [
    Rule::requiredIf($isReferred),
    Rule::prohibitedIf(! $isReferred),
    'uuid',
    'exists:classifications,id',
],
```php

File: api/database/factories/CommunityInterestFactory.php [hunk 4]
```diff
@@ -50,6 +53,9 @@ public function definition(): array
             'procurement_is_sdo' => fn ($attributes) => Community::find($attributes['community_id'])->key === 'procurement'
                 ? $this->faker->boolean()
                 : null,
+            'referral_status' => CommunityReferralStatus::NEW->name,
+            'referral_notes' => $this->faker->optional()->paragraph(),
+            'referral_status_data_updated_at' => $this->faker->optional()->dateTimeBetween('-1 month', 'now'),
```
yonikid15: If a default CommunityInterest has a `'referral_status' => CommunityReferralStatus::NEW->name`, then we shouldn't define the fields below.

packages/fake-data/src/fakeCommunityInterests.ts
```diff
@@ -1,27 +1,54 @@
...
+  const referralFollowUp = [
+    CommunityReferralStatus.NotReferred,
+    CommunityReferralStatus.AvailableForReferral,
+  ].includes(referralStatus)
+    ? null
+    : faker.date
+        .between({ from: FAR_PAST_DATE, to: FAR_FUTURE_DATE })
+        .toISOString()
+        .substring(0, 10);
+
```
yonikid15, line 51: This needs fixing. 

- `CommunityReferralStatus.NEW` shouldn't have a follow up date.
- `CommunityReferralStatus.AvailableForReferral` shouldn't be set to null

api/app/Builders/CommunityInterestBuilder.php
```diff
@@ -22,13 +23,30 @@ public function whereMatchesTalentRequest(?array $filters): self
...
+                    // Match the ids as one Postgres array value, so the number of ids has no limit.
+                    $userIdArray = '{'.$userIds->implode(',').'}';
+
+                    $pendingQuery->whereIn('referral_status', [CommunityReferralStatus::NEW->name, CommunityReferralStatus::PENDING->name])
+                        ->whereRaw('community_interests.user_id = any(?::uuid[])', [$userIdArray]);
+                })->orWhere(function (Builder $availableQuery) use ($filters) {
+                    $qualifiedInClassifications = $filters['qualifiedInClassifications'] ?? null;
+
+                    $userIds = $this->atLevelUserIds($filters, atRequestedClassification: false);
+                    $userIdArray = '{'.$userIds->implode(',').'}';
```
yonikid15, line 41: Suggested by Claude (low priority): **Duplicate `atLevelUserIds` lookup**

Branch 2's user set (available-for-referral) already contains everyone in branch 1's set (new/pending) plus more. Branch 1 is just "branch 2's people, narrowed to those at the requested classification." So instead of querying twice, query once for the bigger set, and get the narrower set by filtering that same result in PHP — no second trip to the database.

Before — two nearly identical queries:

```php
$userIds = $this->atLevelUserIds($filters);                              // query #1
...
$userIds = $this->atLevelUserIds($filters, atRequestedClassification: false); // query #2
```

After — one query, then split the results in PHP:

```php
private function atLevelUsers(array $filters): Collection
{
    return User::query()
        ->whereIsVerifiedGovEmployee()
        ->whereUserAttributesMatchTalentRequest($filters)
        ->with('currentClassification')
        ->get(['id']);
}
```

Call it once, use it twice:

```php
$users = $this->atLevelUsers($filters);

$availableIds = $users->pluck('id');                         // for the AVAILABLE_FOR_REFERRAL branch

$newPendingIds = $qualifiedInClassifications
    ? $users->filter(fn ($u) => $this->userMatchesClassification($u, $qualifiedInClassifications))->pluck('id')
    : $availableIds;                                           // for the NEW/PENDING branch
```

So: one database query gets everyone, and the classification check that used to be a second SQL `whereHas` becomes a plain PHP filter over the rows already in memory.

The catch: this only pays off if that base "verified gov employee" set is small-ish. If it's huge, pulling every row into PHP to filter it there could end up slower than just running the second SQL query. Worth a rough gut-check on that table size before treating this as a clear win.

yonikid15 (changes requested):
Looking real good so far. Found some things to fix/clean up.

api/database/factories/CommunityInterestFactory.php
esizer: This seems odd, there is nothing preventing it from being set when it is new. I'm not sure I understand why we should prevent these from being set.

packages/fake-data/src/fakeCommunityInterests.ts
esizer: I don't understand this. Why should'nt new have a follow yup date? and what does it mean to have available for referall set to null?

api/app/Builders/CommunityInterestBuilder.php
esizer: I might need you to re-explain this. I'm not sure what is being said here.

api/app/GraphQL/Mutations/UpdateCommunityInterestReferralStatus.php
esizer: Updated both in [rename variable](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17645/commits/9b5eb56683d1bacaba6155f59add042089939ce8)

api/app/GraphQL/Validators/UpdateCommunityInterestReferralStatusInputValidator.php
esizer: Added in [prohibit classification if not available](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17645/commits/869fc5fd7c42e318e01130fc493bacccab7d2df7)

api/app/Builders/CommunityInterestBuilder.php
yonikid15: You know what, let's ignore this one :laughing:

api/database/factories/CommunityInterestFactory.php
yonikid15: The design shows that you can only add notes while updating the status. Therefore, these two fields would always be null for a NEW status.

image

packages/fake-data/src/fakeCommunityInterests.ts
yonikid15: Oh sorry that wasn't very clear.. According to the design, a `NEW` status would never be able to receive a follow up date.

If the status is `CommunityReferralStatus.AvailableForReferral` then the follow up date shouldn't be null. According to the business logic only the status `NOT_REFERRED` can be null.

api/database/factories/CommunityInterestFactory.php
esizer: Is there anything other than UI preventing this? Is it considered an impossible state if this happens? From how I understood the business rules, the only thing we care about is that things are required, not prohibited.

The only think we prohibit is classification when the status is not available for refer.

yonikid15: The point of the factory is to mimic realistic data. Shouldn't we strive to make this as accurate as possible?

esizer: What im asking is, how is this unrealistic when the business rules allow it? We should not model our data based on designs. There is nothing in the business ruels that forbids this as it stands.

If you are suggesitong this, we might need to revisit the business rules :sweat_smile:

packages/fake-data/src/fakeCommunityInterests.ts
esizer: Changed in [only enforce fake null when not referred](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17645/commits/34dd62415ab19ca4751e65edd96f0f34a0dff0a2)

api/database/factories/CommunityInterestFactory.php
yonikid15: @tristan-orourke You can be the tie breaker

packages/fake-data/src/fakeCommunityInterests.ts
yonikid15: @tristan-orourke Also here, if the status is `NEW` can the fake community interest have a follow up date?

api/database/factories/CommunityInterestFactory.php
tristan-orourke: I think you're both right! Let's add a business rule to the validator that prohibits setting the status to NEW.

esizer: sorry, but the question is not if we can set the status to new. Its whether it makes sense to have notes if the status is new :sweat_smile:

yonikid15 (approved):
Just need to update the snapshot then it's ready to go

PR #17668, opened 2026-08-06
[Fix] Stop freezing the Pusher auth token at construction time (#17181)

### LINKED ISSUE #17181: 🐛 Prod - lots of 500s on /app/gcdt

brindasasi:
🤖 Resolves #17181

## 👋 Introduction

Pusher grabbed the access token from localStorage once, when the client was first constructed. Since the client is only created once at app startup, if the token ever refreshes or expires, every later reconnect keeps sending the old one and 403s on `/graphql/subscriptions/auth` until the page reloads.

Fix: read the token from localStorage fresh on every channel auth request, using pusher-js's `channelAuthorization.headersProvider`, instead of once at startup.

## 🕵️ Details

The issue is about 500s on `/app/gcdt`, but what we found in the code is a 403 loop on `/graphql/subscriptions/auth`, so here's how those connect. 

I pulled a week of real prod logs and found a group of sessions where the 500s and the 403s land in the same tight time clusters, repeating every 60 to 120 seconds, matching Reverb's normal reconnect rhythm. 

That's the signature of one loop: the token goes stale, a reconnect fires, the auth check rejects it with a 403, and that rejection is what's surfacing back to the client as the 500 at the connection endpoint.

The fix removes the trigger we found clustered with real 500s in prod. 
If that holds, we'd expect those clusters to stop within a day or so of deploying, since this group alone was producing hundreds of them a week.

**Confirmed with a forced repro:** 
        Tightening nginx's `proxy_read_timeout` to 5s forces a reconnect every ~9s, and every single reconnect 403'd.

 It never escalated to a literal 500 in that test though, dev is low-traffic and single-session, so whatever specific condition turns the 403 into a visible 500 seems to need prod's real concurrency and infrastructure to show up. 

`api/app/Reverb/ReverbLogger.php` and the `AppServiceProvider.php` binding came from Eric's #17379, merged into this branch during the investigation. 

## 🧪 Testing

Automated:

Run the test createPusher.test.ts against main it should fail and run against the branch it should pass 

Manual:

**To reproduce** 

1. Check out main branch and sign in to the app 
2. Note the token in Local Storage.
4. Get a second valid token
     4.1. Open a second tab (or incognito window) to the same local app and go through sign-in again. 
      4.2 At the mock login screen, type in a different username (applicant@test.com)  copy that tab's access_token value, call it Token B.
5. Back in your original tab (do not reload), overwrite the access_token value in Local Storage with Token B.
6. Force a Pusher reconnect (restart Reverb, or drop network past the activity timeout).
7. Observe the reconnect still sends the old token and gets 403s.

**To test the fix** 
1. Check out the branch 
2. Repeat the same step above and in the 7th step, observe it carries the tokenB and succeeds


esizer (approved):
Cool! I think this should help considerably.

I think the one thing we need to keep an eye on is that there might still be a tiny edge case but im not concerned about it right now.

The case being that wile this new approach gets the most up to date token in locale strage at the time of auth request, it has no capabuilty of refrehsing that token. This improves the likeliness that topken will be valid considerably but, it does not guarantee it will be valid.

I'd say we go ahead with this since it should reduce errors to near zero (I think) but we should keep an eye on it to see if the scenario i outlined is of any concern to us.

PR #17670, opened 2026-08-06
[Feature] Tooltip component

### LINKED ISSUE #17631: ✨ Tooltip component
### Details

A popup that appears when an element is hovered or focused, showing a hint for sighted users.

### Design file & designer

Image

### 📸 Screenshots

_No response_

### 🌎 Localization

_No response_

### Proposed Implementation

Using [Base UI's](https://base-ui.com/react/components/tooltip) unstyled component as a base. Then adding our personal design layer to it.

### ✅ Acceptance Criteria

- [ ] Popup appears when hovering over text or icon
- [ ] Matches design for light and dark modes
- [ ] Add a story

yonikid15:
🤖 Resolves #17631

## 👋 Introduction

Adds a `Tooltip` component built on Base UI's unstyled primitive.

## 🕵️ Details

- New `Tooltip` in `packages/ui`, using the Base UI component library `@base-ui/react/tooltip`.
- Styled for light/dark mode, matching the design.
- 6 Storybook stories: default, opening from top/bottom, on icon, on button, on plain text.

## 🧪 Testing

1. Run Storybook.
2. Open the `Tooltip` stories.
3. Hover/focus each trigger and check both light and dark modes.

### DIFF
--- a/packages/ui/src/components/Tooltip/Tooltip.stories.tsx
+++ b/packages/ui/src/components/Tooltip/Tooltip.stories.tsx
@@ -36,7 +36,7 @@ export default {
 
 const Template: StoryFn<typeof Tooltip.Root> = () => (
   <Tooltip.Provider>
-    <Tooltip.Root>
+    <Tooltip.Root defaultOpen>
       <Tooltip.Trigger aria-label="Administrative Services">
         CS-02
       </Tooltip.Trigger>
--- a/packages/ui/src/components/Tooltip/Tooltip.stories.tsx
+++ b/packages/ui/src/components/Tooltip/Tooltip.stories.tsx
@@ -1,0 +1,114 @@
+import InformationCircleIcon from "@heroicons/react/20/solid/InformationCircleIcon";
+import type { StoryFn, Meta } from "@storybook/react-vite";
+
+import {
+  GLOBAL_A11Y_EXCLUDES,
+  allModes,
+} from "@gc-digital-talent/storybook-helpers";
+
+import Button from "../Button";
+import IconButton from "../Button/IconButton";
+import Tooltip from "./Tooltip";
+
+export default {
+  component: Tooltip.Root,
+  decorators: [
+    (Story) => (
+      <div className="flex justify-center p-16">
+        <Story />
+      </div>
+    ),
+  ],
+  parameters: {
+    chromatic: {
+      modes: {
+        light: allModes.light,
+        dark: allModes.dark,
+      },
+    },
+    a11y: {
+      context: {
+        exclude: [...GLOBAL_A11Y_EXCLUDES],
+      },
+    },
+  },
+} as Meta;
+
+const Template: StoryFn<typeof Tooltip.Root> = () => (
+  <Tooltip.Provider>
+    <Tooltip.Root>
+      <Tooltip.Trigger aria-label="Administrative Services">
+        CS-02
+      </Tooltip.Trigger>
+      <Tooltip.Popup>Administrative Services</Tooltip.Popup>
+    </Tooltip.Root>
+  </Tooltip.Provider>
+);
+
+export const Default = Template.bind({});
+
+export const OpeningFromTop: StoryFn<typeof Tooltip.Root> = () => (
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/Tooltip/Tooltip.stories.tsx
esizer, line 1: These stories ahould ahve the tooltip open like we do for dialogs. Otherwise, our chromatic snapshots lose their value.

packages/ui/src/components/Tooltip/Tooltip.tsx
esizer, line 1: A few things:

- I would love if we could change how the classes are written here for easier readability and maintainability. Something I find helpful is:
  - using an array rather than a single string to group similar styles
  - avoid arbitrary values where possible and use the built in scaling system of tialwind
- Another thing is, the background of the toolstip matches the site backgroundf which makes it really hard to distinguish. I would use the same colour from the card background/text here. Its a little lighter which helps it standout from the background

esizer (changes requested):

packages/ui/src/components/Tooltip/Tooltip.tsx
yonikid15: Hopefully this meets those requirements.. Fixed in [Refactor styling](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17670/commits/0d7205c2d8878eca3854cbb4515b0bb157f8b515)

packages/ui/src/components/Tooltip/Tooltip.stories.tsx
yonikid15: Oh I missed one, fixed in [Default open all stories](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17670/commits/339df7b57c65ddc138af35aa1cafe44f1a1926a1)

packages/ui/src/components/Tooltip/Tooltip.tsx
esizer: sorry, going to nitpick agian. But, why are these assigned to variables? Im nto sure that is necessary. They can just be inlined in the array on separarte lines.

packages/ui/src/components/Tooltip/Tooltip.tsx
```diff
@@ -4,13 +4,28 @@
...
+function ArrowSvg(props: React.ComponentProps<"svg">) {
+  return (
+    <svg width="20" height="10" viewBox="0 0 20 10" fill="none" {...props}>
+      <path
+        d="M9.66437 2.60207L4.80758 6.97318C4.07308 7.63423 3.11989 8 2.13172 8H0V10H20V8H18.5349C17.5468 8 16.5936 7.63423 15.8591 6.97318L11.0023 2.60207C10.622 2.2598 10.0447 2.25979 9.66437 2.60207Z"
+        className="fill-white dark:fill-gray-600"
+      />
+    </svg>
+  );
+}
```
esizer: I tihnk we could jsut inline this rather than creating a component 🤔 I dont see this being reused anywhere. Also, the classname can just be put on the coponent. Im not sure there is much value add to assign it to a variable.

esizer (changes requested):

packages/ui/src/components/Tooltip/Tooltip.tsx
yonikid15: I was just following the pattern we have in the DropdownMenu. Should I change it there too?

image

esizer: Yeah, i have no idea what I was thinking there. :confused: That must have been an off day fopr meh ahah. I can't think of any reason why I would write a component like that :stuck_out_tongue:

yonikid15: Refactored in [Move arrow svg directly into parent](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17670/commits/698c74e951ba7241900eb90eec4cd6968c761c01).

esizer (approved):

PR #17674, opened 2026-08-06
[Feature] Make development programs optional

### LINKED ISSUE #17619: ✨ Make development program optional for Talent Nomination Events
### Details

The form to create a Talent Nomination Event currently gives an error if you try to create the event without a development program. Adding a development program should be optional though.

This will have an impact on what the nomination is for though. "Development program" is currently one of the options, but requires that you select a specific program you want to nominate them for. This will not be possible if there are no development programs associated with the event. So we need to hide that option. FRONT END hiding only is fine as far as Gray is concerned for this.

And one more place... if this is too hard or annoying, we will figure out a work-around. But the description under the H1 for nominations (Nominate talent for advancement, lateral movement, or development opportunities) also references development opportunities. Ideally, we would just say "advancement or lateral movement" if no development programs have been added to the event.

### Design file & designer

_No response_

### 📸 Screenshots

Image

Image

Image

### 🌎 Localization

_No response_

### Proposed Implementation

_No response_

### ✅ Acceptance Criteria

- [ ] Create Talent Nomination Event submits without adding a Development Program
- [ ] "Development opportunities" is not provided as "Nomination option" during the nomination
- [ ] "Nominate talent for advancement, lateral movement, or development opportunities" page subtitle updated to remove "development opportunities"

petertgiles:
🤖 Resolves #17619 

## 👋 Introduction

Makes development programs optional when creating a nomination event and then hides the options to select them when nominating.

## 🧪 Testing

1. Rebuild
2. Log in as admin@test.com

### New Functionality

- [x] Create an upcoming event with no development programs
- [x] Edit that event
- [x] Create a nomination to the event
    - [x] Subtitle is changed
    - [x] No ability to nominate for development programs
- [x] Create an active event with no development programs
- [x] Edit that event

### Regressions

- [x] Create an upcoming event with development programs
- [x] Edit that event
- [x] Create a nomination to the event
    - [x] Subtitle is the original form
    - [x] Can nominate for development programs
- [x] Create an active event with development programs
- [x] Edit that event

## 📸 Screenshot

image

### DIFF
--- a/apps/web/src/pages/TalentNominations/NominateTalent/components/Details.tsx
+++ b/apps/web/src/pages/TalentNominations/NominateTalent/components/Details.tsx
@@ -306,17 +306,23 @@ const DetailsFields = ({
                 "Description for the lateral movement nomination option on the details step",
             }),
           },
-          {
-            value: "developmentProgram",
-            label: intl.formatMessage(adminMessages.developmentOpportunities),
-            contentBelow: intl.formatMessage({
-              defaultMessage:
-                "The employee would benefit from a development or learning opportunity to help prepare them for their next role.",
-              id: "5MFMo5",
-              description:
-                "Description for the development program nomination option on the details step",
-            }),
-          },
+          ...(event?.includeDevelopmentOpportunities
+            ? [
+                {
+                  value: "developmentProgram",
+                  label: intl.formatMessage(
+                    adminMessages.developmentOpportunities,
+                  ),
+                  contentBelow: intl.formatMessage({
+                    defaultMessage:
+                      "The employee would benefit from a development or learning opportunity to help prepare them for their next role.",
+                    id: "5MFMo5",
+                    description:
+                      "Description for the development program nomination option on the details step",
+                  }),
+                },
+              ]
+            : []),
         ]}
       />
       {noOptionsSelected ? (
--- a/apps/web/src/pages/TalentNominations/NominateTalent/components/Details.tsx
+++ b/apps/web/src/pages/TalentNominations/NominateTalent/components/Details.tsx
@@ -107,6 +107,7 @@ const DetailsTalentNominationEvent_Fragment = graphql(/* GraphQL */ `
   fragment DetailsTalentNominationEvent on TalentNominationEvent {
     id
     includeNineBox
+    includeDevelopmentOpportunities
   }
 `);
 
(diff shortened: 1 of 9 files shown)


bella77-69 (commented):
Hi @petertgiles  👋 
Quick question.... the development opportunities notice still appears when “Nominee’s performance assessment” and “Nominee’s leadership potential” are selected. Should this notice still appear in this case?? or should it only display when “Development opportunities” is selected? 

Image

petertgiles:
> Should this notice still appear in this case?

Good question.  I have no idea.  :sweat_smile:

gobyrne:
Good catch @bella77-69 , let's leave it as-is for now. We can clean it up later if needed. I think it still makes sense for the most part.

apps/web/src/pages/TalentNominations/NominateTalent/components/Details.tsx
```diff
@@ -305,17 +306,23 @@ const DetailsFields = ({
...
-            label: intl.formatMessage(adminMessages.developmentOpportunities),
-            contentBelow: intl.formatMessage({
-              defaultMessage:
-                "The employee would benefit from a development or learning opportunity to help prepare them for their next role.",
-              id: "5MFMo5",
-              description:
-                "Description for the development program nomination option on the details step",
-            }),
-          },
+          ...(event?.includeDevelopmentOpportunities
```
tristan-orourke, line 309: Let's remove this condition. The advice remains accurate, even if that development opportunity would happen in a different nomination event.

bella77-69: I think we should keep the condition. If we remove it the Development option will show up for events that weren't created with development opportunities. Users might select it based on the recommended alert description, but since no development opportunities exist for that event the form will throw error.

bella77-69 (approved):
LGTM! 😁

apps/web/src/pages/TalentNominations/NominateTalent/components/Details.tsx
tristan-orourke: Whoops, thanks @bella77-69, I misunderstood where this text appears. The condition is definitely appropriate! 😅

tristan-orourke (approved):

PR #17688, opened 2026-08-07
Potential fixes for 2 code quality findings

mnigh:
_This PR applies 2/2 suggestions from code quality [AI findings](https://github.com/GCTC-NTGC/gc-digital-talent/security/quality/ai-findings)._

### DIFF
--- a/packages/ui/src/components/DropdownMenu/DropdownMenu.tsx
+++ b/packages/ui/src/components/DropdownMenu/DropdownMenu.tsx
@@ -73,7 +73,7 @@ const Popup = ({
 );
 
 const item = tv({
-  base: "duration-200 ease flex cursor-pointer items-center rounded-md bg-transparent px-6 py-1.5 font-bold underline transition outline-none focus-visible:bg-focus hover:focus-visible:text-black",
+  base: "ease flex cursor-pointer items-center rounded-md bg-transparent px-6 py-1.5 font-bold underline transition duration-200 outline-none focus-visible:bg-focus hover:focus-visible:text-black",
   variants: {
     color: {
       primary:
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -73,7 +73,7 @@ const Popup = ({
 );
 
 const item = tv({
-  base: "transition-200 ease flex cursor-pointer items-center rounded-md bg-transparent px-6 py-1.5 font-bold underline transition outline-none focus-visible:bg-focus hover:focus-visible:text-black",
+  base: "duration-200 ease flex cursor-pointer items-center rounded-md bg-transparent px-6 py-1.5 font-bold underline transition outline-none focus-visible:bg-focus hover:focus-visible:text-black",
```
mnigh: > The class `transition-200` is not a valid Tailwind CSS utility. The correct class for a 200ms transition duration is `duration-200`. Additionally, both `transition-200` and `transition` appear in the same class string, which is redundant once corrected to `duration-200`.

packages/ui/src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -41,7 +41,7 @@ const Popup = ({
 }: PopupProps) => (
   <Menu.Portal {...portalProps}>
     <Menu.Positioner
-      className="max-w-(--available-height)"
+      className="max-h-(--available-height)"
```
mnigh, line 44: > The `className` on `Menu.Positioner` uses `max-w-(--available-height)` (max-width), but the CSS variable `--available-height` refers to height, not width. This is likely intended to be `max-h-(--available-height)` (max-height) to constrain the positioner's height, consistent with the `max-h-[var(--available-height)]` already applied to the popup itself.

esizer (approved):
LGTM! 👍

PR #17694, opened 2026-08-10
[Fix] Flaky nomination test

### LINKED ISSUE #17644: 🐛 Flaky E2E test - Create nomination

The Playwright E2E test `admin/talent-nomination.spec.ts` → "Talent nomination management" → "Create a talent nomination" fails occasionally at [talent-nomination.spec.ts:223](apps/playwright/tests/admin/talent-nomination.spec.ts#L223):

```
Error: expect(locator).toBeVisible() failed
Locator: getByRole('heading', { name: /We've received your nomination/i, level: 2 })
Expected: visible
Timeout: 15000ms
Error: element(s) not found
```

**Expected:** After clicking "Submit nomination" on the Review and Submit step, the confirmation heading "We've received your nomination" should render.

**Actual:** The page goes blank (no form step, no confirmation) and stays that way until the assertion times out — even though the submission succeeded on the back end.

**🕵️ Details**

Trace analysis of the failed run shows this is a front-end cache-staleness bug, not a back-end/network failure:

- The `NominateTalentSubmit` GraphQL mutation completes successfully — `200 OK`, no GraphQL errors: `{"data":{"submitTalentNomination":{"id":"...","__typename":"TalentNomination"}}}`.
- No GraphQL requests are made after that mutation completes — the `NominateTalent` query backing the page is never refetched.
- The DOM is byte-identical from immediately after the mutation resolves through to the 15s timeout — the page never updates.
- [`Success.tsx:45`](apps/web/src/pages/TalentNominations/NominateTalent/components/Success.tsx#L45) only renders the confirmation when `talentNomination.submittedAt` is truthy, and that value comes from the separate `NominateTalent` query's cached data, not from the mutation response.
- The submit handler in [`useMutations.ts:120-129`](apps/web/src/pages/TalentNominations/NominateTalent/useMutations.ts#L120-L129) only strips the `step` search param on success — it never explicitly triggers a refetch/invalidation of the `NominateTalent` query.
- urql's default document cache exchange normally auto-invalidates queries sharing a typename with a mutation's response (here `TalentNomination`), which should trigger an automatic refetch — but the network trace confirms this refetch never fires.
- Net effect: once the `step` param is removed (so `current` is `null`) but `submittedAt` is still stale, both the step-form components and the `Success` component render `null`, leaving the main content area blank until the test times out.

> [!NOTE]
> Full Playwright trace available locally for whoever picks this up — ask in the PR/thread if the trace zip is needed.

**📋 Steps to Reproduce**

1. Log in as an applicant with a nomination in progress for an active talent nomination event.
2. Go to `/en/communities/talent-nominations/{id}` and complete the Instructions, Nominator, Nominee, Details, and Rationale steps.
3. On the Review and Submit step, click "Submit nomination".
4. Watch the Network tab — confirm the `NominateTalentSubmit` mutation returns `200` with no errors.
5. Observe that the page content area goes blank and the "We've received your nomination" heading never appears (a manual page refresh does show it, since it forces the query to refetch with fresh data).

**📸 Screenshot**

Image

**🙋‍♀️ Proposed Solution**

Don't rely on implicit urql cache invalidation here. In the `submit` handler ([`useMutations.ts`](apps/web/src/pages/TalentNominations/NominateTalent/useMutations.ts)), explicitly force the `NominateTalent` query to refresh after a successful submit — e.g. via `client.reexecuteQuery` with `requestPolicy: 'network-only'`, or by passing `context: { additionalTypenames: ['TalentNomination'] }` on the mutation call, so the UI reliably shows the confirmation screen without depending on a page reload.

**✅ Acceptance Criteria**

- [ ] After successfully submitting a nomination, the "We've received your nomination" confirmation screen renders reliably without a manual page refresh.
- [ ] `admin/talent-nomination.spec.ts` → "Create a talent nomination" passes consistently in CI (no timeout waiting for the confirmation heading).
- [ ] The `NominateTalent` query's `submittedAt` field updates immediately in the UI after a successful `NominateTalentSubmit` mutation.

**🔄 Regression**

No

**🔗 Root Cause Pull Request**

_No response_

esizer:
🤖 Resolves #17644 

## 👋 Introduction

Fixes an issue where the talent nomination e2e test spec was flaky.

## 🕵️ Details

We were only waiting that the submit request completed. However, the query for the success step was in-flight when we expected the success heading to appear. This adds another wait for the success query before checking for the heading which should make this test more reliable.

## 🧪 Testing

1. Run the test multiple times
2. Confirm it passess consistently

### DIFF
--- a/apps/playwright/tests/admin/talent-nomination.spec.ts
+++ b/apps/playwright/tests/admin/talent-nomination.spec.ts
@@ -215,6 +215,7 @@ test.describe("Talent nomination management", () => {
       .getByRole("button", { name: /submit nomination/i })
       .click();
     await appPage.waitForGraphqlResponse("NominateTalentSubmit");
+    await appPage.waitForGraphqlResponse("NominateTalent");
     await expect(
       appPage.page.getByRole("heading", {
         name: /We’ve received your nomination/i,


vd1992 (approved):
Looks promising 👍 

Image

PR #17724, opened 2026-08-11
[Refactor] Fix markup defects and mistakes

### LINKED ISSUE #17596: ♻️ HTML related defects to fix
## ♻️ Refactor

While fixing the `Ul`/`Ol` trailing-margin bug in #17557 we found three `<p><Ul></p>` call sites and fixed them in passing. A full sweep of the frontend for the same class of defect turns up **22 more sites on `main`**, including two in `packages/ui` that multiply across the app.

Ten of them are the class React treats as an error — an element that can't be a `<p>` descendant — and it logs `In HTML, <ul> cannot be a descendant of <p>. This will cause a hydration error.` for each. Worth being precise about the cost, because `apps/web/react-router.config.ts` sets `ssr: false`: we render as a SPA, React builds these nodes through DOM APIs, and the invalid tree survives verbatim rather than being repaired. The audits below confirm that — `P > P` and `P > UL` are both present in the rendered DOM. So today this costs us invalid markup and console errors; the day any of this markup is server-rendered or prerendered, the HTML parser closes the `<p>` at the first flow-content child and the tree genuinely changes.

The other twelve are content models the browser accepts and React never checks: `<div>` as a direct child of `<ul>`, `<li>` outside any list, headings inside `<span>`. They render fine. What they cost is structural meaning — a list whose children aren't list items stops describing a list of N items to anything reading the accessibility tree.

Nothing catches these today. The full suite (192 files, 547 tests, including the Chromium Storybook project) reports zero nesting errors — and not because these components go unrendered: it runs the `Accordion`, `Stepper`, `DevelopmentProgramCard` and `PoolAdvertisementPage` stories. The twelve content-model defects aren't in React's tables at all, and the Pool page's are inside collapsed Radix accordion content that never mounts until something expands it (surfacing them took clicking through 25 triggers). React also dedupes each warning by `(child, ancestor)` pair for the life of the process, so a page with six of these can log one line.

<details>
<summary>How the sweep was done</summary>

**Static pass.** A TypeScript-AST checker over 1,212 `.tsx` files (1,662 parsed, including `.ts` for the module graph) reproduces React 19.2.7's own `validateDOMNesting` tables — `isTagValidWithParent`, `findInvalidAncestorForTag`, and the ancestor-scope bookkeeping (`pTagInButtonScope`, `aTagInScope`, `buttonTagInScope`, `listItemTagAutoclosing`, `dlItemTagAutoclosing`). A hit there is a warning React would actually emit. A second rule set covers the content models React doesn't check: list children, orphan `<li>`, flow content inside phrasing-only elements, interactive nesting.

**Component resolution.** Most JSX here is components rather than tags, so each one is resolved to its root DOM element through the real module graph: relative imports, the `~/` alias, workspace package index re-exports, compound-component objects (`Notice.Root`, `TableOfContents.Wrapper`), `forwardRef`/`memo` unwrapping, dynamic tags (`const El = level`, `const Heading = as`), Radix `Slot`/`asChild` and context providers treated as transparent, and third-party roots verified against installed source rather than assumed — react-router's `Link` is `createElement("a", …)`, react-hook-form's `FormProvider` emits no DOM. It also models where each component places its `children`, so a caller's JSX is evaluated at the depth it actually lands.

**Coverage, and what it misses.** That resolution work took unresolved component usages from 5,299 down to 2,387 across 599 names. The remainder is mostly Radix compound primitives, which render `<div>`s — those don't clear `<p>` scope, so the `<p>`-descendant class above is unaffected. What's lost is direct-parent checks underneath those wrappers.

**False positives.** A rule only fires when every candidate tag violates it. Two classes were identified and removed: components that wrap children in `<li>` themselves (`Chips`), and JSX held in object literals or variables, which renders somewhere other than where it's written.

**Runtime pass.** 21 of the 22 sites were then mounted — with real fragment data (`getFragment` is identity at runtime) or through composed Storybook stories — and the resulting DOM audited for non-`<li>` list children, `<li>` parents and phrasing-parent containment, with React's `console.error` captured alongside.

**Why the runtime pass was worth it.** It caught the static work out twice. An early probe searched React 18's wording (`cannot appear as a descendant of`) and reported a false clean until corrected to 19's `cannot be a descendant of`. And the `Accordion.tsx` defect surfaced only as an unexplained `SPAN > H3` in a rendered DOM, which exposed a resolver gap — dynamic tags at use sites — that had hidden it from the static pass.

</details>

## 🕵️ Details

Verification: 15 of the 22 sites were observed individually in a rendered DOM; 6 more are covered by a pattern-level observation plus their source line ranges (the four `p > p` sites in `PoolAdvertisementPage` collapse to one `P.my-0 > P` in the audit, the two `span > p` in `DevelopmentProgramCard` to one `SPAN > P`); one is source-only (see note). Grouped by fix, not by rule. Paths in groups 2–5 are shown relative to `apps/web/src/`, and every link is pinned to e501c85 so the line numbers stay right once these files start changing.

### 1. `packages/ui` — fix once, affects everything

| File | Rendered DOM | Rule broken |
|---|---|---|
| [`Accordion.tsx:239`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/packages/ui/src/components/Accordion/Accordion.tsx#L239) | `button` › `span.flex.grow` › `h2`–`h5` in practice (`as` accepts `h1`–`h6` or `p`) | A heading is flow content. Both `<button>` and `<span>` accept phrasing content only, so neither may contain one — the heading is two levels deep inside elements that can't hold it |
| [`Stepper/Step.tsx:94`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/packages/ui/src/components/Stepper/Step.tsx#L94) | `a` › `span.absolute` › `div` | `<span>` accepts phrasing content only; `<div>` is flow content |

`Accordion.Trigger` has 127 call sites (`as="h3"` ×74, `as="h4"` ×23, `as="h5"` ×1, default `h2` ×29) and `Step` renders in every stepper, so both fixes pay off repo-wide. Neither is in React's warning tables, so nothing flags them — which is how they've survived in the shared package.

### 2. Flow content inside a `<p>` — the ten React reports as errors

| File | Rendered DOM | Rule broken |
|---|---|---|
| [`pages/TalentRequests/components/TalentRequestReferralDialogs/ChangeStatusDialog.tsx:81`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/pages/TalentRequests/components/TalentRequestReferralDialogs/ChangeStatusDialog.tsx#L81) | `p.mb-6.flex-col` › `span` › `IconLabel` = `p.flex` | `IconLabel` is rooted at `<p>`, and a `<p>` cannot contain another |
| [`pages/TalentRequests/components/TalentRequestTrackedUsersInbox/dialogs/StatusDialogLayout.tsx:60`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/pages/TalentRequests/components/TalentRequestTrackedUsersInbox/dialogs/StatusDialogLayout.tsx#L60) | same shape as above | same — the two dialogs share the markup |
| [`pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx:1082`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx#L1082)<br>also [1091](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx#L1091), [1170](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx#L1170), [1179](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx#L1179)<br>outer wrappers: [1068](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx#L1068), [1156](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx#L1156) | `Text` = `p.my-0` › `Text` = `p.my-3` | the local `Text` component is a `<p>`, nested inside itself |
| [`pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx:1254`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx#L1254) | `Text` = `p.my-3` › `Ul` = `ul` | `<ul>` is flow content, so it can't sit inside a `<p>` |
| [`components/RecruitmentProcesses/RecruitmentProcessesPreviewList.tsx:182`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/components/RecruitmentProcesses/RecruitmentProcessesPreviewList.tsx#L182) | `p.mb-6` › `OffPlatformRecruitmentProcessList` = `ul.mb-6.flex-col` | same — a component rooted at `<ul>` inside a `<p>` |
| [`pages/TalentNominations/NominationGroup/components/CurrentPositionExperiences.tsx:147`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/pages/TalentNominations/NominationGroup/components/CurrentPositionExperiences.tsx#L147)<br>also [156](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/pages/TalentNominations/NominationGroup/components/CurrentPositionExperiences.tsx#L156) | `p` (139→176) › `p.my-3` and `Ul` = `ul` | two violations in one paragraph: a nested `<p>` and, alongside it, a `<ul>` |

These six rows are the ten sites React warns about, and the same defect #17557 fixed in three other files. Each renders as authored today (`ssr: false`), so the fix is about valid markup and a clean console rather than a visible bug — with the caveat from the top of the ticket about what happens if this markup ever reaches an HTML parser.

### 3. Broken list structure

| File | Rendered DOM | Rule broken |
|---|---|---|
| [`components/NextRoleAndCareerObjective/NextRoleAndCareerObjective.tsx:77`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/components/NextRoleAndCareerObjective/NextRoleAndCareerObjective.tsx#L77)<br>also [96](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/components/NextRoleAndCareerObjective/NextRoleAndCareerObjective.tsx#L96) | `PreviewList.Root` = `ul` › `Notice.Root` = `div`, `div` | `<ul>` accepts only `<li>` and script-supporting children. The populated branch of this same list contributes an `<li>` via `PreviewList.Item`, so the list mixes both |
| [`pages/Users/UserEmployeeInformationPage/components/NextRoleAndCareerObjective.tsx:72`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/pages/Users/UserEmployeeInformationPage/components/NextRoleAndCareerObjective.tsx#L72)<br>also [95](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/pages/Users/UserEmployeeInformationPage/components/NextRoleAndCareerObjective.tsx#L95) | `div.space-y-6` › `PreviewList.Item` = `li`, `li` | an `<li>` may only be a child of `<ul>`, `<ol>` or `<menu>` — here there's no list at all |
| [`pages/Pools/AssessmentPlanBuilderPage/components/AssessmentStepCard.tsx:144`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/pages/Pools/AssessmentPlanBuilderPage/components/AssessmentStepCard.tsx#L144) | `ul` › `li`, `span.mx-3`, `li`, `span.mx-3`, `li` | `<ul>` accepts only `<li>`; the `aria-hidden` bullet separators are siblings of the items rather than inside them |

The first two files are near-duplicates that made complementary mistakes — one puts non-`li` content in the list, the other puts `li` content outside one. None of it is in React's tables, so nothing warns. What's wrong is structural: a `<ul>` holding two `<div>`s describes a list with no items, and an `<li>` whose parent is a `<div>` belongs to no list at all.

### 4. `<span>` used as a flex container around block content

| File | Rendered DOM | Rule broken |
|---|---|---|
| [`components/DevelopmentProgramCard/DevelopmentProgramCard.tsx:80`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx#L80)<br>also [87](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx#L87) | `span` › `Heading` = `h3` (the `headingAs` default), and `p` | `<span>` accepts phrasing content only; headings and `<p>` are flow content |
| [`components/DevelopmentProgramCard/DevelopmentProgramCard.tsx:92`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx#L92)<br>also [101](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx#L101) | `span.flex.gap-1.5` › `p`, `Chips` = `ul` | same — and this one is explicitly a flex row, so the `<span>` was never doing a phrasing job |
| [`components/ScreeningDecisions/ScreeningDecisionDialogForm.tsx:38`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialogForm.tsx#L38) | `span.mb-3.flex` › `p` | same, inside `ContextBlock` |

All three are layout containers that happen to be `<span>`s. Browsers render them as written, so nothing looks wrong today.

> [!NOTE]
> `ScreeningDecisionDialogForm.tsx:38` is the one site not confirmed against a rendered DOM — its branch needs mocked GraphQL (`useOptions` leaves the form in its pending state otherwise). The markup is intrinsic-only inside a 13-line component (lines 30–42), so the JSX maps 1:1 to the DOM, and it is live code (rendered through `CardOptionGroup`'s `context` slot at line 176).

### 5. Adjacent, cheap to fix in the same files

Not nesting defects, but they're in files the groups above already touch.

| File | Problem |
|---|---|
| [`pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx:1204`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx#L1204)<br>also [1236](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx#L1236) | `className="m-y0"` is not a utility defined anywhere in the project, so the intended `my-0` never applies and `Text`'s default `my-3` stands |
| [`components/ScreeningDecisions/ScreeningDecisionDialogForm.tsx:30`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/e501c85092574a0f6e45f14a6ee4c2c46b796623/apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialogForm.tsx#L30) | `ContextBlock` destructures `key` out of props, but React strips `key` before props reach a component — it is always `undefined`, so the map keys come out as `"undefined-1"`, `"undefined-2"` |

## 🙋‍♀️ Proposed Solution

**Group 1 — `Accordion.tsx`.** Not a `span` → `div` swap: the element is inside the Radix `<button>`, which permits only phrasing content, so a `<div>` or a heading is invalid there either way. Move the rank element out of the button and wrap the trigger with it — the ARIA APG accordion pattern, and what `AccordionPrimitive.Header` renders by default:

```tsx
<AccordionPrimitive.Header asChild>
  <Hx className={header()} {...titleProps}>       {/* Hx = the `as` prop, h2 by default */}
    <AccordionPrimitive.Trigger className={btn()} {...rest}>
      …
      <span className="flex grow flex-col">
        <span className={heading()}>{children}</span>   {/* was <Hx> */}
        {subtitle && <span …>{subtitle}</span>}
      </span>
    </AccordionPrimitive.Trigger>
    …
  </Hx>
</AccordionPrimitive.Header>
```

Two things to decide while doing it: whether the icon/context spans currently rendered as siblings of the trigger should stay inside the heading element (they would become part of the heading's text content for screen readers — scoping the heading to just the trigger avoids that), and Chromatic review, since this changes the DOM of all 127 accordions. `Stepper/Step.tsx:94` is a plain `span` → `div`; it sits inside `<a>` inside `<li>`, where flow content is allowed.

**Group 2.** Same fix as #17557: drop the wrapper that shouldn't be a paragraph.
- `ChangeStatusDialog` / `StatusDialogLayout`: the outer element is a flex column holding two spans, not a paragraph — make it a `<div>` and leave `IconLabel` alone.
- `PoolAdvertisementPage`: split each `<Text className="my-0">msg<Text/><Text/></Text>` into sibling `Text`s, and drop the `Text` wrapper around `<Ul className="mt-3">` at `1253`.
- `RecruitmentProcessesPreviewList`: render `OffPlatformRecruitmentProcessList` directly (it already carries `mb-6`) and keep the `<p>` for the null-state string only.
- `CurrentPositionExperiences`: split the outer `<p>` into `<p>`, `<p class="my-3">`, `<Ul>` siblings.

**Group 3.**
- `NextRoleAndCareerObjective` (component): render the null-state notices as siblings of `PreviewList.Root` rather than inside it, so the `<ul>` only ever contains `PreviewList.Item`s.
- `UserEmployeeInformationPage` variant: wrap the two previews in `PreviewList.Root`, matching the other file. These two files are near-duplicates with complementary bugs — worth aligning rather than patching separately.
- `AssessmentStepCard`: move the separator inside the `<li>` or produce it with CSS (`li + li::before`) instead of a sibling `<span>`.

**Group 4.** `<span>` → `<div>` at all five sites. Both landing spots are safe: the `DevelopmentProgramCard` spans sit inside `li > div > div`, and `ContextBlock` renders into `Field.Context` → `Notice.Root`/`Notice.Content`, both `<div>`s.

> [!NOTE]
> Worth a follow-up ticket to run the checker described above in CI, so these classes can't come back silently.

## ✅ Acceptance Criteria

- [ ] None of the components listed in Details log `In HTML, <x> cannot be a descendant of <y>` when rendered
- [ ] Every `<ul>`/`<ol>` listed contains only `<li>` children, and no `<li>` renders outside a list
- [ ] No `<span>` listed contains block-level content, and no heading or `<div>` renders inside an accordion trigger `<button>`
- [ ] Accordion triggers keep their current heading rank per call site (`as="h3"` ×74, `as="h4"` ×23, `as="h5"` ×1, default `h2` ×29) and their accessible name does not gain the context/icon text
- [ ] Chromatic shows no unintended visual change to the `Accordion`, `Stepper`, `DevelopmentProgramCard` and `PoolAdvertisementPage` stories
- [ ] The surfaces with no story — both talent request status dialogs, `CurrentPositionExperiences`, `RecruitmentProcessesPreviewList`, both `NextRoleAndCareerObjective` variants and `AssessmentStepCard` — are verified by hand or gain a story
- [ ] Every finding listed in Details is either fixed or dismissed as a false positive, with the reasoning recorded on this issue

---
<sub>Findings compiled with Claude Opus 5 (1M context).</sub>

vd1992:
🤖 Resolves #17596

## 👋 Introduction

To attempt to play to the bot's strength I had it try and analyze to find markup problems. Something easier to do with such tooling rather than stumbling upon with the console open by chance

## 🕵️ Details

It did some extensive probing, with me pestering for double-checks and refinements. This resulted in the connected issue being fairly extensive and involved. Gathering the findings, proposing fixes, and plenty of justification or rationale. Making for a harder to parse issue.

So why not let the bot try to handle what it found and see if it makes sense?
Judging by the diff size, it wasn't all that much editing required

### Explanation of the fixes

Nesting elements inside other elements that you aren't supposed to.
Component key that was ending up undefined
Accordion header markup fix, was technically invalid but threw no problems due to implementation's leniency
Tailwind classes that were invalid (likely typos)

> **Invalid HTML nesting — the class React reports as errors (9 sites, 5 files)**
>
> Paragraphs were wrapping content that can't legally live inside a paragraph: bulleted lists, other paragraphs, and components rooted at a list. Browsers tolerate this today only because the app renders client-side; if any of it is ever server-rendered or prerendered, the parser closes the paragraph early and the structure genuinely changes.
> Two status dialogs had a second layer of the same fault that the ticket's proposed fix would have missed — an inner span also holding a paragraph. Re-tagging only the outer element would have moved the problem rather than solved it.
>
>**Accordion heading nested inside the trigger button (affects every accordion in the app)**
>
>The heading element sat inside the `<button>`. A heading is flow content, which is invalid inside a button, and ARIA marks a button's children as presentational — so a conforming assistive technology is permitted to drop the heading entirely. In practice the accessibility tree on main does expose it at the correct level, so there's no observed loss today; the markup was relying on lenient implementation rather than on the spec. The heading now wraps the button, matching the ARIA APG accordion pattern and Radix's own Accordion.Header default. The rendered DOM is otherwise byte-identical, with no class gained or lost.
>
>
>**Lists that weren't lists (5 sites, 3 files)**
>
>Two near-duplicate components had opposite faults: one put non-list content directly inside a `<ul>`, the other rendered list items with no list around them at all. The second had a visible consequence — Tailwind's preflight resets list-style on ul/ol but not on li, and list-style-type is inherited, so `<li>`s adrift in a `<div>` fall back to disc and rendered stray bullets.
>A third placed decorative bullet separators as siblings of the list items rather than inside them, so the `<ul>` had five children where only three were items.
>
> **`<span>` used as a layout container (5 sites, 3 files)**
>
> Five spans were holding headings, paragraphs and a list, where only inline content is allowed. All five were already flex containers or flex items, so the swap to div changes nothing visually.
>
> **Styling that never did anything (4 sites)**
>
> Four Tailwind classes were misspelled in a way that matches no utility (m-y0, m-t18, m-t0, m-y3), so they had never applied. Deleted rather than corrected — these pages passed design review as they actually render, so making the classes work would have silently changed approved layouts.
>
> **A prop React silently discards (1 site)**
>
> A component took its list-key prefix in a prop named key. React strips key before props reach a component, so it always arrived undefined and every generated key came out as "undefined-1", "undefined-2". Renamed to idPrefix.

~ Claude 

## 🧪 Testing

1. Manual exercise of pages
2. Observe Chromatic diff

Places where I manually checked include:

- PoolAdvertisementPage 
- Stepper / Step
- UserEmployeeInformationPage → NextRoleAndCareerObjective 
- Accordion with a subtitle
- CommunityDashboardPage
- AssessmentStepCard
- NextRoleAndCareerObjective — nomination group → Profile tab
- RecruitmentProcessesPreviewList
- CurrentPositionExperiences
- ChangeStatusDialog — talent request → Details tab
- StatusDialogLayout — talent request → Tracking tab

## 📸 Screenshot

One place with changes, but I'd contend that it would be fixing existing problems

#### MAIN

image

#### BRANCH

image

#### RELATED
I would contend this resolved a visual bug. As it now is in sync with

image

### DIFF
--- a/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx
+++ b/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx
@@ -1176,7 +1176,7 @@ export const PoolPoster = ({
                     })}
                   </Accordion.Trigger>
                   <Accordion.Content>
-                    <Text className="m-y0">
+                    <Text>
                       {intl.formatMessage(
                         {
                           defaultMessage:
--- a/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx
+++ b/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx
@@ -933,7 +933,7 @@ export const PoolPoster = ({
                     size="h3"
                     icon={QuestionMarkCircleIcon}
                     color="success"
-                    className="m-t18 mb-0"
+                    className="mb-0"
                   >
                     {sections.moreInfo.title}
                   </TableOfContents.Heading>
--- a/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx
+++ b/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx
@@ -1056,24 +1056,24 @@ export const PoolPoster = ({
                             internalLink(paths.register(), chunks),
                         },
                       )}
-                      <Text>
-                        {intl.formatMessage({
-                          defaultMessage:
-                            "First-time users report spending between 2 to 3 hours on average on their first application. The second time you apply, it should take significantly less time since we'll reuse most of your information. Return users spend on average between 30 minutes to 1 hour on each application.",
-                          id: "adaiV0",
-                          description:
-                            "Text explaining the average time spent on applications",
-                        })}
-                      </Text>
-                      <Text>
-                        {intl.formatMessage({
-                          defaultMessage:
-                            "Applying early also allows hiring managers to review your application sooner. They can begin assessing your application as soon as you submit it, even if the deadline hasn't passed yet.",
-                          id: "ay8mcT",
-                          description:
-                            "Text explaining the benefits of applying early",
-                        })}
-                      </Text>
+                    </Text>
+                    <Text>
+                      {intl.formatMessage({
+                        defaultMessage:
+                          "First-time users report spending between 2 to 3 hours on average on their first application. The second time you apply, it should take significantly less time since we'll reuse most of your information. Return users spend on average between 30 minutes to 1 hour on each application.",
+                        id: "adaiV0",
+                        description:
+                          "Text explaining the average time spent on applications",
+                      })}
+                    </Text>
+                    <Text>
+                      {intl.formatMessage({
+                        defaultMessage:
+                          "Applying early also allows hiring managers to review your application sooner. They can begin assessing your application as soon as you submit it, even if the deadline hasn't passed yet.",
+                        id: "ay8mcT",
+                        description:
+                          "Text explaining the benefits of applying early",
+                      })}
                     </Text>
                   </Accordion.Content>
                 </Accordion.Item>
@@ -1144,24 +1144,24 @@ export const PoolPoster = ({
                             internalLink(paths.support(), chunks),
                         },
                       )}
-                      <Text>
-                        {intl.formatMessage({
-                          defaultMessage:
-                            "We review tickets during business hours. If you report an issue late at night, right before a job advertisement deadline, expect a response within 2 business days. To avoid last-minute issues, we encourage you to submit your application as early as possible.",
-                          id: "b9gGFH",
-                          description:
-                            "Text explaining the average time spent on applications",
-                        })}
-                      </Text>
-                      <Text>
-                        {intl.formatMessage({
-                          defaultMessage:
-                            "If you submit your ticket after the application deadline has passed, we won't be able to assist you, and your application won't be accepted.",
-                          id: "4DfNUW",
-                          description:
-                            "Text explaining the benefits of applying early",
-                        })}
-                      </Text>
+                    </Text>
+                    <Text>
+                      {intl.formatMessage({
+                        defaultMessage:
+                          "We review tickets during business hours. If you report an issue late at night, right before a job advertisement deadline, expect a response within 2 business days. To avoid last-minute issues, we encourage you to submit your application as early as possible.",
+                        id: "b9gGFH",
+                        description:
+                          "Text explaining the average time spent on applications",
+                      })}
+                    </Text>
+                    <Text>
+                      {intl.formatMessage({
+                        defaultMessage:
+                          "If you submit your ticket after the application deadline has passed, we won't be able to assist you, and your application won't be accepted.",
+                        id: "4DfNUW",
+                        description:
+                          "Text explaining the benefits of applying early",
+                      })}
                     </Text>
                   </Accordion.Content>
                 </Accordion.Item>
--- a/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx
+++ b/apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx
@@ -1176,7 +1176,7 @@ export const PoolPoster = ({
                     })}
                   </Accordion.Trigger>
                   <Accordion.Content>
-                    <Text>
+                    <Text className="my-0">
                       {intl.formatMessage(
                         {
                           defaultMessage:
--- a/apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
+++ b/apps/web/src/components/DevelopmentProgramCard/DevelopmentProgramCard.tsx
@@ -76,7 +76,7 @@ const DevelopmentProgramCard = ({
           </DropdownMenu.Root>
         )}
         <div className="flex flex-col items-start gap-3">
-          <span>
+          <div>
             <Heading
               level={headingAs}
               size="h6"
@@ -85,10 +85,10 @@ const DevelopmentProgramCard = ({
               {title}
             </Heading>
             {description && <p>{description}</p>}
-          </span>
+          </div>
 
           {classificationRestrictions?.length ? (
-            <span className="flex gap-1.5">
+            <div className="flex gap-1.5">
               <p>
                 {intl.formatMessage({
                   defaultMessage: "Restricted to",
@@ -105,7 +105,7 @@ const DevelopmentProgramCard = ({
                   </Chip>
                 ))}
               </Chips>
-            </span>
+            </div>
           ) : null}
         </div>
       </div>
--- a/apps/web/src/components/NextRoleAndCareerObjective/NextRoleAndCareerObjective.tsx
+++ b/apps/web/src/components/NextRoleAndCareerObjective/NextRoleAndCareerObjective.tsx
@@ -73,38 +73,42 @@ const NextRoleAndCareerObjective = ({
           <PreviewList.Root>
             {/* If the employee hasn't filled out this section then display null message */}
             {hasAllEmptyFieldsNextRole({ ...nextRole }) ? (
-              <Notice.Root className="my-6">
-                <Notice.Content>
-                  <p>
-                    {intl.formatMessage({
-                      defaultMessage:
-                        "The nominee has not provided information about the next role they’d like to achieve.",
-                      id: "wYka/g",
-                      description:
-                        "Message displayed if nominee hasn't filled out next role info",
-                    })}
-                  </p>
-                </Notice.Content>
-              </Notice.Root>
+              <li>
+                <Notice.Root className="my-6">
+                  <Notice.Content>
+                    <p>
+                      {intl.formatMessage({
+                        defaultMessage:
+                          "The nominee has not provided information about the next role they’d like to achieve.",
+                        id: "wYka/g",
+                        description:
+                          "Message displayed if nominee hasn't filled out next role info",
+                      })}
+                    </p>
+                  </Notice.Content>
+                </Notice.Root>
+              </li>
             ) : (
               <NextRolePreview
                 nextRolePreviewQuery={nextRoleAndCareerObjective}
               />
             )}
             {hasAllEmptyFieldsCareerObjective({ ...careerObjective }) ? (
-              <Notice.Root className="my-6">
-                <Notice.Content>
-                  <p>
-                    {intl.formatMessage({
-                      defaultMessage:
-                        "The nominee has not provided information about their ultimate career objective.",
-                      id: "wL95nl",
-                      description:
-                        "Message displayed if nominee hasn't filled out career objective info",
-                    })}
-                  </p>
-                </Notice.Content>
-              </Notice.Root>
+              <li>
+                <Notice.Root className="my-6">
+                  <Notice.Content>
+                    <p>
+                      {intl.formatMessage({
+                        defaultMessage:
+                          "The nominee has not provided information about their ultimate career objective.",
+                        id: "wL95nl",
+                        description:
+                          "Message displayed if nominee hasn't filled out career objective info",
+                      })}
+                    </p>
+                  </Notice.Content>
+                </Notice.Root>
+              </li>
             ) : (
               <CareerObjectivePreview
                 careerObjectivePreviewQuery={nextRoleAndCareerObjective}
--- a/apps/web/src/components/RecruitmentProcesses/RecruitmentProcessesPreviewList.tsx
+++ b/apps/web/src/components/RecruitmentProcesses/RecruitmentProcessesPreviewList.tsx
@@ -177,20 +177,20 @@ const RecruitmentProcessPreviewList = ({
             description: "Off-platform section information",
           })}
         </p>
-        <p className="mb-6">
-          {offPlatformRecruitmentProcesses?.length ? (
-            <OffPlatformRecruitmentProcessList
-              processesQuery={offPlatformRecruitmentProcesses}
-            />
-          ) : (
-            intl.formatMessage({
+        {offPlatformRecruitmentProcesses?.length ? (
+          <OffPlatformRecruitmentProcessList
+            processesQuery={offPlatformRecruitmentProcesses}
+          />
+        ) : (
+          <p className="mb-6">
+            {intl.formatMessage({
               defaultMessage:
                 "The nominee has not added any off-platform processes.",
               id: "4NtC0R",
               description: "Null state for off-platform section",
-            })
-          )}
-        </p>
+            })}
+          </p>
+        )}
       </div>
     </>
   );
--- a/apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialogForm.tsx
+++ b/apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialogForm.tsx
@@ -24,14 +24,16 @@ const TEXT_AREA_MAX_WORDS = 200;
 
 interface ContextBlockProps {
   messages: string[];
-  key: string;
+  idPrefix: string;
 }
 
-const ContextBlock = ({ messages, key }: ContextBlockProps) => (
+// NOTE: not named `key` — React strips that before props reach a component,
+// so it would always arrive undefined.
+const ContextBlock = ({ messages, idPrefix }: ContextBlockProps) => (
   <div className="mb-3">
     {messages.map((message, index) => (
       <div
-        key={`${key}-${index + 1}`}
+        key={`${idPrefix}-${index + 1}`}
         className="mb-3 flex justify-start gap-3"
       >
         <CheckIcon className="size-6" />
@@ -175,7 +177,7 @@ const ScreeningDecisionDialogForm = ({
                     educationContext ? (
                       <ContextBlock
                         messages={educationContext.messages}
-                        key={educationContext.key}
+                        idPrefix={educationContext.key}
                       />
                     ) : null
                   }
--- a/apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialogForm.tsx
+++ b/apps/web/src/components/ScreeningDecisions/ScreeningDecisionDialogForm.tsx
@@ -30,13 +30,13 @@ interface ContextBlockProps {
 const ContextBlock = ({ messages, key }: ContextBlockProps) => (
   <div className="mb-3">
     {messages.map((message, index) => (
-      <span
+      <div
         key={`${key}-${index + 1}`}
         className="mb-3 flex justify-start gap-3"
       >
         <CheckIcon className="size-6" />
         <p>{message}</p>
-      </span>
+      </div>
     ))}
   </div>
 );
--- a/apps/web/src/pages/CommunityDashboardPage/CommunityDashboardPage.tsx
+++ b/apps/web/src/pages/CommunityDashboardPage/CommunityDashboardPage.tsx
@@ -332,12 +332,7 @@ export const DashboardPage = ({ currentUser }: DashboardPageProps) => {
           )}
         </div>
         <div className="mt-18 mb-12">
-          <Heading
-            size="h4"
-            icon={CogIcon}
-            color="warning"
-            className="m-t0 mb-6"
-          >
+          <Heading size="h4" icon={CogIcon} color="warning" className="mb-6">
             {intl.formatMessage({
               defaultMessage: "Your roles",
               id: "IJlJF1",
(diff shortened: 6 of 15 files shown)


vd1992:
PR diff size is inflated by indentation changes. Which makes up approximately two-thirds of it apparently

>~67% of the diff is re-indentation from JSX changing depth — use the hide-whitespace view; the real change is +89/−53.

apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx
```diff
@@ -1176,7 +1176,7 @@ export const PoolPoster = ({
                     })}
                   </Accordion.Trigger>
                   <Accordion.Content>
-                    <Text className="m-y0">
+                    <Text>
```
yonikid15: We needed that there

```suggestion
                    <Text className="my-0">
```

tristan-orourke:
> PR diff size is inflated by indentation changes

I don't know about anyone else, but I usually look at the number of files affected to gauge a PR's complexity, not the number of lines touched. This PR still has 30+ real change points across 15 files.

yonikid15 (changes requested):
Nice work 🎉 lots of small but important fixes

Just need to add back that class that got removed (but spelt correctly). It's needed to keep the margin consistent with the other sections.

esizer:
> > PR diff size is inflated by indentation changes
> 
> I don't know about anyone else, but I usually look at the number of files affected to gauge a PR's complexity, not the number of lines touched. This PR still has 30+ real change points across 15 files.

I am close but slightly different, I look at a combination of:

 - Lines of code
 - Files modified
 - Area of impact (places I need to review)
 - Total scope
 
Some scenarios of a PR that could:

 - Touch 100 files but be fixing something as minor as a package name or variable typo and still be very easy to quickly review
 - Add 1000 lines of code but it could be a new class with lots of PHPDoc, be used in one specific place and just be a lot unit tests
 - Be a 2 line fix to a function that is used on 50 different pages I need to check
 - A PR could only touch 5-10 files and be 30 line diff but to 10 very different features that are unrelated and take time to individually review with a lot of context switching

apps/web/src/pages/Pools/PoolAdvertisementPage/PoolAdvertisementPage.tsx
vd1992: Good catch! Thanks!

yonikid15 (approved):
Excellent!

vd1992:
@esizer 
Do you have any concerns about this PR? Requests to split any parts out?
You did raise confusion in the issue body. I just went ahead with a PR initially to test how complicated it would turn out. It personally to me was less massive than I first thought. 

Any objections to merging as is?

esizer:
> @esizer Do you have any concerns about this PR? Requests to split any parts out? You did raise confusion in the issue body. I just went ahead with a PR initially to test how complicated it would turn out. It personally to me was less massive than I first thought.
> 
> Any objections to merging as is?

No direct concerns, no. I just had a very diffiuclt time parsing the ticket itself so I had no idea how to evalutate the PR and determine what issues it was attempting to solve to determin if the fixes applied address the issues or if they were the most appropriate.

But, I trust your work and @yonikid15 review :stuck_out_tongue:

vd1992:
> > @esizer Do you have any concerns about this PR? Requests to split any parts out? You did raise confusion in the issue body. I just went ahead with a PR initially to test how complicated it would turn out. It personally to me was less massive than I first thought.
> > Any objections to merging as is?
> 
> No direct concerns, no. I just had a very diffiuclt time parsing the ticket itself so I had no idea how to evalutate the PR and determine what issues it was attempting to solve to determin if the fixes applied address the issues or if they were the most appropriate.
> 
> But, I trust your work and @yonikid15 review 😛

Thank you!
Yeah, the issue I directed to be in-depth and technical. Which works for having rationale well-explained.
The downside I failed to account for... it makes the issue a brick wall 😵

PR #17725, opened 2026-08-12
[Fix] Allow unauthenticated job poster template download

### LINKED ISSUE #17356: 🐛 Error message when trying to download job templates

Issue: I was not able to download a job template. 
What should be happening: I think I should be able to download it signed in or signed out since there is a button there?

### 🕵️ Details

I went to download a job template so I could review and save it to share and was receiving this error when trying to download. 
- I was not signed in when trying to download it
- I tried on both chrome and edge

After logging in, I could only download after verifying my gov status and becoming part of a community. 

### 📋 Steps to Reproduce

1.  Go to the job template library (when not logged in)
2. Click the name of a template
3. Click the download button on the left side

### 📸 Screenshot

Image

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Anonymous users can download job templates

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

esizer:
🤖 Resolves #17356 

## 👋 Introduction

Fixes an issue where unauthenticated users could not download job poster templates.

## 🕵️ Details

We have added a public folder than anyone can read from. We allow unautenticated users to read from this folder. Since job postr templates do not contain sensitive information, we allow them to read these and download them.

This adds a separate disk for public files to make it less likely to leak guarded/unguarded files.

## 🧪 Testing

1. Refresh `make refresh-api`
2. Ensure you aere not logged in
3. Navigate to `/job-templates`
4. Go to one of the templates
5. Attempt to download it
6. Confirm the download works

### DIFF
--- a/api/app/GraphQL/Mutations/DownloadJobPosterTemplateDoc.php
+++ b/api/app/GraphQL/Mutations/DownloadJobPosterTemplateDoc.php
@@ -19,9 +19,11 @@ final readonly class DownloadJobPosterTemplateDoc
             $targetPoster = JobPosterTemplate::findOrFail($args['id']);
             $generator = new JobPosterTemplateGenerator(
                 jobPoster: $targetPoster,
-                dir: FilePath::PUBLIC_PATH,
+                dir: null,
                 lang: App::getLocale(),
             );
+
+            $generator->setDiskName(FilePath::PUBLIC_DISK);
 
             $generator->generate()->write();
 
--- a/api/app/GraphQL/Mutations/DownloadJobPosterTemplateDoc.php
+++ b/api/app/GraphQL/Mutations/DownloadJobPosterTemplateDoc.php
@@ -6,28 +6,22 @@ namespace App\GraphQL\Mutations;
 
 use App\Generators\JobPosterTemplateGenerator;
 use App\Models\JobPosterTemplate;
+use App\Support\FilePath;
 use Illuminate\Support\Facades\App;
-use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Log;
-use Illuminate\Validation\UnauthorizedException;
 
 final readonly class DownloadJobPosterTemplateDoc
 {
     /** @param  array{id: string|null}  $args */
     public function __invoke(null $_, array $args)
     {
-        $user = Auth::user();
-        throw_unless(is_string($user?->id), UnauthorizedException::class);
-
         try {
             $targetPoster = JobPosterTemplate::findOrFail($args['id']);
             $generator = new JobPosterTemplateGenerator(
                 jobPoster: $targetPoster,
-                dir: $user->id,
+                dir: FilePath::PUBLIC_PATH,
                 lang: App::getLocale(),
             );
-
-            $generator->setAuthenticatedUserId($user->id);
 
             $generator->generate()->write();
 
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -4185,6 +4185,7 @@ type Mutation {
     @canFind(ability: "view", find: "id", model: "PoolCandidate")
   downloadJobPosterTemplateDoc(id: UUID!): String
     @canFind(ability: "view", find: "id", model: "JobPosterTemplate")
+    @throttle(maxAttempts: 20)
   downloadNominationDoc(id: UUID!): String
     @guard
     @canFind(ability: "view", find: "id", model: "TalentNominationGroup")
--- a/api/graphql/schema.graphql
+++ b/api/graphql/schema.graphql
@@ -4185,7 +4185,6 @@ type Mutation {
     @canFind(ability: "view", find: "id", model: "PoolCandidate")
   downloadJobPosterTemplateDoc(id: UUID!): String
     @canFind(ability: "view", find: "id", model: "JobPosterTemplate")
-    @throttle(maxAttempts: 20)
   downloadNominationDoc(id: UUID!): String
     @guard
     @canFind(ability: "view", find: "id", model: "TalentNominationGroup")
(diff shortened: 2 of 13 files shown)


api/app/GraphQL/Mutations/DownloadJobPosterTemplateDoc.php
vd1992, line 1: Downloading a template in French fails, English works
Going to `/fr/job-templates/:id` and clicking download

Image

Image

Image

.
In `api/app/Support/FilePath.php`
Claude recommendation below

> FilePath::sanitize() strips the curly apostrophe that French copy actually uses, so widen the allowlist to match what the docblock already claims ("Preserves French characters"):

``` diff
-        $allowedPattern = $preserveExtension ? '/[^\p{L}\p{N}\s\-\_\.]/u' : '/[^\p{L}\p{N}\s\-\_]/u';
+        $allowedPattern = $preserveExtension ? '/[^\p{L}\p{N}\s\-\_\.\'’]/u' : '/[^\p{L}\p{N}\s\-\_\'’]/u';
```

api/graphql/schema.graphql
```diff
@@ -4184,8 +4184,8 @@ type Mutation {
     @guard
     @canFind(ability: "view", find: "id", model: "PoolCandidate")
   downloadJobPosterTemplateDoc(id: UUID!): String
-    @guard
     @canFind(ability: "view", find: "id", model: "JobPosterTemplate")
+    @throttle(maxAttempts: 20)
```
vd1992: The `@throttle` as is isn't that good apparently. Very important to consider further given what we are introducing with this PR

Per Claude

>Plainly, the problem is that the throttle counts requests per caller but has no reliable way to tell callers apart. It identifies them by the IP address the request itself claims, and because the app trusts all proxies, that value comes from a header the caller sends. So a caller picks which counter they're counted against — meaning the only people actually held to 20 per minute are the ones sending ordinary requests. And because the counter isn't tied to this mutation, unrelated traffic can use it up.

Seems like anonymous users could bypass the restriction on the download if we kept it as is.
Recommended improvements

> Name the limit

Use the `name` field in the directive. Prevents unintended overlap with other uses of throttling 

> The global limit is the part that actually bounds abuse

Can bypass per user limits, cannot bypass global limits. Introduce a `RateLimiter` that defines limits there per user as well as for all. The prior added `name` links to this
The `throttle:graphql` doesn't work best for anonymous users as it called by a client-supplied key
Relevant docs you may appreciate https://laravel.com/docs/13.x/routing#rate-limiting

As well, @petertgiles ran into a similar problem with PR https://github.com/GCTC-NTGC/gc-digital-talent/pull/11865
This PR sorta undoes that work

vd1992 (dismissed):
Some high impacts things 👇

api/graphql/schema.graphql
esizer: Umm, I'm a little confused. What exactly should be done here?

vd1992: Create a `RateLimiter` method within `configureRateLimiting()` for `RouteServiceProvider`
With the `for` matching the `name` added to `@throttle` and best to drop the `maxAttempts` from the directive and let the limiter be the sole source to avoid confusion.
Attach limit per user as well as a global limit

Important as it is possible to bypass per caller limits with anonymous downloads
Extend the pattern introduced by Peter in https://github.com/GCTC-NTGC/gc-digital-talent/pull/11865

Sample code with `name: document-generation`
The `by('all')` being the most protective

```
RateLimiter::for('document-generation', function (Request $request) {
    return [
        Limit::perMinute(20)->by(
            $request->user()->id
            ?? $request->cookie('ai_user')
            ?? $request->ip()
        ),
        Limit::perMinute(60)->by('all'),
    ];
});
```

esizer: :confused: This seems to far exceed the scope of this. I'm not sure this is related to the bug fix to allow unauthenticated users to download files.

vd1992: Hm... I'd lean towards caution for this PR since this PR introduces a significant change (anonymous downloading)
This doesn't seem like a big file diff to me, just adding the `document-generation` name `RateLimiter` as the fourth call in this grouping below
https://github.com/GCTC-NTGC/gc-digital-talent/blob/af6542dd3215703d2a84994381ec59db7f009f03/api/app/Providers/RouteServiceProvider.php#L51-L86

Claude did test the snippet for what its worth and
>the global by('all') limit still tripped at 61 when every request arrived with a different identity, which is the part a per-caller limit can't do alone.

But, let's get a third person in the mix. Thoughts on this discussion? @tristan-orourke

api/app/GraphQL/Mutations/DownloadJobPosterTemplateDoc.php
vd1992: We (heh) found a very similar bug that affects a totally unrelated file
So issue created by Claude, edited by me. Link for your reference
https://github.com/GCTC-NTGC/gc-digital-talent/issues/17762

api/graphql/schema.graphql
tristan-orourke: It looks like these document downloads go through the existing api and graphql paths already covered by the RateLimiter. A new `document-generation` RateLimiter seems unnecessary.

tristan-orourke: But it does mean we should be able to remove the `@throttle` directive, and use the normal RateLimiter config like everything else.

tristan-orourke (changes requested):
Looks good to me. Just remove that `throttle` directive and I'll approve

api/graphql/schema.graphql
esizer: Removed in [remove throttle](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17725/commits/0ab09428c3901086a802027f21c512bc0ec93a13)

tristan-orourke (approved):

api/graphql/schema.graphql
vd1992: Okay, that got dug into. 
Just to remind everyone, it appears the applicable global limit would then be `600`. Environment variables

PR #17726, opened 2026-08-12
[Bug] Experience skill validation

### LINKED ISSUE #17406: 🐛 Experience skill connect/sync accepts arbitrary skill IDs
### Bug

All 10 experience create/update mutations pass skill IDs through `Experience::connectSkills()` / `syncSkills()` to `User::addSkills()` (`User.php:447`), which creates `UserSkill` records without checking that the IDs exist in the `skills` table. Only `CreateUpdateWorkExperienceValidator` has any validation for this path, and it has no skill rules. The other 4 experience types have no validator at all.

Related to the class of bug fixed in #17223.

### Steps to Reproduce

1. Authenticate as any applicant.
2. Call `createWorkExperience` (or any experience create/update) with a `skills.connect` array containing a nonexistent skill UUID.
3. The request fails with an internal server error caused by a foreign key violation. This should be caught with input validation before it reaches the database.

### Proposed Solution

Add `Rule::exists('skills', 'id')` validation for skill IDs in experience mutations — either in `CreateUpdateWorkExperienceValidator` and new validators for the other 4 types, or via a shared rule.

### Acceptance Criteria

- [ ] Experience create/update mutations reject skill IDs that do not exist in the `skills` table
- [ ] Existing experience functionality is unaffected

yonikid15:
🤖 Resolves #17406

## 👋 Introduction

Connecting a skill to an experience (Work/Award/Community/Education/Personal) didn't check that the skill ID actually existed.

- Added `@rules(apply: ["exists:skills,id"])` to `ConnectExperienceSkills.id` in `schema.graphql`. Since all five experience types share this input for `connect`/`update`/`sync`, one schema change covers all of them.
- Added an `apiMessages` entry (`ErrorCode.SkillNotFound`) so the error renders proper text on the frontend instead of falling back to "Unknown error".
- `disconnect` intentionally left unvalidated — removing a nonexistent skill ID is a harmless no-op.
- Added tests to `WorkExperienceTest` covering both the rejection and the still-working success case.

## 🧪 Testing

1. Run `php artisan test --filter WorkExperienceTest`.
2. Manually: create/update a experience, with a non-existent skill ID in `skills.connect` — confirm you get a `SKILL_NOT_FOUND` validation error, not a 500. Then repeat with a real skill ID and confirm it connects successfully.


esizer (changes requested):
I think this should be a validator that uses our error codes rather than aloose rule in the schema.

That way we can more eaisly use our error code enum and keep the rules co-located rather that scattered in the schema.

esizer (approved):

esizer (approved):

PR #17755, opened 2026-08-13
[Chore] Assert minimum KLC skill count in talent nomination test

brindasasi:
## Introduction
Adds an explicit assertion that at least 3 "key leadership competencies" skills exist before the test selects 3 of them, so missing seed data fails with a clear message instead of a confusing selector timeout.

## Testing
Run the talent-nomination Playwright spec against seeded data.


mauli0110 (approved):
LGTM

PR #17756, opened 2026-08-13
Code quality finding -Array.prototype.sort mutates the original array in place

brindasasi:
_This PR applies 1/1 suggestions from code quality [AI findings](https://github.com/GCTC-NTGC/gc-digital-talent/security/quality/ai-findings)._

`Array.prototype.sort` mutates the original array in place. Since `poolSkills` is a parameter passed in, this will mutate the caller's array unexpectedly. Consider creating a shallow copy before sorting, e.g. `[...poolSkills].sort(...)`.


tristan-orourke (approved):
Makes sense

PR #17758, opened 2026-08-13
[Fix] Set department on test candidate in TalentRequestMatchesTest

brindasasi:
## Introduction
Sets `department_id` on the in-department test candidate's work experience so `testFiltersByDepartments` actually exercises the department filter instead of relying on unset factory data.

## Testing
`php artisan test --filter TalentRequestMatchesTest`

### DIFF
--- a/api/tests/Feature/TalentRequestMatchesTest.php
+++ b/api/tests/Feature/TalentRequestMatchesTest.php
@@ -465,6 +465,9 @@ class TalentRequestMatchesTest extends TestCase
         $department = Department::factory()->create();
         $pool = Pool::factory()->candidatesAvailableInSearch()->create();
         $inDepartment = $this->matchingUser($pool, [], true);
+        $inDepartmentWorkExperience = $inDepartment->currentGovWorkExperience()->firstOrFail();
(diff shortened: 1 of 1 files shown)


api/tests/Feature/TalentRequestMatchesTest.php
```diff
@@ -465,6 +465,9 @@ public function testFiltersByDepartments(): void
         $department = Department::factory()->create();
         $pool = Pool::factory()->candidatesAvailableInSearch()->create();
         $inDepartment = $this->matchingUser($pool, [], true);
+        $inDepartmentWorkExperience = $inDepartment->currentGovWorkExperience()->firstOrFail();
```
mnigh, line 468: `App\Models\User::currentGovWorkExperience()` is not a method in the user model. did you mean
`latestCurrentGovernmentWorkExperience`?

https://github.com/GCTC-NTGC/gc-digital-talent/blob/ab48f8856331494d8456579537bddb63383cef0c/api/app/Models/User.php#L548

brindasasi: actually this is wrong.. let me close this for now

PR #17761, opened 2026-08-13
[Fix] Flaky nomination test 2

### LINKED ISSUE #17644: 🐛 Flaky E2E test - Create nomination

esizer:
🤖 Resolves #17644 

## 👋 Introduction

Another attempt at fixing the flaky nomination test.

## 🕵️ Details

Last attempt made it less flaky but still not consistent. This moves around the waiting for responses to hopefully make it more consistent :crossed_fingers: 

## 🧪 Testing

1. Run test many times (10-20)
2. Confirm it passes consistently


yonikid15:
Solution by Claude :robot: 

When you hit submit, the app is supposed to show a confirmation screen right away. But **the confirmation screen wasn't tied to the submit at all** — it was tied to a separate, automatic background refresh of the page that was *supposed* to happen right after submitting. Most of the time that refresh fired and everything looked fine. Sometimes it just **didn't fire**, and since the confirmation screen was waiting on it, the heading never appeared — the test just sat there for **70 seconds** waiting for something that was never coming.

I found this by digging into the **network trace** of a failed run — after every other step in the form, there's always a follow-up refresh request. **After the final submit, there isn't.** That's the smoking gun.

There was also a smaller bug in the test itself: it was watching for that refresh request, but sometimes caught a **leftover one from the previous step** instead of one from the actual submit. Fixed that too, but it turned out **not to be the real problem** (still failed at the same rate after).

The actual fix: the submit code already had a flag (`submitting: true`) it set the instant the mutation succeeded — it just **wasn't being used** to show the confirmation screen. Now it is. So the confirmation shows **immediately based on "the submit worked,"** instead of waiting around for a refresh that might never come.


image

image

-------------------------

If you approve, I can commit the changes, or you can do it yourself :smile:

mauli0110:
> Solution by Claude 🤖
> 
> When you hit submit, the app is supposed to show a confirmation screen right away. But **the confirmation screen wasn't tied to the submit at all** — it was tied to a separate, automatic background refresh of the page that was _supposed_ to happen right after submitting. Most of the time that refresh fired and everything looked fine. Sometimes it just **didn't fire**, and since the confirmation screen was waiting on it, the heading never appeared — the test just sat there for **70 seconds** waiting for something that was never coming.
> 
> I found this by digging into the **network trace** of a failed run — after every other step in the form, there's always a follow-up refresh request. **After the final submit, there isn't.** That's the smoking gun.
> 
> There was also a smaller bug in the test itself: it was watching for that refresh request, but sometimes caught a **leftover one from the previous step** instead of one from the actual submit. Fixed that too, but it turned out **not to be the real problem** (still failed at the same rate after).
> 
> The actual fix: the submit code already had a flag (`submitting: true`) it set the instant the mutation succeeded — it just **wasn't being used** to show the confirmation screen. Now it is. So the confirmation shows **immediately based on "the submit worked,"** instead of waiting around for a refresh that might never come.
> 
> image image
> If you approve, I can commit the changes, or you can do it yourself 😄

This solution seems work like 90% all the time:  
image
But I am unsure what did it changed in 'success.tsx' 🤔

vd1992 (changes requested):
Testing this thoroughly in CI really isn't feasible. 
Between the amount of runs needed, time per run, and CI throttling it becomes a multiple day affair, doubly worse if you want to check additional changes. 

So I had Claude try and reproduce the problem locally to test the fixes reliably. 

#### NOTE
Mauli had a failed run but it wasn't for test "Create nomination", so not applicable here making this easier 😃 
#### /NOTE

#### Local reproduction
Problem appears to be a race condition.

Confirmation requires value `submittedAt` which is obtained by `NominateTalent` query and not through Submit. 
Problem arises from near simultaneous re-fetch of `NominateTalent` and submitting. What returns first varies, and that is what causes test flakiness. As submit returning first prevents proper re-fetch by URQL leaving stale cache

Claude local reproduction then depended on manually ensuring Submit arrived first. 

>Reproduction depended on letting the `NominateTalent` refetch hit the server before the submit, then holding its response until after — so the page receives a genuinely stale `submittedAt`. Simply delaying the refetch doesn't reproduce it; the request then reads post-submit data and the page renders fine.

#### Testing against your solution

The problem wasn't resolved and the flakiness still possible

>Ran your branch's spec under the same forced reproduction — fails identically to main. It's test-only, so it can't change whether the confirmation renders; the blank page is upstream of anything the spec waits on.

#### Recommended solution

My bot agrees with the YoniBot on the changes to `Success.tsx`, heh. 

>Agreed on the app change — that's what I tested, and it's what fixes it. The spec change I'd leave out: instrumenting it shows `rationaleQueried` resolving on the previous step's refetch, before the rationale save has even returned, so the submit still lands in the race window. Same leftover-response problem it was meant to solve. The app fix makes it unnecessary either way.

Claude diff generated by me locally, agrees with Yoni

``` diff
--- apps/web/src/pages/TalentNominations/NominateTalent/components/Success.tsx
+++ apps/web/src/pages/TalentNominations/NominateTalent/components/Success.tsx

+import { useLocation, type Location } from "react-router";
+
+type TLocation = Location<{ submitting?: boolean }>;

+  const location: TLocation = useLocation() as TLocation;
+  const justSubmitted = !!location.state?.submitting;
+
-  if (current || !talentNomination.submittedAt) {
+  if (current || (!talentNomination.submittedAt && !justSubmitted)) {
     return null;
   }
```
#### Does it work

Testing against the local setup. This solution _does_ work against the local reproduction running it multiple times. 

>Locally: 13/13 against the forced reproduction (2/2 fail before the change), 25/25 unmodified.
> 
> Separately, the test's await waitForGraphqlResponse("NominateTalent") after submit should go. Force a refetch that never returns and the test times out on that line with the confirmation heading already rendered on the page. It's now waiting on something the fix makes optional.

Up to you to decide what is next

vd1992:
Ugh, spec 1 still flaky. 
BRB giving Claude a piece of my mind

image

vd1992 (approved):
Exercising nomination submission manually and it still works.

Per our chatter.
There appears to be two possible problems at play, one now appears resolved. I'll include a AI suggestion below for the second problem we talked about. 
Then leave it up to you whether to create a separate issue or ignore it if it seems inaccurate.

I will hit approve, assuming no one else has any strong objections. 

Lastly, for your possible interest
Diff:

``` diff
--- a/apps/playwright/utils/talentNominationEvent.ts
+++ b/apps/playwright/utils/talentNominationEvent.ts
@@ -1,4 +1,5 @@
 import type {
+  CommunityDevelopmentProgram,
   CreateTalentNominationEventInput,
   TalentNominationEvent,
 } from "@gc-digital-talent/graphql/schema-types";
@@ -42,11 +43,32 @@ export const createTalentNominationEvent: GraphQLRequestFunc<
   Partial<CreateTalentNominationEventInput>
 > = async (ctx, talentNominationEvent) => {
   const communities = await getCommunities(ctx, {});
-  const firstCommunity = communities[0];
-  const communityId =
-    talentNominationEvent.community?.connect ?? firstCommunity.id ?? "";
-  const communityDevelopmentPrograms =
-    await getCommunityDevelopmentProgramsForCommunity(ctx, { communityId });
+
+  // An event's includeDevelopmentOpportunities is computed server side as
+  // communityDevelopmentPrograms->count() > 0, and the seeders leave it to
+  // chance which communities have any (TalentNominationEventFactory assigns
+  // events to Community::inRandomOrder()). Picking communities[0] blindly can
+  // therefore produce an event where the development opportunities option
+  // never renders, so prefer a community that actually has a program.
+  let communityId = talentNominationEvent.community?.connect ?? "";
+  let communityDevelopmentPrograms: CommunityDevelopmentProgram[] = [];
+  if (communityId) {
+    communityDevelopmentPrograms =
+      await getCommunityDevelopmentProgramsForCommunity(ctx, { communityId });
+  } else {
+    for (const community of communities) {
+      const programs = await getCommunityDevelopmentProgramsForCommunity(ctx, {
+        communityId: community.id,
+      });
+      if (programs[0]?.id) {
+        communityId = community.id;
+        communityDevelopmentPrograms = programs;
+        break;
+      }
+    }
+    communityId = communityId || (communities[0]?.id ?? "");
+  }
+
   const communityDevelopmentProgramsSync = communityDevelopmentPrograms[0]?.id
     ? [
         {
@@ -72,5 +94,14 @@ export const createTalentNominationEvent: GraphQLRequestFunc<
         },
       },
     })
-    .then((res) => res.createTalentNominationEvent);
+    .then((res) => {
+      if (!res.createTalentNominationEvent) {
+        // A denied create returns data.createTalentNominationEvent = null, which
+        // otherwise surfaces much later as a timeout on an unrelated locator.
+        throw new Error(
+          `Failed to create talent nomination event for community ${communityId}`,
+        );
+      }
+      return res.createTalentNominationEvent;
+    });
 };
```

PR #17765, opened 2026-08-13
Fix nominations download crash for archived nominees

### LINKED ISSUE #17715: 🐛 UAT & Prod | Talent Nominations download fail

- In  Prod and UAT the talent nominations download is failing.

**Expected Outcome:** Talent Nominations should get downloaded without any issues

### 🕵️ Details

Env and version: 
-   UAT - 2.79.0
-   Prod - 2.78.0

### 📋 Steps to Reproduce
1. Nominate a user in the system for any event 
2. Soft delete that user - User -> Advanced Tools -> Archive User
3. Open the nomination event
4. Observe the nominee displays as No name provided 
5. Try download the nominees excel includding that nominee - observe error 
6. Try skipping that nominee and download other nominations - observe download passes 

### Prod:

Image

### UAT:

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Soft deleted user's nomination should not be shown in the nomination table

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

brindasasi:
🤖 Resolves #17715

## 👋 Introduction

Nominations download crashed whenever a talent event had a nomination group for an archived (soft-deleted) nominee, since the group still pointed at them but the `nominee` relation resolved to null. Archived nominees should behave like every other archived user on the platform: invisible everywhere, not just in this one export.

## 🕵️ Details

Added a global scope on `TalentNominationGroup` that excludes groups whose nominee is soft-deleted. It reads the live `deleted_at` state, so no migration or backfill is needed, and existing broken records are fixed on deploy. Also fixed an edge case in `connectToTalentNominationGroupIfMissing()` where a second nomination for an already-archived nominee would have hit the group's unique constraint instead of reusing the hidden group.

## 🧪 Testing

1. Archive a nominee who has an existing nomination group
2. Confirm their group no longer appears in the event's nominations table
3. Confirm the nominations download for that event succeeds instead of erroring
4. Restore the nominee, confirm the group reappears

Covered by new automated tests in `TalentNominationGroupTest` and `NominationsExcelGeneratorTest`. Confirmed the tests reproduce the original crash when the fix is reverted.

## 📸 Screenshot

N/A, backend-only fix.


api/app/Models/TalentNomination.php
```diff
@@ -213,8 +213,11 @@ public function connectToTalentNominationGroupIfMissing()
         if (! is_null($this->submitted_at)) {
             // this is a submitted nomination
             if (is_null($this->talent_nomination_group_id)) {
-                // not yet attached to a group
-                $talentNominationGroup = TalentNominationGroup::firstOrCreate(
+                // not yet attached to a group;
+                // look up regardless of the nominee's archived
+                // status so an already-existing group is reused instead of violating the
+                // nominee_id/talent_nomination_event_id uniqueness constraint
+                $talentNominationGroup = TalentNominationGroup::withoutGlobalScope('activeNominee')->firstOrCreate(
```
brindasasi, line 225: This is to protect from an edge case 

1. Nominee Alex gets nominated for Event A → a TalentNominationGroup row is created (nominee_id: Alex, event: A).
2. Alex's user account later gets archived. The group row still physically exists in the table — we never delete it — but the global scope now hides it from every normal query, since whereHas('nominee') fails for an archived nominee.
3. A second person nominates Alex for the same event A again at the same time step 2 is happening.

Eloquent proceeds to create() a new row with the same nominee_id + talent_nomination_event_id.

The insert violates the constraint and throws a database error, crashing the nomination submission.

The above code will prevent this crash from happening.

api/app/Models/TalentNominationGroup.php
```diff
@@ -81,6 +81,11 @@ public function getActivitylogOptions(): LogOptions
     protected static function booted(): void
     {
         TalentNominationGroup::observe(TalentNominationGroupObserver::class);
+
+        // only pull active users ,skip archived ones
+        static::addGlobalScope('activeNominee', function (Builder $query) {
+            $query->whereHas('nominee');
+        });
```
vd1992, line 93: I believe this would be our first implementation of a global scope. 
Looking at the docs, there doesn't seem to be anything of serious concern present

https://laravel.com/docs/13.x/eloquent#anonymous-global-scopes

Digging around however and it appears there is one potential concern of note in that the global scope would apply to all `TalentNominationGroup::class` Eloquent operations. The problem being this also includes things like `update` operations. So mass updates of rows would silently filter with the global scope

This can be easily avoided with `withoutGlobalScope` or using the DB Facade. So long as this is kept in mind.
I think my takeaway would be mass updates should prefer the Facade but that isn't an immediate problem.  

It is just a novel cause for error since as mentioned, first implementation of a global scope

brindasasi: so we have an existing cascade delete model which we introduced in https://github.com/GCTC-NTGC/gc-digital-talent/pull/7593

We covered pool candidates , experiences etc in that PR and cascade soft delete them.
But 3 years later no one remembers that obviously and forgot about nomination group .. thats how this bug surfaced. 

I could go back and do the same thing here but there will be 3 limitations 
1. There are already archived users in the Nomination Group table and we've to go manually soft delete them
2. That model would work only if the user is archived through that eloquent model .. if we run any `DB::table('users')->update(['deleted_at' => now()])`, a bulk script, a data migration, or a future cleanup job that bypasses the model , the bug will surface again
3. Couple years later another developer would forget this soft delete again for a new model and the production issue would surface again. 

Instead always picking up active users would be safer and common practice.

I could add the comment for `withoutGlobalScope`  just in case we needed to update something that involves archived users as well  ..in the future .. currently there is none

esizer (commented):
Unsure if we need this but, it will prevent a group from loading when accessing the group from it. Normally, this would not be an issue except when we need to update the group based on new nominations.

This occurs in the talent nominaton observer where we attempt to update the groups status:

https://github.com/GCTC-NTGC/gc-digital-talent/blob/1192ff990fc33a88b2567eaaec57f47fb5de3d77/api/app/Observers/TalentNominationObserver.php#L21-L23

https://github.com/GCTC-NTGC/gc-digital-talent/blob/1192ff990fc33a88b2567eaaec57f47fb5de3d77/api/app/Observers/TalentNominationObserver.php#L39-L41

I think we have different ways of solving this:

1. Ignore the global scope specifcally in these observers
2. Ignore the global scope in the relationship in `TalentNomination`
3. Use a local scope instead, applied in the generator query

Or, we don't care about the issue of updating the status 🤔

api/app/Observers/TalentNominationGroupObserver.php
```diff
@@ -25,8 +26,11 @@ public function updated(TalentNominationGroup $talentNominationGroup): void
         // if advancement decision changes to approved then record the classification
         if ($talentNominationGroup->getOriginal('advancement_decision') !== TalentNominationGroupDecision::APPROVED->name &&
             $talentNominationGroup->advancement_decision === TalentNominationGroupDecision::APPROVED->name) {
-            $talentNominationGroup->loadMissing('nominee');
-            $talentNominationGroup->classificationAtTimeOfAdvancementApproval()->associate($talentNominationGroup->nominee->currentClassification);
+            // nominee() excludes archived users by default; look up directly with
```
brindasasi, line 29: This is another edge scenario handled here preventing crash ..
1. Nominee has been nominated to an event , but the application is still draft 
2. Nominee has been archived 
3. Nomination has been successfully submitted ( we may need this part to be fixed later to prevent this happen. But currently there is no verification on submission for active nominees and so submission to archived nominees can happen  ) 
4. Now Admin is trying to approve for advancement 
5. This will crash  if we don't call the query with `withTrashed()` to pull archived nominees as well

esizer (approved):
LGTM! 👍

api/app/Models/TalentNominationGroup.php
tristan-orourke: You make a good argument that adding this global scope to new models as necessary is a better solution than trying to cascade the soft delete through every model that will every be connected to User. I'm convinced! :+1:

PR #17770, opened 2026-08-14
[Bug] Next role and career objective validation

### LINKED ISSUE #17034: 🐛 Your next role validation error

in the "Your next role" section on the GC Employee profile there is a notice that says "There are currently unanswered questions in this section. Use the "Edit" button to review and answer any required fields."

However, once you 'Edit' this section there isn't any warning about where the missing information is (infact- I don't think I actually am missing any information - unless the desired work streams is required).

Even when I save it doesn't remove the message.

My recommendation would be to do one of the following options:
a) make the desired work streams required
b) remove the warning once you open the form even if nothing has changed and its not required

### 🕵️ Details

comment from @petertgiles : The validation for the checkbox in the section title and the navigation thinks it is complete, too.

### 📋 Steps to Reproduce

1. go to GC Employee profile 
2. go to your next role 
3. fill out every required field and save

### 📸 Screenshot

Image

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Warning message should only be shown when required fields are missing. Missing optional fields should not trigger it.

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

yonikid15:
🤖 Resolves #17034 

## 👋 Introduction

- Changes hasAnyEmptyFields logic to only check for required fields
-  Warning message should only be shown when required fields are missing. Missing optional fields should not trigger it.

## 🧪 Testing

1. Build front-end `pnpm dev:fresh`
2. Ensure that the warning message only shows if required fields are missing

## 📸 Screenshot

image

image

### DIFF
--- a/apps/web/src/pages/EmployeeProfile/components/CareerObjective/Display.tsx
+++ b/apps/web/src/pages/EmployeeProfile/components/CareerObjective/Display.tsx
@@ -64,14 +64,12 @@ const Display = ({
   return (
     <div className="flex flex-col gap-y-6">
       {hasAnyEmptyFields({
-        careerObjectiveClassification,
         careerObjectiveTargetRole,
-        careerObjectiveJobTitle,
+        careerObjectiveTargetRoleOther,
         careerObjectiveCommunity,
         careerObjectiveCommunityOther,
-        careerObjectiveWorkStreams,
-        careerObjectiveDepartments,
-        careerObjectiveAdditionalInformation,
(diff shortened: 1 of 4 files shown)


esizer (changes requested):
Seems this makes the function name a liar now 😓 Can we update it to be more accurate? 😅

esizer (approved):

PR #17771, opened 2026-08-14
Fix candidate bookmark/flag cache issues without causing jumping in table

### LINKED ISSUE #16166: 🐛 UAT | Flagged candidate is not showing as flagged in candidate application page

### LINKED ISSUE #17704: 🐛 User Flag in processes don't appear to be saved

Under a process, when you flag a user the flag appears to work but other users cannot see it as flagged and if you refresh the page the flag disappears. However, if you click on the flag again, you get a popup saying flag removed so it seems to work but we cannot it the flag.

### 🕵️ Details

_No response_

### 📋 Steps to Reproduce

Go to any process and click on the flag near a users name.
Refresh page
Flag is gone.
Click on the flag icon for that same user and you get a popup saying flag was removed.

### 📸 Screenshot

Image

Image

### 🙋‍♀️ Proposed Solution

Make the flag show when clicked.

### ✅ Acceptance Criteria

_No response_

### 🔄 Regression

No

### 🔗 Root Cause Pull Request

_No response_

tristan-orourke:
🤖 Resolves #16166
Resolves #17704 

## 👋 Introduction
#17704 
This PR attempts to solve one issue without creating another.

#16166 describes an issue where if you view a PoolCandidate as a Community admin, update the flag or bookmark toggle, navigate to a different page, and navigate back, then the flag/toggle appears incorrectly (in the state before you toggled it).

#16167 resolved this issue, but by creating a related one: now toggling the flag/bookmark from a Process/Candidates table could cause rows in the table to jump around. Trying to avoid this in the first place by not refreshing the full cache of PoolCandidates is what led to the somewhat roundabout solution that ultimately caused issue #16166.

This PR resolves the navigate-away-nagivage-back staleness described in #16166 while avoiding jumping rows. It slightly simplifies the CandidateFlag component by removing some unused params, and uses a new hook `useServerSyncedState` which is both simpler than `useControllableState` and more customized for this specific use case.

## 🕵️ Details

How the flags work now:
1. When the page (candidates table or individual candidate) the useSeverSyncedState hook is initialized with the flag value from the overall PoolCandidate query.
2. User clicks a flag/bookmark to toggle it. This triggers a graphql mutation specific to the flag.
3. Button is disabled while mutation is in progress; UI does not update optimistically.
4. The mutation returns the new state of the flag itself, instead of the entire PoolCandidate object. If the mutation was successful, the hook saves the new value locally and displays it; local state now differs from serverValue.
5. If you navigate away and come back, the hook will be initialized with the flag value from the overallPoolCandidate query again - but that value will be stale, coming from the cache, while a new query is kicked off.
6. When the query returns, serverValue will differ from cached serverValue, and useServerSyncedState will adopt it as a new local value.

> [!NOTE]
> Why does the flag/bookmark mutation return only the new state of the boolean, not the entire PoolCandidate object?
> Returning the whole object would cause URQL's document cache to invalidate its cache for all PoolCandidates. This would immediately trigger a new query if you're on the Pool/Candidates table. This would result in re-ordered results, with newly flagged candidates appearing at the top and unflagged candidates moving later, and this re-ordering of table rows is what we want to avoid. 

### Downsides to this solution

1. When you navigate to a new page and navigate back after toggling a flag, it will briefly appear with the stale state before being updated by the new query.
2. The code for handling this specific case of this specific cache implementation still feels rather convoluted.

### Alternative solution

The code would be a lot simpler if we could rely on the normal operation of the cache. The main reason not to is the reordering of the table. So instead of working around the cache, we could consider working around the ordering behaviour:
1. When PoolCandidates query is first completed, we save the ordering of ids in local state
2. We then use that ordering to order the table rows, instead of whatever is returned by the api.
3. But... then we have to be really careful about when we invalidate the cached order. We'd want to maintain the cached order when running mutations on PoolCandidates, but we'd want to invalidate that cache any time the PoolCandidates query is modified (updated sort order or filters or pagination).

### navigate-away-navigate-back
1. Access to GC digital portal using a community role
3. Mark that candidate as flagged or bookmarked from Candidate application page
6. Verify that the flag appears in the correct state you left it in (you may briefly see it in stale state)

### jumping-rows
1. Access GC digital portal using a community role
2. Navigate to a Pool/Process
3. Navigate to Candidates table
4. Scroll down to a candidate that is not flagged or bookmarked, and does not appear at top of table
5. Mark candidate as bookmarked. Bookmark should be displayed, but candidate row should not move.
6. Change filters. Candidate should move to top of table, bookmark still showing.
7. Flag another candidate lower in the table.
8. Navigate to another page, then navigate back to candidates table. After a brief flash of stale state, flagged candidate should appear at top of table.

### DIFF
--- a/apps/web/src/components/CandidateFlag/CandidateFlag.tsx
+++ b/apps/web/src/components/CandidateFlag/CandidateFlag.tsx
@@ -32,16 +32,12 @@ export const PoolCandidate_FlagFragment = graphql(/* GraphQL */ `
 interface CandidateFlagProps {
   candidateQuery: FragmentType<typeof PoolCandidate_FlagFragment>;
   processTitle?: string | null;
-  onFlagChange?: (newIsFlagged: boolean) => void;
-  flagged?: boolean;
   size?: "sm" | "md" | "lg";
 }
 
 const CandidateFlag = ({
   candidateQuery,
   processTitle = null,
-  flagged,
-  onFlagChange,
   size = "md",
 }: CandidateFlagProps) => {
   const intl = useIntl();
@@ -55,8 +51,6 @@ const CandidateFlag = ({
   const [{ isFlagged, isUpdating: isUpdatingFlag }, toggleFlag] =
     useCandidateFlagToggle({
       id: candidate.id,
-      onChange: onFlagChange,
-      value: flagged,
       defaultValue: candidate?.isFlagged ?? false,
       name: candidateName,
       processTitle:
--- a/apps/web/src/hooks/useCandidateBookmarkToggle.test.tsx
+++ b/apps/web/src/hooks/useCandidateBookmarkToggle.test.tsx
@@ -1,0 +1,156 @@
+import { act, renderHook } from "@testing-library/react";
+import type { ReactNode } from "react";
+import type { Client } from "urql";
+import { Provider as GraphqlProvider, CombinedError } from "urql";
+import { fromValue } from "wonka";
+
+import { Providers } from "@gc-digital-talent/vitest-helpers";
+import { toast } from "@gc-digital-talent/toast";
+
+import useCandidateBookmarkToggle from "./useCandidateBookmarkToggle";
+
+vi.mock("@gc-digital-talent/toast", () => ({
+  toast: {
+    success: vi.fn(),
+    error: vi.fn(),
+  },
+}));
+
+/**
+ * Stub urql client returning a fixed mutation result.
+ * Ref: packages/storybook-helpers/decorators/MockGraphqlDecorator.tsx
+ */
+const mockClient = (result: Record<string, unknown>) =>
+  ({
+    executeMutation: vi.fn(() => fromValue(result)),
+  }) as unknown as Client;
+
+const renderBookmarkToggle = (client: Client, defaultValue = false) =>
+  renderHook(
+    (props: { defaultValue: boolean; id: string }) =>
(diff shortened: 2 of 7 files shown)


esizer:
I could be wrong but, maybe a simpler solution would be just adding an independant query for bookmark/flag state in the sidebar and setting it to `netowrk-only` would also solve this :thinking: 

```tsx
const FlagState_Query = graphql(`
  query FlagState(id: UUID!) {
    isBookmarked
    isFlagged
  }
`);

const [{ data, fethcing }] = useQuery({
  query: FlagState_Query,
  context: { requestPolicy: "network-only" }
});
```

This would be an additional query on the page that runs on every load but, it should be relativley inexpensive and ensure it always has the updated value without needing the change much else. :thinking:

tristan-orourke:
> adding an independant query for bookmark/flag state in the sidebar and setting it to network-only 

Do you mean adding a `poolCandidateFlagState(id: UUID!)` query in the graphql schema? Or would this just a second call to the poolCandidate query?

tristan-orourke:
This should resolve #17704  as well

tristan-orourke:
Replaced by a different approach, in #17774

PR #17772, opened 2026-08-14
[Fix] Exclude community-less pools from talent search results and estimated count

### LINKED ISSUE #17732: 🐛 UAT | Pools with no community appear in search results and inflate the estimated candidate count

- While investigating the mismatch in the estimated candidate count, I traced it back to a pool that isn't attached to any community. Looking at the GraphQL response for such pool, community comes back as null, the pool has `a classification and work stream assigned, but no community`. Because it's not tied to a community, it never rolls up under the community grouping, yet it still shows up in the search results and gets counted in the estimated candidates total.
- This traces back to when departments were introduced: a pool can be created under a department where community selection isn't mandatory and any work stream can be picked. So the work stream gets assigned to the pool, but no community does — leaving these orphaned pools.

**Expected Outcome:** A pool that isn't part of any community shouldn't appear on the search page, and its candidates shouldn't be included in the estimated count.

### 🕵️ Details

Env: UAT
Version: 2.79.0

### 📋 Steps to Reproduce

Prerequisite: Make sure to have one pool having workstream but not community [Hint- Create a pool using department admin where community selection is not mandatory but workstream can be selected]. Create a new user, apply to the same process and qualify the candidate.

1. Access GC digital portal
2. Navigate to the find talent page
3. Filter the classification and workstream [similar to the one created]
4. Observe the estimated candidate count
5. Navigate down to results section and observe the pool having no community is present and included the candidate count into estimated candidate

### 📸 Screenshot

https://github.com/user-attachments/assets/5a8eedfb-16c6-4890-aa86-2db988425ef5

Image

Image

### 🙋‍♀️ Proposed Solution

_No response_

### ✅ Acceptance Criteria

- [ ] Estimated candidate count should match with the count present in results by community
- [ ] The pool having no community attached, shouldn't be present in the search request results

### 🔄 Regression

Yes

### 🔗 Root Cause Pull Request

_No response_

yonikid15:
🤖 Resolves #17732 

## 👋 Introduction

Pools without a community attached were still showing up as cards on the talent search page, and were being counted in the "estimated candidates" total even though they were already excluded from the by-community breakdown. This PR excludes community-less pools from talent request matching entirely, so the search results and the estimated count agree.

## 🕵️ Details

 - Added a `whereHasCommunity()` scope to `PoolCandidateBuilder`
 - Wired it into the shared `whereMatchesTalentRequest()` scope, since the by-pool count, by-community count, and overall total all route through it.

## 🧪 Testing

1. Run `php artisan test --filter=TalentRequestMatchesTest`
2. On the search page, confirm a pool with no community attached does not appear as a card and isn't reflected in the estimated candidate count

### DIFF
--- a/api/app/Builders/PoolCandidateBuilder.php
+++ b/api/app/Builders/PoolCandidateBuilder.php
@@ -46,6 +46,16 @@ class PoolCandidateBuilder extends Builder implements TalentRequestMatchable
     }
 
     /**
+     * Scopes the query to exclude PoolCandidates in a pool that isn't attached to a community
+     */
+    public function whereHasCommunity(): self
+    {
+        return $this->whereHas('pool', function ($query) {
+            $query->whereNotNull('community_id');
+        });
+    }
+
+    /**
      * Scopes the query to return PoolCandidates in a pool with one of the specified classifications.
      * If $classifications is empty, this scope will be ignored.
      *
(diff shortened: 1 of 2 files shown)


tristan-orourke (approved):
Code makes sense, tests look reasonable, manual test passes.

PR #17774, opened 2026-08-15
Fix stale bookmark/flag state on the candidate application page

### LINKED ISSUE #16166: 🐛 UAT | Flagged candidate is not showing as flagged in candidate application page

### LINKED ISSUE #17704: 🐛 User Flag in processes don't appear to be saved

tristan-orourke:
🤖 Resolves #16166
~~Resolves #17704~~ unable to reproduce #17704

## 👋 Introduction

An alternative to #17771, taking the approach suggested in [this review comment](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17771#issuecomment-5296298496).

#16166 describes the bookmark/flag buttons on the candidate application page reverting to their pre-toggle state after you navigate away and come back. The buttons are seeded from the page's `PoolCandidateSnapshot` query, and `defaultValue` on the toggle hooks is only read once at mount — so on back-navigation they latch onto the stale cached document and never pick up the network result that follows.

Rather than teach the hooks to adopt a later server value (the `useServerSyncedState` approach in #17771), this PR simply asks the server for the current state in a dedicated `network-only` query, and doesn't render the real buttons until that answer arrives.

## 🕵️ Details

`ApplicationBookmarkFlag.tsx` now uses a `<Pending>` component to render a spinner until fresh data comes in from the server. This avoids initializing the flag and bookmark with stale cache data.

`isBookmarked` and `applicationAssessmentData { isFlagged }` were dropped from `ApplicationBookmarkFlag_Fragment` — they were the stale values, and nothing else on the page reads them.

Nothing about the candidates table changes. The toggle mutations still return a bare `Boolean` rather than a `PoolCandidate`, so URQL's document cache is not invalidated and rows still don't jump around (#16167).

The bookmark and flag buttons in PoolCandidateTable were slightly modified to avoid disabling buttons while the mutation is in progress. Instead, `useCandidateFlagToggle` and `useCandidateBookmarkToggle` avoid triggering a new mutation if one is in progress already. Thus, the behaviour of the table buttons now matches what was already in the sidebar.

### Trade-offs

- One extra small query per candidate page load. That's the cost of the approach, and it's cheap.
- After clicking a flag/bookmark button, there is no user feedback while the mutation is running. This is how it already worked on the sidebar, and the mutation runs fast enough on UAT that users haven't seemed bothered.

### navigate-away-navigate-back (#16166)

1. Sign in with a community role.
2. Open any pool candidate's application page.
3. Toggle the flag and the bookmark on.
4. Navigate to Processes, or any other page.
5. Use the browser back button to return to the application page.
6. A spinner should briefly replace the flag and bookmark buttons, and when they appear they should be in the correct state.

### jumping-rows no-regression (#16167)

1. Sign in with a community role, open a process and its Candidates table.
2. Scroll to a candidate that is neither flagged nor bookmarked and is not near the top.
3. Bookmark it. The icon fills and the row does **not** move.
4. Flag another candidate lower down. Same.
5. Change a filter. Rows re-order as expected, with the bookmark/flag states preserved.
6. Toggle a bookmark or flag, then click a link to go to another page.
7. Navigate back. After a brief moment when a query refresh is happening, the latest toggled flag should appear correctly.

## 📸 Screenshot

---

*Description generated by Claude Code.*

### DIFF
--- a/apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/ApplicationBookmarkFlag.tsx
+++ b/apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/ApplicationBookmarkFlag.tsx
@@ -7,7 +7,7 @@ import BookmarkIconOutline from "@heroicons/react/24/outline/BookmarkIcon";
 import BookmarkIconSolid from "@heroicons/react/24/solid/BookmarkIcon";
 
 import type { ButtonProps } from "@gc-digital-talent/ui";
-import { Button } from "@gc-digital-talent/ui";
+import { Button, Pending } from "@gc-digital-talent/ui";
 import type { FragmentType } from "@gc-digital-talent/graphql";
 import { getFragment, graphql } from "@gc-digital-talent/graphql";
 import { commonMessages } from "@gc-digital-talent/i18n";
@@ -58,37 +58,50 @@ const networkOnlyRequest: Partial<OperationContext> = {
   requestPolicy: "network-only",
 };
 
-const noop = () => {
-  // Placeholder handler for the buttons while their state is still unknown.
-};
-
 interface BookmarkFlagButtonsProps {
+  id: string;
+  name: string;
+  processTitle: string;
   isBookmarked: boolean;
   isFlagged: boolean;
-  bookmarkDisabled?: boolean;
-  flagDisabled?: boolean;
-  onToggleBookmark: () => void;
-  onToggleFlag: () => void;
 }
 
-/** Presentational bookmark and flag buttons, with no knowledge of where their state comes from. */
+/**
+ * Bookmark and flag buttons, owning the toggle mutations and seeded with freshly fetched
+ * server state.
+ *
+ * The toggle hooks only read their `defaultValue` on mount, so this must not be rendered
+ * until the real values are known.
+ */
 const BookmarkFlagButtons = ({
-  isBookmarked,
-  isFlagged,
-  bookmarkDisabled = false,
-  flagDisabled = false,
-  onToggleBookmark,
-  onToggleFlag,
+  id,
+  name,
+  processTitle,
+  isBookmarked: initialIsBookmarked,
+  isFlagged: initialIsFlagged,
 }: BookmarkFlagButtonsProps) => {
   const intl = useIntl();
+  const [{ isFlagged, isUpdating: isUpdatingFlag }, toggleFlag] =
+    useCandidateFlagToggle({
+      id,
+      defaultValue: initialIsFlagged,
+      name,
+      processTitle,
+    });
+  const [{ isBookmarked, isUpdating: isUpdatingBookmark }, toggleBookmark] =
+    useCandidateBookmarkToggle({
+      id,
+      defaultValue: initialIsBookmarked,
+      name,
+    });
 
   return (
     <div className="flex flex-col gap-y-4.5">
       <Button
         {...commonProps}
         icon={isBookmarked ? BookmarkIconSolid : BookmarkIconOutline}
-        disabled={bookmarkDisabled}
-        onClick={onToggleBookmark}
+        disabled={isUpdatingBookmark}
+        onClick={toggleBookmark}
       >
         {isBookmarked
           ? intl.formatMessage({
@@ -106,8 +119,8 @@ const BookmarkFlagButtons = ({
       <Button
         {...commonProps}
         icon={isFlagged ? FlagIconSolid : FlagIconOutline}
-        disabled={flagDisabled}
-        onClick={onToggleFlag}
+        disabled={isUpdatingFlag}
+        onClick={toggleFlag}
       >
         {isFlagged
           ? intl.formatMessage({
@@ -125,53 +138,6 @@ const BookmarkFlagButtons = ({
   );
 };
 
-interface LoadedBookmarkFlagProps {
-  id: string;
-  name: string;
-  processTitle: string;
-  isBookmarked: boolean;
-  isFlagged: boolean;
-}
-
-/**
- * Owns the toggle mutations, seeded with freshly fetched server state.
- *
- * The toggle hooks only read their `defaultValue` on mount, so this is deliberately kept as a
- * separate component: it is not rendered until the real values are known.
- */
-const LoadedBookmarkFlag = ({
-  id,
-  name,
-  processTitle,
-  isBookmarked: initialIsBookmarked,
-  isFlagged: initialIsFlagged,
-}: LoadedBookmarkFlagProps) => {
-  const [{ isFlagged, isUpdating: isUpdatingFlag }, toggleFlag] =
-    useCandidateFlagToggle({
-      id,
-      defaultValue: initialIsFlagged,
-      name,
-      processTitle,
-    });
-  const [{ isBookmarked, isUpdating: isUpdatingBookmark }, toggleBookmark] =
-    useCandidateBookmarkToggle({
-      id,
-      defaultValue: initialIsBookmarked,
-      name,
-    });
-
-  return (
-    <BookmarkFlagButtons
-      isBookmarked={isBookmarked}
-      isFlagged={isFlagged}
-      bookmarkDisabled={isUpdatingBookmark}
-      flagDisabled={isUpdatingFlag}
-      onToggleBookmark={toggleBookmark}
-      onToggleFlag={toggleFlag}
-    />
-  );
-};
-
 interface ApplicationBookmarkFlagProps {
   query: FragmentType<typeof ApplicationBookmarkFlag_Fragment>;
 }
@@ -181,7 +147,7 @@ const ApplicationBookmarkFlag = ({ query }: ApplicationBookmarkFlagProps) => {
   const application = getFragment(ApplicationBookmarkFlag_Fragment, query);
 
   // This query is run separately without a cache policy, to avoid rendering stale values.
-  const [{ data, fetching }] = useQuery({
+  const [{ data, fetching, error }] = useQuery({
     query: ApplicationBookmarkFlagState_Query,
     variables: { id: application.id },
     context: networkOnlyRequest,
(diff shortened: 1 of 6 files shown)


apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/ApplicationBookmarkFlag.tsx
```diff
@@ -111,4 +125,102 @@ const ApplicationBookmarkFlag = ({ query }: ApplicationBookmarkFlagProps) => {
...
+      <BookmarkFlagButtons
+        isBookmarked={false}
+        isFlagged={false}
+        bookmarkDisabled
+        flagDisabled
+        onToggleBookmark={noop}
+        onToggleFlag={noop}
+      />
+    );
+  }
```
esizer: This part is a little confusing to me. If there is no candidate, rendering any buttons seems odd 🤔 Shouldn't we just return `null` or some loading state to let the user know what is going on?

Also, if the queyr fails, this will lead to two disabled buttons that seem to indicate that a candidate is neither flagged or bookmarked (possibly inacurate since it is hardcoded) with no way of knowing why you cannot change that.

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/ApplicationBookmarkFlag.tsx
```diff
@@ -93,7 +106,8 @@ const ApplicationBookmarkFlag = ({ query }: ApplicationBookmarkFlagProps) => {
       <Button
         {...commonProps}
         icon={isFlagged ? FlagIconSolid : FlagIconOutline}
-        onClick={toggleFlag}
+        disabled={flagDisabled}
```
esizer: Ideally, we do not disable buttons. This would essentially remove them from the DOM making them undiscoverable. It also doesn't tell a user why it is dsiabled.

We should prefer using `aria-disabled` and show some message letting users know why something was not possible. (Maybe a toast, or change the button text) 🤔

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/ApplicationBookmarkFlag.tsx
```diff
@@ -35,47 +37,58 @@ const ApplicationBookmarkFlag_Fragment = graphql(/** GraphQL */ `
...
+const networkOnlyRequest: Partial<OperationContext> = {
+  requestPolicy: "network-only",
+};
 
-const ApplicationBookmarkFlag = ({ query }: ApplicationBookmarkFlagProps) => {
-  const intl = useIntl();
-  const application = getFragment(ApplicationBookmarkFlag_Fragment, query);
+const noop = () => {
+  // Placeholder handler for the buttons while their state is still unknown.
+};
```
esizer: This seems a little odd to me. If a user hits this no-op and nothing happens, it might be very confusing. I feel like we should tell the user why nothing happened, otherwise, it could seem like the app is just broken or they encountered a bug.

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/ApplicationBookmarkFlag.tsx
```diff
@@ -111,4 +125,102 @@ const ApplicationBookmarkFlag = ({ query }: ApplicationBookmarkFlagProps) => {
...
+}
+
+/**
+ * Owns the toggle mutations, seeded with freshly fetched server state.
+ *
+ * The toggle hooks only read their `defaultValue` on mount, so this is deliberately kept as a
+ * separate component: it is not rendered until the real values are known.
+ */
+const LoadedBookmarkFlag = ({
+  id,
```
esizer: I'm not understanding why this needs to be a standalone component. Could this not be rolled into a single component with some conditionals in the props 🤔 

My understanding is that this is the "working" version afetr everything has loaded and it passes the appropriate props to a pure version? Though, I'm not so sure the inert version should exist since it could lead to a lot of UX confusion.

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/ApplicationBookmarkFlag.tsx
```diff
@@ -111,4 +125,102 @@ const ApplicationBookmarkFlag = ({ query }: ApplicationBookmarkFlagProps) => {
...
+      defaultValue: initialIsBookmarked,
+      name,
+    });
+
+  return (
+    <BookmarkFlagButtons
+      isBookmarked={isBookmarked}
+      isFlagged={isFlagged}
+      bookmarkDisabled={isUpdatingBookmark}
+      flagDisabled={isUpdatingFlag}
```
esizer: This will cause the button to be removed from the tab order so focus is getting sent to the body when you toggle flags making this a very difficult interactioon for keyboard or AT only users.

esizer (changes requested):
While I think this is better than the other solution. It sitll seems to increase complexity and result in a a degredation of UX, especially those who rely on alternative navigation methods.

The particular issues mostly derive from having disabled controls and no-ops.

tristan-orourke:
This should resolve https://github.com/GCTC-NTGC/gc-digital-talent/issues/17704 as well

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/ApplicationBookmarkFlag.tsx
tristan-orourke: This noop was only used in cases where the buttons were disabled anyway, and was defined here to be stable across renders. But its gone in https://github.com/GCTC-NTGC/gc-digital-talent/pull/17774/commits/c03ef6def811cf01b570f870b1a93a0b88bd1a07

tristan-orourke:
Okay @esizer I've used the Pending component to render a spinner while the flags first load in, instead of disabling the buttons. That also removes any reason to separate the Presentational component from the stateful one, so I've put them back into one component. I think this should address most of your comments.

The one thing I want to ask you about is disabling buttons while a toggle mutation is in progress. On the [candidates table](https://github.com/GCTC-NTGC/gc-digital-talent/blob/ab48f8856331494d8456579537bddb63383cef0c/apps/web/src/components/PoolCandidatesTable/PoolCandidateBookmark.tsx#L54) we disable the flag buttons while a toggle mutation is in progress; in this sidebar, we weren't , and in my testing you often got an error (with a toast) if you clicked several times too quickly. I have changed this sidebar to disable during the mutation to match the table, but if you think its preferrable to avoid disabling entirely I can change the buttons in the table instead.

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/ApplicationBookmarkFlag.tsx
tristan-orourke: See my question in https://github.com/GCTC-NTGC/gc-digital-talent/pull/17774#issuecomment-5329108975

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/ApplicationBookmarkFlag.tsx
tristan-orourke: Presentation and stateful layers rolled back together in https://github.com/GCTC-NTGC/gc-digital-talent/pull/17774/commits/c03ef6def811cf01b570f870b1a93a0b88bd1a07

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/ApplicationBookmarkFlag.tsx
tristan-orourke: See my question about we should avoid disabling buttons even while the mutation is in progress. https://github.com/GCTC-NTGC/gc-digital-talent/pull/17774#issuecomment-5329108975

If its okay to disable while mutation is in progress, I'll follow up with advice on how best to use aria-disabled.

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/ApplicationBookmarkFlag.tsx
tristan-orourke: Now using the Pending component to render a spinner instead https://github.com/GCTC-NTGC/gc-digital-talent/pull/17774/commits/c03ef6def811cf01b570f870b1a93a0b88bd1a07

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/ApplicationBookmarkFlag.tsx
```diff
@@ -93,6 +119,7 @@ const ApplicationBookmarkFlag = ({ query }: ApplicationBookmarkFlagProps) => {
       <Button
         {...commonProps}
         icon={isFlagged ? FlagIconSolid : FlagIconOutline}
+        disabled={isUpdatingFlag}
```
esizer: Disabling this button has an effect on focus manahgmenrt. When abutton becomes disaled, it gets removed from the tab index, forcing a users focus to the `body` element. This means a user must then tab all the way back to the button after toggling it.

I think we should avoid disabling buttons as a general rule so this does not occur.

esizer (changes requested):

tristan-orourke:
I ended up added optimistic updating of the flag/bookmark state, because it's so easy in this case. But I know that's not our usual pattern, so @esizer let me know if you think that is inappropriate, it would be easy to undo.

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/ApplicationBookmarkFlag.tsx
tristan-orourke: Resolved in https://github.com/GCTC-NTGC/gc-digital-talent/pull/17774/commits/c03ef6def811cf01b570f870b1a93a0b88bd1a07

apps/web/src/pages/PoolCandidates/ViewPoolCandidatePage/components/Sidebar/ApplicationBookmarkFlag.tsx
tristan-orourke: Disabled removed in https://github.com/GCTC-NTGC/gc-digital-talent/pull/17774/commits/c03ef6def811cf01b570f870b1a93a0b88bd1a07

esizer (approved):

PR #17785, opened 2026-08-18
[Refactor] Remove full graphql object types 11

esizer:
Related: #17227 

## 👋 Introduction

Removes full object types from the applicaitons and talent request sections.

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Login as any user
3. Confirm no regressions in submitting an application
4. Confirm no regression in submitting a talent request

### DIFF
--- a/apps/web/src/pages/Applications/profileStep/profileStepValidation.tsx
+++ b/apps/web/src/pages/Applications/profileStep/profileStepValidation.tsx
@@ -1,9 +1,9 @@
-import type {
-  Application_PoolCandidateFragment,
-  Pool,
-} from "@gc-digital-talent/graphql";
+import type { Application_PoolCandidateFragment } from "@gc-digital-talent/graphql";
 
-import type { ApplicationBrowserState } from "~/types/applicationStep";
+import type {
+  ApplicationBrowserState,
+  ApplicationStepPool,
+} from "~/types/applicationStep";
 import type {
   PartialUserAbout,
   PartialUserDei,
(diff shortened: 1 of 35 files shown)


apps/web/src/validators/profile/about.ts
```diff
@@ -1,30 +1,41 @@
...
-type PartialLanguage = Pick<LocalizedLanguage, "value"> | null;
+interface AboutPool {
+  areaOfSelection?: GenericLocalizedEnum<PoolAreaOfSelection> | null;
+}
+
+interface PartialLanguage {
+  value: Language;
+}
+
+interface PartialCitizenship {
```
petertgiles: The AI flagged that interfaces like these are repeated across the validators and could be replaced with something like `Pick<GenericLocalizedEnum<T>, "value">`.  Not blocking, of course.

apps/web/src/pages/Applications/profileStep/profileStepValidation.tsx
petertgiles, line 1: Probably pre-existing.  Just noting for myself: it seems like opening the language profile triggers a render loop.  Doesn't happen in prod but that might be because of error suppression or prod build differences.

https://github.com/user-attachments/assets/2433e552-908e-4823-b6fd-d99e798d4f03

petertgiles (approved):
✔️ Made a pool with general questions and screening questions
✔️ Submitted an application to that pool
✔️ Submitted a talent request

apps/web/src/pages/Applications/profileStep/profileStepValidation.tsx
petertgiles: https://github.com/GCTC-NTGC/gc-digital-talent/issues/17795

apps/web/src/validators/profile/about.ts
esizer: That is a good suggesiton, added in [add localized enum value helper type](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17785/commits/f64a2bf9c127a30f7f6e914e48bc9ef8e38f190e)

petertgiles: Nice, that got rid of a lot of manual typing.  :heart:

PR #17788, opened 2026-08-18
[Fix] CSP nonce hydration error

### LINKED ISSUE #17634: 🐛 Hydration mismatch on every page load from unsubstituted CSP nonce in client bundle

On every page load, the browser console logs a React hydration error: "A tree hydrated but some attributes of the server rendered HTML didn't match the client properties. This won't be patched up...". The mismatch is reported for every element in `Layout` ([apps/web/src/root.tsx](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/apps/web/src/root.tsx)) that carries a `nonce` prop: the inline theme-init `<script>`, the stylesheet `<link>`s rendered by `<Links>`, `<ScrollRestoration>`, `<Scripts>`, and the server-config `<script>`.

Expected: no hydration warnings on load — the nonce the client computes should match what was actually served in the HTML.

**🕵️ Details**

- Reproduced locally against `http://localhost:8000`, served through nginx per `infrastructure/conf/nginx-local.conf`.
- `root.tsx` and `entry.server.tsx` hardcode the literal placeholder string `"**CSP_NONCE**"` for every `nonce` prop. Per [infrastructure/conf/nginx-local.conf:245](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/infrastructure/conf/nginx-local.conf#L245) (and the prod equivalent, [nginx-deploy.conf:203](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/infrastructure/conf/nginx-deploy.conf#L203)), nginx's `sub_filter '**CSP_NONCE**' $cspNonce;` rewrites that placeholder to a real per-request nonce (`$cspNonce = $request_id`) — but only inside the `location /` block that serves the static `index.html` ([nginx-local.conf:241](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/infrastructure/conf/nginx-local.conf#L241)).
- The compiled client bundle (`/apps/web/dist/client/assets/*.js`) is served from a **different** `location` block ([nginx-local.conf:225](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/infrastructure/conf/nginx-local.conf#L225)) that has no `sub_filter`. That bundle still contains the literal `"**CSP_NONCE**"` string baked in at build time — this app doesn't do per-request SSR; `entry.server.tsx`'s `handleRequest` only runs once, at build time, per its own file comment.
- So on hydration, the client re-renders `Layout` and computes `nonce="**CSP_NONCE**"` (the literal), while the DOM already has the real, nginx-substituted nonce — hence the mismatch on every nonce'd element.
- `<html suppressHydrationWarning>` ([root.tsx:130](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/apps/web/src/root.tsx#L130)) doesn't help here — it only covers the element it's set on, not descendants — and react-router's `<Scripts>`/`<ScrollRestoration>` explicitly omit `suppressHydrationWarning` from their prop types, so it can't be passed through to them regardless.
- Not a CSP security issue (the real nonce is what's actually enforced), but it pollutes the console on every load and would mask genuine hydration bugs introduced later.

**📋 Steps to Reproduce**

1. Run the web app locally with `make -f Makefile.nix setup_web`, served behind nginx per `infrastructure/conf/nginx-local.conf`.
2. Open the site in a browser (e.g. `http://localhost:8000`) with devtools open.
3. Watch the console during initial page load.
4. Observe a React error — "A tree hydrated but some attributes of the server rendered HTML didn't match the client properties..." — showing repeated `nonce="**CSP_NONCE**"` (client) vs. `nonce=""` (server, as read back) diffs for the theme-init `<script>`, the `<Links>`-rendered stylesheet `<link>`s, and the `<Scripts>`/config `<script>` tags.

**📸 Screenshot**

_No response_

**🙋‍♀️ Proposed Solution**

Stop having the client recompute the placeholder literal; have it read back the real nonce nginx already substituted into the page.

1. Add `<meta name="csp-nonce" content="**CSP_NONCE**" />` in `Layout`'s `<head>` ([root.tsx](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/apps/web/src/root.tsx)). Because it's plain text, nginx's existing `sub_filter` substitutes it the same as every other occurrence of the placeholder — no nginx config changes needed. Using `content` rather than `nonce` matters here: browsers hide/clear the `nonce` **attribute** after use (anti-exfiltration), but `content` is unaffected and stays readable after load.
2. Add a small helper and use it for every `nonce` prop in `Layout`:
   ```tsx
   function getNonce(): string {
     // Build-time render (entry.server.tsx runs once, at build time, to
     // produce the static index.html — see its file comment). Emit the
     // placeholder; nginx's sub_filter swaps it for the real per-request
     // nonce in the HTML it serves.
     if (typeof document === "undefined") return "**CSP_NONCE**";

     // Hydration: the compiled client bundle never goes through nginx's
     // sub_filter, so read the real nonce back out of the DOM instead of
     // recomputing the placeholder.
     return (
       document.querySelector<HTMLMetaElement>('meta[name="csp-nonce"]')
         ?.content ?? ""
     );
   }
   ```
   Replace every `nonce="**CSP_NONCE**"` in `Layout` (theme-init `<script>`, `<Links>`, `<ScrollRestoration>`, `<Scripts>`, server-config `<script>`) with a single `const nonce = getNonce();` reused across all five.
3. Leave `entry.server.tsx`'s own `const nonce = "**CSP_NONCE**";` as-is — it only runs at build time, `document` is never defined there, and it should keep emitting the placeholder.

> [!NOTE]
> This isn't a `suppressHydrationWarning` situation — the client bundle genuinely cannot know the real nonce at build time, since nginx generates a fresh one (`$request_id`) per request. The fix has to make the client read the value back from the already-rendered document, not just silence the warning.

**✅ Acceptance Criteria**

- [ ] No hydration mismatch errors appear in the browser console on initial page load, in any environment where nginx performs the `**CSP_NONCE**` substitution.
- [ ] The `nonce` value used by client-rendered elements (`<Links>`, `<Scripts>`, `<ScrollRestoration>`, inline `<script>`s) matches the real nonce present in the served HTML/CSP header for that request.
- [ ] Inline scripts and stylesheets continue to load correctly under CSP (no new violations reported at `/api/csp-report`) both locally and in a deployed-like (nginx) environment.

**🔄 Regression:** No
**🔗 Root Cause Pull Request:** _No response_

esizer:
🤖 Resolves #17634 

## 👋 Introduction

Fixes a hydration error in react the CSP nonce placeholder value.

## 🕵️ Details

We had the placeholder hardcoded into both the built HTML and the JS files. When nginx sent the HTML with the substituted value, it did not match the placeholder value in the JS bundle when react was hydrating it which lead to the error.

This now keeps the placeholder since `document` is not defined in the build so it still contains the placeholder in the HTML for nginx to replace. However, it should no longet be in the JS bundle since it just has code that executes to get the nonce from the HTML. So, there is no longer anything to be mistmatched when react hydrates.

## 🧪 Testing

1. Run the dev sever `pnpm watch`
2. Navigate to any page
3. Confirm no hydration error appears in the console

### DIFF
--- a/apps/web/src/root.tsx
+++ b/apps/web/src/root.tsx
@@ -9,6 +9,7 @@ import "~/assets/css/tailwind.css";
 
 import type { Route } from "./+types/root";
 import RootErrorBoundary from "./components/Layout/RouteErrorBoundary/RootErrorBoundary";
+import { getDocumentNonce } from "./utils/csp";
 import { makeServerConfigJS } from "./utils/runtime";
 // eslint-disable-next-line import/extensions
 import initTheme from "./utils/initTheme.js?raw";
@@ -125,14 +126,13 @@ interface LayoutProps {
 }
 
 export function Layout({ children }: LayoutProps) {
+  const nonce = getDocumentNonce();
+
   return (
     // Suppress hydration warning coming from `initTheme.js` when running watch mode
     <html lang="en" suppressHydrationWarning>
       <head>
-        <script
-          nonce="**CSP_NONCE**"
-          dangerouslySetInnerHTML={{ __html: initTheme }}
-        />
+        <script nonce={nonce} dangerouslySetInnerHTML={{ __html: initTheme }} />
         <meta charSet="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1.0" />
         <meta name="msapplication-TileColor" content="#9747FF" />
@@ -144,13 +144,13 @@ export function Layout({ children }: LayoutProps) {
           content="$APP_URL/images/digital-talent/banner.jpg"
         />
         <Meta />
-        <Links nonce="**CSP_NONCE**" />
+        <Links nonce={nonce} />
       </head>
       <body className="bg-[#EEF1F9] font-sans text-black dark:bg-gray-700 dark:text-white">
         <div className="isolate">{children}</div>
 
         <ScrollRestoration
-          nonce="**CSP_NONCE**"
+          nonce={nonce}
           getKey={(location) => {
             // if there's a hash then use default behavior since the ScrollToLink will be handling scrolling
             if (location.hash) {
@@ -160,10 +160,10 @@ export function Layout({ children }: LayoutProps) {
             return location.pathname;
           }}
         />
-        <Scripts nonce="**CSP_NONCE**" />
+        <Scripts nonce={nonce} />
 
         <script
-          nonce="**CSP_NONCE**"
+          nonce={nonce}
           dangerouslySetInnerHTML={{
             __html: `
               const filterUnusable = (value) => !value.startsWith("$") && value.length > 0 ? value : undefined;
--- a/apps/web/src/utils/csp.test.ts
+++ b/apps/web/src/utils/csp.test.ts
@@ -1,6 +1,6 @@
 import { getNonce, setNonce } from "get-nonce";
 
-import { applyCspNonce } from "./csp";
+import { applyCspNonce, getDocumentNonce } from "./csp";
 
 /**
  * Limitation: jsdom has no CSP, so it never blanks the `nonce` content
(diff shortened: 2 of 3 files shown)


petertgiles (approved):
Looks great!  Thanks for the fix.

PR #17790, opened 2026-08-18
Merge experience rework feature branch

### LINKED ISSUE #15378: ♻️ Rework experience data models
### ♻️ Refactor

This ticket will likely function as an epic for refactors to all 5 experience types in varying degrees of transformation.

The goal of this ticket is to summarize changes to the experience models to bring them in line with more nuanced data needs. Often this means collecting more data around specific sub-types. This ticket also address the removal of the "Additional details" section on each experience in favor of more targeted `textarea` fields per experience type (an example of one of these changes is provided in the image below, but it varies by experience type). The Figma file linked here provides a side-by-side comparison of each experience model alongside migration path recommendations. Reach out to the design team for suggestions or questions.

Image

> [!NOTE]
> This ticket doesn't cover any major UI changes outside of the addition of the new fields and the rework of "Additional details" 

[Visit the Figma file](https://www.figma.com/board/8MCsCIOJ52c9DeHhD2wMWy/Experience-data-updates--Q4-2025-?node-id=0-1&t=zOS0TwYcC3jyt4VQ-1)

#### Work experience

Most of the effort on work experiences was rolled into the previous updates we did earlier in the year, but there are still a few outstanding changes to be made.

* A rework of how current experience and dates are collected
* The addition of "Management or supervisory status" to external positions
* Collecting the name of an employee's manager for internal positions
* Reworking "Additional details" into a subtler "Key tasks and responsibilities" field

#### Education and certificates

This type has the most drastic changes as we're hoping to collect more nuanced information around the type of education and specific details where needed.

* New types, each with their own options
  * New sub-types for degrees and fellowships
* Changes to "Completion status" and dates
* "Additional details" being moved from a full section into a subtler question

#### Community participation

Most a reordering of fields, but with a few label changes.

* A rework of how current experience and dates are collected
* Reworking "Additional details" into a subtler "Key tasks and responsibilities" field

#### Personal learning

Again, mostly a reordering of fields, but in this case, we're also looking to merge the two `textarea` questions in order to simplify the data.

* A rework of how current experience and dates are collected
* Merging "Experience description" and "Additional details" into a single "Learning description" field

#### Awards and recognition

Primarily label and order updates, with the addition of a new relationship to other experiences.

* The addition of a field that allows the user to link a different experience to the award, with the intent of providing further context to how/when/why the award or recognition was provided

### ✅ Acceptance Criteria

- [ ] Figma has been reviewed and questions have been discussed with the design team
- [ ] [✨ Experience rework - Work #15432](https://github.com/GCTC-NTGC/gc-digital-talent/issues/15432)
- [ ] [✨ Experience rework - Education #15433](https://github.com/GCTC-NTGC/gc-digital-talent/issues/15433)
- [ ] [✨ Experience rework - Community #15434](https://github.com/GCTC-NTGC/gc-digital-talent/issues/15434)
- [ ] [✨ Experience rework - Personal #15435](https://github.com/GCTC-NTGC/gc-digital-talent/issues/15435)
- [ ] [✨ Experience rework - Award #15436](https://github.com/GCTC-NTGC/gc-digital-talent/issues/15436)
- [ ] Narrow and wide screens have been reviewed
- [ ] Light and dark modes have been reviewed
- [ ] Accessibility has been manually reviewed
- [ ] Design has reviewed the final elements and page

JamesHuf:
🤖 Resolves #15378

## 👋 Introduction

Preps the experience rework feature branch to be merged.

## 🧪 Testing

Test all things experience related:
- Creating/updating experiences
- Submitting applications with both new and old experiences
- Viewing experiences as the user 
- Viewing experiences as an admin
- Viewing old and new snapshots
- Downloading profiles
- Downloading snapshots


esizer (approved):
LGTM! 👍 Not sure if we want a secondary review or not? Its a lot of change but we were reviewing each unit so maybe it is not necessary 🤔

JamesHuf:
I was wondering about that as well

api/app/Generators/ApplicationDocGenerator.php
```diff
@@ -30,6 +31,10 @@ public function __construct(protected PoolCandidate $candidate, public ?string $
 
         parent::__construct($fileName, $dir);
         $this->anonymous = false;
+
+        $this->snapshotVersion = $candidate->profile_snapshot['version'] ?
+            $candidate->profile_snapshot['version'] :
+            1;
```
vd1992: **Critical — older applications can't be downloaded**

This reads `['version']` straight off the array. If the key isn't there, PHP raises a warning, and Laravel turns warnings into exceptions. Snapshot versioning was added in #11433 (Sept 2024) and nothing backfilled the older rows, so any snapshot written before then has no `version` key — going back to when the column was added in Aug 2022.

What each caller does with the exception:

- **Single download** — `DownloadApplicationDoc` catches it, logs, and returns null, so the download fails with nothing pointing at the cause.
- **Bulk zip** — the candidate's document is replaced with a generic "unknown error" entry.
- **Excel export** — the same access sits inside `chunk(200)`, so one old candidate fails the whole export.

There's a second effect. This runs in the constructor, before the `MissingProfileSnapshotException` check in `generate()`. So a candidate with no snapshot at all now throws "Trying to access array offset on null" instead of the friendly error.

```suggestion
        $this->snapshotVersion = $candidate->profile_snapshot['version'] ?? 1;
```

Same fix needed at `PoolCandidateExcelGenerator.php:163`.

api/app/Http/Resources/AwardExperienceResource.php
```diff
@@ -23,6 +26,17 @@ class AwardExperienceResource extends JsonResource
      */
     public function toArray($request)
     {
+
+        $relatedExperienceResource = (bool) $this->relatedExperience ? match (get_class($this->relatedExperience)) {
+            WorkExperience::class => new WorkExperienceResource($this->relatedExperience),
```
vd1992, line 32: **Critical — awards linked to a personal experience lose the link in the snapshot**

This arm matches `PersonalExperienceResource::class`, but `get_class()` returns the model, `App\Models\PersonalExperience`. So it never matches and falls through to `default => null`. The other three arms use the model class and work.

If an applicant links an award to a personal experience, the snapshot stores `relatedExperience: null`. Their live profile still shows the link, because that reads through `@morphTo` rather than this resource. So the applicant sees the link and the assessor reading the snapshot doesn't. `ProfileSnapshot::set` returns the existing value when one is already stored, so a snapshot saved this way can't be regenerated afterwards.

`PersonalExperience` is also the only one of the four models not imported at the top of this file, which is probably how it happened.

api/app/Models/EducationExperience.php
```diff
@@ -38,21 +66,77 @@ class EducationExperience extends Experience
...
+                        ? $this->other_fellowship_type
+                        : LanguageHelpers::localizeEnum($this->fellowship_type, FellowshipType::class, $lang);
+                    break;
+                case EducationType::OTHER->name:
+                    $type = $this->other_education_type ?? Lang::get('headings.other_type_of_education', [], $lang);
+                    break;
+                default:
+                    $type = LanguageHelpers::localizeEnum($this->education_type, EducationType::class, $lang);
+            }
+            if ($this->education_type === EducationType::PROFESSIONAL_CERTIFICATION) {
```
vd1992: **High — certification and licence names are missing from downloaded documents**

`education_type` is a plain string column, so comparing it to an enum instance is always false. Neither this branch nor the `LICENSE_ACCREDITATION` one below ever runs. The `switch` just above does the same comparison correctly, with `->name`.

Since neither runs, `$subject` stays as `area_of_study`. The new validator prohibits `areaOfStudy` for these two types, so it's null, and the title comes out as "Professional certification from «Institution»" with no certification name in it.

Two reasons this is easy to miss when testing: migrated rows still hold a copy of the name in `area_of_study` so they look correct, and nothing maps to `LICENSE_ACCREDITATION` in the migration, so every licence record will be newly created.

api/app/Models/EducationExperience.php
```diff
@@ -38,21 +66,77 @@ class EducationExperience extends Experience
...
+        return $this->morphMany(AwardExperience::class, 'related_experience');
+    }
+
+    public function getTitle(?string $lang = 'en', ?int $experienceVersion = null): string
+    {
+        $type = '';
+        $subject = $this->area_of_study;
+        if ((bool) $experienceVersion && $experienceVersion === 1) {
+            // V1 Snapshot representation of an Education Experience
+            $type = LanguageHelpers::localizeEnum($this->type, DegreeType::class, $lang);
```
vd1992: **High — French documents show English education types for older snapshots**

V1 snapshots store the old `EducationType` values, but this localises them against `DegreeType`. Only three values exist in both lists. The rest don't match, so `localizeEnum` falls back to `sanitizeEnum`, which just title-cases the raw value in English:

```
BACHELORS_DEGREE          FR: Baccalauréat à …
DIPLOMA                   FR: Diploma à …
POST_DOCTORAL_FELLOWSHIP  FR: Post Doctoral Fellowship à …
ONLINE_COURSE             FR: Online Course à …
CERTIFICATION             FR: Certification à …
OTHER                     FR: Other à …
```

Because three of the eight do translate correctly, a spot check can easily land on a working one.

This PR replaced the old `education_type` lang keys, so there's nowhere left to get the V1 French labels from. Probably needs a decision on whether to keep the old keys or map the V1 values onto the new enums.

apps/web/src/components/ApplicationSnapshot/ApplicationSnapshot.tsx
```diff
@@ -332,7 +332,10 @@ const ApplicationSnapshot = ({
         </Accordion.Item>
       </Accordion.Root>
 
-      <CareerTimelineSnapshot experiences={experiences} />
+      <CareerTimelineSnapshot
+        experiences={experiences}
+        snapshotVersion={snapshot.version}
```
vd1992, line 339: **High — the skills sections show "Not available" for education on older snapshots**

The version is passed to `CareerTimelineSnapshot` here, but the same `experiences` array goes to `TechnicalEssentialSkillsSnapshot` and `TechnicalAssetSkillsSnapshot` further up without it. Both of those render `SkillSnapshot`, which always uses the new `ExperienceCard`.

So on a V1 snapshot, two sections of the same page disagree:

```
career timeline (gated)   -> Other type of education in Brattice Builder from Schiller, Kovacek and Koss
skills sections (ungated) -> Not available in Brattice Builder from Schiller, Kovacek and Koss
```

Only education is affected — award, community, personal and work names come out the same either way. And since `SkillSnapshot` passes `hideDetails`, it's just the heading, not the card body.

`SkillSnapshot` isn't in this diff, which is probably why it was missed.

api/app/Traits/Generator/GeneratesUserDoc.php
```diff
@@ -411,28 +428,15 @@ public function experience(Section $section, AwardExperience|CommunityExperience
...
-
-            $title = implode(' ', $titleComponents);
+            $title = $experience->getTitle($this->lang, $experienceVersion);
             $section->addTitle($title, $headingRank);
             $section->addText($experience->getDateRange($this->lang));
             $this->addLabelText($section, $this->localize('headings.area_of_study'), $experience->area_of_study);
-            $this->addLabelText($section, $this->localize('common.status'), $this->localizeEnum($experience->status, EducationStatus::class));
+            $this->addLabelText($section, $this->localize('headings.completion_status'), $this->localizeEnum($experience->status, EducationStatus::class));
             $this->addLabelText($section, $this->localize('headings.thesis_title'), $experience->thesis_title);
+            $this->addLabelText($section, $this->localize('headings.license_accreditation'), $experience->license_or_accreditation);
```
vd1992: **Medium — documents print empty labelled lines**

`addLabelText` always writes the label and colon, even when the value is null. Each education type uses at most one of these three fields, so the other two come out as empty lines. A newly created bachelor's degree gets four blank ones — thesis, licence, certification and course name — since the validator prohibits a thesis title at that degree level. Migrated rows kept their thesis titles, so they show three.

It's already visible in the committed snapshot:

```
Thesis or dissertation: 
License or accreditation: 
Certification: 
Course name: et
```

Wrapping these in a check on `education_type` would match how `getTitle` already handles it.

api/database/factories/EducationExperienceFactory.php
```diff
@@ -25,35 +29,67 @@ class EducationExperienceFactory extends Factory
...
+                            ? $this->faker->dateTimeBetween($attributes['start_date'], '2019-12-31')?->format('Y-m-d')
+                            : null,
             'details' => $this->faker->text(),
+            'education_type' => $this->faker->randomElement(EducationType::cases())->name,
+            'status' => fn ($attributes) => (
+                $attributes['education_type'] === EducationType::LICENSE_ACCREDITATION ||
+                $attributes['education_type'] === EducationType::PROFESSIONAL_CERTIFICATION
+            )
+                ? $this->faker->randomElement($generalStatuses)->name
+                : $this->faker->randomElement($licenseOrCertificationStatuses)->name,
```
vd1992: **Medium — the factory generates statuses the form can't show**

Two things are wrong. `$attributes['education_type']` is a string (set with `->name` on the line above), so comparing it to an enum instance is always false. And the branches are the wrong way round — as written, licence and certification would get `$generalStatuses`.

Because the condition never matches, everything falls to `$licenseOrCertificationStatuses`. So `SUCCESS_CREDENTIAL` and `SUCCESS_NO_CREDENTIAL` are never generated at all, and degrees get `SUCCESS`, which `getEducationStatusOptions` only offers for licences and certifications.

Confirmed on a degree carrying `SUCCESS`: the experience card renders it as "Completed" without any trouble, so nothing looks wrong until you open the record for editing. There the completion status radio group comes up with **nothing selected**, on a field marked required — it offers In progress / Completed with a credential / Completed without a credential / Did not complete, and the stored value isn't among them. Any edit to that experience, for any reason, forces the user to re-answer the status, and whatever they pick replaces the original value.

Worth fixing both parts together — fixing only the comparison would activate the swapped branches and make it worse.

api/app/GraphQL/Validators/AwardExperienceInputValidator.php
```diff
@@ -0,0 +1,69 @@
...
+        return [
+            'title' => ['sometimes', 'required', 'string'],
+            'issuedBy' => ['sometimes', 'required', 'string'],
+            'awardedDate' => ['sometimes', 'required', 'date', 'before_or_equal:today'],
+            'awardedTo' => ['sometimes', 'required', Rule::in(array_column(AwardedTo::cases(), 'name'))],
+            'awardedScope' => ['sometimes', 'required', Rule::in(array_column(AwardedScope::cases(), 'name'))],
+            'projectName' => [
+                'nullable',
+                'string',
+                Rule::requiredIf(
```
vd1992, line 44: **Medium — `projectName` is never actually required**

`awarded_to` is the database column name; the GraphQL argument is `awardedTo`. `Validator::arg()` looks up the argument name, so this is always null and the rule never fires. It's the only snake_case `arg()` call in the validators directory.

The form requires the field client-side, so nothing is visibly broken today — it's the server-side backstop that isn't working.

api/app/GraphQL/Validators/EducationExperienceInputValidator.php
```diff
@@ -0,0 +1,198 @@
...
+                    [
+                        'before_or_equal:today',
+                    ]
+                ),
+            ],
+            'endDate' => [
+                'sometimes',
+                'nullable',
+                'date',
+                'after_or_equal:'.$this->arg('startDate'),
```
vd1992, line 158: **Medium — this rule rejects everything when `startDate` is missing**

If `startDate` isn't sent, this builds `after_or_equal:` with nothing after the colon. That doesn't skip the check — it fails every value:

```
Validator::make(['endDate' => '2020-05-05'], ['endDate' => ['nullable','date','after_or_equal:']])
  → fails: "The end date field must be a date after or equal to ."
```

Through the UI an end date only ever has a value when a start date does too, so you'd need to hit the API directly to reach this. But that message is user-facing if it ever fires. Line 179 does the same thing for `prospectiveEndDate`.

This pattern is already on main in a few places — `ChangePoolClosingDateValidator` interpolates `$pool?->closing_date`, which can also be null — so a shared helper might beat fixing this one spot.

apps/web/src/components/ExperienceFormFields/AwardFields.tsx
```diff
@@ -55,120 +130,172 @@ const AwardFields = ({
...
+    // Reset the project name if awarded to changes
+    if (watchAwardedTo !== AwardedTo.MyProject) {
+      resetField("projectName", { keepDirty: false, defaultValue: null });
+    }
+
+    // Set the related experience type
+    const relatedExperienceType = myExperiences.find(
+      (experience) => experience.id === watchRelatedExperienceId,
+    );
+    if (relatedExperienceType) {
```
vd1992: **Medium — a related experience can't be un-set once chosen**

Correcting my original comment: I said clearing the selection was a normal thing to do. It isn't possible at all. `Select` renders its null option as `disabled={!enableNull}`, and this call site doesn't pass `enableNull`, so "Select an option" is an inert placeholder. The field is optional going in but permanent afterwards — you can swap it for a different experience, never back to none.

That also narrows the original point. The effect here only ever calls `setValue` and never clears, so `relatedExperienceType` keeps its previous value when `relatedExperienceId` goes empty. But since the UI can't produce an empty selection, that inconsistent row — type populated, id null — is only reachable through the API. Nothing rejects it there either; the server-side `required_with` guards the opposite direction.

Worth fixing together: pass `enableNull` so the field can be cleared, **and** add an `else` that clears `relatedExperienceType`. Doing only the first would start writing the inconsistent row through the UI.

On the label problem in the screenshot on this thread — `AwardOptions_Query` selects `educationType { value label { localized } }`, but `getExperienceName` hands that label to `getLocalizedName`, which reads `name?.[locale]`, i.e. `label.en` / `label.fr`. Neither is fetched, so it returns the not-available string. `degreeType` has the same shape. The "Other type of education" row renders correctly because that branch uses `otherEducationType` and never touches the label, which is what pins the cause to the label selection rather than the name logic.

api/tests/Unit/HydrationTest.php
```diff
@@ -99,7 +99,7 @@ public function testHydrateRawEnums(): void
             [
                 {
                     "id": "00000000-0000-0000-0000-000000000000",
-                    "type": "CERTIFICATION",
+                    "educationType": "CERTIFICATION",
```
vd1992: **Medium — the tests that covered V1 snapshots were converted to V2**

These two fixtures were the only tests using a V1 education payload — one with a raw string enum, one with a localised object. Both were renamed to `educationType`, so the `'type' => 'type', // preserved for snapshot version 1` mapping is no longer tested anywhere. This assertion also expects `CERTIFICATION`, which isn't a value in the new `EducationType`.

Grouping the related points here rather than leaving four separate comments, because it's one pattern: the fixtures were all updated to V2 at the same time as the code, so nothing is left that can see the compatibility layer.

- The V1 frontend code — 907 lines under `SnapshotV1/`, plus the V1 branches in `getTitle` and `GeneratesUserDoc` — isn't referenced by any test, story or spec in the repo.
- `AwardExperienceFactory` never sets `related_experience_*`, so `SnapshotTest` never compares that field, even though the reference query selects it (including the `PersonalExperience` arm). The seeder, fake data and e2e fixtures don't set it either.
- Both snapshot stories build a user with no `version` key, so Chromatic renders V1 components against V2 fake data. That's the "Not available" in the ScreeningDecisionDialog diff — and it means Chromatic is showing the one situation where the ungated skills sections happen to look right.
- `fakeExperiences.generateEducation` gained `educationType` but not `degreeType`, `certification`, `courseName`, `fellowshipType` or `licenseOrAccreditation`, so the stories render education states that can't exist.

Best value for effort, roughly in order: add a `relatedExperience` state to `AwardExperienceFactory` — the existing snapshot test would then catch the resource bug on its own; keep a V1 payload here alongside the V2 one; and set `version: 2` on the story fixtures, plus one `version: 1` story with V1-shaped experiences.

vd1992 (changes requested):
Went through the full diff, ran the migrations against a local database to check the data changes on real rows, and then manually tested the findings in the running app.

18 findings across 13 comments — one of them groups five related points rather than repeating itself. Two are critical, five high, eleven medium. The two critical ones both touch data that either already exists or can't be recovered once written, so those are the ones I'd want resolved before this merges. The rest can be triaged.

**Ten are confirmed by manual testing**, including both criticals — screenshots are on the individual threads. Three of the findings only surfaced during that testing rather than from reading the diff.

A few things that don't fit in a line comment:

**Green CI doesn't contradict any of this.** Every test builds a fresh snapshot, so nothing exercises the older ones. The application document fixture has a single education experience, and it's the one type where everything works. `UserDocGeneratorTest` has no award experiences at all. And the snapshot files were updated to match the new output, so the blank lines sit in the committed fixture rather than failing an assertion.

**Plenty holds up.** The education form's reset logic covers every field the new validator prohibits. The migration mapping is correct on all 12,088 education rows and round-trips losslessly. The personal experience branch in the document generator handles both snapshot versions properly — a good model for what the education branch could do.

Neither critical issue shows up when working with freshly created data, which is why both got past review and CI.

vd1992:
This PR is massive so I leaned on the bot more to do the reviewing

apps/web/src/components/ApplicationSnapshot/ApplicationSnapshot.tsx
vd1992: Verified

On view candidate application page
Observe:

image

image

api/app/Models/EducationExperience.php
vd1992: Verified
French application download, observe "Online course", screenshot

image

api/app/Traits/Generator/GeneratesUserDoc.php
vd1992: Verified.
Rendering stuff with no values attached. See screenshot (reused it)

image

api/app/Http/Resources/AwardExperienceResource.php
vd1992: Verified
#### Application view
image

#### Logged in as user, viewing timeline

image

api/app/Traits/Generator/GeneratesUserDoc.php
```diff
@@ -354,16 +359,28 @@ function ($a, $b) {
...
-            $section->addTitle($experience->getTitle(), $headingRank);
+            $section->addTitle($experience->getTitle($this->lang), $headingRank);
             $section->addText($experience->getDateRange($this->lang));
             $this->addLabelText($section, $this->localize('headings.awarded_to'), $this->localizeEnum($experience->awarded_to, AwardedTo::class));
+            if ($experience->project_name) {
+                $this->addLabelText($section, $this->localize('headings.project_name'), $experience->project_name);
+            }
             $this->addLabelText($section, $this->localize('headings.issuing_organization'), $experience->issued_by);
             $this->addLabelText($section, $this->localize('headings.awarded_scope'), $this->localizeEnum($experience->awarded_scope, AwardedScope::class));
+            if ($experience->relatedExperience) {
```
vd1992, line 383: **High — profile document download fails for any user with more than one award experience**

`$experience->relatedExperience` is lazy-loaded here, and `DownloadUserDoc` fetches the user with a bare `User::findOrFail()`. With `Model::shouldBeStrict(! $this->app->isProduction())` in `AppServiceProvider`, that throws:

```
LazyLoadingViolationException: Attempted to lazy load [relatedExperience]
on model [App\Models\AwardExperience] but lazy loading is disabled.
```

`DownloadUserDoc` catches it and returns null, so the user gets the generic "There was a problem downloading your file" message with nothing pointing at the cause, and the real error only appears in the logs.

**The trigger is the number of awards, not whether any is linked.** Laravel only arms the guard when a query returns multiple rows — `Builder::hydrate()`:

```php
if (count($items) > 1) {
    $model->preventsLazyLoading = Model::preventsLazyLoading();
}
```

So a user with a single award never trips it, and a user with two or more does, whether or not any of them has a related experience. Confirmed by adding two unlinked awards to a seeded user who had one (started throwing) and removing them again (stopped).

That means this doesn't depend on the new related-experience feature being used — it affects any applicant with two or more awards. Seeded data gives 601 of 602 users exactly one award, which is why it doesn't show up locally by default. Production runs with strict mode off, so there it lazy-loads instead: one extra query per award rather than a failure.

`GeneratesCareerExperienceSheet.php:342` already does this correctly with `->with([... 'relatedExperience'])`.

`UserDocGeneratorTest` has no award experiences in its fixture — the committed snapshot contains a single work experience — which is why this path isn't covered.

api/app/Models/AwardExperience.php
```diff
@@ -42,9 +44,15 @@ class AwardExperience extends Experience
         'awarded_date' => 'awardedDate',
         'awarded_to' => 'awardedTo',
         'awarded_scope' => 'awardedScope',
+        'project_name' => 'projectName',
```
vd1992, line 47: **Medium — application documents never show an award's related experience**

`project_name` was added to this map but the morph columns weren't, so `hydrateSnapshot` drops the relation. Even when the snapshot stored it correctly, the hydrated model comes back empty:

```
                              snapshot JSON              hydrated model
Client Service Commendation   WorkExperience present  →  null
Innovation Prize              WorkExperience present  →  null
```

`GeneratesUserDoc.php:381` then finds nothing and omits the "Related experience" line, so the field never appears in an application document. That line behaves differently depending on the path: application documents render hydrated models, which come back with the relation empty, while profile documents render models straight from the database, which is where the lazy-loading failure on that same line applies instead.

Worth noting the fix isn't two more entries: the snapshot stores `relatedExperience` as a nested object rather than ids, and `hydrateFields` handles scalars and localized enums, not nested models. `PoolCandidateExcelGenerator` already pre-processes nested objects into ids before hydrating work experiences — same shape of problem.

api/app/Traits/Generator/GeneratesUserDoc.php
vd1992: Verified
Admin pages -> user profile -> download profile failed

image

image

api/app/Models/EducationExperience.php
vd1992: Verified

#### Applicant view
image

#### Admin download of user profile
image

api/app/Generators/ApplicationDocGenerator.php
vd1992: Verified
Attempting to download full spreadsheet datasheet for a pool with such a candidate fails
Nothing gets logged for this failure though

image

apps/web/src/components/ExperienceFormFields/AwardFields.tsx
vd1992: Uh, something else verified as well. 
I can't seem to unlink the related experience. Going to the edit page, I cannot see a way to null out the input. 

_**As well**_

Necessary fields not fetched to render names in the input

image

api/database/factories/EducationExperienceFactory.php
vd1992: Verified
In edit mode no option is selected

image

image

api/app/Generators/ApplicationDocGenerator.php
JamesHuf: Should be fixed by [94e5e90](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17790/commits/94e5e901ec9168433385d18f4780567afe1902a0).

api/app/Http/Resources/AwardExperienceResource.php
JamesHuf: Oops! Fixed in [3411582](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17790/commits/3411582008538fcb7d15122a3ea1a021f7077d2f).

api/app/Models/EducationExperience.php
JamesHuf: Fixed in [35221a4](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17790/commits/35221a429086c3bb4ac2ed83f1742965b1dfd43b).

api/app/Models/EducationExperience.php
JamesHuf: I was hoping this could stay gone but fixed in [3f84930](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17790/commits/3f8493065c9f4ed4f2c2f910037ef4b367075329).

apps/web/src/components/ApplicationSnapshot/ApplicationSnapshot.tsx
JamesHuf: Fixed in [963b296](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17790/commits/963b2968acba77bef9d5b63a727765ff231db8b4).

api/app/Traits/Generator/GeneratesUserDoc.php
JamesHuf: I updated this in [6edb014](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17790/commits/6edb014ad72d40644a26029d177b757a8b537dbe) to only show fields when relevant - I kept them in when the field was optional and empty though.

api/database/factories/EducationExperienceFactory.php
JamesHuf: This should be fixed in [6859d36](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17790/commits/6859d365eaec463b5e337d02d9cc2b3a737c2d96).

api/app/GraphQL/Validators/AwardExperienceInputValidator.php
JamesHuf: Fixed in [d91e886](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17790/commits/d91e886b3ea0063372fb3bed7b5dd0d9787fa862).

api/app/GraphQL/Validators/EducationExperienceInputValidator.php
JamesHuf: I don't think this makes sense since it's only a potential problem when using the API directly to set a state that **should** be invalid.

apps/web/src/components/ExperienceFormFields/AwardFields.tsx
JamesHuf: Fixed in [1a58fcf](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17790/commits/1a58fcf4db45586ee3e086537cf77c803ebdd207) and [8bc42fd](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17790/commits/8bc42fd911a02c119b7a693e86205875f6601f0a).

api/tests/Unit/HydrationTest.php
JamesHuf: Ok, I updated several places to consistently be accurate for V2 snapshots - I'm not sure if it's worth it for V1 snapshots which **should** remain the same.
Relevant commits: [ba5c635](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17790/commits/ba5c63595fe3cf3959ca379ad262bf77557c4eb6), [f93f0c8](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17790/commits/f93f0c8324a80c74acc103cd988fadb72fc187df), [2534adf](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17790/commits/2534adf23420ce09653c5495baefbe4187062872), [a1cb067](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17790/commits/a1cb0676e75bb5df8b30ce6f62f37a10cda5e8ca).

api/app/Traits/Generator/GeneratesUserDoc.php
JamesHuf: Fixed in [633c51b](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17790/commits/633c51b92558b61d92d9bdb5f5966363bcac9d7e).

api/app/Models/AwardExperience.php
JamesHuf: Fixed in [113040b](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17790/commits/113040b7f50fa38ecf6342129f07bb03f0b5ac39).

JamesHuf:
~~I'm not sure why the fake experience generator wants to generate duplicate experiences suddenly.~~
Figured it out. Should be ready for re-review now.

api/app/Models/Experience.php
vd1992, line 1: Oh, one more potential thing, I think one or two fields may not be getting plucked from the snapshot. Only `id` and `__typename` are being used from the snapshot. The rest refetched from live data.
On top of the redundancy you have to track in two places

Snapshot experience handling duplicated in
`api/app/Generators/ApplicationDocGenerator.php`
`api/app/Generators/PoolCandidateExcelGenerator.php`

In this file  `api/app/Models/Experience.php`
Might as well handle it together, important since it would be used the same way in multiple files and easier to make sure one doesn't drift and become bugged. This would entail deleting them from their current location

Plop down the related model handling, sample diff to illustrate the intent of what you are moving here from the other two files
Call `setRelation()` allows overriding existed relations and is new

This would be around lines 209-210

``` diff
+use Illuminate\Support\Arr;
 use Illuminate\Support\Carbon;
@@
-                $model = new $hydrationModel();
-                $experiences[] = self::hydrateFields($experience, $fields, $model);
+                // the snapshot nests these as objects; $hydrationFields expects ids
+                if (isset($experience['department'])) {
+                    $experience['departmentId'] = $experience['department']['id'];
+                }
+                if (isset($experience['classification'])) {
+                    $experience['classificationId'] = $experience['classification']['id'];
+                }
+                if (isset($experience['workStreams'])) {
+                    $experience['workStreamIds'] = Arr::map($experience['workStreams'], fn ($value) => $value['id']);
+                }
+
+                $model = self::hydrateFields($experience, $fields, new $hydrationModel());
+
+                // an award's related experience renders from the snapshot, not the live row
+                if (! empty($experience['relatedExperience'])) {
+                    $model->setRelation('relatedExperience', self::hydrateSnapshot([$experience['relatedExperience']])[0] ?? null);
+                }
+
+                $experiences[] = $model;
```

JamesHuf: Nice fix! Added in [192531f](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17790/commits/192531fa1105197cce74992319d36f4c648ab18b).

api/app/Generators/ApplicationDocGenerator.php
vd1992: Critical issue resolved - manually tested

api/app/Http/Resources/AwardExperienceResource.php
vd1992: Critical issue resolved - manually tested

api/app/Models/EducationExperience.php
vd1992: Resolved 

image

api/app/Models/AwardExperience.php
vd1992: Fixed

image

api/app/Traits/Generator/GeneratesUserDoc.php
vd1992: Yep, can download now

api/app/Models/EducationExperience.php
vd1992: Good, good 

image

apps/web/src/components/ApplicationSnapshot/ApplicationSnapshot.tsx
vd1992: Nice

image

api/database/factories/EducationExperienceFactory.php
vd1992: Confirmed 

image

apps/web/src/components/ExperienceFormFields/AwardFields.tsx
vd1992: Both confirmed fixed, thanks!

image

api/app/Traits/Generator/GeneratesUserDoc.php
vd1992: Was tested in other comment chains, fixed

api/app/Models/Experience.php
vd1992: Tested manually, appears to work great

vd1992 (approved):
Okay, okay. I don't know if @esizer wants to double check since frankly there has been a lot of work since that approval.

But, after following up on just about every thread, all the concerns of note appear addressed. 
Fantastic!
This one was a little ridiculous in work involved, great job

PR #17822, opened 2026-08-20
[Refactor] Remove full graphql object types 13

esizer:
Related: #17227 

## 👋 Introduction

Removes full graphql object types form utils and validators.

## 🧪 Testing

I'm actually not entirely sure the best way to test this. These utils are used in a few dfifferent places.

This should ony changes types, so a type check should work. Or, go to the places the utils and validators are used and confirm no in browser errors?


apps/web/src/utils/poolUtils.tsx
```diff
@@ -150,9 +156,13 @@ interface PoolTitleOptions {
   short?: boolean;
 }
 
-type PartialPool = Pick<Pool, "name" | "publishingGroup" | "workStream">;
+interface PartialPool {
+  name?: LocalizedString | null;
+  publishingGroup?: GenericLocalizedEnum<PublishingGroup> | null;
```
vd1992: This is a little hard to follow due to all the chaining. But `PartialPool.publishingGroup` doesn't appear to actually be used for something?

Following the chain to `useAdminPoolPages`
Call
```
  const poolName = getFullPoolTitleLabel(intl, {
    workStream: pool.workStream,
    name: pool.name,
    publishingGroup: pool.publishingGroup,
    classification: pool.classification,
  });
```
Appears to use it
But continuing to follow the function chain to function `poolTitle` and field `publishingGroup` remains unused. 
Is this missing from name generation or can it just be dropped from interface `PartialPool` ?

apps/web/src/utils/experienceUtils.tsx
```diff
@@ -386,12 +380,12 @@ export const getExperienceFormLabels = (
...
  * @param type  ExperienceType
  * @param data  ExperienceFormValues<AllExperienceFormValues>
- * @param hiddenSkills Skill[] | null | undefined
+ * @param hiddenSkills AddedSkill[] | null
  * @returns ExperienceDetailsSubmissionData
  */
 export const formValuesToSubmitData = (
   data: ExperienceFormValues<AllExperienceFormValues>,
-  hiddenSkills: Skill[] | null,
+  hiddenSkills: AddedSkill[] | null,
```
vd1992: When is
`hiddenSkills`
used?
I can't seem to find an instance of this function being called with the input for that argument not being empty

apps/web/src/utils/skillUtils.ts
```diff
@@ -2,16 +2,32 @@ import uniqBy from "lodash/uniqBy";
...
   PoolSkillType,
 } from "@gc-digital-talent/graphql";
 import { SkillLevel, SkillCategory } from "@gc-digital-talent/graphql";
+import type { GenericLocalizedEnum } from "@gc-digital-talent/i18n";
 
 import type { SimpleAnyExperience } from "./experienceUtils";
 
+export interface SkillOption {
+  id: string;
+  key: string;
```
vd1992, line 13: The `SkillOption.key` doesn't appear to be used for anything? If it isn't used for anything, let us get rid of it rather than forcing it to be fetched and used

vd1992 (changes requested):
Nothing too significant, mostly minor

apps/web/src/utils/poolUtils.tsx
esizer: i'm not sure, this is my best attempt at a simple translation only. I have been trying to avoid refactoring as much as possible :sweat_smile:

apps/web/src/utils/experienceUtils.tsx
esizer: You may be correct, though, I have not really been checking to make sure these are used and just translating the types directly. Maybe this is a good opportunity to remove some unuser args? :thinking:

apps/web/src/utils/poolUtils.tsx
esizer: okay, very much required I guess :stuck_out_tongue: Removing it lead to:

```sh
Found 20 errors in 16 files
```

apps/web/src/utils/skillUtils.ts
esizer: I think right now it is required (`Found 4 errors in 2 files.`). Though, eventually, it might not be...

This is making me wonder how many other things we have that we no longer need though :sweat_smile:

apps/web/src/utils/poolUtils.tsx
vd1992: The function doesn't seem to use it though, so wouldn't it just be a whole bunch of calls passing in an argument that is never needed?
But I guess this isn't a big problem since it isn't a big field to fetch or pass in

esizer: Well, I have actualy been hoping that we can remove this entire function. We now have the `displayName` attrbiute on `Pool` which we could use instead. But, I think removing this type would lead to a little more than what the goal of this PR is.

Ideally, we never even need this function (in the near future :crossed_fingers:)

https://github.com/GCTC-NTGC/gc-digital-talent/blob/b8fb0a3de4673072dcdd88c7e455197cd063187d/api/app/Models/Pool.php#L456-L504

apps/web/src/utils/experienceUtils.tsx
esizer: Removed in [remove unused arg](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17822/commits/21e10020d5e6ba18fcea33f384334931cc00ac3d)

apps/web/src/utils/skillUtils.ts
```diff
@@ -2,16 +2,32 @@ import uniqBy from "lodash/uniqBy";
...
-  Skill,
-  PoolSkill,
   PoolSkillType,
 } from "@gc-digital-talent/graphql";
 import { SkillLevel, SkillCategory } from "@gc-digital-talent/graphql";
+import type { GenericLocalizedEnum } from "@gc-digital-talent/i18n";
 
 import type { SimpleAnyExperience } from "./experienceUtils";
 
+export interface SkillOption {
```
vd1992, line 11: This may need nullable/optional
`description`

Page `ApplicationSkills` 
Calls
```
  const categorizedEssentialSkills = categorizeSkill(
    poolSkillsByType.get(PoolSkillType.Essential),
  );
  const categorizedOptionalSkills = categorizeSkill(
    poolSkillsByType.get(PoolSkillType.Nonessential),
  );
```

The page renders skill description which is present in `application.pool.poolSkills` that is directly used in `groupPoolSkillByType`

`categorizeSkill` of input and return `SkillOption` makes no mention of `description`
But since it would be nullable and optional, would it even be useful to add to `SkillOption` ?

apps/web/src/utils/skillUtils.ts
vd1992: I believe that is due to
`<SkillTree />`
With prop
`skill` of type `Skill` from `"@gc-digital-talent/graphql"`

Which isn't until PR 15 of this epic though
So LOL

apps/web/src/utils/poolUtils.tsx
esizer: Removed in [remove publishing group from pool title helper](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17822/commits/9ab48f2d780d436994d31eda66be9cb79aeb3ec9)

apps/web/src/utils/skillUtils.ts
esizer: Hmm, maybe? It does not seem to lead to any errors as-is right now though :thinking:

vd1992: Since it is nullable/optional it wouldn't throw errors ever I think
Would be more about accurate documentation and the interface being thorough

esizer: I'm sorry but im confused, what is inaccurate about it right now? It describes the shape the fucniton accepts and returns. Or are you sayhing it does not? :thinking: 

It can contain more outside of the funciton, the the interface is just defining the fields it needs to execute every branch, not all potentional additional fields that could be contained in the type.

vd1992: Whether we need to extend `SkillOption` with
description?: localized string
At some point. To account for the interface being referenced in `ApplicationSkills` where description is called

To be clear
I haven't worked on any of these PRs nor looked into the codegen changes forcing us to do this. So I am guessing at what things we ought to do rather than sure about for this refactor as a whole

apps/web/src/utils/skillUtils.ts
```diff
@@ -61,22 +77,27 @@ interface InvertedExperience extends SimpleAnyExperience {
...
- * @param { Experience[] } experiences - The collection of experiences with child skills to invert
- * @returns { Skill[] } - The new collection of skills with child experiences
+ * @param experiences - The collection of experiences with child skills to invert
+ * @returns - The new collection of skills with child experiences
  */
 export function invertSkillExperienceTree(
-  experiences: Omit<Experience, "user">[],
+  experiences: ExperienceWithFullSkills[],
 ): InvertedSkillExperience[] {
   const allChildSkills = experiences.flatMap((s) => s.skills).filter(notEmpty);
```
vd1992: One last thing to note. Claude has flagged that function 
`invertSkillExperienceTree`
Is also no longer called anywhere except for the sake of testing in 
`apps/web/src/utils/skillUtils.test.ts`

So it probably should be deleted at some point. Your call as for now or later.

vd1992 (approved):
One last thing to note but not blocking. Other than that...

Great!
Looks good, thanks for bearing with me

apps/web/src/utils/skillUtils.ts
esizer: Oh cool, I would be glad to delete code :laughing: 

[remove unused function](https://github.com/GCTC-NTGC/gc-digital-talent/pull/17822/commits/f9bc409ae5ae13c30001f8a3e095a7114251296f)

PR #17824, opened 2026-08-20
 [Bug] Allow adding development programs to PAST talent nomination events

### LINKED ISSUE #17681: 🐛 Can't add a development program to a `PAST` talent nomination event
Editing a talent nomination event whose status is `PAST` (nominations closed) and trying to add a new development program (professionalization) to it always fails with a validation error, even when the chosen `CommunityDevelopmentProgram` is correctly scoped to the event's own community. This happens for **every** development program, not just misconfigured ones — the picker offers valid options and submission still rejects them.

Expected: adding a development program that legitimately belongs to the event's community should succeed for a `PAST` event, the same way it already works for an `ACTIVE` event.

### 🕵️ Details
Root cause is in [`UpdateTalentNominationEventValidator::rules()`](https://github.com/GCTC-NTGC/gc-digital-talent/blob/main/api/app/GraphQL/Validators/Mutation/UpdateTalentNominationEventValidator.php):
```php
$communityId = $eventStatus === TalentNominationEventStatus::ACTIVE->name
    ? $thisEvent->community_id
    : $this->arg('talentNominationEvent.community.connect');
```
This only falls back to the event's stored `community_id` when the status is `ACTIVE`. For any other status (`UPCOMING` or `PAST`) it instead reads `talentNominationEvent.community.connect` from the mutation input.

But `UpdateTalentEventPage.tsx` routes **both** `ACTIVE` and `PAST` events through `ActiveTalentEventForm` (grouped together as "already launched, limited fields editable"), and `ActiveTalentEventForm`'s `onSubmit` never sends a `community` field at all — by design, community can't be changed once an event launches.

So for a `PAST` event: `$eventStatus !== ACTIVE`, the validator falls to `community.connect`, which is absent from the input, giving `$communityId = null`. The `communityDevelopmentPrograms.sync.*.id` exists-rule then requires `community_id = null OR id already attached` — which no newly-added pivot row can ever satisfy. Every attempt to add a new development program to a `PAST` event fails with `COMMUNITY_DEVELOPMENT_PROGRAM_NOT_FOUND_OR_INVALID`.

Confirmed locally with throwaway PHPUnit feature tests: the same steps against a genuinely `ACTIVE` event (open in past, close in future) succeed, isolating the bug to the `PAST`-status branch specifically.

Relevant files:
- `api/app/GraphQL/Validators/Mutation/UpdateTalentNominationEventValidator.php`
- `apps/web/src/pages/TalentEvents/UpdateTalentEventPage.tsx`
- `apps/web/src/pages/TalentEvents/components/ActiveTalentEventForm.tsx`
- `api/tests/Feature/TalentNominationEventTest.php`

### 📋 Steps to Reproduce
1. As a community admin or community talent coordinator, open a talent nomination event whose close date has already passed (status `PAST`).
2. Go to edit the event and open the "Add an opportunity" dialog under development opportunities.
3. Select any development program belonging to the event's own community and save.
4. Observe that the update fails with a validation error (`COMMUNITY_DEVELOPMENT_PROGRAM_NOT_FOUND_OR_INVALID`), even though the selected program is correctly scoped to the community.

### 📸 Screenshot

Image

### 🙋‍♀️ Proposed Solution
Broaden the validator's condition to cover every status handled by `ActiveTalentEventForm` (i.e. anything that isn't `UPCOMING`) instead of only `ACTIVE`:
```php
$communityId = $eventStatus !== TalentNominationEventStatus::UPCOMING->name
    ? $thisEvent->community_id
    : $this->arg('talentNominationEvent.community.connect');
```
Add a regression test in `api/tests/Feature/TalentNominationEventTest.php` alongside the existing `ACTIVE`-status coverage (`testNewDevelopmentProgramMustBelongToCommunityOnUpdate`, `testStaleCommunityDevelopmentProgramIdDoesNotBlockUpdate`) for the `PAST`-status case.

### ✅ Acceptance Criteria
- [ ] Adding a new, community-scoped development program to a `PAST` talent nomination event succeeds.
- [ ] Adding a development program that does **not** belong to the event's community to a `PAST` event is still rejected with `COMMUNITY_DEVELOPMENT_PROGRAM_NOT_FOUND_OR_INVALID`.
- [ ] Existing `ACTIVE` and `UPCOMING` status validation behaviour is unchanged (no regressions in existing `TalentNominationEventTest` cases).
- [ ] A regression test covering the `PAST`-status case is added to `TalentNominationEventTest`.

### 🔄 Regression
Yes

### 🔗 Root Cause Pull Request
https://github.com/GCTC-NTGC/gc-digital-talent/pull/16366

bella77-69:
🤖 Resolves #17681

## 👋 Introduction

This PR fixes a bug where adding a new development program to a PAST talent nomination event would always fail with a validation error, even when the program was properly scoped to the event's community. 

## 🧪 Testing

1. Build `pnpm dev:fresh`
2. Log in as `talent-coordinator@test.com` and navigate to `admin/talent-events`
3. Open a Talent Event that its status is `PAST`
4. Go to edit talent nomination event and open the "Add an opportunity" dialog under development opportunities 
5. Select any development program belonging to the event's own community and save.
6. Verify the talent nomination event updated successfully.
7. Run tests `TalentNominationEventTest` verify they all pass

## 📸 Screenshot
image


api/app/GraphQL/Validators/Mutation/UpdateTalentNominationEventValidator.php
```diff
@@ -30,7 +30,7 @@ public function rules(): array
         $storedOpenDate = $thisEvent->open_date;
         $storedCloseDate = $thisEvent->close_date ?? null;
 
-        $communityId = $eventStatus === TalentNominationEventStatus::ACTIVE->name
+        $communityId = $eventStatus !== TalentNominationEventStatus::UPCOMING->name
```
petertgiles: I found this bit hard to follow before.  I know this was the proposed solution in the issue but I still find it hard to follow.  Maybe we could change this to pattern matching and make the three options explicit?  And maybe add a comment to explain the options?

api/tests/Feature/TalentNominationEventTest.php
```diff
@@ -425,6 +425,95 @@ public function testNewDevelopmentProgramMustBelongToCommunityOnUpdate()
             ->assertGraphQLValidationError('talentNominationEvent.communityDevelopmentPrograms.sync.0.id', ErrorCode::COMMUNITY_DEVELOPMENT_PROGRAM_NOT_FOUND_OR_INVALID->name);
     }
 
+    // #17681 - a community-scoped development program can be added to a PAST event
+    public function testNewDevelopmentProgramCanBeAddedToPastEvent()
```
petertgiles, line 429: Nice tests!  😄

petertgiles (commented):
Works great!  Small style suggestion but nothing blocking.

petertgiles (approved):

api/app/GraphQL/Validators/Mutation/UpdateTalentNominationEventValidator.php
bella77-69: Thanks @petertgiles  😁 I have updated it and I will wait for you to do review and a thumbs up before I merge.

petertgiles: Nice, I like that much better.  :heart:

